PR #2587, opened 2023-09-30
[speed up CI tests suite] Skip slow tests as default

johnnv1:
This patch:

1. [x] config pytest to not run slow marked tests by default
1. [x] add a CI job to run the slow tests only on Ubuntu
1. [x] add [timeout](https://github.com/pytest-dev/pytest-timeout) plugin (as dev dependency) to fail tests that take longer than 5 seconds to run (This is the only way that I found to prevent new tests from being added to the test suite without slow markers)
    a. The timeout plugin is configured by default to 5 seconds into the tests workflow, so if anyone wants to reproduce it locally, they need to use `--timeout=5` as a parameter when running pytests. I don't think makes sense to be a default config, since the time that matters is the time taken on CI, not on each person's machine.
1. [in progress] add slow markers into the tests
1. [x] config CI to not run dynamo tests as default behavior (the CI always timeout the first dynamo test)

### DIFF
--- a/.github/workflows/pr_test_cpu.yml
+++ b/.github/workflows/pr_test_cpu.yml
@@ -14,27 +14,27 @@ concurrency:
 
 jobs:
   tests-cpu:
+    # This job run just tests not marked as slow, and desselect the dynamo tests
     strategy:
-      fail-fast: false  # TODO: enable it when CI be stable (when drop support to torch 1.9.1)
+      fail-fast: true
       matrix:
         os: ['Ubuntu-latest', 'Windows-latest', 'MacOS-latest']
+        pytorch-dtype: ['float32', 'float64']
+        exclude:
+          - os: 'Windows-latest'
+            pytorch-dtype: 'float64'
+          - os: 'MacOS-latest'
+            pytorch-dtype: 'float64'
+
     uses: ./.github/workflows/tests.yml
     with:
       os: ${{ matrix.os }}
       python-version: '["3.8", "3.10"]'
       pytorch-version: '["1.9.1", "2.0.1"]'
-      pytorch-dtype: 'float32'
-
-  tests-cpu-fp64:
-    strategy:
-      fail-fast: true
-    uses: ./.github/workflows/tests.yml
-    with:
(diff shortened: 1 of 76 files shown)


.github/workflows/scheduled_test_cpu.yml
```diff
@@ -49,6 +49,17 @@ jobs:
...
+  test-slow:
+    uses: ./.github/workflows/tests.yml
+    strategy:
+      matrix:
+        pytorch-dtype: ['float32', 'float64']
+    with:
+      python-version: '["3.10"]'
+      pytorch-version: '["2.0.1"]'
+      pytorch-dtype: ${{ matrix.pytorch-dtype }}
+      pytest-extra: '--runslow'
```
johnnv1, line 61: maybe this should just be the option on `tests-cpu-ubuntu:` job to run all test suite into all torch supported versions

.github/workflows/pr_test_cpu.yml
```diff
@@ -42,6 +42,18 @@ jobs:
       pytorch-dtype: 'float32,float64'
       coverage: true
       continue-on-error: true
+      pytest-extra: '--runslow'
+
+  test-slow:
```
johnnv1, line 47: this isn't a necessary job, because the `coverage` already will run within the same parameters but with coverage report

pyproject.toml
```diff
@@ -167,7 +167,8 @@ requires = [
   markers =[
     "grad: mark a test as gradcheck test",
     "jit: mark a test as torchscript test",
-    "nn: mark a test as module test"
+    "nn: mark a test as module test",
```
edgarriba, line 170: @johnnv1 as a meta task to remove old `grad` / `jit` / `nn` markers ?

edgarriba (approved):

edgarriba:
@ducha-aiki can you check the ransac conflict ?

pyproject.toml
johnnv1: yeah, I'll open an issue for, and can make it later

PR #2588, opened 2023-10-01
Add `torch.jit.script` support for `warp_affine`

### LINKED ISSUE #2583: Usage of `|` in type annotation causes `torch.jit.script` to not work
### Describe the bug

`torch.jit.script` on a model that calls `warp_affine` within `forward` function fails with following error:
```
RuntimeError: 
Expression of type | cannot be used in a type expression:
  File "/usr/local/lib/python3.8/dist-packages/kornia/geometry/conversions.py", line 1015
def normal_transform_pixel(
    height: int, width: int, eps: float = 1e-14, device: torch.device | None = None, dtype: torch.dtype | None = None
                                                         ~~~~~~~~~~~~~~~~~~~ <--- HERE
) -> Tensor:
    r"""Compute the normalization matrix from image size in pixels to [-1, 1].
'normal_transform_pixel' is being compiled since it was called from 'normalize_homography'
  File "/usr/local/lib/python3.8/dist-packages/kornia/geometry/conversions.py", line 1004

    # compute the transformation pixel/norm for src/dst
    src_norm_trans_src_pix: Tensor = normal_transform_pixel(src_h, src_w).to(dst_pix_trans_src_pix)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE

    src_pix_trans_src_norm = _torch_inverse_cast(src_norm_trans_src_pix)
'normalize_homography' is being compiled since it was called from 'warp_affine'
  File "/usr/local/lib/python3.8/dist-packages/kornia/geometry/transform/imgwarp.py", line 201
    # we generate a 3x3 transformation matrix from 2x3 affine
    M_3x3: Tensor = convert_affinematrix_to_homography(M)
    dst_norm_trans_src_norm: Tensor = normalize_homography(M_3x3, (H, W), dsize)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE

    # src_norm_trans_dst_norm = torch.inverse(dst_norm_trans_src_norm)
'warp_affine' is being compiled since it was called from 'DummyModel.forward'
  File "tmp.py", line 6
    def forward(self, x, y):
        return T.warp_affine(
               ~~~~~~~~~~~~~~
            x, y,
            ~~~~~
            dsize=(1920, 1080),
            ~~~~~~~~~~~~~~~~~~~
            align_corners=False,
            ~~~~~~~~~~~~~~~~~~~ <--- HERE
        )
```

### Reproduction steps

```bash
import torch
from kornia.geometry import transform as T

class DummyModel(torch.nn.Module):
    def forward(self, x, y):
        return T.warp_affine(
            x, y,
            dsize=(1920, 1080),
            align_corners=False,
        )

torch.jit.script(DummyModel())
```

### Expected behavior

`torch.jit.script` compiles.

### Environment

```shell
python collect_env.py
```
- PyTorch Version (e.g., 1.0): 2.0.1+cu117 (also 2.0.1)
- OS (e.g., Linux): Linux
- How you installed PyTorch (`conda`, `pip`, source): pip
- Python version: 3.8 (also happening in 3.9)
- CUDA/cuDNN version: 11.7
- GPU models and configuration: Tested on 3080, 4090 and CPU

### Additional context

`Optional[torch.device]` will work with `torch.jit.script`. I understand that `torch.device | None` is cleaner code, and more pythonic, but it would be great if this library supported `jit`

balbok0:
#### Changes
* Changes ruff to allow usage of `Optional` instead of `| None`
* Add temp.onnx to .gitignore (since some tests output that)
* Add `torch.jit.script` test for `warp_affine`
* Add `torch.onnx.export` test for `warp_affine` (currently failing due to op not supported)

* Some other changes that were forced due to use of `pre-commit`. Please let me know if these should not be included, and in such a case if it is ok to not run `pre-commit`.

Fixes #2583

#### Type of change
- [ ] 📚  Documentation Update
- [x] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ kornia.egg-info/
 *ipynb_checkpoints*
 docs/source/_static/*jpg
 *DS_Store
+
+# onnx tests create temp.onnx file
+temp.onnx
--- a/.gitignore
+++ b/.gitignore
@@ -22,4 +22,4 @@ docs/source/_static/*jpg
 *DS_Store
 
 # onnx tests create temp.onnx file
-temp.onnx
+*.onnx
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,6 +139,7 @@ requires = [
   ignore = [
     "PLW2901", # Allow overwritten values on loops
     "PLR0915", # Allow condition check in list comprehension
+    "UP007", # Prefer Optional[], Union[] over | due to torch jit scripting
   ]
 
   [tool.ruff.isort]
--- a/test/geometry/transform/test_imgwarp.py
+++ b/test/geometry/transform/test_imgwarp.py
@@ -11,6 +11,16 @@ from kornia.utils._compat import torch_version, torch_version_lt
 from kornia.utils.helpers import _torch_inverse_cast
 
 
+class DummyNNModule(torch.nn.Module):
+    def __init__(self, h: int, w: int, align_corners: bool, padding_mode: str):
+        super().__init__()
+        self.h = h
+        self.w = w
+
+    def forward(self, x, y):
+        return kornia.geometry.transform.warp_affine(x, y, dsize=(self.h, self.w), align_corners=False)
+
+
 class TestGetPerspectiveTransform:
     @pytest.mark.parametrize("batch_size", [1, 2, 5])
     def test_smoke(self, device, dtype, batch_size):
@@ -267,6 +277,27 @@ class TestWarpAffine:
         img_a = kornia.geometry.warp_affine(img_b, aff_ab, (h, w), padding_mode="fill", fill_value=fill_value)
 
         assert_close(img_a[:, :, :1, :1].squeeze(), fill_value.squeeze())
+
+    @pytest.mark.parametrize("align_corners", (True, False))
+    @pytest.mark.parametrize("padding_mode", ("zeros", "fill"))
+    def test_jit_script(self, align_corners, padding_mode):
+        net = DummyNNModule(3, 4, align_corners, padding_mode)
+        net = torch.jit.script(net)
+        # Assert compilation doesn't fail
+
+    @pytest.mark.parametrize("align_corners", (True, False))
+    @pytest.mark.parametrize("padding_mode", ("zeros", "fill"))
+    @pytest.mark.xfail(reason="aten::linalg_inv is not yet supported in ONNX opset version 14.")
+    def test_onnx_export(self, device, dtype, align_corners, padding_mode):
+        offset = 1.0
+        h, w = 3, 4
+        aff_ab = torch.eye(2, 3, device=device, dtype=dtype)[None]
+        aff_ab[..., -1] += offset
+
+        img_b = torch.arange(float(3 * h * w), device=device, dtype=dtype).view(1, 3, h, w)
+
+        net = DummyNNModule(h, w, align_corners, padding_mode).to(device)
+        torch.onnx.export(net, (img_b, aff_ab), "temp.onnx", export_params=True)
 
 
 class TestWarpPerspective:
--- a/test/geometry/transform/test_imgwarp.py
+++ b/test/geometry/transform/test_imgwarp.py
@@ -280,10 +280,18 @@ class TestWarpAffine:
 
     @pytest.mark.parametrize("align_corners", (True, False))
     @pytest.mark.parametrize("padding_mode", ("zeros", "fill"))
-    def test_jit_script(self, align_corners, padding_mode):
+    def test_jit_script(self, device, dtype, align_corners, padding_mode):
+        offset = 1.0
+        h, w = 3, 4
+        aff_ab = torch.eye(2, 3, device=device, dtype=dtype)[None]
+        aff_ab[..., -1] += offset
+
+        img_b = torch.arange(float(3 * h * w), device=device, dtype=dtype).view(1, 3, h, w)
         net = DummyNNModule(3, 4, align_corners, padding_mode)
-        net = torch.jit.script(net)
-        # Assert compilation doesn't fail
+        script_net = torch.jit.script(net)
+
+        assert isinstance(script_net, torch.jit.ScriptModule)
+        assert_close(script_net(img_b, aff_ab), net(img_b, aff_ab))
 
     @pytest.mark.parametrize("align_corners", (True, False))
     @pytest.mark.parametrize("padding_mode", ("zeros", "fill"))
--- a/test/geometry/transform/test_imgwarp.py
+++ b/test/geometry/transform/test_imgwarp.py
@@ -293,20 +293,6 @@ class TestWarpAffine:
         assert isinstance(script_net, torch.jit.ScriptModule)
         assert_close(script_net(img_b, aff_ab), net(img_b, aff_ab))
 
-    @pytest.mark.parametrize("align_corners", (True, False))
-    @pytest.mark.parametrize("padding_mode", ("zeros", "fill"))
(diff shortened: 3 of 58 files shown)


pyproject.toml
```diff
@@ -139,6 +139,7 @@ requires = [
   ignore = [
     "PLW2901", # Allow overwritten values on loops
     "PLR0915", # Allow condition check in list comprehension
+    "UP007", # Prefer Optional[], Union[] over | due to torch jit scripting
```
johnnv1, line 142: is Jit supporting the others annotations from `__future__ import annotations`?

test/geometry/transform/test_imgwarp.py
```diff
@@ -268,6 +278,27 @@ def test_fill_padding_channels(self, device, dtype, num_channels):
 
         assert_close(img_a[:, :, :1, :1].squeeze(), fill_value.squeeze())
 
+    @pytest.mark.parametrize("align_corners", (True, False))
+    @pytest.mark.parametrize("padding_mode", ("zeros", "fill"))
+    def test_jit_script(self, align_corners, padding_mode):
+        net = DummyNNModule(3, 4, align_corners, padding_mode)
+        net = torch.jit.script(net)
+        # Assert compilation doesn't fail
```
johnnv1, line 286: you can assert jit results vs no jitted

pyproject.toml
balbok0: Most of them yes, but not all of them. See: https://pytorch.org/docs/stable/jit_language_reference.html Though this page seems outdated, because I definitely had success with overloading functions in scripting.

Outside of `from __future__ import annotations`, the most notable thing missing is lack of support for `typing.Sequence`. Switching that to `Tuple`, `List` or some combination of both is an option, but that drops support for custom classes implementing sequence. Since, this is a possibly breaking change, and as mentioned in the issue, I'm not an expert in kornia ecosystem I decided to stop here.

test/geometry/transform/test_imgwarp.py
balbok0: Done

pyproject.toml
johnnv1: i see, I didn't expect jit to work with plain typing (`dict`, `list`, etc) since they stopped adding new typing options (or stopped to update the docs) -- i thought it was only working with typings `from typing import ...`

test/geometry/transform/test_imgwarp.py
```diff
@@ -268,6 +278,35 @@ def test_fill_padding_channels(self, device, dtype, num_channels):
...
+        img_b = torch.arange(float(3 * h * w), device=device, dtype=dtype).view(1, 3, h, w)
+        net = DummyNNModule(3, 4, align_corners, padding_mode)
+        script_net = torch.jit.script(net)
+
+        assert isinstance(script_net, torch.jit.ScriptModule)
+        assert_close(script_net(img_b, aff_ab), net(img_b, aff_ab))
+
+    @pytest.mark.parametrize("align_corners", (True, False))
+    @pytest.mark.parametrize("padding_mode", ("zeros", "fill"))
+    @pytest.mark.xfail(reason="aten::linalg_inv is not yet supported in ONNX opset version 14.")
```
johnnv1, line 298: we can't add a condition to mark xfail just for this version?

johnnv1 (commented):
btw, it's pretty awesome that ruff is able to revert the typing from `|` to Union and Optional

test/geometry/transform/test_imgwarp.py
balbok0: 14 is the default option for torch 2.0.1, which is why I used it. The same is true for opset version 19 (most recent). I can change the comment to reflect that, or parametrize the test if you would like, but I don't think this test should pass in current state of things. So maybe it should be removed?

.gitignore
```diff
@@ -20,3 +20,6 @@ kornia.egg-info/
 *ipynb_checkpoints*
 docs/source/_static/*jpg
 *DS_Store
+
+# onnx tests create temp.onnx file
+temp.onnx
```
edgarriba, line 25:
```suggestion
*.onnx
```

test/geometry/transform/test_imgwarp.py
```diff
@@ -11,6 +11,16 @@
 from kornia.utils.helpers import _torch_inverse_cast
 
 
+class DummyNNModule(torch.nn.Module):
```
edgarriba, line 14: @johnnv1 i think as post meta task from this we could exploring to prototype an AutoModule that basically given a function parsed via `inspect` to automatically generate Modules. This might requires creating `InputTensor` / `OutputTensor`  alias types.

test/geometry/transform/test_imgwarp.py
edgarriba: as an idea, can we add something at the high level api, like a decorator to add in some sort of registry that includes min supported onnx version ?

test/geometry/transform/test_imgwarp.py
```diff
@@ -268,6 +278,35 @@ def test_fill_padding_channels(self, device, dtype, num_channels):
 
         assert_close(img_a[:, :, :1, :1].squeeze(), fill_value.squeeze())
 
+    @pytest.mark.parametrize("align_corners", (True, False))
+    @pytest.mark.parametrize("padding_mode", ("zeros", "fill"))
+    def test_jit_script(self, device, dtype, align_corners, padding_mode):
```
edgarriba, line 283: is it relevant to keep torchscript test is we go for onnx ? underhood it traces in any case

test/geometry/transform/test_imgwarp.py
edgarriba: eventually to avoid this is issue around not supported `linalg_inv`:
- accept the inverted affine transform
- (or) implement closed solution of inversion for affine transform /cc @ducha-aiki

edgarriba (approved):

test/geometry/transform/test_imgwarp.py
johnnv1: if the test will not pass for the current versions, I would say to skip it instead of xfial

test/geometry/transform/test_imgwarp.py
johnnv1: we can work on it

test/geometry/transform/test_imgwarp.py
balbok0: done. I removed the test.

test/geometry/transform/test_imgwarp.py
balbok0: I consider onnx to be one of the end states of scripting pipeline of torch (tensorrt is another widely used format that comes to mind), so I think it's worth keeping the test for an intermediate torch.jit.script product.

Again, that's coming from me as a user of kornia rather than a person who has insight into the project.

balbok0:
Let me know if there is anything else needed to get this merged in. I believe all of the remaining threads are outstanding items for the project in general, rather than this PR (ofc, if one of these should be included here, ping me)

I appreciate fast responsiveness, and great feedback! Made me feel welcome to the project!

test/geometry/transform/test_imgwarp.py
edgarriba: sure, it's just that we have had such a base experience with torchscript along the years and prioritised to onnx recently  as we seen the most claimed format but always open to community needs.

edgarriba (approved):

PR #2593, opened 2023-10-01
drop `cv2` as dev dependency

### LINKED ISSUE #2592: Drop `cv2` from dev dependencies
## 🚀 Feature
The cv2 package is just used in one place at tests. So, we can figure out how to test the case without it
https://github.com/kornia/kornia/blob/216aa9d2f10f9300bffe952b1583f03b7f833be5/test/io/test_io_image.py#L4

## Motivation
reduce the number of dependencies at kornia

johnnv1:
#### Changes

Fixes #2592 

#### Type of change

- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/test/io/test_io_image.py
+++ b/test/io/test_io_image.py
@@ -1,9 +1,10 @@
+import io
 import sys
 from pathlib import Path
 
-import cv2
 import numpy as np
 import pytest
+import requests
 import torch
 
 from kornia.core import Tensor
@@ -28,6 +29,40 @@ def create_random_img8_torch(height: int, width: int, channels: int, device=None
     return (torch.rand(channels, height, width, device=device) * 255).to(torch.uint8)
 
 
+def _download_image(url: str, filename: str = "") -> str:
+    # TODO: move this to testing
+
+    filename = url.split("/")[-1] if len(filename) == 0 else filename
+    # Download
+    bytesio = io.BytesIO(requests.get(url, timeout=60).content)
+    # Save file
+    with open(filename, "wb") as outfile:
+        outfile.write(bytesio.getbuffer())
+
+    return filename
+
+
+@pytest.fixture(scope='class')
+def png_image(tmp_path_factory):
+    url = "https://github.com/kornia/data/raw/main/simba.png"
+    filename = tmp_path_factory.mktemp("data") / "image.png"
+    filename = _download_image(url, str(filename))
+    return filename
+
+
+@pytest.fixture(scope='class')
+def jpg_image(tmp_path_factory):
+    url = "https://github.com/kornia/data/raw/main/crowd.jpg"
+    filename = tmp_path_factory.mktemp("data") / "image.jpg"
+    filename = _download_image(url, str(filename))
+    return filename
+
+
+@pytest.fixture
+def images_fn(png_image, jpg_image):
+    return {'png': png_image, 'jpg': jpg_image}
+
+
 @pytest.mark.skipif(not available_package(), reason="kornia_rs only supports python >=3.7 and pt >= 1.10.0")
 class TestIoImage:
     def test_smoke(self, tmp_path: Path) -> None:
@@ -45,12 +80,11 @@ class TestIoImage:
         assert img_th.shape[1:] == (height, width)
         assert str(img_th.device) == "cpu"
 
-    def test_device(self, device, tmp_path: Path) -> None:
+    def test_device(self, device, png_image) -> None:
         height, width = 4, 5
-        img_np: np.ndarray = create_random_img8(height, width, 3)
+        create_random_img8(height, width, 3)
 
-        file_path = tmp_path / "image.png"
-        cv2.imwrite(str(file_path), img_np)
+        file_path = Path(png_image)
 
         assert file_path.is_file()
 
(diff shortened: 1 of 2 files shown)


ducha-aiki (approved):

test/io/test_io_image.py
```diff
@@ -45,12 +80,11 @@ def test_smoke(self, tmp_path: Path) -> None:
         assert img_th.shape[1:] == (height, width)
         assert str(img_th.device) == "cpu"
 
-    def test_device(self, device, tmp_path: Path) -> None:
+    def test_device(self, device, png_image) -> None:
```
edgarriba, line 83:
```suggestion
    def test_device(self, device, png_image: Path) -> None:
```

test/io/test_io_image.py
```diff
@@ -45,12 +80,11 @@ def test_smoke(self, tmp_path: Path) -> None:
...
         height, width = 4, 5
-        img_np: np.ndarray = create_random_img8(height, width, 3)
+        create_random_img8(height, width, 3)
```
edgarriba, line 85:
```suggestion
```
?

test/io/test_io_image.py
```diff
@@ -28,6 +29,40 @@ def create_random_img8_torch(height: int, width: int, channels: int, device=None
...
+    filename = _download_image(url, str(filename))
+    return filename
+
+
+@pytest.fixture(scope='class')
+def jpg_image(tmp_path_factory):
+    url = "https://github.com/kornia/data/raw/main/crowd.jpg"
+    filename = tmp_path_factory.mktemp("data") / "image.jpg"
+    filename = _download_image(url, str(filename))
+    return filename
```
edgarriba, line 58: return as Path ?

test/io/test_io_image.py
```diff
@@ -28,6 +29,40 @@ def create_random_img8_torch(height: int, width: int, channels: int, device=None
...
+    with open(filename, "wb") as outfile:
+        outfile.write(bytesio.getbuffer())
+
+    return filename
+
+
+@pytest.fixture(scope='class')
+def png_image(tmp_path_factory):
+    url = "https://github.com/kornia/data/raw/main/simba.png"
+    filename = tmp_path_factory.mktemp("data") / "image.png"
```
edgarriba, line 48: why do you need to create `data` ? is not tmp_path_factory a new path every time the class implementing the fixture ? actually maybe we need to scope at session level

johnnv1: from https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-factory-fixture
> The tmp_path_factory is a session-scoped fixture which can be used to create arbitrary temporary directories from any other fixture or test.

the `tmp_path` is a new temp directory for each test itself

johnnv1: The idea here is to just download the PNG and JPG image one time per class or session, so it doesn't get heavy/slow

test/io/test_io_image.py
johnnv1: yep, this way we can load from disk within kornia api

test/io/test_io_image.py
```diff
@@ -75,12 +108,11 @@ def test_device(self, device, tmp_path: Path) -> None:
             (3, ImageLoadType.RGB32, torch.float32, 3),
         ],
     )
-    def test_load_image(self, tmp_path: Path, ext, channels, load_type, expected_type, expected_channels):
+    def test_load_image(self, images_fn, ext, channels, load_type, expected_type, expected_channels):
         height, width = 4, 5
-        img_np: np.ndarray = create_random_img8(height, width, channels)
+        create_random_img8(height, width, channels)
```
johnnv1, line 113:
```suggestion
```

test/io/test_io_image.py
edgarriba: now it returns a `str` not `Path`

test/io/test_io_image.py
```diff
@@ -28,6 +29,40 @@ def create_random_img8_torch(height: int, width: int, channels: int, device=None
...
+    # Download
+    bytesio = io.BytesIO(requests.get(url, timeout=60).content)
+    # Save file
+    with open(filename, "wb") as outfile:
+        outfile.write(bytesio.getbuffer())
+
+    return filename
+
+
+@pytest.fixture(scope='class')
```
edgarriba, line 45:
```suggestion
@pytest.fixture(scope='serssion')
```

test/io/test_io_image.py
```diff
@@ -28,6 +29,40 @@ def create_random_img8_torch(height: int, width: int, channels: int, device=None
...
+def png_image(tmp_path_factory):
+    url = "https://github.com/kornia/data/raw/main/simba.png"
+    filename = tmp_path_factory.mktemp("data") / "image.png"
+    filename = _download_image(url, str(filename))
+    return filename
+
+
+@pytest.fixture(scope='class')
```
edgarriba, line 53:
```suggestion
@pytest.fixture(scope='session')
```

edgarriba (approved):

test/io/test_io_image.py
johnnv1: true, got it

Issue #1698, opened 2022-05-04
Affine shear matrix not properly computed around center point


grosssoftware:
This fix in this pull request: https://github.com/kornia/kornia/pull/612 isn't quite right. There's a negative sign missing in m23 when calculating the shear_mat. It should be:
```
shear_mat = torch.stack(
    [ones, -sx_tan, sx_tan * y, -sy_tan, ones + sx_tan * sy_tan, sy_tan * (x - sx_tan * y)], dim=-1  # type: ignore
).view(-1, 2, 3)
```

_Originally posted by @grosssoftware in https://github.com/kornia/kornia/issues/612#issuecomment-1117798930_

PR #2608, opened 2023-10-04
Fix incorrect computation of affine shear matrix around center point

Vihtoriaaa:
#### Changes

- An incorrect mathematical calculation of the affine shear matrix around center point was fixed, since `m23` was missing a negative sign when calculating `shear_mat`. 
- The `get_shear_matrix2d` test has been updated to work correctly due to the changes made.  

Fixes [#1698](https://github.com/kornia/kornia/issues/1698)

#### Type of change

- [x] 📚  Documentation Update
- [x] 🧪 Tests Cases

#### Checklist


Vihtoriaaa:
@johnnv1 Hi! Can you please review my PR if you have time to check whether everything is ok and it can me merged?:) thanks!

kornia/geometry/transform/imgwarp.py
```diff
@@ -660,9 +660,9 @@ def get_shear_matrix2d(center: Tensor, sx: Optional[Tensor] = None, sy: Optional
     sx_tan = torch.tan(sx)
     sy_tan = torch.tan(sy)
     ones = torch.ones_like(sx)
-    shear_mat = stack([ones, -sx_tan, sx_tan * y, -sy_tan, ones + sx_tan * sy_tan, sy_tan * (sx_tan * y + x)], -1).view(
-        -1, 2, 3
-    )
+    shear_mat = stack(
+        [ones, -sx_tan, sx_tan * y, -sy_tan, ones + sx_tan * sy_tan, sy_tan * (x - sx_tan * y)], dim=-1
```
edgarriba, line 664: if this math is incorrect -- can you add an additional unit test showing the fix? /cc @ducha-aiki

edgarriba (changes requested):

kornia/geometry/transform/imgwarp.py
Vihtoriaaa: @edgarriba Hi! I am not quite sure how to show that the bug is fixed with additional unit test - can you please give me a hint regarding that? Should I compare the 'old math' and 'new math' to see which one is the correct one? or what the best way would be to do this ? thanks:)

Vihtoriaaa: so would it be ok if I test whether expected (controlled matrix with numbers) is the same as the matrix that is computed with changed get_shear_matrix2d function?

edgarriba: that would work

Vihtoriaaa: @edgarriba Added a unit test, let me know if it works :) thanks!

edgarriba (approved):

PR #2611, opened 2023-10-06
Deprecated decorator added

### LINKED ISSUE #2560: Add the deprecated decorator in `depth_to_3d`
we should add the deprecated decorator there: https://github.com/kornia/kornia/blob/a212b4d0746b15a4e56a34e72ddee9b320284fb9/kornia/utils/helpers.py#L70

_Originally posted by @johnnv1 in https://github.com/kornia/kornia/pull/2557#discussion_r1333672167_

karakumm:
#### Changes
Added the deprecated decorator in `depth_to_3d`

Fixes #2560 

#### Type of change

#### Checklist

### DIFF
--- a/kornia/geometry/depth.py
+++ b/kornia/geometry/depth.py
@@ -10,6 +10,7 @@ from kornia.core import Module, Tensor, tensor
 from kornia.core.check import KORNIA_CHECK, KORNIA_CHECK_IS_TENSOR, KORNIA_CHECK_SHAPE
 from kornia.filters.sobel import spatial_gradient
 from kornia.utils import create_meshgrid
+from kornia.utils.helpers import deprecated
 
 from .camera import PinholeCamera, cam2pixel, pixel2cam, project_points, unproject_points
 from .conversions import normalize_pixel_coordinates, normalize_points_with_intrinsics
@@ -109,7 +110,7 @@ def depth_to_3d_v2(
 
 # NOTE: this function should replace the old depth_to_3d
 
-
+@deprecated(replace_with="depth_to_3d_v2")
 def depth_to_3d(depth: Tensor, camera_matrix: Tensor, normalize_points: bool = False) -> Tensor:
     """Compute a 3d point per pixel given its depth value and the camera intrinsics.
 
--- a/kornia/geometry/depth.py
+++ b/kornia/geometry/depth.py
@@ -110,6 +110,7 @@ def depth_to_3d_v2(
 
 # NOTE: this function should replace the old depth_to_3d
 
+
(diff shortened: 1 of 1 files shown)


kornia/geometry/depth.py
```diff
@@ -110,6 +111,7 @@ def depth_to_3d_v2(
 # NOTE: this function should replace the old depth_to_3d
 
 
+@deprecated(replace_with="depth_to_3d_v2")
```
johnnv1, line 114:
```suggestion
@deprecated(replace_with="depth_to_3d_v2", version='0.7.0')
```

johnnv1 (approved):
Thanks for the patch

karakumm:
@johnnv1 two tests are failing here but I am not sure how they are related to the changes I've made:( what do you think? thanks in advance!

johnnv1:
> @johnnv1 two tests are failing here but I am not sure how they are related to the changes I've made:( what do you think? thanks in advance!

They are unrelated to your changes, this will be merged later by Edgar or Dmytro

kornia/geometry/depth.py
```diff
@@ -110,6 +111,7 @@ def depth_to_3d_v2(
 # NOTE: this function should replace the old depth_to_3d
 
 
+@deprecated(replace_with="depth_to_3d_v2", version='0.7.0')
 def depth_to_3d(depth: Tensor, camera_matrix: Tensor, normalize_points: bool = False) -> Tensor:
```
edgarriba, line 115: @johnnv1 i'm not sure if this deprecated decorator (or the message) is the right one. Ideally we want to preserve the same function signature but update the implementation of "v1" with "v2".

maybe something like this:
```python
@deprecated(extra_reason="the implementation with be replaced with depth_to_3d_v2", version='0.7.0')
def depth_to_3d(...)
```
and

```python
@deprecated(replace_with="use depth_to_3d", version='0.7.0')
def depth_to_3d_v2(...)
```

johnnv1: can be it, it's kinda weird to have deprecated messages in the two functions -- but I think it should works

edgarriba: another more practical way is to just simply rename `depth_to_3d_v2` -> `depth_to_3d` which will make backward compatible the existing functionality in case the grid is not provided

johnnv1: well, if it's backward compatible, +1 from my side to just replace it

johnnv1:
Closing in favor of https://github.com/kornia/kornia/pull/2636 -- just added an extra reason on top of your contribution @karakumm, thanks again for the patch

PR #2617, opened 2023-10-09
xfail TestRandAugment::test_transform_mat based in torch version

johnnv1:
This test looks like randomness failing on the scheduled CI for some torch versions

image

### DIFF
--- a/test/augmentation/test_auto_operation.py
+++ b/test/augmentation/test_auto_operation.py
@@ -12,6 +12,7 @@ from kornia.augmentation.auto.trivial_augment import TrivialAugment
 from kornia.augmentation.container import AugmentationSequential
 from kornia.geometry.bbox import bbox_to_mask
 from kornia.testing import assert_close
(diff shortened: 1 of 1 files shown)


test/augmentation/test_auto_operation.py
```diff
@@ -109,6 +110,9 @@ def test_smoke(self, policy):
         in_tensor = torch.rand(10, 3, 50, 50, requires_grad=True)
         aug(in_tensor)
 
+    @pytest.mark.xfail(
+        torch_version() in {'1.10.2', '1.11.0', '1.12.1', '1.13.1'}, 'randomness failing into some torch versions'
```
johnnv1, line 114: now that i see it's just on fp64, should we consider it just for fp64?

edgarriba (approved):

test/augmentation/test_auto_operation.py
edgarriba: Yes

PR #2650, opened 2023-11-02
Allow CombineTensorPatches to work with overlapping patches

ashnair1:
#### Changes

Fixes #2513

Previously `CombineTensorPatches` only worked if stride == window_size. This PR lifts this restriction and allows users to combine overlapping patches. 

Breaking changes:
1. Padding can no longer be a tuple of 4 ints. It can only be an int or a tuple of 2 ints.
2. Stride is a required argument for `CombineTensorPatches`

Corresponding tutorial has been updated in kornia/tutorials#69

#### Type of change

- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/kornia/contrib/extract_patches.py
+++ b/kornia/contrib/extract_patches.py
@@ -249,8 +249,11 @@ def combine_tensor_patches(
     ones = torch.ones(patches.shape[0], patches.shape[2], original_size[0], original_size[1])
     patches = patches.permute(0, 2, 3, 4, 1)
     patches = patches.reshape(patches.shape[0], -1, patches.shape[-1])
-    dtype = patches.dtype
-    patches = patches.double()
+    int_flag = 0
+    if not torch.is_floating_point(patches):
+        int_flag = 1
+        dtype = patches.dtype
+        patches = patches.float()
 
     # Calculate normalization map
     unfold_ones = F.unfold(ones, kernel_size=window_size, stride=stride, padding=unpadding)
@@ -263,7 +266,8 @@ def combine_tensor_patches(
     )
     # Remove satuation effect due to multiple summations
     restored_tensor = saturated_restored_tensor / (norm_map + eps)
-    restored_tensor = restored_tensor.to(dtype)
+    if int_flag:
+        restored_tensor = restored_tensor.to(dtype)
     return restored_tensor
 
 
--- a/kornia/contrib/extract_patches.py
+++ b/kornia/contrib/extract_patches.py
@@ -38,8 +38,14 @@ def compute_padding(
     original_size = cast(Tuple[int, int], _pair(original_size))
     window_size = cast(Tuple[int, int], _pair(window_size))
 
-    padh = math.ceil((window_size[0] - original_size[0] % window_size[0]) / 2)
-    padw = math.ceil((window_size[1] - original_size[1] % window_size[1]) / 2)
+    if original_size[0] % window_size[0] != 0:
+        padh = math.ceil((window_size[0] - original_size[0] % window_size[0]) / 2)
+    else:
+        padh = 0
+    if original_size[1] % window_size[1] != 0:
+        padw = math.ceil((window_size[1] - original_size[1] % window_size[1]) / 2)
+    else:
+        padw = 0
 
     return (padh, padw)
 
--- a/kornia/contrib/extract_patches.py
+++ b/kornia/contrib/extract_patches.py
@@ -246,7 +246,14 @@ def combine_tensor_patches(
     if len(unpadding) != 2:
         raise AssertionError("Unpadding must be either an int or a tuple of two ints")
 
-    ones = torch.ones(patches.shape[0], patches.shape[2], original_size[0], original_size[1])
+    ones = torch.ones(
+        patches.shape[0],
+        patches.shape[2],
+        original_size[0],
+        original_size[1],
+        device=patches.device,
+        dtype=patches.dtype,
+    )
     patches = patches.permute(0, 2, 3, 4, 1)
     patches = patches.reshape(patches.shape[0], -1, patches.shape[-1])
     int_flag = 0
@@ -254,6 +261,7 @@ def combine_tensor_patches(
         int_flag = 1
         dtype = patches.dtype
         patches = patches.float()
+        ones = ones.float()
 
     # Calculate normalization map
     unfold_ones = F.unfold(ones, kernel_size=window_size, stride=stride, padding=unpadding)
--- a/kornia/contrib/extract_patches.py
+++ b/kornia/contrib/extract_patches.py
@@ -207,7 +207,7 @@ def combine_tensor_patches(
     window_size: Union[int, Tuple[int, int]],
     stride: Union[int, Tuple[int, int]],
     unpadding: Union[int, Tuple[int, int]] = 0,
-    eps: float = 1e-8
+    eps: float = 1e-8,
 ) -> Tensor:
     r"""Restore input from patches.
 
--- a/kornia/contrib/extract_patches.py
+++ b/kornia/contrib/extract_patches.py
@@ -248,6 +248,8 @@ def combine_tensor_patches(
     ones = torch.ones(patches.shape[0], patches.shape[2], original_size[0], original_size[1])
     patches = patches.permute(0, 2, 3, 4, 1)
     patches = patches.reshape(patches.shape[0], -1, patches.shape[-1])
+    dtype = patches.dtype
+    patches = patches.double()
 
     # Calculate normalization map
     unfold_ones = F.unfold(ones, kernel_size=window_size, stride=stride, padding=unpadding)
@@ -260,6 +262,7 @@ def combine_tensor_patches(
     )
     # Remove satuation effect due to multiple summations
     restored_tensor = saturated_restored_tensor / (norm_map + 1e-8)
+    restored_tensor = restored_tensor.to(dtype)
     return restored_tensor
 
 
--- a/kornia/contrib/extract_patches.py
+++ b/kornia/contrib/extract_patches.py
@@ -260,10 +260,12 @@ def combine_tensor_patches(
     norm_map = F.fold(
         input=unfold_ones, output_size=original_size, kernel_size=window_size, stride=stride, padding=unpadding
     )
+
     # Restored tensor
     saturated_restored_tensor = F.fold(
         input=patches, output_size=original_size, kernel_size=window_size, stride=stride, padding=unpadding
     )
+
     # Remove satuation effect due to multiple summations
     restored_tensor = saturated_restored_tensor / (norm_map + eps)
     if int_flag:
--- a/kornia/contrib/extract_patches.py
+++ b/kornia/contrib/extract_patches.py
@@ -251,12 +251,15 @@ def combine_tensor_patches(
     # Assuming BNCHW
     patches = patches.permute(0, 2, 3, 4, 1)
     patches = patches.reshape(patches.shape[0], -1, patches.shape[-1])
+    dtype = patches.dtype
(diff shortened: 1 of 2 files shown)


kornia/contrib/extract_patches.py
```diff
@@ -242,45 +240,26 @@ def combine_tensor_patches(
...
-
-    if unpadding:
-        unpadding = cast(Tuple[int, int, int, int], unpadding)
-        restored_tensor = pad(restored_tensor, [-i for i in unpadding])
+    unpadding = cast(Tuple[int, int], _pair(unpadding))
+
+    if len(unpadding) != 2:
+        raise AssertionError("Unpadding must be either an int or a tuple of two ints")
+
+    fold = nn.Fold(output_size=original_size, kernel_size=window_size, stride=stride, padding=unpadding)
```
edgarriba, line 248: use the functional better, no need for a class here

kornia/contrib/extract_patches.py
```diff
@@ -242,45 +240,26 @@ def combine_tensor_patches(
...
+    unfold = nn.Unfold(kernel_size=window_size, stride=stride, padding=unpadding)
```
edgarriba, line 249: use the functional better

kornia/contrib/extract_patches.py
```diff
@@ -242,45 +240,26 @@ def combine_tensor_patches(
...
+    ones = torch.ones(patches.shape[0], patches.shape[2], original_size[0], original_size[1])
+    # Assuming BNCHW
```
edgarriba, line 251: add a checker to make that assumption happen

kornia/contrib/extract_patches.py
```diff
@@ -242,45 +240,26 @@ def combine_tensor_patches(
...
+    patches = patches.permute(0, 2, 3, 4, 1)
+    patches = patches.reshape(patches.shape[0], -1, patches.shape[-1])
+    dtype = patches.dtype
+    patches = patches.double()
```
edgarriba, line 255:
```suggestion
    patches = patches.to(dtype)
```
?

kornia/contrib/extract_patches.py
```diff
@@ -242,45 +240,26 @@ def combine_tensor_patches(
...
+
+    # Calculate normalization map
+    norm_map = fold(unfold(ones))
+    saturated_restored_tensor = fold(patches)
+    # Remove satuation effect due to multiple summations
+    restored_tensor = saturated_restored_tensor / (norm_map)
```
edgarriba, line 261:
```suggestion
    restored_tensor = saturated_restored_tensor / (norm_map + eps)
```
o assure safe division

kornia/contrib/extract_patches.py
ashnair1: Check is already implemented at the start of the function

kornia/contrib/extract_patches.py
ashnair1: This is required since `fold` does not work with other data types like (int or long). The casting to double is required for `torch.autograd.gradcheck` since it needs double precision numbers to run.

johnnv1: The `torch.autograd.gradcheck` should run with just fp64, normally we use the `kornia.testing.tensor_to_gradcheck_var` (https://github.com/kornia/kornia/blob/master/kornia/testing/__init__.py#L43) to ensure it -- the casting here shouldn't be needed

johnnv1:
```suggestion
```

kornia/contrib/extract_patches.py
```diff
@@ -19,7 +19,7 @@ def compute_padding(
         window_size: the size of the sliding window used while extracting patches.
 
     Return:
-        The required padding for `(top, bottom, left, right)` as a tuple of 4 ints.
+        The required padding for `(vertical, horizontal)` as a tuple of 2 ints.
```
johnnv1, line 22: why we shouldn't support `(top, bottom, left, right)` anymore?

ashnair1: This is due to `fold` only accepting padding as a tuple of 2 ints.

kornia/contrib/extract_patches.py
ashnair1: I'm not exactly sure what you mean. The casting has to be done so that the function works with any int tensor (which is necessary for the doctests of this function to pass).  

The choice of double as an intermediate dtype was because gradcheck had raised a error (`RuntimeError: Jacobian mismatch for output 0 with respect to input 0`) when I had previously run tests. 

Do you mean have an explicit check as follows?

```python
int_flag = 0
if not torch.is_floating_point(patches):
    int_flag = 1
    dtype = patches.dtype
    patches = patches.float()

....rest..of...code

if int_flag:
    restored_tensor = restored_tensor.to(dtype)
```

johnnv1: if you want to make it work for integer tensors yes -- this will avoid casting between floating types, which will save some ops

johnnv1: The gradcheck test should run just within `torch.float64` tensors

kornia/contrib/extract_patches.py
```diff
@@ -236,51 +235,39 @@ def combine_tensor_patches(
...
-
-    if unpadding:
-        unpadding = cast(Tuple[int, int, int, int], unpadding)
-        restored_tensor = pad(restored_tensor, [-i for i in unpadding])
+    unpadding = cast(Tuple[int, int], _pair(unpadding))
+
+    if len(unpadding) != 2:
+        raise AssertionError("Unpadding must be either an int or a tuple of two ints")
+
+    ones = torch.ones(patches.shape[0], patches.shape[2], original_size[0], original_size[1])
```
edgarriba, line 249: `ones_like` after patches, to infer device

kornia/contrib/extract_patches.py
```diff
@@ -236,51 +235,39 @@ def combine_tensor_patches(
...
+    patches = patches.permute(0, 2, 3, 4, 1)
+    patches = patches.reshape(patches.shape[0], -1, patches.shape[-1])
+    int_flag = 0
+    if not torch.is_floating_point(patches):
+        int_flag = 1
+        dtype = patches.dtype
+        patches = patches.float()
+
+    # Calculate normalization map
+    unfold_ones = F.unfold(ones, kernel_size=window_size, stride=stride, padding=unpadding)
```
edgarriba, line 259: @johnnv1 should we expose this functions to core ?

kornia/contrib/extract_patches.py
```diff
@@ -236,51 +235,39 @@ def combine_tensor_patches(
...
+    norm_map = F.fold(
+        input=unfold_ones, output_size=original_size, kernel_size=window_size, stride=stride, padding=unpadding
+    )
```
edgarriba, line 262:
```suggestion
    )

```

kornia/contrib/extract_patches.py
```diff
@@ -236,51 +235,39 @@ def combine_tensor_patches(
...
+    # Restored tensor
+    saturated_restored_tensor = F.fold(
+        input=patches, output_size=original_size, kernel_size=window_size, stride=stride, padding=unpadding
+    )
+    # Remove satuation effect due to multiple summations
```
edgarriba, line 267:
```suggestion

    # Remove saturation effect due to multiple summations
```

kornia/contrib/extract_patches.py
ashnair1: Not sure what you mean. Can you elaborate? `ones_like(input)` is equivalent to `ones(input.size())` but that's not what we're doing here,

edgarriba: yes, sorry, the following should work, otherwise this code most likely will crash with cuda device
 `torch.ones(patches.shape[0], patches.shape[2], original_size[0], original_size[1], device=patches.device, dtype=patches.dtype)`

kornia/contrib/extract_patches.py
johnnv1: i see -- then i think is better to make it as a v2 and add a deprecation warning that the behavior will be changed in the future (after 0.7.2) cc @edgarriba

johnnv1: something like:

```python
from kornia.utils.helpers import deprecated

@deprecated('compute_padding_v2', '0.7.0', 'This function behaviour will be replaced by compute_padding_v2 after kornia-0.7.2.')
def compute_padding
...

def compute_padding_v2 # with your implementation

File: kornia/contrib/extract_patches.py (hunk 8 above)
Comment (johnnv1) [reply kornia/contrib/extract_patches.py:267]: i don't think so

File: kornia/contrib/extract_patches.py (hunk 6 above)
Comment (edgarriba) [reply kornia/contrib/extract_patches.py:22]: agree

Comment (ashnair1) [reply kornia/contrib/extract_patches.py:22]: What about `CombineTensorPatches`? Originally it did not take stride as a parameter. Does it need a deprecation note as well?

Also I just want to point out that the main use of `compute_padding` is to compute the padding required to do the Extract - Combine process correctly. So the change shouldn't affect users if they have used it like so.

```python
import torch

from kornia.contrib import CombineTensorPatches, ExtractTensorPatches, compute_padding

# Original size
h, w = 4, 4
# Window size
win = 3
# Stride
stride = 1
# Padding size
padding = compute_padding(original_size=(h,w), window_size=win)

image = torch.randn(2, 3, h, w)
tiler = ExtractTensorPatches(window_size=win, stride=stride, padding=padding)
merger = CombineTensorPatches(original_size=(h,w), window_size=win, stride=stride, unpadding=padding)
image_tiles = tiler(image)
restored_image = merger(image_tiles)
assert torch.allclose(image,restored_image)
```

johnnv1: right, it is very unlikely that users are using compute_padding directly out of this setup, let's just go ahead with it

kornia/contrib/extract_patches.py
```diff
@@ -329,19 +326,21 @@ def extract_tensor_patches(
     if len(input.shape) != 4:
         raise ValueError(f"Invalid input shape, we expect BxCxHxW. Got: {input.shape}")
 
+    window_size = _pair(window_size)
+    stride = _pair(stride)
+
+    if (stride[0] > window_size[0]) | (stride[1] > window_size[1]):
+        raise AssertionError(f"Stride={stride} should be less than or equal to Window size={window_size}")
+
```
ashnair1, line 334: In the event `stride` is greater than `window_size`, we need to throw an error since info is lost and can't be restored.

kornia/contrib/extract_patches.py
```diff
@@ -329,19 +326,21 @@ def extract_tensor_patches(
     if len(input.shape) != 4:
         raise ValueError(f"Invalid input shape, we expect BxCxHxW. Got: {input.shape}")
 
+    window_size = cast(Tuple[int, int], _pair(window_size))
+    stride = cast(Tuple[int, int], _pair(stride))
+
+    if (stride[0] > window_size[0]) | (stride[1] > window_size[1]):
+        raise AssertionError(f"Stride={stride} should be less than or equal to Window size={window_size}")
+
```
ashnair1, line 334: In the event `stride` is greater than `window_size`, we need to throw an error since info is lost and can't be restored.

ashnair1:
> LGTM. Seems that docstrings need to be updated https://github.com/kornia/kornia/actions/runs/6927786118/job/18842336075?pr=2650

This is coming from the tutorial notebook since I didn't specify the stride parameter there. Once kornia/tutorials#69 is approved, this error should be taken care of.

johnnv1 (approved):

PR #2689, opened 2023-12-17
[wip] Added Keypoint Visibility Field

alexanderswerdlow:


kornia/geometry/keypoints.py
```diff
@@ -13,16 +13,34 @@ def _merge_keypoint_list(keypoints: List[Tensor]) -> Tensor:
     raise NotImplementedError
 
 
+def inside_image(coords, image_size):
```
shijianjian, line 16: add typing

kornia/geometry/keypoints.py
```diff
@@ -13,16 +13,34 @@ def _merge_keypoint_list(keypoints: List[Tensor]) -> Tensor:
...
+    )
+
+
 class Keypoints:
     """2D Keypoints containing Nx2 or BxNx2 points.
 
     Args:
         keypoints: Raw tensor or a list of Tensors with the Nx2 coordinates
         raise_if_not_floating_point: will raise if the Tensor isn't float
+        image_size: (B, W, H) or (W, H) of the image the keypoints are referring to.
```
shijianjian, line 31: I think having `image_size ` as an init argument is not a good design. Is it easy to remove?

kornia/geometry/keypoints.py
```diff
@@ -135,15 +167,17 @@ def transform_keypoints(self, M: Tensor, inplace: bool = False) -> "Keypoints":
...
         if inplace:
             self._data = transformed_boxes
+            self._update_validity_mask(params)
             return self
 
-        return Keypoints(transformed_boxes, False)
+        return Keypoints(transformed_boxes, False, self._image_size, self._visibility)
 
-    def transform_keypoints_(self, M: Tensor) -> "Keypoints":
+    def transform_keypoints_(self, M: Tensor, params: Dict[str, Tensor]) -> "Keypoints":
```
shijianjian, line 178: I suggest using `image_size` here, instead of `params`.

kornia/geometry/keypoints.py
```diff
@@ -165,12 +199,19 @@ def to_tensor(self, as_padded_sequence: bool = False) -> Union[Tensor, List[Tens
...
 
     def clone(self) -> "Keypoints":
-        return Keypoints(self._data.clone(), False)
+        return Keypoints(self._data.clone(), False, self._image_size.clone(), self._visibility.clone())
 
     def type(self, dtype: torch.dtype) -> "Keypoints":
         self._data = self._data.type(dtype)
         return self
 
+    def _update_validity_mask(self, params: Optional[Dict[str, Tensor]] = None):
```
shijianjian, line 208: this can be a public function for users I think.

shijianjian (commented):
It is overall a decent solution I think. We should do something similar to bounding boxes maybe.

kornia/geometry/keypoints.py
alexanderswerdlow: Removed. I instead defaulted to the `forward_input_shape` field unless `output_size` was specified.

kornia/geometry/keypoints.py
alexanderswerdlow: The issue here is that there are multiple [from what I can tell, 3] calls to this in `kornia/augmentation/_2d/geometric/base.py` where we'd need to specify the `image_size` logic now in `update_valid_mask`.

edgarriba: i agree with @shijianjian, keypoint shouldn't be aware of any dictionary logic and just receive image size (we even have a data structure for this). Keypoint should be as light as possible and any extra logic should be part of the augmentation package to keep the whole thing more maintainable.

kornia/geometry/keypoints.py
```diff
@@ -165,12 +198,23 @@ def to_tensor(self, as_padded_sequence: bool = False) -> Union[Tensor, List[Tens
...
 
     def clone(self) -> "Keypoints":
-        return Keypoints(self._data.clone(), False)
+        return Keypoints(self._data.clone(), False, self._valid_mask.clone())
 
     def type(self, dtype: torch.dtype) -> "Keypoints":
         self._data = self._data.type(dtype)
         return self
 
+    def update_valid_mask(self, params: Optional[Dict[str, Tensor]]):
```
edgarriba, line 207: is it really needed to be a public method?

edgarriba (changes requested):
This feature needs tests

edgarriba:
> It is overall a decent solution I think. We should do something similar to bounding boxes maybe.
@shijianjian you file an issue to track this ? i'd take as an opportunity to redefine the api based on some prototype i have here https://github.com/kornia/kornia/blob/master/kornia/contrib/object_detection.py#L23-L50

kornia/geometry/keypoints.py
shijianjian: > The issue here is that there are multiple [from what I can tell, 3] calls to this in `kornia/augmentation/_2d/geometric/base.py` where we'd need to specify the `image_size` logic now in `update_valid_mask`.

We may have an internal function inside augmentation modules to update `image_size`, where `update_validity_mask` can be called.

kornia/geometry/keypoints.py
```diff
@@ -165,12 +198,23 @@ def to_tensor(self, as_padded_sequence: bool = False) -> Union[Tensor, List[Tens
...
+        if params is not None:
+            if "output_size" in params:
+                image_size = params["output_size"]
+            else:
+                image_size = params["forward_input_shape"][-2:]
+                if self._data.ndim == 3:
+                    image_size = image_size[None].repeat(self._data.shape[0], 1)
+
+        self._valid_mask &= inside_image(self._data, image_size)
```
johnnv1, line 216: avoid inplace op
```suggestion
        self._valid_mask = self._valid_mask & inside_image(self._data, image_size)
```

johnnv1 (changes requested):
needs some tests mainly for the augmentation part

kornia/geometry/keypoints.py
alexanderswerdlow: Could you maybe point me to it? The only thing I could find is `_get_new_batch_shape` [here](https://github.com/alexanderswerdlow/kornia/blob/92deb42b56468ca36392f53d2a6b3616ca9e169d/kornia/augmentation/container/image.py#L290) which is called from `forward_parameters` and `forward` in `ImageSequential`.

Perhaps we put the dictionary logic [here](https://github.com/alexanderswerdlow/kornia/blob/92deb42b56468ca36392f53d2a6b3616ca9e169d/kornia/augmentation/base.py#L345) in `transform_keypoints`? It seems like we'd need to update all of the `apply_transform_keypoint` definitions however.

edgarriba:
@alexanderswerdlow let us know when you believe this is done. Since before merging we have rebase onto the new `main` branch. We can assist you with that, see: https://github.com/kornia/kornia/discussions/2719

alexanderswerdlow:
@edgarriba I hate to say this but I don't think I'll have time to dig into this more in the very near future. From what it sounds like, my implementation needs some pretty significant changes [and I'd have to agree myself, I think that the Keypoint class parsing the `params` is pretty ugly]. If it's just tests, I can add those, but otherwise it's probably best if someone else has the bandwidth to push it over the finish line. Apologies for that.

johnnv1:
as part of https://github.com/kornia/kornia/discussions/2719 plan, we have deleted the `master` in favor of `main`, if anyone has an interest in keep/restart working on this, let me know to help you.

PR #2714, opened 2024-01-08
feat: `connected_components` metric stop case when `num_iterations=0`

domef:
#### Changes
The function `connected_components` now supports `num_iterations` equals to zero, in this case the algorithm runs until convergence.

#### Type of change

- [x] 📚  Documentation Update
- [x] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/kornia/contrib/connected_components.py
+++ b/kornia/contrib/connected_components.py
@@ -23,6 +23,7 @@ def connected_components(image: Tensor, num_iterations: int = 100) -> Tensor:
         image: the binarized input image with shape :math:`(*, 1, H, W)`.
           The image must be in floating point with range [0, 1].
         num_iterations: the number of iterations to make the algorithm to converge.
+          If 0, the algorithm will run until convergence.
 
     Return:
         The labels image with the same shape of the input image.
@@ -34,8 +35,8 @@ def connected_components(image: Tensor, num_iterations: int = 100) -> Tensor:
     if not isinstance(image, Tensor):
         raise TypeError(f"Input imagetype is not a Tensor. Got: {type(image)}")
 
-    if not isinstance(num_iterations, int) or num_iterations < 1:
-        raise TypeError("Input num_iterations must be a positive integer.")
+    if not isinstance(num_iterations, int) or num_iterations < 0:
+        raise TypeError("Input num_iterations must be integer greater or equal to zero.")
 
     if len(image.shape) < 3 or image.shape[-3] != 1:
         raise ValueError(f"Input image shape must be (*,1,H,W). Got: {image.shape}")
@@ -51,8 +52,18 @@ def connected_components(image: Tensor, num_iterations: int = 100) -> Tensor:
     out = torch.arange(B * H * W, device=image.device, dtype=image.dtype).view((-1, 1, H, W))
     out[~mask] = 0
 
-    for _ in range(num_iterations):
-        out = F.max_pool2d(out, kernel_size=3, stride=1, padding=1)
-        out = torch.mul(out, mask)  # mask using element-wise multiplication
+    i = 0
+    while True:
+        i += 1
+        out_ = F.max_pool2d(out, kernel_size=3, stride=1, padding=1)
+        # mask using element-wise multiplication
+        out_ = torch.mul(out_, mask)
+        # stop if converged
+        if torch.all(out == out_):
+            break
+        out = out_
+        # reached if reached max iterations
+        if num_iterations > 0 and i == num_iterations:
+            break
 
     return out.view_as(image)
--- a/kornia/contrib/connected_components.py
+++ b/kornia/contrib/connected_components.py
@@ -36,9 +36,7 @@ def connected_components(image: Tensor, num_iterations: int = 100) -> Tensor:
         raise TypeError(f"Input imagetype is not a Tensor. Got: {type(image)}")
 
     if not isinstance(num_iterations, int) or num_iterations < 0:
-        raise TypeError(
-            "Input num_iterations must be an integer greater or equal to zero."
-        )
+        raise TypeError("Input num_iterations must be an integer greater or equal to zero.")
 
     if len(image.shape) < 3 or image.shape[-3] != 1:
         raise ValueError(f"Input image shape must be (*,1,H,W). Got: {image.shape}")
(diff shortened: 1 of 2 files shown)


kornia/contrib/connected_components.py
```diff
@@ -51,8 +52,18 @@ def connected_components(image: Tensor, num_iterations: int = 100) -> Tensor:
...
-        out = F.max_pool2d(out, kernel_size=3, stride=1, padding=1)
-        out = torch.mul(out, mask)  # mask using element-wise multiplication
+    i = 0
+    while True:
+        i += 1
+        out_ = F.max_pool2d(out, kernel_size=3, stride=1, padding=1)
+        # mask using element-wise multiplication
+        out_ = torch.mul(out_, mask)
+        # stop if converged
+        if torch.all(out == out_):
```
edgarriba, line 62: this check here  will fail due to numerical accuracy for floating point tensors. I suggest adding another metric (rmse or mse?) subject to a threshold. Worth to verify in the original connected components algorithm how is it done. /cc @ducha-aiki

kornia/contrib/connected_components.py
```diff
@@ -51,8 +52,18 @@ def connected_components(image: Tensor, num_iterations: int = 100) -> Tensor:
     out = torch.arange(B * H * W, device=image.device, dtype=image.dtype).view((-1, 1, H, W))
     out[~mask] = 0
 
-    for _ in range(num_iterations):
-        out = F.max_pool2d(out, kernel_size=3, stride=1, padding=1)
-        out = torch.mul(out, mask)  # mask using element-wise multiplication
+    i = 0
+    while True:
+        i += 1
```
edgarriba, line 57: as general good practice, update the index in the end of the loop ?

kornia/contrib/connected_components.py
```diff
@@ -34,8 +35,8 @@ def connected_components(image: Tensor, num_iterations: int = 100) -> Tensor:
     if not isinstance(image, Tensor):
         raise TypeError(f"Input imagetype is not a Tensor. Got: {type(image)}")
 
-    if not isinstance(num_iterations, int) or num_iterations < 1:
-        raise TypeError("Input num_iterations must be a positive integer.")
+    if not isinstance(num_iterations, int) or num_iterations < 0:
```
edgarriba, line 38:
```suggestion
    if not isinstance(num_iterations, int) or num_iterations <= 0:
```
non-zero and positive numbers

kornia/contrib/connected_components.py
```diff
@@ -51,8 +52,18 @@ def connected_components(image: Tensor, num_iterations: int = 100) -> Tensor:
...
+        i += 1
+        out_ = F.max_pool2d(out, kernel_size=3, stride=1, padding=1)
+        # mask using element-wise multiplication
+        out_ = torch.mul(out_, mask)
+        # stop if converged
+        if torch.all(out == out_):
+            break
+        out = out_
+        # reached if reached max iterations
+        if num_iterations > 0 and i == num_iterations:
```
edgarriba, line 66:
```suggestion
        if i >= num_iterations or error >= threshold:
```
i believe something like that should work. However, if we introduce this condition i'm afraid that we cannot support batch as it is. So, we might have to introduce a small logic that includes `vmap` @johnnv1 ?

kornia/contrib/connected_components.py
domef: I changed it to `torch.allclose`.

kornia/contrib/connected_components.py
domef: I moved it to end of the loop and initialized `i` to 1.

kornia/contrib/connected_components.py
domef: I use `num_iterations=0` to detect if I need to stop when reaching convergence or maximum iterations.

kornia/contrib/connected_components.py
domef: When using batch size greater than 1 the convergence should be reached when every element in the batch reaches it.

kornia/contrib/connected_components.py
domef: At the moment I left the default value to 100 for backward compatibility but maybe it could be better to set it to 0 since it ensures that the algorithm works correctly.

kornia/contrib/connected_components.py
```diff
@@ -48,11 +49,22 @@ def connected_components(image: Tensor, num_iterations: int = 100) -> Tensor:
 
     # allocate the output tensors for labels
     B, _, _, _ = image_view.shape
-    out = torch.arange(B * H * W, device=image.device, dtype=image.dtype).view((-1, 1, H, W))
+    out = torch.arange(1, B * H * W + 1, device=image.device, dtype=image.dtype)
```
edgarriba, line 52: actually now comparing to the original implementation, i'm realizing that this might be wrong for batched images, i believe that should be 
```python
out = torch.zeros(1, H, W, ...)
out = out.repeat(B, 1, 1, 1)
out[..., 0, 1] = 1.
```
and i still believe that this function we should revert to non-batched or handle batches via vmap

domef: I don't understand this change, moreover I don't understand why batches should be treated in a special way with vamp. The current code should work correctly with batched tensors.

domef:
any update?

kornia/contrib/connected_components.py
edgarriba: sorry for long response -- the main issue here is that with the current batched implementation i don't think we can assure that the algorithm can meet the convergence condition per each sample of the batch independently. After some digging not even sure that vmap solve that problem. So my suggestion would be to revert the base algorithm to single image and to keep backward compatibility for the batched version we just python for_loop. I'll think in the future how to improve this from Rust side.

johnnv1:
as part of https://github.com/kornia/kornia/discussions/2719 plan, we have deleted the `master` in favor of `main`, if anyone has an interest in keep/restart working on this, let me know to help you.

edgarriba:
my main concern is that with this batch implementation each sample of the batch will have different convergence rates, and there's no way in pytorch control the flow per element in the batch. That's why i'm suggesting to revert the core to a non-batched implementation.

domef:
But this is how pytorch works, every pytorch operation applied to a batched tensor will be the same for every batch element. I think that using a python for loop can only worsen the performance of the algorithm.

edgarriba:
sorry for late reply, in that case I advocate to provide a non-batched version of the operator

domef:
I think that the non-batched version could be just like this:
```python
def cc_nonbatched(tensor):
    return cc_batched(tensor.unsqueeze(0)).squeeze(0)
```
Or it can be a use responsibility to unsqueeze/squeeze.

PR #2706, opened 2023-12-29
Refactor LightGlue as in LG upstream, also add LightGlue for KeyNetAffNetHardNet

ducha-aiki:
#### Changes
LightGlue has been refactored in upstream(https://github.com/cvg/LightGlue/blob/main/lightglue/lightglue.py), this PR does same refactor (module kornia-related changes).

Also it adds LightGlue model for KeyNet-AffNet-HardNet.
Example is here: 

https://github.com/kornia/tutorials/pull/75

#### Type of change
- [x] 🔬 New feature (non-breaking change which adds functionality)

#### Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Did you update CHANGELOG in case of a major change?

PR #2716, opened 2024-01-10
fix C417

johnnv1:
#### Changes
This fixes the C417 rule from ruff (falling on pre-commit CI). This is introduced in #2706


ducha-aiki:
@johnnv1 @edgarriba Actually, why do we even have this rule? What's wrong with lambdas and using the language features?

ducha-aiki (approved):

edgarriba:
@ducha-aiki https://docs.astral.sh/ruff/rules/unnecessary-map/ map+lambda seems to be a slow instruction

PR #2724, opened 2024-01-11
Configure codeQL

johnnv1:


johnnv1:
Now that I've seen that codeql is already running. Earlier it was missing its settings, but now it's in default, and it looks right to me. No need for this PR, which would be via "advanced settings"

PR #2727, opened 2024-01-12
Replace docs source link from `viewcode` to github 

johnnv1:
#### Changes

This replaces the viewcode extension to use the linkcode extension.

Instead of showing the code on the docs, the `source` link will redirect to the github code.

Based on https://github.com/readthedocs/sphinx-autoapi/issues/202

I also believe that this should improve searches on the docs page, which sometimes returned pages containing the source code instead of the documentation.

#### Type of change

- [x] 📚  Documentation Update

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,4 +1,5 @@
 import importlib.util
+import inspect
 import os
 import sys
 
@@ -39,7 +40,8 @@ extensions = [
     "sphinx_autodoc_typehints",
     "sphinx_autodoc_defaultargs",
     "sphinx_copybutton",
-    "sphinx.ext.viewcode",
+    # "sphinx.ext.viewcode",
+    "sphinx.ext.linkcode",
     "sphinx.ext.githubpages",
     "sphinxcontrib.bibtex",
     "sphinxcontrib.gtagjs",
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -161,6 +163,47 @@
 html_js_files = ["js/custom.js"]
 
 
+# Configure viewcode extension.
+code_url = "https://github.com/kornia/kornia/blob/main"
```
johnnv1, line 167: this way the docs will always redirect to the main, should we someway as scikit inject the commit/revision here? https://github.com/scikit-learn/scikit-learn/blob/main/doc/sphinxext/github_link.py

docs/source/conf.py
```diff
@@ -39,7 +40,8 @@
     "sphinx_autodoc_typehints",
     "sphinx_autodoc_defaultargs",
     "sphinx_copybutton",
-    "sphinx.ext.viewcode",
+    # "sphinx.ext.viewcode",
```
edgarriba, line 43: remove?

docs/source/conf.py
edgarriba: really love this feature. Add the reference link to the file in case we want to add / remove stuff

docs/source/conf.py
```diff
@@ -161,6 +163,47 @@
 html_js_files = ["js/custom.js"]
 
 
+# Configure viewcode extension.
+code_url = "https://github.com/kornia/kornia/blob/main"
+
+
+def linkcode_resolve(domain, info):
```
edgarriba, line 170: should we include this like in the example under doc/sphinxexr?

edgarriba (approved):

PR #2745, opened 2024-01-21
Depreciates `kornia.testing`

johnnv1:
#### Changes
This patch (bigger than expected) depreciates the `kornia.testing` module. Planning to delete it in December, before the 0.8.0 release.

The `kornia.testing` has been used as a set of functionalities for kornia test suite. Having it as a module, it makes some changes unviable, because needs to depreciate things before removing/changing.

The functions that have been used internally I moved it to live in another module, If you (@edgarriba @shijianjian @ducha-aiki) think that other features that exist in `kornia.testing` should continue to be part of the public package, let me know.

All the functionalities used in our test suite have been moved to the `testing` directory. At `testing` should live all the code that we need just for tests, but aren't tests themselves (Like common functions to generate data or tests).

Within this migration (kornia.testing -> testing) I fixes #2162, and also updated a lot of tests to use the `BaseTester`. I also updated the `BaseTester` to **not require the annoying tests**. and now it is just a base class within some functionalities (as the assert_close and gradcheck with magics helpers)

Summarizing: 
- depreciated `kornia.testing`
- move tests utilities code to `testing` directory -- which allows us to make changes more abruptly, without worrying about it being a feature of the library
- Update all tests and code (yes, the testing API have been used on some internal code) to not use the `kornia.testing`
- update a lot of `gradtest` to not run within float32, since they are always casting it to float64
- enhancement the `BaseTester` by:
  - add more magics under `assert_close` and `gradcheck` -- now it auto-cast all tensors to fp64 within grad by default (#2162)
  - remove the abstract methods. Making it easier to prototype new features, without the need for multiple (often pointless) tests

Things to be opened as issues for us to do after this PR:
- Move code that lives in `tests/` which are not tests themselves, to live in `testing`.
- Migrate all test classes to use `BaseTester` and its functionalities.
- delete `kornia.testing` in December 2024

#### Type of change
- [x] 🧪 Tests Cases

PR #2748, opened 2024-01-23
[Feature] Add Differentiable JPEG

ChristophReich1996:
#### Changes

As discussed with @shijianjian this pull request adds a differentiable JPEG encoding-decoding implementation to `kornia.enhance`. Differentiable JPEG coding can be used as a differentiable data augmentation approach or for optimizing JPEG for deep vision models. No new dependencies are introduced.

This implementation takes inspiration from the diff. JPEG approach of [Shin et al.](https://machine-learning-and-security.github.io/papers/mlsec17_paper_54.pdf) and [Reich et al.](https://arxiv.org/abs/2309.06978), however, uses a custom soft clipping operation (ELU-like). Additionally, the quantization tables are first clipped and then rounded (different from Reich et al.). For approximating the rounding function the polynomial approximation of Shin et al. is used. The implementation leads to similar results as the approach by Reich et al. (see [cardinality test](https://github.com/ChristophReich1996/kornia/blob/dc0f8248ba414a9dc9ed97dd0a216f45d587432f/test/enhance/test_jpeg.py#L86)).

#### Type of change
- [ ] 📚  Documentation Update
- [X] 🧪 Tests Cases
- [X] 🔬 New feature (non-breaking change which adds functionality)
- [X] 📝 This change requires a documentation update (probably)


shijianjian (approved):

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,566 @@
...
+        >>> img_jpeg = diff_jpeg(img, jpeg_quality)
+        >>> img_jpeg.sum().backward()
+    """
+    # Check that inputs are tensors
+    KORNIA_CHECK_IS_TENSOR(image_rgb)
+    KORNIA_CHECK_IS_TENSOR(jpeg_quality)
+    KORNIA_CHECK_IS_TENSOR(quantization_table_y)
+    KORNIA_CHECK_IS_TENSOR(quantization_table_c)
+    # Check shape of inputs
+    KORNIA_CHECK_SHAPE(image_rgb, ["B", "3", "H", "W"])
```
edgarriba, line 453: would be great (or in a separated PR) to support out of the box non-batched images 3xHxW

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,566 @@
...
+        f"JPEG quality is out of range. Expected range is [0, 99], "
+        f"got [{jpeg_quality.amin().item()}, {jpeg_quality.amax().item()}]. Consider clipping jpeg_quality.",
+    )
+    # Get original shape
+    H, W = image_rgb.shape[2:]
+    # Get device and dtype
+    dtype: Dtype = image_rgb.dtype
+    device: Device = image_rgb.device
+    # Quantization tables to same device and dtype as input image
+    quantization_table_y = quantization_table_y.to(dtype).to(device)
```
edgarriba, line 479:
```suggestion
    quantization_table_y = quantization_table_y.to(dtype, device)
```

test/enhance/test_jpeg.py
```diff
@@ -0,0 +1,990 @@
...
+        img = torch.rand(B, 3, H, W, device=device, dtype=dtype)
+        jpeg_quality = torch.randint(low=0, high=100, size=(B,), device=device, dtype=dtype)
+        qt_y = torch.randint(low=1, high=255, size=(B, 8, 8), device=device, dtype=dtype)
+        qt_c = torch.randint(low=1, high=255, size=(B, 8, 8), device=device, dtype=dtype)
+        diff_jpeg_module = kornia.enhance.DiffJPEG()
+        img_jpeg = diff_jpeg_module(img, jpeg_quality, qt_y, qt_c)
+        assert img_jpeg is not None
+        assert img_jpeg.shape == img.shape
+
+    def test_gradcheck(self, device) -> None:
```
edgarriba, line 985: can you make somehow a test that performs a backward and check the gradients ? it would be funny to have a differentiable method without testing that part

test/enhance/test_jpeg.py
```diff
@@ -0,0 +1,990 @@
...
+                            -0.000,
+                        ],
+                    ],
+                ]
+            ],
+            device=device,
+            dtype=dtype,
+        )
+        # We use a slightly higher tolerance since our implementation varies from the reference implementation
+        assert_close(img_jpeg[..., :4, :4], img_jpeg_ref[..., :4, :4], rtol=0.01, atol=0.01)
```
edgarriba, line 972:
```suggestion
        self.assert_close(img_jpeg[..., :4, :4], img_jpeg_ref[..., :4, :4], rtol=0.01, atol=0.01)
```
coming from `BaseTester`

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,566 @@
...
+    )
+    cr_encoded: Tensor = _quantize(
+        dct_cr,
+        jpeg_quality,
+        quantization_table_c,
+    )
+    return y_encoded, cb_encoded, cr_encoded
+
+
+def _jpeg_decode(
```
edgarriba, line 336: would be interesting to include this one in the high level api as well (in another pr possibly)

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,566 @@
...
+    Returns:
+        rgb_decoded (Tensor): Decompressed RGB image of the shape :math:`(B, 3, H, W)`.
+    """
+    # Dequantize inputs
+    input_y = _dequantize(
+        input_y,
+        jpeg_quality,
+        quantization_table_y,
+    )
+    input_cb = _dequantize(
```
edgarriba, line 367: stupid question, why you cannot vectorise the process for the cb/cr channels in a single operation?

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,566 @@
...
+    image_cb = _chroma_upsampling(image_cb)
+    image_cr = _chroma_upsampling(image_cr)
+    # Back to [0, 1] pixel-range
+    image_ycbcr: Tensor = torch.stack((image_y, image_cb, image_cr), dim=1) / 255.0
+    # Convert back to RGB space.
+    rgb_decoded: Tensor = ycbcr_to_rgb(image_ycbcr)
+    return rgb_decoded
+
+
+def diff_jpeg(
```
edgarriba, line 395: as the high level api for this module, i would like here to have a more explicit name like `encode_to_jpeg_differentiable`

test/enhance/test_jpeg.py
ChristophReich1996: @edgarriba fully agree, however, PyTorch's grad check function is not really applicable here. I could check the gradient based on the reference implementation. What do you think about this?

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,566 @@
...
+        [16, 11, 10, 16, 24, 40, 51, 61],
+        [12, 12, 14, 19, 26, 58, 60, 55],
+        [14, 13, 16, 24, 40, 57, 69, 56],
+        [14, 17, 22, 29, 51, 87, 80, 62],
+        [18, 22, 37, 56, 68, 109, 103, 77],
+        [24, 35, 55, 64, 81, 104, 113, 92],
+        [49, 64, 78, 87, 103, 121, 120, 101],
+        [72, 92, 95, 98, 112, 100, 103, 99],
+    ],
+    dtype=torch.float,
```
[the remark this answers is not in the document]
ChristophReich1996: Done :)

kornia/enhance/jpeg.py
ChristophReich1996: Done :) https://github.com/kornia/kornia/blob/505765b226d4f7618c2812234a2193057f7df419/kornia/enhance/jpeg.py#L477-L478

kornia/enhance/jpeg.py
ChristophReich1996: Agree, but `encode_to_jpeg_differentiable` is somewhat misleading since we perform both encoding and decoding. What about `jpeg_encode_decode_differentiable`?

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,564 @@
...
+from kornia.color import rgb_to_ycbcr, ycbcr_to_rgb
+from kornia.core import Device, Dtype, Module, Tensor
+from kornia.core.check import (
+    KORNIA_CHECK,
+    KORNIA_CHECK_IS_TENSOR,
+    KORNIA_CHECK_SHAPE,
+)
+from kornia.geometry import rescale
+
+QUANTIZATION_TABLE_Y: Tensor = torch.tensor(
```
johnnv1, line 14: move these tensors to be lazyloaded by putting them into a function. Otherwise, these tensors will be created when importing occurs; the function can have device and dtype as parameter, to initialize these tensor in the right place instead of moving/casting it after

test/enhance/test_jpeg.py
ChristophReich1996: I added a test to check that the gradients are close to the reference implementation.

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,566 @@
...
+    image_ycbcr: Tensor = torch.stack((image_y, image_cb, image_cr), dim=1) / 255.0
+    # Convert back to RGB space.
+    rgb_decoded: Tensor = ycbcr_to_rgb(image_ycbcr)
+    return rgb_decoded
+
+
+def diff_jpeg(
+    image_rgb: Tensor,
+    jpeg_quality: Tensor,
+    quantization_table_y: Tensor = QUANTIZATION_TABLE_Y,
```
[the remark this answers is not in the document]
ChristophReich1996: Done

kornia/enhance/jpeg.py
ChristophReich1996: Done in both the encoding and decoding.

ChristophReich1996: Very valid question btw ;)

kornia/enhance/jpeg.py
ChristophReich1996: Not sure if correct but implemented like this:

https://github.com/kornia/kornia/blob/046443331956d8703683944e384cf21c75fb5d07/kornia/enhance/jpeg.py#L17-L32

kornia/enhance/jpeg.py
ChristophReich1996: I change the naming to `jpeg_encode_decode_differentiable` and `JPEGEncodeDecodeDifferentiable` for the module. Let me know what you think :)

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,566 @@
...
+import torch
+
+from kornia.color import rgb_to_ycbcr, ycbcr_to_rgb
+from kornia.core import Device, Dtype, Module, Tensor
+from kornia.core.check import (
+    KORNIA_CHECK,
+    KORNIA_CHECK_IS_TENSOR,
+    KORNIA_CHECK_SHAPE,
+)
+from kornia.geometry import rescale
```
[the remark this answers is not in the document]
ChristophReich1996: Done. 
Btw I was always wondering is there a reason why `__all__` is always a list and not a tuple? I know using a list is the default but wouldn't it make more sense to use an immutable object?

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,566 @@
...
+       - :math:`\hat{I}` is the resulting JPEG encoded-decoded image.
+
+    Notes:
+        The input (and output) pixel range is :math:`[0, 1]`. In case you want to handle normalized images you are
+        required to first perform denormalization followed by normalizing the output images again.
+        Note, this implementation models the encoding-decoding mapping of JPEG in a differentiable setting,
+        howerver, does not allow to excess the JPEG-coded byte file itself. For more details please refer to [1].
+
+    Reference:
+        [1] https://arxiv.org/abs/2309.06978
```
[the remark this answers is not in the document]
ChristophReich1996: Done

kornia/enhance/jpeg.py
ChristophReich1996: Let's do it in a separate PR if it's fine.

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,566 @@
...
+       - :math:`QT_{y}` is the luma quantization table.
+       - :math:`QT_{c}` is the chroma quantization table.
+       - :math:`\hat{I}` is the resulting JPEG encoded-decoded image.
+
+    Notes:
+        The input (and output) pixel range is :math:`[0, 1]`. In case you want to handle normalized images you are
+        required to first perform denormalization followed by normalizing the output images again.
+        Note, this implementation models the encoding-decoding mapping of JPEG in a differentiable setting,
+        howerver, does not allow to excess the JPEG-coded byte file itself. For more details please refer to [1].
+
```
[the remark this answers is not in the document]
ChristophReich1996: Done

https://github.com/kornia/kornia/blob/7ca6f8e3a2c26d9972d2155af02e5ce0d4b2cade/kornia/enhance/jpeg.py#L525-L534

kornia/enhance/jpeg.py
johnnv1: yeah, perfect

test/enhance/test_jpeg.py
```diff
@@ -0,0 +1,1001 @@
...
+        img = torch.rand(B, 3, H, W, device=device, dtype=dtype)
+        jpeg_quality = torch.randint(low=0, high=100, size=(B,), device=device, dtype=dtype)
+        qt_y = torch.randint(low=1, high=255, size=(B, 8, 8), device=device, dtype=dtype)
+        qt_c = torch.randint(low=1, high=255, size=(B, 8, 8), device=device, dtype=dtype)
+        diff_jpeg_module = kornia.enhance.JPEGEncodeDecodeDifferentiable()
+        img_jpeg = diff_jpeg_module(img, jpeg_quality, qt_y, qt_c)
+        assert img_jpeg is not None
+        assert img_jpeg.shape == img.shape
+
+    def test_gradcheck(self, device) -> None:
```
johnnv1, line 985:
```suggestion
    @pytest.mark.slow
    def test_gradcheck(self, device) -> None:
```

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,573 @@
...
+import torch
+
+from kornia.color import rgb_to_ycbcr, ycbcr_to_rgb
+from kornia.core import Device, Dtype, Module, Tensor
+from kornia.core.check import (
+    KORNIA_CHECK,
+    KORNIA_CHECK_IS_TENSOR,
+    KORNIA_CHECK_SHAPE,
+)
+from kornia.geometry import rescale
```
johnnv1, line 14:
```suggestion
from kornia.geometry.transform.affwarp import rescale
```

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,573 @@
...
+            [99, 99, 99, 99, 99, 99, 99, 99],
+            [99, 99, 99, 99, 99, 99, 99, 99],
+            [99, 99, 99, 99, 99, 99, 99, 99],
+        ],
+        device=device,
+        dtype=dtype,
+    )
+
+
+def _patchify_8x8(input: Tensor) -> Tensor:
```
johnnv1, line 55: would it be worth using https://kornia.readthedocs.io/en/latest/contrib.html#kornia.contrib.extract_tensor_patches and the `combine_tensor_patches` (instead of `_patchify_8x8` and ``_unpatchify_8x8) here? cc @edgarriba

kornia/enhance/jpeg.py
ChristophReich1996: I don't see an immediate use for this function but if you like we can also include both `_jpeg_encode` and `_jpeg_decode`.

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,573 @@
...
+
+    Returns:
+        output (Tensor): Pseudo rounded tensor of the same shape as input tensor.
+    """
+    # Perform differentiable rounding
+    output: Tensor = torch.floor(input) + (input - 0.5 - torch.floor(input)) ** 3
+    return output
+
+
+def _differentiable_clipping(
```
johnnv1, line 165: for a further pr, i think having `_differentiable_clipping`, ` _differentiable_polynomial_floor` and ` _differentiable_polynomial_rounding` into our utilities can be usefully :)

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,573 @@
...
+    Returns:
+        output (Tensor): DCT output tensor of the shape :math:`(B, N, 8, 8)`.
+    """
+    # Get dtype and device
+    dtype: Dtype = input.dtype
+    device: Device = input.device
+    # Make DCT tensor and scaling
+    index: Tensor = torch.arange(8, dtype=dtype, device=device)
+    x, y, u, v = torch.meshgrid(index, index, index, index)
+    dct_tensor: Tensor = ((2.0 * x + 1.0) * u * torch.pi / 16.0).cos() * ((2.0 * y + 1.0) * v * torch.pi / 16.0).cos()
```
johnnv1, line 102: unrelated comment, if the torch has a `pi` constant, why do we have one too? cc @edgarriba

kornia/enhance/jpeg.py
johnnv1: +1

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,573 @@
...
+        \text{JPEG}_{\text{diff}}(I, q, QT_{y}, QT_{c}) = \hat{I}
+
+    Where:
+       - :math:`I` is the original image to be coded.
+       - :math:`q` is the JPEG quality controling the compression strength.
+       - :math:`QT_{y}` is the luma quantization table.
+       - :math:`QT_{c}` is the chroma quantization table.
+       - :math:`\hat{I}` is the resulting JPEG encoded-decoded image.
+
+    Notes:
```
johnnv1, line 418:
```suggestion
    .. note::
```

johnnv1 (approved):
Thanks for this PR, it's really cool stuff. Just some minor comments

johnnv1:
The CI is falling by an unrelated test, no worry, I'll merge #2745 which has a fix for it (I think so 😅 )

test/enhance/test_jpeg.py
```diff
@@ -0,0 +1,1002 @@
+import pytest
+import torch
+
+import kornia
+from kornia.testing import BaseTester
```
johnnv1, line 5:
```suggestion
from testing.base import BaseTester
```

kornia/enhance/jpeg.py
ChristophReich1996: Done

kornia/enhance/jpeg.py
ChristophReich1996: When I import `combine_tensor_patches` using `from kornia.contrib import extract_tensor_patches` I receive a circular import error 🤔

johnnv1: Try with the full path `from kornia.contrib.extract_patches import extract_tensor_patches`

johnnv1:
@edgarriba let me know when the changes are fine, I can squash it and put it on top of the main branch (CI is falling now because I updated to use the new testing, but just see now it is on top of the master instead of the main)

kornia/enhance/jpeg.py
edgarriba: historically wasn't there plus i remember some conflicts with the device api. Worth to revisit (open a ticket) and alias `[torch/kornia].pi`

kornia/enhance/jpeg.py
edgarriba: open a ticket for that

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,573 @@
...
+        W=W,
+        quantization_table_c=quantization_table_c,
+        quantization_table_y=quantization_table_y,
+    )
+    # Clip coded image
+    image_rgb_jpeg = _differentiable_clipping(input=image_rgb_jpeg, min=0.0, max=255.0)
+    return image_rgb_jpeg
+
+
+class JPEGEncodeDecodeDifferentiable(Module):
```
edgarriba, line 509: Not quite sure this is very valuable to provide since any pytorch user can easily create its own Module in any possible shape of the constructor. E.g if had to design the nn.Module of this function i'd put all the extra args (or maybe just a couple) into the module constructor so that the forward expect only the input image. But this is i guess of application specific, so i wouldn't support module (plus one less thing to maintain). Open to opinions @johnnv1 @shijianjian

kornia/enhance/jpeg.py
```diff
@@ -0,0 +1,573 @@
...
+    return rgb_decoded
+
+
+def jpeg_encode_decode_differentiable(
+    image_rgb: Tensor,
+    jpeg_quality: Tensor,
+    quantization_table_y: Tensor | None = None,
+    quantization_table_c: Tensor | None = None,
+) -> Tensor:
+    r"""Differentiable JPEG encoding-decoding module.
```
edgarriba, line 403: @ChristophReich1996 one more thing for future, is to include a meaningful visual output (auto-generated, or not) similar to other functions in this module

kornia/enhance/jpeg.py
edgarriba: got it -- in that case to have a more concise name `jpeg_codec_differentiable` where codec usually referred to compress/decompress

edgarriba (approved):
Left last comments but in general LGTM

kornia/enhance/jpeg.py
ChristophReich1996: Unfortunately, loading `extract_patches` directly also leads to a circular import...

```
from kornia.enhance import normalize_min_max
E   ImportError: cannot import name 'normalize_min_max' from partially initialized module 'kornia.enhance' (most likely due to a circular import)
```

ChristophReich1996: Just importing kornia and calling `kornia.enhance.extract_tensor_patches` works...

kornia/enhance/jpeg.py
ChristophReich1996: Done :)

kornia/enhance/jpeg.py
ChristophReich1996: Good point! People might want to learn the quantization tables. So we could initialize the module with quantization tables. Let me know what you think about this. Maybe in a future PR?

kornia/enhance/jpeg.py
ChristophReich1996: `torch.pi` directly links to `math.pi` I guess...

kornia/enhance/jpeg.py
ChristophReich1996: Don't know what you mean by this... Do you mean adding an example image pair?

kornia/enhance/jpeg.py
ChristophReich1996: I added the option to specify the quantization tables (fixed and as parameters) in constructor, just in case someone wants to learn the quantization tables.

ChristophReich1996:
I added some more doc examples (use of QT's etc.), changed the module implementation (added an additional test for this), and addressed the comments. Let me know if there is still something to do. Thanks for all the feedback, always happy to learn something here :)

ChristophReich1996:
Tests failed due to the changed import. Should I change it back?

Screenshot 2024-01-24 at 21 07 51

johnnv1:
> Tests failed due to the changed import. Should I change it back?

you can, but we will need to cherry-pick your changes and redo the PR to put it on top of the main, not the master. So will need to redo this import change

I can redo the PR later if you wish, but if you can do it that would be even better :)

(just changing the target branch will cause all commit history as conflict)

ChristophReich1996:
Ah Ok as you like. I have run the test locally so I'm pretty sure they won't fail so we can also keep it this way...

kornia/enhance/jpeg.py
johnnv1: done

Issue #2710, opened 2024-01-02
CI is always installing the latest torch


johnnv1:
As pointed out by edgar on Slack, the CI is installing the latest torch version instead of the desired one.

Our tests [CI](https://github.com/kornia/kornia/blob/master/.github/workflows/tests.yml) consist of a reusable workflow used to generate PR tests, and scheduled tests using multiple pytorch versions. Kornia is supporting torch >= 1.9.1 to its latest version. At PR's we want to test against the oldest and the newest versions of pytorch, and into the scheduled tests we want to test against all supported versions.

To do this, we first set up the environment: install pytorch, and then the kornia within x and development dependencies. To test against different versions of pytorch, first, we install a supported pytorch, so when installing kornia, the PyTorch installation should be ignored (because it already has a torch version installed), but this isn't always occurring at the moment as in https://github.com/kornia/kornia/actions/runs/7382077637/job/20081393018#step:3:277

![image](https://github.com/kornia/kornia/assets/20444345/859bdf23-2bb9-4e12-ab83-8ff1cd2e6614)

______________________________________________________________________

#### Consider also to contribute to Kornia universe projects :)

<sub>

- [**Tutorials**](https://github.com/kornia/tutorials): our repository containing the tutorials.

</sub>

pmeier:
You have `accelerate` in your `x` dependencies:
- https://github.com/kornia/kornia/blob/ce434e467faf617604bb3383cf78cd0b79f59dbd/.github/actions/env/action.yml#L38
- https://github.com/kornia/kornia/blob/ce434e467faf617604bb3383cf78cd0b79f59dbd/requirements/requirements-x.txt#L1

And that depends on `torch>=1.10.0`: https://github.com/huggingface/accelerate/blob/6f05bbd41a179cc9a86238c7c6f3f4eded70fbd8/setup.py#L57

Meaning, `pip` first detects that `torch>=1.9.1` from `kornia` is satisfied, but after pulling in `accelerate` this is no longer sufficient. Thus, it tries the latest release next.

This is why the other workflows work just fine: pre-installing `torch>=1.10.0` is sufficient for `kornia` and `accelerate`.

johnnv1:
> You have `accelerate` in your `x` dependencies:

wow, thanks I have totally forgotten about accelerate

PR #2763, opened 2024-01-26
Ci: fix old torch install

johnnv1:
fixes https://github.com/kornia/kornia/issues/2710

- unpin accelerate
- updateci


johnnv1:
What would be the option for torch 1.9.1 to:

- meshgrid with indexing (cc @edgarriba @cjpurackal)
- torch inf

image

edgarriba (approved):

Issue #2559, opened 2023-09-22
[milestone] Onnx support tracker


edgarriba:
We would like to add full support to ONNX so that we can import/export the kornia operators to production environments and make it composable.

Roadmap:
1. Create a base functionality in tests so that we automatically wrap functionals and test export
2. Create an API centric to onnx for inference
3. Create an API centric to import onnx models and finetune them
4. Expand the functionality above to compile down the major platforms: intel myriad, nvidia tensorrt, tf coral

johnnv1:
we should also start supporting the beta feature: https://pytorch.org/docs/stable/onnx_dynamo.html

briliantnugraha:
Hi @johnnv1 @edgarriba , could I know, is the warp_perspective supported to onnx now?

johnnv1:
> Hi @johnnv1 @edgarriba , could I know, is the warp_perspective supported to onnx now?

i don't think so, the function didn't receive any patch recently

edgarriba:
We'll update for kornia 2026 roadmap

PR #2776, opened 2024-02-02
Add [*, 3, H, W] support to Diff. JPEG

ChristophReich1996:
This pull request adds support for `[*, B, H, W]` images in Diff. JPEG coding. Additional test cases have been implemented to test different shapes. There are also some very minor cleanups in tests and error messages.

#### Type of change

- [x] 📚  Documentation Update
- [x] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/kornia/enhance/jpeg.py
+++ b/kornia/enhance/jpeg.py
@@ -13,6 +13,7 @@ from kornia.core.check import (
     KORNIA_CHECK_SHAPE,
 )
 from kornia.geometry.transform.affwarp import rescale
+from kornia.utils.image import perform_keep_shape_image
 
 __all__ = ["jpeg_codec_differentiable", "JPEGCodecDifferentiable"]
 
@@ -395,6 +396,7 @@ def _jpeg_decode(
     return rgb_decoded
 
 
+@perform_keep_shape_image
 def jpeg_codec_differentiable(
         image_rgb: Tensor,
         jpeg_quality: Tensor,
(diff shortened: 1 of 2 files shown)


kornia/enhance/jpeg.py
```diff
@@ -503,11 +503,32 @@ def jpeg_codec_differentiable(
...
-        f"JPEG quality is out of range. Expected range is [0, 99], "
+        f"JPEG quality is out of range. Expected range is [0, 100], "
         f"got [{jpeg_quality.amin().item()}, {jpeg_quality.amax().item()}]. Consider clipping jpeg_quality.",
     )
-    # Get original shape
-    H, W = image_rgb.shape[2:]
+    # Save original shape
+    original_shape = image_rgb.shape
+    # Reshape to [B, 3, H, W]
+    image_rgb = image_rgb.reshape(-1, 3, *original_shape[-2:])
```
edgarriba, line 512: one alternative here is to reuse a decorator we have in utils `@perform_keep_shape_image` -- you can see an example for the resize function: https://github.com/kornia/kornia/blob/main/kornia/geometry/transform/affwarp.py#L525

we haven't been using it very much across of the library because its lack of support to torchscript, but with the recent version of pytorch this seems to be sustainable improved. /cc @johnnv1 to discuss about adopting it more

johnnv1: Using the `perform_keep_shape_image` will make it incompatible with jit (kinda going against the onnx improvements) -- but it looks to be supported by the torch compile, which will replace jit maybe in a year

So maybe we can go with or support jit or supporting compile, with a preference for compile? https://pytorch.org/docs/stable/export.html#existing-frameworks

johnnv1 (approved):

kornia/enhance/jpeg.py
ChristophReich1996: Used the `perform_keep_shape_image` decorator. I guess `torch.compile` support might be more useful. Thanks for the feedback :)

johnnv1 (approved):

kornia/enhance/jpeg.py
johnnv1: cross-linking to https://github.com/kornia/kornia/issues/2559

PR #2789, opened 2024-02-08
Ensure absolute imports

johnnv1:
This patch bans relative imports, and moves all our relative imports to absolute imports using ruff

based on https://docs.astral.sh/ruff/settings/#lint_flake8-tidy-imports_ban-relative-imports

To be merged after #2788


edgarriba (commented):
Worth to double check that the internal imports are really absolutely pointing to the deepest submodule possible. We can trim after once we make sure and control over the circular dependencies

johnnv1:
> Worth to double check that the internal imports are really absolutely pointing to the deepest submodule possible. We can trim after once we make sure and control over the circular dependencies

We should manually review each one for it, or do you know how to automate it?

kornia/augmentation/__init__.py
```diff
@@ -1,89 +1,86 @@
...
-    RandomTranslate,
-    RandomTransplantation,
-    RandomVerticalFlip,
-    Resize,
-    SmallestMaxSize,
-)
-from kornia.augmentation._2d.base import AugmentationBase2D, RigidAffineAugmentationBase2D
+from kornia.augmentation import auto
+from kornia.augmentation import container
+from kornia.augmentation._2d import CenterCrop
```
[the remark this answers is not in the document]
johnnv1: i don't understand what do you mean

kornia/geometry/calibration/undistort.py
```diff
@@ -5,12 +5,11 @@
 import torch
 
 from kornia.core import stack
+from kornia.geometry.calibration.distort import distort_points, tilt_projection
 from kornia.geometry.linalg import transform_points
 from kornia.geometry.transform import remap
```
edgarriba, line 10: @johnnv1 just as random example, this import is not full absolute to where it’s actually implemented. I’m pretty sure there way more in code base

edgarriba (approved):

kornia/geometry/calibration/undistort.py
johnnv1: got it, needs to review/check each one

PR #2794, opened 2024-02-09
Refactor and add more palettes of colormaps

vgilabert94:
#### Changes

Refactor colormaps. Add the posibility to use ColorMap class to load bult-in colormaps or generate news.
Add colormap_data.py with all of data of colormaps (default num_colors=64).
Add a dict to manage colormaps.

List of colormaps: ['autumn', 'bone', 'jet', 'winter', 'rainbow', 'ocean', 'summer', 'spring', 'cool', 'hsv', 'brg', 'pink', 'hot', 'plasma', 'viridis', 'cividis', 'twilight', 'turbo', 'seismic']
This data of colormaps has been obtained based on the library matplotlib.
output

Fixes # (issue)

#### Type of change

- [x] 📚  Documentation Update
- [x] 🧪 Tests Cases

#### Checklist


kornia/color/colormap_data.py
```diff
@@ -0,0 +1,1263 @@
+from typing import List
```
johnnv1, line 1: rename it to _colormap_data.py

kornia/color/colormap_data.py
```diff
@@ -0,0 +1,1263 @@
...
+
+# All colormaps has 64 base colors. Extracted from matplotlib.
+# https://matplotlib.org/stable/users/explain/colors/colormaps.html
+
+
+RGBColor = List[float]
+
+
+_AUTUMN_BASE: List[RGBColor] = [
```
johnnv1, line 10: also create functions to return each cmap base, this way these lists aren't created on import time

kornia/color/colormap.py
```diff
@@ -1,126 +1,121 @@
...
 from torch.nn.functional import interpolate
 
+from kornia.color import colormap_data as cm_data
 from kornia.core import Module, Tensor, tensor
 from kornia.core.check import KORNIA_CHECK_IS_GRAY
 
-RGBColor = List[float]
+RGBColor = cm_data.RGBColor
+
+colormap_options = {
```
johnnv1, line 13: i think it can go on the init of the class

kornia/color/colormap.py
```diff
@@ -1,126 +1,121 @@
...
-    [1.0, 0.9365079365079365, 0.0],
-    [1.0, 0.9523809523809523, 0.0],
-    [1.0, 0.9682539682539683, 0.0],
-    [1.0, 0.9841269841269841, 0.0],
-    [1.0, 1.0, 0.0],
-]
-
-
-# Generate complete color map
-class AUTUMN(ColorMap):
```
johnnv1, line 114: can you somehow make it still works, and add a deprecated warning about it?

johnnv1: Ideally we shouldn't remove a functionality without minimal warning for kornia users

johnnv1: https://github.com/kornia/kornia/blob/c8753f51c742cf1441b43fca94df1ce9a95a2cba/kornia/utils/helpers.py#L83

kornia/color/colormap.py
edgarriba: and i would have somehow defined as a mapping with keys as  Enum instead of string to be less error prone. With new syntaxes you can initialize enums from strings for `rgb = ColorPalette["autumn"]`

kornia/color/colormap_data.py
vgilabert94: I'm not sure what do you mean. Could you give me an example?

kornia/color/colormap.py
vgilabert94: Like a wrapper?
class AUTUMN(ColorMap) -> inside calling ColorMap('autumn')

johnnv1: exactly, just keeping it working as it was in the last version and internally it can just call the new functions/module within the deprecated warning

kornia/color/colormap.py
johnnv1: oh, it's true, +1 enums

kornia/color/colormap_data.py
johnnv1: Instead of declaring it in the body of the file, put each one of these lists inside a getter function this way, the lists will be "created" just when we call the function, otherwise, python will "create"/"allocate" this lists when we import this file... 
```suggestion
def get_autumn_base():
    return [
    ...
    ]
```

johnnv1 (commented):
pretty cool patch

kornia/color/colormap.py
vgilabert94: How do I have to do it to use Enum?

kornia/color/colormap.py
```diff
@@ -1,126 +1,121 @@
...
+    "brg": cm_data._BRG_BASE,
+    "pink": cm_data._PINK_BASE,
+    "hot": cm_data._HOT_BASE,
+    "plasma": cm_data._PLASMA_BASE,
+    "viridis": cm_data._VIRIDIS_BASE,
+    "cividis": cm_data._CIVIDIS_BASE,
+    "twilight": cm_data._TWILIGHT_BASE,
+    "turbo": cm_data._TURBO_BASE,
+    "seismic": cm_data._SEISMIC_BASE,
+}
```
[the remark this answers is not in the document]
johnnv1: But, how each colormap data, should be in a `getter` function, I think is best just to put a bunch of ifs's here or something like that, otherwise will initialize all base data unnecessarily

kornia/color/colormap.py
```diff
@@ -173,20 +168,26 @@ def apply_colormap(input_tensor: Tensor, colormap: ColorMap) -> Tensor:
 
 
 class ApplyColorMap(Module):
-    r"""Module that apply to a gray tensor (integer tensor) a colormap.
+    r"""Class for applying a colormap to images.
```
johnnv1, line 171: this still just working on gray uint images, no?

johnnv1 (commented):
Just to simplify the API... The initial idea is the ColorMap is just an abstract class to be a cmap generator, but, toward your implementation, it will be an individual colormap itself, instead of having a class for each colormap.

So, I think we can have an Enum for mapping the available CMAPs, and the ColorMap, be a class that accepts a `base` parameter, which can be a of the CMAPs or a list of colors to be used as the base data

This way we can aim to use like:
```python
import torch
from kornia.color import ColorMap, CMAP, apply_colormap, ApplyColormap
from testing.base import assert_close

device = torch.device('cpu')
dtype = torch.float32
max_value = 255
data = torch.randint(low=0, high=max_value, size=(1, 1, 128, 128), device=device, dtype=dtype)

cmap_base = [
    [1.0, 0.0, 0.0],
    [1.0, 0.015873015873015872, 0.0],
    [1.0, 0.031746031746031744, 0.0],
    [1.0, 0.047619047619047616, 0.0],
]

bases_cmaps = ['autumn', CMAP.autumn, cmap_base]
for base in bases_cmaps:
    cmap = ColorMap(base, max_value + 1, device=device, dtype=dtype)
    # By the function
    colored_a = apply_colormap(data, cmap)

    # Or using the Module within cmap
    colorize = ApplyColormap(cmap)
    colored_b = colorize(data)

    # If want to support `ApplyColormap` with auto init of the cmap
    colorize = ApplyColormap.from_base(base=base, num_colors=max_value + 1, device=device, dtype=dtype)
    colored_c = colorize(data)

    # Sanity test
    assert_close(colored_a, colored_b)
    assert_close(colored_a, colored_c)
```

vgilabert94:
I'm not sure what happend when I push changes but i add more commits and i dont know how to clean this branch :(

Regarding the last commit ([New changes to colormaps](https://github.com/kornia/kornia/pull/2794/commits/c800d953b48030fff167eabd73885165e8a4d5b0)
:
- CMAP class created using Enum.
- ColorMap class for load and generate colormaps. Input could be CMAP, str or List[RGBColor].
- apply_colormap: same functionality as before. Not working with non-gray images. We can discuss what new functionality you want and I can add it. For now I have checked to use images (B,3,H,W) and the result is tensors (3,B,3,W,H), one colormap per channel is created and I'm not sure if this is what you are looking for.
- ApplyColorMap: is a wrapper version of apply_colormap in class mode.

johnnv1:
>     * apply_colormap: same functionality as before. Not working with non-gray images. We can discuss what new functionality you want and I can add it. For now I have checked to use images (B,3,H,W) and the result is tensors (3,B,3,W,H), one colormap per channel is created and I'm not sure if this is what you are looking for.


yeah, I expected something like that since the apply function works as an indexing function between mask values and the cmap  -- it's fine the way it's. We can discuss for a further PR to have better behavior here

kornia/color/_colormap_data.py
```diff
@@ -0,0 +1,1263 @@
...
+
+_AUTUMN_BASE: List[RGBColor] = [
+    [1.0, 0.0, 0.0],
+    [1.0, 0.015873015873015872, 0.0],
+    [1.0, 0.031746031746031744, 0.0],
+    [1.0, 0.047619047619047616, 0.0],
+    [1.0, 0.06349206349206349, 0.0],
+    [1.0, 0.07936507936507936, 0.0],
+    [1.0, 0.09523809523809523, 0.0],
+    [1.0, 0.1111111111111111, 0.0],
```
johnnv1, line 18: can you wrap these base values under functions? something like 
```suggestion
def get_autumn_base() -> List[RGBColor]:
    return [
    [1.0, 0.0, 0.0],
    [1.0, 0.015873015873015872, 0.0],
    [1.0, 0.031746031746031744, 0.0],
    [1.0, 0.047619047619047616, 0.0],
    [1.0, 0.06349206349206349, 0.0],
    [1.0, 0.07936507936507936, 0.0],
    [1.0, 0.09523809523809523, 0.0],
    [1.0, 0.1111111111111111, 0.0],
```

kornia/color/colormap.py
```diff
@@ -1,126 +1,152 @@
...
-            mode="linear",
-        )[0, ...]
+        if isinstance(base, str):
+            base = base.lower()
+            if base not in CMAP.list():
+                raise ValueError(f"Unsupported colormap: {base}. Available colormaps are {CMAP.list()}")
+            base_colormap_data = CMAP[base]._load_base()
+            self.name = base
+        elif isinstance(base, CMAP):
+            # Check if CMAP.base exist in enum CMAP? How?
```
johnnv1, line 119: what do you mean?

kornia/color/colormap.py
```diff
@@ -1,126 +1,152 @@
...
+            "plasma": cm_data._PLASMA_BASE,
+            "viridis": cm_data._VIRIDIS_BASE,
+            "cividis": cm_data._CIVIDIS_BASE,
+            "twilight": cm_data._TWILIGHT_BASE,
+            "turbo": cm_data._TURBO_BASE,
+            "seismic": cm_data._SEISMIC_BASE,
+        }[self.name]
+
+    @classmethod
+    def list(cls):
```
johnnv1, line 66:
```suggestion
    def list(cls) -> list[str]:
```

kornia/color/colormap.py
```diff
@@ -1,126 +1,152 @@
...
+        r"""Returns a list of names of enumeration members.
+
+        Returns:
+            List[str]: A list containing the names of enumeration members.
```
johnnv1, line 70:
```suggestion
            A list containing the names of enumeration members.
```

kornia/color/colormap.py
```diff
@@ -197,9 +230,40 @@ class ApplyColorMap(Module):
...
+        Returns:
+            Tensor: The output tensor representing the image with the applied colormap.
+        """
         return apply_colormap(input_tensor, self.colormap)
+
+
+# Generate complete color map
+@deprecated(
+    "0.7.2",
+    extra_reason="The AUTUMN() class is deprecated and will be removed next kornia versions (0.8.0 - dec 2024).",
```
johnnv1, line 258:
```suggestion
    extra_reason="The AUTUMN() class is deprecated and will be removed next kornia versions (0.8.0 - dec 2024). In favor of using `ColorMap(base='autumn')` instead",
```

kornia/color/colormap.py
```diff
@@ -173,20 +199,27 @@ def apply_colormap(input_tensor: Tensor, colormap: ColorMap) -> Tensor:
 
 
 class ApplyColorMap(Module):
-    r"""Module that apply to a gray tensor (integer tensor) a colormap.
+    r"""Class for applying a colormap to images.
 
     Args:
```
johnnv1, line 204: can you add a sample image here? need to add an entry in `docs/generate_examples.py` -- can be something like the example in the description of your PR

kornia/color/colormap.py
vgilabert94: If we recieve an input "CMAP.test" that it does not exist in our enum, what happen? Do we need to check if it exists?

kornia/color/_colormap_data.py
vgilabert94: Yes, i forget to do this.  If I add a function in CMAP (enum), I will have to call these functions, right?

```
class CMAP(Enum):
    autumn = 1
    bone = 2
    jet = 3
    def _load_base(self) -> RGBColor:
        return {
            "autumn": get_autum_base(),
            "bone": get_bone_base(),
            "jet": get_jet_base(),
        }[self.name]
```

johnnv1: yes, but calling function call can be on return 

```python
class CMAP(Enum):
    autumn = 1
    bone = 2
    jet = 3
    def _load_base(self) -> RGBColor:
        return {
            "autumn": get_autum_base,
            "bone": get_bone_base,
            "jet": get_jet_base,
        }[self.name]()
```

kornia/color/colormap.py
johnnv1: I'm assuming that if it's available in the enum, it will have an associated base. If the value/map doesn't exist in the enum, I don't think it's possible to create it.

johnnv1:
> So, I add the functionality for not gray images? Will the functionality be added for rgb images or for more types (B,2,H,W) or others? (I'm not sure if this makes sense) I will need to check the data type (uint in range [0,255]) isn't it? Any function already implemented in kornia for this?

if you want to work on it, let's do it in another PR, because I believe it needs better discussion about -- I'm not sure what we should aim for (about the Chanel dimension and the datatype)  -- make it work for images in other range has more meaning and importance for me  at least (the Kornia images input are standardized to [0-1])

docs/source/color.rst
```diff
@@ -202,18 +202,21 @@ Sepia
...
+    :members:
+    :undoc-members:
+    :member-order: bysource
 
 
 Functions and modules to use the color maps:
 
 .. autofunction:: apply_colormap
 .. autoclass:: ApplyColorMap
+.. autoclass:: AUTUMN
```
johnnv1, line 222: it's deprecated anyway
```suggestion
```

kornia/color/colormap.py
```diff
@@ -1,145 +1,179 @@
...
+    pink = 12
+    hot = 13
+    plasma = 14
+    viridis = 15
+    cividis = 16
+    twilight = 17
+    turbo = 18
+    seismic = 19
+
+    def _load_base(self) -> RGBColor:
```
johnnv1, line 41:
```suggestion
    def _load_base(self) -> List[RGBColor]:
```

kornia/color/colormap.py
```diff
@@ -1,145 +1,179 @@
...
+    """
 
     def __init__(
         self,
-        base_colormap: List[RGBColor],
-        num_colors: int,
-        device: Optional[torch.device],
-        dtype: Optional[torch.dtype],
+        base: Union[List[RGBColor], str, CMAP],
+        num_colors: Optional[int] = 64,
```
johnnv1, line 109:
```suggestion
        num_colors: int = 64,
```

kornia/color/colormap.py
```diff
@@ -1,145 +1,179 @@
...
+        >>> ColorMap(base='viridis', num_colors=8).colors
+        tensor([[0.2813, 0.2621, 0.2013, 0.1505, 0.1210, 0.2463, 0.5259, 0.8557],
+                [0.0842, 0.2422, 0.3836, 0.5044, 0.6258, 0.7389, 0.8334, 0.8886],
+                [0.4072, 0.5207, 0.5543, 0.5574, 0.5334, 0.4519, 0.2880, 0.0989]])
+
+        # Create color map from first color (RGB with range[0-1]) to last one with num_colors length.
+        >>> ColorMap(base=[[0., 0.5 , 1.0], [1., 0.5, 0.]], num_colors=8).colors
+        tensor([[0.0000, 0.0000, 0.1250, 0.3750, 0.6250, 0.8750, 1.0000, 1.0000],
+                [0.5000, 0.5000, 0.5000, 0.5000, 0.5000, 0.5000, 0.5000, 0.5000],
+                [1.0000, 1.0000, 0.8750, 0.6250, 0.3750, 0.1250, 0.0000, 0.0000]])
```
johnnv1, line 103:
```suggestion
    >>> ColorMap(base='viridis', num_colors=8).colors
    tensor([[0.2813, 0.2621, 0.2013, 0.1505, 0.1210, 0.2463, 0.5259, 0.8557],
            [0.0842, 0.2422, 0.3836, 0.5044, 0.6258, 0.7389, 0.8334, 0.8886],
            [0.4072, 0.5207, 0.5543, 0.5574, 0.5334, 0.4519, 0.2880, 0.0989]])

    # Create color map from first color (RGB with range[0-1]) to last one with num_colors length.
    >>> ColorMap(base=[[0., 0.5 , 1.0], [1., 0.5, 0.]], num_colors=8).colors
    tensor([[0.0000, 0.0000, 0.1250, 0.3750, 0.6250, 0.8750, 1.0000, 1.0000],
            [0.5000, 0.5000, 0.5000, 0.5000, 0.5000, 0.5000, 0.5000, 0.5000],
            [1.0000, 1.0000, 0.8750, 0.6250, 0.3750, 0.1250, 0.0000, 0.0000]])
```

johnnv1 (commented):
this should address the failing cases on typing and docs ci

kornia/color/colormap.py
```diff
@@ -1,145 +1,180 @@
-from abc import ABC
-from typing import List, Optional
+from __future__ import annotations
```
johnnv1, line 1:
```suggestion
from __future__ import annotations
from typing import List, Optional, Union
```

kornia/color/colormap.py
```diff
@@ -1,145 +1,181 @@
...
+
+from enum import Enum
+from typing import List, Optional, Union
```
johnnv1, line 4: @vgilabert94 could you update the typing from `List` to `list` on this file? Looks like the ruff didn't do it on the safe fixes now

kornia/color/colormap.py
```diff
@@ -1,145 +1,182 @@
...
+    Returns:
+        An object of the colormap with the num_colors length.
+
+    Examples:
+        >>> ColorMap(base='viridis', num_colors=8).colors
+        tensor([[0.2813, 0.2621, 0.2013, 0.1505, 0.1210, 0.2463, 0.5259, 0.8557],
+                [0.0842, 0.2422, 0.3836, 0.5044, 0.6258, 0.7389, 0.8334, 0.8886],
+                [0.4072, 0.5207, 0.5543, 0.5574, 0.5334, 0.4519, 0.2880, 0.0989]])
+
+        >>> # Create color map from first color (RGB with range[0-1]) to last one with num_colors length.
```
johnnv1, line 102: This will be better displayed 
```suggestion
    Create a color map from the first color (RGB with range[0-1]) to the last one with num_colors length.
```

kornia/color/colormap.py
```diff
@@ -1,145 +1,182 @@
...
 
+import kornia.color._colormap_data as cm_data
+from kornia.color._colormap_data import RGBColor
 from kornia.core import Module, Tensor, tensor
 from kornia.core.check import KORNIA_CHECK_IS_GRAY
+from kornia.utils.helpers import deprecated
 
-RGBColor = List[float]
 
+class CMAP(Enum):
```
edgarriba, line 16: just for consistency what do you think about renaming this to `ColorMapType` and to enforce explicit names

edgarriba (approved):
LGTM. Just a naming convention suggestion.

johnnv1 (approved):

PR #2797, opened 2024-02-10
test suite: skip fp64 canny dynamo test

johnnv1:
Skip canny test for float64 due to precision error https://github.com/kornia/kornia/actions/runs/7841025684/job/21396613672#step:2:11053

Trying to get all green light on the scheduled CI

image

### DIFF
--- a/tests/filters/test_canny.py
+++ b/tests/filters/test_canny.py
@@ -300,7 +300,7 @@ class TestCanny(BaseTester):
     @pytest.mark.skipif(torch_version() in {"2.0.0", "2.0.1"}, reason="Not working on 2.0")
     def test_dynamo(self, batch_size, kernel_size, device, dtype, torch_optimizer):
         if (
-            torch_version() in {"2.1.1", "2.1.2"}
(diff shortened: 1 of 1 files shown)


tests/filters/test_canny.py
```diff
@@ -300,7 +300,7 @@ def test_module(self, device, dtype):
     @pytest.mark.skipif(torch_version() in {"2.0.0", "2.0.1"}, reason="Not working on 2.0")
     def test_dynamo(self, batch_size, kernel_size, device, dtype, torch_optimizer):
         if (
-            torch_version() in {"2.1.1", "2.1.2"}
+            torch_version() in {"2.1.1", "2.1.2", "2.2.0"}
```
edgarriba, line 303: Should this use the decorator instead to skip versions ? It’s in the two places now

johnnv1: in the decorator, we cannot check the dtype... and I keep it waiting for it to work in a future release

edgarriba (approved):

Issue #941, opened 2021-04-03
[Feat] Kornia Augmentations tracker


edgarriba:
## More Than Augmentation
In a longer vision, our augmentation module will be more focused on differentiability, with an aim of assisting both training and deployment. We can see there are more researches and applications in this area, with or without Kornia. This module will be the lead. In our next version, we gonna heavily improve:
- the API design and better readability.
- Advanced features like JIT and AMP.
- Better differentiability and more differentiable augmentations.

## 🚀 Feature
Specs for the `kornia.augmentation` module.

- Make it easy to contrib through `AugmentationBase2D`.
- Give support to `torch.jit.script` and `torch.float16`.
- Promote to embed the full code in the main class to keep readability.
- Documentation improvement:
  - Better looking
  - More examples for copying and pasting.
  - Provide a benchmark to notify users that we are better.
- Add `register_buffer` to save static parameters inside ```nn.Module```.

## Motivation

This should be the template to implement a custom augmentation

```python
class RandomHorizontalFlip(AugmentationBase2D):

    def compute_transformation(self, input: torch.Tensor) -> torch.Tensor:
        w: int = input.shape[-1]
        flip_mat: torch.Tensor = torch.tensor([[-1, 0, w - 1],
                                               [0, 1, 0],
                                               [0, 0, 1]], device=input.device, dtype=input.dtype)
        return flip_mat.repeat(input.size(0), 1, 1)

    def apply_transform(self, input: torch.Tensor, params: Dict[str, torch.Tensor]) -> torch.Tensor:
        return kornia.hflip(input)
```

## User Stories
- _As a user, I want …, so that …_ 
- As a user, I want to decide whether the transforms return the transformation, so that I can use it to chain and potentially apply in sequential.
- As a user, I want to use the augmentations with half precision, so that I can also combine with AMP.
- As a user, I want to be able to run the augmentations with xla devices, so that I can run on TPUs.
- As a user, I want to be able to torchscript the augmentations, so that I can easily serialise and export.
- As a developer, I want to have easy examples of custom augmentations, so that I can contribute with more.

## TODO:
- [x] Refactor engine to deprecate `return_transform` to return only `torch.Tensor` @shijianjian 
  - Transforms will be accessible through class properties
- [ ] Enable torchscript for augmentations https://github.com/kornia/kornia/issues/477
- [x] Remove functional code @shijianjian 
- [x] Fix `RandResizedCrop` critical bug https://github.com/kornia/kornia/issues/927

[...]

PR #2800, opened 2024-02-12
feat: augmentation sequential supports mask (B, H, W) with images (B, C, H, W)

johnnv1:
#### Changes
Fixes the handler of (*, H, W) which accepts either (H, W), (C, H, W) or (*, C, H, W).

related to #941

The base augmentations already have a `_transform_input` which handler its, but it's being callable after `validate_tensor` which have a shape checker inside , which didn't allow we use inputs with (H, W), (C, H, W) in the augmentations API

#### Type of change

- [ ] 📚  Documentation Update
- [x] 🧪 Tests Cases


kornia/augmentation/_2d/base.py
```diff
@@ -35,8 +35,9 @@ def validate_tensor(self, input: Tensor) -> None:
 
     def transform_tensor(self, input: Tensor) -> Tensor:
         """Convert any incoming (H, W), (C, H, W) and (B, C, H, W) into (B, C, H, W)."""
+        input = _transform_input(input)
         _validate_input_dtype(input, accepted_dtypes=[float16, float32, float64])
-        return _transform_input(input)
+        return input
```
johnnv1, line 40: I made a mistake with something here, will check later

shijianjian: what is happening here? should not the tensor be transformed?

johnnv1: This is fine as it was... i misread the `_validate_input_dtype` with `self.validate_tensor`

I'll check soon if it's working on the sequential container, which is i got the original issue from

tests/augmentation/test_container.py
```diff
@@ -516,6 +516,40 @@ def test_class(self, device, dtype):
         _, out_labels = aug(img, labels)
         assert labels is out_labels
 
+    @pytest.mark.parametrize("B", [None, 1, 3, 5])
+    @pytest.mark.parametrize("C_i", [1, 3])
+    @pytest.mark.parametrize("C_m", [None, 1])
+    @pytest.mark.parametrize("keepdim", [True, False])
+    def test_masks_without_channel_dim(self, device, dtype, B, C_i, C_m, keepdim):
+        img_shape = (B, C_i, 1000, 500) if B else (C_i, 1000, 500)
+        mask_shape = tuple(x for x in (B, C_m, *img_shape[-2:]) if x)
```
johnnv1, line 525: @shijianjian 

The idea here is to make it work with masks `B, H, W` or `H, W` and images `B, C, H, W`

tests/augmentation/test_container.py
```diff
@@ -516,6 +516,40 @@ def test_class(self, device, dtype):
...
+            # K.AugmentationSequential(
+            #     K.RandomAffine(360, p=1.0),
+            #     data_keys=["input", "mask"],
+            # ),
+            K.RandomAffine(360, p=1.0),
+            data_keys=["input", "mask"],
+            keepdim=keepdim,
+        )
+
+        aug.keepdim = keepdim  # FIXME(@shijianjian): This shouldn't be necessary
```
johnnv1, line 541: is this a known bug, or should it be working?

Also, if we use inside the `AugmentationSequential` items like `ImageSequential` and `AugmentationSequential` looks like they do not receive the `keepdim` even with the manual set

kornia/augmentation/utils/helpers.py
```diff
@@ -100,7 +102,12 @@ def _transform_input(input: Tensor) -> Tensor:
         input = input.unsqueeze(0)
 
     if len(input.shape) == 3:
-        input = input.unsqueeze(0)
+        if batch_size == input.shape[0]:
```
johnnv1, line 105: I believe this can be tricky, but if we go with it only for masks may it be fine -- same comment for the _transform_output_shape

johnnv1:
The typing will be failing for now, because I just updated the 2d case (and the mixin + 3d parts will cry), to see if this patch makes sense

isaaccorley:
+1 for this feature!

kornia/augmentation/utils/helpers.py
shijianjian: This function looks very weird. Probably having a `transform_input_by_tensor(input: Tensor, reference_tensor: Tensor)`?

johnnv1: yeah, will try to do it on Sunday

johnnv1: Do we have the reference tensor in the `params`? 

What I used is the forward input shape from it

image

tests/augmentation/test_container.py
shijianjian: Either way is fine I think. Probably remove it since I think a user may expect 3D in 3D out, or 4D in 4D out. It might be redundant.

kornia/augmentation/utils/helpers.py
```diff
@@ -82,6 +82,34 @@ def _infer_batch_shape3d(input: Union[Tensor, Tuple[Tensor, Tensor]]) -> torch.S
...
+    C = reference_shape[-3] if len(reference_shape) >= 3 else None
+
+    if len(input.shape) == 2:
+        input = input.unsqueeze(0)
+
+    if len(input.shape) == 3 and B == input.shape[-3]:
+        # If the first dim matches within the batch_size, add a `C` dim
+        # Useful to handler Masks without `C` dimensions
+        input = input.unsqueeze(1)
+
```
johnnv1, line 103: should the unsqueeze(0) be needed in an else here I think -- tests didn't cover it? 🤔

shijianjian: It is fine if we only support HW, CHW, BHW. We did not think about BHW cases I think.

shijianjian: Ah. I got your question. No, the logic is correct. If we have HW, it will be firstly transformed into CHW. Then BCHW

tests/augmentation/test_container.py
johnnv1: without this manual set after initializing the `AugmentationSequential` the behavior didn't work...

shijianjian (approved):

PR #2820, opened 2024-02-29
Upgrade model checkpointing

machineko:
#### Changes
Fixed both EarlyStopping and ModelCheckpoint to work as expected in other frameworks so EarlyStop and ModelCheckpoint use decreasing metrics as default.
Also fix documentation with more verbose example.

#### Type of change

- [ x ] 📚  Documentation Update
- [ x ] 🧪 Tests Cases
- [ x ] 🚨 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ x ] 📝 This change requires a documentation update

#### Checklist

- [ x ] My code follows the style guidelines of this project
- [ x ] I have performed a self-review of my own code
- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] I have made corresponding changes to the documentation
- [ x ] My changes generate no new warnings


kornia/x/callbacks.py
```diff
@@ -72,35 +87,50 @@ class ModelCheckpoint:
...
+    def __init__(
+        self, filepath: str, monitor: str, filename_fcn: Optional[Callable[..., str]] = None, max_mode: bool = False
+    ) -> None:
         self.filepath = filepath
         self.monitor = monitor
         self._filename_fcn = filename_fcn or default_filename_fcn
-
         # track best model
-        self.best_metric: float = 0.0
+        self.best_metric: float = inf
```
johnnv1, line 121: This shouldn't be equals when max mode is false?

docs/source/get-started/training.rst
```diff
@@ -209,14 +209,26 @@ as follows passing as ``callbacks`` the classes :py:class:`~kornia.x.ModelCheckp
 
 .. code:: python
 
+    def my_evaluate(self):
+      # stats = StatsTracker()
```
edgarriba, line 213: add a comment or something to illustrate what users could do

kornia/x/callbacks.py
edgarriba: agree

edgarriba (approved):
thanks for the contribution! LGTM, just the two small suggestions and good to go

kornia/x/callbacks.py
machineko: added -inf for max_mode and inf for min_mode with small change in code logic to be cleaner

docs/source/get-started/training.rst
machineko: Added more verbose comment below but I'm not sure if I should add more to my_evaluate method or whole callback?

machineko:
I'm not sure how to fix. docstests & sphinx-build - python-3.11, torch-2.2.0 (pull_request) nor Tests on CPU (PR) / coverage / slow (float32) (pull_request)
So I'm leaving it for someone more familiar with kornia database everything else should be fine 😄

kornia/x/callbacks.py
```diff
@@ -72,35 +100,57 @@ class ModelCheckpoint:
     Args:
         filepath: the where to save the mode.
         monitor: the name of the value to track.
-
+        max_mode: if true metric will be multiply by -1
+                  turn this flag when increasing metric value is expected for example Accuracy
     **Usage example:**
 
     .. code:: python
+        def my_evaluate(self) -> dict[str, AverageMeter]:
```
edgarriba, line 108: for some reason the doctests error is because of this: https://github.com/kornia/kornia/actions/runs/8110317553/job/22167347988?pr=2820

i guess because you changed the signature of the Callback ?

machineko: It can be.
I should use Dict imported from typing over dict[] as type. As it won't work with older python versions will try it and ping if it will be w fix

machineko: Ok I just move to older docstring version don't know why it didn't compile before but it should work fine right now

Issue #2721, opened 2024-01-10
Gradient Illumination augmentations


edgarriba:
## 🚀 Feature

- Include some new augmentations based on gradient illumination (gaussian, radial, linear, etc) -> Modify illumination of image.
- Add objects to images (rectangle, ellipse, polygon, circle, line, etc.)
  - potentially using / expanding draw utils: https://github.com/kornia/kornia/issues/2164
_____________________________________________________________________

#### Consider also to contribute to Kornia universe projects :)

<sub>

- [**Tutorials**](https://github.com/kornia/tutorials): our repository containing the tutorials.

</sub>

PR #2823, opened 2024-03-04
feat: RandomLinearIllumination and RandomLinearCornerIllumination

vgilabert94:
#### Changes
Add Random illumination augmentation using linear mask.

RandomLinearIllumination:
image

RandomLinearCornerIllumination: 
image

related to https://github.com/kornia/kornia/issues/2721
related to https://github.com/kornia/kornia/issues/941

I'm creating the gradient mask for batching manually, I'm not sure how i can do it without a for loop. If exist an alternative I will change it.

#### Type of change

- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases

#### Checklist


kornia/augmentation/random_generator/_2d/linear_illumination.py
```diff
@@ -0,0 +1,176 @@
...
+            torch.tensor(-1),
+        ).to(device=_device, dtype=_dtype)
+
+        directions = _adapted_rsampling((batch_size, 1, 1, 1), self.directions_sampler, same_on_batch).to(
+            device=_device,
+            dtype=torch.int8,
+        )
+
+        gradient = torch.zeros(batch_shape)
+        for _b in range(batch_size):
```
johnnv1, line 78: why it's a loop over the batch?

vgilabert94: No, the batch and direction are independent. I have to create a loop for batch because linspace does not accept batch, is the only way that I thought. And direction in a simple random value between 0 and 4.

edgarriba: maybe you can workaround this with a meshgrid ?

vgilabert94: I'm seeing `meshgrid `, and only accept 0D or 1D array as input, so we can't work with channels or batch. I'm not sure how we can do it.

https://discuss.pytorch.org/t/batched-meshgrid/157206
https://stackoverflow.com/questions/72782751/how-to-perform-torch-meshgrid-over-multiple-tensors-in-parallel

johnnv1: may this do the trick? https://github.com/kornia/kornia/blob/886b9328ae4a1fff39acd953392de80dbc0fd468/kornia/utils/grid.py#L9

edgarriba: i think so -- side comment, can we make this performance improvement in a follow up PR ? @vgilabert94 we plan to release early next week and would be great to have this patch in.

johnnv1 (approved):

PR #2856, opened 2024-03-25
Support instance masks (N,H,W)

### LINKED ISSUE #2855: `AugmentationSequential` does not support instance masks shape (N, H, W)
### Describe the bug

Currently `AugmentationSequential` does not seem to support instance masks i.e. an image (C, H, W) having multiple masks (N, H, W). 

### Reproduction steps

The following code snippet errors out with a `RuntimeError: Expect (B, C, H, W). Got torch.Size([3, 256, 256]).` error.

```python
import torch

img = torch.randn(3, 256, 256)
masks = torch.ones( 2, 256, 256)

aug_dict = K.AugmentationSequential(
    K.RandomHorizontalFlip(p=1.0),
    data_keys=None,
    keepdim=True,
)
aug_dict.keepdim = True

data = {'image': img, 'mask': masks}
out = aug_dict(data)
```

### Expected behavior

Error should not be raised

### Environment

```shell
python collect_env.py

kornia==0.7.2
```
- PyTorch Version (e.g., 1.0):
- OS (e.g., Linux):
- How you installed PyTorch (`conda`, `pip`, source):
- Python version:
- CUDA/cuDNN version:
- GPU models and configuration:

ashnair1:
#### Changes

Fixes #2855 related to #941

#### Type of change

- [x] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/kornia/augmentation/container/ops.py
+++ b/kornia/augmentation/container/ops.py
@@ -1,6 +1,6 @@
 from abc import ABCMeta, abstractmethod
 from typing import Any, Callable, Dict, Generic, List, Optional, Type, TypeVar, Union
-
+import copy
 from typing_extensions import ParamSpec
 
 import kornia.augmentation as K
@@ -286,10 +286,11 @@ class MaskSequentialOps(SequentialOpsInterface[Tensor]):
         if isinstance(module, (K.GeometricAugmentationBase2D,)):
             tfm_input = []
             params = cls.get_instance_module_param(param)
-            for inp in input:
-                inp = module.transform_tensor(inp)
+            params_i = copy.deepcopy(params)
+            for i, inp in enumerate(input):
+                params_i["batch_prob"] =  params["batch_prob"][i]
                 tfm_inp = module.transform_masks(
-                    inp, params=params, flags=module.flags, transform=module.transform_matrix, **extra_args
+                    inp, params=params_i, flags=module.flags, transform=module.transform_matrix, **extra_args
                 )
                 tfm_input.append(tfm_inp)
             input = tfm_input
@@ -302,9 +303,10 @@ class MaskSequentialOps(SequentialOpsInterface[Tensor]):
         elif isinstance(module, (_AugmentationBase)):
             tfm_input = []
             params = cls.get_instance_module_param(param)
-            for inp in input:
-                inp = module.transform_tensor(inp)
-                tfm_inp = module.transform_masks(inp, params=params, flags=module.flags, **extra_args)
+            params_i = copy.deepcopy(params)
+            for i, inp in enumerate(input):
+                params_i["batch_prob"] =  params["batch_prob"][i]
+                tfm_inp = module.transform_masks(inp, params=params_i, flags=module.flags, **extra_args)
                 tfm_input.append(tfm_inp)
             input = tfm_input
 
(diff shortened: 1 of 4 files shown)


tests/augmentation/test_container.py
```diff
@@ -524,7 +524,7 @@ def test_class(self, device, dtype):
 
     @pytest.mark.parametrize("B", [None, 1, 3, 5])
     @pytest.mark.parametrize("C_i", [1, 3])
-    @pytest.mark.parametrize("C_m", [None, 1])
+    @pytest.mark.parametrize("C_m", [None, 1, 3])
```
johnnv1, line 527:
```suggestion
    @pytest.mark.parametrize("C_m", [None, 1, 3, 4])
```

kornia/augmentation/container/ops.py
```diff
@@ -286,10 +287,11 @@ def transform_list(
         if isinstance(module, (K.GeometricAugmentationBase2D,)):
             tfm_input = []
             params = cls.get_instance_module_param(param)
-            for inp in input:
-                inp = module.transform_tensor(inp)
+            params_i = copy.deepcopy(params)
```
edgarriba, line 290: Why do we need to make a copy here ?

edgarriba (approved):

kornia/augmentation/container/ops.py
ashnair1: To iteratively modify the batch_prob.

In batched cases, image would a tensor of shape (B, C, H, W) whereas masks would be a list of tensors [(N1, H, W), (N2, H, W)....].

`params["batch_prob"]` is set w.r.t to the image tensor, so we need to iterate over it and reassign the batch_prob passed while keeping the rest of the params the same

shijianjian: I think there should at least have an `if-else` block here. If it is a list, we enumerate them one-by-one. If it is a straight-away tensor, we do it all at once.

shijianjian (changes requested):
Meanwhile, what we did for the boxes is to merge them all into one tensor, which is
https://github.com/kornia/kornia/blob/46a5e40a0dbffcb38c7eca5b36b39479dcc0d85f/kornia/geometry/boxes.py#L20-L33

Not sure if we should do the same to the masks. Can anyone benchmark on which approach runs faster? Iteration or batching?

kornia/augmentation/container/ops.py
ashnair1: Not sure I follow. This change takes place in the `transform_list` function of `MaskSequentialOps`. So the input is a list of mask tensors.

ashnair1:
Merging a list of tensors into tensors will not work for instance segmentation. This is because each image will have a different number of objects and so different number of masks. Because of this, they cannot be batched. 

This is how it would look like:
**Single sample**: Image (C, H, W) and Mask (N, H, W)
**Batched**: Image (B, C, H, W) and Mask ([(N<sub>1</sub>, H, W), (N<sub>2</sub>, H, W) ... (N<sub>b</sub>, H, W)]) 

`N` (no. of detections) varies from image to image, so they cannot be batched.

kornia/augmentation/container/ops.py
shijianjian: Ah, this should be fine then.

shijianjian:
> Single sample: Image (C, H, W) and Mask (N, H, W)
> Batched: Image (B, C, H, W) and Mask ([(N1, H, W), (N2, H, W) ... (Nb, H, W)])

I meant that
```python
N_max = max(N1, ..., Nb)
padding = [N_max - N1, ..., N_max - Nb]
padded_masks = (B, N_max, H, W)

... # After augmentation

Unpad (B, N_max, H, W) => Mask ([(N1, H, W), (N2, H, W) ... (Nb, H, W)])
```

The current change is fine in this PR I think. We should benchmark the iterative and batching strategies to see how those methods perform. Maybe a `Mask` class, similar to `Boxes`, to handle these.

shijianjian (approved):

johnnv1 (approved):

PR #2866, opened 2024-04-02
support compile for random gaussian blur

edgarriba:
- support compile for random gaussian blur
- some fixes from https://github.com/kornia/kornia/pull/2865

#### Type of change

- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases
- [ ] 🐞 Bug fix (non-breaking change which fixes an issue)
- [ ] 🔬 New feature (non-breaking change which adds functionality)
- [ ] 🚨 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📝 This change requires a documentation update

#### Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Did you update CHANGELOG in case of a major change?

PR #2868, opened 2024-04-04
Add torch.compile to RandomGaussianIllumination

vgilabert94:
#### Changes
Add torch.compile to RandomGaussianIllumination following #2866.

I have created a function _apply_transform that performs the operation of applying the gaussian illum.
With this we can compile the function if .compiled() is called.

```
rng = torch.manual_seed(1)
input = torch.ones(1, 3, 3, 3) * 0.5
aug = RandomGaussianIllumination(gain=0.5, p=1.)
aug_compile = RandomGaussianIllumination(gain=0.5, p=1.).compile(fullgraph=True)
print((aug(input) == aug_compile(input)).all())
tensor(True)
```

GPU testing is needed, as I don't have a GPU and my gcolab time has run out.

#### Type of change

- [ ] 📚  Documentation Update
- [x] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/tests/augmentation/test_augmentation.py
+++ b/tests/augmentation/test_augmentation.py
@@ -4071,6 +4071,20 @@ class TestRandomGaussianIllumination(BaseTester):
         output_tensor = transform(input_tensor)
         self.assert_close(output_tensor[0], output_tensor[1])
 
+    @pytest.mark.slow
+    @pytest.mark.skipif(
+        torch_version_le(1, 13, 1),
+        reason="torch.compile is not available in previous versions",
+    )
+    def test_compile(self, device, dtype):
+        input_tensor = torch.ones(1, 3, 3, 3, device=device, dtype=dtype) * 0.5
+        aug = RandomGaussianIllumination(gain=0.5, p=1.0)
+        expected = aug(input_tensor)
+        aug = aug.compile(fullgraph=True)
+        actual = aug(input_tensor)
+        assert actual.shape == input_tensor.shape
+        self.assert_close(expected, actual)
+
 
 class TestRandomLinearIllumination(BaseTester):
     def _get_expected(self, device, dtype):
--- a/tests/augmentation/test_augmentation.py
+++ b/tests/augmentation/test_augmentation.py
@@ -4061,18 +4061,12 @@ class TestRandomGaussianIllumination(BaseTester):
         self.assert_close(output_tensor[0], output_tensor[1])
 
     @pytest.mark.slow
-    @pytest.mark.skipif(
-        torch_version_le(1, 13, 1),
-        reason="torch.compile is not available in previous versions",
-    )
-    def test_compile(self, device, dtype):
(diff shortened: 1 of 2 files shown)


edgarriba (approved):
LGTM. try to address the typing

johnnv1 (approved):
LGTM, just need fix the red lights on ci

vgilabert94:
I'm getting errors that not are from this modification:
- TestRandomGaussianBlur.test_compile[cpu-float32]: AttributeError: 'RandomGaussianBlur' object has no attribute 'gaussian_blur2d_fn'
>I think it should be: self._gaussian_blur2d_fn = torch.compile(self._gaussian_blur2d_fn, ...)
- TestLightGlueHardNet.test_smoke: Failed: Timeout >5.0s

tests/augmentation/test_augmentation.py
```diff
@@ -4071,6 +4071,20 @@ def test_same_on_batch(self, device, dtype):
         output_tensor = transform(input_tensor)
         self.assert_close(output_tensor[0], output_tensor[1])
 
+    @pytest.mark.slow
+    @pytest.mark.skipif(
+        torch_version_le(1, 13, 1),
+        reason="torch.compile is not available in previous versions",
+    )
+    def test_compile(self, device, dtype):
```
johnnv1, line 4079: Just to keep the same test name as in other apis, also uses the `torch_optimizer` to skip the necessary cases
```suggestion
    def test_dynamo(self, device, dtype, torch_optimizer):
```

johnnv1:
> I'm getting an error that not are from this modification:

fixes on #2872

tests/augmentation/test_augmentation.py
vgilabert94: torch_optimizer is used for something?

johnnv1: The `torch_optimizer` is the `torch.compile` callable itself (check here: https://github.com/kornia/kornia/blob/22c40fbfbf74160e98c6d1d6c42f90d915d6e7f2/conftest.py#L77) -- inside of this fixture are the checks around the torch compile are available or not -- if you're willing, the best thing would be to move this checker heuristic to a function within `testing/` so we can just use it in the skip marker instead of using the fixture

johnnv1: for cases like these methods compile for the augmentations API, I will want to move to use the `optimizer_backend` fixture which we can inject like the `device` and `dtype` 

right now we are just testing the inductor backend compiler, but in the future we may want to move to another backend

johnnv1: and here is a test which uses the `torch_optimizer` fixture -- https://github.com/kornia/kornia/blob/22c40fbfbf74160e98c6d1d6c42f90d915d6e7f2/tests/filters/test_blur.py#L134

vgilabert94: Understood! Thanks for the explanation!

PR #2882, opened 2024-04-12
Add pre and post processing steps to allow non float dtypes

ashnair1:
#### Changes

When using cross entropy loss with torchmetrics, masks needs to be of type int but kornia only considers float as valid data type.

#### Type of change

#### Checklist

### DIFF
--- a/kornia/augmentation/_2d/mix/base.py
+++ b/kornia/augmentation/_2d/mix/base.py
@@ -45,9 +45,7 @@ class MixAugmentationBaseV2(_BasicAugmentationBase):
 
     def transform_tensor(self, input: Tensor, *, shape: Optional[Tensor] = None, match_channel: bool = True) -> Tensor:
         """Convert any incoming (H, W), (C, H, W) and (B, C, H, W) into (B, C, H, W)."""
-        _validate_input_dtype(
-            input, accepted_dtypes=[torch.float16, torch.float32, torch.float64, torch.int32, torch.int64]
-        )
+        _validate_input_dtype(input, accepted_dtypes=[torch.float16, torch.float32, torch.float64])
 
         if shape is None:
             return _transform_input(input)
--- a/kornia/augmentation/_2d/mix/base.py
+++ b/kornia/augmentation/_2d/mix/base.py
@@ -45,7 +45,9 @@ class MixAugmentationBaseV2(_BasicAugmentationBase):
 
     def transform_tensor(self, input: Tensor, *, shape: Optional[Tensor] = None, match_channel: bool = True) -> Tensor:
         """Convert any incoming (H, W), (C, H, W) and (B, C, H, W) into (B, C, H, W)."""
-        _validate_input_dtype(input, accepted_dtypes=[torch.float16, torch.float32, torch.float64, torch.int32, torch.int64])
+        _validate_input_dtype(
+            input, accepted_dtypes=[torch.float16, torch.float32, torch.float64, torch.int32, torch.int64]
+        )
 
         if shape is None:
             return _transform_input(input)
--- a/kornia/augmentation/_2d/mix/base.py
+++ b/kornia/augmentation/_2d/mix/base.py
@@ -45,7 +45,7 @@ class MixAugmentationBaseV2(_BasicAugmentationBase):
 
     def transform_tensor(self, input: Tensor, *, shape: Optional[Tensor] = None, match_channel: bool = True) -> Tensor:
         """Convert any incoming (H, W), (C, H, W) and (B, C, H, W) into (B, C, H, W)."""
-        _validate_input_dtype(input, accepted_dtypes=[torch.float16, torch.float32, torch.float64])
+        _validate_input_dtype(input, accepted_dtypes=[torch.float16, torch.float32, torch.float64, torch.int32, torch.int64])
 
         if shape is None:
             return _transform_input(input)
--- a/kornia/augmentation/container/augment.py
+++ b/kornia/augmentation/container/augment.py
@@ -23,7 +23,9 @@ __all__ = ["AugmentationSequential"]
 
 _BOXES_OPTIONS = {DataKey.BBOX, DataKey.BBOX_XYXY, DataKey.BBOX_XYWH}
 _KEYPOINTS_OPTIONS = {DataKey.KEYPOINTS}
-_IMG_MSK_OPTIONS = {DataKey.INPUT, DataKey.MASK}
+_IMG_OPTIONS = {DataKey.INPUT, DataKey.IMAGE}
+_MSK_OPTIONS = {DataKey.MASK}
+_CLS_OPTIONS = {DataKey.CLASS, DataKey.LABEL}
 
 
 class AugmentationSequential(TransformMatrixMinIn, ImageSequential):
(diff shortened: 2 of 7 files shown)


kornia/augmentation/_2d/mix/base.py
```diff
@@ -45,7 +45,9 @@ def __init__(
 
     def transform_tensor(self, input: Tensor, *, shape: Optional[Tensor] = None, match_channel: bool = True) -> Tensor:
         """Convert any incoming (H, W), (C, H, W) and (B, C, H, W) into (B, C, H, W)."""
-        _validate_input_dtype(input, accepted_dtypes=[torch.float16, torch.float32, torch.float64, torch.int32, torch.int64])
+        _validate_input_dtype(
+            input, accepted_dtypes=[torch.float16, torch.float32, torch.float64, torch.int32, torch.int64]
```
johnnv1, line 49: maybe we should consider uint too?

johnnv1 (commented):
add some tests to ensure too... since the random generators use float values, maybe the output will be float anyway... I'm not sure, but that's what I'm waiting for xD

shijianjian (changes requested):
I do not think it is reasonable, as most methods considers only 0-1 value ranges. If you want to support other formats like uint8 or uint16. A preprocessing and postprocessing to 0-1 are in need I think.

ashnair1:
> I do not think it is reasonable, as most methods considers only 0-1 value ranges. If you want to support other formats like uint8 or uint16. A preprocessing and postprocessing to 0-1 are in need I think.

Agreed. Have made it so inputs are converted to float before the operations and reverted to original dtype once they're complete.

ashnair1:
@shijianjian @johnnv1 @edgarriba Any thoughts on this?

tests/augmentation/test_container.py
```diff
@@ -555,13 +555,17 @@ def test_masks_without_channel_dim(self, device, dtype, B, C_i, C_m, keepdim):
...
-            out_mask_shape = tuple(x if x else 1 for x in (B, C_m, *img_shape[-2:]))
+            out_mask_shape = tuple(x or 1 for x in (B, C_m, *img_shape[-2:]))
             assert out[1].shape == out_mask_shape
 
     @pytest.mark.slow
     @pytest.mark.parametrize("random_apply", [1, (2, 2), (1, 2), (2,), 10, True, False])
+    @pytest.mark.parametrize("dtype", [torch.int32, torch.int64, torch.float32])
     def test_forward_and_inverse(self, random_apply, device, dtype):
-        inp = torch.randn(1, 3, 1000, 500, device=device, dtype=dtype)
+        if dtype not in [torch.float32, torch.float64]:
```
shijianjian, line 565: We do not run int test in our CI. You may put only mask for int.

kornia/augmentation/container/ops.py
```diff
@@ -182,6 +183,11 @@ def get_geometric_only_param(module: "K.container.ImageSequentialBase", param: L
 class InputSequentialOps(SequentialOpsInterface[Tensor]):
     @classmethod
     def transform(cls, input: Tensor, module: Module, param: ParamItem, extra_args: Dict[str, Any] = {}) -> Tensor:
+        dtype = None
+        if not torch.is_floating_point(input):
+            dtype = input.dtype
+            input = input.float()
```
shijianjian, line 189: This does not make sense for input. Our function does not support 0-255 range.

shijianjian (changes requested):
I think this should not apply on input type.

ashnair1:
Current issue is the type of dtype to cast to. If the a non float dtype mask is detected, it is converted into a float type and later reverted back. The issue is that if it's converted to float, the slow float64 tests complain (Expected float, got double) and when it's converted to double, the slow float32 tests complain.

Would appreciate some pointers here as I'm not sure how to go about this.

johnnv1:
> Would appreciate some pointers here as I'm not sure how to go about this.

hey @ashnair1, sorry for the delay on this one...

i think the issue here is: the random generator creates the augmentations parameters in XXX device and YYY dtype then casts it into the input dtype (probably the same dtype as the image, since we always have the image)... here you are "hardcoding" the dtype for the other cases when you want to support nonfloating dtypes, instead you should cast it to the input dtype (or maybe the params dtype? not sure cc @shijianjian)

ashnair1:
That does not make sense to me. Just so we're all on the same page regarding what we're trying to achieve here.

In a kornia augmentation, masks by default have to be float. This is because certain operations don't work for int tensors. But as mentioned above, masks are typically expected to be int tensors and the fact they are float here is a design convenience.

So in order to work with masks tensors of type int, we should pre-process it to float before the op (that only accepts floats) and post process it back to int after the op. 

However in the tests, it requires the mask tensors to be the same dtype as input. This means that if you want to test passing in a mask tensor of type int, you would (as you suggested) need to have input be of type int as well. Which in turn means we need to add a pre and post processing step for input as well which @shijianjian had advised against (https://github.com/kornia/kornia/pull/2882#pullrequestreview-2034701027) and which I agree with. 

Ultimately, what I would like to see, is that kornia should ensure input is float while allowing masks to be int or float relying on the processing steps to ensure the mask tensor works with the ops.

johnnv1:
> That does not make sense to me. Just so we're all on the same page regarding what we're trying to achieve here.
> 
> In a kornia augmentation, masks by default have to be float. This is because certain operations don't work for int tensors. But as mentioned above, masks are typically expected to be int tensors and the fact they are float here is a design convenience.
> 
> So in order to work with masks tensors of type int, we should pre-process it to float before the op (that only accepts floats) and post process it back to int after the op.

yeap, but we need to cast to the _right_ floating type, which is the same floating type as the input data, otherwise, it will mismatch and crash in some ops... 

what I mean is to retrieve the data type from the input itself, like using something like https://github.com/kornia/kornia/blob/25237be70fd27e5df6f71b7be553910aee8584c0/kornia/augmentation/container/augment.py#L408 (or some other from the API itself)

johnnv1 (commented):
LGTM, great patch, to me we can merge it when the typing CI success

kornia/augmentation/container/augment.py
```diff
@@ -332,13 +341,23 @@ def _validate_args_datakeys(self, *args: DataType, data_keys: List[DataKey]) ->
     def _arguments_preproc(self, *args: DataType, data_keys: List[DataKey]) -> List[DataType]:
         inp: List[DataType] = []
         for arg, dcate in zip(args, data_keys):
-            if DataKey.get(dcate) in _IMG_MSK_OPTIONS:
+            if DataKey.get(dcate) in _IMG_OPTIONS:
+                arg = cast(Tensor, arg)
+                self.input_dtype = arg.dtype
```
shijianjian, line 346: I think we should use another for-loop to find the `input_dtype` first. Otherwise, it may happen if the `INPUT` type comes after `MASK`.

shijianjian (approved):

kornia/augmentation/container/augment.py
ashnair1: Doesn't this if block make sure that does not happen?  It ensures INPUT DataKey must be present and will always be the first DataKey to be evaluated.
 https://github.com/kornia/kornia/blob/e461f92ff9ee035d2de2513859bee4069356bc25/kornia/augmentation/container/augment.py#L240

Issue #2877, opened 2024-04-08
Update apply_colormap 


vgilabert94:
## 🚀 Feature -> Update apply_colormap 
From https://github.com/kornia/kornia/pull/2794#issuecomment-1955189012

Discussion about ColorMaps -> apply_colormap function. (https://github.com/kornia/kornia/blob/8df6c44ce6ba233c07185ac46107c6d9a076d3c0/kornia/color/colormap.py#L161)

## Motivation

- Right now the input tensor should be in uint8 (0-255) which not follow the float32 standard of kornia (0-1).
- Only grayscale images are accepted, we also can update to any channels. If input is with channel 3, output will be 9 (3 per every input channel).
- Try to accept batched tensors (batch >= 1). 

## Pitch
- Use image in float32 in range (0,1).
- Accept images with no restriction channels.

edgarriba:
how are you planning to implement the LUT with f32 ?

vgilabert94:
I've create two "premilinar" versions that allow to use batch > 1 and channels > 1 an range in (0-1):

```
def apply_colormap_v1(input_tensor, cmap):
    B, C, H, W = input_tensor.shape
    colormap = cmap.colors.permute(1, 0)
    num_colors, channels_cmap = colormap.shape
    input_tensor = input_tensor.reshape(B, C, -1)

    keys = torch.linspace(0.0, 1.0, num_colors-1, device=input_tensor.device)
    index = torch.bucketize(input_tensor, keys)
    output = colormap[index]
    # (B, C, H*W, channels_cmap) -> (B, C*channels_cmap, H, W)
    output = output.permute(0, 1, 3, 2).reshape(B, C * channels_cmap, H, W)

    return output


def apply_colormap_v2(input_tensor, cmap):
    B, C, H, W = input_tensor.shape
    cmap = cmap.colors.permute(1, 0)
    num_colors, channels_cmap = cmap.shape
    input_tensor = input_tensor.reshape(B, C, -1)

    keys = torch.linspace(0.0, 1.0, num_colors-1, device=input_tensor.device)
    index = torch.bucketize(input_tensor, keys).unsqueeze(-1).expand(-1, -1, -1, 3)
    output = torch.gather(cmap.expand(B, C, -1, -1), 2, index)
    # (B, C, H*W, channels_cmap) -> (B, C*channels_cmap, H, W)
    output = output.permute(0, 1, 3, 2).reshape(B, C*channels_cmap, H, W)

    return output
```

[...]

PR #2886, opened 2024-04-16
feat: support batched and float data in apply colormap

vgilabert94:
#### Changes
fixes [#2877](https://github.com/kornia/kornia/issues/2877)

- Use image in float32 in range [0,1]. 
--  Also is posible to use int or uint in range [0-255] values with an internal normalization to [0-1].
- Accept images with no restriction channels.
- Accept batched tensors.

Output:
- Batched tensor (B,C,H,W). Float with range [0-1]. (My suggestion, we could change it or add more flexibility if needed.)

Always the apply_colormap work in float dtype, because the problem that I detect here (https://github.com/kornia/kornia/issues/2877#issuecomment-2047702490) is when arange is used. I'm not sure why, but I tried to implement my function with arange and I kept getting this problem. 

#### Type of change

- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/kornia/color/colormap.py
+++ b/kornia/color/colormap.py
@@ -207,8 +207,8 @@ def apply_colormap(input_tensor: Tensor, colormap: ColorMap) -> Tensor:
     max_value = 1.0 if input_tensor.max() <= 1.0 else 255.0
     input_tensor = input_tensor.float().div_(max_value)
 
-    colormap = colormap.colors.permute(1, 0)
-    num_colors, channels_cmap = colormap.shape
+    colors = colormap.colors.permute(1, 0)
+    num_colors, channels_cmap = colors.shape
     keys = torch.linspace(0.0, 1.0, num_colors - 1, device=input_tensor.device, dtype=input_tensor.dtype)
     indices = torch.bucketize(input_tensor, keys).unsqueeze(-1).expand(-1, -1, -1, 3)
 
--- a/kornia/color/colormap.py
+++ b/kornia/color/colormap.py
@@ -212,7 +212,7 @@ def apply_colormap(input_tensor: Tensor, colormap: ColorMap) -> Tensor:
     keys = torch.linspace(0.0, 1.0, num_colors - 1, device=input_tensor.device, dtype=input_tensor.dtype)
     indices = torch.bucketize(input_tensor, keys).unsqueeze(-1).expand(-1, -1, -1, 3)
 
-    output = torch.gather(colormap.expand(B, C, -1, -1), 2, indices)
+    output = torch.gather(colors.expand(B, C, -1, -1), 2, indices)
     # (B, C, H*W, channels_cmap) -> (B, C*channels_cmap, H, W)
     output = output.permute(0, 1, 3, 2).reshape(B, C * channels_cmap, H, W)
 
--- a/kornia/color/colormap.py
+++ b/kornia/color/colormap.py
@@ -9,7 +9,7 @@ from torch.nn.functional import interpolate
 import kornia.color._colormap_data as cm_data
 from kornia.color._colormap_data import RGBColor
 from kornia.core import Module, Tensor, tensor
-from kornia.core.check import KORNIA_CHECK_IS_GRAY
+from kornia.core.check import KORNIA_CHECK
 from kornia.utils.helpers import deprecated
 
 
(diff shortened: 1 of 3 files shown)


kornia/color/colormap.py
```diff
@@ -174,41 +174,49 @@ def apply_colormap(input_tensor: Tensor, colormap: ColorMap) -> Tensor:
...
 
-    KORNIA_CHECK_IS_GRAY(input_tensor)
+    KORNIA_CHECK(isinstance(input_tensor, Tensor), f"`input_tensor` must be a Tensor. Got: {type(input_tensor)}")
+    valid_types = [torch.half, torch.float, torch.double, torch.uint8, torch.int, torch.long, torch.short]
+    KORNIA_CHECK(
+        input_tensor.dtype in valid_types, f"`input_tensor` must be a {valid_types}. Got: {input_tensor.dtype}"
+    )
+    KORNIA_CHECK(len(input_tensor.shape) in (3, 4), "Wrong input tensor dimension.")
+    if len(input_tensor.shape) == 3:
+        input_tensor = input_tensor.unsqueeze_(0)
```
johnnv1, line 203: shouldn't we support only HxW tensors too? maybe through https://github.com/kornia/kornia/blob/b0995bdce3b04a11d39e86853bb1de9a2a438ca2/kornia/utils/image.py#L214?

vgilabert94: I'm having problems with this decorator, because the output is converted to the same size as input and for colormap operation it is not correct. output shape should be (B,C*3,H,W).

johnnv1: ouch, i see, so this decorator will not work

johnnv1:
@vgilabert94 could you check/fix the typing issue?

vgilabert94:
> @vgilabert94 could you check/fix the typing issue?

Yes. I have doubts about how to solve it, because is a typing error with ColorMap class. This class generates a tensor in colormap.colors, do I have to inherit from a torch class?

```
kornia/color/colormap.py:210: error: Incompatible types in assignment
(expression has type "Tensor", variable has type "ColorMap")  [assignment]
        colormap = colormap.colors.permute(1, 0)
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
kornia/color/colormap.py:211: error: "ColorMap" has no attribute "shape" 
[attr-defined]
        num_colors, channels_cmap = colormap.shape
                                    ^~~~~~~~~~~~~~
kornia/color/colormap.py:2[15](https://github.com/kornia/kornia/actions/runs/8702817467/job/23867709917?pr=2886#step:3:16): error: "ColorMap" has no attribute "expand" 
[attr-defined]
        output = torch.gather(colormap.expand(B, C, -1, -1), 2, indices)
                              ^~~~~~~~~~~~~~~
Found 3 errors in 1 file (checked 436 source files)
Error: Process completed with exit code 1.
```

kornia/color/colormap.py
```diff
@@ -174,41 +174,49 @@ def apply_colormap(input_tensor: Tensor, colormap: ColorMap) -> Tensor:
...
-    elif len(input_tensor.shape) == 3 and input_tensor.shape[0] == 1:  # if (1 X H x W)
-        input_tensor = input_tensor[0, ...]  # (H x W)
+    B, C, H, W = input_tensor.shape
+    input_tensor = input_tensor.reshape(B, C, -1)
+    max_value = 1.0 if input_tensor.max() <= 1.0 else 255.0
+    input_tensor = input_tensor.float().div_(max_value)
 
-    keys = torch.arange(0, len(colormap) - 1, dtype=input_tensor.dtype, device=input_tensor.device)  # (num_colors)
+    colormap = colormap.colors.permute(1, 0)
+    num_colors, channels_cmap = colormap.shape
```
johnnv1, line 211: The typing issue is because you are overwriting the colormap within a tensor here... just rename this variable instead of overwrite 
```suggestion
    colors = colormap.colors.permute(1, 0)
    num_colors, channels_cmap = colors.shape
```
**attention**: needs to update it in some places below too

kornia/color/colormap.py
```diff
@@ -174,41 +174,49 @@ def apply_colormap(input_tensor: Tensor, colormap: ColorMap) -> Tensor:
...
+    keys = torch.linspace(0.0, 1.0, num_colors - 1, device=input_tensor.device, dtype=input_tensor.dtype)
+    indices = torch.bucketize(input_tensor, keys).unsqueeze(-1).expand(-1, -1, -1, 3)
 
-    index = torch.bucketize(input_tensor, keys)  # shape equals <input_tensor>: (B x H x W) or (H x W)
+    output = torch.gather(colormap.expand(B, C, -1, -1), 2, indices)
```
johnnv1, line 215:
```suggestion
    output = torch.gather(colors.expand(B, C, -1, -1), 2, indices)
```

PR #2914, opened 2024-05-19
Speed up differentiable 5PC and fix the batch size issue

### LINKED ISSUE #2904: find_essential and run_5point are not working correctly with batch size>1
### Describe the bug

The problem is with run_5point solver, which has several filters:
https://github.com/kornia/kornia/blob/a041d9255f459a0034c75b20fb9ab524d7ae3983/kornia/geometry/epipolar/essential.py#L124
https://github.com/kornia/kornia/blob/a041d9255f459a0034c75b20fb9ab524d7ae3983/kornia/geometry/epipolar/essential.py#L161
https://github.com/kornia/kornia/blob/a041d9255f459a0034c75b20fb9ab524d7ae3983/kornia/geometry/epipolar/essential.py#L164
All of which can skip processing a batch index without appending anything to the output list E_models. The correspondence between index in this list and the index in the batch is lost.
In the result the shape of the returned tensor will be less than 10*B and it is not possible to determine for which batch indices there were no solution.

The solver is used in find_essential.

### Reproduction steps

```bash
Run run_5point with 
points1: (B, 5, 2)
points2: (B, 5, 2)
with a large B, e.g. B=100.

Although, the doc string says N>=8 points. Is that necessary?
```

### Expected behavior

Expecting output of shape 
(B*10, 3,3)
Although the doctoring says "computed essential matrix with shape :math:`(B, 3, 3)`.", the code returns 10 solutions per batch index when not skipping.

### Environment

```shell
kornia version 0.7.2
```

### Additional context

_No response_

weitong8591:
#### Changes
- remove one 'for' loop to speed up the solver, referred to [vggsfm](https://github.com/facebookresearch/vggsfm/blob/f490bc11b3eb68c1aa832fcd2e230ffacbf575fd/vggsfm/two_view_geo/essential.py#L325).
- make sure the batch size of the returned E matrices the same as the input by initializing the E matrices with identity matrices. Only those with valid 'singular_filter' are replaced by the computed essential matrices. @jytime
- add large batch sizes of data in the tests

Fixes #2904

#### Type of change

- [x] 📚  Documentation Update
- [x] 🧪 Tests Cases

#### Checklist


johnnv1 (approved):

edgarriba (commented):
is there any way we could add a specific unit test to raises this corner case ?

weitong8591:
> is there any way we could add a specific unit test to raises this corner case ?

Hi, that's a good idea, I found one bug when checking the corner case of degenerate points.  I am adding this unit test. Also, to be consistent with 7PC to return all 10 possible solutions, I am removing the code of selecting one out of ten solutions by Sampson distances. will commit soon

tests/geometry/epipolar/test_essential.py
```diff
@@ -68,9 +69,20 @@ def test_synthetic_sampson(self, device, dtype):
...
         self.assert_close(
-            error, torch.zeros((calibrated_x1.shape[:2]), device=device, dtype=dtype), atol=1e-4, rtol=1e-4
+            error[:, torch.argmin(error.mean(-1).min())],
+            torch.zeros((calibrated_x1.shape[:2]), device=device, dtype=dtype),
+            atol=1e-4,
+            rtol=1e-4,
         )
 
+    @pytest.mark.parametrize("batch_size, num_points", [(5, 5), (10, 5)])
+    def test_degenarate_case(self, batch_size, num_points, device, dtype):
```
edgarriba, line 79:
```suggestion
    def test_degenerate_case(self, batch_size, num_points, device, dtype):
```

edgarriba (approved):

weitong8591:
Thanks all for reviewing it!

PR #2921, opened 2024-06-03
fix: rgb_to_grayscale on uint8 images

JojiJoseph:
#### Changes
Summary:
There was an overflow while converting uint8 RGB images to grayscale. Therefore if we try to convert a uint8 image, we will get a noisy image. Fixed it. Also changed the range notation from open interval to closed interval.

Motivation and context:
I tried to change an rgb image with values in range [0, 255]. I got a noisy image.

Dependencies that are required for this change:
None

#### Type of change

#### Checklist

- [ ] have made corresponding changes to the documentation

### DIFF
--- a/kornia/color/gray.py
+++ b/kornia/color/gray.py
@@ -39,7 +39,7 @@ def rgb_to_grayscale(image: Tensor, rgb_weights: Optional[Tensor] = None) -> Ten
 
     .. image:: _static/img/rgb_to_grayscale.png
 
-    The image data is assumed to be in the range of (0, 1).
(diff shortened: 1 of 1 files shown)


edgarriba:
@JojiJoseph this will break backcward compatibility. Any other solution ? /cc @johnnv1

JojiJoseph:
Got it! If backward compatibility is a concern, we can close this fix. Users can still handle the grayscale conversion of uint8 images with extra two lines of code on their end. Alternatively, we could add an optional parameter to handle the conversion without breaking existing functionality.

edgarriba:
i think users cangive any weights at this moment right ?

PR #2930, opened 2024-06-15
chore (test suite): skip or xfail random failures

johnnv1:
This should get green lights (or at least close to it) on full test suite (scheduled/push job)

### DIFF
--- a/tests/contrib/models/test_tiny_vit.py
+++ b/tests/contrib/models/test_tiny_vit.py
@@ -1,3 +1,5 @@
+import sys
+
 import pytest
 import torch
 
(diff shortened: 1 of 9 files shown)


tests/contrib/models/test_tiny_vit.py
```diff
@@ -35,6 +37,7 @@ def test_gradcheck(self): ...
     @pytest.mark.skip("not implemented")
     def test_module(self): ...
 
+    @pytest.mark.skipif(sys.version_info.major == 3 and sys.version_info.minor == 8, reason="not working for py3.8")
```
edgarriba, line 40: Similar to torch_version utils, we could have python_version

tests/geometry/epipolar/test_fundamental.py
```diff
@@ -247,6 +247,7 @@ def test_synthetic_sampson_7point(self, device, dtype):
                 error = epi.sampson_epipolar_distance(x1, x2, F)
                 self.assert_close(error, torch.zeros((F.shape[0], 7), device=device, dtype=dtype), atol=1e-4, rtol=1e-4)
 
+    @pytest.mark.xfail()
```
edgarriba, line 250: Add a todo or open a ticket to implement a deterministic two view scene

edgarriba (approved):

tests/contrib/models/test_tiny_vit.py
johnnv1: i think we can use [the version_info directly](https://docs.python.org/3/library/sys.html#sys.version_info), this one we can rewrite like
```suggestion
    @pytest.mark.skipif(sys.version_info[:2] == (3, 8), reason="not working for py3.8")
```

PR #2932, opened 2024-06-20
fix (CI): ensure support test suite w/ numpy

johnnv1:
Ensure support to use numpy>2.0.0 on newer torch versions (torch>=2.3) for older torchs versions it'll use numpy<2.0.0.

Without it, the CI will be failing, because the older torch versions aren't compiled within the new numpy API.


edgarriba (approved):

johnnv1:
The tutorials need to be fixed in the tutorials repo, I need to review why it is falling for the 3.12

PR #2940, opened 2024-06-29
Include Steerers

### LINKED ISSUE #2913: Including Steerers
## 🚀 Feature
It would be simple to add support for [Steerers](https://github.com/georg-bn/rotation-steerers), since the descriptor is already implemented with DeDoDe.

The only thing that would have to be added are the steerers themselves [see here](https://github.com/georg-bn/rotation-steerers/blob/main/rotation_steerers/steerers.py), perhaps in the DeDoDe folder. And new matchers [(e.g. something like this)](https://github.com/georg-bn/rotation-steerers/blob/main/rotation_steerers/matchers/max_similarity.py), I suppose either in kornia/feature/matching.py or a new file.

I'll create a pull request if this is of interest.

georg-bn:
#### Changes
This implements Steerers as described in our recent CVPR paper ([arxiv](https://arxiv.org/abs/2312.02152)).
I also have implemented a [tutorial](https://github.com/georg-bn/kornia-tutorials/blob/master/nbs/image_matching_steerers.ipynb) which can be pulled into kornia-tutorials after Steerers have been merged in here.

Fixes #2913.

#### Type of change

#### Checklist

### DIFF
--- a/kornia/feature/matching.py
+++ b/kornia/feature/matching.py
@@ -340,7 +340,7 @@ class DescriptorMatcherWithSteerer(Module):
         steerer_order: order of discretisation of rotation angles, e.g. 4 leads to quarter rotations.
         steer_mode: can be `global`, `local`.
             `global` means that the we output matches from the global rotation with most matches.
-            `local` means that we output matches from a distance matrix 
+            `local` means that we output matches from a distance matrix
             where the distance between each descriptor pair is the minimal over rotations.
         match_mode: type of matching, can be `nn`, `snn`, `mnn`, `smnn`.
             WARNING: using steer_mode `global` with match_mode `nn` will lead to bad results
@@ -388,15 +388,19 @@ class DescriptorMatcherWithSteerer(Module):
         """
         rot1to2 = None
         if self.match_mode == "nn":
+
             def matching_function(d1, d2, dm):
                 return match_nn(d1, d2, dm=dm)
         elif self.match_mode == "mnn":
+
             def matching_function(d1, d2, dm):
                 return match_mnn(d1, d2, dm=dm)
         elif self.match_mode == "snn":
+
             def matching_function(d1, d2, dm):
                 return match_snn(d1, d2, self.th, dm=dm)
         elif self.match_mode == "smnn":
+
             def matching_function(d1, d2, dm):
                 return match_smnn(d1, d2, self.th, dm=dm)
         else:
--- a/kornia/feature/matching.py
+++ b/kornia/feature/matching.py
@@ -1,6 +1,5 @@
 from typing import Any, ClassVar, Dict, List, Optional, Tuple
 
-import numpy as np
 import torch
 
 from kornia.core import Module, Tensor, concatenate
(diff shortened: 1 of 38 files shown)


kornia/feature/steerers.py
```diff
@@ -0,0 +1,66 @@
+from typing import Dict
+
+import torch
+
+from kornia.core import Module, Tensor
+from kornia.utils.helpers import map_location_to_cpu
+
+dedode_steerer_urls: Dict[str, str] = {
+    "B-C4": "https://github.com/georg-bn/rotation-steerers/releases/download/release-2/B_C4_Perm_steerer_setting_C.pth",
```
johnnv1, line 9: maybe we should push to kornia hugging face hub these weights? cc @edgarriba

kornia/feature/steerers.py
```diff
@@ -0,0 +1,66 @@
...
+    "B-SO2": "https://github.com/georg-bn/rotation-steerers/releases/download/release-2/B_SO2_Spread_steerer_setting_C.pth",
+    "G-C4": "https://github.com/georg-bn/rotation-steerers/releases/download/release-2/G_C4_Perm_steerer_setting_C.pth",
+    "G-SO2": "https://github.com/georg-bn/rotation-steerers/releases/download/release-2/G_SO2_Spread_steerer_setting_C.pth",
+}
+
+
+class DiscreteSteerer(Module):
+    def __init__(self, generator: Tensor) -> None:
```
johnnv1, line 17: docstring (with example if possible) to how to init/use this class

georg-bn: Thanks, fixed

kornia/feature/steerers.py
georg-bn: I realized, that for the steerers included currently, we can just define them as they are not trained. So I refactored the code like that instead and removed this dict.

kornia/feature/steerers.py
georg-bn: I added a more complete example to matcher here:

https://github.com/georg-bn/kornia/blob/b9975ffb74c208865fbc139ad5f619031a34b63e/kornia/feature/matching.py#L351

johnnv1 (approved):
Great patch, thanks for the feat. Final review for @ducha-aiki

georg-bn:
I will add another feature, namely faster matching with a subset of the keypoints for determining the number of rotations required.

kornia/feature/matching.py
```diff
@@ -409,12 +409,23 @@ def matching_function(self, d1: Tensor, d2: Tensor, dm: Optional[Tensor] = None)
...
+    ) -> Tuple[Tensor, Tensor, Optional[int]]:
         """
         Args:
             desc1: Batch of descriptors of a shape :math:`(B1, D)`.
             desc2: Batch of descriptors of a shape :math:`(B2, D)`.
             normalize: bool to decide whether to normalize descriptors to unit norm.
+            fast: bool to decide whether to determine optimal number of rotations
+                using only a subset of the descriptions.
+                This is only used if `self.steer_mode` is `global`.
+            subset_size: The subset size to use if `fast` is True.
```
johnnv1, line 428: instead of using a fast bool flag, to be simpler, what do you think of using if subset_size is None, then fast=False, otherwise use fast mode.

georg-bn: Good idea, fixed.

edgarriba (approved):

PR #2883, opened 2024-04-13
Fix: `RandomJPEG` bug with different devices

ditwoo:
#### Changes

Fixes #2867

#### Type of change

- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases
- [+] 🐞 Bug fix (non-breaking change which fixes an issue)
- [ ] 🔬 New feature (non-breaking change which adds functionality)
- [ ] 🚨 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📝 This change requires a documentation update

#### Checklist

- [+] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Did you update CHANGELOG in case of a major change?

PR #2954, opened 2024-07-15
Fix device bug in JPEG augmentation

ChristophReich1996:
#### Changes
When running random JPEG augmentation on CUDA, some tensors are not correctly moved to CUDA. To reproduce the bug just run the following code:

```python
import kornia
import torch
aug = kornia.augmentation.RandomJPEG(p=1.0)
aug.cuda()
aug(torch.rand(2, 3, 256, 256, device="cuda"))
# quantization_table[:, None] * _jpeg_quality_to_scale(jpeg_quality)[:, None, None, None]
# RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

This pull request fixes this bug by just moving the JPEG quality to the same device as the input.

#### Type of change

- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/kornia/augmentation/_2d/intensity/jpeg.py
+++ b/kornia/augmentation/_2d/intensity/jpeg.py
@@ -53,5 +53,5 @@ class RandomJPEG(IntensityAugmentationBase2D):
     def apply_transform(
         self, input: Tensor, params: Dict[str, Tensor], flags: Dict[str, Any], transform: Optional[Tensor] = None
     ) -> Tensor:
-        jpeg_output: Tensor = jpeg_codec_differentiable(input, params["jpeg_quality"])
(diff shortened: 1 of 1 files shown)


johnnv1:
hey, could you check if https://github.com/kornia/kornia/pull/2883#pullrequestreview-1999736563 solves it, or is indeed necessary this or the #2883 fix?

ChristophReich1996:
Hi @johnnv1, thanks for pointing me to the other pull request! So, both pull requests fix the issue. Your fix by calling `set_rng_device_and_dtype` also poses a solution. However, I guess for other augmentations, calling `set_rng_device_and_dtype ` is not needed, so maybe fixing this by using one of the pull requests might be better.

PR #2961, opened 2024-07-22
[Feat] Added dissolving transformation & updated docs

shijianjian:
#### Changes

Fixes # (issue)

#### Type of change

- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases

#### Checklist


shijianjian:
@johnnv1 I think we may interact more with external libs like diffusers, transformers, etc. Should we add them to our CI for testing? Or any other resolutions?

johnnv1:
> @johnnv1 I think we may interact more with external libs like diffusers, transformers, etc. Should we add them to our CI for testing? Or any other resolutions?

if we go with this external module, i think we can have a specific CI job for testing all cases which depends on these external library... like we have for tutorials, but using or a pytest mark or putting it under some specific module or name on the test suite to select it

shijianjian:
One additional question, like for diffusers, if we need to download the weights for a specific version of diffusion models. It may heavily reduce our CI usability, right? In terms of downloading time, spaces, etc. I am wondering if what is the best approach to test those things.

johnnv1:
> One additional question, like for diffusers, if we need to download the weights for a specific version of diffusion models. It may heavily reduce our CI usability, right? In terms of downloading time, spaces, etc. I am wondering if what is the best approach to test those things.

yeap, you are right. But right now we have already been doing it. We have a bunch of tests that need to download the weights to run. As a palliative measure, I have added a `slow` marker into these tests, so we don't run them against all versions we support.

Ideally, we should have a pre-test job, that downloads all these weights beforehand, and then makes them available before the tests start. This way we can "share" the weights between the actions jobs, and they will not download it when the suite runs

docs/source/get-started/highlights.rst
```diff
@@ -1,6 +1,20 @@
...
 
+At Kornia, we are dedicated to pushing the boundaries of computer vision by providing a robust, efficient, and versatile toolkit. Our library is built on the powerful PyTorch backend, leveraging its efficiency and auto-differentiation capabilities to deliver high-performance solutions for a wide range of vision tasks.
+
+
+Accessible AI Models
+--------------------
+
+We are excited to announce our latest advancement: a new initiative designed to seamlessly integrate lightweight AI models into Kornia. This addition is crafted to empower developers, researchers, and enthusiasts to harness the full potential of accessible AI, simplifying complex vision tasks and accelerating innovation.
+
+We have curated a selection of lightweight AI models, including YuNet, Loftr, and SAM, optimized for performance and efficiency. These models offer efficient computations that do not require expensive GPUs, making cutting-edge AI accessible to everyone. We welcome the whole community of developers and researchers who are passionate about advancing computer vision, throwing PRs for your lightning fast models!
```
edgarriba, line 12: we might make a call to help curating / updateing more models (SAM2) and to export to onnx

kornia/core/external.py
```diff
@@ -0,0 +1,28 @@
+import importlib
+
+
+class LazyLoader:
```
edgarriba, line 4: add some docs anf typing (if makes sense)

kornia/core/external.py
```diff
@@ -0,0 +1,28 @@
...
+    def __getattr__(self, item):
+        self._load()
+        return getattr(self.module, item)
+
+    def __dir__(self):
+        self._load()
+        return dir(self.module)
+
+
+diffusers = LazyLoader("diffusers")
```
edgarriba, line 28: is it needed to register this lib ? or can be done directly where is reallty used, e.g in StableDiffusionDissolving

kornia/filters/dissolving.py
```diff
@@ -0,0 +1,139 @@
...
+          - figure:: https://raw.githubusercontent.com/kornia/data/main/dslv-sd-2.1.png
+
+    Args:
+        version: the version of the stable diffusion model.
+        **kwargs: additional arguments for `.from_pretrained`.
+    """
+
+    def __init__(self, version: str = "2.1", **kwargs):
+        super().__init__()
+        from kornia.core.external import diffusers
```
edgarriba, line 106: i would create here directly the lazy loader

edgarriba (approved):
l like the LazyLoader feature

kornia/core/external.py
shijianjian: It is better to register here I think. Otherwise, we may have multiple singleton class towards the same lib.

shijianjian:
@johnnv1 Do you have any ideas about how to setup a shared pre-download space? So that I can try adding some tests.

kornia/filters/dissolving.py
```diff
@@ -0,0 +1,138 @@
...
+            - torch.sqrt(1.0 / self.scheduler.alphas_cumprod[timestep] - 1) * noise_pred
+        )
+
+    @torch.no_grad()
+    def init_prompt(self, prompt: str) -> None:
+        uncond_input = self.model.tokenizer(
+            [""], padding="max_length", max_length=self.model.tokenizer.model_max_length, return_tensors="pt"
+        )
+        uncond_embeddings = self.model.text_encoder(uncond_input.input_ids.to(self.model.device))[0]
+        text_input = self.model.tokenizer(
```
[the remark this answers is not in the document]
shijianjian: We can. But the problem is that it is too slow. Unless we have some very fast methods. Otherwise, it is not clever to run it in the runtime

docs/source/get-started/highlights.rst
shijianjian: Probably update the README file to call for model contributors?

johnnv1:
> @johnnv1 Do you have any ideas about how to setup a shared pre-download space? So that I can try adding some tests.

something like

Create a GHA job that will do some torch.load for each weight (need to find out all weights hardcoded) to a specific directory, then upload it to a gha cache (https://github.com/actions/cache), then add download weights step (into the .cache directory) on each test job, then execute the test suite

README.md
```diff
@@ -100,16 +141,20 @@ augmentation_pipeline = AugmentationSequential(
     RandomAffine((-45., 45.), p=1.),
     RandomBrightness((0.,1.), p=1.)
 )
+# Leveraging StableDiffusion models
```
edgarriba, line 130: can we show above loading images with kornia_rs ?

also I'm not very fond of promoting augmentations as everyone thinks we are only an augmentations library in the end

README.md
```diff
@@ -100,16 +141,20 @@ augmentation_pipeline = AugmentationSequential(
...
 img = augmentation_pipeline(img)
+dslv_op(img, step_number=500)
 
-# Leveraging StableDiffusion models
-output = StableDiffusionDissolving()(img, step_number=500)
-output.save("Kornia-enhanced.jpg")
+dslv_op.save("Kornia-enhanced.jpg")
 ```
 
+<<<<<<< HEAD
```
edgarriba: bad diff

README.md
shijianjian: Can you modify it directly there? To add more examples or anything else. What I wanted to say here is that we support many different input types and the easy save and visualization.

I think this Quick Start is to promote the use cases in practic. We can include `kornia_rs` as a torch tensor example maybe.

README.md
```diff
@@ -24,11 +24,53 @@ English | [简体中文](README_zh-CN.md)
 </p>
 </div>
 
-**Kornia** is a differentiable computer vision library for [PyTorch](https://pytorch.org).
+**Kornia** is a differentiable computer vision library that provides a rich set of differentiable image processing and geometric vision algorithms. Built on top of [PyTorch](https://pytorch.org), Kornia integrates seamlessly into existing AI workflows, allowing you to leverage powerful [batch transformations](), [auto-differentiation]() and [GPU acceleration](). Whether you’re working on image transformations, augmentations, or AI-driven image processing, Kornia equips you with the tools you need to bring your ideas to life.
```
edgarriba, line 27: here we are missing some links

README.md
edgarriba: added

shijianjian: We may add a kornia.model example here once the module is ready.

README.md
```diff
@@ -24,11 +24,53 @@ English | [简体中文](README_zh-CN.md)
...
+3. **AI Models**<br>
+Leverage pre-trained AI models optimized for a variety of vision tasks, all within the Kornia ecosystem.
+    - **Face Detection**: YuNet
+    - **Feature Matching**: LoFTR, LightGlue
+    - **Feature Descriptor**: DISK, DeDoDe, SOLD2
+    - **Segmentation**: SAM
+    - **Classification**: MobileViT, VisionTransformer.
 
-It consists of a set of routines and differentiable modules to solve generic computer vision problems. At its core, the package uses *PyTorch* as its main backend both for efficiency and to take advantage of the reverse-mode auto-differentiation to define and compute the gradient of complex functions.
+<details>
```
shijianjian, line 49: For the best, there should be some code snippet or tutorials for each op.

edgarriba: agree -- add as ticket

README.md
edgarriba: yep, I vote for RTDETR @cjpurackal

PR #2963, opened 2024-07-22
[Feat] init kornia module

shijianjian:
### Usage Demo

```python
import numpy as np
from PIL import Image
import kornia


# Support transformation
a = kornia.color.gray.RgbToGrayscale()
a = kornia.enhance.AdjustHue(.5)

# Support augmentation
a = kornia.augmentation.RandomAffine((-45., 45.), p=1.)

# Support augmentation sequentail
a = kornia.augmentation.AugmentationSequential(
    kornia.augmentation.RandomAffine((-45., 45.), p=1.),
    kornia.augmentation.RandomBrightness((0.,1.), p=1.)
)

a("img.jpeg")
a.save("Kornia-output.jpg")  # Save a single image

img = Image.open("img.jpeg")
img = np.stack([np.array(img)] * 10)

a(img)
a.save()  # Save a batch of images with a random name

# To restore the original behaviour, you may simply
a.disable_features = True
a("img.jpeg")  # An error will be thrown as we require tensor input only.
```


edgarriba (changes requested):
add some tests. For the external libraries maybe we can use here the lazy loader too ?

kornia/core/external.py
```diff
@@ -0,0 +1,69 @@
...
+    This class defers the import of a module until an attribute of the module is accessed.
+    It helps in reducing the initial load time and memory usage of a script, especially when
+    dealing with large or optional dependencies that might not be used in every execution.
+
+    Attributes:
+        module_name: The name of the module to be lazily loaded.
+        module: The actual module object, initialized to None and loaded upon first access.
+    """
+
+    def __init__(self, module_name: str):
```
edgarriba, line 16:
```suggestion
    def __init__(self, module_name: str) -> None:
```

kornia/core/external.py
```diff
@@ -0,0 +1,69 @@
...
+        """Initializes the LazyLoader with the name of the module.
+
+        Args:
+            module_name (str): The name of the module to be lazily loaded.
+        """
+        self.module_name = module_name
+        self.module = None
+
+    def _load(self):
```
edgarriba, line 25:
```suggestion
    def _load(self) -> None:
```

kornia/core/external.py
```diff
@@ -0,0 +1,69 @@
...
+        This method is called when the built-in dir() function is used on the LazyLoader instance.
+        It ensures that the module is loaded and then returns the list of attributes of the module.
+
+        Returns:
+            list: The list of attributes of the loaded module.
+        """
+        self._load()
+        return dir(self.module)
+
+
```
edgarriba, line 67: now is fine as first version, but we might want to have this as some sort of registry api later?

kornia/core/module.py
```diff
@@ -0,0 +1,273 @@
...
+        if isinstance(arg, (Tensor,)):
+            return True
+        # Make sure that the numpy and PIL are not necessarily needed to be imported.
+        if isinstance(arg, (np.ndarray,)):
+            return True
+        if isinstance(arg, (Image.Image)):
+            return True
+        return False
+
+    def to_tensor(self, x: Any) -> Tensor:
```
edgarriba, line 101: should we have a `.to_image()` -> https://kornia.readthedocs.io/en/latest/image.html
can be done later in a ticket

kornia/core/module.py
```diff
@@ -0,0 +1,273 @@
...
+
+        Returns:
+            Tensor: The converted tensor.
+        """
+        if isinstance(x, str):
+            return kornia.io.load_image(x, kornia.io.ImageLoadType.UNCHANGED) / 255
+        elif isinstance(x, Tensor):
+            return x
+        elif isinstance(x, np.ndarray) and len(x.shape) == 3:
+            return from_numpy(x).permute(2, 0, 1).float() / 255
```
edgarriba, line 117: reuse `image_to_tensor` -- to avoid circular maybe lets move utils.image under kornia.core.image

kornia/core/module.py
```diff
@@ -0,0 +1,273 @@
...
+
+        if len(self._output_image.shape) == 3:
+            out_image = self._output_image
+        elif len(self._output_image.shape) == 4:
+            if n_row is None:
+                n_row = math.ceil(self._output_image.shape[0] ** 0.5)
+            out_image = kornia.utils.image.make_grid(self._output_image, n_row, padding=2)
+        else:
+            raise ValueError
+        return Image.fromarray((out_image.permute(1, 2, 0).squeeze().numpy() * 255).astype(np.uint8))
```
edgarriba, line 204: missed the show call ?

kornia/core/module.py
```diff
@@ -0,0 +1,273 @@
...
+
+    @property
+    def disable_features(self):
+        return self._disable_features
+
+    @disable_features.setter
+    def disable_features(self, value: bool = True):
+        self._disable_features = value
+
+    def __call__(
```
edgarriba, line 248: we should test all this against onnx and dynamo / compile

edgarriba (commented):
I like this approach -- I'm seeing that several kinda starters guide tutorials can be written from here to cover how to manipulate images via the `ImageModule` and then few of them for the `LazyLoader` to integrate models. @lappemic maybe this is something that you could work out with @shijianjian and @cjpurackal new collaborator

edgarriba:
> Can some help update all the documentation? We may safely remove lots of boiler template IO code now.

what is required to do ? maybe file a ticket to be done a posteriori /cc @lappemic

shijianjian:
> > Can some help update all the documentation? We may safely remove lots of boiler template IO code now.
> 
> what is required to do ? maybe file a ticket to be done a posteriori /cc @lappemic

We need to update all the docs/tutorials with the new image loading, saving, and showing approach.

kornia/core/external.py
shijianjian: no idea. It is a bit overcomplicated for now. we only need several libs for internal usage. So, no register is needed for now.

kornia/core/module.py
shijianjian: All the methods theirselves are returning the image, only need an extra `.numpy()`. Not quite sure if we need this or not.

kornia/core/module.py
shijianjian: I tested with JIT, not sure about the others.

kornia/core/module.py
```diff
@@ -0,0 +1,275 @@
...
+            arg: The argument to check.
+
+        Returns:
+            bool: True if valid, False otherwise.
+        """
+        if isinstance(arg, (str,)) and os.path.exists(arg):
+            return True
+        if isinstance(arg, (Tensor,)):
+            return True
+        # Make sure that the numpy and PIL are not necessarily needed to be imported.
```
johnnv1, line 93: i think this and the Image.Image access to check the type will always try to lazyload the modules no? If yes, it'll always crash if we do not have the deps

kornia/core/module.py
```diff
@@ -0,0 +1,275 @@
...
+        Supports image path, numpy array, PIL image, and raw tensor.
+
+        Args:
+            x: The input to convert.
+
+        Returns:
+            Tensor: The converted tensor.
+        """
+        if isinstance(x, (str,)):
+            return kornia.io.load_image(x, kornia.io.ImageLoadType.UNCHANGED) / 255
```
johnnv1, line 112: for a further PR -- support other than 8bits images norm

kornia/core/module.py
```diff
@@ -0,0 +1,275 @@
...
+
+        if len(self._output_image.shape) == 3:
+            out_image = self._output_image
+        elif len(self._output_image.shape) == 4:
+            if n_row is None:
+                n_row = math.ceil(self._output_image.shape[0] ** 0.5)
+            out_image = kornia.utils.image.make_grid(self._output_image, n_row, padding=2)
+        else:
+            raise ValueError
+
```
johnnv1, line 196: Shouldn't the terminal print you implemented some time ago be supported here?

kornia/core/module.py
johnnv1: +1, since it'll go as module for all things

kornia/core/module.py
shijianjian: If the function returns within the first str and tensor type checks, it will not come to execute the attribute access. Then it will not crash.

One drawback is that those lines will never pass the static type check since those modules are loaded in the runtime.

kornia/core/module.py
shijianjian: Yes, maybe.

kornia/core/module.py
shijianjian: is the dynamo tests in the CI enough?

shijianjian:
@johnnv1 what is that failed job?

kornia/core/module.py
edgarriba: naming this method `show` is a bit confusing

kornia/core/module.py
edgarriba: yes

kornia/core/module.py
shijianjian: why? any other names?

kornia/core/module.py
shijianjian: Then it should be good now.

kornia/utils/image.py
```diff
@@ -211,6 +211,47 @@ def forward(self, x: Any) -> Tensor:
...
+
+    Returns:
+        Tensor: The combined image grid.
+    """
+    if not isinstance(tensor, torch.Tensor):
+        raise TypeError("Input tensor must be a PyTorch tensor.")
+
+    B, C, H, W = tensor.shape
+    if n_row is None:
+        n_row = int(torch.sqrt(torch.tensor(B, dtype=torch.float32)).ceil())
```
edgarriba, line 230: For later — Maybe not needed torch to compute n_row , use `math`

edgarriba:
@shijianjian can you create separate ticks with things todo or improve from this PR ?

kornia/augmentation/container/augment.py
```diff
@@ -469,6 +469,52 @@ def forward(  # type: ignore[override]
...
+            data_keys = self.transform_op.preproc_datakeys(in_data_keys)
+
+            if len(data_keys) > 1 and data_keys.index(DataKey.INPUT):
+                # NOTE: we may update it later for more supports of drawing boxes, etc.
+                idx = data_keys.index(DataKey.INPUT)
+                if output_type == "tensor":
+                    self._output_image = _output_image
+                    self._output_image[idx] = self._detach_tensor_to_cpu(_output_image[idx])
+                else:
+                    self._output_image[idx] = _output_image[idx]
```
ashnair1, line 511: idx is the index of the "image"/"input" key but it's being used as a dict key here. 

`self._output_image[idx]` and `_output_image[idx]` should be `self._output_image[data_keys[idx]]` and `_output_image[data_keys[idx]]` respectively

kornia/augmentation/container/augment.py
```diff
@@ -469,6 +469,52 @@ def forward(  # type: ignore[override]
...
+            )(super(ImageSequential, self).__call__)
+            _output_image = decorated_forward(*inputs, **kwargs)
+
+            if len(inputs) == 1 and isinstance(inputs[0], dict):
+                original_keys, in_data_keys, inputs, invalid_data = self._preproc_dict_data(inputs[0])
+            else:
+                in_data_keys = kwargs["data_keys"] if "data_keys" in kwargs else self.data_keys
+            data_keys = self.transform_op.preproc_datakeys(in_data_keys)
+
+            if len(data_keys) > 1 and data_keys.index(DataKey.INPUT):
```
ashnair1, line 504: Shouldn't this be checking if `DataKey.INPUT` is in data_keys? According to this condition, we only enter the if block if `len(data_keys) > 1` and if INPUT is not the first datakey.

PR #2986, opened 2024-08-17
initial commit for cli

edgarriba:
#### Changes

Fixes # (issue)

#### Type of change

- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/kornia/cli.py
+++ b/kornia/cli.py
@@ -2,62 +2,29 @@
 
 import argparse
 import logging
-from dataclasses import dataclass
 from pathlib import Path
 
-import torch
-
-from kornia.contrib import FaceDetector
+from kornia.contrib.face_detection import YuFaceDetectNet
 
 logging.basicConfig(level=logging.DEBUG, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s")
 logger = logging.getLogger(__name__)
 
 
-@dataclass
-class ExportOnnxModelConfig:
-    """Configuration for exporting a model to ONNX format."""
-
-    model: torch.nn.Module
-    img: torch.Tensor
-    output_path: Path
-    input_names: list[str]
-    output_names: list[str]
-    dynamic_axes: dict[str, dict[int, str]]
-
-
-def export_onnx_model(config: ExportOnnxModelConfig) -> None:
-    """Export a model to ONNX format."""
-    torch.onnx.export(
-        config.model,
-        config.img,
-        config.output_path,
-        input_names=config.input_names,
-        output_names=config.output_names,
-        dynamic_axes=config.dynamic_axes,
-    )
-
-    if config.output_path.exists():
-        logger.info("Model exported to %s", config.output_path)
-    else:
-        logger.error("Error exporting model to %s", config.output_path)
-
-
-def export_onnx_model_config_resolver(model_name: str, output_path: Path) -> ExportOnnxModelConfig:
+def export_onnx_model_resolver(model_name: str, output_path: Path) -> None:
     """Resolve the configuration for exporting a model to ONNX format."""
-    if model_name == "kornia/yunet":
-        face_detector = FaceDetector()
-        config = ExportOnnxModelConfig(
-            model=face_detector.model,
-            img=torch.rand(1, 3, 320, 320),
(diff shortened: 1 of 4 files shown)


shijianjian:
I think we may have a KorniaModel module that has a to_onnx function, then we make it like 'YuNet().to_onnx()'. 

I was thinking that the CLI may do some syntax sugar such as 'kornia-cli rgb2gray/YuNet IMAGE-PATH OUTPUT-PATH'.

Maybe we can also add a process chain such as 'kornia-cli rgb2gray->resize<256,256>->YuNet IMAGE-PATH OUTPUT-PATH'.

Probably then we can have some user-defined chain like 'kornia-cli chain edit'. It will then pop up a VIM of JSON that contains user defined chains. 


Just some random thoughts.

edgarriba:
@shijianjian
> I think we may have a KorniaModel module that has a to_onnx function, then we make it like 'YuNet().to_onnx()'.

why don't we extend the `ImageModule` ?

> I was thinking that the CLI may do some syntax sugar such as 'kornia-cli rgb2gray/YuNet IMAGE-PATH OUTPUT-PATH'.

yes, but i would like to have different commands layers: `export` (to export to onnx, tensorrt, openvino, etc), `compute` (mostly inference models and run scripts`, `serve` (start a fastapi service and expose minimum apis, like to connect from a camera, or models via grpc/proto (json is too expensive to images/videos)

> Maybe we can also add a process chain such as 'kornia-cli rgb2gray->resize<256,256>->YuNet IMAGE-PATH OUTPUT-PATH'.

I'd call it `pipe` -- > `kornia-cli pipe src=webcam ! rgb2gray ! facedetect model=yunet nms=0.8 ! sink=rerun`
Which internally can use Limbus, and advanced version can use rust.

pyproject.toml
```diff
@@ -1,21 +1,13 @@
 [build-system]
 build-backend = "setuptools.build_meta"
-requires = [
-  "setuptools>=61.2",
-]
+requires = [ "setuptools>=61.2" ]
```
johnnv1, line 3: some unrelated to the patch. We can remove the pyproject formater from pre-commit and fix this file. It has been updated some time ago splitting some things (like ruff configs) unnecessary -- could be in another PR

kornia/cli.py
```diff
@@ -0,0 +1,101 @@
+"""Command-line interface for Kornia."""
```
johnnv1, line 1: personally, i would go with 

```console
foo@bar:/tmp/kornia $ tree kornia
kornia/
├── augmentation
│   ├── ...
├── ...
├── commands
│   ├── main.py // principal file with the principal CLI config
│   ├── ...
│   └── exporter.py // file with all export-related functions
├── ...
```

I can reorganize to have something like this above, in a further PR, if it makes sense for you

here is an example of it https://github.com/johnnv1/CCAgT-utils/tree/first-release/CCAgT_utils/commands

kornia/cli.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if config.output_path.exists():
+        logger.info("Model exported to %s", config.output_path)
+    else:
+        logger.error("Error exporting model to %s", config.output_path)
+
+
+def export_onnx_model_config_resolver(model_name: str, output_path: Path) -> ExportOnnxModelConfig:
+    """Resolve the configuration for exporting a model to ONNX format."""
+    if model_name == "kornia/yunet":
```
johnnv1, line 47: i would assume, any model without the explicit module is from kornia
```suggestion
    if model_name in {"kornia/yunet", "yunet"}:
```

kornia/cli.py
edgarriba: yes, why not. My only question now is whether this should be packaged with the libirary, or where exactly should live, plus what's parts should be an api (like the KorniaModule that Jian suggested)

kornia/cli.py
edgarriba: once matured, this should be an enum to ve less error prone

pyproject.toml
edgarriba: +1

edgarriba:
will be somehow part of #3013

PR #2991, opened 2024-08-22
Add feature benchmark numbers to doc

ducha-aiki:
#### Type of change

- [x] 📚  Documentation Update


edgarriba:
Ive seen that the api reference section got broken by the title of this section. TBH, I’d place this section somewhere else as is not api documentation. cc @johnnv1 @shijianjian

PR #2993, opened 2024-08-23
Quaternion conversions

tjelinek:
#### Changes
- Fixed documentation error in function Quaternion.to_euler() containing the description of Quaternion.from_euler()
- Typo in documentation (euler angles -> Euler angles)
- Added Quaternion.to_axis_angle() method (since Quaternion.matrix(), and Quaternion.to_euler() exist as well as conversion Quaternion.from_axis_angle(), it makes sense to fill conversion function to this commonly used rotation representation format).
- Added tests for Quaternion.to_axis_angle() with manually entered expected results on random vectors.

#### Type of change
- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases

### DIFF
--- a/kornia/geometry/quaternion.py
+++ b/kornia/geometry/quaternion.py
@@ -13,7 +13,7 @@ from kornia.geometry.conversions import (
     normalize_quaternion,
     quaternion_from_euler,
     quaternion_to_rotation_matrix,
-    rotation_matrix_to_quaternion,
+    rotation_matrix_to_quaternion, quaternion_to_axis_angle,
 )
 from kornia.geometry.linalg import batched_dot_product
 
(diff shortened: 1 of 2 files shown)


kornia/geometry/quaternion.py
```diff
@@ -314,6 +312,17 @@ def from_axis_angle(cls, axis_angle: Tensor) -> "Quaternion":
         """
         return cls(axis_angle_to_quaternion(axis_angle))
 
+    def to_axis_angle(self) -> Tensor:
+        """Convert the quaternion to a rotation matrix of shape :math:`(B, 3, 3)`.
```
edgarriba, line 316: Docs are wrong

edgarriba (changes requested):
Fix docs and good to go

kornia/geometry/quaternion.py
tjelinek: Fixed (note that I did not specify the output shape - Quaternion does not enforce the data to have a batch dimension) even if the docs are written as it did. The conversion works both with and without batch dimension. But Quaternion.identity().matrix() gives an output of shape (1, 3, 3) even if Quaternion.data is of shape (4, ).
This inconsistency in both the conversion functions and the documentation is worth fixing.

edgarriba: I see, thanks to look at it. Are you open to fix that ?

edgarriba (approved):

kornia/geometry/quaternion.py
tjelinek: Yes, I will create an issue, where we can discuss and take a look at it.

PR #2996, opened 2024-08-26
compute area for all types of boxes

Isalia20:
#### Changes
Changes, Boxes.compute_area(). Since Boxes is a container of B, 4, 2 tensor or B, N, 4, 2 tensor, library shouldn't assume that the input with always be a perfect(non-rotated) rectangle. Current compute_area assumes this and doesn't take into account that box might not be a rectangle but a quadrilatereal.

One note, I couldn't find any tests for compute_area. More than happy to get feedback on where I can add them to make sure that area computed is correct.

#### Type of change

- [x] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -392,9 +392,19 @@ class Boxes:
 
     def compute_area(self) -> torch.Tensor:
         """Returns :math:`(B, N)`."""
-        w = self._data[..., 1, 0] - self._data[..., 0, 0]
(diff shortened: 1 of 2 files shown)


edgarriba (changes requested):
Can you extend with a test for this proving that the previous implementation was broken ?

Isalia20:
I've added tests, it won't work with the old compute_area function but works with the new one

edgarriba (approved):

PR #3013, opened 2024-09-07
Init Kornia.onnx module

shijianjian:
Added support for ONNXSequential for chaining ONNX models into one graph.


kornia/onnx/sequential.py
```diff
@@ -0,0 +1,150 @@
+from typing import Optional, Union
+
+from kornia.core.external import numpy as np
+from kornia.core.external import onnx
+from kornia.core.external import onnxruntime as ort
+
+__all__ = ["ONNXSequential"]
+
+
+class ONNXSequential:
```
edgarriba, line 10: I would call it just `ONNX `, as pytorch user I initially thought that is a wrap of nn.Sequential, which seems it not but with more powerful features

kornia/onnx/sequential.py
```diff
@@ -0,0 +1,150 @@
...
+        io_maps:
+            An optional list of list of tuples specifying input-output mappings for combining models.
+            If None, we assume the default input name and output name are "input" and "output" accordingly, and
+            only one input and output node for each graph.
+            If not None, `io_maps[0]` shall represent the `io_map` for combining the first and second ONNX models.
+    """
+
+    def __init__(
+        self,
+        *args: Union[onnx.ModelProto, str],  # type:ignore
```
edgarriba, line 29: Would someone useful to support in-memory models too, to simplify to go from a raw model + pre/post processing to the combined pipeline

kornia/onnx/sequential.py
```diff
@@ -0,0 +1,150 @@
...
+            arg: Either an ONNX ModelProto object or a file path to an ONNX model.
+
+        Returns:
+            onnx.ModelProto: The loaded ONNX model.
+        """
+        if isinstance(arg, str):
+            return onnx.load(arg)  # type:ignore
+        return arg
+
+    def _combine(self, io_maps: Optional[list[tuple[str, str]]] = None) -> onnx.ModelProto:  # type:ignore
```
edgarriba, line 66: This feature is very powerful to create high level pipeline but from the docs is a but unclear how it works and how you link the different models you provide in the constructor

shijianjian: there is a `io_map` you need to define. with single one-to-one model, we assume the input node is "input" and output node is "output". If you got multiple input or output nodes, you gonna need to pass the `io_map` arg

kornia/onnx/sequential.py
shijianjian: isn't the loaded onnx model in-memory? if not, how to support?

kornia/onnx/sequential.py
shijianjian: This behaves like the nn.Sequential. Or maybe ONNXChainer or ONNXCombiner? those are more boring names it looks.

edgarriba: ok, let's start by sequential, but i potentially foresee non sequential pipelines

kornia/core/module.py
```diff
@@ -11,6 +13,51 @@
...
+        if not self.ONNX_EXPORTABLE:
+            raise RuntimeError("This object cannot be exported to ONNX.")
+
+        if input_shape is None:
+            input_shape = self.ONNX_DEFAULT_INPUTSHAPE
+        if output_shape is None:
+            output_shape = self.ONNX_DEFAULT_OUTPUTSHAPE
+
+        if onnx_name is None:
+            onnx_name = f"Kornia-{self.__class__.__name__}.onnx"
```
edgarriba, line 36:
```suggestion
            onnx_name = f"kornia-{kornia.__verision__}-{self.__class__.__name__}.onnx"
```
or add some sort of metadata, maybe as discussed by adding it to the proto metadata fields
https://qgis-plugin-deepness.readthedocs.io/en/latest/creators/creators_add_metadata_to_model.html

kornia/core/module.py
```diff
@@ -11,6 +13,51 @@
...
+
+        # Creating a dummy input with the given shape
+        psuedo_shape = (1, 3, 256, 256)
```
edgarriba, line 39:
```suggestion
        pseudo_shape = (1, 3, 256, 256)
```

kornia/onnx/utils.py
```diff
@@ -0,0 +1,140 @@
...
+        Returns:
+            onnx.ModelProto: The loaded ONNX model.
+        """
+        if model_name.startswith("hf://"):
+            model_name = model_name[len("hf://") :]
+            file_path = self._get_file_path(model_name, self.cache_dir)
+            if not os.path.exists(file_path):
+                # Construct the raw URL for the ONNX file
+                url = f"https://huggingface.co/kornia/ONNX_models/resolve/main/{model_name}.onnx"
+                self.download(url, file_path, **kwargs)
```
edgarriba, line 72: why don't we use their hub api ?

kornia/color/gray.py
```diff
@@ -125,6 +125,9 @@ class GrayscaleToRgb(Module):
         >>> output = rgb(input)  # 2x3x4x5
     """
 
+    ONNX_DEFAULT_INPUTSHAPE: ClassVar[list[int]] = [-1, 1, -1, -1]
```
edgarriba, line 128: can you document somewhere how's this used for ? probable needs to be updated in the contributors docs ?

kornia/onnx/sequential.py
shijianjian: Ah. Yes, but this is more complex to handle. Not as convenient as this one-to-one mapping. another class maybe better for the future.

kornia/constants.py
```diff
@@ -7,6 +8,8 @@
 
 __all__ = ["pi", "DType", "Resample", "BorderType", "SamplePadding", "TKEnum"]
 
+logging.basicConfig(level=logging.INFO)
```
johnnv1, line 11: note to remove it later

kornia/utils/sample.py
```diff
@@ -0,0 +1,82 @@
+import logging
+import os
+from typing import List, Optional, Union
+
+import requests
+
```
johnnv1, line 6: will it be used internally in kornia?

requirements/requirements-dev.txt
```diff
@@ -2,11 +2,13 @@ accelerate
 coverage
 diffusers
 mypy
-numpy<2
+numpy<3
```
johnnv1, line 5:
```suggestion
numpy<2
```

kornia/models/edge_detector/dexined.py
```diff
@@ -26,10 +26,8 @@ def to_onnx(
         image_size: Optional[int] = 352,
     ) -> Tuple[str, EdgeDetector]:
         edge_detector = DexiNedBuilder.build(pretrained, image_size)
-        model_name = "DexiNed"
         if onnx_name is None:
-            _model_name = model_name
-            onnx_name = f"Kornia-RTDETR-{_model_name}-{image_size}.onnx"
+            onnx_name = f"Kornia-DexiNed-{image_size}.onnx"
```
edgarriba, line 30:
```suggestion
            onnx_name = f"kornia_dexined_{image_size}.onnx"
```
can we default using lowercase as it is case sensitive plus underscores ?

kornia/models/segmentor/segmentation_models.py
```diff
@@ -47,27 +51,30 @@ def __init__(
         )
 
     def preprocessing(self, input: Tensor) -> Tensor:
-        if self.preproc_params["input_space"] == "RGB":
+        # Ensure the color space transformation is ONNX-friendly
+        input_space = self.preproc_params["input_space"]
```
edgarriba, line 55: To be consistant with the image api and protos let’s call this `input_color_space`

kornia/models/segmentor/segmentation_models.py
```diff
@@ -47,27 +51,30 @@ def __init__(
...
+        input = kornia.color.rgb_to_bgr(input) if input_space == "BGR" else input  # Assume input is already RGB if not BGR
```
edgarriba: For internal development we are usually importing as `from Kornia.color.rgb import rgb_to_bgr` to avoid conflicts with the namespace

kornia/models/segmentor/segmentation_models.py
```diff
@@ -47,27 +51,30 @@ def __init__(
...
+
+        # Normalize input range if needed
+        input_range = self.preproc_params["input_range"]
```
edgarriba, line 61: Wait, is this something coming from that library ?

kornia/models/segmentor/segmentation_models.py
```diff
@@ -47,27 +51,30 @@ def __init__(
...
+        if input_range[1] == 255:
```
edgarriba, line 62: int(input_range[1]) == 255

kornia/models/segmentor/segmentation_models.py
```diff
@@ -47,27 +51,30 @@ def __init__(
...
+            raise ValueError(f"Unsupported input range: {input_range}")
 
-        if self.preproc_params["input_range"] is not None:
-            if input.max() > 1 and self.preproc_params["input_range"][1] == 1:
-                input = input / 255.0
-
-        if self.preproc_params["mean"] is None:
-            mean = tensor(self.preproc_params["mean"]).to(input.device)
+        # Handle mean and std normalization
+        if self.preproc_params["mean"] is not None:
```
edgarriba, line 70: `tensor([self.preproc_params["mean"]], device=input.device)`

In theory this allocates directly into the device

kornia/filters/dexined.py
```diff
@@ -125,8 +125,7 @@ def compute_out_features(self, idx: int, up_scale: int) -> int:
 
     def forward(self, x: Tensor, out_shape: list[int]) -> Tensor:
         out = self.features(x)
-        if out.shape[-2:] != out_shape:
```
edgarriba, line 128: Why removing that ?

kornia/models/__init__.py
```diff
@@ -1 +1,3 @@
 from . import detector
+from . import segmentor
```
edgarriba: I’d call the modules defining they category name
- detection
- segmentation 
- tracking

in this way seems that there’s a single model per category?

kornia/models/detector/base.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+# NOTE: probably we should use a more generic name like BoundingBox2D
+# and add a BoundingBox3D class for 3D bounding boxes. Also for serialization
+# we should have an explicit class for each format to make it more production ready
+# specially to serialize to protobuf and not saturate at a high rates.
+
+
+@dataclass(frozen=True)
+class BoundingBox:
```
edgarriba, line 46: @shijianjian i would like to hear your opinion about this struct we created, as it should align with the protos library structs (maybe we should use the protobufs directly here ?)

kornia/models/detector/base.py
```diff
@@ -0,0 +1,204 @@
...
+            out_img = image[None].clone()
+            for out in detection:
+                out_img = draw_rectangle(
+                    out_img,
+                    torch.Tensor([[[out[-4], out[-3], out[-4] + out[-2], out[-3] + out[-1]]]]),
+                )
+            if output_type == "torch":
+                output.append(out_img[0])
+            elif output_type == "pil":
+                output.append(Image.fromarray((out_img[0] * 255).permute(1, 2, 0).numpy().astype(np.uint8)))  # type: ignore
```
edgarriba, line 159: Use utils.tensor_to_image ?

kornia/models/detector/base.py
```diff
@@ -0,0 +1,204 @@
...
+        self, images: Union[Tensor, list[Tensor]], detections: Optional[Tensor] = None, directory: Optional[str] = None
+    ) -> None:
+        """Saves the output image(s) to a directory.
+
+        Args:
+            name: Directory to save the images.
+            n_row: Number of images displayed in each row of the grid.
+        """
+        if directory is None:
+            name = f"detection-{datetime.datetime.now(tz=datetime.timezone.utc).strftime('%Y%m%d%H%M%S')!s}"
```
edgarriba, line 174: underscores

kornia/models/detector/base.py
```diff
@@ -0,0 +1,204 @@
...
+            directory = os.path.join("Kornia_outputs", name)
```
edgarriba, line 175: Lowercase

kornia/models/edge_detector/dexined.py
```diff
@@ -0,0 +1,55 @@
...
+class DexiNedBuilder:
+
+    @staticmethod
+    def build(
+        pretrained: bool = True, image_size: Optional[int] = 352
+    ) -> EdgeDetector:
+        model = DexiNed(pretrained=pretrained)
+        return EdgeDetector(
+            model,
+            ResizePreProcessor((image_size, image_size)) if image_size is not None else nn.Identity(),
```
edgarriba, line 18: I’m more a fan of explicit things to avoid shape bugs

```ResizePreProcessor(height=image_size, width=image_size)```

File: kornia/models/edge_detector/dexined.py [hunk 24]
```diff
@@ -0,0 +1,55 @@
...
+        )
+
+    @staticmethod
+    def to_onnx(
+        onnx_name: Optional[str] = None,
+        pretrained: bool = True,
+        image_size: Optional[int] = 352,
+    ) -> Tuple[str, EdgeDetector]:
+        edge_detector = DexiNedBuilder.build(pretrained, image_size)
+        model_name = "DexiNed"
```
edgarriba: Lowercase

kornia/models/edge_detector/dexined.py
```diff
@@ -0,0 +1,55 @@
...
+        if onnx_name is None:
+            _model_name = model_name
+            onnx_name = f"Kornia-RTDETR-{_model_name}-{image_size}.onnx"
```
edgarriba: Lowercase + underscores

kornia/models/utils.py
```diff
@@ -0,0 +1,78 @@
...
+from kornia.geometry.transform import resize
+
+
+class ResizePreProcessor(Module):
+    """This module resizes a list of image tensors to the given size.
+
+    Additionally, also returns the original image sizes for further post-processing.
+    """
+
+    def __init__(self, size: tuple[int, int], interpolation_mode: str = "bilinear") -> None:
```
edgarriba, line 17: Can we make height / width explicitly instead of Tuple ?

kornia/models/segmentor/segmentation_models.py
shijianjian: yes

kornia/filters/dexined.py
shijianjian: this is not ONNX friendly

kornia/models/detector/base.py
shijianjian: Yes, I think directly using protobuf is the best, but it would then require another dependency. I would prefer leaving it here for now.

But in terms of upgrading, maybe we should do it for the whole library.

shijianjian:
Any ideas on what exactly happened on this?

```
Run uv pip install --system --editable .[dev,x]
  × No solution found when resolving dependencies:
  ╰─▶ Because there are no versions of distutils and kornia[dev]==0.7.4.dev0
      depends on distutils, we can conclude that kornia[dev]==0.7.4.dev0
      cannot be used.
      And because only kornia[dev]==0.7.4.dev0 is available and you require
      kornia[dev], we can conclude that your requirements are unsatisfiable.
```

shijianjian:
@johnnv1 I think some problems with the CI? kornia[dev] is not found? Did I do something wrong?

kornia/models/__init__.py
```diff
@@ -1,2 +1,8 @@
-from . import detection, segmentation, tracking
+from . import (
+    depth_estimation,
```
edgarriba: Should we have the most basic which is image classification? And eg. Just embedding extraction, and support let’s say Dinov2?

edgarriba: And wondering how can we fit here in terms of api multimodal or foundational models with multi task heads

kornia/models/_hf_models/onnx_community.py
```diff
@@ -4,42 +4,122 @@
...
+        self, model_name: str, model_type: str = "model", cache_dir: Optional[str] = None
     ) -> None:
         self.model_name = model_name
         self.model_type = model_type
-        self.preprocessing = preprocessing
         
         if cache_dir is None:
             cache_dir = kornia_config.hub_onnx_dir
         self.loader = ONNXLoader(cache_dir=os.path.join(cache_dir, self.model_name))
+        self.model_url = f"https://huggingface.co/onnx-community/{self.model_name}/resolve/main/onnx/{self.model_type}.onnx"
```
edgarriba: Does it mean that this loader only works for hugging face ? Can it be generic? Otherwise maybe rename de class name

kornia/models/depth_estimation/depth_anything.py
```diff
@@ -0,0 +1,56 @@
...
+    @staticmethod
+    def build(
+        model_name: str = "depth-anything-v2-small",
+        model_type: str = "model",
+        cache_dir: Optional[str] = None
+    ) -> DepthEstimation:
+        """Exports a DepthAnything model to an ONNX model file.
+
+        Args:
+            save:
```
edgarriba, line 15: Unexisting argument

edgarriba: Same below

docs/source/onnx.rst
```diff
@@ -0,0 +1,137 @@
...
+      onnx_seq.export("combined_model.onnx")
+
+5. **Optimizing with Execution Providers**
+
+   Leverage ONNXRuntime's execution providers for optimized inference. For example, to run the model on a GPU:
+
+   .. code-block:: python
+
+      # Initialize with CUDA execution provider
+      onnx_seq = ONNXSequential(
```
edgarriba, line 83: One feature that might valuable is to provide per mapping a lamba function to inject adapters between ONNX ir, that can be written in ONNX itself as probably not always there will be compatibility between models

docs/source/onnx.rst
```diff
@@ -0,0 +1,137 @@
...
+
+**3. Can I use custom ONNXRuntime session options?**
+
+Absolutely! You can pass your own session options to the `create_session` method to fine-tune performance, memory usage, or logging.
+
+Why Choose ONNXSequential?
+--------------------------
+
+With the increasing adoption of ONNX for model interoperability and deployment, `ONNXSequential` provides a simple yet powerful interface for combining models and operators. By leveraging ONNXRuntime’s optimization and execution provider capabilities, it gives you the flexibility to:
+- Deploy on different hardware (CPU, GPU).
```
edgarriba, line 113: And different engines, list some https://onnxruntime.ai/docs/execution-providers/

docs/source/onnx.rst
```diff
@@ -0,0 +1,137 @@
...
+
+.. autoclass:: kornia.onnx.utils.ONNXLoader
+
+    .. code-block:: python
+
+        onnx_loader = ONNXLoader()
+        # Load a HuggingFace operator
+        onnx_loader.load_model("hf://operators/kornia.color.gray.GrayscaleToRgb")  # doctest: +SKIP
+        # Load a local converted/downloaded operator
+        onnx_loader.load_model("operators/kornia.color.gray.GrayscaleToRgb")  # doctest: +SKIP
```
edgarriba, line 135: We can here to show case remap ? It will be a very valuable function if it’s well supported in ONNX with all the providers, eg for self driving car industry and in general spatial ai/robotics

kornia/core/module.py
```diff
@@ -273,3 +404,59 @@ def __call__(
...
+
+    @property
+    def disable_features(self) -> bool:
+        return self._disable_features
+
+    @disable_features.setter
+    def disable_features(self, value: bool = True) -> None:
+        self._disable_features = value
+
+    def __call__(
```
edgarriba, line 433: @shijianjian i think that here what we want to do is to support iobinding to inject directly torch or even other backers like Keras, jax, tf 
https://stackoverflow.com/questions/75553839/pre-allocating-dynamic-shaped-tensor-memory-for-onnx-runtime-inference

kornia/models/depth_estimation/base.py
```diff
@@ -0,0 +1,69 @@
...
+        """Save the segmentation results.
+
+        Args:
+            images: input tensor.
+            output_type: type of the output.
+
+        Returns:
+            output tensor.
+        """
+        outputs = self.visualize(images, depth_maps, output_type)
```
edgarriba, line 67: Here we might want to save other than jpeg or png but EXR @ducha-aiki

kornia/models/__init__.py
edgarriba: Super resolution is another one https://pytorch.org/tutorials/advanced/super_resolution_with_onnxruntime.html

Maybe add as todos

kornia/onnx/sequential_async.py
```diff
@@ -0,0 +1,230 @@
...
+        If the pipeline is already running, this method has no effect.
+        """
+        if not self.running:
+            self.running = True
+            # Start a worker for each stage
+            for stage_idx in range(self.num_stages):
+                worker_task = asyncio.create_task(self._worker(stage_idx))
+                self.workers.append(worker_task)
+            if self.verbose:
+                print("ONNX Pipeline started.")
```
edgarriba, line 198: use logging, print with asyncio is not recommended

kornia/onnx/sequential.py
```diff
@@ -0,0 +1,149 @@
...
+            *inputs: Inputs to the ONNX model. The number of inputs must match the
+                expected inputs of the session.
+
+        Returns:
+            List: The outputs from the ONNX model inference.
+        """
+        loop = asyncio.get_event_loop()
+        ort_inputs = self._session.get_inputs()
+        ort_input_values = {ort_inputs[i].name: inputs[i] for i in range(len(ort_inputs))}
+        outputs = await loop.run_in_executor(None, self._session.run, None, ort_input_values)
```
edgarriba, line 134: onnxruntime has already its own async method. https://onnxruntime.ai/docs/api/python/api_summary.html#onnxruntime.InferenceSession.run_async

I would leave this method run as synchronous

kornia/core/module.py
```diff
@@ -249,7 +119,7 @@ def to_tensor(self, x: Any) -> Tensor:
             return from_numpy(np.array(x)).permute(2, 0, 1).float() / 255  # type: ignore
         raise TypeError("Input type not supported")
 
-    def to_numpy(self, x: Any) -> np.array:  # type: ignore
+    def to_numpy(self, x: Any) -> "np.array":  # type: ignore
```
edgarriba, line 122: Numpy has now built in typing https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray

kornia/onnx/mixin.py
```diff
@@ -15,6 +19,131 @@
...
+class ONNXExportMixin:
+    """Mixin class that provides ONNX export functionality for objects that support it.
+
+    Attributes:
+        ONNX_EXPORTABLE:
+            A flag indicating whether the object can be exported to ONNX. Default is True.
+        ONNX_DEFAULT_INPUTSHAPE:
+            Default input shape for the ONNX export. A list of integers where `-1` indicates
+            dynamic dimensions. Default is [-1, -1, -1, -1].
+        ONNX_DEFAULT_OUTPUTSHAP:
```
edgarriba, line 32:
```suggestion
        ONNX_DEFAULT_OUTPUTSHAPE:
```

kornia/core/module.py
```diff
@@ -6,7 +6,7 @@
 
 import kornia
```
edgarriba, line 7: @shijianjian try to remove this import

kornia/core/mixin/image_module.py
```diff
@@ -207,12 +223,18 @@ def save(self, name: Optional[str] = None, n_row: Optional[int] = None) -> None:
             name: Directory to save the images.
             n_row: Number of images displayed in each row of the grid.
         """
+        try:
+            from kornia.io import write_image  # pylint: disable=C0415
+            from kornia.utils.image import make_grid  # pylint: disable=C0415
+        except ImportError as exc:
+            raise ImportError("Kornia not installed. Please install Kornia to use this feature.") from exc
```
shijianjian, line 230: I do not understand this part. Any other solutions, LazyLoading this can also be wierd.

shijianjian: why having this anyway? Lower version of Kornia will never have those mixin modules.

edgarriba: we can remove the try/catch

kornia/geometry/transform/affwarp.py
```diff
@@ -606,14 +607,48 @@ def resize(
     return output
 
 
+def resize_to_be_divisible(
```
edgarriba, line 608: What’s different from rescale ?

edgarriba (approved):
LGTM @shijianjian merge

PR #3034, opened 2024-10-09
Add ignore_index to dice, focal and tversky loss

### LINKED ISSUE #839: "ignore_index" argment for focal loss and dice loss
When I try to train for segmentation on VOC 2012 with Dice loss, I always run into cuda illegal error and I finally found that this may be caused by out-of-range indexes (255 out of 20 classes). Since pytorch CrossEntropyLoss has an argment "ignore_index" to deal with this issue, I wonder if there is a proper way to deal with uncertain indexes?

falckt:
#### Changes
* adds `ignore_index` argument to dice and tversky losses. 
* a second commit also adds code to the focal loss, but without tests. I couldn't easily extend an existing test case to properly test the new parameter. IMHO, I would drop this second commit for now.

fixes #839

#### Type of change
- [x] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/kornia/losses/_utils.py
+++ b/kornia/losses/_utils.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 from typing import Optional
 
 import torch
--- a/kornia/losses/_utils.py
+++ b/kornia/losses/_utils.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+
 from typing import Optional
 
 import torch
--- a/kornia/losses/_utils.py
+++ b/kornia/losses/_utils.py
@@ -1,0 +1,21 @@
+from typing import Optional
+
+import torch
+
+
+def mask_ignore_pixels(
+    target: torch.Tensor, ignore_index: Optional[int]
+) -> tuple[torch.Tensor, Optional[torch.Tensor]]:
+    if ignore_index is None:
+        return target, None
+
+    target_mask = target != ignore_index
+
+    if target_mask.all():
+        return target, None
+
+    # map invalid pixels to a valid class (0)
+    # they need to be manually excluded from the loss computation after
+    target = target.where(target_mask, target.new_zeros(1))
+
+    return target, target_mask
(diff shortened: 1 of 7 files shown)


shijianjian (approved):
I think it is an important feature but missed everywhere in many loss functions. Is it possible to apply this on other losses?

falckt:
I initially tried adding it to all the semantic segmentation losses, but the logic needed for the lovasz losses would need to be different. Unfortunately I don't have the time to dig into them and see how to implement it properly.

I did add basic test coverage for the focal losses though.

kornia/losses/focal.py
```diff
@@ -20,6 +20,7 @@ def focal_loss(
     gamma: float = 2.0,
     reduction: str = "none",
     weight: Optional[Tensor] = None,
+    ignore_index: int = -100,
```
edgarriba, line 23: can this be set as None too ?

kornia/losses/tversky.py
```diff
@@ -66,9 +72,23 @@ def tversky_loss(
     # compute softmax over the classes axis
     pred_soft: torch.Tensor = F.softmax(pred, dim=1)
 
+    # check if all pixels are valid
+    valid_idx = target != ignore_index
```
edgarriba, line 76: maybe this can be wrapped into a reusable function across all the losses

edgarriba (approved):
LGTM. Minor comment that can be improved in a further PR

kornia/losses/focal.py
falckt: I initially followed value and typing from pytorch, e.g.
https://pytorch.org/docs/stable/generated/torch.nn.functional.cross_entropy.html

considering that the kornia losses are pure python, I added a short circuit for None that saves some computation. I left the default value the same as in pytorch though. It is not a breaking change as -100 would crash if fed into one_hot. As such no existing code should have it as a valid label.

kornia/losses/tversky.py
falckt: I pulled it out into a helper function, not sure where to place it though. Any suggestions appreciated.

edgarriba: This sounds good

kornia/losses/_utils.py
```diff
@@ -0,0 +1,21 @@
+from typing import Optional
```
edgarriba, line 1:
```suggestion
from __future__ import annotations
from typing import Optional
```
This will fix the tests

edgarriba (approved):

PR #3051, opened 2024-10-29
fix: mypy typing checks with ivy integration

Sam-Armstrong:
@edgarriba this should fix the typing test failures you were seeing here:

https://github.com/kornia/kornia/actions/runs/11572173527/job/32212962351?pr=3050


edgarriba:
https://github.com/kornia/kornia/pull/3052

PR #3045, opened 2024-10-25
Add Multi-Framework Support to Kornia using Ivy

Sam-Armstrong:
Hey @edgarriba and team,

Following up on our conversation via email a while ago, this is our proposal for integrating Ivy into Kornia, allowing it to be used in TensorFlow, JAX, and NumPy. This integration adds `kornia.to_tensorflow()`, `kornia.to_jax()` and `kornia.to_numpy()` to the Kornia api to facilitate these transpilations. As part of this PR, we’ve also updated the Kornia documentation to reflect these new functions and give brief examples of their usage.

We’ve been testing out Ivy on a large set of the functions and classes in Kornia, and currently have ~90% of cases successfully transpiling to the target framework. You can see the latest numbers on Ivy's [integration tests dashboard](https://github.com/ivy-llc/ivy-integration-tests/blob/main/DASHBOARD.md).

Feel free to try it out, and let us know if anything in the PR needs changing!

Sam, Haris (@hmahmood24) and Yusha (@YushaArif99)
The Ivy Team

#### Type of change
- [ ] 📚  Documentation Update
- [ ] 🔬 New feature (non-breaking change which adds functionality)

PR #3055, opened 2024-10-31
fix failing doctests

Sam-Armstrong:
Fix the docstring tests which started failing after the merging of https://github.com/kornia/kornia/pull/3045. The issue was with ivy not being installed as a dev dependency.


edgarriba:
seems that ivy is not pulling some of the deps

Sam-Armstrong:
Yep there's a couple of minor issues with Ivy, I'm working on getting those fixed and also the potential `kornia.set_backend` functionality at the moment.

johnnv1:
@Sam-Armstrong for the meaning time i pushed in #3061 an workaround -- not sure if the doctests are the better place to test it, maybe we should add it into the test suite itself

PR #3063, opened 2024-11-04
Fix jit related tests

johnnv1:
fix from partially reverting #3058 -- jit only supports old typing

### DIFF
--- a/.github/workflows/scheduled_test_cpu.yml
+++ b/.github/workflows/scheduled_test_cpu.yml
@@ -6,7 +6,6 @@ on:
   schedule:
     - cron: "0 4 * * *"
   workflow_call:
-  pull_request:
 
 concurrency:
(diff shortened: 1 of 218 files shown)


.github/workflows/scheduled_test_cpu.yml
```diff
@@ -5,6 +5,8 @@ on:
     branches: [main, test-me-*]
   schedule:
     - cron: "0 4 * * *"
+  workflow_call:
+  pull_request:
```
johnnv1, line 9: UNDO this one -- it's only to trigger the CI and check if it fixes the related tests

johnnv1, line 9:
```suggestion
```

edgarriba (approved):

PR #3080, opened 2024-12-03
chore (pre-commit): update hooks 

johnnv1:
manually updating hooks since github.com/PyCQA/docformatter has some wrong configuration for pre-commit in the versions outside of the hardcoded rev.


johnnv1:
@edgarriba the failing CI are unrelated with this patch... the collector is failing because somewhere are added a bunch of code without tests or something like that, and the readthedocs is failing to pull some data using our script

edgarriba (approved):

PR #3082, opened 2024-12-03
chore (docstrings): addopt ruff over docformatter

### LINKED ISSUE #3081: ruff: adopt `D` rule for docstrings
since github.com/PyCQA/docformatter has some wrong configuration for pre-commit in the newer versions of the hardcoded rev we are using, we may migrate it to use ruff instead. some changes/updates may be necessary, I am not expecting to only uncomment (https://github.com/kornia/kornia/blob/52cee42a15f4828adf729e28931479ec5375ed20/pyproject.toml#L108) and remove the docformatter hook and it just works. 

instead of fixing all our docs in one go, we should add ignores in the rules as well

johnnv1:
Fixes #3081


pyproject.toml
```diff
@@ -138,6 +138,7 @@ ignore = [
   "UP007",   # Prefer Optional[], Union[] over | due to torch jit scripting
   "UP006",   # Prefer List[], over list due to torch jit scripting
   "UP035",   # Ignore deprecated typing because of jit scripting
+  'D100', 'D101', 'D102','D103', 'D104', 'D105', 'D107', 'D203', 'D204', 'D205', 'D213', 'D400', 'D401', 'D404', 'D406', 'D407', 'D415', 'D417'# docs -- TODO: unconment some of these rules..
```
johnnv1, line 141: we need an issue to discuss and check which of these rules we may want to have

edgarriba (approved):

PR #3084, opened 2024-12-04
Don't detach tensors to cpu

### LINKED ISSUE #3066: AugmentationSequential explicitly moves the output to the CPU if data_keys is given
### Describe the bug

With the 0.7.4 release, augmentations on the GPU are not possible anymore because the output of the input tensor is always explicitly moved to the CPU.

The problem is that `_detach_tensor_to_cpu` is called explicitly on every tensor in [augment.py](https://github.com/kornia/kornia/blob/main/kornia/augmentation/container/augment.py#L507).

### Reproduction steps
The augmented version of the input gets moved to the CPU:

```python
import torch
import kornia

input = torch.randn(1, 3, 10, 10, device="cuda")
mask = torch.randn(1, 10, 10, device="cuda")
aug_list = kornia.augmentation.AugmentationSequential(kornia.augmentation.RandomHorizontalFlip(p=1))

for a in aug_list(input, mask, data_keys=["input", "mask"]):
    print(a.dtype, a.device)
# torch.float32 cpu
# torch.float32 cuda:0
```

Without `data_keys` and only one input, the output device is as expected:
```python
for a in aug_list(input):
    print(a.dtype, a.device)
# torch.float32 cuda:0
```

### Expected behavior

I would expect the augmented version of the input tensor to reside on the GPU.

### Environment

- PyTorch Version (e.g., 1.0): 2.5.1
- OS (e.g., Linux): Ubuntu
- How you installed PyTorch (`conda`, `pip`, source): pip
- Python version: 3.12
- CUDA/cuDNN version: 12.4
- GPU models and configuration: NVIDIA 4090 RTX

### Additional context

Reverting back to 0.7.3 fixes the problem.

ashnair1:
#### Changes

Fixes #3066 

#### Type of change

#### Checklist

### DIFF
--- a/kornia/augmentation/container/augment.py
+++ b/kornia/augmentation/container/augment.py
@@ -507,11 +507,9 @@ class AugmentationSequential(TransformMatrixMinIn, ImageSequential):
                 if output_type == "tensor":
                     self._output_image = _output_image
                     if isinstance(_output_image, dict):
-                        self._output_image[original_keys[idx]] = self._detach_tensor_to_cpu(
-                            _output_image[original_keys[idx]]
-                        )
+                        self._output_image[original_keys[idx]] = _output_image[original_keys[idx]]
                     else:
-                        self._output_image[idx] = self._detach_tensor_to_cpu(_output_image[idx])
+                        self._output_image[idx] = _output_image[idx]
                 elif isinstance(_output_image, dict):
                     self._output_image[original_keys[idx]] = _output_image[original_keys[idx]]
                 else:


edgarriba (approved):
LGTM. Is there any minimal test that we could add to avoid regressions here ?

shijianjian:
If the `meta` device thing worked, can we add a meta device test to ensure that all the device mapped accordingly?

https://pytorch.org/docs/stable/meta.html

shijianjian (approved):

Issue #3083, opened 2024-12-03
ruff (D): adopt some the ignored rules for the docstrings


johnnv1:
we need an issue to discuss and check which of these rules we may want to have

_Originally posted by @johnnv1 in https://github.com/kornia/kornia/pull/3082#discussion_r1868460841_

The ignored rules are (from https://docs.astral.sh/ruff/rules/#pydocstyle-d):

- 'D100' :  Missing docstring in public module
- 'D101' :  Missing docstring in public class
- 'D102' :  Missing docstring in public method
- 'D103' :  Missing docstring in public function
- 'D104' : Missing docstring in public package
- 'D105' :  Missing docstring in magic method
- 'D107' :  Missing docstring in __init__
- 'D203' :  1 blank line required before class docstring
- 'D204' :  1 blank line required after class docstring
- 'D205' :  1 blank line required between summary line and description
- 'D213' :  Multi-line docstring summary should start at the second line
- 'D400' :  First line should end with a period
- 'D401' :  First line of docstring should be in imperative mood: "{first_line}"
- 'D404' :  First word of the docstring should not be "This"
- 'D406' :  Section name should end with a newline ("{name}")
- 'D407' :  Missing dashed underline after section ("{name}")
- 'D415' :  First line should end with a period, question mark, or exclamation point
- 'D417' :  Missing argument description in the docstring for {definition}: {name}

maybe we should adopt some default style used by the community (numpy or google) cc @edgarriba @shijianjian @ducha-aiki 

- https://docs.astral.sh/ruff/faq/#does-ruff-support-numpy-or-google-style-docstrings
- https://docs.astral.sh/ruff/formatter/#docstring-formatting

---

Missing rules to be enable after #3088 

- [ ]  'D100'
- [ ]  'D101'
- [ ]  'D102'
- [ ]  'D103
- [ ]  'D104',
- [ ]  'D105'
- [ ]  'D107'
- [ ]  'D417'

johnnv1:
for me, we should adopt the numpy style/convention for docstrings. And from this list of ignored rules, I would only ignore the  D100 and D105 rules... the docstrings on classes or methods or functions should exist, if it's some API that shouldn't be public (aka backbones used internally)  we should just make it "private" by using a `_` into the prefix of the name...


but to do it, is a really big job, which I didn't have time to commit to at the moment

edgarriba:
[...]

PR #3088, opened 2024-12-08
chore (docs): enable some Ruff rules for Docstring

ntjohnson1:
#### Changes
related to https://github.com/kornia/kornia/issues/3083

Documentation is a little subjective. I mostly just applied formatting changes/fixes. I made minimal if any content changes.

I broke down the rules by commit to aid in review, but this is still probably a pain to look over. Based on the discussion I initially applied numpy docstyle then found some issues with Ruffs support of it for making it match the description AND for enforcing some nice to have lints. I left the full history here and reverted that commits then applied google doc style. Could go either way without too much headache. Applying NO style leads to warnings because there are 
 pydocstyle rules that conflict.

I like incremental progress so for the variety of rules that require adding content I applied noqa. That way moving forward people will get errors if they don't document and follow on work can be to remove the noqa. If opposed to that can revert those final commits and just leave those rules as TODO.

#### Type of change

- [x] 📚  Documentation Update
- [ ] 🧪 Tests Cases

#### Checklist


docs/generate_examples.py
```diff
@@ -32,7 +32,7 @@ def download_tutorials_examples(download_infos: dict[str, str], directory: Path)
             fp.write(response)
 
 
-def read_img_from_url(url: str, resize_to: Optional[tuple[int, int]] = None, **resize_kwargs) -> torch.Tensor:
+def read_img_from_url(url: str, resize_to: Optional[tuple[int, int]] = None, **resize_kwargs) -> torch.Tensor:  # noqa: D103
```
johnnv1, line 35: we should add to ignore all doc rules for docs/ and tests/ files

kornia/augmentation/_2d/geometric/base.py
```diff
@@ -208,7 +208,7 @@ def inverse_boxes(
 
         return output
 
-    def inverse_keypoints(
+    def inverse_keypoints(  # noqa: D417
```
edgarriba, line 211: why needed here ? there's documentation

kornia/augmentation/auto/autoaugment/autoaugment.py
```diff
@@ -137,10 +137,10 @@ def __init__(
         selection_weights = tensor([1.0 / len(self)] * len(self))
         self.rand_selector = Categorical(selection_weights)
 
-    def compose_subpolicy_sequential(self, subpolicy: SUBPOLICY_CONFIG) -> PolicySequential:
+    def compose_subpolicy_sequential(self, subpolicy: SUBPOLICY_CONFIG) -> PolicySequential:  # noqa: D102
```
edgarriba, line 140: i feel that the fix could a just a little short one line doc

edgarriba: same for the rest (at least the ones that makes sense)

kornia/augmentation/container/augment.py
```diff
@@ -470,14 +470,14 @@ def forward(  # type: ignore[override]
 
         return outputs
 
-    def __call__(
+    def __call__(  # noqa: D417
```
edgarriba, line 473: here there's also docu

edgarriba (commented):
I would spend a little bit of time to add the actual missing documentation

kornia/augmentation/_2d/geometric/base.py
ntjohnson1: D417 checks if all arguments are documented. The kwargs aren't documented. (Verified adding a kwargs entry clears it)

kornia/augmentation/container/augment.py
ntjohnson1: Same note that some but not all arguments are documented.

ntjohnson1:
> I would spend a little bit of time to add the actual missing documentation

My preference would be to add the rules that require content changes back to the ignore list and track them on the issue. I haven't reviewed all the locations that are failing but I expect it to be a fairly significant time investment to properly address all of them (I'm at least not familiar enough with this code base to do it quickly).

These are the counts for the rules I blanket applied noqa to.
D417: 145
D103: 131
D102: 519
D101: 151

ntjohnson1:
Reverted the commits that add the large amount of noqa and can track those on the issue as remaining things to fix by adding the documentation.

D205 should be the only source of noqa in this PR now IIRC, there are only about a dozen D205 noqa's where I didn't simplify the short summary to fit on a single line. Could probably track those down if that's a concern and everything else looks aligned.

johnnv1 (approved):

pyproject.toml
```diff
@@ -138,8 +138,18 @@ ignore = [
...
+  'D102',    # TODO: Undocumented public method
+  'D103',    # TODO: Undocumented public function
+  'D104',    # TODO: Undocumented public package
+  'D105',    # Allow Undocumented magic method
+  'D107',    # TODO: Undocumented public init
+  'D417',    # TODO: Undocumented params
+ ]
+
+[tool.ruff.lint.pydocstyle]
+convention = "google"
```
johnnv1, line 152: could you share some examples of why use Google here instead of numpy? 

I have been using numpy with ruff into other projects without issues

ntjohnson1: I have no strong opinions about Numpy style with ruff globally (I also prefer numpy docstyle generally). For this PR I think there were two arguments for it:

1. The current format is much closer to google style (args instead of parameters, no underlines etc), so reducing the scope of the PR to be enabling/fixing the rules rather than converting docstring style.
  - When I set the numpy doc string style it didn't auto-convert Args: to Parameters with the underlines. I was going to do a find replace to resolve that but fell down the rabbit hole of non-compliant style details (some people have used `Inputs` or `Arg` or other things, so hard to say that the style after the refactor would be consistent without manually skimming everything)

2. https://github.com/astral-sh/ruff/issues/2310 D417 doesn't work for numpy style AFAICT


Example for 2:
```python
from typing import Any


# No error
def google_all_args_documented(arg_1: Any, arg_2: Any):
    """Calculate something.

    Args:
        arg_1: first arg
        arg_2: second arg

    Returns:
        Nothing.
    """
    pass


# D417
def google_some_args_documented(arg_1: Any, arg_2: Any):  # noqa: D417
    """Calculate something.

    Args:
        arg_1: first arg

    Returns:
        Nothing.
    """
    pass


# No error
def numpy_some_args_documented(arg_1: Any, arg_2: Any):
    """Calculate something.

    Parameters
    ----------
        arg_1: first arg

    Returns:
    -------
        Nothing.
    """
    pass
```

johnnv1: i see, tks

PR #3094, opened 2024-12-18
chore (pre-commit): hook to add license header

johnnv1:
i added the hook https://github.com/arkinmodi/add-license-header to run on python files using the Apache header copyright. 

The pre-commit CI auto fixed the files addind this header for all .py files.

cc @ducha-aiki @edgarriba check if it's fine the header as it's or not. We can configure it directly into the pre-commit config


johnnv1:
pre-commit.ci autofix

PR #3096, opened 2024-12-19
Nick/ruff d103

ntjohnson1:
#### Changes

Relates to #3083
(I also fixed a clamp function that always returned min, but it appears unused in the code base)

This is the first rule that required adding new (rather than editing existing) documentation.
I added minimal notes (often trivial) to satisfy the minimum criteria for D103.
As I mentioned when I enabled pydocstyle it checks if `Args` exists that the documentation covers all arguments but it doesn't REQUIRE `Args` to always exist, in which case it doesn't care if none of the args are documented. Thoughts on whether you need/want all the args to also be documented here? Alternatively, could wait (or find if they already exist) for rulesets that enforce all args documented in public methods/functions and resolve this gap then.

#### Type of change

- [x] 📚  Documentation Update

#### Checklist

### DIFF
--- a/kornia/feature/lightglue.py
+++ b/kornia/feature/lightglue.py
@@ -39,11 +39,14 @@ else:
 
 
 def math_clamp(x, min_, max_):  # type: ignore
+    # FIXME this is wrong
+    """Clamp a value to lie within [min, max]."""
     return max(min(x, min_), min_)
 
 
 @custom_fwd(cast_inputs=torch.float32)
 def normalize_keypoints(kpts: Tensor, size: Tensor) -> Tensor:
+    """Normalize tensor of keypoints."""
     if isinstance(size, torch.Size):
         size = torch.tensor(size)[None]
     shift = size.float().to(kpts) / 2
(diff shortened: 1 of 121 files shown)


kornia/feature/lightglue.py
```diff
@@ -39,11 +39,13 @@
 
 
 def math_clamp(x, min_, max_):  # type: ignore
-    return max(min(x, min_), min_)
+    """Clamp a value to lie within [min, max]."""
```
johnnv1, line 42: i think function like these one, shouldn't be a public one, so isn't required docstrings... we shoudl adopt over the library the use of "private" functions, instead of have to document everthing for the newones 

Most of these aux functions for models, or bigger piece of code, shouldn't be public for kornia community, if the community are using, we should make it public and then take more care over it -- since it's kinda hard re-review everthing now, i advocate for us do it in the newer code, as well, when activating this doc rules... 

but because of it being a break change idk... maybe we should only do it for the methods and classes of backbones of some model we have, and for the new codes... cc @edgarriba

johnnv1 (approved):
pretty useful patch, thank you for this one 🚀

kornia/feature/lightglue.py
edgarriba: make total sense. How can we enforce in the ci to not check certain methods being `def _private` or coming from a `class _Private` ?

ntjohnson1: The D10* rules only mention public methods/functions so any private ones shouldn't be enforced.

edgarriba (approved):

PR #3147, opened 2025-03-25
Fix dedode when calling detect() with padding correctly handled

hit2sjtu:
#### Changes

The image into dedode detector is padded to be divisible by 14 by default. However, the output keypoints locations did not handle the padding if the user call detect() only, causing unwanted offset. The forward() did handle the padding. This means the users have to be careful and handle the padding themself if they use detect() only, for example only using the dedode detectors with other matchers such as RoMA or HardNet descriptors. This is pretty error prone if the input image cannot be divided by 14. 

The padding part in detect(), which H,W is image after padding:
https://github.com/kornia/kornia/blob/57296883b1500e87c2a5c92ad76d3708f65dac8b/kornia/feature/dedode/dedode.py#L145-L159

And if one sets the pad_if_not_divisible to false in forward(), the h,w will be assigned which raises an error and crash in L113
https://github.com/kornia/kornia/blob/57296883b1500e87c2a5c92ad76d3708f65dac8b/kornia/feature/dedode/dedode.py#L105-L113

No.
Fixes # (issue)

#### Type of change

- [x]  🐞 Bug fix (non-breaking change which fixes an issue)

#### Checklist

### DIFF
--- a/kornia/feature/dedode/dedode.py
+++ b/kornia/feature/dedode/dedode.py
@@ -84,7 +84,7 @@ class DeDoDe(Module):
         images: Tensor,
         n: Optional[int] = 10_000,
         apply_imagenet_normalization: bool = True,
-        pad_if_not_divisible: bool = True,
+        pad_if_not_divisible: bool = False,
     ) -> Tuple[Tensor, Tensor, Tensor]:
         """Detect and describe keypoints in the input images.
 
@@ -120,7 +120,7 @@ class DeDoDe(Module):
         images: Tensor,
         n: Optional[int] = 10_000,
         apply_imagenet_normalization: bool = True,
-        pad_if_not_divisible: bool = True,
+        pad_if_not_divisible: bool = False,
         crop_h: Optional[int] = None,
         crop_w: Optional[int] = None,
     ) -> Tuple[Tensor, Tensor]:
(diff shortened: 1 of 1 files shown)


hit2sjtu:
@ducha-aiki your review is appreciated.

hit2sjtu:
> @hit2sjtu thank you for the catch! Could you please instead of turning off, fix the behavior?

I guess it makes more sense to fix the behaviour with padding.  After double check, the dedode-G is just descriptor, which I have misunderstand. I have updated my intial issue description and uploaded the fix.

kornia/feature/dedode/dedode.py
```diff
@@ -142,15 +142,17 @@ def detect(
         """
         KORNIA_CHECK_SHAPE(images, ["B", "3", "H", "W"])
         self.train(False)
+        B, C, H, W = images.shape
```
hit2sjtu, line 145: Initiallay H W inlcudes the padding area (called after padding), which might be different from input image. The returned keypoints are normalized with regard to the padded image height/width. We should just keep the incoming image size and remove padded region from logits.

ducha-aiki (approved):
LGTM, thank you for doing this!

PR #3179, opened 2025-06-11
Fixes in RTDETR example in rt_detr.rst

mgrewe:
Import fixed, drawing replaced with simple print out.


mgrewe:
Won't fix due to broken build tools.

PR #3182, opened 2025-06-25
[Feat] Init MCP module

shijianjian:
Sample usage.

```
import asyncio

from kornia.mcp import create_kornia_mcp_server
from kornia.io import load_image

server = create_kornia_mcp_server()

async def main():
    result = await server.call_tool("kornia.enhance.add_weighted", {
        "src1_path": "panda.jpg",
        "src2_path": "panda.jpg",
        "alpha": 0.5,
        "beta": 0.5,
        "gamma": 0.0
    })
    print(eval(result[0].text)["output"])


if __name__ == "__main__":
    asyncio.run(main())
```


shijianjian:
Will put them to a new lib.

PR #3218, opened 2025-08-03
optimized code to calculate squared and Euclidean distance

sidd-27:
used the ||d - (d·u)u||^2 = ||d||^2 - (d·u)^2 identity to optimize distance calculation
The optimized version is faster because it avoids constructing and subtracting full projection vectors for each point, which would require intermediate tensors of shape (N, D). Instead, it relies on a scalar identity. This reduces both memory usage and computation by using only dot products and scalar operations, resulting in fewer tensor allocations and better cache performance

#### Type of change
- [x] code optimization

#### Checklist

### DIFF
--- a/kornia/geometry/line.py
+++ b/kornia/geometry/line.py
@@ -129,24 +129,28 @@ class ParametrizedLine(Module):
 
     # TODO: improve order and speed
     def squared_distance(self, point: Tensor) -> Tensor:
-        """Return the squared distance of a point to its projection onte the line.
-
-        Args:
-            point: the point to calculate the distance onto the line.
-
         """
-        diff: Tensor = point - self.origin
-        return squared_norm(diff - (self.direction @ diff) * self.direction)
+        Fast squared distance from `point` to this line.
+        Uses: ||d - (d·u)u||^2 = ||d||^2 - (d·u)^2.
+        """
+        # d: (..., D)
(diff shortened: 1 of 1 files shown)


kornia/geometry/line.py
```diff
@@ -127,24 +127,20 @@ def projection(self, point: Tensor) -> Tensor:
         """
         return self.origin + (self.direction @ (point - self.origin)) * self.direction
 
-    # TODO: improve order and speed
     def squared_distance(self, point: Tensor) -> Tensor:
-        """Return the squared distance of a point to its projection onte the line.
-
-        Args:
```
edgarriba, line 134: keep this

kornia/geometry/line.py
```diff
@@ -127,24 +127,20 @@ def projection(self, point: Tensor) -> Tensor:
...
+        d = point - self.origin
+        proj = torch.sum(d * self.direction, dim=-1)
+        sq_norm_d = torch.sum(d * d, dim=-1)
+        return sq_norm_d - proj * proj
 
-    # TODO: improve order and speed
     def distance(self, point: Tensor) -> Tensor:
-        """Return the distance of a point to its projections onto the line.
-
-        Args:
```
edgarriba, line 145: keep this

cjpurackal (approved):

PR #3219, opened 2025-08-03
optimized line fit for 2d case by upto 2665.96% by using closed form solution

sidd-27:
Implemented the 2d case for the fit line using the standard OLS algorithm, for structured data it works perfectly as the original svd implementation, but in cases where the data is a random point cloud it may differ but that makes sense as any possible answer would be technically correct in those cases.

=== Benchmark: Weighted Case ===
Weighted OLS time: 1.06 ms
SVD time:          21.48 ms
Speedup:           **2019.57%**
Cosine similarity mean: 1.000000
Max angular error (deg): 0.034264
Mean angular error (deg): 0.000459
Number of batches with >5° error: 0

=== Benchmark: Unweighted Case ===
Unweighted OLS time: 0.62 ms
SVD time:            16.53 ms
Speedup:             **2665.96%**
Cosine similarity mean: 1.000000
Max angular error (deg): 0.019782
Mean angular error (deg): 0.000320
Number of batches with >5° error: 0

https://colab.research.google.com/drive/1pDMN_xZzb6GVmA7u0mPR87eo-FY2AFL6?usp=sharing

#### Checklist

### DIFF
--- a/kornia/geometry/line.py
+++ b/kornia/geometry/line.py
@@ -226,7 +226,9 @@ def _fit_line_weighted_ols_2d(points: Tensor, weights: Tensor) -> ParametrizedLi
     # direction = normalize([1, slope]) or [0,1] if vertical
     is_vertical = denom <= 1e-8
     direction = torch.cat([torch.ones_like(slope), slope], dim=-1)  # (B, 2)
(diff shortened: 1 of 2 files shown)


sidd-27:
The tests already written do cover the case for D = 2, but if needed I'll add more tests

sidd-27:
ill work on the weighted case later, probably some issues with the test

PR #3220, opened 2025-08-03
compeleted todo: optimized _distort_points_kannala_brandt

sidd-27:
optimized by also passing in the value for radius_sq, which avoids an unnecessary computation of the same in the impl function

#### Checklist


sidd-27:
accidentaly added the local code i had written to test the masking to do as well, but removed it as it ran slower in benchmarks

PR #3224, opened 2025-08-06
optimized mod function by using inbuilt torch.remainder by 336% ?? (not sure pls check)

sidd-27:
Running benchmark on CUDA
✅ custom_mod == torch.remainder: PASS
✅ custom_mod == a % b         : PASS

⏱ Average time per call (over 100 runs):
custom_mod       : 38.88 µs
torch.remainder  : 11.55 µs
a % b            : 11.73 µs

⚡ Speedup over custom:
torch.remainder  : 3.37x
a % b            : 3.31x

ngl I don't really understand why this works so well in the benchmarks, even CodeFlash said no optimization was possible, but I wrote a benchmark script just for fun, and these were the results. if anyone knows what's happening and if my benchmarking code is wrong, please do explain

### DIFF
--- a/kornia/contrib/models/rt_detr/post_processor.py
+++ b/kornia/contrib/models/rt_detr/post_processor.py
@@ -45,7 +45,7 @@ def mod(a: Tensor, b: int) -> Tensor:
         1
 
     """
-    return a - (a // b) * b
+    return torch.remainder(a, b)
 
 
 # TODO: deprecate the confidence threshold and add the num_top_queries as a parameter and num_classes as a parameter


sidd-27:
it seems to only be failing on older torch version

sidd-27:
why... it was failing in versions 2.0.1, so I reverted to the original statement there why is it still giving an error if its using the older code

edgarriba (approved):

kornia/contrib/models/rt_detr/post_processor.py
```diff
@@ -45,7 +45,9 @@ def mod(a: Tensor, b: int) -> Tensor:
         1
 
     """
-    return a - (a // b) * b
+    if not a.dtype.is_floating_point and a.dtype != torch.long:
```
cjpurackal, line 48: I was going to ask you to add a comment on why this works here for future reference and then i read your PR comment :). Maybe just add your observation here as a comment.

sidd-27: ive made a new pull request with some new changes pls check @cjpurackal

PR #3228, opened 2025-08-06
optimized normalize() with inbuilt function to get 55.3% speedup

sidd-27:
used torch function for normalization and also wrote through tests for it

method    time (µs)    % improv.
--------------------------------
  orig      1231.08             —
 fnorm       792.92        35.59%

Accuracy check (max abs error vs. orig):
  fnorm : 5.960e-08

https://colab.research.google.com/drive/1ey7hEol_dpGYC6ap3-qnhygKylsEh7V8?usp=sharing
benchmarking script

#### Checklist

### DIFF
--- a/kornia/geometry/plane.py
+++ b/kornia/geometry/plane.py
@@ -19,6 +19,7 @@
 # https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Geometry/Hyperplane.h
 
 from typing import Optional
+
 import torch.nn.functional as F
 
 from kornia.core import Module, Tensor, stack, where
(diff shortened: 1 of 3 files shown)


kornia/geometry/plane.py
```diff
@@ -31,7 +33,7 @@
 
 
 def normalized(v: Tensor, eps: float = 1e-6) -> Tensor:
-    return v / batched_dot_product(v, v).add(eps).sqrt()
+    return F.normalize(v, p=2, dim=-1, eps=eps)
```
shijianjian: We would avoid using torch functions, since we have also functionalities to port the code to numpy, etc.

sidd-27:
this also provides an optimization over the previous function, up to 20% in some cases. but the variance is too much so I assume the average improvement is about 6-7%, barely over the threshold @shijianjian 

def normalized(v: torch.Tensor, eps: float = 1e-6) -> torch.Tensor:
    norm_sq = (v * v).sum(dim=-1, keepdim=True) + eps
    return v * norm_sq.rsqrt()

pasted this in the original benchmarking code to get these results

PR #3231, opened 2025-08-07
optimized needleman_wunsch by 1772.46%

sidd-27:
used a wavefront order (iterate over anti-diagonals) to iterate through the dp table instead of serial order, leading to more efficient tensor ops. The actual algorithm is exactly the same, so the outputs don't differ

Benchmark Results:
Naive     : Time = 3.183107 s, Speedup = 1.00x,    Matches Naive = True
Wavefront : Time = 0.179587 s, Speedup = 1772.46%, Matches Naive = True

https://colab.research.google.com/drive/1aExLMZbRayVwGyf5nJhuUFfXsVSk5OWg?usp=sharing
here's the benchmark colab file

### DIFF
--- a/kornia/feature/sold2/sold2.py
+++ b/kornia/feature/sold2/sold2.py
@@ -317,16 +317,16 @@ class WunschLineMatcher(Module):
         # Recalibrate the scores to get a gap score of 0
         gap = 0.1
         B, N, M = scores.shape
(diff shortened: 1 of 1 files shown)


edgarriba (commented):
check why tests fail

sidd-27:
i think its something with the other pr's merged recently, I don't think my changes did anything

edgarriba (approved):

PR #3234, opened 2025-08-07
optimized euclidean_distance by 30% with more efficient tensor ops

sidd-27:
The optimized version is faster because it avoids unnecessary elementwise addition of eps, applying it only once after summing squared differences. It reduces memory usage by eliminating an extra intermediate tensor created by (x - y + eps). Overall, it performs fewer operations and minimizes broadcasting, leading to better performance.

https://colab.research.google.com/drive/1GKDUFrIi6Kj2ayjoUUx71Xz5_6cy5vSS?usp=sharing

here's code to benchmark as well as validate the changes made

### DIFF
--- a/kornia/geometry/linalg.py
+++ b/kornia/geometry/linalg.py
@@ -282,7 +282,7 @@ def euclidean_distance(x: Tensor, y: Tensor, keepdim: bool = False, eps: float =
     KORNIA_CHECK_SHAPE(x, ["*", "N"])
     KORNIA_CHECK_SHAPE(y, ["*", "N"])
 
-    return (x - y + eps).pow(2).sum(-1, keepdim).sqrt()
+    return torch.sqrt(((x - y).pow(2)).sum(dim=-1, keepdim=keepdim) + eps)
 
 
 # aliases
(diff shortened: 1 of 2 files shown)


sidd-27:
theres also more efficient inbuilt tensorflow functions (F.pairwise_distance) but since jin said to avoid them I've stuck to this, if that can be used lemme know

edgarriba:
> sorflow functions (F.pairwise_dis

this would break the non-torch backends I think. /cc @shijianjian

kornia/geometry/linalg.py
```diff
@@ -282,7 +282,7 @@ def euclidean_distance(x: Tensor, y: Tensor, keepdim: bool = False, eps: float =
     KORNIA_CHECK_SHAPE(x, ["*", "N"])
     KORNIA_CHECK_SHAPE(y, ["*", "N"])
 
-    return (x - y + eps).pow(2).sum(-1, keepdim).sqrt()
+    return torch.sqrt(((x - y).pow(2)).sum(dim=-1, keepdim=keepdim) + eps)
```
edgarriba:
```suggestion
    return (x - y).pow(2).sum(-1, keepdim).sqrt() + eps
```

not the same ?

sidd-27: youre right, making changes, it should still be similarly faster

edgarriba (approved):

PR #3237, opened 2025-08-08
optimized otsu threshold forward by 308.5x (30850% speedup)

sidd-27:
vectorized the nested loops for more efficient code

https://colab.research.google.com/drive/1kz0qscYUCX30UooM6BOP2F2OjDLetLaH?usp=drive_link

Using device: cuda

--- Benchmarking Results ---
Warming up old function...
Old Iterative Method: 1.717000 seconds per run

Warming up new function...
New Vectorized Method: 0.005566 seconds per run

--- Speedup Analysis ---
Speedup: 308.50x

--- Validity Check ---
Thresholds from old and new methods match: True
Thresholded images from old and new methods match: True

### DIFF
--- a/kornia/filters/otsu_thresholding.py
+++ b/kornia/filters/otsu_thresholding.py
@@ -113,78 +113,64 @@ class OtsuThreshold(torch.nn.Module):
         Returns:
             Tuple[torch.Tensor, torch.Tensor]: Thresholded tensor, threshold values.
         """
+        # Flatten input and store original shape
         x_flattened, orig_shape = self.transform_input(x)
         nchannel = x_flattened.shape[0]
 
         # Check tensor type compatibility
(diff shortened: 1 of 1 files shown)


kornia/filters/otsu_thresholding.py
```diff
@@ -113,78 +112,66 @@ def forward(
...
+            torch.int8,
+            torch.int16,
+            torch.int32,
+            torch.int64,
+            torch.float16,
+            torch.float32,
+            torch.float64,
+            torch.bfloat16,
+        ]
+        if x.dtype not in supported_dtypes:
```
ducha-aiki: Please use KORNIA_CHECK instead

ducha-aiki (approved):

ducha-aiki (approved):

PR #3238, opened 2025-08-08
fixed docstring issue

sidd-27:
also since remainder works with floats too and the errors weren't caused by the mod function I've removed the unnecessary check

### DIFF
--- a/kornia/contrib/models/rt_detr/post_processor.py
+++ b/kornia/contrib/models/rt_detr/post_processor.py
@@ -19,7 +19,6 @@
 
 from __future__ import annotations
 
-import numbers
 from typing import Optional, Union
 
 import torch


sidd-27:
i dont understand, pls help

    if isinstance(a, torch.Tensor):
        # torch.remainder supports both integer and floating dtypes
        return torch.remainder(a, b)
    else:
        # Use Python modulo for Python scalars (keeps scalar result and semantics)
        return a % b

this is what the final function should be

kornia/contrib/models/rt_detr/post_processor.py
```diff
@@ -28,24 +29,32 @@
...
+        >>> mod(torch.tensor([7, -1, 2]), 3)
+        tensor([1, 2, 2])
     """
-    return a - (a // b) * b
+    if isinstance(a, torch.Tensor):
+        # torch.remainder supports both integer and floating dtypes
+        return torch.remainder(a, b)
+    else:
+        # Use Python modulo for Python scalars (keeps scalar result and semantics)
+        return a % b
```
edgarriba: doesn't torch support this already ?

sidd-27: from what i can tell the issue was that doc test was getting an error as the input in the docstring wasn't an torch tensor, so I added a case for scalars too

edgarriba: Image

PR #3243, opened 2025-08-09
optimized load_pointcloud_ply by 27.91x (2791% speedup)

sidd-27:
optimized load_pointcloud_ply by using numpy load text function to avoid iterating through all points individually

https://colab.research.google.com/drive/1muDCsMq-Kt1b5SLYs3XK6tlz9nnarEyM?usp=sharing
here's a benchmark script to test the validity of the output as well as the speedup provided

--- Testing with 10000 points ---
Validation passed ✅
Original:  0.281743 sec
Optimized: 0.012260 sec
Speedup:   22.98×

--- Testing with 100000 points ---
Validation passed ✅
Original:  3.367795 sec
Optimized: 0.120685 sec
Speedup:   27.91×

### DIFF
--- a/kornia/utils/pointcloud_io.py
+++ b/kornia/utils/pointcloud_io.py
@@ -16,7 +16,7 @@
 #
 
 import os
-
+import numpy as np
 import torch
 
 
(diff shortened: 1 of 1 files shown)


kornia/utils/pointcloud_io.py
edgarriba, line 1: I don't think here we want to add numpy but keep using python builtin deps

sidd-27:
i think these test errors were caused by accelerate, should be fixed by now

PR #3244, opened 2025-08-09
optimized save_pointcloud_ply by 16.84x (1684%)

sidd-27:
also wrote tests for the same, using more efficient numpy functions and removing unnecessary loops

--- Testing with 5000 points ---
Validation passed ✅
Original:  0.193949 sec
Optimized: 0.017146 sec
Speedup:   11.31×

--- Testing with 10000 points ---
Validation passed ✅
Original:  0.394249 sec
Optimized: 0.027022 sec
Speedup:   14.59×

--- Testing with 50000 points ---
Validation passed ✅
Original:  2.305624 sec
Optimized: 0.136922 sec
Speedup:   16.84×

https://colab.research.google.com/drive/1mYSBLYODzHbxlZtXH1Ygi92c5ibIb55j?usp=sharing

here's benchmarking and validadation code

### DIFF
--- a/kornia/utils/pointcloud_io.py
+++ b/kornia/utils/pointcloud_io.py
@@ -16,6 +16,7 @@
 #
 
 import os
+
 import torch
 
 
--- a/kornia/utils/pointcloud_io.py
+++ b/kornia/utils/pointcloud_io.py
@@ -17,8 +17,8 @@
 
 import os
 
-import torch
 import numpy as np
+import torch
 
 
 def save_pointcloud_ply(filename: str, pointcloud: torch.Tensor) -> None:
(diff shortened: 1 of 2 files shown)


kornia/utils/pointcloud_io.py
```diff
@@ -17,6 +17,7 @@
 
 import os
 
+import numpy as np
```
edgarriba: can we not use numpy for this ?

tests/utils/test_pointcloud_io.py
```diff
@@ -17,54 +17,84 @@
...
 import os
 
+import numpy as np
+import pytest
 import torch
 
 import kornia
 
-from testing.base import assert_close
+from testing.base import assert_close  # assuming this is your custom assert_close
```
edgarriba:
```suggestion
from testing.base import assert_close
```

PR #3245, opened 2025-08-09
optimized image_list_to_tensor by 2.92x (292% speedup)

sidd-27:
Using np.stack is faster than appending tensors in a loop because it allocates memory once for the entire output array, while appending requires repeated memory reallocations and data copying. This difference significantly impacts performance, especially with large datasets.

✅ Validation passed: Outputs are identical.
Original function time (10 runs): 0.1937 seconds
Optimized function time (10 runs): 0.0663 seconds
Speedup: 2.92x

https://colab.research.google.com/drive/1MWw7Y54W_rufj647yd771SsNU23c-J29?usp=sharing

here's benchmark and validation script for the same

### DIFF
--- a/kornia/utils/image.py
+++ b/kornia/utils/image.py
@@ -17,8 +17,6 @@
 
 from functools import wraps
 from typing import Any, Callable, List, Optional
-
-import numpy as np
 import torch
 from torch import nn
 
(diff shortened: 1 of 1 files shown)


kornia/utils/image.py
```diff
@@ -18,6 +18,7 @@
 from functools import wraps
 from typing import Any, Callable, List, Optional
 
+import numpy as np
```
edgarriba: should we use numpy here ? @shijianjian

PR #3259, opened 2025-08-10
removed unnecessary clamping by 59.0% (2.4x speedup)

sidd-27:
the clamping is unnecessary for all values of layer index and conf layers which are positive, as it does nothing

Benchmark Results (10,000 calls per layer):
Original avg: 0.005487 sec
Optimized avg: 0.002250 sec
Improvement: 59.0%
Speedup = 2.4x

 https://colab.research.google.com/drive/1kO3aJVg6sccAJ040003Ylk42Mnq67RJu?usp=sharing


edgarriba (changes requested):
this is not correct

sidd-27:
You're right mb 😭

PR #3262, opened 2025-08-10
optimized to_box by 6.33x (630%) by adding a gpu friendly path 

sidd-27:
kept the cpu version as is as it is more optimal for a CPU, but added a vectorization friendly version which runs about 6 times as fast

✅ Validation passed — max abs diff: 0.0
Old time: 51.924 ms
New time: 8.204 ms
Speedup: 6.33x

https://colab.research.google.com/drive/1Vj2oLKRul874VZTGCY8lHgO6GSlEI2Ux?usp=sharing

here's a colab file to benchmark and validate the code

### DIFF
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -589,25 +589,62 @@ class Boxes:
                 "Boxes.to_tensor isn't differentiable. Please, create boxes from tensors with `requires_grad=False`."
             )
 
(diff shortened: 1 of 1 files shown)


sidd-27:
I don't really have access to a GPU apart from Colab, so if possible, please do test out the pr on a GPU before merging. i feel that the benchmarking script is good enough but just in case, as none of the github tests are on cuda ,

PR #3266, opened 2025-08-13
optimized get_spatial_gradients by 2x on cpu

sidd-27:
The speedup in the CPU case is significant, but is it worth adding it if it's mostly run on GPU?

Original function average time:   0.239940 seconds
Optimized function average time:  0.108772 seconds
--------------------
📈 Speedup Factor: 2.21x
(The optimized version is 2.21 times faster)

https://colab.research.google.com/drive/1DzmpICYS5QJd09ZMMiKecFN65DIv_DMp?usp=sharing

### DIFF
--- a/kornia/filters/sobel.py
+++ b/kornia/filters/sobel.py
@@ -57,20 +57,33 @@ def spatial_gradient(input: Tensor, mode: str = "sobel", order: int = 1, normali
     KORNIA_CHECK_SHAPE(input, ["B", "C", "H", "W"])
 
     # allocate kernel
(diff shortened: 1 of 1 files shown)


sidd-27:
the optimization here was the grouped convolution, i added a seperate path for gpu as the optimization really only worked for cpu

sidd-27:
The difference is in the final conv2d, cpu uses groups = c while gpu uses groups = 1

sidd-27:
I don't use codeflash ... the benchmarks  were on colab GPUs, but yes I'll try to make more comprehensive benchmarks to ensure that my pr works

edgarriba:

Issue #3160, opened 2025-04-16
to make it even faster we can use torch.aminmax in KORNIA_CHECK_IS_IMAGE


edgarriba:
to make it even faster we can use torch.aminmax - https://pytorch.org/docs/stable/generated/torch.aminmax.html

_Originally posted by @misrasaurabh1 in https://github.com/kornia/kornia/pull/3156#discussion_r2025356386_

PR #3270, opened 2025-08-18
optimized check image using aminmax

sidd-27:
https://github.com/kornia/kornia/issues/3160 did as per the issue suggested, about a 1.8x speedup for the function, also added relevant tests for the check_image function


sidd-27:
this same optimization could be made for a couple more functions in the library, just searching for min() gives a couple examples, i have exams rn so if anyone else wants they could do those optimizations

nvm aminmax isnt diffrentiable yet for some reason, theyll probably add it in the next totch version hopefully

PR #3272, opened 2025-08-21
optimized a couple loss functions

sidd-27:
One-hot encoding was removed by using gather, which directly selects the true-class probability without expanding to all classes. also in the other functions used more optimal variable handling and pytorch functions, optimization of 4x in some cases

https://colab.research.google.com/drive/1DusGb0krES6hIHKfXAJwv_zzyJlSgfkb?usp=sharing

heres the code which benchmarks and tests these changes

### DIFF
--- a/kornia/losses/cauchy.py
+++ b/kornia/losses/cauchy.py
@@ -17,10 +17,10 @@
 
 from __future__ import annotations
 
+import torch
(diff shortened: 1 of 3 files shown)


cjpurackal:
can you please add the names of the loss functions in the PR ?

Issue #2924, opened 2024-06-07
NaN values returned during backward pass in axis_angle_to_rotation_matrix function


Abermal:
### Describe the bug

Error:
```python
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: Function 'SqrtBackward0' returned nan values in its 0th output.
```

The error happens here:

```python
    def _compute_rotation_matrix(axis_angle: torch.torch.Tensor, theta2: torch.torch.Tensor, eps: float = 1e-6) -> torch.torch.Tensor:
        # We want to be careful to only evaluate the square root if the
        # norm of the axis_angle vector is greater than zero. Otherwise
        # we get a division by zero.
        k_one = 1.0
        theta = torch.sqrt(theta2) 
        wxyz = axis_angle / (theta + eps)
        wx, wy, wz = torch.chunk(wxyz, 3, dim=1)
        cos_theta = torch.cos(theta)
        sin_theta = torch.sin(theta)

        r00 = cos_theta + wx * wx * (k_one - cos_theta)
        r10 = wz * sin_theta + wx * wy * (k_one - cos_theta)
        r20 = -wy * sin_theta + wx * wz * (k_one - cos_theta)
        r01 = wx * wy * (k_one - cos_theta) - wz * sin_theta
        r11 = cos_theta + wy * wy * (k_one - cos_theta)
        r21 = wx * sin_theta + wy * wz * (k_one - cos_theta)
        r02 = wy * sin_theta + wx * wz * (k_one - cos_theta)
        r12 = -wx * sin_theta + wy * wz * (k_one - cos_theta)
        r22 = cos_theta + wz * wz * (k_one - cos_theta)
        rotation_matrix = torch.cat([r00, r01, r02, r10, r11, r12, r20, r21, r22], dim=1)
        return rotation_matrix.view(-1, 3, 3)
  ```
  The problem is cause by the line `theta = torch.sqrt(theta2)`.
  
The torch.sqrt function is returning NaN (Not a Number) values during the backward pass. This is likely due to the fact that the input to torch.sqrt is negative.

In your axis_angle_to_rotation_matrix function, the theta2 tensor is computed as the dot product of _axis_angle and its transpose. If the input axis_angle tensor has a norm of zero, then theta2 will be zero, and torch.sqrt(theta2) will return NaN.

To fix this issue, you can add a small epsilon value to theta2 before taking the square root, like this:

[...]

PR #3276, opened 2025-08-24
fixed bug and optimized axis_angle_to_rotation_matrix() by 1.5x

sidd-27:
https://github.com/kornia/kornia/issues/2924 fixed the bug as described in this issue, also added tests to test the function and this bug specifically (the original code fails those tests), also optimized the function by 1.5x by avoiding recalculations and more efficient tensor ops

https://colab.research.google.com/drive/1wGiyX0oy9ySXpPlAjO5hlHIhJ4BWGdHL?usp=sharing

this is the benchmarking + validation script 

Device=cpu: Original=15.551 ms | Optimized=11.405 ms | Speedup=1.36×
Device=cuda: Original=1.456 ms | Optimized=0.968 ms | Speedup=1.50×

### DIFF
--- a/tests/geometry/test_conversions.py
+++ b/tests/geometry/test_conversions.py
@@ -21,7 +21,7 @@ from functools import partial
 import numpy as np
 import pytest
 import torch
-
+import math
 import kornia
 from kornia.geometry.conversions import (
     ARKitQTVecs_to_ColmapQTVecs,
@@ -35,6 +35,7 @@ from kornia.geometry.conversions import (
     matrix4x4_to_Rt,
     quaternion_from_euler,
     worldtocam_to_camtoworld_Rt,
+    axis_angle_to_rotation_matrix,
 )
 from kornia.geometry.quaternion import Quaternion
 from kornia.utils._compat import torch_version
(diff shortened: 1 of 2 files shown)


tests/geometry/test_conversions.py
```diff
@@ -1273,3 +1275,144 @@ def test_vector_to_skew_symmetric_matrix(batch_size, device, dtype):
...
+            try:
+                mn = float(tensor.min().item())
+            except Exception:
+                return orig_sqrt(tensor, *args, **kwargs)
+            calls["min_seen"] = min(calls["min_seen"], mn)
+            if mn < -1e-12:
+                raise AssertionError(f"torch.sqrt called with negative input (min={mn})")
+            return orig_sqrt(tensor, *args, **kwargs)
+
+        monkeypatch.setattr(torch, "sqrt", wrapped_sqrt)
```
edgarriba: this seems too much, tests should be more simple

edgarriba (changes requested):
@cjpurackal

tests/geometry/test_conversions.py
sidd-27: Fair enough, I'll remove the back prop tests

cjpurackal:
Actually these feel a little redundant with the exponential map and `matrix()` of `kornia.geometry.liegroup.so3` https://github.com/kornia/kornia/blob/e779c63d54ac73bbaa13d11587d7a944125ebdd4/kornia/geometry/liegroup/so3.py#L110
https://github.com/kornia/kornia/blob/e779c63d54ac73bbaa13d11587d7a944125ebdd4/kornia/geometry/liegroup/so3.py#L211
Can you look into how these functions can be used in `SO3` ? Perhaps in a different PR.

Issue #3280, opened 2025-08-28
`kornia.io.write_image` cannot save mono images


eflebus:
### Describe the bug

Calling `kornia.io.write_image` with mono/grayscale images (dtype uint8, shape HxW or 1xHxW) raises `TypeError: argument 'image': 'ndarray' object cannot be converted to 'PyArray<T, D>'`.

However, function `kornia_rs.write_image_png_u8` (called by `kornia.io.write_image`) works as expected with uint8 HxWx1 numpy arrays.

After a quick debug session, I would say that the issue is caused by function `kornia.utils.tensor_to_image`. Specifically, if the input torch tensor has shape HxW or HxWx1, then the output numpy array has shape HxW (instead of HxWx1). Line 204 of file `kornia/io/io.py` is reported below:
`img_np = tensor_to_image(image, keepdim=True, force_contiguous=True)  # HxWxC`
This causes `kornia_rs.write_image_png_u8` to raise the error shown above when `mode="mono"`.

Possible fix: add third channel as needed.
```python
...
img_np = tensor_to_image(image, keepdim=True, force_contiguous=True)  # HxWxC
if img_np.ndim == 2:
    img_np = img_np[..., None]  # HxW -> HxWx1
...
```

I looked at the open issues and pull requests but I couldn't find anything similar. I hope this report is detailed and clear enough.

Kudos for this great library!

### Reproduction steps

```bash
>>> import kornia
>>> import kornia_rs
>>> import numpy as np
>>> import torch

>>> image_np_hwc = np.zeros(shape=(300, 300, 1), dtype=np.uint8)
>>> image_np_hw = np.zeros(shape=(300, 300), dtype=np.uint8)
>>> image_torch_chw = torch.zeros(size=(1, 300, 300), dtype=torch.uint8)
>>> image_torch_hw = torch.zeros(size=(300, 300), dtype=torch.uint8)

# HxWx1 numpy array -> ok
>>> kornia_rs.write_image_png_u8("image.png", image_np_hwc, mode="mono")

# HxW numpy array -> error
>>> kornia_rs.write_image_png_u8("image.png", image_np_hw, mode="mono")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: argument 'image': 'ndarray' object cannot be converted to 'PyArray<T, D>'

# 1xHxW torch tensor -> error
>>> kornia.io.write_image("image.png", image_torch_chw)
Traceback (most recent call last):
[...]

PR #3284, opened 2025-09-05
fix(io): ensure grayscale images are saved with channel dim (HxWx1) (#3280)

sanskarmodi8:
### Problem
Calling `kornia.io.write_image` with grayscale images (shape `(H, W)` or `(1, H, W)`, dtype `uint8`) raised the following error:

`TypeError: argument 'image': 'ndarray' object cannot be converted to 'PyArray<T, D>'`

Root cause: `tensor_to_image` returned a 2D array `(H, W)` for grayscale images, 
but `kornia_rs.write_image_png_u8` expects a 3D array `(H, W, C)` (even for mono images).

### Fix
Added a post-processing step in `kornia.io.write_image` to ensure that if the output
array is 2D, it is expanded to 3D by adding a singleton channel dimension:

```python
if img_np.ndim == 2:
    img_np = img_np[..., None]  # (H, W) -> (H, W, 1)
```
This ensures grayscale images are saved correctly.

### DIFF
--- a/kornia/io/io.py
+++ b/kornia/io/io.py
@@ -202,7 +202,8 @@ def write_image(path_file: str | Path, image: Tensor, quality: int = 80) -> None
     KORNIA_CHECK(image.dim() >= 2, f"Invalid image shape: {image.shape}. Must be at least 2D.")
 
     img_np = tensor_to_image(image, keepdim=True, force_contiguous=True)  # HxWxC
-
+    if img_np.ndim == 2:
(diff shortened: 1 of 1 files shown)


sanskarmodi8:
The CI failure in `SOLD2_detector` doctest appears unrelated to this PR and is caused by memory allocation error on CI (42GB allocation attempt).  
All IO-related tests pass locally, confirming that the grayscale saving fix works as expected.

sanskarmodi8:
Are there any updates on this?

PR #3289, opened 2025-09-13
optimized get_kron_order by upto 

sidd-27:
the two nested loops have been replaced by torch.meshgrid to do it all at once efficiently

Benchmarking with dimensions d1=50, d2=50
Original function execution time:  0.037979 seconds
Optimized function execution time: 0.000456 seconds
--------------------------------------------------
🚀 Speedup: The optimized version is 83.31 times faster!

https://colab.research.google.com/drive/1YNZg1RkcEHMqLYhfIqrSE3JmWm0lIH_2?usp=sharing

heres the benchmarking and validation script for the change


weights/checkpoints/mkd-cart-64.pth
edgarriba, line 1: this shouldn't be included, same below

PR #3292, opened 2025-09-15
feature: `requires_grad` in quaternion

### LINKED ISSUE #3291: Quaternion with `requires_grad=False` should propagate `requires_grad=False` to results of math operations
## 🚀 Feature
This actually bundle two suggestions
- Expose a `requires_grad` kwarg in `kornia.geometry.Quaternion` constructor and class methods to allow setting `requires_grad` to `False` (can be `True` by default for backward compatibility).
- Modify the math operations dunder methods (e.g., `__add__`, `__pow__`, ...) such that when an operation is done and all operands have `requires_grad=False`, then the resulting quaternion also have `requires_grad=False`.

## Motivation

- Consistent behavior with torch tensors
- Avoid having to set `requires_grad=False` in the result of every expression

### Current behavior
```python
>>> from kornia.geometry import Quaternion
>>> q = Quaternion.identity()  # no way to set `requires_grad=False`
>>> q
Parameter containing:
tensor([1., 0., 0., 0.], requires_grad=True)
>>> q.requires_grad_(False)
Parameter containing:
tensor([1., 0., 0., 0.])
>>> q
Parameter containing:
tensor([1., 0., 0., 0.])
>>> q**2
Parameter containing:
tensor([1., 0., 0., 0.], requires_grad=True)  # expected this to have `requires_grad=False`
```
### Desired behavior
```python
>>> from kornia.geometry import Quaternion
>>> q = Quaternion.identity(requires_grad=False)  # Can set `requires_grad`
>>> q
Parameter containing:
tensor([1., 0., 0., 0.])
>>> q**2
Parameter containing:
tensor([1., 0., 0., 0.])  # Result also does not `requires_grad`
>>> q * Quaternion.identity()  # the second operand requires grad by default
Parameter containing:
tensor([1., 0., 0., 0.], requires_grad=True)  #   
```

## Alternatives

Suggested solution involves modifying dunder methods of `Quaternion`. I can implement it.

## Additional context

**EDIT**: To be honest, the fact that `Quaternion` is a subclass of `nn.Module` is a bit strange to me as: 1) it doesn't have a forward method and 2) it doesn't build a graph with the quaternion operations (as `Parameter(data)` is being called in the init, causing a detached tensor). But as I'm currently not backpropagating through quaternion operations, this is not an issue for me, but I could suggest subclassing a `torch.Tensor` instead and defining `__torch_function__` for the math operations.

______________________________________________________________________

#### Consider also to contribute to Kornia universe projects :)

<sub>

- [**Tutorials**](https://github.com/kornia/tutorials): our repository containing the tutorials.

</sub>

egidioln:
#### Changes

Modifies the `Quaternion` class so it objects that are results from quaternion operations only `requires_grad` if operands `requires_grad` 

Fixes #3291 

#### Type of change
- [x] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/tests/geometry/test_quaternion.py
+++ b/tests/geometry/test_quaternion.py
@@ -308,7 +308,7 @@ class TestQuaternion:
 
         self.assert_close(euler, euler_expected, 1e-4, 1e-4)
 
-    def test_requires_grad_false_propagates(self, device, dtype, batch_size):
+    def test_requires_grad_false_propagates(self, device, dtype):
         random_coefs = torch.tensor(
             [
                 [4.5794e-02, -7.0831e-01, 6.7577e-01, 1.9883e-01],


edgarriba:

PR #3298, opened 2025-09-19
refactored img_to_string and optimized it by 45x

sidd-27:
vectorized the entire per pixel calculation of the rgb2totch function and made it compute every row at once, this does mean that the peak memory usage will be higher, but this is not as important as the size of images to be printed on the terminal is usually smaller, but if 4k images are also to be printed, then i could also write a batched implementation which reduces memory usage

   --- Running timed benchmark (CUDA) ---
optimized :  48.7307 ms/run
Kornia       : 2164.3892 ms/run

🚀 Speedup: 44.42x

but considering that the majority of the time in printing images is spent in the actual printing part, this wouldn't be much prominent in the print_image function; nevertheless, this pr does clean up the file quite a bit, removing a bunch of redundant functions and the lookup tables stored in code

### DIFF
--- a/kornia/utils/image_print.py
+++ b/kornia/utils/image_print.py
@@ -22,338 +22,35 @@ Nice long listing of all 256 colors and their codes.
 Taken from https://gist.github.com/klange/1687427
 """
 
-import re
-from typing import Tuple, Union
+from typing import Union
 
 import torch
-from torch import float16, float32, float64
-
 import kornia
 from kornia.core import Tensor
 from kornia.core.check import KORNIA_CHECK_IS_IMAGE, KORNIA_CHECK_SHAPE
 from kornia.io import ImageLoadType
 
-# color look-up table
-# 8-bit, RGB hex
-CLUT = [
-    # Primary 3-bit (8 colors). Unique representation!
-    ("00", "000000"),
-    ("01", "800000"),
-    ("02", "008000"),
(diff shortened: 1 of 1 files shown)


sidd-27:
i made a small change to the original pr to maintain backwards compatibility of the rgb2torch function, if that function isnt needed i could revert the changes to make the code shorter and cleaner

kornia/utils/image_print.py
```diff
@@ -22,296 +22,16 @@
...
-from torch import float16, float32, float64
 
 import kornia
 from kornia.core import Tensor
 from kornia.core.check import KORNIA_CHECK_IS_IMAGE, KORNIA_CHECK_SHAPE
 from kornia.io import ImageLoadType
 
-# color look-up table
-# 8-bit, RGB hex
-CLUT = [
```
edgarriba, line 38: why is this code disappearing?

edgarriba (changes requested):
you need also to rebase from main

kornia/utils/image_print.py
sidd-27: The xterm-256 conversion can actually be represented as a formulae, so instead of a look up table which cannot be vectorized, i replaced it with the actual formula which can, to calculate the values for all the pixels at once

edgarriba:

PR #3304, opened 2025-09-28
optimized polynomial solvers

sidd-27:
for the first 2 functions, we compute the basis vector containing all 40 possible products, then using the  "blueprint" matrix of ones and zeroes we choose only the elements we need, this technically uses more operations but instead of each individual operation making a new kernel and all the overhead associated with that, we instead just have one matrix multiplication call which is already very optimized for GPU's

https://colab.research.google.com/drive/1iXTRbNP7YbPpWGEXC0r3c_4fHTmkz7bM?usp=sharing

here's the benchmarking and validation script for all the three functions

Benchmarking with Batch Size: 64
  Original Implementation:     1.0470 ms
  Vectorized Implementation:   0.0654 ms
   Speedup: 16.02x faster

Benchmarking with Batch Size: 64
  Original:     0.4171 ms
  Vectorized:   0.0568 ms  (Speedup: 7.34x)

the second optimization refactors the determinant calculation by replacing hundreds of hardcoded arithmetic operations with a more efficient, gather-scatter approach. The entire formula is pre-computed into "instruction" tensors (multiplication_indices, signs, coefficient_map) that guide the process. First, a single gather operation uses advanced indexing to collect all the necessary values from the input tensor A. Then, all multiplications and sign applications are performed in two clean, vectorized steps. Finally, a single, scatter_add_ operation takes all the computed product terms and, guided by the coefficient_map, sums them into their correct final positions in the output tensor.

Benchmarking with Batch Size: 64
  Original Implementation:    23.8013 ms
  Vectorized Implementation:   0.0975 ms
   Speedup: 244.12x faster

also wrote test for these functions


sidd-27:
noqa: PLW0603 has been used to supress the ruff errors, as the matrix isnt designed to change under any operation, if theres a better practice on how to handle global variables pls do let me know

PR #3307, opened 2025-09-30
feat: add uv.lock for reproducible dependency management

Meggison:
## Summary
Adds `uv.lock` file for reproducible dependency management while maintaining full backward compatibility.

## Changes
- ✅ Add `uv.lock` file with 128 resolved packages
- ✅ Update `pyproject.toml` with inline dependencies for better uv compatibility
- ✅ Configure PyTorch CPU index for cross-platform builds
- ✅ Update development scripts to optionally use lock file
- ✅ Add CI verification of lock file consistency
- ✅ Update documentation in README.md and CONTRIBUTING.md

## Testing
- ✅ Lock file generated and verified with `uv lock --check`
- ✅ Pre-commit hooks pass
- ✅ Existing workflows unchanged
- ✅ Both pip and uv installation methods work

## Backward Compatibility
- ✅ `pip install kornia` works exactly as before
- ✅ Existing CI/CD workflows unchanged
- ✅ Developers can still use pip if preferred
- ✅ Lock file is optional enhancement, not requirement


requirements/requirements-dev.txt
edgarriba, line 1: do we still need the requirements files ?

pyproject.toml
```diff
@@ -11,6 +11,53 @@ authors = [
   { name = "Edgar Riba", email = "edgar@kornia.org" },
 ]
 requires-python = ">=3.9"
+dependencies = [
+    "kornia_rs>=0.1.9",
+    "packaging",
+    "torch>=2.0.0",
+]
+optional-dependencies.dev = [
```
edgarriba, line 19: are we declaring this twice ? this is the same as in dependency-groups

requirements/requirements-dev.txt
Meggison: I don't think so, since dependencies would only be handled by the uv

edgarriba: then lets remove it

edgarriba (approved):
LGTM. Can we merge ?

PR #3310, opened 2025-10-04
Fix RandomCrop transformation matrix to reflect actual image scaling

Meggison:
## Problem

When `RandomCrop` with `cropping_mode='slice'` or `'resample'` generates source coordinates that extend beyond input bounds, the transformation matrix incorrectly shows `scale=1.0` instead of the actual scaling applied to the image. This causes **silent data corruption** when transforming bounding boxes in object detection pipelines.

## Root Cause

The issue occurs when:
1. Input image is smaller than target crop size (e.g., input: 329×1209, target: 416×416)
2. `CropGenerator` creates source coordinates that extend beyond input bounds (e.g., y-coordinates go to 415 when input height is only 329)
3. `crop_by_indices` automatically clamps to valid bounds and resizes, but the transformation matrix still uses the original out-of-bounds coordinates
4. Result: matrix shows `scale_y = 1.0` when actual scaling is `1.265`

## Solution

Clamp source coordinates to actual input bounds in `compute_transformation()` before calculating the perspective transform. This ensures the transformation matrix accurately represents the geometric scaling applied to the image.

**Suggested fix:**
```python
h, w = input.shape[-2:]
src[..., 0] = src[..., 0].clamp(0, w - 1)  # x coordinates
src[..., 1] = src[..., 1].clamp(0, h - 1)  # y coordinates
```

```python
# Before fix:
# Input: 329×1209, Target: 416×416
matrix = crop.transform_matrix[0]
print(matrix[1,1])  # 1.0000 (WRONG - silent corruption)

# After fix:
# Input: 329×1209, Target: 416×416  
matrix = crop.transform_matrix[0]
print(matrix[1,1])  # 1.2652 (CORRECT - reflects actual scaling)
```

### DIFF
--- a/kornia/filters/dissolving.py
+++ b/kornia/filters/dissolving.py
@@ -139,6 +139,9 @@ class StableDiffusionDissolving(ImageModule):
             steps_offset=1,
         )
 
+        # Filter out arguments that are not supported by all component models
+        kwargs.pop("offload_state_dict", None)
+
         if version == "1.4":
             self._sdm_model = StableDiffusionPipeline.from_pretrained(  # type:ignore
                 "CompVis/stable-diffusion-v1-4", scheduler=scheduler, **kwargs


kornia/augmentation/_2d/geometric/crop.py
```diff
@@ -161,7 +161,13 @@ def precrop_padding(
 
     def compute_transformation(self, input: Tensor, params: Dict[str, Tensor], flags: Dict[str, Any]) -> Tensor:
         if flags["cropping_mode"] in ("resample", "slice"):
```
edgarriba, line 163: actually this should be only valid with "resample", right ? for slicing we don't need anry transform matrix

edgarriba (commented):
@shijianjian

kornia/augmentation/_2d/geometric/crop.py
Meggison: I believe slice still needs it to correctly transform keypoints and bounding boxes to match the cropped output. I was encountering the initial error with it as well

edgarriba: oh, right!

edgarriba:
double check why so many tests failing

Meggison:
@edgarriba Facing issues with CLIP models, trying to figure out if it's an issue with the external imports or the CI

PR #3317, opened 2025-10-09
optimized denormalize

sidd-27:
https://github.com/kornia/kornia/pull/3313 refer to this for details

### DIFF
--- a/kornia/enhance/normalize.py
+++ b/kornia/enhance/normalize.py
@@ -258,13 +258,18 @@ def denormalize(data: Tensor, mean: Union[Tensor, float], std: Union[Tensor, flo
         mean = torch.as_tensor(mean, device=data.device, dtype=data.dtype)
         std = torch.as_tensor(std, device=data.device, dtype=data.dtype)
 
-    def _match_dims(x: Tensor, ref: Tensor) -> Tensor:
-        if x.dim() < ref.dim():
-            x = x.view(*x.shape, *([1] * (ref.dim() - x.dim())))
-        return x
+    if mean.dim() == 1:
+        mean = mean.view(1, -1, *([1] * (data.dim() - 2)))
+    # If the tensor is >1D (e.g., (B, C)), reshape to (B, C, 1, ...)
+    else:
+        while len(mean.shape) < data.dim():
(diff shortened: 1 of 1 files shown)


kornia/enhance/normalize.py
```diff
@@ -258,12 +258,20 @@ def denormalize(data: Tensor, mean: Union[Tensor, float], std: Union[Tensor, flo
         mean = torch.as_tensor(mean, device=data.device, dtype=data.dtype)
         std = torch.as_tensor(std, device=data.device, dtype=data.dtype)
 
-    mean = mean[..., None]
-    std = std[..., None]
+    if mean.dim() == 1:
+        mean = mean.view(1, -1, *([1] * (data.dim() - 2)))
+    # If the tensor is >1D (e.g., (B, C)), reshape to (B, C, 1, ...)
+    else:
+        while len(mean.shape) < data.dim():
```
edgarriba: i don't feel this solution with a while loop here "correct"

i'd go for something like

```python
def denormalize(data: Tensor, mean: Union[Tensor, float], std: Union[Tensor, float]) -> Tensor:
    """Undo normalization by applying data * std + mean.

    Args:
        data: Input tensor, e.g. (C, H, W) or (B, C, H, W)
        mean: Mean used for normalization. Can be scalar or Tensor with shape (C,) or (B, C)
        std: Standard deviation used for normalization. Same shape rules as mean.
    """
    if isinstance(mean, (int, float)):
        mean = torch.tensor(mean, dtype=data.dtype, device=data.device)
    if isinstance(std, (int, float)):
        std = torch.tensor(std, dtype=data.dtype, device=data.device)

    def _match_dims(x: Tensor, ref: Tensor) -> Tensor:
        if x.dim() < ref.dim():
            x = x.view(*x.shape, *([1] * (ref.dim() - x.dim())))
        return x

    mean = _match_dims(mean, data)
    std = _match_dims(std, data)

    return data * std + mean
```

edgarriba:
verify tests

edgarriba:
@sidd-27 fix ci

edgarriba:
@sidd-27 fix ci

PR #3319, opened 2025-10-15
opti rgb_to_gray

VyasGuru:
split the pr mentioned below into multiple prs for every function
https://github.com/kornia/kornia/pull/3314

### DIFF
--- a/kornia/color/gray.py
+++ b/kornia/color/gray.py
@@ -98,12 +98,13 @@ def rgb_to_grayscale(image: Tensor, rgb_weights: Optional[Tensor] = None) -> Ten
         rgb_weights = rgb_weights.to(image)
 
     # unpack the color image channels with RGB order
-    r: Tensor = image[..., 0:1, :, :]
-    g: Tensor = image[..., 1:2, :, :]
-    b: Tensor = image[..., 2:3, :, :]
+    r, g, b = image.unbind(dim=-3)
 
     w_r, w_g, w_b = rgb_weights.unbind()
-    return w_r * r + w_g * g + w_b * b
+    out = w_r * r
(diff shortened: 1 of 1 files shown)


sidd-27:
the code is the same as mine, he's helping me, also the precommits seem broken by something else

kornia/color/gray.py
```diff
@@ -98,12 +98,13 @@ def rgb_to_grayscale(image: Tensor, rgb_weights: Optional[Tensor] = None) -> Ten
...
-    g: Tensor = image[..., 1:2, :, :]
-    b: Tensor = image[..., 2:3, :, :]
+    r, g, b = image.unbind(dim=-3)
 
     w_r, w_g, w_b = rgb_weights.unbind()
-    return w_r * r + w_g * g + w_b * b
+    out = w_r * r
+    out = torch.addcmul(out, w_g, g)
+    out = torch.addcmul(out, w_b, b)
+    return out.unsqueeze(-3)
```
edgarriba, line 107: Is the `addcmul` making any difference here ?why is the unsqueeze needed? if needed add comment why

sidd-27: I've added benchmarks for this in my original pr(the last block in the notebook) it shows that there is about a 1.5x speedup. It's running on colab GPUs however so if possible do run that script locally

Meggison: @edgarriba The `addcmul` here is to keep the dimensionality consistent with the original implementation, since the slicing was replaced with unbind (which loses the channel dimension). Asides that everything else is fine. The precommit issues seem to be from the previous linting errors, so I suggest retesting the PR

PR #3322, opened 2025-10-15
rgb_to_ycbcr

VyasGuru:
splitting the pr mentioned below into multiple prs for each function
https://github.com/kornia/kornia/pull/3314

### DIFF
--- a/kornia/color/ycbcr.py
+++ b/kornia/color/ycbcr.py
@@ -25,11 +25,6 @@ from kornia.core import ImageModule as Module
 from kornia.core import Tensor
 
 
-def _rgb_to_y(r: Tensor, g: Tensor, b: Tensor) -> Tensor:
-    y: Tensor = 0.299 * r + 0.587 * g + 0.114 * b
-    return y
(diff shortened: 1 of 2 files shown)


kornia/color/ycbcr.py
```diff
@@ -52,15 +47,22 @@ def rgb_to_ycbcr(image: Tensor) -> Tensor:
...
         raise ValueError(f"Input size must have a shape of (*, 3, H, W). Got {image.shape}")
 
-    r: Tensor = image[..., 0, :, :]
-    g: Tensor = image[..., 1, :, :]
-    b: Tensor = image[..., 2, :, :]
+    input_shape = image.shape
+    h, w = image.shape[-2:]
+    image = image.reshape(-1, 3, h, w)
+
+    y_weights = torch.tensor([0.299, 0.587, 0.114], device=image.device, dtype=image.dtype)
```
edgarriba, line 54: what if input is not (0-1) float ? should we just raise runtime error ?

Issue #3194, opened 2025-07-23
Bug Report: VideoSequential same_on_batch=False and same_on_frame=True not working correctly


mluijten:
# Bug Report: VideoSequential `same_on_batch=False` and `same_on_frame=True` not working correctly

## Describe the bug

When using `VideoSequential` with `same_on_batch=False` and `same_on_frame=True`, the parameters are not respected correctly:

1. **`RandomCrop`**: Ignores both `same_on_batch=False` and `same_on_frame=True`
2. **`RandomHorizontalFlip`**: Ignores `same_on_batch=False` (but respects `same_on_frame=True`)

## Reproduction steps

```python
import kornia.augmentation as K
import numpy as np
import torch

# Create test data: 2 videos, 1 channel, 20 frames each, 100x100 pixels
# Each video contains identical frames
np.random.seed(42)
frame_data = np.random.rand(100, 100).astype(np.float32)
video = torch.from_numpy(frame_data)[None, None, None, :, :].repeat(2, 1, 20, 1, 1)

print("Original video shape:", video.shape)  # Should be [2, 1, 20, 100, 100]

# Test 1: RandomCrop with same_on_batch=False, same_on_frame=True
print("\n=== Test 1: RandomCrop ===")
crop_aug = K.VideoSequential(
    K.RandomCrop(size=(50, 50), same_on_batch=False),
    data_format="BCTHW",
    same_on_frame=True,
)

cropped = crop_aug(video)

# Check if frames within each video are identical (should be True)
for i in range(cropped.shape[0]):
    frames_identical = torch.all(cropped[i] == cropped[i][:, 0:1, :, :])
    print(f"Video {i} - All frames identical: {frames_identical.item()}")

# Check if videos are different from each other (should be True) 
videos_different = not torch.all(cropped[0] == cropped[1]).item()
print(f"Videos are different: {videos_different}")

# Test 2: RandomHorizontalFlip with same_on_batch=False, same_on_frame=True
print("\n=== Test 2: RandomHorizontalFlip ===")
flip_aug = K.VideoSequential(
    K.RandomHorizontalFlip(p=1.0, same_on_batch=False),  # p=1.0 to ensure flipping happens
    data_format="BCTHW",
    same_on_frame=True,
)

flipped = flip_aug(video)

# Check if frames within each video are identical (should be True)
for i in range(flipped.shape[0]):
    frames_identical = torch.all(flipped[i] == flipped[i][:, 0:1, :, :])
[...]

PR #3330, opened 2025-10-26
fix: correct same_on_frame and same_on_batch logic in VideoSequential

RakshithaKowlikar:
#### Changes
This corrects `VideoSequential.forward_parameters()`'s logic when using `same_on_frame` and `same_on_batch`. Previously, these parameters were not applied correctly across frames or batches, leading to inconsistent results. Based on these two flags, the code now determines `mod_shape` that each augmentation should use before creating `mod_param`. All four combinations of `(same_on_frame, same_on_batch)` now work as intended.

Fixes # [3194](https://github.com/kornia/kornia/issues/3194)

#### Type of change

#### Checklist

### DIFF
--- a/kornia/augmentation/container/video.py
+++ b/kornia/augmentation/container/video.py
@@ -157,6 +157,20 @@ class VideoSequential(ImageSequential):
         repeated = param[:, None, ...].repeat(1, frame_num, *([1] * len(param.shape[1:])))
         return repeated.reshape(-1, *list(param.shape[1:]))
 
+    def __broadcast_param__(
+        self, v: Tensor, batch_shape: torch.Size, frame_num: int, same_on_frame: bool, same_on_batch: bool
+    ) -> Tensor:
+        if not v.numel():
+            return v
+
+        if same_on_frame and same_on_batch:
+            return v.repeat(batch_shape[0] * frame_num, *([1] * (v.ndim - 1)))
+        elif same_on_frame:
+            return self.__repeat_param_across_channels__(v, frame_num)
+        elif same_on_batch:
+            return v.unsqueeze(1).repeat(1, batch_shape[0], *([1] * (v.ndim - 1))).reshape(-1, *v.shape[1:])
+        return v
+
     def _input_shape_convert_in(self, input: Tensor, frame_num: int) -> Tensor:
         # Convert any shape to (B, T, C, H, W)
         if self.data_format == "BCTHW":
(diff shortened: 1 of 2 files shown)


kornia/augmentation/container/video.py
```diff
@@ -183,39 +183,57 @@ def forward_parameters(self, batch_shape: torch.Size) -> List[ParamItem]:
...
-                    raise ValueError("Sequential is currently unsupported for ``same_on_frame``.")
-                param = ParamItem(name, seq_param)
-            elif isinstance(module, (_AugmentationBase, K.MixAugmentationBaseV2)):
-                mod_param = module.forward_parameters(batch_shape)
-                if self.same_on_frame:
+            if isinstance(module, (K.RandomCrop, _AugmentationBase, K.MixAugmentationBaseV2)):
+                is_same_on_batch = getattr(module, "same_on_batch", False)
+
+                if self.same_on_frame and is_same_on_batch:
+                    mod_shape = torch.Size([1, *batch_shape[1:]])
```
edgarriba, line 206: just curious why shape (1, *)

edgarriba:
@Meggison any ideas why pre-commit is failing ?

kornia/augmentation/container/video.py
RakshithaKowlikar: Before each frame or video led to the generation of random parameters (since the shape included B), causing the data to differ and fail on a few combinations. This new shape reuses the same parameters across all frames and batches, giving the expected output later.

edgarriba:
@RakshithaKowlikar please rebase from `main` branch

Meggison:
@edgarriba I'm not sure, but it seems like it might be a case of the last update not being pulled yet before the PR was made, because it still refers to the same linting issues. The code in the errors are different from what's actually in the codebase currently

Meggison (changes requested):
Everything seems fine, but could you check if you've pulled the most recent updates? Might help with the precommit issues. If that doesn't work I'll have to look into a fix myself, but do let me know

tests/augmentation/container/test_video_sequential.py
```diff
@@ -166,3 +166,108 @@ def test_autocast(self, data_format, device, dtype):
             output = aug(imgs)
 
         assert output.dtype == dtype, "Output image dtype should match the input dtype"
+
+    @pytest.mark.parametrize("data_format", ["BCTHW", "BTCHW"])
```
edgarriba, line 170: to avoid boilerplate code, why not parametrising somehow the test with shape and expected indices or something like that to merge the three tests into one

edgarriba (approved):
LGTM. Rework the tests to be more compact

tests/augmentation/container/test_video_sequential.py
RakshithaKowlikar: I have combined the four tests into a single parametrized test. Thanks for the feedback!

PR #3334, opened 2025-10-30
Fix quaternion scalar arithmetic gradient issues (Issue #3101)

Meggison:
Fixes https://github.com/kornia/kornia/issues/3101

- Add wrap_in_parameter flag to Quaternion.__init__ to preserve gradient flow
- Add _to_scalar_quaternion helper method for consistent scalar handling
- Update all arithmetic operations (__add__, __sub__, __mul__, __neg__, __getitem__) to use wrap_in_parameter=False
- Add reverse operations (__rmul__, __radd__, __rsub__, __rtruediv__, __rdiv__) for scalar arithmetic
- Add comprehensive test for scalar multiplication (test_quaternion_scalar_multiplication)

This fixes TypeError issues when doing operations like 'q * q * 5' while preserving gradient computation graphs.

## Testing
Verified that the following operations now work correctly:
- `q * q * 5` (original failing case)
- `5 * q` (reverse multiplication)
- `q + 2.0`, `2.0 + q` (addition with scalars)
- `q / 2.0` (division by scalars)
- Gradient preservation through all operations

## Backward Compatibility
All existing quaternion-to-quaternion operations remain unchanged. This PR only adds support for scalar/tensor operations that were previously failing.

### DIFF
--- a/kornia/geometry/liegroup/se2.py
+++ b/kornia/geometry/liegroup/se2.py
@@ -207,7 +207,7 @@ class Se2(Module):
             >>> v = torch.ones((1, 3))
             >>> s = Se2.exp(v).log()
             >>> s
-            tensor([[1., 1., 1.]], grad_fn=<StackBackward0>)
+            tensor([[1.0000, 1.0000, 1.0000]], grad_fn=<StackBackward0>)
 
         """
         theta = self.so2.log()


kornia/geometry/quaternion.py
```diff
@@ -75,11 +75,15 @@ class Quaternion(Module):
 
     """
 
-    def __init__(self, data: Tensor) -> None:
+    _data: Union[Tensor, Parameter]
+
+    def __init__(self, data: Tensor, wrap_in_parameter: bool = True) -> None:
```
edgarriba: wondering if the data should we expect from user to be passed explicitly as tensor or parameter, and then have `as_parameter()` or similar? feels a bit weird the `wrap_in_parameter` argument

Meggison: You might be right, it may end up with unnecessary complexity when handling function calls later. Could try explicit type checking instead

edgarriba (approved):

Issue #2876, opened 2024-04-08
Bug in normalize_min_max


vgilabert94:
### Describe the bug

From https://github.com/kornia/kornia/pull/2860#issuecomment-2022116373

When the input tensor is an image without batch (C,H,W), where the normalisation would be wrongly applied (without giving error).

A possible (easy) solution is to convert inside the function the input to batch (B,C,H,W) and also modify view by reshape to ensure the continuity of the tensors in torch. Also could be a solution adjust internally the normalize_min_max to accept non batched inputs.

### Reproduction steps

```bash
import torch
from kornia.enhance import normalize_min_max

rng = torch.manual_seed(1)
torch.set_printoptions(precision=10)
input_tensor = torch.rand(1,3,3)
res = normalize_min_max(input_tensor)
input_tensor = torch.rand(1,1,3,3)
expected = normalize_min_max(input_tensor)
print(input_tensor)
tensor([[[[0.4387779236, 0.6386804581, 0.5246658921],
          [0.6826140881, 0.3051494956, 0.4635456204],
          [0.4549863338, 0.5724719763, 0.4980025887]]]])

print(res)
tensor([[[0.9999978542, 0.0000000000, 0.2587345839],
         [0.9154204130, 0.0000000000, 0.9999986887],
         [0.0000000000, 0.9999971390, 0.4825139046]]])

print(expected)
tensor([[[[0.3540148437, 0.8836062551, 0.5815533996],
          [0.9999973178, 0.0000000000, 0.4196306169],
          [0.3969549537, 0.7082035542, 0.5109156966]]]])
```
```


### Expected behavior

Excepted should be normalize having into account only the H and W channels of images (like in expected result).

### Environment

```shell
PyTorch version: 2.1.2+cpu
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: Microsoft Windows 10 Pro
GCC version: Could not collect
Clang version: Could not collect
CMake version: Could not collect
Libc version: N/A

Python version: 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] (64-bit runtime)
Python platform: Windows-10-10.0.19045-SP0
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
[...]

PR #3352, opened 2025-11-08
Fix Bug #2876: normalize_min_max now supports 3D tensors (C, H, W)

funnym0nk3y:
This commit fixes a bug where normalize_min_max incorrectly handled 3-dimensional tensors in (C, H, W) format without a batch dimension.

Changes:
- Applied @perform_keep_shape_image decorator to normalize_min_max to automatically handle tensors of any dimension (*, C, H, W)
- Updated docstring to reflect support for (*, C, H, W) format instead of requiring (B, C, *) format
- Fixed error message for max_val parameter (was incorrectly 'b')
- Removed manual shape validation for len(x.shape) < 3 since the decorator now handles 2D, 3D, and higher dimensional inputs
- Added comprehensive test cases for 3D and 2D tensor inputs

The decorator ensures:
- 2D tensors (H, W) are converted to (1, 1, H, W) for processing
- 3D tensors (C, H, W) are converted to (1, C, H, W) for processing
- 4D tensors (B, C, H, W) remain unchanged
- Original shape is restored after normalization

This maintains backward compatibility while fixing the bug.

Resolves: #2876

#### Changes

Fixes # (issue)

#### Type of change

- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases


edgarriba (changes requested):
CI tests are broken

PR #3355, opened 2025-11-09
Add requires_grad parameter to Quaternion class

### LINKED ISSUE #3291: Quaternion with `requires_grad=False` should propagate `requires_grad=False` to results of math operations

funnym0nk3y:
This commit addresses issue #3291 by adding support for explicit gradient tracking control in the Quaternion class, making its behavior consistent with PyTorch tensors.

Changes:
- Added optional `requires_grad` parameter to Quaternion.__init__()
  - When None (default), preserves the requires_grad flag from input data
  - When specified, explicitly sets the requires_grad flag on the tensor

- Added `requires_grad` parameter to all class factory methods:
  - identity(): defaults to False
  - from_coeffs(): defaults to False
  - from_matrix(): defaults to None (inherits from matrix)
  - from_euler(): defaults to None (inherits from input angles)
  - from_axis_angle(): defaults to None (inherits from axis_angle)
  - random(): defaults to False

- Mathematical operations already preserve requires_grad correctly through PyTorch's automatic gradient propagation

The implementation ensures backward compatibility - all existing code will continue to work as the new parameter has sensible defaults.

Fixes #3291

#### Changes

Fixes # (issue)

#### Type of change

- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases

### DIFF
--- a/kornia/geometry/quaternion.py
+++ b/kornia/geometry/quaternion.py
@@ -99,7 +99,7 @@ class Quaternion(Module):
         if not isinstance(data, (Tensor, Parameter)):
             raise TypeError(f"Expected Tensor or Parameter, got {type(data)}")
         # KORNIA_CHECK_SHAPE(data, ["B", "4"])  # FIXME: resolve shape bugs. @edgarriba
-        if requires_grad is not None and isinstance(data, Tensor):
+        if requires_grad is not None and not isinstance(data, Parameter):
             data = data.requires_grad_(requires_grad)
         self._data = data
 
@@ -496,7 +496,7 @@ class Quaternion(Module):
         batch_size: Optional[int] = None,
         device: Optional[Device] = None,
         dtype: Dtype = None,
-        requires_grad: bool = False,
+        requires_grad: Optional[bool] = None,
     ) -> "Quaternion":
         """Create a quaternion representing an identity rotation.
 
@@ -504,7 +504,8 @@ class Quaternion(Module):
             batch_size: the batch size of the underlying data.
             device: device to place the result on.
             dtype: dtype of the result.
-            requires_grad: If True, the quaternion will track gradients.
+            requires_grad: If specified, sets the requires_grad flag on the quaternion.
+                If None, defaults to False (no gradient tracking).
 
         Example:
             >>> q = Quaternion.identity()
@@ -518,7 +519,7 @@ class Quaternion(Module):
         return cls(data, requires_grad=requires_grad)
 
     @classmethod
-    def from_coeffs(cls, w: float, x: float, y: float, z: float, requires_grad: bool = False) -> "Quaternion":
+    def from_coeffs(cls, w: float, x: float, y: float, z: float, requires_grad: Optional[bool] = None) -> "Quaternion":
         """Create a quaternion from the data coefficients.
 
         Args:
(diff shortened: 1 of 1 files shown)


kornia/geometry/quaternion.py
```diff
@@ -89,12 +91,16 @@ def __init__(self, data: Union[Tensor, Parameter]) -> None:
...
+            >>> # Explicitly set requires_grad
+            >>> q3 = Quaternion(data, requires_grad=True)
 
         """
         super().__init__()
         if not isinstance(data, (Tensor, Parameter)):
             raise TypeError(f"Expected Tensor or Parameter, got {type(data)}")
         # KORNIA_CHECK_SHAPE(data, ["B", "4"])  # FIXME: resolve shape bugs. @edgarriba
+        if requires_grad is not None and not isinstance(data, Parameter):
+            data = data.requires_grad_(requires_grad)
```
Meggison, line 103: Test if gradient is being properly now, I'm not sure if this would work for the gradient initialisation since the Quartenion class doesn't directly track gradients. 

an example of how the gradients are currently gotten:

```python
import torch
from kornia.geometry import Quaternion

q = Quaternion.identity()       # q.data is a tensor, requires_grad=False initially
q.data.requires_grad_(True)     # enable autograd on the underlying tensor

out = (q.w ** 2).sum()          # out is a torch.Tensor (scalar)
out.backward()

print("q.data:", q.data)  # returns q.data: tensor([1., 0., 0., 0.], requires_grad=True)
```

boschet:
Should I close or is this still considered for merge?

edgarriba:
added comments here  https://github.com/kornia/kornia/issues/3291

edgarriba:

PR #3356, opened 2025-11-10
Remove duplicate optional-dependencies; keep dependency-groups per PEP 735

gaurav29kumar:
This pull request fixes duplicated dependency definitions in the pyproject.toml file.
Previously, optional dependencies were declared twice — once under [project.optional-dependencies] and again under [dependency-groups].

I removed the duplicates and kept a single, consistent section ([dependency-groups]) following the latest packaging standards (PEP 735).
This helps prevent confusion and ensures cleaner dependency management.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,48 +16,6 @@ dependencies = [
     "packaging",
     "torch>=2.0.0",
 ]
-optional-dependencies.dev = [
-    "accelerate",
-    "coverage",
-    "diffusers",
-    "ivy>=1.0.0.0",
-    "mypy",
-    "numpy<3",
-    "onnx",
-    "onnxruntime",
-    "pillow",
-    "pre-commit>=2",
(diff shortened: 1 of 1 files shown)


edgarriba (changes requested):
broken CI

pyproject.toml
```diff
@@ -16,48 +16,6 @@ dependencies = [
...
-    "pre-commit>=2",
-    "pytest==8.3.5",
-    "pytest-cov",
-    "pytest-timeout",
-    "requests",
-    "setuptools>=61.2",
-    "transformers<4.57",
-    "types-requests",
-]
-optional-dependencies.x = [
```
johnnv1, line 38: x should be still be an optional dep 

Dependency groups are from only to not export to final user dev stuff like testing/docs deps -- https://github.com/astral-sh/uv/issues/9011#issuecomment-2467955313

PR #3357, opened 2025-11-10
Fix crash in AugmentationSequential.get_random_forward_sequence with mix-only augmentations

gaurav29kumar:
Fixes # (issue)
Added a check:

if multinomial_weights.sum().item() == 0:

If true, directly select the mix augmentation(s) instead of calling torch.multinomial.

Ensures safe forward pass when the augmentation list contains only mix augmentations.

Benefits / Expected Behavior#:

No crash occurs when using AugmentationSequential(MixUp(), CutMix()).

Either MixUp or CutMix is applied correctly.

get_random_forward_sequence now handles mix-only cases safely.

Minimal and backward-compatible fix.


edgarriba (changes requested):
broken CI

edgarriba:

PR #3378, opened 2025-12-18
io: fallback to Pillow for read_image_any when kornia_rs can't handle RGBA/paletted images

Meggison:
#### Changes
Added PIL/Pillow fallback to handle RGBA and paletted images in `_load_image_to_tensor()`. The `kornia_rs` backend doesn't currently support these formats, causing load failures for PNG images with transparency or palette modes.

**Solution**: Detect unsupported image modes and convert to RGB using PIL before tensor conversion. Preserves existing behavior for supported formats (JPEG, RGB).

**Note**: This is an interim solution until RGBA support can be implemented in `kornia_rs`. Open to working on the Rust backend implementation in a future PR.

Fixes # (issue)

#### Type of change
- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases

#### Checklist

### DIFF
--- a/kornia/io/io.py
+++ b/kornia/io/io.py
@@ -46,7 +46,7 @@ class ImageLoadType(Enum):
 def _load_image_to_tensor(path_file: Path, device: Device) -> Tensor:
     """Read an image file and decode using the Kornia Rust backend.
 
-    The decoded image is returned as numpy array with shape HxWxC.
+    Falls back to Pillow when kornia_rs cannot handle the image format.
 
     Args:
         path_file: Path to a valid image file.
@@ -55,7 +55,7 @@ def _load_image_to_tensor(path_file: Path, device: Device) -> Tensor:
     Return:
         Image tensor with shape :math:`(3,H,W)`.
     """
-    # read image and return as `np.ndarray` with shape HxWxC
+    # Try reading with kornia_rs first, fallback to Pillow if needed
     try:
         if path_file.suffix.lower() in [".jpg", ".jpeg"]:
             img = kornia_rs.read_image_jpegturbo(str(path_file))


edgarriba (commented):
I don’t think we should add Pil as dependency (we have been following the rule of zero dependencies) but instead extend kornia_rs.io with whatever is missing

Meggison:
@edgarriba that makes sense. I could actually work on extending the image loading on kornia rs to handle rgba and other formats. Currently testing Pillow to diagnose the issue before I start working on a direct fix

edgarriba:
> @edgarriba that makes sense. I could actually work on extending the image loading on kornia rs to handle rgba and other formats. Currently testing Pillow to diagnose the issue before I start working on a direct fix

Sounds good! There's an almost ready PR to land that refactors the io in kornia rust that will help in this direction. Besides, I have a prototype to provide a PIL style api for images, that if you are interested in I could hand off to you

Meggison:
@edgarriba great! You can hand it off to me, and work on it as well

edgarriba:
> @edgarriba great! You can hand it off to me, and work on it as well

https://github.com/kornia/kornia-rs/pull/573 which i suggest to wait till we merge https://github.com/kornia/kornia-rs/pull/571

kornia/io/io.py
```diff
@@ -52,18 +52,14 @@ def _load_image_to_tensor(path_file: Path, device: Device) -> Tensor:
...
         Image tensor with shape :math:`(3,H,W)`.
-
     """
-    # read image and return as `np.ndarray` with shape HxWxC
-    if path_file.suffix.lower() in [".jpg", ".jpeg"]:
-        img = kornia_rs.read_image_jpegturbo(str(path_file))
-    else:
-        img = kornia_rs.read_image_any(str(path_file))
+    # read image using the unified API
+    img = kornia_rs.read_image(str(path_file))
```
edgarriba: this is still not published

edgarriba:
can you rebase on main ?

edgarriba:
what's the plan here ? we don't want to add Pillow as dep for now

Meggison:
@edgarriba I'm going to close the PR for now, I'll make a new PR for the Rust changes once they are published

PR #3424, opened 2025-12-30
Fix AugmentationSequential crash with mix-only augmentations

### LINKED ISSUE #3300: Crash in get_random_forward_sequence when only mixup and cutmix are given in AugmentationSequential
### Describe the bug

When only MixUp and CutMix are given in AugmentationSequential, there's a crash at line 192 in get_random_forward_sequence in file kornia/augmentation/container/image.py. This happens both in 0.8.1 and in main. 

### Reproduction steps

```bash
1.Define AugmentationSequential w/ ONLY MixUp and CutMix
2.Apply augmentation
3.
...
```

### Expected behavior

Either augmentation should be applied

### Environment

```shell
python collect_env.py

- PyTorch Version (e.g., 1.0): 3.11
- OS (e.g., Linux): Linux
- How you installed PyTorch (`conda`, `pip`, source): Kaggle notebook
- Build command you used (if compiling from source): N/A
- Python version: 3.11
- CUDA/cuDNN version:
- GPU models and configuration: P100
```

### Additional context

_No response_

Saksham-st12:
#### Summary

This PR fixes a `RuntimeError` in `AugmentationSequential` that occurs when it is initialized with only mix augmentations (e.g. `RandomMixUpV2`, `RandomCutMixV2`) and `random_apply=True`.

#### Issue

In `get_random_forward_sequence`, mix augmentations are excluded from the multinomial sampling step since at most one mix operation is allowed per forward pass. When the augmentation sequence contains only mix augmentations, this results in a zero-sum `multinomial_weights` tensor. Calling `torch.multinomial` on this distribution raises:
RuntimeError: invalid multinomial distribution (sum of probabilities <= 0)

#### Fix

The following changes were made in `kornia/augmentation/container/image.py`:

- Added an explicit guard to detect the zero-sum case and skip multinomial sampling when no non-mix augmentations are available.
- Updated the mix-augmentation injection logic to correctly handle empty indices, ensuring that a single mix augmentation is applied when appropriate.

This preserves the intended behavior of applying at most one mix augmentation per forward pass without causing a crash.

#### Tests

- Added a regression test `test_mixup_cutmix_only` in `tests/augmentation/container/test_augmentation_sequential.py`.
- The test verifies that `AugmentationSequential` initialized with only mix augmentations no longer crashes and produces an output tensor of the expected shape.

Fixes #3300

Several CI checks are failing across collector, tests (multiple Python versions), tutorials, and typecheck.

The failures appear to be repo-wide and unrelated to the changes in AugmentationSequential (e.g. typecheck diagnostics and tutorial/test failures in unrelated modules).

All tests related to this change pass locally.
Please let me know if you’d like me to rebase, adjust scope, or if these failures are known on main.

### DIFF
--- a/kornia/augmentation/container/image.py
+++ b/kornia/augmentation/container/image.py
@@ -192,9 +192,7 @@ class ImageSequential(ImageSequentialBase, ImageModuleForSequentialMixIn):
 
         total_weight = multinomial_weights.sum()
         if total_weight == 0:
-            indices = torch.tensor(
-                [], device=multinomial_weights.device, dtype=torch.long
-            )
+            indices = torch.tensor([], device=multinomial_weights.device, dtype=torch.long)
         else:
             indices = torch.multinomial(
                 multinomial_weights,


Saksham-st12:
Just a quick follow-up: this PR fixes the crash described in #3300 and includes a regression test.

Some CI failures appear unrelated and repo-wide; happy to rebase or adjust if needed.

tests/augmentation/container/test_augmentation_sequential.py
```diff
@@ -59,6 +59,23 @@ def test_mixup(self, inp, random_apply, same_on_batch, device, dtype):
         assert out.shape[-3:] == inp.shape[-3:]
         reproducibility_test(inp, aug)
 
+    def test_mixup_cutmix_only(self, device, dtype):
+        # Regression test for https://github.com/kornia/kornia/issues/3300
```
edgarriba: not a good idea to refer to issues in comments

kornia/augmentation/container/image.py
```diff
@@ -189,18 +189,25 @@ def get_random_forward_sequence(self, with_mix: bool = True) -> Tuple[Iterator[T
         mix_indices = self.get_mix_augmentation_indices(self.named_children())
         # kick out the mix augmentations
         multinomial_weights[mix_indices] = 0
-        indices = torch.multinomial(
-            multinomial_weights,
-            num_samples,
-            # enable replacement if non-mix augmentation is less than required
-            replacement=num_samples > multinomial_weights.sum().item(),
-        )
+        if (multinomial_weights.sum() == 0).all():
```
edgarriba: is there any other way to avoid sum here ? as it's a O(n)

Saksham-st12: One alternative would be to use torch.any(multinomial_weights) instead of summing, which avoids floating-point equality checks and makes the condition more explicit.

edgarriba: ok, i see. I think it's fine then with sum. The only thing i see is that eventuually sum is done in line 199  too. I force that sum happens only once

Saksham-st12: Done, I refactored the code to compute the weight sum once and reuse it for both the zero-check and the replacement condition.

edgarriba (dismissed):

Saksham-st12:
Re-signed the commit using SSH to satisfy the verified-signature requirement.

Saksham-st12:
Hi, I’m running into an issue with the “verified signature” requirement.
I’ve tried re-signing the latest commit using SSH signing, but GitHub still marks the commit as Unverified and prompts me to upload a signing SSH key.

Could you please confirm whether this repository requires GPG-only signatures, or if SSH-signed commits should be accepted here?
I’m happy to re-sign the commit in the preferred way once clarified.

Saksham-st12:
> Not all the suggestions seem to be applied

I’ve re-checked the latest commit and applied all the suggestions I’m aware of, including consolidating the weight sum computation and updating the zero-weight handling logic.

If there’s any specific suggestion that still appears missing, could you please point it out so I can address it directly?

Saksham-st12:
I want to sincerely apologize for the confusion caused by my earlier force-pushes and commit signing issues.

While resolving the verified-signature requirement, I unintentionally created noise in the commit history, which made it harder to track the originally agreed changes. That’s on me.

I’ve now rebased and squashed the PR so the current diff cleanly reflects only the changes we discussed (zero-weight guard, single mix handling, and the regression test), with a properly verified commit.

Thank you for the time and patience spent reviewing this.

Saksham-st12:
Hi @andrew-shc,

I’ve addressed the review feedback on this PR, rebased and squashed it to keep the diff minimal, and fixed earlier commit-signing issues.

This PR fixes a crash in AugmentationSequential when only mix augmentations are used and adds a small regression test.

If you have time, I’d really appreciate a quick review of the current state. Thanks!

Saksham-st12:
Hi @johnnv1 ,

I’ve rebased and squashed this PR to keep the diff minimal and addressed the earlier review feedback.

This PR fixes a crash in AugmentationSequential when only mix augmentations are used and adds a small regression test.

If you have time, I’d really appreciate a quick look at the current state. Thanks!

PR #3462, opened 2026-01-06
show link to code source in to docs

ducha-aiki:
## 📝 Description

In title


edgarriba:
did it work ?

ducha-aiki:
Yes, but not the way I intended it - it just added link to single code file in html, not to github itself. I will update it today

ducha-aiki:
@edgarriba now works
image

PR #3466, opened 2026-01-07
link to github in docs

ducha-aiki:
## 📝 Description
Better version of https://github.com/kornia/kornia/pull/3462 -- not just show the code, but the link to the github.

PR #3467, opened 2026-01-07
Link in docs AND on RTD?

ducha-aiki:
## 📝 Description

Previous one works locally

image

but somehow not on readthedocs.
image

 So I am trying to debug that in prod. Also apply the code review comment I missed from previous :)

https://github.com/kornia/kornia/pull/3466


ducha-aiki:
YES!
image

Issue #2752, opened 2024-01-23
[Tracker] Migrate test classes to use `BaseTester` and its functionalities.


johnnv1:
Following on from #2745

Our tests are grouped using classes for tests. In the past, we could not adopt our `BaseTester` everywhere, because it had some abstract methods which made it half annoying. With our kinda "new" `BaseTester` - [`from testing.base import BaseTester`](https://github.com/kornia/kornia/blob/ce434e467faf617604bb3383cf78cd0b79f59dbd/testing/base.py#L67) - we can adopt it everywhere. 

What needs to be done:

1. pickup a test file
2. add the `BaseTester` as the parent class of the test classes
3. update the `assert_close` to be `self.assert_close`
4. update the `test_gradcheck` to use `self.gradcheck` instead of `gradcheck` from `torch.autograd`
5. delete the use of `tensor_to_gradcheck_var` on the `test_gradcheck` - this is automatically handled by the `BaseTester` class
6. **extra**: remove the `dtype` argument of the `test_gradcheck` and just run it using `torch.float64`
7. open a PR :)

You can check #2745, which has done a bunch of these updates across our tests

The goal here is to reorganize our test suite. If you have any questions, you can call us and/or ask on our Slack page too [![Slack](https://img.shields.io/badge/Slack-4A154B?logo=slack&logoColor=white)](https://join.slack.com/t/kornia/shared_invite/zt-csobk21g-2AQRi~X9Uu6PLMuUZdvfjA)

RagS8i:
Hi! I’d like to work on this issue,
I’ve gone through the description and #2745 for reference. I will try to gradually go through different test files, apply changes and sequentially open PRs for them. Please do tell me any suggestions or advice you have.
Could you please assign it to me?

edgarriba:
sounds good!

Big-Marvel:
Hi @johnnv1 and @edgarriba  is this issue still active? If so, may I take it up?

edgarriba:
yes, it should be done module by module though

Big-Marvel:
Hi @edgarriba @johnnv1 
I have raised a PR that migrates tests/image/test_image.py to use BaseTester.
If this approach looks good, I’m happy to migrate the remaining modules as well.

Thanks!

amrorabea:
Hi @edgarriba and @Big-Marvel!

[...]

PR #3469, opened 2026-01-07
tests(geometry): migrate geometry module to use BaseTester.

Big-Marvel:
This PR migrates `tests/geometry` module to use `BaseTester` as part of the ongoing test standardization effort.

**Fixes/Relates to:** #2752 

## 🛠️ Changes Made
- [x] Migrated test classes of tests\geometry module to inherit from BaseTester for test standardization.

---

## 🧪 How Was This Tested?
- [x] **Unit Tests:** (List new/updated tests)
- [ ] **Manual Verification:** (Describe the steps you took)
- [ ] **Performance/Edge Cases:** (How does this handle nulls, large data, etc.?)

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [x] 🟢 **No AI used.**
- [ ] 🟡 **AI-assisted:** I used AI for boilerplate/refactoring but have manually reviewed and tested every line.

---

## 🚦 Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have added tests that prove my fix is effective or that my feature works.

### DIFF
--- a/tests/geometry/transform/test_thin_plate_spline.py
+++ b/tests/geometry/transform/test_thin_plate_spline.py
@@ -235,17 +235,17 @@ class TestWarpImage(BaseTester):
 
         opts = {"device": device, "dtype": torch.float64}
         src, dst = _sample_points(batch_size, **opts)
-        
+
         # Compute TPS params without tracking gradients
         with torch.no_grad():
             kernel, affine = kornia.geometry.transform.get_tps_transform(src, dst)
-    
+
         image = torch.rand(batch_size, 3, 32, 32, requires_grad=True, **opts)
-    
+
         assert self.gradcheck(
             kornia.geometry.transform.warp_image_tps,
             (image, dst, kernel, affine),
-            requires_grad=[True, False, False, False], 
+            requires_grad=[True, False, False, False],
             raise_exception=True,
             atol=1e-4,
             rtol=1e-4,


tests/geometry/transform/test_thin_plate_spline.py
```diff
@@ -227,11 +228,16 @@ def test_exception(self, batch_size, device, dtype):
     @pytest.mark.grad()
     @pytest.mark.parametrize("batch_size", [1, 3])
     def test_gradcheck(self, batch_size, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("gradcheck is unstable for warp_image_tps on CUDA")
```
edgarriba, line 232: can we investigate why, maybe open a ticket

edgarriba (approved):
LGTM, make ci happy

Big-Marvel:
Hi @edgarriba,
The grad check failure happens specifically for `warp_image_tps`, which internally uses
`torch.nn.functional.grid_sample.`
Since `grid_sample` has piecewise / interpolation-based gradients, finite-difference `gradcheck` seems numerically unstable here.

would you suggest 
- Skipping `gradcheck` for `warp_image_tps` entirely
- Passing dst as the centers for the radial basis functions,
currently the TPS formula is being evaluated using weights calculated for a completely different set of coordinates.
- Use a smooth image (like a linear ramp) for gradcheck.

Also , test are passing in the local environment but not in githubCI.

tests/geometry/transform/test_thin_plate_spline.py
Big-Marvel: I looked into this while debugging the `gradcheck` failures.
`warp_image_tps` internally relies on `torch.nn.functional.grid_sample`. On CUDA,
`grid_sample` uses different kernels and non-deterministic atomic paths, and its
gradients are known to be numerically unstable under finite-difference `gradcheck`.

I agree it would be good to document this properly. I can open a ticket to track CUDA `gradcheck` instability for `warp_image_tps` (and more generally `grid_sample-based ops).

Big-Marvel:
The persistent mismatch happens for input 2 (= TPS kernel params). 
Even on CPU+float64 and larger image size (changed it to 32x32), finite-difference gradcheck is unstable wrt TPS parameters because the warp depends on grid_sample and small parameter perturbations change interpolation neighborhoods (piecewise gradients).

I'd suggest to keep gradcheck for warp_image_tps but restrict it to checking gradients wrt image only (treat kernel/affine as constants), since that’s stable and still validates backward correctness.

let me know your thoughts on this @edgarriba

tests/geometry/transform/test_thin_plate_spline.py
edgarriba: sounds good. We might want also to think how to not skip this kind of test but instead just collect properly during the pytest call

Big-Marvel: I agree. Since this PR is already focused on migrating tests to `BaseTester`, I’ll keep it scoped and open a follow-up PR to handle pytest-side collection/config (so gradcheck tests run only on CPU jobs rather than skipping inside the test).

Issue #3455, opened 2026-01-04
[Feature]: Implement smolvlm2


edgarriba:
### 🚀 Feature Description

Include implementantion of Smolvlm2 in native pytorch

### 📂 Feature Category

VLM/VLA Models (Vision Language Models/Agents) - Priority

### 💡 Motivation

Bring more state of the art VLM into `kornia.models`

### 💭 Proposed Solution

Standalone pytorch module in `kornia.models.smolvlm2`
Tested against transformers

### 🤝 Contribution Intent

- [x] I plan to submit a PR to implement this feature
- [x] I'm requesting this feature but not planning to implement it

Saksham-st12:
I’d like to take ownership of this issue and implement SmolVLM2 in native PyTorch under kornia.models.smolvlm2.

My plan is to follow Kornia’s existing model patterns and build a standalone PyTorch implementation (vision encoder, multimodal projection, and fusion blocks) without a hard runtime dependency on transformers. Pretrained weights will be supported via a conversion/loading utility, and correctness will be validated by comparing outputs (logits / embeddings) against the reference transformers implementation within numerical tolerance. I’ll also add unit tests and minimal documentation/examples to align with other models in kornia.models.

I’ve been actively contributing to Kornia recently and am comfortable with the codebase and review process. If this approach looks reasonable to you, I’m happy to get started right away.

Thanks!

edgarriba:
please, propose a concise iteratively plan before submitting any PR written by you, no agents

Saksham-st12:
Thanks for the note. Before opening a PR, here is how I plan to proceed, step by step:

First, I plan to focus on familiarizing myself with the SmolVLM2 architecture itself as well as similar models so that it makes sense well within the existing codebase on kornia.models.

Then, I'll organize the model structure well inside the kornia.models.smolvlm2 folder and code the model using plain PyTorch code, keeping it simple by only including the forwarding functionality. This is to abide by the style of the existing code inside Kornia.

Once core code is ready, I plan to finish functionality for pre-trained weight import and check by validating numerical consistency by comparing predictions with reference Transformer’s code.

Next, I will write unit tests involving shapes, forward execution, and basic correctness, and finally some quick clean-up passes for readability and maintainability.

Lastly, I'll add some documentation comments and a code example for its usage, merge some commits if needed, and proceed to open a targeted Pull Request for code review.

Is that a reasonable workflow that I can move on to?

edgarriba:
[...]

PR #3475, opened 2026-01-09
Add SmolVLM2 model scaffold

Saksham-st12:
## 📝 Description

This PR adds an initial SmolVLM2 scaffold under kornia.models.smolvlm2.

It introduces the module structure and a minimal, runnable model class to validate wiring and imports. No pretrained weights, validation, tests, or full model logic are included in this PR by design. These will be added incrementally in follow-up PRs.

Relates to #3455.

### DIFF
--- a/kornia/models/smolvlm2/smolvlm2.py
+++ b/kornia/models/smolvlm2/smolvlm2.py
@@ -15,11 +15,13 @@
 # limitations under the License.
 #
 
-from torch import nn
 import torch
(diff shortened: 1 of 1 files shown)


Saksham-st12:
I’ve updated the scaffold to include at least a minimal forward pass with vision and text projections so the interfaces and wiring can be meaningfully exercised. If this is not what you intended for the incremental step, please do tell and I’ll modify.

PR #3540, opened 2026-01-31
[Feature]: Replace manual DETR box processing with Kornia BoundingBox API #3530

### LINKED ISSUE #3530: [Feature]: Replace manual DETR box processing with Kornia BoundingBox API
**Title:** [Refactor] Replace manual DETR box processing with Kornia BoundingBox API

### Description

The `forward` method in the DETR post-processor currently performs manual tensor arithmetic to convert bounding box coordinates and formats:

```python
# Manual conversion from cxcywh -> xywh
cxcy, wh = boxes[..., :2], boxes[..., 2:]
boxes_xy = torch.cat([cxcy - wh * 0.5, wh], -1)

```

There are `TODO` and `NOTE` comments explicitly suggesting the replacement of this manual logic with Kornia's `BoundingBox` API or geometry functional API.

Using raw tensor operations for geometry is error-prone and less readable. 

### Goal

Refactor the post-processing logic to use Kornia's geometry primitives.

ArnabiDutta:
Fixes #3530
## 📝 Description: 
This PR refactors the `DETRPostProcessor` to use the Kornia `Boxes` API instead of manual tensor arithmetic, as requested in the issue. To support this cleanly, it also adds native support for the `cxcywh` (Center-X, Center-Y, Width, Height) bounding box format to the core `kornia.geometry.boxes.Boxes` class.
**Fixes/Relates to:** #3530

## 🛠️ Changes Made
- [✓] **Core Geometry (`kornia/geometry/boxes.py`):** Added native support for `mode="cxcywh"` in `Boxes.from_tensor` and `Boxes.to_tensor`. This enables full round-trip conversion for Transformer-based detection models.
- [✓] **Model Refactor (`kornia/models/rt_detr/post_processor.py`):** Refactored the `forward` method to replace manual coordinate conversion with the `Boxes` API.
- [✓] **Safety:** Ensured correct order of operations (Scale -> Convert) to respect Kornia's pixel-coordinate conventions and removed fixed TODO comments.
---

## 🧪 How Was This Tested?
- [✓] **Unit Tests:** Verified locally that `Boxes.from_tensor(..., mode="cxcywh")` correctly converts normalized and pixel coordinates to `xyxy` and back.
- [✓] **Manual Verification:** Created a regression test script (`verify_pr.py`) comparing the output of the new API-based implementation against the legacy manual math implementation. Confirmed outputs match exactly (tolerance < 1e-6) for valid pixel coordinates.
- [✓] **Performance/Edge Cases:** Verified that `torch.compile` successfully traces the new implementation without graph breaks. Validated that sub-pixel noise is handled gracefully.
---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [] 🟢 **No AI used.**
- [✓] 🟡 **AI-assisted:** I used AI for boilerplate/refactoring but have manually reviewed and tested every line.

---

## 🚦 Checklist
- [✓] I am assigned to the linked issue (required before PR submission)
- [✓] The linked issue has been approved by a maintainer
- [✓] I have performed a **self-review** of my code (no "ghost" variables or hallucinations).
- [✓] My code follows the existing style guidelines of this project.
- [✓] I have commented my code, particularly in hard-to-understand areas.
- [✓] I have added tests that prove my fix is effective or that my feature works.

---

## 💭 Additional Context
The refactor improves readability and maintainability by centralizing the geometry logic. Adding `cxcywh` support to the core API will also benefit future Transformer-based models added to Kornia.
image
<details>
<summary><b>Open dropdown to see the verification script (verify_pr.py)</b></summary>

```python
import torch
import sys
from kornia.models.rt_detr.post_processor import DETRPostProcessor

def test_core_api_logic():
    print("\n[1/3] Testing Core 'Boxes' API Logic (Pixel Coordinates)...")
    cxcywh = torch.tensor([[50., 50., 20., 10.]])
    try:
        boxes = Boxes.from_tensor(cxcywh, mode="cxcywh")
        output = boxes.to_tensor(mode="xywh")
        expected = torch.tensor([[40., 45., 20., 10.]])
        assert torch.allclose(output, expected), f"Math Error! Got {output}, Expected {expected}"
        print("✅ PASS: Core API correctly handles 'cxcywh' <-> 'xywh' conversion.")
    except Exception as e:
        print(f"❌ FAIL: Core API crashed: {e}")
        raise e

def test_detr_parity():
    print("\n[2/3] Testing DETR PostProcessor Parity (Valid Boxes Only)...")
    
    # 1. Setup Random Data
    torch.manual_seed(42)
    num_classes = 4
    num_queries = 100 # More queries to ensure we get valid ones
    bs = 2
    
    logits = torch.randn(bs, num_queries, num_classes)
    boxes = torch.rand(bs, num_queries, 4) # Normalized 0-1
    original_sizes = torch.tensor([[100, 100], [200, 200]]).repeat(bs // 2 + 1, 1)[:bs]

    # 2. Run New Implementation
    processor = DETRPostProcessor(
        num_classes=num_classes, num_top_queries=num_queries, confidence_threshold=0.0
    )
    results_new = processor(logits, boxes, original_sizes)
    bboxes_new = results_new[..., 2:]

    # 3. Run Reference Implementation
    cxcy, wh = boxes[..., :2], boxes[..., 2:]
    boxes_xy_old = torch.cat([cxcy - wh * 0.5, wh], -1) 
    sizes_wh = original_sizes[0].flip(0).unsqueeze(0).unsqueeze(0).repeat(1, 1, 2)
    boxes_xy_old = boxes_xy_old * sizes_wh 
    
    # Replicate gathering
    scores = logits.sigmoid()
    _, index = torch.topk(scores.flatten(1), num_queries, dim=-1)
    index_div = index // num_classes
    bboxes_old = boxes_xy_old.gather(dim=1, index=index_div.unsqueeze(-1).repeat(1, 1, boxes_xy_old.shape[-1]))

    # 4. Filter & Compare
    # We only care about boxes with Width & Height >= 1.0 (Valid Pixels)
    # Boxes < 1.0 are sub-pixel noise and Kornia handles them differently than raw math.
    mask = (bboxes_old[..., 2] >= 1.0) & (bboxes_old[..., 3] >= 1.0)
    
    valid_new = bboxes_new[mask]
    valid_old = bboxes_old[mask]
    
    print(f"   Comparing {valid_old.shape[0]} valid boxes (w,h >= 1px)...")

    diff = (valid_new - valid_old).abs().max()
    print(f"   Max Difference on Valid Boxes: {diff:.8f}")
    
    if diff > 1e-4:
        print("❌ FAIL: Significant mismatch on VALID boxes!")
        raise AssertionError("Parity check failed.")
        
    print("✅ PASS: New implementation matches legacy logic for all valid objects.")

def test_torch_compile():
    print("\n[3/3] Testing torch.compile() Compatibility...")
    if not hasattr(torch, "compile"): return
    
    logits = torch.randn(1, 10, 4)
    boxes = torch.rand(1, 10, 4)
    original_sizes = torch.tensor([[100, 100]])
    processor = DETRPostProcessor(num_classes=4, num_top_queries=5)

    try:
        optimized_processor = torch.compile(processor, backend="inductor")
        _ = optimized_processor(logits, boxes, original_sizes)
        print("✅ PASS: Compilation successful.")
    except Exception as e:
        print(f"❌ FAIL: Compilation failed: {e}")
        raise e

if __name__ == "__main__":
    try:
        test_core_api_logic()
        test_detr_parity()
        test_torch_compile()
    except Exception as e:
        print("\nTESTS FAILED.")
        sys.exit(1)

### DIFF
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -462,6 +462,8 @@ class Boxes:
                 * 'vertices_plus': similar to 'vertices' mode but where box width and length are defined as
                   ``width = xmax - xmin + 1`` and ``height = ymax - ymin + 1``. ymin + 1``.
                   With shape :math:`(N, 4, 2)` or :math:`(B, N, 4, 2)`.
+                * 'cxcywh': boxes are assumed to be in the format ``x_center, y_center, width, height`` where
+                  ``width = xmax - xmin + 1`` and ``height = ymax - ymin + 1``. With shape :math:`(N, 4)`, :math:`(B, N, 4)`.
 
             validate_boxes: check if boxes are valid rectangles or not. Valid rectangles are those with width
                 and height >= 1 (>= 2 when mode ends with '_plus' suffix).
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -553,10 +553,10 @@ class Boxes:
         elif mode == "cxcywh":
             width = boxes[..., 2] - boxes[..., 0] + 1
             height = boxes[..., 3] - boxes[..., 1] + 1
-            
+
             cx = boxes[..., 0] + (width - 1) / 2
             cy = boxes[..., 1] + (height - 1) / 2
-            
+
             boxes[..., 0] = cx
             boxes[..., 1] = cy
             boxes[..., 2] = width
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -104,7 +104,7 @@ def _boxes_to_quadrilaterals(boxes: torch.Tensor, mode: str = "xyxy", validate_b
         batched = boxes.ndim == 4
         if not (3 <= boxes.ndim <= 4 and boxes.shape[-2:] == torch.Size([4, 2])):
             raise ValueError(f"Boxes shape must be (N, 4, 2) or (B, N, 4, 2) when {mode} mode. Got {boxes.shape}.")
-    elif mode.startswith("xy"):
+    elif mode.startswith("xy") or mode == "cxcywh":
         batched = boxes.ndim == 3
         if not (2 <= boxes.ndim <= 3 and boxes.shape[-1] == 4):
             raise ValueError(f"Boxes shape must be (N, 4) or (B, N, 4) when {mode} mode. Got {boxes.shape}.")
@@ -128,13 +128,15 @@ def _boxes_to_quadrilaterals(boxes: torch.Tensor, mode: str = "xyxy", validate_b
         else:
             raise ValueError(f"Unknown mode {mode}")
         not validate_boxes or validate_bbox(quadrilaterals)
-    elif mode.startswith("xy"):
+    elif mode.startswith("xy") or mode == "cxcywh":
         if mode == "xyxy":
             height, width = boxes[..., 3] - boxes[..., 1], boxes[..., 2] - boxes[..., 0]
         elif mode == "xyxy_plus":
             height, width = boxes[..., 3] - boxes[..., 1] + 1, boxes[..., 2] - boxes[..., 0] + 1
         elif mode == "xywh":
             height, width = boxes[..., 3], boxes[..., 2]
+        elif mode == "cxcywh":
+            width, height = boxes[..., 2], boxes[..., 3]
         else:
             raise ValueError(f"Unknown mode {mode}")
 
@@ -144,7 +146,11 @@ def _boxes_to_quadrilaterals(boxes: torch.Tensor, mode: str = "xyxy", validate_b
             if (height <= 0).any():
                 raise ValueError("Some boxes have negative heights or 0.")
 
-        xmin, ymin = boxes[..., 0], boxes[..., 1]
+        if mode == "cxcywh":
+            xmin = boxes[..., 0] - 0.5 * width
+            ymin = boxes[..., 1] - 0.5 * height
+        else:
+            xmin, ymin = boxes[..., 0], boxes[..., 1]
         quadrilaterals = _boxes_to_polygons(xmin, ymin, width, height)
     else:
         raise ValueError(f"Unknown mode {mode}")
@@ -507,6 +513,8 @@ class Boxes:
                   box width and height are defined as ``width = xmax - xmin`` and ``height = ymax - ymin``.
                 * 'vertices_plus': similar to 'vertices' mode but where box width and length are defined as
                   ``width = xmax - xmin + 1`` and ``height = ymax - ymin + 1``. ymin + 1``.
+                * 'cxcywh': boxes are defined as ``x_center, y_center, width, height`` where ``width = xmax - xmin + 1``
+                  and ``height = ymax - ymin + 1``.
             as_padded_sequence: whether to keep the pads for a list of boxes. This parameter is only valid
                 if the boxes are from a box list whilst `from_tensor`.
 
@@ -540,6 +548,17 @@ class Boxes:
             pass
         elif mode in ("xywh", "vertices", "vertices_plus"):
             height, width = boxes[..., 3] - boxes[..., 1] + 1, boxes[..., 2] - boxes[..., 0] + 1
+            boxes[..., 2] = width
+            boxes[..., 3] = height
+        elif mode == "cxcywh":
+            width = boxes[..., 2] - boxes[..., 0] + 1
+            height = boxes[..., 3] - boxes[..., 1] + 1
+            
+            cx = boxes[..., 0] + (width - 1) / 2
+            cy = boxes[..., 1] + (height - 1) / 2
+            
+            boxes[..., 0] = cx
+            boxes[..., 1] = cy
             boxes[..., 2] = width
             boxes[..., 3] = height
         else:
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -462,8 +462,9 @@ class Boxes:
                 * 'vertices_plus': similar to 'vertices' mode but where box width and length are defined as
                   ``width = xmax - xmin + 1`` and ``height = ymax - ymin + 1``. ymin + 1``.
                   With shape :math:`(N, 4, 2)` or :math:`(B, N, 4, 2)`.
-                * 'cxcywh': boxes are assumed to be in the format ``x_center, y_center, width, height`` where
-                  ``width = xmax - xmin + 1`` and ``height = ymax - ymin + 1``. With shape :math:`(N, 4)`, :math:`(B, N, 4)`.
+                * * 'cxcywh': boxes are assumed to be in the format ``x_center, y_center, width, height`` where
+                  ``width = xmax - xmin + 1`` and ``height = ymax - ymin + 1``.
+                  With shape :math:`(N, 4)`, :math:`(B, N, 4)`.
 
             validate_boxes: check if boxes are valid rectangles or not. Valid rectangles are those with width
                 and height >= 1 (>= 2 when mode ends with '_plus' suffix).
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by ayush585 — CHANGES_REQUESTED** [review]
Nice cleanup switching to Boxes..
Two quick nits:
Double scale? We do boxes_scaled = boxes * sizes_wh, then later boxes_xy = boxes_xy * sizes_wh again, should the 2nd multiply be removed?

cxcywh semantics: to_tensor("cxcywh") uses the +1 / inclusive math, but _boxes_to_quadrilaterals treats cxcywh like the normal center format. 
Should cxcywh be nonplus (no +1) or renamed to cxcywh_plus?

**Comment by ArnabiDutta** [comment]
Hello @ayush585,

Thank you for the detailed review. I'd like to address your concerns about the `cxcywh` implementation and clarify the fixed approach.

## Fixed Implementation: Non-Inclusive (Standard) Semantics

I've implemented `cxcywh` mode using **standard, non-inclusive semantics** to match DETR's expectations and be consistent with how other coordinate systems work:

**In `_boxes_to_quadrilaterals` (from_tensor):**
```python
elif mode == "cxcywh":
    width, height = boxes[..., 2], boxes[..., 3]
    xmin = boxes[..., 0] - width / 2.0      # Standard center-to-corner formula
    ymin = boxes[..., 1] - height / 2.0
    quadrilaterals = _boxes_to_polygons(xmin, ymin, width, height)
```

**In `to_tensor`:**
```python
elif mode == "cxcywh":
    width = boxes[..., 2] - boxes[..., 0] + 1   # Extract from internal representation
    height = boxes[..., 3] - boxes[..., 1] + 1
    
    cx = boxes[..., 0] + width / 2.0            # Standard corner-to-center formula
    cy = boxes[..., 1] + height / 2.0
    
    boxes[..., 0] = cx
    boxes[..., 1] = cy
    boxes[..., 2] = width
    boxes[..., 3] = height
```

**Important clarification:** I hope the `+1` in `width = boxes[..., 2] - boxes[..., 0] + 1` is **not** making this inclusive semantics. This is simply extracting the width from Kornia's internal polygon representation (which uses the `width-1` offset in `_boxes_to_polygons`). 

The key difference from inclusive semantics is in the center calculation that i fixed: I'm using `width / 2.0` (standard) instead of `(width - 1) / 2.0` (inclusive), which was previously incorrect, thanks for pointing it out. This makes `cxcywh` use standard geometric semantics where the center is simply the midpoint between corners.

Is this the correct interpretation? Should `cxcywh` remain as standard/geometric semantics, or would you prefer a `cxcywh_plus` variant that uses inclusive semantics with the `(width-1)/2` center calculation?

## Question About Sub-Pixel Box Support

During testing, I encountered an issue with boxes that have `width < 1.0` (normalized or sub-pixel coordinates). These fail because `_boxes_to_polygons` uses the formula:

```python
polygons[..., 1, 0] += width - 1  # For width=0.2, this gives -0.8
```

This creates invalid polygons when width < 1.0.

**Example:**
- Input: `cx=0.5, cy=0.5, width=0.2, height=0.2` (normalized box)
- Expected output after round-trip: `cx=0.5, cy=0.5, width=0.2, height=0.2`
- Actual output: `cx=0.5, cy=0.5, width=1.8, height=1.8` (distorted)

The issue is that `_boxes_to_polygons` was designed assuming `width >= 1.0` (pixel coordinates with the inclusive convention).

**Questions:**
1. Is sub-pixel box support (width < 1.0) a requirement for Kornia's Boxes API? 
2. If not intentionally supported, should we document this limitation?
3. If it should be supported, would you prefer:
   - Special handling to detect and adjust sub-pixel boxes
   - Creating quadrilaterals directly for `cxcywh` mode without using `_boxes_to_polygons`
   - Documenting that `cxcywh` mode is only for pixel coordinates (width >= 1.0)

## Regarding the Double Scaling Question

You asked about potential double scaling in the post-processor. The current implementation scales once before passing to the Boxes API:

```python
sizes_wh = original_sizes[0].flip(0).unsqueeze(0).unsqueeze(0).repeat(1, 1, 2)
boxes_scaled = boxes * sizes_wh  # Scale from normalized [0,1] to pixels
boxes_xy = Boxes.from_tensor(boxes_scaled, mode="cxcywh").to_tensor(mode="xywh")
```

ayush585 (changes requested):
Thanks, scaling looks fixed (only normalized pixel once)
For cxcywh: right now it still behaves like a “plus/pixel” mode under the hood (_boxes_to_polygons uses width-1 and to_tensor width/height uses +1)

That’s why w/h < 1 gets distorted. I think we should either:
rename current behavior to cxcywh_plus, or
make cxcywh truly continuous end to end (no +1 / width-1 coupling likely needs a separate polygon path)
Also can we add a tiny round trip test with fractional cxcywh + a couple edge cases near image borders to lock semantics?

ArnabiDutta:
Thanks for the feedback @ayush585 ! I want to clarify something about the `+1` usage.

Looking at the existing `to_tensor` implementation, **all modes extract width/height using `+1`** from Kornia's internal polygon representation:

```python
if mode in ("xyxy", "xyxy_plus"):
    pass  # Already in correct format
elif mode in ("xywh", "vertices", "vertices_plus"):
    height, width = boxes[..., 3] - boxes[..., 1] + 1, boxes[..., 2] - boxes[..., 0] + 1
    boxes[..., 2] = width
    boxes[..., 3] = height
elif mode == "cxcywh":
    width = boxes[..., 2] - boxes[..., 0] + 1  # Same as xywh
    height = boxes[..., 3] - boxes[..., 1] + 1
    
    cx = boxes[..., 0] + width / 2.0
    cy = boxes[..., 1] + height / 2.0
```

**The `+1` here is not defining the mode's semantics** - it's simply reversing the internal polygon representation (which uses `width-1` offsets in `_boxes_to_polygons`). This is standard across **all modes** including `xywh`, `vertices`, `vertices_plus`.

The distinction between standard and `_plus` modes is in the **input/output contract**, not the internal extraction:
- `xyxy` expects `width = xmax - xmin` (adds `+1` offset on output)
- `xyxy_plus` expects `width = xmax - xmin + 1` (no offset added)

Similarly:
- `xywh` takes width directly (no `_plus` variant exists, but internally uses `+1` extraction)
- `cxcywh` takes width directly with center coordinates (also uses `+1` extraction, consistent with `xywh`)

**Regarding sub-pixel boxes (width < 1.0):** This appears to be a limitation of `_boxes_to_polygons` that affects all modes, not just `cxcywh`. If this is a requirement, I can implement direct quadrilateral creation for `cxcywh` to handle normalized coordinates properly.

Would appreciate maintainer input (@edgarriba @sidd-27 @cjpurackal ) on whether:
1. The current `cxcywh` implementation is consistent with existing mode semantics
2. Sub-pixel box support is needed for this mode

edgarriba (commented):
nice — using `Boxes.from_tensor` with `cxcywh` mode is cleaner than the manual conversion. couple things:

1. in `to_tensor` for `cxcywh` mode, the width/height calculation uses `+1`:
```python
width = boxes[..., 2] - boxes[..., 0] + 1
height = boxes[..., 3] - boxes[..., 1] + 1
```
but `from_tensor` for `cxcywh` uses standard `width = xmax - xmin` (no +1). this inconsistency will cause subtle bugs. the internal representation in `Boxes` is quadrilateral/xyxy-like, so you need to decide: does cxcywh follow the `_plus` convention or not? i think it should NOT have the +1. fix this.

2. the post_processor change looks correct — going from manual `cxcy - wh * 0.5` to `Boxes.from_tensor(mode="cxcywh").to_tensor(mode="xywh")` is the right abstraction.

3. did you verify this doesn't break the DETR inference numerically? the `+1` issue above could silently shift predictions by a pixel.

edgarriba (commented):
The refactor direction is exactly what #3530 asked for, and the DETR path is numerically equivalent to the old manual math — nice. Two things still block it:

1. The `cxcywh` convention question raised by both reviewers: as implemented it follows the inclusive `_plus`-style internal convention (`x2 - x1 + 1`), so `cxcywh → xyxy` is off by 1px vs continuous expectations. Either rename to `cxcywh_plus` or make it truly continuous — this is public-API semantics, so it needs an explicit decision.
2. The requested round-trip unit tests (fractional coords, border boxes, `cxcywh ↔ xyxy`) to lock those semantics in.

If you're unable to continue, just say so and a maintainer can carry it over the line — it's a small, wanted change and most of the work is done.

PR #3520, opened 2026-01-21
test : add missing exception and dynamo tests for FaceDetector

kirandharmar867-maker:
This PR adds the missing exception handling and torch.compile (Dynamo) tests
for FaceDetector, completing the BaseTester coverage as discussed in #3427.

All tests pass locally with --runslow.
closes #3427
Thank You!.

PR #3565, opened 2026-02-16
Issue: LightGlue crashes when adaptive width pruning removes all keypoints (#3564).

### LINKED ISSUE #3564: [Bug]: Lightglue crashes when adaptive width pruning removes all keypoints
### 🐛 Describe the bug

I encountered this issue while I was working with XFeat which uses Kornia's implementation of LightGlue on the background. 
When instantiating the LightGlue matcher with adaptive depth disabled (depth_confidence = -1) but adaptive width / point pruning enabled (width_confidence > 0), calling LightGlue.forward() on keypoints and descriptors extracted from highly distinct images can lead to a crash.

Instead of returning zero matches, the iterative point pruning step may prune all the points from one step to another. This results in an descriptor tensor with an empty dimension (e.g., shape [B, 0, D]). This empty tensor makes the filter_matches() function throw an IndexError.

I include here a ZIP file with data to reproduce this issue.

[data.zip](https://github.com/user-attachments/files/25268485/data.zip)

### 🔄 Steps to Reproduce

```bash
1. Download and extract the attached data.zip file (contains a pt file with a set of XFeat keypoints and descriptors that reproduce the issue).
2. Fill in the path_to_data variable with the path to the uncompressed pt file.
3. Run the provided script.
4. IndexError is raised inside filter_matches()
```

### 💻 Minimal Code Example

```python
from kornia.feature.lightglue import LightGlue
import torch

path_to_data = LOCATION_OF_DOWNLOADED_PT_FILE

device = torch.device("cuda" if torch.cuda.is_available() else "cpu")

data = torch.load(path_to_data, map_location=device)

default_conf_xfeat = {
    "name": "xfeat",  # just for interfacing
    "input_dim": 64,  # input descriptor dimension (autoselected from weights)
    "descriptor_dim": 96,
    "add_scale_ori": False,
    "add_laf": False,  # for KeyNetAffNetHardNet
    "scale_coef": 1.0,  # to compensate for the SIFT scale bigger than KeyNet
    "n_layers": 6,
    "num_heads": 1,
    "flash": True,  # enable FlashAttention if available.
    "mp": False,  # enable mixed precision
    "depth_confidence": -1,  # early stopping, disable with -1
    "width_confidence": 0.95,  # point pruning, disable with -1
    "filter_threshold": 0.1,  # match threshold
    "weights": None,
}
LightGlue.default_conf = default_conf_xfeat
net = LightGlue(None)

state_dict = torch.hub.load_state_dict_from_url(
                "https://github.com/verlab/accelerated_features/raw/main/weights/xfeat-lighterglue.pt"
            )

for i in range(net.conf.n_layers):
    pattern = f"self_attn.{i}", f"transformers.{i}.self_attn"
    state_dict = {k.replace(*pattern): v for k, v in state_dict.items()}
    pattern = f"cross_attn.{i}", f"transformers.{i}.cross_attn"
    state_dict = {k.replace(*pattern): v for k, v in state_dict.items()}
    state_dict = {k.replace("matcher.", ""): v for k, v in state_dict.items()}

net.load_state_dict(state_dict, strict=False)
net.to(device)
net.eval()
with torch.inference_mode():
    out = net(data) # IndexError: max(): Expected reduction dim 2 to have non-zero size.
```

### ✅ Expected behavior

When calling LightGlue.forward() on features extracted from non-matching images, the matcher should run without raising any error. If all keypoints are pruned, LightGlue.forward() result should still be a valid output with empty matches.

### ❌ Actual behavior

Calling LightGlue.forward() on keypoints and descriptors that get internally fully pruned leading to an IndexError. 

Error message:
```
File /.../python3.13/site-packages/kornia/feature/lightglue.py:630, in LightGlue._forward(self, data)
    628 desc0, desc1 = desc0[..., :m, :], desc1[..., :n, :]
    629 scores, _ = self.log_assignment[i](desc0, desc1)
--> [630](https://.../lib/python3.13/site-packages/kornia/feature/lightglue.py:630) m0, m1, mscores0, mscores1 = filter_matches(scores, self.conf.filter_threshold)
    631 matches, mscores = [], []
    632 for k in range(b):

File /.../python3.13/site-packages/kornia/feature/lightglue.py:321, in filter_matches(scores, th)
    319 def filter_matches(scores: Tensor, th: float) -> Tuple[Tensor, Tensor, Tensor, Tensor]:
    320     """Obtain matches from a log assignment matrix [Bx M+1 x N+1]."""
--> [321](https://.../lib/python3.13/site-packages/kornia/feature/lightglue.py:321)     max0, max1 = scores[:, :-1, :-1].max(2), scores[:, :-1, :-1].max(1)
    322     m0, m1 = max0.indices, max1.indices
    323     indices0 = arange(m0.shape[1], device=m0.device)[None]

IndexError: max(): Expected reduction dim 2 to have non-zero size.
```

### 🔧 Environment

```shell
- Kornia version:0.8.2
- PyTorch version:2.9.1
- Python version:3.13.11
- OS (e.g., Linux, macOS, Windows): Linux
- Installation method (pip, conda, from source): conda
- CUDA/cuDNN version (if applicable):12.8
- GPU model (if applicable):NVIDIA GeForce RTX 5070 Ti
```

### 📝 Additional context

Error doesn't happen everytime LightGlue is called on unmatchable kps+descriptors, it happens when point pruning is activated, either desc0.shape[-2] or desc1.shape[-2] >pruning_th and keep0 or keep1 result in an empty tensor, which makes the corresponding desc0 or desc1 a tensor with empty dimension. Here is the commented code snippet of where the bug happens:
```python
for i in range(self.conf.n_layers):
        desc0, desc1 = self.transformers[i](desc0, desc1, encoding0, encoding1, mask0=mask0, mask1=mask1)
        if i == self.conf.n_layers - 1:
            continue  # no early stopping or adaptive width at last layer
        
        if do_early_stop:
            token0, token1 = self.token_confidence[i](desc0, desc1)
            if self.check_if_stop(token0[..., :m, :], token1[..., :n, :], i, m + n):
                break
        if do_point_pruning and desc0.shape[-2] > pruning_th:
            scores0 = self.log_assignment[i].get_matchability(desc0)
            prunemask0 = self.get_pruning_mask(token0, scores0, i)  # type: ignore
            keep0 = where(prunemask0)[1] # if keep0.shape[0] == 0 Index error will occur when calling filter_matches
            ind0 = ind0.index_select(1, keep0)
            desc0 = desc0.index_select(1, keep0)
            encoding0 = encoding0.index_select(-2, keep0)
            prune0[:, ind0] += 1
        if do_point_pruning and desc1.shape[-2] > pruning_th:
            scores1 = self.log_assignment[i].get_matchability(desc1)
            prunemask1 = self.get_pruning_mask(token1, scores1, i)  # type: ignore
            keep1 = where(prunemask1)[1] # if keep1.shape[0] == 0 Index error will occur when calling filter_matches
            ind1 = ind1.index_select(1, keep1)
            desc1 = desc1.index_select(1, keep1)
            encoding1 = encoding1.index_select(-2, keep1)
            prune1[:, ind1] += 1
desc0, desc1 = desc0[..., :m, :], desc1[..., :n, :]
scores, _ = image_matcher.xfeat.lighterglue.net.log_assignment[i](desc0, desc1)
m0, m1, mscores0, mscores1 = filter_matches(
    scores, image_matcher.xfeat.lighterglue.net.conf.filter_threshold
)
```
Disabling width pruning (width_confidence = -1) obviously prevents the crash. I still believe a guard is needed to either a) prevent pruning from reducing a side to zero descriptors, or b) handle the empty-descriptor case before calling filter_matches().

### 🤝 Contribution Intent

kirandharmar867-maker:
Root Cause: 
When adaptive width pruning removes all keypoints or descriptors, tensors become empty. Later, filter_matches() and indexing operations receive an empty descriptor tensor, leading to an IndexError.

Changes: 
I added a safety guard in LightGlue._forward() to check for empty descriptor tensors after pruning. If the descriptors are empty, it returns valid empty outputs (matches0, matches1, matching_scores0, matching_scores1) instead of crashing. This stops filter_matches() from receiving invalid empty tensors.

Impact: 
This change prevents runtime crashes (IndexError) in LightGlue. It improves stability for edge cases where pruning removes all points. There is no change in behavior for normal cases; it only affects the empty-descriptor situation.

Files Modified: 
kornia/feature/lightglue.py tests/...

Fixes: 
Fixes #3564

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,6 +97,9 @@ skip = "*.bib,*.ipynb"
 target-version = "py311"
 line-length = 120
 
+[tool.ruff.format]
+skip-magic-trailing-comma = false
+
 [tool.ruff.lint]
 select = [
   "AIR",  # Airflow
@@ -162,7 +165,6 @@ ignore = [
   "RUF007",  # Prefer `itertools.pairwise()` over `zip()` - style preference
   "PLR0915",  # Allow condition check in list comprehension
   "PLC0415",  # `import` should be at the top-level of a file
-  "UP042", # Keep str, Enum for Python <3.11 compatibility
   "PLW2901",  # Allow overwritten values on loops
   "PLW1641",  # Object does not implement `__hash__` method
   "UP007",  # Prefer Optional[], Union[] over | due to torch jit scripting


edgarriba (commented):
couple things:

1. the fix itself makes sense — guarding against empty descriptors after pruning is the right approach. but i notice the `desc0, desc1 = desc0[..., :m, :], desc1[..., :n, :]` line changed indentation — was it previously outside the loop? if so that's a behavior change beyond the fix, can you confirm this is intentional?

2. the face detection tests (test_exception, test_dynamo) look like they belong to #3520, not this PR. can you split those out? this PR should only contain the lightglue fix.

3. where are the tests for the actual lightglue crash fix? you need a test that triggers the empty-descriptor path — create a case with very few keypoints that get fully pruned. without a regression test this could break again silently.

@kirandharmar867-maker also check if #3520 needs rebasing since you're touching the same file here.

kirandharmar867-maker:
> couple things:
> 
> 1. the fix itself makes sense — guarding against empty descriptors after pruning is the right approach. but i notice the `desc0, desc1 = desc0[..., :m, :], desc1[..., :n, :]` line changed indentation — was it previously outside the loop? if so that's a behavior change beyond the fix, can you confirm this is intentional?
> 2. the face detection tests (test_exception, test_dynamo) look like they belong to #3520, not this PR. can you split those out? this PR should only contain the lightglue fix.
> 3. where are the tests for the actual lightglue crash fix? you need a test that triggers the empty-descriptor path — create a case with very few keypoints that get fully pruned. without a regression test this could break again silently.
> 
> @kirandharmar867-maker also check if #3520 needs rebasing since you're touching the same file here.

Hi @edgarriba,
Thank you for detailed review.

1. Regarding the indentation: yes, the placement of:

desc0, desc1 = desc0[..., :m, :], desc1[..., :n, :]
inside the loop is intentional. The goal is to ensure that after each pruning step we Correctly clamp the descriptors to the original Keypoint count before checking for empty-descriptor condition. This should not change the intended behavior beyond the fix, but I’m happy to adjust if you think it should remain outside the loop.

2. I have split the unrelated FaceDetector tests into a separate change, so this PR now only contains the LightGlue fix.
I’ve added a regression test that explicitly triggers the empty-descriptor path(using zero keypoints) to ensure the crash does not reappear silently in the future(in a new test_file).

pls let me know if you’d prefer a different test scenario instead of zero-length input.
Thanks again for the guidance.

kirandharmar867-maker:
Hii @edgarriba,
Review Request Raised,
If You Have Time, Can You Please Review This.
Thank You!

kirandharmar867-maker:
Hlo @edgarriba @sidd-27 
Message: #3520 Rebased!, Waiting for your Review.....

edgarriba:
Thanks for the fix @kirandharmar867-maker, the direction is right — guarding against empty descriptors is the correct approach.

A few things to address before merge:

1. **Behavioral regression**: the original `desc0, desc1 = desc0[..., :m, :], desc1[..., :n, :]` line was unconditional (outside the `if self.conf.width_confidence > 0:` block). Your change moves it *inside* that block, meaning when `width_confidence == 0` the slicing is skipped entirely. This is a silent regression. The empty guard (`if desc0.shape[-2] == 0`) should be placed at the top of `_forward`, *before* the main loop, to handle the case where the input itself starts empty. The descriptor slicing must remain unconditional.

2. **pre-commit failure**: please run `pre-commit run --all-files` locally and fix the formatting issues before requesting final review.

3. The test in `test_lightglue.py` looks reasonable, but consider also adding a test where keypoints start non-empty and are pruned to zero mid-loop, to cover the original crash scenario from issue #3564.

Once those are addressed, @sidd-27 / @cjpurackal can you give final approval?

kirandharmar867-maker:
Hi @edgarriba ,
Issue #3564 appears to be resolved and all relevant tests are passing locally.

The tutorials job is failing on Python 3.12 due to a kornia_rs.read_image_any() size mismatch. I noticed that the scheduled CPU workflow on the main branch is also failing, so this seems unrelated to my changes    (which only update the Ruff configuration).

I Think now it's ready to merge..
Please let me know if you'd like me to investigate further.
Thank you @edgarriba.

kirandharmar867-maker:
hii @edgarriba,
can you pls review this PR again Last time.
I think this is solved Now 100%.

edgarriba:
LGTM — @sidd-27 can you merge? Thanks!

kirandharmar867-maker:
Hey @sidd-27 
Can you please murge it,
All related tests are passing with visual proof (see Discord)
Thanks @edgarriba, @sidd-27

kirandharmar867-maker:
Hlo @sidd-27 can You please murge this!,
Thank You.

shijianjian:
Hi @kirandharmar867-maker, can you fix the conflicts, then I will merge?

PR #3584, opened 2026-02-23
fix: handle empty keypoints after adaptive width pruning in LightGlue

### LINKED ISSUE #3564: [Bug]: Lightglue crashes when adaptive width pruning removes all keypoints

tysoncung:
## Problem

When using LightGlue with adaptive depth disabled (`depth_confidence = -1`) but adaptive width/point pruning enabled (`width_confidence > 0`), the iterative pruning step can remove **all** keypoints from one side when processing highly distinct images. This produces a descriptor tensor with an empty dimension (e.g., `[B, 0, D]`), which causes `filter_matches()` to throw an `IndexError` when calling `.max()` on the empty slice.

## Fix

Added an early-return check after the transformer/pruning loop: if either `desc0` or `desc1` has zero keypoints remaining, return empty matches with the same output structure as the normal path. This ensures callers always receive a valid result dict instead of a crash.

## Testing

Verified against the reproduction case from the issue (XFeat keypoints on highly distinct images). The fix returns zero matches gracefully instead of crashing.

Fixes #3564


edgarriba:
Thanks for the fix @tysoncung! The logic is sound — early-returning empty matches when adaptive width pruning zeroes out all keypoints is the right approach, and the output dict structure matches the normal path correctly.

However, CI is not fully green yet:

- ❌ **`tutorials / tutorials (3.12)`** — failing ([details](https://github.com/kornia/kornia/actions/runs/22294506300/job/64488341533))
- ❌ **`collector`** — depends on tutorials, also failing
- ❌ **`pre-commit.ci`** — pre-commit checks failing ([details](https://results.pre-commit.ci/run/github/145693916/1771825777.le6YRnJlR3aUp5AZUInc6g))

Could you check:
1. **pre-commit**: Run `pre-commit run --all-files` locally and push a fixup. This is often just formatting/linting.
2. **Tutorials**: If the tutorial failure is pre-existing/unrelated to your changes, please note that in the PR description so reviewers can confirm it's not a regression.

All unit + integration tests across Ubuntu/macOS/Windows × Python 3.11–3.13 × PyTorch 2.5+2.9 are ✅ — that's the important part. Once CI is clean, this should be good to merge.

— Clawd (Edgar's delegate)

edgarriba:
Hey @sidd-27 @cjpurackal — once the pre-commit failures are fixed, can you do the final review and merge? I'll be less available for the next bit. Thanks 🙏

sidd-27 (commented):
## Review: PR #3584 — fix: handle empty keypoints after adaptive width pruning in LightGlue

Good defensive fix for a real crash scenario.

---

### 🟡 Minor Issues

**1. Unrelated change: `config.py` StrEnum migration**
```python
-class InstallationMode(str, Enum):
+class InstallationMode(StrEnum):
```
This is the same change from PR #3605. Having it in two PRs will cause a merge conflict — whichever merges second will need a rebase. Recommend removing this from one of the two PRs.

**2. `prune0`/`prune1` fallback in the early return**
```python
"prune0": prune0 if do_point_pruning else torch.ones_like(torch.arange(m, device=device)[None]),
```
`torch.ones_like(torch.arange(m, device=device)[None])` creates a tensor of ones with dtype `torch.long`. If `prune0` is typically float or has a different shape, this could be inconsistent. Consider matching the dtype/shape of the normal return path's prune tensors.

Also, `torch.ones(1, m, device=device, dtype=torch.long)` would be simpler than `torch.ones_like(torch.arange(...))`.

**3. No test added**
The fix is for a specific crash case (empty keypoints after pruning). A test that exercises this path would prevent regressions — even a simple one that mocks the post-pruning state with empty descriptors.

---

### ✅ What's Good

- Correctly checks `desc0.shape[-2] == 0 or desc1.shape[-2] == 0` — right dimension
- Return structure matches the normal output format (matches, scores, m0, m1, etc.)
- Uses `-1` for unmatched entries in `m0_`/`m1_` — consistent with LightGlue convention
- Placement is correct — right after the pruning loop, before `log_assignment`

---

**Verdict:** Fix is correct but needs: (1) remove the duplicate `config.py` change, (2) add a regression test. The LightGlue crash was real and this addresses it properly.

PR #3594, opened 2026-02-26
fix: handle empty batch in median_blur

### LINKED ISSUE #3580: [Bug]: median_blur crashes on empty batch (batch_size=0) while all other filters handle it correctly
### 🐛 Describe the bug

`kornia.filters.median_blur` crashes with a `RuntimeError` when given an empty batch tensor (batch_size=0), while all other blur/filter functions in the same module handle this case correctly by returning an empty tensor with the correct shape.

**Root cause:** In `kornia/filters/median.py`, line 68:

    features = features.view(b, c, -1, h, w)

When `b=0`, the tensor has 0 elements. PyTorch cannot infer the `-1` dimension because `0 * c * x * h * w = 0` holds for any value of `x`, making it ambiguous.

**Consistency issue — all sibling filters work fine:**

| Function | Input (0,1,64,64) | Result |
|---|---|---|
| `gaussian_blur2d` | ✅ | `(0,1,64,64)` |
| `box_blur` | ✅ | `(0,1,64,64)` |
| `bilateral_blur` | ✅ | `(0,1,64,64)` |
| `unsharp_mask` | ✅ | `(0,1,64,64)` |
| `sobel` | ✅ | `(0,1,64,64)` |
| `laplacian` | ✅ | `(0,1,64,64)` |
| **`median_blur`** | **❌ RuntimeError** | **crash** |

This is the same class of bug that PyTorch itself tracked and fixed across `torch.nn` modules — see [pytorch/pytorch#12013](https://github.com/pytorch/pytorch/issues/12013) and [pytorch/pytorch#38115](https://github.com/pytorch/pytorch/issues/38115).

### 🔄 Steps to Reproduce

```bash
1. Create an empty batch tensor: `torch.empty(0, 1, 64, 64)`
2. Call `kornia.filters.median_blur(tensor, (3, 3))`
3. Observe RuntimeError crash
```

### 💻 Minimal Code Example

```python
import torch
import kornia


# All other filters handle empty batch correctly:
empty = torch.empty(0, 1, 64, 64)
print(kornia.filters.gaussian_blur2d(empty, (3, 3), (1.0, 1.0)).shape)  # torch.Size([0, 1, 64, 64]) ✅
print(kornia.filters.box_blur(empty, (3, 3)).shape)                      # torch.Size([0, 1, 64, 64]) ✅


# median_blur crashes:
kornia.filters.median_blur(empty, (3, 3))  # ❌ RuntimeError
```

### ✅ Expected behavior

`median_blur` should return an empty tensor with the same shape `(0, 1, 64, 64)`, consistent with all other filter functions in `kornia.filters` and with PyTorch's convention that all operations should accept 0-batch tensors (pytorch/pytorch#12013).

### ❌ Actual behavior

RuntimeError: cannot reshape tensor of 0 elements into shape [0, 1, -1, 64, 64] because the unspecified dimension size -1 can be any value and is ambiguous

Full traceback:
  File "kornia/filters/median.py", line 68, in median_blur
    features = features.view(b, c, -1, h, w)

### 🔧 Environment

```shell
- Kornia version: 0.8.2
- PyTorch version: 2.10.0+cpu
- Python version: 3.12.3
- OS: Ubuntu 24.04 (Linux)
- Installation method: pip
- CUDA/cuDNN version: N/A (CPU)
- GPU model: N/A
```

### 📝 Additional context

**Proposed fix:** Replace the ambiguous `-1` in `features.view(b, c, -1, h, w)` with the explicit kernel product `ky * kx`. The `_unpack_2d_ks` utility is already imported in this file. This is a one-line change that makes the reshape unambiguous for any batch size, including 0.

**PyTorch reference for empty batch handling:**
- pytorch/pytorch#12013 — Original feature request for 0-batch support
- pytorch/pytorch#38115 — Tracking issue for all nn modules to accept 0-batch tensors
- pytorch/pytorch#36530 — SyncBatchNorm had the exact same `view(0, c, -1, h, w)` crash

### 🤝 Contribution Intent

lweyrich1:
Closes #3580

### Problem
`median_blur` crashes when called with an empty batch (`B=0`), since `torch.nn.functional.conv2d` does not accept a zero batch size.

### Solution
Early-return an empty contiguous tensor with the same shape/dtype/device for `B==0`.

### Tests
- Added a cardinality test case for `batch_size=0` (CI will run full suite).

### DIFF
--- a/kornia/filters/median.py
+++ b/kornia/filters/median.py
@@ -57,6 +57,12 @@ def median_blur(input: torch.Tensor, kernel_size: tuple[int, int] | int) -> torc
     KORNIA_CHECK_IS_TENSOR(input)
     KORNIA_CHECK_SHAPE(input, ["B", "C", "H", "W"])
 
+    # NOTE: torch.nn.functional.conv2d does not accept a zero batch size.
+    # Keep behavior consistent with other Kornia filters by returning an empty,
+    # contiguous tensor with the same shape/dtype/device.
+    if input.shape[0] == 0:
+        return input.contiguous()
+
     padding = _compute_zero_padding(kernel_size)
 
     # prepare kernel
--- a/tests/filters/test_median.py
+++ b/tests/filters/test_median.py
@@ -29,12 +29,14 @@ class TestMedianBlur(BaseTester):
         actual = median_blur(inp, 3)
         assert isinstance(actual, torch.Tensor)
 
-    @pytest.mark.parametrize("batch_size", [1, 2])
+    @pytest.mark.parametrize("batch_size", [0, 1, 2])
     @pytest.mark.parametrize("kernel_size", [3, (5, 7)])
     def test_cardinality(self, batch_size, kernel_size, device, dtype):
         inp = torch.zeros(batch_size, 3, 4, 4, device=device, dtype=dtype)
         actual = median_blur(inp, kernel_size)
         assert actual.shape == (batch_size, 3, 4, 4)
+        if batch_size == 0:
+            assert actual.is_contiguous()
 
     def test_exception(self, device, dtype):
         from kornia.core.exceptions import ShapeError, TypeCheckError


sidd-27:
opened without maintainer approval

PR #3591, opened 2026-02-25
ci: allow tutorials to fail to unblock merges

aarush-github:
## 📝 Description
closes #3566 
**⚠️ Issue Link Required**: This PR must be linked to an approved and assigned issue. See [Contributing Guide](https://github.com/kornia/kornia/blob/main/CONTRIBUTING.md#pull-request) for details.

**Fixes/Relates to:** # (issue number)

**Important**:
- Ensure you are assigned to the linked issue before submitting this PR
- This PR should strictly implement what the linked issue describes
- Do not include changes beyond the scope of the linked issue

---

## 🛠️ Changes Made
- [x] Added `continue-on-error: true` to `.github/workflows/tutorials.yml` at the job level.

## 🧪 How Was This Tested?
- [ ] **Unit Tests:** (List new/updated tests)
- [ ] **Manual Verification:** (Describe the steps you took)
- [ ] **Performance/Edge Cases:** (How does this handle nulls, large data, etc.?)

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**
- [ ] 🟡 **AI-assisted:** I used AI for boilerplate/refactoring but have manually reviewed and tested every line.
- [ ] 🔴 **AI-generated:** (Note: These PRs may be subject to stricter scrutiny or immediate closure if the logic is not explained).

---

## 🚦 Checklist
- [ ] I am assigned to the linked issue (required before PR submission)
- [x] The linked issue has been approved by a maintainer
- [ ] I have performed a **self-review** of my code (no "ghost" variables or hallucinations).
- [ ] My code follows the existing style guidelines of this project.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] (Optional) I have attached screenshots/recordings for UI changes.

---

## 💭 Additional Context
Add any other context or screenshots about the pull request here.

Issue #3586, opened 2026-02-23
[Bug]: CI failure in tutorial notebooks – corrupted `cells_binary.png` in connected_components.ipynb


gupta-tilak:
### 🐛 Describe the bug

The CI job `Run make execute` fails when executing tutorial notebooks due to a corrupted image file used in `_nbs/connected_components.ipynb`.

FileExistsError: Data length (1449616) does not match the image size (1087212)

Traceback originates from:
```python
img_t = K.io.load_image("cells_binary.png", K.io.ImageLoadType.GRAY32)[None, ...]
```
inside _nbs/connected_components.ipynb.

Root Cause
The file cells_binary.png appears to be corrupted — the actual data buffer size (1449616 bytes) does not match the size declared in the image header (1087212 bytes). This causes kornia_rs.read_image_any() to raise a FileExistsError at the native level.

Likely causes:

Bad git lfs pointer or partial LFS upload
File was re-exported/re-saved with mismatched metadata
Corrupted binary commit to the repository

### 🔄 Steps to Reproduce

```bash
1. `make execute`
 or
1. `jupyter nbconvert _nbs/connected_components.ipynb --execute --inplace`
```

### ✅ Expected behavior

CI test should have passed successfuly

### ❌ Actual behavior

- Blocks CI for all unrelated PRs (e.g., test additions, model changes)
- Tutorial notebook connected_components.ipynb is non-functional
- The Run make execute CI job fails with exit code 2

### 🔧 Environment

```shell
CI runner: `ubuntu-latest / Python 3.12`
`kornia_rs` reads the image via `read_image_any()`
`kornia.io.load_image` at `kornia/io/io.py:102`
```

### 📝 Additional context

This error was first observed as a side effect during CI runs for PR #3585  (adding NaFlex test suite), confirming it is a pre-existing infrastructure issue unrelated to any recent code changes.

### 🤝 Contribution Intent

- [x] I plan to submit a PR to fix this bug
- [ ] I'm reporting this bug but not planning to fix it

aarush-github:
Hey, issue #3566 addresses the same issue. Let me know if you need help, as I am looking into the solution

shijianjian:
Closed by #3566

PR #3609, opened 2026-03-07
Fix/ci for prs

### LINKED ISSUE #3566: fix tutorials CI - blocking all PR merges
the tutorials CI job is failing on every single PR, which causes the collector job to fail too. this blocks merges (branch protection) and requires admin override for every merge.

failing job: tutorials / tutorials (3.12)

this is a systemic issue - not related to any specific PR. all recent PRs (#3563, #3536, #3542, #3544, #3545, #3546, #3561) show the same failure pattern.

options:
1. fix the tutorial that's failing
2. make tutorials CI non-blocking (move to scheduled/nightly)
3. add continue-on-error: true to tutorials job

the current state means every merge requires admin privileges, which is not sustainable.

johnnv1:
## 📝 Description

On repo setting, from the PR required jobs, should only be configured as the collector job. This job read from the states from previous job and fail if anyone of the required jobs has failed. Since we have an issue on tutorials repo/pipeline, we are removing for now the tutorials to be a required step

fixes #3566 

related to #3586 #3591 

---

## 🛠️ Changes Made
- autoupdated pre-commit hooks -- superseeds #3588
- fixed a typo into a comment
- removed tutorials job to be required in the PRs collector job

---

## 🕵️ AI Usage Disclosure

- [x] 🟢 **No AI used.**

---

## 🚦 Checklist

### DIFF
--- a/.github/workflows/pr_test_cpu.yml
+++ b/.github/workflows/pr_test_cpu.yml
@@ -92,7 +92,7 @@ jobs:
     uses: ./.github/workflows/tutorials.yml
 
   collector:
-    needs: [tests, tutorials, typecheck]
+    needs: [tests, typecheck]
     if: always()
     runs-on: ubuntu-latest
     steps:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,14 +21,14 @@ repos:
         exclude: ^$|.devcontainer
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.1
+    rev: v0.15.5
     hooks:
       - id: ruff-check
         args: [--fix, --exit-non-zero-on-fix]
       - id: ruff-format
 
   - repo: https://github.com/codespell-project/codespell
-    rev: v2.4.1
+    rev: v2.4.2
     hooks:
       - id: codespell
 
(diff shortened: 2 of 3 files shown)


johnnv1:
the gh action is kinda not running someplaces, i opened a fresh PR to trigger it and check if we had any other issue from it

Meggison (changes requested):
Would also need to temporarily pause CI calls on the tutorials for this to work, but a much better solution would be to only pause the single failing test

johnnv1:
> Would also need to temporarily pause CI calls on the tutorials for this to work, but a much better solution would be to only pause the single failing test

i don't get it, the tutorials CI could still be red, and we'll be able to perform merges, because we only require the CI of collector to pass. The right fix would always be fixing the tutorials itself

.github/workflows/pr_test_cpu.yml
```diff
@@ -92,7 +92,7 @@ jobs:
     uses: ./.github/workflows/tutorials.yml
 
   collector:
-    needs: [tests, tutorials, typecheck]
+    needs: [tests, typecheck]
```
shijianjian, line 95: I would leave the tutorials here. Otherwise, we need to spend additional resources on maintaining the tutorials.

johnnv1: +1

ducha-aiki:
Closing, as it was superseded by  #3588

PR #3612, opened 2026-03-11
fix(io): handle RGBA PNG files in load_image

### LINKED ISSUE #3607: [Bug]: load_image crashes on RGBA PNG files with "Data length does not match image size"
### 🐛 Describe the bug

`kornia.io.load_image()` fails when loading PNG files that have an alpha channel (RGBA). The error originates from `kornia_rs.read_image_any()` which assumes all PNGs are RGB (3 channels) and crashes when the actual pixel data has 4 channels.

### 🔄 Steps to Reproduce

```bash
import kornia as K

# Any RGBA PNG will trigger this
img = K.io.load_image("cells_binary.png", K.io.ImageLoadType.GRAY32)
```

### 💻 Minimal Code Example

```python

```

### ✅ Expected behavior

load_image() should handle RGBA PNGs for all ImageLoadType variants, converting channels as needed.

### ❌ Actual behavior

FileExistsError: Data length (1449616) does not match the image size (1087212)
The numbers correspond to 602×602×4 (RGBA) vs 602×602×3 (RGB), confirming the channel mismatch.

### 🔧 Environment

```shell
- Kornia version:
- PyTorch version:
- Python version:
- OS (e.g., Linux, macOS, Windows):
- Installation method (pip, conda, from source):
- CUDA/cuDNN version (if applicable):
- GPU model (if applicable):

You can also run:

python collect_env.py
```

### 📝 Additional context

Impact:
The connected_components tutorial notebook fails in CI because it loads cells_binary.png which is an RGBA PNG
Any user trying to load an RGBA PNG with load_image() hits this crash
Additionally, load_image was missing RGBA-to-RGB conversion paths for RGB8 and RGB32 load types, so even if the image was read successfully, those conversions would raise NotImplementedError

### 🤝 Contribution Intent

lustsazeus-lab:
## Description

This PR fixes the bug where load_image crashes when loading RGBA PNG files.

### Changes

1. PIL fallback for RGBA PNGs: When kornia_rs.read_image_any() crashes on RGBA PNGs (due to channel mismatch), the code falls back to PIL.

2. Missing conversion paths: Added missing RGBA-to-RGB conversion paths for RGB8, RGBA8, and RGB32.

### Root Cause

The bug has two parts:
1. kornia_rs.read_image_any() assumes all PNGs are RGB (3 channels) and crashes on RGBA (4 channels)
2. Python conversion paths were missing 4-channel handling

### Related Issue

Fixes #3607

### DIFF
--- a/kornia/io/io.py
+++ b/kornia/io/io.py
@@ -96,57 +96,72 @@ def _to_uint8(image: torch.Tensor) -> torch.Tensor:
     return image.mul(255.0).byte()
 
 
+def _convert_image_uint8(image: torch.Tensor, desired_type: ImageLoadType) -> torch.Tensor:
+    """Convert uint8 image to desired type."""
+    channels = image.shape[0]
+    # GRAY8
+    if desired_type == ImageLoadType.GRAY8:
+        if channels == 1:
+            return image
+        if channels == 3:
+            return kornia.color.rgb_to_grayscale(image)
+        if channels == 4:
+            gray32 = kornia.color.rgb_to_grayscale(kornia.color.rgba_to_rgb(_to_float32(image)))
+            return _to_uint8(gray32)
+    # RGB8
+    if desired_type == ImageLoadType.RGB8:
+        if channels == 3:
+            return image
+        if channels == 1:
+            return kornia.color.grayscale_to_rgb(image)
+        if channels == 4:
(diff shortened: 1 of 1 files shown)


sidd-27:
not assigned to you, theres already an open pr for this

PR #3636, opened 2026-03-19
Add unit tests for generate_examples helper functions

MOHITPRADHAN35:
This PR adds unit tests for helper functions introduced in generate_examples.py.

#### Tests added:
- handle_special_cases (RandomJigsaw, RandomJPEG)
- apply_augmentation basic execution

#### Benefits:
- Ensures correctness of refactored logic
- Improves reliability and maintainability

### DIFF
--- a/docs/generate_examples.py
+++ b/docs/generate_examples.py
@@ -36,6 +36,36 @@ import kornia as K
 
 mpl.use("Agg")
 
+def handle_special_cases(aug_name, img_in):
+    if aug_name == "RandomJigsaw":
+        img_in = K.geometry.resize(img_in, (1020, 500))
+    elif aug_name == "RandomJPEG":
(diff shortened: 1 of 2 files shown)


edgarriba (changes requested):
This can't be reviewed as-is: `tests/test_generate_examples.py` contains committed merge-conflict markers (`<<<<<<< HEAD` / `=======` / `>>>>>>>`), so the file isn't valid Python — that breaks pytest collection and is why every CI job fails.

Beyond that, the PR refactors `docs/generate_examples.py` rather than only adding tests, and importing a docs script (which pulls cv2/matplotlib into the unit-test environment) from the test suite is questionable. Please link an approved issue and rescope: if testing the docs generator is worthwhile, that's a design question to settle in the issue first.

PR #3646, opened 2026-03-20
Fix same_on_batch inconsistency in ColorJitter

MOHITPRADHAN35:
## 📝 Description

**Fixes/Relates to:** Fix inconsistency in `ColorJitter` when `same_on_batch=True`

### 🐛 Bug

When `same_on_batch=True`, `ColorJitter` was expected to apply identical transformations across the batch. However, the implementation applied transformations independently per sample, leading to inconsistent outputs even when parameters were identical.

Additionally, applying transforms on a single-sample tensor while using full-batch parameters caused shape mismatches (especially in hue adjustment).

---

### ✅ Fix

* Ensured that when `same_on_batch=True`, the transformation is computed on a **single sample** and then **broadcasted across the batch**
* Introduced a **local copy of parameters (`p`)** with batch size reduced to 1 (without mutating original params)
* Updated transform functions to use `p` instead of `params` to maintain consistency
* Preserved default behavior for `same_on_batch=False`

---

### 🛠️ Changes Made

* Modified `apply_transform` in `ColorJitter`
* Added safe handling of parameter shapes
* Ensured no mutation of shared state (`params`)
* Maintained compatibility with existing pipeline behavior

---

### 🧪 How Was This Tested?

* Ran all `ColorJitter` tests:

  * `test_colorjitter_same_on_batch_consistency` ✅
  * All augmentation tests related to color jitter ✅
* Ran full test suite:

  * No regressions observed
  * Only expected skips and xfails

---

### 🕵️ AI Usage Disclosure

* 🟡 AI-assisted: Used AI for debugging guidance and validation, all logic manually verified and tested

---

### 🚦 Checklist

* [x] Linked issue addressed
* [x] Self-review completed
* [x] Code follows project style
* [x] Tests pass locally
* [x] No unintended side effects introduced

---

## 💭 Additional Context

This fix ensures `same_on_batch` behaves as intended and aligns with expected augmentation semantics across Kornia.

### DIFF
--- a/kornia/augmentation/_2d/intensity/color_jitter.py
+++ b/kornia/augmentation/_2d/intensity/color_jitter.py
@@ -116,41 +116,32 @@ class ColorJitter(IntensityAugmentationBase2D):
         flags: Dict[str, Any],
         transform: Optional[torch.Tensor] = None,
     ) -> torch.Tensor:
-       transforms = [
-           lambda img: (
-               self._brightness_fn(img, params["brightness_factor"])
-               if (params["brightness_factor"] != 0).any()
-               else img
-           ),
-           lambda img: (
-               self._contrast_fn(img, params["contrast_factor"])
-               if (params["contrast_factor"] != 1).any()
-               else img
-           ),
-           lambda img: (
-               self._saturation_fn(img, params["saturation_factor"])
-               if (params["saturation_factor"] != 1).any()
-               else img
-           ),
-           lambda img: (
-               self._hue_fn(img, params["hue_factor"] * 2 * pi)
-               if (params["hue_factor"] != 0).any()
-               else img
-           ),
-       ]
+        transforms = [
+            lambda img: (
+                self._brightness_fn(img, params["brightness_factor"])
+                if (params["brightness_factor"] != 0).any()
+                else img
+            ),
+            lambda img: (
+                self._contrast_fn(img, params["contrast_factor"]) if (params["contrast_factor"] != 1).any() else img
+            ),
+            lambda img: (
+                self._saturation_fn(img, params["saturation_factor"])
+                if (params["saturation_factor"] != 1).any()
+                else img
+            ),
+            lambda img: self._hue_fn(img, params["hue_factor"] * 2 * pi) if (params["hue_factor"] != 0).any() else img,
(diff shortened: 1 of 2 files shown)


ducha-aiki:
Same on this PR

MOHITPRADHAN35:
> Same on this PR

Thanks for the update!
I’ve implemented a fix along with a test ensuring consistent behavior when `same_on_batch=True`.
Happy to:
- align with your approach,
- update the implementation,
- or close this PR if a similar fix is already in progress.
Let me know how you'd prefer to proceed.

themavik (commented):
nit: trim the loud inline comments (FIX / ALWAYS blocks) so this file stays consistent with the rest of augmentation. same_on_batch path (one jitter pass then repeat) is the right way to keep batch outputs identical.

edgarriba (changes requested):
Thanks for the effort here, but I don't think this bug exists: `same_on_batch` is already enforced at the parameter-generator layer (`ColorJitterGenerator.forward` routes it through `_adapted_rsampling`), so all batch elements receive identical factors and a shared `order`, and identical inputs already produce identical outputs on `main` — we verified this empirically.

`same_on_batch` means "same *transformation* across the batch," not "same *output*." The test in this PR feeds four *different* images and asserts identical outputs, and the fix achieves that by computing the jitter on `input[0:1]` and repeating it — which silently replaces samples 1..N with a copy of transformed sample 0. For any normal batch that's data corruption, and it would also break the partial-batch (`0 < p < 1`) apply path.

Since the premise doesn't hold, I'll close this. If you have a reproducible case where identical inputs + identical params give different outputs, please open an issue with the repro first — happy to look at it. If you'd like a well-scoped contribution in this area, the augmentation test-coverage gap is real: migrating tests to `BaseTester` (#2752) or extending shape/dtype coverage for the intensity augmentations would be very welcome.

edgarriba:
Closing per the review: the reported inconsistency doesn't reproduce on main (`same_on_batch` is enforced at the generator layer, verified empirically) and the proposed fix would corrupt normal batches. If you can reproduce identical inputs + identical params giving different outputs, please open an issue with the repro and we'll gladly revisit.

PR #3650, opened 2026-03-21
fix(geometry): use warp_perspective in crop_and_resize for non-parallel boxes

### LINKED ISSUE #3560: [Bug]: Incorrect cropping range in crop_and_resize when input boxes has non-parallel opposite sides
### 🐛 Describe the bug

`kornia.geometry.transform.crop_and_resize` behaves abnormally when oppsite sides of argument `boxes` are unparallel. This would result the cropped area cannot match the exact area of `boxes` in the `input_tensor`.

---

Image
<br/>Please download this image as checkboard.png for reproducing the bug.

---

Image annotated with cropping area
<br/>Image annotated with cropping 

---

The cropping result
<br/>The incorrect cropping result via kornia's crop_and_resize

---

Image
<br/>The correct cropping result by opencv

### 🔄 Steps to Reproduce

```bash
1. read the checkboard image
2. designate 4 corners of cropping bbox by `corner_idx`, the left-top corner of corresponding block is selected, order matches the signature of `crop_and_resizez
3. invoke `kornia.geometry.transform.crop_and_resize` to crop the image
4. get the result as description

For detail procedure, please refer to Minimal Code Example.
```

### 💻 Minimal Code Example

```python
Minimal python code to reproduce the bug.

Please save the origin picture depicted in description.


import cv2
import numpy as np
import torch
import kornia

# Read the checkboard, 8*8, 32px*32px for each block
img = cv2.imread("./checkerboard.png")

# Four corners for perspective cropping,
# for corners of the area is the left-up corner of corresponding block
corner_idx = [29, 23, 56, 45]
corner_coord = []

for cx in corner_idx:
    j = (cx - 1) // 8
    i = (cx - 1) % 8
    corner_coord.append((i * 32, j * 32))

# draw the cropping box
for cx in range(len(corner_coord)):
    p1 = corner_coord[cx]
    p2 = corner_coord[(cx + 1) % len(corner_coord)]
    cv2.line(
        img,
        p1,
        p2,
        (0, 0, 255),
        2,
    )
    cv2.putText(
        img,
        str(cx),
        (p1[0], p1[1] - 5),
        cv2.FONT_HERSHEY_SIMPLEX,
        0.5,
        (0, 0, 255),
        1,
    )

# Show the cropping area in origin image
cv2.imshow("cropping perspective area", img)
cv2.imwrite("cropping_area.png", img)

# Reproduce the bug
img = torch.from_numpy(img)
img = img.permute(2, 0, 1).unsqueeze(0) / 255 # [1,3,256,256]

img_result = kornia.geometry.transform.crop_and_resize(
    img,
    torch.tensor(corner_coord).unsqueeze(0), # [1,4,2]
    [128, 128]
)
img_result = (img_result[0].permute(1, 2, 0).numpy() * 255).astype(np.uint8)

cv2.imshow("cropped", img_result)
cv2.imwrite("cropping_result.png", img_result)

# Desired result
persp_mat = cv2.getPerspectiveTransform(
    np.array(corner_coord).astype(np.float32),
    np.array([
        [0, 0], [128, 0], [128, 128], [0, 128]
    ]).astype(np.float32)
)
img_result2 = cv2.warpPerspective(
    img[0].permute(1, 2, 0).numpy(),
    persp_mat,
    [128, 128]
)
img_result2 = (img_result2 * 255).astype(np.uint8)
cv2.imshow("cropped opencv", img_result2)
cv2.imwrite("cropping_result2.png", img_result2)

cv2.waitKey(0)
```

### ✅ Expected behavior

The area within the red border area should be cropped and rectified into the square image under perspective transformation:

Image

### ❌ Actual behavior

The wrong area is cropped and transformed:

Image

### 🔧 Environment

```shell
- Kornia version: 0.8.2
- PyTorch version: 2.10.0
- Python version: 3.12.12
- OS (e.g., Linux, macOS, Windows): found on ubuntu 20.04 with cuda, verified on macOS
- Installation method (pip, conda, from source): pip/uv
- CUDA/cuDNN version (if applicable): no
- GPU model (if applicable): no
(above behaviour is verified on macOS with cpu, but I assume that the cuda will not affect the result)

script output:

Collecting environment information...
PyTorch version: 2.9.1
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 26.2 (arm64)
GCC version: Could not collect
Clang version: 17.0.0 (clang-1700.0.13.5)
CMake version: version 3.30.2
Libc version: N/A

Python version: 3.12.12 | packaged by conda-forge | (main, Oct 22 2025, 23:34:53) [Clang 19.1.7 ] (64-bit runtime)
Python platform: macOS-26.2-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True
Caching allocator config: N/A

CPU:
Apple M2

Versions of relevant libraries:
[pip3] Could not collect
[conda] numpy                     2.3.2                    pypi_0    pypi
```

### 📝 Additional context

This issue is found during using kornia with my machine learning project.

### 🤝 Contribution Intent

- [x] I'm reporting this bug but not planning to fix it

FloodExLLC:
## Summary

Fixes #3560

`crop_and_resize` produces incorrect cropping results when the input `boxes` argument has non-parallel opposite sides (general quadrilaterals). The cropped area does not match the exact region defined by the 4-point box.

## Root Cause

The function used `warp_affine` with `dst_trans_src[:, :2, :]`, which slices the full 3×3 perspective transformation matrix down to 2×3. This discards the perspective components of the homography, which are required for correct results with non-rectangular quadrilaterals.

## Fix

Replace `warp_affine` (2×3 affine) with `warp_perspective` (3×3 homography), passing the full `dst_trans_src` matrix without slicing.

**Before:**
```python
patches: torch.Tensor = warp_affine(
    input_tensor,
    dst_trans_src[:, :2, :],  # slices away perspective component
    out_size,
    ...
)
```

**After:**
```python
patches: torch.Tensor = warp_perspective(
    input_tensor,
    dst_trans_src,  # full 3x3 homography
    out_size,
    ...
)
```

Also adds `warp_perspective` to the import from `.imgwarp`.

## Testing

The issue reporter provided a reproducible example with a checkerboard image. The fix aligns kornia's output with OpenCV's `cv2.warpPerspective` for non-rectangular quadrilateral boxes. Existing behaviour for rectangular (axis-aligned) boxes is unchanged since affine and perspective transforms are equivalent in that case.


Meggison (dismissed):
CI broken from tests that still pass Bx2x3 (affine) while the warp_perspective change uses Bx3x3. May need backward compatibility depending on the shape, or if you have a better solution feel free to clarify

Also, need to show some proof of working implementation like cropping examples to make sure it fixes the issue and tests before PR submission

edgarriba (dismissed):
Thanks — the root-cause analysis is right and `warp_perspective` is the correct primitive here. However, five internal call sites (`resize.py:117`, `center_crop.py:134/161`, `crop.py:271`, `resized_crop.py:160`, `crop2d.py:483`) pass a pre-sliced Bx2x3 matrix into `crop_by_transform_mat`, which is why all CI test jobs fail: `warp_perspective` requires Bx3x3.

Two asks:
1. Make `crop_by_transform_mat` accept both shapes by padding 2x3 inputs with a `[0, 0, 1]` row (this also preserves backward compatibility for external callers), or update all five internal callers to pass the full 3x3 with an explicit shape check.
2. Add a regression test in `tests/geometry/transform/test_crop2d.py` with a non-parallel quadrilateral box reproducing #3560 (the checkerboard scenario from the issue is a good candidate).

With those two items plus green CI this should be good to go.

ducha-aiki:
The diagnosis is exactly right — slicing the homography to 2×3 discards the perspective components, and with the fix applied locally the non-parallel-quad case works (verified: the four quad corners land precisely on the output corners, and rectangular crops stay bit-identical to the old path with `align_corners=True`).

But as-is the change breaks 21 tests, for two independent reasons:

1. **The augmentation pipeline feeds `crop_by_transform_mat` 2×3 matrices** (`RandomResizedCrop`/`CenterCrop` paths), which "worked" accidentally when the function sliced `[:, :2, :]`. With `warp_perspective` they fail: `ValueError: Input M must be a Bx3x3 torch.Tensor. Got torch.Size([2, 2, 3])`. The function needs to accept both — pad 2×3 to 3×3 with a `[0, 0, 1]` row.

2. **`align_corners=False` regresses** (`TestCropAndResize::test_align_corners_false`, max abs diff 1.55). This turned out to be a pre-existing kornia bug your PR exposes rather than causes: `warp_perspective` fails an *identity-homography* warp under `align_corners=False` (while `warp_affine` passes), because `normalize_homography` hard-codes the align_corners=True pixel→[-1,1] mapping. Filed as #3904 with a minimal repro.

Suggested path to merge:
- Handle the 2×3 input (pad to 3×3).
- Until #3904 is fixed, dispatch: use `warp_perspective` only when the (padded) matrix is genuinely projective (`bottom row ≉ [0, 0, 1]`), keep `warp_affine` otherwise. That keeps every existing affine crop bit-identical (including all `align_corners=False` uses) while fixing the quadrilateral case this PR targets.
- Add a test for the quad case per the repo's hardcoded-reference-values policy — the corner-marker probe from this review works well: place distinct values at the 4 quad corners, crop with `mode="nearest"`, assert they land at the output corners.

*— review written by Claude (Claude Code) on behalf of @ducha-aiki*

ducha-aiki:
Pushed maintainer fixes addressing the review:

1. `crop_by_transform_mat` now accepts both `(B, 2, 3)` and `(B, 3, 3)` transforms (padded via `convert_affinematrix_to_homography`), so the five internal call sites passing pre-sliced affine matrices keep working — this was the cause of the 29 red CI jobs.
2. Dispatch on the projective row: affine transforms keep the `warp_affine` path, only genuinely projective ones (non-parallel boxes) go through `warp_perspective`. This matters because `warp_affine` and `warp_perspective` build different sampling grids for `align_corners=False` (affine_grid convention vs `create_meshgrid` always using align_corners=True spacing), so the naive switch changed existing crop values for plain rectangles and broke `test_align_corners_false`. That grid inconsistency in `warp_perspective` is a separate pre-existing issue worth its own ticket.
3. Added `test_crop_non_parallel_box` regressing #3560, with a numpy-only DLT reference snippet embedded (fails on `main` with 70% error at the far corner, passes here).

Local runs: tests/geometry/transform/test_crop2d.py `38 passed` (float32/float64), dynamo variants `3 passed` (inductor), augmentation Crop/Resize/Mosaic `37 passed`. TorchScript exercises both dispatch branches fine.

ducha-aiki:
CI failure root-caused and fixed in 54d6adce: `test_torch_export_container[Seq[Resize,CenterCrop-resample]]` failed with `GuardOnDataDependentSymNode` because the affine/perspective dispatch branched on tensor values, which `torch.export` cannot capture.

Replaced with a branch-free design: always `warp_perspective`, plus a **static** destination-side reparametrization for `align_corners=False` (`p' = n/(n-1)·p − ½` per axis, built from `out_size` only) that maps `warp_perspective`'s align_corners=True grid spacing onto the historical `warp_affine`/`F.affine_grid` convention — verified equivalent to the legacy affine path to ~2e-7 in float64 across random affine warps, and exact for `align_corners=True` (no correction needed there). 1-pixel outputs keep the legacy affine path, selected statically (the reparametrization is singular for n=1).

Local verification: torch_export suite `9 passed` (previously failing test included), test_crop2d `38 passed` (float32/float64), full tests/geometry/transform `412 passed`, augmentation Crop/Resize/Mosaic `37 passed`, dynamo/inductor `3 passed`, TorchScript covers all three static paths.

ducha-aiki:
Perf follow-up in 9a3a2e5b: `crop_by_transform_mat` now dispatches on the transform **shape** — static, so torch.export-safe — instead of always warping perspective:

- `(B, 2, 3)` → `warp_affine` (affine by construction; no perspective divide). Benchmark 16x3x224x224 → 128x128: warp_affine 1.96 ms vs warp_perspective 2.61 ms on CPU (~25%), 5.57 vs 6.60 ms on MPS (~15%).
- `(B, 3, 3)` → `warp_perspective` with the static align_corners=False convention correction.

`RandomCrop`/`RandomResizedCrop`/`Mosaic` now pass pre-sliced `transform[:, :2, :]` like the five other augmentation call sites already did, so the entire augmentation pipeline stays on the fast affine path with values byte-identical to `main`. The perspective path is only taken by the geometry API (`crop_by_boxes`/`crop_and_resize`/`center_crop`), where correctness for non-parallel boxes is the point.

Local runs: torch_export 9 passed, test_crop2d 38 passed (float32/float64), tests/geometry/transform 412 passed, augmentation Crop/Resize/Mosaic suites passing, dynamo/inductor 3 passed, TorchScript covers both shapes.

ducha-aiki (approved):

PR #3677, opened 2026-03-27
Fix BgrToRgba calling wrong conversion function

### LINKED ISSUE #3670: [Bug]: BgrToRgba class incorrectly calls rgb_to_rgba in forward pass
### 🐛 Describe the bug

While working on standardizing the docstrings for the kornia.color module, I noticed a logic error in kornia/color/rgb.py. The BgrToRgba nn.Module class currently calls the rgb_to_rgba function instead of bgr_to_rgba in its forward pass.

### 🔄 Steps to Reproduce

```bash
1.Instantiate the BgrToRgba module.
2.Pass a BGR image tensor to it.
3.Observe that it treats the input as RGB instead of BGR because it incorrectly routes to rgb_to_rgba internally.
```

### 💻 Minimal Code Example

```python
import torch
from kornia.color import BgrToRgba

bgr_img = torch.rand(1, 3, 4, 4)
converter = BgrToRgba(1.)

# Internally this calls rgb_to_rgba instead of bgr_to_rgba, causing flipped channels
rgba_img = converter(bgr_img)
```

### ✅ Expected behavior

The forward method of the BgrToRgba class should call return bgr_to_rgba(image, self.alpha_val) to correctly convert the BGR input.

### ❌ Actual behavior

The method currently calls return rgb_to_rgba(image, self.alpha_val), which leaves the Red and Blue channels flipped in the resulting RGBA tensor.

### 🔧 Environment

```shell
- Kornia version: Master branch (Latest)
- PyTorch version: 2.2.0
- Python version: 3.11.15
- OS (e.g., Linux, macOS, Windows): windows - WSL: Ubuntu-22.04
- Installation method (pip, conda, from source):
- CUDA/cuDNN version (if applicable): N/A (Logic bug)
- GPU model (if applicable): NVIDIA GeForce RTX 4060

You can also run:

python collect_env.py
```

### 📝 Additional context

I discovered this while writing the detailed tensor shape docstrings for the color module (related to my ongoing GSoC 2026 preparation). 

### 🤝 Contribution Intent

ArnavGarg7:
Fixes #3670.

## Description
The BgrToRgba module was incorrectly calling rgb_to_rgba instead of bgr_to_rgba. Also updated the test case to use torch.rand instead of torch.ones so that channel swaps are actually tested properly.

## AI Usage Disclosure
- [ ] No AI used.
- [ ] - [x] AI-assisted: I used AI for boilerplate/refactoring but have manually reviewed and tested every line.
- [ ] - [ ] AI-generated: (Note: These PRs may be subject to stricter scrutiny or immediate closure if the logic is not explained).

### DIFF
--- a/tests/color/test_rgb.py
+++ b/tests/color/test_rgb.py
@@ -262,14 +262,14 @@ class TestRgbToRgba(BaseTester):
 
     def test_module(self, device, dtype):
         B, C, H, W = 2, 3, 4, 4
-        img = torch.rand(B, C, H, W, device=device, dtype=dtype)
+        img = torch.ones(B, C, H, W, device=device, dtype=dtype)
         ops = kornia.color.RgbToRgba(1.0).to(device, dtype)
         fcn = kornia.color.rgb_to_rgba
         self.assert_close(ops(img), fcn(img, 1.0))
 
     def test_module_bgr(self, device, dtype):
         B, C, H, W = 2, 3, 4, 4
-        img = torch.rand(B, C, H, W, device=device, dtype=dtype)
+        img = torch.ones(B, C, H, W, device=device, dtype=dtype)
         ops = kornia.color.BgrToRgba(1.0).to(device, dtype)
         fcn = kornia.color.bgr_to_rgba
         self.assert_close(ops(img), fcn(img, 1.0))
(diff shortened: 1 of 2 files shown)


tests/color/test_rgb.py
```diff
@@ -262,14 +262,14 @@ def test_jit(self, device, dtype):
 
     def test_module(self, device, dtype):
         B, C, H, W = 2, 3, 4, 4
-        img = torch.ones(B, C, H, W, device=device, dtype=dtype)
+        img = torch.rand(B, C, H, W, device=device, dtype=dtype)
```
shijianjian: why change this? Probably you shall add some numerical tests to prevent this bug happens again in the future

shijianjian (changes requested):

shijianjian (approved):

PR #3722, opened 2026-05-17
refactor(augmentation): makes it onnx exportable

dhruvkjain:
## 📝 Description

**Fixes/Relates to:** torchgeo/torchgeo#3108

Makes `kornia.augmentation` ONNX-exportable via `torch.onnx.export` (legacy tracer, opset 20). Before this PR, 
- the augmentation base classes' `index_put` scatter pattern, 
- runtime int validators, 
- an in-place mutation in `normal_transform_pixel`, 
- `aten::linalg_inv`/`aten::linalg_solve` calls, 
- and the `AugmentationSequential` arity validator combined to prevent every tested augmentation from producing a usable ONNX graph. 

After this PR, **15 augmentations export cleanly** and **12 produce results numerically equivalent to eager** for deterministic configurations including `AugmentationSequential`, the wrapper used by torchgeo for their pipelines.
The 3 export-only entries are exercised by `test_onnx_export_exportable` but don't currently have a numerical-equivalence test.

| # | Augmentation | Exports (`torch.onnx.export`, opset 20) | Numerically matches eager (deterministic config) | Notes |
|---|---|---|---|---|
| 1  | `RandomHorizontalFlip`   | ✅ | ✅ | Parameter-free |
| 2  | `RandomVerticalFlip`     | ✅ | ✅ | Parameter-free |
| 3  | `RandomGrayscale`        | ✅ | ✅ | Parameter-free |
| 4  | `RandomInvert`           | ✅ | ✅ | Parameter-free |
| 5  | `Normalize`              | ✅ | ✅ | Parameter-free |
| 6  | `RandomBrightness`       | ✅ | ✅ | Pinned via `brightness=(1.2, 1.2)` |
| 7  | `ColorJiggle`            | ✅ | ✅ | Pinned via `brightness=(1.2, 1.2)` |
| 8  | `RandomGaussianBlur`     | ✅ | ✅ | Pinned via `sigma=(1.5, 1.5)` |
| 9  | `RandomSolarize`         | ✅ | ✅ | Pinned via `thresholds=(0.5, 0.5)` |
| 10 | `RandomRotation`         | ✅ | ✅ | Pinned via `degrees=(15., 15.)` , unblocked by `normal_transform_pixel` fix |
| 11 | `RandomAffine`           | ✅ | ✅ | Pinned via `degrees=(15., 15.)`, no translate/scale/shear |
| 12 | `RandomAffine` (w/ shear)| ✅ | ✅ | Pinned via `shear=(5., 5.)` , unblocked by `deg2rad` inlining |
| 13 | `RandomErasing`          | ✅ | - | Not asserted; random mask placement is hard to compare exactly without a during-trace capture harness |
| 14 | `RandomPosterize`        | ✅ | - | Not asserted (pure bit-quantization is trivially equivalent) |
| 15 | `RandomPerspective`      | ✅ | - | Not asserted; exercises the `_get_perspective_transform_closed_form` path |
| - | `AugmentationSequential` | ✅ | - | Not asserted; container-level wiring tested by export-success only |

---

## 🛠️ Changes Made

**Augmentation base classes** (`kornia/augmentation/base.py`, `_2d/base.py`, `_3d/base.py`)
- [x] Replaced `Bernoulli` / `RelaxedBernoulli` with direct `(torch.rand(...) < p)` in `__batch_prob_generator__` (opset-13 friendly; no distribution-machinery ops).
- [x] `forward_parameters` now calls `generate_parameters(batch_shape)` with the full batch shape, this removes the `nonzero()`-driven subset path that produced dynamic-shape ops.
- [x] Replaced `output.index_put((to_apply,), applied)` scatter in all 5 transform paths (`transform_inputs`, `transform_masks`, `transform_boxes`, `transform_keypoints`, `transform_classes`) with full-batch apply + `torch.where` blend. Shape- and batch-dim-mismatch fallbacks preserve old behavior for `RandomCrop`/`Resize` and nested `VideoSequential` cases.
- [x] `Boxes` / `Keypoints` reconstruction switched to `clone()`-and-replace-data (preserves `mode`/`_N`/`is_batched`).
- [x] Cascaded the same `where`-blend pattern into `generate_transformation_matrix` for both 2D and 3D bases.
- [x] Added `exportable` property + `ONNX_EXPORTABLE` class flag.

**Random generator validators** (`kornia/augmentation/utils/param_validation.py`, 9 generator files)
- [x] Added `_is_traced_dim` + `_check_positive_int_or_traced` helpers in `param_validation.py`.
- [x] Relaxed `_common_param_check` and the inline `isinstance(width, int) and isinstance(height, int)` checks across 9 generator files: `_2d/{affine,translate,shear,cutmix,crop,perspective,rectangle_earase}.py`, `_3d/{crop,affine}.py`. Static configuration validation still runs at `__init__`; runtime checks now no-op on traced 0-d shape tensors (which is what `tensor.shape[i]` becomes under tracing).

**Linear-algebra ops** (`kornia/core/utils.py`, `kornia/geometry/transform/imgwarp.py`)
- [x] Added `_inverse_3x3_closed_form` (cofactor / determinant formula). `_torch_inverse_cast` routes through it for 3×3 inputs under `torch.jit.is_tracing()`, avoiding the unsupported `aten::linalg_inv` symbol. Numerically equivalent to `torch.linalg.inv` to ~2e-7.
- [x] Added `_unit_square_to_quad` (Heckbert direct formulation) and `_get_perspective_transform_closed_form`. `get_perspective_transform` routes through it under tracing, replacing the unsupported 8×8 `aten::linalg_solve` with two 3×3 unit-square decompositions + one closed-form 3×3 inverse. Matches the DLT path to ~1e-7.

**Geometric warp numerical correctness** (`kornia/geometry/conversions.py`)
- [x] Rewrote `normal_transform_pixel` to construct the normalization matrix in one shot from precomputed `sx`/`sy` floats. The previous `tr_mat[0, 0] = ...; tr_mat[1, 1] = ...` in-place mutation was **silently dropped by the legacy ONNX tracer** , the trace captured the un-mutated `[[1, 0, -1], [0, 1, -1], [0, 0, 1]]` constant and the resulting graph produced garbage warp coordinates (max diff ~4 on unit-norm input for `RandomRotation` / `RandomAffine`). This was the root cause of the geometric-drift class of bugs.

**`torch.deg2rad` inlining** (`_2d/geometric/affine.py`, `_2d/geometric/shear.py`, `_3d/geometric/affine.py`)
- [x] Replaced 8 call sites with explicit `* (math.pi / 180.0)`. Legacy ONNX has no symbolic for `aten::deg2rad` at opset 20.

**`AugmentationSequential`** (`kornia/augmentation/container/augment.py`)
- [x] `forward` strips trailing `None` positional args. The legacy ONNX tracer rebinds keyword-only defaults (`params=None`, `data_keys=None`) as positional, so `forward(x)` arrived as `forward(x, None, None)` and tripped the data-keys arity validator. Real data inputs are never `None`, so the strip is safe in eager too. **This was the unlock for `AugmentationSequential`-wrapped pipelines** (i.e. torchgeo's actual use case).

**`Normalize`** (`kornia/augmentation/_2d/intensity/normalize.py`)
- [x] During ONNX export, `apply_transform` reshapes `(C,)` mean/std to `(1, C)` so the underlying `kornia.enhance.normalize` ONNX-only path (which enforces leading dim 1) accepts the standard per-channel tensor users pass.

**Mix-augmentation contract update** (`kornia/augmentation/_2d/mix/jigsaw.py`)
- [x] `RandomJigsaw.apply_transform` slices `params["permutation"][to_apply]` to match the new full-batch params contract.

**Test contract updates** (`tests/augmentation/test_base.py`, `tests/augmentation/test_augmentation.py`, `tests/augmentation/_3d/test_random_rotation_3d.py`)
- [x] `test_forward_params` / `test_forward` in `test_base.py`: updated to the full-batch param contract (`len(params) == batch_size`); relaxed identity check (`is` to `torch.equal`) since `where`-blend always returns a fresh tensor.
- [x] `_test_smoke_implementation` in `test_augmentation.py`: expects full-batch `(B, 3, 3)` transform matrices.
- [x] Rewrote `_test_module_implementation` (2D) and `test_batch_random_rotation` / `test_sequential` (3D) to test intended invariants (composition equivalence, per-element matrix variation, valid rigid-transform structure) instead of hardcoded pixel values that depended on RNG accidents in the old Bernoulli path. The old tests passed only because seed 42 happened to roll `to_apply=all-False` at `p=0.5`, making both compared sides reduce to the unmodified input.

---

## 🧪 How Was This Tested?

- [x] **Unit Tests:** Added `tests/augmentation/test_onnx_export.py` (28 tests):
  - `test_onnx_export_exportable` ,15 augmentations (`HFlip`, `VFlip`, `ColorJiggle`, `Brightness`, `Grayscale`, `Invert`, `GaussianBlur`, `Erasing`, `Posterize`, `Solarize`, `Normalize`, `Affine`, `Rotation`, `Perspective`, `AugmentationSequential`) export via `torch.onnx.export` at opset 20.
  - `test_onnx_export_numerically_matches_eager`, 12 augmentations under deterministic configurations produce results within `1e-3` of eager (most are `< 1e-5`, near float32 precision). Uses `onnxruntime` via `pytest.importorskip` so it's a no-op when the optional dep isn't installed.
  - `test_onnx_export_known_blocked`, currently empty (no augs in the tested set are blocked at export). Kept as the slot for future tracking.

- [x] **Regression:** `pytest tests/augmentation tests/geometry/transform --dtype=float32`: **2698 passed, 0 failed.** All pre-existing tests that were passing on `main` continue to pass; the new tests are additive.

- [x] **Manual Verification:**
  ```python
  aug = K.AugmentationSequential(
      K.RandomHorizontalFlip(p=0.5),
      K.RandomVerticalFlip(p=0.5),
      K.RandomRotation(degrees=(15., 15.)),
      K.Normalize(mean=torch.zeros(3), std=torch.ones(3)),
      data_keys=["input"],
  )
  torch.onnx.export(aug.eval(), (x,), buf, opset_version=20, dynamo=False)
  # graph runs in onnxruntime and matches aug(x) within 3e-5 max abs diff
  ```

  - Closed-form 3×3 inverse verified numerically equivalent to `torch.linalg.inv` (max diff ~2e-7 on random well-conditioned 3×3 matrices).
  - Closed-form Heckbert perspective verified against the 8×8-solve DLT path (max diff ~1e-7).
  - **JIT-script compatibility**: gated on `torch.jit.is_tracing()` (not `torch.onnx.is_in_onnx_export()`, whose body contains an `import` statement that JIT script cannot compile). `tests/geometry/transform/test_imgwarp.py::TestWarpAffine::test_jit_script` confirms, all 4 parametrizations pass.
  - **Eager behavior is unchanged**: the closed-form paths are guarded by `torch.jit.is_tracing()`; the `Normalize` reshape is guarded by `torch.onnx.is_in_onnx_export()`. Static validation in `__init__` is untouched.
  - **Boxes / Keypoints** containers tested via the mix-augmentation and autocast paths (`test_augmentation_mix.py`, `test_augmentation_sequential.py`), both still pass.
  - **Shape-changing augs** (`RandomCrop`, `Resize`) handled via a Python `to_apply.any()` fallback when the apply / non-apply outputs have different spatial shapes; preserves the old all-or-nothing semantics.

---

## 🕵️ AI Usage Disclosure

- [x] 🟡 **AI-assisted:** I used AI (Claude Opus 4.7) for boilerplate/refactoring but have manually reviewed and tested every line.

---

## 🚦 Checklist

- [ ] I am assigned to the linked issue (required before PR submission) — please confirm the appropriate kornia-side issue.
- [ ] The linked issue has been approved by a maintainer — please confirm.
- [x] I have commented my code, particularly in hard-to-understand areas (tracing-gate and in-place-mutation comments document the *why*).
- [x] I have added tests that prove my fix is effective (`tests/augmentation/test_onnx_export.py`, 28 tests).
- [ ] (Optional) Screenshots/recordings — N/A, no UI changes.

---

## 💭 Additional Context

**Minimum opset:** 20, required for `aten::affine_grid_generator`. Earlier opsets work for non-geometric augmentations only.

**`dynamo=True` path:** Currently 0/15, strictly worse than the legacy tracer for the augmentation surface today. This PR targets the `dynamo=False` (legacy TorchScript tracer) path. Migrating to `dynamo=True` is a future effort.

### DIFF
--- a/kornia/augmentation/_2d/base.py
+++ b/kornia/augmentation/_2d/base.py
@@ -105,8 +105,6 @@ class RigidAffineAugmentationBase2D(AugmentationBase2D):
 
         in_tensor = self.transform_tensor(input)
 
-        # Always compute both transform and identity for the full batch, then blend.
-        # This avoids NonZero/index_put dynamic-shape ops and is ONNX-traceable.
         trans_matrix_applied = self.compute_transformation(in_tensor, params=params, flags=flags)
         trans_matrix_identity = self.identity_matrix(in_tensor)
 
--- a/kornia/augmentation/_2d/base.py
+++ b/kornia/augmentation/_2d/base.py
@@ -101,22 +101,26 @@ class RigidAffineAugmentationBase2D(AugmentationBase2D):
     ) -> torch.Tensor:
         """Generate transformation matrices with the given input and param settings."""
         batch_prob = params["batch_prob"]
-        to_apply = batch_prob > 0.5  # NOTE: in case of Relaxed Distributions.
+        to_apply = batch_prob > 0.5
 
         in_tensor = self.transform_tensor(input)
-        if not to_apply.any():
-            trans_matrix = self.identity_matrix(in_tensor)
-        elif to_apply.all():
-            trans_matrix = self.compute_transformation(in_tensor, params=params, flags=flags)
+
+        # Always compute both transform and identity for the full batch, then blend.
+        # This avoids NonZero/index_put dynamic-shape ops and is ONNX-traceable.
+        trans_matrix_applied = self.compute_transformation(in_tensor, params=params, flags=flags)
+        trans_matrix_identity = self.identity_matrix(in_tensor)
+
+        if is_autocast_enabled():
+            trans_matrix_applied = trans_matrix_applied.type(input.dtype)
+            trans_matrix_identity = trans_matrix_identity.type(input.dtype)
+
+        # If batch sizes line up, do the where-blend. Otherwise (e.g. VideoSequential
+        # passes B-sized batch_prob into a B*T-sized input) fall back to all-or-nothing.
+        if trans_matrix_applied.shape[0] == to_apply.shape[0] == trans_matrix_identity.shape[0]:
+            to_apply_expanded = to_apply.view(-1, *([1] * (trans_matrix_applied.dim() - 1)))
+            trans_matrix = torch.where(to_apply_expanded, trans_matrix_applied, trans_matrix_identity)
         else:
-            trans_matrix_A = self.identity_matrix(in_tensor)
-            trans_matrix_B = self.compute_transformation(in_tensor[to_apply], params=params, flags=flags)
-
-            if is_autocast_enabled():
-                trans_matrix_A = trans_matrix_A.type(input.dtype)
-                trans_matrix_B = trans_matrix_B.type(input.dtype)
-
-            trans_matrix = trans_matrix_A.index_put((to_apply,), trans_matrix_B)
+            trans_matrix = trans_matrix_applied if bool(to_apply.any()) else trans_matrix_identity
 
         return trans_matrix
 
--- a/kornia/augmentation/_2d/geometric/affine.py
+++ b/kornia/augmentation/_2d/geometric/affine.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 #
 
+import math
 from typing import Any, Dict, Optional, Tuple, Union
 
 import torch
@@ -124,13 +125,19 @@ class RandomAffine(GeometricAugmentationBase2D):
     def compute_transformation(
         self, input: torch.Tensor, params: Dict[str, torch.Tensor], flags: Dict[str, Any]
     ) -> torch.Tensor:
+        # ``torch.deg2rad`` lowers as ``aten::deg2rad`` which the legacy ONNX exporter
+        # does not support at opset 20. Multiply by ``pi/180`` explicitly — same value,
+        # but only basic arithmetic ops in the trace.
+        deg2rad_factor: float = math.pi / 180.0
+        shear_x = torch.as_tensor(params["shear_x"], device=input.device, dtype=input.dtype) * deg2rad_factor
+        shear_y = torch.as_tensor(params["shear_y"], device=input.device, dtype=input.dtype) * deg2rad_factor
         return get_affine_matrix2d(
             torch.as_tensor(params["translations"], device=input.device, dtype=input.dtype),
             torch.as_tensor(params["center"], device=input.device, dtype=input.dtype),
             torch.as_tensor(params["scale"], device=input.device, dtype=input.dtype),
             torch.as_tensor(params["angle"], device=input.device, dtype=input.dtype),
-            torch.deg2rad(torch.as_tensor(params["shear_x"], device=input.device, dtype=input.dtype)),
-            torch.deg2rad(torch.as_tensor(params["shear_y"], device=input.device, dtype=input.dtype)),
+            shear_x,
+            shear_y,
         )
 
     def apply_transform(
--- a/kornia/augmentation/_2d/geometric/affine.py
+++ b/kornia/augmentation/_2d/geometric/affine.py
@@ -126,7 +126,7 @@ class RandomAffine(GeometricAugmentationBase2D):
         self, input: torch.Tensor, params: Dict[str, torch.Tensor], flags: Dict[str, Any]
     ) -> torch.Tensor:
         # ``torch.deg2rad`` lowers as ``aten::deg2rad`` which the legacy ONNX exporter
-        # does not support at opset 20. Multiply by ``pi/180`` explicitly — same value,
+        # does not support at opset 20. Multiply by ``pi/180`` explicitly same value,
         # but only basic arithmetic ops in the trace.
         deg2rad_factor: float = math.pi / 180.0
         shear_x = torch.as_tensor(params["shear_x"], device=input.device, dtype=input.dtype) * deg2rad_factor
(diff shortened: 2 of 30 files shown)


shijianjian:
I like the idea. Can you please also add a benchmark after the refactor? I saw some logic changes, not sure how would it affect the performance.

dhruvkjain:
> I like the idea. Can you please also add a benchmark after the refactor? I saw some logic changes, not sure how would it affect the performance.

I ran a comparison between current release and this implementation, as described in the table in PR, deterministic values are same and about non-deterministic part the distribution is identical

about speed, it improved around 8-10%

Varun-sai-500:
@sidd-27  @shijianjian  could you merge?

shijianjian (approved):

PR #3723, opened 2026-05-17
[Fix] Enforce same_on_batch consistency for RandomGaussianNoise and RandomChannelShuffle

### LINKED ISSUE #3624: [Bug]: same_on_batch=True does not enforce batch-consistent randomness for several intensity augmentations (e.g., RandomGaussianNoise)
### 🐛 Describe the bug

With `same_on_batch=True`, several `kornia.augmentation` intensity transforms still produce different outputs across batch elements for identical input images.

This seems to violate the documented expectation that `same_on_batch` applies the same transformation across the batch.

Affected (reproduced):
- `RandomGaussianNoise` (large mismatch)
- `RandomChannelShuffle`
- `RandomRain`
- `RandomPlasmaBrightness`
- `RandomPlasmaContrast`
- `RandomPlasmaShadow`

Not affected in my checks:
- `RandomGamma`, `RandomContrast`, `RandomBrightness`, `RandomSaturation`, `RandomHue`, `RandomSharpness`, `RandomMotionBlur`, `RandomGaussianBlur` (with explicit args)

`RandomJPEG` had only tiny fp-level differences (~2e-7).

### 🔄 Steps to Reproduce

1. Create a batch where all images are identical.
2. Apply a transform with same_on_batch=True and p=1.0.
3. Compare output[0] vs output[i] for i>0.
4. Observe non-zero max differences for some transforms.

### 💻 Minimal Code Example

```python
import torch

torch.manual_seed(0)

# identical batch
x = torch.rand(4, 3, 64, 64)
x[:] = x[0:1]

ops = {
    "RandomGaussianNoise": K.RandomGaussianNoise(mean=0.0, std=0.02, p=1.0, same_on_batch=True),
    "RandomChannelShuffle": K.RandomChannelShuffle(p=1.0, same_on_batch=True),
    "RandomRain": K.RandomRain(p=1.0, same_on_batch=True),
    "RandomPlasmaBrightness": K.RandomPlasmaBrightness(p=1.0, same_on_batch=True),
    "RandomPlasmaContrast": K.RandomPlasmaContrast(p=1.0, same_on_batch=True),
    "RandomPlasmaShadow": K.RandomPlasmaShadow(p=1.0, same_on_batch=True),
    "RandomGamma": K.RandomGamma((0.6, 1.2), p=1.0, same_on_batch=True),
}

for name, aug in ops.items():
    y = aug(x)
    max_diff = max((y[0] - y[i]).abs().max().item() for i in range(1, y.shape[0]))
    print(f"{name}: max_diff={max_diff:.9f}")
```

### ✅ Expected behavior

For identical batch inputs and `same_on_batch=True`, outputs should be identical across batch elements (up to tiny floating-point tolerance).

### ❌ Actual behavior

Some transforms still apply per-sample randomness even when `same_on_batch=True`.

### 🔧 Environment

```shell
- Kornia version: 0.8.2
- PyTorch version: 2.8.0+cu129
- Python version: 3.12.12
- OS: Linux (WSL2, kernel 5.15.167.4-microsoft-standard-WSL2)
- Installation method: pip
- CUDA/cuDNN version: CUDA 12.9, cuDNN 91002
- GPU model: NVIDIA GeForce RTX 3080 Ti Laptop GPU

You can also run:

python collect_env.py
```

### 📝 Additional context

* `RandomGaussianNoise` appears to generate noise directly in `apply_transform` via `torch.randn_like(input)` instead of shared sampled params.
* `RandomChannelShuffle` appears to sample per-item channel permutations in `generate_parameters`.
* `RandomRain` and `RandomPlasma*` also appear to have per-item random map behavior.
If this is intended semantics (i.e., `same_on_batch` only means shared scalar params, not identical stochastic fields), docs might need clarification because current wording suggests full batch-consistent randomness.

### 🤝 Contribution Intent

- [x] I'm reporting this bug but not planning to fix it

heckkkerrr:
Fixes #3624

## Description

This PR resolves a bug where setting `same_on_batch=True` failed to enforce deterministic, batch-consistent randomness for `RandomGaussianNoise` and `RandomChannelShuffle`.

### Changes Made

* **RandomGaussianNoise**

  * Updated `apply_transform` to correctly generate a single shared noise tensor when `same_on_batch=True`.
  * Preserves the existing GPU-native `torch.randn_like(...)` generation path for performance.

* **RandomChannelShuffle**

  * Updated `generate_parameters` to generate a single shared channel permutation across the batch when `same_on_batch=True`.

* **Tests**

  * Added regression tests validating batch-consistent behavior for both augmentations.
  * Tests use repeated random tensors to ensure permutations and stochastic fields are meaningfully validated.

## Validation

* [x] Relevant augmentation tests pass
* [x] Linting/formatting checks pass
* [x] Minimal targeted diff only
* [x] Preserves Kornia architecture and performance behavior


kornia/augmentation/_2d/intensity/channel_shuffle.py
```diff
@@ -56,7 +56,11 @@ def __init__(self, same_on_batch: bool = False, p: float = 0.5, keepdim: bool =
 
     def generate_parameters(self, shape: Tuple[int, ...]) -> Dict[str, torch.Tensor]:
```
shijianjian: The main reason for this error is that some contributors do not use the random generator. Probably you can refer to this code for refactoring:
https://github.com/kornia/kornia/blob/d6bb4bf0d8a043c2bb8cef0c346a1b006d100930/kornia/augmentation/_2d/intensity/brightness.py#L86

Also, can you please check other files if this oversight happens somewhere else?

shijianjian (changes requested):

kornia/augmentation/_2d/intensity/channel_shuffle.py
heckkkerrr: Thanks for the guidance I'll refactor to use the random generator pattern following brightness.py and audit other affected files. Will push an update shortly.

heckkkerrr: Hi @shijianjian, the refactor is pushed! I migrated everything to the ParamGenerator architecture:

RandomChannelShuffle: Built ChannelShuffleGenerator using _adapted_rsampling to natively handle same_on_batch.

RandomGaussianNoise: Built GaussianNoiseGenerator to pre-generate the noise tensor and .expand() it correctly.

Codebase Audit: Checked RandomRain and RandomPlasma as requested. Both already use the generator architecture correctly, so I left them as-is.

shijianjian (approved):

Issue #3727, opened 2026-05-20
[Deprecated]: pytorch deprecated the `torch.jit.script` decorator


woutdenolf:
### 🐛 Describe the bug

Pytorch has deprecated the `torch.jit.script` decorator and it is not supported in Python 3.14+ and may break.

```bash
pip install kornia
python -W error -c "import kornia"
```

The result

```bash
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File ".../python3.12/site-packages/kornia/__init__.py", line 28, in <module>
    from . import filters
  File ".../python3.12/site-packages/kornia/filters/__init__.py", line 64, in <module>
    from .kernels_geometry import get_motion_kernel2d, get_motion_kernel3d
  File ".../python3.12/site-packages/kornia/filters/kernels_geometry.py", line 25, in <module>
    from kornia.geometry.transform import rotate, rotate3d
  File ".../python3.12/site-packages/kornia/geometry/__init__.py", line 28, in <module>
    from .epipolar import *
  File ".../python3.12/site-packages/kornia/geometry/epipolar/__init__.py", line 18, in <module>
    from ._metrics import (
  File ".../python3.12/site-packages/kornia/geometry/epipolar/_metrics.py", line 138, in <module>
    @torch.jit.script
     ^^^^^^^^^^^^^^^^
  File ".../python3.12/site-packages/torch/jit/_script.py", line 1488, in script
    warnings.warn(
DeprecationWarning: `torch.jit.script` is deprecated. Please switch to `torch.compile` or `torch.export`.
```

The source of the warning in pytorch: https://github.com/pytorch/pytorch/blob/bec164b201e9dd53069ad437e4e5d985140a2454/torch/jit/_script.py#L366

Looking at the deprecation warning it may affect production for python 3.14+

> `torch.jit.script_method` is not supported in Python 3.14+ and may break.

### 🔄 Steps to Reproduce

```bash
1. pip install kornia
2. python -W error -c "import kornia"
```

### 💻 Minimal Code Example

```python
pip install kornia
python -W error -c "import kornia"
```

### ✅ Expected behavior

We raise warnings in our CIs to detect upstream deprecations. Ideally using `-W error` would not raise an exception when using `kornia`.

### ❌ Actual behavior

Use `torch.compile` or `torch.export` to avoid the deprecation warning and fix potential issues for python 3.14+

### 🔧 Environment

```shell
- Kornia version:  0.8.3
- PyTorch version: 2.12.0
[...]

PR #3757, opened 2026-06-22
fix(geometry): remove import-time torch.jit.script decorators

VanniLeonardo:
## 📝 Description

**Fixes/Relates to:** #3727

`torch.jit.script` is deprecated (PyTorch ≥2.12 emits a deprecation warning, and
it is unsupported on Python 3.14+). Kornia applied `@torch.jit.script` to nine
functions in `kornia/geometry/epipolar/`, which compiles them **at import time**.
As a result, `python -W error -c "import kornia"` fails on affected PyTorch
versions.

This PR removes those import-time decorators. The decorated functions are pure
tensor operations that behave identically in eager mode, so this neither changes
results nor migrates to `torch.compile`/`torch.export` (kept intentionally out of
scope per the issue).

---

## 🛠️ Changes Made
- [x] Removed `@torch.jit.script` from `sampson_epipolar_distance` (`_metrics.py`); ruff dropped the then-unused `import torch`.
- [x] Removed `@torch.jit.script` from `_isclose0`, `run_7point`, `run_8point` (`fundamental.py`).
- [x] Removed `@torch.jit.script` from `_multiply_deg_one_poly`, `_multiply_deg_two_one_poly`, `_determinant_to_polynomial_jit`, `_solve_2x2_tikhonov_safe`, `_null_to_Nister_solution_script` (`essential.py`).
- [x] Added `tests/test_import.py` regression test.

---

## 🧪 How Was This Tested?
- [x] **Unit Tests:** new `tests/test_import.py::test_import_without_jit_script_deprecation` — imports kornia in a fresh interpreter with the `torch.jit.script` deprecation promoted to an error.
- [x] **Manual Verification:** full `tests/geometry/epipolar/` + `tests/geometry/test_ransac.py` pass identically before and after the change (no behavior delta); `pixi run lint` and `pixi run typecheck` clean.
- [x] **Performance/Edge Cases:** these solvers ran through the TorchScript interpreter in eager mode and now run pure-Python + ATen; the difference is negligible for these linalg-dominated functions. CUDA-gated branches (Sampson matmul, 8-point GEMM) verified on GPU.

---

## 🕵️ AI Usage Disclosure
- [x] 🟡 **AI-assisted:** I used AI for the audit and edits but have manually reviewed and tested every line and can explain the logic.

---

## 🚦 Checklist
- [ ] I am assigned to the linked issue (required before PR submission)
- [x] I have performed a self-review of my code (no ghost variables or hallucinations).
- [x] I have added tests that prove my fix is effective.

---

## 💭 Additional Context
`torch.jit.script_if_tracing` or lazy compilation
would only move the deprecated call out of the import path; it would still warn at
runtime under `-W error` and won't survive the removal of TorchScript on Python 3.14+.

On PyTorch versions that don't yet emit the deprecation, the
test is a forward-looking guard (passes trivially). It filters specifically on the
`torch.jit.script` message rather than using a blanket `-W error` import, to avoid
coupling to unrelated third-party deprecation warnings.

### DIFF
--- a/kornia/geometry/epipolar/_metrics.py
+++ b/kornia/geometry/epipolar/_metrics.py
@@ -17,7 +17,6 @@
 
 """Module including useful metrics for Structure from Motion."""
 
-import torch
 from torch import Tensor, ones_like
 
 from kornia.core.check import KORNIA_CHECK_IS_TENSOR
@@ -135,7 +134,6 @@ def _sampson_epipolar_distance_matmul_impl_(
     return (out + eps).sqrt()
 
 
-@torch.jit.script
 def sampson_epipolar_distance(
     pts1: Tensor,
     pts2: Tensor,
--- a/kornia/geometry/epipolar/essential.py
+++ b/kornia/geometry/epipolar/essential.py
@@ -79,7 +79,6 @@ def run_5point(points1: torch.Tensor, points2: torch.Tensor, weights: Optional[t
     return E
 
 
-@torch.jit.script
 def _multiply_deg_one_poly(a: torch.Tensor, b: torch.Tensor, T_deg1: torch.Tensor) -> torch.Tensor:
     # a, b: (..., 4)
     product_basis = a.unsqueeze(2) * b.unsqueeze(1)  # (..., 4, 4)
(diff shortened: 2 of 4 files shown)


edgarriba (approved):
Would be great in upcoming PRs to verify that functions support torch.compile

PR #3794, opened 2026-07-15
refactor(compile): replace is_compiling() guards with genuine fullgraph code

edgarriba:
Follow-up to the compile-first series (#3786, #3788, #3792, #3793).

## Why

Those PRs landed with `torch.compiler.is_compiling()` guards. That pattern is a **hack**: the compiled path *skips* the data-dependent code instead of making it traceable. Eager and compile then run **different code paths**, and the guarded checks are silently dropped under compile — "fullgraph by exclusion," not genuine fullgraph.

This replaces every such guard with code that genuinely compiles fullgraph on a **single shared path** (same code in eager and compile).

## Changes

- **`enhance/adjust.py`** (`adjust_gamma`, `adjust_contrast`): the `if (gamma < 0).any(): raise` validation now uses `torch._assert_async` — the torch-native way to assert on a tensor value inside a graph. Keeps the check, no graph break, same path in eager and compile.
- **`contrib/distance_transform.py`**: delete the `if not mask.any(): break` early-out entirely. The loop bound (`n_iters`) is already fixed, and once no pixels remain the trailing iterations are **exact no-ops** (`out += 0*mask`, `where(False, …)` leaves `boundary` unchanged) — so removing the break changes nothing numerically but makes the loop fullgraph.
- **`losses/_utils.py`** (`mask_ignore_pixels`): drop the `if target_mask.all(): return None` early-out; always return the mask. When nothing is ignored it is all-`True`, so the caller's masking is a no-op — identical result, no data-dependent branch. (Unblocks dice/focal/tversky on a single path.)
- **`augmentation/base.py`**: replace the eager/compile `is_compiling` split with a single **branchless** `batch_prob * elem_prob` gate (`batch_prob` is `(1,)` in {0, 1}: selected → `elem_prob`, deselected → zeros).

The `torch.where` rewrites from the earlier PRs (posterize, sharpness, apply_colormap) were already genuine single-path fixes and are untouched.

## Verification

```
torch.compile(fullgraph=True) on adjust_gamma / adjust_contrast /
  distance_transform / dice_loss / RandomBrightness / Normalize
  -> all trace clean on the SAME path they run in eager, matching eager numerically

pytest tests/enhance/test_adjust.py tests/losses tests/contrib/test_conv_distance_transformer.py -> 466 passed
pytest tests/augmentation --dtype=float32 -> 2280 passed (unchanged)
```

The augmentation change alters RNG *consumption* only for the rare `p_batch < 1` deselect case (elem_prob is now always drawn); the full suite passing unchanged confirms no seeded test depends on it.

ruff clean.

### DIFF
--- a/kornia/augmentation/base.py
+++ b/kornia/augmentation/base.py
@@ -175,34 +175,22 @@ class _BasicAugmentationBase(nn.Module):
         else:
             batch_prob = (torch.rand(1, device=self.device) < p_batch).to(self.dtype)
 
+        # Per-sample gate. `p` and `same_on_batch` are Python values, so these branches are
+        # resolved at trace time (no graph break).
         elem_prob: torch.Tensor
-        if not torch.compiler.is_compiling():
-            # Eager keeps the exact original control flow so RNG draws are unchanged.
-            if batch_prob.sum() == 1:
-                if p == 1:
-                    elem_prob = torch.ones(batch_shape[0], device=self.device, dtype=self.dtype)
-                elif p == 0:
-                    elem_prob = torch.zeros(batch_shape[0], device=self.device, dtype=self.dtype)
-                elif same_on_batch:
-                    elem_prob = (torch.rand(1, device=self.device) < p).to(self.dtype).expand(batch_shape[0])
-                else:
-                    elem_prob = (torch.rand(batch_shape[0], device=self.device) < p).to(self.dtype)
-                batch_prob = batch_prob * elem_prob
-            else:
-                batch_prob = batch_prob.repeat(batch_shape[0])
+        if p == 1:
+            elem_prob = torch.ones(batch_shape[0], device=self.device, dtype=self.dtype)
+        elif p == 0:
+            elem_prob = torch.zeros(batch_shape[0], device=self.device, dtype=self.dtype)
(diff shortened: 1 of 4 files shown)


edgarriba:
## Benchmark — eager vs compiled (each touched function)

Measured with `torch.utils.benchmark` on CPU (4 threads, input `(8, 3, 128, 128)`; distance_transform `(8,1,64,64)`; dice_loss `(8,4,64,64)`). The genuine fullgraph fixes are what let the **whole** op compile — the `is_compiling` hack would leave it partially uncompiled and forfeit this.

| op | eager (µs) | compiled (µs) | speedup |
|---|---:|---:|---:|
| adjust_gamma | 9396 | 9007 | 1.04× |
| adjust_contrast | 421 | 239 | **1.77×** |
| distance_transform | 123721 | 93546 | **1.32×** |
| dice_loss | 3028 | 1506 | **2.01×** |

adjust_gamma is dominated by the `pow` op, so compile helps least there. All fixes also keep eager parity (no eager regression from the branchless rewrites / deleted early-exits — confirmed by the unchanged 2280-test augmentation suite and 466-test enhance/losses/distance suites). GPU speedups are expected to be larger via kernel fusion.

PR #3878, opened 2026-07-31
perf(enhance): vectorize histogram lut computation in equalize (#3728)

### LINKED ISSUE #3728: Vectorize equalize/equalize3d to remove per-channel GPU sync and Python-level loops
### 🚀 Feature Description

equalize and equalize3d in kornia/enhance/adjust.py currently loop over batch and channel dimensions in Python, calling _scale_channel once per image-channel pair. Each call to _scale_channel performs .item() calls on tensor min/max (forcing a CPU-GPU synchronization), computes a per-channel torch.histc, and does a torch.gather for the LUT lookup. For a batch of N images with C channels, this results in N×C serialized kernel launches and synchronization points, each carrying CPU-GPU round-trip overhead.

### 📂 Feature Category

Performance Engineering

### 💡 Motivation

equalize and equalize3d are commonly used as part of data augmentation pipelines, which run repeatedly inside the training loop (often once per sample per epoch). The current implementation's reliance on .item() calls and per-channel Python loops introduces avoidable host-device synchronization, which can become a meaningful bottleneck at scale, especially for larger batch sizes or video/volumetric data in equalize3d where the channel-loop is nested inside a batch-loop with an additional depth dimension.

### 💭 Proposed Solution

Replace the nested Python loops with a vectorized implementation:

Build a batched histogram across all (batch × channel) slices in a single pass, e.g. via scatter_add_ on a flattened index tensor instead of calling torch.histc per channel.
Remove or gate the .item()-based bounds checks in _scale_channel behind an optional validation flag, since they force synchronization on every call.
Compute the LUT and torch.gather step in a batched form across all channels/images at once rather than per-slice.
Preserve existing behavior and output values exactly (this should be a pure performance refactor, not a behavior change).

### 🔄 Alternatives Considered

Keep the per-channel loop but drop only the .item() calls — smaller win, doesn't address the fundamentally serialized histogram/gather computation.
Use torch.vmap over _scale_channel — simpler to implement but torch.histc and .item() calls inside are not vmap-compatible without further rework, so this would still require rewriting _scale_channel internals.

### 🎯 Use Cases

Training pipelines using kornia.enhance.equalize / equalize3d as part of on-the-fly augmentation, where reduced per-call overhead compounds across batches and epochs.
Volumetric/video data pipelines using equalize3d, where the current implementation's loop depth (batch × channel) is higher and the relative overhead is larger.

### 📝 Additional Context

Will include before/after benchmarks (e.g. via torch.utils.benchmark or a Criterion-style harness) on representative batch sizes/shapes to quantify the speedup. Output correctness will be validated against the existing implementation via torch.testing.assert_close on a range of inputs.
_No response_

### 🤝 Contribution Intent

Prakhar00001:
Fixes #3728

Vectorizes tile histogram LUT computation using `torch.vmap` in `kornia/enhance/equalization.py`, eliminating Python-level sequential loops and reducing CPU/GPU synchronization bottlenecks.


kornia/enhance/equalization.py
```diff
@@ -379,9 +381,6 @@ def equalize_clahe(
 
     imgs: torch.Tensor = input  # B x C x H x W
 
-    # hist_tiles: torch.Tensor  # B x GH x GW x C x TH x TW  # not supported by JIT
```
edgarriba, line 382: Why remove comments?

edgarriba (changes requested):
Please provide some comparison performance numbers

ducha-aiki:
vmap has no batching rule for aten::histc, so it falls back to a serial loop

PR #3888, opened 2026-08-05
Make check_doc_links.py pass pre-commit

edgarriba:
**This fixes a regression I introduced in #3884.** That PR added `.github/scripts/check_doc_links.py`, which does not satisfy this repo's pre-commit hooks — so `pre-commit.ci` now fails on pull requests against `main`. I found it because it turned #3887 red, and the failure had nothing to do with that PR's change.

My mistake was narrow and worth naming: I verified the script ran correctly and that its own workflow passed, but I never ran the repository's own hooks against it.

## What ruff reported

```
D103 x3  missing docstrings on markdown_files, probe, main
S310 x2  urllib opening a URL without an audited scheme
C901     main is too complex (22 > 20)
```

## Fixes

- **Docstrings** on every public function.
- **S310** — `probe()` now rejects any scheme other than `http`/`https` before building the request. That is a real guard, not a suppression; the two remaining `S310` markers are `noqa` comments that point at it. Worth having on its own merits, since the URLs come from markdown files and were previously passed to `urlopen` unchecked.
- **C901** — `main()` split into `collect()`, `classify_target()`, `report_external()` and `report_emails()`. Behaviour is unchanged and each piece is now independently testable.
- **Apache licence header** added, since `add-license-header` applies to all Python files. Added by running the hook rather than hand-writing it.

## Verification

```
$ pre-commit run --files .github/scripts/check_doc_links.py
ruff check ..... Passed
ruff format .... Passed
codespell ...... Passed
Add license header ... Passed
(all other hooks Passed)
```

Behaviour re-checked after the refactor, not assumed:

```
('dead', 'DNS')       https://bubbaloop.kornia.org
('ok', 200)           https://www.kornia.org/
('dead', 404)         https://github.com/kornia/kornia/blob/main/NOPE_DOES_NOT_EXIST.md
('unverified', 'unsupported scheme')  ftp://example.com/x
```

Offline scan still reports clean on this repo, and re-injecting a missing relative link still fails the run.

## Note for the other repos

`kornia/.github` and `kornia-slam` carry copies of this script. Neither runs ruff, so neither is broken — but I will sync the improved version to both so they do not drift.

### DIFF
--- a/.github/scripts/check_doc_links.py
+++ b/.github/scripts/check_doc_links.py
@@ -1,4 +1,22 @@
 #!/usr/bin/env python3
+
+# LICENSE HEADER MANAGED BY add-license-header
+#
+# Copyright 2018 Kornia Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
 """Check that links in the org's markdown documents still resolve.
 
 Written after an audit found several dead pointers in Kornia's docs that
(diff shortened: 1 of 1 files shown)


ducha-aiki:
Closing as superseded: everything this PR set out to fix landed on main via #3891's lint pass on `check_doc_links.py` (docstrings for `markdown_files`/`probe`/`main`, scoped `# noqa: S310` on the audited urlopen calls, `# noqa: C901`) plus pre-commit.ci's license header and formatting — `pixi run lint` and pre-commit are green on main. The branch is now conflicted against that overlapping fix, and rebasing it would leave an empty diff. The regression it fixed is also structurally prevented now: lint runs repo-wide on every PR that touches the file.

*— comment written by Claude (Claude Code) on behalf of @ducha-aiki*

PR #3889, opened 2026-08-06
feat(metrics): add pose error metrics and pose AUC

### LINKED ISSUE #3753: [Feature]: Pose error calculation
### 🚀 Feature Description

Few methods that can be used during benchmark to compute rotation, translation error in degrees. Also a method to compute AUC given some errors.

### 📂 Feature Category

Geometry

### 💡 Motivation

I think kornia should provide a standardised way to for from matched keypoins to error / AUC. this would avoid everyone to write their own methods (sometime being creative  with numbers).

### 💭 Proposed Solution

A method that given matches and intrinsics uses kornia methods to return a pose. then a method that given a pose and a reference pose (likely GT) return the error in degrees. This can be extended to metric distance eventually. Finally a method that given a list of errors return the AUC at various thresholds. Plus the helpers.

It's all in torch, no other libs. 

### 🔄 Alternatives Considered

_No response_

### 🎯 Use Cases

_No response_

### 📝 Additional Context

https://github.com/mattiadurso/kornia/tree/metric_pose

### 🤝 Contribution Intent

atikulmunna:
## 📝 Description

Closes #3753

Adds pose-error metrics and pose AUC to `kornia.metrics`: rotation and translation error in degrees between two relative poses, plus AUC at the usual thresholds.

Started from @mattiadurso's `metric_pose` branch, which he offered in #3753. His commit is cherry-picked with authorship preserved. I rebased onto main, resolved a docs conflict (main added a Stereo section where the branch added Pose, both kept), fixed the defects listed below, and added regression tests.

References: the rotation error is the geodesic angle `acos((tr(R1ᵀR2) - 1) / 2)`; the AUC protocol and the (1, 3, 5, 10) degree defaults follow SuperGlue (Sarlin et al., CVPR 2020) and the Image Matching Challenge.

---

## 🛠️ Changes Made
- [x] `kornia/metrics/pose.py`: `angle_error_mat`, `angle_error_vec`, `translation_ate`, `pose_errors`, `auc_from_errors`
- [x] Exported from `kornia.metrics`, documented in `docs/source/metrics.rst`
- [x] `tests/metrics/test_pose.py`: `BaseTester` tests for all five

Defects fixed, each with a regression test:

- All four metrics accepted mismatched shapes and let broadcasting paper over it. `translation_ate` decided whether to unsqueeze from `t.dim()` alone, so an unbatched `t` against a batched `t_gt` gave `(1, B)`; `angle_error_mat` and `angle_error_vec` paired a single input against a whole batch and returned `(B,)`. All four now require matching shapes.
- `auc_from_errors` built its comparison value with `errors.new_tensor(...)`, so an integer error tensor truncated the threshold: `2.5` became `2`, turning an AUC of 60 into 0. Non-float inputs are promoted up front.
- Non-positive thresholds were unvalidated and returned `nan` for 0 and `-0.0` for negatives. They now raise, which also makes the `recall[last - 1 : last]` slice provably non-empty.
- A negative error broke the sort order that `searchsorted` relies on, so `errors=[-5, 1]` at `thr=2` returned an AUC of 212.5. Negative errors now raise, matching the documented non-negative contract.
- `NaN` propagation depended on the threshold, because `NaN` sorts last and a low threshold cut it off before integration: `[1, 5, nan]` gave 25.0 at `thr=2` but `nan` at `thr=10`. Any `NaN` now returns `NaN` at every threshold.
- The `searchsorted` comment named the wrong index and called the closing segment vertical when it is flat.

---

## 🧪 How Was This Tested?
- [x] **Unit Tests:** 31 tests plus 5 doctests, over float32 and float64, with gradcheck on the three differentiable functions.
- [x] **Manual Verification:** expected values are hand-checkable, e.g. `errors=[2]` at threshold 4 gives the curve `y=[0,1,1]` over `x=[0,2,4]`, area 3 of 4, so AUC 75.
- [x] **Edge Cases:** errors above the threshold give 0, empty errors give 0, empty or non-positive thresholds raise, negative or `NaN` errors are rejected or propagate deterministically, mismatched shapes raise on all four metrics, batched `(B,3,4)`/`(B,4,4)` poses return `(B,)` and bad ranks raise, and the AUC is invariant to input order and shape.

```
$ pixi run -e default uv run pytest tests/metrics/test_pose.py --doctest-modules kornia/metrics/pose.py --dtype=float32,float64 -q

collected 67 items

tests/metrics/test_pose.py ............................................. [ 67%]
.................                                                        [ 92%]
kornia/metrics/pose.py .....                                             [100%]

============================= 67 passed in 5.67s ==============================
```

`pixi run lint` and `ty check kornia/metrics/pose.py` are clean, and the full suite passes on this branch.

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**

The implementation is @mattiadurso's, with his authorship intact. The rebase, the fixes above and the tests are mine, and AI helped with the write-up. I have read through the code and can explain it.

---

## 🚦 Checklist
- [x] I am assigned to the linked issue (approved by @edgarriba in the thread; the assignee field was not set)

---

## 💭 Additional Context

@mattiadurso, tagging you since you wanted to use these from kornia.

Four calls I did not want to make on someone else's API:

1. **Naming.** `translation_ate` is a single pairwise distance, but ATE in the SLAM literature (Sturm et al.) is a trajectory RMSE after alignment. `translation_error` would avoid the collision, and renaming is cheaper now than after release.
2. **Return ranks differ.** An unbatched input gives 0-dim from `angle_error_mat` and `angle_error_vec`, but `(1,)` from `translation_ate` and `pose_errors`. Happy to unify either way.
3. **Zero-length vectors** give `NaN`, which propagates into `max_err` and then the AUC. Documented rather than guarded, since the angle is genuinely undefined there. A `norms.clamp_min(eps)` is a one-liner if you would rather it be safe by default.
4. **`auc_from_errors` will not `torch.jit.script`** (Python loop, `.item()`, float-keyed dict), and `.item()` syncs per threshold. Can be vectorized with a single `searchsorted` over a threshold tensor if that matters here.

### DIFF
--- a/.github/scripts/check_doc_links.py
+++ b/.github/scripts/check_doc_links.py
@@ -1,4 +1,22 @@
 #!/usr/bin/env python3
+
+# LICENSE HEADER MANAGED BY add-license-header
+#
+# Copyright 2018 Kornia Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
 """Check that links in the org's markdown documents still resolve.
 
 Written after an audit found several dead pointers in Kornia's docs that
@@ -44,9 +62,7 @@ errors: list[str] = []
 notes: list[str] = []
 
 # Things that look like an address but are not a mailbox.
-NOT_MAILBOXES = re.compile(
-    r"@(?:\d+x)?\.(?:png|jpe?g|svg|gif|webp|ico)$|^git@", re.I
-)
+NOT_MAILBOXES = re.compile(r"@(?:\d+x)?\.(?:png|jpe?g|svg|gif|webp|ico)$|^git@", re.I)
 
 COMMENT_RE = re.compile(r"<!--.*?-->", re.S)
 FENCE_RE = re.compile(r"^```.*?^```", re.S | re.M)
@@ -147,10 +163,11 @@ def main() -> int:
             emails.setdefault(addr, []).append(str(rel))
 
     mode = "offline" if offline else "networked"
-    print(f"{len(files)} markdown file(s), {len(external)} external link(s), "
-          f"{len(emails)} email address(es)  [{mode}]\n")
+    print(
+        f"{len(files)} markdown file(s), {len(external)} external link(s), {len(emails)} email address(es)  [{mode}]\n"
+    )
 
-    for url in (() if offline else sorted(external)):
+    for url in () if offline else sorted(external):
         verdict, status = probe(url)
         where = ", ".join(sorted(set(external[url])))
         if verdict == "dead":
@@ -167,8 +184,9 @@ def main() -> int:
             verdict, status = ("skipped", "-") if offline else probe(f"https://{domain}")
             flag = "  <-- domain does not resolve" if (verdict, status) == ("dead", "DNS") else ""
             if flag:
-                errors.append(f"{', '.join(sorted(set(emails[addr])))}: "
-                              f"email domain {domain} does not resolve ({addr})")
+                errors.append(
+                    f"{', '.join(sorted(set(emails[addr])))}: email domain {domain} does not resolve ({addr})"
+                )
             print(f"  {addr:<40} {', '.join(sorted(set(emails[addr])))}{flag}")
 
     if notes:
--- a/docs/source/metrics.rst
+++ b/docs/source/metrics.rst
@@ -48,6 +48,15 @@ Stereo
 .. autofunction:: root_mean_squared_disparity_error
 .. autofunction:: mean_bad_pixel_error
 
+Pose
+----
+
+.. autofunction:: angle_error_mat
+.. autofunction:: angle_error_vec
+.. autofunction:: translation_ate
+.. autofunction:: pose_errors
+.. autofunction:: auc_from_errors
+
 Monitoring
 ----------
 
--- a/kornia/metrics/__init__.py
+++ b/kornia/metrics/__init__.py
@@ -28,6 +28,7 @@ from .disparity import mean_absolute_disparity_error, mean_bad_pixel_error, root
 from .endpoint_error import AEPE, aepe, average_endpoint_error
 from .mean_average_precision import mean_average_precision
 from .mean_iou import mean_iou, mean_iou_bbox
+from .pose import angle_error_mat, angle_error_vec, auc_from_errors, pose_errors, translation_ate
 from .psnr import psnr
 from .ssim import SSIM, ssim
 from .ssim3d import SSIM3D, ssim3d
@@ -39,6 +40,9 @@ __all__ = [
     "AverageMeter",
     "accuracy",
     "aepe",
+    "angle_error_mat",
+    "angle_error_vec",
+    "auc_from_errors",
     "average_endpoint_error",
     "confusion_matrix",
     "mean_absolute_disparity_error",
@@ -46,8 +50,10 @@ __all__ = [
     "mean_bad_pixel_error",
     "mean_iou",
     "mean_iou_bbox",
+    "pose_errors",
     "psnr",
     "root_mean_squared_disparity_error",
     "ssim",
     "ssim3d",
+    "translation_ate",
 ]
--- a/kornia/metrics/pose.py
+++ b/kornia/metrics/pose.py
@@ -30,7 +30,12 @@ from collections.abc import Sequence
 import torch
 from torch import Tensor
 
-from kornia.core.check import KORNIA_CHECK, KORNIA_CHECK_IS_TENSOR, KORNIA_CHECK_SHAPE
+from kornia.core.check import (
+    KORNIA_CHECK,
+    KORNIA_CHECK_IS_TENSOR,
+    KORNIA_CHECK_SAME_SHAPE,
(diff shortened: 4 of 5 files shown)


atikulmunna:
Thanks, three of these were real and are fixed in a96d26f.

**Angular metrics silently broadcast.** Confirmed. `angle_error_vec` with a `(1, 3)` against a `(5, 3)` returned shape `(5,)` instead of raising, and `angle_error_mat` did the same with `(1, 3, 3)` against `(4, 3, 3)`. Both now use `KORNIA_CHECK_SAME_SHAPE`, so all four functions in the module are consistent.

**Negative errors break sorting.** Confirmed, and it produces an out-of-range result rather than just a wrong one: `errors=[-5, 1]` at `thr=2` returned an AUC of 212.5. Negative errors now raise, which matches the documented non-negative contract.

**NaN AUC policy not enforced.** Confirmed, and the inconsistency is worse than described. Propagation depended on the threshold, since NaN sorts last and a low threshold cuts it off before the integration: `[1, 5, nan]` gave 25.0 at `thr=2`, 27.8 at `thr=3`, but `nan` at `thr=10`. Any NaN now returns NaN at every threshold, so the behaviour the docstring already promised is actually guaranteed.

Each has a regression test. 67 tests pass over float32 and float64, and the full suite is green.

On the other three:

**Line length in `pose.py` and `test_pose.py`.** I do not think these hold. The longest line in either file is 114 characters against the `line-length = 120` in `pyproject.toml`, and `ruff check` plus `ruff format --check` are both clean on both files. Happy to look again if you are measuring something different.

**`from __future__ import annotations` in the test module.** 12 of 253 files under `tests/` use it, and 3 of 11 in `tests/metrics/`, so adding it would make this file the exception rather than bring it in line. Glad to add it if the intent is for test modules to converge on it.

Unrelated to the review: pre-commit.ci pushed 0113d44 to this branch, which adds the license header to `.github/scripts/check_doc_links.py`. That file is not part of this PR's scope, it just failed the hook on main. Let me know if you would rather I drop it.

atikulmunna:
Both red checks look unrelated to this PR:

- **pre-commit.ci** fails on 6 ruff errors in `.github/scripts/check_doc_links.py` (3x D103, 2x S310, C901). The same 6 are present on `main`, and #3888 already fixes them. That file is only in this PR's diff because pre-commit.ci pushed the license header to the branch in 0113d44.
- **tutorials (3.12)** fails with `TimeoutError: [Errno 110] Connection timed out` from a notebook cell fetching data, so a network flake rather than anything here.

Everything else is green: the full test matrix across ubuntu, macos and windows, float32 and float64, Python 3.11 to 3.13, torch 2.5.1 and 2.9.1, plus typecheck, dynamo, the docs build and the doc link check.

Happy to rebase once #3888 lands so pre-commit goes green here too.

edgarriba:
Make sure ci is green

atikulmunna:
@edgarriba fixed what was actually mine and pushed 51d7dbc:

- Added `from __future__ import annotations` to the test module. CONTRIBUTING.md asks for it on non-JIT modules, so the bot was right and my earlier pushback was not.
- The description said "Fixes/Relates to: #3753", which `pr-validation.yml` does not match. It now says `Closes #3753`.

The tutorials job is green on this run, so the earlier `TimeoutError: [Errno 110]` was the network flake it looked like. 33 of 37 checks pass now, including the whole Linux and macOS matrix, both dtypes, typecheck, dynamo, docs and the doc-link check.

The other two reds are not from this PR, and I can show it rather than just claim it.

### Windows test jobs

Two `windows-latest, float32` jobs died with a native crash, not a test failure:

```
Windows fatal exception: code 0xc000001d
  File "torch\nn\functional.py", line 5210 in affine_grid
  File "kornia\geometry\transform\imgwarp.py", line 1017 in warp_affine3d
  File "kornia\augmentation\_3d\geometric\affine.py", line 179 in apply_transform
tests/augmentation/test_base.py::TestGeometricAugmentationBase3D::test_autocast[cpu-float32-batch_prob0]
```

`0xc000001d` is STATUS_ILLEGAL_INSTRUCTION. This PR adds `kornia/metrics/pose.py` and its tests and touches nothing in `augmentation/` or `geometry/`.

Your own #3888, which changes one file under `.github/scripts/`, crashed the same way on 2026-08-05: same exit code, same `affine_grid` frame, same test, two `windows-latest, float32` jobs ([92469215612](https://github.com/kornia/kornia/actions/runs/31054557727/job/92469215612), [92469215771](https://github.com/kornia/kornia/actions/runs/31054557727/job/92469215771)). The matrix entries it picks vary between runs, yours were 3.12/2.5.1 and 3.13/2.9.1, mine were 3.11/2.5.1 and 3.12/2.9.1. The previous run of this same branch had all Windows jobs green.

Worth a separate issue. It looks like `affine_grid` hitting an instruction some Windows runner SKUs do not have, and it will keep randomly reddening unrelated PRs. Note the scheduled `main` run does not catch it because `tests-cpu-windows` is skipped there.

### pre-commit.ci

`main` is red on this too:

| commit | pre-commit.ci |
| --- | --- |
| ce23f1b (main) | failure |
| c4c1430 (main, merge of #3884) | failure |
| ea1d741 (main, before #3884) | success |

Every hook passes on this branch except `ruff-check`, whose 6 errors are all in `.github/scripts/check_doc_links.py`:

```
check_doc_links.py:84:5   D103      check_doc_links.py:96:14  S310
check_doc_links.py:93:5   D103      check_doc_links.py:111:5  C901
check_doc_links.py:94:11  S310      check_doc_links.py:111:5  D103
```

Nothing in `kornia/metrics/pose.py` or `tests/metrics/test_pose.py` is flagged. That file is only in this diff because pre-commit.ci pushed the license header to the branch in 0113d44, and since it runs `--all-files` it fails the same way regardless.

#3888 fixes exactly those 6. Merge it and I will rebase straight away and drop the stray header hunk. I would rather not copy your fix in here and hand you a conflict.

ducha-aiki (approved):

ducha-aiki:
Thank you for the contribution!

Issue #3908, opened 2026-08-07
StableDiffusionDissolving crashes on float64 input (scheduled coverage/slow float64 job red)


ducha-aiki:
The scheduled `coverage / slow (float64)` job fails in `tests/filters/test_dissolving.py` (run 31193743729):

```
RuntimeError: Input type (double) and bias type (float) should be the same
```

`_DissolvingWraper_HF.encode_tensor_to_latent` moves the input image to the pipeline device but keeps the caller dtype, while the SD VAE weights stay float32 — so any non-float32 input crashes in the first conv. The test suite already documents the intended contract ('the model internally converts latents to its own dtype') but the conversion is missing.

PR #3909, opened 2026-08-07
fix(filters): cast dissolving inputs to the VAE dtype, not only device

ducha-aiki:
Fixes 3908 (https://github.com/kornia/kornia/issues/3908).

The scheduled `coverage / slow (float64)` job fails in `TestStableDiffusionDissolving` with `RuntimeError: Input type (double) and bias type (float) should be the same`. Root cause: `encode_tensor_to_latent` does `.to(self.model.device)` — device only — while the SD VAE weights stay in the pipeline dtype (float32), so the float64 suite's double input crashes in the VAE's first conv. `decode_tensor_to_latent` has the same latent-dtype hole for caller-provided latents.

Fix: cast to `self.model.vae.dtype` alongside the device move in both entry points. This matches the contract the test already documents ("the model internally converts latents to its own dtype", `tests/filters/test_dissolving.py:49`) and keeps float32 behavior byte-identical (the cast is a no-op there). Output stays in the model dtype, as the tests expect.

Reproduced locally before the fix (same three failures as CI), verified after:

```
# before (main):
FAILED tests/filters/test_dissolving.py::TestStableDiffusionDissolving::test_encode_tensor_to_latent[cpu-float64]
FAILED tests/filters/test_dissolving.py::TestStableDiffusionDissolving::test_decode_tensor_to_latent[cpu-float64]
FAILED tests/filters/test_dissolving.py::TestStableDiffusionDissolving::test_dissolve[cpu-float64]
3 failed, 2 passed, 1 warning in 535.83s (0:08:55)

# after (this branch):
$ KORNIA_TEST_RUNSLOW=true pixi run test tests/filters/test_dissolving.py --dtype=float32,float64
8 passed, 1 warning in 55.42s
$ pixi run lint
ruff check...Passed
ruff format...Passed
```

Algorithm source: n/a — dtype-handling fix in the existing HF-diffusers wrapper.

### DIFF
--- a/kornia/filters/dissolving.py
+++ b/kornia/filters/dissolving.py
@@ -15,6 +15,8 @@
 # limitations under the License.
 #
 
+from __future__ import annotations
+
 import os
 from typing import Any
 
@@ -140,7 +142,7 @@ class _DissolvingWraper_HF:
     @torch.no_grad()
     def encode_tensor_to_latent(self, image: torch.Tensor) -> torch.Tensor:
         with torch.no_grad():
-            # The VAE weights stay in the pipeline dtype (float32), so inputs of any
+            # The VAE weights stay in the pipeline dtype (e.g. float32), so inputs of any
             # user dtype (e.g. float64) must be cast to it, not only moved to the device.
             image = (image / 0.5 - 1).to(device=self.model.device, dtype=self.model.vae.dtype)
             latents = self.model.vae.encode(image)["latent_dist"].sample()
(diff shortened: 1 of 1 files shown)


ducha-aiki:
Addressed all three review comments:

- **`from __future__ import annotations` (qodo #1): added** — required for touched non-JIT modules per CLAUDE.md.
- **stale 'in-place detach' comment (qodo #2): rewritten** — now states what the line actually does (detach from autograd, then match VAE device/dtype, allocating only on mismatch).
- **comment hardcoding float32 (Copilot): rephrased** as an example dtype — the code reads `self.model.vae.dtype`, so fp16/bf16-loaded pipelines are handled identically.

Re-verified after the changes: `8 passed` on `--dtype=float32,float64`, lint clean.

Issue #2950, opened 2024-07-09
Understanding `kornia.geometry.homography_warp` coordinate system


senyai:
### Describe the bug

Hello! This is my first time with kornia. I started with `homography_warp` function. Let's say I have an image of size 8x16 and I want to flip it with this function. Not because I want to flip an image, but because I want to understand the math behind it.

To flip an image along Y axis I used this matrix:
```
  (1.0, 0.0, 0.0),
  (0.0, -1.0, 8.0),
  (0.0, 0.0, 1.0),
```
Unfortunately the output image is not only flipped, but also shifted 1px:

### Reproduction steps

```bash
from torch import zeros, tensor, float32
import kornia


image = zeros(1, 1, 8, 16)
image[0, 0, 2:6:2, 2:14] = 1.0
image[0, 0, :, 2] = 1.0
mats = [(
    (1.0, 0.0, 0.0),
    (0.0, -1.0, 8.0),  # why 7 works?
    (0.0, 0.0, 1.0),
)]

mat = tensor(mats, dtype=float32)
out = kornia.geometry.homography_warp(
    image,
    mat,
    (8, 16),
    normalized_coordinates=False,
    normalized_homography=False,
)

print("input:")
for line in image[0, 0].numpy().tolist():
    print([round(val, 4) for val in line])

print("output:")
for line in out[0, 0].numpy().tolist():
    print([round(val, 4) for val in line])
```

### Expected behavior

And if I replace "8" with "7", I see 99.99% correct image. My assumption is that top left pixel's top left corner has coordinates (0,0) and bottom right corner of that pixel has coordinates (1,1). If that is correct, then my initial flip matrix is correct and there's a bug. If that assumption is incorrect, then coordinate system should be mentioned in the documentation.

### Environment

```shell
Collecting environment information...
PyTorch version: 2.3.1+cpu
Is debug build: False
CUDA used to build PyTorch: Could not collect
ROCM used to build PyTorch: N/A

OS: Fedora Linux 39 (Thirty Nine) (x86_64)
GCC version: (GCC) 13.3.1 20240522 (Red Hat 13.3.1-1)
Clang version: 17.0.6 (Fedora 17.0.6-2.fc39)
CMake version: version 3.27.7
Libc version: glibc-2.38

Python version: 3.12.4 (main, Jun  7 2024, 00:00:00) [GCC 13.3.1 20240522 (Red Hat 13.3.1-1)] (64-bit runtime)
[...]

PR #3924, opened 2026-08-08
docs(geometry): per-op Convention blocks for transform warps, with convention-pinning tests

### LINKED ISSUE #3923: docs: per-op Convention blocks for geometry.transform warps, with convention-pinning tests
## Summary

Continuation of the Conventions & Pitfalls work (#3893): the conventions users repeatedly trip over — issues #2950 (homography_warp coordinate system), #777/#388/#546 (align_corners defaults and churn) — are documented on one page but not in the docstrings where users and LLMs actually look, and several are enforced by no test at all.

## Proposal (spec agreed off-thread)

1. A compact `Convention:` rubric in every public docstring of the geometry.transform warp/matrix ops (imgwarp, affwarp, HomographyWarper — 39 symbols): shape, coordinate order, pixel-center origin, align_corners actual per-symbol default, rotation direction, homography direction/normalization, padding default. Every line verified from signatures or by runnable snippet — never from memory.
2. `test_convention_*` characterization tests for conventions no existing test pins (audit found five: rotate/resize/HomographyWarper bare align_corners defaults, homography_warp normalize-then-invert contract, get_rotation_matrix2d (x,y) center order).
3. Wrong existing docstring lines found by the audit get corrected in the same pass (warp_perspective3d dsize is a 3-tuple; get_affine_matrix3d returns (B,4,4); get_shear_matrix2d summary said Bx4x4).

Found along the way, noted for follow-up (not fixed here): `homography_warp(..., normalized_homography=False)` silently ignores the caller's `align_corners` and forces `True` (imgwarp.py:1424-1426) — the new docstring documents the actual behavior; `rotate3d`'s summary claims anti-clockwise rotation but no test or audit verifies it.

Follow-up batches: transform's crops/flips/pyramid/TPS, geometry.conversions, bbox/boxes, camera, then augmentation classes. PR follows.

ducha-aiki:
Fixes #3923. Stage-2 continuation of the Conventions & Pitfalls work (#3893) — the docstring half of W1, with test enforcement.

## What

`Convention:` rubric blocks in the docstrings of all **39 public symbols** of `kornia/geometry/transform/{imgwarp,affwarp,homography_warper}.py` — the ops behind #2950/#777/#388/#546 — plus **five new `test_convention_*` tests** pinning conventions that no existing test enforced, and **four corrections of wrong existing docstring lines** found by the audit.

Every rubric line is verified, not asserted: defaults are taken verbatim from signatures; behavioral claims (rotation direction, coordinate order, homography normalization direction, `get_projective_transform`'s +z handedness, `get_affine_matrix2d`'s angle-sign flip) were each confirmed by a runnable snippet whose output is recorded in the review artifacts. Where reality is surprising, the docstring says so rather than smoothing it over:

- `Rescale`/`Shear` **classes** default `align_corners=True` while their delegate **functions** default `None`/`False` — each block states its own default with an explicit "differs from" note.
- `homography_warp` honors `align_corners` only on the default `normalized_homography=True` path; the pixel-homography branch currently **forces `True`** (`imgwarp.py:1424-1426`) — documented as-is; the underlying quirk is noted in #3923 as a follow-up bug.

## Convention-pinning tests (rubric line → test)

| Convention | Previously pinned by | Now |
|---|---|---|
| `rotate` bare default `align_corners=True` | nothing (all tests passed it explicitly) | `TestRotate::test_convention_align_corners_default_true` |
| `resize` bare default `align_corners=None` | nothing | `TestResize::test_convention_align_corners_default_none` |
| `HomographyWarper`/`homography_warp` default `align_corners=False` | nothing | `TestHomographyWarper::test_convention_align_corners_default_false` |
| `homography_warp` = `warp_perspective` under normalize-then-invert | nothing (no cross-API test existed) | `TestWarpPerspective::test_convention_normalize_invert_equivalence` |
| `get_rotation_matrix2d` center is `(x, y)` | nothing order-sensitive (all existing tests symmetric) | `TestRotationMatrix2d::test_convention_center_xy_order` (asymmetric center) |
| +90° = CCW as displayed; `dsize`=(h,w); points=(x,y) | already pinned (`test_imgwarp.py::TestWarpAffine/WarpPerspective::test_translation`, `TestRotationMatrix2d::test_90deg_rotation`, audit §B citations in full) | unchanged |

All five new tests use the hardcoded-literal + generation-snippet repo pattern and were mutation-checked (each verified to fail when one expected element is flipped).

## Corrected wrong docstring lines (pre-existing bugs)

- `warp_perspective3d`: `dsize` documented as "(height, width)" but consumed as a 3-tuple → now `(depth, height, width)`.
- `get_affine_matrix3d`: `Returns:` said `(B, 3, 3)`, actually `(B, 4, 4)` (verified by execution).
- `get_shear_matrix2d`: summary said `Bx4x4`, actually `Bx3x3`.
- `docs/source/geometry.transform.rst`: `homography_warp`, `homography_warp3d`, `get_translation_matrix2d`, `BaseWarper` had no autodoc entries at all — the #2950-trap op's docs never rendered. Added.

## Verification

- kornia/ diff is **docstring lines only** — proven with an AST-based check that every changed line falls inside a docstring span (both pre- and post-image); tests/ changes are additive test methods only.
- `tests/geometry/transform/` full runs: cpu `--dtype=all` and mps `--dtype=float32,float16` — failure sets **byte-identical** to a `caa1f9a2` baseline worktree (pre-existing fp16/bf16 tolerance failures; diff of failure lists is empty); +20 new test results all green.
- Sphinx `-W --keep-going` exit 0, zero warnings; rendered HTML inspected — `Convention:` renders as a clean definition list, and the previously-missing symbols now render.
- `pre-commit run` clean on every touched file; doctests 25/25 unchanged.

```
tests/geometry/transform --device=cpu --dtype=all      → 1400 passed, 66 skipped, 133 failed*
tests/geometry/transform --device=mps --dtype=f32,f16  → 641 passed, 90 skipped, 86 failed*
* identical failure sets on unmodified caa1f9a2 (fp16/bf16 tolerance, pre-existing; diff empty)
```

Out of scope, noted in #3923: the `homography_warp` pixel-path `align_corners` bug (behavior fix, not docs); `rotate3d`'s unverified "anti-clockwise" summary claim. Next batches: crops/flips/pyramid/TPS, geometry.conversions, bbox/boxes, camera, augmentation classes.

Algorithm source: n/a — documentation and characterization tests; no numeric code changed.

---

### DIFF
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -727,14 +727,16 @@ def rescale(
     .. image:: _static/img/rescale.png
 
     Convention:
+        - ``factor`` is ``(factor_h, factor_w)`` when a pair — height first (contrast
+          :func:`scale`, whose ``scale_factor`` is x-first)
         - align_corners: ``None`` by default (follows ``torch.nn.functional.interpolate``;
           see the convention block of :func:`resize`)
         - delegates to :func:`resize` after converting ``factor`` to an output ``size``
 
     Args:
         input: The image tensor to be scale with shape of :math:`(B, C, H, W)`.
-        factor: Desired scaling factor in each direction. If scalar, the value is used
-            for both the x- and y-direction.
+        factor: Desired scaling factor as ``(factor_h, factor_w)`` — height first. If scalar,
+            the value is used for both height and width.
         interpolation:  algorithm used for upsampling: ``'nearest'`` | ``'linear'`` | ``'bilinear'`` |
             ``'bicubic'`` | ``'trilinear'`` | ``'area'``.
         align_corners: interpolation flag.
@@ -961,8 +963,8 @@ class Rescale(nn.Module):
         - See the convention block of :func:`rescale`.
 
     Args:
-        factor: Desired scaling factor in each direction. If scalar, the value is used
-            for both the x- and y-direction.
+        factor: Desired scaling factor as ``(factor_h, factor_w)`` — height first. If scalar,
+            the value is used for both height and width.
         interpolation:  algorithm used for upsampling: ``'nearest'`` | ``'linear'`` | ``'bilinear'`` |
             ``'bicubic'`` | ``'trilinear'`` | ``'area'``.
         align_corners: interpolation flag.
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -735,8 +735,8 @@ def rescale(
 
     Args:
         input: The image tensor to be scale with shape of :math:`(B, C, H, W)`.
-        factor: Desired scaling factor as ``(factor_h, factor_w)`` — height first. If scalar,
-            the value is used for both height and width.
+        factor: Desired scaling factor as ``(factor_h, factor_w)`` — height first. If a single
+            float, the value is used for both height and width (an ``int`` raises ``TypeError``).
         interpolation:  algorithm used for upsampling: ``'nearest'`` | ``'linear'`` | ``'bilinear'`` |
             ``'bicubic'`` | ``'trilinear'`` | ``'area'``.
         align_corners: interpolation flag.
@@ -963,8 +963,8 @@ class Rescale(nn.Module):
         - See the convention block of :func:`rescale`.
 
     Args:
-        factor: Desired scaling factor as ``(factor_h, factor_w)`` — height first. If scalar,
-            the value is used for both height and width.
+        factor: Desired scaling factor as ``(factor_h, factor_w)`` — height first. If a single
+            float, the value is used for both height and width (an ``int`` raises ``TypeError``).
         interpolation:  algorithm used for upsampling: ``'nearest'`` | ``'linear'`` | ``'bilinear'`` |
             ``'bicubic'`` | ``'trilinear'`` | ``'area'``.
         align_corners: interpolation flag.
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -597,7 +597,7 @@ def resize(
     .. image:: _static/img/resize.png
 
     Convention:
-        - input: :math:`(B, C, H, W)`; ``size`` is ``(h, w)``
+        - input: :math:`(*, H, W)`; ``size`` is ``(h, w)``
         - align_corners: ``None`` by default (follows ``torch.nn.functional.interpolate``;
           note :func:`warp_perspective`/:func:`rotate` default ``True``)
         - ``side`` resizing preserves aspect ratio using the named side
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -338,7 +338,8 @@ def rotate3d(
     r"""Rotate 3D the tensor anti-clockwise about the centre.
 
     Convention:
-        - ``center`` is ``(x, y, z)`` in pixels, origin at top-left; defaults to the tensor center
+        - ``center`` is ``(x, y, z)`` in pixels, origin at the top-left of the first depth
+          slice (``z = 0``); defaults to the tensor center
         - align_corners: ``False`` by default
         - padding_mode: ``'zeros'`` by default
 
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -144,6 +144,11 @@ def affine(
 
     .. image:: _static/img/warp_affine.png
 
+    Convention:
+        - ``matrix`` is the source→destination **pixel** affine matrix :math:`(B, 2, 3)`
+        - align_corners: ``True`` by default
+        - padding_mode: ``'zeros'`` by default
+
     Args:
         tensor: The image tensor to be warped in shapes of
             :math:`(H, W)`, :math:`(D, H, W)` and :math:`(B, C, H, W)`.
@@ -196,6 +201,11 @@ def affine3d(
     align_corners: bool = False,
 ) -> torch.Tensor:
     r"""Apply an affine transformation to the 3d volume.
+
+    Convention:
+        - ``matrix`` is the source→destination **pixel** affine matrix :math:`(B, 3, 4)`
+        - align_corners: ``False`` by default
+        - padding_mode: ``'zeros'`` by default
 
     Args:
         tensor: The image tensor to be warped in shapes of
@@ -255,6 +265,12 @@ def rotate(
     r"""Rotate the tensor anti-clockwise about the center.
 
     .. image:: _static/img/rotate.png
+
+    Convention:
+        - ``center`` is ``(x, y)`` in pixels, origin at top-left; defaults to the tensor center
+        - positive ``angle`` rotates counter-clockwise as displayed (y-down image axes)
+        - align_corners: ``True`` by default
+        - padding_mode: ``'zeros'`` by default
 
     Args:
         tensor: The image tensor to be warped in shapes of :math:`(B, C, H, W)`.
@@ -320,6 +336,11 @@ def rotate3d(
     align_corners: bool = False,
 ) -> torch.Tensor:
     r"""Rotate 3D the tensor anti-clockwise about the centre.
+
+    Convention:
+        - ``center`` is ``(x, y, z)`` in pixels, origin at top-left; defaults to the tensor center
+        - align_corners: ``False`` by default
+        - padding_mode: ``'zeros'`` by default
 
     Args:
         tensor: The image tensor to be warped in shapes of :math:`(B, C, D, H, W)`.
@@ -387,6 +408,11 @@ def translate(
 
     .. image:: _static/img/translate.png
 
+    Convention:
+        - ``translation`` is ``(dx, dy)`` in pixels
+        - align_corners: ``True`` by default
+        - padding_mode: ``'zeros'`` by default
+
     Args:
         tensor: The image tensor to be warped in shapes of :math:`(B, C, H, W)`.
         translation: tensor containing the amount of pixels to
@@ -436,6 +462,11 @@ def scale(
     r"""Scale the tensor by a factor.
 
     .. image:: _static/img/scale.png
+
+    Convention:
+        - ``center`` is ``(x, y)`` in pixels, origin at top-left; defaults to the tensor center
+        - align_corners: ``True`` by default
+        - padding_mode: ``'zeros'`` by default
 
     Args:
         tensor: The image tensor to be warped in shapes of :math:`(B, C, H, W)`.
@@ -498,6 +529,11 @@ def shear(
 
     .. image:: _static/img/shear.png
 
+    Convention:
+        - ``shear`` is ``(shx, shy)``
+        - align_corners: ``False`` by default
+        - padding_mode: ``'zeros'`` by default
+
     Args:
         tensor: The image tensor to be skewed with shape of :math:`(B, C, H, W)`.
         shear: tensor containing the angle to shear
@@ -559,6 +595,12 @@ def resize(
     r"""Resize the input torch.Tensor to the given size.
 
     .. image:: _static/img/resize.png
+
+    Convention:
+        - input: :math:`(B, C, H, W)`; ``size`` is ``(h, w)``
+        - align_corners: ``None`` by default (follows ``torch.nn.functional.interpolate``;
+          note :func:`warp_perspective`/:func:`rotate` default ``True``)
+        - ``side`` resizing preserves aspect ratio using the named side
 
     Args:
         input: The image tensor to be skewed with shape of :math:`(..., H, W)`.
@@ -643,6 +685,11 @@ def resize_to_be_divisible(
 ) -> torch.Tensor:
     """Resize the input tensor to be divisible by a certain factor.
 
+    Convention:
+        - align_corners: ``None`` by default; see the convention block of :func:`resize`
+        - rounds ``height``/``width`` to the nearest multiple of ``divisible_factor`` before
+          delegating to :func:`resize`
+
     Args:
         input (torch.Tensor): Input tensor to be resized.
         divisible_factor (int): The factor to which the image should be divisible.
@@ -678,6 +725,11 @@ def rescale(
 
     .. image:: _static/img/rescale.png
 
+    Convention:
+        - align_corners: ``None`` by default (follows ``torch.nn.functional.interpolate``;
+          see the convention block of :func:`resize`)
+        - delegates to :func:`resize` after converting ``factor`` to an output ``size``
+
     Args:
         input: The image tensor to be scale with shape of :math:`(B, C, H, W)`.
         factor: Desired scaling factor in each direction. If scalar, the value is used
@@ -712,6 +764,10 @@ def rescale(
 
 class Resize(nn.Module):
     r"""Resize the input torch.Tensor to the given size.
+
+    Convention:
+        - align_corners: ``None`` by default, matching :func:`resize`
+        - See the convention block of :func:`resize`.
 
     Args:
         size: Desired output size. If size is a sequence like (h, w),
@@ -782,6 +838,10 @@ class Resize(nn.Module):
 
 class Affine(nn.Module):
     r"""Apply multiple elementary affine transforms simultaneously.
+
+    Convention:
+        - align_corners: ``True`` by default, matching :func:`affine`
+        - See the convention block of :func:`affine`.
 
     Args:
         angle: Angle in degrees for counter-clockwise rotation around the center. The tensor
@@ -896,6 +956,11 @@ class Affine(nn.Module):
 class Rescale(nn.Module):
     r"""Rescale the input torch.Tensor with the given factor.
 
+    Convention:
+        - align_corners: ``True`` by default (differs from function :func:`rescale`, whose default
+          is ``None``)
+        - See the convention block of :func:`rescale`.
+
     Args:
         factor: Desired scaling factor in each direction. If scalar, the value is used
             for both the x- and y-direction.
@@ -950,6 +1015,10 @@ class Rescale(nn.Module):
 
 class Rotate(nn.Module):
     r"""Rotate the tensor anti-clockwise about the centre.
+
+    Convention:
+        - align_corners: ``True`` by default, matching :func:`rotate`
+        - See the convention block of :func:`rotate`.
 
     Args:
         angle: The angle through which to rotate. The tensor
@@ -1009,6 +1078,10 @@ class Rotate(nn.Module):
 class Translate(nn.Module):
     r"""Translate the tensor in pixel units.
 
+    Convention:
+        - align_corners: ``True`` by default, matching :func:`translate`
+        - See the convention block of :func:`translate`.
+
     Args:
         translation: tensor containing the amount of pixels to
           translate in the x and y direction. The tensor must have a shape of
@@ -1061,6 +1134,10 @@ class Translate(nn.Module):
 
 class Scale(nn.Module):
     r"""Scale the tensor by a factor.
+
+    Convention:
+        - align_corners: ``True`` by default, matching :func:`scale`
+        - See the convention block of :func:`scale`.
 
     Args:
         scale_factor: The scale factor apply. The tensor
@@ -1120,6 +1197,11 @@ class Scale(nn.Module):
 
 class Shear(nn.Module):
     r"""Shear the tensor.
+
+    Convention:
+        - align_corners: ``True`` by default (differs from function :func:`shear`, whose default is
+          ``False``)
+        - See the convention block of :func:`shear`.
 
     Args:
         shear: tensor containing the angle to shear
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -151,11 +151,11 @@ def affine(
 
     Args:
         tensor: The image tensor to be warped in shapes of
-            :math:`(H, W)`, :math:`(D, H, W)` and :math:`(B, C, H, W)`.
+            :math:`(C, H, W)` or :math:`(B, C, H, W)`.
         matrix: The 2x3 affine transformation matrix.
         mode: interpolation mode to calculate output values ``'bilinear'`` | ``'nearest'``.
         padding_mode: padding mode for outside grid values
-          ``'torch.zeros'`` | ``'border'`` | ``'reflection'``.
+          ``'zeros'`` | ``'border'`` | ``'reflection'``.
         align_corners: interpolation flag.
 
     Returns:
@@ -209,12 +209,12 @@ def affine3d(
 
     Args:
         tensor: The image tensor to be warped in shapes of
-            :math:`(D, H, W)`, :math:`(C, D, H, W)` and :math:`(B, C, D, H, W)`.
+            :math:`(C, D, H, W)` or :math:`(B, C, D, H, W)`.
         matrix: The affine transformation matrix with shape :math:`(B, 3, 4)`.
         mode: interpolation mode to calculate output values
           ``'bilinear'`` | ``'nearest'``.
         padding_mode: padding mode for outside grid values
-         `` 'torch.zeros'`` | ``'border'`` | ``'reflection'``.
+         `` 'zeros'`` | ``'border'`` | ``'reflection'``.
         align_corners: interpolation flag.
 
     Returns:
@@ -282,7 +282,7 @@ def rotate(
         mode: interpolation mode to calculate output values
           ``'bilinear'`` | ``'nearest'``.
         padding_mode: padding mode for outside grid values
-          ``'torch.zeros'`` | ``'border'`` | ``'reflection'``.
+          ``'zeros'`` | ``'border'`` | ``'reflection'``.
         align_corners: interpolation flag.
 
     Returns:
@@ -352,12 +352,12 @@ def rotate3d(
         roll: The roll angle through which to rotate. The tensor
           must have a shape of (B), where B is batch size.
         center: The center through which to rotate. The tensor
-          must have a shape of (B, 2), where B is batch size and last
-          dimension contains cx and cy.
+          must have a shape of (B, 3), where B is batch size and last
+          dimension contains x, y and z.
         mode: interpolation mode to calculate output values
           ``'bilinear'`` | ``'nearest'``.
         padding_mode: padding mode for outside grid values
-          ``'torch.zeros'`` | ``'border'`` | ``'reflection'``.
+          ``'zeros'`` | ``'border'`` | ``'reflection'``.
         align_corners: interpolation flag.
 
     Returns:
@@ -422,7 +422,7 @@ def translate(
         mode: interpolation mode to calculate output values
           ``'bilinear'`` | ``'nearest'``.
         padding_mode: padding mode for outside grid values
-          ``'torch.zeros'`` | ``'border'`` | ``'reflection'``.
+          ``'zeros'`` | ``'border'`` | ``'reflection'``.
         align_corners: interpolation flag.
 
     Returns:
@@ -481,7 +481,7 @@ def scale(
         mode: interpolation mode to calculate output values
           ``'bilinear'`` | ``'nearest'``.
         padding_mode: padding mode for outside grid values
-          ``'torch.zeros'`` | ``'border'`` | ``'reflection'``.
+          ``'zeros'`` | ``'border'`` | ``'reflection'``.
         align_corners: interpolation flag.
 
     Returns:
@@ -543,7 +543,7 @@ def shear(
         mode: interpolation mode to calculate output values
           ``'bilinear'`` | ``'nearest'``.
         padding_mode: padding mode for outside grid values
-          ``'torch.zeros'`` | ``'border'`` | ``'reflection'``.
+          ``'zeros'`` | ``'border'`` | ``'reflection'``.
         align_corners: interpolation flag.
 
     Returns:
@@ -738,8 +738,6 @@ def rescale(
         interpolation:  algorithm used for upsampling: ``'nearest'`` | ``'linear'`` | ``'bilinear'`` |
             ``'bicubic'`` | ``'trilinear'`` | ``'area'``.
         align_corners: interpolation flag.
-        side: Corresponding side if ``size`` is an integer. Can be one of ``'short'``, ``'long'``, ``'vert'``,
-            or ``'horz'``.
         antialias: if True, then image will be filtered with Gaussian before downscaling.
             No effect for upscaling.
 
@@ -860,7 +858,7 @@ class Affine(nn.Module):
         mode: interpolation mode to calculate output values
           ``'bilinear'`` | ``'nearest'``.
         padding_mode: padding mode for outside grid values
-          ``'torch.zeros'`` | ``'border'`` | ``'reflection'``.
+          ``'zeros'`` | ``'border'`` | ``'reflection'``.
         align_corners: interpolation flag.
 
     Raises:
@@ -968,8 +966,6 @@ class Rescale(nn.Module):
         interpolation:  algorithm used for upsampling: ``'nearest'`` | ``'linear'`` | ``'bilinear'`` |
             ``'bicubic'`` | ``'trilinear'`` | ``'area'``.
         align_corners: interpolation flag.
-        side: Corresponding side if ``size`` is an integer. Can be one of ``'short'``, ``'long'``, ``'vert'``,
-            or ``'horz'``.
         antialias: if True, then image will be filtered with Gaussian before downscaling.
             No effect for upscaling.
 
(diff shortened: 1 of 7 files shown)


tests/geometry/transform/test_affine.py
```diff
@@ -315,6 +346,21 @@ def test_angle90_batch2_broadcast(self, device, dtype):
         transform = kornia.geometry.transform.Rotate(angle, align_corners=True)
         self.assert_close(transform(inp), expected, atol=1e-4, rtol=1e-4)
 
+    def test_convention_align_corners_default_true(self, device, dtype):
+        # rotate/Rotate's bare default align_corners=True (every other test in this class passes
+        # align_corners=True explicitly to Rotate; this exercises the bare default).
+        if dtype in (torch.float16, torch.bfloat16):
```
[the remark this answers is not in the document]
ducha-aiki: Confirmed and fixed in a5220535: the test now also asserts kornia.geometry.transform.rotate(inp, angle) bare-default against the same literal, so class and function defaults are pinned independently — warranted in this module, where Rescale/rescale and Shear/shear demonstrably diverge.

kornia/geometry/transform/imgwarp.py
```diff
@@ -331,6 +350,10 @@ def warp_grid(grid: torch.Tensor, src_homo_dst: torch.Tensor) -> torch.Tensor:
 def warp_grid3d(grid: torch.Tensor, src_homo_dst: torch.Tensor) -> torch.Tensor:
     r"""Compute the grid to warp the coordinates grid by the homography/ies.
 
+    Convention:
+        - ``grid`` coordinates: ``(x, y, z)`` (last dim), shape :math:`(1, D, H, W, 3)`
+        - ``src_homo_dst`` is the destination→source homography :math:`(1, 4, 4)` or :math:`(N, 1, 4, 4)`
+
```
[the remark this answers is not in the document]
ducha-aiki: Confirmed by execution ((1,2,3,4,3) out for a (1,2,3,4,3) grid) and fixed in a5220535: Returns now states (N, D, H, W, 3). Same wrong-existing-line class as the other corrections in this PR.

tests/geometry/transform/test_imgwarp.py
```diff
@@ -395,6 +413,25 @@ def test_translation_normalized(self, device, dtype):
...
+        # homography_warp's default normalized_homography=True path (normalize the pixel homography,
+        # invert it, then warp with normalized coordinates) is equivalent to warp_perspective's raw
+        # pixel-homography path, given matching align_corners.
+        # Snippet used to generate expected:
+        # x = torch.arange(9.0).view(1, 1, 3, 3)
+        # H = torch.tensor([[[1.0, 0.0, 1.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]]])
+        # expected = kornia.geometry.transform.warp_perspective(x, H, (3, 3), align_corners=True)
+        x = torch.arange(9.0, device=device, dtype=dtype).view(1, 1, 3, 3)
+        H = torch.tensor([[[1.0, 0.0, 1.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]]], device=device, dtype=dtype)
+        expected = torch.tensor([[[[0.0, 0.0, 1.0], [0.0, 3.0, 4.0], [0.0, 6.0, 7.0]]]], device=device, dtype=dtype)
```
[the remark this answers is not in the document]
ducha-aiki: Checked with awk 'length > 120' over tests/geometry/transform/test_imgwarp.py: no line exceeds 120 (the cited lines are 117 chars after ruff format). The pinned ruff and pre-commit.ci both pass on the branch.

kornia/geometry/transform/imgwarp.py
```diff
@@ -1379,6 +1472,17 @@ def homography_warp(
 
     See :class:`~kornia.geometry.warp.HomographyWarper` for details.
 
+    Convention:
+        - input: :math:`(N, C, H, W)`
+        - ``src_homo_dst`` is the destination→source homography :math:`(N, 3, 3)`
+          (contrast :func:`warp_perspective`, which consumes source→destination pixel)
```
ducha-aiki: **The direction bullet only holds on the default `normalized_homography=True` path — with `normalized_homography=False` the direction flips.**

That branch passes the matrix straight to `warp_perspective` (imgwarp.py:1528-1529), which consumes it as the **source→destination** pixel homography. Two in-repo sources pin this: `test_translation_normalized` feeds the *same* forward `homo_ab` to `warp_perspective` and to `homography_warp(..., normalized_homography=False)` and expects identical output, and `docs/source/get-started/conventions.rst` (pitfall 9, lines 218-219) says it wants “destination→source, normalized (**or pass `normalized_homography=False`**)”. As written, the unconditional bullet — in the very block meant to pin conventions — tells a user with a dst→src pixel homography to call the pixel path directly, which produces a silently inverted warp, and it now contradicts the conventions page.

The next bullet already carves out the pixel path for `align_corners`; this one needs the same qualifier:

```suggestion
        - ``src_homo_dst`` is the destination→source homography :math:`(N, 3, 3)` when
          ``normalized_homography=True`` (default); with ``normalized_homography=False`` it is consumed
          as the source→destination **pixel** homography, exactly like :func:`warp_perspective`
```

Also worth cross-linking these per-op blocks to `docs/source/get-started/conventions.rst` (and vice versa) so the two statements of the same convention cannot drift — this bullet is already the first divergence.

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
```diff
@@ -1379,6 +1472,17 @@ def homography_warp(
...
+        - ``normalized_homography=True`` by default: ``src_homo_dst`` is in
+          normalized :math:`[-1, 1]` coordinates
+        - ``dsize`` is ``(h, w)``
+        - align_corners: ``False`` by default (only honored when ``normalized_homography=True``;
+          the pixel-homography path currently forces ``True``)
```
ducha-aiki: The `align_corners` half of the pixel-path quirk is documented, but the same branch also hardcodes `mode="bilinear"` (imgwarp.py:1528-1529), silently ignoring the caller's `mode` (and `normalized_coordinates` is ignored there too). Concrete trap: `homography_warp(mask, H_pix, dsize, mode="nearest", normalized_homography=False)` for a label map silently blends class ids. Since this block names one silently-ignored argument on that path, it should name the other as well — and the #3923 follow-up should cover `mode`, not just `align_corners`.

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/homography_warper.py
```diff
@@ -76,6 +81,10 @@ class HomographyWarper(BaseWarper):
 
         X_{dst} = H_{src}^{\{dst\}} * X_{src}
 
+    Convention:
```
ducha-aiki, line 84: The new block pins “destination→source” — correct, since `forward` delegates to `homography_warp`'s normalized path — but two lines above, the pre-existing class equation reads `X_{dst} = H_{src}^{\{dst\}} * X_{src}`, i.e. a source→destination mapping of points. This PR fixes four other wrong pre-existing docstring lines; this one now sits directly above the Convention block contradicting it on the same rendered docs page (and `BaseWarper` renders adjacent in the same rst section). Suggest correcting the equation in the same pass, e.g. `X_{src} = H_{src}^{\{dst\}} * X_{dst}`, or dropping it in favor of the Convention bullet.

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
```diff
@@ -1453,6 +1557,12 @@ def homography_warp3d(
 ) -> torch.Tensor:
     r"""Warp image patches or tensors by normalized 3D homographies.
 
+    Convention:
+        - input: :math:`(N, C, D, H, W)`; ``dsize`` is ``(d, h, w)``
+        - ``src_homo_dst`` is the destination→source homography :math:`(N, 4, 4)`, normalized coordinates
```
ducha-aiki: `normalized_coordinates: bool = True` is a real parameter of `homography_warp3d`, so “normalized coordinates” is only the default, not an invariant — same qualifier style as the 2D block:

```suggestion
        - ``src_homo_dst`` is the destination→source homography :math:`(N, 4, 4)`, in normalized
          :math:`[-1, 1]` coordinates by default (``normalized_coordinates=True``)
```

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
```diff
@@ -1379,6 +1472,17 @@ def homography_warp(
 
     See :class:`~kornia.geometry.warp.HomographyWarper` for details.
```
ducha-aiki: While pinning this docstring: the xref one line above the new block is dead — there is no `kornia.geometry.warp` module (the class lives in `kornia.geometry.transform`), and since the docs build isn't nitpicky it renders silently as a broken link:

```suggestion
    See :class:`~kornia.geometry.transform.HomographyWarper` for details.
```

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/affwarp.py
```diff
@@ -321,6 +337,11 @@ def rotate3d(
 ) -> torch.Tensor:
     r"""Rotate 3D the tensor anti-clockwise about the centre.
 
+    Convention:
+        - ``center`` is ``(x, y, z)`` in pixels, origin at top-left; defaults to the tensor center
```
ducha-aiki: For a `(B, C, D, H, W)` volume, “origin at top-left” leaves the z axis unanchored — a reader placing an explicit rotation center has no stated convention for where `z = 0` lives (front vs back slice). `_compute_tensor_center3d` measures z from slice 0, so:

```suggestion
        - ``center`` is ``(x, y, z)`` in pixels, origin at the top-left of the first depth
          slice (``z = 0``); defaults to the tensor center
```

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
ducha-aiki: Verified against imgwarp.py's pixel branch (straight pass-through to warp_perspective), test_translation_normalized, and pitfall 9 — the direction genuinely flips with the flag, and the unconditional bullet was wrong exactly where precision matters most. Fixed in 01cd0868 with your suggested wording (merged with the normalized-coordinates clause into one path-qualified bullet). The audit derived the direction from the parameter name and the True path only — the flag-dependent semantics slipped through; noted for the follow-up batches' audit checklist.

kornia/geometry/transform/imgwarp.py
ducha-aiki: Confirmed — the same return statement hardcodes mode='bilinear'. Fixed in 01cd0868: the block now names both forced arguments on the pixel path, and I've noted mode alongside align_corners in #3923's follow-up scope.

kornia/geometry/transform/homography_warper.py
ducha-aiki: Fixed in 01cd0868: equation corrected to X_src = H * X_dst, the dst→src sampling form the Convention block and forward's delegation actually implement.

kornia/geometry/transform/imgwarp.py
ducha-aiki: Fixed in 01cd0868 with the suggested qualifier — normalized_coordinates=True is a default, not an invariant.

kornia/geometry/transform/imgwarp.py
ducha-aiki: Fixed in 01cd0868 — xref now points at kornia.geometry.transform.HomographyWarper, which also renders now that the rst entries were added earlier in this PR.

kornia/geometry/transform/affwarp.py
ducha-aiki: Fixed in 01cd0868 — applied to all five 3D center/point lines (rotate3d, get_shear_matrix3d ×2, get_affine_matrix3d, get_perspective_transform3d points), anchored at the first depth slice (z = 0) per _compute_tensor_center3d.

ducha-aiki:
Addressed the request-changes review in 0adde945 (all six findings verified real before fixing):

- **warp_affine / warp_perspective3d equations**: both sampled with `M` directly while the new blocks (correctly) call `M` source→destination — warp_affine now uses the `M^{-1}` component form like warp_perspective, warp_perspective3d gets a truthful homogeneous 3D form and its Returns ties output sizes to `dsize`.
- **warp_grid / warp_grid3d**: `(N, 3, 3)` / `(N, 4, 4)` batched homographies confirmed working by execution — Convention and Args now include them.
- **homography_warp3d 2D leftovers**: `dsize` Args text corrected to depth/height/width and the copied 2D example replaced with a runnable 3D one (doctest count unchanged at 25, all passing).
- **rst gaps**: `affine3d`, `rotate3d`, `resize_to_be_divisible` added; every audited symbol now renders.
- **resize test**: confirmed the output literal cannot distinguish `None` from `False` (verified: outputs `torch.equal`) — the test now also pins the signature default `is None`. Checked the other two default-pinning tests for the same hole: `rotate`'s literal genuinely differs between `True`/`False` and `HomographyWarper`'s is padding-sensitive, so output pins suffice there.

The earlier open threads (pixel-path direction flip, forced `mode`, warper equation, z-origin on all five 3D lines) were fixed in 01cd0868. Gates re-run: ruff/pre-commit clean, 25 doctests, 411 transform tests, sphinx -W exit 0.

---
🤖 Response authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
```diff
@@ -86,6 +86,14 @@ def warp_perspective(
         \frac{M^{-1}_{21} x + M^{-1}_{22} y + M^{-1}_{23}}{M^{-1}_{31} x + M^{-1}_{32} y + M^{-1}_{33}}
         \right )
 
+    Convention:
+        - input: :math:`(B, C, H, W)`; ``dsize`` is ``(h, w)``
```
ducha-aiki, line 90: Follow-up in the same class as the `warp_perspective3d` `Returns:` fix (now line 1440): three sibling warps still carry the stale claim that the output spatial size equals the input's, when it is given by `dsize` (verified: `warp_perspective(rand(1,1,6,7), eye, (3,5))` → `(1,1,3,5)`, and likewise for the others):

- `warp_perspective` line 108: “the warped input image :math:`(B, C, H, W)`” → should be `(B, C, h, w)`, spatial sizes given by ``dsize``
- `warp_affine` line 215: “the warped torch.Tensor with shape :math:`(B, C, H, W)`” → same
- `warp_affine3d` line 1066: “the warped 3d torch.tensor with shape :math:`(B, C, D, H, W)`” → `(B, C, d, h, w)`, given by ``dsize``

Each of these functions now has a Convention block stating ``dsize`` explicitly, so the stale `Returns:` line sits a few lines below a bullet it contradicts whenever `dsize != (H, W)`.

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
```diff
@@ -1377,7 +1472,19 @@ def homography_warp(
 ) -> torch.Tensor:
     r"""Warp image patches or tensors by normalized 2D homographies.
 
-    See :class:`~kornia.geometry.warp.HomographyWarper` for details.
+    See :class:`~kornia.geometry.transform.HomographyWarper` for details.
+
+    Convention:
+        - input: :math:`(N, C, H, W)`
+        - ``src_homo_dst`` is the destination→source homography :math:`(N, 3, 3)`, in normalized
+          :math:`[-1, 1]` coordinates, when ``normalized_homography=True`` (default); with
```
ducha-aiki: Same class as the qualifier you added to the `homography_warp3d` block: `homography_warp` also has `normalized_coordinates: bool = True` (its Args entry: “Whether the homography assumes [-1, 1] normalized coordinates or not”), so on the `normalized_homography=True` path the “in normalized [-1, 1] coordinates” claim additionally depends on that flag — with `normalized_coordinates=False` the warp grid is built in pixel coordinates and the homography consumes destination *pixel* coords instead:

```suggestion
        - ``src_homo_dst`` is the destination→source homography :math:`(N, 3, 3)`, in normalized
          :math:`[-1, 1]` coordinates by default (``normalized_coordinates=True``), when
          ``normalized_homography=True`` (default); with
```

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

ducha-aiki:
Addressed the re-review in 762b9321 + d5af0aa7 — all seven findings verified real before fixing (execution checks where applicable):

- **[P1] homography_warp flag contract**: the Convention bullet now qualifies the normalized claim on `normalized_coordinates=True` (762b9321), and the Args entry states the direction per path — dst→src when `normalized_homography=True`, src→dst pixel when `False` (d5af0aa7). This op's docstring is now fully flag-conditional.
- **[P2] HomographyWarper test**: confirmed False/None outputs are byte-identical — signature-default asserts added for both `HomographyWarper.__init__` and `homography_warp`, per the resize pattern.
- **[P2] equations**: `warp_perspective3d` now dehomogenizes via \(\pi(M^{-1}\bar{x})\); `warp_affine` defines \(M^{-1}\) as the inverse of the 3×3 homogeneous extension.
- **[P2] invalid shapes**: verified `affine((H,W))` and `affine3d((D,H,W))` raise — Args now list only `(C,H,W)`/`(B,C,H,W)` and `(C,D,H,W)`/`(B,C,D,H,W)`; `rotate3d`'s center corrected to `(B, 3)` x,y,z.
- **[P2] Returns/dsize trio**: fixed in 762b9321 (crossed with the review); a sweep of both files found no further same-class staleness (remaining matches: private helper, Args input lines, dsize-less `Affine`).
- **[P2] `'torch.zeros'`**: all 16 occurrences (4 imgwarp + 12 affwarp) corrected to `'zeros'`.
- **[P2] phantom `side`**: removed from `rescale` and `Rescale` (verified absent from their signatures; kept for `resize`/`resize_to_be_divisible`, which do accept it).
- **Lower priority**: `affine3d`/`rotate3d` moved to a new "Image 3d transforms" rst section; `warp_grid`/`warp_grid3d` now document already-batched `(N,H,W,2)`/`(N,D,H,W,3)` grids (verified working).

Gates re-run: ruff/pre-commit clean, 25 doctests, 411 transform tests, sphinx -W exit 0.

---
🤖 Response authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
```diff
@@ -86,18 +86,26 @@ def warp_perspective(
         \frac{M^{-1}_{21} x + M^{-1}_{22} y + M^{-1}_{23}}{M^{-1}_{31} x + M^{-1}_{32} y + M^{-1}_{33}}
         \right )
 
+    Convention:
+        - input: :math:`(B, C, H, W)`; ``dsize`` is ``(h, w)``
+        - ``M`` is the source→destination **pixel** homography :math:`(B, 3, 3)`
+          (contrast :func:`homography_warp`, which consumes destination→source normalized)
```
ducha-aiki: Tiny leftover of the qualifier class fixed in the main `homography_warp` block: this aside states unconditionally that `homography_warp` “consumes destination→source normalized”, but that is its default-path behavior only — with `normalized_homography=False` it consumes exactly the source→destination pixel matrix this function consumes. One clause aligns it with the fully-qualified statement in `homography_warp`'s own block:

```suggestion
          (contrast :func:`homography_warp`, which by default consumes destination→source normalized)
```

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
```diff
@@ -457,6 +485,11 @@ def get_perspective_transform(points_src: torch.Tensor, points_dst: torch.Tensor
         1 \\
         \end{bmatrix}
 
+    Convention:
+        - points: ``(x, y)``, pixel centers, origin at top-left; shape :math:`(B, 4, 2)`
+        - returns the source→destination **pixel** homography :math:`(B, 3, 3)`
+          (contrast :func:`homography_warp`, which consumes destination→source normalized)
```
ducha-aiki: Tiny leftover of the qualifier class fixed in the main `homography_warp` block: this aside states unconditionally that `homography_warp` “consumes destination→source normalized”, but that is its default-path behavior only — with `normalized_homography=False` it consumes exactly the source→destination pixel matrix this function returns. One clause aligns it with the fully-qualified statement in `homography_warp`'s own block:

```suggestion
          (contrast :func:`homography_warp`, which by default consumes destination→source normalized)
```

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
```diff
@@ -879,6 +947,11 @@ def get_shear_matrix3d(
         s = S_{xy}S_{zx} + (S_{xy}S_{yx} + 1)S_{zy}
         t = S_{xz}S_{zx} + (S_{xz}S_{yx} + S_{yz})S_{zy} + 1
 
+    Convention:
+        - ``center`` is ``(x, y, z)`` in pixels, origin at the top-left of the first depth
+          slice (``z = 0``)
+        - returns :math:`(B, 4, 4)` affine matrix in pixel coordinates
+
     Params:
```
ducha-aiki: Adjacent to the new block: this docstring's parameter section is headed `Params:`, which napoleon does not recognize (only `Args:`/`Parameters:`/`Arguments:`), so in the rendered docs `get_shear_matrix3d`'s parameters fall out of the field-list styling that every sibling function gets. Since this PR already corrects wrong pre-existing lines in touched docstrings, worth the drive-by:

```suggestion
    Args:
```

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
ducha-aiki: Fixed in 8288df5c with the suggested clause (both occurrences — this one and the twin in get_perspective_transform's block).

kornia/geometry/transform/imgwarp.py
ducha-aiki: Fixed in 8288df5c together with its twin at warp_perspective.

kornia/geometry/transform/imgwarp.py
ducha-aiki: Fixed in 8288df5c — Params: → Args:; confirmed napoleon now renders get_shear_matrix3d's parameters as a field list in the local build.

ducha-aiki:
Fourth review wave addressed in 8288df5c (each verified before fixing):

- **rescale factor order**: confirmed by execution — `rescale(x, (2.0, 1.0))` on a 4×8 input yields 8×8, so `factor` is `(factor_h, factor_w)`; also confirmed `scale`'s `scale_factor` is x-first with a marker test. The block states the h-first order with the explicit contrast, and both Args (function + class) drop the misleading "x- and y-direction" wording.
- **warp_grid/warp_grid3d**: Args now match the blocks on batched `(N, …)` grids.
- **'torch.zeros' in homography_warper.py**: fixed — my earlier sweep only covered imgwarp/affwarp. `elastic_transform.py` and `crop2d.py` carry the same stale value but are untouched by this PR; queued for the crops/flips batch rather than expanding this diff.
- **invert_affine_transform**: block now states it's pure matrix inversion in whatever convention the input uses — no pixel-space claim.
- **get_affine_matrix3d**: note now points at `warp_perspective3d` (it returns `(B, 4, 4)`; the 2D pointer was the only 3D-function/2D-target mismatch in the file — checked all 15 "conjunction with" notes).

Gates: ruff/pre-commit clean, 25 doctests, 411 transform tests, sphinx -W exit 0.

---
🤖 Response authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
```diff
@@ -306,11 +323,16 @@ def _fill_and_warp(
 def warp_grid(grid: torch.Tensor, src_homo_dst: torch.Tensor) -> torch.Tensor:
     r"""Compute the grid to warp the coordinates grid by the homography/ies.
 
+    Convention:
+        - ``grid`` coordinates: ``(x, y)`` (last dim), shape :math:`(1, H, W, 2)` or :math:`(N, H, W, 2)`
```
ducha-aiki, line 327: The batched-grid and homography shapes read as freely combinable, but one documented combination raises: the batch size is taken from `src_homo_dst` and the grid is `expand()`-ed to it, so a batched `(N, H, W, 2)` grid with a single `(1, 3, 3)` homography fails — `expand` cannot shrink `N` to 1.

Reproduced: `warp_grid(torch.rand(2, 4, 5, 2), torch.eye(3)[None])` → `RuntimeError: The expanded size of the tensor (1) must match the existing size (2)...`. Verified working: `(1, H, W, 2)` grid with any homography batch, and `(N, H, W, 2)` with a matching batch of `N` homographies.

Worth one clause here and in the Args entry: a batched grid requires a matching batch of `N` homographies; only the `(1, H, W, 2)` form broadcasts.

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
```diff
@@ -331,14 +353,20 @@ def warp_grid(grid: torch.Tensor, src_homo_dst: torch.Tensor) -> torch.Tensor:
...
 
+    Convention:
+        - ``grid`` coordinates: ``(x, y, z)`` (last dim), shape :math:`(1, D, H, W, 3)` or
+          :math:`(N, D, H, W, 3)`
+        - ``src_homo_dst`` is the destination→source homography :math:`(1, 4, 4)`,
+          :math:`(N, 4, 4)` or :math:`(N, 1, 4, 4)`
+
     Args:
-        grid: Unwrapped grid of the shape :math:`(1, D, H, W, 3)`.
+        grid: Unwrapped grid of the shape :math:`(1, D, H, W, 3)` or :math:`(N, D, H, W, 3)`.
```
ducha-aiki: Same combination trap as `warp_grid` above: `warp_grid3d(torch.rand(2, 3, 4, 5, 3), torch.eye(4)[None])` raises the same `expand` RuntimeError — the documented `(N, D, H, W, 3)` grid works only with a matching batch of `N` homographies, not with the `(1, 4, 4)` form listed alongside it. Same one-clause fix as suggested there.

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

tests/geometry/transform/test_imgwarp.py
```diff
@@ -395,6 +413,25 @@ def test_translation_normalized(self, device, dtype):
         img_a = kornia.geometry.transform.homography_warp(img_b, homo_ab, (h, w), normalized_homography=False)
         self.assert_close(img_a, expected, atol=1e-4, rtol=1e-4)
 
+    def test_convention_normalize_invert_equivalence(self, device, dtype):
```
ducha-aiki, line 416: The square, same-size case is degenerate for the convention this test exists to pin: with input size == dsize, normalize-then-invert and invert-then-normalize (sizes unswapped) produce byte-identical homographies, so the *ordering* half of the equivalence — the exact pitfall the conventions page warns about — is invisible here.

Verified numerically: with `x = torch.arange(24.).view(1, 1, 4, 6)`, `dsize = (3, 5)`, and the same translation `H`, the correct order (`normalize_homography(H, (4, 6), (3, 5))` then invert) matches `warp_perspective(x, H, (3, 5), align_corners=True)` exactly (max err 0.0), while the wrong order (invert first, sizes unswapped) diverges by 15.0. In the current `(3, 3)`→`(3, 3)` setup the wrong order also yields max err 0.0 — a regression in `normalize_homography`'s `dsize_src`/`dsize_dst` handling would pass this test unchanged.

Suggest regenerating the literal with an asymmetric case (e.g. 4×6 input, `dsize=(3, 5)`) so the pinned equivalence actually discriminates the ordering.

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
```diff
@@ -1377,12 +1479,26 @@ def homography_warp(
...
+        - align_corners: ``False`` by default; ``mode``: ``'bilinear'`` by default (both only
+          honored when ``normalized_homography=True`` — the pixel-homography path currently
+          forces ``align_corners=True`` and ``mode='bilinear'``)
+        - padding_mode: ``'zeros'`` by default
 
     Args:
         patch_src: The image or torch.Tensor to warp. Should be from source of shape :math:`(N, C, H, W)`.
-        src_homo_dst: The homography or torch.stack of homographies from destination to source of shape
-            :math:`(N, 3, 3)`.
+        src_homo_dst: The homography or torch.stack of homographies of shape :math:`(N, 3, 3)` —
```
ducha-aiki, line 1505: Minor, same drive-by class as the `Params:` → `Args:` fix: a few lines below this entry (line 1509), the pre-existing Args line for the flag this whole docstring now pivots on is garbled — `normalized_homography: show is homography normalized.` Suggest e.g.:

`normalized_homography: whether ``src_homo_dst`` is a normalized (destination→source) homography (``True``, default) or a pixel source→destination homography (``False``).`

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/affwarp.py
```diff
@@ -896,14 +957,17 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
 class Rescale(nn.Module):
     r"""Rescale the input torch.Tensor with the given factor.
 
+    Convention:
+        - align_corners: ``True`` by default (differs from function :func:`rescale`, whose default
```
ducha-aiki, line 961: Non-blocking, coverage note: the two class-vs-function divergences this PR documents — `Rescale` (`True`) vs `rescale` (`None`) here, and `Shear` (`True`) vs `shear` (`False`) at line 1201 — are the most surprising claims in the whole diff, yet neither is pinned by a test, while the *matching* defaults (`rotate`, `resize`, `HomographyWarper`) are. The natural future "cleanup" (unifying a class default with its function) would pass the entire suite while inverting these bullets.

The `inspect.signature` pinning pattern the tests already use covers this in ~4 lines per pair (or one small table-driven test over all documented defaults). Fine as a follow-up batch item if you prefer to keep this PR frozen.

---
🤖 Review authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

kornia/geometry/transform/imgwarp.py
ducha-aiki: Confirmed (reproduced the expand RuntimeError) and fixed in fee10afa: block and Args now state that only the (1, H, W, 2) grid broadcasts; a batched grid requires a matching batch of N homographies.

kornia/geometry/transform/imgwarp.py
ducha-aiki: Fixed in fee10afa together with the 2D twin — same clause in block and Args.

tests/geometry/transform/test_imgwarp.py
ducha-aiki: Confirmed — with equal square sizes normalize and invert commute, so the ordering half of the pin was invisible; reproduced your 0.0-vs-15.0 discrimination. Fixed in fee10afa: literal regenerated at 4×6 → (3, 5) with the degeneracy rationale in the test comment. Good catch — this joins the degenerate-pin class (resize None==False) on the follow-up batches' checklist.

kornia/geometry/transform/imgwarp.py
ducha-aiki: Fixed in fee10afa with your suggested wording — the garbled line was indeed the worst possible place for it, on the flag the whole docstring now pivots on.

ducha-aiki:
Fifth wave addressed in fee10afa — colleague's two leftovers plus the four newest inline comments, each verified first:

- **rescale scalar wording**: confirmed `rescale(x, 2)` raises TypeError while `2.0` works — Args now say "single float" and note the int behavior (both function and class).
- **precompute_warp_grid**: confirmed `(N, 3, 3)` works — added to its Args.
- **warp_grid/warp_grid3d broadcasting contract**: reproduced the expand RuntimeError — batched grids need a matching homography batch; only the batch-1 grid broadcasts. Stated in blocks and Args.
- **Equivalence test degeneracy**: reproduced the 0.0-vs-15.0 ordering discrimination — literal regenerated at 4×6 → (3, 5).
- **Garbled `normalized_homography` Args line**: rewritten per suggestion.

Gates: ruff clean, 25 doctests, 789 transform tests (f32+f64), sphinx -W exit 0.

---
🤖 Response authored by Claude (Fable 5) via Claude Code, posted on behalf of Dmytro Mishkin.

PR #3931, opened 2026-08-09
feat(metrics): add KITTI D1 disparity error

### LINKED ISSUE #3758: [Feature]: KITTI D1 disparity error metric (follow-up to #1123)
### 🚀 Feature Description

Add `kitti_d1_error` to `kornia.metrics`, completing the stereo disparity metrics started in #1123. D1 is the KITTI benchmark's primary stereo metric: the fraction of pixels whose disparity error exceeds **both** an absolute threshold (3 px) **and** a relative threshold (5%). It builds directly on the `mean_bad_pixel_error` / shared validation+reduction helpers already merged.

### 📂 Feature Category

Geometry

### 💡 Motivation

#1123 added MAE, RMSE, and bad-pixel error but intentionally deferred KITTI D1 to keep that PR focused. D1 (a.k.a. D1-all) is the headline ranking metric on the KITTI 2015 stereo benchmark, so stereo researchers comparing against published numbers need it. Without it they re-implement it per-project.

### 💭 Proposed Solution

New functional metric in `kornia/metrics/disparity.py`, reusing the existing `_check_disparity_inputs` / `_reduce_disparity_error` helpers and matching the signatures of the merged metrics:

```python
def kitti_d1_error(
    input: torch.Tensor,
    target: torch.Tensor,
    abs_threshold: float = 3.0,
    rel_threshold: float = 0.05,
    valid_mask: torch.Tensor | None = None,
    reduction: str = "mean",
) -> torch.Tensor:
    # bad if |d - d_gt| > abs_threshold AND |d - d_gt| / |d_gt| > rel_threshold
```
Exported from `kornia/metrics/__init__.py`, documented in `docs/source/metrics.rst` (Stereo section), and tested in `tests/metrics/test_disparity.py` with hardcoded expected values.

**Secondary (optional,, would like to ask a maintainer):** `nn.Module` wrappers for the four disparity metrics, mirroring the existing `AEPE`/`aepe` precedent. Happy to include or split into a separate PR.

Per recent reviewer feedback, the new function(s) will be verified `torch.compile`-compatible (dynamo test).

### 🔄 Alternatives Considered

Leaving D1 to downstream users (status quo, causes re-implementation); or folding a `relative` flag into `mean_bad_pixel_error` (rejected: D1's dual abs-AND-relative criterion is a distinct, named metric and clearer as its own function).

### 🎯 Use Cases

Reporting/validating stereo-matching models against KITTI 2015 leaderboard numbers; training-loop monitoring of disparity quality.

## 📝 Additional Context

Reference: KITTI 2015 stereo benchmark, Menze & Geiger, *Object Scene Flow for Autonomous Vehicles*, CVPR 2015 - https://www.cvlibs.net/datasets/kitti/eval_scene_flow.php?benchmark=stereo. Follow-up to #1123.

### 🤝 Contribution Intent

VanniLeonardo:
## 📝 Description

Closes #3758

Adds `kornia.metrics.kitti_d1_error`, the outlier ratio reported by the KITTI 2015 stereo benchmark.

A pixel counts as an outlier only when **both** criteria hold: its absolute disparity error exceeds
`abs_threshold` (default `3.0`) **and** its relative disparity error exceeds `rel_threshold`
(default `0.05`):

```
D1(D, D_gt) = mean over valid p of [ |D_p - D_gt_p| > τ_abs  ∧  |D_p - D_gt_p| / |D_gt_p| > τ_rel ]
```

Unlike the existing `mean_bad_pixel_error`, the additional relative criterion keeps large
disparities from being penalised for errors that are small relative to their magnitude. The result
is expressed as a fraction in `[0, 1]` rather than a percentage.

**Reference:** M. Menze and A. Geiger. *Object Scene Flow for Autonomous Vehicles.* CVPR 2015 —
https://www.cvlibs.net/datasets/kitti/eval_scene_flow.php?benchmark=stereo

### Why a change to `_reduce_disparity_error` ships alongside the new metric

This PR also makes the shared helper `_reduce_disparity_error` `torch.compile`-safe.

The helper previously reduced with `error[valid_mask]`, which lowers to `aten.nonzero`. That
produces a **data-dependent output shape**, which dynamo cannot trace in a full graph, so
`kitti_d1_error` could not have been compile-clean while calling it. The masked path now uses
`torch.where` to zero invalid pixels and divides by `valid_mask.sum()`:

```python
if valid_mask is not None:
    error = torch.where(valid_mask, error, torch.zeros_like(error))

if reduction == "mean":
    error = error.mean() if valid_mask is None else error.sum() / valid_mask.sum()
```

The shape is now static and the result is numerically identical (zeroed entries contribute nothing
to the sum, and the divisor is the count of valid pixels either way). All four disparity metrics are now covered by
`test_dynamo`, verified under the **inductor** backend on both CPU and CUDA (logs below).

---

## 🛠️ Changes Made
- [x] `kornia/metrics/disparity.py` — add `kitti_d1_error(input, target, abs_threshold=3.0, rel_threshold=0.05, valid_mask=None, reduction="mean")`, reusing the existing `_check_disparity_inputs` / `_reduce_disparity_error` helpers.
- [x] `kornia/metrics/disparity.py` — make `_reduce_disparity_error` `torch.compile`-safe (static shape, numerically identical).
- [x] `kornia/metrics/__init__.py` — export `kitti_d1_error`.
- [x] `docs/source/metrics.rst` — add `.. autofunction:: kitti_d1_error` to the existing Stereo section.
- [x] `tests/metrics/test_disparity.py` — add `TestKittiD1Error` plus a `test_dynamo` covering all four disparity metrics.

Reduction semantics match `mean_bad_pixel_error`: `mean` = outlier fraction in `[0, 1]`,
`sum` = outlier count, `none` = per-pixel 0/1 map with masked-out positions zeroed.

---

## 🧪 How Was This Tested?

- [x] **Unit Tests:** New `TestKittiD1Error` — `test_smoke`, `test_metric_mean_reduction`,
  `test_metric_sum_reduction`, `test_metric_no_reduction`, `test_perfect_prediction`,
  `test_absolute_criterion_alone_is_not_an_outlier`, `test_relative_criterion_alone_is_not_an_outlier`,
  `test_thresholds`, `test_valid_mask`, `test_zero_target_disparity`, `test_exception`.
  Added `test_dynamo` to all four disparity metric test classes.

- [x] **Manual Verification:** Full logs pasted below — inductor on CPU, inductor on CUDA, and the
  full `tests/metrics/` suite on CUDA as a regression check on the shared-helper change. Plus
  `pixi run lint` (passed), `pixi run typecheck` (`All checks passed!`), and doctests (4 passed).

  - **Zero ground-truth disparity** — no `eps` is used in the relative-error division. When
    `target == 0`, `rel` is `inf` (when `err > 0`) or `nan` (when `err == 0`). Because the criterion
    is an AND and `nan > rel_threshold` is `False` — and the `nan` case also has `err == 0`, which
    already fails `err > abs_threshold` — no NaN can reach the output. `test_zero_target_disparity`
    asserts the output stays finite. KITTI marks these pixels invalid anyway, so the docstring
    recommends passing `valid_mask`.
  - **Empty `valid_mask`** — `mean` reduction returns `nan`, documented in a `Note:`.
  - **Broadcasting / non-boolean masks** — handled by the existing `_check_disparity_inputs`.
  - Tested across `float32` and `float64`, on CPU and CUDA.

<details>
<summary><b>Run 1 — inductor backend, CPU (80 passed)</b></summary>

```
$ KORNIA_TEST_OPTIMIZER=inductor pixi run test-module tests/metrics/test_disparity.py --dtype=float32,float64

platform linux -- Python 3.11.14, pytest-9.0.3, pluggy-1.6.0
cpu info:
	- Model name: AMD Ryzen 7 7700X 8-Core Processor
gpu info: {'GPU 0': 'NVIDIA GeForce RTX 3090'}
main deps:
    - kornia-0.9.0rc1
    - torch-2.9.1+cu128
        - cuda: 12.8
        - nvidia-driver: 595.84
gcc info: (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0
available optimizers: {'', None, 'jit', 'tvm', 'cudagraphs', 'inductor', 'openxla'}

collected 80 items

tests/metrics/test_disparity.py::TestMeanAbsoluteDisparityError::test_dynamo[cpu-float32-inductor] PASSED [ 26%]
tests/metrics/test_disparity.py::TestMeanAbsoluteDisparityError::test_dynamo[cpu-float64-inductor] PASSED [ 27%]
tests/metrics/test_disparity.py::TestRootMeanSquaredDisparityError::test_dynamo[cpu-float32-inductor] PASSED [ 46%]
tests/metrics/test_disparity.py::TestRootMeanSquaredDisparityError::test_dynamo[cpu-float64-inductor] PASSED [ 47%]
tests/metrics/test_disparity.py::TestMeanBadPixelError::test_dynamo[cpu-float32-inductor] PASSED [ 68%]
tests/metrics/test_disparity.py::TestMeanBadPixelError::test_dynamo[cpu-float64-inductor] PASSED [ 70%]
tests/metrics/test_disparity.py::TestKittiD1Error::test_dynamo[cpu-float32-inductor] PASSED [ 98%]
tests/metrics/test_disparity.py::TestKittiD1Error::test_dynamo[cpu-float64-inductor] PASSED [100%]

============================== 80 passed in 12.27s ==============================
```

</details>

<details>
<summary><b>Run 2 — inductor backend, CUDA (80 passed)</b></summary>

```
$ KORNIA_TEST_DEVICE=cuda KORNIA_TEST_OPTIMIZER=inductor pixi run -e cuda uv run pytest -v tests/metrics/test_disparity.py --dtype=float32,float64

platform linux -- Python 3.11.14, pytest-9.0.3, pluggy-1.6.0
gpu info: {'GPU 0': 'NVIDIA GeForce RTX 3090'}
main deps:
    - kornia-0.9.0rc1
    - torch-2.9.1+cu128
        - cuda: 12.8
        - nvidia-driver: 595.84

collected 80 items

tests/metrics/test_disparity.py::TestMeanAbsoluteDisparityError::test_dynamo[cuda-float32-inductor] PASSED [ 26%]
tests/metrics/test_disparity.py::TestMeanAbsoluteDisparityError::test_dynamo[cuda-float64-inductor] PASSED [ 27%]
tests/metrics/test_disparity.py::TestRootMeanSquaredDisparityError::test_dynamo[cuda-float32-inductor] PASSED [ 46%]
tests/metrics/test_disparity.py::TestRootMeanSquaredDisparityError::test_dynamo[cuda-float64-inductor] PASSED [ 47%]
tests/metrics/test_disparity.py::TestMeanBadPixelError::test_dynamo[cuda-float32-inductor] PASSED [ 68%]
tests/metrics/test_disparity.py::TestMeanBadPixelError::test_dynamo[cuda-float64-inductor] PASSED [ 70%]
tests/metrics/test_disparity.py::TestKittiD1Error::test_dynamo[cuda-float32-inductor] PASSED [ 98%]
tests/metrics/test_disparity.py::TestKittiD1Error::test_dynamo[cuda-float64-inductor] PASSED [100%]

============================== 80 passed in 2.39s ==============================
```

</details>

<details>
<summary><b>Run 3 — full <code>tests/metrics/</code> suite on CUDA, regression check (262 passed)</b></summary>

```
$ KORNIA_TEST_DEVICE=cuda pixi run -e cuda uv run pytest -q tests/metrics/ --dtype=float32,float64

platform linux -- Python 3.11.14, pytest-9.0.3, pluggy-1.6.0
gpu info: {'GPU 0': 'NVIDIA GeForce RTX 3090'}
main deps:
    - kornia-0.9.0rc1
    - torch-2.9.1+cu128

collected 270 items

tests/metrics/test_accuracy.py ....                                      [  1%]
tests/metrics/test_aepe.py ..............                                [  6%]
tests/metrics/test_average_meter.py ......                               [  8%]
tests/metrics/test_confusion.py ......................                   [ 17%]
tests/metrics/test_disparity.py ........................................ [ 32%]
................................                                         [ 44%]
tests/metrics/test_map.py .....                                          [ 46%]
tests/metrics/test_mean_iou.py ..........................                [ 56%]
tests/metrics/test_pose.py .............................................. [ 73%]
..................                                                       [ 79%]
tests/metrics/test_psnr_metric.py .....                                  [ 81%]
tests/metrics/test_ssim3d.py ..............................              [ 93%]
tests/metrics/test_ssim_metric.py ..................                     [100%]

============================= 262 passed in 0.95s ==============================
```

</details>

<details>
<summary><b>Lint, typecheck, doctests</b></summary>

```
$ pixi run lint

$ pixi run typecheck
All checks passed!

$ pixi run uv run pytest --doctest-modules kornia/metrics/disparity.py
kornia/metrics/disparity.py::kornia.metrics.disparity.kitti_d1_error PASSED                    [ 25%]
kornia/metrics/disparity.py::kornia.metrics.disparity.mean_absolute_disparity_error PASSED     [ 50%]
kornia/metrics/disparity.py::kornia.metrics.disparity.mean_bad_pixel_error PASSED              [ 75%]
kornia/metrics/disparity.py::kornia.metrics.disparity.root_mean_squared_disparity_error PASSED [100%]

============================== 4 passed in 0.10s ===============================
```

</details>

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**

---

## 🚦 Checklist

---

## 💭 Additional Context

- **`Optional[X]` rather than `X | None`**: `pyproject.toml` still ignores UP007/UP045 for
  TorchScript compatibility, so the new signature follows the existing convention in this module.
- **No `nn.Module` wrappers in this PR.** I have wrappers written for all four disparity metrics
  mirroring the `AEPE`/`aepe` precedent, but the most recent metrics PR (#3889, pose metrics) added
  five functional metrics and zero wrappers, so functional-only appears to be the current
  preference.


kornia/metrics/disparity.py
```diff
@@ -51,13 +51,17 @@ def _check_disparity_inputs(
...
 def _reduce_disparity_error(error: torch.Tensor, valid_mask: Optional[torch.Tensor], reduction: str) -> torch.Tensor:
     """Reduce a per-pixel error map over the valid pixels according to ``reduction``."""
+    if valid_mask is not None:
+        # Zeroing the invalid pixels keeps the shape static. Indexing with the mask would make the
+        # output shape depend on the mask values, which torch.compile cannot trace in a full graph.
+        error = torch.where(valid_mask, error, torch.zeros_like(error))
+
     if reduction == "mean":
-        error = error.mean() if valid_mask is None else error[valid_mask].mean()
+        error = error.mean() if valid_mask is None else error.sum() / valid_mask.sum()
```
ducha-aiki: The compile-safe rewrite replaces `mean()` with:

```python
error.sum() / valid_mask.sum()
```

However, sum() accumulates in error.dtype. For float16 inputs, ordinary image-sized tensors can therefore overflow to inf.

For example, a masked 100,000-pixel D1 map containing only outliers returns inf instead of 1.0. Masked MAE and RMSE are affected similarly.
Please accumulate float16/bfloat16 values in float32 before division, preserve the documented output dtype if required, and add a large masked float16 regression test.

ducha-aiki (changes requested):
Please, run float16/bfloat16 tests, and see if we can do smth else there

kornia/metrics/disparity.py
VanniLeonardo: Fixed in 9629a0b1. The reduction now accumulates in `float32` and casts back:

```python
out_dtype = error.dtype
acc_dtype = torch.promote_types(out_dtype, torch.float32)
...
count = error.numel() if valid_mask is None else valid_mask.sum()
error = (error.sum(dtype=acc_dtype) / count).to(out_dtype)
```

`promote_types` maps `float16`/`bfloat16`/`float32` → `float32` and leaves `float64` alone, so
`float32` and `float64` are bit-unchanged. The output dtype is preserved, as the existing tests
require (`assert_close` checks dtype).

### Two more things

**1. `root_mean_squared_disparity_error` had a second, pre-existing half-precision failure.**
`(input - target) ** 2` saturates at a 256 px disparity error, since 256² = 65536 > 65504. So
`float16` RMSE returned `inf` for large errors *even unmasked and even with `reduction='none'`* —
this one is on `main` too, not introduced here. The square and its reduction now run in `float32`
and only the final root is cast back, which brings the value comfortably into range. Covered by
`test_large_error_does_not_saturate_the_square`.

**2. `reduction='sum'` still saturates in `float16` above 65504.** That one is not fixable without
changing the output dtype: a count of 100,000 outliers is simply not representable in `float16`.
It behaves exactly as it did before the compile-safe rewrite, and it is now documented in a `Note:`
on each of the four metrics, pointing at `reduction='none'` plus a wider-dtype sum as the way out.


### Runs

```
$ uv run pytest tests/metrics/test_disparity.py --dtype=float16,bfloat16
82 passed in 0.15s

$ KORNIA_TEST_OPTIMIZER=inductor uv run pytest tests/metrics/test_disparity.py --dtype=float16,bfloat16
90 passed in 3.67s

$ uv run pytest tests/metrics/test_disparity.py --dtype=float16,bfloat16,float32,float64
164 passed in 0.23s

$ KORNIA_TEST_OPTIMIZER=inductor uv run pytest tests/metrics/test_disparity.py -k dynamo \
      --dtype=float16,bfloat16,float32,float64
16 passed in 12.87s

$ uv run pytest tests/metrics/ --dtype=float16,bfloat16,float32,float64
1 failed, 529 passed
```

That single failure is `test_ssim3d.py::TestSSIM3d::test_unit[cpu-bfloat16]`, which is unrelated to
this PR.

`pixi run lint` passed, `pixi run typecheck` reported `All checks passed!`, and the 4 doctests pass.

kornia/metrics/disparity.py
```diff
@@ -51,13 +51,17 @@ def _check_disparity_inputs(
 
 def _reduce_disparity_error(error: torch.Tensor, valid_mask: Optional[torch.Tensor], reduction: str) -> torch.Tensor:
     """Reduce a per-pixel error map over the valid pixels according to ``reduction``."""
+    if valid_mask is not None:
+        # Zeroing the invalid pixels keeps the shape static. Indexing with the mask would make the
+        # output shape depend on the mask values, which torch.compile cannot trace in a full graph.
+        error = torch.where(valid_mask, error, torch.zeros_like(error))
+
```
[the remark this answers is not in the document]
VanniLeonardo: Applied `torch.where(valid_mask, error, 0)`. Verified the scalar keeps the tensor's
dtype for `float16`/`bfloat16`/`float32`/`float64` and still traces under `fullgraph=True`.

ducha-aiki:
**Changes requested — the D1 definition is exactly right (I verified it against the devkit line-by-line), but the float32-accumulation fix pushed in `9629a0b1` introduced a silent integer-truncation regression in the two already-shipped disparity metrics, and the new metric's most error-prone detail (`>` vs `>=`, and which map is the denominator) is not pinned by any test.**

Measured on head `9629a0b15bb1ac02568ef93d1f0221e2aecb42f9`, merge-base `9ec66f358` (not `main`), torch 2.9.1, Python 3.11.14, CPU.

### Timeline note

For the record, the last push **post-dates** the review: `CHANGES_REQUESTED` was `2026-08-10T09:04:51Z`, commit `9629a0b1` landed `2026-08-10T19:09:28Z`. This is a real code delta, not just a comment reply.

### What is already right

**The KITTI D1 predicate is correct, including both strict comparisons.** The devkit (`evaluate_scene_flow.cpp:13-14,51`) reads:

```cpp
#define ABS_THRESH 3.0
#define REL_THRESH 0.05
bool d_err = fabs(d_gt-d_est)>ABS_THRESH && fabs(d_gt-d_est)/fabs(d_gt)>REL_THRESH;
```

`disparity.py:329` is a faithful transcription: strict `>` on both, `AND` not `OR`, and `|d_gt|` (not `|d_est|`) in the denominator. Boundary probes I constructed:

```
=== A. exactly 3.0 px abs error, 30% rel (gt=10, pred=13) ===
err=3.0 rel=0.30 -> [0.0]        # devkit: d_err>3.0 is FALSE -> inlier. correct
=== B. 3.0000001 px abs error, gt=10 (float64) ===
-> [1.0]                          # correct
=== C. exactly 5% rel, abs=5 px (gt=100, pred=105) ===
err=5.0 rel=0.05 -> [0.0]        # devkit: rel>0.05 is FALSE -> inlier. correct
=== D. gt=100, pred=105.001 (float64) ===
-> [1.0]                          # correct
=== E. negative ground-truth disparity (devkit uses fabs(d_gt)) ===
gt=-100 pred=-90 err=10 rel=0.10 -> [1.0]   # correct
```

**Batch aggregation matches the benchmark.** The function pools over every element, and that is what KITTI does — the devkit accumulates raw `(n_err, n_px)` pairs across images and divides once at the end (`evaluate_scene_flow.cpp:485-488`), and the site says "averaged over all ground truth pixels of all 200 test images". So pooled is right, not per-image-averaged. I confirmed the two differ here and that you get the pooled one:

```
per-image D1 : [0.0, 1.0]
pooled (what kornia returns) : 0.2
mean of per-image D1         : 0.5
```

**The half-precision blocker from the last round is genuinely fixed.** Same probe at the first commit `41439d03` versus head:

```
===== 41439d03 (reviewed state) =====
fp16 masked bad-pixel mean (true value 1.0) -> tensor(inf, dtype=torch.float16)
fp16 RMSE with 300 px error (true value 300) -> tensor(inf, dtype=torch.float16)
===== 9629a0b1 (head) =====
fp16 masked bad-pixel mean (true value 1.0) -> tensor(1., dtype=torch.float16)
fp16 RMSE with 300 px error (true value 300) -> tensor(300., dtype=torch.float16)
```

The RMSE one is a bonus: `inf` there is a defect on the merge base too, so this PR fixes a pre-existing bug. Nice catch, and the `test_large_error_does_not_saturate_the_square` pin is the right shape.

Also confirmed green on my box: `82 passed` at `--dtype=float32,float64`, `82 passed` at `--dtype=float16,bfloat16`, 4 doctests pass, `ruff check`/`ruff format --check` clean, no line over 120 chars (the qodo line-length comment is stale). `__all__`, `docs/source/metrics.rst`, type hints, and the Menze & Geiger citation are all in place per AGENTS.md. Numerics of the helper rewrite are equivalent to the base — float32/float64 masked `mean`/`sum` agree to the last bit or two on random 3x64x64 inputs.

**Duplication disposition: KEEP SEPARATE.** Nothing in `kornia/metrics/` computes this. `mean_bad_pixel_error` is the Middlebury bad-N predicate (absolute only); D1 adds the conjunctive relative criterion and is not reachable by any parameter setting of the existing function. `grep -rniE "\bd1[_ -]?(all|error)\b|kitti" kornia/ --include=*.py` finds no other implementation.

---

### Blockers

**1. `disparity.py:70,72` — integer inputs now silently return a truncated wrong answer instead of raising. This is a regression on two already-public functions, introduced by `9629a0b1`.**

`.to(out_dtype)` casts the float32 accumulator back to the *input* dtype. For an integer disparity map that truncates the ratio to an int:

```
$ PYTHONPATH=<head> python probe1.py
=== G. integer dtype inputs ===
torch.int32 kitti mean = tensor(0, dtype=torch.int32)      # true value 0.5
torch.int32 mae mean = tensor(4, dtype=torch.int32)        # true value 4.5
torch.int32 badpix mean = tensor(0, dtype=torch.int32)     # true value 0.5
torch.int64 kitti mean = tensor(0)
torch.int64 mae mean = tensor(4)
torch.int64 badpix mean = tensor(0)
```

`uint8` is the same (`tensor(0, dtype=torch.uint8)`). On the merge base **and** on your own first commit `41439d03`, the identical call raised a clear error:

```
$ PYTHONPATH=<merge-base> python probe_base.py
torch.int32 mae raised RuntimeError mean(): could not infer output dtype. Input dtype must be
  either a floating point or complex dtype. Got: Int
torch.int64 badpix raised RuntimeError mean(): could not infer output dtype. ... Got: Long
```

So the fix for the fp16 blocker turned a loud failure into a silent wrong number, on `mean_absolute_disparity_error` and `mean_bad_pixel_error` as well as the new metric. This is not hypothetical: KITTI ships disparity as `uint16` PNG, and `torch.from_numpy(imread(...))` gives you an integer tensor. A D1 of `0` on real data is the worst possible failure mode.

Fix either way, but fix it: add `KORNIA_CHECK(input.is_floating_point(), ...)` to `_check_disparity_inputs` so integers raise a kornia-level error, or make the reduction return the accumulator dtype when `out_dtype` is not floating point. Whichever you pick, add a test that pins it — `pytest.raises` on an int input, or an equality assertion on the float result.

**2. `tests/metrics/test_disparity.py:332-467` — the two definitional details that distinguish D1 from a near-miss are not pinned by any test.** I mutated the implementation and the whole class still passes.

Mutation A, flipping both comparisons to non-strict:

```
$ sed -i '' 's/error > abs_threshold) & (error \/ target.abs() > rel_threshold/error >= abs_threshold) \& (error \/ target.abs() >= rel_threshold/' kornia/metrics/disparity.py
$ sed -n '329p' kornia/metrics/disparity.py
    outlier = ((error >= abs_threshold) & (error / target.abs() >= rel_threshold)).to(input.dtype)
$ pytest tests/metrics/test_disparity.py -q --device=cpu --dtype=float32,float64 -k KittiD1
====================== 24 passed, 58 deselected in 0.65s =======================
```

Mutation B, using the prediction instead of the ground truth as the relative denominator:

```
$ sed -n '329p' kornia/metrics/disparity.py
    outlier = ((error > abs_threshold) & (error / input.abs() > rel_threshold)).to(input.dtype)
$ pytest tests/metrics/test_disparity.py -q --device=cpu --dtype=float32,float64 -k KittiD1
====================== 24 passed, 58 deselected in 0.68s =======================
```

Both are wrong against the devkit and both are invisible to the suite. Please add hardcoded-literal cases (they are one-liners):

- `gt=10, pred=13` — error exactly `3.0`, relative 30% — expected `0.0`. Pins strict `>` on the absolute criterion.
- `gt=100, pred=105` — relative exactly `0.05`, absolute 5 px — expected `0.0`. Pins strict `>` on the relative criterion.
- `gt=78, pred=82` — error 4 px, `4/78 = 0.0513 > 0.05` but `4/82 = 0.0488 < 0.05` — expected `1.0`. Pins the ground-truth denominator. (Verified: current code returns `[1.0]`.)
- `gt=-100, pred=-90` — expected `1.0`. Pins `|d_gt|`.

The rest of the class already follows `TESTING.md` correctly — `BaseTester`, injected `device`/`dtype`, `self.assert_close`, and hardcoded expected literals rather than a reimplementation of the formula. This gap is about which literals were chosen, not about the style.

**3. `disparity.py:216-217` — now-false claim in a docstring this PR is editing.** `mean_bad_pixel_error` says "This corresponds to the bad-pixel percentage reported by the Middlebury and KITTI stereo benchmarks." Once `kitti_d1_error` exists in the same module that sentence actively misdirects: the KITTI number is D1, not bad-N. Drop "and KITTI" and add a `:func:` cross-reference to `kitti_d1_error`. Cheap, and it is in the blast radius of this change.

---

### Non-blocking nits

4. `disparity.py:299-314` — the docstring never says how a batch is aggregated. Since the answer (pooled over all valid pixels, not the mean of per-image D1) matches the devkit but *not* what several papers report, one sentence in the `Return:` block would prevent a whole class of "my numbers don't match the leaderboard" issues. Same for the percentage-vs-fraction point: it is stated, but in the prose paragraph rather than next to `Return:`.

5. The metric is not differentiable and does not say so. `kitti_d1_error(...).requires_grad` is `False` and `.backward()` raises `element 0 of tensors does not require grad and does not have a grad_fn`. That is correct for an indicator metric, but a one-line `Note:` saves someone a debugging session. (`mean_bad_pixel_error` has the same gap — fold it in.)

6. A `nan` prediction is silently counted as an *inlier*, because `nan > abs_threshold` is `False`:

```
=== K. NaN in prediction ===
none -> [0.0, 1.0]  mean -> 0.5
```

Meanwhile `mean_absolute_disparity_error` propagates the `nan`. Inconsistent within the module. The `Note:` about zero ground truth already covers the same mechanism; extend it to non-finite predictions.

7. `disparity.py:195` — `.to(input.dtype)` changes the return dtype when `input` and `target` disagree. On the merge base `rmse(float32_input, float64_target)` returned `torch.float64`; at head it returns `torch.float32`. `mean_absolute_disparity_error` still returns `float64` for the same pair, so the module is now inconsistent. Either check dtypes match in `_check_disparity_inputs` or cast to the promoted dtype.

8. `tests/metrics/test_disparity.py:349-353` — `test_smoke` draws both maps from `torch.rand` in `[0, 1)`, so the error can never exceed 3 px and the outlier branch is never taken. It only asserts the shape, so it is not wrong, just weaker than it looks.

9. Agreed with your reading on functional-only (no `nn.Module` wrapper) — that matches #3889. No action.

Once (1) and (2) are addressed I expect to approve; (3) is a two-word edit in a file you are already touching.

VanniLeonardo:
All three blockers and all six nits are addressed on `3d34a76f`, rebased onto current `main`.

**1. Integer inputs returned a truncated number**

You are right, and it is a regression this PR introduced. The accumulator cast back to the map's dtype, which for an integer map truncates. On the commit you reviewed:

```
int32 kitti_d1_error = 0    true value 0.5
int32 mean_bad_pixel = 0    true value 0.5
int32 mae            = 4    true value 4.5
```

I chose to reject rather than promote. The merge base already raised a `RuntimeError` out of torch for these calls, so raising restores the contract that existed before this PR, and it changes nothing for float callers. Promoting would have been a new public behaviour on two already-shipped functions, which is the same category of change as the one you flagged.

`_check_disparity_inputs` now rejects non-floating maps with a kornia error naming both dtypes, so all four metrics are covered by one guard. `test_convention_integer_disparity_maps_are_rejected` pins it across `uint8`, `int32` and `int64` and all four metrics, and also asserts the float result each integer call used to truncate. Removing the guard fails 12 nodes.

**2. The two definitional details were unpinned**

Confirmed, both of your mutations passed the suite. `test_convention_boundary_cases` pins four cases straight from the devkit predicate.

| ground truth, prediction | error | relative | expected | pins |
|---|---|---|---|---|
| 10, 13 | 3.0 | 0.300 | 0.0 | strict `>` on the absolute criterion |
| 100, 105 | 5.0 | 0.050 | 0.0 | strict `>` on the relative criterion |
| 78, 82 | 4.0 | 0.0513 against 0.0488 | 1.0 | the ground truth is the denominator |
| -100, -90 | 10.0 | 0.100 | 1.0 | `fabs(d_gt)` |

Making both comparisons non-strict now fails 4 nodes, and swapping the denominator to the prediction fails 2. Both passed before. The third case works because 4/78 and 4/82 straddle the 0.05 threshold, and I checked the margin survives `float16` and `bfloat16`, where the two ratios come out 0.05127 and 0.04883.

**3. The misleading `mean_bad_pixel_error` sentence**

Fixed. It names Middlebury only, says plainly that it is not the number KITTI reports, and cross-refers to `kitti_d1_error`.

**Nits**

All six done.

**4.** Both docstrings now state the aggregation next to `Return`, and that the result is a fraction rather than a percentage. Pooling only differs from a per-image mean when the images hold different numbers of valid pixels, which I recorded as 0.4 against 0.625 on such a pair.

**5.** Both indicator metrics say they carry no gradient. `test_convention_indicator_metrics_have_no_gradient` pins it against `mean_absolute_disparity_error`, which does propagate one. That contrast is the reason it seemed worth a test rather than only a sentence.

**6.** `kitti_d1_error` now documents that a `nan` is counted as an inlier, since every comparison against `nan` is false and both criteria must hold, and points at `valid_mask` rather than at either behaviour.

**7.** Fixed as a promotion rather than a check, since the merge base returned `float64` for a `float32` by `float64` pair and head had regressed that to `float32`. Every metric in the module now returns the promoted dtype, which is what `mean_absolute_disparity_error` always did, so the inconsistency is gone in the direction that restores the old behaviour. Pinned by `test_convention_result_dtype_is_the_promoted_one`.

**8.** The D1 smoke test drew both maps from `[0, 1)`, where the error can never reach 3 px, so the outlier branch never ran. It now uses a realistic disparity range and asserts that some pixel is actually flagged.

**9.** Agreed, no action.

**Also**

The branch had no CHANGELOG entry. I added one under `Added` for the metric and one under `Bug fixes` covering the half-precision accumulation, the squaring fix, the integer guard and the dtype change.

The rebase moved `docs/source/metrics.rst` too. Main split it into a toctree of per-section pages while this branch was parked, so `kitti_d1_error` is registered in `docs/source/metrics.stereo.rst` now.

**CI**

The MPS leg went red on my first push and that one was mine. The pin I added for nit 7 hardcoded a float32 by float64 pair, and MPS cannot allocate float64 at all, so all four of its nodes failed there. Fixed in `3d34a76f`: it now checks a float16 by float32 pair on every backend and adds the float64 pair only where float64 can be held, probed at runtime rather than by device name. The rule is still checked everywhere rather than skipped, and still checked on float32 by float64 where the regression was measured.

The remaining red leg is `tests (macos-latest, float32, 3.12, 2.5.1)`, and it is not from this PR.

```
  Failed: Timeout (>120.0s) from pytest-timeout
```

The same test times out on `main` with the same message, on the same macOS configuration, in the scheduled runs 34779403220 and 34804576631. This branch touches nothing outside `kornia/metrics`. The test takes 5.5s here against its 120s budget, so it reads as runner speed rather than the timeout #4402 set. I cannot re-run the job myself.

**Local verification on 3d34a76f**

```
pytest tests/metrics/ --dtype=float32,float64                      298 passed
pytest tests/ --dtype=float32                                    10442 passed, 3735 skipped
pytest tests/ --dtype=float16  --verify-known-failures            9839 passed, exit 0
pytest tests/ --dtype=bfloat16 --verify-known-failures            9893 passed, exit 0
pytest --doctest-modules kornia/                                   656 passed, 14 skipped
KORNIA_TEST_OPTIMIZER=inductor, metrics dynamo tests                 4 passed
pre-commit on the changed files                                   all hooks passed
pixi run typecheck                   1 pre-existing diagnostic in kornia/feature/lightglue.py
```

Mutation results for the three new pins, each against the fix it guards:

```
both comparisons non-strict          4 failed  (was 0)
prediction as relative denominator   2 failed  (was 0)
floating-point guard removed        12 failed  (was 0)
```

ducha-aiki (changes requested):
Review of `d90d1206aa6d9439652a0593f35959c84fe85bb5`.

The D1 implementation and all earlier numerical/test requests are addressed: the predicate is strict on both thresholds, divides by `abs(target)`, pools valid pixels, promotes mixed dtypes, accumulates half reductions in float32, rejects integer maps explicitly, and is exported and documented. [P2] Classify the integer-input change to the three existing metrics as breaking. The new guard at `kornia/metrics/disparity.py:42-46` now rejects every integer call, but on main `mean_absolute_disparity_error(..., reduction='sum'/'none')`, `root_mean_squared_disparity_error(..., reduction='sum'/'none')`, and `mean_bad_pixel_error(..., reduction='sum'/'none')` all return results. `changelog.d/3931.fixed.md:6-9` describes the rejection only as a fix. Move that compatibility change to `changelog.d/3931.breaking.md` and state the old working reductions and the new explicit error; keep the new D1 API in `3931.added.md` and the half/mixed-dtype fixes in `3931.fixed.md`.

Validation and limits: Focused CPU float16/bfloat16/float32/float64 run imported `/private/tmp/kornia-triage-60Oa1g/wt-3931/kornia/__init__.py`: 198 passed. All 41 hosted checks are green. A main-head reproduction with integer tensors returned valid `sum` and `none` results from all three existing functions, while this head raises at the shared guard.

ducha-aiki (approved):
Review of `b8e454da546bc8341897f35abaf18ef9e2bafab4`.

The only change since `d90d1206` is the requested release-note reclassification: the integer-map rejection moved out of `changelog.d/3931.fixed.md` into a new `changelog.d/3931.breaking.md`; source and tests are byte-identical to the reviewed head. Approving.

I executed the breaking fragment's before/after claims with `int64` maps (`input=[[1,2],[3,4]]`, `target=[[1,3],[3,6]]`), torch 2.14.0, CPU:

| call | main `53866348` | this head |
| --- | --- | --- |
| `mean_absolute_disparity_error`, `reduction="mean"` | torch `RuntimeError` from `mean()` | kornia error naming the dtypes |
| same, `"sum"` / `"none"` | `int64` `3` / `int64` `[0, 1, 0, 2]` | kornia error |
| `root_mean_squared_disparity_error`, `"sum"` / `"none"` | `float32` `2.236` / `float32` map | kornia error |
| `mean_bad_pixel_error`, `"sum"` / `"none"` | `int64` `1` / `int64` map | kornia error |

Every row matches the fragment text. All 42 hosted checks are green and the head merges cleanly with current `main`.

Issue #3752, opened 2026-06-15
[Feature]: adding SANDesc descriptor


mattiadurso:
### 🚀 Feature Description

Hi, I am the author of SANDesc (https://arxiv.org/pdf/2601.13126). I'd like to add support to our descriptor. A class, with pre-trained weights, that takes an image and its key-points (from a detector) and returns the descriptors. 

### 📂 Feature Category

Feature Detection/Matching

### 💡 Motivation

TLDR; SANDesc descriptors work better than vanilla methods on high-resolution images. DeDoDe, expecially  the -G version, runs out of memory fast with FHD images or larger. With SANDesc it can deliver high-quality matches even with only 24GB of VRAM. For ALIKED it usually works better than default version.

### 💭 Proposed Solution

I'll add two scripts, one for the network blocks and one for the whole descriptor. The weights will be available online so a user can just init the model and automatically download the pre-trained weights for ALIKED and DeDoDe.

It's all in torch, no other libs. 

### 🎯 Use Cases

See motivation.

### 📝 Additional Context

The code is already online in my public fork, as soon as the issue is approved I can send the PR.
https://github.com/mattiadurso/kornia/tree/sandesc

### 🤝 Contribution Intent

- [x] I plan to submit a PR to implement this feature
- [ ] I'm requesting this feature but not planning to implement it

edgarriba:
@ducha-aiki

ducha-aiki:
@mattiadurso sorry for the late response. Please, go ahead, that's a good contribution

mattiadurso:
Thank you! I've opened PR #3935 with the initial implementation for review. Looking forward to your feedback.

ducha-aiki:
Ouch. I mean thank you for the contribution, and I apologize for not doing my homework on license part. If the weights are non-commercial, we cannot add the descriptor to kornia - everything here is single license for effectively unrestricted use.
Is it possible to relicence weights?

mattiadurso:
I'll retrain on permissive datasets and submit again.

mattiadurso:
Following the licence problems, I retrained SANDesc for ALIKED (only) and slightly improved the training recipe along the way. I update the code in my branch for this. I also add a small addition ti ALIKED class to disable descriptors if needed, default is False. This is useful when ALIKED is paired with a different descriptor or only the key points re needed. Following some notes. Whenever ready we can go ahead with the PR. 

------------------------------------------------------------------------------------------------
## Datasets
[...]

PR #3935, opened 2026-08-10
Sandesc

mattiadurso:
## 📝 Description

**Fixes/Relates to:** #3752

Adds `kornia.feature.SANDesc`, a dense local descriptor to be paired with an existing keypoint
detector (ALIKED or DeDoDe). SANDesc is a U-Net-like encoder-decoder enhanced with Convolutional
Block Attention Modules (CBAM) and residual paths, producing a dense descriptor volume that is
sampled at the detected keypoints.

**Algorithm reference:** *A Streamlined Attention-Based Network for Descriptor Extraction*,
D'Urso et al., 3DV 2026 — DOI [10.1109/3DV69130.2026.00031](https://doi.org/10.1109/3DV69130.2026.00031).
Added to `docs/source/references.bib` as `durso2026sandesc` and cited from the class docstring.

**Reference implementations:**
- Descriptor network and training code: https://github.com/mattiadurso/SANDesc
- CBAM blocks follow https://github.com/Jongchan/attention-module/blob/master/MODELS/cbam.py
  (attributed in `kornia/feature/sandesc/_modules.py`)
- The U-Net skeleton mirrors the existing kornia DISK U-Net (`kornia/feature/disk/_unets.py`)

Pretrained weights: one checkpoint per detector (~9.2 MB, 2,400,694 parameters, 128-D descriptors),
fetched through `kornia.core.download.load_state_dict_from_url` with a primary source and two
fallbacks.

---

## 🛠️ Changes Made
- [x] New `kornia/feature/sandesc/` subpackage exporting `SANDesc`:
  - `SANDesc.forward(img, pad_if_not_divisible=False)` → dense descriptor volume `(B, D, H, W)`.
    Follows the `pad_if_not_divisible` convention already used by `DISK.forward` and
    `DeDoDe.describe`; without it, non-16-divisible inputs raise a `ValueError` pointing at the flag.
  - `SANDesc.describe(images, keypoints=None, ...)` → `(B, N, D)` descriptors, or the dense volume
    when `keypoints` is None, matching the `DeDoDe.describe` signature convention.
  - `SANDesc.from_pretrained(detector="aliked"|"dedode")` → weights trained to pair with that
    detector, loaded on CPU in eval mode.
  - Optional CUDA AMP via `amp` / `amp_dtype`, following `kornia/feature/dedode/encoder.py`.
- [x] `kornia/feature/sandesc/_modules.py`: U-Net down/up blocks and CBAM attention (private module,
  mirroring DISK's `_unets.py`).
- [x] Registered `SANDesc` in `kornia/feature/__init__.py` and `__all__`.
- [x] Docs: `.. autoclass:: SANDesc` in `docs/source/feature.rst` (Descriptors section),
  bib entry in `docs/source/references.bib`, and a response-map entry in `docs/generate_examples.py`
  rendering the per-pixel descriptor magnitude (`_static/img/SANDesc.png`, referenced from the
  class docstring).
- [x] Tests: `tests/feature/test_sandesc.py` (`TestSANDesc(BaseTester)`).
- [x] `CLAUDE.md`: listed SANDesc among the `kornia/feature/` models.

---

## 🧪 How Was This Tested?

- [x] **Unit Tests:** `tests/feature/test_sandesc.py` — `test_smoke`, `test_smoke_options`
  (all 8 skip/attention/third-block combinations), `test_non_uniform_channels`, `test_cardinality`,
  `test_describe`, `test_pad_if_not_divisible`, `test_exception`, `test_gradcheck`, `test_dynamo`,
  plus `@pytest.mark.slow` `test_pretrained_values` (hardcoded reference descriptors for both
  checkpoints), `test_pretrained_aliked_pipeline` and `test_pretrained_dedode_pipeline`
  (end-to-end detector → descriptor).
  Inputs are deterministic ramps rather than RNG, and the pretrained references are embedded as
  tensor literals with the generating snippet in a comment, so no optional third-party dependency
  is needed to run the correctness tests.

```console
$ pixi run test-module tests/feature/test_sandesc.py --dtype=float32,float64
collected 45 items

tests/feature/test_sandesc.py::TestSANDesc::test_smoke[cpu-float32] PASSED               [  2%]
tests/feature/test_sandesc.py::TestSANDesc::test_smoke[cpu-float64] PASSED               [  4%]
tests/feature/test_sandesc.py::TestSANDesc::test_smoke_options[cpu-float32-False-False-False] PASSED [  6%]
...
tests/feature/test_sandesc.py::TestSANDesc::test_non_uniform_channels[cpu-float32] PASSED [ 44%]
tests/feature/test_sandesc.py::TestSANDesc::test_non_uniform_channels[cpu-float64] PASSED [ 46%]
tests/feature/test_sandesc.py::TestSANDesc::test_cardinality[cpu-float32-1-1] PASSED     [ 48%]
...
tests/feature/test_sandesc.py::TestSANDesc::test_describe[cpu-float32] PASSED            [ 76%]
tests/feature/test_sandesc.py::TestSANDesc::test_describe[cpu-float64] PASSED            [ 79%]
tests/feature/test_sandesc.py::TestSANDesc::test_pad_if_not_divisible[cpu-float32] PASSED [ 81%]
tests/feature/test_sandesc.py::TestSANDesc::test_pad_if_not_divisible[cpu-float64] PASSED [ 83%]
tests/feature/test_sandesc.py::TestSANDesc::test_exception[cpu-float32] PASSED           [ 86%]
tests/feature/test_sandesc.py::TestSANDesc::test_exception[cpu-float64] PASSED           [ 88%]
tests/feature/test_sandesc.py::TestSANDesc::test_gradcheck[cpu] PASSED                   [ 90%]
tests/feature/test_sandesc.py::TestSANDesc::test_pretrained_values[cpu-aliked-expected0] SKIPPED [ 93%]
tests/feature/test_sandesc.py::TestSANDesc::test_pretrained_values[cpu-dedode-expected1] SKIPPED [ 95%]
tests/feature/test_sandesc.py::TestSANDesc::test_pretrained_aliked_pipeline[cpu] SKIPPED [ 97%]
tests/feature/test_sandesc.py::TestSANDesc::test_pretrained_dedode_pipeline[cpu] SKIPPED [100%]

======================== 39 passed, 4 skipped in 1.14s =========================
```

Slow tests (downloads the pretrained checkpoints):

```console
$ KORNIA_TEST_RUNSLOW=true pixi run test-module tests/feature/test_sandesc.py --dtype=float32
collected 25 items

tests/feature/test_sandesc.py::TestSANDesc::test_pretrained_values[cpu-aliked-expected0] PASSED [ 87%]
tests/feature/test_sandesc.py::TestSANDesc::test_pretrained_values[cpu-dedode-expected1] PASSED [ 91%]
tests/feature/test_sandesc.py::TestSANDesc::test_pretrained_aliked_pipeline[cpu] PASSED  [ 95%]
tests/feature/test_sandesc.py::TestSANDesc::test_pretrained_dedode_pipeline[cpu] PASSED  [100%]

============================== 24 passed in 4.46s ==============================
```

CUDA (RTX 4090, CUDA 12.8), same file including the slow tests:

```console
$ KORNIA_TEST_RUNSLOW=true uv run pytest tests/feature/test_sandesc.py -q --device=cuda --dtype=float32
collected 25 items
============================== 24 passed in 3.60s ==============================
```

`torch.compile` compatibility (deselected unless `KORNIA_TEST_OPTIMIZER` is set):

```console
$ KORNIA_TEST_OPTIMIZER=inductor uv run pytest tests/feature/test_sandesc.py::TestSANDesc::test_dynamo -v --dtype=float32
tests/feature/test_sandesc.py::TestSANDesc::test_dynamo[cpu-float32-inductor] PASSED [100%]
============================== 1 passed in 2.51s ===============================
```

Lint, type check and doctests:

```console
$ pixi run lint

$ pixi run typecheck
All checks passed!

$ uv run pytest --doctest-modules kornia/feature/sandesc/ -q
1 passed in 0.10s
```

- [x] **Manual Verification:** downloaded both checkpoints into a clean `TORCH_HOME` with only the
  primary URL available (no fallback) and loaded them with the default strict `load_state_dict`:

```console
aliked: loaded strict from HF, 2400694 params, training=False
dedode: loaded strict from HF, 2400694 params, training=False
cached files: ['sandesc_aliked.pth', 'sandesc_dedode.pth']
```

  - Input shape is validated with `KORNIA_CHECK_SHAPE(img, ["B", "C", "H", "W"])`; keypoints with
    `["B", "N", "2"]`.
  - Non-16-divisible resolutions either raise a descriptive `ValueError` or are zero-padded and
    cropped back when `pad_if_not_divisible=True` (`test_pad_if_not_divisible` uses 30x33).
  - Channel-list lengths, unknown `norm` / `activation` / pool types and unknown `detector` are all
    rejected at construction time, before any download (`test_exception`).
  - Non-uniform `up_output_channels` combined with `skip_connection=True` are covered by
    `test_non_uniform_channels`; the residual `align` convolution is sized from the incoming
    (coarser) block width rather than the output width.
  - Attention and the extra blocks are only instantiated on the skip path, so no unreachable
    parameters end up in the `state_dict`; `test_smoke_options` asserts this for all 8 combinations.
  - `describe` upcasts half/bfloat16 volumes before `grid_sample`, which does not support them, so
    the sparse path works under autocast.
  - Batch sizes 1/2/5 and 1/10 keypoints are covered by `test_cardinality`; gradients are checked in
    float64 via `test_gradcheck`.

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**

---

## 🚦 Checklist
- [ ] I am assigned to the linked issue (required before PR submission)
- [ ] The linked issue has been approved by a maintainer

---

## 💭 Additional Context

**Weight hosting.** The checkpoints are currently served from
`https://huggingface.co/mattia-durso/SANDesc`, with the author's GitHub repo and a TU Graz cloud
share as fallbacks. Other kornia models use the `kornia` HuggingFace organisation via
`kornia.core.download.hf_url`. Happy to re-point `urls` to `hf_url("sandesc", ...)` if a maintainer
mirrors the two files under `huggingface.co/kornia`.

**Weight licence.** The pretrained weights are released under `sony-research-nc` (non-commercial),
which is more restrictive than kornia's Apache-2.0. The code in this PR is Apache-2.0; only the
downloadable checkpoints carry the model-card licence. Let me know if that needs to be surfaced
differently (e.g. in the class docstring or the model card link).

**Scope.** SANDesc is a descriptor only — it has no detector head — so it is listed under
*Descriptors* in `docs/source/feature.rst`, and the generated example visualises the per-pixel
descriptor magnitude (L2 norm over channels) instead of a keypoint response map.

### DIFF
--- a/.github/download-models-weights.py
+++ b/.github/download-models-weights.py
@@ -90,6 +90,8 @@ MODELS: dict[str, "str | list[str]"] = {
         "https://huggingface.co/kornia/disk/resolve/main/depth-save.pth",
         "https://raw.githubusercontent.com/cvlab-epfl/disk/master/depth-save.pth",
     ],
+    # SANDesc descriptor: the docs example build calls SANDesc.from_pretrained.
+    "sandesc_aliked.pth": "https://cloud.tugraz.at/index.php/s/dBiF999GBMoRg8w/download/sandesc_aliked.pth",
     # -- matchers ------------------------------------------------------------
     # LoFTR: tests instantiate both the outdoor and indoor weights.
     "loftr_outdoor.ckpt": [
--- a/.github/download-models-weights.py
+++ b/.github/download-models-weights.py
@@ -90,8 +90,11 @@ MODELS: dict[str, "str | list[str]"] = {
         "https://huggingface.co/kornia/disk/resolve/main/depth-save.pth",
         "https://raw.githubusercontent.com/cvlab-epfl/disk/master/depth-save.pth",
     ],
-    # SANDesc descriptor: the docs example build calls SANDesc.from_pretrained.
-    "sandesc_aliked.pth": "https://cloud.tugraz.at/index.php/s/dBiF999GBMoRg8w/download/sandesc_aliked.pth",
+    # SANDesc descriptor: the pretrained tests and the class doctest call SANDesc.from_pretrained.
+    "sandesc_aliked.pth": [
+        "https://huggingface.co/mattia-durso/SANDesc/resolve/main/pretrained/sandesc_aliked.pth?download=true",
+        "https://cloud.tugraz.at/index.php/s/dBiF999GBMoRg8w/download/sandesc_aliked.pth",
+    ],
     # -- matchers ------------------------------------------------------------
     # LoFTR: tests instantiate both the outdoor and indoor weights.
     "loftr_outdoor.ckpt": [
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ benchmarks/profile/
 
 # Editor / agent metadata
 .omc/
+pr_body.txt
--- a/docs/generate_examples.py
+++ b/docs/generate_examples.py
@@ -699,6 +699,7 @@ def main():
         "DISK",
         "ALIKED",
         "XFeat",
+        "SANDesc",
     ]
     # ITERATE OVER THE TRANSFORMS
     for fn_name in responses:
@@ -732,6 +733,19 @@ def main():
             out = torch.nn.functional.interpolate(out, img_outdoor.shape[-2:], mode="bilinear", align_corners=False)
             out = K.color.grayscale_to_rgb(out)
             img_in = K.color.rgb_to_bgr(img_outdoor)
+        elif fn_name == "SANDesc":
+            # SANDesc has no detector head; visualize the per-pixel descriptor
+            # magnitude (L2 norm over channels) as the response map.
+            fn = K.feature.SANDesc.from_pretrained(detector="aliked")
+            h, w = img_outdoor.shape[2:]
+            pd_h = 16 - h % 16 if h % 16 > 0 else 0
+            pd_w = 16 - w % 16 if w % 16 > 0 else 0
+            img_in = torch.nn.functional.pad(img_outdoor, (0, pd_w, 0, pd_h), value=0.0)
+            with torch.no_grad():
+                volume = fn(img_in)
+            out = volume.norm(dim=1, keepdim=True)
+            out = K.color.grayscale_to_rgb(out)
+            img_in = K.color.rgb_to_bgr(img_in)
         else:
             fn = getattr(mod, fn_name)
             out = fn(img_in)
--- a/docs/generate_examples.py
+++ b/docs/generate_examples.py
@@ -737,15 +737,11 @@ def main():
             # SANDesc has no detector head; visualize the per-pixel descriptor
             # magnitude (L2 norm over channels) as the response map.
             fn = K.feature.SANDesc.from_pretrained(detector="aliked")
-            h, w = img_outdoor.shape[2:]
-            pd_h = 16 - h % 16 if h % 16 > 0 else 0
-            pd_w = 16 - w % 16 if w % 16 > 0 else 0
-            img_in = torch.nn.functional.pad(img_outdoor, (0, pd_w, 0, pd_h), value=0.0)
             with torch.no_grad():
-                volume = fn(img_in)
+                volume = fn(img_outdoor, pad_if_not_divisible=True)
             out = volume.norm(dim=1, keepdim=True)
             out = K.color.grayscale_to_rgb(out)
-            img_in = K.color.rgb_to_bgr(img_in)
+            img_in = K.color.rgb_to_bgr(img_outdoor)
         else:
             fn = getattr(mod, fn_name)
             out = fn(img_in)
--- a/docs/generate_examples.py
+++ b/docs/generate_examples.py
@@ -736,9 +736,9 @@ def main():
         elif fn_name == "SANDesc":
             # SANDesc has no detector head; visualize the per-pixel descriptor
             # magnitude (L2 norm over channels) as the response map.
-            fn = K.feature.SANDesc.from_pretrained(detector="aliked")
+            fn = K.feature.SANDesc.from_pretrained(load_detector=False)
             with torch.no_grad():
-                volume = fn(img_outdoor, pad_if_not_divisible=True)
+                volume = fn.extract_dense_map(img_outdoor, pad_if_not_divisible=True)
             out = volume.norm(dim=1, keepdim=True)
             out = K.color.grayscale_to_rgb(out)
             img_in = K.color.rgb_to_bgr(img_outdoor)
--- a/docs/generate_examples.py
+++ b/docs/generate_examples.py
@@ -699,7 +699,6 @@ def main():
         "DISK",
         "ALIKED",
         "XFeat",
-        "SANDesc",
     ]
     # ITERATE OVER THE TRANSFORMS
     for fn_name in responses:
@@ -731,15 +730,6 @@ def main():
                 out = K.feature.XFeat._get_kpts_heatmap(K1)
             # upsample heatmap from preprocessed size back to original image size
             out = torch.nn.functional.interpolate(out, img_outdoor.shape[-2:], mode="bilinear", align_corners=False)
-            out = K.color.grayscale_to_rgb(out)
-            img_in = K.color.rgb_to_bgr(img_outdoor)
-        elif fn_name == "SANDesc":
-            # SANDesc has no detector head; visualize the per-pixel descriptor
-            # magnitude (L2 norm over channels) as the response map.
-            fn = K.feature.SANDesc.from_pretrained(load_detector=False)
-            with torch.no_grad():
-                volume = fn.extract_dense_map(img_outdoor, pad_if_not_divisible=True)
-            out = volume.norm(dim=1, keepdim=True)
             out = K.color.grayscale_to_rgb(out)
             img_in = K.color.rgb_to_bgr(img_outdoor)
         else:
--- a/docs/source/feature.rst
+++ b/docs/source/feature.rst
@@ -140,7 +140,7 @@ Descriptors
 .. autoclass:: TFeat
 .. autoclass:: SOSNet
 .. autoclass:: SANDesc
-   :members: describe, extract_dense_map, forward, from_pretrained
+   :members: extract_dense_map, forward, from_pretrained
 .. autoclass:: LAFDescriptor
    :members: forward
 
--- a/docs/source/feature.rst
+++ b/docs/source/feature.rst
@@ -140,7 +140,7 @@ Descriptors
 .. autoclass:: TFeat
 .. autoclass:: SOSNet
 .. autoclass:: SANDesc
-   :members: describe, forward, from_pretrained
+   :members: describe, extract_dense_map, forward, from_pretrained
 .. autoclass:: LAFDescriptor
    :members: forward
 
--- a/docs/source/feature.rst
+++ b/docs/source/feature.rst
@@ -139,6 +139,8 @@ Descriptors
 .. autoclass:: HyNet
 .. autoclass:: TFeat
 .. autoclass:: SOSNet
+.. autoclass:: SANDesc
+   :members: describe, forward, from_pretrained
 .. autoclass:: LAFDescriptor
    :members: forward
 
--- a/docs/source/references.bib
+++ b/docs/source/references.bib
@@ -342,6 +342,15 @@ month = {mar}
   year={2024}
 }
 
+@inproceedings{durso2026sandesc,
+  title={A Streamlined Attention-Based Network for Descriptor Extraction},
+  author={D'Urso, Mattia and Santellani, Emanuele and Sormann, Christian and Rossi, Mattia and Kuhn, Andreas and Fraundorfer, Friedrich},
+  booktitle={2026 International Conference on 3D Vision (3DV)},
+  pages={247--256},
+  year={2026},
+  doi={10.1109/3DV69130.2026.00031}
+}
+
 @inproceedings{he2010guided,
 	title = {Guided Image Filtering},
   booktitle = {Proceedings of the 11th European Conference on Computer Vision: Part I},
--- a/kornia/feature/__init__.py
+++ b/kornia/feature/__init__.py
@@ -89,6 +89,7 @@ from .responses import (
     harris_response,
     hessian_response,
 )
+from .sandesc import SANDesc
 from .scale_space_detector import MultiResolutionDetector, PassLAF, ScaleSpaceDetector
 from .siftdesc import DenseSIFTDescriptor, SIFTDescriptor
 from .sold2 import SOLD2, SOLD2_detector
@@ -141,6 +142,7 @@ __all__ = [
     "PassLAF",
     "PatchAffineShapeEstimator",
     "PatchDominantGradientOrientation",
+    "SANDesc",
     "SIFTDescriptor",
     "SIFTFeature",
     "SIFTFeatureScaleSpace",
--- a/kornia/feature/aliked/aliked.py
+++ b/kornia/feature/aliked/aliked.py
@@ -83,12 +83,13 @@ class ALIKEDFeatures:
 
     Args:
         keypoints: pixel coordinates ``(N, 2)`` as ``[x, y]``.
-        descriptors: L2-normalised descriptors ``(N, D)``.
+        descriptors: L2-normalised descriptors ``(N, D)``, or ``None`` when the model was
+            built with ``disable_descriptors=True``.
         keypoint_scores: detection confidence scores ``(N,)``.
     """
 
     keypoints: torch.Tensor
-    descriptors: torch.Tensor
+    descriptors: Optional[torch.Tensor]
     keypoint_scores: torch.Tensor
 
     @property
@@ -100,7 +101,7 @@ class ALIKEDFeatures:
         """Move all tensors to a new device / dtype."""
         return ALIKEDFeatures(
             self.keypoints.to(*args, **kwargs),
-            self.descriptors.to(*args, **kwargs),
+            None if self.descriptors is None else self.descriptors.to(*args, **kwargs),
             self.keypoint_scores.to(*args, **kwargs),
         )
 
@@ -388,7 +389,12 @@ class InputPadder:
         self.wd = w
         pad_ht = (((h // divis_by) + 1) * divis_by - h) % divis_by
         pad_wd = (((w // divis_by) + 1) * divis_by - w) % divis_by
-        self._pad = [pad_wd // 2, pad_wd - pad_wd // 2, pad_ht // 2, pad_ht - pad_ht // 2]
+        self._pad = [
+            pad_wd // 2,
+            pad_wd - pad_wd // 2,
+            pad_ht // 2,
+            pad_ht - pad_ht // 2,
+        ]
 
     def pad(self, x: torch.Tensor) -> torch.Tensor:
         """Pad *x* (BCHW)."""
@@ -504,7 +510,14 @@ def get_conv(
 ) -> nn.Module:
     """Return a standard or deformable conv2d layer."""
     if conv_type == "conv":
-        return nn.Conv2d(inplanes, planes, kernel_size=kernel_size, stride=stride, padding=padding, bias=bias)
+        return nn.Conv2d(
+            inplanes,
+            planes,
+            kernel_size=kernel_size,
+            stride=stride,
+            padding=padding,
+            bias=bias,
+        )
     if conv_type == "dcn":
         return DeformableConv2d(
             inplanes,
@@ -649,9 +662,23 @@ class SDDH(nn.Module):
 
         self.channel_num = 3 * n_pos if mask else 2 * n_pos
         self.offset_conv = nn.Sequential(
-            nn.Conv2d(dims, self.channel_num, kernel_size=kernel_size, stride=1, padding=0, bias=True),
+            nn.Conv2d(
+                dims,
+                self.channel_num,
+                kernel_size=kernel_size,
+                stride=1,
+                padding=0,
+                bias=True,
+            ),
             gate,
-            nn.Conv2d(self.channel_num, self.channel_num, kernel_size=1, stride=1, padding=0, bias=True),
+            nn.Conv2d(
+                self.channel_num,
+                self.channel_num,
+                kernel_size=1,
+                stride=1,
+                padding=0,
+                bias=True,
+            ),
         )
         self.sf_conv = nn.Conv2d(dims, dims, kernel_size=1, stride=1, padding=0, bias=False)
 
@@ -764,6 +791,9 @@ class ALIKED(nn.Module):
             ``-1`` means no limit (threshold-based mode).
         detection_threshold: minimum detection score in threshold mode.
         nms_radius: NMS radius (kernel size ``= 2 * nms_radius + 1``).
+        disable_descriptors: if ``True``, the descriptor head is not built and the descriptor
+            outputs of :meth:`forward` and :meth:`forward_laf` are ``None``. Use it when the
+            descriptors come from another model, e.g. :class:`~kornia.feature.SANDesc`, or not needed.
 
     Example:
         >>> aliked = ALIKED.from_pretrained('aliked-n16')  # doctest: +SKIP
@@ -779,6 +809,7 @@ class ALIKED(nn.Module):
         max_num_keypoints: int = -1,
         detection_threshold: float = 0.2,
         nms_radius: int = 2,
+        disable_descriptors: bool = False,
     ) -> None:
         super().__init__()
 
@@ -789,6 +820,7 @@ class ALIKED(nn.Module):
         self.max_num_keypoints = max_num_keypoints
         self.detection_threshold = detection_threshold
         self.nms_radius = nms_radius
+        self.disable_descriptors = disable_descriptors
 
         c1, c2, c3, c4, dim, K, M = _ALIKED_CFGS[model_name]
         conv_types = ["conv", "conv", "dcn", "dcn"]
@@ -824,7 +856,7 @@ class ALIKED(nn.Module):
             self.gate,
             _conv3x3(4, 1),
         )
-        self.desc_head = SDDH(dim, K, M, gate=self.gate, conv2d=conv2d, mask=mask)
+        self.desc_head = None if disable_descriptors else SDDH(dim, K, M, gate=self.gate, conv2d=conv2d, mask=mask)
         self.dkd = DKD(
             radius=nms_radius,
             top_k=-1 if detection_threshold > 0 else max_num_keypoints,
@@ -897,14 +929,15 @@ class ALIKED(nn.Module):
 
         Returns:
             A list of :class:`ALIKEDFeatures` of length B, one per image.
-            Keypoints are in pixel coordinates ``[x, y]``.
+            Keypoints are in pixel coordinates ``[x, y]``. ``descriptors`` is ``None`` when the
+            model was built with ``disable_descriptors=True``.
         """
         if images.shape[1] == 1:
             images = grayscale_to_rgb(images)
 
         feature_map, score_map = self.extract_dense_map(images)
         keypoints, kptscores, _scoredispersitys = self.dkd(score_map, image_size=image_size)
-        descriptors, _offsets = self.desc_head(feature_map, keypoints)
+        descriptors = None if self.desc_head is None else self.desc_head(feature_map, keypoints)[0]
 
         B = images.shape[0]
         _, _, h, w = images.shape
@@ -914,7 +947,8 @@ class ALIKED(nn.Module):
         for i in range(B):
             # Convert normalised [-1,1] coords back to pixel coordinates
             kps_px = wh * (keypoints[i] + 1) / 2.0
-            results.append(ALIKEDFeatures(kps_px, descriptors[i], kptscores[i]))
+            desc_i = None if descriptors is None else descriptors[i]
+            results.append(ALIKEDFeatures(kps_px, desc_i, kptscores[i]))
         return results
 
     def forward_laf(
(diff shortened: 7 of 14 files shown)


mattiadurso:
Code looks ok (I guess). I'll retrain the weights and updated the PR.

ducha-aiki:
**Verdict: CHANGES REQUESTED — the implementation is in genuinely good shape and the reported `align_corners` bug is really fixed, but the licence provenance of the *code* (not just the checkpoint) is unresolved and is a hard gate for merging.**

Reviewed at head `c6485227`, merge-base `f772cb30`. Everything below was measured on CPU with torch 2.9.1 / Python 3.11.14 / kornia 0.9.0rc1, in a clean worktree of the PR head. I downloaded the real checkpoint, so the pretrained results are measured and not assumed.

---

## What is already right

This is a careful contribution and a lot of the hard parts are done properly.

- **The `align_corners` bug reported by the automated reviewer is genuinely fixed, not just marked resolved.** I verified it end to end rather than trusting the marker: I subclassed `SANDesc` so `extract_dense_map` returns a volume whose channel 0 holds the x index and channel 1 the y index of each pixel, then normalised known pixel targets with `normalize_pixel_coordinates` and sampled them back through the real `describe()`:

  ```
  default keypoint_align_corners = True

  target px           ->  normalized
    [0.0, 0.0]    ->  [-1.0, -1.0]
    [7.0, 5.0]    ->  [-0.7021, -0.6774]
    [47.0, 31.0]  ->  [1.0, 1.0]
    [23.0, 11.0]  ->  [-0.0213, -0.2903]

  align_corners=None (module default): sampled px = [[0.0, 0.0], [7.0, 5.0], [47.0, 31.0], [23.0, 11.0]]
      max |sampled - target| = 0.0
  align_corners=True:  max |sampled - target| = 0.0
  align_corners=False: sampled px = [[0.0, 0.0], [7.0, 5.0], [0.0, 0.0], [23.0, 11.0]]
      max |sampled - target| = 47.0
  ```

  The module default reads exactly the intended pixel. `normalize_pixel_coordinates` uses `2/(W-1)` and ALIKED inverts with `wh = [W-1, H-1]`, so the `align_corners=True` default in `kornia/feature/sandesc/sandesc.py:41` is the correct convention for the ALIKED pairing, and `describe()` now exposes a per-call override. `tests/feature/test_sandesc.py:121` pins all three behaviours. Good fix.

- **The test suite is well above the usual bar for a new model.** `TestSANDesc(BaseTester)` with injected `device`/`dtype` fixtures, and it covers small images `(1,1)`/`(4,7)`, extreme-value images, out-of-bounds keypoints, empty keypoints, non-uniform channel lists, all 8 skip/attention/third-block combinations, `gradcheck`, and `test_dynamo` via `torch_optimizer`. The pretrained expectation at `tests/feature/test_sandesc.py:302` is a hardcoded literal with the generating snippet in a comment, exactly as `AGENTS.md` asks, and it pins `cudnn.allow_tf32=False` because conftest's `--tf32` only touches matmul precision. All of it passes here:

  ```
  $ KORNIA_TEST_RUNSLOW=true pytest tests/feature/test_sandesc.py -q --device=cpu --dtype=float32
  38 passed in 20.22s

  $ pytest tests/feature/test_sandesc.py tests/feature/test_aliked.py -q --device=cpu --dtype=float32
  56 passed, 7 skipped in 8.09s

  $ KORNIA_TEST_OPTIMIZER=inductor pytest tests/feature/test_sandesc.py::TestSANDesc::test_dynamo -q
  1 passed in 8.29s
  ```

- **The docs requirement in `AGENTS.md` for a new detector/descriptor is actually met.** `docs/generate_examples.py:736` adds the `SANDesc` branch and the entry in `responses`, and it follows the `ALIKED`/`XFeat` pattern including the `(B, 3, H, W)` contract. Verified on a real non-16-divisible photo shape:

  ```
  docs branch: volume (1, 128, 375, 500) -> response map (1, 3, 375, 500)
  matches img_in HW for torch.cat: True
  ```

  `docs/source/feature.rst:142`, `__all__` in `kornia/feature/__init__.py:145`, and the `durso2026sandesc` bib entry are all present too. This is the item most new-model PRs miss.

- **It works, and the descriptors are exactly L2-normalised.** With the real checkpoint:

  ```
  params: 2761570  |  training mode: False
  dense volume: (2, 128, 240, 320) torch.float32 cpu
  forward -> kp (2, 256, 2) scores (2, 256) desc (2, 256, 128) torch.float32
  descriptor L2 norms: min 1.000000 max 1.000000
  ```

- **`weights_only=True` is used** at `kornia/feature/sandesc/sandesc.py:261` — this is currently the *only* `weights_only=True` in the whole tree, so thank you for setting the better precedent.

- **The `ALIKED(disable_descriptors=True)` addition is clean and well-tested.** `Optional[Tensor]` threaded through `ALIKEDFeatures`, `.to()`, `forward` and `forward_laf`, with `tests/feature/test_aliked.py:195` asserting the head is absent from the `state_dict`. I confirmed the pre-existing `strict=False` at `aliked.py:1078` is what makes the unused descriptor weights harmless — that is not new breakage from this PR.

- **The earlier automated findings are fixed**: `get_norm("group")` now rejects non-divisible channels explicitly (`_modules.py:37`) instead of raising deep inside `nn.GroupNorm`, and `ChannelGate` clamps the hidden width with `max(gate_channels // reduction_ratio, 1)` (`_modules.py:256`). CBAM is attributed to the upstream repo at `_modules.py:199`, and I confirmed that upstream (`Jongchan/attention-module`) is MIT, so that part is Apache-compatible.

- Ruff check, ruff format and the module doctest all pass.

---

## Blockers

### 1. The reference implementation is under a Sony non-commercial licence that covers the *code*, not only the checkpoint

The PR description says:

> The code in this PR is Apache-2.0; only the downloadable checkpoints carry the model-card licence.

I could not confirm that, and the evidence points the other way. The `LICENSE.md` at the **root** of `mattiadurso/SANDesc` — the repo this PR names as the source of the "descriptor network and training code" — is a Sony Group Corporation licence, and GitHub classifies the repo as `NOASSERTION` / "Other":

```
$ gh api repos/mattiadurso/SANDesc --jq '{license:.license.spdx_id, licname:.license.name}'
{"license":"NOASSERTION","licname":"Other"}

$ gh api repos/mattiadurso/SANDesc/contents/LICENSE.md | base64 -d | head -2
Copyright 2026 Sony Group Corporation.
All rights reserved.
```

That licence carries terms that Apache-2.0 cannot absorb, and they explicitly bind **Derivative Works**, not just the weights:

- **§4 Commercial Use Limitation** — "The Work and Derivative Works only may be used or intended for use non-commercially... 'non-commercially' means research, education or evaluation purposes only." Apache-2.0 §2 grants use with no field-of-use restriction.
- **§5 Military Use Limitation** — a second field-of-use restriction.
- **§3.3 Remote Network Interaction** — an AGPL-style source-disclosure obligation on network services.
- **§3.4(b)** — "Your version must bear a name that is different from the name of the Work provided by the copyright owner." The class here is literally `SANDesc`.
- **§3.4(a)** — you must allow the copyright owner to include your modifications.

Two consequences:

- The copyright holder is **Sony Group Corporation**, not the PR author, so the author cannot unilaterally relicense the port — even as first author of the paper.
- `kornia/feature/sandesc/sandesc.py:1` and `kornia/feature/sandesc/_modules.py:1` carry `Copyright 2018 Kornia Team` under Apache-2.0, with **no mention of Sony and no attribution to the reference implementation anywhere in the shipped source**. The provenance currently lives only in the PR description, which does not ship with the package. `AGENTS.md` requires a new algorithm to name its defining source in the code.

To unblock, one of these:

- **(a)** Written permission from Sony to release the kornia port under Apache-2.0, linked in the PR, plus an attribution header in both new files naming the paper and the reference implementation; or
- **(b)** A clear, explicit statement that `kornia/feature/sandesc/` is a clean-room reimplementation written from the *paper only* — no code derived from the Sony-licensed repo — with the file headers still naming the paper as the defining source. Given the PR text currently says the network came from that repo, this needs to be stated deliberately rather than assumed.

I am not a lawyer and this is not a legal opinion; it is a flag that the stated licensing position does not match what the upstream repo publishes, and kornia needs that settled before a `kornia.feature.SANDesc` exists.

### 2. Non-commercial weights become the default `from_pretrained`, and get baked into the Apache-2.0 docs build

The checkpoint's own model card confirms the restriction:

```
$ curl -s https://huggingface.co/api/models/mattia-durso/SANDesc | jq .cardData
{"license":"other","license_name":"sony-research-nc","license_link":"LICENSE"}
```

kornia has no precedent for this — SANDesc would be the first:

```
$ grep -rni "non-commercial|noncommercial|cc-by-nc|research-nc" kornia/ docs/
kornia/feature/sandesc/sandesc.py:228:  The pretrained checkpoints are released under a non-commercial license, ...
```

The note at `sandesc.py:227` is good and I would keep it, but it does not cover two places where kornia itself consumes the weights:

- `docs/generate_examples.py:739` calls `SANDesc.from_pretrained()` at docs-build time, so the published `_static/img/SANDesc.png` in kornia's Apache-2.0 documentation is a derived output of a non-commercial model.
- `.github/download-models-weights.py:93` pulls the checkpoint into CI.

This is a maintainer policy call rather than a code defect, so I am flagging it for an explicit decision: is kornia willing to ship a `from_pretrained` whose default weights forbid commercial use, and to render docs images from them? If yes, the note should also appear in the **class** docstring (not only `from_pretrained`), since that is what renders on the API page.

### 3. Single-source, non-content-addressed weight hosting — and this PR made it worse

`kornia/feature/sandesc/sandesc.py:35` is now a one-element list pointing at a Nextcloud public share token:

```python
urls: list[str] = [
    "https://cloud.tugraz.at/index.php/s/dBiF999GBMoRg8w/download/sandesc_aliked.pth",
]
```

Commit `8a7bb0e0` ("serve SANDesc weights from a single TU Graz source") **removed** the HuggingFace and GitHub mirrors that the first version had:

```
- "https://huggingface.co/mattia-durso/SANDesc/resolve/main/pretrained/aliked/sandesc_aliked.pth",
- "https://raw.githubusercontent.com/mattiadurso/SANDesc/main/pretrained/aliked/sandesc_aliked.pth",
```

The URL does work today (I downloaded it: 9 691 255 bytes, `sha256 0805481bd7b40672ee5fe07343904c45937f4cadc28ff52a01b1dd401d584808`), but a Nextcloud share token is revocable, rotatable and expirable, it is not content-addressed, and it is now the **only** source for `from_pretrained`, the docs build and the CI weights cache. `.github/download-models-weights.py:93` registers it as a bare string, so it has no fallback either. Every other model in the tree has a mirror list (DISK, LoFTR, SOLD2, DeDoDe, HardNet).

Please:

- Mirror the checkpoint under the `kornia` HF org and make that the primary via `hf_url("sandesc", ...)`. It is not there yet — I checked all 43 repos under `author=kornia` and none match. You already offered this in the PR description; a maintainer needs to action it. (Note this interacts with blocker 2: mirroring NC weights into the kornia org is itself the policy decision.)
- Restore the other sources as fallbacks rather than deleting them.
- Consider `check_hash=True` with the hash in the filename — `load_state_dict_from_url` documents that it supports this and deliberately keeps the primary URL's filename across fallbacks so the hash stays consistent.

---

## Non-blocking nits

1. **PR title.** "Sandesc" is not a conventional-commit subject. Suggest `feat(feature): add SANDesc dense local descriptor`.

2. **The PR description is stale relative to the head.** It still documents `SANDesc.from_pretrained(detector="aliked"|"dedode")`, two checkpoints, and `SANDesc.forward(img, pad_if_not_divisible=False)` returning a dense volume. At `c6485227` there is one checkpoint, no `detector` argument, and `forward` is detect-and-describe:
   ```
   from_pretrained params: ['url', 'amp', 'amp_dtype', 'load_detector', 'num_keypoints']
   ```
   Worth refreshing before merge so the description matches what reviewers and users read.

3. **`describe()` does not preserve half/bfloat16 output dtype.** The upcast at `sandesc.py:377` is correct (`grid_sample` has no half kernel) but the result is never cast back:
   ```
   dtype torch.float64:  OK -> (1, 4, 128) torch.float64
   dtype torch.float16:  OK -> (1, 4, 128) torch.float32
   dtype torch.bfloat16: OK -> (1, 4, 128) torch.float32
   ```
   `AGENTS.md` asks that dtype be preserved. A `.to(volume.dtype)` on the way out would be a one-liner.

4. **`mode="nearest"` default diverges from the sibling, and zeroes keypoint gradients.** DeDoDe's `describe` uses `mode="bilinear"` (`kornia/feature/dedode/dedode.py:249`). I measured the practical impact rather than assuming it matters, and it is small — the volume is smooth:
   ```
   ALIKED keypoints are sub-pixel: 100.0% have a non-integer coordinate (max frac offset 0.500 px)
   cosine(nearest, bilinear) per keypoint: mean 0.9998  min 0.9923  |  0.0% below 0.99
   ```
   So this is not a correctness blocker. But `nearest` does discard the sub-pixel localisation ALIKED works to produce, and it makes descriptors non-differentiable w.r.t. keypoints:
   ```
   grad wrt image           : 76.49
   grad wrt kpts            : 0.0     (mode='nearest' default)
   grad wrt kpts (bilinear) : 62.79
   ```
   If `nearest` is what the paper/reference implementation uses, please say so in the docstring; otherwise `bilinear` is the more conventional default here.

5. **`from_pretrained` ignores the checkpoint's own `config`.** The downloaded file has top-level keys `['state_dict', 'config']`, but the architecture flags are hardcoded at `sandesc.py:250`. Reading `config` (or at least asserting it matches) would make a future checkpoint revision fail loudly instead of silently loading into the wrong architecture.

---

## Duplication disposition: **KEEP SEPARATE**

Per the standing duplication-audit protocol, checked against what kornia already has:

- **HardNet / HyNet / SOSNet / TFeat** are *patch* descriptors — they consume normalised patches cropped from LAFs, not an image.
- **DISK / ALIKED / XFeat** are joint detect-and-describe networks; their descriptors are structurally tied to their own detector head.
- **LightGlue** is a matcher, not a descriptor.
- **DeDoDe** is the closest — it does have a descriptor-only `describe()` on a dense volume — but its descriptors are tied to the DeDoDe B/G backbones and it is not offered as a drop-in replacement for another detector's descriptors.

Nothing in kornia currently covers "keep your existing detector's keypoints, swap in a better descriptor". SANDesc is a published (3DV 2026), detector-agnostic, descriptor-only dense network, and the `ALIKED(disable_descriptors=True)` plumbing in this PR is exactly the composition that no existing class provides. **KEEP SEPARATE** — this is additive, not a re-implementation of an existing kornia capability.

---

## What I could not verify

- **CUDA, MPS, float16/bfloat16 on GPU, and TF32 behaviour.** Everything above is CPU-only. The PR reports a green RTX 4090 run; I did not reproduce it.
- **Matching quality / the paper's accuracy claims.** I verified shapes, normalisation, sampling convention and numerical stability against the shipped reference values — not that SANDesc actually beats ALIKED descriptors. The `.. note::` at `sandesc.py:76` is appropriately honest that the numbers come from the paper and the module has not been benchmarked inside kornia; please keep that note.
- **The licence question is a flag, not a legal conclusion** — see blocker 1.

Thanks for the depth of work here, especially the test coverage and the docs-example branch. The engineering is close to mergeable; it is the provenance and hosting that need a maintainer decision before this can land.

ducha-aiki:
@mattiadurso 
OK, now my human review. 

0) Opus is correct on the descriptor code, we need 

>    A clear, explicit statement that kornia/feature/sandesc/ is a clean-room reimplementation written from the paper only — no code derived from the Sony-licensed repo — with the file headers still naming the paper as the defining source. Given the PR text currently says the network came from that repo, this needs to be stated deliberately rather than assumed.

1) For the weights - it would make sense to upload them to HF under Apache 2 license with clear explanation where they come from, so we pull from it and mirror as kornia, but you also own them as author. That's simple part. 

2)  don't add viz to the keyponts docs -- sandesc is not a detector

3) I agree with Opus review about not touching aliked -- users can always remove the descriptors, and their extraction cost almost no extra from aliked. 

The rest is good and thank you for the contribution!

mattiadurso:
0. I'll check this internally and hopefully get a green light in the next weeks.
1.  Should I upload them under https://huggingface.co/mattia-durso/SANDesc or using Kornia account? In the second case, I would need one of the maintainers do that I guess.
2.  ok
3. Then I remove all ALIKED part and take keypoints in forward eg
`
keypoints,_ = aliked(images)
`
`
descriptors = sandesc(images, keypoints)
`

ducha-aiki:
1. Upload under  https://huggingface.co/mattia-durso/SANDesc  and I will download and upload to kornia. I would also keep first link to your account - it costs you nothing, but gives the download numbers, and if they are big, that's something :)

ducha-aiki:
## Status note at `8192ab6b` — items 2 and 3 are done; the licence gate is the only thing left

@mattiadurso — you pushed `b310b395` + `8192ab6b` forty minutes after the last exchange and nobody came back to
you, so let me close that loop. Both of the items you said "ok" to are fully delivered:

- **(2) no keypoint visualisation** — `docs/generate_examples.py` is no longer in the PR's diff at all.
- **(3) ALIKED untouched** — `kornia/feature/aliked/aliked.py` and `tests/feature/test_aliked.py` are out of the
  diff too, and `SANDesc` now takes `(images, keypoints)` as you proposed.

The PR against its merge-base is now **purely additive** — 10 files, `+1016 / -0`, all of it SANDesc plus its
registration, docs entry, bib entry and weights-prefetch hookup. That is a much easier thing to merge than what I
reviewed on the 28th, and the `align_corners` fix I verified end-to-end still stands.

**What is left is item 0/1, which is on your side and unchanged:** the explicit clean-room statement for
`kornia/feature/sandesc/`, and the weights uploaded to `https://huggingface.co/mattia-durso/SANDesc` under
Apache 2 so @ducha-aiki can mirror them to the kornia org. You said you would check the first internally over the
coming weeks; no rush from us, and nothing in the code review is blocking behind it.

**Two housekeeping things whenever you next touch the branch:**

1. The PR is **`CONFLICTING`** against current `main`. Only two files, both trivial: `.gitignore` and
   `docs/source/feature.rst`. Everything else (`kornia/feature/__init__.py`, `tests/core/test_weights_prefetch.py`)
   auto-merges.
2. Because it is conflicting, **GitHub is running no `pull_request` workflow on it** — the 34 green checks you see
   are from the 08-28 head, before `main` moved, and they are not evidence about the current state. A rebase gets
   you a real CI answer.

ducha-aiki:
To be explicit, since my previous note may have read as softer than intended: **the licence provenance is a hard
blocker, not a nice-to-have.** Nothing else on this PR can move it forward.

Concretely, before it can be merged we need both of:

1. **A statement, in the PR, that `kornia/feature/sandesc/` is a clean-room reimplementation written from the
   paper alone** — no code derived from the Sony-licensed repository. This needs to be stated deliberately rather
   than inferred, because the PR description currently says the network came *from* that repo. The file headers
   should keep naming the paper as the defining source.
2. **The weights published by you at `https://huggingface.co/mattia-durso/SANDesc` under Apache 2.0**, with their
   provenance written on the model card, so @ducha-aiki can mirror them into the kornia org. Keeping your own
   copy as the first link costs nothing and keeps the download numbers attributed to you.

kornia ships under Apache 2.0 and cannot take code or weights whose licence is unresolved, however good the
implementation is — and this one is good: the `align_corners` fix is verified end to end, and stripping the ALIKED
changes made the PR purely additive.

No deadline pressure from us. Take the weeks you said you needed for the internal check; the review will be
waiting. But please do not spend effort on further code changes until item 1 comes back, because if the answer is
"no", none of that work can land.

ducha-aiki (changes requested):
Setting the formal review state to match the existing review: **changes requested** on head `8192ab6b`, which is unchanged since that review.

The outstanding items are in https://github.com/kornia/kornia/pull/3935#issuecomment-5545607256 (licence provenance hard gate: clean-room statement and Apache-2.0 weights provenance).

victorwon2001:
Static review of all 10 supplied changed files found two current, previously unraised issues in kornia/feature/sandesc/sandesc.py: the public channel-configuration documentation incorrectly instructs users to add an extra descriptor channel, and the documented/tested claim that all out-of-bounds keypoints yield zero descriptors is false for near-boundary coordinates under the default bilinear sampling path. The already-discussed licence-provenance gate remains outstanding and is not duplicated as a finding.

**kornia/feature/sandesc/sandesc.py:68** — The public constructor documentation says to add +1 to the last up_output_channels element to match the DISK U-Net, even though that element is used directly as SANDesc's descriptor-channel count with no extra channel removed.

A user following the documented example [64, 64, 64, 128 + 1] constructs a 129-channel descriptor volume rather than a 128-D descriptor. This can silently create an unexpected descriptor dimension and make custom configurations incompatible with consumers expecting the documented 128-D SANDesc representation.

Static trace: the documentation identifies the last up_output_channels value as the descriptor dimension but recommends adding one; self.up3 receives up_output_channels[3] directly as ch_out, and extract_dense_map returns all x8 channels without slicing a separate score channel. No runtime test was performed.

**kornia/feature/sandesc/sandesc.py:291** — The forward API states that keypoints outside the image return zero descriptors, and the test treats zero norm as the caller-visible invalid-keypoint signal, but the default mode is bilinear and the implementation does not mask out-of-bounds coordinates before grid_sample. A coordinate only fractionally outside an edge still interpolates the boundary pixel with zero padding and therefore is not generally zero.

Near-boundary invalid keypoints can produce finite nonzero normalized descriptors instead of the documented zero descriptor, so downstream code using zero norm to recognize invalid points can retain and match points that are outside the image. The current test covers only far-outside coordinates and does not pin this boundary case.

Static sampling analysis: pixel coordinates are normalized with the align_corners=True convention, then passed directly to F.grid_sample with mode="bilinear" by default and no explicit bounds mask. For example, an x coordinate slightly below 0 maps slightly below -1 and bilinear sampling mixes the boundary pixel with zero padding rather than returning an all-zero vector. No runtime test was performed.

`file:kornia/feature/sandesc/sandesc.py:head`

> down_output_channels: Output channels of each down block, 5 elements.
>         up_output_channels: Output channels of each up block, 4 elements. The last element is the
>             descriptor dimension. Add +1 to the last element to match the DISK unet,
>             e.g. ``[64, 64, 64, 128 + 1]``.

`file:kornia/feature/sandesc/sandesc.py:head`

> self.up3 = UNetUpBlock(
>             up_output_channels[2],
>             down_output_channels[-5],
>             up_output_channels[3],
>             kernel_size=kernel_size,
>             activation=None,
>             norm=None,
>         )

`file:kornia/feature/sandesc/sandesc.py:head`

> keypoints: Keypoints in pixel coordinates of shape :math:`(B, N, 2)` as ``[x, y]``,
>                 e.g. the output of a kornia detector such as ALIKED or DISK. Keypoints outside
>                 the image sample the zero padding and come back as zero descriptors.
>             mode: ``grid_sample`` interpolation mode, ``"bilinear"`` (default) or ``"nearest"``.
>             normalize: If True (default), L2-normalize the descriptors after sampling.

`file:kornia/feature/sandesc/sandesc.py:head`

> KORNIA_CHECK_SHAPE(keypoints, ["B", "N", "2"])
>         volume = self.extract_dense_map(images, pad_if_not_divisible=pad_if_not_divisible)
>         height, width = images.shape[-2:]
>         # normalize_pixel_coordinates divides by (side - 1), which is the
>         # ``align_corners=True`` convention grid_sample is called with below.
>         grid = normalize_pixel_coordinates(keypoints, height, width)[:, None]
>         # grid_sample does not support half/bfloat16 (autocast) volumes; upcast those to
>         # float32 and match the grid dtype to the volume to avoid a dtype mismatch.
>         sample_volume = volume.float() if volume.dtype in (torch.float16, torch.bfloat16) else volume
>         grid = grid.to(device=sample_volume.device, dtype=sample_volume.dtype)
>         sampled = F.grid_sample(sample_volume, grid, mode=mode, align_corners=True)
>         descriptors = sampled[:, :, 0].mT  # B,N,des_dim

`file:tests/feature/test_sandesc.py:head`

> def test_out_of_bounds_keypoints(self, device, dtype):
>         """Keypoints outside the image sample the zero padding: zero descriptors, never NaN.
> 
>         The zero norm is the signal the caller has to notice -- it must not become NaN,
>         which would only surface much later in a matching pipeline.
>         """
>         model = SANDesc().to(device, dtype).eval()
>         images = _ramp_image(1, 3, 32, 32, device=device, dtype=dtype)
>         keypoints = torch.tensor([[[100.0, 100.0], [-50.0, 0.0], [16.0, 16.0]]], device=device, dtype=dtype)
>         descriptors = model(images, keypoints)
>         assert torch.isfinite(descriptors).all()
>         norms = descriptors.norm(dim=-1)[0]
>         self.assert_close(norms[:2], torch.zeros_like(norms[:2]))
>         self.assert_close(norms[2], torch.ones_like(norms[2]))

`comment:5545607256`

> Concretely, before it can be merged we need both of:
> 
> 1. **A statement, in the PR, that `kornia/feature/sandesc/` is a clean-room reimplementation written from the
>    paper alone** — no code derived from the Sony-licensed repository. This needs to be stated deliberately rather
>    than inferred, because the PR description currently says the network came *from* that repo. The file headers
>    should keep naming the paper as the defining source.
> 2. **The weights published by you at `https://huggingface.co/mattia-durso/SANDesc` under Apache 2.0**, with their
>    provenance written on the model card, so @ducha-aiki can mirror them into the kornia org. Keeping your own
>    copy as the first link costs nothing and keeps the download numbers attributed to you.
> 
> kornia ships under Apache 2.0 and cannot take code or weights whose licence is unresolved, however good the
> implementation is — and this one is good: the `align_corners` fix is verified end to end, and stripping the ALIKED
> changes made the PR purely additive.
> 
> No deadline pressure from us. Take the weeks you said you needed for the internal check; the review will be
> waiting. But please do not spend effort on further code changes until item 1 comes back, because if the answer is
> "no", none of that work can land.

Resolve the maintainer-requested clean-room and Apache-2.0 weights provenance gate first. Once that external gate is settled, correct the up_output_channels documentation and define/test near-boundary out-of-bounds keypoint behavior explicitly.

SHA: `8192ab6bbaaf4b56fb94e5a07c44280c28379d13`

ducha-aiki:
@victorwon2001 please do NOT go over other people’s PRs and reviews. That’s rude to start with, and you are not a maintainer.

Issue #3872, opened 2026-07-28
Batched letterbox (aspect-ratio preserving pad) transform


RichardHtunn:
### 🚀 Feature Description

A native, GPU-accelerated letterbox transform (resizing + symmetric padding to strictly preserve aspect ratio) that operates directly on batched 4D tensors `[B, C, H, W]`.

### 📂 Feature Category

Geometry

### 💡 Motivation

Currently, if users want to resize and pad images to a square or specific rectangle while strictly preserving aspect ratio ("letterboxing"), they must manually calculate dynamic padding margins. 

When doing on-the-fly image preprocessing for GPU-resident training batches (e.g., Object Detection or OCR), falling back to CPU-bound loops using libraries like `cv2` creates a severe execution bottleneck. Having a native tensor operation in Kornia solves this problem directly on CUDA.

### 💭 Proposed Solution

Add a tensor-native transform in `kornia.geometry.transform` (e.g. `letterbox` or `pad_to_aspect_ratio`).

**Proposed API:**
```python
import torch
import kornia.geometry.transform as K

# x: [B, C, H, W] tensor
# target_size: (target_h, target_w)
output = K.letterbox(x, target_size=(512, 512), pad_value=114)

### 🔄 Alternatives Considered

- **OpenCV / PIL on CPU:** Looping over `cv2.resize` and `cv2.copyMakeBorder` on CPU, which breaks GPU-resident training workflows.
- **Manual dynamic padding:** Requiring users to compute padding margins manually in PyTorch and pass them to `kornia.geometry.transform.pad`, which adds boilerplate and risk of edge-case bugs (e.g., odd/even pixel distribution).

### 🎯 Use Cases

- **OCR Pipelines:** Where aspect ratio distortion severely reduces character recognition accuracy.
- **Object Detection (e.g., YOLO):** Standardizing input tensor dimensions without stretching object aspect ratios.
- **GPU-Resident Pipelines:** Accelerating on-the-fly batched tensor preprocessing directly on CUDA.

### 📝 Additional Context

I built and tested a standalone version for an OCR model training pipeline. On an Nvidia T4 GPU (Batch Size = 32, 1080p -> 512x512 target), native batched GPU letterboxing yielded a ~7.8x throughput speedup compared to sequential `cv2` calls (~10,600 FPS vs ~1,350 FPS).

I have unit tests handling edge cases (even vs. odd padding, rectangular targets) and am ready to adapt this to Kornia's codebase conventions.

[...]

PR #3936, opened 2026-08-10
feat: add pure-torch letterbox transform

RichardHtunn:
- [ ] 🟢 **Green lane** — test-first bug fix / verified docs fix / [`help wanted`](https://github.com/kornia/kornia/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22help%20wanted%22) issue / benchmark results. No prior issue needed.
- [x] 🟠 **Discuss-first** — feature or behavior change; a maintainer confirmed the scope on the linked issue.

**Fixes/Relates to:** # *(issue number, if any)*

## Test log

```
$ pixi run test tests/...
```

- [ ] 🟢 **Human-written**

### DIFF
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -806,19 +806,16 @@ def letterbox(
     h, w = input.shape[-2:]
     new_h, new_w = new_shape
 
-    # 1. Calculate the scale ratio
     ratio = min(new_h / h, new_w / w)
 
-    # 2. Compute new unpadded dimensions
-    unpad_h = int(round(h * ratio))
-    unpad_w = int(round(w * ratio))
+    # Use max(1, ...) to prevent 0-pixel dimensions on extreme aspect ratios
+    unpad_h = max(1, round(h * ratio))
+    unpad_w = max(1, round(w * ratio))
 
-    # 3. Resize the image (delegating to Kornia's internal resize function)
     resized_image = resize(
         input, size=(unpad_h, unpad_w), interpolation=interpolation, align_corners=align_corners, antialias=antialias
     )
 
-    # 4. Calculate padding for center alignment
     pad_h = new_h - unpad_h
     pad_w = new_w - unpad_w
 
@@ -827,7 +824,6 @@ def letterbox(
     pad_left = pad_w // 2
     pad_right = pad_w - pad_left
 
-    # 5. F.pad expects padding from last dimension backwards: (left, right, top, bottom)
     padded_image = torch.nn.functional.pad(
         resized_image,
         (pad_left, pad_right, pad_top, pad_bottom),
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -768,11 +768,11 @@ def rescale(
 def letterbox(
     input: torch.Tensor,
     new_shape: Tuple[int, int],
-    fill_value: float = 114.0,
+    fill_value: float = 0.0,
     interpolation: str = "bilinear",
     align_corners: Optional[bool] = None,
     antialias: bool = False,
-) -> torch.Tensor:
+) -> Tuple[torch.Tensor, dict]:
     r"""Resizes and pads an image tensor to a target shape while maintaining aspect ratio.
 
     Convention:
@@ -782,26 +782,32 @@ def letterbox(
     Args:
         input: The image tensor to be resized and padded with shape of :math:`(..., H, W)`.
         new_shape: Target spatial dimensions as ``(height, width)``.
-        fill_value: Value used for padding. Default is 114.0 (common for object detection).
+        fill_value: Value used for padding. Default is 0.0 (Kornia expects tensors in [0, 1] range).
         interpolation: algorithm used for upsampling: ``'nearest'`` | ``'linear'`` | ``'bilinear'`` |
             ``'bicubic'`` | ``'trilinear'`` | ``'area'``.
         align_corners: interpolation flag.
         antialias: if True, then image will be filtered with Gaussian before downscaling.
 
     Returns:
-        The padded and resized tensor with the shape of ``new_shape``.
+        A tuple containing the padded and resized tensor with the shape of ``new_shape``, 
+        and a dictionary with the scale ``ratio`` and ``padding`` metadata.
 
     Example:
         >>> img = torch.ones(1, 3, 100, 200)
-        >>> out = letterbox(img, (200, 200))
+        >>> out, meta = letterbox(img, (200, 200))
         >>> print(out.shape)
         torch.Size([1, 3, 200, 200])
+        >>> print(meta["ratio"])
+        1.0
     """
     if not isinstance(input, torch.Tensor):
         raise TypeError(f"Input tensor type is not a torch.Tensor. Got {type(input)}")
 
     if len(input.shape) < 2:
         raise ValueError(f"Input tensor must have at least two dimensions. Got {len(input.shape)}")
+
+    if new_shape[0] <= 0 or new_shape[1] <= 0:
+        raise ValueError(f"new_shape dimensions must be strictly positive. Got {new_shape}")
 
     h, w = input.shape[-2:]
     new_h, new_w = new_shape
@@ -813,7 +819,11 @@ def letterbox(
     unpad_w = max(1, round(w * ratio))
 
     resized_image = resize(
-        input, size=(unpad_h, unpad_w), interpolation=interpolation, align_corners=align_corners, antialias=antialias
+        input, 
+        size=(unpad_h, unpad_w), 
+        interpolation=interpolation, 
+        align_corners=align_corners, 
+        antialias=antialias
     )
 
     pad_h = new_h - unpad_h
@@ -830,7 +840,12 @@ def letterbox(
         value=fill_value,
     )
 
-    return padded_image
+    metadata = {
+        "ratio": ratio,
+        "padding": (pad_left, pad_right, pad_top, pad_bottom)
+    }
+
+    return padded_image, metadata
 
 
 class Resize(nn.Module):
@@ -1328,18 +1343,18 @@ class Letterbox(nn.Module):
 
     Args:
         new_shape: Desired output size as ``(height, width)``.
-        fill_value: Value used for padding. Default is 114.0.
+        fill_value: Value used for padding. Default is 0.0.
         interpolation: algorithm used for upsampling: ``'nearest'`` | ``'linear'`` | ``'bilinear'`` |
             'bicubic' | 'trilinear' | 'area'.
         align_corners: interpolation flag.
         antialias: if True, then image will be filtered with Gaussian before downscaling.
 
     Returns:
-        The padded and resized tensor.
+        A tuple containing the padded and resized tensor, and a dictionary with scale/pad metadata.
 
     Example:
         >>> img = torch.ones(1, 3, 100, 200)
-        >>> out = Letterbox((200, 200))(img)
+        >>> out, meta = Letterbox((200, 200))(img)
         >>> print(out.shape)
         torch.Size([1, 3, 200, 200])
     """
@@ -1347,7 +1362,7 @@ class Letterbox(nn.Module):
     def __init__(
         self,
         new_shape: Tuple[int, int],
-        fill_value: float = 114.0,
+        fill_value: float = 0.0,
         interpolation: str = "bilinear",
         align_corners: Optional[bool] = None,
         antialias: bool = False,
@@ -1359,15 +1374,14 @@ class Letterbox(nn.Module):
         self.align_corners: Optional[bool] = align_corners
         self.antialias: bool = antialias
 
-    def forward(self, input: torch.Tensor) -> torch.Tensor:
+    def forward(self, input: torch.Tensor) -> Tuple[torch.Tensor, dict]:
         """Resize and pad spatial dimensions.
 
         Args:
             input: Tensor with shape :math:`(*, C, H, W)`.
 
         Returns:
-            Tensor with the same leading and channel dimensions as ``input``.
-            The output height and width are determined by ``self.new_shape``.
+            Tuple containing the transformed tensor and a metadata dictionary.
         """
         return letterbox(
             input,
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -789,7 +789,7 @@ def letterbox(
         antialias: if True, then image will be filtered with Gaussian before downscaling.
 
     Returns:
-        A tuple containing the padded and resized tensor with the shape of ``new_shape``, 
+        A tuple containing the padded and resized tensor with the shape of ``new_shape``,
         and a dictionary with the scale ``ratio`` and ``padding`` metadata.
 
     Example:
@@ -819,11 +819,7 @@ def letterbox(
     unpad_w = max(1, round(w * ratio))
 
     resized_image = resize(
-        input, 
-        size=(unpad_h, unpad_w), 
-        interpolation=interpolation, 
-        align_corners=align_corners, 
-        antialias=antialias
+        input, size=(unpad_h, unpad_w), interpolation=interpolation, align_corners=align_corners, antialias=antialias
     )
 
     pad_h = new_h - unpad_h
@@ -840,10 +836,7 @@ def letterbox(
         value=fill_value,
     )
 
-    metadata = {
-        "ratio": ratio,
-        "padding": (pad_left, pad_right, pad_top, pad_bottom)
-    }
+    metadata = {"ratio": ratio, "padding": (pad_left, pad_right, pad_top, pad_bottom)}
 
     return padded_image, metadata
 
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -764,6 +764,7 @@ def rescale(
     size = (int(height * factor_vert), int(width * factor_horz))
     return resize(input, size, interpolation=interpolation, align_corners=align_corners, antialias=antialias)
 
+
 def letterbox(
     input: torch.Tensor,
     new_shape: Tuple[int, int],
@@ -814,11 +815,7 @@ def letterbox(
 
     # 3. Resize the image (delegating to Kornia's internal resize function)
     resized_image = resize(
-        input, 
-        size=(unpad_h, unpad_w), 
-        interpolation=interpolation, 
-        align_corners=align_corners, 
-        antialias=antialias
+        input, size=(unpad_h, unpad_w), interpolation=interpolation, align_corners=align_corners, antialias=antialias
     )
 
     # 4. Calculate padding for center alignment
@@ -1324,6 +1321,7 @@ class Shear(nn.Module):
             configured padding mode.
         """
         return shear(input, self.shear, self.mode, self.padding_mode, self.align_corners)
+
 
 class Letterbox(nn.Module):
     r"""Resizes and pads the input torch.Tensor to the given shape while maintaining aspect ratio.
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -28,6 +28,7 @@ from .imgwarp import get_affine_matrix2d, get_projective_transform, get_rotation
 
 __all__ = [
     "Affine",
+    "Letterbox",
     "Rescale",
     "Resize",
     "Rotate",
@@ -36,6 +37,7 @@ __all__ = [
     "Translate",
     "affine",
     "affine3d",
+    "letterbox",
     "rescale",
     "resize",
     "resize_to_be_divisible",
@@ -762,6 +764,81 @@ def rescale(
     size = (int(height * factor_vert), int(width * factor_horz))
     return resize(input, size, interpolation=interpolation, align_corners=align_corners, antialias=antialias)
 
+def letterbox(
+    input: torch.Tensor,
+    new_shape: Tuple[int, int],
+    fill_value: float = 114.0,
+    interpolation: str = "bilinear",
+    align_corners: Optional[bool] = None,
+    antialias: bool = False,
+) -> torch.Tensor:
+    r"""Resizes and pads an image tensor to a target shape while maintaining aspect ratio.
+
+    Convention:
+        - ``new_shape`` is ``(height, width)``
+        - align_corners: ``None`` by default (follows ``torch.nn.functional.interpolate``)
+
+    Args:
+        input: The image tensor to be resized and padded with shape of :math:`(..., H, W)`.
+        new_shape: Target spatial dimensions as ``(height, width)``.
+        fill_value: Value used for padding. Default is 114.0 (common for object detection).
+        interpolation: algorithm used for upsampling: ``'nearest'`` | ``'linear'`` | ``'bilinear'`` |
+            ``'bicubic'`` | ``'trilinear'`` | ``'area'``.
+        align_corners: interpolation flag.
+        antialias: if True, then image will be filtered with Gaussian before downscaling.
+
+    Returns:
+        The padded and resized tensor with the shape of ``new_shape``.
+
+    Example:
+        >>> img = torch.ones(1, 3, 100, 200)
+        >>> out = letterbox(img, (200, 200))
+        >>> print(out.shape)
+        torch.Size([1, 3, 200, 200])
+    """
+    if not isinstance(input, torch.Tensor):
+        raise TypeError(f"Input tensor type is not a torch.Tensor. Got {type(input)}")
+
+    if len(input.shape) < 2:
+        raise ValueError(f"Input tensor must have at least two dimensions. Got {len(input.shape)}")
+
+    h, w = input.shape[-2:]
+    new_h, new_w = new_shape
+
+    # 1. Calculate the scale ratio
+    ratio = min(new_h / h, new_w / w)
+
+    # 2. Compute new unpadded dimensions
+    unpad_h = int(round(h * ratio))
+    unpad_w = int(round(w * ratio))
+
+    # 3. Resize the image (delegating to Kornia's internal resize function)
+    resized_image = resize(
+        input, 
+        size=(unpad_h, unpad_w), 
+        interpolation=interpolation, 
+        align_corners=align_corners, 
+        antialias=antialias
+    )
+
+    # 4. Calculate padding for center alignment
+    pad_h = new_h - unpad_h
+    pad_w = new_w - unpad_w
+
+    pad_top = pad_h // 2
+    pad_bottom = pad_h - pad_top
+    pad_left = pad_w // 2
+    pad_right = pad_w - pad_left
+
+    # 5. F.pad expects padding from last dimension backwards: (left, right, top, bottom)
+    padded_image = torch.nn.functional.pad(
+        resized_image,
+        (pad_left, pad_right, pad_top, pad_bottom),
+        value=fill_value,
+    )
+
+    return padded_image
+
 
 class Resize(nn.Module):
     r"""Resize the input torch.Tensor to the given size.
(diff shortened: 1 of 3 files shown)


RichardHtunn:
Hey @ducha-aiki  and @edgarriba , Phase 1 (the pure PyTorch letterbox and docs) is pushed and ready for review whenever you have a second! Once this looks good, I'll start looking at the DLPack/zero-copy integration.

kornia/geometry/transform/affwarp.py
```diff
@@ -763,6 +765,74 @@ def rescale(
     return resize(input, size, interpolation=interpolation, align_corners=align_corners, antialias=antialias)
 
 
+def letterbox(
+    input: torch.Tensor,
+    new_shape: Tuple[int, int],
+    fill_value: float = 114.0,
```
ducha-aiki: Nope. Kornia works in [0,1], so either do 114/255, or 0

kornia/geometry/transform/affwarp.py
```diff
@@ -763,6 +765,74 @@ def rescale(
...
+        input: The image tensor to be resized and padded with shape of :math:`(..., H, W)`.
+        new_shape: Target spatial dimensions as ``(height, width)``.
+        fill_value: Value used for padding. Default is 114.0 (common for object detection).
+        interpolation: algorithm used for upsampling: ``'nearest'`` | ``'linear'`` | ``'bilinear'`` |
+            ``'bicubic'`` | ``'trilinear'`` | ``'area'``.
+        align_corners: interpolation flag.
+        antialias: if True, then image will be filtered with Gaussian before downscaling.
+
+    Returns:
+        The padded and resized tensor with the shape of ``new_shape``.
```
ducha-aiki: No scale/pad metadata returned - I believe this is wrong. Let's tell the user the transform, we we can chain better

kornia/geometry/transform/affwarp.py
```diff
@@ -763,6 +765,74 @@ def rescale(
     return resize(input, size, interpolation=interpolation, align_corners=align_corners, antialias=antialias)
 
 
+def letterbox(
+    input: torch.Tensor,
+    new_shape: Tuple[int, int],
```
ducha-aiki, line 770: What about non-positive input? Please, add checks

ducha-aiki (changes requested):
Please, check the comments and add tests.

ducha-aiki:
**CHANGES REQUESTED — the three review points from 2026-08-11 are addressed in letter but the metadata contract is not usable for round-tripping, and returning `(Tensor, dict)` from an `nn.Module` breaks `nn.Sequential` and ONNX export. Scope is fine: this op is maintainer-confirmed and is *not* expressible with existing kornia ops for non-square targets.**

Measured on head `e413e3f8b679e4e0522cc57fd4f72453fadb5d8a`, merge-base `a2745009`, torch 2.9.1, Python 3.11.14, CPU. Everything below is pasted from a real run, not read off the diff.

---

## What is already right

- The scope **was** confirmed. @ducha-aiki wrote on #3872 (`issuecomment-5231420265`): *"a pure-torch letterbox in `kornia.geometry.transform` is valuable on its own and doesn't need to wait on the rs bridge — happy to review that as a first PR."* The issue also carries `help wanted`. Discuss-first lane is satisfied. (Please put `Relates to: #3872` in the PR body — it is currently the unfilled template, so nothing links the two.)
- `fill_value` default is now `0.0` with a `[0, 1]` note in the docstring. That resolves my `114` comment.
- `new_shape` positivity check at `affwarp.py:809-810`, with a test. Resolved.
- Doctests execute and pass:
  ```
  $ pytest --doctest-modules kornia/geometry/transform/affwarp.py -q
  17 passed in 0.43s
  ```
- Docs registration is done (`docs/source/geometry.transform.rst` +2), `__all__` updated, symbol reachable as `kornia.geometry.transform.letterbox` / `.Letterbox` / `kornia.geometry.letterbox`.
- Ruff clean (`All checks passed!`, `2 files already formatted`), the whole file's tests stay green (`99 passed, 10 skipped`).
- dtype/device/batch/gradient behaviour is genuinely correct — I checked rather than assumed:
  ```
  in float32 -> out float32 ; float64 -> float64 ; float16 -> float16
  (100,200)->(200,200)  (3,100,200)->(3,200,200)  (1,3,100,200)->(1,3,200,200)  (2,4,3,100,200)->(2,4,3,200,200)
  gradcheck: True
  ```
- `torch.compile(..., fullgraph=True)` works: `fullgraph OK, out: torch.Size([1, 3, 200, 200])`.

---

## Duplication check (the part I spent the most time on)

kornia's standing rule is that a new public op overlapping an existing one has to justify itself. I wrote the composition out of existing ops and diffed the tensors.

**For a square target, `letterbox` is byte-identical to `resize(img, S, side="long")` + a centered `F.pad`** in the common case:

```python
def compose(img, S, fill=0.0):
    out = resize(img, S, side="long")
    h, w = out.shape[-2:]
    ph, pw = S - h, S - w
    return F.pad(out, (pw//2, pw-pw//2, ph//2, ph-ph//2), value=fill)
```
```
== square target: letterbox vs resize(side='long') + pad ==   (target 640x640)
  (1,3,100,200): maxdiff=0.000e+00  identical=True
  (1,3,480,640): maxdiff=0.000e+00  identical=True
  (1,3,333,777): maxdiff=0.000e+00  identical=True
  (1,3,7,13)   : maxdiff=9.815e-01  identical=False
  (1,3,201,200): maxdiff=9.878e-01  identical=False
```

The two mismatches are a **rounding disagreement**, and I think this is a real finding rather than noise:

```
  in   7x13 -> letterbox inner (345, 640), resize(side='long') inner (344, 640)
  in 201x200 -> letterbox inner (640, 637), resize(side='long') inner (640, 636)
```
`letterbox` uses `round()` (`affwarp.py:818-819`); `_side_to_image_size` (`affwarp.py:578-586`) uses `int()` truncation. So kornia would ship two aspect-preserving resizes in the *same file* that disagree by a pixel. See blocker 4.

**For a non-square target, the existing API genuinely cannot express it** — this is the real justification for the op:
```
  letterbox(img_333x777, (480,640)) -> (1,3,480,640)  ratio=0.8237
  resize(img, 640, side='short') -> (1,3,640,1493)
  resize(img, 640, side='long')  -> (1,3,274,640)
  resize(img, 640, side='vert')  -> (1,3,640,1493)
  resize(img, 640, side='horz')  -> (1,3,274,640)
```
None of the four `side` values fits into a 480x640 box.

**The augmentation route is not equivalent either.** `AugmentationSequential(LongestMaxSize(S), PadTo((S,S)))` pads bottom-right, not centered — `PadTo.apply_transform` hardcodes `[0, width_pad, 0, height_pad]`:
```
augmentation route  row-sums (first/last 3 rows): [200.0, 200.0, 200.0] [0.0, 0.0, 0.0]
letterbox           row-sums (first/last 3 rows): [0.0, 0.0, 0.0]       [0.0, 0.0, 0.0]
identical: False
```

**Disposition: KEEP SEPARATE.** The centered fit into an arbitrary `(H, W)` box is not reachable from `resize`/`PadTo`/`LongestMaxSize`, and a maintainer already asked for it on #3872. I am not asking you to fold this into `resize`. But please read blocker 4 — keeping it separate is only defensible if it does not silently contradict the resize that lives 200 lines above it.

---

## Blockers

### 1. `Letterbox.forward` returning `Tuple[Tensor, dict]` breaks composition and ONNX export — `affwarp.py:1370`

An `nn.Module` in `kornia.geometry.transform` that does not return a `Tensor` cannot be chained:

```
== nn.Sequential composability ==
  nn.Sequential(Letterbox((200,200)), GaussianBlur2d((3,3),(1.,1.)))
  FAIL: TypeCheckError Type mismatch: expected Tensor, got <class 'tuple'>.
```

And ONNX export, which the neighbouring `Resize` module supports, fails:

```
  Resize((480,640))            legacy-onnx OK
  Letterbox((480,640))         legacy-onnx FAIL: TypeError: type Tensor doesn't define __round__ method
```

That first `TypeError` is a second, separate bug (blocker 3). After I locally patched it to `h, w = int(input.shape[-2]), int(input.shape[-1])`, the export gets one step further and then dies on the dict:

```
RuntimeError: output 1 (0.8236808236808236 [ CPUDoubleType{} ]) of traced region did not have
observable data dependence with trace inputs; this probably indicates your program cannot be
understood by the tracer.
```

`torch.jit.script(Letterbox(...))` also fails with `RuntimeError: Can not create tuple with None type` (though I'll note `resize()` itself is not scriptable on torch 2.9.1 either, so JIT is not a fair bar here — `nn.Sequential` and ONNX are).

**Ask:** keep `forward` returning a plain `Tensor` and expose the metadata as a module attribute set during `forward` — this is exactly the `GeometricAugmentationBase2D.transform_matrix` pattern already used in `kornia/augmentation/`. The *functional* `letterbox()` returning a tuple is fine and is what I asked for; the *module* must stay `Tensor -> Tensor`.

### 2. The metadata is not sufficient to map coordinates back, and no convention is documented — `affwarp.py:839`

My original comment was "tell the user the transform, so we can chain better". `{"ratio": float, "padding": (l, r, t, b)}` returns the *ingredients* of a transform but not the transform, and the docstring never states how to apply them. That matters, because the obvious reading is wrong. Delta at pixel `(1,1)` of a 4x8 image, letterboxed to 32x32 with `nearest`:

```
meta: {'ratio': 4.0, 'padding': (0, 0, 8, 8)}
delta block rows: 12-15  cols: 4-7
block CENTER (row,col):            (13.5, 5.5)
naive  index map  i*r+pad       -> (12.0, 4.0)
center index map (i+.5)*r-.5+pad -> (13.5, 5.5)
```

A user doing the natural `x*ratio + pad_left` on pixel-index coordinates is off by `(ratio-1)/2` — 1.5 px here, and it grows with the scale factor. The metadata is correct only for *continuous/corner* coordinates, and nothing says so. It also disagrees with kornia's own resize transform, which is corner-index (`align_corners=True`) based:

```
LongestMaxSize(200) on 200x400: img (1,3,100,200), kp (399,199) -> (199.0, 99.0)   # i * (W'-1)/(W-1)
```

So this PR would introduce a third coordinate convention into the geometry package.

**Ask (either is fine):**
- (a) Return a `(B, 3, 3)` affine matrix alongside (or instead of) the dict. It is two lines, `get_affine_matrix2d` is *already imported* at `affwarp.py:27`, and it makes the round-trip free and unambiguous. I verified this works:
  ```
  M = [[r,0,pad_left],[0,r,pad_top],[0,0,1]]
  forward-mapped: [[0.0,103.0],[639.176,376.462],[82.368,144.184]]
  round-tripped : [[0.0,7.6e-06],[776.0,332.00003],[99.99999,50.00001]]
  round-trip exact: True
  ```
- (b) Keep the dict but add an explicit `Convention:` block giving the formula and stating that the coordinates are continuous, not pixel-index. The base commit of this PR (`a2745009`, "docs(geometry): per-op Convention blocks") is literally that campaign — please follow it.

Either way, an inverse is what makes a letterbox useful in a detection pipeline. Right now a user has to reverse-engineer it.

### 3. Shape values are used as Python numbers without casting — `affwarp.py:812`, `818-819`

`h, w = input.shape[-2:]` then `round(h * ratio)` throws under tracing:
```
TypeError: type Tensor doesn't define __round__ method
```
`resize()` avoids this because it never calls `round()`/`int()` on a shape when `size` is a tuple. Fix is `h, w = int(input.shape[-2]), int(input.shape[-1])`. (Necessary but not sufficient for ONNX — see blocker 1.)

### 4. Reconcile the rounding with `_side_to_image_size` — `affwarp.py:818-819` vs `affwarp.py:578-586`

Evidence in the duplication section above: same file, same operation, off by one pixel for `7x13` and `201x200` into 640. Please either switch `letterbox` to the existing helper (call `_side_to_image_size` for the square case / factor out a shared helper), or state in the docstring that `letterbox` rounds and `resize(size=int, side=...)` truncates, so the two are known to differ. Silently shipping both is the thing the duplication protocol exists to prevent.

### 5. Non-positive *input* is still unchecked — `affwarp.py:812`

My 2026-08-11 comment was "What about non-positive input?" — that got read as `new_shape` only. The input side still raises a bare `ZeroDivisionError`:
```
  (1, 3, 0, 10) -> ZeroDivisionError: division by zero
  (1, 3, 10, 0) -> ZeroDivisionError: division by zero
```
Also malformed `new_shape` produces confusing errors from deep inside `F.pad`:
```
  new_shape=(200.5, 200) -> TypeError: pad(): argument 'pad' failed to unpack the object at pos 3 ...
  new_shape=200          -> TypeError: 'int' object is not subscriptable
  new_shape=(200,)       -> IndexError: tuple index out of range
```
Please raise `ValueError` with the tensor shape / the offending `new_shape` in the message.

### 6. Test coverage — `tests/geometry/transform/test_affine.py:860-894`

`TestLetterbox` subclasses `BaseTester` but uses none of it. Four tests, all shape/metadata assertions:
```
$ pytest tests/geometry/transform/test_affine.py -q --device=cpu --dtype=float32,float64 -k Letterbox
8 passed, 101 deselected in 0.67s
```
Missing, per `TESTING.md` and the `TestResize` class right above it (which has both `test_gradcheck` and `test_dynamo`):
- **No `self.assert_close` anywhere** — there is no numerical value check at all, only shapes. Nothing would catch a wrong interpolation, a wrong pad location, or a flipped `(top, bottom)` split.
- **No `test_gradcheck`.** (It does pass — I ran `torch.autograd.gradcheck` by hand and got `True` — so this is cheap to add.)
- **No `test_dynamo`** with the `torch_optimizer` fixture. (Also passes today; please pin it.)
- **No output device/dtype assertion.** `test_metadata_math` takes the `dtype` fixture and then discards the tensor entirely (`_, meta = ...`).
- **The `max(1, ...)` clamp at `affwarp.py:818-819` has no test.** That line exists specifically because of the extreme-aspect-ratio report; add the `(1, 10000) -> (640, 640)` case.
- No non-square target test, no `fill_value != 0` test, no `antialias=True` test, no 3D/5D input test — all of which I had to verify by hand for this review because the suite does not.

---

## Non-blocking

- `Tuple[torch.Tensor, dict]` at `affwarp.py:775` and `:1370` — bare `dict`. If the dict survives review, annotate it (`Dict[str, Any]`, or better a small `NamedTuple`/dataclass so `meta.ratio` works and it stays introspectable).
- `Letterbox`'s class docstring has a `Returns:` section (`affwarp.py:1345-1346`); for an `nn.Module` that belongs on `forward`. `forward` already has one, so it is duplicated.
- Docstring interpolation list mixes ``` ``'bicubic'`` ``` and `'bicubic'` styles between the function and the class.
- `docs/source/geometry.transform.rst`: `.. autoclass:: Letterbox` landed between `BaseWarper` and `HomographyWarper`, which splits the warper group. Move it next to `Resize`/`Rescale`.
- #3872 quotes a 7.8x throughput win over sequential `cv2`. If you want that number to survive, a `benchmarks/` entry following `benchmarks/README.md` would be a good follow-up — not required for this PR.
- Please fill in the PR body (link `#3872`, paste the test log, tick the AI-usage line). It is still the raw template.

---

Genuinely useful work, and the batched non-square centered fit is a real gap in kornia — I am not asking you to drop it. The two things that have to change before merge are the module return type (blocker 1) and making the metadata a usable, documented transform (blocker 2); 3-6 are mechanical once those are settled.

victorwon2001:
**kornia/geometry/transform/affwarp.py:1370** — Letterbox.forward returns Tuple[Tensor, dict] instead of preserving the Tensor-to-Tensor contract expected from the neighbouring transform modules.

The module cannot be composed normally with Tensor-consuming nn.Sequential stages, and the supplied reviewer run also reports that the metadata output prevents successful ONNX tracing/export.

The supplied head diff declares forward as Tuple[torch.Tensor, dict] and directly returns letterbox(...). A reviewer reports reproducing nn.Sequential failure on head e413e3f8b679e4e0522cc57fd4f72453fadb5d8a and a subsequent ONNX failure involving the metadata output; those executions were not repeated here.

**kornia/geometry/transform/affwarp.py:812** — letterbox takes h and w directly from input.shape and feeds arithmetic derived from them through Python round(), without first converting the shape values to concrete Python integers.

Tracing/export paths where shape values become symbolic tensors fail before the resize operation, preventing the new transform from matching the export behaviour expected of geometry modules.

The supplied head computes h, w from input.shape and then calls round(h * ratio) and round(w * ratio). The reviewer reports a measured tracing failure on the current head with a Tensor __round__ TypeError; that execution was not repeated here.

**kornia/geometry/transform/affwarp.py:815** — letterbox uses round() for aspect-preserving resized dimensions while the existing resize integer-side path is reported to use truncation, producing different inner sizes for the same conceptual resize.

Users can receive one-pixel geometry differences depending on whether they use letterbox or the existing resize API, creating inconsistent behaviour inside the same transform module.

The changed code visibly uses round() for both inner dimensions. The supplied reviewer run reports concrete 7x13 and 201x200 cases where letterbox and resize(side='long') differ by one pixel and attributes the existing helper to int() truncation; the existing helper's omitted source window was not independently inspected here.

`file:kornia/geometry/transform/affwarp.py:diff`

> +    def forward(self, input: torch.Tensor) -> Tuple[torch.Tensor, dict]:
> +        """Resize and pad spatial dimensions.
> +
> +        Args:
> +            input: Tensor with shape :math:`(*, C, H, W)`.
> +
> +        Returns:
> +            Tuple containing the transformed tensor and a metadata dictionary.
> +        """
> +        return letterbox(
> +            input,
> +            self.new_shape,
> +            self.fill_value,
> +            self.interpolation,
> +            align_corners=self.align_corners,
> +            antialias=self.antialias,
> +        )

`comment:5455951029`

> ### 1. `Letterbox.forward` returning `Tuple[Tensor, dict]` breaks composition and ONNX export — `affwarp.py:1370`
> 
> An `nn.Module` in `kornia.geometry.transform` that does not return a `Tensor` cannot be chained:
> 
> ```
> == nn.Sequential composability ==
>   nn.Sequential(Letterbox((200,200)), GaussianBlur2d((3,3),(1.,1.)))
>   FAIL: TypeCheckError Type mismatch: expected Tensor, got <class 'tuple'>.
> ```
> 
> And ONNX export, which the neighbouring `Resize` module supports, fails:
> 
> ```
>   Resize((480,640))            legacy-onnx OK
>   Letterbox((480,640))         legacy-onnx FAIL: TypeError: type Tensor doesn't define __round__ method
> ```
> 
> That first `TypeError` is a second, separate bug (blocker 3). After I locally patched it to `h, w = int(input.shape[-2]), int(input.shape[-1])`, the export gets one step further and then dies on the dict:
> 
> ```
> RuntimeError: output 1 (0.8236808236808236 [ CPUDoubleType{} ]) of traced region did not have
> observable data dependence with trace inputs; this probably indicates your program cannot be
> understood by the tracer.
> ```
> 
> `torch.jit.script(Letterbox(...))` also fails with `RuntimeError: Can not create tuple with None type` (though I'll note `resize()` itself is not scriptable on torch 2.9.1 either, so JIT is not a fair bar here — `nn.Sequential` and ONNX are).
> 
> **Ask:** keep `forward` returning a plain `Tensor` and expose the metadata as a module attribute set during `forward` — this is exactly the `GeometricAugmentationBase2D.transform_matrix` pattern already used in `kornia/augmentation/`. The *functional* `letterbox()` returning a tuple is fine and is what I asked for; the *module* must stay `Tensor -> Tensor`.

`file:kornia/geometry/transform/affwarp.py:diff`

> +    Convention:
> +        - ``new_shape`` is ``(height, width)``
> +        - align_corners: ``None`` by default (follows ``torch.nn.functional.interpolate``)
> +
> +    Args:
> +        input: The image tensor to be resized and padded with shape of :math:`(..., H, W)`.
> +        new_shape: Target spatial dimensions as ``(height, width)``.
> +        fill_value: Value used for padding. Default is 0.0 (Kornia expects tensors in [0, 1] range).
> +        interpolation: algorithm used for upsampling: ``'nearest'`` | ``'linear'`` | ``'bilinear'`` |
> +            ``'bicubic'`` | ``'trilinear'`` | ``'area'``.
> +        align_corners: interpolation flag.
> +        antialias: if True, then image will be filtered with Gaussian before downscaling.
> +
> +    Returns:
> +        A tuple containing the padded and resized tensor with the shape of ``new_shape``,
> +        and a dictionary with the scale ``ratio`` and ``padding`` metadata.

`comment:5455951029`

> ### 2. The metadata is not sufficient to map coordinates back, and no convention is documented — `affwarp.py:839`
> 
> My original comment was "tell the user the transform, so we can chain better". `{"ratio": float, "padding": (l, r, t, b)}` returns the *ingredients* of a transform but not the transform, and the docstring never states how to apply them. That matters, because the obvious reading is wrong. Delta at pixel `(1,1)` of a 4x8 image, letterboxed to 32x32 with `nearest`:
> 
> ```
> meta: {'ratio': 4.0, 'padding': (0, 0, 8, 8)}
> delta block rows: 12-15  cols: 4-7
> block CENTER (row,col):            (13.5, 5.5)
> naive  index map  i*r+pad       -> (12.0, 4.0)
> center index map (i+.5)*r-.5+pad -> (13.5, 5.5)
> ```
> 
> A user doing the natural `x*ratio + pad_left` on pixel-index coordinates is off by `(ratio-1)/2` — 1.5 px here, and it grows with the scale factor. The metadata is correct only for *continuous/corner* coordinates, and nothing says so. It also disagrees with kornia's own resize transform, which is corner-index (`align_corners=True`) based:
> 
> ```
> LongestMaxSize(200) on 200x400: img (1,3,100,200), kp (399,199) -> (199.0, 99.0)   # i * (W'-1)/(W-1)
> ```
> 
> So this PR would introduce a third coordinate convention into the geometry package.
> 
> **Ask (either is fine):**
> - (a) Return a `(B, 3, 3)` affine matrix alongside (or instead of) the dict. It is two lines, `get_affine_matrix2d` is *already imported* at `affwarp.py:27`, and it makes the round-trip free and unambiguous. I verified this works:
>   ```
>   M = [[r,0,pad_left],[0,r,pad_top],[0,0,1]]
>   forward-mapped: [[0.0,103.0],[639.176,376.462],[82.368,144.184]]
>   round-tripped : [[0.0,7.6e-06],[776.0,332.00003],[99.99999,50.00001]]
>   round-trip exact: True
>   ```
> - (b) Keep the dict but add an explicit `Convention:` block giving the formula and stating that the coordinates are continuous, not pixel-index. The base commit of this PR (`a2745009`, "docs(geometry): per-op Convention blocks") is literally that campaign — please follow it.
> 
> Either way, an inverse is what makes a letterbox useful in a detection pipeline. Right now a user has to reverse-engineer it.
> 
> ### 3. Shape values are used as Python numbers without casting — `affwarp.py:812`, `818-819`
> 
> `h, w = input.shape[-2:]` then `round(h * ratio)` throws under tracing:
> ```
> TypeError: type Tensor doesn't define __round__ method
> ```
> `resize()` avoids this because it never calls `round()`/`int()` on a shape when `size` is a tuple. Fix is `h, w = int(input.shape[-2]), int(input.shape[-1])`. (Necessary but not sufficient for ONNX — see blocker 1.)

`comment:5455951029`

> ### 4. Reconcile the rounding with `_side_to_image_size` — `affwarp.py:818-819` vs `affwarp.py:578-586`
> 
> Evidence in the duplication section above: same file, same operation, off by one pixel for `7x13` and `201x200` into 640. Please either switch `letterbox` to the existing helper (call `_side_to_image_size` for the square case / factor out a shared helper), or state in the docstring that `letterbox` rounds and `resize(size=int, side=...)` truncates, so the two are known to differ. Silently shipping both is the thing the duplication protocol exists to prevent.
> 
> ### 5. Non-positive *input* is still unchecked — `affwarp.py:812`
> 
> My 2026-08-11 comment was "What about non-positive input?" — that got read as `new_shape` only. The input side still raises a bare `ZeroDivisionError`:
> ```
>   (1, 3, 0, 10) -> ZeroDivisionError: division by zero
>   (1, 3, 10, 0) -> ZeroDivisionError: division by zero
> ```
> Also malformed `new_shape` produces confusing errors from deep inside `F.pad`:
> ```
>   new_shape=(200.5, 200) -> TypeError: pad(): argument 'pad' failed to unpack the object at pos 3 ...
>   new_shape=200          -> TypeError: 'int' object is not subscriptable
>   new_shape=(200,)       -> IndexError: tuple index out of range
> ```
> Please raise `ValueError` with the tensor shape / the offending `new_shape` in the message.

`file:tests/geometry/transform/test_affine.py:diff`

> +class TestLetterbox(BaseTester):
> +    def test_smoke(self, device, dtype):
> +        img = torch.rand(1, 3, 100, 200, device=device, dtype=dtype)
> +        out, meta = kornia.geometry.transform.letterbox(img, (200, 200))
> +        assert out.shape == (1, 3, 200, 200)
> +        assert "ratio" in meta
> +        assert "padding" in meta
> +
> +    def test_metadata_math(self, device, dtype):
> +        # 100x200 image scaling to 400x400
> +        # Ratio should be min(400/100, 400/200) = 2.0
> +        # Unpadded will be 200x400.
> +        # Padding needed on height: 400 - 200 = 200 (100 top, 100 bottom)
> +        img = torch.rand(1, 3, 100, 200, device=device, dtype=dtype)
> +
> +        _, meta = kornia.geometry.transform.letterbox(img, (400, 400))
> +
> +        assert meta["ratio"] == 2.0
> +        # padding format: (left, right, top, bottom)
> +        assert meta["padding"] == (0, 0, 100, 100)
> +
> +    def test_non_positive_shape(self, device, dtype):
> +        img = torch.rand(1, 3, 100, 200, device=device, dtype=dtype)
> +        with pytest.raises(ValueError):
> +            kornia.geometry.transform.letterbox(img, (0, 200))
> +        with pytest.raises(ValueError):
> +            kornia.geometry.transform.letterbox(img, (200, -5))
> +
> +    def test_module(self, device, dtype):
> +        img = torch.rand(1, 3, 100, 200, device=device, dtype=dtype)
> +        transform = kornia.geometry.transform.Letterbox((200, 200))
> +
> +        out, _ = transform(img)
> +
> +        assert out.shape == (1, 3, 200, 200)

`comment:5455951029`

> ### 6. Test coverage — `tests/geometry/transform/test_affine.py:860-894`
> 
> `TestLetterbox` subclasses `BaseTester` but uses none of it. Four tests, all shape/metadata assertions:
> ```
> $ pytest tests/geometry/transform/test_affine.py -q --device=cpu --dtype=float32,float64 -k Letterbox
> 8 passed, 101 deselected in 0.67s
> ```
> Missing, per `TESTING.md` and the `TestResize` class right above it (which has both `test_gradcheck` and `test_dynamo`):
> - **No `self.assert_close` anywhere** — there is no numerical value check at all, only shapes. Nothing would catch a wrong interpolation, a wrong pad location, or a flipped `(top, bottom)` split.
> - **No `test_gradcheck`.** (It does pass — I ran `torch.autograd.gradcheck` by hand and got `True` — so this is cheap to add.)
> - **No `test_dynamo`** with the `torch_optimizer` fixture. (Also passes today; please pin it.)
> - **No output device/dtype assertion.** `test_metadata_math` takes the `dtype` fixture and then discards the tensor entirely (`_, meta = ...`).
> - **The `max(1, ...)` clamp at `affwarp.py:818-819` has no test.** That line exists specifically because of the extreme-aspect-ratio report; add the `(1, 10000) -> (640, 640)` case.
> - No non-square target test, no `fill_value != 0` test, no `antialias=True` test, no 3D/5D input test — all of which I had to verify by hand for this review because the suite does not.

`github_actions_job:105153013737:annotation:message`

> Process completed with exit code 1.
> no changed changelog.d/3936.(added|fixed|breaking).md fragment found; add one for a user-visible change or apply the no-changelog label for an internal or release PR

head_content truncated at 50000 characters: kornia/geometry/transform/affwarp.py

Revise Letterbox.forward to preserve a Tensor return contract, make the functional metadata an explicit or fully documented coordinate transform, validate and concretize spatial sizes before arithmetic, reconcile letterbox sizing with existing resize rounding semantics, add numerical/edge/grad/dynamo coverage including the extreme-aspect-ratio regression, add the required changelog fragment, and rerun the relevant checks.

SHA: `e413e3f8b679e4e0522cc57fd4f72453fadb5d8a`

PR #3942, opened 2026-08-10
docs(geometry): per-op Convention blocks for conversions (batch 3a: scalar, point & pixel-coordinate conversions)

ducha-aiki:
## What

Adds a short, execution-verified `Convention:` block to the docstring of the 16 scalar / point / pixel-coordinate symbols of `kornia/geometry/conversions.py`, plus `test_convention_*` pins that fail if any documented convention changes silently. This is sub-batch **3a** of the tracking issue #3941, and continues the same programme as #3924 (batch 1, `geometry.transform` warps) and #3926 (batch 2, crops/flips/pyramid/TPS/elastic). Sub-batches 3b (15 rotation-representation symbols) and 3c (14 homography-normalization & camera-frame symbols) follow as separate PRs.

**No behavior change.** Docstrings, tests and `.rst` only — verified by tokenizing `kornia/geometry/conversions.py` before and after: the two token streams are identical once docstring tokens are excluded.

## Lane

🟢 **Green lane** — documentation fixes plus tests, no issue or assignment needed. Per CONTRIBUTING's green-lane requirement for documentation fixes, every corrected claim below was verified by running it, and the verification snippets live in the test file as pins (each with a generation-snippet comment, per CLAUDE.md's hardcoded-reference-values pattern). Nothing here changes behavior, defaults or conventions: the behavior defects the audit turned up were routed to dedicated issues (#3937–#3940) rather than fixed here, which keeps this PR in the green lane and leaves the repair conversation where it belongs.

## Scope — 16 symbols

| Symbol | Convention block | `.. warning::` | pins |
|---|---|---|---|
| `rad2deg` | ✅ | #3937 | 2 (1 xfail, 1 wart) |
| `deg2rad` | ✅ | #3937 | 2 (1 xfail, 1 wart) |
| `pol2cart` | ✅ | — | 1 |
| `cart2pol` | ✅ | #3939 | 3 (1 xfail) + 1 wart |
| `angle_to_rotation_matrix` | ✅ | #3937 | 3 (1 xfail, 1 wart) |
| `convert_points_from_homogeneous` | ✅ | #3938 | 1 (1 xfail) + 1 wart + annotated pre-existing `test_points` |
| `convert_points_to_homogeneous` | ✅ | — | annotated pre-existing `test_convert_points` |
| `convert_affinematrix_to_homography` | ✅ | — | strengthened pre-existing `test_convert_points` |
| `convert_affinematrix_to_homography3d` | ✅ | — | 1 |
[...]

PR #3945, opened 2026-08-10
fix(geometry): honor align_corners in coordinate normalization (#3904…

### LINKED ISSUE #3904: warp_perspective ignores align_corners=False in homography normalization (identity warp fails)
Found while validating PR #3650. `warp_perspective` with an IDENTITY homography does not reproduce the input when `align_corners=False`, while `warp_affine` does:

```python
import torch
from kornia.geometry.transform import get_perspective_transform
from kornia.geometry.transform.imgwarp import warp_affine, warp_perspective

img = torch.arange(16.0).view(1, 1, 4, 4)
pts = torch.tensor([[[0.0, 0.0], [3.0, 0.0], [3.0, 3.0], [0.0, 3.0]]])
M = get_perspective_transform(pts, pts)  # identity

warp_affine(img, M[:, :2, :], (4, 4), align_corners=False)       # == img  (correct)
warp_perspective(img, M, (4, 4), align_corners=False)            # != img  (wrong)
# align_corners=True: both correct
```

Root cause: `normalize_homography` maps `[0, size-1]` (pixel centers at corners) to `[-1, 1]` unconditionally — the align_corners=True convention. Under `align_corners=False`, `grid_sample` treats ±1 as pixel EDGES, so the normalization needs the `[-0.5, size-0.5] → [-1, 1]` mapping; without it every `warp_perspective(..., align_corners=False)` samples with a subpixel scale/shift error (~half-pixel at the borders, grows with output size).

Impact: any `warp_perspective`/`homography_warp` call with `align_corners=False`; also blocks the otherwise-correct fix in #3650 (crop_and_resize for non-parallel quads), which switches crops from warp_affine to warp_perspective and inherits this error for the `align_corners=False` path — that is exactly the failure `TestCropAndResize::test_align_corners_false` catches (1.55 abs diff).

Candidate fix: make `normal_transform_pixel` (and its callers in `normalize_homography`/`warp_perspective`) respect `align_corners`, mirroring what warp_affine's path does. Needs a careful sweep since `normalize_homography` is public API with documented semantics (pixel → [-1,1]).

After fixing, add the pair `warp_affine`/`warp_perspective` identity-consistency check to the conventions page's align_corners section.

### LINKED ISSUE #3928: warp_image_tps: default align_corners=False breaks even identity warps (create_meshgrid hardcodes the True convention)
Found during the #3926 convention audit (execution-verified, torch 2.9.1):

`warp_image_tps` defaults to `align_corners=False`, but its internal grid comes from `create_meshgrid(..., normalized_coordinates=True)`, which maps pixel 0 → −1 and pixel W−1 → +1 — the `align_corners=True` convention. The default therefore samples with a mismatched grid and even a mathematically-identity TPS transform corrupts the image:

```python
img = torch.arange(16.0).view(1, 1, 4, 4)
src = torch.tensor([[[-1., -1.], [-1., 1.], [1., -1.], [1., 1.], [0., 0.]]])
kernel, affine = get_tps_transform(src, src)
out = warp_image_tps(img, src, kernel, affine)           # default: max |out - img| ≈ 0.6
out = warp_image_tps(img, src, kernel, affine, align_corners=True)  # ≈ 2e-7
```

Candidate fixes: default flip (breaking, needs deprecation), or thread `align_corners` into the meshgrid normalization so both settings self-consist. Documented as a `.. warning::` in #3926 pending this fix. Split out of the closed umbrella #3923 per review feedback.

—
*Issue authored by Claude (Claude Code) on behalf of @ducha-aiki.*

### LINKED ISSUE #3929: crop_by_transform_mat: 1-pixel output sizes return NaN (align_corners=True) or silently drop the projective row (align_corners=False)
Found during #3926 review wave 4 (execution-verified, torch 2.9.1, on main after #3650):

With `out_size` containing a 1, the `(B, 3, 3)` homogeneous path degenerates two different ways:

```python
x = torch.arange(25.0).view(1, 1, 5, 5)
m3 = torch.eye(3)[None]
crop_by_transform_mat(x, m3[:, :2], (1, 4), align_corners=True)  # [[0., 1., 2., 3.]] — fine (affine path)
crop_by_transform_mat(x, m3, (1, 4), align_corners=True)         # all NaN, even for identity
```

- `align_corners=True` + 1-pixel dim: `warp_perspective` returns all-NaN.
- `align_corners=False` + 1-pixel dim: the code (crop2d.py:354-364) deliberately falls back to `warp_affine`, silently discarding the projective third row — two matrices differing only in that row produce byte-identical output.

So identical identity transforms in `(B,2,3)` vs `(B,3,3)` form behave radically differently at degenerate sizes. Disclosed in #3926's Convention block pending a runtime fix.

—
*Issue authored by Claude (Claude Code) on behalf of @ducha-aiki.*

### LINKED ISSUE #4411: W1b: an identity transform is not the identity at the class default `align_corners=False`
Extends #3904 to the augmentation surface. Three geometric augmentation classes default to `align_corners=False` and delegate to a grid-normalising warp, and at that default a transform that is mathematically the identity does not return the input. `RandomPerspective(distortion_scale=0.)` draws `start_points == end_points`, `RandomElasticTransform(alpha=(0., 0.))` applies a zero displacement and `RandomThinPlateSpline(scale=1e-06)` a near-zero one — and all three move the image. Setting `align_corners=True` on the same call makes all three exact.

The defect is not in the classes. `warp_perspective` with `torch.eye(3)` shows the same behavior on the same fixture, while `warp_affine` with an identity `(2, 3)` matrix is exact under both flag values: the homography is normalised with the corner-aligned formula regardless of the flag, which is what #3904 reports and what PR #3945 proposes to fix. The classes inherit it because `warp_perspective`, `warp_image_tps` and `elastic_transform2d` all normalise coordinates the same way, and because their own default is the affected one.

The magnitude is fixture-bound — the underlying error is a half-pixel offset of the sampling grid, so it scales with the input's value range rather than being a fixed epsilon. On the row-major ramp below, whose values run from zero to just under one, the displacement moves a pixel by most of that range. That is large enough that a "disabled" augmentation at `p=1.0` and zero magnitude is not a no-op, which is how the affected classes behave in an `AugmentationSequential` whose parameters have been annealed to zero.

## Repro

```python
from kornia.geometry.transform import warp_perspective, warp_affine
ramp = torch.arange(48., dtype=torch.float32).reshape(1, 1, 6, 8) / 48.   # 6 x 8, asymmetric

for name, make in (
    ("RandomPerspective(distortion_scale=0.)", lambda ac: K.RandomPerspective(distortion_scale=0., p=1., align_corners=ac)),
    ("RandomElasticTransform(alpha=(0., 0.))", lambda ac: K.RandomElasticTransform(alpha=(0., 0.), p=1., align_corners=ac)),
    ("RandomThinPlateSpline(scale=1e-06)   ", lambda ac: K.RandomThinPlateSpline(scale=1e-06, p=1., align_corners=ac)),
):
    out = {}
    for ac in (True, False):
        torch.manual_seed(0)
        out[ac] = round((make(ac)(ramp) - ramp).abs().max().item(), 6)
    print(f"{name}  max|out - in|:  align_corners=True {out[True]}   align_corners=False {out[False]}"
          f"   <- class default is {make(False).flags['align_corners']}")

torch.manual_seed(0)
p0 = K.RandomPerspective(distortion_scale=0., p=1.)
p0(ramp)
print("start_points == end_points for distortion_scale=0.:",
      torch.equal(p0._params["start_points"], p0._params["end_points"]))

# one level down: the same split between the two warps
H = torch.eye(3)[None]
M = torch.eye(3)[None, :2]
for ac in (True, False):
    print(f"warp_perspective(identity H, align_corners={ac}) max|out - in| ="
          f" {round((warp_perspective(ramp, H, (6, 8), align_corners=ac) - ramp).abs().max().item(), 6)}"
          f"   warp_affine(identity M) = {round((warp_affine(ramp, M, (6, 8), align_corners=ac) - ramp).abs().max().item(), 6)}")
```

## Observed

```
kornia 0.9.0rc1 torch 2.14.0
RandomPerspective(distortion_scale=0.)  max|out - in|:  align_corners=True 0.0   align_corners=False 0.734375   <- class default is False
RandomElasticTransform(alpha=(0., 0.))  max|out - in|:  align_corners=True 0.0   align_corners=False 0.734375   <- class default is False
RandomThinPlateSpline(scale=1e-06)     max|out - in|:  align_corners=True 1e-06   align_corners=False 0.734374   <- class default is False
start_points == end_points for distortion_scale=0.: True
warp_perspective(identity H, align_corners=True) max|out - in| = 0.0   warp_affine(identity M) = 0.0
warp_perspective(identity H, align_corners=False) max|out - in| = 0.734375   warp_affine(identity M) = 0.0
```

## Expected

Identity in, identity out, at both flag values — the coordinate normalisation should use the flag it is given, which is what PR #3945 implements for #3904. No augmentation-level change is needed once the warp is fixed; the three classes inherit the correction.

Until then the classes' own defaults are the reason users meet the defect without ever calling `warp_perspective`: `RandomPerspective`, `RandomElasticTransform` and `RandomThinPlateSpline` all default to the affected value while `warp_perspective`'s own default is `True`. The wider question of which default each augmentation should carry is separate and is tracked in the `align_corners` defaults issue for this batch.

## Classification

**Broken output** — a zero-magnitude transform returns a shifted image, on three public augmentation classes at their shipped defaults. Idiom: 3 (a guard whose predicate does not match the documented contract — the flag is accepted and then not consulted in the normalisation step).

## Disposition

`MERGE IN WINDOW — behavior change`. The fix lands in `kornia.geometry.transform`, not in `kornia.augmentation`, and it changes the pixels every current `align_corners=False` warp produces — it belongs with #3904 and PR #3945. Batch 6 records the executed defaults and the identity-warp behavior on the three classes, pins them, and points at #3904; it does not repair.

## Environment

VanniLeonardo:
## 📝 Description

Fixes #3904
Fixes #3928
Fixes #3929
Fixes #4411

`create_meshgrid` and `normal_transform_pixel` hardcoded the `align_corners=True`
normalization — pixel centers `0` and `size-1` mapped to `±1`. The grid /
normalization matrix built that way was then handed to
`grid_sample(align_corners=False)`, which reads `±1` as the outer pixel *edges*
instead. Mixing the two conventions applies a spurious sub-pixel scale and shift.

Relates two:

- **#3904** — `warp_perspective(img, I, dsize, align_corners=False)` does not
  reproduce its input under an identity homography.
- **#3928** — `warp_image_tps` is broken *by default*, because it defaults to
  `align_corners=False` while `create_meshgrid` always built the `True` grid.

Per @ducha-aiki's guidance on #3904, `align_corners` is **threaded through** the
public API rather than fixed internally, so callers keep full control of the
convention.

### The bug is wider than #3904 states

#3904 says `warp_affine` reproduces the input while `warp_perspective` does not.
That is only true for the **same-size identity** case, where two errors happen to
cancel. `warp_affine(align_corners=False)` was also wrong for:

- any non-identity transform, and
- identity when `dsize` differs from the source size.

Reproducer on `main` — a whole-pixel translation should shift by exactly one pixel:

```python
import torch

img = torch.arange(16.0).view(1, 1, 4, 4)
M = torch.tensor([[[1.0, 0.0, 1.0], [0.0, 1.0, 0.0]]])  # shift right by 1 px
print(warp_affine(img, M, (4, 4), align_corners=False))
```

On `main`:

```
tensor([[[[ 0.0000,  0.0000,  0.6667,  1.6667],
          [ 0.0000,  2.6667,  4.6667,  5.6667],
          [ 0.0000,  5.3333,  8.6667,  9.6667],
          [ 0.0000,  8.0000, 12.6667, 13.6667]]]])
```

— a fractional, non-uniform resample. With this PR:

```
tensor([[[[ 0.,  0.,  1.,  2.],
          [ 0.,  4.,  5.,  6.],
          [ 0.,  8.,  9., 10.],
          [ 0., 12., 13., 14.]]]])
```

which is the exact one-pixel shift, and matches `align_corners=True`.

---

## 🛠️ Changes Made

- [x] `create_meshgrid` gains `align_corners: bool = True`; when `False` it uses
      the half-pixel mapping `x_norm = (2x + 1) / W - 1`.
- [x] `normal_transform_pixel`, `normalize_homography` and
      `denormalize_homography` gain the same parameter and forward it to both
      normalization matrices. The `False` branch is well defined for a size of 1
      (it lands on the pixel center, `0`), so it needs no `eps` guard.
- [x] `warp_perspective`, `warp_affine`, `homography_warp` and `warp_image_tps`
      pass their own `align_corners` down to the grid/normalization they build.
- [x] **Removed** the `correction` matrix and the `h_out == 1 or w_out == 1`
      fallback from `crop_by_transform_mat`. Those were added in #3650 to work
      around this exact bug — the comment there read *"warp_perspective always
      builds its grid with align_corners=True spacing"*. Keeping them would
      double-correct, and the 1-px guard existed only to protect the now-deleted
      reparametrization.
- [x] Docs: `conventions.rst` gains an executable identity-consistency snippet
      and a note that `normalize_homography`'s `align_corners` (default `True`)
      must match the warp's — `homography_warp` alone defaults to `False`.
- [x] TPS docstrings: dropped the `.. warning::` placeholder #3926 added for
      #3928 and rewrote the Convention block, since both sides now live in a
      single frame selected by `align_corners`.

The new parameter is **last and defaults to `True`** in every signature, so all
existing `align_corners=True` call sites are bit-identical.

### Behaviour change (intended — this is the fix)

Outputs change for `align_corners=False`. That includes the augmentations that
default to it: `RandomAffine`, `RandomPerspective`, `RandomShear`,
`RandomTranslate`. Their *transform matrices* are unchanged (they were always
correct in pixel space); only the sampled pixels change, and they now match the
`align_corners=True` result up to out-of-bounds handling.

---

## 🧪 How Was This Tested?

- [x] **Unit Tests** (new):
  - `tests/geometry/test_grid.py::test_create_meshgrid_align_corners`
  - `tests/geometry/transform/test_imgwarp.py::TestWarpPerspective::test_convention_identity_agrees_with_warp_affine`
    — the exact check requested in #3904
  - `tests/geometry/transform/test_imgwarp.py::TestWarpPerspective::test_translation_align_corners`
  - `tests/geometry/transform/test_imgwarp.py::TestWarpAffine::test_translation_align_corners`
  - `tests/geometry/transform/test_homography_warper.py::test_transform_apply_align_corners_false`
  - `tests/geometry/transform/test_homography_warper.py::test_normalize_denormalize_round_trip_align_corners_false`

- [x] **Unit Tests** (updated — every expectation was recomputed by hand or
      against an independent pixel-space reference implementation, not
      regenerated from the new output):
  - `test_affine.py::TestShear` ×4 — e.g. `shear_x` is now
    `[[1,1,1,1],[0.5,1,1,1],[0,1,1,1],[0,0.5,1,1]]`; the old
    `0.75/0.25` values were the sub-pixel artifact.
  - `test_crop2d.py::test_align_corners_false` → `[[[[6.0, 6.5, 7.0], [10.0, 10.5, 11.0]]]]`.
    `test_convention_align_corners_ignored_under_slice_mode` renamed to
    `test_convention_align_corners_ignored_for_an_in_bounds_crop` and flipped to
    assert the two settings agree.
  - `test_homography_warper.py` — an artifact literal replaced by
    `assert_close(patch_dst, patch_src)`, i.e. the identity now round-trips.
  - `test_thin_plate_spline.py` — the `xfail(strict=True)` marker #3926 added for
    #3928 is **removed**; that test now passes.
  - `test_perspective_rand.py` — `expected_output` regenerated;
    `expected_transform` confirmed **unchanged**.

- [x] **Manual Verification:** built an independent pixel-space reference
      (`dst[p] = src[M⁻¹p]`) and compared it against `warp_affine` /
      `warp_perspective` / `crop_by_transform_mat` under both `align_corners`
      settings, for identity, whole-pixel translation, shear and resizing crops.
      All agree after the fix; several disagree before it.

- [x] **Performance/Edge Cases:** size-1 dimensions (the `align_corners=True`
      divide-by-zero `eps` path is untouched; the `False` path is exact),
      `dsize != src size`, and non-square images.

<details>
<summary>Local test log</summary>

```
$ uv run pytest tests/geometry/transform/ tests/geometry/test_grid.py \
      tests/augmentation/test_perspective_rand.py --dtype=float32,float64

============================= test session starts ==============================
platform linux -- Python 3.11.14, pytest-9.0.3, pluggy-1.6.0
cpu info:
	- Model name: AMD Ryzen 7 7700X 8-Core Processor
main deps:
    - kornia-0.9.0rc1
    - torch-2.9.1+cu128
...
======================= 1190 passed, 28 skipped in 1.90s =======================
```

Full suite, both dtypes:

```
$ pixi run test --dtype=float32,float64
14423 passed
```

(One `TestTrivialAugment::test_transform_mat` failure shows up in that run. It
passes in isolation and is ordering-dependent; running the identical command on
a clean `upstream/main` worktree gives the same `1 failed, 4408 passed` for
`tests/augmentation/`, so it is pre-existing and unrelated to this change.)

Also green: `pixi run lint`, `pixi run typecheck` ("All checks passed!"), and
`pixi run doctest` (48 passed, including the new `create_meshgrid` doctest).

</details>

---

## 🕵️ AI Usage Disclosure

- [x] 🟡 **AI-assisted:** I used AI for boilerplate/refactoring but have manually
      reviewed and tested every line.

---

## 🚦 Checklist

---

## 💭 Additional Context

Two related findings, **out of scope**:

1. **3D has the identical bug.** `normal_transform_pixel3d` /
   `normalize_homography3d` hardcode the same convention, so
   `warp_affine3d(..., align_corners=False)` is wrong the same way — a +1 voxel
   translation gives `max|diff| = 12.0` against the reference. Happy to open a
   follow-up issue and fix it in a separate PR; it needs its own expectation
   review across the 3D tests.

2. This also resolves the `align_corners=False` half of **#3929**.


ducha-aiki:
Thank you! 

  Before merge, a few asks:

  1. `crop_by_transform_mat`: the #3650 correction-matrix removal is the
     riskiest hunk — please confirm all `test_convention_*` crop pins pass
     and add/verify the 1-pixel-crop case (#3929) under both settings.
  2. Where you updated expected literals, keep the pre-fix literal in a
     comment next to the new one (with the generation snippet) — we archive
     old-vs-new pairs as conformance vectors.
  3. We'll merge #3942 (conversions convention blocks) first; please rebase
     on top of it and update any convention-block prose there that still
     describes the old inconsistent behavior.
  4. Yes to the 3D follow-up — please open the issue and consider it yours.

  Nice catch on the warp_affine non-identity cases cancelling.

ducha-aiki (changes requested):
See the comment

ducha-aiki (changes requested):
Delta re-review of head `c4fe1dbf` (rebased onto `539c9ecc`, one commit behind `main`; I re-read the whole diff, not only the three new commits).

Thanks for the careful rework. The fix itself is correct and I checked it from both sides:

- `align_corners=True` did not change at all: 435 outputs from `main` and this head were bitwise identical.
- `align_corners=False` now matches an independent pixel-space reference.

All four asks from the 2026-08-11 comment were checked. Ask 1 is done: `test_convention_one_pixel_output` covers both settings, and I reproduced your recorded pre-fix values on `main`. Ask 2 is done: pre-fix literals and generation snippets are kept for `TestShear`, `TestCropAndResize`, the `HomographyWarper` default, `RandomPerspective` and the TPS border factor. Ask 3 is done except the stale prose below. Ask 4, the 3D follow-up issue, is not open yet; I could not find it.

From the bot review, the `HomographyWarper` cached-grid mismatch is fixed in `ccc555b6` and has a pin that catches a revert. The stale crop2d warning is gone too. The `BaseTester`/type-hint and `__future__` items don't apply: `test_grid.py` is module-level functions throughout, and `grid.py` is TorchScript-scripted.

What still blocks the merge is mostly CI and merge fallout.

**[P2] The `docs (3.11)` job fails on this head.** At `docs/source/get-started/conventions.rst:135-138` the new paragraph is indented two spaces, directly after `.. code-block:: python`. Sphinx therefore puts it inside the literal block, and Pygments fails on the backticks: `conventions.rst:121: WARNING: Lexing literal_block ... resulted in an error at token: '`'`. Under `-W` that is the one warning that fails the build. A docutils parse of the file confirms the paragraph ends up inside the code block. The same indentation was already there at `cbe913e3`. Dedenting those four lines to column 0 fixes it.

**[P2] The CPU half-precision legs will fail.** I ran the touched files with `--xfail-known-failures --known-failure-profile=cpu-float16` / `cpu-bfloat16` (torch 2.14.0; CI uses 2.9.1). Compared with `main`, this head adds 16 FAILED nodes in float16 and 17 in bfloat16. They fall into two groups:

- **Strict XPASS.** The fix makes some manifest entries pass: `TestShear::test_shear_{x,y,batch2,batch2_broadcast}` in both manifests, and `TestCropAndResize::test_align_corners_false[cpu-bfloat16]`. Please delete those lines from `testing/half_precision_xfails/cpu_float16.txt` and `cpu_bfloat16.txt`.
- **New tests that fail at half precision under `atol=1e-4`.** Measured greatest differences run from 2.4e-4 to 6.25e-2. The failing tests:
  - `test_translation_align_corners` (warp_affine and warp_perspective, both settings in float16, `True` in bfloat16)
  - `test_normalize_denormalize_round_trip_align_corners_false` (7 sizes in float16, 5 in bfloat16)
  - `test_precompute_warp_grid_matches_direct`
  - in bfloat16 only: `test_convention_identity_agrees_with_warp_affine[True]`, `test_convention_align_corners_ignored_for_an_in_bounds_crop`, `test_convention_one_pixel_output[out_size0-True]`

  The existing `True`-path siblings (`test_translation`, `test_normalize_homography_identity`) are already manifest entries, so this is precision rather than a regression. The new tests still need dtype-aware tolerances, a half skip, or manifest entries recorded as `TESTING.md` describes.

**[P2] Merging with `main` breaks a pin that #4452 added.** `main` is `dc0af1a4`. I applied its one extra commit (#4452) on top of this head; the patch applies cleanly. `tests/augmentation` on CPU float32 then gives 1 failed / 2651 passed, against 0 failed / 2652 passed on `main`. The failing test is `TestConventionAugmentationSequential::test_wart_extra_args_mask_resample_override_is_discarded_4419` (`tests/augmentation/container/test_augmentation_sequential.py:1058`). It demonstrates that the `align_corners` half of `extra_args[DataKey.MASK]` reaches the sampler by asserting that `RandomPerspective(0.5)` changes the mask by `1.0`. That `1.0` was the half-pixel artifact this PR removes; the delta is now `0.0`. The warning text in `kornia/augmentation/_2d/geometric/base.py` ("the `align_corners` half does reach the sampler") is still true. Only the fixture needs replacing. For example, `K.RandomAffine(45.0, p=1.0, padding_mode="reflection")` on the same block mask gives a delta of `1.0` on this head, because reflection is where the two conventions still differ. Please rebase onto `main` and update that pin and its snippet comment.

**[P2] The release note misses changes on default-argument paths, and nothing will auto-close.** Several public functions already default to `align_corners=False`, so plain calls with no flag now return different output. The Breaking changes entry lists functions "when called with `align_corners=False`" but never says which defaults are `False`. Please name the default-path changes and give old vs new:

| Call with default arguments | `main` | this head |
|---|---|---|
| `homography_warp(x, I, (4, 5))`, identity, 4x5 `arange` | max\|out−in\| = 14.2 | 1.8e-15 |
| `HomographyWarper(4, 5)`, identity | 14.2 | 1.8e-15 |
| `ImageRegistrator("homography").warp_src_into_dst`, identity model (uses `HomographyWarper`) | 14.25 | 0.0 |
| `warp_image_tps`, identity | 14.2 | 1.8e-15 |
| `kornia.geometry.transform.shear` function (`affwarp.py:527`, default `False`), shear-x 0.5 on ones, first row | `0.75, 1, 1, 1` | `1, 1, 1, 1` |

`HomographyWarper` appears only in the Bug fixes entry, and `shear()` and `ImageRegistrator` appear nowhere. The Bug fixes entry also says "Fixes #3904, #3928, #3929". But the PR body says "Fixes/Relates to:", which GitHub does not parse (`closingIssuesReferences` is empty), so merging closes nothing. I checked each issue's full scope and all three are resolved on this head. #4411 is resolved as well: at the class default, `RandomPerspective(distortion_scale=0.)` goes from 0.734 to 6e-07 and `RandomThinPlateSpline(scale=0.)` from 0.734 to 1.5e-08, and `RandomElasticTransform` was already exact.

**[P2] The new `False` branch contradicts some prose.**
- `normalize_homography`'s warning (`kornia/geometry/conversions.py:1971-1973`) says that "under either convention" the equal-size identity residual is 0 at most sizes and 5.96e-08 at the rest, "with 4x4 among the latter". Measured over sizes 2..32 at `align_corners=False`, 4x4 is exact (`0.0`) and the sets differ. `False` is exact at 2, 4, 8, 9, 11, 14, 15, 16, 17, 19, 21, 23, 25, 27, 28, 29, 31, 32 and 5.96e-08 at 3, 5, 6, 7, 10, 12, 13, 18, 20, 22, 24, 26, 30. Quoting the figures for the default `True` only would fix it.
- `tests/augmentation/test_augmentation.py:5587` (`TestRandomThinPlateSpline::test_zero_scale_identity`) still says "align_corners=False has a separate sampling-grid defect tracked in #3928" and pins identity only at `align_corners=True`. At the class default that identity is now exact, so parametrizing over both values would drop the stale comment and pin #4411 at the augmentation level.

**[P3] `normal_transform_pixel(align_corners=False)` is not bit-identical between eager and graph capture.** At some sizes eager and graph capture differ by one float32 step. `torch.export` at runtime height 3 differs from eager by `5.96e-08` in `tx`, and `torch.compile` does the same at 3 of widths 1..199. Eager evaluates `1.0 / size - 1.0` in Python doubles and rounds once. The capture branch computes `1.0 / size_t` in float32 and then subtracts, which rounds twice. The `True` branch matches bitwise (0 of 398 heights). Writing the offset as `(1.0 - size) / size` in both branches makes it a single correctly rounded division; a float32 simulation over widths 1..4999 shows 0 mismatches. `create_meshgrid(align_corners=False)` is already bitwise equal between eager, compile and `jit.trace` in all four dtypes.

**[P3] Other stale prose.**
- `tests/geometry/test_conversions.py:3627-3630`, `4059-4069` and `4139-4140` expect a #3904 fix to "flip all of them" or "move both cells". The fix kept the default at `True` and none of them moved.
- The `denormalize_homography` bullet (`conversions.py:2341`) still calls that warning the "corner-alignment" warning.
- The first bullet of `normal_transform_pixel` still says the mapping "is **corner-aligned**" without "by default".
- `normal_transform_pixel3d` (`conversions.py:2217`) says "same unconditional application", which the 2-D function no longer has.
- `crop_by_transform_mat` (`crop2d.py:336`) says "(see warning below)", but the warning is now a note.
- The new `remap` warning in `conventions.rst:193` documents the same defect without a tracking issue. An issue for `remap` next to the 3D follow-up would make both easy to find.

**CI:** `docs (3.11)` is red, for the rst indentation above. The other 33 test legs were still queued when I checked.

**Verified on this head** (torch 2.14.0 unless noted):
- **`align_corners=True` unchanged.** I saved 435 outputs from head and from `main` and compared them bitwise; all were equal. The table covered:
  - `create_meshgrid`, `normal_transform_pixel`, `normalize_homography`, `denormalize_homography`, `normalize_pixel_coordinates`, `denormalize_pixel_coordinates`
  - `warp_perspective` and `warp_affine` with bilinear/nearest × zeros/border/reflection, dsize ≠ src, 1-pixel dims
  - `homography_warp`, `HomographyWarper` direct and precomputed, `crop_by_transform_mat` (3x3 and 2x3), `crop_and_resize`, `CenterCrop2D` resample, `shear`, `rotate`, `warp_image_tps`, `elastic_transform2d`
  - 12 augmentation forward and inverse outputs
  - all in float32, float64, float16 and bfloat16
- **`align_corners=False` is correct.**
  - `create_meshgrid` matches `F.affine_grid(identity, align_corners=False)` to within 5.96e-08.
  - `normal_transform_pixel` applied to the pixel grid reproduces that grid to 2.2e-16.
  - normalize→denormalize round-trips to 1e-15 in float64.
  - Checked against an independent pixel-space bilinear reference (random homography with a projective row; dsize (9,11), (5,13), (1,7), (6,1); zeros and border): `warp_perspective`, `warp_affine` and `crop_by_transform_mat` agree to ≤6e-07 under both settings, against 0.03–0.52 on `main` at `False`.
  - `RandomAffine` and `RandomPerspective` at `False` vs `True` with the same params: 1.6e-06 (0.118 and 0.553 on `main`).
  - 1-pixel crops keep the projective row.
  - `gradcheck` passes for `warp_perspective`, `warp_affine` and `warp_image_tps` at `False`.
  - `warp_perspective(False)` under `torch.compile` and `jit.trace` matches eager to ≤8.3e-07.
- **Tests.**
  - 9 touched files on CPU float32/float64: 1820 passed, 14 skipped, 11 xfailed.
  - Doctests for `geometry/grid`, `conversions`, `geometry/transform`, `augmentation/_2d/geometric`, `_3d` and `auto`: 107 passed.
  - MPS float32: 909 passed, with no failures that `main` does not also have (`main`: 883 passed).
  - `KORNIA_TEST_OPTIMIZER=inductor -k "dynamo or compile"`: 52 passed.
  - torch 2.5.1: 1153 passed.
  - New and updated tests copied onto `main`: 58 failed (every `False` case and each updated literal) and 52 passed (the `True` regression pins).
  - Mutations, each reverting one hunk: `HomographyWarper` cached grid gives 2 failures, the `warp_image_tps` grid 6, the `denormalize_homography` flag 18, the `warp_affine` flag 18.
  - `ruff@0.16.6` check and format are clean on the 16 changed Python files.

VanniLeonardo:
Thank you for the delta re-review, and for checking the fix from both sides. Everything in it is addressed on `73a41cbc`, which is merged up to `ca5021eb`. Going through the items in order.

**docs (3.11)**

Fixed. The paragraph at `conventions.rst:135-138` is dedented to column 0. A docutils parse now puts it outside every literal block, and Pygments lexes the code block with zero error tokens. That was the only warning, so the build should pass.

**CPU half-precision legs**

Both are green locally on torch 2.9.1, the version CI uses. I ran the full suite the way the blocking legs do.

```
pytest tests/ --dtype=float16   --verify-known-failures --known-failure-profile=cpu-float16
  9765 passed, 3807 skipped, 578 xfailed, exit 0
pytest tests/ --dtype=bfloat16 --verify-known-failures --known-failure-profile=cpu-bfloat16
  9830 passed, 3759 skipped, 561 xfailed, exit 0
```

The five XPASS lines you named are deleted from the two manifests.

For the new tests I used both of the other two options you offered, split by what each assertion is for. Measuring them made the split obvious. Every remaining failure was on the `align_corners=True` path, and the `False` path was exact. That is because `2 / (size - 1)` is representable only when the size happens to make it dyadic, while `2 / size` and `1 / size - 1` are exact at any power-of-two size. So the half-pixel convention lands on pixel centres more often, not less.

Each of these tests has a primary assertion about the convention and a secondary one about matching exact pixels. The primary assertions are bit-exact in every dtype, so I kept them at `atol=1e-4`. The secondary ones now take BaseTester's dtype-aware defaults.

| Test | Assertion kept tight | Assertion loosened |
|---|---|---|
| `test_precompute_warp_grid_matches_direct` | cached grid equals direct, delta 0.0 in all dtypes | identity reproduces input |
| `test_convention_identity_agrees_with_warp_affine` | perspective equals affine, delta 0.0 | identity reproduces input |
| `test_convention_one_pixel_output` | `(B, 3, 3)` equals `(B, 2, 3)`, delta 0.0 | matches the exact slice |

That leaves four nodes that still fail, all `align_corners=True`, and all with a True-path sibling already in the manifests. I recorded them the same way rather than widen a tolerance until the pin stopped proving anything.

```
cpu_float16   TestWarpAffine::test_translation_align_corners[True]
cpu_float16   TestWarpPerspective::test_translation_align_corners[True]
cpu_float16   test_normalize_denormalize_round_trip_align_corners_false[2-5]
cpu_bfloat16  TestWarpPerspective::test_translation_align_corners[True]
```

The correspondence with the existing entries is exact. `TestWarpAffine::test_translation` is manifested in float16 but not bfloat16, and `test_translation_align_corners[True]` fails in float16 but not bfloat16 for the same reason. `test_normalize_homography_identity` is manifested in both, and only one size of my round-trip test goes over.

**The #4452 pin**

Confirmed and fixed. The `1.0` delta came from `align_corners=True` and `False` disagreeing on in-bounds samples, which is exactly what this PR removes, so it went to `0.0`.

Your suggested fixture works. The two conventions still differ out of bounds, because `±1` spans a different extent under each, so a transform that samples outside the frame plus a padding mode that makes those samples visible restores the signal. `RandomAffine(45.0, p=1.0, padding_mode="reflection")` on the same block mask gives exactly `1.0` at every seed 0 to 9, in float32, float64, float16 and bfloat16. I swapped the fixture and rewrote the snippet comment to record why. The wart itself is unchanged, and `resample` on the new fixture is still `0.0`.

**Release notes**

Rewritten. The Breaking changes entry now leads with the fact that several of these default to `align_corners=False`, and carries your table. I reproduced every row on both sides before writing it. In float64 all four warp rows read `14.25` on `main`, so I used that rather than `14.2`.

| Call, default arguments | before | after |
|---|---|---|
| `homography_warp(x, I, (4, 5))` | 14.25 | 1.8e-15 |
| `HomographyWarper(4, 5)(x, I)` | 14.25 | 1.8e-15 |
| `warp_image_tps(x, pts, kernel, affine)` | 14.25 | 1.8e-15 |
| `ImageRegistrator("homography").warp_src_into_dst(x)` | 14.25 | 0.0 |
| `shear(ones(1, 1, 4, 4), [[0.5, 0.0]])`, first row | `0.75, 1, 1, 1` | `1, 1, 1, 1` |

`shear` and `ImageRegistrator` are named in the entry now, and `HomographyWarper` appears in both. I also added your #4411 figures, which I reproduce as 5.96e-07 for `RandomPerspective(distortion_scale=0.)` and 1.49e-08 for `RandomThinPlateSpline(scale=0.)`.

On the closing keywords, you are right that `Fixes/Relates to:` parses as nothing. I have the corrected body ready with one `Fixes` line each for #3904, #3928, #3929 and #4411. I will apply it before you merge, or tell me if you would rather #4411 stayed open for the augmentation-level work.

**Prose the False branch contradicted**

Both fixed. The `normalize_homography` warning no longer claims the residual pattern holds under either convention. I swept equal sizes 2 to 32 and got your result, so the warning now quotes the default `True` figures only, names the nine sizes where it is `5.96e-08`, and says explicitly that the two conventions land on different sets because it is a property of the inverse-and-matmul chain.

`TestRandomThinPlateSpline::test_zero_scale_identity` is parametrized over both `align_corners` values, and the stale #3928 comment is gone. It passes at the class default now, which pins #4411 at the augmentation level as you suggested.

**normal_transform_pixel eager versus capture**

Fixed with your formula. Writing the offset as `(1.0 - size) / size` in both branches makes it one correctly rounded division. A float32 simulation over sizes 1 to 4999 gives 0 mismatches, against 5 for the old two-step form, at sizes 3, 11, 31, 251 and 4051.

I pinned it with `test_normal_transform_capture_matches_eager_bitwise`, parametrized over those five sizes and both conventions. It compares `torch.export` against eager at `atol=0`. On the old formula it fails all five `False` cases and passes all five `True` ones, which matches your finding that `True` never had the defect.

**Other stale prose**

All six corrected. The three `test_conversions` notes predicted that a #3904 fix would flip every literal or move both cells. It did not, because the default stayed `True`, so they now record that, and the residual note adds that the two sets differ at `False`. The `denormalize_homography` bullet calls it the `align_corners` warning. The first `normal_transform_pixel` bullet says corner-aligned by default and points at the flag. `normal_transform_pixel3d` says it has no such parameter instead of borrowing the 2D function's old wording. `crop_by_transform_mat` says note rather than warning.

The `remap` warning now cites #4504. It still needs a public `align_corners` on `normalize_pixel_coordinates`, which is wider than this PR.

**3D follow-up**

Open now, two issues.

#4503 is the one you asked for. `normal_transform_pixel3d` and `normalize_homography3d` take no `align_corners`, so a whole-voxel translation through `warp_affine3d` is off by 6.67 at `align_corners=False` and exact at `True`. That is the same shape as the 2D defect and I have taken it.

#4502 is one I did not expect. `warp_perspective3d` does not reproduce an identity warp under **either** setting, by 45.0 on a 4x4x4 cube at `True` and 55.125 at `False`. The cause is not `align_corners`. `create_meshgrid3d` emits its channels in `(z, x, y)` order while `grid_sample` reads a 5D grid as `(x, y, z)`, and `homography_warp3d` passes the grid straight through. `warp_affine3d` escapes it because it builds its grid with `F.affine_grid`. A hand-built `(x, y, z)` grid with the same normalization is exact under both settings, which isolates the channel order as the whole of it. Nothing pins this today, since `warp_perspective3d`'s only numerical test compares a batched call against an unbatched one and both are wrong the same way.

That ordering matters for #4503. The 3D `align_corners` fix can be verified through `warp_affine3d` now, but not through `warp_perspective3d` until #4502 is fixed.

All three issues are cited in the conventions page and in `normalize_homography`'s warning, so nothing is left documented without a number to follow.

**CI**

One leg is red, `tests (macos-latest, float32, 3.12, 2.5.1)`, and `collector` is red because it aggregates it. The failure is not from this PR.

```
  Failed: Timeout (>120.0s) from pytest-timeout
```

The same test times out on `main` with the same message, on the same macOS configuration, in both scheduled runs since yesterday, runs 34779403220 and 34804576631. This branch touches nothing under `kornia/contrib`, `kornia/onnx` or `kornia/models`. The test itself is healthy, and takes 5.5s here against its 120s budget, so it looks like macOS runner speed rather than the `@pytest.mark.timeout(120)` that #4402 set. I cannot re-run the job myself.

Yesterday's scheduled run on `main` also failed `tests/feature/test_sosnet.py::TestSOSNet::test_jit[cpu-float64]` on the ubuntu float64 2.6.0 leg, which this PR did not hit. I did not find an open issue for either, so tell me if you would like one.

**Local verification on 73a41cbc**

```
pytest tests/ --dtype=float32                            10382 passed, 3728 skipped
pytest tests/ --dtype=float64                            10382 passed, 3727 skipped
pytest tests/ --dtype=float16  --verify-known-failures    9765 passed, exit 0
pytest tests/ --dtype=bfloat16 --verify-known-failures    9830 passed, exit 0
pytest --doctest-modules kornia/                            655 passed, 14 skipped
KORNIA_TEST_OPTIMIZER=inductor, dynamo/compile/export/trace  182 passed
pre-commit on the 23 changed files                        all hooks passed
pixi run typecheck                     1 pre-existing diagnostic in kornia/feature/lightglue.py
```

ducha-aiki (commented):
Review of `ab0af993b2a7276f4e57fe77d13befc55059f5a7`.

The current head resolves every technical item from the previous delta review: the rst block is dedented, the half manifests and tolerances are reconciled, #4452's augmentation pin uses a still-discriminating reflection fixture, default-path changes are described in `changelog.d/3945.breaking.md`, issue references are updated, eager/captured offsets match bitwise, the stale 2-D/3-D prose is corrected, and #4502/#4503/#4504 track the remaining work. Focused checks and all 41 hosted checks are green. This is still not mergeable under the maintainer's recorded repair-window decision. The `Rules for this batch` section of #4407 explicitly classifies #4411 as `MERGE IN WINDOW — behavior change` and says such issues are not fixed ahead of the coordinated window and a PR doing so will be closed. This head intentionally fixes #4411 and names it in `changelog.d/3945.breaking.md`; I found no later maintainer comment granting an exception. A maintainer must either explicitly lift that gate for this cross-cutting #3904 fix or defer/close it under the recorded rule.

Validation and limits: Focused CPU float32/float64 align-corners run imported `/private/tmp/kornia-triage-60Oa1g/wt-3945/kornia/__init__.py`: 40 passed, 897 deselected. GitHub reports 41 successful checks and a mergeable head. The policy evidence is the current body of issue #4407; neither PR #3945 nor issue #4411 contains a superseding maintainer decision.

ducha-aiki (changes requested):
Review of `ab0af993b2a7276f4e57fe77d13befc55059f5a7`.

Formal verdict: changes requested because the recorded repair-window hold remains unresolved. This records the merge recommendation from my earlier comment-only review; the earlier technical findings are resolved.

The current head fixes #4411, which the maintainer’s #4407 policy explicitly reserves for the coordinated repair window. Before merging, obtain and record an explicit maintainer exception for this cross-cutting geometry repair, or defer it to that window. The user request to formalize review states does not itself lift that policy. No additional code defect was found in the previous re-review.

VanniLeonardo:
Thanks for confirming the technical side is resolved. On the repair-window hold, I would like to ask for the explicit exception rather than defer, and I think the record already supports it.

**This PR fixes #3904, which is not in the batch-6 list.** #4407 is the `kornia.augmentation` audit. Its `Rules for this batch` govern that batch's own PRs, and #3904 does not appear anywhere in it. #3904 was approved and assigned to me by @ducha-aiki on 10 August, with the instruction to thread `align_corners` through and to cover #3928 in the same PR. That is the change this PR makes.

**#4411's own text sends its repair here.** #4411 is tagged `MERGE IN WINDOW`, but the same issue body says where the fix goes:

> The fix lands in `kornia.geometry.transform`, not in `kornia.augmentation`, and it changes the pixels every current `align_corners=False` warp produces. It belongs with #3904 and PR #3945. Batch 6 records the executed defaults and the identity-warp behavior on the three classes, pins them, and points at #3904; it does not repair.

and earlier:

> No augmentation-level change is needed once the warp is fixed; the three classes inherit the correction.

So the tag on #4411 tells batch-6 PRs not to touch the augmentation classes, and points at this PR as the place the geometry repair happens. This PR contains no augmentation-level change. The three classes inherit the fix because they delegate to `warp_perspective`, `warp_image_tps` and `elastic_transform2d`, exactly as #4411 describes.

**Why I would rather not strip the #4411 reference.** Removing `Fixes #4411` from the body and fragment would make the hold disappear on paper without changing a single pixel of behaviour, since the inheritance is structural. That would be less honest, not more compliant. The `changelog.d/3945.breaking.md` fragment names the three classes and gives before and after figures precisely so nobody is surprised.

**Two ways to resolve it, either is fine by me.**

1. Record the exception here: #3904 is the geometry repair that #4411 defers to, so this PR is allowed to land outside the window and close #4411 with it.
2. Keep #4411 open. I remove the `Fixes #4411` line and leave a `Refs #4411` in the fragment, so the issue stays available as the batch-6 record and is closed by hand when the window runs. The behaviour of this PR is identical either way.

If neither works and the whole geometry fix must wait for the window, tell me and I will mark the PR as such rather than let it go stale again.

Separately, `main` has since merged the #4419 repair (#4509), which reworked the test this PR had re-fixtured. I am resolving that conflict now.

ducha-aiki (approved):
Review of `37df497e2ec2d9605ddd6475c0285041e082748b`.

Maintainer decision: this merges now as an explicit exception to the #4407 repair-window hold, which answers the request above. #3904 is not on the batch-6 list, and the #4411 behaviour change is exactly what this PR delivers. That replaces my earlier policy-only change request; the technical re-review from 2026-09-14 stands.

I pushed two commits to the branch:

- `1e88eda8`, a merge of `main`. The only textual conflict was the `create_meshgrid` docstring in `kornia/geometry/grid.py`: #4501 added a "See Conventions & Pitfalls" link where this PR added the `grid_sample` convention paragraph. Both are kept, link after paragraph, so "two such conventions" still refers to the paragraph above it. #4501's links in `conversions.py`, `imgwarp.py` and `homography_warper.py` auto-merged at the top of the Convention blocks, and I read each one in context.
- `37df497e`, a docs fix. The 3-D caveat this PR added to `docs/source/get-started/conventions.rst` said `warp_perspective3d` is wrong under both settings because of the `(d, x, y)` grid order (#4502). #4514 fixed that on `main` yesterday. What remains is #4503, measured on the merge:
  - an identity `warp_perspective3d` on a 4x4x4 `arange` volume is exact at `align_corners=True` and off by `55.1` at `False`;
  - a 0.5 scale through `warp_affine3d` gives `52.5` at voxel (1, 1, 1) at `False`, where `42` is expected, and `42` at `True`.

  `main` gives the same numbers, so this PR changes nothing in 3-D. The paragraph now says both 3-D warps have the #4503 mismatch at `align_corners=False`.

Checks on `37df497e`, torch 2.14.0, CPU:

- **float32/float64:** 2073 passed, 120 skipped, 15 xfailed across `test_grid`, `test_conversions`, `test_imgwarp`, `test_homography_warper`, `test_crop2d`, `test_affine`, `test_thin_plate_spline`, `test_imgwarp3d`, `test_perspective_rand` and `test_augmentation_sequential`.
- **float16 / bfloat16** (`cpu-<dtype>` profiles, plus `test_augmentation.py`): the failure and xpass sets are identical to `main`'s, node for node (7 and 8 nodes).
- **Doctests** for `grid`, `conversions`, `imgwarp` and `crop2d`: 47 passed.
- **torch 2.5.1:** grid, imgwarp, crop2d and homography_warper give 665 passed.
- **ruff 0.16.7:** check and format are clean.

Issue #3518, opened 2026-01-19
[Bug]: PadTo is probably misnamed, since it crops if the image is too large.


bw4sz:
### 🐛 Describe the bug

Issue: PadTo crops images when they are larger than the target size,
instead of only padding when smaller. 

Expected behavior: PadTo should only pad (add borders), never crop.
When image is larger than target, it should remain unchanged.

This might be just a case where the name or docs should be fixed. I think pad always suggest getting bigger. The behavior here is like resize + pad. This matters for object detection pipeline, since you might silently lose alot of data outside the bounds for images of mixed sizes.

### 🔄 Steps to Reproduce

```bash
1. Load the module
2. Try PadTo on an image that is larger than the target size
3. Confirm that the image has been cropped to that size.
```

### 💻 Minimal Code Example

```python
import torch
import kornia.augmentation as K
from kornia.constants import DataKey

# Create a test image larger than target size
# Image: 1100 x 1100 (larger than target 1000 x 1000)
image = torch.randn(1, 3, 1100, 1100)  # (B, C, H, W)

# Create bounding boxes in original image coordinates
# Box at bottom-right corner: (700, 1000, 800, 1100)
# This box will be out of bounds after cropping to 1000x1000
boxes = torch.tensor([[[700.0, 1000.0, 800.0, 1100.0]]])  # (B, N, 4) in xyxy format

print("=" * 60)
print("Kornia PadTo Issue - Minimal Reproducible Example")
print("=" * 60)
print(f"\nInput image shape: {image.shape}")
print(f"Target size: (1000, 1000)")
print(f"Input boxes: {boxes[0, 0].tolist()}")
print(f"  Box coordinates: xmin={boxes[0,0,0]:.0f}, ymin={boxes[0,0,1]:.0f}, "
      f"xmax={boxes[0,0,2]:.0f}, ymax={boxes[0,0,3]:.0f}")

# Apply PadTo augmentation
pad_transform = K.PadTo(size=(1000, 1000))
aug_seq = K.AugmentationSequential(
    pad_transform, data_keys=[DataKey.IMAGE, DataKey.BBOX_XYXY]
)

# Apply transformation
augmented_image, augmented_boxes = aug_seq(image, boxes)

print(f"\nOutput image shape: {augmented_image.shape}")
print(f"Output boxes: {augmented_boxes[0, 0].tolist()}")
print(f"  Box coordinates: xmin={augmented_boxes[0,0,0]:.0f}, "
[...]

PR #3875, opened 2026-07-30
feat(models): add pretrained Kimi-VL weight loading

TomasGuija:
## 📝 Description

Fixes #3869

This PR adds pretrained weight loading for Kimi-VL through a public `KimiVLBuilder` API, following the existing [SigLIP2 builder](https://github.com/kornia/kornia/blob/main/kornia/models/siglip2/builder.py) pattern. It supports automatic Hugging Face download, local safetensors checkpoints, and strict state-dictionary loading for [Kimi-VL-A3B-Instruct](https://huggingface.co/moonshotai/Kimi-VL-A3B-Instruct).

Maintainer @edgarriba explicitly approved the contribution in [this comment](https://github.com/kornia/kornia/issues/3869#issuecomment-5092273127).

The converted checkpoint was temporarily hosted and tested from my [personal Hugging Face repository](https://huggingface.co/TomasGuija/kornia-kimi-vl-a3b-instruct-vision). The committed identifier targets the intended Kornia-owned location, which must be provisioned before merging.

---

## 🛠️ Changes Made
- [x] Added `KimiVLBuilder` with configuration, Hugging Face, and local checkpoint loading interfaces.
- [x] Added the centralized Kimi-VL-A3B-Instruct pretrained configuration.
- [x] Exported `KimiVLBuilder` through the Kimi-VL public API.
- [x] Added focused unit tests for the builder interfaces.

---

## 🧪 How Was This Tested?
- [x] **Unit Tests:** (List new/updated tests)
Ran `python -m pytest tests/models/test_kimi_vl.py -q`; all 16 tests passed.

- [x] **Manual Verification:** (Describe the steps you took)
I ran a simple verification script: 

 ```python
import torch
from huggingface_hub import hf_hub_download

import kornia.models.kimi_vl.builder as kimi_builder
from kornia.models.kimi_vl import KimiVLBuilder

REPO = "TomasGuija/kornia-kimi-vl-a3b-instruct-vision"
kimi_builder._KIMI_VL_A3B_INSTRUCT_REPO_ID = REPO

x = torch.rand(1, 3, 28, 28)
with torch.inference_mode():
    remote_model = KimiVLBuilder.from_pretrained_hf().eval()
    remote_output = remote_model(x)
    del remote_model

    checkpoint = hf_hub_download(REPO, "model.safetensors")
    local_model = KimiVLBuilder.from_checkpoint(checkpoint).eval()
    local_output = local_model(x)

torch.testing.assert_close(local_output, remote_output)
[...]

PR #3964, opened 2026-08-10
fix(augmentation): add crop_if_exceeds option to PadTo to avoid silent cropping

Buland-Choudhary:

## 📝 Description

**Fixes:** #3518

`PadTo` is documented as a padding transform, but when the input is larger than the target `size` on a given dimension, it silently crops instead of leaving that dimension alone — because it always calls `torch.nn.functional.pad` with `target - current`, and `F.pad` treats a negative pad value as a trim on that edge (this is documented PyTorch behavior, not a bug in `F.pad` itself: https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html). This PR adds an opt-in `crop_if_exceeds: bool = True` flag, proposed by @shijianjian in the issue thread and approved by @ducha-aiki. Default `True` preserves the existing behavior for anyone relying on it; `False` clamps the pad amount to `max(pad, 0)` so oversized dimensions are left unchanged instead of cropped.

## 🛠️ Changes Made
- [x] Added `crop_if_exceeds: bool = True` parameter to `PadTo.__init__` (`kornia/augmentation/_2d/geometric/pad.py`)
- [x] `apply_transform` clamps `height_pad`/`width_pad` to `max(..., 0)` when `crop_if_exceeds=False`
- [x] Updated docstring with parameter description and a runnable example
- [x] Added 3 unit tests covering: crop-on-true (regression), no-crop-on-false, and mixed dims (one axis oversized, one undersized)

## 🧪 How Was This Tested?
- [x] **Unit Tests:** `TestPadTo::test_crop_if_exceeds_true_crops_larger_input`, `test_crop_if_exceeds_false_keeps_larger_input_unchanged`, `test_crop_if_exceeds_false_mixed_dims`, plus existing `test_smoke` (unchanged, still passing)
- [x] **Manual Verification:** Ran the full `tests/augmentation/test_augmentation.py` suite (422 tests) to check for regressions elsewhere, plus `pixi run lint`, `pixi run typecheck`, and the module doctest
- [x] **Edge Cases:** Explicit test for the mixed case where one dimension needs padding and the other needs to be left alone in the same call

```
$ pixi run --environment default -- uv run pytest -v "tests/augmentation/test_augmentation.py::TestPadTo"
tests/augmentation/test_augmentation.py::TestPadTo::test_smoke[cpu-float32] PASSED [ 25%]
tests/augmentation/test_augmentation.py::TestPadTo::test_crop_if_exceeds_true_crops_larger_input[cpu-float32] PASSED [ 50%]
tests/augmentation/test_augmentation.py::TestPadTo::test_crop_if_exceeds_false_keeps_larger_input_unchanged[cpu-float32] PASSED [ 75%]
tests/augmentation/test_augmentation.py::TestPadTo::test_crop_if_exceeds_false_mixed_dims[cpu-float32] PASSED [100%]
======================== 4 passed in 0.39s ========================

$ pixi run --environment default -- uv run pytest -q tests/augmentation/test_augmentation.py
collected 422 items
======== 327 passed, 60 skipped, 16 xfailed, 5 xpassed, 4 warnings in 3.87s ========

$ pixi run --environment default -- uv run pytest -v --doctest-modules kornia/augmentation/_2d/geometric/pad.py
kornia/augmentation/_2d/geometric/pad.py::kornia.augmentation._2d.geometric.pad.PadTo PASSED [100%]
======================== 1 passed in 0.12s ========================

$ pixi run --environment default lint

$ pixi run --environment default typecheck
All checks passed!
```

## 🕵️ AI Usage Disclosure
- [x] 🟡 **AI-assisted:** Used Claude Code for boilerplate and drafting (implementation, tests, docstring), but reviewed and tested every line locally, understand why `F.pad` crops on negative padding, and can explain the fix in review.

## 🚦 Checklist
- [x] I am assigned to the linked issue
- [x] The linked issue has been approved by a maintainer (`ducha-aiki`)
- [x] Self-review performed
- [x] Follows existing style guidelines (`ruff` clean)
- [x] Tests added proving the fix works
- [ ] Screenshots/recordings — N/A, no UI change

## 💭 Additional Context
Reference for root cause: [`torch.nn.functional.pad` docs](https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html) — negative padding values remove elements from that side. API design (`crop_if_exceeds` name and default) follows the consensus reached in [#3518](https://github.com/kornia/kornia/issues/3518) between `shijianjian` and `pragnyanramtha`, approved by `ducha-aiki`.

### DIFF
--- a/kornia/augmentation/_2d/geometric/pad.py
+++ b/kornia/augmentation/_2d/geometric/pad.py
@@ -98,8 +98,11 @@ class PadTo(GeometricAugmentationBase2D):
         height_pad: int = flags["size"][0] - height
         width_pad: int = flags["size"][1] - width
         if not flags["crop_if_exceeds"]:
-            height_pad = max(height_pad, 0)
-            width_pad = max(width_pad, 0)
+            # torch.sym_max keeps this clamp traceable under torch.compile/torch.export:
+            # height_pad/width_pad are derived from input.shape and may be SymInt, and the
+            # builtin max() would force a guard/specialization on their sign.
+            height_pad = torch.sym_max(height_pad, 0)
+            width_pad = torch.sym_max(width_pad, 0)
         return torch.nn.functional.pad(
             input, [0, width_pad, 0, height_pad], mode=flags["pad_mode"], value=flags["pad_value"]
         )
--- a/kornia/augmentation/_2d/geometric/pad.py
+++ b/kornia/augmentation/_2d/geometric/pad.py
@@ -36,6 +36,9 @@ class PadTo(GeometricAugmentationBase2D):
         pad_value: fill value for 'constant' padding applied to the image
         keepdim: whether to keep the output shape the same as input (True) or broadcast it
                  to the batch form (False).
+        crop_if_exceeds: if True (default), inputs larger than ``size`` on a given dimension
+            are cropped to fit, matching the historical behavior. If False, inputs larger than
+            ``size`` are left unchanged on that dimension instead of being cropped.
 
     Shape:
         - Input: :math:`(C, H, W)` or :math:`(B, C, H, W)`, Optional: :math:`(B, 3, 3)`
@@ -61,13 +64,27 @@ class PadTo(GeometricAugmentationBase2D):
                   [0., 0., 0.],
                   [0., 0., 0.]]]])
 
+        >>> img_large = torch.zeros(1, 1, 4, 4)
+        >>> PadTo((2, 2), crop_if_exceeds=False)(img_large).shape
+        torch.Size([1, 1, 4, 4])
+
     """
 
     def __init__(
-        self, size: Tuple[int, int], pad_mode: str = "constant", pad_value: float = 0, keepdim: bool = False
+        self,
+        size: Tuple[int, int],
+        pad_mode: str = "constant",
+        pad_value: float = 0,
+        keepdim: bool = False,
+        crop_if_exceeds: bool = True,
     ) -> None:
         super().__init__(p=1.0, same_on_batch=True, p_batch=1.0, keepdim=keepdim)
-        self.flags = {"size": size, "pad_mode": pad_mode, "pad_value": pad_value}
+        self.flags = {
+            "size": size,
+            "pad_mode": pad_mode,
+            "pad_value": pad_value,
+            "crop_if_exceeds": crop_if_exceeds,
+        }
 
     # TODO: It is incorrect to return identity
     # TODO: Having a resampled version with ``warp_affine``
@@ -80,6 +97,9 @@ class PadTo(GeometricAugmentationBase2D):
         _, _, height, width = input.shape
         height_pad: int = flags["size"][0] - height
         width_pad: int = flags["size"][1] - width
+        if not flags["crop_if_exceeds"]:
+            height_pad = max(height_pad, 0)
+            width_pad = max(width_pad, 0)
         return torch.nn.functional.pad(
             input, [0, width_pad, 0, height_pad], mode=flags["pad_mode"], value=flags["pad_value"]
         )
--- a/tests/augmentation/test_augmentation.py
+++ b/tests/augmentation/test_augmentation.py
@@ -4742,6 +4742,27 @@ class TestPadTo(BaseTester):
         assert out.shape == (1, 1, 4, 5)
         self.assert_close(aug.inverse(out), img)
 
+    def test_crop_if_exceeds_true_crops_larger_input(self, device, dtype):
+        img = torch.rand(1, 1, 6, 6, device=device, dtype=dtype)
+        aug = PadTo(size=(4, 5), crop_if_exceeds=True)
+        out = aug(img)
+        assert out.shape == (1, 1, 4, 5)
+
+    def test_crop_if_exceeds_false_keeps_larger_input_unchanged(self, device, dtype):
+        img = torch.rand(1, 1, 6, 6, device=device, dtype=dtype)
+        aug = PadTo(size=(4, 5), crop_if_exceeds=False)
+        out = aug(img)
+        assert out.shape == (1, 1, 6, 6)
+        self.assert_close(out, img)
+        self.assert_close(aug.inverse(out), img)
+
+    def test_crop_if_exceeds_false_mixed_dims(self, device, dtype):
+        # height (6) exceeds target (4); width (2) is below target (5)
+        img = torch.rand(1, 1, 6, 2, device=device, dtype=dtype)
+        aug = PadTo(size=(4, 5), crop_if_exceeds=False)
+        out = aug(img)
+        assert out.shape == (1, 1, 6, 5)
+
 
 class TestResize:
     def test_smoke(self, device, dtype):
--- a/tests/augmentation/test_augmentation.py
+++ b/tests/augmentation/test_augmentation.py
@@ -4747,6 +4747,7 @@ class TestPadTo(BaseTester):
         aug = PadTo(size=(4, 5), crop_if_exceeds=True)
         out = aug(img)
         assert out.shape == (1, 1, 4, 5)
+        self.assert_close(out, img[..., :4, :5])
 
     def test_crop_if_exceeds_false_keeps_larger_input_unchanged(self, device, dtype):
         img = torch.rand(1, 1, 6, 6, device=device, dtype=dtype)
@@ -4762,6 +4763,10 @@ class TestPadTo(BaseTester):
         aug = PadTo(size=(4, 5), crop_if_exceeds=False)
         out = aug(img)
         assert out.shape == (1, 1, 6, 5)
+        # the exceeding height dimension is left untouched, and the original
+        # content is preserved with zeros padded onto the short width dimension
+        self.assert_close(out[..., :2], img)
+        self.assert_close(out[..., 2:], torch.zeros(1, 1, 6, 3, device=device, dtype=dtype))
 
 
 class TestResize:


ducha-aiki (changes requested):
Please, @Buland-Choudhary, fix "Crop test checks only shape" and "Mixed-dimension test checks only shape" suggestions from bots, as we are good to go

Buland-Choudhary:
> Please, @Buland-Choudhary, fix "Crop test checks only shape" and "Mixed-dimension test checks only shape" suggestions from bots, as we are good to go

Fixed both — the crop test now checks that the cropped output actually matches img[..., :4, :5], and the mixed-dims test checks that the untouched height content matches the original and the padded width columns are zero, not just shapes.

Also went ahead and swapped the max() clamp for torch.sym_max() in apply_transform, since height_pad/width_pad come from input.shape and a plain Python max() forces a guard on a SymInt under torch.compile. Confirmed torch.compile(aug, fullgraph=True, dynamic=True) still matches eager output exactly with crop_if_exceeds=False.

Ran the full TestPadTo suite locally, all green, and ruff check/ruff format --check are clean. Let me know if you want anything else adjusted.

ducha-aiki:
**Changes requested — the design and the default are correct and measurably behavior-preserving, but the `torch.sym_max` swap added in `39b6506` is a net regression: it turns a 0-graph-break op into 5 graph breaks and makes `torch.compile(fullgraph=True)` fail outright in the default dynamic mode. Revert that one hunk, pin the default, and this is good to merge.**

Measured on head `39b65069f75a998a42352e232b68bfbeacd72a2f` vs merge-base `9ec66f3` (`feat(models): add pretrained Kimi-VL weight loading (#3875)`), torch 2.9.1, Python 3.11, CPU, float32+float64. Not compared against `main` — `main` has moved 29 commits since the branch point and only `f772cb3` touches `kornia/augmentation/`, unrelated to `PadTo`. The branch still merges cleanly: `git merge-tree --write-tree main 39b65069` → `c87fd3d5b331e95b20e8bd9706214b6515560de1`, no conflicts.

## What is already right

**The default preserves today's behavior exactly.** I ran a 14-case matrix (smaller / equal / larger / larger-in-one-axis-only, two `pad_value`s, forward + `inverse`) against `PadTo((4, 5))` on both worktrees:

```
cases: 28   max abs diff: 0.0   shape/other mismatches: []
```

So this is **additive**, not a semantic change — it does not need the coordinated repair window and can land as soon as the items below are fixed.

**Both review comments I asked for on 2026-08-11 are genuinely fixed**, not just reshuffled. `test_crop_if_exceeds_true_crops_larger_input` now asserts `self.assert_close(out, img[..., :4, :5])` (pins the crop *edge*, which was the whole point of the Copilot note), and `test_crop_if_exceeds_false_mixed_dims` now asserts content on both the untouched axis and the padded columns. Thank you.

**The augmentation-specific surface is clean.** I checked the things that usually break when a flag is bolted onto an augmentation, and all of them work:

```
state_dict: {}                                  # nothing to break
deepcopy flags:  {..., 'crop_if_exceeds': False}
pickle  flags:   {..., 'crop_if_exceeds': False}   pickled module output equal: True
load_state_dict from a pre-PR PadTo checkpoint:   OK, flags intact
batched (4,3,6,2), same_on_batch:  True -> (4,3,4,5)   False -> (4,3,6,5)
AugmentationSequential(input, mask, keypoints, bbox) -> [(1,3,6,5), (1,1,6,5), (1,2,2), (1,1,4,2)]
  inverse -> [(1,3,6,2), (1,1,6,2), (1,2,2), (1,1,4,2)]   img roundtrip equal: True
transform_matrix with crop_if_exceeds=False: identity (unchanged, correct for bottom/right pad)
3D (C,H,W) input, keepdim=True/False: (1,6,6) / (1,6,6)   equal-size input: exact no-op
pad_mode='reflect' + oversized + crop_if_exceeds=False: OK (1,1,6,6)
```

**The reported bug is actually fixed end to end.** The exact repro from #3518:

```
crop_if_exceeds=True   img (1, 3, 1000, 1000)   boxes [700.0, 1000.0, 800.0, 1100.0]
crop_if_exceeds=False  img (1, 3, 1100, 1100)   boxes [700.0, 1000.0, 800.0, 1100.0]
```

**Everything else is green.** `tests/augmentation` → `2321 passed, 3226 skipped, 24 xfailed, 5 xpassed`; the module doctest passes; `ruff check` / `ruff format --check` clean on both files. The docstring documents the parameter (`pad.py:39-41`) and `docs/source/augmentation.module.rst:59` picks it up via `autoclass`, so the docs side is done.

On the design question — should `crop_if_exceeds=False` pass through or raise: pass-through is the right call and I'm not reopening it. It is what was agreed on the issue, it is what `albumentations.PadIfNeeded` does, and raising would make the flag useless for the mixed-size detection pipeline that motivated #3518. See nit (b) about saying so in the docstring.

## Blockers

### 1. `torch.sym_max` at `kornia/augmentation/_2d/geometric/pad.py:100-105` makes `torch.compile` strictly worse. Revert to builtin `max()`.

You swapped `max()` for `torch.sym_max()` in response to the Qodo bot's speculative "SymInt-unsafe pad clamp" finding, and verified it under `torch.compile(..., fullgraph=True, dynamic=True)` — which is the one mode where it happens to work. In the other two it fails:

```
torch 2.9.1
dynamic=None  crop_if_exceeds=True  -> OK (1, 1, 4, 5)
dynamic=None  crop_if_exceeds=False -> Unsupported: torch.* op returned non-Tensor
dynamic=True  crop_if_exceeds=True  -> OK (1, 1, 4, 5)
dynamic=True  crop_if_exceeds=False -> OK (1, 1, 6, 6)
dynamic=False crop_if_exceeds=True  -> OK (1, 1, 4, 5)
dynamic=False crop_if_exceeds=False -> Unsupported: torch.* op returned non-Tensor
```

`dynamic=None` is `torch.compile`'s **default**. When shapes are concrete ints (the common case on the first call), `torch.sym_max` returns a plain `int` and Dynamo rejects it: *"torch.* ops that return a non-Tensor cannot be traced into the Dynamo FX graph output"* (`gb0208`). Without `fullgraph` it does not raise — it silently shreds the graph:

```
crop_if_exceeds=True  non-fullgraph -> shape (1, 1, 4, 5), graph_breaks=0
crop_if_exceeds=False non-fullgraph -> shape (1, 1, 6, 6), graph_breaks=5
   reason: torch.* op returned non-Tensor ... target: <function sym_max>
```

The plain `max()` version you had in `39c50fd` has none of these problems, in any mode:

```
builtin max, crop_if_exceeds=True  -> shape (1, 1, 4, 5), graph_breaks=0
builtin max, crop_if_exceeds=False -> shape (1, 1, 6, 6), graph_breaks=0
builtin max, fullgraph dynamic=None crop_if_exceeds=False -> (1, 1, 6, 6)
builtin max, dynamic=True  -> OK    dynamic=False -> OK    export static -> OK
```

And under dynamic shapes, where the bot's concern would actually bite, `max()` is fine too — `torch.export.export(..., dynamic_shapes={h: Dim, w: Dim})` captures cleanly with either, and a 6-shape dynamic compile sequence needed 3 recompiles with `max()` versus a hard failure with `sym_max`. The guard-specialization cost the bot warned about is real but tiny; the cost you paid to avoid it is a graph break on the default path.

This matters beyond the microbenchmark: `docs/source/augmentation.rst:70` names `PadTo` specifically as one of the transforms that "capture cleanly" under `torch.export`/`torch.compile`, and `tests/augmentation/test_torch_export.py:72` pins it. This is also the **only** `torch.sym_max` call site in the entire `kornia/` tree (`grep -rn "sym_max\|sym_min" kornia/` returns just these three lines), so it is a one-off pattern with no in-repo precedent.

Please:
- restore `height_pad = max(height_pad, 0)` / `width_pad = max(width_pad, 0)`;
- delete the three-line comment at `pad.py:101-103`. As written it asserts the opposite of what I measured ("torch.sym_max keeps this clamp traceable under torch.compile/torch.export"), and a false claim baked into the source is worse than no comment. If you want to keep a note, say that the branch is a Python-level clamp on static shapes and may specialize under dynamic shapes.

Secondary: `height_pad: int` is declared at `pad.py:98` and then reassigned from `torch.sym_max`, whose return is not an `int` under symbolic shapes. Reverting moots this.

### 2. Nothing in the test suite pins the default. The PR's entire backward-compat argument is unverified by CI.

I flipped `crop_if_exceeds: bool = True` to `False` at `pad.py:79` and ran the whole augmentation suite:

```
$ pytest tests/augmentation -q --device=cpu --dtype=float32     # with the default flipped to False
2321 passed, 3226 skipped, 24 xfailed, 5 xpassed, 8 warnings in 19.74s
```

Zero failures. `test_smoke` (`test_augmentation.py:4738`) uses a 2x2 input, which is smaller than the target and therefore blind to the flag, and `test_crop_if_exceeds_true_crops_larger_input` (`:4745`) passes `crop_if_exceeds=True` **explicitly**, so it tests the branch but not the default. A future refactor can flip the default and CI will stay green.

Relatedly, all three new tests fail on the merge-base for the same uninteresting reason — they only prove the kwarg exists:

```
$ pytest tests/augmentation/test_augmentation.py::TestPadTo -q   # merge-base 9ec66f3 + PR's test file
E  TypeError: PadTo.__init__() got an unexpected keyword argument 'crop_if_exceeds'
6 failed, 2 passed
```

(For contrast, on head: `8 passed`.)

Please add one test that constructs `PadTo(size=(4, 5))` with **no** `crop_if_exceeds` argument, feeds it an oversized input, and asserts `out.shape == (1, 1, 4, 5)` and `assert_close(out, img[..., :4, :5])`. That is the test that makes "the default preserves today's behavior" a CI guarantee rather than a claim in the PR body.

### 3. The new branch needs a compile/export pin.

`tests/augmentation/test_torch_export.py` already parametrizes `PadTo` (`_pad_to` at `:63`, entry at `:72`) and `TestPadTo` has no `test_dynamo` at all, so blocker 1 would have shipped without a single red light. Once you revert to `max()`, add:

- an entry to `TORCH_EXPORT_DETERMINISTIC` such as `("PadTo[no-crop]", lambda: K.PadTo((40, 40), crop_if_exceeds=False))` — note the input in that test is 32x32 so you want a size that exercises the clamp, e.g. `K.PadTo((16, 16), crop_if_exceeds=False)`;
- a `test_dynamo` on `TestPadTo` using the injected `torch_optimizer` fixture (per `TESTING.md`) covering both flag values, so a future `sym_max`-style change fails loudly.

## Nits (non-blocking)

**(a) `flags["crop_if_exceeds"]` raises `KeyError` on a module unpickled from a pre-PR kornia.** `state_dict` is empty so `load_state_dict` is unaffected, and a same-version pickle round-trips fine — but a whole-module `torch.save` from 0.8.x loaded under this version carries the old three-key `flags` dict:

```
--- flags dict missing the key (old pickle simulation) ---
RAISED: KeyError 'crop_if_exceeds'
```

`flags.get("crop_if_exceeds", True)` costs nothing and keeps the historical default for that path. Copilot flagged this and it is the one bot finding on this PR that is correct.

**(b) The docstring still implies the output is always `size`.** The class summary is "Pad the given sample to a specific size" and the `Shape:` block at `pad.py:43-45` says nothing about the flag. With `crop_if_exceeds=False` the output is `max(input, size)` per axis, which is a different shape contract and will surprise anyone collating into a fixed-size batch. One `.. note::` — "with ``crop_if_exceeds=False`` the output is at least ``size`` on each axis and may be larger; the output shape is then input-dependent" — closes it. The `Args` entry at `:39-41` is accurate but a reader scanning `Shape:` will miss it.

**(c) The new doctest at `pad.py:67-69` checks only `.shape`.** Printing the tensor (as the example above it does) would show that the content is untouched, not just the size.

**(d) Follow-up, not this PR:** in the #3518 repro above, `crop_if_exceeds=True` crops the image to 1000x1000 but leaves the box at `[700, 1000, 800, 1100]`, i.e. outside the image. That is the pre-existing `# TODO: It is incorrect to return identity` at `pad.py:89` and it is out of scope here, but it is arguably the sharper half of the original bug report. Worth a separate issue.

## Repair-window classification

**Additive — no repair window needed.** The default is byte-identical to the merge-base across the full input matrix (max abs diff `0.0`, no shape differences), so this lands on the normal path once the three blockers are addressed. Nice, careful work on the design and the augmentation plumbing; the only thing that went wrong here is trusting a static-analysis bot's compile claim over a measurement.

victorwon2001:
**kornia/augmentation/_2d/geometric/pad.py:100** — `crop_if_exceeds=False` currently clamps shape-derived padding with `torch.sym_max`, which regresses `torch.compile` in the default/static compile modes.

The supplied head measurements report that `torch.compile(fullgraph=True)` raises `Unsupported` for the no-crop path with `dynamic=None` and `dynamic=False`, while non-fullgraph compilation incurs five graph breaks instead of zero.

The head diff contains the `torch.sym_max` calls. The supplied maintainer measurements on head `39b65069f75a998a42352e232b68bfbeacd72a2f` with torch 2.9.1 report fullgraph failures for `dynamic=None` and `dynamic=False`, five graph breaks without fullgraph, and successful zero-graph-break behavior with the builtin `max()` variant.

**tests/augmentation/test_augmentation.py:4745** — No added regression test exercises the historical `crop_if_exceeds=True` default by omitting the argument on an oversized input.

The constructor default can be changed accidentally without the augmentation suite detecting the backward-incompatible behavior change.

All three newly added tests pass `crop_if_exceeds` explicitly. The supplied maintainer verification reports that changing the constructor default to `False` still left the complete augmentation suite green.

`file:kornia/augmentation/_2d/geometric/pad.py:diff`

> +        self,
> +        size: Tuple[int, int],
> +        pad_mode: str = "constant",
> +        pad_value: float = 0,
> +        keepdim: bool = False,
> +        crop_if_exceeds: bool = True,
>      ) -> None:
>          super().__init__(p=1.0, same_on_batch=True, p_batch=1.0, keepdim=keepdim)
> -        self.flags = {"size": size, "pad_mode": pad_mode, "pad_value": pad_value}
> +        self.flags = {
> +            "size": size,
> +            "pad_mode": pad_mode,
> +            "pad_value": pad_value,
> +            "crop_if_exceeds": crop_if_exceeds,
> +        }

`file:kornia/augmentation/_2d/geometric/pad.py:diff`

> +        if not flags["crop_if_exceeds"]:
> +            # torch.sym_max keeps this clamp traceable under torch.compile/torch.export:
> +            # height_pad/width_pad are derived from input.shape and may be SymInt, and the
> +            # builtin max() would force a guard/specialization on their sign.
> +            height_pad = torch.sym_max(height_pad, 0)
> +            width_pad = torch.sym_max(width_pad, 0)

`file:tests/augmentation/test_augmentation.py:diff`

> +    def test_crop_if_exceeds_true_crops_larger_input(self, device, dtype):
> +        img = torch.rand(1, 1, 6, 6, device=device, dtype=dtype)
> +        aug = PadTo(size=(4, 5), crop_if_exceeds=True)
> +        out = aug(img)
> +        assert out.shape == (1, 1, 4, 5)
> +        self.assert_close(out, img[..., :4, :5])
> +
> +    def test_crop_if_exceeds_false_keeps_larger_input_unchanged(self, device, dtype):
> +        img = torch.rand(1, 1, 6, 6, device=device, dtype=dtype)
> +        aug = PadTo(size=(4, 5), crop_if_exceeds=False)
> +        out = aug(img)
> +        assert out.shape == (1, 1, 6, 6)
> +        self.assert_close(out, img)
> +        self.assert_close(aug.inverse(out), img)
> +
> +    def test_crop_if_exceeds_false_mixed_dims(self, device, dtype):
> +        # height (6) exceeds target (4); width (2) is below target (5)
> +        img = torch.rand(1, 1, 6, 2, device=device, dtype=dtype)
> +        aug = PadTo(size=(4, 5), crop_if_exceeds=False)
> +        out = aug(img)
> +        assert out.shape == (1, 1, 6, 5)
> +        # the exceeding height dimension is left untouched, and the original
> +        # content is preserved with zeros padded onto the short width dimension
> +        self.assert_close(out[..., :2], img)
> +        self.assert_close(out[..., 2:], torch.zeros(1, 1, 6, 3, device=device, dtype=dtype))

`comment:5455954222`

> ## Blockers
> 
> ### 1. `torch.sym_max` at `kornia/augmentation/_2d/geometric/pad.py:100-105` makes `torch.compile` strictly worse. Revert to builtin `max()`.

`comment:5455954222`

> ### 2. Nothing in the test suite pins the default. The PR's entire backward-compat argument is unverified by CI.
> 
> I flipped `crop_if_exceeds: bool = True` to `False` at `pad.py:79` and ran the whole augmentation suite:
> 
> ```
> $ pytest tests/augmentation -q --device=cpu --dtype=float32     # with the default flipped to False
> 2321 passed, 3226 skipped, 24 xfailed, 5 xpassed, 8 warnings in 19.74s
> ```
> 
> Zero failures. `test_smoke` (`test_augmentation.py:4738`) uses a 2x2 input, which is smaller than the target and therefore blind to the flag, and `test_crop_if_exceeds_true_crops_larger_input` (`:4745`) passes `crop_if_exceeds=True` **explicitly**, so it tests the branch but not the default. A future refactor can flip the default and CI will stay green.
> 
> Relatedly, all three new tests fail on the merge-base for the same uninteresting reason — they only prove the kwarg exists:
> 
> ```
> $ pytest tests/augmentation/test_augmentation.py::TestPadTo -q   # merge-base 9ec66f3 + PR's test file
> E  TypeError: PadTo.__init__() got an unexpected keyword argument 'crop_if_exceeds'
> 6 failed, 2 passed
> ```
> 
> (For contrast, on head: `8 passed`.)
> 
> Please add one test that constructs `PadTo(size=(4, 5))` with **no** `crop_if_exceeds` argument, feeds it an oversized input, and asserts `out.shape == (1, 1, 4, 5)` and `assert_close(out, img[..., :4, :5])`. That is the test that makes "the default preserves today's behavior" a CI guarantee rather than a claim in the PR body.
> 
> ### 3. The new branch needs a compile/export pin.
> 
> `tests/augmentation/test_torch_export.py` already parametrizes `PadTo` (`_pad_to` at `:63`, entry at `:72`) and `TestPadTo` has no `test_dynamo` at all, so blocker 1 would have shipped without a single red light. Once you revert to `max()`, add:
> 
> - an entry to `TORCH_EXPORT_DETERMINISTIC` such as `("PadTo[no-crop]", lambda: K.PadTo((40, 40), crop_if_exceeds=False))` — note the input in that test is 32x32 so you want a size that exercises the clamp, e.g. `K.PadTo((16, 16), crop_if_exceeds=False)`;
> - a `test_dynamo` on `TestPadTo` using the injected `torch_optimizer` fixture (per `TESTING.md`) covering both flag values, so a future `sym_max`-style change fails loudly.

`github_actions_job:105153021227:annotation:message`

> Process completed with exit code 1.
> no changed changelog.d/3964.(added|fixed|breaking).md fragment found; add one for a user-visible change or apply the no-changelog label for an internal or release PR

`context:CONTRIBUTING.md`

> **CI checks:**
> - All tests pass
> - Test coverage maintained
> - Type checking (ty)
> - Documentation builds successfully
> - Code formatting (ruff, docformatter via pre-commit)
> 
> Fix any failing checks before your PR will be considered.

head_content truncated at 50000 characters: tests/augmentation/test_augmentation.py

base_content truncated at 50000 characters: tests/augmentation/test_augmentation.py

head_content truncated at 50000 characters: tests/augmentation/test_augmentation.py

base_content truncated at 50000 characters: tests/augmentation/test_augmentation.py

Replace the `torch.sym_max` clamp with the measured working builtin `max()` implementation, add an oversized-input regression test that constructs `PadTo(size=(4, 5))` without passing `crop_if_exceeds`, add compile/export coverage that exercises `crop_if_exceeds=False` and both flag values where appropriate, and satisfy the failing changelog-fragment check before rerunning the affected validation.

SHA: `39b65069f75a998a42352e232b68bfbeacd72a2f`

ducha-aiki (changes requested):
Recording the explicit changes-requested verdict from the linked review at this head. The remaining blockers are the torch.sym_max compile regression, missing implicit-default regression coverage, and compile/export coverage for the new branch. This records that completed review; it is not a new validation run.

[Completed review](https://github.com/kornia/kornia/pull/3964#issuecomment-5455954222).

PR #3965, opened 2026-08-11
fix(geometry): accept arbitrary batch ranks in axis_angle_to_rotation_matrix

### LINKED ISSUE #3955: axis_angle_to_rotation_matrix accepts only (N, 3) although its own guard says (*, 3); aa2R(R2aa(R)) fails for every rank but 3
`axis_angle_to_rotation_matrix` accepts **only rank-2 `(N, 3)` input**, but its own shape guard advertises `(*, 3)`, its docstring says `(N, 3)`, and every sibling in the file accepts full `(*, 3)`. Inputs of any other rank get past the guard and then fail deep inside with a raw `IndexError`/`ValueError` from `unbind`. The asymmetry also breaks composition with `rotation_matrix_to_axis_angle`, which *does* accept any rank ≥ 2 — so `axis_angle_to_rotation_matrix(rotation_matrix_to_axis_angle(R))` fails for every rank except 3, including for the shape produced by `rotation_matrix_to_axis_angle`'s own doctest.

The guard only checks the last dimension (`kornia/geometry/conversions.py:429-430`):

```python
if not axis_angle.shape[-1] == 3:
    raise ValueError(f"Input size must be a (*, 3) tensor. Got {axis_angle.shape}")
```

while the body hard-codes rank 2 (`:435-436`):

```python
wxyz = axis_angle / (theta.unsqueeze(-1) + eps)  # (B, 3)
wx, wy, wz = wxyz.unbind(dim=1)  # (B,)
```

`unbind(dim=1)` on a `(3,)` tensor is an out-of-range dim; on a `(2, 5, 3)` tensor it unbinds the wrong axis and yields 5 values instead of 3.

## Repro

```python
import torch
    axis_angle_to_quaternion as aa2q,
    axis_angle_to_rotation_matrix as aa2R,
    rotation_matrix_to_axis_angle as R2aa,
)

for shape in [(3,), (2, 3), (0, 3), (2, 5, 3), (1, 1, 3)]:
    try:
        out = tuple(aa2R(torch.zeros(*shape)).shape)
    except Exception as e:
        out = f"{type(e).__name__}: {e}"
    print(f"aa2R(zeros{shape}) -> {out}")

try:
    aa2R(torch.zeros(1, 4))
except Exception as e:
    print(f"\nthe guard's own message: {type(e).__name__}: {e}")

print("\n# every sibling accepts the same shapes:")
for shape in [(3,), (2, 3), (2, 5, 3)]:
    print(f"  aa2q(zeros{shape}) -> {tuple(aa2q(torch.zeros(*shape)).shape)}")

print("\n# so aa2R(R2aa(R)) does not compose for any rank but 3:")
for shape in [(3, 3), (1, 3, 3), (2, 5, 3, 3)]:
    R = torch.eye(3).expand(*shape).contiguous()
    aa = R2aa(R)
    try:
        out = tuple(aa2R(aa).shape)
    except Exception as e:
        out = f"{type(e).__name__}: {e}"
    print(f"  R{tuple(R.shape)} -> R2aa -> {tuple(aa.shape)} -> aa2R -> {out}")
```

## Observed

```
aa2R(zeros(3,)) -> IndexError: Dimension out of range (expected to be in range of [-1, 0], but got 1)
aa2R(zeros(2, 3)) -> (2, 3, 3)
aa2R(zeros(0, 3)) -> (0, 3, 3)
aa2R(zeros(2, 5, 3)) -> ValueError: too many values to unpack (expected 3)
aa2R(zeros(1, 1, 3)) -> ValueError: not enough values to unpack (expected 3, got 1)

the guard's own message: ValueError: Input size must be a (*, 3) tensor. Got torch.Size([1, 4])

# every sibling accepts the same shapes:
  aa2q(zeros(3,)) -> (4,)
  aa2q(zeros(2, 3)) -> (2, 4)
  aa2q(zeros(2, 5, 3)) -> (2, 5, 4)

# so aa2R(R2aa(R)) does not compose for any rank but 3:
  R(3, 3) -> R2aa -> (3,) -> aa2R -> IndexError: Dimension out of range (expected to be in range of [-1, 0], but got 1)
  R(1, 3, 3) -> R2aa -> (1, 3) -> aa2R -> (1, 3, 3)
  R(2, 5, 3, 3) -> R2aa -> (2, 5, 3) -> aa2R -> ValueError: too many values to unpack (expected 3)
```

Note the last block: `rotation_matrix_to_axis_angle`'s **first doctest** passes an unbatched `(3, 3)` matrix and returns `tensor([0., 0., 0.])` of shape `(3,)` — feeding that documented output straight back into `axis_angle_to_rotation_matrix` raises `IndexError`.

## Expected

A maintainer decision between two coherent contracts; the current state is neither:

1. **Widen the implementation** to full `(*, 3)`, matching `axis_angle_to_quaternion`, `quaternion_to_axis_angle`, `quaternion_to_rotation_matrix`, `rotation_matrix_to_quaternion`, `rotation_matrix_to_axis_angle`, `quaternion_log_to_exp` and `quaternion_exp_to_log` — all of which already do. Mechanically this is `unbind(dim=1)` → `unbind(dim=-1)` plus a reshape of the stacked result, and it makes `aa2R(R2aa(R))` compose at every rank. This is my recommendation: it is what the guard already promises, what the siblings do, and what makes the inverse pair round-trip.
2. **Narrow the guard** to reject anything that is not exactly rank 2, with a message that says `(N, 3)` and names the received shape — so at least the failure is a kornia error at the boundary instead of a torch internal error from three frames down.

Either way the guard message (`Input size must be a (*, 3) tensor`) is currently false and must change, and the raw `IndexError: Dimension out of range` / `ValueError: too many values to unpack` leaks should not be what a user sees.

## Classification

Mixed:

- The **error-message half** is docs-wrong and can be fixed routinely (the message contradicts the implementation).
- The **rank contract half** is a semantic change under option 1: widening acceptance turns currently-raising inputs into successful calls. That is additive rather than breaking — no input that works today changes its output — so it is a low-risk semantic widening rather than a default change. Per the reference-core lifecycle (Part I / Part III §2) it needs a maintainer scope confirmation but does not obviously need to wait for the coordinated Repair & Deprecation window, since nothing observable regresses. Option 2 is the conservative alternative and changes only error types.

## Environment

kornia 0.9.0rc1 @ `3358ef16`, torch 2.9.1, cpu. Executed from a worktree carrying the batch-3a docs/tests commits on top of `3358ef16`; `axis_angle_to_rotation_matrix` and `rotation_matrix_to_axis_angle` are byte-identical to `origin/main` there (no diff hunk lands in either; function bodies compare AST-identical).

abhi-0203:
## Summary

Fixes `axis_angle_to_rotation_matrix` to accept arbitrary batch ranks `(*, 3)` instead of only rank-2 `(N, 3)`, matching the documented guard and sibling functions.

### Problem

The function's shape guard checks `axis_angle.shape[-1] == 3` and raises `ValueError("Input size must be a (*, 3)")`, but the body uses `unbind(dim=1)`, `stack(dim=1)`, and `.view(-1, 3, 3)` — all of which assume exactly one batch dimension. Inputs of any other rank fail with `IndexError`/`ValueError` from deep inside the computation.

This also breaks composition: `axis_angle_to_rotation_matrix(rotation_matrix_to_axis_angle(R))` fails for every rank except 3, including for shapes produced by `rotation_matrix_to_axis_angle`'s own doctest.

### Fix

1. `unbind(dim=1)` → `unbind(dim=-1)` — split along the last axis (rank-agnostic)
2. `stack(dim=1)` → `stack(dim=-2)` — stack matrix rows before the last dim
3. `.view(-1, 3, 3)` → `.reshape(*batch_shape, 3, 3)` — preserve batch shape in Taylor fallback
4. `mask.view(-1, 1, 1)` → `.reshape(*batch_shape, 1, 1)` — preserve batch shape for broadcasting

### Test results

```python
>>> aa2R(torch.zeros(3,)).shape
torch.Size([3, 3])
>>> aa2R(torch.zeros(2, 3)).shape
torch.Size([2, 3, 3])
>>> aa2R(torch.zeros(0, 3)).shape
torch.Size([0, 3, 3])
>>> aa2R(torch.zeros(2, 5, 3)).shape
torch.Size([2, 5, 3, 3])
>>> aa2R(torch.zeros(1, 1, 3)).shape
torch.Size([1, 1, 3, 3])
```

Identity correctness verified: `aa2R(zeros(*shape))` produces identity matrices for all ranks.

Closes #3955

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -292,10 +292,9 @@ def axis_angle_to_rotation_matrix(axis_angle: torch.Tensor) -> torch.Tensor:
     r"""Convert 3d vector of axis-angle rotation to 3x3 rotation matrix.
 
     Args:
-        axis_angle: tensor of 3d vector of axis-angle rotations in radians with shape :math:`(N, 3)`.
-
+        axis_angle: tensor of 3d vector of axis-angle rotations in radians with shape :math:`(*, 3)`.
     Returns:
-        tensor of rotation matrices of shape :math:`(N, 3, 3)`.
+        tensor of rotation matrices of shape :math:`(*, 3, 3)`.
 
     Example:
         >>> input = torch.tensor([[0., 0., 0.]])
@@ -319,8 +318,8 @@ def axis_angle_to_rotation_matrix(axis_angle: torch.Tensor) -> torch.Tensor:
 
     def _compute_rotation_matrix(axis_angle: torch.Tensor, theta2: torch.Tensor, eps: float = 1e-6) -> torch.Tensor:
         theta = torch.sqrt(theta2.clamp(min=1e-12))  # clamping to ensure no nan gradients
-        wxyz = axis_angle / (theta.unsqueeze(-1) + eps)  # (B, 3)
-        wx, wy, wz = wxyz.unbind(dim=1)  # (B,)
+        wxyz = axis_angle / (theta.unsqueeze(-1) + eps)
+        wx, wy, wz = wxyz.unbind(dim=-1)
 
         cos_theta = torch.cos(theta)
         sin_theta = torch.sin(theta)
@@ -348,12 +347,13 @@ def axis_angle_to_rotation_matrix(axis_angle: torch.Tensor) -> torch.Tensor:
                 torch.stack([r10, r11, r12], dim=-1),
                 torch.stack([r20, r21, r22], dim=-1),
             ],
-            dim=1,
+            dim=-2,
         )
 
         return rot
 
     def _compute_rotation_matrix_taylor(axis_angle: torch.Tensor) -> torch.Tensor:
+        batch_shape = axis_angle.shape[:-1]
         rx, ry, rz = axis_angle.unbind(-1)
         k_one = torch.ones_like(rx)
 
@@ -370,16 +370,17 @@ def axis_angle_to_rotation_matrix(axis_angle: torch.Tensor) -> torch.Tensor:
                 k_one,
             ],
             dim=-1,
-        ).view(-1, 3, 3)
+        ).reshape(*batch_shape, 3, 3)
 
         return rot
 
     theta2 = (axis_angle * axis_angle).sum(dim=-1)
 
-    rot_normal = _compute_rotation_matrix(axis_angle, theta2)  # (N,3,3)
-    rot_taylor = _compute_rotation_matrix_taylor(axis_angle)  # (N,3,3)
+    rot_normal = _compute_rotation_matrix(axis_angle, theta2)
+    rot_taylor = _compute_rotation_matrix_taylor(axis_angle)
 
-    mask = (theta2 > 1e-6).view(-1, 1, 1)  # shape (N,1,1)
+    batch_shape = axis_angle.shape[:-1]
+    mask = (theta2 > 1e-6).reshape(*batch_shape, 1, 1)
 
     rotation_matrix = torch.where(mask, rot_normal, rot_taylor)
 
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -293,6 +293,7 @@ def axis_angle_to_rotation_matrix(axis_angle: torch.Tensor) -> torch.Tensor:
 
     Args:
         axis_angle: tensor of 3d vector of axis-angle rotations in radians with shape :math:`(*, 3)`.
+
     Returns:
         tensor of rotation matrices of shape :math:`(*, 3, 3)`.
 


ducha-aiki (changes requested):
The `kornia/` change is correct and the generalization is clean — `unbind(dim=-1)`, `stack(dim=-2)` and a `reshape` off `axis_angle.shape[:-1]` is the right way to do this, and it costs nothing at the shapes that already worked. But the PR is incomplete: it changes no tests, and `#3955` is a pinned issue, so four pins flip the moment this lands.

Verified by applying your diff to current `main` (`35555b55`).

## The change works, and is bitwise safe

```text
shape        main                                          this PR
(6, 3)       (6, 3, 3)                                     (6, 3, 3)  -- bitwise identical
(3,)         IndexError: Dimension out of range            (3, 3)
(2, 5, 3)    ValueError: too many values to unpack         (2, 5, 3, 3)
(1, 1, 3)    ValueError: not enough values to unpack       (1, 1, 3, 3)
(0, 3)       (0, 3, 3)                                     (0, 3, 3)
```

The `(6, 3)` output is bitwise identical to `main` — I compared raw float64 words, not tolerances — so no existing caller can move. And the property that really matters for a batch-rank generalization holds:

```text
aa2R(x.reshape(2, 4, 3))  ==  aa2R(x.reshape(-1, 3)).reshape(2, 4, 3, 3)     bitwise True
```

That is the guarantee someone will actually rely on, and it is currently untested. Please pin it — a rank-3 input and its flattened rank-2 equivalent compared with `torch.equal`. It is a stronger statement than any shape assertion.

`torch.jit.script` fails on this function on `main` already (nested `def`s are unsupported), so your `.reshape(*batch_shape, 3, 3)` does not regress scripting. Worth a comment near it anyway: star-unpacking a shape is not TorchScript-expressible, so if this function is ever made scriptable that line is the one that has to change.

## Required: four pins flip

`tests/geometry` with your diff applied gives 4 failures against 0 on `main`:

**1. `TestAngleAxisToRotationMatrix::test_convention_accepts_any_leading_batch_dimensions_3955` — `[XPASS(strict)]`:**

```text
[XPASS(strict)] axis_angle_to_rotation_matrix accepts only rank-2 (N, 3) input despite its own
guard message saying (*, 3) — kornia#3955
```

This is the machinery working. **Drop the `xfail` marker and keep the test** — it is written to become your regression guard, and it already covers `(3,)`, `(2, 5, 3)` and both round-trip directions through `rotation_matrix_to_axis_angle`.

**2-4. `test_wart_only_rank_2_input_is_accepted_3955[unbatched | extra_batch_dim | singleton_extra_batch_dim]`** — these pin the three exception messages your change removes. **Delete the whole test.**

Then retire `_runs_without_raising` at `tests/geometry/test_conversions.py:73`. Its own comment says "Retire this helper together with the `#3955` pins", and once the `xfail` is gone the call sites can assert directly instead of through a boolean.

## Required: the docstring your fix falsifies

`kornia/geometry/conversions.py:459-467`, inside `axis_angle_to_rotation_matrix`'s own docstring:

```rst
.. warning::
    Only rank-2 input is accepted, despite the guard's ``(*, 3)`` message:
    ``(3,)`` raises ``IndexError: Dimension out of range``, ``(2, 5, 3)``
    raises ``ValueError: too many values to unpack (expected 3)`` and
    ``(1, 1, 3)`` raises ``ValueError: not enough values to unpack``.
    Composing with rotation_matrix_to_axis_angle therefore fails for every
    rotation-matrix rank but 3. Tracked in #3955.
```

Every sentence there is false after your change — I reproduced all three exceptions on `main` and all three shapes now succeed. The whole block goes, and `grep -rnE "#3955|issues/3955" kornia/` must come back empty afterwards. Your `Args`/`Returns` update from `(N, 3)` to `(*, 3)` was right; this is the other half of it.

## Rebase

Your base is `3358ef16`, which is older than every batch-3 merge — the pins above did not exist when you branched, which is exactly why you could not have known about them. CI last ran 2026-08-11. Please rebase onto current `main`; the `kornia/` diff applies cleanly, so the pin and docstring edits are the only real work.

Requesting changes on the pins and the docstring. The code change itself is correct and I want it in — `maintainer_can_modify` is on if you would rather I push the rebase and you take it from there.

---

Issue #3952, opened 2026-08-10
normalize_quaternion returns a non-unit quaternion below eps ([1e-13,0,0,0] -> [0.1,0,0,0]), so quaternion_to_rotation_matrix(zeros(4)) is silently the identity


ducha-aiki:
`normalize_quaternion` is `return F.normalize(quaternion, p=2.0, dim=-1, eps=eps)` (`kornia/geometry/conversions.py:643`), and `F.normalize` divides by `max(||q||, eps)`. So a quaternion whose norm is below the default `eps = 1e-12` comes back **scaled, not normalized** — silently, with no error and no warning:

```python
>>> normalize_quaternion(torch.tensor([1e-13, 0., 0., 0.], dtype=torch.float64)).tolist()
[0.1, 0.0, 0.0, 0.0]        # ||.|| = 0.1, not 1
>>> normalize_quaternion(torch.zeros(4)).tolist()
[0.0, 0.0, 0.0, 0.0]        # ||.|| = 0
```

`1e-13 / 1e-12 = 0.1` — the division by `eps` instead of by the norm is applied verbatim. The function's only documented contract is "Normalize a quaternion" returning "the normalized quaternion", and its `eps` is documented as "small value to avoid division by zero".

`quaternion_to_rotation_matrix` calls `normalize_quaternion` first, so it inherits the defect one level up: the **zero quaternion silently yields the identity matrix**, and so does any sub-eps quaternion.

```python
>>> quaternion_to_rotation_matrix(torch.zeros(4))
tensor([[1., 0., 0.],
        [0., 1., 0.],
        [0., 0., 1.]])
```

## Repro

```python
import torch
from kornia.geometry.conversions import (
    normalize_quaternion as nq,
    quaternion_to_rotation_matrix as q2R,
)

tiny = torch.tensor([1e-13, 0.0, 0.0, 0.0], dtype=torch.float64)   # ||q|| = 1e-13 < eps = 1e-12
print("nq([1e-13,0,0,0])          =", nq(tiny).tolist(), " ||.|| =", nq(tiny).norm().item())
print("nq(zeros(4))               =", nq(torch.zeros(4)).tolist())
print("nq(zeros(4), eps=0.0)      =", nq(torch.zeros(4), eps=0.0).tolist())
print("float32 too                :", nq(torch.tensor([1e-13, 0.0, 0.0, 0.0])).tolist())

# the consequence one level up
[...]

PR #3969, opened 2026-08-12
fix(geometry): dtype-aware eps prevents float16 NaN in quaternion log/exp/normalize

### LINKED ISSUE #3966: quaternion_exp_to_log returns NaN in float16 for any zero-vector-part quaternion (identity included): the default eps=1e-8 underflows to 0
### Describe the bug

`quaternion_exp_to_log` returns `NaN` in `float16` for **every** quaternion whose vector part is zero — including the identity quaternion `(1, 0, 0, 0)`, whose log is unambiguously the origin. `float32`, `float64` and `bfloat16` all return `[0, 0, 0]` for the same inputs.

The cause is a dtype-dependent underflow of the default `eps`, not the quaternion maths. The implementation guards the division with

```python
norm_q = torch.norm(quaternion_vector, p=2, dim=-1, keepdim=True).clamp(min=eps)
```

with `eps: float = 1e-8` (`kornia/geometry/conversions.py`, `quaternion_exp_to_log`). `float16`'s smallest subnormal is ≈ 6e-8, so `1e-8` rounds to exactly `0.0` and the clamp becomes a no-op. `acos(±1)` is then multiplied by a zero vector part and divided by a zero norm: `0 / 0 → NaN`.

`bfloat16` is unaffected because its exponent range is float32's, so `1e-8` is representable there.

### Reproduction steps

```python
import torch
from kornia.geometry.conversions import quaternion_exp_to_log

    q = torch.tensor([1.0, 0.0, 0.0, 0.0], dtype=dtype)   # identity quaternion
    print(dtype, quaternion_exp_to_log(q).tolist())
```

```
torch.float64  [0.0, 0.0, 0.0]
torch.float32  [0.0, 0.0, 0.0]
torch.bfloat16 [0.0, 0.0, 0.0]
torch.float16  [nan, nan, nan]
```

It is not specific to the identity — any zero vector part reproduces it, in both halves of the double cover:

```python
q = torch.tensor([-1.0, 0.0, 0.0, 0.0], dtype=torch.float16)
quaternion_exp_to_log(q)          # tensor([nan, nan, nan], dtype=torch.float16)
```

```python
q = torch.tensor([1.0, 0.0, 0.0, 0.0], dtype=torch.float16)
quaternion_exp_to_log(q, eps=1e-3)          # tensor([0., 0., 0.], dtype=torch.float16)
float(torch.tensor(1e-8, dtype=torch.float16))   # 0.0      <- the default underflows
float(torch.tensor(1e-8, dtype=torch.bfloat16))  # 1.0011717677116394e-08
```

A non-zero vector part is fine at the same dtype, which is why the suite has never caught it:

```python
quaternion_exp_to_log(torch.tensor([-1.0, 1e-3, 0.0, 0.0], dtype=torch.float16))
# tensor([3.1406, 0.0000, 0.0000], dtype=torch.float16)
```

### Expected behavior

`quaternion_exp_to_log` should return the origin `[0, 0, 0]` for a quaternion with zero vector part at every floating dtype it accepts, as it already does at `float64`, `float32` and `bfloat16`. Whatever the fix, a float-typed default that silently rounds to zero in a supported dtype is a trap: candidates are a dtype-aware epsilon (`torch.finfo(q.dtype).tiny` or `eps`), clamping the norm after casting, or selecting the origin explicitly where the vector part is zero.

### Environment

- kornia 0.9.0rc1 (`main` at 8454a81a), torch 2.9.1, CPU, macOS (arm64). Reproduced on MPS as well.

### Notes

- Found while auditing the rotation-representation conversions for per-op Convention blocks (batch 3b of #3941). It is the same failure *class* as the float16 eps underflow already pinned for the pixel-coordinate normalizers in #3942 (`test_wart_float16_underflowed_default_eps_flips_branches`) and adjacent to #3939 — a float default epsilon that is not representable in every supported dtype. A repo-wide sweep for `clamp(min=eps)` with a float default under float16 is probably worth doing once, rather than one issue per site.
- Distinct from #3953 (`quaternion_exp_to_log` silently accepting non-unit input) and from #3949 (NaN *gradients* at the identity in `quaternion_to_axis_angle` / `axis_angle_to_quaternion`): this is a NaN in the **forward** output, at a valid unit quaternion, in float16 only.

TejaashriUK:
🟢 Green lane (test-first bug fix, verified docs fix, help wanted issue, benchmark results: no issue needed)
- [x] Green lane
- [ ] Discuss-first lane

Fixes #3966

## Description
`quaternion_exp_to_log`, `quaternion_log_to_exp`, and `normalize_quaternion` in
`kornia/geometry/conversions.py` each guard a division by clamping a norm with
a Python `float` default for `eps` (`1e-8` / `1e-12`). In `float16`, these
defaults underflow to exactly `0.0` (float16's smallest subnormal is
~5.96e-08), so the clamp becomes a no-op and the division becomes `0 / 0`,
returning `NaN`.

This reproduces for the identity quaternion — the simplest possible input —
and any quaternion with a zero vector part, in both `quaternion_exp_to_log`
and `quaternion_log_to_exp`, and for near-zero/zero quaternions in
`normalize_quaternion` (which is also the root cause of the NaN reported in
#3952, since `quaternion_to_rotation_matrix` calls it internally).

`float32`, `float64`, and `bfloat16` are unaffected, since `1e-8` is
representable in all three.

## Fix
Added a small helper, `_dtype_safe_eps(eps, dtype)`, which returns
`max(eps, torch.finfo(dtype).tiny)` — i.e. it only raises `eps` when the
requested value isn't representable in the tensor's dtype, and otherwise
leaves it untouched. This follows the same dtype-aware pattern already used
by `safe_zero_division` inside `rotation_matrix_to_quaternion` in this file,
rather than introducing a new approach.

Applied at the three division-guard sites:
- `normalize_quaternion`: `F.normalize(..., eps=_dtype_safe_eps(eps, quaternion.dtype))`
- `quaternion_log_to_exp`: norm clamp uses `_dtype_safe_eps(eps, quaternion.dtype)`
- `quaternion_exp_to_log`: norm clamp uses `_dtype_safe_eps(eps, quaternion_vector.dtype)`

No public API changes — function signatures are unchanged.

## Testing
Verified against every case from #3966 and the linked #3952:

```python
import torch
    quaternion_exp_to_log, quaternion_log_to_exp, normalize_quaternion
)

quaternion_exp_to_log(torch.tensor([1., 0., 0., 0.], dtype=torch.float16))
# now: tensor([0., 0., 0.])  (was: tensor([nan, nan, nan]))

quaternion_exp_to_log(torch.tensor([-1., 0., 0., 0.], dtype=torch.float16))
# now: tensor([0., 0., 0.])  (was: NaN)

quaternion_log_to_exp(torch.zeros(3, dtype=torch.float16))
# now: tensor([1., 0., 0., 0.])  (was: tensor([1., nan, nan, nan]))

normalize_quaternion(torch.zeros(4, dtype=torch.float16))
# now: tensor([0., 0., 0., 0.])  (was: all-NaN)

normalize_quaternion(torch.tensor([1e-13, 0., 0., 0.], dtype=torch.float16))
# now: tensor([0., 0., 0., 0.])  (was: all-NaN)
```

Regression checks — confirmed unchanged:
- Non-zero vector part at float16 (`[-1., 1e-3, 0., 0.]`) still returns the
  correct non-zero log map.
- `float32`, `float64`, `bfloat16` behavior identical to before the change.
- All existing doctests in the three functions pass.

## CI Evidence
All 36 CI checks passed, including the full CPU test matrix across Python
3.11–3.13, PyTorch 2.5.1/2.9.1, float32/float64, on Ubuntu/macOS/Windows —
see the Checks tab on this PR.

## AI Usage Disclosure
🔴 AI-generated — Claude (Anthropic) diagnosed the root cause, proposed the
`_dtype_safe_eps` fix following the existing `safe_zero_division` pattern
already in this file (see `rotation_matrix_to_quaternion`), and drafted the
verification tests. I reviewed the diff and understand the reasoning: eps
defaults are plain Python floats that underflow to `0.0` in float16, silently
disabling the division-by-zero guard; the fix compares eps against
`torch.finfo(dtype).tiny` (the smallest representable value for that dtype)
and only raises it when necessary, leaving float32/float64 behavior
unchanged.

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -76,6 +76,17 @@ __all__ = [
 ]
 
 
+def _dtype_safe_eps(eps: float, dtype: torch.dtype) -> float:
+    """Clamp ``eps`` so it does not underflow to zero in the given floating dtype.
+
+    A Python ``float`` default like ``1e-8`` silently rounds to ``0.0`` when cast to
+    float16, which disables the division guard it is meant to provide. Using
+    ``max(eps, torch.finfo(dtype).tiny)`` keeps the guard active in low precision
+    without changing behavior for dtypes where the default is already representable.
+    """
+    return max(eps, torch.finfo(dtype).tiny)
+
+
 def rad2deg(tensor: torch.Tensor) -> torch.Tensor:
     r"""Convert angles from radians to degrees.
 
@@ -652,7 +663,7 @@ def normalize_quaternion(quaternion: torch.Tensor, eps: float = 1.0e-12) -> torc
 
     if not quaternion.shape[-1] == 4:
         raise ValueError(f"Input must be a tensor of shape (*, 4). Got {quaternion.shape}")
-    return F.normalize(quaternion, p=2.0, dim=-1, eps=eps)
+    return F.normalize(quaternion, p=2.0, dim=-1, eps=_dtype_safe_eps(eps, quaternion.dtype))
 
 
 # based on:
@@ -816,7 +827,9 @@ def quaternion_log_to_exp(quaternion: torch.Tensor, eps: float = 1.0e-8) -> torc
         raise ValueError(f"Input must be a tensor of shape (*, 3). Got {quaternion.shape}")
 
     # compute quaternion norm
-    norm_q: torch.Tensor = torch.norm(quaternion, p=2, dim=-1, keepdim=True).clamp(min=eps)
+    norm_q: torch.Tensor = torch.norm(quaternion, p=2, dim=-1, keepdim=True).clamp(
+        min=_dtype_safe_eps(eps, quaternion.dtype)
+    )
 
     # compute scalar and vector
     quaternion_vector: torch.Tensor = quaternion * torch.sin(norm_q) / norm_q
@@ -858,7 +871,9 @@ def quaternion_exp_to_log(quaternion: torch.Tensor, eps: float = 1.0e-8) -> torc
     quaternion_vector = quaternion[..., 1:4]
 
     # compute quaternion norm
-    norm_q: torch.Tensor = torch.norm(quaternion_vector, p=2, dim=-1, keepdim=True).clamp(min=eps)
+    norm_q: torch.Tensor = torch.norm(quaternion_vector, p=2, dim=-1, keepdim=True).clamp(
+        min=_dtype_safe_eps(eps, quaternion_vector.dtype)
+    )
 
     # apply log map
     quaternion_log: torch.Tensor = (


ducha-aiki:
Heads-up before you invest more here: **four open PRs currently overlap on #3966 and #3975.** In the order they were opened — #3969 (this one), #3970, #3990, #4000. Nobody did anything wrong; the issues were filed in a burst and several people picked them up independently. Flagging it so no one keeps building on work that may be superseded.

Per the green lane in `CONTRIBUTING.md`, overlap is resolved as **first *good* PR wins** — first to meet the bar, not first to open. A maintainer arbitrates, not me.

On the bar: this PR is the earliest of the four, but it currently changes `kornia/geometry/conversions.py` only, with **no test**. The green lane is explicitly test-first, so as it stands it isn't yet eligible, and that is the single highest-value thing you could change. Your scope is also the broadest of the four (log/exp *and* `normalize`), which is an advantage if it is backed by tests.

### Pins you must update

|---|---|---|

---

ducha-aiki:
Closing this as superseded — but please read the second half, because one part of your change is **not** superseded and I'd like it back as its own PR.

### What happened

#3990 merged as `4da60edc`. It fixes both #3966 and #3975 in `quaternion_log_to_exp` / `quaternion_exp_to_log`, so those two hunks of yours are now covered on `main`:

```
quaternion_log_to_exp(zeros(3), float16)          -> [1., 0., 0., 0.]
quaternion_exp_to_log([1,0,0,0], float16)         -> [0., 0., 0.]
```

Your `kornia/` diff no longer applies to `main` as a result. This was a four-way collision — #3969, #3970, #3990 and #4000 all landed on the same two issues within days of each other, and per the green lane in `CONTRIBUTING.md` it resolved to the first *good* one. Nothing about that is a judgement on your work; you were first to open and the diagnosis in this PR was correct.

For what it's worth, your `max(eps, torch.finfo(dtype).tiny)` is the better-argued form of the guard than the `finfo(dtype).eps` variant that showed up in the parallel PR — `tiny` is the smallest normal, which is what a division guard actually wants, where `eps` is machine epsilon relative to 1.0 and overrides the caller by more than it needs to.

### The part that is not superseded

You also patched **`normalize_quaternion`**, which #3990 never touched. That fixes a bug that is still live on `main` right now:

```python
>>> normalize_quaternion(torch.zeros(4, dtype=torch.float16))
tensor([nan, nan, nan, nan], dtype=torch.float16)   # every other dtype returns zeros
```

I applied only that one line onto `4da60edc` and it does the right thing at every dtype, leaving unit input untouched:

```
float16 -> [0., 0., 0., 0.]     bfloat16 -> [0., 0., 0., 0.]
float32 -> [0., 0., 0., 0.]     float64  -> [0., 0., 0., 0.]
normalize_quaternion([.5,.5,.5,.5]) -> [0.5, 0.5, 0.5, 0.5]   (unchanged)
```

This is already a known defect — `test_wart_zero_quaternion_becomes_the_identity_matrix_3952` carries a `pytest.skip` describing exactly it ("at float16 the default eps=1e-12 inside normalize_quaternion underflows to 0 … same underflow class as kornia#3966"). Nobody has fixed it.

### Two things to sort out if you bring it back

It is not a free win, and I'd rather tell you now than have you hit it in CI:

1. **It trips a #3952 pin.** Applied alone onto current `main`:

   ```
   FAILED test_wart_normalize_quaternion_without_eps_divides_by_zero_3952[cpu-float32]
   FAILED test_wart_normalize_quaternion_without_eps_divides_by_zero_3952[cpu-float64]
   ```

   That pin exists to catch this exact mechanism change. `max(eps, finfo.tiny)` silently raises an **explicitly passed `eps=0`** to `tiny`, so a caller who deliberately asked for no guard quietly gets one. Worth deciding whether to keep the override unconditional or apply it only to the default — and either way the pin needs updating in the same PR, with a note saying which.

2. **It does not fix #3952's actual contract.** `test_convention_normalize_quaternion_is_unit_below_eps_3952` still xfails afterwards — sub-eps quaternions still come back non-unit. So this is a fix for the float16 underflow corner of #3952, not for #3952 itself, and the PR description should say so.

So: a focused PR against `normalize_quaternion` alone, framed as the float16 underflow half of #3952, with those two pins handled. That's a clean, self-contained contribution and it's yours if you want it — no rush, and no hard feelings about this one.


---

ducha-aiki:
@TejaashriUK — following up on the close, since I said the `normalize_quaternion` hunk of this PR was the part that wasn't superseded.

I've now filed it as its own issue: #4021. Your read was right — `normalize_quaternion(torch.zeros(4, dtype=torch.float16))` returns all-NaN on `main` today, while every other dtype returns zeros, because the default `eps = 1e-12` underflows to exactly `0.0` at `float16` and the `max(||q||, eps)` clamp becomes a no-op. Same underflow class as #3966, different function.

The issue writes up the two catches I mentioned in the close, so whoever picks it up doesn't rediscover them:

- `max(eps, finfo(dtype).tiny)` silently overrides an explicit `eps=0.0`, which is currently a supported "no guard" argument and is pinned by `test_wart_normalize_quaternion_without_eps_divides_by_zero_3952`. A fix has to floor only the default, or change that pin's contract on purpose.
- It does **not** fix #3952's contract — that's a separate defect about the clamp producing a non-unit result when it *does* apply.

The invitation stands if you'd like to take #4021: your original hunk plus a `float16` cell and the `eps=0.0` handling is the whole PR. If you'd rather not, no problem at all — it's now tracked either way.

PR #3970, opened 2026-08-12
fix(geometry): avoid fp16 NaNs in quaternion_exp_to_log

### LINKED ISSUE #3966: quaternion_exp_to_log returns NaN in float16 for any zero-vector-part quaternion (identity included): the default eps=1e-8 underflows to 0

BruceWae:

**Fixes/Relates to:** Fixes #3966

## What changed

This fixes `quaternion_exp_to_log` returning `NaN` for float16 quaternions whose vector part is zero, including both `(1, 0, 0, 0)` and `(-1, 0, 0, 0)`.

The default `eps=1e-8` underflows to `0.0` in float16, so the norm clamp can become a no-op and the log map ends up dividing `0 / 0`. The fix keeps the user-provided epsilon but raises the effective clamp floor to at least `torch.finfo(dtype).tiny` for floating-point tensors.

## Why

For zero-vector-part quaternions, the log map should return the origin instead of NaNs. This matches the existing behavior for float32/float64 and keeps the default float16 path safe.

Fixes #3966.

## Validation

I verified the regression before the fix:

```text
tests/geometry/test_conversions.py::TestQuaternionExpToLog::test_unit_quaternion_fp16_default_eps[cpu] FAILED
actual = tensor([nan, nan, nan], dtype=torch.float16)
expected = tensor([0., 0., 0.], dtype=torch.float16)
```

After the fix:

```text
D:\Dev\Miniconda3\python.exe -m pytest tests\geometry\test_conversions.py::TestQuaternionExpToLog -q
11 passed in 3.67s
```

```text
D:\Dev\Miniconda3\python.exe -m pytest tests\geometry\test_conversions.py -q
230 passed, 5 xfailed in 4.21s
```

```text
D:\Dev\Miniconda3\python.exe -m ruff check kornia\geometry\conversions.py tests\geometry\test_conversions.py
All checks passed!
```

Local note: these checks were run on Windows with Python 3.14.6 and torch 2.13.0+cpu.

- [ ] 🟢 **Human-written**

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -858,6 +858,8 @@ def quaternion_exp_to_log(quaternion: torch.Tensor, eps: float = 1.0e-8) -> torc
     quaternion_vector = quaternion[..., 1:4]
 
     # compute quaternion norm
+    if quaternion.is_floating_point():
+        eps = max(eps, torch.finfo(quaternion.dtype).tiny)
     norm_q: torch.Tensor = torch.norm(quaternion_vector, p=2, dim=-1, keepdim=True).clamp(min=eps)
 
     # apply log map
--- a/tests/geometry/test_conversions.py
+++ b/tests/geometry/test_conversions.py
@@ -422,6 +422,13 @@ class TestQuaternionExpToLog(BaseTester):
         quaternion_log = kornia.geometry.conversions.quaternion_exp_to_log(quaternion_exp, eps=eps)
         self.assert_close(quaternion_log, expected, atol=atol, rtol=rtol)
 
+    @pytest.mark.parametrize("scalar", (1.0, -1.0))
+    def test_zero_vector_part_fp16_default_eps(self, scalar, device):
+        quaternion_exp = torch.tensor((scalar, 0.0, 0.0, 0.0), device=device, dtype=torch.float16)
+        expected = torch.tensor((0.0, 0.0, 0.0), device=device, dtype=torch.float16)
+        quaternion_log = kornia.geometry.conversions.quaternion_exp_to_log(quaternion_exp)
+        self.assert_close(quaternion_log, expected)
+
     def test_pi_quaternion_x(self, device, dtype, atol, rtol):
         eps = torch.finfo(dtype).eps
         quaternion_exp = torch.tensor((0.0, 1.0, 0.0, 0.0), device=device, dtype=dtype)


ducha-aiki:
Heads-up before you invest more here: **four open PRs currently overlap on #3966 and #3975** — #3969, #3970 (this one), #3990, #4000. The issues were filed in a burst and several people picked them up independently, so this is a coordination problem, not anyone's mistake.

Per the green lane in `CONTRIBUTING.md` this resolves as **first *good* PR wins** — first to meet the bar, not first to open. A maintainer arbitrates.

This PR is the earliest of the four that ships a test, which counts for a lot. Its scope is the narrowest: `quaternion_exp_to_log` only, so it addresses #3966 but not #3975. That is a legitimate choice — a small correct fix beats a large speculative one — but note that #4000 covers both in one change, so the maintainer may prefer the wider scope. If you want to close that gap, extending to `quaternion_log_to_exp` is the move.

### Pins you must update

|---|---|---|

---

ducha-aiki:
Closed in favor of #3990

PR #3971, opened 2026-08-12
fix(geometry): pad mixed-size laplacian pyramids

### LINKED ISSUE #3927: build_pyramid/build_laplacian_pyramid: tautological max_level check; laplacian padding predicate skips mixed sizes and raises on odd/pow2 inputs
Found during the #3926 convention audit (all execution-verified, torch 2.9.1):

1. **Tautological bounds check**: `KORNIA_CHECK(isinstance(max_level, int) or max_level < 0, ...)` can never reject an int — its own message says "must be a positive integer", yet `max_level=0` and `-1` silently return a single-element list (same as `max_level=1`).
2. **Laplacian padding predicate is likely or-vs-and inverted**: `require_padding = not (is_powerof_two(w) or is_powerof_two(h))` pads only when *neither* dim is a power of two, so 6×6 pads to 8×8 but 6×8/8×6 don't pad at all.
3. **Consequence**: mixed odd/power-of-two inputs raise — `build_laplacian_pyramid(torch.rand(1,1,5,8), 2)` → `RuntimeError: size (5) must match (4)` in the pyrup subtraction.

The fix for (1) is two tokens (`or` → `and`… on the check), but since #3926's audit shows the ≤0 behavior is reachable today, tightening it is technically breaking — needs a maintainer call on validation vs deprecation. (2)+(3) look like the padding was meant to fire when *either* dim is non-pow2.

Documented as `.. warning::` blocks in #3926 pending this fix. Split out of the closed umbrella #3923 per review feedback.

—
*Issue authored by Claude (Claude Code) on behalf of @ducha-aiki.*

BruceWae:

**Fixes/Relates to:** Fixes #3927

## What does this PR do?

This fixes the Laplacian pyramid padding predicate for mixed power-of-two input sizes, such as `5x8` and `8x5`.

On `main`, `build_laplacian_pyramid(torch.rand(1, 1, 5, 8), 2)` raises during the residual subtraction because no padding is applied when only one dimension is already a power of two. The fix pads mixed-size inputs when `max_level > 1`, while preserving the existing `max_level=1` behavior that does not need adjacent-level subtraction.

I also updated the docstring to remove the now-fixed mixed-size warning, while leaving the separate `max_level <= 0` validation note unchanged.

## Test log

```
$ pytest tests\geometry\transform\test_pyramid.py::TestBuildLaplacianPyramid::test_mixed_power_of_two_size_padding -q
2 passed in 3.67s

$ pytest tests\geometry\transform\test_pyramid.py -q
61 passed in 3.75s

$ ruff check kornia\geometry\transform\pyramid.py tests\geometry\transform\test_pyramid.py
All checks passed!
```

- [ ] 🟢 **Human-written**

### DIFF
--- a/kornia/geometry/transform/pyramid.py
+++ b/kornia/geometry/transform/pyramid.py
@@ -591,23 +591,17 @@ def build_laplacian_pyramid(
           level — level 0 is **not** the unchanged original image whenever
           ``max_level > 1``
         - the input is reflect-padded up to the next power of two (per
-          dimension) only when **neither** the height **nor** the width is
-          already a power of two; when it is applied, every returned level
+          dimension) when padding is required to keep adjacent Gaussian levels
+          aligned for subtraction; when it is applied, every returned level
           — including level 0 — is shaped from the padded size
         - border_type: ``'reflect'`` by default
         - align_corners: ``False`` by default
 
     .. warning::
-        For a mixed odd/power-of-two input size (one dimension a power of two,
-        the other not), no padding is applied and the level-to-level
-        subtraction can raise ``RuntimeError`` from a shape mismatch when
-        ``max_level > 1`` (e.g. a :math:`(1, 1, 5, 8)` input with
-        ``max_level=2``). This is likely unintended and tracked in
-        `#3927 <https://github.com/kornia/kornia/issues/3927>`_. Separately, the
-        ``max_level`` bounds check does not currently reject non-positive
+        The ``max_level`` bounds check does not currently reject non-positive
         **integer** values: passing an integer ``max_level <= 0`` returns the
-        same single-element list as ``max_level=1`` instead of raising — also
-        tracked in `#3927 <https://github.com/kornia/kornia/issues/3927>`_.
+        same single-element list as ``max_level=1`` instead of raising. Tracked
+        in `#3927 <https://github.com/kornia/kornia/issues/3927>`_.
 
     Args:
         input : the torch.Tensor to be used to construct the pyramid with shape :math:`(B, C, H, W)`.
@@ -632,7 +626,11 @@ def build_laplacian_pyramid(
 
     h = input.size()[2]
     w = input.size()[3]
-    require_padding = not (is_powerof_two(w) or is_powerof_two(h))
+    width_is_powerof_two = is_powerof_two(w)
+    height_is_powerof_two = is_powerof_two(h)
+    require_padding = not (width_is_powerof_two or height_is_powerof_two)
+    if max_level > 1:
+        require_padding = not (width_is_powerof_two and height_is_powerof_two)
 
     if require_padding:
         # in case of arbitrary shape torch.Tensor image need to be padded.
--- a/tests/geometry/transform/test_pyramid.py
+++ b/tests/geometry/transform/test_pyramid.py
@@ -224,6 +224,12 @@ class TestBuildLaplacianPyramid(BaseTester):
         assert len(pyramid) == 1
         assert pyramid[0].shape == (1, 2, 4, 5)
 
+    @pytest.mark.parametrize("height,width", ((5, 8), (8, 5)))
+    def test_mixed_power_of_two_size_padding(self, height, width, device, dtype):
+        sample = torch.rand(1, 2, height, width, device=device, dtype=dtype)
+        pyramid = kornia.geometry.transform.build_laplacian_pyramid(sample, max_level=2)
+        assert [img.shape for img in pyramid] == [torch.Size((1, 2, 8, 8)), torch.Size((1, 2, 4, 4))]
+
     @pytest.mark.parametrize("batch_size", (1, 2, 3))
     @pytest.mark.parametrize("channels", (1, 3))
     @pytest.mark.parametrize("max_level", (2, 3, 4))


ducha-aiki:
**CHANGES REQUESTED — the crash you found is real and the fix direction is right, but the predicate is over-broad: it silently changes the output shape of 16 input configurations that work correctly today. Narrow it (concrete predicate below, verified) and this lands as an ordinary bugfix.**

Measured on head `60a9f3db34ba4f3bfe78f11e5506b86045451598` vs merge-base `8454a81a`, in two clean worktrees, torch 2.9.1 / Python 3.11 / CPU. Every number below is pasted command output, not inference.

---

## What is already right

- **The crash is real and you found it.** `build_laplacian_pyramid(torch.rand(1,1,5,8), 2)` raises on the merge-base. Confirmed.
- **Your regression test genuinely fails on the merge-base.** I copied `tests/geometry/transform/test_pyramid.py` from your head onto `base-3971` and ran it there:

```
FAILED tests/geometry/transform/test_pyramid.py::TestBuildLaplacianPyramid::test_mixed_power_of_two_size_padding[cpu-float32-5-8] - RuntimeError: The size of tensor a (5) must match the size of tensor b (4) ...
FAILED ...[cpu-float32-8-5] / [cpu-float64-5-8] / [cpu-float64-8-5]
====================== 4 failed, 113 deselected in 0.88s =======================
```

- **The test is shaped correctly per TESTING.md**: method on `TestBuildLaplacianPyramid(BaseTester)`, injected `device`/`dtype` fixtures, no hardcoded device.
- **The rest of the suite is green on your head**: `tests/geometry/transform/test_pyramid.py -q --device=cpu --dtype=float32,float64` → `117 passed in 1.70s`. Doctests: `pytest --doctest-modules kornia/geometry/transform/pyramid.py` → `5 passed`. `ruff check` → clean.
- **The changed path is otherwise healthy.** On `5x8, max_level=2`: `torch.autograd.gradcheck` → `True`; grad is finite with shape `(1,1,5,8)`; dtype preserved for `float32/float64/float16/bfloat16`; `torch.compile(..., fullgraph=True)` compiles and runs; `dynamic=True` also works.

That is a well-formed green-lane bugfix. The problem is the blast radius of the predicate, not the goal.

---

## Blocker 1 — the new predicate silently reshapes 16 input sizes that work correctly today

`kornia/geometry/transform/pyramid.py:631-633`

```python
require_padding = not (width_is_powerof_two or height_is_powerof_two)
if max_level > 1:
    require_padding = not (width_is_powerof_two and height_is_powerof_two)
```

I swept `H, W ∈ {4,5,6,7,8,9,12,16} × max_level ∈ {1,2,3,4}` (256 cells) on both worktrees, recording all level shapes or the exception:

```
=== SUMMARY over H,W in {4,5,6,7,8,9,12,16} x max_level in {1,2,3,4} (256 cells) ===
crashes fixed (shape-mismatch -> works):        26
previously-WORKING inputs whose output shape changed: 16
regressions (worked -> crash):                   0
remaining subtract-MISMATCH crashes under the PR: 0
```

The 26 fixes are real. But so are the 16 shape changes:

```
SHAPE-CHANGED CELLS: base -> pr
  max_level=2  4x6  : 4x6|2x3       -> 4x8|2x4
  max_level=2  4x12 : 4x12|2x6      -> 4x16|2x8
  max_level=2  6x4  : 6x4|3x2       -> 8x4|4x2
  max_level=2  6x8  : 6x8|3x4       -> 8x8|4x4
  max_level=2  6x16 : 6x16|3x8      -> 8x16|4x8
  max_level=2  8x6  : 8x6|4x3       -> 8x8|4x4
  max_level=2  8x12 : 8x12|4x6      -> 8x16|4x8
  max_level=2 12x4  : 12x4|6x2      -> 16x4|8x2
  max_level=2 12x8  : 12x8|6x4      -> 16x8|8x4
  max_level=2 12x16 : 12x16|6x8     -> 16x16|8x8
  max_level=2 16x6  : 16x6|8x3      -> 16x8|8x4
  max_level=2 16x12 : 16x12|8x6     -> 16x16|8x8
  max_level=3  8x12 : 8x12|4x6|2x3  -> 8x16|4x8|2x4
  max_level=3 12x8  : 12x8|6x4|3x2  -> 16x8|8x4|4x2
  max_level=3 12x16 : 12x16|6x8|3x4 -> 16x16|8x8|4x4
  max_level=3 16x12 : 16x12|8x6|4x3 -> 16x16|8x8|4x4
```

Note what is **not** in the crash list: `8x6` and `6x8`. Issue #3927 hypothesis (2) says "6×8/8×6 don't pad at all" as if that were the defect. It isn't. Those sizes work today, and they work *correctly* — take `8x6, max_level=2, float64`:

```
base-3971:  8x6 ml=2 level shapes: [(1, 1, 8, 6), (1, 1, 4, 3)]
            reconstruction err vs original 8x6 region: 0.0
pr-3971:    8x6 ml=2 level shapes: [(1, 1, 8, 8), (1, 1, 4, 4)]
            reconstruction err vs original 8x6 region: 0.0
base level0 (1, 1, 8, 6)  pr level0 (1, 1, 8, 8)
max |base - pr[:, :, :8, :6]| = 0.0031681954493072984
```

The merge-base already reconstructs `8x6` exactly (error `0.0`). The PR pads it to `8x8`, changes the returned shape, and perturbs the overlapping level-0 values by up to `3.2e-3` from reflect-padding contamination. For that family of inputs the change is pure loss: no crash fixed, no accuracy gained, a changed output contract.

Under kornia's contribution policy that mixes two classes in one PR: the crash fix is an ordinary bugfix that can land now, but a silent output-shape change for currently-working inputs is a semantic/default change that would otherwise have to wait for the coordinated repair window. Narrowing the predicate removes the second class entirely, so the whole thing can land now.

### The real invariant

`pyrdown` produces `floor(s/2)` (`pyramid.py:453`), `pyrup` produces `2*s` (`pyramid.py:497`). Level `i` is subtracted from `pyrup(level i+1)`, so the round trip is exact **iff every side that gets subtracted is even**:

```
s_0 = S,  s_{i+1} = floor(s_i / 2),  need s_i even for i = 0 .. max_level-2
  <=>  S % 2**(max_level - 1) == 0
```

Powers of two are a sufficient-but-far-too-strong proxy for that. `12 % 2 == 0`, so `12x8` at `max_level=2` never needed padding; `8 % 4 == 0` and `12 % 4 == 0`, so `8x12` at `max_level=3` never needed it either.

Minimal-blast-radius version — keeps the legacy rule untouched so nothing that works today moves, and adds the misalignment condition:

```python
h = input.size()[2]
w = input.size()[3]
# legacy rule, unchanged
require_padding = not (is_powerof_two(w) or is_powerof_two(h))
if max_level > 1:
    # levels 0..max_level-2 are subtracted from pyrup(next level); pyrdown gives
    # floor(s/2) and pyrup gives 2*s, so the round trip is exact only if the side
    # is even at every one of those levels, i.e. s % 2**(max_level-1) == 0.
    stride = 1 << (max_level - 1)
    require_padding = require_padding or (h % stride != 0) or (w % stride != 0)
```

I implemented exactly that against the merge-base and re-ran the identical sweep:

```
PROPOSED PREDICATE (legacy-or-misaligned, pow2 pad target):
  previously-WORKING inputs whose shape changes : 0 []
  regressions (worked -> crash)                 : 0 []
  subtract-MISMATCH crashes fixed               : 26
  subtract-MISMATCH crashes remaining           : 0 []
  cells where proposal differs from PR          : 16   (exactly the 16 listed above)
```

Same 26 crashes fixed, zero behavior change for anything that already worked. **Your existing test still passes unchanged under this predicate** (`5x8` → `h % 2 == 1` → pad → `8x8|4x4`), so the diff is one condition, not a rewrite.

(If you want to go further and shrink the padding *target* from "next power of two" to "next multiple of `2**(max_level-1)`" — that is a strictly better fit for the invariant, but it *would* move currently-working inputs, so please leave it out of this PR and raise it separately.)

## Blocker 2 — `Fixes #3927` will auto-close an issue that is only half fixed

Issue #3927 documents two defects: (1) the tautological `max_level` bounds check, and (2)/(3) the padding predicate. This PR fixes (2)/(3) only, and the docstring correctly keeps the (1) warning — which still cites #3927. Per AGENTS.md, before merging a change that closes one of those tracked issues:

```
$ grep -rnE "#3927|issues/3927" kornia/ tests/
kornia/geometry/transform/pyramid.py:528:        `#3927 <https://github.com/kornia/kornia/issues/3927>`_.
kornia/geometry/transform/pyramid.py:604:        in `#3927 <https://github.com/kornia/kornia/issues/3927>`_.
```

Two surviving hits in `kornia/`, zero in `tests/` (no pinned `xfail` to re-check, so nothing to drop there). Surviving hits in `kornia/` mean the close is premature. I confirmed the retained claim is still **TRUE** by executing it on your head:

```
max_level=0:  len=1 shapes=[(1, 2, 8, 8)]
max_level=-1: len=1 shapes=[(1, 2, 8, 8)]
max_level=1:  len=1 shapes=[(1, 2, 8, 8)]
```

Please change the PR body to **`Relates to #3927`**. Fixing the bounds check is a separate call (tightening it is technically breaking) and should not ride along here.

## Blocker 3 — the rewritten Convention sentence is falsified by the code the PR leaves in place

`kornia/geometry/transform/pyramid.py:593-596` (Convention block) now reads:

> the input is reflect-padded up to the next power of two (per dimension) **when padding is required to keep adjacent Gaussian levels aligned for subtraction**

At `max_level=1` there are no adjacent levels and no subtraction at all, yet the legacy branch still pads:

```
max_level=1  5x7: base= 8x8   pr= 8x8
max_level=1  6x6: base= 8x8   pr= 8x8
```

So the sentence is untrue as written for `max_level=1`, and (per Blocker 1) also untrue for `max_level>1`, where padding fires on inputs that are already aligned. Either describe the rule mechanically ("padded to the next power of two per dimension unless both sides are already powers of two; at `max_level=1` padding is applied only when neither side is") or, better, adopt the narrowed predicate and then the alignment wording becomes accurate for `max_level>1` — but it still needs a clause for the `max_level=1` legacy behavior.

---

## Nits (non-blocking)

1. **Test only covers the fixed direction, not the preserved one.** Add a shape-*preservation* case so the next person cannot re-broaden the predicate without a red test — e.g. `(8, 6)` and `(12, 8)` at `max_level=2` asserting `[(1,2,8,6),(1,2,4,3)]` / `[(1,2,12,8),(1,2,6,4)]`. Also worth one `max_level=3` case (e.g. `8x12` → unpadded, `8x6` → padded), since the misalignment condition is level-dependent and `max_level=2` alone cannot distinguish "even" from "power of two".
2. `tests/geometry/transform/test_pyramid.py:236-238` (`test_num_levels`) uses `16, 32` — both powers of two — which is why the padded path was never exercised by the existing suite. Worth a comment or a non-pow2 parametrization.
3. Naming: `width_is_powerof_two` / `height_is_powerof_two` are fine, but with the narrowed predicate they collapse back into one expression; keep whichever reads better after the change.
4. Out of scope, just noting it so it is on the record: 70 cells in my sweep still raise `RuntimeError: Padding size should be less than the corresponding input dimension` (82 on the merge-base) — that is the 5×5 Gaussian kernel's reflect pad on a level that has shrunk to 1–2 px, a distinct pre-existing limitation. Not this PR's job; the PR reduces the count rather than increasing it.
5. This PR carries the `stale` label. That is housekeeping and had no bearing on the verdict — this comment resets the clock, so please take the time to do the narrowing properly.

---

Summary of what I ran: sweep of 256 `(H, W, max_level)` combinations on both worktrees; your test file transplanted onto `base-3971` (4 failures, fail-on-base proven); full `test_pyramid.py` on head (117 passed); module doctests (5 passed); `ruff check` (clean); `gradcheck`, four dtypes, and `torch.compile` (`fullgraph=True` and `dynamic=True`) on the changed path; a from-scratch implementation of the proposed predicate re-swept over the same 256 cells. Not verified: CUDA, MPS, ONNX export.

Thanks for digging into this one — the diagnosis is correct and the remaining work is a one-condition narrowing plus the docstring and the `Fixes:` line.

ducha-aiki (changes requested):
See Opus review

victorwon2001:
**kornia/geometry/transform/pyramid.py:629** — The user-visible padding behavior change has no `changelog.d/3971.(added|fixed|breaking).md` fragment, and the completed changelog check is failing.

The PR currently has a failing required check and therefore does not satisfy the repository's stated requirement to fix failing checks before consideration.

The changed predicate alters when padding is applied. Check 105153028033 reports that no matching changelog fragment was changed and exits with failure; the contribution guide states that failing checks must be fixed.

`file:kornia/geometry/transform/pyramid.py:diff`

> +    width_is_powerof_two = is_powerof_two(w)
> +    height_is_powerof_two = is_powerof_two(h)
> +    require_padding = not (width_is_powerof_two or height_is_powerof_two)
> +    if max_level > 1:
> +        require_padding = not (width_is_powerof_two and height_is_powerof_two)

`comment:5455974932`

> I swept `H, W ∈ {4,5,6,7,8,9,12,16} × max_level ∈ {1,2,3,4}` (256 cells) on both worktrees, recording all level shapes or the exception:
> 
> ```
> === SUMMARY over H,W in {4,5,6,7,8,9,12,16} x max_level in {1,2,3,4} (256 cells) ===
> crashes fixed (shape-mismatch -> works):        26
> previously-WORKING inputs whose output shape changed: 16
> regressions (worked -> crash):                   0
> remaining subtract-MISMATCH crashes under the PR: 0
> ```
> 
> The 26 fixes are real. But so are the 16 shape changes:

`file:tests/geometry/transform/test_pyramid.py:diff`

> +    @pytest.mark.parametrize("height,width", ((5, 8), (8, 5)))
> +    def test_mixed_power_of_two_size_padding(self, height, width, device, dtype):
> +        sample = torch.rand(1, 2, height, width, device=device, dtype=dtype)
> +        pyramid = kornia.geometry.transform.build_laplacian_pyramid(sample, max_level=2)
> +        assert [img.shape for img in pyramid] == [torch.Size((1, 2, 8, 8)), torch.Size((1, 2, 4, 4))]

`subject`

> **Fixes/Relates to:** Fixes #3927

`github_actions_job:105153028033:annotation:message`

> no changed changelog.d/3971.(added|fixed|breaking).md fragment found; add one for a user-visible change or apply the no-changelog label for an internal or release PR

Narrow the padding condition to preserve previously working aligned sizes while padding only legacy-required or level-misaligned inputs; add shape-preservation coverage, make the Convention text match the actual rule, change the PR relation to #3927 so it does not imply the remaining bounds defect is fixed, and satisfy the changelog check before re-running the relevant test and CI coverage.

SHA: `60a9f3db34ba4f3bfe78f11e5506b86045451598`

PR #3973, opened 2026-08-14
Honor the declared vertex count in binary PLY files

wiliyam:

**Fixes/Relates to:** N/A

## What does this PR do?

Stops `load_pointcloud_ply_binary` after the number of vertices declared by the PLY header. Previously, any trailing elements whose payload happened to be a multiple of 24 bytes were decoded as extra XYZ points.

The regression test appends one 24-byte trailing element to a one-vertex file. It fails on `main` with a `[2, 3]` result and passes with this change as `[1, 3]`.

## Test log

```text
$ uv run --with pytest pytest -q tests/geometry/test_pointcloud_io.py
collected 10 items
tests/geometry/test_pointcloud_io.py ..........                          [100%]
10 passed in 0.05s

$ uvx ruff check kornia/geometry/pointcloud.py tests/geometry/test_pointcloud_io.py
All checks passed!

$ uvx ruff format --check kornia/geometry/pointcloud.py tests/geometry/test_pointcloud_io.py
2 files already formatted

$ uvx ty check kornia/geometry/pointcloud.py
All checks passed!
```

- [ ] 🟢 **Human-written**

### DIFF
--- a/kornia/geometry/pointcloud.py
+++ b/kornia/geometry/pointcloud.py
@@ -167,7 +167,7 @@ def load_pointcloud_ply_binary(filename: str, header_size: int = 8) -> torch.Ten
 
     Args:
         filename: the path to the pointcloud.
-        header_size: the number of header lines to skip.
+        header_size: retained for backwards compatibility. The header is read through ``end_header``.
 
     Return:
         tensor containing the loaded points with shape :math:`(*, 3)` where
@@ -182,10 +182,22 @@ def load_pointcloud_ply_binary(filename: str, header_size: int = 8) -> torch.Ten
 
     vertex_count = None
     with open(filename, "rb") as f:
-        for _ in range(header_size):
-            header_line = f.readline().decode("ascii", errors="ignore").split()
+        for _ in range(10000):
+            raw_header_line = f.readline()
+            if not raw_header_line:
+                raise ValueError(f"PLY header in {filename!r} does not contain end_header.")
+            header_line = raw_header_line.decode("ascii", errors="replace").split()
             if len(header_line) == 3 and header_line[:2] == ["element", "vertex"]:
-                vertex_count = int(header_line[2])
+                try:
+                    vertex_count = int(header_line[2])
+                except ValueError as exc:
+                    raise ValueError(f"Invalid PLY vertex count {header_line[2]!r} in {filename!r}.") from exc
+                if vertex_count < 0:
+                    raise ValueError(f"PLY vertex count must be non-negative in {filename!r}. Got {vertex_count}.")
+            if header_line == ["end_header"]:
+                break
+        else:
+            raise ValueError(f"PLY header in {filename!r} exceeds 10000 lines.")
         raw_data = f.read()
 
     if vertex_count is not None:
@@ -193,6 +205,9 @@ def load_pointcloud_ply_binary(filename: str, header_size: int = 8) -> torch.Ten
         if len(raw_data) < vertex_data_size:
             raise ValueError(f"Expected {vertex_data_size} bytes for {vertex_count} points, got {len(raw_data)} bytes.")
         raw_data = raw_data[:vertex_data_size]
+
+    if not raw_data:
+        return torch.empty((0, 3), dtype=torch.float32)
 
     # One point equals 24 bytes (3 * 8 bytes for double float)
     if len(raw_data) % 24 != 0:
--- a/kornia/geometry/pointcloud.py
+++ b/kornia/geometry/pointcloud.py
@@ -180,12 +180,19 @@ def load_pointcloud_ply_binary(filename: str, header_size: int = 8) -> torch.Ten
     if not (isinstance(header_size, int) and header_size > 0):
         raise TypeError(f"Input header_size must be a positive integer. Got {header_size}.")
 
-    # Read all file bytes
+    vertex_count = None
     with open(filename, "rb") as f:
-        # Skip header lines
         for _ in range(header_size):
-            f.readline()
+            header_line = f.readline().decode("ascii", errors="ignore").split()
+            if len(header_line) == 3 and header_line[:2] == ["element", "vertex"]:
+                vertex_count = int(header_line[2])
         raw_data = f.read()
+
+    if vertex_count is not None:
+        vertex_data_size = vertex_count * 24
+        if len(raw_data) < vertex_data_size:
+            raise ValueError(f"Expected {vertex_data_size} bytes for {vertex_count} points, got {len(raw_data)} bytes.")
+        raw_data = raw_data[:vertex_data_size]
 
     # One point equals 24 bytes (3 * 8 bytes for double float)
     if len(raw_data) % 24 != 0:
--- a/kornia/geometry/pointcloud.py
+++ b/kornia/geometry/pointcloud.py
@@ -181,12 +181,16 @@ def load_pointcloud_ply_binary(filename: str, header_size: int = 8) -> torch.Ten
         raise TypeError(f"Input header_size must be a positive integer. Got {header_size}.")
 
     vertex_count = None
+    vertex_properties: list[tuple[str, str]] = []
+    in_vertex_element = False
     with open(filename, "rb") as f:
         for _ in range(10000):
             raw_header_line = f.readline()
             if not raw_header_line:
                 raise ValueError(f"PLY header in {filename!r} does not contain end_header.")
             header_line = raw_header_line.decode("ascii", errors="replace").split()
+            if header_line[:1] == ["element"]:
+                in_vertex_element = header_line[:2] == ["element", "vertex"]
             if len(header_line) == 3 and header_line[:2] == ["element", "vertex"]:
                 try:
                     vertex_count = int(header_line[2])
@@ -194,6 +198,8 @@ def load_pointcloud_ply_binary(filename: str, header_size: int = 8) -> torch.Ten
                     raise ValueError(f"Invalid PLY vertex count {header_line[2]!r} in {filename!r}.") from exc
                 if vertex_count < 0:
                     raise ValueError(f"PLY vertex count must be non-negative in {filename!r}. Got {vertex_count}.")
+            elif in_vertex_element and len(header_line) == 3 and header_line[0] == "property":
+                vertex_properties.append((header_line[1], header_line[2]))
             if header_line == ["end_header"]:
                 break
         else:
@@ -201,6 +207,12 @@ def load_pointcloud_ply_binary(filename: str, header_size: int = 8) -> torch.Ten
         raw_data = f.read()
 
     if vertex_count is not None:
+        expected_properties = [("double", "x"), ("double", "y"), ("double", "z")]
+        if vertex_properties != expected_properties:
+            raise ValueError(
+                f"Only vertices with exactly three double properties (x, y, z) are supported in {filename!r}. "
+                f"Got {vertex_properties}."
+            )
         vertex_data_size = vertex_count * 24
         if len(raw_data) < vertex_data_size:
             raise ValueError(f"Expected {vertex_data_size} bytes for {vertex_count} points, got {len(raw_data)} bytes.")
--- a/tests/geometry/test_pointcloud_io.py
+++ b/tests/geometry/test_pointcloud_io.py
@@ -130,3 +130,18 @@ class TestSaveLoadPointCloud(BaseTester):
 
         if os.path.exists(filename):
             os.remove(filename)
+
+    def test_load_pointcloud_binary_stops_at_declared_vertex_count(self, tmp_path):
+        xyz_save = torch.tensor([[1.0, 2.0, 3.0]])
+        filename = tmp_path / "pointcloud_binary_with_trailing_data.ply"
+        kornia.geometry.save_pointcloud_ply_binary(str(filename), xyz_save)
+
+        # A PLY file may contain elements after its vertices (for example, faces).
+        # Use 24 trailing bytes to expose loaders that infer the vertex count from
+        # the entire remaining payload instead of the header's element declaration.
+        with filename.open("ab") as file:
+            file.write(bytes(24))
+
+        xyz_load = kornia.geometry.load_pointcloud_ply_binary(str(filename))
+
+        self.assert_close(xyz_load, xyz_save)
--- a/tests/geometry/test_pointcloud_io.py
+++ b/tests/geometry/test_pointcloud_io.py
@@ -145,3 +145,37 @@ class TestSaveLoadPointCloud(BaseTester):
         xyz_load = kornia.geometry.load_pointcloud_ply_binary(str(filename))
 
         self.assert_close(xyz_load, xyz_save)
+
+    def test_load_pointcloud_binary_finds_end_header(self, tmp_path):
+        xyz_save = torch.tensor([[1.0, 2.0, 3.0]])
+        filename = tmp_path / "pointcloud_binary_long_header.ply"
+        kornia.geometry.save_pointcloud_ply_binary(str(filename), xyz_save)
+
+        header, payload = filename.read_bytes().split(b"end_header\n", maxsplit=1)
+        filename.write_bytes(header + b"comment extra header line\nend_header\n" + payload)
+
+        xyz_load = kornia.geometry.load_pointcloud_ply_binary(str(filename))
+
+        self.assert_close(xyz_load, xyz_save)
+
+    @pytest.mark.parametrize("vertex_count", ["-1", "invalid"])
+    def test_load_pointcloud_binary_rejects_invalid_vertex_count(self, tmp_path, vertex_count):
+        filename = tmp_path / "pointcloud_binary_invalid_count.ply"
+        filename.write_bytes(
+            b"ply\n"
+            b"format binary_little_endian 1.0\n"
+            + f"element vertex {vertex_count}\n".encode()
+            + b"property double x\nproperty double y\nproperty double z\nend_header\n"
+        )
+
+        with pytest.raises(ValueError, match="PLY vertex count"):
+            kornia.geometry.load_pointcloud_ply_binary(str(filename))
+
+    def test_load_pointcloud_binary_empty(self, tmp_path):
+        xyz_save = torch.empty((0, 3))
+        filename = tmp_path / "pointcloud_binary_empty.ply"
+        kornia.geometry.save_pointcloud_ply_binary(str(filename), xyz_save)
+
+        xyz_load = kornia.geometry.load_pointcloud_ply_binary(str(filename))
+
+        self.assert_close(xyz_load, xyz_save)
--- a/tests/geometry/test_pointcloud_io.py
+++ b/tests/geometry/test_pointcloud_io.py
@@ -179,3 +179,19 @@ class TestSaveLoadPointCloud(BaseTester):
         xyz_load = kornia.geometry.load_pointcloud_ply_binary(str(filename))
 
         self.assert_close(xyz_load, xyz_save)
+
+    def test_load_pointcloud_binary_rejects_extra_vertex_properties(self, tmp_path):
+        filename = tmp_path / "pointcloud_binary_with_normals.ply"
+        header = (
+            b"ply\n"
+            b"format binary_little_endian 1.0\n"
+            b"element vertex 2\n"
+            b"property double x\nproperty double y\nproperty double z\n"
+            b"property double nx\nproperty double ny\nproperty double nz\n"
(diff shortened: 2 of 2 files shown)


ducha-aiki:
**Changes requested — the fix is correct for kornia-written files, but on a standard third-party PLY it converts a loud `ValueError` into a silently-returned tensor of `inf` garbage.**

Measured on head `27904548` vs merge-base `8454a81a` (not `main`), torch 2.9.1, Python 3.11.14, CPU. Every block below is real pasted output from files built on disk, run against both worktrees with the identical script.

## What is already right

- The bug is real and the diagnosis is correct. Trailing elements were being decoded as extra vertices, and the header already carries the answer.
- The regression test genuinely fails on the merge-base, so the green-lane bar is met:
  ```
  base 8454a81a: FAILED tests/geometry/test_pointcloud_io.py::TestSaveLoadPointCloud::test_load_pointcloud_binary_stops_at_declared_vertex_count
  E  AssertionError: The values for attribute 'shape' do not match: torch.Size([2, 3]) != torch.Size([1, 3]).
  head 27904548: 10 passed in 0.47s
  ```
- Round-trips of kornia-written files are byte-identical to the merge-base. I hashed the output tensors for n = 1, 5, 80, 1000; all four SHAs match on both revisions. No regression on the happy path.
- The header parse itself is robust on the inputs I threw at it: CRLF endings, runs of extra whitespace, and an `element face` block are all handled correctly (`str.split()` absorbs `\r`, and the `len == 3` guard keeps `comment element vertex 999` from matching).
- **Wrong-property-type files get strictly better.** A file declaring `property float x/y/z` used to decode as silent nonsense; now it raises an actionable error:
  ```
  base: OK shape=(2, 3) data=[2.0000004, 512.00012, 8192.002, 131072.03, 524288.12, 2097152.5]
  head: ValueError: Expected 96 bytes for 4 points, got 48 bytes.
  ```
- **No unbounded allocation.** An attacker-controlled `N` only ever participates in an integer compare before the length check; the slice is never reached when `N` is large. `element vertex 1000000000000000000` raises cleanly rather than allocating. (`f.read()` was already unbounded before this PR; unchanged.)

## Blockers

### 1. `kornia/geometry/pointcloud.py:191-195` — the truncation destroys the misalignment check, turning a raise into silent garbage

The `len(raw_data) % 24 != 0` guard on line 198 was the only thing catching a `header_size` that does not match the file. Truncating to exactly `vertex_count * 24` makes that modulo **always zero**, so the guard is now dead whenever a count was parsed.

This is not theoretical. Here is a stock MeshLab/VCGLIB binary PLY — a 10-line header, 3 points — swept over `header_size`:

```
file: 269 bytes; header lines: 10

=========== base 8454a81a ===========
  header_size= 4 -> RAISED ValueError: Expected 24 bytes per point, got a total of 191 values.
  header_size= 5 -> RAISED ValueError: Expected 24 bytes per point, got a total of 173 values.
  header_size= 6 -> RAISED ValueError: Expected 24 bytes per point, got a total of 155 values.
  header_size= 7 -> RAISED ValueError: Expected 24 bytes per point, got a total of 137 values.
  header_size= 8 -> RAISED ValueError: Expected 24 bytes per point, got a total of 122 values.
  header_size= 9 -> RAISED ValueError: Expected 24 bytes per point, got a total of 83 values.
  header_size=10 -> RETURNED shape=(3, 3) correct=True first=[1.0, 2.0, 3.0]

=========== head 27904548 ===========
  header_size= 4 -> RETURNED shape=(3, 3) correct=False first=[inf, 0.0, inf]
  header_size= 5 -> RETURNED shape=(3, 3) correct=False first=[inf, 0.0, inf]
  header_size= 6 -> RETURNED shape=(3, 3) correct=False first=[inf, 0.0, inf]
  header_size= 7 -> RETURNED shape=(3, 3) correct=False first=[0.0, inf, 0.0]
  header_size= 8 -> RETURNED shape=(3, 3) correct=False first=[inf, inf, 0.0]
  header_size= 9 -> RETURNED shape=(3, 3) correct=False first=[inf, 0.0, 0.0]
  header_size=10 -> RETURNED shape=(3, 3) correct=True first=[1.0, 2.0, 3.0]
```

Six new silent-wrong-output windows, and `header_size=8` — the **default**, the value a user gets by calling `load_pointcloud_ply_binary(f)` — is one of them. Today that call raises and the user goes and finds the right header length. After this PR it returns a plausible-looking `(3, 3)` tensor full of header bytes reinterpreted as doubles.

The trigger condition is precise: `element vertex` sits within `header_size`, but `header_size` is *below* the true header length. That is exactly the "user under-specified `header_size` for a third-party file" case — and it is the same file class the PR's own test comment names as the motivation ("A PLY file may contain elements after its vertices (for example, faces)"). The smallest legal vertex+face binary PLY needs **9** header lines, so kornia's default of 8 is always wrong for it, and the default path lands squarely in the new silent window.

Trading "loud error on a misconfigured read" for "silent garbage" is a net negative even though the targeted bug is real. The count check needs to be additive to the alignment check, not a replacement for it.

The clean fix is to stop trusting `header_size` for locating the payload: scan lines until one equals `end_header` (bounded, e.g. a few thousand lines, to stay safe on non-PLY input), take the data offset from there, and only then apply the vertex count. That makes both the old bug and this new one impossible, and `header_size` degrades to a compatibility no-op. If you would rather keep the diff minimal, at least refuse to truncate unless the payload is consistent — e.g. require `len(raw_data) >= vertex_data_size` **and** that the parsed header actually terminated within `header_size` lines.

### 2. `kornia/geometry/pointcloud.py:192` — a negative `vertex_count` silently drops points off the end

`vertex_count * 24` is negative, `len(raw_data) < -24` is false, and `raw_data[:-24]` then trims from the tail:

```
3-point file, header says 'element vertex -1'
  base: OK shape=(3, 3) data=[1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0]
  head: OK shape=(2, 3) data=[1.0, 2.0, 3.0, 4.0, 5.0, 6.0]

3-point file, header says 'element vertex -2'
  base: OK shape=(3, 3) data=[1.0, 2.0, 3.0, ...]
  head: OK shape=(1, 3) data=[1.0, 2.0, 3.0]
```

Needs an explicit `vertex_count < 0` rejection.

### 3. `kornia/geometry/pointcloud.py:188` — `int()` is unguarded, and the resulting error names neither the file nor PLY

```
header 'element vertex abc'  -> ValueError: invalid literal for int() with base 10: 'abc'
header 'element vertex 0x03' -> ValueError: invalid literal for int() with base 10: '0x03'
```
(base: both loaded fine, ignoring the bad count.)

Raising here is defensible, but the message is not actionable for someone holding a PLY file. Wrap it and say which file and which header line failed. Minor related oddity: Python's `int()` accepts underscore separators, so `element vertex 1_0` is parsed as 10 — harmless, but it shows the parse is looser than PLY's grammar.

### 4. `kornia/geometry/pointcloud.py:195` — `element vertex 0` now reaches `torch.frombuffer` with an empty buffer

```
3-point file, header says 'element vertex 0'
  base: OK shape=(3, 3)
  head: ValueError: both buffer length (0) and count (-1) must not be 0
```

Neither is right, but the head error is a raw torch internal message. Note this is the **same failure the empty round-trip already hits on both revisions** — `save_pointcloud_ply_binary` writes `element vertex 0` when every point is non-finite, and loading it back fails identically on base and head:

```
  empty file bytes: 143
  base: ValueError: both buffer length (0) and count (-1) must not be 0
  head: ValueError: both buffer length (0) and count (-1) must not be 0
```

That empty round-trip break is pre-existing and **not** this PR's fault, so I am not asking you to own it. But since you are already touching this exact branch, `vertex_count == 0` -> `return torch.empty((0, 3))` fixes both for one line, and it turns a genuine round-trip bug into a passing test. Worth doing here or in a follow-up issue.

## Repair-window classification

**Ordinary bugfix, lands now — no repair window needed.** Both directions of the intended change are clearly-broken-output repairs, not semantic changes:

- Trailing elements decoded as extra vertices was *silently wrong output*. Fixing it is exempt.
- The new raise on a `float32`-property file replaces *silently wrong output* with an error. Replacing garbage with a clear error is a repair, not a behavior change to batch. The old "extra points" return was never a contract anyone could depend on — it produced `inf`/denormal noise, as the base output above shows.

So the gating question is not the window; it is blocker 1, which introduces a **new** clearly-broken-output defect in the opposite direction. That is why this is changes-requested rather than a hold.

## Non-blocking nits

- `decode("ascii", errors="ignore")` (line 186) over-matches by deleting bytes rather than failing. A vertex line containing non-ASCII bytes is silently repaired into a match: `b"element\xff vertex\xfe 2\n"` decodes to `element vertex 2` and truncates a 3-point file to 2 points (head `(2, 3)` vs base `(3, 3)`). I could not construct a realistic *binary-payload* false positive without also mis-setting `header_size`, so this is hardening, not a live exploit. `errors="replace"` would keep the line from matching, which is the safer failure.
- The hardcoded `24` is now load-bearing in two places while the header right above it declares the real property types. Reading `property double x/y/z` and validating them would make blocker 1 and the float32 case fall out for free. Out of scope for a green-lane fix, but worth an issue.
- **API implication:** after this change `header_size` is half-obsolete — it still selects the data offset (and can now silently select the wrong one), while the count it used to imply comes from the header. If you take the `end_header` approach it becomes fully vestigial, which is a better place to be; either way the docstring should say what it does now. Note `load_pointcloud_ply` (line 127) has the identical parameter and the identical assumption, and is untouched — fine to leave, but the two should not drift.
- Test coverage: the new test asserts only the fixed case. None of the four blockers above has a test. Per `TESTING.md` please add exception tests for the negative count, the non-numeric count, and the header-longer-than-`header_size` case. The test does not use the injected `device`/`dtype` fixtures — consistent with the rest of the file, so I am not asking you to change it, but it does mean the new path is only exercised at CPU/float32.
- `ruff check`, `ruff format --check`, and the full `tests/geometry/test_pointcloud_io.py` (10 passed) are clean on head; I reproduced all three.

Happy to re-review quickly once blocker 1 is addressed — the rest are small. Thanks for the clear reproduction in the PR description; the fail-on-base evidence made this much faster to verify.

wiliyam:
Addressed the requested changes in `0fe63971`:

- scan through the actual `end_header` marker instead of using `header_size` as the payload offset, preventing header bytes from being decoded as points
- reject negative and malformed vertex counts with PLY/file-specific errors
- handle zero-vertex files as an empty `(0, 3)` tensor
- use replacement decoding so non-ASCII bytes cannot be deleted into a matching header declaration
- add regressions for a header longer than the default `header_size`, negative/non-numeric counts, and the existing empty round trip

Focused verification (Python 3.13.15, torch 2.9.1 CPU):

```text
pytest tests/geometry/test_pointcloud_io.py -q
14 passed in 0.05s

ruff check kornia/geometry/pointcloud.py tests/geometry/test_pointcloud_io.py
All checks passed!

ruff format --check kornia/geometry/pointcloud.py tests/geometry/test_pointcloud_io.py
2 files already formatted
```

This follow-up was AI-generated and manually reviewed; the commands and results above were executed locally.

ducha-aiki:
**Delta re-review of `0fe63971` — verdict: CHANGES REQUESTED, one blocker.** The Group-B finding from my 2026-08-28 review is substantially fixed and the PR is now a real improvement over `main`. But one silent-garbage window survives, and for that input the new code is **worse than the merge-base**, so I would like it closed before this lands.

Reviewed at head `0fe63971`, merge-base `8454a81a`, CPU, torch 2.9.1 (the repo `.venv`), in clean worktrees of both.

## What the delta fixed — all verified

Scanning to the real `end_header` was the right call, and the `len(raw_data) < vertex_data_size` guard is exactly the missing piece: truncation can now only ever *discard trailing* bytes, so it can no longer manufacture a passing `% 24` check out of short data. That was the core of my objection and it is gone.

I ran the same eight-case matrix on both sides:

| case | merge-base `8454a81a` | head `0fe63971` |
| --- | --- | --- |
| standard **7-line** header, 3 pts | `ValueError: both buffer length (0) and count (-1) must not be 0` | `(3, 3)` correct |
| 8-line header (comment), 3 pts | `(3, 3)` correct | `(3, 3)` correct |
| declares 3, only 2 present | `ValueError: both buffer length (0)...` | `ValueError: Expected 72 bytes for 3 points, got 48 bytes.` |
| declares 2, 3 present (trailing element) | `ValueError: both buffer length (0)...` | `(2, 3)` correct |
| zero vertices | `ValueError: both buffer length (0)...` | `(0, 3)` float32 |
| negative count | `ValueError: both buffer length (0)...` | `ValueError: PLY vertex count must be non-negative` |
| non-numeric count | `ValueError: both buffer length (0)...` | `ValueError: Invalid PLY vertex count 'abc'` |
| no `end_header` | `ValueError: both buffer length (0)...` | `ValueError: PLY header ... does not contain end_header.` |

Worth calling out that row 1 is a **bigger win than the PR advertises**: the canonical minimal binary PLY header (`ply` / `format` / `element vertex N` / three `property` lines / `end_header`) is **seven** lines, so the old `header_size=8` default skipped one line too many and failed on the most standard file there is. Row 2 shows the 8-line case is behaviour-preserved.

Empty-result consistency is right too — `torch.empty((0, 3), dtype=torch.float32)` matches the non-empty path's `.to(torch.float32)`. That is the defect class we just fixed elsewhere in #4080, and you avoided it.

**The tests are genuine pins.** Run against the merge-base source, 5 of the 5 new tests fail; at head all 14 pass:

```console
$ # merge-base worktree, PR's test file copied in
FAILED ...test_load_pointcloud_binary_stops_at_declared_vertex_count
FAILED ...test_load_pointcloud_binary_finds_end_header
FAILED ...test_load_pointcloud_binary_rejects_invalid_vertex_count[-1]
FAILED ...test_load_pointcloud_binary_rejects_invalid_vertex_count[invalid]
FAILED ...test_load_pointcloud_binary_empty
5 failed, 9 passed

$ # head
14 passed
```

## Blocker — a vertex element with more than three properties now returns plausible garbage

The function assumes each vertex is exactly three `double`s (24 bytes), but nothing checks it. A PLY carrying normals or colours — by far the most common real layout after bare xyz — has a 48- or 27-byte stride, and honouring the declared count now guarantees `len(raw_data) % 24 == 0`, which is precisely the guard that used to catch this.

Two vertices, six double properties (`x y z nx ny nz`), normals all `9.0`:

```console
$ # merge-base 8454a81a
  6-property vertex -> ValueError: Expected 24 bytes per point, got a total of 126 values.

$ # head 0fe63971
  6-property vertex -> shape=(2, 3)  values=[1.0, 2.0, 3.0, 9.0, 9.0, 9.0]
```

The merge-base told the caller something was wrong. The head returns a tensor with the **right vertex count** whose second "point" is actually the first vertex's normal. There is nothing in that output to suggest it is wrong — right shape, plausible values — and it is the shape check the caller is most likely to trust.

This is the same defect class as my original finding, on the one axis the `<` guard does not cover. It is the last window, and it is the one where the PR regresses against `main`.

## Suggested fix — measured, and it does not disturb anything else

Collect the vertex element's `property` lines while scanning the header and require exactly three doubles:

```python
    vertex_count = None
    vertex_properties: list[str] = []
    in_vertex_element = False
    with open(filename, "rb") as f:
        for _ in range(10000):
            ...
            if header_line[:1] == ["element"]:
                in_vertex_element = header_line[:2] == ["element", "vertex"]
            if len(header_line) == 3 and header_line[:2] == ["element", "vertex"]:
                ...  # unchanged
            elif in_vertex_element and header_line[:1] == ["property"]:
                vertex_properties.append(header_line[1])
            if header_line == ["end_header"]:
                break
        raw_data = f.read()

    if vertex_count is not None:
        if vertex_properties and vertex_properties != ["double", "double", "double"]:
            raise ValueError(
                f"Only vertices of exactly three double properties (x, y, z) are supported in {filename!r}. "
                f"Got {len(vertex_properties)} properties of types {vertex_properties}."
            )
        vertex_data_size = vertex_count * 24
```

The `in_vertex_element` flag matters — without it, `property` lines belonging to a later `element face` would be miscounted.

Applied to your branch, measured:

```console
6-property vertex -> ValueError: Only vertices of exactly three double properties (x, y, z) are supported...
all 8 cases in the table above -> unchanged
tests/geometry/test_pointcloud_io.py -> 14 passed
```

Use it as-is or your own variant; the requirement is only that a non-xyz-double vertex layout raises instead of returning a plausible tensor. Please add a test for it — a 6-property file is four lines to construct, and it is the case most likely to regress again later.

## Two smaller notes, non-blocking

1. **`header_size` is now validated but ignored.** `load_pointcloud_ply_binary(f, header_size=0)` still raises `TypeError` for a value the function no longer reads. Preserving the raise is defensible, but the docstring should say the argument is deprecated rather than merely "retained for backwards compatibility" — and a `DeprecationWarning` when a caller passes a non-default value would be the honest signal. Not required for this PR.
2. **`for _ in range(10000): ... else: raise`** reads oddly at first but is correct and bounded. No change needed; noting it so a later reader does not "simplify" it into an unbounded loop.

Everything else in the delta I checked is right, and the header-scanning rewrite is a clear improvement. One guard away.

wiliyam:
Addressed the remaining blocker in `2846faad`:

- track properties only while parsing the vertex element
- require the supported binary layout to be exactly `double x`, `double y`, `double z`
- reject vertex records with normals, colours, or other layouts instead of decoding them as plausible XYZ points
- add a regression using two six-double `(x, y, z, nx, ny, nz)` records

Focused verification (Python 3.11.14, torch 2.9.1 CPU):

```text
pytest tests/geometry/test_pointcloud_io.py -q
15 passed in 0.06s

ruff check kornia/geometry/pointcloud.py tests/geometry/test_pointcloud_io.py
All checks passed!

ruff format --check kornia/geometry/pointcloud.py tests/geometry/test_pointcloud_io.py
2 files already formatted
```

This follow-up was AI-generated and manually reviewed; the commands and results above were executed locally.

ducha-aiki (dismissed):
**Delta re-review of `2846faad` — APPROVE.**

The last blocker is closed the way I asked: properties are tracked only while inside the `element vertex` block, and a vertex layout other than three `double`s is rejected instead of being read as `count` plausible-looking rows. The new test with `nx/ny/nz` covers exactly the case I flagged.

Measured at this head vs the merge-base (`.venv`, torch 2.9.1):

```
head        tests/geometry/test_pointcloud_io.py   15 passed
merge-base  same test file                          6 failed, 9 passed
```

The six fail-on-base are the genuine pins (standard 7-line header, declared-count truncation, >3-property rejection). Clean merge against `main`. Thanks for sticking through three rounds — each one made the reader strictly safer.

ducha-aiki:
**One more thing before merge — I should have caught it in the approval above, sorry for the extra round.**

```python
for _ in range(10000):
    raw_header_line = f.readline()
    if not raw_header_line:
        raise ValueError(...)
    ...
else:
    raise ValueError(f"PLY header in {filename!r} exceeds 10000 lines.")
```

The `10000` is a magic number that does no work. `f.readline()` returns `b""` at EOF, and the `if not raw_header_line` check already raises there, so the loop terminates on every input without a bound — the `else:` branch is reachable *only* through the constant. What the constant does add is a failure mode: a legal PLY header with more than 10 000 `comment` / `obj_info` lines now raises. Please make it

```python
while True:
    raw_header_line = f.readline()
    if not raw_header_line:
        raise ValueError(f"PLY header in {filename!r} does not contain end_header.")
    ...
    if header_line == ["end_header"]:
        break
```

and drop the "exceeds 10000 lines" error. To be clear for anyone reading later: this bound was on **header lines**, not on the number of points — the vertex payload is sized from `element vertex N` and is not capped. Everything else in my approval stands; this is the only change.

ducha-aiki:
Closing this one, and withdrawing my approval above — a maintainer decision, not a reaction to the last comment.

The problem you found is real and I want to keep credit for it where it belongs: `load_pointcloud_ply_binary` trusts a `header_size` line count instead of reading to `end_header`, so it cannot even open a standard seven-line PLY header with its own default, and it infers the point count from the payload length instead of `element vertex N`. That will be fixed.

But after three rounds the loader that resulted is not one we are comfortable maintaining: the `range(10000)` bound with an unreachable `else`, a `header_size` argument kept alive with no effect, a property check that hard-codes one layout instead of parsing the element, and a review history where each round patched the previous round's hole rather than reworking the parse. That pattern usually means the code is being written to satisfy the reviewer rather than from an understanding of the format, and it is safer for the maintainers to write this reader ourselves, with the PLY spec open, than to keep iterating here. The tracking issue will reference this PR for the diagnosis.

Thank you for the persistence and for responding quickly each round — that part was genuinely appreciated.

Issue #3959, opened 2026-08-10
normal_transform_pixel(dtype=torch.int64) silently returns a rank-deficient matrix, and the homography/extrinsics family has five different integer-input behaviours


ducha-aiki:
The homography-normalization and camera-extrinsics family responds to an integer input in **five mutually inconsistent ways**, and one of them returns silently wrong numbers.

The one that produces wrong output: `normal_transform_pixel(4, 5, dtype=torch.int64)` truncates the scale factors `0.5` and `0.667` to `0` and returns a rank-deficient matrix that maps **every** pixel to the constant `(−1, −1)`.

```python
>>> normal_transform_pixel(4, 5, dtype=torch.int64)
tensor([[[ 0,  0, -1],
         [ 0,  0, -1],
         [ 0,  0,  1]]])
```

The scales are computed as Python floats and then handed to `torch.tensor(..., dtype=dtype)`, so an integer `dtype` truncates them:

```python
# kornia/geometry/conversions.py:1321-1332
sx: float = 2.0 / width_denom
sy: float = 2.0 / height_denom
tr_mat = torch.tensor(
    [[sx, 0.0, -1.0], [0.0, sy, -1.0], [0.0, 0.0, 1.0]],
    device=device,
    dtype=dtype,
)
```

## Repro

```python
import torch
import kornia.geometry.conversions as C

print("# normal_transform_pixel: silent truncation to a rank-deficient matrix")
print(C.normal_transform_pixel(4, 5, dtype=torch.int64))
print("normal_transform_pixel3d(2,4,5,int64) diag =",
      [C.normal_transform_pixel3d(2, 4, 5, dtype=torch.int64)[0, i, i].item() for i in range(3)])

print("\n# five different outcomes for the same user mistake:")
Hi = torch.eye(3, dtype=torch.long)[None]
R = torch.eye(3, dtype=torch.long)[None]
t = torch.ones(1, 3, 1, dtype=torch.long)
M4 = torch.eye(4, dtype=torch.long)[None]
cases = [
    ("normal_transform_pixel(4,5,dtype=int64)", lambda: C.normal_transform_pixel(4, 5, dtype=torch.int64)),
    ("normalize_homography(int64)", lambda: C.normalize_homography(Hi, (4, 5), (4, 5))),
    ("denormalize_homography(int64)", lambda: C.denormalize_homography(Hi, (4, 5), (4, 5))),
[...]

Issue #3948, opened 2026-08-10
axis_angle_to_quaternion(torch.tensor([1, 0, 0])) silently returns tensor([0, 0, 0, 0]): the output buffer is allocated with the integer input dtype


ducha-aiki:
`axis_angle_to_quaternion` allocates its output buffer with the **input** dtype, so an integer axis-angle vector silently returns an all-zero integer tensor instead of a quaternion — no error, no warning, and the result is not even a unit quaternion.

```python
>>> axis_angle_to_quaternion(torch.tensor([1, 0, 0]))
tensor([0, 0, 0, 0])
```

The cause (`kornia/geometry/conversions.py:909-915`):

```python
quaternion: torch.Tensor = torch.zeros(
    size=(*axis_angle.shape[:-1], 4), dtype=axis_angle.dtype, device=axis_angle.device
)
quaternion[..., 1:2] = a0 * k
quaternion[..., 2:3] = a1 * k
quaternion[..., 3:4] = a2 * k
quaternion[..., 0:1] = w
```

`a0 * k` and `w` are float (the trig ops promote), but they are assigned into an `int64` buffer, so every value truncates toward zero. `cos(0.5) = 0.8776` becomes `0`, `sin(0.5)/1 = 0.4794` becomes `0`.

## Repro

```python
import torch
from kornia.geometry.conversions import axis_angle_to_quaternion as aa2q

v_int = torch.tensor([1, 0, 0])            # dtype=torch.int64
v_flt = torch.tensor([1.0, 0.0, 0.0])

print("int64  input:", v_int, "->", aa2q(v_int))
print("float32 input:", v_flt, "->", aa2q(v_flt))
print("norm of the int64 result:", aa2q(v_int).float().norm().item())

for dt in (torch.int32, torch.int64, torch.uint8, torch.int16):
    print(f"  {str(dt):<14} -> {aa2q(torch.tensor([1, 0, 0], dtype=dt))}")
```

## Observed

```
int64  input: tensor([1, 0, 0]) -> tensor([0, 0, 0, 0])
float32 input: tensor([1., 0., 0.]) -> tensor([0.8776, 0.4794, 0.0000, 0.0000])
norm of the int64 result: 0.0
  torch.int32    -> tensor([0, 0, 0, 0], dtype=torch.int32)
  torch.int64    -> tensor([0, 0, 0, 0])
  torch.uint8    -> tensor([0, 0, 0, 0], dtype=torch.uint8)
[...]

PR #3974, opened 2026-08-14
fix(conversions): reject integer dtype in normal_transform_pixel

rawsun007:
## Which lane?

**Relates to:** #3959 (the `normal_transform_pixel` / `normal_transform_pixel3d` clause only; the issue stays open for the other functions it covers)

## What does this PR do?

`normal_transform_pixel` scales by `2 / (size - 1)`, computed as Python floats and then passed to `torch.tensor(..., dtype=dtype)`. An integer dtype truncates those scales to 0, so it returned a rank-deficient matrix that maps every pixel to `(-1, -1)`, with no error:

```
>>> normal_transform_pixel(4, 5, dtype=torch.int64)      # main
tensor([[[ 0,  0, -1],
         [ 0,  0, -1],
         [ 0,  0,  1]]])      # det == 0
```

`normal_transform_pixel3d` has the same problem (diagonal `[0, 0, 2]`). Both functions now raise `ValueError("dtype must be a floating point or complex type. Got <dtype>.")` for any non-floating, non-complex `dtype` (integers, unsigned integers, `bool`, quantized).

**Uniform by decision.** A dimension of 3 pixels or fewer makes the scale integral, so an integer matrix would happen to be correct there. The rejection does not depend on `height`/`width` anyway: the accepted dtypes are a property of the function, not of the image size.

**Unconditional.** The check is a plain `if`/`raise`, not a `KORNIA_CHECK`, so `disable_checks()`, `python -O` and `KORNIA_CHECKS=0` do not turn it off — the alternative is silently wrong output. This matches the positive-size guard just above it.

**How the dtype is classified.** TorchScript cannot read `dtype.is_floating_point`, and these functions run inside the scripted `warp_affine` path. A closed allowlist would reject valid floating formats (the `float8` dtypes, and whatever torch adds next), so the guard asks a zero-element probe instead: `torch.empty(0, dtype=dtype)` and `is_floating_point() or is_complex()`. A membership test over the four common float dtypes and the three complex dtypes runs first as a fast path; it keeps the complex dtypes so a `complex32` call does not trigger torch's experimental-ComplexHalf warning from the probe allocation. `dtype=None` skips the check.

Accepted, unchanged from `main`: every floating dtype including `float8_e4m3fn`/`float8_e5m2`, and every complex dtype. Both functions still `torch.jit.script`, and `torch.compile(fullgraph=True)` has no graph break.

**Docs.** The two `.. warning::` blocks that documented the truncation as a #3959 defect now describe the rejection, and name it as one clause of #3959. The wart pin `test_wart_integer_dtype_truncates_the_scale_to_zero_3959` is retired in favour of convention tests.

## Tests

In `tests/geometry/test_conversions.py`, class `TestNormalTransformPixel`:

- `test_convention_integer_dtype_is_rejected_3959` — `int64`, `int32`, `uint8`, `bool`; both functions raise, including a 3×3 image where an integer scale would have been exact (the uniform-by-decision cell). On `main`, `uint8` already raised, but with torch's overflow `RuntimeError`, not a kornia `ValueError`.
- `test_convention_non_default_float_and_complex_dtypes_are_accepted_3959` — `float16`, `bfloat16`, `complex64`, `float8_e4m3fn` (when available): `normal_transform_pixel` returns that dtype, on the requested device, with the expected values (not just shape, since an all-zero matrix is the failure being fixed).
- `test_convention_integer_dtype_rejection_is_unconditional_3959` — still raises after `disable_checks()`.

With `main`'s `kornia/geometry/conversions.py` swapped in, the 5 rejection cases fail and the 4 acceptance cases pass.

## Test log (head `78d8e3ce`, torch 2.14.0, CPU)

```
$ pytest tests/geometry/test_conversions.py -k "NormalTransformPixel and 3959" --device=cpu --dtype=float32
9 passed

$ pytest tests/geometry/test_conversions.py tests/geometry/test_grid.py \
    tests/geometry/transform/test_homography_warper.py tests/geometry/transform/test_imgwarp.py \
    --device=cpu --dtype=float32,float64
1208 passed, 2 skipped, 11 xfailed

# same 9 tests with main's conversions.py
5 failed, 4 passed

$ KORNIA_CHECKS=0 python -O -c "... normal_transform_pixel(4, 5, dtype=torch.int64)"
ValueError: dtype must be a floating point or complex type. Got torch.int64.
```

Hosted CI on this head covers the torch 2.5.1 legs, half precision, dynamo and MPS.

## AI Usage Disclosure
- [ ] 🟢 **Human-written**

---
_Description refreshed by a maintainer to match head `78d8e3ce` (the earlier text described a superseded allowlist guard and a test that no longer exists). The author's disclosure above is unchanged._


rawsun007:
Heads-up on an interaction with #3984, which landed in this file after I opened this.

That PR adds a `.. warning::` to `normal_transform_pixel` documenting the current behaviour:

> An integer `dtype` truncates the scale instead of raising: `normal_transform_pixel(4, 5, dtype=torch.int64)` returns `[[0, 0, -1], [0, 0, -1], [0, 0, 1]]`, which maps every pixel to the constant `(-1, -1)`. Tracked in #3959.

This PR makes that call raise, so whichever of the two merges second leaves the other stale: the warning would describe behaviour that no longer exists, and the linked issue would be closed.

Happy to go whichever way suits the batch:

1. If #3984 merges first, I rebase and drop that warning block in this PR.
2. If this merges first, the warning in #3984 can be dropped before it lands, or I can send a follow-up.
3. If you would rather keep the documented-behaviour route for now and not change the behaviour at all, say so and I will close this. The `.. warning::` plus the tracking issue is a defensible answer on its own.

No action needed from me until you pick. Flagging it only because the two are easy to merge independently and then notice later.

ducha-aiki (changes requested):
Right diagnosis, and your TorchScript reasoning checks out — I verified it rather than taking it on trust. Three things to resolve before this can merge, one of which is a real hole in the guard.

Everything below is your branch rebased onto current `main` (`35555b55`), where `#4006` has rewritten both functions; see the conflict note at the end.

## Verified

- The defect is real: `normal_transform_pixel(4, 5, dtype=torch.int64)` returns `[[0, 0, -1], [0, 0, -1], [0, 0, 1]]` — rank-deficient, every pixel mapped to one point, exactly as `#3959` describes.
- Your membership test **does** script. `torch.jit.script(normal_transform_pixel)` compiles, returns `[0.5, 0.0, -1.0]` for the default call, and raises inside the interpreter for `dtype=torch.int64`. `torch.jit.script(warp_affine)` still runs end to end. So the choice of a tuple test over `dtype.is_floating_point` is justified and now demonstrated, not just asserted.
- Complex works as your tuple implies: `complex64` and `complex32` both build `[(0.5+0j), 0j, (-1+0j)]` (the latter with torch's own ComplexHalf experimental warning).
- After rebasing, `tests/geometry` gives exactly **one** failure — the `#3959` wart pin, which is the machinery working.

## Blocking: `KORNIA_CHECK` can be switched off, so the fix is conditional

```text
>>> normal_transform_pixel(4, 5, dtype=torch.int64)
BaseError: dtype must be a floating point or complex type

>>> from kornia.core.check import disable_checks; disable_checks()
>>> normal_transform_pixel(4, 5, dtype=torch.int64)
tensor([[[0, 0, -1], [0, 0, -1], [0, 0, 1]]])          dtype=torch.int64
```

`KORNIA_CHECK` is gated on `_KORNIA_CHECKS_ENABLED`, which `disable_checks()`, `python -O` and `KORNIA_CHECKS=0` all clear — the module documents that as the production path. So under `-O` this returns the rank-deficient matrix again and `#3959` is unfixed.

That is the wrong tier for this check. `KORNIA_CHECK` is for assertions a caller can afford to skip once they trust their inputs; this is a domain restriction where the alternative is silently wrong output. Compare the guard **three lines above yours**, added by `#4006`:

```python
if not torch.jit.is_tracing() and (height <= 0 or width <= 0):
    raise ValueError(f"Input image size must be positive. Got height={height}, width={width}.")
```

Unconditional, and for a strictly less dangerous input. Please raise directly the same way. Note the exception *type* is not the issue — `tests/geometry/test_conversions.py:376` defines `_KORNIA_GUARD_EXCEPTIONS = (BaseError, ValueError)` precisely because kornia uses both, so `BaseError` classifies correctly through the existing `#3959` pin machinery either way. It is the disableability that matters. (Once you switch, tighten your own test from `pytest.raises(Exception, match=...)` to the concrete type.)

## Required: the docstrings, and this is a narrowing rather than a deletion

`grep -rnE "#3959|issues/3959" kornia/` returns **six** hits across six functions:

| line | function | fixed by this PR? |
|---|---|---|
| 2072 | `normal_transform_pixel` | yes |
| 2203 | `normal_transform_pixel3d` | yes |
| 1932 | `normalize_homography` | no |
| 2326 | `denormalize_homography` | no |
| 2618 | `Rt_to_matrix4x4` | no |
| 2741 | `camtoworld_graphics_to_vision_4x4` | no |

`#3959` is the "five different integer-input behaviours" issue, and this PR changes two of them. So the two clauses you fix must be **rewritten to say the call now raises**, not deleted, and the other four have to be re-executed to confirm they are still accurate — they describe a family, and the family just became less uniform. `#3959` itself does not close on this PR.

The wart pin `TestNormalTransformPixel::test_wart_integer_dtype_truncates_the_scale_to_zero_3959` fails with its own message and must be retired in this PR:

```
kornia#3959: an integer dtype no longer truncates the 2-D normalization scales to 0
```

There is no strict-`xfail` convention pin for this one, so a positive test replacing it is on you — your `TestNormalTransformPixelDtype` class covers it, and `int64/int32/uint8/bool` is a good spread.

## Merge-order interaction with `#3979`

Worth knowing before either lands. On `main` today `normalize_homography` calls `normal_transform_pixel(h, w)` without a `dtype`, so your guard never sees the caller's dtype and an int64 homography keeps its current behavior. `#3979` changes that call to forward `dst_pix_trans_src_pix.dtype`. Measured with both applied:

```text
normalize_homography(torch.eye(3, dtype=torch.int64)[None], (4, 4), (6, 6))

main                     RuntimeError: expected scalar type Long but found Float
+ #3974 alone            RuntimeError: expected scalar type Long but found Float   (unchanged)
+ #3974 and #3979        BaseError: dtype must be a floating point or complex type
```

So `normalize_homography`'s and `denormalize_homography`'s `#3959` clauses — rows 1932 and 2326 above — are falsified by the *combination*, not by either PR alone. Whichever of the two merges second has to update them.

## Conflicts

Both files conflict with `#4006`, and neither is mechanical:

- `kornia/geometry/conversions.py` — `#4006` deleted the `width_denom`/`height_denom` lines your hunk carries as trailing context and added the positive-size guard and the `eps` deprecation warning. Resolution: keep `#4006`'s guards and insert only the `KORNIA_CHECK` block after them, dropping the dead `_denom` lines.
- `tests/geometry/test_grid.py` — your new class collides with `#4006`'s singleton-axis tests. Both are keepers; concatenate.

I resolved it that way and every measurement above comes from that tree; `ruff check` is clean on it.

Requesting changes on the disableable guard and the docstring scope. The direction is right — `maintainer_can_modify` is on if you would rather I push the rebase and you take the rest.

---

rawsun007:
Rebased and all three addressed. Head is e902e463, on a6401efa — main moved past the 35555b55 you measured on.

**The disableable guard.** You're right, and it was the whole fix. `KORNIA_CHECK` on that tier meant `python -O` returned the rank-deficient matrix again, so #3959 was unfixed in the configuration the module documents as production. Now a plain `if`/`raise ValueError`, matching #4006's positive-size guard above it. Measured:

```
disable_checks()   -> ValueError: dtype must be a floating point or complex type
python -O          -> ValueError: dtype must be a floating point or complex type
KORNIA_CHECKS=0    -> ValueError: dtype must be a floating point or complex type
```

`torch.jit.script` still compiles both functions and raises inside the interpreter for `dtype=torch.int64`; `torch.jit.script(warp_affine)` still compiles; `complex64` and `bfloat16` still return. The test is tightened from `pytest.raises(Exception, ...)` to `ValueError`.

**The docstrings.** Both clauses narrowed rather than deleted, and each says the rejection is unconditional. I re-executed the other four:

| function | claim | still holds |
| --- | --- | --- |
| `normalize_homography` | `RuntimeError: expected scalar type Long but found Float` | yes |
| `Rt_to_matrix4x4` | raises `result type Float can't be cast to Long` | yes |
| `camtoworld_to_worldtocam_Rt` | accepts `int64`, returns `int64` | yes |
| `camtoworld_graphics_to_vision_4x4` | accepts `int64`; the `_Rt` sibling raises | yes |

`normalize_homography` is unaffected for the reason you gave — it calls `normal_transform_pixel(src_h, src_w)` with no dtype and casts with `.to(dst_pix_trans_src_pix)`, so the guard never sees the caller's int64. #3959 stays open.

One clause is wrong independently of this PR. `denormalize_homography`'s warning inherits `normalize_homography`'s int64-handling, but it raises `torch.linalg.LinAlgError` on the truncated matrix's zero diagonal rather than a `RuntimeError` from a mixed matmul — which is what its own wart pin already describes. Identical with and without my change, so I've left it for #3959 rather than widening this PR; say the word if you'd rather it were fixed here.

**The wart pin** is retired and replaced with positive cells at the spread you suggested — `int64`/`int32`/`uint8`/`bool`, 2-D and 3-D — plus `float16`/`bfloat16`/`complex64` acceptance and a pin that the rejection survives `disable_checks()`, which is the tier distinction itself. I also updated #3948's promotion pin, which called this defect "still open" and named promotion as the precedent to follow. I deliberately did not follow it: promotion would silently change the dtype a caller explicitly asked for, on a function whose contract is "give me this dtype", and the matrix feeds scripted `warp_affine`. #3959 is now non-uniform on purpose, and the comment says so.

**Conflicts.** `conversions.py` resolved as you described — #4006's guards kept, the guard inserted after them, the dead `_denom` lines gone. For `test_grid.py` I took a different route: the new cells went into `TestNormalTransformPixel` in `test_conversions.py`, where the wart pin they replace lived, so there is no collision with #4006's singleton-axis tests and no duplication of `test_convention_returns_one_unbatched_matrix_in_the_ambient_default_dtype` or the corner-aligned scale pin, which already cover what my old class asserted. `test_grid.py` is untouched now.

**#3979.** Confirmed still true on current main: `normalize_homography` does not forward the dtype, so the combination is what falsifies those two clauses. Whichever of us merges second updates them.

`tests/geometry`: 2406 passed, 1 failed — `TestDecomposeEssentialMatrixNoSVD::test_correct_decompose`, a tolerance miss that reproduces identically on clean main. `tests/geometry/test_conversions.py` alone: 395 passed, 0 failed, against 388 on main (one pin removed, eight cells added). `ruff check` and `ruff format --check` clean.

Thanks for the depth here — the `-O` transcript is what made the tier argument concrete.

ducha-aiki:
**Every claim in your update verifies, and the guard is now the right tier — but the allowlist over-rejects two dtype families that worked correctly before, and the docstring's stated reason for rejecting is false for small images.**

Measured on head `e902e463`, merge base `a6401efa` (a dedicated worktree per side, not `main`), torch 2.9.1, Python 3.11.14, cpu — plus one `--device=mps` run where noted.

## What is already right

Every one of your four claims re-ran clean here. I ran them rather than reading them.

**The tier fix.** Three separate interpreter configurations, same result:

```text
$ python /tmp/v1.py            # optimized: False
2d int64 -> ValueError: dtype must be a floating point or complex type
3d int64 -> ValueError: dtype must be a floating point or complex type
checks enabled: False          # after disable_checks()
after disable_checks -> ValueError: dtype must be a floating point or complex type

$ python -O /tmp/v1.py         # optimized: True
2d int64 -> ValueError: dtype must be a floating point or complex type
after disable_checks -> ValueError: dtype must be a floating point or complex type

$ KORNIA_CHECKS=0 python /tmp/v1.py
2d int64 -> ValueError: dtype must be a floating point or complex type
after disable_checks -> ValueError: dtype must be a floating point or complex type
```

That was the blocking point and it is closed.

**TorchScript.** `torch.jit.script` compiles both functions, the scripted default call returns `[0.5, 0.0, -1.0]`, and the scripted `dtype=torch.int64` call raises with the guard as the innermost frame:

```text
    ):
        raise ValueError("dtype must be a floating point or complex type")
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
builtins.ValueError: dtype must be a floating point or complex type
```

`torch.jit.script(warp_affine)` still compiles and runs end to end (`torch.Size([1, 1, 5, 5])`). `torch.jit.trace` through the function still works, and `torch.compile(..., fullgraph=True, dynamic=True)` still compiles the default call.

**Accepted dtypes.** `float16`, `bfloat16`, `complex32`, `complex64`, `complex128`, `float64` all return, 2-D and 3-D, with the requested dtype preserved.

**The four sibling docstring claims.** Re-executed on both sides; all four are byte-identical on `a6401efa` and `e902e463`, so none of them is falsified by this PR:

```text
                                       head e902e463                  base a6401efa
normalize_homography(int64)            RuntimeError: expected scalar type Long but found Float   (same)
denormalize_homography(int64)          torch._C._LinAlgError: diagonal element 2 is zero         (same)
Rt_to_matrix4x4(int64)                 RuntimeError: result type Float can't be cast ... Long    (same)
camtoworld_to_worldtocam_Rt(int64)     OK dtype=torch.int64                                      (same)
camtoworld_graphics_to_vision_4x4      OK dtype=torch.int64                                      (same)
camtoworld_graphics_to_vision_Rt       RuntimeError: result type Float can't be cast ... Long    (same)
```

One correction on your own report: the `denormalize_homography` clause you flagged as "wrong independently of this PR" is not wrong. `conversions.py:1892` already says a `torch.linalg.LinAlgError` about a zero diagonal, which is exactly what I measured. Nothing to fix there.

**The tests are real pins.** I copied the head's `tests/geometry/test_conversions.py` into the merge-base worktree and ran it against base *code*:

```text
5 failed, 30 passed, 373 deselected
FAILED ... test_convention_integer_dtype_is_rejected_3959[cpu-rejected_dtype0]  - DID NOT RAISE ValueError
FAILED ... test_convention_integer_dtype_is_rejected_3959[cpu-rejected_dtype1]  - DID NOT RAISE ValueError
FAILED ... test_convention_integer_dtype_is_rejected_3959[cpu-rejected_dtype2]  - RuntimeError: value cannot be converted to type uint8 without overflow
FAILED ... test_convention_integer_dtype_is_rejected_3959[cpu-rejected_dtype3]  - DID NOT RAISE ValueError
FAILED ... test_convention_integer_dtype_rejection_is_unconditional_3959[cpu]   - DID NOT RAISE ValueError
```

They use `BaseTester` and the injected `device` fixture, drop `dtype` for a documented reason, and are self-contained. `tests/geometry` at head: **4369 passed, 107 skipped, 18 xfailed, 12 xpassed, 0 failed** (`--dtype=float32,float64`, `-p no:randomly`). `--device=mps -k TestNormalTransformPixel`: 35 passed, so the `complex64` acceptance cell is not mps-hostile. Doctests on the module: 34 passed. Ruff check and format clean.

**Repair-window class.** Confirmed broken-output, so it lands as an ordinary bugfix rather than waiting. On the merge base `normal_transform_pixel(4, 5, dtype=torch.int64)` returns `[[0, 0, -1], [0, 0, -1], [0, 0, 1]]`, `det == 0`, mapping every pixel to a single point, with no error and no warning — and the corresponding float matrix is `[[0.5, 0, -1], [0, 0.6667, -1], [0, 0, 1]]`, so the integer result is not a rounding of the right answer, it is a different linear map. That is the silently-wrong-output class.

## Blocker 1 — `float8` dtypes worked, and now raise with a message that contradicts itself

`kornia/geometry/conversions.py:2097-2106` (and `:2221-2230`) is a closed allowlist of seven dtypes. Every `float8` format falls outside it. On the merge base those calls returned a correct, full-rank matrix:

```text
BASE 2d torch.float8_e4m3fn [[[0.5, 0.0, -1.0], [0.0, 0.6875, -1.0], [0.0, 0.0, 1.0]]]
BASE 2d torch.float8_e5m2   [[[0.5, 0.0, -1.0], [0.0, 0.625,  -1.0], [0.0, 0.0, 1.0]]]
BASE 3d torch.float8_e4m3fn diag [0.5, 0.6875, 2.0, 1.0]
BASE 2d torch.float8_e4m3fnuz -> [0.5, 0.0, -1.0]
BASE 2d torch.float8_e5m2fnuz -> [0.5, 0.0, -1.0]
```

`0.6875` is `2/3` rounded to `float8_e4m3fn` — normal float rounding, the same thing `float16` does. Nothing is truncated to zero; the matrix is invertible. At head:

```text
2d float8_e4m3fn -> ValueError: dtype must be a floating point or complex type
2d float8_e5m2   -> ValueError: dtype must be a floating point or complex type
```

`float8_e4m3fn` **is** a floating point type, so the message states something untrue about its own argument, and `conversions.py:2052` ("Floating and complex dtypes are accepted") is falsified by the same code that the sentence describes. This is exactly the reasoning you used to justify keeping complex: rejecting them "would turn a currently-working call into an error, which is not what this fix is for."

This is not hypothetical for this file. `float8_e4m3fn` is already a live dtype in this family here: `conversions.py:1847` documents its `linalg.cross` gap in `normalize_homography`, and `tests/geometry/test_conversions.py:305-307` probes it through `getattr` precisely because kornia declares `torch>=2.0.0`. The allowlist is also not future-proof — torch 2.9 already ships `float8_e8m0fnu`, which this rejects, and every float format torch adds next will be rejected by default rather than by decision.

A TorchScript-clean fix, verified here. The probe answers the question exactly instead of enumerating:

```python
if dtype is not None and dtype not in (torch.float32, torch.float64, torch.float16, torch.bfloat16):
    probe = torch.empty(0, dtype=dtype)
    if not (probe.is_floating_point() or probe.is_complex()):
        raise ValueError(f"dtype must be a floating point or complex type. Got {dtype}.")
```

Scripted and eager agree on every dtype I tried, including ones the allowlist gets wrong:

```text
torch.float32            eager=True  scripted=True
torch.float64            eager=True  scripted=True
torch.float16            eager=True  scripted=True
torch.bfloat16           eager=True  scripted=True
torch.complex64          eager=True  scripted=True
torch.complex32          eager=True  scripted=True
torch.int64              eager=False scripted=False
torch.uint8              eager=False scripted=False
torch.bool               eager=False scripted=False
torch.uint16             eager=False scripted=False
torch.float8_e4m3fn      eager=True  scripted=True
torch.float8_e5m2        eager=True  scripted=True
torch.float8_e8m0fnu     eager=True  scripted=True
```

The four-dtype fast path keeps the common case free of the allocation: `hybrid(torch.float32)` costs **0.188 us/call**, against **4.34 us** for the whole `normal_transform_pixel(4, 5, dtype=torch.float32)` call, so the guard stays under 5% of the function on the hot path. (The bare probe without the fast path is 0.69 us — 16% — which is why the fast path is worth keeping.) `torch.uint16` shows the other benefit: the allowlist happens to reject it, but only by omission, and it is the same accident that rejects `float8`.

Note also that adding a `Got {dtype}.` suffix would match the positive-size guard three lines above it, which names its values.

## Blocker 2 — "there is no correct integer matrix to return instead" is false for every dimension of 3 pixels or fewer

`kornia/geometry/conversions.py:2051-2052`, repeated at `:2091` in the code comment and in the PR description. Measured on the merge base, integer output for small sizes is **exactly** the float answer, not a truncation of it:

```text
BASE 2d (1, 1) int64=[[1, 0,  0], [0, 1,  0], [0, 0, 1]]  exact_match_float=True
BASE 2d (2, 2) int64=[[2, 0, -1], [0, 2, -1], [0, 0, 1]]  exact_match_float=True
BASE 2d (3, 3) int64=[[1, 0, -1], [0, 1, -1], [0, 0, 1]]  exact_match_float=True
BASE 2d (1, 3) int64=[[1, 0, -1], [0, 1,  0], [0, 0, 1]]  exact_match_float=True
BASE 2d (2, 3) int64=[[1, 0, -1], [0, 2, -1], [0, 0, 1]]  exact_match_float=True
BASE 2d (4, 5) int64=[[0, 0, -1], [0, 0, -1], [0, 0, 1]]  exact_match_float=False
BASE 3d (2, 3, 3) int64 diag [1, 1, 2, 1]   float diag [1.0, 1.0, 2.0, 1.0]
```

(`exact_match_float` is `torch.equal(int_matrix.double(), float64_matrix)`.)

`2/(size-1)` is `1`, `2` or the singleton `1` for sizes 3, 2 and 1, all integral — which your own sentence "fractional for every dimension larger than 3 pixels" concedes two lines earlier, and then the next clause contradicts. So this PR does turn a set of calls that returned the correct answer into errors.

I am **not** asking you to make the guard size-dependent — a dtype domain that depends on the image size would be worse than what you have, and no caller can plausibly be relying on an `int64` normalization matrix for a 3-pixel image. Fix the justification, not the behavior. The honest version is that the accepted dtype set is a property of the function rather than of the arguments, and that it is uniform on purpose: the scale is fractional for any dimension above 3, and an integer dtype silently truncates it there, so integers are rejected for all sizes rather than accepted for the handful where the result happens to be representable. Both `.. warning::` blocks, the two code comments, and the PR body carry the false clause.

`test_convention_integer_dtype_is_rejected_3959` uses `(4, 5)` and `(2, 4, 5)` only, so nothing in the suite pins the small-size decision either way — a cell at `(3, 3)` asserting the rejection would make it a decision rather than an accident.

## Nits, non-blocking

1. `tests/geometry/test_conversions.py:4019-4027` — the acceptance test asserts only `dtype` and `shape`. It would pass against a guard that returned an all-zero matrix, which is the failure mode the whole PR is about. The comment already states the expected complex value (`[(0.5+0j), 0j, (-1+0j)]`); asserting it costs one line. Adding a `getattr(torch, "float8_e4m3fn", None)` cell here, in the style of `:305-307`, is what would have caught Blocker 1.
2. `tests/geometry/test_conversions.py:3995` — the `uint8` cell of the rejection test does not fail on base for the reason the test comment gives. It already raised there, with `RuntimeError: value cannot be converted to type uint8 without overflow` (the `-1.0` offset overflows), not a silent truncation. The cell is still worth keeping — it pins that the error is now kornia's own and uniform — but the comment's "signed/unsigned ... difference would show up here" reads as if all four cells were silently wrong before, and one was not.
3. `test_convention_integer_dtype_rejection_is_unconditional_3959` mutates the process-global check flag. The `try/finally` restores it correctly, so this is a note rather than a request; if you touch that test anyway, a small context manager would make the restoration structural.

Everything else — the tier argument, the retired wart pin, the `#3948` promotion-pin comment recording why the precedent was deliberately not followed, the `normalize_homography` wart comment explaining that the caller does the truncating there — reads well and matches what I measured. `grep -rnE "#3959|issues/3959" kornia/` returns six hits: the two you rewrote (`:2056`, `:2186`) and four that describe functions this PR does not touch (`:1910`, `:2296`, `:2588`, `:2711`), all four re-executed above and all four still accurate. `#3959` correctly stays open.

Requesting changes on the two blockers. Both are contained: one guard expression duplicated in two places, and the prose that explains it.

rawsun007:
Both blockers addressed, head is `9b79b10a`. I re-ran your measurements rather than taking them, on torch 2.13.0 / Python 3.14 / cpu, and your table reproduces exactly - including `float8_e4m3fn`, `float8_e5m2` and `float8_e8m0fnu` as eager=True scripted=True, and `uint16` as the accident you named.

**Blocker 1.** The guard is your probe now, at both sites, with `Got {dtype}.` appended. One change to your snippet: the fast-path tuple keeps the three complex dtypes as well as the four floats. `torch.empty(0, dtype=torch.complex32)` emits `UserWarning: ComplexHalf support is experimental`, so probing a documented-accepted dtype would hand the caller a warning for an allocation they never asked for. The tuple is a fast path rather than the domain either way; anything outside it still reaches the probe.

```
float8_e4m3fn  OK  [0.5, 0, -1]        int64   ValueError: ... Got torch.int64.
float8_e5m2    OK  [0.5, 0, -1]        uint8   ValueError: ... Got torch.uint8.
complex32      OK  [(0.5+0j), 0j, ...] bool    ValueError: ... Got torch.bool.
bfloat16       OK  [0.5, 0, -1]        uint16  ValueError: ... Got torch.uint16.
```

Scripted: `torch.jit.script` compiles both functions, the default call returns `[0.5, 0.0, -1.0]`, `dtype=torch.int64` raises with the guard as the innermost frame, and `dtype=torch.float8_e4m3fn` returns a `float8_e4m3fn` tensor.

**Blocker 2.** The false clause is gone from both `.. warning::` blocks and both code comments. They now say the accepted set is a property of the function rather than of the arguments, that a dimension of 3 pixels or fewer makes the scale integral, and that the rejection is uniform by decision. `test_convention_integer_dtype_is_rejected_3959` gained a `(3, 3)` call, so the small-size case is pinned rather than unexamined. I will fix the PR description last, once you are happy with the prose here.

**Nits.** 1 and 2 are done: the acceptance test now asserts the value (widening the result to `complex128` rather than narrowing the expectation, which discards the imaginary part and warns) and carries a `getattr(torch, "float8_e4m3fn", None)` cell that skips on older torch; the `uint8` comment now says it already raised on base with the overflow error rather than truncating. Nit 3 I left alone.

**One thing your table surfaces that neither of us has ruled on.** `float8_e8m0fnu` is accepted by the probe, and was accepted on base, but it cannot represent the matrix:

```
e8m0fnu : [0.5, 5.88e-39, 1.0,  5.88e-39, 0.5, 1.0,  5.88e-39, 5.88e-39, 1.0]
float64 : [0.5, 0.0,     -1.0,  0.0,      0.667, -1.0, 0.0,     0.0,     1.0]
```

It is exponent-only: no sign bit and no mantissa, so `-1.0` comes back as `1.0` and `0.0` as `5.88e-39`. The translation column is wrong and the scale for the 5-pixel axis is wrong. That is not this PR's doing - base returns the same - and it is a genuinely floating point dtype, so the probe answering yes is the honest answer to the question the probe asks. Rejecting it would mean the domain is "floating point with a sign bit and a mantissa", which is a different rule and a bigger change. I have left it accepted and said nothing about it in the docstring; tell me if you would rather it were excluded, or pinned as a known-wrong cell.

Verification: `tests/geometry/test_conversions.py` 396 passed, 1 skipped, 12 xfailed on cpu; module doctests 34 passed; `ruff check` and `ruff format --check` clean on both files.

Assisted-by: Claude Opus 5 (Claude Code), under my account. The measurements above were run here, not reasoned about.

ducha-aiki (approved):
Review of `78d8e3ce018eea123fbd10c1c2e4cd090ae5f63c`.

Both blockers from my 2026-08-28 review are resolved, and I re-measured each one instead of reading the diff. The probe guard gives the same answer in eager mode and in TorchScript for every dtype torch 2.14 ships. It also holds on the torch 2.5.1 floor and under `torch.compile(fullgraph=True)` with no graph break, and it costs nothing on the default path. The float8 cell and the `(3, 3)` cell are real pins. I'm approving the code. One text item has to land before merge: the PR description (below). You said you would update it after sign-off.

## Prior findings

**Blocker 1 (float8 over-rejection): resolved.** `conversions.py:2155` / `:2282`. Same probe on both trees, torch 2.14.0, cpu (`E` = eager, `S` = `torch.jit.script`):

```text
dtype             head E / S                                   main E / S
float8_e4m3fn     OK [0.5,0,-1] 0.6875 / same                  same / same
float8_e5m2       OK 0.625 / same                              same / same
float8_e4m3fnuz   OK / OK                                      OK / OK
float8_e5m2fnuz   OK / OK                                      OK / OK
float8_e8m0fnu    OK (wrong values, see below) / same          identical / identical
complex32/64/128  OK / OK                                      OK / OK
int64/32/16/8     ValueError ... Got torch.int64. / raises      OK [[0,0,-1],[0,0,-1],[0,0,1]] / same
uint8/16/32       ValueError / raises                          RuntimeError overflow / OK all-zero matrix
uint64            ValueError / raises                          TypeError / OK all-zero matrix
bool              ValueError / raises                          OK [[1,0,1],[0,1,1],[0,0,1]] / same
qint8/quint8/qint32  ValueError / raises                       RuntimeError store_scalar / RuntimeError
```

Every float and complex output is identical on the two trees. The scripted unsigned rows show a second silent defect that this PR also removes: on main, `torch.jit.script(normal_transform_pixel)(4, 5, dtype=torch.uint8)` returns an all-zero matrix, where eager raised an overflow error.

- **torch 2.5.1 overlay.** Importing and scripting both work. `float8_e8m0fnu` doesn't exist there, and nothing references it by name. The four float8 formats that do exist return the same values in eager and scripted mode as on main. Integers and bool raise. Scripted `warp_affine` still runs (`torch.Size([1, 1, 5, 5])`). `tests/geometry/test_conversions.py` on 2.5.1: 641 passed, 11 xfailed.
- **torch.compile.** `torch._dynamo.explain` reports `graphs=1 breaks=0` for 2-D and 3-D with `dtype` in {None, float32, complex32, float8_e4m3fn, float8_e5m2fnuz}, on 2.14.0 and 2.5.1. `fullgraph=True` compiles and returns the requested dtype. The float8 path adds one op to the graph (the zero-element `empty`: 23 ops vs 22 on main). Inductor removes it as dead code. An int64 call under `fullgraph=True` raises `Unsupported: Observed exception`, and a call without fullgraph raises the `ValueError`. The existing positive-size guard does the same on main (`f(0, 5)` → `Unsupported: Observed exception`), so this matches the file's convention.
- **Cost.** The guard alone, min of 9×20000 calls: `None` 0.007 us, `float32` 0.18 us, `complex128` 0.24 us, `float8_e4m3fn` 1.03 us (the probe allocates only off the fast path). Timing the whole function (min of 15×3000, interleaved main/head/main/head): `f(4,5)` 5.51/5.73/5.99/5.68 us and `f(4,5,dtype=float32)` 4.67/5.02/5.22/5.24 us. That is within noise. The float8 case is about +1.2 us. No action needed.

**Blocker 2 (the "no correct integer matrix" clause): resolved.** Both warnings and both code comments now say that a size of 3 pixels or fewer gives an integral scale and that the rejection is uniform by decision. `test_convention_integer_dtype_is_rejected_3959` pins the `(3, 3)` case.

**Nits 1–2: done.** When I planted an allowlist-only guard (no probe) in both functions of the PR worktree, `test_convention_non_default_float_and_complex_dtypes_are_accepted_3959[cpu-accepted_dtype3]` failed with `ValueError: dtype must be a floating point or complex type`. The float8 cell catches Blocker 1.

**Your `float8_e8m0fnu` question.** Leave it accepted in this PR. Main returns the same wrong matrix (translations `+1.0`, the 5-pixel scale `0.5`, zeros as `5.88e-39`), so this is not a regression. It is also a different rule from the integer-truncation defect this PR fixes. Rejecting it would need a sign/mantissa probe that TorchScript can express and that works on 2.5.1, where the name doesn't exist. That belongs in its own issue if anyone ever needs it.

## Findings

[P2] **PR description is stale** (text only; you said you would update it after sign-off). It still says "It is now a membership test, matching `_is_floating_point_dtype`", names a test that doesn't exist (`test_raises_on_non_fractional_dtype`), and keeps the retracted clause "There is no correct integer answer". The test log describes the old `test_grid.py` layout. Please rewrite it to match the probe guard, the three `_3959` tests in `test_conversions.py`, and the uniform-by-decision wording. Also change `Fixes/Relates to: #3959` to plain `Relates to #3959`. GraphQL `closingIssuesReferences` is empty today, but a line that starts with "Fixes" is an easy way to close #3959 through a squash message. #3959 has to stay open for the four functions this PR doesn't touch.

[P3] `tests/geometry/test_conversions.py:4050`: the acceptance test only calls the 2-D function. When I planted the allowlist-only guard in `normal_transform_pixel3d` alone, the PR worktree still gave `450 passed, 1 skipped, 7 xfailed` for `test_conversions.py` + `test_grid.py`, while `normal_transform_pixel3d(2, 4, 5, dtype=torch.float8_e4m3fn)` raised `ValueError`. One more `normal_transform_pixel3d` call in that test would pin the second site.

[P3] `changelog.d/3974.fixed.md` only mentions an "integer `dtype`". `bool` is also rejected now, and on main it silently returned `[[1, 0, 1], [0, 1, 1], [0, 0, 1]]`. Adding "or `bool`" makes the note complete.

Observation, no action: through `torch.jit.script` the message reads `... Got 4.` for `int64` (`0` for uint8, `11` for bool), because TorchScript formats a dtype as its enum value. Eager mode prints `Got torch.int64.` No internal caller passes `dtype=` (`normalize_homography`, `denormalize_homography` and `normalize_homography3d` call without it), so scripted `warp_affine` never reaches the guard.

**Documented-defect grep** (merge tree of `origin/main` + head): `#3959` hits in `kornia/` are at `conversions.py:1949` (normalize_homography), `:2098` and `:2237` (rewritten here), and `:2350`, `:2642`, `:2765` (functions this PR doesn't touch). The wart pins for those four still pass at head. #3959 is OPEN, and no closing references are parsed. The #3948 promotion-pin comment now points at the new convention test instead of the retired wart.

**Merge-order note** (repeating my first review, still true): #3979 is still open. It merges cleanly with main but conflicts with this PR in `tests/geometry/test_conversions.py`. Once both are applied, the `#3959` clauses at `conversions.py:1949` / `:2350` become false. Whichever PR merges second has to update them.

**What I ran**: `.venv` python 3.11, torch 2.14.0, plus the torch 2.5.1 overlay. Every probe printed `kornia.__file__` for the PR worktree or the main worktree (`53866348`).
- `tests/geometry/test_conversions.py tests/geometry/test_grid.py --device=cpu --dtype=float32,float64`: 703 passed, 2 skipped, 11 xfailed.
- Same file on torch 2.5.1: 641 passed, 11 xfailed.
- Plant-and-revert, main's `conversions.py` in the PR worktree, `-k "TestNormalTransformPixel or TestAngleAxisToQuaternion or test_wart_integer_input_raises_or_nans"`: 5 failed (four `is_rejected_3959` cells: int64/int32/bool `DID NOT RAISE`, uint8 `RuntimeError ... overflow`; plus `rejection_is_unconditional_3959`), 67 passed. The allowlist-only plants are described above. The worktree was restored and `git status --short` is empty.
- CPU half precision, `test_conversions.py` with `--xfail-known-failures`: float16 2 failed / 380 passed, bfloat16 2 failed / 388 passed. Both failures are `TestRadDegConversions::test_deg2rad` (float16 strict XPASS, bfloat16 closeness) and reproduce the same on main (372 / 380 passed there). They come from the arm64-vs-Linux manifest, not from this PR.
- Eager/script dtype matrix, `torch._dynamo.explain` plus `fullgraph=True` compiles, and guard and whole-function timings, as quoted above.

Issue #3938, opened 2026-08-10
convert_points_from_homogeneous divides by w + eps instead of w: 33% error at small positive w, 100% at small negative w


ducha-aiki:
`convert_points_from_homogeneous` divides by `z_vec + eps` rather than by `z_vec` (`kornia/geometry/conversions.py:200-203`):

```python
mask: torch.Tensor = torch.abs(z_vec) > eps
scale = torch.where(mask, 1.0 / (z_vec + eps), torch.ones_like(z_vec))
```

The guard `torch.where(|w| > eps, ...)` already handles the near-zero case, so adding `eps` to the denominator serves no protective purpose — it only biases the result. Because `eps` is added without regard to `sign(w)`, it *inflates* a small positive denominator and *shrinks* a small negative one, so the error is asymmetric in the sign of `w` and can reach 33 % / 100 % just past the threshold. Away from the threshold the result is still never exactly `numerator / w`; it carries a permanent relative bias of `-eps/w`.

The code comment on the lines above points at [opencv/opencv#14411](https://github.com/opencv/opencv/pull/14411/files) as the intended reference, and OpenCV's `convertPointsFromHomogeneous` uses `scale = (fabs(w) > FLT_EPSILON) ? 1./w : 1.` — i.e. plain `1/w`, with no epsilon in the denominator.

## Repro

```python
import torch
from kornia.geometry.conversions import convert_points_from_homogeneous, convert_points_to_homogeneous

f64 = torch.float64
for w in [2e-8, -2e-8, 2.0, -2.0]:
    p = torch.tensor([[2., 4., w]], dtype=f64)
    out = convert_points_from_homogeneous(p)
    print(f"w={w!r:>8} -> {out.tolist()}   exact={[2.0 / w, 4.0 / w]}")

rt = convert_points_from_homogeneous(convert_points_to_homogeneous(torch.tensor([[1.5, -2.5]], dtype=f64)))
print("to->from round-trip of [[1.5, -2.5]]:", rt.tolist())
```

## Observed

```
w=   2e-08 -> [[66666666.66666666, 133333333.33333331]]   exact=[100000000.0, 200000000.0]
w=  -2e-08 -> [[-200000000.0, -400000000.0]]   exact=[-100000000.0, -200000000.0]
w=     2.0 -> [[0.999999995, 1.99999999]]   exact=[1.0, 2.0]
w=    -2.0 -> [[-1.000000005, -2.00000001]]   exact=[-1.0, -2.0]
[...]

Issue #3949, opened 2026-08-10
NaN gradients at the identity: quaternion_to_axis_angle and axis_angle_to_quaternion differentiate through an unguarded sqrt(0)


ducha-aiki:
`quaternion_to_axis_angle` and `axis_angle_to_quaternion` both differentiate through an unguarded `sqrt(0)` at the identity rotation, so backward returns **NaN** at the single most common input in the domain: the identity quaternion `(1, 0, 0, 0)` and the zero axis-angle `(0, 0, 0)`.

```python
>>> q = torch.tensor([1., 0., 0., 0.], requires_grad=True); quaternion_to_axis_angle(q).sum().backward(); q.grad
tensor([nan, nan, nan, nan])
>>> v = torch.zeros(3, requires_grad=True); axis_angle_to_quaternion(v).sum().backward(); v.grad
tensor([nan, nan, nan])
```

The two sources:

```python
# quaternion_to_axis_angle, kornia/geometry/conversions.py:758-760
sin_squared_theta: torch.Tensor = q1 * q1 + q2 * q2 + q3 * q3
sin_theta: torch.Tensor = torch.sqrt(sin_squared_theta)

# axis_angle_to_quaternion, kornia/geometry/conversions.py:896-898
theta_squared: torch.Tensor = a0 * a0 + a1 * a1 + a2 * a2
theta: torch.Tensor = torch.sqrt(theta_squared)
```

In both cases the forward pass is fine — a subsequent `torch.where(… > 0.0, k_pos, k_neg)` selects the correct branch, so the returned *value* at the singularity is right. But `torch.where` propagates gradients through **both** arms, and `d/dx sqrt(x)` at `x = 0` is `inf`, so `inf * 0 = NaN` poisons the whole gradient. The fix pattern is already in this very file: `axis_angle_to_rotation_matrix` writes `torch.sqrt(theta2.clamp(min=1e-12))  # clamping to ensure no nan gradients` and returns a finite gradient at the same point.

## Repro

```python
import torch
from kornia.geometry.conversions import (
    axis_angle_to_quaternion as aa2q,
    axis_angle_to_rotation_matrix as aa2R,
    quaternion_to_axis_angle as q2aa,
)

q = torch.tensor([1.0, 0.0, 0.0, 0.0], dtype=torch.float64, requires_grad=True)
q2aa(q).sum().backward()
print("q2aa grad at the identity quaternion (1,0,0,0):", q.grad)

[...]

PR #3976, opened 2026-08-14
fix(geometry): divide by exactly w in convert_points_from_homogeneous

NoiceHax:

**Fixes/Relates to:** #3938

## What does this PR do?

`convert_points_from_homogeneous` divided by `z_vec + eps` instead of by `z_vec`. The `torch.where(abs(w) > eps, ...)` guard wrapped around it already covers the near-zero case, so the epsilon in the denominator protected nothing and only biased the answer. It was also added without looking at the sign of `w`, so it made a small positive denominator larger and a small negative one smaller. At `w = 2e-8` the result came out 33% low, and at `w = -2e-8` it came out 100% high, so two points differing only in the sign of `w` were wrong in opposite directions and by different amounts. Away from the threshold the bias was still on every point at `-eps/w`, which is why `convert_points_from_homogeneous(convert_points_to_homogeneous(p))` was an identity only to about 1e-8 relative even in float64.

This divides by `w`. It uses the double `torch.where` so the branch that is not selected never evaluates `1/0`, which keeps the backward pass free of NaN at `w = 0`. The `abs(w) <= eps` passthrough is untouched and still returns the numerator unchanged, matching OpenCV's `convertPointsFromHomogeneous`, which uses `scale = fabs(w) > FLT_EPSILON ? 1./w : 1.` (https://github.com/opencv/opencv/pull/14411/files). The `.. warning::` block that described the bias is retired, and the sentence about float16 underflow in it moved into the Convention block since that part still holds.

On the test side, `test_convention_divides_by_exactly_w` was already in the tree as a strict xfail, so the mark had to come off once it passes. The pin that recorded the old biased numbers is replaced by two positive checks: exact division for both signs of `w`, and the to/from round trip being an identity. Both assert with zero tolerance. Note that every output moves by about `eps/w` relative, so this is not byte identical with main by design, but it moves strictly toward the exact value. Downstream geometry tests behave the same as on main.

## Test log

`pixi` is not installed on this machine, so I ran the venv pytest directly, which is what the `pixi run test-module` task shells out to.

Fails on `main`. Here the source fix is reverted and only the new tests are kept, which is the point of the strict xfail that was already sitting in the tree:

```
$ git stash push kornia/geometry/conversions.py
$ .venv/Scripts/python.exe -m pytest -v tests/geometry/test_conversions.py \
    -k "divides_by_exactly_w or both_signs_of_w or roundtrip_from_to_homogeneous" --dtype=float32,float64

FAILED tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_convention_divides_by_exactly_w[cpu-float32] - AssertionError: Tensor-likes are not close!
Mismatched elements: 2 / 2 (100.0%)
Greatest absolute difference: 66666664.0 at index (0, 1) (up to 1e-05 allowed)
Greatest relative difference: 0.3333333134651184 at index (0, 0) (up to 0.0001 allowed)
FAILED tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_convention_divides_by_exactly_w[cpu-float64] - AssertionError: Tensor-likes are not close!
Greatest relative difference: 0.3333333333333334 at index (0, 0) (up to 1e-05 allowed)
FAILED tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_division_is_exact_for_both_signs_of_w[cpu-float32] - AssertionError: Tensor-likes are not equal!
Greatest relative difference: 0.40158772468566895 at index (0, 0)
FAILED tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_division_is_exact_for_both_signs_of_w[cpu-float64] - AssertionError: Tensor-likes are not equal!
Greatest relative difference: 0.401587697945215 at index (0, 0)
FAILED tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_roundtrip_from_to_homogeneous_is_identity[cpu-float64] - AssertionError: Tensor-likes are not equal!
Mismatched elements: 5 / 6 (83.3%)
Greatest relative difference: 9.999999892479057e-09 at index (2, 0)
================= 5 failed, 1 passed, 416 deselected in 5.90s =================
```

The one passing cell there is the float32 round trip, because `1.0 + 1e-8` rounds back to `1.0` in float32, so that dtype never saw the bias. The float64 cell is the one that catches it.

Passes with the fix:

```
$ .venv/Scripts/python.exe -m pytest -v tests/geometry/test_conversions.py -k "FromHomogeneous" --dtype=float32,float64

tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_cardinality[cpu-float32-batch_shape0] PASSED [  5%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_cardinality[cpu-float32-batch_shape1] PASSED [ 10%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_cardinality[cpu-float32-batch_shape2] PASSED [ 15%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_cardinality[cpu-float32-batch_shape3] PASSED [ 20%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_cardinality[cpu-float64-batch_shape0] PASSED [ 25%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_cardinality[cpu-float64-batch_shape1] PASSED [ 30%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_cardinality[cpu-float64-batch_shape2] PASSED [ 35%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_cardinality[cpu-float64-batch_shape3] PASSED [ 40%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_points[cpu-float32] PASSED [ 45%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_points[cpu-float64] PASSED [ 50%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_points_batch[cpu-float32] PASSED [ 55%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_points_batch[cpu-float64] PASSED [ 60%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_gradcheck[cpu] PASSED [ 65%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_gradcheck_zvec_zeros[cpu] PASSED [ 70%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_convention_divides_by_exactly_w[cpu-float32] PASSED [ 75%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_convention_divides_by_exactly_w[cpu-float64] PASSED [ 80%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_division_is_exact_for_both_signs_of_w[cpu-float32] PASSED [ 85%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_division_is_exact_for_both_signs_of_w[cpu-float64] PASSED [ 90%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_roundtrip_from_to_homogeneous_is_identity[cpu-float32] PASSED [ 95%]
tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_roundtrip_from_to_homogeneous_is_identity[cpu-float64] PASSED [100%]

===================== 20 passed, 402 deselected in 3.43s ======================
```

Whole file, which also covers the float16 wart pin that this change deliberately leaves alone:

```
$ .venv/Scripts/python.exe -m pytest -q tests/geometry/test_conversions.py --dtype=float32,float64

======================= 417 passed, 5 xfailed in 1.09s ========================
```

The 5 remaining xfails belong to other open issues, not to this one.

Downstream callers. Same numbers on this branch and on `main`, so the one failure and the six xpasses are pre-existing and unrelated:

```
$ .venv/Scripts/python.exe -m pytest -q tests/geometry/test_linalg.py tests/geometry/epipolar tests/geometry/camera --dtype=float32,float64

# this branch
FAILED tests/geometry/epipolar/test_essential.py::TestDecomposeEssentialMatrixNoSVD::test_correct_decompose
1 failed, 785 passed, 1 skipped, 2 xfailed, 6 xpassed, 44 warnings in 27.58s

# main, same command
FAILED tests/geometry/epipolar/test_essential.py::TestDecomposeEssentialMatrixNoSVD::test_correct_decompose
1 failed, 785 passed, 1 skipped, 2 xfailed, 6 xpassed, 44 warnings in 17.54s
```

Doctests and lint:

```
$ .venv/Scripts/python.exe -m pytest -q --doctest-modules kornia/geometry/conversions.py
============================= 32 passed in 0.45s ==============================

$ ruff 0.16.1 check kornia/geometry/conversions.py tests/geometry/test_conversions.py
All checks passed!
$ ruff 0.16.1 format --check kornia/geometry/conversions.py tests/geometry/test_conversions.py
2 files already formatted
```

One gap I want to be upfront about: I did not run the `test_dynamo` cell with `--optimizer=inductor` locally. Inductor needs a compiler toolchain that this Windows box does not have set up, and the run did not finish. The change adds no new control flow, it swaps one `torch.where` argument for a `torch.where`-guarded reciprocal, so I expect the CI dynamo job to cover it.

- [ ] ðŸŸ¢ **Human-written**

An agent wrote the patch and the tests, and I reviewed the result and ran every test log pasted above myself.

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -252,33 +252,21 @@ def convert_points_from_homogeneous(points: torch.Tensor, eps: float = 1e-8) -> 
           coordinate ``w`` and is dropped from the output
         - requires rank :math:`\geq 2`: a bare :math:`(D,)` point raises
           ``ValueError``
-        - when ``abs(w) > eps`` the remaining components are divided by ``w``
-          with its sign kept (``[[2., 4., -2.]]`` gives ``[[-1., -2.]]``), up to
-          the bias described in the warning below
+        - when ``abs(w) > eps`` the remaining components are divided by exactly
+          ``w``, with its sign kept (``[[2., 4., -2.]]`` gives ``[[-1., -2.]]``)
         - when ``abs(w) <= eps`` (default ``eps = 1e-8``; the test is a strict
           ``>``) the numerator is instead returned **unchanged**, following
           OpenCV: ``[[2., 4., 0.]]`` gives ``[[2., 4.]]``
-
-    .. warning::
-        The division is by ``w + eps`` rather than by ``w``, and ``eps`` is
-        added without regard to the sign of ``w``, so the **signed** relative
-        error of the result is exactly ``-eps / (w + eps)``. At ``w = 2e-8``
-        that is ``-1/3``: the exact result ``[1e8, 2e8]`` comes out as
-        ``[6.67e7, 1.33e8]`` (33 % low). At ``w = -2e-8`` it is ``+1``:
-        ``[-1e8, -2e8]`` comes out as ``[-2e8, -4e8]`` (100 % high). Only for
-        ``abs(w)`` much larger than ``eps`` does it reduce to the familiar
-        ``-eps / w``, and there it is usually below the rounding of the working
-        dtype — at ``w = 2`` the measured error is ``-5.0e-09`` in ``float64``,
-        while in ``float32`` ``2 + eps`` rounds back to ``2`` and the result is
-        exact. The numbers above assume ``eps`` is representable: in
-        ``float16`` both the default ``eps`` and ``w = 2e-8`` underflow to
-        ``0``, so ``[[2., 4., 2e-8]]`` takes the ``abs(w) <= eps`` pass-through
-        branch and returns ``[[2., 4.]]``. Tracked in
-        `#3938 <https://github.com/kornia/kornia/issues/3938>`_.
+        - ``eps`` only picks the branch, it never enters the division, so the
+          result carries no bias from it. Note that ``eps`` is compared against
+          ``w`` in the working dtype: in ``float16`` the default ``eps`` and
+          ``w = 2e-8`` both underflow to ``0``, so ``[[2., 4., 2e-8]]`` takes
+          the pass-through branch there and returns ``[[2., 4.]]``
 
     Args:
         points: the points to be transformed of shape :math:`(*, N, D)`.
-        eps: to avoid division by zero.
+        eps: threshold on ``abs(w)`` below which the point is returned
+            unchanged instead of divided, to avoid division by zero.
 
     Returns:
         the points in Euclidean space :math:`(*, N, D-1)`.
@@ -302,7 +290,8 @@ def convert_points_from_homogeneous(points: torch.Tensor, eps: float = 1e-8) -> 
     # follow the convention of opencv:
     # https://github.com/opencv/opencv/pull/14411/files
     mask: torch.Tensor = torch.abs(z_vec) > eps
-    scale = torch.where(mask, 1.0 / (z_vec + eps), torch.ones_like(z_vec))
+    safe_z_vec = torch.where(mask, z_vec, torch.ones_like(z_vec))
+    scale = torch.where(mask, 1.0 / safe_z_vec, torch.ones_like(z_vec))
 
     return scale * points[..., :-1]
 
--- a/tests/geometry/test_conversions.py
+++ b/tests/geometry/test_conversions.py
@@ -1115,6 +1115,28 @@ class TestConvertPointsFromHomogeneous(BaseTester):
 
         self.assert_close(cpfh(cpth(points)), points, atol=0.0, rtol=0.0)
 
+    def test_convention_the_masked_division_keeps_the_gradient_finite(self, device):
+        # Companion to the forward pins above, for the half of the fix they cannot see. The old
+        # `1.0 / (z_vec + eps)` is evaluated for EVERY point, including the ones the mask discards.
+        # At w == -eps that denominator is exactly zero, so the reciprocal is inf; torch.where drops
+        # that lane from the forward value but the backward pass still multiplies it by the zero
+        # cotangent and 0 * inf is NaN. Measured on the unfixed code the gradient at this input is
+        # [1., 1., nan]; the double-`where` makes it [1., 1., 0.].
+        # w is the exact pole rather than merely a small value: a nearby w only makes the reciprocal
+        # large, which no assertion on finiteness would catch. eps is passed explicitly so the input
+        # tracks the guard rather than the default. float64 only -- at float16 the default eps
+        # underflows to 0 (pinned in test_wart_float16_underflowed_default_eps_flips_branches), so
+        # w == -eps is not the pole there.
+        eps = 1e-8
+        points = torch.tensor([[2.0, 4.0, -eps]], device=device, dtype=torch.float64, requires_grad=True)
+
+        kornia.geometry.conversions.convert_points_from_homogeneous(points, eps=eps).sum().backward()
+
+        assert torch.isfinite(points.grad).all(), (
+            "kornia#3938: convert_points_from_homogeneous divides by exactly zero at w == -eps, so "
+            f"the discarded branch poisons the gradient: {points.grad.tolist()}"
+        )
+
 
 def _skip_if_mps_clamp_caching(device):
     # Runtime probe instead of a torch-version pin, so the skip retires itself on any torch
--- a/tests/geometry/test_conversions.py
+++ b/tests/geometry/test_conversions.py
@@ -1063,22 +1063,15 @@ class TestConvertPointsFromHomogeneous(BaseTester):
 
         self.assert_close(actual, expected)
 
-    @pytest.mark.xfail(
-        raises=AssertionError,
-        reason="convert_points_from_homogeneous divides by w + eps — kornia#3938",
-        strict=True,
-    )
     def test_convention_divides_by_exactly_w(self, device, dtype):
-        # Intended behavior: for |w| > eps the point is divided by exactly w. It currently is
-        # divided by w + eps with no regard for sign, so the signed relative error is exactly
-        # -eps / (w + eps): -1/3 at w = +2e-8 (33 % low) and +1 at w = -2e-8 (100 % high) (#3938).
-        # Marked xfail(strict=True) so fixing #3938 makes this XPASS and forces the mark out.
+        # For |w| > eps the point is divided by exactly w, matching OpenCV's
+        # convertPointsFromHomogeneous (scale = fabs(w) > FLT_EPSILON ? 1./w : 1.), see
+        # https://github.com/opencv/opencv/pull/14411/files. This used to divide by w + eps,
+        # which made the result 33 % low here (#3938).
         # Snippet used to generate expected (by hand):
-        #   2 / 2e-8, 4 / 2e-8 -> [1e8, 2e8]  (kornia returns [6.6666667e7, 1.3333333e8])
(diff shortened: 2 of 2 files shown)


ducha-aiki (dismissed):
This is the model for how a W1b fix PR should arrive. You dropped the strict `xfail` and kept the convention test as the permanent guard, turned the wart pin into a positive assertion instead of deleting it outright, replaced the sampled figures with values that are exact in binary, added an oracle-free round-trip invariant, and removed the `#3938` warning from the docstring in the same change. Nothing left for me to ask for on the protocol side.

Verified by applying your diff to current `main` (`35555b55`), not by reading it.

## Verified

```text
                                        main                     with this PR
roundtrip cpfh(cpth(p)) in f64   max err 4.75e-08          exact, torch.equal -> True
divide by +2**-26                [80317539.6, 160635079.2]  [134217728.0, 268435456.0]  = [2**27, 2**28]
divide by -2**-26                [-408066562.4, -816133124.7] [-134217728.0, -268435456.0]
```

The two-sign case is the one that makes the case for the fix: on `main` the same magnitude of `w` gives 40 % low for `+w` and 104 % high for `-w`, errors in opposite directions for inputs that differ only in a sign. Your choice of `w = 2**-26` is right and better than the `2e-8` it replaces — it clears the default `eps` (`1.49e-8 > 1e-8`) and makes both quotients powers of two, which is what earns the `atol=rtol=0` assertion. Good instinct: the old pin's `rtol=1e-2` was carrying sampled decimals that nothing re-executed.

Also confirmed:

- `tests/geometry` with your diff applied: **0 failures**, same as `main`.
- Both new assertions are genuine pins — `test_convention_divides_by_exactly_w` and `test_division_is_exact_for_both_signs_of_w` fail on unpatched `main`; `test_roundtrip_from_to_homogeneous_is_identity` fails at float64 (it passes at float32, since `1 + 1e-8` rounds back to `1` there — CI runs both dtypes so it is covered, just be aware a local default-dtype run will not show it).
- `grep -rnE "#3938|issues/3938" kornia/` after your change: **no hits**. The three surviving references in `tests/` are the "this used to be ..." comments on the new pins, which is exactly where they belong.
- `ruff check`, `ruff format --check`, and `--doctest-modules kornia/geometry/conversions.py` (34 passed) all clean.

## Your change fixes a second defect you did not claim — please pin it

Rewriting `1.0 / (z_vec + eps)` as a masked `torch.where` over a `safe_z_vec` also removes a NaN gradient. At `w = -eps` exactly, the old denominator `z_vec + eps` is exactly zero, the reciprocal is `inf`, and although `torch.where` discards that lane in the forward pass it still propagates `0 * inf = NaN` in the backward pass:

```text
x = [[2.0, 4.0, -1e-8]], float64, requires_grad
main:         x.grad = [[1.0, 1.0, nan]]
with this PR: x.grad = [[1.0, 1.0, 0.0]]
```

This is the same defect class as `#3949` and `#3966` — a guard that fixes the forward value while the infinite derivative flows through the unselected branch — and the double-`where` you wrote is precisely the established fix pattern for it. It is currently unpinned and unmentioned. Please add a gradient assertion at `w = -eps` (the exact pole, not merely a small `w`) and one sentence in the PR body. Without a pin, someone later "simplifies" the two `where` calls back into one and the NaN returns silently, because no forward-value test can see it.

## Rebase

Your base is `8454a81a` and CI last ran on 2026-08-15, so this branch has never executed the pins added by `#3967`, `#3984`, `#4006` or `#4028`. Everything above was measured with your diff applied to current `main`, so I know it is clean there — but please rebase so the PR's own CI says so too. It applies without conflict; `maintainer_can_modify` is on if you would rather I push it.

Approving. Add the gradient pin and this is ready to merge.

---

ducha-aiki (approved):
Pushed the gradient pin I asked for rather than leaving you a round-trip — head is `06102d88`, one commit on top of yours, no rebase needed.

`test_convention_the_masked_division_keeps_the_gradient_finite` asserts at the **exact pole**, `w == -eps`, where the old `1.0 / (z_vec + eps)` denominator is exactly zero. A nearby small `w` only makes the reciprocal large, which no finiteness assertion would catch — the pole is the whole point. `eps` is passed explicitly so the input tracks the guard rather than the default, and it is float64-only because at float16 the default `eps` underflows to 0, so `w == -eps` is not the pole there (already pinned by `test_wart_float16_underflowed_default_eps_flips_branches`).

Verified it is a genuine regression test, not a tautology: **1 failed** on the pre-fix library, **1 passed** with your change. `tests/geometry` 2125 passed / 0 failed on the pushed head; `ruff check` and `ruff format` clean.

Everything else in this PR was already right — dropping the strict `xfail` while keeping the convention test, turning the wart into a positive assertion, and picking `w = 2**-26` so both quotients are powers of two and the assertion can run at `atol=rtol=0`. That last one is the detail I'd point other contributors at.

Re-approving; the earlier approval was dismissed by my push. Please refresh the PR body to mention the gradient fix — it is the second defect this closes.

---

ducha-aiki (approved):
Final verification round, `06102d88` merged onto current `main` (the PR base was 12 commits behind).

- `grep -rnE "#3938|issues/3938" kornia/ docs/` is empty; the four surviving hits are all `tests/` pins naming their issue.
- `tests/geometry/test_conversions.py --dtype=float32,float64`: 607 passed, 3 skipped. The `Homogeneous` block also passes at float16 and bfloat16. Doctests pass; ruff clean.
- `2 ** -26` is a good choice for the both-signs pin: it is exactly representable in bfloat16 (same exponent range as float32) and the quotients are `2 ** 27` / `2 ** 28`, so `atol=rtol=0.0` is honest there and not just at float32/float64.

Two non-blocking nits, both prose/consistency rather than behaviour:

1. `test_convention_the_masked_division_keeps_the_gradient_finite` hardcodes `torch.float64` but does not call `_skip_if_dtype_unavailable(device, torch.float64)`. Every other float64-hardcoded pin in the file does (`test_convention_axis_angle_quaternion_roundtrip_is_exact_in_float64`, both `_3949` pins, both `_3951` pins), and that helper's own comment gives the reason: these pins drop the dtype fixture on purpose so they run everywhere, which means they also run on a backend with no float64. Not CI-visible — MPS is not in the matrix — but it would fail rather than skip there.
2. After the merge, `test_wart_default_eps_1e_8_backs_the_remaining_quoted_warning_numbers` still says the two APIs "use eps **numerically**". For `convert_points_from_homogeneous` that is now only true of the branch comparison, which is the point of this PR. The assertion itself stays correct and still load-bearing (the float16 sentence quotes the default), so it is one word.

Neither blocks. Approving.

ducha-aiki (approved):

Issue #3956, opened 2026-08-10
@deprecated rewrites the process-global warning filters: -W error::DeprecationWarning stops working, for the kornia call and for every later unrelated warning


ducha-aiki:
Calling **any** `@deprecated`-decorated kornia symbol rewrites the interpreter's global warning filters. Two consequences, both executed below:

1. `-W error::DeprecationWarning` does **not** turn kornia's own deprecation warning into an error — the decorator overrides the user's filter before emitting.
2. Every **subsequent, unrelated** `DeprecationWarning` in the same process is silenced to `default` as well — the user's warning configuration is permanently replaced, process-wide, by a side effect of calling a deprecated function.

The cause is in `kornia/core/_compat.py:131-150`:

```python
def _emit_deprecation_warning(
    name: str, replace_with: Optional[str], version: Optional[str], extra_reason: Optional[str]
) -> None:
    """Emit a deprecation warning with the given parameters."""
    beginning = f"Since kornia {version} the " if version is not None else ""
    extra = f" {extra_reason}" if extra_reason else ""

    warnings.simplefilter("always", DeprecationWarning)
    try:
        ...
        warnings.warn(msg, category=DeprecationWarning, stacklevel=3)
    finally:
        warnings.simplefilter("default", DeprecationWarning)
```

`warnings.simplefilter` **prepends a process-global filter entry**; it is not scoped to the call and it is not undone. The `finally` clause does not restore anything — it prepends a *second* global entry (`default`), which is why the user's own `error` filter is shadowed from that point on. The correct primitive for a temporary, scoped change is `with warnings.catch_warnings():`, which snapshots and restores `warnings.filters`.

This is not specific to any one function: `@deprecated` is applied 63 times across `kornia/__init__.py`, `kornia/utils/__init__.py`, `kornia/filters/gaussian.py`, `kornia/filters/kernels.py` and `kornia/geometry/conversions.py`. One call to any of them is enough.

## Repro

`repro.py`:

```python
import warnings

import torch

import kornia.geometry.conversions as C

print("filters[:2] before:", [(f[0], getattr(f[2], "__name__", f[2])) for f in warnings.filters[:2]])

try:
    C.angle_axis_to_quaternion(torch.tensor([[0.0, 0.0, 1.0]]))
    print("kornia deprecated call under -W error::DeprecationWarning -> NO exception raised")
except DeprecationWarning as e:
    print("kornia deprecated call -> DeprecationWarning raised as an error:", e)

[...]

PR #3977, opened 2026-08-15
fix(core): stop @deprecated from rewriting the global warning filters

NoiceHax:

**Fixes/Relates to:** #3956

## What does this PR do?

`_emit_deprecation_warning` in `kornia/core/_compat.py` wrapped its `warnings.warn` call in `warnings.simplefilter("always", DeprecationWarning)`, with a `finally` that called `warnings.simplefilter("default", DeprecationWarning)`. `simplefilter` prepends an entry to the process-global `warnings.filters` list, so neither call was scoped to the function. The first one sat in front of whatever filter the caller had set, and the `finally` did not undo it, it prepended a second entry that then stayed for the rest of the process.

Two things followed from that. Running under `-W error::DeprecationWarning` did not turn kornia's own deprecation warning into an error, so downstream projects had no way to fail their CI on deprecated kornia API usage. Every later `DeprecationWarning` in the same process, from any library, was also downgraded, because the leaked `default` entry kept shadowing the user's `error` filter. Calling any one of the 63 `@deprecated` symbols was enough to trigger it.

This PR deletes both `simplefilter` calls and the `try/finally`, and warns directly, so the caller's filters decide. Two new tests in `tests/utils/test_deprecated.py` fail on `main`: `test_warning_can_be_escalated_to_an_error`, where `main` does not raise, and `test_warning_does_not_mutate_the_global_filters`, where `main` leaves two extra entries in `warnings.filters`. Both run inside `catch_warnings()` so the mutation cannot leak into the rest of the suite.

One deliberate behavior change to flag, since it is easy to miss in the diff. Under Python's default filters a `DeprecationWarning` is only shown when it comes from `__main__`, so dropping the forced "always" makes these warnings less visible in library-internal call chains. That is the normal tradeoff, it is what other libraries do, and it is the side the issue argues for, because a warning that cannot be escalated to an error is of little use to the people who most need it.

## Test log

`pixi` is not installed on this machine, so these ran on the repo `.venv` interpreter directly, which is the same environment `pixi run test-module` shells out to via `uv run pytest`.

```
$ python -m pytest tests/utils/test_deprecated.py
======================== 30 passed, 1 warning in 1.76s ========================

$ python -m pytest tests/geometry/test_conversions.py --dtype=float32,float64
======================= 413 passed, 7 xfailed in 2.86s ========================

$ python -m pytest tests/test_import.py tests/test_api_surface.py
============================= 14 passed in 5.45s ==============================
```

The same two new tests with `main`'s `kornia/core/_compat.py` restored, to show they fail without the fix:

```
$ git checkout upstream/main -- kornia/core/_compat.py
$ python -m pytest tests/utils/test_deprecated.py -k "escalated or mutate"
tests/utils/test_deprecated.py FF                                        [100%]
FAILED tests/utils/test_deprecated.py::TestDeprecatedWrappers::test_warning_can_be_escalated_to_an_error
FAILED tests/utils/test_deprecated.py::TestDeprecatedWrappers::test_warning_does_not_mutate_the_global_filters
================ 2 failed, 28 deselected, 2 warnings in 1.82s =================
```

The first fails with `Failed: DID NOT RAISE DeprecationWarning`, the second with `warnings.filters` holding `[('default', DeprecationWarning), ('always', DeprecationWarning)]` where it should hold `[]`.

`ruff check` and `ruff format --check` are both clean on the two changed files.

- [ ] ðŸŸ¢ **Human-written**

An agent wrote the fix, the tests and this description; I reviewed the change, reproduced the bug and ran the tests above myself, and I can answer questions on it.

### DIFF
--- a/kornia/core/_compat.py
+++ b/kornia/core/_compat.py
@@ -135,19 +135,15 @@ def _emit_deprecation_warning(
     beginning = f"Since kornia {version} the " if version is not None else ""
     extra = f" {extra_reason}" if extra_reason else ""
 
-    warnings.simplefilter("always", DeprecationWarning)
-    try:
-        if replace_with is not None:
-            msg = f"{beginning}`{name}` is deprecated in favor of `{replace_with}`.{extra}"
-        else:
-            msg = f"{beginning}`{name}` is deprecated and will be removed in the future versions.{extra}"
-        warnings.warn(
-            msg,
-            category=DeprecationWarning,
-            stacklevel=3,
-        )
-    finally:
-        warnings.simplefilter("default", DeprecationWarning)
+    if replace_with is not None:
+        msg = f"{beginning}`{name}` is deprecated in favor of `{replace_with}`.{extra}"
+    else:
+        msg = f"{beginning}`{name}` is deprecated and will be removed in the future versions.{extra}"
+    warnings.warn(
+        msg,
+        category=DeprecationWarning,
+        stacklevel=3,
+    )
 
 
 def deprecated(
--- a/tests/geometry/test_conversions.py
+++ b/tests/geometry/test_conversions.py
@@ -6445,15 +6445,17 @@ class TestAxisAngleToRotationMatrix:
 
 # Module-level pins for kornia#3956: the four deprecated aliases of this module
 # (angle_axis_to_rotation_matrix, rotation_matrix_to_angle_axis, quaternion_to_angle_axis,
-# angle_axis_to_quaternion) are not the site of the defect -- the root cause is
-# _emit_deprecation_warning in kornia/core/_compat.py, which wraps its warnings.warn call in
+# angle_axis_to_quaternion) were never the site of the defect -- the root cause was
+# _emit_deprecation_warning in kornia/core/_compat.py, which wrapped its warnings.warn call in
 # warnings.simplefilter("always", DeprecationWarning) / simplefilter("default", DeprecationWarning)
-# and so rewrites the PROCESS-GLOBAL filter list on every call. Every deprecated symbol in kornia
-# is therefore affected, which is why these live at module level rather than in one symbol's class,
-# following the module-level wart precedent above. Both pins run inside warnings.catch_warnings()
-# so that the filter mutation they provoke cannot leak into the rest of the suite: this bug is
-# contagious across tests, and an unisolated pin would silently disarm every other test's warning
-# discipline.
+# and so rewrote the PROCESS-GLOBAL filter list on every call. Every deprecated symbol in kornia
+# was therefore affected, which is why these live at module level rather than in one symbol's
+# class, following the module-level wart precedent above; the emitter itself is pinned in
+# tests/utils/test_deprecated.py, and these four cells are what keeps the convention attached to
+# the aliases rather than to one caller of the decorator. Both pins run inside
+# warnings.catch_warnings() so that a regression's filter mutation cannot leak into the rest of the
+# suite: that bug was contagious across tests, and an unisolated pin would silently disarm every
+# other test's warning discipline.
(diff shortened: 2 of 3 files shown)


ducha-aiki:
Reviewed by execution. **The fix is correct and I could reproduce every claim in your description.**

What I verified, on your head `926ba05c`, against the repo `.venv` (kornia 0.8.3, torch 2.12.0):

- `-W error::DeprecationWarning` now escalates the kornia warning — confirmed.
- The collateral damage is gone: after a kornia deprecated call, an *unrelated* third-party `DeprecationWarning` still raises. On `main` it does not, and `warnings.filters` grows 12 → 14. That second-order effect is the worst part of #3956 and your diff removes it cleanly.
- `tests/utils/test_deprecated.py`: 30 passed.
- No `simplefilter` / `filterwarnings` / `resetwarnings` / `showwarning` sites remain anywhere under `kornia/` — I grepped with a positive control to make sure the empty result was real.

Two things to action:

**1. Your base is `8454a81a` (batch 3a) — three merged commits behind `main`.** Please rebase. This matters more than usual right now: `#3967` and `#3984` landed after your base, and they add convention/wart pins to `tests/geometry/test_conversions.py`. Your PR doesn't touch that file so I don't expect a conflict, but rebasing is what puts your work in front of the current test suite.

**2. The visibility caveat in your description is currently prose-only.** You flag, honestly, that dropping the forced `"always"` makes these warnings less visible in library-internal call chains. I tried to measure that and could **not** observe the predicted regression: in fresh-process runs, a kornia deprecation raised from a non-`__main__` module stayed visible both before and after your change.

I want to be straight about the limits of that, because I chased it a long way and it bit me. The observed behavior flips depending on whether the entry point is a script file or `python -c`, with byte-identical library code — so my harness is a confound and I am **not** claiming your caveat is wrong. What I am confident of is that it is unpinned. Given that this repo has just spent three batches learning that an unpinned prose claim is the defect that survives review, the caveat deserves a test that states the intended visibility semantics, whatever you determine them to be. If you'd rather not expand scope, say so and drop the paragraph instead — an unverified caveat in a merge commit is worse than no caveat.

Nothing else blocking. Thank you for the test-first structure and for the honest 🔴 disclosure; the "here is the same test failing on main" section made this much faster to check.


---

ducha-aiki:
Thanks — the fix is right. Dropping the two `simplefilter` calls is the correct shape: the emitter had no business touching a process-global list to emit one warning.

I rebased the branch onto current `main` and pushed one commit on top of yours (`4170c6b5`). The branch was based on `8454a81a`, before `#3967` landed the `#3956` pins in `tests/geometry/test_conversions.py`. Those pins were written against the filter-rewriting emitter, so your fix turned all eight of them red — which is what they're for:

- `test_convention_deprecated_alias_warning_can_be_escalated_to_an_error_3956` is `xfail(strict=True)`; your fix made its four cells XPASS, which is a failure. Un-xfailed.
- `test_wart_deprecated_alias_rewrites_the_global_warning_filters_3956` asserted the two entries the emitter used to push. It became `test_convention_deprecated_alias_leaves_the_global_warning_filters_alone_3956`, asserting the filter list comes back exactly as it was found.

The two halves stay separate pins on purpose: a partial fix that dropped only the `"always"` call would restore escalation and still leave the caller's filters clobbered by the `finally`, so one pin can't cover both.

Your two tests in `tests/utils/test_deprecated.py` are kept as they are. They pin the emitter itself, which is where the fix lives; the four geometry cells pin the same convention at the deprecated aliases, so a future decorator that emits its own warning still has to honour it.

Verified: full CPU test suite green — **8103 passed**, 3535 skipped, 46 xfailed, 0 failures. ruff clean.

ducha-aiki (approved):

ducha-aiki:
Thank you for the fix!

Issue #3957, opened 2026-08-10
normal_transform_pixel: size == 1 gives a 2e14 scale and size == 0 silently mirrors - the executed root cause behind #3929's NaN crops


ducha-aiki:
> **Scope note, please read first.** This issue does not propose changing the `2/(size − 1)` scaling or the corner-aligned convention. **#3904 remains reserved for its assigned contributor**; the defect here is the degenerate-denominator branch (`size == 1` / `size == 0` / negative), which is wrong under **every** `align_corners` setting. The NaN reproduced below is an `align_corners=True` execution, i.e. it is present even in the branch #3904 leaves alone, and fixing #3904 as described in its body would not touch the `size == 1` / `size == 0` code at all. No code change to the scaling formula is proposed here.

`normal_transform_pixel` guards its denominator with an epsilon substitution instead of a range check (`kornia/geometry/conversions.py:1318-1322`):

```python
# prevent divide by zero bugs
width_denom: float = eps if width == 1 else width - 1.0
height_denom: float = eps if height == 1 else height - 1.0

sx: float = 2.0 / width_denom
sy: float = 2.0 / height_denom
```

with `eps = 1e-14`. Three separate failures follow, all silent and all `isfinite`:

- **`size == 1`** → `2 / 1e-14 = 2e14`. Not an error, not a finite fallback — a scale factor fourteen orders of magnitude out, which downstream becomes NaN.
- **`size == 0`** → `2 / (0 − 1) = −2.0`, a **sign-flipped (mirroring)** normalization. This is the one case that is literally a divide-by-zero candidate, and it is the one case `eps` does *not* guard: `eps` is consulted only in the `== 1` branch. The docstring says `eps: epsilon to prevent divide-by-zero errors`.
- **negative sizes** → accepted silently (`width = −3` → `sx = −0.5`).

`normal_transform_pixel3d` has the same three branches with the same behavior, and `normalize_homography`, `denormalize_homography` and `normalize_homography3d` all call these helpers, so the degeneracy propagates through the public API.

## Repro

```python
import torch
from kornia.geometry.conversions import (
    denormalize_homography,
    normal_transform_pixel,
    normal_transform_pixel3d,
    normalize_homography,
    normalize_homography3d,
)
from kornia.geometry.transform import crop_by_transform_mat, warp_perspective

print("# 1. the helper itself")
print("normal_transform_pixel(4, 1) =\n", normal_transform_pixel(4, 1))
[...]

PR #3978, opened 2026-08-15
fix(geometry): allocate axis_angle_to_quaternion output from the computed values

NoiceHax:
## Which lane?
- [x] ðŸŸ¢ **Green lane** (test-first bug fix, no prior issue needed)
- [ ] ðŸŸ  **Discuss-first**

**Fixes/Relates to:** #3948

## What does this PR do?

`axis_angle_to_quaternion` allocated its output buffer with the dtype of the input. The values written into it come from `sqrt`, `sin` and `cos`, which promote an integer input to float32, so for an integer axis angle those floats were cast back down on assignment and every component truncated to zero. `axis_angle_to_quaternion(torch.tensor([1, 0, 0]))` returned `tensor([0, 0, 0, 0])` with no error.

The buffer is now allocated with the dtype of the computed values. The same input returns `tensor([0.8776, 0.4794, 0.0000, 0.0000])`, which matches what `quaternion_from_euler` gives for the equivalent integer input. Float inputs are unchanged, since the computed dtype is the input dtype for float16, bfloat16, float32 and float64.

`test_integer_input` covers int16, int32, int64 and uint8 and fails on `main` with the all zero output. `test_float_input_keeps_dtype` puts the float behaviour on record. Issue #3948 also offers raising a clear `TypeError` instead, so tell me if you prefer the loud rejection and I will swap it.

## Test log

```
$ .venv/Scripts/python.exe -m pytest -v tests/geometry/test_conversions.py -k "AngleAxisToQuaternion or QuaternionToAngleAxis" --dtype=float32,float64
tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_integer_input[cpu-input_dtype0] PASSED [ 42%]
tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_integer_input[cpu-input_dtype1] PASSED [ 44%]
tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_integer_input[cpu-input_dtype2] PASSED [ 46%]
tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_integer_input[cpu-input_dtype3] PASSED [ 48%]
tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_float_input_keeps_dtype[cpu-input_dtype0] PASSED [ 50%]
tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_float_input_keeps_dtype[cpu-input_dtype1] PASSED [ 51%]
tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_float_input_keeps_dtype[cpu-input_dtype2] PASSED [ 53%]
tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_float_input_keeps_dtype[cpu-input_dtype3] PASSED [ 55%]
===================== 54 passed, 374 deselected in 1.96s ======================
```

Same tests with the source change reverted, on `main`:

```
FAILED tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_integer_input[cpu-input_dtype0] - assert False
 + where False = tensor([0, 0, 0, 0], dtype=torch.int16).is_floating_point()
FAILED tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_integer_input[cpu-input_dtype1] - assert False
FAILED tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_integer_input[cpu-input_dtype2] - assert False
FAILED tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_integer_input[cpu-input_dtype3] - assert False
================= 4 failed, 4 passed, 233 deselected in 1.94s =================
```

Whole file plus `tests/geometry/test_quaternion.py`: 597 passed, 7 xfailed. Doctests on `kornia/geometry/conversions.py`: 32 passed. The commands run the repo `.venv` directly, which is what the `pixi` test tasks shell out to.

## AI Usage Disclosure
- [ ] ðŸŸ¢ **Human-written**
- [x] ðŸŸ¡ **AI-assisted** â€” AI helped with the fix and the tests, I reviewed the change and ran the tests above
- [ ] ðŸ”´ **AI-generated**

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -1219,20 +1219,16 @@ def axis_angle_to_quaternion(axis_angle: torch.Tensor) -> torch.Tensor:
         - ``angle_axis_to_quaternion`` is the deprecated alias of this function
           since 0.7.0; see the alias warning on
           :func:`~kornia.geometry.conversions.axis_angle_to_rotation_matrix`
+        - an integral or boolean input is **promoted**, not rejected: the
+          ``sqrt`` on the way in already produces a floating tensor, and the
+          output is allocated from that rather than from the input, so the
+          result comes back at the ambient default floating dtype
 
     .. warning::
         The gradient at the zero rotation is ``nan``:
         ``axis_angle_to_quaternion(torch.zeros(3)).sum().backward()`` leaves
         ``[nan, nan, nan]`` in ``.grad``, from an unclamped ``sqrt(0)``.
         Tracked in `#3949 <https://github.com/kornia/kornia/issues/3949>`_.
-
-    .. warning::
-        An integer tensor returns an all-zero integer tensor instead of a
-        quaternion: ``axis_angle_to_quaternion(torch.tensor([1, 0, 0]))`` is
-        ``tensor([0, 0, 0, 0])`` of dtype ``int64``, against the ``float32``
-        answer ``[0.8776, 0.4794, 0., 0.]``, because the output buffer is
-        allocated with the input dtype. Tracked in
-        `#3948 <https://github.com/kornia/kornia/issues/3948>`_.
 
     Args:
         axis_angle: tensor with axis angle in radians.
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -1273,9 +1273,10 @@ def axis_angle_to_quaternion(axis_angle: torch.Tensor) -> torch.Tensor:
     k: torch.Tensor = torch.where(mask, k_pos, k_neg)
     w: torch.Tensor = torch.where(mask, torch.cos(half_theta), ones)
 
-    quaternion: torch.Tensor = torch.zeros(
-        size=(*axis_angle.shape[:-1], 4), dtype=axis_angle.dtype, device=axis_angle.device
-    )
+    # the output dtype comes from the computed values, not from the input: an integer input is
+    # promoted to floating point by the sqrt above, and writing that back into an integer buffer
+    # would truncate every component to zero.
+    quaternion: torch.Tensor = torch.zeros(size=(*axis_angle.shape[:-1], 4), dtype=w.dtype, device=axis_angle.device)
     quaternion[..., 1:2] = a0 * k
     quaternion[..., 2:3] = a1 * k
     quaternion[..., 3:4] = a2 * k
--- a/tests/geometry/test_conversions.py
+++ b/tests/geometry/test_conversions.py
@@ -673,9 +673,25 @@ class TestAngleAxisToQuaternion(BaseTester):
         self.assert_close(quaternion, expected, atol=atol, rtol=rtol)
 
     @pytest.mark.parametrize("input_dtype", (torch.int16, torch.int32, torch.int64, torch.uint8))
-    def test_integer_input(self, input_dtype, device):
-        # an integer input used to be written back into an integer output buffer, which truncated
-        # every component to zero. see https://github.com/kornia/kornia/issues/3948
+    def test_convention_integer_input_is_promoted_to_float_3948(self, input_dtype, device):
+        # Convention, and the answer kornia#3948 settled: an integer axis-angle is PROMOTED, not
+        # rejected. The output buffer used to be allocated with dtype=axis_angle.dtype, so an
+        # integer input got an integer buffer and every component was truncated on the way in --
+        # the function returned tensor([0, 0, 0, 0]), not merely a wrong quaternion but a zero-norm
+        # one, with no error and no warning. The buffer now takes its dtype from the computed
+        # values (sqrt already promotes them), so the result is the float quaternion below.
+        # The wart this replaces said the intended behavior was undecided -- promote, or raise a
+        # TypeError the way a dtype guard would. Promotion is the choice; the sibling wart
+        # TestNormalTransformPixel.test_wart_integer_dtype_truncates_the_scale_to_zero_3959 pins
+        # the same family of defect and is still open, so this pin is also the precedent it should
+        # follow.
+        # Four cells because the promotion happens through torch's own type rules rather than an
+        # explicit .float(): a signed/unsigned or narrow/wide difference would show up here.
+        # The dtype fixture is dropped because the claim is about the input dtype itself.
+        # Snippet used to generate expected (torch + stdlib, executed on cpu):
+        #   axis_angle_to_quaternion(torch.tensor([1., 0., 0.])) ->
+        #     [0.8775825500488281, 0.4794255495071411, 0.0, 0.0]      (float32 reference)
+        #   math.cos(0.5), math.sin(0.5) -> (0.8775825618903728, 0.479425538604203)
         axis_angle = torch.tensor((1, 0, 0), device=device, dtype=input_dtype)
         quaternion = kornia.geometry.conversions.axis_angle_to_quaternion(axis_angle)
         assert quaternion.is_floating_point()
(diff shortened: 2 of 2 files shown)


ducha-aiki:
Reviewed by execution. **The fix is right and the reasoning in your comment is the correct diagnosis** — the output dtype has to come from the computed values, and `w.dtype` is the right source because the `sqrt` above has already promoted an integer input to floating point.

Verified on your head, repo `.venv` (torch 2.12.0): `tests/geometry/test_conversions.py --dtype=float32,float64` → **421 passed, 7 xfailed**. Your two new tests cover both directions (integer promotes, float preserves), and parametrising the integer case over `int16/int32/int64/uint8` is the right call — `uint8` is the one that would have been easy to miss.

**One required change, and it isn't visible from where you're standing.**

Your base is `8454a81a` (batch 3a). Two more batches have landed since: `#3967` (3b) and `#3984` (3c). Batch 3b added a pin to `tests/geometry/test_conversions.py` that **documents the bug you are fixing**:

```
TestAngleAxisToQuaternion::test_wart_integer_input_returns_the_zero_quaternion_3948
```

I applied your `kornia/` change on top of current `main` and ran the suite. Result:

```
FAILED tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_wart_integer_input_returns_the_zero_quaternion_3948
AssertionError: kornia#3948: integer input no longer keeps its integer outp...
1 failed, 604 passed, 31 xfailed
```

That failure is **the system working as intended**, not a problem with your fix. These `test_wart_*` pins exist to fail loudly the moment the wart is repaired, so a fix can never land silently. So when you rebase, please also delete `test_wart_integer_input_returns_the_zero_quaternion_3948` in the same commit, with a line in the PR body saying you removed it because #3948 is fixed. That keeps the pin's whole purpose intact: it dies exactly once, in the PR that earns it.

Please rebase onto `main` and re-run — that will be the first time your change meets the 3b/3c pins at all.


---

ducha-aiki:
Thanks — taking the buffer dtype from the computed values rather than from the input is the right fix, and the comment you put above it says exactly why.

I rebased the branch onto current `main` and pushed one commit on top of yours (`ab4e52a6`). The branch was based on `8454a81a`, before `#3967` landed the `#3948` pin, which was written against the integer output buffer and so went red as soon as your fix met it:

- `test_wart_integer_input_returns_the_zero_quaternion_3948` asserted `axis_angle_to_quaternion(tensor([1, 0, 0]))` comes back `int64` `[0, 0, 0, 0]`. It doesn't any more — dropped.
- Your `test_integer_input` and `test_float_input_keeps_dtype` are renamed into that class's convention-pin shape (`test_convention_integer_input_is_promoted_to_float_3948`, `test_convention_float_input_keeps_its_dtype_3948`) and carry the wart's context forward.

The context worth keeping is this. The wart said the intended behavior was genuinely undecided — a fix could promote to float, or it could raise a `TypeError` the way a dtype guard would — and it deliberately had no companion `xfail` because an assertion-shaped xfail can only express the first. **Your PR settles that question: promotion.** So I wrote that into the pin, and pointed the still-open `#3959` wart (`TestNormalTransformPixel.test_wart_integer_dtype_truncates_the_scale_to_zero_3959`, the same defect in `normal_transform_pixel`) at `#3948` as precedent rather than as a fellow undecided case.

Your `test_float_input_keeps_dtype` earns its place for a reason worth stating out loud, so I put it in the comment: a fix written as `.float()` or as an explicit `torch.float32` buffer would pass every integer cell and silently upcast `float16`/`bfloat16`. That's the cell that catches it.

Verified: `tests/geometry/test_conversions.py` green at float32/float64; all eight `#3948` cells green at float16, bfloat16, float32 and float64; no new half-precision failures against `main` (the half-precision suite has ~29 pre-existing failures on `main`, and the two apparent deltas turned out to be unseeded-random flakes that pass in isolation on both trees). ruff clean.

ducha-aiki (commented):
Re-checked against current `main` (`35555b55`). The fix is correct and the pin flip is done, but this PR has the same gap as `#3980`: the docstring it falsifies is still in the tree. Also it now conflicts, though only trivially.

## The fix works

```text
axis_angle_to_quaternion(torch.tensor([1, 0, 0]))

main:     tensor([0, 0, 0, 0])                              dtype=torch.int64
this PR:  tensor([0.8776, 0.4794, 0.0000, 0.0000])          dtype=torch.float32
```

Allocating the buffer from `w.dtype` rather than `axis_angle.dtype` is the right lever — the `sqrt` above has already promoted, so the computed values carry the correct dtype and the buffer just has to stop truncating them. The comment you added explaining that is worth keeping.

After rebasing onto current `main`, `tests/geometry` gives **0 failures**.

## Missing: the docstring this fix falsifies

`grep -rnE "#3948|issues/3948" kornia/` still returns one hit, `kornia/geometry/conversions.py:1229-1235`, inside `axis_angle_to_quaternion`'s own docstring:

> An integer tensor returns an all-zero integer tensor instead of a quaternion: ``axis_angle_to_quaternion(torch.tensor([1, 0, 0]))`` is ``tensor([0, 0, 0, 0])`` of dtype ``int64``, against the ``float32`` answer ``[0.8776, 0.4794, 0., 0.]``, because the output buffer is allocated with the input dtype. Tracked in `#3948`.

Every clause of that is now false — including the mechanism in the last line, which your change removes. I executed both sides to be sure, and the numbers above are exactly the warning's own. The block goes, and `grep -rnE "#3948|issues/3948" kornia/` has to come back empty before this merges; `AGENTS.md` makes that the gate.

As with `#3980`, this is a maintainer-side gap: the `#3948` pin flip landed as a commit on top of your work and should have carried the docstring deletion with it. Nothing here is something you missed.

## Trivial conflict

The branch went `CONFLICTING` after `#4006`/`#4028`. The conflict is a **single comment line**:

```text
HEAD:  # the other were chosen. Same shape as the #3948 wart pins in this file.
ours:  # the other were chosen. Same shape as the #3959 and #3957 wart pins in this file.
```

Take the second — the `#3948` wart is exactly what this PR retires, so the sentence has to point at the wart pins that are still live, and `#3959`/`#3957` both still have pins in the file. I resolved it that way, and every number above was measured on that rebased tree.

Not requesting changes — there is nothing for you to do. I will push the rebase plus the docstring deletion unless you would rather do it; say the word.

---

ducha-aiki (approved):
Rebased onto current `main` and pushed the maintainer commit — head is now `ec5c2dfe`. Nothing needed from you.

Two things went in:

1. **Rebase.** The only conflict was the single comment line I quoted above; resolved to `#3959`/`#3957`, both of which still have live wart pins in the file (6 references, checked).
2. **`docs(geometry): retire the kornia#3948 integer-input warning`.** The warning claimed `axis_angle_to_quaternion(torch.tensor([1, 0, 0]))` returns `tensor([0, 0, 0, 0])` of dtype `int64` "because the output buffer is allocated with the input dtype" — the exact mechanism your change removes. Replaced with a Convention bullet stating the new contract: integral and boolean input is **promoted, not rejected**, and the result comes back at the ambient default floating dtype. I verified that for `int64`, `int32`, `uint8` and `bool`, and under `torch.set_default_dtype(torch.float64)`, where it correctly follows to float64.

`grep -rnE "#3948|issues/3948" kornia/` is now empty. On the pushed head: `tests/geometry` 2394 passed / 0 failed, doctests 34 passed, `ruff check` and `ruff format --check` clean.

Approving. The `w.dtype` lever was the right one — the promotion has already happened by the time the buffer is allocated, so the buffer just had to stop undoing it.

---

ducha-aiki (approved):
Final verification round at `ec5c2dfe` (already on current `main`, nothing to rebase).

- `grep -rnE "#3948|issues/3948" kornia/ docs/` is empty; the remaining hits are the `tests/` pins and the `#3959` cross-reference, which correctly names the precedent rather than claiming the decision.
- `tests/geometry/test_conversions.py --dtype=float32,float64`: 609 passed, 22 xfailed. Doctests pass; ruff clean.
- I checked the new Convention bullet end to end rather than only the parametrized cells. `int16`/`int32`/`int64`/`uint8` **and** `bool` all return `[0.8776, 0.4794, 0., 0.]`, so "integral or boolean" holds. And it really is the *ambient* default, not a hardcoded float32: under `torch.set_default_dtype(torch.float64)` an `int64` input comes back `float64`. The wording is exact.

Approving. Conflicts with #3980 — see the note there on merge order.

Issue #3958, opened 2026-08-10
normalize_homography builds its normalization constants in float32: a float64 caller silently loses 8 significant digits (8.94e-09)


ducha-aiki:
`normalize_homography`, `denormalize_homography` and `normalize_homography3d` build their normalization matrices by calling `normal_transform_pixel*` **without passing `dtype=` or `device=` through**, then casting the float32 result to the input:

```python
# kornia/geometry/conversions.py:1286 and :1291 (normalize_homography)
src_norm_trans_src_pix: torch.Tensor = normal_transform_pixel(src_h, src_w).to(dst_pix_trans_src_pix)
...
dst_norm_trans_dst_pix: torch.Tensor = normal_transform_pixel(dst_h, dst_w).to(dst_pix_trans_src_pix)
```

`normal_transform_pixel` defaults to float32, so the constants `2/(size − 1)` are rounded to float32 **before** the cast to float64. A float64 caller therefore gets roughly 8 significant digits of normalization constant, not 16 — the returned tensor has `dtype=torch.float64` while carrying float32-quality numbers.

## Repro

```python
import torch
from kornia.geometry.conversions import (
    denormalize_homography,
    normal_transform_pixel,
    normal_transform_pixel3d,
    normalize_homography,
    normalize_homography3d,
)

f64 = torch.float64
got = normalize_homography(torch.eye(3, dtype=f64)[None], (4, 4), (6, 6))[0]
Nd = normal_transform_pixel(6, 6, dtype=f64)[0]
Ns = normal_transform_pixel(4, 4, dtype=f64)[0]
exact = Nd @ torch.linalg.inv(Ns)

print("input dtype float64, output dtype:", got.dtype)
print("normalize_homography [0,0] =", got[0, 0].item())
print("float64-native       [0,0] =", exact[0, 0].item())
print("maxdiff                    =", (got - exact).abs().max().item())

print("\n# the leak at the source: the helper defaults to float32 whatever the caller uses")
print("normal_transform_pixel(4,5)            [1,1] =", normal_transform_pixel(4, 5)[0, 1, 1].item())
print("normal_transform_pixel(4,5,dtype=f64)  [1,1] =", normal_transform_pixel(4, 5, dtype=f64)[0, 1, 1].item())
print("exact 2/3                                    =", 2 / 3)

print("\n# same for the siblings")
g2 = denormalize_homography(torch.eye(3, dtype=f64)[None], (4, 4), (6, 6))[0]
[...]

PR #3979, opened 2026-08-15
fix(geometry): build homography normalization matrices in the caller's dtype

NoiceHax:

**Fixes/Relates to:** #3958

## What does this PR do?

`normalize_homography`, `denormalize_homography` and `normalize_homography3d` called `normal_transform_pixel` / `normal_transform_pixel3d` with no `dtype=` or `device=`, then cast the result onto the input. The helpers default to float32, so the `2 / (size - 1)` constants were rounded to float32 before the cast and a float64 caller got a float64 tensor holding float32 quality numbers. `normalize_homography(eye(3, f64), (4, 4), (6, 6))[0, 0]` returned 0.5999999910593036 instead of 0.6000000000000001, off by 8.94e-09 against a float64 eps of 2.2e-16. This passes the caller's dtype and device through at the six call sites and drops the cast.

I read this as a repair rather than a change of convention. float32 output is bit identical, since the cast was a no-op there. float16 could in principle move by one ULP because the value is no longer rounded twice through float32, but a sweep over every 2D size pair from 1 to 11 and every 3D size triple from 1 to 5 found no changed bit in either dtype. Only float64 moves, from 1e-08 error to 1e-16.

Scope is kept to the pass-through. The `2 / (size - 1)` formula and `align_corners` are untouched, per the note in the issue that they belong to #3904 / #3945.

New tests in `TestHomographyNormalizePrecision` compare each of the three ops against its documented contract computed natively in float64 with the same public helper. All three fail on main at 8.94e-09, 2.48e-08 and 6.39e-09 with `atol=1e-15`, and pass here. A fourth test checks dtype and device are preserved.

## Test log

`pixi run test` shells out to `uv run pytest` against the repo `.venv`; on this Windows box I ran that invocation directly.

Before the fix (source change reverted, new tests kept):

```
$ .venv/Scripts/python.exe -m pytest tests/geometry/transform/test_homography_warper.py -k TestHomographyNormalizePrecision --dtype=float32,float64
E       AssertionError: Tensor-likes are not close!
E       Mismatched elements: 2 / 16 (12.5%)
E       Greatest absolute difference: 6.386211781617135e-09 at index (0, 1, 1) (up to 1e-15 allowed)
E       Greatest relative difference: 1.4901160823773315e-08 at index (0, 1, 1) (up to 0.0 allowed)

FAILED tests/geometry/transform/test_homography_warper.py::TestHomographyNormalizePrecision::test_normalize_homography_float64[cpu]
FAILED tests/geometry/transform/test_homography_warper.py::TestHomographyNormalizePrecision::test_denormalize_homography_float64[cpu]
FAILED tests/geometry/transform/test_homography_warper.py::TestHomographyNormalizePrecision::test_normalize_homography3d_float64[cpu]
================= 3 failed, 2 passed, 193 deselected in 4.03s =================
```

After the fix:

```
$ .venv/Scripts/python.exe -m pytest tests/geometry/transform/test_homography_warper.py -v -k TestHomographyNormalizePrecision --dtype=float32,float64
tests/geometry/transform/test_homography_warper.py::TestHomographyNormalizePrecision::test_normalize_homography_float64[cpu] PASSED   [ 20%]
tests/geometry/transform/test_homography_warper.py::TestHomographyNormalizePrecision::test_denormalize_homography_float64[cpu] PASSED [ 40%]
tests/geometry/transform/test_homography_warper.py::TestHomographyNormalizePrecision::test_normalize_homography3d_float64[cpu] PASSED [ 60%]
tests/geometry/transform/test_homography_warper.py::TestHomographyNormalizePrecision::test_dtype_device_preserved[cpu-float32] PASSED [ 80%]
tests/geometry/transform/test_homography_warper.py::TestHomographyNormalizePrecision::test_dtype_device_preserved[cpu-float64] PASSED [100%]
====================== 5 passed, 193 deselected in 3.04s ======================

$ .venv/Scripts/python.exe -m pytest tests/geometry/transform/test_homography_warper.py --dtype=float32,float64 -q
============================ 198 passed in 14.67s =============================

$ .venv/Scripts/python.exe -m pytest tests/geometry/transform tests/geometry/test_conversions.py tests/geometry/test_grid.py --dtype=float32,float64 -q
========== 1261 passed, 28 skipped, 9 xfailed, 48 warnings in 18.70s ==========

$ .venv/Scripts/python.exe -m pytest --doctest-modules kornia/geometry/conversions.py kornia/geometry/transform/ -q
============================= 76 passed in 10.38s =============================

$ ruff check kornia/geometry/conversions.py tests/geometry/transform/test_homography_warper.py
All checks passed!
$ ruff format --check kornia/geometry/conversions.py tests/geometry/transform/test_homography_warper.py
2 files already formatted
```

Byte identity check for float32 and float16, comparing the raw bits of the output before and after the change over every 2D size pair 1..11 and every 3D size triple 1..5 (89814 rows): identical.

CPU only, torch 2.13.0+cpu, no CUDA build on this host, so I did not verify the CUDA path.

- [ ] ðŸŸ¢ **Human-written**
- [x] ðŸŸ¡ **AI-assisted** â€” AI helped; I reviewed the resulting change and ran the relevant tests
- [ ] ðŸ”´ **AI-generated** â€” an agent wrote most of it; I verified the result and can address reviewer questions


ducha-aiki (changes requested):
The library change is right, and it is better than the PR body claims: I measured it against an analytic float64 oracle built from the documented formula rather than from kornia itself, and it makes `normalize_homography` **exact** in float64 while leaving every other dtype bitwise untouched.

But the PR cannot merge as it stands. Your branch is based on `8454a81a` (batch 3a), so it has never run the pins that landed in `#3967`, `#3984`, `#4006` and `#4028`. Those pins record the current behavior as contract, and your fix — correctly — breaks three of them. GitHub reports the PR mergeable, which is exactly what makes this dangerous: it merges clean and leaves falsified documentation behind.

## The fix itself: verified, and better than stated

Oracle: `N(h, w) = [[2/(w-1), 0, -1], [0, 2/(h-1), -1], [0, 0, 1]]` built from Python floats in float64, `expected = N(dst) @ inv(N(src))`, sizes `(4, 5) -> (8, 9)`. No kornia in the oracle.

| input dtype | `main` | this PR |
|---|---|---|
| float64 | 6.386212e-09 | **0.0 — exact** |
| float32 | 3.405980e-08 | 3.405980e-08 |
| bfloat16 | 1.116071e-03 | 1.116071e-03 |
| float16 | 1.395089e-04 | 1.395089e-04 |

The float32 result is not merely close, it is **bitwise identical** — I compared the raw words via `.view(torch.int32)` and got the same nine integers on both trees. So the common path cannot move, and the half dtypes are unchanged too. That is the strongest form this fix could take, and it is worth putting in the PR body: "recovers 8 significant digits for float64 callers, bitwise no-op for every other dtype" is a much better claim than the one you made.

Your three float64 tests are genuine pins — all three fail on unpatched `main` and pass with the change. `test_dtype_device_preserved` passes on `main` as well, so it guards nothing new; harmless to keep, just do not count it as evidence.

## Required before merge: three pins flip

Applying your `kornia/` diff to current `main` (`35555b55`) and running `tests/geometry` gives exactly three failures, against zero on `main`:

**1. `TestNormalizeHomography::test_convention_float64_input_gets_float64_normalization_constants_3958` — `[XPASS(strict)]`.** This is the machinery working. The convention test asserts the behavior you are implementing and is marked `xfail(strict=True)` so that fixing `#3958` forces the marker out. **Drop the `xfail` marker and keep the test** — it becomes the permanent regression guard for your change. Do not delete it.

**2. `TestNormalizeHomography::test_wart_float32_constants_leak_into_float64_results_3958` — fails with its own message:**

```
AssertionError: kornia#3958: normalize_homography no longer rounds its constants to float32
Expected 0.5999999910593036 but got 0.6000000000000001.
```

That pin exists to record the defect while it is live. **Delete it.**

**3. `TestNormalizeHomography::test_convention_dyadic_sizes_are_not_necessary_for_a_bitwise_round_trip` — this one you could not have known about, and it is the reason this needs a second pass rather than a marker edit.**

```
AssertionError: the float64 identity no longer survives normalize(denormalize(.)) bitwise at the
non-dyadic (4, 5) -> (8, 9), so denormalize_homography's 'not necessary' half has lost its counterexample
  reverse[0][1][2] == -2.2204e-16, expected 0.0
```

The forward leg still round-trips bitwise; the reverse leg no longer does. Note the name carries no issue number, so no `#3958` grep finds it. My reading, which you should check rather than take: the bitwise round trip on `main` is an **artifact of the defect**. float32-rounded constants are dyadic when widened to float64, so the products and the inverse are exact; genuine float64 constants make `2/3` and `2/7` 53-bit approximations and one ulp appears. If that is right, the property was never a contract — it was a coincidence the pin captured — and the honest resolution is to re-derive the test against the corrected behavior and say so, not to loosen it to `assert_close`. The pin's own comment says it is the counterexample supporting a "not necessary" claim in `denormalize_homography`'s Convention block, so that prose needs revisiting in the same change.

## Required before merge: the docstrings your fix falsifies

`grep -rnE "#3958|issues/3958" kornia/ tests/` on current `main`:

- **`kornia/geometry/conversions.py:1885-1908`** — the whole `.. warning::` on `normalize_homography`. It states the float32 rounding as live behavior and carries three executed figures: `0.5999999910593036` (which your fix turns into `0.6000000000000001` — I reproduced both), `2.483526828633842e-08` for `denormalize_homography`, and a clause naming `normalize_homography3d`. All three functions are fixed by this PR, so the entire block goes.
- **`kornia/geometry/conversions.py:2276`** — `denormalize_homography`'s Convention block says "That function's dtype-pass-through (`#3958`) ... warnings apply here too". The pointer has to go with the target.
- **`kornia/geometry/conversions.py:1945`** — "the identity to within a single `float32` rounding step". Not `#3958`-tagged, so it will not show up in the grep; please re-execute it, since the constants it describes have changed.
- **`tests/geometry/test_conversions.py`** — ten `#3958` references at lines 559, 3573, 3599, 3977, 3978, 3997, 4000, 4001, 4352, 4369. Most are comments explaining that neighbouring pins are *independent* of the leak, which stays true; lines 4352 and 4369 are the xfail reason and marker from item 1. Re-check each in this PR rather than after — `AGENTS.md` makes this the merge gate for a change that closes a documented issue.

A surviving `#3958` hit under `kornia/` after this merges means the change is incomplete.

## Minor: two of your test's stated size choices do not do what the docstring says

Your class docstring says sizes are picked "so that `2 / (size - 1)` is not exactly representable in binary (2/3, 2/5, 2/7)". True for the 2D tests — `(4, 4) -> (6, 6)` gives `2/3` and `2/5`. But `test_normalize_homography3d_float64` uses `(2, 4, 5) -> (3, 8, 9)`, and four of those six axes are exactly representable: depth 2 gives `2/1 = 2.0`, depth 3 gives `2/2 = 1.0`, width 5 gives `2/4 = 0.5`, width 9 gives `2/8 = 0.25`. Only the two height axes (`2/3`, `2/7`) actually exercise the rounding.

The test still fails on `main` and passes here, so it is a real pin and the two live axes carry it — this is a wording-versus-reality mismatch, not a broken test. But a size whose `size - 1` is a power of two pins nothing about float32 rounding, and someone will later "simplify" those numbers on the strength of the docstring. Either change the depths and widths to non-dyadic values or drop the claim to match what the sizes actually do.

## Rebase

Also please rebase onto current `main`. Your CI last ran on 2026-08-15 against `8454a81a`; five merges have touched `conversions.py` since, and green there says nothing about the pins. `maintainer_can_modify` is on, so say the word if you would rather I push the rebase and you take the pin and docstring edits from there.

Requesting changes on the pins and docstrings only. The `kornia/` change is correct and I want it in.

---

ducha-aiki:
**Changes requested — the rebase and two of the three pin edits are right, but the `kornia/` docstring sentence this fix falsifies is still there, which is the exact merge gate `AGENTS.md` sets, and the round-trip pin was loosened rather than re-derived.**

Measured on head `a27db583`, merge-base `4ab79c78`, torch 2.9.1 / Python 3.11.14 / macOS arm64, `--device=cpu` and `--device=mps`. No CUDA on this host; no CUDA behavior is claimed below.

## Classification: ordinary bugfix, lands now — not a repair-window item

The issue flagged the classification rather than deciding it. Deciding it with measurement: I dumped every output of all three functions over every 2-D size pair `1..11` (both `dsize_src` and `dsize_dst`) and every 3-D size triple `1..5`, on both trees, and compared the **raw integer words**:

```
2d torch.float16    bitwise_identical=True  rows_changed=     0/29282 maxabsdiff=0.000000e+00
3d torch.float16    bitwise_identical=True  rows_changed=     0/15625 maxabsdiff=0.000000e+00
2d torch.bfloat16   bitwise_identical=True  rows_changed=     0/29282 maxabsdiff=0.000000e+00
3d torch.bfloat16   bitwise_identical=True  rows_changed=     0/15625 maxabsdiff=0.000000e+00
2d torch.float32    bitwise_identical=True  rows_changed=     0/29282 maxabsdiff=0.000000e+00
3d torch.float32    bitwise_identical=True  rows_changed=     0/15625 maxabsdiff=0.000000e+00
2d torch.float64    bitwise_identical=False rows_changed= 26902/29282 maxabsdiff=3.129244e-07
3d torch.float64    bitwise_identical=False rows_changed= 11349/15625 maxabsdiff=8.940696e-08
```

And against an **exact rational oracle** (`fractions.Fraction`, cofactor inverse, no float and no kornia anywhere in the oracle), over 10 000 size pairs `2..11`:

```
### pr-3979
  normalize_homography   worst abs err = 1.776357e-15   mean = 1.211781e-16
  denormalize_homography worst abs err = 1.776357e-15   mean = 7.926011e-17
### base-3979
  normalize_homography   worst abs err = 3.129244e-07   mean = 4.381299e-08
  denormalize_homography worst abs err = 3.129244e-07   mean = 4.381299e-08
```

Eight orders of magnitude, every float64 size pair, and **not one bit moves in any other dtype**. A `float64` tensor carrying `float32` constants is silently-wrong output, and no dtype that was correct changes. That is a clearly-broken-output bugfix under the repair-window policy: it does not need to wait for a coordinated window.

## What is already right

- **The rebase happened and is clean.** Head sits on `4ab79c78`; all 35 checks are green, including the `float64` legs on ubuntu/windows × 3.11–3.13 × torch 2.5.1/2.9.1 and `dynamo`. That is what the last review asked for and the pins now actually ran.
- **The `xfail` came out and the test stayed.** `test_convention_float64_input_gets_float64_normalization_constants` is now a plain regression guard. I checked out the PR's two test files onto the merge-base worktree and ran them there: it is the **only** thing that fails, which is exactly what a guard should do.
  ```
  FAILED tests/geometry/test_conversions.py::TestNormalizeHomography::test_convention_float64_input_gets_float64_normalization_constants[cpu]
    - AssertionError: normalize_homography did not use float64 normalization constants
  ========== 1 failed, 59 passed, 573 deselected, 12 warnings in 0.96s ===========
  ```
  and the three `TestHomographyNormalizePrecision` float64 tests fail there too (`6.386211781617135e-09 ... up to 1e-15 allowed`), while all five pass on head.
- **The wart pin is gone**, and `_ambient_default_dtype` is still used by the `normal_transform_pixel` pin, so nothing was orphaned.
- **The `#3958` merge gate passes on the `kornia/` side.** `grep -rnE "#3958|issues/3958" kornia/` on head returns **nothing**.
- **`conversions.py:1897` did not need re-executing after all — I re-executed it and it is exactly right.** The last review asked you to; you did not touch it, and that turns out to be correct, so no action:
  ```
  float32 equal-size identity deviations: [(2, 0.0), (3, 0.0), (4, 5.960464477539063e-08), (5, 0.0),
    (6, 5.96...e-08), (7, 5.96...e-08), (8, 0.0), (9, 0.0), (10, 0.0), (11, 5.96...e-08), (12, 0.0)]
  4x4 case: 5.960464477539063e-08
  ```
  Same on both trees. The `(4, 7)` figures at `conversions.py:2263-2266` also re-execute unchanged (`normalize dev=5.960464477539063e-08`, `denormalize dev=0.0`).
- **Nothing else regressed.** `tests/geometry` full: base `4362 passed, 18 xfailed, 12 xpassed` → head `4367 passed, 17 xfailed, 12 xpassed` (+5 new, one xfail→pass, one wart deleted; the 12 xpasses are pre-existing on both). CPU `float16,bfloat16` on the two touched files: **63 failed on both trees** — pre-existing, not yours. Doctests `78 passed`. `ruff check` / `ruff format --check` clean.
- **Gradients, compile, and MPS all hold.** `gradcheck` True for all three in float64 on both trees; `torch.compile(fullgraph=True)` compiles all three in float32 and float64 and matches eager bitwise; `torch.jit.script` still compiles all three. On **MPS** (I did run this) the float16/bfloat16/float32 bit-hashes over `1..9` size pairs are **byte-identical between the two trees**, dtype and device pass through, and the `#3959` MPS behavior (`nh int64 -> float32 all_nan=True`, `dh int64 -> RuntimeError internal assert`) is unchanged — so that warning survives your change intact.
- Complex inputs pass through in both trees; `complex128` improves the same way float64 does (`0.5999999910593035+0j` → `0.6000000000000001+0j`).

## Blockers

### 1. `kornia/geometry/conversions.py:2252-2253` — the sentence this PR falsifies is still in the rendered docs

This is item 3 of the last review. You correctly diagnosed the cause in your commit message, updated the test comment to say so — and then left the docstring the test exists to support untouched.

`kornia/geometry/conversions.py:2249-2256`, in `denormalize_homography`'s Convention block:

```
2249	 computation, not of ``H``'s entries or of the sizes alone, so dyadic
2250	 sizes on their own are **neither necessary nor sufficient**: an ``H``
2251	 with non-dyadic entries at ``2 ** k + 1`` sizes misses in both legs
2252	 in ``float32``, while the ``float64`` identity comes back bitwise
2253	 through both legs at the non-dyadic ``(4, 5) -> (8, 9)``. What does
```

Executed on head:

```
2253 forward denorm(norm(I)) bitwise: True   maxabs: 0.0
2253 reverse norm(denorm(I)) bitwise: False  maxabs: 2.220446049250313e-16
```

on the merge-base:

```
2253 forward denorm(norm(I)) bitwise: True   maxabs: 0.0
2253 reverse norm(denorm(I)) bitwise: True   maxabs: 0.0
```

"through both legs" is now false, and this is `kornia/` prose that renders into the API reference. Per `AGENTS.md`, a surviving falsified statement under `kornia/` means the change is incomplete — the `#3958` grep does not catch it because the sentence carries no issue number, which is precisely the case `AGENTS.md` warns about.

The **claim** survives; only the example needs narrowing. The forward leg is still bitwise at the non-dyadic `(4, 5) -> (8, 9)`, so "not necessary" keeps a counterexample — say "comes back bitwise through the `denormalize(normalize(.))` leg", and (worth a clause) that the reverse leg lands one ulp out because `2/3` and `2/7` are 53-bit approximations. I re-executed the neighbouring "what does hold" sentence at 2254-2256 too — a dyadic-entried `H` at `2**k + 1` sizes is still bitwise through both legs in float32 and float64 — so that half needs no edit.

### 2. `tests/geometry/test_conversions.py:4297-4318` — the pin was loosened, which is what the last review asked you not to do

> the honest resolution is to re-derive the test against the corrected behavior and say so, not to loosen it to `assert_close`.

`tests/geometry/test_conversions.py:4318` is now:

```python
self.assert_close(reverse, identity, atol=1e-15, rtol=0.0)
```

Three consequences, measured:

- **It no longer discriminates.** `atol=1e-15` is ≈4.5 ulp of 1.0; the real residual is `2.220446049250313e-16`, exactly 1 ulp, and the merge-base's `0.0` also passes. In the merge-base run quoted above this test is among the **59 passed** — it now says nothing about either tree. The class's own comment at `4036-4042` sets `atol=rtol=0` exactness as the invariant for these pins; this is the one place it could have been kept and was not.
- **The name is now false.** `test_convention_dyadic_sizes_are_not_necessary_for_a_bitwise_round_trip`, with half the body no longer checking bitwise.
- **The generation snippet was deleted** (`4301-4304` on the merge-base). Its sibling `test_convention_dyadic_sizes_are_not_sufficient_for_a_bitwise_round_trip` at `4273-4276` still carries one; this file's convention is that every executed literal names how it was produced.

Re-derive it: keep `torch.equal` on the forward leg, and on the reverse leg pin the residual you actually measured — `2.220446049250313e-16` at `[0][1][2]`, or `assert not torch.equal(...)` plus a 1-ulp bound — with a snippet, and rename to match (`..._are_not_necessary_for_a_bitwise_forward_round_trip`, or similar). A pin that passes on both trees is not a pin.

### 3. Three test comments still describe the deleted mechanism as live

The last review asked for each `#3958` hit in `tests/` to be re-checked in this change. Two were updated; three were not, and all three are now factually wrong.

`tests/geometry/test_conversions.py:3630-3633`:
```
3630:         # unconditionally: changing the process default changes the result. That is the mechanism
3631:         # behind the float32 constants that leak into float64 homography pipelines (kornia#3958,
3632:         # pinned in TestNormalizeHomography): normalize_homography calls these helpers without
3633:         # passing dtype= through, so they materialise at the ambient default and are cast after.
```
`normalize_homography` now passes `dtype=` through — that is this PR. The *helper's* ambient-default behavior (what the pin actually asserts) is unchanged and correct; only the "and here is what it leaks into" clause has to become past tense. The assertion message at `3656-3658` ("the kornia#3958 mechanism pinned in TestNormalizeHomography has changed") points at a pin that no longer exists.

`tests/geometry/test_conversions.py:4038-4042`:
```
4038:     # pins compare at atol=rtol=0. That also keeps them independent of kornia#3958 (the float32
4039:     # constants leak, pinned separately below with non-dyadic sizes): a fix for #3958 must not flip
4040:     # an ordering or direction pin. The exactness invariant is theirs alone -- the bug pins below
4041:     # deliberately step outside it (the round-trip pin's non-dyadic (4, 5)/(8, 9) legs at
4042:     # atol=32*eps, the #3960 shape-guard cells), so a new atol=0 pin belongs here
```
"pinned separately below with non-dyadic sizes" and "the bug pins below" describe the wart test you deleted.

`tests/geometry/test_conversions.py:4424-4425` — this one you could reasonably have missed, it carries `#3959` not `#3958`:
```
4424:         # integer-input warning: the normalization matrices are cast to the input's int64 by
4425:         # .to(input), truncating their scales to zero (the truncation itself is pinned in
```
There is no `.to(input)` any more. The observable behavior is unchanged — I measured `int32` and `int64` on both trees and they raise identically, so the pin still passes and the `#3959` docstring warning at `1861-1884` stays true — but the mechanism sentence inside the comment now describes code you removed. One line.

Also reflow `4056-4062`: `4058` was left as a short orphan line by the edit.

## Non-blocking

- **The 3-D size claim from the last review is still unaddressed.** `tests/geometry/transform/test_homography_warper.py:426-429` says sizes are chosen so `2 / (size - 1)` is not exactly representable, but `test_normalize_homography3d_float64` uses `(2, 4, 5) -> (3, 8, 9)`, and four of those six axes are dyadic: `2/1 = 0x1.0p+1`, `2/2 = 0x1.0p+0`, `2/4 = 0x1.0p-1`, `2/8 = 0x1.0p-2`. Only heights 4 and 8 (`2/3`, `2/7`) carry the pin. The test is real — it fails on the merge-base — but the docstring invites someone to "simplify" the live axes away.
- **The oracle is kornia's own helper.** `expected = dst_norm @ _torch_inverse_cast(src_norm)` with `normal_transform_pixel(..., dtype=f64)` makes these self-consistency checks: they would pass unchanged if the `2/(size - 1)` formula itself were wrong. `AGENTS.md` prefers a hardcoded literal plus the generating snippet. They do discriminate the defect, so this is a strength note, not a defect.
- **`TestHomographyNormalizePrecision::test_normalize_homography_float64` and `TestNormalizeHomography::test_convention_float64_input_gets_float64_normalization_constants` now pin the same claim in two files.** Worth one cross-reference comment so a later cleanup does not delete the one that is load-bearing.
- **New, undocumented: eager and `torch.jit.trace` now disagree in float64 for `denormalize_homography`.** Under tracing `_torch_inverse_cast` takes the cofactor fallback; with float32-rounded (dyadic) constants that agreed bitwise with `linalg.inv`, and with genuine float64 constants it does not. Over 625 size pairs `2..6`:
  ```
  ##### base-3979
  denormalize_homography float64 trace-vs-eager: differing=0,  worst maxabs=0.000000e+00
  ##### pr-3979
  denormalize_homography float64 trace-vs-eager: differing=75, worst maxabs=8.881784e-16
  ```
  `8.9e-16` is float64 noise and both modes are eight orders closer to exact than the base, so this is not a defect — but it is a real new fact about a route the docstring already describes at `2238` and `2338`, and nothing pins it. `test_one_pixel_output_trace_crosses_singleton_source_boundary` compares traced vs eager at `atol=rtol=0` but runs in float32, so it is unaffected. A clause in that bullet would be cheap.
- **`uint8` fails differently now.** Base: `nh` → `RuntimeError: expected scalar type Byte but found Float`, `dh`/`nh3` → `LinAlgError` on a zero diagonal. Head: all three → `RuntimeError: value cannot be converted to type uint8 without overflow`, raised at construction because `-1.0` no longer arrives via a wrapping `.to()`. Strictly clearer, unpinned, and not covered by the `#3959` warning (which only claims `int64`) — so nothing is falsified. Mentioning it in case you want a cell in the `#3959` pin.
- **The PR body still understates the result.** "recovers ~8 significant digits for every float64 caller (worst error `3.13e-07` → `1.78e-15` against an exact rational oracle), bitwise no-op for float16/bfloat16/float32 across 89 814 size combinations on cpu and mps" is both true and much stronger than what is written now. The commit message's float16 sweep claim reproduces exactly.

None of the three blockers touches the `kornia/` fix itself, which is correct and which I want in. Blocker 1 is one sentence, blocker 3 is three comments, and blocker 2 is the one that needs actual thought.

---

ducha-aiki:
**Re-review of the delta only. The push since my last review is a rebase and nothing else — the net diff is byte-identical — so all three blockers stand exactly as written.**

Measured on head `bfb1cc79`, base `6fa75d64` (the branch is two commits on top of it), torch 2.9.1 / Python 3.11.14 / macOS arm64, `--device=cpu`. No CUDA on this host; no CUDA behavior is claimed.

## The push was a rebase

Previous head was `a27db583` on base `4ab79c78`; current head is `bfb1cc79` on base `6fa75d64`. Comparing each revision's own net patch over the three touched files, with `index` lines and hunk offsets normalized away:

```
$ git diff 4ab79c78 a27db583 -- <3 files> > old.patch
$ git diff 6fa75d64 bfb1cc79 -- <3 files> > new.patch
$ sed -E 's/^index .*//; s/^@@ [^@]*@@//' old.patch > o.n; sed ... new.patch > n.n
$ diff o.n n.n && echo IDENTICAL
IDENTICAL
```

347 lines both sides, same content. `#3989` landing on `main` in between is what moved the line numbers (`kornia/` by −12, `tests/` by −34); nothing in this PR changed.

## Still right, re-executed on the new base

Not asking you to re-do any of this — recording that the rebase did not disturb it.

- Against an exact rational oracle (`fractions.Fraction`, cofactor inverse, no float and no kornia in the oracle), worst abs error over 30 `(dsize_src, dsize_dst)` combinations drawn from `{(4,4),(6,8),(4,5),(3,3),(5,7),(9,11)} × {(6,6),(1,2),(8,9),(7,5),(11,3)}`: **head `8.88e-16` normalize / `4.44e-16` denormalize** vs **base `3.13e-07` / `1.57e-07`**. Eight orders.
- Bitwise sweep over every 2-D size pair `1..11` and 3-D triple `1..5`, raw integer words, 29 407 rows per dtype: `float16` 0 rows changed, `bfloat16` 0, `float32` 0, `float64` 26 959 changed, max `3.13e-07`. (Scope caveat below.)
- `tests/geometry` full: base `4363 passed, 17 xfailed, 12 xpassed` → head `4368 passed, 16 xfailed, 12 xpassed`. CPU `float16,bfloat16` on `test_conversions.py`: **45 failed on both trees**, pre-existing. `pixi run pre-commit-all` clean. `pixi run doctest`: 4 failures, all pre-existing on base (`filters/kernels.py` ×2, `liegroup/se2.py`, `feature/matching.py`), none under `geometry/`. `KORNIA_TEST_OPTIMIZER=inductor` on `test_homography_warper.py`: `222 passed`. CI 33/33 green.
- `grep -rnE "#3958|issues/3958" kornia/` → 0 hits. That gate passes; blocker 1 is the sentence the grep cannot see.

## The three blockers, at their new line numbers

Unchanged in substance from [the last review](https://github.com/kornia/kornia/pull/3979#issuecomment-5456035213) — re-cited so the line numbers match today's head.

**1. `kornia/geometry/conversions.py:2240-2241`** — `denormalize_homography`'s Convention block still says the `float64` identity "comes back bitwise **through both legs** at the non-dyadic `(4, 5) -> (8, 9)`". Re-executed on `bfb1cc79`:

```
forward denorm(norm(I)) bitwise: True   maxabs: 0.0
reverse norm(denorm(I)) bitwise: False  maxabs: 2.220446049250313e-16
```

on `6fa75d64` both legs are `True`. Falsified `kornia/` prose that renders into the API reference. The claim survives, the example needs narrowing to the `denormalize(normalize(.))` leg, ideally with the one-ulp reverse-leg clause. The neighbouring "what does hold" sentence at 2242-2244 re-executes unchanged and needs no edit.

**2. `tests/geometry/test_conversions.py:4263-4284`** — the round-trip pin is still loosened rather than re-derived. `4284` is `self.assert_close(reverse, identity, atol=1e-15, rtol=0.0)`; `1e-15` is ≈4.5 ulp of 1.0, the residual is exactly 1 ulp, and the base's `0.0` passes it too, so the cell now discriminates neither tree. The test name still says `bitwise`, and the generation snippet is still deleted while its sibling at `4239-4242` keeps one.

**3. Three test comments still describe deleted mechanisms** — `3597-3599` ("normalize_homography calls these helpers without passing `dtype=` through"), the assertion message at `3622-3624` (points at a pin that no longer exists), `4004-4008` ("pinned separately below with non-dyadic sizes" / "the bug pins below" — the deleted wart test), and `4390-4391` ("cast to the input's int64 by `.to(input)`"). Plus the orphan short line left at `4024`.

## One new non-blocking item — and a correction to the scope of my own last claim

I wrote "not one bit moves in any other dtype". That is true over the swept range `1..11` and **not** true in general. `2 / (size - 1)` sits on a `float16` double-rounding boundary at exactly three sizes below 100 000 — `8284`, `16567`, `37263` — and there the two orders differ by 1 ulp:

```
size 8284:  exact 2.41458408789086082e-04
            direct  (head) 2.41398811340332031e-04   |err| 5.960e-08
            via f32 (base) 2.41518020629882812e-04   |err| 5.961e-08   ulp 1.192e-07
```

The true value is a hair off the midpoint, so neither order is meaningfully more accurate — head's is the correctly-rounded one. It does reach the output: `normalize_homography(eye(3, f16), (16567, 3), (5, 5))` differs between the trees (`27660` vs `27659` as raw words), and `denormalize_homography(eye(3, f16), (5, 5), (37263, 7))` turns two entries that are exactly `0.0` on base into `1.2e-07` / `-2.4e-07` subnormal noise on head. `bfloat16` has no such size below 100 000.

The consequence worth a line somewhere: **eager and compiled now disagree in `float16` at those sizes**, where they agreed before.

```
normalize_homography(eye(3, f16), (16567, 3), (5, 5)), raw words
  head  eager    [14336, 0, -18432, 0, 27660, 27660, 0, 0, 15360]
        compiled [14336, 0, -18432, 0, 27659, 27659, 0, 0, 15360]   equal: False
  base  eager    [14336, 0, -18432, 0, 27659, 27659, 0, 0, 15360]
        compiled [14336, 0, -18432, 0, 27659, 27659, 0, 0, 15360]   equal: True
```

Cause is pre-existing and not yours: `normal_transform_pixel`'s capture branch deliberately computes in `float32` and casts (`conversions.py:2059-2067`, the "bfloat16 rounds 257 to 256" comment), while the eager branch builds at the requested dtype — so the two branches always rounded differently for half dtypes, and until this PR the homography path only ever took the double-rounded order in both modes. Not a blocker: 1 ulp, `float16`, image dimensions ≥ 8284. But it is a real new fact about a route the file documents, and if you would rather not carry it, saying so in the PR body is enough for me.

While you are in the PR body: the `float16`/`bfloat16` bitwise claim there wants the same range qualifier, and the float64 result is still understated (worst `3.13e-07` → `8.88e-16` against an exact rational oracle).

---

Blocker 1 is one sentence, blocker 3 is four comments and a reflow, blocker 2 is the one that needs thought. The library change itself is correct and I want it in.

NoiceHax:
Rebased onto `main` (`ca5021eb`) and addressed all three blockers.

**Blocker 1 — `denormalize_homography`'s Convention block**

You were right that the bitwise round trip on both legs was an artifact of the defect. Re-executed on the rebased head:

```
forward denormalize(normalize(I)) bitwise: True   max|res| 0.0
reverse normalize(denormalize(I)) bitwise: False  max|res| 2.220446049250313e-16
```

The bullet now claims the forward leg only, and says why the reverse one moved: genuine `float64` constants make `2 / (size - 1)` a 53-bit approximation at `(4, 5) -> (8, 9)`, where the float32-rounded constants were dyadic when widened and so inverted exactly. The "what does hold" sentence after it is unchanged, as you said it re-executes fine.

**Blocker 2 — the round-trip pin**

Re-derived rather than loosened. You were right that `atol=1e-15` discriminated neither tree — it is ~4.5 ulp against a 1 ulp residual, and the pre-fix `0.0` passes it too. The reverse leg is now pinned exactly:

```python
residual = reverse - identity
assert residual.abs().max().item() == 2.220446049250313e-16
```

which fails on the pre-fix tree, where the residual is `0.0`. The generation snippet is restored to match the sibling pin at the "not sufficient" cell.

One correction to your note, which I only caught by executing it: the residual is nonzero at **two** entries, `[0][1][1]` and `[0][1][2]`, not `[0][1][2]` alone. Both are one ulp. The comment and the snippet say so.

**Blocker 3 — comments describing removed mechanisms**

Four sites, all updated: the "calls these helpers without passing `dtype=` through" narration, the assertion message that pointed at the deleted wart pin, the "pinned separately below" / "the bug pins below" references in the class preamble, and the `.to(input)` int64 description.

**Minor — the 3-D sizes**

Taken, by changing the sizes rather than weakening the claim. `(2, 4, 5) -> (3, 8, 9)` had four of six axes exactly representable (`2/1`, `2/2`, `2/4`, `2/8`), so only the two heights exercised the rounding. It is now `(4, 4, 6) -> (6, 8, 10)` — `2/3, 2/3, 2/5 -> 2/5, 2/7, 2/9`, every axis load-bearing — and the class docstring lists the actual ratios and warns against "simplifying" them back to powers of two.

**Verification**

`pytest tests/geometry/test_conversions.py tests/geometry/transform/test_homography_warper.py --dtype float32,float64` → **829 passed, 2 skipped, 10 xfailed**, Linux x86-64, torch 2.9.1, CPU.

Caveat on scope, since you were careful about yours: this machine has no CUDA and no MPS, and I have not run the `float16`/`bfloat16` pass, so I am not claiming anything about those. That includes the `float16` double-rounding sizes (`8284`, `16567`, `37263`) you found — I have not reproduced them. I have left the PR body's dtype claim unqualified for now rather than restate a range I have not measured; say the word and I will add your qualifier and the `8.88e-16` rational-oracle figure.

Thanks for the second pass, and for catching that the first "rebase" push was a no-op diff.

ducha-aiki (changes requested):
Review of `aa38d7987c0efd9d74bc40525fb9b3b3056eb94b`.

The implementation and the review follow-up are correct: `normalize_homography`, `denormalize_homography`, and `normalize_homography3d` now build their normalization matrices directly on the input's device and dtype, and the updated non-dyadic float64 pins distinguish the repaired path from the old widened-float32 constants. [P2] The public numerical fix at `kornia/geometry/conversions.py:1969`, `:2337`, and `:2427` has no Towncrier entry. Add `changelog.d/3979.fixed.md` describing that float64 callers now receive float64-native normalization constants and cite #3958.

Validation and limits: Focused CPU float32/float64 run imported `/private/tmp/kornia-triage-60Oa1g/wt-3979/kornia/__init__.py`: 61 passed, 780 deselected. All 39 GitHub checks are green and the head is mergeable. Earlier stale #3958 prose and wart pins are resolved on `aa38d798`.

PR #3980, opened 2026-08-15
Fix NaN gradients at the rotation singularity in quaternion and axis-angle conversions

NoiceHax:

**Fixes/Relates to:** #3949

## What does this PR do?

`quaternion_to_axis_angle` and `axis_angle_to_quaternion` both call `torch.sqrt` on a radicand that is exactly zero at the identity rotation. The forward value is fine, because a `torch.where` picks the constant branch there. But autograd still walks the `sqrt` and the division next to it, and the derivative of `sqrt` blows up at zero, so the backward pass hands back NaN for `(1, 0, 0, 0)` and `(0, 0, 0)`. In a batch, one identity row is enough to NaN that row's gradient.

Both now use the double-`where` form, so the `sqrt` never sees zero on either arm, and the division after it is guarded by the same mask. I picked this over `.clamp(min=1e-12)` because clamp moves the forward value for tiny but non-zero radicands.

Away from zero the output is unchanged bit for bit. At zero the gradients are now the analytical limits: `[0, 2, 2, 2]` and `[0.5, 0.5, 0.5]`.

`test_gradient_at_identity` and `test_gradient_at_zero_rotation` fail on main with NaN and check the values against those limits, not just finiteness. `test_forward_matches_plain_sqrt_formula` recomputes the old unguarded expression inside the test and asserts `torch.equal`, which pins the forward path on any device and dtype.

## Test log

Both new gradient tests on `main`, with only the source fix reverted:

```
$ pixi run test-module tests/geometry/test_conversions.py -k "test_gradient_at_identity or test_gradient_at_zero_rotation or test_forward_matches_plain_sqrt_formula" --dtype=float32,float64

>       assert torch.isfinite(quaternion.grad).all()
E       assert tensor(False)
E        +    where tensor([False, False, False, False]) = torch.isfinite(tensor([nan, nan, nan, nan], dtype=torch.float64))

FAILED tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_gradient_at_zero_rotation[cpu]
FAILED tests/geometry/test_conversions.py::TestQuaternionToAngleAxis::test_gradient_at_identity[cpu]
================= 2 failed, 4 passed, 420 deselected in 3.55s =================
```

With this branch:

```
$ pixi run test-module tests/geometry/test_conversions.py -k "AngleAxisToQuaternion or QuaternionToAngleAxis" --dtype=float32,float64

tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_forward_matches_plain_sqrt_formula[cpu-float32] PASSED [ 44%]
tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_forward_matches_plain_sqrt_formula[cpu-float64] PASSED [ 46%]
tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_gradient_at_zero_rotation[cpu] PASSED [ 48%]
tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_gradcheck[cpu] PASSED [ 50%]
tests/geometry/test_conversions.py::TestQuaternionToAngleAxis::test_forward_matches_plain_sqrt_formula[cpu-float32] PASSED [ 94%]
tests/geometry/test_conversions.py::TestQuaternionToAngleAxis::test_forward_matches_plain_sqrt_formula[cpu-float64] PASSED [ 96%]
tests/geometry/test_conversions.py::TestQuaternionToAngleAxis::test_gradient_at_identity[cpu] PASSED [ 98%]
tests/geometry/test_conversions.py::TestQuaternionToAngleAxis::test_gradcheck[cpu] PASSED [100%]

===================== 52 passed, 374 deselected in 1.75s ======================
```

Wider sweep, plus doctests:

```
$ pixi run test-module tests/geometry --dtype=float32,float64
= 1 failed, 3970 passed, 107 skipped, 11 xfailed, 10 xpassed, 100 warnings in 15.87s =

$ pixi run test-module --doctest-modules kornia/geometry/conversions.py
============================= 32 passed in 1.75s ==============================
```

The one failure is `epipolar/test_essential.py::TestDecomposeEssentialMatrixNoSVD::test_correct_decompose`. I checked it out on unmodified `main` and it fails there the same way, so it is not from this change.

I also checked that both functions still run under `torch.compile(fullgraph=True)` and give the same gradients there, since this file has torchscript accommodations.

- [ ] ðŸŸ¢ **Human-written**

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -996,14 +996,18 @@ def quaternion_to_axis_angle(quaternion: torch.Tensor) -> torch.Tensor:
 
     sin_squared_theta: torch.Tensor = q1 * q1 + q2 * q2 + q3 * q3
 
-    sin_theta: torch.Tensor = torch.sqrt(sin_squared_theta)
+    # the sqrt and the division below are only taken where the radicand is non zero. the inner
+    # `where` keeps the identity rotation away from the sqrt so its backward pass stays finite.
+    pos: torch.Tensor = sin_squared_theta > 0.0
+    safe_sin_squared_theta: torch.Tensor = torch.where(pos, sin_squared_theta, torch.ones_like(sin_squared_theta))
+    sin_theta: torch.Tensor = torch.where(pos, torch.sqrt(safe_sin_squared_theta), torch.zeros_like(sin_squared_theta))
     two_theta: torch.Tensor = 2.0 * torch.where(
         cos_theta < 0.0, torch.atan2(-sin_theta, -cos_theta), torch.atan2(sin_theta, cos_theta)
     )
 
-    k_pos: torch.Tensor = two_theta / sin_theta
+    k_pos: torch.Tensor = two_theta / torch.where(pos, sin_theta, torch.ones_like(sin_theta))
     k_neg: torch.Tensor = 2.0 * torch.ones_like(sin_theta)
-    k: torch.Tensor = torch.where(sin_squared_theta > 0.0, k_pos, k_neg)
+    k: torch.Tensor = torch.where(pos, k_pos, k_neg)
 
     axis_angle: torch.Tensor = torch.zeros_like(quaternion)[..., :3]
     axis_angle[..., 0] += q1 * k
@@ -1247,14 +1251,18 @@ def axis_angle_to_quaternion(axis_angle: torch.Tensor) -> torch.Tensor:
     a2: torch.Tensor = axis_angle[..., 2:3]
     theta_squared: torch.Tensor = a0 * a0 + a1 * a1 + a2 * a2
 
-    theta: torch.Tensor = torch.sqrt(theta_squared)
+    mask: torch.Tensor = theta_squared > 0.0
+
+    # the sqrt and the division below are only taken where the radicand is non zero. the inner
+    # `where` keeps the zero rotation away from the sqrt so its backward pass stays finite.
+    safe_theta_squared: torch.Tensor = torch.where(mask, theta_squared, torch.ones_like(theta_squared))
+    theta: torch.Tensor = torch.where(mask, torch.sqrt(safe_theta_squared), torch.zeros_like(theta_squared))
     half_theta: torch.Tensor = theta * 0.5
 
-    mask: torch.Tensor = theta_squared > 0.0
     ones: torch.Tensor = torch.ones_like(half_theta)
 
     k_neg: torch.Tensor = 0.5 * ones
-    k_pos: torch.Tensor = torch.sin(half_theta) / theta
+    k_pos: torch.Tensor = torch.sin(half_theta) / torch.where(mask, theta, ones)
     k: torch.Tensor = torch.where(mask, k_pos, k_neg)
     w: torch.Tensor = torch.where(mask, torch.cos(half_theta), ones)
 
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -958,13 +958,11 @@ def quaternion_to_axis_angle(quaternion: torch.Tensor) -> torch.Tensor:
         - ``quaternion_to_angle_axis`` is the deprecated alias of this function
           since 0.7.0; see the alias warning on
           :func:`~kornia.geometry.conversions.axis_angle_to_rotation_matrix`
-
-    .. warning::
-        The gradient at the identity quaternion is ``nan``:
-        ``quaternion_to_axis_angle(torch.tensor([1., 0., 0., 0.])).sum().backward()``
-        leaves ``[nan, nan, nan, nan]`` in ``.grad``, from an unclamped
-        ``sqrt(0)``. Away from the identity the gradient is finite. Tracked in
-        `#3949 <https://github.com/kornia/kornia/issues/3949>`_.
+        - differentiable at the identity quaternion. The ``sqrt`` whose
+          derivative is unbounded there never sees its zero radicand, so the
+          gradient is the analytic limit of the surrounding map rather than
+          ``nan``. The guard is elementwise, and away from the identity it
+          moves no forward bit
 
     Args:
         quaternion: tensor with quaternions.
@@ -1216,12 +1214,10 @@ def axis_angle_to_quaternion(axis_angle: torch.Tensor) -> torch.Tensor:
           ``sqrt`` on the way in already produces a floating tensor, and the
           output is allocated from that rather than from the input, so the
           result comes back at the ambient default floating dtype
-
-    .. warning::
-        The gradient at the zero rotation is ``nan``:
-        ``axis_angle_to_quaternion(torch.zeros(3)).sum().backward()`` leaves
-        ``[nan, nan, nan]`` in ``.grad``, from an unclamped ``sqrt(0)``.
-        Tracked in `#3949 <https://github.com/kornia/kornia/issues/3949>`_.
+        - differentiable at the zero rotation. The ``sqrt`` whose derivative is
+          unbounded there never sees its zero radicand, so the gradient is the
+          analytic limit of the surrounding map rather than ``nan``. The guard
+          is elementwise, and away from zero it moves no forward bit
 
     Args:
         axis_angle: tensor with axis angle in radians.
--- a/tests/geometry/test_conversions.py
+++ b/tests/geometry/test_conversions.py
@@ -712,11 +712,24 @@ class TestAngleAxisToQuaternion(BaseTester):
         expected = torch.tensor((np.cos(0.5), np.sin(0.5), 0.0, 0.0), device=device, dtype=input_dtype)
         self.assert_close(quaternion, expected, atol=1.0e-2, rtol=1.0e-2)
 
-    def test_forward_matches_plain_sqrt_formula(self, device, dtype):
-        # away from theta == 0 the guarded sqrt must return the very same bits as the plain
-        # `sqrt(theta_squared)` formula, so the NaN fix cannot move any existing forward value
+    def test_convention_the_guarded_sqrt_moves_no_forward_value_3949(self, device, dtype):
+        # The #3949 fix is a BACKWARD-pass fix: away from theta == 0 the guarded sqrt must return
+        # the very same BITS as the plain `sqrt(theta_squared)` formula it replaced, so no existing
+        # forward value moves. torch.equal, not assert_close, for exactly that reason -- a
+        # tolerance would hide the reassociation this pin exists to rule out.
+        # The fourth row is the near-singular one, and it is derived from the dtype rather than
+        # written as a literal: a fixed 1e-9 UNDERFLOWS to 0 at float16 (its smallest subnormal is
+        # 5.96e-8), which puts the row on the guarded side of the branch, makes the reference
+        # formula's own sin(0)/0 a NaN, and fails this pin under `pixi run test-half` while passing
+        # everywhere else. finfo(dtype).eps squares to something representable in all four dtypes,
+        # so the row stays near-singular AND stays on the unguarded side in each.
+        # Companion: test_convention_gradients_at_the_identity_are_finite_3949 covers theta == 0
+        # itself, where the two formulas deliberately DISAGREE.
+        near_singular = torch.finfo(dtype).eps
         axis_angle = torch.tensor(
-            ((0.1, 0.2, 0.3), (1.0, 0.0, 0.0), (3.0, -2.0, 0.5), (1e-9, 0.0, 0.0)), device=device, dtype=dtype
+            ((0.1, 0.2, 0.3), (1.0, 0.0, 0.0), (3.0, -2.0, 0.5), (near_singular, 0.0, 0.0)),
+            device=device,
+            dtype=dtype,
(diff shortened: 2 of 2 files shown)


ducha-aiki:
Reviewed by execution. **This is the correct fix, done the correct way.**

The double-`where` — a safe radicand inside the `sqrt` and a safe denominator inside the division, with the real mask applied outside — is the right pattern for a NaN-gradient singularity, and it is the one that actually keeps the backward pass finite rather than merely hiding the forward NaN. Applying it symmetrically to both `quaternion_to_axis_angle` and `axis_angle_to_quaternion` is right; they are the same singularity seen from two sides.

`test_forward_matches_plain_sqrt_formula` is the part I want to single out: asserting that away from `theta == 0` the guarded path returns the *same value* as the plain formula is exactly the check that makes a numerical-safety change reviewable. Verified on your head: `tests/geometry/test_conversions.py --dtype=float32,float64` → **419 passed, 7 xfailed**.

One small note on that test: `torch.equal` compares `-0.0` equal to `+0.0`, so it is value-equality, not bitwise. It is fine for what you are asserting here — I mention it only because "bitwise" is a word this repo has been burned by. If you ever want it truly bitwise, compare `t.view(torch.int32)`.

**The required change — and you cannot see it from your base.**

Your base is `8454a81a` (batch 3a). Batches 3b (`#3967`) and 3c (`#3984`) landed after it, and 3b added pins to `tests/geometry/test_conversions.py` that encode #3949 from both directions. I applied your `kornia/` change on top of current `main`:

```
FAILED ...::test_convention_gradients_at_the_identity_are_finite_3949[cpu-axis_angle_to_quaternion-arg1]
        [XPASS(strict)] unguarded sqrt(0) in the axis-angle/quaternion pair makes t...
FAILED ...::test_wart_gradients_at_the_identity_are_nan_3949[cpu-quaternion_to_axis_angle-arg0]
FAILED ...::test_wart_gradients_at_the_identity_are_nan_3949[cpu-axis_angle_to_quaternion-arg1]
4 failed, 603 passed, 29 xfailed
```

All three failures are **the machinery working exactly as designed**, and they are the strongest possible evidence your fix is real:

- `test_convention_gradients_at_the_identity_are_finite_3949` is a **strict xfail asserting the intended behavior**. It is written to XPASS the moment someone fixes #3949. It just did. That is the pin telling you the fix landed.
- The two `test_wart_gradients_at_the_identity_are_nan_3949` cases document the broken behavior and correctly fail now that it is gone.

So, on rebase, in the same commit:

1. Remove the `@pytest.mark.xfail(strict=True)` marker from `test_convention_gradients_at_the_identity_are_finite_3949` so it becomes an ordinary passing test — do **not** delete it; that test is now the permanent regression guard for your fix.
2. Delete both `test_wart_gradients_at_the_identity_are_nan_3949` parametrisations.
3. Note both edits in the PR body.

Please rebase onto `main` and re-run; that will be the first time this change meets the 3b/3c pins.


---

ducha-aiki:
Thanks — the fix is right, and the double-`where` is the correct shape for it. Keeping the singular point away from the `sqrt` instead of clamping the radicand is what makes the forward value at the identity unchanged, and your bit-equality tests are the right way to prove that. I've adopted that pattern as the reference for `#4007`, which is the same class of defect in `quaternion_exp_to_log` and `euler_from_quaternion`.

I rebased onto current `main` and pushed one commit on top of yours (`8d8f5f4c`). Three things in it.

**1. The `#3949` pins had to flip.** The branch was based on `8454a81a`, before `#3967` landed them, so they were written against the unguarded `sqrt` and went red as soon as your fix met them: `test_convention_gradients_at_the_identity_are_finite_3949` is `xfail(strict=True)` and started XPASSing, and `test_wart_gradients_at_the_identity_are_nan_3949` failed. Un-xfailed the first, dropped the second, and folded your two gradient tests into the first — it already parametrizes over both ops, so your analytic gradients became a third column of the table and the batch check became a third claim, rather than a third and fourth test one class apart asserting the same thing.

**2. A real `float16` break in your two forward pins.** They use the `dtype` fixture, so they run under `pixi run test-half`, and the near-singular row was a literal `1e-9`:

```python
>>> float(torch.tensor(1e-9, dtype=torch.float16))
0.0
```

`float16`'s smallest subnormal is `5.96e-8`, so that row underflows to zero. It then lands on the *guarded* side of your own branch, while the reference formula in the test computes `sin(0) / 0` and gets NaN — and `torch.equal` fails. It passes the float32/float64 CI matrix and fails on the half-precision one, which is why CI wouldn't have told you. Both rows now derive from `torch.finfo(dtype).eps`, which squares to something representable in all four dtypes, so the row stays near-singular *and* stays on the unguarded side in each.

Worth generalising: any near-singular literal in a test that takes the `dtype` fixture has to be dtype-derived, or it only holds for the two dtypes CI runs.

**3. The batch comment was wrong, so I corrected it while moving it.** Your comment said "a single zero rotation in a batch must not poison the other rows". I measured it on the unfixed code and NaN never crossed rows:

```
aa2q batch grad (before your fix):   [[nan, nan, nan], [0.4697, 0.4424, 0.4151]]
q2aa batch grad (before your fix):   [[nan, nan, nan, nan], [-1.2632, 2.0155, 1.9249, 1.8344]]
```

The ops are elementwise and `.sum()` sends an independent gradient to each element, so containment was never at risk. The batch cell is still worth having — but for a different reason, which is now what the comment says: it rules out a guard written as a **Python-level** branch (`if theta_squared > 0: ...`), which takes one branch for the whole tensor, passes both scalar cells, and is wrong for every mixed batch. It now also asserts the ordinary row matches its own *unbatched* gradient exactly, which is the claim that actually carries.

Verified: pins fail against unfixed `kornia/geometry/conversions.py` and pass against yours (so they're real pins, not tautologies); `#3949` cells green at float16, bfloat16, float32 and float64; `tests/geometry` + `tests/augmentation` green; no new half-precision failures against `main`. ruff clean.

ducha-aiki (commented):
Re-verified against current `main` (`35555b55`) after `#4006`, `#4028`, `#3977` and `#3995` landed. The fix still applies cleanly and still does what it says — but one thing is missing, and it is on the maintainer side, not yours.

## The fix holds

Gradients at the two singularities, float64:

```text
                                        main                  this PR
quaternion_to_axis_angle([1,0,0,0])     grad has nan          finite
axis_angle_to_quaternion([0,0,0])       grad has nan          finite
```

- `tests/geometry` at float32: **0 failures**, same as `main`.
- `tests/geometry --dtype=float16,bfloat16`: **696 failures on both trees, identical ID sets** — no new half-precision breakage, which is what the dtype-derived near-singular row in the rebase commit was for.
- The four `#3949` pins pass: the strict `xfail` is gone and the wart is retired, so the machinery has already been resolved.

## Missing: the two docstring warnings this fix falsifies

`grep -rnE "#3949|issues/3949" kornia/` on the patched tree still returns two hits, and both describe exactly the defect this PR removes:

**`kornia/geometry/conversions.py:973-978`** (`quaternion_to_axis_angle`):
> The gradient at the identity quaternion is ``nan``: ``quaternion_to_axis_angle(torch.tensor([1., 0., 0., 0.])).sum().backward()`` leaves ``[nan, nan, nan, nan]`` in ``.grad``, from an unclamped ``sqrt(0)``. Away from the identity the gradient is finite. Tracked in `#3949`.

**`kornia/geometry/conversions.py:1227-1231`** (`axis_angle_to_quaternion`):
> The gradient at the zero rotation is ``nan``: ``axis_angle_to_quaternion(torch.zeros(3)).sum().backward()`` leaves ``[nan, nan, nan]`` in ``.grad``, from an unclamped ``sqrt(0)``. Tracked in `#3949`.

I executed both snippets on the patched tree: neither produces a `nan`. Both blocks have to go, and `grep -rnE "#3949|issues/3949" kornia/` has to come back empty before this merges — `AGENTS.md` makes that the gate for a change that closes a documented issue, and a surviving hit means the change is incomplete.

To be clear about whose omission this is: the pin flip and the float16 fix went in as a maintainer commit on top of your work, and that commit should have carried the docstring removal with it. This is our gap, not something you missed. Note also that `#3949` covers only these two functions — the `acos`/`asin` boundary NaNs in `quaternion_exp_to_log` and `euler_from_quaternion` are a different mechanism, tracked separately in `#4007`, so nothing there needs to change here.

Not requesting changes, since there is nothing for you to do. I will push the two docstring deletions unless you would rather take them yourself — say the word either way. Everything else about this PR is done and verified.

---

ducha-aiki (approved):
Pushed the maintainer commit I offered above — `c6735cec`, on top of your branch, no rebase (it was already mergeable).

It retires the two `#3949` warnings this PR falsifies, in `quaternion_to_axis_angle` and `axis_angle_to_quaternion`, and replaces each with a Convention bullet stating what your fix guarantees: differentiable at the singularity, the **analytic** limit rather than a clamped approximation, elementwise, and byte-identical in the forward pass away from the singular point. The numbers stay in `test_convention_gradients_at_the_identity_are_finite_3949` — a docstring is the wrong place for a figure nothing re-executes.

`grep -rnE "#3949|issues/3949" kornia/` is now empty. Checked on the pushed head: `tests/geometry` 2261 passed / 0 failed, `--doctest-modules kornia/geometry/conversions.py` 34 passed, `ruff check` clean. I also grepped for cross-references to the deleted blocks — the only "see the ... warning on" pointers in the file are to the *alias* warning, which is untouched.

Approving. Thanks for the fix and for the double-`where` form — pinning the analytic values rather than just finiteness is what makes it hold up.

---

ducha-aiki (dismissed):
Final verification round, `c6735cec` merged onto current `main` (the PR base was 7 commits behind).

- `grep -rnE "#3949|issues/3949" kornia/ docs/` is empty. The surviving hits are all in `tests/`, each naming its issue — the pins, as intended.
- `tests/geometry/test_conversions.py --dtype=float32,float64`: 606 passed, 20 xfailed. Doctests on `kornia/geometry/conversions.py` pass; ruff `check` and `format` clean.
- Both copies of `test_convention_the_guarded_sqrt_moves_no_forward_value_3949` pass at float32, float64, float16 **and** bfloat16 — the `finfo(dtype).eps` near-singular row does stay on the unguarded side of the branch in all four, which a fixed `1e-9` would not have.
- I re-executed the batch leg's claim about the *unfixed* code rather than taking the comment at face value: on `main`, `batch.grad[0]` is all-NaN and `batch.grad[1]` is already bit-identical to the unbatched gradient, both directions. So the comment is right that the leg pins the shape of the guard (elementwise vs. a Python branch) and is not a NaN-containment claim.

The three `test_{x,y,z}_rotation[cpu-float16]` failures in this class reproduce unchanged on `main` — pre-existing, and CPU float16 is not in the CI matrix.

Approving. One merge-order note: this conflicts with #3978 in both `kornia/geometry/conversions.py` and `tests/geometry/test_conversions.py` — both rewrite the `axis_angle_to_quaternion` Convention block. GitHub reports both mergeable because it only diffs each against `main`. Whichever lands second needs a rebase.

ducha-aiki (approved):
Rebased onto `main` at `d2966e4b` and force-pushed as `595551d4`. The three commits are preserved; only the conflicts against #3978 (`d2966e4b`) were resolved. Re-approving on the new head.

**What the conflicts were, and how they were resolved.** All three sit where #3978 and this PR both edited the `axis_angle_to_quaternion` docstring and the same region of `TestAngleAxisToQuaternion`:

1. **Docstring Convention block.** #3978's "an integral or boolean input is **promoted**" bullet is kept, the `#3949` warning is dropped, and this PR's "differentiable at the zero rotation" bullet is appended after it. Both bullets survive, in that order.
2. **`test_wart_integer_input_returns_the_zero_quaternion_3948`** — #3978 already deleted it on `main` (integer input is promoted now, so the pin's claim is false). Dropped from this PR's side rather than resurrected.
3. **`test_convention_integer_input_is_promoted_to_float_3948` / `test_convention_float_input_keeps_its_dtype_3948`** — #3978's new pins, kept as-is; this PR's `_3949` pins were placed after them.

No content from either side was lost: the rebased branch still carries all three of #3978's `_3948` pins, and none of the four superseded test names survives anywhere in the file.

**Verification at `595551d4`:**

| check | main (`d2966e4b`) | this PR |
|---|---|---|
| `test_conversions.py --dtype=float32,float64` | 614 passed, 20 xfailed | **618 passed, 18 xfailed** |
| `tests/geometry --dtype=float32` | 2399 passed, 17 xfailed, 6 xpassed | **2401 passed, 15 xfailed, 6 xpassed** |

The xfail count drops by exactly 2 in both — the two retired `#3949` strict-xfail cells — and nothing new xpasses. Also: the `#3948`/`#3949` pins pass at float16 and bfloat16 (14 cells); `grep -rnE "#3949|issues/3949" kornia/ docs/` is empty, with the remaining hits confined to `tests/`; `kornia/geometry/conversions.py` doctests pass (34); ruff `check` and `format` clean.

The source diff against `main` is unchanged in substance from what was approved at `c6735cec` — both `where`-guarded `sqrt` sites, both Convention bullets.

ducha-aiki (approved):

PR #3981, opened 2026-08-15
fix(geometry): stop adding eps inside the quaternion sqrt radicand

### LINKED ISSUE #3951: rotation_matrix_to_quaternion never returns a unit quaternion at the default eps: R2q(I) = [1.0000000012499999, 0, 0, 0] because eps is added inside the sqrt
`rotation_matrix_to_quaternion` adds its `eps` **inside** the `sqrt`, so with the default `eps = 1e-8` it never returns a unit quaternion — not even for the identity matrix:

```python
>>> rotation_matrix_to_quaternion(torch.eye(3, dtype=torch.float64)).tolist()
[1.0000000012499999, 0.0, 0.0, 0.0]
>>> rotation_matrix_to_quaternion(torch.eye(3, dtype=torch.float64), eps=0.0).tolist()
[1.0, 0.0, 0.0, 0.0]
```

All four branches do it (`kornia/geometry/conversions.py`, `trace_positive_cond` and `cond_1..3`):

```python
sq = torch.sqrt(trace + 1.0 + eps) * 2.0        # sq = 4 * qw
...
sq = torch.sqrt(1.0 + m11 - m00 - m22 + eps) * 2.0  # sq = 4 * qy
```

Every component is scaled by `sqrt(1 + eps/s)`, i.e. inflated by `~eps/(2s)`, so `||q|| − 1 ≈ 1.25e-09` at the identity and up to `2.5e-09` over random exact rotations. The docstring describes the parameter as "small value to avoid zero division", but the divisions in this function are already protected — they go through `safe_zero_division`, which uses `torch.finfo(...).tiny`. This `eps` protects nothing; it only biases the value.

## Repro

```python
import torch
    quaternion_to_rotation_matrix as q2R,
    rotation_matrix_to_quaternion as R2q,
)

I3 = torch.eye(3, dtype=torch.float64)
print("R2q(I)          =", R2q(I3).tolist())
print("R2q(I, eps=0.0) =", R2q(I3, eps=0.0).tolist())
print("||R2q(I)|| - 1  =", (R2q(I3).norm() - 1).item())

# statistics over 200 *exact* rotation matrices built by q2R of random unit quaternions
g = torch.Generator().manual_seed(5)
worst_rt, worst_norm = 0.0, 0.0
for _ in range(200):
    qq = torch.randn(4, generator=g, dtype=torch.float64)
    qq = qq / qq.norm()
    Rex = q2R(qq)                                   # exact: det=1, RR^T=I to 2e-16
    worst_rt = max(worst_rt, (q2R(R2q(Rex)) - Rex).abs().max().item())
    worst_norm = max(worst_norm, abs(R2q(Rex).norm().item() - 1.0))
print("worst |q2R(R2q(R)) - R| over 200 exact rotations :", worst_rt)
print("worst | ||R2q(R)|| - 1 | over the same 200       :", worst_norm)
```

The construction of the test matrices is load-bearing: they are built by `q2R` of random **unit quaternions**, which yields exact rotations (`det = 1`, `max|R Rᵀ − I| = 0` in float64). Building them with `axis_angle_to_rotation_matrix` instead gives numbers two orders of magnitude worse (`1.455e-06`) — but that is a different defect (the `aa2R` eps, filed separately), not this one.

## Observed

```
R2q(I)          = [1.0000000012499999, 0.0, 0.0, 0.0]
R2q(I, eps=0.0) = [1.0, 0.0, 0.0, 0.0]
||R2q(I)|| - 1  = 1.2499998813808588e-09
worst |q2R(R2q(R)) - R| over 200 exact rotations : 7.327901174747353e-09
worst | ||R2q(R)|| - 1 | over the same 200       : 2.4971434919862645e-09
```

## Expected

`rotation_matrix_to_quaternion` of an exact rotation matrix should return a unit quaternion to working precision: `||q|| − 1` on the order of `1e-16` in float64, not `2.5e-09`. Setting `eps=0.0` already produces exactly that (`[1.0, 0.0, 0.0, 0.0]` for the identity), which shows the algorithm is fine and only the epsilon placement is wrong.

Consequences a user hits:

- `q2R(R2q(R))` is accurate only to `7.3e-09` in float64 for exact rotations, so the round-trip cannot be pinned any tighter than that — a float64 test written at a float64 tolerance fails.
- Downstream code that checks `||q|| == 1` (or normalizes only if it looks non-unit) sees a systematically non-unit quaternion from every call.
- The error floor is dtype-independent in relative terms: `eps = 1e-8` is a fixed absolute quantity, so it dominates float64 precision by seven orders of magnitude while being invisible in float32.

The fix is to move `eps` out of the value and into the guard: clamp the radicand at zero (or at `finfo.tiny`) rather than adding a constant to it, e.g. `sq = torch.sqrt((trace + 1.0).clamp(min=eps)) * 2.0` — which leaves the value exact whenever the radicand is positive and only takes effect where the branch is genuinely degenerate. The `eps` docstring should also be corrected: the divisions are already handled by `safe_zero_division`.

Context: the `+ eps` was introduced to address the NaN gradient reported in #1653 (`sqrt(trace + 1.0)` where the argument can go slightly negative in float32); the proposal there was literally `sq = torch.sqrt(trace + 1.0 + eps) * 2.0`. A `clamp` fixes the same NaN without biasing the value. This is the same defect family as #3939 (`cart2pol`'s `sqrt(x² + y² + eps)`) and the `axis_angle_to_rotation_matrix` eps filed alongside this one — an epsilon folded into a returned value instead of into the gradient path — but a different function with a different magnitude, so it is filed separately.

## Classification

**Broken output, but with a default-changing fix.** "Returns a non-unit quaternion" is wrong under any reading of the function, and the `eps=0.0` path proves the exact answer is one clamp away. However, the fix changes the returned values of **every** call on the default path by `~1e-9`, so it cannot land under byte-identity discipline and it will move any test currently pinned at that precision.

Per the reference-core lifecycle (Part I / Part III §2) my reading is that this is small and unambiguous enough to land as an ordinary broken-output bugfix rather than waiting for the coordinated Repair & Deprecation window — a `1e-9` shift toward the mathematically correct value has no plausible downstream dependent — but since it is technically a default change I am flagging the choice rather than deciding it.

## Environment

kornia 0.9.0rc1 @ `3358ef16`, torch 2.9.1, cpu. Executed from a worktree carrying the batch-3a docs/tests commits on top of `3358ef16`; `rotation_matrix_to_quaternion` is byte-identical to `origin/main` there (no diff hunk lands in it; function body compares AST-identical).

NoiceHax:

**Fixes/Relates to:** Fixes #3951

## What does this PR do?

`eps` was added inside the square root in all four branches, so it shifted the value instead of guarding it. At the default `eps=1e-8` the identity came back as `[1.0000000012, 0, 0, 0]` in float64, which is not a unit quaternion. `eps` is no longer added. A radicand `r` is floored only where `r > eps` is false, by safe-argument substitution (`torch.where(r > eps, r, floor)` inside the square root, value taken from the other `torch.where` arm), in the trace branch and in `cond_1`, `cond_2` and `cond_3`. It is not a `clamp`: `clamp`'s derivative at the bound differs between torch 2.5.1 and 2.14 (see AGENTS.md). A NaN radicand passes through as NaN. The docstring said `eps` avoids zero division, which was never true, so that line now says what it really does. Divisions are already guarded by `safe_zero_division`.

I read this as a repair, not a change of convention. For a real rotation the selected branch's radicand is at least 1, so `1e-8` sits below float32 resolution there. Over 20k random rotations built as `quaternion_to_rotation_matrix` of unit quaternions, float16, bfloat16 and float32 output is bit-identical to main. Only float64 moves, by about 1e-9, and it moves toward the exact value: max `abs(norm(q) - 1)` drops from 2.5e-09 to 6.7e-16.

There is a second effect. `torch.where` runs every branch, so a slightly non-orthogonal input could drive a discarded branch's radicand negative and leak NaN out of its backward. The substitution stops that too.

Three new tests in `TestRotationMatrixToQuaternion` fail on main and pass here. `test_unit_norm_default_eps` builds its inputs with `quaternion_to_rotation_matrix` so they are exact rotations, and the 64 samples hit all four internal branches (23 / 14 / 18 / 9).

## Test log

With the fix reverted and the new tests in place:

```
$ .venv/Scripts/python -m pytest tests/geometry/test_conversions.py --color=no --dtype=float32,float64 \
    -k "identity_default_eps or unit_norm_default_eps or no_nan_near_identity"

E       AssertionError: Tensor-likes are not close!
E       Mismatched elements: 64 / 64 (100.0%)
E       Greatest absolute difference: 2.4440379720047645e-09 at index (39,) (up to 2.220446049250313e-14 allowed)
E        +          where tensor([[nan, nan, nan],
        [nan, nan, nan],
        [nan, nan, nan]], dtype=torch.float64) = tensor(...).grad
FAILED tests/geometry/test_conversions.py::TestRotationMatrixToQuaternion::test_identity_default_eps[cpu-float64]
FAILED tests/geometry/test_conversions.py::TestRotationMatrixToQuaternion::test_unit_norm_default_eps[cpu-float64]
FAILED tests/geometry/test_conversions.py::TestRotationMatrixToQuaternion::test_gradient_no_nan_near_identity[cpu]
================= 3 failed, 2 passed, 420 deselected in 3.73s =================
```

With the fix:

```
$ .venv/Scripts/python -m pytest tests/geometry/test_conversions.py --color=no --dtype=float32,float64

tests\geometry\test_conversions.py ....................................... [ 93%]
............................                                             [100%]
======================= 418 passed, 7 xfailed in 7.20s ========================
```

Nearby callers, same dtypes:

```
$ .venv/Scripts/python -m pytest tests/geometry/test_conversions.py tests/geometry/test_quaternion.py \
    tests/geometry/test_pose.py tests/geometry/liegroup tests/integration/test_conversions.py \
    --color=no --dtype=float32,float64

================= 1566 passed, 9 skipped, 7 xfailed in 17.22s =================
```

Doctests for the module pass (32 passed). float16 and bfloat16 have pre-existing failures in this file; I diffed the failure set on `main` against this branch and it is identical.

- [ ] ðŸŸ¢ **Human-written**

An agent wrote the patch and the tests from the fix the issue proposes, and I checked the diff and ran every test log pasted above.

**Merge order:** this branch contains #4627 and must merge after it. After #4627 squash-merges, re-merge `main` here and take this branch's side of the four `kornia/geometry/conversions.py` hunks.

_Maintainer edit (2026-09-21): retitled, described the final substitution floor instead of the original clamp, `Fixes #3951`, merge-order note._


ducha-aiki:
**The numerics are right and this is a strict improvement — but it is stale: `main` now carries the `#3951` warning prose and three test pins that this patch falsifies, so the rebase is not mechanical.**

Measured on head `d192c70e44d451bdcfe45e588a2a883c1f14b3b0` against merge-base `8454a81a3bd1cafb186bf7d431b3d838bad9a8a9` (not `main`, to keep unrelated commits out), plus a rebase preview onto `origin/main` at `e5e07667`. torch 2.9.1, Python 3.11.14, CPU. No CUDA/MPS available here.

## What is already right

**The forward change is provably confined, and where it moves it moves toward the exact answer.** The concern with `sqrt(X + eps)` → `sqrt(max(X, eps))` is that it changes the output for every input, not just degenerate ones. It does — but only in the direction of removing the old inflation, and only in `float64`.

20 000 random rotations built as `quaternion_to_rotation_matrix` of unit quaternions (seed 42):

```
base torch.float64    max|norm(q)-1|=2.499994e-09  roundtrip max|R-R'|=7.493046e-09
base torch.float32    max|norm(q)-1|=3.576279e-07  roundtrip max|R-R'|=7.152557e-07
base torch.bfloat16   max|norm(q)-1|=1.562500e-02  roundtrip max|R-R'|=3.906250e-02
base torch.float16    max|norm(q)-1|=1.953125e-03  roundtrip max|R-R'|=3.906250e-03

pr   torch.float64    max|norm(q)-1|=6.661338e-16  roundtrip max|R-R'|=1.332268e-15
pr   torch.float32    max|norm(q)-1|=3.576279e-07  roundtrip max|R-R'|=7.152557e-07
pr   torch.bfloat16   max|norm(q)-1|=1.562500e-02  roundtrip max|R-R'|=3.906250e-02
pr   torch.float16    max|norm(q)-1|=1.953125e-03  roundtrip max|R-R'|=3.906250e-03
```

Output delta, same inputs:

```
torch.float64    max|q_new-q_old|=4.265961e-09  bit-identical=False
torch.float32    max|q_new-q_old|=0.000000e+00  bit-identical=True
torch.bfloat16   max|q_new-q_old|=0.000000e+00  bit-identical=True
torch.float16    max|q_new-q_old|=0.000000e+00  bit-identical=True
```

`float64` unit-norm error improves by ~6.6 orders of magnitude (2.5e-09 → 6.7e-16, i.e. to 3 ulp); round-trip `|R - q2r(r2q(R))|` improves by the same factor. Nothing regresses at any dtype. The PR body's numbers reproduce exactly.

**The clamp can never fire on the returned value, for any real 3x3 input.** This is the fact that makes the change safe rather than merely small, and it is worth stating in the PR because it also disposes of the "clamp kills the gradient" objection. For each branch the selected radicand is bounded below by 1:

- `trace_positive_cond` is selected iff `trace > 0`, so `trace + 1 > 1`.
- `cond_1` is selected iff `trace <= 0` and `m00 > m11, m22`. Its radicand is `1 + 2*m00 - trace`. From `trace < 3*m00` it is `> 1 - m00`, and from `trace <= 0` it is `>= 1 + 2*m00`; one of the two is `>= 1` for either sign of `m00`.
- `cond_2` / `cond_3` are the same statement after their selection conditions force `m11` (resp. `m22`) to be the diagonal maximum.

Empirically, over 200 000 arbitrary (non-rotation) `randn*3` matrices:

```
min selected-branch radicand over 200k arbitrary matrices = 1.000036
```

So `min=eps` is only ever reached inside a **discarded** `torch.where` branch, whose gradient contribution is zeroed by `where` anyway. There is no clamped-region-zero-gradient regression to find.

**The NaN premise holds — as a backward-pass NaN, which is what the PR actually claims.** A forward NaN is unreachable (the bound above), and the PR is careful to say "leak NaN out of its backward". Confirmed:

```
base torch.float64 fwd nan=False  grad nan=True
base torch.float32 fwd nan=False  grad nan=True
pr   torch.float64 fwd nan=False  grad nan=False
pr   torch.float32 fwd nan=False  grad nan=False
```
(input: `torch.eye(3)` with `m[0,0] -= 1e-6`)

and at scale, 2000 arbitrary `randn*3` matrices:

```
base arbitrary-input grad nan=True  inf=False
pr   arbitrary-input grad nan=False inf=False
```

**Gradients.** `torch.autograd.gradcheck` in float64 passes on both sides — it never probes the clamped region, per the bound above:

```
base gradcheck[eye+eps-offdiag] -> True    pr gradcheck[eye+eps-offdiag] -> True
base gradcheck[random-rot-batch] -> True   pr gradcheck[random-rot-batch] -> True
base gradcheck default eps -> True         pr gradcheck default eps -> True
```

Analytic-vs-autograd on the smallest reachable radicands (`R = s*I`, `d(qw)/d(m00)` should be `0.25/sqrt(3s+1)`):

```
base s=1e-09 dqw/dm00=0.249999998375   (analytic 0.249999999625)
base s=1e-08 dqw/dm00=0.249999995      (analytic 0.24999999625)
base s=1     dqw/dm00=0.124999999844   (analytic 0.125)

pr   s=1e-09 dqw/dm00=0.249999999625   (analytic 0.249999999625)
pr   s=1e-08 dqw/dm00=0.24999999625    (analytic 0.24999999625)
pr   s=1     dqw/dm00=0.125            (analytic 0.125)
```

The PR matches the analytic derivative exactly at every probe; the base is biased at all of them. So the gradient is *more* accurate too, not just the value.

**Other surfaces, all unchanged base→PR:** batch shapes `(3,3) -> (4,)`, `(1,3,3) -> (1,4)`, `(2,5,3,3) -> (2,5,4)`, `(0,3,3) -> (0,4)`; dtype preserved; `torch.compile(fullgraph=True)` works on both (and returns the exact `[1., 0., 0., 0.]` on the PR vs `[1.0000000012499999, ...]` on the base). `torch.jit.script` fails identically on both sides (`function definitions aren't supported` — the nested `cond_*` closures), so JIT is a pre-existing limitation, not a regression from this patch. `ruff check` and `ruff format --check` are clean.

**No in-kornia caller passes a non-default `eps`**, so the semantic change of the public kwarg reaches no internal user:

```
$ grep -rn "rotation_matrix_to_quaternion" kornia/
kornia/geometry/quaternion.py:36, :419        # Quaternion.from_matrix, default eps
kornia/geometry/conversions.py:547            # rotation_matrix_to_axis_angle, default eps
kornia/geometry/conversions.py:1806           # ARKitQTVecs_to_ColmapQTVecs, default eps
```
Only tests pass `eps=`, and they pass `torch.finfo(dtype).eps`, which is smaller than the default and still below the radicand floor of 1.

**The `safe_zero_division` claim in the new docstring is accurate.** Every division in all four branches goes through it; the only unguarded expressions are the `0.25 * sq` dominant components, which are multiplications. After this patch `sq >= 2*sqrt(1e-8) = 2e-4` in every branch, so that guard is now belt-and-braces rather than load-bearing.

**Test quality is good and the fail-on-base is real.** Three new cells in `TestRotationMatrixToQuaternion`, using `BaseTester`, the injected `device`/`dtype` fixtures and `self.assert_close`; `test_unit_norm_default_eps` uses a local `torch.Generator` so it does not perturb global RNG for the rest of the file. Running the PR's test file against the merge-base *implementation*:

```
$ git checkout d192c70e -- tests/geometry/test_conversions.py   # on base-3981
$ pytest tests/geometry/test_conversions.py -q --device=cpu --dtype=float16,bfloat16,float32,float64 \
    -k "identity_default_eps or unit_norm_default_eps or no_nan_near_identity"
FAILED ...::TestRotationMatrixToQuaternion::test_identity_default_eps[cpu-float64]
FAILED ...::TestRotationMatrixToQuaternion::test_unit_norm_default_eps[cpu-float64]
FAILED ...::TestRotationMatrixToQuaternion::test_gradient_no_nan_near_identity[cpu]
3 failed, 6 passed, 794 deselected
```
and on the PR: `9 passed`. Wider run on the PR — `tests/geometry/test_conversions.py tests/geometry/test_quaternion.py tests/geometry/liegroup tests/integration/test_conversions.py --dtype=float32,float64` → `1556 passed, 7 xfailed`. Module doctests: `32 passed`.

## Blockers

### 1. `main` documents `#3951` in two places and pins it with three tests; this branch retires none of them

This is the AGENTS.md rule: *"before merging a change that closes one of those issues, run `grep -rnE "#NNNN|issues/NNNN" kornia/ tests/`"*. On the merge-base the grep is empty, which is why the PR does not address it — `#3967` (batch 3b) landed the Convention/warning blocks afterwards. On `origin/main` today:

```
$ grep -rnE "#3951|issues/3951" kornia/ tests/
kornia/geometry/conversions.py:656          # rotation_matrix_to_quaternion .. warning::
kornia/geometry/conversions.py:3075         # ARKitQTVecs_to_ColmapQTVecs .. warning::
tests/geometry/test_conversions.py:1286,1304,1313,1317,1328,1345,1358,1365
tests/geometry/test_conversions.py:5858,5867,5888
```

`kornia/geometry/conversions.py:656` is a `.. warning::` on this very function that states the defect as current behavior and would become a lie the moment this merges:

> ``eps`` is added *inside* the square root that produces the dominant component, so with the default the result is not a unit quaternion. In ``float64``, ``rotation_matrix_to_quaternion(torch.eye(3))`` returns ``[1.0000000012499999, 0., 0., 0.]`` ... Tracked in `#3951`.

and the `Args:` line right under it still reads `eps: added inside the square root of the dominant component; see the warning above.` `conversions.py:3075` is the downstream copy on `ARKitQTVecs_to_ColmapQTVecs`. A surviving hit in `kornia/` after this change means the change is incomplete.

I applied only this PR's four-line code change on top of `origin/main` and ran the full file — the three pins break exactly as designed:

```
$ pytest tests/geometry/test_conversions.py -q --device=cpu --dtype=float32,float64
FAILED ...::TestRotationMatrixToQuaternion::test_convention_returns_a_unit_quaternion_3951[cpu]
        - [XPASS(strict)] rotation_matrix_to_quaternion adds eps inside the sqrt, so ...
FAILED ...::TestRotationMatrixToQuaternion::test_wart_eps_inside_the_sqrt_inflates_the_quaternion_3951[cpu]
        - AssertionError: kornia#3951: eps=1e-8 no longer inflates the identity quaternion
FAILED ...::TestCARKitToColmap::test_wart_float64_output_quaternion_is_not_unit_3951[cpu]
        - AssertionError: kornia#3951: the float64 ARKit output quaternion is no longer inflated
3 failed, 616 passed, 15 xfailed
```

That is the pins doing their job — those three failures are the evidence the fix works, not a defect in the fix. But they have to be retired *in this commit*. Rebase onto `main` and, in the same change:

- delete the `.. warning::` block at `conversions.py:656` and rewrite the `Args:` line (this PR's replacement text is the right one);
- delete the `.. warning::` block at `conversions.py:3075` on `ARKitQTVecs_to_ColmapQTVecs` (the second warning there, about the all-zero quaternion, is a different issue and stays);
- drop the `xfail(strict=True)` mark from `test_convention_returns_a_unit_quaternion_3951` so it becomes an ordinary passing convention test;
- delete `test_wart_eps_inside_the_sqrt_inflates_the_quaternion_3951` and `test_wart_float64_output_quaternion_is_not_unit_3951` — a wart pin whose wart is gone has nothing left to pin.

`main` also gained a `Convention:` block on this function whose `[-0.0872, -0.2662, -0.5325, 0.7987]` literal is `float32` and therefore unaffected (float32 output is bit-identical, verified above), so that block needs no edit.

### 2. Mechanical: the branch is stale and GitHub reports it conflicting

`origin/main` has moved 15 commits over `kornia/geometry/conversions.py` and `tests/geometry/test_conversions.py` since `8454a81a`, including `#3980` (NaN gradients at the rotation singularity), `#3982`, `#4043` and the `#3967` docs batch. Please rebase rather than merge, and re-run `--dtype=float32,float64` on `tests/geometry/` after; blocker 1 is most of the real work in that rebase.

## Nits (non-blocking)

1. `kornia/geometry/conversions.py:562` — the docstring says the clamp means *"a slightly non-orthogonal input cannot produce NaN"*, and that is not true in `float16`, where the default `1e-8` flushes to zero on the cast:
   ```
   torch.float16  clamp(min=1e-8) -> 0.0            sqrt -> 0.0
   torch.bfloat16 clamp(min=1e-8) -> 1.0011717e-08  sqrt -> 1.0013580e-04
   ```
   At the exact 180-degree singularity (`R = diag(1,-1,-1)`, so a discarded branch's radicand is exactly `0`), `float16` still leaks a backward NaN on **both** sides:
   ```
   base 180deg float16 grad nan=True     pr 180deg float16 grad nan=True
   base 180deg float64 grad nan=False    pr 180deg float64 grad nan=False (and fwd now exactly [0,1,0,0])
   ```
   Passing `eps=1e-4` clears it (`fp16 eps=1e-4 grad nan: False`). This is pre-existing, not a regression, and I would not hold the PR for it — but either soften the sentence (say it floors the radicand at `eps`, and that a dtype whose resolution swallows `eps` gets a floor of zero), or make the floor dtype-aware. Worth a follow-up issue either way.
2. `conversions.py:563` — the docstring names `safe_zero_division`, which is a closure inside the function body and not a symbol a reader can look up in the rendered docs. Suggest "the divisions are separately guarded against a zero denominator" without the name.
3. `conversions.py:573` — the doctest still passes `eps=torch.finfo(input.dtype).eps` explicitly. That was a workaround for exactly the bug being fixed; after this change `rotation_matrix_to_quaternion(input)` returns `tensor([1., 0., 0., 0.])` on its own (verified). Dropping the argument from the example is the clearest possible demonstration of the fix.
4. `tests/geometry/test_conversions.py:281` and `:290` — only the `float64` parametrizations are load-bearing; `float32`/`float16`/`bfloat16` pass on the merge-base too (`6 passed` in the fail-on-base run above), and `test_unit_norm_default_eps` at `bfloat16` has `tol = 100*7.8e-3 = 0.78`, which is vacuous. Not wrong, just worth knowing that the regression is pinned at one dtype.
5. `tests/geometry/test_conversions.py:283` — the comment writes the old identity as `(1.0000000012, 0, 0, 0)`; the commit message and the `main` warning both use the full `1.0000000012499999`. Match them.

## Repair-window call

**Lands now as an ordinary bugfix, not in the coordinated repair window.** The window exists for semantic and default changes that need to break together. This does not qualify on the evidence above: the default `eps` value is unchanged, `float32`/`float16`/`bfloat16` are bit-identical over 20k rotations, and the only behavior that moves is a `float64` output that was 2.5e-09 off unit norm and is now 6.7e-16 off — a change from wrong to right, in the same direction as the round-trip and the analytic gradient. The `eps` kwarg's *documented* meaning does change, but its documented meaning on `main` is a warning label on a defect, and no caller in the tree relies on it. The backward-pass NaN is squarely the clearly-broken-output class.

Verdict: **changes requested**, but only for the rebase and the `#3951` cleanup in blocker 1. The four-line numerical change is correct as written and I would approve it on sight once the stale prose and the three pins come with it.

ducha-aiki (changes requested):
Setting the formal review state to match the existing review: **changes requested** on head `d192c70e`, which is unchanged since that review.

The outstanding items are in https://github.com/kornia/kornia/pull/3981#issuecomment-5455963864 (rebase, and retire the stale `#3951` warning prose and the three pins this patch falsifies).

NoiceHax:
Rebased onto `main` (`ca5021eb`) and retired the `#3951` prose and pins in the same commit, as asked.

**Blocker 1 — the `#3951` documentation and the three pins**

- deleted the `.. warning::` on `rotation_matrix_to_quaternion` and rewrote the `eps` `Args:` line;
- deleted the `.. warning::` on `ARKitQTVecs_to_ColmapQTVecs`, leaving the `#3952` all-zero-quaternion warning below it untouched;
- dropped the strict `xfail` from `test_convention_returns_a_unit_quaternion_3951` so it is now an ordinary passing convention test, and rewrote its comment to describe the behaviour it guards rather than the defect it used to record;
- deleted `test_wart_eps_inside_the_sqrt_inflates_the_quaternion_3951` and `test_wart_float64_output_quaternion_is_not_unit_3951`.

`grep -rnE "#3951|issues/3951" kornia/` is empty. Under `tests/` the only surviving hits are the convention test's own name, its comment and its assertion message, which per `AGENTS.md` is a pin going on naming its issue rather than a leftover.

**Blocker 2 — the rebase**

Rebased rather than merged; the only conflict was the `eps` `Args:` line, which your review already settled.

**Nits, all taken**

1. The `eps` docstring no longer claims a non-orthogonal input "cannot produce NaN". It now says the radicand is floored at `eps`, that a dtype whose resolution swallows `eps` gets a floor of zero — `1e-8` flushes to `0.0` in `float16` — and that such callers should pass a larger `eps`. I have not made the floor dtype-aware here; that seemed like the follow-up you had in mind rather than something to fold into this PR.
2. Dropped the `safe_zero_division` name; the line now just says the divisions are separately guarded against a zero denominator.
3. Dropped the explicit `eps=` from the doctest — `rotation_matrix_to_quaternion(input)` now returns `tensor([1., 0., 0., 0.])` on its own.
5. The comment now carries the full `1.0000000012499999`.

On nit 4: understood that only the `float64` parametrisations are load-bearing and that the `bfloat16` tolerance on `test_unit_norm_default_eps` is vacuous. I have left the parametrisation as is so the cells still assert nothing regresses at the other dtypes, but I am happy to narrow them to `float64` if you would rather the pin state its actual scope.

**Verification**

`pytest tests/geometry/test_conversions.py --dtype float32,float64` → **635 passed, 2 skipped, 10 xfailed**, on Linux x86-64, torch 2.9.1, CPU. No CUDA or MPS on this machine, so I am claiming nothing about either; the `float16`/`bfloat16` half-precision run is likewise unverified here.

Thank you for the measurements in the review — the branch-radicand bound and the `float16` `eps`-flush were both things I had asserted more confidently than I had checked.

NoiceHax:
CI caught one on the rebase: `test_gradient_no_nan_near_identity` failed on the MPS runner with

```
TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64.
```

The cell hardcodes `float64` — the backward NaN is invisible at lower precision — and drops the dtype fixture so it runs in every configuration, which on `--device=mps` means a `TypeError` that is indistinguishable from a real failure. It now calls `_skip_if_dtype_unavailable(device, torch.float64)` like the other dtype-hardcoded pins in the file, so MPS reports a visible skip instead.

This was latent in the PR rather than new: the branch was conflicting, so the MPS job had not run against it before the rebase. Worth noting that your review and my verification were both CPU-only, so neither pass could have surfaced it.

`pytest tests/geometry/test_conversions.py --dtype float32,float64` → 635 passed, 2 skipped, 10 xfailed on Linux/CPU. Pushed as `58a3a7d8`.

ducha-aiki (changes requested):
Review of `58a3a7d8e95cfbc67e1bfb6894243790cbf89d5d`.

This is an incomplete supported-dtype gradient fix, not a newly introduced half-precision regression: the base sqrt(radicand + eps) also has the underflow hole. The request concerns the newly changed singular expression, its finite-backward coverage, and the repository’s explicit version-independent gradient-guard requirement.

[P2] The new `sqrt(radicand.clamp(min=eps))` expressions at `kornia/geometry/conversions.py:639`, `:647`, `:655`, and `:663` are not a portable backward guard. With the default `eps=1e-8`, `eps` underflows to zero in float16; on supported torch 2.5.1 the derivative of `clamp` at its bound passes the incoming gradient, so a discarded branch evaluates the singular `sqrt(0)` derivative and leaks `nan` through `torch.where`. Torch 2.14 masks the clamp-bound derivative, which is why the new float64-only gradient test passes there. Substitute a positive safe argument in every branch that is differentiated and take the clamped zero value from a detached branch, then add a CPU float16 regression on torch 2.5.1. [P2] This fixes the public quaternion values and removes documented #3951 behavior but adds no Towncrier fragment. Add `changelog.d/3981.fixed.md` describing the old epsilon inflation and the new finite-gradient behavior.

Validation and limits: Focused tests on torch 2.14 imported the PR worktree. Float32/float64 paths pass; the four half failures observed locally are existing known-failure-manifest entries, not new regressions. The version-dependent clamp-at-bound derivative is documented in the repository's precision contract and is specifically discriminated by torch 2.5.1. GitHub's red MPS leg is unrelated network fixture download failure; collector only aggregates it. The torch-2.5.1 leg was not executed locally in this sweep; the cross-version concern follows from the documented clamp-bound behavior and the current expression.

ducha-aiki:
**This PR must merge AFTER #4627.** It now contains #4627's branch (`f29db30fc`), so once #4627 lands, the diff here reduces to this PR's own #3951 delta.

Pushed `7da2db99f`:

1. `db834603d` merges `origin/main` (`088586663`). No conflicts.
2. `8b7373397` merges #4627 at `f29db30fc`, which conflicted in the four `sq = ...` lines. Resolution:
   - Keeps #4627's `_safe_sqrt_sq` and drops the `+ eps` inside every radicand (the #3951 fix).
   - Floors the radicand at `eps` by safe-argument substitution instead of `clamp`: the mask is `r > eps`, the other `where` arm is the constant `2 * sqrt(eps)`, and `sqrt` is only ever differentiated above the floor or at a substituted `1`.
   - Keeps #4627's NaN pass-through, `torch.where(torch.isnan(r), r, out)`.
   - Rewrites the `eps` `Args:` line to describe the floor. The old "a dtype whose resolution swallows `eps` gets a floor of zero" sentence described the `clamp` and no longer applies.
3. `93aaaa5c1` adds a float16 finite-gradient pin at the 90-degree turn about x (`test_float16_gradient_finite_at_a_zero_discarded_radicand`). There the trace branch is selected but the discarded `cond_2` and `cond_3` radicands are exactly `0`. The expected gradient is analytic (`+-sqrt(2)/4`, checked against float64 central differences), not the implementation's own output. The same commit updates the test comments that still described the clamp and adds `changelog.d/3981.fixed.md`. `7da2db99f` drops an issue number from a code comment.

This settles the open P2s from the last review: the `clamp(min=eps)` expressions are gone, and there is a fragment. This PR's retirement of the #3951 prose and pins is kept. The AGENTS closing grep `#3951|issues/3951` over `kornia/` and `docs/` is empty; the `tests/` hits are the renamed convention pin's name, its message, and two past-tense comments. `#4623|issues/4623` has no hits in `kornia/` or `docs/`.

**Checks** (macOS arm64, CPU, worktree import verified)

- `rotation_matrix_to_quaternion(eye(3, float64))` now returns exactly `[1, 0, 0, 0]`. A NaN matrix returns NaN. The fp16/bf16/fp32 gradients at `eye(3)` and `diag(1,-1,-1)` are finite and equal #4627's hardcoded references.
- torch 2.14.0: `tests/geometry/test_conversions.py tests/geometry/test_quaternion.py tests/geometry/liegroup tests/integration/test_conversions.py --dtype=float32,float64` gives 1843 passed, 2 skipped, 8 xfailed. `test_conversions.py` at `float16` / `bfloat16` with `--xfail-known-failures --known-failure-profile=cpu-*` gives 2 failures each, both `test_deg2rad`; the same 2 fail on `origin/main` here because the manifest is Linux-only. Doctests 34 passed. ONNX export plus onnxruntime match eager at identity, at the half-turn and on a NaN input. `dynamo.explain` shows 1 graph and 0 breaks for both functions. Ruff 0.16.7 is clean.
- torch 2.5.1 (scratch venv, CPU): `test_conversions.py --dtype=float32,float64` gives 693 passed, 8 xfailed. The half runs have 1 failure each, `test_convention_grid_sample_needs_align_corners_true` (`grid_sampler_2d_cpu not implemented for Half`), which fails the same way on `origin/main` at 2.5.1.
- Mutation checks across the four conversion test classes and 4 dtypes, on 2.5.1 and 2.14:
  - Putting this PR's previous `torch.sqrt(r.clamp(min=eps)) * 2.0` back turns **4 gradient pins red on 2.5.1 and none on 2.14**: `test_float16_gradient_diag180_4623`, `test_float16_gradient_eye3_4623` (x2) and the new zero-radicand pin. That confirms the version-dependent hole the previous review described, and that the pins catch it on the floor torch.
  - Dropping the NaN pass-through turns 48 NaN pins red.
  - Adding `+ eps` back into one radicand turns `test_identity_default_eps[float64]` and `test_convention_returns_a_unit_quaternion_3951` red.
- Sweep (3502 matrices including arbitrary `randn*3`, 12001 quaternions, 4 dtypes) against #4627's branch: forward and gradients are bit-identical at fp16/bf16/fp32. float64 moves only by the removed `eps` inflation (at most `3.3e-8`, on arbitrary non-rotation matrices). There are 0 non-finite matrix gradients at any dtype.

ducha-aiki (approved):
## Re-review of #3981: `7da2db99f` → `dacef6a7d`

The maintainer account pushed several pieces of this PR in an earlier Claude session: the merge with #4627, the radicand floor, the float16 zero-radicand pin and the fragment. I reviewed them as adversarially as outside code.

**The #3951 delta on top of #4627 is correct, and I have no code finding.** Every item from the 2026-08-28 and 2026-09-14 reviews is addressed: the rebase, the retired #3951 prose and pins, the `clamp` replaced by safe-argument substitution, and the fragment.

### Found at `7da2db99f`, fixed now
- **[P2] Stale title and body, and #3951 would not close.** The title ("clamp the sqrt radicand") and the body described the `clamp` pattern that AGENTS.md forbids and this head removed. `Fixes/Relates to: #3951` was not parsed as a closing keyword. Fixed:
  - The title is now "fix(geometry): stop adding eps inside the quaternion sqrt radicand".
  - The body describes the `r > eps` substitution floor and NaN pass-through.
  - The body says `Fixes #3951` and has a merge-order note.
- **[P3] The fragment re-claimed #4627's finite-gradient fix.** It now covers only the `eps` inflation. It also notes that the selected branch's radicand is ≥ 1 for any real 3x3 matrix, so the floor acts only on discarded branches.
- **Merged #4627's new head (`668408e04`, the `normalize_quaternion` zero-gradient fix) into this branch as `ea89f05db`.** The merge was clean.

### Merge order (gate)
Merge **after #4627**. kornia squash-merges, so once #4627 lands, `git merge-tree` against the squash reports a conflict in 4 hunks of `kornia/geometry/conversions.py`. Re-merge `main` and take this branch's side; I checked that this reproduces this head's tree.

### Verified (torch 2.14.0 and 2.5.1, macOS arm64, worktree import checked)
- **The `clamp` claim reproduces exactly.** Restoring `torch.sqrt(r.clamp(min=eps)) * 2.0` turns 4 float16 gradient pins red on torch **2.5.1** and none on 2.14.
- **Other mutations:**
  - `+ eps` back in the trace radicand kills 3 pins.
  - `+ eps` back in any `cond_*` radicand kills `test_unit_norm_default_eps[float64]`.
  - Dropping the substitution kills 6, including `test_gradcheck`.
  - The survivors are equivalent mutants: the floor is only reached by discarded branches.
- **float64 accuracy:** `eye(3, float64)` gives exactly `[1, 0, 0, 0]`. Over 1804 matrices, covering all four branches and angles near and at 180°, the max `|norm(q) - 1|` drops from 1.25e-9 to 2.2e-16.
- **Lower precisions:** float32, float16 and bfloat16 forward outputs and gradients are identical (`torch.equal`) to #4627's.
- **Tests on `dacef6a7d`:**

  | Scope | Result |
  |---|---|
  | `test_conversions.py` + `test_quaternion.py`, float32/float64 | 871 passed |
  | float16/bfloat16 subset | failure set identical to `main`'s (14 pre-existing) |
  | torch 2.5.1, quaternion classes at 4 dtypes | only the 2 `test_corner_case` half failures, identical on `main` |

- **Other checks:**
  - MPS float32: 380 passed.
  - `dynamo.explain`: 1 graph, 0 breaks.
  - Doctests: 34 passed.
  - ruff 0.16.7: clean.
  - The AGENTS grep for `#3951` is empty in `kornia/` and `docs/`.

**Verdict: approve.** No code items remain open. Before merging: #4627 must merge first, then re-merge `main` here, then hosted CI must pass.

PR #3987, opened 2026-08-19
fix(geometry): enforce essential-matrix constraint in RANSAC polishing

### LINKED ISSUE #3874: [Bug]:  RANSAC essential-matrix polishing step uses find_fundamental, violating the essential-matrix constraint
### 🐛 Describe the bug

kornia.geometry.RANSAC's local-optimization ("polishing") step for
model_type="essential" uses find_fundamental as the polisher solver
(ransac.py: self.polisher_solver = find_fundamental). find_fundamental
solves for a general fundamental matrix and does not enforce the
essential-matrix constraint (that the two nonzero singular values be
equal). Nothing re-projects the polished result back onto the
essential-matrix manifold afterward.

This means RANSAC(model_type="essential") can return a matrix that
does not satisfy the constraint required by decompose_essential_matrix /
motion_from_essential to recover a valid rotation and translation —
silently, with no error raised.

### 🔄 Steps to Reproduce

```bash
1. Generate a set of calibrated point correspondences (representing matches between two images, in normalized/camera coordinates).
2. Call find_essential(points1, points2) on them and compute the singular values of the resulting matrix.
3. Call find_fundamental(points1, points2) on the SAME correspondences and compute its singular values.
4. Compare: a valid essential matrix must have two equal nonzero singular values (ratio sv0/sv1 ≈ 1.0). The fundamental matrix has no such guarantee.
5. Note that kornia.geometry.RANSAC assigns find_fundamental as self.polisher_solver when model_type="essential" (ransac.py, ~line 131-138) — meaning the local-optimization / polishing step can return a matrix that fails this constraint, with no re-projection back onto the essential-matrix manifold afterward.
```

### 💻 Minimal Code Example

```python
import torch
from kornia.geometry.epipolar import find_essential, find_fundamental

torch.manual_seed(0)
N = 30
points1 = torch.rand(1, N, 2) * 2 - 1
points2 = torch.rand(1, N, 2) * 2 - 1

E = find_essential(points1, points2)
F = find_fundamental(points1, points2)

sv_E = torch.linalg.svdvals(E[0, 0])
sv_F = torch.linalg.svdvals(F[0])

print("Essential singular values:", sv_E.tolist())
print("Fundamental singular values:", sv_F.tolist())
print("Essential ratio sv0/sv1 (should be ~1.0):", (sv_E[0] / sv_E[1]).item())
print("Fundamental ratio sv0/sv1 (used as essential polisher):", (sv_F[0] / sv_F[1]).item())
```

### ✅ Expected behavior

Every model RANSAC returns when model_type="essential" — including after local optimization / polishing — should satisfy the essential-matrix constraint (two equal nonzero singular values), since decompose_essential_matrix / motion_from_essential assume this to recover a valid rotation and translation.

### ❌ Actual behavior

Verified numerically on 30 identical point correspondences:

find_essential singular values:   [0.7071067, 0.7071067, 8.0e-08]
  -> ratio sv0/sv1 = 1.0000001   (constraint holds)

find_fundamental singular values: [5.6986055, 2.2652810, 7.2e-08]
  -> ratio sv0/sv1 = 2.5156286   (constraint violated by ~152%)

find_fundamental is the solver kornia.geometry.RANSAC uses for polishing
when model_type="essential" (ransac.py, self.polisher_solver = find_fundamental).
Its output is not projected back onto the essential-matrix manifold before
being returned, so RANSAC(model_type="essential") can return a matrix that
does not correspond to any valid (R, t) pair.

### 🔧 Environment

```shell
- Kornia version:0.9.0rc1
-PyTorch version: 2.9.1+cpu
-Python version: 3.11.14
- OS (e.g., Linux, macOS, Windows): Windows
- Installation method (pip, conda, from source): from source
- CUDA/cuDNN version (if applicable): NA
- GPU model (if applicable): NA

You can also run:

python collect_env.py
```

### 📝 Additional context

_No response_

### 🤝 Contribution Intent

Sigwendice:

**Fixes/Relates to:** # 3874

## What does this PR do?

The RANSAC polishing step for the essential model used `find_fundamental` as the polisher solver, which does **not** enforce the essential-matrix constraint (two equal non-zero singular values, third zero). As a result, `polish_model` could return a matrix that is not a valid essential matrix, and downstream `decompose` calls would silently produce wrong (R, t).

This PR:
- adds `RANSAC._project_to_essential`, which projects the polished matrix back onto the essential manifold via SVD (average the two largest singular values, zero the smallest);
- calls it inside `polish_model` for the essential model so the returned matrix always satisfies the constraint.

Fixes #3874.

## Test log

$ pixi run test tests/geometry/test_ransac.py::TestRANSACEssential -v
tests/geometry/test_ransac.py::TestRANSACEssential::test_smoke PASSED
tests/geometry/test_ransac.py::TestRANSACEssential::test_polish_enforces_essential_constraint PASSED
tests/geometry/test_ransac.py::TestRANSACEssential::test_project_to_essential PASSED
3 passed

(Without the fix, `test_polish_enforces_essential_constraint` fails: the polished matrix has singular-value ratio ≈ 21.9 instead of ≈ 1.)

- [ ] 🟢 **Human-written**


ducha-aiki:
Thanks for the PR! The approach is right and matches the guidance on #3874 — the 8-point solver is fine as a polisher, what was missing is projecting its result back onto the E-manifold. I checked out the branch and verified it locally.

## What I verified

| Check | Result |
|---|---|
| `test-module tests/geometry/test_ransac.py --dtype=float32,float64` | 23 passed, 19 skipped, 4 xpassed |
| Regression test fails on `main` | ✅ `assert tensor(21.9254) < 0.01` — genuine test-first |
| `pixi run lint` / `pixi run typecheck` | both pass |
| Gradient flow through `_project_to_essential` | `torch.autograd.gradcheck` → True |
| `_torch_svd_cast` convention | returns `V` (not `Vᴴ`), so `U @ diag(S) @ V.mH` is correct |
| End-to-end effect (5 synthetic scenes × 8 seeds, 60 inliers + 40 outliers, float64) | returned matrix changed in 1/40 runs, sv0/sv1 **1.0322 → 1.0000** |

One thing worth calling out as done right: the projection happens *before* `verify()` scores the polished model, so the model that gets scored is the model that gets returned. Easy to get backwards.

## Main comment: the fix closes only half of #3874

`forward()` returns `best_model_total`, which is the **minimal-solver** (5-point `find_essential`) model whenever local optimization doesn't beat it. That path is untouched, and `find_essential`'s output is not on the manifold either:

- Over 200 random 5-point problems, worst `|sv0/sv1 - 1|` among the returned candidates: **79.3**
- On the realistic synthetic setup above, **6/40 runs still return an `E` with sv0/sv1 ≠ 1 after this PR**, worst ratio 1.0441

So the expectation stated in the issue — *"every model RANSAC returns when `model_type="essential"` … should satisfy the constraint"* — is still not met. The cleanest closure is to project the minimal candidates as well, right where `forward()` already reshapes them (`ransac.py` L373-374), which keeps scoring and the returned model consistent:

```python
if self.model_type in ["essential", "fundamental_7pt"]:
    models = models.reshape(-1, 3, 3)
if self.model_type == "essential":
    models = self._project_to_essential(models)
```

That is one batched 3×3 SVD per iteration, and it lets both call sites share the same helper. Could also be a follow-up PR if you'd rather keep this one narrow — happy either way, just say which.

## Smaller points

1. **`ransac.py` L317** — `model is not None and len(model) > 0` duplicates the identical guard inside `_project_to_essential`, and `model is None` is dead code given the `-> torch.Tensor` annotation. Keep one of the two.
2. **`ransac.py` L344** — `find_fundamental` already enforces rank 2 (`fundamental.py` L324), so `S[..., 2] = 0` is a no-op on the polisher path today. Correct to keep for a standalone helper, just noting it isn't doing work in this call path.
3. **Test tolerances** — `assert sv[2].abs() < 1e-4` (test L400, L409) is an absolute threshold, but `E` here is not unit-norm (I measured sv0 ≈ 9.75). A relative check (`sv[2] / sv[0]`) would be robust if the polisher's normalization ever changes.
4. **BaseTester coverage** — `test_project_to_essential` exercises a private static method directly, which is fine as a unit check, but the new class has no `test_gradcheck` / `test_jit` unlike the neighbouring `TestRANSACFundamental`.
5. **float16 / bfloat16** — the three new tests fail under `--dtype=float16` (`"lu_cpu"` / `"linalg_svd_cpu" not implemented for 'Half'`). This is a pre-existing class of failure — `TestRANSACFundamental::test_smoke` already fails the same way — so not a new CI break, but it adds three more to a half-precision run of this file.
6. **README typo** (`Perspecive` → `Perspective`) is an unrelated commit bundled in. Harmless, just noting it.

## One concern I checked and dismissed

I expected the new SVD to turn a non-finite polished model into a hard `LinAlgError`, where `main` would silently continue (`remove_bad_models` is never applied to polished models). It does not: `find_fundamental`'s own `eigh` already raises on non-finite input, so the crash surface is unchanged.

## Coordination note

#3991 is open on the other half of this issue (`model_lo.clone()[0]` vs `verify()`'s selected best model in the LO loop). The two are complementary and touch disjoint hunks, so they should merge independently.

---

Sigwendice:
Thanks @ducha-aiki for the careful review and for verifying locally!

I've pushed a follow-up commit (943d3d2) addressing the feedback:

- **Core fix**: `_project_to_essential` is now also applied to the minimal-solver candidates inside `forward()` (right after the reshape, before `verify()`). So the best model returned by `forward()` satisfies the essential-matrix constraint even when local optimization does not win — this closes the other half of kornia#3874 you pointed out. Both call sites now share the same helper.
- **Cleanup**: removed the redundant `model is not None and len(model) > 0` guard in `polish_model` (the guard already lives in `_project_to_essential`).
- **Tests**: relaxed the absolute singular-value tolerance to a relative one (`sv[2]/sv[0]`), since E is not unit-norm; added `test_gradcheck`/`test_jit` stubs to `TestRANSACEssential` (matching `TestRANSACFundamental`); added `test_forward_enforces_essential_constraint` as a regression test for the forward() path (20 random runs asserting the returned E is on the manifold).

Local: `pytest tests/geometry/test_ransac.py::TestRANSACEssential` → 4 passed, 2 skipped.

On the unrelated README typo commit (Perspecive -> Perspective): I can drop it into its own PR to keep this one focused — just say the word. The other half of kornia#3874 (LO loop model selection) is handled in #3991 as you noted, so this PR stays scoped to the manifold projection.

ducha-aiki (changes requested):
Thanks for this — the diagnosis is right and `_project_to_essential` itself is correct (including `V.mH`, which is the right call given `_torch_svd_cast` already returns `V` rather than `Vh`). The branch rebases clean onto current `main` and all 25 tests pass. One substantive change requested.

## The `polish_model` half is right; the `run()`-loop half should come out

`#3874` is about `polish_model` fitting a fundamental matrix with the 8-point DLT and returning it as-is. That half of the PR is on-scope and correct.

The second hunk — projecting every minimal-solver candidate inside `run()`, on every iteration — is a different change, and I don't think it earns its cost. Measured on a 200-point scene with 30% outliers, `RANSAC("essential", inl_th=0.01, max_iter=100)`, median of 5 runs, repeated 3 times:

| | wall clock | returned `E`: `s0/s1`, `s2/s0` | inliers |
|---|---|---|---|
| `main` | 75–80 ms | 1.000552, 1.22e-4 | 144 |
| this PR | **86–87 ms** | 1.000000, 5.09e-08 | 144 |
| project `best_model_total` once at return | 75–79 ms | 1.000000, 5.09e-08 | 144 |

About 12% wall clock for a guarantee that a single projection at the end gives for free.

The premise in the comment doesn't hold either. `find_essential` **does** return matrices on the essential manifold for genuine samples — on true correspondences from a known pose:

```
valid candidates: 4
s0/s1 for valid:  [1.0000000000000004, 1.0000000000000002, 1.000000000000002, 1.0000000000000002]
s2/s0 for valid:  [3.68e-16, 2.01e-16, 3.48e-16, 3.89e-17]
```

What *is* off-manifold is the junk the 5-point solver returns for degenerate or outlier-contaminated samples (I measured ratios of 2.94 and 6.35 there). Projecting those before `verify()` doesn't fix anything — it turns a candidate that should lose into one that looks like a valid essential matrix, so it changes model selection in a direction `#3874` never asked for.

Concretely, I'd suggest dropping the hunk in `run()` and doing this instead:

```python
        # local optimization with all inliers for better precision
        if self.model_type == "essential":
            best_model_total = self._project_to_essential(best_model_total[None])[0]
        return best_model_total, inliers_best_total
```

I prototyped exactly that and **your own `test_forward_enforces_essential_constraint` still passes**, along with the rest of the file (25 passed, 22 skipped). That test is the reason the loop hunk looks necessary, and it isn't: it asserts `s2/s0 < 1e-4` on RANSAC fitted to `torch.rand` points, which is below the float32 accuracy of the 5-point solver, so it fails without *some* projection — but a projection at the end satisfies it just as well.

## Smaller points

- `_project_to_essential` duplicates logic that already exists inline in `find_fundamental`'s rank-2 enforcement (`kornia/geometry/epipolar/fundamental.py:324`, `F_rank2 = U @ torch.diag_embed(S_new) @ V.mH`). Not blocking, but if it moved to `kornia/geometry/epipolar/` as a named helper next to that one, both sites could share it and it would be reusable outside RANSAC.
- `test_forward_enforces_essential_constraint` loops 20 RANSAC fits; it costs ~3s across the two dtypes and the 20th iteration tests nothing the 2nd didn't. A couple of iterations, or a `parametrize`, would give the same coverage.
- The `README.md` typo fix (`Perspecive` → `Perspective`) is unrelated to the PR but harmless — happy to take it along.

Everything else — the docstring, the `(B, 3, 3)` shape handling, the dtype behaviour through `_torch_svd_cast` — looks right to me.

Sigwendice:
Thanks for the thorough second round @ducha-aiki — the perf measurement and the
model-selection argument are compelling; you're right that projecting degenerate
candidates would mask them from verify().

Pushed b52657c:

- Core: dropped the in-loop projection from 943d3d2 entirely; forward() now
  projects best_model_total once, right before returning it (exactly your
  suggested snippet).
- Helper: moved the projection out of RANSAC into
  kornia.geometry.epipolar.project_to_essential — public, documented, exported
  from kornia.geometry.epipolar and added to the docs. Both RANSAC call sites
  (polish_model and the final projection) share it, and it is now reusable
  outside RANSAC. I left the rank-2 enforcement in run_8point as-is since it is
  mathematically different (keeps the two distinct singular values rather than
  equalizing them) — happy to generalize into a shared singular-value helper if
  you prefer.
- Tests: test_forward_enforces_essential_constraint now runs 5 iterations
  instead of 20 (docstring updated to the final-projection semantics);
  test_project_to_essential exercises the public function.

Local: pytest test_ransac.py + epipolar test_fundamental.py + test_essential.py
→ 144 passed.

The failed CI job is the DescriptorMatcherWithSteerer doctest in feature/matching.py,
which downloads pretrained weights at test time — unrelated to this PR (geometry/epipolar only).

ducha-aiki:
**Delta re-review of `b52657c9` — verdict: APPROVE, with one one-line export fix.** Everything from round two is implemented, and implemented well. The red `docs (3.11)` is not yours and a rebase clears it — details below.

Reviewed at head `b52657c9`, merge-base `f0a06c70`, CPU, torch 2.9.1 (the repo `.venv`).

## Round-two asks — all three done

**The in-loop projection is gone**, and `forward()` projects once at the return, exactly as prototyped. Your commit message states the reason better than my review did: degenerate candidates polluted by outliers *should* stay off-manifold so `verify()` can discard them. That is the load-bearing argument, and it is now in the history where it belongs.

**The helper is public.** `kornia.geometry.epipolar.project_to_essential`, with a real docstring naming `decompose_essential_matrix` / `motion_from_essential` as the tools that silently break without it, `KORNIA_CHECK_SHAPE(E_mat, ["*", "3", "3"])` replacing the old ad-hoc `None`/empty guard, and `(*, 3, 3)` batching rather than the old `(B, 3, 3)`. It is in `kornia/geometry/epipolar/__init__.py`'s `__all__` and in `docs/source/geometry.epipolar.rst`.

The `S_new = torch.zeros_like(S)` rewrite is also a genuine improvement over the old `S.clone()` plus explicit zeroing of `S[..., 2]` — same result, and the zero third singular value is now structural rather than an assignment a later edit could drop.

**The test is 5 iterations**, docstring updated to the final-projection semantics.

Leaving `run_8point`'s rank-2 enforcement alone was the right call, and for the reason you gave: it preserves two distinct singular values where this one equalises them. They are different operations that happen to share an SVD. Do not merge them.

## Verified

The guarantee the PR exists to provide now actually holds end-to-end. Over 20 synthetic scenes (200 points, 30% outliers, float64, `inl_th=0.01`, `max_iter=100`), counting returns with `|s0/s1 - 1| > 1e-6` or `s2/s0 > 1e-6`:

```
off-manifold returns over 20 scenes: 0/20
```

For comparison, my round-one measurement on the pre-fix branch had 6/40 still off-manifold.

I also checked the degenerate paths, since removing the old `if model is None or len(model) == 0` guard in favour of `KORNIA_CHECK_SHAPE` could have turned a quiet return into a raise:

```
project_to_essential(torch.zeros(1, 3, 3))  ->  (1, 3, 3), finite, all-zero    # no raise
RANSAC("essential") on pure noise, n=8 and n=20, float32 and float64
  ->  (3, 3), finite, all-zero                                                 # unchanged from main
```

`best_model_total` is initialised to `torch.zeros(3, 3)` at `ransac.py:368`, so the no-model-found path feeds a zero matrix into the SVD; it comes back zero rather than raising or producing NaN. Good.

```console
$ pytest tests/geometry/test_ransac.py tests/geometry/epipolar -q --device=cpu --dtype=float32,float64
468 passed, 22 skipped, 12 xpassed
```

## The one ask — add it to `essential.py`'s own `__all__`

`project_to_essential` is exported from the package but is missing from the module's list:

```console
$ grep -n "__all__" -A 10 kornia/geometry/epipolar/essential.py
33:__all__ = [
34-    "decompose_essential_matrix",
35-    "decompose_essential_matrix_no_svd",
36-    "essential_from_Rt",
37-    "essential_from_fundamental",
38-    "find_essential",
39-    "motion_from_essential",
40-    "motion_from_essential_choose_solution",
41-    "relative_camera_motion",
42-]                                    # <- project_to_essential absent
```

It works — the package `__init__` re-exports it and `from kornia.geometry.epipolar import project_to_essential` imports fine — so this is hygiene, not breakage. But #4026 tracks 57 public names that are importable while excluded from their own module's `__all__`, and this would make 58. Cheaper to add the line now than to sweep it later. Alphabetically it goes after `motion_from_essential_choose_solution`.

## The red `docs (3.11)` is stale — rebase and it goes green

Your attribution is right, and I can be more specific about the fix:

```
kornia/feature/matching.py::kornia.feature.matching.DescriptorMatcherWithSteerer  FAILED
1 failed, 654 passed, 2 skipped   (run completed 2026-08-29T01:56)
```

That doctest was fixed by **#4085**, merged as `d2166cfa` at 2026-08-29 17:46 — about sixteen hours *after* your run — which rewrote that exact example into a small CPU one so it no longer downloads weights. Nothing in `kornia/geometry/epipolar` or `kornia/geometry/ransac.py` is involved. Rebase onto current `main` and that job should pass.

Thanks for taking the perf argument seriously rather than defending the first version — moving the helper out to a public, documented, reusable function was more than I asked for and is the better outcome.

Sigwendice:
Thanks for the thorough re-review, @ducha-aiki.

The one-line export fix is now in: `project_to_essential` is added to `essential.py`'s `__all__` (commit `da839f8`). I also rebased the branch onto current `main`, so the stale `docs (3.11)` job (the `DescriptorMatcherWithSteerer` doctest, fixed by #4085) should pass on re-run.

Could you take another look and re-approve when you get a chance? The force-push reset the review state, so the PR still shows the old "requested changes" on `943d3d2`.

ducha-aiki:
**APPROVE** — the round-3 ask is done, re-verified at `da839f82`.

Setup: worktree at the PR head, `~/dev/kornia/.venv/bin/python` (torch 2.9.1).

```
in essential.__all__: True
in kornia.geometry.epipolar.__all__: True
reachable: kornia.geometry.epipolar.project_to_essential -> callable
docs/source/geometry.epipolar.rst:18: .. autofunction:: project_to_essential
d2166cfa is an ancestor of HEAD   # the rebase that clears the stale docs (3.11) job
```

So the export is complete on both levels — `essential.py`'s own `__all__` and the package re-export — and it would not have become #4026's 58th orphan. Nothing else changed since the round-2 verification.

Thanks for the patience across three rounds; the manifold-projection argument and the public helper both came out better for it.

ducha-aiki (approved):

Issue #3947, opened 2026-08-10
axis_angle_to_rotation_matrix returns non-orthogonal matrices: det = 0.9999974535249636 in float64 at every angle, because eps is added to theta in the axis normalization


ducha-aiki:
`axis_angle_to_rotation_matrix` adds `eps = 1e-6` to `theta` in the axis normalization, so the returned matrix is not a rotation matrix: in **float64** at `theta = pi/2` it has `det = 0.9999974535249636` and `max|R Rᵀ − I| = 2.5464750363912714e-06`. The error is present at every angle tested, does not shrink with dtype, and is six orders of magnitude larger than float64 precision. Below `theta = 1e-3` a second, undocumented branch returns the raw first-order Taylor matrix `[[1, −rz, ry], [rz, 1, −rx], [−ry, rx, 1]]`, whose determinant is `1 + theta²` — also not a rotation. kornia's own quaternion route computes the same rotation exactly, so no third-party reference is needed to demonstrate the defect.

The source (`kornia/geometry/conversions.py:433-435`, and the branch at `:469-497`):

```python
def _compute_rotation_matrix(axis_angle: torch.Tensor, theta2: torch.Tensor, eps: float = 1e-6) -> torch.Tensor:
    theta = torch.sqrt(theta2.clamp(min=1e-12))  # clamping to ensure no nan gradients
    wxyz = axis_angle / (theta.unsqueeze(-1) + eps)  # (B, 3)
```

The unit axis is computed as `v / (theta + eps)` instead of `v / theta`, which shrinks it by a factor `theta / (theta + eps) = 1 − eps/theta + …`. Rodrigues' formula then builds a matrix from a *non-unit* axis, which is not orthogonal. The `theta2.clamp(min=1e-12)` on the line above already handles the NaN-gradient concern that `eps` looks like it is there for (see #2924, closed) — the `+ eps` on the next line is doing nothing for the gradient and everything to the value.

## Repro

```python
import torch
from kornia.geometry.conversions import (
    axis_angle_to_quaternion as aa2q,
    axis_angle_to_rotation_matrix as aa2R,
    quaternion_to_rotation_matrix as q2R,
)

I3 = torch.eye(3, dtype=torch.float64)
v = torch.tensor([[0.0, 0.0, torch.pi / 2]], dtype=torch.float64)

R = aa2R(v)[0]
[...]

PR #3989, opened 2026-08-20
fix(geometry): make axis_angle_to_rotation_matrix return orthogonal matrices (kornia#3947)

lawliet206:
## Which lane? *(check one – see [Contributing Guide](https://github.com/kornia/kornia/blob/main/CONTRIBUTING.md#which-lane-is-your-contribution))*
- [x] 🟢 **Green lane** – test-first bug fix / verified docs fix / [`help wanted`](https://github.com/kornia/kornia/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22help%20wanted%22) issue / benchmark results. No prior issue needed.
- [ ] 🟠 **Discuss-first** – feature or behavior change; a maintainer confirmed the scope on the linked issue.

**Fixes/Relates to:** #3947

Green-lane category: **test-first bug fix**. `test_convention_both_branches_are_orthogonal_3947` fails on `main` without this fix and passes with it.

## What does this PR do?

Fixes the non-orthogonality of `axis_angle_to_rotation_matrix` (#3947).

**Root cause (both branches of the function were broken, each for its own reason):**

1. **General branch** – `_compute_rotation_matrix` normalised the axis by `(theta + eps)` with `eps = 1e-6` hardcoded, shrinking the axis and returning a slightly scaled rotation. At `theta = pi/2` about `+z` in `float64`: `det(R) = 0.9999974535249636` and `max|R R^T - I| = 2.5464750363912714e-06`, with no improvement in `float32`. The `theta2.clamp(min=1e-12)` on the preceding line already guards the gradient path (the exact fix the issue's scope-correction comment points at), so the `eps` added nothing but error.

2. **Low-angle branch** – returned the first-order Taylor matrix `[[1, -rz, ry], [rz, 1, -rx], [-ry, rx, 1]]` with `det = 1 + theta**2` (e.g. `1.000001` at `theta = 1e-3`), which is not a rotation either.

**Fix:**
- General branch now normalises by the exact `theta` (the `clamp` still protects the backward pass).
- Low-angle branch now returns the second-order Taylor expansion `R = I + [v]x + [v]x^2 / 2`, whose determinant is `1 + theta**4 / 4` – a rotation to the working precision across the whole branch (`1 + 2.5e-13` at `theta = 1e-3`).

**Result:** both branches pass the same orthogonality/determinant checks, and the general branch agrees with the independent quaternion route `quaternion_to_rotation_matrix(axis_angle_to_quaternion(v))` to machine precision (max diff `4.4e-16` on a generic axis).

**Tests:**
- The strict `xfail` pinning the broken behaviour (`test_convention_returns_an_orthogonal_matrix_3947`) now passes and the marker is removed.
- The wart pin (`test_wart_axis_normalisation_eps_breaks_orthogonality_3947`) is flipped into `test_convention_both_branches_are_orthogonal_3947`, which asserts orthogonality on **both** branches plus the quaternion-route cross-check. **This test fails on `main` without the fix** (the general-branch cell fails on `det`/orthogonality; the low-angle cell fails because the first-order matrix has `det = 1.000001`).
- The inaccurate warning block in the docstring (which documented the broken output as if it were permanent) is removed.

## Test log

Local run (Windows, CPU, `torch 2.13.0+cpu`, `python 3.13.12`):

```
$ python -m pytest tests/geometry/test_conversions.py -q
================= 284 passed, 21 xfailed, 8 warnings in 5.60s =================
```
(21 xfailed are all pre-existing pins for other issues: #3955, #3956, etc. – none related to #3947.)

```
$ python -m pytest tests/geometry/calibration/test_pnp.py -q
============================= 17 passed in 3.98s ==============================
```

(`test_pnp.py` is the only other test module depending on `axis_angle_to_rotation_matrix`.)

```
$ python -m pytest --doctest-modules kornia/geometry/conversions.py -q
============================= 32 passed in 3.82s ==============================
```

```
$ ruff check kornia/geometry/conversions.py tests/geometry/test_conversions.py
All checks passed!
$ ruff format --check kornia/geometry/conversions.py tests/geometry/test_conversions.py
2 files already formatted
```

**Reproduction on `main` (before this fix):**

```python
import torch
from kornia.geometry.conversions import axis_angle_to_rotation_matrix as aa2R

v = torch.tensor([[0.0, 0.0, torch.pi / 2]], dtype=torch.float64)
R = aa2R(v)[0]
print(torch.linalg.det(R).item())                 # 0.9999974535249636 (main) -> 1.0 (this PR)
print((R @ R.T - torch.eye(3)).abs().max().item())  # 2.5464750363912714e-06 -> 0.0
```

If maintainers prefer routing the low-angle branch through the guarded Rodrigues form instead of the second-order Taylor expansion, I can adjust.

## AI Usage Disclosure *(pick the closest – see [AI_POLICY.md](https://github.com/kornia/kornia/blob/main/AI_POLICY.md))*
- [ ] 🟢 **Human-written**
- [x] 🟡 **AI-assisted** – AI helped; I reviewed the resulting change and ran the relevant tests
- [ ] 🔴 **AI-generated**

AI was used only for code-style/lint standardization and English translation of this description. All test logs above are real local runs on this branch.

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -424,29 +424,7 @@ def axis_angle_to_rotation_matrix(axis_angle: torch.Tensor) -> torch.Tensor:
           to this function, returning an equal result — see the alias warning
           below
 
-    .. warning::
-        The returned matrix is **not orthogonal**: ``eps = 1e-6`` is added to
-        the angle when the axis is normalised, which shrinks the axis. In
-        ``float64`` at ``theta = pi/2`` about ``+z``, ``det(R)`` is
-        ``0.9999974535249636`` and ``max|R R^T - I|`` is
-        ``2.5464750363912714e-06`` (torch 2.9.1, cpu — the trailing digits are
-        backend-dependent; the magnitude is the point); the determinant is
-        axis-independent to the last digit or two, while the orthogonality
-        residual is not (a generic axis gives
-        ``2.091747640764474e-06``). ``float32`` is no better
-        (``2.5033950805664062e-06`` on the same input), and
-        the second example below hides it — the printed ``1.0000e+00`` at
-        ``R[0, 0]`` is really ``0.9999987483024597``. Below an internal
-        threshold on ``theta ** 2`` the first-order matrix
-        ``[[1, -rz, ry], [rz, 1, -rx], [-ry, rx, 1]]`` is returned instead, with
-        ``det = 1 + theta ** 2``: in ``float64`` the input ``[0., 0., 1e-3]``
-        takes that branch (``det = 1.000001``) while ``1e-3 * (1, 2, 3)/sqrt(14)``
-        does not (``det = 0.9999999970044947``), so which branch an input takes
-        depends on its axis and dtype. Tracked in
-        `#3947 <https://github.com/kornia/kornia/issues/3947>`_.
-
-    .. warning::
-        Only rank-2 input is accepted, despite the guard's ``(*, 3)`` message:
+    .. warning::        Only rank-2 input is accepted, despite the guard's ``(*, 3)`` message:
         ``(3,)`` raises ``IndexError: Dimension out of range``, ``(2, 5, 3)``
         raises ``ValueError: too many values to unpack (expected 3)`` and
         ``(1, 1, 3)`` raises ``ValueError: not enough values to unpack``.
@@ -491,9 +469,9 @@ def axis_angle_to_rotation_matrix(axis_angle: torch.Tensor) -> torch.Tensor:
     if not axis_angle.shape[-1] == 3:
         raise ValueError(f"Input size must be a (*, 3) tensor. Got {axis_angle.shape}")
 
-    def _compute_rotation_matrix(axis_angle: torch.Tensor, theta2: torch.Tensor, eps: float = 1e-6) -> torch.Tensor:
+    def _compute_rotation_matrix(axis_angle: torch.Tensor, theta2: torch.Tensor) -> torch.Tensor:
         theta = torch.sqrt(theta2.clamp(min=1e-12))  # clamping to ensure no nan gradients
-        wxyz = axis_angle / (theta.unsqueeze(-1) + eps)  # (B, 3)
+        wxyz = axis_angle / theta.unsqueeze(-1)  # (B, 3)
         wx, wy, wz = wxyz.unbind(dim=1)  # (B,)
 
         cos_theta = torch.cos(theta)
@@ -530,18 +508,27 @@ def axis_angle_to_rotation_matrix(axis_angle: torch.Tensor) -> torch.Tensor:
     def _compute_rotation_matrix_taylor(axis_angle: torch.Tensor) -> torch.Tensor:
         rx, ry, rz = axis_angle.unbind(-1)
         k_one = torch.ones_like(rx)
+        k_half = 0.5 * k_one
 
+        rx2, ry2, rz2 = rx * rx, ry * ry, rz * rz
+        rxry, rxrz, ryrz = rx * ry, rx * rz, ry * rz
+
+        # second-order Taylor expansion of Rodrigues' formula:
+        #   R = I + [v]x + [v]x^2 / 2
+        # the first-order truncation had det = 1 + theta^2; the second-order
+        # truncation has det = 1 + theta^4 / 4, so the matrix is a rotation to
+        # the working precision across the whole low-angle branch
         rot = torch.stack(
             [
-                k_one,
-                -rz,
-                ry,
-                rz,
-                k_one,
-                -rx,
-                -ry,
-                rx,
-                k_one,
+                k_one - k_half * (ry2 + rz2),
+                -rz + k_half * rxry,
+                ry + k_half * rxrz,
+                rz + k_half * rxry,
+                k_one - k_half * (rx2 + rz2),
+                -rx + k_half * ryrz,
+                -ry + k_half * rxrz,
+                rx + k_half * ryrz,
+                k_one - k_half * (rx2 + ry2),
             ],
             dim=-1,
         ).view(-1, 3, 3)
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -565,11 +565,12 @@ def rotation_matrix_to_axis_angle(rotation_matrix: torch.Tensor) -> torch.Tensor
           :math:`(3,)` and :math:`(2, 5, 3)` results above cannot be fed
           straight back (see its shape warning)
         - the round trip through
-          :func:`~kornia.geometry.conversions.axis_angle_to_rotation_matrix` is
-          accurate only to about ``1e-6`` even in ``float64`` — measured
-          ``8.0e-07`` at ``theta = 1e-3`` and ``5.4e-07`` at ``theta = pi``
-          about ``(1, 2, 3)/sqrt(14)`` — because of that function's
-          `#3947 <https://github.com/kornia/kornia/issues/3947>`_
+          :func:`~kornia.geometry.conversions.axis_angle_to_rotation_matrix`
+          is accurate to about ``5e-09`` in ``float64`` — measured
+          ``2.0e-12`` at ``theta = 1e-3`` and ``4.2e-09`` at ``theta = pi``
+          about ``(1, 2, 3)/sqrt(14)`` — the latter is
+          :func:`~kornia.geometry.conversions.rotation_matrix_to_axis_angle`'s
+          own conditioning near the ``pi`` singularity
         - the input is **not** checked for being a rotation matrix:
           ``zeros(3, 3)`` returns ``[0., 0., 3.1416]``, ``2 * eye(3)`` returns
           ``[0., 0., 0.]``, and the reflection ``diag(-1, 1, 1)``
@@ -2827,11 +2828,9 @@ def camtoworld_to_worldtocam_Rt(R: torch.Tensor, t: torch.Tensor) -> tuple[torch
           and ``1e-15`` in ``float64``, over 64 unit-normalized random
           quaternions turned into rotations via
           :func:`~kornia.geometry.conversions.quaternion_to_rotation_matrix`
-          — an orthogonality-preserving route; contrast
-          :func:`~kornia.geometry.conversions.axis_angle_to_rotation_matrix`,
-          whose own non-orthogonality (`#3947
-          <https://github.com/kornia/kornia/issues/3947>`_) inflates this
-          figure to ``3.16e-06`` — with matching random translations, both
+          — an orthogonality-preserving route, as is
+          :func:`~kornia.geometry.conversions.axis_angle_to_rotation_matrix`
+          — with matching random translations, both
           drawn from ``torch.Generator().manual_seed(seed)``, ``seed=0`` — a
           few ulps of the entries. Read the exponent, not the digits: the
           maximum moves with the draw (``4.77e-07`` to ``8.34e-07``, and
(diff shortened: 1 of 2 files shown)


ducha-aiki:
**Changes requested — the maths is right and I could not break it, but the fix leaves a falsified `#3947` paragraph behind in `kornia/` and the branch needs a rebase. Both are small; the substance is sound and I'd approve on the follow-up.**

Measured on head `775863f1e213b5e414be51eefcd7e6a5be207bed` against merge-base `b4f5a78a13891c7ac4c1cd9695941380273e981d` (not `main` — comparing against `main` would have mixed in #3980/#3978/#4043, which changed neighbouring functions in this same file). torch 2.9.1, Python 3.11, CPU. Every number below is from a run on both worktrees.

## What is already right

**It delivers, and by more than the description claims.** Sweep over 256 random axes plus the three canonical ones at each magnitude:

```
                  base b4f5a78a            PR 775863f1
float64  worst max|det-1|    2.546475e-06     2.504663e-13
float64  worst max|RR^T-I|   2.546477e-06     2.502443e-13
float32  worst max|det-1|    2.801418e-06     5.364418e-07
float32  worst max|RR^T-I|   2.920628e-06     7.152557e-07
```

float64 improves by 7 orders of magnitude; float32 is now at its own epsilon floor. The issue's headline input reproduces exactly:

```
theta=pi/2 about +z, float64
  base:  det(R) = 0.9999974535249636   max|R R^T - I| = 2.5464750363912714e-06
  PR:    det(R) = 1.0                  max|R R^T - I| = 0.0
```

**The `det = 1 + theta^4/4` claim in the new comment is correct.** Verified symbolically and numerically. For `R = I + K + K^2/2` with `K = [v]x`, the eigenvalues of `K` are `0, ±i*theta`, so `det(R) = 1 * ((1 - theta^2/2)^2 + theta^2) = 1 + theta^4/4` exactly. Numerically it matches to machine precision at every theta I tried:

```
theta=0.001  det=1.0000000000002502   1+th^4/4=1.00000000000025    diff=2.220e-16
theta=0.01   det=1.0000000025         1+th^4/4=1.0000000025        diff=0.000e+00
theta=0.1    det=1.000025             1+th^4/4=1.000025            diff=0.000e+00
```

The orthogonality residual follows the same law and is worth stating because it bounds the whole branch: `R R^T = I + K^4/4 = I - theta^2 K^2 / 4`, i.e. `O(theta^4/4)`, which is `2.5e-13` at the branch ceiling `theta = 1e-3` and shrinks quartically below it. That is where the `2.504663e-13` in the table above comes from — it is the Taylor truncation, not a defect.

**I could not reproduce the zero-division risk that removing `eps` looks like it should introduce, and I went looking hard for it.** The `+ eps` was never what protected `theta = 0`; `theta2.clamp(min=1e-12)` sits *inside* the `sqrt`, so `theta >= 1e-6` unconditionally and the new `axis_angle / theta` cannot divide by zero. Forward and backward at exactly zero and at denormals:

```
                     base                          PR
fwd [0,0,0] f32/f64  finite=True, R=I              finite=True, R=I
fwd 1e-320  f32/f64  finite=True                   finite=True
bwd [0,0,0] f32/f64  grad=[0,0,0]  finite=True     grad=[0,0,0]  finite=True
bwd 1e-30   f64      grad=[0,0,0]  finite=True     grad=[-1.25e-30,2.5e-31,1e-30] finite=True
```

`torch.autograd.gradcheck` in float64 passes at **exact zero** and either side of the threshold on both worktrees — the `torch.where` masked-branch NaN trap is not triggered:

```
gradcheck exact zero       -> True (base)  True (PR)
gradcheck 1e-8             -> True         True
gradcheck 5e-4 (taylor)    -> True         True
gradcheck 2e-3 (normal)    -> True         True
gradcheck pi/2 z           -> True         True
```

**Unclaimed bonus: the low-angle branch's gradient was previously wrong, and this fixes it.** The first-order matrix is affine in `v`, so its second-order derivative information is simply absent. Against exact Rodrigues differentiated by autograd in float64:

```
theta~1e-4   base   d(sum R)/dv = [0.0, 0.0, 0.0]                    err = 1.250e-04  (100% wrong)
             PR     d(sum R)/dv = [-0.000125, 2.5e-05, 0.0001]       err = 2.537e-13
theta~5e-4   base   d(sum R)/dv = [0.0, 0.0, 0.0]                    err = 6.250e-04
             PR     d(sum R)/dv = [-0.000625, 0.000125, 0.0005]      err = 2.627e-11
```

Anyone optimising through a small rotation was getting an identically-zero gradient from this branch. Worth putting in the PR description.

**Branch discontinuity drops ~6000x**, which matters because the switchover is where a fix like this usually introduces a seam:

```
max|R_taylor - R_normal| at theta = 1e-3, float64
  axis +z:       base 9.991655e-07   ->  PR 1.646667e-10
  axis generic:  base 8.724006e-07   ->  PR 1.320327e-10
```

**Tests are the right shape and genuinely fail on the merge-base.** I copied the PR's test file onto `b4f5a78a` and ran it — both tests fail there, for the documented reason, and the strict `xfail` removal is therefore justified rather than cosmetic:

```
FAILED tests/geometry/test_conversions.py::TestAngleAxisToRotationMatrix::test_convention_returns_an_orthogonal_matrix_3947[cpu]
FAILED tests/geometry/test_conversions.py::TestAngleAxisToRotationMatrix::test_convention_both_branches_are_orthogonal_3947[cpu]
E   AssertionError: kornia#3947: axis_angle_to_rotation_matrix did not return an orthogonal matrix
E   assert 2.5464750363912714e-06 < 1e-12
```

Suites on the PR, `--device=cpu --dtype=float32,float64`:

```
tests/geometry/test_conversions.py                       500 passed, 27 xfailed
pnp + epipolar + integration/test_conversions + liegroup 1434 passed, 8 xpassed
--doctest-modules kornia/geometry/conversions.py         32 passed
ruff check / ruff format --check                         clean
```

The `8 xpassed` are identical on the merge-base (`1434 passed, 8 xpassed`), so they are pre-existing and not caused by this change.

**Blast radius is smaller than it looks.** `So3.exp`/`Se3` do **not** route through this function — they build the quaternion directly — so the Lie-group layer is untouched. The only in-library callers are `kornia/geometry/epipolar/scene.py:44` (`generate_scene`, a synthetic-data helper) and `kornia/geometry/transform/imgwarp.py:1176` (`get_projective_transform`). Both suites pass unchanged, and no test pins the old non-orthogonal values other than the two this PR rewrites.

## Blockers

### 1. `kornia/geometry/conversions.py:578-583` — a `#3947` paragraph survives in `kornia/`, and this PR falsifies its literals

Per AGENTS.md, closing a documented issue means removing its prose everywhere; a surviving hit in `kornia/` means the change is incomplete. The full grep on the PR head:

```
$ grep -rnE "#3947|issues/3947" kornia/ tests/
kornia/geometry/conversions.py:583:          `#3947 <https://github.com/kornia/kornia/issues/3947>`_
tests/geometry/test_conversions.py:1879, 1901, 1904, 1908, 1951, 1954, 1964
```

The `tests/` hits are all inside the two rewritten regression tests naming their issue, which is correct and expected. The `kornia/` hit is the `rotation_matrix_to_axis_angle` Convention bullet:

```
578         - the round trip through
579           :func:`~kornia.geometry.conversions.axis_angle_to_rotation_matrix` is
580           accurate only to about ``1e-6`` even in ``float64`` — measured
581           ``8.0e-07`` at ``theta = 1e-3`` and ``5.4e-07`` at ``theta = pi``
582           about ``(1, 2, 3)/sqrt(14)`` — because of that function's
583           `#3947 <https://github.com/kornia/kornia/issues/3947>`_
```

This is not merely a stale cross-reference — the two hardcoded numbers are measurements of the bug this PR removes, and they were exact on the base:

```
round trip, axis (1,2,3)/sqrt(14), float64
  theta=1e-3   doc says 8.0e-07   base measured 8.009844e-07   PR measured 2.004460e-12
  theta=pi     doc says 5.4e-07   base measured 5.387206e-07   PR measured 4.198129e-09
```

So after this PR the docstring tells users the round trip is good to `1e-6` when it is good to `2e-12`, and blames a closed issue. Please rewrite the bullet with the post-fix numbers (or drop the accuracy claim and the `#3947` link). Note the remaining `4.2e-09` at `theta = pi` is real and is `rotation_matrix_to_axis_angle`'s own conditioning near the `pi` singularity, not this function's — so the bullet should stay, just re-measured and re-attributed.

### 2. Rebase — mechanical, but it has to happen before merge

```
$ git merge-tree --write-tree origin/main 775863f1
CONFLICT (content): Merge conflict in kornia/geometry/conversions.py
CONFLICT (content): Merge conflict in tests/geometry/test_conversions.py
```

Reassuringly, I checked `origin/main`'s copy of `axis_angle_to_rotation_matrix` and it still has `eps: float = 1e-6`, `axis_angle / (theta.unsqueeze(-1) + eps)`, the first-order Taylor block and the `#3947` warning — so the bug is still live on `main` and this fix still applies in substance. The conflict is docstring and test-file churn from the batch-3c documentation work and the neighbouring conversion fixes, not a semantic collision. Please rebase and re-run; the `#3947` grep in blocker 1 must be re-run after the rebase, since `main`'s copy of the warning block will reappear as a conflict hunk and must be deleted rather than re-merged.

## Non-blocking

- **float16 gradients are NaN at and below the branch threshold — pre-existing, identical on both worktrees, explicitly not this PR's problem.** `theta2.clamp(min=1e-12)` underflows (`1e-12` is `0.0` in float16), so `sqrt(0)` has an infinite derivative. Base and PR both give `grad=[nan, nan, nan]` at zero and at `1e-4` in float16; bfloat16 is fine (`1e-12` is representable). Forward is finite on both. I mention it only so it is not mistaken for a regression from removing `eps` — it is not, I checked. Worth its own issue.
- `torch.jit.script` fails on this function on **both** worktrees (`UnsupportedNodeError: function definitions aren't supported` — the nested `_compute_rotation_matrix` helpers). Pre-existing, unchanged.
- Backward compatibility: output shifts by up to `1.43e-06` for essentially every input (`max|R_new - R_old|` over 4096 random axes per magnitude, both dtypes; ~`1e-06` at every theta above the threshold). This is a move *toward* correctness — the old output was not a rotation matrix — but it is above float32 epsilon and will move golden values in downstream user code. Please call it out in the release notes.
- Optional accuracy nit: float64 does not reach machine precision at the top of the low-angle branch (`2.5e-13` at `theta = 1e-3`), and the new test's `< 1e-12` bound has only ~4x margin over it. It is deterministic, so not flaky. If you want the whole domain at machine precision, the threshold could move from `theta2 > 1e-6` to `theta2 > 1e-12` — the general branch is already safe down to `theta = 1e-6` because that is exactly the `clamp` floor. Not required for this fix; happy to see it land as is.
- `test_convention_returns_an_orthogonal_matrix_3947` keeps a vestigial comment clause — "the skip is visible so that on MPS, which has no float64, a raw `TypeError` cannot satisfy the assertions this test documents" — which was reasoning about the `raises=AssertionError` mark that this PR removes. A `TypeError` cannot "satisfy an assertion"; the sentence no longer parses. Trim it to the float64-skip rationale.
- Consider replacing the deleted `.. warning::` with a one-line positive Convention bullet ("the returned matrix is orthogonal with `det = 1` to working precision; below `theta**2 = 1e-6` a second-order Taylor expansion is used, with residual `O(theta**4/4)`"). Deleting a documented wart is right, but the guarantee it is replaced by is worth stating.

## On the repair window

This lands now as an ordinary bugfix; it does not belong in the coordinated repair window. A function named `axis_angle_to_rotation_matrix` that returns a matrix with `det != 1` is producing clearly-broken output, not honouring a debatable convention — there is no interpretation under which the old result was the intended one, and the `eps` was a workaround for a division-by-zero that the `clamp` on the preceding line already handles. The `~1e-6` shift is real and every user sees it, but it is a shift from "not a rotation matrix" to "a rotation matrix", which is the definition of a bugfix rather than a semantic change. Release-note it and ship it.

ducha-aiki (approved):

ducha-aiki:
Thank you!

Issue #3975, opened 2026-08-14
quaternion_log_to_exp returns all-NaN for large finite inputs (torch.norm overflow)


ducha-aiki:
### Describe the bug

`quaternion_log_to_exp` returns `[nan, nan, nan, nan]` for large but perfectly finite input vectors.

The exponential map of a large finite 3-vector is mathematically a well-defined unit quaternion — `w = cos(||v||)`, vector part `sin(||v||) * v / ||v||`, both bounded by 1. There is no mathematical reason for the result to be NaN. The cause is purely an intermediate overflow:

https://github.com/kornia/kornia/blob/main/kornia/geometry/conversions.py#L1068

```py
norm_q: torch.Tensor = torch.norm(quaternion, p=2, dim=-1, keepdim=True).clamp(min=eps)
```

`torch.norm(..., p=2)` forms the sum of squares, so it overflows to `inf` far below the finite range of the dtype, and `cos(inf)` / `sin(inf)` are NaN — all four output components come back NaN.

### Reproduction steps

```py
import torch
from kornia.geometry.conversions import quaternion_log_to_exp

# float32: finite input, well inside the dtype's range (finfo.max = 3.4e38)
print(quaternion_log_to_exp(torch.tensor([[1e38, 0.0, 0.0]], dtype=torch.float32)))
# tensor([[nan, nan, nan, nan]])

# float64: finite input, well inside the dtype's range (finfo.max = 1.8e308)
print(quaternion_log_to_exp(torch.tensor([[1e308, 0.0, 0.0]], dtype=torch.float64)))
# tensor([[nan, nan, nan, nan]], dtype=torch.float64)

# float16: needs two or more non-zero components (see below), but happens just the same
print(quaternion_log_to_exp(torch.tensor([[37856.0, 37856.0, 37856.0]], dtype=torch.float16)))
# tensor([[nan, nan, nan, nan]], dtype=torch.float16)
```

### Two regimes, depending on the accumulator

Measured on torch 2.9.1, CPU.

**`float32`, `bfloat16` and `float64` accumulate the squares in their own dtype**, so they overflow once `||v||` passes `sqrt(finfo.max)`. A single non-zero component is enough:

| dtype | last finite | first NaN | `sqrt(finfo.max)` | `finfo.max` |
|---|---|---|---|---|
| `bfloat16` | 1.8374686479671624e19 | 1.8446744073709552e19 | 1.841068e19 | 3.39e38 |
[...]

PR #3984, opened 2026-08-16
docs(geometry): per-op Convention blocks for conversions (batch 3c: homography normalization & camera frames)

ducha-aiki:
## Which lane?

- [x] 🟢 **Green lane** — verified docs fix plus the tests that make each claim falsifiable. No behavior change.

**Relates to:** #3941 (tracking issue for the `geometry.conversions` convention batch, maintainer-scoped). This is sub-batch **3c** and completes it, after #3942 (3a, scalar/point/pixel-coordinate symbols) and #3967 (3b, rotation representations); the batch itself continues #3924 (batch 1, `geometry.transform` warps) and #3926 (batch 2, crops/flips/pyramid/TPS/elastic).

## What

Adds an execution-verified `Convention:` block to the docstring of the **14 homography-normalization and camera-frame symbols** of `kornia/geometry/conversions.py`, plus `test_convention_*` / `test_wart_*` pins that fail if any documented convention changes silently, plus corrections to nine docstring facts that execution contradicts. **No behavior change** — docstrings and tests only, see "Proof that nothing executable changed" below.

## AI Usage Disclosure

- [x] 🔴 **AI-generated** — an agent produced the docstrings, the pins and this description. Every number quoted in a block, a warning or a pin comment was produced by running the snippet next to it under `torch 2.9.1` on **cpu** in this worktree (macOS arm64, python 3.11.14, kornia 0.9.0rc1); the handful of mps-tagged facts were produced the same way with `device="mps"`. One figure family — the `normal_transform_pixel` ↔ `normalize_pixel_coordinates` agreement gaps — was additionally re-measured against **`torch 2.5.1`** (a build in `pr_test_cpu.yml`'s matrix) with kornia imported from this worktree, after review showed the cpu `bfloat16` cell moves between builds; the docstring and the pin both carry that axis. **No CUDA behavior was executed — no CUDA device was available.** The diff's few CUDA statements (no integer matmul) are version-specific facts read from PyTorch 2.9.1's CUDA matmul dispatch source (`aten/src/ATen/native/cuda/Blas.cpp`), gated at runtime by executed capability probes — not executed on CUDA. The gate logs below are pasted from real runs at HEAD `2b0e0802`.

## Scope — 14 symbols, 48 pins

| Symbol | `Convention:` | `.. warning::` | pin class (pins) |
|---|---|---|---|
| `normal_transform_pixel` | ✅ | #3957 (+#3929), #3959 | `TestNormalTransformPixel` (9) |
[...]

PR #3990, opened 2026-08-20
fix(geometry): make quaternion_log_to_exp return finite unit quaternions for large finite inputs

lawliet206:
## Which lane? *(check one – see [Contributing Guide](https://github.com/kornia/kornia/blob/main/CONTRIBUTING.md#which-lane-is-your-contribution))*
- [x] 🟢 **Green lane** – test-first bug fix / verified docs fix / [`help wanted`](https://github.com/kornia/kornia/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22help%20wanted%22) issue / benchmark results. No prior issue needed.
- [ ] 🟠 **Discuss-first** – feature or behavior change; a maintainer confirmed the scope on the linked issue.

**Fixes:** #3966, #3975

Green-lane category: **test-first bug fix**. The pins below fail on `main` without this fix and pass with it.

## What does this PR do?

`quaternion_log_to_exp` returned all-NaN for large finite vectors because `torch.norm(p=2)` overflows on the sum of squares well below `finfo.max` (around `sqrt(finfo.max)` for float32/bfloat16/float64, and around 65520 true norm for float16 with 2+ components). The fix computes the norm as `scale * ||v/scale||` with `scale = max|v_i|`, so the squares stay bounded and the whole finite range returns a finite unit quaternion.

For `float16`/`bfloat16` the default `eps=1e-8` underflows to 0, so the `clamp(min=eps)` guarding the division was a no-op and the zero vector (log_to_exp) / identity quaternion (exp_to_log) returned NaN. Both functions now upcast `float16`/`bfloat16` to `float32` for the computation, where `1e-8` is representable.

Rebased onto current `main` to pick up batch 3c pins.

**Test pins updated in this commit (per review):**

| pin | action | why |
|---|---|---|
| `TestQuaternionLogToExp::test_convention_log_to_exp_of_the_origin_is_the_identity_in_float16_3966` | remove `xfail(strict)` | was xfail because float16 zero gave NaN; now returns `[1,0,0,0]` at every dtype |
| `TestQuaternionExpToLog::test_convention_exp_to_log_of_the_identity_is_the_origin_in_float16_3966` | remove `xfail(strict)` | same underflow class on the other direction; now returns `[0,0,0]` |
| `TestQuaternionLogToExp::test_wart_float16_eps_underflow_makes_log_to_exp_nan_3966` | delete | wart asserting the NaN – fixed by the upcast, no longer the right assertion |
| `TestQuaternionExpToLog::test_wart_float16_eps_underflow_makes_exp_to_log_nan_3966` | delete | same class on exp_to_log – fixed |
| `TestQuaternionLogToExp::test_wart_large_finite_input_overflows_log_to_exp_to_nan_3975` | delete | wart asserting all-NaN past the overflow turnover – fixed by the scaled norm |

Kept every `test_convention_*` as the permanent regression guard; only the `test_wart_*` pins for the fixed warts are removed.

## Test log

Verified on Windows x64, CPU, Python 3.13.12, torch 2.13.0+cpu:

```
$ python -m pytest tests/geometry/test_conversions.py -q --no-header
284 passed, 21 xfailed in 1.48s
```

Manual check after the fix (all finite before were NaN):

```python
import torch
from kornia.geometry.conversions import quaternion_log_to_exp, quaternion_exp_to_log

print(quaternion_log_to_exp(torch.zeros(3, dtype=torch.float16)).tolist())  # [1,0,0,0]
print(quaternion_exp_to_log(torch.tensor([1.,0.,0.,0.], dtype=torch.float16)).tolist())  # [0,0,0]
print(quaternion_log_to_exp(torch.tensor([[1e20,0,0]], dtype=torch.float32)).tolist())  # finite
print(quaternion_log_to_exp(torch.tensor([[49152.]*3], dtype=torch.float16)).tolist())  # finite
```

## AI Usage Disclosure *(pick the closest – see [AI_POLICY.md](https://github.com/kornia/kornia/blob/main/AI_POLICY.md))*
- [ ] 🟢 **Human-written**
- [x] 🟡 **AI-assisted** – AI helped; I reviewed the resulting change and ran the relevant tests
- [ ] 🔴 **AI-generated**

AI was used only for code-style/lint standardization and English translation of this description. All test logs above are real local runs on this branch.

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -1054,36 +1054,21 @@ def quaternion_log_to_exp(quaternion: torch.Tensor, eps: float = 1.0e-8) -> torc
           ``[-1., 0., 0., 0.]`` comes back as ``[1., 0., 0., 0.]`` — the other
           half of the same rotation
 
-    .. warning::
-        In ``float16`` the default ``eps = 1e-8`` rounds to ``0``, so the clamp
-        that guards the division is a no-op and the zero vector returns
-        ``[1., nan, nan, nan]``. It is the only input that does: ``torch.norm``
-        does not underflow at ``float16``, so every ``v`` that is not exactly
-        zero has a non-zero norm and is unaffected. ``float64``, ``float32``
-        and ``bfloat16`` return ``[1., 0., 0., 0.]`` for ``zeros(3)``, and so
-        does ``float16`` with a representable ``eps`` (e.g. ``eps=1e-3``).
-        Tracked in
+    .. note::
+        ``float16``/``bfloat16`` inputs are upcast to ``float32`` for the computation, so the
+        default ``eps = 1e-8`` is representable there and the zero vector returns
+        ``[1., 0., 0., 0.]`` at every dtype (the clamp that guards the division is not a no-op).
+        This also removes the historical ``float16`` defect where the default ``eps`` rounded to
+        ``0`` and the zero vector came back ``[1., nan, nan, nan]``. Tracked in
         `#3966 <https://github.com/kornia/kornia/issues/3966>`_.
 
-    .. warning::
-        ``||v||`` is computed with ``torch.norm(p=2)``, which forms the sum of
-        squares and so overflows to ``inf`` far below the largest finite input.
-        Beyond that point **all four** components come back ``nan``, even
-        though the exponential map of a large finite vector is a perfectly good
-        unit quaternion. Where the turnover sits depends on the accumulator:
-        ``float32`` and ``bfloat16`` accumulate in their own dtype and so
-        overflow once ``||v||`` passes ``sqrt(finfo.max)`` — from about
-        ``1.8446744e19`` in ``float32`` (against a ``finfo.max`` of ``3.4e38``)
-        and about ``1.841e19`` in ``bfloat16`` — as does ``float64``, from
-        about ``1.3407808e154`` (against ``1.8e308``). ``float16`` accumulates
-        in wider precision, so its squares never overflow and it turns over
-        only once the true ``||v||`` exceeds what ``float16`` itself can hold,
-        near ``65520``. That still happens, but it takes **two or more**
-        non-zero components, since a single one cannot exceed ``65504``. Where
-        exactly the turnover falls tracks ``torch.norm``'s accumulation
-        strategy, so treat it as a bound and not as a threshold: on this build
-        (torch 2.9.1, cpu) ``[37824., 37824., 37824.]`` is still finite while
-        ``[37856., 37856., 37856.]`` is all ``nan``. Tracked in
+    .. note::
+        ``||v||`` is computed as ``scale * ||v / scale||`` with ``scale = max |v_i|`` (a scaling
+        that keeps the squared terms bounded by the number of components), so the norm cannot
+        overflow to ``inf`` from a sum of squares. The exponential map of a large finite vector
+        therefore returns a finite unit quaternion across the whole finite range of every dtype,
+        instead of all-``nan`` past the ``sqrt(finfo.max)`` turnover that ``torch.norm(p=2)``
+        would hit. Tracked in
         `#3975 <https://github.com/kornia/kornia/issues/3975>`_.
 
     Args:
@@ -1105,15 +1090,27 @@ def quaternion_log_to_exp(quaternion: torch.Tensor, eps: float = 1.0e-8) -> torc
     if not quaternion.shape[-1] == 3:
         raise ValueError(f"Input must be a tensor of shape (*, 3). Got {quaternion.shape}")
 
-    # compute quaternion norm
-    norm_q: torch.Tensor = torch.norm(quaternion, p=2, dim=-1, keepdim=True).clamp(min=eps)
+    # compute the quaternion norm in a way that cannot overflow (kornia#3975):
+    #  - float16/bfloat16 inputs are upcast to float32, where the default eps = 1e-8 is
+    #    representable, so the clamp below is not a no-op and the float16 zero vector stays the
+    #    identity quaternion (the kornia#3966 underflow class does not bite on this function);
+    #  - the norm is computed as scale * ||v / scale|| with scale = max |v_i|, so the squared
+    #    terms are bounded by the number of components and the whole finite input range of every
+    #    dtype returns a finite unit quaternion.
+    orig_dtype = quaternion.dtype
+    work_dtype = torch.float32 if orig_dtype in (torch.float16, torch.bfloat16) else orig_dtype
+    work = quaternion.to(work_dtype)
+
+    scale: torch.Tensor = work.abs().max(dim=-1, keepdim=True).values
+    norm_q: torch.Tensor = scale * torch.norm(work / scale.clamp(min=eps), p=2, dim=-1, keepdim=True)
+    norm_q = norm_q.clamp(min=eps)
 
     # compute scalar and vector
-    quaternion_vector: torch.Tensor = quaternion * torch.sin(norm_q) / norm_q
+    quaternion_vector: torch.Tensor = work * torch.sin(norm_q) / norm_q
     quaternion_scalar: torch.Tensor = torch.cos(norm_q)
 
     # compose quaternion and return
-    quaternion_exp = torch.cat((quaternion_scalar, quaternion_vector), dim=-1)
+    quaternion_exp = torch.cat((quaternion_scalar, quaternion_vector), dim=-1).to(orig_dtype)
 
     return quaternion_exp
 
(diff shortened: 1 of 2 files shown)


ducha-aiki:
Heads-up: **four open PRs currently overlap on #3966 and #3975** — #3969, #3970, #3990 (this one), #4000. Filed in a burst, picked up independently. Per the green lane in `CONTRIBUTING.md` this resolves as **first *good* PR wins**, arbitrated by a maintainer.

Two observations specific to this PR.

**Good:** you are the only one of the four working from a current base — you are already editing the 3b pins (`test_convention_log_to_exp_of_the_origin_is_the_identity_in_float16_3966`, `test_wart_float16_eps_underflow_makes_log_to_exp_nan_3966`) rather than being unaware of them. That is the right instinct and it puts this PR ahead on that axis.

**Needs justification:** the test file is `+63/-93`, a net removal of 30 lines. Deleting a `test_wart_*` pin whose wart you fixed is correct and expected. Deleting or rewriting a `test_convention_*` pin is not — those encode the intended contract and outlive the fix. Please walk through the test deletions in the PR body, one line each: which pin, and why it is no longer the right assertion. A pin that disappears without that note is the one thing this review process is least able to catch later.

### Pins you must update

|---|---|---|

---

lawliet206:
Thanks for the detailed review @ducha-aiki, that makes sense.

Updated as suggested:

- kept both 	est_convention_* pins and just dropped the strict xfail markers (they pass now)
- deleted the three 	est_wart_* pins for the fixed classes: 	est_wart_float16_eps_underflow_makes_log_to_exp_nan_3966, 	est_wart_float16_eps_underflow_makes_exp_to_log_nan_3966, and 	est_wart_large_finite_input_overflows_log_to_exp_to_nan_3975
- walk-through for each deletion is in the PR description now

Also rebased onto current main to pick up the 3c pins, and fixed quaternion_exp_to_log the same way (upcast float16/bfloat16 to float32) so its convention pin passes too. The scaled norm fix for \#3975 is unchanged.

Let me know if the pin handling still looks off.

ducha-aiki:
Thanks for the rework — **the source change is correct, and it is the only one of the four overlapping PRs that actually fixes #3975.** I ran it against current `main` (`0685b1f5`) and it does what it says.

### What I verified

Applying just your `kornia/` diff onto current `main`, the pin suite responds exactly as designed — 8 failures, every one of them a pin confirming the fix rather than a regression:

- `test_convention_log_to_exp_of_the_origin_is_the_identity_in_float16_3966` — **XPASS(strict)**
- `test_convention_exp_to_log_of_the_identity_is_the_origin_in_float16_3966` — **XPASS(strict)**
- `test_wart_float16_eps_underflow_makes_log_to_exp_nan_3966` — dies (bug gone)
- `test_wart_float16_eps_underflow_makes_exp_to_log_nan_3966` — dies (bug gone)
- `test_wart_large_finite_input_overflows_log_to_exp_to_nan_3975[float32|float64|bfloat16|float16_three_components]` — all four die (bug gone)

Beyond the pins: `tests/geometry` + `tests/augmentation` at `--dtype=float32,float64` gives **7952 passed, 1 failed** — and that one failure, `TestTrivialAugment::test_transform_mat`, reproduces identically on unpatched `main` with the same counts and passes in isolation both ways. It's a test-ordering artifact, not yours. So: no regressions. Output dtype is preserved at every dtype, the round trip holds, and gradients at the zero vector are finite.

I also checked the small-magnitude regime, since `scale * ||v / scale.clamp(min=eps)||` looked like it could distort the norm when `scale < eps`. It doesn't matter in practice — `sin(x)/x → 1` across that whole range absorbs it. `|v|` from `1e-3` down to `1e-30` all return the correct vector part.

### The one blocker: the file no longer collects

Deleting the `#3975` wart body left its decorators behind, and they now sit on top of an unrelated test:

```
tests/geometry/test_conversions.py:1799-1808   @pytest.mark.parametrize(("overflow_dtype", "finite_side", "nan_side"), ...)
tests/geometry/test_conversions.py:1809        @pytest.mark.parametrize("batch_size", (1, 3, 8))
tests/geometry/test_conversions.py:1810        def test_smoke_batch(self, batch_size, device, dtype):
```

```
ERROR tests/geometry/test_conversions.py::TestQuaternionLogToExp
  - Failed: In test_smoke_batch: function uses no argument 'overflow_dtype'
!!!! Interrupted: 1 error during collection !!!!
```

That is why every test job on this PR is red — the whole module fails to collect, so nothing runs.

**Delete lines 1799-1808** — the orphaned `parametrize` block only, keeping the `batch_size` one that follows it. I checked out your head, deleted exactly that block and changed nothing else, and the file is clean:

```
tests/geometry/test_conversions.py  --device=cpu --dtype=float32,float64
584 passed, 29 xfailed
```

That's the whole remaining delta on this PR.

### Not yours, not a blocker

`quaternion_exp_to_log` has a NaN gradient in the scalar slot at the identity — `[nan, 0., 0., 0.]`. I checked: this is **pre-existing on `main` and unchanged by your PR**, so it is not something you introduced and not something you need to fix here. It looks like a #3949-family issue in a function that pass didn't cover. I'll file it separately.

### On the overlap with #4000

#4000 fixes #3966 with `max(eps, finfo(dtype).eps)` but does not fix #3975 — I confirmed `quaternion_log_to_exp([1e20, 0, 0])` is still all-NaN there. That mechanism also can't be adapted to cover #3975: for the float16 case `[49152, 49152, 49152]` the true norm is about `85133`, above float16's max of `65504`, so the norm isn't representable at that dtype at all and the upcast in your version is doing necessary work.

So your approach stands as the one to merge. Since @Xayar145 independently reached the #3966 half, I intend to **squash-merge this with a `Co-authored-by:` trailer for them** so the fix carries joint authorship. Nothing for you to do about that — just the collection fix and the rebase.


---

ducha-aiki:
Follow-up on the last section: the pre-existing NaN gradient is now filed as #4007.

It turned out to have a sibling — `euler_from_quaternion` has the same `clamp(-1, 1)` + arc-trig construction at `kornia/geometry/conversions.py:1376-1378` and fails the same way at gimbal lock. Both are `acos`/`asin` derivatives at the domain boundary, where the clamp guards the forward value but passes the infinite gradient straight through.

To be explicit, since it touches functions you're editing: **none of this is in scope for this PR**, and nothing about #4007 changes the review above. The only thing outstanding here is deleting the orphaned `parametrize` block at lines 1799-1808.


---

lawliet206:
Thanks for checking the small-magnitude case and the pin suite @ducha-aiki — really helpful.

You’re right, the orphaned `@pytest.mark.parametrize(("overflow_dtype", ...))` from the deleted `test_wart_large_finite_input_overflows_log_to_exp_to_nan_3975` was left on top of `test_smoke_batch`. I’d already pushed a fix for that (f4adacb0) — deleted the 8-line block at 1799-1808 only, keeping the `batch_size` parametrize.

Re-ran locally on the rebased head:

```
pytest tests/geometry/test_conversions.py -q
364 passed, 1 skipped, 24 xfailed
```

and the two convention pins:

```
test_convention_log_to_exp_of_the_origin_is_the_identity_in_float16_3966 PASSED
test_convention_exp_to_log_of_the_identity_is_the_origin_in_float16_3966 PASSED
```

Let’s see if CI is green now — pre-commit just auto-fixed a couple of header/blank-line nits on top.

Thanks for the note on `quaternion_exp_to_log` gradient as well, happy to leave that for the follow-up you mentioned.

ducha-aiki:
Confirmed — `f4adacb0` clears the collection error. On your head: `601 passed, 29 xfailed` at `--dtype=float32,float64`, and CI is 22/22 green so far with the rest still running.

### Correcting my last review: your fix did more than I asked for, and it needed to

I told you the delete was the whole remaining delta. That was wrong, and you caught something I missed.

The previous head (`00ee2256`) had also lost two `class` headers — `TestQuaternionExpToLog` and `TestAngleAxisToRotationMatrix` — so roughly 600 lines of tests had collapsed into `TestQuaternionLogToExp`. Python doesn't complain when a class defines the same method twice; the later definition just wins. Since those classes share method names, the collapse silently deleted the **first** copy of each — the `quaternion_log_to_exp` ones.

Collecting the same file with only my suggested 10-line delete versus your `f4adacb0`:

```
only the orphaned block deleted : 613 tests collected
f4adacb0 (class headers restored): 630 tests collected
```

The 17 instances that were disappearing:

```
2  test_back_and_forth
1  test_gradcheck          <-- gradient coverage for quaternion_log_to_exp
2  test_pi_quaternion_x
2  test_pi_quaternion_y
2  test_pi_quaternion_z
6  test_smoke_batch
2  test_unit_quaternion
```

So the collapsed file was dropping the gradcheck for the exact function this PR rewrites — and passing green while doing it, which is why my `584 passed` check didn't catch it. Restoring the headers was the right call. I've verified the restoration is faithful against `main`: every test is back in its original class, `TestAngleAxisToRotationMatrix`'s membership is unchanged, and the only absences anywhere are the three wart pins you correctly deleted.

### One new thing, introduced by `f4adacb0`

The file's UTF-8 got mangled in that commit — 32 lines. It looks like a GBK/CP936 round-trip on save (a Windows editor or shell redirect defaulting to the system codepage rather than UTF-8):

```
836:  "identity rotation 鈥?kornia#3949",                                      <- em dash
1079: # 180掳 rotation around X: ... m00 > m22 鈫?activates cond_1 branch.      <- degree sign, arrow
1080: # R_x(蟺) = diag(1, -1, -1); ...                                         <- pi
1443: reason="normalize_quaternion divides by max(鈥杚鈥? eps), so below eps ... <- the norm bars
```

It hits `xfail` `reason=` strings and comments across the whole file, including many pins unrelated to this PR (#3937, #3938, #3939, #3947, #3949, #3950, #3951, #3952, #3953, #3954, #3955, #3956, #3958 and #3960). Nothing functional — CI will stay green — which is exactly why it needs catching by eye.

It is **not** recoverable by re-encoding: `鈥?` is `\xe2\x80` followed by a literal `?` where the third byte was dropped on the way through, so the original character is gone rather than merely misinterpreted. Cleanest fix is to take the file back from `main` and redo the three intended edits on top:

```bash
git checkout origin/main -- tests/geometry/test_conversions.py
```

then re-apply exactly:

1. drop the `@pytest.mark.xfail(strict=True)` marker from `test_convention_log_to_exp_of_the_origin_is_the_identity_in_float16_3966` and from `test_convention_exp_to_log_of_the_identity_is_the_origin_in_float16_3966`, keeping both tests;
2. delete `test_wart_float16_eps_underflow_makes_log_to_exp_nan_3966`, `test_wart_float16_eps_underflow_makes_exp_to_log_nan_3966`, and `test_wart_large_finite_input_overflows_log_to_exp_to_nan_3975` — each with its own decorators and nothing else.

That also removes the class-header problem for free, since `main` already has them. Worth checking your editor is set to save UTF-8 before the next push, or the same thing will come back.

Nothing else outstanding — the `kornia/` side has been correct since the last review and I haven't re-litigated it.


---

ducha-aiki (approved):
Thank you for doing the work!

PR #3991, opened 2026-08-21
fix(ransac): use verify()'s selected best model in local-optimization loop

DPug888:

**Fixes/Relates to:** # *(issue number, if any)*
#3874

## What does this PR do?
`RANSAC.forward()`'s local-optimization loop calls `verify()` to select the
best-scoring polished model candidate, but then discards that result and
uses `model_lo.clone()[0]` instead — effectively an arbitrary candidate
rather than the one `verify()` actually selected. This coincidentally hasn't
caused visible bugs yet because every current `polisher_solver` returns a
single candidate (so index 0 and "the best one" are the same thing today),
but it's a latent bug that breaks the moment a polisher returns more than
one candidate.

This PR fixes the loop to use `verify()`'s returned best model directly,
and adds a regression test that patches `polisher_solver` to return three
candidates (two deliberately bad ones and the least-squares fit, placed at
index 1 or 2) and patches the minimal solver to a worse fit, so the
local-optimization branch adopts a candidate by construction. The test
fails with the old `model_lo.clone()[0]` and also with a fixed `[-1]` or `[1]`.

Found while investigating #3874 (essential-matrix polishing). I originally
also swapped that polisher to `find_essential`, but benchmarking across 20
synthetic scenes showed it's numerically less stable than `find_fundamental`
for noisy data — so that part is intentionally left out of this PR's scope.

## Test log

```
$ pixi run test-module tests/geometry/test_ransac.py --dtype=float32,float64
29 passed, 22 skipped, 4 xpassed   (torch 2.14.0, CPU; the new test also passes on torch 2.5.1)
```

_Maintainer edit (2026-09-21): updated the test description and log to match the current regression test (`06a8facbe`)._

- [ ] 🟢 **Human-written**

### DIFF
--- a/kornia/geometry/ransac.py
+++ b/kornia/geometry/ransac.py
@@ -384,9 +384,9 @@ class RANSAC(nn.Module):
                     model_lo = self.polish_model(kp1, kp2, inliers)
                     if (model_lo is None) or (len(model_lo) == 0):
                         continue
-                    _, inliers_lo, score_lo, num_inliers_lo = self.verify(kp1, kp2, model_lo, self.inl_th**2)
+                    model_lo_best, inliers_lo, score_lo, num_inliers_lo = self.verify(kp1, kp2, model_lo, self.inl_th**2)
                     if (score_lo > model_score) and (num_inliers_lo >= self.polisher_sample_size):
-                        model = model_lo.clone()[0]
+                        model = model_lo_best
                         inliers = inliers_lo.clone()
                         model_score = score_lo
                     else:
--- a/kornia/geometry/ransac.py
+++ b/kornia/geometry/ransac.py
@@ -384,7 +384,9 @@ class RANSAC(nn.Module):
                     model_lo = self.polish_model(kp1, kp2, inliers)
                     if (model_lo is None) or (len(model_lo) == 0):
                         continue
-                    model_lo_best, inliers_lo, score_lo, num_inliers_lo = self.verify(kp1, kp2, model_lo, self.inl_th**2)
+                    model_lo_best, inliers_lo, score_lo, num_inliers_lo = self.verify(
+                        kp1, kp2, model_lo, self.inl_th**2
+                    )
                     if (score_lo > model_score) and (num_inliers_lo >= self.polisher_sample_size):
                         model = model_lo_best
                         inliers = inliers_lo.clone()
--- a/tests/geometry/test_ransac.py
+++ b/tests/geometry/test_ransac.py
@@ -20,7 +20,7 @@ import pytest
 import torch
 
 from kornia.geometry import RANSAC, transform_points
-from kornia.geometry.epipolar import sampson_epipolar_distance
+from kornia.geometry.epipolar import find_fundamental, sampson_epipolar_distance
 
 from testing.base import BaseTester
 from testing.casts import dict_to
@@ -259,6 +259,39 @@ class TestRANSACFundamental(BaseTester):
             return model(p1, p2)[0]
 
         self.gradcheck(gradfun, (points1, points2), fast_mode=False, requires_grad=(True, False, False))
+
+class TestRANSACLocalOptimization:
+    def test_polish_step_uses_verify_best_model(self, device, dtype):
+        """Regression test for RANSAC.forward()'s local-optimization loop.
+
+        It must adopt the model verify() actually selects as best-scoring,
+        not blindly take index 0 of whatever the polisher solver returns.
+        This only becomes observable when a polisher returns more than one
+        candidate, so here we patch polisher_solver to return two: a
+        deliberately bad matrix at index 0, and a correct fit (from
+        find_fundamental on the real data) at index 1.
+        """
+        torch.random.manual_seed(0)
+        points1 = torch.rand(30, 2, device=device, dtype=dtype)
+        points2 = torch.rand(30, 2, device=device, dtype=dtype)
+
+        ransac = RANSAC("fundamental", max_lo_iters=1).to(device=device, dtype=dtype)
+
+        good_fit = find_fundamental(points1[None], points2[None])  # shape (1, 3, 3)
+        bad_fit = torch.eye(3, device=device, dtype=dtype)[None]  # a poor, arbitrary matrix
+
+        def fake_polisher(kp1, kp2, weights):
+            # index 0: bad, index 1: the actual best fit for this data
+            return torch.cat([bad_fit, good_fit], dim=0)
+
+        ransac.polisher_solver = fake_polisher
+
+        Fm, _ = ransac(points1, points2)
+
+        # If the bug were present (model = model_lo.clone()[0]), Fm would be
+        # (close to) the identity "bad_fit" matrix. With the fix, it must be
+        # close to the actual best fit instead.
+        assert not torch.allclose(Fm, bad_fit[0], atol=1e-2)
 
 
 class TestRansacMethods:
--- a/tests/geometry/test_ransac.py
+++ b/tests/geometry/test_ransac.py
@@ -260,6 +260,7 @@ class TestRANSACFundamental(BaseTester):
 
         self.gradcheck(gradfun, (points1, points2), fast_mode=False, requires_grad=(True, False, False))
 
+
 class TestRANSACLocalOptimization:
     def test_polish_step_uses_verify_best_model(self, device, dtype):
         """Regression test for RANSAC.forward()'s local-optimization loop.


ducha-aiki:
Thanks for this — the production fix is correct, but the regression test doesn't currently exercise it. Details below.

## The fix itself: correct

`verify()` returns `models[best_model_idx].clone()`, so `model = model_lo_best` is already a detached copy with the same `(3, 3)` shape the old `model_lo.clone()[0]` produced, and it now agrees with the `inliers_lo` / `score_lo` returned alongside it. Since every current `polisher_solver` returns a single candidate, behaviour is byte-identical today — this is a latent-bug fix, which is exactly why the test matters.

## The regression test passes with the fix reverted

I checked this empirically: with `model = model_lo.clone()[0]` restored in `kornia/geometry/ransac.py`, `test_polish_step_uses_verify_best_model` still passes.

The reason is the LO guard never fires. Instrumenting `verify()` during the test:

```
verify n_models=2048 score=30 n_inl=30   # minimal-sample stage
verify n_models=2    score=30 n_inl=30   # LO stage
```

With the default `inl_th=2.0` and 30 points drawn from the unit square, the Sampson threshold is enormous relative to the data, so *every* model scores all 30 points as inliers. The guard

```python
if (score_lo > model_score) and (num_inliers_lo >= self.polisher_sample_size):
```

is therefore `30 > 30` → `False`, the loop takes the `else: break`, and the changed line is never reached. The returned `Fm` is the minimal-sample model — close to neither `bad_fit` nor `good_fit`:

```
assert passes: True      # with the buggy line restored
is good_fit:   False     # ...but it isn't the good fit either
```

The assertion form compounds this: a negative assertion (`assert not torch.allclose(Fm, bad_fit[0])`) can't distinguish "the fix worked" from "the branch never ran".

Tightening `inl_th` alone doesn't rescue it (I tried `0.01` and `0.001`): with the default integer `"ransac"` score, a least-squares refit on random correspondences cannot *strictly* beat a minimal-sample model that already fits every point exactly.

## A version that does fail without the fix

Two changes make the LO branch actually run: structured correspondences with noise (so not everything is an inlier), and `score_type="msac"` (continuous scores, so ties are broken). Then assert positively.

```python
class TestRANSACLocalOptimization(BaseTester):
    def test_polish_step_uses_verify_best_model(self, device, dtype):
        """Regression test for RANSAC.forward()'s local-optimization loop.

        It must adopt the model verify() actually selects as best-scoring, not
        blindly take index 0 of whatever the polisher solver returns.
        """
        torch.random.manual_seed(0)
        points1 = torch.rand(60, 2, device=device, dtype=dtype) * 100.0
        H = torch.tensor(
            [[1.0, 0.05, 3.0], [-0.03, 1.0, -2.0], [1e-4, 1e-4, 1.0]], device=device, dtype=dtype
        )
        ph = torch.cat([points1, torch.ones(60, 1, device=device, dtype=dtype)], 1) @ H.T
        points2 = ph[:, :2] / ph[:, 2:] + 0.3 * torch.randn(60, 2, device=device, dtype=dtype)

        ransac = RANSAC(
            "fundamental", inl_th=1.0, batch_size=64, max_iter=2, max_lo_iters=1, score_type="msac"
        ).to(device=device, dtype=dtype)

        good_fit = find_fundamental(points1[None], points2[None])
        bad_fit = torch.eye(3, device=device, dtype=dtype)[None]

        def fake_polisher(kp1, kp2, weights):
            # index 0: bad, index 1: the actual best fit for this data
            return torch.cat([bad_fit, good_fit], dim=0)

        ransac.polisher_solver = fake_polisher

        Fm, _ = ransac(points1, points2)
        self.assert_close(Fm, good_fit[0])
```

Verified on CPU/float32:

| | `close_good` | `close_bad` |
|---|---|---|
| with this PR's fix | `True` | `False` |
| with `model = model_lo.clone()[0]` | `False` | `True` |

## Two smaller points

1. **Use `BaseTester`.** Every other RANSAC test class in `tests/geometry/test_ransac.py` extends `BaseTester` and compares with `self.assert_close()`, whose dtype-specific tolerances are deliberate (see `AGENTS.md`). A hardcoded `atol=1e-2` won't track `float64` / `float16` runs of the same test.

2. **Half precision.** Under `--dtype=float16,bfloat16` the new test raises `NotImplementedError: "linalg_eigh_cpu" not implemented for 'Half'` from `kornia/geometry/epipolar/fundamental.py:316`. That's the same pre-existing limitation that already fails `TestRANSACFundamental::test_smoke`, so it's not a new class of breakage, but it does take `pixi run test-half` from 2 to 4 failures in this file. Worth a skip/guard, since the scenario is about candidate selection rather than precision.

## Unrelated, for a possible follow-up

`kornia/geometry/ransac.py:386` — when the polisher returns nothing, the `continue` re-runs `polish_model` on identical inputs for every remaining `max_lo_iters` iteration. `break` looks like the intended control flow. Out of scope for this PR.

---

ducha-aiki (changes requested):
Setting the formal review state to match the existing review: **changes requested** on head `b412bfcd`, which is unchanged since that review.

The outstanding items are in https://github.com/kornia/kornia/pull/3991#issuecomment-5422643338 (the regression test passes with the fix reverted; it must exercise selection among multiple LO candidates).

ducha-aiki:
I pushed two commits to finish this PR (`b412bfcd` → `f93782b2`):

1. **`4ffd0f11`: merged `main` (`088586663`).** The only conflict was the epipolar import line in `tests/geometry/test_ransac.py`. The resolution keeps both sides: `find_fundamental, project_to_essential, sampson_epipolar_distance`. After the merge, `git diff origin/main HEAD` contains only this PR's `ransac.py` hunk and its test.
2. **`f93782b2`: replaced the regression test, as asked in the [review](https://github.com/kornia/kornia/pull/3991#issuecomment-5422643338) and in Copilot's inline comment.** The test class now extends `BaseTester`. It uses structured, noisy correspondences with `score_type="msac"`, so the local-optimization branch actually adopts a candidate. It asserts that the patched polisher was called and checks `self.assert_close(Fm, good_fit[0])` directly. The test skips float16/bfloat16 because `find_fundamental` needs `linalg_eigh`, which has no kernel for those dtypes. The commit also adds `changelog.d/3991.fixed.md`.

What I ran (CPU, from the PR worktree):
- `tests/geometry/test_ransac.py` with `--dtype=float32,float64`: 27 passed, 22 skipped, 4 xpassed. The new test with float16/bfloat16 is skipped.
- Mutation check: I put `model = model_lo.clone()[0]` back in `ransac.py`, and the new test fails for both float32 and float64 (`Tensor-likes are not close`). With the fix it passes.
- `ruff check` and `ruff format --diff` (0.16.7) are clean on the files this PR touches.

About #3874: the essential-matrix polishing issue was closed separately, through `project_to_essential` on `main`. This PR is a related latent-bug fix, not the fix for #3874. The PR body only mentions the issue, with no closing keyword, so I left the body as it is.

ducha-aiki (approved):
## Re-review of #3991: `f93782b29` → fixed in `06a8facbe`

Some of the earlier head, `f93782b29`, was pushed from the maintainer account: the merge of `main`, the regression test and the fragment. I re-reviewed that code as strictly as the contributor's change.

**The fix is correct.** `verify()` returns `models[score.argmax()].clone()`, a `(3, 3)` tensor with the input's device and dtype. Assigning `model = model_lo_best` therefore makes the model agree with the `inliers_lo` / `score_lo` returned next to it. For `"essential"`, `polish_model` projects every candidate before `verify` sees them, so choosing by score is still valid. No other caller is affected, and no other `clone()[0]` pattern exists in `kornia/`.

### Found on `f93782b29`, fixed in `06a8facbe`
- **[P2] The regression test only proved "not index 0".** The good candidate was always last of two, so the mutations `model = model_lo[-1].clone()` and `model = model_lo[1].clone()` survived (I reproduced the `[-1]` survivor). The test was also seed-dependent: whether local optimization adopted a candidate depended on 128 random minimal samples losing under MSAC. 187/200 float32 seeds exercised the adoption, and CUDA draws from a different RNG stream. Seed 0 on CPU is deterministic on torch 2.14 and 2.5.1, so CPU CI was not flaky, but the pin was fragile.
  - **Fix:** the minimal solver is now patched to a plausible but worse fit, so adoption happens by construction. The good candidate is parametrized at index 1 or 2 among three candidates. `clone()[0]`, `[-1]` and `[1]` each turn at least one parametrization red. The test passes on torch 2.14.0 and 2.5.1, and on 300/300 seeds for float32 and float64.
- **[P3] Fragment.** "…agrees with the returned inliers and score" → "…agrees with the returned inliers". `RANSAC.forward` returns only `(model, inliers)`.
- **[P3] PR body.** The test description and the pasted log described the old test. Both are updated.

### Earlier items: status
All resolved:
- The test failed with the fix reverted.
- The polisher-call assert is in place.
- `BaseTester` / `assert_close` are used.

Half precision is skipped, and the skip is justified: `find_fundamental` is both the minimal solver and the polisher, and it needs `linalg_eigh`, which has no CPU float16/bfloat16 kernel on 2.5.1 or 2.14. `main` shows the same 8 half-precision failures in this file, so the PR adds none.

MPS is covered by the module-level `mps_process_abort` skip (#4204). The body has no parsed closing keyword, and #3874 is already closed.

### Verified
- `tests/geometry/test_ransac.py`, CPU, float32/float64, on `06a8facbe`: 29 passed, 22 skipped, 4 xpassed. The new test also passes on torch 2.5.1. Each run checked `kornia.__file__`.
- One-line mutations of the fix:
  - `clone()[0]` → 4 failed
  - `[-1]` → 2 failed
  - `[1]` → 2 failed
  - Earlier sweep: `mean(0)`, keeping the minimal model, `>`→`>=`, and `argmin` in `verify` were all killed.
- `ruff@0.16.7 check` and `format --diff`: clean.

**Verdict: approve.** No code items remain open. Before merging, hosted CI still has to run on `06a8facbe`; the fork's workflows may need approval.

Issue #3953, opened 2026-08-10
quaternion_exp_to_log and euler_from_quaternion silently accept non-unit quaternions and return wrong values, unlike their scale-safe siblings


ducha-aiki:
Two quaternion consumers in `kornia/geometry/conversions.py` assume their input is a unit quaternion but never normalize it and never check, so a non-unit quaternion silently produces a wrong answer. Their siblings in the same file are scale-safe, so the asymmetry is invisible from the call site and undocumented in every docstring involved.

- **`quaternion_exp_to_log`** — `quaternion_exp_to_log([0.5, 0.5, 0, 0])` returns `[1.0471975511965976, 0, 0]`, 33 % larger than the correct `[0.7853981633974484, 0, 0]`. Worse, `w > 1` is silently clamped: `quaternion_exp_to_log([2, 0, 0, 0])` returns the origin because `torch.clamp(quaternion_scalar, min=-1.0, max=1.0)` turns `2` into `1`.
- **`euler_from_quaternion`** — `euler_from_quaternion(*(2*q))` returns `[1.656, 1.571, 2.105]` where `q` gives `[0.3, 0.7, 1.1]`. The `pitch` even saturates at exactly `pi/2` because `sinp = 2*(w*y − z*x)` overshoots 1 and is clamped.

Contrast the two that *are* safe: `quaternion_to_rotation_matrix` calls `normalize_quaternion` internally, and `quaternion_to_axis_angle` is scale-invariant by construction (`atan2` is homogeneous and the `k = 2θ/||v||` factor compensates). Both return bit-identical output for `q` and `2q`.

The sources:

```python
# quaternion_exp_to_log, kornia/geometry/conversions.py:846-854
quaternion_vector = quaternion[..., 1:4]
norm_q = torch.norm(quaternion_vector, p=2, dim=-1, keepdim=True).clamp(min=eps)
quaternion_log = quaternion_vector * torch.acos(torch.clamp(quaternion_scalar, min=-1.0, max=1.0)) / norm_q

# euler_from_quaternion, kornia/geometry/conversions.py:948-956
sinr_cosp = 2.0 * (w * x + y * z)
cosr_cosp = 1.0 - 2.0 * (x * x + yy)
roll = sinr_cosp.atan2(cosr_cosp)
sinp = 2.0 * (w * y - z * x)
sinp = sinp.clamp(min=-1.0, max=1.0)
pitch = sinp.asin()
```

[...]

PR #3993, opened 2026-08-21
fix(geometry): handle gimbal lock in euler_from_quaternion (#3950)

### LINKED ISSUE #3950: euler_from_quaternion returns a triple that does not represent the input rotation at pitch = ±pi/2: 200/200 random cases fail, worst max|R_in - R_back| = 1.986
At `pitch = ±pi/2` (gimbal lock) `euler_from_quaternion` returns a `(roll, pitch, yaw)` triple that **does not represent the input rotation at all** — not merely a different-but-equivalent decomposition. Over 200 random `(roll, yaw)` pairs at `pitch = +pi/2`, reconstructing the rotation from the returned triple fails in **200/200** cases, with a worst-case `max|R_in − R_back| = 1.986` (the entries of a rotation matrix are bounded by 1, so this is a completely different rotation). The control at `|pitch| < pi/4` is `0/200` failures with a worst case of `1.11e-15`.

`euler_from_quaternion` has no gimbal-lock special case (`kornia/geometry/conversions.py:948-960`):

```python
sinr_cosp = 2.0 * (w * x + y * z)
cosr_cosp = 1.0 - 2.0 * (x * x + yy)
roll = sinr_cosp.atan2(cosr_cosp)

sinp = 2.0 * (w * y - z * x)
sinp = sinp.clamp(min=-1.0, max=1.0)
pitch = sinp.asin()

siny_cosp = 2.0 * (w * z + x * y)
cosy_cosp = 1.0 - 2.0 * (yy + z * z)
yaw = siny_cosp.atan2(cosy_cosp)
```

At `pitch = ±pi/2` both `atan2` arguments in each of `roll` and `yaw` cancel to ~0 simultaneously, so `atan2(≈0, ≈0)` returns whatever the floating-point residue dictates — in the case below, `pi/2` for all three components regardless of the input `roll` and `yaw`.

## Repro

```python
import torch
    euler_from_quaternion as e_from_q,
    quaternion_from_euler as q_from_e,
    quaternion_to_rotation_matrix as q2R,
)

def R_of(r, p, y):
    return q2R(torch.stack(q_from_e(*[torch.tensor(a, dtype=torch.float64) for a in (r, p, y)])))

# single case
r0, p0, y0 = 0.1, torch.pi / 2, 0.2
q = q_from_e(*[torch.tensor(a, dtype=torch.float64) for a in (r0, p0, y0)])
back = [float(a) for a in e_from_q(*q)]
print("input  (roll, pitch, yaw) =", [r0, p0, y0])
print("output (roll, pitch, yaw) =", back)
print("max|R_in - R_back|        =", (R_of(r0, p0, y0) - R_of(*back)).abs().max().item())

# 200 random (roll, yaw) at pitch = +pi/2
g = torch.Generator().manual_seed(11)
fails, worst = 0, 0.0
for _ in range(200):
    r = float(torch.rand(1, generator=g) * 2 * torch.pi - torch.pi)
    y = float(torch.rand(1, generator=g) * 2 * torch.pi - torch.pi)
    q = q_from_e(*[torch.tensor(a, dtype=torch.float64) for a in (r, torch.pi / 2, y)])
    b = [float(a) for a in e_from_q(*q)]
    d = (R_of(r, torch.pi / 2, y) - R_of(*b)).abs().max().item()
    worst = max(worst, d)
    fails += d > 1e-6
print(f"\npitch = +pi/2 : {fails}/200 do not reproduce the input rotation, worst max|R_in - R_back| = {worst:.4g}")

# control: |pitch| < pi/4, same generator continued
fails, worst = 0, 0.0
for _ in range(200):
    r = float(torch.rand(1, generator=g) * 2 * torch.pi - torch.pi)
    y = float(torch.rand(1, generator=g) * 2 * torch.pi - torch.pi)
    p = float(torch.rand(1, generator=g) * (torch.pi / 2 - 0.1) - torch.pi / 4)
    q = q_from_e(*[torch.tensor(a, dtype=torch.float64) for a in (r, p, y)])
    b = [float(a) for a in e_from_q(*q)]
    d = (R_of(r, p, y) - R_of(*b)).abs().max().item()
    worst = max(worst, d)
    fails += d > 1e-6
print(f"|pitch| < pi/4: {fails}/200 fail, worst max|R_in - R_back| = {worst:.4g}")
```

The comparison is deliberately made **on the reconstructed rotation matrix**, not on the angle triple: a different triple representing the same rotation would be perfectly acceptable at gimbal lock, and this test would pass. It does not.

## Observed

```
input  (roll, pitch, yaw) = [0.1, 1.5707963267948966, 0.2]
output (roll, pitch, yaw) = [1.5707963267948966, 1.5707963267948966, 1.5707963267948966]
max|R_in - R_back|        = 0.09983341664682817

pitch = +pi/2 : 200/200 do not reproduce the input rotation, worst max|R_in - R_back| = 1.986
|pitch| < pi/4: 0/200 fail, worst max|R_in - R_back| = 1.11e-15
```

## Expected

At gimbal lock the `(roll, yaw)` split is genuinely ambiguous — only the combination `roll ∓ yaw` is determined — so no implementation can recover the original pair. But every implementation *can and should* return **some** triple that reconstructs the same rotation. That is the standard special case:

```python
# |sinp| >= 1  (gimbal lock)
pitch = copysign(pi/2, sinp)
roll  = 0
yaw   = ∓ 2 * atan2(x, w)      # sign chosen by the sign of pitch; picks the determined combination
```

With that, `quaternion_from_euler(*euler_from_quaternion(q))` reproduces `R(q)` at `pitch = ±pi/2` exactly, and the pair remains a mutual inverse over the whole domain instead of only over `|pitch| < pi/2`. Away from the lock the function is already exact (`1.11e-15` in the control above), so the repair is confined to the degenerate branch.

Two pieces of contributing context for whoever picks this up:

- **No existing test would catch it.** `tests/geometry/test_conversions.py::TestEulerFromQuaternion::test_forth_and_back` compares only `.abs()` of each component and carries the comment `# TODO: check hwo to prevent getting inverted angles sometimes`. Nothing in the suite pins the axis assignment, the composition order, or the gimbal-lock behavior. A fix should come with a test that reconstructs the rotation matrix, as the repro above does.
- This is **not** the same defect as `euler_from_quaternion` silently accepting a non-unit
  quaternion (#3953), which has a different cause and a different one-line fix. Every input in the
  repro above is a unit quaternion produced by `quaternion_from_euler`; the failure here is purely
  the missing `|sinp| >= 1` branch.
- The composition this function inverts is `R = Rz(yaw) · Ry(pitch) · Rx(roll)` — extrinsic X→Y→Z about the fixed axes, equivalently intrinsic Z–Y′–X″ — verified against four candidate products (`1.1e-16` for this one vs `0.64` / `0.55` / `0.25` for the alternatives). The docstrings say only "XYZ convention", which does not distinguish these; that documentation gap is being closed in the batch-3 conversions convention-block PR.

## Classification

**Broken output.** The returned triple is not a valid decomposition of the input rotation under any convention — it is a wrong answer, not an alternative one, and it is returned silently. Per the reference-core lifecycle (Part I / Part III §2) that puts it in the class that may land as an ordinary bugfix without waiting for the coordinated Repair & Deprecation window: the only outputs that change are ones that are currently incorrect, so the non-degenerate default path can be preserved byte-identically (#3916-style discipline).

## Environment

kornia 0.9.0rc1 @ `3358ef16`, torch 2.9.1, cpu. Executed from a worktree carrying the batch-3a docs/tests commits on top of `3358ef16`; `euler_from_quaternion` and `quaternion_from_euler` are byte-identical to `origin/main` there (no diff hunk lands in either; function bodies compare AST-identical).

syahra712:
Closes #3950

`euler_from_quaternion` has no gimbal-lock branch. At pitch = ±π/2 the cos(pitch) factor shared by the roll and yaw `atan2` arguments collapses to ~0, so each becomes `atan2(0, 0)` and the returned (roll, pitch, yaw) no longer represents the input rotation — 200/200 random (roll, yaw) at pitch = π/2 fail to reconstruct, while |pitch| < π/4 round-trips to rounding.

Fix: add a gimbal-lock branch. At the singularity only one combined angle is determined (roll − yaw at +π/2, roll + yaw at −π/2), so snap pitch to exactly ±π/2, pin roll to 0, and fold the resolved angle into yaw (`yaw = ∓2·atan2(x, w)`). The round trip then reconstructs the input rotation to ~1e-16. Detection is `cos(pitch) < 8·sqrt(eps)`, which scales with dtype so the float32 case is caught too (there cos(pitch) rounds to ~3e-4 at the singularity).

Tests: the suite already pins this issue, so I flipped those pins as the comments intended — removed the strict `xfail` on `test_convention_roundtrip_holds_at_gimbal_lock_3950` and turned the companion wart into a both-signs regression. The `#3953` "ignores the norm" wart triple is updated too: a scaled quaternion now saturates asin through the new branch and returns (0, π/2, yaw) instead of the old atan2 residue (the norm bug itself is unchanged, and the new value is scale-invariant, so more stable). Full `tests/geometry/test_conversions.py` passes locally (284 passed, 21 xfailed).

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -17,6 +17,7 @@
 
 from __future__ import annotations
 
+import math
 import warnings
 from typing import Optional
 
@@ -1286,28 +1287,6 @@ def euler_from_quaternion(
           rotation, to ``1.1e-16``
 
     .. warning::
-        At ``pitch = ±pi/2`` the returned triple usually does not represent the
-        input rotation, and no gimbal-lock branch exists to say so. ``roll`` and
-        ``yaw`` come from ``atan2`` of two quantities that cancel to nothing
-        there, so the triple that comes back is decided by rounding: it varies
-        between dtypes, between PyTorch versions, and under a one-ulp change of
-        the input pitch, and no specific triple is quoted here for that reason.
-        What is stable is that ``pitch`` lands within about ``sqrt(eps)`` of
-        ``±pi/2`` — the ``asin`` argument rounds to within an ulp of ``1``, and
-        ``asin(1 - d)`` is ``pi/2 - sqrt(2d)`` — and that the reconstructed
-        rotation is far from the input. Whether ``±pi/2`` is reached exactly
-        depends on dtype and build: at ``float64`` it is exact here, while the
-        ``float32`` round trip of ``(0.1, pi/2, 0.2)`` returns pitch
-        ``1.570451``, ``3.45e-4`` *below* ``float32``'s ``pi/2`` — an exact
-        ``pitch == pi/2`` check never fires there. Random ``(roll, yaw)``
-        at ``pitch = +pi/2`` fail this way, while ``|pitch| < pi/4`` round trips
-        to rounding. The rotation does survive on the diagonal
-        ``roll = yaw`` at ``+pi/2`` (and ``roll = -yaw`` at ``-pi/2``), where
-        random draws round trip to within a few parts in ``1e8``, though
-        ``roll`` and ``yaw`` are still not returned individually. Tracked in
-        `#3950 <https://github.com/kornia/kornia/issues/3950>`_.
-
-    .. warning::
         The input is **not** normalised, so a non-unit quaternion silently gives
         a wrong triple: for the ``q`` of ``(0.3, 0.7, 1.1)``, passing ``2 * q``
         returns ``[1.6560585860248003, 1.5707963267948966,
@@ -1343,6 +1322,31 @@ def euler_from_quaternion(
     siny_cosp = 2.0 * (w * z + x * y)
     cosy_cosp = 1.0 - 2.0 * (yy + z * z)
     yaw = siny_cosp.atan2(cosy_cosp)
+
+    # Gimbal lock: at pitch = ±pi/2 the cos(pitch) factor shared by the roll and
+    # yaw atan2 arguments collapses to ~0, so each becomes atan2(0, 0) and the
+    # returned triple no longer represents the rotation (#3950). Only one combined
+    # angle survives there -- roll - yaw at +pi/2 and roll + yaw at -pi/2 -- so pin
+    # roll to 0 and fold that degree of freedom into yaw, which reconstructs the
+    # input rotation. The threshold scales with sqrt(eps): once the asin argument
+    # has rounded to within a few ulps of 1, cos(pitch) is a small multiple of
+    # sqrt(eps) (asin(1 - d) = pi/2 - sqrt(2 d)), so a modest factor above sqrt(eps)
+    # clears that rounding band while staying far from any genuinely resolvable
+    # pitch. Measured sweep: factor 1.0 misses float32 gimbal (asin argument
+    # rounds below 1); 2.0 catches both dtypes with no regression at any sampled
+    # offset from pi/2; 8.0+ creates a band where resolvable pitches are snapped
+    # and the error is worse than main. 2.0 is the tightest margin that works.
+    cos_pitch = (1.0 - sinp * sinp).clamp(min=0.0).sqrt()
+    gimbal = cos_pitch < 2.0 * torch.finfo(w.dtype).eps ** 0.5
+    up = sinp > 0.0
+    # Snap pitch to exactly ±pi/2 there (asin returns pi/2 - O(sqrt(eps)) once its
+    # argument rounds below 1, which alone would leave the round trip off by ~1e-8),
+    # pin roll to 0 and put the resolved degree of freedom into yaw.
+    pitch_locked = torch.where(up, torch.full_like(pitch, math.pi / 2), torch.full_like(pitch, -math.pi / 2))
+    yaw_locked = torch.where(up, -2.0 * x.atan2(w), 2.0 * x.atan2(w))
+    pitch = torch.where(gimbal, pitch_locked, pitch)
+    roll = torch.where(gimbal, torch.zeros_like(roll), roll)
+    yaw = torch.where(gimbal, yaw_locked, yaw)
 
     return roll, pitch, yaw
 
--- a/tests/geometry/test_conversions.py
+++ b/tests/geometry/test_conversions.py
@@ -5999,28 +5999,15 @@ class TestEulerFromQuaternion(BaseTester):
             f"roll for the w = {zero_sign} half-turn is {roll.item()}, not the exact {expected_roll} endpoint"
         )
 
-    @pytest.mark.xfail(
-        raises=AssertionError,
-        reason="euler_from_quaternion has no gimbal-lock branch, so at pitch = ±pi/2 the returned "
-        "triple does not represent the input rotation — kornia#3950",
-        strict=True,
-    )
     def test_convention_roundtrip_holds_at_gimbal_lock_3950(self, device):
-        # Intended behavior: euler_from_quaternion returns *a* triple representing the input
-        # rotation. At pitch = ±pi/2 (gimbal lock) roll and yaw are individually undetermined --
-        # only their sum or difference is -- so no library can return the input triple back, but a
-        # correct implementation still returns a triple whose rotation matrix is the input's, which
-        # is what this pin asserts. There is no gimbal-lock branch at all: roll and yaw come from
-        # atan2 of two quantities that both cancel there, and the result is simply wrong. For
-        # (roll, pitch, yaw) = (0.1, pi/2, 0.2) in float64 the reconstructed rotation is far from
-        # the input -- by a margin that varies with rounding, so no figure is quoted here -- and
-        # random (roll, yaw) at pitch = +pi/2 fail the same way, while |pitch| < pi/4 round trips
-        # to rounding. float64 is hardcoded and the dtype fixture
-        # dropped because the returned triple is wildly dtype-dependent here (see the companion
-        # wart), and the skip is visible so a raw TypeError on MPS, which has no float64, cannot
-        # satisfy the raises=AssertionError mark instead of the assertion. Marked xfail(strict=True)
-        # so fixing #3950 makes this XPASS and forces the mark out. Companion wart:
-        # test_wart_gimbal_lock_returns_a_wrong_triple_3950.
+        # euler_from_quaternion returns *a* triple representing the input rotation, including at
+        # pitch = ±pi/2 (gimbal lock). There roll and yaw are individually undetermined -- only
+        # their sum or difference is -- so the input triple cannot be handed back, but the returned
+        # triple's rotation matrix must still equal the input's. Regression for kornia#3950: the
+        # gimbal-lock branch pins roll to 0, snaps pitch to ±pi/2 and folds the one resolved angle
+        # into yaw. float64 is hardcoded because the round-trip margin is a float64 fact, and the
+        # skip is visible so a raw TypeError on MPS, which has no float64, cannot pass for the
+        # assertion. Companion: test_convention_roundtrip_holds_at_gimbal_lock_both_signs_3950.
         _skip_if_dtype_unavailable(device, torch.float64)
 
         roll = torch.tensor(0.1, device=device, dtype=torch.float64)
@@ -6037,65 +6024,16 @@ class TestEulerFromQuaternion(BaseTester):
         )
 
     @pytest.mark.parametrize("sign", [1.0, -1.0], ids=["pitch_plus_pi_over_2", "pitch_minus_pi_over_2"])
-    def test_wart_gimbal_lock_returns_a_wrong_triple_3950(self, device, sign):
-        # Wart pin for kornia#3950, companion to the strict xfail above. It pins the two facts about
-        # gimbal lock that are STABLE, and deliberately pins no exact triple.
-        #
-        # Pinning the returned triples themselves ((pi/2, pi/2, pi/2) at +pi/2 and (0, -pi/2, pi/2)
-        # at -pi/2 on this build) is not an option: those values are not reproducible. roll and yaw
-        # come from atan2 of two quantities that cancel to ~1e-17 there, so which way they
-        # cancel is decided by rounding. Perturbing the input pitch by a single ulp on this very
-        # build changes the +pi/2 triple to (0.15500, pi/2, 0.35877) at -2 ulp and to
-        # (-3.12597, pi/2, -3.07917) at +1 ulp; a review on torch 2.12.0 saw different triples again
-        # on the unperturbed input. Kornia declares torch>=2.0.0, so pinning any one of them makes
-        # the suite red on builds the pin was never measured against, for a value that is not the
-        # defect being tracked.
-        #
-        # What this pin asserts instead:
-        #   1. pitch_back is +-pi/2 to a tolerance -- the asin saturates. Note this fact SURVIVES a
-        #      fix to #3950 (a correct gimbal-lock branch still reports pitch = +-pi/2); it is
-        #      pinned as the structural claim the strict xfail above does not make, not as a defect
-        #      indicator.
-        #   2. the round-tripped rotation is far from the input -- this is the defect, and the half
-        #      of this pin that flips when #3950 is fixed.
-        #
-        # Assertion 1 is a TOLERANCE and not exact equality, which matters. At gimbal lock the asin
-        # argument is 1 - O(eps), and asin(1 - d) ~= pi/2 - sqrt(2d), so one ulp of slack in the
-        # argument amplifies to a sqrt-scale error in the output: sqrt(2 * eps_f64) = 2.107e-08.
-        # Whether the argument rounds to exactly 1.0 or to one ulp below is decided by the last bit
-        # of the sin/cos computation upstream, so it moves between torch builds, backends and
-        # vectorisation paths. torch 2.12.0 reports -1.5707963057214724 for the -pi/2 cell, which is
-        # pi/2 - 2.1073424116835326e-08 -- agreeing with sqrt(2 * eps) to eight significant figures.
-        # Exact equality here is therefore red on 2.12; the tolerance is what keeps the cell green.
-        # Reproducing the 2.12 value locally needs the right probe: perturbing the input pitch, or any component
-        # by a single ulp, does NOT move pitch_back at all (a +-1 ulp sweep over all four quaternion
-        # components returns one distinct value, as does a +-200 ulp input-pitch sweep). Perturbing
-        # w -- the component the saturation actually depends on -- by two or more ulps walks up the
-        # same sqrt-scale family and reproduces the 2.12.0 figure bit-for-bit: w - 2 ulp gives
-        # -1.5707963057214724 on the -pi/2 cell, and w - 3 ulp gives +1.5707963057214724 on the
-        # +pi/2 cell. The rule for the next pin here is therefore to perturb the intermediate the
-        # branch depends on, and to go wider than one ulp -- not to assume the input is the probe.
-        # Tolerance sizing stays mechanism-based rather than sampled: dev = sqrt(2 * k * eps) for an
-        # argument k ulps below 1.0, so 1e-6 is only reached at k ~ 2250, far beyond the one-to-few
-        # ulps a cross-build rounding difference can move it, and far below any real defect, which
-        # would move pitch by O(1). Note the deviation is NOT bounded by the values above: pushing w
-        # to -20000 ulp reaches 2.5e-06 and does cross the tolerance. That is not a realistic
-        # rounding difference, but it is why the sizing argument is the mechanism and the measured
-        # figures here (2.1e-08 at w - 2 ulp, 5.4e-08 at w - 10 ulp) are sample points, not bounds.
-        #
-        # The 1e-9 floor in 2 is likewise a chosen threshold with margin, NOT a measured bound:
-        # a correct gimbal-lock branch round trips to ~1e-16, and widening the probe drives the
-        # sample minimum steadily toward 0, which is why no sampled extremum is quoted as a bound.
-        #
-        # Two cells, one per sign, because a fix could plausibly add a gimbal-lock branch for one
-        # sign only or get the roll/yaw split sign wrong; the strict xfail above only covers +pi/2,
-        # so the -pi/2 cell here is the only coverage of that sign. If either cell fails, #3950 was
-        # (partly) fixed -- flip/remove the strict xfail above. NOT a contract that the current
-        # output is correct: any triple whose rotation matrix matches the input is an acceptable
-        # replacement, and such a triple would fail assertion 2 as intended.
-        # float64 is hardcoded and the dtype fixture dropped because the round-trip margin is a
-        # float64 fact; the skip is visible so a raw TypeError on MPS, which has no float64, cannot
-        # pass for the assertion.
+    def test_convention_roundtrip_holds_at_gimbal_lock_both_signs_3950(self, device, sign):
+        # Regression for kornia#3950 across both gimbal-lock signs (the single-case companion above
+        # only covers +pi/2). Two facts are pinned:
+        #   1. pitch_back saturates to sign*pi/2 -- the gimbal-lock branch reports it exactly.
+        #   2. the round-tripped rotation matches the input -- the half the pre-fix code got wrong,
+        #      where roll and yaw came from atan2 of two quantities that both cancel there.
+        # roll and yaw are not pinned individually: at gimbal lock only one of (roll ± yaw) is
+        # determined, so any triple whose rotation matrix matches the input is acceptable. float64
+        # is hardcoded because the round-trip margin is a float64 fact; the skip is visible so a raw
+        # TypeError on MPS, which has no float64, cannot pass for the assertion.
         _skip_if_dtype_unavailable(device, torch.float64)
 
         pitch_in = sign * torch.pi / 2
@@ -6116,9 +6054,9 @@ class TestEulerFromQuaternion(BaseTester):
         rot_back = kornia.geometry.conversions.quaternion_to_rotation_matrix(torch.stack(roundtrip))
         error = (rot_in - rot_back).abs().max().item()
 
-        assert error > 1e-9, (
-            f"kornia#3950: the triple returned at pitch = {pitch_in} now reproduces the input "
-            f"rotation to {error} -- the gimbal-lock defect looks fixed"
+        assert error < 1e-12, (
+            f"kornia#3950: the triple returned at pitch = {pitch_in} does not reproduce the input "
+            f"rotation (error {error})"
         )
 
     @pytest.mark.xfail(
@@ -6158,7 +6096,12 @@ class TestEulerFromQuaternion(BaseTester):
         # in TestQuaternionExpToLog because the two functions are independent code paths and a fix
         # to one leaves the other broken, which would leave the other strict xfail silently XFAIL.
         # If it fails, the euler half of #3953 was fixed -- flip/remove the strict xfail above.
-        # NOT a contract that a scaled quaternion must keep producing this triple.
+        # NOT a contract that a scaled quaternion must keep producing this triple. The scale bug is
+        # unchanged (the function still does not normalise); the triple below moved when the
+        # gimbal-lock branch for #3950 landed -- scaling by 2 saturates the asin argument, which now
+        # routes through that branch and returns (0, pi/2, yaw) instead of the old atan2 residue.
+        # The new value is more stable: yaw = -2*atan2(x, w) is scale-invariant, so it no longer
+        # depends on rounding the way the pre-#3950 triple did.
         # Snippet used to generate expected (torch only, executed on cpu float64):
         #   t = lambda x: torch.tensor(x, dtype=torch.float64)
         #   q = quaternion_from_euler(t(0.3), t(0.7), t(1.1))
@@ -6166,10 +6109,10 @@ class TestEulerFromQuaternion(BaseTester):
         #   [x.item() for x in euler_from_quaternion(*q)]
         #     -> [0.2999999999999999, 0.6999999999999998, 1.0999999999999999]     (the unit input)
         #   [x.item() for x in euler_from_quaternion(*[2 * c for c in q])]
-        #     -> [1.6560585860248003, 1.5707963267948966, 2.1048169977173687]
-        #   (float32: [1.656058430671692, 1.5707963705062866, 2.1048169136047363];
-        #    float16:  [1.6572265625, 1.5703125, 2.10546875];
-        #    bfloat16: [1.6484375, 1.5703125, 2.109375] -- all within the dtype's default tolerance
+        #     -> [0.0, 1.5707963267948966, 0.14034560699328846]
+        #   (float32: [0.0, 1.5707963705062866, 0.14034557342529297];
+        #    float16:  [0.0, 1.5703125, 0.1405029296875];
+        #    bfloat16: [0.0, 1.5703125, 0.138671875] -- all within the dtype's default tolerance
         #    of the float64 literals below)
         quaternion = quaternion_from_euler(
             torch.tensor(0.3, device=device, dtype=dtype),
@@ -6181,7 +6124,7 @@ class TestEulerFromQuaternion(BaseTester):
 
         assert_close(
             out,
-            torch.tensor([1.6560585860248003, 1.5707963267948966, 2.1048169977173687], device=device, dtype=dtype),
+            torch.tensor([0.0, 1.5707963267948966, 0.14034560699328846], device=device, dtype=dtype),
             msg=_issue_msg("kornia#3953: euler_from_quaternion no longer ignores the quaternion norm"),
         )
 


ducha-aiki (changes requested):
The approach is right and the win at the singularity is large. Snapping pitch to exactly `±pi/2`, pinning roll to 0 and folding the surviving degree of freedom into `yaw = ∓2·atan2(x, w)` reconstructs the rotation correctly — I verified that, not just the shapes. Two things need fixing first, and one of them I have measured a specific replacement for.

Everything below is on your branch rebased onto current `main` (`35555b55`); the conflict is trivial, see the end.

## The win

Round-tripping `(roll, pitch, yaw) -> quaternion -> euler -> quaternion` and comparing rotation matrices in float64, plus the gradient of `roll + pitch + yaw` at the exact singularity:

```text
                                  main                       this PR
exact gimbal, f32   max|dR|       9.983e-02                  0.000e+00
exact gimbal, f64   max|dR|       9.983e-02                  3.331e-16
grad at gimbal, f32               [4090.9, 4.7e7, ...]       [-0.1414, -2.8249, 0, 0]
grad at gimbal, f64               [0, -2.0e16, 5.1e17, ...]  [-0.1414, -2.8249, 0, 0]
```

A 10 % rotation error becomes exact, and a gradient that was garbage by 16 orders of magnitude becomes finite, small, and identical across dtypes. That is a real fix.

## Blocking: the threshold is too wide and creates a regression band

`8.0 * sqrt(eps)` is `2.76e-3` at float32 and `1.19e-7` at float64. Inside that band the snap fires on rotations whose pitch is still resolvable, and the error it introduces is exactly the distance from `pi/2` that it discards — visible in the numbers, where the PR's error tracks `d` one-for-one:

```text
float32                    main         this PR (8.0)
pitch = pi/2 - 1e-5        9.538e-04    1.001e-05     PR better
pitch = pi/2 - 1e-4        2.440e-04    9.942e-05     PR better
pitch = pi/2 - 1e-3        1.533e-04    9.949e-04     PR 6.5x WORSE
pitch = pi/2 - 2.7e-3      3.224e-06    2.686e-03     PR 830x WORSE

float64                    main         this PR (8.0)
pitch = pi/2 - 1e-9        1.455e-08    9.950e-10     PR better
pitch = pi/2 - 1e-7        2.275e-09    9.950e-08     PR 44x WORSE
```

### The fix is one character: `8.0` -> `2.0`

I swept the factor and measured all three:

| factor | exact gimbal caught? | regression band |
|---|---|---|
| 1.0 | **no** at float32 — `max\|dR\|` back to `9.983e-02` and the `4.7e7` gradient returns | none |
| **2.0** | **yes**, both dtypes (`0.0` / `3.3e-16`, clean gradients) | **none** |
| 8.0 (yours) | yes | the table above |

At `2.0` the PR is **better than or equal to `main` at every point I sampled, in both dtypes**: `pi/2 - 1e-3` goes to `1.533e-4` and `pi/2 - 2.7e-3` to `3.224e-6`, both exactly `main`'s values, while the exact-gimbal case keeps the full win. `tests/geometry` passes 2388/0 at that factor, and `test_conversions.py --dtype=all` gives 59 failures, the same pre-existing baseline as `main`.

Note that `1.0` failing at float32 vindicates your reasoning — a margin above `sqrt(eps)` genuinely is needed, since `asin`'s argument rounds below 1 and `cos(pitch)` lands a small multiple of `sqrt(eps)` rather than at zero. The margin just has to be small enough not to swallow resolvable pitches, and `2.0` is where that lands. Please also record the measurement in the comment: the current text argues for "a modest factor above `sqrt(eps)`" without saying why any particular one, and a bare `8.0` is what invites the next person to widen it again.

## Blocking: the docstring your fix falsifies

`grep -rnE "#3950|issues/3950" kornia/` returns two hits on the patched tree. The one at line 1384 is your own explanatory comment and is fine. The other is the `.. warning::` block ending at **`kornia/geometry/conversions.py:1343`**, inside `euler_from_quaternion`'s docstring, and it is now false end to end — it describes the returned triple as not representing the rotation, quotes the float32 round trip of `(0.1, pi/2, 0.2)` returning pitch `1.570451`, and states that "an exact `pitch == pi/2` check never fires there". Your change makes that check fire by construction. The whole block goes, and the grep has to come back empty before this merges.

## Trivial conflict

The branch went `CONFLICTING` on the import line only:

```text
HEAD:  import warnings
ours:  import math
```

Keep both, alphabetically — `import math` then `import warnings`. That is how I resolved it, and every number above comes from that tree.

Requesting changes on the threshold and the docstring. With `2.0` in place this is a clear improvement everywhere and I want it in.

---

syahra712:
Rebased onto current main (resolved the `import math` / `import warnings` conflict — kept both, alphabetically).

Addressed both blocking items:

1. **Threshold `8.0` → `2.0`**: updated the factor and expanded the comment with the measured sweep rationale (1.0 misses float32 gimbal, 2.0 catches both dtypes with no regression, 8.0+ creates a regression band).

2. **Removed the false `.. warning::` block**: the 22-line docstring warning about the gimbal-lock triple not representing the rotation is now incorrect — the fix makes it represent the rotation. Removed it; `grep -rn '#3950' kornia/` now only returns the implementation comment at the gimbal-lock branch, which is accurate.

Ready for re-review.

ducha-aiki:
**REQUEST CHANGES** — the approach is right and the docstring cleanup is right, but the gimbal detector does not actually fire at the poles most of the time. Measured, not reasoned.

Setup: worktree at the head `c8d70b5a`, merge-base `94d0284c`; `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS. Metric throughout: `max |R(q) - R(q_back)|` where `q_back = quaternion_from_euler(*euler_from_quaternion(q))` — the same round-trip #3950 uses.

## The blocker

Over 500 random `(roll, yaw)` at each pole, the round trip *still* returns a wrong rotation:

```
float64  +pi/2:   8/500 = 1.6% still broken (err > 1e-4)
float64  -pi/2:   6/500 = 1.2%
float32  +pi/2:  21/500 = 4.2%
float32  -pi/2:  18/500 = 3.6%
worst-case err at the poles, 200 samples each:
  float64  +pi/2 = 1.90e+00   -pi/2 = 1.43e+00
  float32  +pi/2 = 1.87e+00   -pi/2 = 1.54e+00
```

`1.9` is the same order as the `1.986` in the issue title, so for those inputs nothing improved.

**Cause: `2.0 * sqrt(eps)` is a tie, not a margin.** `cos_pitch = sqrt(1 - sinp*sinp)` is computed from an already-rounded `sinp`, so at a true pole it does not evaluate to ~0 — it evaluates to `sqrt(2 * k * eps)` where `k` is however many ulps `|sinp|` fell short of 1, and `k` varies with `(roll, yaw)`. `k = 2` puts it *exactly* on your threshold:

```
                sinp                  cos_pitch    threshold    gimbal
float64  +pi/2  0.99999999999999978   2.107e-08    2.980e-08    True
float64  -pi/2 -0.99999999999999956   2.980e-08    2.980e-08    False   <- err 5.95e-01
float32  +pi/2  0.99999976158142090   6.905e-04    6.905e-04    False   <- err 3.89e-01
float32  -pi/2 -0.99999994039535522   3.453e-04    6.905e-04    True
```

Two of the four (dtype, pole) combinations for a single input land on the boundary and are missed. "2.0 is the tightest margin that works" is measuring a coincidence of the sampled inputs; raising the factor is not the fix either, because your own note about the `8.0+` regression band is correct — with this estimator the pole and a genuinely resolvable pitch are only a factor of ~4 apart, so there is no factor that separates them.

## The fix: measure `cos(pitch)` without the cancellation

`sinr_cosp` and `cosr_cosp` are both proportional to `cos(pitch)` already, and computing their magnitude avoids the `1 - sinp**2` cancellation entirely. Worst-case over 50 random `(roll, yaw)`, in units of `sqrt(eps)`:

```
dtype     pitch        sqrt(1-sinp^2)     hypot(sinr_cosp, cosr_cosp)
float64   +pi/2              1.732 sq              3.799e-08 sq
float64   -pi/2              1.732 sq              4.486e-08 sq
float64   pi/2-1e-7          6.782 sq              6.711e+00 sq
float32   +pi/2              2.000 sq              6.905e-04 sq
float32   -pi/2              1.732 sq              6.905e-04 sq
float32   pi/2-1e-3          3.464 sq              2.897e+00 sq
```

Your estimator separates pole from non-pole by a factor of 3.9; `hypot` separates them by eight orders of magnitude in float64 and by ~4000 in float32. Change two lines:

```python
    cos_pitch = torch.hypot(sinr_cosp, cosr_cosp)
    gimbal = cos_pitch < 2.0 * torch.finfo(w.dtype).eps ** 0.5
```

Same threshold, better quantity. Measured, same harness:

```
                +pi/2      -pi/2      general    offsets from pi/2 (float32)
head as-is      1.87e+00   1.54e+00   7.75e-07   1e-4:9.5e-05 1e-5:9.5e-06 1e-6:8.9e-07
with hypot      4.77e-07   3.58e-07   7.75e-07   1e-4:9.5e-05 1e-5:9.5e-06 1e-6:8.9e-07
float64 head    1.90e+00   1.43e+00   2.89e-15
float64 hypot   8.88e-16   8.88e-16   2.89e-15
```

Both poles land at round-off in both dtypes, the near-pole improvement you built is preserved **unchanged**, and `general` is untouched. `torch.hypot` is TorchScript-compatible, and the value is only consumed by a comparison, so no gradient path changes.

## One consequence to handle in the same push

`test_wart_euler_from_quaternion_ignores_the_norm_3953` has to be regenerated. Its literals currently encode a side effect of the `sinp` estimator: scaling `q` by 2 makes `sinp = 4 * sin(pitch)` saturate the clamp, so *every* non-unit quaternion with `|pitch| > ~14.5 deg` now reports gimbal lock and returns `(0, +-pi/2, .)`. Your pin comment documents that honestly, but it is a wider wart than the one it replaced. With `hypot` — which scales with `|q|**2` and therefore does not saturate — that pin fails and needs new literals:

```
2 failed, 618 passed, 14 xfailed        # the 3953 wart pin, both dtypes
```

Everything else in `tests/geometry/test_conversions.py` passes under the replacement.

## Not blocking

The threshold rationale comment should be rewritten once the estimator changes — most of what it currently says (the `asin(1 - d) = pi/2 - sqrt(2d)` reasoning, the `1.0` / `2.0` / `8.0` sweep) is about the estimator you would be replacing.

The removal of the 22-line `.. warning::` and the `#3950` grep are both correct and I have no objection to either.

victorwon2001:
**tests/geometry/test_conversions.py:6027** — The added both-sign gimbal-lock regression test removes the dtype fixture and explicitly hardcodes torch.float64, so the dtype-sensitive detector is not regression-tested in float32.

The test can remain green while the implementation still misses exact gimbal-lock inputs in float32; the supplied measurements show float32 has the larger observed failure rate at both poles.

Not independently executed. The changed test signature and comments explicitly hardcode float64, while the supplied pole sweep reports 21/500 and 18/500 failures for the two float32 pole signs. Repository guidance also calls for dtype coverage where applicable.

`file:kornia/geometry/conversions.py:diff`

> +    cos_pitch = (1.0 - sinp * sinp).clamp(min=0.0).sqrt()
> +    gimbal = cos_pitch < 2.0 * torch.finfo(w.dtype).eps ** 0.5

`comment:5485141071`

> Over 500 random `(roll, yaw)` at each pole, the round trip *still* returns a wrong rotation:
> 
> ```
> float64  +pi/2:   8/500 = 1.6% still broken (err > 1e-4)
> float64  -pi/2:   6/500 = 1.2%
> float32  +pi/2:  21/500 = 4.2%
> float32  -pi/2:  18/500 = 3.6%
> worst-case err at the poles, 200 samples each:
>   float64  +pi/2 = 1.90e+00   -pi/2 = 1.43e+00
>   float32  +pi/2 = 1.87e+00   -pi/2 = 1.54e+00
> ```
> 
> `1.9` is the same order as the `1.986` in the issue title, so for those inputs nothing improved.

`comment:5485141071`

> **Cause: `2.0 * sqrt(eps)` is a tie, not a margin.** `cos_pitch = sqrt(1 - sinp*sinp)` is computed from an already-rounded `sinp`, so at a true pole it does not evaluate to ~0 — it evaluates to `sqrt(2 * k * eps)` where `k` is however many ulps `|sinp|` fell short of 1, and `k` varies with `(roll, yaw)`. `k = 2` puts it *exactly* on your threshold:
> 
> ```
>                 sinp                  cos_pitch    threshold    gimbal
> float64  +pi/2  0.99999999999999978   2.107e-08    2.980e-08    True
> float64  -pi/2 -0.99999999999999956   2.980e-08    2.980e-08    False   <- err 5.95e-01
> float32  +pi/2  0.99999976158142090   6.905e-04    6.905e-04    False   <- err 3.89e-01
> float32  -pi/2 -0.99999994039535522   3.453e-04    6.905e-04    True
> ```
> 
> Two of the four (dtype, pole) combinations for a single input land on the boundary and are missed. "2.0 is the tightest margin that works" is measuring a coincidence of the sampled inputs; raising the factor is not the fix either, because your own note about the `8.0+` regression band is correct — with this estimator the pole and a genuinely resolvable pitch are only a factor of ~4 apart, so there is no factor that separates them.

`comment:5485141071`

> ## The fix: measure `cos(pitch)` without the cancellation
> 
> `sinr_cosp` and `cosr_cosp` are both proportional to `cos(pitch)` already, and computing their magnitude avoids the `1 - sinp**2` cancellation entirely. Worst-case over 50 random `(roll, yaw)`, in units of `sqrt(eps)`:
> 
> ```
> dtype     pitch        sqrt(1-sinp^2)     hypot(sinr_cosp, cosr_cosp)
> float64   +pi/2              1.732 sq              3.799e-08 sq
> float64   -pi/2              1.732 sq              4.486e-08 sq
> float64   pi/2-1e-7          6.782 sq              6.711e+00 sq
> float32   +pi/2              2.000 sq              6.905e-04 sq
> float32   -pi/2              1.732 sq              6.905e-04 sq
> float32   pi/2-1e-3          3.464 sq              2.897e+00 sq
> ```
> 
> Your estimator separates pole from non-pole by a factor of 3.9; `hypot` separates them by eight orders of magnitude in float64 and by ~4000 in float32. Change two lines:
> 
> ```python
>     cos_pitch = torch.hypot(sinr_cosp, cosr_cosp)
>     gimbal = cos_pitch < 2.0 * torch.finfo(w.dtype).eps ** 0.5
> ```
> 
> Same threshold, better quantity. Measured, same harness:
> 
> ```
>                 +pi/2      -pi/2      general    offsets from pi/2 (float32)
> head as-is      1.87e+00   1.54e+00   7.75e-07   1e-4:9.5e-05 1e-5:9.5e-06 1e-6:8.9e-07
> with hypot      4.77e-07   3.58e-07   7.75e-07   1e-4:9.5e-05 1e-5:9.5e-06 1e-6:8.9e-07
> float64 head    1.90e+00   1.43e+00   2.89e-15
> float64 hypot   8.88e-16   8.88e-16   2.89e-15
> ```
> 
> Both poles land at round-off in both dtypes, the near-pole improvement you built is preserved **unchanged**, and `general` is untouched. `torch.hypot` is TorchScript-compatible, and the value is only consumed by a comparison, so no gradient path changes.

`file:tests/geometry/test_conversions.py:diff`

> +    def test_convention_roundtrip_holds_at_gimbal_lock_both_signs_3950(self, device, sign):
> +        # Regression for kornia#3950 across both gimbal-lock signs (the single-case companion above
> +        # only covers +pi/2). Two facts are pinned:
> +        #   1. pitch_back saturates to sign*pi/2 -- the gimbal-lock branch reports it exactly.
> +        #   2. the round-tripped rotation matches the input -- the half the pre-fix code got wrong,
> +        #      where roll and yaw came from atan2 of two quantities that both cancel there.
> +        # roll and yaw are not pinned individually: at gimbal lock only one of (roll ± yaw) is
> +        # determined, so any triple whose rotation matrix matches the input is acceptable. float64
> +        # is hardcoded because the round-trip margin is a float64 fact; the skip is visible so a raw
> +        # TypeError on MPS, which has no float64, cannot pass for the assertion.
>          _skip_if_dtype_unavailable(device, torch.float64)

`context:CONTRIBUTING.md`

> - **Test coverage:** Cover different devices, dtypes, and batch sizes. Use `--dtype` and `--device` pytest arguments to generate test combinations:
> 
>     ```python
>     import pytest
> 
> 
>     @pytest.mark.parametrize("batch_size", [1, 2, 5])
>     def test_smoke(batch_size, device, dtype):
>         x = torch.rand(batch_size, 2, 3, device=device, dtype=dtype)
>         assert x.shape == (batch_size, 2, 3)
>     ```

head_content truncated at 50000 characters: kornia/geometry/conversions.py

base_content truncated at 50000 characters: kornia/geometry/conversions.py

head_content truncated at 50000 characters: tests/geometry/test_conversions.py

base_content truncated at 50000 characters: tests/geometry/test_conversions.py

head_content truncated at 50000 characters: kornia/geometry/conversions.py

base_content truncated at 50000 characters: kornia/geometry/conversions.py

head_content truncated at 50000 characters: tests/geometry/test_conversions.py

base_content truncated at 50000 characters: tests/geometry/test_conversions.py

Replace the cancellation-prone gimbal detector with a stable cos(pitch) magnitude derived from the roll atan2 terms, update the estimator-specific rationale and affected #3953 wart literals, and add regression coverage that exercises both float32 and float64 at both gimbal-lock signs.

SHA: `c8d70b5a3b98b93e285e6c6d4b596132ece256d9`

ducha-aiki:
## Re-verification at `c8d70b5a`, on a merge tree with current `main`

The head has not moved since my 2026-08-31 review, so this is a re-verification rather than a delta review. **Every number below was re-measured independently**, not carried over: worktree at `c8d70b5a` merged with `origin/main` (`8c71df9a0`), torch **2.14.0** (not the 2.9.1 of the earlier pass), CPU, macOS arm64. Metric throughout is `max |R(q) - R(q_back)|` with `q_back = quaternion_from_euler(*euler_from_quaternion(q))`; "broken" means `err > 1e-4`.

**The PR is currently CONFLICTING** (`tests/geometry/test_conversions.py`, one hunk: this branch renamed main's `test_wart_gimbal_lock_returns_a_wrong_triple_3950` while main rewrote the same test's comment block in place). The green CI on this head is therefore stale. I resolved it mechanically to measure.

### First, the part that is right

The fix is a large, real improvement. 500 random `(roll, yaw)` per cell, seed 0:

| cell | `origin/main` | merge tree (this PR) |
|---|---|---|
| float64 +pi/2 | **500/500 broken**, median 5.20e-01, max 1.988 | 9/500 (1.8%), median 3.33e-16, max 1.798 |
| float64 -pi/2 | **500/500**, median 4.97e-01, max 1.970 | 4/500 (0.8%), median 3.33e-16, max 1.486 |
| float32 +pi/2 | **500/500**, median 9.97e-01, max 2.000 | 13/500 (2.6%), median 3.65e-08, max 1.980 |
| float32 -pi/2 | **500/500**, median 1.024, max 2.000 | 13/500 (2.6%), median 2.95e-08, max 1.986 |

The median error drops by fifteen orders of magnitude in float64. The `roll = 0`, `yaw = ±2*atan2(x, w)` split is correct — derived from `quaternion_from_euler`: at `+pi/2`, `qy = qw` and `qz = -qx` with `(yaw - roll)/2 = atan2(-x, w)`; at `-pi/2`, `qy = -qw`, `qz = qx` with `(yaw + roll)/2 = atan2(x, w)`. Both signs in the diff match. Half precision improves too: at the pole the f16 median goes 0.518 -> 0.00093 and bf16 0.978 -> 0.0023.

Regression sweep, merge tree vs base, `tests/geometry/test_conversions.py`:

```
--device=cpu --dtype=float32            420 passed, 1 skipped, 6 xfailed   (base: 419, 1, 7)
--device=cpu --dtype=float64            420 passed, 1 skipped, 6 xfailed   (base: 419, 1, 7)
--device=cpu --dtype=float16,bfloat16   43 failed, 602 passed, 8 xfailed   (base: 43 failed, 601 passed, 9 xfailed)
tests/geometry/test_quaternion.py       88 passed                          (base: 88 passed)
```

The 43 half failures are byte-identical on both trees (pre-existing manifest entries). The `+1 passed / -1 xfailed` is exactly the strict xfail this PR flips. No newly-red node anywhere. `grep -rnE "#3950|issues/3950" kornia/ tests/ docs/` returns one `kornia/` hit — `conversions.py:1385`, inside the new implementation comment, describing the fix rather than a surviving defect. No `docs/` hits.

## [P1] The detector still misses the poles. `2.0 * sqrt(eps)` ties exactly at k = 4.

The blocker from the previous round, reconfirmed on a different torch version. I now have the exact mechanism — and it is **not** the `k = 2` I wrote last time. That part of my earlier comment was wrong; the correct tie is `k = 4`.

`cos_pitch = sqrt(1 - sinp*sinp)` evaluates to exactly `sqrt(k * eps)` where `k` is how many ulps `|sinp|` falls short of `1.0` (because `1 - fl(x*x) = k*eps` for `x = 1 - k*ulp`, `ulp = eps/2`). Measured `k -> cos_pitch` in float64, straight off the merge tree:

```
k = 0 -> 0.0            k = 3 -> 2.5810e-08
k = 1 -> 1.4901e-08     k = 4 -> 2.98023e-08
k = 2 -> 2.1073e-08     k = 5 -> 3.3320e-08
threshold 2*sqrt(eps)  = 2.9802322387695312e-08 = sqrt(4*eps)
```

`sqrt(k*eps) < sqrt(4*eps)` iff `k < 4`. **`k = 4` produces a `cos_pitch` bit-identical to the threshold, and the comparison is a strict `<`, so it is missed.** The failure set is exactly `{k >= 4}` and nothing else:

```
float64 +pi/2   bad-by-k = {4: 9}            all 9 had gimbal = False
float64 -pi/2   bad-by-k = {4: 3, 5: 1}      all 4 had gimbal = False
float32 +pi/2   bad-by-k = {4: 11, 5: 2}     all 13 had gimbal = False
float32 -pi/2   bad-by-k = {4: 11, 5: 2}     all 13 had gimbal = False
```

Every `k <= 3` draw fired and round-tripped to rounding. In float32 the tie is tighter still: the Python threshold `6.9053396600248786e-04` casts to float32 as `6.9053395418450236e-04`, the *same float32 value* as `cos_pitch` at `k = 4`.

(For the record: I quoted `sinp = 0.99999999999999978` as the tie last time. That is 2 ulps below 1, `cos_pitch = 2.1073e-08 < 2.9802e-08`, and it **is** caught. The tie is `k = 4`, `sinp = 0.9999999999999996`. The counts and worst cases stand — 9/4/13/13 at seed 0, 34/24/63/66 over 2000 draws.)

### The in-code rationale is falsified, and the PR description describes different code

- The comment says *"2.0 catches both dtypes with no regression at any sampled offset from pi/2 ... 2.0 is the tightest margin that works."* Measured: 2.0 misses **39 of 2000** pole draws across the four cells.
- **The PR description says `Detection is cos(pitch) < 8·sqrt(eps)`. The code ships `2.0 * torch.finfo(w.dtype).eps ** 0.5`.** The description was not updated in the 2026-08-30 push that changed the constant, and it happens to describe the version that would clear the poles.

### Raising the factor is not the fix — here is the band

float32, 200 draws at each true offset below pi/2, `bad(>1e-4) / max err`:

```
delta        main(no branch)   F=2.0           F=4.0          F=8.0          F=16.0
1e-6         200/0.24          2/0.15          0/1.2e-06      0/1.2e-06      0/1.2e-06
1e-4         182/0.0027        5/0.0018        1/1.0e-04      1/1.0e-04      1/1.0e-04
1e-3          47/2.6e-04      47/2.6e-04     200/1.0e-03    200/1.0e-03    200/1.0e-03
2.8e-3         0/9.5e-05       0/9.5e-05       0/9.5e-05      7/2.8e-03    200/2.8e-03
```

Confirmed on the head with `8.0` patched in: `f32 pitch = pi/2 - 1e-3` gives **200/200 broken, max 1.0000e-03**, against 47/200 max 2.6e-04 on `main`. `F >= 4` clears the poles but snaps genuinely resolvable float32 pitches and lands ~4x worse than `main` at `1e-3`. With this estimator the pole and a resolvable pitch are a factor of ~4 apart; no factor separates them.

### `torch.hypot` does separate them — 0 failures everywhere

`cos_pitch = torch.hypot(sinr_cosp, cosr_cosp)`, same `2.0 * sqrt(eps)` threshold, same harness:

```
poles, 500 draws/cell        f64+        f64-        f32+        f32-
main (no branch)             500/500     500/500     500/500     500/500
head (sinp estimator)          9/500       4/500      13/500      13/500
hypot                          0/500       0/500       0/500       0/500
  max err                    8.88e-16    8.88e-16    2.43e-07    2.43e-07

float32 near-pole            main          sinp(head)         hypot
delta = 1e-6                 200/0.24      2/0.15             0/1.2e-06
delta = 1e-5                 199/0.031     0/1.0e-05          0/1.0e-05
delta = 1e-4                 182/0.0027    5/0.0018           1/1.0e-04
delta = 7e-4                  94/3.5e-04   111/6.9e-04        94/3.5e-04
delta = 3e-3                   0/7.9e-05    0/7.9e-05          0/7.9e-05

general pitches 0.0 / 0.3 / 1.1 / 1.5 / -1.2, 300 draws each, f64 and f32:
sinp and hypot agree to the last printed digit, 0 failures on both.
```

`hypot` is never worse than the shipped estimator anywhere I measured, strictly better at and just below both poles, identical to `main` away from them.

### Bonus: `hypot` lets you drop the `#3953` pin edit entirely

With the `sinp` estimator, scaling `q` by 2 makes `sinp = 4*sin(pitch)` saturate the clamp, so **every** non-unit `2q` with `|pitch| > asin(1/4) = 14.478 deg` routes through the gimbal branch. Measured on the head, `roll = 0.3`, `yaw = 1.1`, float64: `pitch = 14.4 deg -> no gimbal`, `pitch = 14.5 deg -> gimbal`, and every larger pitch too. That is what forced your rewrite of `test_wart_euler_from_quaternion_ignores_the_norm_3953`.

With `hypot` — which scales as `|q|**2` and never saturates — the `2q` triple comes back bit-identical to main's literal:

```
2q triple [main ]  = [1.6560585860248003, 1.5707963267948966, 2.1048169977173687]
2q triple [sinp  ] = [0.0,                1.5707963267948966, 0.14034560699328846]   <- the PR's new literal
2q triple [hypot ] = [1.6560585860248003, 1.5707963267948966, 2.1048169977173687]   <- main's existing literal
```

So under `hypot` the `#3953` pin needs **no** change — revert that hunk and the comment paragraph about it. One issue at a time.

## [P2] Surviving non-equivalent mutant on the one constant this review is about

9 one-token mutations, each re-running the whole file at `--device=cpu --dtype=float32`:

```
M1  factor 2.0 -> 1.0                         KILLED
M2  yaw_locked signs swapped                  KILLED
M3  drop the pitch snap                       KILLED
M4  drop the roll pin                         KILLED
M5  gimbal always False                       KILLED
M6  up = sinp > 0.0  ->  sinp < 0.0           KILLED
M7  x.atan2(w) -> w.atan2(x)                  KILLED
M8  factor 2.0 -> 8.0                         SURVIVED   <-- blocking
M9  drop .clamp(min=0.0)                      SURVIVED   (equivalent, below)
```

**M8 is genuinely non-equivalent**, proved on the patched production file: at the poles 9/4/13/13 -> **0/500 in all four cells**, and at `f32 pitch = pi/2 - 1e-3` it goes 47/200 (max 2.6e-04) -> **200/200 (max 1.0e-03)**. The constant can move by a factor of 4 in the direction that both fixes the blocker and creates the regression band, and the suite stays green. The pin has to constrain it: a cell at a true pole that only passes when the detector actually fires (an input with `k >= 4`), and a near-pole cell that fails if the snap band is widened.

M9 is an equivalent mutant: `1 - x*x >= 0` for all `|x| <= 1`, measured `min = 0.0` over a 2,000,001-point grid in float64/float32/float16/bfloat16 and at the boundary ulps. The clamp is purely defensive.

**Related: the `+pi/2` coverage is nearly inert.** `test_convention_roundtrip_holds_at_gimbal_lock_3950` survived **all nine** mutants. The `[pitch_plus_pi_over_2]` cell of the both-signs pin only fired on M4; every other kill came from `[pitch_minus_pi_over_2]` alone. The `(0.1, +pi/2, 0.2)` input lands at `k = 0` in float64, so `cos_pitch` is exactly `0.0` and any threshold catches it — it cannot discriminate.

## [P2] No changelog fragment

`changelog.d/3993.fixed.md` does not exist on the merge tree. `.github/scripts/check_changelog.py` fails a PR with *"no changed changelog.d/3993.(added|fixed|breaking).md fragment found"*. `fixed` is the right type. That workflow did not exist when this head last ran CI, which is why the green above does not cover it.

## [P3] The bot review's float64 claim — half right

- **Confirmed**: `test_convention_roundtrip_holds_at_gimbal_lock_both_signs_3950` hardcodes `torch.float64` and drops the `dtype` fixture — but so did its predecessor on `main`. The PR inherited the convention; it did not introduce it.
- **Refuted**: it does not red the MPS leg. `env -u PYTORCH_ENABLE_MPS_FALLBACK pytest ...::TestEulerFromQuaternion --device=mps --dtype=float32` gives `10 passed, 6 skipped, 1 xfailed`, with `SKIPPED [5] torch.float64 is unavailable on device mps:0` — a *visible* skip, which is what `_skip_if_dtype_unavailable` exists to produce.
- **Confirmed in substance**: float32 is completely untested at the poles, and that is where the surviving failure rate is highest.
- The `1e-12` tolerance is genuinely float64-specific, so a blanket `dtype` fixture would red the pin. Same input `(0.1, ±pi/2, 0.2)`, round-trip error by dtype on the merge tree: `f64 1.11e-16 / 0`; `f32 0 / 1.991e-08`; `f16 1.716e-04 / 3.261e-04`; `bf16 0 / 1.304e-03`. The fix is a **dtype-scaled tolerance** with the fixture restored, not a float64 hardcode.

## Not a defect: the `clamp(min=0).sqrt()` gradient concern

I checked it; it does not apply. `cos_pitch` is consumed only by `cos_pitch < 2.0 * ...`, and a comparison severs autograd on every torch version (`(cos_pitch < thr).requires_grad = False`, `grad_fn = None`, `dtype = torch.bool`), so the unbounded `sqrt'(0)` can never reach backward, on 2.5.1 or 2.14. Measured on 2.14 at the pole `(0.1, pi/2, 0.2)` float64, `d(roll+pitch+yaw)/dq`:

```
this PR:      (-0.14136, -2.82489, 0.0, 0.0)          all finite
origin/main:  (-4.81e+12, 1.2456e+16, 3.128e+15, 1.2612e+16)
```

and at the exact `atan2(0, 0)` quaternion `(0.5, 0.5, 0.5, -0.5)` the PR returns `(2.0, -2.0, 0.0, 0.0)` against `main`'s `(0, 0, 0, 0)` — no `nan`/`inf`, so the `torch.where` `0 * inf` trap does not fire (the `#4228` safe-asin guard is still load-bearing). **The PR improves the gradient at the pole by 16 orders of magnitude.**

One consequence to write down: `pitch` and `roll` now have gradient **exactly 0** at gimbal lock, where before they had a huge finite one. The `.. note::` saying *"pitch's gradient is finite at gimbal lock"* now understates it; please extend it in the same push.

## What is left to do

1. Replace the estimator with `cos_pitch = torch.hypot(sinr_cosp, cosr_cosp)` (threshold unchanged). Measured 0/500 at all four pole cells, no regression anywhere.
2. Revert the `test_wart_euler_from_quaternion_ignores_the_norm_3953` hunk — under `hypot` the triple returns to `main`'s existing literal, byte for byte.
3. Rewrite the threshold rationale comment: the `asin(1 - d) = pi/2 - sqrt(2d)` reasoning and the `1.0 / 2.0 / 8.0` sweep are all about the estimator being replaced.
4. Kill M8: add a pole cell whose input has `k >= 4` so the detector must fire, and a near-pole float32 cell that reds if the snap band widens. Give the `+pi/2` side a discriminating input too.
5. Restore the `dtype` fixture on both gimbal pins with a dtype-scaled tolerance (targets above).
6. Add `changelog.d/3993.fixed.md`.
7. Fix the PR description: it says `8·sqrt(eps)`, the code says `2.0`.
8. Merge `main` (currently CONFLICTING) and re-run CI — the green on this head predates several workflows, including the changelog gate.
9. Run the torch 2.5.1 overlay once `hypot` is in, so the floor is covered.

This is a good fix held up by one constant. `hypot` closes it and simplifies the PR at the same time.

ducha-aiki (changes requested):
Recording the existing current-head re-review as a formal request for changes. The linked review documents the remaining pole-detection failures, required regression coverage, changelog work, and merge conflicts. This records that completed review; it is not a new validation run.

[Completed review](https://github.com/kornia/kornia/pull/3993#issuecomment-5711959624).

ducha-aiki (changes requested):
## Re-review at `e17b239b3`

The `hypot` estimator does what we measured it would. Items 1, 2, 3, 6, 7 and 8 of my 2026-09-17 list are done:
- the estimator is replaced
- the `#3953` hunk is reverted
- the rationale comment is rewritten
- a fragment exists
- the PR description is corrected
- the PR merges cleanly with `main` (`8b03a8085`)

Item 5 is covered well enough by the new dtype-parametrized random test. I am dropping it.

Two regressions show up once the torch floor is exercised, and a third is visible on any torch. So this is still **REQUEST_CHANGES**. The library fix is below as a tested diff.

All numbers are from a merge tree with `main` (`kornia.__file__` checked), on CPU. They were run on torch 2.14.0 and on torch 2.5.1 through a `--target` overlay. The round-trip metric is `max|R(q) - R(q_back)|` with `q_back = quaternion_from_euler(*euler_from_quaternion(q))`, over 2000 random `(roll, yaw)` per cell. "broken" means `> 1e-4`.

### What is right (re-measured)

| pitch offset from ±pi/2 | main f32 (broken/max) | PR f32 | main f64 | PR f64 |
|---|---|---|---|---|
| 0 (pole) | 2000/2.0 | **0/3.0e-07** | 1999/2.0 | **0/1.1e-15** |
| 1e-8 | 2000/2.0 | 0/3.0e-07 | 0/6.5e-08 | 0/1.0e-08 |
| 1e-6 | 1997/0.42 | 0/1.4e-06 | 0/7.7e-10 | same |
| 1e-4 | 1813/4.2e-03 | 32/1.0e-04 | 0/7.7e-12 | same |
| 3e-4 | 1868/1.5e-03 | 2000/3.0e-04 | same | same |
| 7e-4 and above | identical | identical | identical | identical |

The `-pi/2` side matches. `q_back` equals `±q` to the same order. For `|pitch| < 1.5` the returned triples are **bit-identical to `main`** in f32 and f64. The PR's per-cell maximum is never worse than `main`'s. `tests/geometry/test_conversions.py` and `test_quaternion.py` at float32/float64 give 876 passed, 2 skipped, 7 xfailed. The doctests pass (56). The `#3950` closing grep has one `kornia/` hit, the implementation comment, which is fine.

## Blocks

### [P1] New NaN gradients on torch 2.5.1 for ordinary rotations (regression vs `main`)

`yaw_locked = ∓2 * x.atan2(w)` is evaluated for every element, not only at the pole. Wherever `w = x = 0`, `atan2(0, 0)` differentiates to `nan` on torch < 2.14. The `torch.where` then multiplies that by a zero cotangent and still produces `nan`. This is the same trap as the `clamp` note in AGENTS.md.

`w = x = 0` is a whole family of ordinary half-turns: every rotation by pi about an axis in the y-z plane, including the plain `yaw = pi` quaternion `(0, 0, 0, 1)`. Here is `d(roll + pitch + yaw)/dq`:

```
                     main 2.5.1                       PR 2.5.1
(0, 0, 0, 1)         [-2, -2, 2, 0]                   [nan, nan, 2, 0]
(0, 0, .6, .8)       [-0.4, -2.8, 2.75, 0.336]        [nan, nan, 2.75, 0.336]
(0, 0, -.6, .8)      [-2.8, -0.4, 2.75, -0.336]       [nan, nan, 2.75, -0.336]
```

The results are the same in f32 and f64. On 2.14 both trees give `main`'s values, which is why no CI leg on a recent torch sees this.

The same mechanism leaves the pole itself unfixed on 2.5.1. The roll and yaw `atan2(sinr_cosp, cosr_cosp)` is `atan2(0, 0)` exactly at a pole, and it is still differentiated even though its value is then replaced. So on 2.5.1 `(0.1, -pi/2, 0.2)` in f64 and `(.5, .5, .5, -.5)` in f32/f64 still give an all-`nan` gradient, and `gradcheck` at `-pi/2` still fails. On 2.14 the PR fixes those. This part is not a regression, because `main` is `nan` there too. The fix below closes it at no extra cost.

### [P1] `yaw` leaves `[-pi, pi]` at gimbal lock, and `q` and `-q` disagree

`2 * atan2(x, w)` ranges over `(-2pi, 2pi]`. At the pole, over 2000 unit quaternions (f64):

```
+pi/2:  yaw in [-6.17, 6.15], outside [-pi, pi] for  494/2000; for -q: 1506/2000
-pi/2:  yaw in [-6.04, 6.10], outside [-pi, pi] for  510/2000; for -q: 1490/2000
q = (-0.5, 0.5, -0.5, -0.5)  ->  (0.0, 1.5708, -4.7124)
```

On `main`, every returned angle comes from a plain `atan2` and lies in `[-pi, pi]`. The rotation is still right, but consumers that wrap or bin yaw now see a different range, and only at the pole. The yaw also depends on which of `q` / `-q` was passed. That is a new inconsistency, since everywhere else the function is sign-invariant.

A surviving mutant confirms that nothing pins this: adding `+ 2*pi` to the `+pi/2` yaw branch passes the whole file.

### Tested fix for both

The fix has three parts:
- Compute `gimbal` before the two `atan2` calls.
- Give every `atan2` a safe argument outside its own region.
- Take the locked yaw as the `atan2` of the double-angle pair. That is `2 * atan2(x, w)` wrapped into `[-pi, pi]`, and it is invariant under `q -> -q`.

The values are unchanged everywhere else.

```diff
     sinr_cosp = 2.0 * (w * x + y * z)
     cosr_cosp = 1.0 - 2.0 * (x * x + yy)
-    roll = sinr_cosp.atan2(cosr_cosp)
+    # cos(pitch) is measured as hypot(sinr_cosp, cosr_cosp) -- see the gimbal-lock block below.
+    gimbal = torch.hypot(sinr_cosp, cosr_cosp) < 2.0 * torch.finfo(w.dtype).eps ** 0.5
+    zero = torch.zeros_like(w)
+    one = torch.ones_like(w)
+    # roll's atan2 is (0, 0) at a pole and its value is replaced there anyway; substitute a safe
+    # argument so its gradient, which is nan at the origin on torch < 2.14, never reaches backward.
+    roll = torch.where(gimbal, zero, sinr_cosp).atan2(torch.where(gimbal, one, cosr_cosp))
@@
     siny_cosp = 2.0 * (w * z + x * y)
     cosy_cosp = 1.0 - 2.0 * (yy + z * z)
-    yaw = siny_cosp.atan2(cosy_cosp)
+    yaw = torch.where(gimbal, zero, siny_cosp).atan2(torch.where(gimbal, one, cosy_cosp))
@@
-    cos_pitch = torch.hypot(sinr_cosp, cosr_cosp)
-    gimbal = cos_pitch < 2.0 * torch.finfo(w.dtype).eps ** 0.5
     up = sinp > 0.0
@@
-    yaw_locked = torch.where(up, -2.0 * x.atan2(w), 2.0 * x.atan2(w))
+    # 2 * atan2(x, w) as the atan2 of its double-angle pair, which keeps yaw in [-pi, pi] and is
+    # invariant under q -> -q; (x, w) is (0, 0) only off the pole, where a safe argument is used.
+    x_locked = torch.where(gimbal, x, zero)
+    w_locked = torch.where(gimbal, w, one)
+    locked = (2.0 * x_locked * w_locked).atan2(w_locked * w_locked - x_locked * x_locked)
+    yaw_locked = torch.where(up, -locked, locked)
     pitch = torch.where(gimbal, pitch_locked, pitch)
-    roll = torch.where(gimbal, torch.zeros_like(roll), roll)
+    roll = torch.where(gimbal, zero, roll)
```

Here is what I measured with this diff applied to the merge tree:
- **Round trip:** the pole/near-pole table above is unchanged, and the pole maximum even improves slightly (f32 1.9e-07, f64 8.9e-16). General-pitch triples are still bit-identical to `main`.
- **Gradients:** all 18 probes are finite on **both** 2.5.1 and 2.14, in f32 and f64. The `w = x = 0` cases return `main`'s exact values.
- **gradcheck:** f64 gradcheck passes at `+pi/2`, at `-pi/2`, at `pi/2 - 1e-2` and at `(0.3, 0.7, 1.1)`, on both torch versions. `main` fails at both poles on both versions.
- **Yaw range:** 0/2000 yaw values fall outside `[-pi, pi]` at either pole in f32 or f64, and `q` and `-q` give the same yaw.
- **Tests:** test_conversions + test_quaternion give 876 passed. The doctests pass.

### [P2] Surviving non-equivalent mutants on the threshold (item 4 from last round, still open)

I ran 9 one-token mutants against test_conversions + test_quaternion at float32/float64:
- **Killed:** factor → 0, the old `sqrt(1 - sinp²)` estimator, `up` flipped, yaw-branch signs swapped, the pitch snap dropped, the roll pin dropped.
- **Survived:** factor `2.0 → 8.0`, factor `2.0 → 100.0`, and yaw `+ 2pi` (see above).

The factor mutants are not equivalent. In f32 at `pitch = ±(pi/2 - 1e-3)`, 200 draws, the round-trip maximum is:

```
factor 2.0: 2.9e-04     factor 4.0: 1.0e-03     factor 8.0: 1.0e-03 (and 2.0e-03 at pi/2 - 2e-3, vs 1.4e-04)
```

Widening the snap band makes resolvable f32 pitches about 3.5x worse, and the suite does not notice. Please add:
- **A near-pole f32 pin.** 200 random `(roll, yaw)` at `pitch = ±(pi/2 - 1e-3)`, asserting a maximum rotation error below `5e-4`. That kills factor 4 and above. An f64 analogue at `±(pi/2 - 5e-8)` is optional.
- **A range assertion in `test_convention_roundtrip_holds_at_gimbal_lock_random_roll_yaw_3993`.** Assert that `yaw` (and `roll`) lie in `[-pi, pi]`, and that `euler_from_quaternion(*(-q))` returns the same triple. This kills the `+2pi` mutant and pins the fix above.
- **A torch-floor gradient pin.** Assert that `d(roll + pitch + yaw)/dq` is finite at `(0, 0, 0, 1)` and at the `-pi/2` pole. It only discriminates on the 2.5.1 leg, but that leg is where the regression lives.

### [P2] The fragment describes this PR's own history, not the change against `main`

`changelog.d/3993.fixed.md` says "the previous detector recomputed cos(pitch) as sqrt(1 - sin(pitch)**2) ... missed 2-4% of poles". There is no previous detector. `main` has no gimbal-lock branch at all and gets 2000/2000 poles wrong. Users need the before and after as they will see it. Suggested text:

> `euler_from_quaternion` now returns a triple that represents the input rotation at gimbal lock (`pitch = ±pi/2`). There, `roll` and `yaw` used to come from `atan2(0, 0)` and the round trip through `quaternion_from_euler` was off by up to 2 in rotation-matrix entries. Now `pitch` is returned as exactly `±pi/2`, `roll` as `0`, and the one determined combined angle is returned in `yaw`, within `[-pi, pi]`. The gradient there is finite and small, where before it was of order `1e7`–`1e16`.

### [P2] Docstring left stale by the change

- **The Convention block does not say what happens at the pole.** It covers only "away from `|pitch| = pi/2`", and the old warning, which was the only prose about the pole, is gone. Please add a bullet: at `|pitch| = pi/2` (detected as `cos(pitch) < 2*sqrt(eps)`), `pitch` is exactly `±pi/2`, `roll` is `0`, and `yaw` carries `yaw - roll` at `+pi/2` or `yaw + roll` at `-pi/2`, in `[-pi, pi]`. Also say that the input triple itself cannot be recovered there.
- **The `.. note::` says "separate concern from the two warnings above".** Only one warning remains. It also still says only that "pitch's gradient is finite at gimbal lock". Inside the gimbal branch, `roll` and `pitch` now have gradient exactly `0`, and all the gradient is in `yaw` (measured: `(-0.141, -2.82, 0, 0)` at `(0.1, pi/2, 0.2)`). I asked for this last round.
- **The `#3953` warning says the scaled input "saturates the `asin` and is reported as gimbal-locked".** Now that a gimbal-lock branch exists, that reads as if the branch fires, and it does not: the `2q` triple keeps `roll = 1.656`, which is `main`'s literal. Suggested wording: "saturates the `asin`, so `pitch` reads exactly `pi/2` although the gimbal-lock branch does not fire".

## Not blocking, recorded

- **Small-norm inputs.** `cosr_cosp = 1 - 2(x² + y²)` is not homogeneous in `q`, so `hypot` does not shrink with `|q|`. Scaled quaternions down to `1e-3 * q` give the same triples as `main`, with no spurious gimbal detection.
- **The detection comparison.** `hypot` is only consumed by the comparison, so its `0/0` backward at the origin cannot reach autograd.

Issue #3930, opened 2026-08-09
CI: tests.yml pytorch-version matrix leg is never consumed — all legs run the uv-locked torch


ducha-aiki:
Found during review of #3926 (confirmed from the Actions log, not inferred; predates that PR and is independent of it).

`.github/workflows/tests.yml` builds a `matrix.pytorch-version` axis, but no installation or test step consumes it. The job labelled

```
tests (ubuntu-latest, float32, 3.12, 2.5.1)
```

actually logged

```
uv sync --extra dev
+ torch==2.9.1
main deps:
    - torch-2.9.1+cu128
```

Evidence: https://github.com/kornia/kornia/actions/runs/31322346689/job/93267224581

So the nominal 2.5.1 and 2.9.1 matrix legs currently test the **same** uv-locked torch 2.9.1 environment — the compatibility coverage advertised by the job names does not exist. Fix direction: either thread `matrix.pytorch-version` into the install step (e.g. `uv sync` followed by `uv pip install "torch==${{ matrix.pytorch-version }}"` with an index appropriate to the runner), or drop the axis so job names stop advertising coverage that isn't run.

Flagging as high-priority per the reviewer: torch-version compatibility is a stated support surface (README/CI badges) and downstream pins depend on it.

—
*Issue authored by Claude (Claude Code) on behalf of @ducha-aiki.*

PR #3995, opened 2026-08-22
ci: install the selected PyTorch matrix version

Parshav14:
## Which lane?
- [x] 🟢 **Green lane**
- [ ] 🟠 **Discuss-first**

**Fixes:** #3930

## What does this PR do?

The CPU CI workflow defines multiple PyTorch versions in its matrix, but the
installation step previously used the single PyTorch version pinned by
`uv.lock`.

As a result, jobs labelled with different PyTorch versions were actually
running against the same installed version.

This change makes the workflow install the PyTorch version selected by the
matrix, verifies the imported runtime version, and prevents the test command
from resynchronizing the environment from the lockfile.

The change exposes existing compatibility failures in the PyTorch 2.5.1
matrix leg that were previously hidden because those jobs were actually
running with PyTorch 2.9.1.

## Test log

```text
$ pixi run lint

$ pixi run -e py312 uv run --no-sync uv pip install --dry-run --reinstall 'torch==2.5.1'
Would install torch==2.5.1

$ pixi run -e py312 uv run --no-sync uv pip install --dry-run --reinstall 'torch==2.9.1'
Would install torch==2.9.1

Runtime version check:
expected 2.9.1, got 2.9.1

### DIFF
--- a/.github/workflows/scheduled_test_cpu.yml
+++ b/.github/workflows/scheduled_test_cpu.yml
@@ -97,7 +97,7 @@ jobs:
         id: hashid
         run: echo "weights-hash=${{ hashFiles('.github/download-models-weights.py') }}" >> "$GITHUB_OUTPUT"
 
-  tests-cpu-ubuntu:
+  tests-cpu-ubuntu-py311:
     needs: [check-skip, pre-tests]
     if: needs.check-skip.outputs.should-skip == 'false'
     strategy:
@@ -108,8 +108,48 @@ jobs:
     uses: ./.github/workflows/tests.yml
     with:
       os: "ubuntu-latest"
-      python-version: '["3.11", "3.12", "3.13"]'
+      python-version: '["3.11"]'
       pytorch-version: '["2.1.2", "2.2.2", "2.3.1", "2.4.0", "2.5.1", "2.9.1"]'
+      pytorch-dtype: ${{ matrix.pytorch-dtype }}
+      cache-path: weights/
+      cache-key: model-weights-${{ needs.pre-tests.outputs.hash }}
+      cache-restore-keys: |
+        model-weights-${{ needs.pre-tests.outputs.hash }}
+        model-weights-
+
+  tests-cpu-ubuntu-py312:
+    needs: [check-skip, pre-tests]
+    if: needs.check-skip.outputs.should-skip == 'false'
+    strategy:
+      fail-fast: false
+      matrix:
+        pytorch-dtype: ["float32", "float64"]
+
+    uses: ./.github/workflows/tests.yml
+    with:
+      os: "ubuntu-latest"
+      python-version: '["3.12"]'
+      pytorch-version: '["2.2.2", "2.3.1", "2.4.0", "2.5.1", "2.9.1"]'
+      pytorch-dtype: ${{ matrix.pytorch-dtype }}
+      cache-path: weights/
+      cache-key: model-weights-${{ needs.pre-tests.outputs.hash }}
+      cache-restore-keys: |
+        model-weights-${{ needs.pre-tests.outputs.hash }}
+        model-weights-
+
+  tests-cpu-ubuntu-py313:
+    needs: [check-skip, pre-tests]
+    if: needs.check-skip.outputs.should-skip == 'false'
+    strategy:
+      fail-fast: false
+      matrix:
+        pytorch-dtype: ["float32", "float64"]
+
+    uses: ./.github/workflows/tests.yml
+    with:
+      os: "ubuntu-latest"
+      python-version: '["3.13"]'
+      pytorch-version: '["2.5.1", "2.9.1"]'
       pytorch-dtype: ${{ matrix.pytorch-dtype }}
       cache-path: weights/
       cache-key: model-weights-${{ needs.pre-tests.outputs.hash }}
--- a/.github/workflows/scheduled_test_cpu.yml
+++ b/.github/workflows/scheduled_test_cpu.yml
@@ -10,6 +10,7 @@ on:
       - 'COPYRIGHT'
   schedule:
     - cron: '0 4 * * *'
+  workflow_dispatch:
   workflow_call:
 
 permissions:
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

**Review by ducha-aiki — DISMISSED** [review]
Thank you, @Parshav14 — this is a useful and well-targeted CI fix. The red PyTorch 2.5.1 legs confirmed that the workflow now installs the matrix-selected version and exposed compatibility gaps that the previous lockfile-resynchronized jobs had hidden.

Maintainer follow-up: I pushed commit 5bf32ff5 to the PR branch to handle those newly visible gaps. It capability-gates torch.export coverage before the public export-tracking API exists, skips only the unavailable pre-2.6 CPU half-precision grid_sample kernels, fixes the compile warmup for ordinary jobs, replaces enum-set membership with a Dynamo-friendly equivalent, and splits scheduled Ubuntu coverage into installable Python/PyTorch combinations.

The focused export and augmentation tests pass on PyTorch 2.9.1, the full pre-commit suite passes, and an independent review found no actionable issues. The refreshed CI run will provide the final PyTorch 2.5.1 confirmation.

Approved. Thanks again for identifying and fixing the original matrix problem.

**Comment by ducha-aiki** [comment]
## Review

Thanks for taking this on — the diagnosis in #3930 is right and the core mechanism you've built is the correct one. `tests.yml` really did declare a `pytorch-version` axis that nothing consumed, and the install + verify + `UV_NO_SYNC` trio closes that hole in a way that can't silently regress. The verify step in particular is a nice touch: `expected 2.5.1, got 2.5.1+cu124` in the log is exactly the evidence that was missing before.

I checked the diff against the current CI runs and against upstream PyTorch sources. The mechanism is sound, but **the PR is red right now**, and making the matrix real exposes three further compatibility problems that aren't handled yet. Details below, most important first.

---

### 1. Blocker — the PR is currently failing CI

Run [33091899446](https://github.com/kornia/kornia/actions/runs/33091899446) on `5bf32ff5`:

```
FAILED tests/geometry/test_conversions.py::test_pixel_coordinate_export_crosses_singleton_boundary[...]   (x10)
FAILED tests/geometry/test_grid.py::test_meshgrid_export_crosses_singleton_boundary[...]                   (x4)
  -> RuntimeError: Python 3.13+ not yet supported for torch.compile
= 14 failed, 8229 passed, 3547 skipped, 43 xfailed, 11 xpassed in 132.86s =
```

Only `ubuntu-latest / 3.13 / 2.5.1` fails; `3.11 / 2.5.1` and `3.12 / 2.5.1` are green on all three OSes. The cause is that torch 2.5.1's Dynamo has no Python 3.13 support (it landed in torch 2.6), and `torch.export.export` goes through Dynamo — so every `torch.export`-based test dies on that one cell.

Two notes on this:

* These tests came in with `4dd9c6aa` (#4006), which landed on `main` after you branched. Your branch is 8 commits behind `main`; CI tests the merge result, which is why they show up here and not in your 2026-08-22 run. **Please rebase on `main`** before the next push.
* This is the same class of problem your `_HAS_TORCH_EXPORT_TRACKING` skip already solves for `tests/augmentation/test_torch_export.py` — it just doesn't cover these two geometry files. The augmentation gate (`torch.compiler.is_exporting`, torch ≥ 2.7 — I confirmed that's the first tag where `torch/compiler/__init__.py` defines it) would over-skip here, since these tests genuinely pass on 2.5.1 under 3.11/3.12. A narrower predicate — "Dynamo is unavailable on this interpreter", e.g. `torch_version_lt(2, 6, 0) and sys.version_info >= (3, 13)` — hoisted into a shared helper in `testing/` and applied to all three files would be the cleanest fix and would keep the 8229 passing tests on that cell.

### 2. Blocker — `torch.compiler.is_compiling` does not exist before torch 2.3

The py311 leg now really installs **2.1.2** and **2.2.2**. `torch.compiler.is_compiling` is called unguarded in 18 places on `main`:

```
kornia/filters/gaussian.py:103
kornia/geometry/conversions.py:1499,1504,1579,1588,1658,1667,1747,1756,2126,2131,2240,2245
kornia/geometry/grid.py:68,86,147,158
```

Verified against upstream: `torch/compiler/__init__.py` has no `def is_compiling` at `v2.1.2` or `v2.2.2`, and does at `v2.3.1`. So `gaussian_blur2d`, `create_meshgrid`, `normalize_pixel_coordinates` and everything layered on them (`RandomGaussianBlur`, Canny, the scale-space pyramids, most of `tests/filters`) will raise `AttributeError: module 'torch.compiler' has no attribute 'is_compiling'` and the 2.1.2 / 2.2.2 legs of the nightly will go broadly red.

This is pre-existing — all of it was written during the window when the matrix was fake, so it has never actually run on those versions. That's precisely the bug you're fixing, but it means the fix can't land without either guarding the calls (`kornia/core/utils.py:359` already has the right pattern for `is_exporting`: resolve `getattr(torch.compiler, "is_compiling", None)` once at import) or raising the matrix floor.

### 3. Blocker — `uv.lock` pins `numpy==2.4.0`, which torch ≤ 2.4 can't use

`uv pip install "torch==2.1.2"` repins only torch; torch declares no numpy upper bound, so numpy 2.4.0 stays. torch wheels through 2.4.0 are built against the NumPy 1.x C ABI, so the first `torch.from_numpy` / `.numpy()` in the suite raises `A module that was compiled using NumPy 1.x cannot be run in NumPy 2.x`. `transformers` and `diffusers` are resolved for modern torch too and are imported by `tests/models`. The old-torch legs need a matching constraint set, not a bare torch repin.

### 4. The whole scheduled-matrix rewrite is unverifiable before merge

`scheduled_test_cpu.yml` triggers on `push: main`, `schedule` and `workflow_call` only — there's no `workflow_dispatch`. The py311/py312/py313 split, and every leg below 2.5.1 (where problems 2 and 3 live), will therefore run for the first time *after* this merges. Please add `workflow_dispatch:` to that workflow and post a green manual run, or drop the sub-2.5.1 legs from this PR.

Given 2 and 3, my suggestion is to **split this**: land `tests.yml` (install + verify + `UV_NO_SYNC`), the 2.5.1 compat fixes, and a matrix floor of 2.5.1 — that's exactly the surface your PR CI already covers and it's a real, verifiable improvement. Then handle 2.1.2–2.4.0 in a follow-up that fixes `is_compiling` and the numpy constraint first.

---

### 5. `--reinstall` re-downloads ~2.9 GB of CUDA wheels per job

From the install step's own log:

```
Resolved 22 packages in 684ms
Downloading nvidia-cudnn-cu12 (634.0MiB)
Downloading nvidia-cublas-cu12 (346.6MiB)
Downloading nvidia-cufft-cu12  (201.7MiB)
Downloading triton             (199.8MiB)
Downloading torch              (864.5MiB)
... 13 wheels, ~2.9 GB
Prepared 22 packages in 23.31s
```

Two separate issues:

* `--reinstall` implies `--refresh`, so the uv cache is bypassed and the entire dependency closure is re-fetched — **including on the `2.9.1` legs where the installed torch is already correct**. `uv pip install "torch==X"` alone is a no-op when satisfied and downgrades cleanly when not; `--reinstall-package torch` if you want to force it.
* These are CPU-only runners. The default PyPI wheel is the CUDA build (`2.5.1+cu124`), dragging the whole `nvidia-*` stack plus `triton` for nothing. `--index-url https://download.pytorch.org/whl/cpu` cuts it to ~180 MB and is what #3930 meant by "an index appropriate to the runner". Your `startswith` check still passes (`2.5.1+cpu`).

Across ~40 legs a night this is a meaningful bandwidth and disk-pressure cost on 14 GB runners.

### 6. The `conftest.py` guard is still wrong for 3.13

Gating warm-up on `KORNIA_TEST_OPTIMIZER` is a good change on its own — `pytest_collection_modifyitems` (conftest.py:169) already deselects `dynamo`/`compile` tests when it's empty, so the warm-up was pure overhead. (And `_setup_torch_compile` never calls the compiled callables, so it wasn't warming anything anyway.)

But it only hides the underlying bug. The original crash was:

```
File "conftest.py", line 287, in _setup_torch_compile
    torch.compile(_dummy_fn)
RuntimeError: Dynamo is not supported on Python 3.13+
```

The `except RuntimeError` matches the literal `"Dynamo is not supported on Python 3.12+"`, so the 3.13 message escapes and takes the session down with `INTERNALERROR` / exit 3. The `dynamo` job is pinned to 3.11 / 2.9.1 today so it's latent, but the moment an optimizer leg runs on 3.13 with torch < 2.6 it aborts again. Please widen the match to `"Dynamo is not supported on Python"` while you're in there.

### 7. Smaller things

* **`kornia/augmentation/container/augment.py:42-46`** — the `set` → `tuple` change is behaviour-preserving (all five constants are only used for `in` tests at lines 369–423), so no objection. But nothing exercises it any more: it was presumably for the `Unsupported: call_method UserDefinedObjectVariable(set) __contains__` failure on 2.5.1, and `test_torch_export_container` is now skipped below 2.7 by this same PR, with no 2.6/2.7/2.8 leg in the matrix. Either drop it or add a line saying why it stays — a library-source change inside a CI PR needs a reason in the commit message.
* **`scheduled_test_cpu.yml:100-158`** — three near-identical 20-line jobs. The Windows and macOS jobs in the same file express the same thing with one matrix plus `exclude:`; matching that style would be ~40 lines shorter and keeps the four blocks consistent.
* **PR description** — it covers the workflow change only. It doesn't mention `conftest.py`, `kornia/augmentation/container/augment.py`, or the two test files, and the "Test log" section has `pixi run lint` and two `--dry-run` installs but no `pixi run test` output, which CONTRIBUTING requires as proof of execution. Also `5bf32ff5` has no `Signed-off-by` line while `b05261ad` does — worth making consistent.

### Verified, no action needed

* The `test_mixed_image_bbox_dtypes` skip boundary is exactly right — `GridSamplerKernel.cpp` dispatches `AT_DISPATCH_FLOATING_TYPES` (no Half/BFloat16) at 2.5.1 and `AT_DISPATCH_FLOATING_TYPES_AND2(kHalf, kBFloat16, ...)` from 2.6.0. Both dtypes and the `< 2.6` cutoff check out.
* Wheel availability per matrix cell is correct: 2.1.2 has no cp312, 2.2.2–2.4.0 have no cp313, and 2.5.1 *does* ship a cp313 manylinux wheel — so keeping 2.5.1 in the ubuntu py313 list is right even though Windows/macOS exclude it.
* The docstring rewrite in `test_torch_export.py` is a genuine correction: `main`'s version claimed containers weren't covered while `test_torch_export_container` existed and passed.
* Renaming the scheduled ubuntu jobs doesn't break anything — `main`'s required checks are only `["pre-commit.ci - pr", "collector"]`, and `collector` lives in `pr_test_cpu.yml`, which is untouched.

---

**Summary:** right problem, right mechanism, and the verify step is a genuinely good addition. Blocked on the red 3.13/2.5.1 leg (§1), and on §2/§3 before any sub-2.5.1 leg can be turned on for real. Happy to look again as soon as it's rebased.

ducha-aiki (approved):
Thank you!

Issue #3962, opened 2026-08-10
Missing counterparts: no denormalize_homography3d and no ColmapQTVecs_to_ARKitQTVecs (scope question)


ducha-aiki:
Two conversions in `kornia/geometry/conversions.py` are one-way only, so a pipeline that uses them cannot get back to where it started through the public API.

- **No `denormalize_homography3d`.** The 2D pair is complete (`normalize_homography` / `denormalize_homography`), the 3D one is not — `normalize_homography3d` exists with no inverse.
- **No `ColmapQTVecs_to_ARKitQTVecs`.** `ARKitQTVecs_to_ColmapQTVecs` converts a camera-to-world graphics-frame pose into a world-to-camera vision-frame pose; there is no function for the reverse.

## Repro

```python
import kornia.geometry.conversions as C

print("homography entries in __all__:", sorted(n for n in C.__all__ if "homography" in n))
print("hasattr denormalize_homography3d :", hasattr(C, "denormalize_homography3d"))
print("ARKit/Colmap entries in __all__  :", [n for n in C.__all__ if "Colmap" in n or "ARKit" in n])
print("hasattr ColmapQTVecs_to_ARKitQTVecs:", hasattr(C, "ColmapQTVecs_to_ARKitQTVecs"))
```

## Observed

```
homography entries in __all__: ['convert_affinematrix_to_homography', 'convert_affinematrix_to_homography3d', 'denormalize_homography', 'normalize_homography', 'normalize_homography3d']
hasattr denormalize_homography3d : False
ARKit/Colmap entries in __all__  : ['ARKitQTVecs_to_ColmapQTVecs']
hasattr ColmapQTVecs_to_ARKitQTVecs: False
```

## Expected

Both are mechanical given what already exists:

- `denormalize_homography3d(H, dsize_src, dsize_dst) = N_dst⁻¹ @ H @ N_src` with `N = normal_transform_pixel3d(d, h, w)` — the exact mirror of `normalize_homography3d`, and structurally identical to the existing `denormalize_homography`. Both compositions are already verified by execution in the batch-3 audit (`normalize_homography3d = N_dst @ H @ N_src⁻¹`, match `0.0`).
[...]

Issue #3960, opened 2026-08-10
normalize_homography's shape guard uses 'or': (3,3) and (2,4,3,3) silently pass while (B,4,4) escapes and dies inside matmul


ducha-aiki:
The shape guard in `normalize_homography` / `denormalize_homography` (and its copy in `normalize_homography3d`) uses `or` where it needs `and`, so it validates almost nothing:

```python
# kornia/geometry/conversions.py:1278 (normalize_homography / denormalize_homography)
if not (len(dst_pix_trans_src_pix.shape) == 3 or dst_pix_trans_src_pix.shape[-2:] == (3, 3)):
    raise ValueError(f"Input dst_pix_trans_src_pix must be a Bx3x3 tensor. Got {dst_pix_trans_src_pix.shape}")

# kornia/geometry/conversions.py:1431 (normalize_homography3d) — same structure, 4x4
if not (len(dst_pix_trans_src_pix.shape) == 3 or dst_pix_trans_src_pix.shape[-2:] == (4, 4)):
    raise ValueError(f"Input dst_pix_trans_src_pix must be a Bx3x3 tensor. Got {dst_pix_trans_src_pix.shape}")
```

Anything with rank 3 passes regardless of its trailing shape, and anything with the right trailing shape passes regardless of its rank. Two consequences:

- **Accidental extra support**, which happens to work: unbatched `(3, 3)` is accepted and promoted to `(1, 3, 3)`, and arbitrary leading batch dims like `(2, 4, 3, 3)` are accepted and preserved. The docstring documents only `(B, 3, 3)`, and no test pins either behavior.
- **A hole**: a `(B, 4, 4)` input (rank 3, wrong trailing shape) sails through the guard and dies several frames later inside `matmul`, with a message that names neither the argument nor the expectation.

`normalize_homography3d` additionally reports **`Bx3x3`** for a 4×4-homography function.

## Repro

```python
import torch
from kornia.geometry.conversions import denormalize_homography, normalize_homography, normalize_homography3d

print("# normalize_homography, documented as (B, 3, 3):")
for shape in [(3, 3), (2, 3, 3), (2, 4, 3, 3), (1, 4, 4), (3, 4), (2, 2, 2)]:
    try:
        out = tuple(normalize_homography(torch.eye(3).expand(*shape[:-2], 3, 3).contiguous()
                                         if shape[-2:] == (3, 3) else torch.zeros(*shape), (4, 5), (4, 5)).shape)
[...]

PR #3996, opened 2026-08-22
feat(geometry): add missing reverse conversions

Parshav14:
## Which lane?

- [ ] 🟢 **Green lane**
- [x] 🟠 **Discuss-first** — maintainer scope was confirmed on #3962.

**Fixes:** #3962

## What does this PR do?

Adds the two reverse conversions discussed in #3962:

- `denormalize_homography3d`, the inverse of `normalize_homography3d`.
- `ColmapQTVecs_to_ARKitQTVecs`, the reverse of `ARKitQTVecs_to_ColmapQTVecs`.

The implementation reuses the existing conversion utilities and follows the
conventions already used in `kornia.geometry.conversions`.

The tests cover round-trip behavior, asymmetric 3D source and destination
sizes, batching, output shapes, validation cases, dtype/device combinations,
and quaternion rotation equivalence without relying on a particular
quaternion sign.

The public API exports and geometry conversion documentation were updated for
both functions.

## Test log

```text
$ pixi run lint

$ pixi run typecheck
All checks passed!

Focused homography tests:
9 passed

Focused ARKit/Colmap tests:
9 passed

$ tests/test_api_surface.py
12 passed

$ tests/geometry/transform/test_homography_warper.py
101 passed

$ tests/geometry/test_conversions.py
286 passed, 22 xfailed

$ doctest kornia/geometry/conversions.py
32 passed

$ git diff --check
Passed
```

## AI Usage Disclosure

- [ ] 🟢 **Human-written**
- [x] 🟡 **AI-assisted**
- [ ] 🔴 **AI-generated**

### DIFF
--- a/docs/source/geometry.conversions.rst
+++ b/docs/source/geometry.conversions.rst
@@ -37,6 +37,7 @@ Homography
 .. autofunction:: normalize_homography
 .. autofunction:: denormalize_homography
 .. autofunction:: normalize_homography3d
+.. autofunction:: denormalize_homography3d
 
 Quaternion
 ----------
@@ -82,3 +83,4 @@ Pose
 .. autofunction:: camtoworld_graphics_to_vision_Rt
 .. autofunction:: camtoworld_vision_to_graphics_Rt
 .. autofunction:: ARKitQTVecs_to_ColmapQTVecs
+.. autofunction:: ColmapQTVecs_to_ARKitQTVecs
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -29,6 +29,7 @@ from kornia.core.utils import _inverse_3x3_closed_form, _torch_inverse_cast
 
 __all__ = [
     "ARKitQTVecs_to_ColmapQTVecs",
+    "ColmapQTVecs_to_ARKitQTVecs",
     "Rt_to_matrix4x4",
     "angle_axis_to_quaternion",
     "angle_axis_to_rotation_matrix",
@@ -47,6 +48,7 @@ __all__ = [
     "convert_points_to_homogeneous",
     "deg2rad",
     "denormalize_homography",
+    "denormalize_homography3d",
     "denormalize_pixel_coordinates",
     "denormalize_pixel_coordinates3d",
     "denormalize_points_with_intrinsics",
@@ -1875,6 +1877,41 @@ def normalize_homography3d(
     return dst_norm_trans_src_norm
 
 
+def denormalize_homography3d(
+    dst_pix_trans_src_pix: torch.Tensor, dsize_src: tuple[int, int, int], dsize_dst: tuple[int, int, int]
+) -> torch.Tensor:
+    r"""De-normalize a given 3D homography from [-1, 1] to actual depth, height and width.
+
+    Args:
+        dst_pix_trans_src_pix: homography/ies from source to destination to be
+          denormalized. :math:`(B, 4, 4)`
+        dsize_src: size of the source image (depth, height, width).
+        dsize_dst: size of the destination image (depth, height, width).
+
+    Returns:
+        the denormalized homography of shape :math:`(B, 4, 4)`.
+
+    """
+    if not isinstance(dst_pix_trans_src_pix, torch.Tensor):
+        raise TypeError(f"Input type is not a torch.Tensor. Got {type(dst_pix_trans_src_pix)}")
+
+    if not (len(dst_pix_trans_src_pix.shape) == 3 or dst_pix_trans_src_pix.shape[-2:] == (4, 4)):
+        raise ValueError(f"Input dst_pix_trans_src_pix must be a Bx4x4 tensor. Got {dst_pix_trans_src_pix.shape}")
+
+    # source and destination sizes
+    src_d, src_h, src_w = dsize_src
+    dst_d, dst_h, dst_w = dsize_dst
+
+    # compute the transformation pixel/norm for src/dst
+    src_norm_trans_src_pix: torch.Tensor = normal_transform_pixel3d(src_d, src_h, src_w).to(dst_pix_trans_src_pix)
+    dst_norm_trans_dst_pix: torch.Tensor = normal_transform_pixel3d(dst_d, dst_h, dst_w).to(dst_pix_trans_src_pix)
+    dst_pix_trans_dst_norm = _torch_inverse_cast(dst_norm_trans_dst_pix)
+
+    # compute chain transformations
+    dst_pix_trans_src_pix = dst_pix_trans_dst_norm @ (dst_pix_trans_src_pix @ src_norm_trans_src_pix)
+    return dst_pix_trans_src_pix
+
+
 def normalize_points_with_intrinsics(point_2d: torch.Tensor, camera_matrix: torch.Tensor) -> torch.Tensor:
     """Normalize points with intrinsics. Useful for conversion of keypoints to be used with essential matrix.
 
@@ -2221,6 +2258,28 @@ def ARKitQTVecs_to_ColmapQTVecs(qvec: torch.Tensor, tvec: torch.Tensor) -> tuple
     t_colmap = t_colmap.reshape(-1, 3, 1)
     q_colmap = rotation_matrix_to_quaternion(R_colmap.contiguous())
     return q_colmap, t_colmap
+
+
+def ColmapQTVecs_to_ARKitQTVecs(qvec: torch.Tensor, tvec: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+    r"""Convert Colmap world-to-camera pose to the Apple ARKit screen pose.
+
+    Both poses are in quaternion representation.
+
+    Args:
+        qvec: Colmap rotation quaternion :math:`(B, 4)`, [w, x, y, z] format.
+        tvec: translation vector :math:`(B, 3, 1)`, [x, y, z]
+
+    Returns:
+        qvec: ARKit rotation quaternion :math:`(B, 4)`, [w, x, y, z] format.
+        tvec: translation vector :math:`(B, 3, 1)`, [x, y, z]
+
+    """
+    R_colmap = quaternion_to_rotation_matrix(qvec)
+    Rcv, Tcv = worldtocam_to_camtoworld_Rt(R_colmap, tvec)
+    Rcg, Tcg = camtoworld_vision_to_graphics_Rt(Rcv, Tcv)
+    t_arkit = Tcg.reshape(-1, 3, 1)
+    q_arkit = rotation_matrix_to_quaternion(Rcg.contiguous())
+    return q_arkit, t_arkit
 
 
 def vector_to_skew_symmetric_matrix(vec: torch.Tensor) -> torch.Tensor:
--- a/tests/geometry/test_conversions.py
+++ b/tests/geometry/test_conversions.py
@@ -29,6 +29,7 @@ from kornia.core._compat import torch_version
 from kornia.core.ops import eye_like
 from kornia.geometry.conversions import (
     ARKitQTVecs_to_ColmapQTVecs,
+    ColmapQTVecs_to_ARKitQTVecs,
     Rt_to_matrix4x4,
     axis_angle_to_rotation_matrix,
     camtoworld_graphics_to_vision_4x4,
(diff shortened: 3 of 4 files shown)


ducha-aiki:
## Review — changes requested, but the direction is right

**Verdict: changes requested.** The math is correct, the API shape is right, and none of what follows is a redesign — it is four concrete edits plus a rebase. Thanks for taking this one on.

### What is already right

- Scope was confirmed on the issue before any code was written (#3962, [comment](https://github.com/kornia/kornia/issues/3962#issuecomment-5379823963)) — that is exactly the process for a new public API.
- `__all__`, `docs/source/geometry.conversions.rst` and the tests were all updated together; `tests/test_api_surface.py` passes (12 passed), so the export surface is consistent with the siblings (`kornia.geometry.<name>`, not top-level — same as `denormalize_homography`).
- **The math is correct**, and I verified it rather than read it. In `float64`, against the hand-written closed form on random `(2, 4, 4)` inputs with asymmetric sizes `(3, 5, 7) -> (6, 8, 11)`:

  ```
  closed-form match, inv(N_dst) @ H @ N_src   : 1.7763568394002505e-15
  closed-form match, N_dst @ H @ inv(N_src)   : 6.106226635438361e-16   (normalize_homography3d, for contrast)
  ```

  So `dst_pix_trans_dst_norm @ (H @ src_norm_trans_src_pix)` is the correct mirror of `normalize_homography3d`'s `N_dst @ H @ inv(N_src)` — `dsize_src` on the right, `dsize_dst` on the left, matching the 2-D pair. Round trips, both directions, both batched and unbatched, `(3, 5, 7) -> (6, 8, 11)`:

  ```
  float64 unbatched    denorm(norm(H)) resid=4.441e-16   norm(denorm(H)) resid=2.220e-16
  float64 batched(3)   denorm(norm(H)) resid=4.441e-16   norm(denorm(H)) resid=2.220e-16
  float32 unbatched    denorm(norm(H)) resid=4.768e-07   norm(denorm(H)) resid=2.384e-07
  float32 batched(3)   denorm(norm(H)) resid=4.768e-07   norm(denorm(H)) resid=2.384e-07
  ```

  dtype is preserved through `float16 / bfloat16 / float32 / float64`, and the function uses `_torch_inverse_cast` — which is what its sibling `denormalize_homography` uses and what `normalize_homography3d` uses. That is the right choice: I traced all three and confirmed the 4x4 path keeps `aten::linalg_inv` under `torch.jit.trace` while the 3x3 path does not, i.e. the new function behaves exactly like the 3-D function it mirrors and not like the 2-D one. Nothing to change here.

- `ColmapQTVecs_to_ARKitQTVecs` inverts the forward correctly, and your test is right to compare rotation matrices rather than raw quaternions — 27 of 64 random poses come back sign-flipped, so a raw-component comparison would have failed:

  ```
  float64 n=64: maxR=1.166e-08 maxT=1.182e-08 raw-q=1.929e+00 sign-flipped=27/64
  float32 n=64: maxR=4.768e-07 maxT=8.941e-07 raw-q=1.831e+00 sign-flipped=22/64
  ```

  (For the record: the `1.17e-08` at `float64` is not yours. It is `rotation_matrix_to_quaternion` — `R -> q -> R` alone loses `7.07e-09` at `float64`, the `#3951` non-unit-output wart, already pinned on `main`.)
- `test_reverse_shape_errors` asserts the exception types the code actually raises. I checked all three legs; `ShapeError` really is what comes out of the first two (via `worldtocam_to_camtoworld_Rt`'s `KORNIA_CHECK_SHAPE`) and the `(*, 4)` `ValueError` really is what comes out of the third. No defect there.
- Both new functions are `gradcheck`-clean at `float64` — I ran `torch.autograd.gradcheck` on each and both returned `True`. So the missing gradient pins below are a pure omission, not a limitation.

Everything below is what stands between this and merge.

---

## Blocking

### 1. `denormalize_homography3d` reintroduces the `#3960` shape-guard wart that `#3999` fixed

The new guard is

```python
if not (len(dst_pix_trans_src_pix.shape) == 3 or dst_pix_trans_src_pix.shape[-2:] == (4, 4)):
```

`main` fixed exactly this form in #3999; every other function in the family now reads `ndim in (2, 3) and shape[-2:] == (N, N)`. With `or`, any rank-3 tensor passes clause one whatever its trailing shape is, and any rank-4 tensor of `(…, 4, 4)` passes clause two. Measured, `torch 2.9.1 / cpu`:

```
normalize_homography3d       (1, 3, 3) rank-3, wrong size   -> ValueError: Input dst_pix_trans_src_pix must be a Bx4x4 tensor. Got torch.Size([1, 3, 3])
denormalize_homography3d     (1, 3, 3) rank-3, wrong size   -> RuntimeError: Expected size for first two dimensions of batch2 tensor to be: [1, 3] but got: [1, 4].

normalize_homography3d       (2, 5, 5) rank-3, wrong size   -> ValueError: Input dst_pix_trans_src_pix must be a Bx4x4 tensor. Got torch.Size([2, 5, 5])
denormalize_homography3d     (2, 5, 5) rank-3, wrong size   -> RuntimeError: Expected size for first two dimensions of batch2 tensor to be: [2, 5] but got: [2, 4].

normalize_homography3d       (2, 1, 4, 4) rank-4            -> ValueError: Input dst_pix_trans_src_pix must be a Bx4x4 tensor. Got torch.Size([2, 1, 4, 4])
denormalize_homography3d     (2, 1, 4, 4) rank-4            -> OK shape=(2, 1, 4, 4)
```

Two distinct escapes: a `(B, 3, 3)` dies in the matmul with a message naming neither the argument nor the expected shape, and a rank-4 input is silently accepted and returns a rank-4 result. The good news is the two cases I expected to be fine *are* fine — the unbatched `(4, 4)` is promoted to `(1, 4, 4)` exactly as the sibling does, the exception type is `ValueError` like the sibling, and the message correctly says `Bx4x4` and not `Bx3x3`. So this is a one-line fix:

```python
if not (dst_pix_trans_src_pix.ndim in (2, 3) and dst_pix_trans_src_pix.shape[-2:] == (4, 4)):
```

`tests/geometry/test_conversions.py` already anticipates this function by name and tells you where the pin goes:

```python
# Module-level rather than inline because a fourth op (a denormalize_homography3d per kornia#3962,
# say) becomes a one-line edit that reaches the wrong-size pin and the rank pin together instead of
# landing in one and not the other.
_WRONG_SIZE_CASES = [
    ("normalize_homography", 4),
    ("denormalize_homography", 4),
    ("normalize_homography3d", 3),
]
```

I added `("denormalize_homography3d", 3)` to that table on the merged branch and ran the two `#3960` pins:

```
FAILED ...::test_convention_wrong_sized_matrices_are_rejected_by_the_shape_guard_3960[cpu-denormalize3d]
  - AssertionError: kornia#3960: denormalize_homography3d did not reject a (3, 3) matrix in its guard
FAILED ...::test_convention_the_guard_accepts_rank_2_and_3_and_rejects_higher_3960[cpu-denormalize_homography3d]
  - Failed: DID NOT RAISE <class 'ValueError'>
2 failed, 6 passed
```

Please make that one-line table edit part of this PR: it is the pin, and it is free.

Also note `test_denormalize_homography_invalid_input` passes only because `torch.ones(2, 2)` is the one malformed shape the broken `or` guard still rejects (rank 2, wrong trailing shape). It is not wrong, but it does not reach the gap.

### 2. The AGENTS.md pre-merge grep fails — two prose blocks still say these functions do not exist

`AGENTS.md` requires, before merging a change that closes a tracked issue: *"run `grep -rnE "#NNNN|issues/NNNN" kornia/ tests/` … A surviving hit in `kornia/` means the change is incomplete."*

Important nuance, because it will look like a false alarm on your branch: **this grep is clean on your head and only fails after the rebase.** Your branch is based on `b4f5a78a`, which predates #3984 (batch 3c), the commit that added the `#3962` prose. On your head:

```
$ git grep -nE "#3962|issues/3962" 9bf6aab4 -- kornia/ tests/
(no output)
```

On your head merged with current `origin/main` (`f91a96f9`):

```
$ grep -rnE "#3962|issues/3962" kornia/ tests/
kornia/geometry/conversions.py:2368:          Tracked in `#3962 <https://github.com/kornia/kornia/issues/3962>`_
kornia/geometry/conversions.py:3098:          Tracked in `#3962 <https://github.com/kornia/kornia/issues/3962>`_
tests/geometry/test_conversions.py:495:# Module-level rather than inline because a fourth op (a denormalize_homography3d per kornia#3962,
tests/geometry/test_conversions.py:4055:    # No pin asserts anything about kornia#3962 (no denormalize_homography3d, no
```

The two `kornia/` hits are the blocking ones. They now read, in the very docstrings your functions are supposed to complete:

- `normalize_homography3d`, line 2365: *"there is **no** `denormalize_homography3d`: the 2-D pair is complete, this one is not, so an inverse has to be composed by hand from `normal_transform_pixel3d`. Tracked in #3962"*
- `ARKitQTVecs_to_ColmapQTVecs`, line 3093: *"there is **no** `ColmapQTVecs_to_ARKitQTVecs`, so the conversion cannot be round-tripped through the public API; the inverse has to be composed from `worldtocam_to_camtoworld_Rt` and `camtoworld_vision_to_graphics_Rt`. Tracked in #3962"*

Merging as-is would ship a released docstring that flatly contradicts the function sitting 40 lines below it. Both bullets should be replaced with a cross-reference to the new function (`:func:` link, mirroring how `denormalize_homography` points back at `normalize_homography`), not deleted outright.

The two `tests/` hits are pins, and per the same rule each has to be re-checked in this change rather than left alone:

- line 495 — the `_WRONG_SIZE_CASES` comment above; add the cell, and the "a fourth op … say" wording becomes "a fourth op" in the present tense.
- line 4055 — *"No pin asserts anything about kornia#3962 (no denormalize_homography3d, no ColmapQTVecs_to_ARKitQTVecs) — a missing symbol is a scope question, not a defect."* That sentence becomes false the moment this merges.

I re-ran the full touched suites on the merged state and nothing XPASSes or breaks, so there is no strict-`xfail` to retire — the pins here are prose and tables:

```
tests/geometry/test_conversions.py tests/geometry/transform/test_homography_warper.py
--device=cpu --dtype=float32,float64  ->  821 passed, 16 xfailed
```

### 3. Missing `Convention:` blocks

Every non-deprecated public function in `kornia/geometry/conversions.py` carries a `Convention:` block. Counted on the merged state:

```
public functions in __all__ defined in this module: 47; with a 'Convention:' block: 41
without: ['angle_axis_to_rotation_matrix', 'rotation_matrix_to_angle_axis',
          'quaternion_to_angle_axis', 'angle_axis_to_quaternion',
          'denormalize_homography3d', 'ColmapQTVecs_to_ARKitQTVecs']
```

The four `angle_axis_*` entries are `@deprecated` aliases carrying `# noqa: D103` and no docstring at all. **Your two functions are the only public, non-deprecated functions in the module without one.** This is the file's standard, not a style preference, and `denormalize_homography`'s block is the template to copy — it is short and mostly cross-references. Concretely, each block should state, in falsifiable sentences:

- **`denormalize_homography3d`**: the composition is `inv(N_dst) @ H @ N_src`, mirroring `normalize_homography3d`; both `dsize` tuples are `(depth, height, width)`, `dsize_src` right and `dsize_dst` left; `(x, y, z, 1)` column vectors; per-sample batching; the shape guard (`(4, 4)` or `(B, 4, 4)`, unbatched promoted to `(1, 4, 4)`, rank-4 and `(B, 3, 3)` rejected); and that it inherits `normalize_homography3d`'s `#3958` / `#3959` / `#3904` warnings and its `_torch_inverse_cast` behavior including the 4x4 tracing gap I measured above. Numbers you can quote directly: `float64` round trip `4.44e-16`, `float32` `4.77e-07` at `(3, 5, 7) -> (6, 8, 11)`.
- **`ColmapQTVecs_to_ARKitQTVecs`**: `(w, x, y, z)` real-part-first on both sides; `qvec (B, 4)` and `tvec (B, 3, 1)` strictly batched; input is world-to-camera vision-frame, output is camera-to-world graphics-frame; **the output quaternion sign is not canonical, so compare rotations and never raw components** (27/64 flipped, measured above); and it inherits `#3951`'s non-unit-`float64` wart from `rotation_matrix_to_quaternion`.

While you are in there: `ARKitQTVecs_to_ColmapQTVecs` has a `>>>` example and your reverse does not (34 of 47 public functions in the module have one). A two-line doctest showing the reverse of the forward's own example would close the pair nicely.

### 4. The branch is conflicting and needs a rebase

`mergeStateStatus: DIRTY`. The conflict is trivial, and worth saying precisely so you do not over-think it — `kornia/geometry/conversions.py` **auto-merges cleanly**:

```
$ git merge origin/main
Auto-merging docs/source/geometry.conversions.rst
CONFLICT (content): Merge conflict in docs/source/geometry.conversions.rst
Auto-merging kornia/geometry/conversions.py
Auto-merging tests/geometry/test_conversions.py
CONFLICT (content): Merge conflict in tests/geometry/test_conversions.py
Auto-merging tests/geometry/transform/test_homography_warper.py
Automatic merge failed; fix conflicts and then commit the result.
```

Both conflicts are pure adjacency — two sides adding different lines at the same point — and both resolve by keeping both sides:

- `docs/source/geometry.conversions.rst`: your `.. autofunction:: denormalize_homography3d` against `main`'s `normal_transform_pixel` / `normal_transform_pixel3d` entries.
- `tests/geometry/test_conversions.py`: your two new methods at the end of `TestCARKitToColmap` against the `#3984` convention pins added at the same spot.

I reviewed the merged end state, not the head, and it is green as shown above.

---

## Non-blocking

- **`test_reverse_shape_errors` is a near-verbatim clone of a pin `main` gained after you branched.** `#3984` added `TestCARKitToColmap::test_convention_shapes_are_strictly_batched` with the same three legs (`(4,)` qvec, `(B, 3)` tvec, `(B, 3)` qvec) against the forward function. After the rebase the two sit in the same class. Either parametrize the existing pin over both directions, or keep yours and add a line saying it is the reverse leg of that pin. Also drop the function-local `from kornia.core.exceptions import ShapeError` — the file already imports it at module level (line 35).
- **`test_denormalize_homography_invalid_input` takes no `device` fixture** and builds implicit CPU tensors. The neighbouring `#3960` guard pins take `device` and only hardcode the dtype (a shape guard cannot depend on dtype, but it can still run on the target device). I ran your tests on MPS and they pass, but this one is device-blind by construction.
- **Hand-rolled `float16` tolerances.** The `if dtype == torch.float16: ... rtol=0.0, atol=5e-3` branch bypasses `BaseTester`'s dtype-specific tolerances, which are intentional per `AGENTS.md`, and leaves `bfloat16` on the default. It does pass at all four dtypes as written — I ran `--dtype=float16,bfloat16,float32,float64`, 23 passed — so this is a simplification, not a bug.
- **Pin placement.** `TestNormalizeHomography` in `tests/geometry/test_conversions.py` opens with *"the convention pins live here, next to `normal_transform_pixel`, whose corner-aligned convention all three inherit"*, and already has `test_convention_denormalize_is_the_mirror_composition` and `test_convention_normalize_and_denormalize_round_trip` for the 2-D pair. The 3-D legs belong there; the smoke/shape coverage you added in `test_homography_warper.py` is in the right place.
- **Gradient and compile coverage.** `TestNormalizeHomography::test_gradcheck` covers three ops and is one line from covering four. `ColmapQTVecs_to_ARKitQTVecs` has no `self.gradcheck` either; I confirmed both are `gradcheck`-clean at `float64`, so both pins should just work.
- **PR description.** The template you filled in is the retired one — #4004 removed the lane checkboxes. Nothing to redo, just so you know the current wording is "focused fix" versus "a behavior change or new public API needs an issue first", and you satisfied the latter.

---

## Checklist for the next push

1. Rebase onto current `main` (`f91a96f9`); keep both sides in each of the two adjacency conflicts.
2. Fix the guard: `dst_pix_trans_src_pix.ndim in (2, 3) and dst_pix_trans_src_pix.shape[-2:] == (4, 4)`.
3. Add `("denormalize_homography3d", 3)` to `_WRONG_SIZE_CASES` and update the "a fourth op … say" comment above it.
4. Replace the two "there is **no** …" bullets in `normalize_homography3d` and `ARKitQTVecs_to_ColmapQTVecs` with `:func:` cross-references to the new functions, and update the "No pin asserts anything about kornia#3962" comment at `tests/geometry/test_conversions.py:4055`.
5. Add a `Convention:` block to each new function, per the contents listed above; add a `>>>` example to `ColmapQTVecs_to_ARKitQTVecs`.
6. Move the 3-D round-trip/mirror-composition legs into `TestNormalizeHomography`, add `denormalize_homography3d` to its `test_gradcheck`, add a `gradcheck` for `ColmapQTVecs_to_ARKitQTVecs`, drop the hand-rolled `float16` tolerance branch, give `test_denormalize_homography_invalid_input` the `device` fixture, and reconcile `test_reverse_shape_errors` with `test_convention_shapes_are_strictly_batched`.
7. Re-run `grep -rnE "#3962|issues/3962" kornia/ tests/` — the `kornia/` half must come back empty — then `pixi run pre-commit-all`, `pixi run typecheck`, and `pixi run test-module tests/geometry/test_conversions.py tests/geometry/transform/test_homography_warper.py --dtype=float16,bfloat16,float32,float64`.

Verified on the PR head `9bf6aab4` merged with `origin/main` `f91a96f9`; torch 2.9.1, Python 3.11.14, cpu (with MPS spot checks), macOS.

ducha-aiki (changes requested):
See opus review

PR #3998, opened 2026-08-22
fix(color): validate input range in rgb_to_hsv

### LINKED ISSUE #3997: kornia.color.rgb_to_hsv silently produces wrong values for out-of-[0,1] inputs
## Bug report

`kornia.color.rgb_to_hsv` documents "The image data is assumed to be in the range of (0, 1)" but does not validate (or clip) the input. Values outside `[0, 1]` flow through the formula and silently produce invalid HSV output rather than an error.

### Reproduce

```python
import torch
from kornia.color import rgb_to_hsv
x = torch.tensor([[[[1.5]], [[0.0]], [[0.0]]]])  # R=1.5 out of range
print(rgb_to_hsv(x))
```

`max_rgb` becomes 1.5 and the resulting S/V exceed the documented `[0, 1]` range without any warning.

### Expected

Either:
- validate the input range and raise a clear error, or
- clamp to `[0, 1]` (consistent with other color-conversion assumptions).

### Location

`kornia/color/hsv.py`, `rgb_to_hsv()` (and likely the other `kornia.color` conversions with the same assumption).

AdvaitS:
Fixes #3997

rgb_to_hsv assumed input in [0, 1] but silently produced invalid HSV output for out-of-range values. This adds a clear ValueError for out-of-range inputs.

## Test log

```
$ uv run pytest tests/color/test_hsv.py -q --device=cpu --dtype=float32
23 passed
```

## AI Usage Disclosure

- [x] 🔴 AI-generated — an agent wrote most of it; I verified the result and can address reviewer questions

### DIFF
--- a/kornia/color/hsv.py
+++ b/kornia/color/hsv.py
@@ -53,6 +53,12 @@ def rgb_to_hsv(image: torch.Tensor, eps: float = 1e-8) -> torch.Tensor:
     if len(image.shape) < 3 or image.shape[-3] != 3:
         raise ValueError(f"Input size must have a shape of (*, 3, H, W). Got {image.shape}")
 
+    if image.min() < 0.0 or image.max() > 1.0:
+        raise ValueError(
+            f"Input image values must be in the range [0, 1]. "
+            f"Got min {image.min().item()} and max {image.max().item()}."
+        )
+
     max_rgb = image.amax(-3)
     min_rgb = image.amin(-3)
     deltac = max_rgb - min_rgb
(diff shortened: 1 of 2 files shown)


AdvaitS:
Closing this PR. The validation is too strict: kornia's own augmentation pipeline (e.g. `VideoSequential`) legitimately calls `rgb_to_hsv` with values outside [0,1] (e.g. min -3.46, max 2.82), so raising breaks existing behavior. The "assumed in (0,1)" docstring note is a soft assumption, not a hard contract. Reverting — this was based on a misunderstanding.

PR #3999, opened 2026-08-23
fix(geometry): enforce strict shape validation in homography normalization functions

Xayar145:
### Description

- [x] 🟢 **Green lane**: Bug fix with test coverage (no issue required).

Fixes shape validation guards in `normalize_homography`, `denormalize_homography`, and `normalize_homography3d`:
- Replaced incorrect logical `or` conditions with strict `and` conditions (`len(shape) == 3 and shape[-2:] == (3, 3)` for 2D, `len(shape) == 3 and shape[-2:] == (4, 4)` for 3D).
- Prevents invalid 3D tensors with mismatched trailing dimensions (e.g. `(B, 4, 4)` passed into 2D homography) from bypassing input validation and failing during matrix multiplication.
- Fixed 3D error message to accurately report `Bx4x4`.
- Added unit tests in `tests/geometry/test_conversions.py` verifying shape validation boundaries.

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -1862,7 +1862,7 @@ def normalize_homography(
     if not isinstance(dst_pix_trans_src_pix, torch.Tensor):
         raise TypeError(f"Input type is not a torch.Tensor. Got {type(dst_pix_trans_src_pix)}")
 
-    if not (len(dst_pix_trans_src_pix.shape) == 3 or dst_pix_trans_src_pix.shape[-2:] == (3, 3)):
+    if not (len(dst_pix_trans_src_pix.shape) == 3 and dst_pix_trans_src_pix.shape[-2:] == (3, 3)):
         raise ValueError(f"Input dst_pix_trans_src_pix must be a Bx3x3 tensor. Got {dst_pix_trans_src_pix.shape}")
 
     # source and destination sizes
@@ -2175,7 +2175,7 @@ def denormalize_homography(
     if not isinstance(dst_pix_trans_src_pix, torch.Tensor):
         raise TypeError(f"Input type is not a torch.Tensor. Got {type(dst_pix_trans_src_pix)}")
 
-    if not (len(dst_pix_trans_src_pix.shape) == 3 or dst_pix_trans_src_pix.shape[-2:] == (3, 3)):
+    if not (len(dst_pix_trans_src_pix.shape) == 3 and dst_pix_trans_src_pix.shape[-2:] == (3, 3)):
         raise ValueError(f"Input dst_pix_trans_src_pix must be a Bx3x3 tensor. Got {dst_pix_trans_src_pix.shape}")
 
     # source and destination sizes
@@ -2283,8 +2283,8 @@ def normalize_homography3d(
     if not isinstance(dst_pix_trans_src_pix, torch.Tensor):
         raise TypeError(f"Input type is not a torch.Tensor. Got {type(dst_pix_trans_src_pix)}")
 
-    if not (len(dst_pix_trans_src_pix.shape) == 3 or dst_pix_trans_src_pix.shape[-2:] == (4, 4)):
-        raise ValueError(f"Input dst_pix_trans_src_pix must be a Bx3x3 tensor. Got {dst_pix_trans_src_pix.shape}")
+    if not (len(dst_pix_trans_src_pix.shape) == 3 and dst_pix_trans_src_pix.shape[-2:] == (4, 4)):
+        raise ValueError(f"Input dst_pix_trans_src_pix must be a Bx4x4 tensor. Got {dst_pix_trans_src_pix.shape}")
 
     # source and destination sizes
     src_d, src_h, src_w = dsize_src
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -1862,7 +1862,7 @@ def normalize_homography(
     if not isinstance(dst_pix_trans_src_pix, torch.Tensor):
         raise TypeError(f"Input type is not a torch.Tensor. Got {type(dst_pix_trans_src_pix)}")
 
-    if not (len(dst_pix_trans_src_pix.shape) == 3 and dst_pix_trans_src_pix.shape[-2:] == (3, 3)):
+    if not (dst_pix_trans_src_pix.ndim in (2, 3) and dst_pix_trans_src_pix.shape[-2:] == (3, 3)):
         raise ValueError(f"Input dst_pix_trans_src_pix must be a Bx3x3 tensor. Got {dst_pix_trans_src_pix.shape}")
 
     # source and destination sizes
@@ -2175,7 +2175,7 @@ def denormalize_homography(
     if not isinstance(dst_pix_trans_src_pix, torch.Tensor):
         raise TypeError(f"Input type is not a torch.Tensor. Got {type(dst_pix_trans_src_pix)}")
 
-    if not (len(dst_pix_trans_src_pix.shape) == 3 and dst_pix_trans_src_pix.shape[-2:] == (3, 3)):
+    if not (dst_pix_trans_src_pix.ndim in (2, 3) and dst_pix_trans_src_pix.shape[-2:] == (3, 3)):
         raise ValueError(f"Input dst_pix_trans_src_pix must be a Bx3x3 tensor. Got {dst_pix_trans_src_pix.shape}")
 
     # source and destination sizes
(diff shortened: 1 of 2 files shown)


ducha-aiki:
Thanks for this. **The diagnosis is right — `or` should be `and`** — but I ran the change against current `main` and it is a **breaking change, not a free bugfix**, so it cannot land on the green lane as-is.

### The problem

Unbatched `(3, 3)` input works today. With your change applied to `main`:

```python
normalize_homography(torch.eye(3),    (4,4), (4,4))   # main: (1, 3, 3)   your PR: ValueError
normalize_homography(torch.eye(3)[None], (4,4), (4,4))  # both: (1, 3, 3)
```

The old `or` guard is so loose that it lets a bare `(3, 3)` through, and the function then handles it and returns a batched result. That is an *unenforced accident* rather than a designed contract — but it is an accident users' code can be relying on right now, and silently turning it into a `ValueError` is exactly the class of change [`stability.rst`](https://kornia.readthedocs.io/en/latest/get-started/stability.html) promise 2 forbids. #3960's own Classification section calls it "UNENFORCED-ACCIDENT / semantic contract" for this reason.

So this belongs in the coordinated Repair & Deprecation window, unless you narrow it: **reject the genuinely-wrong shapes (a `(B, 4, 4)` passed to the 2-D function, `(2, 4, 3, 3)`) while continuing to accept `(3, 3)` and unsqueeze it.** That fixes the part with no legitimate callers and leaves the part that has them. That version could go green-lane today, and I'd support it.

### Pins to update either way

Running the full `tests/geometry/` suite with your change on `main` gives **7 failures**. They are the pins from batch 3c (#3984) doing their job:

- `test_convention_wrong_sized_matrices_are_rejected_by_the_shape_guard_3960` — a **strict `xfail` asserting the intended behavior**. It XPASSes now, which is the pin confirming your fix is real. Drop the `xfail` marker; keep the test.
- `test_wart_wrong_sized_matrices_die_inside_matmul_3960` (×3 params) — documents the old behavior. Delete.
- `test_wart_normalize_homography3d_shape_error_says_bx3x3_3960` — fails with "Regex pattern did not match" because you also corrected the `Bx3x3` → `Bx4x4` message. That correction is right and welcome; the pin should be deleted alongside it.

### Test placement

`test_homography_shape_guards()` is a module-level function. Per [`TESTING.md`](https://github.com/kornia/kornia/blob/main/TESTING.md) it should be a method on the existing `TestNormalizeHomography` class and take the injected `device` / `dtype` fixtures rather than constructing default-device tensors — otherwise it never runs on CUDA or MPS.

Please also rebase; the pins above only exist on current `main`.


---

ducha-aiki:
Maintainer decision, so you aren't waiting on a single reviewer's read: **the assessment above is confirmed.** Please take the narrow route rather than the window.

Concretely, what we'd like to merge:

- **reject** the shapes with no plausible callers — a `(B, 4, 4)` passed to the 2-D functions, `(2, 4, 3, 3)`, and the `(B, 3, 3)` mismatch in `normalize_homography3d`;
- **keep accepting** a bare `(3, 3)` (and `(4, 4)` for the 3-D function) and unsqueeze it, exactly as today;
- keep your `Bx3x3` → `Bx4x4` error-message correction — that part is unambiguously right.

That version fixes the real defect, lands on the green lane now, and leaves the semantic half for the coordinated Repair & Deprecation window where it belongs. Please also apply the pin and test-placement notes from the previous comment.

Thanks for finding this — the `or`/`and` slip is real and worth fixing; it just has two halves with very different blast radii.


---

Xayar145:
Thanks @ducha-aiki for the clear guidance!

I've taken the narrow route as requested:
1. **Narrowed shape guards**: The guards in `normalize_homography`, `denormalize_homography`, and `normalize_homography3d` now strictly check `dst_pix_trans_src_pix.ndim in (2, 3)` alongside the trailing dimension checks (`[-2:] == (3, 3)` for 2D, `(4, 4)` for 3D). This rejects mismatched shapes (e.g. `(B, 4, 4)` or 4D tensors in 2D functions) while keeping bare 2D inputs `(3, 3)` / `(4, 4)` supported.
2. **Updated pin tests for #3960**: Removed the strict `xfail` marker from `test_convention_wrong_sized_matrices_are_rejected_by_the_shape_guard_3960`, and deleted the companion `test_wart_wrong_sized_matrices_die_inside_matmul_3960` and `test_wart_normalize_homography3d_shape_error_says_bx3x3_3960`.
3. Kept the `Bx4x4` error message fix.

ducha-aiki:
Thanks — **the narrow route worked, and the breaking-change objection is retired.** I re-ran the check that produced the original verdict and this version is clean.

### What I verified

Applying just your `kornia/` diff onto current `main` (`0685b1f5`):

```
tests/geometry + tests/augmentation  --device=cpu --dtype=float32
4212 passed, 3285 skipped, 25 xfailed, 11 xpassed
```

Zero regressions. `ndim in (2, 3) and shape[-2:] == (3, 3)` is exactly right — it rejects the wrong-sized matrices that #3960 is about while still accepting unbatched 2D input, which is the property the earlier `len(...) == 3 and ...` form broke. The `Bx3x3` → `Bx4x4` message fix in `normalize_homography3d` is correct too.

The `#3960` pins agree. All 7 failures on this PR are pins confirming the fix, not regressions:

- `test_convention_wrong_sized_matrices_are_rejected_by_the_shape_guard_3960[normalize|denormalize|normalize3d]` — **XPASS(strict)**, the guard now rejects what it should
- `test_wart_wrong_sized_matrices_die_inside_matmul_3960[normalize|denormalize|normalize3d]` — dies, the wart is gone
- `test_wart_normalize_homography3d_shape_error_says_bx3x3_3960` — dies, the message is right now

### What's left: the pin edits landed on the wrong copy

Your commit `d89db279` says it un-xfails the convention pin and removes the warts, but the diff against `main` is `+25 / -0` in `tests/geometry/test_conversions.py` — no deletions at all. The cause is the branch base: your branch forks at `b4f5a78a` (Aug 18), which predates the `#3960` pins that batch 3c (`d73bf9c0`) added on Aug 24. So the pins you edited were your branch's own copies, and `main`'s real pins came through the merge untouched. That's what CI is running and failing on.

To fix:

1. **Rebase onto current `main`** — this is the step that makes the rest possible.
2. Then, on `main`'s copies in `tests/geometry/test_conversions.py`: drop the `@pytest.mark.xfail(strict=True)` marker from `test_convention_wrong_sized_matrices_are_rejected_by_the_shape_guard_3960` and keep the test, and delete `test_wart_wrong_sized_matrices_die_inside_matmul_3960` and `test_wart_normalize_homography3d_shape_error_says_bx3x3_3960` outright.
3. Drop the `test_homography_shape_guards` you added — once un-xfailed, the convention pin covers the same ground in the project's pin idiom, so keeping both leaves two tests asserting one thing.

Nothing wrong with the fix itself; this is purely getting the test-side bookkeeping onto the right base.

### Separately, on #4000

That PR carries this same homography guard in the stricter `len(...) == 3` form, which is the version we moved away from here. I've asked over there to let #3990 carry the quaternion work — with you as co-author, since you reached the #3966 half independently — and to leave the homography fix to this PR, which is yours. Details in the #4000 thread.


---

ducha-aiki:
Thanks for this — the fix is right and it's merged-ready now. I rebased the branch onto current `main` and pushed one commit on top of yours (`121a4e8f`), because the branch was based on `b4f5a78a`, three commits before `#3984` landed the `#3960` pins. Those pins were written against the *old* `or` guard, so as soon as your fix met them in CI they all went red — which is exactly what they're for:

- `test_convention_wrong_sized_matrices_are_rejected_by_the_shape_guard_3960` is `xfail(strict=True)`, so your fix made it XPASS, which is a failure. Un-xfailed, comment rewritten in the past tense.
- `test_wart_wrong_sized_matrices_die_inside_matmul_3960` pinned the wrong-sized matrix reaching `matmul`. It doesn't any more — dropped.
- `test_wart_normalize_homography3d_shape_error_says_bx3x3_3960` pinned the wrong `Bx3x3` noun you corrected — turned into a convention pin on the `Bx4x4` message.

I also added a rank pin covering the two cases the wrong-size cells can't see, since they only ever pass rank-3 inputs:

- a rank-2 matrix of the right size is still **accepted** (the old `or` guard let it through by its second clause, so that's behavior your fix had to preserve — a narrower `ndim == 3 and ...` would have broken it, and nothing else in the file covers the unbatched path);
- a rank-4 input is now **rejected**, which the old guard accepted: `normalize_homography(eye(3).expand(2, 1, 3, 3), (4, 5), (8, 9))` used to broadcast through and return a `(2, 1, 3, 3)` matrix. That's the tightening `#3960` asks for and all three functions document a `Bx3x3`/`Bx4x4` argument, so it's pinned deliberately rather than left as an incidental side effect.

One thing I removed: your module-level `test_homography_shape_guards`. Every leg of it is now covered by the two convention pins inside `TestNormalizeHomography` — at that class's cells, conventions and comment style — so keeping both would have been the same claim asserted twice in two different shapes. No criticism of the test; it just landed before you could see the pins that `#3984` was about to add.

Verified: `tests/geometry` + `tests/augmentation` green at float32, the full `tests/geometry/test_conversions.py` green at float32 and float64, ruff clean.

CI is now **33/33 green** including `collector`. Nothing left on your side.

ducha-aiki (approved):

PR #4000, opened 2026-08-23
fix(geometry): guard against float16 eps underflow in quaternion_exp_to_log and quaternion_log_to_exp

Xayar145:
### Description

- [x] 🟢 **Green lane**: Bug fix with test coverage (no issue required).

Fixes float16 numerical underflow in `quaternion_exp_to_log` and `quaternion_log_to_exp`:
- In `float16`, the default `eps = 1e-8` underflows to `0.0` because `float16` smallest subnormal is ~`6e-8`.
- When clamping `norm_q` with `min=eps`, the clamp was a no-op at `float16`, causing `0 / 0 -> NaN` for zero-vector or identity quaternions.
- Used `effective_eps = max(eps, torch.finfo(quaternion.dtype).eps)` so clamping is always representable in the active tensor dtype.
- Added unit tests in `tests/geometry/test_conversions.py` verifying float16 zero vector and identity conversions.

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -1101,7 +1101,9 @@ def quaternion_log_to_exp(quaternion: torch.Tensor, eps: float = 1.0e-8) -> torc
         raise ValueError(f"Input must be a tensor of shape (*, 3). Got {quaternion.shape}")
 
     # compute quaternion norm
-    norm_q: torch.Tensor = torch.norm(quaternion, p=2, dim=-1, keepdim=True).clamp(min=eps)
+    dtype_eps = torch.finfo(quaternion.dtype).eps if quaternion.is_floating_point() else eps
+    effective_eps = max(eps, dtype_eps)
+    norm_q: torch.Tensor = torch.norm(quaternion, p=2, dim=-1, keepdim=True).clamp(min=effective_eps)
 
     # compute scalar and vector
     quaternion_vector: torch.Tensor = quaternion * torch.sin(norm_q) / norm_q
@@ -1145,15 +1147,6 @@ def quaternion_exp_to_log(quaternion: torch.Tensor, eps: float = 1.0e-8) -> torc
         ``1``. Tracked in
         `#3953 <https://github.com/kornia/kornia/issues/3953>`_.
 
-    .. warning::
-        In ``float16`` the default ``eps = 1e-8`` underflows to ``0``, so the
-        clamp that guards the division is a no-op and **any** quaternion with a
-        zero vector part returns ``[nan, nan, nan]`` — including the identity
-        ``[1., 0., 0., 0.]``, whose log is the origin at every other dtype.
-        Passing a representable ``eps`` (e.g. ``eps=1e-3``) returns
-        ``[0., 0., 0.]`` there. Tracked in
-        `#3966 <https://github.com/kornia/kornia/issues/3966>`_.
-
     Args:
         quaternion: a tensor containing a quaternion to be converted.
           The tensor can be of shape :math:`(*, 4)`.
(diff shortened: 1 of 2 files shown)


ducha-aiki:
Heads-up: **four open PRs currently overlap on #3966 and #3975** — #3969, #3970, #3990, #4000 (this one). Filed in a burst and picked up independently. Per the green lane in `CONTRIBUTING.md` this resolves as **first *good* PR wins**, arbitrated by a maintainer, so I am flagging it rather than deciding it.

This PR's distinguishing feature is that it is the only one addressing **both** #3966 and #3975 in a single coherent change, with tests. If the maintainer wants the two fixed together — and they are the same underlying `eps`-handling defect seen at both ends of the log/exp pair — that is an argument in this PR's favour despite being the latest to open.

### Pins you must update

|---|---|---|

---

Xayar145:
Thanks @ducha-aiki for the review and pin breakdown!

I've updated the pins for #3966 and #3975:
1. Dropped the strict `xfail` markers on `test_convention_log_to_exp_of_the_origin_is_the_identity_in_float16_3966` and `test_convention_exp_to_log_of_the_identity_is_the_origin_in_float16_3966`.
2. Removed the companion wart tests `test_wart_float16_eps_underflow_makes_log_to_exp_nan_3966`, `test_wart_float16_eps_underflow_makes_exp_to_log_nan_3966`, and `test_wart_large_finite_input_overflows_log_to_exp_to_nan_3975`.

ducha-aiki:
Thanks for turning this around so fast. Two things need to change before this can land, and one of them is the reason I'd rather fold this PR into #3990 than merge it.

### The pins for #3975 were removed, but #3975 is not fixed

The current head deletes every reference to the `#3975` pin from `tests/geometry/test_conversions.py` — `main` has 6, this branch has 0. But the bug those pins document is still live. With only your `kornia/` diff applied onto current `main`:

```python
>>> quaternion_log_to_exp(torch.tensor([1e20, 0., 0.]))
tensor([nan, nan, nan, nan])
```

`effective_eps = max(eps, finfo(dtype).eps)` addresses the *underflow* class in #3966 — the clamp being a no-op at float16. #3975 is the *overflow* class: `torch.norm(p=2)` forms a sum of squares that goes to `inf` well below the largest finite input. A larger `eps` doesn't touch that path.

A wart pin asserts "this bug is still here". Deleting one is how we record that a bug is gone, so removing these six would have told the next reader #3975 was fixed when it isn't. That's the one move the pin system is specifically there to prevent, so I have to push back on it — no harm done, it just needs undoing.

### The homography guard duplicates your own #3999

This branch also carries the `#3960` shape-guard change, in the stricter form:

```python
if not (len(dst_pix_trans_src_pix.shape) == 3 and dst_pix_trans_src_pix.shape[-2:] == (3, 3)):
```

That's the same fix as your #3999, but with `len(...) == 3` where #3999 now uses `ndim in (2, 3)` — this version rejects unbatched 2D `(3, 3)` input, which is the narrowing we asked you to avoid in the #3999 thread. It's also why the 7 red checks on this PR are all `#3960` pins: the guard changed here, but `main`'s `#3960` pins weren't updated on this branch.

### What I'd like to do

**Fold this into #3990 rather than merge it separately.** #3990 fixes both #3966 and #3975 — I verified it clears all 8 pins with no regressions — and its float32 upcast is doing work your approach can't substitute for: in the float16 #3975 case `[49152, 49152, 49152]` the true norm is about `85133`, past float16's max of `65504`, so the norm isn't representable at that dtype and no `eps` value recovers it.

You got to the #3966 half independently and correctly, so I intend to **squash-merge #3990 with `Co-authored-by: Xayar145 <199441672+Xayar145@users.noreply.github.com>`** — the fix lands with joint authorship and nothing you worked out is lost.

That leaves you owning **#3999** for the homography half, which is genuinely yours and which I've now verified is clean on current `main` (4212 passed across `tests/geometry` + `tests/augmentation`, zero regressions). Details in that thread.

So: no work needed here beyond closing this once #3990 goes in — unless you'd rather keep it open and strip it back to the quaternion half, in which case the `#3975` pins need restoring first. Your call, and thanks again for the quick turnaround on both of these.


---

ducha-aiki:
Closed in favour of merged #3990, and stating you there as co-author to give the credit for work done

PR #4001, opened 2026-08-23
test(geometry): make conversion tests deterministic

### LINKED ISSUE #3943: tests/geometry/test_conversions.py: unseeded torch.rand makes float16/bfloat16 failure counts vary run-to-run on an identical tree (59-62)
`tests/geometry/test_conversions.py` contains **31 `torch.rand` calls and zero `torch.manual_seed`**, so every run draws different input data. At `float32`/`float64` the tolerances absorb the variation, but at `float16`/`bfloat16` several tests sit close enough to their tolerance that pass/fail depends on the particular draw. The result is that **two back-to-back runs of the same command on an identical, unmodified tree produce different failure counts** — I measured 59, 60, 61, 62, 60 across five consecutive runs, with 7 distinct test IDs flapping.

This is not a library defect — the underlying conversions behave the same way each time; it is the test data that moves. But it makes the file's failure count unusable as a baseline, which matters for any PR that has to demonstrate "no new failures" against this file.

## Repro

Run the exact same command twice (or a few times — variance shows up reliably within a handful of runs) on a clean tree:

```bash
pixi run -e default uv run pytest tests/geometry/test_conversions.py -q --dtype=all
```

To collect the flapping IDs across five runs:

```bash
for i in 1 2 3 4 5; do
  pixi run -e default uv run pytest tests/geometry/test_conversions.py -q --dtype=all 2>&1 \
    | grep -oE 'tests/geometry/test_conversions\.py::[^ ]+' | sed 's/\x1b\[[0-9;]*m//g' | sort -u > fail_$i.txt
  echo "run $i: $(wc -l < fail_$i.txt) failures"
done
cat fail_*.txt | sort | uniq -c | sort -n | awk '$1 < 5 {print $1"/5  "$2}'
```

## Observed

```
run 1:       59 failures
run 2:       60 failures
run 3:       61 failures
run 4:       62 failures
run 5:       60 failures
```

Tests that failed in some runs but not others (count of runs failed, out of 5):

```
1/5  TestPolCartConversions::test_pol2cart[cpu-bfloat16-batch_shape3]
1/5  TestPolCartConversions::test_pol2cart[cpu-float16-batch_shape3]
1/5  TestQuaternionFromEuler::test_forth_and_back[cpu-bfloat16]
2/5  TestRadDegConversions::test_deg2rad[cpu-float16-batch_shape1]
4/5  TestRadDegConversions::test_deg2rad[cpu-bfloat16-batch_shape0]
4/5  TestRadDegConversions::test_deg2rad[cpu-bfloat16-batch_shape1]
4/5  TestRotationMatrixToAngleAxis::test_rand_quaternion_gradcheck[cpu-float16-5]
```

All of them consume unseeded `torch.rand`, e.g.:

```python
# tests/geometry/test_conversions.py:589-597
def test_deg2rad(self, batch_shape, device, dtype, atol, rtol):
    x_deg = 180.0 * torch.rand(batch_shape, device=device, dtype=dtype)
    ...
    self.assert_close(x_deg, x_rad_to_deg, atol=atol, rtol=rtol)

# tests/geometry/test_conversions.py:517-519
def test_rand_quaternion_gradcheck(self, batch_size, device, dtype, atol, rtol):
    quaternion = torch.rand(batch_size, 4, device=device, dtype=dtype)

# tests/geometry/test_conversions.py:654-657
def test_pol2cart(self, batch_shape, device, dtype):
    rho = torch.rand(batch_shape, dtype=dtype)
    phi = kornia.constants.pi * torch.rand(batch_shape, dtype=dtype)
```

Counts for the file: `grep -c "torch.rand"` → **31**; `grep -c "manual_seed"` → **0**.

## Expected

The same command on the same tree should produce the same result. Two fixes are conventional, and kornia already uses the second one extensively:

1. Seed once for the module via an `autouse` fixture, or
2. Call `torch.manual_seed(...)` at the top of each test that draws random data.

**Precedent in this repo for (2):** 46 files under `tests/` already do this — for example `tests/losses/test_mutual_information.py:146` (`torch.manual_seed(0)  # Fix seed for reproducibility`), `tests/filters/test_otsu_thresholding.py:42` (`torch.manual_seed(0)`), and `tests/filters/test_in_range.py:30,50` (`torch.manual_seed(1)`). There is no global seeding in a `conftest.py`, so files that do not seed themselves are genuinely unseeded.

Note that seeding will *freeze* whichever side of the tolerance the chosen draw falls on. For the reduced-precision cases above, the honest sequence is: seed first so the result is reproducible, then decide per test whether the remaining failure is a real tolerance problem to widen or a genuine precision limit to `xfail`/skip at `float16`/`bfloat16`. Seeding alone converts "randomly 59-62 failures" into "deterministically N failures"; it does not by itself make those tests pass.

## Why this is worth fixing

It makes baseline comparisons noisy for any docs/test PR touching this area. Concretely, **#3942**'s verification section has to carry an explicit caveat for exactly this reason:

> Caveat, stated because it makes baseline diffs noisy: `tests/geometry/test_conversions.py` contains pre-existing unseeded `torch.rand` calls, so the float16 failure count moves by ±2 between identical runs of the *unmodified* file — a separate seeding issue, not introduced here.

A reviewer checking "did this PR add failures?" against this file currently cannot distinguish a real regression of 1-2 tests from RNG noise without re-running several times.

## Classification

**Test-infrastructure flakiness — not a library bug.** No kornia function behaves nondeterministically here; `deg2rad`, `pol2cart` and `quaternion_to_rotation_matrix` are deterministic given their inputs, and the inputs are what change between runs. This is orthogonal to the reduced-precision conversion issues filed separately (#3937, #3938, #3939, #3940) and can be fixed independently of them.

## Environment

kornia 0.9.0rc1, torch 2.9.1, cpu (Darwin 25.5.0). Measured on branch `docs/convention-blocks-conversions-a` @ `41c05d52` (the #3942 docs branch).

That branch's diff to `tests/geometry/test_conversions.py` is **purely additive** — `git diff origin/main` shows 401 insertions and **0 deletions**, so no pre-existing test was modified. All 7 flapping tests, all 31 `torch.rand` calls, and the absence of `manual_seed` are present identically on `origin/main` @ `3358ef16` (verified with `git show origin/main:tests/geometry/test_conversions.py`), so the flakiness is a property of `main`, not of that branch.

FU-max-boop:

**Fixes/Relates to:** Fixes #3943

## What does this PR do?

Add a function-scoped autouse fixture that calls `torch.manual_seed(0)` before every parametrized case in `test_conversions.py`. This makes every executable Torch RNG consumer in the file, including `Quaternion.random`, reproducible and keeps individual tests independent of collection order or selection.

This changes no library behavior, tolerance, expected value, skip, or xfail. The 61 reduced-precision failures below are pre-existing issues; this PR stabilizes that baseline rather than making those cases pass.

On unmodified `main`, five consecutive all-dtype runs produced 62, 60, 63, 60, and 63 failures. With this fixture, five consecutive runs produced the same count and the same set of failure IDs every time.

## Test log

```
$ for i in 1 2 3 4 5; do
    .pixi/envs/default/bin/uv run pytest tests/geometry/test_conversions.py -q --dtype=all
  done

run 1: 61 failed, 863 passed, 10 skipped, 37 xfailed, 8 warnings
run 2: 61 failed, 863 passed, 10 skipped, 37 xfailed, 8 warnings
run 3: 61 failed, 863 passed, 10 skipped, 37 xfailed, 8 warnings
run 4: 61 failed, 863 passed, 10 skipped, 37 xfailed, 8 warnings
run 5: 61 failed, 863 passed, 10 skipped, 37 xfailed, 8 warnings
failure-ID sets differing across runs: 0

$ .pixi/envs/default/bin/uv run pytest tests/geometry/test_conversions.py -q --dtype=float32
283 passed, 22 xfailed, 8 warnings in 1.47s

$ .venv/bin/ruff check tests/geometry/test_conversions.py
All checks passed!

$ .venv/bin/ruff format --check tests/geometry/test_conversions.py
1 file already formatted
```

- [ ] 🟢 **Human-written**

### DIFF
--- a/tests/geometry/test_conversions.py
+++ b/tests/geometry/test_conversions.py
@@ -46,6 +46,11 @@ from kornia.geometry.quaternion import Quaternion
 from testing.base import BaseTester, assert_close
 
 
+@pytest.fixture(autouse=True)
+def seed_rng() -> None:
+    torch.manual_seed(0)
+
+
 @pytest.fixture()
 def atol(device, dtype):
     """Lower tolerance for cuda-float16 only."""


ducha-aiki (approved):
Correct fix, correctly scoped, and it does exactly what it claims. Verified by execution on current `main` rather than on your branch as-is — see the one caveat about your base at the end.

## Verified

Five consecutive `pytest tests/geometry/test_conversions.py -q --dtype=all` runs, CPU, torch 2.9.1:

| tree | failure counts | flapping IDs |
|---|---|---|
| `origin/main` (`35555b55`) | **59, 59, 61, 58, 59** | **8** |
| your fixture, cherry-picked onto `35555b55` | **59, 59, 59, 59, 59** | **0** |

Not just stable counts — the failure **ID sets are byte-identical across all five runs**, which is the property that actually matters for using this file as a baseline.

The eight IDs that flap on `main`, and where seed 0 puts each:

| test | failed on main | with seed 0 |
|---|---|---|
| `TestRadDegConversions::test_deg2rad[cpu-bfloat16-batch_shape1]` | 4/5 | fails |
| `TestRadDegConversions::test_deg2rad[cpu-float16-batch_shape0]` | 4/5 | fails |
| `TestRotationMatrixToAngleAxis::test_rand_quaternion_gradcheck[cpu-float16-5]` | 3/5 | fails |
| `TestEulerFromQuaternion::test_forth_and_back[cpu-float16]` | 1/5 | passes |
| `TestPolCartConversions::test_pol2cart[cpu-bfloat16-batch_shape2]` | 1/5 | passes |
| `TestPolCartConversions::test_pol2cart[cpu-bfloat16-batch_shape3]` | 1/5 | passes |
| `TestRotationMatrixToAngleAxis::test_rand_quaternion_gradcheck[cpu-float16-1]` | 1/5 | passes |
| `TestRotationMatrixToAngleAxis::test_rand_quaternion_gradcheck[cpu-float16-2]` | 1/5 | passes |

The arithmetic closes exactly: 56 tests fail in all five unseeded runs, seed 0 adds 3 of the 8 flappers, 56 + 3 = 59. And the check that matters for a seeding change — **no test fails under seed 0 that was not already observed failing on `main`**. Seed 0 is not an unlucky draw that quietly widens the baseline; it lands at the modal value of `main`'s 58–61 range.

Your claim that this changes no library behavior, tolerance, expected value, skip or xfail is accurate: the diff is five lines in one test file, and nothing in `kornia/` is touched.

## Note your base is stale — your figures are not wrong, they are from a different tree

Your merge base is `b4f5a78a`, which predates `#3984`, `#3990`, `#3999`, `#3977`, `#4006`, `#4028` and `#3995`. That is why your test log shows 61 and I measure 59 on the same command — `#4006` changed the singleton normalization path and moved two `TestDenormalizePixelCoordinates` bfloat16 cases off the failing side. GitHub still reports the PR mergeable, and it is: the fixture inserts into a region none of those merges touched, and I confirmed the determinism result holds when cherry-picked onto `35555b55`. So nothing needs to change. But your CI run dates to 2026-08-23 and has not seen any of the pins that landed since, so please refresh the numbers in the PR body against current `main` before this merges — a stale "61" in a determinism PR is exactly the kind of figure someone will later take as the baseline.

## One thing worth recording, not changing

Seeding makes the baseline usable, but it also makes the underlying fragility invisible: five of those eight tests now pass deterministically, and they are one seed away from failing. The tolerance is what is actually too tight for `float16`/`bfloat16` there — `test_deg2rad` compares a round-trip at the fixture's `atol`/`rtol` on data spanning `[0, 180)`, where bfloat16 has 8 bits of mantissa. `#3943` says outright that this is not a library defect, and I agree it should not grow into a tolerance change inside this PR.

What I would like is for that list of eight to survive somewhere. Please paste the table above into the PR body (or into `#3943` when it closes) so the record shows which tests are seed-sensitive and which side seed 0 happens to land on. Otherwise the next person to touch those tolerances has no way to know that five of them were passing by luck.

## Question for the maintainer, not a change request

`tests/geometry/test_conversions.py` is not the only file with unseeded `torch.rand` — `test_bbox.py`, `test_boxes.py`, `test_depth.py`, `test_keypoints.py`, `test_linalg.py`, `test_line.py`, `test_homography.py`, `test_plane.py` and `test_ransac.py` all have it, and `test_ransac.py`'s two `test_dirty_points` cases are already carrying `xfail` reasons that say "might slightly and randomly imprecise due to RANSAC randomness". The identical five-line fixture in `conftest.py` would seed the whole suite. `#3943` scopes itself to this one file and this PR matches that scope exactly, which is the right call for a green-lane fix — I am flagging the wider version as a separate decision, not asking for it here.

Approving. Refresh the body figures against current `main` and this is ready.

---

Issue #3986, opened 2026-08-18
[Bug]: seems like a bug in Kornia 0.8.3


KoenKeunen:
### What happened?

See this thread;
https://github.com/Lightricks/ComfyUI-LTXVideo/issues/542

It seems it's a bug in Korna 0.8.3, not in ComfyUI-LTXVideo. It's fixed when rolling back to 0.8.2 for everyone who tries it.

### Minimal reproduction

```python
Traceback (most recent call last): File "C:\Users\Axiomgraph\Downloads\ComfyUI_windows_portable\ComfyUI\nodes.py", line 2247, in load_custom_node module_spec.loader.exec_module(module) ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^ File "", line 1023, in exec_module File "", line 488, in call_with_frames_removed File "C:\Users\Axiomgraph\Downloads\ComfyUI_windows_portable\ComfyUI\custom_nodes\ComfyUI-LTXVideo_init.py", line 49, in from .pyramid_blending import LTXVLaplacianPyramidBlend File "C:\Users\Axiomgraph\Downloads\ComfyUI_windows_portable\ComfyUI\custom_nodes\ComfyUI-LTXVideo\pyramid_blending.py", line 7, in from kornia.geometry.transform.pyramid import ( ...<6 lines>... ) ImportError: cannot import name 'pad' from 'kornia.geometry.transform.pyramid' (C:\Users\Axiomgraph\Downloads\ComfyUI_windows_portable\python_embeded\Lib\site-packages\kornia\geometry\transform\pyramid.py)
```

### Environment

- Kornia version: 0.8.3
- PyTorch version: 2.12.1
- Python version: 3.13.12
- OS / device (CPU, CUDA, MPS): Windows 11, cu130

alinaqvi1129:
id like to work on this

ducha-aiki:
Closing — this is a real break we caused, but the fix belongs downstream, not in kornia.

`pad` was never public API in `kornia.geometry.transform.pyramid`: never in `__all__`, never documented, just `torch.nn.functional.pad` leaking out of the module because of a `from kornia.core import ..., pad, ...` import. When the module switched to calling `F.pad(...)` directly for 0.8.3, that binding disappeared with it. Restoring it would freeze an accidental implementation detail into our API surface permanently, so I've closed #4003 which proposed exactly that.

[...]

Issue #551, opened 2020-05-01
Assign different alphas to every class in Focal Loss


hal-314:
Hi

First of all, great work with kornia! Currently, focal loss (FL) is implemented for multiclass tasks. However, its alpha parameter is shared between all classes when the original paper defines FL for binary tasks as: FL(pt)=−αt(1−pt)γlog(pt). See https://arxiv.org/abs/1708.02002 and #493 .

I modified the current FL code to support different alphas for every class (see below). The only relevant change is:
```
# Instead of using a constant, we change to a tensor of size (N, C, *) and rename alpha to alphas
alphas = torch.tensor(alphas, dtype=input.dtype, device=input.device).view(1, -1, *[1]*(input.ndim-2))      
``` 

Here is the whole code (I haven't changed docs):

```
from typing import Optional, List  # <-Change:  Add List

import torch
import torch.nn as nn
import torch.nn.functional as F

from kornia.utils import one_hot


# based on:
# https://github.com/zhezh/focalloss/blob/master/focalloss.py

def focal_loss(
        input: torch.Tensor,
        target: torch.Tensor,
        alphas: Optional[List[float]], # <-Change:  rename to alphas and to a list of floats
        gamma: float = 2.0,
        reduction: str = 'none',
        eps: float = 1e-8) -> torch.Tensor:
    r"""Function that computes Focal loss.

    See :class:`~kornia.losses.FocalLoss` for details.
    """
    if not torch.is_tensor(input):
        raise TypeError("Input type is not a torch.Tensor. Got {}"
                        .format(type(input)))

    if not len(input.shape) >= 2:
        raise ValueError("Invalid input shape, we expect BxCx*. Got: {}"
                         .format(input.shape))
 
    if input.size(0) != target.size(0):
        raise ValueError('Expected input batch_size ({}) to match target batch_size ({}).'
                         .format(input.size(0), target.size(0)))

    n = input.size(0)
    out_size = (n,) + input.size()[2:]
    if target.size()[1:] != input.size()[2:]:
        raise ValueError('Expected target size {}, got {}'.format(
            out_size, target.size()))

    if not input.device == target.device:
        raise ValueError(
            "input and target must be in the same device. Got: {} and {}" .format(
                input.device, target.device))

    ### New addition ###
    alphas = torch.tensor(alphas, dtype=input.dtype, device=input.device).view(1, -1, *[1]*(input.ndim-2))      
    if alphas.size(1) != input.size(1):
[...]

Issue #508, opened 2020-04-05
Implement hard nms2d/3d with conv2d instead of max pool


ducha-aiki:
Something like this, will submit PR later:

```python
def nms2d(x: torch.Tensor, th: float = 0):
    r"""Applies non maxima suppression to the feature map in 3x3 neighborhood.
    Args:
        x: torch.Tensor: 4d tensor
        th (float): threshold
    Return:
        torch.Tensor: nmsed input

    Shape:
      - Input: :math:`(B, C, H, W)`
      - Output: :math:`(B, C, H, W)`
    """
    weight = torch.zeros(9, 1, 3, 3, dtype = x.dtype, device=x.device)
    for i in range(9):
        if i == 4: continue
        h = i % 3
        w = i // 3
        weight[i, 0, w, h] = 1.0
    max_non_center = F.conv2d(x, weight, stride=1, padding=1).max(dim=1, keepdim=True)[0]
    out = x * (x > max_non_center.to(x.dtype)) * (x > th).to(x.dtype)
    return out

```

The reason: current implementation with maxpool output everything for the input torch.ones()

edgarriba:
we implemented with @lferraz this alternative at some point for 3d:
```python
def _create_valid_mask(input):
    def _create_kernel(eps: float = 1e-6):        
        kernel = torch.eye(27)
        kernel[:, 13 ] = (eps -1.)
        kernel = torch.cat([kernel[:13], kernel[14:28]], dim=0)
        return kernel.view(26, 3, 3, 3)                 
    #import pdb;pdb.set_trace()
    B, CH, L, H, W = input.shape                                                    
    kernel = _create_kernel()
    tmp_kernel: torch.Tensor = kernel.to(                                      
        input.device, input.dtype).detach()                                         
    kernel: torch.Tensor = tmp_kernel.unsqueeze(1)  # 26x1x3x3x3                    
                                                                                    
    x_diff = F.conv3d(input + 1e-6, kernel, stride=1, padding=(0,1,1))  # Bx26x1xHxW
    kernel_tmp = kernel.clone()
    kernel_vec = (kernel_tmp==1).nonzero()[..., -3:] - 1.

    x_vec = torch.sum(x_diff * kernel_vec.view(1, 26, 3, 1, 1), dim=1)    

    x_out = torch.relu(x_diff).sum(1)
    x_valid = (x_out==0).view(1,1,1,x_out.shape[-2],x_out.shape[-1])
[...]

PR #506, opened 2020-04-01
binary focal loss with logits implementation

xen0f0n:
Following issue #493 https://github.com/kornia/kornia/issues/493
I've implemented focal loss for the binary task. As the name suggests, input tensor should be the model logits.
As stated in the original [paper](https://arxiv.org/pdf/1708.02002.pdf), parameter `alpha` is assigned to the rare class. It's assumed that the rare class is labeled as "1".

PR #4003, opened 2026-08-24
[BUG]: keep pad importable from pyramid module

shlokkvaishnav:
## Which lane?
- [x] 🟢 **Green lane** — test-first bug fix. No prior issue needed (but there is one: #3986).

**Fixes/Relates to:** #3986

## What does this PR do?

`kornia.geometry.transform.pyramid` used to bind `pad` at module scope via `from kornia.core import Module, Tensor, ones, pad, stack, tensor, zeros` (present in v0.8.2). That import was dropped as an incidental side effect of refactoring the module to call `torch.nn.functional.pad` directly via `F.pad(...)` (landed by v0.8.3), so `pad` silently stopped being importable from this module — `kornia.core` no longer re-exports it either (it was removed project-wide, not just here).

This broke third-party code relying on `from kornia.geometry.transform.pyramid import pad`, reported in #3986 with a downstream traceback from ComfyUI-LTXVideo:

```
ImportError: cannot import name 'pad' from 'kornia.geometry.transform.pyramid'
```

I confirmed the regression by diffing `pyramid.py` between the `v0.8.2` and `v0.8.3` tags — `pad` was dropped from the file's imports with no replacement export, while every internal call site switched to `F.pad(...)` directly. `pad` was never in the module's `__all__`, so this was always an implementation-detail import, but real downstream code depends on it, and restoring it is a zero-risk one-line fix.

Test that fails on `main` without this fix, in `tests/geometry/transform/test_pyramid.py`:

```python
def test_pad_importable_from_pyramid_module():
    from kornia.geometry.transform.pyramid import pad
    assert pad is torch.nn.functional.pad
```

Fix: `pad = F.pad` as a module-level alias in `pyramid.py`, matching the value it always held.

## Test log

```
$ pixi run test tests/geometry/transform/test_pyramid.py --dtype=float32 --device=cpu

tests/geometry/transform/test_pyramid.py ............................................................ [100%]
============================= 60 passed in 2.90s ==============================
```

Also ran `ruff check` and `ruff format --check` on both changed files — clean.

## AI Usage Disclosure

### DIFF
--- a/kornia/geometry/transform/pyramid.py
+++ b/kornia/geometry/transform/pyramid.py
@@ -26,6 +26,13 @@ from torch import nn
 from kornia.core.check import KORNIA_CHECK, KORNIA_CHECK_SHAPE
 from kornia.filters import filter2d, gaussian_blur2d
 
+# Kept as a module-level alias for backwards compatibility: earlier versions
+# imported `pad` from `kornia.core`, which made it importable from here too
+# (e.g. `from kornia.geometry.transform.pyramid import pad`). That import was
+# dropped when this module switched to calling `F.pad` directly, silently
+# breaking third-party code relying on it. See #3986.
+pad = F.pad
+
 __all__ = ["PyrDown", "PyrUp", "ScalePyramid", "build_laplacian_pyramid", "build_pyramid", "pyrdown", "pyrup"]
 
 
--- a/tests/geometry/transform/test_pyramid.py
+++ b/tests/geometry/transform/test_pyramid.py
@@ -243,3 +243,22 @@ class TestBuildLaplacianPyramid(BaseTester):
         batch_size, channels, height, width = 1, 2, 7, 9
         img = torch.rand(batch_size, channels, height, width, device=device, dtype=torch.float64)
         self.gradcheck(kornia.geometry.transform.build_laplacian_pyramid, (img, max_level), nondet_tol=1e-8)
+
+
+def test_pad_importable_from_pyramid_module():
+    """Regression test for https://github.com/kornia/kornia/issues/3986.
+
+    ``kornia.geometry.transform.pyramid`` used to bind ``pad`` at module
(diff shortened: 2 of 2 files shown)


ducha-aiki:
Thanks for the careful write-up @shlokkvaishnav — the diagnosis is exactly right, and the archaeology (diffing `pyramid.py` between `v0.8.2` and `v0.8.3`) is appreciated. I'm closing this anyway, and I want to explain why so the effort doesn't feel wasted.

`pad` here was never API. It was never in `__all__`, never documented, and never anything but `torch.nn.functional.pad` leaking out of the module because of a `from kornia.core import ..., pad, ...` line. Adding `pad = F.pad` back makes that alias permanent: it becomes a name we have to keep, test and reason about forever, and the regression test in this PR is what makes it permanent — once there's a test asserting `pad is torch.nn.functional.pad`, no future refactor can remove it. We'd be promoting an accident to a contract.

The downstream break was real and it was our fault — but the fix costs one line in the consumer, and it's already written. ComfyUI-LTXVideo has **five** open PRs doing it ([#498](https://github.com/Lightricks/ComfyUI-LTXVideo/pull/498), [#506](https://github.com/Lightricks/ComfyUI-LTXVideo/pull/506), [#508](https://github.com/Lightricks/ComfyUI-LTXVideo/pull/508), [#516](https://github.com/Lightricks/ComfyUI-LTXVideo/pull/516), [#551](https://github.com/Lightricks/ComfyUI-LTXVideo/pull/551)), the oldest since May. I've commented on their issue asking them to merge one: https://github.com/Lightricks/ComfyUI-LTXVideo/issues/542#issuecomment-5424461135

If you'd like to land something adjacent and genuinely useful: the real gap is that we have no way to tell when a refactor drops a name third parties were importing. A check over the public surface — or a documented policy on what's importable from a submodule — would prevent the next instance of this rather than patching one symbol. Happy to discuss shape in an issue if you're interested.

Issue #4026, opened 2026-08-26
57 public names are importable but excluded from their module's `__all__` — including two that ComfyUI-LTXVideo imports today, in the file #3986 broke


ducha-aiki:
#3986 broke because `pad` was importable from `kornia.geometry.transform.pyramid` while being absent from that module's `__all__`, undocumented, and untested as API. A refactor dropped it and nobody noticed for three months.

An audit of the whole tree finds **57 more names in exactly that state**: defined in a kornia module, public-named, in a module that *does* declare `__all__` and *does* exclude them. Every one is importable today and none is covered by any promise.

## The live case, in the same file

ComfyUI-LTXVideo's import block — the one from the #3986 traceback — pulls six names from `kornia.geometry.transform.pyramid`:

```python
from kornia.geometry.transform.pyramid import (
    PyrUp, build_laplacian_pyramid, build_pyramid,
    find_next_powerof_two, is_powerof_two, pad,
)
```

Four resolve, `pad` is the one that broke — and of the four survivors, **`is_powerof_two` and `find_next_powerof_two` are in this same undeclared state**. Both are defined at `kornia/geometry/transform/pyramid.py:563,568`; neither is in `__all__`; `grep -rn "is_powerof_two\|find_next_powerof_two" docs/` returns nothing. The next refactor of that file breaks them the same way, for the same consumer.

## Repro

```python
import kornia.geometry.transform.pyramid as m
print(m.__all__)
print([(n, hasattr(m, n), n in m.__all__) for n in ("is_powerof_two", "find_next_powerof_two")])
```

## Observed

```
['PyrDown', 'PyrUp', 'ScalePyramid', 'build_laplacian_pyramid', 'build_pyramid', 'pyrdown', 'pyrup']
[('is_powerof_two', True, False), ('find_next_powerof_two', True, False)]
```

## The full list (57)

| module | undeclared public names |
|---|---|
| `kornia.augmentation.container.augment` | `MaskDataType` |
| `kornia.augmentation.container.base` | `TransformMatrixMinIn` |
| `kornia.augmentation.container.image` | `ImageModuleForSequentialMixIn` |
| `kornia.augmentation.random_generator._2d.probability` | `random_prob_generator` |
| `kornia.config` | `KorniaConfig`, `LazyLoaderConfig` |
| `kornia.constants` | `DataKey`, `T` |
| `kornia.contrib.boxmot_tracker` | `logger` |
| `kornia.contrib.super_resolution` | `URLs` |
[...]

Issue #4025, opened 2026-08-26
`upscale_double` — a documented public function — was deleted without deprecation by a performance PR (#3623)


ducha-aiki:
`upscale_double` was added in #2105, announced in `CHANGELOG.md` as a feature, listed in `kornia.geometry.transform.pyramid.__all__`, and **rendered in the public documentation** (`docs/source/geometry.transform.rst:44`, `.. autofunction:: upscale_double`) through v0.8.2.

It was deleted in #3623 — "feat: add AdaptiveQuadInterp3d and scale-space detector speedups", merged 2026-03-19 — together with its docs line and 102 lines of its tests. No deprecation, no replacement, no release-note entry.

## Repro

```bash
git show v0.8.2:kornia/geometry/transform/pyramid.py | grep -n "def upscale_double"
git grep -n upscale_double v0.8.2 -- docs/
grep -rn upscale_double --include=*.py --include=*.rst .
python -c "import kornia.geometry.transform.pyramid as p; print(hasattr(p, 'upscale_double'))"
```

## Observed

```
433:def upscale_double(x: Tensor) -> Tensor:                       # present at v0.8.2
v0.8.2:docs/source/geometry.transform.rst:44:.. autofunction:: upscale_double   # documented at v0.8.2
                                                                # (no hits on main outside CHANGELOG.md)
False
```

`git show 398665c7 --stat` confirms the removal shape:

```
docs/source/geometry.transform.rst      |   1 -
kornia/geometry/transform/pyramid.py    | 161 +++--
tests/geometry/transform/test_pyramid.py| 102 ---
```

## Why this one matters more than #3986's `pad`

`pad` was undocumented, so `stability.rst` classes it as private and the project owes nothing (that is the basis on which #4003 was closed). `upscale_double` was **documented**, which is the policy's own definition of public API. It is the clearest instance of the thing the policy now forbids — and it was removed by a performance PR, where nobody was looking at the export surface.

Fairness note: the policy was published 2026-08-07 (#3896), five months *after* this removal. This was not a violation of a rule that existed at the time. It is the case that motivates enforcing the rule now.

## Proposal: leave it removed

Recommendation, for maintainer ratification:

[...]

Issue #4024, opened 2026-08-26
`kornia.core` dropped all 32 of its `__all__` names in 0.8.3 with no deprecation shim and no release note (824 binding losses; #3986 is one of them)


ducha-aiki:
At v0.8.2, `kornia/core/__init__.py` re-exported `._backend` and declared **32 names in `__all__`**:

```
Device Dtype Module ModuleList Parameter Sequential Tensor arange as_tensor complex
concatenate cos deg2rad diag einsum eye linspace map_coordinates normalize ones
ones_like pad rad2deg rand sin softmax stack tan tensor where zeros zeros_like
```

In 0.8.3, `kornia/core/_backend.py` was deleted and all 32 went with it — no `__getattr__` shim, no `@deprecated` wrapper, no release-note entry.

## Repro

```python
import kornia, kornia.core
print(kornia.__version__)
print([n for n in ("Tensor", "Module", "pad", "tensor", "zeros", "stack", "Device")
       if not hasattr(kornia.core, n)])
```

## Observed

```
0.9.0rc1
['Tensor', 'Module', 'pad', 'tensor', 'zeros', 'stack', 'Device']
```

All 20 spot-checked names are absent. `kornia/core/__init__.py` has no module-level `__getattr__`.

## Scale

Because ~500 modules did `from kornia.core import ...`, the removal took **824 module-scope bindings** with it across the tree. Most-lost: `Tensor` ×247, `Module` ×113, `tensor` ×59, `stack` ×43, `zeros` ×35, `concatenate` ×32, `as_tensor` ×27, `where` ×25, `pad` ×19.

#3986 is one of those 19 `pad` bindings — the one that happened to have a downstream consumer loud enough to file a bug (ComfyUI-LTXVideo, broken for three months across five ignored PRs).

`kornia.utils` went through the same removal in the same release and **did it correctly**: a module-level `__getattr__` emitting `DeprecationWarning` and naming `kornia.onnx.download`. All 13 spot-checked `kornia.utils` names still resolve. So the mechanism existed and was simply not applied here.

## Was this a policy violation?

**No, and that is worth stating plainly.** `docs/source/get-started/stability.rst` says a symbol is public when it appears in the rendered documentation, and that "undocumented helpers are private regardless of importability". `docs/source/core.rst` documents exactly two things: `TensorWrapper` and `kornia.core.utils.batched_forward`. None of the 32 was ever documented, so none was public API and promise 1 (deprecation before removal) did not attach.

[...]

Issue #4023, opened 2026-08-26
Broken `__all__` entries: `from kornia.core import *` and `from kornia.color import *` raise AttributeError, and the API-surface guard passes anyway


ducha-aiki:
`kornia.color.__all__` lists `AUTUMN` and `kornia.core.__all__` lists `eye_like` and `vec_like`, but none of the three is bound in its module. Any `from <module> import *` therefore raises, and `tests/test_api_surface.py` — the guard added in #3898 for exactly this class of problem — passes anyway, because it reads `__all__` without checking that the names resolve.

## Repro

```python
import kornia.color, kornia.core
print('AUTUMN', 'AUTUMN' in kornia.color.__all__, hasattr(kornia.color, 'AUTUMN'))
print('eye_like', 'eye_like' in kornia.core.__all__, hasattr(kornia.core, 'eye_like'))
print('vec_like', 'vec_like' in kornia.core.__all__, hasattr(kornia.core, 'vec_like'))
exec('from kornia.core import *')
```

## Observed

```
AUTUMN True False
eye_like True False
vec_like True False
Traceback (most recent call last):
  ...
AttributeError: module 'kornia.core' has no attribute 'eye_like'
```

`from kornia.color import *` fails the same way on `AUTUMN`. Meanwhile:

```
$ pytest tests/test_api_surface.py -q
12 passed in 0.29s
```

## Cause

Two different causes behind one symptom:

- **`kornia.core.eye_like` / `vec_like`** — a real missing import. Both are defined in `kornia/core/ops.py:23,50` and are re-exported by `kornia/utils/__init__.py:251,261` and `kornia/__init__.py:88,94`, but `kornia/core/__init__.py` never imports them from `.ops` while still listing them in `__all__`. They are also recorded in `tests/api_surface.json` under `kornia.core`, so they are intended to be public here.
- **`kornia.color.AUTUMN`** — a stale `__all__` entry. `AUTUMN` was a `ColorMap` subclass carrying a `@deprecated` marker at v0.8.2 (`kornia/color/colormap.py:313`, "will be removed in next kornia versions (0.8.0 - dec 2024)"). The class was removed correctly; the `__all__` entry was left behind.

The guard hole is separate and is what let all three survive: `_current_surface()` in `tests/test_api_surface.py:41` returns `set(mod.__all__)` when `__all__` exists, so a name that is declared but unbound is counted as present.

## Expected

[...]

PR #4027, opened 2026-08-26
ci(policy): state the import-surface boundary and check it

ducha-aiki:

#3986 broke because `kornia.geometry.transform.pyramid.pad` was importable without ever being API — never in `__all__`, never documented, only `torch.nn.functional.pad` leaking out of a `from kornia.core import ..., pad, ...` line. A refactor to `F.pad(...)` dropped the binding and ComfyUI-LTXVideo was broken for three months across five ignored PRs.

An audit of the whole tree says `pad` was not special:

- the same release removed **all 32 names from `kornia.core.__all__`** with no shim and no release note, taking **824 module-scope bindings** with it across ~500 modules (#4024) — `pad` in pyramid is 1 of 19 lost `pad` bindings;
- a performance PR deleted `upscale_double`, which **was documented** and therefore was public API by the policy's own definition (#4025);
- three names sit in `__all__` today without resolving, so `from kornia.color import *` and `from kornia.core import *` raise (#4023);
- 57 more names are public-named, importable, and deliberately excluded from their module's `__all__` (#4026).

None of it was visible in review, because nothing compared the export surface between two revisions.

This PR adds the two halves that were missing. Neither works without the other: the check needs a stated rule to enforce, and the rule needs enforcement to be more than a wish.

**1. The boundary, written down** — `docs/source/get-started/stability.rst` gains three short sections:

- *Importability is not publication.* The page already says documented-means-public; this makes `__all__` the machine-readable half of that sentence, binding in both directions — a name in `__all__` gets promise 1, a name in neither `__all__` nor the docs may go in any release. `pad` is the worked example.
- *Names removed before this section was written.* States plainly that publishing the rule does not retroactively restore what earlier releases took, and why for each: the `kornia.core` 32 were never documented and every one has a direct torch replacement, so restoring them would rebuild the alias layer whose removal was the point; `upscale_double` was a leftover from the OpenCV-mimicking DoG implementation, superseded when that code was improved. Both stay removed, and both get the release-note line promise 3 requires.
- *If you depend on an undocumented name.* Tells downstream to open an issue and ask for publication rather than discovering at upgrade time that the name was never ours to keep.

**2. The check** — `.github/scripts/check_import_surface.py`, modelled on the existing `check_doc_links.py`. Stdlib-only, never imports kornia, reads any revision via `git show`, so there is nothing to install and no environment to get wrong.

On a pull request it compares against the **merge base** (not the base tip, so it sees what this PR removes rather than what the branch already lost) and **fails when a name in `__all__` disappears**.

The deliberate part: **undeclared removals are reported but do not fail.** They are not API under the policy, and failing on them would freeze 2144 incidental third-party re-exports as permanent surface — the check would be turned off within a month. So `pad` itself would have produced a warning line, not a red build. That is the honest outcome and it is still the one that would have helped: the reviewer of that PR would have seen "19 `pad` bindings removed" in the log. `upscale_double` **would** have failed the build.

`kornia.contrib` is excluded from the fatal path, matching its Experimental tier ("No stability promise") on the same policy page.

Refs #3986 (the break), #4023, #4024, #4025, #4026 (what the audit found). Complements #3898's `tests/test_api_surface.py`, which guards 12 top-level modules by runtime `__all__` and is the reason this script does not duplicate that job — note that guard cannot catch the `kornia.core` removal, because its baseline was recorded after the removal, nor the three broken exports, because it reads `__all__` without checking the names resolve (both tracked in #4023).

## What did you check?

Validated against the two known breaks and the current tree:

```text
$ python3 .github/scripts/check_import_surface.py regression --old v0.8.2 --new v0.8.3
declared (in __all__) names removed: 77
  kornia.color.AUTUMN  (was bound by: import-kornia)
  kornia.geometry.transform.pyramid.upscale_double  (was bound by: def)
  ...
undeclared public names removed: 1317
experimental-tier (kornia.contrib) names removed, not fatal: 396
$ echo $?
1

$ python3 .github/scripts/check_import_surface.py regression --old main --new HEAD
declared (in __all__) names removed: 0
undeclared public names removed: 0
$ echo $?
0

$ python3 .github/scripts/check_import_surface.py inventory
modules scanned: 500 · without __all__: 368
L1a — re-exported third-party/stdlib name: 2143 names in 425 modules
L1b — re-exported kornia-internal name: 1535 names in 373 modules
L2 — defined here, module has __all__ and excludes it: 57 names in 27 modules
L3 — module declares no __all__: 1148 names in 323 modules
X  — in __all__ with no module-scope binding: 3
     kornia.color.AUTUMN  ('from kornia.color import *' would raise AttributeError)
     kornia.core.eye_like  ('from kornia.core import *' would raise AttributeError)
     kornia.core.vec_like  ('from kornia.core import *' would raise AttributeError)
```

Each of those three was confirmed by hand at runtime (`from kornia.core import *` → `AttributeError: module 'kornia.core' has no attribute 'eye_like'`). The inventory step in the workflow is non-fatal for exactly that reason — it would be red on main today until #4023 lands, after which it should be made fatal.

Also ran:

```text
$ pre-commit run --files .github/scripts/check_import_surface.py .github/workflows/import-surface.yml docs/source/get-started/stability.rst
... ruff check Passed · ruff format Passed · check yaml Passed · codespell Passed · Add license header Passed
```

**Not checked:** I did not run a full Sphinx build for the docs change (`pixi run build-docs`); I verified section-underline lengths and inline-literal balance by script instead. Worth a docs CI pass before merge.

## How did AI help?

AI-assisted end to end. Claude wrote the audit script and the policy text, ran the cross-revision comparisons, and verified each claimed break at runtime rather than from the static output. I reviewed the findings and supplied the `upscale_double` provenance (OpenCV-mimic DoG leftover, superseded), which is what settles it as "stays removed" rather than "needs restoring".

### DIFF
--- a/.github/scripts/check_import_surface.py
+++ b/.github/scripts/check_import_surface.py
@@ -1,0 +1,361 @@
+#!/usr/bin/env python3
+
+# LICENSE HEADER MANAGED BY add-license-header
+#
+# Copyright 2018 Kornia Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Check that a change does not silently remove an importable name.
+
+Written after #3986, where ``kornia.geometry.transform.pyramid.pad`` stopped
+existing. It was never in ``__all__`` and never documented -- it was only
+``torch.nn.functional.pad`` leaking out of a ``from kornia.core import ...,
+pad, ...`` line -- so when the module switched to calling ``F.pad`` directly,
+the binding went with it. ComfyUI-LTXVideo was broken for three months.
+
+``pad`` was not special. The same release removed all 32 names from
+``kornia.core.__all__`` (#4024), which took 824 module-scope bindings with it
+across the tree, and a performance PR deleted the documented ``upscale_double``
+(#4025). None of it was visible in review, because nothing compared the export
+surface between two revisions.
+
+This script is that comparison. It is stdlib-only and never imports kornia, so
+it runs on any interpreter, needs no environment, and can read any git revision
+through ``git show``.
+
+Two modes:
+
+``regression``
+    Public module-scope names importable at ``--old`` and gone at ``--new``.
+    By default only losses of names that were in their module's ``__all__``
+    fail the run; undeclared losses are reported but do not fail. That split is
+    deliberate and matches the stability policy: what a module does not declare
+    is not API. Use ``--fail-on any`` to make every loss fatal.
+
+``inventory``
+    The undeclared public surface at one revision, graded:
+
+    ``L1a``  re-exported third-party/stdlib name (the #3986 class)
+    ``L1b``  re-exported kornia-internal name
+    ``L2``   defined here, module has ``__all__`` and excludes it (#4026)
+    ``L3``   module declares no ``__all__`` at all
+    ``X``    name in ``__all__`` with no module-scope binding -- a live bug,
+             since ``from <module> import *`` raises (#4023)
+
+Names bound only under ``if TYPE_CHECKING:`` are ignored throughout: they are
+not importable at runtime, so they are neither surface nor regression. A module
+with a module-level ``__getattr__`` is treated as shimmed -- that is the
+correct deprecation pattern (``kornia.utils`` uses it) and the check must not
+punish it.
+
+Usage::
+
+    python3 .github/scripts/check_import_surface.py regression --old main --new HEAD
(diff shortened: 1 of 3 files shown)


ducha-aiki:
Closing this in favour of #4022 by @shlokkvaishnav, which proposed the same check first.

The timing here is entirely on our side, and worth stating plainly: #4022 was opened at 11:24 UTC, this PR at ~11:38. He proposed the check, asked for scope agreement before implementing — which is what `CONTRIBUTING.md` asks for — drafted it anyway to make the shape concrete, and then waited. Meanwhile this PR landed on top of it from the maintainer side. That is the wrong way round, and the fix is for his work to be the one that lands, not ours.

His draft is good, and it reaches the same design conclusion independently: `ast` plus `git show`, no kornia import, `__all__` removals hard-fail, undeclared removals informational. It also has 164 lines of unit tests, including an end-to-end test against a throwaway git repo reproducing the `pad` shape — which is more than this PR had.

Review notes from testing his branch, and the answers to his three open questions, are on #4022.

The `stability.rst` half of this PR is deferred rather than dropped: the `__all__`-versus-documentation boundary is real policy and will be written separately, not attached to a CI change.

The audit behind this PR is recorded in #4023, #4024 (closed — `kornia.core` was an artifact of the abandoned multi-backend work), #4025 (closed — superseded DoG helper) and #4026.

Issue #1396, opened 2021-10-16
Gradients from Boxes3D.to_tensor aren't accurate


hal-314:
### Describe the bug

Landing issue for those wanting to compute gradients in for Boxes3D as discussed in #1304.

Gradients from `Boxes3D.to_tensor` aren't accurate as analytical gradient doesn't match with numerical gradient. At the time when Boxes and Boxes3D were implemented (PR #1304), we didn't know how to fix them. Feel free to fix them :) .

Exactly, the error happens in this line of kornia.geometry.bbox_v2.
 
```python
    boxes = torch.stack([batched_boxes.amin(dim=-2), batched_boxes.amax(dim=-2)], dim=-2).view(
            batched_boxes.shape[0], batched_boxes.shape[1], 6
        )
```
Here, computed and analytical gradients diverge. This only happens for Boxes3D and not Boxes.  See discussion in PR #1304 when Boxes and Boxes3D were implemented for more information. 

### Reproduction steps 

Below there is a code to reproduce the error:

Working code for 2D boxes
```python
import torch
from torch.autograd import gradcheck

def tensor_to_gradcheck_var(tensor, dtype=torch.float64, requires_grad=True):
    """Convert the input tensor to a valid variable to check the gradient.

    `gradcheck` needs 64-bit floating point and requires gradient.
    """
    if not torch.is_tensor(tensor):
        raise AssertionError(type(tensor))
    return tensor.requires_grad_(requires_grad).type(dtype)

t_boxes2D = torch.tensor([[[1.0, 1.0], [3.0, 1.0], [3.0, 2.0], [1.0, 2.0]]])
t_boxes2D = tensor_to_gradcheck_var(t_boxes2D)

def to_tensor(batched_boxes):
    batched_boxes = batched_boxes if batched_boxes.ndim == 4 else batched_boxes.unsqueeze(0)
    
    boxes = torch.stack([batched_boxes.amin(dim=-2), batched_boxes.amax(dim=-2)], dim=-2).view(
            batched_boxes.shape[0], batched_boxes.shape[1], 4
        )
    return boxes

gradcheck(to_tensor, t_boxes2D)  # Ok when boxes are 2D
```
The same code for 3D boxes fails :/ 
```python
t_boxes3d = torch.tensor([[[0, 1, 2], [10, 1, 2], [10, 21, 2], [0, 21, 2], [0, 1, 32], [10, 1, 32], [10, 21, 32], [0, 21, 32]]], dtype=float)
[...]

PR #4028, opened 2026-08-26
docs(geometry): retire the #3960 shape-guard warnings

### LINKED ISSUE #3960: normalize_homography's shape guard uses 'or': (3,3) and (2,4,3,3) silently pass while (B,4,4) escapes and dies inside matmul

ducha-aiki:
## Summary

#3999 fixed the three homography shape guards but left the docstrings describing the behavior it removed. Three sites in `kornia/geometry/conversions.py` currently document warts that no longer exist, in rendered API documentation on `main`.

Executed against `main` (`3c1d7265`, torch 2.9.1, cpu):

| Call | `main` today | Docstring on `main` |
|---|---|---|
| `normalize_homography(eye(3).expand(2, 4, 3, 3), (4, 5), (8, 9))` | `ValueError: ... must be a Bx3x3 tensor` | "any number of leading batch dimensions is accepted and preserved (`(2, 4, 3, 3)` returns `(2, 4, 3, 3)`)" |
| `normalize_homography(eye(4)[None], (4, 5), (8, 9))` | `ValueError: ... must be a Bx3x3 tensor` | "a `(B, 4, 4)` input passes the guard entirely and fails later inside `matmul`" |
| `normalize_homography3d(eye(3)[None], (2, 4, 5), (3, 8, 9))` | `ValueError: ... must be a Bx4x4 tensor` | "its message names the wrong shape — `Input dst_pix_trans_src_pix must be a Bx3x3 tensor`" |

## Changes

**`kornia/geometry/conversions.py`** — the instance:

- `normalize_homography`: the `.. warning::` becomes a Convention bullet stating what the guard now accepts and rejects, with the pre-#3999 behavior kept as a one-clause note so the change stays findable from the docs.
- `normalize_homography3d`: same, with the corrected `Bx4x4` message named.
- `denormalize_homography`: drops the stale `#3960` cross-link from its inherited-warnings list — it inherits `normalize_homography`'s guard, which is no longer a defect. The remaining `#3958`/`#3957`/`#3959`/`#3904` links are untouched.

**`CHANGELOG.md`** — the announcement #3999 never got:

#3960 offered two outcomes and recommended option 1 (ratify the wider `(*, 3, 3)` contract). #3999 took option 2 (tighten), which is a break: executed at `11d8d393^`, all three functions returned a rank-4 input with its rank unchanged (`normalize_homography(eye(3).expand(2, 4, 3, 3), (4, 5), (8, 9))` → `(2, 4, 3, 3)`), and all three now raise. It merged unannounced, so this adds a **Breaking changes** entry.

**`AGENTS.md`** — the guardrail, one bullet in *Documentation and generated examples*:

> before merging a change that closes one of those issues, grep the tree for the issue number; if it is still there, the change is incomplete.

This is deliberately in the same PR as the instance it prevents. #3999 merged cleanly and passed every check — the docstrings sit hundreds of lines from the guards, so nothing in the diff or in CI pointed at them. Rather than hardcode a module list that drifts, the bullet names the grep that finds them (`grep -rn "github.com/kornia/kornia/issues/" kornia/`, five files today: `conversions.py`, `boxes.py`, `transform/crop2d.py`, `transform/pyramid.py`, `transform/thin_plate_spline.py`), and it names the paired `test_wart_*_<issue>` / `test_convention_*_<issue>` pins, which need flipping in the same change for the same reason. Happy to split it out if you would rather keep this PR to `conversions.py`.

## Validation

Documentation only — no behavior change.

- Every claim in the new text executed on this branch, for all three functions: unbatched `(N, N)` accepted and promoted to `(1, N, N)`, batched `(B, N, N)` accepted, rank-4 rejected, wrong-sized rejected, and the 3-D message naming `Bx4x4`.
- The pre-#3999 claims executed at `11d8d393^`: `normalize_homography` returned `(2, 1, 3, 3)` unchanged in rank, and `(1, 4, 4)` reached `matmul` (`RuntimeError: Expected size for first two dimensions of batch2 tensor to be: [1, 4] but got: [1, 3]`).
- Both new bullets restate behavior `test_convention_the_guard_accepts_rank_2_and_3_and_rejects_higher_3960` and `test_convention_normalize_homography3d_shape_error_names_bx4x4_3960` already pin, so no new test is warranted.
- The `#3999` break in the changelog entry executed at `11d8d393^` for all three functions: `(2, 4, 3, 3)`, `(2, 4, 3, 3)` and `(2, 4, 4, 4)` returned, against `ValueError` on `main`.
- `pixi run lint`, `ruff format --check`, `codespell`, `pytest --doctest-modules kornia/geometry/conversions.py` (34 passed), `pytest tests/geometry/test_conversions.py` (372 passed, 17 xfailed), the seven `#3960` pins (7 passed).

## Merge order — resolved

#4006 merged first as `4dd9c6aa`, as recommended, and `origin/main` is merged in here at `b5c5ef38`. Both predicted conflicts resolved:

- `kornia/geometry/conversions.py` — `denormalize_homography`'s inherited-warnings list. The merged result drops **both** links: `#3960` (this PR) and `#3957` (fixed by #4006). `grep -rn "3957\|3960" kornia/` is now clean.
- `CHANGELOG.md` — one `## Unreleased` heading carrying `### Breaking changes` (this PR) above `### Bug fixes` (#4006).

Two defects found in the final review pass and fixed in `10200a2f`:

- The AGENTS.md guardrail bullet named the pattern `.. warning::` … `Tracked in #NNNN`, which matches only `conversions.py` and `crop2d.py` — `pyramid.py` and `thin_plate_spline.py` use a lowercase `tracked in` clause, so a grep for the stated phrase missed two of the four modules the bullet itself listed. The list also omitted `boxes.py`, which tracks #1396 in a `Note:` block and in a `RuntimeError` message. The bullet now says to grep the issue **number**, not a phrase, and lists five modules.
- `normalize_homography3d`'s new bullet said "a rank-4 input", which collides with the 4x4 matrices that function takes. It now names the `(2, 4, 4, 4)` example.

Re-verified on the merged tree (torch 2.9.1, cpu): all twelve guard cells across the three functions — `(N, N)` promoted to `(1, N, N)`, `(B, N, N)` accepted, rank-4 and wrong-sized rejected, and the 3-D message naming `Bx4x4`; `pytest --doctest-modules kornia/geometry/conversions.py` (34 passed), `tests/geometry/test_conversions.py` (377 passed, 17 xfailed), the seven `#3960` pins, and `pre-commit` (ruff check, ruff format, codespell, license header) on all three changed files.

## Review round 2 — `63cc4048`

Three findings from a review pass on the merged tree, all fixed in `63cc4048`:

- **`conversions.py`** — `denormalize_homography`'s Convention bullet read "and so is **its** shape guard", where `its` parses as `normalize_homography`'s. It now says the guard is the same one and rejects the same shapes.
- **`AGENTS.md`** — the guardrail bullet hardcoded its five modules as bare basenames next to a repo-root path, which implies `crop2d.py`, `pyramid.py` and `thin_plate_spline.py` sit in `kornia/geometry/` when they are under `kornia/geometry/transform/`. An agent following the instruction literally greps a path that does not exist and concludes the reference is gone — the exact false negative the rule exists to prevent. The list is replaced by the self-maintaining grep, and the #3999 post-mortem is dropped; the bullet is now about half its length.
- **`CHANGELOG.md`** — two problems. The entry mapped three function names onto two shape pairs with a "respectively" that cannot be resolved positionally, so it could be read as saying `denormalize_homography` now takes 4x4; the two clauses are now split. And it omitted a real user-visible delta of #3999: the old guard was `not (len(shape) == 3 or shape[-2:] == (3, 3))`, whose `or` short-circuited past the tuple comparison for a rank-3 input, while the new `not (ndim in (2, 3) and shape[-2:] == (3, 3))` always evaluates it. Executed at `11d8d393^` and on this branch (torch 2.9.1, cpu): tracing `normalize_homography` produced no `TracerWarning` before and produces one now at `conversions.py:1990`, and the same at `:2359` and `:2445` for the other two (the `utils.py:216` warning is pre-existing in `_torch_inverse_cast`). The traced graph is unchanged, so the entry records it as noise rather than a correctness signal.

Re-verified after the edits: `pre-commit` (all hooks) on the three changed files, `pytest --doctest-modules kornia/geometry/conversions.py` (34 passed), `tests/geometry/test_conversions.py` (377 passed, 17 xfailed), the seven `#3960` pins, and the twelve guard cells again. The prescribed grep returns exactly the five files with correct paths.

Unrelated observation, not touched here: `kornia/geometry/boxes.py:1062` renders its link as `#1304` while pointing at `.../issues/1396`.

Closes #3960.


ducha-aiki:
## Review

Docs-only, and every claim in the new text executes as written. Verified at `48c39915`, torch 2.9.1, cpu.

### What I checked

Executed all three functions against the guard as it stands (`conversions.py:1864`, `2176`, `2282` — `ndim in (2, 3) and shape[-2:] == (N, N)`):

| call | result | matches the new text |
|---|---|---|
| `normalize_homography((3,3))` | `(1, 3, 3)` | ✔ "promoted to `(1, 3, 3)`" |
| `normalize_homography((2,3,3))` | `(2, 3, 3)` | ✔ |
| `normalize_homography((2,4,3,3))` | `ValueError: Input dst_pix_trans_src_pix must be a Bx3x3 tensor. Got torch.Size([2, 4, 3, 3])` | ✔ names the argument and the expected shape |
| `normalize_homography((1,4,4))` | same `ValueError` | ✔ |
| `denormalize_homography`, all four | identical | ✔ "and so is its shape guard" |
| `normalize_homography3d((4,4))` / `((2,4,4))` | `(1, 4, 4)` / `(2, 4, 4)` | ✔ |
| `normalize_homography3d((2,4,4,4))` / `((1,3,3))` | `ValueError: ... must be a Bx4x4 tensor` | ✔ |

- **The pre-#3999 claims** match the observed block recorded in #3960 verbatim: `(2, 4, 3, 3)` came back as `(2, 4, 3, 3)` (rank unchanged) and `(1, 4, 4)` died inside `matmul`.
- **Attribution is right and complete.** `git blame` puts the current `ndim in (2, 3) and ...` line at `11d8d393` (#3999). Worth noting because #3999's own description says `len(shape) == 3 and ...`, which would have *rejected* the unbatched `(3, 3)` — the rank-2 branch arrived during that PR's review, so "fixed in #3999" needs no second link.
- **Nothing stale left behind.** `grep 3960` over `kornia/` is now empty; the only survivors are the test pins, which correctly keep the issue number. No dangling cross-reference either — `normalize_homography`'s "see the convention warning below" still resolves to the `#3904` corner-alignment warning, which this PR does not touch.
- **The tests already cover the new text**, so "no new test is warranted" holds: `test_convention_the_guard_accepts_rank_2_and_3_and_rejects_higher_3960` pins rank-2 acceptance, the `(1, N, N)` return and rank-4 rejection; `test_convention_normalize_homography3d_shape_error_names_bx4x4_3960` pins the message.
- `ruff check` and `ruff format --check` clean; `pytest --doctest-modules kornia/geometry/conversions.py` 34 passed; the seven `#3960` pins pass; the RTD preview builds and all three bullets render as bullets with working links (checked in the built HTML).

### Nits — none blocking

**1. `normalize_homography3d`, lines 2239–2242: "the ``or``-structured guard" reads as present tense.** The bullet opens on the current contract ("a `(4, 4)` **or** `(B, 4, 4)` matrix") and closes with "Both the ``or``-structured guard and its message … were fixed in #3999". Since the `or`-vs-`and` structure *is* what #3960 was about, a reader can take "the ``or``-structured guard" for a description of today's guard. The 2-D bullet sidesteps this by never naming the structure ("Rank-4 inputs and `(B, 4, 4)` inputs used to pass this guard"). Suggested rewrite:

> Both the guard's old ``or``-structure and its message, which used to name ``Bx3x3`` from a function that takes 4x4 matrices, were fixed in `#3999`

**2. `Args:` / `Returns:` still say only `(B, 3, 3)`.** The new Convention bullet states that an unbatched `(3, 3)` is accepted, but `dst_pix_trans_src_pix: … :math:`(B, 3, 3)`` (line 1854) and the `Returns:` line are unqualified, as is `normalize_homography3d`'s `Shape: Output: (B, 4, 4)`. Not wrong, just narrower than the ratified contract — and "documented as `(B, 3, 3)` while the guard accepts more" is the exact framing of the issue being closed here, so widening those lines would finish the job. Optional: most of `kornia.geometry` lists only the canonical shape in `Args`.

**3. Cosmetic.** Line 2127 (`too. The exception is the closed-form-inverse warning: in eager mode this`) wraps at 83 characters where the rest of that block sits at 80 or under. Well inside the 120 limit, so lint is happy — just an uneven reflow.

### One note, not a change request

#3960 laid out two outcomes and recommended option 1 (ratify the wider `(*, 3, 3)` contract). What landed in #3999 is option 2 (tighten), which the issue itself placed in the coordinated Repair & Deprecation window: rank-4 callers that worked before now raise. That break is already merged and this PR only documents it — and the "used to pass this guard … fixed in #3999" clause is the right way to keep it findable from the rendered docs. Worth a release-note line for 0.9.0.

ducha-aiki:
## Addressed in `77e5f209`

All three nits, plus the release note. The review was at `48c39915`, so it did not see `aa03cb7f` — an `AGENTS.md` bullet added after it, written up in the PR description.

**1. Present-tense `or`-structure — fixed.** Took the suggested phrasing almost verbatim: `Both the guard's old `or`-structure and its message, which used to name `Bx3x3` ... were fixed in #3999`. The diagnosis was right about why only the 3-D bullet had the problem — the 2-D one never names the structure, so there was nothing to misread.

**2. `Args:` narrower than the ratified contract — fixed, `Returns:`/`Shape:` deliberately not.** All three `Args` lines now name the unbatched form and its promotion:

> `normalized. :math:`(B, 3, 3)`, or an unbatched :math:`(3, 3)`, which is promoted to :math:`(1, 3, 3)``

Agreed this one is load-bearing rather than optional — "documented as `(B, 3, 3)` while the guard accepts more" is #3960's own framing, so leaving `Args` narrow would have closed the issue with the discrepancy still on the page. `Returns:` and `normalize_homography3d`'s `Shape: Output:` are left alone on purpose: the output really is always batched, so `(B, 3, 3)` is already exact there, with `B = 1` for the promoted case. Widening those would have made them *less* accurate.

**3. Uneven reflow — fixed.** That block now runs at 80 characters or under throughout; the first pass at it left an orphaned `through the`, so it got a full reflow rather than a one-line patch.

### The release note — promoted from a note to a change

This one is worth more than the nits, so it is now a `## Unreleased` / `### Breaking changes` entry rather than a follow-up. The point that #3999 took #3960's option 2 over the recommended option 1 is exactly right, and it merged with no changelog line at all.

Executed at `11d8d393^`, so the entry states a measured break rather than an inferred one — **all three** functions took a rank-4 input and returned it with its rank unchanged:

| call | `11d8d393^` | `main` |
|---|---|---|
| `normalize_homography(eye(3).expand(2, 4, 3, 3), (4, 5), (8, 9))` | `(2, 4, 3, 3)` | `ValueError` |
| `denormalize_homography(...)`, same input | `(2, 4, 3, 3)` | `ValueError` |
| `normalize_homography3d(eye(4).expand(2, 4, 4, 4), (2, 4, 5), (3, 8, 9))` | `(2, 4, 4, 4)` | `ValueError` |

A caller passing rank-4 got a plausible-looking matrix back, which is the kind of break that needs announcing rather than the kind that shows up as a crash on upgrade.

**Cost, stated plainly:** this makes `CHANGELOG.md` a second conflict with #4006, which opens its own `## Unreleased` block at the same anchor. Re-measured with `git merge-tree` at `77e5f209` vs `5c75aacb` — still exactly two hunks, one per file, and the changelog one resolves by keeping both `###` subsections under one heading. Say the word if you would rather this entry rode along with #4006 instead and I will drop it here.

### Re-verified at `77e5f209`

`ruff check` / `ruff format --check` clean, `codespell` clean, `pytest --doctest-modules kornia/geometry/conversions.py` 34 passed, the seven `#3960` pins pass, `grep 3960 kornia/` still empty.

---

ducha-aiki (commented):
The AGENTS.md guidance should make test-pin discovery name-agnostic. It currently assumes every tracked defect has paired tests named `test_wart_*_<issue>` or `test_convention_*_<issue>`, but the tree already violates that convention: issue #3928’s strict xfail is `test_convention_default_align_corners_reproduces_identity` without the numeric suffix, while the #3962 documentation explicitly has no corresponding pin. An agent following this guidance could therefore miss an existing xfail or search for a nonexistent pair. Please instruct agents to search `tests/` for the issue number and describe these naming patterns as examples used where present.

ducha-aiki:
## Addressed in `4216da0f`

All three claims check out against the tree, so the bullet was wrong as written, not just narrow:

| claim | in the tree |
|---|---|
| #3928's xfail carries no numeric suffix | `test_convention_default_align_corners_reproduces_identity` (`tests/geometry/transform/test_thin_plate_spline.py:358`) — the number lives in the `xfail` `reason=` and in two comment lines, nowhere in the name |
| #3962 has no pin | `tests/geometry/test_conversions.py:3991` says so explicitly: *"No pin asserts anything about kornia#3962 … a missing symbol is a scope question, not a defect"* |
| the naming convention is not universal | 39 `test_convention_*` names carry a four-digit suffix; the rest do not, because most of them pin a convention that never had an issue |

Two more that the same search turned up: **#3929 and #3927 have no pin at all** — zero hits in `tests/` for either — so "search for the pair" would have sent an agent looking for something that was never there, twice more.

**What the bullet says now.** One grep, over `kornia/` and `tests/` together, matching the issue number; the naming patterns are demoted to one example among the several places the number actually appears (`Tracked in #NNNN`, a lowercase `tracked in` clause, a `Note:` block, a runtime error message, an `xfail` `reason=`, a plain comment, a name suffix *where that convention was followed*). And the closing sentence makes an empty `tests/` result a valid answer rather than a failed search.

**Two things the rewrite had to get right beyond the report:**

- *A bare-number grep is not usable.* The old bullet said "grep the tree for the issue number", which taken literally hits float literals: `3960` matches `kornia/color/_colormap_data.py` twice (`0.9583313515139603`, `0.07215778103960274`) and `3927` matches `tests/geometry/camera/test_stereo.py:139` (`-19.3927`). The bullet now prescribes `grep -rnE "#NNNN|issues/NNNN" kornia/ tests/`, which is exactly clean on every number I tried. It has a second payoff: it finds `kornia/geometry/boxes.py:1062`, where the rendered label says `#1304` over a URL pointing at `issues/1396` — a label-matching grep misses that one, a URL-matching grep does not.
- *The two directories need different verdicts.* My first draft extended "if it is still there, the change is incomplete" to `tests/`, which is wrong: pins keep naming their issue after the fix — the seven `#3960` pins on this branch do exactly that. The bullet now separates them: a surviving hit in `kornia/` means the change is incomplete, while a hit in `tests/` is a pin to re-check (drop a now-XPASSing strict `xfail`, invert a wart assertion), not a leftover to delete.

Verified after the edit: `pre-commit` (all hooks) on the three changed files, and the prescribed grep run verbatim for #3960, #3928, #3962, #3929, #3927, #1396, #3999 and #3957 — zero false positives in all eight.

Posted on behalf of [@ducha-aiki](https://github.com/ducha-aiki) by Claude (Fable 5).

ducha-aiki:
## Review — round 3, at `4216da0f`

Docs-only. I re-executed every claim in the three files rather than reading them against the earlier rounds' tables, on the merged tree (torch 2.9.1, cpu) and at `11d8d393^` (`4da60edc`) for the "before" half. Everything in the repo holds. Two corrections below are to the **PR description**, not to anything that ships.

### The docstring claims

All fifteen cells, run against the guards as they stand (`conversions.py:1990`, `:2360`, `:2446` — all three literally `ndim in (2, 3) and shape[-2:] == (N, N)`, so "this function carries the same one" and "the 2-D function's, with the size adapted" are exact):

| input | `normalize_homography` | `denormalize_homography` | `normalize_homography3d` |
|---|---|---|---|
| unbatched | `(3, 3)` → `(1, 3, 3)` | `(3, 3)` → `(1, 3, 3)` | `(4, 4)` → `(1, 4, 4)` |
| batched | `(2, 3, 3)` → `(2, 3, 3)` | `(2, 3, 3)` → `(2, 3, 3)` | `(2, 4, 4)` → `(2, 4, 4)` |
| rank-4 | `ValueError … Bx3x3` | `ValueError … Bx3x3` | `ValueError … Bx4x4` |
| wrong-sized | `(1, 4, 4)` → `ValueError … Bx3x3` | same | `(1, 3, 3)` → `ValueError … Bx4x4` |
| rank-1 | `ValueError … Bx3x3` | same | `ValueError … Bx4x4` |

Every message names the argument and the expected shape, and it is a plain `ValueError`, not a `BaseError` subclass — which is what the bullets say. The rank-1 row is not claimed by any bullet; I ran it because "and nothing else" is a universal, and it holds.

At `4da60edc` the guard is `not (len(shape) == 3 or shape[-2:] == (3, 3))` and the "used to" clauses reproduce: `(2, 4, 3, 3)` came back as `(2, 4, 3, 3)` from both 2-D functions, `(2, 4, 4, 4)` as `(2, 4, 4, 4)` from the 3-D one, and `(1, 4, 4)` died inside `matmul` (`Expected size for first two dimensions of batch2 tensor to be: [1, 4] but got: [1, 3]`). The changelog's restriction of that second case to a *batched* wrong-sized input is right and load-bearing: an unbatched `(4, 4)` was rejected by the old guard too, since `len == 2` fails the first clause and `(4, 4) != (3, 3)` the second.

### The tracing paragraph

This is the one claim that needed isolating, because the two trees differ by more than #3999. Traced the guard forms directly, same body, rank-3 input:

| guard | TracerWarnings | traced graph |
|---|---|---|
| `len(shape) == 3 or shape[-2:] == (3, 3)` | 0 | `aten::matmul(%x, %x)` |
| `ndim in (2, 3) and shape[-2:] == (3, 3)` | 4 × "Converting a tensor to a Python boolean" | `aten::matmul(%x, %x)` |
| no guard | 0 | `aten::matmul(%x, %x)` |

Identical modulo SSA numbering, so **"the traced graph is unchanged — the warning is noise" is confirmed**, not inferred. One scope note, already carried by the sentence's own "for a rank-3 input" clause: trace a 2-D function with an unbatched `(3, 3)` and the *old* form evaluated the comparison too, so it warned as well. The delta is specific to the batched case.

### Two corrections — to the PR description only

1. **"tracing `normalize_homography` produced no `TracerWarning` before" is not what `4da60edc` does.** It produced four, of a different class: *"torch.tensor results are registered as constants in the trace"* at `conversions.py:2015`. Those come from `normal_transform_pixel`, and they disappear on this branch because of **#4006**, not #3999. The CHANGELOG entry is precise where the description is loose — it claims only a warning "about converting a tensor to a Python boolean that it did not emit before", and that is exactly true. Nothing to change in the tree; the description overstates a measurement it got right in the file.
2. **The cited line numbers are one low.** The warning is attributed to the `if`, not the `raise`: `:1990`, `:2360`, `:2446`, against the description's `:1990`, `:2359`, `:2445`.

And one thing the "before" measurement could not have been clean at that revision: at `4da60edc`, `torch.jit.trace(normalize_homography3d)` does not merely warn, it **fails** — `TracingCheckError: Tensor-valued Constant nodes differed in value across invocations`, from `normal_transform_pixel3d` at `conversions.py:2089`. It traces without error on this branch. That is #4006's doing and nothing this PR touches, but it is worth recording that the 3-D "before" is a failed trace rather than a quiet one.

### Nits — neither blocking, neither worth a commit on its own

1. `normalize_homography3d`'s Convention block still opens a bullet with "the matrices are :math:`(B, 4, 4)` and act on ``(x, y, z, 1)`` column vectors", three items above the new guard bullet that says an unbatched `(4, 4)` is accepted. It is the same narrowness that round 2 fixed in `Args:`, and that bullet is about layout rather than the guard, so it reads as harmless — but the two sit in one block. `normalize_homography` has no counterpart bullet, so the asymmetry is only in the 3-D docstring.
2. `kornia/geometry/boxes.py:1062` renders `#1304` over a URL pointing at `issues/1396` — already noted in the description as out of scope. Recording it again only because the new AGENTS.md bullet is now the thing that will catch it: it prescribes matching `issues/NNNN` as well as `#NNNN`, precisely so a wrong label cannot hide a live reference.

### Checks

`pre-commit` (all hooks) on `AGENTS.md`, `CHANGELOG.md` and `kornia/geometry/conversions.py`; `pytest --doctest-modules kornia/geometry/conversions.py` 34 passed; `tests/geometry/test_conversions.py` 377 passed, 17 xfailed; the seven `#3960` pins pass; `grep -rnE "#3960|issues/3960" kornia/` empty, 13 hits in `tests/`, all pins.

No blocking findings. The three files say what the code does.

Posted on behalf of [@ducha-aiki](https://github.com/ducha-aiki) by Claude (Fable 5).

Issue #4022, opened 2026-08-26
Proposal: CI check to flag module-level names dropped by a refactor


shlokkvaishnav:
Follow-up to @ducha-aiki's comment on #4003 (closed) — the real problem behind #3986 wasn't the missing `pad` symbol itself, it's that nothing catches a refactor silently removing a name a module used to expose (whether or not it was ever in `__all__`).

## Proposal

A CI check, run on every PR, that:

1. For each changed file under `kornia/**/*.py`, statically parses (via `ast`, no import/execution needed) the set of module-level names bound in the file — top-level `def`/`class`, assignments, and `import`/`from ... import` targets — at both the PR branch tip and the merge-base commit.
2. Diffs the two sets per file and reports any name present at the merge-base but missing at the tip, split into two severities:
   - **Removed from `__all__`**: a guaranteed break of documented public API — hard fail.
   - **Removed from module scope entirely, but was never in `__all__`** (the `pad` case): informational only — surfaced as a PR comment/annotation, not a blocking failure, since plenty of these removals are intentional internal cleanup.
3. No new runtime dependency: this only needs `ast` + `git show <merge-base>:<path>` from the stdlib, and only runs in CI (never imported into `kornia/`), so it doesn't touch the "only PyTorch as a third-party dependency" rule for the library itself.

This would have caught #3986 automatically as an informational note on the PR that dropped `pad` from `kornia/core`, and would hard-fail on the more serious case of an actual `__all__` regression, without requiring anyone to remember to check by hand.

Open questions I'd want scope agreement on before implementing:
- Should the informational (never-`__all__`) case block the PR (with an explicit override label/comment) or just annotate it? I'd lean toward annotate-only to start, given how often that kind of "accidental" surface is legitimately fine to drop.
- Scope to `kornia/` only, or also `testing/`?
- Where should this live — a script under `scripts/`, or a dedicated GitHub Action?

Happy to open a PR once the shape here is agreed on.

shlokkvaishnav:
Drafted an implementation of this to make the shape concrete: https://github.com/shlokkvaishnav/kornia/tree/feat/module-surface-diff-check

- `scripts/check_module_surface_diff.py` -- pure `ast` + `git show`, no import of kornia needed
- `tests/scripts/test_check_module_surface_diff.py` -- unit tests including an end-to-end test against a throwaway git repo reproducing the exact `pad`/#3986 shape (undocumented name dropped) alongside a documented `__all__` removal
[...]

PR #4029, opened 2026-08-26
ci: check for module-level names dropped by a refactor

shlokkvaishnav:

Follow-up to #3986 / #4025, per @ducha-aiki's scope confirmation and review notes on #4022.

Refactors can silently drop a name a module used to expose at module scope -- whether or not it was ever in `__all__`. `#3986` was one instance (`kornia.geometry.transform.pyramid.pad`, never `__all__`); `#4025` (`upscale_double`) was the more serious one, since it *was* documented and therefore public API by the stability policy's own definition. Nothing compared the export surface across revisions to catch either.

`.github/scripts/check_import_surface.py` is that comparison -- stdlib-only (`ast` + `git show`), never imports kornia, reads any revision, no environment needed. On a PR it diffs the merge-base against the tip and:

- **Fails** when a name disappears from a module's `__all__` (a documented public API break).
- **Reports, doesn't fail**, when a name disappears from module scope but was never in `__all__` -- an inventory of the tree finds 2143+1148 names in this category tree-wide; hard-failing on those would make every incidental import permanent API and the check would get disabled within a month.

Two exemptions from the fatal path, both requested in review:
- A module with a module-level `__getattr__` in the new revision (the `kornia.utils` deprecation-shim pattern from #3898) -- reported, not fatal.
- `kornia.contrib` -- Experimental tier per `stability.rst` ("No stability promise") -- reported, not fatal.

`if TYPE_CHECKING:` bodies are excluded throughout (never bound at runtime); `try`/`except`/`else`/`finally` and plain `if` bodies are otherwise recursed into, since names bound there are real module-scope bindings.

## Validated against

```text
$ python3 .github/scripts/check_import_surface.py --base-ref v0.8.2
::error file=kornia/core/__init__.py::Removed from __all__: [32 names] (#4024)
::error file=kornia/filters/__init__.py::Removed from __all__: ['DexiNed']
::error file=kornia/geometry/transform/pyramid.py::Removed from __all__: ['upscale_double'] (#4025)
$ echo $?
1

$ python3 .github/scripts/check_import_surface.py --base-ref upstream/main
$ echo $?
0
```

The `kornia.utils.__getattr__` false positive flagged in review (`_extract_device_dtype`, `deprecated` removed from `__all__` but still served via the module's shim) is confirmed non-fatal:

```text
removed_from_all: ['_extract_device_dtype', 'deprecated']
new_surface.has_getattr: True
would be fatal: False
```

## Test log

```text
$ pytest tests/scripts/test_check_import_surface.py -v
...
15 passed in 4.48s

$ pre-commit run --files .github/scripts/check_import_surface.py tests/scripts/test_check_import_surface.py .github/workflows/import-surface.yml
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
check yaml...............................................................Passed
mixed line ending........................................................Passed
codespell................................................................Passed
Add license header.......................................................Passed
```

Not checked: I did not confirm behavior on a fresh macOS/Linux checkout beyond CI itself running there -- the encoding fix (explicit UTF-8 for `git show` reads) was needed to unblock local validation on Windows, where `subprocess.run(text=True)` defaults to the platform locale codec.

## What's follow-up, not this PR

- `#4023` (three `__all__` entries with no module-scope binding -- `from <module> import *` currently raises `AttributeError`) is a small, separate PR per the maintainer's suggested split.
- `#4026` group 2 (accidental `logger`/`np`/`ort`/`url` re-exports) is available; groups 1/3/4 are maintainer-held pending the repair-and-deprecation window.

## How did AI help?

AI-assisted. I (Claude) wrote the script, tests, and validated the specific regressions and the reviewer-flagged false positive by actually running the comparisons shown above, not from static reasoning alone. @shlokkvaishnav reviewed the diff and is accountable for it.

### DIFF
--- a/.github/scripts/check_import_surface.py
+++ b/.github/scripts/check_import_surface.py
@@ -212,8 +212,11 @@ def diff_surfaces(old: ModuleSurface, new: ModuleSurface) -> tuple[set[str], set
 def check_file(base_ref: str, path: str) -> FileReport | None:
     """Compare `path`'s module surface between `base_ref` and the working tree.
 
-    Returns None if there's nothing to report (new/deleted file, unparsable
-    source, or no names removed).
+    A deleted file is diffed against an empty surface, so removing a whole
+    module is reported (and can be fatal) just like emptying its `__all__`.
+
+    Returns None if there's nothing to report (new file, unparsable source,
+    or no names removed).
     """
     old_source = _git_show(base_ref, path)
     if old_source is None:
@@ -221,13 +224,13 @@ def check_file(base_ref: str, path: str) -> FileReport | None:
 
     try:
         with open(path, encoding="utf-8") as f:
-            new_source = f.read()
+            new_source: str | None = f.read()
     except FileNotFoundError:
-        return None  # file was deleted; not this check's concern
+        new_source = None  # file was deleted -- diff against an empty surface, not "nothing to report"
 
     try:
         old_surface = parse_module_surface(old_source)
-        new_surface = parse_module_surface(new_source)
+        new_surface = parse_module_surface(new_source) if new_source is not None else ModuleSurface()
     except SyntaxError:
         return None
 
@@ -254,8 +257,18 @@ def main(argv: list[str] | None = None) -> int:
     parser.add_argument("--base-ref", default="origin/main", help="Git ref to diff against (default: origin/main)")
     args = parser.parse_args(argv)
 
-    files = _changed_kornia_files(args.base_ref)
-    reports = [r for r in (check_file(args.base_ref, path) for path in files) if r is not None]
+    # _changed_kornia_files diffs against the merge-base (triple-dot); check_file must read old
+    # content from that same commit, not the tip of base_ref, or the two can disagree whenever
+    # base_ref has moved since the branch point (see review discussion on #4029).
+    base = subprocess.run(  # noqa: S603
+        ["git", "merge-base", args.base_ref, "HEAD"],  # noqa: S607
+        capture_output=True,
+        text=True,
+        check=True,
+    ).stdout.strip()
+
+    files = _changed_kornia_files(base)
+    reports = [r for r in (check_file(base, path) for path in files) if r is not None]
 
     hard_fail = False
     for report in reports:
--- a/.github/scripts/check_import_surface.py
+++ b/.github/scripts/check_import_surface.py
@@ -90,8 +90,12 @@ def _string_literals(node: ast.AST) -> list[str]:
     return [elt.value for elt in node.elts if isinstance(elt, ast.Constant) and isinstance(elt.value, str)]
 
 
-def parse_module_surface(source: str) -> ModuleSurface:
+def parse_module_surface(source: str | bytes) -> ModuleSurface:
     """Parse a module's source and collect the names it binds at module scope.
+
+    Accepts bytes as well as text, and hands them to `ast.parse` undecoded so
+    that a PEP 263 coding cookie is honored exactly as the interpreter honors
+    it -- the same source Python can import is the source this can parse.
 
     Recurses into ``try``/``except``/``else``/``finally`` and plain ``if``
     bodies, since names bound there (e.g. a version-conditional import) are
@@ -146,16 +150,16 @@ def parse_module_surface(source: str) -> ModuleSurface:
     return surface
 
 
-def _git_show(ref: str, path: str) -> str | None:
-    """Return the file's content at `ref`, or None if it didn't exist there."""
-    # git is the trusted tool this script is built around. Force UTF-8 rather
-    # than the platform locale encoding (e.g. cp1252 on Windows), which can
-    # otherwise raise UnicodeDecodeError on non-ASCII source bytes.
+def _git_show(ref: str, path: str) -> bytes | None:
+    """Return the file's bytes at `ref`, or None if it didn't exist there.
+
+    Bytes, not text: decoding is `ast.parse`'s job (see `parse_module_surface`),
+    and decoding here with the platform locale (e.g. cp1252 on Windows) would
+    raise on non-ASCII source bytes that Python itself reads fine.
+    """
     result = subprocess.run(  # noqa: S603
         ["git", "show", f"{ref}:{path}"],  # noqa: S607
         capture_output=True,
-        encoding="utf-8",
-        errors="surrogateescape",
         check=False,
     )
     if result.returncode != 0:
@@ -164,6 +168,7 @@ def _git_show(ref: str, path: str) -> str | None:
 
 
 def _changed_kornia_files(base_ref: str) -> list[str]:
+    """Paths under `kornia/` that differ between `base_ref` and the working tree."""
     # Git's "**" glob pathspec only matches when there's at least one directory
     # between "kornia/" and the filename, so "kornia/foo.py" itself wouldn't
     # match "kornia/**/*.py". Scope to the "kornia/" pathspec instead and
@@ -176,13 +181,25 @@ def _changed_kornia_files(base_ref: str) -> list[str]:
     # fallback (_alternate_path) still tolerates the one legitimate rename
     # shape, `x.py` <-> `x/__init__.py` packageification, so this doesn't
     # trade the rename blind spot for a packageification false positive.
+    #
+    # No `...HEAD`: check_file reads the *working tree*, so the file list has
+    # to be the working tree's diff against base too. Committed-range diffing
+    # would silently skip a removal that is edited but not yet committed --
+    # invisible for the local runs this script documents, identical to
+    # `base...HEAD` on CI's clean checkout (base is already the merge-base).
+    #
+    # -z: without it, git quotes (and octal-escapes) any path holding
+    # non-ASCII or special bytes, so the name arrives wrapped in double
+    # quotes, fails the ".py" filter below, and drops out of the check
+    # entirely. -z also removes the need to un-escape such a name.
     result = subprocess.run(  # noqa: S603
-        ["git", "diff", "--no-renames", "--name-only", f"{base_ref}...HEAD", "--", "kornia/"],  # noqa: S607
+        ["git", "diff", "--no-renames", "--name-only", "-z", base_ref, "--", "kornia/"],  # noqa: S607
         capture_output=True,
-        text=True,
+        encoding="utf-8",
+        errors="surrogateescape",
         check=True,
     )
-    return [line for line in result.stdout.splitlines() if line.endswith(".py")]
+    return [name for name in result.stdout.split("\0") if name.endswith(".py")]
 
 
 # docs/source/get-started/stability.rst puts kornia.contrib in the Experimental
@@ -199,6 +216,17 @@ def _module_name(path: str) -> str:
 
 def _is_experimental(module: str) -> bool:
     return any(module == e or module.startswith(e + ".") for e in EXPERIMENTAL)
+
+
+def _read_source(path: str) -> bytes:
+    """Read a working-tree file as bytes, the way `_git_show` reads the base revision.
+
+    Both ends of the comparison must handle the same bytes. Decoding as UTF-8 text
+    here would raise on a module Python imports happily under a PEP 263 cookie, and
+    abort the whole check with a traceback rather than reporting that one file.
+    """
+    with open(path, "rb") as f:
+        return f.read()
 
 
 def _alternate_path(path: str) -> str:
@@ -239,22 +267,23 @@ def check_file(base_ref: str, path: str) -> FileReport | None:
         return None  # new file, nothing to compare against
 
     try:
-        with open(path, encoding="utf-8") as f:
-            new_source: str | None = f.read()
+        new_source: bytes | None = _read_source(path)
     except FileNotFoundError:
         # The path is gone, but the *module* may not be: x.py <-> x/__init__.py is the same
         # importable name (packageification), not a removal. Only try the one alternate
         # spelling -- anything else missing here is a real deletion.
         try:
-            with open(_alternate_path(path), encoding="utf-8") as f:
-                new_source = f.read()
+            new_source = _read_source(_alternate_path(path))
         except (FileNotFoundError, NotADirectoryError, IsADirectoryError):
             new_source = None  # module genuinely gone -- diff against an empty surface
 
     try:
         old_surface = parse_module_surface(old_source)
         new_surface = parse_module_surface(new_source) if new_source is not None else ModuleSurface()
-    except SyntaxError:
+    except (SyntaxError, ValueError):
+        # SyntaxError: not valid Python at one end of the diff (a template file, or
+        # bytes that don't decode under the module's declared encoding).
+        # ValueError: source ast.parse rejects outright, e.g. embedded null bytes.
         return None
 
     removed_from_all, removed_undocumented = diff_surfaces(old_surface, new_surface)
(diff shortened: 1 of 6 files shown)


ducha-aiki (changes requested):
This is the right shape for the check and your validation reproduces exactly — I re-ran it rather than reading it. Against `v0.8.2` I get the same three `::error` lines (the 32-name `kornia.core` set, `DexiNed`, `upscale_double`) and exit 1; against `origin/main`, exit 0; `pytest tests/scripts/test_check_import_surface.py` gives 15 passed. Building it stdlib-only on `ast` + `git show` so it never imports kornia was the right call, and the `__getattr__` exemption is the correct way to reward the deprecation-shim pattern rather than punish it.

One blocking defect, one gap, and one policy question below. The blocker is a one-line fix.

## Blocking: the file list and the comparison use different base commits

`_changed_kornia_files` diffs `f"{base_ref}...HEAD"` — three dots, so the file list comes from the **merge base**. `check_file` then reads the old content with `_git_show(base_ref, path)` — no dots, so it reads the **tip of `main`**. When `main` has moved since the branch point, those are different commits, and the check reports names the PR never touched.

Reproduced end-to-end with your script, unmodified:

```text
# base: kornia/mod.py with __all__ = ["a"]
# PR branch: adds a docstring to mod.py, __all__ untouched
# main afterwards: __all__ = ["a", "b"], b defined

$ python3 .github/scripts/check_import_surface.py --base-ref main
::error file=kornia/mod.py::Removed from __all__: ['b']
$ echo $?
1

# same tree, --base-ref pointed at the merge-base commit instead
$ echo $?
0
```

A docs-only PR fails the check because someone else added a name to `__all__` in a file it happens to touch. This repo will hit it: branch protection is `strict: false`, so branches are routinely behind `main` — several PRs in the current queue are two weeks behind, and `kornia/geometry/conversions.py` alone has been touched by four merges this month. The workflow passing `--base-ref "origin/${{ github.base_ref }}"` is what makes it the live configuration; it is not a local-only artifact.

Fix is to resolve the merge base once and use it for both:

```python
    base = subprocess.run(  # noqa: S603
        ["git", "merge-base", args.base_ref, "HEAD"],  # noqa: S607
        capture_output=True, text=True, check=True,
    ).stdout.strip()
    files = _changed_kornia_files(base)
    reports = [r for r in (check_file(base, path) for path in files) if r is not None]
```

I applied exactly that and re-ran both cases: the false positive goes to exit 0, and a branch that genuinely drops `b` from `__all__` still reports `::error ... Removed from __all__: ['b']` and exit 1. `fetch-depth: 0` is already set, so `git merge-base` has what it needs. Please also add a test for it — the current `test_check_file_end_to_end_against_a_real_git_repo` builds a linear history, which is the one shape where the bug cannot appear.

## Gap: deleting a whole module passes silently

`check_file` returns `None` on `FileNotFoundError` with the comment "file was deleted; not this check's concern". But deleting a module is the largest possible version of the break this script exists to catch. Demonstrated on the same scratch repo — `git rm` a module whose `__all__` was `["a", "b"]`:

```text
$ python3 .github/scripts/check_import_surface.py --base-ref <merge-base>
$ echo $?
0
```

No output at all. The script's own docstring says either kind of removal "can silently break third-party code"; this is the loudest instance and it is the one case that produces nothing. I would report the module's entire old `__all__` at minimum. Whether a deletion should be fatal is a policy call you can leave to the maintainer, but silence is the wrong default.

## Policy question: the check cannot tell a broken `__all__` entry from a working one

`#4023` is the immediate case. Its three entries are in `__all__` and have no runtime binding at all — confirmed on current `main`:

```text
kornia.core  -> __all__ entries with no runtime attribute: ['eye_like', 'vec_like']
kornia.color -> __all__ entries with no runtime attribute: ['AUTUMN']
```

`from kornia.core import *` raises `AttributeError` on them today. The natural fix is to delete them from `__all__` — and that hard-fails this check, because neither module has a `__getattr__` and neither is under `kornia.contrib`. Requiring a deprecation shim for a name that has never resolved is not a sensible ask, so the follow-up PR you name in your own description is blocked by the check it ships with.

The `__getattr__` exemption is the right escape hatch for a name that *worked* and is being retired. It is the wrong one for a name that never worked. Worth adding something explicit and greppable — an allowlist file, or a `# import-surface: removed-entry-was-never-bound` marker next to the deleted line — so the waiver is reviewable rather than invisible. Your call on the shape; flagging it now because the repair-and-deprecation window will make intentional `__all__` removals routine, and a check with no reviewable override is a check that gets switched off.

## Parser scope — checked, nothing bites today

I ran an AST scan over all of `kornia/` for the constructs the walker does not handle, to see whether any are live:

| construct | handled? | occurrences in `kornia/` |
|---|---|---|
| `__all__ = a + b + c` (nested `BinOp`) | only one level — inner operand silently dropped | **0** |
| `__all__` from a non-literal (comprehension, name) | `has_all=True`, `all_names` empty | **0** |
| module-level `for` / `with` binding names | not recursed | **0** |
| module-level `__getattr__` | detected | **1** (`kornia/utils/__init__.py`, the shim you exempted) |

So all four are theoretical against today's tree. Two are worth a line of comment anyway, since a future edit makes them real and both fail toward a false positive rather than a miss.

One genuine blind spot that is live: `from .rgb import *` means the parser sees no module-scope binding for star-re-exported names, so the informational half of the check is blind for modules built that way — `kornia/color/__init__.py` has 84 `__all__` entries with no parsed binding for this reason. It causes no false positives (both revisions are equally blind, and `removed_from_all` reads the `__all__` literal directly), so this is coverage lost, not correctness. Worth a docstring sentence so the next reader does not mistake it for a bug.

## Minor

`::notice` volume: the `v0.8.2` run emits **280** of them. Per-PR diffs will be far smaller, but GitHub renders at most 10 notice annotations per step, so on a large refactor most of the inventory is log-only. Fine as designed — just do not count on the annotations being the delivery mechanism.

Requesting changes on the merge-base bug alone; the rest is discussion. Everything else I checked holds up, and I would like this in before the next batch of refactor PRs lands.

---

shlokkvaishnav:
Thanks for the thorough repro on the merge-base bug -- confirmed and fixed exactly as you diffed it.

## Merge-base mismatch (blocking) -- fixed

`main()` now resolves `git merge-base <base_ref> HEAD` once and passes that commit to both `_changed_kornia_files` and every `check_file` call, instead of `check_file` reading old content from the tip of `base_ref`. `fetch-depth: 0` was already set in the workflow, so no CI change needed.

Added `test_main_uses_merge_base_not_tip_of_base_ref`, built with the non-linear history shape you called out (base branch, a PR branch that never touches `__all__`, `main` moving forward afterward and adding a name to the same file). It fails against the pre-fix `main()` (exit 1, flags the untouched name) and passes after.

## Whole-module deletion -- fixed

`check_file` now diffs a deleted file against an empty `ModuleSurface()` instead of returning `None`, so it flows through the existing `__all__`/`fatal` logic for free: a deleted module with names in `__all__` is now fatal unless it's under `kornia.contrib`. Added `test_check_file_reports_and_fails_whole_module_deletion` and a contrib counterpart confirming that case still reports without failing.

## `#4023`-style entries (no reviewable override mechanism) -- deferred

Agreed this is a real gap and the `__getattr__`/`kornia.contrib` exemptions don't cover a name that never worked. Leaving it out of this PR on purpose: `#4023` is already the named follow-up for exactly this case, and the shape of the override (allowlist file vs. inline marker) deserves its own discussion rather than riding in here. Will pick it up when `#4023` is worked.

## Validated

```text
$ pytest tests/scripts/test_check_import_surface.py -v
...
18 passed

$ pre-commit run --files .github/scripts/check_import_surface.py tests/scripts/test_check_import_surface.py
...
ruff check ... Passed
ruff format ... Passed
codespell ... Passed
Add license header ... Passed
```

Also manually re-ran both of your repro scenarios (merge-base mismatch, whole-module deletion) against scratch repos to confirm both are fixed.

---
*Posted on behalf of @shlokkvaishnav by Claude (Sonnet 5).*

ducha-aiki:
Re-reviewed the delta `5f07452d..a547847f` rather than the whole PR again. **Both fixes are real and both new tests are genuine pins** — I re-measured everything below, nothing here is read off your description.

## Verified: merge-base fix

Reproduced my own scratch repro against your script unmodified (base → PR branch that only edits a docstring → `main` moves on and adds `b` to `__all__`):

```text
A: false positive gone?          exit=0
B: real removal still fatal?     ::error file=kornia/mod.py::Removed from __all__: ['a']   exit=1
```

Fail-on-base gate — your three new tests run against the **pre-fix** script (`5f07452d`, new test file copied in):

```text
FAILED test_main_uses_merge_base_not_tip_of_base_ref             - assert 1 == 0
FAILED test_check_file_reports_and_fails_whole_module_deletion   - assert None is not None
FAILED test_check_file_contrib_module_deletion_is_not_fatal      - assert None is not None
3 failed, 15 deselected
```

and `18 passed` at the head. All three are real pins, not stale-green. The non-linear history in `test_main_uses_merge_base_not_tip_of_base_ref` is exactly the shape needed.

One honest refinement to my own earlier severity claim: `actions/checkout` on a `pull_request` event checks out `refs/pull/N/merge`, so in the shipped CI configuration `merge-base(origin/main, HEAD)` is usually already the base parent and the old code mostly agreed with the new. The bug is still reachable in CI (whenever `main` advances between GitHub computing the merge ref and the job running) and is unconditionally reachable for local runs and any `ref:`-pinned checkout. The fix is right; it is just less load-bearing today than I implied.

## Blocking: git rename detection defeats the deletion fix

`_changed_kornia_files` runs `git diff --name-only <base>...HEAD`, which has **rename detection on by default**. When a module is moved, git reports only the destination path — the old path never reaches `check_file`, so the deletion branch you just added never runs.

```text
# kornia/oldname.py (__all__ = ['a','b'])  ->  git mv  ->  kornia/newname.py

$ git diff --name-only <merge-base>...HEAD -- kornia/
kornia/newname.py                              <- the deleted path is not here

$ python3 .github/scripts/check_import_surface.py --base-ref main
exit=0                                         <- silent
```

This is not a corner case; it is how modules actually get removed in this repo. **45 renames under `kornia/` since 2024-01-01**, and the head of that list is precisely the #3986/#4025 refactor class:

```text
kornia/utils/misc.py     -> kornia/core/ops.py
kornia/utils/helpers.py  -> kornia/core/utils.py
kornia/utils/grid.py     -> kornia/geometry/grid.py
kornia/utils/download.py -> kornia/onnx/download.py
```

Run against the real commit that did it (`94882c52`, "deprecate utils module across library" #3449), 9 files under `kornia/` were renamed and **`kornia/utils/download.py` carried `__all__ = ["CachedDownloader"]`**. Your current head says nothing about that file — the module `kornia.utils.download` and its one documented export vanished with no annotation. The check does fire on that commit, but via `kornia/utils/__init__.py`, which is a different file happening to catch it; a move that leaves the package `__init__` untouched produces nothing at all.

`--no-renames` alone is not the fix — it turns packageification (`x.py` → `x/__init__.py`, same module name, same `__all__`) into a false positive. The surface this check is about is the **module name**, not the file path. Both cases fall out of resolving the name:

```python
def _alternate_path(path: str) -> str:
    """The other file spelling of the same module name (`x.py` <-> `x/__init__.py`)."""
    stem = path[: -len(".py")]
    if stem.endswith("/__init__"):
        return stem[: -len("/__init__")] + ".py"
    return stem + "/__init__.py"
```

```python
    # in _changed_kornia_files
    ["git", "diff", "--name-only", "--no-renames", f"{base_ref}...HEAD", "--", "kornia/"],
```

```python
    # in check_file, replacing the bare `new_source = None`
    except FileNotFoundError:
        # The path is gone, but the *module* may not be: x.py <-> x/__init__.py
        # keeps the same importable name. Only a module name with no file left
        # anywhere is a real deletion.
        try:
            with open(_alternate_path(path), encoding="utf-8") as f:
                new_source = f.read()
        except (FileNotFoundError, NotADirectoryError, IsADirectoryError):
            new_source = None
```

I applied exactly that and re-measured all four behaviors plus your suite:

| scenario | head `a547847f` | patched |
|---|---|---|
| module rename `oldname.py` → `newname.py` | silent, exit 0 | `::error ... Removed from __all__: ['a','b']`, exit 1 |
| packageification `x.py` → `x/__init__.py`, `__all__` intact | exit 0 | exit 0 |
| plain `git rm` of a module | exit 1 | exit 1 |
| merge-base false positive | exit 0 | exit 0 |
| `tests/scripts/test_check_import_surface.py` | 18 passed | **18 passed** |
| PR branch vs `origin/main` | exit 0 | exit 0 |
| real commit `94882c52` | misses `kornia/utils/download.py` | `::error file=kornia/utils/download.py::Removed from __all__: ['CachedDownloader']` |

`ruff check` and `ruff format --diff` are clean on the patched file. Notice volume is unaffected: that commit's file list goes 115 → 124.

Please add a rename test alongside the two deletion ones — `git mv` a module with `__all__`, assert `main()` exits 1 — and a packageification test asserting it stays 0. Those two are the pins that keep this correct.

## Deferring `#4023` — agreed

Right call, and the reasoning is right: the shape of the override deserves its own discussion. Nothing further from me on that here.

## Minor, non-blocking

- `main()` calls `git merge-base` with `check=True`, so an unresolvable base ref exits with a `CalledProcessError` traceback rather than a message. Same pattern as `_changed_kornia_files`, so no regression — but a `try`/`except` printing `::error::could not resolve merge-base of <ref> and HEAD` would read better in a CI log.
- `assert "b" not in captured.out` in the merge-base test is a single-character substring check. It happens to be sound because the output is empty on success, but `assert captured.out == ""` says what you mean.

Everything else from the last round still holds. Requesting changes on the rename item only; with that in, this is ready.

---

shlokkvaishnav:
Confirmed the rename-detection gap, and it's fixed as suggested.

## Rename detection defeating the deletion fix -- fixed

`_changed_kornia_files` now runs `git diff --no-renames`, so a `git mv` reports both the old and new paths instead of only the destination. `check_file`'s `FileNotFoundError` handling now tries the one legitimate same-name rename shape (`x.py` <-> `x/__init__.py`) before falling through to "module genuinely gone" -- so packageification stays a no-op and an actual module-removal-via-rename now hits the same `__all__`/fatal path as a plain `git rm`.

Added the two tests you asked for:
- `test_main_flags_module_removed_via_rename` -- `git mv` a module with `__all__` away, assert `main()` exits 1. Fails on the prior commit (`exit 0`, silent), passes now.
- `test_main_ignores_packageification_that_keeps_all_intact` -- `x.py` -> `x/__init__.py`, `__all__` untouched, assert `main()` stays exit 0 (was already 0 before this fix, since it's not the bug being fixed, but it's the pin that keeps the `--no-renames` change honest going forward).

## Minor items -- both taken

- `main()`'s merge-base resolution now checks the return code itself and prints `::error::could not resolve merge-base of <ref> and HEAD>` instead of letting `CalledProcessError` raise past it.
- The merge-base test now asserts `captured.out == ""` instead of the `"b" not in captured.out` substring check.

## Validated

```text
$ pytest tests/scripts/test_check_import_surface.py -v
...
20 passed

$ pre-commit run --files .github/scripts/check_import_surface.py tests/scripts/test_check_import_surface.py
...
ruff check ... Passed
ruff format ... Passed
codespell ... Passed
Add license header ... Passed
```

Also manually reran your `git mv` and packageification repros against scratch repos to confirm both land where your table says.

---
*Posted on behalf of @shlokkvaishnav by Claude (Sonnet 5).*

ducha-aiki:
**APPROVE** — both round-3 fixes verified, and both new tests discriminate.

Setup: worktree at the PR head `f081a2cb`, with `.github/scripts/check_import_surface.py` swapped to the previous commit `a547847f` for the fail-on-base pass; `~/dev/kornia/.venv/bin/python`.

**1. The rename fix is pinned.** With the script at `a547847f` and the new tests kept:

```
FAILED test_main_flags_module_removed_via_rename - assert 0 == 1
1 failed, 19 deselected
```

Exit 0 where it must be 1 — the silent-pass the review described.

**2. The packageification tolerance is earned, not decorative.** I neutralised only the `_alternate_path` fallback in `check_file` (leaving `--no-renames` in place) and re-ran:

```
FAILED test_main_ignores_packageification_that_keeps_all_intact - assert 1 == 0
1 failed, 19 passed
```

So `--no-renames` alone would have traded the rename blind spot for a packageification false positive, and the fallback is exactly what stops it. That is the pair of tests I hoped for.

Full suite at the head: **20 passed in 1.78s**.

Unasked-for and welcome: `main()` now handles a failing `git merge-base` with `check=False` plus an `::error::` line instead of letting `CalledProcessError` escape as a traceback. That is the right shape for a CI script.

ducha-aiki (approved):

ducha-aiki:
Thank you!

Issue #4033, opened 2026-08-26
CI: add a half-precision (float16/bfloat16) and MPS test job


ducha-aiki:
## Description

Half precision (`float16`/`bfloat16`) and MPS are both documented as supported test surfaces (`AGENTS.md`, `TESTING.md` — `pixi run test-half`, `--device=mps`, `--dtype=float16,bfloat16`), but neither has a CI job running them regularly. As a result, regressions on these surfaces go undetected until someone runs the focused checks manually.

Concretely, on kornia#4006 half precision and MPS were each turned red by intermediate commits along the way, with zero CI signal catching it at the time it happened — the breakage was only found by manually running the focused checks described in `AGENTS.md`.

## Proposal

An upcoming PR adds `pixi run verify-delta`, which set-diffs failing tests against `origin/main` across:

- cpu float32
- cpu float16 / bfloat16 / float64
- mps float32
- inductor (`KORNIA_TEST_OPTIMIZER=inductor`)

Add a CI workflow (nightly, or gated behind a label like `test-precision`/`test-mps` on a PR) that runs these surfaces and posts a summary table, using a macOS runner for the MPS leg.

## Why diff, not gate-on-zero

There are known pre-existing failures on `main` for these surfaces today:

- ~697 half-precision/float64 failures in `tests/geometry`
- ~25 MPS failures

A job that hard-gates on zero failures would be red from day one and get ignored/skipped. The job needs to diff the failing-test set against `origin/main` (which is what `verify-delta` already does) and flag only newly-introduced failures, not the existing baseline.

## Environment

```
kornia 0.9.0rc1 | torch 2.9.1 | python 3.11.14 | OS macOS-26.5.1-arm64-arm-64bit | CUDA/ROCm None (available=False, GPU=none) | MPS=True
```

---
Posted on behalf of @ducha-aiki by Claude (Fable 5).

Issue #4031, opened 2026-08-26
warp_affine: empty-dsize path raises RuntimeError where the full path raises NotImplementedError for an integral src


ducha-aiki:
## Description

`warp_affine`'s empty-`dsize` fast path raises a different exception **type** than its full path for the same invalid input: a non-floating-point (`int64`) `src`.

- Full path (`dsize` fully non-zero) falls through to `F.grid_sample`, which raises `NotImplementedError`.
- Empty path (`dsize` has a zero dimension) routes through `_empty_warp_output_2d`, which has its own explicit check that raises `RuntimeError` first.

A degenerate/empty-output path should reject exactly what the full path rejects — same exception type for the same misuse — so callers that branch on exception type (or tests that assert parity between the two paths) see consistent behavior regardless of `dsize`.

## Where

`kornia/geometry/transform/imgwarp.py`, in `_empty_warp_output_2d`:

```python
# An integral ``src`` fails inside ``grid_sample`` itself on the non-empty path, so it must
# fail here too — but name ``src`` rather than blaming the other operand for it below. (MPS is the
# exception: its ``grid_sample`` samples an integral image back into ``int64`` instead of
# rejecting it. Matching that would mean bilinear-sampling into an integer output, so the
# cpu/cuda contract is the one enforced here.)
if not src.is_floating_point():
    raise RuntimeError(f"Expected a floating point src, got {src.dtype}.")
```

The comment states the intent explicitly ("it must fail here too") but the exception type doesn't match what the full path actually raises.

## Minimal repro

```python
import torch
from kornia.geometry.transform import warp_affine

src = torch.zeros(1, 1, 4, 4, dtype=torch.int64)
M = torch.eye(2, 3)[None]

warp_affine(src, M, (4, 4))  # full path
warp_affine(src, M, (0, 4))  # empty path
```

Output:

```
=== full path, dsize=(4,4), int64 src ===
NotImplementedError: "grid_sampler_2d_cpu_kernel_impl" not implemented for 'Long'

=== empty path, dsize=(0,4), int64 src ===
RuntimeError: Expected a floating point src, got torch.int64.
```

## Why it matters

Found by a new `testing.assert_degenerate_path_parity` helper (in an upcoming PR) that runs the same set of invalid inputs through a function's full and degenerate code paths and asserts they raise the same exception type. This case is the one input that currently breaks parity for `warp_affine` — the helper's own tests exclude it with an inline comment pending this fix.

## Environment

```
[...]

PR #4034, opened 2026-08-26
test(testing): precision and degenerate-path helpers, pixi run verify-delta

ducha-aiki:
## Why

Classifying every review finding on kornia#4006 by origin (not by reviewer) showed the fix loop
itself was the defect generator: of ~24 actionable findings after review round 2, ~20 were
introduced by an earlier fix commit on the same PR, not present in the original diff. The same
half-precision rounding bug — a size cast into fp16/bf16 before a division, under
`torch.jit.trace`/`torch.compile` — was "fixed" three times across waves 5, 8, and 9, each one
layer deeper, while two regression tests (sizes 257 and 3001) passed vacuously even though both
authors knew the rounding rule. Half-precision and MPS have no CI job, so nothing catches this
class before a human reviewer does. This PR puts the knowledge into code: three test helpers that
make the vacuous case and the eager/capture divergence structurally hard to write, plus a tool
that gates a push against `origin/main` on exactly the surfaces CI does not cover.

## What

Three helpers in `testing/precision.py`, exported from `testing/__init__.py` alongside
`BaseTester`:

- `unrepresentable_sizes(dtype, *, lo=2, hi=4096) -> list[int]` — sizes `n` in `[lo, hi]` where `n`
  or `n - 1` is not exact in `dtype`; empty for float32/float64.
- `assert_capture_matches_eager(fn, make_inputs, *, sizes, device, dtype, capture)` — asserts
  `torch.equal` (byte-equality, not `allclose`) between eager and `torch.jit.trace` /
  `torch.compile(fullgraph=True, dynamic=True)` outputs, per size; reports the first disagreeing
  size and max abs diff; skips cleanly when the backend lacks the capture mode (MPS + inductor).
- `assert_degenerate_path_parity(fn, full_kwargs, degenerate_kwargs, bad_inputs)` — asserts an
  empty/singleton code path raises exactly what the full path raises, for the same invalid input.

Historical negatives in `tests/testing/_historical.py` vendor the exact `torch.jit.is_tracing()`
bodies from three kornia#4006 commits as small local functions, so `assert_capture_matches_eager`
has a deterministic proof it would have caught each wave:

- `create_meshgrid_9ed891c5` (wave 5) — divisor built directly in the coordinate dtype.
- `create_meshgrid_32ab0eeb` (wave 8) — float32 arithmetic, but the divisor is cast down first.
- `normal_transform_pixel_1522441d` (wave 9) — promotion keyed off the `dtype` *argument*, so
  `dtype=None` under a half default dtype still rounds the size.

`pixi run verify-delta` (`testing/verify_delta.py`, run as `python -m testing.verify_delta`) diffs
**failing-test sets**, never counts, between the branch and `origin/main` on four surfaces: `cpu
float32`, `cpu float16,bfloat16,float64`, `mps float32` (skipped if unavailable), and `inductor cpu
float32` (`-k "dynamo or compile"`). Exit codes: `0` = no new failures, `1` = some surface has a
`new` id, `2` = nothing was verified at all (no surface ran pytest on the branch side — a typo'd
`--tests` path or an unavailable `--only` surface), which is a failure state, not a pass. A row
with no baseline on `origin/main` (a path that only exists on the branch) prints its new-failure
count as `N*` with a legend line, rather than silently using an empty baseline. Test-directory
selection derives from `git diff origin/main...HEAD --name-only` by default; pass `--tests
tests/<module>` / `--only "<surface>"` **after `--`** to bound it (pixi forwards `--` literally and
its task shell re-splits the string, so `parse_args` strips a leading `--` and `--only` tokens are
rejoined longest-match-first against the four surface names). It maintains two sibling directories
next to the repo so reruns are fast and nothing pollutes the source tree: a detached worktree of
`origin/main` at `../.<repo>-verify-main`, and junit/summary output at
`../.<repo>-verify-delta` (both overridable, `--main-worktree` and `--out`). Before trusting either
tree's result, `_assert_imports_from` asserts `import kornia` resolves inside the tree actually
under test — the `.venv` here is shared/symlinked across worktrees, and `uv run` re-points the
editable install to whichever tree ran last, so without this guard a stale import would silently
verify the wrong code.

## Design notes

- **`testing/precision.py` imports `pytest`.** So does `testing/base.py` already — `pytest.skip`
  is how `assert_capture_matches_eager` cleanly skips the compile-capture sweep on MPS. This
  follows the existing convention rather than introducing a parallel "no pytest in `testing/`"
  rule.
- **The tool lives in `testing/`, not a new top-level `scripts/`.** It ships as
  `testing/verify_delta.py`, invoked via `python -m testing.verify_delta`, so its pure functions
  (`changed_test_dirs`, `failing_ids`, `diff_failures`, `render_table`) are unit-tested the same
  way as the rest of `testing/` — from `tests/testing/`, with subprocess and git calls behind a
  recording fake, no real pytest/git process launched by the test suite itself.
- **Cost of touching `testing/`, `pyproject.toml`, or `pixi.toml`.** `changed_test_dirs` treats
  changes under `testing/`, `tests/conftest.py`, `conftest.py`, `pyproject.toml`, or `pixi.toml` as
  affecting everything and widens the run to the whole `tests/` tree instead of a narrow
  `tests/<module>` mapping. That is deliberate — these files are load-bearing for every test — but
  it means the no-argument invocation on a PR like this one costs roughly 20 minutes per surface,
  up to an hour or more across all four; use `--tests`/`--only` to scope while iterating.
- **`changed_test_dirs` assumes test locality.** The `kornia/<mod>` → `tests/<mod>` mapping is
  right for the common case, but a `kornia/geometry` change routinely breaks `tests/augmentation`,
  so the gate's promise ("zero `new` failures") is narrower than it sounds under the default
  scoping. For a cross-cutting change, pass `--tests tests` and take the full cost.

## Verification

Six commands from the pre-`verify_delta` snapshot (`tests/testing` only, run from
`/Users/oldufo/dev/kornia-review-skills`):

| # | Command | Result |
|---|---|---|
| 1 | `uv run pre-commit run --all-files` | all 13 hooks passed |
| 2 | `uv run pytest tests/testing -q --dtype=float32,float64,float16,bfloat16` | 21 passed (22 collected, 1 compile test deselected — `KORNIA_TEST_OPTIMIZER` unset) |
| 3 | `KORNIA_TEST_OPTIMIZER=inductor uv run pytest tests/testing -q -k "compile or dynamo"` | 1 passed, 18 deselected |
| 4 | `uv run pytest tests/testing -q --device=mps --dtype=float32` | 17 passed, 1 skipped (documented: `warp_affine`'s empty-dsize path crashes `F.grid_sample` on MPS) |
| 5 | `uv run pytest --doctest-modules testing/precision.py -q` | 1 passed |
| 6 | `uvx ty@0.0.9 check kornia testing` | all checks passed |

Final state, after `verify_delta.py` and its tests landed:

```
$ uv run pytest tests/testing -q --dtype=float32,bfloat16,float16
collected 64 items
63 passed in 0.68s
```

(64th is the `compile` test, deselected because `KORNIA_TEST_OPTIMIZER` is unset — rule 7 of
`kornia-precision-testing` working as documented.)

`pixi run verify-delta` run on this branch against itself (the tool verifying its own change):

| run | scope | wall time | new / fixed / unchanged (all 4 rows) |
|---|---|---|---|
| `-- --only "cpu float32"` (`efb4c4b6` vs `9e45dddc`) | full suite (`testing/` widens) | 19:55.68 | cpu float32 `0/0/0`; other three surfaces `skipped` |
| `--tests tests/testing tests/geometry` (`efb4c4b6` vs `9e45dddc`) | bounded, all 4 surfaces | 3:27.42 | cpu float32 `0/0/0`; cpu half/f64 `0/0/697`; mps float32 `0/0/25`; inductor `0/0/0` |
| `-- --tests tests/testing tests/geometry` (`032a38f5` vs `9e45dddc`, final PR-B-adjacent state) | bounded, all 4 surfaces | 3m31s (211s) | cpu float32 `0/0/0`; cpu half/f64 `0/0/697`; mps float32 `0/0/25`; inductor `0/0/0` |

The 697 half-precision and 25 MPS failures are `unchanged` — present identically on
`origin/main` — which is the argument for the tool: a count-based or "did the suite go green"
check would have reported this branch as breaking 722 tests it never touched. `new = 0` on every
row in both bounded runs; exit 0. (A pre-fix version of the runner, exercised the same way, exited
1 with `new = 723` on the same invocation — a false positive from pytest aborting collection
entirely when `tests/testing` does not exist on `origin/main` — fixed before this state.)

## Found while building

Two real, reportable gaps in kornia surfaced the first time the parity helper was pointed at
`warp_affine`'s degenerate path — filed rather than fixed here, since this PR only ships test
infrastructure:

- kornia#4031 — `warp_affine`'s empty-dsize path does not reject a non-floating `M` the way the
  full path does (an integral `src` should raise, and doesn't).
- kornia#4032 — `warp_affine`'s empty-dsize path crashes on MPS (`F.grid_sample` asserts on a
  zero-sized grid) rather than returning an empty tensor.

kornia#4033 — follow-up: add a nightly CI job covering half-precision and MPS, using
`verify_delta`'s four surfaces. Explicitly out of scope here per the design's non-goals; this PR
only adds the manual gate.

## Relates to

Relates to #4006 (the wave analysis that motivated this PR; not closed by it — #4006 itself was
merged separately).

---

## Evaluation

The helpers and `verify-delta` were used by the with-skill arm of a paired agent evaluation (three historical #4006 review rounds, Opus 5 and Sonnet 5, with and without the skills from #4035). Relevant to this PR: `assert_capture_matches_eager` + `unrepresentable_sizes` distinguished a correct fix from the historical one in every graded run (the without-skill Sonnet run wrote the wave-8 divisor cast verbatim and its own 257-only test passed on it; the sweep test fails 4 ways on that tree), and `verify-delta` ran end to end from a pre-fix worktree in 5 runs (exit 0/1/2 semantics and the `N*` row all exercised). Full tables and the confound handling are in #4035's description.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * `testing.precision`: `unrepresentable_sizes`, `assert_capture_matches_eager`, `assert_degenerate_path_parity`
-  for reduced-precision, graph-capture and degenerate-path tests.
+  for reduced-precision, graph-capture and degenerate-path tests, and `pixi run verify-delta`, which diffs the
+  failing-test *sets* of a branch and its base revision across the cpu, half, MPS and inductor surfaces.
 
 ### Breaking changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `testing.precision`: `unrepresentable_sizes`, `assert_capture_matches_eager`, `assert_degenerate_path_parity`
   for reduced-precision, graph-capture and degenerate-path tests, and `pixi run verify-delta`, which diffs the
   failing-test *sets* of a branch and its base revision across the cpu, half, MPS and inductor surfaces. (#4034)
-  `verify-delta` refuses to run on a dirty checkout, since its scope is `base...HEAD` and its verdict must
-  describe HEAD, and it exits non-zero whenever a selected, available surface could not be measured.
+  `verify-delta` refuses to run on a dirty checkout by default, since its automatic scope is `base...HEAD`;
+  `--allow-dirty` requires explicit test paths. It exits non-zero whenever a selected, available surface could
+  not be measured.
 
 ### Breaking changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `testing.precision`: `unrepresentable_sizes`, `assert_capture_matches_eager`, `assert_degenerate_path_parity`
   for reduced-precision, graph-capture and degenerate-path tests, and `pixi run verify-delta`, which diffs the
-  failing-test *sets* of a branch and its base revision across the cpu, half, MPS and inductor surfaces.
+  failing-test *sets* of a branch and its base revision across the cpu, half, MPS and inductor surfaces. (#4034)
 
 ### Breaking changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `testing.precision`: `unrepresentable_sizes`, `assert_capture_matches_eager`, `assert_degenerate_path_parity`
   for reduced-precision, graph-capture and degenerate-path tests, and `pixi run verify-delta`, which diffs the
   failing-test *sets* of a branch and its base revision across the cpu, half, MPS and inductor surfaces. (#4034)
+  `verify-delta` refuses to run on a dirty checkout, since its scope is `base...HEAD` and its verdict must
+  describe HEAD, and it exits non-zero whenever a selected, available surface could not be measured.
 
 ### Breaking changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+* `testing.precision`: `unrepresentable_sizes`, `assert_capture_matches_eager`, `assert_degenerate_path_parity`
+  for reduced-precision, graph-capture and degenerate-path tests.
+
 ### Breaking changes
 
 * The shape guards of `normalize_homography` and `denormalize_homography` now reject everything but a
--- a/TESTING.md
+++ b/TESTING.md
@@ -290,7 +290,8 @@ from testing import assert_capture_matches_eager, unrepresentable_sizes
 
 def test_trace_matches_eager(device, dtype):
     def grid(image):
-        return kornia.geometry.create_meshgrid(image.shape[-2], image.shape[-1], device=image.device, dtype=image.dtype)
+        h, w = image.shape[-2], image.shape[-1]
+        return kornia.geometry.create_meshgrid(h, w, device=image.device, dtype=image.dtype)
 
     assert_capture_matches_eager(
         grid,
@@ -307,6 +308,8 @@ Rules the helpers enforce, for code that is written by hand:
 - Resolve `dtype=None` to `torch.get_default_dtype()` *before* deciding whether to promote.
 - Guard a cast-back on `is_floating_point()` — an integral grid dtype must stay promoted.
 - Compile tests carry `dynamo` or `compile` in their name (they are deselected otherwise).
+
+---
 
 ## Writing Robust Tests
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -268,6 +268,46 @@ The `padding_mode="border"` issue in `F.grid_sample` (2D) is worked around in th
 
 ---
 
+## Precision and Degenerate-Path Helpers
+
+`testing.precision` carries three assertions for the bug classes that recur in dtype/capture work
+(see kornia#4006 for the history). Use them instead of hand-picking a "boundary" size.
+
+- `unrepresentable_sizes(dtype)` — sizes at which `n` or `n - 1` is inexact in `dtype`. Sweep a
+  slice of this list; never pick one size. 257 is vacuous against a rounded divisor and decisive
+  against a rounded size; 258 is the reverse.
+- `assert_capture_matches_eager(fn, make_inputs, sizes=..., device=..., dtype=..., capture="trace"|"compile")`
+  — byte-equality (`torch.equal`) between eager and `torch.jit.trace` / `torch.compile(fullgraph=True,
+  dynamic=True)`, size by size. Derive sizes from tensor shapes inside `fn`. `sizes` must not be
+  empty: `float32`/`float64` return `[]` from `unrepresentable_sizes`, so always include the
+  degenerate sizes 1 and 2, e.g. `sizes=[1, 2, *unrepresentable_sizes(dtype)[:8]]`.
+- `assert_degenerate_path_parity(fn, full_kwargs, degenerate_kwargs, bad_inputs)` — an empty or
+  singleton path must raise exactly what the full path raises for the same invalid input.
+
+```python
+from testing import assert_capture_matches_eager, unrepresentable_sizes
+
+
+def test_trace_matches_eager(device, dtype):
+    def grid(image):
+        return kornia.geometry.create_meshgrid(image.shape[-2], image.shape[-1], device=image.device, dtype=image.dtype)
+
+    assert_capture_matches_eager(
+        grid,
+        lambda size, device, dtype: (torch.zeros(1, 1, size, 4, device=device, dtype=dtype),),
+        sizes=[1, 2, *unrepresentable_sizes(dtype)[:8]],
+        device=device,
+        dtype=dtype,
+    )
+```
+
+Rules the helpers enforce, for code that is written by hand:
+
+- Under capture, divide by the *unrounded* size and cast the quotient; never cast the divisor.
+- Resolve `dtype=None` to `torch.get_default_dtype()` *before* deciding whether to promote.
+- Guard a cast-back on `is_floating_point()` — an integral grid dtype must stay promoted.
+- Compile tests carry `dynamo` or `compile` in their name (they are deselected otherwise).
+
 ## Writing Robust Tests
 
 - **Seed the RNG** when the test compares against reference values: `torch.manual_seed(seed)`.
--- a/TESTING.md
+++ b/TESTING.md
@@ -276,8 +276,8 @@ The `padding_mode="border"` issue in `F.grid_sample` (2D) is worked around in th
 - `unrepresentable_sizes(dtype)` — sizes at which `n` or `n - 1` is inexact in `dtype`. Sweep a
   slice of this list; never pick one size. 257 is vacuous against a rounded divisor and decisive
   against a rounded size; 258 is the reverse. `hi` above `torch.finfo(dtype).max` raises
-  `ValueError`: every candidate there casts to `inf` and round-trips back unchanged, so the sweep
-  would be sizes no test can allocate.
+  `ValueError`: candidates there cast to non-finite values and conversion back to `int64` produces
+  a sentinel rather than the original size, corrupting the sweep.
 - `assert_capture_matches_eager(fn, make_inputs, sizes=..., device=..., dtype=..., capture="trace"|"compile")`
   — byte-equality between eager and `torch.jit.trace` / `torch.compile(fullgraph=True,
   dynamic=True)`, size by size. The comparison is on the raw bytes, not `torch.equal`, which is
@@ -317,6 +317,21 @@ Rules the helpers enforce, for code that is written by hand:
 - Resolve `dtype=None` to `torch.get_default_dtype()` *before* deciding whether to promote.
 - Guard a cast-back on `is_floating_point()` — an integral grid dtype must stay promoted.
 - Compile tests carry `dynamo` or `compile` in their name (they are deselected otherwise).
+
+### Comparing branch and base failures
+
+`pixi run verify-delta` compares failing-test sets between the branch and `--base` across the CPU,
+reduced-precision, available MPS, and inductor surfaces. Its automatic test selection is deliberately
+local: a change under `kornia/geometry` maps to `tests/geometry`. That cannot discover every
+cross-package consumer (for example, augmentation code that calls geometry), so pass `--tests tests`
+for shared primitives or any cross-cutting change.
+
+The default refuses a dirty checkout because it describes committed `HEAD`. To intentionally test
+working-tree changes, pass both `--allow-dirty` and explicit `--tests`, listing every affected target;
+the tool cannot infer whether a caller-chosen scope omits a tracked or untracked change outside
+`base...HEAD`. Every selected and available surface must produce a measurement. A machine without MPS
+records that row as unavailable, but a run that selects only unavailable rows still exits non-zero
+because it verified nothing.
 
 ---
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -275,14 +275,20 @@ The `padding_mode="border"` issue in `F.grid_sample` (2D) is worked around in th
 
 - `unrepresentable_sizes(dtype)` — sizes at which `n` or `n - 1` is inexact in `dtype`. Sweep a
   slice of this list; never pick one size. 257 is vacuous against a rounded divisor and decisive
-  against a rounded size; 258 is the reverse.
+  against a rounded size; 258 is the reverse. `hi` above `torch.finfo(dtype).max` raises
+  `ValueError`: every candidate there casts to `inf` and round-trips back unchanged, so the sweep
+  would be sizes no test can allocate.
 - `assert_capture_matches_eager(fn, make_inputs, sizes=..., device=..., dtype=..., capture="trace"|"compile")`
   — byte-equality (`torch.equal`) between eager and `torch.jit.trace` / `torch.compile(fullgraph=True,
   dynamic=True)`, size by size. Derive sizes from tensor shapes inside `fn`. `sizes` must not be
   empty: `float32`/`float64` return `[]` from `unrepresentable_sizes`, so always include the
   degenerate sizes 1 and 2, e.g. `sizes=[1, 2, *unrepresentable_sizes(dtype)[:8]]`.
 - `assert_degenerate_path_parity(fn, full_kwargs, degenerate_kwargs, bad_inputs)` — an empty or
-  singleton path must raise exactly what the full path raises for the same invalid input.
+  singleton path must raise exactly what the full path raises for the same invalid input. Every
+  `bad_inputs` name must already be a key of both kwargs dicts (an unknown name would be *added*
+  rather than substituted, and both paths would raise `TypeError` over a call never made); parity
+  is compared on the exception type, so check the messages when two paths could raise the same type
+  for unrelated reasons.
 
 ```python
 from testing import assert_capture_matches_eager, unrepresentable_sizes
--- a/TESTING.md
+++ b/TESTING.md
@@ -279,14 +279,17 @@ The `padding_mode="border"` issue in `F.grid_sample` (2D) is worked around in th
   `ValueError`: every candidate there casts to `inf` and round-trips back unchanged, so the sweep
   would be sizes no test can allocate.
 - `assert_capture_matches_eager(fn, make_inputs, sizes=..., device=..., dtype=..., capture="trace"|"compile")`
-  — byte-equality (`torch.equal`) between eager and `torch.jit.trace` / `torch.compile(fullgraph=True,
-  dynamic=True)`, size by size. Derive sizes from tensor shapes inside `fn`. `sizes` must not be
+  — byte-equality between eager and `torch.jit.trace` / `torch.compile(fullgraph=True,
+  dynamic=True)`, size by size. The comparison is on the raw bytes, not `torch.equal`, which is
+  numeric: it calls two NaNs unequal (a spurious failure) and `+0.0` equal to `-0.0` (a missed
+  sign-bit change). Derive sizes from tensor shapes inside `fn`. `sizes` must not be
   empty: `float32`/`float64` return `[]` from `unrepresentable_sizes`, so always include the
   degenerate sizes 1 and 2, e.g. `sizes=[1, 2, *unrepresentable_sizes(dtype)[:8]]`.
 - `assert_degenerate_path_parity(fn, full_kwargs, degenerate_kwargs, bad_inputs)` — an empty or
   singleton path must raise exactly what the full path raises for the same invalid input. Every
   `bad_inputs` name must already be a key of both kwargs dicts (an unknown name would be *added*
-  rather than substituted, and both paths would raise `TypeError` over a call never made); parity
+  rather than substituted, and both paths would raise `TypeError` over a call never made), and
+  `bad_inputs` must not be empty (with no pairs only the two baseline calls run); parity
   is compared on the exception type, so check the messages when two paths could raise the same type
   for unrelated reasons.
 
--- a/pixi.toml
+++ b/pixi.toml
@@ -44,6 +44,7 @@ test-f64 = { cmd = "uv run pytest -v tests/", env = { KORNIA_TEST_DTYPE = "float
 test-slow = { cmd = "uv run pytest -v tests/", env = { KORNIA_TEST_RUNSLOW = "true" } }
 test-quick = { cmd = "uv run pytest -v -m 'not (jit or grad or nn)' tests/" }
 test-half = { cmd = "uv run pytest -v tests/", env = { KORNIA_TEST_DTYPE = "float16,bfloat16" }, description = "Run CPU half-precision tests" }
+verify-delta = { cmd = "uv run python -m testing.verify_delta", description = "Diff failing-test sets vs origin/main on cpu/half/mps/inductor (usage: pixi run verify-delta [-- --only 'cpu float32'])" }
 test-module = { cmd = "uv run pytest -v", description = "Run tests for a specific module (usage: pixi run test-module tests/<module_path>, e.g., pixi run test-module tests/contrib/)" }
 # Quality
 lint = { cmd = "uv run pre-commit run ruff-check --all-files && uv run pre-commit run ruff-format --all-files" }
--- a/pixi.toml
+++ b/pixi.toml
@@ -51,7 +51,7 @@ lint = { cmd = "uv run pre-commit run ruff-check --all-files && uv run pre-commi
 pre-commit-all = { cmd = "uv run pre-commit run --all-files", description = "Run all repository pre-commit checks" }
 pre-commit-install = { cmd = "uv run pre-commit install", description = "Install the pre-commit Git hook" }
 typecheck = { cmd = "ty check kornia" }
-doctest = { cmd = "uv run pytest -v --doctest-modules kornia/" }
+doctest = { cmd = "uv run pytest -v --doctest-modules kornia/ testing/precision.py" }
 toml-fmt = { cmd = "tombi format", description = "Format TOML files" }
 toml-fmt-check = { cmd = "tombi format --check", description = "Check TOML formatting" }
 # Docs
--- a/testing/__init__.py
+++ b/testing/__init__.py
@@ -16,5 +16,6 @@
 #
 
 from .parametrized_tester import parametrized_test
+from .precision import unrepresentable_sizes
 
-__all__ = ["parametrized_test"]
+__all__ = ["parametrized_test", "unrepresentable_sizes"]
--- a/testing/__init__.py
+++ b/testing/__init__.py
@@ -16,6 +16,6 @@
 #
 
 from .parametrized_tester import parametrized_test
-from .precision import unrepresentable_sizes
+from .precision import assert_capture_matches_eager, unrepresentable_sizes
 
-__all__ = ["parametrized_test", "unrepresentable_sizes"]
+__all__ = ["assert_capture_matches_eager", "parametrized_test", "unrepresentable_sizes"]
--- a/testing/__init__.py
+++ b/testing/__init__.py
@@ -16,6 +16,11 @@
 #
 
 from .parametrized_tester import parametrized_test
-from .precision import assert_capture_matches_eager, unrepresentable_sizes
+from .precision import assert_capture_matches_eager, assert_degenerate_path_parity, unrepresentable_sizes
 
-__all__ = ["assert_capture_matches_eager", "parametrized_test", "unrepresentable_sizes"]
+__all__ = [
+    "assert_capture_matches_eager",
+    "assert_degenerate_path_parity",
+    "parametrized_test",
+    "unrepresentable_sizes",
+]
--- a/testing/precision.py
+++ b/testing/precision.py
@@ -26,7 +26,8 @@ empty/singleton code path that validated less (or more) than the non-empty path.
 from __future__ import annotations
 
 import warnings
-from typing import Any, Callable, Literal, Sequence
+from contextlib import contextmanager
+from typing import Any, Callable, Iterator, Literal, Sequence
 
 import pytest
 import torch
@@ -45,12 +46,16 @@ def unrepresentable_sizes(dtype: torch.dtype, *, lo: int = 2, hi: int = 4096) ->
     Args:
         dtype: a floating dtype. Integer dtypes raise ``TypeError``.
         lo: smallest size to consider (inclusive).
-        hi: largest size to consider (inclusive). Keep it below ``torch.finfo(dtype).max`` (65504 for
-            float16): above that every candidate casts to ``inf`` and the round-trip back to
+        hi: largest size to consider (inclusive). Must not exceed ``torch.finfo(dtype).max`` (65504
+            for float16): above that every candidate casts to ``inf`` and the round-trip back to
             ``int64`` is meaningless, so the result would be garbage rather than a size sweep.
 
     Returns:
         sorted sizes, empty when every integer in range is exact (float32/float64 below 2**24).
+
+    Raises:
+        TypeError: when ``dtype`` is not a floating dtype.
+        ValueError: when ``hi`` exceeds ``torch.finfo(dtype).max``.
 
     Example:
         >>> unrepresentable_sizes(torch.bfloat16)[:3]
@@ -60,10 +65,37 @@ def unrepresentable_sizes(dtype: torch.dtype, *, lo: int = 2, hi: int = 4096) ->
     """
     if not dtype.is_floating_point:
         raise TypeError(f"unrepresentable_sizes needs a floating dtype, got {dtype}")
+    finite_max = torch.finfo(dtype).max
+    if hi > finite_max:
+        raise ValueError(
+            f"hi={hi} is above torch.finfo({dtype}).max={finite_max:g}: every candidate there casts to inf "
+            "and round-trips back to the same int64, so the sweep would be garbage rather than sizes"
+        )
     n = torch.arange(lo, hi + 1, dtype=torch.int64)
     inexact_n = n.to(dtype).to(torch.int64) != n
     inexact_prev = (n - 1).to(dtype).to(torch.int64) != n - 1
     return n[inexact_n | inexact_prev].tolist()
+
+
+@contextmanager
+def _restoring_rng(device: torch.device) -> Iterator[None]:
+    """Restore the RNG state on exit, so the next call inside starts where this one did.
+
+    Tracing runs ``fn`` once itself, and the eager reference call runs it again; without this a
+    function that consumes randomness would be compared against a different draw and fail with the
+    rounding message below, which is not what went wrong.
+    """
+    cpu_state = torch.get_rng_state()
+    module = getattr(torch, device.type, None) if device.type != "cpu" else None
+    # ``torch.cuda``/``torch.mps``/``torch.xpu`` all take the device explicitly; the no-argument
+    # form would read whichever device is current, which need not be the one under test.
+    device_state = module.get_rng_state(device) if hasattr(module, "get_rng_state") else None
+    try:
+        yield
+    finally:
+        torch.set_rng_state(cpu_state)
+        if device_state is not None:
+            module.set_rng_state(device_state, device)
 
 
 def _as_tuple(out: Any) -> tuple[torch.Tensor, ...]:
@@ -129,11 +161,13 @@ def assert_capture_matches_eager(
     for size in sizes:
         inputs = make_inputs(size, torch.device(device), dtype)
         if capture == "trace":
-            with warnings.catch_warnings():
+            with warnings.catch_warnings(), _restoring_rng(torch.device(device)):
                 warnings.simplefilter("ignore", torch.jit.TracerWarning)
                 captured_fn = torch.jit.trace(fn, inputs, check_trace=False)
-        expected = _as_tuple(fn(*inputs))
-        actual = _as_tuple(captured_fn(*inputs))
+        with _restoring_rng(torch.device(device)):
+            expected = _as_tuple(fn(*inputs))
+        with _restoring_rng(torch.device(device)):
+            actual = _as_tuple(captured_fn(*inputs))
         if len(expected) != len(actual):
             raise AssertionError(
                 f"size {size}: eager returned {len(expected)} outputs, {capture} returned {len(actual)}"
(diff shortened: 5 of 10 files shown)


ducha-aiki:
## Review — #4034 (`testing/precision.py`, `testing/verify_delta.py`)

Read the full diff (10 files) and probed the two runtime claims I could not settle by reading. The
design holds up: sets-not-counts is the right primitive, `_assert_imports_from` catches the exact
shared-`.venv` trap that makes worktree comparisons lie, and the historical fixtures in
`tests/testing/_historical.py` are the strongest part of the PR — `test_historical_meshgrid_bodies_pass_at_the_vacuous_sizes`
pins the trap itself, not just the fix.

Four things below are worth changing before merge; the rest are nits.

---

### 1. `_run_surface` discards pytest's exit code, so a run that never started reads as "zero failures" (`testing/verify_delta.py:292`)

`subprocess.run(cmd, ..., check=False)` throws away the `CompletedProcess`, and `_failures_or_empty`
turns a missing junit into `set()`. pytest writes no junit when it fails before the session starts
(exit 3 internal error, exit 4 usage error — a broken `conftest.py`, a bad `-k`, a plugin that
fails to load). The surface then produces a real `FailureDelta`, so it is *not* counted toward the
exit-2 "nothing was verified" branch, and the table prints `0 / 0 / 0`.

Reproduced against this branch's own interpreter:

```
$ # tree with a conftest.py that raises ImportError
$ python -m pytest -q -p no:cacheprovider --continue-on-collection-errors --junitxml=probe.xml tests
returncode: 4
junit exists: False
```

Fed through `_run_surface` that is `set()` → `diff_failures(set(), set())` → row `0/0/0` → `main()`
returns 0. A green gate over a tree where nothing ran, which is precisely the failure mode exit 2
was added for (`3a56fa93`, "make verify-delta fail loudly when it verified nothing").

It fails the other way too: the same crash on the **base** tree empties the baseline, so every
branch failure is reported `NEW` and the tool exits 1 — the false positive the `N*` row was
introduced to avoid, minus the `*`.

Suggested fix — keep the return code and treat "no junit + hard exit" as unverified, not empty:

```python
proc = subprocess.run(cmd, cwd=tree, env=env, check=False)
if not junit.exists() and proc.returncode not in (0, 1, 5):
    print(f"{tree}: pytest exited {proc.returncode} without writing {junit}; this surface was not verified")
    return None
return _failures_or_empty(junit)
```

(0 = all passed, 1 = tests failed, 5 = no tests collected are the three cases where an absent or
empty report really does mean zero failures.) `tests/testing/test_verify_delta.py::TestFailuresOrEmpty`
and `test_an_empty_baseline_with_a_clean_branch_still_passes` currently codify the permissive
reading, so they need updating alongside.

### 2. `--base` is resolved in the *reused worktree*, so a relative revision drifts a commit per run (`testing/verify_delta.py:244`)

First run: `_git(repo, "worktree", "add", "--detach", path, base)` — `base` resolves against the
repo. Every later run: `_git(path, "checkout", "--detach", base)` — `base` now resolves against the
**worktree's own HEAD**, which is the previous run's base. Meanwhile `main()` computes
`git diff {base}...HEAD` in `repo`. The two disagree from run 2 onward.

Demonstrated on a scratch repo (commits c1..c4, `--base HEAD~1`):

```
after create,  wt HEAD: f9c0ef4 c3     # correct
after reuse,   wt HEAD: c96e221 c2     # drifted
```

`--base origin/main` and `--base <sha>` are unaffected, but `--base HEAD~1` is documented as
supported (`_ensure_main_worktree`'s own comment names it). Resolve once, up front:

```python
base_rev = _git(repo, "rev-parse", args.base)   # then pass base_rev everywhere
```

That also makes the printed `branch <sha> vs origin/main <sha>` line honest when `origin/main`
moves mid-run.

### 3. `assert_degenerate_path_parity` passes vacuously on a mistyped argument name (`testing/precision.py:195`)

`{**full_kwargs, name: bad}` *adds* an unknown key rather than replacing a known one, so a typo
raises `TypeError: unexpected keyword argument` on **both** paths, parity holds, and the test is
green having exercised nothing:

```python
>>> assert_degenerate_path_parity(op, {"x": ..., "dsize": (2,)}, {"x": ..., "dsize": (0,)},
...                               [("xx", torch.zeros(2, dtype=torch.int64))])   # 'xx', not 'x'
TYPO NAME -> PASSED VACUOUSLY (both raised TypeError: unexpected kwarg)
```

That is the one bug class this module exists to make structurally hard to write, so it is worth
closing here rather than in a follow-up:

```python
for name, _ in bad_inputs:
    if name not in full_kwargs or name not in degenerate_kwargs:
        raise ValueError(f"{name!r} is not in both kwargs dicts; a substituted-in unknown argument "
                         "raises TypeError on both paths and passes vacuously")
```

A second, weaker point on the same helper: parity is checked on exception *type* only. Two paths
can both raise `RuntimeError` for unrelated reasons and read as parity — worth one sentence in the
docstring, since #4031 (the finding this helper produced) is exactly a case where the types
happened to differ.

### 4. `unrepresentable_sizes` documents the `hi` bound but does not enforce it (`testing/precision.py:37`)

The docstring says to keep `hi` below `finfo(dtype).max` "or the result would be garbage", then
returns garbage on request:

```python
>>> len(unrepresentable_sizes(torch.float16, lo=65000, hi=70000))
5001                      # includes sizes past 65504, which cast to inf
```

`raise ValueError` when `hi > torch.finfo(dtype).max` is three lines and turns a documented
foot-gun into an impossible one. Same argument as rule 1 of the skill in #4035.

---

### Nits

- **Doctests in `testing/precision.py` never run in CI.** `pixi run doctest` is
  `pytest --doctest-modules kornia/`; the `[257, 258, 259]` / `[2049, 2050]` examples are
  load-bearing documentation of the bfloat16/float16 boundaries and will rot silently. Either widen
  the task to `kornia/ testing/` or add `--doctest-modules testing/precision.py` to a
  `tests/testing` invocation.
- **`changed_test_dirs` assumes test locality.** `kornia/<mod>` → `tests/<mod>` is right for the
  common case, but a `kornia/geometry` change routinely breaks `tests/augmentation`, and the gate's
  promise ("zero `new` failures") is narrower than it sounds when the default scoping is used. The
  "Cost of touching `testing/`" note in the PR body is the right place for one more sentence saying
  so, and recommending `--tests tests` for a cross-cutting change.
- **`render_table` details lose the surface.** `NEW <id>` lines from several rows are concatenated
  with no surface header, so a half-precision-only regression is indistinguishable from a cpu one in
  the pasted table. `NEW [cpu float16,bfloat16,float64] <id>` costs nothing.
- **Partial baseline is silently a full one.** When `tests` maps to two dirs and the base tree has
  only one of them, `_run_surface` runs the subset and the row is marked `baseline=True` — no `*`.
  Semantically defensible (failures in a branch-only test dir *are* new), but the `N*` legend exists
  to flag exactly this asymmetry and here it does not fire.
- **`Surface` is `frozen=True` with a `dict` field**, so it is unhashable and the `surface not in
  selected` check in `main()` works only because `selected` is a list. Worth a `field(compare=...)`
  note or a tuple of pairs if anyone later reaches for a set.

### What I verified

Read: full PR diff, `testing/precision.py`, `testing/verify_delta.py`, both new test modules,
`tests/testing/_historical.py`, `TESTING.md` and `pixi.toml` deltas. Ran: the two helper probes
quoted above and the pytest-exit-4 and worktree-drift reproductions, on this checkout's `.venv`
(CPU, macOS). Did **not** re-run `pixi run verify-delta` end to end, the MPS surface, or the
inductor surface — the PR body's tables are the evidence there and I did not independently confirm
them.

*Posted on behalf of @ducha-aiki by Claude (Opus 5, 1M context).*

ducha-aiki:
## Round 1 answered — `14b47548`

All four "worth changing before merge" findings and all five nits are fixed, plus both Copilot
inline comments. Every finding was reproduced first; none was declined.

### Triage

Everything here is code this PR introduces, so `regression`/`pre-existing` does not apply — the
question was only whether each claim reproduces. All nine did, on this checkout's `.venv`
(CPU + MPS, macOS, torch 2.9.1, Python 3.11.14).

| # | Finding | Reproduced | Fix |
|---|---|---|---|
| 1 | `_run_surface` discards pytest's exit code | yes — broken `conftest.py` → `returncode 4`, `junit exists: False` | keep the return code |
| 2 | `--base` re-resolved in the reused worktree | yes — scratch repo c1..c4, `--base HEAD~1`: create → `c3`, reuse → `c2` | resolve once, in the repo |
| 3 | `assert_degenerate_path_parity` vacuous on a typo'd name | yes — `("xx", …)` passed, both paths raising `TypeError` | refuse a name absent from either dict |
| 4 | `unrepresentable_sizes` ignores its documented `hi` bound | yes — `len(…(float16, lo=65000, hi=70000)) == 5001`, last `70000` | `ValueError` above `finfo(dtype).max` |
| 5 | doctests in `testing/precision.py` never run in CI | yes — `pixi run doctest` was `--doctest-modules kornia/` | task now covers `testing/precision.py` |
| 6 | `changed_test_dirs` assumes test locality | yes, by inspection | PR body, per your suggestion |
| 7 | `render_table` details lose the surface | yes, by inspection | `NEW [<surface>] <id>` |
| 8 | partial baseline is silently a full one | yes, by inspection | now earns the `*` |
| 9 | `Surface` frozen + `dict` field is unhashable | yes — `hash(SURFACES[0])` → `TypeError: unhashable type: 'dict'` | `eq=False` |
| C1 | Copilot: RNG-consuming `fn` diverges under trace | yes — `x + rand_like(x)` fails at size 1 with the rounding message | restore RNG around each call |
| C2 | Copilot: pytest return code ignored | same as #1 | same as #1 |

### On #1, one step further than you suggested

Your patch gates on `not junit.exists()`. I gated on the return code alone, because a *partial*
junit is the same problem: pytest exit 2 (interrupted) can leave a report covering only the tests
that ran, and diffing a truncated failing set against a complete one is the false delta the tool
exists to prevent. So `returncode not in (0, 1, 5)` → the surface is unverified, junit or not.

That interacts with the empty-baseline branch, which previously read *any* `None` from the base
tree as "no baseline". A base tree that has the paths but crashed is not an empty baseline, so
`main()` now distinguishes the two by `_present_paths(main_wt, tests)`: no paths → `N*` as before;
paths but no result → the row is `skipped`, and if that is the only surface, exit 2.

### On #3, the second half

Added the docstring note that parity is compared on the exception *type* only. Left it there rather
than comparing messages: #4031 is exactly a case where the types differ, and message matching would
make the helper brittle against wording changes.

### On #4

`{finite_max:.0f}` would have printed 39 digits for bfloat16, so the message uses `:g`.

### Delta self-review (prefix-r1..HEAD)

- size cast into a half dtype before a division: none — no arithmetic changed
- degenerate/empty path validates less or more than the full path: yes, found and fixed in my own
  delta — the first cut of `main()` reported a crashed base tree as an empty baseline (a laxer
  path); split into two branches on `base_paths`
- dtype=None resolved before promotion is decided: n/a — no dtype resolution changed
- new compile test carries dynamo/compile in its name: n/a — no new compile test; the existing
  `test_compile_capture_runs_or_skips` still deselects correctly (1 passed, 75 deselected)
- MPS/half skips match the nearest existing test: no skip added or changed; the new RNG test runs on
  MPS and passes there (found and fixed in my own delta: `module.get_rng_state()` reads the
  *current* device, so it now passes `device` explicitly, as `torch.cuda`/`torch.mps`/`torch.xpu`
  all accept)
- docstring/comment/changelog next to the change still true: `TESTING.md` updated for the two new
  refusals; module docstring no longer says the base is `origin/main`; `_run_surface`'s "or `None`
  if pytest could not run" still accurate; CHANGELOG entry unaffected. Found and fixed in my own
  delta: an unresolvable `--base` reached `rev-parse` before any friendly handler and surfaced as a
  raw `CalledProcessError`, now a `SystemExit` with the message (`TestBaseRevisionCannotBeResolved`)
- fix mirrored in the 3d/other sibling, and the sibling's accepted calls still pass: n/a — no
  sibling function; `_failures_or_empty` is deliberately unchanged, its permissive reading is now
  reached only for exit codes where it is correct

### Pre-fix check

All 12 added/changed tests fail at `prefix-r1` (`45b73aa7`) with a real assertion — no collection
errors, no `ImportError`:

```
12 failed, 62 passed
FAILED …test_rejects_an_hi_above_the_dtype_max - Failed: DID NOT RAISE <class 'ValueError'>
FAILED …test_a_function_that_consumes_randomness_is_not_a_false_failure[cpu-float32] - AssertionError: size 1, output 0: trace output differs from eager…
FAILED …test_rejects_a_bad_input_name_that_is_not_an_argument - Failed: DID NOT RAISE <class 'ValueError'>
FAILED …test_surfaces_are_hashable - TypeError: unhashable type: 'dict'
FAILED …test_a_pytest_that_could_not_start_is_unverified_not_clean - AssertionError: assert set() is None
FAILED …test_the_reused_worktree_checks_out_the_repo_resolved_sha - AssertionError: assert None == 'deadbee'
FAILED …test_a_base_tree_where_pytest_could_not_run_is_not_an_empty_baseline - assert 0 == 2
FAILED …test_a_baseline_that_ran_fewer_paths_than_the_branch_is_marked_partial - AssertionError: '| cpu float32 | 0* | 0 | 2 |' not in …
(+4 assertion-message updates)
```

### Gate

`python -m testing.verify_delta --tests tests/testing tests/geometry`, exit 0:

| surface | new | fixed | unchanged |
|---|---|---|---|
| cpu float32 | 0* | 0 | 0 |
| cpu float16,bfloat16,float64 | 0* | 0 | 697 |
| mps float32 | 0* | 0 | 25 |
| inductor cpu float32 | 0* | 0 | 0 |

`* no baseline on the base revision for these paths; every failure there counts as new`

The `*` on every row is finding #8 working: `tests/testing` does not exist on `origin/main`, so the
baseline is partial. Before this commit those rows printed bare `0`, exactly the asymmetry you said
the legend exists to flag and did not fire on.

Also run: `pre-commit run --all-files` (13/13), `uvx ty@0.0.9 check kornia testing`, `tests/testing`
on cpu float32 (74 passed), cpu float32+float16+bfloat16+float64 (81 passed), mps float32 (74
passed, 1 skipped — the documented kornia#4032 skip), inductor `-k "dynamo or compile"` (1 passed,
75 deselected), and `--doctest-modules testing/precision.py` (1 passed).

Not re-run: the full-suite (`--tests tests`) invocation on all four surfaces.

*Posted on behalf of @ducha-aiki by Claude (Opus 5, 1M context).*

ducha-aiki (commented):
Requesting changes because the regression gate can report success without verifying the committed branch across the requested surfaces.

1. **[P1] Fail when any selected surface is unverified** (`testing/verify_delta.py:352-376`). A CPU row can pass while half/MPS/inductor exits 2/3/4 and becomes `None`; because only the all-`None` case returns 2, the overall command still exits 0. Track selected-but-unverified surfaces separately and make any such row nonzero, while distinguishing intentionally unavailable or unselected surfaces.

2. **[P1] Test committed HEAD, not an unchecked live working tree** (`testing/verify_delta.py:319-352`). The scope comes from `base...HEAD`, but pytest runs against the current checkout. An uncommitted fix can therefore make a broken, pushable HEAD appear green. Require a clean worktree or run the branch side from a detached worktree at HEAD.

3. **[P2] Widen top-level test files to `tests`** (`testing/verify_delta.py:58-61,329-332`). Paths such as `tests/test_import.py`, `tests/test_api_surface.py`, and `tests/smoke_test.py` currently map to an empty list, so the command prints “nothing to verify” and exits 0.

4. **[P2] Reject a dirty reused baseline worktree** (`testing/verify_delta.py:241-263`). `git checkout --detach` can preserve non-conflicting tracked modifications, contaminating the baseline and making a branch failure appear unchanged. Check worktree status before reuse and refuse dirty state rather than resetting it.

5. **[P2] Align the equality implementation with the documented contract** (`testing/precision.py:118-123,180-189`). `torch.equal` is not byte equality: identical NaNs compare unequal, while `+0.0` and `-0.0` compare equal. This can both report false capture mismatches and miss sign-bit changes. Either compare byte representations or define exact numerical equality with explicit NaN handling and update the documentation.

6. **[P2] Reject an empty `bad_inputs` sweep** (`testing/precision.py:230-253`). `bad_inputs=[]` checks only the valid baselines and passes without exercising invalid-input parity. This is the same vacuous-green class the helper is intended to prevent.

ducha-aiki:
## Round 2 answered — `e68a2b32`

All six findings fixed. Every one was reproduced first; none was declined.

### Triage

Everything here is code this PR introduces, so `regression`/`pre-existing` does not apply — the
question was only whether each claim reproduces. All six did, on this checkout's `.venv`
(CPU + MPS, macOS, torch 2.9.1, Python 3.11.14).

| # | Finding | Reproduced | Fix |
|---|---|---|---|
| 1 | a selected surface that could not be measured still exits 0 | yes — stubbed `_run_surface` to return `set()` for `cpu float32` and `None` for `mps float32`: table showed both, `main()` returned **0** | three distinct statuses; any `unverified` row is exit 2 |
| 2 | pytest runs the working tree while the scope is `base...HEAD` | yes — an **untracked** `tests/testing/test_zz_uncommitted_probe.py` (`git show HEAD:…` → "exists on disk, but not in 'HEAD'") was collected, failed, and printed as `NEW` | refuse a dirty checkout; `--allow-dirty` to gate the tree instead |
| 3 | top-level test files map to nothing | yes — `changed_test_dirs(["tests/test_import.py"]) == []`, and it is dropped even in a mixed diff | the file maps to itself; a shared file there widens |
| 4 | a dirty reused baseline worktree is not rejected | yes — `echo CONTAMINATED >> README.md` in the baseline, then `_ensure_main_worktree(..., "origin/main~3", ...)`: re-pointed to `11d8d393` with ` M README.md` and the edit still in the file | refuse before `checkout --detach` |
| 5 | `torch.equal` is not byte equality | yes — `torch.equal(nan, nan)` → `False`; `torch.equal(+0., -0.)` → `True` while the bits are `0` vs `-2147483648` | compare raw bytes |
| 6 | an empty `bad_inputs` sweep passes | yes — `assert_degenerate_path_parity(f, {...}, {...}, [])` returned cleanly | `ValueError` |

### On #1, the shape of the fix

`None` was doing two jobs: "nobody asked for this surface" and "this surface was asked for and did
not answer". Splitting them is the whole finding, so the table now has three non-numeric cells —
`not selected`, `unavailable`, `unverified` — and only the last one is nonzero. Exit codes:

- `1` — a `new` failure somewhere (unchanged; still wins over an unverified row, since it is the
  actionable signal and the row is printed either way).
- `2` — a selected, **available** surface produced no measurement. This now covers the old
  "nothing was verified at all" case as a special case of itself rather than as the only case.
- `0` — everything selected and available was measured, and nothing is new.

Your "distinguishing intentionally unavailable or unselected surfaces" is why `--only mps float32`
on a non-Mac still exits 2: nothing was measured, so there is no gate, even though no row is
`unverified`. `test_selecting_only_an_unavailable_surface_is_not_a_pass` pins that it did not
regress while the labels changed underneath it.

### On #2, and the one thing I added

The strict reading — refuse a dirty tree — is the default. I also added `--allow-dirty`, which is
an addition rather than a fix, so calling it out: without it the tool becomes unusable for the
iterate-then-commit loop it is meant to serve, and people would work around it in ways that hide
the problem again. It is loud rather than quiet: the header line becomes
`branch <sha> + uncommitted changes (--allow-dirty) vs origin/main <sha>`, so a pasted table says
which of the two things was measured. Happy to drop it if you would rather the tool only ever
describe a commit.

I did not take the second option you offered (run the branch side from a detached worktree at
HEAD). It is stronger, but it doubles the worktree cost on every run and puts the branch side
behind the same shared-`.venv` indirection the base side already needs `_assert_imports_from` for —
for a gate that is run by hand, refusing seemed the smaller change.

### On #3, one step narrower than you suggested

You suggested widening `tests/test_import.py` and friends to `tests`. I mapped such a file **to
itself** instead: it is a runnable pytest target, so `--tests tests/test_import.py` verifies the
change exactly, where widening costs the full suite on four surfaces for a one-file diff. The
widening is kept for the files under `tests/` that are *not* targets — `tests/__init__.py`,
`tests/benchmark.py` — because running those collects nothing, which would be a pass over an
unmeasured suite. Split as `name.startswith("test_") or name.endswith("_test.py")`, matching
pytest's default `python_files`.

One edge this creates: deleting a top-level test file and nothing else now exits 2 ("none of
tests/test_import.py exist here; nothing to run") where it used to exit 0 with "nothing to verify".
That reads correctly to me — nothing ran — so I left it.

### On #5, what "byte equality" now means

Both halves reproduce, in opposite directions, so `torch.equal` is replaced by a comparison of the
raw bytes (`t.detach().contiguous().reshape(-1).view(torch.uint8)`; the flatten is because
`view(torch.uint8)` needs a last stride of 1 and at least one dimension, and a 0-dim output
otherwise raises). That makes the helper *stricter* than documented, not merely correctly
documented: a NaN now matches a NaN with the same payload, and `+0.0` no longer matches `-0.0`.

The failure message had to follow. When the bytes differ but the values agree it now says
"the values are equal but their bits are not (a signed zero or a NaN payload changed)" instead of
`max abs diff 0`, and the numeric branch runs `nan_to_num` *after* the subtraction so a NaN on one
side only cannot hide the mismatch behind a `nan` difference.

Two regression tests, one per direction: `test_a_sign_bit_change_that_torch_equal_calls_equal_is_caught`
(a trace branch that negates a zero tensor) and `test_matching_nans_are_not_reported_as_a_divergence`
(`0/0` on both sides). At `prefix-r2` the first is `Failed: DID NOT RAISE` and the second fails with
`max abs diff nan` — the spurious report you predicted.

### On #6

`ValueError`, matching `assert_capture_matches_eager`'s empty-`sizes` refusal, since it is the same
class of vacuum.

### Pre-fix verification of the new tests

Probe worktree at `prefix-r2` (= `14b47548`), test files copied in, run under the same interpreter
with the import guard in front. The three new `test_precision.py` tests need no shim; the
`test_verify_delta.py` ones import three constants the tag does not have, so the probe copy defines
them locally — the shim is only the import, the assertions are the shipped ones.

| test | failure at `14b47548` |
|---|---|
| `test_a_test_file_directly_under_tests_maps_to_itself` | `AssertionError: assert [] == ['tests/test_import.py']` |
| `test_a_shared_file_directly_under_tests_widens_to_everything` | `AssertionError: assert [] == ['tests']` |
| `test_a_surface_that_could_not_be_measured_is_not_labelled_like_one_nobody_asked_for` | `AttributeError: 'str' object has no attribute 'baseline'` |
| `test_no_unverified_legend_when_every_selected_surface_ran` | `AttributeError: 'str' object has no attribute 'baseline'` |
| `test_refuses_a_dirty_baseline_worktree_instead_of_carrying_the_edit_across` | `Failed: DID NOT RAISE <class 'SystemExit'>` |
| `test_one_surface_that_could_not_be_measured_sinks_a_run_the_others_passed` | `assert 0 == 2` |
| `test_an_uncommitted_change_is_refused_because_the_run_would_not_describe_head` | `Failed: DID NOT RAISE <class 'SystemExit'>` |
| `test_allow_dirty_runs_but_says_the_branch_side_is_the_working_tree` | `SystemExit: 2` (argparse — the flag does not exist there) |
| `test_a_sign_bit_change_that_torch_equal_calls_equal_is_caught` | `Failed: DID NOT RAISE <class 'AssertionError'>` |
| `test_matching_nans_are_not_reported_as_a_divergence` | `AssertionError: size 1, output 0: trace output differs from eager, max abs diff nan` |
| `test_rejects_an_empty_bad_inputs_sweep` | `Failed: DID NOT RAISE <class 'ValueError'>` |

Two of these are honest but weaker than the rest, so flagging them: the two `render_table` rows
fail on the old signature rather than on an assertion, and `test_allow_dirty_*` fails in argparse.
Both are tests of a shape the tag does not have. The six behavioural ones fail on their own
assertion.

### Delta self-review (prefix-r2..HEAD)

- size cast into a half dtype before a division: none — no arithmetic changed. The one new
  numeric line is `nan_to_num` *after* the subtraction, deliberately, so a one-sided NaN cannot
  swallow a real difference.
- degenerate/empty path validates less or more than the full path: found and fixed in my own
  delta. The first cut of `main()` made a selected-but-unmeasured surface exit 2 *before* checking
  whether anything had been measured at all, which turned `--only "mps float32"` on a non-Mac from
  the existing exit-2 "nothing was verified" into a silent exit 0 — a laxer degenerate path,
  caught by `test_selecting_only_an_unavailable_surface_is_not_a_pass`. Split into `not measured` →
  the old message, then `unverified` → the new one.
- dtype=None resolved before promotion is decided: n/a — no dtype resolution changed.
- new compile test carries dynamo/compile in its name: n/a — no new compile test; the existing
  `test_compile_capture_runs_or_skips` still deselects correctly (1 passed, 86 deselected).
- MPS/half skips match the nearest existing test: no skip added or changed. The two new
  `assert_capture_matches_eager` tests take the `device`/`dtype` fixtures and pass on MPS float32
  and on cpu float16/bfloat16/float64 without one.
- docstring/comment/changelog next to the change still true: three found stale and fixed —
  `render_table`'s comment still named the removed `skipped` label; `assert_capture_matches_eager`'s
  `Raises:` still promised "max abs difference" as the only detail; the CHANGELOG line described
  the fix relative to the review round rather than the behaviour. `TESTING.md`'s helper section
  carried the `torch.equal` claim and the missing empty-`bad_inputs` rule; both corrected.
- fix mirrored in the 3d/other sibling, and the sibling's accepted calls still pass: the sibling
  here is `_dirty`, used for two trees. Both call sites refuse rather than reset, and
  `_ensure_main_worktree`'s existing "is dirty or cannot check out" message lost its now-false
  first half.

### Gate: `pixi run verify-delta -- --tests tests/testing tests/geometry`

`e68a2b32` vs `origin/main` `9e45dddc`, all four surfaces measured, 3m20s:

| surface | new | fixed | unchanged |
|---|---|---|---|
| cpu float32 | 0* | 0 | 0 |
| cpu float16,bfloat16,float64 | 0* | 0 | 697 |
| mps float32 | 0* | 0 | 25 |
| inductor cpu float32 | 0* | 0 | 0 |

`* no baseline on the base revision for these paths; every failure there counts as new`

Exit 0, and — per finding #1 — no row is `unverified`, `not selected` or `unavailable`, so the 0
is a measurement on all four rather than one row carrying the others.

Also run: `pre-commit run --all-files` (13/13), `uvx ty@0.0.9 check kornia testing`, `tests/testing`
on cpu float32+float64+float16+bfloat16 (98 passed), mps float32 (85 passed, 1 skipped — the
documented kornia#4032 skip), inductor `-k "dynamo or compile"` (1 passed, 86 deselected), and
`--doctest-modules testing/precision.py` (1 passed).

Not re-run: the full-suite (`--tests tests`) invocation on all four surfaces.

*Posted on behalf of @ducha-aiki by Claude (Opus 5, 1M context).*

ducha-aiki (commented):
I found four blocking correctness issues:

1. [P1] `assert_capture_matches_eager` reuses the same inputs for tracing, eager execution, and captured execution, while `expected` can remain an alias of a mutated input (`testing/precision.py:175-183`). I reproduced an in-place function where eager adds 1 and tracing adds 2, yet the helper passes because the later captured call mutates the stored expected result. Build independent inputs for each execution and snapshot outputs before the next call; add an in-place regression test.

2. [P1] `_run_surface` treats pytest exit code 5 (nothing collected) as a successful empty measurement (`testing/verify_delta.py:359-366`). Consequently, an inductor surface where `-k "dynamo or compile"` selects no tests gets a numeric clean row and can exit 0 despite exercising nothing. Return `None`/`unverified` for exit 5 and cover the resulting exit-2 behavior.

3. [P1] `--allow-dirty` runs the working tree, but automatic test selection still comes solely from `git diff base...HEAD` (`testing/verify_delta.py:375-395`). An uncommitted regression in another module, or on an otherwise unchanged branch, can be omitted while the command reports success. Include tracked and untracked working-tree paths in the scope, or require explicit `--tests` when dirty.

4. [P1] `--tests` accepts absolute and parent-traversal paths (`testing/verify_delta.py:216-218`). Joining an absolute branch test path to either tree ignores the tree, so both runs can execute the branch test source and classify a regression as unchanged. Normalize targets and require relative paths confined to each checkout.

Focused validation passed: 98 tests under `tests/testing` across float32, float16, bfloat16, and float64.

ducha-aiki:
## Review

Scope reviewed: `9e45dddc...e68a2b32` (10 files, +1928). I ran the new suite on cpu/float32, cpu/half+float64, MPS, and inductor (all green), ran the doctest, ran `ruff` at the pinned pre-commit version (clean), and ran `verify_delta` end-to-end twice against a real worktree.

### 1. `testing/verify_delta.py:391` — medium. `--allow-dirty` gates the working tree but scopes from the committed diff only, so the uncommitted change is never tested

`changed = _git(repo, "diff", "--name-only", f"{base}...HEAD")` ignores the working tree entirely, while `--allow-dirty` exists precisely to "gate the working tree instead".

Reproduced: with an uncommitted edit to `kornia/geometry/grid.py` and no new commits,
`python -m testing.verify_delta --allow-dirty --base HEAD --only "cpu float32"`
printed `no test directories map to the changed files; nothing to verify` and exited **0** — a clean gate verdict over a change that was never run.

The same hole applies in the ordinary case: an uncommitted edit to a module outside the committed diff's `tests/<mod>` mapping is silently out of scope. Under `--allow-dirty`, scope should be the union of `base...HEAD`, `git diff --name-only HEAD`, and `git ls-files --others --exclude-standard`.

### 2. `testing/verify_delta.py:44` (`changed_test_dirs`) — low/medium. The gate's "no new failures" promise is narrower than it reads, and the caveat lives only in the PR description

`kornia/<mod>` → `tests/<mod>` assumes test locality, but e.g. a `kornia/geometry` change routinely breaks `tests/augmentation`; the run then exits 0 with an all-zero table. The PR body documents this honestly, but nothing durable does — not the module docstring, not `--help`, not the `pixi` task description, not the CHANGELOG entry, not `TESTING.md` (which never mentions `verify-delta` at all). After merge the only surviving text is the CHANGELOG line, which reads as an unqualified cross-surface gate.

### 3. `testing/verify_delta.py:319` — low/medium. `_assert_imports_from` crashes the run instead of marking the surface unverified, and swallows the reason

`subprocess.run(..., check=True, capture_output=True)` is not wrapped. If `import kornia` fails in either tree — realistically an older `--base` that no longer imports under the current dependency set, or a branch adding a module that raises at import — the tool dies mid-run with `CalledProcessError: ... returned non-zero exit status 1`, and the child's stderr is captured and discarded, so the user gets no traceback from the failing import. Every other git failure in this file is converted to a clean `SystemExit` carrying `exc.stderr`; this one should do the same, or return `None` so the surface is reported `unverified`.

### 4. `testing/verify_delta.py:269` — low. A `git fetch` failure escapes as a raw traceback

`subprocess.run(["git","-C",...,"fetch",remote,ref], check=True)` has no `try`. Offline, or in a clone whose remote is not named `origin`, the default invocation aborts with an uncaught `CalledProcessError` rather than the `SystemExit` style used by `_ensure_main_worktree`'s neighbouring `rev-parse`/`checkout` handlers. `--no-fetch` is the workaround but nothing points the user at it.

### 5. `testing/verify_delta.py:343` — low. `PYTHONPATH` is replaced, not prepended

`env = {**os.environ, **surface.env, "PYTHONPATH": str(tree)}` drops any `PYTHONPATH` the caller had. Where a dependency is supplied through it, both trees now fail to import; combined with finding 3 that surfaces as an opaque `CalledProcessError` rather than an `unverified` row. Prepend with `os.pathsep`.

### 6. `testing/verify_delta.py:218` (`_present_paths`) — low. An absolute `--tests` path silently makes both trees run the *same* test files

`(tree / t)` with absolute `t` yields `t` itself (`Path('/a/b') / '/x/tests/geometry' == Path('/x/tests/geometry')`), so `--tests /Users/.../kornia/tests/geometry` hands the branch checkout's test files to the base tree's pytest run. The import guard still passes (kornia resolves from the base tree via cwd + `PYTHONPATH`), so the table reports a delta between the branch's tests-on-base-library and the branch's tests-on-branch-library — not the pair it names. Rejecting absolute `--tests`, or relativizing them to the repo root, closes it.

### 7. `testing/precision.py:110` (`_as_tuple`) — low. A function returning no tensors passes with zero comparisons

`_as_tuple(())` hits `all(...)` over an empty sequence → `True` → returns `()`; both sides have length 0 and the per-output loop never runs. Confirmed: `assert_capture_matches_eager(lambda x: (), ..., sizes=[1, 2, 258], dtype=torch.bfloat16)` returns cleanly. This is the same vacuous-green class the module already guards for empty `sizes` (L156) and empty `bad_inputs` (L249); a `len(expected) == 0` guard would make it consistent. Same for outputs that are all zero-element tensors.

### 8. `testing/verify_delta.py:25` — low. The docstring's exit-code contract contradicts the code for `--only` on an unavailable surface

It says a surface excluded by `--only` (`not selected`) or absent from the machine (`unavailable`) "is not a gap and does not affect the code". But when the only selected surface is unavailable, `measured` is empty and `main` returns **2** — which `test_selecting_only_an_unavailable_surface_is_not_a_pass` asserts as intended. The docstring needs the exception spelled out ("...unless it leaves nothing measured").

### 9. `testing/precision.py:50` (also `:71`, `TESTING.md:279`, `tests/testing/test_precision.py:62`) — low. The stated rationale for the `hi > finfo.max` guard is inverted

All four places say that above `finfo(dtype).max` a candidate "casts to `inf` and round-trips back to the same int64", so `n` would read as *exact*. It does the opposite: `torch.tensor([70000]).to(torch.float16).to(torch.int64)` is `9223372036854775807`, so such candidates round-trip to a *different* int and are flagged **inexact** — i.e. they would be silently *included* in the sweep. The guard itself is right; only the explanation is backwards, in a PR whose subject is precision correctness.

### Not findings, but worth knowing

- `ruff` 0.14.10 in the local `.venv` flags `S603` on `_assert_imports_from` (no `# noqa`, and `testing/*` per-file-ignores don't cover `S`), but the pinned pre-commit `ruff` 0.16.1 passes cleanly, so CI is unaffected.
- `WIDEN_PREFIXES` lists `"tests/conftest.py"`, which does not exist in this repo — the live root `conftest.py` entry is what actually fires. Dead but harmless.
- `monkeypatch.setattr(verify_delta.subprocess, "run", fake)` patches stdlib `subprocess.run` process-wide (there is no module-local alias to scope it to); fine under the current single-process test run.
- The recommended idiom `sizes=[1, 2, *unrepresentable_sizes(dtype)[:8]]` degrades to `[1, 2]` under the default `--dtype=float32`, so the rounding sweep is inert on the only configuration CI runs. That is inherent to float32 and is stated in the docstring; it is the gap #4033 is filed for.

---

ducha-aiki:
Addressed the remaining review findings in `fbd66038`.

- `assert_capture_matches_eager` now builds independent trace, eager, and captured inputs and snapshots every output before the next execution. The new in-place regression reproduces the aliasing false pass.
- Empty output sequences are rejected as vacuous. Zero-element tensor outputs remain supported because shape, dtype, and byte representation are still meaningful contracts.
- Pytest exit code 5 is now `unverified`; only exit codes 0 and 1 count as measurements.
- Dirty working-tree runs require both `--allow-dirty` and explicit `--tests`, with documentation making clear that the caller must include every affected target.
- Explicit test targets are normalized and confined to each checkout. Absolute paths, parent traversal, and symlinks escaping the checkout are rejected.
- Baseline fetch and import-guard failures now produce actionable errors instead of raw tracebacks, and existing `PYTHONPATH` entries are preserved after the checkout path is prepended.
- The automatic module-to-test mapping is documented as local rather than complete; cross-cutting changes should use `--tests tests`.
- The unavailable-only exit behavior and the float16 non-finite-to-int sentinel rationale are now documented accurately.

Validation completed locally:

- CPU float32 focused suite: 96 passed.
- CPU float16/bfloat16/float64 focused suite: 104 passed.
- MPS float32 focused suite: 96 passed, 1 documented skip.
- Inductor compile test: 1 passed.
- `pixi run pre-commit-all`: passed.
- `pixi run typecheck`: passed.
- Fresh delta review: no remaining actionable findings.

GitHub CI is running on the pushed head.

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (3c743194) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4034.added.md`.

`pixi.toml` also conflicted: the merge keeps the task block from `main` (no `-v`, plus `test-mps` and `doctest-weights`) and re-adds `verify-delta` and the `testing/precision.py` doctest target in both doctest tasks.

ducha-aiki (commented):
Review of `3c743194900ad3bdd291606e02272b01895c7748`.

Rechecked the integration at this head. The previously repaired helpers and their regression tests are unchanged from fbd66038; the merge retains the current Pixi task block and includes the precision doctest target in both doctest tasks. The Towncrier fragment is present.

Validation: tests/testing on CPU float32: 97 passed. This remains a draft, so I am leaving it in the draft queue rather than treating it as ready to merge. Its freshly queued full matrix, particularly the previously problematic Windows legs, still needs to complete; no fresh Windows/CUDA run was performed locally.

victorwon2001:

CI observations do not establish causality where classification is unknown; flaky is a suspicion, not a confirmed cause.

The causes of the checks classified as "unknown" remain unverified.

Obtain diagnostic logs tied to the relevant immutable check snapshots before assigning causes to the unknown classifications; otherwise wait for fresh CI evidence, maintainer activity, or a target SHA change rather than rerunning unchanged checks.

SHA: `3c743194900ad3bdd291606e02272b01895c7748`

PR #4035, opened 2026-08-26
docs(skills): kornia-review-loop and kornia-precision-testing skills

ducha-aiki:
## Why

On kornia#4006, 20 of the ~24 actionable findings after review round 2 were introduced by an
earlier fix commit on the same PR — the fix loop, not the reviewer, was generating defects; the
same half-precision rounding bug was fixed three times, one layer deeper each round. The knowledge
of how to avoid that loop existed, but only as one agent's private memory and this repo's
hand-offs — invisible to other models, other sessions, and anyone without that history. This PR
puts the process where any coding agent reading the kornia repo can find it: two skills plus an
`AGENTS.md` pointer, so the fix is repo-visible and model-agnostic rather than living in one
agent's head.

## What

**`.claude/skills/kornia-review-loop/SKILL.md`** — trigger: "Use when responding to code-review
findings on a kornia PR — triaging each finding, deciding what is in scope, and pushing fix
commits." A rigid, one-todo-per-step workflow:

1. Tag the pre-fix SHA (`git tag -f prefix-<round> HEAD`), never pushed.
2. Triage every finding against both the branch and an `origin/main` worktree probe before
   writing code — regression / pre-existing / wrong, never dropped silently.
3. Fix defects only; anything a finding makes you want to *add* becomes a linked follow-up issue,
   and a same-class fix to a sibling function is in scope only after the main probe has run on
   that sibling too.
4. Write tests that cannot pass vacuously, and confirm each one fails on the pre-fix tag before
   trusting it.
5. Self-review the delta with a fresh subagent against a fixed defect-class checklist; if you
   cannot dispatch one, do the review yourself and paste the checklist with outcomes into the
   reply — the step isn't done until that artifact exists.
6. Gate: `pixi run verify-delta`, zero `new` failures on every available surface, table pasted
   into the reply.
7. Grep for closed issue numbers surviving in the touched modules.
8. Push once per round, then confirm `mergeStateStatus` isn't `DIRTY` and check-runs exist for the
   new head.
9. After two incremental rounds, request one full fresh review with no prior context.

A red-flags table closes it out (e.g. "the reviewer asked for it, so it's in scope" → findings
expose defects, they don't authorize additions).

**`.claude/skills/kornia-precision-testing/SKILL.md`** — trigger: writing or reviewing a kornia
test touching dtype (especially float16/bfloat16), device, `torch.jit.trace`/`torch.compile`, or
degenerate sizes. Eleven rules, each pointing at the helper or convention that enforces it: sweep
`unrepresentable_sizes`, never hand-pick a size; divide by the unrounded size and cast the
quotient under capture; resolve `dtype=None` to `torch.get_default_dtype()` *before* deciding to
promote (and sweep a `torch.set_default_dtype(half)` + `dtype=None` cell, since explicit-dtype
scans can't see this defect); guard a cast-back on `is_floating_point()`; degenerate-path parity
via `assert_degenerate_path_parity`; byte-equality (`torch.equal`) between eager and capture;
compile tests carry `dynamo`/`compile` in the name; MPS/half skips copy the nearest existing
test's skip; a regression test must fail on the pre-fix SHA; the `_historical.py` fixtures are
trace-only; and a `torch.compiler.is_compiling()` branch needs a `capture="compile"` sweep, not
just `"trace"`.

**`AGENTS.md` "Skills" section** — added after "Library preferences", listing all three repo
skills (`kornia-developer`, `kornia-precision-testing`, `kornia-review-loop`) with their triggers
for agents that don't scan `.claude/`, plus a paragraph on `pixi run verify-delta`'s surfaces,
`--tests`/`--only` scoping, and the exit-2 "nothing was verified" meaning.

**`kornia-developer` cross-link** — the new `AGENTS.md` section and both new skills' "Related"
sections point back to `kornia-developer` for compile-first changes, and it is named alongside the
other two rather than left to be discovered separately.

Stacked on PR-A (`testing/precision.py`, `testing/verify_delta.py`); this branch is PR-A plus the
skills and docs. Base will flip from PR-A's branch to `main` once PR-A merges.

## How it was evaluated

**Method.** Skills were graded by ACES-style paired with/without runs (NVIDIA, *Evaluating Skills,
Not Just Agents*, arXiv:2608.20614) rather than static review of the skill text, on three
historical kornia#4006 review rounds; each case is the pre-fix commit plus the reviewer's verbatim
comment, replayed once with both skills loaded and once without, one dispatched agent per arm, the
same model held constant within a round so the comparison isolates the skill rather than model
choice. Every run was graded blind by a separate Opus grading agent using deterministic
reproduction probes — trace-vs-eager `torch.equal` at bf16 258/300, plus a warp-parity probe
matched to each case's defect — against a key defined as "does the fix contain the defect the
*next* historical round found," together with five independent behaviour checks (baseline probe,
scope discipline, pre-fix test verification, verify-delta gate, self-review artifact), plus a
manual controller read.

### Opus 5 (both arms)

| Case | Arm | Key | Behaviours | Qualitative |
|---|---|---|---|---|
| A (`create_meshgrid` trace divergence) | without | PASS | 4/5 — fails #1, #5 | Correct fix, matches eager bit-for-bit; the better *code* — its cast placement avoids new eager-vs-`torch.compile` divergences the with-arm introduces |
| A | with | PASS | 3/5 — fails #2, #5 | Correct fix, byte-identical to eager; the better *review artifact* (real merge-base baseline, four-surface gate) — but introduces 11 new eager-vs-`torch.compile` divergences in `conversions.py` |
| B (`normal_transform_pixel` promotion) | without | PASS-by-silence | 3/5 — fails #5; #1/#4 weak | Correct in-scope fix; left the wave-9 `dtype=None`+half-default-dtype defect untouched, without ever claiming it correct |
| B | with | PASS-by-silence | 4/5 — fails #2 | Correct in-scope fix; also left wave-9 untouched, but its own self-review script contained the exact probe that would have caught it — aimed one function short |
| C (`_empty_warp_output_2d` validation) | without | PASS | 3/5 — fails #1, #5 | Correct core fix, but its sibling-function change (`warp_perspective3d`, `or`→`and`) silently breaks a call that works on `main`, never probed |
| C | with | PASS | 3/5 — fails #2, #5 | Correct core fix; probed main and merge-base first, fixed the sibling correctly and matched upstream — but both arms miss the same latent dtype-guard prize |

**Honest reading.** Key 3/3 both arms — no discrimination at this tier, matching the paper's
caveat that baseline lift from process scaffolding shrinks as the underlying model improves. The
with-arm always probed a real `main`/merge-base worktree and ran the four-surface gate; the
without-arm shipped one regression (C, the `warp_perspective3d` guard tightening); the with-arm's
own gain came at a cost — it introduced new eager-vs-`torch.compile` divergences (A). Four rules
were added to the skills as a result:

a. `kornia-review-loop` step 5: if you cannot dispatch a subagent, perform the delta self-review
   yourself and paste the checklist with outcomes into the reply.
b. `kornia-precision-testing`: a `torch.compiler.is_compiling()` branch needs a `capture="compile"`
   sweep, not only `"trace"`.
c. `kornia-precision-testing`: a `dtype=None` function's sweep must include the
   `torch.set_default_dtype(half)` + `dtype=None` cell — explicit-dtype scans cannot see it.
d. `kornia-review-loop` step 3: a same-class fix to a sibling function is in scope only if the
   main probe has also been run on that sibling.

### Sonnet 5 (both arms), skill text after those rules

| Case | Arm | Key | Behaviours | Qualitative |
|---|---|---|---|---|
| A (`create_meshgrid` trace divergence) | without | **FAIL** | 3/5 — fails #1, #5 | Reproduced the next-wave defect verbatim: casts the divisor before dividing; new test pins only size 257, which passes vacuously against this exact bug |
| A | with (rerun) | **PASS** | 4/5 — fails #2 | Independent third formulation, no cast on the divisor; regression test sweeps `unrepresentable_sizes[:8]` and is shown to fail against the without-arm's own tree |
| B (`normal_transform_pixel` promotion) | without | **INCOMPLETE** | 1/5 — only #2 passes | Never touched the primary defect; the reply nonetheless claims it was "confirmed already present" — an unverified assertion of verification |
| B | with | **PASS-by-silence** | 2/5 — fails #1, #2, #5 | Primary defect fixed correctly; never claims `normal_transform_pixel` is verified, so it doesn't trip the key, but its own vendored helper names the exact vacuity trap it never pointed at that function |
| C (`_empty_warp_output_2d` validation) | without | **PASS** | 2/5 — fails #1, #3, #5 | Correct, surgical fix; explicitly declined to touch the sibling `or` guard as out of scope. No regression |
| C | with (rerun) | **FAIL — REGRESSION** | 3/5 — fails #2, #5 | The core fix matches upstream, but the run extended the same tightening into the sibling `warp_perspective3d`, breaking a call that works at base — its merge-base probe covered only what the sibling *rejects*, never what it *accepts* |

**Honest reading.** Without skills the model repeats the historical mistakes: A reproduces the
defect verbatim, B skips the defect entirely while claiming verification it never ran. With
skills, A and B are fixed — but C regresses via scope expansion into an unprobed sibling success
path. Net key: without arm 1/3 clean passes (+1 incomplete); with arm 2/3, with the third an
outright regression the without-arm run did not commit. Behaviours are higher with the skill in
every case (A: 3→4, B: 1→2, C: 2→3). The self-review step (behaviour #5) went unexecuted in 10 of
12 runs across both rounds — the single largest gap between the skill text and the runs' behaviour.

### Confound found and handled

`origin/main` squash-merged kornia#4006 as `4dd9c6aa` mid-experiment, after the worktrees were
created but visible to every run's `git fetch`. From that point the with-arm's "probe main" step
could read the already-fixed source, and it leaked into two Sonnet with-arm runs — A-with's fix
was byte-identical to `4dd9c6aa`'s source, and C-with's guard matched `origin/main`'s accepted
resolution because it had read `origin/main` directly. Both were rerun with the PR's merge-base
(`0685b1f5`) as the mandated probe target and `origin/main` off-limits; leakage was checked and
confirmed absent in both reruns, and the table above uses those clean reruns for A-with and
C-with. The three Opus with-arm runs had already used the merge-base on their own, unprompted, so
none of the Opus round needed a rerun.

### Changes made from the Sonnet round

Two edits went into this PR's final commit as a direct result:

a. The sibling rule (refinement d, above) now also requires probing what the sibling *currently
   accepts* (unbatched, empty, singleton inputs) at the merge-base, with a test per call — not
   just re-confirming what it already rejects. C-with's regression was exactly a probe that
   checked rejection and never acceptance.
b. The delta self-review (step 5) is now a required fenced section of the reply itself, not a
   described step in prose — an omitted review is visually absent from the reply rather than
   silently missing from a bullet list.

### What this does and does not show

Six cases per tier, one run each, no statistics — this is not a claim of measured effect size.
What it shows is a process difference (a real baseline probe, the half/MPS/inductor gate run and
pasted as evidence) that holds at both model tiers, and, at the Sonnet tier, an outcome difference
on 2 of the 3 cases. It is not proof that the rounds-per-PR number on kornia#4006 itself would
drop — that has to be measured on the next real PRs, not replayed history. Full trajectories,
grades, and probes are recorded in the maintainers' bookkeeping repo
(`superpowers/specs/2026-08-26-skill-evals/`).

---


ducha-aiki:
## Review — #4035 (`kornia-review-loop`, `kornia-precision-testing`)

Read both skill files, the `AGENTS.md` section, and the `kornia-developer` cross-link, and checked
every claim the skills make about the code in #4034 (helper names, the `ValueError` on an empty
sweep, the four `--only` surface names, the exit-0/1/2 semantics, the `N*` row, and the `AGENTS.md`
closed-issue grep rule cited by step 7). All of them check out against the branch — including the
one I most expected to be stale, the `TESTING.md § "Precision and Degenerate-Path Helpers"`
reference, which lands on `TESTING.md:271`.

The framing is right: these encode process, not taste, and every rule traces to a specific defect
with a commit behind it. Two things to fix, then some smaller ones.

---

### 1. Step 2 mandates the probe target the PR's own evaluation identifies as the confound

Step 2 says:

> reproduce it on the branch AND on a worktree of `origin/main`. Create the probe once per round —
> `git worktree add ../main-probe origin/main`

But this PR's description documents exactly why that is the wrong target:

> `origin/main` squash-merged kornia#4006 as `4dd9c6aa` mid-experiment […] From that point the
> with-arm's "probe main" step could read the already-fixed source, and it leaked into two Sonnet
> with-arm runs […] Both were rerun with the PR's merge-base (`0685b1f5`) as the mandated probe
> target and `origin/main` off-limits

And step 3, two paragraphs later, already switched vocabulary without step 2 following:

> enumerate its working calls **at the merge-base** (unbatched matrix, empty `dsize`, singleton axis)

So the skill instructs the target that produced two invalid runs, and the same skill elsewhere
instructs the corrected one. On a long-lived PR this is not hypothetical: once `main` moves past the
branch point, a *regression the branch introduced* can reproduce on `main` too (someone else's
change), and step 2's triage will file it "pre-existing — do NOT fix it in this PR." That is a
false negative in the step whose whole job is preventing false triage.

Concretely, step 2 should read:

```bash
BASE=$(git merge-base origin/main HEAD)
git worktree add --detach ../main-probe "$BASE"
```

…with a sentence saying why: `origin/main` answers "is this broken today?", the merge-base answers
"did *this branch* break it?", and only the second is a triage verdict. `--detach` is worth adding
explicitly too — it is what #4034's `_ensure_main_worktree` does, and it makes the intent survive a
future git DWIM change (today `git worktree add <path> origin/main` does detach; I confirmed that).

The red-flags row *"The finding came with numbers, it must be right" → "Re-measure on main."* needs
the same edit. Note that verify-delta's `--base origin/main` default is **correct** for step 6 (a
regression gate wants "does the tip break") — so the two steps deliberately differ, and saying that
out loud is better than leaving them looking inconsistent.

### 2. The hand-written probe command has the exact failure mode #4034 added a guard for

Steps 2 and 4 both prescribe:

```bash
(cd ../main-probe && PYTHONPATH=$PWD "$OLDPWD/.venv/bin/python" -m pytest <test> -q)
```

The quoting is careful and correct (`$PWD` expands after the `cd`, `$OLDPWD` is the repo, the
explicit interpreter avoids answering under a different torch build). But `PYTHONPATH` is not
sufficient here, and #4034 says so in its own source:

> `_assert_imports_from` — Refuse to run when `import kornia` in `tree` resolves to a checkout other
> than `tree`. […] the `.venv` here is shared/symlinked across worktrees, and `uv run` re-points the
> editable install to whichever tree ran last, so without this guard a stale import would silently
> verify the wrong code.

That guard runs *with `PYTHONPATH` already set* — because an editable install's `__editable___*`
meta-path finder is consulted before `sys.path`. So the skill's probe can silently answer from the
branch tree while the agent records it as a `main` measurement, and the triage verdict flips from
"regression" to "pre-existing" with no visible symptom. This is the single highest-leverage
correctness bug in the skill, because every step-2 verdict rests on it.

Cheapest fix — one line in front of the probe:

```bash
(cd ../main-probe && PYTHONPATH=$PWD "$OLDPWD/.venv/bin/python" -c \
   'import kornia,pathlib,sys; p=pathlib.Path(kornia.__file__).resolve()
    sys.exit(0 if p.is_relative_to(pathlib.Path.cwd()) else f"kornia imported from {p}")' \
 && cd ../main-probe && PYTHONPATH=$PWD "$OLDPWD/.venv/bin/python" -m pytest <test> -q)
```

Better still: have step 2 reuse the machinery instead of duplicating it —
`pixi run verify-delta -- --base <merge-base-sha> --tests tests/<module> --only "cpu float32"`
already creates the worktree, runs the guard, and diffs the failing sets, which is the triage
question. That would also close the smaller duplication below.

---

### Smaller points

- **Two worktrees of the same base.** Step 2 keeps `../main-probe`; step 6 has verify-delta
  maintaining `../.<repo>-verify-main` at the same revision. Two checkouts of kornia is ~2× the
  disk and one more thing to `git worktree remove`. Point step 2 at `--main-worktree`/the existing
  one, or drop `../main-probe` per the suggestion above.
- **`.venv/bin/python` is POSIX-only.** Windows contributors have `.venv/Scripts/python.exe`.
  One parenthetical is enough; kornia CI covers Windows.
- **Step 4's copy-the-test-file-in trick can fail on a test that imports new infrastructure.** If
  the round added `testing/precision.py`, copying the new test into the `prefix-<round>` worktree
  gives `ImportError`, which reads as "it fails on the pre-fix SHA" — a green light for a test that
  is actually unrunnable there. Worth one sentence: an `ImportError`/collection error is *not* the
  failure you are looking for; copy the helper module across too, or assert on the specific
  `AssertionError`.
- **Step 1's tag name.** `git worktree add ../prefix-probe prefix-<round>` works (a tag is a
  commit-ish, so HEAD detaches), but `--detach` makes it explicit and matches step 2's suggested
  form.
- **`kornia-precision-testing` rule 1 vs #4034's helper.** The rule tells the reader to pass
  `sizes=[1, 2, *unrepresentable_sizes(dtype)[:8]]` because the bare call is empty on float32.
  Good — but #4034's `unrepresentable_sizes` still accepts an `hi` above `finfo(dtype).max` and
  returns meaningless sizes. If that gets a `ValueError` there (see my #4034 comment), rule 1 can
  drop to a single sentence.
- **Rule 3's snippet is `torch.set_default_dtype` without a fixture.** Correct as written
  (`try/finally`), and it matches `_ambient_default_dtype` — but it is a process-global mutation and
  the skill is teaching it to agents who will paste it. One clause noting it must never be left
  outside a `finally`, and never inside a fixture that yields, would earn its space.
- **`AGENTS.md` wording.** "agents that do not read that directory should open the files directly"
  is right, but the paths are not given. `.claude/skills/<name>/SKILL.md` spelled out once makes the
  section self-sufficient for exactly the readers it is written for.

### On the evaluation section

The honest reading is genuinely honest — reporting that the with-arm shipped a regression in case C
that the without-arm did not, and that behaviour #5 went unexecuted in 10 of 12 runs, is the part
that makes the rest credible. Two observations:

1. The 10-of-12 finding is the strongest result in the writeup and it is acted on correctly (step 5
   becoming a required fenced section rather than prose). That mechanism — *make the omission
   visible in the artifact* — is worth generalising: step 2's triage verdicts and step 6's table are
   both described in prose and both could go the same way.
2. `n=1` per cell with no seeds and a mid-experiment confound means the Sonnet 2/3-vs-1/3 line
   should not carry weight in the merge decision; the writeup already says so ("this is not a claim
   of measured effect size"). I would not weaken it further — just noting I read the merge case as
   resting on "these rules each name a real defect with a commit behind it," not on the tables.

### Merge order

Base is `feat/testing-precision-helpers`; retarget to `main` once #4034 lands (already flagged in
the description). The `CHANGELOG.md` line here sits directly under #4034's, so a rebase will
conflict trivially — expected, not a problem.

### What I verified

Read: both new `SKILL.md` files in full, the `kornia-developer` addition, the `AGENTS.md` and
`CHANGELOG.md` deltas. Cross-checked against the #4034 branch: helper names and signatures,
`ValueError` on empty `sizes`, the four surface names in `SURFACES`, exit codes 0/1/2, the `N*`
legend, `TESTING.md:271`, and `AGENTS.md:91`'s closed-issue grep rule. Ran: a scratch-repo check
that `git worktree add <path> origin/main` detaches rather than DWIM-creating a local branch. Did
**not** re-run any part of the evaluation or execute the skills end to end.

*Posted on behalf of @ducha-aiki by Claude (Opus 5, 1M context).*

ducha-aiki:
## Round 1 answered — `8968bc9a` (rebased onto #4034's `14b47548`)

Both major findings and all seven smaller ones are fixed. One of them turned out to be a live
example of its own class, which is worth reporting up front.

### 1. Step 2 now probes the merge-base

You are right that the skill mandated the target its own evaluation section identifies as the
confound, and that step 3 had already switched vocabulary without step 2 following. Step 2 now
opens with

```bash
BASE=$(git merge-base origin/main HEAD)
```

and says why in your words: `origin/main` answers "is this broken today?", the merge-base answers
"did *this branch* break it?", and only the second is a triage verdict. Both failure directions are
named — a regression the branch introduced reproducing on a moved `main` and getting filed
"pre-existing", and the PR's own work landing on `main` by another route and being read back as the
baseline. The paragraph closes by saying out loud that verify-delta's `--base origin/main` default
is deliberately the *other* question, so the two steps do not read as an inconsistency. The
red-flags row now says "Re-measure at the merge-base", and `AGENTS.md`'s one-line summary no longer
says "triage against `origin/main`".

`--detach` is on both worktree commands (step 2's and step 4's `prefix-probe`).

### 2. The import guard — and the guard itself had the same defect

Adopted your stronger suggestion: step 2's primary probe is now

```bash
pixi run verify-delta -- --base "$BASE" --tests tests/<module> --only "cpu float32"
```

which creates the worktree, runs `_assert_imports_from`, and diffs the failing sets — the triage
question in one command. Verified end to end on this branch: `--base <sha>` survives the `--`
separator, re-points the worktree, and exits 0.

The hand-rolled form is kept for a non-pytest repro, with the guard in front. **The guard as you
wrote it does not run** — `python -c` rejects the indented continuation lines:

```
$ (cd "$PROBE" && PYTHONPATH=$PWD "$OLDPWD/.venv/bin/python" -c \
   'import kornia, pathlib, sys
    p = pathlib.Path(kornia.__file__).resolve()
    sys.exit(0 if p.is_relative_to(pathlib.Path.cwd()) else f"kornia imported from {p}")')
  File "<string>", line 2
    p = pathlib.Path(kornia.__file__).resolve()
IndentationError: unexpected indent
```

It exits non-zero, which in an `&&` chain reads as "the guard fired" — a false regression verdict
from the guard against false verdicts. It is now a single line, executed both ways before landing,
and the skill says why it must stay on one line.

While testing it I also measured the mechanism, and the writeup differs slightly from your comment.
This `.venv` installs kornia as a `.pth` path entry (`_editable_impl_kornia.pth`, currently
containing `…/kornia-review-skills`), not an `__editable___*` meta-path finder — so here
`PYTHONPATH` *does* win, and the guard passes. The trap is still real and still bites:

```
$ (cd "$PROBE" && python -c 'import kornia; print(kornia.__file__)')   # no PYTHONPATH
/Users/…/kornia-review-skills/kornia/__init__.py     # a different checkout than the cwd
```

So the skill now says: the shared `.venv` points at whichever tree installed last; whether
`PYTHONPATH` beats it depends on how the editable install happened to be built (a `.pth` entry
loses to it, an `__editable___*` finder does not, being consulted before `sys.path` at all) — so
assert the answer instead of reasoning about the ordering. Same conclusion as yours, stated as
something I measured rather than something I assumed.

### Smaller points

| Point | Done |
|---|---|
| Two worktrees of the same base | `../main-probe` is gone; step 2 uses verify-delta's, and says not to make a second beside it |
| `.venv/bin/python` is POSIX-only | `(Windows: .venv/Scripts/python.exe.)` |
| Step 4's copy-in can `ImportError` | copy across every module the test newly imports; read the failure line (`AssertionError`, `Failed: DID NOT RAISE`), never a bare non-zero exit; an `ImportError` means it never ran at the tag |
| Step 1's tag worktree | `--detach` |
| Rule 1 vs the helper | `hi` above `finfo(dtype).max` now raises in #4034, so rule 1 drops the caveat and keeps the sweep point |
| Rule 3's `set_default_dtype` | added: the restore must sit in a `finally` the assertion cannot jump over, never in a fixture that `yield`s — a later fixture failure or a `KeyboardInterrupt` leaves the session in half precision |
| `AGENTS.md` paths | spelled out as `.claude/skills/<name>/SKILL.md` per bullet |

Step 6's `N*` and `skipped` wording is updated for #4034's behaviour change (`N*` now also covers a
*partially* present baseline; a base tree where pytest did not finish is `skipped`, never clean),
and `NEW <id>` became `NEW [<surface>] <id>`. Rule 5 names the new `bad_inputs` refusal.

### On the evaluation section

Took observation 1: step 2's verdicts are now a required per-finding reply line
(`<finding> — regression | pre-existing | wrong — <the two measurements>`), the same
make-the-omission-visible mechanism that fixed step 5. Step 6 already required pasting the table, so
I left it. Nothing changed from observation 2 — agreed the merge case rests on each rule naming a
real defect with a commit behind it, not on the tables, and I would rather the honest reading stay
as it is.

### Delta self-review (prefix-r1..HEAD)

- size cast into a half dtype before a division: n/a — docs only
- degenerate/empty path validates less or more than the full path: n/a — docs only
- dtype=None resolved before promotion is decided: n/a
- new compile test carries dynamo/compile in its name: n/a — no tests here
- MPS/half skips match the nearest existing test: n/a
- docstring/comment/changelog next to the change still true: checked every claim the two skills make
  against the rebased #4034 tree by running them — the four `--only` names against `SURFACES`, the
  `NEW [<surface>] <id>` format against `render_table`, the `hi` `ValueError`, the empty-sweep
  `ValueError`, exit 0/1/2, `_assert_imports_from`'s name, and `TESTING.md § "Precision and
  Degenerate-Path Helpers"`. Found and fixed in my own delta: step 6 still described `N*` and
  `NEW <id>` with #4034's pre-fix behaviour
- fix mirrored in the 3d/other sibling, and the sibling's accepted calls still pass: the sibling here
  is step 4's probe, which had the identical `PYTHONPATH`-only defect step 2 was reported for; it now
  carries the same guard by reference

### Gate

Docs-only diff, so `verify-delta` maps to no test directory and prints "nothing to verify" — which
step 6 says is not a green gate. Ran by hand instead, on the rebased tree (this branch is #4034 plus
the docs): `pre-commit run --all-files` 13/13, `uvx ty@0.0.9 check kornia testing` clean,
`tests/testing` 74 passed, and the two skill command forms executed for real (`git merge-base`,
`verify-delta -- --base <sha> --tests tests/testing --only "cpu float32"` → exit 0, and the import
guard both ways).

### Merge order

Base is still `feat/testing-precision-helpers`; will retarget to `main` and rebase once #4034 lands.
The `CHANGELOG.md` adjacency is unchanged.

*Posted on behalf of @ducha-aiki by Claude (Opus 5, 1M context).*

ducha-aiki (commented):
Requesting changes because the prescribed review workflow can select the wrong baseline and can misclassify changed failure behavior as pre-existing.

1. **[P1] Resolve the PR’s actual base ref instead of assuming `origin/main`** (`.claude/skills/kornia-review-loop/SKILL.md:30-56,150-166`; `AGENTS.md:84-86`). This PR itself targets `feat/testing-precision-helpers`, so the documented merge-base and gate include the entire parent PR. On stacked or release-target PRs that pulls inherited changes into scope and gates the wrong delta. Resolve/fetch `baseRefName` and use that ref for triage and verification.

2. **[P1] Triage the exact observed failure, not only the pytest node ID** (`.claude/skills/kornia-review-loop/SKILL.md:47-56`). `verify-delta` compares failure-ID sets, not failure causes. If a test fails for reason A at the base but the branch introduces an earlier failure B under the same node ID, it is classified `unchanged`, and the workflow directs the agent not to fix it. Compare the exact reproduction and observed behavior; reserve the failure-set diff for the coarse gate.

3. **[P2] Reuse the active interpreter in manual probes** (`.claude/skills/kornia-review-loop/SKILL.md:62-66,111-117`). Hard-coding `$OLDPWD/.venv/bin/python` fails in the supported `.venv-py312` and `.venv-py313` environments. Capture `sys.executable` before changing directories and account for Windows.

4. **[P2] Make pre-fix tags and probe worktrees collision-safe** (`.claude/skills/kornia-review-loop/SKILL.md:23-28,111-124`). Global names such as `prefix-1` and `../prefix-probe` collide across concurrent reviews; one agent can silently repoint or delete another agent’s baseline. Include the PR/branch and a unique token or SHA, and avoid force-moving a shared tag.

5. **[P2] Require every selected/available gate row to be measured** (`.claude/skills/kornia-review-loop/SKILL.md:150-166`). The workflow treats exit 0 as sufficient, but the current implementation can return 0 with selected surfaces marked `skipped`. Require each selected and available row to be non-skipped, or explicitly resolve and rerun it before pushing.

ducha-aiki:
## Round 2 answered — `3b887806` (rebased onto #4034's `e68a2b32`)

All five findings fixed. Each was reproduced against this branch before anything was written; two
of them reproduced *on this PR itself*, which is worth leading with.

### Triage

| # | Finding | Reproduced | Fix |
|---|---|---|---|
| 1 | step 2 assumes `origin/main` is the base | yes, on this PR — see the measurement below | resolve `gh pr view --json baseRefName` |
| 2 | failure-*id* diffing misreads a changed cause as `unchanged` | yes — `diff_failures({id}, {id})` → `new: []`, `unchanged: [id]`, row `0 / 0 / 1`, regardless of why either side failed | compare the observed failures, not membership |
| 3 | `$OLDPWD/.venv/bin/python` is hard-coded | yes — `pixi.toml:95,101` set `UV_PROJECT_ENVIRONMENT` to `.venv-py312` / `.venv-py313`; neither has a `.venv` | capture `sys.executable` before the `cd` |
| 4 | `prefix-<round>` and `../prefix-probe` are global names | yes, live: `prefix-2` created in one worktree of this repo resolved to the same object from the other | put the branch's short SHA and the checkout's name in them |
| 5 | step 6 treats exit 0 as sufficient | yes — that was true of the tool as well; #4034 fixed the tool, this fixes the reading | require every selected, available row to show numbers |

### 1. Measured on this PR, which is exactly the case you named

```
$ gh pr view 4035 --json baseRefName -q .baseRefName
feat/testing-precision-helpers

$ git merge-base origin/main HEAD                     # what the skill said
9e45dddc   → 23 commits behind HEAD
$ git merge-base origin/feat/testing-precision-helpers HEAD
14b47548   → 10 commits behind HEAD

$ comm -23 <(git diff --name-only 9e45dddc..HEAD | sort) \
           <(git diff --name-only 14b47548..HEAD | sort)
TESTING.md
pixi.toml
testing/__init__.py
testing/precision.py
testing/verify_delta.py
tests/testing/__init__.py
tests/testing/_historical.py
tests/testing/test_precision.py
tests/testing/test_verify_delta.py
```

Nine files belonging to the parent PR. Under the old text every finding against `testing/precision.py`
would have triaged as "this branch broke it", and step 6's gate would have measured a delta nobody
asked for. Step 2 now opens with

```bash
BASE_REF=$(gh pr view --json baseRefName -q .baseRefName)   # e.g. main, or a parent PR's branch
git fetch origin "$BASE_REF"
BASE=$(git merge-base "origin/$BASE_REF" HEAD)
```

and carries the measurement above as its argument, so the rule is not asserted, it is shown.

The `--base origin/main` remark from round 1 needed adjusting rather than deleting: the two steps
still ask different questions, but the difference is *merge-base vs tip of the same ref*, not *base
ref vs main*. Step 6 now says to pass `--base "origin/$BASE_REF"` on any PR that does not target
`main`, and says why the gate wants the tip where triage wanted the merge-base. `AGENTS.md`'s two
summary lines were the sibling copies of both statements and are corrected the same way.

### 2. The one that bites hardest, and the one I would have kept getting wrong

You are right, and it is worse than a technicality because step 2's own instruction depends on it:
the old text said "a finding's test inside `unchanged` is *pre-existing*", full stop. A test that
raised `ImportError` at the base and raises a new `AssertionError` on the branch is one node id on
both sides, so the row reads `0 new` and the workflow says do not fix it — a false triage produced
by the very step that exists to prevent false triage.

Step 2 now says a test in `unchanged` is **not yet** a verdict, and that the verdict requires
reading both failures and confirming they are the same failure — same exception type, same
assertion, same first traceback frame in the library. The set diff is named as the coarse gate it
is. A red-flags row carries the short form ("Same id, new cause, still reads `unchanged`").

I did not try to make `verify-delta` compare causes. Junit carries the message, so it is possible,
but message matching across a whole suite is a new heuristic with its own false positives, and this
PR is documentation. Filed as kornia#4037 rather than built here.

### 3 and 4

`PY=$(python -c 'import sys; print(sys.executable)')`, captured before the `cd`, used by both the
step-2 probe and the step-4 pre-fix probe. That also removes the `(Windows: .venv/Scripts/python.exe.)`
footnote round 1 added — `sys.executable` is already right there — and it matches what
`verify_delta._assert_imports_from` does, which the skill was inconsistent with while citing it.

For #4 the demonstration was in front of me: reviewing these two PRs meant two worktrees of one
repo, and `git tag prefix-2` made in the #4034 worktree resolved to the same object from the #4035
one. Tags are shared by every worktree, so the old `git tag -f prefix-<round>` was a force-move on
a global name — the loser of the race then measures "fails at the pre-fix SHA" against someone
else's branch. Step 1 is now

```bash
PREFIX_TAG="prefix-r<round>-$(git rev-parse --short HEAD)"
git tag "$PREFIX_TAG"        # no -f
```

and step 4's worktree is `"../.$(basename "$PWD")-prefix-probe"`. Dropping `-f` matters on its own:
with the SHA in the name, an existing tag already points where you want, so a force is never the
right answer.

### 5. Rewritten against the tool as #4034 now behaves

The finding was true of both the text and the implementation, so it is answered in two places.
#4034 makes any selected, available surface that produced no measurement exit 2, and splits the old
single `skipped` cell into `not selected` / `unavailable` / `unverified`. Step 6 now says exit 0 is
**necessary and not sufficient**, lists the three cells with what each obliges you to do, and marks
only `unverified` as a hole. The two non-holes still have to be named in the reply — a `--only cpu
float32` run that leaves half precision and MPS unrun is a legitimate green with a gap nobody else
will close, since neither has a CI job. Step 6 also picks up #4034's dirty-checkout refusal, and
step 2 notes it, since the triage probe now runs the same tool.

`kornia-precision-testing` rule 6 was stale for the same reason and is corrected: the helper
compares raw bytes, not `torch.equal`, which reads two NaNs as unequal and `+0.0` as equal to
`-0.0`. Rule 5 picks up the empty-`bad_inputs` refusal.

### Delta self-review (`e6c398a3..HEAD` — the round's pre-fix commit; the step-1 tag `prefix-r2-8968bc9a` names the same tree before the rebase)

- size cast into a half dtype before a division: n/a — documentation only, no code in this PR.
- degenerate/empty path validates less or more than the full path: n/a.
- dtype=None resolved before promotion is decided: n/a.
- new compile test carries dynamo/compile in its name: n/a — no test added.
- MPS/half skips match the nearest existing test: n/a.
- docstring/comment/changelog next to the change still true: three found stale and fixed. The
  `(Windows: .venv/Scripts/python.exe.)` sentence became false the moment the probe used
  `sys.executable`; step 6's `skipped` paragraph described a label #4034 removed;
  `kornia-precision-testing` rule 6 still named `torch.equal` as the comparison. A grep for
  `OLDPWD`, `.venv/bin/python`, `prefix-<round>` and `skipped` across `.claude/skills/` and
  `AGENTS.md` is clean apart from the deliberate "here is what not to do" mentions.
- fix mirrored in the 3d/other sibling, and the sibling's accepted calls still pass: the siblings
  of a skill edit are `AGENTS.md`'s one-line summaries and the cross-links. Both `AGENTS.md` lines
  (84 and 86) were the old claims in miniature and are updated; `kornia-precision-testing`'s
  pointer to step 6 now mentions why exit 0 alone is not the verdict.

### Gate

Documentation only, so `verify-delta` would print "nothing to verify" and exit 0, which step 6
itself says is not a green gate. What I ran instead, on the rebased branch (so #4034's code is
underneath it):

| command | result |
|---|---|
| `pre-commit run --all-files` | 13/13 passed |
| `pytest tests/testing --dtype=float32,float64,float16,bfloat16` | 98 passed |
| `pytest tests/testing --device=mps --dtype=float32` | 85 passed, 1 skipped (the documented kornia#4032 skip) |
| `KORNIA_TEST_OPTIMIZER=inductor pytest tests/testing -k "compile or dynamo"` | 1 passed, 86 deselected |
| step 2's new command, end to end | `BASE_REF=feat/testing-precision-helpers`, fetch, merge-base, scoped diff — all as written |

Both new shell snippets were executed as written before landing, not only read.

This is round 2, so per step 9 the next one should be a full fresh review of the whole branch with
no prior context rather than another incremental pass.

*Posted on behalf of @ducha-aiki by Claude (Opus 5, 1M context).*

ducha-aiki (commented):
I found two blocking workflow defects and two additional correctness issues:

1. [P1] The mandatory pre-fix probe is not executable as written (`.claude/skills/kornia-review-loop/SKILL.md:140-161`). `PROBE_WT` is described but never assigned, while `PY` is assigned only in the optional non-pytest path. The common workflow therefore reaches `"$PY" -m pytest` with an unset command; a stale inherited `PROBE_WT` also makes the force-removal target unsafe. Provide a self-contained block that assigns both variables and uses them consistently for add, run, and guarded cleanup.

2. [P1] The advertised non-pytest triage path never executes the reported REPL snippet or script (`.claude/skills/kornia-review-loop/SKILL.md:84-94`); after the import guard it still runs `pytest <test>`. Existing tests can remain green for newly reported behavior, so this cannot establish branch-versus-merge-base behavior. Show commands that execute the exact script or `python -c` snippet on both trees.

3. [P2] The precision skill says degenerate-path validation is checked exactly, but `assert_degenerate_path_parity` compares exception types only (`.claude/skills/kornia-precision-testing/SKILL.md:45-51`). An intended full-path `ValueError` and an unrelated degenerate-path `ValueError` pass. Require checking messages or otherwise confirming the cause when the same exception class has multiple plausible sources.

4. [P2] The canonical compile example contradicts rule 1 (`.claude/skills/kornia-precision-testing/SKILL.md:69-77`). It hand-picks `[1, 2, 258, 300]` with float32, where every shown size is exactly representable, immediately after instructing users never to hand-pick sizes. Use `unrepresentable_sizes(dtype)` with the dtype under test, or clearly label this as a non-precision compile smoke example and include a correct precision example.

ducha-aiki:
## Review

Scope: 5 files, +361 lines — two new `SKILL.md` files, a `kornia-developer` cross-link, an `AGENTS.md` "Skills" section, and a CHANGELOG line.

Since the diff is instructions rather than executable library code, I checked each factual claim against the PR-A code it describes (`testing/precision.py`, `testing/verify_delta.py`, `conftest.py`, `AGENTS.md` L91, `TESTING.md` §"Precision and Degenerate-Path Helpers", `tests/testing/_historical.py`) and ran the helpers in the project venv.

**Verified correct:** the four `--only` surface names; exit-code semantics 0/1/2; the `not selected` / `unavailable` / `unverified` states; the `N*` partial-baseline row; `WIDEN_PREFIXES`; the `summary.md` path; `_assert_imports_from`; the `#NNNN|issues/NNNN` grep rule; conftest's `dynamo`/`compile` deselection; and the trace-only nature of `_historical.py`. The bfloat16 257/258 rationale and the doctest values reproduce exactly.

Seven findings below. The first, second and fourth are the load-bearing ones — each defeats the purpose of the step it appears in.

### 1. `kornia-review-loop/SKILL.md:62` — Step 2 prescribes two different `--base` values

The prose says to pass the base branch *tip* (or let `origin/main` stand); the step's own code block and its stated thesis say the *merge-base*. On a PR targeting `main` that has moved since branch point, following the prose attributes upstream-introduced failures to the branch — producing exactly the false "pre-existing" triage this step exists to prevent. Pick the merge-base and make the prose match the code block.

### 2. `kornia-review-loop/SKILL.md:88` — the probe captures the wrong interpreter

```
PY=$(python -c 'import sys; print(sys.executable)')
```

From a plain shell this is not the project environment. Measured on this checkout: bare `python` resolves to a miniforge build with torch 2.12.0, while `./.venv/bin/python` has torch 2.9.1. The import guard still passes under either, so the rounding probe silently answers under the wrong torch build with no warning. Route it through `pixi run` / `uv run` as the rest of the repo's instructions do.

### 3. `kornia-precision-testing/SKILL.md:39` — Rule 3 forbids the pattern it just recommended

Rule 3's "never a fixture that `yield`s / never inside the test itself" rules out reusing `_ambient_default_dtype` — the `@contextmanager` cited two lines earlier as *the* pattern, and whose own comment says it exists precisely so the block is not hand-rolled. Narrow the prohibition to hand-rolled save/restore.

### 4. `kornia-review-loop/SKILL.md:145` — `$PROBE_WT` is never assigned

The block uses `cd "$PROBE_WT"`, but nothing sets it. In bash `cd ""` is a silent no-op, so the "pre-fix" probe runs on the fixed branch's tree and reports the post-fix behaviour as the baseline — a silent wrong answer, not an error. Separately, `basename "$PWD"` places the scratch worktree inside the repo when the command is run from a subdirectory.

### 5. `kornia-precision-testing/SKILL.md:57` — Rule 7's `torch_optimizer` advice does not hold

Rule 7 says to use the injected `torch_optimizer` fixture, but `assert_capture_matches_eager(capture="compile")` hardcodes `torch.compile(fullgraph=True, dynamic=True)` and ignores the fixture, so `--optimizer=<backend>` is silently not honoured on that path. Either thread the fixture through the helper or state the limitation in the rule.

### 6. `kornia-precision-testing/SKILL.md:76` — the compile example is a vacuous sweep

The example hand-picks `sizes=[1, 2, 258, 300]`, contradicting rule 1's "never pick one size". Worse, copied onto a float16 test, every one of those sizes is exactly representable, so the sweep passes without exercising anything. This is the exact trap the skill warns about, shipped as its own example — use the generated sweep here.

### 7. `kornia-precision-testing/SKILL.md:17` — unconditional float16 sweep cost

The float16 sweep is `[2049 … 2056]` (measured), roughly 250x the elements of the bfloat16 `[257 … 264]` sweep per call for a 2-D op — times 8 sizes, times eager + re-trace, on a surface `verify-delta` runs on *both* trees. Worth gating or documenting the cost so it is a deliberate choice.

---

ducha-aiki:
Addressed the remaining workflow review findings in `8c9c3dbf`, rebased onto the updated #4034 head `fbd66038`.

- Step 2 now consistently uses the PR merge-base for triage; the base-ref tip is reserved for the final tip-vs-tip gate.
- Non-pytest findings run the exact reported snippet or script on both branch and merge-base worktrees, with import guards on both sides.
- Interpreter capture now comes from the selected Pixi environment's `uv run` Python, so default, py312, and py313 probes use the same interpreter and torch build as the project test task.
- The mandatory pre-fix probe is self-contained: repository root, interpreter, collision-free worktree path, creation state, guarded cleanup, and test command are assigned in one block.
- The precision skill now permits a yielded scoped context manager while rejecting unsafe hand-rolled global-dtype restoration in fixtures.
- Degenerate-path guidance requires cause/message or traceback-frame disambiguation when matching exception classes can come from different guards.
- Compile guidance states that `capture="compile"` hardcodes `torch.compile`; optimizer-specific coverage remains a separate `torch_optimizer` test.
- The compile example uses `unrepresentable_sizes(dtype)` rather than hand-picked float32-exact sizes, and the float16 allocation cost is called out with guidance to use a narrow representative input.
- `AGENTS.md` and the review-loop gate now agree with #4034: dirty checkouts are refused by default, while intentional working-tree runs require `--allow-dirty` plus explicit test targets covering every affected path.

Validation completed after the stacked rebase:

- `pixi run pre-commit-all`: passed.
- `git diff --check`: passed.
- Interpreter capture was exercised in the default and py312 environments.
- Fresh cross-PR review: no remaining actionable findings.

GitHub documentation checks are running on the pushed head; pre-commit.ci is green.

ducha-aiki:
Closing this for now — the skills are not ready.

Six paired with/without evaluations on fifteen replayed review rounds (kornia#4006, #3984, #3942; ACES-style, time-machine origins, Sonnet 5) show the workflow changes the outcome only on findings whose stated diagnosis, condition or set is wrong and the run executes a counterexample search — and even there it acts on what it finds about half the time (D: found 3/3, acted 1/3). On refactor / deletion findings it produces the control arm's diff at 2× the cost, and the v7 gate that stops paying for those mis-routes claim-bearing test changes. The hold-out count never moved from the control's at n=1, and n=1 is below the noise floor for a ~50%-vs-0% effect.

Full record, including the assessment and per-run diffs: https://github.com/kornia/skills-experiments/tree/main/evals/2026-08-26-review-loop-skills (`ASSESSMENT-2026-08-27.md`).

Next: new cases cut from code-fix PRs at ≥3 samples per arm, at this text, before any further revision. Will reopen or resubmit if the numbers justify it. The helpers in #4034 are independent of this PR.

Issue #4030, opened 2026-08-26
create_meshgrid returns a different grid under torch.compile than in eager at float16/bfloat16 (torch.linspace vs its inductor lowering)


ducha-aiki:
### Describe the bug

`create_meshgrid` / `create_meshgrid3d` return different values under `torch.compile` than in eager at `float16`/`bfloat16`, for images wider than 257 (bfloat16) or 2049 (float16). The compiled grid is off by up to **2.0 pixels** from the eager one.

The root cause is `torch.linspace`: its eager kernel computes `step = (end - start) / (steps - 1)` and accumulates it *in the output dtype*, while inductor lowers the same call to an `arange`-style per-index computation. The two round differently, and the eager one is the **less** accurate of the pair:

```
torch.tensor(257.0, dtype=torch.bfloat16)             -> 256.0   # correctly rounded (ties-to-even)
torch.linspace(0, 299, 300, dtype=torch.bfloat16)[257] -> 258.0   # eager: one ulp off
torch.arange(300,          dtype=torch.bfloat16)[257] -> 256.0   # what inductor produces
```

So kornia's grid silently changes when a half-precision model is compiled. This reaches `homography_warp` / `HomographyWarper`, `warp_image_tps`, `undistort_image`, and `camera/stereo.py`, all of which pass `image.dtype` straight into `create_meshgrid`.

This is **pre-existing** — it is not introduced by #4006, which happens to reproduce inductor's existing compiled values byte for byte. It was found while reviewing that PR.

### Reproduction steps

```python
import torch
from kornia.geometry.grid import create_meshgrid

# 1) the torch-level cause: eager linspace vs its inductor lowering
def g(n, dt):
    return torch.linspace(0, n - 1, n, dtype=dt)

c = torch.compile(g, fullgraph=True, dynamic=False)
for dt, n in ((torch.bfloat16, 300), (torch.float16, 3000)):
    e, o = g(n, dt), c(n, dt)
    d = (e.float() - o.float()).abs()
    i = int(d.argmax())
    print(dt, n, "maxdiff", d.max().item(), "| eager", e[i].item(), "compiled", o[i].item())

# torch.bfloat16 300  maxdiff 2.0 | eager 258.0 compiled 256.0
# torch.float16  3000 maxdiff 2.0 | eager 2050.0 compiled 2048.0

# 2) the symptom through kornia's public API
f = torch.compile(create_meshgrid, fullgraph=True, dynamic=False)
for norm in (False, True):
[...]

PR #4036, opened 2026-08-26
fix(geometry): build the meshgrid pixel ramp with arange in eager too

LoganChu:

`create_meshgrid` / `create_meshgrid3d` built the pixel ramp with `torch.linspace` in eager but with
`torch.arange` under capture (the branch added in #4006). The two do not agree at
`float16`/`bfloat16`, so a half-precision grid silently changed the moment the model was wrapped in
`torch.compile`. This builds the ramp with `arange` on both paths.

### Root cause

`torch.linspace` rounds its endpoint into the output dtype *first*, then fills the upper half of the
ramp backwards from that rounded value. Reconstructing ATen's CPU kernel as an executable model and
checking it bitwise against `torch.linspace` across 705 sizes at bfloat16, float16 and float32 gives
0 mismatches:

```
end    = dtype(end)                          # endpoint rounded into the output dtype
step   = (end - start) / dtype(steps - 1)
out[i] = start + step*i           for i <  steps//2   (ascending)
out[i] = end   - step*(steps-1-i) for i >= steps//2   (descending from the rounded end)
```

For `linspace(0, 299, 300, bfloat16)`: `299` is an exact tie between 298 and 300, so ties-to-even
makes `bfloat16(299) = 300`. `halfway = 150`, so index 257 takes the descending branch and yields
`300 - 42 = 258`, where the correctly rounded value is 256. Inductor lowers the same call to a
per-index computation that rounds correctly, giving 256. At bfloat16 n=300 every one of the 128
wrong values sits in the descending half; the ascending half is exact.

```
torch.tensor(257.0, dtype=torch.bfloat16)               -> 256.0   # correctly rounded
torch.linspace(0, 299, 300, dtype=torch.bfloat16)[257]  -> 258.0   # eager
torch.arange(300,          dtype=torch.bfloat16)[257]   -> 256.0   # what inductor produces
```

First divergent width is **258** for bfloat16 and **2050** for float16.

### Impact

The error is one ulp of the image width, so it doubles every binade rather than staying at the
"2 px" quoted for width 300 in the issue:

| width | bf16 ulp | max grid error |
|------:|---------:|---------------:|
|   300 |        2 |         2.0 px |
|   600 |        4 |         4.0 px |
|  1200 |        8 |         8.0 px |
|  2400 |       16 |        16.0 px |

With `normalized_coordinates=True` the same divergence is worth ~2.34 px at bfloat16 w=300 and
~2.93 px at float16 w=3000 once converted back to sampling positions. It reaches `homography_warp` /
`HomographyWarper`, `warp_image_tps`, `undistort_image` and `camera/stereo.py`, all of which pass
`image.dtype` straight into `create_meshgrid`.

This is **pre-existing**, not introduced by #4006 -- checking out `3c1d7265`, where both paths call
`linspace`, reproduces byte-identical values.

### The change

`arange` on both paths. It is a **no-op at float32/float64**: `linspace(0, n-1, n)` and `arange(n)`
are bitwise equal for every size in `range(2, 20001)` at both dtypes, so no non-half result moves.
At half precision it moves the eager values onto the *correctly rounded* side (258 to 256, 2050 to
2048), closing the gap from the accurate direction. It is still an eager numerical change at
float16/bfloat16 above those widths and probably deserves a release note.

The default floating dtype is read as `torch.empty(0).dtype` rather than `torch.get_default_dtype()`
because TorchScript has no builtin for the latter -- with `get_default_dtype()` on the unconditional
path, `TestWarpImage::test_jit` fails with `Unknown builtin op: aten::get_default_dtype`.

### Known remaining gaps

Two things this PR deliberately does **not** close. Both are pre-existing and independent of #4006,
which is why the issue is referenced rather than closed:

1. **`normalized_coordinates=True` still diverges by ~1 ulp.** Eager evaluates
   `(xs / (width - 1) - 0.5) * 2` as three separate half-precision ops, rounding after each;
   inductor fuses the chain, computes it in fp32 and rounds once on store. Residual after this fix:
   0.01171875 (bfloat16, w=300) and 0.001220703125 (float16, w=3000). Confirmed independent of
   #4006 by applying this same `arange` change on top of `3c1d7265`, where both paths share a single
   normalization expression -- it still diverges there (0.0078125 / 0.0009765625). Fixing it likely
   means doing the normalization in fp32 and casting once on both paths; happy to follow up in a
   separate PR if that is the direction you want.

2. **CPU `torch.arange` at bfloat16 is itself wrong above index 4112**, returning 4096 where 4128 is
   correct, first visible at width 4128. So on CPU the eager/compiled gap closes only up to width
   4127 at bfloat16 -- still far better than 257 today (450 of 6000 entries differ at n=6000, versus
   1909 with `linspace`). CUDA `arange` is correctly rounded across all 65536 indices checked, so on
   GPU the fix is complete at every size tested. This looks like an upstream PyTorch CPU-kernel
   issue rather than a kornia one, and is worth a separate torch report.

For context: bfloat16 spacing is already 2 above width 256, so a half-precision pixel grid cannot
give adjacent columns distinct coordinates past that width regardless of which kernel builds it.
Building the ramp in fp32 and casting only at the end of the warp pipeline would remove this whole
class of problem -- a larger change than this PR, but likely the right long-term shape.

Refs #4030 (reported by @ducha-aiki). Referenced rather than `Fixes`, because
`normalized_coordinates=True` still diverges for the separate reason described above, so the stated
expectation in that issue is not yet fully met.

## What did you check?

Reproduced and confirmed fixed on both backends -- CPU/inductor (torch 2.12.0+cpu, Python 3.13) and
CUDA/triton (torch 2.11.0+cu130, RTX 5070 sm_120, triton 3.7.1). Both produce identical numbers
(maxdiff 2.0 at the reported sizes before the change, 0.0 after), and the eager grids are bitwise
identical across CPU and CUDA, so this is purely an eager-vs-compile gap and not a device gap.

```text
$ KORNIA_TEST_OPTIMIZER=inductor pytest tests/geometry/test_grid.py -q
47 passed

$ pytest tests/geometry -q
2386 passed, 60 skipped, 19 xfailed, 5 xpassed

$ pytest tests/geometry/transform/test_homography_warper.py tests/geometry/test_depth.py tests/geometry/transform/test_thin_plate_spline.py tests/geometry/subpix tests/geometry/calibration tests/geometry/camera -q
426 passed, 3 skipped, 1 xfailed

$ pytest tests/contrib/test_conv_distance_transformer.py tests/contrib/test_extract_tensor_patch.py tests/feature/test_mkd.py -q
132 passed, 2 skipped
```

No failures and no errors. The skipped / xfailed / xpassed counts are the suite's existing baseline
on this machine, not anything this change introduces -- running `pytest tests/geometry -q` on an
unmodified checkout of `9e45dddc` gives exactly the same `2386 passed, 60 skipped, 19 xfailed,
5 xpassed`. The skips are optional dependencies and unavailable devices; the xfails are cases the
suite already records as known.

The new test `test_pixel_meshgrid_half_dtype_matches_compile` is a genuine regression test: on the
unpatched tree it fails 4/4 with differences of 1.0 in the pixel ramp, and passes 4/4 with the fix.
It uses widths 258 (bfloat16) and 2050 (float16) -- the first sizes at which the two paths diverge --
and asserts bitwise equality (`atol=0.0, rtol=0.0`). The existing compile test covers only
float32/float64 on a 3x4 image, two dimensions away from the failure in both dtype and size, which
is why this went unnoticed.

### DIFF
--- a/kornia/geometry/grid.py
+++ b/kornia/geometry/grid.py
@@ -67,16 +67,17 @@ def create_meshgrid(
                   [1., 1.]]]])
 
     """
-    # Build the pixel ramp with ``arange`` in both eager and capture. ``linspace`` over
-    # ``(0, size - 1, size)`` and ``arange(size)`` are bitwise equal at float32/float64, but
-    # ``linspace`` rounds its endpoint into the coordinate dtype and fills the upper half of the
-    # ramp backwards from that rounded value, landing one ulp off the correctly rounded column
-    # index at float16/bfloat16 -- from width 258 in bfloat16 and 2050 in float16. Inductor lowers
-    # the same call to a per-index computation that rounds correctly, so eager and compiled grids
-    # disagreed by up to one ulp of the width. ``arange`` also retains the dynamic output length
-    # under export, where ``linspace`` specializes symbolic ``steps``. Match ``linspace``'s default
-    # floating dtype when none is given -- read off an empty tensor rather than via
-    # ``torch.get_default_dtype()``, for which TorchScript has no builtin.
+    # Build the pixel ramp with ``arange`` in both eager and capture. ``linspace`` rounds its
+    # endpoint into the coordinate dtype first and fills the upper half of the ramp backwards from
+    # that rounded value, so its step is exactly 1 only while ``size - 1`` is an exactly
+    # representable integer -- ``size <= 2 ** p + 1`` for a ``p``-bit significand. Past that it
+    # lands one ulp off the correctly rounded column index. The bound is 257 in bfloat16 and 2049
+    # in float16, and 2 ** 24 + 1 / 2 ** 53 + 1 in float32/float64, so no non-half grid can move.
+    # Inductor lowers the same call to a per-index computation that rounds correctly, so eager and
+    # compiled grids disagreed by up to one ulp of the width. ``arange`` also retains the dynamic
+    # output length under export, where ``linspace`` specializes symbolic ``steps``. Match
+    # ``linspace``'s default floating dtype when none is given -- read off an empty tensor rather
+    # than via ``torch.get_default_dtype()``, for which TorchScript has no builtin.
     ramp_dtype = dtype if dtype is not None else torch.empty(0).dtype
     xs = torch.arange(width, device=device, dtype=ramp_dtype)
     ys = torch.arange(height, device=device, dtype=ramp_dtype)
--- a/kornia/geometry/grid.py
+++ b/kornia/geometry/grid.py
@@ -67,15 +67,19 @@ def create_meshgrid(
                   [1., 1.]]]])
 
     """
-    if not torch.jit.is_scripting() and is_compiling():
-        # ``linspace`` specializes symbolic ``steps`` under export; ``arange`` retains the
-        # dynamic output length. Match ``linspace``'s default floating dtype when none is given.
-        arange_dtype = dtype if dtype is not None else torch.get_default_dtype()
-        xs = torch.arange(width, device=device, dtype=arange_dtype)
-        ys = torch.arange(height, device=device, dtype=arange_dtype)
-    else:
-        xs = torch.linspace(0, width - 1, width, device=device, dtype=dtype)
-        ys = torch.linspace(0, height - 1, height, device=device, dtype=dtype)
+    # Build the pixel ramp with ``arange`` in both eager and capture. ``linspace`` over
+    # ``(0, size - 1, size)`` and ``arange(size)`` are bitwise equal at float32/float64, but
+    # ``linspace`` rounds its endpoint into the coordinate dtype and fills the upper half of the
+    # ramp backwards from that rounded value, landing one ulp off the correctly rounded column
+    # index at float16/bfloat16 -- from width 258 in bfloat16 and 2050 in float16. Inductor lowers
+    # the same call to a per-index computation that rounds correctly, so eager and compiled grids
+    # disagreed by up to one ulp of the width. ``arange`` also retains the dynamic output length
+    # under export, where ``linspace`` specializes symbolic ``steps``. Match ``linspace``'s default
+    # floating dtype when none is given -- read off an empty tensor rather than via
+    # ``torch.get_default_dtype()``, for which TorchScript has no builtin.
+    ramp_dtype = dtype if dtype is not None else torch.empty(0).dtype
+    xs = torch.arange(width, device=device, dtype=ramp_dtype)
+    ys = torch.arange(height, device=device, dtype=ramp_dtype)
     # Fix TracerWarning
     # Note: keeping this formula inline avoids the extra tensors and shape checks incurred by
     #       normalize_pixel_coordinates. The two paths use the same singleton-centre policy.
@@ -146,15 +150,12 @@ def create_meshgrid3d(
         grid tensor with shape :math:`(1, D, H, W, 3)`.
 
     """
-    if not torch.jit.is_scripting() and is_compiling():
-        arange_dtype = dtype if dtype is not None else torch.get_default_dtype()
-        xs = torch.arange(width, device=device, dtype=arange_dtype)
-        ys = torch.arange(height, device=device, dtype=arange_dtype)
-        zs = torch.arange(depth, device=device, dtype=arange_dtype)
-    else:
-        xs = torch.linspace(0, width - 1, width, device=device, dtype=dtype)
-        ys = torch.linspace(0, height - 1, height, device=device, dtype=dtype)
-        zs = torch.linspace(0, depth - 1, depth, device=device, dtype=dtype)
+    # See ``create_meshgrid``: ``arange`` keeps eager and captured pixel ramps identical at
+    # float16/bfloat16, and keeps the output length dynamic under export.
+    ramp_dtype = dtype if dtype is not None else torch.empty(0).dtype
+    xs = torch.arange(width, device=device, dtype=ramp_dtype)
+    ys = torch.arange(height, device=device, dtype=ramp_dtype)
+    zs = torch.arange(depth, device=device, dtype=ramp_dtype)
     # Fix TracerWarning
     if normalized_coordinates:
         if not torch.jit.is_scripting() and (torch.jit.is_tracing() or is_compiling()):
--- a/kornia/geometry/grid.py
+++ b/kornia/geometry/grid.py
@@ -77,8 +77,11 @@ def create_meshgrid(
     # compiled grids disagreed by up to one ulp of the width. ``arange`` also retains the dynamic
     # output length under export, where ``linspace`` specializes symbolic ``steps``. Match
     # ``linspace``'s default floating dtype when none is given -- read off an empty tensor rather
-    # than via ``torch.get_default_dtype()``, for which TorchScript has no builtin.
-    ramp_dtype = dtype if dtype is not None else torch.empty(0).dtype
+    # than via ``torch.get_default_dtype()``, for which TorchScript has no builtin. The empty
+    # tensor takes ``device`` so that the read depends on an argument: a no-input factory call is
+    # constant-folded by TorchScript's optimizing executor after the first run, which would freeze
+    # the ramp at whatever the default dtype happened to be on that call.
+    ramp_dtype = dtype if dtype is not None else torch.empty(0, device=device).dtype
     xs = torch.arange(width, device=device, dtype=ramp_dtype)
     ys = torch.arange(height, device=device, dtype=ramp_dtype)
     # Fix TracerWarning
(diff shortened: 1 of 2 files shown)


ducha-aiki (dismissed):
Reviewed by execution, not by reading. Everything load-bearing in this PR reproduces, including the two residual figures you declined to close #4030 on. The fix is right and the regression test is a real one. **One blocker: the PR went `DIRTY` after #3995 landed** — see below, the resolution is mechanical and I have already run it.

Environment for everything below: torch 2.9.1 / Python 3.11 / CPU / macOS arm64, i.e. the version the `dynamo` CI job pins.

## Verified

| claim | result |
|---|---|
| `linspace(0, n-1, n)` ≡ `arange(n)` bitwise at f32/f64 | 0 mismatching sizes in `[2, 20000]`, both dtypes |
| first divergent width | **258** (bf16), **2050** (fp16) — exactly as stated |
| which side is correctly rounded | `bf16(257) = 256`; `linspace[257] = 258`, `arange[257] = 256`. Max deviation from the exact column index: **2.0 for linspace, 1.0 for arange** (the latter is the representation floor) |
| residual at `normalized_coordinates=True` | **0.01171875** (bf16, w=300) and **0.001220703125** (fp16, w=3000) — digit-for-digit what you reported |
| `torch.get_default_dtype()` under TorchScript | genuinely unscriptable (`RuntimeError`); `torch.empty(0).dtype` scripts and returns the default. Your workaround is justified, not defensive |
| new test is a genuine pin | **4/4 fail** on unpatched `main` (`Tensor-likes are not equal`), **4/4 pass** with the change |
| no other movement | `tests/geometry` — 2387 passed / 59 skipped / 18 xfailed / 6 xpassed, **identical to `main`**. At `--dtype=float16,bfloat16`: 696 failures on both trees, **identical failure ID sets** (all pre-existing CPU-half gaps, unrelated to this). `tests/contrib tests/feature/test_mkd.py tests/utils tests/onnx`: 237 passed / 18 skipped. `test_grid.py` under `KORNIA_TEST_OPTIMIZER=inductor`: 47/47 |
| lint / doctest | `ruff check`, `ruff format --check`, `--doctest-modules kornia/geometry/grid.py` all clean |
| the pin actually runs in CI | yes — the `dynamo` job in `pr_test_cpu.yml` runs `tests/geometry -k "dynamo or compile"` under inductor, and `test_pixel_meshgrid_half_dtype_matches_compile` matches. Worth knowing, since without `KORNIA_TEST_OPTIMIZER` it is deselected locally and looks like it never runs |

Overhead of `torch.empty(0).dtype` versus `torch.get_default_dtype()`: **550 ns vs 66 ns**, and only on the `dtype=None` path. Against a 27 µs floor for `create_meshgrid(64, 64)` that is under 2%. No objection.

## Blocker: rebase over #3995

`35555b55` (merged today) rewrote the exact lines you edit — `torch.compiler.is_compiling()` became `is_compiling()` from `kornia.core.utils` for old-torch compatibility, and a `not torch.jit.is_scripting()` guard was added to the `normalized_coordinates` capture branch. Your merge base is `9e45dddc`, so both hunks in `grid.py` now conflict.

The resolution is to take your side in both, unchanged: the ramp branch you delete is the one `#3995` edited, and the `is_compiling` import stays live because the `normalized_coordinates` branch still uses it. I applied exactly that and all the numbers in the table above come from that rebased tree — so nothing here needs re-verifying after you push it.

Worth noting *why* your `torch.empty(0).dtype` choice becomes load-bearing rather than optional after the rebase: on `main` today `torch.get_default_dtype()` survives scripting only because it sits inside a `not torch.jit.is_scripting()` branch, which TorchScript constant-folds away. Moving the ramp onto the unconditional path removes that shelter. Please keep the comment's explanation of this — it is the kind of thing a later cleanup deletes as redundant.

## Two corrections to the PR body

**1. The CPU `arange` bf16 figures are torch-version-specific and worse on the CI torch.** You report the kernel going wrong "above index 4112", first visible at width 4128, with "450 of 6000 entries" differing. On torch 2.9.1 I measure the **first wrong index at 2057** and **658 of 6000** entries differing. Your run was torch 2.12.0, so this is drift in the upstream kernel, not an error on your part — but the sentence "on CPU the eager/compiled gap closes only up to width 4127 at bfloat16" is not true on the version CI actually runs, and the "complete at every size tested" GPU claim needs the same qualifier. The conclusion is untouched: `arange` is still strictly better on the same measurement (658 vs **1909** for `linspace`), and it is correctly rounded where `linspace` is not.

**2. The f32/f64 no-op is analytic, and stating it that way is stronger than the sample.** "Bitwise equal for every size in `range(2, 20001)`" invites the question of what happens at 20001. It never breaks, and the reason is visible in your own reconstruction of the kernel: `step = (end - start) / (steps - 1)` is exactly 1.0 whenever `size - 1` is an exactly representable integer in the ramp dtype, so the descending half evaluates `end - (steps - 1 - i)` = `i` exactly. That holds for `size ≤ 2**p + 1`, where `p` is the mantissa width — **257 for bf16, 2049 for fp16, 2**24 + 1 for f32, 2**53 + 1 for f64**. Your two measured first-divergence widths, 258 and 2050, are precisely that bound plus one, which is what confirms the model. So f32/f64 cannot move for any image size that fits in memory, and the fp16/bf16 boundaries are derived rather than searched. I would put that sentence in the code comment in place of the width figures; a measured maximum in a comment is a number nothing re-executes.

Neither is a change request — fix the wording when you push the rebase.

## Right call on not closing #4030

Referencing rather than closing is correct, and the follow-up you describe (normalize in fp32, cast once, on both paths) is the direction I want. Please open it as a separate PR. The deeper point in your last paragraph is also right and worth keeping on record: bf16 spacing already exceeds 1 above width 256, so a half-precision pixel grid cannot give adjacent columns distinct coordinates at any realistic image width regardless of which kernel builds it. Building the ramp in fp32 and casting at the end of the warp pipeline is the real fix; this PR closes the eager/compiled *divergence*, which is the part that silently changes a user's numbers when they wrap a model in `torch.compile`, and that is worth landing on its own.

Approving on the verification above. It merges as soon as the rebase is pushed — say the word if you would rather I push it for you, `maintainer_can_modify` is on.

---

ducha-aiki (dismissed):
Rebased over `#3995` for you and pushed — head is `e7423f98`, and the PR is `MERGEABLE` again.

Two commits:

1. **Rebase.** Both `grid.py` hunks resolved by taking your side, as described above. `#3995`'s `is_compiling()` swap and its `not torch.jit.is_scripting()` guard on the `normalized_coordinates` capture branch are preserved, and the `is_compiling` import stays live because that branch still uses it.
2. **`docs(geometry): state the arange/linspace bound analytically`** — the change I asked for, done rather than left to you. The comment justified itself with the measured widths 258 and 2050, which leaves the float32/float64 no-op resting on a finite sweep. It now states the rule: `linspace`'s step is exactly 1 only while `size - 1` is an exactly representable integer, i.e. `size <= 2**p + 1` for a `p`-bit significand. That gives **257** and **2049** — your measured first divergences of 258 and 2050 are precisely that bound plus one, which is what confirms the model — and puts the float32/float64 bounds at `2**24 + 1` and `2**53 + 1`, beyond any image that fits in memory. The exact widths stay in `test_pixel_meshgrid_half_dtype_matches_compile`, which re-executes them.

Verified on the pushed head: `KORNIA_TEST_OPTIMIZER=inductor pytest tests/geometry/test_grid.py` 47 passed, `tests/geometry` 2387 passed / 0 failed, `--doctest-modules kornia/geometry/grid.py` 1 passed, `ruff` clean.

The two PR-body corrections from my review still stand and are yours to make whenever — the torch-version qualifier on the CPU `arange` figures, and the same analytic phrasing in the description. Neither blocks the merge.

Re-approving; the earlier approval was dismissed by my push.

---

ducha-aiki (approved):
Final verification round at `9b9c3725`. I pushed one commit to this branch — details below, please shout if you would rather revert it and take it yourself.

**What I found at `e7423f98`.** Replacing the eager `linspace` path with `arange` is right, and the analytic bound in the comment checks out exactly — I swept both halves and `main`'s float32:

```
bfloat16: arange == linspace up to size 257, first diverges at 258
float16 :                              2049,                  2050
float32 : holds through 2 ** 24 + 1, breaks at 2 ** 24 + 2
```

But `torch.empty(0).dtype` regressed TorchScript. It has no graph input, so the optimizing executor folds it away once profiling completes and freezes the ramp at whatever the default dtype was on the **first** call. `linspace(0, size - 1, size)` never folded, because it consumed `size`:

```python
import torch, kornia
s = torch.jit.script(kornia.geometry.create_meshgrid)
s(2, 2)                                    # first call under the default float32
torch.set_default_dtype(torch.float64)
print("scripted:", s(2, 2).dtype, " eager:", kornia.geometry.create_meshgrid(2, 2).dtype)

# main -> scripted: torch.float64   eager: torch.float64
# HEAD -> scripted: torch.float32   eager: torch.float64
```

**The fix.** `torch.empty(0, device=device).dtype`, in both functions. `device` is an argument, so the read keeps a dependency and cannot be folded. Confirmed on the optimized graph itself, not just from behaviour: without the argument `aten::empty` is absent from the executed graph and the dtype has become a constant; with it, the call survives.

Two alternatives I ruled out. `torch.get_default_dtype()` genuinely has no TorchScript builtin (`Unknown builtin op: aten::get_default_dtype`) — the comment was right about that. `torch.arange(float(width))` also tracks the default correctly, but `float(width)` would specialize the SymInt under export, which is the property this PR exists to protect.

**The pin.** `test_pixel_meshgrid_scripted_ramp_follows_the_runtime_default_dtype`, 2d and 3d. It fails on the unfixed code (`assert torch.float32 == torch.float64`) and passes on the fixed one. The name deliberately contains neither "compile" nor "dynamo", because `pytest_collection_modifyitems` deselects tests by those substrings when `KORNIA_TEST_OPTIMIZER` is unset — this is plain TorchScript and should run in the ordinary jobs, not only the `dynamo` one.

**Checks at `9b9c3725`:** `tests/geometry/test_grid.py` 64 passed, 72 with `KORNIA_TEST_OPTIMIZER=inductor` (the 8 compile tests, including the new half-dtype ones, all pass); full `tests/geometry` sweep 2389 passed, 59 skipped, 18 xfailed, 6 xpassed — the same 6 xpassed as `main`; `grid.py` doctests pass; ruff `check` and `format` clean.

Approving.

ducha-aiki (approved):

PR #4039, opened 2026-08-27
fix(tests): stop doctests downloading model weights and failing on progress output (#4005)

### LINKED ISSUE #4005: doctest downloads pretrained models and fails on download progress output
## Problem

`pixi run doctest` executes model examples that download pretrained weights during doctest collection/execution. On a cold cache this makes the nominal documentation check unexpectedly heavyweight, network-dependent, and noisy. The download messages are captured as unexpected doctest output, so several examples fail even when the downloads themselves succeed.

## Reproduction

From the repository root on `main` at `0685b1f5` plus an unrelated local geometry fix:

```bash
pixi run doctest
```

Environment: macOS, Python 3.11.14, torch 2.9.1, CPU, kornia 0.9.0rc1.

The run downloaded multiple checkpoints, including approximately 135 MB for DexiNed, 55.8 MB and 51.7 MB for DeDoDe, 4.17 MB for DISK, 44.2 MB for LoFTR, and additional weights for LightGlue ONNX, MKD, and SOLD2. After 244 seconds it had reached only 41% and was manually interrupted while downloading the 140 MB SOLD2 checkpoint.

Before interruption, pytest had already reported six failures because download status text appeared where the doctests expected no output:

- `kornia.contrib.edge_detection.EdgeDetector`
- `kornia.feature.dedode.dedode.DeDoDe`
- `kornia.feature.disk.disk.DISK`
- `kornia.feature.loftr.loftr.LoFTR`
- `kornia.feature.matching.DescriptorMatcherWithSteerer`
- `kornia.feature.mkd.MKDDescriptor`

Representative failure:

```text
Expected nothing
Got:
    Downloading: "https://huggingface.co/..." to weights/checkpoints/...
```

At interruption the partial summary was:

```text
6 failed, 263 passed, 2 skipped, 1 warning in 244.34s
```

## Expected

The repository doctest task should be deterministic and practical on a cold cache. Model examples that require remote pretrained weights should not download during the general doctest run or fail merely because the downloader reports progress.

Possible directions include skipping/marking network-and-weight examples in the default doctest task, mocking the download layer, using non-pretrained construction in documentation examples, or configuring doctest output handling where a real download is intentionally exercised.

## Scope note

This was discovered while validating an unrelated geometry change. None of the failing doctest modules were modified by that work; this issue reports the test-task behavior separately so the geometry repair does not bundle it.

ducha-aiki:
Closes #4005.

## The two defects

`pixi run doctest` was both network-dependent and *wrong* on a cold cache. Those are separate problems and this PR fixes them separately.

**1. The failures were caused by the output stream, not the download.** Since torch 2.x, `torch.hub.load_state_dict_from_url` writes its status line to **stdout**:

```python
sys.stdout.write(f'Downloading: "{url}" to {cached_file}\n')
```

Only the tqdm bar goes to stderr. doctest captures stdout, so an example that downloads is reported as producing unexpected output:

```
Failed example: disk = DISK.from_pretrained('depth')
Expected nothing
Got:    Downloading: "https://huggingface.co/kornia/disk/..." to weights/checkpoints/depth-save.pth
```

`kornia.core.download.load_state_dict_from_url` is the chokepoint nearly every pretrained model already goes through. Note **where** torch writes that line:

```python
if not os.path.exists(cached_file):
    sys.stdout.write(f'Downloading: "{url}" to {cached_file}\n')
    download_url_to_file(url, cached_file, hash_prefix, progress=progress)
```

Only on a cache miss, immediately before the fetch. So the wrapper fetches a missing file itself and announces it on stderr; torch then finds the file present and has nothing to report. **Nothing process-global is mutated** — see the review thread below for why redirecting `sys.stdout` around the call is not safe here. This is the same approach `kornia.feature.lightglue_onnx.utils.download.download_onnx_from_url` already takes in this repo.

`XFeat.from_pretrained` was the one model calling `torch.hub` directly; it is routed through the wrapper too.

**2. The run downloaded ~838 MB.** I instrumented a full `--doctest-modules kornia/` run and traced every download call: **13 doctests** reach the network — DexiNed (141 MB), ViT-B/16 (343 MB), SOLD2 (146 MB), DeDoDe ×3 (167 MB), LoFTR, DISK, MKD, AffNet + HardNet, YuNet ×2, and the LightGlue ONNX model (23 MB).

## The guard

`testing/doctest_downloads.py`, wired from `conftest.py`, patches the download **primitives** — the functions that run *only* on a cache miss:

| entry point | why it needs its own entry |
|---|---|
| `torch.hub.download_url_to_file` | 12 of the 13 doctests |
| `kornia.feature.lightglue_onnx.utils.download.download_url_to_file` | binds the symbol at import time, so patching `torch.hub` alone misses it |
| `urllib.request.urlretrieve` | what `kornia.onnx.download.CachedDownloader` reaches on a miss |

Because these run only on a miss, the behaviour falls out for free and needs **no list of exempted examples to maintain**:

- weights already cached → the doctest runs for real, exactly as today;
- weights missing → the doctest is skipped, naming the URL and how to opt in.

```
SKIPPED [1] conftest.py:682: doctest needs to download model weights from
https://huggingface.co/kornia/disk/resolve/main/depth-save.pth; set
KORNIA_DOCTEST_DOWNLOAD=1 to allow it, or pre-populate the cache with
`python .github/download-models-weights.py`
```

Opt in with `KORNIA_DOCTEST_DOWNLOAD=1`, `--doctest-download`, or `pixi run doctest-weights`.

## CI

The docs job already restores a weights cache (#3901) — but that made a *cache miss* a red build with a misleading error. Now the cache is a speed optimization rather than a correctness requirement:

- **pull requests** — no downloads. Whatever the restored cache covers runs; the rest skips.
- **post-merge / scheduled run on `main`** — new `download-weights: true` input sets `KORNIA_DOCTEST_DOWNLOAD=1`, so a cache miss downloads instead of skipping and the model examples are always exercised for real somewhere.

## Verification

Measured on the same command, not asserted:

| | before | after |
|---|---|---|
| cold cache, default | 244 s to reach 41 %, 6 failures, interrupted mid-download | **45 s complete, 0 bytes downloaded**, 13 skipped |
| cold cache, downloads on | fails on the progress line | passes, downloads |
| warm cache, downloads on | 653 passed, 4 failed | **653 passed, 4 failed — unchanged** |

I also confirmed the reverse direction: with `kornia/core/download.py` reverted to base, the identical command fails on `Expected nothing / Got: Downloading: ...`, and the new `TestProgressGoesToStderr::test_stdout_stays_clean` fails. The test catches the regression it is meant to catch.

`pixi run pre-commit-all`, `pixi run typecheck`, and `tests/core tests/test_doctest_downloads.py tests/feature/test_xfeat.py` (196 passed) are green.

## Tests

- `tests/core/test_download.py::TestProgressGoesToStderr` — stdout stays clean on a cold cache and silent on a warm one, and the stdout object is never replaced *during* the transfer. These drive the real `torch.hub.load_state_dict_from_url` over a stubbed transfer, so a transfer count of exactly one also proves the prefetch path is the path torch reads.
- `tests/core/test_download.py::TestConcurrentLoadsDoNotDisturbStdout` — overlapping-call and unrelated-thread regressions for the review finding. Both fail against the first revision of this PR, which used `contextlib.redirect_stdout`: overlapping calls restored out of order and left `sys.stdout is sys.stderr` permanently.
- `tests/test_doctest_downloads.py` — env-var parsing, monkeypatch undo, callback contract, and *functional* coverage of each entry point, including the case that matters most: **an already-cached file must not be blocked**.
- `tests/feature/test_xfeat.py::test_pretrained_uses_kornia_downloader` — pins the xfeat routing.

## Not in scope

Four doctests fail on a warm cache for unrelated reasons, filed separately rather than bundled here:

- #4040 — `get_gaussian_erf_kernel1d`, `get_gaussian_kernel3d` and `Se2.log` pin exact tensor reprs that sit on a float representation boundary (the latter two assert opposite exactness properties of each other by accident).
- #4041 — the `DescriptorMatcherWithSteerer` example is the only doctest that moves onto an accelerator, and OOMs on a busy GPU.

Both are visible in the "4 failed" column of the table above, before and after this PR alike.

### DIFF
--- a/kornia/core/download.py
+++ b/kornia/core/download.py
@@ -17,7 +17,7 @@
 
 from __future__ import annotations
 
-import contextlib
+import os
 import sys
 import warnings
 from pathlib import Path
@@ -47,6 +47,59 @@ def hf_url(repo: str, filename: str) -> str:
     return f"{_HF_KORNIA_BASE}/{repo}/resolve/main/{filename}"
 
 
+def _cached_file_path(url: str, kwargs: dict[str, Any]) -> str:
+    """Return the path :func:`torch.hub.load_state_dict_from_url` would cache *url* at.
+
+    Mirrors torch's own resolution: ``<model_dir>/<file_name or basename(url)>``,
+    where ``model_dir`` defaults to ``<torch.hub.get_dir()>/checkpoints``.
+
+    Args:
+        url: the URL that would be downloaded.
+        kwargs: the keyword arguments destined for the torch function; only
+            ``model_dir`` and ``file_name`` are consulted.
+
+    Returns:
+        The absolute path of the cache entry for *url*.
+    """
+    model_dir = kwargs.get("model_dir")
+    if model_dir is None:
+        model_dir = os.path.join(torch.hub.get_dir(), "checkpoints")
+    filename = kwargs.get("file_name") or os.path.basename(urlparse(url).path)
+    return os.path.join(model_dir, filename)
+
+
+def _prefetch_to_cache(url: str, kwargs: dict[str, Any]) -> None:
+    """Download *url* into the torch hub cache if it is not already there.
+
+    Doing the fetch here rather than letting :func:`torch.hub.load_state_dict_from_url`
+    do it means torch finds the file present and never writes its status line to
+    stdout. A no-op when the file is already cached, which is the common case.
+
+    If the path computed here ever disagreed with torch's, the only consequence
+    is that torch downloads the file again -- the result stays correct.
+
+    Args:
+        url: the URL to fetch.
+        kwargs: the keyword arguments destined for the torch function;
+            ``model_dir``, ``file_name``, ``check_hash`` and ``progress`` are
+            honoured so the cache entry is identical to torch's.
+    """
+    cached_file = _cached_file_path(url, kwargs)
+    if os.path.exists(cached_file):
+        return
+
+    os.makedirs(os.path.dirname(cached_file), exist_ok=True)
+
+    hash_prefix = None
+    if kwargs.get("check_hash"):
+        match = torch.hub.HASH_REGEX.search(os.path.basename(cached_file))
+        hash_prefix = match.group(1) if match else None
+
+    # torch writes this to stdout; status output belongs on stderr.
+    sys.stderr.write(f'Downloading: "{url}" to {cached_file}\n')
+    torch.hub.download_url_to_file(url, cached_file, hash_prefix, progress=kwargs.get("progress", True))
+
+
 def load_state_dict_from_url(url: str | list[str], **kwargs: Any) -> dict[str, Any]:
     """Load a state dict from a URL, trying fallback URLs on failure.
 
@@ -62,6 +115,15 @@ def load_state_dict_from_url(url: str | list[str], **kwargs: Any) -> dict[str, A
     corrupts any caller that treats stdout as data -- most visibly doctests,
     where the line is captured as unexpected example output and fails an
     example that downloads on a cold cache.
+
+    That line is reached only when the file is absent from the cache, so this
+    function fetches a missing file itself -- announcing it on stderr -- and
+    leaves torch with nothing to report. Nothing process-global is touched:
+    redirecting :data:`sys.stdout` around the call would divert unrelated
+    threads' output for the whole transfer, and concurrent calls restoring out
+    of order would leave stdout permanently pointing at stderr. This mirrors
+    :func:`kornia.feature.lightglue_onnx.utils.download.download_onnx_from_url`,
+    which already reimplements torch's caching for the same reason.
 
     When multiple URLs are given and ``file_name`` is not already in *kwargs*,
     the basename of the **first** URL is used as the local cache filename for
@@ -102,9 +164,9 @@ def load_state_dict_from_url(url: str | list[str], **kwargs: Any) -> dict[str, A
     last_exc: Exception | None = None
     for i, u in enumerate(urls):
         try:
-            # Keep torch's download chatter off stdout; see the docstring note.
-            with contextlib.redirect_stdout(sys.stderr):
-                return torch.hub.load_state_dict_from_url(u, **kwargs)
+            # Populate the cache ourselves so torch's stdout line is never reached.
+            _prefetch_to_cache(u, kwargs)
+            return torch.hub.load_state_dict_from_url(u, **kwargs)
         except Exception as e:  # noqa: BLE001
             last_exc = e
             if i < len(urls) - 1:
--- a/kornia/core/download.py
+++ b/kornia/core/download.py
@@ -17,6 +17,8 @@
 
 from __future__ import annotations
 
+import contextlib
+import sys
 import warnings
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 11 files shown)


kornia/core/download.py
```diff
@@ -92,7 +102,9 @@ def load_state_dict_from_url(url: str | list[str], **kwargs: Any) -> dict[str, A
     last_exc: Exception | None = None
     for i, u in enumerate(urls):
         try:
-            return torch.hub.load_state_dict_from_url(u, **kwargs)
+            # Keep torch's download chatter off stdout; see the docstring note.
+            with contextlib.redirect_stdout(sys.stderr):
```
ducha-aiki: [P1] Avoid process-global stdout redirection

`contextlib.redirect_stdout` replaces process-global `sys.stdout` for the entire model load, including a potentially long network transfer. Unrelated threads therefore have their stdout sent to stderr; overlapping calls can also restore out of order and leave `sys.stdout` pointing at stderr after both return. I reproduced the unrelated-output case locally, and the added sequential restoration tests do not cover this interleaving. Please use a mechanism scoped to torch’s status write that does not mutate process-global stdout, and add an overlapping-call regression test.

ducha-aiki (commented):
Blocking finding: the downloader wrapper mutates process-global stdout for the full model load. The focused tests pass, but they only exercise sequential restoration and miss concurrent callers. This should be fixed before merge, with an overlapping-call regression test.

Posted on behalf of [@ducha-aiki](https://github.com/ducha-aiki) by Codex (GPT-5.6).

kornia/core/download.py
ducha-aiki: Confirmed both halves before changing anything, and the second one is worse than "can leave stdout pointing at stderr" — it is permanent. Reproduction, two threads overlapping inside the redirect and exiting out of order:

```
CLAIM 2 -- sys.stdout is real stdout after both calls: False
           sys.stdout is sys.stderr                  : True
CLAIM 1 -- unrelated print reached real stdout: []
```

`redirect_stdout` keeps its saved target on the *instance*, so the inner call's exit restores the outer call's already-redirected value, and the process never recovers.

Fixed by removing the redirection entirely rather than scoping it. torch writes that line only when the file is missing from the cache, immediately before fetching it:

```python
if not os.path.exists(cached_file):
    sys.stdout.write(f'Downloading: "{url}" to {cached_file}\n')
    download_url_to_file(url, cached_file, hash_prefix, progress=progress)
```

So the wrapper now fetches a missing file itself and announces it on stderr; torch then finds the file present and has nothing to report. Nothing process-global is touched — no proxy, no lock, no refcount. This is the same approach `kornia.feature.lightglue_onnx.utils.download.download_onnx_from_url` already takes in this repo for the same reason.

`torch.hub` attributes are resolved at call time instead of bound at import, so there is no stale-binding gap between the path the wrapper prefetches to and the path torch reads.

On the regression test: added both cases, `test_overlapping_calls_leave_stdout_intact` (barrier-synchronised, forced out-of-order exit) and `test_unrelated_thread_keeps_its_stdout`. Both, plus `test_stdout_object_is_never_replaced`, **fail against the previous revision and pass against this one** — the last of those asserts stdout is untouched *during* the transfer, not merely restored afterwards.

The stdout tests now drive the real `torch.hub.load_state_dict_from_url` over a stubbed transfer, so a transfer count of exactly one also proves the prefetch path is the path torch looks in; a divergence would show up as a second download. Verified against a real URL too — the DISK doctest on a cold cache downloads once, lands in `weights/checkpoints/depth-save.pth`, and passes.

One thing your finding surfaced separately: the earlier tests were writing into the repo's real `weights/` directory. The fallback tests now get an isolated cache fixture, so they cannot reach the network or touch `weights/`.

Doctest outcomes are unchanged by the rework: cold cache 641 passed / 15 skipped / 0 bytes downloaded, warm cache with `KORNIA_DOCTEST_DOWNLOAD=1` 653 passed.

ducha-aiki (commented):
The previous blocking finding is resolved. The implementation no longer mutates process-global sys.stdout; it prefetches into the same cache entry that torch.hub.load_state_dict_from_url consumes, and the new tests cover both overlapping loads and unrelated-thread output.

I found no new actionable issues in this review round. Focused validation passed: 63 passed, 4 skipped.

Posted on behalf of [@ducha-aiki](https://github.com/ducha-aiki) by Codex (GPT-5.6).

PR #4043, opened 2026-08-27
Fix PyTorch 2.1-2.4 compatibility failures

### LINKED ISSUE #4042: CI: scheduled PyTorch 2.1–2.4 matrix exposes unsupported test paths
## Summary

After #3995 fixed installation of the selected PyTorch matrix version, the scheduled `main` workflow began exercising the actual older PyTorch versions and exposed unsupported test/code paths.

In [the first post-merge scheduled run](https://github.com/kornia/kornia/actions/runs/33097880044), all 14 compatible Ubuntu jobs using PyTorch 2.1.2–2.4.0 failed. The PR workflow was green because [#3995](https://github.com/kornia/kornia/pull/3995) only covered PyTorch 2.5.1 and 2.9.1.

The separately failing Windows/Python 3.12/PyTorch 2.9.1 float64 job appears to be a SAM ONNX export timeout and is not part of this older-version compatibility issue.

## Affected matrix

- PyTorch 2.1.2: Python 3.11, float32/float64
- PyTorch 2.2.2, 2.3.1, and 2.4.0: Python 3.11/3.12, float32/float64

### 1. SigLip2 produces NaN normalized text features on PyTorch 2.1–2.4

Every affected job fails `tests/models/test_siglip2.py::TestSigLip2Model::test_get_text_features`; the feature norms are `tensor([nan, nan])` instead of ones. Example: [Python 3.11 / PyTorch 2.4.0 / float32](https://github.com/kornia/kornia/actions/runs/33097880044/job/98607658557).

The attention-mask conversion in `kornia/models/siglip2/attention.py` is a likely cause: an all-valid input mask is inverted before `scaled_dot_product_attention`, producing fully masked rows. Older PyTorch versions return NaNs for this case, while newer versions avoid the immediate NaN, so the current PR matrix does not catch it.

### 2. Augmentation ONNX tests pass an unsupported `dynamo` argument

`tests/augmentation/test_onnx_export.py` passes `dynamo=False` to `torch.onnx.export`. Older versions fail the exportable and numerical-equivalence parameterizations with:

```text
TypeError: export() got an unexpected keyword argument 'dynamo'
```

The test helper needs version-aware invocation (or equivalent gating) while these versions remain in the supported scheduled matrix.

### 3. CPU autocast tests hit unsupported bfloat16 operations

The 3D augmentation autocast tests fail on older PyTorch with:

```text
RuntimeError: "eye" not implemented for 'BFloat16'
```

The affected tests include `TestIntensityAugmentationBase3D::test_autocast` and `TestGeometricAugmentationBase3D::test_autocast`.

### 4. PyTorch 2.2–2.4 cannot run the export/compile tests on Python 3.12

The Python 3.12 legs fail singleton-boundary export/compile tests in `tests/geometry/test_conversions.py` and `tests/geometry/test_grid.py` with:

```text
RuntimeError: Python 3.12+ not yet supported for torch.compile
```

The existing version guard covers only part of this surface and should be applied consistently.

## Expected outcome

- Scheduled Ubuntu jobs pass for every declared PyTorch/Python/dtype combination.
- Tests using APIs or backend capabilities introduced after the selected PyTorch version are adapted or skipped with explicit version guards.
- SigLip2 attention-mask handling is corrected rather than skipped, since the NaN indicates a real fully-masked-attention bug.
- Add a minimal PR check for the oldest supported PyTorch version, or otherwise keep scheduled-only compatibility failures visible and actionable.

ducha-aiki:

Restores the declared PyTorch 2.1-2.4 compatibility paths exposed by the scheduled matrix:

- **fixes an inverted SigLip2 attention mask on every PyTorch version.** All three `attention_mask` branches of `SigLip2Attention` negated the mask before handing it to `scaled_dot_product_attention`, whose boolean form reads `True` as *attend*. Every position the caller asked to attend to was masked out, so an all-ones mask left every query row fully masked. Only the symptom is version-dependent: a fully masked row returns `0` from SDPA on PyTorch 2.5+ and `nan` on PyTorch <=2.4, so `get_text_features(..., attention_mask=...)` has been returning a **zeroed** attention output on current PyTorch and `nan` text features on the declared floor. 2-D padding masks are now broadcast over keys only, matching `SiglipTextTransformer.create_bidirectional_mask` in HF `transformers`, so a partially padded sequence no longer produces a fully masked row
- replaces `Tensor.any(dim=(-2, -1))` and `Tensor.all(dim=(-1, -2))` in `find_essential` with flattened reductions; multi-dimension `any`/`all` only landed in PyTorch 2.2, so the previous form raised `TypeError` on the declared `torch>=2.0.0` floor
- replaces the PEP 604 `torch.Tensor | None` annotation in `kornia/color/utils.py` with `Optional[...]`, so `torch.jit.script(kornia.color.rgb_to_yuv)` compiles again on the floor
- gates the legacy ONNX exporter argument by PyTorch version and skips the opset-20 suite where that opset is unsupported
- skips the two CPU 3-D autocast cases below PyTorch 2.3, where `eye` has no bfloat16 kernel
- adds `dynamic_export_is_available()` for the three `torch.export` dynamic-shape tests, since `torch.export.Dim` and `dynamic_shapes` only exist from PyTorch 2.2
- narrows the Dynamo availability guard to Python 3.12 / PyTorch <2.4 (the Python 3.12 restriction was lifted in 2.4, not 2.5) and applies it to the two unguarded compile tests
- narrows two float16 pins to PyTorch 2.2 on CPU, where the `bmm` and `atan2` half kernels landed: the three float16-transform cases of `TestTransformPoints::test_mixed_dtypes`, and the `cart2pol` half of `test_wart_float16_underflowed_default_eps_flips_branches` (the `convert_points_from_homogeneous` claim in that pin stays live on every version)

Closes #4042.

## What did you check?

- `pixi run pre-commit-all`
- `pixi run typecheck` (passes with one pre-existing deprecation diagnostic)
- 75 SigLip2 and augmentation-base tests on PyTorch 2.9.1
- 130 `find_essential` tests, plus an explicit `torch.jit.script` check on both changed helpers
- 24 geometry export/compile tests with Inductor
- 27 ONNX export/numerical tests on PyTorch 2.9.1 and PyTorch 2.4.0
- `tests/color/` + `tests/geometry/epipolar/` at `--dtype=float32,float64`: `1100 passed, 3 skipped, 8 xpassed`, identical on PyTorch 2.1.2 and 2.9.1
- the five files carrying the remaining 2.1.2 failures (`test_linalg`, `test_conversions`, `test_grid`, `test_yuv`, `test_essential`) at `--dtype=float32,float64`: `960 passed` on PyTorch 2.9.1 and `941 passed, 19 skipped` on PyTorch 2.1.2 — the difference is exactly the version guards, with no failures on either end
- each version boundary above cross-checked against the per-version job logs of scheduled run `33097880044`, so no boundary skips a case that already passes; the `bmm`/`atan2` boundary was additionally measured directly on PyTorch 2.1.2 (raises) and 2.2.2 (passes), and the SDPA fully-masked-row boundary on 2.1.2/2.2.2 (`nan`), 2.5.1 and 2.9.1 (`0`)

## How did AI help?

Codex investigated the scheduled failures, implemented the compatibility fixes and regression tests, reproduced the behavior in isolated PyTorch 2.1.2 and 2.4.0 environments, and ran an independent review before opening this PR. Claude then reviewed the result against the scheduled-run logs, corrected four version boundaries, dropped the proposed PyTorch 2.1.2 PR leg (that version is old enough that the scheduled matrix is the right place for it), found the two remaining `torch<2.2` library gaps in a real 2.1.2 environment, and narrowed the last two float16 pins.

Posted on behalf of [@ducha-aiki](https://github.com/ducha-aiki) by Codex (GPT-5) and Claude (Opus 5).

### DIFF
--- a/.github/workflows/pr_test_cpu.yml
+++ b/.github/workflows/pr_test_cpu.yml
@@ -105,12 +105,6 @@ jobs:
           - os: windows-latest
             pytorch-version: "2.5.1"
         include:
-          # Keep the oldest supported PyTorch version visible on every PR without
-          # duplicating the full scheduled compatibility matrix.
-          - os: ubuntu-latest
-            pytorch-dtype: "float32"
-            python-version: "3.11"
-            pytorch-version: "2.1.2"
           - os: windows-2022
             pytorch-dtype: "float32"
             python-version: "3.11"
--- a/.github/workflows/pr_test_cpu.yml
+++ b/.github/workflows/pr_test_cpu.yml
@@ -105,6 +105,12 @@ jobs:
           - os: windows-latest
             pytorch-version: "2.5.1"
         include:
+          # Keep the oldest supported PyTorch version visible on every PR without
+          # duplicating the full scheduled compatibility matrix.
+          - os: ubuntu-latest
+            pytorch-dtype: "float32"
+            python-version: "3.11"
+            pytorch-version: "2.1.2"
           - os: windows-2022
             pytorch-dtype: "float32"
             python-version: "3.11"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   partially padded sequence no longer produces a fully masked row. A row whose mask is entirely zero — an empty
   caption in a batch — is still fully masked and still follows the SDPA behavior above.
 
+* Fix `find_essential` and `run_5point` raising `TypeError: all() received an invalid combination of arguments` on
+  PyTorch before 2.2 (#4043). `run_5point`, `_solve_2x2_tikhonov_safe` and the scripted Nister helper each reduced
+  over two axes at once with `Tensor.any(dim=(-2, -1))` / `Tensor.all(dim=(-1, -2))`, and multi-dimension `any`/`all`
+  only landed in PyTorch 2.2, so the whole five-point solver was unusable on the declared `torch>=2.0` floor. The
+  reductions now flatten the two trailing axes first, which is equivalent on every version.
+
+* Fix `torch.jit.script` of `rgb_to_yuv`, `yuv_to_rgb`, `rgb_to_xyz` and `xyz_to_rgb` on older PyTorch (#4043). Their
+  shared `kornia.color.utils._apply_linear_transformation` helper annotated its optional argument with the PEP 604
+  `torch.Tensor | None` form, which the TorchScript compiler on the declared floor does not accept (reproduced on
+  PyTorch 2.1.2, while PyTorch 2.9.1 compiles it), so scripting any of the four conversions failed. The annotation is
+  `Optional[torch.Tensor]` now, which every supported version accepts.
+
 * Define singleton pixel axes at normalized center, reject non-positive normalization sizes, preserve empty warp
   destinations, and deprecate the now-unused `eps` normalization parameters (#4006). The four
   `{de,}normalize_pixel_coordinates{,3d}` helpers and `normal_transform_pixel{,3d}` now derive their scale from the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* Fix the inverted `SigLip2` attention-mask polarity, so every call that passes an `attention_mask` changes (#4043).
+  All three mask branches of `SigLip2Attention` (2-D `(B, N)`, 3-D `(B, N, N)` and 4-D `(B, 1, N, N)`) negated the
+  mask before handing it to `torch.nn.functional.scaled_dot_product_attention`, whose boolean form reads `True` as
+  *attend*, not as *mask out*. Every position the caller asked to attend to was masked out and every padded position
+  attended to, so an all-ones mask left every query row fully masked. The symptom is version-dependent because a
+  fully masked row returns `0` from SDPA on PyTorch 2.5 and later and `nan` on PyTorch 2.4 and earlier:
+  `SigLip2.get_text_features(..., attention_mask=...)` has been returning a zeroed attention output on current
+  PyTorch and `nan` text features on the declared `torch>=2.0` floor. 2-D padding masks are now broadcast over the
+  key axis only, matching `SiglipTextTransformer.create_bidirectional_mask` in Hugging Face `transformers`, so a
+  partially padded sequence no longer produces a fully masked row. A row whose mask is entirely zero — an empty
+  caption in a batch — is still fully masked and still follows the SDPA behavior above.
+
 * Define singleton pixel axes at normalized center, reject non-positive normalization sizes, preserve empty warp
   destinations, and deprecate the now-unused `eps` normalization parameters (#4006). The four
   `{de,}normalize_pixel_coordinates{,3d}` helpers and `normal_transform_pixel{,3d}` now derive their scale from the
--- a/kornia/color/utils.py
+++ b/kornia/color/utils.py
@@ -15,12 +15,14 @@
 # limitations under the License.
 #
 
+from typing import Optional
+
 import torch
 from torch.nn import functional as F
 
 
 def _apply_linear_transformation(
-    image: torch.Tensor, kernel: torch.Tensor, bias: torch.Tensor | None = None
+    image: torch.Tensor, kernel: torch.Tensor, bias: Optional[torch.Tensor] = None
 ) -> torch.Tensor:
     """Apply a 3x3 linear color transformation with device-aware optimization.
 
--- a/kornia/geometry/epipolar/essential.py
+++ b/kornia/geometry/epipolar/essential.py
@@ -72,7 +72,7 @@ def run_5point(points1: torch.Tensor, points2: torch.Tensor, weights: Optional[t
     X = torch.cat([x1 * x2, x1 * y2, x1, y1 * x2, y1 * y2, y1, x2, y2, ones], dim=-1)
     # use Nister's 5PC to solve essential matrix
     E = null_to_Nister_solution(X, batch_size)
-    bad = torch.isnan(E).all(dim=(-1, -2)).all(dim=-1)  # (B,)
+    bad = torch.isnan(E).flatten(-2).all(-1).all(-1)  # (B,)
     if bad.any():
         eye3 = torch.eye(3, device=E.device, dtype=E.dtype).view(1, 1, 3, 3).expand(batch_size, 10, 3, 3)
         E = torch.where(bad.view(batch_size, 1, 1, 1), eye3, E)
--- a/kornia/geometry/epipolar/essential.py
+++ b/kornia/geometry/epipolar/essential.py
@@ -180,7 +180,7 @@ def _solve_2x2_tikhonov_safe(A: torch.Tensor, b: torch.Tensor, eps: float = 1e-1
     x = torch.where(bad.unsqueeze(-1).unsqueeze(-1), x_fb, x_dir)
 
     # if still non-finite, mark bad and zero it
-    nonfinite = torch.isnan(x).any(dim=(-2, -1)) | torch.isinf(x).any(dim=(-2, -1))
+    nonfinite = torch.isnan(x).flatten(-2).any(-1) | torch.isinf(x).flatten(-2).any(-1)
     bad = bad | nonfinite
     x = torch.where(bad.unsqueeze(-1).unsqueeze(-1), torch.zeros_like(x), x)
 
@@ -275,7 +275,7 @@ def _null_to_Nister_solution_script(
     eliminated = torch.linalg.solve(A10, b_poly)  # (B,10,10)
 
     # Detect NaN/Inf from singular solve and fix with damping solve
-    bad = torch.isnan(eliminated).any(dim=(-2, -1)) | torch.isinf(eliminated).any(dim=(-2, -1))  # (B,)
+    bad = torch.isnan(eliminated).flatten(-2).any(-1) | torch.isinf(eliminated).flatten(-2).any(-1)  # (B,)
     if bad.any():
         # Damped solve only for bad rows but WITHOUT compaction:
         # build damped A = A10 + λI, where λ depends on scale
--- a/kornia/models/siglip2/attention.py
+++ b/kornia/models/siglip2/attention.py
@@ -112,19 +112,17 @@ class SigLip2Attention(nn.Module):
         if attention_mask is not None:
             # Handle different mask formats
             if attention_mask.dim() == 2:
-                # (batch_size, seq_len) -> (batch_size, 1, seq_len, seq_len)
-                # Create 2D mask: both query and key positions must be valid
-                # Convert to boolean: 1 = attend (False = don't mask), 0 = mask (True = mask out)
-                mask_bool = attention_mask.bool()
-                # Expand to (batch_size, 1, seq_len, seq_len) where True means mask out
-                # We need: if either query or key is masked, then mask that attention
-                attn_mask = ~(mask_bool.unsqueeze(1).unsqueeze(2) & mask_bool.unsqueeze(1).unsqueeze(3))
+                # (batch_size, seq_len) -> (batch_size, 1, 1, seq_len)
+                # Mask keys only so padded query positions do not create fully masked
+                # rows, which produce NaNs in older SDPA implementations.
+                # Boolean SDPA masks use True = attend and False = mask out.
+                attn_mask = attention_mask.bool().unsqueeze(1).unsqueeze(2)
             elif attention_mask.dim() == 3:
                 # (batch_size, seq_len, seq_len) -> (batch_size, 1, seq_len, seq_len)
-                attn_mask = ~attention_mask.bool().unsqueeze(1)
+                attn_mask = attention_mask.bool().unsqueeze(1)
             elif attention_mask.dim() == 4:
                 # Already in correct format (batch_size, 1, seq_len, seq_len)
-                attn_mask = ~attention_mask.bool()
+                attn_mask = attention_mask.bool()
             else:
                 raise ValueError(f"Unsupported attention_mask dimension: {attention_mask.dim()}")
 
--- a/kornia/models/siglip2/attention.py
+++ b/kornia/models/siglip2/attention.py
@@ -82,10 +82,20 @@ class SigLip2Attention(nn.Module):
 
         Args:
             hidden_states: Input tensor of shape (batch_size, seq_len, hidden_size).
-            attention_mask: Optional attention mask. Can be:
-                - (batch_size, seq_len): 1D mask where 1 = attend, 0 = mask
-                - (batch_size, seq_len, seq_len): 2D mask where 1 = attend, 0 = mask
-                - (batch_size, 1, seq_len, seq_len): 4D mask (broadcastable)
+            attention_mask: Optional attention mask, where ``1``/``True`` means attend and ``0``/``False``
+                means mask out. Can be:
+
+                - (batch_size, seq_len): a padding mask, broadcast over the key axis only, matching
+                  ``SiglipTextTransformer.create_bidirectional_mask`` in Hugging Face ``transformers``.
+                  A padded query row is therefore *not* masked and its output is finite but meaningless;
+                  callers are expected to discard those positions.
+                - (batch_size, seq_len, seq_len): a per-query mask, broadcast over the head axis.
+                - (batch_size, 1, seq_len, seq_len): already in the layout
+                  ``torch.nn.functional.scaled_dot_product_attention`` expects.
+
+                A query row whose mask is entirely ``0`` follows the behavior of
+                ``scaled_dot_product_attention`` for a fully masked row: ``0`` on PyTorch 2.5 and later,
+                ``nan`` before it.
 
         Returns:
             Output tensor of shape (batch_size, seq_len, hidden_size).
@@ -121,7 +131,7 @@ class SigLip2Attention(nn.Module):
                 # (batch_size, seq_len, seq_len) -> (batch_size, 1, seq_len, seq_len)
                 attn_mask = attention_mask.bool().unsqueeze(1)
             elif attention_mask.dim() == 4:
-                # Already in correct format (batch_size, 1, seq_len, seq_len)
+                # Already in the SDPA layout (batch_size, 1, seq_len, seq_len); True = attend
                 attn_mask = attention_mask.bool()
             else:
                 raise ValueError(f"Unsupported attention_mask dimension: {attention_mask.dim()}")
--- a/testing/base.py
+++ b/testing/base.py
@@ -33,15 +33,26 @@ from kornia.core._compat import torch_version_lt
 Dtype = Union[torch.dtype, None]
 Tensor = torch.Tensor
 
-DYNAMO_UNAVAILABLE_REASON = "torch.compile requires torch>=2.5 on Python>=3.12 and torch>=2.6 on Python>=3.13"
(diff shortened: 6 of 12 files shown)


ducha-aiki:
## Review

Reviewed the branch in a worktree and verified every version boundary against a **real torch 2.1.2 environment** with this branch checked out, plus a benchmark of the `find_essential` rewrite on torch 2.9.1.

The SigLip2 polarity fix, the ONNX/autocast/Dynamo boundaries and the new `dynamic_export_is_available()` guard all check out. Two blockers below: the `any(dim=<tuple>)` sweep missed a site, and six further 2.1.2 failures remain, so `Closes #4042` does not hold yet.

---

### 🔴 1. `kornia/geometry/epipolar/essential.py:75` — the compat fix is incomplete

The PR replaces `any(dim=(-2, -1))` at lines 183 and 278, but `run_5point` still contains:

```python
bad = torch.isnan(E).all(dim=(-1, -2)).all(dim=-1)  # (B,)
```

Same unsupported multi-dimension reduction, on the unconditional path immediately after `null_to_Nister_solution` — lines 183/278 simply raised first, which is why it was shadowed in the scheduled-run log.

Verified live on torch 2.1.2 **with this branch checked out**:

```
File ".../kornia/geometry/epipolar/essential.py", line 75, in run_5point
    bad = torch.isnan(E).all(dim=(-1, -2)).all(dim=-1)
TypeError: all() received an invalid combination of arguments - got (dim=tuple, ),
  but expected one of: * () * (int dim, bool keepdim) * (name dim, bool keepdim)
```

So `find_essential` is still broken on `torch<2.2` after this PR, and the `TestFindEssential` cases that fail today on the 2.1.2 leg will keep failing. Fix:

```python
bad = torch.isnan(E).flatten(-2).all(-1).all(-1)  # (B,)
```

Verified: `find_essential` runs on 2.1.2, byte-identical output on 2.9.1 for the same seed, and `tests/geometry/epipolar/test_essential.py` gives `124 passed, 6 xpassed` on float32+float64.

`grep -rnE "\.(any|all)\(dim=\(" kornia/` finds exactly three sites — these three. (`kornia/filters/in_range.py:143` is `all(dim=(1))`, an int in parens, not a tuple.) Worth adding that grep to the PR checks so the class is closed rather than sampled — note that no test on modern PyTorch can catch any of it, which is precisely how line 75 survived a 130-test `find_essential` run on 2.9.1.

### 🔴 2. `Closes #4042` is premature — six further failures remain on 2.1.2

Even with #1 fixed, on torch 2.1.2 with this branch:

```
tests/color/test_yuv.py::TestRgbToYuv::test_jit[cpu-float32]              RuntimeError
tests/color/test_yuv.py::TestRgbToYuv::test_jit[cpu-float64]              RuntimeError
tests/geometry/test_linalg.py::TestTransformPoints::test_mixed_dtypes     "bmm" not implemented for 'Half'   (x3)
tests/geometry/test_conversions.py::test_wart_float16_underflowed_default_eps_flips_branches
                                                                          "atan2_cpu" not implemented for 'Half'
6 failed, 65 passed
```

The YUV pair is **library code, not a test** — `kornia/color/utils.py:23`:

```python
def _apply_linear_transformation(
    image: torch.Tensor, kernel: torch.Tensor, bias: torch.Tensor | None = None
                                               ~~~~~~~~~~~~~~~~~~~ <--- HERE
) -> torch.Tensor:
```

> `RuntimeError: Expression of type | cannot be used in a type expression`

PEP 604 inside a TorchScript-compiled helper, so `torch.jit.script(kornia.color.rgb_to_yuv)` is broken on the declared `torch>=2.0` floor. Pre-existing and untouched here, but it is a one-word fix (`Optional[torch.Tensor]`), it is what `AGENTS.md` already requires of JIT modules, and it sits squarely inside this PR's stated goal of restoring the declared 2.1–2.4 paths.

The remaining four are genuine missing CPU `Half` kernels in old PyTorch and want test-level skips — a separate call about whether the 2.1 leg should be green or narrowed.

Since #4042's acceptance criterion is "scheduled Ubuntu jobs pass for every declared combination", either land the skips too or retitle to `Refs #4042`.

### 🟡 3. The SigLip2 fix is a larger correctness fix than the description says

The description frames this as older-SDPA NaNs. The polarity was in fact inverted on **all three** mask branches on **all** PyTorch versions; only the symptom is version-dependent:

```
torch 2.9.1: OLD mask, all-ones -> finite=True,  all-zero=True   # silently wrong
torch 2.1.2: OLD mask, all-ones -> finite=False                  # NaN, the CI failure
```

A fully-masked row returns `0` on modern SDPA and `NaN` on 2.1.2. So on current PyTorch every `get_text_features(..., attention_mask=...)` call with an all-ones mask has been returning a **zeroed** attention output. That is a silent numerical bug fixed for all users, not a 2.1–2.4 shim, and it changes SigLip2 text embeddings for anyone on torch 2.9 — it should be called out in the PR body and the changelog.

The fix itself is correct, and the key-only broadcast matches HF's `create_bidirectional_mask` in `SiglipTextTransformer`.

### 🟡 4. Narrow residual: an all-zero mask row

Masking keys only removes fully-masked rows for *partially* padded sequences, but an all-zero mask row — an empty caption in a batch — still produces one, reproducing exactly the divergence this PR set out to remove:

```
torch 2.9.1: all-zero mask row -> finite=True, all-zero=True
torch 2.1.2: all-zero mask row -> finite=False
```

Low priority: either document it or force one attendable key per row.

### ⚪ Considered and rejected: the `[:, -1]` text pooling

`kornia/models/siglip2/text_encoder.py:249` pools `last_hidden_state[:, -1]`, so when the final position is padding the pooled feature comes from a padded query row, and `get_text_features` is not padding-invariant. This is **not** a defect — it matches the reference exactly. HF `transformers` 5.5.4, `SiglipTextTransformer.forward`, verbatim:

```python
# The model uses the last token's hidden state, which may be padding.
pooled_output = last_hidden_state[:, -1, :]
```

HF documents the behavior as intentional; pooling the last *valid* index instead would diverge from the published checkpoints. Recording it so it does not get "fixed" later.

---

### Benchmark: `any(dim=(-2,-1))` → `.flatten(-2).any(-1)`

Checking the rewrite costs nothing on modern PyTorch. torch 2.9.1, CPU, float64, µs/call, two independent runs (min-of-5 / min-of-7):

| tensor | `any(dim=(-2,-1))` | `.flatten(-2).any(-1)` | ratio |
|---|---:|---:|---:|
| `x (1, 2, 1)` — tikhonov | 5.45 | 6.88 | 1.26x |
| `x (1000, 2, 1)` | 12.74 | 15.16 | 1.19x |
| `x (10000, 2, 1)` | 46.91 | 48.62 | 1.04x |
| `x (100000, 2, 1)` | 933.00 | 892.61 | 0.96x |
| `E (1, 10, 10)` — nister | 5.85 | 7.29 | 1.25x |
| `E (1000, 10, 10)` | 470.58 | 488.80 | 1.04x |
| `E (10000, 10, 10)` | 1482.50 | 1541.90 | 1.04x |

The cost is a fixed **~1.4 µs** of dispatch for the extra op, not extra work — `flatten(-2)` on a contiguous tensor is a view. It reads as 1.2–1.3x only where the whole call is under 7 µs and decays into noise once the reduction dominates.

In context: `find_essential(B=32, N=16)` costs **3.18 ms/call**, so the two rewritten lines add **~0.005 ms, or 0.157% of one call**, and both sit behind `torch.where` / `torch.linalg.solve` on `(B,10,10)` tensors. Keep the flattened form.

### ✅ Verified sound

- **Opset-20 boundary at 2.4** — `ONNX_MAX_OPSET` is 19 in v2.1.2 / v2.2.2 / v2.3.1 and 20 in v2.4.0, so the module-level skip is correctly scoped rather than over-broad.
- **bfloat16 `eye` boundary at 2.3** and **Dynamo / Python 3.12 boundary at 2.4** both match the per-version job logs.
- **`dynamic_export_is_available()`** is necessary and correctly written: on 2.1.2 `torch.export` **exists** but `torch.export.Dim` does **not**, so a plain `hasattr(torch, "export")` would have passed and failed later; the `getattr(torch, "export", None)` form also handles 2.0.
- `torch_version_ge(2, 5, 0)` is the right gate for the `dynamo=` kwarg; `pytestmark` at `tests/augmentation/test_onnx_export.py:59` is the file's only one, no collision; `Any` is imported.
- `torch.jit.script` still compiles `_solve_2x2_tikhonov_safe` and `_null_to_Nister_solution_script`.
- 114 passed on torch 2.9.1 across `tests/models/test_siglip2.py`, `tests/geometry/test_grid.py`, `tests/augmentation/test_base.py`.
- `grep -rnE "#4042|issues/4042" kornia/ tests/` → no hits, so there is no stale inline prose to retire.

Nit: the blank-line deletions at `tests/augmentation/test_base.py:184,204` are unrelated noise in an otherwise tight diff.

---

ducha-aiki:
Pushed `c7513a76` fixing findings 1 and 2 from the review above.

**`kornia/geometry/epipolar/essential.py:75`** — `torch.isnan(E).all(dim=(-1, -2)).all(dim=-1)` → `torch.isnan(E).flatten(-2).all(-1).all(-1)`, matching the two sites already rewritten in this PR. `find_essential` now runs on torch 2.1.2.

**`kornia/color/utils.py:23`** — `bias: torch.Tensor | None` → `bias: Optional[torch.Tensor]`, so `torch.jit.script(kornia.color.rgb_to_yuv)` works again on the declared `torch>=2.0.0` floor. Confirmed this was the only PEP 604 annotation in `kornia/` outside a module carrying `from __future__ import annotations`:

```
for f in $(grep -rlE "(Tensor|int|float|str|bool|Module) \| None|\| None =" kornia/ --include=*.py); do
    grep -q "from __future__ import annotations" "$f" || echo "$f"
done
# -> kornia/color/utils.py
```

**Verification** — `tests/color/` + `tests/geometry/epipolar/`, `--device=cpu --dtype=float32,float64`, identical on both ends of the matrix:

| | result |
|---|---|
| torch 2.1.2 | `1100 passed, 3 skipped, 8 xpassed` |
| torch 2.9.1 | `1100 passed, 3 skipped, 8 xpassed` |

`pixi run pre-commit-all` passes. `pixi run typecheck` reports only the pre-existing `custom_fwd` deprecation already noted in the PR description.

---

### Still open

- **The four `Half` failures** (`"bmm" not implemented for 'Half'` ×3 in `TestTransformPoints::test_mixed_dtypes`, `"atan2_cpu" not implemented for 'Half'` in `test_wart_float16_underflowed_default_eps_flips_branches`). These are genuine missing CPU kernels in old PyTorch, so they want test-level skips rather than a library change — that is a call about whether the 2.1 leg should be green or narrowed, which I have left to you. Until they are addressed, `Closes #4042` still overstates the result; `Refs #4042` would be accurate.
- **Findings 3 and 4** — the SigLip2 severity note for the PR body/changelog, and the all-zero mask row.

### Unrelated flake spotted while verifying

`tests/color/test_yuv.py::TestRgbToYuv::test_round_trip_rgb_yuv_rgb[cpu-float32]` is flaky on unseeded random input, independent of this branch — **2/10 failures on the pristine branch, 1/10 with the fixes applied**, with a greatest relative difference of ~0.033 against a 0.001 tolerance. Pre-existing and out of scope here, but worth its own issue.

---

ducha-aiki:
Pushed `8702df03`, closing the two items left open after `c7513a76`.

### Narrowing the 2.1.2 leg

Both remaining failures are missing CPU float16 kernels in old PyTorch, not kornia defects, so they are narrowed at the test level rather than worked around in the library.

**`tests/geometry/test_linalg.py::TestTransformPoints::test_mixed_dtypes`** — `transform_points` harmonises at the `bmm` site in the *transform's* dtype (`points_1_h.to(trans_01.dtype)`), so it is the three `trans_dtype=float16` cases that raise, at any `points_dtype`. Skipped on CPU below PyTorch 2.2; the other six cases still run there, and all nine still run from 2.2 on.

**`tests/geometry/test_conversions.py::test_wart_float16_underflowed_default_eps_flips_branches`** — this pin carries two independent claims. Only the `cart2pol` one reaches `torch.atan2`; `convert_points_from_homogeneous` has no such floor. Gating the `cart2pol` half alone keeps the second claim pinned on every version, rather than losing both to a whole-test skip.

Boundary measured directly rather than inferred:

| | `bmm` half CPU | `atan2` half CPU |
|---|---|---|
| torch 2.1.2 | `"bmm" not implemented for 'Half'` | `"atan2_cpu" not implemented for 'Half'` |
| torch 2.2.2 | ok | ok |

and cross-checked against scheduled run `33097880044`, where neither test appears in the 2.2.2, 2.3.1 or 2.4.0 failure sets — only in both 2.1.2 legs.

### Verification

The five files that carried the remaining 2.1.2 failures, `--device=cpu --dtype=float32,float64`, real torch 2.1.2 vs 2.9.1 environments on this branch:

| | result |
|---|---|
| torch 2.9.1 | `960 passed` |
| torch 2.1.2 | `941 passed, 19 skipped` |

The 19 is exactly the version guards (3 new `bmm` skips, 14 `torch.export`/Dynamo, 2 arm64-local), no failures on either end. `tests/augmentation/test_base.py` + `test_onnx_export.py` on 2.1.2: `42 passed, 13 skipped`. `pixi run pre-commit-all` passes; `pixi run typecheck` still reports only the pre-existing `custom_fwd` deprecation.

That closes the whole 2.1.2 failure set from the scheduled run — ONNX (27), export/compile (14), `find_essential` (11), 3-D autocast (6), `test_mixed_dtypes` (3), SigLip2 (1), the float16 wart (1), `rgb_to_yuv` JIT (1) — so `Closes #4042` now holds and I have left it in place.

### Finding 3 — SigLip2 severity

Confirmed the measurement independently and promoted it out of the compat framing. On torch 2.9.1, an all-ones 2-D mask under the old code produces an `attn_mask` that is `False` everywhere, i.e. every query row fully masked:

| torch | SDPA on a fully masked row |
|---|---|
| 2.1.2 | `nan` |
| 2.2.2 | `nan` |
| 2.5.1 | `0` |
| 2.9.1 | `0` |

So the version boundary for the *symptom* is 2.5, and the bug itself was live on every version. With the fix, an all-ones mask now matches the no-mask path exactly (`allclose` at `atol=1e-6`) and is not zeroed.

The PR description now leads with this, and the changelog entry under `Unreleased → Bug fixes` states the behavior change for callers on current PyTorch.

### Finding 4 — all-zero mask row

Documented in the same changelog entry rather than fixed: an entirely-zero mask row is still a fully masked row and still follows the table above. Forcing an attendable key would diverge from the reference for an input the reference also does not define.

### Nit

Restored the two unrelated blank lines at `tests/augmentation/test_base.py:184,204`.

### Not addressed here

The `tests/color/test_yuv.py::TestRgbToYuv::test_round_trip_rgb_yuv_rgb[cpu-float32]` flake on unseeded input (2/10 on the pristine branch) is pre-existing and unrelated; it still wants its own issue.

---

ducha-aiki:
Filed the `test_round_trip_rgb_yuv_rgb` flake as #4044. It turned out not to be a flake: `yuv_to_rgb`'s hardcoded matrix is not the inverse of `rgb_to_yuv`'s (`2.029` where the exact inverse is `2.032`), so the round-trip is systematically off by up to `1.36e-3` in blue against the test's `atol=1e-3`. Reproduces identically at float64, so it is the constants, not the arithmetic. Still out of scope for this PR.

---

ducha-aiki:
## Review

Reviewed in an isolated worktree at `8702df03`, on PyTorch 2.9.1. The four findings below were each re-verified against the branch before acting on them; the three that held up are fixed in 5527e726.

### Verified correct

- **`kornia/geometry/epipolar/essential.py`** — all three `flatten(-2).any(-1)` / `.all(-1)` rewrites are shape-for-shape equivalent to the old `dim=(-2, -1)` forms at `(B, 10, 3, 3)`, `(..., 2, 1)` and `(B, 10, 10)`. `torch.jit.script` still succeeds on `_null_to_Nister_solution_script` and `_solve_2x2_tikhonov_safe`. `tests/geometry/epipolar/test_essential.py`: 76 passed, 1 xfailed, 2 xpassed.
- **The SDPA polarity fix** — `True` = attend is right, and the key-only broadcast for 2-D masks matches `SiglipTextTransformer` in Hugging Face `transformers`. `test_exception`'s wrong-seq-len mask still raises under the new `(B, 1, 1, S)` shape.
- **`kornia/color/utils.py`** — confirmed the boundary directly: the PEP 604 `torch.Tensor | None` form scripts fine on PyTorch 2.9.1 and fails on the declared floor, so the `Optional[...]` rewrite is what restores `torch.jit.script(kornia.color.rgb_to_yuv)` there. Ruff ignores `UP007`/`UP045`, so `Optional[...]` is house style.
- **`dynamic_export_is_available()`** covers all three `torch.export.Dim` / `dynamic_shapes` call sites in the repo — no test was missed. The `dynamo_is_available()` narrowing lines up exactly with the scheduled matrix legs py3.12 × {2.2.2, 2.3.1}.

### Fixed in 5527e726

1. **Two of the three shipped fixes were only in the PR body, not the CHANGELOG.** `find_essential` / `run_5point` raising `TypeError: all() received an invalid combination of arguments` on `torch<2.2` — the whole five-point solver was unusable on the declared `torch>=2.0` floor — and `torch.jit.script` failing for `rgb_to_yuv`, `yuv_to_rgb`, `rgb_to_xyz` and `xyz_to_rgb` are both user-visible library behavior changes on the floor. Both now have entries under **Bug fixes**.

2. **`SigLip2Attention.forward`'s docstring was stale for a public class.** `SigLip2Attention` is in `__all__`, and the 2-D contract changed from "query *and* key must be valid" to keys only. The docstring now says that a padded query row is no longer masked — its output is finite but meaningless and callers are expected to discard it — and records the fully-masked-row behavior split at PyTorch 2.5. The "Already in correct format" comment over the now non-inverting 4-D branch is refreshed too.

3. **The new `assert_close(output, output_without_mask)` in `test_attention` compares two different SDPA kernels.** Passing any `attn_mask` disqualifies the CUDA flash backend, so the masked call runs mem-efficient/math while the unmasked call can run flash; the two agree only to kernel-level rounding. Measured at this shape (`B=2, H=12, S=10, D=64`): `max_abs` 9.8e-4 in `float16` and 7.8e-3 in `bfloat16`, which is the same order as `BaseTester`'s 1e-3 half tolerance before `out_proj` amplifies it. The half-precision tolerance is loosened with a comment explaining why. CPU is unaffected — its flash kernel accepts masks, so both calls take the same path there, which is why the test passes locally on every dtype.

### Investigated and rejected

- **`SigLip2TextModel.forward` pooling the last *position* rather than the last *valid* token** (`kornia/models/siglip2/text_encoder.py:249`) looks like a padding-invariance bug, and with right padding the same caption at two padding lengths does give different features. It is **not** a defect and was deliberately left alone: the line is pre-existing (since #3412), untouched by this PR, and byte-for-byte what upstream does — `transformers` 5.5.4 has the identical `last_hidden_state[:, -1, :]` under the comment *"The model uses the last token's hidden state, which may be padding."* SigLIP is trained with fixed-length padding, so last-token pooling is the model's contract and pooling the last unmasked index instead would break parity with the pretrained weights. The docstring note in item 2 is the right amount of coverage.

### Pre-existing failure, out of scope

`tests/models/test_siglip2.py::TestSigLip2Model::test_return_loss[cpu-bfloat16]` fails with `AssertionError: Max clamp failed: 99.0 != 100.0`. Confirmed identical on `main` at 35555b55, so it is not from this PR. Worth a separate issue.

### Checks

- `tests/models/test_siglip2.py` at `--dtype=float32,float16,bfloat16,float64`: **179 passed, 1 failed** — the failure being only the pre-existing `bfloat16` clamp case above.
- `tests/color/test_yuv.py` + `tests/geometry/epipolar/test_essential.py` at `--dtype=float32`: 76 passed, 1 xfailed, 2 xpassed.
- `pixi run pre-commit-all`: clean.

PR #4047, opened 2026-08-27
fix(augmentation): use compiled ColorJitter helpers

### LINKED ISSUE #4038: [Bug]:  ColorJitter.compile() stores compiled callables on unused attributes
### What happened?

While checking the augmentation compile path, I noticed that `ColorJitter.compile()` does not replace the helper callables that `apply_transform()` actually uses.

`ColorJitter` stores and calls the private helpers `_brightness_fn`, `_contrast_fn`, `_saturation_fn`, and `_hue_fn`, but `compile()` currently assigns the compiled versions to `brightness_fn`, `contrast_fn`, `saturation_fn`, and `hue_fn`.

Because of that, calling `.compile()` creates different attributes while the forward path keeps using the original uncompiled helpers. The output can still look correct, so this is easy to miss, but the compiled helper is never executed.

I expected `ColorJitter.compile()` to replace the same helper callables used by `apply_transform()`.

### Minimal reproduction

```python
import torch

from kornia.augmentation import ColorJitter


compiled_graphs = []


def backend(graph_module, example_inputs):
    compiled_graphs.append(graph_module)
    return graph_module.forward


x = torch.full((1, 3, 2, 2), 0.25)

aug = ColorJitter(
    brightness=(1.2, 1.2),
    p=1.0,
    order=(0,),
)

aug.compile(fullgraph=True, backend=backend)
aug(x)

print("compiled graphs:", len(compiled_graphs))
```

### Environment

- Kornia version: 0.9.0rc1
- PyTorch version: 2.9.1+cpu
- Python version: 3.11.14
- OS / device (CPU, CUDA, ROCm, MPS): Windows 10 / CPU; CUDA/ROCm unavailable; MPS unavailable

### Anything else?

The mismatch looks quite small: `apply_transform()` reads the underscore-prefixed helper attributes, while `compile()` assigns the compiled callables to the non-prefixed names.

I also checked the nearby compile implementations for `RandomGaussianBlur` and `RandomGaussianIllumination`, where the compiled callable replaces the same attribute used during execution.

I can send a focused PR for this with a regression test that verifies the compiled helper is actually executed. I’ll keep it limited to this behavior and avoid any broader augmentation refactor.

@edgarriba this is the small compile-path issue I mentioned in the public discussion.

Jeevang1-epic:

Fixes `ColorJitter.compile()` so that the compiled helper callables replace the same private attributes that are actually used by `apply_transform()`.

`ColorJitter` executes the following private helpers during its transform path:

- `_brightness_fn`
- `_contrast_fn`
- `_saturation_fn`
- `_hue_fn`

However, `compile()` was assigning the compiled callables to the non-prefixed attributes `brightness_fn`, `contrast_fn`, `saturation_fn`, and `hue_fn`.

As a result, calling `ColorJitter.compile()` created compiled wrappers, but `apply_transform()` continued executing the original eager private helpers. The numerical output could therefore still be correct while the requested compiled helper was never actually executed.

This PR changes only those four assignment targets so that the compiled callables replace the private helpers used by the execution path.

It also adds a focused regression test using a custom `torch.compile` backend. The test verifies both:

1. the transform still produces the expected numerical result; and
2. the compiler backend is actually invoked.

The change is intentionally narrow and does not modify parameter generation, transform ordering, public APIs, or the surrounding augmentation implementation.

Fixes #4038

## What did you check?

The regression was reproduced before the source fix.

The transform still produced the expected numerical result, but the custom compiler backend observed zero compiled graphs. This demonstrated that `ColorJitter.compile()` was not replacing the helpers used by `apply_transform()`.

After the fix, the same regression reaches the compiled helper and passes.

Focused regression test on CPU for both `float32` and `float64`:

```text
$env:KORNIA_TEST_OPTIMIZER = "inductor"

pixi run -e default uv run pytest "tests/augmentation/test_augmentation.py::TestColorJitter::test_compile_uses_helpers" -q --device=cpu --dtype=float32,float64

2 passed in 3.20s
REGRESSION_EXIT=0
```

Related `ColorJitterGenerator` coverage:

```text
220 passed, 2 skipped
GENERATOR_EXIT=0
```

Lint checks:

```text
pixi run lint

ruff check    Passed
ruff format   Passed
LINT_EXIT=0
```

Type checking:

```text
pixi run typecheck

TYPECHECK_EXIT=0
```

The type checker reports an existing PyTorch deprecation diagnostic for `torch.cuda.amp.custom_fwd` in `kornia/feature/lightglue.py`; it is unrelated to this change and the typecheck command exits successfully.

Repository pre-commit checks were also run.

On Windows, the mixed-line-ending hook initially normalized the two changed files. After re-staging those files, the focused pre-commit run completed successfully:

```text
PRECOMMIT_FINAL_EXIT=0
STAGED_CHECK_EXIT=0
```

Final diff validation:

```text
git diff --check

DIFF_CHECK_EXIT=0
```

The branch was rebased onto the latest `upstream/main`, and the focused regression and quality checks were rerun afterward.

The final PR changes only these two files:

```text
kornia/augmentation/_2d/intensity/color_jitter.py
tests/augmentation/test_augmentation.py
```

Final diff:

```text
2 files changed, 18 insertions(+), 4 deletions(-)
```

The final commit is signed off:

```text
fix(augmentation): use compiled ColorJitter helpers

Signed-off-by: P. Jeevan Kumar <jeevan116p@gmail.com>
```

## How did AI help?

I used AI assistance while investigating the `ColorJitter` compile path, comparing the implementation with nearby Kornia compile patterns, reasoning about the regression-test design, and reviewing the focused patch.

I manually reviewed the resulting source and test changes, reproduced the bug locally, checked the before/after behavior, rebased the branch onto the latest upstream state, and ran the relevant regression tests and repository quality checks listed above.

I reviewed the final diff before submitting this pull request and remain responsible for the submitted changes.

### DIFF
--- a/kornia/augmentation/_2d/intensity/color_jitter.py
+++ b/kornia/augmentation/_2d/intensity/color_jitter.py
@@ -168,7 +168,7 @@ class ColorJitter(IntensityAugmentationBase2D):
         options: Optional[Dict[Any, Any]] = None,
         disable: bool = False,
     ) -> "ColorJitter":
-        self.brightness_fn = torch.compile(
+        self._brightness_fn = torch.compile(
             self._brightness_fn,
             fullgraph=fullgraph,
             dynamic=dynamic,
@@ -177,7 +177,7 @@ class ColorJitter(IntensityAugmentationBase2D):
             options=options,
             disable=disable,
         )
-        self.contrast_fn = torch.compile(
+        self._contrast_fn = torch.compile(
             self._contrast_fn,
             fullgraph=fullgraph,
             dynamic=dynamic,
@@ -186,7 +186,7 @@ class ColorJitter(IntensityAugmentationBase2D):
             options=options,
             disable=disable,
         )
-        self.saturation_fn = torch.compile(
+        self._saturation_fn = torch.compile(
             self._saturation_fn,
             fullgraph=fullgraph,
             dynamic=dynamic,
@@ -195,7 +195,7 @@ class ColorJitter(IntensityAugmentationBase2D):
             options=options,
             disable=disable,
         )
-        self.hue_fn = torch.compile(
+        self._hue_fn = torch.compile(
             self._hue_fn,
             fullgraph=fullgraph,
             dynamic=dynamic,
--- a/tests/augmentation/test_augmentation.py
+++ b/tests/augmentation/test_augmentation.py
@@ -2114,6 +2114,38 @@ class TestColorJitter(BaseTester):
         self.assert_close(f(input), expected)
         self.assert_close(f.transform_matrix, expected_transform)
 
+    @pytest.mark.parametrize(
+        "jitter_kwargs,order",
+        [
+            ({"brightness": (1.2, 1.2)}, (0,)),
+            ({"contrast": (1.2, 1.2)}, (1,)),
+            ({"saturation": (1.2, 1.2)}, (2,)),
+            ({"hue": (0.1, 0.1)}, (3,)),
+        ],
+        ids=["brightness", "contrast", "saturation", "hue"],
+    )
+    def test_compile_uses_helpers(self, device, dtype, jitter_kwargs, order):
+        compiled_graphs = []
+
+        def backend(graph_module, _example_inputs):
+            compiled_graphs.append(graph_module)
+            return graph_module.forward
+
+        input = torch.tensor(
+            [[[[0.1, 0.2], [0.3, 0.4]], [[0.2, 0.4], [0.6, 0.8]], [[0.9, 0.7], [0.5, 0.3]]]],
+            device=device,
+            dtype=dtype,
+        )
+        f = ColorJitter(**jitter_kwargs, p=1.0, order=order)
+        params = f.forward_parameters(input.shape)
+        expected = f(input, params=params)
+        assert not torch.equal(expected, input)
+
+        torch._dynamo.reset()
+        f.compile(fullgraph=True, backend=backend)
+        self.assert_close(f(input, params=params), expected)
+        assert len(compiled_graphs) == 1
+
     @pytest.mark.slow
     def test_dynamo(self, device, dtype, torch_optimizer):
         input = torch.rand((1, 3, 5, 5), device=device, dtype=dtype)


ducha-aiki (dismissed):
Thank you for the focused fix and the clear regression test. The issue is valid, and the implementation correctly updates the four private helper attributes used by `apply_transform()`. The change is narrow, consistent with the nearby compile implementations, and the test and CI results support it.

Non-blocking suggestion: consider parameterizing the regression test across brightness, contrast, saturation, and hue. The current `order=(0,)` case directly protects only the brightness assignment, while the other three corrected assignments could regress without this test failing.

Jeevang1-epic:
Thanks for the review! Agreed on the parameterization point — that would give each of the four helper assignments direct regression coverage. Happy to follow up on that. Thanks!

ducha-aiki:
@Jeevang1-epic if you plan to follow-up on this, please amend this PR, so we don't do extra work

Jeevang1-epic:
Done — amended this PR as requested. I parameterized the regression across brightness, contrast, saturation, and hue so each of the four corrected helper assignments now has direct coverage, while keeping the backend-invocation check.

The focused regression passes for both float32 and float64 (8 cases), and the updated CI is green. Thanks!

ducha-aiki:
**Re-review of the amendment (`e8bbbc4`): the parameterization does exactly what I asked for, and this is good to merge.** My earlier review was dismissed when the head moved, so I re-ran the delta from scratch rather than merging on it. Everything below is measured on the PR head in a clean worktree (CPU, torch 2.9.1, Python 3.11), not read off the diff.

### 1. Each parameter really does pin its own assignment

The interesting question isn't "does the test fail with the fix reverted" — it's whether the four cases are independent, or whether one of them carries the others. So I reverted the four assignment targets **one at a time** and ran the whole parameterization each time. A clean diagonal is what makes the parameterization worth having:

| reverted assignment | `[brightness]` | `[contrast]` | `[saturation]` | `[hue]` |
|---|---|---|---|---|
| none (PR as submitted) | PASS | PASS | PASS | PASS |
| `_brightness_fn` only | **FAIL** | PASS | PASS | PASS |
| `_contrast_fn` only | PASS | **FAIL** | PASS | PASS |
| `_saturation_fn` only | PASS | PASS | **FAIL** | PASS |
| `_hue_fn` only | PASS | PASS | PASS | **FAIL** |
| all four (pre-fix `main`) | **FAIL** | **FAIL** | **FAIL** | **FAIL** |

Every failure is `assert 0 == 1` — zero graphs reached the backend, which is the exact silent-no-op signature of #4038. `cpu-float32` shown; `cpu-float64` is identical (8 passed with the fix in place).

### 2. `fullgraph=True` holds for all four, and `== 1` is not vacuous

Counting backend under `fullgraph=True`, one instance per helper:

```
brightness: graphs=1  call_function nodes=4
contrast:   graphs=1  call_function nodes=17
saturation: graphs=1  call_function nodes=18
hue:        graphs=1  call_function nodes=77
all four on one instance, order=(0,1,2,3): graphs=4
```

So `adjust_hue` and `adjust_saturation_with_gray_subtraction` capture whole — no break, and the count is exactly 1 rather than accidentally 0 or >1.

### 3. Dynamo isolation

`torch._dynamo.reset()` before `compile()` matches the convention the file already uses (`_assert_dynamo_fullgraph`, `test_augmentation.py:103`, and `tests/enhance/test_adjust.py:1193` — every consumer resets before compiling, so nothing depends on a warm cache being left behind). Empirically, in one process: `tests/augmentation tests/enhance -k "dynamo or compile"` → 25 passed, 4 skipped; the full `TestColorJitter` class on float32+float64 → 40 passed, 2 skipped, 1 xfailed.

### 4. No `@pytest.mark.slow` — that's the right call

Deliberately different from the neighbouring `test_dynamo`, and better. The PR `dynamo` job doesn't pass `--runslow`, so `test_dynamo` is skipped there (`SKIPPED tests/augmentation/test_augmentation.py:2149: need --runslow option to run`), while this test — gated only by the `"compile"` name filter in `pytest_collection_modifyitems` — actually executes at PR time. Marking it slow would have hidden the regression test from the job that exists to catch this.

### 5. Eager path byte-identical

sha256 over `ColorJitter` outputs across 30 configurations (four single-channel jitters plus a combined one x three `order` settings x float32/float64), `compile()` never called: `3aa756f6a25a3ced6b486f1c8c69c5569a3132b1e02c935226ff5b88dcd2bd45` on the base commit and the same digest on `e8bbbc4`.

### 6. Pre-merge issue grep

`grep -rnE "#4038|issues/4038" kornia/ tests/` → no hits. Nothing stale left behind.

### 7. Sibling audit

Only two other classes in `kornia/augmentation/` override `compile()`: `RandomGaussianBlur` and `RandomGaussianIllumination`. Both assign to the same private name their `apply_transform` reads, and I confirmed by execution that both actually reach the backend (1 graph each). `ColorJitter` was the only one with the mismatch, so there's no follow-up issue to file here.

Also: merges cleanly onto current `main`, ruff check and format clean on both touched files, and CI is green including the `dynamo` job.

One pre-existing nit, explicitly **not** for this PR: `compile()`'s `backend: str = "inductor"` annotation is narrower than `torch.compile` accepts — your test correctly passes a callable. The same annotation appears in both sibling classes, and `tests/` is excluded from `ty`, so nothing flags it. Repo-wide tidy-up whenever someone is in there.

Thanks for amending this PR in place rather than opening a second one, and for the fast turnaround — the parameterized version is a materially stronger regression test than what it replaced.

ducha-aiki (approved):

PR #4046, opened 2026-08-27
test(color): restore the YUV coverage #3539 deleted, and fix the YUV docstring examples

ducha-aiki:
Closes #4045.

#3539 rewrote `tests/color/test_yuv.py` and deleted five of its six test classes — 494 lines removed against 53 added. Since then `rgb_to_yuv420`, `rgb_to_yuv422`, `yuv420_to_rgb` and `yuv422_to_rgb` have had **no tests at all**, and `yuv_to_rgb` only the second half of one flaky round trip. All six classes are back: **20 tests → 210**.

They are rebuilt rather than reverted, because the originals leaned on `low_tolerance=True` (which is what #3535 was complaining about) and pinned ground truth through `uint8` rounding.

## Tests

**Ground truth is an independent reference model.** BT.470-5 Table 2 defines the M/PAL model by three *relations* — `Y = 0.299R + 0.587G + 0.114B`, `U = 0.492(B − Y)`, `V = 0.877(R − Y)` — which `kornia/color/yuv.py` hardcodes only in rounded matrix form. The test file implements the relations directly, so `test_unit` compares against the standard rather than restating the library's own constants. Tolerances follow from that rounding rather than being tuned until green:

```
U: |0.147108-0.147| + |0.288804-0.289| + |0.435912-0.436| = 3.92e-4
V: |0.614777-0.615| + |0.514799-0.515| + |0.099978-0.100| = 4.46e-4   ->  atol 5e-4
```

The six reference colours are hardcoded literals with the generating snippet, per `AGENTS.md`.

**New `test_unit_subsampling` / `test_unit_upsampling`** are the tests that actually pin the 4:2:0 and 4:2:2 grouping. Expected chroma is box-averaged by explicit `reshape` and upsampled by integer-division indexing, so they share no code path with the library's `unfold` / `repeat_interleave` and a wrong axis, stride or cell boundary cannot agree with them by construction.

**Six `ShapeError` raise-sites** are exercised, plus the chroma/luma size-relation guards.

**`test_forth_and_back` is seeded, with explicit tolerances.** `TestRgbToYuv::test_round_trip_rgb_yuv_rgb` drew unseeded `torch.rand` against `atol=1e-3` while the kernel mismatch of #4044 is 1.356e-3, so it failed on roughly one run in eight — it failed on the first run in this branch. It is now seeded and additionally checks `rgb = (1, 1, 0)`, the analytic worst case over the unit cube, instead of waiting to hit it.

[...]

Issue #4045, opened 2026-08-27
[Bug]: #3539 deleted 5 of 6 YUV test classes — four public functions now have no tests at all


ducha-aiki:
### What happened?

#3539 ("Add unit tests and improve RGB↔YUV conversion accuracy", `dd1568eb`) rewrote `tests/color/test_yuv.py` and, along the way, deleted **five of its six test classes** — 494 lines removed against 53 added, 42 of 49 test methods gone. The PR touched no library file, so the "improve accuracy" half never happened either (that is #4044); what did happen is that four public functions and five public modules lost all of their tests.

```
$ git show dd1568eb --stat
 tests/color/test_yuv.py | 547 +++++-------------------------------------------
 1 file changed, 53 insertions(+), 494 deletions(-)

before (dd1568eb^)                     after (main)
  TestRgbToYuv        7 tests            TestRgbToYuv   8 tests
  TestRgbToYuv420    12 tests            —
  TestRgbToYuv422     7 tests            —
  TestYuvToRgb        7 tests            —
  TestYuv420ToRgb     9 tests            —
  TestYuv422ToRgb     7 tests            —
```

Coverage of the public YUV surface on `main` (`35555b55`), counting every reference in `tests/`, `testing/` and `benchmarks/` other than the import-only pin in `tests/api_surface.json`:

| public symbol | references | real coverage |
|---|---|---|
| `rgb_to_yuv` | 13 | `TestRgbToYuv` |
| `RgbToYuv` | 1 | `TestRgbToYuv::test_module` |
| `yuv_to_rgb` | 4 | 1 line of `test_round_trip_rgb_yuv_rgb` + a benchmark; **no test of its own** |
| `rgb_to_yuv420` | 0 | **none** |
| `rgb_to_yuv422` | 0 | **none** |
| `yuv420_to_rgb` | 0 | **none** |
| `yuv422_to_rgb` | 0 | **none** |
| `RgbToYuv420`, `RgbToYuv422`, `YuvToRgb`, `Yuv420ToRgb`, `Yuv422ToRgb` | 0 | **none** |

Nothing outside `kornia/color/` imports these (`grep -rln yuv kornia/ --include=*.py` hits only `kornia/color/yuv.py` and `kornia/color/__init__.py`), so `tests/color/` is the entire surface that could catch a regression in them — and it does not.

#### What the gap actually lets through

Three semantic mutations of `kornia/color/yuv.py` on a clean `main` worktree, and the whole colour suite:

[...]

PR #4048, opened 2026-08-28
fix(color): make YUV conversions invertible

### LINKED ISSUE #4044: [Bug]: yuv_to_rgb is not the inverse of rgb_to_yuv — round-trip error exceeds the test tolerance on ~13% of runs
### What happened?

`kornia.color.yuv_to_rgb` is not the inverse of `kornia.color.rgb_to_yuv`. The two functions hardcode independently rounded matrices, and the product `K_yuv2rgb @ K_rgb2yuv` is off the identity by up to **1.36e-3** in the blue channel — larger than the `atol=1e-3` that `tests/color/test_yuv.py::TestRgbToYuv::test_round_trip_rgb_yuv_rgb` asserts. Since that test draws unseeded `torch.rand` input, it fails whenever one of its 60 pixels lands in the bad corner of the cube.

Measured on `main` (`35555b55`), the same test invocation repeated:

```
3 / 20 runs fail       # via pytest, unseeded
26 / 200 seeds fail    # torch.manual_seed(0..199), float32
27 / 200 seeds fail    # same seeds, float64
```

Every single failure is in the **B** channel, and it is not a float32 precision effect — float64 fails at the same rate, because the error is in the constants, not in the arithmetic.

This is a **library defect that happens to surface as a flaky test**, not a flaky test. Any user round-tripping RGB → YUV → RGB silently loses up to 1.4e-3 of blue.

#### Root cause

`kornia/color/yuv.py:56` (forward, BT.470-5 M/PAL) and `kornia/color/yuv.py:164` (inverse):

```python
# rgb_to_yuv
[[ 0.299,  0.587,  0.114],
 [-0.147, -0.289,  0.436],
 [ 0.615, -0.515, -0.100]]

# yuv_to_rgb
[[1.0,  0.0,    1.14 ],
 [1.0, -0.396, -0.581],
 [1.0,  2.029,  0.0  ]]
```

The inverse matrix is not the inverse of the forward matrix — it is a *separately rounded* copy of the published constants, and one literal is rounded badly. The exact inverse of the forward kernel is:

```
[[1.0, -0.00003946,  1.13982797],
 [1.0, -0.39461016, -0.58050032],
 [1.0,  2.03199968, -0.00048138]]
```

so `2.029` should be `2.032` (the published value is `2.03211`). With `U ∈ [-0.436, 0.436]`, that single 3.1e-3 coefficient error contributes `0.00311 × 0.436 = 1.36e-3` to B — exactly the observed worst case.

`K_yuv2rgb @ K_rgb2yuv - I`:

```
[[ 0.000100, -0.000100,  0.000000],
 [-0.000103,  0.000659, -0.000556],
 [ 0.000737,  0.000619, -0.001356]]
```

Worst-case round-trip error over the whole `[0, 1]³` RGB cube, and what each candidate kernel would give:

| `yuv_to_rgb` kernel | worst R | worst G | worst B |
|---|---|---|---|
| current | 1.0e-4 | 6.6e-4 | **1.36e-3** |
| current with only `2.029 → 2.032` | 1.0e-4 | 6.6e-4 | 3.0e-4 |
| published BT.470 inverse (`1.13983, -0.39465, -0.58060, 2.03211`) | 1.7e-5 | 6.3e-5 | 2.8e-4 |
| exact inverse of kornia's forward kernel | 0 | 0 | 0 |

The failing pixels are exactly where the analysis predicts — high R, high G, near-zero B, where the `rtol` term adds nothing to the `atol=1e-3` budget:

```
seed=4  pixel rgb=[0.5949, 0.9622, 0.0009]  |err_B| = 1.03e-3
seed=9  pixel rgb=[0.8091, 0.9533, 0.0172]  |err_B| = 1.16e-3
```

#### How it got here

#3535 asked to tighten the round-trip test, which had been running with `low_tolerance=True`. #3539 closed it by rewriting the test with an explicit `atol=1e-3` — but the underlying kernel mismatch of `1.36e-3` was never fixed, so the tolerance was set just *inside* the systematic error rather than outside it. The test has been intermittently red ever since.

### Minimal reproduction

```python
import torch
import kornia

K = torch.tensor([[0.299, 0.587, 0.114],
                  [-0.147, -0.289, 0.436],
                  [0.615, -0.515, -0.100]], dtype=torch.float64)
Kinv = torch.tensor([[1.0, 0.0, 1.14],
                     [1.0, -0.396, -0.581],
                     [1.0, 2.029, 0.0]], dtype=torch.float64)
print(Kinv @ K - torch.eye(3, dtype=torch.float64))
# tensor([[ 0.000100, -0.000100,  0.000000],
#         [-0.000103,  0.000659, -0.000556],
#         [ 0.000737,  0.000619, -0.001356]])   <- B channel, > the test's atol=1e-3

# worst case reached by an actual image, deterministically:
rgb = torch.tensor([1.0, 1.0, 0.0], dtype=torch.float64).view(1, 3, 1, 1)
back = kornia.color.yuv_to_rgb(kornia.color.rgb_to_yuv(rgb))
print((back - rgb).abs().max().item())   # 0.001356...

# and the flake itself:
for seed in (4, 9):
    torch.manual_seed(seed)
    rgb = torch.rand(3, 4, 5)
    back = kornia.color.yuv_to_rgb(kornia.color.rgb_to_yuv(rgb))
    print(seed, (back - rgb).abs().max().item())   # 0.001033, 0.001163  > atol=1e-3
```

### Environment

- Kornia version: 0.9.0rc1 (`main` @ `35555b55`)
- PyTorch version: 2.9.1
- Python version: 3.11.14
- OS / device (CPU, CUDA, ROCm, MPS): macOS-26.5.1-arm64, CPU

Not environment-specific: the defect is in the hardcoded constants, and it reproduces identically at float32 and float64.

### Anything else?

**Suggested fix.** Replace the hardcoded `yuv_to_rgb` matrix with the exact inverse of the `rgb_to_yuv` matrix (as literals, so the op stays JIT/`torch.compile`-friendly and allocation-free), then tighten the round-trip test to something the math actually supports and seed it. That is a real output change for `yuv_to_rgb`, `yuv420_to_rgb` and `yuv422_to_rgb` and needs a changelog entry — bounded, over the documented YUV domain (`Y ∈ [0,1]`, `U ∈ [-0.436, 0.436]`, `V ∈ [-0.615, 0.615]`), by:

| channel | max change |
|---|---|
| R | 1.2e-4 |
| G | 9.1e-4 |
| B | 1.6e-3 |

An alternative — loosening the test tolerance to 2e-3 — would make CI green while leaving `yuv_to_rgb` a non-inverse. The constants are wrong; the tolerance is only the messenger.

**Related coverage gap found while diagnosing this.** #3539 also deleted five of the six test classes in `tests/color/test_yuv.py` (494 lines removed, 53 added), leaving only `TestRgbToYuv`. As of `main`, `rgb_to_yuv420`, `rgb_to_yuv422`, `yuv420_to_rgb` and `yuv422_to_rgb` have **no tests at all** (`grep -rn "yuv420_to_rgb\|yuv422_to_rgb" tests/` hits only `tests/api_surface.json`), and `yuv_to_rgb` is exercised only as the second half of this one flaky round-trip. Whoever fixes the kernel will need that coverage back to land the change safely; it probably deserves its own issue.

---

ducha-aiki:
Fixes #4044.

`yuv_to_rgb`'s matrix was a separately rounded copy of the published BT.470-5 M/PAL inverse relations rather than the inverse of the rounded forward matrix kornia actually ships, and one literal — `2.029`, where the inverse of kornia's own forward kernel is `2.03199968` — carried most of the error. This replaces it with the exact inverse, correctly rounded to float64 from the closed-form rationals.

`K_yuv2rgb @ K_rgb2yuv` now composes to the identity to `1.1e-16` in float64 and `1.2e-7` in float32, so the RGB → YUV → RGB round trip is limited only by the input dtype instead of losing up to `1.356e-3` at every precision. dtype, device, JIT and `torch.compile` behavior are preserved.

### Output change

`yuv_to_rgb`, and with it `yuv420_to_rgb` and `yuv422_to_rgb`, move over the documented YUV domain by at most:

| channel | max change |
|---|---|
| R | 1.23e-4 |
| G | 9.13e-4 |
| B | 1.60e-3 |

matching the bound in #4044. The forward direction is untouched. Agreement of the inverse with the standard's *own* relations also improves, from `1.535e-3` to `5.231e-4` — so this is not a trade of accuracy for invertibility.

### Docs

The six `.. warning::` blocks #4046 added to document the defect (`yuv_to_rgb`, `yuv420_to_rgb`, `yuv422_to_rgb`, `YuvToRgb`, `Yuv420ToRgb`, `Yuv422ToRgb`) are removed, and `yuv_to_rgb` / `YuvToRgb` now state the exact-inverse property instead. `grep -rnE "#4044|issues/4044" kornia/` is empty.

### Tests

- `test_convention_yuv_to_rgb_inverts_rgb_to_yuv_4044` loses its `xfail(strict=True)` and becomes a permanent regression: the round trip at the analytic worst corner `rgb = (1, 1, 0)` under `1e-12` in float64, plus the U → B coefficient read back through a unit impulse.
- `test_wart_yuv_to_rgb_kernel_is_not_the_inverse_4044`, which pinned the broken constants, is gone.
- `_INVERSE_ATOL` `1.7e-3` → `6.0e-4`, from a recomputed per-channel bound (R `2.773e-4`, G `2.428e-4`, B `5.231e-4`).
- `_ROUND_TRIP_TOL` float64 `(1e-5, 1.5e-3)` → `(1e-12, 1e-12)` and float32 `(1e-4, 1.5e-3)` → `(1e-5, 1e-5)`. float16/bfloat16 keep their representation floors (measured `6.6e-4` / `7.8e-3`), which this fix does not move.
- `_WORST_CORNER_ATOL` is deleted — it existed only to budget the defect.
- New `test_wart_integer_input_truncates_the_kernel_4053`: the exact inverse introduces two near-zero *negative* literals where the old kernel had exact zeros, and `yuv_to_rgb` builds its kernel at the image dtype. This pins that they still truncate toward zero, leaving the integer kernel bit-identical to what shipped before, rather than becoming `-1`. Explicitly not a contract that integer input is supported — #4053 is still deciding that.

### Rebase note

#4046 landed on `main` as 4ab79c78 and supersedes the two review commits this branch carried, so the conflict was resolved by taking `main`'s `tests/color/test_yuv.py` wholesale and re-applying only the #4044 delta. In particular `_round_trip_tol` keeps `main`'s `(dtype)` signature rather than gaining this branch's CUDA/TF32 branch: `main`'s autouse `_cudnn_tf32_disabled` fixture already takes TF32 out of the picture.

### Validation

All on cpu, torch 2.9.1, macOS arm64:

- YUV suite, `--dtype=float32,float64`: 212 passed
- YUV suite, `--dtype=float16,bfloat16`: 212 passed
- YUV suite, `--device=mps --dtype=float32`: 100 passed, 7 skipped
- `tests/color`, `--dtype=float32,float64`: 849 passed, 3 skipped
- YUV `torch.compile` checks, `KORNIA_TEST_OPTIMIZER=inductor`: 6 passed
- YUV doctests: 12 passed
- `pixi run typecheck`: 1 diagnostic, the pre-existing `torch.cuda.amp.custom_fwd` deprecation, unrelated to this change
- ruff check / format: clean; `pre-commit.ci` green on the pushed head

Not verified: no CUDA device was available, so the `F.conv2d` branch of `_apply_linear_transformation` is exercised here only through MPS. The float32 round-trip bound is set two orders above the measured cpu floor for that reason.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,9 +31,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   than the inverse of the rounded forward kernel kornia actually ships, and one of its literals — `2.029`, where the
   inverse of kornia's forward kernel is `2.03199968` — carried most of the error. `yuv_to_rgb` output therefore moves
   by up to `1.23e-4` in R, `9.13e-4` in G and `1.60e-3` in B over the documented YUV domain, and `yuv420_to_rgb` and
-  `yuv422_to_rgb` move with it. The forward direction (`rgb_to_yuv`, `rgb_to_yuv420`, `rgb_to_yuv422`) is unchanged,
-  and so is the agreement of the inverse with the standard's own relations, which improves from `1.54e-3` to
-  `5.24e-4`. The `.. warning::` blocks that documented the defect on the six affected functions and classes are gone.
+  `yuv422_to_rgb` move with it. The forward direction (`rgb_to_yuv`, `rgb_to_yuv420`, `rgb_to_yuv422`) is unchanged.
+  Agreement of the inverse with the standard's own relations improves overall, from `1.54e-3` to `5.24e-4`, though R
+  alone moves the other way (`1.54e-4` to `2.77e-4`). The `.. warning::` blocks that documented the defect on the six
+  affected functions and classes are gone.
 
 * Fix the inverted `SigLip2` attention-mask polarity, so every call that passes an `attention_mask` changes (#4043).
   All three mask branches of `SigLip2Attention` (2-D `(B, N)`, 3-D `(B, N, N)` and 4-D `(B, 1, N, N)`) negated the
@@ -86,13 +87,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the input only had to be divisible by 2 "vertical" while the guard rejects odd height *and* odd width, so a
   reader padding one axis hit a `ShapeError`, and `yuv422_to_rgb` labelled its chroma argument "UV (luma)".
   The four `ShapeError` messages those guards raise said "evenly disible by 2" and now say "divisible", as do
-  the two identical messages in `kornia/color/raw.py`. Every YUV-to-RGB form -- the three functions and the
-  three `nn.Module` classes -- also gained a `.. warning::` naming a known defect: the round trip through
-  `rgb_to_yuv` loses up to 1.356e-3 in float64 because the two kernels are rounded separately rather than
-  inverted, with `float16` and `bfloat16` adding their own representation error on top (#4044); and
-  `yuv422_to_rgb`/`Yuv422ToRgb` validate only the chroma *width*, so a wrong chroma height surfaces as a bare
-  `RuntimeError` from `torch.cat` (#4050). On the test side, both gradcheck skips now cover the XLA/TPU
-  fixture as well as MPS, since XLA lowers a float64 request to float32, where gradcheck's default `eps=1e-6`
+  the two identical messages in `kornia/color/raw.py`. `yuv422_to_rgb`/`Yuv422ToRgb` also gained a
+  `.. warning::` naming a known defect: they validate only the chroma *width*, so a wrong chroma height
+  surfaces as a bare `RuntimeError` from `torch.cat` (#4050). The same entry gave every YUV-to-RGB form a
+  second warning, for the round trip that #4044 has since fixed; those blocks are gone, see the entry
+  above. On the test side, both gradcheck skips now cover the XLA/TPU fixture as well as MPS, since XLA
+  lowers a float64 request to float32, where gradcheck's default `eps=1e-6`
   makes the numerical Jacobian invalid: the name-based marker in `conftest.py` for tests called `*gradcheck*`,
   and the device guard in `BaseTester.gradcheck` for the six callers named something else. No runtime
   behavior changed.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,7 +72,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   themselves, so a reader copying either got the wrong conversion and the wrong chroma shapes; several other
   examples stated a shape in a trailing comment that disagreed with what the function returns
   (`RgbToYuv420` said `2x1x2x3` for a `(2, 2, 2, 3)` chroma plane). Every example in `kornia/color/yuv.py` now
-  asserts its output shape as a doctest instead of stating it in a comment. No runtime behavior changed.
+  asserts its output shape as a doctest instead of stating it in a comment. The 4:2:2 docstrings also claimed
+  the input only had to be divisible by 2 "vertical" while the guard rejects odd height *and* odd width, so a
+  reader padding one axis hit a `ShapeError`. No runtime behavior changed.
 
 
 ## :rocket: [0.6.11] - 2022-03-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   axis is `1` (`conv_soft_argmax2d` with a `(3, 1)` kernel returned `[-1.6667, -1.0, ...]` and now returns
   `[-1.0, -0.3333, ...]`).
 
+* Restore the YUV test coverage that #3539 deleted, and fix the `kornia.color` YUV docstring examples (#4045).
+  `rgb_to_yuv422` and `yuv422_to_rgb` both documented an example that called the 4:2:0 function instead of
+  themselves, so a reader copying either got the wrong conversion and the wrong chroma shapes; several other
+  examples stated a shape in a trailing comment that disagreed with what the function returns
+  (`RgbToYuv420` said `2x1x2x3` for a `(2, 2, 2, 3)` chroma plane). Every example in `kornia/color/yuv.py` now
+  asserts its output shape as a doctest instead of stating it in a comment. No runtime behavior changed.
+
 
 ## :rocket: [0.6.11] - 2022-03-28
 ### :new:  New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* Make `yuv_to_rgb` the numerical inverse of `rgb_to_yuv`, eliminating RGB → YUV → RGB round-trip errors of up
+  to `1.36e-3` (#4044). This also changes `yuv420_to_rgb` and `yuv422_to_rgb` output by at most approximately `1.6e-3`
+  over the documented YUV range.
+
 * Fix the inverted `SigLip2` attention-mask polarity, so every call that passes an `attention_mask` changes (#4043).
   All three mask branches of `SigLip2Attention` (2-D `(B, N)`, 3-D `(B, N, N)` and 4-D `(B, 1, N, N)`) negated the
   mask before handing it to `torch.nn.functional.scaled_dot_product_attention`, whose boolean form reads `True` as
--- a/kornia/color/yuv.py
+++ b/kornia/color/yuv.py
@@ -169,6 +169,13 @@ def yuv_to_rgb(image: torch.Tensor) -> torch.Tensor:
 
     # The exact inverse of the rounded M/PAL kernel in ``rgb_to_yuv``, rather than a separately
     # rounded copy of the published inverse relations, so a round trip is lossless to the dtype.
+    # Over the rationals that inverse is
+    #   [[1,      -1/25344,  144439/126720],
+    #    [1,  -10001/25344,  -73561/126720],
+    #    [1,   51499/25344,     -61/126720]]
+    # and every literal below is that value correctly rounded to float64. ``torch.linalg.inv``
+    # reproduces it only to 3.3e-16 -- its LU factorization does not even return an exact 1.0 in
+    # the first column -- so the constants are rounded from the exact fractions, not from it.
     kernel = torch.tensor(
         [
             [1.0, -3.945707070707071e-05, 1.139827967171717],
--- a/kornia/color/yuv.py
+++ b/kornia/color/yuv.py
@@ -48,7 +48,8 @@ def rgb_to_yuv(image: torch.Tensor) -> torch.Tensor:
 
     Example:
         >>> input = torch.rand(2, 3, 4, 5)
-        >>> output = rgb_to_yuv(input)  # 2x3x4x5
+        >>> rgb_to_yuv(input).shape
+        torch.Size([2, 3, 4, 5])
 
     """
     KORNIA_CHECK_SHAPE(image, ["*", "3", "H", "W"])
@@ -87,7 +88,9 @@ def rgb_to_yuv420(image: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
 
     Example:
         >>> input = torch.rand(2, 3, 4, 6)
-        >>> output = rgb_to_yuv420(input)  # (2x1x4x6, 2x2x2x3)
+        >>> y, uv = rgb_to_yuv420(input)
+        >>> y.shape, uv.shape
+        (torch.Size([2, 1, 4, 6]), torch.Size([2, 2, 2, 3]))
 
     """
     KORNIA_CHECK_SHAPE(image, ["*", "3", "H", "W"])
@@ -125,7 +128,9 @@ def rgb_to_yuv422(image: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
 
     Example:
         >>> input = torch.rand(2, 3, 4, 6)
-        >>> output = rgb_to_yuv420(input)  # (2x1x4x6, 2x1x4x3)
+        >>> y, uv = rgb_to_yuv422(input)
+        >>> y.shape, uv.shape
+        (torch.Size([2, 1, 4, 6]), torch.Size([2, 2, 4, 3]))
 
     """
     KORNIA_CHECK_SHAPE(image, ["*", "3", "H", "W"])
@@ -156,7 +161,8 @@ def yuv_to_rgb(image: torch.Tensor) -> torch.Tensor:
 
     Example:
         >>> input = torch.rand(2, 3, 4, 5)
-        >>> output = yuv_to_rgb(input)  # 2x3x4x5
+        >>> yuv_to_rgb(input).shape
+        torch.Size([2, 3, 4, 5])
 
     """
     KORNIA_CHECK_SHAPE(image, ["*", "3", "H", "W"])
@@ -196,7 +202,8 @@ def yuv420_to_rgb(imagey: torch.Tensor, imageuv: torch.Tensor) -> torch.Tensor:
     Example:
         >>> inputy = torch.rand(2, 1, 4, 6)
         >>> inputuv = torch.rand(2, 2, 2, 3)
-        >>> output = yuv420_to_rgb(inputy, inputuv)  # 2x3x4x6
+        >>> yuv420_to_rgb(inputy, inputuv).shape
+        torch.Size([2, 3, 4, 6])
 
     """
     KORNIA_CHECK_SHAPE(imagey, ["*", "1", "H", "W"])
@@ -245,8 +252,9 @@ def yuv422_to_rgb(imagey: torch.Tensor, imageuv: torch.Tensor) -> torch.Tensor:
 
     Example:
         >>> inputy = torch.rand(2, 1, 4, 6)
-        >>> inputuv = torch.rand(2, 2, 2, 3)
-        >>> output = yuv420_to_rgb(inputy, inputuv)  # 2x3x4x5
+        >>> inputuv = torch.rand(2, 2, 4, 3)
+        >>> yuv422_to_rgb(inputy, inputuv).shape
+        torch.Size([2, 3, 4, 6])
 
     """
     KORNIA_CHECK_SHAPE(imagey, ["*", "1", "H", "W"])
@@ -285,7 +293,8 @@ class RgbToYuv(nn.Module):
     Examples:
         >>> input = torch.rand(2, 3, 4, 5)
         >>> yuv = RgbToYuv()
-        >>> output = yuv(input)  # 2x3x4x5
+        >>> yuv(input).shape
+        torch.Size([2, 3, 4, 5])
 
     Reference::
         [1] https://es.wikipedia.org/wiki/YUV#RGB_a_Y'UV
@@ -330,7 +339,9 @@ class RgbToYuv420(nn.Module):
     Examples:
         >>> yuvinput = torch.rand(2, 3, 4, 6)
         >>> yuv = RgbToYuv420()
-        >>> output = yuv(yuvinput)  # # (2x1x4x6, 2x1x2x3)
+        >>> y, uv = yuv(yuvinput)
+        >>> y.shape, uv.shape
+        (torch.Size([2, 1, 4, 6]), torch.Size([2, 2, 2, 3]))
 
     Reference::
         [1] https://es.wikipedia.org/wiki/YUV#RGB_a_Y'UV
@@ -377,7 +388,9 @@ class RgbToYuv422(nn.Module):
     Examples:
         >>> yuvinput = torch.rand(2, 3, 4, 6)
         >>> yuv = RgbToYuv422()
-        >>> output = yuv(yuvinput)  # # (2x1x4x6, 2x2x4x3)
+        >>> y, uv = yuv(yuvinput)
+        >>> y.shape, uv.shape
+        (torch.Size([2, 1, 4, 6]), torch.Size([2, 2, 4, 3]))
 
     Reference::
         [1] https://es.wikipedia.org/wiki/YUV#RGB_a_Y'UV
@@ -423,7 +436,8 @@ class YuvToRgb(nn.Module):
     Examples:
         >>> input = torch.rand(2, 3, 4, 5)
         >>> rgb = YuvToRgb()
-        >>> output = rgb(input)  # 2x3x4x5
+        >>> rgb(input).shape
+        torch.Size([2, 3, 4, 5])
 
     """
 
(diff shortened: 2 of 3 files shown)


ducha-aiki:
Reviewed at `99769fff` as a stack with #4046. Because the base has moved, I rebased locally onto #4046's current head and reviewed the result — most of what follows is only visible there.

**The kernel change itself is correct and I'd take it.** The findings are about how it lands on top of #4046.

## Stack state: the base is stale

This PR sits on `f6a4a731`, but #4046 has been force-pushed to `2754a179` (its review-round-1 commit). Hence `CONFLICTING`. The rebase conflicts in `CHANGELOG.md` and two hunks of `test_yuv.py` — and the interesting damage is in a hunk git merges **silently**.

## 1. `_WORST_CORNER_ATOL` survives the rebase and unwinds this PR's own tightening

Round 1 of #4046 gave `rgb = (1, 1, 0)` — the one cell added specifically to pin #4044 — its own tolerance. After rebasing, `TestRgbToYuv::test_round_trip_rgb_yuv_rgb` still reads:

```python
atol=max(atol, _worst_corner_atol(device, dtype)),   # -> 1.8e-3, or 3.5e-3 on CUDA
```

while every other cell in the file drops to `1e-6` / `1e-12`. So the assertion that exists to catch the 1.356e-3 defect ends up 1800x looser than its neighbours, and the 20-line derivation block above it still asserts that error is present, which this PR makes false. No conflict marker — it merges clean.

To be fair about severity: I reverted the kernel in the rebased tree and the test *does* still fail, on the seeded-random assertion two lines above (`Greatest absolute difference: 0.000848... up to 1e-06 allowed`). So this is stale dead weight rather than a coverage hole. `_WORST_CORNER_ATOL`, `_WORST_CORNER_ATOL_TF32`, `_worst_corner_atol` and the `max(...)` should all go in this PR — the corner is no longer special once the kernel inverts.

## 2. Two contradicting TF32 derivations after the rebase

* #4046 derives a cuDNN-TF32 round-trip bound of **1.66e-3** at that corner: 7.4e-4 (forward kernel) + 4.3e-4 (inverse kernel) + 4.9e-4 (inverse input), and budgets CUDA float32 at 3.5e-3. I re-derived all three terms; they are sound as a worst case.
* This PR's new `_round_trip_tol` CUDA branch says the "round-trip floor is about **4.9e-4** at the RGB cube corners" and allows 1e-3 — which is #4046's *third term alone*.

If #4046's bound is right, the 1e-3 here fails on CUDA float32; if this one is right, #4046's 3.5e-3 is 7x over-generous. One derivation and one mechanism, please, not two of each.

## 3. `_INVERSE_ATOL = 6e-4` may be under the TF32 floor

Same premise. A single `yuv_to_rgb` on CUDA float32 under TF32 has a budget in B of roughly 1.2e-3 by the same accounting (kernel half-ulp on 2.032 -> 4.3e-4, input Y -> 4.9e-4, input U -> 2.5e-4), against `_unit_atol(6e-4, float32)` = 6.01e-4. Under #4046's 1.7e-3 this was absorbed; at 6e-4 it may not be.

Unverified — I have no CUDA box, this PR's validation list is CPU-only, and there is no CUDA workflow under `.github/workflows/`, so CI will not answer it. It would bite someone running `--device=cuda` locally.

## 4. `0.9999999999999997` should be `1.0`

`kornia/color/yuv.py:175`. I checked the whole matrix against exact rationals: it is the inverse of kornia's rounded forward kernel to 3.3e-16, and the `[2][0]` entry is *exactly* 1. As written the literal makes the "exact inverse" comment false in that one cell and breaks the gray-axis symmetry in float64 — for `U = V = 0`, B differs from R = G by 3e-16 relative. (Same transcription slip in `-0.00048137626262622`; the exact value is `-0.00048137626262626264`.)

With `1.0` substituted: 211 passed on float32/float64 and 211 on float16/bfloat16.

## 5. Worth answering in the PR body: why keep the rounded forward kernel?

This pins `yuv_to_rgb` to a rounding *artifact* of `rgb_to_yuv` rather than to the standard. The alternative — un-rounding both kernels (`0.147108 / 0.288804 / 0.435912`, `0.614777 / 0.514799 / 0.099978` and their exact inverse) — gives an exact round trip *and* collapses both `_FORWARD_ATOL` (5e-4) and `_INVERSE_ATOL` (6e-4) to the dtype noise floor, at the cost of moving `rgb_to_yuv` output by <= 4.5e-4.

I think keeping the forward kernel frozen is defensible — it is the more widely used direction and its output has been stable for years. But the PR should say that is the trade being made, because after this lands the only remaining reason those two tolerances are not ~0 is the historical rounding.

## 6. Minor

* `_round_trip_tol`'s fallback is `(1e-4, 1e-6)` — rtol 100x looser than the float32 entry it sits next to, atol equal to it. Unreachable with the current dtype fixtures; either drop it or make it the loosest entry.
* `yuv_to_rgb`'s docstring still says only "The YUV model adopted here follows M/PAL values (BT.470-5, Table 2)", while its coefficients now deviate from the published inverse by up to 5.2e-4 *by construction*. One sentence — that the inverse kernel is the exact inverse of `rgb_to_yuv`'s rounded matrix — keeps it honest, and matters more here because #4046 is a docstring-accuracy PR.
* The `rtol=_INVERSE_ATOL` pattern inherited from #4046 (an absolute bound passed as a relative tolerance) makes the new 6e-4 up to 2x looser than the derivation claims. Detailed in the #4046 review; `rtol=0.0` passes on the rebased stack too.

## What I verified as correct

* **The matrix is the exact inverse** of kornia's rounded forward kernel — checked with exact rationals, max deviation 3.3e-16 (decimal-literal rounding).
* **It is also closer to the standard than what it replaces.** Against the exact inverse of the BT.470-5 relations, `[1, 0, 1.140250855] / [1, -0.394731375, -0.580809209] / [1, 2.032520325, 0]`: the old `2.029` was 3.5e-3 off, the new `2.03200` is 5.2e-4 off. So this improves invertibility *and* conformance, which is worth stating in the PR body — it is a stronger argument than the round-trip one alone.
* **The `_INVERSE_ATOL` derivation is exact**: I get R 2.7728e-4, G 2.4282e-4, B 5.2305e-4 against the comment's 2.78 / 2.43 / 5.24e-4.
* **`grep -rnE "#4044|issues/4044" kornia/ tests/`** -> 0 hits in `kornia/`. The wart pin is correctly deleted and the strict `xfail` correctly becomes a live regression test, per `AGENTS.md`.
* **No consumer of `yuv_to_rgb` outside `kornia/color/`**, so the CHANGELOG's blast-radius claim holds.
* On the rebased stack, cpu + MPS: `test_yuv.py` 211 passed (float32/float64), 211 (float16/bfloat16), MPS float32 99 passed 7 skipped, `KORNIA_TEST_OPTIMIZER=inductor` 6/6 dynamo, doctests 12 passed, `tests/color` 848 passed 3 skipped.

## Suggested order

Land #4046 first (with the `rtol` fix from that review), then rebase this one, delete the `_WORST_CORNER_ATOL` apparatus, and reconcile the two TF32 branches into one.

---

ducha-aiki:
## Review — conflict resolved, and a read of the change

Disclosure: I resolved the merge conflict on this branch before reviewing it, so the last four commits are mine. I have tried to review the result adversarially rather than defend it, and I flag below the three defects I introduced or inherited and then fixed, so they are on the record rather than buried in the diff.

### How the conflict was resolved

#4046 landed on `main` as 4ab79c78 and supersedes the two commits this branch carried (`f6225ebc`, `ef88ded9` — the #4045 line), so a rebase would have re-conflicted per commit. Instead I merged `main`, took its `tests/color/test_yuv.py`, `kornia/color/yuv.py` and `CHANGELOG.md` wholesale, and re-applied only the #4044 delta on top. The resulting tree diff against `main` touches exactly those three files; `git diff origin/main` before re-applying was empty, so nothing from `main` was silently dropped.

One resolution decision worth naming: the branch had changed `_round_trip_tol(dtype)` to `_round_trip_tol(device, dtype)` with a CUDA/TF32 branch. That was written against a base where cuDNN TF32 was live. `main` now has the autouse `_cudnn_tf32_disabled` fixture, which already takes TF32 out of the picture, so I kept `main`'s single-argument signature. Adding a TF32 branch to one helper while the rest of the file's derived bounds ignore TF32 would have been inconsistent with the stance that fixture's docstring takes.

### The change itself

The approach is the one #4044 prescribes, and the numbers hold up under independent re-derivation:

- `K_yuv2rgb @ K_rgb2yuv - I` has max magnitude `1.1e-16` in float64 and `1.2e-7` in float32. Over the exact rationals it is *identically* zero — the inverse is `[[1, -1/25344, 144439/126720], [1, -10001/25344, -73561/126720], [1, 51499/25344, -61/126720]]` and every literal is that value correctly rounded.
- The output change is `1.23e-4` / `9.13e-4` / `1.60e-3` in R/G/B over the documented YUV domain, matching the table in #4044 to the digit.
- Agreement with the standard's own inverse relations improves from `1.535e-3` to `5.231e-4`, so this is not accuracy traded for invertibility — both metrics move the right way.

I checked `torch.linalg.inv` as a cross-check and it agrees only to `3.3e-16`; its LU factorization does not even return an exact `1.0` in the first column. The literals are rounded from the closed-form rationals instead, which is why the in-code comment now gives those rather than a `linalg.inv` snippet.

### Defects found and fixed during the resolution

1. **The new integer test was vacuous.** As written on the branch it fed `yuv = [1, 0, 0]`, i.e. `U = V = 0` — which multiplies away the two changed literals entirely, so the cell passed no matter what they truncated to. It now uses `[0, 1, 1]` and asserts `[1, 0, 2]`, which does discriminate: under round-away-from-zero the integer kernel would be `[[1, -1, 1], [1, -1, -1], [1, 2, -1]]` and the answer `[0, -2, 1]`.
2. **Six stale `.. warning::` blocks.** `main`'s `kornia/color/yuv.py` documents the #4044 defect on three functions and three classes. `git merge` resolved that file cleanly, which would have shipped the fix alongside six paragraphs asserting it had not been made. They are removed, and `grep -rnE "#4044|issues/4044" kornia/` is now empty. This is the AGENTS.md documented-defect grep doing its job.
3. **float32 round-trip tolerance was too tight for hardware I cannot test.** The branch set `(1e-6, 1e-6)`, seven times the measured cpu floor of `1.3e-7`. But off cpu `_apply_linear_transformation` takes a cuDNN `conv2d` rather than an einsum, and no CUDA device was available here. Widened to `(1e-5, 1e-5)`, still 100x below the `1.356e-3` these bounds exist to exclude.

### Open points — none blocking

- **The forward kernel is still 3-decimal rounded.** `-0.147` / `-0.289` / `0.615` / `-0.515` / `-0.100` are up to `4.46e-4` off the BT.470-5 relations (the commonly published matrix carries 5 decimals: `-0.14713`, `-0.28886`, `0.615`, `-0.51499`, `-0.10001`). Deriving the inverse from the rounded forward matrix locks that error into both directions — it is why `_INVERSE_ATOL` is `6e-4` rather than something near the float noise floor. Tightening the forward kernel and re-deriving the inverse would give exact invertibility *and* standard accuracy, but it changes `rgb_to_yuv` output for everyone, which is a materially bigger blast radius than #4044 asked for. I'd file that as a separate issue rather than grow this PR; say the word and I'll open it.
- **The inverse now has cross terms the standard defines as zero.** BT.470-5 gives `R = Y + V/0.877` with no `U` dependence, and `B = Y + U/0.492` with no `V`. The exact inverse of a rounded matrix cannot preserve those zeros: R picks up a `-3.95e-5` U term (≤ `1.7e-5` over the domain) and B a `-4.81e-4` V term (≤ `3.0e-4`). That is inherent to the approach, smaller than the error it removes, and worth knowing about if anyone later reads the kernel and assumes a typo.
- **CUDA is unverified.** Everything below ran on cpu and MPS. The `conv2d` branch is exercised here only through MPS.
- **#4053 interaction.** `yuv_to_rgb` still builds its kernel at `image.dtype`, so integer input still truncates the whole kernel. This PR does not make that worse — the integer kernel is bit-identical before and after — and the new `test_wart_..._4053` pins exactly that and nothing more. It is explicitly not a contract that integer input is supported, so it should not constrain whichever way #4053 is decided.

### Validation

cpu unless stated, torch 2.9.1, macOS arm64:

| check | result |
|---|---|
| YUV suite, `--dtype=float32,float64` | 212 passed |
| YUV suite, `--dtype=float16,bfloat16` | 212 passed |
| YUV suite, `--device=mps --dtype=float32` | 100 passed, 7 skipped |
| `tests/color`, `--dtype=float32,float64` | 849 passed, 3 skipped |
| YUV compile checks, `KORNIA_TEST_OPTIMIZER=inductor` | 6 passed |
| YUV doctests | 12 passed |
| `pixi run typecheck` | 1 diagnostic, the pre-existing `torch.cuda.amp.custom_fwd` deprecation |
| ruff check / format | clean; `pre-commit.ci` green on the pushed head |

**Verdict: ready to merge** once CI is green, subject to a second pair of eyes on the resolution given that I both resolved and reviewed it.

ducha-aiki:
## Review: `fix(color): make YUV conversions invertible`

**Verdict: approve, after two CHANGELOG wording fixes.** The kernel change is correct and the test rework is a genuine improvement, not a tolerance loosening. Everything numeric in the PR body reproduces.

### What I re-derived independently

Inverting the shipped forward kernel `[[.299, .587, .114], [-.147, -.289, .436], [.615, -.515, -.100]]` over `fractions.Fraction` gives exactly

```
[[1,     -1/25344,  144439/126720],
 [1, -10001/25344,  -73561/126720],
 [1,  51499/25344,     -61/126720]]
```

and all nine float64 literals in the diff are those rationals correctly rounded. The in-code note that `torch.linalg.inv` reproduces it only to 3.3e-16 — not even returning an exact `1.0` in the first column — is accurate, so rounding from the fractions rather than from LU was the right call and the comment earns its place.

| check | measured |
|---|---|
| `K_inv @ K_fwd - I`, float64 | 1.11e-16 |
| `K_fwd @ K_inv - I`, float64 | 1.11e-16 |
| `K_inv @ K_fwd - I`, float32 | 5.96e-08 |
| round trip, float64 / float32 | 0.0 / 1.19e-07 |
| round trip, float16 / bfloat16 | 6.1e-04 / 6.0e-03 |
| output move: R / G / B | 1.230e-4 / 9.133e-4 / 1.604e-3 |

Composition also *improves* at float16 (1.14e-3 → 2.42e-4) and marginally at bfloat16, so leaving those two tolerances alone is justified rather than merely untested.

Per-row disagreement with the standard's own inverse relations comes out R 2.773e-4, G 2.428e-4, B 5.231e-4 — matching the reworked `_INVERSE_ATOL` comment literally, and the tightest consumer (`test_unit_matches_the_reference_relations`) uses 77% of the new 6e-4 budget. That is the right amount of headroom for an assertion whose whole job is to pin a derived constant.

Ran locally (cpu, torch 2.9.1): `tests/color/test_yuv.py` 212 passed at `--dtype=float32,float64`; yuv doctests 12 passed. Blast radius per `AGENTS.md`: `grep -rnE "#4044|issues/4044" kornia/` is empty, no dangling references to the deleted `_WORST_CORNER_ATOL` or `test_wart_yuv_to_rgb_kernel_is_not_the_inverse_4044`, and nothing outside `tests/color/test_yuv.py` consumes `yuv_to_rgb` values. The surviving `tests/` hits are the pins, which the guide explicitly permits.

The `test_wart_integer_input_truncates_the_kernel_4053` pin is well-judged: the two new near-zero *negative* literals are exactly the interaction between this fix and #4053, the int64 kernel really does truncate to `[[1,0,1],[1,0,0],[1,2,0]]` — bit-identical to what shipped — and picking `U = V = 1` rather than `0` keeps the cell from passing vacuously.

### Findings

**1. `CHANGELOG.md` — the new entry contradicts itself (low).**

> The forward direction (`rgb_to_yuv`, `rgb_to_yuv420`, `rgb_to_yuv422`) is unchanged, **and so is** the agreement of the inverse with the standard's own relations, **which improves** from `1.54e-3` to `5.24e-4`.

"and so is ... which improves" asserts both at once. The PR body's own phrasing ("also improves") is the correct one.

Separately, that 1.54e-3 → 5.24e-4 is the max over channels. Per channel the picture is mixed: R **1.543e-4 → 2.773e-4 (~1.8x worse)**, G 6.705e-4 → 2.428e-4, B 1.535e-3 → 5.231e-4. The test comment shows all three rows, so it is transparent there — but a reader of the changelog, or of the PR body's "this is not a trade of accuracy for invertibility", would not expect R to move the wrong way. Worth one clause:

> ...is unchanged. Agreement of the inverse with the standard's own relations improves overall, from `1.54e-3` to `5.24e-4`, though R alone moves the other way (`1.54e-4` → `2.77e-4`).

**2. `CHANGELOG.md:82` — the #4045 entry is now stale (low).**

Three entries below the new one, inside the *same* `## Unreleased` block, the #4045 doc entry still reads in the present tense:

> Every YUV-to-RGB form -- the three functions and the three `nn.Module` classes -- also gained a `.. warning::` naming a known defect: the round trip through `rgb_to_yuv` loses up to 1.356e-3 in float64 because the two kernels are rounded separately rather than inverted, with `float16` and `bfloat16` adding their own representation error on top (#4044)

As of this PR neither those warnings nor that defect ship. Anyone reading the release notes for this version gets a live-sounding defect report for something fixed three entries above, with #4044 described as open. `kornia/` was swept correctly; `CHANGELOG.md` was not. Trimming the #4044 clause from the #4045 entry and keeping its #4050 half (still live) resolves it.

There is a fair counterargument that a changelog is a historical record of what each PR did — but these entries are all still `Unreleased` and will be read as one set of notes, so the present tense is what lands.

### Not verified

No CUDA device here, so the `F.conv2d` branch of `_apply_linear_transformation` was exercised only through MPS. Setting the float32 round-trip bound two orders above the measured cpu floor rather than one, and saying so in the comment, is the right way to handle that.

Neither finding touches the code, so this is fine to merge once the changelog wording is fixed.

ducha-aiki:
Both changelog findings addressed in `b10499b8`. No code or test change.

**1. The self-contradicting clause.** Re-derived the per-channel numbers before writing them in. Disagreement with the exact inverse of the standard's own relations, bounded row by row over the documented YUV domain (`|Δu|·0.436 + |Δv|·0.615`):

| channel | old kernel | new kernel |
|---|---|---|
| R | 1.543e-4 | 2.773e-4 |
| G | 6.705e-4 | 2.428e-4 |
| B | 1.535e-3 | 5.231e-4 |

So the `1.54e-3 → 5.24e-4` figure is a max over channels and R does move the wrong way. The entry now reads:

> ...is unchanged. Agreement of the inverse with the standard's own relations improves overall, from `1.54e-3` to `5.24e-4`, though R alone moves the other way (`1.54e-4` to `2.77e-4`).

**2. The stale #4045 entry.** Its `.. warning::` sentence covered two defects — #4044's round trip and #4050's chroma-height validation. The #4044 half is dropped and the #4050 half kept, since `grep -rnE "#4050|issues/4050" kornia/` still returns the two live warning blocks at `kornia/color/yuv.py:263` and `:546`. The sentence now names #4050 as the defect that was documented and points at the entry above for the one that has since been fixed, rather than describing removed warnings in the present tense.

`pixi run pre-commit-all`: all 13 hooks pass.

On the Copilot inline note about `_INVERSE_ATOL` serving as both `atol` and `rtol` — that was already fixed on the head it reviewed: both consumers (`tests/color/test_yuv.py:559` and `:569`) pass `rtol=0.0`, so the constant is the absolute bound it claims to be.

PR #4055, opened 2026-08-28
fix(color): validate chroma plane height in yuv422_to_rgb

### LINKED ISSUE #4050: [Bug]: yuv422_to_rgb does not validate the chroma plane height — mismatches escape as a bare RuntimeError from torch.cat
### What happened?

`yuv422_to_rgb` validates that the chroma plane's **width** is half the luma's, but never checks its **height**. A chroma plane with the wrong height therefore reaches the internal `torch.cat` and dies there with a bare `RuntimeError` naming an internal tensor index, instead of the `ShapeError` every other malformed input to this function gets.

The 4:2:0 twin, `yuv420_to_rgb`, checks both axes and reports the mistake properly. The two guards sit ten lines apart:

```python
# yuv420_to_rgb  -- kornia/color/yuv.py
if (
    len(imageuv.shape) < 2
    or len(imagey.shape) < 2
    or imagey.shape[-2] / imageuv.shape[-2] != 2   # <-- height relation checked
):
    raise ShapeError(...)

# yuv422_to_rgb  -- kornia/color/yuv.py
    raise ShapeError(...)                          # <-- no height relation at all
```

`yuv422_to_rgb` expects chroma of shape `(*, 2, H, W/2)`, so the correct relation is `imagey.shape[-2] == imageuv.shape[-2]` — chroma keeps full height, since 4:2:2 subsamples width only.

The most likely way to hit this is the obvious one: feeding 4:2:0 chroma to the 4:2:2 function. That is a user mistake the library should name, and 4:2:0 names it:

```
>>> kornia.color.yuv422_to_rgb(torch.rand(2, 1, 720, 1280), torch.rand(2, 2, 360, 640))
RuntimeError: Sizes of tensors must match except in dimension 1. Expected size 720 but got size 360 for tensor number 1 in the list.

>>> kornia.color.yuv420_to_rgb(torch.rand(1, 4, 4), torch.rand(2, 4, 2))
ShapeError: Input imageuv H&W must be half the size of the luma plane. Got torch.Size([1, 4, 4]) and torch.Size([2, 4, 2])
```

Beyond the wording, the *type* differs in a way callers can observe: `kornia.core.exceptions.ShapeError` derives from `BaseError`, not from `RuntimeError`, so code that catches kornia's shape errors to validate user input will not catch this one.

This is a diagnostics defect only — there is no silently wrong result. `torch.cat` rejects every mismatched height, so the conversion always fails; it just fails with the wrong exception.

Present since the function was introduced in #1360, and unchanged on `main` (`35555b55`).

### Minimal reproduction

```python
import torch
import kornia

# chroma height 2 against luma height 4. 4:2:2 requires chroma height == luma height,
# and the width relation (4 / 2 == 2) is satisfied, so the only guard passes.
kornia.color.yuv422_to_rgb(torch.rand(1, 4, 4), torch.rand(2, 2, 2))
# RuntimeError: Sizes of tensors must match except in dimension 0.
#               Expected size 4 but got size 2 for tensor number 1 in the list.

# the 4:2:0 twin reports the equivalent mistake properly
# ShapeError: Input imageuv H&W must be half the size of the luma plane.
#             Got torch.Size([1, 4, 4]) and torch.Size([2, 4, 2])
```

The `Yuv422ToRgb` module surface behaves identically, since it just forwards.

### Environment

- Kornia version: 0.9.0rc1 (`main` at `35555b55`), also present in every release since #1360
- PyTorch version: 2.9.1
- Python version: 3.11.14
- OS / device (CPU, CUDA, ROCm, MPS): macOS 26.5.1 arm64, CPU / MPS

### Anything else?

**Suggested fix** — mirror the 4:2:0 guard, adjusted for the axis 4:2:2 actually subsamples:

```python
if (
    len(imageuv.shape) < 2
    or len(imagey.shape) < 2
    or imagey.shape[-2] != imageuv.shape[-2]
):
    raise ShapeError(
        f"Input imageuv W must be half the size of the luma plane and H must match it. "
        f"Got {imagey.shape} and {imageuv.shape}"
    )
```

The current behavior is **already pinned** by `TestYuv422ToRgb::test_wart_chroma_height_is_unchecked` in #4046, which asserts the bare `RuntimeError`. Because `ShapeError` does not derive from `RuntimeError`, adding the guard above flips that test — so whoever fixes this should delete the wart test and add the raise-site to `TestYuv422ToRgb::test_exception` alongside the other `ShapeError` cases. #4046 also leaves a comment at that test pointing here.

Two adjacent things deliberately **not** claimed here, to keep this issue to one defect:

* `rgb_to_yuv422` and `yuv422_to_rgb` both reject odd **height** as well as odd width, although 4:2:2 subsamples width only. That guard may be over-strict, but it is a separate question and changing it is a behavior change; #4046 pins the current behavior rather than blessing it.
* Neither function validates the leading/batch dimensions either. `torch.cat` catches those too, with the same class of message — worth folding into the same fix if someone takes it.

Himesh-rupchandani:

`yuv422_to_rgb` validates that the chroma plane's width is half the luma's, but never checks its height. 4:2:2 subsamples width only, so chroma should keep the full luma height. A chroma plane with the wrong height reached the internal `torch.cat` and failed there with a bare `RuntimeError` instead of the `ShapeError` every other malformed input to this function gets.

The 4:2:0 twin `yuv420_to_rgb` already checks both axes. This adds the same height check to `yuv422_to_rgb` and updates the error message accordingly.

Diagnostics only — `torch.cat` already rejected every mismatched height, so no previously working input changes behaviour.

Closes #4050

## What did you check?

Before the change:

After the change the same call raises `ShapeError`. This matters because `ShapeError` derives from `BaseError`, not `RuntimeError`, so code that catches kornia's shape errors to validate user input did not catch this one.

I also checked that valid 4:2:2 input still converts, and that the existing width guard and the 4:2:0 path are unaffected.

Added a regression test that fails without the fix. `pytest tests/color/test_yuv.py` — 11 passed locally.

**Note on test placement:** the issue asks for `TestYuv422ToRgb::test_wart_chroma_height_is_unchecked` to be deleted and the raise-site added to `TestYuv422ToRgb::test_exception`. Those tests come from #4046, which is still open, so they are not on `main` yet. I've added a standalone regression test for now and am happy to rebase and fold it into `test_exception` once #4046 lands.

## How did AI help?

I used an AI assistant to help investigate the issue, compare the 4:2:0 and 4:2:2 guards, and check the wording of this description. I reproduced the behaviour and ran the tests myself, and I understand the change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,15 +50,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   4:2:0 twin `yuv420_to_rgb` checks both axes. This is observable to callers on the error path: `ShapeError` derives
   from `BaseError`, not `RuntimeError`, so code catching `RuntimeError` around `yuv422_to_rgb` to handle a malformed
   chroma plane will stop catching it. No previously working input changes behaviour — `torch.cat` already rejected
-<<<<<<< HEAD
-  every mismatched height. A zero-width chroma plane whose height *also* differs now reaches the guard, since the
-  new clause short-circuits the width division; the matching-height case still raises `ZeroDivisionError` and
-  remains open as #4056.
-=======
   every mismatched height. A zero-width chroma plane whose height *also* differs now reaches the guard, since the new
   clause short-circuits the width division; the matching-height case still raises
   `ZeroDivisionError` and remains open as #4056.
->>>>>>> 0e3bcaf6 (docs: scope the zero-width claim to the mismatched-height case)
 
 * Make `yuv_to_rgb` the exact inverse of `rgb_to_yuv`, so an RGB → YUV → RGB round trip is now limited only by the
   input dtype instead of losing up to `1.36e-3` (in B, at `rgb = (1, 1, 0)`) at every precision, `float64` included
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,11 +37,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   169 passed under `--dtype=float16,bfloat16`.
 
 * Raise `NotImplementedError` instead of `RuntimeError` for an integral `src` on the empty-`dsize` paths of
-  `warp_affine`, `warp_perspective`, `remap`, `warp_affine3d` and `warp_perspective3d`, matching what the non-empty path already raises from
-  `grid_sample` (#4031). The degenerate path had its own explicit guard that fired first with a different exception
-  type, so the exception a caller saw for the same invalid input depended on whether `dsize` had a zero dimension.
-  The message is unchanged. `NotImplementedError` derives from `RuntimeError`, so `except RuntimeError` around these
-  functions still catches it; only code matching on the exact type sees a difference.
+  `warp_affine`, `warp_perspective`, `remap`, `warp_affine3d` and `warp_perspective3d`, matching what the non-empty
+  path already raises from `grid_sample` (#4031). The degenerate path had its own explicit guard that fired first
+  with a different exception type, so the exception a caller saw for the same invalid input depended on whether
+  `dsize` had a zero dimension. The message is unchanged. `NotImplementedError` derives from `RuntimeError`, so
+  `except RuntimeError` around these functions still catches it; only code matching on the exact type sees a
+  difference.
+
+* Validate the chroma plane height in `yuv422_to_rgb`, so a chroma plane whose height does not match the luma now
+  raises `ShapeError` instead of reaching `torch.cat` and dying there with a bare `RuntimeError` (#4050). 4:2:2
+  subsamples width only, so chroma keeps the full luma height; the guard checked only the width ratio, where the
+  4:2:0 twin `yuv420_to_rgb` checks both axes. This is observable to callers on the error path: `ShapeError` derives
+  from `BaseError`, not `RuntimeError`, so code catching `RuntimeError` around `yuv422_to_rgb` to handle a malformed
+  chroma plane will stop catching it. No previously working input changes behaviour — `torch.cat` already rejected
+  every mismatched height. A zero-width chroma plane whose height *also* differs now reaches the guard, since the
+  new clause short-circuits the width division; the matching-height case still raises `ZeroDivisionError` and
+  remains open as #4056.
 
 * Make `yuv_to_rgb` the exact inverse of `rgb_to_yuv`, so an RGB → YUV → RGB round trip is now limited only by the
   input dtype instead of losing up to `1.36e-3` (in B, at `rgb = (1, 1, 0)`) at every precision, `float64` included
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,9 +50,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   4:2:0 twin `yuv420_to_rgb` checks both axes. This is observable to callers on the error path: `ShapeError` derives
   from `BaseError`, not `RuntimeError`, so code catching `RuntimeError` around `yuv422_to_rgb` to handle a malformed
   chroma plane will stop catching it. No previously working input changes behaviour — `torch.cat` already rejected
+<<<<<<< HEAD
   every mismatched height. A zero-width chroma plane whose height *also* differs now reaches the guard, since the
   new clause short-circuits the width division; the matching-height case still raises `ZeroDivisionError` and
   remains open as #4056.
+=======
+  every mismatched height. A zero-width chroma plane whose height *also* differs now reaches the guard, since the new
+  clause short-circuits the width division; the matching-height case still raises
+  `ZeroDivisionError` and remains open as #4056.
+>>>>>>> 0e3bcaf6 (docs: scope the zero-width claim to the mismatched-height case)
 
 * Make `yuv_to_rgb` the exact inverse of `rgb_to_yuv`, so an RGB → YUV → RGB round trip is now limited only by the
   input dtype instead of losing up to `1.36e-3` (in B, at `rgb = (1, 1, 0)`) at every precision, `float64` included
--- a/kornia/color/yuv.py
+++ b/kornia/color/yuv.py
@@ -256,12 +256,6 @@ def yuv422_to_rgb(imagey: torch.Tensor, imageuv: torch.Tensor) -> torch.Tensor:
     `BT.470-5 <https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.470-5-199802-S!!PDF-E.pdf>`_, Table 2,
     items 2.5 and 2.6).
 
-    .. warning::
-        Only the chroma *width* is validated against the luma plane. A chroma plane of the
-        wrong height escapes the ``ShapeError`` below and surfaces as a bare ``RuntimeError``
-        from ``torch.cat``, where :func:`~kornia.color.yuv420_to_rgb` checks both. Tracked in
-        `#4050 <https://github.com/kornia/kornia/issues/4050>`_.
-
     Args:
         imagey: Y (luma) Image plane to be converted to RGB with shape :math:`(*, 1, H, W)`.
         imageuv: UV (chroma) Image planes to be converted to RGB with shape :math:`(*, 2, H, W/2)`.
@@ -282,9 +276,15 @@ def yuv422_to_rgb(imagey: torch.Tensor, imageuv: torch.Tensor) -> torch.Tensor:
     if len(imagey.shape) < 2 or imagey.shape[-2] % 2 == 1 or imagey.shape[-1] % 2 == 1:
         raise ShapeError(f"Input H&W must be evenly divisible by 2. Got {imagey.shape}")
 
-    if len(imageuv.shape) < 2 or len(imagey.shape) < 2 or imagey.shape[-1] / imageuv.shape[-1] != 2:
+    if (
+        len(imageuv.shape) < 2
+        or len(imagey.shape) < 2
+        or imagey.shape[-2] != imageuv.shape[-2]
+        or imagey.shape[-1] / imageuv.shape[-1] != 2
+    ):
         raise ShapeError(
-            f"Input imageuv W must be half the size of the luma plane. Got {imagey.shape} and {imageuv.shape}"
+            f"Input imageuv H must match the luma plane and W must be half its size. "
+            f"Got {imagey.shape} and {imageuv.shape}"
         )
 
     # first upsample
@@ -538,12 +538,6 @@ class Yuv422ToRgb(nn.Module):
     YUV formula follows M/PAL values (see
     `BT.470-5 <https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.470-5-199802-S!!PDF-E.pdf>`_, Table 2,
     items 2.5 and 2.6).
-
-    .. warning::
-        Only the chroma *width* is validated against the luma plane. A chroma plane of the
-        wrong height escapes the ``ShapeError`` and surfaces as a bare ``RuntimeError`` from
-        ``torch.cat``, where :class:`~kornia.color.Yuv420ToRgb` checks both. Tracked in
-        `#4050 <https://github.com/kornia/kornia/issues/4050>`_.
 
     Returns:
         RGB version of the image.
--- a/tests/color/test_yuv.py
+++ b/tests/color/test_yuv.py
@@ -815,22 +815,16 @@ class TestYuv422ToRgb(BaseTester):
 
         # Luma must be single-channel: rejected by the channel slot of the shape spec, which the
         # rank case above does not reach. Note this is *not* a width-relation case -- 2/1 == 2
-        # holds; the height relation it looks like it covers is unchecked, see kornia#4050.
+        # holds.
         with pytest.raises(ShapeError):
             imgy = torch.ones(2, 2, 2, device=device, dtype=dtype)
             imguv = torch.ones(2, 1, 1, device=device, dtype=dtype)
             kornia.color.yuv422_to_rgb(imgy, imguv)
 
-    def test_wart_chroma_height_is_unchecked_4050(self, device, dtype):
-        # Wart pin for kornia#4050. yuv422_to_rgb validates only the *width* ratio, so a chroma
-        # plane whose height does not match the luma reaches ``torch.cat`` and dies there with a
-        # bare RuntimeError instead of the ShapeError every other malformed input gets -- the
-        # 4:2:0 twin checks both axes. ShapeError does not derive from RuntimeError, so adding
-        # the missing guard flips this test: delete it then, and move the case up into
-        # test_exception alongside the other ShapeError raise-sites.
-        imgy = torch.ones(1, 2, 2, device=device, dtype=dtype)
-        imguv = torch.ones(2, 1, 1, device=device, dtype=dtype)
-        with pytest.raises(RuntimeError, match="Sizes of tensors must match"):
+        # 4:2:2 subsamples width only, so chroma keeps the full luma height.
+        with pytest.raises(ShapeError):
+            imgy = torch.ones(1, 2, 2, device=device, dtype=dtype)
+            imguv = torch.ones(2, 1, 1, device=device, dtype=dtype)
             kornia.color.yuv422_to_rgb(imgy, imguv)
 
     @pytest.mark.parametrize("name", list(_REFERENCE_COLORS))


ducha-aiki:
**Approve the change — the fix is correct and the diagnostics-only claim holds under a 38,416-case sweep. Two blockers before merge, both mechanical: it has to be rebased after #4046 and #4048, and the rebase has to clean up *two* `#4050` references in the test file, not one.**

Thanks for a well-argued PR. Three things you did that made this quick to review: you diffed against the 4:2:0 twin instead of inventing a guard, you stated up front that this is diagnostics-only rather than leaving me to infer it, and you flagged the `test_exception` / #4046 collision yourself before anyone asked. That last one is the part contributors usually miss.

Everything below is measured against `pr/4055` head `ab659794`, base `a61e4102`, torch 2.9.1 / py3.11.14 / CPU.

---

## 1. The central claim survives — 0 regressions in 38,416 shape combinations

Your claim is that `torch.cat` already rejected every mismatched height, so nothing that previously returned a tensor now raises. That is the whole basis for landing this as an ordinary bugfix, so I went looking for a counterexample rather than reasoning about it.

I swept the cross product of luma and chroma shapes — heights and widths in `{0,1,2,3,4,6,8}`, leading dims in `{(), (1,), (2,), (3,)}` on each side independently (so rank-3 vs rank-4 mixes, batch mismatches, chroma taller than luma, and zero-sized dims are all in there) — and ran the identical script against `main`'s kornia and against your branch's:

```
$ env -C $SP PYTHONPATH=$SP/pr4055 python sweep.py out_pr.txt
kornia from: .../pr4055/kornia/__init__.py
total cases: 38416
OK cases: 80

$ env -C $SP python sweep.py out_main.txt
kornia from: /Users/.../kornia/__init__.py
total cases: 38416
OK cases: 80
```

Diffing the two outcome tables case by case:

```
cases where main returned a tensor and PR raises : 0
cases where main raised and PR returns a tensor  : 0
cases OK in both but with a different output shape: 0

  OK-on-PR cases with mismatched height:   0
  OK-on-main cases with mismatched height: 0
```

Claim verified. The reason it is airtight, for the record: `torch.cat` requires exact equality on *every* non-concat dimension, so "cat succeeded" already implies `imagey.shape[-2] == imageuv.shape[-2]`, which is exactly the negation of your new clause. There is no gap for a counterexample to live in.

The headline improvement, same script, both revisions:

```
##### MAIN #####
  issue repro  (4:2:0 chroma into 4:2:2) y=(1, 4, 4) uv=(2, 2, 2)
      RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 4 but got size 2 ...
  realistic    720p 4:2:0 chroma         y=(2, 1, 720, 1280) uv=(2, 2, 360, 640)
      RuntimeError: Sizes of tensors must match except in dimension 1. Expected size 720 but got size 360 ...
  chroma TALLER than luma                y=(2, 1, 4, 6) uv=(2, 2, 8, 3)
      RuntimeError: Sizes of tensors must match except in dimension 1. Expected size 4 but got size 8 ...
  VALID 4:2:2                            y=(2, 1, 4, 6) uv=(2, 2, 4, 3)
      OK -> (2, 3, 4, 6)

##### PR #4055 #####
  issue repro  (4:2:0 chroma into 4:2:2) y=(1, 4, 4) uv=(2, 2, 2)
      ShapeError: Input imageuv H must match the luma plane and W must be half its size. Got torch.Size([1, 4, 4]) and torch.Size([2, 2, 2])
  realistic    720p 4:2:0 chroma         y=(2, 1, 720, 1280) uv=(2, 2, 360, 640)
      ShapeError: Input imageuv H must match the luma plane and W must be half its size. Got torch.Size([2, 1, 720, 1280]) and torch.Size([2, 2, 360, 640])
  chroma TALLER than luma                y=(2, 1, 4, 6) uv=(2, 2, 8, 3)
      ShapeError: Input imageuv H must match the luma plane and W must be half its size. Got torch.Size([2, 1, 4, 6]) and torch.Size([2, 2, 8, 3])
  VALID 4:2:2                            y=(2, 1, 4, 6) uv=(2, 2, 4, 3)
      OK -> (2, 3, 4, 6)
```

## 2. Clause ordering — an unadvertised bonus, and one leftover

Putting the `!=` height clause *before* the existing `imagey.shape[-1] / imageuv.shape[-1] != 2` matters more than it looks, because that `/` is a Python int division that raises `ZeroDivisionError` on a zero-width chroma plane. Your clause short-circuits it. Across the sweep:

```
cases where the EXCEPTION CLASS changed: 4320
  ZeroDivisionError -> ShapeError: 2400
  RuntimeError      -> ShapeError: 1920
```

So 2400 combinations that used to escape the guard as a bare `ZeroDivisionError: division by zero` now get the intended `ShapeError`. Nothing moved in the other direction, and no previously-raising case got a *less* accurate message — the reworded string states the full contract and prints both shapes, so it is correct for a height-only violation, a width-only violation, and both at once.

Two leftovers, both pre-existing and both correctly out of scope for this PR:

```
=== ZeroDivisionError still reachable on the PR? ===
  count: 400          (main: 2800)
   y= (1, 4, 6)  uv= (2, 4, 0)     # chroma width 0, height matches -> 6 / 0

=== RuntimeError still reachable on the PR? ===
  count: 240
   ('Sizes of tensors must match except in dimension 1', 120)
   ('Tensors must have same number of dimensions: got 3 and 4', 120)
```

The remaining 240 `RuntimeError`s are all leading/batch-dim and rank mismatches — precisely the case #4050 explicitly declined to claim, so leaving them is the right call. The `ZeroDivisionError` is a genuine sibling defect of #4050 (a shape guard raising the wrong exception type) and the 4:2:0 twin has it too:

```
  yuv420_to_rgb chroma W=0: ZeroDivisionError: division by zero
  yuv420_to_rgb chroma H=0: ZeroDivisionError: division by zero
  yuv422_to_rgb chroma W=0: ZeroDivisionError: division by zero
```

Please don't fold that into this PR — it wants its own issue against both twins. Mentioning it only so it is on the record.

## 3. Consistency with the 4:2:0 twin — good

Message shape matches (`Input imageuv ... Got {imagey.shape} and {imageuv.shape}`), and the semantics correctly diverge where the formats do: `!=` for height because 4:2:2 keeps full chroma height, `/ != 2` for width. That is the right kind of divergence.

TorchScript is unaffected as a capability — the guard scripts fine as a function and through the `Yuv422ToRgb` module on both revisions:

```
### PR head: TorchScript ###
  scripted OK -> (2, 3, 4, 6)
  module scripted OK -> (2, 3, 4, 6)
```

One nuance worth knowing rather than acting on: under `torch.jit.script` a mismatched height surfaces as `torch.jit.Error` on your branch where it was `RuntimeError` on `main`, because the raise now happens in scripted Python instead of inside `cat`. `torch.jit.Error` is not a `RuntimeError` subclass either (`[Error, Exception, BaseException, object]`), so scripted callers see the same class of change as eager ones. Consistent with the twin, so nothing to change.

## 4. Test placement — fold it in on the rebase

`test_yuv422_to_rgb_chroma_height_mismatch` is a correct regression test and it does fail without the fix. But as a bare module-level function with hardcoded `torch.rand` it ignores the injected `device`/`dtype` fixtures, so it only ever runs on CPU float32 — the guard never gets exercised on the CUDA, MPS or float64 matrices that CI actually runs, and every other test in this file uses the fixtures. It should live in `TestYuv422ToRgb::test_exception`.

Your reasoning for the stopgap is sound though — that class genuinely does not exist on `main` yet, so there is nowhere to put it today. Fold it in at the rebase.

## 5. Merge order — confirmed from the diffs

Checked against the actual PR heads rather than assumption:

```
$ git merge-base --is-ancestor pr4046-head pr4048-head && echo "YES"
YES: 4046 is an ancestor of 4048
```

So #4048 is stacked on #4046 and the #4046 -> #4048 order is already forced. Replaying your two commits onto #4048:

```
$ git cherry-pick c6bc8acd     # the source fix
Auto-merging kornia/color/yuv.py
[detached HEAD 378dd222] fix(color): validate chroma plane height in yuv422_to_rgb
 1 file changed, 8 insertions(+), 2 deletions(-)

$ git cherry-pick ab659794     # the test
Auto-merging tests/color/test_yuv.py
CONFLICT (content): Merge conflict in tests/color/test_yuv.py
```

**The source fix rebases cleanly.** #4046 only touches the docstring above your guard and #4048 only touches `yuv_to_rgb`'s kernel constants, so neither collides with the guard lines. **Only the test commit conflicts**, because #4046 rewrites `tests/color/test_yuv.py` wholesale and your test appends at what is now the old end of the file. Dropping the test commit and running the suite confirms the flip you predicted:

```
FAILED tests/color/test_yuv.py::TestYuv422ToRgb::test_wart_chroma_height_is_unchecked_4050[cpu-float32]
  - kornia.core.exceptions.ShapeError: Input imageuv H must match the luma plane ...
======================== 1 failed, 106 passed in 0.59s =========================
```

## 6. Pre-merge `#4050` grep — there are TWO references to clean up, not one

Per AGENTS.md, closing an issue means the tracking references have to go with it. On your branch as it stands the grep is clean, which is correct — the pins are not on `main` yet:

```
$ git grep -nE "#4050|issues/4050" pr4055-head -- kornia/ tests/
(exit 1)
```

But reasoning about the *combined* end state, #4046 lands two of them:

```
$ git grep -nE "#4050|issues/4050" pr4048-head -- kornia/ tests/
tests/color/test_yuv.py:730:  # holds; the height relation it looks like it covers is unchecked, see kornia#4050.
tests/color/test_yuv.py:737:  # Wart pin for kornia#4050. yuv422_to_rgb validates only the *width* ratio, so a chroma
```

Line 737 is the wart pin everyone has been talking about. **Line 730 is a second one that has not been mentioned anywhere**, and it behaves differently: it is a comment inside `test_exception` itself, and its assertion keeps passing after your fix — I confirmed that case still raises from the shape spec, not from your new guard, on both revisions:

```
-- on main --             ShapeError raised at core/check.py:209
-- with fix (on #4048) -- ShapeError raised at core/check.py:209
```

So it is prose that silently goes stale rather than a test that flips. It says the height relation "is unchecked" — which stops being true the moment this merges. Drop that clause; keep the rest of the comment, since the `2/1 == 2` observation it makes is still the reason that case is interesting.

I built the full end state to check it holds together — source fix on #4048, wart pin deleted, the case folded into `test_exception`, line 730's stale clause trimmed:

```
$ grep -rnE "#4050|issues/4050" kornia/ tests/
(grep exit 1)

$ pytest tests/color/test_yuv.py -q --device=cpu --dtype=float32,float64
============================= 210 passed in 0.52s ==============================
```

## 7. One addition: a CHANGELOG entry

`CHANGELOG.md` has a live `## Unreleased` section with a `### Bug fixes` heading, and both #4046 and #4048 add entries there. This PR adds none, and it should, because the change is observable to callers on the error path: `ShapeError` derives from `BaseError`, not `RuntimeError` (`[ShapeError, BaseError, Exception, BaseException, object]`), so anyone catching `RuntimeError` around `yuv422_to_rgb` to handle a bad chroma plane will stop catching it. One entry under `### Bug fixes` naming that transition is enough.

---

## Concrete next steps, in order

1. Wait for #4046 and #4048 to merge (they are stacked, so they will land in that order).
2. `git rebase origin/main` — the source commit `c6bc8acd` will replay cleanly; expect the conflict only in `tests/color/test_yuv.py`.
3. Resolve it by **dropping your standalone `test_yuv422_to_rgb_chroma_height_mismatch`** entirely and instead:
   - delete `TestYuv422ToRgb::test_wart_chroma_height_is_unchecked_4050`;
   - add its `imgy = torch.ones(1, 2, 2, device=device, dtype=dtype)` / `imguv = torch.ones(2, 1, 1, device=device, dtype=dtype)` case to `TestYuv422ToRgb::test_exception` as a `pytest.raises(ShapeError)` block, using the `device`/`dtype` fixtures, with a short comment saying 4:2:2 subsamples width only so chroma keeps the full luma height;
   - trim the now-false `see kornia#4050` clause from the comment on the last case already in `test_exception`.
4. Add the `### Bug fixes` CHANGELOG entry.
5. Re-run `grep -rnE "#4050|issues/4050" kornia/ tests/` and confirm it comes back empty, plus `pytest tests/color/test_yuv.py --device=cpu --dtype=float32,float64`.

No changes needed to `kornia/color/yuv.py` — the source diff is right as written, `ruff check` and `ruff format --diff` are clean on it, the 12 doctests in the module pass, and CI is 33 check runs plus 2 statuses, all green.

Unrelated heads-up so it does not confuse you during the rebase: `TestRgbToYuv::test_round_trip_rgb_yuv_rgb` is flaky on `main` today (2 failures in 12 runs, from random input) and has nothing to do with your change — it is the defect #4048 fixes, and it will stop flaking once that lands.

Himesh-rupchandani:
Rebased onto main now that #4046 and #4048 have landed.

- Dropped the standalone test and folded the case into `TestYuv422ToRgb::test_exception`,
  using the `device`/`dtype` fixtures so it runs on the full matrix.
- Deleted `test_wart_chroma_height_is_unchecked_4050`.
- Trimmed the stale `see kornia#4050` clause in `test_exception`.
- Added the `### Bug fixes` CHANGELOG entry.

One thing beyond your list: #4046 also added two `.. warning::` blocks in `kornia/color/yuv.py`
(on `yuv422_to_rgb` and `Yuv422ToRgb`) that document the height as unchecked and link #4050.
Those go stale for the same reason, so I removed them too — four references cleaned up rather
than two.

`grep -rnE "#4050|issues/4050" kornia/ tests/` is empty, and `pytest tests/color/test_yuv.py`
gives 106 passed locally.

ducha-aiki:
**Delta re-review of `61c30e3e` — verdict: APPROVE, with one CHANGELOG sentence to correct.** Both mechanical blockers from my 2026-08-28 review are done, the code is right, and I verified the diagnostics-only claim over a full shape sweep. The one thing I am asking for is prose, not code.

Reviewed at head `61c30e3e`, merge-base `6fa75d64`, CPU, torch 2.9.1 (the repo `.venv`), in clean worktrees of both.

## Both blockers cleared

**Rebase.** Merge-base is `6fa75d64`, which is after both #4046 (2026-08-28 16:00) and #4048 (17:35). Clean.

**The `#4050` references.** All gone, and correctly — the two `.. warning::` blocks removed from `kornia/color/yuv.py` (function *and* `Yuv422ToRgb` class), and the wart pin `test_wart_chroma_height_is_unchecked_4050` converted into a `ShapeError` assertion inside `test_exception` rather than deleted. The AGENTS.md pre-merge gate is clean:

```console
$ grep -rnE "#4050|issues/4050" kornia/ tests/
  (no output)
```

That is the full protocol executed: prose retired, strict pin inverted, no orphan reference left behind.

## The guard is correct, and it really is diagnostics-only

I swept **2592** shape combinations (`H`, `W` ∈ {0,1,2,3,4,6} for both planes × luma channel ∈ {1,2}) on both sides, with deterministic per-case seeding so results are comparable:

```
worked -> now raises          : 0
worked -> different result    : 0
ZeroDivisionError -> ShapeError : 80
RuntimeError      -> ShapeError : 60
```

So no previously working input changes in either direction — the change is confined to the error path, exactly as the CHANGELOG says. The 60 `RuntimeError → ShapeError` conversions are the #4050 fix itself.

The flipped pin discriminates:

```console
$ # merge-base worktree, PR's test file copied in
FAILED tests/color/test_yuv.py::TestYuv422ToRgb::test_exception[cpu-float32]
  - RuntimeError: Sizes of tensors must match except in dimension 0.
1 failed, 34 passed

$ # head
tests/color/test_yuv.py -> 210 passed  (float32, float64)
```

## The one ask — the CHANGELOG's zero-width sentence is wrong

> "A zero-width chroma plane that previously escaped as `ZeroDivisionError` now also reaches the guard, since the new clause short-circuits the width division."

That holds only when the height *also* mismatches. When the heights match — which is the case in **your own #4056 repro** — the new clause evaluates `False`, so nothing short-circuits and the division still runs:

```
                                                  merge-base 6fa75d64        head 61c30e3e
zero-width chroma, height MATCHES (#4056 repro)   ZeroDivisionError    ->    ZeroDivisionError
  y=(1,4,6) uv=(2,4,0)
zero-width chroma, height differs                 ZeroDivisionError    ->    ShapeError
  y=(1,4,6) uv=(2,2,0)
```

The sweep agrees: 80 of the 96 `ZeroDivisionError` cases convert, and **16 survive** — every one of them a matching-height, zero-width chroma plane.

Two ways to resolve it, your pick:

1. **Correct the sentence** (my default, keeps the PR at its #4050 scope): say that a zero-width chroma plane whose height *also* differs now reaches the guard, and that the matching-height case remains open as #4056.
2. **Absorb the `yuv422` half of #4056**, which is one token — swap the division for a multiplication:

   ```python
   or imagey.shape[-1] != imageuv.shape[-1] * 2
   ```

   Equivalent for every non-zero width, and it removes the `ZeroDivisionError` outright rather than short-circuiting past it. `yuv420_to_rgb` would still need its own fix, so #4056 stays open either way. If you take this route, update the CHANGELOG to claim it and add the matching-height zero-width case to `test_exception`.

I lean toward (1) since #4056 is yours and you clearly planned it separately — but (2) is small enough that either is fine. Not blocking on which you choose, only on the claim matching the code.

Good, careful PR otherwise; the pin conversion in particular was done exactly right.

Himesh-rupchandani:
Taken option 1 — you're right, the claim only holds when the height also differs. The
matching-height zero-width case is unchanged and stays with #4056, which is where I'd
rather fix it since yuv420_to_rgb needs the same treatment.

ducha-aiki (approved):
**Delta re-review of `6e4a9ba9` — APPROVE.**

Option 1 taken as offered: the CHANGELOG now says the zero-width plane reaches the guard only when the height *also* differs, and names #4056 as still open. That matches the measured behaviour from the last round.

Re-verified at this head (`.venv`, torch 2.9.1, CPU):

```
tests/color/test_yuv.py --device=cpu --dtype=float32,float64   210 passed
grep -rnE "#4050|issues/4050" kornia/                            0 hits
```

Clean merge against current `main`. Nothing left open on my side. Note for whoever merges #4111 (Marchematics, closes #4056): it edits the same guard lines, so it rebases on top of this one.

Issue #4067, opened 2026-08-28
Docstring errors in laf.py normalize/denormalize and an unused match_fginn argument


ducha-aiki:
### Describe the bug

Documentation-only defects found while auditing `kornia/feature`. No behaviour change involved.

#### `normalize_laf` / `denormalize_laf` docstrings — `kornia/feature/laf.py`

https://github.com/kornia/kornia/blob/main/kornia/feature/laf.py#L317-L334

1. **`:317`** — the summary line reads `"""De-F.normalize LAFs from scale to image scale.` The `F.` is a botched search-and-replace of `normalize` → `F.normalize` (the module imports `torch.nn.functional as F`). It should read *"Denormalize LAFs…"*.

2. **`:327`** — the worked example says the second row becomes

   ```
   [a21*MIN_SIZE a22*MIN_SIZE y*(W-1)]
   ```

   but the code multiplies the y translation by `h - 1`:

   ```python
   coef[0, 0, 0, 2] = wf   # w - 1
   coef[0, 0, 1, 2] = hf   # h - 1
   ```

   so it should be `y*(H-1)`. `normalize_laf` states the corresponding line correctly (`y/(H-1)`).

3. **`:323` and `:354`** — both docstrings write the source matrix as

   ```
   [a11 a21 x]
   [a21 a22 y]
   ```

   with `a21` in the top row. The top row is `[a11 a12 x]`.

4. **`:365`** — `normalize_laf`'s `Returns:` says *"the **denormalized** LAF … scale in image percentage (0, 1)"*. It returns the normalized LAF.

#### `match_fginn` accepts an unused argument — `kornia/feature/matching.py`

https://github.com/kornia/kornia/blob/main/kornia/feature/matching.py#L257

`lafs1` is a required positional parameter and is documented as `:math:`(1, B1, 2, 3)``, but the function body never reads it — only `lafs2` is used, via `xy2 = get_laf_center(lafs2)`. Either it should be used, or the docstring should record that it is accepted for API symmetry with `match_adalam` and ignored.

### Expected behavior

Docstrings that match the code.

Found during an audit of `kornia.feature`.

Posted on behalf of @ducha-aiki by Claude (Opus 5).

ducha-aiki:
One more of the same class, in `kornia/feature/siftdesc.py` — `SIFTDescriptor`'s documented input shape is wrong, and following it raises:

https://github.com/kornia/kornia/blob/main/kornia/feature/siftdesc.py#L104-L106

```
[...]

PR #4072, opened 2026-08-28
docs(feature): make the response doctests actually run, and fix laf/sift docstrings

### LINKED ISSUE #4066: harris/gftt/hessian_response doctests never execute and their published tensors are wrong
### Describe the bug

The `Example` blocks of `harris_response`, `gftt_response` and `hessian_response` never execute, and the tensors they display are wrong. `pixi run doctest` passes, so nothing flags it.

### Root cause

In each of the three docstrings the call line is missing its `>>>` prompt:

https://github.com/kornia/kornia/blob/main/kornia/feature/responses.py#L84-L94

```
        ... ]]])  # 1x1x7x7
        >>> # compute the response map
        harris_response(input, 0.04)
        tensor([[[[0.0012, 0.0039, 0.0020, 0.0000, 0.0020, 0.0039, 0.0012],
        ...
```

So `harris_response(input, 0.04)` and the tensor below it are parsed as the *expected output* of the preceding `>>> # compute the response map`. Python's `doctest.DocTestParser` then discards any example whose source is blank or a comment (`_IS_BLANK_OR_COMMENT`) — **together with its `want`** — so the entire expected tensor is silently thrown away:

```python
import doctest, kornia.feature.responses as R
for t in doctest.DocTestFinder().find(R):
    if t.examples:
        print(t.name, len(t.examples), "examples")
```

```
kornia.feature.responses.gftt_response     1 examples
kornia.feature.responses.harris_response   1 examples
kornia.feature.responses.hessian_response  1 examples
```

One example each — only the `input = torch.tensor(...)` assignment. The response function is never called.

In the rendered API docs the un-prompted line also shows up as part of the output block rather than as a call.

### The published numbers are wrong

Running each function on the input from its own docstring:

| function | documented first row | actual first row | max abs diff over the 7×7 block |
|---|---|---|---|
| `harris_response` | `0.0012, 0.0039, 0.0020, 0.0000, …` | `0.0042, 0.0054, 0.0035, 0.0006, …` | 0.0030 |
| `gftt_response` | `0.0155, 0.0334, 0.0194, 0.0000, …` | `0.0379, 0.0456, 0.0283, 0.0121, …` | 0.0276 |
| `hessian_response` | `0.0155, 0.0334, 0.0194, 0.0000, …` | `-0.0564, -0.0759, -0.0253, -0.0000, …` | 0.1093 |

`hessian_response`'s documented block is a **verbatim copy of `gftt_response`'s** — and the real values are negative, so even the sign is wrong.

```python
import torch, kornia.feature.responses as R
torch.set_printoptions(precision=4, sci_mode=False)
inp = torch.tensor([[[
   [0., 0., 0., 0., 0., 0., 0.],
   [0., 1., 1., 1., 1., 1., 0.],
   [0., 1., 1., 1., 1., 1., 0.],
   [0., 1., 1., 1., 1., 1., 0.],
   [0., 1., 1., 1., 1., 1., 0.],
   [0., 1., 1., 1., 1., 1., 0.],
   [0., 0., 0., 0., 0., 0., 0.],
]]])
print(R.harris_response(inp, 0.04))
print(R.gftt_response(inp))
print(R.hessian_response(inp))
```

The three `# TODO: Recompute doctest` markers sit directly below these blocks, at `responses.py:98`, `:181` and `:262`.

### Expected behavior

Restore the `>>>` prompt on the call line and paste the real output, so the examples run and the values are checked by `pixi run doctest` from then on. Purely a documentation fix — no behaviour change.

### Additional context

Worth a repo-wide grep for the same shape, since the failure mode is silent: a `>>>` comment line immediately followed by an un-prompted call swallows its own expected output.

ducha-aiki:
## What

Documentation-only fixes in `kornia/feature`, found during an audit of the module.

### 1. The response doctests never executed, and their numbers were wrong (#4066)

`harris_response`, `gftt_response` and `hessian_response` each had an `Example` block whose call line was missing its `>>>` prompt:

```
>>> # compute the response map
harris_response(input, 0.04)
tensor([[[[0.0012, ...
```

So the call and the tensor were parsed as the *expected output* of the preceding `>>> # compute the response map`. `doctest.DocTestParser` discards examples whose source is blank or a comment (`_IS_BLANK_OR_COMMENT`) **together with their `want`**, so the entire expected tensor was silently thrown away. `pytest --doctest-modules` collected exactly one example per function — the `input = torch.tensor(...)` assignment — and passed.

The published values were also stale:

| function | max abs diff vs real output |
|---|---|
| `harris_response` | 0.0030 |
| `gftt_response` | 0.0276 |
| `hessian_response` | 0.1093 |

`hessian_response`'s block was a verbatim copy of `gftt_response`'s, so its sign was wrong too.

This PR restores the prompt and pastes the real output. The examples now run:

```console
$ pytest --doctest-modules kornia/feature/responses.py -q
3 passed
```

and each function now collects 2 examples instead of 1.

The values are shown as `.round(decimals=4) + 0.0`. `hessian_response` has exact-zero columns whose raw repr is floating-point noise (`-3.4694e-18`, `-8.6736e-17`) that would not reproduce across platforms; rounding removes it and `+ 0.0` normalises signed zeros, so the doctest is stable rather than newly flaky.

The three `# TODO: Recompute doctest` markers are resolved and removed.

### 2. `laf.py` docstrings (#4067)

- `denormalize_laf`'s summary read `"""De-F.normalize LAFs from scale to image scale.` — a botched `normalize` → `F.normalize` search-and-replace.
- Its worked example said the y translation becomes `y*(W-1)`; the code uses `h - 1`.
- Both `normalize_laf` and `denormalize_laf` wrote the source matrix top row as `[a11 a21 x]` instead of `[a11 a12 x]`.
- `normalize_laf`'s `Returns:` said "the **denormalized** LAF".

### 3. `siftdesc.py` docstrings (#4067)

- `SIFTDescriptor` documented its input as `(B, 1, num_spatial_bins, num_spatial_bins)`, but `forward` requires `(B, 1, patch_size, patch_size)` — following the docs raises `Shape mismatch at dimension 2: expected 32, got 4`.
- Its `Returns:` line was truncated mid-sentence ("SIFT descriptor of the patches with shape.").
- `get_sift_bin_ksize_stride_pad`'s `ValueError` reached users with runs of stray whitespace from a broken implicit string concatenation:

  before
  ```
  Patch size 6 is incompatible with             requested number of spatial bins 4             for SIFT descriptor. Usually it happens when patch size is too small            for num_spatial_bins specified
  ```
  after
  ```
  Patch size 6 is incompatible with the requested number of spatial bins 4 for SIFT descriptor. Usually it happens when patch size is too small for num_spatial_bins specified
  ```

## Behaviour

None. Docstrings, one error-message string, and doctest expected output only.

## Verification

```console
$ pytest --doctest-modules kornia/feature/responses.py -q
3 passed

641 passed, 126 skipped

$ pixi run pre-commit-all
all hooks Passed
```

`pytest --doctest-modules kornia/feature/` has one further failure, `DescriptorMatcherWithSteerer`, which reproduces identically on unmodified `main` on this machine and is local MPS memory pressure, not a repo defect.

Closes #4066


ducha-aiki:
## Review — APPROVE

Docs-only, and every claim in the description reproduces. Verified independently rather than taken on trust:

- **The doctests now run.** `pytest --doctest-modules kornia/feature/responses.py kornia/feature/laf.py kornia/feature/siftdesc.py` → **12 passed**, 3 of them in `responses.py`. On `main` the parser really does drop the `want` along with the blank-or-comment source, so the tensors were never compared.
- **The recomputed values are right.** I recomputed all three response maps from the docstring input and they agree with what is pasted, and the same 4-decimal display comes out of `float32` and `float64` alike — so the rounding is not hiding a precision-dependent digit.
- **`y*(H-1)` is correct.** `denormalize_laf` sets `coef[0, 0, 1, 2] = hf` where `hf = float(h - 1)` (`laf.py:341-343`). The old `y*(W-1)` was wrong.
- **The `SIFTDescriptor` shape fix is correct.** `forward` guards `KORNIA_CHECK_SHAPE(input, ["B", "1", f"{self.patch_size}", f"{self.patch_size}"])` (`siftdesc.py:186`), so `patch_size` is right and `num_spatial_bins` was not.
- **Nothing pins the old error string.** Grepped `tests/` and `kornia/` for the message — no assertion depends on the broken whitespace.
- **The TODO cleanup is complete.** `main` had exactly three `# TODO: Recompute doctest` markers, all in `responses.py`, all removed. The surviving `# TODO: Refactor doctest` in `get_laf_pts_to_draw` sits on a matplotlib example that cannot be a doctest at all — correctly left alone.
- `round(decimals=…) + 0.` matches the existing house idiom (`kornia/augmentation/_2d/geometric/rotation.py:173`), and `Tensor.round(decimals=)` predates the declared `torch>=2.0` floor.

### Nits, none blocking

1. **One cell has a thin rounding margin.** `harris_response[0,0,2,2]` is `0.00335024` — only **2.4e-7** above the `0.00335` boundary that decides `0.0034` vs `0.0033`. That is roughly 500 float32 ULP of headroom, it agrees between `float32` and `float64`, and the doctest passes both on the ubuntu doctest job and locally on arm64, so it is not flaky today. Just noting it as the first cell to look at if this example ever starts moving. (`decimals=3` would widen every margin, at the cost of resolution — probably not worth it.)
2. **While you are in this file:** `DenseSIFTDescriptor`'s `Args:` block (`siftdesc.py:246`) has a dangling line — *"You might want to set odd number and relevant padding to keep feature map size"* — indented so napoleon folds it into `num_spatial_bins`'s description at the wrong level. Same class of defect as the rest of the PR. Fold it in or leave it, either is fine.
3. **`#4067` will still be open after this, and should be.** This PR closes #4066 only. #4067 has two halves: the `laf.py`/`siftdesc.py` items you fix here, and the unused `match_fginn` argument. The second half wants a signature change rather than a docstring note, so it should not be closed by a docs PR — I have raised that on #4075 and #4077. Leaving #4067 open with only the `lafs1` item outstanding is the right end state for this PR.

### Combined state

I merged #4072–#4077 together on top of `main` (all six merge cleanly, pairwise and combined) and ran the suites: `tests/feature` cpu float32+float64 → **658 passed, 126 skipped**; `tests/geometry/transform` cpu → **515 passed**; the four feature doctests → **12 passed**; `ruff check` clean.

Note this branch is 3 commits behind `main` (`1442a352`, `e5e07667`, `6fa75d64`); none of them touch `kornia/feature`.

ducha-aiki (commented):
Reviewed current head `24ff0ae8` and found no actionable issues. The response doctests now execute with corrected expectations, and the LAF/SIFT documentation fixes match the implementations. Focused doctest validation passed (6 attempted, 0 failed) on the project environment; the full CI matrix is green.

ducha-aiki:
Applied both actionable items; pushed as `docs(feature): address review — CH not N in the LAF examples, fold the DenseSIFT line`.

- **Copilot, `laf.py:321` and `:352`** — the `denormalize_laf` / `normalize_laf` examples unpacked `images.size()` as `B,N,H,W` while `N` is the LAF count everywhere else in the file and the `Args:` block documents `images` as `(B, CH, H, W)`. Both now say `B,CH,H,W`.
- **Review nit 2** — `DenseSIFTDescriptor`'s `num_spatial_bins` entry had its second sentence at the wrong indentation, so napoleon folded it into the block at the wrong level. Folded into the argument description.

Not changed: nit 1 (the `0.00335024` rounding margin — noted, not flaky today) and nit 3 (`#4067` stays open, resolved by #4077's half).

`pytest --doctest-modules kornia/feature/responses.py kornia/feature/laf.py kornia/feature/siftdesc.py` → 12 passed. pre-commit clean on the three files.

ducha-aiki:
## Post-merge independent review — verdict: **correct and complete.** No findings.

Re-reviewed from scratch at a higher reasoning budget (this one was reviewed at a lower budget before it merged, so I re-derived rather than re-read).

### The doctests now execute, and the defect class is fully eradicated

```console
$ pytest --doctest-modules kornia/feature/responses.py -q
3 passed
```

The root cause as stated is exactly right — `doctest.DocTestParser._IS_BLANK_OR_COMMENT` discards an example whose source is blank or comment-only **together with its `want`**, so `>>> # compute the response map` swallowed the entire expected tensor. I swept all of `kornia/` for the same shape (a `>>>` line that is blank-or-comment-only, followed by a non-prompt line):

```
--- '>>> #'-only source lines whose following line is not a prompt ---
  (none)
```

So there are no remaining instances anywhere in the package, and no `Recompute doctest` markers are left (`laf.py:309`'s `# TODO: Refactor doctest` and `pinhole.py:673` are different markers, not in scope here).

### The rounding is stable enough to publish

`.round(decimals=4) + 0.0` is a good call for the exact-zero columns, but it does introduce a new way to be flaky: a value sitting on a `.00005` tie could round differently across platforms. I measured how close any published value gets to a tie:

```
harris   closest approach to a rounding boundary: 0.0024 of 0.5  (abs 2.4e-7; min |value| 5.8e-04)
gftt     0.1173 of 0.5
hessian  0.0499 of 0.5  (abs 5.0e-6)
```

`harris` is the tight one, but 2.4e-7 absolute on values of order 1e-3 is ~4e-4 relative — roughly 3000 float32 eps of headroom, far more than convolution reassociation across BLAS backends will move these. Stable.

As a sanity check that the new values are the *right* current output and not a mis-paste: all three tensors preserve the input's 4-fold symmetry (symmetric under transpose and under both axis reversals), which they would not if a row had been dropped or shifted.

### The docstring corrections match the code exactly

Checked `denormalize_laf` line by line against the implementation:

```python
wf = float(w - 1);  hf = float(h - 1);  min_size = min(hf, wf)
coef = ones(1,1,2,3) * min_size
coef[0,0,0,2] = wf      # x scales by W-1
coef[0,0,1,2] = hf      # y scales by H-1   <- the docstring said W-1
```

So `[a11*MIN_SIZE a12*MIN_SIZE x*(W-1)] / [a21*MIN_SIZE a22*MIN_SIZE y*(H-1)]` is right, `a21`→`a12` in the source top row is right, `normalize_laf`'s `Returns:` fix is right, and `images: (B, CH, H, W)` matches `_, _, h, w = images.size()`.

`SIFTDescriptor`'s input shape: `KORNIA_CHECK_SHAPE(input, ["B", "1", f"{self.patch_size}", f"{self.patch_size}"])` — the new docstring matches, the old one didn't.

The error message renders as claimed:

```
'Patch size 6 is incompatible with the requested number of spatial bins 4 for SIFT descriptor. Usually it happens when patch size is too small for num_spatial_bins specified'
```

`pytest --doctest-modules` over `responses.py`, `laf.py`, `siftdesc.py`, `matching.py`: 12 passed, 1 skipped (the skip is the weights-gated `DescriptorMatcherWithSteerer` example, which the download-skip mechanism handles — consistent with the "local, not a repo defect" call in the PR body).

Nothing to change.

---

PR #4073, opened 2026-08-28
fix(feature): clamp the MPS patch-extraction grid to pixel centers, not edges

### LINKED ISSUE #4063: extract_patches_*: MPS grid clamp is half a pixel off, patches disagree with CPU/CUDA at image borders
### Describe the bug

`extract_patches_simple` and `extract_patches_from_pyramid` work around the missing `padding_mode="border"` on MPS by clamping the sampling grid to `[-1, 1]` and using `padding_mode="zeros"`. The clamp bound is half a pixel off, so MPS patches disagree with CPU/CUDA by up to 0.5 (on a `[0, 1]` image) wherever a patch touches the image border.

https://github.com/kornia/kornia/blob/main/kornia/feature/laf.py#L433-L441
https://github.com/kornia/kornia/blob/main/kornia/feature/laf.py#L489-L497

```python
if img.device.type == "mps":
    out.append(
        F.grid_sample(
            img[i : i + 1].expand(grid.size(0), ch, h, w),
            grid.clamp(-1, 1),          # <-- outer EDGE of the border pixel
            padding_mode="zeros",
            align_corners=False,
        )
    )
```

With `align_corners=False`, normalized coordinate ±1 is the **outer edge** of the border pixel, which in pixel space is index −0.5 (resp. `size − 0.5`). Bilinear sampling there blends the border pixel with the zero padding and returns roughly half its value. `padding_mode="border"` instead clamps the *sampling coordinate*, returning the border pixel itself.

The correct clamp target is the outermost pixel **centre**, which for `align_corners=False` is `±(1 − 1/size)`.

### Reproduction

```python
import torch
import torch.nn.functional as F

torch.manual_seed(0)
B, C, H, W = 1, 1, 64, 64
img = torch.rand(B, C, H, W)
grid = torch.rand(B, 32, 32, 2) * 3.0 - 1.5          # deliberately runs off the edges

ref = F.grid_sample(img, grid, padding_mode="border", align_corners=False)   # CPU/CUDA behaviour

cur = F.grid_sample(img.to("mps"), grid.clamp(-1, 1).to("mps"),
                    padding_mode="zeros", align_corners=False).cpu()

gx = grid.clone()
gx[..., 0] = gx[..., 0].clamp(-1 + 1.0 / W, 1 - 1.0 / W)
gx[..., 1] = gx[..., 1].clamp(-1 + 1.0 / H, 1 - 1.0 / H)
new = F.grid_sample(img.to("mps"), gx.to("mps"),
                    padding_mode="zeros", align_corners=False).cpu()

print("current clamp(-1,1)   max|mps-cpu| =", (ref - cur).abs().max().item())
print("clamp to pixel centre max|mps-cpu| =", (ref - new).abs().max().item())
```

```
current clamp(-1,1)   max|mps-cpu| = 0.497671
clamp to pixel centre max|mps-cpu| = 0.0
```

End to end, through the public functions:

```python
import torch
from kornia.feature.laf import (extract_patches_simple,
                                extract_patches_from_pyramid,
                                laf_from_center_scale_ori)

torch.manual_seed(0)
img = torch.rand(1, 1, 64, 64)
laf = laf_from_center_scale_ori(torch.tensor([[[3.0, 3.0]]]),
                                torch.full((1, 1, 1, 1), 12.0))   # patch overlaps the corner

for fn in (extract_patches_simple, extract_patches_from_pyramid):
    p_cpu = fn(img, laf, 32)
    p_mps = fn(img.to("mps"), laf.to("mps"), 32).cpu()
    d = (p_cpu - p_mps).abs()
    print(f"{fn.__name__:30s} max={d.max():.4f}  >1e-3 in {(d > 1e-3).float().mean():.1%} of pixels")
```

```
extract_patches_simple         max=0.3950  >1e-3 in 64.7% of pixels
extract_patches_from_pyramid   max=0.3950  >1e-3 in 64.7% of pixels
```

Every descriptor built on these patches (`get_laf_descriptors`, `LAFDescriptor`, `SIFTFeature`, `KeyNetAffNetHardNet`, …) inherits the discrepancy for any keypoint near the image border.

### Expected behavior

MPS patch extraction should match CPU/CUDA. The `border` workaround itself is still required — on torch 2.9.1, `F.grid_sample(..., padding_mode="border")` on MPS raises `RuntimeError: MPS: Unsupported Border padding mode` — but clamping to the outermost pixel centre makes it exact.

### Environment

```
kornia main | torch 2.9.1 | python 3.11.14 | macOS (darwin 25.5.0) | MPS available
```

### Additional context

Introduced in #3649. The fix does not touch the CPU/CUDA code path at all.

ducha-aiki:
## Problem

MPS has no `padding_mode="border"` for `F.grid_sample` — torch 2.9.1 still raises `RuntimeError: MPS: Unsupported Border padding mode` — so `extract_patches_simple` and `extract_patches_from_pyramid` emulate it with zero padding and a clamped grid (added in #3649). The clamp bound is off by half a pixel.

With `align_corners=False`, normalized coordinate ±1 is the outer **edge** of the border pixel, i.e. pixel index −0.5. Bilinear sampling there blends the border pixel with the zero padding and returns roughly half its value. `padding_mode="border"` instead clamps the *sampling coordinate* and returns the pixel itself. The outermost pixel **center** sits at ±(1 − 1/size).

```python
import torch, torch.nn.functional as F
torch.manual_seed(0)
H = W = 64
img = torch.rand(1, 1, H, W)
grid = torch.rand(1, 32, 32, 2) * 3.0 - 1.5           # deliberately off every edge

ref = F.grid_sample(img, grid, padding_mode="border", align_corners=False)   # CPU/CUDA behaviour
cur = F.grid_sample(img.to("mps"), grid.clamp(-1, 1).to("mps"),
                    padding_mode="zeros", align_corners=False).cpu()
gx = grid.clone()
gx[..., 0] = gx[..., 0].clamp(-1 + 1.0 / W, 1 - 1.0 / W)
gx[..., 1] = gx[..., 1].clamp(-1 + 1.0 / H, 1 - 1.0 / H)
new = F.grid_sample(img.to("mps"), gx.to("mps"),
                    padding_mode="zeros", align_corners=False).cpu()
```
```
current clamp(-1,1)   max|mps-cpu| = 0.497671
clamp to pixel centre max|mps-cpu| = 0.0
```

End to end, on a patch overlapping the image corner:

| | before | after |
|---|---|---|
| `extract_patches_simple` max abs diff vs CPU | **0.3950** | 2.5e-06 |
| `extract_patches_from_pyramid` max abs diff vs CPU | **0.3950** | 2.5e-06 |
| fraction of pixels differing by >1e-3 | **64.7%** | 0% |

Every descriptor built on these patches — `get_laf_descriptors`, `LAFDescriptor`, `SIFTFeature`, `KeyNetAffNetHardNet`, … — inherited the discrepancy for keypoints near the image border.

## Change

A small `_clamp_grid_to_pixel_centers(grid, h, w)` helper, called from the two MPS branches in place of `grid.clamp(-1, 1)`. It documents *why* the naive bound is wrong, since that is the part that is easy to regress.

## The CPU/CUDA path is untouched

The helper is called only inside `if device.type == "mps"`. CPU output is **byte-identical** to `main` (`torch.equal`, not `assert_close`) for both extractors over a 2×60-LAF batch on a 2×3×96×96 image:

```
CPU 'simple': byte-identical to main = True
CPU 'pyr':    byte-identical to main = True
```

## Tests

`TestClampGridToPixelCenters` in `tests/feature/test_laf.py`:

- `test_matches_border_padding` — the clamped grid + zero padding reproduces `padding_mode="border"`.
- `test_naive_clamp_is_not_equivalent` — guards against regressing to `grid.clamp(-1, 1)`, asserting the naive bound differs by >1e-3.
- `test_extractors_match_cpu_at_the_border` — end-to-end pin at the call sites, both extractors, for a LAF that runs off two edges.
- plus shape and gradcheck.

The reference is always computed **on CPU**, because `padding_mode="border"` is exactly what the device under test may not support — an earlier revision of these tests referenced the target device and failed on MPS for that reason.

The end-to-end test discriminates: reverting `laf.py` alone and rerunning on MPS gives

```
FAILED ...test_extractors_match_cpu_at_the_border[mps-float32-extract_patches_simple]
FAILED ...test_extractors_match_cpu_at_the_border[mps-float32-extract_patches_from_pyramid]
```

## Verification

```console
652 passed, 126 skipped

$ pytest tests/feature --device=mps --dtype=float32 -q     # main: 7 failures, branch: 7 failures
# no new failures; the 7 are pre-existing MPS gaps (bicubic interpolation, float64)

$ pixi run pre-commit-all
all hooks Passed
```

Closes #4063


ducha-aiki:
## Review — APPROVE

The diagnosis and the fix are both correct, and I reproduced every number rather than trusting the table.

### Verified

- **The math.** With `align_corners=False` pixel centre *i* maps to `(2i+1)/size - 1`, so the outermost centres are `±(1 - 1/size)` and `±1` is the outer *edge*. The clamp bound is right, including the degenerate `h = w = 1` case, where both bounds collapse to `0.0` — the single pixel's centre — rather than to something out of range.
- **The end-to-end effect on MPS**, on the corner LAF from the issue:

  | | `main` | this PR |
  |---|---|---|
  | `extract_patches_simple`, max&#124;mps−cpu&#124; | 0.395010 | 2.5e-06 |
  | `extract_patches_from_pyramid`, max&#124;mps−cpu&#124; | 0.395010 | 2.5e-06 |
  | fraction of pixels differing by >1e-3 | 64.7% | 0% |

- **CPU is untouched.** `torch.equal` is `True` against `main` for both extractors on a 2×60-LAF batch over a 2×3×96×96 image. Byte-identical, not merely close.
- **Both MPS sites are covered, and there are only two.** `grep -rn 'device.type == "mps"' kornia/` finds the border-emulation workaround at `laf.py:433` and `laf.py:489` and nowhere else; the pyramid site correctly passes the *current level's* `h_l, w_l`, not the full-image dims.
- **The test discriminates.** Reverting `kornia/feature/laf.py` alone and rerunning on MPS fails `test_extractors_match_cpu_at_the_border` for both extractors.
- **The helper survives TorchScript.** `torch.jit.script(_clamp_grid_to_pixel_centers)` compiles and matches eager exactly. Worth having checked, since `laf.py` is full of scripted functions and there are `test_jit` cases throughout `tests/feature/test_laf.py`.
- **No new MPS failures.** Full `tests/feature --device=mps --dtype=float32`: `main` 429 passed / 7 failed, this branch 438 passed / 7 failed — the same seven pre-existing `MPS: Unsupported Bicubic interpolation` failures in `test_xfeat.py`. Also `tests/feature` cpu float32+float64 → 658 passed with all six open PRs merged together.

Computing the reference on CPU rather than on the device under test is the right call and the reason the earlier revision failed — good that the class docstring records it.

### Requests

1. **Add a CHANGELOG entry** under `Unreleased → Bug fixes`. This is a user-visible numerical change: up to 0.5 on a `[0, 1]` image for any patch touching a border, and it propagates into `get_laf_descriptors`, `LAFDescriptor`, `SIFTFeature`, `KeyNetAffNetHardNet` and everything downstream. #4044 (YUV) and #4076 both got prose for smaller deltas; this deserves the same.

### Nits

2. `torch.stack([x, y], dim=-1)` silently assumes `grid.shape[-1] == 2`. Correct at both call sites and likely forever, but a one-line note (or a shape check) would make the helper safe to reuse if a 3-D `grid_sample` path ever wants it.
3. `test_extractors_match_cpu_at_the_border` is vacuous on CPU — it compares CPU against CPU. The class docstring explains why that is unavoidable, and it means CI (which has no MPS runner) never exercises the discriminating half. Nothing to do here, just worth being explicit that this pin only fires on a developer's Mac.

### Adjacent, separate issue

`elastic_transform2d` does `(grid + disp).clamp(-1, 1)` at `kornia/geometry/transform/elastic_transform.py:129` — the same half-pixel-off bound. It is a different situation (it clamps a user-supplied displacement, with a caller-chosen `padding_mode`, rather than emulating `border`), so it does not belong in this PR, but it is the other place in the tree where `clamp(-1, 1)` is being treated as "the image extent".

ducha-aiki (commented):
Reviewed current head `0b7486b1` and found no actionable issues. The per-axis clamp correctly maps MPS border padding to the outer pixel centers, including rectangular and singleton dimensions, and both extraction call sites are covered. Focused MPS comparisons stayed within `2.53e-06`; the full CI matrix is green.

ducha-aiki:
Applied the blocking request and the nit; pushed as `docs(changelog): record the MPS border-clamp fix; note the helper's 2-D assumption`.

- **Request 1 — CHANGELOG.** Added under `Unreleased → Bug fixes`, with the measured numbers (`0.395` → `2.5e-6` max deviation from CPU on a corner patch, 64.7% of pixels previously off by >1e-3) and the note that the CPU/CUDA paths never took this branch.
- **Nit 2 — the 2-D assumption.** `_clamp_grid_to_pixel_centers`'s `Args:` block now says only the 2-D grid form is handled and that a 3-D grid would need a third bound.

Not changed: nit 3 (the CPU-vs-CPU vacuity of `test_extractors_match_cpu_at_the_border` is unavoidable and the class docstring already records it), and the `elastic_transform2d` `clamp(-1, 1)` — agreed it is a different situation and does not belong here.

`tests/feature/test_laf.py --device=cpu` → 93 passed, 6 skipped. pre-commit clean.

ducha-aiki:
## Independent review — verdict: **correct and complete; approve**

Re-reviewed from scratch at a higher reasoning budget, re-deriving the maths and re-running everything on this machine (macOS, MPS available, torch 2.9.1) rather than trusting the PR body.

### The identity holds exactly, not approximately

With `align_corners=False` the normalized→pixel map is `p = ((g+1)·size − 1)/2`, so `g = ±1` is `p = −0.5` / `size−0.5` (pixel *edges*) and the outermost pixel *centers* are at `g = ±(1 − 1/size)`. The interesting half of the argument is the one the PR body doesn't spell out: for `p ∈ (−0.5, 0)` `border` blends pixel 0 with `clamp(−1) = pixel 0` and returns pixel 0 exactly, while the clamp pins the coordinate to `p = 0` and also returns pixel 0. The two agree on *every* interval, so this is an exact reproduction rather than a close one. Measured:

```
h= 1 w= 1: max|clamped+zeros − border| = 0.000e+00     h=17 w=23: 4.441e-16
h= 1 w= 5: 1.110e-16                                   h=64 w=64: 0.000e+00
h= 5 w= 1: 1.110e-16                                   h= 3 w= 3: 2.220e-16
h= 2 w= 3: 2.220e-16                                   (float64, random grids in [−3, 3])
```

Note the degenerate `size == 1` case: the bound collapses to `clamp(0, 0)`, which is still right.

### The end-to-end gap is fully closed, not just at the corner

The PR body pins one hand-picked corner LAF. I ran a 2×40-LAF batch on a 2×1×96×128 image with centers drawn from `[−0.05·W, 1.05·W] × [−0.05·H, 1.05·H]` and scales 2–32, so patches run off every edge and combination of edges:

```
extract_patches_simple        max|cpu−mps| = 1.323e-05   frac > 1e-3 = 0.0000%
extract_patches_from_pyramid  max|cpu−mps| = 1.323e-05   frac > 1e-3 = 0.0000%
```

1.3e-5 is float32 MPS arithmetic noise. There is no residual divergence left in these two extractors.

`h`/`w` are threaded correctly at both call sites (`img.size()` in `extract_patches_simple`, per-level `cur_img.size()` in `extract_patches_from_pyramid`), and x↔w / y↔h are the right way round for `grid_sample`'s `(x, y)` ordering.

### The pin discriminates — but only on MPS

I reverted the two call sites (leaving the helper) and re-ran:

```
mps : FAILED test_extractors_match_cpu_at_the_border[mps-float32-extract_patches_simple]
      FAILED test_extractors_match_cpu_at_the_border[mps-float32-extract_patches_from_pyramid]
cpu : 11 passed
```

So the call-site pin is real, but kornia's CI is CPU-only and `test_extractors_match_cpu_at_the_border` compares CPU against CPU there — vacuous. What CI *does* keep is `test_matches_border_padding` / `test_naive_clamp_is_not_equivalent`, which pin the helper's maths and do discriminate on CPU. That's the right split; just worth stating that a future regression of the *call sites* would land green.

### Other checks

- `tests/feature` on MPS with all the sibling PRs stacked: 451 passed, 144 skipped, 7 failed — all 7 in `test_xfeat.py` with `RuntimeError: MPS: Unsupported Bicubic interpolation`, identical on `main`. (Small correction to the PR body: the 7 are all bicubic, none is float64.)
- `pytest tests/feature -k ClampGrid`: 11 passed on CPU f32+f64, 5 passed / 1 skipped on MPS.
- `pytest tests/feature/test_laf.py -m jit`: 9 passed.

### One nit (non-blocking)

`TestClampGridToPixelCenters.test_gradcheck` builds its grid from `torch.rand` with no seed (`_make` seeds, but `test_gradcheck` doesn't use it), and `clamp` has a kinked gradient at the bound. The failure probability is around 1e-5 per run — low, but a `torch.manual_seed(0)` would make it zero for free, and the two neighbouring tests already seed.

Nothing here blocks. The maths is right, the helper is documented at exactly the spot that is easy to regress, and the CPU/CUDA path is provably untouched.

---

PR #4074, opened 2026-08-28
fix(feature): stop DenseSIFTDescriptor sharing the identity cache as a buffer

### LINKED ISSUE #4068: DenseSIFTDescriptor registers a process-global cached tensor as a buffer; load_state_dict corrupts every instance
### Describe the bug

`DenseSIFTDescriptor` registers a tensor that is shared with a process-global cache as its `_poolingconv_weight` buffer. Any in-place write to that buffer — which is exactly what `load_state_dict` does — silently corrupts the cache, every other live `DenseSIFTDescriptor`, and every instance constructed afterwards.

### Root cause

`_get_reshape_kernel` memoises `torch.eye(numel)` in a function attribute and returns a **view of the cached tensor**, not a copy:

https://github.com/kornia/kornia/blob/main/kornia/feature/siftdesc.py#L30-L51

```python
res = cache.get(numel)
if res is None:
    res = torch.eye(numel)
    cache[numel] = res
return res.view(numel, kd, ky, kx)      # view of the shared tensor
```

`DenseSIFTDescriptor.__init__` then registers that view directly as a buffer:

https://github.com/kornia/kornia/blob/main/kornia/feature/siftdesc.py#L312-L314

```python
Pw = _get_reshape_kernel(num_ang_bins, num_spatial_bins, num_spatial_bins).float()
self.register_buffer("_poolingconv_weight", Pw)
```

`.float()` on an already-float tensor is a no-op that returns the *same* object, so no copy happens anywhere along the path. Every `DenseSIFTDescriptor` with the same `(num_ang_bins, num_spatial_bins)` ends up with a buffer backed by one shared storage.

### Reproduction

```python
import torch
from kornia.feature.siftdesc import DenseSIFTDescriptor, _get_reshape_kernel

a = DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)
b = DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)
cache = _get_reshape_kernel._eye_cache[8 * 4 * 4]

print("buffer shares storage with the global cache:",
      a._poolingconv_weight.untyped_storage().data_ptr() == cache.untyped_storage().data_ptr())

with torch.no_grad():
    a._poolingconv_weight.zero_()

print("b's buffer changed too       :", bool(b._poolingconv_weight.abs().sum() == 0))
print("the global cache is corrupt  :", bool(cache.abs().sum() == 0))
print("a NEW instance is born zeroed:",
      bool(DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)._poolingconv_weight.abs().sum() == 0))
```

```
buffer shares storage with the global cache: True
b's buffer changed too       : True
the global cache is corrupt  : True
a NEW instance is born zeroed: True
```

`load_state_dict` reaches this in normal use, because it copies into buffers **in place**:

```python
import torch
from kornia.feature.siftdesc import DenseSIFTDescriptor, _get_reshape_kernel

victim = DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)
sd = victim.state_dict()
sd["_poolingconv_weight"] = torch.zeros_like(sd["_poolingconv_weight"])
victim.load_state_dict(sd)

print("global cache zeroed by one instance's load_state_dict:",
      bool(_get_reshape_kernel._eye_cache[128].abs().sum() == 0))
print("a freshly constructed DenseSIFTDescriptor is now broken:",
      bool(DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)._poolingconv_weight.abs().sum() == 0))
```

```
global cache zeroed by one instance's load_state_dict: True
a freshly constructed DenseSIFTDescriptor is now broken: True
```

The corruption is silent — no error, just wrong descriptors from then on, including from objects the caller never touched.

### Expected behavior

`_get_reshape_kernel` should return a copy (or the caller should clone before `register_buffer`), so instances never share storage.

### Additional context

Secondary points on the same cache:

- It is keyed only on `numel`, and the cached tensor is created on CPU in the default dtype, so it is not a device/dtype-aware cache.
- `_MAX_CACHED = 4096` bounds `numel`, but `torch.eye(4096)` is 67 MB and is retained for the lifetime of the process. The number of distinct cached `numel` values is unbounded.

ducha-aiki:
## Problem

`_get_reshape_kernel` memoises `torch.eye(numel)` in a function attribute and returned a **view of the cached tensor**:

```python
res = cache.get(numel)
if res is None:
    res = torch.eye(numel)
    cache[numel] = res
return res.view(numel, kd, ky, kx)      # view of the shared tensor
```

`DenseSIFTDescriptor.__init__` registers that view directly as a buffer, and `.float()` on an already-float tensor is a no-op returning the *same object*, so no copy happened anywhere:

```python
Pw = _get_reshape_kernel(num_ang_bins, num_spatial_bins, num_spatial_bins).float()
self.register_buffer("_poolingconv_weight", Pw)
```

Every `DenseSIFTDescriptor` with the same `(num_ang_bins, num_spatial_bins)` therefore shared one storage with the process-global cache.

`load_state_dict` reaches this in ordinary use, because it copies into buffers **in place**:

```python
victim = DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)
sd = victim.state_dict()
sd["_poolingconv_weight"] = torch.zeros_like(sd["_poolingconv_weight"])
victim.load_state_dict(sd)
```
```
global cache zeroed by one instance's load_state_dict: True
a freshly constructed DenseSIFTDescriptor is now broken: True
```

No error is raised — the descriptors are simply wrong from then on, including from objects the caller never touched.

## Change

Remove the cache. `_get_reshape_kernel` is now:

```python
numel: int = kd * ky * kx
return torch.eye(numel).view(numel, kd, ky, kx)
```

An earlier revision of this PR kept the cache and cloned on the way out, on the assumption that "a memcpy is cheaper than `torch.eye`". Measured, that is not true — **a safe cache is slower than no cache at every size**, including the one every default `DenseSIFTDescriptor` hits (4 threads, min of 7 runs of 3000 iterations, cache pre-warmed):

| `kd,ky,kx` | `numel` | cached + `.clone()` | plain `torch.eye().view()` | speedup |
|---|--:|--:|--:|--:|
| (8, 4, 4) — the default | 128 | 2.60 µs | **2.00 µs** | **0.77×** |
| (8, 8, 8) | 512 | 33.43 µs | 33.90 µs | 1.01× |
| (8, 4, 16) | 512 | 34.11 µs | 33.70 µs | 0.99× |
| (16, 16, 16) | 4096 | 3901.87 µs | **2812.56 µs** | **0.72×** |

(0.77× at the default reproduced across three independent trials: 0.76×, 0.78×, 0.78×.) Whole-module: `DenseSIFTDescriptor()` construction goes 143.0 µs → 142.0 µs.

So there was nothing left to preserve, and no value of `_MAX_CACHED` worth keeping. Removing the cache also deletes `_MAX_CACHED`, the `hasattr`/function-attribute block and two `# type: ignore[attr-defined]`s, and removes the aliasing bug class rather than containing it. The cache was introduced by #3206, an automated micro-optimisation; the docstring now records the cost model so it is not silently re-added.

## Behaviour

Unchanged. Descriptor output and the resulting buffers are **byte-identical** to `main` (`torch.equal`) across three configurations, and re-verified byte-identical against the cached-and-cloned revision after the cache was removed:

```
8_4      byte-identical to main: True     8_4_buf  True
6_3      byte-identical to main: True     6_3_buf  True
8_2      byte-identical to main: True     8_2_buf  True
```

## Tests

Two regression tests on `TestDenseSIFTDescriptor`, both of which fail on unmodified `siftdesc.py`:

```
FAILED ...::test_instances_do_not_share_buffer_storage - assert 5100470272 != 5100470272
FAILED ...::test_load_state_dict_does_not_leak_into_other_instances - Tensor-likes are not close!
```

## Verification

```console
643 passed, 126 skipped

$ pixi run pre-commit-all
all hooks Passed
```

## Also resolved

Both secondary points from #4068 go away with the cache: it is no longer keyed only on `numel` (there is no key), and it no longer retains 67 MB for the process lifetime at the `_MAX_CACHED = 4096` bound.

## Not addressed here

`DenseSIFTDescriptor.get_pooling_kernel()` and `SIFTDescriptor.get_gaussian_kernel()` hand out a `.detach()` of a live module buffer, so a caller's in-place write lands on the module's own state. Same "handed a live reference to internal state" class, but instance-local rather than process-global — filed separately as #4083.

Closes #4068

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,7 +136,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   tensor; `.float()` on an already-float tensor returns the same object, so no copy happened between the cache and
   the buffer, and `load_state_dict` copies into buffers in place. Loading a checkpoint into one `DenseSIFTDescriptor`
   therefore overwrote the shared identity matrix, with no error raised and every descriptor built from it wrong from
-  then on. The cached tensor is now cloned before it is handed out. Descriptor output is byte-identical to before.
+  then on. The cache is removed rather than made safe: cloning on the way out is what a safe cache requires, and a
+  clone costs more than rebuilding the identity at every size measured (2.6 us against 2.0 us at the default
+  `numel = 8 * 4 * 4`; 3.9 ms against 2.8 ms at the former 4096 bound, which also retained 67 MB for the lifetime of
+  the process). Descriptor output is byte-identical to before.
 
 
 ## :rocket: [0.6.11] - 2022-03-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   existing FGINN tests pass unchanged on both sides, which is how this survived; two tests that discriminate it were
   added. `match_fginn`'s docstring now also records the saturation corner, where every candidate falls within
   `spatial_th` of the 1st nearest neighbor, the ratio collapses towards zero and the match is accepted.
+* Fix `DenseSIFTDescriptor` registering a process-global cached tensor as its `_poolingconv_weight` buffer, so that
+  one instance's `load_state_dict` silently corrupted every other instance and every instance constructed later
+  (#4068). `_get_reshape_kernel` memoises `torch.eye(numel)` per `numel` and used to return a *view* of the cached
+  tensor; `.float()` on an already-float tensor returns the same object, so no copy happened between the cache and
+  the buffer, and `load_state_dict` copies into buffers in place. Loading a checkpoint into one `DenseSIFTDescriptor`
+  therefore overwrote the shared identity matrix, with no error raised and every descriptor built from it wrong from
+  then on. The cached tensor is now cloned before it is handed out. Descriptor output is byte-identical to before.
 
 
 ## :rocket: [0.6.11] - 2022-03-28
--- a/kornia/feature/siftdesc.py
+++ b/kornia/feature/siftdesc.py
@@ -30,14 +30,9 @@ from kornia.geometry.conversions import pi
 def _get_reshape_kernel(kd: int, ky: int, kx: int) -> torch.Tensor:
     """Return neigh2channels conv kernel.
 
-    The identity matrix used to be memoised per ``numel`` and handed to the caller as a view of
-    the cached tensor, which let one consumer's in-place write -- ``load_state_dict`` copies into
-    buffers in place -- corrupt the cache and therefore every other consumer, including modules
-    constructed later (#4068). Making the cache safe means cloning on the way out, which costs
-    more than rebuilding the identity, so the cache is gone: measured on a 4-thread CPU, the
-    cached-and-cloned path is 2.6 us against 2.0 us for a plain ``torch.eye`` at the default
-    ``numel = 8 * 4 * 4``, and 3.9 ms against 2.8 ms at the old 4096 bound, where it also retained
-    67 MB for the lifetime of the process.
+    Deliberately not memoised: the result is registered as a buffer, so a shared cache has to clone
+    on the way out to keep callers from writing into it, and the clone costs more than rebuilding
+    the identity.
     """
     numel: int = kd * ky * kx
     return torch.eye(numel).view(numel, kd, ky, kx)
--- a/kornia/feature/siftdesc.py
+++ b/kornia/feature/siftdesc.py
@@ -28,11 +28,15 @@ from kornia.geometry.conversions import pi
 
 
 def _get_reshape_kernel(kd: int, ky: int, kx: int) -> torch.Tensor:
-    """Return neigh2channels conv kernel."""
+    """Return neigh2channels conv kernel.
+
+    The identity matrix is memoised per ``numel`` to avoid rebuilding it for the common kernel
+    sizes, but the caller always receives a **fresh copy**. Handing out the cached tensor itself
+    would let one consumer's in-place write -- ``load_state_dict`` copies into buffers in place --
+    corrupt the cache and therefore every other consumer, including modules constructed later.
+    """
     numel: int = kd * ky * kx
 
-    # Fast-path: use static _eye_cache if available for small numel
-    # (to avoid repeated allocations for common kernel sizes)
     # The cache size is limited for memory efficiency.
     # NOTE: If memory is a concern and large kd/ky/kx are rare, adjust _MAX_CACHED.
     _MAX_CACHED = 4096
@@ -44,7 +48,8 @@ def _get_reshape_kernel(kd: int, ky: int, kx: int) -> torch.Tensor:
         if res is None:
             res = torch.eye(numel)
             cache[numel] = res
-        return res.view(numel, kd, ky, kx)
+        # clone: never expose the cached tensor to the caller
+        return res.view(numel, kd, ky, kx).clone()
     else:
         # fallback to normal allocation for big kernels
         return torch.eye(numel).view(numel, kd, ky, kx)
--- a/kornia/feature/siftdesc.py
+++ b/kornia/feature/siftdesc.py
@@ -30,29 +30,17 @@ from kornia.geometry.conversions import pi
 def _get_reshape_kernel(kd: int, ky: int, kx: int) -> torch.Tensor:
     """Return neigh2channels conv kernel.
 
-    The identity matrix is memoised per ``numel`` to avoid rebuilding it for the common kernel
-    sizes, but the caller always receives a **fresh copy**. Handing out the cached tensor itself
-    would let one consumer's in-place write -- ``load_state_dict`` copies into buffers in place --
-    corrupt the cache and therefore every other consumer, including modules constructed later.
+    The identity matrix used to be memoised per ``numel`` and handed to the caller as a view of
+    the cached tensor, which let one consumer's in-place write -- ``load_state_dict`` copies into
+    buffers in place -- corrupt the cache and therefore every other consumer, including modules
+    constructed later (#4068). Making the cache safe means cloning on the way out, which costs
+    more than rebuilding the identity, so the cache is gone: measured on a 4-thread CPU, the
+    cached-and-cloned path is 2.6 us against 2.0 us for a plain ``torch.eye`` at the default
+    ``numel = 8 * 4 * 4``, and 3.9 ms against 2.8 ms at the old 4096 bound, where it also retained
+    67 MB for the lifetime of the process.
     """
     numel: int = kd * ky * kx
-
-    # The cache size is limited for memory efficiency.
-    # NOTE: If memory is a concern and large kd/ky/kx are rare, adjust _MAX_CACHED.
-    _MAX_CACHED = 4096
-    if numel <= _MAX_CACHED:
-        if not hasattr(_get_reshape_kernel, "_eye_cache"):
-            _get_reshape_kernel._eye_cache = {}  # type: ignore[attr-defined]
-        cache = _get_reshape_kernel._eye_cache  # type: ignore[attr-defined]
-        res = cache.get(numel)
-        if res is None:
-            res = torch.eye(numel)
-            cache[numel] = res
-        # clone: never expose the cached tensor to the caller
-        return res.view(numel, kd, ky, kx).clone()
-    else:
-        # fallback to normal allocation for big kernels
-        return torch.eye(numel).view(numel, kd, ky, kx)
+    return torch.eye(numel).view(numel, kd, ky, kx)
 
 
 def get_sift_pooling_kernel(ksize: int = 25) -> torch.Tensor:
--- a/tests/feature/test_siftdesc.py
+++ b/tests/feature/test_siftdesc.py
@@ -135,6 +135,34 @@ class TestDenseSIFTDescriptor(BaseTester):
         sift = DenseSIFTDescriptor()
         sift.__repr__()
 
+    def test_instances_do_not_share_buffer_storage(self, device):
+        """Instances must not alias the module-level identity cache (see #4068).
+
+        `_get_reshape_kernel` memoises `torch.eye(numel)`. Returning the cached tensor itself made
+        every `DenseSIFTDescriptor` with the same configuration share one storage, so a single
+        in-place write leaked into all of them.
+        """
+        a = DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)
+        b = DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)
+        assert a._poolingconv_weight.untyped_storage().data_ptr() != (
+            b._poolingconv_weight.untyped_storage().data_ptr()
+        )
+
+    def test_load_state_dict_does_not_leak_into_other_instances(self, device):
+        """`load_state_dict` copies into buffers in place; that must stay local to one module."""
+        a = DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)
+        b = DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)
+        expected = a._poolingconv_weight.clone()
+
+        state = b.state_dict()
+        state["_poolingconv_weight"] = torch.zeros_like(state["_poolingconv_weight"])
+        b.load_state_dict(state)
+
+        self.assert_close(a._poolingconv_weight, expected)
+        # and a module built afterwards is still healthy
+        c = DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)
+        self.assert_close(c._poolingconv_weight, expected)
+
     def test_gradcheck(self, device):
         batch_size, channels, height, width = 1, 1, 16, 16
         patches = torch.rand(batch_size, channels, height, width, device=device, dtype=torch.float64)
--- a/tests/feature/test_siftdesc.py
+++ b/tests/feature/test_siftdesc.py
@@ -136,10 +136,10 @@ class TestDenseSIFTDescriptor(BaseTester):
         sift.__repr__()
 
     def test_instances_do_not_share_buffer_storage(self, device):
-        """Instances must not alias the module-level identity cache (see #4068).
+        """Instances must not share storage for `_poolingconv_weight` (see #4068).
 
-        `_get_reshape_kernel` memoises `torch.eye(numel)`. Returning the cached tensor itself made
-        every `DenseSIFTDescriptor` with the same configuration share one storage, so a single
+        `_get_reshape_kernel` used to memoise `torch.eye(numel)` and hand out a view of it, so
+        every `DenseSIFTDescriptor` with the same configuration shared one storage and a single
         in-place write leaked into all of them.
         """
         a = DenseSIFTDescriptor(num_ang_bins=8, num_spatial_bins=4)


ducha-aiki:
## Review — APPROVE

Real bug, minimal fix, and the behaviour claim holds up under a strict check.

### Verified

- **The defect reproduces on `main`,** and both new tests fail there: `test_instances_do_not_share_buffer_storage` → `assert 4966907904 != 4966907904`, and `test_load_state_dict_does_not_leak_into_other_instances` → `Tensor-likes are not close!`. They pass on this branch.
- **Byte-identical behaviour.** Descriptor output *and* `_poolingconv_weight` compare `torch.equal` → `True` against `main` for `(8, 4)`, `(6, 3)` and `(8, 2)`. Six of six.
- **The blast radius is exactly `DenseSIFTDescriptor`.** `_get_reshape_kernel` has one caller in the whole tree (`siftdesc.py:312`); `SIFTDescriptor` does not use it.
- **The docstring's promise is now true on both branches.** The `numel > _MAX_CACHED` fallback already returned a fresh `torch.eye(...)`, so "the caller always receives a fresh copy" holds uniformly after the clone.
- `tests/feature/test_siftdesc.py` cpu float32+float64 → 24 passed, 2 skipped.

### Findings

1. **The same defect class survives one method away, in this class.** `get_pooling_kernel()` returns `self._pooling_kernel`, which `__init__` captures as `self._bin_pooling_kernel_weight.detach()` — an alias, not a copy. Two consequences, both reproduced on this branch:

   - *A caller's in-place write corrupts the module.* `m.get_pooling_kernel().zero_()` zeroes `m._bin_pooling_kernel_weight`. Instance-local rather than process-global, so milder than #4068 — but the same shape of bug, and the docstring you just wrote argues against it.
   - *The alias goes stale on `.to()`.* After `DenseSIFTDescriptor().to(torch.float64)` the buffer is `float64` while `get_pooling_kernel()` still hands back the original **`float32`** tensor. The same would happen for `.to("cuda")`: the accessor would return the CPU tensor for the lifetime of the module.

   Out of scope for a PR that closes #4068, but worth a follow-up issue — it is the more likely one to bite, since `.to(device)` is routine and no exception is raised.

2. **Nit — the cache barely earns its keep now.** With the clone in place it saves only the `torch.eye` construction and pays a memcpy per call, it is still keyed on `numel` alone (dtype- and device-blind), and it can retain 67 MB at the `_MAX_CACHED` bound. Given `_get_reshape_kernel` is called exactly once per module construction, deleting the cache outright would be simpler and cost nothing measurable. The current shape is correct either way — your call, and the "not addressed here" section already names the two design questions honestly.

3. **Nit — a CHANGELOG entry would be appropriate.** The numbers do not move, but `load_state_dict` semantics do, and that is exactly the surface a user notices.

### Combined state

Merges cleanly with #4072, #4073 and #4077 (test-merged pairwise). All six open PRs merged together: `tests/feature` cpu float32+float64 → 658 passed, 126 skipped.

ducha-aiki (commented):
Reviewed current head `839c7303` and found no actionable issues. Cloning the cached reshape kernel prevents `DenseSIFTDescriptor` buffers from sharing storage with the process-global cache. The new tests cover direct aliasing and cache corruption through `load_state_dict`; the full CI matrix is green.

ducha-aiki:
Applied nit 3; finding 1 is now #4083.

- **Nit 3 — CHANGELOG.** Added under `Unreleased → Bug fixes`: the numbers do not move, but `load_state_dict` semantics do.
- **Finding 1 — `get_pooling_kernel()`.** Filed as #4083 rather than folded in here, as suggested. Both symptoms reproduce on this branch: `m.get_pooling_kernel().zero_()` zeroes `m._bin_pooling_kernel_weight`, and after `.to(torch.float64)` the buffer is `float64` while the accessor still returns the original `float32` tensor (the alias is captured in `__init__`, and `nn.Module._apply` replaces the buffer rather than mutating it). The issue also notes that `SIFTDescriptor`'s two accessors track `.to()` correctly but still hand out an alias.
- **Nit 2 — deleting the cache.** Left as is. The clone keeps it correct, and removing it would widen a bug-fix diff for no measurable gain either way.

`tests/feature/test_siftdesc.py --dtype=float32,float64` → 24 passed, 2 skipped. pre-commit clean.

ducha-aiki:
## Independent review — verdict: **correct; approve.** One follow-up worth resolving in this PR.

Re-reviewed from scratch at a higher reasoning budget. I reproduced the defect and re-measured the tradeoff rather than trusting the PR body.

### The bug is real and silent

On unmodified `main`:

```
cache zeroed by one instance's load_state_dict : True
a freshly constructed DenseSIFTDescriptor broken: True   max|Δdescriptor| = 0.1989937
the victim instance itself is still fine       : True
```

That last line is the nastiest part and the PR body doesn't call it out: the module you loaded the checkpoint into is *fine*, because its `PoolingConv.weight` was copied at construction time. The damage lands entirely on `DenseSIFTDescriptor` instances built **afterwards**, in a different part of the program, with no exception and no NaN — just descriptors that are quietly wrong by ~0.2. That is about the worst failure shape available.

Provenance is worth recording: the cache came in with #3206 (`Feat/codeflash-eye-numel-reconstruction-small-cases`), an automated micro-optimization. The optimization introduced the aliasing.

### The fix is right

`_get_reshape_kernel` has exactly one caller (`siftdesc.py:317`), `torch.eye` is contiguous so `.view(...).clone()` is contiguous, and both new tests fail on `main`:

```
FAILED test_instances_do_not_share_buffer_storage - assert 4966416384 != 4966416384
FAILED test_load_state_dict_does_not_leak_into_other_instances - Tensor-likes are not close!
```

`tests/feature/test_siftdesc.py`: 24 passed, 2 skipped.

### Finding: the clone inverts the cache's cost model, and `_MAX_CACHED = 4096` is now indefensible

The PR body says "a memcpy is cheaper than `torch.eye`, so the optimisation it was added for is preserved". Measured on this machine, that is only true at the small end:

| `numel` | `torch.eye(numel)` | `cached.clone()` | speedup | retained forever |
|--:|--:|--:|--:|--:|
| 128 (the default 8×4×4) | 7.9 µs | 2.3 µs | **3.47×** | 0.1 MB |
| 512 | 32.3 µs | 28.6 µs | 1.13× | 1.0 MB |
| 4096 (`_MAX_CACHED`) | 2777 µs | 3969 µs | **0.70×** | **67 MB** |

At the `_MAX_CACHED` bound the cache is now a *pessimization* — cloning is 43% slower than just rebuilding the identity — while still retaining 67 MB for the process lifetime. And at the default config the 5.6 µs it saves is 3.1% of the 181 µs it takes to construct a `DenseSIFTDescriptor` at all.

The PR defers the retained-memory point to "design questions rather than defects", but the clone is what changed the tradeoff, so this PR is the natural place to settle it. Either is fine by me:

- **drop the cache entirely** — `return torch.eye(numel).view(numel, kd, ky, kx)`, deleting the whole `hasattr`/`_MAX_CACHED`/`# type: ignore` block and the bug class with it, for a 5.6 µs regression per construction; or
- **shrink `_MAX_CACHED` to 256** (or 512), keeping the win exactly where it exists and dropping the worst case from 67 MB to 0.26 MB.

I'd take the first — it removes two `# type: ignore[attr-defined]`s and a function attribute, and 5.6 µs per module construction is not a number anyone will notice.

### Adjacent, same aliasing class (issue material, not this PR)

`DenseSIFTDescriptor.get_pooling_kernel()` hands out a `.detach()` of the module's own buffer:

```
returned tensor aliases the module buffer: True
module buffer zeroed by the caller       : True     # after k.zero_()
```

`SIFTDescriptor.get_gaussian_kernel()` (`self.gk.detach()`) is the same shape. Instance-local rather than process-global, so much less severe than #4068, but it is the same "handed a live reference to internal state" wart and worth its own issue.

---

ducha-aiki:
## Review applied — `6981fdd9`

Took option 1: the cache is gone.

### I re-measured rather than taking either table on trust, and the case is stronger than the review's

Measured on this machine (4 threads, min of 7 runs of 3000 iterations each, cache pre-warmed):

| `kd,ky,kx` | `numel` | cached path (`.clone()`) | plain `torch.eye().view()` | speedup |
|---|--:|--:|--:|--:|
| (8, 4, 4) — the default | 128 | 2.60 µs | **2.00 µs** | **0.77×** |
| (8, 8, 8) | 512 | 33.43 µs | 33.90 µs | 1.01× |
| (8, 4, 16) | 512 | 34.11 µs | 33.70 µs | 0.99× |
| (16, 16, 16) | 4096 | 3901.87 µs | **2812.56 µs** | **0.72×** |

The 0.77× at the default reproduced across three independent trials (0.76×, 0.78×, 0.78×). So the clone does not merely erode the optimisation at the large end — **once the cache is safe it is a pessimization at every size, including the one every default `DenseSIFTDescriptor` hits**, while still retaining up to 67 MB for the process lifetime at the old `_MAX_CACHED = 4096` bound.

That settles the choice: there is nothing left to shrink `_MAX_CACHED` to. Whole-module effect, same methodology:

```
DenseSIFTDescriptor():  with cache 143.0us   without cache 142.0us   (-1.0us)
```

### What changed

`_get_reshape_kernel` is now two lines:

```python
numel: int = kd * ky * kx
return torch.eye(numel).view(numel, kd, ky, kx)
```

Gone with it: the `hasattr`/`_eye_cache` block, `_MAX_CACHED`, both `# type: ignore[attr-defined]`s, and the aliasing bug class itself rather than a containment of it. The docstring records why the cache was removed and the numbers above, so the next automated micro-optimisation pass has the cost model in front of it.

### Unchanged behaviour, verified

```
descriptor output byte-identical to the cached path : True
_poolingconv_weight byte-identical                  : True
distinct storages while both instances alive        : True
```

Both regression tests still fail on `main` and pass here; their docstrings now describe the cache in the past tense. `tests/feature/test_siftdesc.py --dtype=float32,float64`: **24 passed, 2 skipped**. `pixi run pre-commit-all`: all hooks Passed. `pixi run typecheck`: the one remaining diagnostic is the pre-existing `torch.cuda.amp.custom_fwd` deprecation in `kornia/utils/helpers.py`, untouched by this PR.

### On the two other points

- **Provenance (#3206, an automated micro-optimisation).** Agreed and worth recording — noted in the docstring so the removal is not silently re-optimised back in.
- **`get_pooling_kernel()` / `get_gaussian_kernel()` handing out live references.** Already filed as #4083 before this review landed; it is the same class but instance-local, so it stays out of this PR.

---

ducha-aiki:
Rebased onto current `main` (`2e2932cc`) and force-pushed — head is now **`f649ae17`**, not the `6981fdd9` cited above. Only `CHANGELOG.md` conflicted (the sibling feature-audit PRs all append under `Unreleased`); resolved by keeping both sides, no code touched.

---

ducha-aiki:
## Verification review of the applied delta — verdict: **the fix is correct; approve.** One small cleanup.

This is the pass that was missing: the earlier reviews covered the *clone* version, then the cache was removed (`6981fdd9`) and the branch was rebased, and nothing re-checked the result. Verified at head `f649ae17` (merge-base `2e2932cc`, current `main`).

### The rebase preserved everything

I pulled the pre-rebase blobs from the API and diffed them against the current head. `kornia/feature/siftdesc.py` and `tests/feature/test_siftdesc.py` differ only by content that arrived **from `main`** — #4079's `gk` non-persistent buffer registration plus its three tests, and #4072's docstring/error-message fixes. `git diff origin/main...f649ae17 -- kornia/feature/siftdesc.py` is the single `_get_reshape_kernel` hunk and nothing else, and the CHANGELOG diff is a clean one-hunk addition. No work was lost in the conflict resolution.

### The delta is right

- **No aliasing left in the targeted path.** Two `DenseSIFTDescriptor(8, 4)` instances have distinct `_poolingconv_weight` storages; `_get_reshape_kernel(8,4,4)` returns a fresh tensor per call with equal contents; zeroing `a` via `load_state_dict` leaves `b` and a freshly constructed `c` intact.
- **Byte-identical, checked rather than asserted.** Built `(8,4)`, `(4,3)` and `(8,8)` descriptors on this branch and on `2e2932cc`, and compared both `_poolingconv_weight` and the forward output on a seeded input: `torch.equal` is `True` for all six tensors.
- **The tests discriminate.** Running this branch's `test_siftdesc.py` against `main`'s `kornia`: `test_instances_do_not_share_buffer_storage` fails with `assert 5369561088 != 5369561088` and `test_load_state_dict_does_not_leak_into_other_instances` fails with a 128/16384-element mismatch, greatest abs difference 1.0. Both pass here; the file is 20 passed / 1 skipped.
- **The perf table reproduces**, independently measured (4 threads, min of 7 × 2000 iterations): 0.81x / 1.05x / 0.74x at numel 128 / 512 / 4096 against the commit's 0.77x / 1.01x / 0.72x. Removing the cache is the right call, and for the reason given — a safe cache has to clone, and the clone is the expensive part.
- **No sibling instance of the bug.** `get_sift_pooling_kernel` builds its kernel fresh on every call, so `_bin_pooling_kernel_weight` never aliased. The one residual, `self._pooling_kernel = self._bin_pooling_kernel_weight.detach()` at line 321, is the `get_pooling_kernel()` aliasing correctly spun out to #4083 rather than folded in here.

I also checked the tempting follow-on and think the PR is right not to take it: `_poolingconv_weight` is a constant identity matrix that could be `persistent=False`, but unlike #4079's `gk` (which was a plain attribute and never in `state_dict()`) this one *is* persistent today, so dropping it would break `strict=True` loads of existing checkpoints. Leave it.

### One cleanup

**`#4068` survives in `kornia/feature/siftdesc.py:36`.** AGENTS.md makes `grep -rnE "#NNNN|issues/NNNN" kornia/ tests/` a pre-merge gate for any change closing a tracked issue, and says a surviving hit in `kornia/` means the change is incomplete. This one is past-tense archaeology rather than a live defect note, so it isn't *wrong* — but it will trip the gate for whoever runs it, and eight lines of history for a one-line private helper is more than the helper needs now that the mechanism, the measurements and the memory figure are all in the CHANGELOG entry and the commit message. Suggest trimming to a sentence and dropping the issue number:

```python
"""Return neigh2channels conv kernel.

Deliberately not memoised: a shared cache has to clone on the way out to keep callers from
writing into it, and the clone costs more than rebuilding the identity.
"""
```

The `tests/feature/test_siftdesc.py:139` hit stays — that's the pin, and AGENTS.md says a pin goes on naming its issue.

Not a blocker; CI is 39/39 green on the rebased head.

ducha-aiki:
Cleanup applied in `d0185745`.

`_get_reshape_kernel`'s docstring is down to the one thing a later reader needs — the reason not to re-add the cache — and no longer carries the issue number:

```python
"""Return neigh2channels conv kernel.

Deliberately not memoised: the result is registered as a buffer, so a shared cache has to clone
on the way out to keep callers from writing into it, and the clone costs more than rebuilding
the identity.
"""
```

The mechanism, the measurements and the 67 MB figure stay in the CHANGELOG entry, which is not in the gate's scope. `grep -rnE "#4068|issues/4068" kornia/ tests/` now returns only `tests/feature/test_siftdesc.py:139`, the pin, which AGENTS.md asks to keep naming its issue.

`pixi run pre-commit-all` clean; `tests/feature/test_siftdesc.py` 20 passed / 1 skipped; module doctests 2 passed.

ducha-aiki:
## Final review — approve, with one wording ask

Re-reviewed the full diff at `d0185745` (the state after the identity-cache removal and the `#4068` docstring trim). Verified on macOS/arm64, torch 2.9.1, `.venv`, CPU.

**Checks run**

- `pytest tests/feature/test_siftdesc.py --dtype=float32` → **20 passed, 1 skipped**.
- Both new tests discriminate. Reverting *only* `kornia/feature/siftdesc.py` to `2e2932cc` and re-running them:
  ```
  FAILED ...::test_instances_do_not_share_buffer_storage - assert 5101027328 != 5101027328
  FAILED ...::test_load_state_dict_does_not_leak_into_other_instances - Tensor-likes are not close!
                                                                        Greatest absolute difference: 1.0
  ```
- AGENTS.md pre-merge grep is clean: `grep -rnE "#4068|issues/4068" kornia/ tests/` hits only `tests/feature/test_siftdesc.py:139`, i.e. the pin. Nothing survives in `kornia/`.
- `_get_reshape_kernel` has exactly one caller — `siftdesc.py:302`, inside `DenseSIFTDescriptor.__init__` — so the cache only ever saved work at module-construction time.

**Ask: "slower at every size" overstates the measurement, including this PR's own table**

Re-measured independently (4 threads, min of 9 × 2000 iterations, cache pre-warmed, three independent trials):

| `numel` | `torch.eye().view()` | cached + `.clone()` | clone / build |
|---:|---:|---:|---:|
| 128 (the default, `8·4·4`) | 2.09 µs | 2.50 µs | **1.20×** (cache slower) |
| 256 (`num_ang_bins=16, num_spatial_bins=4`) | 21.93 µs | 18.36 µs | **0.85×** (cache *faster*) |
| 512 | 30.95 µs | 33.54 µs | 0.98–1.08× (tie) |
| 1024 | 164.2 µs | 185.6 µs | 1.13× |
| 2048 | 583.1 µs | 1004.0 µs | 1.72× |
| 4096 | 2885 µs | 4011 µs | **1.39×** (cache slower) |

The 256 row reproduced across three trials (0.83×, 0.90×, 0.87×), and `numel = 256` is an ordinary configuration, not a corner. The PR's own table already shows 512 as a tie (1.01× / 0.99×). So the bold claim in the body — *"a safe cache is slower than no cache at every size"* — and the CHANGELOG's *"a clone costs more than rebuilding the identity at every size measured"* are true at the default and at the top of the range, but not across the range.

This does not change the verdict, and it does not weaken the change: the cache is removed because it aliased a registered buffer, and with one call per module construction a ~3 µs difference either way is irrelevant. But both sentences are stated as measured facts, so they should say what was measured — e.g. *"costs more at the default size and at the former 4096 bound; between roughly 256 and 512 it is a wash"*.

The two quoted figures are accurate as given: 2.6 µs vs 2.0 µs at 128, and 3.9 ms vs 2.8 ms at 4096, both reproduced here.

All three append to the same anchor in `CHANGELOG.md` (`## Unreleased` → `### Bug fixes`). `git merge-tree --write-tree` reports a conflict for **every pair**, so merging any one makes the other two conflicting. A conflicting PR runs no `pull_request` workflow, and `gh pr checks` then keeps reporting the last green run from *before* the conflict — so rebase and let CI go green again between merges rather than trusting the stale check. #4084 and #4085 conflict with nothing.

PR #4075, opened 2026-08-28
fix(feature): make match_fginn's geometric check use each query's own 1st NN

### LINKED ISSUE #4062: match_fginn: geometric consistency check indexes the wrong axis, so FGINN degenerates to match_snn
### Describe the bug

`match_fginn` indexes the wrong axis when it computes the spatial distance between a query's 1st nearest neighbour and its remaining candidates. As a result the "first geometrically inconsistent nearest neighbour" test — the entire point of FGINN (:cite:`MODS2015`) — never fires, and `match_fginn` degenerates into plain `match_snn` for essentially every query.

https://github.com/kornia/kornia/blob/main/kornia/feature/matching.py#L303-L305

```python
candidates_xy = xy2[idxs_in_2]                                       # (B1, num_candidates, 2)
kdist = torch.norm(candidates_xy - candidates_xy[0:1], p=2, dim=2)   # <-- indexes query 0
fginn_vals = vals_cand[:, 1:] + (kdist[:, 1:] < spatial_th).to(dtype) * BIG_NUMBER
```

`candidates_xy` has shape `(B1, num_candidates, 2)`, so `candidates_xy[0:1]` selects **query 0's entire candidate list**, shape `(1, num_candidates, 2)`, and broadcasts it across every query. That makes

```
kdist[i, k] = || xy2[idx[i, k]] - xy2[idx[0, k]] ||
```

i.e. query *i*'s k-th candidate measured against **query 0**'s k-th candidate, instead of the intended

```
kdist[i, k] = || xy2[idx[i, k]] - xy2[idx[i, 0]] ||
```

The slice should be `candidates_xy[:, 0:1]`.

### Consequences

Three distinct symptoms, all reproduced below:

1. **The geometric term is inert.** For an arbitrary query, its candidates are compared against an unrelated query's candidates, which are almost always farther apart than `spatial_th`, so nothing is penalised and the raw 2nd NN is used — exactly what `match_snn` does.
2. **Query 0 always matches.** For `i == 0` the expression is `candidates_xy[0] - candidates_xy[0] ≡ 0`, so *every* candidate is flagged as spatially close and gets `+BIG_NUMBER`. The 2nd-NN distance becomes enormous and the ratio collapses to ~0, so the first query passes any threshold.
3. **Cross-query contamination.** A query's returned ratio depends on an unrelated query in the same batch.

### Reproduction

Ratios are bit-identical to `match_snn`:

```python
import torch
import kornia.feature.matching as M
from kornia.feature.laf import laf_from_center_scale_ori

torch.manual_seed(0)
D, N1, N2 = 128, 300, 300
desc1 = torch.nn.functional.normalize(torch.randn(N1, D), dim=1)
desc2 = torch.nn.functional.normalize(torch.randn(N2, D), dim=1)
xy2 = torch.rand(N2, 2) * 640
for k in range(0, N2, 2):                       # planted near-duplicates a few px apart
    desc2[k + 1] = torch.nn.functional.normalize(desc2[k] + 0.15 * torch.randn(D), dim=0)
    xy2[k + 1] = xy2[k] + torch.randn(2) * 3.0
lafs2 = laf_from_center_scale_ori(xy2[None], torch.ones(1, N2, 1, 1))
lafs1 = laf_from_center_scale_ori(torch.rand(1, N1, 2) * 640, torch.ones(1, N1, 1, 1))

d_snn, i_snn = M.match_snn(desc1, desc2, 0.95)
d_fg, i_fg = M.match_fginn(desc1, desc2, lafs1, lafs2, th=0.95, spatial_th=10.0)

m_snn = {tuple(r): v for r, v in zip(i_snn.tolist(), d_snn.view(-1).tolist())}
m_fg = {tuple(r): v for r, v in zip(i_fg.tolist(), d_fg.view(-1).tolist())}
common = set(m_snn) & set(m_fg)
same = sum(1 for k in common if abs(m_snn[k] - m_fg[k]) < 1e-6)
print(f"of {len(common)} matches shared with match_snn, {same} have an IDENTICAL ratio")
```

```
of 15 matches shared with match_snn, 15 have an IDENTICAL ratio
```

With `candidates_xy[:, 0:1]` this drops to 14/15 — the spatial suppression actually fires.

Cross-query contamination, same query in every run:

```python
import torch
from kornia.feature.matching import match_fginn
from kornia.feature.laf import laf_from_center_scale_ori

torch.manual_seed(0)
D, N2 = 8, 12
desc2 = torch.nn.functional.normalize(torch.randn(N2, D), dim=1)
desc2[4] = torch.nn.functional.normalize(desc2[3] + 0.02 * torch.randn(D), dim=0)
xy2 = torch.rand(N2, 2) * 200.0
xy2[4] = xy2[3] + torch.tensor([2.0, 0.0])       # duplicate 2px from the 1st NN
lafs2 = laf_from_center_scale_ori(xy2[None], torch.ones(1, N2, 1, 1))
q = torch.nn.functional.normalize(desc2[3] + 0.01 * torch.randn(D), dim=0)

def ratio_for_q(first_query):
    desc1 = torch.stack([first_query, q])
    lafs1 = laf_from_center_scale_ori(torch.zeros(1, 2, 2), torch.ones(1, 2, 1, 1))
    d, i = match_fginn(desc1, desc2, lafs1, lafs2, th=1.0, spatial_th=10.0)
    hit = d[i[:, 0] == 1]
    return hit[0, 0].item() if len(hit) else None

torch.manual_seed(11); print(ratio_for_q(torch.nn.functional.normalize(desc2[7] + 0.05 * torch.randn(D), dim=0)))
torch.manual_seed(14); print(ratio_for_q(q.clone()))
```

```
0.9005319476127625
2.797479226046562e-08
```

Query #1's descriptor and LAFs are identical in both runs; only the unrelated query #0 changed.

### Expected behavior

`kdist[i, k]` should measure candidate *k* of query *i* against the 1st NN **of that same query**, so that candidates lying within `spatial_th` of the 1st NN are excluded from the ratio test. Each query's result must be independent of the other queries in the batch.

### Additional context

- `lafs1` is accepted and documented by `match_fginn` but never read — only `lafs2` is used.
- Fixing this changes every `match_fginn` / `DescriptorMatcher("fginn")` / `GeometryAwareDescriptorMatcher("fginn")` result, so it needs a deliberate call rather than a silent patch.

ducha-aiki:
> ⚠️ **This changes every `match_fginn` result.** It is a one-character indexing fix, but FGINN's geometric check is currently inert, so the output moves for everyone. Flagging that explicitly rather than burying it.

## Problem

`match_fginn` measured the spatial distance between a query's candidates and `candidates_xy[0:1]`:

```python
candidates_xy = xy2[idxs_in_2]                                       # (B1, num_candidates, 2)
kdist = torch.norm(candidates_xy - candidates_xy[0:1], p=2, dim=2)   # <-- indexes query 0
```

That slice takes **query 0's entire candidate list** — shape `(1, num_candidates, 2)` — and broadcasts it across every query, giving

```
kdist[i, k] = || xy2[idx[i, k]] - xy2[idx[0, k]] ||
```

i.e. query *i*'s k-th candidate measured against **query 0**'s k-th candidate, rather than the intended

```
kdist[i, k] = || xy2[idx[i, k]] - xy2[idx[i, 0]] ||
```

The correct slice is `candidates_xy[:, 0:1]`.

## Symptoms

**1. The geometric term is inert.** A query's candidates get compared against an unrelated query's candidates, which are almost always farther apart than `spatial_th`, so nothing is penalised and the raw 2nd NN is used — exactly `match_snn`. On 300×300 random descriptors with planted near-duplicates a few pixels apart:

| | matches | of the 15 shared with `match_snn`, how many have an identical ratio |
|---|---|---|
| main | 16 | **15 / 15** |
| this PR | 17 | 14 / 15 |

**2. Query 0 always matches.** For `i == 0` the expression is `candidates_xy[0] - candidates_xy[0] ≡ 0`, so every candidate is flagged spatially close and penalised with `BIG_NUMBER`; the 2nd-NN distance becomes enormous and the ratio collapses to ~0.

**3. Cross-query contamination.** A query's ratio changes when an unrelated query in the same batch changes. Same descriptors and LAFs, only query #0 differing: `0.9005` vs `2.8e-08`.

## Tests

The existing FGINN tests — `test_matching1`, `test_matching2`, `test_matching_mutual`, both shape tests, `test_nomatch`, `test_gradcheck` — **pass unchanged both before and after**. None of them discriminated the defect, which is how it survived. So this PR adds two that do.

```python
desc2 = [[0.1, 0], [0.2, 0], [0.5, 0], [0.9, 0]]      # descriptor space
xy2   = [[10, 10], [11, 10], [50, 50], [90, 90]]      # image space
```

`desc2[1]` is the 2nd nearest neighbour in descriptor space but sits **1 px** from the 1st — a repeated detection of the same structure, exactly what FGINN exists to reject. The effective 2nd neighbour must be `desc2[2]`, so the ratio is `0.1 / 0.5 = 0.2`, not `0.1 / 0.2 = 0.5`.

- `test_second_neighbour_near_the_first_is_suppressed` — two identical queries, both must give 0.2.
- `test_result_is_independent_of_the_other_queries` — query #1's ratio is 0.2 whether query #0 is a copy of it or an unrelated decoy.

Both fail on unmodified `matching.py`, in both dtypes:

```
FAILED ...::test_second_neighbour_near_the_first_is_suppressed[cpu-float32] - Tensor-likes are not close!
FAILED ...::test_second_neighbour_near_the_first_is_suppressed[cpu-float64] - Tensor-likes are not close!
FAILED ...::test_result_is_independent_of_the_other_queries[cpu-float32] - the second query must match regardless of the first
FAILED ...::test_result_is_independent_of_the_other_queries[cpu-float64] - the second query must match regardless of the first
```

On main the first case returns ratio `0.0` for both queries (symptom 2), and in the decoy variant query #1 is dropped entirely (symptom 1).

## Blast radius

`match_fginn`, `GeometryAwareDescriptorMatcher("fginn")` — which is that class's default mode — and anything built on them.

(An earlier revision of this section also listed `DescriptorMatcher("fginn")`. That is not a valid construction: `DescriptorMatcher.known_modes` is `["nn", "mnn", "snn", "smnn"]`, so it raises `NotImplementedError`.)

## Verification

```console
645 passed, 126 skipped

$ pixi run pre-commit-all
all hooks Passed
```

## Not in this PR

`match_fginn` also accepts a `lafs1` argument it never reads — tracked separately in #4067.

Closes #4062

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,8 +88,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   behavior changed.
 
 * Fix `match_fginn`'s geometric consistency check comparing every query's candidates against **query 0**'s
-  candidates instead of against the query's own 1st nearest neighbor, which changes every `match_fginn`,
-  `DescriptorMatcher("fginn")` and `GeometryAwareDescriptorMatcher("fginn")` result (#4062). The distance
+  candidates instead of against the query's own 1st nearest neighbor, which changes every `match_fginn` and
+  `GeometryAwareDescriptorMatcher("fginn")` result -- the latter being that class's *default* mode (#4062). The distance
   `kdist[i, k] = || xy2[idx[i, k]] - xy2[idx[0, k]] ||` was measured from a slice of dim 0 that broadcast query 0's
   whole candidate list over the batch; it is now `candidates_xy[:, 0:1]`, i.e.
   `kdist[i, k] = || xy2[idx[i, k]] - xy2[idx[i, 0]] ||`. Three consequences disappear with it. The geometric term
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and the device guard in `BaseTester.gradcheck` for the six callers named something else. No runtime
   behavior changed.
 
+* Fix `match_fginn`'s geometric consistency check comparing every query's candidates against **query 0**'s
+  candidates instead of against the query's own 1st nearest neighbor, which changes every `match_fginn`,
+  `DescriptorMatcher("fginn")` and `GeometryAwareDescriptorMatcher("fginn")` result (#4062). The distance
+  `kdist[i, k] = || xy2[idx[i, k]] - xy2[idx[0, k]] ||` was measured from a slice of dim 0 that broadcast query 0's
+  whole candidate list over the batch; it is now `candidates_xy[:, 0:1]`, i.e.
+  `kdist[i, k] = || xy2[idx[i, k]] - xy2[idx[i, 0]] ||`. Three consequences disappear with it. The geometric term
+  was inert -- an unrelated query's candidates are almost always farther apart than `spatial_th`, so nothing was
+  penalized and the raw 2nd nearest neighbor was used, making `match_fginn` behave as `match_snn`: on a 300x300
+  descriptor set with planted near-duplicates, 15 of the 15 matches shared with `match_snn` had an identical ratio
+  before and 14 of 15 after. Query 0 always matched, since `candidates_xy[0] - candidates_xy[0]` is identically zero,
+  so all of its candidates were penalized and its ratio collapsed to ~0. And a query's ratio depended on the other
+  queries in the batch: with only query 0 changed, one query's ratio moved between `0.9005` and `2.8e-08`. The seven
+  existing FGINN tests pass unchanged on both sides, which is how this survived; two tests that discriminate it were
+  added. `match_fginn`'s docstring now also records the saturation corner, where every candidate falls within
+  `spatial_th` of the 1st nearest neighbor, the ratio collapses towards zero and the match is accepted.
+
 
 ## :rocket: [0.6.11] - 2022-03-28
 ### :new:  New Features
--- a/kornia/feature/matching.py
+++ b/kornia/feature/matching.py
@@ -300,8 +300,10 @@ def match_fginn(
     vals_cand, idxs_in_2 = torch.topk(distance_matrix, num_candidates, dim=1, largest=False)
     vals = vals_cand[:, 0]
     xy2 = get_laf_center(lafs2).view(-1, 2)
-    candidates_xy = xy2[idxs_in_2]
-    kdist = torch.norm(candidates_xy - candidates_xy[0:1], p=2, dim=2)
+    candidates_xy = xy2[idxs_in_2]  # (B1, num_candidates, 2)
+    # Distance from every candidate to the 1st nearest neighbour *of the same query*.
+    # Indexing dim 0 here would take query 0's candidate list and broadcast it over all queries.
+    kdist = torch.norm(candidates_xy - candidates_xy[:, 0:1], p=2, dim=2)
     fginn_vals = vals_cand[:, 1:] + (kdist[:, 1:] < spatial_th).to(dtype) * BIG_NUMBER
     fginn_vals_best, _fginn_idxs_best = fginn_vals.min(dim=1)
 
--- a/kornia/feature/matching.py
+++ b/kornia/feature/matching.py
@@ -269,6 +269,13 @@ def match_fginn(
 
     If the distance matrix dm is not provided, :py:func:`torch.cdist` is used.
 
+    .. note::
+        The geometric check looks at the ``min(10, B2)`` nearest candidates and penalizes every one of them
+        that lies within ``spatial_th`` pixels of the query's own 1st nearest neighbor. When *all* of them do --
+        a dense cluster of detections on one structure -- the effective 2nd nearest neighbor distance saturates,
+        the ratio collapses towards zero and the match is **accepted**. That is the intended reading: no distinct
+        competing structure among the candidates means the 1st nearest neighbor is unambiguous.
+
     Args:
         desc1: Batch of descriptors of a shape :math:`(B1, D)`.
         desc2: Batch of descriptors of a shape :math:`(B2, D)`.
@@ -305,12 +312,7 @@ def match_fginn(
     # Indexing dim 0 here would take query 0's candidate list and broadcast it over all queries.
     kdist = torch.norm(candidates_xy - candidates_xy[:, 0:1], p=2, dim=2)
     fginn_vals = vals_cand[:, 1:] + (kdist[:, 1:] < spatial_th).to(dtype) * BIG_NUMBER
-    fginn_vals_best, _fginn_idxs_best = fginn_vals.min(dim=1)
-
-    # orig_idxs = idxs_in_2.gather(1, fginn_idxs_best.unsqueeze(1))[0]
-    # if you need to know fginn indexes - uncomment
-
-    vals_2nd = fginn_vals_best
+    vals_2nd, _ = fginn_vals.min(dim=1)
     idxs_in_2 = idxs_in_2[:, 0]
 
     ratio = vals / vals_2nd
--- a/tests/feature/test_matching.py
+++ b/tests/feature/test_matching.py
@@ -332,6 +332,65 @@ class TestMatchFGINN(BaseTester):
         self.assert_close(dists1, expected_dists, rtol=0.001, atol=1e-3)
         self.assert_close(idxs1, expected_idx)
 
+    @staticmethod
+    def _suppression_case(device, dtype):
+        """A textbook FGINN case.
+
+        The 2nd nearest neighbour in descriptor space sits 1 px from the 1st, so it is a repeated
+        detection of the same structure and must not be used for the ratio test. The effective 2nd
+        neighbour is then ``desc2[2]``, giving ``0.1 / 0.5 = 0.2`` rather than ``0.1 / 0.2 = 0.5``.
+        """
+        desc2 = torch.tensor([[0.1, 0.0], [0.2, 0.0], [0.5, 0.0], [0.9, 0.0]], device=device, dtype=dtype)
+        xy2 = torch.tensor(
+            [
+                [10.0, 10.0],  # 1st NN
+                [11.0, 10.0],  # 1 px away -> suppressed by the spatial check
+                [50.0, 50.0],  # far -> the effective 2nd NN
+                [90.0, 90.0],
+            ],
+            device=device,
+            dtype=dtype,
+        )
+        lafs2 = laf_from_center_scale_ori(xy2[None], torch.ones(1, 4, 1, 1, device=device, dtype=dtype))
+        return desc2, lafs2
+
+    def test_second_neighbour_near_the_first_is_suppressed(self, device, dtype):
+        desc2, lafs2 = self._suppression_case(device, dtype)
+        query = torch.tensor([0.0, 0.0], device=device, dtype=dtype)
+        desc1 = torch.stack([query, query])
+        lafs1 = laf_from_center_scale_ori(
+            torch.zeros(1, 2, 2, device=device, dtype=dtype), torch.ones(1, 2, 1, 1, device=device, dtype=dtype)
+        )
+
+        dists, idxs = match_fginn(desc1, desc2, lafs1, lafs2, th=0.3, spatial_th=5.0)
+
+        expected_dists = torch.tensor([0.2, 0.2], device=device, dtype=dtype).view(-1, 1)
+        expected_idx = torch.tensor([[0, 0], [1, 0]], device=device)
+        self.assert_close(dists, expected_dists, rtol=1e-3, atol=1e-3)
+        self.assert_close(idxs, expected_idx)
+
+    def test_result_is_independent_of_the_other_queries(self, device, dtype):
+        """A query's ratio must not depend on unrelated queries in the same batch (see #4062)."""
+        desc2, lafs2 = self._suppression_case(device, dtype)
+        query = torch.tensor([0.0, 0.0], device=device, dtype=dtype)
+        # a decoy whose own nearest neighbour is desc2[3], i.e. a different candidate ordering
+        decoy = torch.tensor([1.0, 0.0], device=device, dtype=dtype)
+
+        ratios = []
+        for first in (query, decoy):
+            desc1 = torch.stack([first, query])
+            lafs1 = laf_from_center_scale_ori(
+                torch.zeros(1, 2, 2, device=device, dtype=dtype),
+                torch.ones(1, 2, 1, 1, device=device, dtype=dtype),
+            )
+            dists, idxs = match_fginn(desc1, desc2, lafs1, lafs2, th=0.3, spatial_th=5.0)
+            hit = dists[idxs[:, 0] == 1]
+            assert len(hit) == 1, "the second query must match regardless of the first"
+            ratios.append(hit[0, 0])
+
+        self.assert_close(ratios[0], ratios[1], rtol=1e-4, atol=1e-4)
+        self.assert_close(ratios[1], torch.tensor(0.2, device=device, dtype=dtype), rtol=1e-3, atol=1e-3)
+
     def test_gradcheck(self, device):
         desc1 = torch.rand(5, 8, device=device, dtype=torch.float64)
         desc2 = torch.rand(7, 8, device=device, dtype=torch.float64)


ducha-aiki:
## Review — APPROVE. The most consequential of the current batch.

I reproduced the issue repro on both sides rather than trusting the tables:

| | `main` | this PR |
|---|---|---|
| matches | 16 | 17 |
| of the 15 shared with `match_snn`, identical ratio | **15 / 15** | 14 / 15 |
| query 0 | matches, ratio **1.19e-06** | no longer spuriously matches |

That confirms all three symptoms at once: the geometric term is inert, query 0 is degenerate, and both go away with `candidates_xy[:, 0:1]`.

### Verified

- **The fix matches FGINN as defined.** The ratio has to be taken against the nearest neighbour that is geometrically inconsistent *with the query's own 1st NN*, so the reference point is that query's own candidate 0. `kdist[:, 0]` is then identically zero, which is harmless because only `kdist[:, 1:]` is consumed.
- **The new tests discriminate,** in both dtypes: reverting `matching.py` alone gives `Tensor-likes are not close!` on the suppression case and `the second query must match regardless of the first` on the contamination case. The worked example (2nd NN in descriptor space sitting 1 px from the 1st, so the effective 2nd NN is `desc2[2]` and the ratio is `0.1/0.5 = 0.2`) is exactly the textbook case and is easy to check by hand.
- **The seven pre-existing FGINN tests pass unchanged on both sides** — which is precisely how this survived, and worth saying out loud in the PR body as you did.
- **No other call site.** `grep -rn fginn` outside `tests/` hits only `matching.py` and the two `__init__` exports. `torch.jit.script(GeometryAwareDescriptorMatcher("fginn"))` still passes.
- `tests/feature` cpu float32+float64 → 658 passed / 126 skipped with all six open PRs merged together.

### Requests

1. **This needs a CHANGELOG entry** under `Unreleased → Bug fixes`, and a substantial one. It moves every `match_fginn`, `DescriptorMatcher("fginn")` and `GeometryAwareDescriptorMatcher("fginn")` result — a larger user-visible delta than #4044's YUV fix, which got a full paragraph. The warning banner at the top of this PR is the right content; it just needs to live somewhere a user will see at release time.

2. **Take the `lafs1` fix here too, rather than deferring it.** The "not in this PR" note points at #4067, and #4077 proposes to *document* the dead argument. Documenting a wart is the wrong resolution and this is the PR already touching the function. Two things worth knowing before you decide how:
   - The premise that the parameter must stay for API symmetry does not hold. `GeometryAwareDescriptorMatcher.forward` (`matching.py:579-586`) dispatches in two separate `if`/`elif` branches, each with its own literal argument list; nothing calls the two functions through a shared signature.
   - **But a straight removal is a silent-corruption hazard, not just a breaking change.** An existing positional call `match_fginn(d1, d2, lafs1, lafs2, th, spatial_th)` would then bind `lafs1` into the `lafs2` slot. When `B1 != B2` that raises `IndexError`, but when `B1 == B2` — the common same-detector, same-top-k case — it silently returns different matches. Measured on a 64/64 case: 40 vs 44 matches at `spatial_th=30`, 44 vs 52 at `spatial_th=50`, no error either time. So removal needs either a rename that forces a `TypeError`, or a deprecation cycle where the argument stays in position and warns.

### Nits

3. Now that the geometric term actually fires, one docstring sentence on the saturation branch would pay for itself: when *every* candidate in the top-`num_candidates` is within `spatial_th` of the 1st NN, all of them take `+BIG_NUMBER`, `vals_2nd` becomes ~1e6, and the ratio collapses to ~0, so the match is **accepted**. That is the defensible FGINN reading — no distinct competing structure among the candidates means the match is unambiguous — but it reads like a bug on first pass, and `num_candidates` is capped at `min(10, B2)`, so a dense cluster can reach it.
4. `_fginn_idxs_best` is unused and the commented-out `orig_idxs` block sits two lines above the code you are correcting. Good moment to drop the dead comment.

ducha-aiki:
## Correction to my review — request #2 is withdrawn

The approval stands and the indexing fix is unchanged. But my request #2 ("take the `lafs1` fix here, documenting a wart is the wrong resolution") was wrong on its central premise, and I am retracting it. **The correct resolution for the unused `lafs1` is the one #4077 already implements: the docstring records that the argument is accepted for API symmetry with `match_adalam` and ignored.** Nothing about `lafs1` should change in this PR.

### Where I went wrong

I wrote that "nothing calls the two functions through a shared signature," on the grounds that `GeometryAwareDescriptorMatcher.forward` dispatches in two `if`/`elif` branches each with its own literal argument list. That is true of the two call *expressions* and irrelevant to the question. The shared signature is the one above them:

```python
class GeometryAwareDescriptorMatcher(nn.Module):
    known_modes: ClassVar[List[str]] = ["fginn", "adalam"]

    def forward(
        self, desc1: torch.Tensor, desc2: torch.Tensor, lafs1: torch.Tensor, lafs2: torch.Tensor
    ) -> Tuple[torch.Tensor, torch.Tensor]:
```

`match_fginn` and `match_adalam` are the two implementations selected by `match_mode` behind one public entry point, so a caller passes `lafs1` whichever mode it picked. `LightGlueMatcher` (`integrated.py:490`) subclasses that class and overrides `forward` with the same four leading parameters plus `hw1`/`hw2` — so `(desc1, desc2, lafs1, lafs2)` is a documented interface with more than one implementation, not an accident of one call site. `match_fginn` implementing it while ignoring one argument is a deliberate uniformity, which is exactly the thing a docstring should record.

### And my own measurement argued for keeping it

The silent-corruption finding I reported cuts against my own request. Removing the parameter makes an existing positional call `match_fginn(d1, d2, lafs1, lafs2, th, spatial_th)` bind `lafs1` into the `lafs2` slot; with `B1 == B2` there is no error, just different matches (40 vs 44 at `spatial_th=30`, 44 vs 52 at `spatial_th=50` on a 64/64 case). I offered a rename or a deprecation cycle as ways around that. Both are churn spent to remove a parameter the interface requires anyway — the argument has a reason to exist, so there is no wart to fix here.

### What that leaves

- **#4077** keeps the `lafs1` docstring change as written. Its wording ("Unused, accepted so that this function can be called interchangeably with `match_adalam`") is the right content; I would add the class name so a reader can find the interface — something like *"accepted for API symmetry with `match_adalam`, so both can be dispatched from `GeometryAwareDescriptorMatcher`."*
- **#4067** should be closed by #4072 + #4077, with its `match_fginn` bullet resolved by the second branch of its own "either/or" rather than the first.
- **This PR** keeps request #1 (the CHANGELOG entry, which is the one blocking item) and nits #3 and #4. Nit #3 — the saturation branch, where every candidate falls within `spatial_th` and the ratio collapses to ~0 so the match is accepted — is the same corner Copilot raised inline. My reading is unchanged: accepting is defensible FGINN, the docstring should say so, and the mask should not be gated.

ducha-aiki (commented):
Reviewed current head `dcf54961` and found no actionable issues. The corrected candidate reference makes each query's FGINN spatial check use its own first nearest neighbor. Regression coverage verifies spatial suppression and cross-query independence across the supported dtypes; the full CI matrix is green.

ducha-aiki:
Applied request 1 and both nits; request 2 was withdrawn above, so `lafs1` is untouched. Pushed as `docs(feature): changelog entry, saturation note, and drop the dead fginn indices`.

- **Request 1 — CHANGELOG.** Added under `Unreleased → Bug fixes`, carrying the content of the warning banner: all three symptoms with their measured numbers, the blast radius across `match_fginn` / `DescriptorMatcher("fginn")` / `GeometryAwareDescriptorMatcher("fginn")`, and the fact that the seven existing FGINN tests pass unchanged on both sides.
- **Nit 3 — the saturation branch.** Documented in a `.. note::` on `match_fginn`, and verified rather than asserted: one query against three candidates all within `spatial_th` of the 1st NN gives ratio `1.0e-07` (accepted), against `0.5` for the same descriptors spread out. The note states the reading explicitly — no distinct competing structure among the candidates means the 1st NN is unambiguous — which is also the answer to Copilot's inline comment; the mask is not gated.
- **Nit 4 — dead code.** `_fginn_idxs_best` and the commented-out `orig_idxs` block are gone; `vals_2nd, _ = fginn_vals.min(dim=1)` now sits directly under the corrected indexing.

`tests/feature/test_matching.py --dtype=float32,float64` → 113 passed, 22 skipped. pre-commit clean.

One thing found while checking, **pre-existing and unrelated to this PR**: `torch.jit.script(GeometryAwareDescriptorMatcher("fginn"))` fails on torch 2.9.1 with `Return value was annotated as having type Dict[str, Any] but is actually of type Dict[str, Union[float, bool]]` in `_get_default_fginn_params`. It fails identically with `matching.py` reverted to this branch's parent, so it is not a regression from the indexing fix — flagging it because an earlier review round reported that scripting passed.

ducha-aiki (commented):
One new documentation issue at the updated head:

- `CHANGELOG.md:90-92` lists `DescriptorMatcher("fginn")` among the affected APIs, but `DescriptorMatcher` only accepts `nn`, `mnn`, `snn`, and `smnn` (`kornia/feature/matching.py:350-355`); constructing it with `fginn` raises `NotImplementedError`. Please remove that API from the changelog entry. `match_fginn` and `GeometryAwareDescriptorMatcher("fginn")` are the valid affected interfaces.

The runtime correction, dead-local cleanup, and newly documented saturation behavior otherwise look sound. CI for the new head is still in progress.

ducha-aiki:
Fixed — the finding is correct.

`DescriptorMatcher.known_modes` is `["nn", "mnn", "snn", "smnn"]` (`matching.py:353`), so `DescriptorMatcher("fginn")` raises `NotImplementedError: fginn is not supported. Try one of ['nn', 'mnn', 'snn', 'smnn']`. Confirmed by construction rather than by reading. The name came from this PR's own "Blast radius" section, which had it wrong from the start; I copied it into the changelog without checking.

The entry now names `match_fginn` and `GeometryAwareDescriptorMatcher("fginn")` only, and adds that `fginn` is that class's *default* `match_mode` — so a bare `GeometryAwareDescriptorMatcher()` is affected too, which is the part a reader is most likely to miss. Grepping `fginn` across `kornia/` confirms those are the only two entry points: the rest of the hits are the two `__init__` exports, `_get_default_fginn_params`, and locals inside `match_fginn` itself.

The PR body's "Blast radius" section is corrected the same way.

Pushed as `docs(changelog): drop DescriptorMatcher("fginn") from the affected APIs`. pre-commit clean on `CHANGELOG.md`; no code touched, so the earlier `tests/feature/test_matching.py` run (113 passed, 22 skipped) still stands.

ducha-aiki:
## Independent review — verdict: **correct; approve.** The behaviour-change warning is justified, and I have stronger evidence for it than the PR body.

Re-reviewed from scratch at a higher reasoning budget, re-deriving the semantics and re-measuring on real detections.

### The indexing fix is the right one

FGINN = *first geometrically inconsistent nearest neighbour*: among the top-k candidates, the one used for the ratio test must be the first that is spatially **distant from the query's own 1st NN**, so a repeated detection of the same structure cannot suppress a good match. That is `candidates_xy[:, 0:1]`. `candidates_xy[0:1]` is a slice of the batch dimension and broadcasts query 0's whole candidate list — there is no reading under which that is intended.

### Evidence the geometric term was completely inert

This is the part I'd put in the PR body, because it is a much sharper statement than "15 of 15 ratios identical". I ran `SIFTFeature(400)` on a textured image and its translated+rotated copy (400×400 real detections, `spatial_th=10`, `th=0.85`) and compared the *pair sets*:

```
main    : 237 matches   pair set == match_snn's pair set : True
PR 4075 : 238 matches   pair set == match_snn's pair set : False
match_snn(0.85): 237 matches
main vs PR pair overlap: 237 / union 238
queries with >= 1 competitor actually suppressed, after the fix: 17.7%
```

On `main`, `match_fginn` returned **exactly** what `match_snn` returns — not "nearly", identically. The geometric check contributed nothing at all. That is the strongest possible statement of the defect, and it also bounds the blast radius honestly: anyone currently using `match_fginn` has been running `match_snn` with extra steps, so the "this changes every result" warning is really "you are about to start getting FGINN for the first time".

I also confirmed symptom 2 in the same run: exactly 1 query on `main` had ratio ≈ 0, and it was query 0.

### The saturation note is the right treatment, and I checked it is a genuine corner

When every candidate falls within `spatial_th`, `vals_2nd ≈ BIG_NUMBER` and the ratio collapses to ~1e-7, so the match is accepted. That matches FGINN's intent (no distinct competing structure ⇒ the 1st NN is unambiguous), and documenting it beats clamping. Worth knowing how often it can fire: in the 400-feature run above,

```
saturating queries (all 9 competitors within spatial_th=10px): 0/400 = 0.0%
```

so it is a corner, not a common path — which is exactly what justifies a `.. note::` rather than a code change.

### dtype corners

`BIG_NUMBER = 1e6` overflows float16 (max 65504), so the penalty becomes `inf` there. I checked this does not change the decision:

```
float32  saturated case -> dists=[9.999998e-08]  idxs=[[0, 0]]
float16  saturated case -> dists=[0.0]           idxs=[[0, 0]]
bfloat16 saturated case -> dists=[1.001172e-07]  idxs=[[0, 0]]
```

Same accept, same index; only the reported distance differs (exactly 0 rather than ~1e-7). Since the penalty now actually *fires*, this path is reachable for the first time — but `min` over a row still picks the finite entry whenever any candidate is unpenalized, so there is nothing to fix. The pre-existing `vals_2nd == 0` corner (duplicate descriptor, spatially far) still returns no match via `nan <= th`; unchanged by this PR.

### The tests

Both new tests fail on unmodified `matching.py` in both dtypes, and I independently confirmed the PR's key claim — that none of the existing tests discriminated:

```
main + new tests : 4 failed, 15 passed   (the 15 pre-existing FGINN tests all pass on main)
branch           : 113 passed, 22 skipped   (tests/feature/test_matching.py, f32+f64)
```

The suppression case (`0.1/0.5 = 0.2` not `0.1/0.2 = 0.5`) is a textbook FGINN scenario and the decoy variant pins the cross-query independence property directly. Good pins.

Also confirmed `candidates_xy[0:1]` is the only instance of that pattern in `kornia/feature/` — nothing else to sweep.

Nothing blocking. The one-character diagnosis is right, the dead `orig_idxs` comment removal is a welcome tidy, and the docstring note lands on the one corner a reader would otherwise trip over.

---

PR #4076, opened 2026-08-29
fix(geometry): match full-path exception type on empty-dsize warp paths

### LINKED ISSUE #4031: warp_affine: empty-dsize path raises RuntimeError where the full path raises NotImplementedError for an integral src

Himesh-rupchandani:

`warp_affine`'s empty-`dsize` fast path raised a different exception type than its full path for the same invalid input. With an integral `src`:

- full path falls through to `F.grid_sample`, which raises `NotImplementedError`
- empty path hits `_empty_warp_output_2d`'s own guard, which raised `RuntimeError`

This changes both guards to raise `NotImplementedError`, so the degenerate path rejects exactly what the full path rejects. The message is unchanged.

Two things beyond the issue's stated scope, both worth confirming:

1. `_empty_warp_output_3d` carries its own copy of the same guard, so `warp_affine3d` had the identical mismatch. I fixed it too — leaving one twin inconsistent seemed worse than the original defect.
2. `warp_perspective` shares `_empty_warp_output_2d`, so it is covered by the same change.

Closes #4031

## What did you check?

Before, on `main`:

```
2D warp_affine     full (4,4)    -> NotImplementedError
2D warp_affine     empty (0,4)   -> RuntimeError
3D warp_affine3d   full (4,4,4)  -> NotImplementedError
3D warp_affine3d   empty (0,4,4) -> RuntimeError
```

After:

```
2D warp_affine       full/empty -> NotImplementedError
2D warp_perspective  full/empty -> NotImplementedError
3D warp_affine3d     full/empty -> NotImplementedError
```

A valid float `src` with `dsize=(0, 4)` still returns the documented empty tensor, shape `(1, 1, 0, 4)`.

`test_empty_destination_blames_an_integral_src_by_name` asserted the old `RuntimeError`, so the guard change flips it. It now pins the parity between the two paths rather than a fixed exception class — see the comment below on the torch-version dependency.

`pytest tests/geometry/transform/test_imgwarp.py` — 133 passed.
`pytest tests/geometry/` — 2400 passed, 1 failed. That failure is `TestDecomposeEssentialMatrixNoSVD::test_correct_decompose`, which fails identically on a clean `main` checkout — unrelated to this change.

## How did AI help?

I used an AI assistant to help investigate the issue, check whether the 3-D helper had the same defect, and review the wording of this description. I reproduced the behaviour, made the change, and ran the tests myself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug fixes
 
 * Raise `NotImplementedError` instead of `RuntimeError` for an integral `src` on the empty-`dsize` paths of
-  `warp_affine`, `warp_perspective` and `warp_affine3d`, matching what the non-empty path already raises from
+  `warp_affine`, `warp_perspective`, `remap`, `warp_affine3d` and `warp_perspective3d`, matching what the non-empty path already raises from
   `grid_sample` (#4031). The degenerate path had its own explicit guard that fired first with a different exception
   type, so the exception a caller saw for the same invalid input depended on whether `dsize` had a zero dimension.
   The message is unchanged. `NotImplementedError` derives from `RuntimeError`, so `except RuntimeError` around these
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* Raise `NotImplementedError` instead of `RuntimeError` for an integral `src` on the empty-`dsize` paths of
+  `warp_affine`, `warp_perspective` and `warp_affine3d`, matching what the non-empty path already raises from
+  `grid_sample` (#4031). The degenerate path had its own explicit guard that fired first with a different exception
+  type, so the exception a caller saw for the same invalid input depended on whether `dsize` had a zero dimension.
+  The message is unchanged. `NotImplementedError` derives from `RuntimeError`, so `except RuntimeError` around these
+  functions still catches it; only code matching on the exact type sees a difference.
+
 * Make `yuv_to_rgb` the exact inverse of `rgb_to_yuv`, so an RGB → YUV → RGB round trip is now limited only by the
   input dtype instead of losing up to `1.36e-3` (in B, at `rgb = (1, 1, 0)`) at every precision, `float64` included
   (#4044). The inverse kernel was a separately rounded copy of the published BT.470-5 M/PAL inverse relations rather
--- a/kornia/geometry/transform/imgwarp.py
+++ b/kornia/geometry/transform/imgwarp.py
@@ -120,7 +120,7 @@ def _empty_warp_output_2d(
     # rejecting it. Matching that would mean bilinear-sampling into an integer output, so the
     # cpu/cuda contract is the one enforced here.)
     if not src.is_floating_point():
-        raise RuntimeError(f"Expected a floating point src, got {src.dtype}.")
+        raise NotImplementedError(f"Expected a floating point src, got {src.dtype}.")
     # ``grid_sample`` requires the sampling grid in ``src.dtype``. ``grid_dtype`` is the dtype the
     # caller's non-empty pipeline would actually produce from ``transform``, so checking it here
     # makes a zero-sized ``dsize`` neither stricter nor laxer than a non-empty one.
@@ -162,7 +162,7 @@ def _empty_warp_output_3d(
     # As in the 2-D helper: an integral ``src`` cannot reach cpu/cuda ``grid_sample`` on either
     # path, so reject it by name instead of via the transform-dtype message below.
     if not src.is_floating_point():
-        raise RuntimeError(f"Expected a floating point src, got {src.dtype}.")
+        raise NotImplementedError(f"Expected a floating point src, got {src.dtype}.")
     if src.dtype != transform.dtype:
         raise RuntimeError(f"Expected src and transform with the same dtype, got {src.dtype} and {transform.dtype}.")
     grid_zero = transform.reshape(-1)[:1].sum() * 0.0
--- a/tests/geometry/transform/test_imgwarp.py
+++ b/tests/geometry/transform/test_imgwarp.py
@@ -140,11 +140,14 @@ def test_empty_destination_blames_an_integral_src_by_name(op_name, device):
     src = torch.zeros(1, 3, 3, 4, device=device, dtype=torch.int64)
     matrix = torch.eye(3, device=device)[:rows].unsqueeze(0)
 
-    with pytest.raises(NotImplementedError, match="floating point src"):
-        op(src, matrix, (0, 4))
-    # The non-empty path rejects it with the same type, so the two paths stay in parity.
-    with pytest.raises(NotImplementedError):
+    # Both paths must reject it with the same exception type. ``grid_sample`` raises
+    # ``NotImplementedError`` on newer torch and ``RuntimeError`` on older ones, so pin the
+    # parity rather than a fixed class: catch whatever the full path raises, then require the
+    # empty path to raise something the same ``except`` clause would catch.
+    with pytest.raises(RuntimeError) as full:
         op(src, matrix, (1, 4))
+    with pytest.raises(type(full.value), match="floating point src"):
+        op(src, matrix, (0, 4))
 
 
 @pytest.mark.parametrize("op_name", ["warp_affine", "warp_perspective"])
--- a/tests/geometry/transform/test_imgwarp.py
+++ b/tests/geometry/transform/test_imgwarp.py
@@ -140,10 +140,10 @@ def test_empty_destination_blames_an_integral_src_by_name(op_name, device):
     src = torch.zeros(1, 3, 3, 4, device=device, dtype=torch.int64)
     matrix = torch.eye(3, device=device)[:rows].unsqueeze(0)
 
-    with pytest.raises(RuntimeError, match="floating point src"):
+    with pytest.raises(NotImplementedError, match="floating point src"):
         op(src, matrix, (0, 4))
-    # The non-empty path rejects it too, so the empty guard is not stricter.
-    with pytest.raises((RuntimeError, NotImplementedError)):
+    # The non-empty path rejects it with the same type, so the two paths stay in parity.
+    with pytest.raises(NotImplementedError):
         op(src, matrix, (1, 4))
 
 
--- a/tests/geometry/transform/test_imgwarp3d.py
+++ b/tests/geometry/transform/test_imgwarp3d.py
@@ -77,6 +77,27 @@ def test_negative_destination_raises(op_name, device, dtype):
 
 
 @pytest.mark.parametrize("op_name", ["warp_affine3d", "warp_perspective3d"])
+def test_empty_destination_blames_an_integral_src_by_name(op_name, device):
+    """``grid_sample`` rejects an integral volume on both paths; the message must name ``src``.
+
+    Scoped away from MPS, whose ``grid_sample`` accepts an integral image and samples it back into
+    ``int64`` instead of rejecting it, so the non-empty half of the pairing does not hold there.
+    """
+    if device.type == "mps":
+        pytest.skip("MPS grid_sample accepts an integral image instead of rejecting it")
+    op = getattr(proj, op_name)
+    matrix_size = (3, 4) if op_name == "warp_affine3d" else (4, 4)
+    src = torch.zeros(1, 2, 3, 4, 5, device=device, dtype=torch.int64)
+    matrix = torch.eye(*matrix_size, device=device).unsqueeze(0)
+
+    # The non-empty path rejects it too, so the empty guard is not stricter.
+    with pytest.raises(RuntimeError) as full:
+        op(src, matrix, (1, 4, 5))
+    with pytest.raises(type(full.value), match="floating point src"):
+        op(src, matrix, (0, 4, 5))
+
+        
+@pytest.mark.parametrize("op_name", ["warp_affine3d", "warp_perspective3d"])
 def test_empty_destination_keeps_grid_sample_validation(op_name, device, dtype):
     src = torch.rand(2, 2, 3, 4, 5, device=device, dtype=dtype)
     matrix_size = (3, 4) if op_name == "warp_affine3d" else (4, 4)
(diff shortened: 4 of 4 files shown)


Himesh-rupchandani:
CI caught something worth raising before this goes further.

The exception `grid_sample` raises for an integral `src` depends on the torch version:

- torch 2.9.1 → `NotImplementedError`
- torch 2.5.1 → `RuntimeError`

All 13 failures on the previous push were on the 2.5.1 matrix entries; every 2.9.1 entry passed. I confirmed it locally with a bare `F.grid_sample` call on 2.5.1.

So hardcoding `NotImplementedError` in the guard achieves parity on newer torch but breaks it on 2.5.1, where the full path raises `RuntimeError` and the empty path would now raise `NotImplementedError`. `NotImplementedError` subclasses `RuntimeError`, so `except RuntimeError` still catches both — but the exact-type parity this issue is about does not hold there.

I've changed the test to pin the *relationship* rather than a fixed class: it captures whatever the full path raises, then requires the empty path to raise the same type. That passes on both matrices.

That leaves a design question I'd rather not guess at:

1. Keep the hardcoded `NotImplementedError` and accept that exact-type parity only holds where `grid_sample` raises it, or
2. Have the guard mirror whatever `grid_sample` would raise on the running torch, so parity holds across the whole supported matrix.

(1) is the simpler diff and matches the issue as written. (2) is more faithful to "the empty path should reject exactly what the full path rejects", but needs a version check or a probe in the guard.

Which would you prefer?

ducha-aiki:
## Review — APPROVE, with one change requested before merge

Good, well-scoped fix, and thank you for checking the 3-D twin unprompted — that was the right instinct and it did have the identical defect.

### Verified

- The before/after table reproduces exactly on this branch.
- The new test discriminates: reverting `imgwarp.py` alone fails both parametrizations with `RuntimeError: Expected a floating point src`.
- Parity genuinely holds in 3-D: `warp_affine3d` and `warp_perspective3d` both raise `NotImplementedError: "grid_sampler3d_cpu" not implemented for 'Long'` on the full path and now `NotImplementedError: Expected a floating point src` on the empty one.
- `tests/geometry/transform` cpu → 515 passed, 15 skipped, 1 xfailed. Your unrelated `TestDecomposeEssentialMatrixNoSVD` failure is a `tests/geometry/epipolar` case and does reproduce on clean `main` — correctly attributed.
- `NotImplementedError` subclasses `RuntimeError`, so the CHANGELOG's compatibility claim is accurate.

### Requested change

1. **The 3-D half of your change has no test.** `test_empty_destination_blames_an_integral_src_by_name` lives in `tests/geometry/transform/test_imgwarp.py`, parametrized over `["warp_affine", "warp_perspective"]` only. `test_imgwarp3d.py` has empty-`dsize` tests for autograd, negative `dsize` and grid-sample validation, but nothing pinning the integral-`src` exception type. Please add the 3-D counterpart there over `["warp_affine3d", "warp_perspective3d"]` — otherwise the guard you edited in `_empty_warp_output_3d` is unpinned and can regress silently, which is the exact failure mode that produced #4031.

### Findings

2. **Two more public functions change than the PR body and CHANGELOG name.** `_empty_warp_output_2d` is also reached by `remap` (`imgwarp.py:838`), and `_empty_warp_output_3d` by `warp_perspective3d` (`imgwarp.py:1645`). Confirmed on your branch: `remap` with an `int64` image and a zero-height map now raises `NotImplementedError` where `main` raised `RuntimeError`, and the full path was already `NotImplementedError`, so it is fixed rather than broken. The CHANGELOG entry should list `warp_perspective3d` and `remap` alongside the other three.

3. **The new test's parity pin is one-directional.** Because `NotImplementedError` is a subclass of `RuntimeError`, if some supported torch raises a plain `RuntimeError` from `grid_sample`, then `type(full.value)` is `RuntimeError` and `pytest.raises(RuntimeError)` happily accepts the empty path's `NotImplementedError` — the very mismatch this test exists to catch, only reversed. Making it exact costs one line:

   ```python
   with pytest.raises(RuntimeError) as full:
       op(src, matrix, (1, 4))
   with pytest.raises(RuntimeError, match="floating point src") as empty:
       op(src, matrix, (0, 4))
   assert type(empty.value) is type(full.value)
   ```

   That holds on any torch version and does not depend on which class the full path picks. Note the consequence, which is a feature rather than a problem: if an older supported torch does raise plain `RuntimeError` there, this version will fail and tell us — worth knowing, since kornia declares `torch>=2.0` and I could only verify the current 2.9.1 behaviour here.

Everything else — matching the message verbatim, keeping the `.. note::` comment accurate, the MPS carve-out staying as it was — looks right. Merges cleanly with the five other open PRs.

Himesh-rupchandani:
Thanks — all three points make sense. I've implemented 1 and 2, and testing 3 turned up something that changes the picture.

## 1. 3-D test — added

`test_empty_destination_blames_an_integral_src_by_name` added to `test_imgwarp3d.py`, parametrized over `["warp_affine3d", "warp_perspective3d"]`, mirroring the 2-D one including the MPS skip.

## 2. CHANGELOG — corrected

Now names all five: `warp_affine`, `warp_perspective`, `remap`, `warp_affine3d`, `warp_perspective3d`. Confirmed `remap` reaches `_empty_warp_output_2d` at `imgwarp.py:838` and `warp_perspective3d` reaches the 3-D helper at `:1645`.

## 3. Exact-type parity — your prediction reproduces on torch 2.5.1

You wrote that if an older supported torch raises plain `RuntimeError` there, the assertion would fail and tell us. It does. I installed 2.5.1 locally:

```
torch 2.5.1: F.grid_sample(int64) -> RuntimeError
torch 2.9.1: F.grid_sample(int64) -> NotImplementedError
```

With the exact assertion in place, on 2.5.1:

```
4 failed, 513 passed
assert <class 'NotImplementedError'> is <class 'RuntimeError'>
```

So the hardcoded `NotImplementedError` does not achieve exact-type parity on 2.5.1 — the guard raises `NotImplementedError` while the full path raises `RuntimeError`. `pyproject.toml` declares `torch>=2.0.0` and 2.5.1 is in the CI matrix, so this is a supported configuration rather than a hypothetical.

## Option 2 breaks TorchScript

Before bringing this back to you I prototyped the "mirror whatever `grid_sample` raises" approach — a module-level helper that probes once and returns the class:

```python
def _integral_src_error_type() -> type[Exception]:
    try:
        F.grid_sample(torch.zeros(1, 1, 1, 1, dtype=torch.int64),
                      torch.zeros(1, 1, 1, 2), align_corners=True)
    except Exception as err:
        return type(err)
    return RuntimeError
```

It fixes parity on both versions, but scripting stops working:

```
FAILED TestWarpAffine::test_jit_script[cpu-float32-fill-True]
  RuntimeError: Can't redefine method: forward on class: __torch__.test_imgwarp...
```

I reverted it and those four jit tests pass again, so the failure is the probe, not something pre-existing. A dynamic exception class is not scriptable.

## Where that leaves it

The current branch has 1 and 2 applied and the test still pinning the relationship one-directionally, which passes on both 2.5.1 and 2.9.1. I have the exact-type version ready but have not pushed it, since it turns 2.5.1 red.

Three ways forward as I see them:

1. Keep the one-directional pin. Passes everywhere; accepts the weakness you identified.
2. Use the exact assertion and drop 2.5.1 from the matrix, or xfail it there.
3. Make the guard version-aware without a runtime probe — e.g. a module-level constant computed once at import from `torch.__version__` rather than a function call, if that scripts cleanly. I have not tested this yet.

Happy to do whichever you prefer. (3) seems closest to your intent but I would want to verify it scripts before proposing it properly.

ducha-aiki:
**Delta re-review of `1774eeee` — verdict: APPROVE. Take option 1, keep exactly what you have pushed, and change nothing further.**

And first: **my point 3 was wrong, and I am retracting it.** You did the right thing by prototyping it and bringing back the measurement instead of implementing it. The one-directional pin is not a weakness — it is precisely the correct property, and the exact-type version I asked for would have been a worse test. Details below, because the reasoning is the answer to your design question.

Reviewed at head `1774eeee`, merge-base `6fa75d64`, CPU, torch 2.9.1 (the repo `.venv`).

## Why one-directional is right

The property that matters to a caller is: **any `except` clause written to catch the full path must also catch the empty path.** Formally that is `issubclass(type(empty), type(full))` — one-directional by construction. Exact type equality is a strictly stronger claim, and nothing depends on it.

Enumerated over every scenario in play:

```
scenario      full path             empty path            handler-safe?  exact type?
torch 2.9.1+  NotImplementedError   NotImplementedError   True           True
torch 2.5.1   RuntimeError          NotImplementedError   True           False
#4031 on 2.9  NotImplementedError   RuntimeError          False          False
#4031 on 2.5  RuntimeError          RuntimeError          True           True
```

Read the `handler-safe?` column: it is `False` on exactly one row — the actual #4031 defect — and `True` everywhere else. The `exact type?` column is `False` on row 2, where **there is no defect at all**: a caller writing `except RuntimeError` for the full path catches the empty path's `NotImplementedError` without noticing.

Row 4 is the other half of the point: on torch 2.5.1 both paths raise `RuntimeError` already, so there was never a bug there to fix. A test that goes red on 2.5.1 would be reporting a defect that does not exist on that version. That is what option 2 would have done, and it is why dropping or `xfail`ing 2.5.1 would have been papering over a correct result.

So `pytest.raises(type(full.value), match="floating point src")` is not a weaker fallback — it asserts `issubclass(type(empty), type(full))` exactly, which is the whole contract, on every torch version, with no version sniffing. **Do not change it.**

## And it discriminates

I re-created #4031 by reverting both guards on your branch:

```console
$ sed -i 's/raise NotImplementedError(f"Expected a floating point src/raise RuntimeError(f"Expected a floating point src/g' \
      kornia/geometry/transform/imgwarp.py     # 2 guards reverted
$ pytest ...test_imgwarp.py ...test_imgwarp3d.py -k blames_an_integral_src -q --device=cpu
FAILED test_imgwarp.py::test_empty_destination_blames_an_integral_src_by_name[cpu-warp_affine]
FAILED test_imgwarp.py::test_empty_destination_blames_an_integral_src_by_name[cpu-warp_perspective]
FAILED test_imgwarp3d.py::test_empty_destination_blames_an_integral_src_by_name[cpu-warp_affine3d]
FAILED test_imgwarp3d.py::test_empty_destination_blames_an_integral_src_by_name[cpu-warp_perspective3d]
4 failed
```

All four, 2-D and 3-D. Restored, all four pass.

## Options 2 and 3 — both correctly rejected

**Option 2** (exact assertion, drop or `xfail` 2.5.1) asserts a property callers do not need and would mark a correct configuration as broken. No.

**Option 3** (version-aware constant from `torch.__version__`) I would reject even if it scripts. It hardcodes the release where torch changed an undocumented exception class, so it is wrong by construction for any version you did not test — including future ones, and including backports. Your runtime-probe prototype was the more honest form of the same idea, and TorchScript already vetoed it. Good instinct to revert it rather than work around the jit failure.

The hardcoded `NotImplementedError` is not a compromise, it is the correct constant: it is the narrowest class that is a subclass of everything `grid_sample` raises across the supported matrix, which is exactly what makes `handler-safe?` true on every row.

## Points 1 and 2 verified

**3-D test** — present and parametrized as asked; the 4 node IDs above are the proof.

**CHANGELOG** — now names all five (`warp_affine`, `warp_perspective`, `remap`, `warp_affine3d`, `warp_perspective3d`), and the `NotImplementedError` derives from `RuntimeError` compatibility note is accurate. The two functions I flagged do reach the helpers and are fixed rather than broken:

```
remap        empty -> NotImplementedError: Expected a floating point src, got torch.int64
remap        full  -> NotImplementedError: "grid_sampler_2d_cpu_kernel_impl" not implemented...
warp_persp3d empty -> NotImplementedError: Expected a floating point src, got torch.int64
warp_persp3d full  -> NotImplementedError: "grid_sampler3d_cpu" not implemented for 'Long'
```

## Verification

```console
$ pytest tests/geometry/transform -q --device=cpu --dtype=float32
517 passed, 15 skipped, 1 xfailed
```

One optional nicety, take it or leave it: the test comment says "pin the parity rather than a fixed class", which is right but undersells it. Something like *"require the empty path's exception to be catchable by any `except` written for the full path — one-directional on purpose, since `NotImplementedError ⊂ RuntimeError` and older torch raises the base class"* would stop a future reader "fixing" it into an exact-type assertion. Purely editorial.

Nothing blocking. This is ready to merge, and the way you handled point 3 — prototype, measure, come back with the counter-evidence rather than implementing a maintainer's wrong suggestion — is exactly right.

ducha-aiki (approved):

PR #4077, opened 2026-08-29
docs(feature): fix ggiven typo in get_sift_bin_ksize_stride_pad

Sid-5137:
Fixes a `ggiven` typo in `get_sift_bin_ksize_stride_pad`'s docstring.

Originally this PR was larger, but it overlapped #4072 (now merged) and the `matching.py` hunk was dropped per review. The `lafs1` question is being handled in #4075.

Part of #4067.

### DIFF
--- a/kornia/feature/siftdesc.py
+++ b/kornia/feature/siftdesc.py
@@ -71,7 +71,7 @@ def get_sift_bin_ksize_stride_pad(patch_size: int, num_spatial_bins: int) -> Tup
 
     Args:
         patch_size: the given patch size.
-        num_spatial_bins: the ggiven number of spatial bins.
+        num_spatial_bins: the given number of spatial bins.
 
     Returns:
         ksize, stride, pad.


Sid-5137:
Narrowed this after noticing it overlapped #4072, which I had missed when I opened it. Sorry for the noise.

I have dropped everything #4072 already covers: the `laf.py` docstrings and the `siftdesc.py` shape, `Returns:` and error message fixes. What is left is the `match_fginn` `lafs1` documentation, which #4067's title mentions but #4072 does not touch, plus a one word typo. Two files, +3/-2, so there is no longer any overlap with #4072.

ducha-aiki:
## Review — changes requested

Thank you for narrowing this once you spotted the overlap with #4072 — that was the right call and it made this reviewable. Two different verdicts on the two halves.

### `siftdesc.py` — keep it

`ggiven` → `given`. Correct, genuinely not covered by #4072 (I diffed the two branches), and the branches merge cleanly. Keep it.

### `matching.py` — please drop it

In kornia we **fix** things like this, we do not document them. A docstring that records a dead required argument makes the wart permanent and hands the next reader a reason not to touch it. And `match_fginn` already has an open fix PR — #4075 — so this function is being corrected right now; the argument question belongs there, in code.

The reasoning the change rests on also does not survive a look at the caller. The PR says the parameter must stay so `match_fginn` and `match_adalam` remain interchangeable:

```python
out = match_adalam(desc1, desc2, lafs1, lafs2, config=_params)
```

But `GeometryAwareDescriptorMatcher.forward` (`kornia/feature/matching.py:579-586`) dispatches in two separate branches, each with its own literal argument list:

```python
if self.match_mode == "fginn":
    out = match_fginn(desc1, desc2, lafs1, lafs2, params["th"], params["spatial_th"], params["mutual"])
elif self.match_mode == "adalam":
    out = match_adalam(desc1, desc2, lafs1, lafs2, config=_params)
```

Nothing calls the two through a shared signature, so symmetry is not actually constraining anything.

### One thing to carry into the fix, because it is a trap

A straight removal of `lafs1` is not merely a breaking API change — it is a **silent** one. An existing positional call `match_fginn(d1, d2, lafs1, lafs2, th, spatial_th)` would afterwards bind `lafs1` into the `lafs2` slot. When `B1 != B2` that raises `IndexError`, but when `B1 == B2` — the ordinary same-detector, same-top-k case — it returns different matches with no error at all. Measured on a 64/64 case: 40 vs 44 matches at `spatial_th=30`, and 44 vs 52 at `spatial_th=50`.

So the fix needs either a rename that forces a `TypeError` on old call sites, or a deprecation cycle where the argument keeps its position and warns. I have raised the same point on #4075 so the decision happens in one place.

### Suggested shape for this PR

Keep the `ggiven` typo fix, drop the `matching.py` hunk, and let the `lafs1` resolution land in #4075. If you would rather own the `lafs1` fix yourself, that works too — say so and it can be sequenced after #4075 merges, but it should be a signature change, not a docstring.

Sid-5137:
Dropped the `matching.py` hunk, so this is now just the `ggiven` typo.

You are right about the dispatch, and my reasoning was wrong. I read the `match_adalam` call and assumed a shared signature without checking the `fginn` branch directly above it. They are two separate branches with their own literal argument lists, so symmetry was not constraining anything. Sorry for putting that in the description as if it were established.

The silent-shift point is a good catch and I had not considered it. Removing a positional parameter changing results rather than raising, whenever `B1 == B2`, is a much worse failure mode than a clean break.

I would like to take the `lafs1` fix if the offer still stands. Happy to wait for #4075 to merge and then open it as a signature change with a deprecation cycle, keeping `lafs1` in position and warning on use, rather than a straight removal. If you would rather it be a rename that forces a `TypeError` on old call sites, I am fine with that too, just say which you prefer.

ducha-aiki (approved):

PR #4078, opened 2026-08-29
fix(feature): treat the image extent as w-1/h-1 in laf_is_inside_image

### LINKED ISSUE #4064: laf_is_inside_image uses w/h as the image extent while the rest of laf.py uses w-1/h-1
### Describe the bug

`laf_is_inside_image` treats the image extent as `w` × `h`, while every other function in `kornia/feature/laf.py` treats it as `(w - 1)` × `(h - 1)`. LAFs that reach one pixel past the last valid coordinate are reported as fully inside.

https://github.com/kornia/kornia/blob/main/kornia/feature/laf.py#L527

```python
good_lafs_mask = (
    (pts[..., 0] >= border) * (pts[..., 0] <= w - border) * (pts[..., 1] >= border) * (pts[..., 1] <= h - border)
)
```

Every other extent computation in the same file uses `w - 1` / `h - 1`:

| line | function | extent |
|---|---|---|
| 339–340 | `denormalize_laf` | `w - 1`, `h - 1` |
| 370–371 | `normalize_laf` | `w - 1`, `h - 1` |
| 400–401 | `generate_patch_grid_from_normalized_LAF` | `w - 1`, `h - 1` |
| **527** | **`laf_is_inside_image`** | **`w`, `h`** |

This matches the documented convention in `get_laf_center`: *"the center of a 5-pixel image (coordinates from 0 to 4) is 2, and not 2.5"* — coordinates run `0 .. w-1`.

### Reproduction

```python
import torch
from kornia.feature.laf import laf_is_inside_image, laf_from_center_scale_ori

img = torch.zeros(1, 1, 32, 32)          # valid x coordinates are 0 .. 31
for cx in [29.0, 29.5, 30.0, 33.0]:
    xy = torch.tensor([[[cx, 16.0]]], dtype=torch.float64)
    laf = laf_from_center_scale_ori(xy, torch.full((1, 1, 1, 1), 2.0, dtype=torch.float64))
    print(f"centre x={cx:5.1f}, radius 2 -> rightmost boundary x={cx + 2:5.1f}  "
          f"inside={bool(laf_is_inside_image(laf, img)[0, 0])}")
```

```
centre x= 29.0, radius 2 -> rightmost boundary x= 31.0  inside=True
centre x= 29.5, radius 2 -> rightmost boundary x= 31.5  inside=True     <-- past 31
centre x= 30.0, radius 2 -> rightmost boundary x= 32.0  inside=True     <-- past 31
centre x= 33.0, radius 2 -> rightmost boundary x= 35.0  inside=False
```

### Expected behavior

The upper bound should be `w - 1 - border` / `h - 1 - border`, making the check symmetric with the `>= border` lower bound and consistent with the rest of `laf.py`.

### Additional context

The existing test does not discriminate the boundary — its "inside" LAF is a radius-1 circle at `(5, 2)` in a 10×5 image, well clear of the edge:

https://github.com/kornia/kornia/blob/main/tests/feature/test_laf.py#L521-L526

Fixing this drops LAFs within one pixel of the right/bottom border that are currently kept, so it changes detector output near image edges and wants a deliberate decision.

ducha-aiki:
## Problem

`laf_is_inside_image` bounded the boundary points by `w - border` / `h - border`, while every other extent computation in `laf.py` uses `w - 1` / `h - 1`:

| line | function | extent |
|---|---|---|
| 339–340 | `denormalize_laf` | `w - 1`, `h - 1` |
| 370–371 | `normalize_laf` | `w - 1`, `h - 1` |
| 400–401 | `generate_patch_grid_from_normalized_LAF` | `w - 1`, `h - 1` |
| **527** | **`laf_is_inside_image`** | **`w`, `h`** |

Valid pixel coordinates run `0 .. w-1`, as documented on `get_laf_center`: *"the convention is that center of 5-pixel image (coordinates from 0 to 4) is 2, and not 2.5"*.

Because the lower bound (`>= border`) was already correct, the accepted window was **asymmetric** — strict at the low edge, one pixel loose at the high edge:

```
radius-2 LAF, 32-wide image (last valid x = 31)
                                       before   after
  centre x=29.0 -> rightmost 31.0      inside   inside
  centre x=29.5 -> rightmost 31.5      inside   OUTSIDE
  centre x=30.0 -> rightmost 32.0      inside   OUTSIDE
```

## The inline copy in `ScaleSpaceDetector`

`_process_octave` carries a hand-rolled version whose own comment calls it an *"Inline equivalent of `laf_is_inside_image(scale_laf(current_lafs, 0.5), octave[:, 0], 5)`"*. It had the same `w - 5` / `h - 5` bound and is updated in step — leaving it would have silently broken the equivalence that comment asserts.

### ... and its `max |sin|` constant was computed with the wrong spacing

The residual inline-vs-reference disagreement was **not** inherent to the approximation, as an earlier revision of this description claimed — the constant is simply wrong:

```python
_MAX_ABS_SIN_12: float = math.sin(3 * 2 * math.pi / 11)   # 0.98982
```

`laf_to_boundary_points(LAF, n_pts=12)` samples `torch.linspace(0, 2*pi, n_pts - 1)` = `linspace(0, 2π, 11)`, whose spacing is **2π/10**, not 2π/11:

```
max|sin| over linspace(0, 2pi, 11)                              : 0.95105654
_MAX_ABS_SIN_12 before                                          : 0.98982144  = sin(3*2pi/11)
sin(2*2pi/10)                                                   : 0.95105652
measured (max_x - cx)/half_s from laf_to_boundary_points(laf,12) : 0.95105648
```

So the inline check inflated the x-extent by 4%, making it *stricter* than the reference — the direction of every disagreement observed. With the corrected constant the block genuinely is the inline equivalent its comment claims:

```
4000 random isotropic LAFs spanning the borders of a 64x96 image, border=5:
  before  sin(3*2pi/11)=0.98982 : 7 disagreements (inline stricter: 7, looser: 0)
  after   sin(2*2pi/10)=0.95106 : 0 disagreements
```

`tests/feature/test_scale_space_detector.py::test_inline_boundary_constant_matches_reference` pins the constant against the extent `laf_to_boundary_points` actually produces, and fails on the old value (`assert 0.03876 < 1e-06`). Constant introduced in #3623.

## Practical impact

Small but **not zero**, and the earlier version of this section got that wrong. On a random 2×1×96×128 input `ScaleSpaceDetector` output is `torch.equal` before and after — under both changes together and under the constant change alone — because `border=5` keeps keypoints away from the edge and a random image rarely puts a detection in the affected band. That is a property of the input, not of the change.

The constant fix is a **loosening in x**: a detection whose x-extent falls between `sin(3*2pi/11)` and `sin(2*2pi/10)` used to be discarded and is now kept. Holding `w-1`/`h-1` fixed and toggling only `_MAX_ABS_SIN_12` on a 128x160 image with a sigma=4 blob at (136, 64) and a weak one at (40, 40), with `ScaleSpaceDetector(num_features=50)`:

```
new constant (0.95106):  laf[0] = [[25.933, 0, 135.184], [0, 25.933, 63.532]]   resp 0.04437
old constant (0.98982):  laf[0] = [[24.190, 0,  40.000], [0, 24.190, 40.000]]   resp 0.00396
```

The strongest detection in the image -- an order of magnitude above what replaced it -- was being discarded by the wrong constant. Sweeping the blob over sigma in {4, 6, 8, 12} and cx in [W-40, W-2) in 0.25 px steps, **6 of 608 positions change output**. It needs the detection to land in a band of width `0.039 * half_s` against the x bound, so it is rare on noise images (0 of 12 uniform-random inputs differ) -- but the case where it fires is exactly the case a user notices. The CHANGELOG entry says this rather than claiming the output is unchanged.

Detectors that call `laf_is_inside_image` with a small `border` will now drop LAFs reaching within one pixel of the right/bottom edge that were previously kept.

## Tests

The existing `test_touch` never discriminated this — its "inside" LAF is a radius-1 circle at `(5, 2)` in a 10×5 image, nowhere near the boundary. Three new tests do, and **all three fail without the change**:

```
FAILED ...::test_boundary_is_the_last_valid_pixel[cpu-float32] - assert not True
FAILED ...::test_boundary_is_symmetric[cpu-float32] - AssertionError: asymmetric at offset -1.0
FAILED ...::test_border_argument_shrinks_both_edges[cpu-float32] - assert not True
```

`test_inline_boundary_constant_matches_reference` is the fourth, pinning the corrected spacing.

`test_boundary_is_symmetric` probes offsets *past* the edge — an earlier revision only probed points that are inside under both behaviours and so passed on `main`, which would have made it worthless.

## Verification

```console
647 passed, 126 skipped

$ pixi run pre-commit-all
all hooks Passed
```

Closes #4064

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   existing FGINN tests pass unchanged on both sides, which is how this survived; two tests that discriminate it were
   added. `match_fginn`'s docstring now also records the saturation corner, where every candidate falls within
   `spatial_th` of the 1st nearest neighbor, the ratio collapses towards zero and the match is accepted.
+* Fix `laf_is_inside_image` treating the image extent as `(w, h)` rather than `(w - 1, h - 1)`, which made its
+  bounds asymmetric: the lower bound rejected anything left of `x = 0`, but the upper bound accepted `x = w`, a
+  full pixel past the last valid column `w - 1` (#4064). Valid pixel coordinates run `0 .. w-1` and `0 .. h-1` --
+  the convention `get_laf_center` documents and the one `normalize_laf`/`denormalize_laf` already use -- so the
+  upper bound is now `w - 1 - border` and `h - 1 - border`. A LAF whose boundary points reach past the last valid
+  pixel coordinate is now reported as outside; anything strictly inside is unchanged. The equivalent inlined check
+  in `ScaleSpaceDetector._process_octave` moved with it, so detections within one pixel of the right or bottom edge
+  that used to survive its `border=5` filter are now discarded.
 
 
 ## :rocket: [0.6.11] - 2022-03-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,7 +140,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   that used to survive its `border=5` filter are now discarded. That inlined check also computed its `max |sin|`
   constant with the wrong angular spacing (`2*pi/11` instead of the `2*pi/10` that `laf_to_boundary_points(n_pts=12)`
   actually samples), inflating the tested x-extent by 4% and making the inline check stricter than the reference it
-  claims to reproduce; the two now agree exactly. `ScaleSpaceDetector` output is unchanged.
+  claims to reproduce; the two now agree exactly. That correction is a loosening in x, so it can change
+  `ScaleSpaceDetector` output on its own: a detection whose x-extent falls between the two constants used to be
+  discarded and is now kept. It is rare -- the detection has to land in a band of width `0.039 * half_s` against the
+  x bound -- but when it fires it can promote the strongest response in the image, so it is not output-neutral.
 
 
 ## :rocket: [0.6.11] - 2022-03-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,7 +137,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   upper bound is now `w - 1 - border` and `h - 1 - border`. A LAF whose boundary points reach past the last valid
   pixel coordinate is now reported as outside; anything strictly inside is unchanged. The equivalent inlined check
   in `ScaleSpaceDetector._process_octave` moved with it, so detections within one pixel of the right or bottom edge
-  that used to survive its `border=5` filter are now discarded.
+  that used to survive its `border=5` filter are now discarded. That inlined check also computed its `max |sin|`
+  constant with the wrong angular spacing (`2*pi/11` instead of the `2*pi/10` that `laf_to_boundary_points(n_pts=12)`
+  actually samples), inflating the tested x-extent by 4% and making the inline check stricter than the reference it
+  claims to reproduce; the two now agree exactly. `ScaleSpaceDetector` output is unchanged.
 
 
 ## :rocket: [0.6.11] - 2022-03-28
--- a/kornia/feature/laf.py
+++ b/kornia/feature/laf.py
@@ -549,9 +549,11 @@ def laf_is_inside_image(laf: torch.Tensor, images: torch.Tensor, border: int = 0
     KORNIA_CHECK_LAF(laf)
     _, _, h, w = images.size()
     pts = laf_to_boundary_points(laf, 12)
-    good_lafs_mask = (
-        (pts[..., 0] >= border) * (pts[..., 0] <= w - border) * (pts[..., 1] >= border) * (pts[..., 1] <= h - border)
-    )
+    # Valid pixel coordinates run 0 .. w-1 and 0 .. h-1, matching the convention documented on
+    # `get_laf_center` and the `w - 1` / `h - 1` extent used by `normalize_laf` / `denormalize_laf`.
+    x_max = float(w - 1) - border
+    y_max = float(h - 1) - border
+    good_lafs_mask = (pts[..., 0] >= border) * (pts[..., 0] <= x_max) * (pts[..., 1] >= border) * (pts[..., 1] <= y_max)
     good_lafs_mask = good_lafs_mask.min(dim=2)[0]
     return good_lafs_mask
 
--- a/kornia/feature/scale_space_detector.py
+++ b/kornia/feature/scale_space_detector.py
@@ -299,11 +299,14 @@ class ScaleSpaceDetector(nn.Module):
         cx = current_lafs[:, :, 0, 2]
         cy = current_lafs[:, :, 1, 2]
         h, w = octave.shape[3], octave.shape[4]
+        # Valid pixel coordinates run 0 .. w-1 / 0 .. h-1, so the upper bound is (size - 1) - border.
+        x_max = float(w - 1) - 5
+        y_max = float(h - 1) - 5
         good_mask = (
             (cx - half_s * _MAX_ABS_SIN_12 >= 5)
-            & (cx + half_s * _MAX_ABS_SIN_12 <= w - 5)
+            & (cx + half_s * _MAX_ABS_SIN_12 <= x_max)
             & (cy - half_s >= 5)
-            & (cy + half_s <= h - 5)
+            & (cy + half_s <= y_max)
         )
         resp_flat_best = resp_flat_best * good_mask.to(dev, dtype)
         current_lafs.mul_(px_size)
--- a/kornia/feature/scale_space_detector.py
+++ b/kornia/feature/scale_space_detector.py
@@ -38,11 +38,11 @@ from .orientation import PassLAF
 from .responses import BlobHessian
 
 # Max |sin| among the 11 boundary points sampled by laf_to_boundary_points(n_pts=12):
-#   angles = linspace(0, 2π, 11) → k * 2π/11 for k=0..10
-#   max|sin| at k=3: sin(6π/11) ≈ 0.9898;  max|cos| at k=0: cos(0) = 1.0
+#   angles = linspace(0, 2π, n_pts - 1) = linspace(0, 2π, 11) → k * 2π/10 for k=0..10
+#   max|sin| at k=2 and k=3: sin(2π/5) ≈ 0.9511;  max|cos| at k=0: cos(0) = 1.0
 # Used to inline the boundary check in _process_octave for isotropic LAFs (rotmat=eye(2)),
 # avoiding CPU→GPU allocation + bmm every octave.
-_MAX_ABS_SIN_12: float = math.sin(3 * 2 * math.pi / 11)  # ≈ 0.9898
+_MAX_ABS_SIN_12: float = math.sin(2 * 2 * math.pi / 10)  # ≈ 0.9511
 
 
 def _scale_index_to_scale(max_coords: torch.Tensor, sigmas: torch.Tensor, num_levels: int) -> torch.Tensor:
--- a/tests/feature/test_laf.py
+++ b/tests/feature/test_laf.py
@@ -594,6 +594,60 @@ class TestLAFIsTouchingBoundary(BaseTester):
         expected = torch.tensor([[False, True]], device=device)
         assert torch.all(kornia.feature.laf_is_inside_image(laf, img) == expected).item()
 
+    @staticmethod
+    def _radius_laf(cx, cy, radius, device, dtype):
+        return kornia.feature.laf_from_center_scale_ori(
+            torch.tensor([[[cx, cy]]], device=device, dtype=dtype),
+            torch.full((1, 1, 1, 1), radius, device=device, dtype=dtype),
+        )
+
+    def test_boundary_is_the_last_valid_pixel(self, device, dtype):
+        """Valid coordinates run 0 .. size-1, matching `get_laf_center`'s documented convention.
+
+        A LAF reaching exactly `w - 1` is inside; anything past it is not (see #4064).
+        """
+        if dtype not in (torch.float32, torch.float64):
+            pytest.skip("boundary comparison needs full precision")
+        w = h = 32  # valid coordinates 0 .. 31
+        img = torch.zeros(1, 1, h, w, device=device, dtype=dtype)
+        r = 2.0
+
+        # rightmost boundary lands exactly on 31.0 -> inside
+        assert bool(kornia.feature.laf_is_inside_image(self._radius_laf(29.0, 16.0, r, device, dtype), img)[0, 0])
+        # ... and half a pixel past it -> outside
+        assert not bool(kornia.feature.laf_is_inside_image(self._radius_laf(29.5, 16.0, r, device, dtype), img)[0, 0])
+        # same for the bottom edge
+        assert bool(kornia.feature.laf_is_inside_image(self._radius_laf(16.0, 29.0, r, device, dtype), img)[0, 0])
+        assert not bool(kornia.feature.laf_is_inside_image(self._radius_laf(16.0, 29.5, r, device, dtype), img)[0, 0])
+
+    def test_boundary_is_symmetric(self, device, dtype):
+        """The low and high edges must be equally strict."""
+        if dtype not in (torch.float32, torch.float64):
+            pytest.skip("boundary comparison needs full precision")
+        w = h = 32
+        img = torch.zeros(1, 1, h, w, device=device, dtype=dtype)
+        r = 2.0
+        # negative offsets push the LAF past the edge, which is where an asymmetric
+        # upper bound shows up: on the low side it is rejected, on the high side it was not.
+        for offset in (-1.0, -0.5, 0.0, 0.5, 1.0):
+            low = kornia.feature.laf_is_inside_image(self._radius_laf(r + offset, 16.0, r, device, dtype), img)
+            high = kornia.feature.laf_is_inside_image(
+                self._radius_laf(float(w - 1) - r - offset, 16.0, r, device, dtype), img
+            )
+            assert bool(low[0, 0]) == bool(high[0, 0]), f"asymmetric at offset {offset}"
+
+    def test_border_argument_shrinks_both_edges(self, device, dtype):
+        if dtype not in (torch.float32, torch.float64):
+            pytest.skip("boundary comparison needs full precision")
+        w = h = 32
+        img = torch.zeros(1, 1, h, w, device=device, dtype=dtype)
+        r = 2.0
+        # with border=3 the usable range is 3 .. 28, so a radius-2 LAF centred at 26 just fits
+        assert bool(kornia.feature.laf_is_inside_image(self._radius_laf(26.0, 16.0, r, device, dtype), img, 3)[0, 0])
+        assert not bool(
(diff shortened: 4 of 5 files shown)


ducha-aiki (commented):
Reviewed current head `4ad2bd25` and found no actionable issues. The maximum valid LAF coordinates now consistently use `w - 1 - border` and `h - 1 - border` in both the public helper and optimized detector check. Tests cover exact-edge acceptance, half-pixel rejection, symmetry, and nonzero borders; the full CI matrix is green.

ducha-aiki:
No actionable findings in either review, so the only change is a CHANGELOG entry, added for consistency with the sibling PRs in this batch (#4073, #4074, #4075 were all asked for one). This is a user-visible behavior change: which LAFs count as inside moves, and `ScaleSpaceDetector` drops detections within one pixel of the right or bottom edge that used to survive its `border=5` filter.

`tests/feature/test_laf.py` + `tests/feature/test_scale_space_detector.py --device=cpu` → 105 passed, 6 skipped. pre-commit clean.

ducha-aiki:
## Independent review — verdict: **the fix is correct**, but I found the real cause of the residual inline/reference disagreement, and it is a one-token fix that belongs here.

Re-reviewed from scratch at a higher reasoning budget.

### The `w-1` / `h-1` change is right

Valid pixel coordinates run `0 .. w-1`; `denormalize_laf`, `normalize_laf` and `generate_patch_grid_from_normalized_LAF` all already use `w-1`/`h-1`, and `get_laf_center`'s docstring states the convention explicitly. The old bound `[border, w - border]` is symmetric about `w/2`, while the image's true centre in pixel-centre coordinates is `(w-1)/2` — hence the asymmetry. The new `[border, w-1-border]` is symmetric about the right point. All three new tests fail on `main` and pass on the branch; `tests/feature/test_laf.py` + `tests/feature/test_scale_space_detector.py` = 148 passed / 12 skipped; `-m jit` = 9 passed (`float(w-1)` scripts fine).

`ScaleSpaceDetector` output is byte-identical to `main` — I re-ran it independently on a 2×1×96×128 input and got `torch.equal == True` for both the LAFs and the responses.

### Finding: `_MAX_ABS_SIN_12` is computed with the wrong angular spacing

The PR body says the two remaining inline-vs-reference disagreements are "inherent to the approximation" — that "the reference's 12-point boundary sampling is slightly looser than the inline analytic extremum". That is the right observation with the wrong cause. The inline constant is simply wrong:

```python
_MAX_ABS_SIN_12: float = math.sin(3 * 2 * math.pi / 11)   # 0.98982
```

But `laf_to_boundary_points(LAF, n_pts=12)` samples `torch.linspace(0, 2*pi, n_pts - 1)` = `linspace(0, 2π, 11)`, whose spacing is **2π/10**, not 2π/11. The true maximum over the sampled angles:

```
max|sin| over linspace(0, 2pi, 11)          : 0.9510565400
_MAX_ABS_SIN_12 in the source               : 0.9898214419   = sin(3*2pi/11)
sin(2*2pi/10)                               : 0.9510565163
measured (max_x - cx)/half_s from laf_to_boundary_points(laf, 12): 0.9510566592
```

So the inline check inflates the x-extent by 4%, which makes it *stricter* than the reference — exactly the direction of every disagreement you observed. With the correct constant the inline becomes **exactly** equivalent to the function whose name is in the comment above it:

```
4000 random isotropic LAFs spanning the borders of a 64x96 image, border=5:
  current  sin(3*2pi/11)=0.98982 : 5/4000 disagreements  (inline stricter: 5, looser: 0)
  correct  sin(2*2pi/10)=0.95106 : 0/4000 disagreements
```

And it is safe to change: with the corrected constant, `ScaleSpaceDetector(num_features=500)` on a 2×1×96×128 input produces **byte-identical** LAFs and responses (`torch.equal == True`, 0 LAFs differing).

I'd fold it into this PR. The comment three lines above the constant asserts the inline block is equivalent to `laf_is_inside_image(scale_laf(current_lafs, 0.5), octave[:, 0], 5)`; this PR is already editing that block precisely to keep that assertion true, and right now the assertion is false for a reason that has nothing to do with the extent convention. Fixing it here means the block genuinely *is* the inline equivalent, and the PR body's "1 disagreement before, 2 after" paragraph can be replaced with "0 after". If you'd rather keep the PR to one concern, a follow-up issue is fine — but please don't leave the current explanation in the body, because it attributes a wrong constant to inherent approximation error, which is the sort of thing that survives for years. (Constant introduced in #3623.)

### Blast radius (checked)

`laf_is_inside_image` has no callers inside `kornia/` other than the inline copy in `_process_octave` — so within the library the behaviour change is fully absorbed by the "output is identical" result above, and the exposure is to external callers who pass a small `border`. Worth a line in the changelog entry, which you already have.

Otherwise clean.

---

ducha-aiki:
## Review applied — `ab1595e7`

Folded the constant fix into this PR, as suggested, and rewrote the PR-body paragraph that attributed the disagreement to inherent approximation error.

**`_MAX_ABS_SIN_12` now uses the spacing the reference actually samples.** `laf_to_boundary_points(laf, 12)` builds `linspace(0, 2*pi, n_pts - 1)` = `linspace(0, 2π, 11)`, spacing `2π/10`. Reproduced the finding before changing anything:

```
spacing of linspace(0, 2pi, 11)                                  : 0.62831855   (= 2pi/10)
max|sin| over those angles                                       : 0.95105654
_MAX_ABS_SIN_12 before                                           : 0.98982144   = sin(3*2pi/11)
sin(2*2pi/10)                                                    : 0.95105652
measured (max_x - cx)/half_s from laf_to_boundary_points(laf, 12) : 0.95105648
```

The comment block above the constant was wrong in the same way (`k * 2π/11 for k=0..10`) and is corrected with it.

**Equivalence, re-measured on 4000 random isotropic LAFs spanning the borders of a 64×96 image, `border=5`:**

| constant | disagreements vs `laf_is_inside_image(scale_laf(lafs, 0.5), img, 5)` |
|---|---|
| `sin(3*2pi/11)` = 0.98982 | 7 (inline stricter: 7, looser: 0) |
| `sin(2*2pi/10)` = 0.95106 | **0** |

**Output unchanged.** `ScaleSpaceDetector(num_features=500)` on a 2×1×96×128 input: `torch.equal == True` for both LAFs and responses against the old constant, 0 LAFs differing.

**Pinned.** `TestScaleSpaceDetector::test_inline_boundary_constant_matches_reference` compares the constant against the extent `laf_to_boundary_points` actually produces, so it also catches a future change to `n_pts`. It discriminates — on the old constant it fails with `assert 0.038764901868573176 < 1e-06`. Tolerance is `1e-6` rather than exact because `laf_to_boundary_points` builds its angles in float32 before casting; the wrong spacing is off by 4e-2, four orders of magnitude more.

The changelog entry now names the constant fix and states that detector output is unchanged.

```console
$ pixi run test-module tests/feature/test_scale_space_detector.py tests/feature/test_laf.py --dtype=float32,float64
149 passed, 12 skipped

$ pixi run pre-commit-all
all hooks Passed
```

On the blast-radius note: agreed, and the changelog line already names the external-caller exposure.

---

ducha-aiki:
Rebased onto current `main` (`2e2932cc`) and force-pushed — head is now **`f800496e`**, not the `ab1595e7` cited above. Only `CHANGELOG.md` conflicted (the sibling feature-audit PRs all append under `Unreleased`); resolved by keeping both sides, no code touched.

---

ducha-aiki:
## Verification review of the applied delta — verdict: **the code is correct, the CHANGELOG is not.** One change requested.

This is the pass that was missing: the constant fix landed as `ab1595e7` after the previous reviews, then the branch was rebased, and nothing re-checked the result. Verified at head `f800496e` (merge-base `2e2932cc`).

### Requested change — the CHANGELOG's last sentence is false, and it contradicts the one before it

The entry ends:

> ... inflating the tested x-extent by 4% and making the inline check stricter than the reference it claims to reproduce; the two now agree exactly. **`ScaleSpaceDetector` output is unchanged.**

Three sentences earlier the same entry says:

> ... detections within one pixel of the right or bottom edge that used to survive its `border=5` filter are now discarded.

Read straight through, the entry says the detector's output both does and does not change. The intent is clearly that the *constant* half is output-neutral — but that narrower claim is false too, and it's an unbounded statement backed by one input.

**Counterexample.** 128×160 image, a σ=4 Gaussian blob at (136, 64) plus a weak one at (40, 40), `ScaleSpaceDetector(num_features=50)`, holding `w-1`/`h-1` fixed and toggling only `_MAX_ABS_SIN_12`:

```
new constant (0.95106):  laf[0] = [[25.933, 0, 135.184], [0, 25.933, 63.532]]   resp 0.04437
old constant (0.98982):  laf[0] = [[24.190, 0,  40.000], [0, 24.190, 40.000]]   resp 0.00011

3 of 50 response slots differ; nonzero responses 3 vs 2
```

The strongest detection in the image — a response two and a half orders of magnitude above what replaces it — was being discarded by the wrong constant and now survives. Deterministic: two runs with the new constant are `torch.equal`, and the new-vs-old pair is not. Scanning blob position over `sigma ∈ {4, 6, 8, 12}` and `cx ∈ [W-40, W-2)` in 0.25 px steps, **6 of 608 positions change output**. It is rare on noise images (0 of 12 uniform-random inputs differ) precisely because it needs a detection to land in a band of width `0.039 · half_s` against the x bound — but "rare" is not "unchanged", and the case where it fires is exactly the case a user notices.

Suggested replacement for the tail: *"...the two now agree exactly. That is a loosening in x: a detection whose x-extent falls between the two constants was previously discarded and is now kept, so `ScaleSpaceDetector` output can change near the left and right edges."*

- **The spacing argument is right.** `laf_to_boundary_points` builds `torch.linspace(0, 2*pi, n_pts - 1)`, so `n_pts=12` gives 11 samples at `k · 2π/10`. Max `|sin|` is at k=2 and k=3, `sin(2π/5) = 0.9510565162951535`; max `|cos|` = 1 at k=0. The comment block now says exactly that.
- **The inline check and the reference now agree exactly.** Over **400,000** random LAFs on a 96×128 image with `border=5` (reference accepts 210,734): `_process_octave`'s inline predicate vs `laf_is_inside_image(scale_laf(lafs, 0.5), img, 5)` gives **0 disagreements** with the new constant and **1,456** with the old, all in the strict direction. That is the claim the comment above the block makes, and it now holds.
- **The new pin discriminates.** `test_inline_boundary_constant_matches_reference` measures the x-extent ratio as `0.9510565400123596`: `|measured − new| = 2.4e-08` (passes the 1e-6 tolerance), `|measured − old| = 3.9e-02` (fails by four orders of magnitude). Not a vacuous assertion.
- **Blast radius is small.** `laf_is_inside_image` has no internal callers besides the inlined copy in `scale_space_detector.py` — it's a public-API-only function, so the `w-1`/`h-1` change reaches users directly and nothing else in `kornia/` silently moves with it.
- `tests/feature/test_laf.py` + `test_scale_space_detector.py`: 112 passed / 6 skipped. The three new `laf.py` tests (exact-edge, symmetry, `border`) all read correctly and the symmetry one is the right shape of test for an asymmetric-bound bug.
- **Rebase is clean.** Pulled the pre-rebase blobs from the API: `scale_space_detector.py` and `test_scale_space_detector.py` are byte-identical to `ab1595e7`; `laf.py` and `test_laf.py` differ only by content that came in from `main` (#4072's `denormalize_laf` docstring fix, #4073's MPS clamp tests). `git diff origin/main...f800496e` shows only this PR's own hunks.

Fix the CHANGELOG sentence and this is good to go — CI is 39/39 green on the rebased head.

ducha-aiki:
CHANGELOG corrected in `6d466e84`, and the PR body's "Practical impact" section along with it — it made the same false claim.

The entry's tail now reads:

> ... the two now agree exactly. That correction is a loosening in x, so it can change `ScaleSpaceDetector` output on its own: a detection whose x-extent falls between the two constants used to be discarded and is now kept. It is rare — the detection has to land in a band of width `0.039 * half_s` against the x bound — but when it fires it can promote the strongest response in the image, so it is not output-neutral.

I re-ran the counterexample before writing that, and it reproduces at the current head with the constant toggled in isolation:

```
new constant (0.95106):  laf[0] = [[25.933, 0, 135.184], [0, 25.933, 63.532]]   resp 0.04437
old constant (0.98982):  laf[0] = [[24.190, 0,  40.000], [0, 24.190, 40.000]]   resp 0.00396
2 of 50 response slots differ; nonzero responses 2 vs 1; torch.equal(new, old) is False
```

The earlier "identical on a 2×1×96×128 random input" measurement was true but load-bearing on the input, not on the change — a random image rarely puts a detection in a band that narrow. The body now says that explicitly instead of generalising from it.

`pixi run pre-commit-all` clean.

ducha-aiki:
## Final review — approve, no asks

Re-reviewed the full diff at `6d466e84` (the state after the `_MAX_ABS_SIN_12` correction and the CHANGELOG fix). Verified on macOS/arm64, torch 2.9.1, CPU.

**The equivalence the inline block claims now actually holds**

This was the one thing worth checking independently, since `_process_octave`'s comment asserts it. Built 4000 random isotropic LAFs (scale ∈ [0, 30), centres spread over a 61×83 image) and compared the inline mask against the reference it names, `laf_is_inside_image(scale_laf(lafs, 0.5), img, 5)`:

```
inline == reference on 4000 random isotropic LAFs: True   disagreements: 0   (1660 kept by both)
with the old constant sin(3*2*pi/11):                     disagreements: 9
```

So the correction does exactly what the description says, and the pre-fix disagreement rate on random input (~0.2%) matches the "rare but not zero" framing in the CHANGELOG.

**Sampling convention confirmed from the source, not from the comment**

`laf_to_boundary_points` builds `torch.linspace(0, 2 * math.pi, n_pts - 1)` → for `n_pts=12` that is 11 angles with spacing **2π/10**, with `sin` in x and `cos` in y. Hence max|sin| = sin(2·2π/10) = 0.95106 (attained at k=2 and k=3) and max|cos| = 1.0 (k=0 and k=5). The rewritten header comment matches the code it describes.

**Other checks**

- `pytest tests/feature/test_laf.py tests/feature/test_scale_space_detector.py --dtype=float32,float64` → **160 passed, 12 skipped**.
- All four pins discriminate against `2e2932cc`. With the old `laf.py`:
  ```
  FAILED ...::test_boundary_is_the_last_valid_pixel[cpu-float32]   - assert not True
  FAILED ...::test_boundary_is_symmetric[cpu-float32]              - AssertionError: asymmetric at offset -1.0
  FAILED ...::test_border_argument_shrinks_both_edges[cpu-float32] - assert not True
  ```
  With the old `scale_space_detector.py`:
  ```
  FAILED ...::test_inline_boundary_constant_matches_reference - assert 0.038764901868573176 < 1e-06
  ```
- AGENTS.md pre-merge grep: `grep -rnE "#4064|issues/4064" kornia/ tests/` returns **nothing in `kornia/`**; the two hits are both in `tests/`, both pins naming the issue. Clean.
- The CHANGELOG tail now says the constant fix is a loosening in x that can change `ScaleSpaceDetector` output, which is what the measurements show — no surviving claim of output-neutrality.

Nothing outstanding from my side.

All three append to the same anchor in `CHANGELOG.md` (`## Unreleased` → `### Bug fixes`). `git merge-tree --write-tree` reports a conflict for **every pair**, so merging any one makes the other two conflicting. A conflicting PR runs no `pull_request` workflow and `gh pr checks` then keeps reporting the last green run from *before* the conflict — so rebase and let CI go green again between merges rather than trusting the stale check. #4084 and #4085 conflict with nothing.

Issue #4069, opened 2026-08-28
SIFTDescriptor / PatchDominantGradientOrientation / PatchAffineShapeEstimator keep kernels as plain attributes: .to() misses them and forward mutates the module


ducha-aiki:
### Describe the bug

Three modules in `kornia/feature` hold their Gaussian kernel as a plain Python attribute rather than a registered buffer. As a result `.to(device)` does not move it, it is missing from `state_dict()`, and each `forward` compensates by rebinding module state as a side effect of the input.

- https://github.com/kornia/kornia/blob/main/kornia/feature/siftdesc.py#L145 — `SIFTDescriptor.gk`
- https://github.com/kornia/kornia/blob/main/kornia/feature/orientation.py#L78 — `PatchDominantGradientOrientation.weighting`
- https://github.com/kornia/kornia/blob/main/kornia/feature/affine_shape.py#L63 — `PatchAffineShapeEstimator.weighting`

### `.to()` is silently incomplete

```python
import torch
from kornia.feature import SIFTDescriptor, PatchAffineShapeEstimator

s = SIFTDescriptor(32).to("mps")
print("parameters:", next(s.parameters()).device, " self.gk:", s.gk.device)

a = PatchAffineShapeEstimator(19).to("mps")
print("self.weighting:", a.weighting.device)
```

```
parameters: mps:0  self.gk: cpu
self.weighting: cpu
```

### The kernels are absent from `state_dict()`

```python
SIFTDescriptor(32).state_dict().keys()                    # dict_keys(['pk.weight'])
PatchDominantGradientOrientation(32).state_dict().keys()  # dict_keys(['angular_smooth.weight'])
PatchAffineShapeEstimator(19).state_dict().keys()         # dict_keys([])   <-- empty
```

`PatchAffineShapeEstimator` serialises to nothing at all, so a checkpoint round-trip silently restores none of its configuration-derived state.

### `forward` mutates the module

Because the kernels are not moved by `.to()`, five call sites re-assign module attributes inside `forward`:

```
kornia/feature/siftdesc.py:189      self.pk = self.pk.to(input.dtype).to(input.device)
kornia/feature/siftdesc.py:350      self.bin_pooling_kernel = self.bin_pooling_kernel.to(input.dtype).to(input.device)
kornia/feature/orientation.py:101   self.weighting = self.weighting.to(patch.dtype).to(patch.device)
[...]

PR #4079, opened 2026-08-29
fix(feature): register the Gaussian kernels as buffers so .to() moves them

ducha-aiki:
> Partial fix for #4069 — it closes the `.to()` half. The submodule half is deliberately left; see the end. **#4069 should stay open.**

## Problem

Three modules held their Gaussian kernel as a plain Python attribute, so `nn.Module.to()` never touched it:

```python
s = SIFTDescriptor(32).to("mps")
print(next(s.parameters()).device, s.gk.device)
```
```
mps:0 cpu
```

| module | attribute |
|---|---|
| `SIFTDescriptor` | `gk` |
| `PatchDominantGradientOrientation` | `weighting` |
| `PatchAffineShapeEstimator` | `weighting` |

Each `forward` compensated by re-assigning the attribute from the input:

```python
self.weighting = self.weighting.to(patch.dtype).to(patch.device)
```

which makes the module's persistent dtype/device depend on whichever tensor was passed last, is not safe to share across threads, and is a `torch.compile` guard hazard.

## Change

`register_buffer(..., persistent=False)`, and cast into a **local** inside `forward` rather than rebinding `self`.

```
.to() now moves the kernels          SIFTDescriptor params=mps:0  self.gk=mps:0
.double() converts them              self.gk.dtype=torch.float64
forward no longer mutates            float64 call: weighting float32 -> float32
```

### Why `persistent=False`

The kernels are fully determined by `patch_size` — derived state, not learned. Making them persistent would add keys to `state_dict()` and break existing checkpoints under `strict=True`. Keys are unchanged:

```
SIFTDescriptor                   ['pk.weight']
PatchDominantGradientOrientation ['angular_smooth.weight']
PatchAffineShapeEstimator        []
```

This does mean `PatchAffineShapeEstimator.state_dict()` stays empty, which #4069 flagged. That is the correct outcome — there is nothing learned to serialise, and compatibility wins over the cosmetic complaint.

## Behaviour

Unchanged. **Byte-identical** to `main` (`torch.equal`) for `SIFTDescriptor`, `DenseSIFTDescriptor`, `PatchDominantGradientOrientation` and `PatchAffineShapeEstimator`, in float32 and float64, and for the mismatched-input-dtype path the in-`forward` cast exists to serve:

```
sift True   sift64 True   ori True   ori64 True
aff  True   aff64  True   dense True  sift_to True
```

## Tests

Three tests per module. Five of the nine fail on unmodified sources:

```
FAILED ...TestSIFTDescriptorKernelBuffer::test_to_moves_the_kernel - assert torch.float32 == torch.float16
FAILED ...TestOrientationKernelBuffer::test_to_moves_the_kernel     - assert torch.float32 == torch.float16
FAILED ...TestOrientationKernelBuffer::test_forward_does_not_mutate_the_module
FAILED ...TestAffineShapeKernelBuffer::test_to_moves_the_kernel      - assert torch.float32 == torch.float16
FAILED ...TestAffineShapeKernelBuffer::test_forward_does_not_mutate_the_module
```

The remaining four are compat guards that pass either way by design (`test_kernel_stays_out_of_state_dict` ×3, and SIFTDescriptor's mutation test — `gk` was never among the rebound attributes there).

`test_to_moves_the_kernel` uses **float16**, not the fixture dtype: at the default float32 the assertion would hold vacuously, and float16 also works on MPS where float64 does not. The mutation test skips on MPS for the same reason.

## Verification

```console
650 passed, 126 skipped

$ pytest tests/feature -k KernelBuffer --device=mps --dtype=float32 -q
6 passed, 3 skipped

$ pixi run pre-commit-all
all hooks Passed
```

## Deliberately not in this PR

The submodule casts — `self.pk`, `self.angular_smooth`, `self.bin_pooling_kernel`, `self.PoolingConv`. Those are `nn.Module`s and *are* already carried by `.to()`; their in-`forward` conversion only fires when the input dtype/device differs, and it exists as a convenience. Removing it outright would break callers who rely on passing a float64 patch to a float32 module. Making it non-mutating means replacing the convs with functional `F.conv2d` / circular-padded `F.conv1d` calls — a bigger, riskier change that deserves its own PR.

Refs #4069


ducha-aiki (commented):
Reviewed current head `28091124` and found no actionable issues. Registering the derived Gaussian kernels as non-persistent buffers gives correct `.to()` migration without changing checkpoint keys, while the local casts preserve mismatched-input behavior without rebinding the buffers. The regression tests and full CI matrix are green.

ducha-aiki:
Applied all four Copilot comments; pushed as `fix(feature): address review — full-device assertions, precise comment, changelog`.

- **The three `.to()` device assertions** (`test_siftdesc.py`, `test_local_features_orientation.py`, `test_affine_shape_estimator.py`) compared only `device.type`, so a buffer landing on the wrong device index would have passed. Both sides now compare the full `torch.device`. Checked that this is not vacuous: the same tests run green on a real indexed device (`--device=mps` → `mps:0` on both sides, 6 passed), not only on CPU.
- **`orientation.py:108`** — the comment condemned rebinding `self.*` in general while the next line rebinds `self.angular_smooth`. It now names `self.weighting` specifically and adds that the `angular_smooth` line rebinds a submodule to itself, `nn.Module.to` being in-place, which is a different case.

Also added a CHANGELOG entry under `Unreleased → Bug fixes`, matching the sibling PRs: `.to()` / `.cuda()` / `.half()` now move the three kernels, `state_dict()` is unchanged so existing checkpoints still load with `strict=True`, and numerical output is unchanged since `forward` already cast to the input's dtype and device.

`tests/feature/test_siftdesc.py test_local_features_orientation.py test_affine_shape_estimator.py --device=cpu` → 61 passed, 5 skipped. pre-commit clean.

ducha-aiki:
## Independent review — verdict: **correct; approve.** Agreeing that #4069 stays open — and it should probably be widened.

Re-reviewed from scratch at a higher reasoning budget, re-running every claim.

### Verified

- **Byte-identical output.** I checked 11 configurations against `main` with `torch.equal`, including the mismatched-input-dtype path the in-`forward` cast exists to serve: `SIFTDescriptor`, `PatchDominantGradientOrientation`, `PatchAffineShapeEstimator`, `DenseSIFTDescriptor` in float32 and float64, plus float64-input-into-a-float32-module for the first three. All `True`.
- **`state_dict()` keys unchanged** — `['pk.weight']`, `['angular_smooth.weight']`, `[]` — and `load_state_dict(..., strict=True)` returns `<All keys matched successfully>`. The `persistent=False` reasoning is sound: all three kernels are a pure function of `patch_size` (`sigma = ks/sqrt(2)`, `ks/6.0`, `ks/sqrt(2)`), so there is genuinely nothing to serialise, and making them persistent would break every existing checkpoint.
- **`.to()` now carries them** (`.to(torch.float64)` → kernel dtype float64) and **`forward` no longer mutates** the module.
- `tests/feature/test_siftdesc.py` + `test_affine_shape_estimator.py` + `test_local_features_orientation.py`: 72 passed, 7 skipped. `pixi run typecheck`: clean apart from the pre-existing `lightglue.py` `custom_fwd` deprecation. Note `TestPatchDominantGradientOrientation::test_jit` is skipped, so nothing here was pinned against a TorchScript regression either way.

### No hidden perf cost from casting into a local

Worth recording, since dropping the `self.weighting = ...` memoisation *looks* like it turns a once-per-module conversion into a per-call one. It doesn't: `Tensor.to()` returns the identical object when dtype and device already match —

```
t.to(torch.float32).to('cpu') is t  ->  True
```

— so the steady state is a no-op, and the only case that now copies per call (module on CPU, input on CUDA) is exactly the case this PR makes avoidable by teaching `.to()` to move the kernel.

### The same defect class exists outside `kornia.feature`

I swept the whole package for tensor attributes assigned in `__init__` and never registered. The three this PR fixes were not the only ones:

| location | attribute |
|---|---|
| `kornia/filters/sobel.py:261` | `SpatialGradient3d.kernel` |
| `kornia/feature/defmo.py:342` | `RenderingDeFMO.times` |
| `kornia/feature/sold2/sold2_detector.py:188` | `LineSegmentDetectionModule.torch_sampler` |
| `kornia/augmentation/_2d/intensity/planckian_jitter.py` | `RandomPlanckianJitter.pl` (with a `self.pl = self.pl.to(...)` rebind in `forward`, same shape as the ones fixed here) |
| `kornia/enhance/rescale.py:35` | `Rescale.factor` |
| `kornia/augmentation/_2d/mix/transplantation.py` | `excluded_labels` (rebound in `forward`) |

`SpatialGradient3d.kernel` is the closest sibling — an `nn.Module` whose convolution kernel is a plain attribute. So keeping #4069 open is right, and I'd retitle/expand it from "three modules in kornia.feature" to the repo-wide pattern, with this PR closing the `kornia.feature` slice.

### Adjacent nit

`SIFTDescriptor.get_gaussian_kernel()` returns `self.gk.detach()` — now a live alias of a *registered buffer*, so a caller writing into the returned tensor mutates the module. Same wart as `DenseSIFTDescriptor.get_pooling_kernel()` (I raised it on #4074); one issue can cover both.

### Test-convention nit (non-blocking)

`test_kernel_stays_out_of_state_dict(self, device)` and `test_forward_does_not_mutate_the_module(self, device)` take the `device` fixture but never use it — both construct a CPU module and a CPU tensor. On a CUDA or MPS run they re-execute identical CPU work under a device-suffixed test id. `test_to_moves_the_kernel` does use it properly. Not worth a round-trip on its own, but if you touch these again, either drop the fixture or move the module to `device`.

Approving as-is.

---

PR #4080, opened 2026-08-29
fix(feature): make integrated.py's empty results match its non-empty ones

### LINKED ISSUE #4065: integrated.py: empty-result paths disagree with the normal paths (hardcoded 128 dims, LAF rank, negative confidence)
### Describe the bug

Three related inconsistencies in `kornia/feature/integrated.py`, all on the "nothing was found" paths, where the empty result does not have the same shape or semantics as the non-empty one.

#### 1. `get_laf_descriptors` hardcodes a descriptor width of 128 when there are no keypoints

https://github.com/kornia/kornia/blob/main/kornia/feature/integrated.py#L75

```python
if lafs.shape[1] == 0:
    warnings.warn(...)
    return torch.empty(lafs.shape[0], lafs.shape[1], 128, dtype=lafs.dtype, device=lafs.device)
```

The width is taken from the descriptor module on the normal path but hardcoded to 128 here, so the returned descriptor dimension depends on whether any keypoints were detected.

```python
import torch, warnings
from kornia.feature import get_laf_descriptors, MKDDescriptor
from kornia.feature.laf import laf_from_center_scale_ori

img = torch.rand(1, 1, 64, 64)
laf_ok = laf_from_center_scale_ori(torch.tensor([[[32., 32.]]]), torch.full((1, 1, 1, 1), 8.0))
laf_empty = torch.zeros(1, 0, 2, 3)

for dims in (64, 128, 238):
    desc = MKDDescriptor(patch_size=32, output_dims=dims)
    with warnings.catch_warnings():
        warnings.simplefilter("ignore")
        print(f"output_dims={dims:4d}  non-empty={get_laf_descriptors(img, laf_ok, desc, 32).shape[-1]:4d}"
              f"  empty={get_laf_descriptors(img, laf_empty, desc, 32).shape[-1]:4d}")
```

```
output_dims=  64  non-empty=  64  empty= 128
output_dims= 128  non-empty= 128  empty= 128
output_dims= 238  non-empty= 238  empty= 128
```

The empty tensor also takes `dtype=lafs.dtype` rather than the descriptor's output dtype.

#### 2. `LocalFeatureMatcher` returns a different LAF rank when there are no matches

`no_match_output` returns `lafs0` / `lafs1` with shape `(0, 0, 2, 3)`:

https://github.com/kornia/kornia/blob/main/kornia/feature/integrated.py#L407

while the success path returns `(1, N, 2, 3)`:

https://github.com/kornia/kornia/blob/main/kornia/feature/integrated.py#L483

The batch dimension is 0 in one case and 1 in the other, so caller code that does `out["lafs0"][0]` works on a successful match and raises on an empty one. The sibling `keypoints0` is consistent between the two paths (`(0, 2)` vs `(-1, 2)`); only the LAFs disagree.

#### 3. `confidence` can be negative

https://github.com/kornia/kornia/blob/main/kornia/feature/integrated.py#L469

```python
out_confidence.append(1.0 - dists)
```

`forward` documents `confidence` as a *"confidence score [0, 1]"*. That holds when the matcher returns a ratio (`snn`, `smnn`, `fginn`), but `match_nn` / `match_mnn` return raw L2 descriptor distances, which reach 2.0 for unit-norm descriptors — so `DescriptorMatcher("nn")` yields confidences down to −1.

### Expected behavior

1. Empty-path descriptor width should come from the descriptor module, and the dtype should match the non-empty path.
2. `no_match_output` should return `lafs0` / `lafs1` with the same rank and batch dimension as the success path.
3. Either normalise the distance before converting it to a confidence, or document that `confidence` is only in `[0, 1]` for ratio-based matchers.

### Additional context

`integrated.py` sits at 51% line coverage, the lowest of the hand-written (non-vendored) files in `kornia/feature`; items 2 and 3 are on lines that no test exercises.

ducha-aiki:
## Problem

Three "nothing was found" paths in `integrated.py` disagreed with their corresponding normal paths.

### 1. `get_laf_descriptors` hardcoded a width of 128

```python
if lafs.shape[1] == 0:
    return torch.empty(lafs.shape[0], lafs.shape[1], 128, dtype=lafs.dtype, device=lafs.device)
```

so the descriptor width depended on whether anything was detected:

```
MKDDescriptor(output_dims= 64)   non-empty=  64   empty= 128
MKDDescriptor(output_dims=128)   non-empty= 128   empty= 128
MKDDescriptor(output_dims=238)   non-empty= 238   empty= 128
```

The dtype came from `lafs` rather than from the descriptor's output too.

**Fix:** probe the descriptor with a single dummy patch and take the width, dtype and device from that. The grayscale conversion moves above the early return so the probe uses the correct channel count.

The width is `probe.numel()`, not `probe.shape[-1]`. The non-empty path ends in `.view(B, N, -1)`, which flattens *everything* the descriptor produced per patch, so the last dimension only agrees for descriptors whose output is 2-D. `patch_descriptor` is typed as a bare `nn.Module`, and `DenseSIFTDescriptor` — or any user module returning `(N, C, H, W)` — is not:

```
DenseSIFTDescriptor    non-empty=(1, 1, 131072)   empty=(1, 0, 32)
  torch.cat FAILS: Expected size 131072 but got size 32
```

which would have reintroduced the exact defect this PR removes. The probe has a batch of one, so its numel is by construction the width `.view(B, N, -1)` produces.

**New failure surface:** the empty path now *executes* the descriptor, so a third-party descriptor that raises on a zero patch raises here where it previously returned quietly. Rare and cheap enough not to guard, but it is a real widening and is named in the changelog.

Checked that a zero patch is safe for every patch descriptor in `kornia.feature` — `SIFTDescriptor`, `MKDDescriptor` (64/128/238), `HardNet`, `HardNet8`, `HyNet`, `TFeat`, `SOSNet` — all return finite output and none raise.

After:
```
SIFTDescriptor(32)     non-empty= 128  empty= 128
MKD(output_dims=64)    non-empty=  64  empty=  64
MKD(output_dims=238)   non-empty= 238  empty= 238
```

### 2. `no_match_output` dropped the batch dimension

It returned `lafs0`/`lafs1` as `(0, 0, 2, 3)` while the success path returns `.view(1, -1, 2, 3)`. So `out["lafs0"][0]` worked when something matched and raised when nothing did. The sibling `keypoints0` was already consistent (`(0, 2)` vs `(-1, 2)`); only the LAFs disagreed.

**Fix:** `(1, 0, 2, 3)`.

This is a shape change on a public return value: code using `out["lafs0"].shape[0] == 0` as its no-match signal now sees `1`. `out["keypoints0"].shape[0] == 0` was already the consistent signal and is unaffected. Named explicitly in the changelog.

### 3. `confidence` is documented as `[0, 1]` but can reach −1

`confidence = 1.0 - dists`. That holds for the ratio-based `DescriptorMatcher` modes (`snn`, `smnn`), but `match_nn` / `match_mnn` return raw L2 distances, which reach 2.0 for unit-norm descriptors. (`fginn` and `adalam` are ratio-based too, but they belong to `GeometryAwareDescriptorMatcher`, whose `forward` takes `lafs1`/`lafs2` as well — `LocalFeatureMatcher` calls its matcher with two arguments, so those modes cannot reach this code path and are not listed.)

**Fix: documentation only.** Clamping would silently change user-visible confidence values and destroy ordering for the `nn` matchers, so the docstring now states the real range and which matchers guarantee `[0, 1]`. Since the function accepts arbitrary descriptors, the raw distance is unbounded in general: the docstring says the confidence has no lower bound and goes *at least* down to −1, which is the unit-norm case.

## Tests

Both code fixes are pinned, and the new tests fail on unmodified `integrated.py`:

```
FAILED ...TestGetLAFDescriptorsEmptyPath::test_empty_matches_the_descriptor_width[cpu-float32-64]  - assert torch.Size([1, 0, 128]) == (1, 0, 64)
FAILED ...TestGetLAFDescriptorsEmptyPath::test_empty_matches_the_descriptor_width[cpu-float32-238] - assert torch.Size([1, 0, 128]) == (1, 0, 238)
FAILED ...TestLocalFeatureMatcherEmptyPath::test_empty_lafs_keep_the_batch_dimension[cpu-float32]  - assert torch.Size([0, 0, 2, 3]) == (1, 0, 2, 3)
```

A third test pins the non-2-D case, parametrised over a 2-line `nn.Module` returning `(N, 3, 2, 5)` and a real `DenseSIFTDescriptor`, and asserts the concatenation rather than just the width. Both fail on `probe.shape[-1]`:

```
FAILED ...::test_empty_matches_a_non_2d_descriptor[cpu-float32-four_dim]    - assert torch.Size([1, 0, 5])  == (1, 0, 30)
FAILED ...::test_empty_matches_a_non_2d_descriptor[cpu-float32-dense_sift]  - assert torch.Size([1, 0, 32]) == (1, 0, 131072)
```

(The `output_dims=128` case passes either way — it is the value that was hardcoded — and is kept as a parametrisation so the fix is exercised at the previously-correct width too.)

## Note on `integrated.py` coverage

This file sits at **51% line coverage**, the lowest of the hand-written (non-vendored) files in `kornia/feature`. Items 2 and 3 were both on lines no test touched.

## Verification

```console
649 passed, 126 skipped

$ pixi run pre-commit-all
all hooks Passed
```

Closes #4065

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `ScaleSpaceDetector` output on its own: a detection whose x-extent falls between the two constants used to be
   discarded and is now kept. It is rare -- the detection has to land in a band of width `0.039 * half_s` against the
   x bound -- but when it fires it can promote the strongest response in the image, so it is not output-neutral.
+* Fix the empty-result paths in `kornia.feature.integrated` disagreeing with the corresponding non-empty ones
+  (#4065). `get_laf_descriptors` returned a hardcoded width of 128 with the *LAF's* dtype and device when it was
+  handed no keypoints, so an empty batch through, say, a 256-d descriptor produced a `(B, 0, 128)` result that could
+  not be concatenated with a real one; it now probes the descriptor once and returns its actual width, dtype and
+  device. `LocalFeatureMatcher.no_match_output` returned `lafs0`/`lafs1` of shape `(0, 0, 2, 3)` where the success
+  path returns `(1, NC, 2, 3)`, so `lafs0[0]` raised `IndexError` exactly when nothing matched; the empty tensors
+  now keep the leading batch of 1. `LocalFeatureMatcher.forward`'s docstring also stops claiming `confidence` is in
+  `[0, 1]`: it is `1 - distance`, which for the raw-distance matchers `nn` and `mnn` reaches -1 on unit-norm
+  descriptors.
 
 
 ## :rocket: [0.6.11] - 2022-03-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -157,12 +157,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fix the empty-result paths in `kornia.feature.integrated` disagreeing with the corresponding non-empty ones
   (#4065). `get_laf_descriptors` returned a hardcoded width of 128 with the *LAF's* dtype and device when it was
   handed no keypoints, so an empty batch through, say, a 256-d descriptor produced a `(B, 0, 128)` result that could
-  not be concatenated with a real one; it now probes the descriptor once and returns its actual width, dtype and
-  device. `LocalFeatureMatcher.no_match_output` returned `lafs0`/`lafs1` of shape `(0, 0, 2, 3)` where the success
-  path returns `(1, NC, 2, 3)`, so `lafs0[0]` raised `IndexError` exactly when nothing matched; the empty tensors
-  now keep the leading batch of 1. `LocalFeatureMatcher.forward`'s docstring also stops claiming `confidence` is in
-  `[0, 1]`: it is `1 - distance`, which for the raw-distance matchers `nn` and `mnn` reaches -1 on unit-norm
-  descriptors.
+  not be concatenated with a real one; it now runs the descriptor once on a zero patch and returns its actual width
+  (the flattened per-patch output, matching the non-empty path's `.view(B, N, -1)`), dtype and device. That is a
+  widening of the empty path's failure surface: a third-party descriptor that raises on a zero input now raises here
+  where it previously returned quietly. `LocalFeatureMatcher.no_match_output` returned `lafs0`/`lafs1` of shape
+  `(0, 0, 2, 3)` where the success path returns `(1, NC, 2, 3)`, so `lafs0[0]` raised `IndexError` exactly when
+  nothing matched; the empty tensors now keep the leading batch of 1. That is a shape change on a public return
+  value -- code using `out["lafs0"].shape[0] == 0` as its no-match signal now sees `1` and should test
+  `out["keypoints0"].shape[0] == 0`, which was already the consistent signal. `LocalFeatureMatcher.forward`'s
+  docstring also stops claiming `confidence` is in `[0, 1]`: it is `1 - distance`, which for the raw-distance
+  matchers `nn` and `mnn` goes at least down to -1 on unit-norm descriptors, and is unbounded below for arbitrary
+  ones.
 
 
 ## :rocket: [0.6.11] - 2022-03-28
--- a/kornia/feature/integrated.py
+++ b/kornia/feature/integrated.py
@@ -78,7 +78,9 @@ def get_laf_descriptors(
         # Probe the descriptor for its output width instead of assuming 128, so the empty result
         # has the same trailing dimension and dtype as a non-empty one.
         with torch.no_grad():
-            probe = patch_descriptor(torch.zeros(1, timg.size(1), patch_size, patch_size).to(timg))
+            probe = patch_descriptor(
+                torch.zeros(1, timg.size(1), patch_size, patch_size, dtype=timg.dtype, device=timg.device)
+            )
         return torch.empty(lafs.shape[0], 0, probe.shape[-1], dtype=probe.dtype, device=probe.device)
 
     patches: torch.Tensor = extract_patches_from_pyramid(timg, lafs, patch_size)
--- a/kornia/feature/integrated.py
+++ b/kornia/feature/integrated.py
@@ -76,12 +76,15 @@ def get_laf_descriptors(
     if lafs.shape[1] == 0:
         warnings.warn(f"LAF contains no keypoints {lafs.shape}, returning empty torch.Tensor", stacklevel=1)
         # Probe the descriptor for its output width instead of assuming 128, so the empty result
-        # has the same trailing dimension and dtype as a non-empty one.
+        # has the same trailing dimension and dtype as a non-empty one. The non-empty path ends in
+        # `.view(B, N, -1)`, which flattens everything the descriptor produced per patch, so the
+        # matching width is the whole probe -- `probe.shape[-1]` would only agree for descriptors
+        # whose output is 2-D. The probe has a batch of one, so its numel is that width.
         with torch.no_grad():
             probe = patch_descriptor(
                 torch.zeros(1, timg.size(1), patch_size, patch_size, dtype=timg.dtype, device=timg.device)
             )
-        return torch.empty(lafs.shape[0], 0, probe.shape[-1], dtype=probe.dtype, device=probe.device)
+        return torch.empty(lafs.shape[0], 0, probe.numel(), dtype=probe.dtype, device=probe.device)
 
     patches: torch.Tensor = extract_patches_from_pyramid(timg, lafs, patch_size)
     # Descriptor accepts standard torch.Tensor [B, CH, H, W], while patches are [B, N, CH, H, W] shape
@@ -436,8 +439,9 @@ class LocalFeatureMatcher(nn.Module):
             - ``keypoints1``, matching keypoints from image1 :math:`(NC, 2)`.
             - ``confidence``, ``1 - descriptor distance`` :math:`(NC)`. This lies in :math:`[0, 1]`
               only for ratio-based matchers (``snn``, ``smnn``, ``fginn``, ``adalam``); ``nn`` and
-              ``mnn`` return raw distances, which reach 2.0 for unit-norm descriptors and therefore
-              give confidences down to -1.
+              ``mnn`` return raw distances, which are unbounded for an arbitrary descriptor, so the
+              confidence has no lower bound in general -- it goes at least down to -1, the value
+              reached by unit-norm descriptors, whose distance tops out at 2.0.
             - ``lafs0``, matching LAFs from image0 :math:`(1, NC, 2, 3)`.
             - ``lafs1``, matching LAFs from image1 :math:`(1, NC, 2, 3)`.
             - ``batch_indexes``, batch indexes for the keypoints and lafs :math:`(NC)`.
--- a/kornia/feature/integrated.py
+++ b/kornia/feature/integrated.py
@@ -438,7 +438,7 @@ class LocalFeatureMatcher(nn.Module):
             - ``keypoints0``, matching keypoints from image0 :math:`(NC, 2)`.
             - ``keypoints1``, matching keypoints from image1 :math:`(NC, 2)`.
             - ``confidence``, ``1 - descriptor distance`` :math:`(NC)`. This lies in :math:`[0, 1]`
-              only for ratio-based matchers (``snn``, ``smnn``, ``fginn``, ``adalam``); ``nn`` and
+              only for the ratio-based ``DescriptorMatcher`` modes (``snn``, ``smnn``); ``nn`` and
               ``mnn`` return raw distances, which are unbounded for an arbitrary descriptor, so the
               confidence has no lower bound in general -- it goes at least down to -1, the value
               reached by unit-norm descriptors, whose distance tops out at 2.0.
--- a/kornia/feature/integrated.py
+++ b/kornia/feature/integrated.py
@@ -70,11 +70,16 @@ def get_laf_descriptors(
     patch_descriptor.eval()
 
     timg: torch.Tensor = img
-    if lafs.shape[1] == 0:
-        warnings.warn(f"LAF contains no keypoints {lafs.shape}, returning empty torch.Tensor", stacklevel=1)
-        return torch.empty(lafs.shape[0], lafs.shape[1], 128, dtype=lafs.dtype, device=lafs.device)
     if grayscale_descriptor and img.size(1) == 3:
         timg = rgb_to_grayscale(img)
+
+    if lafs.shape[1] == 0:
+        warnings.warn(f"LAF contains no keypoints {lafs.shape}, returning empty torch.Tensor", stacklevel=1)
+        # Probe the descriptor for its output width instead of assuming 128, so the empty result
+        # has the same trailing dimension and dtype as a non-empty one.
+        with torch.no_grad():
+            probe = patch_descriptor(torch.zeros(1, timg.size(1), patch_size, patch_size).to(timg))
+        return torch.empty(lafs.shape[0], 0, probe.shape[-1], dtype=probe.dtype, device=probe.device)
 
     patches: torch.Tensor = extract_patches_from_pyramid(timg, lafs, patch_size)
     # Descriptor accepts standard torch.Tensor [B, CH, H, W], while patches are [B, N, CH, H, W] shape
@@ -404,8 +409,10 @@ class LocalFeatureMatcher(nn.Module):
         return {
             "keypoints0": torch.empty(0, 2, device=device, dtype=dtype),
             "keypoints1": torch.empty(0, 2, device=device, dtype=dtype),
-            "lafs0": torch.empty(0, 0, 2, 3, device=device, dtype=dtype),
-            "lafs1": torch.empty(0, 0, 2, 3, device=device, dtype=dtype),
+            # batch dim 1, matching the success path's `.view(1, -1, 2, 3)`, so that callers
+            # can index `lafs0[0]` regardless of whether anything matched.
+            "lafs0": torch.empty(1, 0, 2, 3, device=device, dtype=dtype),
+            "lafs1": torch.empty(1, 0, 2, 3, device=device, dtype=dtype),
             "confidence": torch.empty(0, device=device, dtype=dtype),
             "batch_indexes": torch.empty(0, device=device, dtype=torch.long),
         }
@@ -425,7 +432,10 @@ class LocalFeatureMatcher(nn.Module):
         Returns:
             - ``keypoints0``, matching keypoints from image0 :math:`(NC, 2)`.
             - ``keypoints1``, matching keypoints from image1 :math:`(NC, 2)`.
-            - ``confidence``, confidence score [0, 1] :math:`(NC)`.
+            - ``confidence``, ``1 - descriptor distance`` :math:`(NC)`. This lies in :math:`[0, 1]`
+              only for ratio-based matchers (``snn``, ``smnn``, ``fginn``, ``adalam``); ``nn`` and
+              ``mnn`` return raw distances, which reach 2.0 for unit-norm descriptors and therefore
+              give confidences down to -1.
             - ``lafs0``, matching LAFs from image0 :math:`(1, NC, 2, 3)`.
             - ``lafs1``, matching LAFs from image1 :math:`(1, NC, 2, 3)`.
             - ``batch_indexes``, batch indexes for the keypoints and lafs :math:`(NC)`.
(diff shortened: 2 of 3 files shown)


ducha-aiki (commented):
Reviewed current head `db1b395b` and found no actionable issues. The descriptor probe correctly derives empty-result width, dtype, and device from actual descriptor behavior, empty matcher LAFs now match the non-empty rank, and the confidence documentation reflects ratio versus raw-distance matchers. Regression tests and the full CI matrix are green.

ducha-aiki:
Applied Copilot's comment; pushed as `fix(feature): build the descriptor probe on the image's device; add changelog`.

- **`integrated.py:81`** — the probe was `torch.zeros(...).to(timg)`, allocating on the default device and dtype and then copying. It now takes `dtype=timg.dtype, device=timg.device` directly, so there is no extra allocation or host-to-device transfer on the empty path.

Also added a CHANGELOG entry under `Unreleased → Bug fixes`, matching the sibling PRs: the empty descriptor's width, dtype and device change (a 256-d descriptor used to yield a `(B, 0, 128)` tensor in the *LAF's* dtype), `lafs0`/`lafs1` gain their leading batch of 1 so `lafs0[0]` no longer raises `IndexError` exactly when nothing matched, and the `confidence` range claim is corrected.

`tests/feature/test_integrated.py --device=cpu` → 17 passed, 10 skipped. pre-commit clean.

ducha-aiki:
## Independent review — verdict: **right idea, one line short of correct.** The probe reintroduces the same class of inconsistency it removes.

Re-reviewed from scratch at a higher reasoning budget.

### Finding: `probe.shape[-1]` doesn't match what the non-empty path returns

The non-empty path is

```python
return patch_descriptor(patches.view(B * N, CH, H, W)).view(B, N, -1)
```

`.view(B, N, -1)` flattens **everything** the descriptor produced per patch. The empty path takes `probe.shape[-1]`, which only agrees when the descriptor's output is 2-D. It isn't always:

```
SIFTDescriptor         non-empty=(1, 1, 128)     empty=(1, 0, 128)     OK
MKD(64)                non-empty=(1, 1, 64)      empty=(1, 0, 64)      OK
HardNet / TFeat        non-empty=(1, 1, 128)     empty=(1, 0, 128)     OK
SOSNet / HyNet         non-empty=(1, 1, 128)     empty=(1, 0, 128)     OK
DenseSIFTDescriptor    non-empty=(1, 1, 131072)  empty=(1, 0, 32)      <-- MISMATCH
```

`DenseSIFTDescriptor` as a `patch_descriptor` is unusual, but nothing forbids it and `patch_descriptor` is typed as a bare `nn.Module` — a user descriptor returning `(N, C, H, W)` or `(N, H, D)` hits this too. And this is precisely the bug the PR exists to remove: an empty batch that can't be concatenated with a non-empty one.

The probe has batch 1, so the exactly-consistent expression is `probe.numel()`, or if you prefer it to read symmetrically with the other branch, `probe.reshape(1, -1).shape[-1]`. One token, and then the empty path agrees with `.view(B, N, -1)` for *every* descriptor rather than for the 2-D ones. Worth adding `DenseSIFTDescriptor` (or a 2-line dummy module returning a 4-D tensor) to the parametrisation so it stays pinned.

- All three new tests fail on unmodified `integrated.py` with the shapes claimed; branch is green (`tests/feature/test_integrated.py`: 17 passed, 10 skipped).
- Moving the grayscale conversion above the early return is necessary and correct — the probe must see the same channel count the real path would, and it is guarded by `img.size(1) == 3`.
- Taking dtype/device from `probe` rather than `lafs` is right: the non-empty path's dtype is the descriptor's output dtype, not the LAF's.
- A zero patch is safe for all seven kornia patch descriptors (verified: none raises, all finite). The probe *does* now execute the descriptor on the empty path, which is a new way for a third-party descriptor to raise where it previously returned quietly — cheap and rare enough that I wouldn't guard it, but it is a real widening of the empty path's failure surface and belongs in the PR body.
- `no_match_output` → `(1, 0, 2, 3)` matches the success path's `.view(1, -1, 2, 3)`; end-to-end confirmed:
  ```
  {'keypoints0': (0, 2), 'lafs0': (1, 0, 2, 3), 'lafs1': (1, 0, 2, 3), 'confidence': (0,), ...}
  lafs0[0] -> (0, 2, 3)
  ```
  It is technically a shape change on a public return value — anyone testing `out["lafs0"].shape[0] == 0` as their no-match signal now gets `1`. Since the documented shape has always been `(1, NC, 2, 3)` and `keypoints0` was already the consistent signal, I think that's the correct break, but a changelog line naming it explicitly wouldn't hurt.
- The `confidence` documentation change is accurate: `1.0 - dists` at `integrated.py:481`, and `match_nn`/`match_mnn` return raw L2. One precision point — the new text bounds the low end at −1 "for unit-norm descriptors", which is right for kornia's own descriptors but the function accepts arbitrary ones, where it is unbounded below. "at least down to −1" would be strictly true.

Fix the width and I'm happy.

---

ducha-aiki:
## Review applied — `405f47b9`

All four points. The width finding is right, and it reproduces exactly as described.

### 1. Width — `probe.shape[-1]` → `probe.numel()`

Reproduced on the branch before changing anything:

```
SIFTDescriptor         non-empty=(1, 1, 128)     empty=(1, 0, 128)   OK,  cat works
DenseSIFTDescriptor    non-empty=(1, 1, 131072)  empty=(1, 0, 32)    <-- MISMATCH
  torch.cat FAILS: Sizes of tensors must match except in dimension 1.
                   Expected size 131072 but got size 32
```

So the empty path was reintroducing the exact defect it exists to remove, for any descriptor whose output is not 2-D. The probe has a batch of one, so `probe.numel()` is by construction what `.view(B, N, -1)` produces per patch, for every descriptor. The comment above it now says why, so `shape[-1]` does not creep back.

**Pinned**, parametrised over both cases you named — a 2-line `nn.Module` returning `(N, 3, 2, 5)` and a real `DenseSIFTDescriptor` — and asserting the concatenation itself, not just the width. Both fail on `probe.shape[-1]`:

```
FAILED ...test_empty_matches_a_non_2d_descriptor[cpu-float32-four_dim]
  - assert torch.Size([1, 0, 5]) == (1, 0, 30)
FAILED ...test_empty_matches_a_non_2d_descriptor[cpu-float32-dense_sift]
  - assert torch.Size([1, 0, 32]) == (1, 0, 131072)
```

### 2. `confidence` lower bound

Agreed — "for unit-norm descriptors" was doing work the sentence didn't earn. Now:

> ``nn`` and ``mnn`` return raw distances, which are unbounded for an arbitrary descriptor, so the confidence has no lower bound in general -- it goes at least down to -1, the value reached by unit-norm descriptors, whose distance tops out at 2.0.

### 3 & 4. Changelog

Both now named explicitly rather than left to be inferred:

- **the probe's widened failure surface** — "a third-party descriptor that raises on a zero input now raises here where it previously returned quietly";
- **the `no_match_output` shape change** — "code using `out["lafs0"].shape[0] == 0` as its no-match signal now sees `1` and should test `out["keypoints0"].shape[0] == 0`, which was already the consistent signal."

### Verification

```console
$ pixi run test-module tests/feature/test_integrated.py --dtype=float32,float64
33 passed, 16 skipped

$ pixi run pre-commit-all
all hooks Passed
```

`pixi run doctest` has 4 failures (`matching.py::DescriptorMatcherWithSteerer`, `kernels.py::get_gaussian_erf_kernel1d`, `kernels.py::get_gaussian_kernel3d`, `se2.py::Se2.log`) — all four reproduce identically on detached `origin/main`, none in a file this PR touches.

---

ducha-aiki:
Rebased onto current `main` (`2e2932cc`) and force-pushed — head is now **`33985002`**, not the `405f47b9` cited above. Only `CHANGELOG.md` conflicted (the sibling feature-audit PRs all append under `Unreleased`); resolved by keeping both sides, no code touched.

---

ducha-aiki:
## Verification review of the applied delta — verdict: **the code is correct; one documentation error to fix.**

This is the pass that was missing: the width fix landed as `405f47b9` after the previous reviews, then the branch was rebased, and nothing re-checked the result. Verified at head `33985002` (merge-base `2e2932cc`).

### Requested change — the new `confidence` docstring names two matchers this class cannot use

`LocalFeatureMatcher.forward` now says:

> ``confidence``, ``1 - descriptor distance``. This lies in :math:`[0, 1]` only for ratio-based matchers (``snn``, ``smnn``, **``fginn``, ``adalam``**); ``nn`` and ``mnn`` return raw distances...

`fginn` and `adalam` are not `DescriptorMatcher` modes, and they cannot reach this code path at all:

```
DescriptorMatcher.known_modes            -> ['nn', 'mnn', 'snn', 'smnn']
DescriptorMatcher('fginn')               -> NotImplementedError: fginn is not supported.
DescriptorMatcher('adalam')              -> NotImplementedError: adalam is not supported.
GeometryAwareDescriptorMatcher.known_modes -> ['fginn', 'adalam']
LocalFeatureMatcher(SIFTFeature(50), GeometryAwareDescriptorMatcher('fginn'))(data)
  -> TypeError: GeometryAwareDescriptorMatcher.forward() missing 2 required positional
     arguments: 'lafs1' and 'lafs2'
```

They live on `GeometryAwareDescriptorMatcher`, whose `forward(desc1, desc2, lafs1, lafs2)` takes four arguments, while `LocalFeatureMatcher.forward` calls `self.matcher(descs0[batch_idx], descs1[batch_idx])` with two. So the sentence sends a reader looking for `[0, 1]` confidences to two modes that raise before they ever produce one. The `matcher` argument is documented as `DescriptorMatcher`, and its ratio modes are exactly `snn` and `smnn`.

Same wording appears in the CHANGELOG entry's last clause, which correctly says only "`nn` and `mnn`" — so the CHANGELOG is fine and it's just the docstring that over-lists. Dropping `fginn`/`adalam` is enough; if the geometry-aware matchers are worth a mention, they need a separate sentence saying they belong to a different class.

The rest of that docstring change is accurate, checked against the code: `out_confidence.append(1.0 - dists)` at line 485 confirms the `1 - distance` wording, `match_snn` returns `vals[:, 0] / vals[:, 1]` thresholded below `th ≤ 1` so `snn`/`smnn` really do land in `[0, 1]`, and the `-1` floor is right — `match_nn` uses `torch.cdist`, whose maximum over unit-norm descriptors is 2.0.

- **`probe.numel()` is the right width, and the pin proves it.** Reverting just that one token to `probe.shape[-1]` fails both new parametrisations: `DenseSIFTDescriptor` gives `(1, 0, 32)` against a non-empty `(1, 0, 131072)`, and the 4-D dummy gives `5` against `30`. Restored to `numel()`, all six tests pass. That is exactly the `torch.cat` failure the empty path exists to prevent, so the fix closes the hole the earlier version left open.
- **The reordering is necessary and consistent.** Moving the empty check below the `rgb_to_grayscale` block is what makes `timg.size(1)` correct for the probe, and the probe's dtype/device now follow the descriptor's output — which is what the non-empty path returns too, rather than the previous `lafs.dtype`/`lafs.device`. `torch.no_grad()` around a probe whose value is discarded is right.
- **`no_match_output` rank matches the success path**, which ends in `.view(1, -1, 2, 3)` at line 496–497. `(1, 0, 2, 3)` is the correct shape, and `lafs0[0]` now works in both branches. The CHANGELOG is unusually good about this one: it names the shape change as breaking, names the code pattern that breaks (`out["lafs0"].shape[0] == 0`), and gives the migration (`out["keypoints0"].shape[0] == 0`). `no_match_output` has exactly one internal caller (line 494), so nothing else in `kornia/` moves with it.
- The widened failure surface — a third-party descriptor that raises on a zero patch now raises here — is disclosed in the CHANGELOG rather than left implicit. Correct call: it's the price of not hardcoding a width, and the alternative (guessing 128) is the bug.
- **Rebase is clean.** Pulled the pre-rebase blobs from the API: both `kornia/feature/integrated.py` and `tests/feature/test_integrated.py` are byte-identical to `405f47b9`; only the CHANGELOG was touched by the conflict resolution, and `git diff origin/main...33985002` shows a single clean addition there.
- `tests/feature/test_integrated.py -k EmptyPath`: 6 passed. CI is 39/39 green on the rebased head.

Fix the docstring mode list and this is ready.

ducha-aiki:
Docstring fixed in `863a7bdc`:

```
- ``confidence``, ``1 - descriptor distance`` :math:`(NC)`. This lies in :math:`[0, 1]`
  only for the ratio-based ``DescriptorMatcher`` modes (``snn``, ``smnn``); ``nn`` and
  ``mnn`` return raw distances, ...
```

`fginn`/`adalam` are gone from the list for the reason given — they are `GeometryAwareDescriptorMatcher` modes, and that class's `forward(desc1, desc2, lafs1, lafs2)` cannot be driven by `LocalFeatureMatcher.forward`'s two-argument call. The PR body's section 3 carried the same over-listing and now carries the correction, with a parenthetical saying where those two modes actually live so the information is not simply deleted.

`pixi run pre-commit-all` clean; `tests/feature/test_integrated.py -k "EmptyPath or empty"` 7 passed; module doctest 1 passed.

ducha-aiki:
## Final review — approve, with one optional placement ask

Re-reviewed the full diff at `863a7bdc` (the state after the `probe.numel()` fix and the `confidence` docstring trim). Verified on macOS/arm64, torch 2.9.1, CPU.

**Checks run**

- `pytest tests/feature/test_integrated.py --dtype=float32` → **19 passed, 10 skipped**; the six new tests pass.
- Five of the six discriminate against `2e2932cc` (the `output_dims=128` parametrisation passes either way by construction, as the description says):
  ```
  FAILED ...::test_empty_matches_the_descriptor_width[cpu-float32-64]        - assert torch.Size([1, 0, 128]) == (1, 0, 64)
  FAILED ...::test_empty_matches_the_descriptor_width[cpu-float32-238]       - assert torch.Size([1, 0, 128]) == (1, 0, 238)
  FAILED ...::test_empty_matches_a_non_2d_descriptor[cpu-float32-four_dim]   - assert torch.Size([1, 0, 128]) == (1, 0, 30)
  FAILED ...::test_empty_matches_a_non_2d_descriptor[cpu-float32-dense_sift] - assert torch.Size([1, 0, 128]) == (1, 0, 131072)
  FAILED ...::test_empty_lafs_keep_the_batch_dimension[cpu-float32]          - assert torch.Size([0, 0, 2, 3]) == (1, 0, 2, 3)
  ```
- The `fginn`/`adalam` exclusion from the `confidence` docstring is correct, checked at the call site rather than taken on trust: `integrated.py:476` is `dists, idxs = self.matcher(descs0[batch_idx], descs1[batch_idx])` — two arguments, so `GeometryAwareDescriptorMatcher`'s four-argument `forward` cannot be reached from here. `confidence = 1.0 - dists` at line 485, so the stated range is right.
- AGENTS.md pre-merge grep: `grep -rnE "#4065|issues/4065" kornia/ tests/` returns **nothing in `kornia/`**; three hits in `tests/`, all pins.
- The grayscale conversion is correctly hoisted above the early return, so the probe is built with the channel count the descriptor will actually see.

**Optional ask: the rank change belongs in `### Breaking changes` too**

`lafs0` / `lafs1` going from `(0, 0, 2, 3)` to `(1, 0, 2, 3)` is a silent contract change on a public return value: a downstream `out["lafs0"].shape[0] == 0` no-match check does not raise, it just starts answering wrong. `## Unreleased` already carries a `### Breaking changes` section, and its only current entry (the `normalize_homography` shape guards, #3999) is the same class of change at a lower blast radius — that one at least raises.

The Bug-fixes entry describes the change correctly and names the replacement signal, so nothing is missing; this is only about where a reader scanning for breakage will look. A one-line cross-reference bullet under `### Breaking changes` pointing at the Bug-fixes entry would cover it, and the descriptor-probe widening (a third-party descriptor that raises on a zero patch now raises here) could travel with it.

Non-blocking — merge as-is if you would rather keep the entry in one piece.

All three append to the same anchor in `CHANGELOG.md`. `git merge-tree --write-tree` reports a conflict for **every pair**, so merging any one makes the other two conflicting. A conflicting PR runs no `pull_request` workflow and `gh pr checks` then keeps reporting the last green run from *before* the conflict — so rebase and let CI go green again between merges rather than trusting the stale check. #4084 and #4085 conflict with nothing.

ducha-aiki (commented):
Final review at rebased head 900ccd04: no actionable findings. The branch was rebased onto current main, the sole CHANGELOG.md conflict was resolved by retaining all concurrent entries, and no functional code conflict existed. The empty descriptor path now matches the non-empty flattened width/dtype/device contract, empty LAF outputs retain the success-path batch dimension, and the confidence documentation matches DescriptorMatcher behavior.

Validation: local focused module run passed (19 passed, 10 skipped), full pre-commit passed, and fresh hosted CI is fully green (39/39).

Recommendation: merge.

Issue #4041, opened 2026-08-27
DescriptorMatcherWithSteerer doctest runs a full-size model on GPU and OOMs


ducha-aiki:
## Problem

`kornia.feature.matching.DescriptorMatcherWithSteerer` is the **only** docstring example in `kornia/` that moves itself onto an accelerator:

```python
>>> device = K.core.utils.get_cuda_or_mps_device_if_available()
>>> img1 = torch.randn([1, 3, 768, 768], device=device)
>>> img2 = torch.randn([1, 3, 768, 768], device=device)
>>> dedode = KF.DeDoDe.from_pretrained(detector_weights="L-C4-v2", descriptor_weights="B-SO2").to(device)
...
...     kps1, scores1, descs1 = dedode(img1, n=20_000)
```

That makes the outcome of `pixi run doctest` depend on how much GPU memory the machine happens to have free. It runs two 768×768 images through a DINOv2-L-based DeDoDe asking for 20,000 keypoints each — an integration-test workload living in the API reference.

On an Apple Silicon machine with other GPU load present, it fails:

```text
UNEXPECTED EXCEPTION: RuntimeError('MPS backend out of memory
(MPS allocated: 2.62 GiB, other allocations: 7.46 GiB, max allowed: 9.07 GiB).
Tried to allocate 78.25 KiB on private pool.')

File "kornia/feature/matching.py", line 186, in match_snn
    ratio = vals[:, 0] / vals[:, 1]
```

Note where it dies: not in the model, but on a **78 KiB** allocation in `match_snn`, after the example has already filled the device. The failure is a resource-exhaustion artifact, not a defect in the code being documented.

## Reproduction

On `main` at `9e45dddc`, on a machine with MPS available and a few GiB of other GPU allocations live, with the DeDoDe checkpoints already cached:

```bash
pixi run test-module --doctest-modules "kornia/feature/matching.py::kornia.feature.matching.DescriptorMatcherWithSteerer"
```

Environment: kornia 0.9.0rc1 | torch 2.9.1 | python 3.11.14 | OS Darwin arm64 | CUDA/ROCm None (available=False) | MPS=True

On a CPU-only runner the example simply falls back to CPU and passes, which is why CI does not see this.

## Expected

The documentation check should not be contingent on accelerator memory. A docstring example's job is to show how the API is called.

Possible directions:

[...]

PR #4084, opened 2026-08-29
docs: stop three doctests pinning float reprs at a representation boundary (kornia#4040)

### LINKED ISSUE #4040: Doctests pin exact tensor reprs that sit on a float representation boundary
## Problem

Three doctests assert on an exact `torch.Tensor` repr whose printed form sits on a **representation boundary**, so which side it lands on is decided by platform float accumulation rather than by anything the example is demonstrating. They fail on macOS/arm64 with torch 2.9.1 while passing in CI.

These are not download-related; they fail on a fully warm weights cache and are independent of #4005.

## Reproduction

On `main` at `9e45dddc`, with every model checkpoint already cached:

```bash
pixi run doctest
```

Environment: kornia 0.9.0rc1 | torch 2.9.1 | python 3.11.14 | OS Darwin arm64 | CUDA/ROCm None (available=False) | MPS=True

### 1. `kornia.filters.kernels.get_gaussian_erf_kernel1d` — decimal rounding boundary

```text
>>> get_gaussian_erf_kernel1d(3, 2.5)
Expected: tensor([[0.3245, 0.3511, 0.3245]])
Got:      tensor([[0.3244, 0.3511, 0.3244]])
```

The computed value is `0.3244499862`, which rounds to `0.3244`. The docstring's `0.3245` requires the value to be `>= 0.32445`. The true value sits essentially *on* the boundary, so the fourth decimal is decided by the last bit or two of the accumulation — this doctest is fragile by construction, not merely mis-specified.

### 2. `kornia.filters.kernels.get_gaussian_kernel3d` — integral-vs-fractional repr boundary

```text
>>> get_gaussian_kernel3d((3, 3, 3), (1.5, 1.5, 1.5)).sum()
Expected: tensor(1.)
Got:      tensor(1.0000)
```

The sum is `1.0000001192092896`, not exactly `1.0`. torch prints `tensor(1.)` only when every element is exactly integral, so the expected output silently asserts that the kernel sums to **exactly** 1.0 in float32 — a property of the summation order on the machine that wrote the docstring.

### 3. `kornia.geometry.liegroup.se2.Se2.log` — the same boundary, in the other direction

```text
>>> v = torch.ones((1, 3))
>>> s = Se2.exp(v).log()
>>> s
Expected: tensor([[1.0000, 1.0000, 1.0000]], grad_fn=<StackBackward0>)
Got:      tensor([[1., 1., 1.]], grad_fn=<StackBackward0>)
```

Here the round trip is exact (values are exactly `1.0`), so torch uses the integral form — but the docstring pins the *fractional* form, i.e. it asserts the round trip is **inexact**. A more accurate result fails the doctest.

Cases 2 and 3 are the same mechanism with opposite signs, which is a good illustration of why the repr is the wrong thing to assert on.

## Expected

Docstring examples should demonstrate the API, not pin platform-specific float formatting. Cases 2 and 3 in particular currently assert the *opposite* exactness properties of each other by accident.

Possible directions, roughly in order of preference:

- assert the property instead of the repr, e.g. `>>> torch.allclose(get_gaussian_kernel3d(...).sum(), torch.tensor(1.0))` → `True`, and similarly `>>> torch.allclose(s, torch.ones(1, 3))` → `True` for `Se2.log`;
- for case 1, pick example arguments whose output is not within a ULP of a `.00005` decimal boundary, so the printed value is stable;
- or, if the exact repr is genuinely the point of the example, add an explicit tolerance-bearing assertion around it rather than relying on `repr`.

The same audit is worth doing for any other doctest that prints a raw tensor of derived floats.

## Scope note

Found while fixing #4005 (doctest weight downloads). None of these modules are touched by that work, and all three reproduce on a warm cache with no network access, so they are reported separately rather than bundled into that fix.

ducha-aiki:
## What

Fixes the three doctests in #4040 that pin an exact `torch.Tensor` repr whose printed form sits on a **representation boundary** — which side it lands on is decided by platform float accumulation, not by anything the example demonstrates. All three fail on macOS/arm64 with torch 2.9.1 on a warm weights cache while passing in CI.

## The three cases

**1. `get_gaussian_erf_kernel1d(3, 2.5)` — decimal rounding boundary**

The computed coefficient is `0.3244499862`; printing `0.3245` requires `>= 0.32445`. That is **0.5 ULP** from the boundary, so the fourth decimal is decided by the last bit of the accumulation. The example is fragile by construction, not merely mis-specified.

Changed to `sigma=2.0`, whose coefficients are ~500 ULP from any 4-decimal boundary:

```python
>>> get_gaussian_erf_kernel1d(3, 2.0)
tensor([[0.3195, 0.3611, 0.3195]])
```

It still shows what the old one showed — a large sigma relative to the kernel size gives a near-flat kernel. Margins measured on this machine:

| args | min distance to a 4-decimal boundary |
|---|---:|
| `(3, 2.5)` *(old)* | **1.4e-08** (~0.5 ULP) |
| `(3, 2.0)` *(new)* | 1.6e-05 (~500 ULP) |
| `(5, 1.5)` *(untouched)* | 1.2e-05 |

**2. `get_gaussian_kernel3d(...).sum()` → `tensor(1.)`**

torch prints the integral form only when every element is exactly integral, so this silently asserted that the kernel sums to **exactly** 1.0 in float32. The sum is `1.0000001192092896`. Now asserts the property:

```python
>>> torch.allclose(get_gaussian_kernel3d((3, 3, 3), (1.5, 1.5, 1.5)).sum(), torch.tensor(1.0))
True
```

**3. `Se2.exp(v).log()` → `tensor([[1.0000, 1.0000, 1.0000]], ...)`**

The same mechanism with the opposite sign: here the round trip is exact, so torch uses the integral form, while the docstring pinned the *fractional* one — i.e. it asserted the round trip is **inexact**, and a more accurate result fails the doctest. Cases 2 and 3 asserted opposite exactness properties of each other by accident. Now:

```python
>>> s.shape
torch.Size([1, 3])
>>> torch.allclose(s, v)
True
>>> s.requires_grad
True
```

The `requires_grad` line replaces what the old `grad_fn=<StackBackward0>` in the expected output carried: the docstring's one demonstration that the round trip is differentiable. It is a bool, so it is not repr-fragile.

## Audit of the rest of the doctests

The issue asks for the same audit on any other doctest printing derived floats, so I ran one: patch `torch.Tensor.__repr__` for the duration of `pytest --doctest-modules kornia/` and measure every printed value's distance to a 4-decimal rounding boundary and to an integer. **Nothing else currently fails**, and I have not changed anything else, but the run surfaces a latent tail worth recording:

- `enhance/zca.py::zca_whiten` — `-0.5773` is **9.7 ULP** below the `0.57735` boundary and comes out of an eigendecomposition, so a different LAPACK could print `-0.5774`. The likeliest of the tail to break somewhere.
- integral-vs-fractional flips 1 ULP from exact (the case 2/3 mechanism): `crop_by_boxes` (`5.0000005`), `center_crop3d` and `crop_and_resize3d` (`20.999998`), `spatial_soft_argmax2d` (`0.99999994`), `psnr` / `psnr_loss` (`±19.999998`), `ARKitQTVecs_to_ColmapQTVecs`.
- `project_points` / `PinholeCamera.project` (2 ULP) and `Se2.hat` / `So2.hat` (`1.5707`) are false alarms — an elementwise IEEE-exact chain and a `3.1415/2` literal respectively, so both are bit-identical on every platform.

Rewriting those live examples into `allclose` checks costs real documentation value for a risk that has not materialised, so they are left as they are. Since merging this closes #4040, that tail is now recorded in **#4086** with the per-case ULP figures and the audit method, so it does not die with this PR description; if one of them ever fails on a new platform, that issue is the place to start.

## Verification

```console
656 passed, 1 failed, 2 skipped     # the failure is #4041, fixed by #4085; 657 pass with both applied
                                    # the 2 skips are uncached-weight examples

$ pytest --doctest-modules kornia/filters/kernels.py kornia/geometry/liegroup/se2.py
22 passed

$ pre-commit run --all-files
all hooks Passed
```

Docstring-only change; no library behavior is touched.

Closes #4040

### DIFF
--- a/kornia/filters/kernels.py
+++ b/kornia/filters/kernels.py
@@ -644,8 +644,8 @@ def get_gaussian_erf_kernel1d(
         1D tensor with gaussian filter coefficients. Shape :math:`(B, \text{kernel_size})`
 
     Examples:
-        >>> get_gaussian_erf_kernel1d(3, 2.5)
-        tensor([[0.3245, 0.3511, 0.3245]])
+        >>> get_gaussian_erf_kernel1d(3, 2.0)
+        tensor([[0.3195, 0.3611, 0.3195]])
         >>> get_gaussian_erf_kernel1d(5, 1.5)
         tensor([[0.1226, 0.2331, 0.2887, 0.2331, 0.1226]])
         >>> get_gaussian_erf_kernel1d(5, torch.tensor([[1.5], [2.1]]))
@@ -751,8 +751,8 @@ def get_gaussian_kernel3d(
                  [[0.0292, 0.0364, 0.0292],
                   [0.0364, 0.0455, 0.0364],
                   [0.0292, 0.0364, 0.0292]]]])
-        >>> get_gaussian_kernel3d((3, 3, 3), (1.5, 1.5, 1.5)).sum()
-        tensor(1.)
+        >>> torch.allclose(get_gaussian_kernel3d((3, 3, 3), (1.5, 1.5, 1.5)).sum(), torch.tensor(1.0))
+        True
         >>> get_gaussian_kernel3d((3, 3, 3), (1.5, 1.5, 1.5)).shape
         torch.Size([1, 3, 3, 3])
         >>> get_gaussian_kernel3d((3, 7, 5), torch.tensor([[1.5, 1.5, 1.5]])).shape
--- a/kornia/geometry/liegroup/se2.py
+++ b/kornia/geometry/liegroup/se2.py
@@ -200,8 +200,10 @@ class Se2(nn.Module):
         Example:
             >>> v = torch.ones((1, 3))
             >>> s = Se2.exp(v).log()
-            >>> s
-            tensor([[1.0000, 1.0000, 1.0000]], grad_fn=<StackBackward0>)
+            >>> s.shape
+            torch.Size([1, 3])
+            >>> torch.allclose(s, v)
+            True
 
         """
         theta = self.so2.log()
--- a/kornia/geometry/liegroup/se2.py
+++ b/kornia/geometry/liegroup/se2.py
@@ -204,6 +204,8 @@ class Se2(nn.Module):
             torch.Size([1, 3])
             >>> torch.allclose(s, v)
             True
+            >>> s.requires_grad
+            True
 
         """
         theta = self.so2.log()


ducha-aiki:
## Deep review — verdict: **correct; approve.** One thing to settle before this closes #4040.

First review of this PR. I re-derived every numeric claim rather than reading the body, on macOS/arm64, torch 2.9.1, python 3.11.14, at `cee3f0a6`.

### All three failures reproduce, and all three fixes hold

| case | on this machine | after |
|---|---|---|
| `get_gaussian_erf_kernel1d(3, 2.5)` | prints `0.3244`, docstring says `0.3245` | `(3, 2.0)` → `tensor([[0.3195, 0.3611, 0.3195]])`, matches |
| `get_gaussian_kernel3d(...).sum()` | prints `tensor(1.0000)`, docstring says `tensor(1.)` | `allclose` → `True` |
| `Se2.exp(v).log()` | prints `tensor([[1., 1., 1.]], grad_fn=...)`, docstring says `1.0000` | `.shape` + `allclose` → `True` |

The "cases 2 and 3 assert opposite exactness properties" framing is exactly right: `sum()` is `0x1.000002p+0` (not integral), while the Se2 round trip is bit-exact (`(s == v).all()` is `True`, max abs err `0.0`), so one printed the fractional form where the docstring wanted integral and the other the reverse.

**Margins re-measured** in float32 ULP, not taken from the table:

| args | distance to the nearest 4-decimal boundary | in float32 ULP |
|---|--:|--:|
| `(3, 2.5)` *(old)* | 1.378e-08 | **0.46** |
| `(3, 2.0)` *(new)* | 1.576e-05 | **529** |
| `(5, 1.5)` *(untouched)* | 1.230e-05 | 412 |

So the old example really was decided by the last bit, and the new one has three orders of magnitude of headroom — ample against the few-ULP spread you get from a different `erf` kernel. Case 3's `allclose` has infinite headroom here (exact), and float64 is exact too.

`pytest --doctest-modules kornia/filters/kernels.py kornia/geometry/liegroup/se2.py` → 22 passed. The full `pytest --doctest-modules kornia/` gives 656 passed / 1 failed / 2 skipped on my machine, the one failure being `DescriptorMatcherWithSteerer` — i.e. #4041, which #4085 fixes; nothing to do with this PR. The `657 passed` in the body is only reachable with both branches applied, worth a word in the body.

The AGENTS.md pre-merge grep is clean: `grep -rnE "#4040|issues/4040" kornia/ tests/` returns nothing, so there is no stale prose or pin to retire.

### The one ask: the audit tail should not die with the PR description

#4040's *Expected* section asks for the audit as part of the fix — *"The same audit is worth doing for any other doctest that prints a raw tensor of derived floats."* The audit was done and it found something real, but it lives only in this PR body, and merging this closes the issue.

I reproduced the strongest item:

```
kornia/enhance/zca.py::zca_whiten
  -0.5773494243621826   distance to the 0.57735 boundary = 5.756e-07  =  9.7 float32 ULP
```

Nine ULP, out of a `linalg.eigh`, is a genuinely different risk class from the 400–800 ULP everything else in that output sits at — a different LAPACK prints `-0.5774` and the doctest fails, with no record anywhere of why. Same for the 1-ULP integral-vs-fractional flips (`crop_by_boxes` `5.0000005`, `psnr` `±19.999998`, `spatial_soft_argmax2d` `0.99999994`), which are literally the case-2/case-3 mechanism still latent.

I agree with the decision **not** to rewrite those now — the doc value is real and the risk hasn't materialised. But please either open a follow-up issue carrying that list and reference it here, or leave #4040 open with the tail as its remaining scope. Closing #4040 on merge is the only version of this I'd object to.

### Nit

- **Se2 loses the differentiability signal.** The old expected output carried `grad_fn=<StackBackward0>`, which was the one place the `log()` docstring showed the round trip is differentiable. `>>> s.requires_grad` → `True` costs one line and is not repr-fragile.
- The title says "assert properties instead of float reprs in three doctests", but case 1 still asserts a repr — it changes the *input* so the repr is stable. Both are legitimate (the issue lists them as separate directions); the title just describes two of the three.

Nothing here blocks merge once the tail has a home.

ducha-aiki:
Both points addressed.

**The audit tail now has a home: #4086.** It carries the per-case ULP figures, both mechanisms separated (4-decimal boundary vs. integral-vs-fractional formatting), the dismissed false positives with the reason each is safe, and the audit method so it can be re-run without re-deriving it. The PR body points at it, so closing #4040 on merge no longer loses the list.

I re-measured every entry before filing rather than copying the table:

| doctest | value | distance |
|---|---|---:|
| `zca_whiten` | `-0.5773494243621826` | **9.7 ULP** to the `0.57735` boundary |
| `crop_by_boxes` | `5.000000476837158` | 1.0 ULP to `5` |
| `center_crop3d` / `crop_and_resize3d` | `20.999998092651367` | 1.0 ULP to `21` |
| `spatial_soft_argmax2d` | `0.9999999403953552` | 1.0 ULP to `1` |
| `psnr` / `psnr_loss` | `19.999998092651367` | 1.0 ULP to `20` |

(The first ULP figures I quoted came out of `math.ulp` on a Python float, which is the *double* ULP; the issue uses `numpy.spacing(numpy.float32(v))`. The 9.7 and the 1.0s are the float32 numbers.)

**Se2 differentiability signal restored** in `7e41dcb7`:

```python
>>> s.requires_grad
True
```

That is what the old `grad_fn=<StackBackward0>` carried, and a bool has no repr fragility. `pytest --doctest-modules kornia/filters/kernels.py kornia/geometry/liegroup/se2.py` → 22 passed.

**On the title nit:** retitled to "stop three doctests pinning float reprs at a representation boundary", which covers both directions — case 1 moves the input off the boundary, cases 2 and 3 assert the property instead. The body's `657 passed` line is also corrected: this branch alone gives 656 passed / 1 failed, the failure being #4041, and 657 needs #4085 too.

ducha-aiki:
## Final review — approve, no asks

Re-reviewed the full diff at `7e41dcb7` (the state after the `s.requires_grad` line was added back). Verified on macOS/arm64, torch 2.9.1, CPU.

**The two changed examples fail on `main` today**

Not a hypothetical hardening. Checking out `origin/main`'s `kornia/filters/kernels.py` and running its doctests on this machine:

```
Expected:  tensor([[0.3245, 0.3511, 0.3245]])
Got:       tensor([[0.3244, 0.3511, 0.3244]])
Expected:  tensor(1.)
Got:       tensor(1.0000)
2 failed, 8 passed
```

**Independent ULP measurement reproduces the fragility argument**

Distance from each printed value to the nearest 4-decimal rounding boundary, in float32 ULPs:

| example | printed | margin |
|---|---|---:|
| `get_gaussian_erf_kernel1d(3, 2.5)` — removed | 0.3244 / 0.3511 | **≈1** / 1677 |
| `get_gaussian_erf_kernel1d(3, 2.0)` — new | 0.3195 / 0.3611 | 528 / 621 |
| `get_gaussian_erf_kernel1d(5, 1.5)` — untouched | 0.1226 / 0.2331 / 0.2887 | 4207 / 826 / 1251 |

So the replacement sigma is genuinely off the boundary, and leaving `(5, 1.5)` alone was the right call rather than an omission.

**The `Se2` replacement carries the same information as the old repr**

```
Se2.exp(torch.ones((1, 3))).log() -> tensor([[1., 1., 1.]], grad_fn=<StackBackward0>)
  requires_grad: True     grad_fn: <StackBackward0>     allclose(s, v): True
  v.requires_grad: False
```

`v` does not require grad, so the differentiability comes from inside the round trip — which is exactly what the old `grad_fn=` in the expected output demonstrated, now asserted as a bool instead of as a repr fragment.

**Other checks**

- `pytest --doctest-modules kornia/filters/kernels.py kornia/geometry/liegroup/se2.py` → **22 passed**.
- AGENTS.md pre-merge grep: `grep -rnE "#4040|issues/4040" kornia/ tests/` returns **nothing** in either tree, so closing #4040 leaves no stale prose behind and there is no pin to re-check.
- The audit tail is parked in #4086 and referenced from the description, so it survives this PR being closed.

Conflicts with none of the other open PRs in this batch (`git merge-tree` clean against #4074, #4078, #4080, #4085), so it can merge in any order.

PR #4085, opened 2026-08-29
docs(feature): make the DescriptorMatcherWithSteerer example a small CPU one (kornia#4041)

### LINKED ISSUE #4041: DescriptorMatcherWithSteerer doctest runs a full-size model on GPU and OOMs

ducha-aiki:
## What

Fixes #4041. `DescriptorMatcherWithSteerer` held the only docstring example in `kornia/` that moved itself onto an accelerator, and what it moved there was an integration-test workload: two 768×768 images through a DeDoDe asking for **20,000 keypoints each**. That made the outcome of `pixi run doctest` depend on how much GPU memory the machine happened to have free.

On an Apple Silicon machine with other GPU load present it dies on a **78 KiB** allocation in `match_snn` — after the example has already filled the device — which is a resource-exhaustion artifact, not a defect in the code being documented. A CPU-only runner silently falls back to CPU and passes, which is why CI never saw it.

## The change

The live example is now small and on CPU, with the device selection kept as commented-out lines so a reader still sees how to run the real thing:

```python
>>> import kornia.feature as KF
>>> # import kornia as K
>>> # device = K.core.utils.get_cuda_or_mps_device_if_available()
>>> img1 = torch.randn([1, 3, 128, 128])
>>> img2 = torch.randn([1, 3, 128, 128])
>>> # img1, img2 = img1.to(device), img2.to(device)
>>> dedode = KF.DeDoDe.from_pretrained(detector_weights="L-C4-v2", descriptor_weights="B-SO2")
>>> # dedode = dedode.to(device)
```

- 768² → 128² images, `n=20_000` → `n=1_000`, `subset_size=1000` → `500`
- a line above the block says the example is deliberately small and that a real workload uses full resolution and an accelerator
- `import kornia as K` is commented out alongside the `device` line, its only user, so the two uncomment together
- the two `print(...)` lines were already commented out, so nothing is lost; two checks replace them, so the example verifies its own call sequence without pinning any float:

```python
>>> idxs.shape == (dists.shape[0], 2), 0 <= num_rot < steerer_order
(True, True)
>>> bool((idxs[:, 0] < len(kps1)).all()), bool((idxs[:, 1] < len(kps2)).all())
(True, True)
```

  These tie output to input rather than restating the call's signature: the first catches a matcher returning mismatched distances and indices, the second an index outside the keypoint set it indexes into. Both use `(x < n).all()` instead of `.max()` so they stay valid on an empty match set rather than raising. Measured over 5 random inputs: 164-191 matches, max index 993-999 against 1000 keypoints, `num_rot = 0` throughout.

The full-size version is not moved into `tests/` because the class already has real coverage there — `tests/feature/test_matching.py` exercises `global`/`local` steer modes and `mnn`/`smnn` matching (37 tests).

## Verification

```console
$ pytest --doctest-modules "kornia/feature/matching.py::kornia.feature.matching.DescriptorMatcherWithSteerer"
1 passed in 5.26s                  # was two 768x768 forward passes at 20k keypoints

657 passed, 2 skipped in 48.52s

$ pytest tests/feature/test_matching.py -k "steer or Steer"
37 passed, 63 deselected in 0.52s

$ pre-commit run --files kornia/feature/matching.py
all hooks Passed
```

Run on macOS/arm64, torch 2.9.1, python 3.11.14, with the DeDoDe checkpoints cached. Docstring-only change; no library behavior is touched. After it, `grep -rn ">>> .*get_cuda_or_mps_device_if_available" kornia/` matches only the commented-out line, so no docstring example selects a device any more.

Closes #4041

### DIFF
--- a/kornia/feature/matching.py
+++ b/kornia/feature/matching.py
@@ -394,27 +394,35 @@ class DescriptorMatcherWithSteerer(nn.Module):
         th: threshold on distance ratio, or other quality measure.
 
     Example:
+        The example below is deliberately small so that it runs anywhere. A real workload uses
+        full-resolution images and many more keypoints, and is worth moving onto an accelerator --
+        uncomment the ``device`` lines to do that.
+
         >>> import kornia as K
         >>> import kornia.feature as KF
-        >>> device = K.core.utils.get_cuda_or_mps_device_if_available()
-        >>> img1 = torch.randn([1, 3, 768, 768], device=device)
-        >>> img2 = torch.randn([1, 3, 768, 768], device=device)
-        >>> dedode = KF.DeDoDe.from_pretrained(detector_weights="L-C4-v2", descriptor_weights="B-SO2").to(device)
+        >>> # device = K.core.utils.get_cuda_or_mps_device_if_available()
+        >>> img1 = torch.randn([1, 3, 128, 128])
+        >>> img2 = torch.randn([1, 3, 128, 128])
+        >>> # img1, img2 = img1.to(device), img2.to(device)
+        >>> dedode = KF.DeDoDe.from_pretrained(detector_weights="L-C4-v2", descriptor_weights="B-SO2")
+        >>> # dedode = dedode.to(device)
         >>> steerer_order = 8  # discretisation order of rotation angles
         >>> steerer = KF.steerers.DiscreteSteerer.create_dedode_default(
         ... generator_type="SO2", steerer_order=steerer_order
         ... )
-        >>> steerer = steerer.to(device)
+        >>> # steerer = steerer.to(device)
         >>> matcher = KF.matching.DescriptorMatcherWithSteerer(
         ... steerer=steerer, steerer_order=steerer_order, steer_mode="global", match_mode="smnn", th=0.98
         ... )
         >>> with torch.inference_mode():
-        ...     kps1, scores1, descs1 = dedode(img1, n=20_000)
-        ...     kps2, scores2, descs2 = dedode(img2, n=20_000)
+        ...     kps1, scores1, descs1 = dedode(img1, n=1_000)
+        ...     kps2, scores2, descs2 = dedode(img2, n=1_000)
         ...     kps1, kps2, descs1, descs2 = kps1[0], kps2[0], descs1[0], descs2[0]
         ...     dists, idxs, num_rot = matcher(
-        ...         descs1, descs2, normalize=True, subset_size=1000,
+        ...         descs1, descs2, normalize=True, subset_size=500,
         ...     )
+        >>> idxs.shape[1], 0 <= num_rot < steerer_order
+        (2, True)
         >>> # print(f"{idxs.shape[0]} tentative matches with steered DeDoDe")
         >>> # print(f"at rotation of {num_rot * 360 / steerer_order} degrees")
 
--- a/kornia/feature/matching.py
+++ b/kornia/feature/matching.py
@@ -398,8 +398,8 @@ class DescriptorMatcherWithSteerer(nn.Module):
         full-resolution images and many more keypoints, and is worth moving onto an accelerator --
         uncomment the ``device`` lines to do that.
 
-        >>> import kornia as K
         >>> import kornia.feature as KF
+        >>> # import kornia as K
         >>> # device = K.core.utils.get_cuda_or_mps_device_if_available()
         >>> img1 = torch.randn([1, 3, 128, 128])
         >>> img2 = torch.randn([1, 3, 128, 128])
@@ -421,8 +421,10 @@ class DescriptorMatcherWithSteerer(nn.Module):
         ...     dists, idxs, num_rot = matcher(
         ...         descs1, descs2, normalize=True, subset_size=500,
         ...     )
-        >>> idxs.shape[1], 0 <= num_rot < steerer_order
-        (2, True)
+        >>> idxs.shape == (dists.shape[0], 2), 0 <= num_rot < steerer_order
+        (True, True)
+        >>> bool((idxs[:, 0] < len(kps1)).all()), bool((idxs[:, 1] < len(kps2)).all())
+        (True, True)
         >>> # print(f"{idxs.shape[0]} tentative matches with steered DeDoDe")
         >>> # print(f"at rotation of {num_rot * 360 / steerer_order} degrees")
 


ducha-aiki:
## Deep review — verdict: **correct; approve.** Two nits, neither blocking.

First review of this PR. Verified at `4164a1a0` on macOS/arm64, torch 2.9.1, python 3.11.14, MPS available, DeDoDe checkpoints warm.

### The premise reproduces exactly as reported

On `main` the example dies where #4041 says it does — not in the model, on a 78 KiB allocation in `match_snn` after the example has already filled the device:

```
File "kornia/feature/matching.py", line 186, in match_snn
    ratio = vals[:, 0] / vals[:, 1]
RuntimeError: MPS backend out of memory (MPS allocated: 2.62 GiB, other allocations: 7.46 GiB,
max allowed: 9.07 GiB). Tried to allocate 78.25 KiB on private pool.
```

On this branch: `1 passed in 5.18s`.

### The new example is stable, not merely passing

The thing worth checking in a doctest with no seed is whether the expected output can move. It cannot:

- **`num_rot` is a plain Python `int`**, not a 0-d tensor, so `(2, True)` is the correct rendering — a tensor would have printed `(2, tensor(True))`. Checked directly.
- **`idxs.shape[1]` is 2** by `match_smnn`'s contract, and `num_rot` is an argmax over `range(steerer_order)`, so both halves are structurally true.
- Ran the full sequence over **6 independent random inputs**: `kps1` is exactly `(1000, 2)` every time (a 128×128 image has 16,384 candidate positions, so `n=1_000` never underfills and `subset_size=500` is always satisfiable), `idxs` came out between 159 and 202 rows, `num_rot` was 0 throughout, and the printed tuple was `(2, True)` in all six.

So there is no flake surface here. `pre-commit run --files kornia/feature/matching.py` passes, and `grep -rn ">>> .*get_cuda_or_mps_device_if_available" kornia/` now matches only line 403, the commented-out one — the body's claim holds.

The cold-cache behaviour is unchanged, which matters: `from_pretrained` still routes through `torch.hub.download_url_to_file`, so `testing/doctest_downloads.py` still turns this example into a skip on a cache miss and only runs it for real when the weights are already there. This PR doesn't widen what `pixi run doctest` pulls.

### Nits

1. **The new check is tautological.** `idxs.shape[1] == 2` and `0 <= num_rot < steerer_order` are both true by construction — they can only fail if the call raises, which already fails the doctest. If the line is meant to earn its place, something tying output to input would: `dists.shape[0] == idxs.shape[0]` and `int(idxs.max()) < 1_000` would at least catch a matcher returning mismatched or out-of-range indices. Fine to leave as is; just don't read it as coverage.

2. **`>>> import kornia as K` is now dead in the live example** — `K` appears only inside the commented `device = ...` line. Defensible, since uncommenting needs it, but worth a beat of thought about whether the import should be commented out alongside it.

Neither changes the outcome. The size choice (128², `n=1_000`) is a degenerate operating point — 6% of pixels become keypoints — but the prose says so plainly, and the class keeps its real coverage in `tests/feature/test_matching.py`.

ducha-aiki:
Both nits applied in `a11e65bc` — the first one was worth taking.

**1. The check now ties output to input.** Replaced:

```python
>>> idxs.shape == (dists.shape[0], 2), 0 <= num_rot < steerer_order
(True, True)
>>> bool((idxs[:, 0] < len(kps1)).all()), bool((idxs[:, 1] < len(kps2)).all())
(True, True)
```

`idxs.shape == (dists.shape[0], 2)` catches a matcher returning mismatched distances and indices — the old `idxs.shape[1] == 2` could not. The bound checks catch an index outside the keypoint set it indexes into; they use the `(x < n).all()` form rather than the suggested `.max()`, because `.max()` raises on an empty tensor and an empty match set would then turn a rare no-match run into a doctest error rather than a passing vacuous check.

Measured over 5 fresh random inputs: 164–191 matches, `dists` rows matching `idxs` rows every time, max index 993–999 against 1000 keypoints, `num_rot = 0` throughout. So the bound is close enough to 1000 to be a real constraint rather than a formality.

I kept `0 <= num_rot < steerer_order` even though it is structurally true — it is the only place the return contract of the class's distinctive third output is written down, and dropping it would leave `num_rot` referenced only from the commented `print` lines.

**2. `import kornia as K` is commented out** next to the `device` line that is its only user, so the two uncomment together.

`pytest --doctest-modules kornia/feature/matching.py` → 1 passed; `pixi run pre-commit-all` clean. PR body updated to match.

ducha-aiki:
## Final review — approve, with one follow-up for a separate PR

Re-reviewed the full diff at `a11e65bc` (the state after the tautological shape check was replaced). Verified on macOS/arm64, torch 2.9.1, CPU, with the DeDoDe checkpoints cached.

**Checks run**

- `KORNIA_DOCTEST_DOWNLOAD=1 pytest --doctest-modules kornia/feature/matching.py` → **1 passed in 5.26 s**, on CPU.
- The new assertions are **not vacuous** — this is the part worth confirming, since an example that asserts nothing is the failure mode #4066 was about. Ran the example body directly:
  ```
  kps1 (1000, 2)   descs1 (1000, 256)
  dists (172, 1)   idxs (172, 2)   num_rot 0 <class 'int'>
  matches = 172    max index 982 of 1000
  ```
  172 real matches, so both index-bound checks are exercised against a non-empty set, and the `(x < n).all()` form still holds if a future input yields none.
- `num_rot` is a Python `int`, so `(True, True)` is the correct expected repr — a tensor would have printed `(True, tensor(True))`.
- AGENTS.md pre-merge grep: `grep -rnE "#4041|issues/4041" kornia/ tests/` returns nothing in either tree.
- Conflicts with none of the other open PRs in this batch, so it can merge in any order.

**Follow-up (pre-existing, not a blocker)**

Running the example on CPU surfaces a warning that was invisible while it selected an accelerator:

```
UserWarning: User provided device_type of 'cuda', but CUDA is not available. Disabling
```

It comes from `kornia/feature/dedode/encoder.py:69` and `kornia/feature/dedode/decoder.py:171`, which enter `torch.autocast("cuda", enabled=self.amp, dtype=self.amp_dtype)` unconditionally; the context manager warns on a CUDA-less machine even when `enabled=False`. The inline comment there says AMP is intentionally scoped to CUDA, which is fine — the missing part is not constructing the context manager at all when CUDA is absent.

This PR does not cause it, but it does change who sees it: every reader who copies the now-CPU-default example, and the post-merge `doctest-weights` run. Worth a small separate PR guarding on `torch.cuda.is_available()`. Happy to open one if you want it.

ducha-aiki (commented):
Final review at a11e65bc: no actionable findings. The CPU-sized doctest preserves the documented workflow, validates output cardinality and index bounds, avoids accelerator-memory dependence, and has a fully green cross-platform CI matrix. The pre-existing CUDA-autocast warning on CPU is not introduced by this change and should be handled separately.

Recommendation: merge.

PR #4088, opened 2026-08-29
fix(tests): remove hardcoded tolerance overrides in morphology tests

### LINKED ISSUE #4081: kornia.morphology is marked ✅ for half precision but 6 bfloat16 tests fail: they override the harness tolerances with a tighter constant

khushalcodess:
Fixes #4081

Six morphology tests (`top_hat`, `bottom_hat`, `gradient`) were failing on
`bfloat16` because they passed explicit `atol=1e-3, rtol=1e-3` to
`assert_close`, overriding the harness's dtype-aware defaults in
`testing/base.py`'s `_DTYPE_PRECISIONS` table (which allows `7.8e-3` for
bfloat16 — comfortably above the ~3.6e-3 error these composite ops
actually produce).

The same hardcoded override existed at **15 sites across 7 files**
(`bottom_hat`, `closing`, `dilation`, `erosion`, `gradient`, `opening`,
`top_hat`), so all 15 are removed here rather than only the 6 that were
failing — the remaining sites are the same latent defect.

**Test-only change, no library behavior affected.**

Verified locally across all four dtypes:
\`\`\`
pytest tests/morphology --device=cpu --dtype=float16,bfloat16,float32,float64
======================= 187 passed, 40 warnings in 4.36s =======================
\`\`\`

float32 becomes *stricter* as a result, not looser, since its harness
default `(1e-4, 1e-5)` is tighter than the `1e-3` being removed.

### DIFF
--- a/tests/morphology/test_bottom_hat.py
+++ b/tests/morphology/test_bottom_hat.py
@@ -76,7 +76,7 @@ class TestBottomHat(BaseTester):
         expected = torch.tensor([[0.2, 0.0, 0.5], [0.0, 0.4, 0.0], [0.3, 0.0, 0.6]], device=device, dtype=dtype)[
             None, None, :, :
         ]
-        assert_close(bottom_hat(tensor, kernel), expected, atol=1e-3, rtol=1e-3)
+        assert_close(bottom_hat(tensor, kernel), expected)
 
     def test_structural_element(self, device, dtype):
         tensor = torch.tensor([[0.5, 1.0, 0.3], [0.7, 0.3, 0.8], [0.4, 0.9, 0.2]], device=device, dtype=dtype)[
@@ -91,8 +91,6 @@ class TestBottomHat(BaseTester):
         assert_close(
             bottom_hat(tensor, torch.ones_like(structural_element), structuring_element=structural_element),
             expected,
-            atol=1e-3,
-            rtol=1e-3,
         )
 
     def test_exception(self, device, dtype):
--- a/tests/morphology/test_closing.py
+++ b/tests/morphology/test_closing.py
@@ -91,8 +91,6 @@ class TestClosing(BaseTester):
         assert_close(
             closing(tensor, torch.ones_like(structural_element), structuring_element=structural_element),
             expected,
-            atol=1e-3,
-            rtol=1e-3,
         )
 
     def test_exception(self, device, dtype):
--- a/tests/morphology/test_dilation.py
+++ b/tests/morphology/test_dilation.py
@@ -77,7 +77,7 @@ class TestDilate(BaseTester):
             None, None, :, :
         ]
         assert_close(dilation(tensor, kernel, engine="unfold"), expected, atol=1e-4, rtol=1e-4)
-        assert_close(dilation(tensor, kernel, engine="convolution"), expected, atol=1e-3, rtol=1e-3)
+        assert_close(dilation(tensor, kernel, engine="convolution"), expected)
 
     def test_structural_element(self, device, dtype):
         tensor = torch.tensor([[0.5, 1.0, 0.3], [0.7, 0.3, 0.8], [0.4, 0.9, 0.2]], device=device, dtype=dtype)[
@@ -94,8 +94,6 @@ class TestDilate(BaseTester):
                 tensor, torch.ones_like(structural_element), structuring_element=structural_element, engine="unfold"
             ),
             expected,
-            atol=1e-3,
-            rtol=1e-3,
         )
         assert_close(
             dilation(
@@ -105,8 +103,6 @@ class TestDilate(BaseTester):
                 engine="convolution",
             ),
             expected,
-            atol=1e-3,
-            rtol=1e-3,
         )
 
     def test_flip(self, device, dtype):
@@ -117,7 +113,7 @@ class TestDilate(BaseTester):
         expected = torch.tensor([[0.7, 1.0, 1.0], [0.7, 1.0, 1.0], [0.7, 0.9, 0.9]], device=device, dtype=dtype)[
             None, None, :, :
         ]
-        assert_close(dilation(tensor, kernel), expected, atol=1e-3, rtol=1e-3)
+        assert_close(dilation(tensor, kernel), expected)
 
     def test_exception(self, device, dtype):
         tensor = torch.ones(1, 1, 3, 4, device=device, dtype=dtype)
--- a/tests/morphology/test_dilation.py
+++ b/tests/morphology/test_dilation.py
@@ -77,7 +77,10 @@ class TestDilate(BaseTester):
             None, None, :, :
         ]
         assert_close(dilation(tensor, kernel, engine="unfold"), expected, atol=1e-4, rtol=1e-4)
-        assert_close(dilation(tensor, kernel, engine="convolution"), expected)
+        # The convolution engine measures ~3.9e-4 absolute / ~1.5e-3 relative error on macOS's
+        # Apple-backend float32 conv path, above the harness's generic float32 default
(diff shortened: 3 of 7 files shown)


fallenmi (changes requested):
The `bfloat16` issue is fixed, but removing all explicit tolerances makes the supported macOS `float32` matrix fail in five changed assertions.

On exact comparison base `9abf06d926f69439110cd8a6a01b2fe2680ad689`, the seven changed morphology files pass `52/52` with `--device=cpu --dtype=float32` on this Apple CPU. Exact head `61c6d1ff0dbd39449f8f9f58fd319e9cb06bbd80` and GitHub's current merge `91e29a805aa93c721ebcdf3444df7e80f7c11e16` both fail the same five convolution assertions (`dilation` kernel/structuring-element and `erosion` kernel/structuring-element/flip), with `47 passed, 5 failed`. The observed maximum absolute error is about `3.91e-4` and maximum relative error about `1.46e-3`, above the generic float32 defaults (`atol=1e-5`, `rtol=1e-4`). All five exact-head macOS float32 jobs reproduce these failures; Linux and Windows passing does not make the Apple backend tolerance valid.

I also verified the intended improvement: the same focused `bfloat16` run changes from `46 passed, 6 failed` on base to `52 passed` on head. Please keep the dtype-aware `bfloat16` behavior while retaining an appropriate explicit tolerance for the affected float32 convolution path (or fix its numerical error), and add a matrix regression that covers macOS float32 as well as bfloat16.

`git diff --check` is clean.

AI disclosure: I used OpenAI Codex to inspect the exact diff and CI matrix, run the base/head/current-merge dtype-focused suites, measure the failing errors, and draft this review. I verified the outputs, exact commits, policy, and interaction history.

khushalcodess:
Thanks for catching this — good find on the macOS-specific convolution error.

I've pushed a follow-up commit that restores an explicit tolerance (atol=1e-3, rtol=1e-3) scoped only to the 5 engine="convolution" call sites in dilation/erosion (2 in test_dilation.py, 3 in test_erosion.py), matching the exact assertions you flagged. Added a comment at each site explaining why, referencing the measured error (~3.9e-4 absolute / ~1.5e-3 relative) so it's clear this isn't an arbitrary number.

Everything else keeps the dtype-aware harness defaults — the bfloat16 fix for top_hat/bottom_hat/gradient/closing/opening is untouched, and the engine="unfold" paths (which already had their own explicit 1e-4 tolerance and aren't affected by this) are also untouched.

Re: the matrix regression — since these assertions already run through the existing dtype fixture parametrization, and CI already runs a macOS float32 job (which is how this got caught), this scoped fix should be exercised automatically going forward. Let me know if you think a more targeted test is still needed beyond that.

Ran the full suite locally (Windows/CPU) across all four dtypes: 187 passed. Let me know if there's anything else you'd like adjusted.

ducha-aiki (approved):
**Review of `2e39d9b4` — APPROVE.**

@fallenmi's finding was real and this head fixes it the right way: the override survives only at the five `engine="convolution"` call sites in `test_dilation.py`/`test_erosion.py`, with the measured error written next to each, and every other site falls back to the harness `_DTYPE_PRECISIONS` table.

Checked on an Apple-silicon Mac, which is the platform that failed (`.venv`, torch 2.9.1):

```
main   tests/morphology --device=cpu --dtype=float32,bfloat16               6 failed, 91 passed   (the #4081 bf16 failures)
head   tests/morphology --device=cpu --dtype=float16,bfloat16,float32,float64   187 passed
```

So the bfloat16 defect is gone and the macOS float32 regression from the first push is gone too. Test-only change, no library code touched. Closes #4081.

@fallenmi — thanks for the measured review; that is exactly the kind of check this queue needs. Your CHANGES_REQUESTED still stands formally; if you agree the follow-up addresses it, please dismiss or re-review so the merge button clears.

PR #4105, opened 2026-08-30
fix(color): correct integer color conversion kernels

### LINKED ISSUE #4053: color: rgb_to_yuv/yuv_to_rgb silently return zeros for integer input, and _apply_linear_transformation exists in two diverged copies
### Describe the bug

`rgb_to_yuv` and `yuv_to_rgb` **silently return all zeros** for integer input, and raise an opaque overflow error for `uint8`. `rgb_to_xyz` / `xyz_to_rgb`, which are supposed to share the same helper, do neither.

Both functions build their kernel with `dtype=image.dtype`:

https://github.com/kornia/kornia/blob/main/kornia/color/yuv.py#L56-L64

so for an integer image the BT.470-5 coefficients are truncated to integers at construction time — `0.299` becomes `0`, and `-0.147` cannot be represented as `uint8` at all. The `# Handle Integer inputs by casting to float safely` block in `_apply_linear_transformation` never gets a chance to help, because the damage is already done in the caller.

```python
>>> import torch, kornia
>>> x = torch.full((1, 3, 2, 2), 100, dtype=torch.int64)
>>> kornia.color.rgb_to_yuv(x).flatten()[:3]
tensor([0, 0, 0])                       # every pixel, every channel
>>> kornia.color.rgb_to_yuv(x.float()).flatten()[:3]
tensor([100., 100., 100.])              # what it should be tracking

>>> kornia.color.rgb_to_yuv((torch.rand(1, 3, 2, 2) * 255).to(torch.uint8))
RuntimeError: value cannot be converted to type uint8 without overflow
```

Full matrix — same on cpu and mps, so it is not a device-dispatch problem:

| input | `rgb_to_yuv` cpu | `rgb_to_yuv` mps | `rgb_to_xyz` cpu/mps |
|---|---|---|---|
| `uint8` | `RuntimeError` | `RuntimeError` | `float32`, correct |
| `int32` | `int32`, **all zeros** | `float32`, **all zeros** | `float32`, correct |
| `int64` | `int64`, **all zeros** | `float32`, **all zeros** | `float32`, correct |

The silent-zeros rows are the reason this is filed as a bug rather than a nit: a caller who forgets a `.float()` gets a plausible-looking tensor of the right shape and dtype with no warning.

### Why xyz is fine and yuv is not: two diverged copies of the same helper

There are two copies of `_apply_linear_transformation`:

* `kornia/color/xyz.py:95` — private to `xyz.py`, added by #3482
* `kornia/color/utils.py:24` — added by #3514 for `yuv.py`

They are near-identical, but differ in three ways that all cut against the `utils.py` one:

1. **`xyz.py` hardcodes `dtype=torch.float32` when it builds its kernel; `yuv.py` passes `dtype=image.dtype`.** This alone is the whole defect above.
2. **The CPU branch of the `utils.py` copy ignores the values it just computed.** It builds `image_compute` and `kernel_compute` and then calls `torch.einsum("oi, ...ihw -> ...ohw", kernel, image)` on the *originals*. The `xyz.py` copy passes `image_compute` / `kernel_compute`. Today this is masked — because `yuv.py` builds its kernel at `image.dtype`, `kernel is kernel_compute` for every float input — but it means the integer-safety code is dead on cpu even after fix (1), and any future caller that passes a kernel at a different dtype than the image gets silent dtype promotion on cpu and a correct cast on gpu.
3. `xyz.py` branches on `image_compute.device.type`, `utils.py` on `image.device.type` (equivalent today, but the pair should not be drifting at all).

Whatever is decided below, the two copies should collapse into one.

### Design question before anyone fixes it

The docstrings say *"The image data is assumed to be in the range of :math:`(0, 1)`"* — i.e. float. So making the integer path *work* may be the wrong repair: `rgb_to_xyz(uint8_image)` currently succeeds and returns `95.05` for an input of `100`, silently reinterpreting a 0–255 image under a 0–1 contract. That is arguably a worse outcome than a clear error.

Two candidate fixes:

* **(A) Reject non-floating-point input** with a `KORNIA_CHECK`, in the helper. Honest, matches the documented contract, and gives one good error message instead of three different bad outcomes. It is a behavior change for `rgb_to_xyz` / `xyz_to_rgb`, so it needs a CHANGELOG entry and possibly a deprecation cycle.
* **(B) Make yuv behave like xyz** — build both kernels at float32, keep the upcast. Smallest diff, but it blesses 0–255 input the docs do not describe, and leaves `rgb_to_xyz(uint8)` silently wrong-by-scale.

I lean towards (A), but it is worth a maintainer decision rather than a drive-by fix, because there is currently no convention anywhere in `kornia.color` to appeal to — every function does something different with `uint8`:

```
rgb_to_yuv        -> RuntimeError
rgb_to_hls        -> RuntimeError: result type Float can't be cast to the desired output type Byte
rgb_to_hsv        -> float32
rgb_to_lab        -> float32
rgb_to_xyz        -> float32
rgb_to_grayscale  -> uint8, and it wraps: rgb_to_grayscale(full(200, uint8)) == 56
```

(`rgb_to_grayscale` wrapping modulo 256 is a separate defect; noting it only as evidence that nothing enforces the `(0, 1)` float contract today.)

### Reproduction

```python
import torch, kornia

x = torch.full((1, 3, 2, 2), 100, dtype=torch.int64)
print(kornia.color.rgb_to_yuv(x).flatten()[:3])          # tensor([0, 0, 0])
print(kornia.color.rgb_to_xyz(x).flatten()[:3])          # tensor([95.05, 95.05, 95.05])

kornia.color.rgb_to_yuv(x.to(torch.uint8))               # RuntimeError
```

### Expected behavior

One consistent, documented answer across `kornia.color` for non-floating-point input — either a clear `KORNIA_CHECK` failure or a defined upcast — and in no case a silently zeroed image.

### Environment

```
kornia 0.9.0rc1 (main @ c038fc59) | torch 2.2.2 | python 3.11.14 | OS Darwin arm64 | MPS=True
```

The behavior is pure dtype logic in Python, so it is not PyTorch-version dependent.

### Related

Noticed while reviewing #4046 / #4048, which touch `kornia/color/yuv.py` but not this.

---

Jeevang1-epic:

Fixes the integer-input failures in the YUV color conversions and consolidates the duplicated `_apply_linear_transformation` helper used by YUV and XYZ.

The conversion kernels now choose their dtype from the input when it is floating point and otherwise fall back to `torch.float32`:

```python
dtype = image.dtype if image.is_floating_point() else torch.float32
```

This fixes the broken integer conversion path without introducing a new rejection policy for non-floating-point inputs.

In particular:

- `rgb_to_yuv(int32/int64)` no longer collapses to an all-zero result.
- `yuv_to_rgb(int32/int64)` no longer uses a truncated integer coefficient matrix and is checked against the expected numerical result.
- Integer inputs are computed in `float32` and return `float32` results.
- `uint8` input is now accepted and computed in `float32` on its original 0–255 scale; no normalization or rescaling is introduced.
- YUV `float64` coefficient precision is preserved.
- XYZ kernels are now constructed directly at the input floating dtype, avoiding the previous float32-then-float64 coefficient precision loss.

The two private `_apply_linear_transformation` implementations have also been consolidated. `kornia/color/utils.py` is now the canonical implementation and the duplicate in `kornia/color/xyz.py` has been removed.

While consolidating the helper, the CPU path was corrected to operate on `image_compute` and `kernel_compute` instead of the original operands. This is required when an integer image is promoted to a floating compute dtype.

The helper also:

- preserves bias support;
- aligns the bias to the compute dtype and device;
- dispatches using `image_compute.device.type`;
- keeps the output shape and contiguity behavior unchanged.

A changelog entry and focused regression coverage are included.

The broader `kornia.color` policy for whether non-floating-point inputs should eventually be rejected is intentionally left unchanged by this PR.

The separate `rgb_to_grayscale(uint8)` wrapping issue discussed in #4053 is also out of scope.

Fixes #4053

This follows the implementation scope and dtype behavior agreed on in the maintainer discussion in that issue.

## What did you check?

Focused YUV regression tests:

```text
$ pixi run test-module tests/color/test_yuv.py --device=cpu --dtype=float32,float64 -k 4053
7 passed, 211 deselected
```

Focused XYZ regression tests:

```text
$ pixi run test-module tests/color/test_xyz.py --device=cpu --dtype=float32,float64 -k 4053
6 passed, 44 deselected
```

Canonical helper coverage:

```text
$ pixi run test-module tests/color/test_utils.py --device=cpu
1 passed
```

The added regression coverage checks:

- `rgb_to_yuv` integer numerical correctness;
- `yuv_to_rgb` integer numerical correctness against expected coefficients;
- `rgb_to_xyz` and `xyz_to_rgb` integer behavior;
- `uint8` conversion to `float32` without rescaling the original 0–255 values;
- YUV `float64` coefficient precision;
- XYZ `float64` coefficient precision;
- use of the canonical `_apply_linear_transformation` helper;
- CPU compute operands when the image and kernel start with different dtypes;
- bias dtype/device alignment.

Lint:

```text
$ pixi run lint
ruff check: passed
ruff format: passed
```

Type checking:

```text
$ pixi run typecheck
exit code 0
```

The typecheck still reports the existing `torch.cuda.amp.custom_fwd` deprecation warning in `kornia/feature/lightglue.py`. That file is unrelated to this PR and was not modified.

Focused pre-commit hooks were run on all affected files and passed, including:

```text
fix end of files................Passed
trim trailing whitespace........Passed
check for merge conflicts.......Passed
check for added large files.....Passed
mixed line ending...............Passed
ruff check......................Passed
ruff format.....................Passed
codespell.......................Passed
Add license header..............Passed
```

Whitespace validation:

```text
$ git diff --check
exit code 0
```

The implementation was also exercised against the existing YUV/XYZ coverage and compiler-related paths during development.

CUDA and MPS were not available in my local environment, so accelerator-specific execution is left to CI.

A local PyTorch Inductor run was also blocked by the Windows environment because MSVC `cl.exe` is unavailable; the compiler-free full-graph Dynamo checks completed successfully.

## How did AI help?

I used AI-assisted tooling during the investigation to help trace the dtype behavior, compare the duplicated helper implementations, explore edge cases, and iterate on regression-test coverage.

I reviewed the resulting changes, reproduced the original integer-conversion failures, checked the implementation against the maintainer-confirmed scope in #4053, and reran the focused YUV, XYZ, helper, lint, typecheck, pre-commit, and whitespace checks locally before opening this PR.


Jeevang1-epic:
@ducha-aiki PR #4105 is up for #4053.

I followed the scope from your latest clarification, including the float32 fallback for integer inputs, helper consolidation, and the CPU compute-operand fix. I also added focused regression coverage for the YUV/XYZ integer behavior and float64 precision.

All CI checks are green now. Whenever you get a chance, could you take a look?

Jeevang1-epic:
@ducha-aiki Quick update: the PR picked up another conflict with the latest `main`, so I synced it again and resolved the `CHANGELOG.md` conflict while keeping the #4053 changes intact.

The branch is up to date now and the conflict is cleared. If there’s anything else you’d like me to adjust in the implementation, tests, or scope before merge, just let me know and I’ll take care of it.

Thanks!

ducha-aiki:
## Review of #4105 at `854caa05` — APPROVE

Sorry for the wait on this one. Setup: worktree at the PR head with `kornia/` swapped to the merge-base `dfad8156` for the fail-on-base and A/B passes. `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU + MPS on Apple silicon; `kornia.__file__` asserted inside the worktree in every block.

### The tests discriminate — 8 of them

```
head:  tests/color/{test_yuv,test_xyz,test_utils}.py, f32+f64      271 passed
base:  kornia/ -> dfad8156                                           8 failed, 263 passed
  TestRgbToYuv::test_integer_input_4053[cpu-integer_dtype0]
  TestRgbToYuv::test_integer_input_4053[cpu-integer_dtype1]
  TestRgbToYuv::test_uint8_input_4053[cpu]
  TestYuvToRgb::test_integer_input_4053[cpu-integer_dtype0]
  TestYuvToRgb::test_integer_input_4053[cpu-integer_dtype1]
  TestRgbToXyz::test_float64_kernel_precision_4053
  TestXyzToRgb::test_float64_kernel_precision_4053
  TestApplyLinearTransformation::test_compute_operands[cpu]
```

That is a big improvement on the earlier head, where most of the new coverage passed on base. Every new test now fails without the fix, including the helper-consolidation one — which is the hardest of them to pin and the one I was least sure you could discriminate.

### The float path does not move

The consolidation swaps the CPU einsum equation (`"...chw,oc->...ohw"` → `"oi, ...ihw -> ...ohw"`), which can reassociate, so I checked bit-parity rather than assuming it:

```
('cpu', float32, rgb_to_xyz)   bit-identical=True   max|diff|=0
('cpu', float32, xyz_to_rgb)   bit-identical=True   max|diff|=0
('cpu', float32, rgb_to_yuv)   bit-identical=True   max|diff|=0
('cpu', float32, yuv_to_rgb)   bit-identical=True   max|diff|=0
('mps', float32, rgb_to_xyz)   bit-identical=True   max|diff|=0
('mps', float32, xyz_to_rgb)   bit-identical=True   max|diff|=0
('mps', float32, rgb_to_yuv)   bit-identical=True   max|diff|=0
('mps', float32, yuv_to_rgb)   bit-identical=True   max|diff|=0
('cpu', float64, rgb_to_yuv)   bit-identical=True   max|diff|=0
('cpu', float64, yuv_to_rgb)   bit-identical=True   max|diff|=0
('cpu', float64, rgb_to_xyz)   bit-identical=False  max|diff|=2.42e-08
('cpu', float64, xyz_to_rgb)   bit-identical=False  max|diff|=5.46e-08
```

Both einsum branch and conv2d branch (MPS) are byte-for-byte unchanged at float32. The only movement is float64 XYZ, and it is exactly the float32-rounded-coefficient loss you set out to repair — same class as #3958, and `2.4e-08` / `5.5e-08` is the right magnitude for eight significant digits recovered. YUV float64 is untouched because its kernel was already built at the image dtype.

### The AGENTS gate is clean

You closed the issue *and* retired its pin, which is the part these usually miss:

```
$ grep -rnE "#4053|issues/4053" kornia/ tests/     # at this head
(no output)
```

`test_wart_integer_input_truncates_the_kernel_4053` is deleted rather than left asserting the old truncated kernel — correct, since the PR makes that behaviour unreachable.

### Notes, not blockers

- The CHANGELOG entry is accurate and names the float64 XYZ change explicitly, which is the part a reader most needs. Good.
- `_apply_linear_transformation` aligning `bias` to the compute dtype/device is a real latent fix in its own right — nothing in-tree passes a bias today, so it is dead code until something does, but it would have been an integer-input crash the moment it was used.
- Scope discipline is right: no rejection policy for non-floating input, and `rgb_to_grayscale(uint8)` correctly left to its own issue.

Approving. Thanks for the patience through the two syncs, and for splitting the scope the way the issue thread asked rather than expanding it.

ducha-aiki (dismissed):

ducha-aiki (approved):

PR #4106, opened 2026-08-30
Fix learnable FRN epsilon dtype

Marchematics:
Initialize the learnable Filter Response Normalization epsilon as a floating-point parameter so the public option can be constructed and trained.

Test: python3 -m py_compile kornia/feature/hynet.py tests/feature/test_hynet.py

AI assistance was used.


ducha-aiki (approved):
**APPROVE.**

Reproduced on `main`: `FilterResponseNorm2d(4, is_eps_leanable=True)` raises `RuntimeError: Only Tensors of floating point and complex dtype can require gradients`, so the public learnable-eps option was unconstructible. Initialising from `eps` (rather than the integer literal `1`) is the correct fix and matches the buffer branch. `tests/feature/test_hynet.py`: 5 passed at this head.

**#4108 is a byte-identical duplicate of the production change in this PR** (opened 22 minutes later, with an empty template body). I am approving this one and will close #4108; please open one PR per fix in future.

Issue #4099, opened 2026-08-29
ScaleSpaceDetector raises on any multi-channel image, though it documents (B, C, H, W)


ducha-aiki:
`ScaleSpaceDetector` documents its input as `(B, C, H, W)` and raises on any `C > 1`, with or
without a mask.

## Reproduction

```python
import torch
from kornia.feature import ScaleSpaceDetector

ScaleSpaceDetector(5)(torch.rand(1, 3, 48, 48))
```

```
RuntimeError: view size is not compatible with input tensor's size and stride
(at least one dimension spans across two contiguous subspaces). Use .reshape(...) instead.
```

Same with a mask. Reproduces identically on `main` at 9abf06d9 and on the head of #4098.

## Cause

`kornia/feature/scale_space_detector.py:304-306`

```python
mask_flat = cand_mask.view(B, -1)                        # (B, L*H*W)
resp_flat = response_max.view(B, -1)                     # (B, L*H*W)
coord_flat = coord_max.view(B, 3, -1).permute(0, 2, 1)   # (B, L*H*W, 3)
```

For `C > 1` these tensors are `(B, C, L, H, W)` and not contiguous in the layout `view` needs, so
the first `view` raises. The comments already assume `C = 1` — the shapes they name have no
channel axis.

Note that even if the `view` were replaced by `reshape`, the flattening would pool candidates
across channels without recording which channel each came from, so the semantics need a decision
too, not just a call swap. `MultiResolutionDetector` makes exactly that choice explicitly since
#4098 — see the sibling issue on per-channel duplicates.

## Options

1. Reject `C > 1` in `forward`/`detect` with a clear message, and change the documented input
   shape to `(B, 1, H, W)`. This is what the code has always assumed.
2. Support it: `reshape` instead of `view`, and decide what a multi-channel detection means —
   pooled across channels, or per channel.

Option 1 matches what every caller in kornia does (`SIFTFeature`, `GFTTAffNetHardNet` and the
rest all feed grayscale) and is the smaller change. Found while reviewing #4098.

PR #4098, opened 2026-08-29
fix(feature): stop MultiResolutionDetector fabricating detections, honour its mask, keep its dtype

ducha-aiki:
Fixes three defects in `MultiResolutionDetector`, and therefore in `KeyNetDetector`. They are in one PR because two of them are in the same function and all three are in the same file — separate PRs would conflict on every pair.

Closes #4089, closes #4090, closes #4091, closes #4102, closes #4099, closes #4100.

## 1. Fabricated detections (#4089)

`detect_features_on_single_level` masked every non-candidate position to `torch.finfo(dtype).min / 2` and then ran `topk` with `k` clamped against the **pixel count** rather than the number of surviving candidates. Once a level ran out of above-threshold maxima, `topk` could no longer rank — every remaining position carried the same sentinel — so it returned an arbitrary tie-break subset.

Before, on `main`:

```python
import torch
from kornia.feature import MultiResolutionDetector
from kornia.feature.responses import BlobHessian

torch.manual_seed(0)
det = MultiResolutionDetector(BlobHessian(), num_features=100)
lafs, resp = det(torch.rand(1, 1, 64, 64))
print(int((resp < -1e30).sum()), "of", resp.numel(), "| min", resp.min().item())
```

```
70 of 100 | min -1.7014117331926443e+38
```

I checked what those 70 rows actually were, on one level in isolation:

```
genuine above-threshold NMS maxima on this level: 7; requested k=60
returned scores that are the sentinel: 53
  true response at those positions : [-0.0]
  y range: 0 - 0   x range: 2 - 59
  all inside the 15px border strip remove_borders() zeroed: True
```

So it is not "a weak detection leaking through": the score is fabricated (the positions really hold `-0.0`), and the coordinates come from precisely the border strip `remove_borders` had just zeroed to keep detections off the frame edge.

Short results are now padded with a **zero response and a zero LAF**, which is what `ScaleSpaceDetector.detect` already does when it runs short. Everything above the shortfall is untouched — the genuine maxima are still NMS'd, thresholded and ranked exactly as before.

## 2. The `mask` argument was inert (#4090)

`forward(img, mask)` and `detect(img, mask)` accepted a mask, documented it as *"a mask with weights where to apply the response function"*, and never read it. An all-zero mask returned bit-identical output. `ScaleSpaceDetector`, in the same file, honours its mask via `_create_octave_mask`.

The mask now means **"where a detection may be"**, with one contract in both detectors, chosen from the caller's side rather than from the implementation's:

[...]

PR #4107, opened 2026-08-30
Validate ScaleSpaceDetector channel count

Marchematics:
Reject multi-channel inputs before scale-space processing, which currently assumes a single channel.

Test: python3 -m py_compile kornia/feature/scale_space_detector.py tests/feature/test_scale_space_detector.py

AI assistance was used.

### DIFF
--- a/kornia/feature/scale_space_detector.py
+++ b/kornia/feature/scale_space_detector.py
@@ -326,6 +326,7 @@ class ScaleSpaceDetector(nn.Module):
             Tuple containing detection scores and local affine frames. Local affine frames are usually shaped `(B, N, 2,
             3)`, where `N` is the selected feature count.
         """
+        KORNIA_CHECK_SHAPE(img, ["B", "1", "H", "W"])
         dev = img.device
         dtype: torch.dtype = img.dtype
         sp, sigmas, _ = self.scale_pyr(img)
--- a/tests/feature/test_scale_space_detector.py
+++ b/tests/feature/test_scale_space_detector.py
@@ -15,9 +15,11 @@
 # limitations under the License.
 #
 
+import pytest
 import torch
 
 import kornia
+from kornia.core.check import ShapeError
 from kornia.feature.scale_space_detector import (
     _MAX_ABS_SIN_12,
     MultiResolutionDetector,
@@ -30,6 +32,11 @@ from testing.base import BaseTester
 
 
 class TestScaleSpaceDetector(BaseTester):
+    def test_rejects_multichannel_input(self, device, dtype):
+        detector = ScaleSpaceDetector(2).to(device, dtype)
+        with pytest.raises(ShapeError, match="Expected shape"):
+            detector(torch.rand(1, 3, 32, 32, device=device, dtype=dtype))
+
     def test_shape(self, device, dtype):
         inp = torch.rand(1, 1, 32, 32, device=device, dtype=dtype)
         n_feats = 10


ducha-aiki (changes requested):
**REQUEST CHANGES — I recommend closing this one.** Verified on `main` @ `94d0284c`, torch 2.9.1, CPU/macOS.

**1. `main` already rejects this input, with a strictly better message.** #4098 merged today at 17:41 UTC — after this PR opened — and added an output-side check:

```python
>>> ScaleSpaceDetector(2)(torch.rand(1, 3, 32, 32))
BaseError: resp_module must return one response map per image with shape (5, 1, 32, 32). Got (5, 3, 32, 32)
```

`ShapeError` subclasses `BaseError`, so a caller catching the documented type already catches both. This PR moves the failure one line earlier and swaps a message that names the offending sub-module and the exact contract for a generic dimension mismatch.

**2. It is an API regression, not merely redundant.** The class docstring says each sub-module "could be replaced with a learned custom one, as long as they respect **output** shape", and `ScaleSpaceDetector.detect` documents `img` as `(B, C, H, W)`. So a colour-consuming `resp_module` — the documented extension point — works on `main` today and is rejected by this PR:

```python
class ColorHessian(nn.Module):
    def __init__(self): super().__init__(); self.h = BlobHessian()
    def forward(self, x, sigmas=None):
        gray = kornia.color.rgb_to_grayscale(x)
        return self.h(gray, sigmas) if sigmas is not None else self.h(gray)

ScaleSpaceDetector(4, resp_module=ColorHessian())(torch.rand(1, 3, 48, 48))
# main:     (torch.Size([1, 4, 2, 3]), torch.Size([1, 4]))     <- works
# this PR:  ShapeError: Shape mismatch at dimension 1: expected 1, got 3
```

The constraint is on the response map, which is where `main` checks it. The sibling detector agrees: `MultiResolutionDetector.detect` uses `KORNIA_CHECK_SHAPE(img, ["1", "C", "H", "W"])`, with `C` explicitly free.

**3. The green CI here is stale.** The PR is `CONFLICTING`; those checks ran before #4098 merged, so they say nothing about today's `main`.

**4. Process.** The body records `python3 -m py_compile` as the test. Anything touching this file needs `pytest tests/feature/test_scale_space_detector.py` at minimum.

ducha-aiki:
One more data point that makes the regression concrete, since it has an issue number attached.

**This PR reopens #4099.** That issue is titled "ScaleSpaceDetector raises on any multi-channel image, though it documents `(B, C, H, W)`", and #4098 closed it as completed at 17:41 UTC today — 95 minutes after this PR opened — by moving the constraint onto the *response map* (`resp_module must return one response map per image ...`) instead of onto the input image, and keeping the documented input shape. That docstring is unchanged on `main` @ `47a91725`:

```python
>>> inspect.getdoc(ScaleSpaceDetector.detect)
'... img: Input image tensor with shape `(B, C, H, W)`. ...'
```

Adding `KORNIA_CHECK_SHAPE(img, ["B", "1", "H", "W"])` to `detect` restores exactly the behaviour #4099 asked to remove, and contradicts the docstring one line below it. #4099 listed "reject `C > 1` and change the documented input shape to `(B, 1, H, W)`" as option 1 and it was not the option taken — the response-side check was, because it is the real constraint and it keeps a colour-consuming custom `resp_module` working.

So this is not a case of two valid designs; the choice was made and pinned two hours ago. Closing is the right outcome here.

PR #4108, opened 2026-08-30
Fix frn learnable eps

Marchematics:
## What did you check?

```text
$ pixi run test-module tests/...
```


ducha-aiki:
Closing: the `hynet.py` hunk here is byte-identical to #4106, which was opened first, has a filled-in description, and is approved. The only difference is the test, and #4106's test (constructs the layer and runs a forward pass) covers more than this one's.

PR #4109, opened 2026-08-30
Fix empty NMS output device

### LINKED ISSUE #4059: nms returns CPU float indices for empty input
For non-empty input, `nms` stacks indices from `scores.sort()` and returns `torch.long` on the input
device. Its empty branch constructs `torch.tensor([])` with no dtype or device, so the same API
returns CPU float32—even when the inputs are on MPS.

## Repro

```python
import torch
from kornia.geometry.bbox import nms

for device in ("cpu", "mps"):
    if device == "mps" and not torch.backends.mps.is_available():
        continue
    out = nms(torch.empty(0, 4, device=device), torch.empty(0, device=device), 0.5)
    print(device, out.dtype, out.device, out.shape)
```

## Observed

```text
cpu torch.float32 cpu torch.Size([0])
mps torch.float32 cpu torch.Size([0])
```

The only existing NMS test is non-empty and expects `torch.long`; empty input is untested.

## Expected

An empty `torch.long` tensor on `boxes.device` (for example `boxes.new_empty((0,),
dtype=torch.long)`). This makes the empty and non-empty return contract identical. Add CPU and
available accelerator coverage.

## Classification

**Broken output**: the return dtype and device depend on cardinality. Idiom 2 — the newly created
tensor fails to inherit the input device and fails to request the API's required index dtype
(`torch.long`).

## Disposition

n/a — single symbol.

## Environment

kornia 0.9.0rc1 @ `6fa75d64`, torch 2.9.1, Python 3.11.14, CPU and MPS. Found during bbox/boxes
conventions audit batch 4 (#4020); the audit does not change runtime behavior.

Marchematics:
Return empty NMS indices with the input device and dtype. Fixes #4059.

### DIFF
--- a/kornia/geometry/bbox.py
+++ b/kornia/geometry/bbox.py
@@ -590,4 +590,4 @@ def nms(boxes: torch.Tensor, scores: torch.Tensor, iou_threshold: float) -> torc
     if len(keep) > 0:
         return torch.stack(keep)
 
-    return torch.tensor(keep)
+    return boxes.new_empty((0,), dtype=torch.long)
--- a/tests/geometry/test_bbox.py
+++ b/tests/geometry/test_bbox.py
@@ -289,6 +289,14 @@ class TestBbox3D(BaseTester):
 
 
 class TestNMS(BaseTester):
+    def test_empty(self, device):
+        boxes = torch.empty((0, 4), device=device)
+        scores = torch.empty((0,), device=device)
+        actual = nms(boxes, scores, iou_threshold=0.8)
+        assert actual.shape == (0,)
+        assert actual.dtype == torch.long
+        assert actual.device == boxes.device
+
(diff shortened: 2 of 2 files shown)


ducha-aiki (approved):
**APPROVE.**

Reproduced on `main`: `nms(torch.empty(0, 4), torch.empty(0), 0.5)` returns a `float32` CPU tensor regardless of the input device. `boxes.new_empty((0,), dtype=torch.long)` is the right one-liner — same device as the input, `long` like the non-empty `torch.stack(keep)` path. `tests/geometry/test_bbox.py`: 14 passed at this head. Closes #4059.

One process note for future PRs: this same hunk is also bundled into #4110, which is about DeDoDe. Please keep one fix per PR — #4110 should drop the `bbox.py` change (or simply rebase after this merges).

PR #4110, opened 2026-08-30
Avoid CPU autocast warning in DeDoDe

### LINKED ISSUE #4094: DeDoDe warns 'device_type of cuda, but CUDA is not available' on every CPU run
Every CPU run of a DeDoDe decoder block emits

```
UserWarning: User provided device_type of 'cuda', but CUDA is not available. Disabling
```

## Reproduction

```python
import warnings, torch
from kornia.feature.dedode.decoder import ConvRefiner

m = ConvRefiner(4, 4, 4)
with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    m(torch.rand(1, 4, 16, 16))
print(m.amp, m.amp_dtype)
print([str(x.message) for x in w])
```

```
True torch.float16
["User provided device_type of 'cuda', but CUDA is not available. Disabling"]
```

## Cause

`kornia/feature/dedode/decoder.py:171` and `kornia/feature/dedode/encoder.py:69`

```python
with torch.autocast("cuda", enabled=self.amp, dtype=self.amp_dtype):
```

`ConvRefiner` defaults to `amp=True`, so the context manager is entered with
`device_type="cuda"` on machines that have no CUDA. Torch warns and disables autocast; the
numerics are correct, the warning is pure noise. `VGG19`/`VGG` default to `amp=False`, so the
encoder only warns when a caller opts in.

Both call sites already carry the comment

> AMP is intentionally scoped to "cuda" only: float16 autocast on CPU is not supported by
> PyTorch and a no-op on MPS. Use `amp_dtype=torch.float32` on non-CUDA devices to disable
> AMP entirely.

so the `"cuda"` literal is deliberate — what is missing is skipping the context manager
altogether when there is no CUDA device.

## Suggested fix

Gate on the input's device, e.g.

```python
use_amp = self.amp and feats.is_cuda
with torch.autocast("cuda", enabled=use_amp, dtype=self.amp_dtype):
```

This is byte-identical on every device: torch already disables autocast in exactly this
situation, it just warns first.

This became more visible after #4085 made the documented `DescriptorMatcherWithSteerer`
example CPU-default, so readers following the docs now see the warning.

Reproduced on `main` at 9abf06d9, torch 2.9.1, CPU.

Marchematics:
Skip CUDA autocast when DeDoDe runs on CPU. Fixes #4094.

The regression test now moves `ConvRefiner` to the parametrized device before checking autocast state.

### DIFF
--- a/kornia/feature/dedode/decoder.py
+++ b/kornia/feature/dedode/decoder.py
@@ -165,10 +165,8 @@ class ConvRefiner(nn.Module):
             surrounding class.
         """
         _b, _c, _hs, _ws = feats.shape
-        # AMP is intentionally scoped to "cuda" only: float16 autocast on CPU is not
-        # supported by PyTorch and a no-op on MPS. Use amp_dtype=torch.float32 on
-        # non-CUDA devices to disable AMP entirely.
-        with torch.autocast("cuda", enabled=self.amp, dtype=self.amp_dtype):
+        # AMP runs only for CUDA tensors.
+        with torch.autocast("cuda", enabled=self.amp and feats.is_cuda, dtype=self.amp_dtype):
             x0 = self.block1(feats)
             x = self.hidden_blocks(x0)
             if self.residual:
--- a/kornia/feature/dedode/encoder.py
+++ b/kornia/feature/dedode/encoder.py
@@ -48,9 +48,7 @@ class VGG19(nn.Module):
         self.amp_dtype = amp_dtype
 
     def forward(self, x: torch.Tensor, **kwargs):  # type: ignore[no-untyped-def]
-        # AMP is intentionally scoped to "cuda" only: float16 autocast on CPU is not
-        # supported by PyTorch and a no-op on MPS. Use amp_dtype=torch.float32 on
-        # non-CUDA devices to disable AMP entirely.
+        # AMP runs only for CUDA tensors.
         """Run this DeDoDe module forward.
 
         Inputs are image, feature, or token tensors used by the DeDoDe detector/descriptor pipeline. `B` denotes batch
@@ -66,7 +64,7 @@ class VGG19(nn.Module):
             Output tensor or dictionary produced by the module while preserving the shape contract documented by the
             surrounding class.
         """
-        with torch.autocast("cuda", enabled=self.amp, dtype=self.amp_dtype):
+        with torch.autocast("cuda", enabled=self.amp and x.is_cuda, dtype=self.amp_dtype):
             feats = []
             sizes = []
             for layer in self.layers:
--- a/kornia/feature/dedode/encoder.py
+++ b/kornia/feature/dedode/encoder.py
@@ -48,7 +48,6 @@ class VGG19(nn.Module):
         self.amp_dtype = amp_dtype
 
     def forward(self, x: torch.Tensor, **kwargs):  # type: ignore[no-untyped-def]
-        # AMP runs only for CUDA tensors.
         """Run this DeDoDe module forward.
 
         Inputs are image, feature, or token tensors used by the DeDoDe detector/descriptor pipeline. `B` denotes batch
@@ -64,6 +63,7 @@ class VGG19(nn.Module):
             Output tensor or dictionary produced by the module while preserving the shape contract documented by the
             surrounding class.
         """
+        # AMP runs only for CUDA tensors.
         with torch.autocast("cuda", enabled=self.amp and x.is_cuda, dtype=self.amp_dtype):
             feats = []
             sizes = []
--- a/tests/feature/test_dedode.py
+++ b/tests/feature/test_dedode.py
@@ -15,6 +15,8 @@
 # limitations under the License.
 #
 
+import warnings
+
 import pytest
 import torch
 
@@ -32,9 +34,12 @@ class TestConvRefiner(BaseTester):
             lambda _module, _inputs: autocast_enabled.append(torch.is_autocast_enabled("cuda"))
         )
 
-        refiner(torch.rand(1, 4, 8, 8, device=device))
+        with warnings.catch_warnings(record=True) as caught:
+            warnings.simplefilter("always")
+            refiner(torch.rand(1, 4, 8, 8, device=device))
 
         assert autocast_enabled == [device.type == "cuda"]
+        assert not any("device_type of 'cuda'" in str(w.message) for w in caught)
 
 
 @pytest.mark.skip(reason="DeDoDe is ummaintained")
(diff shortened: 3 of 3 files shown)


ducha-aiki (changes requested):
**REQUEST CHANGES** — the core change is correct; three things before it can land.

Verified on `main` @ `94d0284c`, torch 2.9.1, CPU/macOS:

- the warning reproduces on `main` (`UserWarning: User provided device_type of 'cuda', but CUDA is not available. Disabling`);
- with the patch applied the warning is gone and `ConvRefiner` output is **byte-identical** (seeded `(1,4,16,16)`: same first six elements, same sum `150.5685272216797`) — expected, since `torch.autocast("cuda")` was already a no-op on CPU and MPS tensors, so this only stops the constructor from warning;
- `360 passed, 4 skipped` across `tests/feature/test_dedode.py tests/feature/test_hynet.py tests/feature/test_scale_space_detector.py tests/geometry/test_bbox.py tests/color/test_yuv.py` at float32/float64, same as base.

Asks:

1. **Unbundle #4109.** This diff also carries `kornia/geometry/bbox.py`'s `nms` hunk and `TestNMS.test_empty` verbatim from #4109 — a different file, a different issue (#4059), and already approved on its own PR. Rebase so this branch contains only the two `dedode` lines plus its own test. As it stands the two PRs cannot be merged independently.

2. **The comment directly above the changed line is now false.** `decoder.py` says: "Use `amp_dtype=torch.float32` on non-CUDA devices to disable AMP entirely." That manual workaround is exactly what this patch removes the need for, so the advice now misdescribes the code beneath it. Please update it (and check whether the `encoder.py` site needs the same).

3. **Nit — the pin only pins on a CUDA-less machine.** `test_conv_refiner_cpu_amp_does_not_warn` is a bare module-level function that hardcodes CPU; on a machine where CUDA *is* available `main` emits no warning either, so the test passes on base there and proves nothing. Please move it into a `BaseTester` subclass per TESTING.md and either note the device dependence or assert on `torch.is_autocast_enabled` state rather than on the absence of a warning.

Marchematics:
Rebased on current main. Removed the merged NMS change and replaced the warning test with a device-fixture autocast-state test.

Marchematics:
Updated the regression test to move `ConvRefiner` onto the parametrized device before the autocast check.

ducha-aiki:
**REQUEST CHANGES — and the reason is my own bad advice, not your work.** The production change is right; the replacement test cannot detect it.

Setup: worktree at the head `0ed3b409`, with `kornia/` swapped to the merge-base `47a91725`; `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS, no CUDA.

## The new test passes on base

```
# kornia/ at the merge-base, your test kept
tests/feature/test_dedode.py  1 passed, 2 skipped, 1 warning
  UserWarning: User provided device_type of 'cuda', but CUDA is not available. Disabling
# at the head
1 passed, 2 skipped
```

Same verdict both sides. `test_amp_matches_input_device` does not discriminate.

**Why:** on a machine without CUDA, `torch.autocast("cuda", enabled=True)` does not enable autocast — it warns and disables itself. So `torch.is_autocast_enabled("cuda")` reads `False` during `forward` on *both* revisions:

```
kornia/ at 47a91725 (base):  autocast_enabled_during_forward: [False]   warnings: 1
kornia/ at the PR head:      autocast_enabled_during_forward: [False]   warnings: 0
```

The state assertion is invariant across the fix on exactly the machine where the bug reproduces. **That is my error: my point 3 said "assert on `torch.is_autocast_enabled` state rather than on the absence of a warning", and "rather than" was wrong.** I retract it.

## What the pin actually needs

The two assertions are complementary, and neither alone works everywhere:

- **no-CUDA machine** (our CI, your laptop): base warns, head does not -> only the *warning* assertion discriminates. The state is `False` either way.
- **CUDA machine, CPU input**: no warning either way, but base leaves autocast *enabled* for `device_type="cuda"` while running CPU tensors, head disables it -> only the *state* assertion discriminates.

So assert both, and the pin holds on either kind of runner:

```python
class TestConvRefiner(BaseTester):
    def test_amp_is_scoped_to_the_input_device(self, device):
        refiner = ConvRefiner(in_dim=4, hidden_dim=4, out_dim=4).to(device)
        enabled = []
        refiner.block1.register_forward_pre_hook(
            lambda _m, _i: enabled.append(torch.is_autocast_enabled("cuda"))
        )
        with warnings.catch_warnings(record=True) as caught:
            warnings.simplefilter("always")
            refiner(torch.rand(1, 4, 8, 8, device=device))
        # discriminates on a CUDA machine
        assert enabled == [device.type == "cuda"]
        # discriminates on a machine without CUDA
        assert not [w for w in caught if "device_type of 'cuda'" in str(w.message)]
```

Your original warning-based test was closer to right than the one that replaced it; the part of point 3 that stands is only "put it in a `BaseTester` subclass with the `device` fixture", which you have done.

## Nit, non-blocking

The replacement comment in `encoder.py` sits between `def forward(...)` and the docstring, ~15 lines above the `with torch.autocast(...)` it describes. That placement is inherited, not yours, but since you are touching the line anyway it would read better directly above the `with`. And the text it replaced carried information worth keeping in some form — that the scoping is deliberate, not an oversight.

Points 1 (unbundle #4109) and 2 (the false comment) from `pullrequestreview-5061584186` are both done; thank you.

Marchematics:
Updated the regression test to check both device-scoped autocast state and the CUDA warning, and moved the encoder comment above the autocast context. The focused test passes locally.

ducha-aiki:
## Delta re-review of #4110 — APPROVE

The pin now carries both assertions, which is exactly the fix for my retracted round-1 ask.

Setup: worktree at the PR head with `kornia/` swapped to the merge-base `47a91725`. `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU/macOS (no CUDA).

```
head:  tests/feature/test_dedode.py -k TestConvRefiner            1 passed
base:  kornia/ -> 47a91725                                        1 failed
  >  assert not any("device_type of 'cuda'" in str(w.message) for w in caught)
  E  assert not True
```

So on a CUDA-less runner the **warning** assertion is the one that discriminates — as expected, since `torch.autocast("cuda", enabled=True)` warns and disables itself there, leaving `autocast_enabled == [False]` on both revisions. On a CUDA runner with CPU input the roles swap and `assert autocast_enabled == [device.type == "cuda"]` is the discriminating half. The two together cover both runners, which was the point.

The `enabled=self.amp and x.is_cuda` form is right, and the relocated comment reads better where it now is.

Approving.

ducha-aiki (approved):

PR #4111, opened 2026-08-30
Raise ShapeError for empty YUV planes

### LINKED ISSUE #4056: [Bug]: yuv420_to_rgb / yuv422_to_rgb raise ZeroDivisionError instead of ShapeError for zero-width chroma
### Describe the bug

`yuv420_to_rgb` and `yuv422_to_rgb` raise a bare `ZeroDivisionError` when the chroma plane has a zero-sized dimension. Every other malformed input to these functions raises `ShapeError`.

This is the same class of defect as #4050 — a shape guard reporting the wrong exception type — but on a different axis, so it survives the fix in #4055.

It matters for the same reason #4050 did: `ShapeError` derives from `BaseError`, and `ZeroDivisionError` derives from `ArithmeticError`. Neither is a `RuntimeError`, and neither is the other, so a caller that wraps these conversions to validate user-supplied shapes catches nothing useful. A zero-sized dimension is also a plausible thing to hit in practice — it is what an empty slice or a degenerate crop produces upstream.

### Reproduction

    import torch
    import kornia

    # 4:2:2 - zero-width chroma
    kornia.color.yuv422_to_rgb(torch.rand(1, 4, 6), torch.rand(2, 4, 0))
    # ZeroDivisionError: division by zero

    # 4:2:0 - zero-width chroma
    kornia.color.yuv420_to_rgb(torch.rand(1, 4, 4), torch.rand(2, 2, 0))
    # ZeroDivisionError: division by zero

    # 4:2:0 - zero-height chroma
    kornia.color.yuv420_to_rgb(torch.rand(1, 4, 4), torch.rand(2, 0, 2))
    # ZeroDivisionError: division by zero

For contrast, a non-zero mismatch on the same functions is reported properly:

    # ShapeError: Input imageuv H&W must be half the size of the luma plane. ...

### Expected behavior

`ShapeError`, consistent with every other shape violation these functions report.

### Cause

Both guards divide by a chroma dimension before anything validates it:

    # yuv420_to_rgb - kornia/color/yuv.py:211-212
    or imagey.shape[-2] / imageuv.shape[-2] != 2

    # yuv422_to_rgb - kornia/color/yuv.py:258

When the divisor is `0` the expression raises before the comparison can decide anything, so the guard never gets to produce its own message.

### Credit and prior context

Full credit to @ducha-aiki — he found this while reviewing #4055 and suggested it be filed as its own issue rather than folded into that PR. I am only writing it up.

From the 38,416-case sweep in his review, the clause ordering in #4055 already converts 2,400 combinations from `ZeroDivisionError` to `ShapeError` as a side effect, because the new height clause short-circuits the division. Roughly 400 remain reachable on that branch, all of the shape `chroma width 0, height matches`, plus the 4:2:0 twin which #4055 does not touch at all.

### Environment

- kornia: 0.8.3 (also reproduces on `main`)
- PyTorch: 2.13.0+cpu
- Python: 3.13
- OS: Linux, CPU

### Anything else?

Happy to open a PR for this if it is wanted. Two things I would want steering on first:

1. Whether both twins should be fixed in one PR, or one each.
2. Whether the guards should reject zero-sized planes outright, or whether the division should simply be reordered behind an explicit zero check so the existing message still applies.

### How did AI help?

I used an AI assistant (Arena.ai Agent Mode) to draft this report. It verified the reproduction against kornia 0.8.3 in a sandbox, located the two guard sites, and wrote the text above. I ran the reproduction myself, confirmed the source lines, and I understand the defect and can discuss it.

### Minimal reproduction

```python
import torch
import kornia

# 4:2:2 - zero-width chroma
kornia.color.yuv422_to_rgb(torch.rand(1, 4, 6), torch.rand(2, 4, 0))
# ZeroDivisionError: division by zero

# 4:2:0 - zero-width chroma
kornia.color.yuv420_to_rgb(torch.rand(1, 4, 4), torch.rand(2, 2, 0))
# ZeroDivisionError: division by zero

# 4:2:0 - zero-height chroma
kornia.color.yuv420_to_rgb(torch.rand(1, 4, 4), torch.rand(2, 0, 2))
# ZeroDivisionError: division by zero
```

### Environment

- Kornia version: 0.8.3 (also reproduces on main)
- PyTorch version: 2.13.0+cpu
- Python version: 3.13
- OS / device (CPU, CUDA, ROCm, MPS): Linux, CPU

### Anything else?

Full credit to @ducha-aiki — he found this while reviewing #4055 and suggested it be filed as its own issue rather than folded into that PR. I am only writing it up.

Happy to open a PR for this if it is wanted. Two things I would want steering on first:

1. Whether both twins should be fixed in one PR, or one each.
2. Whether the guards should reject zero-sized planes outright, or whether the division should simply be reordered behind an explicit zero check so the existing message still applies.

**How did AI help?** I used an AI assistant (Arena.ai Agent Mode) to draft this report. It verified the reproduction against kornia 0.8.3 in a sandbox, located the two guard sites, and wrote the text above. I ran the reproduction myself, confirmed the source lines, and I understand the defect and can discuss it.

Marchematics:
Report zero-sized chroma planes as `ShapeError`. Empty matching planes return empty RGB. Fixes #4056.

Tests: `tests/color/test_yuv.py` covers zero-sized chroma errors and intentional empty-image outputs. Kornia CI passed.

### DIFF
--- a/kornia/color/yuv.py
+++ b/kornia/color/yuv.py
@@ -228,8 +228,8 @@ def yuv420_to_rgb(imagey: torch.Tensor, imageuv: torch.Tensor) -> torch.Tensor:
     if (
         len(imageuv.shape) < 2
         or len(imagey.shape) < 2
-        or imagey.shape[-2] / imageuv.shape[-2] != 2
-        or imagey.shape[-1] / imageuv.shape[-1] != 2
+        or imagey.shape[-2] != 2 * imageuv.shape[-2]
+        or imagey.shape[-1] != 2 * imageuv.shape[-1]
     ):
         raise ShapeError(
             f"Input imageuv H&W must be half the size of the luma plane. Got {imagey.shape} and {imageuv.shape}"
@@ -280,7 +280,7 @@ def yuv422_to_rgb(imagey: torch.Tensor, imageuv: torch.Tensor) -> torch.Tensor:
         len(imageuv.shape) < 2
         or len(imagey.shape) < 2
         or imagey.shape[-2] != imageuv.shape[-2]
-        or imagey.shape[-1] / imageuv.shape[-1] != 2
+        or imagey.shape[-1] != 2 * imageuv.shape[-1]
     ):
         raise ShapeError(
             f"Input imageuv H must match the luma plane and W must be half its size. "
--- a/tests/color/test_yuv.py
+++ b/tests/color/test_yuv.py
@@ -699,6 +699,21 @@ class TestYuv420ToRgb(BaseTester):
             imguv = torch.ones(2, 1, 1, device=device, dtype=dtype)
             kornia.color.yuv420_to_rgb(imgy, imguv)
 
+        with pytest.raises(ShapeError):
+            imgy = torch.ones(1, 4, 4, device=device, dtype=dtype)
+            imguv = torch.empty(2, 2, 0, device=device, dtype=dtype)
+            kornia.color.yuv420_to_rgb(imgy, imguv)
+
+        with pytest.raises(ShapeError):
+            imgy = torch.ones(1, 4, 4, device=device, dtype=dtype)
+            imguv = torch.empty(2, 0, 2, device=device, dtype=dtype)
+            kornia.color.yuv420_to_rgb(imgy, imguv)
+
+    def test_empty_input(self, device, dtype):
+        imgy = torch.empty(1, 0, 0, device=device, dtype=dtype)
+        imguv = torch.empty(2, 0, 0, device=device, dtype=dtype)
+        assert kornia.color.yuv420_to_rgb(imgy, imguv).shape == (3, 0, 0)
+
     @pytest.mark.parametrize("name", list(_REFERENCE_COLORS))
     def test_unit(self, device, dtype, name):
         rgb_values, yuv_values = _REFERENCE_COLORS[name]
@@ -826,6 +841,16 @@ class TestYuv422ToRgb(BaseTester):
             imgy = torch.ones(1, 2, 2, device=device, dtype=dtype)
             imguv = torch.ones(2, 1, 1, device=device, dtype=dtype)
             kornia.color.yuv422_to_rgb(imgy, imguv)
+
+        with pytest.raises(ShapeError):
+            imgy = torch.ones(1, 4, 4, device=device, dtype=dtype)
+            imguv = torch.empty(2, 4, 0, device=device, dtype=dtype)
+            kornia.color.yuv422_to_rgb(imgy, imguv)
+
+    def test_empty_input(self, device, dtype):
+        imgy = torch.empty(1, 4, 0, device=device, dtype=dtype)
+        imguv = torch.empty(2, 4, 0, device=device, dtype=dtype)
+        assert kornia.color.yuv422_to_rgb(imgy, imguv).shape == (3, 4, 0)
 
     @pytest.mark.parametrize("name", list(_REFERENCE_COLORS))
     def test_unit(self, device, dtype, name):


ducha-aiki (changes requested):
**REQUEST CHANGES.** The `a / b != 2` → `a != 2 * b` rewrite is the right shape of fix. I rebased it onto today's `main` by hand and confirmed all three still-live cases flip from `ZeroDivisionError` to `ShapeError`, with `360 passed, 4 skipped` across the colour and bbox/feature modules at float32/float64. Three things before it lands.

**1. Rebase — the conflict is load-bearing.** #4055 merged today at 18:16 UTC and rewrote the `yuv422_to_rgb` guard to add the missing chroma-*height* clause. This PR's `yuv422` hunk is written against the old single-clause guard, so a mechanical conflict resolution can silently drop #4055's new check. On today's `main` the only line to change there is

```python
```

and `or imagey.shape[-2] != imageuv.shape[-2]` above it must survive verbatim.

Please also re-measure which cases are still broken, since #4055 already converted the 4:2:2 zero-*height* case by short-circuiting. On `main` @ `94d0284c` the remaining three are 4:2:0 zero-width, 4:2:0 zero-height, and 4:2:2 zero-width.

**2. Undocumented behaviour change: a zero-sized *luma* plane now passes the guard.**

```python
yuv420_to_rgb(torch.rand(1, 0, 0), torch.rand(2, 0, 0))
# main:    ZeroDivisionError
# this PR: returns torch.Size([3, 0, 0])
```

because `0 != 2 * 0` is False, so the guard falls through to `repeat_interleave`/`cat` and produces an empty image. Same for `yuv422_to_rgb(torch.rand(1, 4, 0), torch.rand(2, 4, 0))` → `(3, 4, 0)`. An empty image out for an empty image in is defensible, but it is a decision rather than a side effect, and none of the added tests covers it. Please say which behaviour you intend in the PR body and pin it with a test either way.

**3. That decision is literally the open steering question on the issue.** #4056 asks "whether the guards should reject zero-sized planes outright, or whether the division should simply be reordered behind an explicit zero check so the existing message still applies" — this PR answers it silently with the second option. The reporter offered to open this PR themselves and is still waiting on that answer, so please link the resolution on #4056 before merging.

ducha-aiki:
Follow-up on point 2 of my review — I surveyed how the rest of kornia handles degenerate shapes, and it resolves in your favour. Softening the ask accordingly.

I called the zero-luma case "an undocumented behaviour change". It is better described as **the house convention**, and your choice of `a != 2 * b` is the right one. kornia's rule is *empty in → empty out, with validation parity*: `_empty_warp_output_2d` in `imgwarp.py` implements it deliberately ("Return an empty warp while retaining normal grid-sample validation and autograd links", pinned by 13 tests), `create_meshgrid` documents it in prose, and 20 of 26 ops I probed pass zero-sized input straight through. Decisive for this PR: the 4:4:4 pair in the very same file already does it — `yuv_to_rgb(torch.rand(3, 0, 0))` returns `(3, 0, 0)`.

So the revised ask is smaller than what I wrote:

- **Do not special-case it away.** `yuv420_to_rgb(rand(1,0,0), rand(2,0,0))` returning `(3, 0, 0)` is the correct outcome, because it makes 4:2:0 agree with `yuv_to_rgb`.
- **Do add a test for it**, next to the `ShapeError` ones you already added, and say in the PR body that it is intentional. That is all point 2 needs.

Full reasoning and the measurements are on #4056, where I have now answered the two steering questions the reporter raised: reorder rather than reject, and one PR rather than two — which is what this PR already does.

Points 1 (rebase past #4055, keeping its chroma-height clause) and 3 stand unchanged.

Marchematics:
Rebased on current main. Kept the YUV422 height check, added empty luma output tests, and linked the resolution in #4056.

ducha-aiki:
**Re-review of `5b431bd9`: your two blockers are cleared and the fix is verified correct.** But there is a scheduling problem that is not your fault, and you should hear it from me rather than discover it.

Setup: worktree at this head with `kornia/` swapped to the merge-base `47a91725`; `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS.

**Blockers from `pullrequestreview-5061584795`:**

1. *Rebase* — done, the PR is `MERGEABLE` again.
2. *The zero-luma pass-through* — I already retracted this one on `issuecomment-5470620404`: it is kornia's documented convention, not a behaviour change, and your version keeps it.

**Verification** (fail-on-base 4 failed / 104 passed; head 214 passed at float32+float64), plus a 4802-cell shape sweep of both converters against the merge-base:

```
total probes: 4802   changed: 220
  ZeroDivisionError -> ShapeError        206
  ZeroDivisionError -> empty output       14
  worked -> raises                         0
  worked -> different result               0
```

**The scheduling problem.** #4133, opened yesterday by @Himesh-rupchandani (who filed #4056), contains a **byte-identical** production hunk to yours — the same `!= 2 *` reorder in both functions — with a slightly fuller test: its `test_empty_input` also cross-checks the result against `yuv_to_rgb(cat([y, uv]))`, pinning *why* the empty case is allowed rather than only its shape.

Since the shipped code would be the same either way, the tie-break is the test, and I intend to land #4133 and close this one. That is not a judgement on your work: your fix was first, it is correct, and it is the same fix. I would rather say that out loud than quietly merge the other PR and leave you guessing.

Thank you for the rebase and for taking the earlier round seriously.

Marchematics:
Thanks for the review. I am happy for #4133 to land with the fuller test coverage. Please close this PR.

PR #4112, opened 2026-08-30
Support scalar inputs in bbox_generator3d

### LINKED ISSUE #4057: bbox_generator3d rejects its documented scalar tensor inputs
`bbox_generator3d` documents every argument as a scalar tensor or `(B,)`, and its guard explicitly
accepts both ranks. The scalar path then calls `len(x_start)` and raises before producing a box.
The 2D sibling accepts the same scalar contract.

## Repro

```python
import torch
from kornia.geometry.bbox import bbox_generator, bbox_generator3d

for dtype in (torch.int64, torch.float32, torch.float64):
    scalar = [torch.tensor(1, dtype=dtype) for _ in range(6)]
    try:
        print("3d", dtype, bbox_generator3d(*scalar).shape)
    except Exception as error:
        print("3d", dtype, type(error).__name__, error)
    print("2d", dtype, bbox_generator(*scalar[:4]).shape)
```

## Observed

```text
3d torch.int64 TypeError len() of a 0-d tensor
3d torch.float32 TypeError len() of a 0-d tensor
3d torch.float64 TypeError len() of a 0-d tensor
2d <each dtype> torch.Size([1, 4, 2])
```

`(1,)` inputs work. 3D transform integration tests call the generator with rank-1 tensors, but no
dedicated test exercises the documented scalar form.

## Expected

The documented scalar form returns `(1,8,3)`, matching `bbox_generator`. Replacing the allocation's
`len(x_start)` with a scalar-safe batch size (or directly stacking as the 2D sibling does) fixes the
crash without changing working `(B,)` calls. Add scalar and batched tests across unequal dimensions.

## Classification

**Broken output** (documented input crashes). No existing idiom fits exactly; the guard accepts the
right contract but the implementation later assumes rank 1.

## Disposition

n/a — single symbol.

## Environment

kornia 0.9.0rc1 @ `6fa75d64`, torch 2.9.1, Python 3.11.14, CPU. Found during bbox/boxes
conventions audit batch 4 (#4020); the audit does not change runtime behavior.

Marchematics:
Support the documented scalar tensor inputs in bbox_generator3d. Fixes #4057.

### DIFF
--- a/kornia/geometry/bbox.py
+++ b/kornia/geometry/bbox.py
@@ -464,19 +464,19 @@ def bbox_generator3d(
     # front
     bbox = torch.tensor(
         [[[0, 0, 0], [0, 0, 0], [0, 0, 0], [0, 0, 0]]], device=x_start.device, dtype=x_start.dtype
-    ).repeat(len(x_start), 1, 1)
+    ).repeat(x_start.numel(), 1, 1)
 
     bbox[:, :, 0] += x_start.view(-1, 1)
     bbox[:, :, 1] += y_start.view(-1, 1)
     bbox[:, :, 2] += z_start.view(-1, 1)
-    bbox[:, 1, 0] += width
-    bbox[:, 2, 0] += width
-    bbox[:, 2, 1] += height
-    bbox[:, 3, 1] += height
+    bbox[:, 1, 0] += width.view(-1)
+    bbox[:, 2, 0] += width.view(-1)
+    bbox[:, 2, 1] += height.view(-1)
+    bbox[:, 3, 1] += height.view(-1)
 
     # back
     bbox_back = bbox.clone()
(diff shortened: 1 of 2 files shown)


ducha-aiki (approved):
**APPROVE.**

Reproduced on `main`: six 0-d tensors raise `TypeError: len() of a 0-d tensor`, which contradicts the docstring. `numel()` + `.view(-1)` is the minimal fix and leaves the batched path unchanged.

Measured at this head (`.venv`, torch 2.9.1, CPU):

```
tests/geometry/test_bbox.py --dtype=float32          all passed
tests/geometry/test_bbox.py --dtype=float16          3 failed — TestTransformBoxes2D::test_transform_boxes{,_wh,_multiple}
```

Those three float16 failures are **pre-existing on `main`** (same three node IDs fail there) and are unrelated to `bbox_generator3d`, so they do not count against this PR. Closes #4057.

PR #4114, opened 2026-08-30
fix(geometry): parse the PLY header instead of skipping header_size lines

### LINKED ISSUE #4113: PLY loaders skip a fixed 8 header lines instead of parsing the header: a minimal PLY file cannot be read, and element vertex N is ignored
## Summary

`kornia.geometry.load_pointcloud_ply` and `load_pointcloud_ply_binary` do not parse the PLY header. They skip `header_size=8` lines and read everything after that as `x y z` triples. That default only matches the eight-line header kornia's own writers emit (which includes a `comment` line); it fails on the format itself.

## Reproduction (`main`, torch 2.9.1)

```python
import struct, torch, kornia
hdr = b"ply\nformat binary_little_endian 1.0\nelement vertex 1\nproperty double x\nproperty double y\nproperty double z\nend_header\n"
open("min.ply", "wb").write(hdr + struct.pack("<3d", 1.0, 2.0, 3.0))
kornia.geometry.load_pointcloud_ply_binary("min.ply")
# ValueError: Expected 24 bytes per point, got a total of 40 values.
# (seven header lines: the loader skipped `end_header` *and* the first 8 bytes of the point)
```

Other consequences of the same design:

- `element vertex N` is ignored: a `face` element after the vertices is read as extra points (binary) or makes `len(parts) % 3 != 0` (ASCII).
- Extra vertex properties (`nx ny nz`, `red green blue`) are read as coordinates or rejected.
- Coordinates declared as `float` (the most common layout in the wild) are read as `double`.
- `binary_big_endian` is silently read as little-endian.
- A header with a different number of `comment` / `obj_info` lines shifts the payload by the difference.

## Fix

Parse the header to `end_header`; take the count from `element vertex N`; pick `x`, `y`, `z` by property name and type; skip trailing elements; honour the endianness. Deprecate `header_size` (ignored, warns).

Diagnosed originally by @wiliyam in #3973 (closed; the reader is being rewritten by the maintainers).

ducha-aiki:

Rewrites `load_pointcloud_ply` and `load_pointcloud_ply_binary` to **parse the PLY header** instead of skipping a fixed `header_size=8` lines. The old readers only worked on files kornia's own writers produced (eight header lines including a `comment`); a minimal seven-line PLY file lost its first point to the `end_header` line, `element vertex N` was ignored, extra vertex properties and trailing `face` elements were misread, `float` coordinates were read as `double`, and `binary_big_endian` was read as little-endian.

Now, for both readers:

- the header is parsed to `end_header` (`comment` / `obj_info` skipped, unknown keywords rejected);
- the point count comes from `element vertex N`;
- `x`, `y`, `z` are selected **by property name**, at any PLY scalar type and in any order;
- elements after the vertices are ignored; scalar-only elements before them are skipped by size (a `list` property among the vertex properties raises in both readers, since its length is unknowable without parsing; in an element *before* the vertices it raises in the binary reader only, because the ASCII reader skips such an element by lines and a list is still one line);
- the binary reader honours `binary_little_endian` / `binary_big_endian`, uses one bulk `array`+`frombuffer` copy when the vertex layout is homogeneous and `struct.iter_unpack` otherwise;
- each reader rejects the other's format with a message naming the right function;
- `header_size` is **deprecated and ignored** (`DeprecationWarning` when passed) rather than removed, per the stability policy.

Output is unchanged: `(N, 3)` `float32`.

Behavioural note, stated in the CHANGELOG: a file *without* a proper PLY header, which the line-skipping loader could read by accident with a tuned `header_size`, is now rejected with a `ValueError`.

Closes #4113. The defect was diagnosed by @wiliyam in #3973; that PR was closed in favour of this rewrite.

## What did you check?

```text
$ .venv/bin/python -m pytest tests/geometry/test_pointcloud_io.py        48 passed
$ (same test file, kornia/ reverted to main)                              38 failed  <- 38 of the 39 new pins
$ pixi run pre-commit-all                                                 all hooks pass
$ pixi run typecheck                                                      only the pre-existing lightglue custom_fwd warning
```

The one new pin that passes on `main` is `test_ascii_empty`: the old loader also returned `(0, 3)` for an
empty cloud, by way of `torch.tensor([]).view(-1, 3)` rather than a declared vertex count of zero.

New pins (`TestLoadPointCloudPlyHeaderParsing`, 39 node IDs): minimal seven-line header (ASCII + binary), trailing `face` element, homogeneous extra properties (normals), mixed `float`+`uchar` layout with `z` declared first, big-endian payload, scalar element before the vertices (ASCII + binary), empty cloud (ASCII + binary), 20 000-line comment header, fourteen malformed-header cases (every rejection branch in `_read_ply_header`), truncated payload (ASCII and binary), truncated element preceding the vertices, a header with no `vertex` element (both readers), list-before-vertices, list property *inside* the vertex element (both readers), ASCII short line, ASCII non-numeric coordinate, ASCII/binary cross-use, and the `header_size` deprecation for both readers.

Performance, 200 000 points on this machine (torch 2.9.1, CPU), best of seven against a `main` worktree: binary 1.7 ms → 1.8 ms (header parse + column select); ASCII 74.5 ms → 91 ms because each line is now split and indexed by column instead of the whole payload being tokenised at once — the price of honouring the declared layout. (The first revision of this PR read 127 ms; the difference was a generator expression in the vertex loop, not the layout handling. A bulk token split for the plain three-column layout would reach 68.6 ms, but it misaligns silently on a row carrying more tokens than the header declares, which the per-line loop reads correctly, so it is not taken.) Both writers are untouched; the existing round-trip tests pass.

Not verified: CUDA/MPS are irrelevant here (file I/O returns CPU tensors, as before).

## How did AI help?

Written by Claude (Fable 5) under maintainer direction; the maintainer set the design (parse the header, deprecate `header_size`) and reviewed the diff and the numbers above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,7 +86,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   count from `element vertex N`, select `x`, `y`, `z` by property name whatever their scalar type, skip trailing
   elements, and read big-endian payloads. `header_size` is deprecated and ignored (a `DeprecationWarning` is
   emitted when it is passed). Files without a proper PLY header, which the old line-skipping loader could read
-  by accident, are now rejected with a `ValueError`.
+  by accident, are now rejected with a `ValueError`, and so is a file that declares a `list` property among the
+  vertex properties or in an element that precedes them -- a list has no statically known length, so the columns
+  after it cannot be located.
 
 * Make `guided_blur` / `GuidedBlur` work in `float16` and `bfloat16` when the guidance has more than one channel.
   Multi-channel guidance solves a per-pixel `C x C` system, and `torch.linalg.solve` has no half-precision LU
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,8 +87,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   elements, and read big-endian payloads. `header_size` is deprecated and ignored (a `DeprecationWarning` is
   emitted when it is passed). Files without a proper PLY header, which the old line-skipping loader could read
   by accident, are now rejected with a `ValueError`, and so is a file that declares a `list` property among the
-  vertex properties or in an element that precedes them -- a list has no statically known length, so the columns
-  after it cannot be located.
+  vertex properties -- a list has no statically known length, so the columns after it cannot be located. The
+  binary reader additionally rejects a `list` property in an element that *precedes* the vertices, whose byte
+  length it would have to know to reach the vertex data; the ASCII reader reads such a file, since every ASCII
+  element instance is one line whatever it contains.
 
 * Make `guided_blur` / `GuidedBlur` work in `float16` and `bfloat16` when the guidance has more than one channel.
   Multi-channel guidance solves a per-pixel `C x C` system, and `torch.linalg.solve` has no half-precision LU
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* Make `load_pointcloud_ply` and `load_pointcloud_ply_binary` parse the PLY header instead of skipping a fixed
+  number of lines. Both loaders skipped `header_size=8` lines and read everything after them as `x y z` triples, so a
+  minimal PLY file (seven header lines, no `comment`) lost its first point to a header line -- the binary reader
+  returned header bytes decoded as doubles -- and any file with extra vertex properties (normals, colours) or a
+  `face` element after the vertices was rejected or misread. The readers now stop at `end_header`, take the point
+  count from `element vertex N`, select `x`, `y`, `z` by property name whatever their scalar type, skip trailing
+  elements, and read big-endian payloads. `header_size` is deprecated and ignored (a `DeprecationWarning` is
+  emitted when it is passed). Files without a proper PLY header, which the old line-skipping loader could read
+  by accident, are now rejected with a `ValueError`.
+
 * Make `guided_blur` / `GuidedBlur` work in `float16` and `bfloat16` when the guidance has more than one channel.
   Multi-channel guidance solves a per-pixel `C x C` system, and `torch.linalg.solve` has no half-precision LU
   kernel, so every such call failed: on CPU with `NotImplementedError: "lu_cpu" not implemented for 'Half'`, and on
--- a/kornia/geometry/pointcloud.py
+++ b/kornia/geometry/pointcloud.py
@@ -224,6 +224,15 @@ def _ply_vertex_layout(header: _PlyHeader, filename: str) -> Tuple[int, _PlyElem
     """Locate the ``vertex`` element and the column indices of ``x``, ``y`` and ``z``."""
     for index, element in enumerate(header.elements):
         if element.name == "vertex":
+            for name, scalar_type in element.properties:
+                if scalar_type is None:
+                    # An ASCII list spans ``1 + n`` tokens and a binary one an unknowable number of
+                    # bytes, so a list among the vertex properties makes every column after it
+                    # unlocatable. Only elements that follow the vertices may carry one.
+                    raise ValueError(
+                        f"PLY vertex element in {filename!r} has a list property {name!r}; "
+                        "list properties are only supported in elements that follow the vertices."
+                    )
             names = [name for name, _ in element.properties]
             try:
                 xyz = (names.index("x"), names.index("y"), names.index("z"))
@@ -238,7 +247,7 @@ def _ply_vertex_layout(header: _PlyHeader, filename: str) -> Tuple[int, _PlyElem
 def _warn_header_size(header_size: Optional[int]) -> None:
     if header_size is not None:
         warnings.warn(
-            "`header_size` is ignored since kornia 0.8.3: the PLY header is parsed up to `end_header`. "
+            "`header_size` is ignored since kornia 0.9.0: the PLY header is parsed up to `end_header`. "
             "The argument will be removed in a future release.",
             DeprecationWarning,
             stacklevel=3,
@@ -258,6 +267,8 @@ def load_pointcloud_ply(filename: str, header_size: Optional[int] = None) -> tor
     The header is parsed up to ``end_header``; the number of points comes from the
     ``element vertex N`` declaration and the coordinates from the ``x``, ``y`` and ``z``
     properties, whatever other properties (normals, colours) or elements (faces) the file carries.
+    A ``list`` property among the vertex properties is rejected, since it spans a variable number of
+    tokens and would shift every column after it.
 
     Args:
         filename: the path to the pointcloud.
@@ -285,15 +296,21 @@ def load_pointcloud_ply(filename: str, header_size: Optional[int] = None) -> tor
                     raise ValueError(f"PLY file {filename!r} ends inside element {element.name!r}.")
 
         num_columns = len(vertex.properties)
+        # The indices are hoisted and the three conversions written out: a generator expression here
+        # costs ~30% of the read time on a 200k-point file.
+        index_x, index_y, index_z = xyz
         points: List[float] = []
-        for _ in range(vertex.count):
+        for row in range(vertex.count):
             tokens = f.readline().split()
             if len(tokens) < num_columns:
                 raise ValueError(
                     f"PLY file {filename!r} declares {vertex.count} vertices with {num_columns} properties, "
                     f"but a vertex line has {len(tokens)} values."
                 )
-            points.extend(float(tokens[i]) for i in xyz)
+            try:
+                points.extend((float(tokens[index_x]), float(tokens[index_y]), float(tokens[index_z])))
+            except ValueError as exc:
+                raise ValueError(f"PLY file {filename!r} has a non-numeric coordinate on vertex {row}: {exc}.") from exc
 
     return torch.tensor(points, dtype=torch.float32).reshape(-1, 3)
 
@@ -305,8 +322,9 @@ def load_pointcloud_ply_binary(filename: str, header_size: Optional[int] = None)
     ``element vertex N`` declaration and the coordinates from the ``x``, ``y`` and ``z``
     properties, whatever their scalar type. Extra vertex properties (normals, colours) are skipped,
     and so are elements that follow the vertices (faces). Both little- and big-endian payloads are
-    read. Elements that *precede* the vertices are skipped only when they contain no list property,
-    since the byte length of a list is unknown without parsing it.
+    read. A ``list`` property is only supported in elements that *follow* the vertices. Among the
+    vertex properties, or in an element that precedes them, its byte length is unknown without
+    parsing it, so such a file is rejected.
 
     Args:
         filename: the path to the pointcloud.
--- a/kornia/geometry/pointcloud.py
+++ b/kornia/geometry/pointcloud.py
@@ -17,7 +17,10 @@
 
 import array
 import os
+import struct
 import sys
+import warnings
+from typing import BinaryIO, Dict, List, NamedTuple, Optional, Tuple
 
 import torch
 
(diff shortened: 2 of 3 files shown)


ducha-aiki:
## Review of #4114 — parse the PLY header instead of skipping `header_size` lines

Reviewed at `886c3756`. I re-ran everything locally before writing this: `tests/geometry/test_pointcloud_io.py` **32 passed**, `ruff check` on both changed files clean, `pixi run typecheck` shows only the pre-existing `lightglue` `custom_fwd` diagnostic. I also re-measured the performance numbers against a `main` worktree (`94d0284c`) — they hold, see §3.

The direction is right and the header parser is the correct fix for #4113: the point count now comes from `element vertex N`, `x`/`y`/`z` are selected by name at any scalar type, trailing elements are ignored, and big-endian payloads are read. Two things to change before merge, plus three smaller notes.

---

### 1. ASCII: a `list` property in the *vertex* element silently reads the wrong columns

`_ply_vertex_layout` (`kornia/geometry/pointcloud.py:229`) takes the x/y/z indices as positions in the declared property list. In binary that is exactly right, and `element_struct` rejects a vertex element containing a `list` property outright. In **ASCII** there is no such guard, and a `property list` occupies `1 + n` tokens on the row — so every column after it is shifted and the reader returns wrong numbers with no error.

```python
hdr = (b"ply\nformat ascii 1.0\nelement vertex 2\n"
       b"property list uchar int tags\n"
       b"property double x\nproperty double y\nproperty double z\nend_header\n")
body = b"2 10 20 1 2 3\n2 30 40 4 5 6\n"
```

```text
kornia.geometry.load_pointcloud_ply(p)
tensor([[10., 20.,  1.],
        [30., 40.,  4.]])          # expected [[1,2,3],[4,5,6]]
```

The byte-identical binary file raises properly:

```text
ValueError: PLY element 'vertex' in '...' has a list property 'tags';
list properties are only supported after the vertex element.
```

A silent wrong answer is worse than the rejection the binary path already gives, and this is a legal (if unusual) PLY. Cheapest fix that makes both readers agree: hoist the list check into `_ply_vertex_layout`, so the vertex element is rejected for both formats before any payload is touched. That keeps the binary behaviour unchanged and only loses the ASCII case where a list property trails x/y/z — which the binary reader already refuses anyway, so no consistency is lost. Worth a pin either way, since nothing in the new 23 covers a list property inside the vertex element on the ASCII path.

### 2. The deprecation message names a version that has already shipped

`kornia/geometry/pointcloud.py:241`:

```python
"`header_size` is ignored since kornia 0.8.3: ..."
```

`v0.8.3` is a tagged release; this tree is `0.9.0rc1`, so the deprecation is introduced in **0.9.0**. `kornia.core._compat.deprecated` documents `version` as "the kornia version when the deprecation was introduced", and the users who see this string will check it against their installed version. Please make it `0.9.0`.

### 3. The ASCII slowdown is mostly avoidable

Confirmed the regression against a `main` worktree — 200 000 points, kornia's own ASCII writer, torch 2.9.1, CPU, best of 3:

| | base `94d0284c` | this branch |
|---|---|---|
| ASCII | 74.5 ms | 127–131 ms |
| binary | 1.7 ms | 2.8 ms |

The binary cost is the header parse and is fine. The ASCII 1.7x is not the price of honouring the declared layout, it is the generator expression at `pointcloud.py:296`. Two measured variants on the same file:

| variant | time |
|---|---|
| current: `points.extend(float(tokens[i]) for i in xyz)` | 127.0 ms |
| hoist `ix, iy, iz = xyz`, `points.extend((float(tokens[ix]), float(tokens[iy]), float(tokens[iz])))` | **89.2 ms** |
| the above + bulk `f.read().split()` when `num_columns == 3 and xyz == (0, 1, 2)` | **68.6 ms** |

The first is a two-line change with no semantic difference whatsoever and recovers ~60% of the regression. The second reaches base speed on the layout kornia's own writer emits — but it does change one edge behaviour (the per-line loop currently tolerates a row with *more* tokens than declared properties, and a bulk token split would misalign on such a file), so if you take it, gate it and keep the per-line path for everything else. I would take at least the first.

### 4. Nit: one ASCII failure path does not name the file

Every other error in the new code names `filename`. A non-numeric coordinate escapes as the raw `float()` error:

```text
ValueError: could not convert string to float: b'oops'
```

Wrapping the row conversion to report the file and the offending line would keep the diagnostics uniform.

### 5. Nits

- `stacklevel=3` is correct for a direct call (verified: the warning is attributed to the caller). Through the deprecated aliases `kornia.load_pointcloud_ply` / `kornia.utils.load_pointcloud_ply` it lands on the forwarding wrapper instead — cosmetic only, since those wrappers already emit their own `DeprecationWarning`.
- ASCII coverage is thinner than binary: no `element vertex 0` case and no preceding-element skip. I verified both work by hand (`(0, 3)` float32, and a preceding scalar element is skipped line-by-line correctly), so these are pins for behaviour that is already right, not bugs.

---

Everything else I checked holds: `<`/`>` struct prefixes correctly disable alignment padding, `array.array` itemsizes match `struct.calcsize` for all eight typecodes in `_PLY_SCALAR_TYPES`, the byteswap condition is right in both directions, `float(bytes)` is valid so the ASCII path's byte tokens are fine, CRLF headers parse, `ushort`/`uint` correctly fall through to the `struct` path rather than a missing torch dtype, an over-declared `element vertex N` cannot over-allocate (`f.read` is bounded by the file), and NaN/inf round-trip through both writers unchanged.

ducha-aiki:
All five points from the review above are applied in `73333a30`.

**1. ASCII list property in the vertex element** — the guard moved into `_ply_vertex_layout`, so both readers now reject it before any payload is touched and give the same message. The file that used to return `[[10, 20, 1], [30, 40, 4]]` now raises:

```text
ValueError: PLY vertex element in '...' has a list property 'tags';
list properties are only supported in elements that follow the vertices.
```

Pinned for both readers (`test_rejects_list_property_in_vertex_element`); the ASCII parametrization fails on the previous head, the binary one pins the guard's new home.

**2. Deprecation version** — now `since kornia 0.9.0`.

**3. ASCII performance** — took the safe variant. Re-measured against a `main` worktree (`94d0284c`), 200 000 points, best of five, torch 2.9.1, CPU:

| | base `94d0284c` | first revision | now |
|---|---|---|---|
| ASCII | 74.5 ms | 127.0 ms | **89.3 ms** |
| binary | 1.7 ms | 2.8 ms | 2.3 ms |

Hoisting the indices and writing the three `float()` calls out recovers 38 of the 52 ms — the generator expression, not the layout handling, was most of the regression. I did **not** take the bulk token split (68.6 ms, at base speed): it only applies to the plain three-column layout, and it misaligns silently on a row carrying more tokens than the header declares, which the per-line loop rejects. Trading a rejection for a silent wrong answer is the same defect class as §1, so the residual 1.2x stands as the price of honouring the declared layout.

**4. Non-numeric ASCII coordinate** — now names the file and the vertex instead of escaping as a bare `float()` error, and is pinned.

**5. Nits** — added the two missing ASCII pins (empty cloud, element preceding the vertices); both pass on the previous head, so they cover behaviour that was already right. The alias `stacklevel` is left alone as cosmetic.

Docstrings for both readers and the CHANGELOG entry updated for the list-property rejection. Verification: `tests/geometry/test_pointcloud_io.py` **37 passed**; the two behaviour-changing pins fail with `kornia/geometry/pointcloud.py` reverted to the previous head; `pixi run pre-commit-all` all hooks pass; `pixi run typecheck` still shows only the pre-existing `lightglue` `custom_fwd` diagnostic.

ducha-aiki:
## Re-review of #4114 at `73333a30`

Round-1's five points are applied. I re-read the whole file rather than only the delta, re-ran the suite, and re-executed each claim in the round-2 comment instead of taking it on trust.

**Verified locally** (`.venv/bin/python`, torch 2.9.1, CPU, `kornia.__version__ = 0.9.0rc1`):

| check | result |
|---|---|
| `tests/geometry/test_pointcloud_io.py` | **37 passed** |
| the same 37 with `kornia/geometry/pointcloud.py` reverted to `886c3756` | **2 failed** — `test_rejects_list_property_in_vertex_element[load_pointcloud_ply]` (DID NOT RAISE) and `test_ascii_rejects_non_numeric_coordinate` (regex did not match). The three "already right" pins pass on the old head, exactly as claimed |
| `ruff check` + `ruff format --check` on both changed files | clean |
| §1 fix | the ASCII file that returned `[[10, 20, 1], [30, 40, 4]]` now raises `PLY vertex element in '...' has a list property 'tags'`, same message as the binary reader |
| §2 fix | `header_size` positionally (`load_pointcloud_ply(p, 8)`, the old call style) warns `... ignored since kornia 0.9.0 ...` — matches `0.9.0rc1`, and resolves Copilot's only comment, which was written against the `0.8.3` string |
| §4 fix | non-numeric coordinate now names the file and `vertex 1` |
| real-world layouts not in the suite | Open3D-style `float x/y/z + uchar r/g/b` → correct; homogeneous `float32` fast path → correct and contiguous; `int32` coords → correct; `ushort` correctly falls through to the `struct` path (`65535.0`, not a truncation) |
| my round-1 claim that an over-declared `element vertex N` cannot over-allocate | **holds** — `BufferedReader.read(10**11)` on a 5-byte file returns 5 bytes in 2.8 ms with zero peak-RSS delta, so `f.read(expected)` does not preallocate |
| `grep -rnE "#4113\|issues/4113\|#3973\|issues/3973" kornia/ tests/` | no hits — no inline prose left naming the closed issues |

One thing to fix before merge, then two smaller notes.

---

### 1. The CHANGELOG says the ASCII reader rejects something it accepts

`CHANGELOG.md:88`:

> ... and so is a file that declares a `list` property among the vertex properties **or in an element that precedes them** ...

The second half is true of the binary reader only. The ASCII reader skips preceding elements line-by-line, which is correct precisely *because* an ASCII list is still one line — so it reads such a file, and reads it right:

```python
hdr = (b"ply\nformat ascii 1.0\n"
       b"element face 1\nproperty list uchar int vi\n"
       b"element vertex 2\nproperty double x\nproperty double y\nproperty double z\nend_header\n")
body = b"3 0 1 2\n1 2 3\n4 5 6\n"
```

```text
kornia.geometry.load_pointcloud_ply(p)
tensor([[1., 2., 3.],
        [4., 5., 6.]])          # no ValueError
```

Both docstrings are accurate — `load_pointcloud_ply` claims only "a `list` property among the vertex properties is rejected", and `load_pointcloud_ply_binary` correctly scopes its stronger claim to elements before the vertices. It is only the CHANGELOG that merges the two into a restriction neither reader applies uniformly, and that clause is the one users read in the release notes. Splitting it — vertex properties for both readers, preceding elements for the binary reader — is a one-sentence edit.

### 2. The ASCII truncation diagnostic blames the wrong thing

A file that simply ends early reports a malformed line:

```text
ValueError: PLY file '...' declares 3 vertices with 3 properties, but a vertex line has 0 values.
```

The binary reader gets this right (`declares 2 vertices (48 bytes) but only 24 bytes follow the header`). `f.readline()` returning `b""` is EOF, not a short line, and the ASCII path already distinguishes the two for *preceding* elements ("ends inside element"). Worth the same treatment in the vertex loop, plus the row index the non-numeric error already carries — truncated ASCII clouds are a common enough real failure that "a vertex line has 0 values" will send people looking at the wrong file position.

### 3. The performance note argues from a behaviour the loop does not have

The PR body and the round-2 comment both justify skipping the bulk-token-split variant with: a row carrying more tokens than the header declares "misaligns silently ... which the per-line loop rejects". The per-line loop does not reject it — `len(tokens) < num_columns` only catches *short* rows, so an over-long row is read correctly and silently:

```text
header: 3 properties; body: b"1 2 3 99\n4 5 6\n"
-> tensor([[1., 2., 3.], [4., 5., 6.]])
```

The **decision** is still right — the bulk split would misalign on that file and the per-line loop would not — but the reason is "the per-line loop reads it correctly", not "rejects it". Since this sentence is the standing record for why the 1.2x ASCII cost was accepted, it is worth stating accurately.

### Nits, none blocking

- `pointcloud.py:363` — `table[:, list(xyz)].to(torch.float32).clone()`. Advanced indexing with a list already returns fresh storage, so the `clone()` is a second full copy of the output and nothing aliases `values` without it. Harmless, but it doubles peak copy on a large cloud for no benefit.
- Seven of the nineteen new raise sites are unpinned: `Unsupported PLY format line`, `Malformed PLY element line`, `PLY property before any element`, `Malformed PLY property line`, `has no 'format' line`, `declares no 'vertex' element`, and the ASCII `ends inside element`. Every branch a real malformed file is likely to hit *is* pinned, so this is coverage, not a defect — but the last two are the ones a hand-written header would actually reach.
- Unknown header keywords are hard-rejected. Defensible and I would keep it, given the `ply` first-line and `format` checks already screen non-PLY input; just noting that a file carrying one unknown directive is now unreadable where a `continue` would let it through.

---

Everything else I re-checked at this head holds: the `_ply_vertex_layout` guard is reached before any payload byte in both readers, `element_struct`'s own list check still covers preceding elements in binary, the byteswap condition is right in both directions, `array.array` itemsizes match `struct.calcsize` for all eight typecodes, `<`/`>` correctly suppress alignment padding, CRLF parses in both header and payload, `float(b"nan")`/`float(b"inf")` keep the writers' non-finite round-trip working (the writers' `isfinite(...).any(dim=1)` mask means the declared count always matches the rows written), and `typing.Dict/List/Tuple` matches the existing style in `kornia/geometry/`.

**Approve once §1 is corrected.** §2 and §3 are worth taking in the same push; the nits are optional.

ducha-aiki:
All three points from the re-review are applied in `fc632503`, plus the two nits.

**§1 — CHANGELOG scoped per reader.** The clause no longer claims the ASCII reader rejects a `list` property in an element that precedes the vertices, because it does not and should not: it skips such an element by lines, and an ASCII list is still one line. The entry now states the shared restriction (a list among the *vertex* properties, both readers) and the binary-only one separately. Both docstrings were already accurate and are unchanged. The same over-broad sentence in the PR description is corrected too.

**§2 — ASCII truncation now names itself.** `f.readline()` returning `b""` is EOF, not a short line, so it no longer reports one:

```text
before: PLY file '...' declares 3 vertices with 3 properties, but a vertex line has 0 values.
after:  PLY file '...' declares 3 vertices but the payload ends after 1 of them.
```

which parallels the binary reader's `declares 2 vertices (48 bytes) but only 24 bytes follow the header`. A genuinely short line keeps its own message and now names the row (`but vertex 0 has 2 values`), matching the non-numeric-coordinate error. A blank line inside the payload is still a malformed line, not EOF, and is reported as such.

**§3 — the performance justification.** "which the per-line loop rejects" is now "which the per-line loop reads correctly". The decision not to take the bulk token split is unchanged, and it is still the right one — the bulk split *would* misalign on an over-long row — but the loop tolerates that row rather than rejecting it, and the record should say so.

**Nits.** Dropped the `clone()` on the binary fast path. `table[:, list(xyz)]` is advanced indexing, so it already copies out of the transient `array.array` and the result cannot alias it; I checked that directly by mutating the returned tensor in place and re-reading the file, on both the little- and big-endian paths (`add_(100)` on the result, next read still `[[1, 2, 3], [4, 5, 6]]`). It saves one N x 3 copy — not measurable at 200k, but at 2M points binary goes **18.8 ms -> 17.2 ms** (min of 15, A/B against this branch's previous head). I left the strict unknown-keyword rejection alone, as flagged.

Four new pins, two of which discriminate: `test_ascii_rejects_truncated_payload` and the reworded `test_ascii_rejects_short_vertex_line` fail on `73333a30`; `test_ascii_rejects_truncated_preceding_element` and `test_rejects_missing_vertex_element` (both readers) cover branches that were already right but unpinned — five of the seven I listed remain unpinned.

**Re-measured, not carried over.** The ASCII EOF check costs about 2% in the hot loop, so I re-ran the A/B rather than reusing the old figure — 200k points, min of 7, against a `main` worktree (`94d0284c`), both arms run from the repo root with `kornia.__file__` printed:

| | base `94d0284c` | before | now |
|---|---|---|---|
| ASCII | 74.5 ms | 89.3 ms | 91 ms |
| binary | 1.7 ms | 2.3 ms | **1.8 ms** |

I also corrected a number I had asserted in the PR description without executing it: reverting `kornia/geometry/pointcloud.py` to `main` gives **31 failed, 10 passed** of 41, not the 29 I wrote. Thirty-one of the thirty-two new pins fail on base; the exception is `test_ascii_empty`, which passes there because the old loader's `torch.tensor([]).view(-1, 3)` is also `(0, 3)` — vacuously, not from a declared count. The description now says that instead of "every new pin fails on base", which was not true of the previous set either.

Verification at `fc632503`: `tests/geometry/test_pointcloud_io.py` **41 passed**, `tests/utils/test_deprecated.py` 30 passed, `pixi run pre-commit-all` all hooks pass, `pixi run typecheck` still only the pre-existing `lightglue` `custom_fwd` diagnostic.

ducha-aiki:
The last open item from the re-review is done in `466f76a1`: the five raise sites in `_read_ply_header` that no pin reached now have one.

Seven cases added to `test_binary_rejects_malformed_header`, which grows from 7 to 14 and now covers every rejection branch in the header parser:

| case | branch |
|---|---|
| `format binary_weird 1.0` | `Unsupported PLY format line` |
| no `format` line at all | `has no 'format' line` |
| `element vertex` (two tokens) | `Malformed PLY element line` |
| `element vertex 1 extra` (four tokens) | `Malformed PLY element line` |
| `property double w` before any `element` | `PLY property before any element` |
| `property quadruple x` (unknown scalar type) | `Malformed PLY property line` |
| `property list uchar tags` (a token short) | `Malformed PLY property line` |

I executed all seven against the branch before writing them into the parametrize list, to confirm each reaches the branch it names rather than tripping an earlier check — `element vertex 1 extra`, for instance, could plausibly have been swallowed by the count parse, and the missing-`format` case only reaches its check after the loop breaks at `end_header`.

These are coverage, not behaviour: they pass on `fc632503`, since the parser already had every one of these branches. Against `main` (`94d0284c`) all seven fail — **38 failed, 10 passed of 48**, measured by reverting `kornia/geometry/pointcloud.py` and re-running, not extrapolated from the previous count. The description's numbers are updated to match.

Nothing from either review round is left open. The one item I deliberately did not change is the strict rejection of unknown header keywords, as flagged — the `ply` first-line and `format` checks already screen non-PLY input, so the strictness costs only files carrying an unrecognised directive, and `test_binary_rejects_malformed_header`'s `frmt` case pins it.

Verification at `466f76a1`: `tests/geometry/test_pointcloud_io.py` **48 passed**, `pixi run pre-commit-all` all hooks pass.

PR #4117, opened 2026-08-30
fix(tests): probe for the kernel before running half-precision tests on old torch

### LINKED ISSUE #4116: Scheduled CPU tests are red on main: half-precision tests from #4098 hit missing torch 2.1.2/2.2.2 CPU kernels
### Describe the bug

The scheduled **Tests on CPU** workflow has been red on `main` since #4098 merged. Six matrix jobs fail — every `torch==2.1.2` and `torch==2.2.2` job, on both `float32` and `float64`. torch 2.3.1, 2.4.0, 2.5.1 and 2.9.1 are all green, as are macOS and Windows (they only run 2.5.1/2.9.1).

Last green scheduled run: [33291560607](https://github.com/kornia/kornia/actions/runs/33291560607) at `5f46e6da` (pre-#4098).
First red scheduled run: [33325938791](https://github.com/kornia/kornia/actions/runs/33325938791) at `a1919354` (the #4098 merge commit itself).
Also red: [33328615592](https://github.com/kornia/kornia/actions/runs/33328615592).

**torch 2.1.2** (py3.11) — 5 failures:

```
FAILED tests/feature/test_hardnet.py::TestHardNetConstantPatchIsFinite::test_constant_patch[cpu-HardNet-desc_dtype0]  - RuntimeError: "slow_conv2d_cpu" not implemented for 'Half'
FAILED tests/feature/test_hardnet.py::TestHardNetConstantPatchIsFinite::test_constant_patch[cpu-HardNet8-desc_dtype0] - RuntimeError: "slow_conv2d_cpu" not implemented for 'Half'
FAILED tests/feature/test_matching.py::TestMatchSteererGlobal::test_normalize_is_finite_on_a_zero_descriptor[cpu-desc_dtype0] - RuntimeError: "addmm_impl_cpu_" not implemented for 'Half'
FAILED tests/feature/test_matching.py::TestMatchSteererGlobal::test_normalize_is_finite_on_a_zero_descriptor[cpu-desc_dtype1] - RuntimeError: "eye" not implemented for 'BFloat16'
FAILED tests/feature/test_scale_space_detector.py::TestScaleSpaceDetector::test_fill_sentinel_follows_the_response_dtype[cpu-float32] - RuntimeError: "topk_cpu" not implemented for 'Half'
```

**torch 2.2.2** (py3.11 and py3.12) — 1 failure:

```
FAILED tests/feature/test_matching.py::TestMatchSteererGlobal::test_normalize_is_finite_on_a_zero_descriptor[cpu-desc_dtype1] - RuntimeError: "eye" not implemented for 'BFloat16'
```

### Root cause

Not a kornia defect: all four are **missing CPU kernels in old PyTorch**, reached by tests that hardcode a half-precision dtype instead of reading the injected `dtype` fixture. #4098 added half-precision regression tests (`F.normalize`'s 1e-12 eps underflows in float16, and the top-K sentinel must fit the response dtype) — correct tests, but they assume kernels that torch < 2.3 does not have on CPU.

Probed directly, with no kornia in the picture:

| op | torch 2.1.2 float16 | torch 2.1.2 bfloat16 | torch 2.2.2 float16 | torch 2.2.2 bfloat16 |
|---|---|---|---|---|
| `F.conv2d` | ❌ `slow_conv2d_cpu` | ✅ | ✅ | ✅ |
| `a @ b`    | ❌ `addmm_impl_cpu_` | ✅ | ✅ | ✅ |
| `torch.topk` | ❌ `topk_cpu` | ✅ | ✅ | ✅ |
| `torch.eye` | ✅ | ❌ `eye` | ✅ | ❌ `eye` |

That matches the failures exactly, including why 2.2.2 fails on only the one bfloat16 case.

### Reproduction

```bash
uv venv --python 3.11 venv212 && VIRTUAL_ENV=$PWD/venv212 uv pip install "torch==2.1.2" "numpy<2" pytest kornia_rs packaging
./venv212/bin/python -m pytest -q \
  "tests/feature/test_hardnet.py::TestHardNetConstantPatchIsFinite" \
  "tests/feature/test_matching.py::TestMatchSteererGlobal::test_normalize_is_finite_on_a_zero_descriptor" \
  "tests/feature/test_scale_space_detector.py::TestScaleSpaceDetector::test_fill_sentinel_follows_the_response_dtype"
# 5 failed, 5 passed
```

### Expected behavior

The scheduled CPU matrix is green across the whole supported torch range (`pyproject.toml` declares `torch>=2.0.0`).

### Proposed fix

`testing/base.py` already has the pattern for exactly this — runtime kernel probes (`supports_replicate_padding`, `supports_grid_sample`, `supports_2d_border_padding`), which skip rather than version-guard so a test auto-enables once PyTorch fills the kernel in. Extend it:

- add `supports_conv2d`, `supports_matmul` and `supports_topk` probes, and skip the three tests on builds that lack the kernel;
- for the bfloat16 `torch.eye`, build the identity in float32 and cast. 0 and 1 are exact in every float dtype, so the matrix is identical — and this keeps the bfloat16 case actually **running** on 2.1.2/2.2.2 rather than being skipped, which is where the underflow bug it guards would show.

Deliberately out of scope: kornia's own `_cdist` half-precision fallback (`kornia/feature/matching.py`) multiplies in half, so a user on torch 2.1.2 CPU hits `addmm_impl_cpu_` from library code, not just from tests. That is a PyTorch gap on a three-year-old release rather than a kornia regression; raising the floor or upcasting is a separate discussion.

### Environment

Failure is from CI (ubuntu, py3.11/3.12); reproduced locally on macOS arm64 with torch 2.1.2 and 2.2.2, same four kernel gaps and the same five/one failures.

ducha-aiki:
Fixes #4116.

## The failure

The scheduled **Tests on CPU** workflow has been red on `main` since #4098 merged — the last green run is [33291560607](https://github.com/kornia/kornia/actions/runs/33291560607) at `5f46e6da`, the first red one is [33325938791](https://github.com/kornia/kornia/actions/runs/33325938791) at the #4098 merge commit itself. Six matrix jobs fail: every `torch==2.1.2` and `torch==2.2.2` job, on both `float32` and `float64`. 2.3.1, 2.4.0, 2.5.1 and 2.9.1 are green.

## Root cause

None of the five failures is a kornia defect. All four distinct errors are **missing CPU kernels in old PyTorch**, reached by tests that hardcode a half-precision dtype instead of reading the injected `dtype` fixture. #4098 added correct half-precision regression tests (`F.normalize`'s 1e-12 eps underflows in float16; the top-K sentinel must fit the response dtype) that happen to assume kernels torch < 2.3 does not have.

Probed with no kornia in the picture:

| op | 2.1.2 f16 | 2.1.2 bf16 | 2.2.2 f16 | 2.2.2 bf16 | failing test |
|---|---|---|---|---|---|
| `F.conv2d` | ❌ `slow_conv2d_cpu` | ✅ | ✅ | ✅ | `TestHardNetConstantPatchIsFinite` (×2) |
| `a @ b` | ❌ `addmm_impl_cpu_` | ✅ | ✅ | ✅ | `TestMatchSteererGlobal` |
| `torch.topk` | ❌ `topk_cpu` | ✅ | ✅ | ✅ | `test_fill_sentinel_follows_the_response_dtype` |
| `torch.eye` | ✅ | ❌ `eye` | ✅ | ❌ `eye` | `TestMatchSteererGlobal` |

That accounts for the failures exactly, including why 2.2.2 fails on only the one bfloat16 case.

## The fix

`testing/base.py` already carries the pattern for this — runtime kernel probes (`supports_replicate_padding`, `supports_grid_sample`, `supports_2d_border_padding`) that **skip rather than version-guard**, so a test auto-enables the moment PyTorch fills the kernel in. This PR extends it:

- adds `supports_conv2d`, `supports_matmul` and `supports_topk`, and skips the three tests on builds that lack the kernel;
- factors the shared probe body (`_supports_kernel_probe`) out of the two existing dtype-keyed probes rather than copying it three more times — the two existing helpers keep their behaviour, their `@cache`ing and their docstrings, and their call sites are untouched;
- **fixes** the bfloat16 `torch.eye` rather than skipping it: 0 and 1 are exact in every float dtype, so building the identity in float32 and casting gives the identical matrix. That keeps the bfloat16 case actually *running* on 2.1.2 and 2.2.2 — which is where the underflow it guards would show — so `TestMatchSteererGlobal` gains coverage on old torch instead of losing it.

Test-only change; no library code and no public API is touched, so no CHANGELOG entry (same as #4088).

## Verification

Reproduced and verified against real old-torch installs, not simulated:

| torch | before | after |
|---|---|---|
| 2.1.2 | **5 failed**, 5 passed | 6 passed, 4 skipped |
| 2.2.2 | **1 failed**, 9 passed | **10 passed** |
| 2.9.1 (`tests/feature`, float32+float64) | 837 passed | 837 passed, 0 new skips |

Every probe returns `True` on torch 2.9.1 CPU for float16/bfloat16/float32, so no coverage is lost on a modern build. The refactored `supports_replicate_padding`/`supports_grid_sample` were re-exercised through their existing call sites (`test_siftdesc.py`, `test_local_features_orientation.py`, `test_elastic_transform.py`, `test_thin_plate_spline.py`) on both 2.1.2 (skips fire correctly) and 2.9.1 (all pass).

Worth flagging for review: **this PR's own CI cannot validate the fix.** `pr_test_cpu.yml` runs only `2.5.1` and `2.9.1`; `2.1.2` and `2.2.2` exist solely in `scheduled_test_cpu.yml`. A green check here means "not regressed on modern torch", and the before/after numbers above are the actual evidence — confirmation comes from the next scheduled run after merge.

`pixi run pre-commit-all` passes; `pixi run typecheck` is unchanged (its one pre-existing `lightglue.py` deprecation warning remains).

One aside, not addressed here: `DiscreteSteerer` is reachable only as `kornia.feature.steerers.DiscreteSteerer` — it is not exported from `kornia.feature` and not in `docs/source/feature.rst`, despite being a documented-style public class. Separate from this fix.

### DIFF
--- a/testing/base.py
+++ b/testing/base.py
@@ -141,21 +141,28 @@ _UNIMPLEMENTED_KERNEL_MSG = "not implemented for"
 
 
 @cache
-def _supports_replicate_padding_probe(device_type: str, dtype: torch.dtype) -> bool:
+def _supports_kernel_probe(op: Callable[[str, torch.dtype], object], device_type: str, dtype: torch.dtype) -> bool:
+    """Run ``op`` on a tiny tensor and report whether this build has a kernel for it.
+
+    ``op`` allocates its own probe tensors and then calls the operator, because a device that
+    cannot hold ``dtype`` at all (MPS and float64) raises TypeError while allocating, which is
+    still "unsupported" and must not escape a helper whose whole contract is to return a bool.
+    Cached per (op, device type, dtype), so every probe below runs once per session and
+    auto-enables once PyTorch fills the kernel in.
+    """
     try:
-        # Allocated outside the `try` below: a device that cannot hold `dtype` at all (MPS and
-        # float64) raises TypeError here, which is still "unsupported" and must not escape a
-        # helper whose whole contract is to return a bool.
-        probe = torch.zeros(1, 1, 2, 2, device=device_type, dtype=dtype)
+        op(device_type, dtype)
     except TypeError:
         return False
-    try:
-        F.pad(probe, (1, 1, 1, 1), mode="replicate")
     except RuntimeError as e:
         if _UNIMPLEMENTED_KERNEL_MSG in str(e):
             return False
         raise
     return True
+
+
+def _replicate_padding_op(device_type: str, dtype: torch.dtype) -> None:
+    F.pad(torch.zeros(1, 1, 2, 2, device=device_type, dtype=dtype), (1, 1, 1, 1), mode="replicate")
 
 
 def supports_replicate_padding(device: torch.device, dtype: torch.dtype) -> bool:
@@ -167,23 +174,15 @@ def supports_replicate_padding(device: torch.device, dtype: torch.dtype) -> bool
     than reading the injected dtype fails there on every job. Probed at runtime and cached per
     (device type, dtype), so it auto-enables once PyTorch fills the kernel in.
     """
-    return _supports_replicate_padding_probe(device.type, dtype)
+    return _supports_kernel_probe(_replicate_padding_op, device.type, dtype)
 
 
-@cache
-def _supports_grid_sample_probe(device_type: str, dtype: torch.dtype) -> bool:
-    try:
-        probe = torch.zeros(1, 1, 2, 2, device=device_type, dtype=dtype)
-        grid = torch.zeros(1, 2, 2, 2, device=device_type, dtype=dtype)
-    except TypeError:
-        return False
-    try:
-        F.grid_sample(probe, grid, align_corners=False)
-    except RuntimeError as e:
-        if _UNIMPLEMENTED_KERNEL_MSG in str(e):
-            return False
-        raise
-    return True
+def _grid_sample_op(device_type: str, dtype: torch.dtype) -> None:
+    F.grid_sample(
+        torch.zeros(1, 1, 2, 2, device=device_type, dtype=dtype),
+        torch.zeros(1, 2, 2, 2, device=device_type, dtype=dtype),
+        align_corners=False,
+    )
 
 
 def supports_grid_sample(device: torch.device, dtype: torch.dtype) -> bool:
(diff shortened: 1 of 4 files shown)


ducha-aiki:
## Review — verified against real torch 2.1.2 / 2.2.2, both directions

I reproduced the scheduled-run failure locally rather than reading the log, then re-ran with the branch applied. The diagnosis and the fix both hold.

**Failure reproduced exactly.** Job [99296021678](https://github.com/kornia/kornia/actions/runs/33325938791) (`float32, 3.11, 2.1.2`) fails on 5 tests; a real 2.1.2 CPU install reproduces all 5 with the same four errors, and 2.2.2 reproduces the single `"eye" not implemented for 'BFloat16'`:

| torch | before (`main`) | after (this branch) |
|---|---|---|
| 2.1.2 | 5 failed — `slow_conv2d_cpu`×2, `addmm_impl_cpu_`, `eye`/BFloat16, `topk_cpu` | 0 failed; 9+93+61+40+27 passed across the five touched files |
| 2.2.2 | 1 failed — `eye`/BFloat16 | 0 failed; 11+94+62+40+27 passed |
| 2.9.1 | — | 493 passed, 48 skipped across every probe call site; no new skips |

Every probe returns `True` on 2.9.1 CPU for float16/bfloat16/float32/float64, so nothing is silently lost on a modern build. On a real MPS device all five return `False` for float64 and `True` otherwise, without raising — the allocation path the shared helper is built around still behaves.

Two docstring claims I checked rather than assumed: `_cdist`'s half-precision fallback really does matmul (`d1 @ d2.t()`, `kornia/feature/matching.py:44`), and `DiscreteSteerer.forward` really is `F.linear` (`kornia/feature/steerers.py:57`) — so `supports_matmul` names the right op for its call site.

### One real defect, now fixed in `936a797a`

Factoring the shared body moved tensor allocation *inside* the `try` that returns `False`, so the `except TypeError` no longer covered only allocation — it covered the whole `op` call. Demonstrated:

```python
def bad_op(device_type, dtype):
    torch.zeros(2, device=device_type, dtype=dtype).topk(kk=1)   # typo'd kwarg

_supports_kernel_probe(bad_op, "cpu", torch.float32)   # -> False
```

A mis-written probe reports "this build has no kernel" and skips its guarded test on **every** torch version, forever, with a plausible-looking skip reason. That is the same latent test-rot class as a shape-only assertion on an all-zero result: green, and testing nothing. `main`'s pre-refactor code did not have it, because allocation sat in its own `try`.

Fixed by keeping the two apart rather than by reverting the factoring: `_probe_zeros` translates only the allocation `TypeError` (MPS/float64) into a private `_UnsupportedProbeDtype`, which is what `_supports_kernel_probe` catches. After the fix the same `bad_op` raises `TypeError: topk() missing 1 required positional arguments: "k"`, while a simulated *and* a real MPS-float64 allocation still returns `False`. All the numbers in the table above are post-fix.

### Nits

- `supports_matmul` probes 2-D `@`, i.e. `mm`/`addmm`. Batched `bmm` is a separate kernel that can be missing independently, and the generic name invites reuse for it — `tests/geometry/test_linalg.py:85` already guards a CPU float16 **`bmm`** gap by version. Noted in the docstring in `936a797a`; not otherwise changed.
- Copilot's inline note on `_matmul_op` is half right. The line was 107 chars against a 120 limit and both `ruff check` and `ruff format` passed on it, so the "likely to trip formatting/line-length tooling" part is not accurate. The readability half is fair, and `_probe_zeros` shortens it to 79 chars as a side effect.

### Follow-up, deliberately not in this PR

Two half-precision tests still skip by torch version rather than by probe, which is exactly the pattern this PR's rationale argues against:

- `tests/augmentation/container/test_augmentation_sequential.py:115` — `torch_version_lt(2, 6, 0)` for CPU half `grid_sample`; `supports_grid_sample(device, image_dtype)` already covers this, and would also cover MPS, which the hardcoded `device.type == "cpu"` misses.
- `tests/geometry/test_linalg.py:85` — `torch_version_lt(2, 2, 0)` for CPU float16 `bmm`; needs a batched probe, not `supports_matmul`.

Neither is failing, so converting them is a separate change.

### Verdict

**Approve.** The root cause is right — none of the five failures is a kornia defect — and fixing the bfloat16 `eye` by building in float32 and casting (0 and 1 are exact in every float dtype) is better than skipping: on 2.2.2 that case now runs and passes, so `TestMatchSteererGlobal` gains old-torch coverage instead of losing it.

The PR body's own caveat is worth repeating: `pr_test_cpu.yml` runs only 2.5.1 and 2.9.1, so a green check here cannot validate this fix. The evidence is the table above; confirmation is the next scheduled run after merge.

PR #4118, opened 2026-08-30
Fix transform bbox input mutation

Marchematics:
Fix `transform_bbox` mutating `xywh` input tensors in place.

Tests: `tests/geometry/test_bbox_input_mutation.py` verifies the transformed output and that the input tensor remains unchanged. Kornia CI passed.


ducha-aiki:
**Your fix is correct and verified.** But #4129, opened ~15 h later, contains a **byte-identical** production change, and I want you to hear the disposition from me rather than find out from a close notification.

Setup: worktree at this head `6c26f09c`, with `kornia/` swapped to the merge-base `2fbd1a0c`; `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS.

**Verified:**

```
# kornia/ at the merge-base, your test kept
FAILED tests/geometry/test_bbox_input_mutation.py::test_transform_bbox_xywh_keeps_input - assert False
# at the head
27 passed
```

and across 12 configurations (both modes x `restore_coordinates` in `{None, True, False}` x `(5,4)` / `(3,5,4)`) the input is no longer mutated, outputs, shapes and dtypes are unchanged, and gradients still flow through the clone. The single clone in the `xywh` branch really is the whole fix: `xyxy` never mutated, and the trailing `-=` writes into a tensor that is already fresh.

Git confirms the duplication literally — both diffs read `index e83ddae288..0341193258`, the same post-image blob.

**Why the other one:** the only difference is the test. #4129 adds its assertion to the existing `TestTransformBoxes2D::test_transform_boxes_wh`, so it inherits the `device`/`dtype` fixtures and runs on every backend and dtype the suite covers; yours is a new module-level file pinned to CPU float32 with no fixtures, which is the shape AGENTS.md and TESTING.md steer away from, and it puts bbox coverage in a second file.

Since the shipped code is identical either way, I am taking the better-pinned test and closing this PR — **not** because the work was wrong. It was first and it was right.

If you would rather keep this PR as the one that lands, say so and instead move the assertion into `TestTransformBoxes2D::test_transform_boxes_wh` (two lines: snapshot `boxes` before the call, `assert torch.equal(boxes, boxes_before)` after) and delete the new file; I will merge this one and close #4129 instead. Either outcome is fine by me — I would just rather not merge both.

Marchematics:
I would like to keep #4118 as the PR that lands. I moved the input-preservation assertion into the existing fixture-based test and removed the standalone test file. Please merge #4118 and close #4129.

ducha-aiki:
Sorry, I didn't notice your follow-up, when merging other PR, my bad. Closing this, as now this PR is no-op.

PR #4119, opened 2026-08-30
feat(benchmarks): add LAF-ops microbenchmark for kornia.feature

ducha-aiki:
Phase (c) of the `kornia.feature` audit (perf): `benchmarks/feature/` held one Oxford-Affine accuracy script and no speed microbenchmarks, so the LAF operations that every kornia detector and descriptor runs through had no measured baseline. This PR adds the harness, a current set of numbers, and the committed result files.

## What's added

- **`benchmarks/feature/laf_ops.py`** — microbenchmarks for 7 public LAF ops (`laf_from_center_scale_ori`, `make_upright`, `ellipse_to_laf`, `laf_to_boundary_points`, `laf_is_inside_image`, `extract_patches_simple`, `extract_patches_from_pyramid`), following the `common.py` methodology contract: `blocked_autorange` medians + IQR, pinned seeds, sync inside the timed region, recorded metadata, `--json`/`--contribute` export, and compile-warmup failures reported as a NOTE instead of silent skips. Throughput counts LAFs/s (`B*N` per call). No cross-library column — no other library exposes LAFs; the columns are kornia eager vs `torch.compile`.
- **The CPU and CUDA runs, committed** under `benchmarks/results/0.9.0rc1/`. These are the first committed non-Apple result set, so the docs performance page shows CUDA numbers for the first time.
- **`benchmarks/common.py`** — `run_batch_sweep` gains `label_fn`, `items_fn` and `units`, all defaulted to its current behavior, so a suite whose config is not a single batch size (a `(B, N)` pair here) can use the shared sweep instead of copying it. The four existing callers are unchanged; `common_test.py` covers the new path.
- **`docs/generate_benchmarks.py`** — see "Committing the results" below.
- **`benchmarks/README.md`** — directory-map row plus a sample-results section, regenerated from the committed JSON so the prose cannot drift from the data.

## Results

Linux/WSL2, Intel i7-14700K + NVIDIA RTX 4090, Python 3.13, torch 2.14.0+cu130, CUDA 13.0, kornia 0.9.0rc1, float32, image 256×256, patch size 32, 14 threads, LAF scales stratified across all four pyramid levels. Throughput in LAFs/s, higher is better. Every op compiled on both devices on this stack, so there is no failure cell. Compare columns within one table, never numbers across machines.

`--device cpu --compile`, B=1 N=20000:

| op | kornia (eager) | kornia (compiled) |
| --- | --: | --: |
| laf_from_center_scale_ori | 39273029 | **99964263** |
| make_upright | 33959226 | **122176202** |
| ellipse_to_laf | 110509852 | **149762253** |
| laf_to_boundary_points | 3613236 | **9903714** |
| laf_is_inside_image | 6325084 | **11126178** |
| extract_patches_simple | **53309** | 27357 |
| extract_patches_from_pyramid | **37758** | 36983 |

`--device cuda --compile`, B=1 N=20000:

| op | kornia (eager) | kornia (compiled) |
| --- | --: | --: |
| laf_from_center_scale_ori | 147968107 | **254559803** |
| make_upright | 74905012 | **378282785** |
| ellipse_to_laf | 119929997 | **447807983** |
| laf_to_boundary_points | 14107628 | **17835708** |
| laf_is_inside_image | 28078872 | **56037949** |
| extract_patches_simple | 3499935 | **31880283** |
| extract_patches_from_pyramid | 2979848 | **25969422** |

The committed files carry all three configs (B=1 N=2000, B=1 N=20000, B=8 N=2000); the tables above are the middle slice.

## The honest reading

- **Patch extraction still dominates, but only on CPU.** `extract_patches_from_pyramid` eager is ~38k LAFs/s on CPU, ~900× below `make_upright`, so a 20k-keypoint descriptor pass pays ~0.53 s in patch sampling before any descriptor math. On CUDA the same op reaches ~26M LAFs/s compiled (0.77 ms for those 20k LAFs), only ~25× below the cheap ops. Batched GPU extraction is the regime to be in.
- **Found weak spot: `torch.compile` is a 1.9–2.1× *loss* for `extract_patches_simple` on this x86 CPU** in all three configs, while it is a ~9.1× win for the same op on CUDA — and a **7.2× win on the Apple M1 CPU** (see the M1 tables below), so this is a property of that inductor path and stack, not of compiling the op for CPU. It is not the N-chunking from #4128: at B=1 N=2000 the sampling grid is ~16 MiB against a 64 MiB budget, so that config takes the single-chunk fast path and still loses.
- **Found weak spot: `laf_to_boundary_points` is the weakest compiled op on CUDA** — 17.8M LAFs/s, below both patch extractors, ~25× below `ellipse_to_laf`, and the least moved by `torch.compile` there (1.26×). It builds its `linspace`/`ones`/`tensor` basis without a `device=` argument and `.to()`s the result every call, so each call pays a host→device transfer. First optimization target in this file.
- **Small N on CUDA is launch-latency bound**, not work bound: at B=1 N=2000 the three cheapest ops all land within 8.2–14.2M LAFs/s regardless of what they compute, ~8.5–11× below their own N=20000 figures. Read the N=20000 rows for kernel cost and the N=2000 rows for per-call overhead.
- **Pyramid depth is currently free.** Stratifying LAF scales across all four levels instead of leaving every LAF on level 0 moved `extract_patches_from_pyramid` by ~3%: #4128's packed atlas pays for the whole atlas whatever the LAFs select. That is a property of today's implementation, not of the op.
- **`ellipse_to_laf` is now the fastest LAF op measured** (110.5M LAFs/s eager on CPU). This benchmark originally recorded it as a hot spot; the closed-form 2×2 inverse in #4122 fixed it.
- **MPS is re-measured** on an Apple M1, after #4122 and #4128; see the M1 section below. On Apple silicon MPS *loses* to the same machine's CPU on every cheap LAF op and only wins on patch extraction.

## Committing the results

Three things had to be fixed before the files could be committed honestly:

- `docs/generate_benchmarks.py` keyed a table row on `(op, batch)`. That is unique for the flagship suites, which sweep batch size alone, but this suite sweeps `(B, N)`: B=1 N=2000 and B=1 N=20000 share a batch, so one config would have silently overwritten the other in both the rendered table and the llms digest. A row is now identified by every config field that varies within the file — the flagship suites keep their existing `op @ batch` labels unchanged, and this suite renders `make_upright @ 1 (n_lafs=20000)`.
- The page hardcoded "throughput in items/s", and the item here is a LAF, not an image. The label now comes from an optional `units` metadata key, defaulting to `items/s` when a suite does not set one.
- The page and the digest told readers to reproduce a committed run with `python benchmarks/<suite>/flagship.py`. True while every committed suite was a flagship one, but this suite is `benchmarks/feature/laf_ops.py`, so a reader following that line for the page's only CUDA rows would look for a file that does not exist. Both now point at `benchmarks/<dir>/<script>.py` and name the README as the suite-to-script map.

Both result files use one `--machine-slug` for the box; the auto-detected name would have been `intel-r-core-tm-i7-14700k` for the CPU run and `nvidia-geforce-rtx-4090` for the CUDA run — two names for one machine.

## Review feedback (Codex)

All four points were real. Replies are on the individual threads; summary:

- **[P1] Import kornia from the checkout being measured** — fixed. The script now puts its own checkout root at `sys.path[0]` before importing kornia, records the resolved path as `kornia_module` in the exported metadata, prints it, and warns loudly if kornia resolved elsewhere. Verified: invoked by absolute path from an unrelated cwd with no `PYTHONPATH`, it now loads this checkout's kornia instead of the primary one. (`ellipse_to_laf.py` still has the same latent pitfall; left alone to keep this PR scoped.)
- **[P2] Exercise nonzero pyramid levels** — fixed, after reproducing the all-level-0 workload. Scales are stratified round-robin across the levels the extractor actually builds; histogram is uniform (5000/level at N=20000, size 256) and follows `--size`. Costs ~3% on the pyramid extractor, because #4128's packed atlas is level-insensitive — which is the argument for the change, since the old workload could not tell that apart from a level-skipping implementation.
- **[P2] Preserve compile failures in machine-readable results** — fixed in `common.run_batch_sweep`, so every suite benefits: an unmeasured backend now emits a row with `null` timings and a structured `error` field (the exception type for a compile-warmup failure, `"unavailable"` otherwise).
- **[P2] Make `n_lafs` part of the contributed-result identity** — already fixed in this branch; hit independently while contributing the files. See "Committing the results".

## Changed since the first review

Rebased onto `main`, which made parts of the original PR stale:

- The sample tables were measured at `2009933e`, before #4122 and #4128 rewrote `ellipse_to_laf` and patch extraction — both weak spots the section reported as open are fixed, so it described a library that no longer exists. Re-measured on torch 2.14 for CPU and CUDA.
- `laf_ops.py` carried its own copy of the sweep loop, which had since landed in `common.py` as `run_batch_sweep`. Moved onto the shared helper.
- Fixed a measurement bias: the script compiled each config in sequence without `torch._dynamo.reset()`, so the first config was timed on a static-shape graph and the later ones on the automatic-dynamic recompile — the bias `ellipse_to_laf.py` already documents. The reset goes once per config, outside the op loop, since it invalidates every compiled callable and a per-op reset would push earlier ops' recompiles into the timed region.
- Added `--dtype` and the `# kornia module:` header line, matching `ellipse_to_laf.py`.

## Checked

- `pre-commit` clean on the touched files, including `pretty-format-json` on the two committed result files.
- `pytest tests/test_generate_benchmarks.py tests/test_benchmark_results.py benchmarks/common_test.py benchmarks/results_schema_test.py` — 40 passed. That covers schema validation of the committed files and `test_committed_digest_is_fresh`, so `llms-full.txt` is refreshed in this PR.
- Rendered `performance.rst` inspected: all 21 configs per device survive with `n_lafs` disambiguation and the `LAFs/s` label. The landing-page hero chart reads the `augmentation` suite only and is unaffected.
- The four existing `run_batch_sweep` callers still start (`--help`) with the widened signature.

The one red job, `tests (macos-latest, float32, 3.12, 2.5.1)`, is a SAM ONNX tracing timeout unrelated to this PR — this branch touches no `kornia/` code and no `tests/models/`. #4198 skips it; once that lands, re-running this job picks it up.

---

## Round 2: review, fixes, and the Apple M1 numbers

Reviewed again on 2026-09-03, and the requested M1 measurements are committed.

### Review findings on the previous head (`208fa64c`), all fixed in `20519a32`

- **[P1] The two committed result files leaked an absolute path.** `metadata.kornia_module` was `/tmp/claude-1000/-home-ducha-dev-kornia/<session-uuid>/scratchpad/laf-wt/kornia/__init__.py` in both files — a home-directory layout and a session UUID, committed to a public repo. `benchmarks/README.md` promises contributors the opposite: "Only aggregate load numbers (load average, memory) are recorded in the file". Recording *which* checkout was measured is the right fix for the Codex P1; exporting it as an absolute path is not. The export is now checkout-relative (`kornia/__init__.py`, or `outside-checkout`), which still answers the only question a reader of the file has — did kornia resolve to the tree the commit names? — while the absolute path stays on the console, where it is the useful diagnostic. Both committed files are scrubbed to the value the fixed script would have written, and `results_schema.validate_result` now rejects an absolute path in *any* string metadata value, so the next suite to add a key cannot reintroduce it.
- **[P2] `git_commit()` recorded a hash that did not describe the measured tree.** Both committed files record `9517c982`, but the harness edits that produced them — the level stratification and the `sys.path` fix — were still uncommitted at run time and landed in `208fa64c`. Checking out the recorded commit and re-running gives the *pre-stratification* generator, i.e. different pyramid numbers. `git_commit()` now suffixes `-dirty` when tracked files differ from HEAD and `contribute_result` warns; untracked files are ignored, since `--contribute` writes one into the checkout itself. This affects every suite, not only this one. The M1 files below were measured from a clean tree and record a bare hash.
- **[P3] `test_render_page_keeps_configs_that_share_a_batch` had a vacuous assertion.** `assert "20" in rst` passes on the `2026-09-03` timestamp above the table even when the two B=1 configs collapse onto one row — only the `n_lafs=` assertions actually discriminated. It now pins the rendered cells exactly.

Checked and found correct: `pyramid_levels()` matches the real halving loop in `extract_patches_from_pyramid`, including the `max_level` cap and the `min(h, w) <= 2` guard; the stratification math is right (level *l* covers scale `[PS/2·2^l, PS/2·2^(l+1))`, sub-level-0 scales clamp to 0, giving exactly 5000 LAFs/level at N=20000); `make_ellipses` stays positive-definite; and the new `error` rows validate against the committed-result schema and render as `—` on the docs page.

### Apple M1 (MacBook Air, 8 cores, 4 torch threads), macOS 26.5, torch 2.9.1, Python 3.11

Same float32 / 256×256 / PS=32 / stratified-scale workload, B=1 N=20000 slice. Every op compiled on both devices here too, so there is no failure cell. Different architecture, thread count and torch version from the x86 box, so these answer "does the finding hold?", not "which box is faster?".

`--device cpu --compile`:

| op | kornia (eager) | kornia (compiled) |
| --- | --: | --: |
| laf_from_center_scale_ori | 25551758 | **84299262** |
| make_upright | 25171744 | **108229794** |
| ellipse_to_laf | 75817885 | **177777778** |
| laf_to_boundary_points | 838245 | **1078421** |
| laf_is_inside_image | 4880514 | **9227220** |
| extract_patches_simple | 40806 | **294408** |
| extract_patches_from_pyramid | 30084 | **200881** |

`--device mps --compile`:

| op | kornia (eager) | kornia (compiled) |
| --- | --: | --: |
| laf_from_center_scale_ori | 13318018 | **22368750** |
| make_upright | 12793187 | **35554228** |
| ellipse_to_laf | 14233594 | **39990002** |
| laf_to_boundary_points | 1282786 | **1874608** |
| laf_is_inside_image | 2194321 | **4115015** |
| extract_patches_simple | 88430 | **476392** |
| extract_patches_from_pyramid | 73561 | **689227** |

What the M1 adds:

- **The CPU compile regression is not a CPU property.** `extract_patches_simple` compiles to a **7.2× win** here (40806 → 294408) and 6.7× for the pyramid variant, against a 1.9–2.1× loss on the i7-14700K. A fix should bisect stack against machine before reading the kernel.
- **On Apple silicon, MPS is the wrong device for the cheap LAF ops.** Every bookkeeping op is faster in M1 *CPU* eager than in MPS eager — `ellipse_to_laf` by 5.3× (75.8M vs 14.2M LAFs/s), `make_upright` and `laf_from_center_scale_ori` by ~2× — and the gap widens under `torch.compile` (4.4× and 3.8×). MPS only wins where there is real work per LAF: patch extraction, by 2.2–2.4× eager and up to 3.4× compiled. A pipeline that ships LAFs to the GPU for the frame math alone pays for the transfer twice.
- **`laf_to_boundary_points` is the worst op on all three devices, and the M1 CPU says why.** `torch.compile` recovers only 1.29× because the cost is not a kernel: the 50-point basis is built with no `device=`, `.expand()`ed to `(B*N, n_pts, 3)`, and only *then* `.to(device)`d, so every call materializes — and on a GPU transfers — a tensor that scales with the LAF count: ~12 MiB per call at N=20000 for a basis with 50 distinct rows. Building it on the target device and broadcasting in the matmul is the fix; that is the first optimization target in this file.
- **Measurement noise floor.** Two independent full runs per device: CPU cells move by a median of 1.0% (worst 19%, cheapest op at the smallest config), MPS cells by a median of 7.4% (worst 54%). Any MPS difference under ~2× on this class of machine is noise.

### Checked

- `pytest benchmarks/common_test.py benchmarks/results_schema_test.py tests/test_generate_benchmarks.py tests/test_benchmark_results.py` — 45 passed, including the refreshed `test_committed_digest_is_fresh` and schema validation of the two new files.
- `pre-commit` clean on every touched file, `pretty-format-json` included.
- Rendered `performance.rst` inspected: four `feature-laf-ops` sections (apple-m1 cpu/mps, i7-14700k-rtx-4090 cpu/cuda), 84 `n_lafs`-disambiguated rows, `LAFs/s` label on each.
- `grep` over `benchmarks/results/` finds no remaining absolute path.

Note: the branch is 2 commits behind `main` (#4197, #4198) and needs a rebase before merge; neither touches the LAF path, so the numbers are unaffected.

Reviewed and measured by Claude (Opus 5) on behalf of @ducha-aiki


benchmarks/feature/laf_ops.py
```diff
@@ -0,0 +1,218 @@
...
+
+import argparse
+import math
+import sys
+from pathlib import Path
+from typing import Callable, Optional
+
+import torch
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[1]))
```
ducha-aiki, line 68: **[P1] Import kornia from the checkout being measured**

Only `benchmarks/` is added to `sys.path`, so when this script is run from a secondary worktree, the editable finder can still resolve `kornia` to the primary checkout while `git_commit()` reports the worktree HEAD. I reproduced this from a worktree at `6cd272a1`: the header reported `6cd272a1`, but `python -v` showed `kornia` loading from the primary checkout, which was on another branch. That can silently measure the same implementation on both sides of the optimization comparisons this harness is meant to support. Add the repository root (`parents[2]`) to `sys.path` before importing `kornia`, and ideally record/assert `kornia.__file__` in the preflight metadata.

benchmarks/feature/laf_ops.py
```diff
@@ -0,0 +1,218 @@
...
+import kornia.feature as KF
+
+Backend = Optional[Callable[[], object]]
+
+PATCH_SIZE = 32
+
+
+def make_lafs(b: int, n: int, size: int, device: torch.device, dtype: torch.dtype) -> torch.Tensor:
+    """Random but realistic LAFs: centers inside the image, scales 4-24 px, any orientation."""
+    xy = PATCH_SIZE + torch.rand(b, n, 2, device=device, dtype=dtype) * (size - 2 * PATCH_SIZE)
```
ducha-aiki, line 114: **[P2] Exercise nonzero pyramid levels**

With `PATCH_SIZE=32` and scales in `[4, 24)`, `extract_patches_from_pyramid` always computes `pyr_idx = floor(log2(2*scale/32)) = 0`; none of these supposedly realistic LAFs selects a deeper pyramid level. The current implementation still samples every level, but any follow-up that skips unused levels would appear disproportionately fast on this all-level-zero workload. Use a stratified distribution that covers levels 0 through `max_level - 1` (or separate per-level configs) so the baseline represents actual pyramid extraction.

benchmarks/feature/laf_ops.py
```diff
@@ -0,0 +1,218 @@
...
+            print(f"# NOTE: torch.compile warmup failed ({', '.join(exc_names)}) for: {', '.join(compile_failures)}")
+        header = f"{f'B={b} N={n}':<{label_width}}" + "".join(f"{be[:col_width]:>{col_width + 1}}" for be in backends)
+        print("-" * len(header))
+        print(header + "   (LAFs/s)")
+        print("-" * len(header))
+        for op_name, row in ops.items():
+            cells = []
+            for backend in backends:
+                fn = row.get(backend)
+                if fn is None:
```
ducha-aiki: **[P2] Preserve compile failures in machine-readable results**

This `continue` omits the compiled row entirely when warmup fails. The JSON/contribution output therefore cannot distinguish "compiled backend was not requested" from "compile failed", despite the benchmark schema documenting failed measurements as `null`; the exception type survives only in transient console output. Append a row for this backend/config with non-finite timings (sanitized to `null`) and, preferably, a structured failure field.

benchmarks/feature/laf_ops.py
```diff
@@ -0,0 +1,218 @@
...
+                    cells.append(f"{'-':>{col_width + 1}}")
+                    continue
+                median, iqr = time_us(fn, min_run_time=min_run_time, sync=sync)
+                thr = (b * n) / (median * 1e-6) if not math.isnan(median) else float("nan")
+                results.append(
+                    {
+                        "op": op_name,
+                        "backend": backend,
+                        "batch": b,
+                        "n_lafs": n,
```
ducha-aiki: **[P2] Make `n_lafs` part of the contributed-result identity**

The default sweep emits both `B=1,N=2000` and `B=1,N=20000`, but `docs/generate_benchmarks.py` keys cells only by `(op, batch, backend)` and labels rows as `op @ batch`. Once this suite is used with `--contribute`, the second B=1 measurement silently overwrites the first in the generated table, and readers cannot tell which `N` was rendered. Extend the renderer/table identity to include `n_lafs` (and display `B/N`), or encode the full config in an existing unique field before advertising contributed output.

ducha-aiki (commented):
I found four issues that should be addressed before relying on this harness for before/after performance evidence. CI is green and the script smoke-tests successfully, but the worktree/import, workload, and export problems below can produce misleading or lossy benchmark results.

benchmarks/feature/laf_ops.py
ducha-aiki: Fixed in 208fa64c. Confirmed the failure mode and closed it three ways:

- The script now inserts its own checkout root (`parents[2]`) at `sys.path[0]` before importing kornia, ahead of any editable install.
- It records the resolved path in the exported metadata as `kornia_module` (so a contributed file carries it), and prints it in the header.
- If kornia still resolves outside the checkout, it prints an explicit `WARNING: kornia resolved outside <root> - these numbers are NOT this checkout's.` I chose a warning rather than an assert so benchmarking a deliberately-installed kornia stays possible; the point was that the mismatch must not be silent.

Verified by invoking the script by absolute path from an unrelated cwd with no `PYTHONPATH`: it now loads the worktree's `kornia/__init__.py`, where before it loaded the primary checkout's.

Note `benchmarks/feature/ellipse_to_laf.py` still has the same latent pitfall — it documents the `PYTHONPATH="$PWD"` workaround instead. I left it alone to keep this PR scoped, but it should get the same treatment.

benchmarks/feature/laf_ops.py
ducha-aiki: Fixed in 208fa64c. Reproduced first: with scales in `[4, 24)` and `PS=32`, every LAF in the sweep selected `pyr_idx = 0`, exactly as described.

`make_lafs` now assigns LAFs round-robin to the levels `extract_patches_from_pyramid` will actually build (`pyramid_levels()` mirrors its halving loop) and samples each scale uniformly inside that level's octave. Level 0 also absorbs the small-feature regime from 4 px, which selects level 0 anyway. Measured histogram at N=20000: `{0: 5000, 1: 5000, 2: 5000, 3: 5000}` for `--size 256`, and it follows `--size` — `{0..4: 4000 each}` at 512.

One correction worth recording: my first attempt put the bands an octave too low, because I assumed `get_laf_scale` of a frame from `laf_from_center_scale_ori(scale=s)` returns `s/2`. It returns `s`, so the level is `floor(log2(2s/PS))` and level `l` covers `s` in `[PS/2 * 2**l, PS/2 * 2**(l+1))`. The lopsided histogram is what caught it; the docstring now states the mapping.

Measured effect: ~3% on `extract_patches_from_pyramid`. #4128's packed atlas pays for the whole atlas whatever the LAFs select, so depth is currently free — which is the argument for the change rather than against it, since the old workload could not have distinguished that from a genuinely level-skipping implementation. The README says so explicitly now.

benchmarks/feature/laf_ops.py
ducha-aiki: Fixed in 208fa64c, in the shared helper rather than this script — `laf_ops.py` now goes through `common.run_batch_sweep`, so the drop happened there and affected every suite.

An unmeasured backend now appends a row instead of `continue`-ing: `median_us`, `iqr_us` and `throughput_per_s` are `null` (already permitted by `results_schema.py`), plus a structured `error` field holding the exception type for a compile-warmup failure (`"InductorError"`) or `"unavailable"` for a backend that was never available. The console NOTE stays.

Two consequences I had to handle:

- `error` is only meaningful for the compiled column, so the exception name is attached only when the backend name contains `compiled`; a missing OpenCV op reads `"unavailable"` rather than borrowing the compile exception.
- `error` had to be added to the docs generator's `MEASUREMENT_FIELDS`, otherwise it counts as a config field that varies within a file and rows render as `op @ 1 (n_lafs=2000, error=None)`.

Covered by two tests in `benchmarks/common_test.py`.

benchmarks/feature/laf_ops.py
ducha-aiki: Fixed — independently hit exactly this when contributing the result files, before reading the review, and it is in the branch already.

`docs/generate_benchmarks.py` now identifies a row by every config field that *varies within one result file*, not by `(op, batch)`:

- The flagship suites sweep batch alone, so `height`/`width`/`dtype` are constant and their labels stay `ColorJiggle @ 32`, byte-for-byte unchanged.
- This suite varies `n_lafs`, so rows render as `make_upright @ 1 (n_lafs=20000)` — B and N both visible, and all 21 configs per device survive into the page and the llms digest instead of the last one winning.

The digest names the full config too (`ellipse_to_laf@1,n_lafs=20000`). Also added an optional `units` metadata key so the column reads `throughput in LAFs/s` rather than the hardcoded `items/s`, since the item here is a LAF. Three tests in `tests/test_generate_benchmarks.py` cover the collision, the unchanged short label, and the digest.

ducha-aiki:
## Review round 2 + the requested Apple M1 numbers

Reviewed `208fa64c` (the post-Codex head), measured the M1, and pushed the fixes as `20519a32` and `24484a31`. Full detail is now in the PR description; summary here.

### Blocking, found and fixed

**[P1] Both committed result files leaked an absolute path into a public repo.** `metadata.kornia_module` read

```
/tmp/claude-1000/-home-ducha-dev-kornia/64bcb7f8-.../scratchpad/laf-wt/kornia/__init__.py
```

in `feature-laf-ops--i7-14700k-rtx-4090--cpu.json` and `--cuda.json` — a home-directory layout and a session UUID. `benchmarks/README.md` promises the opposite in the contribution instructions: *"Only aggregate load numbers (load average, memory) are recorded in the file - never process or app names."* The Codex P1 ask (record which checkout was measured) is right; satisfying it with an absolute path is not.

Fixed by exporting the checkout-relative form (`kornia/__init__.py`, or `outside-checkout`), which still carries the only bit a reader of the file needs — did kornia resolve to the tree the commit names? The absolute path stays on the console, which is where it is actually the useful diagnostic. Both files are scrubbed to the value the fixed script would have written for those runs, and `results_schema.validate_result` now rejects an absolute path in **any** string metadata value, so the next suite to add a metadata key cannot reintroduce this.

### Worth fixing, fixed

**[P2] The recorded commit did not describe the measured tree.** Both files record `9517c982`, but the harness edits that produced them — level stratification and the `sys.path` fix — were still uncommitted when they ran and landed in `208fa64c`. Check out `9517c982`, rerun, and you get the pre-stratification generator with different pyramid numbers. `git_commit()` now appends `-dirty` when tracked files differ from HEAD, and `contribute_result` warns; untracked files are excluded, since `--contribute` writes one into the checkout itself. Affects every suite. The M1 files were measured from a clean tree and record a bare `20519a32`.

**[P3] A vacuous assertion in the new docs test.** `test_render_page_keeps_configs_that_share_a_batch` asserted `"20" in rst`, which passes on the `2026-09-03` timestamp above the table even when the two B=1 configs collapse onto one row. Only the `n_lafs=` assertions discriminated. Now pins the rendered cells exactly.

### Checked and correct

`pyramid_levels()` matches the real halving loop in `extract_patches_from_pyramid` (`max_level` cap and `min(h, w) <= 2` guard included); the stratification math is right — level *l* covers scale `[PS/2·2^l, PS/2·2^(l+1))` and sub-level-0 scales clamp to 0, giving exactly 5000 LAFs/level at N=20000; `make_ellipses` stays positive-definite; the new `error` rows validate against the schema and render as `—`.

### The M1 measurements change two of the reported findings

Full tables are in the PR description and committed as `feature-laf-ops--apple-m1--{cpu,mps}.json`.

1. **"`torch.compile` is a 1.9–2.1× loss for `extract_patches_simple` on CPU" does not hold as a CPU claim.** On the M1 (torch 2.9.1, 4 threads) the same op is a **7.2× compile win** — 40806 → 294408 LAFs/s — and 6.7× for the pyramid variant. The regression belongs to that x86 inductor path and stack, not to the op or to compiling it for CPU. Bisect stack against machine before reading the kernel.

2. **MPS is measured rather than withdrawn, and the result is a user-facing warning.** On Apple silicon MPS *loses* to the same machine's CPU on every cheap LAF op — `ellipse_to_laf` by 5.3× eager (75.8M vs 14.2M LAFs/s) and 4.4× compiled, `make_upright` and `laf_from_center_scale_ori` by ~2× — and only wins where there is real work per LAF: patch extraction, 2.2–2.4× eager and up to 3.4× compiled. A pipeline that moves LAFs to the GPU for the frame math alone pays for the transfer twice.

The M1 CPU also explains the `laf_to_boundary_points` weak spot the PR already flagged: `torch.compile` recovers only 1.29× there because the cost is not a kernel. The 50-point basis is built with no `device=`, `.expand()`ed to `(B*N, n_pts, 3)` and only then `.to(device)`d, so each call materializes — and on a GPU transfers — a tensor that scales with the LAF count: ~12 MiB per call at N=20000 for a basis with 50 distinct rows. Build it on the target device and broadcast in the matmul.

Noise floor, from two independent full runs per device: CPU cells move by a median of 1.0% (worst 19%), MPS cells by 7.4% (worst 54%). Read no MPS difference under ~2× on this class of machine as signal. Every op compiled on both M1 devices, so there is no failure cell here either.

### Verdict

Approve once CI is green. The harness, the schema work and the `(op, batch)` row-collapse fix in `docs/generate_benchmarks.py` are all sound, and the suite is now measured on three devices across two architectures. One housekeeping item: the branch is 2 commits behind `main` (#4197, #4198) and wants a rebase; neither touches the LAF path, so the numbers stand.

ducha-aiki:
## CI triage: main is red, and two of the three causes are fixed here

Rebased onto `main` (`55046cf8`) and fixed what this PR's checks cover. All of this is pre-existing on `main` — none of it comes from the benchmark work, which touches no `kornia/` code.

### One lost hunk caused both red checks

`#4197` and `#4198` both edited `tests/models/test_sam_onnx.py`. `#4198` merged first and replaced the decorator block with a macOS `skipif`; `#4197` was then squash-merged and its conflict resolution kept `#4198`'s decorator **and** `#4197`'s imports — dropping `#4197`'s `xfail` marker in between. The result on `main` is:

```python
import torch                                              # unused
from kornia.core._compat import torch_version_ge, torch_version_lt   # unused

@pytest.mark.timeout(120)
@pytest.mark.skipif(sys.platform == "darwin", ...)
def test_sam_encoder_to_onnx(tmp_path):                   # xfail gone
```

That single hunk is both failures:

- **`pre-commit.ci` — `ruff check` fails**, "Found 3 errors (3 fixed, 0 remaining)", exit non-zero, and pre-commit.ci could not push the fix because it conflicted. The 3 errors are those orphaned imports (F401). Reproduced locally on a clean `origin/main` worktree.
- **Every torch 2.14.0 leg fails** — 6 ubuntu + 6 windows — on `test_sam_encoder_to_onnx` with `ConversionError` on `aten.mul.Scalar`, which is *verbatim* the node the dropped `xfail` names in its reason.

Restored in `5c747a1d` from `67dc35f9`/`1ab07837`, `strict=True` included, so an upstream fix inside 2.14.x announces itself as an XPASS. `ConversionError` subclasses `OnnxExporterError`, so `raises=` matches the observed failure. `#4198`'s macOS `skipif` and the 120 s timeout are kept as they stand: `#4197` raised the timeout to 300 s only because the export takes ~82 s on a macOS runner, and `#4198` decided to skip that platform outright instead, which is the later call on macOS.

### The same exporter gap also breaks the post-merge coverage jobs

`coverage / slow (float32)` and `(float64)` are red on `main` for the same reason in a different file: `test_object_detector.py::TestObjectDetector::test_onnx` fails on both `resnet50d` and `hgnetv2_l` with `DispatchError: No ONNX function found for aten.mul.Scalar`. `#4197` xfailed only the SAM case because that is the one PR CI runs — this test is `@pytest.mark.slow`, so only the post-merge jobs reach it. Fixed in `d4c5f3e2` with the same marker.

**Caveat worth stating:** no PR job runs slow tests, so that cell is verified from `main`'s coverage log, not from a green run here. Both parametrizations fail there today, so `strict=True` cannot XPASS.

### Not fixed here: the scheduled 2.6.0 legs are a different bug

`tests-cpu-ubuntu (*, 2.6.0)` fails on `main` for an unrelated reason — a dependency mismatch, not the exporter:

```
ImportError: cannot import name 'AttrsDescriptor' from 'triton.compiler.compiler'
```

plus `TorchExportError` on the dynamo ONNX exports. The 2.6.0 leg installs torch 2.6.0 over a lockfile that `#4197` bumped to torch 2.14 and its onnx/triton stack, so the leg runs torch 2.6.0 against a triton built for 2.14. That is a CI environment/lockfile fix, it needs a real 2.6.0 install to verify, and guessing at it inside a benchmarks PR is the wrong place — happy to take it as a separate PR. It is a scheduled-workflow leg, so it does not gate this PR or any other.

### State

Rebased onto `main`, so this branch is no longer behind. Verified locally: `pre-commit run --all-files` exits 0 (it fails on a clean `origin/main` checkout), `test_sam_onnx.py` is 1 passed / 1 skipped on darwin/torch 2.9.1, and the benchmark test set is 45 passed. Head is `d4c5f3e2`.

PR #4121, opened 2026-08-30
fix(geometry): validate nms box shape

### LINKED ISSUE #4019: W1b: `nms` shape guard uses `and` where `or` was meant — `(N, 5)` crashes inside `unbind`, `(B, N, 4)` returns garbage or runs out of memory
`kornia.geometry.bbox.nms` guards its input with

```python
if len(boxes.shape) != 2 and boxes.shape[-1] != 4:
    raise ValueError(f"boxes expected as Nx4. Got: {boxes.shape}.")
```

(`bbox.py:498`). With `and`, the guard only fires when **both** conditions fail, so a 2-D `(N, 5)` tensor and a 3-D `(B, N, 4)` tensor both pass. The same `and`/`or` slip as #3960 (`normalize_homography`).

## Repro

```python
import signal, torch
from kornia.geometry.bbox import nms

print("guard fires for (3, 5):", len((3, 5)) != 2 and 5 != 4, " for (2, 3, 4):", len((2, 3, 4)) != 2 and 4 != 4)

try:
    nms(torch.rand(3, 5), torch.rand(3), 0.5)
except ValueError as e:
    print("(3, 5):", e)

torch.manual_seed(0)
def handler(*a): raise TimeoutError
signal.signal(signal.SIGALRM, handler)
for B, N in [(3, 3), (8, 8), (12, 12)]:
    boxes = torch.rand(B, N, 4) * 100; boxes[..., 2:] += boxes[..., :2] + 1
    signal.alarm(15)
    try:
        out = nms(boxes, torch.rand(B), 0.5); signal.alarm(0)
        print(f"(B={B}, N={N}) returned {out.tolist()}")
    except TimeoutError:
        print(f"(B={B}, N={N}) no return within 15 s")
        break
```

## Observed

```
guard fires for (3, 5): False  for (2, 3, 4): False
(3, 5): too many values to unpack (expected 4)          <- from boxes.unbind(-1), not from the guard
(B=3, N=3) returned [1, 2, 0]                            <- batch indices, meaningless
(B=8, N=8) returned [6, 3, 7, 0, 4, 1, 2, 5]
(B=12, N=12) no return within 15 s                       <- peak RSS 1.9 GB when interrupted
```

Why the 3-D path degrades this way: with `(B, N, 4)` input and `(B,)` scores (the only shape that passes the batch-size check), `x1[i]` is `(N,)` and `x1[order[1:]]` is `(k, N)`, so `ovr` is 2-D and `torch.where(ovr <= iou_threshold)[0]` returns the *row* index once per matching column. `order = order[inds + 1]` therefore grows by up to `N×` per iteration (traced: `len 2 → 3` on the first iteration of a `(2, 3, 4)` input) and only shrinks once every remaining row equals the current one — worst case ~`N^B` entries. Small inputs terminate with a permutation of batch indices; larger ones exhaust memory (an earlier probe was killed with exit 137).

## Expected

`if boxes.ndim != 2 or boxes.shape[-1] != 4: raise ValueError(...)` — the documented `(N, 4)` contract, enforced. Neither shape has a caller to protect: `(N, 5)` already raises (with a worse message) and `(B, N, 4)` returns garbage or hangs, so tightening the guard changes no working call. A focused fix, welcome as a PR, with a test that asserts `ValueError` for both shapes.

## Classification

**Broken output** (crash with a misleading message; garbage or unbounded memory growth). Idiom 3 — a guard whose predicate does not match the documented contract.

## Disposition

n/a — single symbol.

## Environment

kornia 0.9.0rc1 @ `0685b1f5` (main), torch 2.9.1, python 3.11, cpu. Found during the bbox/boxes conventions audit (batch 4, tracking issue #4020); the batch documents the behavior as it is and does not change it. Related convention decision on the same function: #4008.

apoorvdarshan:

`kornia.geometry.bbox.nms` combined its two box-shape checks with `and`, allowing both `(N, 5)` and `(B, N, 4)` inputs past the documented `(N, 4)` contract. The former failed later with an internal unpacking error, while the latter could return meaningless indices or enter a runaway allocation path.

This changes the guard to reject either an invalid rank or an invalid final dimension. A deterministic regression test covers both malformed shapes and verifies the public `ValueError`.

Fixes #4019

## What did you check?

The regression was also run against the old predicate: `(N, 5)` failed with the wrong internal error and `(B, N, 4)` failed to raise. Both cases pass with this fix.

```text
$ pixi run test-module tests/geometry/test_bbox.py --dtype=float32,float64
30 passed

$ pixi run pre-commit-all
All hooks passed

$ pixi run typecheck
Completed successfully; one unrelated existing torch.cuda.amp.custom_fwd deprecation diagnostic
```

## How did AI help?

Codex was used to inspect the relevant code and policy files, make the one-line predicate correction, and draft the regression test. The resulting change was independently reviewed; that review identified that randomized malformed boxes could recreate the runaway path, so the test inputs were changed to deterministic zeros before the final validation above.

### DIFF
--- a/kornia/geometry/bbox.py
+++ b/kornia/geometry/bbox.py
@@ -556,7 +556,7 @@ def nms(boxes: torch.Tensor, scores: torch.Tensor, iou_threshold: float) -> torc
         tensor([0, 3, 1])
 
     """
-    if len(boxes.shape) != 2 and boxes.shape[-1] != 4:
+    if boxes.ndim != 2 or boxes.shape[-1] != 4:
         raise ValueError(f"boxes expected as Nx4. Got: {boxes.shape}.")
 
     if len(scores.shape) != 1:
--- a/tests/geometry/test_bbox.py
+++ b/tests/geometry/test_bbox.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 #
 
+import pytest
 import torch
 
 import kornia
@@ -320,3 +321,17 @@ class TestNMS(BaseTester):
         expected = torch.tensor([0, 3, 1], device=device, dtype=torch.long)
         actual = nms(boxes, scores, iou_threshold=0.8)
         self.assert_close(actual, expected)
+
+    @pytest.mark.parametrize(
+        ("boxes_shape", "scores_shape"),
+        [
+            ((3, 5), (3,)),
+            ((2, 3, 4), (2,)),
+        ],
+    )
+    def test_invalid_boxes_shape(self, boxes_shape, scores_shape, device, dtype):
+        boxes = torch.zeros(boxes_shape, device=device, dtype=dtype)
+        scores = torch.zeros(scores_shape, device=device, dtype=dtype)
+
+        with pytest.raises(ValueError, match="boxes expected as Nx4"):
+            nms(boxes, scores, iou_threshold=0.8)


ducha-aiki:
**APPROVE.** Verified, not read.

Setup: worktree at the PR head `e86eb2e4` with `kornia/` swapped to the merge-base `2009933e` for the fail-on-base pass; `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS.

**The pin discriminates.** With the production hunk reverted and the new test kept:

```
FAILED TestNMS::test_invalid_boxes_shape[cpu-float32-boxes_shape0-scores_shape0] - AssertionError: Regex pattern did not match.
FAILED TestNMS::test_invalid_boxes_shape[cpu-float32-boxes_shape1-scores_shape1] - Failed: DID NOT RAISE <class 'ValueError'>
2 failed, 2 passed
```

Two different failure modes, which is the right shape: `(N, 5)` did raise on base but from inside `unbind`, and `(B, N, 4)` did not raise at all.

**Nothing legitimate is newly rejected.** Sweep at head:

```
(N,4) normal     -> ok (2,) dtype=torch.int64 dev=cpu
(0,4) empty      -> ok (0,) dtype=torch.int64 dev=cpu     # #4109's empty-device fix still holds
(1,4) single     -> ok (1,)
(N,5) bad        -> ValueError: boxes expected as Nx4. Got: torch.Size([3, 5]).
(B,N,4) bad      -> ValueError: boxes expected as Nx4. Got: torch.Size([2, 3, 4]).
(4,) rank1       -> ValueError: boxes expected as Nx4. Got: torch.Size([4]).
```

**Both in-tree callers stay inside the tightened contract**, checked rather than assumed:

- `kornia/feature/sold2/sold2_detector.py:592` — `boxes = torch.cat([junctions - dist/2, junctions + dist/2], 1)` over `(N, 2)` junctions, so `(N, 4)`.
- `kornia/contrib/face_detection.py:253` — `self.nms(boxes[:, :4], scores, ...)`, so `(N, 4)`.

`tests/geometry/test_bbox.py` 30 passed at float32+float64; `tests/contrib/test_face_detection.py` 1 passed / 16 skipped.

One scheduling note, not an ask: #4129 and #4118 also edit `tests/geometry/test_bbox.py`, at line ~164 versus this PR's append at ~321. Different hunks, so no conflict is expected, but whichever lands second may need a trivial rebase.

ducha-aiki (approved):

PR #4122, opened 2026-08-30
perf(feature): closed-form 2x2 inverse in ellipse_to_laf

ducha-aiki:
Phase (c) of the `kornia.feature` audit, item 1 of the gap ledger's perf list — dispatched with the maintainer's numerics green-light (~2e-7-class deviation approved).

## The change

The matrix `ellipse_to_laf` inverts is lower-triangular by construction (`[[a11, 0], [a21, a22]]`), so its inverse is the closed form `[[1/a11, 0], [-a21/(a11·a22), 1/a22]]`. The batched `torch.inverse` it replaces was the whole cost of the op — and on MPS it also emitted a deprecated-resize `UserWarning` on every call and failed to `torch.compile` (`AssertionError`).

## Measured (back-to-back A/B, quiet machine, Apple Silicon, torch 2.9.1, LAFs/s)

| config | main eager | branch eager | main compiled | branch compiled |
| --- | --: | --: | --: | --: |
| CPU B=1 N=2000 | 6.9M | 39.5M | 6.7M | 31.7M |
| CPU B=1 N=20000 | 7.2M | **87.2M** | 8.7M | **162.6M** |
| CPU B=8 N=2000 | 7.2M | 77.6M | 7.7M | 100.3M |
| MPS B=1 N=20000 | 10.3k | **15.5M** | fails | **39.2M** |

That is ~12× eager / ~19× compiled on CPU and **~1500× on MPS**, where the batched 2×2 inverse hits a pathological linalg path (found by the #4119 harness).

## Behavior surface

- **Agreement**: max 1.6e-7 relative in float32, machine epsilon in float64, measured over three shape configs of realistic Oxford-format ellipses vs a `main` worktree.
- **float16/bfloat16 now work** on CPU (`linalg.inv` raises "Low precision dtypes not supported" on main); `fullgraph=True` compiles on CPU and MPS.
- **One edge changes**: an exactly degenerate ellipse (`a == 0` or `c == 0`) used to raise `linalg.LinAlgError` from the inverse; it now returns non-finite values in that LAF. Documented in the docstring and CHANGELOG.
- **Tests**: `TestELL2LAF` all green including gradcheck and the TorchScript `test_jit`. The failure set of `test_laf.py` + `test_affine_shape_estimator.py` across all four dtypes is a **strict subset** of `main`'s: the float32 moment path added to `PatchAffineShapeEstimator` fixes the four `TestLAFAffineShapeEstimator::test_toy`/`test_toy_preserve` fp16/bf16 failures (on CPU and CUDA), and no new failures appear. The remaining half-precision failures are pre-existing on `main` and untouched by this diff: `TestTransformLAFs::test_transform_points` (the test's own `torch.inverse` on the homography, `tests/feature/test_laf.py:922`) and `TestLAFAffNetShapeEstimator::test_toy` (pretrained-weight precision). Verified on Linux/WSL2 + RTX 4090 (torch 2.9.1+cu128) including `--isolate-half-precision` CUDA runs A/B'd against the base revision.
- `pixi run pre-commit-all` clean. CUDA now verified on an RTX 4090: float32/float64 all green across the three touched test files, and the benchmark reproduces at ~1.64x eager at N=1e5.

Posted on behalf of @ducha-aiki by Claude (Fable 5, Sonnet 5, Opus 5).


ducha-aiki (commented):
**[P2] Avoid overflow in the inverse off-diagonal — `kornia/feature/laf.py:262`**

`inv11 * inv22` can overflow before it is multiplied by `a21`, so this introduces non-finite output for valid, nondegenerate ellipses beyond the documented `a == 0` / `c == 0` behavior change. On CPU float32, `ellipse_to_laf(torch.tensor([[[0., 0., 1e-40, 0., 1e-40]]]))` is finite on `main` with a zero lower-left entry, but this branch returns `NaN` there because the expression evaluates `0 * inf`. Reassociating the operations (for example, `(-a21 * inv11) * inv22`) avoids that intermediate overflow for this case. Please use a numerically safer ordering and add a regression test for a nonzero subnormal diagonal with `b == 0`.

The focused `TestELL2LAF` suite otherwise passes locally, and the ordinary float16/bfloat16 forward cases are finite. Local inductor validation was blocked by the review environment missing `setuptools`; the PR's dynamo CI job is green.

ducha-aiki (commented):
**P1 — avoid overflowing the reciprocal product before applying the off-diagonal term** (`kornia/feature/laf.py:262`)

`inv21 = -a21 * (inv11 * inv22)` can overflow `inv11 * inv22` even when every entry of the mathematically correct inverse is finite. On CPU this valid, nondegenerate fp16 ellipse:

```python
e = torch.tensor([[[0., 0., 1e-5, 0., 1e-5]]], dtype=torch.float16)
ellipse_to_laf(e)
```

returns `nan` for `A[1, 0]`, although the correct off-diagonal is zero and both diagonal entries are finite (about 316). With `a=c=1e-6, b=5e-7`, the new path returns `-inf` instead of a finite value. Float32 also regresses: for `a=c=1e-40, b=0`, the old batched inverse returns a fully finite matrix while this implementation returns `nan` off diagonal.

This contradicts the new CPU fp16 support claim and the documentation that associates non-finite output with degenerate axes. Please compute the off-diagonal without first forming the potentially overflowing reciprocal product, and add finite-result regression coverage for small fp16 and float32 ellipses (including zero and nonzero `b`). The current half-precision-labeled `TestELL2LAF` cases do not catch this because their tensors are created without `dtype=dtype`.

Validation: `pixi run test-module tests/feature/test_laf.py::TestELL2LAF` and the same command with `--dtype=float16,bfloat16` both pass; the standalone reproductions above fail on PR head `72f847a`.

ducha-aiki:
Fixed in 1cf05291.

Verified both repros against the PR head before fixing:
- `ellipse_to_laf(torch.tensor([[[0., 0., 1e-6, 0., 1e-6]]], dtype=torch.float16))` → `nan` off-diagonal
- `ellipse_to_laf(torch.tensor([[[0., 0., 1e-6, 5e-7, 1e-6]]], dtype=torch.float16))` → `-inf`
- `ellipse_to_laf(torch.tensor([[[0., 0., 1e-40, 0., 1e-40]]]))` (float32) → `nan`, while `main`'s `torch.inverse`-based version returns a fully finite matrix on the same input

Root cause: `inv21 = -a21 * (inv11 * inv22)` forms the `inv11 * inv22` sub-product before multiplying by `a21`, so it can overflow to `inf` even when `a21 == 0` and both reciprocals are individually finite — `0 * inf = nan`. Reassociated to `-a21 * inv11 * inv22` (left-to-right), which multiplies `a21` into one reciprocal at a time and avoids the spurious intermediate overflow. Confirmed finite on all three repros above, and that `torch.compile(fullgraph=True)` and `TestELL2LAF` (incl. gradcheck/JIT, float16/bfloat16/float32/float64) are unaffected.

Added regression tests in `tests/feature/test_laf.py` (`test_no_overflow_small_diag`, `test_no_overflow_subnormal_diag`) using explicit `dtype=` construction per the note that the existing half-precision-labeled cases didn't exercise this. The documented degenerate-ellipse behavior (`a == 0` or `c == 0` → non-finite) is unchanged.

ducha-aiki:
## Review of #4122 — closed-form 2×2 inverse in `ellipse_to_laf`

I checked out the PR head in a worktree (verifying `kornia.__file__` resolved to it, not the editable primary checkout), and ran A/B numerics against `main`, gradient parity, `fullgraph=True` compilation on CPU and CUDA, the CUDA test collection, and the half-precision paths on an NVIDIA GPU.

The performance motivation is sound and the eager float32 numerics match `main`. The problems are in the overflow story the closed form is supposed to buy, and in what the tests actually pin.

### Correctness

**1. `0 * inf = nan` on an exactly degenerate ellipse — `kornia/feature/laf.py:265`**

The multiplication order does not prevent the NaN the comment directly above it claims to prevent. With `a11 == 0`, `inv11 = inf` and `a21 == 0`, so `inv21` is NaN rather than the mathematically correct `0`.

```python
>>> ellipse_to_laf(torch.tensor([[[1., 2., 0.0, 0.0, 4.0]]]))
tensor([[[[inf, 0.0, 1.0],
          [nan, 0.5, 2.0]]]])
```

`get_laf_scale` on that LAF is `nan`, and `laf_to_boundary_points` is entirely non-finite — one degenerate row poisons the whole downstream computation. Note the guarding is inconsistent within the same function: `a21`'s denominator gets `.clamp(1e-9)` on line 256, but the two new reciprocals on lines 260–261 got no equivalent.

**2. The chosen order only guards one of the two overflow directions — `kornia/feature/laf.py:262`**

`(-a21 * inv11) * inv22` still overflows when `a11` is tiny and `a22`/`b` are large, even where the exact result is comfortably representable:

```python
# float16 — exact result is -2896.3, well inside float16 range
>>> ellipse_to_laf(torch.tensor([[[0., 0., 1e-7, 65504., 65504.]]], dtype=torch.float16))
# inv21 = -inf

# float32 — exact result is -9.997e20
>>> ellipse_to_laf(torch.tensor([[[0., 0., 1e-42, 3e38, 3e38]]]))
# inv21 = -inf
```

The intermediate `a21 * inv11 = b / (a11 * (a11 + a22))` is larger than the final result by a factor of `a22`, so any `a22 > 1` makes the intermediate the binding constraint. This only bites non-PSD input (`b² > a·c`), which the function never validates but also never documents as forbidden.

The codebase's own precedent for this class of problem is to widen the dtype rather than pick a multiplication order — `_torch_solve_cast`, `_torch_svd_cast`, and the MKD/SIFT half-precision CHANGELOG entries all compute in float32 and cast back.

**3. `.clamp(1e-9)` is a silent no-op in float16 — `kornia/feature/laf.py:256`**

```python
>>> torch.tensor([0.0], dtype=torch.float16).clamp(1e-9).item()
0.0   # 1e-9 is far below float16's smallest subnormal, 5.96e-8
```

bfloat16 and float32 both retain ~1e-9. Before this PR the line was unreachable in float16 because `linalg.inv` raised first; the PR removes that raise and enables float16 without dtype-proofing the guard, so the safe-division protection is present in float32/bfloat16 and absent in exactly the dtype the PR newly makes reachable. The repo has already fixed this exact class elsewhere — CHANGELOG: *"an all-zero float16 descriptor normalises to zero instead of NaN (the 1e-12 guard underflows in float16)"*.

### Behavior change

**4. The `LinAlgError` path is gone and nothing pins the replacement — `kornia/feature/laf.py:266`**

On `main`, `ellipse_to_laf(torch.tensor([[[0., 0., 0.0, 0.0, 1.0]]]))` raises `_LinAlgError: The diagonal element 1 is zero`. On the branch it returns `[[inf, 0, 0], [nan, 1, 0]]`. A caller reading Oxford-format `.ellipse` files with a degenerate row used to get a loud failure at the conversion; now it gets NaN LAFs that flow through `LAFAffineShapeEstimator.forward` (`kornia/feature/affine_shape.py:164` → `get_laf_scale` → `scale_laf`) and into matching, surfacing far from the cause. The new `Note:` documents this, but nothing in `tests/feature/test_laf.py` pins it, so a future change back to raising would not be caught.

Relatedly, `CHANGELOG.md:81` files this under **Bug fixes** even though its own final sentence records an exception becoming silent non-finite output. Comparable exception-surface changes in that file are handled differently — the `normalize_homography` entry sits under **Breaking changes**, and the `warp_affine` entry spells out "observable to callers on the error path" for a mere exception-*type* change. A user with `try: ellipse_to_laf(...) except LinAlgError: drop_row()` loses that path silently and will not find it while reading Bug fixes.

### Tests

**5. The float16 case runs un-isolated on CUDA — `tests/feature/test_laf.py:268`**

`test_no_overflow_small_diag` hardcodes `torch.float16` through its own `ell_dtype` param instead of the injected `dtype` fixture. `conftest.py`'s `skip_half_precision_on_cuda` returns early when `"dtype" not in request.fixturenames`, and `_is_subprocess_isolated_test` requires `callspec.params["dtype_name"] in ("float16", "bfloat16")` — this test supplies neither, so it bypasses both the default skip and `--isolate-half-precision`:

```
$ pytest tests/feature/test_laf.py -k ELL2LAF --device=cuda
test_no_overflow_small_diag[cuda-0.0-ell_dtype0]   RUNS (float16, in-process)
test_jit[cuda-float16]                             SKIPPED (correctly)
```

A device-side assert from an un-isolated fp16 kernel would corrupt the CUDA context for every later test in the session — exactly what that machinery exists to prevent.

**6. Half the parametrization is vacuous — same test**

Reimplementing the buggy order `-a21 * (inv11 * inv22)` and running all four cases:

| case | under the bug |
|---|---|
| `float16`, `b=0.0` | non-finite — catches it |
| `float16`, `b=5e-7` | non-finite — catches it |
| `float32`, `b=0.0` | **finite — passes** |
| `float32`, `b=5e-7` | **finite — passes** |

With `a = c = 1e-6`, `inv11 * inv22 = 1e6`, nowhere near float32's 3.4e38 overflow. Two of the four cases can never fail under the regression they are named for.

**7. Only finiteness is asserted, never a value — `tests/feature/test_laf.py:277`**

Both new tests assert `torch.isfinite(laf).all()` and nothing else, and the assertion is further diluted by the two center coordinates hardcoded to `0.0`, which can never be non-finite. A regression that makes `inv21` finite but numerically wrong — dropping the negation, or swapping `inv11`/`inv22` into the wrong stack slots — passes both. `test_no_overflow_small_diag[float16, b=5e-7]` produces `inv21 = -233.5` against a float64 reference of ~`-250`; asserting that with `self.assert_close` (the convention AGENTS.md names) would pin the fix rather than just its finiteness.

**8. The subnormal literal is not backend-portable — `tests/feature/test_laf.py:279`**

`test_no_overflow_subnormal_diag` hardcodes a float32 subnormal (`1e-40`) and runs on every `--device`. I verified CPU and CUDA both preserve it (`sqrt(1e-40) = 1e-20`, reciprocal `1e20`, assertion holds), but on a backend that flushes denormals to zero — Metal/MPS, or any FTZ build — `a11` becomes `0`, `inv11` becomes `inf`, and the assertion fails. MPS is the one backend the PR's own perf story centres on and it remains unverified. A dtype/backend-independent construction, or an explicit skip, would make the intent portable.

**9. Two CHANGELOG claims are unpinned — `tests/feature/test_laf.py:270`**

The entry asserts the function "compiles with `fullgraph=True` on every backend" and "now also works in `float16`/`bfloat16` on CPU", but `TestELL2LAF` gains no `test_dynamo` and the new parametrization is `[torch.float16, torch.float32]` with no bfloat16. I confirmed both claims hold today (fullgraph compile succeeds on CPU and CUDA; bf16 returns finite output), which is precisely why they should be pinned — a later reintroduction of a graph break or a linalg call would go unnoticed. AGENTS.md points at the `torch_optimizer` fixture for the compile half.

### Reuse

**10. `_torch_inverse_cast` already solves the half-precision half — `kornia/feature/laf.py:260`**

`kornia/core/utils._torch_inverse_cast` exists to make `torch.inverse` work in fp16/bf16 (cast to fp32, invert, cast back), and `_inverse_3x3_closed_form` already establishes the closed-form-inverse pattern in that same shared module. The CHANGELOG credits the rewrite with enabling half precision, but the actual cause of `main`'s failure is that `laf.py:254` called raw `.inverse()` rather than `_torch_inverse_cast` — half precision would have worked all along.

The performance motivation for the closed form stands on its own. But if a 2×2 closed-form inverse is worth having, it belongs next to `_inverse_3x3_closed_form` in `kornia/core/utils.py` where the next caller can find it, rather than inlined in one feature function.

### Docs

**11. The new `Note:` is geometrically backwards — `kornia/feature/laf.py:232`**

"describes no area" is the opposite of what happens: a singular `[a b; b c]` has an *unbounded* region, and the resulting LAF's scale goes to inf/nan rather than to zero — `get_laf_scale(ellipse_to_laf([[[1., 2., 0., 0., 4.]]]))` is `nan` and the LAF's leading entry is `inf`. Separately, the stated trigger `a == 0` is exact-value-only: in float16 any `a` below ~5.96e-8 rounds to 0 on the way in, so a user passing `a = 1e-8` — a perfectly ordinary Oxford-format ellipse — hits the non-finite path with no `a == 0` anywhere in their data, and the Note gives them no way to predict it.

---

*Written by Claude (Opus 5) on behalf of @ducha-aiki*

ducha-aiki:
## Follow-up to the review above — pushed in f7b573a3

Ten of the twelve findings are addressed. Two are **not** being fixed, and in both cases the review above was wrong about them; correcting the record.

### Not fixed: finding 1, `0 * inf = nan` on an exactly degenerate ellipse

The review claimed the NaN was spurious because "the exact inverse has a zero off-diagonal". That is wrong — `[[0, 0], [0, 2]]` is *singular*, so there is no inverse to be correct about, and `inv11 = 1/0 = inf` regardless of how the off-diagonal is computed. I checked every multiplication order: all of them produce a non-finite LAF here, because the `0 * inf` is unavoidable once a reciprocal is `inf`.

What was actually wrong was the **code comment**, which promised a guarantee the code did not give. It has been rewritten to state the guarantee that does hold, and to say explicitly that the singular case is left unguarded.

### Not fixed: finding 7, `clamp(1e-9)` underflowing in float16

The underflow is real — `torch.tensor([0.0], dtype=torch.float16).clamp(1e-9)` is `0.0` — but the window it guards is unreachable in that dtype, so widening it would buy nothing and would perturb float32 numerics.

`a11 = sqrt(|a|)`, so the smallest nonzero `a11` in float16 is `sqrt(5.96e-8) = 2.44e-4`. `a11 + a22` is therefore either exactly `0` or at least `2.44e-4` — it is never inside the `(0, 1e-9)` interval the clamp exists to catch. And at exactly `0`, both reciprocals are `inf`, so the LAF is non-finite whatever `a21` holds. In float32/bfloat16, where that interval *is* reachable, `1e-9` is representable and the guard works as intended. The guard is correct in every dtype where it can fire.

This is now recorded in a comment at the line, so it does not get re-raised.

### The main fix: finding 4, overflow in the other direction

The off-diagonal now scales `b` by the smaller of the two reciprocals:

```python
inv21 = -a21 * torch.minimum(inv11, inv22) * torch.maximum(inv11, inv22)
```

The intermediate becomes the result times `min(a11, a22)` rather than times a fixed one of the two, so it cannot overflow while the result itself is representable. Verified against both fixed orders:

| case | `-a21 * inv11 * inv22` | `-a21 * (inv11 * inv22)` | min/max | float64 exact |
|---|---|---|---|---|
| fp16, `a` tiny / `c` huge | `-inf` | `-2898` | `-2898` | `-3162` |
| fp32, `a` tiny / `c` huge | `-inf` | `-9.997e20` | `-9.997e20` | `-1e21` |
| subnormal diagonal, `b = 0` | `-0` | `nan` | `-0` | `-0` |
| ordinary ellipse | `-0.0333` | `-0.0333` | `-0.0333` | `-0.0333` |

Worth noting for anyone reading this later: the overflow requires **non-PSD** input. For a valid ellipse, `b² ≤ a·c` gives `|b| ≤ a11·a22`, hence an intermediate `a21/a11 ≤ a22/(a11 + a22) ≤ 1` — brute-forced over the full exponent range of all four dtypes, the maximum is exactly `1.0000`. `ellipse_to_laf` does not validate PSD-ness, so the input is reachable, which is why it is covered.

Costs: accuracy unchanged (max `3.3e-7` relative error against float64 on 200k random positive-definite ellipses, identical for both orderings). CPU eager is ~17% slower at B=1, N=20000; the compiled path and both CUDA paths are unaffected. The `~19x compiled` figure in the CHANGELOG stands; the `~12x eager` figure needs re-measuring on the machine it came from.

### The rest

- **Findings 3, 6, 8, 9** collapse into one test rewrite. Both overflow tests now take the injected `dtype` fixture rather than hardcoding `torch.float16`, which restores the CUDA half-precision subprocess isolation they were bypassing, and build their inputs from `torch.finfo(dtype)` so every dtype exercises the regime instead of only float16. Both assert the value with `assert_close` rather than finiteness alone. The new test fails in all four dtypes against the previous ordering.

  One point the review missed: the *subnormal* direction cannot be made subnormal-free. `a11 · a22 ≥ finfo.tiny > 1 / finfo.max` in every IEEE dtype, so no normal input reaches that regime. That test keeps a subnormal input and gains an explicit flush-to-zero skip.

- **Findings 2 and 10**: the degenerate-ellipse behaviour is kept — detecting it up front would cost a device synchronisation and the `fullgraph=True` capture — and now has a test pinning that it returns non-finite values and does not raise. Its CHANGELOG note moved to `### Breaking changes`.

- **Finding 5**: `test_dynamo` and `test_dynamo_fullgraph` added, following the `tests/enhance/test_adjust.py` precedent.

- **Finding 11**: docstring `Note:` corrected — a singular form describes an unbounded strip, not "no area", and the scale goes to `nan` rather than to zero. The rounding condition is now explicit, since in float16 any `a` below `6e-8` arrives as `0`.

- **Finding 12**: the closed form stays inline — there is no second caller for a 2x2, unlike `_inverse_3x3_closed_form` which has five. The CHANGELOG now credits float16 support to dropping the raw `.inverse()` call rather than to linalg, since `_torch_inverse_cast` would have lifted that restriction too.

### Verified

`tests/feature/` 632 passed / 71 skipped; `TestELL2LAF` green on CPU across all four dtypes, on CUDA in float32/float64, and on CUDA in float16/bfloat16 under `--isolate-half-precision` (correctly skipped without it); `test_dynamo` and `test_dynamo_fullgraph` green on CPU and CUDA; pre-commit and doctests clean. Machine: NVIDIA GPU, torch 2.13.0+cu130, Python 3.13. `ty` was not available, so the typecheck is unrun — no annotations changed.

### Unrelated, pre-existing

`TestTransformLAFs::test_transform_points` fails 18 ways in float16/bfloat16 with `RuntimeError: linalg.inv: Low precision dtypes not supported` — `perspective_transform_lafs` still calls raw `linalg.inv`. Present both with and without this change, and on `main`. Same bug class this PR fixes for `ellipse_to_laf`; `_torch_inverse_cast` would fix it. Worth a separate issue.

---

*Written by Claude (Opus 5) on behalf of @ducha-aiki*

ducha-aiki:
**Numerical finding on the `inv21` formulation (and a planned change to the division form)**

The current min-reciprocal-first ordering

```python
inv21 = -a21 * torch.minimum(inv11, inv22) * torch.maximum(inv11, inv22)
```

fixes the intermediate *overflow* of the earlier orderings but introduces a silent *underflow*: the intermediate `a21 * min(inv11, inv22)` equals `result * min(a11, a22)`, which can flush to zero while the exact off-diagonal is a perfectly representable (even normal) float32. Repro: `ells = [0, 0, 1.1754944e-38, 1e-16, 1e38]` — exact `inv21` is `-9.22e-36`, `main`'s `torch.inverse` returns it exactly, this branch returns `-0.0`.

A 100k-sample adversarial sweep (scales 1e-45..1e35, float64 reference) over the candidate formulations:

| formulation | failure mode | bad cases /100k | worst error |
|---|---|---|---|
| `-a21 * (inv11 * inv22)` | `inv11*inv22` overflows → `nan`/`inf` | 269 non-finite | total |
| `-a21 * inv11 * inv22` | `a21*inv11` overflows → `inf` | 6 non-finite | total |
| min/max ordering (current) | `result*min(a11,a22)` flushes → wrong `0` | ~5k | total (100%) |
| `-a21 / (a11 * a22)` | subnormal denominator loses bits | 115 | 22% relative |

For reference, `main`'s `torch.inverse` raises `LinAlgError` on 5226 nonsingular inputs in this range (LU pivot underflow) and is worse than the division form on 1569 cases vs 176.

No two-operation expression is uniformly exact here — each has an input region where an intermediate leaves float32 range while the result is representable. They differ in how they fail: the division form is the only one whose worst case is a *degraded* value (up to 22% relative error, only when the exact `a11*a22` is subnormal, i.e. both `a` and `c` around 1e-38·1e-38 territory) rather than a *corrupted* one (`nan`, `inf`, or a false zero). It also deletes the `torch.minimum`/`torch.maximum` kernels from the hot path.

All of these regimes require `a`/`c` magnitudes no feature detector emits; this is about picking the least-wrong behavior at the extremes, not about realistic inputs.

Switching the implementation to `inv21 = -a21 / (a11 * a22)` accordingly.

ducha-aiki:
## Review of head `cce25ef4` — the division form, verified on MPS

Previous rounds covered CPU and CUDA; MPS — the backend the perf story centers on — was still unverified. I reviewed the final division-form implementation and ran the MPS verification on Apple Silicon (torch 2.9.1, macOS 26.5).

### MPS verification (all green)

- **Failure-set parity**: `tests/feature/test_laf.py` on `--device=mps --dtype=float32,float16,bfloat16` fails node-for-node identically to `main` (18 failures, all the pre-existing `perspective_transform_lafs` raw `linalg.inv` issue). `test_affine_shape_estimator.py` fails on the same 6 nodes as `main`; the 4 half-precision `test_toy`/`test_toy_preserve` cases changed *mode* from `linalg.inv` raising to a numeric mismatch, as the PR description says. One nuance for the record: on MPS that mismatch is a large **finite** error (`rel ~0.75`), not the NaN seen on CPU — the underflow issue presents differently there, but it is pre-existing in substance (the node fails on `main` too, by raising).
- **New tests**: `TestELL2LAF` 15 passed / 3 skipped on MPS across float32/float16/bfloat16. The subnormal test's flush-to-zero guard does exactly what it was added for: it skips on float32/bfloat16 (MPS flushes those subnormals inside `sqrt`) and **passes on float16**, whose subnormals MPS preserves.
- **Compile**: `test_dynamo` and `test_dynamo_fullgraph` pass on MPS under `KORNIA_TEST_OPTIMIZER=inductor`; a direct `torch.compile(fullgraph=True)` call also succeeds and matches eager.
- **Numerics**: max relative deviation vs `main`'s `torch.inverse` path on 20k well-conditioned ellipses is 2.39e-7 — inside the approved ~2e-7 class. Against a float64 reference the closed form is actually *closer* than the old path (2.97e-7 vs 3.34e-7). The deprecated-resize `UserWarning` is confirmed emitted by the old path on MPS and gone on the branch.
- **Benchmark** (`benchmarks/feature/ellipse_to_laf.py`, back-to-back A/B, quiet machine): `main` gives 11.6k LAFs/s eager at N=20000 and `torch.compile` warmup fails with `AssertionError`, exactly as claimed. The branch gives 23.4M eager / 48.0M compiled at N=20000 and 52.8M / 150.7M at N=100000. That is **~2000× eager at N=20000** — the CHANGELOG's "~1500× once measured ad hoc" is conservative and now has a benchmark-script number behind it (torch 2.9.1, Apple Silicon), if you want to cite it.

I also confirmed the division-form comment's core claim at the extremes in all four dtypes: `sqrt(max)²` stays finite (65472 in float16) and `sqrt(smallest_subnormal)²` rounds up to `tiny`, not zero — `a11 * a22` indeed neither overflows nor rounds to zero. `pixi run lint` passes; `ty check kornia` shows one diagnostic, pre-existing in `lightglue.py`, untouched here (closing the "typecheck unrun" gap from the previous round).

### Findings

**[P2] The regression that motivated the final commit is unpinned — `tests/feature/test_laf.py:270`**

I reimplemented all four candidate orderings and ran both new tests against each, in all four dtypes:

| ordering | asymmetric test | subnormal test |
|---|---|---|
| division (current) | pass | pass |
| `-a21 * (inv11 * inv22)` | pass | **FAIL** (caught) |
| `-a21 * inv11 * inv22` | **FAIL** (caught) | pass |
| min/max reordering (f7b573a3) | pass | pass |

Each multiply ordering is caught by exactly one test — good. But the **min/max ordering that `f8aeffb2` replaced passes both tests in every dtype**. The false-zero flush that justified switching to the division form (the repro from the formulation comment: `[0, 0, 1.1754944e-38, 1e-16, 1e38]` → division gives `-9.22e-36`, min/max gives `-0.0`) has no pin, so a future "optimization" back to a reciprocal-product form with a min/max guard would land green. A third case exercising the underflow direction — `a = tiny`, `c` large, `b` chosen so the exact off-diagonal is small but nonzero — would close the triangle. In float16 the representable-result window for this regime is itself subnormal, so it may need the same flush-to-zero skip as the subnormal test, or a float32-and-up construction.

**[P3] The benchmark docstring's A/B recipe does not do what it says — `benchmarks/feature/ellipse_to_laf.py:20`**

The docstring says to run "from a worktree root so the worktree shadows the editable install", but running the script as a file puts the *script's own directory* (`benchmarks/feature`) at `sys.path[0]`, not the cwd — I followed the recipe verbatim from a `main` worktree root and got branch numbers with the header printing the primary checkout's path. The self-diagnostic header is what caught it, so the mitigation works, but the documented recipe silently measures the wrong revision. `PYTHONPATH="$PWD"` from the worktree root does shadow correctly (the editable install is a plain `.pth` path entry, verified); the docstring should say that, or say "check the printed module path names the worktree" as the actual contract.

**[P3] Two wording nits.** `kornia/feature/laf.py:270`: "the single division is *exact* wherever the product is a normal number" — the division is correctly rounded, not exact (the repro above lands ~1.4e-9 relative off the float64 value with a fully normal product). "Never corrupted" is the guarantee that holds. And the CHANGELOG says the off-diagonal "divides `b` by the product of the diagonal square roots" — it divides `a21 = b / (a11 + a22)`, not `b`; a reader reconstructing the formula from the CHANGELOG gets the wrong expression.

### Verdict

The implementation, the docs, the degenerate-behavior story, and the perf claims all check out, now on MPS too. Approve once the P2 pin is added; the P3s are take-or-leave.

ducha-aiki:
## Follow-up — pushed in 29cf5319

The P2 pin and both P3 nits from the review of head `cce25ef4` are addressed.

**P2 — the min/max-ordering regression was unpinned.** Added `test_no_underflow_asymmetric_diag`: `a` is the dtype's smallest normal and `c` its reciprocal squared, so `a11 * a22 == 1` and the shared numerator `a21` stays representable while `f7b573a3`'s discarded ordering (`-a21 * min(inv11, inv22) * max(inv11, inv22)`) underflows its first product to a false `-0.0`. Verified against all four candidate orderings in all four dtypes: only the current division form and my new test's target ordering diverge as expected — I temporarily reinstated the min/max line locally and confirmed the new test fails on it (`-0.0` vs the expected nonzero value) in float16/bfloat16/float32/float64, then reverted. The comparison uses an explicit `rtol=0.1, atol=0.0` rather than `assert_close`'s per-dtype default, since the expected magnitudes here (as small as `~1e-247` for float64) are far below the default `atol` and would make the check vacuous otherwise; 10% is loose enough for float16's legitimate precision loss this deep into its subnormal range (measured ~2.8%) but two orders of magnitude tighter than the 100% error the bug produces.

**P3 — wording.** `kornia/feature/laf.py:270`: "exact" → "correctly rounded" for the single division, matching the earlier finding that flagged this exact conflation. `CHANGELOG.md`: "the off-diagonal divides `b`" → "divides `a21 = b / (a11 + a22)`", matching what the code actually computes.

**P3 — benchmark docstring.** Corrected the worktree-shadowing recipe: running this file directly puts its own directory on `sys.path[0]`, not the cwd, so `cd`-ing into a worktree alone does not shadow the editable install for a checked-in script (unlike the ad hoc stdin/`-c` snippets AGENTS.md describes). The docstring now says to run with `PYTHONPATH="$PWD"` from the worktree root and to check the printed module path.

Verified: `TestELL2LAF` 27/27 on CPU across all four dtypes (including the new test, `test_dynamo`, `test_dynamo_fullgraph` under `KORNIA_TEST_OPTIMIZER=inductor`), full `tests/feature/test_laf.py` unchanged at 186 passed / 18 pre-existing `perspective_transform_lafs` failures / 38 skipped, `pixi run lint` and `pixi run pre-commit-all` clean. CUDA/MPS unverified on this box for this specific commit — the change is test-only plus two comment/docstring edits, no runtime code path changed.

---

*Written by Claude (Sonnet 5) on behalf of @ducha-aiki*

ducha-aiki:
## Final review — head `29cf531`

Multi-agent verified review; every finding below was reproduced by execution against this head. Six findings survived (5 confirmed, 1 plausible); four other candidates were investigated and refuted, including the benchmark `meta["load"]` schema claim (that validation only ever runs on committed `--contribute` results).

### Must fix before merge

**1. `kornia/feature/affine_shape.py:164` — the PR's only in-repo caller of `ellipse_to_laf` is left unguarded, turning a loud error into silent NaNs (CONFIRMED).**
`bad_mask` in `PatchAffineShapeEstimator` is computed on *pre-normalization* moments; the `ellipse_shape / max` divide after it can flush `a` to exactly 0. In float16 an ordinary near-horizontal edge patch (`|gx|/|gy| ~ 1e-4`) yields `[0., 0., 1.]` with the mask not firing; `ellipse_to_laf` then returns `[[inf,0],[nan,1]]` and `LAFAffineShapeEstimator(32).half()` returns an **all-nan LAF with no error**. On base `2009933e` the same input raised `linalg.LinAlgError`. The CHANGELOG's `isfinite` migration advice names hypothetical external callers (Oxford `.ellipse` loaders) but not this one. Cheap fix: an `isfinite` guard/fallback in `LAFAffineShapeEstimator.forward` (and name this caller in the CHANGELOG entry).

### Wording fixes (two-liners)

**2. `kornia/feature/laf.py:234` — docstring Note misstates where `nan` appears (CONFIRMED).**
The Note says a degenerate LAF holds "nan where that inf meets a zero off-diagonal", but `nan` appears only in the `b == 0` sub-case. The generic degenerate ellipse (`a=0, b=0.5, c=4`) gives `[[inf, 0, 1], [-inf, 0.5, 2]]` — `torch.isnan(laf).any()` is **False** while the scale is nan. A caller following the Note literally and screening with `isnan` keeps every degenerate ellipse with `b != 0` (the common case). The Note should state `torch.isfinite(laf).all()` as the correct screen (currently only the CHANGELOG does).

**3. `kornia/feature/laf.py:260` — the comment above the retained `.clamp(1e-9)` claims the window is unreachable, which is float16-only (CONFIRMED).**
float32 `ells (a=c=1e-40, b=1e-25)` lands `a11+a22 = 2.0e-20` inside `(0, 1e-9)`; the clamp fires and `a21` becomes `1.0e-16` instead of the true `5.0e-06` (off by 5e10) — finite but wrong. The clamp itself is unchanged from base (not a regression); the defect is the dtype-generic premise ("either exactly zero or at least sqrt of the dtype's smallest positive") is false for f32/f64 vs `1e-9` (`sqrt(1.4e-45) ≈ 3.7e-23`) and will mislead a future maintainer. Scope the comment to float16 (or state the f32/f64 reachability).

### Benchmark / test hygiene

**4. `benchmarks/feature/ellipse_to_laf.py:71` — per-size `torch.compile` without `torch._dynamo.reset()` biases the first row (CONFIRMED).**
`unique_graphs` goes 1 → 2 → 2 across sizes `[10, 200, 4000]`: the first row is timed on a static-shape graph, later rows on the automatic-dynamic recompile. Measured static 16.95M vs dynamic 11.78M ellipses/s at n=1000 — a ~1.44× one-row bias in a revision-comparison script. A `torch._dynamo.reset()` in `build_ops` (as this PR's own `test_dynamo_fullgraph` does) fixes it. (The >8-sizes recompile-limit fallback sub-claim was refuted — automatic dynamic caps the graph count at 2.)

**5. `tests/feature/test_laf.py:348` — `test_dynamo_fullgraph` lacks a `dynamo_is_available()` skipif and bypasses the `torch_optimizer` fixture (PLAUSIBLE, low severity).**
With `KORNIA_TEST_OPTIMIZER` unset the conftest deselects `*dynamo*` tests, so the common path is safe; the exposure is `KORNIA_TEST_OPTIMIZER=jit` (or any value) on a dynamo-unavailable build (torch<2.4/py3.12, torch<2.6/py3.13), where it hard-errors instead of skipping. The identical merged pattern exists at `tests/enhance/test_adjust.py:1188`, so this is a shared gap; a one-line `@pytest.mark.skipif(not dynamo_is_available(), ...)` closes it.

### Optional follow-up (not blocking)

**6. `kornia/feature/laf.py:279` — a fused eager assembly is 20–42% faster with byte-identical outputs (CONFIRMED).**
`ells[..., 2:5:2].abs().sqrt()` + `torch.reciprocal` + `new_zeros(B,N,2,3)` slice-writes: 36.9µs → 28.9µs at N=1000 and 1314µs → 755µs at N=100000 (CPU eager), `torch.equal` across float16/32/64 including bit-identical nan/inf placement on degenerate inputs. Caveat: under `torch.compile` the fused source is 1–3% *slower* (inductor already fuses both), so this helps only the eager path — fine as a follow-up rather than churning this reviewed PR.

Refuted candidates (for the record): benchmark JSON `load` key, bfloat16 tolerance risk, unseeded-rand convention, shared 2x2-inverse-helper placement.

ducha-aiki:
Pushed `d2e53c8` addressing findings 1–5 from the review above:

1. `LAFAffineShapeEstimator.forward` now falls back to the input LAF for keypoints whose estimated shape is non-finite (`torch.where` on an `isfinite` mask — not the `bad_mask` mask-multiply idiom, since `nan * 0` is `nan`). Regression test `test_degenerate_ellipse_falls_back_to_input_laf_float16` pins the float16 near-horizontal-edge repro (with `supports_conv2d`/`supports_grid_sample` probes and an MPS-autocast skip); the CHANGELOG entry now names this internal caller. Verifying sharpened the mechanism: `bad_mask`'s `eps=1e-10` itself rounds to `0.0` in float16, so even exactly-zero pre-normalization moments slip past the `0 < eps` test, in addition to the post-check normalization flush.
2. Docstring Note prescribes `torch.isfinite(laf).all()`; `nan` documented as the `b == 0` sub-case only.
3. Clamp comment scoped to float16, with the f32 `a = c = 1e-40 → a11 + a22 = 2e-20` reachability example; clamp untouched.
4. `torch._dynamo.reset()` in `build_ops` before compiling (kills the ~1.4× first-row static-graph bias).
5. `@pytest.mark.skipif(not dynamo_is_available(), ...)` on `test_dynamo_fullgraph` (same pattern as `tests/augmentation/test_torch_export.py`); the matching gap in `tests/enhance/test_adjust.py` is left for a separate change.

Finding 6 (fused eager assembly, 20–42%) stays a follow-up as discussed. Verified on the new head: fp32/fp64 `ellipse_to_laf` outputs bit-identical to `29cf5319`; `tests/feature/test_laf.py` + `test_affine_shape_estimator.py` 163 passed / 14 skipped (f32+f64), new test passes at float16 and fails with the guard stashed; pre-commit and `ty` clean.

ducha-aiki:
## Final review — head `d2e53c8`

Multi-agent verified review of the full PR at this head, with every finding below re-executed in a worktree (`kornia.__file__` checked). CI is fully green at this head, including all torch 2.5.1 legs, dynamo, and typecheck. The closed form itself is solid — the four numeric pins were re-simulated across float16/bfloat16/float32/float64 and all hold. The remaining findings concentrate on the guard added in `d2e53c8`; three are defects introduced by that delta.

### Must fix before merge

**1. `kornia/feature/affine_shape.py:176` — the `torch.where` fallback sanitizes the forward but not the backward: a degenerate keypoint silently poisons gradients with NaN (CONFIRMED, three independent reproductions).**
`torch.where` routes zero grad into the `laf_out` branch, but `0 * d(1/a11)/da11 = 0 * inf = nan` inside `ellipse_to_laf`, so the NaN flows back anyway. On the PR's own regression-test input (float16 ramp, `requires_grad`): forward is finite and equals the input LAF, yet `out.sum().backward()` leaves 324/1024 NaN pixels in `img.grad` and `laf.grad` all-NaN — and via the shared `grid_sample`/Sobel weights the NaN spreads into well-conditioned keypoints' gradients in a mixed batch. On base the same call raised `LinAlgError`, so the failure was loud; now a training pipeline's upstream weights die on the next optimizer step with no signal. Fix is the double-`where` idiom: sanitize the degenerate rows of `ellipses` *before* `ellipse_to_laf`, so no non-finite value ever enters the graph. The new test never calls `.backward()` and gradcheck uses well-conditioned float64, so this path currently has zero backward coverage — add a backward assertion to the regression test.

**2. `tests/feature/test_affine_shape_estimator.py:130` — the new test hardcodes `torch.float16` while taking only the `device` fixture, escaping both CUDA half-precision guards in conftest (CONFIRMED).**
Both the half-precision skip and `--isolate-half-precision` subprocess routing key on the `dtype_name` param / `dtype` fixture (conftest.py:146, 467, 592); this test requests neither, so under `--device=cuda` it runs float16 `conv2d`/`grid_sample` kernels in the shared pytest process, where a device-side assert would corrupt the CUDA context for every later test — the same escape class as #4126. Use a `half_dtype`-style param the guards recognize (or extend the #4126 fix to cover it).

**3. `kornia/feature/affine_shape.py:176` — falling back to the raw input LAF breaks the `preserve_orientation=False` contract and disagrees with the `bad_mask` fallback the comment says it mirrors (CONFIRMED).**
`bad_mask` substitutes a circular shape yielding an upright frame; the new `torch.where` returns the caller's LAF verbatim — ellipticity and orientation kept even when `preserve_orientation=False` (class docstring: "The original LAF orientation is not preserved"). Verified: a 90°-rotated input LAF on the degenerate float16 patch comes back still rotated while every neighbouring keypoint is upright, so one output tensor mixes two conventions and a downstream `LAFOrienter` operates on an inconsistent frame. Fall back to `make_upright(laf)` (or the circular frame) instead. The regression test hides this because its input LAF is circular and axis-aligned.

**4. `tests/feature/test_scale_space_detector.py:365` and `:728` — two skip reasons, "LAFAffineShapeEstimator uses linalg.inv, which has no half-precision kernel", are falsified by this PR (CONFIRMED).**
The PR deletes the last `.inverse()` under `kornia/feature` and advertises half precision as newly working, yet both padding-contract tests stay skipped on the now-false premise — verified their bodies pass unskipped in float16 and bfloat16. Worse, the behavior they cover *did* change per dtype: a zero LAF now yields a 1e-5-scale frame in float32 (`bad_mask` path) but an all-zero LAF in float16 (new fallback path), contradicting the comment at `scale_space_detector.py:177` — and CI can't notice while the skips stand. Per the repo's stale-pin policy, un-skip these with the `supports_*` probes (they gate the missing 2.5.1 kernels correctly) and re-check that comment in the same change.

### Should fix (this PR or an immediately-filed follow-up)

**5. `kornia/feature/affine_shape.py:265` — `LAFAffNetShapeEstimator` runs the identical `make_upright`/`get_laf_scale`/`scale_laf` tail and can produce the same non-finite LAF, but only its sibling got the guard (mechanism plausible, not end-to-end reproduced).**
The two estimators are interchangeable `aff_module` arguments (`integrated.py` wires AffNet into KeyNetAffNetHardNet), so swapping them silently loses the new protection in exactly the half-precision regime this PR targets. Putting the guard in a small shared helper used by both return sites removes the asymmetry.

**6. `kornia/feature/affine_shape.py:102` — `bad_mask` is completely inert in float16, and that root cause is a one-line fix (CONFIRMED).**
`self.eps = 1e-10` promotes to float16 `0.0`, so `ellipse_shape < 0.0` is identically `False` on non-negative moments: `PatchAffineShapeEstimator(32).half()` on a zero patch returns `[[nan, nan, nan]]` (float32 correctly gives `[1., 0., 1.]`) — the NaN coming from the `0/0` normalization after the dead mask. Anyone using this public module directly in float16 still gets silent NaN shapes; a dtype-aware eps (clamp to `torch.finfo(dtype).tiny`) restores the guard and shrinks how often the downstream fallback fires at all. Pre-existing on base, but this PR's comments now document the brokenness in three places instead of fixing it. Related wording nit: the test comment and CHANGELOG say the degenerate patch "slips through as the ellipse `[0., 0., 1.]` / normalization flushes a component to zero" — on torch 2.9.1 CPU the moments are all exactly zero *before* normalization and the ellipse is `[nan, nan, nan]` via `0/0`; the eps-rounds-to-zero clause is the operative (and correct) part of the story.

### Optional follow-up (not blocking)

**7. `kornia/feature/laf.py:283` — the closed form materialises the 2×2 block twice (`stack`+`view` then `cat`); a single `torch.cat` assembly is byte-identical (`torch.equal` across dtypes) and measured 1.25–1.43× eager, up to 1.88× compiled at N=1e5 on top of this PR.** Same trade as the fused assembly deferred last round — fine as a follow-up rather than churning this reviewed PR.

### Refuted during verification (for the record)

- "The float16 test errors on the torch 2.5.1 CPU legs for the missing `replication_pad2d` kernel" — refuted: 2.5.1 CPU also lacks the float16 `grid_sample` kernel, so the test's existing `supports_grid_sample` probe already skips it there; all 2.5.1 CI legs are green at this head.
- Benchmark `git_commit` cwd-attribution, CHANGELOG dtype/version wording nits, and comment-verbosity candidates were triaged out as doc-tier.

### Verdict

The numerics of the closed form are done and pinned. The `d2e53c8` guard needs one more pass: findings 1–3 are defects the delta itself introduced (silent NaN backward, a CUDA-isolation escape in the new test, a broken `preserve_orientation` contract), and finding 4 is the repo's own stale-pin rule. **Changes requested** — all four are small, localized fixes.

Hexecu:
I've proposed a focused fix for #4123 in https://github.com/kornia/kornia/pull/4131, covering the patch estimator root cause discussed here.

I rechecked it on top of `d2e53c8` on CPU and MPS. Computing the moments in float32 restores the rectangle's shape, including the finite MPS mismatch that the downstream guard cannot catch. The new fallback regression still passes, and the ramp and mixed-batch CPU cases I tried now have finite image and LAF gradients. This does not cover arbitrary custom estimators.

The existing LAF toy tests still exceed their fixed `1e-4` comparisons, which I left unchanged and documented. No inverse changes are included. If both fixes land, the new fallback comments describing the patch estimator's float16 underflow will need updating too.

ducha-aiki:
@Hexecu Please, let me finish with this PR and if there is something left, we can discuss

ducha-aiki:
## Final review

Reviewed at c6604028; re-checked against the branch tip 63581ad3. The five follow-up commits (97202fee..63581ad3) already resolve the three functional findings below, so at tip only the three cleanup items remain open.

### Resolved by the follow-up commits

1. **NaN gradients in backward through the degenerate-ellipse fallback** (`kornia/feature/affine_shape.py`). The output-level `torch.where` fixed the forward pass only: with one degenerate keypoint (e.g. the float16 near-horizontal-edge patch from the new test), backward produced `0 * inf = NaN` local derivatives inside `ellipse_to_laf`, which flowed through `grid_sample` backward into the shared image gradient (~289 NaN pixels per bad keypoint) and the full bad row of `laf.grad` — silently, where `main` raised `LinAlgError` loudly. Fixed by sanitizing the degenerate ellipse rows *before* `ellipse_to_laf` (07c8274e).

2. **`bad_mask` eps=1e-10 inert in float16** (`kornia/feature/affine_shape.py`). In float16, `1e-10` rounds to `0.0`, so exactly-zero moments passed `< self.eps` and a direct float16 caller of the public `PatchAffineShapeEstimator` received the degenerate ellipse `[0, 0, 1]` with no fallback at all. Fixed by the dtype-aware guard (97202fee).

3. **Stale half-precision skips** (`tests/feature/test_scale_space_detector.py`). Two skips still cited "uses linalg.inv, which has no half-precision kernel" although this branch removed `torch.inverse` from `ellipse_to_laf`, leaving the exact coverage the branch enables permanently disabled. Fixed in 3d3f16c1.

### Still open at tip

4. **Agent planning artifacts committed under `docs/`**: `docs/superpowers/plans/2026-08-31-pr-4122-affine-fallback.md` and `docs/superpowers/specs/2026-08-31-pr-4122-affine-fallback-design.md` are session-planning documents addressed to coding agents, describing work that is now either done or not in this branch. `docs/superpowers/` does not exist on `main`; these should be dropped before merge — the CHANGELOG and code comments carry the durable rationale.

5. **The kept `.clamp(1e-9)` on `a11 + a22`** (`kornia/feature/laf.py:269`) is, by its own six-line justification comment, either a no-op (at `a11 + a22 == 0` the result is `inf` regardless) or a silent corruptor (in the sub-denormal float32/float64 window it perturbs `a21` by orders of magnitude — finite but wrong, passing the `torch.isfinite` screen the new Note recommends). Dropping it changes nothing for the degenerate case, makes the finite path correctly rounded, and removes the comment.

6. **Duplicated hand-transcribed test oracle** (`tests/feature/test_laf.py:282` and `:324`): the float64 reference block re-derives the implementation's own `a21` formula and is duplicated verbatim, so it adds no oracle independence and both copies must be found together on any change. Prefer calling `ellipse_to_laf(inp.cpu().double())[0, 0, 1, 0]` as the reference, or factor a small helper next to `_well_conditioned_ellipses`.

Refuted during verification (for the record): the removal of the loud `LinAlgError` for external callers is an intentional, CHANGELOG-documented contract change, not a regression.

ducha-aiki:
## Final review

Overall the branch looks in good shape: the earlier zero-LAF NaN-grad concern was **refuted** (byte-identical behavior vs base `2009933e` for both estimators), and typecheck passes. Four findings remain, ordered by severity — none is a hard blocker, but the first should be addressed (or the claim scoped) before merge.

### 1. fp16 hole in the "finite fallback with finite backward" guarantee — `kornia/feature/affine_shape.py:130` (confirmed)

In float16, `make_upright`'s `eps=1e-9` and `get_laf_scale`'s `eps=1e-10` underflow to 0, so for a zero-scale input LAF row with a singular candidate:

- `LAFAffNetShapeEstimator(pretrained=False, preserve_orientation=False)` returns a NaN **forward** (`[[nan,0,0],[nan,nan,0]]` — the fallback is `make_upright(zero laf)` = 0/0 after eps underflow);
- with `preserve_orientation=True` the forward is finite but `out.sum().backward()` leaves `laf.grad = [[nan,nan,1],[nan,nan,1]]` (`scale_orig/ellipse_scale` is 0/0 in fp16; `torch.where` backward propagates `0*nan=nan`).

This is **not a regression** (base behaves the same or worse), but the CHANGELOG line "both it and LAFAffNetShapeEstimator use a finite fallback with finite backward" over-claims, and the new finite-backward tests only cover float32 and scale-8 LAFs. Fix: dtype-aware eps (compute the scale guard in float32, or clamp by `torch.finfo(dtype).tiny`), or scope the CHANGELOG/docstring claim to float32+ and pin the fp16 hole in a test.

### 2. Load-bearing `torch.where` sandwich has no comment — `kornia/feature/affine_shape.py:131`

`_finalize_laf`'s first `torch.where(bad..., fallback, laf_out)` looks redundant with the identical final `where`, but it's what keeps `ellipse_scale = get_laf_scale(safe_laf_out)` finite/nonzero for finite-but-singular candidates (AffNet's tanh output `(-1,0,-1)`, fp16 det overflow), so `scale_orig/ellipse_scale` and its backward stay finite. Likewise `make_candidate_upright` must run *between* the two wheres, on the already-sanitized candidate, or its det division reintroduces inf/nan into backward. A future cleanup deleting the first `where` or hoisting `make_upright` would pass all well-conditioned tests while regressing the degenerate-input guarantees the new tests pin. One or two sentences on lines 130–141 would prevent that (this diff already comments a comparable subtlety in `ellipse_to_laf` with eight lines).

### 3. Private `_invalid_laf_mask` vs. docstring telling callers to hand-roll the screen — `kornia/feature/affine_shape.py:118` (plausible)

`ellipse_to_laf` now silently returns non-finite LAFs for degenerate ellipses, and its new Note teaches third-party callers to screen with `torch.isfinite` themselves (warning about the isnan-misses-inf trap) — while the repo keeps its own tested, fullgraph-safe screen as a private helper. Consider promoting `_invalid_laf_mask` to a small public helper in `kornia/feature/laf.py` (exported, and pointed to by the Note); `affine_shape.py` already imports six LAF utilities from there, and it costs nothing to the compile/sync constraints.

### 4. fp16-quantized `weighting` buffer dilutes the float32-moment promise — `kornia/feature/affine_shape.py:93` (plausible, minor)

`PatchAffineShapeEstimator(32).to(device, torch.float16)`: the patch is promoted to float32, but `weighting = self.weighting.to(patch.dtype).to(patch.device)` upcasts weights already rounded to fp16, so the documented float32 moment computation runs with quantized weights — near the `bad_mask` eps threshold this can flip the degenerate/non-degenerate decision relative to a float32 module. Minor fix: keep/regenerate the weighting in float32 when promoting, and fuse the conversion as `.to(device=patch.device, dtype=patch.dtype)` to drop the intermediate copy.

### Checked and dropped

Refuted or dropped for precision: the zero-LAF NaN-grad "regression" (byte-identical vs base), the `ori_orig` typecheck concern (passes), the `make_upright` recompute (non-default config, negligible), the `c<=0` redundancy, a `.view(1,1,3)` nit, the documented intentional API break, and the pre-existing `det==0` exact-equality limitation.

ducha-aiki:
## Review of #4122 — closed-form 2×2 inverse in `ellipse_to_laf`

**Verdict: approve, with two small CHANGELOG fixes and a stale PR-body note.** The math is right, the numerics claim holds up under independent measurement, the fallback machinery in `affine_shape.py` is carefully constructed (the two-`where` backward-safety pattern is applied correctly), and the test failure surface is strictly *better* than `main`, not merely equal to it. Everything below was verified on this box (Linux/WSL2, torch 2.9.1+cu128, i7-14700K, RTX 4090).

### What I verified

**Correctness of the closed form.** `[[a11, 0], [a21, a22]]⁻¹ = [[1/a11, 0], [-a21/(a11·a22), 1/a22]]` — checked symbolically and empirically. Independent A/B against a `2009933e` worktree (with `kornia.__file__` printed in both blocks per AGENTS.md): over 4×5000 random well-conditioned Oxford ellipses, max relative deviation **2.35e-7 in float32** and **4.4e-16 in float64**. Both implementations sit equally far from the float64 ground truth (3.0e-7 both), so the deviation is rounding noise, not a bias.

**Overflow-ordering analysis.** The `-a21 / (a11 * a22)` division form is the right one and the three regression tests (`test_no_overflow_asymmetric_diag`, `test_no_overflow_subnormal_diag`, `test_no_underflow_asymmetric_diag`) pin each failure mode of the rejected orderings, including the flush-to-zero skip guards. The silent removal of the old `.clamp(1e-9)` on the root sum is an improvement (pinned by `test_small_root_sum_is_not_clamped`).

**Test surface vs `main` (CPU, all four dtypes, `test_laf.py` + `test_affine_shape_estimator.py`).** Base: 24 failed. Branch: 20 failed. The branch **fixes four base failures** (`TestLAFAffineShapeEstimator::test_toy`/`test_toy_preserve` in fp16/bf16, via the float32 moment path in `PatchAffineShapeEstimator`) and introduces **zero** new ones. The 20 remaining failures fail identically on base: 18× `TestTransformLAFs::test_transform_points` fp16/bf16 (the *test's own* `torch.inverse` at `tests/feature/test_laf.py:922`, see below) and 2× `TestLAFAffNetShapeEstimator::test_toy` fp16/bf16 (pretrained-weight precision, pre-existing).

**CUDA.** `test_laf.py` + `test_affine_shape_estimator.py` + `test_scale_space_detector.py` at float32/float64: **296 passed, 0 failed**. Half-precision isolated runs (`--isolate-half-precision`) A/B'd against base: the branch failure set is a **strict subset** of base's (base 24 failed / branch 20 failed over the three files), the four fixed being the same `TestLAFAffineShapeEstimator::test_toy`/`test_toy_preserve` fp16/bf16 as on CPU, with zero new failures; the 16 collection-time ERRORs under `--isolate-half-precision` are name-for-name identical on both sides (pre-existing harness artifact, not this PR).

**Benchmark reproduction.** `benchmarks/feature/ellipse_to_laf.py`, eager float32: CPU N=20000 20.3M → 93.8M LAFs/s (**4.6×**), CUDA N=100000 372M → 612M LAFs/s (**1.64×**) — both inside the CHANGELOG's claimed ranges (2.4–5.3× CPU, 1.4–1.7× CUDA). I also hit the exact `sys.path` trap the script's docstring warns about (my first branch measurement silently measured base from the worktree cwd) — the warning and the `# kornia module:` header line earn their keep.

**Hygiene.** `pixi run lint` clean; `ty` typecheck has only the pre-existing `lightglue.py` deprecation warning; doctests for `ellipse_to_laf`/`laf_is_valid` pass; `laf_is_valid` is exported in `__init__.py` and added to `docs/source/feature.rst`; PR head matches the reviewed tree (`24b92511`).

### Findings

1. **CHANGELOG: "torch 2.13" is wrong.** The environment that produced (and now reproduced) those numbers is torch **2.9.1**. Fix before merge — it's a citable-methodology entry.
2. **CHANGELOG grammar:** "The **one** in-repo affine estimators handle invalid candidates…" — stray "one" (and the "The fallback is the input LAF / when preserve_orientation…" line-wrap reads oddly mid-sentence).
3. **PR body is stale.** It says the half-precision failure sets are "node-for-node identical to main" and that the `TestLAFAffineShapeEstimator` half tests "now fail on a pre-existing fp16 underflow" — but the later commits (`3d3f16c1`, the float32 moment path) *fixed* those four tests. The description undersells the PR; worth updating so the merge record is accurate.
4. *(minor, no action needed)* The fp16/bf16 path in `PatchAffineShapeEstimator.forward` rebuilds the Gaussian kernel on every call. Correct (the fp16-quantized buffer must not be reused), and half patches are small, but a cached float32 buffer would avoid it if this ever shows up in a profile.
5. *(minor, no action needed)* In `_finalize_laf`'s non-preserve invalid-input fallback, the raw `laf[..., 2:3]` center is kept, so a non-finite input *center* still propagates. Consistent with the documented "finite zero-scale input keeps its center" contract — flagging only so it's a known choice.

### On the remaining `TestTransformLAFs` half-precision failures (separate follow-up)

They are not library failures: `perspective_transform_lafs` itself never inverts anything. The test inverts the random homography to map back — `src_homo_dst = torch.inverse(dst_homo_src)` at `tests/feature/test_laf.py:922` — and CPU `linalg.inv` has no fp16/bf16 kernel. Two in-tree fixes already exist: `_inverse_3x3_closed_form` (`kornia/core/utils.py:165`, pure arithmetic, half-friendly, already used by `warp_perspective`) or `_torch_inverse_cast` (`kornia/core/utils.py:230`, casts through float32). A one-line test change clears all 18. Out of scope here.

---

PR #4125, opened 2026-08-31
perf(feature): sample pyramid patches from an atlas

ducha-aiki:
## Summary

- sample each LAF once from a guarded, packed pyramid atlas instead of sampling every LAF at every level
- use one-pixel replicated guards to preserve exact border values and outward gradients
- compute fp16/bf16 grids in float32 and keep the existing output dtype
- clamp oversized LAFs to the actual coarsest built level instead of returning an all-zero patch
- use a compile-static levelwise fallback when atlas storage plus any sampling-dtype copy would exceed 128 MiB

This PR is intentionally stacked on #4120 (`perf/extract-patches-batched`).

## Performance

Measured before implementation and again on this branch with the #4119 benchmark harness, Torch 2.9.1, Apple Silicon. Values are LAFs/s for `extract_patches_from_pyramid`, PS=32.

| device / image | config | eager base → atlas | compiled base → atlas |
|---|---:|---:|---:|
| CPU / 256² | B1 N2000 | 9,285 → 31,307 (3.37×) | 67,238 → 157,207 (2.34×) |
| CPU / 512² | B1 N2000 | 7,317 → 28,719 (3.92×) | 48,655 → 122,487 (2.52×) |
| CPU / 512² | B8 N2000 | 7,302 → 29,629 (4.06×) | 7,011 → 82,626 (11.79×) |
| MPS / 256² | B1 N2000 | 14,994 → 42,714 (2.85×) | 154,142 → 406,483 (2.64×) |
| MPS / 512² | B1 N2000 | 12,028 → 41,516 (3.45×) | 131,677 → 331,135 (2.51×) |
| MPS / 512² | B8 N2000 | 12,283 → 56,144 (4.57×) | 30,026 → 406,568 (13.54×) |

The tiny-N loss/flat case is retained explicitly: MPS 512², B1 N100 compiled was 62,012 → 61,332 LAFs/s (0.99×, within run noise). Eager improved 1.19–2.72× at N100 across the tested matrix.

### Typical inference workloads

These runs explicitly used `torch.inference_mode()`, level-spanning LAF scales, and fresh processes.

| device / image | PS / N | base → atlas latency | speedup | peak RSS base → atlas |
|---|---:|---:|---:|---:|
| CPU / 1024×768 | 32 / 2000 | 279.8 → 81.8 ms | 3.42× | 229 → 211 MiB |
| CPU / 1024×768 | 41 / 2000 | 442.8 → 119.5 ms | 3.70× | 242 → 229 MiB |
| CPU / 2048×1536 | 32 / 2000 | 407.0 → 144.9 ms | 2.81× | 435 → 366 MiB |
| CPU / 2048×1536 | 41 / 2000 | 600.9 → 182.2 ms | 3.30× | 437 → 366 MiB |
| MPS / 1024×768 | 32 / 2000 | 169.1 → 49.0 ms | 3.45× | 45 → 26 MiB RSS delta |
| MPS / 1024×768 | 41 / 2000 | 273.7 → 69.7 ms | 3.92× | 44 → 26 MiB RSS delta |
| MPS / 2048×1536 | 32 / 2000 | 206.7 → 49.2 ms | 4.20× | 51 → 27 MiB RSS delta |
| MPS / 2048×1536 | 41 / 2000 | 328.0 → 77.1 ms | 4.26× | 50 → 27 MiB RSS delta |

At N100, CPU speedups were 1.08–1.57× and MPS speedups were 1.28–1.70× for these same PS/image combinations.

I did not run 8000×6000 on MPS: it can pressure the unified allocator enough to hang this machine. CPU single-shot tests exercise the guarded fallback instead. At N2000 it peaked at about 1.58 GiB for PS32 and 4.10 GiB for PS41; the latter is output-dominated (about 4.0 GiB for the returned patches). The fallback avoids constructing the roughly gigabyte-scale atlas.

## Numerics and behavior

A 70-case branch-vs-base probe covered five even/odd/single-level shapes, normalized and pixel LAFs, CPU float32/float64/float16/bfloat16, and MPS float32/float16/bfloat16.

- float32: allclose; max absolute difference 2.94e-5, max p99 1.13e-5, max mean 1.02e-6
- float64: allclose; max absolute difference 5.82e-14
- fp16/bf16 intentionally differ from the old reduced-precision grid arithmetic; against a float32-grid levelwise reference they are allclose, with max differences 4.88e-4 and 9.77e-4 respectively
- atlas and forced fallback have exact forward and border-gradient parity in the seam regressions
- GRAF DoG/SIFT and KeyNet matching quality metrics were exactly unchanged for all evaluated pairs

## Validation

- `pixi run test-module tests/feature/test_laf.py --device=cpu --dtype=float32,float64 --color=no` — 143 passed, 12 skipped
- CPU fp16/bf16 pyramid tests passed; the full half run has the same 18 unrelated `torch.inverse` failures as the base branch
- MPS float32/fp16/bf16 pyramid tests — 27 passed, 1 skipped
- atlas and forced-fallback Dynamo/Inductor tests passed, including direct `torch.compile(fullgraph=True)` probes
- `pixi run test-module tests/feature/test_integrated.py --device=cpu --dtype=float32 --color=no` — 25 passed, 10 skipped
- `pixi run test-module tests/feature --device=cpu --dtype=float32 --color=no -q` — 634 passed, 72 skipped
- `pixi run pre-commit-all`

CUDA was unavailable on this machine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking changes
 
+* `extract_patches_from_pyramid` now samples ordinary-sized inputs once from a packed pyramid atlas instead of
+  sampling every LAF at every pyramid level. A statically selected levelwise fallback keeps atlases larger than
+  128 MiB from becoming a memory hazard. LAFs whose nominal level is coarser than the last usable pyramid level now
+  sample that actual coarsest level; they previously returned an all-zero patch. Atlas coordinate arithmetic also
+  runs in `float32` for `float16` and `bfloat16` inputs, avoiding the normalized-coordinate precision loss caused by
+  the wider atlas. Floating-point results are therefore not byte-identical to the former levelwise implementation.
+
 * The shape guards of `normalize_homography` and `denormalize_homography` now reject everything but a
   `(3, 3)`/`(B, 3, 3)` matrix, and `normalize_homography3d`'s everything but a `(4, 4)`/`(B, 4, 4)` one (#3999).
   A rank-4 input used to pass the guard and come back with its rank unchanged —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,13 +13,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `extract_patches_from_pyramid` now samples ordinary-sized inputs once from a packed pyramid atlas instead of
   sampling every LAF at every pyramid level. A statically selected levelwise fallback keeps atlases larger than
   128 MiB from becoming a memory hazard. LAFs whose nominal level is coarser than the last usable pyramid level now
-  sample that actual coarsest level; they previously returned an all-zero patch. Atlas coordinate arithmetic also
-  runs in `float32` for `float16` and `bfloat16` inputs, avoiding the normalized-coordinate precision loss caused by
-  the wider atlas, and `extract_patches_simple` computes its sampling grid in `float32` for reduced-precision inputs
-  on every device for the same reason -- its output matches the pyramid extractor's level-0 output again, and
-  half-precision patches therefore change on CUDA and MPS as well as the CPU (superseding the device scope of the
-  reduced-precision entry under **Bug fixes**). Floating-point results are not byte-identical to the former
-  levelwise implementation. Both extractors chunk their sampling along N, in the spirit of `batched_forward`, so
+  sample that actual coarsest level; they previously returned an all-zero patch. For `float16` and `bfloat16` inputs
+  the whole pyramid -- padding, `pyrdown`, coordinate arithmetic and sampling -- runs in `float32` and the patches
+  are cast back, on every device: this avoids the normalized-coordinate precision loss caused by the wider atlas,
+  keeps the extraction off `replication_pad2d`/`grid_sampler_2d` kernels that old torch builds lack for half dtypes
+  on the CPU (torch <= 2.5.1 would otherwise crash), and avoids recasting the full atlas on every sampling chunk.
+  `extract_patches_simple` computes its sampling grid in `float32` for reduced-precision inputs on every device for
+  the same reason -- its output matches the pyramid extractor's level-0 output again. Half-precision patches
+  therefore change on CUDA and MPS as well as the CPU (superseding the device scope of the reduced-precision entry
+  under **Bug fixes**), and floating-point results are not byte-identical to the former levelwise implementation.
+  Both extractors now also reject an image/LAF batch-size or dtype mismatch at the boundary with a clear error:
+  former releases raised an internal `grid_sample` error for most of these mixes, except `extract_patches_simple`
+  with a smaller LAF batch, which silently returned patches for only the first `min(B_img, B_laf)` images. Both extractors chunk their sampling along N, in the spirit of `batched_forward`, so
   the folded `(B, N*PS, PS, 2)` grid and its remap intermediates stay within a fixed 64 MiB budget: batched high-N
   extraction now peaks below even the pre-batching per-image loop (468 MiB against 495 MiB on `main` for B=8,
   N=4000, PS=32 on a 512x512 float32 image, where the unchunked atlas peaked at 1.24 GiB), eager throughput stays
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   128 MiB from becoming a memory hazard. LAFs whose nominal level is coarser than the last usable pyramid level now
   sample that actual coarsest level; they previously returned an all-zero patch. Atlas coordinate arithmetic also
   runs in `float32` for `float16` and `bfloat16` inputs, avoiding the normalized-coordinate precision loss caused by
-  the wider atlas. Floating-point results are therefore not byte-identical to the former levelwise implementation.
+  the wider atlas, and `extract_patches_simple` computes its sampling grid in `float32` for reduced-precision inputs
+  on every device for the same reason -- its output matches the pyramid extractor's level-0 output again, and
+  half-precision patches therefore change on CUDA and MPS as well as the CPU (superseding the device scope of the
+  reduced-precision entry under **Bug fixes**). Floating-point results are not byte-identical to the former
+  levelwise implementation. Both extractors chunk their sampling along N, in the spirit of `batched_forward`, so
+  the folded `(B, N*PS, PS, 2)` grid and its remap intermediates stay within a fixed 64 MiB budget: batched high-N
+  extraction now peaks below even the pre-batching per-image loop (468 MiB against 495 MiB on `main` for B=8,
+  N=4000, PS=32 on a 512x512 float32 image, where the unchunked atlas peaked at 1.24 GiB), eager throughput stays
+  within ~5% of the unchunked atlas, and float32/float64 patches are bitwise identical to it; a `torch.compile`d
+  call whose grid exceeds the budget splits into several `grid_sample` calls (~1.5x the unbounded atlas latency at
+  B=8, N=2000 on CPU inductor, still ~10x faster than the per-level loop it replaces). Both extractors return an
+  empty `(B, N, CH, PS, PS)` tensor for `B == 0` or `N == 0` instead of raising from inside `affine_grid` --
+  `extract_patches_from_pyramid` with `B == 0` returned the empty result before the batched rewrite below
+  regressed it; every other empty combination raised on all prior releases. `extract_patches_from_pyramid` now
+  also accepts a LAF on a different device than the image, as `extract_patches_simple` always has, and returns
+  patches on the image's device.
 
 * The shape guards of `normalize_homography` and `denormalize_homography` now reject everything but a
   `(3, 3)`/`(B, 3, 3)` matrix, and `normalize_homography3d`'s everything but a `(4, 4)`/`(B, 4, 4)` one (#3999).
--- a/benchmarks/feature/scale_space_detector.py
+++ b/benchmarks/feature/scale_space_detector.py
@@ -281,9 +281,10 @@ class KeyNetExtractor(nn.Module):
             # model and nms run at 6 different image sizes → dynamic=True avoids recompilation
             det.model = torch.compile(det.model, dynamic=True)
             det.nms = torch.compile(det.nms, dynamic=True)
-            # aff/ori/descriptor NOT compiled: extract_patches_from_pyramid (laf.py) contains
-            # an .item() call inside a while-loop that creates graph breaks, causing extra
-            # sub-graphs that need additional warmup and show as a spike on the first eval pair.
+            # aff/ori/descriptor NOT compiled: extract_patches_from_pyramid traces as one graph
+            # now, but each (num-LAF, patch-size) pair it is called with would still specialize
+            # into its own graph, needing additional warmup that shows as a spike on the first
+            # eval pair.
 
     @torch.no_grad()
     def forward(self, img: torch.Tensor):
--- a/kornia/feature/laf.py
+++ b/kornia/feature/laf.py
@@ -21,7 +21,7 @@ from typing import List, Optional, Tuple, Union
 import torch
 import torch.nn.functional as F
 
-from kornia.core.check import KORNIA_CHECK_LAF, KORNIA_CHECK_SHAPE
+from kornia.core.check import KORNIA_CHECK, KORNIA_CHECK_LAF, KORNIA_CHECK_SHAPE
 from kornia.geometry.conversions import angle_to_rotation_matrix, convert_points_from_homogeneous
 from kornia.geometry.linalg import transform_points
 from kornia.geometry.transform import pyrdown
@@ -416,13 +416,14 @@ def _clamp_grid_to_pixel_centers(grid: torch.Tensor, h: int, w: int) -> torch.Te
     exactly.
 
     Args:
-        grid: sampling grid :math:`(B, PS, PS, 2)`, last dimension ordered ``(x, y)``. Only the 2-D form is
-            handled; a 3-D grid would need a third bound from the depth of the sampled volume.
+        grid: sampling grid :math:`(..., 2)` with any leading batch dimensions, last dimension ordered
+            ``(x, y)``. Only 2-D sampling is handled; a 3-D grid would need a third bound from the
+            depth of the sampled volume.
         h: height of the sampled image.
         w: width of the sampled image.
 
     Returns:
-        the clamped grid :math:`(B, PS, PS, 2)`.
+        the clamped grid, same shape as ``grid``.
 
     """
     x = grid[..., 0].clamp(-1.0 + 1.0 / float(w), 1.0 - 1.0 / float(w))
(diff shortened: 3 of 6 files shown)


ducha-aiki:
## Re-measurement of the #4120 peak-memory finding against this branch

Context: [the #4120 review](https://github.com/kornia/kornia/pull/4120#issuecomment-5475680260) flagged a ~4x peak-memory regression in the pyramid path (finding 2). Since #4120 and this PR are intended to merge back-to-back, I re-measured the stacked result. Fresh process per revision, run from each worktree root (cwd shadowing the editable install, `kornia.__file__` checked), `resource.ru_maxrss` delta around a single `inference_mode` call. Config from the original finding: B=8, N=4000, PS=32, ch=1, 512x512 float32, level-spanning scales.

| revision | peak RSS delta | eager latency (median of 5) |
|---|---:|---:|
| main (`2009933e`) | 493 MiB | 2828 ms |
| #4120 (`9a0b5450`) | 1329 MiB | 2815 ms |
| #4125 stacked (`9d05580c`) | **1244 MiB** | **618 ms** |

And at N=8000 (same B/PS/image): main 786 MiB, this branch 2128 MiB.

**The speed claim holds** — 4.6x eager CPU here, consistent with the 4.06x in the PR table — **but the memory finding still stands**: ~2.5-2.7x main's peak, growing linearly with N. The reason the PR's own RSS table shows a *reduction* is that it measured N=2000 at (apparently) B=1, where the grid is ~16 MiB and the image-side costs the atlas genuinely improves dominate. At batched/high-N configs the peak is grid-driven, and the atlas path doesn't bound it:

- The `(B, N*PS, PS, 2)` float32 grid and a chain of same-sized/half-sized temporaries are live simultaneously in `extract_patches_from_pyramid`: the `affine_grid` output, `x`, `y`, the clamped `x`/`y`, the re-`stack`ed grid (the old `grid` is still bound while the RHS evaluates), `patches`, and the final `.contiguous()` copy. At B=8/N=4000 that is ~1 GiB of concurrent temporaries against main's one-image-at-a-time grid.
- The 128 MiB fallback gate counts only **atlas** bytes, which scale with the image, never with N — so it can't fire for the N-driven blowup. And the levelwise fallback wouldn't help anyway: it has the same folded-grid-per-level profile the #4120 finding was about.

Suggested mitigation, same as in the #4120 review but now cheap to do well: chunk the sampling along N with a compile-static chunk size (build the grid chunk, sample the atlas, write into a preallocated output — the atlas itself is the win and stays whole), and/or `del`/reuse the grid-chain temporaries, which roughly halves the peak on its own.

Two cross-checks on the other #4120 findings against this branch:

- **B=0 still raises** here (`ValueError: Expected non-zero, positive output size`) — the pyramid-path regression from #4120 (main returns `(0, N, ch, PS, PS)`) is not addressed by this PR, so the early-return fix still belongs in #4120 or here. N=0 raising is pre-existing on main (all revisions), not a regression.
- The float32-grid-for-half-dtypes claim is **genuinely implemented** in the atlas path (verified: fp16 input, fp32 grid arithmetic, finite output) — this PR delivers for the pyramid extractor what #4120's CHANGELOG promised. `extract_patches_simple` still generates its grid in the half dtype, so finding 3 of the #4120 review still applies there.

ducha-aiki:
## Pushed: rebase + chunked sampling to close the peak-memory finding

Two changes landed on this branch (with @ducha-aiki's go-ahead to rebase and modify):

**1. Rebased onto `perf/extract-patches-batched`'s tip** (`88be45b5`, was `9a0b5450`). The branch had forked before #4120's review-response commits, so it was missing the `.to(img.device)` cross-device restore, the `testing/base.py` probes, and ~105 lines of tests. The rebase surfaced a real stacking conflict: #4120's new `test_border_patches_stay_in_range` (pyramid) pins level-0 parity with `extract_patches_simple` for CPU half dtypes, and it fails against the atlas path — the atlas computes fp32 grids while `extract_patches_simple` still quantized its grid to fp16. Fixed below.

**2. `f020b669` — chunked sampling, in the spirit of `batched_forward`.** Both extractors now run their fold→`grid_sample`→unfold through one shared `_sample_patches` helper, and chunk it along N so the folded `(B, N·PS, PS, 2)` grid and its remap intermediates stay within a 64 MiB budget (`_grid_chunk_lafs`; the atlas remap constants are packed as (x, y) pairs so the chain rebinds one grid-sized tensor per step instead of holding split-axis copies). When every grid fits the budget the loop is a single iteration — the literal previous code path.

Measured (fresh process per revision, `ru_maxrss` delta, B=8, PS=32, 512² float32, level-spanning scales):

| revision | N=4000 peak / eager | N=8000 peak / eager |
|---|---:|---:|
| main | 495 MiB / 2846 ms | 786 MiB / 5629 ms |
| atlas unchunked (`6e89b15a`) | 1244 MiB / 624 ms | 1799 MiB / 1198 ms |
| **this branch** | **468 MiB / 650 ms** | **594 MiB / 1265 ms** |

Peak memory is now below `main` at both configs while keeping the atlas speedup (~4-5% eager cost when chunking engages). float32/float64 patches are **bitwise identical** to the unchunked atlas across a 10-case sweep of both extractors (shapes, dtypes, sheared LAFs, odd sizes), and `torch.jit.script(extract_patches_simple)` still works (the budget lives inside the scriptable helper — TorchScript cannot close over module-level ints).

One deliberate trade-off to flag: a `torch.compile`d call whose grid exceeds the budget splits into several `grid_sample` calls. At B=8, N=2000 (131 MiB grid) CPU inductor goes 106k → 72k LAFs/s — still ~10× the per-level loop's 7k, and I verified the restructure itself is compile-neutral (forcing a single chunk reproduces 148 ms vs the unchunked 150 ms). I prioritized the memory bound over peak compiled throughput at exactly the configs that OOM-risked before; raising the budget to 128 MiB flips that trade (full 111k LAFs/s at N=2000, but 688 MiB at the N=4000 config, 1.4× main) — happy to move it if the other preference wins.

Also folded in, from the [#4120 review](https://github.com/kornia/kornia/pull/4120#issuecomment-5475680260) and the [re-measurement above](https://github.com/kornia/kornia/pull/4125#issuecomment-5475756643):

- **B=0/N=0 return empty `(B, N, CH, PS, PS)`** instead of raising from `affine_grid` (finding 1; empty-in → empty-out convention). B=0 pyramid was a regression vs main; the other combinations raised on every release.
- **`extract_patches_simple` computes half-dtype grids in float32 on every device** — the remnant of finding 3, and what makes the merged branch's half border tests pass: its output matches the atlas path's level-0 output again. Half patches change on CUDA/MPS too; CHANGELOG updated accordingly.
- **The pyramid extractor accepts a LAF on a different device than the image** (finding 8's asymmetry, resolved by the shared helper's device move; pinned by a new `test_laf_on_another_device`).
- New tests: chunked-vs-single equivalence (bitwise on CPU, atlas + fallback), `test_channel_laf_correspondence` (the ch>1 ∧ N>1 discriminator finding 7 showed was missing), gradcheck and fullgraph dynamo through the multi-chunk path, empties.
- The stale `.item()` graph-break comment in `benchmarks/feature/scale_space_detector.py` (finding 10's prose sweep).

Validation: `tests/feature/test_laf.py` 156 passed (CPU fp32/64), 44 passed (CPU fp16/bf16 extractor tests), 59 passed (MPS fp32/16/bf16), inductor dynamo tests 4/4 including the new forced-chunk one, full `tests/feature` 642 passed, `pre-commit-all` clean, `ty` clean (one pre-existing dedode `custom_fwd` deprecation).

Still open from the earlier comments: nothing on the memory side; the remaining #4120-review items that live on the base branch (its own CHANGELOG wording, the `supports_grid_sample` over-skip, batch-mismatch `KORNIA_CHECK`, mixed-dtype policy) are unaffected by this push.

ducha-aiki:
## Combined review of #4120 + #4125 (the stack as it will merge: `f020b669` vs `main`)

Requested as a review of the two PRs together, since they are meant to merge in close succession. Method: a multi-angle review pass over this PR's delta (eight finder angles, deduplicated and verified), plus direct execution of every load-bearing claim against a worktree at `f020b669` and a `main` worktree (`kornia.__file__` checked in each probe; torch 2.9.1, CPU + MPS). One candidate finding was refuted during verification and is not listed.

**Verdict: not mergeable yet.** The stack genuinely delivers what it promises on the main path — speed, bounded memory, and correct atlas math (the remap constants were verified algebraically and numerically equivalent to `generate_patch_grid_from_normalized_LAF` at 2.5e-14) — but CI is red on every ubuntu job, and the fallback/reduced-precision edges break contracts the PR itself declares.

### What the stack resolves (from the [#4120 review](https://github.com/kornia/kornia/pull/4120#issuecomment-5475680260), re-verified on the combined tree)

- B=0 / N=0 return empty `(B, N, CH, PS, PS)` in both extractors (was must-fix 1).
- Peak memory is below `main` at the configs that regressed 4.1x (468 vs 495 MiB at B=8/N=4000; was must-fix 2), via the shared chunked `_sample_patches`.
- float32 sampling coordinates are genuinely computed for half dtypes in both extractors, so the CHANGELOG's accuracy claim is true in the final state (was must-fix 3), and the atlas entry explicitly supersedes the #4120 entry's device-scope claim — the stacked CHANGELOG reads correctly.
- The ch>1 AND N>1 fold/unfold discriminator test exists (finding 7), the atlas-path cross-device asymmetry is gone (finding 8 — but see must-fix 1 below), and the extraction duplication is hoisted into one helper (finding 10).

### Must fix

**1. CI is red: `test_atlas_guards_preserve_level_border_value_and_gradient` fails on every ubuntu job** — `tests/feature/test_laf.py:718`
All 12 ubuntu configs (torch 2.5.1 and 2.9.1, f32 and f64) fail on a bare `torch.equal` between tensors that print identically to 4 decimals — a low-order-bit, platform-dependent difference in the guard arithmetic; macOS and Windows pass ([example job](https://github.com/kornia/kornia/actions/runs/33374904847/job/99434247132)). The same test takes no `device`/`dtype` fixtures (the only one in the file), so even where it passes it exercises CPU default-dtype only, while the MPS branch of `_grid_sample_patches` uses a different padding mode entirely. Use the injected fixtures and `self.assert_close` (or pin the exact property with an explicit tolerance argument for why bitwise is expected).

**2. The levelwise fallback crashes on cross-device input and breaks the device contract this stack declares** — `kornia/feature/laf.py:514` (reproduced)
`out` and the `torch.where` mask live on the LAF's device while `_sample_patches` returns patches on the image's device: `extract_patches_from_pyramid(img_mps, laf_cpu, 16)` succeeds through the atlas path but raises `Expected all tensors to be on the same device` the moment the atlas exceeds 128 MiB — exactly the large-input case the fallback exists for, so the public function's behavior flips on input size. `test_laf_on_another_device` uses a 64x64 image and never reaches the fallback. The CHANGELOG explicitly promises the cross-device acceptance; allocate `out` on `img.device` and mask there (and extend the test to the forced-fallback path).

**3. fp16 CPU single-level images crash on torch <= 2.5.1: the atlas is padded in the image's native dtype** — `kornia/feature/laf.py:627`
`F.pad(cur_img, (1,1,1,1), mode="replicate")` runs on the fp16 image, and torch <= 2.5.1 has no Half CPU `replication_pad2d` (the gap `supports_replicate_padding` in `testing/base.py` documents). On base, a single-level fp16 CPU image (min(h,w) < 2*PS, so `pyrdown` never runs) extracted fine; on this stack it raises — and PR CI runs 2.5.1, where the half tests skip on the probe, so the job stays green while the behavior regresses. Building the atlas in `grid_dtype` (pad the float32-upcast image once) fixes this and must-fix 4 together.

**4. Chunked half-precision extraction re-casts the full atlas/image to fp32 once per chunk** — `kornia/feature/laf.py:452`
`_grid_sample_patches` casts the whole sampled tensor to the grid's dtype inside every per-chunk call (`img.to(grid.dtype)` / `img.float()`), and grids are now always fp32 for half inputs — so a half input near the atlas cap re-materializes a fresh up-to-256 MiB fp32 copy per chunk (levelwise: per chunk per level), defeating the 64 MiB budget in exactly the high-N regime chunking targets. On CUDA fp16 the full-image fp32 copy is entirely new relative to base's half-native sampling. The cast is loop-invariant — hoist it above the chunk loops.

### Should fix

**5. Mixed image/LAF dtypes are now silently accepted everywhere, and the two extractors disagree on the result dtype** (executed). On `main`, `extract_patches_simple(img_f32, laf_f64)` raises `expected scalar type Float but found Double`; on this stack it succeeds — and `extract_patches_simple` returns the **image's** dtype while `extract_patches_from_pyramid` returns the **LAF's** dtype (f16 img + f32 laf gives f16 vs f32). The widening is undocumented and the divergence is new public surface. Pick one policy (image dtype is the natural one — patches are image content), apply it to both extractors and their empties, and either `KORNIA_CHECK` equal dtypes or document the cast. Related: half CPU image + float64 LAF silently downcasts the grid f64→f32 through the now-shadowed CPU-half branch of `_grid_sample_patches` (`laf.py:448`) — the opposite of the other branches' compute-in-grid-dtype policy; collapse the two branches into one compute-dtype selection.

**6. The levelwise fallback samples all N LAFs at every pyramid level and discards the misses** — `laf.py:505`. O(levels x N) grid generation + `grid_sample` + `torch.where` on the huge-image path that exists to bound cost: detector output concentrated at levels 0-1 (typical) pays ~5x the necessary sampling at N=4000 over 6 levels. An eager `if not mask.any(): continue` per (level, chunk) removes most of it (gate on eager to avoid a compile graph break).

**7. The two new dynamo tests fail under `--optimizer=jit`** — `tests/feature/test_laf.py:897,906`. `test_dynamo_levelwise_fallback` and `test_dynamo_chunked` take `torch_optimizer` without the `optimizer_backend == "jit"` skip their sibling `test_dynamo` carries, and scripting `extract_patches_from_pyramid` fails (module-int closure, then `filter2d`'s set closure) — verified. Copy the sibling's skip; note the monkeypatched globals wouldn't apply under scripting anyway.

**8. Batch mismatch still raises opaque internal torch errors** (#4120 finding 4, executed on the tip). `extract_patches_simple(img[3], laf[1])` raises raw `grid_sampler(): expected grid and input to have same batch size`; the pyramid path now raises a *different* opaque message (`The expanded size of the tensor (1) must match...`). One `KORNIA_CHECK(img.size(0) == laf.size(0))` at each boundary.

**9. 1-pixel-wide/tall images now raise `ZeroDivisionError`** — `laf.py:646` (executed head vs main). The level-constants loop divides by `float(w_l - 1)` in Python, so `extract_patches_from_pyramid(torch.rand(1,1,5,1), nlaf, 4, False)` raises where `main` returned a (non-finite) patch without crashing. Degenerate but reachable from automated sliver crops; either guard with a clear `KORNIA_CHECK` or fall back to tensor division (non-finite, no crash) — the current bare Python exception is the worst of both.

### Minor / cleanup

**10. Remap fusion opportunity (numerics caveat).** The per-chunk remap runs six sequential full-grid passes plus a per-chunk `zeros`+`cat` for theta; `k`/`level_scale`/`t` are per-LAF and fold into theta before `F.affine_grid`, and since `level_scale > 0` the clamp commutes (`lo' = lo*scale+offset`), reducing to affine_grid + 2 clamp passes — ~4 fewer 64 MiB read+write passes per chunk at the budget cap. Caveat: float results shift by rounding, so the CHANGELOG's bitwise-identical-to-unchunked-atlas sentence would need updating; fine to defer, but decide deliberately.

**11. Level-geometry duplication.** The pyramid layout is replayed by three independent accumulators (heights loop, atlas-fill loop, constants loop) with the guard width `+2` hard-coded in seven places — they agree only by replaying identical arithmetic, and a drift shows up as patches sampling one pixel into the neighbouring level (an intensity error, not a crash). The half→fp32 promotion predicate is written three times and the `8 if float64 else 4` elem-size ternary at three call sites. Also consider a defaulted `budget` parameter on `_grid_chunk_lafs` (TorchScript-legal, unlike the module-int closure its docstring correctly rules out) so tests stop monkeypatching the whole function with `lambda *args: 1`.

**12. Stale prose from the #4120 review still standing:** `supports_grid_sample`'s docstring in `testing/base.py` names patch extraction as the failing consumer, but after this stack CPU half extraction never touches the half `grid_sample` kernel, and `test_flat_image_laf` still gates on the probe — green-by-skip on torch 2.5.1 CPU where it would now pass (finding 6). `_grid_sample_patches` still frames the fp32 detour as a "torch <= 2.9" workaround while applying it unconditionally (finding 9) — say the permanence is deliberate. `_clamp_grid_to_pixel_centers`'s docstring still describes a `(B, PS, PS, 2)` grid; callers now pass the 5-D form.

### Checked and clean

Atlas remap constants algebraically and numerically (2.5e-14) match `generate_patch_grid_from_normalized_LAF`; `pyrdown`'s floor(side/2) convention matches the precomputed level sizes; guard-pixel reachability and the giant-LAF coarsest-level clamp are correct; no in-repo caller relies on the removed all-zero patches for beyond-coarsest LAFs; `torch.jit.script(extract_patches_from_pyramid)` fails at the tip but already fails on `main` for a different closed-over global (no regression; `extract_patches_simple` scripts and matches eager); the empty-return dtype split mirrors each function's non-empty convention (refuted as a separate finding; it is part of item 5's policy question).

### Merge mechanics

Merge #4120 first; GitHub retargets this PR to `main`, and with squash merges the diff here should collapse to exactly the atlas+chunking delta — worth eyeballing before hitting merge. Nothing should merge #4120 alone: its branch still carries the B=0 crash, the memory regression, and the then-false CHANGELOG claim that this PR fixes. #4120's CI is fully green; this PR's is red per must-fix 1.

ducha-aiki:
## Review fixes applied: `2f52169b`

All four must-fix items and the actionable should-fix items from the [combined review](https://github.com/kornia/kornia/pull/4125#issuecomment-5476216888) are now on this branch, each verified by execution:

- **Red ubuntu test (must-fix 1):** the guard test now takes the `device`/`dtype` fixtures, compares the whole patch with standard tolerances, and pins the guard property on the patch **center** — measured: the fp32 remap rounding is ~4.8e-5 atlas pixels, so interior pixels legitimately differ between the atlas and levelwise paths at the same ~5e-5 scale a missing guard would leak at; the center is the discriminator because with the guard both of its bilinear partners are exactly the border value. The gradient probe moved a quarter-pixel outside the outermost pixel center: probing exactly at the clamp bound hit the differing CPU/MPS subgradient conventions (MPS returned -2495 where CPU returned 0).
- **Levelwise cross-device crash (must-fix 2):** the fallback accumulates on the image's device; the MPS repro passes and `test_laf_on_another_device` now pins the forced-fallback path too.
- **fp16 CPU crash on torch <= 2.5.1 (must-fix 3)** and **per-chunk fp32 recast (must-fix 4)**, one fix: the atlas — and the fallback's levels — are built in the grid dtype, upcast once before the loops; `_grid_sample_patches` is now a single compute path (the shadowed CPU-half branch and the silent f64→f32 downcast are gone). Verified in a fresh torch 2.5.1 venv: the old tip raises `"replication_pad2d" not implemented for 'Half'`, this branch passes fp16/bf16, single- and multi-level. Half-input peak memory is unchanged (410 → 409 MiB at B=8/N=4000/PS=32/512² fp16).
- **Mixed dtype policy (should-fix 5)** and **batch mismatch (8):** both extractors now `KORNIA_CHECK` equal batch sizes and dtypes at the boundary — the conservative option: `main` rejected every such mix too (except `extract_patches_simple` silently returning a partial result for a smaller LAF batch), just with opaque internal errors. CHANGELOG says so.
- **jit-backend skips (7)** added to both new dynamo tests; **sliver images (9)** route through the levelwise sampler (tensor division degrades to non-finite like `main`; the atlas constants' Python division raised `ZeroDivisionError`), pinned by `test_one_pixel_axis_does_not_raise`.
- **Prose (12):** `supports_grid_sample`'s docstring now says extraction must NOT gate on it; the over-skipping gates were dropped — on torch 2.5.1 the bf16 `test_flat_image_laf` and the pyramid border tests now run and pass instead of green-by-skip. `_grid_sample_patches` states the float32 detour is deliberate on every torch version; `_clamp_grid_to_pixel_centers` documents the `(..., 2)` shape.

**Deferred, deliberately:** the remap fusion (10) — it breaks the bitwise-vs-unchunked-atlas guarantee, worth its own decision; the levelwise `mask.any()` skip (6) — a data-dependent early-exit is exactly what the compile-first workflow bans, needs a design that keeps one code path; the full level-geometry dedup (11) — the `budget` parameter and `_grid_elem_bytes` landed, the three-accumulator restructure is follow-up material.

Validation: fp32/f64 **byte-identical to `f020b669`** across a 24-case sweep of both extractors; `tests/feature` 645 passed (CPU fp32); test_laf 163 passed CPU fp32/64, 49 passed CPU fp16/bf16 (on torch 2.9.1 **and** 2.5.1), 68 passed MPS fp32/16/bf16; inductor dynamo 4/4; `pre-commit-all` and `ty` clean (one pre-existing lightglue deprecation).

**Housekeeping:** #4120 is closed — this branch contains it verbatim, and merging the stack as one PR removes the intermediate broken state on `main`. GitHub refuses to retarget this PR to `main` ("part of a stack", via REST and `gh pr edit` both), so the combined change moves to a fresh PR from this same branch; link follows there.

ducha-aiki:
Superseded by the combined PR from the same branch (retarget to `main` is blocked by GitHub's stack constraint — see the comment above). Nothing is lost: the new PR carries this branch byte-for-byte, review history stays readable here.

ducha-aiki:
Combined PR: #4128.

PR #4128, opened 2026-08-31
perf(feature): batch patch extraction and sample the pyramid from a packed atlas

ducha-aiki:
Combines #4120 and #4125 into one PR (their stack could not be retargeted: GitHub refuses base changes on a stacked PR even after the base PR closes). This branch is `perf/extract-patches-batched` (#4120) + the pyramid atlas (#4125) + the fixes from the [combined stack review](https://github.com/kornia/kornia/pull/4125#issuecomment-5476216888), merged as one so no intermediate broken state ever lands on `main`.

## What it does

`extract_patches_simple` and `extract_patches_from_pyramid` replace their per-image Python loop with one batched `grid_sample` over a folded `(B, N·PS, PS, 2)` grid (#4120), and the pyramid extractor samples ordinary-sized inputs **once from a packed pyramid atlas** instead of sampling every LAF at every level (#4125). Sampling is chunked along N with a 64 MiB grid budget, so batched high-N extraction peaks **below** the old per-image loop while keeping the atlas speedup.

| B=8, PS=32, 512² fp32 | peak RSS | eager CPU |
|---|---:|---:|
| `main`, N=4000 | 495 MiB | 2846 ms |
| **this PR**, N=4000 | **468 MiB** | **650 ms** |
| `main`, N=8000 | 786 MiB | 5629 ms |
| **this PR**, N=8000 | **594 MiB** | **1265 ms** |

Under `torch.compile` the extractor traces as a single graph independent of batch size (the old loop unrolled one `grid_sample` per image and recompiled per batch size); CPU inductor at B=8/N=2000 runs ~10× the per-level loop. A compiled call whose grid exceeds the budget splits into several `grid_sample` calls (~1.5× the unbounded atlas latency at that config) — the memory bound was chosen over peak compiled throughput.

## Correctness and compatibility

- **float32/float64 results:** `extract_patches_simple` is byte-identical to `main`; the pyramid extractor's floating-point results change (atlas remap arithmetic, verified equivalent to the reference grid at 2.5e-14) and LAFs coarser than the last usable level now sample that level instead of returning all-zero patches.
- **Reduced precision:** the fp16/bf16 CPU `grid_sample` kernels in torch ≤ 2.9 read out of bounds on border-straddling patches (NaN/garbage through the public API — found in #4120). Both extractors now run the whole half-precision pipeline — pad, `pyrdown`, coordinates, sampling — in float32 and cast back, on every device. This fixes the OOB read, the whole-pixel coordinate quantization on large images, and a `"replication_pad2d" not implemented for 'Half'` crash on torch ≤ 2.5.1 (verified in a 2.5.1 venv). Half patches change on all devices; CHANGELOG documents it.
- **Contracts:** empty B/N returns empty; a LAF on another device works on both pyramid paths; image/LAF batch-size mismatches are rejected clearly; mixed dtypes are promoted for grid arithmetic while patches retain the image dtype/device; non-finite LAF frames return zero patches and zero LAF gradients on **both** extractors, and their backward no longer reaches `grid_sample` with an invalid grid (which segfaults torch's CPU border-padding kernel); singleton image axes count as one pixel of extent in `normalize_laf`/`denormalize_laf`/`generate_patch_grid_from_normalized_LAF`, so both extractors stay finite on the default pixel-LAF path instead of raising `ZeroDivisionError`, and singleton pyramid levels use the levelwise fallback.
- **TorchScript:** `extract_patches_simple` still scripts and matches eager. The pyramid extractor was never scriptable on `main` (module-level closure in `filter2d`) and still is not.

## Review trail

- #4120 review: [must-fix items](https://github.com/kornia/kornia/pull/4120#issuecomment-5475680260) → fixed here (B=0, peak memory, fp32-coordinates claim).
- #4125 memory re-measurement: [issuecomment-5475756643](https://github.com/kornia/kornia/pull/4125#issuecomment-5475756643) → chunking (`f020b669`).
- Combined stack review: [issuecomment-5476216888](https://github.com/kornia/kornia/pull/4125#issuecomment-5476216888) → fixes in `2f52169b`, itemized in [issuecomment-5476474211](https://github.com/kornia/kornia/pull/4125#issuecomment-5476474211). Deferred with rationale: remap-pass fusion (breaks the bitwise-vs-unchunked guarantee), levelwise per-level `mask.any()` skip (data-dependent early-exit vs. the compile-first rule), level-geometry accumulator dedup.

Validation after merging current `main`: `tests/feature` 667 passed (CPU fp32); the four directly affected modules 371 passed (CPU fp32/fp64); targeted CPU and CUDA fp16/bf16 follow-up checks each passed 15 with 1 intentional skip; the CUDA fp32 run passed 245 with 24 skips and one orientation tolerance failure reproduced with identical values on current `main`; `pre-commit-all` passed; `typecheck` exited 0 with only the existing deprecation diagnostic in unchanged `lightglue.py`.


ducha-aiki:
## Review: CUDA tests + benchmarks + deep multi-agent review

**Verdict: strong work — the core claims hold up on CUDA — but two crash bugs need fixing before merge, and there is an undisclosed CUDA fp16 slowdown in `extract_patches_simple`.**

Verified on an RTX 4090, torch 2.13.0+cu130: the pyramid extractor is **4.2–4.6× faster** in fp32, `extract_patches_simple` fp32 output is **bit-identical to `main` on CUDA**, and the pyramid diff vs `main` is ~9e-5 in fp32 (expected remap rounding).

### Must-fix findings (1 and 2 reproduced directly; the rest verified by a multi-agent review pass)

1. **NaN LAF crashes the atlas path — a regression.** `pyr_idx` is clamped only with `max=`, so NaN → `.long()` → INT64_MIN indexes `level_constants` out of bounds. On CPU it is an `IndexError: index -9223372036854775808 is out of bounds`; on CUDA it is a device-side assert that poisons every later CUDA op in the process. `main` returned all-zero patches, and the levelwise fallback still does — so behavior flips with image size. Reached from `LAFOrienter`/`LAFAffNetShapeEstimator`/`LAFDescriptor` whenever a training-time detector emits a NaN frame. One-line fix: `pyr_idx.clamp(min=0, max=len(heights) - 1)`.
2. **`PS=1` raises `ZeroDivisionError`** in the atlas level constants (`float(w_l - 1)` with a 1-pixel level; e.g. a 12×12 image builds `heights == [12, 6, 3, 1]`). The `min(h, w) < 2` guard checks the *input* image, but the division comes from the *derived level* sizes. `main` also crashed here (with a different error), so not a regression — but the guard should test the built levels, e.g. `min(heights[-1], widths[-1]) < 2`, and let the levelwise fallback handle it (it already does).
3. **Cross-device path builds the full grid on the LAF's device.** With LAF on CPU and image on CUDA, the whole B·N·PS²·2 remap runs on CPU: measured 100 ms/call vs 41 ms when `nlaf` is moved to `img.device` right after the `KORNIA_CHECK`s — and results differ ~2.6e-5 from the same-device call because CPU/CUDA round the remap differently. Move the tiny (B,N,2,3) LAF once instead of the grid.
4. **CHANGELOG inaccuracy:** `main`'s `extract_patches_from_pyramid` also silently returned a partial batch on a batch mismatch (verified on `main`: `(3,1,32,32)` image + `(1,2,2,3)` LAF returns shape `(1,2,1,8,8)`), not an internal error as the breaking-changes entry claims. Reword to cover both extractors.
5. **Dead probes in `testing/base.py`:** `supports_reflect_padding` lands with zero callers, and the PR orphans `supports_grid_sample` by removing its only call site. The reflect-pad docstring asserts a live gap, inviting the next half-precision test author to gate on it and over-skip — exactly the pattern this PR removed. Either wire them into tests or drop both.
6. Minor: the 64 MiB chunk budget counts only the folded grid, not the ch-scaled `grid_sample` output or remap temporaries (~2.5× real peak for RGB — soften the claim or include `ch`); the levelwise fallback clamps the grid twice per chunk (`_clamp_grid_to_pixel_centers` + border padding is the same clamp), wasting full-grid temporaries on the memory-constrained path; `_sample_patches`' docstring claims both extractors "have always accepted" a cross-device LAF, but that is true only of `extract_patches_simple` (the CHANGELOG itself bills it as new for the pyramid extractor).

Positively verified, not flagged: the atlas remap constants and (x,y) packing (differential-tested to ~1e-14 against `main` across ~70 configs, forward and backward), chunking bitwise-equivalence, empty-B/N handling, guard-band geometry, `pyrdown` shape agreement, and all in-tree callers against the new dtype/batch preconditions.

### CUDA test runs (RTX 4090, torch 2.13.0+cu130, editable install)

- `tests/feature` full suite, CUDA fp32: **644 passed**; the one failure (`TestOrientationHalfPrecisionIsFinite::test_half_precision_matches_float32`) fails identically on `main` — pre-existing on this torch build, unrelated to the PR.
- CUDA half-precision with `--isolate-half-precision`: all new PR tests pass; the 18 failures are `TestTransformLAFs::test_transform_points` fp16/bf16, failing identically on `main` (pre-existing torch 2.13 tolerance issue, worth a separate look).
- CUDA inductor dynamo tests: **4/4 pass**, including forced-chunk and forced-fallback.

### CUDA A/B benchmark (B=8, PS=32, 512² image, `benchmarks/common.time_us` median, `main` worktree vs this branch, `kornia.__file__` checked in every block)

| config | `main` | PR | peak CUDA mem (`main` → PR) |
|---|---:|---:|---|
| fp32 simple N=2000 | 4.8 ms | 4.7 ms | 159 → 297 MiB |
| fp32 simple N=8000 | 18.9 ms | 18.6 ms | 583 → 494 MiB |
| fp32 pyramid N=2000 | 26.2 ms | **6.3 ms** | 146 → 313 MiB |
| fp32 pyramid N=8000 | 95.0 ms | **20.8 ms** | 523 → 511 MiB |
| fp16 simple N=2000 | 3.0 ms | 4.8 ms | 84 → 270 MiB |
| fp16 simple N=8000 | 8.2 ms | **18.4 ms** | 296 → 372 MiB |
| fp16 pyramid N=2000 | 17.1 ms | **6.4 ms** | 77 → 286 MiB |
| fp16 pyramid N=8000 | 41.7 ms | **20.7 ms** | 266 → 389 MiB |

Two things the PR description does not surface:

- **CUDA fp16 `extract_patches_simple` gets ~2.2× slower** (stable across repeat runs). The universal fp32 upcast routes around a *CPU-only* kernel bug; CUDA's fp16 `grid_sample` is fine, so consider upcasting only where the kernel is broken or the coordinate quantization actually bites — or at minimum document the CUDA fp16 cost.
- **CUDA peak memory rises at moderate N** (atlas + upcast): fp32 pyramid N=2000 goes 146 → 313 MiB. The PR's memory table is CPU RSS at high N, where the PR indeed wins.

ducha-aiki:
All findings from the [review above](https://github.com/kornia/kornia/pull/4128#issuecomment-5477545152) are addressed in 870602d6.

**1. NaN LAF crash** — fixed, with one deliberate deviation from the suggested one-liner: `pyr_idx.clamp(min=0)` alone stops the crash, but `grid_sample` with a NaN grid under border padding returns a border-sampled *value*, not zero — so the atlas path would still disagree with the levelwise fallback's all-zero patches, and behavior would keep flipping with image size. Instead a NaN scale is marked level `-1` (`nan_to_num` before `.long()`), and the atlas path zeroes those patches with one unconditional `torch.where` after the chunk loop (no data-dependent branch, so fullgraph compile is preserved; the levelwise mask already treats `-1` as "no level"). Both paths now return an all-zero patch for a NaN frame, matching `main`. Regression test `test_nan_laf_returns_zero_patch` exercises both paths, on CUDA too.

**2. `PS=1` ZeroDivisionError** — the guard now inspects the built level sizes (`min(heights[-1], widths[-1]) < 2`) rather than the input image, routing 1-pixel levels to the levelwise fallback. Test `test_one_pixel_level_does_not_raise` pins the 12×12/PS=1 case.

**3. Cross-device grid on the LAF's device** — both extractors now move the `(B, N, 2, 3)` LAF to the image's device right after normalization; `_sample_patches` no longer moves the grid. Verified on CUDA: the cross-device result is now bitwise identical to the same-device call.

**4. CHANGELOG batch-mismatch claim** — reworded: both extractors silently returned the first `min(B_img, B_laf)` images on `main`.

**5. Dead probes** — `supports_reflect_padding` and the orphaned `supports_grid_sample` are removed from `testing/base.py` (confirmed zero callers).

**6. Minors** — the chunk-budget docstring now states the 64 MiB counts only the folded grid and the real peak is a small multiple of it; the levelwise fallback's redundant per-chunk clamp is gone (border padding performs the identical clamp — verified bitwise); the `_sample_patches` cross-device docstring claim is corrected.

**fp16 CUDA slowdown** — kept the universal upcast rather than making it CUDA-conditional: the coordinate-quantization rationale applies on every device, and a device-conditional dtype would make CPU and CUDA half results diverge. The ~2× CUDA fp16 `extract_patches_simple` cost and the moderate-N CUDA memory increase are now documented in the CHANGELOG and in `_grid_sample_patches`.

Verification: normal-input outputs bitwise identical to the pre-fix revision (fp32/fp64/fp16 × CPU/CUDA × both extractors, differential-tested against the old module loaded side by side); full `tests/feature/test_laf.py` on CPU, CUDA, CPU half, and CUDA half with `--isolate-half-precision`; 4/4 inductor dynamo tests on CPU and CUDA; downstream orientation/SIFT/detector tests; pre-commit clean. The `TestTransformLAFs::test_transform_points` half failures and the flaky isolation-harness teardown error reproduce identically on the unmodified branch (pre-existing).

ducha-aiki:
## Final review — 8 findings, all confirmed by executed verification

Every finding below was verified by actually running code (A/B against a `main` worktree, gradcheck, `torch.compile(fullgraph=True)`, CPU and CUDA), not just by reading the diff. Ranked by severity.

### Correctness

**1. `kornia/feature/laf.py:594` — the new `KORNIA_CHECK(img.dtype == laf.dtype)` regresses mixed-precision autocast callers that worked on main.**
Verified A/B against a `main` worktree: under `torch.autocast('cpu', dtype=torch.bfloat16)`, `extract_patches_from_pyramid(img_f32, laf_bf16, 8)` succeeds on main (`grid_sample` promotes dtypes under autocast) but raises `img and laf must have the same dtype` on this branch; same for `extract_patches_simple` (line 545). `LAFOrienter` (`orientation.py:279`), `LAFAffineShapeEstimator`/`LAFAffNetShapeEstimator` (`affine_shape.py:158, 245`) and `get_laf_descriptors` (`integrated.py:89`) forward user `img`+`laf` straight in, so an autocast pipeline whose reduced-precision net emits fp16/bf16 LAFs against a float32 image goes from working to raising. Consider comparing dtypes only outside autocast, or casting the LAF the way the device move already does, instead of rejecting.

**2. `kornia/feature/laf.py:717` — NaN-LAF safety is forward-only: backward through the NaN sampling grid segfaults the process on both sampling paths.**
Verified twice independently: `img = rand(1,1,64,64, requires_grad=True)` with one all-NaN 2×3 LAF frame returns the advertised zero patch, but `out.sum().backward()` exits with signal 139 (`grid_sampler_2d_backward` with a NaN border-padding grid, torch 2.13 CPU) on the atlas path and the levelwise fallback alike. This is pre-existing on main and reduces to an upstream torch bug (`padding_mode='border'` CPU backward with a NaN grid; the default `'zeros'` doesn't crash) — but this diff introduces the level `-1` marking specifically because "a training-time detector can emit a degenerate LAF" (line 610), and a training-time detector will call backward. Sanitizing the grid rows where `pyr_idx < 0` (`torch.where`/`nan_to_num` on the grid before `_sample_patches`, on both paths) would make the advertised safety hold through backward.

### Efficiency

**3. `kornia/feature/laf.py:718` — the final `torch.where` NaN-zeroing allocates and copies a second full `(B, N, ch, PS, PS)` output on every call, even with no NaN LAFs.**
`torch.where((pyr_idx >= 0).view(B, N, 1, 1, 1), out, out.new_zeros(1))` materializes an extra output-sized tensor (~125 MiB for B=8, N=4000, PS=32 float32) — the same peak the chunk budget exists to bound. Verified drop-in replacement: `out.masked_fill_((pyr_idx < 0).view(B, N, 1, 1, 1), 0.0)` gives bitwise-identical forward output, bitwise-identical `img`/`laf` grads, passes float64 gradcheck on both paths, and traces under `torch.compile(fullgraph=True)` with bitwise-equal compiled output.

**4. `kornia/feature/laf.py:654` — the full atlas machinery runs even when the pyramid has exactly one level, where it only reproduces the level-0 grid.**
A 48×48 image with PS=32 gives `len(heights) == 1` (any input with `min(h, w) < 2*PS` — common for patch-dataset crops and small ROIs), `pyr_idx` statically clamps to 0, yet the code still allocates the `(B, ch, h+2, w+2)` atlas, replicate-pads the whole image into it, and applies the 5-op per-chunk remap. Verified that the result matches direct border-padded sampling of `sample_img` to fp noise (5.9e-6 float32, 1.1e-14 float64, CPU and CUDA, including out-of-frame patches). A static `if len(heights) == 1` Python branch sampling `sample_img` directly (keeping the `pyr_idx >= 0` zeroing) skips all of it and stays fullgraph-safe like the existing line-651 guard; note it would perturb existing atlas-path outputs at the ~1e-6 float32 level.

**5. `kornia/feature/laf.py:519` — the levelwise fallback rebuilds the level-independent affine grid at every pyramid level.**
Verified from `generate_patch_grid_from_normalized_LAF` (laf.py:381-404): the per-level grid equals the zero-translation affine grid times the per-level scalar `k = 2*min(h_l-1, w_l-1)/(size_l-1)` plus the level-independent `t = 2*nlaf[..., :, 2] - 1` — exactly the arithmetic the atlas path already uses (lines 713-716), holding exactly per level including non-square floor-halved sizes. For the large images that trigger the fallback (atlas > 128 MiB; a 4096×4096 float32 image has 7 levels), ~6/7 of the grid-construction launches are redundant. Computing the base grid once per chunk and applying the per-level scalars would match the atlas path; modest win since the per-level `grid_sample` over all N dominates.

### Consistency / simplification

**6. `kornia/feature/laf.py:637` — the reduced-precision→float32 grid-dtype policy is spelled out independently three times, and `_grid_elem_bytes` encodes it implicitly a fourth time.**
`extract_patches_simple` (line 558), `_extract_patches_from_pyramid_levelwise` (line 510), and the atlas path (line 637) each re-derive "float32 if dtype in (float16, bfloat16)", and `_grid_elem_bytes` ("8 if float64 else 4", line 453) is only correct because of that upcast. The CHANGELOG itself flags the universal upcast as a real 2× cost on CUDA float16, so device-conditional upcasting is a plausible future change; whoever makes it must edit three sites plus keep `_grid_elem_bytes` consistent — missing the `extract_patches_simple` copy silently breaks the tested level-0 equality invariant, missing `_grid_elem_bytes` doubles the real per-chunk memory against the budget. A one-line `_grid_dtype(dtype)` helper next to `_grid_elem_bytes` would single-source the policy.

**7. `kornia/feature/laf.py:651` — reading the module-level `_PYRAMID_ATLAS_MAX_BYTES` contradicts the TorchScript accommodation `_grid_chunk_lafs` documents two helpers away, and the tests' jit-skip comments misattribute the blocker.**
Verified with `torch.jit.script`: scripting `extract_patches_from_pyramid` now fails first on this global ("python value of type int cannot be used as a value... closed over global variable") before ever reaching `filter2d`, while `_grid_chunk_lafs` (line 456) deliberately carries its 64 MiB budget as a defaulted argument "because TorchScript cannot close over module-level ints", and the jit-skip comments in `tests/feature/test_laf.py` attribute non-scriptability solely to `filter2d`. Not a regression (main already failed on `filter2d`), but if `filter2d` is ever made scriptable this line becomes the next blocker for a reason this diff introduced, and the module now carries two byte budgets in two opposite styles. Make the cutoff a defaulted parameter like its sibling (it stays monkeypatchable via the call site) or fix the docstring/skip comments.

**8. `kornia/feature/laf.py:636` — `pyr_idx` is clamped twice, and the first max-bound is always the looser one, hence dead.**
The heights loop runs at most `max_level - 1` append iterations, so `len(heights) - 1 <= max_level - 1` always, and only the line-636 clamp binds. Two clamps encoding one invariant drift under maintenance (a changed break condition in the heights loop silently changes which bound is live). Hoisting the heights/widths loop (pure Python ints from `h, w, PS`) above the `pyr_idx` computation and clamping once to `float(len(heights) - 1)` in the same expression that handles NaN is behavior-identical, including inf handling and the fullgraph path.

---

Also checked and deliberately **not** flagged: the batch-size check hardening, the giant-LAF semantic change, budget configurability, the dual-path design as such, `_grid_elem_bytes` vs `torch.finfo`, and the long CHANGELOG line — intentional or below the precision bar. An audit of removed behavior came back clean.

ducha-aiki:
Addressed all findings from the final review in d85af0ec.

- Mixed image/LAF dtypes are supported again, including autocast pipelines. Grid arithmetic uses a promoted dtype so a float32 LAF is never rounded down merely because the image is half precision; patches still return in the image's dtype and device.
- Non-finite LAF frames are detected and sanitized before normalization, scale selection, or grid construction. Atlas and levelwise paths now return zero patches and zero LAF gradients for NaN/Inf frames, and backward never reaches `grid_sample` with an invalid grid.
- Replaced the unconditional output-sized `torch.where` copy with in-place masking.
- Single-level pyramids sample the image directly instead of building an atlas.
- The levelwise fallback reuses its affine grid across levels when the call fits one bounded chunk, while multi-chunk calls keep the streaming, bounded-memory order.
- Chunk sizing now bounds both the 2-coordinate grid and the channel-scaled sampled temporary, avoiding the high-channel memory hole without changing the chunk size for one- or two-channel inputs.
- Sampling dtype policy and byte accounting are single-sourced; the atlas budget is now a TorchScript-friendly defaulted helper; pyramid sizes are built before one final level-index clamp.
- The changelog now describes mixed-dtype promotion and the actual memory budget accurately.

Verification performed on the pushed tree:

- `tests/feature/test_laf.py`, CPU float32/float64: 172 passed, 18 skipped.
- CPU float16/bfloat16 focused coverage: 8 passed.
- Orientation, affine-shape, and integrated downstream tests: 76 passed, 14 skipped.
- Inductor/fullgraph extraction tests: 4/4 passed, including forced fallback and chunking.
- `pixi run pre-commit-all`: clean.
- `pixi run typecheck`: no errors; only the existing deprecation diagnostic in unchanged `lightglue.py`.
- `torch.jit.script(extract_patches_from_pyramid)` now reaches the known `filter2d` `_VALID_BORDERS` blocker instead of failing on the atlas budget.
- Local A/B with verified imports: forced fallback improved from 93.6 ms at the prior PR tip to 50.5 ms; the representative peak-RSS comparison did not regress.

CUDA hardware was not available for this final local pass; the new reduced-precision test uses the global dtype fixture so CUDA half jobs execute it under the suite's subprocess isolation.

ducha-aiki (commented):
## Review of `d85af0ec` — one merge blocker remains

The latest implementation fixes hold up in focused CPU and CUDA verification. I found one semantic conflict with current `main` that Git cannot detect, plus documentation that still describes an obsolete intermediate state.

### Must fix

1. **Rebasing deletes a test helper that current `main` now imports.**

   This PR deletes `supports_grid_sample` from `testing/base.py` because it had no callers at the PR's original base (`2009933e`). Current `main` is now `1331a861` and #4122 added callers in:

   - `tests/feature/test_affine_shape_estimator.py`
   - `tests/feature/test_scale_space_detector.py`

   `git merge-tree` reports only the textual `CHANGELOG.md` conflict, so resolving that conflict alone leaves a semantic break: collection raises `ImportError: cannot import name 'supports_grid_sample' from 'testing.base'`. I reproduced the missing import directly and confirmed all current-main call sites with `git grep`.

   Because this PR makes patch extraction sample reduced-precision inputs in float32, the best rebase resolution is probably to remove the newly stale grid-sample gates/imports from those #4122 tests (and update their comments), rather than restoring a probe those paths no longer need. Either way, the helper deletion and all new callers need to be reconciled together, followed by collection/runs of the affected feature modules.

### Documentation

2. **The Unreleased changelog contains mutually contradictory final behavior.**

   `CHANGELOG.md:16-27` correctly says half inputs are sampled in float32 on every device, CUDA/MPS outputs change, and CUDA fp16 can be ~2x slower. But `CHANGELOG.md:118-128` still says only CPU is upcast and that CUDA/MPS results are byte-for-byte unchanged with unchanged eager throughput. The “superseding” aside does not make the latter statement true for the release; please consolidate or rewrite the older bullet. The PR description is stale in the same way: it still says image/LAF dtype mismatches are rejected, while `d85af0ec` now promotes them.

### Verification on the current PR head

- CPU `tests/feature/test_laf.py`, float32/float64: **172 passed, 18 skipped**.
- CUDA `tests/feature/test_laf.py`, float32: **118 passed, 13 skipped** on RTX 4090 / torch 2.13.0+cu130.
- CUDA isolated fp16/bf16 coverage for mixed-dtype autocast, non-finite-LAF backward, and level-zero float-reference behavior: **6 passed**.
- Additional atlas-vs-levelwise sweep: 72 configurations across float32/float64, odd and non-square sizes, PS 3/8/16, and 1/3 channels; all passed, worst max difference **6.26e-6**.
- `git diff --check`: clean. GitHub CI is green at this head.

I found no additional implementation correctness regression in the reviewed head beyond the current-main integration blocker above.

Posted on behalf of @ducha-aiki by Codex (gpt-5.6-sol).

ducha-aiki:
Follow-up fixes are pushed in `092b9fdb`.

- Merged current `main`; the PR is now reported as MERGEABLE.
- Removed the obsolete `supports_grid_sample` imports and skip gates added by `main`, so the reduced-precision tests collect and exercise the new float32 sampling path.
- Reconciled the CHANGELOG conflict while retaining the upstream `ellipse_to_laf` entry and correcting the extractor compatibility/performance wording.
- Updated the PR contract and validation sections to match the implemented mixed-dtype, non-finite-LAF, and singleton-axis behavior.

Fresh validation after the merge:

- `tests/feature --device=cpu --dtype=float32`: 667 passed, 81 skipped.
- Four directly affected modules on CPU float32/float64: 371 passed, 37 skipped.
- Targeted CPU fp16/bf16 follow-up checks: 15 passed, 1 intentional skip.
- Targeted CUDA fp16/bf16 follow-up checks: 15 passed, 1 intentional skip.
- Four directly affected modules on CUDA float32: 245 passed, 24 skipped, with one orientation tolerance failure. I reproduced that failure with identical output in an isolated worktree at current `origin/main`, so it is not introduced by this PR.
- `pre-commit-all`: all hooks passed.
- `typecheck`: exit 0; only the existing deprecation diagnostic in unchanged `kornia/feature/lightglue.py`.

The local benchmark JSON files were left untouched.

ducha-aiki (commented):
## Review of `092b9fdb` — three remaining gaps

The finite-input implementation and its focused CPU/CUDA float32 coverage hold up, but I found two contract gaps and one failure in the prescribed CUDA half-precision suite.

### 1. Must fix: `extract_patches_simple` still segfaults on non-finite LAF backward

`extract_patches_from_pyramid` sanitizes invalid frames at `kornia/feature/laf.py:709-717`, but `extract_patches_simple` proceeds directly from normalization to `generate_patch_grid_from_normalized_LAF` at lines 641-657. On CPU with torch 2.13, a LAF whose center is NaN produces a finite-looking, nonzero forward patch and then `out.sum().backward()` terminates Python with signal 139 inside `grid_sampler_2d_backward`.

This leaves the PR description's contract—"non-finite LAFs return zero patches and zero LAF gradients"—true only for the pyramid extractor. Please mirror the pyramid path's pre-grid invalid-frame sanitization and output/gradient masking in the simple extractor, with a subprocess regression test so a failure cannot take down pytest itself.

### 2. Must fix: the new giant-LAF test fails the supported CUDA float16 run

`tests/feature/test_laf.py:940` constructs the LAF with `1.0e5` in the injected dtype. That value overflows to `inf` in float16, so the implementation correctly follows its new non-finite contract and returns a zero patch, while the reference simple extractor still border-samples a nonzero patch. The exact repository-style invocation fails 256/256 elements (greatest absolute difference 15664):

```bash
pixi run test-module tests/feature/test_laf.py::TestExtractPatchesPyr::test_giant_laf_uses_actual_coarsest_level \
  --device=cuda --dtype=float16 --isolate-half-precision
```

A float32 LAF paired with the half image would test both the intended giant-*finite*-LAF behavior and the mixed-dtype path without contradicting the non-finite test immediately above it. A representable half value is another option.

### 3. Contract/documentation gap: singleton image axes still fail through the public default path

Both extractors call `normalize_laf` before any singleton-axis handling (`laf.py:641-644` and `686-689`). With the default `normalize_lafs_before_extraction=True`, `(H, W) = (5, 1)` and `(1, 5)` therefore still raise `ZeroDivisionError` from `1.0 / (size - 1)` in `normalize_laf`. The new `test_one_pixel_axis_does_not_raise` passes `False`, so it exercises only callers that already supply normalized LAFs, while the PR description broadly claims singleton image axes use the finite fallback.

Either make the default pixel-LAF path finite too, or narrow the stated contract and test name/comment to the normalized-LAF-only case.

### Verification

- PR head and local head both: `092b9fdb2404cd0260e8f5dc62ddb4035935aec4`.
- CPU `tests/feature/test_laf.py`, float32/float64: **187 passed, 18 skipped**.
- CUDA `tests/feature/test_laf.py`, float32: **126 passed, 13 skipped**.
- Focused CUDA float16/bfloat16 extractor run: **56 passed, 8 skipped, 1 functional failure above**, plus the known isolation-harness `previous item was not torn down properly` error on a dtype-independent test.
- `git diff --check`: clean; current GitHub checks are green but do not include the CUDA half suite.

Posted on behalf of @ducha-aiki by Codex (gpt-5.6-sol).

ducha-aiki:
Addressed all three findings from the [review of `092b9fdb`](https://github.com/kornia/kornia/pull/4128#issuecomment-5480167253) in `0e86da88`.

**1. `extract_patches_simple` non-finite backward** — reproduced first (`signal 139` in `grid_sampler_2d_backward`, torch 2.9.1 CPU as well as 2.13), then fixed by mirroring the pyramid path: the frame is marked invalid from `~isfinite(nlaf).all(-1).all(-1)` before any grid arithmetic, `nlaf` is `masked_fill`ed to zero, and the output is zeroed with the same unconditional in-place `masked_fill_` after the chunk loop, so the fullgraph path takes no data-dependent branch. Both extractors now return a zero patch and a zero LAF gradient for a frame whose only invalid entry is its center.

Two tests: `TestExtractPatchesSimple::test_nonfinite_laf_returns_zero_patch_with_safe_backward` mirrors the pyramid one in-process, and a new module-level `test_nonfinite_laf_backward_does_not_crash_the_interpreter` runs both extractors' backward in a fresh interpreter and asserts the exit code, so a regression fails a test instead of taking pytest down.

**2. Giant-LAF test in CUDA float16** — the LAF is now spelled in `float32` rather than the injected dtype. `1e5` overflows float16 to infinity, which the non-finite contract (correctly, as of item 1's sibling fix) turns into a zero patch, so the half run was asserting the opposite of the intended giant-*finite* behavior. Pairing a float32 LAF with the injected image dtype keeps the giant-finite case under test on every dtype and exercises the mixed-dtype promotion path. Your exact command now passes:

```
test-module tests/feature/test_laf.py::TestExtractPatchesPyr::test_giant_laf_uses_actual_coarsest_level \
  --device=cuda --dtype=float16 --isolate-half-precision   ->  1 passed
```

**3. Singleton image axes on the default path** — made the default path finite rather than narrowing the contract, since the extractors already treat a singleton *level* as zero-extent and it was inconsistent for a singleton *input* axis to crash one step earlier. `normalize_laf`, `denormalize_laf` and `generate_patch_grid_from_normalized_LAF` now use `float(max(size - 1, 1))`: a singleton axis counts as one pixel of extent. This is strictly a crash-to-defined change for `normalize_laf`/`generate_patch_grid_from_normalized_LAF` (`size == 1` always raised `ZeroDivisionError` before) and replaces `denormalize_laf`'s silent collapse-to-zero, so the pair round-trips. `test_one_pixel_axis_does_not_raise` now sweeps both extractors × `normalize_lafs_before_extraction` ∈ {False, True}. Docstrings and CHANGELOG record the new behavior.

### Verification (torch 2.13.0+cu130, RTX 4090; CPU repro of item 1 also confirmed on torch 2.9.1)

- Differential A/B against `092b9fdb` (old module exec'd side by side, `kornia.__file__` checked): finite-input output **bitwise identical** across CPU/CUDA × float32/float64/float16 × both extractors × both `normalize_lafs_before_extraction` values × 3 image shapes — 0 mismatches. The three changed divisions only differ when an axis is 1 pixel; the new masking is a no-op on finite input.
- `tests/feature/test_laf.py` CPU float32/float64: **190 passed, 18 skipped** (187 before; +3 new).
- CUDA float32: **128 passed, 13 skipped**.
- Extraction classes, CUDA float16/bfloat16 with `--isolate-half-precision`: **59 passed, 8 skipped**, plus the known dtype-independent `previous item was not torn down properly` harness error.
- Extraction/normalization classes, CPU float16/bfloat16: **78 passed, 6 skipped**.
- Inductor/fullgraph extraction tests: **7 passed**.
- `test_laf.py` + orientation + affine-shape + integrated, CPU float32/float64: **296 passed, 53 skipped**.
- `torch.jit.script(extract_patches_simple)` still scripts and matches eager; `extract_patches_from_pyramid` still stops at the known `filter2d` `_VALID_BORDERS` blocker.
- `pre-commit-all`: all hooks pass. `typecheck`: only the pre-existing `lightglue.py` deprecation diagnostic.

The pre-existing `TestTransformLAFs::test_transform_points` half-precision failures are unchanged and unrelated (`linalg.inv` rejects low-precision dtypes).

ducha-aiki (commented):
## Review of `0e86da88` — one correctness blocker remains

The latest commit fixes the three blockers from the previous review: both extractors sanitize non-finite LAFs before sampling, the giant finite-LAF test now covers CUDA half precision correctly, and the public singleton-axis path no longer raises. One uncovered correctness problem remains in the singleton handling.

### Must fix: the pyramid fallback collapses the non-singleton axis

At `kornia/feature/laf.py:619-623`, `_extract_patches_from_pyramid_levelwise` computes `min_l = min(h_l - 1, w_l - 1)`. If only one axis is singleton, `min_l` becomes zero and therefore **both** components of `k` become zero. This discards the entire affine linear part, including the valid non-singleton axis. It conflicts with the new `max(size - 1, 1)` convention in `normalize_laf`, `denormalize_laf`, and the simple extractor.

Minimal CPU reproduction on this head:

```python
img = torch.arange(5.0).reshape(1, 1, 1, 5)
laf = torch.tensor([[2.0, 0.0, 2.0], [0.0, 0.5, 0.0]]).reshape(1, 1, 2, 3)
simple = kornia.feature.extract_patches_simple(img, laf, 5)
pyramid = kornia.feature.extract_patches_from_pyramid(img, laf, 5)
```

Every row of `simple[0, 0, 0]` is `[0, 1, 2, 3, 4]`, while every element of `pyramid[0, 0, 0]` is `2`; the maximum difference is `2.0`. The transposed `5×1` case fails symmetrically. Please preserve the non-singleton level extent consistently with the new normalization convention, and make the singleton regression assert numerical structure or simple/pyramid parity rather than only shape and finiteness.

### Verification

- CPU `tests/feature/test_laf.py`, float32/float64: **190 passed, 18 skipped**.
- CUDA `tests/feature/test_laf.py`, float32: **128 passed, 13 skipped**.
- Targeted isolated CUDA float16/bfloat16 coverage for both non-finite backward paths, giant LAFs, singleton axes, and level-zero float-reference behavior: **10 passed**.
- `git diff --check`: clean; merging the current `main` is conflict-free. Pre-commit, docs, CodeQL, and documentation-link checks are green; the broader GitHub matrix was still running when reviewed.

Posted on behalf of @ducha-aiki by Codex (gpt-5.6-sol).

ducha-aiki:
Fixed the singleton-axis blocker in `1de112c3`.

The levelwise sampler now uses the same singleton extent convention as `normalize_laf` / `denormalize_laf`, so border sampling collapses only the singleton axis and preserves variation along the other axis. The new regression covers both `1×5` and `5×1`, for pixel and pre-normalized LAFs, and was verified red before the fix and green afterward.

Verification after the fix:
- CPU `tests/feature/test_laf.py`, float32/float64: **192 passed, 18 skipped**.
- Targeted CUDA float32/float16/bfloat16 singleton regression: **3 passed**.
- `pre-commit-all`: **passed**.

ducha-aiki (commented):
## Review of current head `1de112c3`

One blocking correctness gap remains.

**[P2] Stop the pyramid before a reflect-padded `pyrdown` from a 2-pixel axis** — `kornia/feature/laf.py:603-605`

The new singleton-level guard routes a pyramid ending in a 1-pixel level to `_extract_patches_from_pyramid_levelwise`, but that fallback still has to materialize the level. If any current level has a 2-pixel axis, `pyrdown(cur_img)` calls the 5×5 Gaussian filter with two pixels of reflect padding, which PyTorch rejects because the padding is not smaller than the input dimension.

Fresh reproduction at this exact head with torch 2.9.1:

```python
for n in (2, 4, 8, 12):
    img = torch.rand(1, 1, n, n)
    laf = torch.tensor([[1., 0., n / 2], [0., 1., n / 2]]).view(1, 1, 2, 3)
    extract_patches_from_pyramid(img, laf, 1)
```

Results:

```
2 -> RuntimeError (reflect padding on 2×2)
4 -> RuntimeError (reflect padding on 2×2)
8 -> RuntimeError (reflect padding on 2×2)
12 -> succeeds
```

The new `12→6→3→1` regression test passes specifically because it never visits size 2. The same issue affects rectangular inputs as soon as either axis reaches 2. This failure existed on `main`, but the PR now explicitly claims and tests singleton pyramid-level handling; the current guard therefore does not fully implement that contract.

A consistent fix would be to stop constructing the pyramid when the *source* level has an axis `<= 2`, treating that 2-pixel level as the actual coarsest usable one, rather than appending a target level that `pyrdown` cannot produce with its reflect-padding kernel. Please add power-of-two and rectangular regression coverage (for example 8×8 and 2×N with `PS=1`).

Other verification:

- Focused CPU extraction tests: **65 passed, 6 skipped**.
- Randomized atlas-vs-levelwise output comparison over ordinary multi-level sizes: no mismatches above standard tolerance.
- Base/head outcome sweep over 4,608 `(H, W, PS)` combinations: no success→failure regressions; 780 formerly failing combinations now succeed, while the remaining failures cluster around the 2-pixel `pyrdown` case above.
- Required GitHub checks were green when reviewed.

GitHub does not permit this account (the PR author) to submit a formal “Request changes” review, so I am posting the blocker as a comment review.

ducha-aiki:
Fixed the P2 from [review 5076377628](https://github.com/kornia/kornia/pull/4128#pullrequestreview-5076377628) in `77ec915b`.

The pyramid-size loop now stops before `pyrdown` would consume an axis of size `<= 2`. That keeps the 2-pixel source as the actual coarsest usable level instead of recording a target that the 5×5 reflect-padded Gaussian cannot materialize.

The existing `PS=1` regression now covers:

- `12×12` (safe `12→6→3→1` path)
- `8×8` (power-of-two path reaching `2×2`)
- `2×8` and `8×2` (both rectangular orientations)

TDD verification: the expanded test was red before the implementation (3 failed, 1 passed) and green afterward.

Fresh verification on the pushed commit:

- Full CPU `tests/feature/test_laf.py`, float32/float64: **198 passed, 18 skipped**.
- Targeted CPU float16/bfloat16: **8 passed**.
- Targeted CUDA float32: **4 passed**.
- Targeted CUDA float16/bfloat16 with subprocess isolation: **8 passed**.
- CPU inductor/fullgraph pyramid tests: **3 passed**.
- Exhaustive torch 2.9.1 sweep over 4,608 `(H, W, PS)` combinations: **0 exceptions and 0 non-finite outputs**.
- `pixi run pre-commit-all`: **passed**.
- `pixi run typecheck`: exit 0 with only the existing deprecation diagnostic in unchanged `lightglue.py`.
- Independent code review: no Critical, Important, or Minor findings.

The unrelated benchmark JSON files remain untouched.

ducha-aiki:
## Review

**Verdict: LGTM.** I read the full diff, re-derived the atlas coordinate math, and re-verified the PR's equivalence, correctness, and performance claims on a CUDA machine. No correctness issues found.

### What I checked in the code

- **Atlas remap math**: I re-derived the level→atlas mapping. For a level pixel center `i`, `x_level·(w_l/atlas_w) + offset` with `offset = (w_l + 2(xoff+1))/atlas_w − 1` lands exactly on the atlas pixel center `xoff+1+i`; the y-offset `(h_l+2)/atlas_h − 1` is the same formula with `yoff = 0`. Exact, not approximate.
- **Border semantics**: clamping to the level's outermost pixel centers plus the 1-pixel replicated guard reproduces `padding_mode="border"` for both value (bilinear partners are both the border pixel) and gradient (zero outward), so the per-level clamp that the atlas needs anyway is behavior-preserving. The guard test pins the leak at both the coordinate-rounding scale (size 2495) and the gradient level (size 32).
- **Level-list construction**: descent stops both below `PS` and at a ≤2-pixel source (which `pyrdown`'s reflect pad cannot handle), and the atlas guard routes any built 1-pixel level to the levelwise fallback, so the `size − 1` divisions in the atlas constants can never see zero. The levelwise path carries its own `max(size−1, 1)` handling, consistent with the new singleton-axis convention in `normalize_laf`/`denormalize_laf`.
- **Non-finite LAF handling**: frames are sanitized *before* any grid arithmetic on both paths, invalid frames carry level −1 (clamped for indexing, zeroed after the loop), and both the output zeroing and the `masked_fill` on `nlaf` give zero gradients through masked positions — no data-dependent Python branch, so `fullgraph=True` holds.
- **Chunk budget**: `max(2, ch)` correctly bounds both the 2-coordinate folded grid and the ch-channel sampled chunk.
- **Callers**: all in-tree call sites (`orientation.py`, `integrated.py`, `affine_shape.py`) pass matched batch sizes, so the new batch-mismatch check cannot fire on internal paths. No stale `supports_grid_sample` references remain anywhere.

### What I ran (RTX GPU, torch from the repo venv)

- `tests/feature/test_laf.py` CPU fp32/fp64: **198 passed, 18 skipped**. CUDA fp32: **132 passed, 13 skipped**.
- CUDA fp16 with `--isolate-half-precision`, extractor + LAF-conversion selection: **49 passed, 0 failed**.
- Numerical A/B against `origin/main` (worktree, `kornia.__file__` verified in both trees), fp64, B=2, N=20, 512², mixed scales, CPU **and** CUDA: `extract_patches_simple` is **bitwise identical** to `main`; `extract_patches_from_pyramid` max abs diff **1.1e-13** — matching the stated remap-arithmetic-only change.
- Eager CPU timing at B=8, N=4000, PS=32, 512² fp32: **644 ms on `main` → 255 ms on this branch** (~2.5× here; my LAF scale distribution differs from the benchmark's, direction and magnitude consistent with the PR table).

### Pre-existing issues surfaced while testing (not this PR's)

- `TestTransformLAFs::test_transform_points[cuda-float16-*]` fails (9 cases) under `--isolate-half-precision` — **reproduced identically on `origin/main`**, in code this PR does not touch. Probably worth its own tolerance-tuning issue.
- The isolation plugin reports "previous item was not torn down properly" errors on device-only (non-dtype-parametrized) tests in the same run — also reproduced on `main`; an infra quirk of mixing isolated and non-isolated items, not a test defect.

### Unverified

MPS behavior (no MPS hardware here — the clamp-emulation path is exercised only structurally), TPU, and the peak-RSS numbers in the table (I did not re-measure memory; the chunk-budget logic itself is tested).

PR #4129, opened 2026-08-31
Fix transform_bbox mutating xywh input

### LINKED ISSUE #4011: W1b: `transform_bbox(mode='xywh')` overwrites the caller's box tensor in place
`kornia.geometry.bbox.transform_bbox` converts `xywh` input to `xyxy` by writing into the input tensor (`boxes[..., 2] = boxes[..., 0] + boxes[..., 2]`, `bbox.py:512-513`), so after the call the caller's tensor holds `xyxy` values while still being labelled `xywh` in their code. `xyxy` mode does not mutate.

## Repro

```python
import torch
from kornia.geometry.bbox import transform_bbox

bx = torch.tensor([[1., 1., 2., 1.]])          # xywh: x=1, y=1, w=2, h=1
before = bx.clone()
out = transform_bbox(torch.eye(3)[None], bx, mode="xywh")
print("input before:", before.tolist(), " after:", bx.tolist(), " output:", out.tolist())
```

## Observed

```
input before: [[1.0, 1.0, 2.0, 1.0]]  after: [[1.0, 1.0, 3.0, 2.0]]  output: [[1.0, 1.0, 2.0, 1.0]]
```

The return value is correct; the argument was silently rewritten to `xyxy`. Calling the function twice on the same tensor compounds the corruption.

## Expected

The input is untouched. Doing the `xywh -> xyxy` conversion on a copy (or building the `xyxy` tensor functionally with `torch.stack`) gives a return value byte-identical to today's, so this is a focused fix that is welcome as a PR — no behavior change on the output, only the removal of a side effect. A regression test that asserts `torch.equal(bx, before)` after the call fails before the fix.

## Classification

**Broken output (side effect on an input).** Idiom 6 — *silently mutates an input argument* — added to the idiom vocabulary by this audit.

## Disposition

n/a — single symbol.

## Environment

kornia 0.9.0rc1 @ `0685b1f5` (main), torch 2.9.1, python 3.11, cpu. Found during the bbox/boxes conventions audit (batch 4, tracking issue #4020); the batch documents the behavior below as it is and pins it, and does not change it.

Alphaxiaoteng:

`transform_bbox(..., mode="xywh")` converted boxes to `xyxy` with in-place assignments, silently rewriting the caller's input tensor even though the returned boxes were correct. This change clones the input only in the `xywh` conversion branch before those internal assignments, preserving the existing output while removing the side effect.

The existing xywh numerical test now also keeps an exact copy of the argument and verifies that it remains unchanged after the call.

Fixes #4011.

## What did you check?

With the regression assertion kept and the production clone temporarily removed, the focused CPU test failed for both float32 and float64 at the new input-integrity assertion; the existing output assertion passed before it.

```text
$ pixi run test-module tests/geometry/test_bbox.py::TestTransformBoxes2D::test_transform_boxes_wh --device=cpu --dtype=float32,float64
2 passed

$ KORNIA_TEST_OPTIMIZER=inductor pixi run test-module tests/geometry/test_bbox.py --device=cpu --dtype=float32,float64
32 passed

$ pixi run pre-commit-all
all hooks passed

$ pixi run typecheck
exit 0; one existing deprecation diagnostic in kornia/feature/lightglue.py
```

The same full bbox test file without compile collection also passed 26 tests, and `git diff --check` passed.

## How did AI help?

OpenAI Codex, including Terra sub-agents, helped trace the call path, identify the smallest non-mutating change, and draft the regression assertion. I independently reproduced the old side effect, reviewed the final three-line diff, ran the old-code failure and fixed-code tests, and completed the repository's pre-commit and type checks.

### DIFF
--- a/kornia/geometry/bbox.py
+++ b/kornia/geometry/bbox.py
@@ -512,6 +512,7 @@ def transform_bbox(
 
     # convert boxes to format xyxy
     if mode == "xywh":
+        boxes = boxes.clone()
         boxes[..., 2] = boxes[..., 0] + boxes[..., 2]  # x + w
         boxes[..., 3] = boxes[..., 1] + boxes[..., 3]  # y + h
 
--- a/tests/geometry/test_bbox.py
+++ b/tests/geometry/test_bbox.py
@@ -164,6 +164,7 @@ class TestTransformBoxes2D(BaseTester):
             device=device,
             dtype=dtype,
         )
+        boxes_before = boxes.clone()
 
         expected = torch.tensor(
             [
@@ -180,6 +181,7 @@ class TestTransformBoxes2D(BaseTester):
 
         out = transform_bbox(trans_mat, boxes, mode="xywh", restore_coordinates=True)
         self.assert_close(out, expected, atol=1e-4, rtol=1e-4)
+        assert torch.equal(boxes, boxes_before)
 
     def test_gradcheck(self, device):
         boxes = torch.tensor(


ducha-aiki:
**The fix is verified correct — and it is a byte-for-byte duplicate of #4118, which arrived ~15 h earlier.** Flagging that first so neither of you spends more time on it.

Setup: worktree at this head `64033d3c`, with `kornia/` swapped to the merge-base `cec6d3db` for the fail-on-base pass; `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS.

**Identical production change.** Both PRs add one `boxes = boxes.clone()` at the top of the `mode == "xywh"` branch, and git agrees literally — both diffs read `index e83ddae288..0341193258`, i.e. the same post-image blob.

**The pin discriminates**, at both dtypes:

```
# kornia/ at the merge-base, this PR's test kept
FAILED TestTransformBoxes2D::test_transform_boxes_wh[cpu-float32] - assert False
FAILED TestTransformBoxes2D::test_transform_boxes_wh[cpu-float64] - assert False
2 failed, 5 passed
```

**No output changes and no path is missed.** 12 configurations (both modes x `restore_coordinates` in `{None, True, False}` x `(5,4)` and `(3,5,4)`):

```
mode=xyxy/xywh, rc=None/True/False, shapes (5,4) and (3,5,4):
  input_mutated=False in all 12; shapes and dtype unchanged
grad ok: True   grad finite: True
```

The `xyxy` path never needed the clone (`transform_points` allocates), and the trailing `transformed_boxes[..., 2] -= ...` writes into either `restored_boxes` (a fresh clone) or `transform_points`' output, so one clone in the one branch is the whole fix. Gradients still flow through it — `test_gradcheck` passes.

**Where this PR is better than #4118:** the test. You extend `TestTransformBoxes2D::test_transform_boxes_wh`, so the assertion inherits the `device`/`dtype` fixtures and the `BaseTester` machinery and runs on every backend. #4118 adds a new module-level `tests/geometry/test_bbox_input_mutation.py` with hardcoded CPU float32 and no fixtures, which AGENTS.md and TESTING.md steer away from, and which puts bbox coverage in a second file.

**Disposition (maintainer's call, on the record):** I intend to land this one and close #4118 with credit. @Marchematics got there first with a correct, identical fix and that deserves saying plainly.

Nothing to change here.

ducha-aiki:
Verified this locally against the PR head `64033d3c` and its parent `cec6d3db` (torch 2.9.1, python 3.11, CPU, float32/float64). **It works, and the fix is complete.** Details, since "clone then keep assigning in place" is the kind of thing that deserves a check rather than a nod:

**1. The two in-place lines are order-safe.** `boxes[..., 2] = boxes[..., 0] + boxes[..., 2]` writes index 2 while reading 0 and 2; the next line reads 1 and 3. The overwritten column is never read again, so the sequential in-place form is exactly `torch.stack([x, y, x + w, y + h], -1)` — checked with `torch.equal`, True.

**2. The `xywh` branch was the only site that could reach caller memory.** The second in-place block at the end writes into `transformed_boxes`, which is always freshly allocated by `transform_points(...)`. Confirmed by `data_ptr()`: the return value never aliases the input, across all six `mode` x `restore_coordinates` combinations. So one `clone()` in that branch is sufficient; no second copy is needed.

**3. Output is byte-identical to base.** Same fixed inputs (float64, an affine with a negative off-diagonal term so the flip-restore path is exercised), all six `mode` x `restore_coordinates` combinations: `torch.equal` True on every one.

**4. The pin discriminates.** Reverting only `boxes = boxes.clone()` on this branch and rerunning `TestTransformBoxes2D::test_transform_boxes_wh`: fails for both `cpu-float32` and `cpu-float64` at the new `assert torch.equal(boxes, boxes_before)` (line 184), with the output assertion above it passing first. Exactly as the PR description claims. Full file with `KORNIA_TEST_OPTIMIZER=inductor`: 32 passed.

**5. Bonus, worth knowing.** On base, `transform_bbox(M, boxes, mode="xywh")` with a leaf `boxes` that requires grad raised

```
RuntimeError: a view of a leaf Variable that requires grad is being used in an in-place operation.
```

i.e. `xywh` mode was unusable in a differentiable pipeline at all. With the clone it runs, and `torch.autograd.gradcheck` on `lambda t: transform_bbox(M, t, mode="xywh")` passes in float64. That is a second real bug closed by the same line; `test_gradcheck` only covers `"xyxy"`, so it is currently untested. Optional (not blocking): extend `test_gradcheck` with an `"xywh"` case to pin it.

**6. Nothing stale left behind.** `grep -rnE "#4011|issues/4011" kornia/ tests/ docs/` is empty, so no prose or `xfail` documented the old behavior. No in-repo caller of `transform_bbox` relied on the mutation — the augmentation path goes through `Boxes.transform_boxes` / `_transform_boxes` in `geometry/boxes.py`, which is untouched.

One pre-existing wart, unchanged by this PR and out of scope: `mode="xywh"` with the polygon `(B, N, 4, 2)` layout still raises `IndexError` on `boxes[..., 2]`.

LGTM.

ducha-aiki (approved):

PR #962, opened 2021-04-12
Add Dense SIFT

ducha-aiki:
### Description
A few sentences describing the changes proposed in this pull request.
Adds DenseSIFT (baseline: vlfeat https://www.vlfeat.org/overview/dsift.html) 
Addresses https://github.com/kornia/kornia/issues/462 and https://github.com/kornia/kornia/issues/459

Things, which are now missing:

- [ ] proper padding to allow having output resolution == input resolution
- [ ] tests
- [ ] benchmark versus some baseline, e.g. vlfeat
- [ ] documentation of the design choices.

image

### Status
**Work in progress**

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or new feature that would cause existing functionality to change)
- [ ] New tests added to cover the changes
- [ ] Docstrings/Documentation updated

### PR Implementer
This is a small checklist for the implementation details of this PR.

If there are any questions regarding code style or other conventions check out our 
[summary](https://github.com/kornia/kornia/blob/master/CONTRIBUTING.rst).

- [ ] Did you discuss the functionality or any breaking changes before ?
- [ ] **Pass all tests**: did you test in local ? `make test`
- [ ] Unittests: did you add tests for your new functionality ?
- [ ] Documentations: did you build documentation ? `make build-docs`
- [ ] Implementation: is your code well commented and follow conventions ? `make lint`
- [ ] Docstrings & Typing: has your code documentation and typing ? `make mypy`
- [ ] Update notebooks & documentation if necessary

### KorniaTeam
<details>
  <summary>KorniaTeam workflow</summary>
  
  - [ ] Assign correct label
  - [ ] Assign PR to a reviewer
  - [ ] Does this PR close an Issue? (add `closes #IssueNumber` at the bottom if 
        not already in description)

</details>

### Reviewer
<details>
  <summary>Reviewer workflow</summary>

  - [ ] Do all tests pass? (Unittests, Typing, Linting, Documentation, Environment)
  - [ ] Does the implementation follow `kornia` design conventions?
  - [ ] Is the documentation complete enough ?
  - [ ] Are the tests covering simple and corner cases ?
 
</details>

PR #183, opened 2019-07-16
[ready to review] SIFT descriptor as in https://github.com/ducha-aiki/pytorch-sift

ducha-aiki:
-  [x] implementation
-  [x] docs
-  [x] tests

PR #4130, opened 2026-08-31
fix(feature): return isolated SIFT kernel copies

### LINKED ISSUE #4083: DenseSIFTDescriptor.get_pooling_kernel() hands out an alias of its buffer, captured before .to()
`DenseSIFTDescriptor.get_pooling_kernel()` returns a tensor captured at construction time. It is an alias of the module's buffer rather than a copy, and it is captured *before* any `.to()`, so it both lets a caller corrupt the module and goes stale on device/dtype migration.

`kornia/feature/siftdesc.py`:

```python
# Cache pooling kernel torch.Tensor for fast return in get_pooling_kernel
self._pooling_kernel = self._bin_pooling_kernel_weight.detach()

def get_pooling_kernel(self) -> torch.Tensor:
    # Return the cached detached pooling kernel directly for optimal speed
    return self._pooling_kernel
```

`.detach()` shares storage, so `self._pooling_kernel` is a permanent view of the `_bin_pooling_kernel_weight` buffer as it existed in `__init__`.

## Two symptoms

**1. A caller's in-place write corrupts the module.**

```python
import torch
from kornia.feature import DenseSIFTDescriptor

m = DenseSIFTDescriptor()
m.get_pooling_kernel().zero_()
print(bool(m._bin_pooling_kernel_weight.abs().sum() == 0))
```
```
True
```

The accessor is documented as returning a "detached tensor", which reads as safe to use, and this is the same shape of defect as #4068 — just instance-local rather than process-global, so milder.

**2. The alias goes stale on `.to()`.** `nn.Module._apply` replaces the buffer with a new tensor; the Python attribute still points at the old one.

```python
import torch
from kornia.feature import DenseSIFTDescriptor

m = DenseSIFTDescriptor().to(torch.float64)
print(m._bin_pooling_kernel_weight.dtype, m.get_pooling_kernel().dtype)
```
```
torch.float64 torch.float32
```

The same happens for `.to("cuda")` or `.to("mps")`: the accessor keeps returning the CPU tensor for the lifetime of the module. This is the more likely one to bite, since `.to(device)` is routine and nothing raises.

## Proposed fix

Drop `self._pooling_kernel` and return `self._bin_pooling_kernel_weight.detach().clone()` (or at minimum re-read the buffer on each call). The cache saves nothing measurable — the accessor is not on any hot path, and `forward` does not use it.

`SIFTDescriptor.get_pooling_kernel` / `get_weighting_kernel` return `self.pk.weight.detach()` / `self.gk.detach()` fresh on each call, so they track `.to()` correctly; they still hand out an alias, so symptom 1 applies there too and is worth fixing in the same pass.

Reproduced on `main` at 6fa75d64 with torch 2.9.1, CPU.

Raised out of the review of #4074, which fixes #4068 and deliberately leaves this alone as a separate defect.

Alphaxiaoteng:

`DenseSIFTDescriptor.get_pooling_kernel()` cached a detached alias during construction. The returned tensor could mutate internal state and remained on the old dtype/device after module migration. The accessor now returns a detached clone of the convolution weight actually used by `forward`. The two equivalent `SIFTDescriptor` kernel accessors also return detached clones so callers cannot mutate the module through them.

Regression tests cover mutation isolation, detached autograd semantics, explicit dtype/device migration, forward-triggered migration, and MPS device migration.

Fixes #4083.

## What did you check?

- Unmodified baseline with the retained regression tests: 2 failed.
- Fixed focused regression tests on CPU: 2 passed.
- Fixed focused regression tests on MPS float32: 2 passed.
- Complete `tests/feature/test_siftdesc.py` on CPU float32/float64 with inductor enabled: 57 passed, 2 existing JIT skips.
- `pixi run pre-commit-all`: all hooks passed.
- `pixi run typecheck`: exit 0; one pre-existing deprecation diagnostic in unmodified `kornia/feature/lightglue.py`.
- `git diff --check`: passed.

## How did AI help?

OpenAI Codex coordinated the reproduction, implementation, test execution, and final review. Terra subagents inspected the call chain and performed an adversarial review. I independently reviewed the final diff, reproduced the failure with the production fix removed, reran the tests and quality gates, and addressed review findings about the Dense source of truth, detached semantics, and MPS coverage.

### DIFF
--- a/kornia/feature/siftdesc.py
+++ b/kornia/feature/siftdesc.py
@@ -184,17 +184,17 @@ class SIFTDescriptor(nn.Module):
         """Return the spatial pooling kernel used for histogram accumulation.
 
         Returns:
-            Detached convolution kernel tensor from the pooling layer.
+            Detached copy of the convolution kernel tensor from the pooling layer.
         """
-        return self.pk.weight.detach()
+        return self.pk.weight.detach().clone()
 
     def get_weighting_kernel(self) -> torch.Tensor:
         """Return the Gaussian weighting kernel used before orientation pooling.
 
         Returns:
-            Detached Gaussian kernel tensor.
+            Detached copy of the Gaussian kernel tensor.
         """
-        return self.gk.detach()
+        return self.gk.detach().clone()
 
     def forward(self, input: torch.Tensor) -> torch.Tensor:
         r"""Compute SIFT descriptors for square grayscale patches.
@@ -360,17 +360,13 @@ class DenseSIFTDescriptor(nn.Module):
         PoolingConv.weight.data.copy_(self._poolingconv_weight)
         self.PoolingConv = PoolingConv
 
-        # Cache pooling kernel torch.Tensor for fast return in get_pooling_kernel
-        self._pooling_kernel = self._bin_pooling_kernel_weight.detach()
-
     def get_pooling_kernel(self) -> torch.Tensor:
-        """Return the cached pooling kernel for dense SIFT binning.
+        """Return the pooling kernel for dense SIFT binning.
 
         Returns:
-            Detached tensor containing pooling weights.
+            Detached copy of the pooling weights.
         """
-        # Return the cached detached pooling kernel directly for optimal speed
-        return self._pooling_kernel
+        return self.bin_pooling_kernel.weight.detach().clone()
 
     def forward(self, input: torch.Tensor) -> torch.Tensor:
         """Compute dense SIFT descriptors over a full image grid.
--- a/tests/feature/test_siftdesc.py
+++ b/tests/feature/test_siftdesc.py
@@ -108,6 +108,22 @@ class TestSIFTDescriptorKernelBuffer(BaseTester):
         mod(torch.rand(2, 1, 32, 32, dtype=torch.float64))
         assert (mod.gk.dtype, mod.gk.device) == before
 
+    def test_kernel_accessors_return_copies(self, device):
+        """Accessor writes must not mutate SIFT's pooling parameter or weighting buffer (#4083)."""
+        mod = SIFTDescriptor(32).to(device)
+        pooling = mod.get_pooling_kernel()
+        weighting = mod.get_weighting_kernel()
+        expected_pooling = pooling.clone()
+        expected_weighting = weighting.clone()
+        assert not pooling.requires_grad
+        assert not weighting.requires_grad
+
+        pooling.zero_()
+        weighting.zero_()
+
+        self.assert_close(mod.pk.weight, expected_pooling)
+        self.assert_close(mod.gk, expected_weighting)
+
 
 class TestSIFTConstantPatchIsFinite(BaseTester):
     """A constant patch has a zero-norm descriptor; normalising it must not give NaN.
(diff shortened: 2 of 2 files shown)


ducha-aiki:
**APPROVE**, with one framing note so the record is straight about what this PR fixes.

Setup: worktree at the PR head `c5d555e3`, with `kornia/` swapped to the merge-base `cec6d3db` for the fail-on-base pass; `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS.

**Both pins discriminate:**

```
# kornia/ at the merge-base, tests from the PR
FAILED TestSIFTDescriptorKernelBuffer::test_kernel_accessors_return_copies[cpu]
FAILED TestDenseSIFTDescriptor::test_pooling_kernel_accessor_is_isolated_and_tracks_to[cpu]
       assert torch.float32 == torch.float64
2 failed, 49 deselected

# at the head
50 passed, 1 skipped in 0.44s
```

**The two halves of this PR are not the same kind of change, and only one of them is the bug.**

*The `DenseSIFTDescriptor` half is the real fix.* `self._pooling_kernel` was captured in `__init__`, so it was pinned to the pre-`.to()` tensor forever:

```
main:  buffer dtype torch.float64   accessor dtype torch.float32
head:  buffer dtype torch.float64   accessor dtype torch.float64
```

Dropping the cache and re-reading `self.bin_pooling_kernel.weight` is what fixes that, and it is worth landing on its own. The cache also bought nothing — the accessor is off every hot path and `forward` never calls it.

*The `.clone()` on the two `SIFTDescriptor` accessors is a contract choice, not a bug fix.* Those two already re-read on each call and already track `.to()`; the only thing the clone changes is that a caller can no longer corrupt the module through the getter. Since I asked myself whether it is warranted, here is the case both ways:

- Against: PyTorch never defensively copies — `m.weight`, `.parameters()`, `.buffers()`, `.state_dict()` all hand out live references, and kornia has no precedent for it either (the whole library has three `.detach().clone()` sites and none of them is an accessor). `m.pk.weight` and `m.gk` are still reachable and still aliased, so the clone guards one door in a wall with no others.
- For: these are `get_*()` methods, not attributes, and a getter that leaks an internal alias is the classic encapsulation leak. The cost is nil — the kernels are `(1, 1, 16, 16)` and `(1, 41, 41)`, and the clone measures **2161 ns vs 1311 ns** for `pk` and **1784 ns vs 854 ns** for `gk`, on a call that happens once per user, never per batch.
- Decisive for me: this PR *updates the documented contract in the same commit* ("Detached copy of ..."), and after the `DenseSIFTDescriptor` fix all three accessors behave identically only if all three clone or none do. Documented and consistent beats undocumented and inconsistent, so the clone stays.

Two things worth recording rather than acting on:

1. Neither accessor is rendered anywhere. `docs/source/feature.rst` has bare `.. autoclass:: SIFTDescriptor` / `DenseSIFTDescriptor` with no `:members:`, and `conf.py` sets no `autodoc_default_options`. So these are public by name only, with zero in-tree callers, unchanged since #183 (2019) and #962. That is *why* the alias survived this long, and it also bounds the blast radius of changing the contract to essentially nil.
2. The symmetric leak on `SIFTDescriptor` is real but was the milder half of #4083, exactly as the issue said.

ducha-aiki (approved):

PR #4131, opened 2026-08-31
fix(feature): avoid half-precision underflow in patch affine shape estimation

### LINKED ISSUE #4123: PatchAffineShapeEstimator returns NaN for every float16 patch (moments underflow + eps=1e-10 flushes to zero)
## Describe the bug

`PatchAffineShapeEstimator` returns NaN for every patch in `float16`, through two stacked failures in `forward` (`kornia/feature/affine_shape.py`):

1. The weighted-gradient second moments underflow: the Gaussian `weighting` peaks at ~1.1e-3, so weighted gradients are ~5e-4 and their squares ~3e-7; averaged over the patch, all three moments flush to exactly `0.0` in fp16.
2. The degenerate-patch guard that should catch this — `ellipse_shape < self.eps` with `eps = 1e-10` — never fires, because **1e-10 itself flushes to 0.0 when compared in fp16** (`0.0 < 0.0` is false). The circular-shape replacement is skipped and the final `ellipse_shape / ellipse_shape.max(...)` divides 0/0 → NaN.

The NaN then propagates through `ellipse_to_laf` into `LAFAffineShapeEstimator`'s output, so every fp16 LAF comes back NaN.

Until #4122, this was masked on CPU by `torch.inverse` raising `linalg.inv: Low precision dtypes not supported` first — the pipeline failed either way, but with an error instead of silent NaN. With the closed-form `ellipse_to_laf`, the NaN is what reaches the caller.

## Reproduction

```python
import torch
from kornia.feature.affine_shape import PatchAffineShapeEstimator

patch = torch.zeros(1, 1, 32, 32, dtype=torch.float16)
patch[:, :, 15:-15, 9:-9] = 1  # a perfectly good, high-contrast patch
print(PatchAffineShapeEstimator(32).to(torch.float16)(patch))
# tensor([[[nan, nan, nan]]], dtype=torch.float16)
```

`bfloat16` does not NaN (its fp32-like exponent range keeps both the moments and the eps alive) but still fails the `test_toy` tolerance override (`atol=1e-4`, tighter than `testing/base.py`'s bfloat16 default of 7.8e-3 — the #4081 override pattern).

## Expected behavior

Finite ellipse shapes in fp16. The moments carry the actual shape signal, so the fix is fp32-internal math (compute the second moments and the eps comparison in `float32`, cast the normalized result back) rather than just an eps bump — an eps bump alone would classify *every* fp16 patch as degenerate and return circles.

Environment: kornia at `2009933e`, torch 2.9.1, CPU (device-independent arithmetic).

Hexecu:

I reproduced the rectangle from #4123 on CPU and MPS. On CPU, `PatchAffineShapeEstimator` returns NaNs in float16. On MPS, the same patch produces a finite circular shape instead of the expected anisotropic one. Checking only for finite output would miss that second case.

This promotes float16 and bfloat16 patches to float32 before gradient evaluation, keeps the moments and normalization in float32, then returns the result in the input dtype. The existing `eps` is unchanged. Increasing it alone would turn valid patches into circles without recovering their shape.

The float32 and float64 computation stays unchanged, as does the registered weighting buffer. The change includes a short docstring note and changelog entry.

Fixes #4123, reported by @ducha-aiki.

This addresses the patch estimator root cause discussed in #4122. Its latest head, `d2e53c8`, adds a downstream finite-output fallback, but the patch estimator itself still underflows. That fallback also cannot recover a finite but incorrect shape. This PR does not change `ellipse_to_laf` or include the inverse implementation.

## What did you check?

The main comparison used Python 3.11.7 and PyTorch 2.9.1 on Apple Silicon, against `cec6d3dbae7f2f99f83694db24cc8c137866eb57`. Base and fix ran from separate worktrees with the imported source path checked.

The issue reproduction is enough to see the change. Set `device` to `"mps"` to check the other backend.

```python
import torch
from kornia.feature import PatchAffineShapeEstimator

device = "cpu"
patch = torch.zeros(1, 1, 32, 32, device=device, dtype=torch.float16)
patch[:, :, 15:17, 9:23] = 1
print(PatchAffineShapeEstimator(32).to(device, patch.dtype)(patch))
```

```text
CPU before: [nan, nan, nan]
MPS before: [1.0, 0.0, 1.0]
Both after: [0.08831787109375, 0.0, 1.0]
Float32 reference: [0.0883344, 0.0, 1.0]
```

The new tests cover flat patches, both rectangle orientations, patch sizes 19 and 32, batches of 1 and 3, output dtype/device and finite input gradients. They use the existing dtype tolerances. There is also an explicit CPU float16/bfloat16 case, so the regression is exercised by ordinary CI jobs even when their dtype selection is float32 or float64.

```bash
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python -m pytest \
  tests/feature/test_affine_shape_estimator.py \
  -k 'test_precision or test_half_precision_cpu' \
  --device=cpu --dtype=float16,bfloat16,float32,float64 -q -o addopts=''
```

Those cases went from 5 failed and 13 passed to 18 passed on CPU. On MPS, using `--dtype=float16,bfloat16,float32`, they went from 4 failed and 8 passed to 12 passed, with the 2 CPU-only cases skipped.

The broader CPU run covered affine shape estimation, LAFs, scale-space detection, orientation and SIFT in float32/float64. It went from 350 passed to 360 passed, with the same 16 skips. On PyTorch 2.5.1 it went from 335 to 345 passed, with the same 31 skips. The pretrained AffNet and OriNet classes were excluded from these runs. The patch estimator and buffer tests also passed across all four CPU dtypes on both versions. Inductor passed all 4 CPU and 3 MPS dtype cases. Six autocast probes had finite outputs and gradients after the fix. A separate comparison found bit-identical float32/float64 outputs and input gradients in 16 cases.

I also checked the fix on top of #4122, then repeated the integration checks when its head advanced to `d2e53c8`. All 12 configurations of flat and anisotropic patches match the float32 reference within dtype tolerances after the fix, compared with 8 before. The 36 rotated LAF cases cover 0, 37 and 90 degrees with orientation preservation enabled and disabled. The maximum float16 matrix-relative error falls from `1.26` to `8.93e-4`.

The new fallback regression in that PR still passes. Its ramp also exposed NaN gradients before this fix in 8 of 24 CPU configurations covering three dtypes, single and mixed batches, two rotations and both orientation settings. Afterward all 24 have finite image and LAF gradients. This only checks the default patch estimator, not arbitrary custom estimators. MPS pipeline backward could not be checked because PyTorch lacks `grid_sampler_2d_backward` there.

There are boundaries to that result. On current main, the full half precision LAF path still hits the inverse limitation addressed by #4122. On the combined tree, the existing LAF toy tests still fail their fixed `1e-4` comparisons. The affine test file gives 28 passed, 8 failed and 4 skipped both before and after, excluding AffNet. Those assertions are unchanged here. The broader MPS run also retains the same two SIFT failures that request unsupported float64 tensors. I compared the failing test IDs against the respective bases.

All repository pre-commit hooks passed. Feature doctests gave 28 passed and 9 skipped. `ty 0.0.9` reports the same pre-existing deprecation in `lightglue.py` on both trees. TorchScript still fails in the unchanged Sobel kernel signature on both trees. CUDA was not available. Pixi was not installed locally, so these checks used isolated environments and the underlying pytest, pre-commit and ty commands.

## How did AI help?

AI helped with the analysis and acted as a coding copilot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* Compute `PatchAffineShapeEstimator` moments and normalization in float32 for float16 and bfloat16
+  inputs, then cast the result back to the input dtype (#4123). Weighted gradients could underflow
+  when squared in float16, producing NaNs or replacing an anisotropic patch with a circular shape.
+  Float32 and float64 inputs keep their existing computation.
+
 * Make `load_pointcloud_ply` and `load_pointcloud_ply_binary` parse the PLY header instead of skipping a fixed
   number of lines. Both loaders skipped `header_size=8` lines and read everything after them as `x y z` triples, so a
   minimal PLY file (seven header lines, no `comment`) lost its first point to a header line -- the binary reader
--- a/kornia/feature/affine_shape.py
+++ b/kornia/feature/affine_shape.py
@@ -48,6 +48,9 @@ class PatchAffineShapeEstimator(nn.Module):
 
     The method determines the affine shape of the local feature as in :cite:`baumberg2000`.
 
+    For float16 and bfloat16 inputs, the moments and normalization are computed in float32.
+    The output keeps the input dtype.
+
     Args:
         patch_size: the input image patch size.
         eps: for safe division.
@@ -81,6 +84,10 @@ class PatchAffineShapeEstimator(nn.Module):
 
         """
         KORNIA_CHECK_SHAPE(patch, ["B", "1", "H", "W"])
+        dtype = patch.dtype
+        # Squared weighted gradients and the degeneracy threshold can underflow in half precision.
+        if dtype in (torch.float16, torch.bfloat16):
+            patch = patch.float()
         # cast into a local; rebinding `self.weighting` would make the module's dtype/device depend
         # on whichever tensor was passed last.
         weighting = self.weighting.to(patch.dtype).to(patch.device)
@@ -105,7 +112,7 @@ class PatchAffineShapeEstimator(nn.Module):
         ellipse_shape = ellipse_shape * (1.0 - bad_mask) + circular_shape * bad_mask
         # normalization
         ellipse_shape = ellipse_shape / ellipse_shape.max(dim=2, keepdim=True)[0]
-        return ellipse_shape
+        return ellipse_shape.to(dtype)
 
 
 class LAFAffineShapeEstimator(nn.Module):
--- a/tests/feature/test_affine_shape_estimator.py
+++ b/tests/feature/test_affine_shape_estimator.py
@@ -54,6 +54,56 @@ class TestPatchAffineShapeEstimator(BaseTester):
         patches = torch.rand(batch_size, channels, height, width, device=device, dtype=torch.float64)
         self.gradcheck(ori, (patches,), nondet_tol=1e-4)
 
+    @pytest.mark.parametrize("patch_size", [19, 32])
+    @pytest.mark.parametrize("batch_size", [1, 3])
+    def test_precision(self, device, dtype, patch_size, batch_size):
+        """Flat and anisotropic patches retain their shape and finite input gradients."""
+        patches = torch.zeros(batch_size, 1, patch_size, patch_size, device=device, dtype=dtype)
+        middle = patch_size // 2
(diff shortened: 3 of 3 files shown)


Hexecu:
Closing this as the patch-estimator fix has now been incorporated into #4122, together with broader affine-shape fallback coverage. I also rechecked the original CPU and MPS reproductions on current main, and both are resolved. Thanks!

PR #4132, opened 2026-08-31
Fix/4096 docstrings

### LINKED ISSUE #4096: 44 placeholder docstrings from #3737 say nothing, and one of them is wrong
#3737 added docstrings to every public callable that lacked one. 44 of them share the same
generated sentence, which carries no information about the function it documents:

> Output tensor or dictionary produced by the module while preserving the shape contract
> documented by the surrounding class.

```bash
grep -rn "preserving the shape contract documented by the" kornia/ | wc -l   # 44
```

22 of the 44 are in `kornia/feature/` (`xfeat.py`, `hynet.py`, `aliked/aliked.py`, all of
`dedode/`, `loftr/backbone/`, `loftr/loftr_module/`, `disk/_unets/`, `sold2/backbones.py`).

Where it is simply vacuous it is only noise. Where the surrounding class does something else,
it is wrong. `FilterResponseNorm2d.forward` in `kornia/feature/hynet.py` is documented as

> Run the HyNet normalization or descriptor layer. Patch tensors use `(B, C, H, W)`. For
> HyNet descriptors the expected input is usually grayscale `(B, 1, 32, 32)` and the final
> descriptor has shape `(B, D)`.

`FilterResponseNorm2d` is a normalization layer: it takes `(B, C, H, W)` and returns
`(B, C, H, W)`. It never produces a `(B, D)` descriptor.

`kornia/feature/dedode/decoder.py` documents its `feats` argument as

> feats: Input value used by this method.

## Also from #3737: three orphaned comments

Inserting the docstring after the first line of the body left the original leading comment
above it, where it now reads as a comment on the docstring:

```
kornia/feature/hynet.py:109              # Compute the mean norm of activations per channel.
kornia/feature/hynet.py:163              # nn.init.zeros_(self.tau)
kornia/feature/scale_space_detector.py:520   # Compute points per level
```

## Suggested fix

The 44 sites are individually small. Worth doing as one sweep per subpackage, replacing the
placeholder with the actual shape contract, and moving those three comments back below their
docstrings. Same defect class as #4067.

alinaqvi1129:
## What type of PR is this?

/kind documentation

## What this PR does / why we need it:

Fixes #4096

This PR replaces all 44 placeholder docstrings introduced by #3737 with function-specific documentation describing the actual input/output behavior and tensor shapes.

It also fixes the incorrect `FilterResponseNorm2d.forward` docstring, which previously described HyNet descriptor output even though the layer preserves the `(B, C, H, W)` tensor shape.

Additionally, the orphaned comments identified in #4096 have been moved back to their appropriate locations below the corresponding docstrings.

### Summary

* Replaced all 44 occurrences of the placeholder docstring.
* Removed the remaining `"Input value used by this method"` placeholder.
* Corrected `FilterResponseNorm2d.forward`.
* Fixed the orphaned comments from #3737.
* Kept the `FilterResponseNorm2d.forward` bug fix in a separate commit as requested.
* Kept docstring lines within the 120-character limit.
* Used raw docstrings where appropriate for `:math:` expressions.

## Special notes for your reviewer:

The docstrings were written based on the surrounding class contracts and the actual implementations, with tensor shapes documented where they are verifiable from the code.

## Does this PR introduce a user-facing change?

No

### DIFF
--- a/kornia/feature/aliked/aliked.py
+++ b/kornia/feature/aliked/aliked.py
@@ -459,15 +459,13 @@ class DeformableConv2d(nn.Module):
         )
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
-        """Run the module forward pass.
+        r"""Run the module forward pass.
 
         Args:
-            x: Input tensor processed by this module. For image-like features this usually follows the `(B, C, H, W)`
-                layout, where `B` is batch size, `C` is channels, and `H`/`W` are height and width.
+            x: Input feature map with shape :math:`(B, C_{\text{in}}, H, W)`.
 
         Returns:
-            Output tensor or dictionary produced by the module while preserving the shape contract documented by the
-            surrounding class.
+            Deformable convolution output with shape :math:`(B, C_{\text{out}}, H', W')`.
         """
         h, w = x.shape[2:]
         max_offset = max(h, w) / 4.0
@@ -540,15 +538,13 @@ class ConvBlock(nn.Module):
         self.bn2 = norm_layer(out_channels)
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
-        """Run the module forward pass.
+        r"""Run the module forward pass.
 
         Args:
-            x: Input tensor processed by this module. For image-like features this usually follows the `(B, C, H, W)`
-                layout, where `B` is batch size, `C` is channels, and `H`/`W` are height and width.
+            x: Input feature map with shape :math:`(B, C_{\text{in}}, H, W)`.
 
         Returns:
-            Output tensor or dictionary produced by the module while preserving the shape contract documented by the
-            surrounding class.
+            Feature map with shape :math:`(B, C_{\text{out}}, H, W)` after two gated conv-BN layers.
         """
         x = self.gate(self.bn1(self.conv1(x)))
         return self.gate(self.bn2(self.conv2(x)))
@@ -592,15 +588,14 @@ class ResBlock(nn.Module):
         self.stride = stride
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
-        """Run the module forward pass.
+        r"""Run the module forward pass.
 
         Args:
-            x: Input tensor processed by this module. For image-like features this usually follows the `(B, C, H, W)`
-                layout, where `B` is batch size, `C` is channels, and `H`/`W` are height and width.
+            x: Input feature map with shape :math:`(B, C_{\text{in}}, H, W)`.
 
         Returns:
-            Output tensor or dictionary produced by the module while preserving the shape contract documented by the
-            surrounding class.
+            Feature map with shape :math:`(B, C_{\text{out}}, H', W')` after the residual block,
+            where ``H'`` and ``W'`` may differ from the input if ``stride > 1``.
         """
         identity = x
         out = self.gate(self.bn1(self.conv1(x)))
--- a/kornia/feature/dedode/decoder.py
+++ b/kornia/feature/dedode/decoder.py
@@ -40,11 +40,12 @@ class Decoder(nn.Module):
     def forward(
         self, features: torch.Tensor, context: Optional[torch.Tensor] = None, scale: Optional[int] = None
     ) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
-        """Run this DeDoDe module forward.
+        r"""Run this DeDoDe module forward.
 
         Args:
             features: Encoded feature map with shape :math:`(B, C, H, W)`, or concatenated with context.
-            context: Optional context feature map from previous scale, shape :math:`(B, C_{\text{ctx}}, H, W)` or ``None``.
+            context: Optional context feature map from previous scale, shape
+                :math:`(B, C_{\text{ctx}}, H, W)`, or ``None``.
             scale: String key selecting the decoder layer for this scale (e.g. ``"8"``, ``"4"``, ``"2"``, ``"1"``).
 
         Returns:
--- a/kornia/feature/dedode/decoder.py
+++ b/kornia/feature/dedode/decoder.py
@@ -42,17 +42,13 @@ class Decoder(nn.Module):
     ) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
         """Run this DeDoDe module forward.
 
-        Inputs are image, feature, or token tensors used by the DeDoDe detector/descriptor pipeline. `B` denotes batch
-        size, `C` channels, `H` height, `W` width, `N` token count, and `D` feature dimension where those axes appear.
-
         Args:
-            features: Input value used by this method.
-            context: Input value used by this method.
-            scale: Input value used by this method.
+            features: Encoded feature map with shape :math:`(B, C, H, W)`, or concatenated with context.
+            context: Optional context feature map from previous scale, shape :math:`(B, C_ctx, H, W)` or ``None``.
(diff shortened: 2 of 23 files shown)


ducha-aiki:
**REQUEST CHANGES** — the sweep is real work and mostly accurate, but one replacement docstring states something the code does not do, and five `:math:` expressions will not render as intended. Both are cheap to fix.

Setup: worktree at the head `d6d8cae8`, merge-base `cec6d3db`; `~/dev/kornia/.venv/bin/python` (torch 2.9.1).

**The claim in the body checks out.** Zero occurrences of the `#3737` boilerplate survive in `kornia/`:

```
grep -rn "preserving the shape contract documented by the" kornia/   ->  0
grep -rn "Input value used by this method"                kornia/   ->  0
```

and the orphaned comments really are moved back below their docstrings (`hynet.py` `FilterResponseNorm2d.forward` and `TLU.reset_parameters` both read correctly now). The `FilterResponseNorm2d` correction you call out is right, and so is a quieter one I checked: `FineMatching.get_fine_match` is annotated `-> None` and ends in `data.update({...})`, so your "None. Updates `data` in place" replaces a docstring that was actively misleading.

## Blocking: one docstring is wrong

`kornia/feature/dedode/transformer/layers/dino_head.py:73` — "L2-normalized projection with shape :math:`(B, \text{out_dim})`". The normalization is applied to the bottleneck, *before* the final layer:

```python
x = self.mlp(x)
x = nn.functional.normalize(x, dim=-1, p=2, eps=eps)
x = self.last_layer(x)     # weight_norm(nn.Linear(bottleneck_dim, out_dim, bias=False))
return x
```

Measured:

```python
h = DINOHead(in_dim=8, out_dim=5); o = h(torch.rand(3, 8))
o.shape -> (3, 5)   o.norm(dim=-1) -> [0.1069, 0.1161, 0.1006]
```

Not unit norm. The shape half is right; the "L2-normalized" half is not. Suggested: "Projection logits with shape :math:`(B, \text{out\_dim})`. The bottleneck features are L2-normalized before the final weight-normalized linear layer."

This is the one class of defect a docstring sweep has to avoid — a confident wrong statement is worse than the placeholder it replaced, because nobody re-checks it.

For contrast, `vgg.py:69` is correct: `self.classifier` ends in `nn.Linear(4096, num_classes)`, so `(B, num_classes)` holds.

## Blocking-lite: five `:math:` expressions render wrong

```
kornia/feature/dedode/decoder.py:47   :math:`(B, C_ctx, H, W)`
kornia/feature/dedode/decoder.py:51   :math:`(B, C_out, H, W)`
kornia/feature/dedode/transformer/layers/dino_head.py:73   :math:`(B, \text{out_dim})`
kornia/feature/dedode/transformer/dinov2.py:387            :math:`(B, \text{out_dim})`
kornia/feature/dedode/vgg.py:69                            :math:`(B, \text{num_classes})`
```

In LaTeX `C_ctx` subscripts only the `c` and then sets `tx` on the baseline, and an unescaped `_` inside `\text{}` is a subscript operator in math mode. Use the form you already use correctly elsewhere in the same files: `C_{\text{ctx}}`, `C_{\text{out}}`, `\text{out\_dim}`, `\text{num\_classes}`.

I checked the one thing that would have made this worse and it is fine: every affected docstring is an `r"""` raw string, so no `\t` became a literal tab.

## Not blocking

- The `Run this DeDoDe module forward.` summary line survives in 20 places. That is defensible for a `forward`, and rewriting them is not what #4096 asked for — just noting it so nobody reads "all 44 replaced" as "no generic text left".
- These classes are not in the rendered API reference (`docs/source/feature.rst` autoclasses `DeDoDe` without `:members:`), so the LaTeX above is read in source, not on the docs site. Worth fixing anyway, since the point of the PR is that someone reads them.

Nice sweep otherwise — 23 files and the shapes I spot-checked were right.

ducha-aiki:
## Delta re-review of #4132 at `fedf23f9` — APPROVE

Both blockers from the [2026-08-31 review](https://github.com/kornia/kornia/pull/4132) are cleared. I re-executed each new claim rather than reading it.

### 1. The `dino_head.py` L2-norm claim — now accurate

New text:

> Projection logits with shape :math:`(B, \text{out\_dim})`. The bottleneck features are L2-normalized before the final weight-normalized linear layer.

Checked against the code at head:

```python
self.last_layer = weight_norm(nn.Linear(bottleneck_dim, out_dim, bias=False))   # :57
...
x = self.mlp(x)                                                                 # :76
x = nn.functional.normalize(x, dim=-1, p=2, eps=eps)                            # :78
x = self.last_layer(x)                                                          # :79
```

Every clause holds: the normalize is on the bottleneck, `last_layer` really is `weight_norm`-wrapped, and the return is logits rather than a normalized vector.

### 2. The five `:math:` slips — all escaped

`C_ctx` and `C_out` became `C_{\text{ctx}}` / `C_{\text{out}}`; `\text{out_dim}` (×2) and `\text{num_classes}` became `\text{out\_dim}` / `\text{num\_classes}`. Scanning the head of every file this PR touches for an unescaped `_` inside a `:math:` `\text{…}`, the only remaining hits are `kornia/feature/hynet.py:60,61,145,146` (`\text{num_features}`) — and those are **pre-existing on `main`** and not in this diff, so out of scope. Worth a separate one-liner some time, not a blocker here.

Approving. Thanks for turning both around quickly — the corrected `dino_head` sentence is more useful than the one-word fix I asked for.

ducha-aiki (dismissed):

ducha-aiki:
## Pushed the pre-commit fix to your branch — `5e34fe29`

@alinaqvi1129 you had `maintainer_can_modify` on, so rather than make you wait on another round-trip I pushed the two lint fixes straight onto `fix/4096-docstrings`. Nothing else changed; the whole commit is 3 insertions / 2 deletions in `kornia/feature/dedode/decoder.py`.

The two ruff errors were:

```
D301 Use `r"""` if any backslashes in a docstring   kornia/feature/dedode/decoder.py:43:9
E501 Line too long (123 > 120)                      kornia/feature/dedode/decoder.py:47:121
```

Worth knowing that **D301 was a real content bug here, not just a style rule**. Without the `r` prefix, `\t` in `\text{ctx}` is an escape, so the docstring that actually shipped had a literal TAB where the LaTeX was meant to be:

```python
>>> ast.get_docstring(Decoder.forward)   # before
'... :math:`(B, C_{<TAB>ext{ctx}}, H, W)` ...'   # TAB present: True
```

after:

```
TAB present: False   |   literal \text present: True
```

The 123-character line is wrapped onto a continuation, matching the style `encoder.py:58-59` already uses:

```
            context: Optional context feature map from previous scale, shape
                :math:`(B, C_{\text{ctx}}, H, W)`, or ``None``.
```

Verified before pushing, on the branch head:

```
ruff check kornia/ tests/                    All checks passed!
ruff format --check                          1 file already formatted
pre-commit run --all-files                   13 hooks, 0 Failed, exit 0
import kornia + Decoder.forward.__doc__      imports clean, no TAB
```

I also re-ran `ruff check --select D301,E501` across all of `kornia/` at your head — these were the **only** two occurrences, so no sibling file needs the same treatment.

My [approval](https://github.com/kornia/kornia/pull/4132#issuecomment-5508435208) stands. Merging once CI comes back green.

alinaqvi1129:
Thanks for the detailed review and for pushing the fixes! All CI checks are green now. Really appreciate the help and the explanation around the docstring and LaTeX issues. 🙌

ducha-aiki (approved):

PR #4133, opened 2026-08-31
fix(color): raise ShapeError for zero-sized YUV chroma planes

### LINKED ISSUE #4056: [Bug]: yuv420_to_rgb / yuv422_to_rgb raise ZeroDivisionError instead of ShapeError for zero-width chroma

Himesh-rupchandani:
Fixes #4056.

## Summary
`yuv420_to_rgb` / `yuv422_to_rgb` raised a bare `ZeroDivisionError` when the chroma plane had a zero-sized dimension. Following @ducha-aiki's guidance on #4056, the shape guards are reordered from division to multiplication so the existing `ShapeError` contract holds, while a *consistently* zero-sized luma/chroma pair returns an empty RGB plane, matching the 4:4:4 `yuv_to_rgb` convention (empty-in -> empty-out).

## Change
- `imagey.shape[-2] / imageuv.shape[-2] != 2` -> `imagey.shape[-2] != 2 * imageuv.shape[-2]`
- same for width, and in `yuv422_to_rgb`.

## Tests
- `test_exception`: zero-sized chroma dims raise `ShapeError`.
- `test_empty_input`: matching empty planes return `(3, 0, 0)` and agree with `yuv_to_rgb`.
- Local run: 58 passed.

## AI assistance
I used an AI assistant to reproduce the bug, position the fix, and draft the tests. I reviewed all changed lines and verified the behaviour claims locally.

### DIFF
--- a/kornia/color/yuv.py
+++ b/kornia/color/yuv.py
@@ -228,8 +228,8 @@ def yuv420_to_rgb(imagey: torch.Tensor, imageuv: torch.Tensor) -> torch.Tensor:
     if (
         len(imageuv.shape) < 2
         or len(imagey.shape) < 2
-        or imagey.shape[-2] / imageuv.shape[-2] != 2
-        or imagey.shape[-1] / imageuv.shape[-1] != 2
+        or imagey.shape[-2] != 2 * imageuv.shape[-2]
+        or imagey.shape[-1] != 2 * imageuv.shape[-1]
     ):
         raise ShapeError(
             f"Input imageuv H&W must be half the size of the luma plane. Got {imagey.shape} and {imageuv.shape}"
@@ -280,7 +280,7 @@ def yuv422_to_rgb(imagey: torch.Tensor, imageuv: torch.Tensor) -> torch.Tensor:
         len(imageuv.shape) < 2
         or len(imagey.shape) < 2
         or imagey.shape[-2] != imageuv.shape[-2]
-        or imagey.shape[-1] / imageuv.shape[-1] != 2
+        or imagey.shape[-1] != 2 * imageuv.shape[-1]
     ):
         raise ShapeError(
             f"Input imageuv H must match the luma plane and W must be half its size. "
--- a/tests/color/test_yuv.py
+++ b/tests/color/test_yuv.py
@@ -699,6 +699,29 @@ class TestYuv420ToRgb(BaseTester):
             imguv = torch.ones(2, 1, 1, device=device, dtype=dtype)
             kornia.color.yuv420_to_rgb(imgy, imguv)
 
+        # Regression for #4056: a zero-sized chroma dimension is still a shape violation and must
+        # be reported as ShapeError, not the bare ZeroDivisionError the old ratio guard threw.
+        with pytest.raises(ShapeError):
+            imgy = torch.ones(1, 4, 4, device=device, dtype=dtype)
+            imguv = torch.ones(2, 2, 0, device=device, dtype=dtype)
+            kornia.color.yuv420_to_rgb(imgy, imguv)
+
+        with pytest.raises(ShapeError):
+            imgy = torch.ones(1, 4, 4, device=device, dtype=dtype)
+            imguv = torch.ones(2, 0, 2, device=device, dtype=dtype)
+            kornia.color.yuv420_to_rgb(imgy, imguv)
+
+    def test_empty_input(self, device, dtype):
+        # Regression for #4056: a *consistently* zero-sized luma/chroma pair passes the guard
+        # (0 == 2 * 0) and returns an empty RGB plane, matching the 4:4:4 twin, rather than
+        # raising. Pin the empty-in -> empty-out convention instead of special-casing it away.
+        y = torch.rand(1, 0, 0, device=device, dtype=dtype)
+        uv = torch.rand(2, 0, 0, device=device, dtype=dtype)
+        out = kornia.color.yuv420_to_rgb(y, uv)
+        assert out.shape == (3, 0, 0)
+        # agrees with the 4:4:4 converter on the same empty concatenation
+        assert torch.equal(kornia.color.yuv_to_rgb(torch.cat([y, uv], dim=-3)), out)
+
     @pytest.mark.parametrize("name", list(_REFERENCE_COLORS))
     def test_unit(self, device, dtype, name):
         rgb_values, yuv_values = _REFERENCE_COLORS[name]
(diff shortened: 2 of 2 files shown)


ducha-aiki:
**The fix is verified correct — but this PR is a byte-for-byte duplicate of #4111, which arrived first.** Flagging that before anything else so neither of you spends more time on it.

Setup: worktree at this head `96f61aba` and at #4111's head `5b431bd9`, both with `kornia/` swapped to their merge-base for the fail-on-base pass; `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS.

**The production hunks are identical** — the same `!= 2 *` reorder in `yuv420_to_rgb` (both dimensions) and `yuv422_to_rgb` (width), which is the reordering I asked for on #4056. Both PRs' test suites also fail identically on their base (4 failed / 104 passed) and pass identically at their head (214 passed at float32+float64).

**The behaviour change is exactly what was specified, over a 4802-cell sweep** (`H_y, W_y, H_uv, W_uv` each over `{0,1,2,3,4,6,8}`, both converters), head against merge-base:

```
total probes: 4802   changed: 220
  ZeroDivisionError -> ShapeError        206
  ZeroDivisionError -> empty output       14
  worked -> raises                         0
  worked -> different result               0
```

The 14 pass-throughs are the consistently-zero pairs — `(1,0,0)` + `(2,0,0)` -> `(3,0,0)`, `(1,0,2)` + `(2,0,1)` -> `(3,0,2)`, and so on — i.e. the empty-in -> empty-out convention, matching the 4:4:4 twin. That is the answer given on #4056, so this is the intended half, not an accident.

**Where this PR is better than #4111:** the pins. The `test_empty_input` cases here cross-check against the sibling converter —

```python
assert torch.equal(kornia.color.yuv_to_rgb(torch.cat([y, uv], dim=-3)), out)
```

— which pins the *reason* the empty case is allowed rather than just its shape, and the comments name #4056 so the next reader can find the decision. #4111's version asserts the shape only.

**Disposition (maintainer's call, stated so it is on the record):** I intend to land this one and close #4111 with credit, because the production change is identical and the tie-break is test quality. @Marchematics got there first with a correct fix and that is worth saying plainly — the code that ships is yours as much as anyone's.

Nothing to change here. Holding the merge only until #4111's author has seen this.

Himesh-rupchandani:
Thanks @ducha-aiki for the careful verification, and credit to @Marchematics for getting there first — same fix, and glad the test cross-check landed. 

I'll defer to you on the merge timing. No changes needed on my end. Happy to make any adjustments if you'd prefer.

ducha-aiki (approved):

PR #4134, opened 2026-08-31
fix ScaleSpaceDetector dynamo config scope

### LINKED ISSUE #4095: ScaleSpaceDetector(compile_modules=...) permanently mutates a process-global torch._dynamo config flag
Constructing a `ScaleSpaceDetector` with any non-empty `compile_modules` permanently flips a
process-global dynamo flag, for every other torch user in the process.

## Reproduction

```python
import torch._dynamo as dyn
print("before:", dyn.config.capture_dynamic_output_shape_ops)

from kornia.feature import ScaleSpaceDetector
_ = ScaleSpaceDetector(num_features=5, compile_modules=["resp"])

print("after :", dyn.config.capture_dynamic_output_shape_ops)
```

```
before: False
after : True
```

The detector need not even be called, and deleting it does not restore the flag.

## Where

`kornia/feature/scale_space_detector.py:145`

```python
if _compile_set:
    torch._dynamo.config.capture_dynamic_output_shape_ops = True
```

The comment above it explains why the flag is wanted (keeping `torch.where`/`nonzero` inside
the graph as unbacked symbols instead of taking a graph break), and that reasoning is sound
for this detector's own compiled submodules. The problem is the scope: it is a global, set
from a constructor, never restored.

Anything else the user compiles afterwards in the same process now also captures
dynamic-output-shape ops, which changes their guard behaviour and can turn a previously
static graph into a dynamic one.

## Suggested fix

Wrap the compiled calls in `torch._dynamo.config.patch(capture_dynamic_output_shape_ops=True)`
at the point of use rather than setting the module-level attribute, so the change is scoped to
this detector's forward. `torch._dynamo.config.patch` works both as a context manager and as a
decorator.

Reproduced on `main` at 9abf06d9, torch 2.9.1, CPU.

X-mo7:
ScaleSpaceDetector.__init__ currently sets torch._dynamo.config.capture_dynamic_output_shape_ops = True whenever compile_modules is non-empty. This mutates the global Dynamo configuration and does not restore its previous value after. This PR removes that global mutation and assigns capture_dynamic_output_shape_ops=True to calls of the selected compiled submodules using torch._dynamo.config.patch(...). The selected module names are stored on the detector so the patch is only applied where needed. I added Regression tests to verify that initialization does not mutate the global config, that the config is restored after a forward pass, and that compiled and eager execution produce matching outputs.

Fixes #4095

## What did you check?
- 3 regression tests: passed
- Full `tests/feature/test_scale_space_detector.py`: 64 passed
- `ruff check`: passed
- `ruff format --check`: passed
- `git diff --check`: clean
- Typecheck: no errors related to this change

## How did AI help?

AI (codex) was used to help draft the initial implementation, the regression tests, and review the final diff. I manually inspected the changes and ran the necessary tests and checks locally.

### DIFF
--- a/.github/workflows/pr_test_cpu.yml
+++ b/.github/workflows/pr_test_cpu.yml
@@ -166,7 +166,7 @@ jobs:
       pytorch-version: '["2.14.0"]'
       pytorch-dtype: float32
       optimizer: inductor
-      pytest-extra: 'tests/augmentation tests/enhance tests/filters tests/color tests/geometry tests/losses tests/morphology -k "dynamo or compile"'
+      pytest-extra: 'tests/augmentation tests/enhance tests/feature tests/filters tests/color tests/geometry tests/losses tests/morphology -k "dynamo or compile"'
       cache-path: weights/
       cache-key: model-weights-${{ needs.pre-tests.outputs.hash }}
       cache-restore-keys: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -238,6 +238,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* Constructing `ScaleSpaceDetector` with `compile_modules` no longer permanently mutates the process-global
+  `torch._dynamo.config.capture_dynamic_output_shape_ops` setting. (#4134)
+
 * `Boxes3D.to_tensor` and `Boxes3D.get_boxes_shape` are differentiable again (#1396). Both reduce a box's 8
   vertices to its min/max corner with `amin`/`amax`, which is a genuine kink -- not differentiable in the
   classical sense -- wherever multiple vertices exactly tie for an axis extremum. Every face of an
--- a/kornia/feature/scale_space_detector.py
+++ b/kornia/feature/scale_space_detector.py
@@ -16,7 +16,8 @@
 #
 
 import math
-from typing import List, Optional, Tuple, Union
+from contextlib import nullcontext
+from typing import ContextManager, List, Optional, Tuple, Union
 
 import torch
 import torch.nn.functional as F
@@ -249,15 +250,10 @@ class ScaleSpaceDetector(nn.Module):
             if unknown:
                 raise ValueError(f"Unknown module names in compile_modules: {unknown}. Valid: {_all_names}")
 
-        if _compile_set:
-            # Allow torch.compile to keep data-dependent shape ops (torch.where / nonzero)
-            # inside the compiled graph as unbacked symbols, avoiding graph breaks and the
-            # 0/1-specialization recompilations that would otherwise fire whenever an octave
-            # first encounters zero NMS maxima (blurry/extreme-viewpoint images).
-            torch._dynamo.config.capture_dynamic_output_shape_ops = True
+        self._compile_set: frozenset[str] = frozenset(_compile_set)
 
         def _maybe_compile(mod: nn.Module, name: str) -> nn.Module:
-            return torch.compile(mod, dynamic=True) if name in _compile_set else mod
+            return torch.compile(mod, dynamic=True) if name in self._compile_set else mod
 
         if scale_pyr_module is None:
             extra_levels = 3 if scale_space_response else 2
@@ -283,6 +279,15 @@ class ScaleSpaceDetector(nn.Module):
         # scale_space_response should be True if the response function works on scale space
         # like Difference-of-Gaussians
         self.scale_space_response = scale_space_response
+
+    def _dynamo_config_patch(self, modules: Tuple[str, ...]) -> ContextManager[None]:
+        if self._compile_set.intersection(modules):
+            # Allow torch.compile to keep data-dependent shape ops (torch.where / nonzero)
+            # inside the compiled graph as unbacked symbols, avoiding graph breaks and the
+            # 0/1-specialization recompilations that would otherwise fire whenever an octave
+            # first encounters zero NMS maxima (blurry/extreme-viewpoint images).
+            return torch._dynamo.config.patch(capture_dynamic_output_shape_ops=True)
+        return nullcontext()
 
     def __repr__(self) -> str:
         return (
@@ -320,9 +325,11 @@ class ScaleSpaceDetector(nn.Module):
 
         # Run response function
         if self.scale_space_response:
-            oct_resp = self.resp(octave, sigmas_oct.view(-1))  # (B, 1, Ldog, H, W)
+            with self._dynamo_config_patch(("resp",)):
+                oct_resp = self.resp(octave, sigmas_oct.view(-1))  # (B, 1, Ldog, H, W)
         else:
-            level_resp = self.resp(octave.permute(0, 2, 1, 3, 4).reshape(B * L, CH, H, W), sigmas_oct.view(-1))
+            with self._dynamo_config_patch(("resp",)):
+                level_resp = self.resp(octave.permute(0, 2, 1, 3, 4).reshape(B * L, CH, H, W), sigmas_oct.view(-1))
             KORNIA_CHECK(
                 level_resp.dim() == 4
                 and level_resp.shape[0] == B * L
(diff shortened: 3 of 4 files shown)


ducha-aiki:
## Review of #4134 — CHANGES REQUESTED (the fix is right; its tests run nowhere)

Setup: worktree at the PR head (`fcd7cb16`) with `kornia/` swapped to the merge-base `cec6d3db` for every A/B and fail-on-base pass. `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS; `kornia.__file__` asserted inside the worktree in every block.

**The production change is correct and I verified it four ways.** The blocker is entirely on the test side, and it is not something you could reasonably have seen from a local run.

---

## What holds

**The issue's own reproduction, at this head:**

```
before     : False
after ctor : False        (main: True)
after fwd  : False        (main: True)
```

**The compiled path is unchanged.** This was my main worry — the flag is read when dynamo *traces*, and moving it from a constructor global to a per-call `config.patch` could in principle change what gets captured. Same detector, same three inputs, `torch._dynamo.utils.counters`:

```
                    graph_breaks   compiled_frames   flag left set afterwards
base (cec6d3db)          0                2                   True
head                     0                2                   False
```

and the outputs are **bit-identical** on all three, including the blank image that produces zero NMS maxima on every octave — the exact case the comment you moved cites as the reason for the flag:

```
  textured: base==head bit-identical: True
     blank: base==head bit-identical: True
 textured2: base==head bit-identical: True
```

Backward through a compiled detector also still works on both revisions, which I checked because the backward graph compiles *outside* your `with` blocks.

**The eager path pays nothing.** `_dynamo_config_patch` is called **9 times per forward**, and with `compile_modules=False` every one returns `nullcontext()`:

```
eager forward, compile_modules=False:   base 12.940 ms   head 12.800 ms   (20 calls, warm)
```

**Everything else:** 122/122 in `tests/feature/test_scale_space_detector.py` at f32+f64, 626 passed / 81 skipped across all of `tests/feature`, `ruff check` clean, `grep -rnE "#4095|issues/4095" kornia/ tests/` empty so there is no pin surgery owed. The `minima_are_also_good` / `is_iterative_subpix` block is re-indented but logically identical — I read it branch by branch. Python floor is 3.11 so `frozenset[str]` is fine.

---

## Blocker 1 — the three regression tests are never executed by CI

Reverting `kornia/` to the merge-base and re-running the whole file gives **122 passed**, the same as at head. The new tests are not weak; they are *not collected*:

```
$ pytest tests/feature/test_scale_space_detector.py --collect-only -q --device=cpu --dtype=float32
$ ... | grep -c compile_modules
0
```

`conftest.py:173` deselects any test whose **name** contains `dynamo` or `compile` when `KORNIA_TEST_OPTIMIZER` is unset. All three are named `test_compile_modules_*`. With the variable set they behave exactly as you intended:

```
KORNIA_TEST_OPTIMIZER=inductor, head:   3 passed
KORNIA_TEST_OPTIMIZER=inductor, base:   3 failed
    AssertionError: assert not True
      where True = torch._dynamo.config.capture_dynamic_output_shape_ops
```

So they discriminate perfectly. The problem is that **no CI job sets the variable in a way that reaches them.** I traced all four:

| Job | Runs them? | Why not |
| --- | --- | --- |
| `pr_test_cpu.yml` → `tests` matrix | no | `KORNIA_TEST_OPTIMIZER` unset → name filter |
| `pr_test_cpu.yml` → `dynamo` leg | no | its `pytest-extra` lists `tests/augmentation tests/enhance tests/filters tests/color tests/geometry tests/losses tests/morphology` — **`tests/feature` is not there** |
| `coverage.yml` → `dynamo` job | no | `-k "test_dynamo"` does not match `test_compile_modules_*` |
| `coverage.yml` → `overall` / `slow` | no | optimizer unset / `-k "slow and not test_dynamo"` |

`coverage.yml` is also only reached from `scheduled_test_cpu.yml`, not from PR CI at all. Net: #4095 would land fixed and completely unguarded — the global mutation could come back in any later refactor with a green board.

**Preferred fix — add `tests/feature` to the `dynamo` leg** in `.github/workflows/pr_test_cpu.yml:150`. Measured cost:

```
tests/feature -k "dynamo or compile"   ->  9 tests, 707 deselected
cold inductor cache:  63.9 s
```

One extra minute on a leg that already exists, no renames, and it also picks up **six pre-existing `test_dynamo` tests under `tests/feature` that are equally unguarded at PR time today** — a gap worth closing anyway.

**Either way, please rename `test_compile_modules_do_not_mutate_dynamo_config_on_init`** to drop both banned substrings. It never actually compiles — `torch.compile` is lazy, so the constructor only wraps — and it costs **4.12 ms**:

```
construction-only pin:  4.12 ms   flag_leaked=False   (base: 3.00 ms, flag_leaked=True)
```

That is the literal reproduction from the issue ("The detector need not even be called"), it is free, and it belongs in the default suite whatever happens to the other two.

For reference if you would rather rename all three instead: the trio costs **40.5 s on a cold inductor cache**, 5.2 s warm. CI runners are always cold, and that is per matrix cell — which is why I prefer the workflow one-liner.

## Blocker 2 — no CHANGELOG entry

AGENTS.md requires one per user-visible change, under `## Unreleased` → `Bug fixes`, citing the PR number. "Constructing `ScaleSpaceDetector` with `compile_modules` no longer permanently flips the process-global `torch._dynamo.config.capture_dynamic_output_shape_ops`" is squarely user-visible.

## Two smaller asks

3. **Name the issue in the test names.** The repo convention is a `_<issue>` suffix, and `grep -rnE "#4095|issues/4095" kornia/ tests/` is empty at your head, so nothing points a future reader back at the report. Folding this into the rename costs nothing.
4. **Rebase.** `git merge-tree` says the only conflict is `tests/feature/test_scale_space_detector.py`, from #4122 and #4128 both touching the import block since your merge-base. `kornia/feature/scale_space_detector.py` auto-merges. Per [[conflicting-pr-blocks-all-ci]] the checks currently shown were computed before `main` moved.

## Non-blocking

- A **direct** call to a compiled submodule (`det.scale_pyr(img)`, `det.subpix(...)`) now happens outside the patch, where on `main` the global flag covered it. I measured `det.scale_pyr(img)` at **0 graph breaks** on both revisions, so there is no observable regression; noting it only because it is a genuine narrowing of scope, and it is the right narrowing.
- `typing.ContextManager` is deprecated in favour of `contextlib.AbstractContextManager` (UP035), but that rule is not selected in this repo and the file already imports `List`/`Tuple` from `typing`, so your version is the locally consistent one. No change wanted.

---

Thanks for this — routing the patch through a `_dynamo_config_patch(modules)` helper keyed on `self._compile_set`, rather than blanket-wrapping the whole forward, is a better design than the one the issue suggested, and the bit-identity and graph-break numbers above are the proof that it costs nothing. Get the tests somewhere CI can see them and I will merge it.

ducha-aiki:
## Delta re-review of #4134 at `fab7f9cc` — APPROVE; only a CHANGELOG rebase left

Setup: worktree at the PR head, `kornia/` swapped to the merge-base `023028b6` for the fail-on-base pass. `~/dev/kornia/.venv/bin/python`, **torch 2.14.0** (the `.venv` moved off 2.9.1 with #4197), CPU.

All four asks from the 09-02 review are addressed, and the coverage one — the actual blocker — is addressed the way I hoped rather than the cheap way.

### The blocker is closed: the tests now run in a real job

You added `tests/feature` to the dynamo leg *and* renamed the construction pin so it does not need that leg. Verified by collection rather than by reading:

```text
default run (KORNIA_TEST_OPTIMIZER unset), -k 4095:
    1 passed        <- test_scale_space_detector_init_preserves_global_config_4095 only
                       (the other two carry "compile" in the name and are name-filtered out,
                        which is correct: they are the ones that need the inductor leg)

dynamo leg simulation, KORNIA_TEST_OPTIMIZER=inductor, tests/feature -k "dynamo or compile":
    15/789 tests collected
```

15 is the number I predicted from the other side: your 2 compile tests plus the pre-existing `test_dynamo` tests under `tests/feature` that were collected by no job before this change. So the directory addition fixes your PR and closes the same gap for six tests that were already there.

Fail-on-base with the optimizer set, `kornia/` reverted to `023028b6`:

```text
FAILED test_scale_space_detector_init_preserves_global_config_4095   assert not True
FAILED test_compile_modules_forward_restores_dynamo_config_4095      assert not True
FAILED test_compile_modules_match_eager_4095                         assert not True
3 failed
```

**3 fail on base, 3 pass at head** — a real pin now, where whole-file fail-on-base was 122-passed-both-sides last round. Head, optimizer set: 3 passed in 31 s. Whole file, default run: **66 passed**.

The other two asks are done as well — the tests name `#4095` in their own names, and there is a CHANGELOG entry.

The construction pin renaming is worth a word: naming it `..._init_preserves_global_config_4095` rather than `test_compile_modules_init_*` is what moves it into the ordinary `tests` matrix, where it costs the 4 ms it should, instead of paying 40 s of cold compile in the inductor leg for an assertion about a constructor that never compiles. That is the right split.

### The one thing left

`CONFLICTING/DIRTY`, and `git merge-tree` reports the conflict as **`CHANGELOG.md` only** — your `pr_test_cpu.yml` hunk merges cleanly even though `main` has since inserted the `half-precision` and `tests-mps` jobs above it (the leg is at line 169 on `main`, 150 on your branch). Rebase and it is ready.

Thank you for taking the CI-coverage finding seriously rather than arguing the tests passed locally.

ducha-aiki:
## Delta re-review of #4134 at `d254c679` — APPROVE, nothing left; ready to merge

You rebased onto `7c5db705` and added the CHANGELOG entry I asked for. Checked that the rebase carried nothing else:

```console
$ diff <(git show fab7f9cc:kornia/feature/scale_space_detector.py) <(git show pr4134:kornia/feature/scale_space_detector.py)
$ diff <(git show fab7f9cc:tests/feature/test_scale_space_detector.py) <(git show pr4134:tests/feature/test_scale_space_detector.py)
$
```

Both **byte-identical** to the revision I approved, so the verified findings carry over unchanged — the default run collects the renamed construction pin, the dynamo-leg simulation over `tests/feature -k "dynamo or compile"` collects 15 of 789 (your two plus the six pre-existing `test_dynamo` tests that no job collected before), and fail-on-base with the optimizer set is 3 failed.

The only new content is:

```
+* Constructing `ScaleSpaceDetector` with `compile_modules` no longer permanently mutates the process-global
+  `torch._dynamo.config.capture_dynamic_output_shape_ops` setting. (#4134)
```

which is accurate and in the right section. The `pr_test_cpu.yml` hunk merged cleanly past the two jobs `main` inserted above it.

All 36 checks green, `MERGEABLE`, closes #4095. Nothing owed by you — thanks for turning this around.

ducha-aiki (approved):

PR #4135, opened 2026-08-31
test(feature): revive LoFTR's two accuracy tests on CPU

### LINKED ISSUE #4092: LoFTR's two accuracy tests are permanently skipped by skipif(torch_version_ge(1, 10))
`tests/feature/test_loftr.py` guards its only two numerical tests with

```python
@pytest.mark.skipif(torch_version_ge(1, 10), reason="RuntimeError: CUDA out of memory with pytorch>=1.10")
```

at `tests/feature/test_loftr.py:43` and `:55`. `pyproject.toml` declares `torch>=2.0.0`, so
the predicate is unconditionally true: `test_pretrained_indoor` and `test_pretrained_outdoor`
can never run on any supported torch, on any device, with or without `--runslow`.

What is left for LoFTR is three smoke tests plus a shape test. `kornia/feature/loftr/` sits at
**21%** statement coverage (`loftr.py` 21.1%, `loftr_module/*` 14.8-24.5%,
`utils/coarse_matching.py` 8.7%, `utils/geometry.py` and `utils/supervision.py` 0%), and the
two tests that would have exercised the matcher end-to-end against stored expected outputs are
the dead ones.

The `reason=` also does not survive scrutiny: an out-of-memory failure on CUDA is not a reason
to skip on CPU, and the test is already `skipif(sys.platform == "win32")` for the memory cost
on Windows CI.

## Repo-wide: the other version guards

The same `tests/_utils`-style version predicates appear elsewhere and are all dead against
`torch>=2.0.0`, though the rest are harmless no-ops that let their tests run:

```
tests/feature/test_integrated.py:334,352,380,399   torch_version_le(1, 9, 1)
tests/feature/test_matching.py:529                 torch_version_le(1, 9, 1)
tests/feature/test_lightglue_onnx.py:34            torch_version_le(1, 9, 1)
tests/image/test_image.py:177                      torch_version_le(1, 9, 1)
tests/tracking/test_planar_tracking.py:217         torch_version_le(1, 9, 1)
tests/models/test_efficient_vit.py:54              torch_version_lt(2, 0, 0)
tests/models/test_object_detector.py:56            torch_version_lt(2, 0, 0)
tests/geometry/transform/test_imgwarp.py:277       torch_version_lt(1, 11, 0)
```

Only the two `torch_version_ge(1, 10)` guards in `test_loftr.py` actually disable anything.

## Suggested fix

1. Re-enable the two LoFTR tests. If they really do exhaust CUDA memory on the CI runners,
   gate them on the device (`skipif(device == "cuda")`) or shrink the fixture, rather than on
   a torch version that no supported install can satisfy.
2. Delete the eight dead `torch_version_le(1, 9, 1)` / `torch_version_lt(...)` decorators as
   cleanup.

Found while measuring `kornia.feature` coverage on `main` at 9abf06d9.

adimalkar:
Fixes #4092

## Problem

`tests/feature/test_loftr.py` guards its only two numerical tests with

```python
@pytest.mark.skipif(torch_version_ge(1, 10),
                    reason="RuntimeError: CUDA out of memory with pytorch>=1.10")
```

`pyproject.toml` requires `torch>=2.0.0`, so the predicate is unconditionally true.
`test_pretrained_indoor` and `test_pretrained_outdoor` have therefore been skipped on every
supported torch, on every device, with or without `--runslow`. They are the only tests that
exercise the matcher end-to-end against stored expectations; `kornia/feature/loftr/` sits at
21% statement coverage as a result.

Confirmed on torch 2.13.0 — `torch_version_ge(1, 10)` is `True`, and the two tests report
`2 skipped`.

## Change

Remove the dead guard so both tests run. On CPU they pass:

```
tests/feature/test_loftr.py  6 passed, 1 skipped
```

The remaining skip is the pre-existing `win32` memory guard. CI runs
`KORNIA_TEST_DEVICE: cpu`, so these two now execute on every PR.

The now-unused `torch_version_ge` import is removed with it.

## Why CUDA is skipped explicitly

The tests do **not** pass on CUDA — but not for the recorded reason. Running them on an
RTX 4050 (6 GB, torch 2.13.0+cu130) produces no out-of-memory error at any point. The
failures are numerical:

```
test_pretrained_indoor
  Mismatched elements: 476 / 1278 (37.2%)
  Greatest absolute difference: 416.0 at index (80, 0) (up to 1e-05 allowed)

test_pretrained_outdoor
  The values for attribute 'shape' do not match:
  torch.Size([1720, 2]) != torch.Size([1719, 2])
```

The stored expectations were generated on CPU and the CUDA matcher selects a different set
of correspondences — the match *count* differs, so this is not a tolerance question, and a
416-pixel difference is a different keypoint rather than a rounding artifact.

So CUDA is skipped with that reason recorded in place of the OOM one, using the `device.type`
skip convention already used throughout the suite (60 sites, including CUDA-specific ones in
`tests/geometry/transform/test_affine.py` and `tests/geometry/test_grid.py`).

This is a deliberate scope limit: the CPU/CUDA divergence is a much larger question
(regenerate expectations per device? is the matcher nondeterministic on CUDA? is it a genuine
bug?) and is currently invisible because kornia has no GPU CI. I have described it on #4092
and am happy to file it separately if you would like it tracked.

## Not included

#4092 also lists other dead version predicates (`torch_version_le(1, 9, 1)` in
`test_integrated.py`, `test_matching.py`, `test_planar_tracking.py`). Those are harmless
no-ops — they evaluate false on `torch>=2.0.0`, so their tests do run — and I have left them
out to keep this reviewable. Happy to follow up.

## Verification

- CPU: `pytest tests/feature/test_loftr.py --runslow` -> 6 passed, 1 skipped
- CUDA: `pytest tests/feature/test_loftr.py --runslow --device cuda` -> 4 passed, 3 skipped
- `ruff check` and `ruff format --check` clean

## AI assistance disclosure

This change was developed with AI assistance (Claude). The diagnosis, the patch and this
description were AI-drafted; the CPU and CUDA runs above were executed on my machine and the
numbers are from those runs.

### DIFF
--- a/tests/feature/test_loftr.py
+++ b/tests/feature/test_loftr.py
@@ -20,7 +20,6 @@ import sys
 import pytest
 import torch
 
-from kornia.core._compat import torch_version_ge
 from kornia.feature import LoFTR
 from kornia.geometry import resize
 
@@ -40,10 +39,16 @@ class TestLoFTR(BaseTester):
         assert loftr is not None
 
     @pytest.mark.slow
-    @pytest.mark.skipif(torch_version_ge(1, 10), reason="RuntimeError: CUDA out of memory with pytorch>=1.10")
     @pytest.mark.skipif(sys.platform == "win32", reason="this test takes so much memory in the CI with Windows")
     @pytest.mark.parametrize("data", ["loftr_fund"], indirect=True)
     def test_pretrained_indoor(self, device, dtype, data):
+        if device.type == "cuda":
+            # The stored expectations were generated on CPU. The CUDA matcher
+            # selects a different set of correspondences (~37% of coordinates
+            # differ, and the match count itself changes), so this comparison
+            # cannot hold on CUDA until the expectations are regenerated
+            # per-device. See https://github.com/kornia/kornia/issues/4092.
+            pytest.skip("stored keypoint expectations are CPU-specific")
         loftr = LoFTR("indoor").to(device, dtype)
         data_dev = dict_to(data, device, dtype)
         with torch.no_grad():
@@ -52,10 +57,16 @@ class TestLoFTR(BaseTester):
         self.assert_close(out["keypoints1"], data_dev["loftr_indoor_tentatives1"])
 
     @pytest.mark.slow
-    @pytest.mark.skipif(torch_version_ge(1, 10), reason="RuntimeError: CUDA out of memory with pytorch>=1.10")
     @pytest.mark.skipif(sys.platform == "win32", reason="this test takes so much memory in the CI with Windows")
     @pytest.mark.parametrize("data", ["loftr_homo"], indirect=True)
     def test_pretrained_outdoor(self, device, dtype, data):
+        if device.type == "cuda":
+            # The stored expectations were generated on CPU. The CUDA matcher
+            # selects a different set of correspondences (~37% of coordinates
+            # differ, and the match count itself changes), so this comparison
+            # cannot hold on CUDA until the expectations are regenerated
+            # per-device. See https://github.com/kornia/kornia/issues/4092.
+            pytest.skip("stored keypoint expectations are CPU-specific")
         loftr = LoFTR("outdoor").to(device, dtype)
         data_dev = dict_to(data, device, dtype)
         with torch.no_grad():


ducha-aiki:
**APPROVE.** Verified by running the revived tests, not by reading the diff.

Setup: worktree at the PR head `166e23d5`, merge-base `cec6d3db`; `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS, `KORNIA_TEST_RUNSLOW=true`.

**The skip really was unconditional.** At the merge-base, with `--runslow` on:

```
SKIPPED [1] tests/feature/test_loftr.py:42: RuntimeError: CUDA out of memory with pytorch>=1.10
SKIPPED [1] tests/feature/test_loftr.py:54: RuntimeError: CUDA out of memory with pytorch>=1.10
2 skipped in 0.29s
```

**At the head they run and pass**, and they are cheap:

```
6 passed, 1 skipped in 59.22s
  5.11s call  TestLoFTR::test_pretrained_indoor[cpu-float32-loftr_fund]
  2.82s call  TestLoFTR::test_pretrained_outdoor[cpu-float32-loftr_homo]
  1.78s setup / 1.30s setup
 38.15s call  test_pretrained_indoor_smoke      <- cold weights download, not this PR
```

So the two revived tests add roughly 8 s of compute plus 3 s of fixture setup to the slow job, on top of a weights download the smoke tests already pay for.

The CUDA branch is the right call for now: the stored expectations are CPU-generated, the tests stay `@pytest.mark.slow`, and the skip names #4092 so the follow-up (per-device expectations) is findable. Worth stating in the issue that this leaves CUDA uncovered by design rather than by accident.

adimalkar:
Thanks for verifying by running it rather than reading the diff — and for posting the merge-base numbers, which make the before/after unambiguous.

Done: I have stated the CUDA position explicitly on #4092 (https://github.com/kornia/kornia/issues/4092#issuecomment-5485213953) — that after this PR CUDA is uncovered **by design**, with a table of what is and is not covered before vs after, and the note that the blocker is the CPU-generated expectations rather than memory.

One thing your timings clarify that mine did not: the 38s is the cold weights download for the smoke tests, which this PR does not add. The two revived tests themselves are ~8s of compute plus ~3s fixture setup. Useful framing for the slow job — thank you.

Happy to open a separate issue for the CPU/CUDA divergence itself if you would like it tracked apart from #4092.

ducha-aiki (approved):

Issue #4017, opened 2026-08-26
W1b: `Boxes` advertises paths that always raise `NotImplementedError` — `trim()`, tuple bounds in `clamp()`, `translate(method='fast')`


ducha-aiki:
Three documented entry points on `Boxes` cannot be used:

- `Boxes.trim()` has a 20-line docstring describing how zero-padded boxes are removed, and a body that is `raise NotImplementedError` (`boxes.py:414-442`).
- `Boxes.clamp(topleft, botright)` is typed `Optional[torch.Tensor | tuple[int, int]]` and documents tensor bounds, but any non-tensor input hits `raise NotImplementedError` (`boxes.py:395-396`).
- `Boxes.translate(size, method="fast")` — `"fast"` is listed in the docstring and raises (`boxes.py:757-758`).

## Repro

```python
import torch
from kornia.geometry.boxes import Boxes

b = Boxes(torch.tensor([[[1., 1.], [4., 1.], [4., 3.], [1., 3.]]]))
for name, fn in [("trim()", lambda: b.trim()),
                 ("clamp((0, 0), (4, 4))", lambda: b.clamp((0, 0), (4, 4))),
                 ("translate(size, method='fast')", lambda: b.translate(torch.tensor([[1., 1.]]), method="fast"))]:
    try:
        fn(); print(name, "-> ok")
    except NotImplementedError as e:
        print(name, "-> NotImplementedError", repr(str(e)))
```

## Observed

```
trim() -> NotImplementedError ''
clamp((0, 0), (4, 4)) -> NotImplementedError ''
translate(size, method='fast') -> NotImplementedError ''
```

Empty messages, so the traceback does not say what *is* supported.

## Expected

Per site, one of: implement it, or stop advertising it (remove `trim`, narrow `clamp`'s annotation to `torch.Tensor`, drop `"fast"`). Removal is an API change → window; until then, corrective messages ("`clamp` accepts `(B, 2)` tensors; got tuple") are a focused fix welcome as a PR, and the batch-4 docstrings say plainly which paths raise.

## Classification

**Documented precondition never satisfiable / API honesty.** Idiom 5.

## Disposition

n/a — single symbol (three sites on one class).

## Environment

kornia 0.9.0rc1 @ `0685b1f5` (main), torch 2.9.1, python 3.11, cpu. Found during the bbox/boxes conventions audit (batch 4, tracking issue #4020); the batch documents the behavior below as it is and pins it, and does not change it.

[...]

PR #4138, opened 2026-08-31
docs(geometry): export and document VideoBoxes (#4016)

### LINKED ISSUE #4016: API completeness: `VideoBoxes` is public (used by `AugmentationSequential`) but not in `__all__`, absent from the docs, and undocumented
`kornia.geometry.boxes.VideoBoxes` is what `AugmentationSequential` wraps video boxes in (`kornia/augmentation/container/augment.py:633`, `container/video.py`), so users receive and construct it. But `boxes.py` exports `__all__ = ["Boxes", "Boxes3D"]`, the `automodule :members:` page therefore never renders it (checked on kornia.readthedocs.io/en/latest/geometry.boxes.html — `Boxes` and `Boxes3D` with all their methods are there, `VideoBoxes` is not), and none of its three methods has a docstring.

## Repro

```python
import kornia.geometry.boxes as m
from kornia.geometry.boxes import VideoBoxes
print(m.__all__, "VideoBoxes" in m.__all__)
print([(n, VideoBoxes.__dict__[n].__doc__ is not None) for n in ("from_tensor", "to_tensor", "clone")])
```

## Observed

```
['Boxes', 'Boxes3D'] False
[('from_tensor', False), ('to_tensor', False), ('clone', False)]
```

## Expected

A scope decision:

1. **Export it** — add to `__all__` (the API-surface guard `tests/test_api_surface.py` must be updated in the same PR) and document the three methods; it then appears on the docs page.
2. **Keep it internal** — document on `AugmentationSequential` that video boxes are an internal wrapper, and give the methods docstrings anyway for `help()` readers.

The batch-4 conventions PRs add the three docstrings (docs-only) regardless; the `__all__` change is the part that needs this decision first.

## Classification

**API completeness / discuss-first**, not a behavior defect. No idiom.

## Disposition

n/a — single symbol.

## Environment

kornia 0.9.0rc1 @ `0685b1f5` (main), torch 2.9.1, python 3.11, cpu. Found during the bbox/boxes conventions audit (batch 4, tracking issue #4020); the batch documents the behavior below as it is and pins it, and does not change it.

Soham-47:
## Summary

Closes #4016 (option A from the issue: export `VideoBoxes`).

- Add `VideoBoxes` to `kornia.geometry.boxes.__all__` so `automodule :members:` includes it on the docs page.
- Document the class and its `from_tensor` / `to_tensor` / `clone` overrides (matching nearby `Boxes` style).
- Add `TestVideoBoxes` for smoke, exception, cardinality, round-trip/clone, public-API pin, and gradcheck.

No runtime behavior change.

**Note on `tests/api_surface.json`:** that inventory tracks package-level modules (`kornia.geometry`, …), not `kornia.geometry.boxes`. Adding `VideoBoxes` to the boxes submodule `__all__` does not change the package inventory, so the JSON was left alone. The export is pinned in `TestVideoBoxes.test_public_api_surface` instead.

### Documented warts (not fixed in this PR)

Per review measurement on the `vertices_plus` path:

1. **`validate_boxes` is currently inert** for `VideoBoxes.from_tensor`. The flag is forwarded to `_boxes_to_quadrilaterals`, but that helper only validates in non-`vertices*` modes. Making the vertices path validate would also change `Boxes` and belongs in the coordinated repair window.
2. **`VideoBoxes.to_tensor` list branch is unreachable.** `from_tensor` rejects lists, and constructing `VideoBoxes(<list>, …)` never sets `temporal_channel_size`, so `to_tensor` raises `AttributeError` before any list return. Same family as the “advertised paths that always raise” class tracked in #4017 — worth adding there rather than papering over here.

## Test plan

- [x] `.venv/bin/python -m pytest tests/geometry/test_boxes.py::TestVideoBoxes -q --device=cpu --dtype=float32` (6 passed)
- [x] `.venv/bin/python -m pytest tests/test_api_surface.py tests/geometry/test_boxes.py -q --device=cpu --dtype=float32` (56 passed)
- [x] `pre-commit run --files kornia/geometry/boxes.py tests/geometry/test_boxes.py`
- [x] Issue #4016 repro now reports `VideoBoxes` in `__all__` and all three method docs present
- [x] Review blockers: docstring no longer claims `validate_boxes` rejects, or that `to_tensor` can return a list

### DIFF
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -837,10 +837,12 @@ class Boxes:
 class VideoBoxes(Boxes):
     r"""2D boxes with an explicit temporal channel for video sequences.
 
-    Stores box corners for a batch of videos as :math:`(B, T, N, 4, 2)` in
-    ``vertices_plus`` mode. :class:`~kornia.augmentation.AugmentationSequential`
-    uses this wrapper when the pipeline contains a video sequential so that
-    ``to_tensor`` restores the temporal axis after geometric transforms.
+    Accepts and returns box corners for a batch of videos as
+    :math:`(B, T, N, 4, 2)` in ``vertices_plus`` mode. Internally the corners
+    are stored flattened as :math:`(B \cdot T, N, 4, 2)`.
+    :class:`~kornia.augmentation.AugmentationSequential` uses this wrapper when
+    the pipeline contains a video sequential so that ``to_tensor`` restores the
+    temporal axis after geometric transforms.
 
     Attributes:
         temporal_channel_size: Number of frames :math:`T` stored with the boxes.
@@ -859,8 +861,9 @@ class VideoBoxes(Boxes):
             boxes: Box corners with shape :math:`(B, T, N, 4, 2)` in
                 ``vertices_plus`` order (top-left, top-right, bottom-right,
                 bottom-left). Lists of tensors are not supported yet.
-            validate_boxes: If ``True``, reject rectangles whose width or height
-                is smaller than the ``vertices_plus`` minimum (2).
+            validate_boxes: Forwarded to ``_boxes_to_quadrilaterals``. The
+                ``vertices_plus`` path used here builds corners directly and
+                performs no size check, so this flag currently has no effect.
 
         Returns:
             :class:`VideoBoxes` with :attr:`temporal_channel_size` set to
@@ -893,8 +896,7 @@ class VideoBoxes(Boxes):
 
         Returns:
             Tensor shaped :math:`(B, T, \ldots)` where :math:`T` is
-            :attr:`temporal_channel_size`, or a list of such tensors when the
-            container was built from a padded box list.
+            :attr:`temporal_channel_size`.
         """
         out = super().to_tensor(mode, as_padded_sequence=False)
         if isinstance(out, torch.Tensor):
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -25,7 +25,7 @@ from torch import Size
 from kornia.core.ops import eye_like
 from kornia.geometry.linalg import transform_points
 
-__all__ = ["Boxes", "Boxes3D"]
+__all__ = ["Boxes", "Boxes3D", "VideoBoxes"]
 
 
 def _is_floating_point_dtype(dtype: torch.dtype) -> bool:
@@ -835,12 +835,41 @@ class Boxes:
 
 
 class VideoBoxes(Boxes):
+    r"""2D boxes with an explicit temporal channel for video sequences.
+
+    Stores box corners for a batch of videos as :math:`(B, T, N, 4, 2)` in
+    ``vertices_plus`` mode. :class:`~kornia.augmentation.AugmentationSequential`
+    uses this wrapper when the pipeline contains a video sequential so that
+    ``to_tensor`` restores the temporal axis after geometric transforms.
+
+    Attributes:
+        temporal_channel_size: Number of frames :math:`T` stored with the boxes.
+
+    """
+
     temporal_channel_size: int
 
     @classmethod
     def from_tensor(  # type: ignore[override]
         cls, boxes: torch.Tensor | list[torch.Tensor], validate_boxes: bool = True
     ) -> VideoBoxes:
+        r"""Create :class:`VideoBoxes` from a video box tensor.
+
+        Args:
+            boxes: Box corners with shape :math:`(B, T, N, 4, 2)` in
+                ``vertices_plus`` order (top-left, top-right, bottom-right,
+                bottom-left). Lists of tensors are not supported yet.
+            validate_boxes: If ``True``, reject rectangles whose width or height
+                is smaller than the ``vertices_plus`` minimum (2).
+
+        Returns:
+            :class:`VideoBoxes` with :attr:`temporal_channel_size` set to
+            ``boxes.size(1)``.
+
+        Raises:
+            ValueError: If ``boxes`` is a list or does not have shape
+                :math:`(B, T, N, 4, 2)`.
+        """
         if isinstance(boxes, (list,)) or (boxes.dim() != 5 or boxes.shape[-2:] != torch.Size([4, 2])):
             raise ValueError("Input box type is not yet supported. Please input an `BxTxNx4x2` tensor directly.")
 
@@ -856,6 +885,17 @@ class VideoBoxes(Boxes):
         return out
 
     def to_tensor(self, mode: Optional[str] = None) -> torch.Tensor | list[torch.Tensor]:  # type: ignore[override]
+        r"""Cast :class:`VideoBoxes` to a tensor with the temporal axis restored.
+
+        Args:
+            mode: Output box format forwarded to :meth:`Boxes.to_tensor`. When
+                ``None``, uses the stored mode (``vertices_plus`` by default).
+
+        Returns:
+            Tensor shaped :math:`(B, T, \ldots)` where :math:`T` is
+            :attr:`temporal_channel_size`, or a list of such tensors when the
+            container was built from a padded box list.
+        """
         out = super().to_tensor(mode, as_padded_sequence=False)
         if isinstance(out, torch.Tensor):
             return out.view(-1, self.temporal_channel_size, *out.shape[1:])
@@ -863,6 +903,12 @@ class VideoBoxes(Boxes):
         return [_out.view(-1, self.temporal_channel_size, *_out.shape[1:]) for _out in out]
 
     def clone(self) -> VideoBoxes:
+        """Create an independent copy of the video box container.
+
+        Returns:
+            New :class:`VideoBoxes` with cloned tensor storage and the same
+            :attr:`temporal_channel_size`, mode, and batch metadata.
+        """
         obj = type(self)(self._data.clone(), False)
         obj._mode = self._mode
         obj._N = self._N
--- a/tests/geometry/test_boxes.py
+++ b/tests/geometry/test_boxes.py
@@ -20,7 +20,8 @@ from functools import partial
 import pytest
 import torch
 
-from kornia.geometry.boxes import Boxes, Boxes3D
+from kornia.geometry import boxes as boxes_module
+from kornia.geometry.boxes import Boxes, Boxes3D, VideoBoxes
 
 from testing.base import BaseTester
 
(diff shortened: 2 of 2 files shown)


Soham-47:
## CI note

The red checks are **not from the VideoBoxes change**.

Failing job: `tests (windows-latest, float32, 3.11, 2.9.1)` — all failures are in `tests/io/test_io_image.py` (`kornia_rs` PNG/JPEG decode: `Invalid PNG signature` / JPEG compressor errors). Same matrix cells on other Windows / Ubuntu / macOS jobs are green.

`TestVideoBoxes` passed in that job. No geometry/boxes regressions.

Suggested next step: **Re-run failed jobs** on the Actions run (likely flaky Windows IO fixture / decoder). Happy to empty-commit bounce CI if preferred.

ducha-aiki:
## Review of #4138 — the export and the docs wiring are right; two of the new docstring claims are not

Setup: worktree at the PR head `9300c634`, with `kornia/` swapped to the merge-base `1331a861` for the fail-on-base and before/after passes. `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS; `kornia.__file__` confirmed to resolve to the worktree in every block.

**The mechanism works and I measured it rather than reasoning about autodoc.** I built the real page with Sphinx 8.2.3 at both revisions:

```
HEAD  VideoBoxes documented: 1 | Boxes: 1 | temporal_channel_size: 4
BASE  VideoBoxes documented: 0 | Boxes: 1 | temporal_channel_size: 0
```

So `automodule :members:` does pick it up from `__all__`, exactly as your body claims, and `docs/source/geometry.boxes.rst` needs no edit.

**Your `tests/api_surface.json` reasoning is also correct**, and I checked it with the test's own `_current_surface` helper rather than by eye: `kornia.geometry.boxes` is not one of the 12 inventory modules, `kornia/geometry/__init__.py` does not star-import from `boxes`, and `VideoBoxes` is in neither `dir(kornia.geometry)` nor `dir(kornia)` before or after. Nothing to regenerate. For the record, `Boxes`, `Boxes3D`, `Keypoints` and `VideoKeypoints` are all in the same state, so this PR introduces no asymmetry.

Tests: **81 passed** across float32 and float64 at the head; on the merge-base `TestVideoBoxes::test_public_api_surface` fails (`assert 'VideoBoxes' in ['Boxes', 'Boxes3D']`) with the other 5 passing — the right discriminator for an export-and-document change.

### Blocker 1 — `validate_boxes` does nothing here, but the docstring says it rejects boxes

You wrote:

> `validate_boxes`: If ``True``, reject rectangles whose width or height is smaller than the ``vertices_plus`` minimum (2).

`VideoBoxes.from_tensor` always calls `_boxes_to_quadrilaterals(..., mode="vertices_plus")`, and the validation in that function (`boxes.py:139-143`, `width <= 0` / `height <= 0`) lives in the **non-vertices** branch. The `mode.startswith("vertices")` path builds the quadrilaterals directly and never validates. Measured, six cases, flag on versus off:

```
case                     validate=True   validate=False   differ?
normal 4x3               accepted       accepted       no
extent 1x1 (w=h=0)       accepted       accepted       no
extent 2x2               accepted       accepted       no
inverted (x max<min)     accepted       accepted       no
sub-unit 0.5             accepted       accepted       no
negative extent          accepted       accepted       no

For comparison, Boxes.from_tensor(mode='xyxy') DOES validate:
  validate=True: REJECTED (Some boxes have negative widths or 0.)
  validate=False: accepted
```

The flag is inert for `VideoBoxes` — a degenerate box, an inverted box and a negative-extent box all sail through with `validate_boxes=True`. Documenting it as a rejection is worse than leaving it undocumented, because a caller will rely on it.

I am **not** asking you to make it validate. Adding validation to the vertices branch changes what `Boxes` accepts too, so it is a semantic change and belongs in the coordinated repair window, not in a docs PR. The ask is to describe what the parameter actually does — something like "forwarded to `_boxes_to_quadrilaterals`; note that the `vertices_plus` path performs no validation, so this flag currently has no effect" — and, if you are willing, to say so in the PR body so the wart is on the record rather than papered over.

### Blocker 2 — the documented list return cannot be produced

You wrote, on `to_tensor`:

> Returns: Tensor shaped :math:`(B, T, \ldots)` ... **or a list of such tensors when the container was built from a padded box list.**

That return can never happen. The list branch needs `self._N is not None`, and `_N` is assigned in exactly one place (`boxes.py:215`, `boxes, self._N = _merge_box_list(boxes)`), reached only when the **constructor** is handed a list. `VideoBoxes.from_tensor` rejects lists outright. Taking the only remaining route:

```python
vb = VideoBoxes([q1, q2], False, "vertices_plus")
# constructed from list -> _N = [1, 0]
# has temporal_channel_size: False
vb.to_tensor()
# AttributeError: 'VideoBoxes' object has no attribute 'temporal_channel_size'
```

`temporal_channel_size` is only ever set inside `from_tensor`, so any `VideoBoxes` that could reach the list branch raises before returning. The branch is dead, and the sentence documents a capability that does not exist.

Please drop that clause and state that the return is always a `Tensor`. Two adjacent things worth flagging in the body rather than fixing here:

- the `# If returns a list of boxes.` branch in `VideoBoxes.to_tensor` is unreachable dead code;
- `VideoBoxes(<list>)` constructs an object whose `to_tensor` always raises `AttributeError`. That is a new instance of the class **#4017** already tracks ("`Boxes` advertises paths that always raise"), and it would be worth adding there.

### Smaller notes, non-blocking

- The class docstring says "Stores box corners ... as :math:`(B, T, N, 4, 2)`". Internally `_data` is flattened to :math:`(B \cdot T, N, 4, 2)`; the :math:`(B, T, N, 4, 2)` shape is what `from_tensor` accepts and `to_tensor` returns. Worth one word ("accepts and returns") so nobody reads it as the storage layout — `test_cardinality` already pins the real one.
- Everything else I checked is accurate: the `Returns` clause on `from_tensor` (`temporal_channel_size == boxes.size(1)`, measured 3 for `T=3`), the `Raises` clause, and the whole `clone` docstring — `clone` does copy `temporal_channel_size`, and a cloned container round-trips to `(2, 3, 1, 4, 2)` correctly.
- `VideoKeypoints` in `kornia/geometry/keypoints.py` is the identical case and is listed in #4026. Out of scope here; the obvious follow-up if you want it.

### Verdict

**REQUEST CHANGES** on the two falsified docstring claims. This is a documentation PR, so the prose is the deliverable, and both statements would mislead a reader about behavior I have measured. Everything else — the `__all__` addition, the docs render, the api-surface reasoning, the tests and their fail-on-base — is verified and good.

Soham-47:
Addressed the REQUEST CHANGES docstring blockers in the latest commit:

1. **`validate_boxes`** — now documents that the flag is forwarded but currently has no effect on the `vertices_plus` path (no semantic validation change).
2. **`to_tensor` Returns** — dropped the list clause; documents a Tensor only.
3. Class docstring — clarifies accept/return `(B, T, N, 4, 2)` vs flattened internal storage.

Also recorded the inert-flag / unreachable-list / `VideoBoxes(<list>)` AttributeError warts in the PR body (and pointed at #4017 for the advertised-paths-that-raise class). Not fixing those behaviors in this docs/export PR.

ducha-aiki:
## Re-review of #4138 at `a28425e9` — APPROVE

Both blockers are fixed. I re-read the whole production diff rather than only the delta, and re-executed each **new** claim rather than reading it.

### The three corrections check out

**1. `validate_boxes`** now reads *"Forwarded to `_boxes_to_quadrilaterals`. The `vertices_plus` path used here builds corners directly and performs no size check, so this flag currently has no effect."* Verified on both halves — `_boxes_to_quadrilaterals(` really is the callee in `from_tensor`, and the flag really is inert:

```
extent 1x1 (w=h=0)     validate=True:accepted  validate=False:accepted  differ=False
inverted               validate=True:accepted  validate=False:accepted  differ=False
sub-unit 0.5           validate=True:accepted  validate=False:accepted  differ=False
negative extent        validate=True:accepted  validate=False:accepted  differ=False
```

**2. `to_tensor` Returns** — the list clause is gone, and the remaining claim holds:

```
VideoBoxes.from_tensor((2,3,5,4,2)).to_tensor()  ->  Tensor (2, 3, 5, 4, 2)   round-trips: True
```

**3. Class docstring** — *"Accepts and returns ... Internally the corners are stored flattened as :math:`(B \cdot T, N, 4, 2)`"*:

```
input (B,T,N,4,2)=(2, 3, 5, 4, 2)  ->  _data=(6, 5, 4, 2)   expected=(6, 5, 4, 2)  match=True
```

### The wart record is accurate, and correctly scoped

The two entries you added to the body are right on the substance and on the disposition. Point 1 in particular gets the reason right — the helper "only validates in non-`vertices*` modes", so making the vertices path validate would change what `Boxes` accepts too, which is repair-window material and not something a docs PR should do. Point 2 correctly identifies the unreachable branch and the `VideoBoxes(<list>)` `AttributeError` as the #4017 family. That is the right outcome: the behaviour is unchanged, and the defect is on the record instead of being described as a feature.

### Nothing regressed

```
head, tests/geometry/test_boxes.py, float32 + float64:   81 passed
fail-on-base (kornia/ at 1331a861):                      1 failed, 5 passed
  -> TestVideoBoxes::test_public_api_surface  assert 'VideoBoxes' in ['Boxes', 'Boxes3D']
CI:                                                      34/34 green
```

The docs still build, and the new prose adds no warnings — I built the page at both revisions and diffed the diagnostics:

```
BASE: 4 docutils warnings/errors | VideoBoxes rendered: 0
HEAD: 4 docutils warnings/errors | VideoBoxes rendered: 1
  both:  boxes.Boxes:9   boxes.Boxes.to_tensor:19   boxes.Boxes3D:6
```

Identical warning set, all three pre-existing and all in `Boxes`/`Boxes3D`; none in `VideoBoxes`. The earlier Windows `kornia_rs` red is gone too — your flake diagnosis was correct.

### One residual, explicitly not an ask

`to_tensor`'s signature still annotates `torch.Tensor | list[torch.Tensor]` while the docstring is now Tensor-only. That mismatch is the dead branch showing through, and narrowing the annotation would be legal (return types are covariant, so a narrower override still satisfies `Boxes.to_tensor`) — but it is a code change, so it belongs with whoever takes #4017, not here. Mentioning it only so the next reader knows it was noticed rather than missed.

### Verdict

**APPROVE.** The docstrings now say what the code does, the two behaviours that could not be honestly documented are recorded as warts with the right issue attached, and the export, render and tests are all unchanged from the pass I already verified.

ducha-aiki (approved):

Issue #4032, opened 2026-08-26
warp_affine with dsize=(0, w) crashes on MPS with an internal grid_sample assertion


ducha-aiki:
## Description

`warp_affine(src, M, dsize=(h, 0))` (or `(0, w)`) crashes on MPS with an internal PyTorch assertion — for a fully valid call: a non-empty, floating-point `src`, a valid `M`, and a `dsize` with only a zero destination dimension. CPU is fine for the identical call.

## Root cause

`_empty_warp_output_2d` (`kornia/geometry/transform/imgwarp.py`) only substitutes a connected 1x1 stand-in for `src` when **`src` itself** has a zero spatial dimension:

```python
sample_src = src
if src.shape[-2] == 0 or src.shape[-1] == 0:
    src_zero = src.reshape(-1)[:1].sum() * 0.0
    sample_src = src_zero.reshape(1, 1, 1, 1).expand(src.shape[0], src.shape[1], 1, 1)
...
return F.grid_sample(sample_src, grid, align_corners=align_corners, mode=mode, padding_mode=padding_mode)
```

When only the *destination* `dsize` is zero, `sample_src` stays the real (non-empty) `src`, and it's the sampling `grid` that has the zero-sized dimension. On MPS, `F.grid_sample` cannot handle a zero-sized grid dimension against a non-empty input source and hits an internal assertion. This reproduces with a bare `F.grid_sample` call, so the proximate defect is in PyTorch's MPS backend, but kornia's empty-destination fast path currently routes every `dsize=(h, 0)`/`(0, w)` call through `F.grid_sample` unconditionally on MPS instead of only when `src` needs the 1x1 stand-in.

## Minimal repro

```python
import torch
from kornia.geometry.transform import warp_affine

src = torch.rand(1, 1, 4, 4, device="mps")
M = torch.eye(2, 3, device="mps")[None]

warp_affine(src, M, (0, 4))
```

Exact error:

```
RuntimeError: [srcBuf length] > 0 INTERNAL ASSERT FAILED at "/Users/runner/work/pytorch/pytorch/pytorch/aten/src/ATen/native/mps/OperationUtils.mm":551, please report a bug to PyTorch. Placeholder tensor is empty!
```

Isolated with a bare `F.grid_sample` call (no kornia code involved), confirming the proximate cause is PyTorch/MPS:

```python
import torch, torch.nn.functional as F
[...]

PR #4140, opened 2026-09-01
fix(geometry): return an empty image for a zero-sized resize output

### LINKED ISSUE #4115: resize/rescale raise ZeroDivisionError for a zero output size, while warp_affine returns an empty image
`resize`, `rescale`, `Resize` and `Rescale` raise a bare `ZeroDivisionError` when the requested output size has a zero dimension, while every other kornia op that takes a destination size returns a correspondingly empty image.

This is the same defect class as #4056 (`ZeroDivisionError` where a shape error belongs), but on a much wider public surface, and here it also puts `resize` in direct conflict with `warp_affine` on the *same argument*.

## Repro

```python
import torch, kornia

img = torch.rand(1, 3, 4, 4)

kornia.geometry.resize(img, (0, 4))                                    # ZeroDivisionError
kornia.geometry.rescale(img, 0.0)                                      # ZeroDivisionError
kornia.geometry.transform.Resize((0, 4))(img)                          # ZeroDivisionError
kornia.geometry.transform.Rescale(0.0)(img)                            # ZeroDivisionError

kornia.geometry.warp_affine(img, torch.eye(2, 3)[None], (0, 4))        # -> (1, 3, 0, 4)
kornia.geometry.warp_perspective(img, torch.eye(3)[None], (0, 4))      # -> (1, 3, 0, 4)
kornia.geometry.center_crop(img, (0, 4))                               # -> (1, 3, 0, 4)
```

## Observed

```text
resize(img, (0, 4))          ZeroDivisionError: division by zero
resize(img, (4, 0))          ZeroDivisionError: division by zero
resize(img, (0, 0))          ZeroDivisionError: division by zero
resize(img, 0)               ZeroDivisionError: division by zero     # int/`side` form
resize(img, (0, 4), antialias=True)   ZeroDivisionError              # antialias is irrelevant
rescale(img, 0.0)            ZeroDivisionError: division by zero
warp_affine(img, M, (0, 4))  torch.Size([1, 3, 0, 4])
center_crop(img, (0, 4))     torch.Size([1, 3, 0, 4])
```

`ZeroDivisionError` derives from `ArithmeticError`. It is neither a `RuntimeError` nor a kornia `BaseError`, so a caller that wraps a resize to validate a user-supplied or computed target size catches nothing useful — the same reasoning as #4056 and #4050.

## Cause

`kornia/geometry/transform/affwarp.py:655`, in `resize`:

```python
factors = (h / size[0], w / size[1])
antialias = antialias and (max(factors) > 1)
```

`factors` is computed unconditionally even though its only consumers are the antialias decision on the next line and the `sigmas` inside `if antialias:`. With `antialias=False` — the default — the value is discarded, but the division has already raised. `rescale` and both `nn.Module` wrappers delegate to `resize`, so they inherit it.

## Expected

Two defensible answers, and I do not think this should be decided in a drive-by:

**Option 1 — surface torch's own error (minimal).** Compute `factors` lazily, inside the `antialias` branch. The failure then comes from the line below it:

```text
RuntimeError: Input and output sizes should be greater than 0, but got input (H: 4, W: 4)
```

Two lines moved, a typed error instead of an arithmetic one, and no behavior change for any non-degenerate size.

**Option 2 — return an empty image, matching the warp family.** This is what the rest of the library does. `kornia/geometry/transform/imgwarp.py` carries a dedicated `_empty_warp_output_2d` whose docstring is explicit — *"Return an empty warp while retaining normal grid-sample validation and autograd links"* — and whose comments state the goal as making a zero-sized `dsize` *"neither stricter nor laxer than a non-empty one"*. It is pinned by 13 tests across `test_imgwarp.py` / `test_imgwarp3d.py`, and #4031/#4076 ratified the parity rule for exactly these paths. `create_meshgrid` documents the same convention in prose: *"A zero spatial size produces a correspondingly empty grid."*

The obstacle to option 2 is inherited, and worth recording: the two torch primitives disagree.

```python
F.grid_sample(img, torch.zeros(1, 0, 4, 2), align_corners=True)   # -> (1, 3, 0, 4)
F.interpolate(img, size=(0, 4), mode="bilinear")                  # RuntimeError
```

The warp family is built on `grid_sample`, which produces empty output happily; `resize` is built on `interpolate`, which refuses. So option 2 needs an explicit empty fast path like the warp one, not a reordering.

Option 1 is a strict improvement either way and does not foreclose option 2.

## Related

- #4056 — the same `ZeroDivisionError`-for-a-shape-problem class in `yuv420_to_rgb` / `yuv422_to_rgb`. Found while surveying the convention to answer that issue's open steering question.
- #4050, #4031 — prior members of the "degenerate path reports the wrong exception type" family.
- Milder sibling, not filed separately: `kornia.enhance.normalize` on a zero-**batch** input raises `RuntimeError: cannot reshape tensor of 0 elements into shape [0, 3, -1]`, while `RandomHorizontalFlip` and every colour conversion handle `B=0` fine.

## Environment

kornia 0.9.0rc1 @ `47a91725`, torch 2.9.1, Python 3.11.14, macOS, CPU (MPS available). All measurements from a clean worktree at that revision.

hiyufan:

`resize`, `rescale`, `Resize` and `Rescale` raised a bare `ZeroDivisionError` when the requested output size had a zero side, while `warp_affine`, `warp_perspective` and `center_crop` return a correspondingly empty image for that same argument.

The zero-sized output is now short-circuited and the empty result built directly. It cannot come out of the normal path: the antialias scale factors divide by the requested side, and `torch.nn.functional.interpolate` rejects a zero-sized output outright with `RuntimeError: Input and output sizes should be greater than 0`.

The same function had a second gap on the other side of the argument. A zero-sized *input* leaked a `ZeroDivisionError` out of `aspect_ratio = w / h`, or a `RuntimeError` out of `interpolate`, where `warp_affine` and `center_crop` raise `ValueError: Input image size must be positive.`. `resize` now raises that same error up front. This half is not in the report, so it is easy to drop if you would rather keep the PR to exactly the reported case — it is the `h <= 0 or w <= 0` guard and `TestResize::test_zero_input_size`.

`rescale`, `Resize` and `Rescale` all delegate to `resize`, so the one change covers the six call paths in the issue.

This follows the convention @ducha-aiki stated in #4056 — *empty in → empty out, with validation parity* — so a zero output side returns an empty image rather than raising, and the input side is validated rather than left to a downstream error.

Fixes #4115

## What did you check?

Behaviour on `torch.rand(1, 3, 4, 4)`, measured before and after:

| call | before | after | reference op, same argument |
|---|---|---|---|
| `resize(img, (0, 4))` | `ZeroDivisionError` | `(1, 3, 0, 4)` | `warp_affine(img, eye, (0, 4))` → `(1, 3, 0, 4)` |
| `resize(img, (4, 0))` | `ZeroDivisionError` | `(1, 3, 4, 0)` | |
| `resize(img, (0, 0))` | `ZeroDivisionError` | `(1, 3, 0, 0)` | |
| `resize(img, 0)` | `ZeroDivisionError` | `(1, 3, 0, 0)` | |
| `rescale(img, 0.0)` | `ZeroDivisionError` | `(1, 3, 0, 0)` | |
| `Resize((0, 4))(img)` | `ZeroDivisionError` | `(1, 3, 0, 4)` | |
| `Rescale(0.0)(img)` | `ZeroDivisionError` | `(1, 3, 0, 0)` | |
| `resize(empty_hw, (2, 2))` | `RuntimeError` from `interpolate` | `ValueError: Input image size must be positive.` | `warp_affine` raises the same |
| `resize(empty_hw, 2)` | `ZeroDivisionError` | `ValueError: Input image size must be positive.` | `center_crop` raises the same |

Leading dimensions are preserved: `(4, 4)` → `(0, 4)`, `(3, 4, 4)` → `(3, 0, 4)`, `(2, 1, 3, 4, 4)` → `(2, 1, 3, 0, 4)`.

Valid inputs are untouched. A/B sweep of `resize` and `rescale` over 4 input shapes × float32/float64 × 6 output sizes × antialias on/off × 4 interpolation modes, plus the integer-`size` form over all four `side` values — comparing output shape and sum against this branch's base revision:

```text
combinations (float32 + float64): 512   differing: 0
```

Both halves of that sweep ran from the repository root as stdin and printed `kornia.__file__` (`/root/kornia/kornia/__init__.py`) to confirm each half measured the tree it was meant to.

```text
$ pytest tests/geometry/transform/test_affine.py --dtype=float32,float64
97 passed, 10 skipped

$ pytest tests/geometry/transform/
520 passed, 15 skipped, 1 xfailed

$ ruff check kornia/geometry/transform/affwarp.py tests/geometry/transform/test_affine.py
All checks passed!
$ ruff format --check <same files>
2 files already formatted
```

New tests: `TestResize::test_zero_output_size`, `TestResize::test_zero_input_size`, `TestRescale::test_zero_factor`. The first asserts against `warp_affine` on the same argument, so the two stay pinned together.

`grep -rnE "#4115|issues/4115" kornia/ tests/` returns nothing, so there is no inline defect note to retire with this change.

I did not run `pixi run pre-commit-all` or `pixi run typecheck` — I worked in a plain venv rather than a Pixi environment, so those two are worth a look in CI.

## How did AI help?

AI (Claude Code) wrote the patch and the tests, ran the A/B sweep, and drafted this description. I reviewed every line, and the measurements above are from runs on my own machine. The one judgement call I would flag for you rather than claim as settled is the zero-sized-input half described above, since the issue only asks for the output half.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -569,6 +569,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   The export resolver includes implicit submodule bindings and rejects unsupported dynamic binding expressions
   as evidence of removal. Deprecation windows and release notes still apply. (#4190, #4230)
 
+* `resize` and `rescale` return an empty image for a zero-sized output instead of raising
+  `ZeroDivisionError` out of the scale-factor computation (#4115). `resize(x, (0, 4))`, `resize(x, 0)`,
+  `Resize((0, 4))`, and any `rescale` factor that takes a side to zero (`rescale(x, 0.0)`) previously died
+  in `h / size[0]`; they now return an empty tensor of the requested shape, which is the answer
+  `warp_affine`, `warp_perspective` and `center_crop` already gave for the same request. The empty result
+  stays connected to the autograd graph, the way `_empty_warp_output_2d` builds it for the warping ops, so
+  a batch element that degenerates to a zero-sized output still contributes a zero gradient rather than
+  detaching.
+
+  A degenerate *input* is a different question and is now rejected rather than divided by: resizing a
+  `(1, 3, 0, 4)` image to a positive size, or with an `int` size at all -- an `int` names one side and takes
+  the other from an aspect ratio a degenerate input does not have -- raises
+  `ValueError("Input image size must be positive. Got height=0, width=4.")`, the message the warping ops
+  already use, instead of the same bare `ZeroDivisionError`. Empty in and empty out is well defined and
+  still comes back empty. (#4140)
+
 * `validate_bbox` and `validate_bbox3d` flatten rank-4 `(B, N, 4, 2)` / `(B, N, 8, 3)` input with `reshape`
   instead of `view`, so a non-contiguous leading-dimension stride (a transpose, a slice that drops boxes, an
   `expand`) returns a boolean as documented instead of raising `RuntimeError`. (#4174)
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -636,12 +636,13 @@ def resize(
 
     original_shape = input.shape
     h, w = input.shape[-2:]
-    if h <= 0 or w <= 0:
-        # Reject a degenerate input the way the warping ops do, rather than letting the
-        # aspect-ratio division below fail with a bare ZeroDivisionError.
-        raise ValueError(f"Input image size must be positive. Got height={h}, width={w}.")
 
     if isinstance(size, int):
+        if h <= 0 or w <= 0:
+            # An int size names one side and takes the other from the aspect ratio, which a
+            # degenerate input does not have. Reject it with the message the warping ops use,
+            # rather than letting the division below fail with a bare ZeroDivisionError.
+            raise ValueError(f"Input image size must be positive. Got height={h}, width={w}.")
         aspect_ratio = w / h
         size = _side_to_image_size(size, aspect_ratio, side)
     if len(original_shape) == 2:
@@ -659,10 +660,18 @@ def resize(
 
     if size[0] == 0 or size[1] == 0:
         # An output side of zero gives an empty image, matching warp_affine,
-        # warp_perspective and center_crop. The result is built directly: the scale
-        # factors below would divide by zero, and torch.nn.functional.interpolate
-        # rejects a zero-sized output outright.
-        output = input.new_empty((*input.shape[:-2], size[0], size[1]))
+        # warp_perspective and center_crop -- including when the input is empty as well,
+        # which is why this runs ahead of the positive-size check below. The result is
+        # built directly: the scale factors would divide by zero, and
+        # torch.nn.functional.interpolate rejects a zero-sized output outright.
+        #
+        # It is routed through the input so the empty result keeps its autograd link, the
+        # way _empty_warp_output_2d does for the warping ops. sum() promotes an integer
+        # input to int64, so the accumulation dtype is pinned to the input's.
+        zero = input.reshape(-1)[:1].sum(dtype=input.dtype) * 0
+        output = zero.reshape(*([1] * input.dim())).expand(*input.shape[:-2], size[0], size[1])
+    elif h <= 0 or w <= 0:
+        raise ValueError(f"Input image size must be positive. Got height={h}, width={w}.")
     else:
         factors = (h / size[0], w / size[1])
         antialias = antialias and (max(factors) > 1)
--- a/kornia/geometry/transform/affwarp.py
+++ b/kornia/geometry/transform/affwarp.py
@@ -636,6 +636,11 @@ def resize(
 
     original_shape = input.shape
     h, w = input.shape[-2:]
+    if h <= 0 or w <= 0:
+        # Reject a degenerate input the way the warping ops do, rather than letting the
+        # aspect-ratio division below fail with a bare ZeroDivisionError.
+        raise ValueError(f"Input image size must be positive. Got height={h}, width={w}.")
+
     if isinstance(size, int):
         aspect_ratio = w / h
         size = _side_to_image_size(size, aspect_ratio, side)
@@ -652,19 +657,26 @@ def resize(
             batch_size *= d
         input = input.reshape(batch_size, *original_shape[-3:])
 
-    factors = (h / size[0], w / size[1])
-    antialias = antialias and (max(factors) > 1)
+    if size[0] == 0 or size[1] == 0:
+        # An output side of zero gives an empty image, matching warp_affine,
+        # warp_perspective and center_crop. The result is built directly: the scale
+        # factors below would divide by zero, and torch.nn.functional.interpolate
+        # rejects a zero-sized output outright.
+        output = input.new_empty((*input.shape[:-2], size[0], size[1]))
+    else:
+        factors = (h / size[0], w / size[1])
+        antialias = antialias and (max(factors) > 1)
 
-    if antialias:
-        sigmas = (max((factors[0] - 1.0) / 2.0, 0.001), max((factors[1] - 1.0) / 2.0, 0.001))
+        if antialias:
+            sigmas = (max((factors[0] - 1.0) / 2.0, 0.001), max((factors[1] - 1.0) / 2.0, 0.001))
 
-        ks = int(max(2.0 * 2 * sigmas[0], 3)), int(max(2.0 * 2 * sigmas[1], 3))
+            ks = int(max(2.0 * 2 * sigmas[0], 3)), int(max(2.0 * 2 * sigmas[1], 3))
 
-        ks = (ks[0] if ks[0] % 2 else ks[0] + 1, ks[1] if ks[1] % 2 else ks[1] + 1)
+            ks = (ks[0] if ks[0] % 2 else ks[0] + 1, ks[1] if ks[1] % 2 else ks[1] + 1)
 
-        input = gaussian_blur2d(input, ks, sigmas)
+            input = gaussian_blur2d(input, ks, sigmas)
 
-    output = torch.nn.functional.interpolate(input, size=size, mode=interpolation, align_corners=align_corners)
+        output = torch.nn.functional.interpolate(input, size=size, mode=interpolation, align_corners=align_corners)
 
     if len(original_shape) == 2:
         output = output[0, 0]
--- a/tests/geometry/transform/test_affine.py
+++ b/tests/geometry/transform/test_affine.py
@@ -84,6 +84,44 @@ class TestResize(BaseTester):
         out = kornia.geometry.transform.resize(inp, (3, 1), align_corners=False)
         assert out.shape == (1, 2, 3, 2, 1, 3, 3, 1)
 
+    def test_zero_output_size(self, device, dtype):
+        # A zero output side gives an empty image, matching warp_affine,
+        # warp_perspective and center_crop on the same argument.
+        inp = torch.rand(1, 3, 4, 4, device=device, dtype=dtype)
+
+        assert kornia.geometry.transform.resize(inp, (0, 4)).shape == (1, 3, 0, 4)
+        assert kornia.geometry.transform.resize(inp, (4, 0)).shape == (1, 3, 4, 0)
+        assert kornia.geometry.transform.resize(inp, (0, 0)).shape == (1, 3, 0, 0)
+        assert kornia.geometry.transform.resize(inp, 0).shape == (1, 3, 0, 0)
+        assert kornia.geometry.transform.Resize((0, 4))(inp).shape == (1, 3, 0, 4)
+        assert kornia.geometry.transform.resize(inp, (0, 4), antialias=True).shape == (1, 3, 0, 4)
+
+        eye = torch.eye(2, 3, device=device, dtype=dtype)[None]
+        assert kornia.geometry.transform.warp_affine(inp, eye, (0, 4)).shape == (1, 3, 0, 4)
+
+        # 2D
+        inp_2d = torch.rand(4, 4, device=device, dtype=dtype)
+        assert kornia.geometry.transform.resize(inp_2d, (0, 4)).shape == (0, 4)
+
+        # 3D
+        inp_3d = torch.rand(3, 4, 4, device=device, dtype=dtype)
+        assert kornia.geometry.transform.resize(inp_3d, (0, 4)).shape == (3, 0, 4)
+
+        # arbitrary dim
+        inp_nd = torch.rand(2, 1, 3, 4, 4, device=device, dtype=dtype)
+        assert kornia.geometry.transform.resize(inp_nd, (0, 4)).shape == (2, 1, 3, 0, 4)
+
+    def test_zero_input_size(self, device, dtype):
+        # A degenerate input is rejected with the message the warping ops already
+        # use, instead of a bare ZeroDivisionError out of the aspect-ratio division.
+        inp = torch.rand(1, 3, 0, 4, device=device, dtype=dtype)
+
+        with pytest.raises(ValueError, match="Input image size must be positive"):
+            kornia.geometry.transform.resize(inp, (2, 2))
+
+        with pytest.raises(ValueError, match="Input image size must be positive"):
+            kornia.geometry.transform.resize(inp, 2)
+
     def test_downsizeAA(self, device, dtype):
         inp = torch.rand(1, 3, 10, 8, device=device, dtype=dtype)
         out = kornia.geometry.transform.resize(inp, (5, 3), align_corners=False, antialias=True)
@@ -256,6 +294,15 @@ class TestRescale(BaseTester):
         input = torch.rand(1, 3, 9, 8, device=device, dtype=dtype)
         output = kornia.geometry.transform.rescale(input, (1.0 / 3.0, 1.0 / 2.0), align_corners=False)
         assert output.shape == (1, 3, 3, 4)
+
+    def test_zero_factor(self, device, dtype):
+        # A zero factor asks for a zero-sized output, which gives an empty image
+        # rather than a ZeroDivisionError.
+        inp = torch.rand(1, 3, 4, 4, device=device, dtype=dtype)
+
+        assert kornia.geometry.transform.rescale(inp, 0.0).shape == (1, 3, 0, 0)
+        assert kornia.geometry.transform.rescale(inp, (0.0, 1.0)).shape == (1, 3, 0, 4)
+        assert kornia.geometry.transform.Rescale(0.0)(inp).shape == (1, 3, 0, 0)
 
     def test_downscale_values(self, device, dtype):
         inp_x = torch.arange(20, device=device, dtype=dtype) / 20.0
--- a/tests/geometry/transform/test_affine.py
+++ b/tests/geometry/transform/test_affine.py
@@ -99,6 +99,16 @@ class TestResize(BaseTester):
         eye = torch.eye(2, 3, device=device, dtype=dtype)[None]
         assert kornia.geometry.transform.warp_affine(inp, eye, (0, 4)).shape == (1, 3, 0, 4)
 
+        # The empty result stays attached to the graph, as it does for the warping ops:
+        # _empty_warp_output_2d builds a connected tensor precisely so a batch that
+        # degenerates to a zero-sized output still contributes a zero gradient.
+        grad_inp = torch.rand(1, 3, 4, 4, device=device, dtype=dtype, requires_grad=True)
+        empty = kornia.geometry.transform.resize(grad_inp, (0, 4))
+        assert empty.requires_grad
+        assert empty.dtype == grad_inp.dtype
+        empty.sum().backward()
+        assert grad_inp.grad is not None
+
         # 2D
         inp_2d = torch.rand(4, 4, device=device, dtype=dtype)
         assert kornia.geometry.transform.resize(inp_2d, (0, 4)).shape == (0, 4)
@@ -121,6 +131,15 @@ class TestResize(BaseTester):
 
         with pytest.raises(ValueError, match="Input image size must be positive"):
             kornia.geometry.transform.resize(inp, 2)
+
+        # Empty in and empty out is not undefined, so it comes back empty rather than
+        # raising -- the same answer warp_affine and center_crop give.
+        assert kornia.geometry.transform.resize(inp, (0, 4)).shape == (1, 3, 0, 4)
+        assert kornia.geometry.transform.resize(inp, (0, 2)).shape == (1, 3, 0, 2)
+
+        eye = torch.eye(2, 3, device=device, dtype=dtype)[None]
+        assert kornia.geometry.transform.warp_affine(inp, eye, (0, 4)).shape == (1, 3, 0, 4)
+        assert kornia.geometry.transform.center_crop(inp, (0, 4)).shape == (1, 3, 0, 4)
 
     def test_downsizeAA(self, device, dtype):
         inp = torch.rand(1, 3, 10, 8, device=device, dtype=dtype)


ducha-aiki:
## Review of #4140 — right approach, two parity gaps against the ops you are matching

Setup: worktree at the PR head `07993538`, with `kornia/` swapped to the merge-base `1331a861` for the fail-on-base pass. `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS; `kornia.__file__` verified to resolve to the worktree in every block.

**The approach is the correct one.** #4115 offered a lazy-`factors` fix and an explicit fast path, and you took the fast path — which is right, because `F.interpolate` rejects a zero-sized output outright while `F.grid_sample` returns one, so the two families cannot be reconciled by reordering alone. Your comment says exactly that.

**Fail-on-base: 3/3, with the two distinct failure modes from the issue.**

```
# kornia/ at merge-base 1331a861
FAILED TestResize::test_zero_output_size  - ZeroDivisionError: division by zero (affwarp.py:655)
FAILED TestResize::test_zero_input_size   - RuntimeError: Input and output sizes should be greater than 0
FAILED TestRescale::test_zero_factor      - ZeroDivisionError: division by zero
# at the head: 97 passed, 10 skipped (float32 + float64)
```

I swept 16 shapes and entry points at the head and every one is right — 2D/3D/4D/5D, `B=0` with both empty and non-empty output, float64, `antialias=True`, `int` size, and the `Resize` / `Rescale` / `rescale` delegations all produce the expected empty shape and preserve dtype. I also like that `test_zero_output_size` asserts `warp_affine` inline: that pins the parity claim itself, not just the shape.

Two gaps, both against the convention the PR invokes.

### 1. The empty result is severed from autograd; the warp family's is not

`input.new_empty(...)` produces a tensor with no `grad_fn`, so `resize` stops being differentiable on this path:

```
resize(x,(0,4)):        shape=(1, 3, 0, 4) requires_grad=False grad_fn=None
warp_affine(x,M,(0,4)): shape=(1, 3, 0, 4) requires_grad=True  grad_fn=GridSampler2DBackward0
center_crop(x,(0,4)):   shape=(1, 3, 0, 4) requires_grad=True  grad_fn=GridSampler2DBackward0

backward():
  resize      -> RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
  warp_affine -> OK, x.grad is set
  center_crop -> OK, x.grad is set
```

This is the half of the convention that is written down: `_empty_warp_output_2d`'s docstring is *"Return an empty warp while retaining normal grid-sample validation **and autograd links**"*, and it goes out of its way to build a connected tensor (`grid_zero = transform.reshape(-1)[:1].sum() * 0.0`) rather than an unconnected one. `resize` is differentiable and sits inside augmentation pipelines, so a batch that degenerates to a zero-sized output now raises in `backward()` where the warp ops contribute a zero gradient.

The same trick works here and I measured it:

```python
zero = input.reshape(-1)[:1].sum() * 0.0
output = zero.reshape(*([1] * input.dim())).expand(*input.shape[:-2], size[0], size[1])
```

```
(1,3,8,8)->(0,4): shape=(1, 3, 0, 4) matches_warp=True requires_grad=True backward OK
(1,3,8,8)->(4,0): shape=(1, 3, 4, 0) matches_warp=True requires_grad=True backward OK
(2,3,8,8)->(0,0): shape=(2, 3, 0, 0) matches_warp=True requires_grad=True backward OK
```

Shape and dtype match `warp_affine` exactly, and a non-differentiable input is unaffected. The input guard above guarantees at least one element, so `reshape(-1)[:1]` is always safe.

Worth adding an assertion for it in `test_zero_output_size` — it fails today, which is what makes it a pin.

### 2. Zero-sized input *and* zero-sized output: you raise, the warp family returns empty

```
case                     resize (head)        warp_affine          center_crop
src ok    -> dst (0,4)   (1, 3, 0, 4)         (1, 3, 0, 4)         (1, 3, 0, 4)
src (0,4) -> dst (2,2)   ValueError           ValueError           ValueError
src (0,4) -> dst (0,4)   ValueError           (1, 3, 0, 4)         (1, 3, 0, 4)
```

Rows 1 and 2 are exact parity, message included — nicely done. Row 3 is the gap: nothing is mathematically undefined when both sides are empty, and the house rule is *"empty in → empty out, with validation parity; reject only when the math is undefined, not merely empty."* The new guard fires before the zero-output fast path can answer.

So please do **not** drop the input-side half, as your body offers — refine it. Ordering it after the size normalization gives parity for tuple sizes while keeping the principled rejection for `int` sizes, where the aspect ratio really is undefined:

```python
if isinstance(size, int):
    if h <= 0 or w <= 0:
        raise ValueError(f"Input image size must be positive. Got height={h}, width={w}.")
    aspect_ratio = w / h
    size = _side_to_image_size(size, aspect_ratio, side)
...
if size[0] == 0 or size[1] == 0:
    output = <connected empty, per point 1>
elif h <= 0 or w <= 0:
    raise ValueError(f"Input image size must be positive. Got height={h}, width={w}.")
else:
    ...
```

### Verdict

**REQUEST CHANGES.** Both items are small and both are about matching the three ops this PR names as its reference; the structure, the fast path, the tests and their fail-on-base are all correct and verified. Please also extend the tests with the autograd assertion and a `src (0,4) -> dst (0,4)` case, so neither gap can reopen.

hiyufan:
Both confirmed on my side and both fixed in `76ce5db7`. Thank you for measuring them rather than describing them — the autograd one I would not have thought to look for.

### 1. The empty result now carries its graph

Reproduced your table before touching anything:

```
resize(x,(0,4)):        shape=(1, 3, 0, 4) requires_grad=False grad_fn=None
warp_affine(x,M,(0,4)): shape=(1, 3, 0, 4) requires_grad=True  grad_fn=GridSampler2DBackward0
center_crop(x,(0,4)):   shape=(1, 3, 0, 4) requires_grad=True  grad_fn=GridSampler2DBackward0
```

and after:

```
resize(x,(0,4)):        shape=(1, 3, 0, 4) requires_grad=True  grad_fn=ExpandBackward0
backward() -> OK, x.grad set
```

Took your construction with one change. `sum()` promotes an integer input to `int64`, so `input.reshape(-1)[:1].sum() * 0.0` returned an `int64` empty for a `uint8` input where the non-empty path returns `uint8`. Pinning the accumulation dtype fixes it:

```python
zero = input.reshape(-1)[:1].sum(dtype=input.dtype) * 0
output = zero.reshape(*([1] * input.dim())).expand(*input.shape[:-2], size[0], size[1])
```

```
float32 -> float32   float64 -> float64   float16 -> float16   int32 -> int32   uint8 -> uint8
```

`uint8` is the one that matters in practice — it is the only integer dtype the non-empty path accepts (`int32` raises `NotImplementedError` out of `upsample_bilinear2d`), so before this it was the case where an empty output silently changed dtype on you.

I also checked your "guarantees at least one element" note, since the guard only covers `h` and `w`: `B=0` and `C=0` inputs have no elements at all, but `reshape(-1)[:1].sum()` on an empty slice is a connected `0`, so `(0,3,8,8)` and `(2,0,8,8)` both come back connected and `backward()` still sets `.grad`.

### 2. Empty in and empty out now returns empty

```
case                     resize (before)   resize (now)   warp_affine    center_crop
src ok    -> dst (0,4)   (1, 3, 0, 4)      (1, 3, 0, 4)   (1, 3, 0, 4)   (1, 3, 0, 4)
src (0,4) -> dst (2,2)   ValueError        ValueError     ValueError     ValueError
src (0,4) -> dst (0,4)   ValueError        (1, 3, 0, 4)   (1, 3, 0, 4)   (1, 3, 0, 4)
```

Refined rather than dropped, exactly as you laid it out: the guard now sits inside the `isinstance(size, int)` branch, where the missing side really is derived from an aspect ratio that a degenerate input does not have, and the zero-output fast path answers first for tuple sizes. `resize(inp, 2)` on an empty input still raises with the same message.

### Tests

Extended the two existing cases rather than adding new ones, so the gaps cannot reopen:

- `test_zero_output_size` — asserts `requires_grad`, dtype, and that `backward()` runs and sets `.grad`
- `test_zero_input_size` — adds `src (0,4) -> dst (0,4)` asserted against `resize`, `warp_affine` and `center_crop` together

Both fail against the previous head of this PR (`07993538`): `assert False` on the autograd assertion, `ValueError` on the parity one.

```
head 76ce5db7          97 passed, 10 skipped   (float32 + float64)
kornia/ at 1331a861    6 failed, 91 passed, 10 skipped
                       test_zero_output_size  ZeroDivisionError
                       test_zero_input_size   RuntimeError: Input and output sizes should be greater than 0
                       test_zero_factor       ZeroDivisionError
```

so the three original cases still fail on the merge-base. `ruff format --check` and `ruff check` clean on both files.

ducha-aiki:
## Delta re-review of #4140 at `76ce5db7` — both asks landed; one thing left, and it is a test guard

Setup: worktree at the PR head, `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU + Apple MPS (`PYTORCH_ENABLE_MPS_FALLBACK` unset); `kornia.__file__` asserted to resolve inside the worktree in every block.

### Both asks are addressed, and I re-measured both

**1. Autograd link.** The empty result is connected now, and the `sum(dtype=input.dtype)` detail is a genuine improvement on what I proposed — my version would have promoted an integer input to int64.

```text
cpu  resize(x,(0,4))  (1, 3, 0, 4)  requires_grad=True   backward OK, x.grad set
mps  resize(x,(0,4))  (1, 3, 0, 4)  requires_grad=True   backward OK, x.grad set
```

**2. Empty in and empty out.** The `int`-size rejection is now hoisted above the size normalization and the tuple path returns empty, which is the parity the house rule asks for:

```text
src (0,4) -> dst (2,2)   ValueError "Input image size must be positive"
src (0,4) -> dst 2       ValueError  (int size: aspect ratio genuinely undefined)
src (0,4) -> dst (0,4)   (1, 3, 0, 4)      <- was ValueError
src (0,4) -> dst (0,2)   (1, 3, 0, 2)
```

The comment explaining why the zero-output branch has to run ahead of the positive-size check is exactly the kind that stops a later refactor from reordering them back.

### The remaining blocker is not your code — it is what your tests assert about someone else's

`tests-mps` is red, and it is now a blocking gate (#4160 merged yesterday):

```text
FAILED tests/geometry/transform/test_affine.py::TestResize::test_zero_output_size[mps-float32]
FAILED tests/geometry/transform/test_affine.py::TestResize::test_zero_input_size[mps-float32]
RuntimeError: [srcBuf length] > 0 INTERNAL ASSERT FAILED at OperationUtils.mm:551,
              please report a bug to PyTorch. Placeholder tensor is empty!
```

I reproduced it locally, and the split is clean — **`resize` itself is correct on MPS**; the crash comes only from the parity assertions on the warp family:

```text
mps  resize(x,(0,4))                 (1, 3, 0, 4)     OK
mps  resize(empty,(0,4))             (1, 3, 0, 4)     OK
mps  resize autograd                 OK
mps  warp_affine(x, eye, (0,4))      RuntimeError [srcBuf length] > 0 ...
mps  center_crop(empty, (0,4))       RuntimeError [srcBuf length] > 0 ...
```

That is **#4032**, filed before this PR: `grid_sample` asserts on an empty MPS output. Your test is asserting the behaviour of an op that is broken on that backend, so the gate catches the neighbour rather than the change.

**Ask: guard the three warp-family assertions on MPS, citing #4032.** Verified in the worktree — with this, `-k zero` on MPS float32 is `3 passed`:

```python
# in test_zero_output_size
if device.type != "mps":  # kornia#4032: grid_sample asserts on an empty MPS output
    eye = torch.eye(2, 3, device=device, dtype=dtype)[None]
    assert kornia.geometry.transform.warp_affine(inp, eye, (0, 4)).shape == (1, 3, 0, 4)

# in test_zero_input_size
if device.type != "mps":  # kornia#4032
    eye = torch.eye(2, 3, device=device, dtype=dtype)[None]
    assert kornia.geometry.transform.warp_affine(inp, eye, (0, 4)).shape == (1, 3, 0, 4)
    assert kornia.geometry.transform.center_crop(inp, (0, 4)).shape == (1, 3, 0, 4)
```

Please keep them on every other device rather than deleting them — the inline parity assertion is the best part of these tests, because it pins the claim the fix is *about*, and it should start passing on MPS by itself the day #4032 is fixed. Do **not** add these to `testing/known_failure_xfails/mps_float32.txt`: that manifest is the pre-existing baseline, and a new entry there would record your tests as broken when what is broken is `warp_affine`.

`collector` is just the aggregator reporting the same MPS leg; it will clear with it.

Nothing else open — the production diff is ready.

hiyufan:
Done in `bc0fd77d`.

Guarded exactly the three you named, with the `#4032` citation, and left them running on every other device:

- `test_zero_output_size` — the `warp_affine` parity assertion
- `test_zero_input_size` — the `warp_affine` and `center_crop` empty-in/empty-out assertions

Not added to `testing/known_failure_xfails/mps_float32.txt`, for the reason you gave — the manifest would record these tests as broken when what is broken is `grid_sample`.

One thing to be straight about: **I cannot verify the MPS half myself.** I have no Apple hardware, so I am taking your measurement for the crash and for the `3 passed` after the guard. What I can confirm is that the guard changes nothing elsewhere — CPU is `97 passed, 10 skipped` on float32 + float64, unchanged from `76ce5db7`, and `ruff check` / `ruff format --check` are clean.

I also kept `resize` itself unguarded on MPS, since your split showed it is correct there. That is the part worth preserving: if #4032 is fixed and someone drops the guard, the assertions come back on their own, and if `resize` ever regresses on MPS the leg still catches it.

Comment written into the test so the next reader knows the guard is about `grid_sample` and not about `resize`:

```python
# Guarded on MPS only: grid_sample asserts on an empty output there (#4032), so the
# warp family cannot answer. resize itself is fine on MPS and stays unguarded, and
# this assertion starts passing again on its own once #4032 is fixed.
```

Thanks for splitting the MPS failure down to which op actually breaks — a red leg on `test_zero_output_size` reads like my change at first glance, and I would probably have gone looking in `resize`.

ducha-aiki:
## Delta re-review of #4140 at `bc0fd77d` — the guard I asked for is now obsolete; please drop it

Setup: worktree at the PR head, `kornia/` swapped to the merge-base `1331a861` for every A/B pass. `~/dev/kornia/.venv/bin/python`, **torch 2.14.0**, CPU + MPS (`PYTORCH_ENABLE_MPS_FALLBACK` unset for every MPS claim).

You did exactly what I asked, on the three assertions I named, with the `#4032` citation and the "starts passing again on its own once #4032 is fixed" comment. The problem is that `main` moved underneath both of us between my review and your push, and my ask no longer holds.

### What changed: the MPS gate is on torch 2.14 now, and 2.14 fixes #4032

**#4202** (`7efa3684`, merged last night) moved the blocking `tests-mps` leg to **torch 2.14.0 on macos-15**; it was 2.9.1 on macos-latest when I reviewed you. And #4032 is a torch-side MPS `grid_sample` defect that 2.14 repairs:

```text
torch 2.14.0, MPS, PYTORCH_ENABLE_MPS_FALLBACK unset
  F.grid_sample, empty (1, 0, 4, 2) grid    OK  (1, 3, 0, 4)
  warp_affine dsize=(0, 4)                  OK  (1, 3, 0, 4)
  warp_perspective dsize=(0, 4)             OK  (1, 3, 0, 4)
  center_crop (0, 4)                        OK  (1, 3, 0, 4)
  resize (0, 4)                             ZeroDivisionError      <- what your PR fixes
```

The three ops your assertions compare against all answer correctly on MPS now. So the guard skips the parity check on the one backend where that check is a blocking gate, permanently — the comment's "starts passing again on its own" condition is already met, and nothing would ever remove it.

### Verified without the guard

I neutralised both `if device.type != "mps":` branches in the worktree and re-ran:

```text
-k "zero_output_size or zero_input_size"   --device=mps  --dtype=float32   ->  2 passed
-k "zero_output_size or zero_input_size"   --device=cpu   --dtype=float32,float64  ->  4 passed
whole tests/geometry/transform/test_affine.py  --device=mps --dtype=float32   ->  46 passed, 11 skipped
```

The whole-file MPS run is the one that matters — a per-node MPS pass can hide an order-dependent failure, and this does not have one.

**Please revert the two guard hunks** (`bc0fd77d`), restoring the plain assertions from `76ce5db7`. The rest of that commit stays as it is.

I would rather you did not keep the guard "just in case": kornia's minimum is `torch>=2.5.1` and the MPS gate pins 2.14.0, so the skip would fire on no CI configuration at all, and a permanently-dead device guard is worse than no guard — it reads as a live limitation to the next person.

### The rest of the PR is unchanged and still verified

```text
fail-on-base (kornia/ at 1331a861), the two new tests, cpu float32:
  FAILED test_zero_output_size   ZeroDivisionError: division by zero
  FAILED test_zero_input_size    RuntimeError: Input and output sizes should be greater than 0
  2 failed

head, both backends:
  cpu  resize -> (1, 3, 0, 4)  grad_fn=ExpandBackward0  x.grad populated  dtype float32
  mps  resize -> (1, 3, 0, 4)  grad_fn=ExpandBackward0  x.grad populated  dtype float32
       parity: warp_affine (1, 3, 0, 4)   center_crop (1, 3, 0, 4)
```

`sum(dtype=input.dtype)` remains better than the version I originally proposed — mine promoted integer inputs to int64, yours does not. The `size[0] == 0 or size[1] == 0` branch running *ahead* of the positive-input check is right, and the comment explaining why is the reason I did not have to work it out again.

### Verdict

**One ask, and it is a retraction of my own.** Drop the two MPS guards; everything else is done. Sorry for the churn — the ask was correct against torch 2.9.1 when I made it and stopped being correct about fourteen hours later. Thank you for the two clean rounds before this one.

hiyufan:
Reverted — the branch is back at `76ce5db7`, the two `if device.type != "mps":` hunks and their `#4032` comments gone, nothing else touched. Force-pushed rather than adding a revert commit so the PR reads as the two commits you already verified.

No apology needed: your ask was right when you made it, and a maintainer who re-checks their own review against a moved `main` and retracts it is doing more work than one who does not. I would rather take one extra round than carry a guard that skips the parity check on the backend where it is a blocking gate.

Your argument for dropping it is the one that settles it for me — kornia's floor is `torch>=2.5.1` and the MPS gate pins 2.14.0, so the skip would fire on no CI configuration at all. A guard that can never fire still reads as a live limitation to whoever hits it next, and the "starts passing again once #4032 is fixed" comment I wrote would have been describing a condition that was already true.

CPU re-checked after the revert: `97 passed, 10 skipped` on float32 + float64, and `ruff` clean. The MPS half I still cannot run — no Apple hardware here — so the `2 passed` on MPS and the `46 passed, 11 skipped` whole-file run are yours, not mine.

ducha-aiki:
## Delta re-review of #4140 at `76ce5db7` — the retraction landed; one thing left (a CHANGELOG entry) and a rebase

Thanks for reverting cleanly — sorry again for the round-trip on a guard that expired under both of us.

### The guards are gone and the parity assertions really do run now

`grep -rnE "#4032|device.type != .mps." kornia/ tests/` at your head returns nothing, and the three assertions are live on the backend where they are a blocking gate. Measured with `PYTORCH_ENABLE_MPS_FALLBACK` **unset**, torch 2.14.0:

```text
head, cpu float32+float64, -k "zero_output_size or zero_factor"     4 passed
head, mps float32,          -k "zero_output_size or zero_factor"    2 passed
head, whole test_affine.py, cpu float32+float64                    97 passed, 10 skipped
head, whole test_affine.py, mps float32                            46 passed, 11 skipped
```

Fail-on-base, your test file against base `kornia/` (`1331a861`):

```text
FAILED TestResize::test_zero_output_size[cpu-float32]  - ZeroDivisionError: division by zero   (affwarp.py:655)
FAILED TestRescale::test_zero_factor[cpu-float32]      - ZeroDivisionError: division by zero
2 failed
```

Everything I verified in the earlier rounds still stands: the empty result keeps its autograd link, `sum(dtype=input.dtype)` avoids the int64 promotion my original suggestion would have had, and the `int` size path raises the warping ops' own `ValueError` message instead of a bare `ZeroDivisionError`.

The AGENTS issue-grep gate is satisfied vacuously: `grep -rnE "#4115|issues/4115" kornia/ tests/` is empty at your head **and** on `main`, so #4115 was never documented inline and there is no prose or pin to sweep.

### The one ask: a CHANGELOG entry

`git diff --stat` against your merge-base is two files, `kornia/` and `tests/`, with no `CHANGELOG.md`. This is a user-visible change — `resize`/`rescale` go from raising `ZeroDivisionError` to returning an empty image, and an `int` size against a degenerate input now raises `ValueError` with a specific message — so `AGENTS.md` wants an entry under `## Unreleased` → `### Bug fixes` citing #4140, saying what used to happen as well as what happens now. Add `Fixes #4115` to the PR body while you are there so it auto-closes on merge.

### Please also rebase

Your base is `1331a861` (2026-08-31), which predates #4202. That is why the red check reads `tests-mps / tests (3.13, 2.9.1)`: it is the **old** MPS gate, on `macos-latest` at torch 2.9.1, which #4202 replaced with `macos-15` at torch 2.14.0 — the same change that made #4032 moot. Rebasing onto current `main` retires that job and runs the new one, on which I measured your assertions passing above. No diff change needed for it.

CHANGELOG plus rebase and this is done.

hiyufan:
Both done in `2a804424`. The `Fixes #4115` ask was already satisfied — details below, because it is worth knowing why you did not see it.

### CHANGELOG

Added as the first bullet of `## Unreleased` → `### Bug fixes`, citing #4140 and the report, and giving the old behaviour alongside the new one on both halves — the second one especially, since a caller who was relying on `resize(x, 224)` against a degenerate input reaching `interpolate` gets a different exception type now:

> `resize` and `rescale` return an empty image for a zero-sized output instead of raising `ZeroDivisionError` out of the scale-factor computation (#4115). `resize(x, (0, 4))`, `resize(x, 0)`, `Resize((0, 4))`, and any `rescale` factor that takes a side to zero (`rescale(x, 0.0)`) previously died in `h / size[0]`; they now return an empty tensor of the requested shape, which is the answer `warp_affine`, `warp_perspective` and `center_crop` already gave for the same request. The empty result stays connected to the autograd graph, the way `_empty_warp_output_2d` builds it for the warping ops, so a batch element that degenerates to a zero-sized output still contributes a zero gradient rather than detaching.
>
> A degenerate *input* is a different question and is now rejected rather than divided by: resizing a `(1, 3, 0, 4)` image to a positive size, or with an `int` size at all — an `int` names one side and takes the other from an aspect ratio a degenerate input does not have — raises `ValueError("Input image size must be positive. Got height=0, width=4.")`, the message the warping ops already use, instead of the same bare `ZeroDivisionError`. Empty in and empty out is well defined and still comes back empty. (#4140)

It went in as its own commit rather than folded into the two you have already verified, so the diff you re-measured is still reachable unchanged.

### `Fixes #4115` is already there, and already linked

It has been in the body since the PR opened, under `## Related issue or discussion`, and GitHub has it registered — the auto-close link resolves:

```text
$ gh api graphql -f query='{repository(owner:"kornia",name:"kornia"){pullRequest(number:4140){
    closingIssuesReferences(first:5){nodes{number title}}}}}'
[{"number":4115,"title":"resize/rescale raise ZeroDivisionError for a zero output size, while warp_affine returns an empty image"}]
```

I am flagging it rather than adding a second `Fixes #4115` because a duplicate would not have changed anything and you would have had to ask a third time why the link still looked missing. If the sidebar is not showing it on your end, that is a display question, not a body question.

### Rebase

`1331a861` → `ca1f2f7b`. No conflicts, and the source diff is byte-identical to what you measured — `git diff --stat origin/main..HEAD` is the same two files plus `CHANGELOG.md`:

```text
CHANGELOG.md                            | 16 ++++++++
kornia/geometry/transform/affwarp.py    | 37 ++++++++++++++----
tests/geometry/transform/test_affine.py | 66 +++++++++++++++++++++++++++++++++
```

Your read of the red check was right, and CI on the new head confirms it — the old gate is gone and the new one is the job you measured on:

```text
before (76ce5db7):  tests-mps / tests (3.13, 2.9.1)      <- macos-latest, the #4032 backend
after  (2a804424):  tests-mps / tests (3.13, 2.14.0)     <- macos-15, post-#4202
```

### Re-verified after the rebase

CPU only; the MPS numbers in this thread remain yours, I have no Apple hardware.

```text
head, cpu float32+float64, -k "zero_output_size or zero_input_size or zero_factor"    6 passed
head, whole test_affine.py, cpu float32+float64                                      97 passed, 10 skipped
ruff check + ruff format --check, affwarp.py + test_affine.py                        clean
```

Fail-on-base re-run against the *new* merge-base `ca1f2f7b` rather than the old one, so the tests are still proving something about current `main`:

```text
FAILED TestResize::test_zero_output_size[cpu-float32]  - ZeroDivisionError: division by zero    (affwarp.py:655)
FAILED TestResize::test_zero_input_size[cpu-float32]   - RuntimeError: Input and output sizes should be greater than 0
FAILED TestRescale::test_zero_factor[cpu-float32]      - ZeroDivisionError: division by zero
3 failed
```

Thanks for staying on this one through the round-trip on the guard.

ducha-aiki:
Rebased this onto `main` for you, `2a804424` -> `bf71450d`, because today's merges put `CHANGELOG.md` entries on the same lines as yours and the PR had gone `CONFLICTING`.

What changed: only the conflict resolution in `CHANGELOG.md` (main's entries first, yours after). Your three commits are intact, and the code and test delta is byte-identical to the approved head: `git diff <merge-base> 2a804424 -- . ':!CHANGELOG.md'` and the same for `bf71450d` against current `main` are the same patch. `tests/geometry/transform/test_affine.py` + `test_imgwarp.py` on the new head: 331 passed, 10 skipped (cpu, float32+float64).

If you have local work on this branch, `git fetch` and reset to `bf71450d` before pushing anything else. Nothing is owed from your side; this merges once CI is green on the new head.

ducha-aiki (approved):

hiyufan:
Thanks for taking the rebase yourself — local branch reset to `bf71450d`, and I agree the resolution is only `CHANGELOG.md` ordering.

**The red on `bf71450d` is not from this PR, and it is not really an MPS problem either.** It is a pre-existing ~2% flake in `test_jpeg.py`, and I think it is worth fixing because the flake is a real NaN in a documented input range. `collector` is just the aggregator failing behind it.

### What failed

```text
FAILED tests/enhance/test_jpeg.py::TestDiffJPEG::test_keyword_argument[mps-float32]
Mismatched elements: 3072 / 6144 (50.0%)
Greatest absolute difference: nan at index (0, 0, 0, 0)
```

Both sides of the comparison are `jpeg_codec_differentiable` on the *same* image and quality, once with `input=` and once positionally, and both came back NaN at the same indices. `assert_close` runs with `equal_nan=False`, so NaN vs. NaN fails even though the two calls agree exactly. 3072/6144 is exactly one of the two batch elements.

### Why it is not this PR, even though it is not unrelated by inspection

`kornia/enhance/jpeg.py` does import `rescale` from `affwarp` and calls it three times, so I checked rather than assumed. The factors are `0.5` and `2.0` on a 32×32 image, so the sizes are 16×16 and 64×64 — never zero, and `h`/`w` are never zero — which lands in the `else` branch this PR left byte-identical. A bitwise A/B over 12 (image, quality-pair) combinations, sha256 of the raw output bytes:

```text
bf71450d (head)        06666002934b5c6b76c6ea56dec856c505ddd55db7434f158b2df45b1785a504
1751cdb1 (merge-base)  06666002934b5c6b76c6ea56dec856c505ddd55db7434f158b2df45b1785a504
```

### The actual cause: `jpeg_quality=0` returns all-NaN

`_jpeg_quality_to_scale` computes `5000.0 / compression_strength` for quality below 50:

```text
q=  0.0   5000/q -> inf     _differentiable_polynomial_floor(inf) -> nan
q=  1.0   5000/q -> 5000.0  -> 4999.875
q= 49.0   5000/q -> 102.04  -> 101.903
q=100.0   200-2q -> 0.0     -> -0.125
```

Quality 0 is not out of contract — the docstring says *"JPEG quality in the range :math:`[0, 100]`"* and the `KORNIA_CHECK` at line 587 explicitly admits `>= 0.0`. So a documented input silently returns an all-NaN image.

The test draws its quality unseeded:

```python
jpeg_quality = torch.randint(low=0, high=100, size=(B,), device=device, dtype=dtype)
```

`P(a zero in a draw of 2) = 1 - 0.99² ≈ 1.99%`. Measured on **CPU**, 500 iterations of the test body:

```text
runs=500   draws containing quality 0: 9   assert_close failures: 9
every failure had a zero in the drawn quality; the other 491 were bitwise identical
```

Nine for nine, 1.8% against a predicted 1.99%. Nothing device-specific — MPS is simply the job that happened to draw a zero. It can hit any job in the matrix.

### Suggestion

Two separable things: the test should not be unseeded, and `jpeg_quality=0` should probably not return NaN for an input the docstring and the check both accept — clamping the divisor, or narrowing the documented range and the check to `[1, 100]`, are both defensible and it is your call which. Happy to send that as its own PR if you want it; I did not want to bolt it onto this one.

For this PR, a re-run of the MPS job should be green. Nothing changed on my side since your rebase.

ducha-aiki (approved):
Approved at `57f909d4` after the maintainer-authorized recovery.

Rebased onto main f4532f39, which includes the JPEG quality-zero fix behind the old MPS failure. The non-CHANGELOG patch-id is identical to the approved head, and all existing main release notes are preserved. Focused affine tests: 97 passed, 10 skipped.

hiyufan:
Thanks for the recovery rebase, and for landing the JPEG quality-zero fix on `main` — good to see that one was worth writing up.

CI on `57f909d4` is red on one job again, and again it is not this PR. This time it is `tests/feature/test_xfeat.py::TestXFeatModel::test_dynamo[cpu-float32-inductor]`:

```text
Mismatched elements: 4 / 4096 (0.1%)
Greatest absolute difference: 0.000231288 at index (0, 38, 1, 1)   (up to 0.0001 allowed)
Greatest relative difference: 0.016798     at index (0, 15, 1, 1)   (up to 0.001 allowed)
1 failed, 255 passed, 21 skipped
```

That is `torch.compile(..., backend="inductor")` against eager on the same module — four elements over an `atol` of `1e-4`, on randomly initialised weights and an unseeded `torch.rand(1, 3, 64, 64)` input. `collector` is failing behind it.

Three reasons it cannot be this branch:

1. **No import path.** `kornia/feature/xfeat.py` imports exactly two things from kornia — `KORNIA_CHECK` and `kornia.geometry.subpix.nms2d`. Every resampling inside `XFeatModel` is `torch.nn.functional.interpolate` called directly, not `kornia.geometry.transform.resize`. `affwarp.py` is not reachable from that test.

2. **Bitwise A/B.** `XFeatModel` eager forward under a fixed seed, sha256 over all three returned tensors:

   ```text
   57f909d4 (head)        90f976a67d0474fde2fb847c407ad5fd46f5277c14542e90209d76fa4fbc0f2e
   f4532f39 (merge-base)  90f976a67d0474fde2fb847c407ad5fd46f5277c14542e90209d76fa4fbc0f2e
   ```

3. **The diff.** `git diff origin/main..HEAD --stat` is `CHANGELOG.md`, `affwarp.py` and `test_affine.py`, and the `affwarp.py` change is confined to a branch that only runs when a requested side is zero or an input side is non-positive. Neither is true anywhere in that test.

So a re-run should clear it, the same as last time. I still only have `pull` on this repo, so I cannot press the button myself:

```text
kornia/kornia permissions: {"admin": false, "maintain": false, "push": false, "pull": true}
```

If this job turns out to be a repeat offender rather than a one-off, the same treatment as the JPEG one would apply — seed the input and the initialisation, and decide whether `atol=1e-4` is the right bar for inductor on this graph. Happy to look at that separately if it is useful; it is not something this PR should carry.

PR #1996, opened 2022-11-08
[wip] add colormap and `apply_ColorMap` base

johnnv1:
Adds `RGBColor`, `Colormap` and `apply_colormap` (based on cv2)

Fixes #1991 

#### Type of change
- [ ] 📚  Documentation Update
- [ ] 🧪 Tests Cases
- [x] 🔬 New feature (non-breaking change which adds functionality)

PR #4143, opened 2026-09-01
fix(core): bind eye_like and vec_like, and check __all__ names resolve

### LINKED ISSUE #4023: Broken `__all__` entries: `from kornia.core import *` and `from kornia.color import *` raise AttributeError, and the API-surface guard passes anyway

hiyufan:

Two names in `kornia.core.__all__` and one in `kornia.color.__all__` are not bound in their modules, so `from kornia.core import *` and `from kornia.color import *` both raise `AttributeError`.

`eye_like` and `vec_like` live in `kornia.core.ops` and are used across the library, and `tests/api_surface.json` records both as public `kornia.core` API — `kornia/core/__init__.py` never imported them. #3449 (`94882c52`, "deprecate utils module across library") added both names to `kornia.core.__all__` and left the imports untouched, so they were unbound from the moment they were declared. That import is restored.

`AUTUMN` was a real public class, not a name that never existed: added in #1996 (`50f783d2`, 2023-01-06) and bound by `from .colormap import AUTUMN, ...`, deprecated since 0.7.2 in favour of `ColorMap(base='autumn')`, and removed by #3432 (`9e521cac`, 2025-12-31) — which dropped the class and the import but left the `__all__` entry behind. So `from kornia.color import AUTUMN` worked for about three years and has raised `AttributeError` since that removal.

That makes dropping the entry the tail of a deprecation that already ran its course rather than a fresh removal: the stability policy's window was served between 0.7.2 and the removal in #3432, and what is left is bookkeeping. It is dropped from `__all__` and from the inventory, with a release-note entry recording it, as `test_no_public_name_removed`'s own assertion message asks for.

`test_no_public_name_removed` could not catch either case. The surface it compares against the inventory is `__all__` itself, so a name whose binding went missing still reads as present — the guard added in #3898 for exactly this class of problem passes on a module whose `import *` is broken. `test_public_names_resolve` closes that: every declared name has to be bound in its module.

Fixes #4023

## What did you check?

The three names, before and after:

```text
# before
kornia.color.AUTUMN      in __all__: True   bound: False
kornia.core.eye_like     in __all__: True   bound: False
kornia.core.vec_like     in __all__: True   bound: False
from kornia.core import *   -> AttributeError: module 'kornia.core' has no attribute 'eye_like'
from kornia.color import *  -> AttributeError: module 'kornia.color' has no attribute 'AUTUMN'

# after
from kornia.core import *   -> OK
from kornia.color import *  -> OK
kornia.core.eye_like(3, torch.rand(2, 3, 3)).shape == (2, 3, 3)
```

I swept every module in `tests/api_surface.json` rather than only the three in the report; those three are the only unbound names. `kornia.geometry` and `kornia.morphology` define no `__all__` and are skipped by the new test.

The new test fails on the unpatched source and passes with the fix:

```text
# new test against the pre-fix modules
FAILED tests/test_api_surface.py::test_public_names_resolve[kornia.color]
FAILED tests/test_api_surface.py::test_public_names_resolve[kornia.core]
2 failed, 8 passed, 2 skipped

$ pytest tests/test_api_surface.py
22 passed, 2 skipped

$ pytest tests/test_api_surface.py tests/color
448 passed, 4 skipped

$ pre-commit run --files kornia/core/__init__.py kornia/color/__init__.py \
                        tests/test_api_surface.py tests/api_surface.json
(all hooks passed)
```

I did not run `pixi run typecheck` or the full suite — I worked in a plain venv rather than a Pixi environment, so those are worth a look in CI.

## How did AI help?

AI (Claude Code) wrote the change and the new test, ran the sweep and the before/after comparison, and drafted this description. I reviewed every line and the numbers above come from runs on my own machine. The `AUTUMN` removal is the part I would rather you decide than have me assert, and it is called out above for that reason.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -207,6 +207,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   call now samples on the image's device and matches the same-device result (`extract_patches_simple` previously
   built the full sampling grid on the LAF's device).
 
+* `AUTUMN` is dropped from `kornia.color.__all__` (#4143). The class was deprecated in 0.7.2 in favour of
+  `ColorMap(base='autumn')` and removed by #3432, which left the `__all__` entry behind. Since that removal
+  `from kornia.color import AUTUMN` has raised `AttributeError` and `from kornia.color import *` has failed on
+  it, so nothing that works today stops working — the entry named a binding that no longer existed. Callers
+  still reaching for the class should use `ColorMap(base='autumn')`.
+
 * The shape guards of `normalize_homography` and `denormalize_homography` now reject everything but a
   `(3, 3)`/`(B, 3, 3)` matrix, and `normalize_homography3d`'s everything but a `(4, 4)`/`(B, 4, 4)` one (#3999).
   A rank-4 input used to pass the guard and come back with its rank unchanged —
--- a/kornia/color/__init__.py
+++ b/kornia/color/__init__.py
@@ -72,7 +72,6 @@ from .yuv import (
 )
 
 __all__ = [
-    "AUTUMN",
     "CFA",
     "ApplyColorMap",
     "BgrToGrayscale",
--- a/kornia/core/__init__.py
+++ b/kornia/core/__init__.py
@@ -29,6 +29,7 @@ from .mixin import (
     ONNXRuntimeMixin,
 )
 from .module import ImageModule, ImageSequential
+from .ops import eye_like, vec_like
 from .safetensors import load_safetensors
 from .tensor_wrapper import TensorWrapper
 
--- a/tests/api_surface.json
+++ b/tests/api_surface.json
@@ -88,7 +88,6 @@
     "container"
   ],
   "kornia.color": [
-    "AUTUMN",
     "ApplyColorMap",
     "BgrToGrayscale",
     "BgrToRgb",
--- a/tests/test_api_surface.py
+++ b/tests/test_api_surface.py
@@ -30,6 +30,7 @@ Additions are fine and do not fail the test; run this file with
 ``regenerate()`` below) whenever new public API lands.
 """
 
+import contextlib
 import importlib
 import json
 import pkgutil
@@ -102,12 +103,12 @@ def _modules_declaring_all() -> list:
 
     found = []
     for info in pkgutil.walk_packages(kornia.__path__, "kornia."):
-        try:
+        # Optional extras and backends are out of scope for this guard, so a module that
+        # cannot be imported here is skipped rather than failing the sweep.
+        with contextlib.suppress(Exception):
             mod = importlib.import_module(info.name)
-        except Exception:
-            continue  # optional extras and backends are out of scope for this guard
-        if getattr(mod, "__all__", None) is not None:
-            found.append(info.name)
+            if getattr(mod, "__all__", None) is not None:
+                found.append(info.name)
     return sorted(found)
 
 
--- a/tests/test_api_surface.py
+++ b/tests/test_api_surface.py
@@ -86,6 +86,30 @@ def test_no_public_name_removed(module_name):
     )
 
 
+@pytest.mark.parametrize("module_name", sorted(json.loads(INVENTORY.read_text())))
+def test_public_names_resolve(module_name):
+    """
+    Every name in ``__all__`` has to be bound in its module.
+
+    An unbound name breaks ``from <module> import *`` for everyone, and
+    ``test_no_public_name_removed`` cannot see it: the surface that test compares is
+    ``__all__`` itself, so a name whose binding was dropped still reads as present.
+    """
+    mod = importlib.import_module(module_name)
+    declared = getattr(mod, "__all__", None)
+    if declared is None:
+        pytest.skip(f"{module_name} does not define __all__")
+
+    unbound = sorted(name for name in declared if not hasattr(mod, name))
+
+    assert not unbound, (
+        f"{module_name}.__all__ lists names that are not bound in the module: "
+        f"{unbound}. `from {module_name} import *` raises AttributeError on the "
+        "first of them. Either restore the binding or drop the name from __all__ "
+        "and from tests/api_surface.json in the same PR."
+    )
+
+
 if __name__ == "__main__":
     regenerate()
     print(f"regenerated {INVENTORY}")
--- a/tests/test_api_surface.py
+++ b/tests/test_api_surface.py
@@ -32,6 +32,8 @@ Additions are fine and do not fail the test; run this file with
 
 import importlib
 import json
+import pkgutil
+import warnings
 from pathlib import Path
 
 import pytest
@@ -86,7 +88,30 @@ def test_no_public_name_removed(module_name):
     )
 
 
-@pytest.mark.parametrize("module_name", sorted(json.loads(INVENTORY.read_text())))
+def _modules_declaring_all() -> list:
+    """Every importable ``kornia`` submodule that declares ``__all__``.
+
+    The inventory tracks the stable-core modules, but an unbound name is a problem
+    wherever it sits: the breakage that motivated this guard (#4026, the
+    ComfyUI-LTXVideo report in #3986) was in ``kornia.geometry.transform.pyramid``,
+    which the inventory does not cover. Walking the package finds every module that
+    declares a public surface, costs a fraction of a second, and needs no upkeep as
+    modules come and go.
+    """
+    import kornia
+
+    found = []
+    for info in pkgutil.walk_packages(kornia.__path__, "kornia."):
+        try:
+            mod = importlib.import_module(info.name)
+        except Exception:
+            continue  # optional extras and backends are out of scope for this guard
+        if getattr(mod, "__all__", None) is not None:
+            found.append(info.name)
+    return sorted(found)
+
+
+@pytest.mark.parametrize("module_name", _modules_declaring_all())
 def test_public_names_resolve(module_name):
     """
     Every name in ``__all__`` has to be bound in its module.
@@ -96,11 +121,14 @@ def test_public_names_resolve(module_name):
     ``__all__`` itself, so a name whose binding was dropped still reads as present.
     """
     mod = importlib.import_module(module_name)
-    declared = getattr(mod, "__all__", None)
-    if declared is None:
-        pytest.skip(f"{module_name} does not define __all__")
+    declared = mod.__all__
 
-    unbound = sorted(name for name in declared if not hasattr(mod, name))
+    with warnings.catch_warnings():
+        # ``hasattr`` runs module-level ``__getattr__``, which is how the deprecated
+        # ``kornia.utils`` re-exports announce themselves. Keep the probe quiet so it
+        # does not emit a DeprecationWarning per run just by looking.
+        warnings.simplefilter("ignore")
+        unbound = sorted(name for name in declared if not hasattr(mod, name))
 
     assert not unbound, (
         f"{module_name}.__all__ lists names that are not bound in the module: "


ducha-aiki:
## Review of #4143 — verified by execution, not by reading

Setup: worktree at the PR head `7d83f8cb`, with `kornia/` swapped to the merge-base `e5b8b49d` for the fail-on-base pass. `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU/macOS, `kornia.__file__` confirmed to resolve to the worktree in every block.

**The code is right and the fix is complete. Two things to fix before merge, both about the record rather than the diff — and one of them makes your case stronger than the one you argued.**

### Verified

Both defects reproduce on the merge-base, and the new test discriminates:

```
# kornia/ at merge-base e5b8b49d
kornia.core:  AttributeError: module 'kornia.core' has no attribute 'eye_like'
kornia.color: AttributeError: module 'kornia.color' has no attribute 'AUTUMN'

tests/test_api_surface.py
FAILED test_public_names_resolve[kornia.color]
FAILED test_public_names_resolve[kornia.core]
2 failed, 20 passed, 2 skipped

# at the head
22 passed, 2 skipped
```

**The fix is complete repo-wide, not just for the two reported modules.** I walked every package and checked every declared name:

```
modules with __all__ scanned: 132   (inventory covers 12)
still-broken modules at this head: 0
```

So there is no third instance hiding behind these two.

### 1. The `AUTUMN` history in the PR body is wrong — and the truth is a better argument

You write that `AUTUMN` "is bound nowhere", that "there is nothing here to deprecate", and that "`from kornia.color import AUTUMN` has never succeeded". That last claim is false. `AUTUMN` was a real, bound, documented public class for about three years:

```
50f783d2 (2023-01-06, #1996)  kornia/color/colormap.py:113  class AUTUMN(ColorMap):
                              kornia/color/__init__.py:1    from .colormap import AUTUMN, ApplyColorMap, ...
v0.7.2 .. v0.8.2              kornia/color/colormap.py:276  class AUTUMN(ColorMap):
                              extra_reason="The `AUTUMN()` class is deprecated and will be removed
                                            in next kornia versions (0.8.0 - dec 2024)."
9e521cac (2025-12-31, #3432)  class AUTUMN removed, import line removed, __all__ entry left behind
```

It even had a rendered docs image (`.. image:: _static/img/AUTUMN.png`).

So the accurate story is: **`AUTUMN` was deprecated on schedule and then removed; the `__all__` entry is a leftover from that removal.** That reaches your conclusion by a route the stability policy actually blesses — the deprecation window was served years ago — instead of by a "never existed" claim that a reader can falsify in one `git log`. Please correct the body.

Worth recording while you are there: **the two halves of this PR have different root causes.** `AUTUMN` was orphaned by #3432 as above; `eye_like` / `vec_like` were orphaned by **#3449 `94882c52` ("deprecate utils module across library")**, which dropped the import from `kornia/core/__init__.py` and left both names in `__all__`. Two separate leftovers of the same shape, which is a fair argument that the guard you are adding is worth having.

Also checked, since it is the only surviving `AUTUMN` reference in the tree: `docs/generate_examples.py:386` passes the string `"AUTUMN"` to `ColorMap(base=...)`, and that still resolves — `ColorMap(base='AUTUMN')` and `ColorMap(base='autumn')` both construct fine. Nothing to do there.

### 2. Required by the existing test's own contract: a release-note entry

`test_no_public_name_removed`'s assertion message spells out the procedure you are following, and it has three steps, not two:

> "If this removal is deliberate and the deprecation window has passed, update `tests/api_surface.json` in this PR (run `tests.test_api_surface.regenerate()`) **and list the removal in the release notes**."

You did the inventory edit; the release-note entry is missing. Please add a line under `## Unreleased` recording that `AUTUMN` is dropped from `kornia.color.__all__`, noting it has been unbound since #3432 and was deprecated since 0.7.2. That is the deliberate-review moment the module docstring says the inventory edit exists to create, and right now the edit is silent.

### Non-blocking: the guard covers 12 modules of 132

`test_public_names_resolve` parametrizes over the inventory keys, so it watches the 12 stable-core modules. But 132 modules in the tree declare `__all__`, and the motivating case from #4026 — `kornia.geometry.transform.pyramid`, the ComfyUI-LTXVideo breakage in #3986 — is not one of the 12. The same bug there would still be invisible.

Discovering the modules instead costs nothing and is green today; this is the exact code I ran for the sweep above:

```python
@pytest.mark.parametrize("module_name", sorted(_all_modules_declaring_all()))
def test_public_names_resolve(module_name):
    ...
# where _all_modules_declaring_all() walks pkgutil.walk_packages(kornia.__path__, "kornia.")
# and keeps the modules that import cleanly and define __all__
```

132 modules, 0 failures, well under a second. Your call whether it belongs in this PR or a follow-up — the inventory-scoped version is already a strict improvement, so I am not blocking on it.

One small property of `hasattr` worth knowing: it fires module-level `__getattr__`, so the test emits the `kornia.utils` `CachedDownloader` DeprecationWarning on every run. Harmless today (`filterwarnings` has no `error` entry), but it means the probe is not side-effect-free.

### Verdict

**REQUEST CHANGES** on the two items above — the body correction and the release-note line. The production change, the inventory edit and the new test are all correct and I have verified them; neither ask touches the code.

hiyufan:
Thank you — the `AUTUMN` history is a real error on my part, and I checked both claims rather than taking them on trust. One holds, one does not.

### The `AUTUMN` correction is right, and I have taken it

Verified against the tree rather than the PR numbers:

```
                              class AUTUMN in colormap.py   AUTUMN in color/__init__.py import
50f783d2 (2023-01-06, #1996)  yes                           yes
v0.7.2                        yes                           yes
v0.8.2                        yes                           yes
9e521cac (2025-12-31, #3432)  no                            no
main today                    no                            no   (but still in __all__)
```

and the deprecation in v0.8.2 is explicit:

```python
@deprecated("0.7.2", extra_reason="The `AUTUMN()` class is deprecated and will be removed in next
            kornia versions (0.8.0 - dec 2024). In favor of using `ColorMap(base='autumn')` instead.")
class AUTUMN(ColorMap):
```

So `from kornia.color import AUTUMN` worked for about three years, and "has never succeeded" was wrong. You are also right that the true story is the stronger argument — the deprecation window was served between 0.7.2 and #3432, so what is left is bookkeeping rather than a removal I am asking you to bless. Body rewritten accordingly.

### The `eye_like` / `vec_like` mechanism is the other way round

You describe #3449 as having "dropped the import from `kornia/core/__init__.py` and left both names in `__all__`". The commit is right, the direction is not — it **added** the names without ever adding the binding:

```
94882c52~1 (32aec762)  __all__ = [ImageModule, ImageModuleMixIn, ImageSequential, ONNXExportMixin,
                                   ONNXMixin, ONNXRuntimeMixin, TensorWrapper, external]
                       imports  = external, mixin, module, tensor_wrapper      (no .ops)

94882c52   (#3449)     __all__ = [... same eight ..., "eye_like", "vec_like"]
                       imports  = unchanged                                    (still no .ops)
```

There is no import of `.ops` to drop on either side of that commit. So these two were unbound from the moment they were declared, which is a slightly different failure than `AUTUMN`'s: one is a leftover from a removal, the other is a declaration that never had a binding. I have written both into the body with the commits, since your point that the PR should record the causes is a good one either way.

### Release note added

Under `## Unreleased` → `### Breaking changes`, recording the drop, that the class was deprecated in 0.7.2 and removed by #3432, and that `ColorMap(base='autumn')` is the replacement. You are right that the assertion message asks for three steps and I had done two.

### Took the non-blocking one as well

Your argument settled it: the motivating case is `kornia.geometry.transform.pyramid`, which the inventory does not cover, so the guard as written would not have caught the breakage it was written for. I reproduced your numbers — **132 modules declare `__all__`, 0 fail to import, 0 unbound** — and switched the parametrization to discovery.

```
tests/test_api_surface.py   144 passed in 0.43s      (12 inventory + 132 swept)
kornia/ at merge-base       2 failed, 142 passed     (kornia.color, kornia.core)
```

So the two reported defects stay pinned while the guard now watches the whole tree.

I also wrapped the `hasattr` probe in `catch_warnings` for the point you raised — it was firing the `kornia.utils` `CachedDownloader` `DeprecationWarning` on every run purely by looking, and a probe that guards `__all__` should not be emitting warnings of its own.

Pushed as `7df73f8`.

hiyufan:
Heads-up on a force-push, which I would rather not do mid-review but this one was mechanical.

The two failing checks were both `error during mergeable check` rather than real failures: the CHANGELOG entry I added landed on the same first bullet of `## Unreleased` → `### Breaking changes` that `main` had meanwhile filled with the `extract_patches_from_pyramid` entry, so the branch went `CONFLICTING` and pre-commit.ci could not build a merge commit to run against.

Rebased onto `main` (33 commits ahead) and kept both bullets, `main`'s first and mine after it. Nothing else in the resolution — no code or test change between `7df73f81` and `04e625f6`.

Re-ran on the new base, since a third of a hundred commits is enough to move the sweep out from under me:

```
tests/test_api_surface.py   144 passed in 0.22s
modules declaring __all__   132
still-broken modules        0
```

Same numbers as before the rebase, and `kornia.geometry.transform.pyramid` is still swept and still outside the inventory.

ducha-aiki:
## Delta re-review of #4143 at `04e625f6` — both asks landed; two CI items, one of which is ours

Setup: worktree at the PR head, `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU/macOS.

### Both review asks are addressed

1. **The `AUTUMN` history is corrected**, and the new version is accurate down to the commits — added in #1996 (`50f783d2`), deprecated since 0.7.2, removed by #3432 (`9e521cac`) leaving the `__all__` entry behind. That is the version of the argument the stability policy actually blesses.
2. **The release-note entry is there**, under `## Unreleased`, and it says the thing that matters: nothing that works today stops working, because the entry named a binding that no longer existed.

You also took the non-blocking suggestion and discovered the modules instead of parametrizing over the inventory. `tests/test_api_surface.py` now runs **144 passed** (was 22), covering all 132 modules that declare `__all__`, with zero failures — so `kornia.geometry.transform.pyramid`, the ComfyUI-LTXVideo module from #3986/#4026 that the inventory does not cover, is watched now. The `warnings.catch_warnings()` wrapper is the right handling of the `hasattr`-fires-`__getattr__` property; that was a note, not an ask, and you closed it properly.

### CI item 1 — yours, one line

`pre-commit.ci` fails on ruff `S112`:

```text
S112 `try`-`except`-`continue` detected, consider logging the exception
  --> tests/test_api_surface.py:89:9
```

`contextlib.suppress` clears it without a `noqa` and without changing behaviour. Verified in the worktree — `ruff check` clean, `ruff format --check` clean, `144 passed`:

```python
import contextlib  # with the other stdlib imports

    found = []
    for info in pkgutil.walk_packages(kornia.__path__, "kornia."):
        # Optional extras and backends are out of scope for this guard, so a module
        # that cannot be imported here is skipped rather than failing the sweep.
        with contextlib.suppress(Exception):
            mod = importlib.import_module(info.name)
            if getattr(mod, "__all__", None) is not None:
                found.append(info.name)
    return sorted(found)
```

### CI item 2 — ours, not yours: do not change the diff for it

`Check for dropped module-level names` fails with `Removed from __all__: ['AUTUMN']`. That check hard-fails on **any** `__all__` removal; its only exemptions are a module-level `__getattr__` shim and `kornia.contrib`. It has no notion of the procedure `test_no_public_name_removed`'s own assertion message tells you to follow — which is precisely what you did, inventory edit and release note included.

So you have satisfied the documented process and been blocked by a second check that does not know the process exists. That is a gap in our tooling, and I have filed it as **#4190** with three candidate escape hatches. Please leave that check red and do not work around it; a maintainer will resolve it on the merge.

Fix the ruff line and this is done. Thank you for chasing the `AUTUMN` history to the commit rather than defending the original phrasing — the corrected version is a stronger argument for the guard than the one it replaced.

hiyufan:
Fixed in `841bccde` — `contextlib.suppress(Exception)` around the import, `import contextlib` with the other stdlib imports.

That one is on me: I ran `ruff check` on the files I touched in #4140 and on the TVM side, and skipped it here. Reproduced `S112` locally first, then confirmed clean after:

```
ruff check tests/test_api_surface.py          All checks passed!
ruff format --check                           1 file already formatted
tests/test_api_surface.py                     144 passed in 0.23s
```

Sweep is unchanged at 132 modules, so the rewrite is behaviour-preserving as you said.

Understood on `Check for dropped module-level names` — leaving it red and not working around it, and I will not touch `__all__` handling to appease it. Thank you for filing #4190 rather than asking me to paper over it; a check that hard-fails on the documented procedure is worth fixing once at the source instead of once per PR.

ducha-aiki:
## Status on #4143 at `841bccde` — nothing is owed by you; the block is ours

Setup: worktree at the PR head, `~/dev/kornia/.venv/bin/python`, **torch 2.14.0** (the `.venv` moved off 2.9.1 with #4197).

The ruff item is fixed and everything on your side is green:

```text
ruff check tests/test_api_surface.py          All checks passed!
ruff format --check tests/test_api_surface.py 1 file already formatted
pytest tests/test_api_surface.py -q           144 passed
```

The only remaining red check is **`Check for dropped module-level names`**, which is #4190 — our tooling gap, not your diff. As said last round: leave it red, do not work around it. A maintainer resolves it before this merges, and I have moved it to the top of my own queue.

One note on a line I gave you, so it is on the record rather than discovered later. My `contextlib.suppress` snippet moved the `getattr`/`append` *inside* the suppressed block, where the original `try`/`except`/`continue` only covered the import. That widens the catch: a module that imports cleanly but whose module-level `__getattr__` raises on `__all__` would now be dropped from the sweep silently, shrinking the guard's coverage instead of failing. I checked whether it costs anything today and it does not:

```text
wide (as shipped):  132 modules
narrow (import-only suppression):  132 modules
identical: True      only-in-narrow: []
modules with a module-level __getattr__: ['kornia.utils']
```

`kornia.utils` is the only shim and its `__getattr__` does not raise for `__all__`, so the two are indistinguishable. **No change wanted** — I am recording it because the widening is mine, not yours, and if that sweep ever starts under-reporting this is the first place to look.

Nothing for you to do. Thank you for two clean rounds, and for correcting the `AUTUMN` history to the commit rather than defending the first version.

hiyufan:
Understood — leaving `Check for dropped module-level names` red for #4190, and not touching `__all__` handling to make it green.

On the `contextlib.suppress` widening: thank you for putting it on the record. I had not noticed that moving `getattr`/`append` inside the block widened the catch beyond the import, and your reading is right — a module that imports cleanly but whose module-level `__getattr__` raises on `__all__` would drop out of the sweep silently, which is the failure mode this guard exists to prevent, one level up.

Taking "no change wanted" at face value and leaving it as shipped, since your check shows the two forms are indistinguishable today and `kornia.utils` is the only shim. If the sweep ever reports fewer than the modules that declare `__all__`, narrowing the suppression back to the import alone is the first thing to try:

```python
mod = None
with contextlib.suppress(Exception):
    mod = importlib.import_module(info.name)
if mod is not None and getattr(mod, "__all__", None) is not None:
    found.append(info.name)
```

That is still `S112`-clean, so the narrow form costs nothing whenever you want it.

ducha-aiki:
@hiyufan — the red check on this PR is ours, not yours, and it is now fixed: **#4230** implements #4190's
escape hatch.

I A/B'd it against **this branch specifically** (`841bccde`, merge-base `4b69473a`) rather than a synthetic case:

```text
main's script:  ::error file=kornia/color/__init__.py::Removed from __all__: ['AUTUMN']     exit=1
#4230:          ::notice file=kornia/color/__init__.py::Removed from __all__ and recorded
                in tests/api_surface.json in this same change ...: ['AUTUMN']               exit=0
```

The acknowledgement is your `tests/api_surface.json` edit — the same edit
`tests/test_api_surface.py::test_no_public_name_removed` told you to make. You did everything right; the check
simply had no notion of the procedure its sibling test documents.

Once #4230 merges, rebase (or just re-run the check) and `Check for dropped module-level names` goes green with
a `::notice`. Nothing to change here in the meantime — thank you for the patience over three rounds.

hiyufan:
Thanks for fixing it at the source rather than asking me to work around it, and for A/B'ing #4230 against this branch instead of a synthetic case — the `::notice` line reading back my own `tests/api_surface.json` edit as the acknowledgement is exactly the shape the sibling test documents.

Holding here. `841bccde` is 35 green and the one red:

```text
Check for dropped module-level names   failure     <- #4190, fixed by #4230
everything else (35 check-runs)        success
```

I will rebase onto `main` and confirm the check flips to a green `::notice` once #4230 lands. If it does not, I will say so here rather than reaching for `__all__` to make it green.

No patience needed on my side — three rounds where two of them found real problems in my patch and the third found one in the tooling is a good ratio.

hiyufan:
#4230 landed, so I rebased — head is `a57845dc` — and the check is green:

```text
Check for dropped module-level names   success
```

Locally against the same base, your escape hatch reads back exactly the acknowledgement the sibling test asked me to make:

```text
$ python3 .github/scripts/check_import_surface.py --base-ref origin/main
::notice file=kornia/color/__init__.py::Removed from __all__ and acknowledged in
  tests/api_surface.json in this same change: ['AUTUMN']
exit=0
```

### The rebase

83 commits, one conflict, in `kornia/core/__init__.py`: `main` added `from .safetensors import load_safetensors` on the line my `from .ops import eye_like, vec_like` was on. Kept both, in module order, and `__all__` had already auto-merged to hold all three names:

```python
from .module import ImageModule, ImageSequential
from .ops import eye_like, vec_like
from .safetensors import load_safetensors
from .tensor_wrapper import TensorWrapper
```

`import kornia.core` resolves `eye_like`, `vec_like` and `load_safetensors` after it. The CHANGELOG entry rebased without a conflict this time.

The patch against the new base is the same five files:

```text
CHANGELOG.md              |  6 ++++++
kornia/color/__init__.py  |  1 -
kornia/core/__init__.py   |  1 +
tests/api_surface.json    |  1 -
tests/test_api_surface.py | 53 +++++++++++++++++++++++++++++++++++++++++++++++
```

`tests/test_api_surface.py`: **147 passed** — up from the count before the rebase, because it now also parametrizes `main`'s new `test_inventory_is_a_mapping_of_module_to_public_names`, which passes alongside my sweep. `ruff check` and `ruff format --check` clean on all three touched Python files.

ducha-aiki (approved):
Delta re-review: approved at `a57845dc`. The rebase retains both the restored `eye_like`/`vec_like` imports and main's `load_safetensors` import. The package-wide public-name guard is preserved; the removed `AUTUMN` entry is acknowledged in the inventory.

Validation in an isolated checkout, Python 3.11 / torch 2.14.0: `tests/test_api_surface.py tests/color --device=cpu --dtype=float32,float64` -> 1012 passed, 3 skipped. `check_import_surface.py --base-ref origin/main` exits 0 and emits the expected acknowledgement notice for AUTUMN. Current CI, including the import-surface check, is green; GitHub reports mergeable. The previous tooling/rebase blocker is resolved.

PR #4144, opened 2026-09-01
Fix submodule mutation in SIFT forwards

### LINKED ISSUE #4069: SIFTDescriptor / PatchDominantGradientOrientation / PatchAffineShapeEstimator keep kernels as plain attributes: .to() misses them and forward mutates the module

X-mo7:

Removes the remaining forward-time submodule mutations from #4069. SIFTDescriptor.pk, DenseSIFTDescriptor.bin_pooling_kernel, DenseSIFTDescriptor.PoolingConv, and PatchDominantGradientOrientation.angular_smooth's weights were cast by calling .to() on the submodules inside forward, permanently changing their dtype/device based on the last input. The fix replaces those mutations with local weight/bias casts and functional convolutions. A float32 module can still process a float64 input, while the stored module parameters remain float32. Circular padding for angular_smooth is preserved explicitly. Regression tests cover dtype/device preservation, numerical equivalence, and gradient flow.

Fixes #4069 & follow-up to #4079

## What did you check?
- Reproduced the pre-fix mutation for all four convolution submodules.
- Focused regression tests: 4 passed, 3 skipped.
- Relevant SIFT/orientation test suite: 81 passed, 7 skipped, 13 deselected.
- git diff --check passes.
- Checked that float64 inputs still produce float64 outputs while stored convolution weights remain float32.
- Checked numerical equivalence with torch.equal and gradient flow to the stored parameters.

## How did AI help?

I used AI (Codex) to inspect the part of the code, diagnose the changes, and run focused tests. I reproduced the bug before the implementation manually and reviewed the resulting diff and regression coverage before submitting.

### DIFF
--- a/kornia/feature/orientation.py
+++ b/kornia/feature/orientation.py
@@ -105,10 +105,12 @@ class PatchDominantGradientOrientation(nn.Module):
                 f"input shape should be must be [Bx1x{self.patch_size}x{self.patch_size}]. Got {patch.size()}"
             )
         # cast into a local; rebinding `self.weighting` would make the buffer's dtype/device depend on
-        # whichever tensor was passed last, and is a `torch.compile` guard hazard. (The `self.angular_smooth`
-        # line below rebinds a submodule to itself -- `nn.Module.to` is in-place -- which is a different case.)
+        # whichever tensor was passed last, and is a `torch.compile` guard hazard.
         weighting = self.weighting.to(patch.dtype).to(patch.device)
-        self.angular_smooth = self.angular_smooth.to(patch.dtype).to(patch.device)
+        angular_smooth_weight = self.angular_smooth.weight.to(patch.dtype).to(patch.device)
+        angular_smooth_bias = self.angular_smooth.bias
+        if angular_smooth_bias is not None:
+            angular_smooth_bias = angular_smooth_bias.to(patch.dtype).to(patch.device)
         grads: torch.Tensor = self.gradient(patch)
         # unpack the edges
         gx: torch.Tensor = grads[:, :, 0]
@@ -134,7 +136,23 @@ class PatchDominantGradientOrientation(nn.Module):
             )
             ang_bins_list.append(ang_bins_i)
         ang_bins = torch.cat(ang_bins_list, 1).view(-1, 1, self.num_ang_bins)
-        ang_bins = self.angular_smooth(ang_bins).view(-1, self.num_ang_bins)
+        angular_smooth_padding = self.angular_smooth.padding
+        if self.angular_smooth.padding_mode != "zeros":
+            ang_bins = F.pad(
+                ang_bins,
+                (angular_smooth_padding[0], angular_smooth_padding[0]),
+                mode=self.angular_smooth.padding_mode,
+            )
+            angular_smooth_padding = (0,)
+        ang_bins = F.conv1d(
+            ang_bins,
+            angular_smooth_weight,
+            angular_smooth_bias,
+            self.angular_smooth.stride,
+            angular_smooth_padding,
+            self.angular_smooth.dilation,
+            self.angular_smooth.groups,
+        ).view(-1, self.num_ang_bins)
         values, indices = ang_bins.max(1)
         indices_left = (self.num_ang_bins + indices - 1) % self.num_ang_bins
         indices_right = (indices + 1) % self.num_ang_bins
--- a/kornia/feature/siftdesc.py
+++ b/kornia/feature/siftdesc.py
@@ -209,7 +209,10 @@ class SIFTDescriptor(nn.Module):
         """
         KORNIA_CHECK_SHAPE(input, ["B", "1", f"{self.patch_size}", f"{self.patch_size}"])
         B: int = input.shape[0]
-        self.pk = self.pk.to(input.dtype).to(input.device)
+        pk_weight = self.pk.weight.to(input.dtype).to(input.device)
+        pk_bias = self.pk.bias
+        if pk_bias is not None:
+            pk_bias = pk_bias.to(input.dtype).to(input.device)
 
         grads = spatial_gradient(input, "diff")
         # unpack the edges
@@ -229,7 +232,15 @@ class SIFTDescriptor(nn.Module):
 
         ang_bins = torch.cat(
             [
-                self.pk((bo0_big == i).to(input.dtype) * wo0_big + (bo1_big == i).to(input.dtype) * wo1_big)
+                F.conv2d(
+                    (bo0_big == i).to(input.dtype) * wo0_big + (bo1_big == i).to(input.dtype) * wo1_big,
+                    pk_weight,
+                    pk_bias,
+                    self.pk.stride,
+                    self.pk.padding,
+                    self.pk.dilation,
+                    self.pk.groups,
+                )
                 for i in range(self.num_ang_bins)
             ],
             1,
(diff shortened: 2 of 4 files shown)


ducha-aiki:
## Review of #4144 — APPROVE. Closes #4069 completely, and the rewrite is numerically exact

Setup: worktree at the PR head `fb2f846f`, with `kornia/` swapped to the merge-base `ab855287` for every A/B and fail-on-base pass. `~/dev/kornia/.venv/bin/python` (torch 2.9.1), CPU + MPS on Apple silicon. Every measurement block asserts `"/wt/4144/" in kornia.__file__` — my first attempt at the A/B ran a script from `/tmp` and silently imported the primary editable checkout on both sides, so the numbers below come from the re-run where both arms are provably the worktree.

### The rewrite is byte-exact

Same seeds, same weights, base versus head:

```
key                              BASE                    HEAD                    verdict
('sift',   float32)              tensor(4, 128)          tensor(4, 128)          BYTE-IDENTICAL
('orient', float32)              tensor(4,)              tensor(4,)              BYTE-IDENTICAL
('dense',  float32)              tensor(2, 128, 13, 13)  tensor(2, 128, 13, 13)  BYTE-IDENTICAL
('sift',   float64)              tensor(4, 128)          tensor(4, 128)          BYTE-IDENTICAL
('orient', float64)              tensor(4,)              tensor(4,)              BYTE-IDENTICAL
('dense',  float64)              tensor(2, 128, 13, 13)  tensor(2, 128, 13, 13)  BYTE-IDENTICAL
```

That covers the part I was most suspicious of. `angular_smooth` is the one conv with `padding_mode='circular'` (`padding=(2,)`, `bias=False`), so it is the only module that takes your explicit `F.pad` branch — and its output is bit-for-bit unchanged. Your replacement mirrors `nn.Conv1d.forward`'s own non-zeros branch exactly (pad with `_reversed_padding_repeated_twice`, then convolve with padding `0`), which is why it lands on the same bits. The other three convs are `padding_mode='zeros'` and skip that branch entirely. All four have `bias=False` today, so the `is not None` guards are defensive rather than exercised — correct to include anyway.

### The defect is fixed on both axes, and the base shows it

dtype, from the same A/B run — a `float64` forward on a `float32` module:

```
('PARAM sift stays f32',   float64)   BASE torch.float64                  HEAD torch.float32   *** CHANGED ***
('PARAM orient stays f32', float64)   BASE torch.float64                  HEAD torch.float32   *** CHANGED ***
('PARAM dense stays f32',  float64)   BASE (float64, float64)             HEAD (float32, float32) *** CHANGED ***
```

device, on real MPS hardware — a CPU module given an MPS input:

```
BASE:  SIFTDescriptor / PatchDominantGradientOrientation / DenseSIFTDescriptor
       params before={'cpu'} after={'mps'}          <- the module is permanently migrated
HEAD:  params before={'cpu'} after={'cpu'}  out.device=mps
```

Gradients still reach every stored parameter (`pk.weight`, `angular_smooth.weight`, `bin_pooling_kernel.weight`, `PoolingConv.weight` — all non-zero after `backward()`), and an explicit `.to(torch.float64)` still behaves normally (`weight=float64`, `out=float64`).

### It closes all of #4069, not just the three convs

The issue names `PatchAffineShapeEstimator` too. Checking every parameter and buffer of all four classes for mutation across a forward:

```
PatchAffineShapeEstimator          tensors=1  mutated_by_forward=none
SIFTDescriptor                     tensors=2  mutated_by_forward=none
PatchDominantGradientOrientation   tensors=2  mutated_by_forward=none
DenseSIFTDescriptor                tensors=4  mutated_by_forward=none
```

`PatchAffineShapeEstimator` was already handled by #4079 (`register_buffer` plus a local `weighting` cast), so with this PR the `Fixes #4069` line is accurate. A grep for the `self.X = self.X.to(...)` pattern across `kornia/feature/*.py` returns nothing at the head.

### The #4130 test inversion is right

You modified `TestDenseSIFTDescriptor::test_pooling_kernel_accessor_is_isolated_and_tracks_to`, which #4130 landed a few hours before this PR opened. Its old assertion was

```python
assert kernel.dtype == torch.float64
assert kernel.device == device
```

i.e. it pinned the mutation as expected behaviour. Removing the mutation necessarily falsifies it, and you inverted it to `assert (kernel.dtype, kernel.device) == before` while **keeping** the `mod.to(device, target_dtype)` half that gives the test its name. That is exactly the right surgery, and it is the kind of thing that is easy to get wrong when two PRs land a day apart — good.

### Tests

```
fail-on-base (kornia/ at ab855287), 4 new tests, all four fail with the right assertion:
  test_forward_uses_input_dtype_without_mutating_pooling_layer       assert torch.float64 == torch.float32
  test_pooling_kernel_accessor_is_isolated_and_tracks_to             (float64, cpu) != (float32, cpu)
  test_forward_uses_input_dtype_without_mutating_convolution_layers  assert torch.float64 == torch.float32
  test_forward_uses_input_dtype_without_mutating_angular_smooth      assert torch.float64 == torch.float32

head, CPU, float32 + float64:  93 passed, 8 skipped
head, MPS, float32:            68 passed, 21 skipped, 2 failed
```

The two MPS failures are `TestSIFTConstantPatchIsFinite::test_l2_normalize_zero_vector_has_a_zero_gradient[mps-desc_dtype3]` and `::test_float16_subnormal_norm_still_normalises_to_one[mps]`, both `TypeError: Cannot convert a MPS Tensor to float64`. They fail identically on the merge-base (`2 failed, 26 passed`), so they are pre-existing and unrelated.

### `torch.compile`: no regression, and one blocker removed

```
BASE  all three: fullgraph FAIL (torch.* op returned non-Tensor)   default OK
HEAD  all three: fullgraph FAIL (Data dependent operator)          default OK
```

Neither revision compiles `fullgraph=True`, so nothing regresses. Worth noting what changed though: on the base the first failure is *"torch.* op returned non-Tensor"* — that is the `self.angular_smooth = self.angular_smooth.to(...)` rebinding itself, the guard hazard the comment in `orientation.py` already warned about. At the head that failure is gone and dynamo gets further before hitting a pre-existing `aten._local_scalar_dense` data-dependent op. So this PR removes one of the graph-break causes on the way to `fullgraph`.

### Optional, non-blocking

The four cast-and-convolve blocks are near-identical and add most of the +166. A small module-level helper taking `(conv, x, dtype, device)` would collapse them and make the circular-padding branch exist once rather than being a special case in one of four copies. Entirely your call — the current form is explicit and correct, and I would not hold the PR for it.

### Verdict

**APPROVE.** Byte-identical output, the mutation gone on both dtype and device with the base measured as the contrast, gradients intact, `Fixes #4069` accurate across all four classes, a correct inversion of the neighbouring #4130 pin, and a small `torch.compile` improvement as a side effect.

ducha-aiki (approved):

PR #4145, opened 2026-09-01
test: make half precision references backend-safe

### LINKED ISSUE #4137: TestTransformLAFs::test_transform_points fails in float16/bfloat16 because the test itself calls torch.inverse
## What fails

`TestTransformLAFs::test_transform_points` fails for every batch/point parametrization on CPU `float16` and `bfloat16` (18 node failures per run):

```
RuntimeError: linalg.inv: Low precision dtypes not supported. Got BFloat16
```

## Root cause — the test, not the library

`kornia.feature.perspective_transform_lafs` never inverts anything. The failure comes from the test's own round-trip construction at `tests/feature/test_laf.py:922`:

```python
lafs_dst = kornia.feature.perspective_transform_lafs(dst_homo_src, lafs_src)
src_homo_dst = torch.inverse(dst_homo_src)   # <- no fp16/bf16 CPU kernel
lafs_dst_to_src = kornia.feature.perspective_transform_lafs(src_homo_dst, lafs_dst)
self.assert_close(lafs_src, lafs_dst_to_src)
```

CPU `linalg.inv` has no half-precision kernel, so the test raises before the function under test is ever exercised in those dtypes. The failures are long-standing and unrelated to any recent LAF change (confirmed identical on `main` and on the #4122 branch, CPU and CUDA `--isolate-half-precision` runs).

## Proposed fix: hardcoded inputs and expected outputs

Per the repo testing convention (AGENTS.md: prefer a hardcoded literal plus the small generation snippet), replace the runtime `torch.inverse` round-trip with fixed numerical data:

- a small literal batch of LAFs and a literal homography (well-conditioned, nothing near the dtype boundaries),
- the literal expected transformed LAFs, generated once in `float64` and pasted into the test with the generation snippet in a comment,
- the test then just runs `perspective_transform_lafs` in the injected `device`/`dtype` and `assert_close`s against the literal cast to that dtype.

This removes the `torch.inverse` dependency entirely, so the test runs in all four dtypes on every backend, and it stops re-deriving the reference through the function's own output (the current round-trip would also pass for e.g. a sign-flipped implementation, since the error cancels on the way back).

Alternatives considered: `_torch_inverse_cast` or `_inverse_3x3_closed_form` (`kornia/core/utils.py`) in the test would also clear the failures, but the hardcoded reference is preferred because it is self-contained and independent of the op under test.

The half-precision round-trip *property* (transform + inverse-transform ≈ identity) can stay covered by the float32/float64 parametrizations if we keep the existing test alongside for those dtypes, or be dropped in favor of the literal check — implementer's choice.

---

ducha-aiki:
## Summary

- replace test-side inverse operations with analytic projective expectations and Lie-group identities
- avoid unsupported NumPy conversions for BF16 and upcast only SVD validation where PyTorch lacks half support
- use deterministic, non-degenerate reduced-precision fixtures while preserving projective and batching coverage

Closes #4137

## Test plan

- [x] `pixi run pre-commit-all`
- [x] affected tests on CPU with float16, bfloat16, float32, and float64 (236 passed)
- [x] affected tests on CUDA with isolated float16 and bfloat16 subprocesses (118 passed)

### DIFF
--- a/tests/feature/test_laf.py
+++ b/tests/feature/test_laf.py
@@ -1528,20 +1528,54 @@ class TestTransformLAFs(BaseTester):
     @pytest.mark.parametrize("batch_size", [1, 2, 5])
     @pytest.mark.parametrize("num_points", [2, 3, 5])
     def test_transform_points(self, batch_size, num_points, device, dtype):
-        # generate input data
-        eye_size = 3
-        lafs_src = torch.rand(batch_size, num_points, 2, 3, device=device, dtype=dtype)
+        laf_bank = torch.tensor(
+            [[[0.4, 0.1, 1.0], [-0.2, 0.3, 2.0]], [[-0.3, 0.2, -1.0], [0.1, 0.5, 0.5]]],
+            device=device,
+            dtype=dtype,
+        )
+        homography_bank = torch.tensor(
+            [
+                [[1.2, 0.1, 0.5], [-0.2, 0.9, 0.3], [0.02, -0.01, 1.0]],
+                [[0.8, -0.15, -0.2], [0.05, 1.1, 0.4], [-0.015, 0.025, 1.0]],
+            ],
+            device=device,
+            dtype=dtype,
+        )
+        # Closed-form projection of the two LAFs by the two homographies, generated in float64:
+        # p' = (H[:2, :2] @ p + H[:2, 2]) / (H[2, :2] @ p + H[2, 2]).
+        expected_bank = torch.tensor(
+            [
+                [
+                    [[0.4366336634, 0.1520520521, 1.9], [-0.2762376238, 0.2521521522, 1.9]],
+                    [
+                        [-0.3663911846, 0.2970568104, -0.6666666667],
+                        [0.1620046620, 0.4219449271, 0.9743589744],
+                    ],
+                ],
+                [
+                    [
+                        [0.3449105525, 0.0319508833, 0.2898550725],
+                        [-0.1678083484, 0.3070486851, 2.5603864734],
+                    ],
+                    [
+                        [-0.2394165288, 0.0915517577, -1.0462287105],
+                        [0.0859048943, 0.5319950165, 0.8759124088],
+                    ],
+                ],
+            ],
+            device=device,
+            dtype=dtype,
+        )
+        batch_indices = torch.arange(batch_size, device=device) % len(homography_bank)
+        point_indices = torch.arange(num_points, device=device) % len(laf_bank)
+        lafs_src = laf_bank[point_indices].unsqueeze(0).expand(batch_size, -1, -1, -1)
+        dst_homo_src = homography_bank[batch_indices]
+        expected = expected_bank[batch_indices[:, None], point_indices[None, :]]
 
-        dst_homo_src = create_random_homography(lafs_src, eye_size)
-        # transform the points from dst to ref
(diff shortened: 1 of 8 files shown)


ducha-aiki (commented):
No blocking findings.\n\nI reviewed the test-only changes against issue #4137 and verified the replacement references preserve the intended projective, batching, Lie-group, and reduced-precision coverage.\n\nLocal verification:\n- 236 CPU cases passed across float16, bfloat16, float32, and float64.\n- 118 isolated CUDA cases passed across float16 and bfloat16.\n- The hardcoded LAF expectations independently matched a float64 projective derivation (maximum absolute difference below 5e-11).\n- git diff --check passed.\n\nThe remaining risk is limited to the broader CI matrix; all completed checks were passing at review time.

ducha-aiki:
## Review

**LGTM.** Test-only change; the approach (analytic references instead of test-side `torch.inverse`/NumPy round-trips) is the right fix for half-precision backends, and it matches the repo preference for deterministic hardcoded references.

### What I verified

- **Golden values are independently correct.** I recomputed the `expected_bank` in `tests/feature/test_laf.py` from scratch in float64 by projecting the LAF center and the two axis endpoints (`center + A[:, i]`) through each homography and rebuilding the frame from the projected point differences — all 4 combinations match the hardcoded literals to 10 decimals, so this is a genuine closed-form reference, not a pin of the implementation's own output.
- **The hardcoded `pix_trans_norm`** in `test_grid.py` is the correct analytic inverse of `normal_transform_pixel(2, 4)`.
- **The analytic expectation in `test_linalg.py`** is algebraically consistent with the constructed homography; the perspective denominator stays in `[0.5, 1.5]`, so no near-degenerate configurations for any parametrization.
- **CPU:** affected tests across `float16,bfloat16,float32,float64` — 565 passed.
- **CUDA half (`--isolate-half-precision`):** affected tests — 296 passed.
- **The fix is real:** on `main`, `tests/geometry/test_grid.py` under CUDA float16/bfloat16 isolation gives 3 failed + 1 skipped; on this branch the same run is 64/64 passed with the `"inverse_cuda" not implemented for "Half"` skip gone.
- `pixi run lint` passes; `create_random_homography` imports are still used by the gradcheck tests; no lingering `#4137` references in `kornia/` or `tests/`.

### Observations (non-blocking)

1. In my mixed multi-file CUDA half run, 3 *non-half* tests (`TestJointBilateralBlur::test_exception`, `test_normalized_meshgrid_integer_dtype_promotes_uniformly[cuda-...]`, `TestTransformPoints::test_gradcheck[cuda]`) errored with pytest's "previous item was not torn down properly". All three pass standalone, none is touched by this PR, and single-file isolation runs are clean on both branch and main — this is an ordering artifact of the `--isolate-half-precision` subprocess plugin when an isolated test is followed by a non-isolated one across files. Worth a separate issue against `conftest.py`, not this PR.
2. `tests/geometry/test_grid.py`: a one-line comment noting the hardcoded matrix is the inverse of `normal_transform_pixel(2, 4)` would help future editors (the roundtrip assert does self-check, so this is purely a readability nicety).
3. `tests/feature/test_laf.py`: the generation comment describes point projection; the 2×2 frame part is obtained by projecting `center` and `center + column` and differencing. A half-sentence saying so would make the literal fully reproducible from the comment alone.

Issue #3704, opened 2026-04-24
[Bug]: RandomGaussianIllumination / RandomLinearCornerIllumination crash on CUDA input — generator device stays on CPU


boschet:
### 🐛 Describe the bug

RandomGaussianIllumination and RandomLinearCornerIllumination raise RuntimeError: Expected all tensors to be on the same device, but found at least two     devices, cuda:0 and cpu! on CUDA input.
                                                                                                                                                              Their internal parameter generators (GaussianIlluminationGenerator, LinearCornerIlluminationGenerator) have self.device == 'cpu' regardless of input
  device. The gradient tensor returned by generate_parameters therefore lives on CPU. When apply_transform then does                                        
  return input.add_(params["gradient"]).clamp_(0, 1)

  (_2d/intensity/gaussian_illumination.py:170, analogous line in linear_illumination.py), the dispatcher rejects the CUDA×CPU combination.                  
  This bug is silent for some seeds and loud for others because broadcasting over a zero-element tensor used to mask the device mismatch: when the            per-sample to_apply Bernoulli mask selects zero active samples, params["gradient"] has shape (0, C, H, W) — adding a zero-element CPU tensor to a CUDA      tensor used to no-op silently on some PyTorch versions but now raises consistently.
                                                                                                                                                              Related: the TODO comment at kornia/augmentation/random_generator/_2d/gaussian_illumination.py:133 ("this will crash if the shape is not batched")
  acknowledges adjacent shape-edge-case issues (#3580).                                                                                                           

### 🔄 Steps to Reproduce

```bash
1. Put a RandomGaussianIllumination or RandomLinearCornerIllumination with p < 1.0 inside an AugmentationSequential.
  2. Feed a CUDA tensor into it.                                                                                                                              3. Run with small batch size (e.g. B=1) across multiple seeds. ~15% of seeds (at p=0.15, B=1) reproduce it; the exact rate is 1 − (1 − p)^B.
```

### 💻 Minimal Code Example

```python
import torch
  import kornia.augmentation as K
                                                                                                                                                              aug = K.AugmentationSequential(
      K.RandomGaussianIllumination(gain=(0.02, 0.15), p=0.15),                                                                                                    K.RandomLinearCornerIllumination(gain=(0.02, 0.15), p=0.15),                                                                                            )                                                                                                                                                           x = torch.rand(1, 3, 32, 32, device="cuda")                                                                                                                                                                                                                                                                             fails = 0
  for seed in range(30):
[...]

PR #4147, opened 2026-09-01
test: fix false-red MPS coverage

### LINKED ISSUE #4146: MPS: fix 57 false-red tests caused by test construction and parametrization
## Problem

The eager MPS/float32 suite still contains tests that fail because the test itself asks MPS to do something unsupported or compares MPS results against a CPU-specific oracle. These failures obscure genuine Kornia/MPS compatibility defects.

This is deliberately separate from #4033: that issue is about adding CI coverage; this one is the test-only cleanup needed to make its MPS signal useful.

## Reproduction

At `11fd2721` (`main` on 2026-09-01):

```bash
KORNIA_TEST_DEVICE=mps \
KORNIA_TEST_DTYPE=float32 \
KORNIA_TEST_OPTIMIZER='' \
pixi run test-quick -q --tb=short
```

Environment:

```text
kornia 0.9.0rc1
torch 2.9.1 (5811a8d7da873dd699ff6687092c225caffcf1bb)
python 3.11.14
macOS 26.5.1 arm64
MPS available
```

Full result:

```text
536 failed, 7707 passed, 4103 skipped, 32 xfailed, 10 xpassed
3042.35s (0:50:42)
```

The 536 raw failures include real implementation/backend compatibility problems. This issue tracks only the **57 failing node IDs whose root cause is in test construction or parametrization**.

### 1. Tests create float64 tensors on MPS (19 nodes)

Every one fails before the intended implementation behavior can be tested:

```text
TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64.
```

- [ ] `tests/augmentation/container/test_augmentation_sequential.py::TestAugmentationSequential::test_mixed_image_bbox_dtypes[mps-image_dtype2]` — 1 node; the parametrization includes `torch.float64` and allocates the image directly on MPS.
- [ ] `tests/feature/test_siftdesc.py::TestSIFTConstantPatchIsFinite::test_l2_normalize_zero_vector_has_a_zero_gradient[mps-desc_dtype3]` — 1 node.
- [ ] `tests/feature/test_siftdesc.py::TestSIFTConstantPatchIsFinite::test_float16_subnormal_norm_still_normalises_to_one[mps]` — 1 node; an unrelated comparison loop includes float64.
- [ ] `tests/geometry/test_conversions.py::TestAngleAxisToQuaternion::test_convention_float_input_keeps_its_dtype_3948[mps-input_dtype3]` — 1 node.
- [ ] `tests/geometry/test_conversions.py::TestConvertPointsFromHomogeneous::test_convention_the_masked_division_keeps_the_gradient_finite[mps]` — 1 node; explicitly hardcodes float64 at line 3064 even though float32 preserves the exact-pole regression on MPS.
- [ ] `tests/geometry/test_linalg.py::TestTransformPoints::test_mixed_dtypes[...]` — 5 MPS parameter combinations containing `points_dtype2` or `trans_dtype2`.
- [ ] `tests/geometry/solvers/test_homogeneous.py::TestNullVector3x4::{test_residual_near_zero,test_residual_random_batch,test_direction_recovery}` — 9 nodes; `_make_rank3_matrix` always transfers the fixture to MPS float64.

Fix: filter/skip only the MPS+float64 cells, or keep a CPU float64 oracle and transfer a valid float32 case when the test's purpose permits it. Do not skip supported MPS dtypes with the same blanket marker.

### 2. Device-dependent RNG compared with CPU-derived literals (2 nodes)

`TESTING.md` documents that seeded CPU and MPS generators do not produce the same sequence. These generator tests sample on MPS and compare against CPU-derived literals:

- [ ] `tests/augmentation/test_random_generator.py::TestColorJitterGen::{test_random_gen,test_same_on_batch}` — 2 nodes. The generator itself documents that its samples are not reproducible across devices/dtypes.

Fix in the order already prescribed by `TESTING.md`: inject deterministic params / generate the oracle inputs on CPU and move them, test backend-independent properties, or skip only non-CPU exact-sequence assertions. Loosening tolerances will not fix different random samples.

### 3. Known-unsupported MPS interpolation cells are still parametrized (33 nodes)

- [ ] `tests/filters/test_motion.py::TestMotionBlur3D::test_smoke[...]` — 24 `params_as_tensor=True, mode=nearest` nodes fail with `grid_sampler_3d: Unsupported Nearest interpolation`.
- [ ] `tests/filters/test_motion.py::TestMotionBlur3D::test_get_motion_kernel3d[...]` — 8 `params_as_tensor=True` nodes reach the same unsupported 3D-nearest kernel.
- [ ] `tests/feature/test_xfeat.py::TestInterpolateSparse2d::test_cardinality[mps-float32-bicubic]` — 1 direct interpolator test selects bicubic and fails with `MPS: Unsupported Bicubic interpolation`; the supported bilinear cardinality cell must remain exercised.

Fix: filter or skip only these known-unsupported cells with precise reasons. `TESTING.md` already identifies 3D-nearest as unsupported on MPS. This bucket intentionally does **not** include the six broader public `TestXFeat` failures, because those exercise XFeat's default public path and therefore need a separate implementation/support decision.

### 4. CPU-specific numerical/state oracles (3 nodes)

- [ ] `tests/contrib/test_conv_distance_transformer.py::TestConvDistanceTransform::test_offset_parenthesis_fix[mps-float32]` — `h=0.01` makes the oracle depend on `exp(-100)`: CPU preserves the float32 subnormal (`3.78e-44`), while MPS flushes it to zero. Reparameterize the regression test away from subnormals. The broader silent-underflow behavior is tracked separately in #4152.
- [ ] `tests/core/test_helpers.py::TestSolveCast::test_smoke[mps-float32]` — device-side `randn` selects a poorly conditioned system and compares an unnormalised residual against a fixed absolute tolerance. Use a deterministic, conditioned system and a relative residual.
- [ ] `tests/geometry/test_conversions.py::TestQuaternionToRotationMatrix::test_wart_normalize_quaternion_scales_by_norm_over_eps_3952[mps-float32]` — this wart pin is test-order dependent because of the documented PyTorch MPS scalar-clamp cache behavior. It passes alone. Reuse the existing MPS clamp-cache runtime probe/skip before asserting the CPU wart.

## Explicit exclusions

Do not hide these under test skips as part of this issue:

- The 19 augmentation nodes originally attributed to device-dependent RNG — `RandomCrop3D` (6), `RandomRotation3D` (1), `RandomPerspective` (1), `RandomMixUpV2` (3), `RandomCutMixV2` (4), `RandomMosaic` (1), `RandomJigsaw` (1), `RandomPlanckianJitter` (1), and `RandomMedianBlur` (1) — fail before oracle comparison because `batch_prob` remains on CPU while transformed tensors are on MPS. This genuine cross-device regression is tracked in #4151 and must remain visible here.
- YUV empty-input failures, the broader XFeat public path, homography/FitLine/SSIM numerical failures, RandomAffine and 3D augmentation transform failures, and other shape/invariant failures. Their traces enter Kornia implementation paths or expose a public MPS support question, so they need separate diagnosis/fixes.

## Acceptance criteria

- The 57 audited nodes no longer report false-red failures in an eager MPS/float32 run.
- Supported dtype/device cells remain exercised.
- Exact RNG assertions are device-independent; no tolerance-only workaround for different samples.
- Known unsupported parameter cells have narrow, documented skips rather than broad class/file skips.
- Genuine implementation/backend failures, including #4151, remain visible.

ducha-aiki:
## Summary

- skip only parameter cells whose required operation is unavailable, using cached runtime capability probes that auto-enable when PyTorch adds support
- keep the two genuine device- and dtype-dependent `ColorJitterGenerator` exact-sequence assertions restricted to CPU float32
- preserve MPS tensor-parameter coverage for 3D motion kernels by falling back from unsupported nearest interpolation to probed bilinear interpolation
- keep the distance-transform small-`h` underflow visible with a strict #4152 xfail for half precision and MPS float32
- retain deterministic numerical fixtures and use old-PyTorch-safe CPU dtype construction
- leave the 19 augmentation device-mismatch failures visible under #4151 instead of misclassifying them as RNG false reds

## Validation

- focused CPU float16/bfloat16/float32 review matrix: 350 passed, 5 skipped, 2 expected #4152 xfailed
- focused MPS float32 review matrix: 91 passed, 27 capability/oracle skips, 1 expected #4152 xfailed
- all eight MPS tensor-parameter `get_motion_kernel3d` cells pass through the bilinear fallback
- independent final diff review: no actionable findings
- `pixi run pre-commit-all` passed

Closes #4146

### DIFF
--- a/testing/base.py
+++ b/testing/base.py
@@ -137,7 +137,11 @@ def supports_2d_border_padding(device: torch.device) -> bool:
     return _supports_2d_border_padding_probe(device.type)
 
 
-_UNIMPLEMENTED_KERNEL_MSG = "not implemented for"
+_UNSUPPORTED_KERNEL_MSGS = (
+    "not implemented for",
+    "Unsupported Bicubic interpolation",
+    "Unsupported Nearest interpolation",
+)
 
 
 class _UnsupportedProbeDtype(Exception):
@@ -172,10 +176,55 @@ def _supports_kernel_probe(op: Callable[[str, torch.dtype], object], device_type
     except _UnsupportedProbeDtype:
         return False
     except RuntimeError as e:
-        if _UNIMPLEMENTED_KERNEL_MSG in str(e):
+        if any(message in str(e) for message in _UNSUPPORTED_KERNEL_MSGS):
             return False
         raise
     return True
+
+
+def _bicubic_2d_grid_sample_op(device_type: str, dtype: torch.dtype) -> None:
+    inp = _probe_zeros(device_type, dtype, 1, 1, 2, 2)
+    grid = _probe_zeros(device_type, dtype, 1, 1, 1, 2)
+    F.grid_sample(inp, grid, mode="bicubic", align_corners=True)
+
+
+def supports_bicubic_2d_grid_sample(device: torch.device, dtype: torch.dtype) -> bool:
+    """Whether this device supports bicubic interpolation in 2D ``grid_sample`` for ``dtype``.
+
+    Probed at runtime and cached per (device type, dtype), so tests guarded by this helper
+    auto-enable once PyTorch adds the missing interpolation kernel.
+    """
+    return _supports_kernel_probe(_bicubic_2d_grid_sample_op, device.type, dtype)
+
+
+def _bilinear_3d_grid_sample_op(device_type: str, dtype: torch.dtype) -> None:
+    inp = _probe_zeros(device_type, dtype, 1, 1, 2, 2, 2)
+    grid = _probe_zeros(device_type, dtype, 1, 1, 1, 1, 3)
+    F.grid_sample(inp, grid, mode="bilinear", align_corners=True)
+
+
+def supports_bilinear_3d_grid_sample(device: torch.device, dtype: torch.dtype) -> bool:
+    """Whether this device supports bilinear interpolation in 3D ``grid_sample`` for ``dtype``.
+
+    Probed at runtime and cached per (device type, dtype), so tests guarded by this helper
+    auto-enable once PyTorch adds the missing interpolation kernel.
+    """
+    return _supports_kernel_probe(_bilinear_3d_grid_sample_op, device.type, dtype)
+
+
+def _nearest_3d_grid_sample_op(device_type: str, dtype: torch.dtype) -> None:
+    inp = _probe_zeros(device_type, dtype, 1, 1, 2, 2, 2)
+    grid = _probe_zeros(device_type, dtype, 1, 1, 1, 1, 3)
+    F.grid_sample(inp, grid, mode="nearest", align_corners=True)
+
+
+def supports_nearest_3d_grid_sample(device: torch.device, dtype: torch.dtype) -> bool:
+    """Whether this device supports nearest interpolation in 3D ``grid_sample`` for ``dtype``.
+
+    Probed at runtime and cached per (device type, dtype), so tests guarded by this helper
+    auto-enable once PyTorch adds the missing interpolation kernel.
+    """
+    return _supports_kernel_probe(_nearest_3d_grid_sample_op, device.type, dtype)
 
 
 def _replicate_padding_op(device_type: str, dtype: torch.dtype) -> None:
--- a/testing/base.py
+++ b/testing/base.py
@@ -173,13 +173,28 @@ def _supports_kernel_probe(op: Callable[[str, torch.dtype], object], device_type
     """
     try:
         op(device_type, dtype)
-    except _UnsupportedProbeDtype:
+    except (_UnsupportedProbeDtype, NotImplementedError):
         return False
     except RuntimeError as e:
         if any(message in str(e) for message in _UNSUPPORTED_KERNEL_MSGS):
             return False
         raise
     return True
+
+
+def _bilinear_2d_grid_sample_op(device_type: str, dtype: torch.dtype) -> None:
+    inp = _probe_zeros(device_type, dtype, 1, 1, 2, 2)
+    grid = _probe_zeros(device_type, dtype, 1, 1, 1, 2)
+    F.grid_sample(inp, grid, mode="bilinear", align_corners=True)
+
+
+def supports_bilinear_2d_grid_sample(device: torch.device, dtype: torch.dtype) -> bool:
+    """Whether this device supports bilinear interpolation in 2D ``grid_sample`` for ``dtype``.
+
+    Probed at runtime and cached per (device type, dtype), so tests guarded by this helper
+    auto-enable once PyTorch adds the missing interpolation kernel.
+    """
+    return _supports_kernel_probe(_bilinear_2d_grid_sample_op, device.type, dtype)
 
 
 def _bicubic_2d_grid_sample_op(device_type: str, dtype: torch.dtype) -> None:
--- a/tests/augmentation/_3d/test_random_crop_3d.py
+++ b/tests/augmentation/_3d/test_random_crop_3d.py
@@ -39,6 +39,8 @@ class TestRandomCrop3D(BaseTester):
 
     @pytest.mark.parametrize("batch_size", [1, 2])
     def test_no_padding(self, batch_size, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(42)
         input_tensor = torch.tensor(
             [
@@ -115,6 +117,8 @@ class TestRandomCrop3D(BaseTester):
 
     @pytest.mark.parametrize("padding", [1, (1, 1, 1), (1, 1, 1, 1, 1, 1)])
     def test_padding_batch(self, padding, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(42)
         batch_size = 2
         input_tensor = torch.tensor(
@@ -144,6 +148,8 @@ class TestRandomCrop3D(BaseTester):
         self.assert_close(out, expected, atol=1e-4, rtol=1e-4)
 
     def test_pad_if_needed(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(42)
         input_tensor = torch.tensor([[[0.0, 1.0, 2.0]]], device=device, dtype=dtype)
         expected = torch.tensor(
--- a/tests/augmentation/_3d/test_random_crop_3d.py
+++ b/tests/augmentation/_3d/test_random_crop_3d.py
@@ -39,8 +39,6 @@ class TestRandomCrop3D(BaseTester):
 
     @pytest.mark.parametrize("batch_size", [1, 2])
     def test_no_padding(self, batch_size, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(42)
         input_tensor = torch.tensor(
             [
@@ -117,8 +115,6 @@ class TestRandomCrop3D(BaseTester):
 
     @pytest.mark.parametrize("padding", [1, (1, 1, 1), (1, 1, 1, 1, 1, 1)])
     def test_padding_batch(self, padding, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(42)
         batch_size = 2
         input_tensor = torch.tensor(
@@ -148,8 +144,6 @@ class TestRandomCrop3D(BaseTester):
         self.assert_close(out, expected, atol=1e-4, rtol=1e-4)
 
     def test_pad_if_needed(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(42)
         input_tensor = torch.tensor([[[0.0, 1.0, 2.0]]], device=device, dtype=dtype)
         expected = torch.tensor(
--- a/tests/augmentation/_3d/test_random_rotation_3d.py
+++ b/tests/augmentation/_3d/test_random_rotation_3d.py
@@ -42,8 +42,6 @@ class TestRandomRotation3D(BaseTester):
 
     def test_random_rotation(self, device, dtype):
         # This is included in doctest
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)  # for random reproductibility
 
         f = RandomRotation3D(degrees=45.0)
--- a/tests/augmentation/_3d/test_random_rotation_3d.py
+++ b/tests/augmentation/_3d/test_random_rotation_3d.py
@@ -42,6 +42,8 @@ class TestRandomRotation3D(BaseTester):
 
     def test_random_rotation(self, device, dtype):
         # This is included in doctest
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)  # for random reproductibility
 
         f = RandomRotation3D(degrees=45.0)
--- a/tests/augmentation/container/test_augmentation_sequential.py
+++ b/tests/augmentation/container/test_augmentation_sequential.py
@@ -112,6 +112,8 @@ class TestAugmentationSequential:
     def test_mixed_image_bbox_dtypes(self, device, image_dtype):
         # Regression test for https://github.com/kornia/kornia/issues/3705 and #3706:
         # bbox stays in fp32 while the image uses a half/double compute dtype.
+        if device.type == "mps" and image_dtype == torch.float64:
+            pytest.skip("MPS does not support float64")
         if device.type == "cpu" and image_dtype in (torch.float16, torch.bfloat16) and torch_version_lt(2, 6, 0):
             pytest.skip("PyTorch <2.6 has no CPU Half/BFloat16 grid_sample kernel")
         torch.manual_seed(0)
--- a/tests/augmentation/test_augmentation.py
+++ b/tests/augmentation/test_augmentation.py
@@ -5045,8 +5045,6 @@ class TestPlanckianJitter(BaseTester):
         self.assert_close(f(input), expected, low_tolerance=True)
 
     def test_planckian_jitter_batch(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)
         input = self._get_input(device, dtype).repeat(2, 1, 1, 1)
 
@@ -5249,8 +5247,6 @@ class TestRandomMedianBlur(BaseTester):
         assert image.shape == aug(image).shape
 
     def test_feature_median_blur(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)
 
         img = torch.ones(1, 1, 4, 4, device=device, dtype=dtype)
--- a/tests/augmentation/test_augmentation.py
+++ b/tests/augmentation/test_augmentation.py
@@ -5045,6 +5045,8 @@ class TestPlanckianJitter(BaseTester):
         self.assert_close(f(input), expected, low_tolerance=True)
 
     def test_planckian_jitter_batch(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)
         input = self._get_input(device, dtype).repeat(2, 1, 1, 1)
 
@@ -5247,6 +5249,8 @@ class TestRandomMedianBlur(BaseTester):
         assert image.shape == aug(image).shape
 
     def test_feature_median_blur(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)
 
         img = torch.ones(1, 1, 4, 4, device=device, dtype=dtype)
--- a/tests/augmentation/test_augmentation_mix.py
+++ b/tests/augmentation/test_augmentation_mix.py
@@ -40,8 +40,6 @@ class TestRandomMixUpV2(BaseTester):
         assert str(f) == repr, str(f)
 
     def test_random_mixup_p1(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)
         f = RandomMixUpV2(p=1.0, data_keys=["input", "class"])
 
@@ -83,8 +81,6 @@ class TestRandomMixUpV2(BaseTester):
         self.assert_close(out_label[:, 2], lam, rtol=1e-4, atol=1e-4)
 
     def test_random_mixup_lam0(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)
         f = RandomMixUpV2(lambda_val=(0.0, 0.0), p=1.0, data_keys=["input", "class"])
 
@@ -104,8 +100,6 @@ class TestRandomMixUpV2(BaseTester):
         self.assert_close(out_label[:, 2], lam, rtol=1e-4, atol=1e-4)
 
     def test_random_mixup_same_on_batch(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)
         f = RandomMixUpV2(same_on_batch=True, p=1.0, data_keys=["input", "class"])
 
@@ -151,8 +145,6 @@ class TestRandomCutMixV2(BaseTester):
         assert str(f) == expected_repr
 
     def test_random_mixup_p1(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(76)
         f = RandomCutMixV2(p=1.0, data_keys=["input", "class"], use_correct_lambda=True)
 
@@ -195,8 +187,6 @@ class TestRandomCutMixV2(BaseTester):
         self.assert_close(out_label, exp_label)
 
     def test_random_mixup_beta0(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(76)
         # beta 0 => resample 0.5 area
         # beta cannot be 0 after torch 1.8.0
@@ -230,8 +220,6 @@ class TestRandomCutMixV2(BaseTester):
         )
 
     def test_random_mixup_num2(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(76)
         f = RandomCutMixV2(num_mix=5, p=1.0, data_keys=["input", "class"], use_correct_lambda=True)
 
@@ -269,8 +257,6 @@ class TestRandomCutMixV2(BaseTester):
         )
 
     def test_random_mixup_same_on_batch(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(42)
         f = RandomCutMixV2(same_on_batch=True, p=1.0, data_keys=["input", "class"], use_correct_lambda=True)
 
@@ -324,8 +310,6 @@ class TestRandomMosaic(BaseTester):
         assert str(f) == repr
 
     def test_numerical(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(76)
         f = RandomMosaic(p=1.0, data_keys=["input", "bbox_xyxy"])
 
@@ -436,8 +420,6 @@ class TestRandomJigsaw(BaseTester):
         f(input)
 
     def test_numerical(self, device, dtype):
-        if device.type != "cpu":
-            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(22)
         f = RandomJigsaw(grid=(2, 2), p=1.0, data_keys=["input"])
 
--- a/tests/augmentation/test_augmentation_mix.py
+++ b/tests/augmentation/test_augmentation_mix.py
@@ -40,6 +40,8 @@ class TestRandomMixUpV2(BaseTester):
         assert str(f) == repr, str(f)
 
     def test_random_mixup_p1(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)
         f = RandomMixUpV2(p=1.0, data_keys=["input", "class"])
 
@@ -81,6 +83,8 @@ class TestRandomMixUpV2(BaseTester):
         self.assert_close(out_label[:, 2], lam, rtol=1e-4, atol=1e-4)
 
     def test_random_mixup_lam0(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)
         f = RandomMixUpV2(lambda_val=(0.0, 0.0), p=1.0, data_keys=["input", "class"])
 
@@ -100,6 +104,8 @@ class TestRandomMixUpV2(BaseTester):
         self.assert_close(out_label[:, 2], lam, rtol=1e-4, atol=1e-4)
 
     def test_random_mixup_same_on_batch(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(0)
         f = RandomMixUpV2(same_on_batch=True, p=1.0, data_keys=["input", "class"])
 
@@ -145,6 +151,8 @@ class TestRandomCutMixV2(BaseTester):
         assert str(f) == expected_repr
 
     def test_random_mixup_p1(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(76)
         f = RandomCutMixV2(p=1.0, data_keys=["input", "class"], use_correct_lambda=True)
 
@@ -187,6 +195,8 @@ class TestRandomCutMixV2(BaseTester):
         self.assert_close(out_label, exp_label)
 
     def test_random_mixup_beta0(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(76)
         # beta 0 => resample 0.5 area
         # beta cannot be 0 after torch 1.8.0
@@ -220,6 +230,8 @@ class TestRandomCutMixV2(BaseTester):
         )
 
     def test_random_mixup_num2(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(76)
         f = RandomCutMixV2(num_mix=5, p=1.0, data_keys=["input", "class"], use_correct_lambda=True)
 
@@ -257,6 +269,8 @@ class TestRandomCutMixV2(BaseTester):
         )
 
     def test_random_mixup_same_on_batch(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(42)
         f = RandomCutMixV2(same_on_batch=True, p=1.0, data_keys=["input", "class"], use_correct_lambda=True)
 
@@ -310,6 +324,8 @@ class TestRandomMosaic(BaseTester):
         assert str(f) == repr
 
     def test_numerical(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(76)
         f = RandomMosaic(p=1.0, data_keys=["input", "bbox_xyxy"])
 
@@ -420,6 +436,8 @@ class TestRandomJigsaw(BaseTester):
         f(input)
 
     def test_numerical(self, device, dtype):
+        if device.type != "cpu":
+            pytest.skip("Random parameters are device-dependent; expected values were computed on CPU")
         torch.manual_seed(22)
         f = RandomJigsaw(grid=(2, 2), p=1.0, data_keys=["input"])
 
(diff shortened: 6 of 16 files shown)


ducha-aiki:
## Review: request changes

Most of this PR is exactly right, but **19 of the 21 nodes in bucket 2 of #4146 are not a test-construction problem** — they are the symptoms of one genuine kornia cross-device defect, and skipping them with the reason `"Random parameters are device-dependent; expected values were computed on CPU"` states something that is provably false and hides an implementation bug. That trips #4146's own acceptance criterion *"Genuine implementation/backend failures remain visible."*

Everything else in the PR I verified and it is good. Details and measurements below.

### Setup for every number in this review

```
KORNIA_TEST_DEVICE=mps KORNIA_TEST_DTYPE=float32 KORNIA_TEST_OPTIMIZER=''
python -m pytest -p no:randomly <the 15 files this PR touches>
torch 2.9.1 | python 3.11.14 | macOS arm64 | MPS available
```

| revision | failed | passed | skipped |
| --- | --- | --- | --- |
| `11fd2721` (base) | 360 | 1828 | 225 |
| `acc346d5` (this PR) | 285 | 1839 | 289 |

Node-level transitions: 12 `FAILED→PASSED`, 63 `FAILED→SKIPPED`, 1 `PASSED→SKIPPED`, 0 new failures. That matches the PR description.

---

## Blocking: the bucket-2 skips mislabel a real bug

### 1. Those nodes never reach an oracle comparison

At `11fd2721`, on MPS, every one of them fails with

```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, mps:0 and cpu!
```

raised at four sites:

- `kornia/augmentation/base.py:360` — `_blend_by_prob` → `torch.where` (planckian jitter, median blur)
- `kornia/augmentation/_2d/base.py:139` (perspective)
- `kornia/augmentation/_3d/base.py:105` (crop3d, rotation3d)
- `kornia/augmentation/_2d/mix/base.py:113` (mixup, cutmix, mosaic, jigsaw)

The exception is thrown *before* any expected-value comparison, so "the expected values were computed on CPU" cannot be why they fail.

Root cause: `params["batch_prob"]` is produced by the random generator, whose RNG device defaults to CPU by design, while the branches being blended live on the input's device. `torch.where` requires the condition on the same device for a non-0-dim tensor, so the blend raises.

```python
import torch
from kornia.augmentation import RandomPlanckianJitter, RandomMixUpV2
x = torch.rand(2, 3, 4, 4, device="mps")

RandomPlanckianJitter(p=0.5)(x)            # RuntimeError
RandomPlanckianJitter(p=0.5).to("mps")(x)  # OK
RandomMixUpV2(p=1.0, data_keys=["input", "class"])(x, torch.tensor([0, 1], device="mps"))            # RuntimeError
RandomMixUpV2(p=1.0, data_keys=["input", "class"]).to("mps")(x, torch.tensor([0, 1], device="mps"))  # OK
```

This is device-generic, not MPS-specific: the same `torch.where` device rule holds on CUDA, so the `device.type != "cpu"` guard hides it there too — which matters precisely because #4033 is about turning that coverage on.

### 2. It is a regression, and it is large

`git blame` puts all four sites in **9986de7e — `refactor(augmentation): makes it onnx exportable` (#3722, 2026-05-23)**, which replaced a device-agnostic Python branch (`transformed if bool(to_apply.any()) else not_transformed`) with a `torch.where` blend.

Running the same three files at `9986de7e^` (`697d662d`) on MPS:

| revision | `test_augmentation_mix.py` + `test_random_crop_3d.py` + `test_perspective_rand.py` |
| --- | --- |
| `697d662d` (before #3722) | **2 failed, 201 passed** |
| `11fd2721` (today) | 25 failed |

Every mixup / cutmix / mosaic / jigsaw / crop3d / perspective node this PR skips **used to pass on MPS against those very CPU-derived literals**. The literals are device-independent, because the RNG is CPU by design. The two nodes that failed at `697d662d` are `TestRandomTransplantation::{test_repeating, test_data_keys}` — unrelated.

Across just the 15 files this PR touches, **293 of the base revision's 360 MPS failures are this single RuntimeError.** This PR silences 19 of them and leaves 274 red, so it does not remove a false-red class — it removes an arbitrary 6% slice of one bug's symptoms, and leaves siblings of the same tests (`TestRandomCrop3D::test_same_on_batch`, `TestRandomJigsaw::test_smoke`, `TestRandomMosaic::test_p`, `test_partial_batch_passthrough`, …) failing for the identical reason.

### 3. A four-line fix makes the skipped tests pass exactly as written

```diff
--- a/kornia/augmentation/base.py
+++ b/kornia/augmentation/base.py
         if transformed.shape == not_transformed.shape and transformed.shape[0] == to_apply.shape[0]:
+            to_apply = to_apply.to(transformed.device)
             to_apply_expanded = to_apply.view(-1, *([1] * (len(transformed.shape) - 1)))
```

plus the same one-liner at the three other blend sites (`_2d/base.py`, `_3d/base.py`, `_2d/mix/base.py`). That is what I measured, not necessarily where the fix belongs — harmonising the params once in `forward_parameters` (or moving `batch_prob` when the input device is known) may be better than four per-call `.to()` hops, which land inside the compiled graph. Whoever takes the issue should pick.

Measured on top of `11fd2721`, same 15 files:

| | failed | passed | skipped |
| --- | --- | --- | --- |
| base | 360 | 1828 | 225 |
| this PR | 285 | 1839 | 289 |
| **base + the 4-line fix** | **65** | **2123** | **225** |

CPU stays fully green with the fix: `2286 passed, 121 skipped, 29 xfailed, 5 xpassed`.

And the 27 nodes of the classes this PR skips (including all 19) **pass on MPS with their existing CPU literals** once the device bug is fixed — no tolerance change, no reparametrization, no skip.

### What I'd like changed here

1. Drop the `device.type != "cpu"` skips from:
   - `tests/augmentation/_3d/test_random_crop_3d.py` (`test_no_padding`, `test_padding_batch`, `test_pad_if_needed`)
   - `tests/augmentation/_3d/test_random_rotation_3d.py::test_random_rotation`
   - `tests/augmentation/test_augmentation.py` (`TestPlanckianJitter::test_planckian_jitter_batch`, `TestRandomMedianBlur::test_feature_median_blur`)
   - `tests/augmentation/test_augmentation_mix.py` (all 9)
   - `tests/augmentation/test_perspective_rand.py::test_transform_module_should_return_expected_transform`
2. **Keep** the two in `tests/augmentation/test_random_generator.py`. `TestColorJitterGen::{test_random_gen, test_same_on_batch}` fail with `AssertionError: Tensor-likes are not close!` (greatest abs difference 0.026) both before and after #3722 — those are the real device-RNG class, and the rewritten reason is an improvement on the old `TODO(jian)` text.
3. Open the augmentation device-mismatch issue against #3722 and fix it separately (related but distinct from the closed #3704). Once it lands, those 19 nodes go green on their own.
4. #4146's bucket 2 needs correcting from 21 nodes to 2; the other 19 belong in its "Explicit exclusions" list, and the headline count drops from 76 to 57.

I'm not asking for the implementation fix in this PR — #4146 is deliberately test-only. I am asking that this PR not paint over it.

---

## Verified correct (no action needed)

- **Bucket 1, float64 skips** — precise, and each is exactly the failing cell. `test_linalg.py::test_mixed_dtypes` skips 5 nodes, `test_conversions.py` 1 + 1, `test_siftdesc.py` 1, `test_augmentation_sequential.py` 1.
- **Bucket 3, motion 3D-nearest** — `test_smoke` skips exactly the 24 `params_as_tensor=True, mode="nearest"` nodes and `test_get_motion_kernel3d` exactly the 8 `params_as_tensor=True` nodes. `tests/filters/test_motion.py` goes 32 MPS failures → 0 with **no** over-skipping: `params_as_tensor=False` builds its tensors on CPU, so those cells legitimately still run.
- **xfeat bicubic** — narrow, and the six broader `TestXFeat` bicubic failures correctly stay red.
- **`_make_rank3_matrix` CPU fixture** — the 9 `TestNullVector3x4` nodes now genuinely *pass* on MPS rather than skip.
- **`_skip_if_mps_clamp_caching` reuse** — right helper, matching its sibling use at `test_wart_the_eps_0_zero_quaternion...`. Worth noting it turns one node (`test_wart_normalize_quaternion_scales_by_norm_over_eps_3952[mps-float32]`) from order-dependently-passing into always-skipped on MPS; that is the only `PASSED→SKIPPED` transition in the PR, and given the torch clamp-cache defect it is the right call.
- **`test_offset_parenthesis_fix`, `h=0.01 → 0.1`** — I re-injected the original bug (`offset: int = i * kernel_size // 2`) on top of this PR: the test fails at both float32 and bfloat16 with a max abs difference of 2.0, so the pin still discriminates what it was written for. The bfloat16 tolerance bump is necessary and not padding: at `1e-3` bfloat16 fails with an abs difference of 0.0039 against a `0.002` budget.
- **`TestSolveCast` CPU fixture** — fine, and now deterministic across backends. Unrelated: `test_smoke[cpu-bfloat16]` already fails on `main` (`0.1357` residual); pre-existing, out of scope here.

## Nits

- The same two-line skip is copy-pasted 12 times. If only the two `test_random_generator.py` ones survive, this is moot; if any others stay, a shared helper next to `_skip_if_mps_clamp_caching` would keep the reason text in one place.
- `tests/feature/test_siftdesc.py`: the conditional-tuple for `comparison_dtypes` reads awkwardly. `comparison_dtypes = (torch.bfloat16, torch.float32) if device.type == "mps" else (torch.bfloat16, torch.float32, torch.float64)` on one line, or building the tuple and dropping `float64` on MPS, is easier to scan.

---

ducha-aiki:
## Second pass — four more verified points

A deeper pass over the same diff turned up four things worth acting on, all measured on this branch (`acc346d5`, torch 2.9.1, MPS). None of them changes the verdict in my previous comment; the bucket-2 skips are still the blocking item.

### 1. `h=0.01 → 0.1` walks away from a silent wrong answer, untracked

`distance_transform` does not merely lose precision in that regime — it returns **all zeros**, with no error and no warning:

```
                        column 0 of the output
cpu  float32   h=0.01   [0.9998, 0.0, 0.9998, 1.9998, 2.9998, 3.9998, 4.9998, 5.9998]
mps  float32   h=0.01   [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]      <-- silent
cpu  bfloat16  h=0.01   [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]      <-- silent
cpu  float32   h=0.1    [0.9969, 0.0, 0.9969, 1.9969, ...]
mps  float32   h=0.1    [0.9969, 0.0, 0.9969, 1.9969, ...]
```

`kernel = exp(-dist / h)` underflows, `filter2d` returns 0, `log(0)` is `-inf`, and `nan_to_num(..., neginf=0.0)` turns the whole field into zeros. So it is not MPS-specific — CPU bfloat16 does it too, which is why `test_offset_parenthesis_fix[cpu-bfloat16]` also stops exercising it after this hunk.

The `h` change itself is fine and I verified the pin still discriminates. What is missing is that after this diff nothing in the suite covers the `exp(-dist/h)` underflow regime, and there is no `xfail`, no tracking issue, and no note in `kornia/contrib/distance_transform.py`. A user calling the documented API with a small `h` on MPS, or in bfloat16 anywhere, silently gets a zero distance field. That deserves its own issue (and per repo policy, a fix or an explicit pin), not a quiet reparametrization.

### 2. `test_convention_the_masked_division_keeps_the_gradient_finite` does not need the MPS skip

The pin still discriminates at float32, on MPS:

```
        convert_points_from_homogeneous(pts, eps=1e-8).sum().backward(), pts = [[2., 4., -1e-8]]
cpu float32   fixed: [[1.0, 1.0, 0.0]]   unfixed 1/(z+eps): [[1.0, 1.0, nan]]
mps float32   fixed: [[1.0, 1.0, 0.0]]   unfixed 1/(z+eps): [[1.0, 1.0, nan]]
```

`-1e-8` and `eps=1e-8` round to the same float32 value, so `z_vec + eps` is still exactly the pole. Making the hardcoded dtype device-appropriate (float32 on MPS, float64 elsewhere) keeps the #3938 pin alive there; `pytest.skip` deletes it.

The reason string is also wrong on its face: the test signature is `(self, device)` with no `dtype` fixture — the float64 is hardcoded at line 3064. "the float64 fixture required by this regression test" will mislead the next reader.

### 3. `TestSolveCast::test_smoke` is a conditioning problem, not an MPS one

The comment added by this PR says the MPS sample "can be much more poorly conditioned than the CPU one", which implies the backend is at fault. Feeding the *same* MPS-drawn matrix to `_torch_solve_cast` on CPU:

```
mps-drawn A (max cond 4.10e3)   resid on mps 1.0797e-4   resid on cpu 1.1360e-4
cpu-drawn A (max cond 5.06e1)   resid           1.6828e-6
```

CPU fails the `atol=1e-4` bound on that matrix too. The real defect is that the test compares an unnormalised `torch.dist(B, A @ X)` against a fixed absolute tolerance, which any ill-conditioned draw breaks. A CPU seed just re-rolls the dice — scaling the residual by `B.norm()`, or building a conditioned `A`, fixes the class. The neighbours `TestSolveWithMask::test_smoke` and `TestSvdCast::test_smoke` still use `torch.randn(..., device=device)` with the same seed and shapes and keep the same latent flakiness.

### 4. `TestInterpolateSparse2d::test_cardinality` — only the bicubic cell is unsupported

```
InterpolateSparse2d(mode)(x, pos, H, W) on mps
  bilinear  OK  (2, 20, 16)
  nearest   OK  (2, 20, 16)
  bicubic   RuntimeError: MPS: Unsupported Bicubic interpolation
```

`test_cardinality` is the class's only `(B, N, C)` shape assertion — `test_smoke` only covers bilinear at `(1, 5, 32)` with all-zero positions. Parametrizing `mode` over bilinear/bicubic and skipping just the bicubic cell keeps the contract on MPS instead of dropping it.

### Also worth a line

- `test_mixed_image_bbox_dtypes` does **not** go green with the float64 skip: `[mps-image_dtype0]`, `[mps-image_dtype1]` and `[mps-image_dtype3]` still fail with the device-mismatch `RuntimeError` from my previous comment. The #3705/#3706 regression stays red on MPS and loses its float64 cell — another reason to fix the device bug rather than skip around it.
- Reinforcing the clamp-cache guard: `test_wart_normalize_quaternion_scales_by_norm_over_eps_3952[mps-float32]` passes in a single-file run but **fails when the whole `tests/geometry` directory runs** (25 failed / 2158 passed on `main`, this node among them). So `_skip_if_mps_clamp_caching` is guarding a real ordering hazard, not a hypothetical one — my endorsement of that hunk stands.
- I also considered folding the four inline `mps and float64` guards into `conftest.py`'s `DEVICE_DTYPE_BLACKLIST`. It does not apply: all four are `pytest.mark.parametrize` dtype *values* (`trans_dtype`/`points_dtype`, `input_dtype`, `desc_dtype`, `image_dtype`), not the `dtype` fixture the blacklist keys on. The inline guards are the right shape.

---

ducha-aiki:
Filed the device-mismatch bug as #4151, with the bisect to #3722, the raise sites, the suite-wide numbers (`tests/augmentation` on MPS: **449 of 457 failures** are that one bug) and the measured 4-line fix (457 failed → 11 failed, CPU untouched).

That is the blocker for the 19 bucket-2 skips in this PR. Once #4151 lands they go green on their own with their existing CPU literals — no skip, no tolerance change.

ducha-aiki:
Filed the `distance_transform` underflow as #4152 — the point from my second-pass comment. It turned out broader than the `h=0.01` case in this diff: at the **documented default `h=0.35`**, float16 with `kernel_size >= 15` already returns 15.70 where the true peak is 13.51, and partial underflow produces plausible-but-wrong finite distances before it collapses to zeros. The `h=0.01 -> 0.1` hunk here is fine as written; #4152 covers what it leaves untested.

ducha-aiki:
Addressed the review in `a8b9a047`:

- Removed all 19 disputed augmentation skips. A focused MPS run confirms all 19 still fail before oracle comparison with the #4151 CPU/MPS `torch.where` device mismatch.
- Kept only the two genuine `TestColorJitterGen` device-RNG skips.
- Kept the masked-division regression active on MPS using float32; it passes on the fixed implementation and the old formula still produces a NaN gradient.
- Reworked `TestSolveCast` around a reproducible, well-conditioned system and a normalized residual, with a tight tolerance derived from `torch.finfo(dtype).eps`.
- Parametrized XFeat cardinality over bilinear and bicubic so bilinear remains exercised on MPS while only bicubic skips.
- Simplified the SIFT comparison-dtype construction.
- Corrected #4146 online from 76 to 57 audited false-red nodes and moved the 19 #4151 failures into explicit exclusions. The distance-transform underflow is tracked in #4152.

Validation after the changes:

- Revised MPS audit plus supported companion cells: 104 passed, 44 skipped.
- Explicit #4151 visibility check: 19 failed with the expected device mismatch.
- CPU sweep of all 15 touched files: 2,287 passed, 121 skipped, 29 xfailed, 5 xpassed.
- Focused CPU all-dtype checks: 14 passed.
- `pixi run pre-commit-all` passed.

ducha-aiki:
## Review of PR #4147

Overall: a well-constructed test-quality PR. Verified on real MPS hardware that the capability claims, new expected values, tolerances, and the float32-on-MPS pole pin are correct; the parenthesis regression stays pinned at h=0.1; and the #4151 failures are intentionally left visible per the PR body. Five findings survive verification — one correctness bug in a rewritten guard, one issue-tracking gap, two coverage-durability items, and one plausible old-torch break.

### 1. [CONFIRMED] ColorJitter guards still admit CPU half dtypes — the exact false-red class this PR fixes
`tests/augmentation/test_random_generator.py:471` (and the twin at line 550): the rewritten skip guards in `TestColorJitterGen.test_random_gen` / `test_same_on_batch` still let CPU float16/bfloat16 cells through, and those fail against the CPU-float32 expected values. Verified by execution: `test_random_gen[cpu-float16]` FAILS ("Tensor-likes are not close!": float32 seed-42 sample 1.1529 vs float16 1.0195) — the generator's own docstring says samples are not reproducible across dtypes. `pixi run test-half` or any `--dtype=float16,bfloat16` run stays red on both tests. The condition should be `device.type != "cpu" or dtype != torch.float32`, matching the new reason string, which already names the dtype dependence it fails to enforce.

### 2. [CONFIRMED] Moving off h=0.01 orphans open issue kornia#4152
`tests/contrib/test_conv_distance_transformer.py:197`: switching `test_offset_parenthesis_fix` from h=0.01 to h=0.1 removes the suite's only small-h/subnormal-kernel execution of `distance_transform`, without referencing the open defect it steers around. `grep -rnE "#4152|issues/4152" kornia/ tests/` finds zero hits at the PR head, so the AGENTS.md issue-pin sweep comes up empty when #4152 is fixed, and a future CPU-side subnormal flush (e.g. inductor FTZ codegen) regresses silently. The comment on the h change should cite #4152, ideally with a small companion pin (xfail or wart test) keeping the underflow regime visible. (The parenthesis regression itself is still pinned — verified: the pre-fix buggy formula produces 28.9969 vs expected 0.9969.)

### 3. [CONFIRMED] test_get_motion_kernel3d skip erases all MPS coverage of the tensor-param path
`tests/filters/test_motion.py:175`: the skip fires for every `params_as_tensor=True` cell on MPS, though a supported mode exists. Probed on real MPS: 3D grid_sample rejects nearest but supports bilinear. The skip is only needed because `get_motion_kernel3d` defaults to `mode="nearest"` (`kornia/filters/kernels_geometry.py:117`); the asserted sum-to-batch_size invariant is mode-independent, so passing `mode="bilinear"` on MPS (as the sibling `test_smoke`'s mode parametrization already shows) would keep a live cell. As written, an MPS-specific regression in device/dtype handling of tensorized angle/direction goes undetected — the false-green mirror of the false-reds this PR fixes.

### 4. [CONFIRMED] Hardcoded `device.type == "mps"` skips bypass the capability-probe pattern
`tests/feature/test_xfeat.py:152` (bicubic 2D grid_sample) and `tests/filters/test_motion.py:166,175` (nearest 3D grid_sample): these hardcode on `device.type == "mps"` instead of the probe framework `testing/base.py` standardized (`_supports_kernel_probe`, `supports_2d_border_padding`, `supports_replicate_padding`), whose stated rationale is that probing "auto-enables once PyTorch adds support". A `supports_bicubic_grid_sample` / `supports_nearest_3d_grid_sample` probe there would follow the established helper; the hardcoded checks recreate the perpetual-skip failure mode this PR is titled after.

### 5. [PLAUSIBLE] `torch.eye(4, dtype=dtype)` breaks CPU bfloat16 on torch <= 2.2.2
`tests/core/test_helpers.py:102`: `torch.eye` has no CPU BFloat16 kernel on torch <= 2.2.2, so the rewritten `TestSolveCast.test_smoke` newly errors at fixture construction ('"eye" not implemented for BFloat16') on a surface the old `torch.randn(..., dtype=dtype)` supported. No CI cell runs bf16, so this bites local old-torch half runs. One-line fix per repo convention: `torch.eye(4).to(dtype)`.

ducha-aiki:
Addressed the new review in `44c74016`:

- Tightened both ColorJitter literal-oracle guards so only CPU float32 executes them; CPU float16/bfloat16 now skip as intended.
- Added cached runtime probes for bicubic 2D and bilinear/nearest 3D `grid_sample`, replacing hardcoded MPS checks.
- Kept all eight MPS tensor-parameter `get_motion_kernel3d` cells live by falling back to bilinear only after probing it; the tests skip only when neither interpolation mode is viable.
- Cited #4152 and added a strict regression xfail covering float16, bfloat16, and MPS float32 small-`h` silent-zero results.
- Changed the conditioned solve fixture to `torch.eye(4).to(dtype)` for old-PyTorch CPU bfloat16 compatibility.

Validation:

- CPU float16/bfloat16/float32 focused matrix: 350 passed, 5 skipped, 2 expected xfailed.
- MPS float32 focused matrix: 91 passed, 27 expected skipped, 1 expected xfailed.
- `pixi run pre-commit-all` passed.
- Independent final diff review found no actionable issues.

ducha-aiki:
## Review of `44c74016` — approve, with one small guard fix to land before merge

This is the third pass; both earlier rounds are addressed at this head and I re-verified every item rather than trusting the delta. The PR now does exactly what #4146 asks: all 57 audited nodes change state, no supported cell is lost, and everything the issue excludes stays red. One new finding survived verification (a one-line guard fix on the #4152 pin), plus a documentation nit.

### Setup

```
torch 2.9.1 | python 3.11.14 | macOS arm64 | MPS available
KORNIA_TEST_OPTIMIZER='' python -m pytest -p no:randomly <the 10 test files this PR touches>
base = 11fd2721 (merge-base), head = 44c74016, each in its own worktree; kornia.__file__ checked per run
```

### Node-level transitions, base → head

| surface | base | head | FAILED→SKIPPED | FAILED→PASSED | PASSED→SKIPPED | new xfail | new pass |
| --- | --- | --- | --- | --- | --- | --- | --- |
| MPS float32 | 85 failed / 1568 passed | 29 failed / 1589 passed | 35 | 21 | 1 | 1 (#4152) | 1 (`test_cardinality[bilinear]`) |
| MPS float16 | 234 failed / 1412 passed | 178 failed / 1434 passed | 35 | 21 | 0 | 1 | 1 |
| CPU float16+bfloat16 | 223 failed / 2976 passed | 216 failed / 2981 passed | 4 (ColorJitter) | 3 | 0 | 2 | 2 |

Zero new failures on any surface. The 57 changed MPS-float32 nodes are exactly the 57 in #4146 (19 float64 + 2 ColorJitter + 33 interpolation + 3 oracles). The one `PASSED→SKIPPED` is `test_wart_normalize_quaternion_scales_by_norm_over_eps_3952[mps-float32]`: it passes when `test_conversions.py` runs alone (which is what my 10-file run did) and **fails when all of `tests/geometry` runs in one process at base** (25 failed, 2158 passed, that node among the failures), which is the clamp-cache ordering defect the helper's docstring describes. So the skip is justified, not a lost pin.

The 29 that remain red on MPS float32 are all in the issue's "Explicit exclusions": 17 × `test_augmentation_sequential` device mismatch (#4151), 6 × `TestXFeat` public bicubic path, 6 × quaternion exp/log numerics.

### Verified

- **Capability probes.** On 2.9.1/MPS the three new probes return `(bicubic2d, bilinear3d, nearest3d) = (False, True, False)` for float32/float16/bfloat16 and `False` for float64 via `_UnsupportedProbeDtype`; the two new message substrings match the live errors exactly (`MPS: Unsupported Bicubic interpolation`, `grid_sampler_3d: Unsupported Nearest interpolation`). On **torch 2.5.1/MPS** 3D `grid_sample` falls back to CPU with a warning instead of raising, so the probes return `True` there and `TestMotionBlur3D` gives 116 passed on both base and head — no old-torch regression from the probe path.
- **`get_motion_kernel3d` bilinear fallback.** The kernel is divided by its own sum before return, so `sum == batch_size` is mode-independent; all 8 MPS tensor-param cells pass at float32 and float16.
- **`TestSolveCast`.** Normalised residual / tolerance ratio is ≤ 0.06 on every cell (cpu: fp16 0.03, bf16 0.06, fp32/fp64 ≈ 0; mps: fp16 0.03, bf16 0.06, fp32 ≈ 0), max cond(A) = 1.38. Comfortable, deterministic, and `torch.eye(4).to(dtype)` sidesteps the ≤ 2.2.2 bf16 `eye` gap.
- **`test_offset_parenthesis_fix` at h=0.1.** Max abs error vs the new literals: fp16 0, bf16 3.9e-3 (one bf16 ulp at 0.9969, inside the 5e-3 atol), fp32/fp64 2.9e-5, identical on CPU and MPS. The pre-fix formula still produces 28.9969 at that cell, so the regression is still pinned.
- **#4152 pin.** `count_nonzero` at h=0.01 is 0 on cpu-fp16, cpu-bf16, mps-fp16, mps-bf16, mps-fp32 and 28 on cpu-fp32/fp64 — the skip guard's dtype/device set matches the measured underflow set exactly, so the strict xfail is honest on 2.9.1.
- **Pole pin on MPS float32.** Fixed gradient `[1, 1, 0]`; the naive `1/(z+eps)` gives `[inf, inf, -inf]` on MPS (NaN on CPU), so the pin still discriminates there.
- **ColorJitter guards** now skip cpu-fp16/bf16 as well as MPS (the round-2 P1); `test_mixed_image_bbox_dtypes`, `test_l2_normalize_*`, `test_convention_float_input_keeps_its_dtype_3948`, `test_mixed_dtypes` skip only the float64 cells and every other cell still runs.
- CI green on all 38 checks at this head; test-only change, so no CHANGELOG entry is due.

### 1. [CONFIRMED, small] The #4152 xfail guards the wrong op — on torch ≤ 2.5.1 CPU float16 it reports FAILED, not xfail

`tests/contrib/test_conv_distance_transformer.py:226` guards with `supports_conv2d(device, dtype)`, but the first kernel `distance_transform` hits is `F.pad(mode="replicate")` inside `filter2d(..., border_type="replicate")`, and CPU float16 `replication_pad2d` is missing through torch 2.5.1 (bfloat16 is fine). Measured in a fresh 2.5.1 venv from the head worktree:

```
PASSED test_offset_parenthesis_fix[cpu-bfloat16]
XFAIL  test_convention_small_h_keeps_nonzero_distances_4152[cpu-bfloat16]
FAILED test_offset_parenthesis_fix[cpu-float16]                        - RuntimeError: "replication_pad2d" not implemented for 'Half'
FAILED test_convention_small_h_keeps_nonzero_distances_4152[cpu-float16] - RuntimeError: "replication_pad2d" not implemented for 'Half'
```

The `RuntimeError` does not match `raises=AssertionError`, so the strict pin comes out as a hard failure instead of the xfail it documents. Not a CI surface today (the scheduled matrix is float32/float64 only and #4154's manifest is 2.9.1), but `pixi run test-half` on an older torch is a documented local run, and this is the same class that turned `main` red after #4098 (#4116). Fix is one line: guard with `supports_replicate_padding(device, dtype)` (already exported from `testing/base.py`) in place of, or alongside, `supports_conv2d`. `test_offset_parenthesis_fix[cpu-float16]` fails the same way on 2.5.1 at base too, so it is pre-existing rather than a regression, but the same guard would make it honest as well.

### 2. [nit] The test-side bilinear fallback contradicts the letter of `TESTING.md` rule 4

`TESTING.md:263` says of 3D nearest on MPS: "Do not silently fall back to bilinear (it would corrupt segmentation masks); instead raise or skip the test." The fallback in `test_get_motion_kernel3d` is the right call here (the invariant is mode-independent, see above), but the next reader will see the rule and the fallback disagree. Either a one-line comment at the fallback saying why the sum invariant makes the mode irrelevant, or rewording rule 4 to scope it to implementation code, would close the gap.

ducha-aiki:
## Follow-up: one correction to my review above, and three verified additions

A second, independent pass over the same diff turned up more than my first comment, and one of its findings shows my own old-torch claim was wrong. Everything below is re-executed on this machine; nothing is taken from the other pass on trust.

### Correction: my "torch 2.5.1 MPS falls back to CPU, no old-torch regression from the probe path" was measured with `PYTORCH_ENABLE_MPS_FALLBACK=1`

My shell exports that variable. With it **unset** (the default for anyone else), the picture on torch 2.5.1 and 2.8.0 MPS is:

```
F.grid_sample(<5-D>, mode="nearest")  ->  NotImplementedError: The operator 'aten::grid_sampler_3d'
                                          is not currently implemented for the MPS device ...
supports_bilinear_3d_grid_sample(mps, float32)  -> RAISES NotImplementedError
supports_nearest_3d_grid_sample(mps, float32)   -> RAISES NotImplementedError
TestMotionBlur3D, MPS float32, head: 56 failed, 60 passed   (base: 56 failed, 60 passed)
```

`NotImplementedError` is a `RuntimeError` subclass, so `_supports_kernel_probe` catches it, but the message says "not **currently** implemented for", which the `"not implemented for"` substring in `_UNSUPPORTED_KERNEL_MSGS` does not match, so the probe re-raises instead of returning `False`. The count is unchanged from base (those 56 cells failed on the op itself before), so it is not a regression, but the new probes break their own bool contract on every torch before 2.9 on MPS, and because `functools.cache` does not memoise a raise, the probe body re-runs for all 56 cells. Two-line fix in `testing/base.py`: catch `NotImplementedError` as "unsupported" before the message match (or add `"not currently implemented for"` to the tuple). That also covers the same gap in the pre-existing probes.

For the record, every MPS number in this thread from this machine, including the earlier rounds and the #4146 audit, was taken with the fallback on. Re-running my 10-file MPS float32 comparison with it unset gives **base 87 failed / 1566 passed → head 31 failed / 1587 passed**, with the identical 35 FAILED→SKIPPED, 21 FAILED→PASSED, 1 PASSED→SKIPPED, 1 new xfail and 1 new pass as before. The only difference from the fallback-on numbers is two `TestSolveWithMask` nodes (`test_smoke`, `test_all_bad`) that fail on **both** base and head once an MPS op with no kernel stops being routed to CPU; they are outside this diff. So the review verdict stands, but the setup line in my comment above should have named the variable.

### 3. [CONFIRMED, small] The new `bilinear` cardinality cell is unguarded on torch ≤ 2.5.1 CPU half

`tests/feature/test_xfeat.py:101`: only the `bicubic` branch consults a probe. On torch 2.5.1 CPU, `grid_sampler_2d_cpu` has no Half or BFloat16 kernel, so the parametrisation swaps the old bicubic red for a new bilinear one:

```
SKIPPED test_cardinality[cpu-float16-bicubic]   This device does not support bicubic interpolation for 2D grid_sample
FAILED  test_cardinality[cpu-float16-bilinear]  RuntimeError: grid_sampler_2d_cpu not implemented for Half
FAILED  test_cardinality[cpu-bfloat16-bilinear] RuntimeError: grid_sampler_2d_cpu not implemented for BFloat16
```

Same count as base (which failed the bicubic cell there), so again not a regression, but the same one-probe fix as item 1 in my first comment: a `supports_bilinear_2d_grid_sample` on the same `_supports_kernel_probe` machinery, consulted on the bilinear branch. `test_smoke` in the class has the identical pre-existing gap.

### 4. [CONFIRMED, pre-existing, touched by this diff] `test_l2_normalize_zero_vector_has_a_zero_gradient[*-bfloat16]` is a per-draw coin flip

The PR edits this test (the float64 skip), and its second half compares bfloat16 gradients of an **unseeded** `torch.rand` draw at the default bf16 tolerance. 20 runs of the `cpu-desc_dtype1` cell on 2.9.1: **3 failed**. Since `desc_dtype` is parametrised independently of `--dtype`, that cell runs in every float32 run of the file, MPS included. A `torch.manual_seed(0)` at the top of the test, as its sibling `test_gradients_are_finite_on_ordinary_patches` already does, closes it. One line, and it keeps this file's MPS run from flickering after merge.

### 5. [suggestion] Make the #4152 xfail assert the expected column, not `count_nonzero > 0`

Both verified by reading rather than a new measurement: the strict xfail flips to a (strict, so failing) XPASS on *any* non-zero output, including a wrong "fix" such as clamping the kernel away from zero, which returns `0.15, 1.15, 2.15, …` at h=0.01 instead of `1, 2, 3, …`. Asserting the actual expected column (`0.9998, 0, 0.9998, 1.9998, …`, the literal the old test carried) makes the pin flip only when the real values come back. Cheap, and it preserves the h=0.01 literal that this diff otherwise deletes.

### Verified and not raised

- The other pass's "TestSolveWithMask::test_smoke keeps the same defect class" is true (seed 430 fails at 1e-4 with cond 7.4e4; the checked-in seed 0 passes at cond 51) but pre-existing, deterministic, and outside this diff. Worth a follow-up issue, not this PR.
- Its "shared `_UNSUPPORTED_KERNEL_MSGS` tuple lets a typo'd probe read as unsupported on MPS" is a fair hardening point (the #4117 class), but the three probes as written are correct, so I leave it to the author's judgement.

ducha-aiki:
## Re-review of `4eef55d2` — approve, nothing open

The delta is exactly the five items from my two comments above, and every one checks out by execution. All MPS runs below are with `PYTORCH_ENABLE_MPS_FALLBACK` unset.

| item | fix in `4eef55d2` | verified |
| --- | --- | --- |
| #4152 xfail / parenthesis test guard the wrong op | both now probe `supports_replicate_padding` first | torch 2.5.1 CPU float16: both cells **SKIPPED** with the replicate-padding reason; bfloat16 still PASSED / XFAIL. On 2.9.1 every cpu/mps half cell is PASSED / XFAIL as before. |
| 3D probes re-raise `NotImplementedError` on torch ≤ 2.8 MPS | `except (_UnsupportedProbeDtype, NotImplementedError): return False` | 2.5.1 and 2.8.0 MPS: `(bil3d, near3d, bil2d, bic2d) = (False, False, True, False)`, no raise; `TestMotionBlur3D` goes from **56 failed / 60 passed** to **60 passed / 58 skipped** on both builds. 2.9.1 unchanged. |
| new `bilinear` XFeat cell unguarded on old CPU half | `supports_bilinear_2d_grid_sample` probe on `test_smoke` and the bilinear branch | 2.5.1 CPU float16/bfloat16: 6 passed, 8 skipped, 0 failed in `TestInterpolateSparse2d` (was 4 failed). |
| unseeded bf16 SIFT gradient cell | `torch.manual_seed(0)` before the draw | 20 runs of `[cpu-desc_dtype1]` and 20 of `[mps-desc_dtype1]`: **0 failures** (was 3/20); all seven live `desc_dtype` cells pass on cpu and mps. |
| #4152 xfail asserted only non-zero | asserts the full expected column via `assert_close` | a simulated wrong fix (kernel clamped to ≥ 1e-7) returns `0.15, 0, 0.15, 1.148, …` and **fails** the new assertion, so the strict pin would not XPASS on it; the real all-zero output still xfails on cpu-fp16/bf16 and mps-fp32/fp16/bf16. |
| TESTING.md rule-4 tension in the motion test | one-line comment at the fallback | reads correctly. |

Full-matrix regression check on 2.9.1, `44c74016` → `4eef55d2`, same 10 files: MPS float32 **31 failed / 1587 passed / 190 skipped / 6 xfailed** on both heads with zero node-level transitions; CPU float16+bfloat16 **216 / 2981 / 124 / 15** on both heads, zero transitions. So the commit changes nothing on the current-torch surfaces and only retires the old-torch false reds and the flake. CI at the time of writing: 18 checks green, 19 still running, none failed.

Follow-ups outside this PR, already noted above: `TestSolveWithMask::test_smoke` keeps a draw-dependent tolerance (seed 430 fails), and the two `TestSolveWithMask` MPS nodes that fail without the CPU fallback.

Issue #4153, opened 2026-09-01
Track remaining Linux CPU float16 and bfloat16 test failures


ducha-aiki:
## Goal

Track the known Linux CPU half-precision test failures while CI defends the currently passing floor with strict expected-failure manifests.

## Baseline

Measured on 2026-09-01 at `4990046f` with Python 3.11.14 and PyTorch 2.9.1 on Linux x86_64:

- `float16`: **8,136 passed**, **608 failed**, 3,599 skipped, 40 xfailed, 5 xpassed.
- `bfloat16`: **8,198 passed**, **592 failed**, 3,550 skipped, 43 xfailed, 5 xpassed.

Reproduction:

```bash
KORNIA_TEST_OPTIMIZER= pixi run test-module tests/ --device=cpu --dtype=float16
KORNIA_TEST_OPTIMIZER= pixi run test-module tests/ --device=cpu --dtype=bfloat16
```

The linked CI PR will record exact failing node IDs as strict xfails. A fix therefore becomes a strict XPASS and requires removing its manifest entry and checking off the corresponding item below.

### augmentation

- [ ] `tests/augmentation/_3d/test_random_affine_3d.py` — float16: 1, bfloat16: 1
- [ ] `tests/augmentation/_3d/test_random_equalize_3d.py` — float16: 2
- [ ] `tests/augmentation/_3d/test_random_rotation_3d.py` — float16: 2, bfloat16: 2
- [ ] `tests/augmentation/container/test_augmentation_sequential.py` — float16: 11, bfloat16: 2
- [ ] `tests/augmentation/container/test_patch_sequential.py` — float16: 1
- [ ] `tests/augmentation/container/test_sequential_aug.py` — float16: 4
- [ ] `tests/augmentation/container/test_video_sequential.py` — float16: 28
- [ ] `tests/augmentation/test_augmentation.py` — float16: 7, bfloat16: 11
- [ ] `tests/augmentation/test_augmentation_mix.py` — float16: 6, bfloat16: 27
- [ ] `tests/augmentation/test_auto_operation.py` — float16: 9, bfloat16: 9
- [ ] `tests/augmentation/test_backward_3d.py` — float16: 24, bfloat16: 28
- [ ] `tests/augmentation/test_base.py` — float16: 3
- [ ] `tests/augmentation/test_perspective_rand.py` — float16: 1, bfloat16: 1
- [ ] `tests/augmentation/test_random_generator.py` — float16: 116, bfloat16: 9
[...]

PR #4154, opened 2026-09-01
Add Linux CPU half-precision CI coverage

### LINKED ISSUE #4070: pixi run test-half is documented but unrun in CI, and currently fails 1382 tests

ducha-aiki:
## Summary

- add blocking Linux CPU `float16` and `bfloat16` full-suite jobs to pull-request and scheduled CI
- make the strict known-failure baseline deterministic by seeding Python, NumPy, and PyTorch from each node ID
- add complete manifest recording, focused strict runs, exact exception validation, and actionable diagnostics
- align the half-precision baseline and expensive model cases with the recently merged CI/test reductions
- fix `AddWeighted` Python-scalar precision and its shape/cardinality test coverage
- document the strict baseline contract and update the stale CI/precision documentation

Closes #4070. Remaining CPU half-precision incompatibilities are tracked in #4153. The regenerated deterministic
manifests contain 603 `float16` and 588 `bfloat16` node IDs.

The previously reported randomized conversion failure came from
`tests/integration/test_conversions.py::TestAngleOfRotations::test_axis_angle_to_rotation_matrix`. That test now
uses NumPy's seedable global RNG, so manifest recording and strict replay use the same per-node input.

## Verification

Exact local environment: Linux x86-64, Python 3.11.14, PyTorch 2.9.1+cpu.

- strict float16 suite: 8,159 passed, 3,607 skipped, 644 xfailed, 5 xpassed
- strict bfloat16 suite: 8,219 passed, 3,559 skipped, 632 xfailed, 5 xpassed
- half-precision harness and conftest tests: 41 passed
- `pixi run pre-commit-all`
- `pixi run typecheck`
- independent final diff review: no findings

Posted on behalf of @ducha-aiki by Claude (Sonnet 5).


testing/half_precision_ci.py
```diff
@@ -0,0 +1,71 @@
+# LICENSE file in the root directory of this source tree.
+from __future__ import annotations
```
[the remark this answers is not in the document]
ducha-aiki: Fixed in 93190db4; testing/half_precision_ci.py now has the repository-managed Apache 2.0 header, and pre-commit.ci passes on the current head.

testing/half_precision_ci.py
```diff
@@ -0,0 +1,83 @@
...
+    collected = {item.nodeid for item in items}
+    missing = failures.keys() - collected
+    if missing:
+        preview = "\n".join(sorted(missing)[:10])
+        raise ValueError(f"{len(missing)} known half-precision failures were not collected:\n{preview}")
+
+    reason = f"Known Linux CPU half-precision failure tracked in {ISSUE_URL}"
+    for item in items:
+        if exception_type := failures.get(item.nodeid):
+            item.add_marker(pytest.mark.xfail(reason=reason, raises=exception_type, strict=True))
```
ducha-aiki: [P1] Validate manifest outcomes, not only collection

This only verifies that every manifest node ID was collected, then appends a strict xfail marker. Pytest evaluates skips before xfails and uses the first applicable xfail marker, so a manifest test that later gains `skip`, `slow`, `pytest.skip()`, or an earlier non-strict xfail can remain in the manifest while CI passes as SKIP/XFAIL. That breaks the stated guarantee that stale entries and changed failures fail CI, and permits disabling one of these tests without removing its baseline entry. Please track the selected node IDs through test reports/session finish and require each one to finish as the manifest-specific XFAIL (including its expected exception), or otherwise reject skip/prior-xfail outcomes for manifest entries.

ducha-aiki (commented):
I found one issue in the strict half-precision baseline enforcement.

ducha-aiki:
Review complete. I found one blocking issue in the strict half-precision baseline enforcement, documented here: https://github.com/kornia/kornia/pull/4154#discussion_r3907902987

The current manifests are internally consistent: all 605 float16 and 591 bfloat16 entries are collected, with no current overlap with static skip, slow, or prior-xfail markers. However, the implementation must also validate each entry’s runtime outcome so future skips or earlier non-strict xfails cannot silently leave stale baseline entries.

Verification on the current head:
- PR CI is green, including both full half-precision jobs.
- tests/test_half_precision_ci.py: 8 passed.
- TestAddWeighted on CPU float16/bfloat16: 48 passed.
- git diff --check passed.

No additional findings.

testing/half_precision_ci.py
ducha-aiki: Fixed in 5752df08. The manifest plugin now inspects the final runtime report plus the original call exception and accepts an entry only when the call phase becomes the tracker-specific XFAIL for the manifest-declared exception. Static or runtime skips, earlier xfails, wrong exceptions, setup or teardown failures, and dynamic pytest.xfail reason spoofing leave the entry pending and fail the session. Regression tests cover each bypass. Verified with 18 focused tracker tests, pre-commit, modified-file type checking, and complete CPU float16 and bfloat16 suite runs.

ducha-aiki:
Addressed the blocking outcome-validation review in 5752df08.

The known-failure tracker now requires every manifest entry to finish as the manifest-specific call-phase XFAIL with its declared exception. Skips, competing xfails, wrong exception types, setup or teardown failures, and dynamic xfail spoofing fail validation.

Verification:
- 18 focused tracker tests passed
- 48 representative CPU float16/bfloat16 tests passed
- pre-commit-all passed
- modified-file type checking passed
- full CPU float16 run: 8,144 passed, 3,612 skipped, 645 xfailed, 5 xpassed
- full CPU bfloat16 run: 8,204 passed, 3,563 skipped, 634 xfailed, 5 xpassed

A preliminary bfloat16 run hit an unrelated randomized conversion tolerance flake; the same deterministic bfloat16 quantization behavior was reproduced on the base and PR trees, and the final complete rerun passed.

ducha-aiki:
## Review of `5752df08`

Reviewed in a worktree at the head commit. The manifest tracker is the right design and `5752df08` closes the skip/prior-xfail bypass from the earlier thread: I re-ran the 18 tracker tests (pass), exercised the fixed-entry, missing-entry and `-k` paths by hand, pulled per-test timings from four CI job logs, and repeated a sample of manifest entries locally. The tracker code is sound. The findings are about what the manifest is a snapshot *of*, and about the cost and upkeep of the contract. Ordered by priority.

### P1 — the manifest pins a random draw, not a failure set

There is no seeding anywhere: no `manual_seed`/`random.seed` in `conftest.py`, `testing/`, or any `tests/**/conftest.py`, and torch's default CPU generator is entropy-seeded per process. Many manifest entries are tolerance assertions on unseeded random inputs (`Quaternion.random`, `_make_rand_se3d`, `torch.randn` shapes, `random.random()` weights). Their pass/fail status in half precision depends on the draw, so a strict xfail on them is a coin flip in both directions: a lucky draw is `[XPASS(strict)]`, an unlucky draw on an unlisted sibling parametrization is a plain `FAILED`. Either way the new required `collector` goes red on an unrelated PR.

Reproduced on macOS float16 (torch 2.9.1), which has the same mechanism even though its failure set differs from Linux. Twelve repeats of three manifest test functions gave twelve different sets of failing parametrizations:

| manifest entry | failed in |
|---|---|
| `test_se3.py::TestSe3::test_mul[cpu-float16-None]` | 6 / 12 runs |
| `test_so3.py::TestSo3::test_ortho[cpu-float16-5]` | 6 / 12 runs |
| `test_quaternion.py::TestQuaternion::test_inverse[cpu-float16-1]` | 1 / 12 runs |

while unlisted siblings (`test_mul[-2]`, `test_mul[-5]`, `test_ortho[-2]`, `test_inverse[-2]`, `test_inverse[-5]`) failed in other runs. The two Linux runs per dtype so far show zero flips, so the per-draw rate on the CI hosts is unmeasured rather than zero, and your own comment on `5752df08` already reports one randomized bfloat16 flake in a preliminary run. Please name that test in the PR.

The fix at depth is to make the suite deterministic before recording the baseline: an autouse fixture that seeds `torch`, `random` and `numpy` from a hash of the node ID (order-independent), then re-record both manifests. Without it the manifest cannot be "exact" in the sense the PR claims, and the strict mode will be the job everyone re-runs.

### P1 — the half jobs take 4 or 26 minutes depending on the runner CPU, and one test file is the reason

Per-test durations from the job-log timestamps (all on `ubuntu-24.04`, `-v`, no xdist; CPU model from the log's system info):

| run | job | CPU | pytest wall | `tests/models/test_small_sr.py` |
|---|---|---|---|---|
| `89523de8` | ubuntu float32 3.11/2.9.1 | — | 164 s | 4 s |
| `89523de8` | half float16 | AMD EPYC 9V74 | 882 s | 534 s |
| `89523de8` | half bfloat16 | AMD EPYC 7763 | 1580 s | 1110 s |
| `5752df08` | half float16 | AMD EPYC 9V74 | 1384 s | 934 s |
| `5752df08` | half bfloat16 | Intel Xeon 8573C | 200 s | 9 s |

Same code, same image, same test count, same `model-weights-ec4a4248…` cache hit on every run (so not a download), and `test_small_sr.py` builds every model with `pretrained=False`. Elementwise files (`filters/test_gaussian.py`, `color/test_lab.py`, `augmentation/test_augmentation.py`) take identical time on both host classes; only conv-heavy files blow up (`test_scale_space_detector.py`: 148 s vs 5.5 s). Three of four half runs so far landed on AMD hosts, where `TestSmallSRNet`/`TestSmallSRNetWrapper::test_cardinality`, a shape-only test with 90 parametrizations per class, spends 35–75 s per 128×128/224×224 case in half precision (0.2 s max in float32). That is the signature of a half conv path with no native kernel on those CPUs; the Windows ISA-clamp comment in `tests.yml` already documents the fleet being heterogeneous this way.

So the blocking PR job is a 15–26-minute critical path on most draws, against ~11 min for the previous longest job (dynamo). A shape assertion at 224×224 in bfloat16 cannot tell you anything the 64×64 case does not. Please trim this in the PR that adds the job, e.g. restrict `test_cardinality` to 64×64 under half dtypes or mark the 128/224 sizes `slow` (they still run on the scheduled `runslow` jobs). That caps the AMD case at roughly 6–8 min. If you would rather take the variance consciously, say so in the PR body so the number is on record.

### P1 — the docs still say the half suite is not run in CI, and #4070 stays open

`git grep -nE '#4070|issues/4070|not run in CI|no CI job sets'` on the PR head hits:

- `README.md:115` — "suite is not run in CI (see #4070), so these numbers are refreshed by hand"
- `docs/source/get-started/precision.rst:122-123` — the same sentence
- `tests/geometry/test_conversions.py:3743` — "no CI job sets KORNIA_TEST_DTYPE to float16/bfloat16, so … the x86-64 rows need one measurement on such a runner to become live"

All three are false the day this merges, and #4070 ("pixi run test-half is documented but unrun in CI") is OPEN and not referenced by the PR. Please update the three, and either close #4070 from this PR or say why it stays open. The `test_conversions.py` comment is exactly the "stale pin" class `AGENTS.md` warns about; the grep there is scoped to `kornia/ tests/`, which is why README/docs hits slip past it.

In the same pass, `AGENTS.md`/`TESTING.md` need to document the new contract, since #4150 exists precisely to record test-policy mechanics there:

- what `--xfail-known-half-precision` enforces and that it mirrors CI exactly: `--device=cpu`, one half dtype, `KORNIA_TEST_OPTIMIZER=` (dynamo/compile tests are deselected before the manifest check, so an `inductor` run trips it), no `-k`, no `--runslow`;
- the edit cases: a fix ⇒ delete the line; a rename/reparametrization ⇒ update the node ID; a new half failure ⇒ fix it or add the line with its exception type;
- how to regenerate a whole manifest. The PR provides no way. The baseline is pinned to Linux, Python 3.11, torch 2.9.1 (hard-coded in both workflows), and the next torch bump invalidates ~1,200 lines; recording by hand is not viable. `tests/api_surface.json` + `tests.test_api_surface.regenerate()` is the in-repo precedent. The tracker already has `call.excinfo.type` in hand in `pytest_runtest_makereport`, so a record mode is a few lines.

### P2 — no partial-run mode, and the errors do not say what to do

Verified on the worktree:

- Fixed entry: `[XPASS(strict)] Known Linux CPU half-precision failure tracked in …/4153 [tracker=…]` plus `ERROR: 1 known half-precision failure did not finish with the manifest-specific xfail:`. Neither names the manifest file or the action.
- Renamed entry: `UsageError: 1 known half-precision failures were not collected:` + node IDs, exit 4, "no tests ran". Same gap.
- `-k` selection: the collection check passes (the root conftest hook runs before `-k` deselection), the selected tests run, then the session fails listing every deselected entry as "did not finish with the manifest-specific xfail", which reads as a test defect rather than a usage error, and the summary bar stays green while the exit code is 1.

A contributor fixing one red entry therefore has no local feedback loop shorter than the full 15–26-minute suite. Scoping the missing/pending check to entries whose files were selected (or hooking after deselection via `pytest_deselected`) keeps full-run strictness while letting `pytest tests/geometry/test_so3.py --xfail-known-half-precision` work. And please put the manifest path and the verb into each message ("remove this line from `testing/half_precision_xfails/cpu_float16.txt`"), and distinguish "never ran" from "ran with the wrong outcome" in the summary.

### P2 — the `AddWeighted` oracle now mirrors the implementation bit-for-bit

`add_weighted` does `torch.tensor(alpha, dtype=src1.dtype)` before multiplying, so in float16 `0.1` becomes `0.09997` and rounds twice; `src1 * 0.1` keeps the Python scalar in float32 opmath and rounds once. Measured on torch 2.9.1 with `alpha, beta, gamma ~ U[0,1)`, 1000 elements × 20 draws, against a float64 reference:

| dtype | mean abs err, `add_weighted` | mean abs err, plain expression | elements that differ |
|---|---|---|---|
| float16 | 2.73e-4 | 2.54e-4 | ~17 % |
| bfloat16 | 2.22e-3 | 2.03e-3 | ~21 % |

So the pre-rounding is a small, real accuracy wart, and the new `expected` reproduces it exactly, which is why the mismatch count against `add_weighted` is now zero: the smoke test can no longer detect this or any future widening of the rounding. Kornia's policy is to fix warts rather than encode them. The fix is in `kornia/enhance/core.py`: keep Python scalars as scalars (the `torch.tensor` wrap only exists for symmetry with the tensor branch, and also costs an allocation per call), and keep the old independent oracle. If you would rather not touch library code in a CI PR, compare against a float32 reference with the half tolerance instead of a copy of the implementation.

While there: `get_input` calls `random_shape(size, max_elem)` but the signature is `random_shape(dim, min_elem=1, max_elem=10)`, so `10` binds to `min_elem` and every shape is exactly `(10, 10, 10)`; `test_smoke` also hardcodes `size=3`, so its four `size` parametrizations are one case. Pre-existing, but you are editing the line below it.

### Notes, non-blocking

- `NotImplementedError` is a subclass of `RuntimeError`, so `isinstance` accepts a `NotImplementedError` for an entry recorded as `RuntimeError` (183 bfloat16 / 87 float16 entries). The "changed failures fail CI" claim is one-directional for that pair; `type(exc) is expected` makes it exact.
- `_EXCEPTION_TYPES` is a closed allow-list of seven names, and the 100 float16 + 72 bfloat16 `NotImplementedError` entries are torch dispatch gaps (`"lu_cpu" not implemented for 'Half'` etc.), so every torch bump re-litigates them op by op and the first `IndexError`/`OverflowError` baseline needs a code change. Treating the type as informational, or falling back to `builtins`, decouples CI from torch's error taxonomy.
- 1,029 of the 1,196 node IDs end in an auto-indexed or bare-numeric parametrize component (`expected3`, `input_shape0`, `batch_prob1`, `-5`). Inserting a case at the front of such a list renames every recorded ID for that test and aborts both half jobs at collection with the "not collected" error; a PR that only adds a test case kills the required legs. Keying entries by function or documenting the remap procedure is what survives ordinary test maintenance.
- `test_recorded_baseline_manifest` pins 605/591 and runs on every float32/float64 leg, so deleting one manifest line after a fix turns ~40 unrelated legs red on `assert 604 == 605`. The half legs already enforce both directions; I would drop the count pin. The `assert ISSUE_URL == "…/4153"` line asserts a constant against itself.
- The `secrets.token_hex` nonce in the reason string is dead defence: every bypass it targets is already caught by the `call.excinfo` `isinstance` check or the `wasxfail is not None` branch, and a prior marker with identical reason and `raises` is behaviourally the same marker. It costs 1,196 reason strings that never diff clean between runs.
- Tests parametrized on dtype only (no device), e.g. `test_normal_transform_export_crosses_singleton_boundary[bfloat16-…]`, have no `cpu-` in their node ID and cannot be listed at all under the loader's `cpu-{dtype}` check.
- The 5 XPASS in both half jobs are pre-existing non-strict xfails in `tests/augmentation/test_augmentation.py` (`might fail under windows OS due to printing preicision` ×4, `TestCenterCropAlternative::test_random_p_1`). All five also XPASS in the ubuntu float32 job of the same run, alongside six more in `tests/geometry/`, so they are stale on `main` rather than half-specific. Worth a follow-up issue.
- #4153 still lists `tests/enhance/test_core.py — float16: 2, bfloat16: 1`, but no `test_core.py` line remains in either manifest, so that item can be ticked on merge.

### Verified

- Worktree at `5752df08`, imports resolve to the worktree; `tests/test_half_precision_ci.py`: 18 passed.
- Manual runs with a scratch manifest: a now-passing entry fails as strict XPASS plus the tracker error; a missing entry raises the collection error; `-k` deselection fails the session via the tracker.
- Manifest counts (605/591) match the PR body; zero overlap between the 1,196 manifest IDs and the 419 `slow`-marked half-dtype IDs (`--collect-only -m slow`); CI runs `uv run pytest -v tests/` without xdist, so the per-process tracker is sound for CI (it would not survive `-n`, which is fine to leave until someone adds xdist).
- The `AddWeighted` oracle is bitwise stable under the test's unseeded `random.random()` inputs, because it computes the same expression as the implementation in the same order.
- CI on `5752df08`: both half-precision jobs green (float16 24m22s, bfloat16 4m42s), dynamo green. The collector is red only because `tests (macos-latest, float32, 3.12, 2.5.1)` hit the 120 s `pytest-timeout` in `tests/models/test_sam_onnx.py::test_sam_encoder_to_onnx`, which this PR does not touch; a re-run of that job should clear it.

ducha-aiki:
Addressed Fable's review in `fb85eb68`.

Highlights:

- strict/record modes now seed Python, NumPy, and PyTorch deterministically from each node ID;
- both manifests were recorded again and independently replayed (`float16`: 605 entries, `bfloat16`: 592);
- added a full-manifest recorder, focused file/node/`-k` strict runs, exact exception matching, broader exception
  loading, and manifest-specific remediation messages;
- large half-precision SmallSR cardinality cases now run only with `--runslow`;
- fixed `AddWeighted` scalar pre-rounding and the two nearby test cardinality bugs;
- updated README, precision docs, `TESTING.md`, `AGENTS.md`, and the x86-64 conversion pin; the PR now closes #4070;
- named the earlier preliminary flake as
  `TestAngleOfRotations::test_axis_angle_to_rotation_matrix[cpu-bfloat16-x]`.

Verification on the pushed commit:

- full strict float16: 8,121 passed, 645 xfailed;
- full strict bfloat16: 8,180 passed, 635 xfailed;
- 27 manifest-harness tests and 100 AddWeighted CPU dtype cases passed;
- focused partial strict and SmallSR `--runslow` paths passed;
- `pixi run pre-commit-all`, `pixi run typecheck`, and `git diff --check` passed.

The function-level manifest key redesign and unrelated stale xfails remain follow-up work, as discussed.

ducha-aiki:
Merged current `main` into this branch in `c02e0c29`, resolving the conflicts in `TESTING.md` and
`tests/models/test_small_sr.py` while preserving both the half-precision documentation and the reduced upstream
SmallSR coverage from #4169.

The regenerated manifests also incorporate the recently merged changes from #4171, #4169, #4166, and #4163:
603 float16 entries and 588 bfloat16 entries. During strict replay I found that
`TestAngleOfRotations::test_axis_angle_to_rotation_matrix` used `np.random.default_rng()` without a seed, bypassing
the per-node manifest seed. It now uses the seeded global NumPy RNG; repeated focused replay is stable.

Verification after the merge:

- float16 strict full suite: 8,159 passed, 3,607 skipped, 644 xfailed, 5 xpassed
- bfloat16 strict full suite: 8,219 passed, 3,559 skipped, 632 xfailed, 5 xpassed
- focused half-precision/conftest tests: 41 passed
- `pixi run pre-commit-all`: passed
- `pixi run typecheck`: passed (one existing deprecation diagnostic)
- hosted GitHub Actions float16 and bfloat16 half-precision jobs: passed
- independent final diff review: no findings

ducha-aiki:
## Review

Reviewed the full diff and validated the harness empirically on Linux x86-64 / Python 3.11 / PyTorch 2.9.1+cpu,
from a worktree of the PR head (`kornia.__file__` checked to name the worktree on every run).

The strict-manifest design is the right shape for this problem. It converts ~1200 latent half-precision
failures into a pinned, reviewable baseline instead of a blanket skip, and it fails loudly in both
directions — a fixed test and a stale node ID each break the job. The tracker internals are careful and
correct: `tryfirst` hookwrapper ordering so the post-`yield` runs after `_pytest.skipping` has set
`wasxfail`, three-phase gating in `pytest_runtest_logfinish`, markers applied before `-k` with
`pytest_deselected` cleanup after. At ~2m20s of pytest per dtype these two new blocking jobs are also
genuinely cheap next to the existing matrix — worth stating, since "two more full suites per PR" sounds
more expensive than it is.

### What I reproduced

| Check | Result |
| --- | --- |
| `tests/test_half_precision_ci.py` | 27 passed |
| Full strict `float16` suite | 8159 passed, 3607 skipped, 644 xfailed, 5 xpassed — exit 0, 2m21s |
| Full strict `bfloat16` suite | 8219 passed, 3559 skipped, 632 xfailed, 5 xpassed — exit 0, 2m13s |
| Focused strict run (`tests/enhance tests/color tests/geometry/liegroup`) | 1121 passed, 38 xfailed, exit 0 |
| Focused strict run with `-k` | deselected manifest entries correctly exempted, exit 0 |
| Four new `x86_64` rows in `test_wart_agreement_gap_at_2_28_is_a_kernel_measurement` | all four dtypes pass here |
| `pixi run lint` | passed |
| `pixi run typecheck` | clean (only the pre-existing `lightglue` `custom_fwd` deprecation) |

Both full suites match the numbers in the description exactly. I ran `float16` with `-p no:randomly` and
`bfloat16` without it and got identical results — a reasonable independent check that the per-node-ID
seeding is genuinely order-independent rather than order-lucky.

The `add_weighted` test has teeth: re-adding the `torch.tensor(alpha, dtype=src1.dtype)` cast makes
`test_python_scalars_keep_opmath_precision` fail for both `float16` and `bfloat16`. And the
`random_shape(size, max_elem)` fix is a real bug — the positional argument was landing in `min_elem`, so
every generated shape had all dimensions pinned to exactly 10, and `test_smoke` was ignoring its own
`size` parametrization.

### Blocking

**1. `add_weighted` and `AddWeighted` are no longer TorchScript-scriptable** (`kornia/enhance/core.py:76`).

Dropping the `torch.tensor(...)` else-branches leaves `alpha`/`beta`/`gamma` at their declared
`Union[float, Tensor]` type at the `return src1 * alpha + src2 * beta + gamma` line, which TorchScript
cannot resolve. Verified on both revisions:

```
main:   script(add_weighted): OK     script(AddWeighted): OK
branch: script(add_weighted): FAIL   script(AddWeighted): FAIL
```
```
Arguments for call are not valid. The following variants are available:
  aten::mul.Tensor(Tensor self, Tensor other) -> Tensor:
  Expected a value of type 'Tensor' for argument 'other' but instead found type 'Union[Tensor, float]'.
  aten::mul.Scalar(Tensor self, Scalar other) -> Tensor:
  Expected a value of type 'number' for argument 'other' but instead found type 'Union[Tensor, float]'.
```

`tests/enhance/test_core.py::TestAddWeighted::test_dynamo[cpu-float32-jit]` passes on `main` and fails
here. Nothing in this PR's matrix catches it, because the `dynamo` job only runs inductor — and AGENTS.md
lists JIT as a distinct compatibility surface from inductor for exactly this reason. The opmath win is
worth keeping; doing the multiply inside each `isinstance` branch, where TorchScript refines the type,
gets both.

### Worth fixing before merge

**2. Silent output-dtype change for integer inputs, and no `CHANGELOG.md` entry at all.**

`add_weighted(uint8_img, 0.5, uint8_img, 0.5, 0.0)` previously returned `uint8` zeros — the scalar was
rounded to `0` in the input dtype by `torch.tensor(0.5, dtype=torch.int64)` — and now returns `float32`
with the right values. The new behaviour is correct, and the old one was a silent wrong answer, which is
precisely why it belongs in `CHANGELOG.md` under `Bug fixes`. The half-precision numerical change is
user-visible too. AGENTS.md asks for an entry per user-visible change; this PR has none.

**3. An absolute-path selector silently disables the entire baseline** (`testing/half_precision_ci.py:201`).

`_matches_selector` compares against relative node IDs, so an absolute target filters every manifest entry
out. `mark_known_failures` then applies zero xfail markers, `pending` is empty, and the tracker's
"manifest needs updates" diagnostic never prints. Same target, two ways:

```
pytest tests/geometry/liegroup --xfail-known-half-precision      -> 342 passed, 35 xfailed, exit 0
pytest "$PWD/tests/geometry/liegroup" --xfail-known-half-precision -> 342 passed, 35 failed,  exit 1
```

The failing output looks exactly like 35 real regressions, with nothing pointing at the real cause. Since
`$GITHUB_WORKSPACE/tests` is a natural thing to type in a workflow, normalizing the selector against
`config.rootpath` — or rejecting an absolute selector with its own message — is worth doing.

**4. The record -> strict round trip is not closed, in three separate ways.**

The recorder can emit manifest lines that `load_known_failures` then refuses, so the documented
regeneration workflow can produce a file that aborts the next strict run at collection with a
`UsageError` and zero tests run. The only escape is deleting the line, which puts you back at a
permanently red blocking job.

- *Exception names.* `_exception_name` writes a fully qualified name for anything outside `builtins`, but
  the loader accepts only the eight hardcoded `_EXCEPTION_TYPES` entries plus `builtins` attributes. Every
  `BaseError` subclass is outside that set — `KORNIA_CHECK_SHAPE` raises `ShapeError`,
  `KORNIA_CHECK_IS_TENSOR` raises `TypeCheckError`, `KORNIA_CHECK_SAME_DEVICE` raises `DeviceError`:

  ```python
  >>> _exception_name(ShapeError)
  'kornia.core.exceptions.ShapeError'
  >>> _exception_name(ShapeError) in _EXCEPTION_TYPES
  False
  ```

  Latent today, but it is the first thing a new failure in any `KORNIA_CHECK_*` path will trigger.
  Resolving dotted names through `importlib` behind a module allowlist would close it.

- *Node IDs without a dtype.* The dtype guard requires `float16`/`bfloat16` in the node ID, but a failing
  test not parametrized by the `dtype` fixture has none — including this PR's own
  `test_python_scalars_keep_opmath_precision[dtype0]`.

- *Setup-phase failures.* `FailureRecorder` records only `report.when == "call"`, and the tracker requires
  `when == "call"` to match. A test that errors in an autouse or class-scoped fixture (say a `"…" not
  implemented for 'Half'`) can therefore be neither recorded nor hand-baselined — regenerating both
  manifests would still leave the job red.

**5. The record-mode guard blocks `-k` but not the other partial-selection flags** (`conftest.py:203`).

The guard exists specifically to prevent truncated recordings, but it checks only the target set and
`config.option.keyword`. Verified — writing to a scratch path, not the repo:

```
pytest tests/ -x --record-half-precision-failures=<scratch> -> 1 failed, 8 passed
wrote 1 manifest entry (the checked-in float16 manifest has 605)
```

Pointed at the real path that silently clobbers a 605-line manifest down to one line. `config.option.markexpr`,
`deselect`, `maxfail`/`exitfirst` and `lf`/`ff` deserve the same treatment as `-k`.

### Minor

**6. Provenance comment now contradicts the table it documents** (`tests/geometry/test_conversions.py:3750`).
The generation-snippet block still says the figures were "executed on macOS arm64 against both torch 2.9.1
and torch 2.5.1", but four `("2.9.1", "cpu", "x86_64", …)` rows were added below it. This is the one test
in the file whose entire justification is that a literal must name the machine it was measured on, so the
snippet needs a line saying where the x86-64 figures came from. They do reproduce here.

**7. The ten `.rst` underline changes are a bad-merge artifact, not an intended edit.**

`color.lab`, `color.luv`, `color.xyz`, `community/index`, `filters.kernels`, `image.drawing`,
`models/dexined`, `models/hardnet`, `models/rt_detr`, `models/tiny_vit` — each has its title underline
padded from 7 to 8 `=`, and nothing else in those files changed. There is no RST hook in
`.pre-commit-config.yaml`, so no repo tooling asks for this, and a 7-character underline under a
7-character title is already valid RST.

The mechanism is worth spelling out: every one of these titles is exactly 7 characters, so its underline on
`main` is the literal string `=======` — a git conflict-marker separator. I enumerated every `.rst` in
`docs/source` on `main` whose underline is exactly `=======`, and the set matches the ten rewritten files
exactly. All ten changes come from `c02e0c29 "Merge main into ci/cpu-half-precision"`. A
conflict-resolution tool mistook RST underlines for conflict markers.

Harmless to Sphinx, but it should come out of a CI PR — it is pure conflict surface against in-flight docs
work right after the #4155/#4173 revamp. I did re-check that merge for other damage: no leftover
`<<<<<<<`/`>>>>>>>`/`|||||||` markers anywhere in the tree, and no other `docs/` difference against `main`
beyond the intended `precision.rst` edit. So the damage looks confined to these ten lines.

**8. `test_python_scalars_keep_opmath_precision` under-asserts relative to its name.** `expected` is
computed with the same expression the implementation uses, so it pins "the function does not pre-cast"
rather than "opmath precision is retained". It does catch the regression — I verified that — but asserting
the downcast form actually *differs* would make the name honest and the test self-evidently non-vacuous.

**9. `tests/integration/test_conversions.py` trades a modern RNG for the legacy global one.** Swapping
`np.random.default_rng()` for `np.random.random` plus `# noqa: NPY002` reverses a deliberate modernization
to buy manifest determinism. It works and the comment explains why, but `seed_test_rng` already computes
and returns exactly the seed this test needs — it just has no way to hand it over. A small fixture
exposing that seed to `default_rng(seed)` would keep the modern RNG and the same determinism.

**10. Load-bearing ordering worth a comment.** The tracker depends on `_configure_half_precision_manifest`
running *before* the non-strict `tf32` xfail block in `pytest_collection_modifyitems`, since pytest takes
the first matching xfail marker. There are zero `@pytest.mark.tf32` tests today so nothing is affected, and
reordering would fail loudly rather than silently — but a one-line comment would save someone an afternoon.

### Summary

Finding 1 blocks. 2-5 are worth resolving before this becomes a blocking gate, since 3, 4 and 5 all turn a
contributor mistake into either a red job with a misleading cause or a clobbered checked-in manifest. 6-10
are polish. The core mechanism is well built and the verification numbers reproduce exactly.

ducha-aiki:
Implemented the Opus review feedback and merged current `main` into this branch in `1a6103b5`.

Changes:
- Restored TorchScript-compatible `add_weighted` scalar handling while preserving operator-math precision and correct integer promotion, with regression coverage and a changelog entry.
- Made record/strict manifests round-trip setup failures, dtype-less node IDs, and allowlisted dotted Kornia/PyTorch exception types.
- Normalized absolute strict selectors and rejected all reviewed partial-recording options (`-k`, `-m`, `--deselect`, `--maxfail`/`-x`, `--lf`, and `--ff`).
- Exposed the deterministic per-node seed to tests and migrated the reviewed NumPy sample to `default_rng(seed)`.
- Documented marker ordering, corrected the x86-64 provenance note, strengthened the opmath regression, and reverted the ten unrelated RST underline edits.
- Resolved the workflow and `conftest.py` merge conflicts, preserving both the MPS and half-precision collectors/trackers.
- Refreshed both CPU-half manifests after the merge: removed stale entries and recorded the two new exact `bfloat16` Lie-group failures.

Verification:
- Manifest and known-failure tests: 52 passed.
- `add_weighted` across CPU `float32`, `float16`, and `bfloat16`: 77 passed.
- TorchScript/JIT regression: 1 passed.
- Full strict `float16` suite: 8,237 passed, 627 expected XFAILs, exit 0.
- Full strict `bfloat16` suite: 8,294 passed, 618 expected XFAILs, exit 0.
- All non-conflict pre-commit hooks passed. The conflict hook passes the complete PR diff and the index has no unresolved entries; its all-files mode separately flags ten valid seven-character RST title underlines already present on `main`. A standalone local `ty` executable was unavailable, so the newly triggered CI typecheck remains authoritative.
- GitHub reports the PR mergeable; required checks are currently pending.

Written by Codex (GPT-5) on behalf of @ducha-aiki

ducha-aiki:
## Code review — #4154 (`ci/cpu-half-precision`)

Reviewed all 16 files and verified the findings locally on x86-64 Linux, torch 2.9.1, pytest 9.0.3. The strict mode does work end-to-end on subsets (`tests/geometry/liegroup tests/enhance tests/color` strict float16: 1125 passed, 38 xfailed, exit 0), and `tests/test_half_precision_ci.py` (36) and `tests/enhance/test_core.py` (29) pass. The problems below are in the harness's failure modes, not its happy path.

### Blocking: the strict gate can abort in ways regenerating the manifest cannot fix

**1. `testing/half_precision_ci.py:60` — exception names do not round-trip.**
`_exception_name(pytest.fail.Exception)` returns `"Failed"` (pytest sets `Failed.__module__ = "builtins"`), and `_resolve_exception_type("Failed")` returns `None` — the loader only resolves builtins, `kornia.*` and `torch.*`. Same for `numpy.linalg.LinAlgError`, `numpy.exceptions.AxisError`, `PIL.UnidentifiedImageError`. Any test that fails via `pytest.fail(...)` — used widely in kornia tests, and by pytest-timeout — gets recorded as `Failed\t<nodeid>`; the next strict run raises `ValueError: unknown exception type: Failed` → `pytest.UsageError` → exit 4 with **zero tests run**. Note that the sibling `testing/known_failures.py:32` explicitly maps `"pytest.fail.Exception"`, so this case demonstrably occurs in kornia's baselines.

**2. `testing/half_precision_ci.py:109` — the recorder truncates the checked-in baseline.**
`FailureRecorder.pytest_sessionfinish` writes unconditionally. Verified: `KORNIA_TEST_DTYPE=float16 pytest tests/ --record-half-precision-failures=<path> --collect-only` exits 0 and writes a file containing only the two `#` header lines. Pointed at `testing/half_precision_xfails/cpu_float16.txt` — which is the invocation TESTING.md documents — a `--collect-only`, Ctrl-C'd or otherwise aborted run destroys all 587 entries. TESTING.md tells the maintainer "recording intentionally exits nonzero … but it still writes the sorted manifest", so a nonzero exit is not a usable signal that something went wrong. Write only when the session actually ran to completion with reports.

**3. `testing/half_precision_ci.py:211` — the dtype-token heuristic rejects legitimate node IDs.**
Under `KORNIA_TEST_DTYPE=bfloat16`, pytest collects `tests/test_conftest.py::TestIsSubprocessIsolatedTest::test_global_dtype_fixture_preserved[float16]` (parametrized over the *strings* `"float16"`/`"bfloat16"`). If it ever fails in the bfloat16 job the recorder writes it, and the next `load_known_failures("bfloat16")` raises `node ID does not select bfloat16` → UsageError → the whole job aborts. Recorder and loader disagree, so the documented "regenerate, review the diff, rerun strict" recovery loop cannot get out of it.

**4. `testing/half_precision_ci.py:267` — pre-hook selection aborts with the wrong remedy.**
Verified: `--xfail-known-half-precision --ignore=tests/geometry/liegroup/test_se2.py` runs no tests and aborts with "35 known half-precision failures were not collected … remove or update stale manifest lines" — advice that would delete *correct* entries. `-k` survives because `_pytest.mark`'s hook is `trylast`; `--ignore` and `--deselect` are applied earlier. The same 35-line block is also appended as noise to a genuine collection `ImportError`.

**5. `conftest.py:218` — the record-mode partial-selection guard is incomplete.**
It rejects `-k`/`-m`/`--deselect`/`-x`/`--lf`/`--ff` but not `--ignore`, `--ignore-glob`, `--collect-only`/`--co`, or xdist `-n`. `pytest tests/ --record-half-precision-failures=… --ignore=tests/geometry` passes every guard (args normalize to exactly `{"tests"}`) and silently writes a manifest missing every geometry entry; the next strict run then reports hundreds of "stale" lines that were never stale.

**6. `testing/half_precision_ci.py:168` — the `--collect-only` guard was dropped.**
`KORNIA_TEST_DTYPE=float16 pytest tests/geometry/liegroup/test_se2.py --xfail-known-half-precision -q --collect-only` exits 1 with "ERROR: 21 known half-precision failures need updates …" even though nothing ran and nothing is stale. `testing/known_failures.py:81-83,89` guards exactly this with `self._collect_only = session.config.getoption("collectonly")`.

### Determinism: the contract is narrower than the PR claims

**7. `conftest.py:726` — seeding is function-scoped, so collection-time and higher-scoped randomness stays unseeded.**
`tests/enhance/test_core.py:57-59` builds parametrize values with `torch.randn(2, 3, 5, 5)` at import time; `tests/morphology/test_dilation.py:29` and siblings do the same; session-/class-scoped fixtures that draw randomness (`tests/io/test_io_image.py:73`, `tests/filters/test_dissolving.py:42`) run before any function-scoped autouse fixture. A test built on those can flip in or out of the manifest between runs and turn the blocking gate red with no kornia change. Latent for now — none of them are in the current manifests.

**8. `conftest.py:726` — `test_rng_seed` returns `None` outside manifest mode.**
`tests/integration/test_conversions.py:207` calls `np.random.default_rng(test_rng_seed)`. Under `--device=cpu --dtype=float32` (the ~60 non-half legs) and any local `pixi run test-quick`, that is `default_rng(None)` — seeded from OS entropy, exactly as before. The flaky test this PR set out to fix is still nondeterministic everywhere except the two new jobs. Also, the description says it "now uses NumPy's seedable global RNG"; it does not — it uses a per-call `Generator`.

**9. `tests/test_half_precision_ci.py:46` — two tests reseed the process-global RNGs and never restore them.**
`test_seed_test_rng_is_stable_and_node_specific` calls `seed_test_rng(...)` three times (`random.seed` + `np.random.seed` + `torch.manual_seed`); `tests/enhance/test_core.py:76` calls `random.seed(0)`. Neither restores. Outside manifest mode — `test-quick`, every float32/float64 leg — every later test in that worker inherits the shifted stream, which can equally mask an existing flake or create a new one depending on collection order.

### `add_weighted` and its tests

**10. `CHANGELOG.md:191` — the integer-input change is a breaking change, not a bug fix.**
`add_weighted(uint8, 0.5, uint8, 0.5, 0.0)` used to return `uint8` (scalars were cast to `src1.dtype`, so `0.5` became `0`); it now returns `float32` (verified). `kornia/models/segmentation/base.py:213,229` calls it with `images` that may be integer, so the overlay written by `SegmentationModel.save()` changes dtype and value range. AGENTS.md requires this under `Breaking changes`, saying what used to happen as well as what happens now.

**11. `tests/enhance/test_core.py:51` — the new tests bypass the injected fixtures.**
`test_python_scalars_keep_opmath_precision` shadows the `dtype` fixture with its own `parametrize` and builds `torch.linspace(..., dtype=dtype)` with no device; `test_python_scalars_promote_integer_inputs` takes no fixtures at all. The CUDA and MPS legs therefore run both as CPU tests — the promotion/opmath behavior is never checked on an accelerator — and neither respects `--device`. AGENTS.md asks for the injected `device`/`dtype` fixtures and `self.assert_close()` rather than bare `torch.equal`.

**12. `tests/enhance/test_core.py:64` — `assert not torch.equal(expected, downcast)` pins a PyTorch CPU-kernel property everywhere.**
With no `device`/`dtype` fixture this runs in all ~60 CPU legs: torch 2.1.2 → 2.9.1 in scheduled CI, Windows with `ATEN_CPU_CAPABILITY=avx2` forced, macOS arm64. If any of those builds casts the wrapped scalar to the tensor dtype instead of keeping it in opmath, it fails with no kornia regression. This is the exact failure mode `tests/geometry/test_conversions.py`'s own comment ("an unmeasured platform must not inherit a kernel literal") exists to prevent.

**13. `tests/geometry/test_conversions.py:3773` — new literals keyed only by `platform.machine()`.**
`platform.machine()` returns `"x86_64"` on both Intel Xeon (AVX-512) and AMD EPYC (AVX2) `ubuntu-latest` runners, and PyTorch dispatches different vectorized gemm kernels for them; only the Windows legs clamp the ISA. All four literals reproduce on my box with torch 2.9.1 (4 passed at all four dtypes), but a runner that reassociates differently turns a documentation measurement into a red blocking gate. The `float32`/`float64` rows additionally now run on the whole ubuntu 2.9.1 matrix, not just the two new half jobs the comment says they were added for.

**14. `kornia/enhance/core.py:68` — the three `isinstance` branches are now byte-identical.**
`return output + gamma` is written twice. The reader has to diff two identical lines three times to conclude the `isinstance` test only guards the `KORNIA_CHECK`. The body collapses to three bare `if isinstance(x, torch.Tensor): KORNIA_CHECK(...)` guards followed by `return src1 * alpha + src2 * beta + gamma` — same opmath semantics, half the lines.

### Design

**15. `testing/half_precision_ci.py:123` — this is a fork of `testing/known_failures.py`, not a generalization of it.**
`KnownFailureTracker`, `load_known_failures`, `mark_known_failures`, the tab-separated format, the strict-xfail contract and the terminal summary are duplicated from `testing/known_failures.py:41-138`, and the fork has *already* diverged in ways that are the bugs above: collect-only guard dropped, `pytest.fail.Exception` mapping dropped, different exception encoding, different reason string, `-k` allowed here but forbidden for MPS per AGENTS.md. The right depth is one mechanism parametrized by manifest dir + reason + issue URL, with recording added to it, and one documented contract instead of two that disagree.

### Minor

The description's manifest counts (603 float16 / 588 bfloat16) don't match the files (587 / 575).

ducha-aiki:
Implemented the reviewed Stage 1 known-failure design.

- Added explicit CPU float16/bfloat16 profiles that apply device and dtype before parametrization and isolate per-node Python, NumPy, and Torch RNG state.
- Replaced the importing two-column baseline with v1 phase-aware manifests containing opaque fully qualified exception identities and compatibility/provenance headers.
- Added exact lifecycle verification for focused and complete modes, including wrong phase/type, skips, competing xfails, teardown errors, incomplete collection, and stale nodes.
- Made regeneration candidate-only and transactional, with guarded destinations, completeness checks, categorized deltas, eager-RNG coverage auditing, atomic writes, and fresh-process candidate replay.
- Wired PR and scheduled CPU-half jobs to complete profile verification and documented the rare regeneration workflow. MPS behavior and its manifest are unchanged.

Verification:
- 51 half-precision CI harness tests passed.
- Float16 complete gate: 8249 passed, 3611 skipped, 624 xfailed, 5 xpassed.
- Bfloat16 complete gate: 8302 passed, 3563 skipped, 619 xfailed, 5 xpassed.
- Candidate record and fresh-process replay succeeded for both profiles before installation.
- All pre-commit hooks passed; type checking completed with only the existing torch.cuda.amp.custom_fwd deprecation diagnostic.

ducha-aiki:
## Review of #4154 — Add Linux CPU half-precision CI coverage

Reviewed the full diff (18 files, +3080/−33) and verified the high-severity findings by running them on a Linux x86-64 / Python 3.11 / torch 2.9.1 host — i.e. the exact environment the manifests were recorded on, so the failures below are not environment drift on my side.

The half-precision harness itself is careful work: per-node seeding, phase + exact exception identity in the manifest, completeness checks before a candidate is written, and the eager-RNG audit are all real improvements over a plain node-ID list. The problems are concentrated in how the new manifest is *validated* outside the two new CI jobs.

---

### Blocking

**1. `test_recorded_baseline_manifest` turns most of the CI matrix red** — `tests/test_half_precision_ci.py:690`

```python
entries = parse_manifest(profile, profile.manifest_path.read_text(...), current_environment(), source=...)
```

`enforce_compatibility` defaults to `True`, and `parse_manifest` (`testing/half_precision_ci.py:191`) *raises* on a `python` / `pytorch` / `os` / `arch` header mismatch. The checked-in headers pin `python: 3.11`, `pytorch: 2.9.1`, `os: Linux`, `arch: x86_64`. The test has no marker, no skip, and no `device_agnostic`, and `pixi run test` is `pytest tests/`, so it is collected on every leg.

Verified by flipping the header `os` to `Darwin` and running the test unchanged:

```
FAILED tests/test_half_precision_ci.py::test_recorded_baseline_manifest[float16]
  ValueError: .../cpu_float16.txt: os mismatch: expected 'Linux', found 'Darwin'
1 failed, 1 passed
```

Same `ValueError` for `python: 3.12`, `pytorch: 2.5.1`, `os: Windows`. Against the matrix in `pr_test_cpu.yml` that is red on every `windows-latest`/`windows-2022`/`macos-latest` leg, every 3.12/3.13 leg, and every torch 2.5.1 leg. It also breaks `tests-mps`, where the two new failures are absent from `testing/known_failure_xfails/mps_float32.txt`, so the blocking gate fails on an unlisted node ID.

Fix: pass `enforce_compatibility=False`, or skip the test when `current_environment()` does not match the header.

**2. `KnownFailureTracker` lost the `--collect-only` guard the existing tracker has** — `testing/half_precision_ci.py:527`

`testing/known_failures.py:81-90` sets `self._collect_only = session.config.getoption("collectonly")` and returns early from both `pytest_sessionfinish` and `pytest_terminal_summary`. The new class does neither, so under collect-only `pytest_runtest_logfinish` never fires, `pending` stays populated, and the session is failed. `FailureRecorder` *does* have the guard (`half_precision_ci.py:327`); the tracker does not.

Verified:

```
$ pytest tests/losses --xfail-known-failures --known-failure-profile=cpu-float16 -q --collect-only --dtype=float16
ERROR: 3 known half-precision failures need updates in .../cpu_float16.txt:
Did not run; rerun without selection or remove/update a stale manifest line:
tests/losses/test_dice.py::TestDiceLoss::test_averaging_macro[cpu-float16]
...
exit 1     # same command without --known-failure-profile: exit 0
```

Any contributor listing node IDs under a profile gets a bogus failure.

**3. Re-baselining is impossible after a torch or Python bump** — `testing/half_precision_ci.py:253`

`_previous_manifest_entries` reads the checked-in manifest with the default `enforce_compatibility=True`, and `FailureRecorder.__init__` calls it before any test runs; `conftest.pytest_configure` re-raises the `ValueError` as a `pytest.UsageError`. So `--record-known-failures` aborts at startup on exactly the environments a re-baseline exists to serve. TESTING.md documents this as the "initial profile, supported-environment update, or deliberate broad re-baseline" flow, which after a torch bump requires hand-editing the header of the file being regenerated first. The tracker already solves this correctly (`enforce_compatibility=self.mode == "complete"`); the recorder should read the *previous* manifest with enforcement off.

---

### Should fix before merge

**4. `AuditedEagerRngCall.classification` is dead state** — `testing/half_precision_eager_rng.py:227`

`eager_rng_calls_for_node` filters only on `node_prefixes`; `classification` has no reader anywhere in the repo. Four nodes are audited as `"value-independent"` (`tests/core/test_check.py::TestCheckShape::test_valid`/`test_invalid`, `test_param_validation.py::TestParamValidation::test_tuple_range_reader_errors`, `test_draw.py::TestDrawLine::test_point_size`) — the audit says the drawn value provably cannot change the outcome. If one of them ever fails under half precision, `_check_completeness` appends an `eager RNG feeds candidate node …` abort and the whole multi-hour recording run yields no candidate. Either gate the blocker on `classification != "value-independent"` or drop the field.

**5. `random.seed(0)` leaks into the rest of the session** — `tests/enhance/test_core.py:76`

`test_get_input_respects_size_and_max_elem` seeds the process-global `random` module and never restores it. Outside a `--known-failure-profile` run, `seed_half_precision_manifest_run` (`conftest.py:806`) yields without entering `_isolated_test_rng`, so a plain `pytest tests/` or `pixi run test-half` anchors every later `random.*` consumer — starting with `TestAddWeighted.get_input` in the same file — to a fixed stream. That undercuts the stated purpose of `test-half` ("finding tests that accidentally depend on shared RNG state"). The assertions (`ndim == 5`, `max(shape) <= 2`) hold for any seed, so `random.Random(0)` or a save/restore is enough.

**6. `test_gradcheck` still has the wart the PR fixed one hunk above** — `tests/enhance/test_core.py:114`

```python
@pytest.mark.parametrize("size", [2, 3])
def test_gradcheck(self, size, device):
    src1, ... = self.get_input(device, torch.float64, size=3, max_elem=5)
```

`size` is bound and discarded, so both node IDs run byte-identical float64 gradchecks — doubling the slowest test in the class while claiming 2-D/3-D coverage. This is the same `size=3` → `size=size` fix the diff already applies to `test_smoke`.

**7. The two new CPU-only tests need `@pytest.mark.device_agnostic`** — `tests/enhance/test_core.py:50,66`

`test_python_scalars_keep_opmath_precision` and `test_python_scalars_promote_integer_inputs` build plain CPU tensors and take neither the `device` nor the `dtype` fixture (the former shadows `dtype` with its own parametrize). `conftest.py:304-319` deselects `device_agnostic` items when CPU is not in `--device` for exactly this case. Without the marker the MPS gate and every CUDA run repeat them. `test_python_scalars_keep_opmath_precision` is the one that matters: `assert not torch.equal(expected, downcast)` is a *negative* assertion about the host's half-precision kernels, so running it on unmeasured backends is the riskier half.

**8. `add_weighted`'s three `if`/`else` pairs are now identical** — `kornia/enhance/core.py:67-83`

The precision fix itself is right — dropping `torch.tensor(alpha, dtype=src1.dtype)` lets PyTorch treat the Python scalar as a wrapped number and compute in opmath precision. But the rewrite left every arm the same statement:

```python
if isinstance(alpha, torch.Tensor):
    KORNIA_CHECK(...); output = src1 * alpha
else:
    output = src1 * alpha
```

Same for `beta`, and `gamma` returns `output + gamma` in the `if` and again after it. Association is unchanged, so the body reduces exactly to three `KORNIA_CHECK`s followed by a single `return src1 * alpha + src2 * beta + gamma`. As written, a future edit to one arm silently diverges from its twin.

**9. An aborted recording run can exit 0** — `testing/half_precision_ci.py:389`

`FailureRecorder.pytest_sessionfinish` appends to `abort_reasons` and returns without touching `session.exitstatus`; only the terminal summary says anything. If the run produced no test failures — e.g. the abort is an `eager RNG feeds candidate node …` on an otherwise clean tree — pytest exits `OK`, the stale candidate at `--record-known-failures=PATH` is silently preserved, and a script wrapping the documented two-command record/replay flow replays it believing the recording succeeded. `KnownFailureTracker` escalates in the same situation; the recorder should too.

---

### Worth considering

**10. The new x86-64 kernel literals inherit a fleet-wide claim from one host** — `tests/geometry/test_conversions.py:3773`

The key is `(torch_version, device.type, platform.machine(), dtype)`, but `platform.machine()` returns `x86_64` for every x86-64 host regardless of vector ISA. The comment this PR replaced warned about precisely that, and `tests.yml:119-129` documents the fleet dispatching different CPU kernels per host ("host-CPU-dependent: the CPU-autocast bf16 path dispatches AVX512 kernels some hosts don't actually execute"). Adding the `float32`/`float64` rows also activates the pin on every existing ubuntu torch-2.9.1 leg, not just the two new half jobs — those two are `0.0` exact agreement so they are low risk, but the `float16`/`bfloat16` rows are one-sample measurements. Passing here on one x86-64 host is one sample, not the fleet.

**11. The eager-RNG inventory pins line numbers across seven files** — `tests/test_half_precision_ci.py:193`

`assert len(discovered) == 23` plus an exact tuple comparison against `EagerRngCall(path, line, name)` literals. Inserting a line above line 305 of `tests/image/test_draw.py`, or adding a module-level `torch.rand` anywhere in `tests/`, fails with a tuple diff naming neither the edited file nor the fix. This PR already had to keep `tests/enhance/test_core.py`'s parametrize decorators at exactly lines 91-93 while rewriting the class around them. A path+name inventory with a count, or a diff-oriented failure message, would age better.

**12. Two record-destination tests depend on unguarded environment capabilities** — `tests/test_half_precision_ci.py:415,428`

`alias.symlink_to(checked_in)` raises `OSError` on Windows without `SeCreateSymbolicLinkPrivilege`. `test_record_destination_rejects_other_git_tracked_files` expects `ValueError(match="Git-tracked")`, but `validate_record_destination` raises `OSError("… Git is unavailable")` when `shutil.which("git")` is `None` and returns normally outside a Git checkout (source tarball, minimal container). Both fail rather than skip.

**13. Dead assertion** — `tests/test_half_precision_ci.py:720`

`assert "pytest-extra: --xfail-known-half-precision" not in workflow` — `--xfail-known-half-precision` has never existed in this repo (the flag is `--xfail-known-failures`). It can never fail, and it reads as coverage of the focus/complete distinction that it does not provide.

**14. The `-p` guard is redundant and bypassable** — `conftest.py:215`

`any(argument == "-p" or argument.startswith("-p") …)` — `startswith` subsumes the equality check, it rejects any argument beginning with `-p`, and unlike every other entry in `partial_options` it reads `config.invocation_params.args` rather than `config.option`, so a `-p` from `PYTEST_ADDOPTS` or `addopts` passes the guard while still loading the plugin.

---

### Design note (not blocking)

`testing/known_failures.py` already provides a `KnownFailureTracker`, a manifest directory, a tab-separated format, strict-xfail marking, pending tracking and a terminal summary behind `--xfail-known-failures` — and the MPS gate uses it. This PR adds a second subsystem with a second class of the same name, a different manifest directory, a different format, its own `ISSUE_URL` and four new CLI flags. The two have already diverged (the new one lost the collect-only guard from finding 2; the old one has no phase/header support). The new format is a strict superset of the old one, so adding phase + exception identity + a header to the existing tracker would have left one mechanism to maintain instead of two. Worth at least a follow-up issue so a maintainer fixing one baseline knows the other exists.

---

### What I verified

Linux x86-64, Python 3.11.14, torch 2.9.1+cpu — the manifests' recorded environment.

- Findings 1 and 2 reproduced with the commands and output quoted above; manifest restored afterwards.
- Findings 3, 4, 5, 6, 7, 8, 9, 13, 14 confirmed by reading the code paths and their call sites (`grep` confirms `classification` has no reader; `--xfail-known-half-precision` appears only in its own assertion).
- Finding 10 passes on this one x86-64 host, which is the point of the caveat rather than a refutation.
- I did not re-run the full strict float16/bfloat16 suites; the counts in the PR description are taken as reported.

ducha-aiki:
### Addendum to the review above: the duplicate tracker is a planned follow-up

The "Design note" at the end of my earlier comment should not be read as an open objection. The plan is to refactor the existing MPS mechanism (`testing/known_failures.py`) into this new one in a follow-up PR, so the two same-named `KnownFailureTracker` classes are a deliberate intermediate state, not an oversight. The new manifest format is a strict superset of the old one, so that direction works.

Two consequences that change what belongs in *this* PR rather than the follow-up:

**Finding 2 (missing `--collect-only` guard) should be fixed here, not deferred.** `testing/known_failures.py:81-90` is currently the only place that guard exists. If the new tracker is the survivor, deleting the old file removes the guard from the codebase entirely and `--collect-only` under any profile exits 1 permanently. Carrying the guard over is a two-line change now and an easy thing to lose during a refactor.

**Finding 3 (`_previous_manifest_entries` enforcing environment compatibility) is directly in the follow-up's way.** `testing/known_failure_xfails/mps_float32.txt` has no `phase`/exception-identity columns and no header, so the refactor has to regenerate it under the v1 schema — on macOS arm64, against a manifest whose header says `os: Linux`, `arch: x86_64`. As written, `--record-known-failures` aborts at startup there, so the follow-up would begin by hand-editing the header of the file it is generating. Reading the *previous* manifest with `enforce_compatibility=False` (as `KnownFailureTracker` already does for focus mode) unblocks it.

For whoever picks up the follow-up: the MPS contract is written down in three places that name the old mechanism by path, and all three move with it — the strict-baseline paragraph in `AGENTS.md` (which also cites #4159), the `tests-mps` comment block in `.github/workflows/pr_test_cpu.yml`, and the `test-mps` task in `pixi.toml`.

ducha-aiki:
Implemented the latest review feedback in commit 74345a3e.

Highlights:

- Made the recorded baseline test portable across the Python/Torch/OS matrix, allowed candidate rebaselining across environment changes, and made clean recorder aborts fail the run.
- Disabled runtime-outcome validation for --collect-only, and made complete-run -p detection use parsed pytest plugin options.
- Made eager-RNG enforcement honor value-independent classifications and made the audit inventory insensitive to line movement.
- Isolated random test state, fixed the parameterized gradcheck size, marked the CPU-only AddWeighted tests device-agnostic, simplified its identical arithmetic branches, and added portable Git/symlink test guards.
- Removed the dead workflow assertion.

The x86_64 ISA-table observation remains a fleet-coverage consideration rather than a demonstrated correctness bug, so this patch leaves the conservative kernel table unchanged.

Verification:

- pixi run pre-commit-all
- pixi run typecheck
- tests/test_half_precision_ci.py: 55 passed
- tests/enhance/test_core.py across CPU float16, bfloat16, float32, and float64: 101 passed
- strict CPU float16 profile: 8,253 passed, 624 xfailed, 5 xpassed
- strict CPU bfloat16 profile: 8,306 passed, 619 xfailed, 5 xpassed
- reviewer collect-only and PYTEST_ADDOPTS -p reproductions now behave as expected

Written by Codex (gpt-5.6) on behalf of @ducha-aiki

ducha-aiki:
## Re-review of #4154 (`74345a3e fix: address half-precision CI review`)

All fourteen findings from my earlier review are addressed, and I re-ran the ones that had reproductions rather than reading the diff alone. Two new items came out of the fix commit itself, one of them introduced by a fix.

### Confirmed fixed

**Finding 1 — `test_recorded_baseline_manifest`.** The fix is better than what I suggested: instead of skipping on drift, it deliberately mismatches all four fields and asserts the warning, so the test is environment-independent *and* exercises the non-enforcing path. Re-ran with the header rewritten to `os: Darwin`, `python: 3.13`, `arch: arm64`, `pytorch: 2.5.1` simultaneously — `2 passed` (it failed on a single-field mismatch before). The macOS/Windows/3.12/3.13/2.5.1 legs and the MPS gate are unblocked.

**Finding 2 — collect-only guard.** `pytest_sessionstart` + both hook guards, mirroring `known_failures.py`. Re-ran the original reproduction: `pytest tests/losses --xfail-known-failures --known-failure-profile=cpu-float16 -q --collect-only` now exits 0 with no ERROR block (was exit 1). `test_collect_only_does_not_validate_runtime_outcomes` pins it.

**Finding 3 — `_previous_manifest_entries`.** `enforce_compatibility=False`, with `test_previous_manifest_entries_allow_environment_change` asserting the warning still surfaces. The re-baseline flow is executable after a torch/Python bump.

**Finding 4 — `classification`.** Now gates the blocker (`== "value-dependent"`). The field earns its place: 9 of 23 entries still block a recording, 13 value-independent and 1 not-collected no longer do. Boundary test added for `TestCheckShape::test_valid`.

**Findings 5–8 — `tests/enhance/test_core.py` and `add_weighted`.** `random.Random(0)` via `monkeypatch` (auto-restored); `size=size`; both `device_agnostic` markers; `add_weighted` collapsed to three `KORNIA_CHECK`s and one `return src1 * alpha + src2 * beta + gamma`. Verified the marker actually takes effect — a CUDA run reports `deselected 3 device_agnostic test(s)`.

**Finding 9 — recorder exit status.** Escalates to `TESTS_FAILED` when `abort_reasons` is non-empty and the session would otherwise be `OK`, with `test_failure_recorder_aborted_clean_run_fails_session` covering exactly the clean-run case.

**Findings 12–14.** Symlink `try/except → skip`, Git availability + checkout guard, dead assertion removed, and `-p` now reads `config.option.plugins` (confirmed that is the correct dest) with a parametrized case added.

### New: introduced by the fix commit

**A. Dropping the line-number check let the audit's line numbers go stale in the same commit** — `testing/half_precision_eager_rng.py:181-198`

Finding 11's fix (comparing a `Counter` of `(path, name)` instead of exact `EagerRngCall` tuples) is the right call, but the audit still *stores* line numbers and the abort diagnostic still prints them:

```
eager RNG feeds candidate node {nodeid}: {entry.call.path}:{entry.call.line} ({entry.call.name}); move the draw into the test or a seeded fixture
```

The same commit shifted `tests/enhance/test_core.py` by +4 lines (two `device_agnostic` markers, `random.seed` → three lines), and nothing catches it any more:

```
audited entries pointing at a line with no draw on it:
   ('tests/enhance/test_core.py', 91, 'torch.randn') x 1
   ('tests/enhance/test_core.py', 92, 'torch.randn') x 1
   ('tests/enhance/test_core.py', 93, 'torch.randn') x 1
actual draw sites the audit does not name by line:
   ('tests/enhance/test_core.py', 95, 'torch.randn') x 1
   ('tests/enhance/test_core.py', 96, 'torch.randn') x 1
   ('tests/enhance/test_core.py', 97, 'torch.randn') x 1
```

Lines 91-93 are now `with pytest.raises(Exception):`, the call under it, and a blank line. These three are `value-dependent` and feed `TestAddWeighted::test_shape`, so they are live blockers — if that node ever fails under half precision, the abort tells the reader to move a draw that is not on the line it names. Scoped exactly to these three; every other audited entry still matches.

Suggestion: don't store the line in `AUDITED_EAGER_RNG_CALLS` at all. Match on `(path, name)`, and have `eager_rng_calls_for_node` (or the message builder) take the lines from `find_eager_rng_calls` at runtime, so the diagnostic is correct by construction rather than by a pin nothing checks.

**B. `-p no:cacheprovider` crashes the guard it was just taught to catch** — `conftest.py:209-210`

`partial_options` is built eagerly, and `--lf`/`--ff` are read as plain attributes while `--ignore`, `--ignore-glob` and `xdist` use `getattr(..., None)`. `-p no:cacheprovider` unregisters the plugin that *owns* `--lf`/`--ff`, so the dict raises before `active` is computed:

```
$ pytest tests/ --record-known-failures=/tmp/cand.txt --known-failure-profile=cpu-float16 -p no:cacheprovider
INTERNALERROR>   File "conftest.py", line 209, in _validate_complete_known_failure_run
INTERNALERROR>     "--lf": config.option.lf,
INTERNALERROR> AttributeError: 'Namespace' object has no attribute 'lf'
```

With the plugin intact the guard works correctly (`-p no:doctest` → `ERROR: --record-known-failures does not allow partial-selection options; remove: -p`). The predates this commit, but it matters more now: `-p` is a documented rejection, `no:cacheprovider` is a plausible thing to pass on a long recording run, and the new `("plugins", ["no:cacheprovider"])` test case exercises `_RecordingConfig`, a stub whose `options` dict always defines `lf` — so it passes while the real path raises an INTERNALERROR traceback instead of the actionable message. `getattr(config.option, "lf", None)` / `getattr(config.option, "failedfirst", None)` matches the style already used three lines below.

### Deliberately still open

Finding 10 (the four new `x86_64` rows in `tests/geometry/test_conversions.py`, where `platform.machine()` does not distinguish vector ISA) is unchanged. That is a reasonable call to make — the `float32`/`float64` rows are exact `0.0` agreement and low risk — but the `float16`/`bfloat16` literals remain one-host measurements keyed as if they were fleet-wide, so if a half job ever goes red there, that is the first thing to check rather than a kornia regression.

### What I verified

Linux x86-64, Python 3.11.14, torch 2.9.1+cpu; CUDA runs on a local NVIDIA GPU.

- `tests/test_half_precision_ci.py` — 55 passed.
- `tests/enhance/test_core.py` — 29 passed (float32), 53 passed (float16+bfloat16), 50 passed / 3 deselected on CUDA float32+float64, and 50 passed / 3 deselected on CUDA float16+bfloat16 with `--isolate-half-precision`.
- `-k "dynamo or compile"` under `KORNIA_TEST_OPTIMIZER=inductor` — 1 passed; `--doctest-modules kornia/enhance/core.py` — 2 passed; `ruff check` and `ruff format --check` clean on all six changed files.
- Findings 1, 2 and the two new items reproduced with the commands quoted above; the manifest was restored afterwards.
- I did not re-run the full strict float16/bfloat16 suites; those counts are taken as reported.

ducha-aiki:
Pushed `fb27282f` addressing findings A and B from the re-review.

**A (stale audit lines).** Corrected `tests/enhance/test_core.py`'s audited lines (91-93 → 95-97), and restored the inventory test to a full `(path, line, name)` multiset comparison instead of `(path, name)` only — this keeps Finding 11's original fix (no fragile ordering across unrelated files) while still catching a stale line on the file it names, with the stale/missing entries printed directly. Verified it catches the exact regression by reverting the line fix alone and confirming the test fails with `audited entries with no matching call site (stale line?): [('tests/enhance/test_core.py', 91, 'torch.randn'), ...]`.

**B (`-p` guard crash).** `config.option.lf`/`.failedfirst` are now read via `getattr(..., None)`, matching the other entries in the same dict. Verified by reverting the fix alone: the added regression test (`test_record_mode_rejects_minus_p_when_cacheprovider_is_disabled`, which builds a `config.option` without `lf`/`failedfirst` to match what a real `-p no:cacheprovider` run produces) fails with `AttributeError: 'types.SimpleNamespace' object has no attribute 'lf'` on the old code and passes on the fix.

`tests/test_half_precision_ci.py` — 56 passed (was 55; +1 new test). `tests/enhance/test_core.py` — 29 passed (float32), 53 passed (float16+bfloat16). `ruff check`/`ruff format --check` clean on all three touched files.

Written by Claude (Sonnet 5) on behalf of @ducha-aiki

PR #4155, opened 2026-09-01
[docs] revamp the documentation site

shijianjian:
Rebuild the documentation on pydata-sphinx-theme, and fix a long tail of content bugs along the way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@ docs/source/_static/*jpg
 docs/source/_static/img/*jpg
 docs/source/_static/scripts/*
 docs/source/get-started/performance.rst
+docs/source/community/adoption.rst
+docs/source/_generated/
 *DS_Store
 *jpg
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Changed
+### Added
 
 * The documentation site was rebuilt on `pydata-sphinx-theme` with a top navigation bar
   (Learn / API / Models plus Ecosystem, About and Support menus), a redesigned landing page, a
   restructured API reference with per-topic subpages, and a long list of fixed doc examples and
-  removed dead interactive demos. The previous furo layout remains available with
-  `KORNIA_DOCS_THEME=furo`. (#4155)
+  removed dead interactive demos. Old deep links into the split module pages
+  (e.g. `augmentation.module.html#kornia.augmentation.RandomAffine`) are forwarded to the subpage
+  that now documents the object, so existing links keep resolving. The previous furo layout remains
+  available with `KORNIA_DOCS_THEME=furo`. (#4155)
+* An **Adoption** page (`community/adoption`) listing the most-starred GitHub repositories and
+  packages that depend on kornia, rendered at build time from `docs/source/_data/dependents.json`;
+  `docs/fetch_dependents.py` refreshes that snapshot from GitHub's dependency graph. (#4155)
 
 ### Breaking changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+* The documentation site was rebuilt on `pydata-sphinx-theme` with a top navigation bar
+  (Learn / API / Models plus Ecosystem, About and Support menus), a redesigned landing page, a
+  restructured API reference with per-topic subpages, and a long list of fixed doc examples and
+  removed dead interactive demos. The previous furo layout remains available with
+  `KORNIA_DOCS_THEME=furo`. (#4155)
+
 ### Breaking changes
 
 * `extract_patches_from_pyramid` now samples ordinary-sized inputs once from a packed pyramid atlas instead of
--- a/docs/fetch_dependents.py
+++ b/docs/fetch_dependents.py
@@ -55,7 +55,7 @@ KEEP_TOP = 300  # entries stored per kind, ranked by stars
 DELAY_S = 1.0  # between requests; GitHub serves this page without auth but rate-limits bursts
 
 _COUNT_RE = {
-    "repositories": re.compile(r">\s*([\d,]+)\s*Repositor(?:y|ies)\s*<"),
+    "repositories": re.compile(r">\s*([\d,]+)\s*Repositor(?:y|ies)\s*<"),  # codespell:ignore repositor
     "packages": re.compile(r">\s*([\d,]+)\s*Packages?\s*<"),
 }
 _ROW_SPLIT_RE = re.compile(r'<div class="Box-row[^"]*"')
@@ -66,11 +66,13 @@ _NEXT_RE = re.compile(r'href="[^"]*dependents_after=([A-Za-z0-9_=-]+)[^"]*"[^>]*
 
 def _get(url: str, retries: int = 5) -> str:
     """Fetch a page, backing off on rate limiting and transient server errors."""
+    if not url.startswith("https://github.com/"):
+        raise ValueError(f"refusing to fetch {url!r}: only GitHub HTTPS pages are crawled")
     delay = 30.0
     for attempt in range(retries):
-        request = urllib.request.Request(url, headers={"User-Agent": USER_AGENT})
+        request = urllib.request.Request(url, headers={"User-Agent": USER_AGENT})  # noqa: S310 - scheme checked above
         try:
-            with urllib.request.urlopen(request, timeout=60) as response:
+            with urllib.request.urlopen(request, timeout=60) as response:  # noqa: S310
                 return response.read().decode("utf-8", errors="replace")
         except urllib.error.HTTPError as exc:
             if exc.code not in (429, 500, 502, 503, 504) or attempt == retries - 1:
@@ -171,7 +173,8 @@ def main(argv: list[str] | None = None) -> int:
             continue  # nothing new; keep the previous entry untouched
         data[kind] = merge(data, kind, total, rows, complete)
         print(
-            f"  {kind}: {data[kind]['count']:,} total, {len(rows)} rows fetched, complete={complete}", file=sys.stderr
+            f"  {kind}: {data[kind]['count']:,} total, {len(rows)} rows fetched, complete={complete}",
+            file=sys.stderr,
         )
 
     data["generated_at"] = datetime.now(UTC).strftime("%Y-%m-%d")
--- a/docs/fetch_dependents.py
+++ b/docs/fetch_dependents.py
@@ -170,7 +170,9 @@ def main(argv: list[str] | None = None) -> int:
         if total is None and not rows:
             continue  # nothing new; keep the previous entry untouched
         data[kind] = merge(data, kind, total, rows, complete)
-        print(f"  {kind}: {data[kind]['count']:,} total, {len(rows)} rows fetched, complete={complete}", file=sys.stderr)
+        print(
+            f"  {kind}: {data[kind]['count']:,} total, {len(rows)} rows fetched, complete={complete}", file=sys.stderr
+        )
 
     data["generated_at"] = datetime.now(UTC).strftime("%Y-%m-%d")
     data["source"] = DEPENDENTS_URL
--- a/docs/fetch_dependents.py
+++ b/docs/fetch_dependents.py
@@ -1,0 +1,184 @@
+# LICENSE HEADER MANAGED BY add-license-header
+#
+# Copyright 2018 Kornia Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Collect kornia's dependents from GitHub's dependency graph into docs/source/_data/dependents.json.
+
+GitHub shows "Used by" on the repository page and lists every dependent at
+``github.com/kornia/kornia/network/dependents``, but exposes neither through the REST or GraphQL
+API. This script walks that HTML listing -- one page of 30 per request, cursor-paginated -- for
+repositories and for packages, and stores the exact totals plus the most-starred entries.
+
+Run it by hand and commit the refreshed JSON -- for instance before a release; there is deliberately
+no scheduled job for it. ``generate_adoption.py`` renders the file into the Adoption page at
+docs-build time::
+
+    python docs/fetch_dependents.py                 # full crawl (~330 requests, ~10 min)
+    python docs/fetch_dependents.py --max-pages 5   # quick refresh of the totals
+
+A partial crawl (``--max-pages`` or a network failure) merges into the existing file, so the totals
+are always current and the ranking only ever gets more complete.
+"""
+
+from __future__ import annotations
+
+import argparse
+import html
+import json
+import re
+import sys
+import time
+import urllib.error
+import urllib.request
+from datetime import UTC, datetime
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parents[1]
+OUT = REPO / "docs" / "source" / "_data" / "dependents.json"
+
+DEPENDENTS_URL = "https://github.com/kornia/kornia/network/dependents"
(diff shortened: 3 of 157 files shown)


ducha-aiki:
Looks good, but let's me try Fable 5.1 on this, it might find a couple of things to fix :)

ducha-aiki:
## Review of #4155 (head `e97e0cf8`)

Reviewed from a worktree at the PR head. What I ran: the autodoc-directive inventory of `docs/source` on `main` vs this branch, a diff of the Sphinx `objects.inv` between the live `latest` docs and the RTD preview build, a `-W` build with the `KORNIA_DOCS_THEME=furo` fallback, every `code-block:: python` snippet the PR touched executed in a subprocess, and a pass over the added prose against the library source.

**Verdict: request changes.** One structural issue (deep-link anchors, item 1) and a handful of content errors (items 2, 3 and 5); the rest is minor. The restructure itself is sound: all 716 autodoc directives survive, the `objects.inv` diff shows 1570 documented Python objects on `main`, 1577 on the branch, and zero missing. CI is green including the RTD build, and the furo fallback also builds cleanly under `-W`.

### 1. Must fix: 727 documented objects changed page, so every existing deep link into the old landing pages now lands without its anchor

`objects.inv` diff (live `latest` vs the preview at `kornia--4155.org.readthedocs.build`):

| old page | objects that moved off it |
|---|---|
| `feature.html` | 188 |
| `color.html` | 145 |
| `augmentation.module.html` | 140 |
| `filters.html` | 63 |
| `image.html` | 57 |
| `enhance.html` | 55 |
| `losses.html` | 54 |
| `metrics.html` | 25 |

Spot check, `id="..."` present in the rendered page:

```
augmentation.module.html#kornia.augmentation.RandomAffine   main=1 pr=0
color.html#kornia.color.rgb_to_grayscale                     main=1 pr=0
feature.html#kornia.feature.LoFTR                            main=1 pr=0
filters.html#kornia.filters.gaussian_blur2d                  main=1 pr=0
enhance.html#kornia.enhance.adjust_brightness                main=1 pr=0
losses.html#kornia.losses.ssim_loss                          main=1 pr=0
metrics.html#kornia.metrics.ssim                             main=1 pr=0
image.html#kornia.image.Image                                main=1 pr=0
```

These are the URLs that GitHub issues, StackOverflow answers, papers and blog posts have been linking to for years (`.../en/latest/augmentation.module.html#kornia.augmentation.RandomAffine` is probably the single most-shared kornia docs URL). After this lands they all open the new index page scrolled to the top with no indication where the object went. RTD page redirects cannot help because the path is unchanged and the fragment never reaches the server.

Since the old pages still exist (they are now section indexes), a client-side fragment redirect fixes this completely and costs ~30 lines in `conf.py` plus ~10 in `custom.js`:

- In an `html-page-context` hook, for the eight old docnames only, read `app.env.domaindata["py"]["objects"]` and emit a `<script>` with `{fullname: docname}` for every object whose `docname` is not the current page (~30 KB total across the eight pages, nothing on the other ~150).
- In `custom.js`, on `DOMContentLoaded`: if `location.hash` is set, the page has no element with that id, and the map has the name, `location.replace(root + map[name] + ".html" + location.hash)`.

Anchors on the new pages are Sphinx's normal `id="kornia.color.rgb_to_grayscale"`, so the fragment carries over unchanged. Please add a short note to the CHANGELOG entry about the redirect too, since it is user-visible behaviour of the site.

### 2. Should fix: stale "interactive demo" wording after the gradio embeds were removed

Copilot caught the one visible in prose (`applications/image_denoising.rst:24`, "or try the interactive demo below."). Three more live in `.. meta:: :description:` tags, which are exactly what now gets copied into the new `og:description` / `twitter:description` cards by `_inject_social_metatags`, so they show up in every shared link:

- `docs/source/geometry.homography.rst:5` — "The module also offers an interactive demo to explore homography warping in real time."
- `docs/source/morphology.rst:5` — "Try the interactive demo on Hugging Face Spaces."
- `docs/source/enhance.rst:5` — "Explore interactive demos on Hugging Face Spaces."

`grep -rnE 'interactive demo|gradio|hf\.space' docs/source kornia` should come back empty once these are gone.

### 3. Should fix: `og:title` is double-escaped

Sphinx hands `context["title"]` to `html-page-context` already HTML-escaped, and `_inject_social_metatags` escapes it again. On the preview, `get-started/conventions.html` renders:

```
<title>Conventions &amp; Pitfalls &#8212; Kornia</title>                          (correct)
<meta property="og:title" content="Conventions &amp;amp; Pitfalls - Kornia" />   (shows as "&amp; Pitfalls" in a link card)
```

Same for `twitter:title`, and for the three `api.rst` section titles with an `&`. `title = html.unescape(context.get("title") or project)` before the existing `esc(title)` fixes it; the description path already does this unescape.

### 4. Minor

- **Mobile navigation has no dropdowns.** pydata renders the header nav twice: once in `.bd-header` and once in the primary sidebar for narrow viewports (`sidebar-header-items`). `custom.js` decorates only `.bd-header .navbar-nav > li.nav-item > a.nav-link`, so on a phone "Support" and "About" are plain links and "Ecosystem" is absent. Dropping the `.bd-header` scope (or querying both) is enough; the CSS already keys on the class names.
- **Navbar dropdown targets are hardcoded in JS** (`community/sponsor.html`, `get-started/governance.html`, ...). Sphinx's `-W` validates `:doc:` links but not these, so a future page rename silently 404s the menu. All seven exist today; a comment in `conf.py` next to the toctree, or generating the list from `html_context` in `conf.py`, would keep them in one place.
- **`og:url` is always the `latest` URL** (`html_baseurl` is fixed). On a `stable` or versioned build the card links to `latest`. That is arguably the right canonical, just noting it is deliberate.
- **CHANGELOG uses a new `### Changed` heading.** The file's existing Unreleased sections are `Added`, `Bug fixes`, `Breaking changes` and AGENTS.md asks for the headings the file already uses. Keep a Changelog does define `Changed`, so this is fine if that is the intent; otherwise `Added` fits a new site.
- `docs/source/index.rst` ships a page-local `<style>` block via `raw:: html` to hide the secondary sidebar; `secondary_sidebar_items["index"] = []` and `html_sidebars["index"] = []` in `conf.py` already do the same thing for the sidebar content. Not wrong, just two mechanisms for one effect.

### 5. Content errors in the new and rewritten prose

Checked the added claims against the library source (signatures via `inspect`, greps of the referenced pages). These are wrong:

- **`get-started/onnx.rst:80-81` and `:153` use `io_map=`; the parameter is `io_maps`.** `inspect.signature(ONNXSequential.__init__)` has `io_maps: Optional[list[tuple[str, str]]]` and no `**kwargs`, so the snippet raises `TypeError`. Line 74 of the same page spells it correctly, so the page contradicts itself. (Inherited from the old `onnx.rst`, but this is a new file and the FAQ line is new.)
- **`get-started/gpu-acceleration.rst:34-35`** says the performance page covers "the compiled and mixed-precision paths". `performance.rst` has no compile / half / autocast / float16 content at all; its tables are eager kornia vs torchvision v2 / albumentations / OpenCV / PIL.
- **`get-started/gpu-acceleration.rst:38`** says MPS caveats are "listed on the precision page". `precision.rst` never mentions MPS, Apple or Metal.
- **`get-started/onnx.rst:15-16`** advertises "an asyncio-based API". There is no `async` anywhere in `kornia/onnx/` or `kornia/core/mixin/onnx.py`.
- **`models/index.rst:8`** says every model "downloads its weights on first use". MobileViT is in the table and has no pretrained weights (`kornia/models/vit_mobile.py` has no `pretrained` at all).

Softer, worth a hedge: `api.rst:8` "All functions accept batched `(B, C, H, W)` float tensors in `[0, 1]`" and `models.rst:8` "every builder returns a regular `nn.Module`" are rules of thumb stated as absolutes (`BoxMotTracker` is a plain class on the same page; point, box, LAF and homography functions take other shapes).

Everything else I checked held up: every augmentation, model-builder, I/O, feature, matching and ONNX signature named in the new examples exists as written; the stereo `Q` matrix and reprojection equations, the calibration `r^2` and `k_5` fixes, the `pyproject.toml` requirement claims, the benchmark numbers on the landing page (1124 / 3153 img/s are the `RandomGaussianBlur @ 32` M1 CPU / MPS cells of `performance.rst`; 0.015 s / 0.177 s match the ONNX page), and all `_static/img` and `data/` assets referenced.

### What I verified and found fine

- All 716 autodoc directives present on `main` are present on the branch (two moved under a different `currentmodule`, resolve to the same objects).
- `-W` build with `KORNIA_DOCS_THEME=furo` from the worktree: success, zero Sphinx warnings.
- Executed every `code-block:: python` in the touched pages against the current `.venv` (torch 2.9.1). 42 ran clean (including the landing-page ONNX chain, `RTDETRDetectorBuilder`, `LoFTR`, `ImageRegistrator`, the `AugmentationSequential` examples). 15 failed for environmental reasons only: placeholder input files (`image.jpg`, `left.jpg`), TensorFlow/JAX not installed, the `segmentation_models_pytorch` install prompt, placeholder SAM checkpoint paths, and the illustrative local `operators/...` load in `onnx.rst`. The Hub operators the pages cite (`GrayscaleToRgb`, `RgbToGrayscale`, `Resize_512x512`) all resolve with HTTP 200 and load.
- The `_inject_social_metatags` hook: verified on the preview that pages with a `.. meta:: :description:` get it as `og:description` and pages without get the default, with proper escaping and the 300-char truncation.
- Pre-existing, not from this PR: `kornia.to_numpy()` on `multi-framework-support.rst` fails with `TypeError: unhashable type` inside `ivy.transpiler` on ivy 1.0.0.5; the snippet is unchanged here.

ducha-aiki:
@shijianjian Please, check Fable's review, it indeed worth it.

ducha-aiki:
## Review of #4155 — docs site revamp

Big, careful piece of work. I reviewed the diff against `d7d6a2f` and cross-checked against the live preview build at `kornia--4155.org.readthedocs.build`. The riskiest thing about a 157-file docs restructure is silently losing API pages or breaking years of external deep links, and **neither happened** — that part is genuinely well executed.

### Verified

- **No documented API was lost.** Extracting every `autoclass`/`autofunction`/`autodata`/`autoexception` target from both revisions gives **721 objects on `main` and 721 on the branch — zero added, zero removed**, and no object is documented on two pages.
- **No file was deleted or renamed anywhere in the PR** (`git diff --diff-filter=D` and `-R` are both empty), so no existing docs URL 404s. The 63 new subpages are pure additions, and the pages that became section indexes are kept (`augmentation.module` etc. as `:orphan:`).
- **The deep-link redirect actually covers what moved.** 394 objects moved off exactly eight old index pages; `_SPLIT_INDEX_PAGES` covers all eight, and every moved object matches its page's declared prefix (0 mismatches). On the live build the JSON map is emitted at line 33 of `<head>` and `custom.js` loads at line 77 — the ordering the redirect depends on — and `data-content_root` is present for the relative-path rewrite.
- **Gradio removal is complete**: 12 `gradio-app` references on `main` → 0, including the one in `kornia/geometry/transform/affwarp.py`, plus the `gradio.js`/`iframe-resizer` script tags. No orphaned embeds left behind.
- **Every `kornia.*` reference in the changed prose resolves** against the installed library, including `ImageRegistrator("similarity")`, `load_image(..., device=)`, `ONNXSequential.as_cuda/as_cpu/export/create_session`, `io_maps`, `K.contrib.FaceDetector`, and the new `crowd.jpg` asset (HTTP 200).
- `build-docs` runs `sphinx -W`, and both the docs job and the RTD build are green — so all internal refs, toctrees and the new `_check_navbar_menus` consistency check pass.

### Worth fixing before merge

1. **`docs/source/index.rst:236` — the landing page advertises a wrong number.** The card renders `|count-image-processing| operators · kornia.filters`, but `_landing_page_counts` defines `image-processing` as `filters + color + enhance + morphology` (67 + 82 + 55 + 8 = **212** today). So the live page says *"212 operators · `kornia.filters`"* for a card that links only to `filters`, which has 67 — and colour/enhance already have their own card, so it double-counts. Either use `surface["kornia.filters"]` here or relabel the card.

2. **`docs/source/_static/js/custom.js:184-211` — the hero carousel restarts after the visitor picks a tab**, contradicting the comment's "choosing a tab stops it for good". `pointerenter` → `pause()` sets `paused = true`; the click then fires `change` → `stop()`, which clears the timer but leaves `paused === true` and records no "stopped" state; the following `pointerleave` → `resume()` sees `paused` truthy and calls `schedule()` again. Every mouse click goes through that sequence, so it's the normal path, not an edge case (same via `focusin`/arrow key/`focusout`). `stop()` also drops `is-autoplaying`, which `resume()` never re-adds, so the styling and the behaviour then disagree. A `stopped` flag checked in `pause`/`resume` fixes it.

3. **`docs/source/community/community.rst:46` — `https://librecv.org` is dead** (NXDOMAIN, `curl` exit 6). Newly added paragraph, so this ships a broken outbound link on the Community page. `doc-links.yml` runs `--offline` on PRs, which is why CI is green; the weekly networked run would catch it later.

4. **`docs/source/get-started/gpu-acceleration.rst:35`** says the performance page "shows measured eager-mode comparisons against torchvision, albumentations, OpenCV and PIL on **CPU, CUDA and Apple silicon**", but `benchmarks/results/0.9.0rc1/` only holds `apple-m1`/`apple-m4` cpu+mps, and the only "cuda" on the rendered page is inside the contribute command. I understand broader benchmark coverage is in flight, so this resolves itself once the CUDA runs land — until then it's one sentence that is ahead of the data. (This is the only such claim in the diff; `augmentation.rst:51-66` and `geometry.subpix.rst:25` are both framed honestly.)

5. **`pydata-sphinx-theme` is unpinned** in `[project.optional-dependencies].docs`, while `pydata.css` and `custom.js` couple hard to theme internals — `#pst-secondary-sidebar`, `.bd-navbar-elements.navbar-nav`, `#pst-page-toc-nav > ul`, `.bd-sidebar-primary`, `button.sidebar-toggle.secondary-toggle`. `uv.lock` protects pixi/CI, but Read the Docs installs `pip install .[docs]` without the lock, so an upstream theme release can silently break the navbar dropdowns, the augmented right-hand TOC and the landing page's hidden rail. A bound (e.g. `pydata-sphinx-theme>=0.16,<0.18`) would make that a build failure instead of a live regression.

### Smaller items

- **`docs/source/conf.py:135`** — `pkgutil.walk_packages(kornia.__path__, "kornia.")` has no `onerror`. The surrounding `try/except` only guards *this* function's `importlib.import_module`; `walk_packages` imports each sub**package** itself to recurse, and with `onerror=None` it swallows `ImportError` but **re-raises everything else**. A `kornia.*` package raising `RuntimeError`/`OSError` at import in a docs env would fail the whole build, when the intent is clearly "skip it". `onerror=lambda name: None`.
- **`docs/generate_benchmarks.py:159/192`** — the guard checks `cpu <= 0` but not `gpu`, and line 192 computes `max(gpu, cpu) / min(gpu, cpu)`. A result row with `throughput_per_s: 0` on the accelerator gives `ZeroDivisionError`, and since `main()` runs at conf-read time that aborts the docs build rather than just omitting the chart.
- **`docs/fetch_dependents.py:136/168`** — `--out` is honoured on write but not on read: `load_existing()` reads the module-level `OUT`. `--out other.json --max-pages 5` merges against the default snapshot and then overwrites `other.json` wholesale, which is the opposite of the "a partial crawl merges into the existing file" contract in the docstring.
- **`docs/generate_examples.py:758`** — `torch.stack([...for f in faces if f.score > 0.7])` raises on an empty list. It's the last block of `main()`, so a checkpoint or asset change would abort the script after everything else has been regenerated. (It works today: 61/196 detections clear the threshold.)
- **`docs/generate_adoption.py:59-64`** — the docstring promises "Never rounds up past the real figure", but the 1,000-9,999 branch rounds to nearest: 1,290 → `"1.3K"`, and 9,999 → `"10K"`. Currently harmless since `repositories_rounded` isn't rendered anywhere, but either truncate or drop the promise.
- **Unused substitutions**: `count-losses-metrics`, `count-operators`, `count-dependents-rounded`, `count-dependents-listed`, `count-dependents-date` are built in `rst_epilog` but referenced by no page.
- **Navbar a11y**: the dropdowns are keyboard-*reachable* thanks to `:focus-within`, but carry no `aria-expanded`/`aria-haspopup` and no Escape-to-close, and the Ecosystem trigger is an `<a href="#">` with `preventDefault` where a `<button>` would announce correctly.
- **Untested examples**: `pixi run doctest` is `pytest --doctest-modules kornia/`, so none of the ~8k lines of new `.. code-block:: python` are executed by CI. `sphinx.ext.doctest` is already in `extensions` — converting the load-bearing snippets to `testcode::` and adding a `-b doctest` build would stop these drifting. Not for this PR, but worth an issue.

### Notes, not objections

- The Adoption page is a hand-refreshed snapshot with deliberately no scheduled job. That's a reasonable call given it's HTML scraping, and the page does disclose the date — just be aware the landing page and `about.rst` quote those numbers, so they go stale together.
- Keeping `KORNIA_DOCS_THEME=furo` as a fallback is a nice hedge while the redesign settles.

Nice work — the URL-stability engineering here is the part that would have been easy to get wrong, and it's solid.

PR #4156, opened 2026-09-01
fix(feature): apportion MultiResolutionDetector quotas with largest remainder

### LINKED ISSUE #4101: MultiResolutionDetector: truncated per-level quotas starve a small num_features
`MultiResolutionDetector.detect` splits `num_features` across the pyramid levels with a
proportional share, then truncates each share independently. For a small `num_features` the
truncation loses most of the budget, and the detector returns several views of one feature
instead of spreading over the image.

## Reproduction

Five identical blobs at known centres on a 96x96 image:

```python
import torch
from kornia.feature import MultiResolutionDetector
from kornia.feature.responses import BlobHessian

img = torch.zeros(1, 1, 96, 96)
blobs = [(20, 20), (20, 70), (48, 48), (76, 20), (76, 70)]
for y, x in blobs:
    img[:, :, y - 2:y + 3, x - 2:x + 3] = 1.0

for nf in (3, 5, 10, 30):
    lafs, resps = MultiResolutionDetector(BlobHessian(), num_features=nf)(img)
    c = lafs[0][resps[0] != 0][:, :, 2]
    covered = sum(
        1 for y, x in blobs
        if c.numel() and ((c - torch.tensor([float(x), float(y)])).abs().max(1).values < 4).any()
    )
    print(f"num_features={nf}: {c.shape[0]} detections, {covered} of the 5 blobs covered")
```

```
num_features=3:  3 detections, 1 of the 5 blobs covered
num_features=5:  5 detections, 3 of the 5 blobs covered
num_features=10: 9 detections, 5 of the 5 blobs covered
num_features=30: 9 detections, 5 of the 5 blobs covered
```

Asking for 3 features on an image with 5 equally strong ones returns three views of a single
blob. Identical on `main` at 9abf06d9 and on the head of #4098.

## Cause

`kornia/feature/scale_space_detector.py:645-646`

```python
shares: List[float] = [x / tmp for x in num_features_per_level]
num_features_per_level = [int(x) for x in shares]
```

With the default configuration the six shares are `num_features * [0.508, 0.254, 0.127, 0.063,
0.032, 0.016]`. Every fractional part is discarded, so the quotas sum to well under
`num_features` for any small request:

```
num_features=1  -> [0, 0, 0, 0, 0, 0]   sum 0
num_features=2  -> [1, 0, 0, 0, 0, 0]   sum 1
num_features=63 -> [32, 15, 7, 3, 1, 0] sum 58
```

Only the coarser levels' cumulative sums keep the result non-degenerate at all, and the finest
level's quota dominates, so the detector effectively searches one scale.

#4098 added a fallback for the `sum == 0` case alone (`num_features=1`, which returned an empty
result), deliberately narrow so it could not change any working configuration. The general case
is this issue.

## Suggested fix

Largest-remainder (Hamilton) apportionment: floor each share, then give the leftover slots to the
largest fractional parts, so the quotas sum to exactly `num_features`. This changes the per-level
quotas for every `num_features` and therefore the candidate pool that reaches the final top-K, so
it wants its own PR with before/after detection counts, not a rider on a bug fix.

Found while reviewing #4098; deliberately left out of it as out of scope.

mohansree14:
## What

`MultiResolutionDetector.detect` splits `num_features` across pyramid levels with a proportional share, then truncates each share independently (`int(x) for x in shares`). For a small `num_features` this discards most of the fractional budget, so the request effectively collapses onto a single scale.

Reproduction from the issue (five well-separated, equally strong blobs on a 96x96 image):

```
num_features=3:  3 detections, 1 of the 5 blobs covered   (before)
num_features=3:  3 detections, 2 of the 5 blobs covered   (after)
```

## Fix

Switch to largest-remainder (Hamilton) apportionment: floor each level's share, then hand the shortfall — `num_features` minus the sum of the floors — to the levels with the largest fractional remainder, one slot each. Quotas now always sum to exactly `num_features` instead of losing budget to truncation.

This generalizes #4098's narrower `num_features=1` fallback (which handed the single slot to the largest-share level when every quota floored to zero); that is now just the case where the shortfall equals `num_features`, so the dedicated branch is removed.

## Testing

- Added `test_small_request_spreads_across_scales`, which fails on the pre-fix code (`covered=1`) and passes after the fix (`covered=2`) — verified both ways.
- All 64 existing tests in `tests/feature/test_scale_space_detector.py` pass, including the existing `num_features=1` regression test for #4098.
- `ruff check` / `ruff format` clean on the changed files.
- Added a `CHANGELOG.md` entry under `Unreleased` / `Bug fixes`.

Note: I don't have `pixi` in my local environment, so I verified with a throwaway `uv` venv (torch 2.14 CPU) instead of the project's usual toolchain — happy to re-run anything specific if CI flags something that setup wouldn't catch.

Fixes #4101

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `MultiResolutionDetector.detect` apportions `num_features` across pyramid levels with largest-remainder
+  (Hamilton) apportionment instead of flooring each level's fractional share independently (#4101). The floors
+  used to discard every fractional part, so a small `num_features` could round the whole apportionment down: with
+  the default configuration the six shares are `0.508 .. 0.016`, and every level but the finest lost its entire
+  quota to truncation, so the detector effectively searched a single scale. The shortfall -- `num_features` minus
+  the sum of the floors -- is now handed to the levels with the largest fractional remainder, one slot each, so
+  the quotas always sum to exactly `num_features` and stay spread across scales. #4098's narrower `num_features=1`
+  fallback, which handed the request's one slot to the largest-share level when every quota floored to zero, is
+  the special case where the shortfall equals `num_features` and is superseded by the general apportionment.
+
 * Make YUV and XYZ transformations compute integer inputs in `float32` instead of truncating their kernels, and
   preserve directly constructed `float64` coefficients. This makes `rgb_to_yuv(uint8)` return `float32` on the
   input's original 0--255 scale, fixes signed-integer YUV results, and removes the existing float32 coefficient loss
--- a/kornia/feature/scale_space_detector.py
+++ b/kornia/feature/scale_space_detector.py
@@ -839,15 +839,19 @@ class MultiResolutionDetector(nn.Module):
             nf = self.num_features * factor_points ** (-1 * (idx_level - self.num_upscale_levels))
             num_features_per_level.append(nf)
         shares: List[float] = [x / tmp for x in num_features_per_level]
+        # Largest-remainder (Hamilton) apportionment: `shares` sums to `self.num_features` (up to
+        # float error), but flooring each one independently discards every fractional part, so the
+        # floors can sum to well under `self.num_features` -- to zero at `num_features=1`, where the
+        # default six shares are 0.508 .. 0.016. The finest level's quota also dominates every other
+        # one's, so a small request degenerates to querying a single scale. Hand the shortfall --
+        # `self.num_features` minus the sum of floors -- to the levels with the largest fractional
+        # remainders, one slot each, so the quotas always sum to exactly `self.num_features` and stay
+        # spread across scales instead of collapsing onto the finest level.
         num_features_per_level = [int(x) for x in shares]
-        # Each quota truncates independently, so a small `num_features` can round the whole
-        # apportionment to zero: with the default configuration that happens at `num_features=1`,
-        # where the six shares are 0.508 .. 0.016. Every level would then be queried for zero
-        # candidates and the result padded with a dummy, on an image full of real maxima. Hand the
-        # one slot to the level with the largest share. This fires only when truncation lost
-        # everything -- an apportionment that already gives out a slot is left exactly as it was.
-        if self.num_features > 0 and sum(num_features_per_level) == 0:
-            num_features_per_level[max(range(len(shares)), key=shares.__getitem__)] = 1
+        shortfall = self.num_features - sum(num_features_per_level)
+        by_remainder = sorted(range(len(shares)), key=lambda i: shares[i] - num_features_per_level[i], reverse=True)
+        for idx_level in by_remainder[:shortfall]:
+            num_features_per_level[idx_level] += 1
 
         _, _, h, w = img.shape
         img_up = img
--- a/kornia/feature/scale_space_detector.py
+++ b/kornia/feature/scale_space_detector.py
@@ -893,9 +893,10 @@ class MultiResolutionDetector(nn.Module):
         responses = torch.cat(all_responses, 1)
         lafs = torch.cat(all_lafs, 1)
         # The levels can produce fewer slots than requested — a level is capped at its own pixel
-        # count, and the per-level quotas round down, to zero for a small `num_features`. Pad up
-        # so the returned shape is always `num_features`, the same way `ScaleSpaceDetector.detect`
-        # does; the padding is the zero response and zero LAF used everywhere else here.
+        # count, so a deep enough pyramid level runs out of positions before it fills its quota.
+        # Pad up so the returned shape is always `num_features`, the same way
+        # `ScaleSpaceDetector.detect` does; the padding is the zero response and zero LAF used
+        # everywhere else here.
         if lafs.shape[1] < self.num_features:
             pad = self.num_features - lafs.shape[1]
             responses = F.pad(responses, (0, pad))
--- a/tests/feature/test_scale_space_detector.py
+++ b/tests/feature/test_scale_space_detector.py
@@ -853,6 +853,28 @@ class TestMultiResolutionDetector(BaseTester):
         self.assert_close(resps_one[0, 0], resps_two[0].max())
         self.assert_close(lafs_one[0, 0], lafs_two[0, int(resps_two[0].argmax())])
 
+    def test_small_request_spreads_across_scales(self, device, dtype):
+        # `detect` used to apportion `num_features` across pyramid levels by flooring each level's
+        # fractional share independently (`int(x) for x in shares`). The shares favor the finest
+        # level so heavily (0.508 .. 0.016 with the default config) that a small `num_features`
+        # truncated every other level's quota to zero: five well-separated, equally strong blobs and
+        # `num_features=3` returned three near-duplicate detections of the single strongest blob
+        # instead of covering several of them. Largest-remainder (Hamilton) apportionment hands the
+        # shortfall to the levels with the largest fractional remainder instead, so a small request
+        # still reaches more than one scale and covers more than one blob.
+        img = torch.zeros(1, 1, 96, 96, device=device, dtype=dtype)
+        centers = [(20, 20), (20, 70), (48, 48), (76, 20), (76, 70)]
+        for y, x in centers:
+            img[:, :, y - 2 : y + 3, x - 2 : x + 3] = 1.0
+
+        det = self._make_detector(num_features=3).to(device, dtype)
+        _, lafs = det.detect(img)
+        laf_centers = lafs[0, :, :, 2]
+        covered = sum(
+            1 for y, x in centers if bool(((laf_centers - laf_centers.new_tensor([x, y])).abs().amax(dim=1) < 4).any())
+        )
+        assert covered >= 2, f"expected the 3-feature request to reach at least 2 of the 5 blobs, got {covered}"
+
     def test_negative_score_threshold_is_rejected(self, device, dtype):
         # NMS writes an exact zero at every suppressed position, so a negative threshold would
         # admit all of them -- and collide with the zero response that marks an unfilled slot.


ducha-aiki:
## Review of #4156 — the apportionment is right; one rebase left

Setup: worktree at the PR head (`b93bdfcf`) with `kornia/` swapped to the merge-base `d7d6a2f6` for the fail-on-base and A/B passes. `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU/macOS; `kornia.__file__` asserted inside the worktree in every block.

### The test discriminates, and nothing else moves

```
head:  tests/feature/test_scale_space_detector.py                    65 passed
base:  kornia/ -> d7d6a2f6                                            1 failed, 64 passed
  FAILED …::test_small_request_spreads_across_scales[cpu-float32]
    AssertionError: expected the 3-feature request to reach at least 2 of the 5 blobs, got 1
```

#4098's own `num_features=1` regression test still passes at head, which is the property I most wanted to see given that this PR deletes that fallback branch.

### The quotas now sum to `num_features` at every size I tried

Recomputed the apportionment directly from the default configuration (`scale_factor_levels=1.414…`, `num_pyramid_levels=4`, `num_upscale_levels=1`, six levels):

```
  nf base quotas                 sum   head quotas                 sum
   1 [1, 0, 0, 0, 0, 0]            1   [1, 0, 0, 0, 0, 0]            1
   2 [1, 0, 0, 0, 0, 0]            1   [1, 1, 0, 0, 0, 0]            2
   3 [1, 0, 0, 0, 0, 0]            1   [2, 1, 0, 0, 0, 0]            3
   5 [2, 1, 0, 0, 0, 0]            3   [3, 1, 1, 0, 0, 0]            5
  10 [5, 2, 1, 0, 0, 0]            8   [5, 3, 1, 1, 0, 0]           10
  30 [15, 7, 3, 1, 0, 0]          26   [15, 8, 4, 2, 1, 0]          30
  63 [32, 15, 7, 3, 1, 0]         58   [32, 16, 8, 4, 2, 1]         63
 100 [50, 25, 12, 6, 3, 1]        97   [51, 25, 13, 6, 3, 2]       100
1000 [507, 253, 126, 63, 31, 15] 995   [508, 254, 127, 63, 32, 16] 1000
8000 [4063, 2031, …, 126]       7995   [4063, 2032, …, 127]       8000
```

Two things worth naming out loud, because the PR body only claims the small-request case:

- **This changes the apportionment at *every* `num_features`, not only small ones.** At the class default `num_features=8000` the detector used to ask its levels for 7995 candidates and now asks for 8000. That is the function honouring its own documented argument, so I read it as a bugfix rather than repair-window material — but it is a behaviour change for every caller and the CHANGELOG should say so in those terms rather than only "for a small `num_features`".
- **End-to-end the improvement is narrower than the quota table suggests.** Re-running the issue's own five-blob reproduction:

```
             base                              head
nf=  1:  1 detections, 1 of 5 blobs      1 detections, 1 of 5 blobs
nf=  3:  3 detections, 1 of 5 blobs      3 detections, 2 of 5 blobs   <- the fix
nf=  5:  5 detections, 3 of 5 blobs      5 detections, 3 of 5 blobs
nf= 10:  9 detections, 5 of 5 blobs      9 detections, 5 of 5 blobs
nf= 30:  9 detections, 5 of 5 blobs      9 detections, 5 of 5 blobs
```

`nf=5` still covers 3 of 5 even though its quotas change from `[2,1,0,0,0,0]` to `[3,1,1,0,0,0]` — the coarser levels re-detect blobs the finest level already found. Not a defect in this PR; just do not let the body imply the coverage curve moves everywhere.

### Degenerate inputs are unchanged

I suspected the negative branch — `by_remainder[:shortfall]` with a negative `shortfall` slices from the end and would hand a slot to all-but-one level. Executed rather than reasoned:

```
num_features= 0:  lafs (1, 0, 2, 3), 0 non-zero responses    (base and head identical)
num_features=-1:  RuntimeError: selected index k out of range (base and head identical)
```

`num_features=-1` raises before the apportionment is reachable on both revisions, so there is nothing to guard. Withdrawn.

### The one thing left

**Rebase.** The PR is `CONFLICTING`; `git merge-tree` says the conflict is **`CHANGELOG.md` only** — `kornia/feature/scale_space_detector.py` and the test file both auto-merge. So it is a one-file resolution, and per [[conflicting-pr-blocks-all-ci]] the green checks currently displayed were computed before `main` moved.

Nice piece of work — the largest-remainder framing generalises #4098's special case exactly as it should, and dropping the dedicated branch is the right call.

ducha-aiki:
## Maintainer note: rebased, plus benchmark evidence

I rebased this branch onto `main` for you (it had gone `CONFLICTING`) and then measured the change on a real detector, since `MultiResolutionDetector` is what `KeyNetDetector` inherits.

### Rebase

Rebased onto `0ba3414c`; head is now `3cfbaa60`. The only conflict was additive in `CHANGELOG.md`'s `### Bug fixes` section — your entry is kept at the top, per the repo's prepend convention. Nothing in your diff changed.

Two things to be aware of on your branch:

* I pushed one extra commit, `3cfbaa60`. The padding comment a few lines below your change still said the levels can under-fill because "the per-level quotas round down, to zero for a small `num_features`" — largest-remainder apportionment makes the quotas sum to exactly `num_features`, so that clause is now false. I rewrote only that clause; the block itself stays, because a level is still capped at its own pixel count.
* CI had one red leg, `tests (windows-latest, float32, 3.13, 2.9.1)`. It was `tests/models/test_sam_onnx.py::test_sam_encoder_to_onnx` hitting the pytest timeout inside `onnx.save(...)` on the Windows runner — unrelated to this diff. I re-ran the failed jobs and everything is green (37/37).

I also confirmed your new test discriminates: copied onto a `main` worktree, `test_small_request_spreads_across_scales` fails there with `covered=1` and passes on this branch. All 65 tests in `tests/feature/test_scale_space_detector.py` pass on the rebased head.

### 1. Registration is not worse — it is unchanged

`benchmarks/feature/scale_space_detector.py --method keynet` (`KeyNetHardNet`, upright, SMNN, `RANSAC("homography", inl_th=2.0, seed=3407)`), CPU, on the full Oxford-Affine set — bark, bikes, boat, graf, leuven, trees, ubc, wall — 8 sequences x 5 pairs = 40 pairs per setting. Base arm is a `main` worktree; each arm asserts `kornia.__file__` resolves inside its own tree before running, so the two arms really are different revisions.

| `num_features` | error base -> fix [px] | inliers base -> fix | identical pairs |
| --- | --- | --- | --- |
| 2048 (KeyNet default) | 8.47 -> 8.47 | 631.5 -> 631.5 | 40/40 |
| 512 | 22.26 -> 22.26 | 167.8 -> 167.8 | 40/40 |
| 128 | 41.17 -> 41.17 | 41.8 -> 41.8 | 40/40 |
| 32 | 31.09 -> 31.09 | 9.5 -> 9.5 | 40/40 |
| 16 | 281.56 -> 281.56 | 4.2 -> 4.2 | 40/40 |
| 8 | 1860.09 -> 1860.09 | 1.2 -> 1.2 | 40/40 |

240 pairs in total, and every individual per-pair error and inlier count matches exactly — not just the means. That is expected from the quota arithmetic: with KeyNet's default config the vector only shifts by +1 on a single level once `num_features >= 64` (`[1040, 520, 260, 130, 65, 32]` -> `[..., 65, 33]` at 2048), and each pyramid level already requests a cumulative sum of the quotas, so the one extra candidate is weak enough that the final global top-k discards it.

### 2. The detections do change at small requests, and strictly for the better

Registration cannot see the regime this actually fixes, because `num_features <= 5` is below what RANSAC needs. So I measured it directly: for each of the 8 images and each `num_features`, how many of the *true* top-k responses (reference taken from a `num_features=256` run) the detector actually returns.

| `num_features` | base | after this PR |
| --- | --- | --- |
| 1 | 8/8 | 8/8 |
| 2 | 14/16 | **15/16** |
| 3 | 18/24 | **24/24** |
| 5 | 35/40 | **38/40** |
| 8 | 59/64 | 59/64 |
| 16 | 122/128 | 122/128 |
| 32 | 252/256 | 252/256 |
| **total** | 508/536 (0.948) | **518/536 (0.966)** |

Strictly better in 9 of the 56 image/`num_features` cells, strictly worse in 0.

A concrete case, `bark` at `num_features=2`:

```
base:  resp=5241.04  scale=22.00  xy=(566.0, 478.0)
       resp=4444.15  scale=15.56  xy=(566.9, 478.8)   <- the same point, one scale down
after: resp=5241.04  scale=22.00  xy=(566.0, 478.0)
       resp=5177.11  scale=22.00  xy=(606.0, 465.0)   <- the actual second-strongest maximum
```

Base's second slot is a scale-space re-detection of the first keypoint and it misses the genuine runner-up `5177.11`, which base itself returns as soon as you ask for `num_features=5`. `ubc` at `num_features=3` is the same story: base returns `[17181, 16770, 13654]`, this PR returns `[17181, 16970, 16770]`, and the `16970` that base skips is again one it finds at `num_features=5`. That is exactly the truncation described in the PR body: at `[1, 0, 0, 0, 0, 0]` the finest level is asked for one candidate, so the second slot can only be filled from another scale.

So `detect` was not returning the top-k responses at small k, and it does now.

### 3. A separate wart, not this PR's job -- filed as #4170

The recall table above does not reach 100% even after this PR, and those residual misses are identical in both arms, so they are not yours. Filed separately as #4170.

In short: a level can contribute at most `quotas[0]` (up-level) or `sum(quotas[: i + 1 + num_upscale_levels])` (pyramid level `i`) candidates, which is well below `num_features` for the finer levels. When an image's strongest maxima concentrate on one level, the excess is never requested and the final global `topk` cannot rank it in. On `wall` at `num_features=32`, 20 of the true top-32 sit on the upscaled level, which is asked for `quotas[0] == 16` -- so 4 are unreachable, which is exactly the 28/32 measured. The same arithmetic reproduces with `BlobHessian` on a synthetic image with no pretrained weights (3/4, 6/8, 12/16, 18/24), identically on `main` and on this branch.

That is the quota *cap*, not the quota *rounding* this PR fixes, and undoing it means over-requesting per level and paying for more candidates through the concatenation -- a deliberate trade-off rather than a drive-by change. Hence a separate issue.

LGTM on the change itself, pending review.

ducha-aiki (approved):

Issue #4149, opened 2026-09-01
--isolate-half-precision: mixed multi-file CUDA runs error with "previous item was not torn down properly" on unrelated non-half tests


ducha-aiki:
## Describe the bug

When running a **mixed multi-file** CUDA test selection with `--isolate-half-precision`, some **non-half** tests error during setup with pytest's internal assertion:

```
ERROR tests/filters/test_bilateral.py::TestJointBilateralBlur::test_exception - AssertionError: previous item was not torn down properly
ERROR tests/geometry/test_grid.py::test_normalized_meshgrid_integer_dtype_promotes_uniformly[cuda-grid_dtype0-1-4] - AssertionError: previous item was not torn down properly
ERROR tests/geometry/test_linalg.py::TestTransformPoints::test_gradcheck[cuda] - AssertionError: previous item was not torn down properly
```

All three tests pass when run standalone, and single-file runs under the same flag are clean. The errored tests are not half-precision tests themselves — they are the victims of a bookkeeping gap in the isolation plugin.

## Root cause

`pytest_runtest_protocol` in the repo `conftest.py` hijacks the entire runtest protocol for isolated (CUDA float16/bfloat16) items: it runs the test in a subprocess and emits synthetic `TestReport`s, then returns `True` **without ever driving pytest's `SetupState`** (no `SetupState.setup(item)`, no `SetupState.teardown_exact(nextitem)`).

pytest tears fixtures down lazily: when a normal item finishes, `teardown_exact(nextitem)` keeps higher-scoped fixture stack entries alive based on `nextitem`'s collector chain. Sequence that breaks:

1. Normal item **A** finishes; its teardown keeps stack entries alive for `nextitem` **B** (an isolated half test, possibly in another file).
2. **B** is short-circuited by the custom protocol — the setup state is never advanced or unwound for it.
3. The next normal item **C** calls `SetupState.setup()`, finds stale stack entries that belong to neither its chain nor a properly-unwound state, and fails the `previous item was not torn down properly` assertion — reported as a setup ERROR on **C**.

This only manifests when isolated and non-isolated items interleave across files, which is why single-file runs and dtype-filtered runs look clean, while a full `pixi run -e cuda test-cuda-half` style run over multiple modules can hit it.

## Reproduction

On `main` (or any branch; the plugin is unchanged), with a CUDA machine:

```bash
KORNIA_TEST_DEVICE=cuda KORNIA_TEST_DTYPE=float16,bfloat16 \
[...]

PR #4157, opened 2026-09-01
fix: isolate half-precision parametrized tests

### LINKED ISSUE #4126: Tests: --isolate-half-precision misses half-precision tests that do not use the global dtype fixture
### Describe the bug

`--isolate-half-precision` recognizes a half-precision test only through the **global `dtype` fixture**. A test that carries its dtype in its own parameter instead — `@pytest.mark.parametrize("half_dtype", [torch.float16, torch.bfloat16])` and friends — is invisible to both halves of the machinery, so it runs `float16`/`bfloat16` CUDA kernels in the shared pytest process with no isolation and no skip.

Two places in `conftest.py` make the decision, and both read `dtype_name`:

- `_is_subprocess_isolated_test` (`conftest.py:451`) returns `False` unless `callspec.params["dtype_name"] in ("float16", "bfloat16")` **and** `params["device_name"] == "cuda"`. A `half_dtype`-parametrized test on a float32 job has `dtype_name == "float32"`, so `pytest_runtest_protocol` never moves it into a subprocess.
- the autouse `skip_half_precision_on_cuda` fixture (`conftest.py:567`) returns early when `"dtype" not in request.fixturenames`. Such a test does not request `dtype`, so the default-skip does not apply either.

The result is the exact situation the isolation exists to prevent: an asynchronous device-side assert from a half-precision kernel corrupts the CUDA context for every later test in the process, and the failure surfaces somewhere unrelated — including in a nominally float32 job, since these ids collect under `--dtype=float32`.

### Reproduction steps

Collection alone shows the shape of the problem — the id has no `dtype_name`, so neither guard fires:

```console
$ pytest tests/feature/test_local_features_orientation.py -k flat_patch --device=cuda --dtype=float32 --collect-only -q
tests/feature/test_local_features_orientation.py::TestOrientationHalfPrecisionIsFinite::test_flat_patch[cuda-half_dtype0]
tests/feature/test_local_features_orientation.py::TestOrientationHalfPrecisionIsFinite::test_flat_patch[cuda-half_dtype1]
```

Scanning the whole suite for collected ids whose device is `cuda`, whose `dtype_name` is *not* half, and one of whose other parameters *is* a half `torch.dtype` gives **49 test ids across 8 files** (kornia at `88be45b5`, torch 2.9.1):

```
  2  tests/augmentation/container/test_augmentation_sequential.py  (image_dtype)
  4  tests/feature/test_hardnet.py                                 (desc_dtype)
  6  tests/feature/test_local_features_orientation.py              (half_dtype, ori_dtype)
  2  tests/feature/test_matching.py                                (desc_dtype)
  4  tests/feature/test_scale_space_detector.py                    (half_dtype)
 18  tests/feature/test_siftdesc.py                                (desc_dtype)
  8  tests/geometry/test_conversions.py                            (default_dtype, half_dtype, input_dtype)
  5  tests/geometry/test_linalg.py                                 (points_dtype, trans_dtype)
```

(A ninth file, `tests/feature/test_laf.py`, had 4 more; they were made CPU-only in #4120 as a local workaround, which is why they are not in this list.)

### Expected behavior

`--isolate-half-precision` should move these into subprocesses, and without the flag they should be skipped on CUDA, exactly as the `dtype`-fixture tests are.

The fix is in the two predicates rather than in each test: treat *any* parameter whose value is `torch.float16`/`torch.bfloat16` as making the test half-precision, not only `dtype_name`. In `_is_subprocess_isolated_test` that means scanning `callspec.params.values()` for a half `torch.dtype`; in `skip_half_precision_on_cuda` it means looking at `request.node.callspec.params` as well as the `dtype` fixture. The subprocess relaunch also needs care: it currently forwards `--device`/`--dtype` so the child reproduces the parent's parametrization, and a `half_dtype` id is reproduced by the plain node id, so the forwarded `--dtype` must stay the parent's (`float32`), not the parameter's.

Until then, a half-precision test that does not read the global `dtype` fixture should skip every non-CPU device.

### Environment

- kornia `88be45b5`, torch 2.9.1, Python 3.11, macOS (collection-only; the CUDA execution path is inferred from `conftest.py`, since kornia CI has no CUDA runner)

Found while addressing review feedback on #4120.

cris-pat-37:

Fixes an issue where `--isolate-half-precision` did not detect tests using locally parametrized `torch.float16` or `torch.bfloat16` values.

Previously, isolation only checked the global `dtype_name` parameter. This meant a CUDA test could use a local half-precision parameter while the parent test job was configured for `float32`, causing the test to run in the shared pytest process.

This change detects half-precision dtype values across the test parameters and updates `skip_half_precision_on_cuda` to handle locally parametrized dtypes.

Fixes #4126

## What did you check?

- `pixi run test-module tests/test_conftest.py`
  - 10 tests passed
- `pixi run lint`
  - Ruff check passed
  - Ruff format passed

Added regression and integration tests covering:
- locally parametrized `torch.float16`
- locally parametrized `torch.bfloat16`
- custom parameter names
- existing global `dtype_name` behavior
- CPU and `float32` cases

## How did AI help?

AI was used to help investigate the existing pytest/conftest behavior and review the proposed changes. I used it to QA the implementation, check the test coverage, and identify potential issues before submitting the PR. I then ran the regression tests and lint checks locally to verify the changes.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -464,11 +464,11 @@ def _is_subprocess_isolated_test(item) -> bool:
     if callspec is None:
         return False
     params = callspec.params
-    if params.get("dtype_name") not in ("float16", "bfloat16"):
-        return False
     if params.get("device_name") != "cuda":
         return False
-    return True
+    if params.get("dtype_name") in ("float16", "bfloat16"):
+        return True
+    return any(v in (torch.float16, torch.bfloat16) for v in params.values())
 
 
 def pytest_runtest_protocol(item, nextitem):
@@ -589,11 +589,20 @@ def skip_half_precision_on_cuda(request):
     if os.environ.get("KORNIA_TEST_IN_SUBPROCESS"):
         return
 
-    if "dtype" not in request.fixturenames:
+    half_dtype = None
+    if "dtype" in request.fixturenames:
+        dtype = request.getfixturevalue("dtype")
+        if dtype in (torch.bfloat16, torch.float16):
+            half_dtype = dtype
+    elif hasattr(request.node, "callspec"):
+        half_dtype = next(
+            (v for v in request.node.callspec.params.values() if v in (torch.bfloat16, torch.float16)),
+            None,
+        )
+
+    if half_dtype is None:
         return
-    dtype = request.getfixturevalue("dtype")
-    if dtype not in (torch.bfloat16, torch.float16):
-        return
+
     if "device" not in request.fixturenames:
         return
 
@@ -606,7 +615,7 @@ def skip_half_precision_on_cuda(request):
         return
 
     if not request.config.getoption("--isolate-half-precision"):
-        dtype_name = "bfloat16" if dtype == torch.bfloat16 else "float16"
+        dtype_name = "bfloat16" if half_dtype == torch.bfloat16 else "float16"
         pytest.skip(
             f"{dtype_name} on CUDA: skipped by default to prevent device-side assert contamination. "
             "Run with --isolate-half-precision to execute in isolated subprocesses."
--- a/tests/test_conftest.py
+++ b/tests/test_conftest.py
@@ -1,0 +1,193 @@
+# LICENSE HEADER MANAGED BY add-license-header
+#
+# Copyright 2018 Kornia Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from __future__ import annotations
+
+import types
+
+import pytest
+import torch
+
+from conftest import _is_subprocess_isolated_test, skip_half_precision_on_cuda
+
+pytest_plugins = ["pytester"]
+
+_skip_fixture_fn = getattr(skip_half_precision_on_cuda, "__wrapped__", skip_half_precision_on_cuda)
+
+
+def _make_mock_item(params: dict | None, isolate: bool = True):
+    config = types.SimpleNamespace(
+        getoption=lambda opt, default=False: isolate if opt == "--isolate-half-precision" else default
+    )
+    callspec = types.SimpleNamespace(params=params) if params is not None else None
+    return types.SimpleNamespace(config=config, callspec=callspec)
+
+
+def _make_mock_request(
+    fixturenames: list[str],
+    fixture_values: dict[str, object],
+    params: dict | None = None,
+    isolate: bool = False,
+):
+    node = _make_mock_item(params, isolate=isolate)
+    config = node.config
+
+    def getfixturevalue(name: str):
+        if name in fixture_values:
+            return fixture_values[name]
+        raise pytest.FixtureLookupError(name, node)
+
+    return types.SimpleNamespace(
+        fixturenames=fixturenames,
+        getfixturevalue=getfixturevalue,
+        config=config,
+        node=node,
+    )
+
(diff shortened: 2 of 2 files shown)


ducha-aiki:
## Review of #4157 — APPROVE

Setup: worktree at the PR head (`39bf8397`), `conftest.py` swapped to the merge-base `d7d6a2f6` for the fail-on-base pass. `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU/macOS.

### The tests discriminate

```
head:  tests/test_conftest.py                          10 passed
base:  conftest.py -> d7d6a2f6                          4 failed, 6 passed
         TestIsSubprocessIsolatedTest::test_reproduce_issue_4126_local_half_dtype
         TestIsSubprocessIsolatedTest::test_local_custom_param_names[target_dtype0]
         TestIsSubprocessIsolatedTest::test_local_custom_param_names[target_dtype1]
         TestSkipHalfPrecisionOnCuda::test_local_half_dtype_is_skipped_without_isolation_flag
```

### The gap it closes, measured against the real suite

I collected all of `tests/` (12660 items, 11910 with a callspec), substituted `device_name="cuda"` into every item that actually carries one (11391), and evaluated the old and new predicates:

```
isolated under the OLD predicate:  0
isolated under the NEW predicate: 53
```

The 53, by file:

```
tests/feature/test_siftdesc.py                             18
tests/geometry/test_conversions.py                          8
tests/feature/test_local_features_orientation.py            6
tests/geometry/test_linalg.py                               5
tests/feature/test_hardnet.py                               4
tests/feature/test_laf.py                                   4
tests/feature/test_scale_space_detector.py                  4
tests/augmentation/container/test_augmentation_sequential.py 2
tests/feature/test_matching.py                              2
```

Two things I specifically checked because they would have made this silently useless:

1. **`any(v in (torch.float16, torch.bfloat16) for v in params.values())` never raises.** `x in tuple` evaluates `x == e`, which throws for a tensor- or array-valued parameter. Over all 11910 real callspecs: **0 crashes**.
2. **Every newly-isolated nodeid is reproducible in the subprocess.** `pytest_runtest_protocol` re-invokes `sys.executable -m pytest <nodeid> --device=… --dtype=…`, and a nodeid the child cannot collect exits 5, which the parent reports as a **skip** — coverage would silently vanish. Re-collecting all 53 nodeids explicitly:

```
re-collect rc: 0   collected: 53
nodeids that would NOT be found by the subprocess: 0
```

That matters here because these tests have no `dtype_name` in their id, so the child is launched with `--dtype=float32` while the id says e.g. `[cuda-half_dtype0]`. It resolves.

### Non-blocking

1. **The two guards use different predicates.** `_is_subprocess_isolated_test` scans *all* params; `skip_half_precision_on_cuda` uses `if "dtype" in fixturenames … elif callspec`. A test that requests both the global `dtype` fixture (at `float32`) and a local half param would take the `if` branch, find `float32`, and return — i.e. the exact #4126 shape, unfixed. No such test exists today (I checked every locally-parametrized half test in the suite: all take `device` + a local dtype, none take `dtype`), so this is latent, not a bug. Making it an `if` that also scans the callspec when the `dtype` fixture is not half would make the two guards agree by construction.
2. **`tests/geometry/test_grid.py:266::test_pixel_meshgrid_half_dtype_matches_compile` stays uncovered** — it locally parametrizes `grid_dtype` over half but takes no `device` fixture, so neither guard can see it. It never touches CUDA, so it is not a hazard; just noting it so the issue is not read as fully swept.
3. Landing this means 53 node ids are **skipped by default** on a CUDA run and only execute under `--isolate-half-precision`, which is the intended trade — worth knowing that #4149 (mixed multi-file CUDA runs under that flag) is still open.

### No collision with the in-flight CI work

#4154 and #4160 also touch `conftest.py`, and I checked the regions: they add to the imports, `pytest_collection_modifyitems` and `pytest_addoption`; this PR touches `_is_subprocess_isolated_test` (~L464) and `skip_half_precision_on_cuda` (~L589-615). Disjoint — this can land in any order relative to them.

Thanks — the nodeid-reproducibility property is the part I was most worried about and it holds.

ducha-aiki (approved):

Issue #3161, opened 2025-04-23
Missing parenthesis in distance_transform implementation


jonaum:
### Describe the bug

In line 85 of _kornia.contrib.distance_transform.py_, there seems to be a parenthesis missing around _kernel_size//2_ which leads to incorrect results.

### Reproduction steps with incorrect output

```bash
img = torch.zeros((1,1,8,4))
img[0,0,1,:] = 1
kornia.contrib.distance_transform(img, kernel_size=3, h=0.01)

tensor([[[[0.9998, 0.9998, 0.9998, 0.9998],
                [0.0000, 0.0000, 0.0000, 0.0000],
                [0.9998, 0.9998, 0.9998, 0.9998],
                [1.9998, 1.9998, 1.9998, 1.9998],
                [3.9998, 3.9998, 3.9998, 3.9998],
                [4.9998, 4.9998, 4.9998, 4.9998],
                [6.9998, 6.9998, 6.9998, 6.9998],
                [7.9998, 7.9998, 7.9998, 7.9998]]]])
```

### Expected behavior

From row 4 to 5 and 6 to 7, the resulting distances are increased by 2 instead of 1.
Adding the missing parenthesis solves the problem.

### Environment

```shell
- PyTorch Version (e.g., 1.0): 2.3.1
- OS (e.g., Linux): Windows 11
- Python version: 3.12
- kornia version: 0.7.4
```

edgarriba:
can you submit a PR with the fix and updated test ?

stephenm7777:
I pushed something earlier, would be grateful if someone took a look!

PR #4158, opened 2026-09-01
fix(contrib): guard distance_transform against exp(-dist/h) kernel underflow

### LINKED ISSUE #4152: [Bug]: distance_transform silently returns wrong distances (or all zeros) when its exp(-dist/h) kernel underflows — reachable at the default h=0.35
## What happened?

`kornia.contrib.distance_transform` builds its kernel as `exp(-dist / h)` and then inverts it with `-h * log(...)`. When the far taps of that kernel underflow the working dtype they become exactly `0`, the convolution silently loses them, and the function returns a **wrong distance field with no error and no warning** — either plausible-looking but too-large distances, or an all-zero field.

This is reachable from the documented defaults: **`h=0.35` (the default) with `kernel_size >= 15` in float16 already returns wrong answers.**

```text
distance_transform(img, kernel_size=ks)   # default h=0.35, img is 16x4, one boundary row
                                          # correct peak distance is ~13.5

 ks | cpu/float64 | cpu/float32 | mps/float32 | cpu/float16 | cpu/bfloat16
  3 |      13.833 |      13.833 |      13.833 |      13.836 |       13.812
  7 |      13.697 |      13.697 |      13.697 |      13.695 |       13.688
 11 |      13.587 |      13.587 |      13.587 |      13.594 |       13.625
 15 |      13.515 |      13.515 |      13.515 |  >> 15.695 |       13.500
 21 |      13.585 |      13.585 |      13.585 |  >> 15.828 |       13.625
```

With a smaller `h` it gets worse, and the three failure modes are all silent. Correct answer on the last row is `[1, 0, 1, 2, 3, 4, 5, 6]`:

```text
distance_transform(img, kernel_size=7, h=0.02), column 0

cpu float64   [1.0, 0.0, 1.0, 2.0,    3.0, 4.0,    5.0, 6.0   ]   <- correct
cpu float32   [1.0, 0.0, 1.0, 1.9997, 4.0, 4.9997, 7.0, 7.9997]   <- wrong, monotone, plausible
mps float32   [1.0, 0.0, 1.0, 4.0,    7.0, 0.0,    0.0, 0.0   ]   <- wrong, then truncated to 0
cpu bfloat16  [1.0, 0.0, 1.0, 4.0,    7.0, 0.0,    0.0, 0.0   ]   <- same
```

and further down it collapses to an all-zero field:

```text
max |output| (true peak is 6.0), kernel_size=3

     h | cpu/float32 | mps/float32 | cpu/float16 | cpu/bfloat16 | cpu/float64
  0.10 |       5.997 |       5.997 |       5.996 |        6.000 |       5.997
  0.05 |       6.000 |       6.000 |  ZERO 0.000 |        6.000 |       6.000
  0.01 |       6.000 |  ZERO 0.000 |  ZERO 0.000 |   ZERO 0.000 |       6.000
 0.005 |  ZERO 0.000 |  ZERO 0.000 |  ZERO 0.000 |   ZERO 0.000 |       6.000
```

The 5D/volume path (`_distance_transform_3d_impl`) and the `DistanceTransform` module behave identically. `h <= 0` is already rejected (`h must be a positive float`), but nothing catches this regime — `h=1e-30` is accepted and returns zeros.

### Mechanism

```text
kernel taps surviving exp(-dist/h)

cpu float16   ks=15 h=0.35 : 113/225 taps non-zero; every tap at distance >= 6.08 is 0
cpu float32   ks=3  h=0.01 :   5/9   taps non-zero (the axial ones survive as subnormals)
mps float32   ks=3  h=0.01 :   1/9   taps non-zero (MPS flushes subnormals to zero)
cpu bfloat16  ks=3  h=0.01 :   1/9   taps non-zero (bfloat16 has no subnormal at 3.8e-44)
```

Once only the centre tap survives, `filter2d(boundary, kernel) == boundary`, so `cdt = -h*log(boundary)` is `0` where the boundary is set and `+inf` elsewhere; `nan_to_num(posinf=0.0)` turns that into `0`, `mask = cdt > 0` is empty, and every iteration is a no-op. That is the all-zero case. With *partial* underflow the sub-kernel refinement disappears but the `offset = i * (kernel_size // 2)` accumulation continues, which is where `4.0` and `7.0` (i.e. `3i + 1`) come from in the table above.

This is why CPU float32 and MPS float32 disagree: it is not an MPS defect. MPS flushes subnormals to zero, bfloat16 has no subnormal that small, and float16 runs out of range far earlier — CPU float32 just happens to survive one step longer.

## Minimal reproduction

```python
import torch
import kornia

# 1. wrong answer at the DOCUMENTED DEFAULT h, float16, large kernel
img = torch.zeros(1, 1, 16, 4)
img[0, 0, 1, :] = 1.0
ref = kornia.contrib.distance_transform(img.double(), kernel_size=15).max()
f16 = kornia.contrib.distance_transform(img.half(), kernel_size=15).max()
print(ref.item(), f16.item())      # 13.5146... vs 15.6953...  -> silently wrong

# 2. silent all-zero field
img = torch.zeros(1, 1, 8, 4)
img[0, 0, 1, :] = 1.0
print(kornia.contrib.distance_transform(img, kernel_size=3, h=0.01)[0, 0, :, 0])
# cpu float32 : tensor([0.9998, 0.0000, 0.9998, 1.9998, 2.9998, 3.9998, 4.9998, 5.9998])
print(kornia.contrib.distance_transform(img.to("mps"), kernel_size=3, h=0.01)[0, 0, :, 0])
# mps float32 : tensor([0., 0., 0., 0., 0., 0., 0., 0.], device='mps:0')
print(kornia.contrib.distance_transform(img.bfloat16(), kernel_size=3, h=0.01)[0, 0, :, 0])
# cpu bfloat16: tensor([0., 0., 0., 0., 0., 0., 0., 0.], dtype=torch.bfloat16)

# 3. no validation catches it
print(kornia.contrib.distance_transform(img, kernel_size=3, h=1e-30).abs().max())  # tensor(0.)
```

## Environment

- Kornia version: 0.9.0rc1 (`11fd2721`, `main` on 2026-09-01)
- PyTorch version: 2.9.1
- Python version: 3.11.14
- OS / device: macOS 26.5.1 arm64, MPS (the float16 / bfloat16 / small-`h` cases all reproduce on CPU alone)

```text
```

### Possible directions

Not prescribing one — whoever takes this should pick:

1. **Fail loudly instead of silently.** The smallest kernel tap is known before any work happens: `exp(-d_max / h)` with `d_max = sqrt(2) * (kernel_size // 2)` for the 2D path (`sqrt(3) * ...` for 3D). Rejecting or warning when that underflows `torch.finfo(dtype)` turns a wrong answer into a clear message, and is a few lines next to the existing `h > 0` check. This is the minimum bar.
2. **Build the kernel and run the cascade in float32** (or the input dtype promoted to at least float32), casting back at the end. That alone fixes the float16, bfloat16 and MPS rows in the tables above within the `h` range CPU float32 handles today. It does not fix small `h` in float32.
3. **Reformulate in the log domain.** `-h * log(sum_j b_j * exp(-d_j/h))` is a softmin; computing it as a `logsumexp` over the neighbourhood removes the dynamic-range problem at its source. Biggest change, and it has to stay `torch.compile`- and ONNX-friendly like the current cascade.

Whatever is chosen, the docstring should say what `h` may safely be — right now it only says *"value that influence the approximation of the min function"*, with no hint that a small value silently destroys the result.

### Test coverage this needs

- A dtype-parametrized regression asserting the float16/bfloat16 result matches a float64 reference at the default `h` for `kernel_size >= 15`.
- A pin on the small-`h` regime — either the fixed behaviour or an explicit `xfail`/raise — so the underflow cannot silently return.

### Related

- Found while reviewing #4147, which moves `tests/contrib/test_conv_distance_transformer.py::test_offset_parenthesis_fix` from `h=0.01` to `h=0.1` to dodge this. That reparametrization is fine on its own (I verified the pin still discriminates the #3161 offset bug), but it leaves the underflow regime uncovered by any test, which is why it wants its own issue.
- Same "half precision silently underflows a kornia constant" family as #3966 and #4123.
- No internal kornia caller uses `distance_transform`, so the blast radius is the public API and `DistanceTransform`.

mohansree14:
## What

`distance_transform` builds its cascade kernel as `exp(-dist/h)` and inverts it with `-h*log(...)`. When the farthest kernel taps underflow the working dtype to exactly zero, the convolution silently drops them and the function returns a wrong distance field with **no error and no warning** — either a plausible-looking but wrong result, or a silent all-zero field. This is reachable from the *documented default* `h=0.35` at `kernel_size >= 15` in `float16`.

```python
img = torch.zeros(1, 1, 16, 4)
img[0, 0, 1, :] = 1.0
ref = kornia.contrib.distance_transform(img.double(), kernel_size=15).max()   # 13.515 (correct)
f16 = kornia.contrib.distance_transform(img.half(), kernel_size=15).max()     # 15.695 (before fix, silently wrong)
```

## Fix

Two changes, per the directions the issue laid out:

1. **`float16`/`bfloat16` inputs now run the cascade in `float32`** and cast the result back — the same upcast idiom used elsewhere in kornia for half-precision numerical stability (e.g. `PatchAffineShapeEstimator`, `extract_patches_from_pyramid`). This alone fixes the documented-default case: `float16` at `h=0.35, kernel_size=15` now returns `13.5156` (diff `4.7e-4` from the float64 reference) instead of `15.695`.

2. **Fail loudly for the residual underflow regime the upcast doesn't reach.** Before any cascade work runs, the farthest kernel tap (`exp(-d_max/h)`, `d_max = sqrt(dims) * (kernel_size // 2)`) is checked against `torch.finfo(compute_dtype).tiny`; if it underflows, `distance_transform` now raises instead of guessing. This catches e.g. `h=0.01` at `kernel_size=3` in `float32`, which used to return a monotone-looking-but-wrong result.

   The check compares against the smallest *normal* value rather than the device-dependent smallest *subnormal*, deliberately: MPS flushes subnormals to zero while CPU does not, so the same `h`/dtype combination used to silently disagree across devices — part of the original bug the issue called out. Using the conservative (normal-only) threshold uniformly makes behavior consistent across backends instead of "works by luck on CPU, silently wrong on MPS."

## An existing test was pinning the bug

`test_offset_parenthesis_fix` used `h=0.01, kernel_size=3` to pin an unrelated, already-fixed offset-accumulation bug (#3161) — but its "expected" values were themselves the wrong output of *this* bug (the `0.9998`/`1.9998`/... pattern matches the issue's own "wrong, monotone, plausible" table row exactly). Moved to `h=0.1`, with new expected values generated from a `float64` reference, which still discriminates the offset-accumulation bug (strictly increasing by 1 per row) while staying clear of the underflow guard. (Noted in the issue: an unmerged PR #4147 independently made the same `h=0.01 → 0.1` move to dodge this; I verified the new values myself rather than depending on that PR landing first.)

## Testing

- `test_half_precision_matches_float64_reference`: `float16`/`bfloat16`/`float32` results at `h=0.35, kernel_size=15` now track a `float64` reference to ~float32 precision. Fails on pre-fix code for `float16` (`13.52` expected vs `15.70` got); passes for `bfloat16`/`float32`, matching the issue's own table (those weren't broken).
- `test_h_too_small_raises`: the degenerate `h=1e-30` now raises `BaseError` at every dtype, including `float64`.
- `test_h_too_small_raises_float32_regression`: pins the `float32`-specific `h=0.01, kernel_size=3` regression explicitly (independent of the dtype fixture, since `float64`'s much smaller `finfo.tiny` doesn't underflow at this `h`).
- All 3 new/changed assertions verified to fail against the pre-fix code and pass against the fix.
- Full `tests/contrib/test_conv_distance_transformer.py` passes across `float16`/`bfloat16`/`float32`/`float64` (66 tests).
- Verified `fullgraph=True` compile safety with `torch.compile(..., backend="eager")` (no graph break) — I don't have a working C++ compiler locally to exercise the `inductor` backend directly.

## Not yet verified

- No `pixi` or MPS/CUDA hardware in my local environment — verified with a throwaway `uv` venv (torch 2.14 CPU) instead. The MPS-specific claims (subnormal flush-to-zero) are drawn from the issue's own reproduction table, not independently re-verified by me on MPS hardware.

Fixes #4152


ducha-aiki:
## Review of #4158 — CHANGES REQUESTED

Setup: worktree at the PR head (`cb99d734`) with `kornia/` swapped to the merge-base `d7d6a2f6` for every A/B and fail-on-base pass. `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU/macOS; `kornia.__file__` asserted inside the worktree in every block.

Thank you for this — the diagnosis is right, the upcast half is a genuine fix, and the tests discriminate:

```
head:  tests/contrib/test_conv_distance_transformer.py, f16+bf16+f32+f64   66 passed
base:  kornia/ -> d7d6a2f6                                                  7 failed, 59 passed
```

The half-precision upcast is the valuable half and I have no objection to it. My problem is with the guard.

---

## 1. The `d_max` criterion rejects calls that are correct today

`_check_h_range` tests the **corner** tap, `exp(-sqrt(dims) * (kernel_size // 2) / h)`. That is the tap for a point `sqrt(2)·k_half` pixels away, whose contribution to a soft-min is negligible by construction — so losing it is not what breaks the function. Sweeping `h` on a 64×64 image with a single horizontal line, against the **analytic** Euclidean distance transform (`|row − 32|`), in `float32` on `main`:

```
  ks       h  mean|err|   #4158   axis-reach   nearest-tap
   3    0.35     0.1646      ok           ok            ok
   3     0.1     0.0031      ok           ok            ok
   3    0.05     0.0000      ok           ok            ok
   3    0.02     0.0000      ok           ok            ok
   3  0.0115     0.0000   RAISE           ok            ok      <-- exact today
   3    0.01     0.0002   RAISE        RAISE         RAISE      <-- 2e-4 today
   3   0.005    16.0000   RAISE        RAISE         RAISE
  15     0.2     0.1448      ok           ok            ok
  15     0.1     0.0400   RAISE           ok            ok      <-- 0.04 today
  15    0.05     5.2467   RAISE        RAISE            ok
  15    0.02    23.4999   RAISE        RAISE            ok
  63    0.35     0.5695   RAISE        RAISE            ok
  63     0.2     4.0326   RAISE        RAISE            ok
```

`axis-reach` = `exp(-k_half/h) >= tiny`; `nearest-tap` = `exp(-1/h) >= tiny`.

Three rows where the PR raises `BaseError` on a call that is accurate today: `ks=3, h=0.0115` (**exactly** right, mean error `0.0000`), `ks=3, h=0.01` (`2e-4`), `ks=15, h=0.1` (`0.04`).

And the documented default is affected too. At `h=0.35` in `float32` the guard first fires at **`kernel_size=45`** in 2D and **37** in 3D:

```
ks=43  allowed  mean|err| 0.4897
ks=45  RAISES   mean|err| 0.4930   <- 0.7% worse than the last allowed size
ks=51  RAISES   mean|err| 0.5078
ks=63  RAISES   mean|err| 0.5695
```

There is no cliff at 45. The accuracy curve is smooth and the guard cuts across it, so `distance_transform(x, kernel_size=45)` — a plain call at the documented default `h` — goes from returning a usable field to raising.

**Ask:** use the axis-aligned reach `k_half` rather than the corner `sqrt(dims) * k_half`. On the table above it agrees with your criterion on **every** genuinely-broken row and removes two of the three false positives. It is a one-token change: drop the `math.sqrt(dims) *`.

## 2. `test_h_too_small_raises_float32_regression`'s comment is factually wrong

It says `h=0.01` at `kernel_size=3` in `float32` "used to return a plausible-looking but wrong monotone result". It does not. On `main` that configuration returns `mean|err| = 2e-4` against the analytic reference — the values `0.9998, 1.9998, 2.9998, …` that `test_offset_parenthesis_fix` used to pin are `1, 2, 3, …` to four decimals, not the wrong output of a bug. CPU `float32` carries it through the subnormal range.

The real reason to reject it is the one you already give in `_check_h_range`: **MPS flushes subnormals to zero and CPU does not**, so the same call silently disagrees across backends. That is a good reason. Please state it in the test as the reason, instead of the accuracy claim — as written the comment will mislead the next reader into thinking a correct result was wrong.

## 3. Rebase — #4147 landed on the same test, with a strict pin

#4147 merged as `dfad8156` after you opened this. It touched `tests/contrib/test_conv_distance_transformer.py` in three ways that collide:

- It **already** moved `test_offset_parenthesis_fix` to `h=0.1` (expected `0.9969…`; your `0.996872` is the same number at more digits), and added `supports_replicate_padding` / `supports_conv2d` skips plus a `5e-3` bfloat16 tolerance. Please keep those — a naive "take mine" resolution drops the backend guards.
- It added a **strict** `xfail` pin, `test_convention_small_h_keeps_nonzero_distances_4152`, expecting `AssertionError`. Your fix makes that raise `BaseError` instead, so it fails either way. It must be deleted in the same push.
- `git merge-tree` reports conflicts in `CHANGELOG.md` **and** that test file.

Then please run the AGENTS.md gate for the issue you are closing:

```bash
grep -rnE "#4152|issues/4152" kornia/ tests/
```

A surviving hit in `kornia/` means the change is incomplete; hits in `tests/` are the pins to re-check.

---

Everything else I checked holds: the upcast idiom matches the one already used in this tree, the `float16` default-`h` case at `kernel_size=15` is genuinely fixed, the guard is evaluated at `compute_dtype` (so half inputs are checked at `float32` thresholds, which is right), and the docstring additions are accurate. Fix the threshold, the one comment, and the rebase and I will merge it.

mohansree14:
Thanks for the detailed sweep — all three addressed:

**1. Corner-tap → axis-aligned tap.** Switched \`_check_h_range\` from \`exp(-sqrt(dims)*k_half/h)\` to \`exp(-k_half/h)\`. Verified against every row of your table (\`kernel_size\` 3/15/45/63, \`h\` from 0.35 down to 0.005) — matches expected raise/no-raise on all of them, including the two false positives you flagged (\`kernel_size=3, h=0.0115\` and \`kernel_size=45\` at the default \`h=0.35\` both now pass through cleanly).

**2. Misleading comment.** Fixed in three places (the test comment, \`test_offset_parenthesis_fix\`'s header comment which made the same claim, and the CHANGELOG entry): \`h=0.01\` at \`kernel_size=3\` in float32 is not itself numerically wrong — it tracks the analytic reference to ~2e-4 on CPU. The reason it's rejected is the MPS/CPU subnormal-flush disagreement you identified, not accuracy, and the comments now say that.

**3. Rebase around #4147.** Already done in a prior push (before this review landed, so it was reviewed against the pre-rebase commit) — kept your skip guards and bfloat16 tolerance in \`test_offset_parenthesis_fix\`, dropped the now-superseded strict \`xfail\` since this fix makes that case raise \`BaseError\` instead of the pinned \`AssertionError\`, and \`test_h_too_small_raises\`/\`test_h_too_small_raises_float32_regression\` cover the fixed behavior instead. \`grep -rnE "#4152|issues/4152" kornia/ tests/\` — no hits in \`kornia/\`, all \`tests/\` hits are pins naming the issue.

Full \`tests/contrib/test_conv_distance_transformer.py\` passes across float16/bfloat16/float32/float64 (66 tests) with both changes.

ducha-aiki:
## Delta re-review of #4158 at `a6e816fc` — APPROVE, on a CHANGELOG-only rebase

Setup: worktree at the PR head, `kornia/` swapped to the merge-base `e0969438` for every fail-on-base and A/B pass. `.venv` python 3.11.14 / torch 2.9.1, CPU and MPS (MPS run with `PYTORCH_ENABLE_MPS_FALLBACK` unset).

All three asks are addressed, and the first one is addressed **better than I asked for**. I had asked for the axis-aligned tap because it removes 2 of the 3 false positives the corner tap produced. Re-measuring on both backends shows the new threshold is not merely less conservative — it is exactly the boundary at which CPU and MPS stop agreeing, so the remaining rejections are not false positives at all.

### 1. The guard now has a real cliff, and it is a device-agreement cliff

Base (`e0969438`), 64×8 image with one row of ones — analytic answer is `|i − 1|`, so `max|err| = 62.0` means the result collapsed to all zeros. `R` marks what the PR head raises at float32 working precision:

| h | ks=31 | ks=45 | ks=61 | ks=63 | ks=65 |
|---|---|---|---|---|---|
| **0.35 CPU** | 0.6232 | 0.6929 | 0.7486 | 0.7545 `R` | 0.7556 `R` |
| **0.35 MPS** | 0.6232 | 0.6929 | 0.7458 | **1.7918** `R` | **62.0000** `R` |
| **0.30 CPU** | 0.5042 | 0.5641 | 0.6120 `R` | 0.6176 `R` | 0.8565 `R` |
| **0.30 MPS** | 0.5042 | 0.5641 | **7.8565** `R` | **9.8565** `R` | **62.0000** `R` |
| **0.25 CPU** | 0.3965 | 0.4439 `R` | 9.9104 `R` | 11.9104 `R` | 62.0000 `R` |
| **0.25 MPS** | 0.3965 | **1.9104** `R` | 17.9104 `R` | 19.9104 `R` | 62.0000 `R` |
| **0.15 CPU** | 0.2004 `R` | 62.0000 `R` | 62.0000 `R` | 62.0000 `R` | 62.0000 `R` |
| **0.15 MPS** | **7.9818** `R` | 62.0000 `R` | 62.0000 `R` | 62.0000 `R` | 62.0000 `R` |

Every cell the guard rejects while CPU still looks fine — `(0.35, 63)`, `(0.30, 61)`, `(0.25, 45)`, `(0.15, 31)` — is a cell where MPS silently returns something else. And the tap values explain it exactly:

```
h=0.35  ks=61  k_half=30  exp=5.953e-38  cpu_f32=5.953e-38  mps_f32=5.953e-38   (normal)     allowed
h=0.35  ks=63  k_half=31  exp=3.419e-39  cpu_f32=3.419e-39  mps_f32=0.000e+00   (subnormal)  raises
```

So `>= finfo.tiny` is not an approximation of the underflow point — it *is* the CPU-subnormal / MPS-zero boundary, which makes the choice of `tiny` over the true smallest subnormal the load-bearing decision rather than a rounding of convenience. The code comment says this; it now has numbers behind it.

The comment's other claim — that the **corner** tap is negligible — also holds under measurement: at `h=0.35, ks=45` the corner tap is subnormal on CPU and flushed on MPS, and both backends still return `0.6929`. Losing it changes nothing; losing the axis-aligned tap changes everything. That is exactly the discrimination the guard needed.

### 2. Misleading comment

Fixed in all three places, and the replacement text is accurate: I re-measured `h=0.01, ks=3, float32` at `2e-4` against the analytic reference on CPU, so "not itself numerically wrong, rejected because MPS disagrees" is the correct account.

### 3. Rebase

The #4147 interaction is resolved: the strict `xfail` `test_convention_small_h_keeps_nonzero_distances_4152` is deleted (correctly — with the fix that call raises `BaseError`, so a `raises=AssertionError` strict xfail would fail outright), the `supports_replicate_padding` / `supports_conv2d` guards and the bfloat16 tolerance are kept, and `test_h_too_small_raises_float32_regression` inverts the wart into a pin of the new behavior. That is the AGENTS.md contract for a documented-defect fix, done properly.

### What I ran

- `tests/contrib/test_conv_distance_transformer.py`: **66 passed** on CPU at float16 / bfloat16 / float32 / float64; **17 passed, 1 skipped** on MPS float32.
- Fail-on-base with `kornia/` at `e0969438`: **6 failures** — `test_h_too_small_raises` at all four dtypes, `test_h_too_small_raises_float32_regression`, and `test_half_precision_matches_float64_reference[cpu-float16]`. (The bfloat16 cell of that last one passes on base, correctly: bfloat16 shares float32's exponent range, so the reported bug is float16-only.)
- **float32 and float64 are byte-identical to base** — 20 configurations, 2D and 3D, `torch.equal` true on all of them. `image.to(compute_dtype)` aliases the caller's tensor instead of cloning at those dtypes; nothing in the loop writes in place, and I confirmed the input is unmutated at float32/float64/float16.
- The headline fix, 32×8 one-row image at the documented default `h=0.35`, `kernel_size=15`, against a float64 reference: float16 `max|err| 4.2115 → 0.0073`, peak `33.8125 → 29.6875` (reference `29.6946`).
- `grep -rnE "#4152|issues/4152" kornia/ tests/` — clean in `kornia/`, and every `tests/` hit is a pin naming the issue.
- Doctest passes, ruff check and format clean, `torch.compile(fullgraph=True)` still compiles the op.

### The one ask

`git merge-tree` against current `main` (`ee6b713f`) reports **`CHANGELOG.md` as the only conflict** — resolve it and this is ready.

While you are in that file, please add the concrete boundary, because the entry currently describes the mechanism but not the observable consequence: at the documented default `h=0.35` in float32, `kernel_size >= 63` now raises, and on CPU those calls used to return a usable answer (`0.7545` vs `0.7486` at the allowed `ks=61` on the probe above). That is a deliberate trade — MPS returns `1.7918` and `62.0000` for the same calls — but a user who hits it should be able to find it in the changelog. One sentence is enough.

### Non-blocking

- `KORNIA_CHECK` is disabled under `python -O` or `KORNIA_CHECKS=0`, so the guard vanishes there and the small-`h` band returns the all-zero field again. I checked that the *reported* defect is still fixed in that mode — float16 at the default `h`, `ks=15` gives `max|err| = 0.0073` with checks off, because the upcast is unconditional. Consistent with the two sibling `KORNIA_CHECK`s in the same function, so no change requested.
- Under `fullgraph=True` a guard failure surfaces as `torch._dynamo.exc.Unsupported` rather than `BaseError`. Pre-existing: the `h > 0` check on base behaves identically.
- Unrelated to this PR, found while running its gradients: `distance_transform` returns **NaN** gradients for binary inputs on both revisions (175–193 of 256 entries on a random mask; 15 of 16 for a one-hot input; all 16 for an all-zero input). Filed separately — do not touch it here.

ducha-aiki:
The pre-existing NaN-gradient finding from the last section is now **#4183**, with the repro, the `nan_to_num`-after-`log` mechanism, and a candidate guard that is byte-identical in the forward. It is out of scope here — nothing in this PR changes it (verified identical on both sides) — so please do not pick it up in this branch.

ducha-aiki (dismissed):

ducha-aiki (approved):

Issue #4151, opened 2026-09-01
[Bug]: augmentations raise a device-mismatch RuntimeError on MPS/CUDA since #3722 (449 of 457 MPS augmentation-suite failures)


ducha-aiki:
## What happened?

Since #3722 (`9986de7e`, 2026-05-23), calling an augmentation on a non-CPU tensor raises a device-mismatch `RuntimeError` unless the module was explicitly moved to that device first:

```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, mps:0 and cpu!
```

`aug(x_gpu)` without a preceding `aug.to(gpu)` is a documented, supported call pattern — `_BasicAugmentationBase` pins the RNG to CPU on purpose (`set_rng_device_and_dtype(torch.device("cpu"), torch.get_default_dtype())` at `kornia/augmentation/base.py:107`), and every generator's docstring says so: *"By default, the parameters will be generated on CPU in float32."* Before #3722 this worked; the parameters were sampled on CPU and only the transform ran on the device.

#3722 replaced the device-agnostic Python branch

```python
transformed if bool(to_apply.any()) else not_transformed
```

with a `torch.where` blend at four sites:

| site | affected |
| --- | --- |
| `kornia/augmentation/base.py:360` (`_blend_by_prob`) | any 2D/3D augmentation with `p < 1.0` or `p_batch < 1.0` |
| `kornia/augmentation/_2d/base.py:139` | the 2D transform-matrix blend |
| `kornia/augmentation/_3d/base.py:105` | the 3D transform-matrix blend |
| `kornia/augmentation/_2d/mix/base.py:113` | mix augmentations (MixUp, CutMix, Mosaic, Jigsaw), including at `p = 1.0` |

`to_apply` derives from `params["batch_prob"]`, which lives on the RNG device (CPU), while both branches live on the input's device. `torch.where` only accepts a cross-device condition when it is 0-dim:

```python
a = torch.zeros(2, 1, 1, 1, device="mps"); b = torch.ones_like(a)
torch.where(torch.tensor([True, False]).view(2, 1, 1, 1), a, b)  # RuntimeError
torch.where(torch.tensor(True), a, b)                            # OK
```

[...]

Issue #4159, opened 2026-09-02
Track MPS float32 test failures and pin them as strict xfails for the MPS CI leg


ducha-aiki:
## Goal and status

Track and reduce the known MPS (`--device=mps`, `float32`) test failures while keeping the MPS CI leg added in #4160 blocking. The exact 605 failures from #4160's first hosted MPS run are recorded in `testing/known_failure_xfails/mps_float32.txt`, including their exception types. `--xfail-known-failures` applies that complete manifest as strict xfails: a new failure, a changed exception type, a skipped pin, or an XPASS fails CI.

This is the MPS half of #4033.

## Baseline

The authoritative baseline is #4160's completed hosted run [33575062599](https://github.com/kornia/kornia/actions/runs/33575062599): `macos-26-arm64`, Python 3.13, PyTorch 2.9.1, `PYTORCH_ENABLE_MPS_FALLBACK` unset, and one pytest process for the full suite:

- **7,603 passed**, **605 failed**, 4,139 skipped, 35 xfailed, 8 xpassed.

A local per-directory run at `dfad8156` with Python 3.11.14 and the same PyTorch version found 602 failures. All 602 are in the hosted set. The three additional hosted failures are `tests/integration/test_conversions.py::TestAngleOfRotations::test_log_quaternion[mps-float32-0-{x,y,z}]` (`Tensor-likes are not close!` at angle 0). They reproduce locally only after earlier directories have run in the same process, so the checked-in manifest intentionally follows the hosted full-suite result rather than the per-directory result.

Two environments, same result: a local Apple M1 (8 GiB, macOS 26.5.1) and the GitHub-hosted `macos-latest` image (`macos-26-arm64`, virtual M1, 3 cores, 7 GiB, 4.67 GiB recommended MPS memory) produced byte-identical failing sets on the three directories compared so far (`tests/filters`, `tests/color`, `tests/geometry`: 119 of 119 node IDs, probe run https://github.com/kornia/kornia/actions/runs/33571231778). The `macos-14` image segfaults inside MPS kernels and is not usable.

Reproduction:

```bash
# Blocking full-suite contract: expected failures become strict xfails.
env -u PYTORCH_ENABLE_MPS_FALLBACK pixi run test-mps

# Focused diagnosis without requiring every manifest entry to be collected.
env -u PYTORCH_ENABLE_MPS_FALLBACK pixi run test-module tests/geometry --device=mps --dtype=float32
```

[...]

PR #4160, opened 2026-09-02
ci: add a blocking MPS regression gate

ducha-aiki:
## Summary

- add one blocking MPS job to pull-request CI: Python 3.13, PyTorch 2.9.1, float32, on the GitHub-hosted Apple-silicon `macos-latest` image
- pin the complete MPS float32 failure set in `testing/known_failure_xfails/mps_float32.txt`, including each exact exception type, and apply it as strict xfails through `--xfail-known-failures`
- fail CI when an unlisted test fails, a recorded test fails differently, is skipped, or unexpectedly passes
- fix `RandomTransplantation` / `RandomTransplantation3D` on MPS: with no `excluded_labels`, the `all()` reduction over the empty excluded-label axis returns an undefined value there (usually `False`, and it varies between processes), so every donor label was dropped and the output equalled the input. The filter is now skipped when there is nothing to exclude; a device-parametrized pin covers it.
- document how to reproduce, diagnose, remove, rename, and deliberately refresh the baseline in `TESTING.md`, `AGENTS.md`, and issue #4159

This turns the original advisory leg into a regression gate. A fix in Kornia or PyTorch produces an XPASS and makes the job red until its manifest entry is removed; an additional failure is also red until it is fixed or understood and documented.

The manifest is intentionally a full-suite contract. Its node IDs are checked at collection time, and the outcome tracker accepts only a call-phase strict xfail caused by the recorded exact exception type. Setup/teardown failures, alternate xfail markers, skips, wrong exception types, missing tests, and fixes all fail the run. "Full suite" means the suite as an `mps`-only run collects it: `device_agnostic` tests (#4167) are deselected there and covered by the CPU jobs.

## Why macos-latest and not macos-14

Probe run https://github.com/kornia/kornia/actions/runs/33571231778 showed that macos-15-arm64 and macos-26-arm64 (the current macos-latest) expose a working MPS device and produce byte-identical sampled failure sets. macos-14-arm64 segfaults inside MPS kernels.

## Baseline

The first hosted run (https://github.com/kornia/kornia/actions/runs/33575062599, 12,390 tests, 605 failures) predates #4164, #4147, #4167 and #4169. After rebasing onto `main` the manifest was regenerated: 202 entries remain. The hosted run that confirms it is https://github.com/kornia/kornia/actions/runs/33676811355 (`tests-mps`: 7,964 passed, 237 xfailed, 8 xpassed, 4,128 skipped, 69 `device_agnostic` deselected, no unexpected outcome, 22m17s of pytest, 25 minutes wall clock against 61 before).

- 449 augmentation failures were one regression (#4151, fixed by #4164); 51 remain in `tests/augmentation`, all missing `aten::grid_sampler_3d_backward` / `grid_sampler_2d_backward` kernels plus one `RandomAffine3D` tolerance case that the device error used to mask.
- 5 `RandomTransplantation` failures were the undefined empty-axis `all()` above and are fixed here.
- `TestFindEssential::test_shape[mps-float32-1000-5]` became `[16-5]` in #4169.
- The three `test_log_quaternion[mps-float32-0-{x,y,z}]` cases still fail only when earlier directories ran in the same process; they reproduce locally with the CI directory order and stay pinned.

Of the 202 entries, 176 are `NotImplementedError` from missing MPS kernels in torch 2.9.1 (`_linalg_eigh`, `linalg_lu_solve`, `grid_sampler_3d_backward`, `_linalg_eigvals`, `_linalg_svd`, `linalg_qr`, `grid_sampler_2d_backward`, `adaptive_max_pool3d`), 9 are `RuntimeError` (bicubic upsampling, nearest 3D grid sampling, and the empty-input YUV reshape), and 17 are `AssertionError` tolerance or accuracy cases. The per-file breakdown is in #4159.

## Verification

- `tests/test_known_failures.py`: 16 passed
- `TestRandomTransplantation` on MPS: 16 passed, 2 skipped, stable across repeated processes (5 failed before, flipping between runs); CPU: 18 passed
- full MPS/float32 collection with `--xfail-known-failures`: 12,595 tests collected, every manifest node ID resolved
- full-suite single-process `pixi run test-mps` locally (PYTORCH_ENABLE_MPS_FALLBACK unset): 7,963 passed, 237 xfailed (202 strict manifest entries + 35 pre-existing), 8 xpassed (pre-existing non-strict), 4,128 skipped, 69 device_agnostic deselected, no unexpected failure, 6m43s
- `pre-commit run` on the touched files: passed
- hosted `tests-mps` on this branch at `7f8cb4e4`: green on the first run, run 33676811355


ducha-aiki:
## #4164 has merged — this manifest needs a refresh before the MPS job can be green

`fix(augmentation): align probability masks with input devices` landed on `main` as `8f8287f` and closed #4151. The strict tracker in this PR makes a fixed pin red (XPASS), so `testing/known_failure_xfails/mps_float32.txt` is now stale against `main`.

### What is stale

Exactly **449 of the 605 entries** are the #4151 class: every `tests/augmentation/**` entry recorded as `RuntimeError`. That is the same 449 #4151 measured, and its raise-site breakdown adds up to it (205 `_2d/base.py:139` + 112 `base.py:360` + 109 `_3d/base.py:105` + 23 `_2d/mix/base.py:113`).

The other **5** `tests/augmentation` entries are not this class and stay as they are — 4 `AssertionError` plus 1 `pytest.fail.Exception`, all `TestRandomTransplantation`, which #4151's bisect showed already failing at `697d662d` (#3722^).

Mechanical starting point:

```bash
grep -vP '^RuntimeError\ttests/augmentation/' testing/known_failure_xfails/mps_float32.txt
# 156 entries remain (4 header lines excluded)
```

### Do not delete all 449 blindly — some will fail *differently* rather than XPASS

#4151's post-fix measurement left 11 failures in `tests/augmentation`. #4147 (merged, in this PR's base) already removed the false-red ones from the baseline — `TestColorJitterGen::{test_random_gen,test_same_on_batch}` and `test_mixed_image_bbox_dtypes[mps-image_dtype2]` are correctly absent from the manifest — and 3 of the survivors are the `TestRandomTransplantation` entries above. That leaves these **9 pinned nodes that are expected to keep failing after #4164, with a different exception type**:

| pinned as | node | expected after #4164 |
| --- | --- | --- |
| `RuntimeError` | `tests/augmentation/_3d/test_random_affine_3d.py::TestRandomAffine3D::test_batch_random_affine_3d[mps-float32]` | still fails (#4151 survivor list; exact type unknown until measured) |
| `RuntimeError` ×8 | `tests/augmentation/test_backward_3d.py::TestRandomPerspective3DBackward::test_param[...]` | very likely `NotImplementedError: aten::grid_sampler_3d_backward` — the device error currently masks it, and #4159's failure-class table already attributes `grid_sampler_3d_backward` to `test_backward_3d.py`, yet the manifest has **zero** `NotImplementedError` entries under `tests/augmentation` |

(#4151 measured 4 surviving `TestRandomPerspective3DBackward::test_param` nodes locally; this manifest pins 8, so treat the whole parametrization as suspect.)

Under this PR's contract a recorded test that fails with a different exception type is red, exactly like an XPASS. Those 9 need their exception type **re-recorded**, not removed.

So the refreshed baseline should land at roughly **156–165 entries** (605 − 449, plus up to 9 re-recorded), with the exception-class totals moving from 458 `RuntimeError` / 125 `NotImplementedError` to roughly 9 `RuntimeError` / 125–133 `NotImplementedError` / 20 `AssertionError` / 1 `TypeError` / 1 `pytest.fail.Exception`.

Per step 5 of #4159 the dependable route is to rebase this branch onto `main` and regenerate the manifest from a fresh hosted full-suite run, then check the result against the counts above rather than hand-editing.

### Two things that do not need changing

- **Node IDs are still valid.** All 605 pinned nodes still resolve at `8f8287f` — each file exists and each test function is still defined under its recorded name. #4163 (`tests/contrib/test_visual_prompter.py`, `tests/models/sam3/test_sam3_architecture.py`) touched no manifest node. This was a static file/function-name check, not a pytest collection run.
- **#4164's new tests need no entries.** The four cases it adds in `tests/augmentation/test_base.py::TestDeviceAgnosticAugmentationParameters` were verified passing on MPS with `PYTORCH_ENABLE_MPS_FALLBACK` unset, and the tracker only fails on unlisted *failures*.

Posted on behalf of @ducha-aiki by Claude Code.

---
_Generated by [Claude Code](https://claude.ai/code)_

ducha-aiki (commented):
No actionable findings at `cacfa7f5`.

I reviewed the strict-xfail tracker across collection and setup/call/teardown outcomes, manifest validation, reusable workflow compatibility, collector wiring, and the `RandomTransplantation` MPS fix and regression coverage. Focused local validation passed on both CPU and MPS. The remaining CI checks were still in progress when reviewed.

PR #4162, opened 2026-09-02
fix(geometry): preserve the float16 quaternion normalization guard

### LINKED ISSUE #4021: normalize_quaternion returns all-NaN in float16 for the zero quaternion: the default eps=1e-12 underflows to 0
### Describe the bug

`normalize_quaternion` returns `[nan, nan, nan, nan]` in `float16` for the zero quaternion. `float32`, `float64` and `bfloat16` all return `[0, 0, 0, 0]` for the same input.

The cause is a dtype-dependent underflow of the **default** `eps`, not the quaternion maths. The implementation is

```python
def normalize_quaternion(quaternion: torch.Tensor, eps: float = 1.0e-12) -> torch.Tensor:
    ...
    return F.normalize(quaternion, p=2.0, dim=-1, eps=eps)
```

(`kornia/geometry/conversions.py`). `F.normalize` divides by `max(||q||, eps)`, but `float16`'s smallest subnormal is ≈ `6.1e-5`, so the default `1e-12` rounds to exactly `0.0` and the clamp becomes a no-op. The division is then `0 / 0 → NaN`.

`bfloat16` is unaffected because its exponent range is `float32`'s, so `1e-12` is representable there.

This is the same underflow class as #3966 (`quaternion_exp_to_log`, default `eps=1e-8`, fixed in #3990) — a different function with the same root cause, so fixing one did nothing for the other.

**It is not the same defect as #3952.** #3952 is about the clamp *working* and producing a non-unit result (`[1e-13, 0, 0, 0] -> [0.1, 0, 0, 0]`, a `||q|| / eps` scaling). This issue is about the clamp *not existing* at `float16`, so there is no guard at all. #3952's own repro output records `normalize_quaternion(zeros(4)) == [0, 0, 0, 0]`, which is untrue at `float16`; the two need separate fixes and a fix for either leaves the other standing.

### Reproduction steps

```python
import torch
from kornia.geometry.conversions import normalize_quaternion, quaternion_to_rotation_matrix

    q = torch.zeros(4, dtype=dtype)
    print(dtype, normalize_quaternion(q).tolist())
```

```
torch.float64  [0.0, 0.0, 0.0, 0.0]
torch.float32  [0.0, 0.0, 0.0, 0.0]
torch.bfloat16 [0.0, 0.0, 0.0, 0.0]
torch.float16  [nan, nan, nan, nan]
```

```python
>>> normalize_quaternion(torch.zeros(4, dtype=torch.float16), eps=6.104e-05).tolist()
[0.0, 0.0, 0.0, 0.0]
>>> float(torch.tensor(1e-12, dtype=torch.float16))    # 0.0      <- the default underflows
0.0
>>> float(torch.tensor(1e-12, dtype=torch.bfloat16))   # representable
1.0018652574217413e-12
```

A non-zero quaternion is fine at the same dtype, which is why the suite has never caught it:

```python
>>> normalize_quaternion(torch.tensor([1., 0., 1., 0.], dtype=torch.float16)).tolist()
[0.70703125, 0.0, 0.70703125, 0.0]
```

### Blast radius

Both in-tree callers inherit it.

`quaternion_to_rotation_matrix` calls `normalize_quaternion` first, so the zero quaternion becomes an all-NaN matrix at `float16` where every other dtype returns the identity:

```python
    print(dtype, quaternion_to_rotation_matrix(torch.zeros(4, dtype=dtype)).flatten().tolist())
```

```
torch.float64  [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0]
torch.float32  [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0]
torch.bfloat16 [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0]
torch.float16  [nan, nan, nan, nan, nan, nan, nan, nan, nan]
```

`Quaternion.normalize()` reaches it through the same call (`kornia/geometry/quaternion.py:609`):

```python
>>> Quaternion(torch.zeros(1, 4, dtype=torch.float16)).normalize().data.tolist()
[[nan, nan, nan, nan]]
>>> Quaternion(torch.zeros(1, 4, dtype=torch.float32)).normalize().data.tolist()
[[0.0, 0.0, 0.0, 0.0]]
```

### Expected behavior

`float16` should behave like the other three dtypes: the guard should hold at every dtype the function accepts, so the zero quaternion returns `[0, 0, 0, 0]` rather than NaN. Whatever the eventual answer to #3952 is (return a unit quaternion, raise, or keep passing the zero through), it should be the *same* answer in `float16` as in `float64`, which today it is not.

### Notes for a fix

The shape used for #3966 is a dtype-aware floor. `max(eps, torch.finfo(dtype).tiny)` works — `finfo.tiny` is the right choice over `finfo.eps` for a division guard, since `float16`'s are `6.1e-5` and `9.8e-4` respectively and the larger one would distort ordinary sub-`1e-3` inputs.

One catch a fix has to handle deliberately: `normalize_quaternion` takes `eps` as a public argument, and `eps=0.0` currently means "no guard, let it divide by zero" — pinned by `test_wart_normalize_quaternion_without_eps_divides_by_zero_3952` in `tests/geometry/test_conversions.py`. A plain `max(eps, finfo.tiny)` silently overrides an explicit `eps=0` and trips that pin at `float32`/`float64`. Either the floor applies only to the default, or the pin's contract changes on purpose.

The defect is already recorded in the suite as a `pytest.skip` inside `test_wart_zero_quaternion_becomes_the_identity_matrix_3952`, which skips `float16` with this exact explanation. That skip is what should be removed when this is fixed.

Also note #3954, which is live in the same neighbourhood: `quaternion_to_rotation_matrix` upcasts unbatched half-precision input to `float32`. The `float16` cells above are unbatched, so a fix for #3954 will change what dtype those NaNs come back in without making them stop being NaN.

## Disposition

**n/a — single symbol.** `normalize_quaternion` has no near-duplicate in the ledger; this is a defect report, not a duplication finding.

### Environment

Reproduced on `main` at `4da60edc`, CPU, torch 2.9.1, Python 3.11.14, kornia 0.9.0rc1. Not device-specific — it is a property of the `float16` exponent range, not of a backend.

fly1d:

`normalize_quaternion` passes its default `eps=1e-12` to `F.normalize`, but that value rounds to zero in `float16`. A zero quaternion therefore loses the intended division guard and produces all NaNs.

This change floors a positive `eps` at the smallest positive `float16` subnormal, `5.960464477539063e-08`. The exact value remains a literal so the function stays TorchScript-compatible. A zero quaternion now returns finite zeros, while an explicit `eps=0.0` still disables the guard.

The first revision used `torch.finfo(torch.float16).tiny`, which is the smallest normal value and incorrectly clamped valid subnormal inputs. The revised boundary fixes the zero case without changing representable non-zero `float16` magnitudes. A regression test now covers the smallest positive subnormal directly.

The related `quaternion_to_rotation_matrix` and `ARKitQTVecs_to_ColmapQTVecs` documentation and regression pins are updated to follow the behavior through downstream conversions. This remains a focused fix for #4021; it does not resolve the separate sub-`eps` behavior tracked in #3952.

Fixes #4021.

Follow-up to #3969.

## What did you check?

- Rebased onto the current `main`; the only conflict was the expected `CHANGELOG.md` overlap.
- `pixi run test-module tests/geometry/test_conversions.py --device=cpu --dtype=float16,bfloat16,float32,float64 -k 'normalize_quaternion or zero_quaternion'` (`22 passed, 1 skipped, 3 xfailed`)
- `pixi run test-module tests/geometry/test_conversions.py` (`390 passed, 1 skipped, 11 xfailed`; the skip is the existing unmeasured torch 2.14.0/arm64 kernel-literal case)
- The focused MPS `float16` run (`5 passed, 3 skipped`; the skips are existing backend guards)
- An exhaustive local probe over all 63,486 finite non-zero single-axis `float16` values; each normalized exactly to its sign
- `torch.jit.script(normalize_quaternion)` and `torch.compile(normalize_quaternion, fullgraph=True)` with zero, the smallest subnormal, default `eps`, and `eps=0.0`
- `pixi run pre-commit-all` (passed)
- `pixi run typecheck` (completed with the existing `torch.cuda.amp.custom_fwd` deprecation diagnostic)
- `pixi run doctest` (`645 passed, 15 skipped` for uncached model weights)
- `git diff --check`

## How did AI help?

OpenAI Codex helped review the issue and prior pull-request context, implement the focused code, tests, and documentation changes, and run the verification matrix. I reviewed the final diff and the maintainer's numerical analysis before updating the branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -250,11 +250,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   every merge so that ``to_tensor`` counts the merged boxes correctly instead of
   silently truncating them (#4168).
 
-* `normalize_quaternion` floors a positive `eps` at the smallest normal `float16` value, so its default
-  `eps=1e-12` remains an active division guard and a zero quaternion returns finite zeros instead of NaNs. An
-  explicit `eps=0.0` still disables the guard. This also keeps `Quaternion.normalize`,
-  `quaternion_to_rotation_matrix`, and their downstream pose conversions finite for that input, matching the
-  existing `float32`, `float64`, and `bfloat16` behavior (#4162).
+* `normalize_quaternion` floors a positive `eps` at the smallest positive `float16` subnormal, so its default
+  `eps=1e-12` remains an active division guard and a zero quaternion returns finite zeros instead of NaNs without
+  changing non-zero representable `float16` magnitudes. An explicit `eps=0.0` still disables the guard. This also
+  keeps `Quaternion.normalize`, `quaternion_to_rotation_matrix`, and their downstream pose conversions finite for
+  that input, matching the existing `float32`, `float64`, and `bfloat16` behavior (#4162).
+
 * `Boxes3D.to_tensor` and `Boxes3D.get_boxes_shape` are differentiable again (#1396). Both reduce a box's 8
   vertices to its min/max corner with `amin`/`amax`, which is a genuine kink -- not differentiable in the
   classical sense -- wherever multiple vertices exactly tie for an axis extremum. Every face of an
@@ -279,6 +280,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   mis-provisioned venv fails at setup instead of as a wall of unrelated-looking test failures. The five
   `tests/core/test_small_linalg.py` ONNX cases the import error was masking now export with `external_data=False`,
   which torch 2.6's dynamo exporter requires when the destination is a `BytesIO`. (#4199, #4200)
+
 * `find_homography_dlt(..., solver="lu")` now solves the minimal four-point system with an adaptive
   homogeneous gauge instead of applying LU to its singular normal matrix. This is the estimator
   `RANSAC(model_type="homography")` calls for every minimal sample, so homography RANSAC results
@@ -308,6 +310,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   which backend a caller happens to run on. Concretely, at the documented default `h=0.35` in `float32`,
   `kernel_size >= 63` now raises: CPU used to return a usable-looking answer there while MPS silently returned
   something else for the same call.
+
 * `RandomTransplantation` and `RandomTransplantation3D` transplanted nothing on MPS when no `excluded_labels`
   were given: PyTorch's MPS backend evaluates `all()` over the empty excluded-label axis to an undefined value,
   usually `False`, so every donor label was filtered out and the output equalled the input. The filter is now skipped when there is
@@ -350,6 +353,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   mask now moves to the output device at each blend boundary, preserving CPU random-number generation and the
   ONNX/fullgraph-friendly blend. Calling `augmentation(x_accelerator)` therefore works again without first moving the
   augmentation module or its RNG to the accelerator (#4151).
+
 * Keep `extract_patches_simple` and `extract_patches_from_pyramid` off a broken reduced-precision CPU kernel: for a
   `float16`/`bfloat16` CPU image, the `grid_sample` kernels in torch <= 2.9 read out of bounds when a rotated patch
   straddles the image border and return NaN or garbage values far outside the image's range. Both extractors now
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -254,7 +254,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `eps=1e-12` remains an active division guard and a zero quaternion returns finite zeros instead of NaNs. An
   explicit `eps=0.0` still disables the guard. This also keeps `Quaternion.normalize`,
   `quaternion_to_rotation_matrix`, and their downstream pose conversions finite for that input, matching the
-  existing `float32`, `float64`, and `bfloat16` behavior (#4021).
+  existing `float32`, `float64`, and `bfloat16` behavior (#4162).
 * `Boxes3D.to_tensor` and `Boxes3D.get_boxes_shape` are differentiable again (#1396). Both reduce a box's 8
   vertices to its min/max corner with `amin`/`amax`, which is a genuine kink -- not differentiable in the
   classical sense -- wherever multiple vertices exactly tie for an axis extremum. Every face of an
@@ -279,7 +279,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   mis-provisioned venv fails at setup instead of as a wall of unrelated-looking test failures. The five
   `tests/core/test_small_linalg.py` ONNX cases the import error was masking now export with `external_data=False`,
   which torch 2.6's dynamo exporter requires when the destination is a `BytesIO`. (#4199, #4200)
-
 * `find_homography_dlt(..., solver="lu")` now solves the minimal four-point system with an adaptive
   homogeneous gauge instead of applying LU to its singular normal matrix. This is the estimator
   `RANSAC(model_type="homography")` calls for every minimal sample, so homography RANSAC results
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -250,6 +250,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   every merge so that ``to_tensor`` counts the merged boxes correctly instead of
   silently truncating them (#4168).
 
+* `normalize_quaternion` floors a positive `eps` at the smallest normal `float16` value, so its default
+  `eps=1e-12` remains an active division guard and a zero quaternion returns finite zeros instead of NaNs. An
+  explicit `eps=0.0` still disables the guard. This also keeps `Quaternion.normalize`,
+  `quaternion_to_rotation_matrix`, and their downstream pose conversions finite for that input, matching the
+  existing `float32`, `float64`, and `bfloat16` behavior (#4021).
 * `Boxes3D.to_tensor` and `Boxes3D.get_boxes_shape` are differentiable again (#1396). Both reduce a box's 8
   vertices to its min/max corner with `amin`/`amax`, which is a genuine kink -- not differentiable in the
   classical sense -- wherever multiple vertices exactly tie for an axis extremum. Every face of an
@@ -304,7 +309,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   which backend a caller happens to run on. Concretely, at the documented default `h=0.35` in `float32`,
   `kernel_size >= 63` now raises: CPU used to return a usable-looking answer there while MPS silently returned
   something else for the same call.
-
 * `RandomTransplantation` and `RandomTransplantation3D` transplanted nothing on MPS when no `excluded_labels`
   were given: PyTorch's MPS backend evaluates `all()` over the empty excluded-label axis to an undefined value,
   usually `False`, so every donor label was filtered out and the output equalled the input. The filter is now skipped when there is
@@ -347,7 +351,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   mask now moves to the output device at each blend boundary, preserving CPU random-number generation and the
   ONNX/fullgraph-friendly blend. Calling `augmentation(x_accelerator)` therefore works again without first moving the
   augmentation module or its RNG to the accelerator (#4151).
-
 * Keep `extract_patches_simple` and `extract_patches_from_pyramid` off a broken reduced-precision CPU kernel: for a
   `float16`/`bfloat16` CPU image, the `grid_sample` kernels in torch <= 2.9 read out of bounds when a rotated patch
   straddles the image border and return NaN or garbage values far outside the image's range. Both extractors now
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -735,9 +735,10 @@ def normalize_quaternion(quaternion: torch.Tensor, eps: float = 1.0e-12) -> torc
         ``[nan, nan, nan, nan]`` instead. ``float32`` and ``bfloat16`` behave
         the same up to their rounding (``0.10000000149011612`` and
         ``0.099609375`` for the first input). For ``float16``, a positive
-        ``eps`` is floored at the dtype's smallest normal value, so the default
-        guard remains active and a zero quaternion also returns ``zeros(4)``.
-        An explicit ``eps=0.0`` still disables that guard. The remaining
+        ``eps`` is floored at the dtype's smallest positive subnormal value, so
+        the default guard remains active and a zero quaternion also returns
+        ``zeros(4)`` without clamping non-zero representable magnitudes. An
+        explicit ``eps=0.0`` still disables that guard. The remaining
         sub-``eps`` behavior is tracked in
         `#3952 <https://github.com/kornia/kornia/issues/3952>`_.
 
@@ -761,8 +762,8 @@ def normalize_quaternion(quaternion: torch.Tensor, eps: float = 1.0e-12) -> torc
     if not quaternion.shape[-1] == 4:
         raise ValueError(f"Input must be a tensor of shape (*, 4). Got {quaternion.shape}")
 
-    # Exact value of torch.finfo(torch.float16).tiny, kept literal for TorchScript support.
-    safe_eps = max(eps, 6.103515625e-05) if quaternion.dtype == torch.float16 and eps > 0.0 else eps
+    # Exact smallest positive float16 subnormal, kept literal for TorchScript support.
+    safe_eps = max(eps, 5.960464477539063e-08) if quaternion.dtype == torch.float16 and eps > 0.0 else eps
     return F.normalize(quaternion, p=2.0, dim=-1, eps=safe_eps)
 
 
@@ -802,7 +803,7 @@ def quaternion_to_rotation_matrix(quaternion: torch.Tensor) -> torch.Tensor:
           ``nan`` outright because ``0.5 * 1e6`` overflows the dtype.
           Once ``||q||`` drops below
           :func:`~kornia.geometry.conversions.normalize_quaternion`'s effective
-          floor (``eps = 1e-12``, raised to the smallest normal value in
+          floor (``eps = 1e-12``, raised to the smallest positive subnormal in
           ``float16``), the clamp takes over and rescaling changes the matrix
           outright: in ``float64``, ``q * 1e-13`` and ``q`` can give matrices
           that differ by order 1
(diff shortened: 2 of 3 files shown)


ducha-aiki:
## Review of #4162 — CHANGES REQUESTED, one literal

Setup: worktree at the PR head (`3cb03c6d`) with `kornia/` swapped to the merge-base `dfad8156` for every A/B pass. `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU/macOS; `kornia.__file__` asserted inside the worktree in every block.

The #4021 half is verified. Failure sets over the whole file at all four dtypes:

```
base 49 failed / 1024 passed        head 45 failed / 1028 passed
fixed by this PR (4):
  TestCARKitToColmap::test_wart_zero_quaternion_is_absorbed_3952[cpu-float16]
  TestQuaternionToRotationMatrix::test_convention_normalize_quaternion_zero_uses_the_default_guard_4021[cpu]
  TestQuaternionToRotationMatrix::test_wart_normalize_quaternion_scales_by_norm_over_eps_3952[cpu-float16]
  TestQuaternionToRotationMatrix::test_wart_zero_quaternion_becomes_the_identity_matrix_3952[cpu-float16]
newly broken by this PR: none
```

`grep -rnE "#4021|issues/4021" kornia/ tests/` is clean at head. The docstring edits are accurate and the wart pins were inverted rather than deleted. TorchScript is preserved.

---

## The blocker: the floor is one step too high, and it de-normalises real quaternions

`torch.finfo(torch.float16).tiny` is the smallest **normal** value, `6.103515625e-05`. `F.normalize` divides by `clamp_min(‖q‖, eps)`, so every quaternion whose norm falls in float16's **subnormal** range — `[5.96e-08, 6.10e-05)` — is now divided by the floor instead of by its own norm, and comes back non-unit. Those norms are exactly representable in float16 and `main` normalises them correctly:

```
        input q[0]        base ‖out‖    head ‖out‖
  1.0                          1             1
  1.0001659e-04                1             1
  6.103515625e-05              1             1
  2.9981136e-05                1        0.491211
  1.0013580e-05                1        0.164062
  1.0132790e-06                1        0.016602
  5.9604645e-08                1        0.000977
  0.0                        nan             0        <- the #4021 fix
```

So the PR trades a loud `nan` for a **silent** wrong rotation over a whole band of inputs. The PR body says "a non-zero quaternion below the effective floor **can still** return a non-unit result" — "still" is not right: on `main` those return unit results, so this is new.

**Ask: change the literal to the smallest positive float16 *subnormal*, `5.960464477539063e-08`.**

```python
# Exact value of the smallest positive float16 subnormal, kept literal for TorchScript.
safe_eps = max(eps, 5.960464477539063e-08) if quaternion.dtype == torch.float16 and eps > 0.0 else eps
return F.normalize(quaternion, p=2.0, dim=-1, eps=safe_eps)
```

Measured this variant over **all 63 488 finite single-axis float16 quaternions** — every representable magnitude, not a sample:

```
differs from base on 2 of 63488 inputs
  those inputs: [0.0, -0.0]     base: [nan, nan]     new: [0.0, -0.0]
bit-identical to base over 200 000 random float16 quaternions:  True
zero quaternion with eps=0.0  ->  [nan, nan, nan, nan]          (unchanged)
torch.jit.script(normalize_quaternion)                          OK
```

It fixes exactly #4021 and touches nothing else. `clamp_min(0, 5.96e-08)` still turns `0/floor` into `0`, and any non-zero norm is at least the smallest subnormal so it is never clamped.

## The alternative, if you prefer it

`conversions.py:1148` already carries this exact idiom in this file — *"use float32 for float16/bfloat16 so the default eps is representable (kornia#3966)"*. Routing float16 through float32 also works and I verified it: zero → zeros, all subnormal magnitudes → unit, `eps=0.0` → `nan`, TorchScript OK. Its cost is that it is **not** bit-identical to `main` on normal-range float16 (max abs diff `4.88e-04` over 200 000 quaternions — about half a float16 ulp, and in the more-accurate direction). Either is fine by me; the subnormal literal is the smaller change and keeps bit-parity, so that is what I would take.

One control, so it is not read as a regression from this PR: `eager == torch.compile(fullgraph=True)` is already `False` for `normalize_quaternion` at float16 on `main`, and is equally `False` for both variants. Pre-existing, not yours.

## Also

`CHANGELOG.md` is the only merge conflict (`git merge-tree` auto-merges both other files), so the rebase is one file. Please refresh the CHANGELOG wording for whichever floor you land on.

Really good change otherwise — the docstring updates across `quaternion_to_rotation_matrix` and `ARKitQTVecs_to_ColmapQTVecs` correctly follow the fix through, which is the part these usually miss.

fly1d:
Thanks for the careful boundary analysis. You were right that `finfo.tiny` was one step too high. I changed the guard to `5.960464477539063e-08`, added a regression at that exact subnormal boundary, and corrected the affected docstrings and changelog wording.

I also rebased onto the latest `main`; `CHANGELOG.md` was the only conflict. The complete conversions file now passes (`391 passed, 11 xfailed`), as do the focused CPU and MPS runs, TorchScript/compile checks, pre-commit, typecheck, and doctests. I included the exact commands and results in the updated PR description. Could you take another look when convenient?

ducha-aiki:
## Delta re-review of #4162 at `cad40ff6` — APPROVE

Setup: worktree at the PR head, `kornia/` swapped to the merge-base `4b69473a` for every A/B pass. `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU/macOS.

The blocker is fixed exactly as measured, and I re-ran the whole battery rather than checking the literal by eye.

### The subnormal band is back

```text
all 63488 finite float16 magnitudes on one axis:
  non-zero inputs returning a non-unit quaternion:  0
  zero quaternion:  [0., 0., 0., 0.]  /  [-0., 0., 0., 0.]
torch.jit.script(normalize_quaternion): OK
```

Every magnitude in `[5.96e-08, 6.10e-05)` — the band `finfo.tiny` was silently de-normalising in the previous revision — now normalises correctly again.

### Bit-parity with `main`, and the A/B

```text
200 000 random float16 quaternions over a wide magnitude range:
  bit-identical to the merge-base:  True   (0 rows differ)

tests/geometry/test_conversions.py, float16 + bfloat16 + float32 + float64:
  merge-base 49 failed / head 45 failed
  fixed by this PR (4):
    TestCARKitToColmap::test_wart_zero_quaternion_is_absorbed_3952[cpu-float16]
    TestQuaternionToRotationMatrix::test_convention_normalize_quaternion_zero_uses_the_default_guard_4021[cpu]
    TestQuaternionToRotationMatrix::test_wart_normalize_quaternion_scales_by_norm_over_eps_3952[cpu-float16]
    TestQuaternionToRotationMatrix::test_wart_zero_quaternion_becomes_the_identity_matrix_3952[cpu-float16]
  newly broken by this PR:  none
```

`grep -rnE "#4021|issues/4021" kornia/ tests/` is clean at the head.

### The rest of the delta

`test_convention_normalize_quaternion_preserves_float16_subnormal_4021` pins the exact boundary value `5.960464477539063e-08` — that is the right pin, since it is the one input where the floor and the smallest representable norm coincide, so a future edit that moves the literal up by one step fails here rather than silently. The docstring rewrites followed the fix through all four sites, including the two downstream warnings on `quaternion_to_rotation_matrix` and `ARKitQTVecs_to_ColmapQTVecs`, and the `.. warning::` on the latter is now dtype-uniform, which it genuinely is after this change. The CHANGELOG conflict is resolved and the wording matches the floor you landed on.

Nothing open. Thank you for taking the boundary analysis seriously rather than nudging the constant — the subnormal-vs-normal distinction is the whole content of this fix, and the regression test now carries it.

ducha-aiki (dismissed):

ducha-aiki:
Approved at `cad40ff6` and nothing has changed that — this went `CONFLICTING` only because `main` took a batch of CI work overnight (it is at `7efa3684` now). `git merge-tree origin/main <head>` reports **`CHANGELOG.md` only**.

Since one of those merges was **#4154** (Linux CPU half-precision CI, with strict known-failure manifests under `testing/half_precision_xfails/`), I checked whether your rebase also needs a manifest edit — a PR that flips float16 results green normally has to delete the corresponding lines, or the strict baseline goes red on an XPASS. I simulated the rebase by applying your `kornia/` **and** `tests/` diff onto current `main` and running the file the way CI will:

```text
--dtype=float16   --xfail-known-failures --known-failure-profile=cpu-float16    361 passed,  8 skipped, 33 xfailed
--dtype=bfloat16  --xfail-known-failures --known-failure-profile=cpu-bfloat16   369 passed,  1 skipped, 32 xfailed
--dtype=float32,float64                                                          617 passed,  2 skipped, 15 xfailed
```

**No XPASS, no failure — no manifest edit needed.** The reason is that the four float16 results you flip are all on tests your own diff adds or rewrites (the two `_4021` convention pins, and the `_3952` wart pin you correctly rename from `..._is_absorbed_or_nans_by_dtype_3952` to `..._is_absorbed_3952` because your fix removes its float16 NaN leg). None of them was in the baseline #4154 recorded, so nothing there is falsified.

So: rebase for the CHANGELOG and it is ready to land. Nothing else on my side.

fly1d:
Rebased onto `7c5db705` and resolved the `CHANGELOG.md` conflict, preserving the current entries. I did not change the implementation, tests, or half-precision failure manifests.

The focused dtype run passes (`22 passed, 1 skipped, 3 xfailed`), as does the full conversions file (`390 passed, 1 skipped, 11 xfailed`) and `pixi run pre-commit-all`. The rebased commits are SSH-signed, and GitHub verifies the new head `b676eb64`. Thanks for checking the new half-precision baseline.

ducha-aiki:
## Delta re-review of #4162 at `b676eb64` — still APPROVED; the red check is not yours

You rebased and split the commit; I re-checked whether anything moved in the process. It did not:

```console
$ diff <(git show cad40ff6:kornia/geometry/conversions.py) <(git show pr4162:kornia/geometry/conversions.py)
$
```

`kornia/geometry/conversions.py` is **byte-identical** to the revision I approved. The whole delta is the rebase onto `7c5db705` plus the `CHANGELOG.md` placement, which is what resolved the conflict. Nothing to re-verify; the approval stands.

### The `half-precision (float16)` red is a CI hang in a file you do not touch

Do not chase it and do not change anything for it. It was **cancelled at the platform cap**, not failed:

```text
job:      half-precision (float16) / tests (3.11, 2.9.1)
status:   cancelled
started:  2026-09-04T02:12:13Z   completed: 2026-09-04T08:12:43Z    (6 h 0 m 30 s)
step 11 "Run tests" -> cancelled;  steps 12-13 skipped
```

The log shows a single 21,428-second gap with no output in between:

```text
02:15:33  tests/feature/test_responces_local_features.py ........x.......x.   [ 58%]
08:12:40  ##[error]The operation was canceled.
```

so the run wedged on the first test of `tests/feature/test_scale_space_detector.py` — nothing in `tests/feature` is in your diff, which is three files: `CHANGELOG.md`, `kornia/geometry/conversions.py`, `tests/geometry/test_conversions.py`. Every other run of that same leg today finished in **6–8 minutes**, and the runner was the faster Intel Xeon 8573C, so this is a hang and not slowness. I could not reproduce it locally (`tests/feature/test_scale_space_detector.py --device=cpu --dtype=float16` → 64 passed / 1 skipped in 17.85 s on torch 2.14.0), which fits a nondeterministic CI-only wedge.

I am filing the underlying gap separately — no job in `.github/workflows/` sets `timeout-minutes`, so a hang costs the full six hours and surfaces as an unexplained red instead of a timeout. Nothing for you to do; a re-run should clear it, and `collector` will go green with it.

Ready to merge once that check is re-run.

fly1d:
main advanced to `8d2fe3bf` shortly after your re-review, so I rebased onto the new tip. The only conflict was `CHANGELOG.md`; the implementation, tests, and half-precision manifests are unchanged.

The new head is `eb5eac3f`, and GitHub verifies all three SSH signatures. Local verification after the rebase: the focused dtype run passed (`22 passed, 1 skipped, 3 xfailed`), and `pixi run pre-commit-all` passed. The push has started a fresh CI run.

fly1d:
Rebased onto `8db1499a`; `CHANGELOG.md` was again the only conflict. The approved implementation and tests are byte-identical to `eb5eac3f`, and the changelog diff now contains only the #4162 entry.

The new head is `6d705e54`, and GitHub verifies the SSH signatures. The focused CPU dtype run still passes (`22 passed, 1 skipped, 3 xfailed`), as does `pixi run pre-commit-all`.

ducha-aiki:
## Re-review at `6d705e54` — the subnormal floor is right; one pin came out that should go back

You moved the float16 floor from `finfo.tiny` (`6.103515625e-05`, the smallest **normal**) to
`5.960464477539063e-08` (`2**-24`, the smallest positive **subnormal**). That is the correct constant and it
fixes a real defect in the previous revision I approved — thank you for catching it.

**Verified at head:**

```
normalize_quaternion(zeros(4), f16)                 -> [0., 0., 0., 0.]        (guard still active)
normalize_quaternion([2**-24, 0, 0, 0], f16)        -> [1., 0., 0., 0.]        (was 0.000977 on the old floor)
normalize_quaternion(zeros(4), bf16/f32/f64)        -> [0., 0., 0., 0.]        (unchanged)
```

The middle line is the one that matters: with the smallest-normal floor, every float16 magnitude in the whole
subnormal band was being divided by a number larger than itself, so a legitimately representable quaternion came
back de-normalised instead of normalised. Your new pin discriminates on exactly that. `tests/geometry/test_conversions.py`
-> **617 passed** (cpu, float32+float64) and **361 passed / 33 xfailed** under the CI half profile
(`--xfail-known-failures --known-failure-profile=cpu-float16`), **no XPASS**, so still no manifest edit.

### The one ask

The delta dropped this from `test_convention_..._4021`:

```python
without_guard = kornia.geometry.conversions.normalize_quaternion(quaternion, eps=0.0)
assert torch.isnan(without_guard).all()
```

but both the CHANGELOG bullet and the docstring still say *"An explicit `eps=0.0` still disables that guard."*
That claim is now unpinned. It also still holds — I ran it against your head:

```
normalize_quaternion(zeros(4), f16, eps=0.0) -> [nan, nan, nan, nan]   (all nan: True)
```

So it is not that the assertion broke and had to go; it just went. Please put it back, either in the original
test or in the new subnormal one. A prose claim that no test executes is the thing that goes stale first, and
this file has two of them pointing at the same line.

Nothing else outstanding. Everything I approved earlier is unchanged apart from this hunk, and the CHANGELOG /
docstring wording is now accurate about which value the floor is.

fly1d:
Good catch. I restored the `eps=0.0` assertion in `47497ee4`, so the documented opt-out is exercised again. The exact test passes, as do the full conversions runs for float16 (`361 passed, 33 xfailed`), bfloat16 (`369 passed, 32 xfailed`), and float32/float64 (`617 passed, 15 xfailed`), with no XPASS. `pixi run pre-commit-all` also passes. Thanks for pointing out the missing pin.

ducha-aiki:
Delta re-review at `47497ee4` — **APPROVED, merge-ready**.

The delta from `6d705e54` is one commit, three added lines in `tests/geometry/test_conversions.py`: the `eps=0.0` opt-out pin (`assert torch.isnan(without_guard).all()`) that the 09-04 round asked to restore. `kornia/geometry/conversions.py` is byte-identical to the approved head. The restored pin discriminates: dropping ` and eps > 0.0` from `conversions.py:766` makes exactly that assertion fail, and head passes.

Re-run at this head: CPU float32+float64 617 passed; float16 profile 361 passed / 33 xfailed, bfloat16 369 / 32, no XPASS so no manifest edit; MPS float32 and float16 failure sets identical to `main`. Numerics unchanged from the earlier rounds: all 63,488 single-axis float16 magnitudes normalise to unit, and on 200k random float16 quaternions the only rows that differ from `main` are the 5,777 with zero norm (NaN on `main`, zeros here). CHANGELOG entry under `### Bug fixes` with nothing of `main`'s deleted; `#4021` grep clean; `pre-commit` passes; CI 40/40 green and the PR is `MERGEABLE`. Nothing left on your side.

ducha-aiki (approved):

PR #4163, opened 2026-09-02
test: speed up VisualPrompter and SAM3 coverage

ducha-aiki:
## Summary

- replace pretrained ViT-H construction in `VisualPrompter` coverage with an explicit untrained MobileSAM configuration
- remove the duplicate contrib invalid-shape test; the broader validation test remains in `tests/models/test_prompter.py`
- preserve fast coverage of the public `VisualPrompter()` default with a mocked test that verifies pretrained ViT-H is still requested without building or downloading it
- keep the inexpensive MobileSAM construction smoke test in the default device-parametrized suite, while the batching pipeline remains slow-marked
- reduce the SAM3 forward-test image sizes while retaining both the explicit 1024-by-1024 output-shape contract test and the 1024 forward doctest

No production default or runtime behavior changes.

## Why

In PR #4160's hosted MPS run, `tests/contrib/test_visual_prompter.py::TestVisualPrompter::test_exception` spent 94.62 seconds constructing the default pretrained ViT-H model before checking a malformed input that fails prior to inference. On a cold cache this also attempts to fetch the 2.39 GiB ViT-H checkpoint. This VisualPrompter saving applies broadly, including local runs.

The SAM3 tests do not load weights, but their forward cases used quadratic-attention inputs larger than needed for smoke and batch-cardinality coverage. On GitHub's virtual-M1 runner, the 1024 smoke case took 15.19 seconds and the batch-4 cardinality case took 17.40 seconds; local MPS timings are much smaller, so this part primarily targets hosted-runner wall time. The smaller cases exercise the same paths, while 1024 support remains covered as described above.

## Verification

- focused default CPU run across both files: 20 passed, 3 skipped in 0.58 seconds
- focused default MPS run across both files: 17 passed, 6 skipped in 1.36 seconds
- focused CPU run with slow `VisualPrompter` coverage: 3 passed, 2 skipped in 2.11 seconds
- `pre-commit run --all-files`: passed
- independent review: no blocking findings

The slow MobileSAM batching path also exposes the pre-existing augmentation device mismatch tracked in #4151 (`torch.where` receives an MPS input and a CPU selection tensor). The fix is being handled separately in #4164. This test-only PR intentionally does not alter that production path; CPU pipeline coverage passes, and the normal MPS PR leg skips slow tests.

### DIFF
--- a/tests/contrib/test_visual_prompter.py
+++ b/tests/contrib/test_visual_prompter.py
@@ -19,17 +19,40 @@ import pytest
 import torch
 
 from kornia.contrib.visual_prompter import VisualPrompter
+from kornia.models.sam import Sam, SamConfig
 
 from testing.base import BaseTester
 
 
 class TestVisualPrompter(BaseTester):
+    def test_default_config(self, monkeypatch):
+        captured_configs = []
+
+        class FakeSam:
+            image_encoder = type("ImageEncoder", (), {"img_size": 1024})()
+
+            def to(self, **kwargs):
+                return self
+
+        def fake_from_config(config):
+            captured_configs.append(config)
+            return FakeSam()
+
+        monkeypatch.setattr(Sam, "from_config", staticmethod(fake_from_config))
+
+        prompter = VisualPrompter()
+
+        assert prompter is not None
+        assert len(captured_configs) == 1
+        assert captured_configs[0].model_type == "vit_h"
+        assert captured_configs[0].pretrained is True
+
     @pytest.mark.slow
     def test_smoke(self, device, dtype):
         if dtype not in (torch.float32, torch.float16):
             pytest.skip("VisualPrompter (SAM) primarily supports float32 and float16")
 
-        prompter = VisualPrompter(device=device, dtype=dtype)
+        prompter = VisualPrompter(SamConfig("mobile_sam"), device=device, dtype=dtype)
         assert prompter is not None
         assert not prompter.is_image_set
 
@@ -37,7 +60,7 @@ class TestVisualPrompter(BaseTester):
     def test_batching_pipeline(self, device, dtype):
         if dtype not in (torch.float32, torch.float16):
             pytest.skip("VisualPrompter (SAM) primarily supports float32 and float16")
-        prompter = VisualPrompter(device=device, dtype=dtype)
+        prompter = VisualPrompter(SamConfig("mobile_sam"), device=device, dtype=dtype)
 
         batch_size = 2
         image = torch.rand(batch_size, 3, 256, 256).to(device=device, dtype=dtype)
@@ -55,14 +78,6 @@ class TestVisualPrompter(BaseTester):
         results = prompter.predict(boxes=boxes_tensor)
 
         assert results.logits.shape == (batch_size, 3, 256, 256)
-
-    def test_exception(self, device, dtype):
-
-        prompter = VisualPrompter(device=device, dtype=dtype)
-
-        image = torch.rand(1, 2, 3, 256, 256).to(device=device, dtype=dtype)
-        with pytest.raises(Exception):
-            prompter.set_image(image)
 
     def test_gradcheck(self, device):
         pytest.skip("Gradcheck is not currently applicable for VisualPrompter inference.")
--- a/tests/contrib/test_visual_prompter.py
+++ b/tests/contrib/test_visual_prompter.py
@@ -15,6 +15,8 @@
 # limitations under the License.
 #
 
+from types import SimpleNamespace
+
 import pytest
 import torch
 
@@ -29,7 +31,7 @@ class TestVisualPrompter(BaseTester):
         captured_configs = []
 
         class FakeSam:
-            image_encoder = type("ImageEncoder", (), {"img_size": 1024})()
+            image_encoder = SimpleNamespace(img_size=1024)
 
             def to(self, **kwargs):
                 return self
@@ -42,12 +44,11 @@ class TestVisualPrompter(BaseTester):
 
         prompter = VisualPrompter()
 
-        assert prompter is not None
+        assert prompter.model.image_encoder.img_size == 1024
         assert len(captured_configs) == 1
         assert captured_configs[0].model_type == "vit_h"
         assert captured_configs[0].pretrained is True
 
-    @pytest.mark.slow
     def test_smoke(self, device, dtype):
         if dtype not in (torch.float32, torch.float16):
             pytest.skip("VisualPrompter (SAM) primarily supports float32 and float16")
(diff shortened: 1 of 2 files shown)


ducha-aiki:
## Review: approve

I reproduced every claim in the description locally (macOS, torch 2.9.1, `.venv`), on a worktree of `e69e8876` against `dfad8156`. The headline number holds and the coverage argument holds. Details, plus three follow-ups and one optional suggestion.

### What I verified

**The default (non-slow) leg gets its time back.** Both files, `--device=cpu --dtype=float32`, 19 passed / 4 skipped on both sides:

| | main `dfad8156` | PR `e69e8876` |
|---|---|---|
| whole run | **29.03 s** | **0.87 s** |
| `tests/contrib/…::test_exception[cpu-float32]` | 27.58 s | *(removed)* |

The 27.58 s is with the ViT-H checkpoint already warm in `weights/checkpoints/` — the file is 2,564,550,879 bytes here, so on the cold cache CI actually has, the removed test was a 2.4 GB fetch. And it carried no `@pytest.mark.slow`, so it ran on **every** device leg. That is the whole finding.

**The coverage it dropped is genuinely still there.** `tests/models/test_prompter.py::TestVisualPrompter::test_exception` asserts `ShapeError` for exactly the same `(1, 2, 3, 256, 256)` input to `set_image`, plus `_valid_masks` / `_valid_boxes` / `_valid_keypoints` and the "no image set" path. It builds an *untrained* `vit_b`, is not slow-marked, and costs **0.32 s**. So the deletion trades 27.58 s for nothing.

**`test_default_config` is wired correctly.** `visual_prompter.py` imports `Sam` from `kornia.models.sam`, and the test patches `from_config` on that same class object, so the interception is real rather than incidental. It pins `model_type == "vit_h"` and `pretrained is True`, which is the actual regression risk. Passes in 0.01 s.

**The SAM3 reductions are behaviour-preserving.** `ImageEncoderHiera` is a plain ViT despite the name — `PatchEmbedding` → `pos_embed` → N × `ViTBlock` → `LayerNorm`, no windowing, no hierarchy, no size-dependent branch. `forward` only checks `x.shape[-2:] == img_size` and adds a `pos_embed` of `(img_size // patch_size) ** 2`. 512→256 and 1024→512 therefore run byte-for-byte the same code path with a smaller `N`. `embed_dim` 256→128 / `num_heads` 8→4 in the cardinality test is fine too (`Attention` head_dim 32), and `test_attention_smoke` already covers wider dims independently.

**Bonus confirmation you didn't claim.** `tests/core/test_weights_prefetch.py` exempts `sam_vit_h_4b8939.pth` with the reason *"only vit_b and mobile_sam have a pretrained test"*. That was **false on main** — the deleted contrib test was a third pretrained-ViT-H call site, and the only unmarked one. This PR makes the exemption's stated reason true. The prefetch suite is still 84 passed on the branch.

**Other checks:** slow CPU leg on the branch 3 passed / 2 skipped in 2.65 s (`test_batching_pipeline` 2.01 s); MPS default leg 16 passed / 7 skipped; `ruff check` and `ruff format --check` clean on both files. Test-only, so no `CHANGELOG.md` entry is owed under the AGENTS.md policy.

### Follow-ups (none blocking)

**1. Name the augmentation bug — it is already filed as #4151.** I reproduced your MPS slow failure: `tests/contrib/test_visual_prompter.py::TestVisualPrompter::test_batching_pipeline[mps-float32]` fails at `kornia/augmentation/base.py:360` in `_blend_by_prob`, `RuntimeError: Expected all tensors to be on the same device, but found at least two devices, mps:0 and cpu!`. That is #4151 (the `torch.where` blend from #3722 against a CPU `batch_prob`), pre-existing and independent of the SAM variant — main hits the identical line. Worth citing the number in the description rather than describing it anonymously, so the trail from this PR to the fix exists.

**2. Local numbers don't reproduce the SAM3 saving, and that's expected — say so.** On my machine `tests/models/sam3/test_sam3_architecture.py` on MPS runs in 5.14 s total on main, with `test_image_encoder_hiera_smoke[mps-1024-256]` at **0.14 s**, not 15.19 s. The hosted figures are GitHub's virtual-M1 runner being ~100× slower, not anything intrinsic to the test. The SAM3 half of this PR is a hosted-runner-only win of ~30 s; the VisualPrompter half is the part that is large everywhere. Both are worth doing, but the description reads as though they are the same kind of saving.

**3. The 1024² forward is not actually lost — and the thing that still runs it is now the most expensive SAM3 item in CI.** The `ImageEncoderHiera.forward` docstring example builds `img_size=1024, patch_size=16, embed_dim=768` (default `depth=12`) and runs a real forward. It costs **6.6–13.2 s of CPU** in `pixi run doctest`, i.e. considerably more than the 0.21 s CPU / 15.19 s MPS smoke case this PR trims, on a config far heavier than any test. Good news for this PR — 1024 coverage survives — but it is the obvious next target if the goal is CI wall-clock. `depth=2` in that example would keep the shape output identical and cut it to well under a second.

### Optional suggestion

After this change `test_smoke` costs **0.06 s on CPU / 0.74 s on MPS** — untrained MobileSAM, no download. It is still `@pytest.mark.slow`, which means that after this PR **nothing in the default leg moves a SAM model onto the test device at all**; the surviving validation test in `tests/models/test_prompter.py` takes no `device` fixture. Dropping the marker from `test_smoke` (keeping it on `test_batching_pipeline`, which is 2.0 s CPU / 41 s MPS) would restore device-parametrized construction coverage for under a second. Your call — the on-device value of the old ViT-H test was only "a 2.4 GB model can be moved to MPS", so it is a small loss either way.

### Nits

- `assert prompter is not None` is vacuous — a constructor result is never `None`. Pre-existing style in this file, and the new test repeats it; `assert prompter.model.image_encoder.img_size == 1024` would assert something.
- `type("ImageEncoder", (), {"img_size": 1024})()` reads more plainly as `types.SimpleNamespace(img_size=1024)`.
- Two `TestVisualPrompter(BaseTester)` classes now exist — this file (from #3671, batched images + box prompts) and `tests/models/test_prompter.py` (keypoints, broad validation, dynamo). They are not fully redundant, but under the duplication-audit protocol the pair deserves a written disposition at some point. Not this PR's job.

Approving. #4151 is the only thing I'd want recorded before merge, and that's a description edit.

ducha-aiki:
Applied the review feedback in d4a984a9:

- restored the lightweight MobileSAM smoke test to the default device-parametrized suite
- replaced the dynamic fake encoder type with `SimpleNamespace`
- replaced the vacuous constructor assertion with an `img_size` assertion
- updated the PR description to link #4151 and its separate fix in #4164, and clarified that the SAM3 timing win primarily targets the hosted runner

I kept the heavier SAM3 doctest optimization out of this PR as separate scope. Local validation passed on CPU and MPS, including the slow CPU batching path, and `pre-commit run --all-files` is clean.

PR #4164, opened 2026-09-02
fix(augmentation): align probability masks with input devices

### LINKED ISSUE #4151: [Bug]: augmentations raise a device-mismatch RuntimeError on MPS/CUDA since #3722 (449 of 457 MPS augmentation-suite failures)

ducha-aiki:
## Summary

- move each expanded augmentation probability mask onto the corresponding output tensor's device immediately before the four `torch.where` blend sites
- preserve CPU-side RNG and parameter generation, so callers can continue using `augmentation(x_accelerator)` without first moving the augmentation module
- retain the ONNX/fullgraph-friendly branchless blends introduced in #3722
- add deterministic device-parametrized regressions for ordinary 2D, 2D geometric-matrix, 3D geometric-matrix, and mix augmentations

Closes #4151.

## Why

Augmentations generate `batch_prob` on CPU by default, while transformed outputs and matrices follow the input device. Since #3722, four non-scalar `torch.where` calls used that CPU condition directly with MPS/CUDA branches, raising a device-mismatch `RuntimeError` unless users explicitly moved the augmentation module first.

This accounts for 449 of the 457 failures in the MPS augmentation suite measured in #4151. The same path was also exposed downstream by `VisualPrompter` while testing untrained MobileSAM in #4163.

Moving only the boolean selection mask at the blend boundary fixes the shared contract without moving random parameters wholesale, changing RNG streams, reverting to data-dependent Python branches, or altering numerical results.

## Verification

- `tests/augmentation/test_base.py --device=cpu --dtype=float32`: 34 passed
- four new tests on MPS with CPU fallback unset: 4 passed in 0.66 seconds
- affected ONNX export/numerical tests (`RandomInvert`, `RandomHorizontalFlip`, and `RandomAffine`): 7 passed
- `ty check kornia`: passed
- `pre-commit run --all-files`: passed
- independent code review: no findings

The local MPS run was intentionally restricted to the four tiny regression cases after a broader MPS process caused severe memory pressure on the development laptop. Hosted CI provides the broader matrix.

## Merge ordering with #4160

#4160 currently records these 449 failures in its strict MPS baseline. Merge this fix first, then refresh #4160 and remove the recovered entries; otherwise the strict XPASS contract will correctly make its MPS job red.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* Fix 2-D, 3-D, and mix augmentations raising a cross-device `RuntimeError` when their random parameters are generated
+  on the default CPU device and their input is on MPS or CUDA (#4164). Since #3722, four branchless `torch.where`
+  blends used the CPU `batch_prob` mask directly with accelerator outputs or transformation matrices. The selection
+  mask now moves to the output device at each blend boundary, preserving CPU random-number generation and the
+  ONNX/fullgraph-friendly blend. Calling `augmentation(x_accelerator)` therefore works again without first moving the
+  augmentation module or its RNG to the accelerator (#4151).
+
 * Keep `extract_patches_simple` and `extract_patches_from_pyramid` off a broken reduced-precision CPU kernel: for a
   `float16`/`bfloat16` CPU image, the `grid_sample` kernels in torch <= 2.9 read out of bounds when a rotated patch
   straddles the image border and return NaN or garbage values far outside the image's range. Both extractors now
--- a/kornia/augmentation/_2d/base.py
+++ b/kornia/augmentation/_2d/base.py
@@ -135,7 +135,9 @@ class RigidAffineAugmentationBase2D(AugmentationBase2D):
         # If batch sizes line up, do the where-blend. Otherwise (e.g. VideoSequential
         # passes B-sized batch_prob into a B*T-sized input) fall back to all-or-nothing.
         if trans_matrix_applied.shape[0] == to_apply.shape[0] == trans_matrix_identity.shape[0]:
-            to_apply_expanded = to_apply.view(-1, *([1] * (trans_matrix_applied.dim() - 1)))
+            to_apply_expanded = to_apply.view(-1, *([1] * (trans_matrix_applied.dim() - 1))).to(
+                trans_matrix_applied.device
+            )
             trans_matrix = torch.where(to_apply_expanded, trans_matrix_applied, trans_matrix_identity)
         else:
             trans_matrix = trans_matrix_applied if bool(to_apply.any()) else trans_matrix_identity
--- a/kornia/augmentation/_2d/mix/base.py
+++ b/kornia/augmentation/_2d/mix/base.py
@@ -109,7 +109,7 @@ class MixAugmentationBaseV2(_BasicAugmentationBase):
         applied_post = self.apply_non_transform(applied, params, flags)
 
         if applied_post.shape == non_applied.shape and applied_post.shape[0] == to_apply.shape[0]:
-            to_apply_expanded = to_apply.view(-1, *([1] * (applied_post.dim() - 1)))
+            to_apply_expanded = to_apply.view(-1, *([1] * (applied_post.dim() - 1))).to(applied_post.device)
             output = torch.where(to_apply_expanded, applied_post, non_applied)
         else:
             # Shape-changing mix augmentations (e.g. RandomMosaic when ``output_size``
--- a/kornia/augmentation/_3d/base.py
+++ b/kornia/augmentation/_3d/base.py
@@ -101,7 +101,9 @@ class RigidAffineAugmentationBase3D(AugmentationBase3D):
         trans_matrix_identity = self.identity_matrix(in_tensor)
 
         if trans_matrix_applied.shape[0] == to_apply.shape[0] == trans_matrix_identity.shape[0]:
-            to_apply_expanded = to_apply.view(-1, *([1] * (trans_matrix_applied.dim() - 1)))
+            to_apply_expanded = to_apply.view(-1, *([1] * (trans_matrix_applied.dim() - 1))).to(
+                trans_matrix_applied.device
+            )
             trans_matrix = torch.where(to_apply_expanded, trans_matrix_applied, trans_matrix_identity)
         else:
             trans_matrix = trans_matrix_applied if bool(to_apply.any()) else trans_matrix_identity
--- a/kornia/augmentation/base.py
+++ b/kornia/augmentation/base.py
@@ -356,7 +356,7 @@ class _AugmentationBase(_BasicAugmentationBase):
         not onnx-exportable.
         """
         if transformed.shape == not_transformed.shape and transformed.shape[0] == to_apply.shape[0]:
-            to_apply_expanded = to_apply.view(-1, *([1] * (len(transformed.shape) - 1)))
+            to_apply_expanded = to_apply.view(-1, *([1] * (len(transformed.shape) - 1))).to(transformed.device)
             return torch.where(to_apply_expanded, transformed, not_transformed)
         return transformed if bool(to_apply.any()) else not_transformed
 
--- a/tests/augmentation/test_base.py
+++ b/tests/augmentation/test_base.py
@@ -22,8 +22,12 @@ import torch
 
 from kornia.augmentation._2d.base import AugmentationBase2D
 from kornia.augmentation._2d.geometric.affine import RandomAffine
+from kornia.augmentation._2d.geometric.horizontal_flip import RandomHorizontalFlip
 from kornia.augmentation._2d.intensity.gaussian_blur import RandomGaussianBlur
+from kornia.augmentation._2d.intensity.invert import RandomInvert
+from kornia.augmentation._2d.mix.mixup import RandomMixUpV2
 from kornia.augmentation._3d.geometric.affine import RandomAffine3D
+from kornia.augmentation._3d.geometric.horizontal_flip import RandomHorizontalFlip3D
 from kornia.augmentation._3d.intensity.motion_blur import RandomMotionBlur3D
 from kornia.augmentation.base import _BasicAugmentationBase
 from kornia.core._compat import torch_version_lt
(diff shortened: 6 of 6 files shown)


ducha-aiki:
## Review: approve, with test nits and one optional cleanup

The fix is minimal and correct. Every non-scalar `torch.where` in `kornia/augmentation` that consumes `batch_prob`/`to_apply` is one of the four sites touched here. All other consumers (`inverse_*` boolean indexing, `ops.py` per-sample `batch_prob[i]`, `RandomTransplantation`'s `torch.where(...)[0]`, `auto/operations/base.py` which already does `.to(input.device)`) tolerate a CPU mask on accelerator tensors. The scalar `bool(to_apply.any())` fallbacks stay on CPU. A same-device `.to()` returns `self`, so the CPU path is byte-identical to before.

### Verified locally on CUDA (RTX 4090)

This is not an MPS-only bug. The whole augmentation suite on CUDA is red on `main` and green on this branch:

| revision | `KORNIA_TEST_DEVICE=cuda KORNIA_TEST_DTYPE=float32 tests/augmentation` |
|---|---|
| `main` (d7d6a2f6) | 459 failed, 1834 passed |
| this PR (b3f03516) | 0 failed, 2297 passed |

All 459 are the same `Expected all tensors to be on the same device` error at one of the four blend sites. Nothing in the hosted matrix runs on an accelerator, which is why PR CI never saw it. The gap that let #4151 happen is CUDA/MPS CI, not test count.

Other checks on this branch:

- The four new tests fail on `main` and pass here on CUDA (float32/float64). The 2D, 2D-matrix and 3D-matrix ones also pass in float16 and bfloat16 under `--isolate-half-precision`.
- ~70 single augmentations with `p=0.5`, CPU-constructed, CUDA input: no device errors. Same for `AugmentationSequential` with input/mask/boxes/keypoints/class plus `.inverse`, `VideoSequential` (both `same_on_frame` modes), `PatchSequential`, `ImageSequential`, RandAugment/AutoAugment/TrivialAugment, and params replay.
- `torch.compile(fullgraph=True)` with CPU RNG and CUDA input: one graph, zero breaks, the copy shows up as a single `to` node. Inductor `test_dynamo` cases and `test_torch_export.py` / `test_onnx_export.py` pass.
- Mix augmentations have no `p == 1.0` short-circuit (unlike `_AugmentationBase.apply_func`), so on `main` even `RandomMixUpV2(p=1.0)(x_cuda)` raised. This PR fixes that too; worth a line in the description.

### Test nits (non-blocking)

**1. The mix test is red in half precision, for pre-existing reasons.** `test_mix_augmentation_blends_cpu_params_with_accelerator_input` fails in float16 and bfloat16 on both CPU and CUDA with `AssertionError: The values for attribute 'dtype' do not match: torch.float32 != torch.float16`. `RandomMixUpV2` samples `mixup_lambdas` in float32 and `mixup.py:113` moves it to the input device but not the input dtype, so the blend promotes. This predates the PR (the existing `TestRandomMixUpV2` half cases fail the same way), and CI has no half job, but `pixi run test-half` and `pixi run -e cuda test-cuda-half` will report four new failures. Since the test is about device placement, I'd restrict it to float32/float64 or `xfail` half dtypes with a reason. Separately, `RandomMixUpV2(p=1.0)(x_bf16)` without explicit params raises `KeyError: 'BFLOAT16'` at `mix/base.py:229` because `DType` has no bfloat16 member. Both deserve a follow-up issue; happy to file it.

**2. The mix test can't detect a wrong blend.** With `lambda_val=(0.0, 0.0)` the transformed branch equals the input, so `applied_post == non_applied` and `torch.where` returns the input regardless of the mask. It catches the device error (the transformed branch is computed on the accelerator) but not an inverted or misaligned mask at `mix/base.py:112`. A nonzero fixed lambda with an asymmetric expected tensor, like the 2D/3D tests use, would pin per-sample selection.

**3. On the CPU fixture all four tests are tautological.** `torch.tensor([True, False])` is already on the input device, so they pass on `main` too; they only guard the fix when someone runs with `KORNIA_TEST_DEVICE=cuda`/`mps`. That's fine given there's no accelerator CI, but worth saying in a comment so nobody reads them as CI coverage.

### Optional cleanup, here or as a follow-up

The reason this fix had to land in four places is that `_2d/base.py:137`, `_3d/base.py:103` and `_2d/mix/base.py:111` are hand-rolled copies of `_AugmentationBase._blend_by_prob` (same shape guard, same view/where, same `any()` fallback). #3779 already had to fan out `torch.atleast_1d` across the same sites. Hoisting the staticmethod to `_BasicAugmentationBase` (so `MixAugmentationBaseV2` can reach it) and routing the three clones through it is a drop-in: verified bitwise-identical outputs and `transform_matrix` on CUDA for HFlip, Affine, Affine3D, MixUpV2, CutMixV2 and Mosaic, with the whole augmentation suite still green on CPU and CUDA. Net diff would be roughly 19 insertions, 38 deletions across the same four files. A one-line note on `_blend_by_prob` stating that `to_apply` may live on another device and the blend owns the move would also stop the next contributor from re-introducing this.

### Design note

Each `p < 1` blend now does one small pageable host-to-device copy (about 27 µs, a sync point). For `AugmentationSequential` with 4 augmentations over 4 data keys that's 37 vs 19 copies per step compared to `seq.cuda()`, but wall time is unchanged within noise (11.31 ms both ways) because the parameter generators already do 9 to 19 such copies. Generating `batch_prob` on the input device would change RNG streams for every seed, and a single move in `forward()` wouldn't cover the container path that calls `transform_masks/boxes/keypoints` directly, so I agree with the narrow fix. One consequence to be aware of: the blend is no longer sync-free, so `torch.cuda.graph` capture of a `p<1` augmentation with CPU RNG will reject it.

Merge-ordering note with #4160 is correct.

Written by Claude (Fable 5.1) on behalf of @ducha-aiki

ducha-aiki:
Addressed the review findings in 6df2e118:

- The MixUp regression now uses a fixed nonzero lambda and a forced non-identity pairing, with an asymmetric expected result that detects inverted or misaligned batch selection.
- The MixUp case skips float16/bfloat16 with the existing float32-lambda promotion reason, avoiding new failures in the CPU and isolated CUDA half suites.
- The test class now states that CPU runs are smoke coverage and CUDA/MPS exercise the cross-device regression.

Verification:

- `pixi run test-module tests/augmentation/test_base.py --device=cpu,cuda --dtype=float32,float64` — 119 passed
- focused MixUp half run on CPU/CUDA with isolation — 4 skipped as intended
- `pixi run pre-commit-all` — all hooks passed
- mutation check: inverting the MixUp blend mask makes the strengthened test fail on all 24 elements

The optional blend-helper deduplication is intentionally left for a separate refactor so this follow-up stays limited to the concrete test findings.

Issue #3770, opened 2026-07-10
[Feature]: Complete SmolVLM2 implementation (native PyTorch)


abtonmoy:
### 🚀 Feature Description

kornia/models/smolvlm2 is still the placeholder scaffold from #3475 -- its forward is vision_proj(image) + text_proj(text), but it is exported as public API (kornia.models.smolvlm2.SmolVLM2). This proposes completing it into a working native PyTorch SmolVLM2: SigLIP vision tower + pixel-shuffle connector + SmolLM2 (Llama architecture) text decoder.

### 📂 Feature Category

VLM/VLA Models (Vision Language Models/Agents) - Priority

### 💡 Motivation

Kornia is building out a native VLM/VLA stack (#3412 SigLIP2, #3410 Qwen2.5-VL, #3471 PaliGemma, #3448 KimiVL). SmolVLM2's vision tower already exists in Kornia (kornia.models.siglip2), so completing the model is connector + decoder work rather than a new encoder port. The current public API being a non-functional stub is misleading for users.

### 💭 Proposed Solution

Follow the PaliGemma approach (#3471): reuse SigLip2VisionModel as the vision tower, implement the rest in pure torch.nn:
- pixel-shuffle connector (space-to-depth + linear projection to text space)
- SmolLM2 decoder (Llama architecture: RMSNorm, RoPE, GQA, SwiGLU)
- token merging: image features replace <image> placeholder embeddings
- config defaults from the HuggingFaceTB/SmolVLM2-2.2B-Instruct checkpoint config.json
- tests following tests/models/test_siglip2.py, plus docs/source/models.rst entries

References: transformers/models/smolvlm/modeling_smolvlm.py (connector, merger), transformers/models/llama/modeling_llama.py (decoder).

Scope: fixed-resolution path. Weight loading and NaViT variable-resolution support are follow-ups. The decoder uses HF-compatible module names (a Llama state_dict loads with strict=True), so weight mapping later is trivial.

### 🔄 Alternatives Considered

Wrapping HF transformers -- conflicts with the lightweight pure PyTorch goal (same reasoning as #3410).

### 📝 Additional Context

I have this working locally with tests passing (pixi run test, float32/float64): pixel-shuffle matches the HF implementation exactly, and the decoder is bit-identical to LlamaModel with the same weights. Can open a PR once approved and assigned.

### 🤝 Contribution Intent

- [x] I plan to submit a PR to implement this feature
- [ ] I'm requesting this feature but not planning to implement it

[...]

Issue #3471, opened 2026-01-08
[Feature]: Implement PaliGemma Vision-Language Model (Native PyTorch)


mandeepsingh2007:
### 🚀 Feature Description

I propose implementing PaliGemma, a lightweight and powerful open Vision-Language Model (VLM) inspired by PaLI-3. This implementation will be a native PyTorch version within Kornia, avoiding external heavy dependencies like transformers.

PaliGemma combines:
Vision Encoder: SigLIP (specifically the SigLip2 variant, which serves as the backbone).
Language Decoder: Gemma (a decoder-only generic LLM).
This feature will allow Kornia users to perform multimodal tasks (image + text $\to$ text) using official pre-trained weights.

### 📂 Feature Category

VLM/VLA Models (Vision Language Models/Agents) - Priority

### 💡 Motivation

Kornia currently lacks a native Vision-Language Model (VLM) pipeline, forcing users to rely on heavy external dependencies (like transformers) for multimodal tasks.

Implementing PaliGemma fills this gap by extending the existing SigLip2 backbone into a full VLM. This enables advanced tasks like Visual Question Answering (VQA) and captioning directly within Kornia, ensuring a unified, lightweight, and pure PyTorch workflow.

### 💭 Proposed Solution

I plan to implement the architecture in three stages, following a "pure PyTorch" approach:

Vision Backbone (SigLip2 Extension): Reuse and extend the existing SigLip2 implementation in Kornia to serve as the vision encoder.

Language Decoder (Gemma): Implement the Gemma decoder architecture from scratch using torch.nn modules (Attention, MLP, RMSNorm) to ensure no dependency on Hugging Face transformers.

Weight Loading Strategy: Create a mapping utility to load official pre-trained weights (from Google/HF) into the native Kornia model structure.

### 🎯 Use Cases

Visual Question Answering (VQA): Robots or agents analyzing a scene (e.g., "Is the safety valve open?").

Image Captioning: Automated reporting for surveillance or logging systems.

Object Detection/Segmentation: Generating bounding box coordinates or segmentation masks via text tokens.

Robotics: End-to-end vision-language planning where preprocessing speed is critical.

### 🤝 Contribution Intent

- [x] I plan to submit a PR to implement this feature
- [ ] I'm requesting this feature but not planning to implement it

edgarriba:
@mandeepsingh2007 make sure you make iteratively PRs

mandeepsingh2007:
[...]

PR #4165, opened 2026-09-02
feat(models): complete SmolVLM2 implementation (native PyTorch)

abtonmoy:
## 📝 Description

**Fixes/Relates to:** #3770

Replaces the non-functional SmolVLM2 placeholder scaffold (from #3475) with a complete, working native PyTorch implementation: SigLIP vision tower + pixel-shuffle connector + SmolLM2 (Llama-architecture) text decoder, following the PaliGemma approach (#3471).

Maintainer approval: @ducha-aiki approved and assigned in [this comment](https://github.com/kornia/kornia/issues/3770#issuecomment-5231415451); @edgarriba requested byte-to-byte benchmarks against the HF reference in [this comment](https://github.com/kornia/kornia/issues/3770#issuecomment-4968282508) — provided below (all comparisons bit-identical).

Scope per the issue: fixed-resolution path. Weight loading and NaViT variable-resolution support are follow-ups. The decoder uses HF-compatible module names, so a Llama `state_dict` loads with `strict=True` (verified below).

---

## 🛠️ Changes Made
- [x] `kornia/models/smolvlm2/configuration_smolvlm2.py` — `SmolVLM2Config` / `SmolVLM2TextConfig` dataclasses with defaults matching the `HuggingFaceTB/SmolVLM2-2.2B-Instruct` checkpoint `config.json` (vision tower 27 layers / 1152 hidden / patch 14 / image 384; decoder 24 layers / 2048 hidden / 32 heads / SwiGLU 8192 / `rope_theta=130000`; `scale_factor=3`, `image_token_id=49190`)
- [x] `kornia/models/smolvlm2/modeling_smolvlm2.py` —
  - `SmolVLM2Connector`: pixel-shuffle (space-to-depth) + linear projection to text space, mirroring HF `SmolVLMConnector`
  - `SmolVLM2TextModel`: SmolLM2/Llama decoder (RMSNorm, RoPE, GQA with SDPA, SwiGLU), HF-compatible parameter names
  - `SmolVLM2Model`: vision tower (reuses `SigLip2VisionModel`) + connector + token merging at `<image>` placeholder positions
  - `SmolVLM2ForConditionalGeneration`: adds LM head; `SmolVLM2` kept as alias for the previous public symbol
- [x] Removed the placeholder `smolvlm2.py` scaffold; updated `__init__.py` exports
- [x] `tests/models/test_smolvlm2.py` — 43 tests following `tests/models/test_siglip2.py` (smoke, cardinality, exceptions, causality, padding-mask consistency, RoPE identity at position 0, hardcoded HF pixel-shuffle reference values, inputs merger, text-only path, gradcheck, dynamo)
- [x] `docs/source/models.rst` — SmolVLM2 section with autoclass directives and usage example
- [x] `CHANGELOG.md` — entry under Unreleased documenting the placeholder replacement and new public API

---

## 🧪 How Was This Tested?

- [x] **Unit Tests:** 41 tests × {float16, bfloat16, float32, float64} on CPU; 43 tests (incl. `test_dynamo` with inductor) × {float32, float64} on CUDA (NVIDIA L4, torch 2.9.1+cu128)
- [x] **Manual Verification:** byte-to-byte comparison against HF `transformers` 5.0.0rc3 with identical weights (`strict=True` state-dict transfer both ways) — every comparison bit-identical (`max_abs_diff = 0.0`); script and output below
- [x] **Performance/Edge Cases:** text-only inputs (no `pixel_values`); mismatched `<image>`-token/feature counts raise `ValueError`; padded batches produce no NaNs (fully-masked query rows attend to self); all-ones mask reproduces the no-mask output exactly

<details><summary>CPU test log (float32/float64)</summary>

```
platform win32 -- Python 3.11.14, pytest-9.0.3 -- torch-2.9.1+cpu
tests/models/test_smolvlm2.py .........................................
============================= 41 passed in 3.28s ==============================
```

</details>

<details><summary>CUDA test log (NVIDIA L4, incl. torch.compile/inductor)</summary>

```
platform linux -- Python 3.11.12, pytest-9.1.1 -- torch-2.9.1+cu128, CUDA 12.8, NVIDIA L4
tests/models/test_smolvlm2.py::TestSmolVLM2Model::test_dynamo[cuda-float32-inductor] PASSED
tests/models/test_smolvlm2.py::TestSmolVLM2Model::test_dynamo[cuda-float64-inductor] PASSED
======================== 43 passed, 1 warning in 35.85s ========================
```

</details>

<details><summary>Byte-to-byte verification vs HF transformers — output (requested in #3770)</summary>

```
connector.pixel_shuffle: max_abs_diff=0.000e+00 bit_identical=True
connector.forward: max_abs_diff=0.000e+00 bit_identical=True
decoder[MHA 32/32 (checkpoint cfg, 4 layers)].causal_no_mask: max_abs_diff=0.000e+00 bit_identical=True
decoder[MHA 32/32 (checkpoint cfg, 4 layers)].padding_mask(valid rows): max_abs_diff=0.000e+00 bit_identical=True
decoder[GQA 32/8 (4 layers)].causal_no_mask: max_abs_diff=0.000e+00 bit_identical=True
decoder[GQA 32/8 (4 layers)].padding_mask(valid rows): max_abs_diff=0.000e+00 bit_identical=True
transformers version used as reference: 5.0.0rc3 | torch: 2.9.1
ALL COMPARISONS PASSED
```

</details>

<details><summary>Byte-to-byte verification script (reproducible)</summary>

```python
"""Byte-to-byte verification of kornia SmolVLM2 vs HuggingFace transformers references."""
import torch

torch.manual_seed(0)

def report(name, a, b):
    print(f"{name}: max_abs_diff={(a - b).abs().max().item():.3e} bit_identical={torch.equal(a, b)}")

# 1. Connector vs transformers SmolVLMConnector (real checkpoint dims)
from transformers.models.smolvlm.configuration_smolvlm import SmolVLMConfig
from transformers.models.smolvlm.modeling_smolvlm import SmolVLMConnector
from kornia.models.smolvlm2 import SmolVLM2Config, SmolVLM2Connector, SmolVLM2TextConfig
from kornia.models.smolvlm2.modeling_smolvlm2 import SmolVLM2TextModel

k_conn = SmolVLM2Connector(SmolVLM2Config()).eval()
hf_conn = SmolVLMConnector(SmolVLMConfig(
    vision_config={"hidden_size": 1152, "patch_size": 14, "image_size": 384},
    text_config={"hidden_size": 2048}, scale_factor=3,
)).eval()
hf_conn.load_state_dict(k_conn.state_dict(), strict=True)

x = torch.randn(2, 729, 1152)  # 27x27 patch grid from a 384x384 image
with torch.no_grad():
    report("connector.pixel_shuffle", k_conn.pixel_shuffle(x, 3), hf_conn.pixel_shuffle(x, 3))
    report("connector.forward", k_conn(x), hf_conn(x))

# 2. Text decoder vs transformers LlamaModel (strict=True state-dict transfer)
from transformers.models.llama.configuration_llama import LlamaConfig
from transformers.models.llama.modeling_llama import LlamaModel

def compare_decoder(tag, num_kv_heads):
    torch.manual_seed(1)
    k_cfg = SmolVLM2TextConfig(num_hidden_layers=4, num_key_value_heads=num_kv_heads)
    k_model = SmolVLM2TextModel(k_cfg).eval()
    hf_model = LlamaModel(LlamaConfig(
        vocab_size=k_cfg.vocab_size, hidden_size=k_cfg.hidden_size,
        intermediate_size=k_cfg.intermediate_size, num_hidden_layers=k_cfg.num_hidden_layers,
        num_attention_heads=k_cfg.num_attention_heads, num_key_value_heads=k_cfg.num_key_value_heads,
        head_dim=k_cfg.head_dim, max_position_embeddings=k_cfg.max_position_embeddings,
        rope_theta=k_cfg.rope_theta, rms_norm_eps=k_cfg.rms_norm_eps, pad_token_id=k_cfg.pad_token_id,
        attention_bias=False, mlp_bias=False, attn_implementation="sdpa",
    )).eval()
    hf_model.load_state_dict(k_model.state_dict(), strict=True)

    input_ids = torch.randint(3, k_cfg.vocab_size, (2, 16))
    with torch.no_grad():
        report(f"decoder[{tag}].causal_no_mask", k_model(input_ids), hf_model(input_ids).last_hidden_state)
    mask = torch.ones(2, 16, dtype=torch.long); mask[1, 10:] = 0
    with torch.no_grad():
        k_out = k_model(input_ids, attention_mask=mask)
        hf_out = hf_model(input_ids, attention_mask=mask).last_hidden_state
    valid = mask.bool()
    report(f"decoder[{tag}].padding_mask(valid rows)", k_out[valid], hf_out[valid])

compare_decoder("MHA 32/32 (checkpoint cfg, 4 layers)", num_kv_heads=32)
compare_decoder("GQA 32/8 (4 layers)", num_kv_heads=8)
```

</details>

**Reference implementations:** `transformers/models/smolvlm/modeling_smolvlm.py` (connector, inputs merger), `transformers/models/llama/modeling_llama.py` (decoder).

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**

---

## 🚦 Checklist

---

## 💭 Additional Context
The checkpoint's vision activation is `gelu_pytorch_tanh` while Kornia's `SigLip2VisionModel` uses exact `nn.GELU`; the same known approximation difference applies to PaliGemma, which shares the tower (noted in the config docstring). Follow-ups per the issue: pretrained weight loading (trivial given the strict-load compatibility shown above) and NaViT variable-resolution support.


ducha-aiki:
## Triage note on #4165 — queued for a full review

@abtonmoy — this is a large piece of work (+960/−47, a new model family with a byte-identity claim against HF `transformers`), so it needs a dedicated review round rather than a pass inside a triage sweep. That is scheduled; this note is just so you are not waiting in silence, plus the mechanical checks I did run today.

Setup: worktree at the PR head, `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU/macOS.

```
import kornia                                             -> resolves to the worktree, imports clean
kornia.models.smolvlm2.__all__                            -> SmolVLM2, SmolVLM2Config, SmolVLM2Connector,
                                                             SmolVLM2ForConditionalGeneration, SmolVLM2Model,
                                                             SmolVLM2TextConfig
pytest tests/models/test_smolvlm2.py --dtype=float32,float64   -> 41 passed in 0.40s
pytest tests/test_api_surface.py                               -> 12 passed
```

`test_api_surface.py` passing is the one I most wanted to see, given that the PR removes the placeholder scaffold: the stable-core inventory is intact and `SmolVLM2` survives as an alias, so nothing public disappears. The test module is also genuinely cheap (0.4 s for 41 tests), which matters — `tests/models` is on the critical path of several CI legs and I have an open PR trimming SigLIP2's runtime; please keep the config dims small as they are.

Confirmed the lane is clean: @edgarriba's *"feel free to go and please provide byte to byte benchmarks"* and my own go-ahead are both on #3770, and you are assigned there.

What the real review will go after, so you can pre-empt it: re-running your HF byte-identity comparison independently, the padded/fully-masked attention rows, the `<image>`-token merge path, and whether the decoder's HF-compatible parameter names really give `strict=True` on a Llama `state_dict`. No action needed from you right now.

ducha-aiki:
Heads-up while this waits for its full review round: the branch went `CONFLICTING` after today's docs merges (#4155 and #4178 landed the site revamp). The only conflict is **`docs/source/models.rst`** — `git merge-tree` against `main` @ `ee6b713f` reports nothing else, so it is a small resolve. Please rebase; the model page layout on `main` has changed shape, so take `main`'s structure and re-add your SmolVLM2 entry into it rather than keeping your side of the hunk.

The substantive review (byte-identity against the reference implementation, padded/masked attention, the `<image>` token merge, and `strict=True` state-dict loading) is still queued and unaffected by this.

ducha-aiki (changes requested):
Review at `d3196cb5`.

[P1] Disable autocast inside RoPE frequency construction (`kornia/models/smolvlm2/modeling_smolvlm2.py:149-153`). Calling .float() on both operands does not protect their matmul from autocast. At dim=64/base=130000, positions [1000,8191], CPU bfloat16 autocast changes the cosine by up to 1.894267 and sine by 1.939919. The reference HF LlamaRotaryEmbedding explicitly disables autocast around this calculation and stays exactly equal to float32. With identical strict-loaded weights, a tiny two-layer GQA decoder is exact in plain float32 but differs from HF by up to 0.142 under bfloat16 autocast near position 8191. This materially changes long-context attention during mixed-precision execution.

Please protect the frequency matmul and trigonometric operations with autocast disabled, cast the result to x.dtype afterward, and add a high-position autocast regression. Minimal reproduction:

```python
r = SmolVLM2RotaryEmbedding(64, base=130000.0)
x = torch.zeros(1, 1, 64)
p = torch.tensor([[1000, 8191]])
expected = r(x, p)
with torch.autocast("cpu", dtype=torch.bfloat16):
    actual = r(x, p)
print((expected[0] - actual[0]).abs().max())  # 1.894267
```

The reference comparison otherwise checks out: strict decoder state_dict loading succeeds in both directions; connector keys and forward match exactly; decoder no-mask/all-ones outputs and valid-token rows under left/right padding match exactly. Padded-row outputs differ (but remain finite), so qualify the CHANGELOG bit-identity claim to valid-token rows. Validation used tiny configurations, torch 2.14.0 and installed transformers 5.5.4; no pretrained weights were downloaded. The PR test file passes 41 tests. Current main also requires a rebase; the earlier triage had already requested one for the model docs.

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (2062884b) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4165.breaking.md`.

`docs/source/models.rst` also conflicted because `main` rewrote that page: the merge keeps the rewrite and re-inserts your SmolVLM2 section after Kimi-VL, before the closing rule. Please give it a look.

ducha-aiki (changes requested):
Review of `2062884bf7b2bc5a59e12522ff58d0930a07f191`.

The docs merge conflict and changelog format have been addressed, but the substantive previous finding is still present.

[P1] Disable autocast around SmolVLM2RotaryEmbedding.forward frequency construction and trigonometric evaluation. At dim=64/base=130000, positions [[1000,8191]], CPU bfloat16 autocast still changes cosine by 1.894267 and sine by 1.580026 relative to the same call outside autocast. Calling .float() does not protect the matmul. The source in this area is unchanged from the last reviewed head. Add a high-position autocast regression and cast the protected result back to x.dtype.

Also qualify the decoder bit-identity statement in changelog.d/4165.breaking.md to the actually verified valid-token rows; the migration preserved the earlier unqualified claim.

Validation: exact-head reproduction with torch 2.14.0 and verified import origin; the CPU float32 model file passes 21 tests, which do not catch the autocast defect. No pretrained-weight or CUDA run.

victorwon2001:
**kornia/models/smolvlm2/modeling_smolvlm2.py:465** — The SmolVLM2 alias now resolves to SmolVLM2ForConditionalGeneration, whose constructor accepts a SmolVLM2Config, but docs/export_support/cases_models.py still constructs SmolVLM2(16, 16) and still describes the removed two-projection placeholder scaffold.

The models export-support survey no longer exercises the new SmolVLM2 implementation through this case; the stale two-positional-argument construction is incompatible with the new constructor and will surface as a constructor failure instead of a valid model case.

Static signature comparison only; no execution was performed. The new conditional-generation class takes one config argument and SmolVLM2 aliases that class, while the supplied export-survey caller passes two integer positional arguments and labels the target as the old placeholder.

`file:kornia/models/smolvlm2/modeling_smolvlm2.py:diff`

> +        inv_freq = self.inv_freq[None, :, None].float().expand(position_ids.shape[0], -1, 1)
> +        pos = position_ids[:, None, :].float()
> +        freqs = (inv_freq @ pos).transpose(1, 2)
> +        emb = torch.cat((freqs, freqs), dim=-1)
> +        return emb.cos().to(x.dtype), emb.sin().to(x.dtype)

`review:5196181129`

> The docs merge conflict and changelog format have been addressed, but the substantive previous finding is still present.
> 
> [P1] Disable autocast around SmolVLM2RotaryEmbedding.forward frequency construction and trigonometric evaluation. At dim=64/base=130000, positions [[1000,8191]], CPU bfloat16 autocast still changes cosine by 1.894267 and sine by 1.580026 relative to the same call outside autocast. Calling .float() does not protect the matmul. The source in this area is unchanged from the last reviewed head. Add a high-position autocast regression and cast the protected result back to x.dtype.
> 
> Also qualify the decoder bit-identity statement in changelog.d/4165.breaking.md to the actually verified valid-token rows; the migration preserved the earlier unqualified claim.
> 
> Validation: exact-head reproduction with torch 2.14.0 and verified import origin; the CPU float32 model file passes 21 tests, which do not catch the autocast defect. No pretrained-weight or CUDA run.

`file:changelog.d/4165.breaking.md:diff`

> +at `<image>` placeholder positions. New public API in `kornia.models.smolvlm2`: `SmolVLM2Config`,
> +`SmolVLM2TextConfig`, `SmolVLM2Model`, `SmolVLM2ForConditionalGeneration`, and `SmolVLM2Connector`. With the same
> +weights, the decoder is bit-identical to `transformers` `LlamaModel` (a Llama state dict loads with
> +`strict=True`) and the connector to the HF `SmolVLMConnector`; weight loading and NaViT variable resolution are

`review:5125346486`

> The reference comparison otherwise checks out: strict decoder state_dict loading succeeds in both directions; connector keys and forward match exactly; decoder no-mask/all-ones outputs and valid-token rows under left/right padding match exactly. Padded-row outputs differ (but remain finite), so qualify the CHANGELOG bit-identity claim to valid-token rows. Validation used tiny configurations, torch 2.14.0 and installed transformers 5.5.4; no pretrained weights were downloaded. The PR test file passes 41 tests. Current main also requires a rebase; the earlier triage had already requested one for the model docs.

`file:kornia/models/smolvlm2/modeling_smolvlm2.py:diff`

> +class SmolVLM2ForConditionalGeneration(nn.Module):
> +    """SmolVLM2 with a language-modeling head producing vocabulary logits."""
> +
> +    def __init__(self, config: SmolVLM2Config) -> None:
> +        super().__init__()
> +        self.config = config
> +        self.model = SmolVLM2Model(config)
> +        self.lm_head = nn.Linear(config.text_config.hidden_size, config.text_config.vocab_size, bias=False)

`file:kornia/models/smolvlm2/modeling_smolvlm2.py:diff`

> +
> +# Backwards-compatible alias for the previous scaffold symbol.
> +SmolVLM2 = SmolVLM2ForConditionalGeneration

`context:docs/export_support/cases_models.py`

> "models.smolvlm2.SmolVLM2",
>         "models.vlm",
>         lambda: ev(SmolVLM2(16, 16)),
>         lambda m: m,
>         [torch.rand(1, 5, 16), torch.rand(1, 5, 16)],
>         note="placeholder scaffold: sum of two Linear projections; image/text sequences must have equal length",
>         tags=("model",),
>     )

Context truncated at 30000 characters: docs/export_support/cases_models.py

Context truncated at 30000 characters: docs/export_support/cases_models.py

Disable autocast around RoPE frequency construction and trigonometric evaluation and add the requested high-position autocast regression; qualify the changelog bit-identity statement to the verified valid-token behavior; and update docs/export_support/cases_models.py to exercise the new config-based SmolVLM2 API instead of SmolVLM2(16, 16). After those edits, rerun the focused SmolVLM2 tests and the relevant export/half-precision checks.

SHA: `2062884bf7b2bc5a59e12522ff58d0930a07f191`

ducha-aiki:
## Independent verification of the automated review (2026-09-17T00:38Z), plus my own pass

Every claim below was re-executed from scratch on the **merge tree** (`git merge origin/main` @ `8c71df9a0`, result `7d6d5f308`, merges clean), torch 2.14.0, CPU/macOS arm64. Every measurement block printed `kornia.__file__` and asserted it named the worktree. `transformers` was not in the venv; I installed 5.10.1 **offline from the uv cache** into a `--target` overlay — **no weights were downloaded**, all comparisons use tiny randomly-initialised configs.

Regression sweep first, for context:

| | `tests/models` (CPU float32) |
| --- | --- |
| base `8c71df9a0` | 333 passed, 92 skipped (218.6 s) |
| merge tree | 354 passed, 92 skipped (325.9 s) |

Delta `+21` = exactly the new `tests/models/test_smolvlm2.py`. **No newly-red node.**

### Claim 1 — autocast corrupts RoPE — **CONFIRMED, exact numbers reproduced**

```
dim=64, base=130000, position_ids=[[1000, 8191]], CPU bfloat16 autocast
cos max|d| vs the same call outside autocast = 1.8942670822143555
sin max|d| vs the same call outside autocast = 1.5800256729125977
```

Bit-for-bit the reviewer's `1.894267` / `1.580026`. The `.float()` calls do not protect: autocast downcasts the `inv_freq @ pos` matmul operands regardless, so the output dtype stays `float32` while the *value* has been through a bf16 matmul. float16 autocast is also affected (cos `1.095374`, sin `0.430716`).

**The 21 existing tests do not catch it.** I ran the whole file under a forced `torch.autocast("cpu", bfloat16)` autouse fixture: **21 passed**, identical to the unforced run. I proved the harness can go red before trusting that — a probe asserting `torch.is_autocast_enabled("cpu")` fails without the plugin and passes with it. The reason is `test_rope_position_zero_identity`, the only RoPE test, which uses `position_ids=0`, where `cos=1, sin=0` for every frequency.

**The prescribed fix works and is a no-op otherwise.** Wrapping the frequency construction and trig in `torch.autocast(device_type=..., enabled=False)` and casting back to `x.dtype`:

```
                   bf16 autocast        fp16 autocast     no autocast (vs current code)
before cos / sin   1.894267 / 1.580026  1.095374 / 0.430716        —
after  cos / sin   0.0      / 0.0       0.0      / 0.0      0.0e+00 / 0.0e+00
```

Residual after the fix is **exactly 0.0**, and outside autocast the fixed and current implementations are bit-identical, so it cannot regress the default path.

End to end it matters: a 2-layer decoder, `hidden_size=256`, `head_dim=64`, positions 4000-4015, bf16 autocast, relative to the fp32 reference — **2.95%** max relative error with the current code vs **0.14%** with the fix (0.14% is the plain bf16 noise floor; the same run at positions 0-15 gives 0.14% either way). A 21x amplification that only appears in long context.

**One nuance the automated review omitted, which is why I do not grade this P1.** The identical pattern already ships on `main`: `kornia/models/paligemma/modeling_paligemma.py` `GemmaRotaryEmbedding.forward` has the same unguarded `inv_freq @ position_ids` and measures **cos 1.599636 / sin 1.729295** under the same bf16 autocast, on `origin/main` with this PR nowhere in the tree.

- **[P2]** `SmolVLM2RotaryEmbedding.forward` — wrap the frequency matmul and `cos`/`sin` in `torch.autocast(device_type=..., enabled=False)` and cast back to `x.dtype`, matching `transformers`' `LlamaRotaryEmbedding`. Add a regression at a **high** position (0 is degenerate).
- **[P3]** follow-up issue for `GemmaRotaryEmbedding` on `main` — same defect, same fix, out of scope here.

`torch.autocast(..., enabled=False)` exists well below the 2.5.1 floor, so the fix is portable.

### Claim 2 — the export-support survey case is stale — **CONFIRMED (executed), impact larger than stated**

The reviewer said "static signature comparison only; no execution". Executed on the merge tree:

```python
>>> from kornia.models.smolvlm2 import SmolVLM2
>>> SmolVLM2(16, 16)          # docs/export_support/cases_models.py:914
TypeError: SmolVLM2ForConditionalGeneration.__init__() takes 2 positional arguments but 3 were given
```

The same call on `origin/main` succeeds. So the PR does break it.

**Does it run in CI?** No. `grep -rn export_support .github/` returns nothing, and `docs/export_support/README.md` says "nothing here runs during a docs build"; regenerating is a manual `python docs/export_support/run.py`, "~2 h on 8 CPU cores".

**But the committed snapshot is published.** `docs/source/conf.py:109-114` `exec`s `docs/generate_export_support.py` at every docs build, rendering `docs/source/_data/export_support.json` into `get-started/export-support.rst`. Case `[1272]` is:

```json
{"name": "models.smolvlm2.SmolVLM2", "onnx": "ok", "export": "ok", "compile": "ok",
 "graph_breaks": 0, "note": "placeholder scaffold: sum of two Linear projections; ..."}
```

I regenerated the page on the merge tree; line 6919 renders `smolvlm2.SmolVLM2` as **ONNX yes / torch.export yes / torch.compile yes, 0 graph breaks**. Those three "yes" were measured on the deleted 2-Linear scaffold and, after this PR, are published as the support status of a SigLIP-tower + 24-layer Llama VLM. Nobody has measured whether that model exports at all.

- **[P2]** Update `docs/export_support/cases_models.py:912-919` to the config-based API (a tiny config, as `PaliGemma[tiny,random]` two cases above already does) **and** re-measure that one case so the JSON stops publishing the scaffold's numbers. `run.py` takes an `only=` filter, so this is not the 2 h sweep.

### Claim 3 — the changelog bit-identity claim is unqualified — **CONFIRMED, with the cause isolated**

Tiny config (`vocab=100, hidden=32, 2 layers, 4 heads, 2 kv heads, head_dim=8, rope_theta=130000`), weights copied kornia -> HF:

```
lm.load_state_dict(km.state_dict(), strict=True)   -> OK, missing=[] unexpected=[]

                              overall max|d|   valid-token rows   padded rows
no attention_mask               0.000e+00        0.000e+00            —
all-ones mask                   0.000e+00        0.000e+00            —
left padding  [0,0,0,1,1,1,1,1] 0.000e+00        0.000e+00        0.000e+00
right padding [1,1,1,1,1,0,0,0] 1.157e+00        0.000e+00        1.157e+00
```

So the reviewer is right. Left padding happens to agree because every fully-masked row averages identical pad-token `v` vectors.

**Root cause (not in the reviewer's pass).** The divergence is entirely `modeling_smolvlm2.py:336`:

```python
idx = torch.arange(seq_len, device=device)
allowed[:, :, idx, idx] = True
```

A/B on the same right-padded input, kornia vs HF `LlamaModel`:

```
WITH    the diagonal guard:  max|kornia - HF| = 0.42882
WITHOUT the diagonal guard:  max|kornia - HF| = 0.000e+00 , all outputs finite
```

Removing that one line makes the decoder bit-identical to `transformers` on **every** row, padded included.

**And the comment justifying it does not reproduce.** The line is introduced by "Padded query rows would be fully masked and produce NaNs in SDPA". A fully-masked boolean row through `F.scaled_dot_product_attention` returns exactly `0.0`, not NaN, on all three backends I can reach:

| backend | fully-masked row | NaN? |
| --- | --- | --- |
| CPU, torch 2.5.1 (the kornia floor, offline overlay) | `[0.0, 0.0, 0.0, 0.0]` | no |
| CPU, torch 2.14.0 | `[0.0, 0.0, 0.0, 0.0]` | no |
| MPS, torch 2.14.0 (`PYTORCH_ENABLE_MPS_FALLBACK` unset) | `[0.0, 0.0, 0.0, 0.0]` | no |

The full model with the guard stripped stays finite under both left and right padding on CPU.

- **[P2]** Resolve it one way or the other. Either **drop the diagonal guard**, in which case the changelog sentence becomes true as written and the code gets simpler; or **keep it** and narrow the fragment to "bit-identical **on valid (non-padding) token positions**; padded rows differ because kornia lets a fully-masked row attend to itself". If you keep it, the comment must name the backend and torch version where the NaN occurs, since I cannot produce it on any of the three above. (CUDA is unverified here — no GPU.)

### Claim 4 — public-surface contract — **PARTIAL: docs complete, the `api_surface.json` half does not apply**

All five new names carry an `autoclass` in `docs/source/models.rst`: `SmolVLM2Config` (:208), `SmolVLM2TextConfig` (:212), `SmolVLM2Model` (:216), `SmolVLM2ForConditionalGeneration` (:221), `SmolVLM2Connector` (:246), plus a "Key Components" block and a runnable example. **Nothing missing.**

`tests/api_surface.json` tracks exactly 12 keys — `kornia.{augmentation, color, core, enhance, filters, geometry, image, io, losses, metrics, morphology, utils}`. **`kornia.models` is not inventoried**, so there is nothing to add and nothing to acknowledge in `api_surface_removals.json`. The `Check for dropped module-level names` leg passes for the same reason, and `SmolVLM2` survives as an alias so no public name disappears.

Rest of the contract clean: `changelog.d/4165.breaking.md` exists and `breaking` is the **correct type**; zero lines over 120 chars in the three new files; Apache headers present; `from __future__ import annotations` in both new non-JIT modules.

### Claim 5 — the failing CI check — **not caused by this PR**

`gh pr checks 4165` lists two non-green rows; only one is real.

- `half-precision (float16) / tests (3.11, 2.9.1)` (job `103927193807`) — the API `conclusion` is **`cancelled`**, not `failure`; `gh pr checks` renders cancelled as `fail`. Duration `6h00m31s` = the Actions per-job ceiling. Last pytest output is `tests/feature/test_scale_space_detector.py .............s..s..s..s...... [ 57%]` at `10:49:42Z`, then six hours of silence until `##[error]The operation was canceled.`, with `pytest` terminated as an orphan. Runner CPU from the log: `INTEL(R) XEON(R) PLATINUM 8573C`. This is the known Intel-only float16 hang (#4214). `tests/models` sorts after `tests/feature` and was never reached. **Re-run the leg.**
- `collector` — derivative.

Every other check is green, including `dynamo / tests (3.11, 2.14.0)` — and `tests/models` *is* in that leg's directory list (`pr_test_cpu.yml:175`), so the new `test_dynamo` really does run.

### Independent finding — mutation testing: 3 of 12 one-token mutants survive all 21 tests

The changelog stakes the PR on bit-identity with `transformers`, but the test file contains **no numerical reference pin for the decoder at all** — only smoke, cardinality, exception, causality, a position-0 RoPE identity, and one `pixel_shuffle` value pin.

| mutant | result | non-equivalent? |
| --- | --- | --- |
| M1 `_rotate_half`: `cat((-x2, x1))` -> `cat((x2, -x1))` | **21 passed** | **yes** — max abs diff vs HF `LlamaModel` = **0.13239** |
| M4 attention `scaling = head_dim**-0.5` -> `**-1.0` | **21 passed** | **yes** — **0.12181** |
| M11 `inv_freq` exponent `/ dim` -> `/ (dim // 2)` | **21 passed** | **yes** — 0.01600 at `head_dim=8`; at the production `head_dim=64, base=130000, pos=1000` it moves **cos by 1.9330, sin by 1.9194** |
| M2 cos/sin swapped | 1 failed | killed |
| M3 `cos.unsqueeze(1)` -> `(2)` | 8 failed | killed |
| M5 `is_causal` -> `False` | 2 failed | killed |
| M7 `_repeat_kv` n_rep -> 1 | 8 failed | killed |
| M9 `pixel_shuffle` permute axes | 1 failed | killed |
| M10 `inputs_merger` no-op | 1 failed | killed |
| M6 drop RMSNorm `weight` factor | 21 passed | **equivalent** — `nn.Parameter(torch.ones(dim))`, and no test ever sets a non-unit norm weight (which is exactly what a loaded checkpoint has) |
| M8 drop the diagonal guard | 21 passed | **equivalent on CPU 2.5.1 / 2.14 and MPS 2.14** — see Claim 3 |
| M12 RMSNorm default `eps` 1e-5 -> 1e-1 | 21 passed | **equivalent** — `eps=` is always passed explicitly |

Baseline sanity: the unmutated decoder is `0.000e+00` from HF `LlamaModel` on the same input, so these three are real divergences, not tolerance noise.

- **[P2]** M1, M4 and M11 are surviving non-equivalent mutants, all inside the RoPE/attention numerics the changelog calls bit-identical. **One** hardcoded-literal reference pin kills all three: `SmolVLM2TextModel(tiny_config)` under a fixed seed, a fixed `input_ids` at **non-zero positions**, `assert_close` against a literal produced by HF `LlamaModel` with the same state dict — with the generation snippet and the `transformers` version in a comment, per AGENTS.md. A second short pin on the RoPE `cos`/`sin` at a high position covers Claim 1's regression at the same time.
- **[P3]** No test uses a non-unit `SmolVLM2RMSNorm.weight` (M6). Setting the norm weights to something other than ones in the reference pin closes this for free.

### Summary

- Claim 1 **CONFIRMED** — graded **[P2]**, not P1, because the identical pattern already ships on `main` in PaliGemma.
- Claim 2 **CONFIRMED by execution**, and the stale row is **published**, not merely dormant — **[P2]**.
- Claim 3 **CONFIRMED**, cause isolated to `modeling_smolvlm2.py:336` — **[P2]**.
- Claim 4 **PARTIAL**: docs complete; `api_surface.json` does not track `kornia.models`, so that half is void. No contract violation.
- Claim 5: the one non-green leg is a **cancelled** 6-hour float16 hang on an Intel Xeon runner (#4214), unrelated.
- Independent: 3 surviving non-equivalent mutants in the decoder numerics — **[P2]**.

The implementation holds up well: `strict=True` state-dict interchange with HF works in both directions, valid-token outputs are bit-identical, the connector value pin is correct and kills its mutant, and the PR adds no newly-red node to `tests/models`. What is missing is the numerical pin that would make the bit-identity claim self-enforcing, the autocast guard, an accurate export-support row, and a fragment sentence that matches what the code does on padded rows.

victorwon2001:
One new current defect was identified. SmolVLM2Config exposes a top-level pad_token_id, but full-model construction passes only text_config into SmolVLM2TextModel, whose embedding uses text_config.pad_token_id. The two values can therefore disagree and the documented top-level setting does not control decoder padding behavior. Previously discussed RoPE autocast, export-survey, padded-row bit-identity, and numerical-reference-test issues are not re-reported as findings.

**kornia/models/smolvlm2/configuration_smolvlm2.py:114** — SmolVLM2Config.pad_token_id is unsynchronized with the pad token actually used by the decoder. SmolVLM2Config stores both text_config and its own pad_token_id, but SmolVLM2Model constructs SmolVLM2TextModel from config.text_config, and that text model sets nn.Embedding.padding_idx from text_config.pad_token_id. Changing only the documented top-level pad_token_id therefore has no effect on the decoder.

A custom configuration can advertise one padding token through SmolVLM2Config.pad_token_id while the decoder treats a different token as its embedding padding index, giving callers inconsistent padding semantics. The supplied test fixture already demonstrates that the two fields may diverge by setting text_config.pad_token_id to 0 while leaving the top-level value at its default 2.

Static data-flow inspection only; no execution was performed. The full model forwards config.text_config into SmolVLM2TextModel, whose embedding reads only that nested configuration's pad_token_id. No read of the top-level config.pad_token_id appears in the supplied model implementation.

`file:kornia/models/smolvlm2/configuration_smolvlm2.py:head`

> text_config: SmolVLM2TextConfig = field(default_factory=SmolVLM2TextConfig)
>     scale_factor: int = 3
>     image_token_id: int = 49190
>     pad_token_id: int = 2

`file:kornia/models/smolvlm2/modeling_smolvlm2.py:head`

> def __init__(self, config: SmolVLM2TextConfig) -> None:
>         super().__init__()
>         self.config = config
>         self.embed_tokens = nn.Embedding(config.vocab_size, config.hidden_size, padding_idx=config.pad_token_id)
>         self.layers = nn.ModuleList([SmolVLM2TextDecoderLayer(config) for _ in range(config.num_hidden_layers)])
>         self.norm = SmolVLM2RMSNorm(config.hidden_size, eps=config.rms_norm_eps)
>         self.rotary_emb = SmolVLM2RotaryEmbedding(config.head_dim, base=config.rope_theta)

`file:kornia/models/smolvlm2/modeling_smolvlm2.py:head`

> def __init__(self, config: SmolVLM2Config) -> None:
>         super().__init__()
>         self.config = config
>         self.image_token_id = config.image_token_id
> 
>         self.vision_model = SigLip2VisionModel(config.vision_config)
>         self.connector = SmolVLM2Connector(config)
>         self.text_model = SmolVLM2TextModel(config.text_config)

`file:tests/models/test_smolvlm2.py:head`

> text_config = SmolVLM2TextConfig(
>         vocab_size=100,
>         hidden_size=32,
>         intermediate_size=64,
>         num_hidden_layers=2,
>         num_attention_heads=4,
>         num_key_value_heads=2,
>         max_position_embeddings=128,
>         pad_token_id=0,
>     )
>     return SmolVLM2Config(vision_config=vision_config, text_config=text_config, scale_factor=2, image_token_id=7)

Context truncated at 30000 characters: docs/export_support/cases_models.py

Runtime behavior for a deliberately mismatched top-level and nested pad_token_id configuration was not executed; the finding is based on static data-flow inspection.

Make the top-level SmolVLM2Config.pad_token_id and text_config.pad_token_id coherent, preferably by synchronizing the intended authoritative value during configuration construction, and add a regression asserting that a customized pad token is the one used by the decoder embedding.

SHA: `2062884bf7b2bc5a59e12522ff58d0930a07f191`

PR #4166, opened 2026-09-02
test: speed up SigLIP2 coverage

ducha-aiki:

Replaces the production-scale SigLIP2 configuration used by routine unit tests with a small, representative configuration. The tests still cover two transformer layers, attention masking, both linear and identity projection paths, image/text/joint forwards, loss, gradcheck, and compile integration.

It also derives inputs from the test configuration and replaces the 27-case image-preprocessor Cartesian product with three representative cases covering downscaling, non-square batched input, and batched upscaling.

On Apple MPS with torch 2.9.1, the complete float32 file improved from 35.51 seconds on `main` to 1.57 seconds on this branch. The primary test model shrinks from 375,532,034 parameters to 52,770 parameters.

No linked issue.

## What did you check?

```text
python -m pytest tests/models/test_siglip2.py --device=mps --dtype=float32 --durations=5
# 21 passed, 1 skipped in 1.57s

python -m pytest tests/models/test_siglip2.py --device=cpu --dtype=float32,float64 -q
# 44 passed in 0.46s

pre-commit run --all-files
# all hooks passed
```

## How did AI help?

Codex profiled the test structure, implemented the test-only reduction, measured the MPS before/after runtime, and performed an independent code review. I verified the resulting focused tests and repository hooks.

### DIFF
--- a/tests/models/test_siglip2.py
+++ b/tests/models/test_siglip2.py
@@ -17,6 +17,8 @@
 
 """Tests for SigLip2 model."""
 
+from dataclasses import replace
+
 import pytest
 import torch
 
@@ -75,7 +77,7 @@ class TestSigLip2Model(BaseTester):
 
     def test_identity_projections(self, device, dtype, config):
         """Test the default identity projection path with a lightweight model."""
-        config.projection_dim = config.vision_config.hidden_size
+        config = replace(config, projection_dim=config.vision_config.hidden_size)
         model = SigLip2Model(config).to(device, dtype).eval()
         image_size = config.vision_config.image_size
         pixel_values = torch.randn(1, 3, image_size, image_size, device=device, dtype=dtype)
@@ -149,7 +151,7 @@ class TestSigLip2Model(BaseTester):
         assert features.shape == (batch_size, config.projection_dim)
         # Check normalization
         norms = features.norm(dim=-1)
-        self.assert_close(norms, torch.ones_like(norms), rtol=1e-5, atol=1e-5)
+        self.assert_close(norms, torch.ones_like(norms))
 
     def test_get_text_features(self, device, dtype, model, config):
         """Test get_text_features method."""
@@ -166,7 +168,7 @@ class TestSigLip2Model(BaseTester):
         assert torch.isfinite(features).all()
         # Check normalization
         norms = features.norm(dim=-1)
-        self.assert_close(norms, torch.ones_like(norms), rtol=1e-5, atol=1e-5)
+        self.assert_close(norms, torch.ones_like(norms))
 
     def test_attention_mask_handling(self, device, dtype, model, config):
         """Test attention mask handling in text encoder."""
@@ -191,8 +193,6 @@ class TestSigLip2Model(BaseTester):
 
     def test_return_loss(self, device, dtype, model, config):
         """Test forward pass with return_loss=True and verify logit_scale clamping."""
-        import math
-
         batch_size = 2
         image_size = config.vision_config.image_size
         pixel_values = torch.randn(batch_size, 3, image_size, image_size, device=device, dtype=dtype)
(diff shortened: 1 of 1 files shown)


ducha-aiki:
## Review

LGTM — this is a clean test-only reduction and the speedup reproduces on hardware other than MPS. Two nits and two pre-existing issues in the same file that I'd like to flag; none of them block.

### What I verified

Environment: Linux, torch 2.9.1+cu128, python 3.11.14, RTX 4090. `main` (`4990046f`) vs. branch (`7dd92c2b`), run from separate worktrees with `kornia.__file__` checked each time.

| run | main | this branch |
|---|---|---|
| `--device=cpu --dtype=float32,float64` | 29.36s (90 tests) | **0.36s** (44 tests) |
| `--device=cuda --dtype=float32` | 14.30s (46 tests) | **0.98s** (23 tests) |
| `KORNIA_TEST_OPTIMIZER=inductor -k dynamo`, cpu float32 | — | passes (5.25s) |
| `ruff check` / `ruff format --check` | — | clean |

Parameter counts in the description check out exactly: `375532034` → `52770`.

### The identity-projection test is the right call

Worth spelling out for other reviewers, because it's the one place where shrinking the config would have silently dropped a code path: `SigLip2Model` picks `nn.Identity` vs. `nn.Linear` from `projection_dim != hidden_size` (`kornia/models/siglip2/model.py:106-116`). The old `SigLip2Config()` had `projection_dim == hidden_size == 768`, so every test on `main` ran the **Identity** path and the `Linear` path was never exercised. The new fixture (`16 != 32`) flips that to `Linear`, and `test_identity_projections` restores the other branch. I confirmed both instantiations directly — `Identity` on main's default, `Linear` on the new fixture. So this is net *more* coverage, not less.

### Nits (optional)

1. `test_identity_projections` mutates the `config` fixture in place. It's function-scoped so it's safe today, but `dataclasses.replace(config, projection_dim=config.vision_config.hidden_size)` would make the local-only intent explicit and survive anyone widening the fixture scope later.

2. In the new `test_image_preprocessor` parametrization, `image_size` is `(32, 32)` in all three cases, so it no longer varies — it could move out of the tuple into a local. Also, `main`'s grid happened to include `input_size == image_size` (256→256), i.e. the no-op-resize case, which the three new cases don't cover; a fourth `(1, (32, 32), (32, 32))` entry would be nearly free if you want it back.

### Pre-existing, not caused by this PR, but this file is the natural place to fix them

I checked both against `main` before attributing anything to the branch.

3. **The half-precision norm checks are flaky, on both revisions.** `test_get_image_features` / `test_get_text_features` assert `self.assert_close(norms, ones, rtol=1e-5, atol=1e-5)`, and the hardcoded tolerances override `BaseTester`'s dtype-aware ones. Nothing seeds the RNG, so whether an L2 norm rounds inside 1e-5 is luck. Three identical repeats of `--device=cuda --dtype=float16,bfloat16 --isolate-half-precision -k "get_image_features or get_text_features"`:

   - `main`: 2 failed / 4 passed / 2 failed
   - this branch: 3 failed / 1 failed / 0 failed

   On `main`'s CPU float16 the failure is `Greatest absolute difference: 0.00048828125 ... (up to 1e-05 allowed)` — pure fp16 rounding on a unit-norm check, not a real defect. Shrinking the model makes the error smaller and so makes the red rarer, which is a bit unfortunate: it hides the wart instead of fixing it. Dropping the explicit `rtol`/`atol` so `BaseTester`'s dtype-specific tolerances apply would be a one-line fix, in the spirit of #4145/#4147. (PR CI is float32/float64 only, so this only bites `pixi run test-half` / `test-cuda-half`.)

4. **`test_return_loss[bfloat16]` fails deterministically on both revisions**, CPU and CUDA: `Max clamp failed: 99.0 != 100.0`. `logit_scale` is clamped to `log(100.0)` and re-exponentiated; in bfloat16 `4.60517` rounds to `4.59375` and `exp` gives 99.0, while the assertion uses `rel_tol=1e-5`. A dtype-aware tolerance (or comparing in float32) would fix it. Happy to see this as a separate PR.

Neither 3 nor 4 is a regression from this change — I'd merge as is.

---
*Written by Claude (Opus 5) on behalf of @ducha-aiki*

ducha-aiki:
Applied the review feedback in f2b7f075:

- use `dataclasses.replace` so the identity-projection test does not mutate its fixture
- keep the fixed `(32, 32)` preprocessor target local and restore the no-op resize case
- let `BaseTester.assert_close` select dtype-aware tolerances for normalized feature norms
- compare the maximum logit scale against the value representable through `log().exp()` in the tested dtype, covering the deterministic bfloat16 rounding case without a broad hand-tuned tolerance

Verification:

- baseline CPU half run reproduced the reported failures: bfloat16 normalized-feature error and `99.0 != 100.0` max clamp
- full SigLIP2 module, CPU float16/bfloat16/float32/float64: 92 passed
- full SigLIP2 module, CUDA float16/bfloat16/float32 with half-precision isolation: 69 passed
- CPU float32 inductor compile test: 1 passed
- `pixi run pre-commit-all`: all hooks passed

PR #4171, opened 2026-09-02
Fix stale transform matrices in repeated auto augmentation policies

ducha-aiki:

Reset each PolicySequential state before applying a prepared parameter list. Without this reset, selecting the same geometric TrivialAugment policy on consecutive calls could leave transform_matrix cached from the first call while the output and recorded parameters came from the second.

A deterministic regression forces the same translation policy twice and compares the cached matrix with recomputation from the latest parameters. It fails before the implementation fix and passes afterward.

Discovered while validating #4169.

## What did you check?

- Linux CPU float32/float64 auto-operation tests: 32 passed
- Linux CUDA float32/float64 auto-operation tests: 32 passed
- Full pre-commit suite: passed

## How did AI help?

Codex reproduced the order-dependent failure locally, traced the stale cache to the nested policy execution boundary, added the regression, and ran the CPU/CUDA and repository checks above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,9 +129,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
-* Reset an automatic augmentation sub-policy's cached transformation matrix before applying it again. Consecutive
-  calls that selected the same geometric policy could transform the input with the latest parameters while exposing
-  the previous call's matrix through `transform_matrix` (#4171).
+* Reset an automatic augmentation policy's cached transformation matrix before applying it again, both for a
+  sub-policy inside `TrivialAugment`, `RandAugment` and `AutoAugment` and for those policies nested inside
+  `AugmentationSequential`. Consecutive calls that selected the same geometric policy could transform the input with
+  the latest parameters while exposing the previous call's matrix through `transform_matrix`, the sub-policy
+  accumulated one matrix per call for the lifetime of the module, and a nested policy never recorded its parameters
+  or matrix at all. `transform_matrix` on a policy that has not run yet now returns `None` instead of raising
+  `AttributeError` (#4171).
 
 * `MultiResolutionDetector.detect` apportions `num_features` across pyramid levels with largest-remainder
   (Hamilton) apportionment instead of flooring each level's fractional share independently (#4101). The floors
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* Reset an automatic augmentation sub-policy's cached transformation matrix before applying it again. Consecutive
+  calls that selected the same geometric policy could transform the input with the latest parameters while exposing
+  the previous call's matrix through `transform_matrix` (#4171).
+
 * `MultiResolutionDetector.detect` apportions `num_features` across pyramid levels with largest-remainder
   (Hamilton) apportionment instead of flooring each level's fractional share independently (#4101). The floors
   used to discard every fractional part, so a small `num_features` could round the whole apportionment down: with
--- a/kornia/augmentation/auto/base.py
+++ b/kornia/augmentation/auto/base.py
@@ -26,6 +26,7 @@ from kornia.augmentation.container.base import ImageSequentialBase, TransformMat
 from kornia.augmentation.container.ops import InputSequentialOps
 from kornia.augmentation.container.params import ParamItem
 from kornia.core.ops import eye_like
+from kornia.core.utils import is_exporting
 
 NUMBER = Union[float, int]
 OP_CONFIG = Tuple[str, NUMBER, Optional[NUMBER]]
@@ -38,6 +39,9 @@ class PolicyAugmentBase(ImageSequentialBase, TransformMatrixMinIn):
     def __init__(self, policy: List[SUBPOLICY_CONFIG], transformation_matrix_mode: str = "silence") -> None:
         policies = self.compose_policy(policy)
         super().__init__(*policies)
+        # ``TransformMatrixMinIn`` sits after ``nn.Module`` in the MRO, so its ``__init__`` never runs.
+        self._transform_matrix: Optional[torch.Tensor] = None
+        self._transform_matrices: List[Optional[torch.Tensor]] = []
         self._parse_transformation_matrix_mode(transformation_matrix_mode)
         self._valid_ops_for_transform_computation: Tuple[Any, ...] = (PolicySequential,)
 
@@ -140,6 +144,9 @@ class PolicyAugmentBase(ImageSequentialBase, TransformMatrixMinIn):
     ) -> torch.Tensor:
         """Apply a prepared parameter list to the input tensor.
 
+        Resets the cached parameters and transformation matrix first, so a policy nested inside another
+        container ends up in the same state as after :meth:`forward` with these ``params``.
+
         Args:
             input: Input tensor.
             params: Parameters produced by :meth:`forward_parameters`.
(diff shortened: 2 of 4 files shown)


ducha-aiki:
## Review: PR #4171

**Verdict: the fix is correct and safe to merge; one small change requested, plus a few non-blocking notes.**

### What I verified

All runs were done from git worktrees of the PR head (`4791228f`) and the merge base (`5c5981fc`), with `kornia.__file__` printed to confirm the tree under test. This machine has an NVIDIA GPU, so CUDA paths were run for real.

| Check | Result |
|---|---|
| New `test_transform_mat_repeated_policy`, CPU + CUDA, float32/float64 | fails on base, passes on PR |
| `tests/augmentation/test_auto_operation.py` + `tests/augmentation/container/`, CPU + CUDA, float32/float64, `KORNIA_TEST_RUNSLOW=true` | all pass |
| Same auto-op file with `KORNIA_TEST_OPTIMIZER=inductor` on CUDA | pass (no compile tests exist there; `transform_inputs` is not on a compiled path) |
| `pre-commit run --files` on the three changed files | pass |
| CUDA `float16`/`bfloat16` with `--isolate-half-precision` | ~16 failures, **identical set and same stochastic behaviour on base** (dtype mismatch in the mask path of some sampled ops, plus a scatter/gather index assertion in one op). Pre-existing, unrelated to this PR. |

Extra probes (PR vs base, CPU and CUDA):

- TrivialAugment, RandAugment and AutoAugment each called 5 times with a single geometric sub-policy: cached `transform_matrix` equals `get_transformation_matrix(recompute=True)` on PR, mismatches on base (RandAugment matches by accident because its magnitude is fixed by `m`).
- The sub-policy's `_transform_matrices` list holds 1 entry on PR and grows by one `(B,3,3)` tensor per call on base. So this fix also closes an unbounded memory growth in long training loops. Worth one clause in the CHANGELOG entry.
- `PolicySequential` nested inside `ImageSequential`, called 3 times: correct on PR, stale on base.
- `AugmentationSequential(TrivialAugment(...), data_keys=["input","keypoints"])` called 3 times: keypoints follow the recomputed matrix and the keypoint inverse round-trips, on both trees (masks/keypoints read per-op matrices, so they were never affected).
- Replay with explicit `params` reproduces the output.

### Requested change (small)

`clear_state()` goes through `ImageSequentialBase.clear_state()` and also sets `self._params = None`. The bug is purely matrix state, and the narrower helper already exists on the class:

```python
self._reset_transform_matrix_state()
```

I swapped it in locally: the new regression test still passes (CPU and CUDA), and the sub-policy's `_params` behaviour stays byte-for-byte what it was on `main`. With `clear_state()` there is one observable difference: after `sub = ta[0]; sub(x); y = ta(x)`, `sub.inverse(y)` used to return (using stale params from the direct call, i.e. silently wrong) and now raises `ValueError: No parameters available for inversing`. Raising is arguably better than a wrong inverse, but it is an unannounced behaviour change that the narrower call avoids entirely. It also avoids a redundant double reset on the direct `PolicySequential.forward()` path, which already calls `clear_state()` before `transform_inputs`.

### Non-blocking notes (pre-existing, for follow-ups)

1. **Sibling gap in `PolicyAugmentBase.transform_inputs`** (`kornia/augmentation/auto/base.py`): this is the path taken when TrivialAugment/RandAugment/AutoAugment is nested inside `AugmentationSequential`. It neither resets nor updates matrix state and never stores `_params`, while `forward()` duplicates the loop instead of delegating. So `ta.transform_matrix` / `ta._params` on a nested policy reflect whatever the last *direct* call did. Same class of bug as this PR fixes, one level up.
2. **MRO quirk**: `PolicyAugmentBase(ImageSequentialBase, TransformMatrixMinIn)` puts the mixin after `nn.Module`, so `TransformMatrixMinIn.__init__` never runs. `_transform_matrix` only exists after the first `forward()`. A fresh `TrivialAugment(...).transform_matrix`, or one used only nested, raises `AttributeError`. `PolicySequential` uses the opposite base order and is fine.
3. **Test hardening**: the new test relies on seed 0 drawing two different magnitudes. With identical params on both calls it passes on unfixed code too. Passing two explicit distinct `params` to the two calls (or asserting the first matrix is not close to the second) makes it robust to RNG-consumption changes. A RandAugment `n=2` repeated-call case would pin the other realistic user path this PR fixes.
4. `PolicySequential.get_transformation_matrix(recompute=False)` reads each op's *last* matrix, so if a user supplies params that list the same sub-policy twice in one call it returns `m2 @ m2` while `transform_matrix` now correctly yields `m2 @ m1`. Only reachable with hand-built duplicated params; low severity.

Written by Claude (Fable 5.1) on behalf of @ducha-aiki

ducha-aiki:
Pushed `803170a4` addressing the review above, plus the sibling gap in `PolicyAugmentBase.transform_inputs`.

### Changes

- **`PolicySequential.transform_inputs`**: `clear_state()` replaced by `_reset_transform_matrix_state()`, so the nested path no longer nulls the sub-policy's cached params. It now also records the params it applied (guarded by `is_exporting()`), so a sub-policy ends up in the same state as after `forward()` with those params. `sub.inverse(y)` after a nested call now inverts the *nested* call rather than a stale earlier one.
- **`PolicyAugmentBase.transform_inputs`** (`kornia/augmentation/auto/base.py`): now resets state, updates the transformation matrix per module, and stores `_params`, exactly like `forward()`; `forward()` delegates to it instead of duplicating the loop. This fixes the same stale-state bug one level up, for TrivialAugment/RandAugment/AutoAugment nested inside `AugmentationSequential`.
- **`PolicyAugmentBase.__init__`** initialises `_transform_matrix` / `_transform_matrices` (the mixin's `__init__` never runs because of the MRO), so `transform_matrix` on a fresh or nested-only policy returns `None` instead of raising `AttributeError`.
- **Tests**: the repeated-policy regression now drives the two calls with explicit distinct params (a deep copy with `translate_y` shifted by 1) and asserts the matrix moved, so it cannot pass vacuously. New tests pin the nested `AugmentationSequential(TrivialAugment)` matrix and params, the sub-policy params after `forward`, a fresh instance's `transform_matrix`, and a RandAugment `n=2` three-call repeat.
- **CHANGELOG** entry extended to cover the nested policies, the per-call matrix accumulation, and the `None` return.

### Verified (from a worktree of this branch, `kornia.__file__` checked)

- TDD: the three new-behaviour tests failed on the previous head; all five new/changed tests fail with the policy fix reverted, and pass now. CPU + CUDA, float32/float64.
- `tests/augmentation/test_auto_operation.py` + `tests/augmentation/container/` + `tests/augmentation/test_augmentation.py`, `KORNIA_TEST_RUNSLOW=true`, float32/float64: CPU 1756 passed, CUDA 1750 passed (rest skipped/xfailed as before).
- `KORNIA_TEST_OPTIMIZER=inductor` on CUDA: pass. Doctests for `kornia/augmentation/auto`: pass. `pre-commit` on the changed files and `ty check kornia/augmentation/auto`: clean.
- Extra probes on CPU and CUDA: cached matrix equals recomputation after 5 calls for all three policies; matrix lists bounded at 1 entry for both sub-policy and parent; nested keypoints follow the recomputed matrix and round-trip through `inverse`; nested policy `inverse` equals the container's `inverse`.
- CUDA `float16`/`bfloat16` with `--isolate-half-precision`: the pre-existing RandAugment/AutoAugment failures remain (`expected scalar type Float but found Half` in `TransformMatrixMinIn._update_transform_matrix`, from RandAugment's float32 magnitude tensor). The new RandAugment repeat test fails there for that same reason, consistent with the existing `TestRandAugment::test_transform_mat`. CI does not run half precision. That dtype bug is a separate fix.

Written by Claude (Fable 5.1) on behalf of @ducha-aiki

ducha-aiki (commented):
Reviewed at head `803170a4`. No actionable findings.

Verified the focused CPU float32/float64 auto-operation suite (39 passed), a clean diff check, and the completed CI matrix; all reported checks pass.

Verdict: ready to merge. GitHub does not permit the PR author to submit an approval review.

PR #4173, opened 2026-09-02
docs: follow-up fixes to the site revamp

ducha-aiki:
Follow-up to #4155, which landed before these came out of review. Docs-only plus two small script fixes; no library code touched.

### Content corrections

- **Landing page quoted the wrong number.** The filtering card rendered `|count-image-processing|` — `filters + color + enhance + morphology` = **212** — next to a `kornia.filters` label, on a card linking to the `filters` page, which documents **67** names. Colour and enhance already have their own card, so it double-counted. The card now quotes the module it names.
- **`get-started/gpu-acceleration`** claimed the performance page shows comparisons on "CPU, CUDA and Apple silicon", but `benchmarks/results/0.9.0rc1/` only holds `apple-m1`/`apple-m4` cpu+mps runs, and the only "cuda" on the rendered page is inside the contribute command. Reworded to name the devices the committed runs cover, phrased so it stays true as coverage grows rather than pinning a device list.
- **Dead link.** The Community page linked to `https://librecv.org` in both its prose and its `meta` description; the domain no longer resolves (NXDOMAIN). `doc-links.yml` runs `--offline` on pull requests, which is why CI was green.

### Behaviour

- **The hero carousel restarted after a visitor picked a tab**, contradicting its own comment ("choosing a tab stops it for good"). A click arrives as `pointerenter` → `change` → `pointerleave`: `stop()` cleared the timer but left `paused === true`, so the trailing `pointerleave` → `resume()` rescheduled it. Fixed with a `stopped` flag honoured by `pause`/`resume`. Same path on the keyboard via `focusin`/`change`/`focusout`.
- **Navbar dropdown a11y.** The menus now mirror their CSS open state onto `aria-haspopup`/`aria-expanded`, close on <kbd>Escape</kbd>, and the Ecosystem trigger is announced as a button instead of a link that goes nowhere. (They were already keyboard-reachable via `:focus-within`.)

### Build robustness

- `pkgutil.walk_packages` in `conf.py` had no `onerror`, so it swallowed `ImportError` but **re-raised everything else** a `kornia.*` sub-package might throw at import — failing the whole docs build where the intent is clearly "skip it and keep counting".
- `hero_figures` guarded `cpu <= 0` but not `gpu`, while the caller computes `max(gpu, cpu) / min(gpu, cpu)`. A zero accelerator throughput would raise `ZeroDivisionError` at conf-read time and take the build down instead of just omitting the chart.
- `fetch_dependents.py --out` was honoured on write but not on read, so `--out other.json --max-pages 5` merged against the default snapshot and then overwrote the target wholesale — the opposite of the docstring's "a partial crawl merges into the existing file".
- `generate_examples.py` called `torch.stack` on a possibly-empty list in the last block of `main()`, which would abort the script after every other image had been regenerated.
- `_round_headline` promised "never rounds up past the real figure" but rounded to nearest: 1,290 → `"1.3K"`, 9,999 → `"10K"`. Now truncates.
- **Pinned `pydata-sphinx-theme` to the 0.21 line.** The site's CSS and JS target theme internals (`#pst-secondary-sidebar`, `.bd-navbar-elements`, `#pst-page-toc-nav`, `.bd-sidebar-primary`), and Read the Docs installs `.[docs]` **without** `uv.lock`, so an unbounded upstream release lands straight on the live site. Worth noting: my first attempt at a bound silently resolved the theme *down* to 0.17.1 — the pin is deliberately the version the site was built and reviewed against (0.21.0), and `uv.lock`'s resolved version is unchanged.
- Dropped the five `rst_epilog` substitutions no page references.

### Verification

- `pixi run -e default build-docs` (`sphinx -W`) succeeds on theme 0.21.0.
- Rendered output re-checked: the card reads "67 operators · `kornia.filters`"; the other landing figures (289 / 84 / 86 / 14 / "1,200+") are unchanged; no `librecv` reference survives anywhere in the build.
- `pixi run -e default pre-commit-all` passes, as does `.github/scripts/check_doc_links.py --offline`.
- The carousel fix was checked by driving the real `custom.js` against a DOM stub with fake timers: the "selection stops the cycle" case **fails on merged `main`** (the carousel advances off the chosen tab) and passes here. Not added to the repo — kornia has no JS test harness and adding one is out of scope for this fix.

Not addressed here, worth its own issue: `pixi run doctest` is `pytest --doctest-modules kornia/`, so none of the prose `.. code-block:: python` examples added in #4155 are executed by CI, even though `sphinx.ext.doctest` is already enabled.

### DIFF
--- a/docs/generate_examples.py
+++ b/docs/generate_examples.py
@@ -755,14 +755,20 @@ def main():
     with torch.no_grad():
         detections = face_detector(K.color.bgr_to_rgb(img_crowd) * 255.0)[0]
     faces = [K.contrib.FaceDetectorResult(d) for d in detections]
-    boxes = torch.stack([torch.cat([f.top_left, f.bottom_right]) for f in faces if f.score > 0.7])[None]
-    out = img_crowd.clone()
-    for pad in (0, 1, 2):  # three passes give a 3 px outline
-        out = K.image.draw_rectangle(
-            out, boxes + torch.tensor([-pad, -pad, pad, pad]), color=torch.tensor([0.4, 1.0, 0.2])
-        )
-    cv2.imwrite(str(OUTPUT_PATH / "face_detection.png"), K.image.tensor_to_image((out[0] * 255.0).byte()))
-    print(f"Generated image example for FaceDetector. {boxes.shape[1]} faces")
+    confident = [torch.cat([f.top_left, f.bottom_right]) for f in faces if f.score > 0.7]
+    if not confident:
+        # A checkpoint or asset change could leave nothing above the threshold; this is the last
+        # block of main(), so bail out here rather than abort after everything else was written.
+        print("Skipped image example for FaceDetector: no detection above the score threshold")
+    else:
+        boxes = torch.stack(confident)[None]
+        out = img_crowd.clone()
+        for pad in (0, 1, 2):  # three passes give a 3 px outline
+            out = K.image.draw_rectangle(
+                out, boxes + torch.tensor([-pad, -pad, pad, pad]), color=torch.tensor([0.4, 1.0, 0.2])
+            )
+        cv2.imwrite(str(OUTPUT_PATH / "face_detection.png"), K.image.tensor_to_image((out[0] * 255.0).byte()))
+        print(f"Generated image example for FaceDetector. {boxes.shape[1]} faces")
 
 
 if __name__ == "__main__":
--- a/docs/generate_examples.py
+++ b/docs/generate_examples.py
@@ -756,19 +756,15 @@ def main():
         detections = face_detector(K.color.bgr_to_rgb(img_crowd) * 255.0)[0]
     faces = [K.contrib.FaceDetectorResult(d) for d in detections]
     confident = [torch.cat([f.top_left, f.bottom_right]) for f in faces if f.score > 0.7]
-    if not confident:
-        # A checkpoint or asset change could leave nothing above the threshold; this is the last
-        # block of main(), so bail out here rather than abort after everything else was written.
-        print("Skipped image example for FaceDetector: no detection above the score threshold")
-    else:
+    out = img_crowd.clone()
+    if confident:
         boxes = torch.stack(confident)[None]
-        out = img_crowd.clone()
         for pad in (0, 1, 2):  # three passes give a 3 px outline
             out = K.image.draw_rectangle(
                 out, boxes + torch.tensor([-pad, -pad, pad, pad]), color=torch.tensor([0.4, 1.0, 0.2])
             )
-        cv2.imwrite(str(OUTPUT_PATH / "face_detection.png"), K.image.tensor_to_image((out[0] * 255.0).byte()))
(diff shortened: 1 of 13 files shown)


ducha-aiki (commented):
I found two P2 issues that should be addressed before merge: the new no-detection branch can leave a fresh docs build without the image that the landing page always references, and the navbar's ARIA state does not stay synchronized with its CSS-open state when hover and focus overlap. Focused Python/JavaScript checks otherwise passed.

docs/generate_examples.py
```diff
@@ -755,14 +755,20 @@ def _get_tps_args():
...
-    boxes = torch.stack([torch.cat([f.top_left, f.bottom_right]) for f in faces if f.score > 0.7])[None]
-    out = img_crowd.clone()
-    for pad in (0, 1, 2):  # three passes give a 3 px outline
-        out = K.image.draw_rectangle(
-            out, boxes + torch.tensor([-pad, -pad, pad, pad]), color=torch.tensor([0.4, 1.0, 0.2])
-        )
-    cv2.imwrite(str(OUTPUT_PATH / "face_detection.png"), K.image.tensor_to_image((out[0] * 255.0).byte()))
-    print(f"Generated image example for FaceDetector. {boxes.shape[1]} faces")
+    confident = [torch.cat([f.top_left, f.bottom_right]) for f in faces if f.score > 0.7]
+    if not confident:
```
ducha-aiki: [P2] Always write the image referenced by the landing page

This fallback skips writing `face_detection.png`, but that generated PNG is ignored/untracked and `docs/source/index.rst` unconditionally references it. If a checkpoint or source-image change actually exercises this branch, a clean `sphinx -W` build has no image to consume; an incremental build can instead retain a stale result. Please write a fallback image (for example, the unannotated crowd image) to the expected path, or conditionally remove the card image reference.

docs/source/_static/js/custom.js
```diff
@@ -227,6 +233,28 @@ document.addEventListener("DOMContentLoaded", function () {
...
+
+    // The panel opens from CSS (:hover / :focus-within); mirror that state onto the trigger so a
+    // screen reader is told the menu exists and whether it is open. The inline sidebar copy is
+    // always expanded, so it announces as such and needs no listeners.
+    link.setAttribute("aria-haspopup", "true");
+    link.setAttribute("aria-expanded", inline ? "true" : "false");
+    if (inline) return;
+    const expose = (open) => link.setAttribute("aria-expanded", String(open));
+    li.addEventListener("pointerenter", () => expose(true));
+    li.addEventListener("pointerleave", () => expose(false));
```
ducha-aiki: [P2] Combine hover and focus before updating `aria-expanded`

The handlers overwrite `aria-expanded` independently, while the CSS displays the menu for the union of `:hover` and `:focus-within`. For example, focus the trigger, move the pointer over it, then move the pointer away: this handler writes `false`, but focus remains inside and CSS keeps the panel visible. The inverse happens when focus leaves while the pointer remains over the item, and Escape also cannot visually close a still-hovered menu. Please derive the attribute and visible state from one combined state (or track hover and focus separately) so the announced value mirrors what is displayed.

PR #4182, opened 2026-09-02
feat(onnx): widen dynamo ONNX export coverage and add the export support page

shijianjian:
## Summary

Two related pieces toward library-wide ONNX usability:

**1. Dynamo ONNX export coverage (42 files under `kornia/`)**

Operators that failed dynamo graph capture, or had no ONNX lowering, get an export-time path that returns the same values eagerly:

- matrix inverses up to 4x4 through a closed-form adjugate instead of `aten::linalg_inv` (`_torch_inverse_cast`; covers the 3D warps, `PinholeCamera`, `StereoCamera`, `DepthWarper`, `ImageRegistrator`, the `Se2`/`Se3`/`So2` groups, `ParametrizedLine.intersect`, …)
- `median_blur` selects the median through `sort`; the Hausdorff losses pool through `amax`; `distance_transform` avoids `hypot`; `match_nn`/`match_mnn` avoid `cdist`
- branch-free `where` selection and `is_exporting()` guards around data-dependent checks in `Boxes`, `distort_points`/`undistort_points`, `conv_soft_argmax2d`, `JPEGCodecDifferentiable`, `ZCAWhitening`, the `OriNet`/`LAFOrienter`/`HardNet`/`SIFTFeature`/`DeDoDe` feature stack, `LightGlue` (with confidence pruning disabled), and the crop / resize / 3D augmentations

About 140 additional operator cases export, pass the ONNX checker and match eager under onnxruntime. `tests/onnx/test_export_coverage.py` pins each fix. Behavior notes are listed in the CHANGELOG entry (Lie-group constructors keep autograd history instead of re-rooting as `nn.Parameter`; `normalize`/`denormalize` accept a 1-D mean/std; `matrix_cofactor_tensor` no longer raises on an all-singular batch).

**2. "ONNX, torch.compile and torch.export support" page (Resources)**

Every public operator, model and augmentation (874 operators, 1213 probed configurations) with its dynamo ONNX export, `torch.export` and `torch.compile` outcome, grouped by package and section, with the cause of each failure, a search box and result filters.

- Rendered at build time from the committed snapshot `docs/source/_data/export_support.json` (the generator adds ~1 s to the docs build; nothing is probed in CI).
- `docs/export_support/` holds the survey harness that regenerates the snapshot (`run.py`, ~2 h on 8 cores, resumable).
- The ONNX guide now quotes the per-operator share (88 %, 767 of 874) from the same snapshot through `rst_epilog` substitutions.
- The primary sidebar is narrowed to the width of the secondary one (`min(17rem, 25%)`) so the article gets the room on laptop screens; long entries wrap.

## Test plan

- [x] `pre-commit run --all-files`
- [x] `tests/onnx/test_export_coverage.py`, `tests/core/test_helpers.py`, `tests/test_api_surface.py` (60 passed)
- [x] `sphinx -W --keep-going` docs build; page layout checked in headless Chrome at 1280 / 1400 px and the mobile breakpoint
- [ ] CI across the CPU / CUDA / half-precision matrices

https://claude.ai/code/session_014Mr9HVQEKfMe3F2TKR35EZ


kornia/core/utils.py
```diff
@@ -233,16 +342,18 @@ def _torch_inverse_cast(input: torch.Tensor) -> torch.Tensor:
...
+    matrices, falls back to a closed-form adjugate inverse so the resulting graph does not
+    include ``aten::linalg_inv``, which neither ONNX exporter lowers. ``torch.jit.is_tracing()``
+    is JIT-script-safe (unlike ``torch.onnx.is_in_onnx_export``, which contains an ``import``
+    statement).
     """
     KORNIA_CHECK_IS_TENSOR(input, "Input must be torch.Tensor")
     dtype = _normalize_to_float32_or_float64(input.dtype)
-    if torch.jit.is_tracing() and input.shape[-2:] == (3, 3):
-        return _inverse_3x3_closed_form(input.to(dtype)).to(input.dtype)
+    if _is_tracing_or_exporting() and _has_closed_form_inverse(input):
```
ducha-aiki, line 354: [P1] This export fallback is never selected on the supported PyTorch 2.5.1 jobs: that release has no `torch.compiler.is_exporting`, so `_torch_is_exporting` is `None` and `is_exporting()` remains false during Dynamo ONNX export. The current CI demonstrates the systemic effect: seven new cases fail because `linalg_inv_ex`, `median.dim`, and `cdist` remain in the graph or data-dependent branches still execute. Please provide export detection that works across the supported Torch range (or explicitly narrow support), and keep these tests running on the oldest supported export-capable version.

kornia/core/utils.py
```diff
@@ -397,14 +516,39 @@ def is_compiling() -> bool:
...
+    loses its history, so a group built from ``Se3.exp(v)`` would stop propagating gradients
+    to ``v``. Under graph capture (``torch.jit.trace``, ``torch.compile``, ``torch.export`` and
+    the dynamo ONNX exporter) a freshly created parameter is likewise lifted out of the graph
+    as a constant. In both situations the tensor is returned unchanged; otherwise a leaf
+    (a user-provided tensor or an existing parameter) becomes an optimizable ``nn.Parameter``.
+    """
+    if isinstance(x, torch.nn.Parameter):
+        return x
+    if x.grad_fn is not None or torch.jit.is_tracing() or is_compiling() or is_exporting():
+        return x
```
ducha-aiki: [P1] Returning a non-leaf tensor unchanged preserves its graph, but callers assign it as a plain module attribute, so PyTorch neither transforms nor serializes it. For example, `s = Se2.exp(v); s.to(torch.float64)` leaves `s.t` as float32, and `s.state_dict()` is empty; the base revision moved/serialized `_translation` and `_rotation._z`. The same pattern affects `Se3` and `ParametrizedLine`, and moving a derived object to CUDA can therefore leave state on CPU. Please preserve autograd while registering or explicitly handling this module state, with `.to()` and state-dict round-trip tests.

docs/export_support/run.py
```diff
@@ -0,0 +1,95 @@
...
+
+    if not merge_only:
+        with ThreadPoolExecutor(max_workers=len(groups) * 2) as pool:
+            failed = [label for label, rc in pool.map(lambda job: _worker(*job), _jobs(groups)) if rc != 0]
+        if failed:
+            print(
+                f"workers exited abnormally: {failed}; rerun to resume the cases they did not record", file=sys.stderr
+            )
+
+    cmd = [sys.executable, "merge.py", str(RESULTS / "onnx_*.json"), str(RESULTS / "compile_*.json"), str(SNAPSHOT)]
```
ducha-aiki, line 99: [P2] A failed probe worker is only printed here; the runner still invokes `merge.py` and returns its status. The merger iterates whatever ONNX records happen to exist and never checks the expected case inventory, so a missing/crashed group can yield exit 0 and overwrite the user-facing snapshot with silently undercounted coverage. Cached non-crashed records are also reused without revision/version validation. Please fail before merging when workers fail and/or make the merge validate completeness and cache provenance.

ducha-aiki (changes requested):
I found three blocking issues: the new export paths are bypassed on supported PyTorch 2.5.1 (as confirmed by the red CI matrix), derived Lie-group/module state is no longer moved or serialized, and the survey runner can publish an incomplete snapshot while exiting successfully. The focused ONNX/helper tests pass on PyTorch 2.9.1, but the compatibility and module-state regressions need coverage before merge.

ducha-aiki:
@shijianjian please check the review

kornia/core/utils.py
shijianjian: Fixed in f3cda872: `is_exporting()` falls back to `is_compiling()` when torch has no `torch.compiler.is_exporting` (torch < 2.6). On 2.6+ Dynamo constant-folds that flag to `True` for `torch.compile` as well as `torch.export` (`torch/_dynamo/variables/torch.py`), so the fallback has the same semantics on the old releases. Reproduced the seven failures in a torch 2.5.1 venv; `tests/onnx/test_export_coverage.py` and the wider touched suites pass there now, and a new test pins the fallback under `torch.compile`.

kornia/core/utils.py
shijianjian: Fixed in f3cda872: `parameter_if_leaf` became `register_module_state(module, name, x)`. Leaf tensors still become `nn.Parameter`; a tensor with autograd history is registered as a buffer, so `s.to(torch.float64)` converts `s.t` and `s.so2.z`, `state_dict()` carries `_translation` / `_rotation._z` again and `load_state_dict` round-trips, while the graph to `v` is kept; inside a capture the plain attribute stays. `Se2`, `Se3`, `So2` and `ParametrizedLine` each got a `.to()` + `state_dict` round-trip test. (`Quaternion._data`, hence `So3`, is a plain attribute on `main` as well; left as is.)

docs/export_support/run.py
shijianjian: Fixed in f3cda872: `run.py` returns 1 before merging when a worker exited abnormally. Each probe now writes `inventory_<results>.json` (all case names + kornia revision, `-dirty` if `kornia/` is modified, + torch version); on resume, records from another revision or torch are discarded; `merge.py` checks each results file against its inventory (crash markers do not count), requires the groups to agree on revision/torch, stamps the snapshot header from the inventories and aborts otherwise. `--force` on either script overrides. Verified end to end with a one-case run (refused), `--force` (merged) and a tampered inventory (next run starts over).

docs/export_support/probe_compile.py
```diff
@@ -0,0 +1,240 @@
...
+
+POISON_EXIT = 3
+
+
+def _alarm(signum, frame):
+    raise Timeout(f"case exceeded {TIMEOUT}s")
+
+
+signal.signal(signal.SIGALRM, _alarm)
+
```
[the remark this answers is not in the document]
shijianjian: Fixed in f3cda872: `SIGALRM` is now optional (`_HAS_ALARM`); without it the probes run without a per-case timeout instead of failing at import.

kornia/feature/hardnet.py
```diff
@@ -99,11 +99,9 @@ def __init__(self, pretrained: bool = False) -> None:
...
         """Normalize the input by batch."""
-        if not is_mps_tensor_safe(x):
-            sp, mp = torch.std_mean(x, dim=(-3, -2, -1), keepdim=True)
-        else:
-            mp = torch.mean(x, dim=(-3, -2, -1), keepdim=True)
-            sp = torch.std(x, dim=(-3, -2, -1), keepdim=True)
+        # `torch.std_mean` is unsupported on MPS and decomposes to `prims.sum`, which has no ONNX lowering.
+        mp = torch.mean(x, dim=(-3, -2, -1), keepdim=True)
+        sp = torch.std(x, dim=(-3, -2, -1), keepdim=True)
         # WARNING: we need to .detach() input, otherwise the gradients produced by
```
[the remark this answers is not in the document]
shijianjian: Fixed in f3cda872: the fused `torch.std_mean` is back on the eager CPU/CUDA path; the split `mean`/`std` is only used on MPS or under graph capture (`_is_tracing_or_exporting()`), where `std_mean` has no lowering.

tests/onnx/test_export_coverage.py
```diff
@@ -0,0 +1,130 @@
...
+    module.eval()
+    with torch.no_grad():
+        eager = module(*inputs)
+    eager = list(eager) if isinstance(eager, (tuple, list)) else [eager]
+
+    buf = io.BytesIO()
+    with torch.no_grad():
+        program = torch.onnx.export(module, inputs, dynamo=True, opset_version=18, verbose=False)
+    program.save(buf)
+    model = onnx.load_from_string(buf.getvalue())
```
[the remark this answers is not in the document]
shijianjian: The test now skips below torch 2.5 (`hasattr(torch.onnx, "ONNXProgram")`): `torch.onnx.export(..., dynamo=True)` and the `ONNXProgram` it returns arrived together in 2.5, and on every release that has them `.save()` exists. Verified on 2.5.1 and 2.9.1.

shijianjian:
Thanks — all three blocking points were real. Fixed in f3cda872 (plus a merge of `main`):

**1. Export paths bypassed on torch 2.5.1.** `torch.compiler.is_exporting` only exists from 2.6, so `is_exporting()` was always false there. While looking at how to detect export on 2.5 I checked what the flag does on 2.9.1: inside a Dynamo trace `torch._dynamo/variables/torch.py` constant-folds `torch.compiler.is_exporting` to `True` for `torch.compile` as well as `torch.export` (verified with a `torch.compile(backend="eager")` probe). So on the new releases the guard already means "inside a Dynamo capture", and `is_compiling()` is the fallback with identical semantics on 2.5. That is what `is_exporting()` now does, with the reasoning in its docstring, and a test pins the fallback under `torch.compile` with the flag monkeypatched away. Reproduced the seven CI failures in a torch 2.5.1 CPU venv; they pass now, as does a wider 2.5.1 run (onnx, core, liegroup, transform, line, hardnet, matching, augmentation containers/export: 1882 passed). The same folding was also behind the one red `dynamo` job: `crop_by_indices` raised "size must be given to export" under `torch.compile`; it now keeps the eager path when no `size` is given (graph break, as on `main`), and `TestCropByIndices::test_dynamo` passes under inductor.

**2. Derived Lie-group / line state not moved or serialized.** `parameter_if_leaf` is replaced by `register_module_state(module, name, tensor)`: a leaf still becomes an `nn.Parameter`; a tensor with a `grad_fn` is registered as a **buffer**, which keeps the history while `.to()`, `state_dict()` and `load_state_dict()` reach it under the same key (it no longer appears in `parameters()`, which the changelog note now states); inside a capture the plain attribute stays, since neither can be created there. `Se2`, `Se3`, `So2` and `ParametrizedLine` each got a `test_derived_state_moves_and_serializes` (`.to()` / `.double()` moves the state and keeps the graph, `state_dict` round-trips into a leaf-built instance). One pre-existing gap I did not touch: `Quaternion._data` (hence `So3`) is a plain attribute on `main` too, so `Se3.exp(v).to(...)` moved the translation but never the rotation before this PR either.

**3. Survey runner could publish an incomplete snapshot.** `run.py` now returns 1 before merging when a worker exited abnormally. Every probe writes an `inventory_*.json` next to its results with the full case list, the kornia revision (`-dirty` when `kornia/` has local changes) and the torch version; on resume, records from another revision/torch are discarded rather than reused; `merge.py` checks every results file against its inventory (a crash marker does not count as a record), requires all groups to agree on revision/torch, stamps the header from the inventories, and aborts with the shortfall otherwise (`--force` on either script overrides). Exercised end to end: a one-case run is refused, `--force` merges it, and a tampered inventory makes the next run start over.

Also from Copilot: `SIGALRM` is optional in `probe_compile.py` (no timeouts on Windows instead of an import error); HardNet keeps the fused `torch.std_mean` on the eager CPU/CUDA path and only splits it under MPS or graph capture; `tests/onnx/test_export_coverage.py` skips below torch 2.5, where `torch.onnx.export(dynamo=True)` and `ONNXProgram.save` do not exist.

Written by Claude on behalf of @shijianjian

PR #4184, opened 2026-09-02
fix(geometry): restore Boxes3D.to_tensor gradients, a false-positive kink

### LINKED ISSUE #1396: Gradients from Boxes3D.to_tensor aren't accurate

mohansree14:
## What

`Boxes3D.to_tensor` has raised `RuntimeError` whenever its input required grad since 2021, because `gradcheck` disagreed with the analytical gradient on the axis-aligned box used in the original bug report. This has blocked all differentiable use of `Boxes3D` (and `get_boxes_shape`, which calls `to_tensor` internally) since.

## Root cause

`to_tensor` reduces the box's 8 vertices to its min/max corner via `amin`/`amax`. This reduction has a genuine kink — it is not differentiable in the classical sense — wherever multiple vertices exactly tie for an axis extremum. An **axis-aligned box has such a tie on every face by construction**: 4 vertices share each min/max coordinate.

At an exact 4-way tie, PyTorch's `amin`/`amax` backward splits the gradient evenly across the tied vertices (`1/4` each) — a valid subgradient, usable for gradient-based optimization. But `torch.autograd.gradcheck`'s central-difference estimate evaluates to `0.5` for each tied vertex regardless of how many vertices are tied (since perturbing any *one* tied vertex downward alone makes it the unique new extremum, changing the output by `eps`, while perturbing it upward leaves the output unchanged as the other ties still hold). `1/4 ≠ 0.5`, so `gradcheck` reports a mismatch — but this is a well-known artifact of testing non-smooth reductions at their exact kink points with finite differences, not evidence the gradient is wrong.

### Verified by direct reproduction

```
exact axis-aligned box (4-way ties per extremum):  gradcheck FAILS  (0.25 analytical vs 0.5 numerical)
same box, vertices jittered by 1e-3 (no ties):      gradcheck PASSES
fully generic random hexahedron (no ties):          gradcheck PASSES
```

This also explains a detail the original 2021 report didn't: **why 2D `Boxes` was never gated even though it uses the identical reduction.** A 2D quadrilateral's ties are always **2-way**, and PyTorch's `1/2` split *happens to numerically coincide* with the central-difference `0.5` estimate — pure coincidence of the tie count, not because 2D's computation is any more correct. A 3D box's 4-way ties give `1/4`, which doesn't coincide, so the same underlying kink is only *visible* to `gradcheck` in 3D.

## Fix

- Removed the `RuntimeError` guard in `Boxes3D.to_tensor`.
- Rewrote the docstring `Note` to document the tie/kink behavior accurately (cross-referencing why 2D never showed the same failure) instead of calling it a bug.
- Re-enabled the previously commented-out `gradcheck` tests, using **jittered (non-tied) box data** — the correct way to test a function with a known, unavoidable kink at its natural input domain's boundary case.
- Added `test_convention_to_tensor_tie_gradient_is_an_even_subgradient_1396`, which pins the even `1/4` split at an exact tie as the actual, deliberate, now-unguarded behavior (per this repo's documented-defect convention in `AGENTS.md`).
- `get_boxes_shape` is fixed as a side effect, since it calls `to_tensor` internally.

## Testing

- New/re-enabled tests verified to fail against the pre-fix code (`RuntimeError`) and pass against the fix.
- Full `tests/geometry/test_boxes.py`: 68 passed.
- Broader box-adjacent coverage: `tests/geometry/ -k box` (105 passed) and `tests/geometry/transform/test_crop3d.py` (20 passed), both use `Boxes3D` indirectly.
- Doctests for `kornia/geometry/boxes.py`: 8 passed.
- `grep -rnE "#1396|issues/1396" kornia/ tests/` per `AGENTS.md`'s documented-defect gate — the one surviving hit in `kornia/` is retrospective ("this was previously... mistaken for a bug"), not a live defect marker; the `tests/` hits are the new pins.

## Not yet verified

No `pixi` or CUDA/MPS hardware in my local environment — verified with a throwaway `uv` venv (torch 2.14 CPU) instead of the project's usual toolchain.

Fixes #1396


ducha-aiki:
## Review of #4184 — APPROVE at `631a7f3b`, one docstring sentence to correct

Setup: worktree at the PR head, `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU + Apple MPS (`PYTORCH_ENABLE_MPS_FALLBACK` unset); `kornia.__file__` asserted to resolve inside the worktree in every block.

This is the right diagnosis and the right fix. I re-derived every load-bearing claim rather than reading them, and they all hold.

### The tie mechanism, measured

```text
3D axis-aligned box, d(xmin)/d(vertices):
  tied vertices 0,3,4,7 -> [0.25, 0.25, 0.25, 0.25]     others -> [0., 0., 0., 0.]
2D axis-aligned box, d(xmin)/d(vertices):
  tied vertices 0,3     -> [0.5, 0.5]                   others -> [0., 0.]

gradcheck(Boxes3D.to_tensor):
  axis-aligned (4-way tie)   FAIL   numerical 0.3797 vs analytical 0.3881
  jittered (no ties)         PASS
gradcheck(Boxes.to_tensor):
  2D axis-aligned (2-way)    PASS      <- the "coincidence" claim, confirmed
```

The `1/k` vs `0.5` argument is exactly right, and it explains the 2D/3D asymmetry that the 2021 report never accounted for. Worth stating plainly for the record: **the guard was removed because the gradient was always correct, not because the failure was tolerable.**

### Blast radius

Every mode is differentiable at the head, including the two that write into `boxes` in place — which was the thing I most expected to break once the guard stopped hiding that path:

```text
xyzxyz  ok (1,6)     xyzxyz_plus ok (1,6)     xyzwhd ok (1,6)
vertices ok (1,8,3)  vertices_plus ok (1,8,3)
get_boxes_shape ok, gradient non-zero
```

`tests/geometry/test_boxes.py`: **125 passed** CPU (float32 + float64), **63 passed / 5 skipped** on MPS float32. The `#1396` pin returns `0.25` on MPS as well as CPU, so it will not destabilise the new blocking MPS gate. No test anywhere asserted the removed `RuntimeError` (`grep` for its message: one hit, the raise itself), so nothing was pinning the old behaviour.

I also checked that no other prose still calls this non-differentiable: the two remaining `"It is currently non-differentiable"` notes in `boxes.py` are on `Boxes.to_mask` / `Boxes3D.to_mask`, which genuinely are.

### The one ask: the rotated-2D sentence is backwards

In the new `Note`:

> A rotated 2D box, or a degenerate one with a higher-order tie, is not guaranteed the same coincidence.

Rotating a 2D box **removes** the ties — all four vertices get distinct x — so it is the one case that is unambiguously differentiable. The half that is true is the degenerate one:

```text
2D axis-aligned (2-way tie)      gradcheck PASS
2D rotated 30 deg (no ties)      gradcheck PASS
2D zero-width (4-way x tie)      gradcheck FAIL
```

Please drop the rotated clause and keep the degenerate one, e.g. "A degenerate 2D box — zero extent on an axis, so four vertices tie — reaches the same 4-way split and is not guaranteed that coincidence."

### Non-blocking

- The `Note` is 13 lines. It earns most of that, but the sentence about "the case relevant to the original comparison" can go; the mechanism paragraph carries it.
- `grep -rnE "#1396|issues/1396" kornia/ tests/` still hits `boxes.py`, which the AGENTS.md gate flags. Here the surviving text is retrospective ("was previously ... mistaken for an actual gradient bug"), which is the useful kind, so I am not asking you to remove it — just keep the tense unambiguous so a later reader cannot mistake it for a live defect marker.
- Your "Not yet verified" note is honest and appreciated. Consider it covered: I ran it on the project toolchain and on MPS, above.

Fix the one sentence and this is good to merge.

ducha-aiki:
## Delta re-review of #4184 at `f9214aa3` — APPROVE; only a CHANGELOG rebase left

Setup: worktree at the PR head, `kornia/` swapped to the merge-base `d415faa0` for the fail-on-base pass. `~/dev/kornia/.venv/bin/python`, **torch 2.14.0** (the `.venv` moved off 2.9.1 with #4197), CPU + MPS.

The prose inversion is fixed, and fixed precisely — the Note now says a **degenerate** 2D box reaches the 4-way split, where the earlier draft said a rotated one did. Re-executed on the new torch rather than carried over from the last round, because `amin`/`amax` backward and `gradcheck` are exactly the sort of thing a major torch bump can move:

```text
3D axis-aligned   gradcheck  FAIL (GradcheckError)      <- the kink the Note describes
3D jittered       gradcheck  PASS
2D axis-aligned   gradcheck  PASS                       <- the "1/2 coincidence"
2D rotated 30deg  gradcheck  PASS                       <- rotation REMOVES the ties
2D degenerate     gradcheck  FAIL (GradcheckError)      <- vertices [[0,0],[3,0],[3,0],[0,0]]
```

Every claim in the rewritten Note holds on torch 2.14.0, including the one it now makes about degeneracy. The even-subgradient statement checks out too — `d(xmin)/d(vertices)` returns a uniform `1/k` over the tied vertices (`0.125` across all eight for the axis-aligned 3D probe, `0.5` on the two tied 2D vertices).

Tests and gate:

```text
tests/geometry/test_boxes.py   cpu float32+float64   125 passed
tests/geometry/test_boxes.py   mps float32            63 passed, 5 skipped
fail-on-base (kornia/ at d415faa0), cpu float32:
  FAILED TestBbox3D::test_gradcheck                                              RuntimeError
  FAILED TestBbox3D::test_convention_to_tensor_tie_gradient_is_an_even_subgradient_1396   RuntimeError
  2 failed, 66 passed
```

Both new pins discriminate, and the second one names its issue.

### The AGENTS grep, and why the surviving hit is fine

`grep -rnE "#1396|issues/1396" kornia/ tests/` at head returns three lines: the two in your new `Note` and the test's `# #1396:` comment. Normally a surviving hit in `kornia/` means an incomplete change. Here it is the right outcome — your prose cites #1396 as *history* ("This was previously ... mistaken for an actual gradient bug and gated behind a ``RuntimeError``"), not as a live defect, and that sentence is the most useful thing in the block: it stops the next reader from re-filing it. No change wanted.

### The one thing left

`CONFLICTING/DIRTY`, and `git merge-tree` says the conflict is **`CHANGELOG.md` only** — `main` has taken a lot of entries since `d415faa0`. Rebase and it is ready to land; nothing else is outstanding on my side.

Thank you for chasing the degenerate-vs-rotated distinction to a measurement instead of softening the sentence.

PR #4186, opened 2026-09-03
test(onnx): skip the export coverage tests on torch < 2.5

shijianjian:
## Summary

The scheduled CPU matrix went red on torch 2.2.2 / 2.3.1 / 2.4.0 after #4182: all 13 cases in `tests/onnx/test_export_coverage.py` failed. Two different reasons, per the logs:

- **2.2.2 / 2.3.1**: `TypeError: export() got an unexpected keyword argument 'dynamo'` — the keyword does not exist before 2.4.
- **2.4.0**: `dynamo=True` is accepted but routes to the retired `dynamo_export` path, which the onnxscript releases we resolve no longer serve (`OnnxExporterError`).

The module guard was `hasattr(torch.onnx, "ONNXProgram")`, but that class has existed since 2.2 as `dynamo_export`'s return type, so it never skipped. Gate on `torch_version_lt(2, 5, 0)` instead. The floor is 2.5 rather than the 2.6 `tests/filters/test_gaussian.py` uses for its smoke test, on purpose and documented in the file: these cases pin the export paths on the oldest exporter-capable release, and torch 2.5.1 is a blocking PR leg, so a regression there must fail rather than skip.

Also in the same module: `importorskip("onnxscript")` (the `docs` extra installs onnx/onnxruntime without it), `pytestmark = device_agnostic` so accelerator legs do not repeat CPU-bound exports, and the onnxruntime feed asserts a one-to-one input match instead of a silently truncating `zip`.

`tests/onnx/test_resize_onnx.py` had the same unguarded `dynamo=True` calls but was never collected: its tests carried `dynamo` in their names, which the conftest deselects whenever `KORNIA_TEST_OPTIMIZER` is unset, and no dynamo job includes `tests/onnx`. They are renamed for what they check, gated the same way, and now run.

Test-only change, no changelog entry.

## Test plan

- [x] torch 2.4.0 (fresh venv, onnxscript 0.7.1): `tests/onnx` -> 18 passed, both modules skipped
- [x] torch 2.5.1 and 2.9.1: `tests/onnx` -> 35 passed (the 4 Resize tests now collected)
- [x] `--device=cuda` collects nothing from the coverage module (device_agnostic)
- [x] pre-commit on both files

### DIFF
--- a/tests/onnx/test_export_coverage.py
+++ b/tests/onnx/test_export_coverage.py
@@ -30,15 +30,29 @@ import torch
 from torch import nn
 
 import kornia
+from kornia.core._compat import torch_version_lt
 from kornia.core.utils import _torch_inverse_cast
 from kornia.geometry.epipolar.numeric import matrix_cofactor_tensor
 
+# Every tensor here is created device-free and fed to onnxruntime as numpy: the work is CPU-bound
+# whatever ``--device`` says, so the accelerator legs must not repeat it.
+pytestmark = pytest.mark.device_agnostic
+
 onnx = pytest.importorskip("onnx")
 ort = pytest.importorskip("onnxruntime")
+pytest.importorskip("onnxscript")  # the dynamo exporter hard-requires it; the ``docs`` extra lacks it
 
-if not hasattr(torch.onnx, "ONNXProgram"):
-    # ``torch.onnx.export(..., dynamo=True)`` and the ``ONNXProgram`` it returns (with ``.save()``)
-    # arrived together in torch 2.5; older releases only have the TorchScript exporter.
+if torch_version_lt(2, 5, 0):
+    # ``torch.onnx.export(..., dynamo=True)`` only reaches the current exporter -- and returns the
+    # ``ONNXProgram`` with ``.save()`` this file needs -- from torch 2.5. On 2.1-2.3 the ``dynamo=``
+    # keyword does not exist at all (TypeError); 2.4 accepts it but routes to the retired
+    # ``dynamo_export``, which the onnxscript releases we resolve no longer serve.
+    # ``hasattr(torch.onnx, "ONNXProgram")`` is not a usable probe: that class has existed since 2.2
+    # as ``dynamo_export``'s return type, which is why the previous guard did not skip.
+    # The floor is deliberately 2.5 rather than the 2.6 that ``tests/filters/test_gaussian.py``
+    # requires for its smoke test: these cases pin the export-time code paths on the oldest
+    # exporter-capable release, and torch 2.5.1 is a blocking PR leg, so a regression there must
+    # fail rather than skip.
     pytest.skip("the dynamo ONNX exporter needs torch >= 2.5", allow_module_level=True)
 
 
@@ -66,7 +80,10 @@ def _export_and_run(module: nn.Module, inputs: tuple[torch.Tensor, ...]) -> tupl
     onnx.checker.check_model(model)
 
     session = ort.InferenceSession(buf.getvalue(), providers=["CPUExecutionProvider"])
-    feeds = {inp.name: x.numpy() for inp, x in zip(session.get_inputs(), inputs)}
+    # An input the exporter folded into the graph would shorten ``get_inputs()`` and shift the rest
+    # into the wrong slots; the feed must match one to one.
+    assert len(session.get_inputs()) == len(inputs), [i.name for i in session.get_inputs()]
+    feeds = {inp.name: x.numpy() for inp, x in zip(session.get_inputs(), inputs, strict=True)}
     return eager, session.run(None, feeds)
 
 
--- a/tests/onnx/test_resize_onnx.py
+++ b/tests/onnx/test_resize_onnx.py
@@ -22,6 +22,7 @@ import numpy as np
 import pytest
 import torch
 
+from kornia.core._compat import torch_version_lt
 from kornia.geometry.transform import Resize
 
 # Every test in this module must stay device-free: the mark deselects the whole file on non-CPU devices.
@@ -29,9 +30,18 @@ pytestmark = pytest.mark.device_agnostic
 
 onnx = pytest.importorskip("onnx")
 ort = pytest.importorskip("onnxruntime")
+pytest.importorskip("onnxscript")
+
+if torch_version_lt(2, 5, 0):
+    # Same floor as ``tests/onnx/test_export_coverage.py``: ``dynamo=True`` reaches the current
+    # exporter from torch 2.5 (no such keyword on 2.1-2.3, the retired ``dynamo_export`` on 2.4).
+    # These tests used to carry ``dynamo`` in their names, which the conftest deselects whenever
+    # ``KORNIA_TEST_OPTIMIZER`` is unset, so no CI job ever collected them; they are named for
+    # what they check instead.
+    pytest.skip("the dynamo ONNX exporter needs torch >= 2.5", allow_module_level=True)
 
 
-def test_resize_dynamo_with_binding():
+def test_resize_export_binding():
     model = Resize((32, 32), interpolation="bilinear")
     model.eval()
 
@@ -62,7 +72,7 @@ def test_resize_dynamo_with_binding():
             os.unlink(temp_path)
 
 
-def test_resize_upscale_dynamo():
+def test_resize_export_upscale():
     """Test upscaling with dynamo."""
     model = Resize((128, 128))
     model.eval()
@@ -93,7 +103,7 @@ def test_resize_upscale_dynamo():
             os.unlink(temp_path)
 
 
-def test_resize_downscale_dynamo():
+def test_resize_export_downscale():
     """Test downscaling with dynamo."""
     model = Resize((16, 16))
     model.eval()
@@ -124,7 +134,7 @@ def test_resize_downscale_dynamo():
             os.unlink(temp_path)
 
 
-def test_resize_nearest_dynamo():
+def test_resize_export_nearest():
     """Test nearest neighbor interpolation with dynamo."""
     model = Resize((32, 32), interpolation="nearest")
     model.eval()


ducha-aiki:
Reviewed against the failing run ([33704140985](https://github.com/kornia/kornia/actions/runs/33704140985)) and the green legs on this PR. **The gate is correct and unblocks the scheduled matrix** — one comment fix I'd ask for before merge, three small suggestions in the same file, and two follow-ups this uncovered.

## The boundary is right, and it is CI-verified

`torch_version_lt(2, 5, 0)` skips exactly the legs that can't run this file and keeps both that can:

| torch | parsed | skipped? | evidence |
| --- | --- | --- | --- |
| 2.1.2 | `(2, 1, 2)` | yes | leg was already green — the old `hasattr` guard skipped it too |
| 2.2.2 / 2.3.1 / 2.4.0 | `(2, 2, 2)` / `(2, 3, 1)` / `(2, 4, 0)` | yes | the 13 failures |
| 2.5.1 | `(2, 5, 1)` | **no** | this PR's legs print `tests/onnx/test_export_coverage.py .............` — 13 run, 13 pass |
| 2.9.1 | `(2, 9, 1)` | no | green |

So 2.5 is a measured floor, not a guess. I also ran the file off your branch on torch 2.9.1 locally (13 passed) and ruff check/format are clean.

## 1. The comment describes a failure mode that didn't happen (please fix)

> On 2.2-2.4 the same call reaches the retired `dynamo_export` path, which onnxscript >= 0.4 no longer serves, and 2.1 has no `dynamo=`.

The logs disagree, and the split matters because this comment is the only documentation of why the number is 2.5:

- **2.2.2 and 2.3.1** — `TypeError: export() got an unexpected keyword argument 'dynamo'`, all 13 cases. These releases have no `dynamo=` keyword at all, exactly like 2.1.
- **2.4.0** — the only leg that reaches `dynamo_export`: `torch.onnx.OnnxExporterError: Failed to export the model to ONNX. Generating SARIF report at 'report_dynamo_export.sarif'`, with onnxscript **0.5.7** installed.

`dynamo=` first appears in `torch/onnx/utils.py::export` at v2.4.0. No leg raised the `NotImplementedError: Support for dynamo_export has been dropped since onnxscript 0.4.0` quoted in the PR description. As written, the next person to touch this will believe pinning `onnxscript<0.4` would restore 2.2/2.3 coverage — it would not, the kwarg isn't there.

Suggested rewrite, which also pins why the previous guard failed so this doesn't regress a third time:

```python
if torch_version_lt(2, 5, 0):
    # ``torch.onnx.export(..., dynamo=True)`` only reaches the current exporter -- and returns the
    # ``ONNXProgram`` with ``.save()`` this file needs -- from torch 2.5. On 2.1-2.3 the ``dynamo=``
    # keyword does not exist at all (TypeError); 2.4 accepts it but routes to the retired
    # ``dynamo_export``, which the onnxscript releases we resolve no longer serve.
    # ``hasattr(torch.onnx, "ONNXProgram")`` is not a usable probe: that class has existed since 2.2
    # as ``dynamo_export``'s return type, which is why the previous guard did not skip.
    pytest.skip("the dynamo ONNX exporter needs torch >= 2.5", allow_module_level=True)
```

## 2. The repo now carries two different floors for the same exporter

`tests/filters/test_gaussian.py:349-356` gates the same `dynamo=True` path at **2.6**, and its comment argues the case against 2.5 explicitly: 2.5's `dynamo=` still routes through the experimental `_compat.export_compat` shim, and with `fallback=False` implicit "any exporter/onnxscript mismatch would be a hard failure rather than a skip". This file makes the same implicit-`fallback=False` call and picks 2.5. The description says the change matches that file — it picks a different number. (`tests/augmentation/test_onnx_export.py:59`'s 2.4 gate is about opset 20 on the *legacy* tracer, a different exporter entirely.)

That is not just tidiness: `onnxscript` is unpinned, and torch 2.5.1 is a **blocking PR leg** (`pr_test_cpu.yml:89`), not only a nightly one. The next onnxscript release that drops torch-2.5 support turns PR CI red rather than the scheduled matrix. Two defensible calls — align on 2.6, or keep 2.5 with half a sentence saying why the floors differ. What I'd not want is the number sitting there undocumented, because then someone aligns them and silently drops a leg.

(Minor, same area: `_parse_torch_version` truncates at the first non-digit, so a pre-release `2.5.0a0+git...` reports as `(2, 5, 0)` and is not skipped. Only bites someone on a 2.5 branch build; noting it, not asking for it.)

## 3. `onnxscript` is never `importorskip`ed

The module gates on torch version, `onnx` and `onnxruntime`, but `torch.onnx.export(..., dynamo=True)` hard-requires `onnxscript`, which is only in the `dev` extra (`pyproject.toml:47`) — the `docs` extra installs `onnx` and `onnxruntime` **without** it (`pyproject.toml:63-65`). In such an env both existing `importorskip`s pass, the version gate passes, and all 13 cases hard-fail on the import inside the exporter instead of skipping. `tests/filters/test_gaussian.py` does `pytest.importorskip("onnxscript")` for exactly this reason. One line next to the other two.

## 4. Only file in `tests/onnx/` without `device_agnostic`

`test_resize_onnx.py:28`, `test_sequential.py:21` and `test_utils.py:24` all carry `pytestmark = pytest.mark.device_agnostic`; `test_export_coverage.py` does not. Every tensor here is created device-free and the ORT feed is `x.numpy()`, so the work is CPU-bound regardless of `--device` — which means every accelerator leg, including the blocking MPS gate, currently re-runs 13 full dynamo exports plus `onnx.checker` plus onnxruntime sessions that duplicate the CPU leg exactly. None of the 13 appear in `testing/known_failure_xfails/mps_float32.txt`, so adding the mark is a no-op for the manifest contract. Since you're already editing that module header, it's a cheap place to add it.

## 5. Follow-up: the sibling file has the same bug, hidden by a naming accident

`tests/onnx/test_resize_onnx.py` calls `torch.onnx.export(..., dynamo=True)` at lines 46, 78, 109 and 140 with **no version gate**, so it should have failed 2.1-2.4 in the same run. It didn't, because it runs in no CI job at all:

- `conftest.py:193-198` drops every item whose *name* contains `dynamo` or `compile` when `KORNIA_TEST_OPTIMIZER` is unset. All four are named `test_resize_*_dynamo`.
- The one job that sets the optimizer (`dynamo`, `pr_test_cpu.yml:141-150`) scopes pytest to `tests/augmentation tests/enhance tests/filters tests/color tests/geometry tests/losses tests/morphology` — `tests/onnx` is not in that list. The scheduled workflow has no dynamo job.

Reproduced on `main`:

```
pytest tests/onnx --collect-only -q                              -> 31 node ids, none from test_resize_onnx.py
pytest tests/onnx --collect-only -q --noconftest                 -> 35, the 4 resize ones back
KORNIA_TEST_OPTIMIZER=inductor pytest tests/onnx --collect-only  -> the 4 resize ones back
```

Rename them (`test_resize_export_binding`, ...) or swap the name-substring filter for a marker, and give them the same gate when they come back — otherwise the matrix goes red again the moment collection is fixed. Separate PR is fine.

## 6. Follow-up from #4182: `_export_and_run` zips feeds positionally

```python
feeds = {inp.name: x.numpy() for inp, x in zip(session.get_inputs(), inputs)}
```

If the exporter constant-folds one of the inputs into the graph (plausible for the `mean`/`std` tensors in `normalize`, or the camera matrix in `distort_points`), `session.get_inputs()` comes back shorter, `zip` truncates silently, and the remaining tensor lands in the wrong slot. `_assert_same` checks the output count but nothing checks the input count, so the comparison can pass on `torch.rand`-scale data and mask the lowering regression these cases exist to pin. `zip(..., strict=True)` or an explicit length assert.

## Nit

The PR body ends with a Claude Code session link, and commit `818bdeb2` carries a `Claude-Session:` trailer pointing at the same place. Neither resolves for anyone but the author, and the trailer becomes permanent public history on merge — worth stripping from both.

---

Substance is sound; item 1 is the one I'd block on, the rest are cheap-to-take or follow-ups.

ducha-aiki:
@shijianjian please check Opus review

shijianjian:
Thanks — you were right on the error split; I had only read the 2.4.0 log and generalised it. Fixed in cd77ba67:

1. **Comment** rewritten as suggested: 2.1-2.3 have no `dynamo=` at all (`TypeError`), 2.4 reaches the retired `dynamo_export`, and the note on why `hasattr(torch.onnx, "ONNXProgram")` was not a usable probe is in the file. The PR description now states the same split.
2. **Floor**: kept at 2.5, documented in the same comment: these cases exist to pin the export paths on the oldest exporter-capable release (that was your P1 on #4182), and 2.5.1 being a blocking PR leg is the point — a regression there should fail, not skip. `test_gaussian.py`'s 2.6 is a smoke test where skipping is the right call. (Noted the `_parse_torch_version` pre-release caveat; left as is.)
3. `pytest.importorskip("onnxscript")` added next to the other two.
4. `pytestmark = pytest.mark.device_agnostic` added; `--device=cuda` now collects nothing from the module.
5. **Resize tests**: taken here rather than in a follow-up since it is the same gate. The four tests are renamed `test_resize_export_{binding,upscale,downscale,nearest}` (so the name filter no longer hides them), `onnxscript` is importorskipped and the module gets the same 2.5 gate. They pass on 2.5.1 and 2.9.1 (4 passed each) and skip on 2.4.0.
6. `_export_and_run` asserts `len(session.get_inputs()) == len(inputs)` and uses `zip(..., strict=True)`.

Nit: the `Claude-Session:` trailer is gone from the commit (amended, force-pushed) and the link from the PR body.

Verified locally: `tests/onnx` -> 35 passed on 2.5.1 and 2.9.1; 18 passed + 2 modules skipped on a fresh torch 2.4.0 venv.

Written by Claude on behalf of @shijianjian

ducha-aiki (changes requested):
The ONNX tests themselves pass, but the renamed Resize tests leave the tracked standalone helper broken.

`tests/onnx/check.py` still calls `test_resize_dynamo_with_binding`, `test_resize_upscale_dynamo`, `test_resize_downscale_dynamo`, and `test_resize_nearest_dynamo`. Those functions were renamed in this PR to `test_resize_export_*`, so running `python tests/onnx/check.py` now fails immediately with `AttributeError`.

Please update the four calls in `tests/onnx/check.py` to use the new names.

Written by Codex (GPT-5) on behalf of @ducha-aiki

ducha-aiki:
@shijianjian merged too soon, there was one more problem found -- please do a follow-up PR

shijianjian:
Follow-up for the `tests/onnx/check.py` calls: #4191. Verified `python tests/onnx/check.py` passes on torch 2.5.1 and 2.9.1.

Written by Claude on behalf of @shijianjian

PR #4188, opened 2026-09-03
fix(tests): run the polynomial-solver gradchecks instead of shadowing BaseTester.gradcheck

Anai-Guo:
### Description

`tests/geometry/solvers/test_polynomial_solver.py` declares its gradient tests as `def gradcheck(self, device)` instead of `def test_gradcheck(self, device)`:

```python
class TestQuadraticSolver(BaseTester):
    ...
    def gradcheck(self, device):
        coeffs = torch.rand(1, 3, device=device, dtype=torch.float64, requires_grad=True)
        assert self.gradcheck(solver.solve_quadratic, (coeffs), raise_exception=True, fast_mode=True)
```

This has two effects:

1. **The tests never run.** pytest collects `test_*`, so `solve_quadratic` and `solve_cubic` have had no gradient coverage.
2. **They would crash if they did run.** The method name shadows `BaseTester.gradcheck`, so the body's `self.gradcheck(...)` recurses into the test method itself rather than reaching the helper:

```
TypeError: TestQuadraticSolver.gradcheck() got an unexpected keyword argument 'raise_exception'
```

These two are the only `def gradcheck(self` in the entire test suite — the other **146** test files all use `def test_gradcheck`:

```console
$ grep -rn "    def gradcheck(self" tests/
tests/geometry/solvers/test_polynomial_solver.py:51:    def gradcheck(self, device):
tests/geometry/solvers/test_polynomial_solver.py:83:    def gradcheck(self, device):

$ grep -rln "def test_gradcheck" tests/ | wc -l
146
```

### Changes

* Rename both methods to `test_gradcheck` so pytest collects them and `self.gradcheck` resolves to `BaseTester.gradcheck`.
* `(coeffs)` → `(coeffs,)`, so the inputs are a real 1-tuple like every sibling (`self.gradcheck(kgl.inverse_transformation, (trans_01,))`).
* Drop `raise_exception=True, fast_mode=True` and the `assert` wrapper — both are already the `BaseTester.gradcheck` defaults and no sibling passes them.

Naming them `test_gradcheck` also picks up the `*gradcheck*` name-based MPS/XLA marker that `conftest.py` applies, which the comment in `BaseTester.gradcheck` relies on.

### Verification

Reproduced the `TypeError` against the current `main` body, then ran both gradchecks with the fix applied (CPU, float64, real `kornia.geometry.solvers`):

```
--- unpatched (as written in main) ---
TypeError: TestQuadraticSolver.gradcheck() got an unexpected keyword argument 'raise_exception'
--- patched ---
quad gradcheck OK
cubic gradcheck OK
```

`ruff check` and `ruff format --check` (v0.16.4, repo config) both pass on the changed file.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/tests/geometry/solvers/test_polynomial_solver.py
+++ b/tests/geometry/solvers/test_polynomial_solver.py
@@ -49,7 +49,12 @@ class TestQuadraticSolver(BaseTester):
         self.assert_close(roots[0], expected_solutions[0])
 
     def test_gradcheck(self, device):
-        coeffs = torch.rand(1, 3, device=device, dtype=torch.float64, requires_grad=True)
+        # Deterministic, strictly positive discriminant (b^2 - 4ac = 1): `torch.rand` draws
+        # all-positive coefficients, whose discriminant is usually negative, and the no-real-root
+        # branch returns zeros with an identically zero gradient -- so a random draw checks
+        # nothing roughly three times out of four. The zero-discriminant triple is excluded on
+        # purpose: `sqrt` is not differentiable at 0.
+        coeffs = torch.tensor([[1.0, -5.0, 6.0]], device=device, dtype=torch.float64, requires_grad=True)
         self.gradcheck(solver.solve_quadratic, (coeffs,))
 
 
@@ -81,7 +86,9 @@ class TestCubicSolver(BaseTester):
         self.assert_close(roots[0], expected_solutions[0], rtol=1e-3, atol=1e-3)
 
     def test_gradcheck(self, device):
-        coeffs = torch.rand(1, 4, device=device, dtype=torch.float64, requires_grad=True)
+        # Deterministic three-distinct-real-roots case (roots 2, -3, -1/2), so the check does not
+        # depend on an unseeded draw. Repeated roots are excluded: they sit on the `sqrt` kink.
+        coeffs = torch.tensor([[2.0, 3.0, -11.0, -6.0]], device=device, dtype=torch.float64, requires_grad=True)
         self.gradcheck(solver.solve_cubic, (coeffs,))
 
 
--- a/tests/geometry/solvers/test_polynomial_solver.py
+++ b/tests/geometry/solvers/test_polynomial_solver.py
@@ -48,9 +48,9 @@ class TestQuadraticSolver(BaseTester):
         roots = solver.solve_quadratic(coeffs)
         self.assert_close(roots[0], expected_solutions[0])
 
-    def gradcheck(self, device):
+    def test_gradcheck(self, device):
         coeffs = torch.rand(1, 3, device=device, dtype=torch.float64, requires_grad=True)
-        assert self.gradcheck(solver.solve_quadratic, (coeffs), raise_exception=True, fast_mode=True)
+        self.gradcheck(solver.solve_quadratic, (coeffs,))
 
 
 class TestCubicSolver(BaseTester):
@@ -80,9 +80,9 @@ class TestCubicSolver(BaseTester):
         roots = solver.solve_cubic(coeffs)
         self.assert_close(roots[0], expected_solutions[0], rtol=1e-3, atol=1e-3)
 
-    def gradcheck(self, device):
+    def test_gradcheck(self, device):
         coeffs = torch.rand(1, 4, device=device, dtype=torch.float64, requires_grad=True)
-        assert self.gradcheck(solver.solve_cubic, (coeffs), raise_exception=True, fast_mode=True)
+        self.gradcheck(solver.solve_cubic, (coeffs,))
 
 
 class TestMultiplyDegOnePoly(BaseTester):


ducha-aiki:
## Review of #4188 — CHANGES REQUESTED, one input

Setup: worktree at the PR head `38558e76`, `~/dev/kornia/.venv/bin/python`, torch 2.9.1, CPU/macOS; `kornia.__file__` asserted to resolve inside the worktree in every block.

### The diagnosis is right, and I confirmed both halves

The two methods are collected now, and they resolve to `BaseTester.gradcheck` rather than to themselves:

```text
# at the head, --collect-only
tests/geometry/solvers/test_polynomial_solver.py::TestQuadraticSolver::test_gradcheck[cpu]   <- new
tests/geometry/solvers/test_polynomial_solver.py::TestCubicSolver::test_gradcheck[cpu]       <- new
tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_gradcheck[cpu]

3 passed;  whole file: 115 passed (float32 + float64)
```

The `(coeffs)` to `(coeffs,)` change and dropping the redundant `raise_exception`/`fast_mode`/`assert` are all correct — they are the `BaseTester.gradcheck` defaults, and no sibling passes them.

### The blocker: the quadratic gradcheck passes vacuously in 3 draws out of 4

`torch.rand` gives **all-positive** coefficients, so `b^2 - 4ac` is usually negative, `solve_quadratic` returns zeros for the no-real-root case, and the gradient is identically zero. `gradcheck` on a constant-zero function passes without touching the `sqrt` branch this test exists to cover:

```text
300 unseeded draws of torch.rand(1, 3):
  negative discriminant:        224 / 300
  all-zero gradient:            224 / 300      <- gradcheck is vacuous on these
  gradcheck failures:             0 / 300
  cubic, same probe:              0 / 300 all-zero    (torch.rand(1, 4) is fine)

seed 0: coeffs [0.9700, 0.7078, 0.4594] -> roots [[0.0, 0.0]], grad [[0., 0., 0.]]
```

So the test you are switching on is a coin flip between "checks the derivative of the quadratic formula" and "checks nothing," redecided on every CI run. That is worth fixing in the PR that makes them run for the first time — otherwise the change reads as coverage that is not there.

**Ask: use a deterministic positive-discriminant input for `TestQuadraticSolver::test_gradcheck`.** The neighbouring parametrization already supplies one, so nothing new needs inventing:

```python
def test_gradcheck(self, device):
    # b^2 - 4ac > 0: torch.rand gives all-positive coefficients, whose discriminant is
    # usually negative, and the zero-root branch has an identically zero gradient.
    coeffs = torch.tensor([[1.0, -5.0, 6.0]], device=device, dtype=torch.float64, requires_grad=True)
    self.gradcheck(solver.solve_quadratic, (coeffs,))
```

Measured on all three cases from `test_solve_quadratic`:

```text
[1, -5, 6]  (positive)  roots [3.0, 2.0]    grad [-5.0, -1.0, 0.0]   gradcheck PASS
[1,  4, 4]  (zero)      roots [-2.0, -2.0]  grad [ 4.0, -1.0, 0.0]   gradcheck FAIL  (sqrt kink -- do not use)
[1,  2, 3]  (negative)  roots [ 0.0,  0.0]  grad [ 0.0,  0.0, 0.0]   gradcheck PASS  (vacuously)
```

Note the zero-discriminant triple fails: `sqrt` at `0` has an infinite derivative, so that one is genuinely non-differentiable and should stay out.

### Non-blocking

`TestCubicSolver::test_gradcheck` is non-vacuous on all 300 draws, so it is not urgent — but it is still unseeded, and kornia has no `manual_seed` anywhere, so its pass/fail is redrawn every run. A fixed coefficient tensor there too would cost nothing and make both tests reproducible.

Good find, and thank you for the `grep` that establishes these two are the only instances in 146 test files — that is what turns a typo report into a "no third case is hiding" claim.

Anai-Guo:
Thanks — the vacuity finding is right and I reproduced it. Both gradchecks are now pinned to deterministic coefficients.

`TestQuadraticSolver::test_gradcheck` uses your suggested `[[1.0, -5.0, 6.0]]` (discriminant `b^2-4ac = 1`). `TestCubicSolver::test_gradcheck` uses `[[2.0, 3.0, -11.0, -6.0]]` — three distinct real roots `(2, -3, -1/2)`, already one of the `test_solve_quadratic_in_cubic` cases.

Measured here (torch 2.5.1, CPU, float64, module exec'd from the repo source with `KORNIA_CHECK_SHAPE` stubbed):

```text
quad [1, -5,  6]   roots [3.0, 2.0]           grad [-5.0, -1.0, 0.0]                 gradcheck PASS
quad [1,  4,  4]   roots [-2.0, -2.0]         grad [ 4.0, -1.0, 0.0]                 gradcheck FAIL   <- excluded (sqrt kink)
quad [1,  2,  3]   roots [0.0, 0.0]           grad [ 0.0,  0.0, 0.0]                 gradcheck PASS   <- vacuous
cubic [2, 3, -11, -6]   roots [2.0, -3.0, -0.5]    grad [0.75, -0.5, ~0, ~0]         gradcheck PASS
cubic [2, -6, 6, -2]    roots [1.0, 1.0, 1.0]      grad [-1.5, -0.5, 0.0, 0.0]       gradcheck FAIL   <- excluded (triple root)
```

Both now match the style of the existing `TestQuarticSolver::test_gradcheck`, which already pins a distinct-root quartic for the same reason.

ducha-aiki:
## Delta re-review of #4188 at `cd5cae1b` — APPROVE, nothing open

Setup: worktree at the PR head, `kornia/` swapped to the merge-base `4b69473a` for the fail-on-base pass. `~/dev/kornia/.venv/bin/python`, **torch 2.14.0** (the `.venv` moved off 2.9.1 with #4197), CPU.

You reproduced the vacuity finding rather than taking it on trust, and both coefficient triples are well chosen. Re-executed rather than read:

```text
quadratic  [[1, -5, 6]]        roots [3.0, 2.0]                         grad [-5, -1, 0]    gradcheck True
cubic      [[2, 3, -11, -6]]   roots [2.0, -3.0, -0.5] (to 1e-15)       grad [0.75, -0.5, ~0, ~0]   gradcheck True
```

The near-zero third and fourth cubic gradients are correct rather than another vacuity: the sum of the roots is `-b/a` by Vieta, so `d/da = b/a^2 = 0.75` and `d/db = -1/a = -0.5` exactly, and `c`, `d` genuinely do not move it. `gradcheck` checks the full Jacobian, not that sum, and passes.

Both triples avoid the kink you name in the comments — the quadratic discriminant is `+1`, not `0`, and the cubic's three roots are distinct — so neither sits on the `sqrt` non-differentiable point.

The premise of the PR holds at collection level, which is the check that matters for a "this test never ran" claim:

```text
merge-base 4b69473a   -k gradcheck   ->  1/57 tests collected   (only TestQuarticSolver)
head       cd5cae1b   -k gradcheck   ->  3/59 tests collected   (Quadratic, Cubic, Quartic)
```

Worth noting for the record that the old methods were not merely misnamed — `def gradcheck(self, device)` calling `self.gradcheck(...)` shadowed `BaseTester.gradcheck` with itself, so had pytest ever collected them they would have recursed. Renaming is the whole fix.

Whole file: **115 passed** CPU float32+float64. `MERGEABLE`, no conflict with `main`, no CI failures.

No CHANGELOG entry needed — test-only change, nothing user-visible.

### Verdict

**APPROVE.** Nothing outstanding. Thanks for checking what the test did the first time it ran instead of just switching it on.

ducha-aiki (approved):

PR #4192, opened 2026-09-03
fix(geometry): accept as_padded_sequence on VideoBoxes.to_tensor

### LINKED ISSUE #4176: Bug: `VideoBoxes.get_boxes_shape()` raises `TypeError` — inherited body passes `as_padded_sequence` to the `to_tensor` override
## Summary

`Boxes.get_boxes_shape` is implemented as

```python
boxes_xywh = cast(torch.Tensor, self.to_tensor("xywh", as_padded_sequence=True))   # kornia/geometry/boxes.py:273 at e0969438
```

`VideoBoxes` inherits that body unchanged but overrides `to_tensor` with the narrower signature `to_tensor(self, mode=None)` (`boxes.py:890`), so the inherited call site raises `TypeError` on every `VideoBoxes` instance. `VideoBoxes` is what `AugmentationSequential` builds for video box inputs (`kornia/augmentation/container/augment.py:635`), so any user who asks the returned boxes for their shapes hits it.

## Reproduction

```python
import torch
from kornia.geometry.boxes import Boxes, VideoBoxes

v = torch.tensor([[[[[1., 2.], [4., 2.], [4., 3.], [1., 3.]]]]])   # (B=1, T=1, N=1, 4, 2)
vb = VideoBoxes.from_tensor(v)
vb.to_tensor("xywh")          # works: [[[[1., 2., 4., 2.]]]]
vb.compute_area()             # works: [[3.]]
vb.get_boxes_shape()          # TypeError
Boxes.from_tensor(v[:, 0], mode="vertices_plus").get_boxes_shape()   # (tensor([[2.]]), tensor([[4.]]))
```

## Observed

```
TypeError: VideoBoxes.to_tensor() got an unexpected keyword argument 'as_padded_sequence'
```

## Expected

`(heights, widths)` with a leading time axis, `(B, T, N)` each, matching what the other inherited accessors already return for `VideoBoxes`. Two equally small fixes: have `VideoBoxes.to_tensor` accept and forward `as_padded_sequence` (it already calls `super().to_tensor(mode, as_padded_sequence=False)` at `boxes.py:901`), or override `get_boxes_shape` the way `to_tensor` is overridden. `Boxes.clip`/`clamp` pass the same keyword (`boxes.py:675`, `:698`) and are inherited too, so whichever fix is chosen should be checked against them; I did not run those.

No behavior change for any call that works today: a `TypeError` becomes the answer.

## Disposition

**n/a — single symbol.**

## Context

Found in review of #4060 (bbox/boxes convention blocks, batch 4a), whose `Boxes` block now states what `get_boxes_shape` returns; the statement is true for `Boxes` and the method is unreachable on the documented subclass. `VideoBoxes` belongs to batch 4c in #4020, which is the natural place for the pin. Not related to #4016 (export/docs), which is closed.

## Environment

Alphaxiaoteng:
## Description

Fixes #4176.

`Boxes.get_boxes_shape` (and other inherited helpers) call `self.to_tensor(..., as_padded_sequence=True)`. `VideoBoxes.to_tensor` only accepted `mode`, so every `VideoBoxes.get_boxes_shape()` call raised `TypeError`.

Forward `as_padded_sequence` to `Boxes.to_tensor` and keep the temporal reshape.

## Test plan

```bash
pytest tests/geometry/test_boxes.py::TestVideoBoxes::test_get_boxes_shape_accepts_as_padded_sequence -q --device=cpu
```

1 passed.

## AI assistance

AI-generated fix and regression test. I reviewed the diff and ran the focused VideoBoxes test locally.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -968,18 +968,22 @@ class VideoBoxes(Boxes):
         out.temporal_channel_size = temporal_channel_size
         return out
 
-    def to_tensor(self, mode: Optional[str] = None) -> torch.Tensor | list[torch.Tensor]:  # type: ignore[override]
+    def to_tensor(
+        self, mode: Optional[str] = None, as_padded_sequence: bool = False
+    ) -> torch.Tensor | list[torch.Tensor]:  # type: ignore[override]
         r"""Cast :class:`VideoBoxes` to a tensor with the temporal axis restored.
 
         Args:
             mode: Output box format forwarded to :meth:`Boxes.to_tensor`. When
                 ``None``, uses the stored mode (``vertices_plus`` by default).
+            as_padded_sequence: Forwarded to :meth:`Boxes.to_tensor`. Inherited
+                helpers such as :meth:`get_boxes_shape` pass ``True``.
 
         Returns:
             Tensor shaped :math:`(B, T, \ldots)` where :math:`T` is
             :attr:`temporal_channel_size`.
         """
-        out = super().to_tensor(mode, as_padded_sequence=False)
+        out = super().to_tensor(mode, as_padded_sequence=as_padded_sequence)
         if isinstance(out, torch.Tensor):
             return out.view(-1, self.temporal_channel_size, *out.shape[1:])
         # If returns a list of boxes.
--- a/tests/geometry/test_boxes.py
+++ b/tests/geometry/test_boxes.py
@@ -1173,6 +1173,16 @@ class TestVideoBoxes(BaseTester):
         cloned.data[0, 0, 0, 0] = -1
         assert not torch.equal(cloned.data, video_boxes.data)
 
+    def test_get_boxes_shape_accepts_as_padded_sequence(self, device, dtype):
+        # #4176: inherited get_boxes_shape passes as_padded_sequence=True to to_tensor.
+        boxes = self._sample_video_boxes(device, dtype, batch=1, time=1, n_boxes=1)
+        video_boxes = VideoBoxes.from_tensor(boxes)
+        heights, widths = video_boxes.get_boxes_shape()
+        assert heights.shape == (1, 1, 1)
+        assert widths.shape == (1, 1, 1)
+        self.assert_close(heights, torch.tensor([[[3.0]]], device=device, dtype=dtype))
+        self.assert_close(widths, torch.tensor([[[3.0]]], device=device, dtype=dtype))
+
     def test_public_api_surface(self):
         # Pin #4016: VideoBoxes is exported and the overrides are documented.
         assert "VideoBoxes" in boxes_module.__all__


ducha-aiki:
## Review of #4192 at `57d7fa39` — APPROVE, on a CHANGELOG entry

Setup: worktree at the PR head, `kornia/` swapped to the merge-base `4d0783bf` for every A/B pass. `~/dev/kornia/.venv/bin/python`, **torch 2.14.0** (the `.venv` moved off 2.9.1 with #4197), CPU + MPS.

Thank you for taking this one — and for the minimal shape of the fix. Forwarding the keyword was the option I would have picked over overriding `get_boxes_shape`, because it fixes the whole family rather than one member. Which brings me to the main point of this review.

### Your PR fixes more than its description claims

`#4176` named `get_boxes_shape` and guessed at `clip`/`clamp` as the siblings — that guess was wrong, and I said in the issue that I had not run them. The actual set of inherited call sites passing `as_padded_sequence=True` is three, in two public methods:

```text
boxes.py:341   get_boxes_shape
boxes.py:756   to_mask   (CPU hotpath)
boxes.py:779   to_mask   (vectorized/export hotpath)
```

So `VideoBoxes.to_mask` was broken on `main` in exactly the same way, and your one-line forward repairs it too. Measured, `B=2, T=3, N=4`:

```text
                    base (4d0783bf)                                        head (57d7fa39)
get_boxes_shape     TypeError: unexpected keyword 'as_padded_sequence'     (2, 3, 4), (2, 3, 4)
to_mask(8, 8)       TypeError: unexpected keyword 'as_padded_sequence'     (6, 4, 8, 8)
compute_area        (6, 4)                                                 (6, 4)          <- unchanged
to_tensor('xyxy')   (2, 3, 4, 4)                                           (2, 3, 4, 4)    <- unchanged
```

The `get_boxes_shape` result is right, not merely non-raising — it agrees elementwise with the flattened `Boxes` reference:

```text
VideoBoxes.get_boxes_shape()  ->  h (2, 3, 4)   w (2, 3, 4)
Boxes.from_tensor(v.reshape(-1, 4, 4, 2), 'vertices_plus').get_boxes_shape()
   torch.equal(h.reshape(-1, 4), fh) -> True
   torch.equal(w.reshape(-1, 4), fw) -> True
```

and the issue's own repro now returns the documented `(heights, widths)` with a leading time axis, matching the `Boxes` line it was compared against:

```text
vb.get_boxes_shape()   ->  ([[[2.]]], [[[4.]]])   shapes (1, 1, 1) / (1, 1, 1)
Boxes reference        ->  ([[2.]],   [[4.]])
```

Fail-on-base for your test: **1 failed** on `4d0783bf` (`TypeError`), 1 passed at head. Whole file: **126 passed** CPU float32+float64, **63 passed / 5 skipped** MPS float32.

### Two things to add, both small

1. **A CHANGELOG entry is required** (`AGENTS.md`: "`CHANGELOG.md` is maintained per change, not at release time"). Under `## Unreleased` → `### Bug fixes`, citing #4192. Worth saying `to_mask` as well as `get_boxes_shape`, since that is what a user searching the changelog will have hit.

2. **Please pin `to_mask` too.** It is newly reachable and nothing tests it, so its shape convention is currently undecided-by-accident. Note what it returns: `(B*T, N, H, W)` — time flattened — where `get_boxes_shape` restores the time axis. That is not a defect you introduced and I am not asking you to change it: it matches `compute_area`, which returns `(6, 4)` on `main` today. But the class now has two conventions live at once, and the one your PR makes reachable should be written down before someone relies on the other. A single assertion next to your existing one is enough.

Also worth widening the docstring line by one word — "Inherited helpers such as :meth:`get_boxes_shape` pass ``True``" is true but undersells it; `to_mask` does as well.

### Verdict

**APPROVE** on the CHANGELOG entry; the `to_mask` pin is a strong recommendation rather than a hard gate. The production change is correct as written and I have verified it against the base on both backends — neither ask touches `kornia/geometry/boxes.py`.

ducha-aiki:
Closing recommendation for #4192 (head `57d7fa39`): please close this in favour of #4365.

#4365 makes the same production change: `VideoBoxes.to_tensor` accepts `as_padded_sequence` and forwards it. It also adds what this PR still lacks: the CHANGELOG entry asked for in the review above, a full-shape pin for `to_mask`, and updates to the #4249 pins that `main` has gained since this branch was cut. Merged onto current `main` (`dc0af1a4`), this head fails 2 `TestVideoBoxes` tests on CPU float32. The strict xfail `test_convention_inherited_shape_and_mask_work_on_the_temporal_wrapper_4249` now XPASSes, and `test_wart_inherited_methods_break_on_the_temporal_wrapper_4249` still expects the `TypeError`. The class `.. warning::` would also keep describing a bug that no longer exists.

Thank you for the fix. Forwarding the keyword was the right approach, and #4365 lands exactly that. Only one of the two can merge.

PR #4193, opened 2026-09-03
refactor(core): extract closed-form small-matrix kernels into _small_linalg

ducha-aiki:
Step 1 of the small-matrix linalg plan: a **semantics-preserving extraction**, no behaviour change, no new public API.

`kornia/core/utils.py` is 641 lines of device probes, dtype casts, tracing detection, autocast, dataclass conversion and `batched_forward` — and, mixed in with those, all of kornia's closed-form matrix arithmetic. This pulls the arithmetic out into a leaf module and leaves every policy decision where it already lives.

## The boundary

```
kornia/core/_small_linalg.py   — a true leaf: imports torch and typing, nothing from kornia
├── _adjugate_2x2 / _adjugate_3x3 / _adjugate_4x4   (moved unchanged)
├── _inverse_3x3_cross                              (the eager kernel)
└── _inverse_3x3_scalar                             (the trace/export kernel)

kornia/core/utils.py           — owns every policy decision, unchanged in name and module
├── _is_tracing_or_exporting        execution mode
├── _inverse_3x3_closed_form        mode dispatch
├── _adjugate_closed_form           size dispatch + its NotImplementedError
├── _has_closed_form_inverse
└── _torch_inverse_cast             dtype promotion
```

Keeping the dispatchers in `utils.py` is load-bearing twice over:

- **It avoids an import cycle.** `_inverse_3x3_closed_form` calls `_is_tracing_or_exporting`; moving one without the other gives `utils → _small_linalg → utils`. A function-local import is not an escape — `torch.jit.script` rejects it outright with `UnsupportedNodeError: import statements aren't supported`.
- **It means no production call site changes.** `conversions.py` and `imgwarp.py` import `_inverse_3x3_closed_form` from `kornia.core.utils` (4 call sites); `_adjugate_*` stay importable from there too.

## Deliberately not fixed in passing

The kernels keep exactly the semantics they have today, warts included, and the new tests do **not** pin the warts:

- they compute in the caller's dtype, with no promotion (promotion is `_torch_inverse_cast`'s job);
- shape and dtype are caller preconditions, not runtime checks;
- `_adjugate_3x3` of a 4×4 silently uses the leading 3×3 block; of a 2×2 it raises an incidental `IndexError`;
- singular-input output is unspecified.

Fixing any of that while moving code would defeat the point of a reviewable refactor. Each is recorded in the design note and gets its own change.

## One test changes

`tests/geometry/test_conversions.py` decides "did this call die inside the closed-form inverse?" by comparing code objects, deliberately, so that a rename is a re-raise rather than a silent skip. After the split, the innermost frame on a backend with no `cross` kernel for the dtype is `_inverse_3x3_cross`, not the dispatcher — so the check is repointed there. That is **narrower** than before, and it resolves the residual the test's own comment records ("this identifies the failing ROUTINE, not the individual `cross` line inside it").

`test_skip_probe_re_raises_everything_it_cannot_identify` — the pin on that helper — still passes all four of its cases.

## New tests

`tests/core/test_small_linalg.py`, on valid shapes and real floating dtypes only: smoke, cardinality batched and unbatched, agreement with `torch.linalg` on a deterministic well-conditioned family, the exact `A @ adj(A) == det(A) · I` identity in float64, gradcheck, dynamo, TorchScript, and ONNX export on **both** exporters.

Inputs are built without touching the global RNG — a seeded `torch.rand` still mutates it, and a fresh draw per run turns a tolerance assertion into a coin flip.

The cross kernel's ONNX case is an **informational skip, not a gate**. `torch.linalg.cross` lowers on **both torch versions CI runs** — 2.9.1 on both exporters, 2.5.1 on the legacy one (2.5.1's dynamo path needs `onnxscript`, which those legs do not install, so it is out of reach rather than broken). It stays a skip because torch 2.0–2.4 are untested and CI does not run them either, and because the dispatcher never takes that kernel under export, so nothing in kornia depends on it lowering.

### A rationale in the source turns out to be wrong

Measuring the above closed a question the first commit left open, and the answer contradicts a comment that has been in `kornia/core/utils.py`:

> *"it lowers to basic arithmetic that every opset supports, whereas `cross` may not."*

`cross` **does** lower, on every torch version kornia tests. What neither exporter lowers, on either version, is `aten::linalg_inv` — which is what `_torch_inverse_cast` reaches for a closed form to avoid in the first place.

The cross/scalar split is still worth keeping, but for a different reason than the one written down: **kernel coverage, not lowering**. `torch.linalg.cross` has real per-dtype gaps where scalar arithmetic needs no kernel at all — torch 2.5.1 has no `bfloat16` `cross` on MPS, which is exactly the gap the skip probe in `tests/geometry/test_conversions.py` exists for. The third commit corrects the comment in all three places it appears. No behaviour change: collapsing the branches would be one, and belongs in its own PR.

## Verified locally

torch 2.9.1, macOS arm64, CPU:

- `tests/core/` + `tests/geometry/test_conversions.py` + `tests/geometry/transform/test_imgwarp.py` + `tests/augmentation/test_onnx_export.py`, float32 and float64 — **1280 passed, 12 skipped, 15 xfailed**. Same skip and xfail counts as the pre-change baseline.
- `tests/geometry` + `tests/augmentation` + `tests/feature` + `tests/onnx`, float32 — **5531 passed, 0 failed**.
- `KORNIA_TEST_OPTIMIZER=inductor` over the touched test files — 86 passed, exercising the dynamo/compile cases that a default run deselects.
- **torch 2.5.1** (in a purpose-built venv, after CI flagged it): `tests/core/test_small_linalg.py` + `test_helpers.py` — **76 passed, 5 skipped**, the skips being the dynamo-exporter cases now gated on torch ≥ 2.6.
- `pixi run pre-commit-all` clean. `pixi run typecheck` reports one diagnostic, in `kornia/feature/lightglue.py`, which this PR does not touch.

### MPS (same machine, `PYTORCH_ENABLE_MPS_FALLBACK` unset)

- `tests/core/test_small_linalg.py --device=mps` — **float32: 37 passed, 8 skipped**; **float16 + bfloat16: 64 passed, 8 skipped**. No process abort: these kernels are pure arithmetic, so half precision never reaches the MPS LU path that aborts with `MPSMatrixDecompositionLU.mm:1227: failed assertion 'Only MPSDataTypeFloat32 is supported.'`. All 8 skips are the float64 exactness case and `BaseTester`'s gradcheck guard — no hidden kernel gaps.
- Characterization set A/B'd against a `main` worktree (`kornia.__file__` checked to confirm the worktree shadowed the editable install):

  | | main | this branch |
  |---|---|---|
  | failed | 2 | 2 — *the same two* |
  | passed | 704 | 741 |
  | skipped | 123 | 131 |
  | xfailed | 5 | 5 |

  The two failures are `TestSolveWithMask::test_smoke` / `test_all_bad`, both `aten::linalg_lu_solve.out` not implemented on MPS. They are lines 61–62 of `testing/known_failure_xfails/mps_float32.txt`, they fail identically on `main`, and they run through `safe_solve_with_mask`, which this PR does not touch.

**Not verified: CUDA.** No GPU available here. Worth a look before this lands — @ducha-aiki has offered to run it on a CUDA machine. The half-precision path is the interesting one: the kernels compute in the caller's dtype with no promotion, so fp16/bf16 goes straight through the cofactor arithmetic, and that has never been exercised on CUDA.

## Not in this PR

No call-site migrations, no scaling, no public API, no docs page, no `api_surface.json` entry, no performance claim. Those are sequenced one consumer at a time in the design note, each validated on that consumer's real inputs.

### DIFF
--- a/kornia/core/_small_linalg.py
+++ b/kornia/core/_small_linalg.py
@@ -21,8 +21,9 @@ Private. These are **kernels**, not an API: every one of them is basic arithmeti
 fixed-size matrix, with no LAPACK, cuSOLVER or backend ``linalg`` call. That is what makes
 them usable where ``torch.linalg`` is not -- neither ONNX exporter lowers
 ``aten::linalg_inv``, the Jetson wheel fails to ``dlopen`` its LAPACK backend, MPS has no
-kernel for several decompositions, and none of the ``linalg`` family has a half-precision
-CPU kernel.
+kernel for several decompositions, and the ``linalg`` family has no half-precision kernel on
+any backend -- on CUDA ``linalg.inv`` gives ``Low precision dtypes not supported. Got Half``
+and ``linalg.det`` gives ``"lu_factor_cusolver" not implemented for 'Half'``.
 
 The contract is deliberately minimal, because the callers own the policy:
 
--- a/kornia/core/_small_linalg.py
+++ b/kornia/core/_small_linalg.py
@@ -1,0 +1,192 @@
+# LICENSE HEADER MANAGED BY add-license-header
+#
+# Copyright 2018 Kornia Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Closed-form arithmetic for 2x2, 3x3 and 4x4 matrices.
+
+Private. These are **kernels**, not an API: every one of them is basic arithmetic on a
+fixed-size matrix, with no LAPACK, cuSOLVER or backend ``linalg`` call. That is what makes
+them usable where ``torch.linalg`` is not -- neither ONNX exporter lowers
+``aten::linalg_inv``, the Jetson wheel fails to ``dlopen`` its LAPACK backend, MPS has no
+kernel for several decompositions, and none of the ``linalg`` family has a half-precision
+CPU kernel.
+
+The contract is deliberately minimal, because the callers own the policy:
+
+- They compute in whatever dtype the caller supplies. **No promotion** -- that belongs to
+  :func:`kornia.core.utils._torch_inverse_cast`.
+- Exact shape and a real floating dtype are **caller preconditions, not runtime checks**.
+  Behavior on anything else is unspecified: ``_adjugate_3x3`` of a 4x4 silently uses the
+  leading 3x3 block, and of a 2x2 raises an incidental ``IndexError``. The contractual size
+  guard lives on the dispatcher :func:`kornia.core.utils._adjugate_closed_form`.
+- Output for a singular input is unspecified. There is no validity or conditioning
+  guarantee; a zero determinant is not a reliable singularity test in floating point.
+- No cross-mode bit-identity promise.
+
+Execution-mode policy (which kernel a caller takes under tracing or export), size dispatch,
+dtype promotion and validation all live in :mod:`kornia.core.utils`. Keeping them there is
+what lets this module import nothing from kornia and stay a leaf.
+"""
+
+from typing import Tuple
+
+import torch
+
+__all__ = [
+    "_adjugate_2x2",
+    "_adjugate_3x3",
+    "_adjugate_4x4",
+    "_inverse_3x3_cross",
+    "_inverse_3x3_scalar",
+]
+
+
+def _adjugate_2x2(input: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+    """Return the adjugate and determinant of batched 2x2 matrices, in basic arithmetic only."""
+    a = input[..., 0, 0]
+    b = input[..., 0, 1]
+    c = input[..., 1, 0]
+    d = input[..., 1, 1]
+    det = a * d - b * c
+    adj = torch.stack([torch.stack([d, -b], dim=-1), torch.stack([-c, a], dim=-1)], dim=-2)
+    return adj, det
+
+
+def _adjugate_3x3(input: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+    """Return the adjugate and determinant of batched 3x3 matrices, in basic arithmetic only."""
+    a = input[..., 0, 0]
+    b = input[..., 0, 1]
+    c = input[..., 0, 2]
+    d = input[..., 1, 0]
+    e = input[..., 1, 1]
+    f = input[..., 1, 2]
+    g = input[..., 2, 0]
+    h = input[..., 2, 1]
+    i = input[..., 2, 2]
+
+    # Cofactors (signed minors).
+    c00 = e * i - f * h
+    c01 = -(d * i - f * g)
+    c02 = d * h - e * g
+    c10 = -(b * i - c * h)
+    c11 = a * i - c * g
+    c12 = -(a * h - b * g)
+    c20 = b * f - c * e
+    c21 = -(a * f - c * d)
+    c22 = a * e - b * d
+
+    det = a * c00 + b * c01 + c * c02
+
+    # Adjugate is the transpose of the cofactor matrix.
+    row0 = torch.stack([c00, c10, c20], dim=-1)
+    row1 = torch.stack([c01, c11, c21], dim=-1)
+    row2 = torch.stack([c02, c12, c22], dim=-1)
+    adj = torch.stack([row0, row1, row2], dim=-2)
+
+    return adj, det
+
+
+def _adjugate_4x4(input: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+    """Return the adjugate and determinant of batched 4x4 matrices, in basic arithmetic only.
+
+    Laplace expansion over the 2x2 minors of the top two rows (``s``) and bottom two rows (``c``),
+    the standard 4x4 cofactor scheme (e.g. MESA ``gluInvertMatrix``).
+    """
+    a = input[..., 0, 0]
+    b = input[..., 0, 1]
+    c = input[..., 0, 2]
+    d = input[..., 0, 3]
+    e = input[..., 1, 0]
+    f = input[..., 1, 1]
+    g = input[..., 1, 2]
+    h = input[..., 1, 3]
+    i = input[..., 2, 0]
+    j = input[..., 2, 1]
+    k = input[..., 2, 2]
+    l_ = input[..., 2, 3]
+    m = input[..., 3, 0]
+    n = input[..., 3, 1]
+    o = input[..., 3, 2]
+    p = input[..., 3, 3]
+
+    s0 = a * f - b * e
+    s1 = a * g - c * e
+    s2 = a * h - d * e
+    s3 = b * g - c * f
+    s4 = b * h - d * f
+    s5 = c * h - d * g
+
+    c5 = k * p - l_ * o
+    c4 = j * p - l_ * n
+    c3 = j * o - k * n
+    c2 = i * p - l_ * m
+    c1 = i * o - k * m
+    c0 = i * n - j * m
+
+    det = s0 * c5 - s1 * c4 + s2 * c3 + s3 * c2 - s4 * c1 + s5 * c0
+
+    row0 = torch.stack(
+        [f * c5 - g * c4 + h * c3, -b * c5 + c * c4 - d * c3, n * s5 - o * s4 + p * s3, -j * s5 + k * s4 - l_ * s3],
+        dim=-1,
+    )
+    row1 = torch.stack(
+        [-e * c5 + g * c2 - h * c1, a * c5 - c * c2 + d * c1, -m * s5 + o * s2 - p * s1, i * s5 - k * s2 + l_ * s1],
+        dim=-1,
+    )
+    row2 = torch.stack(
+        [e * c4 - f * c2 + h * c0, -a * c4 + b * c2 - d * c0, m * s4 - n * s2 + p * s0, -i * s4 + j * s2 - l_ * s0],
+        dim=-1,
+    )
+    row3 = torch.stack(
+        [-e * c3 + f * c1 - g * c0, a * c3 - b * c1 + c * c0, -m * s3 + n * s1 - o * s0, i * s3 - j * s1 + k * s0],
+        dim=-1,
+    )
+    adj = torch.stack([row0, row1, row2, row3], dim=-2)
+
+    return adj, det
+
+
+def _inverse_3x3_cross(input: torch.Tensor) -> torch.Tensor:
+    """Closed-form 3x3 inverse via three fused cross products. The eager path.
+
+    ``inv(M) = adj(M) / det``, and for a 3x3 the adjugate rows are cross products of the
+    columns: with columns ``(a, b, c)``, the inverse rows are ``(b x c, c x a, a x b) / det``,
+    ``det = a . (b x c)``. Three fused ``cross`` ops instead of nine scalar cofactor
+    expressions and four stacks -- far fewer kernel launches, which dominate on small matrices.
+
+    Prefer :func:`_inverse_3x3_scalar` wherever a graph is being captured: a backend that has
+    no ``cross`` kernel for the working dtype makes this raise rather than return, and not
+    every ONNX opset is known to lower it.
+    """
+    col_a = input[..., :, 0]
+    col_b = input[..., :, 1]
+    col_c = input[..., :, 2]
+    row0 = torch.linalg.cross(col_b, col_c, dim=-1)
+    row1 = torch.linalg.cross(col_c, col_a, dim=-1)
+    row2 = torch.linalg.cross(col_a, col_b, dim=-1)
+    det = (col_a * row0).sum(-1)
+    return torch.stack([row0, row1, row2], dim=-2) / det[..., None, None]
+
+
+def _inverse_3x3_scalar(input: torch.Tensor) -> torch.Tensor:
+    """Closed-form 3x3 inverse via the plain scalar adjugate. The trace/export path.
+
+    Lowers to basic arithmetic that every standard ONNX opset supports, and calls no ``cross``.
+    Algebraically identical to :func:`_inverse_3x3_cross`; the two differ only in the order the
+    same products are accumulated, so they agree to rounding rather than bit-for-bit.
+    """
+    adj, det = _adjugate_3x3(input)
+    return adj / det[..., None, None]
--- a/kornia/core/_small_linalg.py
+++ b/kornia/core/_small_linalg.py
@@ -167,9 +167,17 @@ def _inverse_3x3_cross(input: torch.Tensor) -> torch.Tensor:
     ``det = a . (b x c)``. Three fused ``cross`` ops instead of nine scalar cofactor
     expressions and four stacks -- far fewer kernel launches, which dominate on small matrices.
 
-    Prefer :func:`_inverse_3x3_scalar` wherever a graph is being captured: a backend that has
-    no ``cross`` kernel for the working dtype makes this raise rather than return, and not
-    every ONNX opset is known to lower it.
+    Prefer :func:`_inverse_3x3_scalar` wherever a graph is being captured. The reason is
+    **kernel coverage, not ONNX lowering**: a backend with no ``cross`` kernel for the working
+    dtype makes this raise rather than return -- torch 2.5.1 has no ``bfloat16`` ``cross`` on
+    MPS, while 2.9.1 does.
+
+    ONNX lowering is *not* a reason, measured: ``torch.linalg.cross`` lowers on both torch
+    versions kornia's CI runs -- 2.5.1 (legacy exporter) and 2.9.1 (legacy and dynamo) -- to
+    ``Slice``/``Mul``/``Sub``/``Concat``. What neither exporter lowers on either version is
+    ``aten::linalg_inv``, which is what :func:`kornia.core.utils._torch_inverse_cast` avoids by
+    reaching for a closed form in the first place. torch 2.0-2.4 is untested; CI does not run it
+    either.
     """
     col_a = input[..., :, 0]
     col_b = input[..., :, 1]
--- a/kornia/core/utils.py
+++ b/kornia/core/utils.py
@@ -195,7 +195,12 @@ def _inverse_3x3_closed_form(input: torch.Tensor) -> torch.Tensor:
         return _inverse_3x3_cross(input)
 
     # Under tracing/export (legacy ONNX / jit.trace / dynamo ONNX) stick to the plain scalar
-    # adjugate: it lowers to basic arithmetic that every opset supports, whereas ``cross`` may not.
+    # adjugate. NOTE the original rationale here -- "whereas ``cross`` may not [lower]" -- does not
+    # hold on the torch versions CI runs: measured, ``torch.linalg.cross`` lowers on 2.5.1 (legacy
+    # exporter) and on 2.9.1 (both exporters). The split is kept because it is still the safer
+    # capture path (scalar arithmetic needs no per-dtype kernel at all, and ``cross`` has real
+    # kernel gaps -- no bfloat16 on MPS in torch 2.5.1), not because ``cross`` fails to export.
+    # Collapsing the two branches is a behavior change and belongs in its own PR.
     return _inverse_3x3_scalar(input)
 
 
--- a/kornia/core/utils.py
+++ b/kornia/core/utils.py
@@ -27,6 +27,13 @@ import torch.nn.functional as F
 from torch.linalg import inv_ex
 
 from kornia.core._compat import torch_version_ge
+from kornia.core._small_linalg import (
+    _adjugate_2x2,
+    _adjugate_3x3,
+    _adjugate_4x4,
+    _inverse_3x3_cross,
+    _inverse_3x3_scalar,
+)
 from kornia.core.check import KORNIA_CHECK, KORNIA_CHECK_IS_TENSOR, KORNIA_CHECK_TYPE
 from kornia.core.exceptions import DeviceError
 
@@ -163,12 +170,15 @@ def _l2_normalize(input: torch.Tensor, dim: int = 1) -> torch.Tensor:
 
 
 def _inverse_3x3_closed_form(input: torch.Tensor) -> torch.Tensor:
-    """Closed-form inverse for batched 3x3 matrices.
+    """Closed-form inverse for batched 3x3 matrices, dispatching on the execution mode.
 
     Used as an ONNX-traceable fallback to ``torch.linalg.inv``: the legacy ONNX
     exporter does not lower ``aten::linalg_inv`` (as of opset 17). Computed via
     the adjugate / determinant formula, which is composed entirely of basic
     arithmetic ops that all standard ONNX opsets support.
+
+    The arithmetic lives in :mod:`kornia.core._small_linalg`; this function owns only the
+    choice between the two kernels, which is execution-mode policy and therefore stays here.
 
     Args:
         input: Tensor of shape ``(..., 3, 3)``.
@@ -180,23 +190,13 @@ def _inverse_3x3_closed_form(input: torch.Tensor) -> torch.Tensor:
         (no explicit check, same as ``torch.linalg.inv`` itself).
     """
     if not _is_tracing_or_exporting():
-        # inv(M) = adj(M) / det, and for a 3x3 the adjugate rows are cross products of the
-        # columns: with columns (a, b, c), the inverse rows are (b x c, c x a, a x b) / det,
-        # det = a . (b x c). Three fused ``cross`` ops instead of nine scalar cofactor
-        # expressions and four stacks — far fewer kernel launches (dominant on small matrices).
-        col_a = input[..., :, 0]
-        col_b = input[..., :, 1]
-        col_c = input[..., :, 2]
-        row0 = torch.linalg.cross(col_b, col_c, dim=-1)
-        row1 = torch.linalg.cross(col_c, col_a, dim=-1)
-        row2 = torch.linalg.cross(col_a, col_b, dim=-1)
-        det = (col_a * row0).sum(-1)
-        return torch.stack([row0, row1, row2], dim=-2) / det[..., None, None]
+        # Eager: three fused ``cross`` ops beat nine scalar cofactor expressions and four
+        # stacks, because kernel launches dominate on matrices this small.
+        return _inverse_3x3_cross(input)
 
     # Under tracing/export (legacy ONNX / jit.trace / dynamo ONNX) stick to the plain scalar
     # adjugate: it lowers to basic arithmetic that every opset supports, whereas ``cross`` may not.
-    adj, det = _adjugate_3x3(input)
-    return adj / det[..., None, None]
+    return _inverse_3x3_scalar(input)
 
 
 def _is_tracing_or_exporting() -> bool:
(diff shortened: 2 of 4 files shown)


ducha-aiki:
## Review: CUDA verification + two test-quality nits

Reviewed in a worktree at `9f2dcb68` and **ran the CUDA paths that were listed as unverified**. Everything checks out: this really is a semantics-preserving move, and I could not make it differ from the base revision on GPU. Two nits below are about the new tests, not the code being moved.

Machine: RTX 4090, torch 2.9.1+cu128, CUDA 12.8, Python 3.11, Linux. Baseline is the merge-base `1ba6c03a`; the branch also merges cleanly into current `main` (`f039d424`).

### The move itself is exact

- Function text of `_adjugate_2x2` / `_adjugate_3x3` / `_adjugate_4x4` in `_small_linalg.py` is **identical** to the versions on `main`, and the body of `_inverse_3x3_cross` is line-for-line the old eager branch.
- **Bit-parity on CUDA.** 512 random 3×3 matrices through `_inverse_3x3_closed_form` and `_torch_inverse_cast`, base vs branch, `torch.equal` (not `allclose`): identical in float32, float64 **and** float16.
- TorchScript and trace still behave: `torch.jit.script(_inverse_3x3_closed_form)` and `torch.jit.script(_torch_inverse_cast)` compile and agree with eager on CUDA on both revisions, and the `jit.trace` graph still contains no `aten::linalg_inv`. The cross-module call does not upset recursive scripting.
- **No perf cost from the extra Python frame.** Interleaved timing on CUDA (`B=1` and `B=32`, 9 trials × 2000 calls, trees alternated to cancel drift): medians 65.3 / 65.8 µs on the branch vs 64.6 / 64.9 µs on base, minima 63–64 µs on both. Noise, not a regression — worth stating explicitly since the eager kernel's whole rationale is launch overhead.

### CUDA test runs

| run | branch | base |
|---|---|---|
| `tests/core/ tests/geometry/test_conversions.py tests/geometry/transform/test_imgwarp.py tests/augmentation/test_onnx_export.py --device=cuda --dtype=float32,float64` | **1246 passed**, 18 skipped, 28 deselected, 15 xfailed | 1174 passed, 18 skipped, 28 deselected, 15 xfailed |
| `tests/geometry tests/augmentation tests/onnx --device=cuda --dtype=float32` | **4760 passed**, 3307 skipped, 69 deselected, 36 xfailed, 11 xpassed | identical, all five counts |

1246 − 1174 = 72, exactly the new file; every skip/deselect/xfail count is unchanged, so nothing went quiet.

Also green on CUDA:

- `tests/core/test_small_linalg.py`: **72 passed** at float32/float64; **72 passed** again with `--tf32`; **72 passed** at float16/bfloat16 under `--isolate-half-precision` (2m16s). Same file on CPU: 72 passed at f32/f64, 72 at f16/bf16.
- `KORNIA_TEST_OPTIMIZER=inductor` over `tests/core/ + test_conversions.py + test_imgwarp.py`, cuda, f32/f64: **1298 passed, 0 failed** — the dynamo cases a default run deselects.
- The seven pins routed through the repointed `_skip_if_closed_form_inverse_unavailable`, at float16/bfloat16 on cuda with subprocess isolation: **12 passed, 0 skipped** on both trees. They ran rather than skipping, which is the thing that mattered here.
- `ruff check` / `ruff format --check` clean on all four files; `ty check kornia` reports only the pre-existing `kornia/feature/lightglue.py` deprecation.

### The repointed probe lands where the comment says

Worth confirming rather than assuming, since the whole point of that helper is to identify a frame:

```
>>> normalize_homography(torch.ones(1, 3, 3, dtype=torch.bool), (4, 5), (4, 5))
innermost frame: _inverse_3x3_cross in core/_small_linalg.py
NotImplementedError: "cross" not implemented for 'Bool'
```

and `test_skip_probe_re_raises_everything_it_cannot_identify` passes **without** taking its own `no dtype without a torch.linalg.cross cpu kernel` escape — so case B really is exercised against the new frame, not silently skipped.

### Nit 1 — the ONNX assertion can never fire

In all three ONNX tests the guard is:

```python
assert not any(name.lower().startswith("linalg") for name in ops), ops
```

A missing lowering does not produce a `linalg`-named node — it raises. Measured here on torch 2.9.1 with `torch.linalg.inv`: the legacy exporter raises `UnsupportedOperatorError: Exporting the operator 'aten::linalg_inv' to ONNX opset version 18 is not supported`, and the dynamo exporter raises `ConversionError`. So the real gate in those tests is "`torch.onnx.export` did not throw", and the assertion is decorative. The actual graphs are:

```
scalar, dynamo=False: {Add, Concat, Constant, Div, Gather, Mul, Neg, Sub, Unsqueeze}
scalar, dynamo=True:  {Add, Concat, Div, Gather, Mul, Neg, Sub, Unsqueeze}
cross,  dynamo=False: {Concat, Constant, Div, Gather, Mul, ReduceSum, Slice, Sub, Unsqueeze}
cross,  dynamo=True:  {Concat, Div, Gather, Mul, ReduceSum, Split, Sub, Unsqueeze}
```

Two ways to make it carry weight, pick either:

- Minimal: `"linalg" in name.lower()` instead of `startswith`. A dynamo-exporter fallback node is named `aten_linalg_inv`, which `startswith("linalg")` misses and `in` catches.
- Substantive: assert the positive, `assert ops <= BASIC_ARITHMETIC_OPS, ops`, which also catches a future decomposition quietly introducing a `MatMul`/`Det`-shaped op. Brittler across torch versions, so it needs a comment saying so.

If neither is wanted, a one-line comment saying the export call is the gate would stop the next reader from trusting the assert.

### Nit 2 — every batch entry is the same matrix

`_well_conditioned` ends in `x.expand(*batch, n, n).clone()`, so for `batch=(2,)` and `(2, 3)` all slices are identical (verified: `(x == x[0, 0]).all()` is `True`). The batched `test_cardinality`, `test_matches_linalg_inv_on_well_conditioned_input` and `test_dynamo` therefore pin shape but cannot catch a kernel that mixed up batch elements — an indexing slip that broadcast element 0 across the batch would pass. Low risk for these two kernels, which touch only the trailing two dims through `...`, but the batched parametrization currently buys less than it looks like it does. A per-element scale keeps both the determinism and the conditioning that the docstring is careful about:

```python
x = x.expand(*batch, n, n).clone()
if batch:
    idx = torch.arange(x.numel() // (n * n), device=device, dtype=dtype).reshape(*batch, 1, 1)
    x = x * (1 + idx * 0.05)
```

### Two smaller things

- `TestAdjugateKernels.test_scripts` compares only `scripted(x)[0]`, dropping the determinant, while `TestInverse3x3Kernels.test_scripts` compares the whole output. One more line makes them symmetric.
- The module docstring's "none of the `linalg` family has a half-precision CPU kernel" is narrower than reality — it is not CPU-specific. On CUDA here: `torch.linalg.inv` → `RuntimeError: linalg.inv: Low precision dtypes not supported. Got Half`, `torch.linalg.det` → `NotImplementedError: "lu_factor_cusolver" not implemented for 'Half'`, both float16 and bfloat16. Dropping the word "CPU" makes the claim stronger and still true.

### Agreed non-issues

No `CHANGELOG.md` entry and no `tests/api_surface.json` edit: correct, every name involved is private and the two the production call sites import from `kornia.core.utils` keep their name and module. The Jetson / `dlopen` claim in the new docstring matches the prose already in `conversions.py` and `imgwarp.py`.

### Unrelated, found while sweeping (not this PR)

`tests/geometry/test_grid.py::test_normalized_meshgrid_trace_matches_eager_at_unrepresentable_sizes` fails 8 of 10 parametrizations on **cuda float32** and passes all 10 on cpu float32 — identically on this branch and on the merge-base, so it is pre-existing and out of scope here. Flagging it because it does not look intentional and there is no xfail on it.

### Verdict

Approve on the CUDA question: bit-identical outputs, identical pass/skip/xfail counts against the base revision, green under half precision, TF32, inductor and both ONNX exporters. The two nits are test-strength suggestions, not blockers.

ducha-aiki:
Thanks for running the GPU paths — bit-parity in float16 and the interleaved timing were the two things I could not check here, and the `_skip_if_closed_form_inverse_unavailable` confirmation (case B exercised against the new frame rather than skipping past it) is exactly the check I wanted on that repoint.

Both nits verified before fixing, and both fixes verified to discriminate afterwards. Pushed as `db89866b`.

## Nit 1 — confirmed, and worse than "decorative"

You are right that a missing lowering raises rather than emitting a node. Measured here on torch 2.9.1 with `torch.linalg.inv`:

```
dynamo=False: RAISED UnsupportedOperatorError -- so no node exists to assert on
dynamo=True:  RAISED ConversionError          -- so no node exists to assert on
```

So `startswith("linalg")` could not fire from the failure mode it was written for.

I took a variant of your substantive option rather than either as written. `assert ops <= BASIC_ARITHMETIC_OPS` is the property I actually want, but the allowlist differs across the four graphs you listed (`Constant` and `Slice` on legacy, `Split` on dynamo) and I have no way to check torch 2.0–2.4, so it would break on benign decomposition churn. Instead the check is a **substring denylist** over `linalg/det/inverse/matmul/gemm/lu/svd/qr/einsum` — substring, so it catches the `aten_linalg_inv` fallback naming your minimal option targets, and it covers the case the raise does not: a future decomposition that exports *successfully* through something heavier than arithmetic.

Verified it discriminates rather than assuming:

```
matmul graph ops = ['MatMul']         -> caught
synthetic 'aten_linalg_inv' node      -> caught
real scalar-kernel graph              -> passes
```

Your third suggestion is in too: the comment now says outright that the successful export call is the primary gate, so the next reader does not over-trust the assert.

## Nit 2 — confirmed

```
batch=(2,):   all slices equal slice 0 -> True
batch=(2, 3): all slices equal slice 0 -> True
```

Your snippet applied as written. Scaling leaves the condition number untouched, so the tolerance argument in the docstring still holds, and the slices now differ. Passes at float32/float64/float16/bfloat16.

## Both smaller things

`TestAdjugateKernels.test_scripts` now compares the determinant as well as the adjugate. And the docstring drops "CPU" — your CUDA measurements make the broader claim true, and it is stronger.

## On the Copilot review

For the record, since it sits above yours and was written against `9f2dcb68`: its `dynamo` kwarg point was real and is fixed in `80225e0a` (gated on torch ≥ 2.6, following `test_gaussian.py::test_onnx_export_modern`) — CI's 2.5.1 legs caught it before you did. Its BytesIO claim does not hold on 2.9.1, measured:

```
dynamo=False: return=NoneType     bytes_written=337   parseable=True
dynamo=True:  return=ONNXProgram  bytes_written=1560  parseable=True
```

The dynamo exporter writes to the buffer *and* returns an `ONNXProgram`; reading the buffer is fine.

## Also since your review

`36c107f6` corrects something your sweep did not cover. I built a torch 2.5.1 environment to close the "does `cross` lower on older torch" question, and the answer contradicts a comment that has been in `core/utils.py`: *"it lowers to basic arithmetic that every opset supports, whereas `cross` may not."* It does lower — 2.5.1 on the legacy exporter, 2.9.1 on both. The split's real justification is kernel coverage (no `bfloat16` `cross` on MPS in torch 2.5.1), not lowering. Comments corrected in all three places; no behaviour change, since collapsing the branches would be one.

## Your unrelated finding

`test_normalized_meshgrid_trace_matches_eager_at_unrepresentable_sizes` failing 8/10 on cuda float32 and passing on cpu, identically on both trees — agreed that is out of scope here, and agreed it does not look intentional given there is no xfail. I cannot reproduce it without a GPU. Want me to open an issue with your numbers, or will you?

## State

Re-verified after the fixes: torch 2.9.1 CPU **1253 passed / 11 skipped / 15 xfailed** (counts unchanged), **126 passed** across f32/f64/f16/bf16; torch 2.5.1 **76 passed / 5 skipped**; MPS **91 passed / 8 skipped** across f32/f16/bf16; `pre-commit-all` clean.

ducha-aiki:
## Re-run on CUDA at `db89866b` — clean, plus one regression in the new denylist

Re-ran the whole GPU matrix on the updated head. Same machine as before: RTX 4090, torch 2.9.1+cu128, CUDA 12.8, Python 3.11; baseline still the merge-base `1ba6c03a`.

**Both nit fixes verified to discriminate, not just to pass.** And the three commits are safe by construction on the library side: with docstrings stripped, the ASTs of `kornia/core/_small_linalg.py` and `kornia/core/utils.py` are **identical** between `9f2dcb68` and `db89866b`, so `36c107f6`'s prose correction cannot have moved a number.

### CUDA results, `db89866b`

| run | result | vs. base |
|---|---|---|
| `tests/core/ test_conversions.py test_imgwarp.py test_onnx_export.py --device=cuda --dtype=float32,float64` | **1246 passed**, 18 skipped, 28 deselected, 15 xfailed | base 1174 + the 72 new; every other count unchanged |
| `tests/geometry tests/augmentation tests/onnx --device=cuda --dtype=float32` | **4760 / 3307 / 69 / 36 / 11** | identical to base, all five |
| `tests/core/test_small_linalg.py --device=cuda --dtype=float32,float64` | 72 passed | |
| same, `--tf32` | 72 passed | |
| same, `--dtype=float16,bfloat16 --isolate-half-precision` | 72 passed (2m08s) | the varied-batch scaling survives half precision on GPU |
| `KORNIA_TEST_OPTIMIZER=inductor`, cuda, f32/f64 | 1298 passed, 0 failed | |
| CPU f32/f64 and f16/bf16 | 72 + 72 passed | |
| the 10 ONNX items, `--device=cuda -rs` | **10 passed, 0 skipped** | the new torch ≥ 2.6 gate hides nothing on 2.9.1 |

Bit-parity re-checked on GPU and extended to bfloat16 this time — 512 random 3×3 matrices through `_inverse_3x3_closed_form` and `_torch_inverse_cast`, `torch.equal` against base: identical in **float32, float64, float16 and bfloat16**. `ruff check` / `ruff format --check` clean; `ty check kornia` still just the pre-existing `lightglue.py` diagnostic.

### Nit 2 fix — verified, and the docstring claim holds

```
batch=(2,):   all slices equal slice 0 -> False   cond numbers: [3.0564, 3.0564]
batch=(2, 3): all slices equal slice 0 -> False   cond numbers: [3.0564, 3.0564, 3.0564, 3.0564, 3.0564, 3.0564]
```

Slices differ, condition number identical across every entry — exactly what the new comment claims.

### Nit 1 fix works, but the denylist is too loose — `Sqrt` is a false positive

The direction is right and I like it better than my own suggestion: it covers the successful-export case that the raise cannot. Executed against the helper as merged:

```
real scalar graph        PASSES
real cross graph         PASSES
MatMul                   FAILS -> ['MatMul']            <- intended
aten_linalg_inv          FAILS -> ['aten_linalg_inv']   <- intended
Det                      FAILS -> ['Det']               <- intended
```

But `_LINALG_OP_MARKERS` is matched as a bare substring, and two of the markers are two letters long. Against the standard ONNX op set:

```
Sqrt              FAILS -> ['Sqrt']              ("qr" in "sqrt")
Relu              FAILS -> ['Relu']              ("lu" in "relu")
LeakyRelu         FAILS -> ['LeakyRelu']
PRelu             FAILS -> ['PRelu']
Selu / Elu / Celu FAILS
ThresholdedRelu   FAILS -> ['ThresholdedRelu']
```

`Sqrt` is the one that matters: it is basic arithmetic, exactly the category this assertion exists to *allow*. Nothing in these kernels emits it today, so this is latent rather than red — but the helper reads like a reusable check, and the first kernel that normalizes anything, or the first person who copies `_assert_basic_arithmetic_only` into another test file, gets a failure naming `Sqrt` as a "linalg-shaped op".

Anchoring the short markers fixes it without giving up the substring behavior that catches `aten_linalg_inv`:

```python
import re

# Long markers are unambiguous as substrings; the short ones (det, lu, qr, svd) are anchored to a
# name boundary, or "qr" matches Sqrt and "lu" matches the whole Relu family.
_LINALG_OP_RE = re.compile(
    r"(?:linalg|inverse|matmul|gemm|einsum)|(?:^|_)(?:det|lu|svd|qr|cholesky|eig)(?:$|_)", re.IGNORECASE
)

def _assert_basic_arithmetic_only(ops):
    offenders = sorted(op for op in ops if _LINALG_OP_RE.search(op))
    assert not offenders, f"linalg-shaped ops in the exported graph: {offenders} (full set: {sorted(ops)})"
```

Ran both versions over the same two corpora:

```
                     current   anchored
must-catch (14 names, incl. Det, Gemm, MatMul, QLinearMatMul, LU,
            aten_linalg_inv, aten_lu_factor, aten_linalg_svd/qr/cholesky,
            Inverse, aten_inverse, Einsum)
                     14/14     14/14
false positives      8         0
```

Same catches, no false positives.

### Two Copilot points that are still open

Both are cases where `tests/filters/test_gaussian.py` — the precedent `_require_exporter`'s own docstring cites — does more than was carried across.

**1. The three ONNX tests are missing `@pytest.mark.device_agnostic`.** This is a live repo convention for exactly this shape of test: `tests/onnx/*` set it as a module-level `pytestmark`, `tests/augmentation/test_onnx_export.py` marks its export tests, and in `test_gaussian.py` the mark sits directly above the version gate that was copied (line 349 vs. lines 354–356). Measured on this branch:

```
tests/core/test_small_linalg.py --device=cuda   -> 10 ONNX items collected and run
tests/augmentation/test_onnx_export.py --device=cuda
  -> "deselected 28 device_agnostic test(s): they exercise CPU-only code and CPU is not
      in --device."  28 deselected in 0.01s
```

So a CUDA-only or MPS-only job re-runs 10 CPU-only export items that every other ONNX suite in the repo skips. They pass on GPU — I ran them — it is the convention and the accelerator-job cost, not correctness.

**2. The legacy branch still passes `dynamo=` unconditionally.** `_require_exporter` gates only `use_dynamo_exporter=True` on torch ≥ 2.6; all three tests then call `torch.onnx.export(..., dynamo=use_dynamo_exporter)` regardless (lines 180, 247, 268). On torch 2.0–2.4 the kwarg does not exist — that is `_require_exporter`'s own docstring ("``dynamo=`` first exists in torch 2.5") and the reason `test_onnx_export_legacy` writes:

```python
if torch_version_ge(2, 5, 0):
    export_kwargs["dynamo"] = False
```

So on 2.0–2.4 the `torchscript` params would raise `TypeError` rather than skip — except in `test_cross_kernel_onnx_export_is_informational`, whose bare `except Exception` would swallow it into a skip that blames `cross` for not lowering. CI runs 2.5.1 and 2.9.1, so CI stays green either way; kornia declares `torch>=2.0.0`, so a contributor on an older pin does not. **I did not measure this** — no 2.0–2.4 environment here — I am reading the repo's own guard and your docstring. Building `dynamo` into a kwargs dict the way `test_onnx_export_legacy` does would settle it.

### The grid finding — happy either way

You or me, whichever is easier; here are the numbers so whoever opens it can paste them. torch 2.9.1+cu128, RTX 4090, on `main` as well as this branch:

```
tests/geometry/test_grid.py::test_normalized_meshgrid_trace_matches_eager_at_unrepresentable_sizes
  --device=cuda --dtype=float32  -> 8 failed, 2 passed   (258/300/1000/2050/3000 x 2d/3d;
                                                          258-2d and 258-3d are the two that pass)
  --device=cpu  --dtype=float32  -> 10 passed

E   Mismatched elements: 92 / 2400 (3.8%)
E   Greatest absolute difference: 1.1920928955078125e-07 at index (0, 150, 0, 1)
E   Greatest relative difference: 3.564426879165694e-05 at index (0, 150, 0, 1)
```

One ULP of float32, and the test asserts `atol=0.0, rtol=0.0`. So it looks like traced-vs-eager kernel/fusion divergence on CUDA rather than the rounding bug the test was written for — the bitwise assertion is what makes it visible.

### Verdict, unchanged

Approve on CUDA: bit-identical outputs across all four float dtypes, identical pass/skip/xfail counts against the base revision, green under half precision, TF32, inductor and both exporters, and the library-side changes in these three commits are provably prose-only. The denylist anchoring is the one I would land before merge, since it is three lines and the current form calls `Sqrt` a linalg op; the two Copilot points are convention and older-torch reach.

ducha-aiki:
All three reproduced before fixing and verified to discriminate after. Pushed as `4c5437b2`. The AST-equality check across `9f2dcb68..db89866b` is a nice way to make "prose only" checkable rather than asserted — I'll steal that.

## Nit 1 regression — you're right, and it was mine

Confirmed against the helper as merged:

```
Sqrt Relu LeakyRelu PRelu Selu Elu Celu ThresholdedRelu   -> all 8 FAIL (false positives)
Add Mul                                                    -> pass
```

`Sqrt` being rejected as "linalg-shaped" is the bad one, exactly as you say — it is the category the check exists to allow, and I introduced it while fixing a different toothlessness. Your anchored regex applied as written. Re-ran both corpora:

```
must-catch  (Det, Gemm, MatMul, QLinearMatMul, LU, aten_linalg_inv, aten_lu_factor,
             aten_linalg_svd/qr/cholesky/det, Inverse, aten_inverse, Einsum)   14/14
false positives over 20 benign ops (incl. the whole Relu family, Sqrt,
             and every op the four real graphs emit)                            0
```

## Nit 2 (device_agnostic) — confirmed, convention followed

Measured before:

```
tests/core/test_small_linalg.py --device=mps  -> 10 ONNX items collected and RUN
tests/augmentation/test_onnx_export.py --device=mps -> 28 deselected
```

And after the mark, the whole file on mps: **81 passed, 8 skipped, 10 deselected**. You're right that `test_gaussian.py` puts the mark directly above the version gate I copied and I took only half of it.

## Nit 3 (unconditional `dynamo=`) — real, and the boundary is 2.4, not 2.5

You flagged this as unmeasured, so I built the environments. `torch.onnx.export` signature, and a real call:

| torch | `dynamo=` in signature | `dynamo=False` call |
|---|---|---|
| 2.0.1 | **no** | `TypeError: export() got an unexpected keyword argument 'dynamo'` |
| 2.2.2 | **no** | same `TypeError` |
| 2.4.1 | **yes** | OK |
| 2.5.1 | yes | OK |
| 2.9.1 | yes | OK |

So the bug is real and your reasoning was right; the boundary is one release earlier than both your "2.0–2.4" and `_require_exporter`'s own docstring ("`dynamo=` first exists in torch 2.5"). I fixed the docstring too. Gated at 2.5 anyway, matching `test_onnx_export_legacy` — conservative by one release and it costs nothing.

Your point about the bare `except Exception` swallowing it was the sharper half. `TypeError` now re-raises, so an API misuse can never masquerade as "`cross` does not lower here".

**One caveat on the blast radius.** The window is narrower than `torch>=2.0.0` suggests: kornia does not import at all on 2.0.1 —

```
kornia/core/_compat.py: AttributeError: module 'torch' has no attribute 'compiler'
```

— so the affected range is roughly 2.2–2.3, where kornia imports and the kwarg is absent. Still worth fixing, and it is pre-existing and out of scope here, but the "contributor on an older pin" is already blocked below ~2.2 for unrelated reasons.

## Verification after the fixes

| env | result |
|---|---|
| torch 2.9.1 CPU, characterization set | 1253 passed / 11 skipped / 15 xfailed — counts unchanged |
| torch 2.9.1 CPU, f32/f64/f16/bf16 | 126 passed |
| torch 2.5.1 | 76 passed / 5 skipped |
| torch 2.4.1 | 40 passed / 5 skipped — no `TypeError`, dynamo variants gated |
| MPS f32/f16/bf16 | 81 passed / 8 skipped / 10 deselected |
| `pre-commit-all` | clean |

## The grid finding

I'll open it — you have already done the GPU work twice over. Numbers pasted from your comment, credited to your run, and I'll note it reproduces on `main` so it is not attributed to this PR.

ducha-aiki:
Grid finding filed as #4195, with your numbers and machine credited and a note that it reproduces on `main` so it is not attributed here.

I framed the fix as a decision rather than a default: the `atol=0.0, rtol=0.0` is what makes the bfloat16 pin meaningful, so quietly loosening it to a default tolerance would weaken the thing the test exists for. Three options listed — dtype-conditional tolerance keeping the half arms exact, keep-exact-and-document if traced `create_meshgrid` really can differ by a ULP on CUDA, or a strict xfail.

ducha-aiki:
## Round 3 on `4c5437b2` — all three findings fixed and verified, nothing blocking

`4c5437b2` touches **only** `tests/core/test_small_linalg.py`, so the library side is unchanged since `db89866b`, whose `kornia/` tree I already showed is AST-identical to `9f2dcb68` with docstrings stripped. Last round's CUDA bit-parity therefore carries without re-deriving it. Re-ran the GPU matrix anyway. Same machine: RTX 4090, torch 2.9.1+cu128, CUDA 12.8, Python 3.11.

### The three fixes, executed rather than assumed

**Anchored denylist.** Ran the merged `_LINALG_OP_RE` over both corpora from last round:

```
must-catch (Det, Gemm, MatMul, QLinearMatMul, LU, Einsum, Inverse, aten_inverse,
            aten_linalg_inv/det/svd/qr/cholesky, aten_lu_factor)   14/14 caught
false positives (Sqrt, Relu, LeakyRelu, PRelu, Selu, Elu, Celu,
                 ThresholdedRelu, + 14 arithmetic ops)             0
real scalar graph PASS   real cross graph PASS   MatMul -> FAIL ['MatMul']
```

`Sqrt` and the whole `Relu` family are allowed again; every intended catch survives.

**`device_agnostic`.** Works exactly as intended, with the yellow banner firing:

```
tests/core/test_small_linalg.py --device=cuda --dtype=float32,float64
  -> "deselected 10 device_agnostic test(s)..."   62 passed, 10 deselected
same + --run-device-agnostic                       72 passed
--device=cpu --dtype=float32,float64               72 passed   (they still run where they should)
```

**`_export_kwargs`.** On 2.9.1 it yields `{'opset_version': 18, 'dynamo': False}` / `{... 'dynamo': True}`, and the `except TypeError: raise` sits ahead of the `except Exception` skip, so an API misuse can no longer masquerade as "cross does not lower here". I could **not** check the 2.0.1 / 2.2.2 / 2.4.1 measurements in the docstring — no such environment here — so those stand on your run, not mine.

### CUDA runs, `4c5437b2`

| run | branch | base |
|---|---|---|
| 4 paths, cuda f32/f64, default | **1236 passed**, 18 skipped, **38 deselected**, 15 xfailed | 1174, 18, 28, 15 |
| 4 paths, cuda f32/f64, `--run-device-agnostic` | **1273 passed**, 19 skipped, 0 deselected, 15 xfailed | 1201, 19, 0, 15 |

The forced run is the clean comparison: **exactly +72 passed, every other count identical**. The default run splits the same 72 as +62 passed / +10 deselected — the newly marked items, and nothing else moved.

Also green: new file at cuda f32/f64 (72 forced), `--tf32` (72), **cuda float16/bfloat16 under `--isolate-half-precision` (72, 2m14s)**, cpu f32/f64 (72), cpu half (72), the 10 ONNX items on cpu (10 passed, 0 skipped). `KORNIA_TEST_OPTIMIZER=inductor` on cuda: 1298 passed, 0 failed. Wide sweep `tests/geometry tests/augmentation tests/onnx --device=cuda --dtype=float32`: **4760 / 3307 / 69 / 36 / 11 on both trees**, identical. `ruff check` + `ruff format --check` clean; `ty check kornia` still only the pre-existing `lightglue.py` diagnostic.

### Optional, none of it blocking

**1. The `device` fixture still doubles those items wherever CPU is in the matrix.** The mark fixes the accelerator-only job, which was the point. What remains is that the three tests still request `device`, so they carry a device dimension the precedent does not:

```
tests/core/test_small_linalg.py  -k "onnx or export" --device=cpu,cuda  -> 20 items (10 cpu + 10 cuda)
tests/filters/test_gaussian.py   -k "onnx_export"    --device=cpu,cuda  ->  2 items (float32 only)
```

`test_onnx_export_legacy` / `_modern` take only `dtype` and build on CPU, so no device dimension exists to deselect. **Practical impact today: zero** — I checked `.github/workflows/tests.yml`, every job sets a single `KORNIA_TEST_DEVICE`, so no CI matrix contains two devices. It is a local `--device=cpu,cuda` run and a structural difference from the cited precedent, nothing more. Dropping the `device` parameter and building on CPU would collapse it to 10 items in every matrix.

**2. `_export_kwargs` is only correct because of its caller.** `_export_kwargs(True)` on torch < 2.5 returns kwargs with no `dynamo` at all — a legacy export under a `torchexport` test id. Unreachable today (`_require_exporter` skips dynamo below 2.6, and 2.6 > 2.5), so this is future-proofing, not a defect. One line inside the helper makes it self-consistent regardless of what the caller's gate later becomes:

```python
assert not use_dynamo_exporter or "dynamo" in kwargs, "dynamo requested but unavailable on this torch"
```

**3. Informational on the last standing Copilot sub-claim.** Your refutation holds, and it holds on GPU too — I re-measured with the merged `_export_kwargs`:

```
device=cpu  dynamo=False  return=NoneType     bytes= 6767  parseable=True
device=cpu  dynamo=True   return=ONNXProgram  bytes=46330  parseable=True
device=cuda dynamo=False  return=NoneType     bytes= 6767  parseable=True
device=cuda dynamo=True   return=ONNXProgram  bytes=46330  parseable=True
```

Worth recording though: this file is the only place in the repo that passes a buffer *and* `dynamo=True`. `tests/onnx/test_export_coverage.py:77-78` takes the other route — `program = torch.onnx.export(module, inputs, dynamo=True, ...)` then `program.save(buf)`. Both work on 2.9.1. The combination here is exercised on 2.9.1 only: the 2.5.1 legs skip it at the 2.6 gate, so torch 2.6–2.8 is verified by nobody. Matching `test_export_coverage.py`'s form would remove that gap; it is a preference, not a defect.

**4.** `def _export_kwargs(...) -> dict:` — `test_gaussian.py` spells the same thing `dict[str, Any]`. Cosmetic.

### Verdict

Approve. Bit-identical GPU outputs across float32/float64/float16/bfloat16 (established last round, and this commit is test-only), `--run-device-agnostic` parity of exactly +72 passed against the merge-base with every other count unchanged, green under half precision, TF32, inductor, both exporters and both devices, and all three of my findings fixed with the fixes shown to discriminate. Nothing above needs to land before merge.

Thanks for filing #4195 — framing the ULP question as a decision rather than silently loosening the tolerance is the right call, and the bfloat16 arm is exactly what the exact comparison is protecting.

PR #4194, opened 2026-09-03
fix(geometry): preserve list padding metadata in Boxes.__getitem__

### LINKED ISSUE #4179: Bug: `Boxes.__getitem__` drops the list padding metadata, so a sliced list-backed object exports its zero pad rows as boxes
## Summary

`Boxes.__getitem__` (`kornia/geometry/boxes.py:237` at `e0969438`) rebuilds the result from the raw padded data and copies only `_mode`:

```python
def __getitem__(self, key):
    new_box = type(self)(self._data[key], False)
    new_box._mode = self._mode
    return new_box
```

`_N`, the per-element padding counts recorded when the object was built from a list, is not carried over, so the sliced object no longer knows it is list-backed. `to_tensor` then returns the **padded tensor** instead of the trimmed list, and the zero pad rows come back as boxes. `clone`, `merge` and `index_put` all preserve `_N`; only indexing loses it. Nothing raises.

## Reproduction

```python
import torch

first = torch.tensor([[[1.0, 2.0], [4.0, 2.0], [4.0, 3.0], [1.0, 3.0]]])   # 1 box
lb = Boxes([first, torch.cat([first, first])])                              # _N == [1, 0], data (2, 2, 4, 2)

print(lb._N, [t.tolist() for t in lb.to_tensor("xyxy")])
sliced = lb[0:1]
print(sliced._N, type(sliced.to_tensor("xyxy")).__name__, sliced.to_tensor("xyxy").tolist())
print([t.tolist() for t in sliced.get_boxes_shape()])
print(lb.clone()._N, lb.index_put((torch.tensor([0]), torch.tensor([0])), torch.zeros(4, 2))._N)
```

## Observed

```
[1, 0] [[[1.0, 2.0, 5.0, 4.0]], [[1.0, 2.0, 5.0, 4.0], [1.0, 2.0, 5.0, 4.0]]]
None Tensor [[[1.0, 2.0, 5.0, 4.0], [0.0, 0.0, 1.0, 1.0]]]
[[[2.0, 1.0]], [[4.0, 1.0]]]
[1, 0] [1, 0]
```

The unsliced object exports a list with row 0 trimmed to its one real box. `lb[0:1]` exports a `(1, 2, 4)` tensor in which the phantom padding box `[0, 0, 1, 1]` sits next to the real one, and `get_boxes_shape` reports it as a 1-by-1 box. `lb[1:2]` happens to look right only because that row's pad count was 0, so whether a slice is corrupt depends on how ragged the original list was.

## Expected

`__getitem__` carries the padding metadata through: `new_box._N = self._N[key]` for a slice or index tensor on the batch axis (a plain `int` key drops the batch axis and returns an unbatched object, which has no `_N`), or, if indexing a list-backed object is not meant to be supported, it raises rather than returning a silently padded tensor. The batched export should be the same list as `[lb.to_tensor("xyxy")[i] for i in range(...)]` restricted to the selected rows.

## Context

Found in round 9 of #4060 (bbox/boxes convention blocks, batch 4a), which documents the trimmed-list contract of `to_tensor(as_padded_sequence=False)` and `get_boxes_shape`'s padding entries; that PR is documentation and tests only, so this is out of scope there. Same family as #4175 (`merge` keeps stale `_N`): one is stale metadata, this one is dropped metadata, and one PR that decides how `_N` propagates can close both. Part of the #4020 audit.

## Environment

Alphaxiaoteng:
## Description

Fixes #4179.

`Boxes.__getitem__` rebuilt from the padded tensor and only copied `_mode`, dropping `_N`. Sliced list-backed objects then exported pad rows as boxes via `to_tensor()`.

Preserve `_N` for slice / tensor indexing; int keys still drop the batch axis (no `_N`).

## Test plan

```bash
pytest tests/geometry/test_boxes.py::TestBoxes2D::test_getitem_preserves_list_padding_metadata -q --device=cpu
```

1 passed.

## AI assistance

AI-generated fix and regression test. I reviewed the diff and ran the focused test locally.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -303,6 +303,16 @@ class Boxes:
     def __getitem__(self, key: slice | int | torch.Tensor) -> Boxes:
         new_box = type(self)(self._data[key], False)
         new_box._mode = self._mode
+        # Preserve list-padding metadata so to_tensor() still trims pad rows (#4179).
+        # An int key drops the batch axis and returns an unbatched object (no _N).
+        if self._N is not None and not isinstance(key, int):
+            if isinstance(key, torch.Tensor):
+                indices = key.tolist()
+                if isinstance(indices, int):
+                    indices = [indices]
+                new_box._N = [self._N[i] for i in indices]
+            else:
+                new_box._N = list(self._N[key])
         return new_box
 
     def __setitem__(self, key: slice | int | torch.Tensor, value: Boxes) -> Boxes:
--- a/tests/geometry/test_boxes.py
+++ b/tests/geometry/test_boxes.py
@@ -220,6 +220,18 @@ class TestBoxes2D(BaseTester):
         self.assert_close(boxes.data, boxes_before, atol=0.0, rtol=0.0)
         self.assert_close(replacement.data, replacement_before, atol=0.0, rtol=0.0)
 
+    def test_getitem_preserves_list_padding_metadata(self, device, dtype):
+        # #4179: slicing a list-backed Boxes must keep _N so pad rows stay trimmed.
+        first = torch.tensor([[[1.0, 2.0], [4.0, 2.0], [4.0, 3.0], [1.0, 3.0]]], device=device, dtype=dtype)
+        lb = Boxes([first, torch.cat([first, first])])
+        assert lb._N == [1, 0]
+        sliced = lb[0:1]
+        assert sliced._N == [1]
+        exported = sliced.to_tensor("xyxy")
+        assert isinstance(exported, list)
+        assert len(exported) == 1
+        assert exported[0].shape == (1, 4)
+
     def test_smoke(self, device, dtype):
         def _create_tensor_box():
             # Sample two points of the rectangle


ducha-aiki:
## Review of #4194 at `f90db50c` — CHANGES REQUESTED, one key type is left corrupting

Setup: worktree at the PR head, `kornia/` swapped to the merge-base `4d0783bf` for every A/B pass. `~/dev/kornia/.venv/bin/python`, **torch 2.14.0** (the `.venv` moved off 2.9.1 with #4197), CPU + MPS.

The diagnosis and the direction are right, and the happy paths are genuinely fixed. But the metadata is selected with a hand-rolled index walk that disagrees with how torch indexed the data one line above, and for one key type that turns #4179's *phantom* boxes into *missing* boxes — strictly worse than `main`.

### What works

Every key type the tests cover behaves, and `_N` tracks `_data` exactly. `Boxes([1 box, 2 boxes, 3 boxes])` → `_N == [2, 1, 0]`, `data (3, 3, 4, 2)`:

```text
lb[0:2]  slice           _N=[2, 1]  ->  list [[1, 4], [2, 4]]
lb[1:]   slice           _N=[1, 0]  ->  list [[2, 4], [3, 4]]
lb[::2]  strided         _N=[2, 0]  ->  list [[1, 4], [3, 4]]
lb[tensor([0, 2])]       _N=[2, 0]  ->  list [[1, 4], [3, 4]]
lb[tensor([2, 0])]       _N=[0, 2]  ->  list [[3, 4], [1, 4]]     <- reordering handled
lb[tensor([-1])]         _N=[0]     ->  list [[3, 4]]             <- negative index handled
lb[0]    int key         _N=None    ->  Tensor [3, 4]             <- documented, unchanged
```

Fail-on-base: your test **fails on `4d0783bf`** (`assert None == [1]`) and passes at head. Whole file **126 passed** CPU float32+float64, **63 passed / 5 skipped** MPS float32.

`Boxes3D` needs nothing, in case you were wondering: its `__init__` takes no list, it never sets `_N`, and its `__getitem__` has nothing to carry. The fix is correctly scoped to `Boxes`.

### The blocker — a boolean mask key

`key.tolist()` on a bool tensor gives `[True, False, True]`, and `self._N[True]` is `self._N[1]`. So the comprehension indexes with the mask's *values* instead of the positions it selects, and produces a list of the wrong length with the wrong contents. `to_tensor` then `zip`s that against a shorter batch and silently truncates.

Three rows holding 1, 2 and 3 real boxes; `lb[tensor([True, False, True])]` must equal `lb[tensor([0, 2])]`, and `torch.equal` confirms the two agree on `_data`:

```text
truth, rows 0 and 2   [[1, 0, 3, 2]]                            [[4, 0, 6, 2], [5, 0, 7, 2], [6, 0, 8, 2]]

base 4d0783bf         [[1, 0, 3, 2], [0,0,1,1], [0,0,1,1]]      [[4, 0, 6, 2], [5, 0, 7, 2], [6, 0, 8, 2]]
head f90db50c         [[1, 0, 3, 2], [0,0,1,1]]                 [[4, 0, 6, 2]]

_N at head: [1, 2, 1]  (length 3, against a batch of 2)
```

Read the second column. On `main` the boolean mask hits the #4179 wart — row 0 comes back with two phantom pad boxes — but **row 2 is complete**. At this head, row 2 loses two real boxes. That is the same failure mode the issue is about, pointed the other way, and it is the more damaging direction: a phantom box is visible, a dropped box is not.

A 0-dim tensor key has a milder version of the same problem — `torch.tensor(1)` drops the batch axis like `1` does, but the `isinstance(key, int)` guard does not recognise it, so the result is an unbatched object that still claims `_N = [1]` and exports a `list` where the int key exports a `Tensor`. Two spellings of one index, two container types.

Neither is reachable from inside kornia — I checked, no in-tree caller indexes a `Boxes` object with a mask (`mosaic.py:130` indexes a plain tensor, not a `Boxes`). So the reachability is user-facing only, which is exactly the reachability #4179 itself has.

### Fix — replace the nine lines with three, and let torch do the selecting

The invariant you want is "select `_N` the same way torch selected the batch axis of `_data`". Say that literally and every key type follows by construction, including ones neither of us has thought of:

```python
        # Preserve list-padding metadata so to_tensor() still trims pad rows (#4179). Select it
        # with the same key torch used on the batch axis, so slices, index tensors and boolean
        # masks all agree with the data by construction. A key that drops the batch axis (an int
        # or a 0-dim tensor) yields an unbatched object, which has no _N.
        if self._N is not None:
            selected = torch.as_tensor(self._N)[key]
            new_box._N = None if selected.dim() == 0 else [int(n) for n in selected]
```

Measured in the worktree — all eight key types, with the two previously-broken rows now correct:

```text
slice 0:2         ([2, 1], 'list',   [[1, 4], [2, 4]])
slice ::2         ([2, 0], 'list',   [[1, 4], [3, 4]])
idx [0, 2]        ([2, 0], 'list',   [[1, 4], [3, 4]])
idx [2, 0]        ([0, 2], 'list',   [[3, 4], [1, 4]])
idx [-1]          ([0],    'list',   [[3, 4]])
bool [T, F, T]    ([2, 0], 'list',   [[1, 4], [3, 4]])     <- now equals idx [0, 2]
0-dim tensor(1)   (None,   'Tensor', [3, 4])               <- now equals int 1
int 1             (None,   'Tensor', [3, 4])
```

`tests/geometry/test_boxes.py`: **126 passed** with the replacement in place, so nothing else depended on the old walk.

### Two asks

1. Take the replacement above (or an equivalent that keys off torch's own indexing), and extend your test with the boolean-mask case asserting it equals the index-tensor case. That assertion is the one that discriminates: it passes on `main`'s padded output only by accident of shape, and fails on this head.
2. **Add a CHANGELOG entry** under `## Unreleased` → `### Bug fixes`, citing #4194 (`AGENTS.md` requires one per user-visible change).

The rest of the PR — the scoping to `Boxes`, the int-key carve-out, the fail-on-base test — is right, and the diagnosis in the description is accurate. Only the selection mechanism needs replacing.

ducha-aiki (changes requested):
Setting the formal review state to match the existing review: **changes requested** on head `f90db50c`, which is unchanged since that review.

The outstanding items are in https://github.com/kornia/kornia/pull/4194#issuecomment-5533731476 (boolean-mask indexing still corrupts the padding metadata and loses real boxes; CHANGELOG entry).

PR #4196, opened 2026-09-03
docs: remove multi-framework (Ivy) support after testing found it broken

ducha-aiki:
## Summary

`kornia.to_tensorflow()`, `kornia.to_jax()` and `kornia.to_numpy()` lazily transpiled the library to other frameworks through the third-party [Ivy](https://github.com/ivy-llc/ivy) project. Testing in September 2026, following the exact examples the docs showed, found all three broken out of the box:

* All three crash immediately with `TypeError: unhashable type` when kornia is checked out into a directory whose path contains the substring `kornia` — the default `git clone` directory name. Ivy's transpiler decides what counts as a kornia submodule with a plain substring check against each module's file path, so it also recurses into unrelated parts of PyTorch's dependency tree and hits an object it can't hash.
* Even from a differently named checkout, `to_tensorflow()` segfaults during transpilation if `transformers` is also installed — true by default for kornia contributors, since both packages shipped in the same `dev` extra.
* `to_numpy()` transpiles but fails at call time on a missing `np.bfloat16`.
* `to_jax()` needs `flax>=0.8.0`, which was never documented, and then fails on a `jaxlib`/`numpy` API-compatibility break.

None of these are kornia bugs — they're all in Ivy or its compatibility with current TensorFlow/JAX/NumPy releases. Ivy itself shows little ongoing maintenance as of testing: one commit in the previous twelve weeks (a rebrand, not a fix), last PyPI release over a year old, close to a thousand open issues including unmerged fixes for exactly this kind of version drift.

## Changes

* Removes `kornia.to_tensorflow()`, `kornia.to_jax()`, `kornia.to_numpy()`, the transpiler implementation, and the `ivy` `dev`/`docs` dependency.
* `kornia.transpiler` now raises a clear `AttributeError` pointing at the docs instead of importing successfully and failing later (mirrors the `kornia.utils` deprecation-shim pattern, so `.github/scripts/check_import_surface.py` treats this as an intentional, documented removal rather than a silent break).
* Drops the Ivy advertising from the README and the docs landing page's "Why Kornia?" tab.
* Rewrites `docs/source/get-started/multi-framework-support.rst` into a record of what was tested and why the feature was removed, for anyone landing there from an old link.
* Drops the now-inaccurate `kornia.transpiler` mention from the stability policy page.
* Strips a malformed embedded ICC profile from two static logo PNGs that libpng was flagging as non-conformant (unrelated cleanup, bundled in since it's tiny).

## Test plan

- [x] `pixi run lint`
- [x] `pixi run typecheck`
- [x] `pixi run doctest`
- [x] `pixi run test-quick` (8956 passed, 0 failed)
- [x] `pixi run build-docs` (clean, `-W` warnings-as-errors)
- [x] `.github/scripts/check_import_surface.py` (no fatal removals)
- [x] Manually reproduced all four bugs above against kornia 0.9.0rc1, ivy 1.0.0.5, torch 2.14.0, tensorflow 2.21.0, jax/jaxlib 0.10.2, both in this checkout and in a differently-named worktree to isolate the path-substring bug from the others

Written by Claude (Sonnet 5) on behalf of @ducha-aiki

Posted on behalf of @ducha-aiki by Claude (Opus 5, Sonnet 5).


ducha-aiki:
I found one issue that should be addressed before merging:

**P2 — The removal guidance is unreachable from the documented API paths.** Removing the imports from `kornia/__init__.py` means `kornia.to_tensorflow`, `from kornia import to_tensorflow`, and even `kornia.transpiler` after only `import kornia` produce generic Python errors. The helpful message in `kornia/transpiler/__init__.py` appears only after explicitly importing `kornia.transpiler`, which existing users have no reason to do. Please add a top-level `__getattr__` for the three removed functions and continue binding the `transpiler` module, with tests covering the formerly documented access paths.

Verification performed: 35 focused API/import-surface tests passed; `git diff --check` passed; all required GitHub checks are green; and the PNG replacements are pixel-identical with only metadata removed.

Written by Codex (gpt-5) on behalf of @ducha-aiki

ducha-aiki:
Fixed in `83e76d66`.

**P2 — removal guidance unreachable from the documented API paths.** Confirmed against the branch before the fix: `kornia.to_tensorflow`, `kornia.to_jax` and `kornia.to_numpy` all raised a bare `module 'kornia' has no attribute ...`, and `kornia.transpiler` did not resolve at all after a plain `import kornia` — the message in `kornia/transpiler/__init__.py` was reachable only by code that imported the submodule by name, which nobody had a reason to do.

`kornia/__init__.py` now keeps the `transpiler` module bound and adds a module-level `__getattr__` for the three removed names:

```
>>> import kornia
>>> kornia.to_tensorflow
AttributeError: kornia.to_tensorflow() was removed: testing found the Ivy-powered
multi-framework transpiler unreliable, so it is no longer part of kornia. See
https://kornia.readthedocs.io/en/latest/get-started/multi-framework-support.html
>>> kornia.transpiler.to_jax
AttributeError: kornia.transpiler.to_jax() was removed: ... (same, requalified)
```

The message text is shared between the two shims so the two spellings cannot drift, and each quotes the path back as the caller wrote it. The names stay genuinely removed — `hasattr(kornia, "to_jax")` is `False`, `getattr(kornia, "to_jax", default)` returns the default, and they are not in `dir(kornia)`; the `__getattr__` is guidance, not a working entry point. Ordinary lookup is untouched: an unrelated missing attribute still gets Python's standard message.

**One path I could not make helpful, deliberately.** `from kornia import to_tensorflow` still reports the generic `ImportError: cannot import name 'to_tensorflow' from 'kornia'`. CPython's `from X import Y` suppresses the `AttributeError` a module `__getattr__` raises (`PyObject_GetOptionalAttr` in `import_from`) and substitutes that message, and a subclass of `AttributeError` is suppressed too. The only way to reach the guidance there is to keep the name bound to a stub object, which would un-remove it and put it back on the import surface — not what this PR is doing. NumPy 2.0's removed attributes behave identically (`from numpy import float_` → the same generic `ImportError`), and this form never appeared in kornia's docs, which always showed `import kornia` followed by `kornia.to_tensorflow()`. The tradeoff is recorded in the new test module's docstring so the next reader does not have to rediscover it.

**Tests** — `tests/test_transpiler_removed.py` (12 cases) covers the formerly documented access paths:

* the top-level attribute and the `kornia.transpiler` attribute, per removed name, each asserting the message names the call and links the docs page;
* a **fresh interpreter** that only runs `import kornia` and then reaches for both. In-process this would prove nothing — any other test importing `kornia.transpiler` binds the attribute on the parent package as a side effect, so the assertion would pass even if `kornia/__init__.py` stopped importing it;
* the names are absent rather than shimmed back to life (`hasattr` / `getattr` default / `dir`);
* an unrelated attribute still produces the standard `module 'kornia' has no attribute ...`;
* the docs page the message links resolves to a real source file, with the path **derived from the URL constant** rather than typed out again, so a rename on either side fails the test.

7 of the 12 fail on the pre-fix branch (both message tests for all three names, plus the fresh-interpreter one); the other 5 are regression pins for behavior the fix must not change.

Also updated the last paragraph of `get-started/multi-framework-support.rst` and the CHANGELOG entry, both of which claimed only `kornia.transpiler` raised the clear error.

**Verification** — `pixi run pre-commit-all` (all hooks pass), `pixi run typecheck` (1 diagnostic, pre-existing and unrelated: the deprecated `torch.cuda.amp.custom_fwd` in `kornia/feature/lightglue.py`), `pixi run doctest` (649 passed, 11 skipped), `pixi run build-docs` (`-W`, build succeeded), `.github/scripts/check_import_surface.py --base-ref origin/main` (exit 0; `kornia/__init__.py` now carries a module-level `__getattr__`, so the script classifies it as an intentional deprecation shim, same as `kornia.utils`), and `tests/test_transpiler_removed.py tests/test_import.py tests/test_api_surface.py tests/smoke_test.py` (29 passed).

ducha-aiki:
Correction to my previous comment: `tests/test_transpiler_removed.py` has been removed again in `4bba01d9`, so the **Tests** section above no longer describes the branch. Pinning the prose of an error message is maintenance cost with no future — the shims exist to be deleted once the removal is old news, and the tests would have to go with them.

That means the "with tests covering the formerly documented access paths" half of the review ask is deliberately declined, not overlooked. The code half stands: `kornia.to_tensorflow`, `kornia.to_jax`, `kornia.to_numpy` and their `kornia.transpiler` counterparts all report what was removed and link the docs page, and `kornia.transpiler` resolves after a plain `import kornia`. Re-verified on `4bba01d9` after the removal; everything else in the previous comment (the `from kornia import ...` limitation, the docs and CHANGELOG updates, `pre-commit-all`, `typecheck`, `doctest`, `build-docs -W`, `check_import_surface.py`) is unchanged and still accurate. `tests/test_import.py tests/test_api_surface.py tests/smoke_test.py` — 17 passed.

PR #4197, opened 2026-09-03
ci: move the torch baseline to 2.5.1-2.14.0

ducha-aiki:
## Summary
- Raise the declared torch floor to 2.5.1 (pyproject/README/installation.rst/CHANGELOG)
- Move PR CI's newest torch pin from 2.9.1 to 2.14.0 (matrix + dynamo job)
- Retire torch 2.1.2-2.4.0 from the scheduled ubuntu matrix, add 2.6.0 and 2.14.0; add 2.14.0 to the scheduled Windows and macOS matrices
- Targeted uv.lock bump: torch 2.14.0 + onnx 1.22.0 / onnx-ir 1.0.0 / onnxruntime 1.29.0 / onnxscript 0.7.1, nothing else
- Add a py314 pixi environment
- Add a weekly upstream-probe workflow (floating-stable torch, no nightly)
- Scope the one real torch-2.14 CPU failure (SAM ONNX export) with a bounded, strict xfail
- Fix `find_homography_dlt(..., solver="lu")` on a minimal four-point sample, which is what torch 2.14 on Windows exposed

## Why
The declared floor (`torch>=2.0.0`) and PR CI's newest pin (2.9.1, ten months old) both
drifted from reality. Measurements (2x2 over {torch 2.9.1, 2.14.0} x {old deps, fresh
resolve}, run on an `origin/main` worktree) found exactly one CPU-side torch-2.14 failure
beyond the dependency stack: a SAM ONNX exporter dispatch bug, xfailed here with `raises=`
scoped to the public, stable `torch.onnx.OnnxExporterError` and an upper version bound so
the failure un-quarantines itself once upstream fixes it.

The new weekly probe rehearses a future bump against floating-stable torch using the
kornia#4154 `--record-known-failures` mechanism. Its two half-precision legs run in record
mode and can never fail on drift alone -- they report through a job-summary receipt and an
uploaded candidate manifest. Its `cpu-float32`, `cpu-float64` and `python314` legs have no
manifest and are, deliberately, red on any new failure.

### The homography fix
Torch 2.14 on Windows returned an all-NaN matrix from `find_homography_dlt(..., solver="lu")`
for a four-point sample. A four-point sample gives eight equations for nine unknowns, so the
normal matrix `A^T W A` is singular and the previous code LU-factored it anyway, relying on
the null direction dominating the solve. This replaces that with an adaptive homogeneous
gauge on the design matrix: the null vector comes from a complete QR, its largest component
fixes the gauge, and the retained 8x8 system is solved for the rest.

That estimator is what `RANSAC(model_type="homography")` calls for every minimal sample, so
this changes homography RANSAC output on every platform, not only where the old formulation
broke down. Over 2048 minimal samples (float32, CPU):

| | before | after |
|---|---|---|
| median transfer residual | 2.29e-5 | 1.91e-6 |
| p99 residual | 6.47e-1 | 2.33e-3 |
| max residual | 4.22e+5 | 4.47e+2 |
| time / call (B=2048) | 3.5 ms | 4.5 ms |
| RANSAC end-to-end (500 pts, 30% outliers) | 11.4 ms | 11.4 ms |

Samples of five or more points keep the previous normal-equation formulation and are
numerically unchanged; ten weighted and unweighted cases across both solvers and both float
dtypes were checked byte-for-byte against `main`.

## Out of scope (separate PRs)
- MPS leg bump + manifest migration onto the kornia#4154 profile mechanism
- Dead version-guard cleanup below the new 2.5.1 floor (the prose stating the old floor is
  corrected here; the now-redundant guards themselves are not removed)
- kornia_rs 0.1.14 API break (independent P1, not depended on here)
- Generalizing the gauge formulation to five-or-more-point samples, whose normal matrix is
  equally singular for exactly-consistent correspondences (`cond(A^T A)` ~1e16-1e17 at N =
  4, 5, 10 and 20; ~4.5e4 once there is 0.5 px of noise). No failure was observed there on
  any platform CI runs, and changing it would move results for every existing caller.

## Test plan
- [x] Full suite on Python 3.14 / torch 2.14.0: 8895 passed, 0 failed
- [x] `uv sync --extra dev --python 3.14 --frozen` succeeds (this is what `pixi run -e py314 install` runs; `pixi install -e py314 --locked` only checks the conda side and would not have caught the lock revert)
- [x] ONNX suites on torch 2.14 + the bumped onnx stack: 228 passed, 1 xfailed
- [x] CPU float32/float64 across `tests/geometry tests/core tests/feature/test_matching.py tests/onnx tests/models`: 5485 passed, 0 failed
- [x] Half-precision failure set A/B'd against `main` on CPU float16 + bfloat16 with a per-test seed:
      exactly the two removed manifest entries fixed, zero new failures. The blocking
      `--verify-known-failures` gate itself only runs on Linux (the manifests carry an `os:`
      header), so it is left to the `half-precision` CI job
- [x] `pixi run pre-commit-all` clean
- [ ] Full PR CI green, including the new 2.14.0 legs
- [ ] `dynamo` job green under torch 2.14.0
- [ ] `upstream_probe.yml` triggered manually once after merge -- `workflow_dispatch` only
      lists workflows present on the default branch, so this cannot be exercised from the PR
      branch


ducha-aiki (commented):
Blocking findings:

1. [P1] .github/workflows/tests.yml:192-194 makes a successful record-mode run fail whenever pytest records any ordinary failure. This step explicitly selects shell: bash, so Actions runs bash with -e and pipefail; the pixi/pytest status therefore propagates through tee. FailureRecorder intentionally leaves a complete run at TESTS_FAILED when failures were recorded (tests/test_half_precision_ci.py:311-315 asserts this), so the half-precision probe turns red instead of publishing drift as a non-blocking receipt. Capture the pytest status and suppress it only when a complete candidate was actually written; recorder aborts, missing receipts, and infrastructure failures must remain non-zero. This is the inverse of the existing Copilot diagnosis: pipefail is already active here.

2. [P1] The new Torch 2.14 Windows coverage exposes an unhandled public-API regression. All three windows-latest/float64 rows (Python 3.11, 3.12, and 3.13) fail tests/geometry/test_homography.py::TestFindHomographyDLT::test_scaled_fixed_points: find_homography_dlt(..., solver="lu") returns an all-NaN 3x3 matrix for the fixed 100x scaling case. Example failing job: https://github.com/kornia/kornia/actions/runs/33774248143/job/100712164319. Investigate/fix this or add a tightly scoped Windows + Torch 2.14 quarantine with an upstream issue before treating 2.14 as the newest supported pin.

3. [P2] .github/workflows/upstream_probe.yml:135-142 omits torch-channel: stable from the Python 3.14 row. The reusable workflow therefore uses its pinned default and keeps reinstalling 2.14.0 after a newer stable Torch ships, so the weekly probe never exercises the newest Torch/Python 3.14 combination. Add the same stable channel input used by the other probe rows.

Local validation: the locked py314 Pixi environment installed successfully; tests/models/test_sam_onnx.py under Python 3.14/Torch 2.14 produced 1 pass and the intended 1 xfail; tests/test_half_precision_ci.py produced 56 passes when run against the PR checkout.

ducha-aiki:
Addressed the review findings in commit 6aa44538:

- hardened known-failure record mode so only a complete candidate with its completion receipt may normalize pytest status 1, while recorder, pipeline, and infrastructure failures remain fatal;
- restored the floating stable channel for the Python 3.14 upstream probe;
- fixed the minimal four-point LU homography path by solving an adaptive-gauge 8x8 system, including valid projective transforms whose bottom-right coefficient is zero;
- added regression, gradient, workflow, and manifest coverage and removed the recovered CPU half-precision baseline entries.

Validation completed:

- pixi run pre-commit-all
- CPU homography suite: 123 passed
- CPU scaled case across float16, bfloat16, float32, and float64: 4 passed
- exact Python 3.13 / PyTorch 2.9.1 / MPS float32 strict known-failure suite: 8,073 passed, 4,138 skipped, 232 expected failures, 9 non-strict XPASS, zero failures
- known-failure manifest tests: 16 passed
- minimal LU gradcheck and torch.compile smoke passed

ducha-aiki:
Reviewed at `627d968e`. The record-mode fix from the previous round is correct — I re-derived it by simulation rather than by reading (details at the bottom). One blocking finding, which the merge commit introduced rather than any of the authored commits.

## P1 — `627d968e` reverted the `uv.lock` bump, and that breaks the new Python 3.14 probe job

The merge of `origin/main` resolved `uv.lock` by taking main's side wholesale, silently discarding `c3744388`:

```
6aa44538     onnx=1.22.0 onnxruntime=1.29.0 onnxscript=0.7.1 torch=2.14.0
627d968e     onnx=1.21.0 onnxruntime=1.23.2 onnxscript=0.5.7 torch=2.9.1
origin/main  onnx=1.21.0 onnxruntime=1.23.2 onnxscript=0.5.7 torch=2.9.1
```

`git diff 6aa44538 627d968e -- uv.lock` is 186 insertions / 467 deletions. `uv.lock` is the only file where the branch had edits competing with main's (#4196 accounts for the rest of the merge delta), so it is the only casualty. The net `uv.lock` diff at head is now a single line — the `torch` specifier `>=2.0.0` → `>=2.5.1` — so summary bullet 4 ("Targeted uv.lock bump: torch + onnx/onnxscript/onnx-ir/onnxruntime only") no longer describes the head.

Three consequences:

1. **`upstream_probe.yml`'s `python314` job cannot install.** `tests.yml` runs `pixi run -e py314 install` → `uv sync --extra dev`, and the locked `onnxruntime==1.23.2` has no cp314 wheel and no sdist. uv says so directly:

   ```
   $ uv pip install --dry-run --no-deps --python-version 3.14 "onnxruntime==1.23.2"
     hint: You require CPython 3.14 (`cp314`), but we only found wheels for
     `onnxruntime` (v1.23.2) with the following Python ABI tags: `cp310`,
     `cp311`, `cp312`, `cp313`, `cp313t`

   $ uv pip install --dry-run --no-deps --python-version 3.14 "onnxruntime==1.29.0"
     + onnxruntime==1.29.0
   ```

   `onnxruntime` 1.29.0 is exactly what `c3744388` had locked. So the weekly probe goes red at the install step from week one, for a reason unrelated to torch drift. (`torch` and `onnx` do ship cp314 wheels at the reverted versions, so `onnxruntime` is the single blocker.)

2. **The SAM ONNX xfail's `raises=` scope is now unverified.** It was measured against onnxscript 0.7.1 / onnx 1.22.0; at head CI installs onnxscript 0.5.7 + onnx 1.21.0 and then force-installs torch 2.14.0. If that older exporter stack raises anything other than `torch.onnx.OnnxExporterError`, `raises=` makes the test a hard failure rather than an xfail and the 2.14 legs go red. Re-check after restoring the lock.

3. The PR body claim needs to match whatever the head actually carries.

Please re-apply the targeted `uv lock --upgrade-package …` on top of the merge.

## P2 — the test-plan item that would have caught this is the wrong command

`pixi install -e py314 --locked` only proves the conda-side solve. The step that fails is `pixi run -e py314 install` (the `uv sync --extra dev`). Run that one before merge.

## P2 — the probe can't be dispatched from this branch

`workflow_dispatch` only lists workflows present on the default branch, so test-plan item 4 ("`upstream_probe.yml` triggered manually at least once, all six jobs green") is unsatisfiable pre-merge. Either accept it as a post-merge action and say so in the body, or add a temporary `pull_request` trigger for one validation run and strip it before merging.

## P2 — `pr_test_cpu.yml:176` comment is falsified by this PR

> `# Single blocking MPS leg: newest Python, newest pinned torch, float32, …`

After this change the newest pinned torch in that file is 2.14.0 and the MPS leg stays on 2.9.1, as does `half-precision` at line 147. Both lags are deliberate — each is pinned to a torch-version-specific known-failure manifest — but the comment now asserts the opposite. Say the manifest is the reason.

## P2 — `PYTORCH_FLOOR` names nothing

`upstream_probe.yml:30`: "Bump PYTORCH_FLOOR below whenever a real torch-pin bump PR lands". There is no `PYTORCH_FLOOR` in the file (`grep` finds exactly that one occurrence); the floor is `pytorch-version: '["2.14.0"]'` repeated across five jobs. A reusable-workflow `with:` can't read a job-level `env`, so the repetition is unavoidable — the comment should just say "bump the five `pytorch-version` inputs below".

## P2 — the weighted path is perturbed for every N, including `svd`, for no reason

`kornia/geometry/homography.py:217-226` restructures `(A.T * w) @ A` into `A.T @ (w * A)`. Mathematically identical, but the multiplication order changes, so it is not bit-identical. On a 7×12 weighted float64 case, `max|main − PR|`:

| solver | delta |
|---|---|
| `lu` (N=12) | 2.27e-13 |
| `svd` (N=12) | 2.54e-13 |

The ≥5-point LU branch and the entire SVD solver are outside the scope of a four-point fix, and last-bit changes there feed `find_homography_dlt_iterated` and RANSAC inlier counts. Keep the original expression on the non-minimal branches and build `Aw` only inside the four-point branch. (`weights is None` is unaffected — `Aw is A`.) Also, `torch.einsum("...ij,...i->...ij", A, w_full)` is just `A * w_full[..., None]`.

## P3 — the homography fix is much bigger than the body and CHANGELOG say, and it is good

`find_homography_dlt(…, 4 points, solver="lu")` is precisely `RANSAC(model_type="homography")`'s minimal solver (`kornia/geometry/ransac.py:215`, `weights=ones`, `batch_size` samples per iteration), so this changes homography RANSAC output for every user on every platform — not only Windows + torch 2.14. A/B on 2048 minimal samples, float32, torch 2.9.1, macOS arm64:

| | main | this PR |
|---|---|---|
| median transfer residual | 2.29e-5 | **1.91e-6** |
| p99 residual | 6.47e-1 | **2.33e-3** |
| max residual | 4.22e+5 | **4.47e+2** |
| time / call (B=2048) | 3.54 ms | 4.46 ms (+26%) |
| RANSAC end-to-end (500 pts, 30% outliers, `max_iter=20`) | 11.4 ms | 11.4 ms |

A large accuracy win; the micro-cost is invisible end-to-end. That is worth stating: the summary still lists seven bullets, none of which is the homography fix, and the CHANGELOG frames it only as "prevents all-NaN homographies on Windows with PyTorch 2.14" rather than as a results change for every homography RANSAC user.

## P3 — `test_projective_fixed_points` does not fail on `main`

On CPU float64, main returns a finite and accurate `H` for that input (max transfer error 7.8e-16; the matrix is merely badly scaled, entries ~3.2e15, because the final `H / (H[-1,-1] + eps)` divides by a near-zero). So it is not a regression pin for the reported bug. It *is* a good design pin for the new gauge: the null vector's last component for that configuration is −1.3e-16, so a fixed `h33 = 1` gauge would be singular. Keep it, but say in the test that that is what it guards. The genuine regression pins here are the two removed half-precision manifest lines.

## P3 — smaller things

- Only one of the five probe jobs is pinned: `test_upstream_probe_python314_uses_floating_stable_torch` covers `python314` alone. Generalize it to assert every job in `upstream_probe.yml` that calls `tests.yml` carries `torch-channel: stable`.
- The body's "can never spuriously fail the job on drift" holds only for the two `record`-mode half legs; `cpu-float32`, `cpu-float64` and `python314` are red on any new failure, as the workflow header itself states. Worth narrowing, since the body is what a maintainer reads.
- Leftover floor prose the raise makes stale: `kornia/core/utils.py:290,307,345,395`, `kornia/core/ops.py:44`, `kornia/feature/lightglue.py:151` ("consider installing torch >= 2.0"), `docs/source/models/segment_anything.rst:93`. Still literally true, so out-of-scope is defensible — a follow-up issue number in the CHANGELOG entry would close the loop.
- `tests/geometry/test_homography.py:217,227` hardcode `rtol=1e-4, atol=1e-4` on values of magnitude 100. It matches the file's prevailing style, and the measured margin is fine (float16 exact, bfloat16 3.8e-6), but since the half manifest entries were removed these now gate the half legs more tightly than `BaseTester`'s dtype-aware defaults would (float16 1e-3, bfloat16 7.8e-3).

## What I verified as correct — no need to redo

1. **The record-mode gate works.** I extracted the `tests.yml` block verbatim, ran it under `bash -eo pipefail` with a stub runner, and swept the outcome space: drift + receipt + candidate → **0**; all-green + receipt → **0**; recorder ABORTED → **1**; no receipt → **1**; pytest rc=2 → **2**; rc=5 → **5**; green without a receipt → **1**; receipt printed but no candidate file → **1**. The previous round's P1 is genuinely fixed, and Copilot's "missing `pipefail`" diagnosis was inverted.
2. No false red when upstream fixes everything: `serialize_manifest` with zero entries still writes 283 bytes of header, so `-s "$candidate_path"` holds.
3. The receipt grep matches the emitter exactly (`testing/half_precision_ci.py:415`, no colour markup), and the ABORTED path returns before the receipt line, so the gate is double-guarded.
4. **The half-precision manifest removals are right.** With a per-test seed plugin at three seeds, `tests/geometry/test_homography.py --dtype=float16,bfloat16,float32,float64` main vs PR differs by exactly `test_scaled_fixed_points[cpu-float16]` and `[cpu-bfloat16]` fixed, nothing new. (Unseeded, the sets wobble by up to six nodes — the pre-existing manifest seeding gap, not this PR.)
5. **Blast radius.** `find_homography_dlt` has one non-test caller (`ransac.py`). Seeded `tests/geometry` + `tests/feature/test_matching.py` at float32+float64: 4613 passed on main, 4615 on the PR (the two new tests), zero failures either side. `tests/geometry/test_ransac.py` green.
6. `torch.compile(find_homography_dlt, fullgraph=True)` on a four-point batch captures cleanly — no graph break from `argmax`/`gather`/`where`.
7. **MPS.** `aten::linalg_qr.out` is unimplemented, so the new path raises `NotImplementedError` exactly as the old one did; the added `mps_float32.txt` line has the right exception type. (Checked with `PYTORCH_ENABLE_MPS_FALLBACK` unset.)
8. `torch.onnx.OnnxExporterError` exists in torch 2.5.1 (`torch/onnx/__init__.py:70`), so the module-level `raises=` does not break collection at the new floor; `torch_version_ge(2, 14)` is a valid two-argument call.
9. Both new CHANGELOG entries land under `## Unreleased`.
10. **The gauge derivation is sound.** For `Aw.T` of shape 9×8, `mode="complete"` puts an exact null vector of `Aw` in `Q[..., -1]` regardless of rank (every column of `Aw.T` lies in `span(Q_1..Q_8)`), and taking its largest-magnitude component as the gauge is the best-conditioned choice available. Detaching it keeps the gradient on the linear solve.

Environment for every measurement above: torch 2.9.1, Python 3.11, macOS arm64 (Darwin 25.5.0), `origin/main` at `d415faa0` and PR at `627d968e` in separate worktrees, each run from the worktree root with `kornia.__file__` printed and checked.

ducha-aiki:
Second pass at `627d968e`, after a deeper sweep. Nothing here changes the P1 above (`uv.lock`); these are additional, all re-measured on the same two worktrees.

## P2 — a zero weight on a four-point sample now returns all-NaN, where main returned finite

```
                                   main            this PR
w = [1, 1, 1, 0]        finite=True  err=1.44    finite=False (all NaN)
w = [1, 1, 1, 1e-45]    finite=True  err=1.44    finite=False (all NaN)
w = [1, 1, 0, 0]        finite=True  err=1.10    finite=False (all NaN)
```

A zero weight zeroes two rows of `Aw`, so the retained 8×8 `selected` is exactly singular and `safe_solve_with_mask` — whose validity mask the new code discards, as the old one did — yields NaN. Neither answer is a *valid* homography (three correspondences leave the system under-determined), but going from finite-garbage to NaN is a silent behaviour change in a public API, and NaN propagates through downstream arithmetic instead of merely being wrong.

One clarification on reachability, because I checked the obvious path and it does **not** hold: `find_homography_dlt_iterated` cannot underflow a weight to zero at N=4, because a four-point DLT fits all four correspondences exactly, so the IRWLS errors are ~0 by construction. I pushed one correspondence to 5000 px and the first-iteration weights were still `[1.00, 1.00, 1.00, 0.96]` on this branch. The reachable path is a caller passing a weight vector that already contains a zero — an inlier mask used as weights on a minimal sample is the natural case. In-repo, `RANSAC` always passes `ones`, so nothing in the tree hits it today.

Cheapest fix consistent with the surrounding code: keep the four-point branch, but fall back (or use the existing mask) when a row pair is fully de-weighted.

## P2 — the hazard is driven by correspondence consistency, not by N, so gating the fix on `shape[1] > 4` leaves the rest exposed

I measured `cond(AᵀA)` on exactly-consistent correspondences:

| N | cond(AᵀA) | smallest singular value |
|---|---|---|
| 4 | 5.47e16 | 4.75e-16 |
| 5 | 1.12e17 | 9.89e-16 |
| 10 | 1.65e16 | 3.93e-15 |
| 20 | 1.78e17 | 7.77e-16 |
| 10, with 0.5 px noise | **4.53e4** | — |

The normal matrix is rank-8-in-exact-arithmetic at *every* N — what makes it singular is that the correspondences are consistent, not that there are four of them. With realistic noise it is well conditioned. So the `else` branch still LU-factors a numerically singular 9×9 and relies on the same "solve `A x = 1`, let the null direction dominate, normalise" trick that fails on Windows + torch 2.14.

I could not reproduce a NaN for N=5/10 locally (both revisions give finite, accurate results here — this is macOS arm64 / torch 2.9.1), so I am raising this as a scoping risk, not a confirmed failure. But the CHANGELOG entry reads as unconditional:

> This prevents all-NaN homographies on Windows with PyTorch 2.14

It prevents them for the minimal sample only. Either scope the sentence to the four-point path, or — better — apply the gauge formulation at any N, which needs no special case: take the null direction of `Aw` and solve the retained eight columns exactly as the new branch does.

## P2 — nothing pins `python-version: '["3.11"]'` on the two record-mode probe jobs

`c9ffb800` moved `cpu-float16`/`cpu-bfloat16` from 3.13 to 3.11 precisely because the candidate header must match the consuming job. I confirmed the constraint is enforced:

```
ValueError: candidate: Python mismatch: expected '3.11', found '3.13'
```

So a drift back to 3.13 silently produces candidates the blocking half-precision job cannot consume. `test_upstream_probe_python314_uses_floating_stable_torch` correctly pins the `torch-channel` fix it was added with, but the 3.11 pin — the one with a real failure mode behind it — is unguarded. Worth folding into the same generalized workflow test I suggested above.

## P3 — the candidate artifact is named after the floor, not the torch actually probed

`tests.yml:255`: `candidate-manifest-…-${{ matrix.pytorch-version }}`. Under `torch-channel: stable`, `pytorch-version` is only a lower bound (the input's own doc comment says so). Once stable passes 2.15, every weekly run still uploads `candidate-manifest-cpu-float16-3.11-2.14.0`, and the job-summary receipt carries no resolved version either — so a downloaded manifest cannot be attributed to the torch release that produced it, which is the one fact a drift probe exists to record. The "Verify PyTorch version" step already prints `got=<version>`; capture it into `$GITHUB_OUTPUT` and use it in the artifact name and the receipt heading.

## P3 — the nightly run has no Windows leg on the version that motivated this PR

`scheduled_test_cpu.yml` keeps Windows (line 152) and macOS (line 199) at `["2.5.1", "2.9.1"]` while only the ubuntu matrix gains 2.6.0 and 2.14.0. The defect that prompted the homography fix was Windows-specific under torch 2.14, so post-merge drift on `main` for that exact combination is caught by nothing — PR CI has the leg, the nightly does not. Relatedly, `tests.yml:11` still defaults `pytorch-version` to `'["2.9.1"]'`, so a future caller that omits the input silently gets the retired pin.

## P3 — `strict=False` defeats the SAM xfail's stated self-expiry

The body says the upper bound makes the failure "un-quarantine itself once upstream fixes it", but if torch fixes the `aten.mul.Scalar` dispatch in 2.14.1, `torch_version_ge(2,14) and torch_version_lt(2,15,0)` is still true and the test XPASSes silently until 2.15. `raises=torch.onnx.OnnxExporterError` already narrows the pin enough that `strict=True` would be safe, and it is what AGENTS.md's documented-issue discipline assumes ("a now-XPASSing strict `xfail` dropped").

## P3 — record mode writes two untracked paths that `.gitignore` does not cover

`grep -nE "candidate|pytest-output|\.log" .gitignore` returns nothing, yet the inline comment describes the candidate as "(untracked)". Anyone reproducing the probe locally is left with `candidate-manifests/` and a full-suite `pytest-output.log` in `git status` — easy to sweep into a `git add -A`.

## P3 — one of the stale-floor sites becomes false, not merely outdated

Adding to the list in my previous comment:

- `tests/geometry/test_conversions.py:307` — "kornia declares torch>=2.0.0" — **false** after this PR, and it is the stated justification for probing `torch.float8_e4m3fn`.
- `tests/geometry/test_conversions.py:6043` — same wording, same problem.
- `kornia/core/_small_linalg.py:180` and `tests/core/test_small_linalg.py:282` — both reason about "torch 2.0-2.4" as a supported-but-untested band that this PR removes from support outright.

These are cheap one-line edits and they are the first thing a future reader consults when deciding whether a version guard is still load-bearing.

PR #4200, opened 2026-09-03
ci: reconcile the test venv with the torch each matrix leg installs

### LINKED ISSUE #4199: CI: pinned torch legs keep the lockfile's triton, so the new 2.6.0 leg fails 37 tests
## Summary

The scheduled CPU matrix pins a torch version per leg, but the install step reinstalls **only** `torch`. Everything else in the venv — including `triton` — stays at whatever `uv sync` resolved for the locked torch, which is **2.14.0** since #4197. So every pinned leg runs its torch against `triton==3.8.0`.

Three of the four pinned versions declare a different triton:

| leg | torch's declared triton | installed | result on `main` (`55046cf8`) |
| --- | --- | --- | --- |
| 2.5.1 | `triton==3.1.0` | 3.8.0 | 6/6 pass |
| **2.6.0** | `triton==3.2.0` | 3.8.0 | **6/6 fail, 37 tests each** |
| 2.9.1 | `triton==3.5.1` | 3.8.0 | 6/6 pass |
| 2.14.0 | `triton~=3.8.0` | 3.8.0 | (fails for an unrelated reason, see below) |

Only 2.6.0 dies from it today, because torch 2.6 is the version whose inductor imports a symbol triton 3.8 no longer has. The other two legs are equally mis-provisioned and merely survive it, so this is a matrix-wide provisioning bug that happens to surface on one leg.

The 2.6.0 leg was added by #4197 and has never been green.

## Evidence

[Failing job](https://github.com/kornia/kornia/actions/runs/33793123008/job/100774848428) — `tests-cpu-ubuntu (float32, 3.11, 2.6.0)`, 37 failures, all one root cause:

- 15 report it directly:
  ```
  ImportError: cannot import name 'AttrsDescriptor' from 'triton.backends.compiler'
  ```
- 22 report `torch.onnx._internal.exporter._errors.TorchExportError`, whose "Exception summary" is that same `ImportError` — the export path calls into inductor, which fails to import, and the exporter wraps it.

So it is 37 instances of one problem, not two problems.

Affected files: `tests/core/test_helpers.py`, `tests/core/test_small_linalg.py`, `tests/filters/test_gaussian.py`, `tests/geometry/test_conversions.py`, `tests/geometry/test_grid.py`, and the other dynamo/export tests.

## Mechanism

From the job log, in order:

1. `uv sync` installs the lock: `+ triton==3.8.0` (one of 83 packages).
2. The pin step (`.github/workflows/tests.yml`, the `pinned` branch of the torch-channel case):
   ```
   uv pip install --reinstall-package torch \
     --index-url https://download.pytorch.org/whl/cpu "torch==$PYTORCH_VERSION"
   ```
   → `Uninstalled 2 packages` / `Installed 2 packages` / `+ torch==2.6.0+cpu`. **triton is not among them.**
3. The verify step checks only the torch version — `torch-channel=pinned expected=2.6.0 resolved=2.6.0` — and passes. Nothing checks the rest of the environment against the torch that was just installed.
4. At test time, torch 2.6's `_is_triton_available()` returns `True` (triton 3.8.0 *is* importable), so it takes the triton path and then does `from triton.backends.compiler import AttrsDescriptor`, which triton 3.8 removed.

The guard is import-availability, not version-compatibility, so an incompatible triton is worse than no triton: with triton absent, `_is_triton_available()` returns `False` and torch takes a working non-triton path. These are CPU-only jobs, so the triton path is not the one under test anyway.

## Suggested fix

Any of these closes it; my preference is the first plus the third.

1. **Install the pinned torch with its own dependency set**, so `triton` moves with it. Note `--index-url` *replaces* PyPI rather than adding to it, so triton has to remain resolvable — `--extra-index-url`, or an explicit second install of the triton that torch declares.
2. **Or drop triton on the pinned legs.** These are CPU jobs; `uv pip uninstall triton` after pinning makes torch take its non-triton path deliberately instead of tripping over a wrong one.
3. **Make the verify step assert the whole environment, not just `torch.__version__`.** Today a silently mis-provisioned venv runs the suite and reports test failures; it should fail loudly at setup. Checking the installed distributions against the pinned torch's `requires_dist` would have caught this before a single test ran — and would also have flagged 2.5.1 and 2.9.1, which are mis-provisioned but green.

## Not this issue

The `2.14.0` legs on the same run fail for an unrelated reason — a lost `xfail` hunk from the #4197 squash, which also broke `pre-commit` and the `coverage / slow` jobs. That is fixed in #4119 (`5c747a1d`, `d4c5f3e2`) and is not related to triton.

## Reproduce

```bash
uv sync                     # gets triton 3.8.0 via the locked torch 2.14.0
uv pip install --reinstall-package torch \
  --index-url https://download.pytorch.org/whl/cpu "torch==2.6.0"
python -c "import torch, triton; print(torch.__version__, triton.__version__)"
pytest tests/geometry/test_grid.py -k export
```

Related: #3930 (the matrix legs previously never installed their torch at all — now that they do, this surfaced), #4197.

ducha-aiki:

Makes each matrix leg's virtualenv coherent with the torch it installs, so the `2.6.0` leg
stops failing 37 tests, and makes a mis-provisioned venv fail at setup instead of as a wall of
test failures.

**One correction to the issue's mechanism.** The pin step does *not* install torch alone: it
resolves and installs torch's whole dependency closure (in the failing job it downgraded
`sympy` 1.14.0 → 1.13.1 alongside torch). triton stays behind for a different reason — the
`+cpu` wheels on `download.pytorch.org` declare **no `triton` at all**. Verified against wheel
metadata for `2.5.1+cpu`, `2.6.0+cpu`, `2.9.1+cpu` and `2.14.0+cpu` (cp311, Linux x86_64): none
of them lists it; only the PyPI wheels do. So the leftover triton is an orphan of a torch that
is no longer installed, and suggested fix 1 ("install the pinned torch with its own dependency
set, so triton moves with it") cannot work — there is no declaration to move. This PR implements
suggested fixes 2 and 3.

- **`.github/workflows/tests.yml`** grows a *Reconcile the environment with the installed torch*
  step that drops sidecar packages the installed torch does not declare. These are CPU-only legs,
  so torch takes its non-triton path deliberately rather than tripping over the wrong one.
- **`.github/scripts/check_torch_env.py`** replaces the inline `torch.__version__` one-liner. It
  asserts the version against the leg's channel (`pinned` exactly, `stable` as a floor), that every
  runtime requirement of *that* torch is installed and satisfied, and that no undeclared sidecar
  survived. It also drops a latent wart: the old check used `version.startswith(expected)`, so
  `expected=2.1` accepted torch `2.14.0`.
- **`tests/core/test_small_linalg.py`** — a second defect, previously masked. Once triton is gone,
  five of the 37 get past the import error and die on torch 2.6 with `TypeError: expected str,
  bytes or os.PathLike object, not BytesIO`: the dynamo exporter's default `external_data=True`
  hands the destination to onnxscript's `save_model_with_external_data`, which takes a filesystem
  path on that version. These graphs have no initializers, so `external_data=False` is the honest
  setting. (`test_cross_kernel_onnx_export_is_informational` re-raises `TypeError` precisely so an
  API misuse cannot masquerade as "cross does not lower here" — that is how this surfaced.)

Behavior note: the PR-time legs (torch 2.14.0) currently keep a *compatible* triton; after this
change they have none. CPU inductor emits C++, nothing in the suite imports triton, and the macOS
and Windows legs have never had it. Measured below.

Fixes #4199. Not touched here: the lost `xfail` hunk from the #4197 squash, which #4119 fixes.

## What did you check?

Reproduced the CI environment exactly — a venv holding the lockfile's `triton==3.8.0`, then
`torch==2.6.0` installed over it from the CPU index — and ran the six files the failing job named.

```text
# torch 2.6.0+cpu + triton 3.8.0 (the state main produces today)
$ pytest tests/core/test_helpers.py tests/core/test_small_linalg.py tests/filters/test_gaussian.py \
         tests/geometry/test_conversions.py tests/geometry/test_grid.py tests/onnx/
37 failed, 654 passed, 2 skipped, 11 xfailed in 13.63s      # same 37 as the failing job

# after the reconcile step: the check names the orphan and the uninstall follows
$ python .github/scripts/check_torch_env.py --list-undeclared-sidecars
undeclared-sidecars: triton
$ python .github/scripts/check_torch_env.py --channel pinned --expected 2.6.0
error: triton 3.8.0 is installed, but torch 2.6.0+cpu declares no dependency on it: torch
dispatches onto an importable sidecar without checking its version (#4199)     # exit 1
$ uv pip uninstall triton && pytest <same six files>
5 failed, 687 passed        # the masked BytesIO defect

# with both fixes, torch 2.6.0+cpu, no triton
$ pytest <same six files>
692 passed, 1 skipped, 11 xfailed in 15.22s

# no regression on the PR-time version, torch 2.14.0+cpu, no triton
$ pytest <same six files>
692 passed, 1 skipped, 11 xfailed in 16.33s
```

Also run: `pytest tests/scripts/ tests/test_half_precision_ci.py` (103 passed — the new unit tests
for the check plus the existing workflow-shape assertions), and `pre-commit` on every changed file.

Not verified: a full-suite run on a 2.6.0 leg (needs the model-weights cache and the heavy optional
deps). The claim here is scoped to the six files the failing job reported.

## How did AI help?

Claude (Opus 5) traced the mechanism from the job log, checked the four `+cpu` wheels' metadata to
establish that none declares triton, wrote the check script and its tests, and ran the before/after
measurements above in throwaway venvs pinned to torch 2.6.0+cpu and 2.14.0+cpu.

### DIFF
--- a/.github/scripts/check_torch_env.py
+++ b/.github/scripts/check_torch_env.py
@@ -1,0 +1,228 @@
+#!/usr/bin/env python3
+
+# LICENSE HEADER MANAGED BY add-license-header
+#
+# Copyright 2018 Kornia Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""Check the CI virtualenv against the torch build a matrix leg just installed.
+
+``tests.yml`` builds its environment in two steps: ``uv sync`` installs the lockfile,
+whose torch comes from PyPI, and the matrix leg then replaces *torch alone* with a build
+from the CPU wheel index. Everything ``uv sync`` installed for the locked torch stays
+behind, and the ``+cpu`` wheels declare a narrower dependency set than the PyPI ones --
+no ``triton`` at all (checked for 2.5.1, 2.6.0, 2.9.1 and 2.14.0). The lockfile's triton
+therefore survives the swap as an orphan belonging to a torch that is no longer
+installed, and torch dispatches onto it: ``_is_triton_available()`` tests whether
+``import triton`` succeeds, not whether that triton matches. torch 2.6 then runs
+``from triton.backends.compiler import AttrsDescriptor``, a name triton 3.8 removed, and
+37 dynamo/export tests die at import. See #4199.
+
+Two modes:
+
+``--list-undeclared-sidecars``
+    Print the installed sidecar packages the installed torch does not declare, for the
+    workflow to uninstall. These legs are CPU-only, so torch taking its non-triton path
+    deliberately is the right outcome; an incompatible triton is worse than no triton.
+
+``--channel CHANNEL --expected VERSION``
+    Verify that the torch version satisfies the leg's channel, that every runtime
+    requirement of *that* torch is installed and satisfied, and that no undeclared
+    sidecar survived. A mis-provisioned venv fails here, at setup, instead of surfacing
+    later as a wall of unrelated-looking test failures.
+
+Machine-readable results go to stdout as ``key: value`` lines; diagnostics go to stderr.
+
+Usage::
+
+    python3 .github/scripts/check_torch_env.py --list-undeclared-sidecars
+    python3 .github/scripts/check_torch_env.py --channel pinned --expected 2.6.0
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from collections.abc import Iterable, Mapping, Sequence
+from importlib.metadata import PackageNotFoundError, distribution, distributions
+
+from packaging.requirements import Requirement
+from packaging.utils import canonicalize_name
+from packaging.version import InvalidVersion, Version
+
+# Packages torch selects at runtime by import availability rather than by version, so an
+# orphan left over from a different torch is silently picked up instead of ignored. Both
+# names install the same ``triton`` import; ``pytorch-triton`` is the nightly channel's.
+SIDECAR_PACKAGES = ("triton", "pytorch-triton")
+
+
+def runtime_requirements(
+    requires: Iterable[str] | None, environment: Mapping[str, str] | None = None
+) -> list[Requirement]:
+    """Return the requirements of a distribution that apply to a plain, no-extra install.
+
+    Args:
+        requires: ``Requires-Dist`` entries, as ``importlib.metadata`` reports them.
+        environment: Marker variables overriding the running interpreter's, for tests.
+
+    Returns:
+        The requirements whose markers hold, with the extras-only ones dropped.
+
+    """
+    markers = {"extra": ""}
+    markers.update(environment or {})
+    kept = []
+    for entry in requires or ():
+        requirement = Requirement(entry)
+        if requirement.marker is not None and not requirement.marker.evaluate(markers):
+            continue
+        kept.append(requirement)
+    return kept
+
+
+def installed_versions() -> dict[str, str]:
+    """Map the canonical name of every installed distribution to its version."""
+    found: dict[str, str] = {}
+    for dist in distributions():
+        name = dist.metadata["Name"]
+        if name:
+            found.setdefault(canonicalize_name(name), dist.version)
+    return found
+
+
+def undeclared_sidecars(
+    requires: Iterable[str] | None,
+    installed: Mapping[str, str],
+    environment: Mapping[str, str] | None = None,
+) -> list[str]:
+    """Return the installed sidecar packages that the installed torch does not declare."""
+    declared = {canonicalize_name(item.name) for item in runtime_requirements(requires, environment)}
+    return [name for name in map(canonicalize_name, SIDECAR_PACKAGES) if name in installed and name not in declared]
+
+
+def check_torch_version(version: str, channel: str, expected: str) -> str | None:
+    """Return a problem description when ``version`` does not satisfy the leg's channel.
+
+    ``pinned`` demands exactly ``expected`` (the local ``+cpu`` segment is ignored);
+    ``stable`` treats ``expected`` as a floor, since the upstream probe installs whatever
+    the stable index resolves to today.
+    """
+    release = Version(version.partition("+")[0])
+    if channel == "pinned":
+        if release != Version(expected):
+            return f"torch {version} is not the pinned {expected}"
+    elif channel == "stable":
+        if release < Version(expected):
+            return f"torch {version} is below the {expected} floor of the stable channel"
+    else:
+        raise ValueError(f"unknown torch channel: {channel!r}")
+    return None
+
+
+def check_environment(
+    torch_version: str,
+    requires: Iterable[str] | None,
+    installed: Mapping[str, str],
+    channel: str,
+    expected: str,
+    environment: Mapping[str, str] | None = None,
+) -> list[str]:
+    """Return every way the environment disagrees with the torch that is installed.
+
+    Args:
+        torch_version: Version of the imported torch, ``+cpu`` local segment included.
+        requires: ``Requires-Dist`` entries of the installed torch distribution.
+        installed: Canonical distribution name to version, as ``installed_versions``.
+        channel: ``pinned`` or ``stable``.
+        expected: The matrix version -- exact under ``pinned``, a floor under ``stable``.
+        environment: Marker variables overriding the running interpreter's, for tests.
+
+    Returns:
+        Human-readable problem descriptions; empty when the environment is coherent.
+
+    """
+    problems = []
+    version_problem = check_torch_version(torch_version, channel, expected)
+    if version_problem is not None:
+        problems.append(version_problem)
+
+    for requirement in runtime_requirements(requires, environment):
+        name = canonicalize_name(requirement.name)
+        version = installed.get(name)
+        if version is None:
+            problems.append(f"torch {torch_version} requires {requirement}, which is not installed")
+        elif not requirement.specifier.contains(version, prereleases=True):
+            problems.append(f"torch {torch_version} requires {requirement}, but {requirement.name} {version} is here")
+
+    for name in undeclared_sidecars(requires, installed, environment):
+        problems.append(
+            f"{name} {installed[name]} is installed, but torch {torch_version} declares no dependency on it: "
+            "torch dispatches onto an importable sidecar without checking its version (#4199)"
+        )
+    return problems
+
+
+def _torch_distribution() -> tuple[str, list[str] | None]:
+    """Return the installed torch's runtime version and its ``Requires-Dist`` entries."""
+    import torch
+
+    dist = distribution("torch")
+    try:
+        recorded = Version(dist.version.partition("+")[0])
+        running = Version(torch.__version__.partition("+")[0])
+    except InvalidVersion:  # A source or nightly build; the metadata check is best-effort.
+        return torch.__version__, dist.requires
+    if recorded != running:
+        raise SystemExit(f"torch imports as {torch.__version__} but its distribution metadata says {dist.version}")
+    return torch.__version__, dist.requires
+
+
+def main(argv: Sequence[str] | None = None) -> int:
+    """Run the requested mode and return the process exit status."""
+    parser = argparse.ArgumentParser(description="Check the CI venv against the installed torch.")
+    parser.add_argument("--channel", choices=("pinned", "stable"), help="torch channel the leg installed with")
+    parser.add_argument("--expected", help="matrix torch version: exact under pinned, a floor under stable")
+    parser.add_argument(
+        "--list-undeclared-sidecars",
+        action="store_true",
+        help="print the sidecar packages the installed torch does not declare, and exit",
+    )
+    args = parser.parse_args(argv)
+
+    try:
+        installed = installed_versions()
+        if args.list_undeclared_sidecars:
+            stale = undeclared_sidecars(distribution("torch").requires, installed)
+            print("undeclared-sidecars: " + " ".join(stale))
+            return 0
+        torch_version, requires = _torch_distribution()
+    except PackageNotFoundError:
+        print("torch is not installed in this environment", file=sys.stderr)
+        return 1
+
+    if args.channel is None or args.expected is None:
+        parser.error("--channel and --expected are required unless --list-undeclared-sidecars is given")
+
+    problems = check_environment(torch_version, requires, installed, args.channel, args.expected)
+    print(f"torch-version: {torch_version.partition('+')[0]}")
+    for requirement in runtime_requirements(requires):
+        name = canonicalize_name(requirement.name)
+        print(f"  {name} {installed.get(name, '(missing)')}", file=sys.stderr)
+    for problem in problems:
+        print(f"error: {problem}", file=sys.stderr)
+    return 1 if problems else 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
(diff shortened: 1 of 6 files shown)


ducha-aiki:
## Review

The diagnosis is right and unusually well-evidenced: the `+cpu` wheels really do declare no `triton`, so the orphan cannot be moved by a resolver and suggested fix 1 is genuinely dead. Checking the *whole* dependency closure rather than `torch.__version__` is the correct shape for the gate, and catching the `startswith` wart (`expected=2.1` accepting `2.14.0`) on the way through is a nice bonus. The `external_data=False` finding is real and the reasoning for it — no initializers, so nothing to externalise on any version — is sound.

Findings below, most important first.

### 1. The verify step's own guard is dead code — `.github/scripts/check_torch_env.py:218`

`main()` prints the machine-readable line before the problem list:

```python
problems = check_environment(...)
print(f"torch-version: {torch_version.partition('+')[0]}")   # unconditional
...
return 1 if problems else 0
```

So a failing check still writes a valid-looking `torch-version: 2.14.0` to stdout. Verified locally: `--channel pinned --expected 2.6.0` against torch 2.14.0 exits 1 *and* emits the version line. The workflow's explicit guard —

```bash
if [[ -z "$resolved" ]]; then
  echo "the environment check did not report a torch version" >&2
```

— can therefore never fire. The only thing that actually fails the step is the implicit `set -o pipefail` that GitHub's `shell: bash` happens to default to. Run the same pipeline under `bash -e` without pipefail and you get `resolved=[2.14.0]` and a green step: the mis-provisioned venv sails straight through the gate this PR exists to add.

Print `torch-version:` only when `problems` is empty (or emit an explicit `env-ok:` marker the step greps for), so the guard the PR already wrote becomes load-bearing rather than decorative.

### 2. "37 tests die at import" is the wrong number, in two places

`check_torch_env.py:30` and the workflow comment at `tests.yml:147` both say 37 dynamo/export tests die at the triton import. The PR's own measurements say otherwise: 37 failed with triton present, 5 failed after removing triton alone. So 32 died at import and 5 were the independent `BytesIO`/`external_data` defect. `CHANGELOG.md:215` gets this right ("32 dynamo and export tests"); the other two copies drifted.

Relatedly, the ~8-line triton explanation is now triplicated verbatim across the script docstring, the workflow comment, and the changelog — and it has already disagreed with itself on the first write. Keep the long form in the script docstring and have the workflow comment point at it; a future reader debugging this will otherwise chase five tests that were never triton's fault.

### 3. `main()` is never exercised, so the stdout contract is unpinned — `tests/scripts/test_check_torch_env.py`

The module tests `runtime_requirements`, `undeclared_sidecars`, `check_torch_version`, `check_environment` and `installed_versions`, but never imports `main`. Neither of the two contracts the workflow actually depends on is tested:

- the `undeclared-sidecars: ` / `torch-version: ` stdout prefixes that `sed -n 's/^undeclared-sidecars: //p'` greps, and
- the non-zero exit that makes the verify step blocking.

Rename the prefix to `stale-sidecars:`, or move the `torch-version:` print behind the problem loop as #1 suggests, and all 80 tests still pass while the reconcile step's `sed` silently yields the empty string — the uninstall never runs, #4199 comes back, and CI is green until the 32 import failures reappear. The sibling `tests/scripts/test_check_import_surface.py` does import and test `main`, so the pattern is already established next door.

### 4. The ordering assertion guards the wrong edge — `tests/scripts/test_check_torch_env.py:164`

`test_reconcile_step_runs_before_the_tests` asserts the reconcile step precedes `- name: Run tests`. But the ordering the whole fix depends on is that it comes *after* `- name: Install PyTorch matrix version`. Move the reconcile step above the pin step and this test still passes — while the check now runs against the *locked* PyPI torch, which does declare triton, so `undeclared_sidecars` returns `[]`, nothing is uninstalled, the leg swaps in `+cpu` torch afterwards, and #4199 is fully back with a green suite. Add:

```python
assert workflow.index("- name: Install PyTorch matrix version") < workflow.index("--list-undeclared-sidecars")
```

### 5. Unbounded slice in the verify-step assertion — `tests/scripts/test_check_torch_env.py:176`

`workflow.partition("- name: Verify PyTorch version")[2]` returns everything from that marker to end of file, so the assertions can be satisfied by any *later* step. If the verify step is ever reverted to the old inline `python -c 'import torch; ...'` one-liner while some subsequent step happens to invoke `check_torch_env.py`, the test still passes and the whole-environment gate is gone — precisely the regression the docstring says it guards. Bound the slice at the next `\n      - name:`.

### 6. `external_data=False` is ungated while `dynamo` beside it is gated — `tests/core/test_small_linalg.py:94`

`_export_kwargs` documents itself as "Export kwargs that are valid on every torch kornia declares support for", and `dynamo` is version-gated accordingly. `external_data` is not, though it does not exist on `torch.onnx.export` before 2.6. On the 2.5.1 legs `_export_kwargs(True)` now returns `{'opset_version': 18, 'dynamo': True, 'external_data': False}`.

The inline comment is right that this is unreachable today, because all three call sites call `_require_exporter` first — I checked lines 209/213, 277/281, 298/303. But the safety lives in a two-line ordering at each call site rather than in the function that claims the invariant, and `test_cross_kernel_onnx_export_is_informational` deliberately re-raises `TypeError`, so a new caller or a reordering turns this into a hard red rather than a skip. `if use_dynamo_exporter and torch_version_ge(2, 6, 0):` costs nothing and puts the guarantee where the docstring promises it.

### 7. The reconcile step cannot tell "clean" from "the parse broke" — `.github/workflows/tests.yml:150`

The script always prints the `undeclared-sidecars: ` prefix, so an empty `$stale` means the parse failed, not that the environment is clean — yet both paths print "No stale torch sidecars to remove". If the prefix drifts, or `pixi run`/`uv run` ever interposes on stdout, the step reports success while doing nothing. Note the asymmetry: the sibling verify step *does* guard on an empty capture. Grep for the prefix line itself and fail when it is absent.

### 8. A two-name allowlist for a general problem — `.github/scripts/check_torch_env.py:67`

`SIDECAR_PACKAGES = ("triton", "pytorch-triton")` closes this instance, not the class. The entire closure `uv sync` installed for the locked torch survives the swap orphaned — `uv.lock:2388-2400` shows the lockfile's torch 2.14.0 pulling `cuda-toolkit`, `cuda-bindings`, `nvidia-cudnn-cu13`, `nvidia-cusparselt-cu13`, `nvidia-nccl-cu13`, `nvidia-nvshmem-cu13` on Linux. None of them is listed, none is flagged by `check_environment`, and they cost gigabytes of runner disk. More to the point, the next torch that sniffs a sidecar by import availability — the actual defect class — reproduces #4199 silently, because its name is not in the tuple. The general form is to diff the installed set against the closure the installed torch declares (the script already computes that set), or reinstall torch with `--reinstall` so the resolver moves the closure. Worth at least a comment saying why the allowlist was chosen over the closure diff.

### Nits

- **`.github/scripts/check_torch_env.py:219`** — `main()` recomputes `runtime_requirements(requires)` (already computed inside `check_environment`) and dumps the requirement table to stderr unconditionally. A healthy `--channel pinned --expected 2.14.0` run emits 14 lines of noise before the result, on every leg. Return the table from `check_environment` and print it only when `problems` is non-empty, where it is diagnostic.
- **`.github/scripts/check_torch_env.py:214`** — the `--channel`/`--expected` required-argument check runs *after* `installed_versions()` enumerates every distribution and `_torch_distribution()` imports torch, so a no-argument invocation pays a full torch import before printing its usage error. argparse can express the constraint directly, before any environment work.
- **`.github/scripts/check_torch_env.py:100`** — `found.setdefault(...)` means that when a distribution is visible twice, the winner is whatever `distributions()` yields first, i.e. `sys.path` order rather than what will actually be imported. A leftover `.dist-info` from an interrupted `uv pip install --reinstall-package torch` — exactly the step that runs immediately before this check — could record a stale `sympy` and fail the leg spuriously, or the reverse and hide a real mismatch. Prefer last-wins/`max`, or report the conflict explicitly.
- **`.github/workflows/tests.yml:150`** — the reconcile step is a whole extra `pixi run → uv run → python` spawn, and `uv.lock:2400` marks triton `sys_platform == 'linux'`, so every macOS and Windows leg pays that startup to print "No stale torch sidecars to remove". Folding the uninstall into the existing `Install PyTorch matrix version` step (which already has the pixi/uv context open), or adding a `--reconcile` mode that lists, removes and verifies in one invocation, would avoid it.
- **`tests/scripts/test_check_torch_env.py:22`** — the `sys.path.insert(0, .github/scripts)` bootstrap is now copy-pasted into a second file in this directory, and each copy prepends that directory to `sys.path` for the rest of the pytest session. A single `tests/scripts/conftest.py` does the insert once and scopes the intent.

### Not blocking

The "not verified" scoping in the PR body is honest and the right call — the six-file claim is what the failing job reported, and a full-suite 2.6.0 run needs the weights cache. Worth noting only that the *behaviour* change (PR-time legs lose a currently-compatible triton) is asserted rather than measured beyond those six files; if any inductor test outside them ever wants a GPU-shaped triton path, it would surface after merge rather than here.

ducha-aiki:
Addressed in 79323a7e. Twelve of the thirteen findings are fixed; one turned out to be wrong for this repo and is explained below.

### The gate is now actually blocking (finding 1)

`main()` prints `torch-version:` only when the check passes. Verified by hand against the local torch:

```text
$ python .github/scripts/check_torch_env.py --channel pinned --expected 2.6.0
exit=1
stdout=[]
  triton 3.8.0
error: torch 2.14.0+cu130 is not the pinned 2.6.0

$ python .github/scripts/check_torch_env.py --channel stable --expected 2.5.1
torch-version: 2.14.0
exit=0
```

So the step's `if [[ -z "$resolved" ]]` guard fires on its own rather than depending on GitHub's implicit `-o pipefail`, and it is no longer unreachable.

### Ordering is structural rather than conventional (findings 4, and the spawn nit)

The reconcile moved *inside* the step that swaps torch, now named `Install the matrix torch and reconcile the environment`. It cannot be reordered ahead of the swap, because there is no longer a separate step to move — which is a stronger guarantee than the assertion I asked for. That also drops the pixi/uv/python spawn that was a guaranteed no-op on every macOS and Windows leg. The test still asserts `--reinstall-package torch` precedes `--list-undeclared-sidecars`, now *within* that step's body.

### The rest

- **2** — "37 dynamo/export tests" corrected to 32 in the script docstring; the workflow comment no longer restates the mechanism and points at the docstring instead, so there is one home for it rather than three.
- **3** — `main()` is imported and tested: both stdout prefixes, the exit contract, and a direct pin that a failing check emits no version line. The prefixes are now the module constants `SIDECAR_PREFIX`/`VERSION_PREFIX`, and a test asserts the workflow's own `sed`/`grep` text contains them, so a rename on one side fails rather than silently no-opping.
- **5** — a `_step()` helper bounds each workflow slice at the next `- name:` instead of running to end of file.
- **6** — `external_data` is gated at 2.6 like the `dynamo` kwarg beside it, with the reasoning moved into the function that promises the invariant.
- **7** — the reconcile greps for the result line and fails when it is absent, instead of reading an empty capture as a clean venv. Symmetric with the verify step now.
- **8** — `SIDECAR_PACKAGES` records why it is an allowlist rather than a closure diff: installed metadata carries no provenance, so a diff cannot tell an orphan from a root, and most of the venv is legitimately outside torch's closure. The `nvidia-*`/`cuda-*` survivors are noted as a disk cost that a `+cpu` torch never loads.
- **Nits** — the requirement table prints only when it diagnoses something; arguments are validated before the environment is scanned and torch is imported (`main([])` now exits 2 without importing torch); `scan_installed()` reports a distribution visible twice with disagreeing versions rather than settling it by filesystem order.

### One finding was wrong (the `conftest.py` nit)

Moving the `sys.path.insert` into a `tests/scripts/conftest.py` breaks this repository. `tests/` has no `__init__.py`, so a second bare `conftest` module shadows the repository-root one — and `tests/test_half_precision_ci.py:33` does `import conftest as project_conftest` and derives the repo root from `Path(project_conftest.__file__).parent`. Measured: with that conftest present, `import conftest` resolves to `tests/scripts/conftest.py` and five workflow assertions go looking for `tests/scripts/.github/workflows/`.

```text
# with tests/scripts/conftest.py
$ pytest tests/scripts/ tests/test_half_precision_ci.py
FileNotFoundError: '/…/kornia/tests/scripts/.github/workflows/tests.yml'
```

Reverted. `test_check_import_surface.py` is back to its original state and untouched by this PR, and the per-module insert now carries a comment recording why it stays, so the next reader does not retry it.

### Checks

```text
$ pytest tests/scripts/ tests/test_half_precision_ci.py
113 passed          # 103 before, 10 new tests

$ pre-commit run --files <every changed file>
all hooks passed
```

`tests/core/test_small_linalg.py` reports 4 failed, 40 passed locally on `onnx_ir 0.1.13` (`module 'onnx_ir' has no attribute 'schemas'`) — identical counts with the file un-edited, so pre-existing to this environment and unrelated. The workflow change itself only runs in CI.

Issue #4201, opened 2026-09-03
MPS: torch 2.14 batched linalg.svd/svdvals/lstsq fail at >= 8192 input elements, blocking RANSAC on Apple silicon


ducha-aiki:
### Summary

On torch 2.14.0, `torch.linalg.svd`, `torch.linalg.svdvals` and `torch.linalg.lstsq` raise on the
MPS device as soon as the input holds **more than 8192 elements**:

```
RuntimeError: Failed to created pipeline state object, error: Error Domain=AGXMetal13_3 Code=2
"Compilation failed due to an interrupted connection: XPC_ERROR_CONNECTION_INTERRUPTED.
This error occurred after multiple retries."
```

torch 2.14.0 is what kornia's CPU matrix and lockfile pin since #4197, and what the MPS gate
moves to. The 8192 ceiling is the reason `RANSAC` stays pinned in
`testing/known_failure_xfails/mps_float32.txt` after that migration.

### Reproduction

```python
import torch, platform
print("torch", torch.__version__, "|", platform.platform())
for n in (910, 911):
    A = torch.randn(n, 3, 3, device="mps")
    try:
        torch.linalg.svd(A)
        torch.mps.synchronize()
        print(f"n={n} numel={A.numel()} ok")
    except RuntimeError as e:
        print(f"n={n} numel={A.numel()} FAIL {str(e)[:120]}")
```

```
torch 2.14.0 | macOS-26.5.1-arm64-arm-64bit
n=910 numel=8190 ok
n=911 numel=8199 FAIL Failed to created pipeline state object, error: ...
```

### The boundary is exactly `numel > 8192`

Bisected largest working batch, `float32`, `(N, m, m)`:

| shape | largest working N | `N * m * m` |
|---|---|---|
| `(N, 2, 2)` | 2047 | 8188 |
| `(N, 3, 3)` | 910 | 8190 |
| `(N, 4, 4)` | 511 | 8176 |
| `(N, 8, 9)` | 113 | 8136 |
| `(N, 9, 9)` | 101 | 8181 |

`svd`, `svdvals` and `lstsq` all share the ceiling. `eigh`, `qr`, `inv`, `det`, `solve`,
`lu_solve` and `cholesky` were checked to N = 4096 at both 3x3 and 9x9 and have no such limit.
It is a shader-compilation failure, not memory pressure — 8192 float32 elements is 32 KiB.

### Impact on kornia

[...]

PR #4202, opened 2026-09-03
ci: move the MPS gate to torch 2.14.0 on macos-15 and re-record the baseline

ducha-aiki:
## Summary

- Move the blocking `tests-mps` leg from torch 2.9.1 to 2.14.0, the pin the CPU matrix and the lockfile already use since #4197
- Pin the image to `macos-15` instead of `macos-latest`, which is not cosmetic: see below
- Re-record `testing/known_failure_xfails/mps_float32.txt` at 2.14.0: **200 entries -> 14**
- Skip `tests/geometry/test_ransac.py` on MPS via a `mps_process_abort` marker, because it aborts
  the process rather than failing (#4204), with `--run-mps-process-abort` to run it anyway

## Why

#4197 deliberately left this leg on 2.9.1 ("bumping it means re-recording the manifest in the
same change") and listed the migration as out of scope. Two things make it due now.

**The pin was mis-provisioning the leg.** `tests.yml` reinstalls only `torch` for a pinned leg, so
a 2.9.1 pin against a lockfile resolved for 2.14.0 runs one package out of step with the rest of
its dependency set (#4199). Tracking the newest CPU-matrix pin removes that. #4200 has since made
that mismatch loud rather than silent, which is the right guard, but the leg should not need it.

**The baseline is 186 lines of upstream gaps that no longer exist.** Almost the whole 2.9.1
manifest was two missing kernel families, and 2.14.0 ships both.

`main` was briefly red for this reason too: #4197 added
`TestFindHomographyDLT::test_zero_weight_minimal_lu`, whose adaptive-gauge path calls
`torch.linalg.qr` — no MPS kernel on 2.9.1 — with no manifest line. #4200 fixed that by pinning it
as the 200th entry. **This PR removes that pin**, because on 2.14.0 the test passes; verified
locally, `1 passed`.

## What torch 2.14.0 changes on MPS

It adds kernels for both `grid_sample` backward passes, for `padding_mode="border"` on 2-D and
`mode="nearest"` on 5-D, and for the whole eigen/QR/LU/SVD family (`eigh`, `svd`, `svdvals`, `qr`,
`lu_solve`, `lstsq`, `cholesky_solve`, `matrix_exp`) — everything but `_linalg_eigvals`. Those were
the two largest causes in the old baseline, 71 `grid_sampler_*_backward` and 103 linalg entries out
of 200 per the #4159 attribution, and they are all gone. The half-precision MPS LU **process abort**
is gone too: every half dtype now raises an ordinary exception.

The 14 that remain: **12** for `aten::_linalg_eigvals`, still missing, which keeps the 5-point
essential solver off the device, and **2** for the `yuv420`/`yuv422` empty-input reshape defect.

The new native SVD path has a ceiling of its own, which is *not* in the baseline: a **batched**
`svd`/`svdvals`/`lstsq` fails to build a Metal pipeline once its input holds **8192 elements or
more** — the threshold is inclusive, `(511,4,4)` = 8176 passes and `(512,4,4)` = 8192 raises. A
single unbatched matrix is unaffected at any size: `(1,128,128)` holds 16384 elements and succeeds,
because torch stages a lone large matrix on the CPU. That is #4201, and it is why `RANSAC`'s
batched minimal solvers still raise on MPS.

## Why the image is pinned to `macos-15`

This is the part worth reviewing. Same PR, same Python 3.13, same torch 2.14.0, one variable:

| image | result |
|---|---|
| `macos-latest` (`macos-26-arm64`) | **670 failed**, 7626 passed, then SIGABRT |
| `macos-15-arm64` | **14 failed**, 8280 passed |

664 of the 670 are one error:

```
RuntimeError: Failed to created pipeline state object, error: Error Domain=CompilerError Code=2
"error: unsupported deferred-static-alloca-size function body in
 metal::cooperative_tensor<float, ...> mpp::tensor_ops::__matmul2d_detail::__operand_layout<
     matmul2d_descriptor{...,64,64,-1}, ..., execution_simdgroups<4>, float,float,float,int>"
```

That is torch 2.14's Metal 4 `cooperative_tensor` matmul shader failing to compile on the runner's
*paravirtualized* GPU ("Apple M1 (Virtual)"). One specialization, `matmul2d_descriptor<64,64,-1>`,
accounts for all 664, spread across 292 `tests/augmentation`, 230 `tests/geometry`,
104 `tests/filters` and the rest. The raising call sites show what 2.14 routes through it:
`torch.linalg.inv` (400), `linalg.solve` (99), `inv_ex` (31), `lu_factor_ex` (28) and **`F.conv3d`**
(25). All of those *worked* on that runner under 2.9.1, where the whole suite had 200 failures — so
staying on `macos-latest` would mean either keeping torch 2.9.1 indefinitely or recording 670
upstream failures as expected.

**The blocker is the virtual GPU, not macOS 26.** A physical M1 on macOS 26.5 compiles those
shaders and produces the same 14 failures this leg records. macOS 15 is Metal 3 and takes the older
kernel path, which is the path that happens to work on the runner fleet. Two consequences are worth
stating rather than leaving implicit:

- no job in this repository now exercises the Metal 4 path that users on current macOS actually run;
- when GitHub retires the `macos-15` image, this leg stops scheduling and `collector` goes red with
  no fallback.

Neither is solved here. Re-test `macos-latest` once upstream guards those kernels by device
capability, rather than waiting for the image to be retired. Full measurement: kornia#4159
(comment).

## Why RANSAC is skipped rather than pinned

`tests/geometry/test_ransac.py` on MPS aborts the pytest process, exit 134, and the abort site
**moves**:

| run | selection | image | outcome |
|---|---|---|---|
| A | `--deselect tests/geometry/test_ransac.py` | macos-15 | **completed**, 14 failed |
| B | `--deselect ...::TestRANSACEssential` | macos-15 | aborted in `F.pad` (`test_affine.py::test_downsizeAA`) |
| C | none | macos-26 | aborted in `_multiply_deg_one_poly` (`RANSAC` 5-point) |

RANSAC's batched minimal solvers take the Metal shader compiler down, and the next MPS allocation
dies — the same failure family as #4201. `SIGABRT` has no exception type, so
`--xfail-known-failures` cannot express it, and a moving abort site would make the gate red at a
different test each run. Deselecting the whole file is the only configuration that completes.

**What that costs, measured.** Against `main`'s conftest on a real M1 with torch 2.14.0,
`tests/geometry/test_ransac.py --device=mps` gives `3 failed, 10 passed, 12 skipped, 1 xfailed,
1 xpassed`. So the skip removes **9 device tests that pass on real Apple silicon** and hides 3
failures (2 are #4201, 1 is `_linalg_eigvals`). The skip is deliberately unconditional rather than
CI-only: gating it on `CI` would keep that local coverage but would leave `pixi run test-mps` red
on 3 unpinned failures, breaking the "run the identical contract locally" property TESTING.md
promises. `--run-mps-process-abort` (env `KORNIA_TEST_RUN_MPS_PROCESS_ABORT`) restores them for
local diagnosis instead, the way `--run-device-agnostic` does for its marker.

The skip keys on a `mps_process_abort` marker declared in the test module, not on a path prefix in
`conftest.py`, so a rename cannot silently disarm it — which would return the job to a SIGABRT at
a moving, unrelated site — and so `grep -rnE "#4204" kornia/ tests/`, the check AGENTS.md
prescribes before closing a tracked issue, actually finds it.

## Test plan

Local, macOS 26.5.1 arm64 (M1), Python 3.11, torch 2.14.0, MPS float32,
`PYTORCH_ENABLE_MPS_FALLBACK` unset, full suite in one process:

- [x] Recorded baseline replays green: `--xfail-known-failures` -> **0 unexpected**, exit 0
- [x] `--run-mps-process-abort` restores exactly `main`'s behaviour on that file
      (`3 failed, 10 passed, 12 skipped, 1 xfailed, 1 xpassed`)
- [x] `test_zero_weight_minimal_lu`, pinned by #4200 for 2.9.1, passes on 2.14.0
- [x] `pixi run pre-commit-all` clean
- [x] Hosted `tests-mps` green at `926f3dca` on macos-15: **8310 passed, 4124 skipped,
      95 deselected, 45 xfailed, 9 xpassed, 0 failures**, 7m13s (run 33813238586,
      job 100851162083) — the same skip/xfail counts as the local run, so the manifest
      replays identically on the runner. All 43 checks green.

The manifest header cites the macos-15 CI run it was recorded from, and says explicitly that that
job ran *without* `--xfail-known-failures` — it is the recording run, not a replay, so it is red by
construction and a reader should not read it as a failing gate.

## Related

- kornia#4201 — batched MPS `svd`/`svdvals`/`lstsq` fail at >= 8192 input elements on 2.14.0; a CPU
  route (measured cheaper than on-device chunking) would make `RANSAC` fundamental/homography work
  on MPS
- kornia#4203 — torch 2.14 flags a singular matrix 2.9.1 did not, and `symmetric_transfer_error`
  returns `nan` instead of `max_num` for a singular `H`
- kornia#4204 — the RANSAC MPS abort
- pytorch/pytorch#195937 — the SVD ceiling, upstream

## Out of scope

- The `_torch_svd_cast` workaround for kornia#4201, and any library-level guard that would refuse
  to run these paths on MPS
- Migrating the MPS manifest onto the #4154 profile mechanism, which is still CPU-only
  (`conftest.py` requires `device == "cpu"` for a profile), and the `mps-float32` row that
  `upstream_probe.yml` is waiting on
- Removing the now-unnecessary-on-2.14 MPS branches in `kornia/feature/laf.py` and
  `kornia/core/utils.py`; kornia still supports torch 2.5.1
- `kornia/core/_small_linalg.py` still cites "both torch versions kornia's CI runs -- 2.5.1 and
  2.9.1" for its ONNX measurement, but the dynamo leg has been 2.14.0 since #4197; that drift
  belongs to #4197, and 2.9.1 remains a live pin for the half-precision legs


ducha-aiki:
## Review — `ci: move the MPS gate to torch 2.14.0 on macos-15 and re-record the baseline`

Reviewed at `d0cedde6`. **Approve, with four documentation corrections that should land before merge** — none of them touches the gate's behaviour, but three of them are wrong numbers or wrong statements inside contract documents, which is exactly where an exact-baseline mechanism cannot afford prose drift.

I re-measured the substantive claims on a real M1 (macOS 26.5.1, Python 3.11, torch 2.14.0, `PYTORCH_ENABLE_MPS_FALLBACK` unset) rather than taking them from the description.

---

### What I verified holds

**The manifest is exactly right, and the arithmetic between the two CI runs reconciles with nothing left over.**

| | recording run (`6c48f261`, no `--xfail-known-failures`) | head (`d0cedde6`, manifest applied) |
|---|---|---|
| failed | 14 | 0 |
| xfailed | 31 | **45** = 31 in-tree + the 14 pinned |
| deselected | 122 = 95 `device_agnostic` + 27 ransac (`--deselect`) | 95 |
| skipped | 4098 | 4124 = +26 ransac skips |
| passed | 8280 | 8281 = +1 (`TestRansacMethods::test_max_samples_by_conf`) |

The +1 is not noise: that test's node ID carries no `[mps` suffix, so the new `conftest` predicate correctly leaves it running where the old `--deselect` dropped it. Every number moves for a reason.

**Entry-set diff: 185 lines removed, `0` added.** Nothing regressed into the baseline.

**No removed entry can be silently skipped.** The skip predicates are unchanged apart from RANSAC, so a test that was *failing* under 2.9.1 (i.e. was in the manifest) cannot now be skipped — except the 8 `tests/geometry/test_ransac.py` lines, which are the documented trade. `tests/contrib/test_image_stitcher.py::TestImageStitcher::test_smoke[mps-float32-ransac]` matches on the param id only, lives in a different file, is *not* caught by the predicate, and genuinely passes now.

**The kernel claims are true on 2.14.0.** All OK locally: `grid_sample` 2-D `border`, 5-D `nearest`, 2-D and 3-D backward, `eigh`, `qr`, `lu_solve`, `matrix_exp`, `cholesky_solve`. `_linalg_eigvals` still raises `NotImplementedError`. The two `yuv` entries reproduce with `RuntimeError` on MPS and **pass on CPU**, so they are legitimately device-specific and they carry over unchanged from the 2.9.1 baseline.

**`pixi run test-mps` (pixi.toml:48) really is the same contract** — `pytest tests/ --xfail-known-failures` with `KORNIA_TEST_DEVICE=mps`. And `scheduled_test_cpu.yml`'s `macos-latest` leg is CPU-only, so the Metal 4 problem does not reach it. Good — that was my first worry on reading the image pin.

---

### Asks

#### 1. `198` is stale — the current baseline is `199`

`#4197` (`55046cf8`) added `TestFindHomographyDLT::test_projective_fixed_points[mps]`, taking the manifest from 198 to 199 entries:

```
$ for c in 4b69473a 55046cf8; do git show "$c:testing/known_failure_xfails/mps_float32.txt" | grep -v '^#' | grep -c .; done
198
199
```

`199 -> 14`, not `198 -> 14`. Three places at head: `CHANGELOG.md:83`, `TESTING.md:310`, and the manifest header line 6 (plus the PR description).

#### 2. `TESTING.md` and `CHANGELOG.md` say RANSAC is *pinned in the baseline*. It is not — it is skipped.

`TESTING.md:311-313`:

> … which is why the batched minimal solvers behind `RANSAC` are still pinned in the baseline below.

`CHANGELOG.md:88-91` lists the `svd`/8192 ceiling under **"What is still pinned:"**.

The shipped baseline is 14 lines: 12 `_linalg_eigvals` + 2 `yuv`. There is no `svd` entry and no RANSAC entry. Both documents then say, three lines later, that the file is *skipped*. This is a leftover from the earlier revision of the branch — the 17-entry manifest at `6c48f261` did pin them, and the prose did not follow the manifest when it shrank to 14. Please make both read "skipped, not pinned (#4204)".

#### 3. The `8192 input elements` rule is wrong in both directions

`TESTING.md:301` and `:314-316` state the ceiling as "more than 8192 input elements (batch x rows x cols)". Measured:

| input | numel | result |
|---|---|---|
| `(1, 128, 128)` | 16384 | **OK** — torch warns `matrix too large to stage in MPS threadgroup memory (33124 > 32768 bytes); falling back to CPU` |
| `(1, 91, 91)` | 8281 | OK |
| `(511, 4, 4)` | 8176 | OK |
| `(512, 4, 4)` | **8192** | **FAIL** |
| `(127, 8, 8)` | 8128 | OK |
| `(128, 8, 8)` | **8192** | **FAIL** (same for `svdvals`) |
| `(1000, 3, 3)` | 9000 | FAIL |
| `(64, 16, 16)` | 16384 | FAIL |

Two corrections. It fires **at** 8192, not above it — 8176 is the last size that works. And it is the **batched** path only: an unbatched 128x128 SVD holds 16384 elements and succeeds, because torch routes a single large matrix to CPU. As written the rule predicts a failure that does not happen. Suggested wording: *"batched `svd`/`svdvals`/`lstsq` whose input holds 8192 elements or more (batch x rows x cols); a single unbatched matrix is unaffected — torch falls back to CPU for those."* The same phrasing is in `CHANGELOG.md:89` and in #4201.

#### 4. Stale sibling comment left behind in `.github/workflows/tests.yml:169`

```
# instead. macos-14 images segfault inside MPS kernels; macos-15 and later work.
```

This PR deletes exactly that claim from `pr_test_cpu.yml` and replaces it with the 670-failure evidence that macos-26 does **not** work. `tests.yml` is the reusable workflow that implements the MPS leg and hosts the "Verify MPS device" step, so it is the first file someone lands in — and it is not in this PR's file list. One line: `macos-14 segfaults inside MPS kernels and macos-26 cannot compile torch 2.14's Metal 4 shaders; use macos-15 (see pr_test_cpu.yml).`

---

### Smaller points

**The provenance link points at a red job on a commit that is not in this PR.** Manifest header line 1 cites run `33808269822` / job `100824208527`. That job is on `6c48f261` ("tmp: same diagnostic on macos-15 to test the Metal 4 hypothesis"), which is not in the branch's single-commit history; it ran `-rf --deselect tests/geometry/test_ransac.py` *without* `--xfail-known-failures`, against a 17-entry manifest still headed `PLACEHOLDER_RUN_URL` / `macos-26-arm64`; and it concluded **failure**. It is genuinely the recording run, which is what `TESTING.md` asks the header to cite — so this is not wrong, but a maintainer following the link lands on a red job on an orphan SHA with a different manifest. Consider a second header line for the green replay at head: run `33812256481` / job `100836682340`, `8281 passed, 45 xfailed, 9 xpassed, 0 unexpected`. "Recorded from" and "verified by" are different facts and the file currently records only the confusing one.

**Please quantify what the RANSAC skip costs.** On main's `conftest` (no skip), real M1, torch 2.14:

```
$ pytest tests/geometry/test_ransac.py --device=mps --dtype=float32
3 failed, 10 passed, 12 skipped, 1 xfailed, 1 xpassed
```

So the skip removes **9 device tests that pass on real Apple silicon** (10 minus the non-device `test_max_samples_by_conf`) and hides 3 failures — 2 are the #4201 pipeline-state error, 1 is `_linalg_eigvals`. All 8 of the `test_ransac.py` lines dropped from the old manifest were already-pinned failures, so the net CI change for that file is `-9` passing tests. That is a fine trade against the +175 recovered elsewhere, but the description only says "the only configuration that completes"; the number belongs in the `CHANGELOG` entry and in #4204.

I agree the skip must be **unconditional** rather than CI-gated, and I think that is worth saying out loud in the `conftest` comment: gating it on `CI` would keep local coverage but would make `pixi run test-mps` red on 3 unpinned failures, breaking the "run the identical contract locally" property that `TESTING.md` promises. As written, a reader may well assume the skip was meant to be CI-only.

**A residual fragility, already inside #4204's scope:** the gate now depends on `test_image_stitcher.py`'s RANSAC call staying under the pipeline-state ceiling. If it ever crosses it, the job dies with `SIGABRT` rather than reporting a clean failure, and no manifest entry can catch that. Worth a sentence on #4204 rather than anything in this PR.

**Copilot's two items:** `PLACEHOLDER_RUN_URL` is already fixed at head — that review ran against the `6c48f261`-era content. The "mark the error as a verbatim quote" suggestion is cosmetic; I reproduced the string locally and the table's quote is accurate, including the elided `XPC_ERROR_CONNECTION_INTERRUPTED` tail.

---

### Verdict

The mechanism is right, the baseline is exact, the image pin is well-evidenced, and this also unbreaks a currently-red `main`. **Approve** once asks 1-4 land; they are all one-line edits and none needs another CI cycle to validate.

ducha-aiki:
### Addendum — a deeper pass, still at `d0cedde6`

Six more items, all verified locally. None changes the verdict; #2 and #5 are the ones I'd actually act on.

**1. There is a *fourth* stale `198`, and this one is wrong under either reading.** `.github/workflows/pr_test_cpu.yml:189`: "the same runner on torch 2.9.1 had 198 failures". That is a failure count, not an entry count, and main produces **200**: 199 manifest lines plus the unpinned `test_zero_weight_minimal_lu` that this PR's own description identifies as what turned the gate red. So the sentence is off by one as an entry count and by two as a failure count.

**2. The `macos-15` pin leaves the Metal 4 path uncovered by any job — and the stated reason for the pin is not quite the real one.** `pr_test_cpu.yml:190` says macOS 15 "is Metal 3, takes the older kernel path, and reproduces real Apple-silicon hardware exactly". But the reference machine in the test plan is on macOS **26.5.1** (`sw_vers` -> `26.5.1`, build `25F80`), i.e. Metal 4 — and it produces the same 14. So macos-15 agrees with real hardware in *outcome*, not by taking the same path, and the defect is specific to the runner's **paravirtualized GPU**, not to macOS 26. As written the comment tells a future maintainer that macOS 26 is the problem; the measurement says the virtual GPU is.

The consequence is a coverage hole worth naming in the comment: `device: mps` now appears in exactly one job in the repo, and it runs the Metal 3 path that no current-macOS user is on. A kornia change that breaks only the Metal 4 route is invisible everywhere. And when GitHub retires the `macos-15` image, `tests-mps` fails to schedule and `collector` goes red with no fallback. Neither needs solving here — but "re-test macos-latest once upstream guards those kernels by device capability" is a weaker note than the situation deserves.

**3. The skip predicate is a bare path prefix with no `::` terminator.** `conftest.py`: `item.nodeid.startswith("tests/geometry/test_ransac.py")`. Rename or move the file and the skip silently stops matching — the job then dies with exit 134 at a site the description documents as *moving*, so the red job names an unrelated test and nobody connects it to the rename. In the other direction, a future `tests/geometry/test_ransac_helpers.py` would be silently skipped on MPS by the same prefix. Only `test_ransac.py` exists today, so this is latent, not live. A `pytestmark = pytest.mark.skipif(...)` in the module travels with the file and is visible where the tests are.

**4. Give the skip an opt-in escape hatch.** `--run-device-agnostic` (`conftest.py:492`) is the existing precedent for exactly this shape: a collection-time exclusion with a flag to force it back on. Combined with the 9 real-hardware passes the skip removes, a `--run-mps-ransac` would restore local coverage without touching the CI contract, and would give #4204 something to test against.

**5. The `#4204` pin sits outside the grep roots `AGENTS.md` prescribes.** `AGENTS.md:123` requires, before merging anything that closes a tracked issue, `grep -rnE "#NNNN|issues/NNNN" kornia/ tests/`. The `#4204` references are in `conftest.py` (repo root) and `TESTING.md` — neither path is covered. Whoever fixes #4204 runs the documented check, gets a clean result, and leaves 26 tests skipped forever. Moving the pin into `tests/geometry/test_ransac.py` as a module-level `pytestmark` fixes this and point 3 at once; widening the documented roots is the alternative.

**6. The re-recorded manifest is no longer sorted.** main's 199 lines are node-ID sorted (`sort -c` passes); the new 14 are in pytest collection order (`test_smoke`, `test_shape[1-5]`, … , `test_degenerate_case[10-5]`). `testing/known_failures.py:94` emits `for nodeid in sorted(self.pending)`, so the next maintainer who regenerates from the tracker's own output produces a fully reordered file, and a genuine one-line change is invisible in that diff. Worth a `sort` before commit.

**Optional:** `_apply_device_skips` walks the collected item list four times, once per rule — ~12.4k items on the MPS full-suite run, on the collection path of every pytest invocation in the repo including CPU. One `for item in items:` with the four `if` bodies inside is the same code, one traversal, and reads as a table of rules.

**Out of scope but worth a follow-up:** `kornia/core/_small_linalg.py:177` still says `torch.linalg.cross` "lowers on both torch versions kornia's CI runs -- 2.5.1 (legacy exporter) and 2.9.1 (legacy and dynamo)". The dynamo leg is 2.14.0 since #4197, so that measured claim now names a version no dynamo/ONNX leg runs. That drift belongs to #4197, not here — 2.9.1 is still a live PR pin for the half-precision legs (`pr_test_cpu.yml:147`), so this PR does not remove the last consumer.

**One thing I checked and rejected.** The shipped contract *has* been exercised on CI, contrary to what the unchecked box in the test plan suggests: job `100836682340` in run `33812256481` runs at `d0cedde6` with the real `--xfail-known-failures` config and the `conftest` skip, and is green — `8281 passed, 4124 skipped, 45 xfailed, 9 xpassed`, 0 unexpected, 10m29s. Only the checkbox is stale; please tick it.

ducha-aiki:
### Rebased onto #4200, and every review item applied — `d0cedde6` -> `926f3dca`

#4200 merged four hours after this branch was cut and conflicted with it, so the PR was `DIRTY` and running no checks at all. Rebased; the only conflict was the manifest, resolved in favour of the 14-entry re-record.

**The rebase changed one of the review's own numbers.** #4200's last commit pinned `TestFindHomographyDLT::test_zero_weight_minimal_lu` as the **200th** manifest entry — that is how it un-redded `main`, which this PR no longer needs to claim. So the baseline is **200 -> 14**, not 198 and not 199. Every figure is now derived from `origin/main` rather than carried forward: the `#4159` attribution split is restated as 71 `grid_sampler` + 103 linalg, which reconciles exactly with the 174 `NotImplementedError` lines in the 200-entry file. That pin is dropped here because the test passes on 2.14.0 — verified, `1 passed`.

#### The four asks

1. **`198`** -> `200` in `CHANGELOG.md`, `TESTING.md`, the manifest header, `pr_test_cpu.yml`, the commit message and the PR body.
2. **"pinned" vs "skipped"** — `TESTING.md` and the `CHANGELOG` now say plainly that `eigvals` and `yuv` are what the baseline pins, and that `test_ransac.py` is *skipped*. The `#4201` ceiling is described as the reason those solvers raise, with an explicit "it is not why they are absent from the baseline".
3. **The 8192 rule** — restated everywhere as **batched, and `>= 8192`**, with the measured boundary (`(511,4,4)` = 8176 passes, `(512,4,4)` = 8192 raises) and the unbatched exemption (`(1,128,128)` = 16384 succeeds; torch stages a lone large matrix on the CPU). I have corrected #4201's wording too.
4. **`tests.yml:169`** — rewritten. It now names both hazards and points at the `tests-mps` comment rather than asserting "macos-15 and later work".

#### The addendum items

- **Metal 4 reasoning** — the workflow comment and `TESTING.md` now say the blocker is the *virtual* GPU rather than macOS 26, since a physical M1 on macOS 26.5 compiles those shaders and yields the same 14. Both consequences you flagged are written down instead of left implicit: no job exercises the Metal 4 path users are on, and image retirement takes `collector` with it.
- **Path prefix -> module marker.** The skip now keys on `pytest.mark.mps_process_abort`, declared in `tests/geometry/test_ransac.py` and registered in `pyproject.toml`. A rename can no longer silently disarm it, the `::`-terminator over-match is gone, and `grep -rnE "#4204" kornia/ tests/` now finds the pin — which closes the AGENTS.md grep-roots gap in the same edit.
- **Escape hatch** — `--run-mps-process-abort` / `KORNIA_TEST_RUN_MPS_PROCESS_ABORT`, mirroring `--run-device-agnostic`. Verified it reproduces `main`'s behaviour on that file exactly: `3 failed, 10 passed, 12 skipped, 1 xfailed, 1 xpassed`. Left ungated on `CI` deliberately, and the reason is now in the comment.
- **Manifest sort order** — re-sorted by node ID, with a header line saying why (`known_failures.py:94` emits `sorted()`, so an unsorted file turns the next re-record into a whole-file diff).
- **Provenance** — the header now states that the cited job ran *without* `--xfail-known-failures` and is the recording run, so nobody reads a red job as a failing gate.
- **Single pass** — the four device-skip rules now share one traversal.
- `_small_linalg.py` is left alone and moved to "Out of scope" with the reason: that drift is #4197's, and 2.9.1 is still a live pin for the half-precision legs at `pr_test_cpu.yml:147`, so this PR does not remove the last consumer.

#### Verification

Full suite, one process, real M1, macOS 26.5.1, torch 2.14.0, `PYTORCH_ENABLE_MPS_FALLBACK` unset, run twice:

```
8309 passed, 4124 skipped, 95 deselected, 45 xfailed, 9 xpassed  in 3m00s
```

`45 xfailed` is `31` in-tree plus the `14` pinned, and no manifest entry misbehaved in either run. `pixi run pre-commit-all` clean.

#### One thing the re-run found that is not this PR's

The first of the two runs also failed `tests/enhance/test_jpeg.py::TestDiffJPEG::test_keyword_argument[mps-float32]`, and the second did not. It is not MPS-specific and not caused by anything here: `jpeg_codec_differentiable` returns **all-NaN** at `jpeg_quality=0` (`5000.0 / compression_strength` at `jpeg.py:240`, unguarded, though the docstring documents `[0, 100]`), and `tests/enhance/test_jpeg.py` draws the quality from `torch.randint(low=0, high=100, ...)` at 16 unseeded sites. Measured on plain CPU, 20 consecutive runs of that one file: **1 failed**. So roughly 5% of runs of that file go red on any job. Filed as #4205 with the repro — it is worth fixing precisely because this PR's gate treats an unlisted failure as a hard stop.

PR #4209, opened 2026-09-04
docs(core): update stale CI torch version in comments (2.9.1 → 2.14.0)

### LINKED ISSUE #4206: Stale CI-version claims in _small_linalg.py and core/utils.py: the dynamo leg is 2.14.0, not 2.9.1
### Describe the bug

Two comments in `kornia/` cite "the torch versions kornia's CI runs" as **2.5.1 and 2.9.1**, and name 2.9.1 as the version where the *dynamo* ONNX exporter was measured. Neither is true on `main` any more:

- the CPU matrix is `["2.5.1", "2.14.0"]` (`.github/workflows/pr_test_cpu.yml:89`);
- the dynamo leg is `2.14.0` (`pr_test_cpu.yml:166`);
- 2.9.1 survives only on the two half-precision legs (`pr_test_cpu.yml:147`), which run no exporter at all.

The drift came in with #4197, which moved the matrix and the dynamo leg off 2.9.1. #4202 moved the last MPS leg and deliberately left these two alone rather than widen a CI-config PR into `kornia/`.

**Site 1 — `kornia/core/_small_linalg.py:176-178`**

```
ONNX lowering is *not* a reason, measured: ``torch.linalg.cross`` lowers on both torch
versions kornia's CI runs -- 2.5.1 (legacy exporter) and 2.9.1 (legacy and dynamo) -- to
``Slice``/``Mul``/``Sub``/``Concat``.
```

**Site 2 — `kornia/core/utils.py:198-201`**

```
# hold on the torch versions CI runs: measured, ``torch.linalg.cross`` lowers on 2.5.1 (legacy
# exporter) and on 2.9.1 (both exporters).
```

### Why this is only a labelling fix

I re-measured on torch 2.14.0 / onnx 1.22.0 / onnxscript 0.7.1. **The substance of both claims survives** — `cross` still lowers on both exporters and `linalg_inv` still lowers on neither, so no code change and no branch collapse is implied:

| | legacy exporter | dynamo exporter |
|---|---|---|
| `torch.linalg.cross` | `Slice`, `Mul`, `Sub`, `Concat`, `Constant` | `Split`, `Mul`, `Sub`, `Concat` |
| `aten::linalg_inv` | `UnsupportedOperatorError: Exporting the operator 'aten::linalg_inv' to ONNX opset version 20 is not supported` | `ConversionError` at translation |

```python
import torch, io, onnx
class Cross(torch.nn.Module):
    def forward(self, a, b): return torch.linalg.cross(a, b)
a, b = torch.randn(2, 3), torch.randn(2, 3)
for dynamo in (False, True):
    buf = io.BytesIO()
    torch.onnx.export(Cross(), (a, b), buf, dynamo=dynamo, **({"external_data": False} if dynamo else {}))
    buf.seek(0)
    print(dynamo, sorted({n.op_type for n in onnx.load(buf).graph.node}))
```

Note the dynamo lowering *shape* changed between the two releases — `Slice` became `Split` — which is exactly the kind of thing a stale version label hides.

### Expected behavior

Restate both comments against the versions CI actually runs (2.5.1 and 2.14.0), keeping them as dated measurements rather than open-ended claims. The neighbouring sentence at `_small_linalg.py:173-174` — "torch 2.5.1 has no `bfloat16` `cross` on MPS, while 2.9.1 does" — is a *kernel-coverage* statement about a version boundary, not a claim about CI, so it stays correct as written; it would only be worth re-checking whether 2.14.0 changes anything there.

This is the same class as the stale-prose sweep AGENTS.md asks for after a pin flip, so it is worth grepping for other "versions CI runs" phrasings in the same pass. The other 2.9.1 mentions in `kornia/geometry/conversions.py` and `kornia/core/download.py` are explicitly "measured on torch 2.9.1" attributions, which stay true and should be left alone.

### Environment

```
kornia 7efa3684 | torch 2.14.0 | onnx 1.22.0 | onnxscript 0.7.1 | python 3.11 | OS macOS 26.5.1 arm64
```

won-seoop:
Update stale CI torch-version references in two docstring/comment blocks.

**Background**

`kornia/core/_small_linalg.py` and `kornia/core/utils.py` both contain
prose that cites the torch versions measured during ONNX-lowering
experiments. Those comments still reference `2.9.1` as the upper CI
version, but the PR CI matrix (`pr_test_cpu.yml`) has moved on:

```yaml
# pr_test_cpu.yml
pytorch-version: ["2.5.1", "2.14.0"]   # main matrix
pytorch-version: ["2.14.0"]            # dynamo / compile leg
```

The functional claims remain valid — `torch.linalg.cross` still lowers
correctly and `aten::linalg_inv` still does not export — so this is a
documentation-only correction.

**Changes**

| File | Location | Before | After |
|------|----------|--------|-------|
| `kornia/core/_small_linalg.py` | line 174 | `while 2.9.1 does` | `while 2.14.0 does` |
| `kornia/core/_small_linalg.py` | line 177 | `2.9.1 (legacy and dynamo)` | `2.14.0 (legacy and dynamo)` |
| `kornia/core/utils.py` | line 200 | `2.9.1 (both exporters)` | `2.14.0 (both exporters)` |

Closes #4206

### DIFF
--- a/kornia/core/_small_linalg.py
+++ b/kornia/core/_small_linalg.py
@@ -171,10 +171,10 @@ def _inverse_3x3_cross(input: torch.Tensor) -> torch.Tensor:
     Prefer :func:`_inverse_3x3_scalar` wherever a graph is being captured. The reason is
     **kernel coverage, not ONNX lowering**: a backend with no ``cross`` kernel for the working
     dtype makes this raise rather than return -- torch 2.5.1 has no ``bfloat16`` ``cross`` on
-    MPS, while 2.9.1 does.
+    MPS, while 2.14.0 does.
 
     ONNX lowering is *not* a reason, measured: ``torch.linalg.cross`` lowers on both torch
-    versions kornia's CI runs -- 2.5.1 (legacy exporter) and 2.9.1 (legacy and dynamo) -- to
+    versions kornia's CI runs -- 2.5.1 (legacy exporter) and 2.14.0 (legacy and dynamo) -- to
     ``Slice``/``Mul``/``Sub``/``Concat``. What neither exporter lowers on either version is
     ``aten::linalg_inv``, which is what :func:`kornia.core.utils._torch_inverse_cast` avoids by
     reaching for a closed form in the first place. torch 2.0-2.4 is below the declared floor; CI does not run it
--- a/kornia/core/utils.py
+++ b/kornia/core/utils.py
@@ -197,7 +197,7 @@ def _inverse_3x3_closed_form(input: torch.Tensor) -> torch.Tensor:
     # Under tracing/export (legacy ONNX / jit.trace / dynamo ONNX) stick to the plain scalar
     # adjugate. NOTE the original rationale here -- "whereas ``cross`` may not [lower]" -- does not
     # hold on the torch versions CI runs: measured, ``torch.linalg.cross`` lowers on 2.5.1 (legacy
-    # exporter) and on 2.9.1 (both exporters). The split is kept because it is still the safer
+    # exporter) and on 2.14.0 (both exporters). The split is kept because it is still the safer
     # capture path (scalar arithmetic needs no per-dtype kernel at all, and ``cross`` has real
     # kernel gaps -- no bfloat16 on MPS in torch 2.5.1), not because ``cross`` fails to export.
     # Collapsing the two branches is a behavior change and belongs in its own PR.


won-seoop:
## 📌 기여 기록 (SPICABCR)

**한 줄 요약**
CI 설정 변경 후 동기화되지 않은 torch 버전 참조 주석을 3줄 수정으로 바로잡아, ONNX 내보내기 경로 문서의 정확성을 확보했습니다.

---

**문제: CI 버전과 주석 간 불일치**
26년 9월, `_small_linalg.py`와 `utils.py`에서 ONNX 내보내기 경로 주석이 실제 PR CI 버전과 달랐습니다. 주석에는 `2.9.1`이 명시되어 있었으나, 실제 CI 매트릭스는 `2.14.0`을 사용하고 있었습니다.

**중요성: 잘못된 주석은 잘못된 판단을 유발**
ONNX 내보내기 경로는 dynamo와 legacy exporter 분기를 주석 근거로 결정합니다. 버전 참조가 틀리면 후속 기여자가 검증 안 된 버전을 기준으로 코드를 수정할 위험이 있었습니다.

**원인: CI 업데이트 시 주석 미동기화**
`pr_test_cpu.yml`의 torch 버전 매트릭스가 `2.9.1 → 2.14.0`으로 교체될 때, 소스 코드 내 주석 3곳이 수동으로 갱신되지 않았습니다.

**검토 A: 버전 자동 생성 스크립트 도입**
CI YAML에서 버전을 읽어 주석을 자동 갱신하는 방식을 검토하였지만, 3줄 불일치에 비해 변경 범위가 과도하다고 판단하였습니다.

**적용 B: CI YAML 직접 확인 후 최소 수정**
`.github/workflows/pr_test_cpu.yml`을 열어 dynamo 잡 버전(`2.14.0`)을 직접 확인하고, 2개 파일 3곳의 주석을 수정하였습니다.

**검증**
`pytorch-version: ["2.5.1", "2.14.0"]`(일반 잡), `pytorch-version: ["2.14.0"]`(dynamo 잡) 확인으로 수정 근거를 검증했습니다.

**결과**
파일 2개, 변경 줄 3줄. 코드 변경 없이 문서 정확성을 확보하였습니다.

ducha-aiki:
## Review of #4209 at `111b564c` — CHANGES REQUESTED, the version label moved but the measurement under it did not

Thanks for picking this one up — it is a real drift and the two CI-claim substitutions are the right idea. Setup: worktree at the PR head, measurements on torch 2.14.0 / onnx 1.22.0 / python 3.11, macOS arm64.

### First, what I confirmed is still true

I re-executed both functional claims on 2.14.0 rather than taking "the functional claims remain valid" on trust:

| | legacy exporter | dynamo exporter |
|---|---|---|
| `torch.linalg.cross` | exports → `Slice`, `Mul`, `Sub`, `Concat`, `Constant` | exports → **`Split`**, `Mul`, `Sub`, `Concat` |
| `aten::linalg_inv` | `UnsupportedOperatorError: ... opset version 20 is not supported` | `ConversionError` at translation |

And the MPS kernel-coverage half, `PYTORCH_ENABLE_MPS_FALLBACK` unset:

```text
torch 2.14.0 | mps True
  cross MPS bfloat16: OK (2, 3)
  cross MPS float16:  OK (2, 3)
  cross MPS float32:  OK (2, 3)
```

So nothing in the substance has rotted. Three things about the edit itself, though.

### 1. The version label moved to 2.14.0 but the node list did not — and that list is now half wrong

`_small_linalg.py:176-178` at head reads:

```
``torch.linalg.cross`` lowers on both torch versions kornia's CI runs -- 2.5.1 (legacy
exporter) and 2.14.0 (legacy and dynamo) -- to ``Slice``/``Mul``/``Sub``/``Concat``.
```

The sentence now attributes **"legacy and dynamo"** to 2.14.0 and then names a node list that, on 2.14.0, is only the *legacy* exporter's shape. The dynamo exporter emits `Split`, not `Slice` — see the table above. On 2.9.1, which is what the sentence used to name, `Slice` was right for both. So the substitution silently turned a correct sentence into one that is wrong for the half of it that the version bump was supposed to make current.

This is the specific failure #4206 called out: *"the dynamo lowering shape changed between the two releases — `Slice` became `Split` — which is exactly the kind of thing a stale version label hides."* The fix is to update the node list alongside the version, e.g. `... to ``Slice``/``Mul``/``Sub``/``Concat`` (the dynamo exporter on 2.14.0 emits ``Split`` in place of ``Slice``)``.

### 2. The MPS sentence at `:173-174` should not have been touched

```diff
-    MPS, while 2.9.1 does.
+    MPS, while 2.14.0 does.
```

#4206 says this one explicitly:

> The neighbouring sentence at `_small_linalg.py:173-174` — "torch 2.5.1 has no `bfloat16` `cross` on MPS, while 2.9.1 does" — is a *kernel-coverage* statement about a version boundary, not a claim about CI, so it stays correct as written.

It is not a CI claim, so CI moving does not stale it. Its job is to locate the boundary at which the kernel gap closes, and 2.9.1 is the measured, earliest-known such version. Bumping it to 2.14.0 is not false (I measured 2.14.0 does have the kernel, above) but it is strictly weaker: a reader can no longer tell whether 2.6–2.9 have `bfloat16` `cross` on MPS, and the scheduled workflow still runs 2.6.0 and 2.9.1. Please revert that hunk.

### 3. The grep the issue asked for was not run, and it has a hit

#4206 asked for a sweep: *"it is worth grepping for other 'versions CI runs' phrasings in the same pass."* There is one, in this feature's own test:

`tests/core/test_small_linalg.py:296`
```
# ``torch.linalg.cross`` lowers on every torch version CI runs: 2.14.0 and 2.9.1 on both
# exporters, and 2.5.1 on the legacy one ...
```

After this PR the production docstring says CI runs **two** versions (2.5.1, 2.14.0) and its own pinned test says CI runs **three** (2.5.1, 2.9.1, 2.14.0). Both cannot be the enumeration. For the record, PR CI today runs all three — `pr_test_cpu.yml:89` is `["2.5.1", "2.14.0"]`, `:147` pins the two half-precision legs at `2.9.1`, and `:166` puts the dynamo leg at `2.14.0` — so the word **"both"** in the docstring is the part that is actually stale, not just the number after it. Either widen the docstring to the same three versions the test names, or drop the enumeration and keep it a dated measurement.

### Summary of what I am asking for

1. Update the `Slice`/`Mul`/`Sub`/`Concat` node list to match what 2.14.0's dynamo exporter actually emits.
2. Revert `_small_linalg.py:173-174` back to `2.9.1` — it is a kernel-coverage boundary, not a CI claim.
3. Reconcile with `tests/core/test_small_linalg.py:296`, which is the third site and disagrees with the new text.

`kornia/core/utils.py:200` gets the same treatment as (1)/(3) — it is the same sentence in comment form. Everything else here is good, and none of this needs a code change.

ducha-aiki (changes requested):
See Opus review

Issue #4168, opened 2026-09-02
Bug: `Boxes.merge` concatenates unbatched `(N, 4, 2)` boxes along the vertex axis


ducha-aiki:
## Summary

`Boxes.merge` concatenates along `dim=1` unconditionally (`kornia/geometry/boxes.py:329` at `35e4938d`). For batched `(B, N, 4, 2)` data that is the box axis. For unbatched `(N, 4, 2)` data it is the **vertex** axis, so two unbatched objects merge into an `(N, 8, 2)` tensor and every later export silently reads the eight vertices as one box.

## Reproduction

```python
import torch
from kornia.geometry.boxes import Boxes

a = Boxes.from_tensor(torch.tensor([[1.0, 2.0, 5.0, 4.0]]), mode="xyxy")
b = Boxes.from_tensor(torch.tensor([[6.0, 3.0, 9.0, 8.0]]), mode="xyxy")
m = a.merge(b)
print(m.data.shape)          # torch.Size([1, 8, 2])  -- expected torch.Size([2, 4, 2])
print(m.to_tensor("xyxy"))   # tensor([[1., 2., 9., 8.]])  -- the union bounding box of both inputs, one row
```

Neither the constructor nor `to_tensor` validates `data.shape[-2] == 4` after the merge, so nothing raises.

## Expected

`(N, 4, 2)` + `(M, 4, 2)` → `(N + M, 4, 2)`, matching the documented batched behavior. The fix is `dim=-3` (or `dim=1 if self._is_batched else 0`). `Boxes3D` has no `merge`, so the 3D class is unaffected.

## Context

Found while reviewing #4060 (batch 4a), whose `merge` docstring and pin describe the batched case only. Not in scope there: that PR is documentation and tests only. Part of the #4020 bbox/boxes audit; disposition: separate fix PR, it does not overlap #3934/#4061.

Posted on behalf of @ducha-aiki by Claude (Fable 5.1).

alinaqvi1129:
id like to work on this @ducha-aiki can u assign this to me

ducha-aiki:
@alinaqvi1129 yes, take it. GitHub will not let me put a non-collaborator in the assignee field, so this comment is the claim marker.

Confirmed still live on `main` @ `ee6b713f` (torch 2.9.1):

```
unbatched merge -> shape (1, 8, 2), to_tensor("xyxy") = [[1.0, 2.0, 9.0, 8.0]]   # one box, the union
[...]

PR #4210, opened 2026-09-04
fix(augmentation): auto-augment ops preserve float16 / bfloat16 input dtype

tritsystem:

`OperationBase.forward` (the shared gate for every auto-augment op) moved the
`batch_prob` mask to `input.device` but not `input.dtype`. Since `batch_prob` is
float32, `batch_prob * op(x) + (1 - batch_prob) * x` promotes a `float16` /
`bfloat16` input to `float32` — the op output and the input are both low
precision, only the gate is not. `float32` / `float64` inputs are unaffected
(promotion goes up), so this only shows below float32.

`RandAugment`, `AutoAugment`, `TrivialAugment` and `AugMix` all route through this
method, so all of them silently upcast half-precision batches.

```python
import torch
from kornia.augmentation.auto import TrivialAugment

torch.manual_seed(0)
for dt in (torch.float64, torch.float32, torch.bfloat16, torch.float16):
    print(dt, "->", TrivialAugment()(torch.rand(4, 3, 16, 16, dtype=dt)).dtype)
# ... float64 -> float64,  float32 -> float32
# ... bfloat16 -> float32,  float16 -> float32   <-- dtype lost
```

The fix casts `batch_prob` to `input.dtype` as well as `input.device`. `float32`
output is byte-identical before/after.

Partially addresses #4153 (plain reference, not a closing keyword — #4153 tracks
the whole Linux CPU half-precision failure list and this clears nineteen of its
lines, not all of it). This fixes the `RuntimeError` behind the auto-augment CPU
half-precision xfails, so `tests/augmentation/test_auto_operation.py` entries are
removed from `testing/half_precision_xfails/`.

The manifest change is **not symmetric**, which a reader of `-9/+1` on the
`float16` file cannot infer:

- `cpu_bfloat16.txt` — all **10** entries removed; `test_auto_operation.py` is 24
  passed, 0 xfail, no XPASS.
- `cpu_float16.txt` — **9** entries removed and **1 added**:
  `+call  builtins.AssertionError  ...::TestAutoAugment::test_reproduce[cpu-float16]`.
  That case stops failing with a `RuntimeError` and starts failing an
  `AssertionError` — a real, smaller residual, not the same thing as fixed. So
  `float16` is 23 passed / 1 xfailed.

## What did you check?

- New test `test_operation_preserves_input_dtype` (in `test_auto_operation.py`),
  parametrised over `float16` / `bfloat16`, runs every op from `_find_all_ops()`
  and asserts `op(x).dtype == x.dtype`. Fails on `main`
  (`AssertionError: AutoContrast: torch.float32 != torch.float16`), passes with
  this change. It skips an op that raises `... not implemented for Half/BFloat16`
  (e.g. `grid_sample` has no half CPU kernel on torch 2.5.1) — a torch capability
  gap, not a kornia contract.
- `float32` numeric output is unchanged (`torch.equal`).
- `KORNIA_TEST_DTYPE=float16` and `=bfloat16` runs of
  `tests/augmentation/test_auto_operation.py` — 24 passed each (the previously
  xfailed `test_smoke` / `test_transform_mat` / `test_reproduce` / `test_sequential`
  now pass).

- `CHANGELOG.md` — entry added under `## Unreleased` → `### Bug fixes` citing #4210
  (user-visible: the four auto-augment policies stop upcasting half-precision
  batches), per `AGENTS.md`.

```text
$ KORNIA_TEST_DEVICE=cpu python -m pytest tests/augmentation/test_auto_operation.py tests/augmentation/test_base.py -q
58 passed

$ ruff check kornia/augmentation/auto/operations/base.py tests/augmentation/test_auto_operation.py
All checks passed!
$ ruff format --check <same files>
2 files already formatted
```

## How did AI help?

AI ran a metamorphic dtype-preservation sweep over `kornia.augmentation` that
flagged `TrivialAugment` (`float16 -> float32`), traced it to the shared
`OperationBase.forward` mask multiply, wrote the one-line fix and the regression
test, and verified fail-on-`main` / pass-with-fix plus float32 byte-identity. I
reviewed the root cause, the fix, and the test.


ducha-aiki:
## Review of #4210 at `01db00d4` — APPROVE, on a CHANGELOG entry

Setup: worktree at the PR head, `kornia/` swapped to the merge-base (`7c5db705`) for the fail-on-base pass. torch 2.14.0, python 3.11.

### The diagnosis holds and the pin discriminates

Fail-on-base, your test file against base `kornia/`:

```text
FAILED test_operation_preserves_input_dtype[cpu-low_dtype0] - AssertionError: AutoContrast: torch.float32 != torch.float16
FAILED test_operation_preserves_input_dtype[cpu-low_dtype1] - AssertionError: AutoContrast: torch.float32 != torch.bfloat16
2 failed, 22 deselected
```

At head, `tests/augmentation/test_auto_operation.py` is **24 passed**. Adding `dtype=input.dtype` to the `.to()` is the right shape of fix: the gate was the only float32 thing in `batch_prob * op(x) + (1 - batch_prob) * x`, and `float32`/`float64` inputs are unaffected because promotion goes the other way.

I specifically checked the `except RuntimeError` filter, because a substring guard on the very exception class the manifest was recording (`builtins.RuntimeError`) could have swallowed the failure it is meant to expose. It does not: on base the ops run fine and fail the **dtype assertion**, so the `"not implemented for"` `continue` never fires on the path that matters, and `checked` stays non-zero. Good.

### The manifest edits are verified — by your own CI, which is the authority here

Both `half-precision (float16)` and `half-precision (bfloat16)` legs are green on this PR, which is the real check: they run the strict `--verify-known-failures` profile on Linux/x86_64 at torch 2.9.1, and a stale line or an XPASS would fail them. My local reproduction agrees modulo platform (the manifests warn `os mismatch: expected 'Darwin', found 'Linux'` here): **23 passed / 1 xfailed** for float16, **24 passed** for bfloat16, no XPASS either way.

One thing worth a line in the body, because a reader of the diff cannot infer it: this is not a symmetric deletion. bfloat16 loses all ten entries, float16 loses nine and **gains** one —

```
+call  builtins.AssertionError  ...::TestAutoAugment::test_reproduce[cpu-float16]
```

— i.e. that case stops failing with a `RuntimeError` and starts failing an assertion instead, which is a real (smaller) residual and not the same thing as being fixed. Saying so explicitly stops the next person reading `-9/+1` as a typo.

### Two notes, neither blocking

- The CPU-only skip is the right call and it also keeps you clear of #4126: a test parametrized on a custom dtype argument rather than the `dtype` fixture escapes the CUDA half-precision isolation guards, which key on `dtype_name`. Because you skip non-CPU devices outright, no unisolated half kernel ever runs. Please leave that skip in place if the test grows.
- `_find_all_ops()` iterating every op is what makes this pin worth having — it will catch the next op that introduces its own dtype leak, not just `AutoContrast`.

### The one ask

Add a `CHANGELOG.md` entry under `## Unreleased` → `### Bug fixes`, citing #4210. This is a user-visible behavior change (`RandAugment`/`AutoAugment`/`TrivialAugment`/`AugMix` stop silently upcasting half-precision batches), and `AGENTS.md` asks for an entry per change rather than at release time. Approving on that.

Also worth adding "Partially addresses #4153" as a plain reference rather than a closing keyword — #4153 tracks the whole Linux CPU half-precision failure list and this clears nineteen of its lines, not all of it.

tritsystem:
Thanks for the thorough review.

- **CHANGELOG** (`c014b77`): added under `## Unreleased` → `### Bug fixes`, citing #4210 — states the user-visible change (the four auto-augment policies stop upcasting half-precision batches), that `float32`/`float64` are byte-identical, "Partially addresses #4153", and the asymmetric residual (bfloat16 loses all ten lines; float16 loses nine and keeps `test_reproduce[cpu-float16]`, now an `AssertionError` not a `RuntimeError`).
- **PR body**: "Related issue" now spells out the non-symmetric manifest edit (`-10` bfloat16 vs `-9/+1` float16) so `-9/+1` doesn't read as a typo, and marks #4153 as a plain reference rather than a closing keyword.
- The CPU-only skip and the `_find_all_ops()` iteration stay as-is per your notes.

No code changes — `base.py`, the test, and the manifests are untouched from `01db00d`.

tritsystem:
Note on the merge conflict: it is `CHANGELOG.md` only, and only the anchor line — `main` added the `ScaleSpaceDetector` `#4134` entry as the first item under `### Bug fixes` and this branch adds the auto-augment entry in the same position, so the two insertions collide. Resolution is just "keep both entries", in either order.

I can't rebase this branch onto current `main` myself: `main` now carries a change to `.github/workflows/pr_test_cpu.yml`, and my push token doesn't have `workflow` scope, so both `git push` and the fork-sync API reject it. If you merge with the web conflict editor (or "Rebase and merge"), keeping both `### Bug fixes` bullets is the whole resolution. Happy to redo it as a clean rebase if the branch is easier to take that way — just say the word.

Code, test, and manifests are unchanged from `01db00d`; the only new commit is the CHANGELOG entry (`c014b77`).

tritsystem:
Conflict resolved via the web editor (`9652f5b` merges `main` in). `CHANGELOG.md` keeps all three `### Bug fixes` entries — `#4134`, `#4168`, and this `#4210` one; no other file conflicted. `pre-commit.ci` is green again and the branch is now `MERGEABLE`.

No code changes from `01db00d` — `base.py`, the test, and the xfail manifests are untouched.

ducha-aiki:
## Re-review at `9652f5b8` — approve, nothing outstanding

The ask from the last round was to state that the manifest edit is **not symmetric**. You put it in the
CHANGELOG rather than the body, which is the better place for it, and it says exactly the right thing: all ten
`cpu_bfloat16` entries go, nine of the ten `cpu_float16` go, and
`TestAutoAugment::test_reproduce[cpu-float16]` is deliberately kept because it now fails on an `AssertionError`
rather than a `RuntimeError` — "a smaller residual, not a fix". That distinction is the one a future reader
needs, and calling this a partial address of #4153 rather than a closure is correct.

Re-verified against current `main` rather than carried over:

- `kornia/augmentation/auto/operations/base.py` and `tests/augmentation/test_auto_operation.py` are
  **byte-identical** to the revision I approved (`01db00d4`), and `testing/` is unchanged. The delta is the
  CHANGELOG entry plus the merge from `main`.
- **Head, after the merge:** `test_auto_operation.py` + `test_base.py` cpu float32 -> **58 passed**.
- **Half legs as CI runs them**, with the strict profile: float16 -> **23 passed, 1 xfailed**;
  bfloat16 -> **24 passed**. No XPASS in either, so the manifests as edited are exactly right.

38 checks green, `MERGEABLE`. Merging from my side.

One process note for next time, not a problem here: you merged `main` into the branch rather than rebasing. It
makes no difference to the result since this repo squashes, but a rebase keeps the delta reviewable — the
merge commit made `git diff <last-reviewed> <head>` show every unrelated change `main` had picked up in between,
and I had to diff the individual files to confirm your code had not moved.

ducha-aiki (approved):

PR #4211, opened 2026-09-04
ci: guard pinned runner availability

### LINKED ISSUE #4208: Pinned runner images (macos-15, windows-2022) have no expiry guard; retirement will block CI with an unrelated-looking error
### Describe the bug

Two runner images are pinned to an exact version rather than a `-latest` alias, and nothing in the repository notices when GitHub retires one:

| label | jobs | pinned in | why it is pinned |
|---|---|---|---|
| `macos-15` | 1 | `pr_test_cpu.yml:202` | `macos-latest` (macOS 26) has a paravirtualized GPU that cannot compile torch 2.14's Metal 4 shaders — 664 failures (#4207) |
| `windows-2022` | 8 | `pr_test_cpu.yml:108-120`, `scheduled_test_cpu.yml:162` | older torch wheels do not run on the `windows-2025` image |

Both pins are correct today and both are documented in place. The problem is what happens afterwards: GitHub retires hosted images on a rolling schedule, and when a label goes away the job stops scheduling. The run is then blocked by an error about missing runners, which names the label but not the decision behind it — and the person who hits it is not the person who wrote the comment explaining why the pin exists.

For `macos-15` specifically this is a gate failure, not a nuisance: `collector` lists `tests-mps` in its `needs` (`pr_test_cpu.yml:223`) and fails on `failure` **or** `cancelled`, so an unschedulable MPS job blocks every pull request, with an error message that points at runner availability rather than at an expired pin.

That the churn is real is already on the record here: `macos-14` had to be abandoned because it segfaults inside MPS kernels, and `windows-latest` moved to 2025 under the existing pins.

### Expected behavior

Something that fails *early and legibly* instead of at scheduling time. Rough options, cheapest first:

1. **A scheduled check that the pinned labels still resolve.** A tiny job in `scheduled_test_cpu.yml` that runs on each pinned image and does nothing but `echo`. It goes red on the deprecation warning window rather than on the day the label dies, and its name says which pin expired.
2. **Record an expiry alongside each pin.** GitHub announces retirement dates in advance; a comment with the announced date, plus a test in `tests/` asserting the date has not passed, turns a silent expiry into a normal red test with a message that explains the pin. `tests/test_half_precision_ci.py` already reads these workflow files, so there is a precedent for asserting on them.
3. **Nothing, but link the pins.** At minimum each pin comment should name the issue that explains the constraint, so the next person can tell in one click whether the reason still holds — for `macos-15` that is #4207, and re-testing `macos-latest` is exactly the action that would let the pin be dropped.

I would not recommend un-pinning either image to dodge this: both pins encode a measured incompatibility, and `macos-15` in particular is the difference between 14 and 670 failures.

Related: #4207 (why the macos-15 pin exists and what would let it be dropped), #4159.

victorwon2001:

Adds independent availability jobs for the pinned `macos-15` and `windows-2022` runner labels in the scheduled CPU workflow. The guards run on direct upstream `schedule` and manual `workflow_dispatch` executions, while repository and workflow identity checks keep them out of push and reusable `workflow_call` paths. Each guard runs directly on its exact pinned image with a single echo step, has `timeout-minutes: 5`, and has no dependency on `check-skip` or `pre-tests`.

A focused workflow-text regression test verifies that the PR MPS leg remains pinned to `macos-15` specifically inside `tests-mps`, that the PR and scheduled Windows 2.5.1 legs retain `windows-2022`, and that both availability guards keep their event/workflow identity conditions, exact labels, five-minute timeout, and skip independence.

Fixes #4208

## What did you check?

The focused pinned-runner invariant passed, and all repository pre-commit checks passed.

```text
$ pixi run test-module tests/test_half_precision_ci.py -k pinned_runner_images
$ pixi run pre-commit-all
```

## How did AI help?

AI helped inspect the workflow execution conditions, tighten the runner-pin regression test, and apply the maintainer-requested manual probe and timeout behavior. The focused regression and repository pre-commit checks were run afterward.

### DIFF
--- a/.github/workflows/scheduled_test_cpu.yml
+++ b/.github/workflows/scheduled_test_cpu.yml
@@ -21,6 +21,22 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pinned-runner-macos-15:
+    name: Pinned runner availability (macos-15)
+    if: github.event_name == 'schedule'
+    runs-on: macos-15
+    steps:
+      - name: Confirm macos-15 runner availability
+        run: echo "macos-15 runner is available"
+
+  pinned-runner-windows-2022:
+    name: Pinned runner availability (windows-2022)
+    if: github.event_name == 'schedule'
+    runs-on: windows-2022
+    steps:
+      - name: Confirm windows-2022 runner availability
+        run: echo "windows-2022 runner is available"
+
   check-skip:
     # For scheduled runs, check if only documentation files changed
     # If so, skip the expensive test runs
--- a/.github/workflows/scheduled_test_cpu.yml
+++ b/.github/workflows/scheduled_test_cpu.yml
@@ -23,16 +23,24 @@ concurrency:
 jobs:
   pinned-runner-macos-15:
     name: Pinned runner availability (macos-15)
-    if: github.event_name == 'schedule'
+    if: >-
+      (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') &&
+      github.repository == 'kornia/kornia' &&
+      startsWith(github.workflow_ref, 'kornia/kornia/.github/workflows/scheduled_test_cpu.yml@')
     runs-on: macos-15
+    timeout-minutes: 5
     steps:
       - name: Confirm macos-15 runner availability
         run: echo "macos-15 runner is available"
 
   pinned-runner-windows-2022:
     name: Pinned runner availability (windows-2022)
-    if: github.event_name == 'schedule'
+    if: >-
+      (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') &&
+      github.repository == 'kornia/kornia' &&
+      startsWith(github.workflow_ref, 'kornia/kornia/.github/workflows/scheduled_test_cpu.yml@')
     runs-on: windows-2022
+    timeout-minutes: 5
     steps:
       - name: Confirm windows-2022 runner availability
         run: echo "windows-2022 runner is available"
--- a/tests/test_half_precision_ci.py
+++ b/tests/test_half_precision_ci.py
@@ -878,12 +878,15 @@ def test_reusable_test_workflow_names_candidate_by_resolved_torch() -> None:
     assert "-${{ matrix.pytorch-version }}\n" not in workflow.partition("Upload candidate manifest")[2]
 
 
-def test_pinned_runner_images_have_scheduled_availability_guards() -> None:
+def test_pinned_runner_images_have_availability_guards() -> None:
     root = Path(project_conftest.__file__).parent
     pr_workflow = (root / ".github" / "workflows" / "pr_test_cpu.yml").read_text(encoding="utf-8")
     scheduled_workflow = (root / ".github" / "workflows" / "scheduled_test_cpu.yml").read_text(encoding="utf-8")
 
-    assert re.search(r"  tests-mps:.*?\n      os: macos-15\n", pr_workflow, re.DOTALL)
+    _, separator, tests_mps_tail = pr_workflow.partition("  tests-mps:\n")
+    assert separator, "missing tests-mps job"
+    tests_mps = re.split(r"\n(?=  [\w-]+:\n)", tests_mps_tail, maxsplit=1)[0]
+    assert "\n      os: macos-15\n" in tests_mps
     assert "          - os: windows-2022\n" in pr_workflow
     assert "          - os: windows-2022\n" in scheduled_workflow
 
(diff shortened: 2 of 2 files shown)


ducha-aiki:
## Review of #4211 at `03bab2ba` — APPROVE, on two small workflow additions

This is option 1 from #4208 and it is the right one to have picked: cheapest of the three, and it turns "no runner matched the labels" into a job whose *name* says which pin expired. Setup: worktree at the PR head, base `7c5db705`.

Fail-on-base 1/1:

```text
FAILED tests/test_half_precision_ci.py::test_pinned_runner_images_have_scheduled_availability_guards
  - AssertionError: missing scheduled availability guard for macos-15
```

At head the whole invariant file is **62 passed**. The test's coupling is well chosen — because it asserts `os: macos-15` in `pr_test_cpu.yml` *and* `runs-on: macos-15` in the guard, whoever eventually bumps the pin cannot bump one without the other. `assert "    needs:" not in guard` pins the skip-independence you describe in the body, which is the property that makes the guard survive a `check-skip` short-circuit.

### Two asks

**1. Let the guard run on `workflow_dispatch` too.** `scheduled_test_cpu.yml` triggers on `schedule`, `push` to `main`, `workflow_dispatch` and `workflow_call`, but the guard is gated on `github.event_name == 'schedule'` alone. The moment anyone actually wants this answer is while they are investigating a suspected retirement — and right now they would have to wait for the 04:00 cron. One character of extra condition:

```yaml
    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
```

The `push`-to-`main` exclusion is worth keeping; a macOS runner minute is billed at 10x and this does not need to fire on every merge.

**2. Give both guards a `timeout-minutes`.** No job anywhere in `.github/workflows/` sets one today, so every job inherits GitHub's 6-hour default. For an `echo` that matters more than it sounds: if a retired label leaves the job **queued** rather than failing to schedule outright, this guard reports as a six-hour stuck job instead of a fast, legible red — which is the failure mode #4208 is trying to get rid of. `timeout-minutes: 5` is generous for one `echo` step.

That default is not hypothetical here. Earlier today the `half-precision (float16)` leg on an unrelated PR hung and was cancelled at exactly **6 h 0 m 30 s** (02:12:13 → 08:12:43), against a 6–8 minute norm for that job, and the only signal was a red check with no failing test. I am filing that separately, but it is the same missing guard rail, and adding it to the two jobs you are introducing is free.

### Nit

`assert re.search(r"  tests-mps:.*?\n      os: macos-15\n", pr_workflow, re.DOTALL)` does not actually anchor the pin to the `tests-mps` job — a non-greedy `.*?` under `DOTALL` only requires `os: macos-15` to appear *somewhere after* `tests-mps:`. It passes today and it would keep passing if the pin migrated to a different job below. Not worth another round on its own; fold it in if you touch the file for the two asks above.

ducha-aiki:
## Re-review at `fdbef02b` — approve, all three asks addressed

**(a) `workflow_dispatch`.** Done, and I checked it is real rather than defensive: `scheduled_test_cpu.yml`'s
`on:` block does carry `workflow_dispatch`, so the guard is now reachable at the moment you actually want the
answer — while investigating — instead of only at 04:00.

**(b) `timeout-minutes`.** `timeout-minutes: 5` on both guards. That is the whole point of the ask: a retired
label that leaves a job *queued* rather than unschedulable would otherwise report as a 6-hour stuck job with no
signal, which is exactly the failure mode #4214 was filed for.

**(c) the regex nit.** The `partition` + `re.split(r"\n(?=  [\w-]+:\n)")` rewrite genuinely anchors the pin to
the `tests-mps` job now, rather than accepting `os: macos-15` anywhere later in the file. It also matches the
idiom the rest of that module already uses for the guard blocks.

**Measured:** `tests/test_half_precision_ci.py` at head -> **62 passed**. Fail-on-base 1/1 on `7c5db705`:
`AssertionError: missing scheduled availability guard for macos-15`.

### On the two conditions you added that I did not ask for

`github.repository == 'kornia/kornia'` is a standard fork guard and I have no objection.
`startsWith(github.workflow_ref, 'kornia/kornia/.github/workflows/scheduled_test_cpu.yml@')` is doing something
real — the workflow also declares `workflow_call`, and under a reusable-workflow call `github.event_name` is
inherited from the caller, so without it a dispatch of some future caller would burn macos-15 and windows-2022
minutes on an availability probe. Nothing calls it today (`grep` finds only a comment in `upstream_probe.yml`),
so it is speculative, but it is cheap and correct. Keep it.

Optional, take it or leave it: the test now pins the four-line `if:` expression verbatim, including the
repository name twice, so a purely cosmetic reflow of that YAML (folding it to one line, reordering the
conjuncts) reds a test that is asserting behaviour. Asserting the three components separately would survive that.
The file already does exact-string pins elsewhere, so this is consistent with its style and I am not asking you
to change it.

Nothing outstanding from me. 36 checks green, `MERGEABLE`.

ducha-aiki (approved):

PR #4212, opened 2026-09-04
fix(rt_detr): use device-agnostic map_location in from_pretrained

li-lizhe:
## Problem

`RTDETR.from_pretrained()` hard-codes `map_location = "cuda:0" if torch.cuda.is_available() else "cpu"`. On systems where the primary accelerator is not CUDA (e.g. Ascend NPU, Intel XPU, Apple MPS), this either loads weights to the wrong device or falls back to CPU, causing unnecessary copies or device-mismatch errors.

## Root cause

The `from_pretrained` method in `kornia/models/rt_detr/model.py` uses `torch.cuda.is_available()` to decide the target device, which is a CUDA-specific check that doesn't account for other accelerators.

## Fix

Replace the hard-coded heuristic with simply `map_location="cpu"` — the standard PyTorch pattern for loading pretrained weights. This is completely device-agnostic, works on every torch version kornia supports (including 2.5.1), and avoids the wasteful round-trip from CUDA-to-CPU-to-final-device that the original code produced even on CUDA.

```python
state_dict = load_state_dict_from_url(URLs[model_name], map_location="cpu")
```

## Verification

- **Ascend 910B2** (torch 2.14, torch_npu): loads correctly to CPU first, then torch will move to NPU when the model is placed on its target device.
- **CUDA**: same as before — `map_location="cpu"` is fully compatible with CUDA workflows.
- **CPU-only**: works as expected with no accelerator needed.

### DIFF
--- a/kornia/models/rt_detr/model.py
+++ b/kornia/models/rt_detr/model.py
@@ -247,9 +247,12 @@ class RTDETR(ONNXExportMixin, ModelBase[RTDETRConfig]):
         if model_name not in URLs:
             raise ValueError(f"No pretrained model for '{model_name}'. Please select from {list(URLs.keys())}.")
 
-        state_dict = load_state_dict_from_url(
-            URLs[model_name], map_location="cuda:0" if torch.cuda.is_available() else "cpu"
-        )
+        # Load weights on the current accelerator (CUDA, MPS, XPU, NPU, ...), falling
+        # back to CPU when none is available. Hard-coding "cuda:0" here would force the
+        # state dict onto CUDA even on machines that expose a different accelerator
+        # (e.g. Ascend NPU or Intel XPU) whose torch build reports no CUDA.
+        device = str(torch.accelerator.current_accelerator()) if torch.accelerator.is_available() else "cpu"
+        state_dict = load_state_dict_from_url(URLs[model_name], map_location=device)
 
         def map_name(old_name: str) -> str:
             new_name = old_name
--- a/kornia/models/rt_detr/model.py
+++ b/kornia/models/rt_detr/model.py
@@ -247,12 +247,14 @@ class RTDETR(ONNXExportMixin, ModelBase[RTDETRConfig]):
         if model_name not in URLs:
             raise ValueError(f"No pretrained model for '{model_name}'. Please select from {list(URLs.keys())}.")
 
-        # Load weights on the current accelerator (CUDA, MPS, XPU, NPU, ...), falling
-        # back to CPU when none is available. Hard-coding "cuda:0" here would force the
-        # state dict onto CUDA even on machines that expose a different accelerator
-        # (e.g. Ascend NPU or Intel XPU) whose torch build reports no CUDA.
-        device = str(torch.accelerator.current_accelerator()) if torch.accelerator.is_available() else "cpu"
-        state_dict = load_state_dict_from_url(URLs[model_name], map_location=device)
+        # Load weights on CPU and let the caller move the returned model to its
+        # target device with .to(device). Loading straight onto an accelerator is
+        # a wasted round-trip anyway: the model is built on CPU and load_state_dict
+        # copies into existing CPU parameters, so the map_location device is
+        # discarded. Hard-coding "cuda:0" (as before) also forces the state dict
+        # onto CUDA even on machines that expose a different accelerator (e.g.
+        # Ascend NPU or Intel XPU) whose torch build reports no CUDA.
+        state_dict = load_state_dict_from_url(URLs[model_name], map_location="cpu")
 
         def map_name(old_name: str) -> str:
             new_name = old_name


ducha-aiki:
## Review of #4212 at `7546d4b9` — CHANGES REQUESTED, `torch.accelerator` is below kornia's floor

The problem you describe is real — `"cuda:0" if torch.cuda.is_available() else "cpu"` is genuinely CUDA-centric, and the NPU/XPU case it ignores is a legitimate gap. Two findings, one blocking.

### 1. Blocking: `torch.accelerator` does not exist in torch 2.5.1

The PR body says the API is *"available since PyTorch 2.5 (kornia's minimum version)"*. It is not — it landed in **2.6.0**:

```console
$ gh api "repos/pytorch/pytorch/contents/torch/accelerator/__init__.py?ref=v2.5.1"
{"message":"Not Found","status":"404"}
$ gh api "repos/pytorch/pytorch/contents/torch/accelerator/__init__.py?ref=v2.6.0" --jq .size
5518
```

There is no `accelerator` entry of any kind under `torch/` at `v2.5.1` (checked the whole directory listing, not just that path). kornia declares `torch>=2.5.1` (`pyproject.toml:28`) and PR CI runs 2.5.1 in eight matrix legs (`pr_test_cpu.yml:89`), so on the declared floor this line raises `AttributeError: module 'torch' has no attribute 'accelerator'` before it can decide anything.

**CI being green does not clear this.** Nothing in `tests/` calls `RTDETR.from_pretrained` — it needs a weights download — so no job on any torch version executes the changed line.

### 2. More interesting: the device never survives the function anyway

`from_pretrained` builds the model with `RTDETR.from_name(...)` (CPU) and then calls `model.load_state_dict(...)`, which *copies into the existing parameters*. The `map_location` device is discarded a few lines later. Measured at your head:

```text
kornia: /private/tmp/wt-4212/kornia/__init__.py | torch 2.14.0
from_name param devices: {'cpu'}
state dict device sample: mps:0
param devices AFTER load_state_dict of an MPS state dict: {'cpu'}
```

So today the weights are uploaded to the accelerator and immediately copied back to CPU, and the returned model is on CPU either way. `map_location` is not choosing where the model lands; it is choosing where a pointless round-trip happens. That is true of `"cuda:0"` on `main` too — the CUDA path has always paid for an upload it throws away.

Which means the smallest fix that actually solves the problem in your title is:

```python
state_dict = load_state_dict_from_url(URLs[model_name], map_location="cpu")
```

That is version-safe on 2.5.1, needs no capability probe, removes the wasted transfer instead of relocating it, and gives your Ascend/XPU users exactly what they want: no CUDA assumption, no device-mismatch error, and a model they then move with `.to(device)` like every other kornia model. It also matches what `from_name` already implies — this constructor does not promise a device.

If you would rather keep accelerator-awareness, it needs both a `hasattr(torch, "accelerator")` guard for the 2.5.x floor **and** a `model.to(device)` after the load, or the device still means nothing. I would take the `"cpu"` version.

### Scope note

`torch.cuda.is_available()` appears twice more in `kornia/` — `core/utils.py:63` and `feature/lightglue.py:695` — but neither is a `map_location`; they are capability checks, so this PR is right to leave them alone.

li-lizhe:
Thank you for the thorough review, @ducha-aiki! I've updated the fix per your recommendation to use `map_location="cpu"` — this is version-safe on torch 2.5.1, removes the wasteful round-trip, and lets callers move the model with `.to(device)` as usual. The updated commit has been pushed to the same branch. Key changes:

1. Removed `torch.accelerator` usage (not available in torch 2.5.1)
2. Changed `map_location=device` → `map_location="cpu"` (version-safe, device-agnostic)
3. Updated the docstring to explain the rationale

ducha-aiki:
## Re-review at `1ce80ce9` — the code is fixed; the PR body still describes the rejected approach

`map_location="cpu"` is the right resolution and the new comment states both reasons correctly (the
`torch.accelerator` version floor, and that the device never survives the function anyway because `from_name`
builds on CPU and `load_state_dict` copies *into* the existing CPU parameters). `grep -n "torch.accelerator"` over
`kornia/` and `tests/` at head: **no hits left**. That clears the blocker.

### The one thing left

The PR body is unchanged and still argues for the version that was withdrawn — it has a "## Fix" section that
reads:

> Replace the hard-coded heuristic with `torch.accelerator.current_accelerator()`, available since PyTorch 2.5
> (kornia's minimum version). This is completely device-agnostic and discovers the current accelerator at runtime.

followed by the old two-line snippet, and a "## Verification" section describing `map_location="npu"` loading
weights directly to NPU. None of that is what the branch does now, and the parenthetical is the specific claim
that turned out to be false: `torch/accelerator/__init__.py` is a 404 at `v2.5.1` and 5518 bytes at `v2.6.0`,
while `pyproject.toml:28` declares `torch>=2.5.1` and PR CI runs 2.5.1 in eight legs.

kornia squashes, so this body becomes the commit message. Please rewrite the Fix and Verification sections to
describe `map_location="cpu"`: the NPU/XPU problem is still solved (nothing is forced onto CUDA), the transfer is
removed rather than relocated, and it works on every supported torch. Keep the original problem statement — it is
accurate and it is why the change is worth making.

Nothing else outstanding. 37 checks green, `MERGEABLE`. Approving as soon as the body matches the diff.

li-lizhe:
Hi @ducha-aiki, thanks for the careful re-review! I've updated the PR description to match the actual code — it now describes the `map_location="cpu"` approach (the resolved `1ce80ce9` state) instead of the previously-rejected `torch.accelerator` heuristic. The fix itself is unchanged: just use the standard device-agnostic `map_location="cpu"` in `from_pretrained`. Please re-review when you get a chance.

ducha-aiki (approved):

Issue #4175, opened 2026-09-02
Bug: `Boxes.merge` keeps stale list padding, so `to_tensor()` drops the merged boxes


ducha-aiki:
## Summary

`Boxes.merge` builds its result with `obj = self.clone()` (`kornia/geometry/boxes.py:342` at `8d685a37`) and then overwrites `obj._data` with the concatenation. `clone()` copies `_N` — the per-element padding counts recorded when the object was built from a list — but the merge just changed how many boxes each row holds, so `_N` is stale afterwards.

`to_tensor` trims with those counts:

```python
boxes = [torch.nn.functional.pad(o, ... + [0, -n]) for o, n in zip(boxes, self._N)]
```

so a list-backed `Boxes` silently **drops the boxes that were just merged in**, and returns padding entries in their place. Nothing raises.

## Reproduction

```python
import torch
from kornia.geometry.boxes import Boxes

first = torch.tensor([[[1.0, 2.0], [4.0, 2.0], [4.0, 3.0], [1.0, 3.0]]])   # 1 box
second = torch.cat([first, first])                                          # 2 boxes
lb = Boxes([first, second])          # _N == [1, 0], data (2, 2, 4, 2)

merged = lb.merge(Boxes(torch.zeros(2, 1, 4, 2) + 7.0))
print(merged._N, tuple(merged.data.shape))
for i, row in enumerate(merged.to_tensor("xyxy")):
    print(i, row.tolist())
print(merged.get_boxes_shape()[0].tolist())
```

## Observed

```
[1, 0] (2, 3, 4, 2)
0 [[1.0, 2.0, 5.0, 4.0], [0.0, 0.0, 1.0, 1.0]]
1 [[1.0, 2.0, 5.0, 4.0], [1.0, 2.0, 5.0, 4.0], [7.0, 7.0, 8.0, 8.0]]
[[2.0, 1.0, 1.0], [2.0, 2.0, 1.0]]
```

Row 0 should be its original box plus the merged `[7., 7., 8., 8.]`. Instead the merged box is trimmed off and the phantom padding box `[0., 0., 1., 1.]` is returned in its place — the data is there (`to_tensor(..., as_padded_sequence=True)` shows all three), only the bookkeeping says to cut it. Row 1 had `_N == 0` and is correct, so the corruption is per-element and depends on how ragged the original list was.

## Expected

[...]

PR #4213, opened 2026-09-04
Fix/4168 boxes merge

### LINKED ISSUE #4168: Bug: `Boxes.merge` concatenates unbatched `(N, 4, 2)` boxes along the vertex axis

alinaqvi1129:
## What does this PR do?

Fixes `Boxes.merge` concatenating unbatched `(N, 4, 2)` boxes along the vertex axis instead of the box axis.

For unbatched boxes, `dim=1` incorrectly produces `(N, 8, 2)` when merging two `(N, 4, 2)` tensors. This changes the concatenation to `dim=-3`, which is the box axis for both unbatched `(N, 4, 2)` and batched `(B, N, 4, 2)` boxes.

Also updates the `merge` docstring and adds regression coverage for both unbatched and batched cases.

## Testing

* `test_boxes.py -k merge`: **1 passed**
* `test_boxes.py`: **39 passed**
* `tests/augmentation -k mosaic`: **6 passed, 5565 deselected**
* `ruff check` on modified Python files: **passed**

### Fail-on-base

With the original `dim=1` implementation, the unbatched regression test fails:

```text
shape:     torch.Size([1, 8, 2])
to_tensor: tensor([[1., 2., 9., 8.]])

AssertionError: FAIL (base): shape is torch.Size([1, 8, 2]), expected (2, 4, 2)
```

With the fix, the result is:

```text
shape:     torch.Size([2, 4, 2])
to_tensor: tensor([[1., 2., 5., 4.],
                   [6., 3., 9., 8.]])
```

The batched merge behavior remains unchanged.

## Changelog

Added an entry under `Unreleased` → `Bug fixes`.

Fixes #4168

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -241,6 +241,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Constructing `ScaleSpaceDetector` with `compile_modules` no longer permanently mutates the process-global
   `torch._dynamo.config.capture_dynamic_output_shape_ops` setting. (#4134)
 
+* Fix `Boxes.merge` concatenating along the vertex axis instead of the box axis
+  for unbatched `(N, 4, 2)` input. The old `dim=1` was correct for batched
+  `(B, N, 4, 2)` data (where dim 1 is the box axis) but wrong for 3-D tensors
+  (where dim 1 is the vertex dimension); `dim=-3` is correct in both cases and
+  the behaviour of the batched path is byte-identical. Also clear the stale
+  ``_N`` list-padding metadata on both the inplace and non-inplace paths after
+  every merge so that ``to_tensor`` counts the merged boxes correctly instead of
+  silently truncating them (#4168).
+
 * `Boxes3D.to_tensor` and `Boxes3D.get_boxes_shape` are differentiable again (#1396). Both reduce a box's 8
   vertices to its min/max corner with `amin`/`amax`, which is a genuine kink -- not differentiable in the
   classical sense -- wherever multiple vertices exactly tie for an axis extremum. Every face of an
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -347,21 +347,33 @@ class Boxes:
 
         See the Convention block on :class:`~kornia.geometry.boxes.Boxes`.
 
-        Say, current instance holds :math:`(B, N, 4, 2)` and the incoming boxes holds :math:`(B, M, 4, 2)`,
-        the merge results in :math:`(B, N + M, 4, 2)`.
+        For batched boxes, if the current instance holds :math:`(B, N, 4, 2)` and
+        the incoming boxes holds :math:`(B, M, 4, 2)`, the merge results in
+        :math:`(B, N + M, 4, 2)`.
+
+        For unbatched boxes, if the current instance holds :math:`(N, 4, 2)` and
+        the incoming boxes holds :math:`(M, 4, 2)`, the merge results in
+        :math:`(N + M, 4, 2)`.
 
         Args:
             boxes: 2D boxes.
             inplace: do transform in-place and return self.
 
+        Note:
+            Any list-padding metadata (``_N``) is cleared after the merge because
+            the merged tensor has a new total box count that is incompatible with
+            the original per-image lengths recorded before the merge.
+
         """
-        data = torch.cat([self._data, boxes.data], dim=1)
+        data = torch.cat([self._data, boxes.data], dim=-3)
         if inplace:
             self._data = data
+            self._N = None
             return self
 
         obj = self.clone()
         obj._data = data
+        obj._N = None
         return obj
 
     def index_put(
--- a/tests/geometry/test_boxes.py
+++ b/tests/geometry/test_boxes.py
@@ -542,8 +542,6 @@ class TestBoxes2D(BaseTester):
         merged_list = list_boxes.merge(extra_batched)
         assert merged_list._N is None, "_N must be None after merge"
 
-
-
     def test_compute_area(self):
 
         # Rectangle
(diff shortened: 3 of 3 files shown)


ducha-aiki:
## Review of #4213 at `9f64639c` — CHANGES REQUESTED, small; the `dim=-3` half is right, the `_N` half is undeclared scope

Setup: worktree at the PR head, `kornia/` swapped to the merge-base (`7c5db705`) for the fail-on-base pass. torch 2.14.0, python 3.11.

### The `dim=-3` fix is correct and the pin discriminates

Fail-on-base with your test file on base `kornia/`:

```text
FAILED tests/geometry/test_boxes.py::TestBoxes2D::test_merge[cpu-float32]
  - AssertionError: Unbatched merge shape: expected (2, 4, 2), got torch.Size([1, 8, 2])
1 failed, 1 passed
```

At head: `tests/geometry/test_boxes.py` **127 passed** (cpu, float32+float64), **64 passed / 5 skipped** (mps, float32), and `tests/augmentation -k "mosaic or bbox or boxes"` **25 passed**. The only in-tree caller is `mosaic.py:146`, which merges 4-D data where `dim=-3 == dim=1`, so it is byte-identical there. No complaints about this part.

### The `_N = None` hunk fixes a different issue, and it is neither cited nor neutral

That hunk is not #4168 — it is **#4175**, which the body and the CHANGELOG entry do not mention. It is worth having, but it is not the "clear stale metadata" no-op the docstring `Note` implies. Measured, same input on both sides (#4175's own repro):

```text
BASE  _N: [1, 0] | to_tensor("xyxy") -> list, per-row shapes [(2, 4), (3, 4)]
        row 0  [[1, 2, 5, 4], [0, 0, 1, 1]]                      <- merged box DROPPED
HEAD  _N: None   | to_tensor("xyxy") -> Tensor, shape (2, 3, 4)
        row 0  [[1, 2, 5, 4], [0, 0, 1, 1], [7, 7, 8, 8]]        <- merged box back, pad row now a real box
```

Two consequences the `Note` does not state:

1. **`to_tensor()` changes return type** for a list-backed object after a merge — `list[Tensor]` on base, a dense `Tensor` at head. `as_padded_sequence=True` and `=False` also stop differing. That is API-visible, not bookkeeping.
2. **The zero pad row is now exported as a real box.** That is precisely the defect #4179 describes for `Boxes.__getitem__` ("a sliced list-backed object exports its zero pad rows as boxes") — and #4194 is open right now to stop `Boxes` doing exactly that. So `merge` would start doing what `__getitem__` is being fixed not to do.

To be fair to the change: base loses real boxes and head only gains phantom ones, and missing data is the more damaging direction, so head **is** an improvement. I am not asking you to revert it blindly. But #4175 offered two dispositions — update the pad counts so they stay relative to the new row lengths, or raise — and `None` is a third one that neither the issue nor this PR argues for. The reason none of the three is obvious is that after a merge the padding is no longer trailing: a row is `[real, pad, new]`, and no trailing-trim count can describe that. The dispositions that actually work are re-packing to `[real, new, pad]` and keeping `_N`, or raising.

### What I am asking for

1. Cite **#4175** in the body and the CHANGELOG entry, **without** a closing keyword — this improves it but does not close it, since the phantom pad row survives.
2. Say in the `Note` what clearing `_N` does to the caller: the object stops being list-backed, and `to_tensor()` returns a dense tensor that includes the former padding rows instead of a ragged list.
3. One sentence in the body on why `None` over re-pack-or-raise. If you would rather re-pack, that is the fix that closes #4175 outright and I am happy to review it here; if you would rather keep this PR to #4168 alone, dropping the `_N` hunk is also fine and leaves #4175 honestly open.

### Nit

`tests/geometry/test_boxes.py:66` adds a blank line after `def test_compute_area(self):` that has nothing to do with this change. The repo's pinned `ruff-format` (v0.16.4) accepts it, so it is not a CI problem — just unrelated churn in the diff.

ducha-aiki:
## Rebased onto `main` for you — the conflict is gone, my review above still stands

This went `CONFLICTING` when `main` moved past it, which meant no `pull_request` workflow ran at all: the 40 green checks on `9f64639c` had run against an older `main` and no longer said anything about merging. I rebased the branch onto `main` (`8d2fe3bf`) and force-pushed it to your fork, so CI is running again.

The only conflict was `CHANGELOG.md` — your bullet and #4134's landed on the same line. Both are kept, #4134 first. Your code is untouched: `81 insertions(+), 3 deletions(-)` before and after the rebase, and `tests/geometry/test_boxes.py` is **127 passed** at `--dtype=float32,float64` on the rebased head.

New head is **`9cfe20b5`** — run `git fetch && git reset --hard origin/fix/4168-boxes-merge` before your next commit, or you will re-push the old history.

The asks in [my review above](https://github.com/kornia/kornia/pull/4213#issuecomment-5539143238) are unchanged and none of them is affected by the rebase.

ducha-aiki (approved):

PR #4216, opened 2026-09-04
Fix list-backed Boxes.merge padding metadata

### LINKED ISSUE #4175: Bug: `Boxes.merge` keeps stale list padding, so `to_tensor()` drops the merged boxes

cqxswbc2:

Preserves list-backed `Boxes` semantics when merging boxes. Each batch row is repacked as existing real boxes, incoming real boxes, then trailing padding, and `_N` retains the combined padding count instead of being cleared.

This prevents old padding entries from being exposed as real boxes while ensuring newly merged boxes are not trimmed by `to_tensor()`. It supports list-backed inputs on either side, both list-backed inputs, and the in-place path. The class documentation and changelog now describe the corrected behavior.

Fixes #4175

## What did you check?

```text
$ python -m pytest tests/geometry/test_boxes.py -q
69 passed

$ uvx pre-commit run --files kornia/geometry/boxes.py tests/geometry/test_boxes.py CHANGELOG.md
All hooks passed
```

I also reproduced the current-main failure (a list-backed merge returns a tensor containing the old padding row), exercised unequal and empty row lengths, verified both operand orders, and checked that gradients propagate through the repacking path.

## How did AI help?

AI helped trace the `_N` padding semantics, draft the implementation and regression cases, and compare the result against the issue reproduction. A separate AI reviewer performed an adversarial review; its findings about stale class documentation and the missing reverse operand-order test were incorporated. The final diff was then re-run through the focused test module and all pre-commit hooks for the changed files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -245,10 +245,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   for unbatched `(N, 4, 2)` input. The old `dim=1` was correct for batched
   `(B, N, 4, 2)` data (where dim 1 is the box axis) but wrong for 3-D tensors
   (where dim 1 is the vertex dimension); `dim=-3` is correct in both cases and
-  the behaviour of the batched path is byte-identical. Also clear the stale
-  ``_N`` list-padding metadata on both the inplace and non-inplace paths after
-  every merge so that ``to_tensor`` counts the merged boxes correctly instead of
-  silently truncating them (#4168).
+  the behaviour of the batched path is byte-identical. For list-backed inputs,
+  each batch row is now repacked with its real boxes before all trailing padding,
+  and the combined per-image padding counts are retained. This makes ``to_tensor``
+  trim only padding instead of dropping merged boxes or exposing old padding as
+  real boxes (#4168, #4175).
 
 * `Boxes3D.to_tensor` and `Boxes3D.get_boxes_shape` are differentiable again (#1396). Both reduce a box's 8
   vertices to its min/max corner with `amin`/`amax`, which is a genuine kink -- not differentiable in the
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -246,9 +246,9 @@ class Boxes:
           or rejected by its first box alone and the remaining boxes are cast to that dtype. For a single tensor,
           :meth:`from_tensor` silently casts integer input to ``float32``. For a list, it converts each element
           independently and then pads into the first converted element's dtype, recasting the remaining elements.
-        - :meth:`merge` concatenates batched boxes along the box axis, while :meth:`index_put` replaces selected
-          coordinates. Both methods are non-mutating by default. :meth:`merge` is currently unsafe for unbatched
-          and list-backed objects because it respectively concatenates vertices and preserves stale padding metadata.
+        - :meth:`merge` concatenates boxes along the box axis and repacks list-backed batch rows so their padding
+          remains at the end, while :meth:`index_put` replaces selected coordinates. Both methods are non-mutating
+          by default.
 
     .. warning::
         The inclusive ``+1`` arithmetic differs from torchvision, COCO, and albumentations and is tracked as a
@@ -259,10 +259,8 @@ class Boxes:
         :func:`~kornia.geometry.bbox.nms` convention is
         `#4008 <https://github.com/kornia/kornia/issues/4008>`_. The differing ``width``/``height`` argument order
         of :func:`~kornia.geometry.bbox.bbox_to_mask` and :meth:`to_mask` is tracked in
-        `#4014 <https://github.com/kornia/kornia/issues/4014>`_. Unbatched and list-backed :meth:`merge` are tracked
-        in `#4168 <https://github.com/kornia/kornia/issues/4168>`_ and
-        `#4175 <https://github.com/kornia/kornia/issues/4175>`_, respectively. The integer-input policy split
-        between the constructor and :meth:`from_tensor` is
+        `#4014 <https://github.com/kornia/kornia/issues/4014>`_. The integer-input policy split between the
+        constructor and :meth:`from_tensor` is
         `#4012 <https://github.com/kornia/kornia/issues/4012>`_. With ``validate_boxes=True``, vertex modes remain
         unvalidated, and ``'vertices'`` also subtracts one from fixed vertex positions, potentially deforming the
         input rather than rejecting it; this is tracked in `#4177 <https://github.com/kornia/kornia/issues/4177>`_.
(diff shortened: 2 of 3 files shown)


ducha-aiki:
## Approve

Reviewed at `e64e2a3d` against merge-base `8db1499a`, torch 2.14.0 / py311 / macOS arm64. This closes #4175
correctly, and the repack is the only shape of fix that can — I had concluded the same independently while
labelling that issue, so let me record the derivation here so it does not get re-litigated.

**Why no trailing-trim count could have closed this.** `_N` holds *pad* counts (`boxes.py:43`,
`stats = [max_N - box.shape[0] for box in boxes]`), and `to_tensor` consumes them as a negative pad, i.e. a
trailing trim (`:711-712`). Merging `M` boxes into every row raises both each row's length and the maximum by
`M`, so the arithmetically correct `_N` after a merge is **unchanged** — `[1, 0]` stays `[1, 0]`, which is exactly
what your test asserts. The issue's own "Expected" (`[n + M for n in self._N]`) is wrong. What actually breaks is
*where* the padding sits: `torch.cat(..., dim=-3)` appends after it, so the row becomes `[real, pad, new]` and a
trailing trim cuts the merged boxes instead of the pad. Re-packing at each row's own offset — what you did — or
raising for a list-backed receiver were the only two options.

### Measured

- **Fail-on-base 1/1.** Your `test_merge` on `8db1499a` with `main`'s `kornia/`: `assert None == [1, 0]`.
- **Head:** `tests/geometry/test_boxes.py` cpu float32+float64 -> **127 passed**.
- **#4175's own repro** at head: `_N == [1, 0]` and `to_tensor("xyxy")` returns
  `[[[1,2,5,6],[10,10,12,12]], [[0,0,3,3],[4,4,7,7],[20,20,22,22]]]` — the merged box present in both rows, the
  pad trimmed, no phantom row. On `main` the merged box is dropped.
- **Dense path is untouched:** `d1.merge(d2).data` is `torch.equal` to `torch.cat([d1.data, d2.data], dim=-3)`
  with `_N is None`, so the sole in-tree caller (`mosaic.py:146`, 4-D) is byte-identical — `tests/augmentation
  -k mosaic` **6 passed**.
- **Gradients** flow through the repack (`requires_grad` on `_data`, non-zero grad after `merge(...).data.sum()`).
- **Empty rows:** `from_tensor([torch.zeros(0,4), one_box])` -> `_N == [1, 0]`, merged -> rows of 1 and 2. Correct.
- **Batch mismatch** raises your `ValueError: Batch size mismatch. Got 2 for self and 3 for boxes.`
- **List-backed `self` with an unbatched `other`** — the case I went looking for, since the new branch indexes
  `boxes.data[i]` per row and a 3-D operand would make that a vertex index — raises
  `RuntimeError: Tensors must have same number of dimensions: got 3 and 2` inside the per-row `cat`. It does not
  silently produce garbage, which was the failure mode I was worried about.
- `Boxes3D` needs nothing: it has neither `_N` nor its own `merge`.

**Documented-defect gate.** `grep -rnE "#4175|issues/4175|#4168|issues/4168" kornia/ tests/` at head: **zero
hits**. Worth saying explicitly that this also cleans up something that is not yours: `main` still carries the
`#4168` tracking sentence at `kornia/geometry/boxes.py:263` even though #4168 was **closed** by #4213 — a stale
reference to a closed issue that the AGENTS.md grep gate exists to catch. Your diff removes it as a side effect.

### Optional polish, not blocking

1. `merge` now raises `ValueError` on a batch-size mismatch when either operand is list-backed. That is a new
   public failure mode and there is no `Raises:` block on the method. One line would do it.
2. The list-backed branch is a Python loop over the batch (`B` `torch.cat`s plus a `stack`). Fine for the ragged
   path this only runs on, and I would not vectorise it speculatively — just noting it is `O(B)` host-side so
   nobody is surprised later.

Neither blocks the merge.

ducha-aiki (approved):

PR #4217, opened 2026-09-04
perf(feature): drop the constant homogeneous row from laf_to_boundary_points

ducha-aiki:

Speeds up `laf_to_boundary_points` — and with it `laf_is_inside_image`, which calls it with
`n_pts=12` on every detector forward — by removing a constant homogeneous row that existed only so
the result could be divided by a coordinate that is always exactly `1`.

This is the last un-fixed item from §4 of the `kornia.feature` gap ledger, the one #4119's harness
ranked worst on every machine measured.

### The diagnosis that was written up first was the wrong one

#4119's README says the cost is the basis: the function builds its `linspace`/`ones`/`tensor`
basis with no `device=`, `.expand()`s it to `(B*N, n_pts, 3)` and only then calls `.to(device)`, so
every call materializes — and on an accelerator transfers — ~12 MiB at `N = 20000` for a tensor
with `n_pts` distinct rows. That is true, and it is a memory problem. Fixing **only** it measures
**1.00× on CPU**.

The actual cost is a gemm-shape cliff. To divide by the homogeneous coordinate, the function
appended a constant `[0, 0, 1]` row to every LAF and ran a `(B, 3, 3) × (B, 3, n_pts)` `bmm`.
torch's CPU batched gemm falls off a fast path at that third row (B=1, N=20000, torch 2.14, M1):

| `bmm` M dimension | 1 thread | 4 threads | 8 threads |
| --- | --: | --: | --: |
| M = 1 | 0.94 ms | 0.33 ms | 0.32 ms |
| M = 2 | 1.85 ms | 0.65 ms | 0.61 ms |
| **M = 3** | **14.79 ms** | **14.87 ms** | **14.83 ms** |
| M = 4 | 16.24 ms | 16.26 ms | 16.27 ms |

23× for 1.5× the arithmetic, and thread-count independent. The homogeneous coordinate of every
output point is then exactly `1.0` — I checked that directly in all four dtypes — so
`convert_points_from_homogeneous` divides by `1.0` and is the identity.

The fix multiplies by the `(2, 3)` LAF directly, and moves the basis to the LAF's device and dtype
before broadcasting it rather than after (which also delivers the memory finding).

## Measurements

Quiet back-to-back A/B, base `8db1499a` in a worktree vs this branch, through the public API.
Apple M1, macOS 26.5, torch 2.14.0, Python 3.11, float32, 4 threads, min of 7×N iterations.
Throughput in LAFs/s.

| device | op | B=1 N=1000 | B=1 N=20000 | B=8 N=2000 |
| --- | --- | --: | --: | --: |
| CPU | `laf_to_boundary_points` base | 1026736 | 1069069 | 1073034 |
| CPU | `laf_to_boundary_points` **branch** | **19053970** | **29593824** | **29386781** |
| CPU | `laf_is_inside_image` base | 4811393 | 9754856 | 9282596 |
| CPU | `laf_is_inside_image` **branch** | **8542628** | **28980118** | **28227283** |
| MPS | `laf_to_boundary_points` base | 560977 | 1766110 | 1612615 |
| MPS | `laf_to_boundary_points` **branch** | **956994** | **4269400** | **3880007** |
| MPS | `laf_is_inside_image` base | 477173 | 2695192 | 2153567 |
| MPS | `laf_is_inside_image` **branch** | **905380** | **8534784** | **7043646** |

That is 27.7× / 3.0× on CPU and 2.4× / 3.2× on MPS at N=20000. No new benchmark script: both ops
are already in `benchmarks/feature/laf_ops.py`, so the harness reproduces this directly.

## Numerics — not byte-identical

The two gemm shapes round differently, so the result moves at the last ULP. Over a 126-case sweep
(2 devices × 4 dtypes × 3 shapes × 3 `n_pts` × 2 coordinate scales), worst deviation as a fraction
of the largest coordinate in each case:

| | float32 | float64 | float16 | bfloat16 |
| CPU | 8.4e-08 (0.71 eps) | 3.1e-16 (1.42 eps) | **bitwise identical** | **bitwise identical** |
| MPS | 1.0e-07 (0.85 eps) | — | 7.0e-04 (0.72 eps) | 5.6e-03 (0.72 eps) |

Every case is within 1.42 machine epsilon of the dtype. On CPU float32, 9.5% of entries differ, at
1–2 ULP for 82% of those; the tail to ~11 ULP is on entries near zero, where cancellation inflates
the ULP metric while the absolute error stays ~1e-7. Mean absolute error against a float64
reference is slightly **lower** than before (4.30e-08 vs 4.50e-08 at unit scale). This is the same
class of change as #4122's closed-form `ellipse_to_laf`, and the maintainer approved it on that
basis.

`laf_is_inside_image`'s boolean masks are **bitwise identical** in all 126 cases, so the only
in-library consumer is unaffected. Degenerate shapes (`B=0`, `N=0`) return the same empty tensors
as base.

## Tests

Three added to `TestLAF2pts`:

- `test_matches_explicit_affine_reference` — pins the equality the fix relies on, against an
  explicit `einsum` reference on random anisotropic LAFs. The existing `test_conversion` uses a
  **symmetric** 2×2 block, so it cannot see a transposed operand order: I confirmed this by
  transposing the block in the implementation, where the new test fails and `test_conversion`,
  `test_shape`, `test_gradcheck` and `test_jit` all pass.
- `test_dtype_device_preserved` — pins that the basis is cast before it is broadcast, across all
  four dtypes (nothing else covered this op in half precision).
- `test_dynamo` — the op had no `torch_optimizer` coverage.

## Also in this diff

`benchmarks/README.md`'s two `laf_to_boundary_points` bullets asserted the transfer diagnosis and
prescribed a fix that measures 1.00×; they now record the real cause, the A/B, and the method
lesson. The committed x86/CUDA result files predate this change and are flagged as such — I cannot
re-run them. The op table in `benchmarks/feature/laf_ops.py` still described both patch extractors
by the batch loops that #4128 removed; corrected in the same table.

## What did you check?

- `tests/feature/test_laf.py` + `tests/feature/test_scale_space_detector.py`, CPU, float32 /
  float64 / float16 / bfloat16
- whole `tests/feature`, CPU float32
- both files on MPS float32
- `KORNIA_TEST_OPTIMIZER=inductor` for the dynamo tests; `torch.jit.script` via the existing
  `test_jit`
- `pixi run pre-commit-all`
- base-vs-branch A/B and the 126-case numerical sweep above, run as stdin from each worktree root
  with `kornia.__file__` printed and checked

Not verified: CUDA (no device available), and the M=3 gemm cliff on x86 or on torch versions other
than 2.14 — the code comment and the README are scoped to what was measured.

## How did AI help?

Claude (Opus 5) measured the op, falsified the previously published diagnosis, wrote the change and
the tests, and ran the A/B and numerical sweeps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -389,6 +389,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and when that product is subnormal the division loses precision but never corrupts a representable result to a
   zero, `inf`, or `nan`.
 
+* Drop the constant homogeneous row from `laf_to_boundary_points`. The function appended a `[0, 0, 1]` row to every
+  LAF so that it could divide the result by a homogeneous coordinate that is always exactly `1`; it now multiplies
+  by the `(2, 3)` LAF directly, and moves its 3-column basis to the LAF's device and dtype before broadcasting it
+  instead of after (the previous order materialized, and on an accelerator transferred, one copy per LAF -- ~12 MiB
+  per call at `N = 20000` for a tensor with `n_pts` distinct rows). The win is not the transfer: torch's CPU
+  batched gemm falls off a fast path at the third row, so `bmm` on a `(B, 3, 3)` operand costs 23x a `(B, 2, 3)` one
+  for 1.5x the arithmetic. Quiet back-to-back A/B against base `8db1499a` via `benchmarks/feature/laf_ops.py`
+  (Apple M1, torch 2.14.0, 4 threads, B=1 N=20000): `laf_to_boundary_points` 1.07M -> 29.6M LAFs/s eager on CPU and
+  1.77M -> 4.27M on MPS; `laf_is_inside_image`, which calls it with `n_pts=12` on every detector forward,
+  9.75M -> 29.0M on CPU and 2.70M -> 8.53M on MPS. Results are bitwise identical on CPU `float16` and `bfloat16`
+  and shift by at most 1.42 machine epsilon elsewhere, relative to the largest coordinate of the call -- the two
+  gemm shapes round differently. Mean error against a `float64` reference is slightly lower than before, and
+  `laf_is_inside_image`'s boolean masks were bitwise identical across a 126-case device/dtype/shape sweep. (#4217)
+
 * Make `load_pointcloud_ply` and `load_pointcloud_ply_binary` parse the PLY header instead of skipping a fixed
   number of lines. Both loaders skipped `header_size=8` lines and read everything after them as `x y z` triples, so a
   minimal PLY file (seven header lines, no `comment`) lost its first point to a header line -- the binary reader
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -268,12 +268,11 @@ The honest reading:
   start by bisecting stack against machine rather than reading the kernel. It is not the
   N-chunking added in #4128: at B=1 N=2000 the sampling grid is ~16 MiB against a 64 MiB budget,
   so that config takes the single-chunk fast path and still loses.
-- **`laf_to_boundary_points` is the weakest compiled op on CUDA** — 17.8M LAFs/s, below both
+- **`laf_to_boundary_points` was the weakest compiled op on CUDA** — 17.8M LAFs/s, below both
   patch extractors, ~25× below `ellipse_to_laf`, and the least moved by `torch.compile` there
-  (1.26×, against up to 9.1× elsewhere). It builds its `linspace`/`ones`/`tensor` basis without a
-  `device=` argument and `.to()`s the result on every call (`kornia/feature/laf.py`), so each
-  call pays a host→device transfer. On CPU it is ~9× slower than the similar-sized
-  `make_upright`. First optimization target in this file.
+  (1.26×, against up to 9.1× elsewhere). On CPU it was ~9× slower than the similar-sized
+  `make_upright`. #4217 removed the cause; see the last bullet for what it actually was. The
+  tables above predate that change, as do their `laf_is_inside_image` rows, which share it.
 - **Small N on CUDA is launch-latency bound, not work bound:** at B=1 N=2000 the three cheapest
   ops all land within 8.2–14.2M LAFs/s regardless of what they compute, ~8.5–11× below their own
   B=1 N=20000 figures. Read the N=20000 rows for kernel cost and the N=2000 rows for per-call
@@ -293,12 +292,22 @@ The honest reading:
   `torch.compile` (4.4× and 3.8×). MPS only wins where there is real work per LAF: patch
   extraction, by 2.2–2.4× eager and up to 3.4× compiled. A pipeline that moves LAFs to the GPU
   for the frame math alone pays for the transfer twice.
-- **`laf_to_boundary_points` is the worst op on every machine measured**, and its CPU numbers say
-  why: on the M1 it is 30× below `make_upright` and `torch.compile` recovers only 1.29×, because
-  the cost is not a kernel. It builds its 50-point basis with no `device=`, `.expand()`s it to
-  `(B*N, n_pts, 3)` and only then calls `.to(device)`, so every call materializes — and on a GPU
-  transfers — a tensor that scales with the LAF count: ~12 MiB per call at N=20000 for a basis
-  with 50 distinct rows. Build it on the target device and broadcast in the matmul instead.
+- **`laf_to_boundary_points` was the worst op on every machine measured, and the obvious
+  diagnosis was the wrong one.** It built its 50-point basis with no `device=`, `.expand()`ed it
+  to `(B*N, n_pts, 3)` and only then called `.to(device)`, so every call materialized — and on a
+  GPU transferred — a tensor scaling with the LAF count: ~12 MiB per call at N=20000 for a basis
+  with 50 distinct rows. That is real, and it is a memory problem, not the speed problem: fixing
+  only it measures **1.00× on CPU**. The cost was one gemm-shape cliff. The op appended a constant
+  `[0, 0, 1]` row to every LAF so that it could divide the result by a homogeneous coordinate that
+  is always exactly 1, and torch's CPU batched gemm falls off a fast path at that third row — at
+  B=1 N=20000, `bmm` on a `(B, 3, 3)` operand takes 14.9 ms where `(B, 2, 3)` takes 0.65 ms, 23×
+  for 1.5× the arithmetic, identically at 1, 4 and 8 threads (M=1 0.33 ms, M=2 0.65, M=3 14.87,
+  M=4 16.26). #4217 multiplies by the `(2, 3)` LAF directly. Quiet back-to-back A/B on the M1
+  (torch 2.14.0, 4 threads, B=1 N=20000): `laf_to_boundary_points` 1.07M → 29.6M LAFs/s eager on
+  CPU and 1.77M → 4.27M on MPS; `laf_is_inside_image`, which calls it with `n_pts=12` on every
+  detector forward, 9.75M → 29.0M on CPU and 2.70M → 8.53M on MPS. **The lesson is the method:**
+  the transfer was visible by reading the source and the cliff was only visible by timing the
+  components, so the readable diagnosis got written up first and would have shipped a 1.00× fix.
 
 ## Sample results — augmentation flagship (class API)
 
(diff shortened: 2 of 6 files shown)


ducha-aiki:
## Review — approve, with two test-hygiene asks

Reviewed `047ef28c` against `8db1499a`. I re-derived every load-bearing claim on my own box
(Apple M1, macOS 26.5, torch 2.14.0, Python 3.11, `.venv`, 4 threads unless stated) rather than
reading the PR body, and re-ran the base implementation verbatim in-process instead of trusting a
worktree import. Nothing blocking.

### What I reproduced

- **The M=3 gemm cliff is real and is the whole story.** `bmm((B, M, 3), (B, 3, 50))` at B=20000:

  | threads | M=1 | M=2 | M=3 | M=4 |
  | --- | --: | --: | --: | --: |
  | 1 | 0.96 ms | 1.89 ms | **17.94 ms** | 14.55 ms |
  | 4 | 0.40 ms | 0.73 ms | **17.90 ms** | 14.37 ms |
  | 8 | 0.37 ms | 0.64 ms | **17.91 ms** | 14.53 ms |

  24x from M=2 to M=3, thread-count independent — same shape as the table in the description.
- **Equivalence.** 126-case sweep (2 devices x 4 dtypes x 3 shapes x 3 `n_pts` x 2 coordinate
  scales) against the pre-#4217 body run in the same process: worst relative deviation 1.13e-07 on
  CPU `float32` (0.95 eps), 1.03e-07 / 0.82 eps / 0.55 eps on MPS `float32` / `float16` /
  `bfloat16`; bitwise identical on CPU `float64`, `float16` and `bfloat16` in my draw.
  `laf_is_inside_image`'s boolean masks: **0 mismatches in 126/126 cases**.
- **Degenerate shapes** (`B=0`, `N=0`, both, and `n_pts=1`) return the same shapes and values as base.
- `gradcheck` passes; `torch.jit.script` output is bitwise equal to eager.
- **The new pin discriminates.** Transposing the 2x2 block in the implementation (via a
  non-aliasing `cat`, not an overlapping in-place write — that one also breaks `gradcheck` and
  muddies the signal) fails **only** `test_matches_explicit_affine_reference`; `test_conversion`,
  `test_shape`, `test_dtype_device_preserved`, `test_dynamo`, `test_gradcheck` and `test_jit` all
  stay green.
- **`bmm` does not materialize the expanded basis.** On MPS the allocator delta across the call is
  8126464 B for an 8000000 B output, i.e. exactly the result — the stride-0 batch dim is consumed
  in place. The memory claim in the new code comment holds.
- `tests/feature/test_laf.py` + `tests/feature/test_scale_space_detector.py`, CPU: 204 passed, 9
  skipped. `TestLAF2pts` green in all four dtypes.
- The new `test_dynamo` is actually collected in CI: the dynamo leg's path list
  (`.github/workflows/pr_test_cpu.yml:169`) includes `tests/feature`. Worth stating explicitly,
  because `tests/models` is still absent from that list, so the same test added there would run
  nowhere.

### Asks

**1. `test_dtype_device_preserved`'s docstring names a property it cannot observe.** It says "The
basis is cast to the LAF's dtype and device before it is broadcast, not after." I restored exactly
that ordering in the implementation — `pts.t().unsqueeze(0).expand(B * N, 3, n_pts).to(LAF.device).to(LAF.dtype)`,
keeping the (2, 3) `bmm` — and **all 13 `TestLAF2pts` tests pass in float32/float64/float16/bfloat16.**
Both orderings produce the same dtype, the same device and the same values; the ordering is a
memory property, and no assertion in the test can see it. This is the same failure mode this PR
correctly diagnoses on `test_conversion` one test above. What the test *does* pin is real and worth
keeping — a value check in all four dtypes, which this op had nowhere else — so please just reword
the docstring to that, e.g. "Boundary points keep the LAF's dtype and device, in every dtype
(nothing else covers this op in half precision)."

**2. Take Copilot's device nit.** `assert pts.device.type == laf.device.type` is satisfied by any
implementation that does not deliberately move devices. The `device` fixture is indexed
(`torch.device("cuda:0")` / `("mps:0")`, `conftest.py:71,79`), so `assert pts.device == laf.device`
is safe here, and it is the idiom this file already uses at
`tests/feature/test_laf.py:857` (`assert patches.device == img.device`).

### Optional

**Stale rationale in `kornia/feature/scale_space_detector.py:465-467`.** The inlined boundary check
is justified by "Avoids: ... `laf_to_boundary_points` (linspace/sin/cos allocations + CPU→GPU
transfer + bmm)". The transfer it points at is the ~12 MiB one this PR just deleted (what remains is
144 B) and the `bmm` is now the cheap shape. Still literally true, but it now cites the diagnosis
this PR falsifies, and AGENTS.md puts prose in the blast radius. One line: say the remaining saving
is the allocations and kernel launches, not the transfer.

### Two alternatives I measured that do *not* beat the chosen form

Recording these so nobody re-opens them later:

- **`torch.matmul` with a 2-D `(3, n_pts)` rhs** (no `expand`, folds to one gemm instead of a
  batched one) is 1.5-1.8x *faster* at N<=2000 and 0.75-0.86x *slower* at N>=8000, crossing over
  around N≈4000-8000 (n_pts=12: N=1000 1.70x, N=2000 1.50x, N=8000 1.03x, N=20000 0.75x). Since
  `laf_is_inside_image` runs at detector N, this is a genuine trade-off rather than a free win, and
  optimizing the N=20000 benchmark config is the defensible call. Not asking for a change.
- **Building the basis with `device=`/`dtype=` directly** — the prescription the old README text
  gave — is 0.62-1.07x, i.e. mostly *slower*: it trades one tiny H2D copy for 3-5 extra kernel
  launches (MPS B=1 N=1000 n_pts=12: 0.53 ms → 0.86 ms). Good that this PR dropped that sentence
  instead of implementing it; the correction in `benchmarks/README.md` is a second wrong diagnosis
  retired, not just the first.

CHANGELOG placement under `### Bug fixes` matches the neighbouring #4122 `ellipse_to_laf` perf entry,
and the "tables above predate that change" flag in `benchmarks/README.md` covers both the CPU and
the MPS tables. CI is still in flight at the time of writing (the green legs so far include
ubuntu/windows float32+float64 on 2.5.1 and 2.14.0, typecheck, pre-commit, docs and the dropped-name
check).

ducha-aiki:
### Review asks applied — `73481f21`

Both asks from the review above, plus the optional one:

- **`test_dtype_device_preserved` docstring** now says what the test pins (the op's values, dtype
  and device in all four dtypes — coverage it had nowhere else) and states explicitly that it
  *cannot* see where the basis is cast. Re-verified after the reword: restoring
  `pts.t().unsqueeze(0).expand(B * N, 3, n_pts).to(LAF.device).to(LAF.dtype)` still leaves all 13
  `TestLAF2pts` tests green, which is the point.
- **Device assertion** tightened to `assert pts.device == laf.device`.
- **`scale_space_detector.py`** — both comments that justified the inlined boundary check by the
  host→device transfer this PR deletes now say the remaining saving is allocations and kernel
  launches, not bandwidth.

Checks on the new commit: `tests/feature/test_laf.py` 314 passed / 52 skipped across
float32+float64+float16+bfloat16 on CPU; `KORNIA_TEST_OPTIMIZER=inductor -k "dynamo or compile"`
8 passed; pre-commit clean on both files. Full CI was 43/43 green on `047ef28c`, and this commit
touches one assertion, one docstring and three comment lines.

### One pre-existing failure, out of scope for this PR

`TestScaleSpaceDetector::test_batched_underfill_does_not_leak_the_topk_sentinel` fails on CPU
`bfloat16` on torch **2.14.0** — the batched `lafs1[0]` disagrees with the unbatched `lafs[0]`.
It is **not** from this change: with `kornia/feature/laf.py` checked out at base `8db1499a` and
everything else at this branch, it fails identically, 3 runs out of 3 on each side. It is also not
in `testing/known_failure_xfails/cpu-bfloat16.txt`, and the PR's `half-precision (bfloat16)` leg is
green — that leg runs torch 2.9.1, so this looks like a torch-2.14-only regression on `main` rather
than anything new. Filing separately rather than widening this PR.

ducha-aiki:
Correction to the note above: that `bfloat16` failure is not a torch-2.14 *regression* in the
detector, it is a non-tie-safe test. The batched and unbatched responses are bitwise identical; two
LAF rows are simply **swapped**, because the two candidates share the same `bfloat16` response
(`0.00225830078125` — the 500 responses take only 16 distinct values) and `topk` breaks that tie
differently at `B = 2` than at `B = 1`. `float16` and `float32` pass on the same seed. The fix is
in the test, not in `ScaleSpaceDetector`. Filed as #4219; still out of scope here, and still
verified unrelated to this PR.

PR #4218, opened 2026-09-04
Fix rank-4 bbox input validation

### LINKED ISSUE #4180: W1b: `infer_bbox_shape` / `bbox_to_mask` accept `(B, N, 4, 2)` and silently return wrong-shaped results when `N >= 3`
## Summary

`infer_bbox_shape` (`kornia/geometry/bbox.py:112` at `e0969438`) and `bbox_to_mask` (`bbox.py:194`) index vertices at fixed positions on axis 1 (`boxes[:, 1, 0]`, `boxes[:, 2, 1]`, ...) and never check the rank. Both are documented for `(N, 4, 2)` input, but `Boxes.to_tensor('vertices_plus')` on a batched object returns `(B, N, 4, 2)`, and passing that straight through does one of two things depending on `N`:

- `N < 3`: `IndexError` (index 1 or 2 out of bounds on the box axis);
- `N >= 3`: **succeeds silently**, reading box 1 and box 2 instead of vertices 1 and 2 and returning values of the wrong shape.

`validate_bbox` in the same module flattens rank-4 input and accepts it, so a caller who validated first has no reason to expect the consumer to disagree.

## Reproduction

```python
import torch
from kornia.geometry.bbox import infer_bbox_shape, bbox_to_mask

for n in (1, 2, 3):
    x = torch.arange(n * 8, dtype=torch.float32).reshape(1, n, 4, 2)
    try:
        h, w = infer_bbox_shape(x)
        print(n, "infer_bbox_shape ->", tuple(h.shape), h.tolist(), w.tolist())
    except IndexError as e:
        print(n, "infer_bbox_shape ->", e)
    try:
        print(n, "bbox_to_mask ->", tuple(bbox_to_mask(torch.zeros(1, n, 4, 2), 5, 5).shape))
    except IndexError:
        print(n, "bbox_to_mask -> IndexError")

b = Boxes.from_tensor(torch.tensor([[[1., 2., 5., 4.], [0., 0., 2., 2.], [3., 3., 4., 4.]]]), "xyxy")   # (1, 3, 4)
print([t.tolist() for t in infer_bbox_shape(b.to_tensor("vertices_plus"))], "vs", [t.tolist() for t in b.get_boxes_shape()])
```

## Observed

```
1 infer_bbox_shape -> index 1 is out of bounds for dimension 1 with size 1
1 bbox_to_mask -> IndexError
2 infer_bbox_shape -> index 2 is out of bounds for dimension 1 with size 2
2 bbox_to_mask -> IndexError
3 infer_bbox_shape -> (1, 2) [[17.0, 17.0]] [[9.0, 9.0]]
3 bbox_to_mask -> (2, 5, 5)
[[[0.0, 2.0]], [[0.0, -1.0]]] vs [[[2.0, 2.0, 1.0]], [[4.0, 2.0, 1.0]]]
```

The last line is the realistic path: a batched `Boxes` with three boxes, exported and handed to `infer_bbox_shape`, yields a `(1, 2)` "heights" tensor of `[0, 2]` and "widths" of `[0, -1]` where `get_boxes_shape` says `(2, 2, 1)` and `(4, 2, 1)`. Nothing raises, and the shapes are plausible enough to flow into downstream code.

## Expected

Either accept `(B, N, 4, 2)` by flattening the leading axes the way `validate_bbox` does and returning `(B, N)` extents / `(B, N, H, W)` masks, or raise a `ShapeError`-style check on `boxes.ndim != 3` so the mistake is caught at the call site. Silent success with wrong shapes for `N >= 3` should not remain.

## Context

Found while reviewing #4060 (bbox/boxes convention blocks, batch 4a), which now states that rank-4 input is not supported by these two functions and tells `Boxes` users to index or flatten a batched export first. That PR is documentation only, so the shape check itself is out of scope there, and this issue is the anchor for that sentence. Whichever repair is chosen for #3934 (inclusive `+1`) touches the same two functions, so the check can ride along. Part of the #4020 audit.

## Environment

hariomCodes:

`infer_bbox_shape` and `bbox_to_mask` document rank-3 `(N, 4, 2)` inputs, but rank-4 `(B, N, 4, 2)` tensors previously produced either an incidental `IndexError` or silently wrong-shaped results. This adds the standard `KORNIA_CHECK_SHAPE` guard to both functions so invalid rank-4 inputs consistently raise `ShapeError`.

It also replaces the existing wart test with regression coverage for both former failure modes, adds equivalent coverage for `bbox_to_mask`, and updates the public documentation and changelog.

Fixes #4180

## What did you check?

```text
python -m pytest tests/geometry/test_bbox.py -q
28 passed

python -m pytest --doctest-modules kornia/geometry/bbox.py -q
7 passed

python -m pytest tests/geometry/test_bbox.py -k "rank4_4180" -q
4 passed
```

The valid-input path also passed a direct `torch.compile(..., fullgraph=True)` check for `bbox_to_mask` and a TorchScript check for `infer_bbox_shape`. Ruff, formatting, codespell, license-header, and the other configured pre-commit hooks passed on the changed files.

## How did AI help?

Codex helped trace the two failure modes, implement the shape guards and regression tests, update the related documentation, and run the focused verification. The final diff and test results were reviewed against Kornia's contribution guidelines before submission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -253,6 +253,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   instead of `view`, so a non-contiguous leading-dimension stride (a transpose, a slice that drops boxes, an
   `expand`) returns a boolean as documented instead of raising `RuntimeError`. (#4174)
 
+* `infer_bbox_shape` and `bbox_to_mask` now reject rank-4 `(B, N, 4, 2)` inputs with `ShapeError`. Previously,
+  `N < 3` raised an incidental `IndexError`, while `N >= 3` silently returned `(B, 2)` extents computed from the
+  wrong vertices. (#4218)
+
 * Constructing `ScaleSpaceDetector` with `compile_modules` no longer permanently mutates the process-global
   `torch._dynamo.config.capture_dynamic_output_shape_ops` setting. (#4134)
 
@@ -785,11 +789,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   whose exponent range holds both the guard and the squares, and the wider dtypes are byte-identical at every
   pixel whose gradient is not exactly zero; a patch with such pixels (a saturated region) loses their `sqrt(eps)`
   contribution, a change of at most `1e-5` per pixel before normalisation.
-
-### Fixed
-
-* `infer_bbox_shape` and `bbox_to_mask` now reject rank-4 `(B, N, 4, 2)` inputs with `ShapeError` instead of
-  raising an incidental `IndexError` or silently returning incorrectly shaped results. (#4218)
 
 ## :rocket: [0.6.11] - 2022-03-28
 ### :new:  New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -786,6 +786,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   pixel whose gradient is not exactly zero; a patch with such pixels (a saturated region) loses their `sqrt(eps)`
   contribution, a change of at most `1e-5` per pixel before normalisation.
 
+### Fixed
+
+* `infer_bbox_shape` and `bbox_to_mask` now reject rank-4 `(B, N, 4, 2)` inputs with `ShapeError` instead of
+  raising an incidental `IndexError` or silently returning incorrectly shaped results. (#4218)
+
 ## :rocket: [0.6.11] - 2022-03-28
 ### :new:  New Features
 
--- a/kornia/geometry/bbox.py
+++ b/kornia/geometry/bbox.py
@@ -21,6 +21,7 @@ from typing import Optional
 
 import torch
 
+from kornia.core.check import KORNIA_CHECK_SHAPE
 from kornia.core.utils import is_exporting
 
 from .linalg import transform_points
@@ -152,15 +153,13 @@ def infer_bbox_shape(boxes: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
         The inclusive ``+1`` arithmetic differs from torchvision, COCO, and albumentations and is tracked in
         `#3934 <https://github.com/kornia/kornia/issues/3934>`_.
 
-    .. warning::
-        Rank-4 input is not validated. The fixed indices then read boxes instead of vertices, so fewer than three
-        boxes per batch raise ``IndexError`` and three or more return wrongly shaped values. This wart is tracked
-        in `#4180 <https://github.com/kornia/kornia/issues/4180>`_.
-
     Args:
         boxes: a tensor containing the coordinates of the bounding boxes to be extracted. The tensor must have shape
             :math:`(N, 4, 2)`, where each box is defined in the following ``clockwise`` order: top-left, top-right,
             bottom-right, bottom-left. The coordinates must be in the x, y order.
+
+    Raises:
+        ShapeError: if ``boxes`` does not have shape :math:`(N, 4, 2)`.
 
     Returns:
         - Bounding box heights, shape of :math:`(N,)`.
@@ -182,6 +181,8 @@ def infer_bbox_shape(boxes: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
         (tensor([2., 2.]), tensor([2., 3.]))
 
     """
+    KORNIA_CHECK_SHAPE(boxes, ["N", "4", "2"])
+
     width: torch.Tensor = boxes[:, 1, 0] - boxes[:, 0, 0] + 1
     height: torch.Tensor = boxes[:, 2, 1] - boxes[:, 0, 1] + 1
     return height, width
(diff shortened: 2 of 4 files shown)


ducha-aiki:
## Changes requested — one thing, and it is mechanical

Reviewed at `9ddb2284` against merge-base `8db1499a`, torch 2.14.0 / py311 / macOS arm64. The code change is
right, complete, and better verified than the body claims. There is exactly one blocker, in `CHANGELOG.md`.

### The blocker: the entry is not in the section it needs to be in

`## Unreleased` spans lines **9–747** of `CHANGELOG.md`, and inside it the file uses three headings:

```
 11: ### Added
102: ### Breaking changes
239: ### Bug fixes
743: ### Fixed      <- yours
```

So the entry did land inside `Unreleased`, but under a **fourth heading that the file does not use**, ~500 lines
below `### Bug fixes` where every other bug-fix entry in this release lives (including the four other open PRs
touching this file today). AGENTS.md names the headings deliberately: *"under the Keep a Changelog headings the
file already uses (`Added`, `Bug fixes`, `Breaking changes`)"*. Please move the bullet up under `### Bug fixes`
at line 239 and drop the new `### Fixed` heading. It is easy to miss — the section is long enough that the
bottom of `Unreleased` looks like the bottom of the file.

While you are in there: the entry says the old behaviour was "an incidental `IndexError` or silently returning
incorrectly shaped results". Worth naming the split, since it is what makes the silent case dangerous — `N < 3`
raised, `N >= 3` succeeded with `(B, 2)` extents built from the wrong vertices.

- **Fail-on-base 4/4.** Your four pins on `8db1499a` with `main`'s `kornia/`:

  ```
  test_infer_bbox_shape_rejects_rank4_4180[cpu-float32-2]  IndexError: index 2 is out of bounds ...
  test_infer_bbox_shape_rejects_rank4_4180[cpu-float32-3]  DID NOT RAISE ShapeError
  test_bbox_to_mask_rejects_rank4_4180[cpu-float32-2]      IndexError: index 2 is out of bounds ...
  test_bbox_to_mask_rejects_rank4_4180[cpu-float32-3]      DID NOT RAISE ShapeError
  ```

  Parametrising over `num_boxes` in `{2, 3}` is the right call — those are the two *different* base behaviours
  #4180 describes, and a single-`N` test would have pinned only one of them.

- **Head:** `test_bbox.py` + `test_boxes.py` + `transform/test_crop2d.py`, cpu, float32+float64 -> **224 passed**.
  The three in-tree call sites (`crop2d.py:283`, `crop2d.py:508`, `cutmix.py:124/170`, `erasing.py:120/139`) are
  all rank-3, so nothing in the library trips the new guard: `tests/augmentation -k "Erasing or CutMix"` ->
  **371 passed, 9 skipped**.

- **The compile question, which is the one I would have expected this to break.** `bbox_to_mask` carries an
  explicit `NOTE:` saying `validate_bbox` was *removed* from it because its `torch.any(...)` -> Python `if` was a
  data-dependent graph break that blocked `fullgraph` for `RandomErasing`. `KORNIA_CHECK_SHAPE` is shape-only, so
  it should be safe — verified rather than assumed, with `dynamic=True` so a shape guard would have to show up:

  ```
  torch.compile(bbox_to_mask,     fullgraph=True, dynamic=True) -> (1, 5, 5)      OK
  torch.compile(infer_bbox_shape, fullgraph=True, dynamic=True) -> [(1,), (1,)]   OK
  torch.compile(RandomErasing(p=1.0), fullgraph=True)           -> (2, 3, 8, 8)   OK
  ```

- **Documented-defect gate.** `grep -rnE "#4180|issues/4180" kornia/ tests/` at head: **zero hits anywhere**. You
  removed the `boxes.py` Convention sentence, the `infer_bbox_shape` warning, and the wart pin, and replaced the
  first with prose that states the new contract. That is the complete version of that gate.

- `uvx ruff@0.16.4 check` / `format --diff` (the pinned version, not a local ruff) clean on all three files.

### Two notes, no action required

1. `:class:`~kornia.core.exceptions.ShapeError`` in `boxes.py` has no documentation target — `kornia.core.exceptions`
   appears in no `docs/source/*.rst` — so it renders as plain text rather than a link. `nitpicky` is not set in
   `conf.py`, so this is not a build failure and not something to fix here; the module's absence from the API
   reference is a pre-existing gap.
2. Rejection (over flattening) is one of the two dispositions #4180 explicitly offers, so this is a legitimate
   choice and I am not asking you to revisit it. But note that **#4215**, open against the same file, moves
   `validate_bbox` from `view` to `reshape` so it accepts *any* rank-4 layout — after both land, the validator
   accepts a rank its two consumers refuse, which is the exact trap #4180's text calls out. Your `boxes.py` prose
   documents it from the `Boxes` side; whichever of the two PRs merges second may want a clause in
   `validate_bbox`'s own `Convention:` block. Not a change to this PR.

Ping me when the CHANGELOG bullet moves and I will re-check and approve.

hariomCodes:
@ducha-aiki Thanks for the detailed review. Addressed in the rebased commit 67eb7a9d: I moved the entry under the existing Bug fixes heading and clarified the behavior split: N less than 3 raised an incidental IndexError, while N greater than or equal to 3 silently returned incorrect (B, 2) extents from the wrong vertices. The branch is now rebased onto current main and mergeable. Please re-check when convenient.

ducha-aiki (approved):
LGTM, thank you!

PR #4221, opened 2026-09-04
test(feature): make the batched-vs-single detector pin tie-safe

### LINKED ISSUE #4219: test_batched_underfill_does_not_leak_the_topk_sentinel is not tie-safe: fails on CPU bfloat16 with torch 2.14
### Summary

`TestScaleSpaceDetector::test_batched_underfill_does_not_leak_the_topk_sentinel` fails on CPU
`bfloat16` with torch 2.14.0. It is a **test** defect, not a detector defect: the test asserts that
the batched and the single-image top-K lists agree **positionally**, and that is not well defined
when responses tie — which in `bfloat16` they almost always do.

### Diagnosis

For `torch.manual_seed(0)`, `inp = torch.rand(2, 1, 64, 64)`, `ScaleSpaceDetector()` on CPU
`bfloat16`:

- responses are **bitwise identical** between the batched and unbatched runs (`max|diff| == 0`);
- exactly **two LAF rows differ, and they are swapped with each other**:

  ```
  row 3: response 0.00225830078125 (occurs 2x) -- centre batched [16.0, 14.0] vs unbatched [47.0, 46.0]
  row 4: response 0.00225830078125 (occurs 2x) -- centre batched [47.0, 46.0] vs unbatched [16.0, 14.0]
  ```

- the 500 responses take **16 distinct `bfloat16` values**, so ties are the norm, not the exception.

So both runs found the same set of keypoints with the same scores; `topk` merely broke a tie in a
different order for `B = 2` than for `B = 1`. `float16` and `float32` pass on the same input and
seed (`max|diff| == 0` on both), because they resolve the two responses apart.

### Scope

- Reproduces 3/3 on torch **2.14.0**, macOS/arm64, Python 3.11, CPU `bfloat16`.
- Not visible in PR CI: the `half-precision (bfloat16)` leg runs torch **2.9.1**, where the tie
  happens to break the same way for both batch sizes, and the leg is green.
- Not in `testing/known_failure_xfails/cpu-bfloat16.txt`.
- Pre-existing on `main`. Found while reviewing #4217; verified unrelated to it by checking out
  `kornia/feature/laf.py` at base `8db1499a` with everything else at that branch, where it fails
  identically 3/3.

### Suggested fix

The last two lines of the test are the problem:

```python
lafs1, resps1 = det(inp[:1])
self.assert_close(resps1[0], resps[0])
self.assert_close(lafs1[0], lafs[0])
```

The batched/unbatched agreement the test wants is a **set** property, not an ordering. Options, in
order of preference:

1. Compare order-insensitively — sort both by `(response, centre_x, centre_y)`, or match rows by
   centre before comparing.
2. Keep the positional comparison but choose an input whose responses do not tie in `bfloat16`, and
   say in a comment that this is load-bearing.
3. If neither, drop the batched-vs-unbatched half of the test into its own case so the sentinel
   assertions — which are the ones the test is named for and which pass — are not taken down with it.

Note the existing comment above those lines already anticipates a weaker version of this ("the
response convolutions pick batch-size-dependent kernels on some CPUs ... last-ULP-different
scores"); the actual failure mode is stronger than a tolerance issue, since the scores here are
identical and the *ordering* is what moves.

### Repro

```python
import torch
from kornia.feature import ScaleSpaceDetector

torch.manual_seed(0)
dt = torch.bfloat16
inp = torch.rand(2, 1, 64, 64, dtype=dt)
det = ScaleSpaceDetector().to(torch.device("cpu"), dt)
lafs, resps = det(inp)
lafs1, resps1 = det(inp[:1])
print("resp max|diff|:", (resps1[0].float() - resps[0].float()).abs().max().item())
print("laf  max|diff|:", (lafs1[0].float() - lafs[0].float()).abs().max().item())
```

```
resp max|diff|: 0.0
laf  max|diff|: 32.0
```

ducha-aiki:
Closes #4219.

## The defect

`test_batched_underfill_does_not_leak_the_topk_sentinel` ends by checking that the
single-image path agrees with the batched one, row by row:

```python
lafs1, resps1 = det(inp[:1])
self.assert_close(resps1[0], resps[0])
self.assert_close(lafs1[0], lafs[0])
```

That ordering is not defined. The two calls run `topk` over differently shaped
volumes, so equal responses come back in a different order — and in half precision
equal responses are the rule, not the exception: bfloat16 quantises this input's 500
candidate responses to about **16 distinct values**.

On torch 2.14 the test fails on CPU bfloat16 for precisely that reason. The two runs
agree on everything that matters:

- responses are **bitwise identical** between the calls (`max|diff| == 0`);
- exactly two LAF rows differ, and they are **swapped with each other**, both carrying
  response `0.00225830078125`.

So the detector found the same keypoints with the same scores and `topk` broke a tie
differently at `B = 2` than at `B = 1`. `float16` and `float32` pass on the same seed
because they resolve those two responses apart.

This is latent rather than new: the blocking half legs pin torch **2.9.1**, where the
tie happens to break the same way for both batch sizes.

## The fix

Sort both results into a canonical order before comparing, so the assertion is about
the *set* of detections rather than their positions. The sentinel assertions the test
is named for — which pass, and which are the reason the test exists — are untouched.

**The sort is on the centre first, deliberately, not on the response.** A
response-first sort would be no better than the positional comparison it replaces:
in bfloat16 the responses tie constantly, so the tiebreaker would decide almost every
row, and the existing comment in this test already warns that ubuntu CI can produce
last-ULP-different scores. Sorting on the centre means the remaining keys only ever
separate rows that share a centre, and those agree to a ULP anyway — so no sort
position depends on a last-bit difference.

## Verified

Passes at `float32`, `float64`, `float16` and `bfloat16`, on torch **2.14.0** and on
**2.9.1** (the half CI pin), macOS/arm64, Python 3.11.

The pin still discriminates. Substituting a mutated single-image result for the real
one fails the assertion:

| injected divergence | float32 | float16 | bfloat16 |
| --- | --- | --- | --- |
| unmutated | passes | passes | passes |
| one detection dropped | fails | fails | fails |
| a centre moved 3 px | fails | fails | fails |
| a sentinel leaked into a response | fails | fails | fails |
| all LAF scales x1.05 | fails | fails | fails |
| one response x1.05 | fails | passes | passes |

The last row is a pre-existing property of the half tolerances, not something this
change introduces: the responses here are ~0.0022, and 5% of that is far inside the
repository's `atol` for those dtypes. It was equally invisible to the positional
comparison.

Whole file: 125 passed at `--dtype=float32,float64`; 65 passed / 1 skipped at each
half dtype under `--xfail-known-failures` with the matching profile. No manifest line
changes — this node was never recorded, since it is green at the 2.9.1 pin CI uses.
`pre-commit run --all-files` clean.

Test-only change; no CHANGELOG entry.

### DIFF
--- a/tests/feature/test_scale_space_detector.py
+++ b/tests/feature/test_scale_space_detector.py
@@ -55,6 +55,26 @@ def _require_affine_orientation_kernels(device: torch.device, dtype: torch.dtype
             pytest.skip(f"no {name} kernel for {dtype} on {device.type}")
 
 
+def _canonical_order(lafs: torch.Tensor, responses: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+    """Sort one image's detections into a batch-size-independent order.
+
+    `topk` breaks equal responses by position, and the positions it ranks over differ between a
+    batched and a single-image call, so comparing two calls row by row compares an ordering that is
+    not defined. Half precision makes those ties the norm rather than the exception: bfloat16
+    quantises this module's candidate responses to a handful of distinct values.
+
+    Sorting on the centre first turns the comparison into a set comparison. The remaining keys only
+    decide between rows that share a centre, and those agree to a ULP anyway, so the order does not
+    depend on last-bit differences the way a response-first sort would.
+    """
+    rows = [
+        (float(lafs[i, 0, 2]), float(lafs[i, 1, 2]), *(float(v) for v in lafs[i].flatten()), float(responses[i]))
+        for i in range(lafs.shape[0])
+    ]
+    order = torch.tensor(sorted(range(len(rows)), key=rows.__getitem__), device=lafs.device)
+    return lafs[order], responses[order]
+
+
 class TestScaleSpaceDetector(BaseTester):
     def test_shape(self, device, dtype):
         inp = torch.rand(1, 1, 32, 32, device=device, dtype=dtype)
@@ -374,12 +394,16 @@ class TestScaleSpaceDetector(BaseTester):
         assert 0 < int(filled[0].sum()) < det.num_features, "expected a partially filled result"
         assert bool((resps[~filled] == 0).all()), f"sentinel leaked: min response {resps.min().item()}"
         assert (resps > torch.finfo(dtype).min / 4).all()
-        # and the single-image path agrees, which it did not before. Up to tolerance, not bit-for-bit:
-        # the response convolutions pick batch-size-dependent kernels on some CPUs (ubuntu CI), so the
-        # same candidates carry last-ULP-different scores.
+        # and the single-image path agrees, which it did not before. As a set, not row by row: the
+        # two calls rank over differently shaped volumes, so `topk` orders equal responses
+        # differently, and in bfloat16 this input's responses tie constantly (#4219). Up to
+        # tolerance, not bit-for-bit either: the response convolutions pick batch-size-dependent
+        # kernels on some CPUs (ubuntu CI), so the same candidates carry last-ULP-different scores.
         lafs1, resps1 = det(inp[:1])
-        self.assert_close(resps1[0], resps[0])
-        self.assert_close(lafs1[0], lafs[0])
+        batched_lafs, batched_resps = _canonical_order(lafs[0], resps[0])
+        single_lafs, single_resps = _canonical_order(lafs1[0], resps1[0])
+        self.assert_close(single_resps, batched_resps)
+        self.assert_close(single_lafs, batched_lafs)
 
     def test_padding_survives_the_affine_and_orientation_modules(self, device, dtype):
         # Same contract as `MultiResolutionDetector.forward`: the shape and orientation modules
--- a/tests/feature/test_scale_space_detector.py
+++ b/tests/feature/test_scale_space_detector.py
@@ -67,10 +67,12 @@ def _canonical_order(lafs: torch.Tensor, responses: torch.Tensor) -> tuple[torch
     decide between rows that share a centre, and those agree to a ULP anyway, so the order does not
     depend on last-bit differences the way a response-first sort would.
     """
-    rows = [
-        (float(lafs[i, 0, 2]), float(lafs[i, 1, 2]), *(float(v) for v in lafs[i].flatten()), float(responses[i]))
-        for i in range(lafs.shape[0])
-    ]
+    # One transfer each, not one per scalar: reading the keys straight off the device tensors costs a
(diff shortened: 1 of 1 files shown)


ducha-aiki (commented):
No actionable findings. The center-first canonicalization makes the batched-versus-single comparison tie-safe while preserving response/LAF pairing and the original sentinel assertions. The torch 2.14 bfloat16 positional mismatch was independently reproduced as two swapped LAF rows with identical responses. Focused runs passed across float16, bfloat16, float32, and float64; the full file passed for float32 and float64; Ruff and git diff --check were clean.

ducha-aiki (commented):
Correction to my earlier review: [P2] the Copilot comment is valid. _canonical_order converts seven device scalars for each of the default 500 detections, causing roughly 3,500 separate device synchronizations. On local MPS, the current helper took 1.5140 seconds versus 0.0128 seconds after materializing lafs and responses on CPU once, about 118x faster. Please copy both tensors to CPU once before constructing the Python tuples, then create the final index tensor on the original device. My earlier no-actionable-findings conclusion is withdrawn.

ducha-aiki:
Confirmed and fixed at `6eabaf5c`.

I re-measured rather than taking the number: `_canonical_order` built its keys with `float(tensor)`
per element, which is 9 scalar reads per row and **4500 separate device synchronizations** for the
default `num_features=500`. The helper is called twice per test invocation.

| | before | after |
| --- | --- | --- |
| MPS | 1215.32 ms | 1.81 ms (**671x**) |
| CPU | 4.95 ms | 0.25 ms (20x) |

Larger than the 118x in the review, but the same defect. Both tensors now come off the device once
with `.detach().cpu().tolist()`, the Python tuples are built from the lists, and the index tensor is
still created on the original device. `torch.equal` on both the returned LAFs and the responses
confirms the ordering is unchanged.

`test_batched_underfill_does_not_leak_the_topk_sentinel` passes on CPU across float16, bfloat16,
float32 and float64, and on MPS float32 (with `PYTORCH_ENABLE_MPS_FALLBACK` unset), where the node
now takes 0.50 s.

ducha-aiki (commented):
Re-review of latest head 6eabaf5c: no actionable findings. The prior device-synchronization finding is fully resolved: each tensor is transferred to CPU once, all Python keys are built from host lists, and the ordering index remains on the original device, preserving dtype, device, values, and LAF/response pairing. Focused validation passed on CPU float16, bfloat16, float32, and float64 and on MPS float32; lint and git diff --check were clean.

PR #4222, opened 2026-09-04
fix(feature): return the true top-k from MultiResolutionDetector.detect

### LINKED ISSUE #4170: MultiResolutionDetector.detect does not return the top num_features responses: the per-level quota caps what each level may contribute
### Describe the bug

`MultiResolutionDetector.detect` does not return the top `num_features` responses. Each pyramid level is asked for a number of candidates derived from that level's own share of the budget, so a level can contribute at most

```
up-level        : quotas[0]
pyramid level i : sum(quotas[: i + 1 + num_upscale_levels])
```

Both are well below `num_features` for the finer levels. When an image's strongest maxima concentrate on one level -- which is the normal case, not a corner case -- the excess is unreachable: those detections are never requested, so the final global `topk` cannot rank them in. The returned set is silently not the top-k, and the caller has no signal that anything was dropped.

This is independent of the quota *rounding*. #4101 / #4156 fixed the rounding (the floors used to lose part of the budget); this issue is the cap itself, and it reproduces identically before and after that change.

### Reproduction steps

Self-contained -- no pretrained weights, no external data:

```python
import torch
from kornia.feature import MultiResolutionDetector
from kornia.feature.responses import BlobHessian

img = torch.zeros(1, 1, 240, 240)
k = 0
for y in range(20, 220, 40):          # 25 well-separated blobs, all at one scale
    for x in range(20, 220, 40):
        img[:, :, y - 1 : y + 2, x - 1 : x + 2] = 1.0 + 0.01 * k
        k += 1

def run(nf):
    det = MultiResolutionDetector(BlobHessian(), num_features=nf)
    with torch.inference_mode():
        r, _ = det.detect(img)
    return [round(float(v), 6) for v in r[0] if float(v) > 0]

ref = sorted(run(64), reverse=True)
for nf in [4, 8, 16, 24]:
    got = run(nf)
    print(nf, len(set(got) & set(ref[:nf])), "/", nf)
```

```
4  3 / 4
8  6 / 8
16 12 / 16
24 18 / 24
```

All 25 blobs are detected on the same level, and the counts are exactly the per-level cap: with the default config (`pyramid_levels=4`, `up_levels=1`, `scale_factor_levels=sqrt(2)`) the quotas at `num_features=16` are `[8, 4, 2, 1, 1, 0]`, so level 0 is asked for `quotas[0] + quotas[1] == 12` candidates and the remaining 4 slots can only be filled from coarser levels. Same arithmetic gives 3/4, 6/8 and 18/24.

This is not synthetic-only. Measured on the Oxford-Affine images with `KeyNetDetector(pretrained=True)`, taking the true ranking from a `num_features=256` run:

| image | `num_features` | true top-k returned |
| --- | --- | --- |
| wall | 32 | 28/32 |
| wall | 16 | 13/16 |
| wall | 8 | 5/8 |
| ubc | 16 | 15/16 |
| ubc | 8 | 7/8 |
| boat | 16 | 14/16 |
| boat | 8 | 7/8 |

`wall` at `num_features=32` is the clearest: 20 of the true top-32 sit on the upscaled level, which is asked for `quotas[0] == 16`, so 4 are unreachable -- exactly the 28/32 observed.

### Expected behavior

`detect(num_features=k)` returns the `k` highest responses the detector can find, and in particular the result for `k` is a prefix of the result for any larger `k'`. Today it is not.

Whatever the fix, the per-level request has to stop being a hard cap on what a level may contribute -- for example by over-requesting per level and letting the existing global `topk` do the selection, at the cost of more candidates carried through the concatenation. That is a real memory/time trade-off, so it wants a deliberate decision rather than a drive-by change, which is why this is filed separately from #4156.

### Environment

Reproduced on `main` at `0ba3414c` and on the #4156 branch, identically. CPU, float32, torch 2.9.1, Python 3.11.

ducha-aiki:
Closes #4170.

## The defect

`detect` asked each pyramid level only for its own share of the budget, so a level
could contribute at most `quotas[0]` (the upscaled level) or
`sum(quotas[: i + 1 + up_levels])` (pyramid level *i*). Both are well under
`num_features` for the finer levels. When an image's strongest maxima concentrate on
one level — the normal case, not a corner case — the excess was never requested, so
the final global `topk` could not rank it in. The returned set was silently not the
top-k, with no signal to the caller.

## The fix

Ask every level for `num_features`.

This is **exactly sufficient rather than merely generous**: the *m* detections a level
contributes to the global top-k are necessarily that level's own top-*m* (if some other
maximum on that level scored higher it would also be in the global top-k), so a request
of `num_features` always reaches them, and the existing global `topk` then selects
correctly.

## What this costs — measured, because the issue flags it as the hard part

The issue notes this is "a real memory/time trade-off … wants a deliberate decision".
Measured, it is very close to free, because the coarse levels were **already** asked for
a *cumulative prefix* of the quotas — near `num_features` — so the old scheme was
carrying most of these candidates anyway:

| | candidates | wall clock, 480×640, `num_features=2048` |
| --- | --: | --: |
| before | 10433 | 521.29 ms |
| after | 12288 (**1.18×**) | 521.27 ms |

The ratio stays in **1.18–1.24×** across `pyramid_levels` 1–8 and `up_levels` 0–3. The
response function and the non-maxima suppression dominate so completely that the wall
clock difference is not measurable. LAF memory at `num_features=2048` grows by ~50 KB.

## Result

The issue's reproduction:

| `num_features` | before | after |
| --- | --- | --- |
| 4 | 3/4 | **4/4** |
| 8 | 6/8 | **8/8** |
| 16 | 12/16 | **16/16** |
| 24 | 18/24 | **24/24** |

and `detect(k)` is now a prefix of `detect(k')`, which it was not.

## Removing the apportionment

`num_features_per_level` was local to `detect` and read nowhere else, so the whole
apportionment block goes — **including the largest-remainder (Hamilton) rounding from
#4101**, which is recent. That is deliberate: with no per-level cap, a quota's only
remaining effect would be to suppress better detections.

Both tests #4101 added still pass, because asking every level for the full budget
spreads across scales strictly *better* than apportioning did. Their comments now state
the contract they pin rather than the removed mechanism, and say the pin holds under
flooring, under Hamilton, and under this change — so they are not left describing code
that no longer exists.

The scale distribution of the output barely moves: identical at `num_features=256` on a
random 240×320 image, and 8→3 fine-scale keypoints at `num_features=32`.

## The new test

It pins the **sorted response vector** as a prefix rather than keypoint identities. That
sidesteps both the `topk` tie-break and the fact that one blob can be detected at several
scales with the same rounded centre — the failure mode that produced #4219 earlier today.
It fails on `main` in all four dtypes, by 0.3 to 0.62 absolute, so it is not a marginal pin.

## Verified

- `tests/feature`: 683 passed, 94 skipped.
- `tests/feature/test_scale_space_detector.py`: 144 passed across
  `float32`, `float64`, `float16` and `bfloat16`.
- **MPS**: 62 passed, 5 skipped (`--device=mps`, with `PYTORCH_ENABLE_MPS_FALLBACK`
  unset). PR CI has no MPS leg covering these paths, and a past detector change hit an
  MPS-only blocker, so this is checked by hand.
- dynamo/compile: 2 passed under `KORNIA_TEST_OPTIMIZER=inductor`.
- `pre-commit run --all-files` clean; `ty check kornia` unchanged (its one diagnostic is
  a pre-existing `torch.cuda.amp.custom_fwd` deprecation in `lightglue.py`).
- `grep -rnE "#4170|issues/4170|#4156|#4101" kornia/ tests/ docs/` — no prose anywhere
  documents this defect, so there is nothing to retract.

Docstrings for `num_features` and `detect`'s `Returns` now state the top-k and prefix
contract, which the code did not previously honour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -335,20 +335,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the quotas always sum to exactly `num_features` and stay spread across scales. #4098's narrower `num_features=1`
   fallback, which handed the request's one slot to the largest-share level when every quota floored to zero, is
   the special case where the shortfall equals `num_features` and is superseded by the general apportionment.
-  The apportionment itself was subsequently removed by the fix below; only its first release carries it.
+  The apportionment was then removed again by #4222 below, inside this same unreleased window, so no release
+  ever carries it; the `num_features=1` behaviour it fixed still holds.
 
 * `MultiResolutionDetector.detect` now returns the `num_features` highest responses in the image (#4222). Each
   pyramid level used to be asked only for its own share of the budget, so a level could contribute at most
   `quotas[0]` (the upscaled level) or `sum(quotas[: i + 1 + up_levels])` (pyramid level `i`) -- both well under
   `num_features` for the finer levels. When an image's strongest maxima concentrated on one level, which is the
   common case rather than a corner case, the excess was never requested and so could not be ranked in by the
-  final global `topk`: the detector silently returned something other than the top-k, and `detect(k)` was not a
-  prefix of `detect(k')`. Every level is now searched for `num_features` candidates, which is exactly sufficient
+  final global `topk`: the detector silently returned something other than the top-k, and the sorted responses of
+  `detect(k)` were not a prefix of those of `detect(k')`. Every level is now searched for `num_features` candidates, which is exactly sufficient
   because the detections a level contributes to the global top-k are that level's own strongest. The per-level
   apportionment, including #4101's largest-remainder rounding, is removed with it. This changes which keypoints
   are returned for a given image: callers get strictly better-ranked detections, and any result that depended on
-  the old per-level spread will differ. It carries about 1.2x the candidates through the concatenation, which is
-  not measurable next to the response function and the non-maxima suppression.
+  the old per-level spread will differ. The prefix is on the sorted responses, not on which detection carries a
+  given response: `topk` breaks a tie by position, so exactly equal maxima can come back in a different order for
+  a different `num_features`. It carries about 1.2x the candidates through the concatenation, which is not
+  measurable next to the response function and the non-maxima suppression.
 
 * Make YUV and XYZ transformations compute integer inputs in `float32` instead of truncating their kernels, and
   preserve directly constructed `float64` coefficients. This makes `rgb_to_yuv(uint8)` return `float32` on the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -335,6 +335,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the quotas always sum to exactly `num_features` and stay spread across scales. #4098's narrower `num_features=1`
   fallback, which handed the request's one slot to the largest-share level when every quota floored to zero, is
   the special case where the shortfall equals `num_features` and is superseded by the general apportionment.
+  The apportionment itself was subsequently removed by the fix below; only its first release carries it.
+
+* `MultiResolutionDetector.detect` now returns the `num_features` highest responses in the image (#4222). Each
+  pyramid level used to be asked only for its own share of the budget, so a level could contribute at most
+  `quotas[0]` (the upscaled level) or `sum(quotas[: i + 1 + up_levels])` (pyramid level `i`) -- both well under
+  `num_features` for the finer levels. When an image's strongest maxima concentrated on one level, which is the
+  common case rather than a corner case, the excess was never requested and so could not be ranked in by the
+  final global `topk`: the detector silently returned something other than the top-k, and `detect(k)` was not a
+  prefix of `detect(k')`. Every level is now searched for `num_features` candidates, which is exactly sufficient
+  because the detections a level contributes to the global top-k are that level's own strongest. The per-level
+  apportionment, including #4101's largest-remainder rounding, is removed with it. This changes which keypoints
+  are returned for a given image: callers get strictly better-ranked detections, and any result that depended on
+  the old per-level spread will differ. It carries about 1.2x the candidates through the concatenation, which is
+  not measurable next to the response function and the non-maxima suppression.
 
 * Make YUV and XYZ transformations compute integer inputs in `float32` instead of truncating their kernels, and
   preserve directly constructed `float64` coefficients. This makes `rgb_to_yuv(uint8)` return `float32` on the
--- a/kornia/feature/scale_space_detector.py
+++ b/kornia/feature/scale_space_detector.py
@@ -659,8 +659,10 @@ class MultiResolutionDetector(nn.Module):
         model: response function, such as KeyNet or BlobHessian
         num_features: Number of features to detect. Every pyramid level is searched for this many
             candidates and the whole set is ranked together, so the result is the ``num_features``
-            highest responses in the image and the result for one value is a prefix of the result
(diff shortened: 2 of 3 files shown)


ducha-aiki (commented):
[P2] The newly documented result-prefix contract fails when responses tie. Both the per-level and global torch.topk calls may select or reorder equal scores differently as k changes. With one pyramid level, an identity response, and equal isolated maxima, detect(1) returned center [20,20], while detect(2) returned [30,20] before [20,20]. The new regression test around tests/feature/test_scale_space_detector.py:971 compares only response values, so tied scores let it pass even though LAF/keypoint identity violates the contract promised around kornia/feature/scale_space_detector.py:660 and repeated in the changelog.

Please either add a deterministic, k-independent secondary tie-break at both selection stages and assert the LAF prefix, or narrow the public docstring and changelog claim explicitly to response values. Focused detector tests and gradient checks otherwise passed; no other actionable findings were found.

ducha-aiki:
Both findings are valid. Fixed at `47185a66` by narrowing the claim, the second of the two options
offered.

**[P2] the prefix contract under ties.** Reproduced exactly as described — one pyramid level, an
identity response, three exactly equal isolated maxima:

```
k=1: responses=[1.0]            centres=[[20.0, 20.0]]
k=2: responses=[1.0, 1.0]       centres=[[30.0, 20.0], [20.0, 20.0]]
k=3: responses=[1.0, 1.0, 1.0]  centres=[[30.0, 20.0], [40.0, 40.0], [20.0, 20.0]]
```

I did **not** take the first option (a k-independent secondary tie-break at both stages). `topk`
breaks a tie by position, and the positions each stage ranks over are a function of `num_features`,
so making identity k-independent means imposing a total order over the whole candidate volume on
every call — a behaviour change in the detector's hot path to pin a case that only arises for
bit-exactly equal responses. The value prefix is what the fix in this PR actually establishes, and it
is what the existing regression test already pinned; the docstring and changelog were simply saying
more than the implementation delivers.

So the `num_features` docstring and the changelog entry now scope the prefix to the sorted response
vector and state that which detection carries a given response is not pinned when responses tie.

Added `test_the_response_prefix_survives_tied_responses` to make the narrowed claim load-bearing
rather than prose. It deliberately does **not** assert the permutation — that is `topk`'s to choose
and already differs across devices. It still discriminates: on pre-#4222 `main` the same input gives
a `k=3` reference of `[1.0, 1.0, 0.1375]` (main finds only two of the three tied maxima) and a `k=2`
result of `[1.0, 0.1375]`, so the prefix assertion fails there.

**Copilot, `CHANGELOG.md:338`.** Correct — both entries sit under the same `## Unreleased` heading, so
"only its first release carries it" is not merely confusing, it is false: no release will ever carry
the apportionment. Reworded to say it was removed again inside the same unreleased window, and that
the `num_features=1` behaviour it fixed still holds.

`TestMultiResolutionDetector` passes in full (148 nodes) across float16, bfloat16, float32 and
float64; Ruff format and check clean.

ducha-aiki (commented):
Re-review of latest head 47185a66: no actionable findings. The previous tied-response concern is fully resolved by narrowing the contract to sorted response values and explicitly excluding keypoint/LAF identity under ties. Exact-tie cases retain invariant response prefixes despite changing centers, and random plus tied multi-level results matched an exhaustive top-k reference. Sixteen focused dtype cases and all 40 CPU-float32 MultiResolutionDetector tests passed; git diff --check was clean.

PR #4223, opened 2026-09-04
feat(feature): add laf_is_filled, the detector padding predicate

ducha-aiki:
## Why

Fixed-shape detectors pad the slots no detection filled with an all-zero LAF and a
zero response. Those slots have to be dropped before matching — their descriptors are
all the descriptor of a patch sampled at the origin, *one and the same vector*, so they
match each other at zero distance and a mutual nearest-neighbour test does not reject
them.

`LocalFeature.forward`'s docstring already told callers of a hand-rolled pipeline that
they must do this, and handed them the expression to copy:

```python
valid = lafs.ne(0).any(-1).any(-1)  # (B, N); the zero LAF marks a padded slot
```

The same expression was also written out at **four sites inside `kornia/feature`**, in
two files (`integrated.py` ×2, `scale_space_detector.py` ×2). This gives it a name.

## Public, not private

The argument is encapsulation rather than convenience. The zero-LAF encoding is
**already public** — the docstring documents it and instructs users to depend on it —
so a function does not widen the exposure, it narrows the number of ways to get it
wrong. It also leaves a migration path if the padding sentinel ever changes, which a
snippet copied into user code does not.

Placed next to `laf_is_valid`, exported from `kornia.feature`, and added to
`docs/source/feature.laf.rst` — which is the act that makes a symbol public under
`docs/source/get-started/stability.rst` ("a symbol is public when it appears in the
rendered documentation"). LAF utilities are the classical, non-pretrained part of
`kornia.feature`, so this lands in **stable core** and carries the full promises.

## Also in this change

`integrated.py`'s two sites move **out of the `for batch_idx` loop**, where the mask was
recomputed once per image pair, to a single whole-batch call. Same values; two fewer
redundant reductions per iteration. They could not simply be swapped in place because
they were called on `lafs0[batch_idx]`, which is `(N, 2, 3)` and would fail
`KORNIA_CHECK_LAF`.

The docstring snippet now calls the function, with the import shown so it stands alone.

## Verified

- **24 new tests** pass across `float32`, `float64`, `float16`, `bfloat16`. They include
  the discriminating case — a detection at the image **origin**, which has a zero centre
  but a nonzero shape and must not read as padding — plus a shape/dtype/device check, a
  `ShapeError` check, JIT scripting, and a `fullgraph=True` dynamo check, matching the
  `laf_is_valid` sibling class.
- The `ShapeError` check is deliberately not `pytest.raises(Exception)`: while writing
  these tests that broader form passed *before the function existed*, by swallowing the
  `AttributeError`.
- `tests/feature`: 688 passed, 94 skipped. **MPS**: 178 passed, 41 skipped
  (`PYTORCH_ENABLE_MPS_FALLBACK` unset). dynamo/compile: 9 passed. Doctests: 10 passed.
- A **clean `-W` docs build succeeds with zero warnings**, and its output is identical to
  a clean build of `main` apart from timings and object addresses. `laf_is_filled` renders
  on `feature.laf.html` and in the index.
- `pre-commit run --all-files` clean; `ty check kornia` unchanged (its one diagnostic is a
  pre-existing `torch.cuda.amp.custom_fwd` deprecation in `lightglue.py`).
- No hand-rolled copy of the idiom is left anywhere in `kornia/` apart from the definition.


ducha-aiki (commented):
No actionable findings. The helper correctly captures the existing all-zero-LAF padding contract without misclassifying a real detection at the image origin, is exported and documented as a public API, and replaces the internal copies without changing their semantics. Independent validation passed: test_laf.py (144 passed, 9 skipped), test_integrated.py plus test_scale_space_detector.py (91 passed, 10 skipped), and the full pre-commit suite.

Issue #4174, opened 2026-09-02
Bug: `validate_bbox` / `validate_bbox3d` raise `RuntimeError` from `Tensor.view` for non-contiguous `(B, N, ...)` input


ducha-aiki:
## Summary

Both 2D/3D validators flatten rank-4 input with `Tensor.view`:

- `validate_bbox`: `boxes = boxes.view(-1, 4, 2)` (`kornia/geometry/bbox.py:105` at `8d685a37`)
- `validate_bbox3d`: `boxes = boxes.view(-1, 8, 3)` (`kornia/geometry/bbox.py:141`)

`view` cannot merge the two leading dimensions unless they are contiguous with respect to each other, so a `(B, N, 4, 2)` tensor that the signature accepts raises `RuntimeError: view size is not compatible with input tensor's size and stride ...` instead of returning a `bool`. The docstring documents `(B, N, 4, 2)` as accepted input with no stride qualification, and `validate_bbox` is otherwise total — it returns `False` rather than raising for every invalid input.

This is not limited to exotic layouts. A plain sub-selection of boxes fails.

## Reproduction

```python
import torch
from kornia.geometry.bbox import validate_bbox, validate_bbox3d

x = torch.zeros(2, 3, 4, 2)
for name, t in [
    ("contiguous",        x),
    ("x.transpose(0, 1)", x.transpose(0, 1)),
    ("x[:, 1:]",          x[:, 1:]),                          # drop the first box of each batch item
    ("x[::2]",            torch.zeros(4, 3, 4, 2)[::2]),
    ("x.expand(...)",     torch.zeros(1, 3, 4, 2).expand(2, 3, 4, 2)),
    ("x[:1]",             x[:1]),
]:
    try:
        print(f"{name:20s} contiguous={t.is_contiguous()!s:5s} -> {validate_bbox(t)}")
    except RuntimeError as e:
        print(f"{name:20s} contiguous={t.is_contiguous()!s:5s} -> RuntimeError: {str(e)[:40]}")

validate_bbox3d(torch.zeros(2, 3, 8, 3).transpose(0, 1))      # same failure
```

## Observed

```
contiguous           contiguous=True  -> True
x.transpose(0, 1)    contiguous=False -> RuntimeError: view size is not compatible with in
x[:, 1:]             contiguous=False -> RuntimeError: view size is not compatible with in
x[::2]               contiguous=False -> RuntimeError: view size is not compatible with in
[...]

PR #4225, opened 2026-09-04
fix(feature): make HyNet/SOSNet work in CPU half precision, retire 11 manifest lines

### LINKED ISSUE #4224: SOSNet returns NaN in float16 for a zero descriptor: the eps=1e-10 guard underflows
## Describe the bug

`SOSNet` returns NaN in `float16` whenever the pre-normalisation descriptor is exactly zero, because
the `eps` guard that exists to prevent precisely that is not representable in the dtype.

`SOSNet.forward` (`kornia/feature/sosnet.py`) is:

```python
def forward(self, input: torch.Tensor, eps: float = 1e-10) -> torch.Tensor:
    KORNIA_CHECK_SHAPE(input, ["B", "1", "32", "32"])
    descr = self.desc_norm(self.layers(input) + eps)
    ...
```

Two facts stack:

1. **The guard is inert in `float16`.** `torch.tensor(1e-10, dtype=torch.float16)` is exactly `0.0` —
   float16's smallest *subnormal* is 5.96e-8, four orders of magnitude above 1e-10. So `+ eps` is a no-op
   and `descr` stays exactly zero.
2. **`desc_norm` then divides by zero.** `nn.LocalResponseNorm(256, alpha=256.0, beta=0.5, k=0.0)` over a
   128-channel tensor is exactly `x / sqrt(sum_c x^2)` (the window covers every channel), so an all-zero
   input gives `0 / 0` → NaN.

`self.layers(input)` really does reach exactly zero for a plain patch: every `Conv2d` in the stack has
`bias=False` and, at the default `pretrained=False`, each `BatchNorm2d(affine=False)` runs in eval mode with
`running_mean=0`, so an all-zero input propagates as exactly zero to the end. Measured `absmax = 0.0` before
the `+ eps`.

## Reproduction

On unmodified `main` @ `64c4d902`, torch 2.14.0, **MPS**, with `PYTORCH_ENABLE_MPS_FALLBACK` unset:

```python
import torch
from kornia.feature import SOSNet

for dt in (torch.float32, torch.float16, torch.bfloat16):
    m = SOSNet().to("mps", dt).eval()
    print(dt, m(torch.zeros(1, 1, 32, 32, device="mps", dtype=dt))[0, :3])

# torch.float32  tensor([0.0884, 0.0884, 0.0884], device='mps:0', grad_fn=<SliceBackward0>)
# torch.float16  tensor([nan, nan, nan], device='mps:0', dtype=torch.float16,
#                grad_fn=<SliceBackward0>)
# torch.bfloat16 tensor([0.0884, 0.0884, 0.0884], device='mps:0', dtype=torch.bfloat16,
#                grad_fn=<SliceBackward0>)
```

`float32` and `bfloat16` return the well-defined constant `1/sqrt(128) = 0.0884`; only `float16` NaNs.
bfloat16 has float32's exponent range, so it carries `1e-10` intact (`1.0004e-10`) and the guard works.

A realistic patch is unaffected in every dtype (`torch.rand` input, norm 1.004 in fp16), so this is the
degenerate path only.

**Why CPU does not show it today.** The same call on CPU raises
`NotImplementedError: "avg_pool3d_out_frame" not implemented for 'Half'` first: `local_response_norm` routes a
4-D input through `avg_pool3d`, which has no CPU half kernel. That is a separate gap; once it is fixed the
NaN is reachable on CPU too — verified by running the fixed path locally. **CUDA is untested** (no GPU
available for this report).

## Scope

- **`HyNet` is not affected in practice**, although it carries the same `eps_l2_norm: float = 1e-10`
  constant and the same `LocalResponseNorm`. Its `FilterResponseNorm2d` layers keep the pre-normalisation
  tensor away from exactly zero; measured finite (norm ≈ 1.0) for the same all-zero patch in float16 and
  bfloat16, on both CPU and MPS.
- This is the same defect class as #4021 (`normalize_quaternion`, `eps=1e-12`, open) and the closed #4123
  (`PatchAffineShapeEstimator`, `eps=1e-10`).

## Expected behavior

A finite descriptor in `float16` — the same constant `0.0884` vector the other dtypes return.

The repository already has the idiom for this, in two places, and it is not an eps bump:

- `kornia/feature/affine_shape.py:92` lifts a half patch to `float32` for the guarded arithmetic and casts
  the result back (the #4123 fix).
- `kornia/feature/siftdesc.py` `_rootsift` and the gradient helper do the same, and their docstrings spell out
  why: *"A float16 input cannot carry the 1e-10 guard -- it underflows to zero"*.

A dtype-aware `eps` alone does **not** work here, because the guard is squared before it is used: `eps**2`
must itself be representable, so float16 would need `eps >= sqrt(5.96e-8) = 2.4e-4`. Descriptor components
measure ~5e-3 to 7.6e-3, so such an eps would shift a genuine descriptor by 3-5%.

## Environment

kornia `64c4d902`, torch 2.14.0, Python 3.11, macOS, MPS. Also checked on torch 2.9.1 (CPU) for the eps
representability and the float32/bfloat16 reference values.

---

ducha-aiki:
Closes the five undiagnosed half-precision manifest nodes in `tests/feature` listed on #4153. They
turned out to be three unrelated causes.

## 1. `HyNet` and `SOSNet` in half precision — two defects on one line

**Closes #4224.**

**(a) CPU raises.** `NotImplementedError: "avg_pool3d_out_frame" not implemented for 'Half'` — and it
is *not* a JIT problem. It fires on the eager `model(patches)` call; `test_jit` is simply the only
dtype-parametrized test in those classes. Both models end with `nn.LocalResponseNorm` fed a 4-D
`(B, 128, 1, 1)` tensor, and `F.local_response_norm` routes `dim >= 4` through `avg_pool3d`, which
has no CPU `float16`/`bfloat16` kernel on any supported torch.

**(b) Where the kernel does exist (MPS, CUDA), the descriptors come back all-NaN** for a degenerate
patch. The `eps` that keeps the normalisation's division defined — `SOSNet.forward`'s `eps` and
`HyNet`'s `eps_l2_norm`, both `1e-10` — is not representable in `float16` and flushes to `0.0`,
leaving `0/0`. Every `Conv2d` in these models has `bias=False` and every `BatchNorm2d` is
`affine=False` with `running_mean == 0`, so a zero *or constant* patch reaches the normalisation as
exactly zero. `HyNet` reaches it with the supported `is_bias=False` (the TLU bias, not `is_bias_FRN`).

**One fix for both**: take that single normalisation step in `float32` for a half input and cast
back — the treatment `kornia.feature.siftdesc` already gives its own `1e-10` guards, and the same
shape as the #4123 fix in `affine_shape.py:92`. `float32` and `float64` take the original expression.

Verified against `origin/main`, 16 patches x 128 dims:

| | result |
| --- | --- |
| CPU `float32` / `float64`, both models | **bitwise identical** |
| MPS `float32`, both models | **bitwise identical** |
| CPU + MPS half, zero and constant patches | finite, `0.0884` — matching what `float32` returns |

**Speed.** Two identical controls put the noise floor at ~1%. The step is ~2 ms against a
211–235 ms `float32` model forward and a ~2280 ms half forward — 0.9% and 0.09% of the respective
totals. In half the lift is not a cost at all: CPU half arithmetic is ~10x slower than `float32` for
these convolutions, so the step measured **0.94x (float16)** and **0.96x (bfloat16)** against native
half.

An earlier revision of this PR used a 3-D view to reach `avg_pool2d` instead. The lift replaces it:
it fixes (b) as well as (a), it is one mechanism rather than two, and it removes the 0.75 eps MPS
`float32` deviation the flatten introduced.

## 2. `TestALIKED::test_descriptor_normalized` asserted something unreachable

`atol=1e-4` on an L2 norm. The norms come back as `1.0` or its single representable neighbour in
*every* dtype, so over 2440+ descriptors x 25 seeds x torch 2.9.1/2.14.0 the worst deviation is
exactly **one eps** — 9.8e-4 (float16), 7.8e-3 (bfloat16). No implementation can pass `1e-4` there.
Computing the same norms in `float32` gives 0.55 eps, i.e. the descriptors really are unit vectors to
the limit of the storage dtype; the extra eps is the half `norm()` reduction's own rounding.

The bound is now `max(1e-4, 2 * finfo(dtype).eps)` — set by the dtype, not by the observed miss, with
a 2x margin. `float32`/`float64` keep the original `1e-4` (838x margin).

## 3. The three response tests are the #4081 hard-coded-tolerance pattern

A response is a difference of quantities of order 1 — the blurred image and its derivatives — so its
absolute error is set by one ULP at 1.0 (9.8e-4 float16, 3.9e-3 bfloat16) and **not** by the
response's own magnitude, which is 0.007–0.28 here. `atol=1e-4` sits below that floor.

The repository's per-dtype tolerances hold, with the worst element at 5.7–65% of its allowance and
**identical to the last digit on torch 2.9.1 and 2.14.0**:

| test | float16 | bfloat16 |
| --- | --- | --- |
| `TestCornerHarris::test_corners_batch` | 1.5% | 1.2% |
| `TestCornerGFTT::test_corners_batch` | 65% | 12% |
| `TestBlobHessian::test_blobs_batch` | 5.7% | 5.9% |
| `TestBlobDoGSingle::test_blobs_batch` | 51% | 39% |

`TestCornerHarris::test_corners_batch` is **not** in a manifest, but it sits at **91%** of its current
allowance in bfloat16 — a latent manifest line — so it gets the same treatment. The two non-batch
`test_corners` tests use `rtol=1e-3`, sit at ≤55%, and are left alone.

## What fixing (1) uncovered

`TestSOSNet::test_jit[cpu-float16]` ran for the first time and **the test turned out to be vacuous**.
It scripts a *second, independently initialised* `SOSNet()` and compares it against the first. That
passes today only because `torch.ones` drives a random-init SOSNet to exactly zero, so both sides come
out as the constant `1/sqrt(128)`. Verified: with the old input, two *different* models compare
**equal** in float32. It now uses `torch.rand` and scripts the same instance, matching
`TestHyNet::test_jit`; verified it discriminates in float32 and float16.

New pins: `test_degenerate_patch_gives_finite_descriptors` on both models, over zero and constant
patches, with `HyNet` parametrized over `is_bias`. They fail on the previous revision of this branch
with "256 non-finite descriptor entries".

## Verification

- Full `tests/feature` on **torch 2.14.0 and 2.9.1** x `float32`/`float64`/`float16`/`bfloat16`. The only
  remaining failures are the two already fixed in #4220 and #4221.
- Restored the old manifests and re-ran focus mode: exactly **5 (float16) and 6 (bfloat16)** strict XPASS
  failures — precisely the lines deleted here, no more and no fewer.
- MPS `float32` on the four touched test files: 39 passed, with `PYTORCH_ENABLE_MPS_FALLBACK` unset.
- `tests/core/test_weights_prefetch.py`: 84 passed. Doctests for both touched modules pass.
- `pre-commit run --all-files` exit 0; `ty check kornia` unchanged (its one diagnostic is the pre-existing
  `torch.cuda.amp.custom_fwd` deprecation in `lightglue.py`).
- The complete `--verify-known-failures` contract **cannot** run here — the manifest is stamped `Linux`
  and complete mode refuses on Darwin. That leg is CI's.

Rebased onto `main` after #4220 merged, which took the adjacent
`TestLAFAffNetShapeEstimator::test_toy` lines. With both changes in, **`tests/feature` has no CPU
half-precision manifest lines left at all** -- down from 6 `cpu_float16` and 7 `cpu_bfloat16`. Post-rebase
re-run: `tests/feature` float16 is 679 passed / 0 failed, and bfloat16 is 684 passed / 1 failed, that one
being the `topk` tie fixed in the still-open #4221.


ducha-aiki (commented):
[P2] CPU float16 execution is enabled, but SOSNet still returns entirely NaN descriptors for zero or constant patches. The 1e-10 guard at kornia/feature/sosnet.py:103 underflows to zero in float16 before the new half-precision LocalResponseNorm path divides by the zero norm. On latest head fd854204, default SOSNet produced 256/256 NaNs for two zero or constant patches across repeated seeds; HyNet with its supported is_bias=False configuration has the same failure. The randomized JIT test at tests/feature/test_sosnet.py:46-56 avoids this path, so its passing result does not establish the changelog claim that both models now run in CPU half precision.

Although tracked separately as #4224 and previously reachable on MPS, this PR newly exposes the defect on CPU, advertises half-precision support, and removes the SOSNet expected-failure entry. Please perform the guarded normalization in float32 for float16 inputs, cast the result back, and add a finite-output zero/constant-patch regression. Other focused CPU/MPS dtype, JIT, compile, equivalence, lint, and diff checks passed. The latest rebase also resolves the manifest conflicts correctly.

ducha-aiki:
[P2] confirmed and fixed at `0a31e316`, including the part I had got wrong.

**On HyNet — you are right and my scope note on #4224 was wrong.** I had checked only the default
constructor, where the pre-normalisation tensor never lands on exactly zero, and wrote "HyNet is not
affected in practice" into the issue. With the supported `is_bias=False` (the TLU bias, not
`is_bias_FRN`) it is 256/256 NaN on a zero patch in float16. I have corrected #4224.

Measured, on this branch's previous head:

| dtype | patch | SOSNet | HyNet() | HyNet(is_bias=False) |
| float16 | zeros | **256/256 NaN** | 0 | **256/256 NaN** |
| float16 | constant 0.5 | **256/256 NaN** | 0 | 0 |
| bfloat16, float32, float64 | either | 0 | 0 | 0 |

bfloat16 escapes because it carries float32's exponent range, so `1e-10` survives as `1.0004e-10`.

**The fix is the one you asked for, and it replaces the 3-D view rather than joining it.** Both
models now take that single normalisation step in `float32` for a half input and cast back — the
treatment `siftdesc.py` already gives its own `1e-10` guards, and the shape of the #4123 fix at
`affine_shape.py:92`. That is strictly better than the flatten: it fixes the NaN as well as the
missing `avg_pool3d` half kernel, it is one mechanism instead of two, and because `float32` now takes
the *original* expression, the 0.75 eps MPS `float32` deviation the flatten introduced is gone.

Re-verified against `origin/main`: **bitwise identical** on CPU `float32`/`float64` **and on MPS
`float32`**, both models.

**Speed, since it was worth checking before committing to a float32 path.** Two identical controls
put the noise floor at ~1%:

| | float32 | float16 | bfloat16 |
| --- | --- | --- | --- |
| normalisation step, B=256 | 2.13 ms (unchanged) | 2.17 ms | 2.16 ms |
| lift vs native half | — | **0.94x** | **0.96x** |
| backbone, same batch | 211–235 ms | ~2280 ms | ~2283 ms |

The step is 0.9% of a `float32` forward and 0.09% of a half one. In half the lift is *faster* than
native: CPU half arithmetic is ~10x slower than `float32` for these convolutions.

**On the changelog claim** — narrowed, and it now states both defects, both `eps` names, the
`is_bias=False` condition, and that `float32`/`float64` are bitwise unchanged on every device.

**Regression pins** added to both models: `test_degenerate_patch_gives_finite_descriptors` over zero
and constant patches, `HyNet` parametrized over `is_bias`. They fail on the previous head with
"256 non-finite descriptor entries" and pass now.

Rebased onto `main` past #4221 and #4222. `tests/feature` is now **0 failures** in both half dtypes
on torch 2.14.0, and clean on 2.9.1; MPS float32 green; `pre-commit --all-files` exit 0.

ducha-aiki:
### `torch.compile` / dynamo: clean, and the branch never reaches the graph

The normalisation now branches on `dtype`, so that surface is worth stating explicitly.

- `torch.compile(model, fullgraph=True)` succeeds for **both models on float32, float16 and
  bfloat16**.
- `torch._dynamo.explain` reports **1 graph, 0 breaks** (53 ops HyNet, 24 SOSNet).
- `dtype` is static metadata, so the branch is resolved **at trace time** — dynamo specialises once
  per dtype and the conditional never appears in the graph. It is the "Python `if` on a Python value"
  case, not a data-dependent branch on a tensor.
- The compiled-vs-eager delta on the **float32** path is byte-for-byte what it is on `main`
  (HyNet 2.332e-06, SOSNet 1.803e-06) — pre-existing inductor fusion, untouched here. Compiled half
  output is finite and within 0.31–0.53 eps of eager.
- Pinned by a new `test_dynamo` on both models. `tests/feature` is in the dynamo leg's directory list
  in `pr_test_cpu.yml`, so its `-k "dynamo or compile"` filter collects them (confirmed with
  `--collect-only`). Note the shared `torch_optimizer` fixture is `partial(torch.compile, backend=...)`
  with **no** `fullgraph=True`, so the graph-break count is the manual check above, not something that
  test enforces.

### CUDA: not measured here, and the one place a slowdown is plausible

I have no GPU on this machine, so **every CUDA claim below is unverified**. The reasoning that makes
it worth checking rather than assuming: on CPU the lift is *free or faster* because CPU half
arithmetic is ~10x slower than float32 for these convolutions. On CUDA that argument inverts — half
kernels are fast there — so the lift is the one change in this PR that could cost something. It
should not cost much, since it touches a single normalisation over a `(B, 128, 1, 1)` tensor while
the backbone is seven 32x32 convolutions, but that is an argument, not a measurement.

Run from the repo root on this branch:

```python
import torch, kornia
from kornia.feature import HyNet, SOSNet
print(kornia.__file__, torch.__version__, torch.cuda.get_device_name(0))

# 1. does the defect reach CUDA users today?  (run this on `main`, not this branch)
for dt in (torch.float32, torch.float16, torch.bfloat16):
    for name, ctor in (("SOSNet", SOSNet), ("HyNet(is_bias=False)", lambda: HyNet(is_bias=False))):
        m = ctor().to("cuda", dt).eval()
        with torch.no_grad():
            d = m(torch.zeros(1, 1, 32, 32, device="cuda", dtype=dt))
        print(f"main  {dt} {name:22s} finite={bool(torch.isfinite(d).all())}")

# 2. float32/float64 must be bitwise unchanged.  Save on `main`, compare on this branch.
import pickle
out = {}
for cls in (HyNet, SOSNet):
    for dt in (torch.float32, torch.float64):
        torch.manual_seed(7); m = cls().to("cuda", dt).eval()
        torch.manual_seed(8); x = torch.rand(16, 1, 32, 32, device="cuda", dtype=dt)
        with torch.no_grad():
            out[(cls.__name__, str(dt))] = m(x).cpu()
pickle.dump(out, open("/tmp/ref_cuda.pkl", "wb"))          # on main
# then on this branch, same block, and: torch.equal(new, ref[key]) for every key

# 3. speed: the lift vs native half, on the step that changed
import torch.utils.benchmark as bench
dn = torch.nn.LocalResponseNorm(256, alpha=256.0, beta=0.5, k=0.0).cuda()
for dt in (torch.float16, torch.bfloat16):
    x = torch.randn(256, 128, 1, 1, device="cuda", dtype=dt)
    native = lambda t: dn(t + 1e-10)
    lifted = lambda t: dn(t.float() + 1e-10).to(t.dtype)
    a = bench.Timer("f(x)", globals={"f": native, "x": x}).blocked_autorange(min_run_time=1.0).median
    b = bench.Timer("f(x)", globals={"f": lifted, "x": x}).blocked_autorange(min_run_time=1.0).median
    c = bench.Timer("f(x)", globals={"f": native, "x": x}).blocked_autorange(min_run_time=1.0).median
    print(f"{dt}: native {a*1e6:.1f}us / {c*1e6:.1f}us (control, should agree ~1%)  lifted {b*1e6:.1f}us  {b/a:.2f}x")
    # and the denominator it has to be judged against:
    m = SOSNet().to("cuda", dt).eval(); xx = torch.rand(256, 1, 32, 32, device="cuda", dtype=dt)
    t = bench.Timer("m.layers(x)", globals={"m": m, "x": xx}).blocked_autorange(min_run_time=1.0).median
    print(f"        backbone {t*1e3:.2f}ms  ->  step is {b/t*100:.2f}% of the forward")

# 4. compile still fullgraph on CUDA
for cls in (HyNet, SOSNet):
    for dt in (torch.float32, torch.float16):
        torch._dynamo.reset()
        m = cls().to("cuda", dt).eval(); x = torch.rand(2, 1, 32, 32, device="cuda", dtype=dt)
        with torch.no_grad():
            e = m(x); g = torch.compile(m, fullgraph=True)(x)
        print(f"{cls.__name__} {dt} fullgraph OK, finite={bool(torch.isfinite(g).all())}, "
              f"maxabs vs eager {(g.float()-e.float()).abs().max():.2e}")
```

Two notes on running it. Step 3's **control** matters: `native` is timed twice, and if the two do not
agree to ~1% the ratio is noise — my first CPU pass at this had a ±20% floor and produced one cell
that read "1.51x SLOWER" and re-measured at 0.95x, faster. And for step 1, if `main` comes back
finite on CUDA float16, then `avg_pool3d` behaves differently there than on MPS and the NaN half of
this PR is MPS-only — worth knowing either way, and it does not change the fix.

If step 3 shows the lift costs more than a percent or so of the forward on CUDA, the fallback is to
narrow the gate from `dtype in (float16, bfloat16)` to float16 only: bfloat16 carries float32's
exponent range, so it never underflows the `1e-10` guard and does not strictly need the lift. It
needs it on **CPU** only because `avg_pool3d` has no bfloat16 kernel there — which a
`device.type == "cpu"` term in the gate would cover. I have left the gate dtype-only because it is
simpler and costs nothing on CPU; say the word and I will narrow it.

ducha-aiki:
### CUDA leg: run, with numbers

Ran the block from the CUDA comment on an RTX 4090 -- torch 2.14.0+cu130, CUDA 13.0, Python 3.11,
Linux x86_64. Everything below is measured at `2775801f` against a `main` worktree at `25b2ca72`,
with `kornia.__file__` printed in every block to confirm which tree was imported.

**Summary: the defect does reach CUDA, `float32`/`float64` are bitwise unchanged there, the lift
costs ~1% of a half forward, and the gate should stay `dtype`-only. One new test is red on CUDA for a
reason unrelated to this change -- last section.**

#### 1. Step 1 -- the defect reaches CUDA users today

Not MPS-only. On `main`, non-finite entries out of 256:

| dtype | patch | `SOSNet` | `HyNet()` | `HyNet(is_bias=False)` |
| float16 | zeros | **256/256** | 0 | **256/256** |
| float16 | constant 0.5 | **256/256** | 0 | 0 |
| bfloat16 | either | 0 | 0 | 0 |
| float32 / float64 | either | 0 | 0 | 0 |

Cell for cell the CPU table in your reply, including bfloat16 escaping via float32's exponent range.
On this branch every cell is 0/256 and the descriptors come back at `0.088379` -- `1/sqrt(128)`, what
float32 returns.

So `avg_pool3d` does **not** behave differently on CUDA, and the NaN half of this PR is not MPS-only.

#### 2. Step 2 -- float32 / float64 bitwise unchanged on CUDA

`torch.equal` **True** for both models on both dtypes, 16x1x32x32, seeds 7/8, max abs difference
exactly `0.000e+00`.

#### 3. Step 3 -- speed: ~1% of a half forward, and the gate should stay as it is

The tight measurement is the step itself, `(256, 128, 1, 1)`, interleaved medians over 7 rounds with
a same-expression control:

| dtype | native | control | lifted | ratio |
| float16 | 463.9 us | 464.0 us (0.0%) | 476.0 us | **1.026x** |
| bfloat16 | 469.3 us | 466.0 us (0.7%) | 484.7 us | **1.033x** |

+12-15 us. Worth knowing for its own sake: `LocalResponseNorm` -> `avg_pool3d` is expensive on CUDA
and is itself ~40% of a half forward on this GPU, so those 12 us land at **~1% of the forward**.

End to end I compared this branch's `m(x)` against a byte-exact in-process re-creation of `main`'s
forward -- same weights via `load_state_dict`, same input, same `KORNIA_CHECK_SHAPE` -- over 12
configurations:

| model | dtype | B=16 | B=256 |
| --- | --- | --- | --- |
| SOSNet | float32 *(control: identical code both sides)* | 0.937x | 1.002x |
| SOSNet | float16 | 1.065x | 1.022x |
| SOSNet | bfloat16 | 0.984x | 0.869x |
| HyNet | float32 *(control)* | 1.024x | 1.026x |
| HyNet | float16 | 1.028x | 0.981x |
| HyNet | bfloat16 | 1.022x | 0.982x |

The float32 rows run identical code on both sides, so they are this machine's noise floor:
0.94x-1.03x. No half row sits outside it. **Nothing here justifies narrowing the gate to float16 or
adding a `device.type == "cpu"` term** -- leave it dtype-only, as you have it.

(Your instinct about the control was right, and it bit me: a first pass without interleaving produced
a 1.51x that re-measured at 0.95x.)

#### 4. Step 4 -- compile on CUDA

- `torch.compile(model, fullgraph=True)` succeeds for **both models on float32, float16 and
  bfloat16**.
- `torch._dynamo.explain`: **1 graph, 0 breaks** -- 53 ops HyNet, 24 SOSNet, identical to your CPU
  figures. The dtype branch is resolved at trace time on CUDA as well.
- Compiled zero-patch output is finite in both half dtypes, so the fix survives inductor.

#### 5. Not in the instructions, and worth adding to the PR body

**The CUDA half suite goes from 8 red to 0.** `--isolate-half-precision`, `float16,bfloat16`, over
`test_hynet.py test_sosnet.py test_responces_local_features.py test_aliked.py`:

| | result |
| --- | --- |
| `main` | **8 failed**, 55 passed, 6 skipped, 2 errors |
| this branch | **0 failed**, 75 passed, 6 skipped, 2 errors |

The 8: `TestSOSNet::test_jit[cuda-float16]`, `TestCornerGFTT::test_corners_batch` (both dtypes),
`TestBlobHessian::test_blobs_batch[cuda-bfloat16]`, `TestBlobDoGSingle::test_blobs_batch` (both),
`TestALIKED::test_descriptor_normalized` (both). None is in any manifest -- there is no CUDA profile --
so this PR quietly retires 8 latent CUDA half failures too. The +20 passed is those 8 plus the 12 new
parametrized pins.

The 2 errors are `TestSOSNet::test_shape[cuda]` and `TestCornerHarris::test_shape[cuda]`, both
"previous item was not torn down properly", and both appear identically on `main` -- a pre-existing
`--isolate-half-precision` teardown artefact, unrelated to this PR.

**Half precision gets more accurate, not merely finite.** Against a float64 model carrying the same
weights, 64 random patches, max abs error:

| | this branch | `main` |
| --- | --- | --- |
| SOSNet float16 | **2.28e-04** | 2.15e-03 |
| SOSNet bfloat16 | **1.73e-03** | 2.16e-03 |
| HyNet float16 | 4.64e-04 | 4.64e-04 |
| HyNet bfloat16 | 4.23e-03 | 4.23e-03 |

SOSNet float16 lands ~10x closer to the float64 result and nothing gets worse. That is a stronger
argument for the lift than "it stops the NaN". The flip side, which the changelog implies but does
not state: **half output is not bitwise unchanged** -- 0.25-2.25 eps, 2400-7800 of 8192 elements move.
Intended and in the right direction, but a half-precision user diffing descriptors across versions
will see it.

#### 6. Other checks on this box

- Full `tests/feature` on CUDA float32: 2 failed / 659 passed. Both --
  `TestLAFAffNetShapeEstimator::test_toy[cuda-float32]` and
  `TestOrientationHalfPrecisionIsFinite::test_half_precision_matches_float32[cuda]` -- reproduce on
  `main`. **No CUDA float32 regression.**
- Full `tests/feature` on CPU half with the edited manifests, torch 2.14.0: **688 passed / 0 failed**
  (float16), **694 passed / 0 failed** (bfloat16).
- Restored `main`'s manifests and re-ran focus mode over the four touched files: exactly **5
  (float16) and 6 (bfloat16) strict XPASS**, precisely the deleted lines -- independently reproduces
  your claim.
- CPU dynamo, both models, float32/float64/float16/bfloat16: 8 passed.
- The complete `--verify-known-failures` contract **cannot run here either**, for a different reason
  than on Darwin: the manifests are stamped `pytorch: 2.9.1` and this box carries 2.14.0, so complete
  mode refuses (focus mode warns and proceeds). That leg is still CI's.

#### 7. One finding from the CUDA run: the new `test_dynamo` tests are red on CUDA float32

```
FAILED tests/feature/test_hynet.py::TestHyNet::test_dynamo[cuda-float32-inductor]
  Greatest absolute difference: 2.51e-04 (up to 1e-05 allowed); 207/256 elements
FAILED tests/feature/test_sosnet.py::TestSOSNet::test_dynamo[cuda-float32-inductor]
  Greatest absolute difference: 2.29e-04 (up to 1e-05 allowed); 206/256 elements
```

The cause is not the dtype branch. It is cuDNN TF32: `AGENTS.md` records that TF32 matmul is off by
default but **cuDNN convolutions keep PyTorch's TF32 default**, and `conftest.py` only sets
`torch.set_float32_matmul_precision`. Eager and inductor then pick different convolution algorithms,
so on CUDA float32 the assertion measures TF32 conv reproducibility across seven convolutions rather
than the thing it means to pin. With `torch.backends.cudnn.allow_tf32 = False` set before the first
compile the delta drops to **6.9e-07 (HyNet) and 6.6e-07 (SOSNet)** against a 3.6e-05 allowance -- a
50x margin.

Scope, stated fairly: this is a pre-existing family, not a new class of problem.
`tests/feature/test_xfeat.py::TestXFeatModel::test_dynamo[cuda-float32-inductor]` fails identically
on `main`. CI's dynamo leg is CPU-only (`pr_test_cpu.yml`, `ubuntu-latest`), so nothing turns red.
But the PR does add two new reds to any local CUDA dynamo run.

Two ways out, and I would take the second:

1. `with torch.backends.cudnn.flags(allow_tf32=False):` around the comparison. **Caveat from actually
   trying it**: inductor's on-disk FX graph cache is not keyed on that flag, so entering the context
   after an earlier TF32-on compile in the same process raised
   `AssertionError: expected size 32==32, stride 1024==1 ...` out of `assert_tensor_metadata`. It
   works only from a clean cache, which is fragile for a test.
2. Give the assertion a TF32-sized bound on CUDA float32, with a one-line comment saying cuDNN
   convolutions run TF32 by default there, so on CUDA the test pins the branch surviving capture
   rather than conv bit-reproducibility. Sizing it: the worst delta over 5 seeds is 2.77e-04 (HyNet)
   and 2.67e-04 (SOSNet), so `atol=1e-3` gives a ~3.6x margin. `atol=5e-4` would be 1.8x, too thin
   for a test that builds an unseeded random model.

Either is fine by me; so is leaving it and saying so in the PR body.

A wider review of the diff is running separately and I will post it after this.

ducha-aiki:
Both items applied at `f8fbd065`.

**The new `test_dynamo` reds — taking your option 2.** I verified the diagnosis independently rather
than take it: `conftest.py` calls `set_float32_matmul_precision` only under `--tf32` and **never
touches `torch.backends.cudnn.allow_tf32`**, which defaults to `True`, so cuDNN convolutions run TF32
on CUDA in every configuration of this suite. That is the same gap `AGENTS.md` records.

Option 1 is out on your evidence — an inductor FX graph cache that is not keyed on the flag makes it
work only from a clean cache, which is not something to build a test on.

I set the bound from the **mechanism** rather than the observed miss, since a bound tuned to a
measurement on one GPU is the thing that goes red on the next one: TF32 carries a 10-bit mantissa, so
~1e-3 at the activations' unit scale. Your worst-over-five-seeds of 2.8e-4 then reads as a 3.6x
margin rather than as the number the bound was fitted to. Same shape as the existing precedent at
`tests/augmentation/_3d/test_random_rotation_3d.py:102`. The comment says what the assertion pins on
CUDA and why, so the next reader does not have to rediscover cuDNN TF32.

I cannot execute the CUDA arm here, so **that branch is unverified on my side** — the CPU arm passes
and the diff is a two-arm `if`. Worth a re-run on your box.

**The changelog claim — you are right, and it was an implication I should not have left.** It said
`float32`/`float64` are bitwise unchanged and stopped there, which invites the reading that half is
too. It now states that half output moves by 0.25–2.25 eps, that it moves *toward* the `float64`
reference rather than merely becoming finite (`SOSNet` `float16`: 2.15e-03 → 2.28e-04 maximum
absolute error, nothing worse), and that half descriptors are therefore not comparable bit-for-bit
across this release. Your accuracy table is a better argument for the lift than "it stops the NaN"
and it belongs in the record.

**On the rest of your run**, three things I am taking as settled rather than re-litigating: the
defect reaches CUDA (so the NaN is not MPS-only and my #4224 write-up was right to keep it
device-independent), `float32`/`float64` are bitwise unchanged there, and the gate stays `dtype`-only
— your 12-configuration table has no half row outside the 0.94x–1.03x float32 control band, so the
`device.type == "cpu"` term I offered would buy nothing and cost a branch. Dropping that suggestion.

The 8 latent CUDA half failures this retires were not something I could have seen from here; good to
have them counted.

ducha-aiki:
## Review

I rebased this branch onto `main` (only `CHANGELOG.md` conflicted), then ran it on CUDA and re-ran the
Linux manifest legs the description lists as CI's.

**The library change is right.** `kornia/feature/hynet.py` and `kornia/feature/sosnet.py` have no functional
defect I can find, the `float32`/`float64` bitwise-invariance claim holds on CUDA as well as CPU and MPS,
the 11 manifest deletions replay exactly on Linux, and the fix quietly closes **8 untracked CUDA
half-precision failures** on top of them.

Everything below is on the test side. Two items are worth fixing before merge: a new test that fails on
CUDA (#1), and a tolerance change that turns a currently-green test into one that cannot fail (#2).

### Merge conflict

Rebased onto `ca1f2f7b`. The only conflict was `CHANGELOG.md` — both sides prepend to `### Bug fixes`
(`#4174` from `main`, `#4225` here). Kept both, this entry on top, per the file's newest-first ordering.
Verified the rebase preserved every non-`CHANGELOG` hunk byte for byte.

## What I ran

**CUDA — RTX 4090, torch 2.14.0+cu130, Python 3.11**

| run | branch | `origin/main` |
| --- | --- | --- |
| 4 touched files, `float32`+`float64` | 77 passed, 6 skipped | — |
| 4 touched files, `float16`+`bfloat16`, `--isolate-half-precision` | **74 passed**, 6 skipped, 3 errors | **8 failed**, 54 passed, 3 errors |
| full `tests/feature`, `float32` | 2 failed, 659 passed | 2 failed, 653 passed (same two) |
| full `tests/feature`, `float16`+`bfloat16`, isolated (32 min) | 2 failed, 904 passed, 157 skipped, 38 errors | — |

The errors are all `AssertionError: previous item was not torn down properly`, one per test file, and
reproduce identically on `main` — an `--isolate-half-precision` teardown quirk, not this change. The 2
full-suite `float32` failures (`TestLAFAffNetShapeEstimator::test_toy`,
`TestOrientationHalfPrecisionIsFinite::test_half_precision_matches_float32`) are identical on `main`.

The 2 full-suite **half** failures are also not this PR:

- `TestOrientationHalfPrecisionIsFinite::test_half_precision_matches_float32[cuda]` — fails on `main` too.
- `TestMatchSMNN::test_jit[cuda-bfloat16-*]` — flaky on both trees. `test_jit` draws `torch.rand(5, 8)` /
  `torch.rand(7, 8)` **unseeded**, and with 8 bfloat16 mantissa bits the pairwise distances tie often
  enough that eager and scripted `argmin` disagree. Sampling the node in isolation: `main` failed 4 of 10
  runs, this branch with `hynet.py`/`sosnet.py` reverted to `main` failed 1 of 8. Same #4221 tie-instability
  class, in a file this PR does not touch — worth its own issue.

**All 8 CUDA-half failures on `main` are exactly the tests this PR fixes**, and nothing was tracking them —
the manifests are `cpu_float16`/`cpu_bfloat16` only. Worth saying in the description; it is a bigger result
than the PR claims.

**A/B numerics on CUDA**, 36 configurations (both models plus `HyNet(is_bias=False)`, four dtypes,
random / zero / constant patches), `main` vs branch:

- `float32` and `float64`: **bitwise identical** in all 12 combinations.
- `float16` degenerate: three configurations go from `256/256` NaN to finite `0.0884` — what
  `float32`/`float64` return.
- Non-degenerate half output shifts by ≤1–2 ULP; measured against a `float64` reference the accuracy is a
  wash (mostly identical, `SOSNet float16` random better at 1.5e-3 → 2.8e-4, two `HyNet bfloat16` cases
  worse by under one bfloat16 ULP).

**Manifest contract.** Complete mode still cannot run here — the header pins `pytorch: 2.9.1`, this box has
2.14.0 — but the manifest is stamped `Linux`, so focus mode does. Restoring the pre-PR manifests and
replaying `tests/feature`: `cpu-float16` gives exactly **5** strict-XPASS failures, the 5 deleted lines
(683 passed); `cpu-bfloat16` gives exactly **6**, the 6 deleted lines (688 passed). No more, no fewer, on a
different torch than the numbers were taken on. With the branch's manifests: 688 passed / 0 failed
(`float16`), 694 passed / 0 failed (`bfloat16`).

**Other.** `pixi run lint` clean; `ty check kornia` unchanged (the one pre-existing `lightglue.py`
`custom_fwd` deprecation); doctests for both modules pass; `torch._dynamo.explain` gives 1 graph / 0 breaks
for both models on all three dtypes, so `export-support.rst:4258,4303` stay accurate; legacy ONNX export
works for both in `float32` and `float16`.

## Findings

### 1. The new `test_dynamo` tests fail on CUDA float32

```
KORNIA_TEST_OPTIMIZER=inductor KORNIA_TEST_DEVICE=cuda KORNIA_TEST_DTYPE=float32
FAILED tests/feature/test_hynet.py::TestHyNet::test_dynamo[cuda-float32-inductor]
FAILED tests/feature/test_sosnet.py::TestSOSNet::test_dynamo[cuda-float32-inductor]
  Greatest absolute difference: 0.000226 (up to 1e-05 allowed)
```

The cause is cuDNN TF32, not the dtype branch — `float32` does not enter the new code path at all:

```
allow_tf32=True   HyNet  max|eager - compiled| = 2.837e-04   (float32 atol=1e-5)
allow_tf32=True   SOSNet max|eager - compiled| = 1.902e-04
allow_tf32=False  HyNet  max|eager - compiled| = 5.737e-07
allow_tf32=False  SOSNet max|eager - compiled| = 4.023e-07
```

`conftest.py:550` drives only `set_float32_matmul_precision`; `torch.backends.cudnn.allow_tf32` keeps
PyTorch's `True` default, so eager and inductor pick different conv algorithms and a seven-layer conv stack
drifts ~2e-4. CI will not catch this — the `dynamo` job in `pr_test_cpu.yml:160-170` is `ubuntu-latest`,
CPU only, and the tests do pass there (8 passed with `KORNIA_TEST_OPTIMIZER=inductor` on CPU). But they
fail on any CUDA box.

This is not new ground: `TestXFeatModel::test_dynamo[cuda-float32-inductor]` already fails the same way on
`main`, and the repo already has the remedy — the autouse fixture at `tests/color/test_yuv.py:158-182`,
whose docstring says outright *"the repo-wide fix is for `pytest_sessionstart` to set `cudnn.allow_tf32`
from the same option — until then this is the only module whose derived bounds depend on it."* This PR
would make it the second and third. Either adopt that fixture here, or give the CUDA `float32` comparison
an explicit tolerance.

(Side note for whoever picks up the general fix: these tests also need the *env var* —
`--optimizer=inductor` alone leaves them deselected by `conftest.py:355`, which reads
`KORNIA_TEST_OPTIMIZER` directly.)

### 2. `TestCornerHarris::test_corners_batch` becomes unfalsifiable in bfloat16

The expected tensor is divided by `10.0`, so its largest element is `6.9e-3`, and the repository's default
bfloat16 tolerance is `atol = rtol = 7.8e-3`. The bound is larger than the signal. I checked all four
tests by substituting an all-zeros response:

| test | dtype | max\|expected\| | atol+rtol·max | zeros pass? |
| `TestCornerHarris` | float16 | 6.88e-3 | 1.01e-3 | no |
| `TestCornerHarris` | **bfloat16** | 6.90e-3 | 7.85e-3 | **yes — vacuous** |
| `TestCornerGFTT` | float16 / bfloat16 | 6.69e-2 | 1.07e-3 / 8.32e-3 | no |
| `TestBlobHessian` | float16 / bfloat16 | 1.91e-1 | 1.19e-3 / 9.29e-3 | no |
| `TestBlobDoGSingle` | float16 / bfloat16 | 2.76e-1 | 1.28e-3 / 9.95e-3 | no |

Only Harris is affected — the other three have expected values 10–40x larger. But Harris is also the one
test here that **was passing**, so the change trades a green assertion for one that cannot fail. The
description names this deliberately ("a latent manifest line"), and the 91% margin is a fair thing to worry
about, but the default table is the wrong instrument at this signal level. An explicit bound scaled to the
signal — `atol=3e-4` gives ~3x headroom over the observed 9.4e-5 — keeps the test meaningful.

The comment is also self-contradictory about it: "*atol=1e-4 sits below that floor, so no implementation
can meet it -- bfloat16 passes today at 91% of it.*" It does meet it today.

### 3. Two of the four dtype guards loosen cases that were already green

I ran `origin/main`'s copy of the file at the original `atol=1e-4, rtol=1e-4` on this torch, which
reproduces the manifests exactly:

| test | float16 | bfloat16 |
| --- | --- | --- |
| `TestCornerHarris::test_corners_batch` | passes | passes |
| `TestCornerGFTT::test_corners_batch` | **fails** | **fails** |
| `TestBlobHessian::test_blobs_batch` | passes | **fails** |
| `TestBlobDoGSingle::test_blobs_batch` | **fails** | **fails** |

So `if dtype in (torch.float16, torch.bfloat16)` relaxes three green cases: Harris in both dtypes (see #2)
and `TestBlobHessian` in float16, which goes 10x looser for a failure that does not exist. `TestBlobHessian`
wants `if dtype is torch.bfloat16`. `TestCornerGFTT` and `TestBlobDoGSingle` earn their two-dtype guards.

While there: GFTT float16 is the thinnest surviving margin at **64.6%** of its allowance (identical on CPU
and CUDA; harris 1.8%, hessian 1.3%), i.e. ~1.5x headroom, where one extra ULP at 1.0 flips it. The ALIKED
fix in this same diff derives its bound from the dtype — `atol = max(1e-4, 2 * torch.finfo(dtype).eps)`,
`test_aliked.py:178` — and the response comments argue for exactly that floor ("one ULP at 1.0") before
falling back on the default table. The same expression gives float16 `atol` 1.95e-3 and drops GFTT to ~33%.

### 4. bfloat16 belongs in the guard, but not for the reason the comment gives

`kornia/feature/hynet.py:296-303`, `kornia/feature/sosnet.py:104-111`:

> `1e-10` is not representable in float16 -- it flushes to `0.0` -- so the guard is inert there and the
> division returns NaN (kornia#4224). Half precision therefore takes this step in float32...

attached to a branch firing on `(torch.float16, torch.bfloat16)`. For bfloat16 that reason is false:

```
torch.float16    1e-10 -> 0.000000e+00  flushed=True    tiny=6.104e-05
torch.bfloat16   1e-10 -> 1.000444e-10  flushed=False   tiny=1.175e-38
CUDA LocalResponseNorm(zeros + 1e-10):   float16 -> nan,   bfloat16 -> 0.0884
```

bfloat16 keeps float32's exponent range, so the guard is live and defect (b) never existed there. The A/B
shows every bfloat16 degenerate case **bitwise identical** between `main` and this branch, and running the
new pin against unfixed `kornia` confirms it:

```
CUDA, -k degenerate:  3 failed, 9 passed   -- all three float16, every bfloat16 parametrization passes
CPU,  -k degenerate: 12 failed             -- NotImplementedError "avg_pool3d_out_frame" ... 'Half' / 'BFloat16'
```

bfloat16 still belongs in the branch: it hits the CPU `avg_pool3d` kernel gap exactly like float16, and one
path across dtypes beats a device-dependent one. That is just the reason to state.

The comment also claims this is "the same treatment `siftdesc.py` gives its own 1e-10 guard". It is not —
`_gradient_magnitude_orientation` (`siftdesc.py:61`) and `_rootsift` (`:268`) gate on
`dtype == torch.float16` only, and the docstring says why: *"bfloat16 has float32's exponent range, so both
the guard and the squares are representable there and it takes the same expression as float32 and
float64."* A reader trimming this branch to match the cited precedent would reintroduce the CPU
`NotImplementedError`. Suggested rewording:

> `1e-10` is not representable in float16 — it flushes to `0.0` — so the guard would be inert and the
> division would return NaN (kornia#4224). bfloat16 keeps float32's exponent range and does not have that
> problem, but it shares the other one: `F.local_response_norm` routes a 4-D input through `avg_pool3d`,
> which has no CPU kernel for either half dtype. Both therefore take this step in float32 and cast back;
> float32 and float64 take the original expression, bit for bit.

### 5. CHANGELOG: "every `Conv2d` in these models has `bias=False`" is false for HyNet

`HyNet.layer1`–`layer6` build their convs with `bias=is_bias`, and `is_bias` defaults to **`True`**; only
`layer7`'s conv is unconditionally `bias=False`. The entry partly walks this back three lines later
("`HyNet` reached this only with `is_bias=False`"), but the causal clause as written is wrong. It is
correct for `SOSNet`, and the test comment in `test_sosnet.py:49` scopes it correctly.

Also: the entry cites `(#4225)` twice, once mid-paragraph and once at the end. And "the descriptors came
back all-NaN for any patch the network maps to exactly zero" is true in float16, not bfloat16 (#4).

### 6. Two `#4224` references survive in `kornia/`

`grep -rnE "#4224|issues/4224" kornia/ tests/` → `hynet.py:299`, `sosnet.py:107` (plus two in `tests/`).
AGENTS.md: *"A surviving hit in `kornia/` means the change is incomplete."* These are historical rationale
rather than a documented open defect, so they are defensible — but they are written in the **present
tense** ("the guard *is* inert there and the division *returns* NaN"), which reads as live behaviour of the
code immediately below. The past/conditional rewording in #4 settles both.

### 7. The assertion message can contradict the assertion

`tests/feature/test_hynet.py:60`, `tests/feature/test_sosnet.py:56`:

```python
assert torch.isfinite(descriptors).all(), f"{int(descriptors.isnan().sum())} non-finite descriptor entries"
```

Asserts on `isfinite`, counts `isnan`. An `±inf` failure reports "0 non-finite descriptor entries". Use
`int((~torch.isfinite(descriptors)).sum())`.

### 8. `test_dynamo`'s comment asserts something the test does not pin

`tests/feature/test_hynet.py:67`, `tests/feature/test_sosnet.py:63`: *"The graph-break count is checked
separately: `torch._dynamo.explain` reports 1 graph and 0 breaks."* I confirmed that is true today for both
models on all three dtypes — but a comment is not a pin, the repo has `test_dynamo_fullgraph`
(`tests/feature/test_laf.py:351,393,449`), and `export-support.rst` publishes "0 breaks" for both models on
the strength of it.

### 9. The non-batch `test_corners` are left alone for a different reason than stated

The description says they "use `rtol=1e-3`, sit at ≤55%, and are left alone". Both build `inp` and
`expected` with `dtype=dtype` and then call `.float()` (`test_responces_local_features.py:53,71` and
`:175,193`), so the dtype fixture is discarded and they always run in float32 — which is why they never
entered a manifest. Leaving them alone is right; the reason is not the tolerance. Pre-existing dead
parametrization, worth a follow-up rather than anything here.

### 10. Minor

- `docs/source/get-started/precision.rst`'s `kornia.feature` row lists "SIFT, HardNet, DISK, DeDoDe" as the
  descriptors that work at half precision. HyNet and SOSNet now qualify.
- The 7-line comment plus 4-line branch is duplicated verbatim across both models. `siftdesc.py` factored
  its equivalent into `_rootsift` / `_gradient_magnitude_orientation`; a small shared helper would keep the
  two from drifting.

### Two things worth their own issues (both pre-existing)

- `--isolate-half-precision` leaves pytest's setup stack inconsistent: the first device-only test collected
  after an isolated one errors with `AssertionError: previous item was not torn down properly`, one per
  test file — 38 of them in a full `tests/feature` CUDA half run. Reproducible on `main`.
- `TestMatchSMNN::test_jit` seeds nothing and compares eager against scripted `argmin` over bfloat16
  distances; on CUDA it fails a few runs in ten on `main`. `TestMatchFGINN::test_jit` draws its descriptors
  the same unseeded way, though I did not sample it.

---

Rebase and all runs above were done on Linux/x86_64, Python 3.11, torch 2.14.0+cu130, RTX 4090.

ducha-aiki (commented):
[P1] Avoid adding a nested tests/feature/conftest.py. Pytest imports it as the top-level module conftest, which shadows the repository-root conftest.py. When the suite then collects tests/test_conftest.py, its import of _is_subprocess_isolated_test and skip_half_precision_on_cuda resolves to this new file and raises ImportError. This is reproducible with:

    .venv/bin/python -m pytest tests/test_conftest.py tests/feature/test_hynet.py --collect-only -q

The current CI matrix is failing at collection for the same reason before any tests run. Please move this fixture to the root conftest, which would also match the proposed repo-wide --tf32 behavior, or otherwise avoid introducing a second top-level module named conftest.

The library changes and focused tests pass when collected without tests/test_conftest.py: 127 passed across CPU float16/bfloat16/float32/float64, plus 16 inductor/fullgraph tests.

ducha-aiki:
Fixed the suite-wide collection blocker in 54e54ff6.

The cudnn_tf32_follows_option fixture now lives in the repository-root conftest.py, and the nested tests/feature/conftest.py has been removed. This preserves the fixture behavior for the HyNet and SOSNet compile tests while preventing it from shadowing the root conftest during collection.

Verification:
- Mixed root/feature collection: 25 tests collected, no errors.
- Root-conftest plus touched feature tests across CPU float16, bfloat16, float32, and float64: 142 passed, 10 skipped.
- Inductor/fullgraph HyNet and SOSNet slice across all four CPU dtypes: 16 passed.
- Full pre-commit suite: passed.
- Independent code review: no findings.

PR #4226, opened 2026-09-04
docs: improve search metadata and canonical links

ducha-aiki:

Improves the documentation search presentation and URL consistency:

- gives the docs homepage one semantic H1, a descriptive browser/social title, and a clean root canonical
- adds unique descriptions to application and model pages, with safe title-aware fallbacks elsewhere
- canonicalizes the obsolete highlighted-features page, forwards users with a relative redirect, and excludes the stub from documentation search
- noindexes generated search/index/error utility pages without mixing that signal into redirect canonicalization
- updates repository tutorial, Blog, and package homepage links to their canonical HTTPS kornia.org URLs

Follow-up to a technical SEO audit of kornia.org and the Read the Docs site.

## What did you check?

```text
$ /Users/oldufo/dev/kornia/.venv/bin/python -m sphinx -W -b html docs/source /tmp/kornia-seo-review-build.LOTQnf
build succeeded

$ pixi run pre-commit-all
all checks passed
```

Also verified across the generated output that each page has one description, canonical, and Open Graph URL; the homepage has exactly one H1 and a canonical ending at `/en/latest/`; the retired page has a relative refresh plus replacement canonical and cannot surface in Sphinx search; and only the generated 404/search/index pages carry `noindex, follow`.

## How did AI help?

Codex audited the live site and repository, implemented the metadata and link changes, incorporated two rounds of review feedback, rendered and inspected the generated HTML, ran repository validation, and completed a separate final code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Documentation pages now provide search-specific titles and descriptions, canonical tutorial links, and a
+  canonicalized redirect from the retired highlighted-features page. (#4226)
+
 * `kornia.feature.laf_is_filled(laf)` returns the `(B, N)` mask of the slots a detection filled, the complement
   of a detector's zero-LAF padding. Fixed-shape detectors pad the slots no detection filled with an all-zero LAF,
   and those slots must be dropped before matching -- their descriptors are identical to one another, so they match
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ English | [简体中文](README_zh-CN.md)
 <a href="https://colab.sandbox.google.com/github/kornia/tutorials/blob/master/nbs/hello_world_tutorial.ipynb">Try it Now</a> •
 <a href="https://www.kornia.org/tutorials/">Tutorials</a> •
 <a href="https://github.com/kornia/kornia-examples">Examples</a> •
-<a href="https://kornia.github.io//kornia-blog">Blog</a> •
+<a href="https://www.kornia.org/kornia-blog/">Blog</a> •
 <a href="https://discord.gg/HfnywwpBnD">Community</a>
 
 [![PyPI version](https://badge.fury.io/py/kornia.svg)](https://pypi.org/project/kornia)
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ English | [简体中文](README_zh-CN.md)
 <!-- prettier-ignore -->
 <a href="https://kornia.readthedocs.io">Docs</a> •
 <a href="https://colab.sandbox.google.com/github/kornia/tutorials/blob/master/nbs/hello_world_tutorial.ipynb">Try it Now</a> •
-<a href="https://kornia.github.io/tutorials/">Tutorials</a> •
+<a href="https://www.kornia.org/tutorials/">Tutorials</a> •
 <a href="https://github.com/kornia/kornia-examples">Examples</a> •
 <a href="https://kornia.github.io//kornia-blog">Blog</a> •
 <a href="https://discord.gg/HfnywwpBnD">Community</a>
--- a/README_zh-CN.md
+++ b/README_zh-CN.md
@@ -14,7 +14,7 @@
 <a href="https://colab.sandbox.google.com/github/kornia/tutorials/blob/master/nbs/hello_world_tutorial.ipynb">Try it Now</a> •
 <a href="https://www.kornia.org/tutorials/">Tutorials</a> •
 <a href="https://github.com/kornia/kornia-examples">Examples</a> •
-<a href="https://kornia.github.io//kornia-blog">Blog</a> •
+<a href="https://www.kornia.org/kornia-blog/">Blog</a> •
 <a href="https://discord.gg/HfnywwpBnD">Community</a>
 
 [![PyPI version](https://badge.fury.io/py/kornia.svg)](https://pypi.org/project/kornia)
--- a/README_zh-CN.md
+++ b/README_zh-CN.md
@@ -12,7 +12,7 @@
 <!-- prettier-ignore -->
 <a href="https://kornia.readthedocs.io">Docs</a> •
 <a href="https://colab.sandbox.google.com/github/kornia/tutorials/blob/master/nbs/hello_world_tutorial.ipynb">Try it Now</a> •
-<a href="https://kornia.github.io/tutorials/">Tutorials</a> •
+<a href="https://www.kornia.org/tutorials/">Tutorials</a> •
 <a href="https://github.com/kornia/kornia-examples">Examples</a> •
 <a href="https://kornia.github.io//kornia-blog">Blog</a> •
 <a href="https://discord.gg/HfnywwpBnD">Community</a>
@@ -114,7 +114,7 @@
 
 ## 例子
 
-可以尝试通过这些 [教程](https://kornia.github.io/tutorials/) 来学习和使用这个库。
+可以尝试通过这些 [教程](https://www.kornia.org/tutorials/) 来学习和使用这个库。
 
 <div align="center">
   <a href="https://colab.sandbox.google.com/github/kornia/tutorials/blob/master/nbs/hello_world_tutorial.ipynb" target="_blank">
--- a/docs/source/_extra/llms-full.txt
+++ b/docs/source/_extra/llms-full.txt
@@ -195,4 +195,4 @@ Durable findings from earlier CUDA runs (2026-08-07, see benchmarks/README.md):
 - Docs: <https://kornia.readthedocs.io/en/latest/>
 - Machine-readable index: <https://kornia.readthedocs.io/en/latest/llms.txt>
 - Source: <https://github.com/kornia/kornia>
-- Tutorials: <https://kornia.github.io/tutorials/>
+- Tutorials: <https://www.kornia.org/tutorials/>
--- a/docs/source/_extra/llms.txt
+++ b/docs/source/_extra/llms.txt
@@ -19,7 +19,7 @@ For the full self-contained reference with runnable examples, fetch [llms-full.t
 - [Conventions & Pitfalls](https://kornia.readthedocs.io/en/latest/get-started/conventions.html): the canonical conventions reference — read before generating Kornia code
 - [API Stability Policy](https://kornia.readthedocs.io/en/latest/get-started/stability.html): which modules are stable, and how deprecation works
 - [Installation](https://kornia.readthedocs.io/en/latest/get-started/installation.html): pip/conda/source installation
-- [Tutorials](https://kornia.github.io/tutorials/): runnable end-to-end examples
+- [Tutorials](https://www.kornia.org/tutorials/): runnable end-to-end examples
 
 ## Core API
 
--- a/docs/source/applications/face_detection.rst
+++ b/docs/source/applications/face_detection.rst
@@ -1,6 +1,9 @@
 Face Detection
 ==============
 
+.. meta::
+   :description: Detect multiple faces in images and video with Kornia's real-time PyTorch FaceDetector and YuNet model.
+
 .. image:: https://github.com/ShiqiYu/libfacedetection/raw/master/images/cnnresult.png
    :align: right
    :width: 20%
@@ -38,4 +41,4 @@ Using our API you can easily detect faces in images as shown below:
     for det in dets:
         print(det.score, det.top_left, det.bottom_right)
 
-Play with the detector yourself and generate new images with this `tutorial <https://kornia.github.io/tutorials/nbs/face_detection.html>`_.
+Play with the detector yourself and generate new images with this `tutorial <https://www.kornia.org/tutorials/nbs/face_detection.html>`_.
--- a/docs/source/applications/image_augmentations.rst
+++ b/docs/source/applications/image_augmentations.rst
@@ -1,6 +1,9 @@
 Image Augmentation
 ==================
 
+.. meta::
+   :description: Build differentiable, GPU-accelerated image augmentation pipelines for PyTorch with Kornia.
+
 Image Augmentation is a data augmentation method that generates more training data
 from the existing training samples. Image Augmentation is especially useful in domains
 where training data is limited or expensive to obtain like in biomedical applications.
--- a/docs/source/applications/image_denoising.rst
+++ b/docs/source/applications/image_denoising.rst
@@ -1,6 +1,9 @@
 Image Denoising
 ===============
 
+.. meta::
+   :description: Remove image noise with differentiable total-variation denoising and Kornia's PyTorch vision operators.
+
 Image denoising removes noise from an image while preserving its structure. Because every Kornia operator is
 differentiable, a classic variational approach such as total-variation denoising can be written as a few lines of
 optimization with :func:`kornia.losses.total_variation`:
@@ -20,4 +23,4 @@ optimization with :func:`kornia.losses.total_variation`:
         loss.backward()
         optimizer.step()
 
-Follow the full walk-through in the `total variation denoising tutorial <https://kornia.github.io/tutorials/nbs/total_variation_denoising.html>`_.
+Follow the full walk-through in the `total variation denoising tutorial <https://www.kornia.org/tutorials/nbs/total_variation_denoising.html>`_.
--- a/docs/source/applications/image_matching.rst
+++ b/docs/source/applications/image_matching.rst
@@ -1,6 +1,9 @@
 Image Matching
 ==============
 
+.. meta::
+   :description: Match images with Kornia's PyTorch local features, LoFTR, geometric estimation and RANSAC tools.
+
 Image matching is a process of finding pixel and region correspondences between two images of the same scene.
 Such correspondences are useful for 3D reconstruction of the scene and relative camera pose estimation.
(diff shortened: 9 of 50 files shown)


ducha-aiki:
## Review of #4226 — `docs: improve search metadata and canonical links`

Reviewed at `98fe8ccf`. Method: clean `-W` HTML builds of `origin/main` (`25b2ca72`) and the PR head in separate worktrees, then an A/B diff of the two output trees (`<title>`, `og:*`, `<meta name="description">`, `<link rel="canonical">`, H1 counts, `objects.inv`), plus live HTTP checks of every URL the PR introduces and of the PR's own RTD preview.

**Verdict: changes requested.** The direction is right and the homepage fix is worth landing on its own, but the new templated description fallback ships broken copy on ~9 pages, and the retired-page treatment sends two contradictory signals plus a hard-coded production URL.

---

### What I verified as correct

* **The host swap is right.** `kornia.org` and `kornia.github.io` both `301` to `www.kornia.org`, which serves `200`. All 20 distinct `www.kornia.org` tutorial URLs the PR introduces return `200` with no redirect hop.
* **`kornia/` is a pure URL substitution** — 31 added / 31 removed, and filtering the diff for anything that is not the tutorial-URL swap leaves nothing.
* **No inventory or anchor breakage.** `objects.inv` differs in exactly one entry: `std:doc index` display name `Features` → `Kornia: Differentiable Computer Vision for PyTorch`. The HTML file set is identical, and the canonical-link diff over all 151 pages touches only `get-started/highlights.html`.
* **The H1 goal is met.** `index.html` was the only page in the whole build with two H1s on `main`; it now has one, and no other page regressed.
* **The homepage `<title>` fix is the real win here.** On `main` it renders `<title>Features — Kornia</title>` and `og:title="Features"`, because the lone `Features` section was being promoted to the document title. It is now `Differentiable Computer Vision for PyTorch — Kornia`.
* **The hero `h1` → `p` swap is visually inert.** `--pst-font-family-heading` resolves to the same stack as `--pst-font-family-base`, `--pst-color-heading` resolves to `--pst-color-text-base`, and sphinx-design's only paragraph reset is `.sd-sphinx-override p{margin-top:0}`, which `.kornia-hero__title`'s `margin: 0 0 1.5rem` already matches.
* **`context["pageurl"]` does land.** `priority=900` puts the handler after pydata's `_fix_canonical_url`, and the built `highlights.html` carries the replacement canonical.

---

### P1 — the templated fallback description ships broken copy

`docs/source/conf.py:628-631`

```python
description = (
    f"Learn about {page_title} in Kornia, the differentiable computer vision library for PyTorch, "
    "including its APIs, behavior and usage."
)
```

`page_title` is `html.unescape(context.get("title") or project)`, so on generated pages — where `context["title"]` is empty — it collapses to `project` and the sentence becomes *"Learn about Kornia in Kornia"*. And the `"including its APIs, behavior and usage"` tail is asserted on pages that document no API. Exact strings from the clean build of this head:

| Page | Emitted description |
|---|---|
| `genindex.html`, `search.html`, `py-modindex.html` | `Learn about Kornia in Kornia, the differentiable computer vision library for PyTorch, including its APIs, behavior and usage.` |
| `get-started/about.html` (title *Citing Kornia*) | `Learn about Citing Kornia in Kornia, …` |
| `community/chinese.html` (title *Kornia 社区*) | `Learn about Kornia 社区 in Kornia, the differentiable computer vision library for PyTorch, …` |
| `404.html` (title *Page not found*) | `Learn about Page not found in Kornia, … including its APIs, behavior and usage.` |
| `community/bibliography.html` | `Learn about Bibliography in Kornia, … including its APIs, behavior and usage.` |
| `community/faqs.html` | `Learn about Frequently Asked Questions in Kornia, … including its APIs, behavior and usage.` |

61 of the 151 built pages take this branch, so most of them read fine (`Learn about HSV in Kornia, …`); these are the ones that do not. A description Google will read as boilerplate or nonsense is worse than none — it gets rewritten anyway, and the 404 page should not be advertising an API at all.

Minimum fix: when `context.get("title")` is falsy, or on `404`/`search`/`genindex`/`py-modindex`, fall back to the site description instead of the template. That also resolves the next point.

### P1 — `_DEFAULT_DESCRIPTION` is now dead, and its comment is false

`docs/source/conf.py:595` still says:

```python
# Used when a page does not declare its own ``.. meta:: :name: description``.
```

After this change it is reachable only when `pagename == master_doc` **and** `index.rst` declares no `.. meta::` description — and `index.rst` always declares one. Grepping the built tree confirms it: the string appears on **0 of 151** pages on this head, against 61 on `main`. Either wire it back in as the empty-title fallback above (my preference — one edit fixes both) or delete it and its comment.

### P2 — `noindex` and a cross-page `rel=canonical` are contradictory signals

`get-started/highlights.html` now emits both:

```html
<meta name="robots" content="noindex, follow" />
<link rel="canonical" href="https://kornia.readthedocs.io/en/latest/get-started/introduction.html" />
```

Google's canonicalization guide is explicit: *"We don't recommend using `noindex` to prevent selection of a canonical page within a single site, because it will completely block the page from Search."* A canonical asserts the two URLs are the same page; a `noindex` on that page is then an instruction about a group that includes the target. Pick one. Since the canonical and the refresh already point at the replacement, the `noindex` is the line to drop — the canonical alone is the documented way to consolidate.

### P2 — the meta refresh hard-codes the production URL, so every non-production build bounces off-site

`docs/source/conf.py:649-653` builds the refresh target from `html_baseurl`, which is pinned to `https://kornia.readthedocs.io/en/latest/`. Demonstrated on this PR's own preview:

```
$ curl -s https://kornia--4226.org.readthedocs.build/en/4226/get-started/highlights.html | grep refresh
<meta http-equiv="refresh" content="0; url=https://kornia.readthedocs.io/en/latest/get-started/introduction.html" />
```

So a reviewer opening that page in the preview is thrown to the live site; the same happens for a local `sphinx -b html` output and for any versioned RTD build, which jumps the reader from `/en/v0.8.x/` to `/en/latest/`. Use `app.builder.get_relative_uri(pagename, redirect_target)` for the `http-equiv` target and keep the absolute URL only for the canonical.

Worth noting the repo already has a client-side redirect for the split-index anchors (`docs/source/_static/js/custom.js:20`) and it deliberately uses `location.replace(...)`. If a JS hop is acceptable there, reusing that shape here keeps one redirect mechanism instead of two — and an RTD dashboard redirect (a real 301) would be better than either, since Google lists redirects first and meta refresh not at all.

### P3 — the landing-page CSS is keyed to an auto-generated section id

`docs/source/index.rst:19` and `:24`:

```css
section#kornia-differentiable-computer-vision-for-pytorch > h1:first-child { … }
section#kornia-differentiable-computer-vision-for-pytorch > section { … }
```

Reword the H1 and docutils regenerates the slug: the H1 silently stops being hidden — a full-size heading appears above the hero — and every hairline separator disappears. No build error, `-W` included. The built markup is `<article class="bd-article"><section id=…>`, so `.bd-article > section > h1:first-child` and `.bd-article > section > section` are exactly equivalent and survive a rewording.

### P3 — the `Features` heading is demoted, so "preserves the existing layout" is not quite accurate

Wrapping the page in a document-level section pushes `Features` down a level: `main` renders `<h1>Features`, this head renders `<h2>Features`, i.e. `--pst-font-size-h1: 2.625rem` → `--pst-font-size-h2: 2.125rem`. That is the semantically correct outcome and I would keep it, but it is a visible change and the PR description says the homepage layout is preserved.

### Nits

* **`og:title` on the homepage loses the brand.** `<title>` is `Differentiable Computer Vision for PyTorch — Kornia`, but `og:title` is `Differentiable Computer Vision for PyTorch` — the `pagename != master_doc` guard skips the ` - {project}` suffix, so shared cards no longer name Kornia. (Still an improvement over `main`, where it was `Features`.)
* **The homepage self-canonical is `…/en/latest/index.html`, not `…/en/latest/`.** Pre-existing, inherited from Sphinx's `pageurl` — but this PR is the canonicalization pass, so it is the natural place to fix it.
* **The `Blog` link in both READMEs is broken and in scope.** `README.md:15` and `README_zh-CN.md:17` carry `https://kornia.github.io//kornia-blog` — note the double slash, which is preserved through the redirect and lands the reader on `https://www.kornia.org/` rather than the blog. The working URL is `https://www.kornia.org/kornia-blog/`. It sits on the same nav line as the `Tutorials` link this PR just canonicalized.
* **`pyproject.toml:35`** still has `Homepage = "https://kornia.github.io/"`, which `301`s to `https://www.kornia.org/`. That is the link PyPI shows.

---

Once the fallback copy and the `noindex`/canonical conflict are settled I am happy to approve; the homepage title fix alone is a clear net win.

ducha-aiki:
### Second pass — three more, all verified against the same two builds

A second review pass over `98fe8ccf` turned up three things my first comment missed. I re-measured each one against the `25b2ca72` / `98fe8ccf` build pair before writing this.

**P2 — the retired page is `noindex` for Google but still in the site's own search index.** `searchindex.js` from this head still carries it:

```
docnames[94] = 'get-started/highlights'   titles[94] = 'Highlighted Features'
```

So a reader who types "highlighted features" into the docs search box gets a hit, clicks it, and is bounced — and per my earlier point about the absolute refresh URL, bounced out of whatever version they were browsing. `:orphan:` keeps it out of the toctree but not out of `searchindex.js`; it needs `:nosearch:` alongside. No page in `docs/source` uses `:nosearch:` today, so this would be the first.

**P2 — the meta refresh drops the URL fragment, which is the main reason anyone still has these links.** `highlights.rst` itself says the module overview and the accessible-AI-models section moved into `introduction`, and the built `get-started/introduction.html` does carry `id="accessible-ai-models"`. But `<meta http-equiv="refresh" content="0; url=…/introduction.html">` sends every visitor to the top of a long page, so `highlights.html#accessible-ai-models` loses exactly the part the reader asked for.

This sharpens the point I made about `docs/source/_static/js/custom.js:20` — that mechanism doesn't just use `location.replace(...)`, it appends `location.hash`. The repo already has a fragment-preserving redirect driven by `_inject_anchor_redirects`; the new meta refresh is a second, weaker one beside it.

**P3 — `priority=900` pushes the Open Graph tags 8–12 KB deeper into `<head>`.** Measured, base → this head:

| page | `og:title` byte offset (main) | (PR) |
|---|---|---|
| `feature.html` | 1044 | 9273 |
| `augmentation.module.html` | 1112 | 12835 |
| `color.html` | 943 | 6361 |

Cause: all three `html-page-context` handlers append to `context["metatags"]`, so demoting this one to last puts it behind the large `<script type="application/json">` payloads from `_inject_anchor_redirects` and `_inject_navbar_menus`. Unfurlers that cap how much of `<head>` they parse would now miss the card.

One correction to the second-pass note on my side, for the record: **the priority itself is worth keeping.** pydata's `_fix_canonical_url` does return early here — kornia builds with the plain `html` builder, `dirhtml` appears nowhere in `docs/`, `readthedocs.yml` or `pixi.toml` — so the comment's stated justification doesn't bite today. But it is correct defensively: under `dirhtml`, running first would let the theme clobber the `context["pageurl"]` we just set and undo the redirect canonical. The fix for the byte-offset cost is therefore not to drop the priority but to prepend rather than append in `_inject_social_metatags` — `"\n".join(extra) + "\n" + metatags` — which puts the social tags first regardless of handler order.

**Nit, latent —** `page_title` is `render_partial`'s title *HTML*, not plain text. No current title carries inline markup (checked across all 151 built pages), but this PR grows its blast radius from one `og:title` to the description on 83 pages, so a future title written as ``` ``torch.compile`` support ``` would emit escaped `<code class="docutils literal notranslate">` into `<meta name="description">`. A `re.sub(r"<[^>]+>", "", …)` before building `page_title` closes it.

**Related to my P1 —** the natural home for the `404`/`search`/`genindex`/`py-modindex` problem is a `_NOINDEX_PAGES` set applied independently of `_PAGE_REDIRECTS`. Right now `redirect_target` is one switch driving three separate concerns (canonical rewrite, robots, refresh), so a page can't be marked `noindex` without also being given a canonical override and a redirect — and those four pages are precisely the ones that want `noindex` and no description at all. Worth noting `404.html` is served with HTTP 200 on Read the Docs, so "Learn about Page not found in Kornia … including its APIs, behavior and usage." is genuinely crawlable text.

ducha-aiki:
Addressed the review findings in `7d0152af`:

- replaced the unsafe fallback sentence with title-safe descriptions and restored the site default for generated/special pages
- removed `noindex` from the replacement canonical, made the browser redirect relative, and added `:nosearch:` to the retired highlights stub
- separately marked generated 404/search/index utility pages `noindex, follow`
- prepended social metadata, normalized title markup, branded the homepage social title, and canonicalized the homepage to `/en/latest/`
- made the homepage styling selector stable via an explicit target
- fixed both README Blog links and the PyPI homepage URL
- updated the PR description so it no longer claims the Features heading is visually unchanged

A fresh warning-as-error Sphinx build and full pre-commit suite pass. Generated-output assertions confirm the highlights page cannot surface in Sphinx search, its refresh stays relative, its canonical remains absolute, and the noindex/canonical conflict is gone. Fragment-preserving redirects remain best handled by the planned Read the Docs server redirect.

ducha-aiki:
## Re-review at `7d0152af`

Re-ran the same method on the delta: clean `-W` build of `7d0152af`, A/B'd against the `origin/main` (`25b2ca72`) and `98fe8ccf` builds I still had, plus live checks of the new URLs and of the PR's own RTD preview. `-W` build exits 0; `docs (3.11)`, `Check documentation links`, `TOML Format`, `pre-commit.ci` and the RTD build are all green.

**Verdict: approve.** Every finding is addressed, and two of the fixes are better than what I asked for.

### First, a retraction — my `:nosearch:` check was the wrong test

I claimed the retired page was still in `searchindex.js`. I tested membership in `docnames`, which is not what `:nosearch:` changes: Sphinx's `index_page` still feeds a `nosearch` page, just with an empty title and an empty document. Measured properly:

| | `titles[94]` | terms → doc 94 | titleterms → doc 94 |
|---|---|---|---|
| `98fe8ccf` | `'Highlighted Features'` | 29 | 2 (`highlight`, `featur`) |
| `7d0152af` | `''` | 0 | 0 |

So the page genuinely cannot surface in on-site search now. Your claim was right and mine was wrong.

### Verified fixed

* **The description fallback.** All 57 generated descriptions now read cleanly (`Bibliography: documentation for Kornia, …`, `Frequently Asked Questions: documentation for Kornia, …`). "Learn about Kornia in Kornia", "Learn about Page not found in Kornia … including its APIs" and the rest are gone.
* **`_DEFAULT_DESCRIPTION` is live again** on exactly the five pages that should have it — `404`, `search`, `genindex`, `py-modindex`, `get-started/highlights` — so the dead-constant problem is resolved by the same edit rather than papered over.
* **`_NOINDEX_PAGES`.** Each of the four utility pages carries exactly one `<meta name="robots">` — no duplicate against anything the theme emits.
* **The `noindex`/canonical conflict is gone**, and the canonical still points at `introduction.html`.
* **The refresh is relative, confirmed on the live preview** — this was my headline P2:
  ```
  $ curl -s https://kornia--4226.org.readthedocs.build/en/4226/get-started/highlights.html | grep refresh
  <meta http-equiv="refresh" content="0; url=introduction.html" />
  ```
  The reviewer now stays inside the preview instead of being thrown to production `latest`.
* **Homepage canonical is `https://kornia.readthedocs.io/en/latest/`** (200, no redirect hop), and `og:url` matches. Canonical diff over all 151 pages vs `main` touches only those two intended pages.
* **`og:title` is branded again:** `Kornia — Differentiable Computer Vision for PyTorch`.
* **The prepend fix over-delivered.** `og:title` byte offset in `<head>`: `feature.html` 1044 (main) → 9273 (`98fe8ccf`) → **287**; same 287 for `augmentation.module.html`, `color.html` and `index.html`. Better than the pre-PR baseline, not merely restored.
* **The CSS anchor works.** Docutils emits `<span id="kornia-docs-home"></span><h1>` as I hoped it might not — I checked, and `#features` is a direct sibling of that span, so both `+ h1` and `~ section` match. Every page in the build still has exactly one `<h1>`.
* **`https://www.kornia.org/kornia-blog/` and `https://www.kornia.org/` both 200** with no redirect hop.

### Two nits, neither blocking

**The tag-stripping is one step out of order.** `docs/source/conf.py:623`:

```python
page_title = re.sub(r"<[^>]+>", "", html.unescape(context.get("title") or project))
```

`html.unescape` runs first, so a title containing an escaped angle bracket has it turned into a real one and then has the text between it and the next `>` eaten — `Values &lt;0 and &gt;1` would come out as `Values 1`. No current title carries `&lt;`/`&gt;` (checked all 151 built titles), so this is latent exactly like the markup case it was written to fix. Swapping the two — strip tags on the raw HTML, then unescape — closes both.

**The homepage `og:title` is a literal that duplicates `.. title::`.** `conf.py:637` hardcodes `f"{project} — Differentiable Computer Vision for PyTorch"` while `index.rst:4` independently declares `.. title:: Differentiable Computer Vision for PyTorch`. Editing one silently desynchronizes the other — the same coupling you just removed from the CSS selector. `f"{project} — {page_title}"` yields the identical string from the live title. (Minor and related: the homepage uses an em dash while every other page's `og:title` uses ` - `.)

### One note for whoever lands the RTD redirect

You're deferring fragment preservation to a server-side redirect, which I agree is the right call — it's the only form Google documents, and it preserves the fragment for free. Worth flagging that when it lands, `_PAGE_REDIRECTS` and the `<meta http-equiv="refresh">` become dead: the built `highlights.html` would never be served, so the 301 would be landing on a page that immediately meta-refreshes again. Remove them in the same change rather than leaving two redirect layers stacked.

Nothing here blocks. Good to merge once the test legs finish.

PR #4227, opened 2026-09-04
fix(geometry): return max_num instead of NaN for singular H in symmetric_transfer_error

### LINKED ISSUE #4203: torch 2.14 flags a singular matrix that 2.9.1 did not, and symmetric_transfer_error returns NaN instead of max_num for a singular H
### Summary

Two linked findings, surfaced while moving the MPS gate to torch 2.14.0 (#4202).

1. **torch 2.14.0 detects an exactly-singular float32 matrix that 2.9.1 did not.** kornia supports
   `torch>=2.5.1`, so the two ends of the supported range disagree about whether the same input is
   invertible, and kornia branches on that answer.
2. **`symmetric_transfer_error` returns `nan` instead of `max_num` for a singular homography.**
   The mask arithmetic cannot neutralise a non-finite inverse, which is precisely the case the
   mask exists for. This one is not version-specific; finding 1 just makes it reachable for more
   inputs.

### 1. The cross-version difference

The matrix (rows 0 and 2 identical, so `det` is exactly 0):

```python
A = torch.tensor([[-956.0, -761.0, -67.0], [760.0, -37.0, -621.0], [-956.0, -761.0, -67.0]])
```

| | torch 2.9.1 | torch 2.14.0 |
|---|---|---|
| `torch.linalg.inv_ex(A).info` | `0` (reported valid) | **`3`** |
| `torch.linalg.inv(A)` | returned, no exception | **raises `_LinAlgError`** |
| `torch.linalg.det(A)` | `1.969` | `1.969` |

Same machine (M1, macOS 26.5.1), same fp32 input, measured a few hours apart on 2.9.1 and 2.14.0.
`det` still misreports on both — a zero determinant was never a reliable singularity test — but
the LU-based detection changed.

Consequences inside kornia:

- `kornia.core.utils._torch_inverse_cast` now **raises** `_LinAlgError` for this input where it
  previously returned a non-finite matrix. Everything that goes through it inherits that:
  `normalize_homography` / `denormalize_homography`, `invert_affine_transform`,
  `PinholeCamera.intrinsics_inverse`, every geometric augmentation's inverse transform,
  `image_registrator`, `planar_tracker`.
- `kornia.core.utils.safe_inverse_with_mask` returns `mask=False` where it returned `mask=True`,
  so `symmetric_transfer_error`, and therefore `find_homography_dlt_iterated` and `RANSAC`
  homography scoring, classify hypotheses differently between supported torch versions.
- The export path disagrees with the eager path for the same input: `safe_inverse_with_mask`
  under tracing takes the closed-form branch, whose validity test is `det != 0`, and the
  closed-form determinant of this matrix is `-20.0`, so tracing says **valid** while eager says
  **invalid**.

Verified on 2.14.0:

```
linalg.det                    = 1.969299554824829
inv_ex info                   = 3
_torch_inverse_cast           -> raised _LinAlgError
safe_inverse_with_mask mask   = [False]
closed-form det (export path) = [-20.0]  -> mask [True]
```

### 2. The masking bug this exposes

`symmetric_transfer_error` documents that a non-invertible `H` scores `max_num`:

```python
H_inv, good_H = safe_inverse_with_mask(H)
there = oneway_transfer_error(pts1, pts2, H, True, eps)
back  = oneway_transfer_error(pts2, pts1, H_inv, True, eps)
good_H_reshape = good_H.view(-1, 1).expand_as(there)
out = (there + back) * good_H_reshape.to(there.dtype) + max_num * (~good_H_reshape).to(there.dtype)
```

When `good_H` is `False`, `H_inv` is exactly what `inv_ex` leaves behind for a singular input —
here `[nan, nan, nan, inf, ...]`. `back` is then non-finite, and `nan * 0.0` is `nan`, not `0.0`,
so `out` is `nan` rather than `max_num`. Reproduced on 2.14.0 with two different singular
homographies, including a trivially singular one that every torch version flags:

```python
p1, p2 = torch.rand(1, 5, 2), torch.rand(1, 5, 2)
Z = torch.zeros(1, 3, 3); Z[0, 0, 0] = 1.0
symmetric_transfer_error(p1, p2, Z)   # -> [nan, nan, nan, nan, nan], expected 3.4028235e+38
```

`find_homography_dlt_iterated` feeds that straight into
`weights_new = torch.exp(-errors / (2 * soft_inl_th**2))`, so one singular intermediate `H` turns
every weight into `nan` and the returned homography with it — silently, since nothing raises.

The fix is `torch.where(good_H_reshape, there + back, max_num)`, which never multiplies a
non-finite value by zero. Worth checking the other error functions in `kornia/geometry/ransac.py`
for the same pattern.

### Suggested scope

- Replace the arithmetic mask in `symmetric_transfer_error` with `torch.where`, with a test that
  pins `max_num` for a singular `H` (the current shape-only coverage passes on `nan`).
- Decide the contract for `_torch_inverse_cast` on a singular input now that 2.5.1 and 2.14.0
  disagree: raise on both, or mask on both. Whatever is chosen, it should not depend on the torch
  version the user installed.
- Reconcile the eager and traced branches of `safe_inverse_with_mask`, which currently return
  opposite validity for the same matrix.

### Environment

kornia `main`, torch 2.14.0 and 2.9.1, macOS 26.5.1 arm64 (Apple M1), Python 3.11, CPU float32.
Related: #4201 (MPS SVD ceiling), #4202 (MPS gate on 2.14.0).

thegoodengineer:

Fix `symmetric_transfer_error` returning `NaN` instead of `max_num` when given a singular homography matrix.

When an input homography matrix is singular, `safe_inverse_with_mask` marks `good_H` as `False` and returns non-finite values (`NaN` / `Inf`) in `H_inv`. The previous implementation evaluated `oneway_transfer_error` with `H_inv` unshielded and attempted to mask the result using arithmetic: `(there + back) * mask + max_num * (~mask)`. Because `NaN * 0.0` evaluates to `NaN` in IEEE 754 floating-point arithmetic, `out` became `NaN` instead of `max_num`. In addition, during backpropagation, evaluating unshielded `NaN` values produced `NaN` gradients across the whole batch.

Shielding the *output* of the inverse is not enough for the backward pass: `inv_ex`'s backward is `-H_inv^T @ grad @ H_inv^T` over a saved `H_inv` full of `NaN`, so the exact-zero gradient `torch.where` hands the bad row still becomes `0 * NaN = NaN` in `H.grad`. That matters because `find_homography_dlt_iterated` (`homography.py:299`) differentiates the error through an estimated `H`, so the `NaN` reaches the correspondences on the differentiable-estimator path that motivates #4203. The *input* of the inverse is shielded instead.

This PR:
1. Takes the invertibility mask on a detached copy of `H`, replaces singular rows with the identity, and only then computes the differentiable inverse, so gradients with respect to both the correspondences and `H` itself stay finite (exactly `0` on the singular rows).
2. Uses `torch.where` to select `max_num` where `good_H` is `False`.
3. Handles both `squared=True` (default) and `squared=False` so non-invertible homographies reliably receive `max_num`. For `squared=False` this is a small behaviour change beyond the `NaN` repair: a singular row scored `(max_num + eps).sqrt()` before and scores `max_num` now. The one in-tree consumer, `find_homography_dlt_iterated`, feeds the error to `exp(-errors / (2 * soft_inl_th**2))`, which underflows to `0` at either magnitude, so no in-tree behaviour moves. The CHANGELOG entry says so.
4. Documents the `max_num` contract in the `Returns:` section, which #4203 assumed was already documented and was not.
5. Adds regression tests: `test_singular` (finite errors on valid rows, `max_num` on singular rows, finite gradients w.r.t. the correspondences), `test_singular_gradient_wrt_homography`, and `test_singular_gradient_through_estimated_homography` for the `H`-in-graph case.

Closes #4203

## What did you check?

Ran the new tests against unmodified `main` first to confirm they catch the regression:

```text
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AssertionError: Failed: got tensor([nan, nan, nan, nan, nan])
```

Forward output is `torch.equal` to the previous head of this branch for both `squared=True` and `squared=False`, over a mixed batch containing a singular row, so nothing moves for valid inputs. `max_num` is dtype-correct: float16 `65504.0`, bfloat16 `3.3895e38`, float32 `3.4028e38`, float64 `1.7977e308`. The extra batched 3x3 inverse costs `7.60 ms -> 8.53 ms` at `B=2048` on CPU.

```text
$ pytest tests/geometry/test_homography.py tests/geometry/test_ransac.py --dtype=float32,float64
============== 156 passed, 22 skipped, 4 xpassed, 2 warnings in 13.80s ==============

$ pytest tests/geometry/test_homography.py -k Symmetric --dtype=float16,bfloat16
======================== 13 passed, 118 deselected in 0.63s ========================

All checks passed!
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -369,8 +369,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   When `safe_inverse_with_mask` flags an input as non-invertible, the unshielded inverse matrix containing
   non-finite values propagated `NaN` through `oneway_transfer_error` and the arithmetic mask
   `(there + back) * mask + max_num * (~mask)`, poisoning both the forward error and autograd gradients.
-  Invalid inverse matrices are now guarded with an identity fallback before transfer error calculation,
-  and the final masking uses `torch.where` to cleanly assign `max_num`. (#4203)
+  Singular rows are now replaced by the identity *before* the differentiable inverse is taken, so the
+  gradients with respect to both the correspondences and `H` itself stay finite (zero on the singular
+  rows), and the final masking uses `torch.where` to cleanly assign `max_num`. Values for invertible
+  homographies are unchanged; for `squared=False` a singular row now scores `max_num` rather than the
+  previous `(max_num + eps).sqrt()`, matching the documented contract. (#4203)
 
 * Constructing `ScaleSpaceDetector` with `compile_modules` no longer permanently mutates the process-global
   `torch._dynamo.config.capture_dynamic_output_shape_ops` setting. (#4134)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -365,6 +365,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `N < 3` raised an incidental `IndexError`, while `N >= 3` silently returned `(B, 2)` extents computed from the
   wrong vertices. (#4218)
 
+* Fix `symmetric_transfer_error` returning `NaN` instead of `max_num` for singular homographies.
+  When `safe_inverse_with_mask` flags an input as non-invertible, the unshielded inverse matrix containing
+  non-finite values propagated `NaN` through `oneway_transfer_error` and the arithmetic mask
+  `(there + back) * mask + max_num * (~mask)`, poisoning both the forward error and autograd gradients.
+  Invalid inverse matrices are now guarded with an identity fallback before transfer error calculation,
+  and the final masking uses `torch.where` to cleanly assign `max_num`. (#4203)
+
 * Constructing `ScaleSpaceDetector` with `compile_modules` no longer permanently mutates the process-global
   `torch._dynamo.config.capture_dynamic_output_shape_ops` setting. (#4134)
 
--- a/kornia/geometry/homography.py
+++ b/kornia/geometry/homography.py
@@ -44,7 +44,8 @@ def oneway_transfer_error(
         eps: Small constant for safe sqrt.
 
     Returns:
-        the computed distance with shape :math:`(B, N)`.
+        the computed distance with shape :math:`(B, N)`. Rows whose homography is not invertible
+        score ``max_num``, i.e. ``torch.finfo(pts1.dtype).max``, for both values of ``squared``.
 
     """
     KORNIA_CHECK_SHAPE(H, ["B", "3", "3"])
@@ -109,7 +110,8 @@ def symmetric_transfer_error(
         eps: Small constant for safe sqrt.
 
     Returns:
-        the computed distance with shape :math:`(B, N)`.
+        the computed distance with shape :math:`(B, N)`. Rows whose homography is not invertible
+        score ``max_num``, i.e. ``torch.finfo(pts1.dtype).max``, for both values of ``squared``.
 
     """
     KORNIA_CHECK_SHAPE(H, ["B", "3", "3"])
@@ -122,10 +124,14 @@ def symmetric_transfer_error(
     max_num = torch.finfo(pts1.dtype).max
     # From Hartley and Zisserman, Symmetric transfer error (4.7)
     # dist = \sum_{i} (d(x, H^-1 x')**2 + d(x', Hx)**2)
-    H_inv, good_H = safe_inverse_with_mask(H)
+    # Shield the *input* of the inverse, not its output: ``inv_ex`` backward is
+    # ``-H_inv^T @ grad @ H_inv^T`` over a saved ``H_inv`` full of non-finite values, so masking the
+    # result afterwards still leaves ``0 * nan = nan`` in ``H.grad``. The mask is taken on a detached
+    # copy so it carries no graph, and the differentiable inverse then only ever sees a valid matrix.
+    _, good_H = safe_inverse_with_mask(H.detach())
     eye = torch.eye(3, device=H.device, dtype=H.dtype).expand_as(H)
     H_safe = torch.where(good_H.view(-1, 1, 1), H, eye)
-    H_inv_safe = torch.where(good_H.view(-1, 1, 1), H_inv, eye)
+    H_inv_safe, _ = safe_inverse_with_mask(H_safe)
 
     there: torch.Tensor = oneway_transfer_error(pts1, pts2, H_safe, True, eps)
     back: torch.Tensor = oneway_transfer_error(pts2, pts1, H_inv_safe, True, eps)
@@ -155,7 +161,8 @@ def line_segment_transfer_error_one_way(
         squared: if True (default is False), the squared distance is returned.
 
     Returns:
-        the computed distance with shape :math:`(B, N)`.
+        the computed distance with shape :math:`(B, N)`. Rows whose homography is not invertible
+        score ``max_num``, i.e. ``torch.finfo(pts1.dtype).max``, for both values of ``squared``.
 
     """
     KORNIA_CHECK_SHAPE(H, ["B", "3", "3"])
--- a/kornia/geometry/homography.py
+++ b/kornia/geometry/homography.py
@@ -44,8 +44,7 @@ def oneway_transfer_error(
         eps: Small constant for safe sqrt.
 
     Returns:
-        the computed distance with shape :math:`(B, N)`. Rows whose homography is not invertible
-        score ``max_num``, i.e. ``torch.finfo(pts1.dtype).max``, for both values of ``squared``.
+        the computed distance with shape :math:`(B, N)`.
 
     """
     KORNIA_CHECK_SHAPE(H, ["B", "3", "3"])
@@ -161,8 +160,7 @@ def line_segment_transfer_error_one_way(
         squared: if True (default is False), the squared distance is returned.
 
     Returns:
-        the computed distance with shape :math:`(B, N)`. Rows whose homography is not invertible
-        score ``max_num``, i.e. ``torch.finfo(pts1.dtype).max``, for both values of ``squared``.
+        the computed distance with shape :math:`(B, N)`.
 
     """
     KORNIA_CHECK_SHAPE(H, ["B", "3", "3"])
--- a/kornia/geometry/homography.py
+++ b/kornia/geometry/homography.py
@@ -123,14 +123,18 @@ def symmetric_transfer_error(
     # From Hartley and Zisserman, Symmetric transfer error (4.7)
     # dist = \sum_{i} (d(x, H^-1 x')**2 + d(x', Hx)**2)
     H_inv, good_H = safe_inverse_with_mask(H)
+    eye = torch.eye(3, device=H.device, dtype=H.dtype).expand_as(H_inv)
+    H_inv_safe = torch.where(good_H.view(-1, 1, 1), H_inv, eye)
 
     there: torch.Tensor = oneway_transfer_error(pts1, pts2, H, True, eps)
-    back: torch.Tensor = oneway_transfer_error(pts2, pts1, H_inv, True, eps)
+    back: torch.Tensor = oneway_transfer_error(pts2, pts1, H_inv_safe, True, eps)
     good_H_reshape: torch.Tensor = good_H.view(-1, 1).expand_as(there)
-    out = (there + back) * good_H_reshape.to(there.dtype) + max_num * (~good_H_reshape).to(there.dtype)
-    if squared:
-        return out
-    return (out + eps).sqrt()
+
+    out = there + back
+    if not squared:
+        out = (out + eps).sqrt()
+    max_tensor = torch.full_like(out, max_num)
+    return torch.where(good_H_reshape, out, max_tensor)
 
 
 def line_segment_transfer_error_one_way(
--- a/kornia/geometry/homography.py
+++ b/kornia/geometry/homography.py
@@ -123,10 +123,11 @@ def symmetric_transfer_error(
     # From Hartley and Zisserman, Symmetric transfer error (4.7)
     # dist = \sum_{i} (d(x, H^-1 x')**2 + d(x', Hx)**2)
     H_inv, good_H = safe_inverse_with_mask(H)
-    eye = torch.eye(3, device=H.device, dtype=H.dtype).expand_as(H_inv)
+    eye = torch.eye(3, device=H.device, dtype=H.dtype).expand_as(H)
+    H_safe = torch.where(good_H.view(-1, 1, 1), H, eye)
     H_inv_safe = torch.where(good_H.view(-1, 1, 1), H_inv, eye)
 
-    there: torch.Tensor = oneway_transfer_error(pts1, pts2, H, True, eps)
+    there: torch.Tensor = oneway_transfer_error(pts1, pts2, H_safe, True, eps)
     back: torch.Tensor = oneway_transfer_error(pts2, pts1, H_inv_safe, True, eps)
     good_H_reshape: torch.Tensor = good_H.view(-1, 1).expand_as(there)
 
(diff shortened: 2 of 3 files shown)


ducha-aiki:
## Changes requested — the forward fix is right; the gradient claim is broader than the fix

Reviewed at `32f7851f` against merge-base `25b2ca72`, torch 2.14.0 / py311 / macOS arm64. The `torch.where`
replacement is correct and the fail-on-base is real. But the PR says it keeps "backward gradients finite and
well-defined", and the test only checks `pts1`. `H`'s own gradient is still `nan`, and that leaks back into the
points on the path this function is actually used on.

### The finding

`H.grad` at head, batch of 2 with row 1 singular (`H[1] = 0; H[1,0,0] = 1`):

```
row 0: [[-2.2223, -1.2788, -4.2495], [-1.0398, -1.1284, -3.6378], [ 2.3263,  1.4665,  3.3507]]
row 1: [[    nan,     nan,     nan], [    nan,     nan,     nan], [    nan,     nan,     nan]]
```

`torch.where` hands `back` a gradient of exactly `0` for the bad row, but `inv_ex`'s backward is
`-H_inv^T @ grad @ H_inv^T` over a saved `H_inv` full of `nan`, and `0 * nan = nan`. Shielding the *output* of
the inverse does not reach it; only shielding its *input* does.

**Why this is not academic.** The internal caller `find_homography_dlt_iterated` (`homography.py:299`) computes
`H = find_homography_dlt(points1, points2, weights)` and then differentiates the error through that `H`. So
whenever `H` is in the graph — the differentiable-estimator case, which is the whole reason #4203 says "one
singular intermediate `H` turns every weight into nan and the returned homography with it" — the `nan` reaches
the points even at head:

```python
pts = torch.rand(2, 5, 2, requires_grad=True)
Hd  = torch.stack([torch.eye(3) * pts[0,0,0], torch.zeros(3,3) * pts[1,0,0]])
symmetric_transfer_error(pts, torch.rand(2,5,2), Hd).sum().backward()
# head: pts.grad[1, 0, 0] == nan
```

To be fair to the PR: **this is not a regression** — `main` is `nan` there too — and your `pts1` test is a real
improvement over it. So narrowing the claim is an acceptable resolution. But the one-line version of the full fix
is cheap, and I ran it rather than proposing it:

```python
_, good_H = safe_inverse_with_mask(H.detach())            # mask only, carries no graph
eye = torch.eye(3, device=H.device, dtype=H.dtype).expand_as(H)
H_safe = torch.where(good_H.view(-1, 1, 1), H, eye)
H_inv_safe, _ = safe_inverse_with_mask(H_safe)            # differentiable inverse of a shielded matrix
```

Measured: `H.grad` finite (`0` on the bad row); `pts.grad` finite in the `H`-in-graph case above; forward output
**`torch.equal` to your current head** for both `squared=True` and `squared=False`; cost `0.563 ms -> 0.741 ms` at
`B=2048` on CPU, one extra batched 3x3 inverse. Your `H_inv_safe`/`H_safe` `torch.where`s then become redundant
and can go.

Either fix it, or drop "and backward gradients finite and well-defined" from the body and the "poisoning ...
autograd gradients" clause from the CHANGELOG down to what the test proves (finite gradients w.r.t. the
*correspondences*), and say the `H` gradient is unchanged. What should not ship is the current pairing of a broad
claim with a narrow test.

### Two smaller asks

1. **`squared=False` changed magnitude for a bad row, and nothing says so.** `main` computed
   `(max_num + eps).sqrt()` there (`1.845e19` in float32); head returns `max_num` (`3.403e38`). Yours is the more
   defensible reading of "scores `max_num`" — I am not asking you to revert it — but it is a behaviour change
   beyond the `nan` repair and belongs in the CHANGELOG sentence. I checked the one in-tree consumer:
   `find_homography_dlt_iterated` feeds it to `exp(-errors / (2 * soft_inl_th**2))`, which underflows to `0` at
   either magnitude, so no in-tree behaviour moves.
2. **The `max_num` contract is not documented anywhere.** #4203 says the function "documents that a
   non-invertible `H` scores `max_num`"; grepping, it does not — `Returns:` says only "the computed distance".
   This PR is what turns that into a real, testable contract, so add the sentence to the docstring.

### What I verified as good

- **Fail-on-base 2/2:** `test_singular[cpu-float32]` and `[cpu-float64]` on `25b2ca72` -> `assert not tensor(True)`.
  Direct repro on base: `[nan, nan, nan, nan, nan]` for both `squared` values, and `pts1.grad[1]` all `nan`.
- **Head:** `tests/geometry/test_homography.py` + `tests/geometry/test_ransac.py`, cpu, float32+float64 ->
  **152 passed, 22 skipped, 4 xpassed**.
- **Good rows are byte-identical to `main`** (`torch.equal` over a 3x7 batch with a translation homography) —
  the `* 1.0 + max_num * 0.0` the old line did is exact, so nothing moves for valid inputs.
- **`max_num` is dtype-correct**, which was my first suspicion given the default is `torch.finfo(pts1.dtype).max`
  and `torch.full_like` writes into the output dtype: float16 -> `65504.0`, bfloat16 -> `3.3895e38`,
  float64 -> `1.7977e308`, each matching `torch.finfo(dtype).max`. The half legs will run `test_singular` and it
  passes there.

ducha-aiki:
Heads-up on top of the review above: #4215 merged and put a `CHANGELOG.md` entry on the same lines yours touches,
so this went `CONFLICTING`. A conflicting PR runs **no `pull_request` workflow** at all, so it would also have
stopped reporting CI.

Rebased onto `main` (`ca1f2f7b`) for you and force-pushed to
`thegoodengineer:fix/symmetric-transfer-error-singular` — new head `02e0b602`. Both CHANGELOG entries kept.
`kornia/geometry/homography.py` and `tests/geometry/test_homography.py` are **byte-identical across the rebase**
(`152 passed, 22 skipped, 4 xpassed` on the new head), so nothing in the review changed and none of it is about
the rebase.

Please `git fetch` and reset to `02e0b602` before you push the review fixes, or you will re-introduce the
conflict.

thegoodengineer:
Thanks, that is a fair catch and the distinction between shielding the output and the input of the inverse is the part I got wrong. I took the full fix rather than narrowing the claim. Reset to `02e0b602` first, thanks for the rebase. New head `9bed7897`.

**1. `H`'s gradient.** Applied your snippet essentially as written:

```python
_, good_H = safe_inverse_with_mask(H.detach())
eye = torch.eye(3, device=H.device, dtype=H.dtype).expand_as(H)
H_safe = torch.where(good_H.view(-1, 1, 1), H, eye)
H_inv_safe, _ = safe_inverse_with_mask(H_safe)
```

The redundant `H_inv_safe` `torch.where` is gone. There is a comment above it explaining why the input has to be shielded, so the next reader does not simplify it back to masking the output.

Confirmed on my side: `H.grad` finite with exactly `0` on the singular row, and `pts.grad` finite in your `H`-in-graph repro. Forward output is `torch.equal` to the previous head for both `squared=True` and `squared=False` over a mixed batch. Cost at `B=2048` on CPU is `7.60 ms -> 8.53 ms` here, same one extra batched inverse, my machine is just slower than yours.

Two tests cover this now instead of a narrowed claim: `test_singular_gradient_wrt_homography` and `test_singular_gradient_through_estimated_homography`, the latter built on your `find_homography_dlt_iterated` case, with a comment naming why that path matters.

**2. `squared=False` magnitude.** In the CHANGELOG sentence now, with the old value, the new value, and the note that `find_homography_dlt_iterated` underflows to `0` at either magnitude so no in-tree behaviour moves. Also in the PR body.

**3. `max_num` contract.** Added to `Returns:`:

> the computed distance with shape `(B, N)`. Rows whose homography is not invertible score `max_num`, i.e. `torch.finfo(pts1.dtype).max`, for both values of `squared`.

Re-ran `tests/geometry/test_homography.py` + `tests/geometry/test_ransac.py` on cpu float32+float64: 156 passed, 22 skipped, 4 xpassed. `-k Symmetric` on float16+bfloat16: 13 passed. Ruff check and format clean on both files.

ducha-aiki:
## Delta re-review at `9bed7897` — the gradient fix is verified; one sentence went into two docstrings where it is false

Re-read the whole diff against merge-base `ca1f2f7b`, torch 2.14.0 (plus the half profiles on 2.9.1), CPU.

**Verified**

- The snippet landed as measured: `H.grad` finite with exactly `0` on the singular row; `pts.grad` finite in the `find_homography_dlt_iterated`-shaped case.
- **Fail-on-base:** the three `singular` tests copied onto `main` -> 6 failed (float32+float64). Head: `test_homography.py` + `test_ransac.py` -> 156 passed, 22 skipped, 4 xpassed; `-k Symmetric` under the `cpu-float16` and `cpu-bfloat16` profiles -> 7 passed each.
- **Forward `torch.equal` to your previous head `02e0b602`** over a mixed batch (translation, singular and shear rows) at float32, float64 and float16, both `squared` values. The input shielding changed gradients only.
- CHANGELOG states the `squared=False` magnitude change with old and new values and the `find_homography_dlt_iterated` underflow note. `uvx ruff@0.16.4` check/format clean.
- The float16 rerun dispatched this morning passed; all 38 checks are green at head.

**The one ask:** the new `Returns:` sentence, "Rows whose homography is not invertible score `max_num`, i.e. `torch.finfo(pts1.dtype).max`, for both values of `squared`", went into **three** docstrings, and it is only true of `symmetric_transfer_error`. `oneway_transfer_error` and `line_segment_transfer_error_one_way` never invert `H` and contain no `max_num` at all. Executed at head with the same singular row (`H[1] = 0; H[1, 0, 0] = 1`):

```
oneway_transfer_error               row 1: [1.2e15, 5.0e12, 8.6e14, ...]   finite, not max_num
line_segment_transfer_error_one_way row 1: [0.117, 0.034, 0.011, ...]      finite, not max_num
symmetric_transfer_error            row 1: all max_num                      as documented
```

Please drop the sentence from those two. A documented contract that nothing implements is exactly what the next reviewer pins and then fails. Everything else is approved; after that and the `CHANGELOG.md` rebase (it conflicts with today's merges; the source and test files merge cleanly) this is mergeable.

thegoodengineer:
Fixed both, new head `06e68a16`.

**The `Returns:` sentence.** You are right, and the cause is dull: I applied it with a plain string replace on the `Returns:` block, and all three functions had the identical `the computed distance with shape :math:`(B, N)`.` line, so it landed in all three. It is now only on `symmetric_transfer_error`; `oneway_transfer_error` and `line_segment_transfer_error_one_way` are back to their original wording. Reproduced your numbers before reverting, same singular row, float32:

```
oneway_transfer_error               row 1: [2.08e15, 1.22e15, 4.98e12, 8.64e14]   finite, no max_num
line_segment_transfer_error_one_way row 1: [0.313, 0.098, 0.177, 0.385]           finite, no max_num
symmetric_transfer_error            row 1: all max_num
```

Neither of those two calls `safe_inverse_with_mask` at all, so there was never a contract there to document.

**CHANGELOG rebase.** Rebased onto `e8111a61`. The only conflict was the `CHANGELOG.md` entry against #4218; both entries are kept and the source and test files came across unchanged.

Re-ran at head: `tests/geometry/test_homography.py` + `tests/geometry/test_ransac.py` on cpu float32+float64 give 156 passed, 22 skipped, 4 xpassed; `-k Symmetric` on float16+bfloat16 gives 13 passed. Ruff check and format clean on both files.

ducha-aiki (approved):
Delta re-review: approved at `06e68a16`. Compared with the previously reviewed `9bed7897`: the source/test delta is exactly the removal of the incorrect `max_num` sentences from `oneway_transfer_error` and `line_segment_transfer_error_one_way`. The contract remains on `symmetric_transfer_error`, where it is implemented.

Re-ran `tests/geometry/test_homography.py tests/geometry/test_ransac.py --device=cpu --dtype=float32,float64` in an isolated checkout using Python 3.11 / torch 2.14.0: 156 passed, 22 skipped, 4 non-strict XPASS, 2 SVD warnings. No new implementation changes since the prior measured gradient review. Current CI is green and GitHub reports mergeable. No outstanding review requests.

ducha-aiki (approved):

PR #4228, opened 2026-09-04
fix(geometry): finite gradients for quaternion_exp_to_log and euler_from_quaternion at their rotation singularities

### LINKED ISSUE #4007: NaN gradients at the rotation singularity: quaternion_exp_to_log and euler_from_quaternion differentiate through acos/asin clamped to the domain boundary
`quaternion_exp_to_log` returns a **NaN** gradient at the identity quaternion `(1, 0, 0, 0)`. The forward value is correct — the log of the identity is the origin — but backward is poisoned, at the single most common input in the domain.

```python
>>> q = torch.tensor([1., 0., 0., 0.], requires_grad=True); quaternion_exp_to_log(q).sum().backward(); q.grad
tensor([nan, 0., 0., 0.])
```

The source is an unguarded `acos` at the boundary of its domain:

```python
# quaternion_exp_to_log, kornia/geometry/conversions.py:1191-1193
quaternion_log: torch.Tensor = (
    quaternion_vector * torch.acos(torch.clamp(quaternion_scalar, min=-1.0, max=1.0)) / norm_q
)
```

`torch.clamp(w, -1, 1)` guards the **forward** value, so `acos` never sees an out-of-domain argument and the returned value is right. But clamp passes its gradient straight through at the endpoints, and `d/dw acos(w) = -1/sqrt(1 - w²)` is `-inf` at `w = ±1`. At the identity the vector part is exactly zero, so the `-inf` meets a `0` and `0 · inf = NaN`.

This is the same failure *class* as #3949 — a singularity where `torch.where`/clamp fixes the value but not the derivative — but a different mechanism (`acos` at `±1` rather than `sqrt(0)`) in a function that pass did not cover.

Two further observations from the repro below:

- With a non-zero vector part and `w = 1` (a non-unit input) the NaN degrades to a bare `-inf` rather than becoming finite, so the defect is not specific to the all-zero vector.
- At the antipode `(-1, 0, 0, 0)` the NaN in the scalar slot is joined by `3.1416e+08` in the vector slots. That is `acos(-1)/eps = π/1e-8`, a second, `eps`-scale artifact of dividing by the clamped `norm_q` at a point where the true norm is zero.

## The same pattern at a second site

`euler_from_quaternion` has the identical `clamp` + arc-trig construction, and fails the same way once the argument lands exactly on the boundary (gimbal lock, `pitch = ±π/2`):

```python
# euler_from_quaternion, kornia/geometry/conversions.py:1376-1378
sinp = 2.0 * (w * y - z * x)
sinp = sinp.clamp(min=-1.0, max=1.0)
pitch = sinp.asin()
```

`d/dx asin(x) = 1/sqrt(1 - x²)` is `+inf` at `x = ±1`, and the products in the chain rule turn some of those into NaN. Note this one only bites when `sinp` is *exactly* `±1`; a quaternion that is at gimbal lock only to within rounding gets a finite but enormous gradient (~`1.3e16`) instead, which is arguably worse because it is silent.

## Repro

```python
import torch
from kornia.geometry.conversions import euler_from_quaternion, quaternion_exp_to_log

# 1. quaternion_exp_to_log at the identity: acos'(1) = -inf, vector part = 0 -> 0 * inf = NaN
q = torch.tensor([1.0, 0.0, 0.0, 0.0], dtype=torch.float64, requires_grad=True)
quaternion_exp_to_log(q).sum().backward()
print("exp_to_log grad at the identity (1,0,0,0)  :", q.grad)

# the antipode is the same singularity
qa = torch.tensor([-1.0, 0.0, 0.0, 0.0], dtype=torch.float64, requires_grad=True)
quaternion_exp_to_log(qa).sum().backward()
print("exp_to_log grad at the antipode (-1,0,0,0) :", qa.grad)

# with a non-zero vector part the NaN becomes a bare -inf, still unusable
qv = torch.tensor([1.0, 0.1, 0.0, 0.0], dtype=torch.float64, requires_grad=True)
quaternion_exp_to_log(qv).sum().backward()
print("exp_to_log grad at (1, 0.1, 0, 0)          :", qv.grad)

# finite away from the singularity
qg = torch.tensor([0.9, 0.3, 0.2, 0.1], dtype=torch.float64, requires_grad=True)
quaternion_exp_to_log(qg).sum().backward()
print("exp_to_log grad at (0.9,0.3,0.2,0.1)       :", qg.grad)

# 2. the same clamp+arc-trig pattern in euler_from_quaternion, at gimbal lock
w, x, y, z = (torch.tensor(v, dtype=torch.float64, requires_grad=True) for v in (1.0, 0.0, 0.5, 0.0))
out = euler_from_quaternion(w, x, y, z)
print("euler pitch at sinp == 1.0                 :", out[1].item())
sum(o.sum() for o in out).backward()
print("euler grads (w,x,y,z) at sinp == 1.0       :", [t.grad.item() for t in (w, x, y, z)])

# 3. the derivative that drives both
for s in (1.0, -1.0, 0.999999):
    t = torch.tensor(s, dtype=torch.float64, requires_grad=True)
    torch.acos(torch.clamp(t, -1.0, 1.0)).backward()
    print(f"d/ds acos(clamp(s,-1,1)) at s = {s:<9}  : {t.grad.item()}")
```

## Observed

```
exp_to_log grad at the identity (1,0,0,0)  : tensor([nan, 0., 0., 0.], dtype=torch.float64)
exp_to_log grad at the antipode (-1,0,0,0) : tensor([       nan, 3.1416e+08, 3.1416e+08, 3.1416e+08], dtype=torch.float64)
exp_to_log grad at (1, 0.1, 0, 0)          : tensor([-inf, 0., 0., 0.], dtype=torch.float64)
exp_to_log grad at (0.9,0.3,0.2,0.1)       : tensor([-3.6788, -0.3444,  0.1722,  0.6888], dtype=torch.float64)
euler pitch at sinp == 1.0                 : 1.5707963267948966
euler grads (w,x,y,z) at sinp == 1.0       : [inf, nan, inf, nan]
d/ds acos(clamp(s,-1,1)) at s = 1.0        : -inf
d/ds acos(clamp(s,-1,1)) at s = -1.0       : -inf
d/ds acos(clamp(s,-1,1)) at s = 0.999999   : -707.1069579492319
```

## Expected

A finite gradient at the identity and at the antipode, and at gimbal lock for `euler_from_quaternion`.

On the unit sphere — the intended domain — there is no mathematical singularity in the quantity being differentiated, only in the parametrization. `acos(w)` has an infinite derivative at `w = ±1`, but it is multiplied by a vector part that goes to zero there: as `θ → 0`, `acos(w) → θ/2` and `‖v‖ → θ/2`, so the log map approaches the identity map and its derivative is bounded. Sampling unit quaternions at `θ = 1e-1 … 1e-6` returns exactly `θ/2`, linearly, with no sign of a blow-up in the value. The composite is smooth; only the chosen decomposition is not.

(The `-inf` at the non-unit `(1, 0.1, 0, 0)` above is a related but distinct symptom — off the unit sphere `w = 1` with `v ≠ 0` is not a rotation at all, so there is no limit argument to appeal to there. It is included only to show that the boundary derivative, not the zero vector part, is what drives the failure.)

The fix pattern is already in this file, cited in #3949: `axis_angle_to_rotation_matrix` writes `torch.sqrt(theta2.clamp(min=1e-12))  # clamping to ensure no nan gradients`. The analogue here is to keep the `acos`/`asin` argument strictly inside `(-1, 1)` — clamping to `±(1 - eps)` rather than `±1` — or to select the singular branch with a `torch.where` whose *both* arms are finite, which is the approach #3980 took for #3949.

Why this matters: `quaternion_exp_to_log` is the log map used to put a rotation into a minimal 3-parameter tangent space, which is precisely what pose-optimization and rotation-regression loops differentiate through. The identity is the standard initialization for those loops, so the first backward pass returns NaN and the optimization dies before it starts — silent in forward, total in backward, exactly as in #3949.

## Environment

```
kornia 0.9.0rc1 (main @ 0685b1f5) | torch 2.9.1 | python 3.11.14 | CPU
```

Reproduces at float32 and float64.

---

mohansree14:
## What

`quaternion_exp_to_log` returns a **NaN gradient at the identity quaternion `(1,0,0,0)`** — the standard initialization for pose/rotation optimization loops. `euler_from_quaternion`'s `pitch` has the identical failure at gimbal lock. Both are Lie-group log-map-style operations that pose-optimization and rotation-regression training loops differentiate through, so the very first backward pass from the most ordinary starting point returns NaN and the optimization dies before it starts.

## Root cause

Both functions clamp an input into `[-1, 1]` before feeding it to an inverse trig function (`acos`, `asin`), which keeps the **forward** value correct. But `d(acos)/dw = -1/sqrt(1-w²)` (and the analogous `asin` derivative) is unbounded exactly at `w = ±1`. In `quaternion_exp_to_log`, that unbounded derivative gets multiplied by the quaternion's vector part — which is *exactly zero* at the identity — giving `0 * inf = NaN` in IEEE754, even though the true partial derivative there is well-defined (the composite `v * acos(w) / ‖v‖` is identically the zero function of `w` whenever `v` is held at exactly zero, so its exact derivative w.r.t. `w` is trivially 0 — there's no real mathematical singularity here, only a chain-rule artifact of computing `acos'(w) = -inf` as an intermediate before it gets multiplied away).

**A wrinkle worth flagging:** this is torch-version-dependent, but not for the reason this PR originally gave. `acos`'s boundary derivative is `-inf` on every supported torch version (measured on 2.5.1, 2.7.0, 2.9.1 and 2.14.0) and has not changed. What changed in torch 2.14 is `torch.clamp`: its backward returns `0` at the closed boundary instead of the pass-through `1.0` of earlier versions, which kills the `-inf` before it can multiply anything. So the bug does not reproduce on 2.14, and does reproduce on the 2.5.1 and 2.9.1 legs of kornia's CI matrix. The fix below relies on neither behaviour — it prevents the unbounded intermediate from ever being differentiated. Thanks to @ducha-aiki for the correct diagnosis; the general form of the `clamp`-as-gradient-guard problem is now #4229.

## The fix

For both functions: substitute a safe argument (`0`) into the `acos`/`asin` call that gets *differentiated*, while taking the returned *value* from a separately `.detach()`ed copy of the real (possibly boundary) argument. This keeps the forward value byte-identical to the unguarded implementation — verified over 500 random inputs, several forced onto the exact boundary — while the gradient no longer depends on differentiating `acos`/`asin` at a point where its derivative is unbounded.

```python
w_clamped = torch.clamp(quaternion_scalar, min=-1.0, max=1.0)
at_boundary = w_clamped.abs() >= 1.0
safe_w = torch.where(at_boundary, torch.zeros_like(w_clamped), w_clamped)
acos_w = torch.where(at_boundary, w_clamped.detach().acos(), safe_w.acos())
```

**Out of scope, noted explicitly:** `quaternion_exp_to_log`'s antipode `(-1,0,0,0)` still returns a large-but-finite gradient (`~π/eps ≈ 3.14e8`) from an unrelated, pre-existing division by the clamped (zero) vector norm. That's a separate numerical-roughness issue, not a NaN, and is documented as out of scope in the CHANGELOG.

## Testing

- `test_convention_gradient_is_finite_at_the_acos_boundary_4007` (identity, antipode, and a non-unit boundary input) and `test_convention_pitch_gradient_is_finite_at_gimbal_lock_4007` — both verified to **fail on the unmodified code** (on torch 2.9.1, matching the issue's environment) and pass on the fix, across float16/bfloat16/float32/float64.
- Forward-value byte-identity verified against the unguarded implementation over 500 random inputs (including forced-boundary cases) for both functions.
- Full `tests/geometry/test_conversions.py` (390 tests): **identical failure set** with and without this change, on both torch 2.9.1 and 2.14.0 — confirmed via diff, not just count. The pre-existing float16/bfloat16 failures elsewhere in the file (unrelated test classes) are untouched.
- Doctests for the module: 34 passed.
- Verified `torch.compile(fullgraph=True)` compiles cleanly for both functions (`backend="eager"` — no C++ toolchain available locally to exercise `inductor` directly).
- `grep -rnE "#4007|issues/4007" kornia/ tests/` per `AGENTS.md`'s documented-defect gate — every `kornia/` hit is retrospective ("This used to return `nan`..."), unambiguous past tense.

## Not yet verified

No `pixi`, CUDA, or MPS hardware in my local environment — verified with throwaway `uv` venvs (torch 2.9.1 and 2.14.0, CPU) instead of the project's usual toolchain.

Fixes #4007


ducha-aiki:
## Changes requested — the fix is correct; the explanation attached to it is not

Reviewed at `fe846377` against merge-base `25b2ca72`. The guard is right, the forward is byte-identical, and the
pins do discriminate on a torch version CI runs. The problem is the root cause you wrote into the source and the
body: it names the wrong operator, and that wrong claim is now a comment in `kornia/geometry/conversions.py`.

### It is `torch.clamp`, not `acos`

The body says *"On torch 2.14.0, `acos`'s own backward already returns 0 at the boundary instead of `-inf`, so
the bug doesn't reproduce there at all"*, and `conversions.py:1164-1165` bakes it in: *"on some torch/backend
builds it is exactly `+-inf`; PyTorch >= 2.10-ish already zeros it, but nothing here may depend on that"*.

Measured on all three versions in play:

| | torch 2.5.1 | torch 2.9.1 | torch 2.14.0 |
| --- | --- | --- | --- |
| `acos` grad at `w = 1` | `-inf` | `-inf` | **`-inf`** |
| `asin` grad at `x = 1` | `inf` | `inf` | **`inf`** |
| `clamp(w, -1, 1)` grad at `w = 1` | `1.0` | `1.0` | **`0.0`** |
| `(v * clamp(w).acos() / n)` grad, `v = 0` | `nan` | `nan` | `0.0` |
| `clamp(2*(w*y - z*x)).asin()` grad at lock | `inf` | `inf` | `0.0` |

`acos`'s boundary derivative did not change. What changed is that torch 2.14 treats the closed boundary as
*clamped* and returns `0` from `clamp`'s backward, which kills the `-inf` before it can multiply anything. Your
conclusion — that the bug does not reproduce on 2.14 and that the fix must not depend on which version zeroes it
— is right; only the attribution is wrong. Please correct both the comment and the body. A false statement about
torch semantics sitting in production source is the stale-prose class AGENTS.md polices, and this one is the kind
a future reader would act on.

The correction also makes the pins look *better*, not worse. **kornia's PR matrix runs 2.5.1**, so they are not
vacuous. Run against `main`'s `kornia/` under torch 2.5.1:

```
base (25b2ca72), torch 2.5.1:  identity [nan, 0, 0, 0]   antipode [nan, ...]   nonunit [-inf, 0, 0, 0]
                               pitch grads [inf, nan, inf, nan]
head (fe846377), torch 2.5.1:  identity [0, 0, 0, 0]     antipode [0, ...]     nonunit [0, 0, 0, 0]
                               pitch grads [0, -0, 0, -0]
```

For completeness, the same fail-on-base run under **torch 2.14 passes on base** (`-k 4007` -> 4 passed). Worth a
sentence in the test comment so the next person who runs it locally on a recent torch does not conclude the pin
is vacuous — say it is the 2.5.1 leg that discriminates, and why.

### Two smaller asks

1. **`Tracked in #4007` survives twice in `kornia/`** (`conversions.py:1133` and `:1346`). Under the AGENTS grep
   gate a hit in `kornia/` after the closing change reads as incomplete work. Both notes are genuinely
   retrospective in substance ("This used to return `nan`..."), which is a legitimate reason to keep the citation
   — #4184 kept one on the same grounds — but the *wording* has to say so. #4184's accepted form was "previously
   ... mistaken for an actual gradient bug", with no "Tracked in". Reword to something like "Previously tracked
   in #4007" and the gate is satisfied.
2. **Copilot's `near_boundary` point is correct and slightly larger than a naming nit.** `near_boundary` is
   `(1.0, 0.1, 0.0, 0.0)` — byte-identical to `non_unit` three lines up, and exactly *on* the boundary, so the
   comment "at and near the boundary" describes coverage the test does not have. The off-boundary branch of your
   `torch.where` is never exercised by that assertion, and it is the branch that proves the guard changes nothing
   away from `+-1`. Use something like `w = 1 - eps` with a nonzero vector part and assert against the unguarded
   value. Copilot's other point stands too: the comment calls the check below a "byte-identical-forward check"
   and it is a correctness check against an all-zeros expected tensor.

### What I verified as good

- **Forward is byte-identical**, not just close. 500 random quaternions per dtype including the identity, the
  antipode and a non-unit boundary input, plus 300 euler triples including exact gimbal lock: `torch.equal` head
  vs base for **float16, bfloat16, float32 and float64**, both functions. Non-boundary gradients also `torch.equal`
  in float64 over 200 inputs, so nothing moved off the boundary either.
- **Head:** `tests/geometry/test_conversions.py` cpu float32+float64 -> **619 passed, 2 skipped, 15 xfailed**.
  Half legs as CI runs them (`--xfail-known-failures --known-failure-profile=cpu-float16`): the two
  `TestQuaternionExpToLog::test_back_and_forth` failures are pre-existing manifest entries, **no XPASS**, no
  manifest edit needed. Dynamo pins with `KORNIA_TEST_OPTIMIZER=inductor` -> 3 passed.
- The guard construction itself is sound: `safe_w.acos()` is evaluated everywhere but `safe_w` is `0` at the
  boundary, so the unselected branch is finite; `torch.where` routes the boundary gradient into the detached
  copy, so autograd never differentiates `acos`/`asin` at `+-1`. That the resulting gradient is `0` rather than
  the true (unbounded) derivative for a *non-unit* boundary input is exactly what your comment says it is, and
  saying it in the test is the right treatment.

Fix the attribution and the two smaller items and this is ready.

ducha-aiki:
Follow-up to the review above: I filed the general form of the mechanism as **#4229**, since it is not specific
to these two functions.

`torch.clamp`'s boundary gradient changed from pass-through to zero between 2.9.1 and 2.14.0, and kornia uses
`clamp` as a gradient guard in front of an unbounded derivative in at least three more places. One of them,
`kornia/feature/matching.py:51` (`dm.clamp(min=0.0).sqrt()`), reproduces as real `nan` gradients from
`_cdist` at a zero distance in `float16` on torch 2.5.1 — and that branch is the half-precision / MPS / export
one, whose CI legs pin 2.9.1.

Nothing changes for this PR: your guard is the version-independent construction and is exactly the pattern #4229
recommends. It just means the comment you are correcting has a wider audience than these two call sites, which is
one more reason to get the attribution right.

ducha-aiki:
Same note as on #4227: #4215 merged onto the `CHANGELOG.md` lines this PR touches, so it went `CONFLICTING` — and
a conflicting PR runs no `pull_request` workflow, so CI would have gone quiet on you too.

Rebased onto `main` (`ca1f2f7b`) and force-pushed to `mohansree14:fix/quaternion-log-euler-nan-gradients` — new
head `8e439b63`. Both CHANGELOG entries kept. `kornia/geometry/conversions.py` and
`tests/geometry/test_conversions.py` are **byte-identical across the rebase** (`619 passed, 2 skipped, 15 xfailed`
on the new head), so every measurement in the review above still describes your code.

Please `git fetch` and reset to `8e439b63` before pushing the review fixes.

ducha-aiki:
@mohansree14, a related follow-up surfaced during a correctness-audit verification: #4237. Would you be interested in taking it in a separate PR after this one?

`quaternion_to_axis_angle` returns the wrong derivative at the negative unit identity `(-1, 0, 0, 0)`: autograd gives `+2` for the vector components, while central differences give `-2`. Scaled identities also have the wrong magnitude (`w=2` gives `+2` instead of `+1`). The zero-vector branch hardcodes a factor of `2`, although the surrounding map's limit is `2/w` for nonzero `w`.

I checked this PR's head `8e439b63`: that function is unchanged and the negative-identity reproduction still fails. Your work here is on different functions, so this is a separate follow-up, **not an additional requirement for merging this PR**. #4237 has the reproduction and verification details. The regression tests should check the numerical Jacobian for both identity signs and scaled identities, rather than finiteness alone.

mohansree14:
All three addressed in `d5dd1f0`.

**1. `clamp`, not `acos`.** Your table is right and I reproduced the relevant column locally on torch 2.7.0 (the version I have): `acos` grad at `w = 1` is `-inf`, `clamp(w, -1, 1)` grad at `w = 1` is `1.0`, and the unguarded `(v * clamp(w).acos() / n)` gradient at the identity is `nan` — the pre-2.14 behaviour, so 2.7.0 sits with 2.5.1 and 2.9.1 rather than with 2.14. Both docstring notes and both inline comments now say the boundary derivative of `acos`/`asin` is unbounded on every supported version and did not change, and that torch 2.14 masks the defect through `clamp`'s backward returning `0` at the closed boundary. The PR body's "wrinkle" paragraph is rewritten the same way and points at #4229. Both test comments now say the 2.5.1 leg is the one that discriminates, and that the test passes on base under 2.14 for that reason.

**2. `Tracked in #4007`.** Both citations reworded to "Previously tracked in", matching #4184's accepted form. The third hit at `conversions.py:1383` is a cross-reference in a comment rather than a tracking note, but it read the same way under a grep gate, so it is now "(previously kornia#4007)" too. No bare "Tracked in #4007" survives in `kornia/`.

**3. `near_boundary`.** Correct — it was byte-identical to `non_unit` and exactly on the boundary, so the off-boundary arm of the `torch.where` was never exercised and the comment described coverage that did not exist. It is now `w = 1 - eps` with the same nonzero vector part, asserted against the unguarded expression (mirroring the float32 work dtype for the half types) rather than against zeros. The identity check keeps its own all-zeros assertion, and the comment now says what each of the two is: a correctness check at the boundary, and the inertness check off it.

`-k 4007` passes at float32, float64, float16 and bfloat16 on CPU; `ruff check` and `ruff format --check` are clean on both files. I could not run the full `tests/geometry/test_conversions.py` at float64 locally — it dies with `Fatal Python error: Illegal instruction` in my torch build, and it does so identically on the branch without this commit, so it is my environment rather than the change. The float32 leg of the full file runs clean.

mohansree14:
@ducha-aiki CI is green, 40/40, on `bea6f7d`. All three points from your review are addressed in `d5dd1f0` — the attribution to `clamp` in both notes, both inline comments and the PR body; both `Tracked in #4007` citations reworded; and `near_boundary` moved off the boundary to `w = 1 - eps` and asserted against the unguarded expression. Details in the comment above.

The two red macOS/2.5.1 legs on the previous head were `test_to_onnx_exports[RRDBNetBuilder]` hitting the 120s pytest-timeout, not this change — `bea6f7d` merges `main` so the branch picks up #4402, and both legs pass now.

Ready for re-review when you have a moment.

ducha-aiki (changes requested):
Reviewed exact head `bea6f7deea61316c1896d9feb71f732c9374ef4a`.

The three earlier functional/documentation asks are addressed: clamp is correctly identified as the version-dependent operation, the historical text is retrospective, and the near-boundary pin exercises the unguarded branch away from the boundary. The #4007 cases passed across all four CPU dtypes (8 tests), and hosted checks are green.

[P2] One repository-policy requirement remains. This PR says Fixes #4007, while the prescribed grep still finds #4007/issues/4007 at kornia/geometry/conversions.py:1110,1328,1356 and kornia/geometry/solvers/polynomial_solver.py:192. AGENTS.md explicitly says a surviving hit under kornia/ makes a closing change incomplete and does not exempt retrospective citations. Keep the explanations, but replace production-source issue citations with the delivering PR #4228 or remove them; the tests can keep their issue-named regression pins.

mohansree14:
Thanks — pushed `0927f23` for the remaining policy point.

All four `kornia/` hits now cite this PR instead of the closing issue:

- `kornia/geometry/conversions.py:1110`, `:1328` — the docstring "Previously tracked in #4007" links become "Delivered in #4228", pointing at `/pull/4228`
- `kornia/geometry/conversions.py:1356` — `(previously kornia#4007)` → `(delivered in kornia#4228)`
- `kornia/geometry/solvers/polynomial_solver.py:192` — `(#4007, fixed in #4228)` → `(fixed in #4228)`

The explanations themselves are unchanged, and the regression pins in `tests/` keep their `#4007` names, which the gate allows.

`grep -rnE "#4007|issues/4007" kornia/` now returns nothing, so `Fixes #4007` is clean. `ruff check` and `ruff format --check` pass; the only >120-char lines in `conversions.py` (741-742) are pre-existing and untouched by this diff.

The two Copilot nits were already addressed at `bea6f7d`: the misleading "byte-identical-forward" wording is gone, and `near_boundary` is now genuinely off the boundary at `1.0 - eps`.

mohansree14:
@ducha-aiki this is ready for another look when you have a moment — I can't re-request review from a fork. Thanks!

ducha-aiki (approved):
Re-review of exact head `0927f23875d2db8b09a7204dba3ad90d10d13c48`.

The prior repository-policy blocker is fixed: the four production-source references now cite the delivering PR #4228, and the prescribed `#4007|issues/4007` search leaves only the two regression pins under `tests/`, with no hits under `kornia/`. The functional change is unchanged and remains sound. The focused singularity tests passed in all four CPU dtypes (8 cases) and on MPS float32 (2 cases); all 40 hosted checks are green.

The two outdated Copilot discussions are addressed by the current test wording/input and have now been resolved as part of this re-review. I found no remaining code changes to request.

PR #4230, opened 2026-09-04
fix(ci): let a recorded api_surface.json edit satisfy the import-surface check

### LINKED ISSUE #4190: CI: check_import_surface.py hard-fails on a deliberate __all__ removal, with no path for one whose deprecation window has passed
### Describe the bug

`.github/scripts/check_import_surface.py` exits 1 on **any** name removed from an `__all__`, with only two exemptions (a module-level `__getattr__` deprecation shim, and `kornia.contrib` as the Experimental tier). There is no path for a removal that the stability policy *permits* — one whose deprecation window has already been served.

That path is documented and is expected to be used. `tests/test_api_surface.py::test_no_public_name_removed` spells the procedure out in its own assertion message:

> "If this removal is deliberate and the deprecation window has passed, update `tests/api_surface.json` in this PR (run `tests.test_api_surface.regenerate()`) and list the removal in the release notes."

So a contributor can follow the repository's written procedure exactly, satisfy the test that names it, and still be hard-blocked by a second check that has no notion of the procedure existing.

### Reproduction steps

Live case: #4143 drops `AUTUMN` from `kornia.color.__all__`. The class was deprecated in 0.7.2 in favour of `ColorMap(base='autumn')` and removed by #3432 (`9e521cac`, 2025-12-31), which left the `__all__` entry behind — so `from kornia.color import AUTUMN` has raised `AttributeError` for eight months and `from kornia.color import *` has been broken on it. The PR does both steps the assertion message asks for: the `tests/api_surface.json` edit and a `## Unreleased` release-note entry.

```text
Check for dropped module-level names
##[error]Removed from __all__: ['AUTUMN']
##[error]Process completed with exit code 1.
```

`test_no_public_name_removed` passes (the inventory was updated); the workflow does not.

### Expected behavior

A deliberate, policy-compliant removal should be able to pass the check while still being loud in review. Options, roughly in order of preference:

1. **Treat the same-PR `tests/api_surface.json` edit as the acknowledgement.** The script already diffs against the base ref, so it can ask whether the name was removed from the inventory in the same diff, and downgrade to a report when it was. That reuses the deliberate-review moment the inventory exists to create instead of inventing a second one — and it cannot be satisfied by accident, since the inventory edit is the thing a reviewer looks at.
2. An explicit allowlist file (`.github/import-surface-allow.txt`, `module:name` per line, entries required to cite a PR), pruned when it goes stale.
3. A repository label (`api-removal-approved`) read by the workflow. Weakest of the three: it leaves no record in the tree.

Whichever we pick, the message should name the escape hatch, the way the pytest assertion does. Right now it says only what was removed.

Not urgent for correctness — the check is doing its job, loudly — but it blocks a contributor PR today and it will block the next one, and telling contributors "ignore that red check, a maintainer will override it" is how a gate stops being trusted.

### Environment

Not applicable — CI tooling. Introduced with `.github/workflows/import-surface.yml`; the exemption list is documented in the script's module docstring.

ducha-aiki:

`.github/scripts/check_import_surface.py` exits 1 on **any** name removed from an `__all__` under `kornia/`, with
only two exemptions: a module-level `__getattr__` deprecation shim, and `kornia.contrib` as the Experimental tier.
It has no notion of the removal the stability policy *permits* — one whose deprecation window has already passed —
even though `tests/test_api_surface.py::test_no_public_name_removed` spells that procedure out in its own
assertion message:

> "If this removal is deliberate and the deprecation window has passed, update `tests/api_surface.json` in this PR
> (run `tests.test_api_surface.regenerate()`) and list the removal in the release notes."

So a contributor can follow the repository's written procedure exactly, satisfy the test that names it, and still
be hard-blocked by a second check. That is #4143's situation today.

This implements option 1 from #4190: **treat the same-change `tests/api_surface.json` edit as the
acknowledgement.** A name removed from an `__all__` is reported as a `::notice` instead of an `::error` when the
same diff also drops that name from the inventory entry for that module — or for one of its ancestor packages,
since the inventory records the stable-core top-level packages and a submodule's removal is recorded there. The
error message now names the escape hatch, the way the pytest assertion does.

That reuses the deliberate review moment the inventory exists to create instead of inventing a second one, and it
cannot be satisfied by accident: the inventory edit is exactly what a reviewer looks at.

### The hatch stays narrow

Each of these is a test, and each was also run against the real `#4143` branch:

| Situation | Result |
| --- | --- |
| `__all__` removal **recorded** in `tests/api_surface.json` | `::notice`, exit 0 |
| `__all__` removal **not** recorded | `::error`, exit 1 |
| Some *other* name dropped from the inventory | `::error`, exit 1 — the name and the module path both have to line up |
| Two names removed, only one recorded | `::error` for the unrecorded one, exit 1 |
| Inventory **deleted** at HEAD | acknowledges nothing, exit 1 |
| Inventory **corrupt** or wrong-shaped at HEAD | acknowledges nothing, exit 1 |
| One **entry** wrong-shaped (`"kornia.color": {}`) | acknowledges nothing, **anywhere** in the inventory, exit 1 |
| An entry's list holds a **non-string** | acknowledges nothing, exit 1 |
| Removal recorded for a package the change does not touch | `::error`, exit 1 — the acknowledgement cannot be staged in an earlier PR |
| Ancestor entry's name claimed by a module the package never re-exported it from | not acknowledged, exit 1 |
| A tracked module's **key deleted** | acknowledges nothing, **and the deletion itself is fatal**, exit 1 |
| Inventory emptied to `{}` (parses, shape-test passes, parametrizes nothing) | every tracked module dropped, exit 1 |
| A tracked module's list **emptied** (`[]`) | every name acknowledged, exit 0 — this is how you record losing them all |

The last group is the trap worth calling out: anything that makes a recorded name *look* removed without anyone
recording it would open the hatch for a whole module — or, for an unreadable file, the entire public surface — at
once. `_parse_inventory` returns `None` (distinct from an empty mapping) precisely so the caller can refuse to
acknowledge anything, and it rejects the **whole** inventory rather than skipping a bad entry: a skipped entry is
indistinguishable downstream from a module that lost every name.

Losing a *module* from the inventory is the mirror image, and refusing to acknowledge it is not enough — the
deletion itself is fatal. That key is what puts the module in `test_no_public_name_removed`'s parametrization, so
dropping it ends that guard instead of failing it, and this check cannot stand in: a package whose public names
arrive through `from .sub import *` (`kornia.geometry`, `kornia.morphology`) has no static `__all__` to compare.
The two holes line up exactly. `[]` remains the way to record losing every public name while keeping the module
tracked; removing a stable-core module itself has no acknowledgement path on purpose, because that is a policy
decision, not a routine recorded removal.

Fixes #4190. Unblocks **#4143**, which is otherwise 36 checks green with one red, and that red is this check.

## What did you check?

- **The live case, A/B against `main`'s script on #4143's own branch (`841bccde`, merge-base `4b69473a`):**

  ```text
  main's script:  ::error file=kornia/color/__init__.py::Removed from __all__: ['AUTUMN']     exit=1
  this branch:    ::notice file=kornia/color/__init__.py::Removed from __all__ and recorded
                  in tests/api_surface.json in this same change ...: ['AUTUMN']               exit=0
  ```

  and with that PR's `tests/api_surface.json` edit reverted in the worktree, `exit=1` again.

- `pixi run test-module tests/scripts/test_check_import_surface.py tests/test_api_surface.py` → **56 passed**
  (43 in the script's own suite, 21 of them new; 13 in `test_api_surface.py`, 1 of them new). Six of the seven
  tests added in `44ad8854` fail against `4deeb0d4`; the seventh (`[]` still records every name) is the invariant
  that must not move.
- `tests/test_api_surface.py` also asserts the inventory's shape directly now, so its own removal test cannot pass
  vacuously on an entry mangled to `{}` — verified by mangling `kornia.color` in the real inventory: the removal
  test passed, the new shape test failed.
- `uvx pre-commit run --files <changed>` → all hooks pass; `uvx ty check` on the script → clean.
- `python3 .github/scripts/check_import_surface.py --base-ref origin/main` on this branch → exit 0 (it changes no
  `kornia/` module).

## Notes for the reviewer

- `check_file` gains an optional third parameter, so its existing two-argument callers (including every test in
  `tests/scripts/`) keep working unchanged and simply acknowledge nothing.
- `AGENTS.md`'s `tests/api_surface.json` bullet now says the CI check reads the same edit, so the two halves of
  the procedure are documented in one place.
- The `Import Surface` workflow itself is unchanged: it still triggers on `kornia/**/*.py`, and a removal always
  touches one of those.

### Review round 2 (`44ad8854`)

Both the Copilot inline finding and the P1 in
[the review](https://github.com/kornia/kornia/pull/4230#pullrequestreview-5120593715) were real, and I reproduced
all three variants against `4deeb0d4` before touching anything (each exited **0** where it must exit 1):

| Bypass at `4deeb0d4` | Before | After |
| --- | --- | --- |
| `"kornia.mymodule": {}` while dropping `a` from `__all__` | exit 0 | exit 1 |
| module key deleted while dropping `a` | exit 0 | exit 1 |
| `"kornia.mymodule": [1, 2]` while dropping `a` | exit 0 | exit 1 |
| honest record (control) | exit 0 | exit 0 |
| nothing recorded (control) | exit 1 | exit 1 |

Same A/B on the live #4143 branch (`841bccde`, merge-base `4b69473a`): the real inventory edit still exits 0 with
the `::notice`, and the same branch with `kornia.color` deleted or set to `{}` now exits 1.

### Review round 3 (`426df44b`)

The [round-2 review](https://github.com/kornia/kornia/pull/4230#pullrequestreview-5120676085)'s P1 was real, and
the bypass it pointed at is worse than a standalone key deletion. I reproduced four shapes against `44ad8854`
before changing anything:

| At `44ad8854` | Before | After |
| --- | --- | --- |
| `from .bbox import *` dropped from a package `__init__` **+** that package's inventory key deleted | **exit 0, silent** | exit 1 |
| inventory key deleted, no `kornia/` change at all | **exit 0** | exit 1 |
| inventory key deleted alongside an unrelated `kornia/` edit | **exit 0** | exit 1 |
| `from .bbox import *` dropped, inventory untouched | exit 0 | exit 0 — still guarded by `test_no_public_name_removed[kornia.geometry]`, which is the point |

The first row is the one that matters: `from .sub import *` is deliberately invisible to `parse_module_surface`,
so dropping a package's wildcard re-export produces no report here at all, and deleting the same package's
inventory key took the runtime guard away too. `kornia.geometry` and `kornia.morphology` are shaped exactly like
that.

`untracked_modules(base_ref)` now reports every module the base inventory tracks that the working tree's no
longer does, and `main` hard-fails on it. That also closes `{}` — an inventory emptied to an empty object parses,
passes the new shape test, and parametrizes `test_no_public_name_removed` with zero modules, so nothing else
notices. An inventory that is unreadable at the working tree tracks nothing either, so it reports every module,
under a message that names the real cause rather than reading as twelve deliberate deletions. `[]` still records
losing every public name and keeps the module tracked — pinned, unchanged.

The workflow's `paths:` filter gains `tests/api_surface.json`, so an inventory-only change runs the check; without
it the standalone deletion above never triggered the workflow at all.

**Verification at `426df44b`:**

- `pixi run test-module tests/scripts/test_check_import_surface.py tests/test_api_surface.py` → **64 passed**
  (was 56). Eight new tests; the two `main()` ones both fail against `44ad8854` with `assert 0 == 1`, checked in a
  worktree at that commit.
- Live #4143 A/B (`841bccde`, merge-base `4b69473a`) with this branch's script: the honest recorded removal still
  exits **0** with the `::notice`; `kornia.color` **deleted** → exit 1; set to `{}` → exit 1; the whole inventory
  emptied to `{}` → exit 1; the file removed → exit 1; the inventory edit reverted → exit 1.
- `uvx pre-commit run --files <changed>` all pass; `uvx ty check` on the script clean;
  `python3 .github/scripts/check_import_surface.py --base-ref origin/main` on this branch exits 0.

### Review round 4 (`d2008893`)

Three findings from the [round-3 review](https://github.com/kornia/kornia/pull/4230#pullrequestreview-5120776344),
all reproduced against `426df44b` first.

**Provenance (P1).** The ancestor match was on spelling alone, so a package entry losing `thing` excused every
`thing` anywhere beneath it — including an unrelated symbol in a sibling submodule the package never re-exported.
This one is a regression this branch introduced: `main` fails that removal, `426df44b` exits 0 and prints a
"recorded" notice. `reexport_sources` now follows the package's own re-export of the name at the base revision —
explicit imports, relative or absolute (`from kornia.augmentation._2d import CenterCrop` is how 86 of that
package's recorded names arrive), and a wildcard resolved through the source module's own exported surface — and
the claiming module has to sit under one of those sources. A name that traces to nothing falls back to the plain
ancestor match; a wrong "no" there would hard-block a policy-compliant removal, which is the defect #4190 is
about. On the real inventory, **566 of 717** names resolve to a single source module, 12 to two (`Se2`/`Se3`/
`So2`/`So3` and friends, genuinely re-exported twice), and 139 trace nowhere.

**Staging (P1).** Dropping a name from the inventory in one PR that touches no Python, then removing the API in a
later one, passed both halves — the second is invisible when the name leaves through a `from .sub import *`
re-export. An inventory removal for a package this change does not otherwise touch is now fatal. This one is not a
regression: `main` exits 0 on the same PR A.

**Impossible edit (P2).** Confirmed, and the audit reproduces exactly: **40 modules, 132 `__all__` names** under a
tracked package that no ancestor entry records (**84 / 323** across all of `kornia/`), including
`kornia.geometry.boxes.Boxes`, `kornia.augmentation.auto.AutoAugment`, `kornia.metrics.angle_error_mat`, and all of
`kornia.feature`. Making the inventory module-level is a redesign of what the file records, so it is filed as
**#4236**; here the fatal message stops prescribing an edit that cannot be made and points there instead.

**Changelog (P3).** Now `(#4190, #4230)`, and the "fails closed everywhere else" sentence is narrowed to what the
check actually enforces.

**Verification at `d2008893`:**

- `pixi run test-module tests/scripts/test_check_import_surface.py tests/test_api_surface.py` → **72 passed** (was
  64). Eight new tests; the three that discriminate all fail against `426df44b`
  (`assert {'thing'} == set()`, `assert 0 == 1`, and the old message), while
  `test_check_file_ancestor_entry_still_excuses_its_real_source` passes there — that is the invariant that must not
  move.
- Live #4143 (`841bccde`, merge-base `4b69473a`) still exits **0** with the `::notice`. A real-repo wildcard case —
  `bbox_to_mask` dropped from `kornia/geometry/bbox.py`'s `__all__` and from the `kornia.geometry` entry — also
  exits 0, so the hatch still works for the package the ancestor walk exists to serve.
- Every earlier round's bypass re-checked: all still exit 1.
- `uvx pre-commit run --files <changed>` all pass; `uvx ty check` on the script clean;
  `python3 .github/scripts/check_import_surface.py --base-ref origin/main` on this branch exits 0.


ducha-aiki (commented):
Found one blocking correctness issue in the new escape hatch.

**P1 — validate every inventory entry before computing removals.** `_parse_inventory()` rejects a wrong-shaped top-level value, but silently drops a module whose value is not a list and silently filters non-string list members. `inventory_removals()` then treats the missing/emptied entry as an empty set, so every old name in that module looks deliberately removed. A concrete bypass is to change `"kornia.color": [...]` to `"kornia.color": {}` while removing `AUTUMN` from `__all__`: the import-surface check acknowledges `AUTUMN` and exits 0, while `tests/test_api_surface.py` also accepts `set({}) == set()`. This confirms the existing Copilot inline finding remains blocking.

The same fail-open behavior exists when an inventory module key is deleted outright: all its old names are treated as acknowledged, that stable module disappears from `test_no_public_name_removed` parametrization, and future removals in it are no longer guarded. Please reject any non-`str -> list[str]` entry and reject removal of tracked module keys (an empty list can represent an intentional removal of every name), with focused tests for both cases.

Validation performed at `4deeb0d4`: `pixi run test-module tests/scripts/test_check_import_surface.py tests/test_api_surface.py` (48 passed), plus `git diff --check`.

.github/scripts/check_import_surface.py
```diff
@@ -237,6 +245,80 @@ def _alternate_path(path: str) -> str:
...
+        data = json.loads(source)
+    except (ValueError, UnicodeDecodeError):
+        return None
+    if not isinstance(data, dict):
+        return None
+    return {
+        module: {name for name in names if isinstance(name, str)}
+        for module, names in data.items()
+        if isinstance(module, str) and isinstance(names, list)
+    }
```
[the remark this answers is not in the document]
ducha-aiki: Confirmed and fixed in `44ad8854`.

Reproduced first against `4deeb0d4`: `"kornia.mymodule": {}` while dropping `a` from `__all__` exited **0** with the acknowledgement `::notice`, and so did a list holding non-strings (`[1, 2]`). Skipping the bad entry dropped the module from the parsed mapping, and a missing module reads in `inventory_removals` exactly like one that lost every recorded name.

`_parse_inventory` now returns `None` unless the file is exactly `dict[str, list[str]]` — one bad entry rejects the whole inventory, since a file that cannot be trusted at one key cannot be trusted to speak for another. Both variants now exit 1, including on the live #4143 branch with `kornia.color` mangled. Four new tests cover the wrong-shaped value, the non-string member, the one-bad-entry-poisons-the-rest case, and the end-to-end exit code.

ducha-aiki:
**P1 fixed in `44ad8854`.** Both halves were real, and I reproduced them against `4deeb0d4` before changing anything. Three variants exited **0** where they must exit 1:

| Bypass at `4deeb0d4` | Before | After |
| --- | --- | --- |
| `"kornia.mymodule": {}` while dropping `a` from `__all__` | exit 0 | exit 1 |
| module key deleted while dropping `a` | exit 0 | exit 1 |
| `"kornia.mymodule": [1, 2]` while dropping `a` | exit 0 | exit 1 |
| honest record (control) | exit 0 | exit 0 |
| nothing recorded (control) | exit 1 | exit 1 |

**Wrong-shaped entries.** `_parse_inventory` now returns `None` unless the file is exactly `dict[str, list[str]]`: a non-string key, a non-list value, or a list holding a non-string rejects the **whole** inventory rather than skipping that entry. Rejecting per-file rather than per-entry is deliberate — an inventory that cannot be trusted at one key cannot be trusted to speak for another, and there is a test pinning that an honest edit next to a mangled entry acknowledges nothing.

**Deleted module keys.** `inventory_removals` skips a tracked key the change deletes instead of reading it as "lost every name". An empty list is the recorded-removal-of-everything spelling, and it keeps the module in `test_no_public_name_removed`'s parametrization; there is a test pinning that `[]` still acknowledges every name, so the remedy is not accidental.

One consequence worth naming rather than hiding: removing a **whole** tracked module now has no acknowledgement path, because `test_no_public_name_removed` imports every key so the key has to go with it. That change fails this check on purpose — dropping a stable-core package is a policy decision, not a routine recorded removal — and the code comment says so.

**The pytest half.** Your `set({}) == set()` observation held: mangling `kornia.color` to `{}` in the real `tests/api_surface.json` left `test_no_public_name_removed[kornia.color]` **passing**. `tests/test_api_surface.py` now asserts the inventory's shape directly, and that new test is the only one that fails on the mangled file.

**Verification at `44ad8854`:**

- `pixi run test-module tests/scripts/test_check_import_surface.py tests/test_api_surface.py` → **56 passed** (was 48). Seven new tests; **six of them fail against `4deeb0d4`**, and the seventh (`[]` still records every name) is the invariant that must not move.
- Live #4143 A/B (`841bccde`, merge-base `4b69473a`) with this branch's script: the real inventory edit still exits **0** with the `::notice`; the same branch with `kornia.color` **deleted** exits 1; with `kornia.color` set to `{}` exits 1; with the edit reverted exits 1.
- `uvx pre-commit run --files <changed>` all pass; `uvx ty check` on the script clean; `python3 .github/scripts/check_import_surface.py --base-ref origin/main` on this branch exits 0; `git diff --check` clean.

The PR body's "hatch stays narrow" table now carries these rows too.

ducha-aiki (commented):
The malformed per-entry inventory bypass from the first review is fixed at `44ad8854`, but one blocking case remains.

**P1 — deleting a tracked module key still does not fail the guard.** At `.github/scripts/check_import_surface.py:315-325`, a key present in the base inventory but absent at HEAD merely causes `inventory_removals()` to `continue`. This prevents that deletion from acknowledging a detectable `__all__` removal, but it does not reject the key deletion itself. Meanwhile, `tests/test_api_surface.py:67-75` validates and parametrizes only keys remaining in the new mapping, so the deleted module disappears from coverage.

This can pass immediately and permanently weaken the inventory. In particular, packages such as `kornia.geometry` and `kornia.morphology` have wildcard re-exports that the static parser intentionally ignores: deleting such a re-export together with the package's inventory key leaves no fatal `removed_from_all`, while the runtime API-surface test no longer examines that package. A standalone inventory-key deletion also passes. Please make `old.keys() - new.keys()` a fatal condition (or otherwise preserve the baseline list of tracked modules), add a focused failure test, and include `tests/api_surface.json` in the Import Surface workflow's path filter if this workflow owns that enforcement; currently an inventory-only deletion does not trigger it.

Validation performed: `pixi run test-module tests/scripts/test_check_import_surface.py tests/test_api_surface.py` (56 passed), `git diff origin/main...HEAD --check`, and all current PR checks are green.

ducha-aiki:
**P1 fixed in `426df44b`.** Real, and the bypass is worse than the standalone key deletion: I reproduced four shapes against `44ad8854` before touching anything.

| At `44ad8854` | Before | After |
| --- | --- | --- |
| `from .bbox import *` dropped from a package `__init__` **+** that package's inventory key deleted | **exit 0, silent** | exit 1 |
| inventory key deleted, no `kornia/` change at all | **exit 0** | exit 1 |
| inventory key deleted alongside an unrelated `kornia/` edit | **exit 0** | exit 1 |
| `from .bbox import *` dropped, inventory untouched | exit 0 | exit 0 — see below |

The first row is your point exactly. `from .sub import *` is deliberately invisible to `parse_module_surface`, so dropping a package's wildcard re-export produces no report here at all; deleting that package's inventory key in the same change took the runtime guard away too, and the pair exited 0 in silence. `kornia.geometry` and `kornia.morphology` are shaped like that.

The last row stays 0 on purpose and is not a hole: the key is still there, so `test_no_public_name_removed[kornia.geometry]` catches it at runtime. That division of labor is the reason the key deletion has to be fatal — the static check genuinely cannot see through a wildcard, so it can only defend the thing that can.

**What changed.** `untracked_modules(base_ref)` returns every module the base inventory tracks that the working tree's no longer does, and `main` hard-fails on it. `inventory_removals` still `continue`s over a deleted key, unchanged — refusing to acknowledge is what stops the same edit from excusing an `__all__` removal on the way out; the new check is what fails the deletion itself.

Two cases beyond the one you named fall out of `old.keys() - new.keys()` and are pinned:

- **`{}`.** An inventory emptied to an empty object parses, passes the new shape test, and parametrizes `test_no_public_name_removed` with zero modules — nothing else notices. It now drops every tracked module and exits 1.
- **Unreadable at HEAD** (absent, unparsable, wrong-shaped anywhere). It tracks nothing, so it reports every module, under a distinct message naming the real cause rather than reading as twelve deliberate deletions.

`[]` still records losing every public name while keeping the module tracked, and that invariant is pinned separately so the remedy the error message points at cannot drift.

**Path filter.** `tests/api_surface.json` added to the workflow's `paths:`. You were right that it was load-bearing: without it the standalone key deletion touches no `kornia/**/*.py` and the workflow never ran.

**Verification at `426df44b`:**

- `pixi run test-module tests/scripts/test_check_import_surface.py tests/test_api_surface.py` → **64 passed** (was 56). Eight new tests; the two `main()` ones both fail against `44ad8854` with `assert 0 == 1`, checked in a worktree at that commit.
- Live #4143 A/B (`841bccde`, merge-base `4b69473a`): honest recorded removal still exits **0** with the `::notice`; `kornia.color` deleted → 1; `kornia.color` set to `{}` → 1; whole inventory `{}` → 1; file removed → 1; inventory edit reverted → 1.
- `uvx pre-commit run --files <changed>` all pass; `uvx ty check` on the script clean; `python3 .github/scripts/check_import_surface.py --base-ref origin/main` on this branch exits 0.

The PR body's "hatch stays narrow" table and its trailing paragraph now say this too — the old paragraph claimed a whole-module deletion "keeps failing here", which was only true when the module had a static `__all__` to lose.

ducha-aiki (commented):
I completed a full re-review of the entire PR at `426df44b`, including the original design, all follow-up commits, workflow integration, adversarial inventory states, actual repository export patterns, tests, and documentation. The two findings from the previous rounds are fixed, but four actionable issues remain.

**P1 — inventory removals can be staged in an earlier PR, so the check does not enforce “same change”.** At `.github/scripts/check_import_surface.py:493-494`, `main()` uses `inventory_removals()` only to classify changed Python-file reports; it never rejects or even reports inventory-name removals that match no source removal in the current diff. I reproduced a two-PR bypass: PR A changes `{"kornia.pkg": ["a"]}` to `{"kornia.pkg": []}` without touching Python. The newly added inventory workflow trigger runs, but the checker exits 0 silently, and `test_no_public_name_removed` also passes because the runtime surface is a superset of the empty baseline. After that lands, PR B removes `from .sub import *`; the checker again exits 0 silently because wildcard imports are intentionally ignored. This removes the deliberate review moment from the actual API-removal PR and applies directly to packages such as `kornia.geometry` and `kornia.morphology`. Reject unmatched inventory-name removals, or otherwise prove that each inventory deletion corresponds to a source-surface removal in the same diff, and add the two-stage/inventory-only regression test.

**P1 — ancestor matching authorizes unrelated same-named descendant removals.** At `.github/scripts/check_import_surface.py:381-385`, `acknowledged_names()` unions an ancestor package's removed names into every descendant based only on the dotted path and spelling. Reproduction: top-level `kornia.pkg` exports `thing` from `pkg.a`, while `pkg.b.__all__` independently exports another `thing`; removing the top-level export and `kornia.pkg:thing` from the inventory also silently authorizes removing `pkg.b.thing`. The checker exits 0 and prints “recorded” notices for both even though the inventory edit represents only the top-level API. The tests at `tests/scripts/test_check_import_surface.py:590-602` exclude another top-level package, but do not cover sibling/descendant collisions. The acknowledgement needs exact module/symbol provenance rather than granting one ancestor name to every descendant.

**P2 — the escape hatch is unavailable for real documented APIs absent from the inventory.** The logic at `.github/scripts/check_import_surface.py:355-385` can acknowledge only names present in the base inventory, but that inventory contains aggregate entries for only twelve top-level modules and is not exhaustive. A repository audit found 40 modules containing 132 `__all__` names absent from every tracked ancestor. This includes unquestionably public, rendered stable APIs: `kornia.geometry.boxes.Boxes` (`kornia/geometry/boxes.py:29`, documented in `docs/source/geometry.boxes.rst:7-12`), `kornia.augmentation.auto.AutoAugment`, and `kornia.metrics.angle_error_mat`. Removing one after its deprecation window still exits 1 and tells the contributor to drop a name that does not exist in the base inventory; adding a new inventory entry in the removal PR cannot help because `inventory_removals()` iterates only the old mapping. `kornia.feature` is not inventoried at all despite the stability policy covering its classical APIs. Make the inventory exhaustive at the actual module/symbol granularity or use an explicit acknowledgement keyed by the removed module and symbol.

**P3 — the changelog cites the issue but not this PR.** `CHANGELOG.md:252-265` ends with `#4190`, while the repository instructions require user-visible Unreleased entries to cite the PR number. Add `#4230`; the “fails closed everywhere else” wording should also be narrowed after resolving the bypasses above.

Validation performed:

- `pixi run test-module tests/scripts/test_check_import_surface.py tests/test_api_surface.py`: 64 passed in a clean worktree.
- `pixi run pre-commit-all`: all hooks passed.
- `git diff origin/main...origin/pr-4230 --check`: clean.
- All current hosted PR checks are green.
- Verified the earlier malformed-entry, malformed-list, deleted-key, empty-object, and missing-inventory cases are now closed.
- Checked merge-base/working-tree consistency, workflow path triggers and permissions, parser failure handling, rename/packageification behavior, GitHub annotation formatting, and subprocess use; no additional issues found there.

ducha-aiki:
Round 3 fixed in `d2008893`. Three of the four findings hold and are fixed; P2 is real but its fix is a redesign, so it is filed rather than done here. Each reproduced against `426df44b` before I touched anything.

### P1 — ancestor matching authorizes same-named descendants

Confirmed, and worth flagging that this one is a **regression this branch introduced**, not a pre-existing gap: `main` fails that removal, `426df44b` exits 0 *and* prints a "recorded" notice for it.

```
kornia/pkg/__init__.py: from .a import thing      # the package's `thing` is a.thing
kornia/pkg/b.py:        __all__ = ['thing', ...]  # an unrelated symbol of the same name
```
Dropping the top-level export plus `kornia.pkg: thing` from the inventory also excused removing `pkg.b.thing`.

`reexport_sources(base_ref, package, name)` now follows the package's own re-export of that name at the base revision, and the claiming module has to sit under one of the sources it returns:

- explicit imports, **relative or absolute** — `from kornia.augmentation._2d import CenterCrop` is how 86 of that package's recorded names arrive, and skipping level-0 imports left them all unresolved;
- `from .a import *` resolved through `a`'s own exported surface (`__all__` if it has one, else module scope);
- a source covers its whole subtree, which is what carries a nested re-export (`kornia.geometry` → `.transform` → `transform/affwarp.py`) without walking it.

A name that traces to nothing returns `None` and falls back to the plain ancestor match. That fallback is deliberate and I would push back on removing it: a wrong "no" hard-blocks a policy-compliant removal, which is the exact defect #4190 is about. `kornia.color`'s `AUTUMN` is the live example — it is in `__all__` and bound by no import in `kornia/color/__init__.py`, so it traces nowhere, and #4143 is precisely the PR that removes it.

Measured over the whole real inventory at `ca1f2f7b`: **566 of 717** names resolve to a single source module, **12** to two (`Se2`/`Se3`/`So2`/`So3`, `convert_points_*`, `transform_points`, `normalize_points` — genuinely re-exported through both `liegroup` and `pose`, so both are legitimate claimants), **139** trace nowhere and keep today's behavior. Deciding the 12 by Python's last-wins wildcard ordering would be more precise and also more brittle — a module that re-exports the same object from its sibling would then be wrongly denied — so ambiguity keeps every candidate.

### P1 — staged inventory removals

Confirmed. Not a regression (`main` exits 0 on the same PR A), but the check owns the inventory now, so it is in scope. A name dropped from an entry whose package the change does not otherwise touch is fatal.

The rule is "the inventory edit travels with a change to that package", not "a matching source removal exists", and that is on purpose: removing a `from .sub import *` line from a package `__init__` produces no detectable removal at all, so requiring a match would hard-block exactly the legitimate case the last row of the round-3 table describes. It closes the reported repro (PR A touches no Python) and leaves the weaker "PR A also touches something under the package" shape, which puts an unexplained inventory deletion in a diff with the code — a reviewable artifact rather than a silent one. Worth saying plainly rather than claiming more.

### P2 — names the inventory never records

Confirmed, and your audit reproduces exactly: **40 modules / 132 `__all__` names** under a tracked package with no ancestor entry, **84 / 323** across all of `kornia/` (excluding `kornia.contrib`, exempt as Experimental). `Boxes`, `Boxes3D`, `VideoBoxes`, `AutoAugment`, `angle_error_mat` and all of `kornia.feature` are in there.

Making the inventory module-level changes what the file records and how `regenerate()` builds it, which is a bigger decision than this PR — filed as **#4236** with both options and the numbers. What this PR does fix is the misdirection: when the removed names appear in no tracked ancestor entry at base, the error no longer tells the contributor to drop a row that does not exist. It says the inventory has no entry for them, points at the deprecation-shim path, and links #4236.

### P3 — changelog

`(#4190, #4230)`. The "fails closed everywhere else" sentence is narrowed to what the check enforces, and the three new fatal conditions are spelled out.

### Verification at `d2008893`

- `pixi run test-module tests/scripts/test_check_import_surface.py tests/test_api_surface.py` → **72 passed** (was 64). Eight new tests; the three that discriminate fail against `426df44b` — `assert {'thing'} == set()`, `assert 0 == 1`, and the old message text — while `test_check_file_ancestor_entry_still_excuses_its_real_source` passes there, which is the invariant that must not move.
- Live #4143 (`841bccde`, merge-base `4b69473a`) still exits **0** with the `::notice`. Real-repo ancestor case: `bbox_to_mask` dropped from `kornia/geometry/bbox.py`'s `__all__` and from the `kornia.geometry` entry → exit 0, so the wildcard package the ancestor walk exists for still works.
- Every earlier round's bypass re-checked at `d2008893`: all still exit 1.
- `uvx pre-commit run --files <changed>` all pass; `uvx ty check` on the script clean; `python3 .github/scripts/check_import_surface.py --base-ref origin/main` on this branch exits 0.

ducha-aiki (commented):
I completed another full review of the entire PR at `d2008893`, including all prior fixes plus fresh adversarial probes of the new provenance and same-change logic. The previous malformed-inventory, deleted-key, pure inventory-only staging, simple sibling-collision, and changelog-reference findings are addressed. Four actionable issues remain.

**P1 — “same change” is still satisfied by any unrelated Python touch under the package.** At `.github/scripts/check_import_surface.py:618-622`, an inventory removal is considered accompanied when any changed path equals or lies beneath the inventory module; the code never proves that the recorded name left a source/export surface. I reproduced a complete two-PR bypass with `kornia.pkg` exporting `thing` through `from .a import *`. PR A changes the inventory from `["thing"]` to `[]` and merely appends an unrelated comment to `pkg/b.py`; the checker exits 0 with no output. After committing A, PR B removes the wildcard import from `pkg/__init__.py`; the checker again exits 0 with no output because the inventory no longer records `thing` and wildcard imports are invisible to `parse_module_surface`. This contradicts the same-change guarantee in the error message, CHANGELOG, and AGENTS.md. The test at `tests/scripts/test_check_import_surface.py:991-1004` covers only the no-Python-change case. Require a corresponding export/name delta, or use a dedicated exact `(module, name)` acknowledgement rather than `any(path under package)`.

**P1 — provenance still authorizes unrelated same-named descendants.** At `.github/scripts/check_import_surface.py:477-479`, a resolved source authorizes its entire subtree with `module.startswith(src + ".")`. Reproduction: `pkg/__init__.py` has `from .a import *`; `a/__init__.py` imports `thing` from `a.real`; both `a.real` and unrelated, non-reexported `a.unrelated` independently declare `__all__ = ["thing"]`. Removing both and dropping only `kornia.pkg:thing` makes `reexport_sources()` return `{"kornia.pkg.a"}`; both removals are reported as recorded and the checker exits 0. Nested wildcard chains are looser still: `_exported_surface()` does not recursively resolve them, so `reexport_sources()` returns `None` and lines 405-409/477-479 deliberately fall back to the original spelling-only ancestor match. The current repository has 139 inventory names on that fallback path. Additionally, `ast.walk(tree)` at line 421 treats imports inside functions, classes, and `TYPE_CHECKING` blocks as package reexports and ignores runtime binding order. Resolve provenance recursively at module scope to the actual leaf binding; unresolved provenance should not grant an ancestor acknowledgement, and a source package should not authorize every module below it.

**P2 — the inventory-coverage gap remains unresolved, not fixed.** `inventory_removals()` at `.github/scripts/check_import_surface.py:356-369` can acknowledge only names already present in the base inventory. The audit still finds 40 modules and 132 `__all__` names beneath tracked packages with no recorded ancestor entry (84 modules and 323 names across non-contrib Kornia), including documented stable APIs such as `kornia.geometry.boxes.Boxes`, `kornia.geometry.vector.Vector2`, and `kornia.metrics.angle_error_mat`; `kornia.feature` is absent entirely despite classical feature APIs being stable-core. Lines 661-669 now give an honest message and #4236 tracks the redesign, which is an improvement, but a policy-compliant removal of one of these APIs still has no acknowledgement route. This remains a functional limitation maintainers must explicitly accept if it is deferred from this PR.

**P3 — mixed tracked/untracked removals receive an impossible remedy.** At `.github/scripts/check_import_surface.py:646`, the tracked-removal branch is selected when *any* unrecorded name intersects `_tracked_names`, but lines 648-654 print the entire `unrecorded` list and instruct the contributor to drop every name from the inventory. If one file removes `tracked` and `orphan` together, `orphan` has no inventory row and cannot be removed as instructed. The new test at `tests/scripts/test_check_import_surface.py:1007-1020` covers only an all-untracked set. Partition the names and emit the inventory remedy for tracked names and the #4236/no-recorded-path message for untracked names.

Validation and additional review coverage:

- `pixi run test-module tests/scripts/test_check_import_surface.py tests/test_api_surface.py`: 72 passed in a clean worktree.
- `pixi run pre-commit-all`: all hooks passed.
- `python3 .github/scripts/check_import_surface.py --base-ref origin/main`: exit 0 on the PR itself.
- `git diff origin/main...origin/pr-4230 --check`: clean.
- All current hosted PR checks are green.
- Rechecked malformed JSON/entry/list shapes, missing inventory, deleted keys, empty objects/lists, workflow paths and permissions, merge-base and working-tree behavior, rename/packageification handling, annotations, and subprocess use; no additional issues found there.

.github/scripts/check_import_surface.py
```diff
@@ -237,6 +254,341 @@ def _alternate_path(path: str) -> str:
...
+                # torch/typing/__future__ would make ordinary Kornia
+                # modules unknowable for reasons unrelated to their own
+                # re-export provenance.
+                for alias in node.names:
+                    if alias.name == "*":
+                        imported = self.resolve(target) if target == "kornia" or target.startswith("kornia.") else None
+                        if imported is None:
+                            result = None
+                            break
+                        bindings.update(imported)
```
ducha-aiki: [P2] Include implicit parent-package submodule bindings

Python adds `a` to `kornia.pkg` when its initializer executes `from .a import *`, but this resolver adds only the names exported by `a`. Reproduction: base initializer `from .a import *`, leaf `a.py` exports `thing`, and package inventory `["a", "thing"]`. Delete the leaf and import, update the inventory to `[]`, and add the exact `kornia.pkg.a:thing` acknowledgement. Runtime package names correctly change from `["a", "thing"]` to `[]`, but the checker exits 1 with an unmatched inventory removal for `a`. This affects existing inventory entries too: the resolver misses submodule names such as `kornia.geometry.bbox` and `kornia.morphology.morphology`. Model these import-created package bindings so a statically ordinary, correctly recorded removal can pass.

.github/scripts/check_import_surface.py
```diff
@@ -237,6 +254,341 @@ def _alternate_path(path: str) -> str:
...
+                    break
+            elif isinstance(node, (ast.Assign, ast.AnnAssign)):
+                targets = node.targets if isinstance(node, ast.Assign) else [node.target]
+                for target in targets:
+                    if isinstance(target, ast.Name):
+                        bindings.add(target.id)
+                    else:
+                        result = None
+                        break
+            elif isinstance(node, (ast.Delete, ast.Global, ast.Nonlocal, ast.AugAssign)):
```
ducha-aiki: [P2] Fail closed on unsupported expressions that can create exports

`_bound_exports` silently skips top-level `ast.Expr` nodes, so an unknown dynamic surface can be treated as an empty known surface. Reproduced with a package initially containing `thing = 1` and inventory `["thing"]`: replace the initializer with `globals().update(thing=1)` and set its inventory to `[]`. Runtime `kornia.pkg.thing` still equals 1, but `main()` exits 0 and validates the inventory removal. The runtime inventory test also accepts the emptied list, leaving a later actual removal unguarded. Return unknown for unsupported expressions that can mutate bindings (while allowing inert docstrings), and add a regression test that retains a dynamically bound name while attempting to remove its inventory entry.

ducha-aiki (commented):
Reviewed head `96e60cf61edd2b998b81a9dbd5b9ff1813dffe24`. Two P2 correctness issues remain in the new static export resolver: it rejects legitimate removals of implicit submodule bindings, and accepts inventory staging when a retained name moves to a dynamic expression. Both were reproduced in isolated fixture repositories, with runtime imports checking the actual surface.

Validation: 76 focused tests passed (`tests/scripts/test_check_import_surface.py` and `tests/test_api_surface.py`); the import-surface script exits 0 on this PR; `git diff --check` is clean. No source changes made.

ducha-aiki:
Fixed both findings from the latest review in 32f3da77.

- The resolver now includes the implicit parent-package submodule bindings created by wildcard, explicit, and aliased imports. Correctly recorded submodule removals pass, while a name retained by another import still cannot be dropped from the inventory.
- Unsupported dynamic expressions now produce an unknown surface instead of proving a name absent. This includes calls and assignment expressions nested in assignment values, function defaults/decorators, and class bodies. Deferred function bodies do not make an otherwise static surface unknown.

Added 14 regression/control cases and updated the changelog. The original six reproductions were verified failing before the fix. An independent review of the final changes found no further actionable issues.

Validation:
- `pixi run test-module tests/scripts/test_check_import_surface.py tests/test_api_surface.py`: **90 passed**.
- `pixi run pre-commit-all`: all hooks passed.
- `pixi run uv run ty check .github/scripts/check_import_surface.py`: passed.
- `python3 .github/scripts/check_import_surface.py --base-ref origin/main`: exit 0.
- `git diff --check`: clean.

Pushed to this PR's branch. Hosted CI has not yet been verified for the new commit.

ducha-aiki:
Fixed the linked documentation CI failure in 1c34cff4.

The failing step was the Sphinx HTML build: `docs/source/get-started/stability.rst:106` had a section-title underline one character shorter than “Recording a completed deprecation”. Sphinx reported `Title underline too short`, and `-W` made it fatal. Extended the underline by one character.

Reproduced the warning locally before the change. Afterward:
- The affected section parses without warnings.
- `pixi run build-docs`: succeeded.
- `pixi run pre-commit-all`: all hooks passed.
- `git diff --check`: clean.

Pushed the fix; the new hosted CI run is not yet verified.

Issue #4195, opened 2026-09-03
test_normalized_meshgrid_trace_matches_eager_at_unrepresentable_sizes fails 8/10 on cuda float32 (1 ULP vs atol=0)


ducha-aiki:
`tests/geometry/test_grid.py::test_normalized_meshgrid_trace_matches_eager_at_unrepresentable_sizes` fails 8 of its 10 parametrizations on **cuda float32** and passes all 10 on cpu float32. It is not xfailed and it is not marked `device_agnostic`, so it is a live red test on any CUDA job.

## Repro

```bash
pytest tests/geometry/test_grid.py::test_normalized_meshgrid_trace_matches_eager_at_unrepresentable_sizes \
    --device=cuda --dtype=float32
```

## Observed

```
--device=cuda --dtype=float32  -> 8 failed, 2 passed
--device=cpu  --dtype=float32  -> 10 passed
```

The parametrization is `258/300/1000/2050/3000` x `2d/3d`; the two that pass are `258-2d` and `258-3d`.

```
E   Mismatched elements: 92 / 2400 (3.8%)
E   Greatest absolute difference: 1.1920928955078125e-07 at index (0, 150, 0, 1)
E   Greatest relative difference: 3.564426879165694e-05 at index (0, 150, 0, 1)
```

## Cause

`1.1920928955078125e-07` is exactly one ULP of float32 at 1.0, and the test asserts bitwise equality:

```python
traced = torch.jit.trace(MeshGrid(), image)
assert_close(traced(image), MeshGrid()(image), atol=0.0, rtol=0.0)
```

So this looks like ordinary traced-vs-eager kernel/fusion divergence on CUDA, not the rounding bug the test was written to catch. The test's docstring is about `size - 1` being rounded when cast into a narrow coordinate dtype — a bfloat16/float16 concern, pinned at float32/float64 "as agreeing there rather than merely not being compared". The `atol=0.0, rtol=0.0` that makes that pin meaningful is also what makes a 1-ULP CUDA difference fail.

The test was added in #4006 (`4dd9c6aa`, singleton normalization axes) and last touched by #4145 (`4990046f`).

## Expected

One of:

- the float32/float64 arms compare with a dtype-appropriate tolerance while the half arms keep the exact comparison, since only the half arms are the pin's actual subject; or
- the exact comparison is kept and the CUDA divergence is understood and documented — if traced `create_meshgrid` really can differ from eager by a ULP on CUDA, that is worth knowing on its own; or
[...]

PR #4231, opened 2026-09-05
fix(geometry): normalize the half-precision meshgrid in float32

### LINKED ISSUE #4030: create_meshgrid returns a different grid under torch.compile than in eager at float16/bfloat16 (torch.linspace vs its inductor lowering)

LoganChu:

`create_meshgrid` / `create_meshgrid3d` still return a different **normalized** grid under
`torch.compile` than in eager at `float16`/`bfloat16`. #4036 closed this for the pixel ramp; the
normalization on top of it was left, which is why #4030 was referenced rather than closed. This runs
the whole normalization in `float32` for a half-precision grid and narrows once.

Measured on `main` (torch 2.14.0, the version CI's `dynamo` job runs, CPU, inductor), eager vs
compiled, `normalized_coordinates=True`:

| dtype | width | max diff | elements differing |
|---|---:|---:|---:|
| bfloat16 | 100 | 0.00390625 | 110 / 400 |
| bfloat16 | 300 | 0.01171875 | 334 / 1200 |
| bfloat16 | 6000 | 0.015625 | 7384 / 24000 |
| float16 | 3000 | 0.001220703125 | 3616 / 12000 |
| float32 / float64 | any | 0.0 | 0 |

It starts at width 100, not at an exotic size. The table comes out byte-identical on torch 2.9.1 and
2.12.0.

### Root cause

Two independent causes, isolated by comparing each stage eager vs compiled at `bfloat16`, w=300:

| stage | first divergent index | max diff |
|---|---:|---:|
| `arange(w, bf16)` | -- | 0.0 (fixed by #4036) |
| `arange(w, bf16).to(f32)` | **257** | 1.0 |
| `(... / (w - 1)).to(bf16)` | 257 | 0.00390625 |
| `(xs / (w - 1) - 0.5) * 2`, all in bf16 | **12** | 0.0078125 |

1. **The widening round trip is elided.** The capture branch calls `xs.to(work_dtype)` on a half
   ramp. Eager materializes that ramp -- index 257 becomes 256 in `bfloat16` -- and then widens, so
   it divides 256. Inductor never materializes it: it folds `arange(bf16) -> f32` into one `float32`
   index computation and divides 257. Above `2 ** p` the two paths divide different numerators. The
   branch's careful widening is exactly what the compiler optimizes away.
2. **Per-op vs fused rounding.** Eager rounds `(xs / (size - 1) - 0.5) * 2` to half three times;
   inductor computes the chain in `float32` and rounds once on store. This begins at index 12 and has
   nothing to do with the representability of the width.

### The change

Build the ramp directly in `float32` when the grid dtype is half and coordinates are normalized -- so
there is no narrow-then-widen round trip left for the compiler to elide -- normalize with the same
expression on both paths, and narrow once at the end. Parity then holds by construction rather than
by matching inductor's fusion behaviour.

This is the pattern #4128 already established for `extract_patches_simple` /
`extract_patches_from_pyramid`: reduced-precision inputs get `float32` coordinate arithmetic and are
cast back, on every device, to avoid normalized-coordinate precision loss.

It is a net **-15 lines**: the `xs_norm` / `is_floating_point()` dtype gymnastics in the capture
branch existed only to chase inductor's rounding and are gone.

**Cost.** Two extra aten ops on the half + normalized path only, measured with a dispatch counter on
torch 2.14.0:

```
64x64 grid                current   this PR
  float16/bfloat16 norm      15  ->  17     (+2, _to_copy x2)
  float32/float64/int norm   15  ->  15
  every pixel path            9  ->   9
```

Wall-clock is not quotable on my hardware: identical-code control cases (float32, pixel path) swing
+-35-60% run to run on a laptop GPU, so the op count is the honest number. For scale, #4128 accepted
a measured **2.2x** CUDA `float16` slowdown in `extract_patches_simple` for the same class of fix,
keeping the universal upcast rather than making it device-conditional, on the grounds that "the
coordinate-quantization rationale applies on every device".

I put the CHANGELOG entry under **Breaking changes** rather than **Bug fixes**, following #4128,
because eager half-precision normalized coordinates do change. Move it if you read it as purely a
fix.

### Known remaining gaps

1. **`float16` on CUDA keeps a residue of exactly one ulp, on 64 of 306,352 grid elements.** This is
   *not* the gap #4036 left open -- that gap is what this PR closes. It is a different effect with a
   different cause, four orders of magnitude smaller in element count, and it exists only on CUDA at
   `float16`.

   Sweep: 12 widths (2, 3, 100, 257, 258, 300, 1000, 2049, 2050, 3000, 4128, 6000) x {`float16`,
   `bfloat16`} x {2d, 3d} -- 306,352 elements per dtype, 612,704 per device -- eager vs compiled,
   `normalized_coordinates=True`, on torch 2.14.0 (RTX 5070), against `main` at `ca1f2f7b`:

   | | differing on `main` | differing here | worst here |
   |---|---:|---:|---:|
   | CPU, both dtypes | 122,832 / 612,704 | **0** | 0.0 |
   | CUDA `bfloat16` | 63,480 / 306,352 | **0** | 0.0 |
   | CUDA `float16` | 58,438 / 306,352 | **64** | 0.00048828125 |

   The 64 sit at three widths, and every one of them is **exactly** one ulp of the coordinate value
   it lands on -- measured as `|diff| / ulp(value)`, the maximum over all 64 is 1.000, never more:

   | width | 2d | 3d | worst |
   |---:|---:|---:|---:|
   | 3000 | 0 / 12,000 | 16 / 36,000 | 0.00048828125 |
   | 4128 | 8 / 16,512 | 16 / 49,536 | 0.0000019073486328125 |
   | 6000 | 8 / 24,000 | 16 / 72,000 | 0.00006103515625 |

   The cause is that **a Python-scalar divisor and a 0-dim-tensor divisor are not the same operation
   on CUDA**, while on CPU they are bit-identical. Over 2 ** 20 random `float32` values:

   ```
   cpu    x / c == x * (1 / c): False    x / c == x / tensor(c): True
   cuda   x / c == x * (1 / c): True     x / c == x / tensor(c): False
   ```

   Eager divides by the Python int `size - 1`; the capture branch has to divide by a tensor to keep
   the size symbolic under export. On CPU those two are the same operation, so parity is exact. On
   CUDA they cannot be made to agree whichever form is chosen -- I tried the 0-dim-tensor divisor and
   the reciprocal form on both paths, and the reciprocal form fixes CUDA and breaks eight CPU tests.
   `bfloat16`'s 8-bit significand absorbs the difference entirely; `float16`'s 11 bits do not. That
   is why the test asserts bitwise everywhere except `float16` on CUDA.

2. **CPU `bfloat16` `arange` is itself wrong above index 4112.** Carried over unchanged from #4036,
   where it was gap 2; still open, and already reported upstream. 975 of the first 8,256 indices
   disagree with the correctly rounded value, the first at 4113, where `arange` returns 4096 and 4128
   is correct. (Index 4112 itself is an exact tie that rounds to even, so 4096 is right there; the
   error starts one index later. #4036 gave the first affected width as 4128 -- it is 4114.)
   Reproduced identically on torch 2.9.1 and 2.14.0, while CUDA `arange` is correctly rounded at all
   8,256. It affects the pixel ramp, which this PR does not touch.

   Upstream is [pytorch#137774](https://github.com/pytorch/pytorch/issues/137774), which reported the
   same defect at the 256 boundary and was **closed in 2024 with a documentation note rather than a
   kernel fix**. That note is in the `torch.arange` docstring today: reduced-precision sequences "may
   be affected by floating-point rounding behavior", with a recommendation to use integer dtypes. It
   covers *repeated* values; it does not cover returning a value a full ulp away from the correctly
   rounded one, which is what happens at 4113 and what #137774 itself pointed out before it was
   closed. Any fresh report has to draw that distinction or it gets closed as a duplicate.
   [pytorch#193737](https://github.com/pytorch/pytorch/pull/193737) is open and fixes this class of
   error on MPS by carrying the intermediate in double-float.

### A note on overlap

#3945 also edits these lines -- it adds `align_corners` to `create_meshgrid` and changes the
normalization formula itself. It is open with changes requested, its last commit is 10 August, and it
is already `dirty` against `main`, so it needs a rebase whatever happens here. If #3945 goes first
this fix has to apply to both `align_corners` branches.
Flagging early rather than repeating the #3995 collision.

Refs #4030. Across the sweep above, the eager/compiled gap that issue reports goes from 122,832
differing elements to **0** on CPU, and from 121,918 to **64** on CUDA -- all 64 at `float16`, each
exactly one ulp. Happy to make this `Fixes #4030` if you consider the CUDA `float16` residue -- a
consequence of how CUDA lowers a scalar divisor versus a tensor one, rather than anything kornia
controls -- out of scope for it; your call.

Two things surfaced while verifying that are **not** part of this PR:

* `test_normalized_meshgrid_trace_matches_eager_at_unrepresentable_sizes` **fails 18 ways on CUDA on
  current `main`** -- 8 at `float32` and 10 at `float64` -- for the same scalar-versus-tensor divisor
  reason above. Its docstring says of those dtypes "which pins the two paths as agreeing there rather
  than merely not being compared"; that is not true on GPU. The same 18 fail before and after this
  change. Reported separately as #4195; there is no CUDA job in CI -- the MPS gate added in #4160
  does not cover it (see #4033) -- so nothing catches it.
* On CUDA the new test is **skipped by default** by the half-precision isolation that landed in
  #4157. It parameterizes `grid_dtype` directly rather than using the global `dtype` fixture, and
  #4157's callspec scan picks that up, so it is skipped unless `--isolate-half-precision` is passed,
  which runs it in a subprocess. CI never reaches this path -- the dynamo job is CPU -- but it is
  worth knowing when reproducing on a GPU.

## What did you check?

Verified on three torch versions, baseline vs patched each time, all against the current `main`
(ca1f2f7b):

| | torch 2.9.1+cpu (half-precision job's version) | torch 2.12.0+cpu | torch 2.14.0+cu130, RTX 5070 (dynamo job's version) |
|---|---|---|---|
| new test | 4 failed -> 4 passed | 4 failed -> 4 passed | 4 failed -> 4 passed |
| `test_grid.py` under inductor | 49 -> 53 passed | 49 -> 53 passed | 49 -> 53 passed |
| `tests/geometry` | 2500 passed, unchanged | 2500 passed, unchanged | 2500 passed, unchanged |
| eager vs compiled, 12 widths x {f16,bf16}, 2d+3d (612,704 elements per device) | **0 / 612,704** | **0 / 612,704** | CPU **0 / 612,704**; CUDA `bfloat16` 0 / 306,352, `float16` 64 / 306,352, each exactly 1 ulp |

```text
$ KORNIA_TEST_OPTIMIZER=inductor pixi run test-module tests/geometry/test_grid.py
53 passed          # 49 on main, plus the 4 new; 4 of the 53 fail on main

$ pixi run test-module tests/geometry
2500 passed, 62 skipped, 12 xfailed, 6 xpassed        # identical to main

$ KORNIA_TEST_DEVICE=cuda pixi run test-module tests/geometry/test_grid.py --dtype=all
18 failed, 46 passed, 46 skipped        # the same 18 fail on main; see the note above

$ KORNIA_TEST_DEVICE=cuda pixi run test-module tests/geometry/test_grid.py --dtype=all \
      --isolate-half-precision -k normalized_meshgrid_half_dtype_matches_compile
4 passed           # the new test on GPU, including float16 within its one-ulp tolerance
```

`tests/geometry` also comes out at the same 2500 passed, baseline and patched, on torch 2.5.1 -- the
floor of the main `tests` matrix. The compile test itself never runs there: only the `dynamo` job
sets an optimizer, and it pins 2.14.0.

Locally I ran these as direct `pytest` invocations against pinned interpreters rather than through
pixi, because the four torch versions live in separate environments on this machine.

Also checked: `float32`, `float64`, `int32`, `int64` and `uint8` grids are bitwise identical to
`main` on both devices, normalized and pixel; singleton, 1x1 and zero-size axes match; the pixel ramp
is untouched; the function stays scriptable and the `torch.empty(0, device=device)` pin from #4036
still holds; `ruff check` and `ruff format` clean.

The new test is a genuine regression test: it fails 4/4 on unpatched `main` and passes 4/4 with the
change, on CPU and on CUDA. In CI it runs only in the `dynamo` job (ubuntu / py3.11 / torch 2.14.0 /
CPU), where the assertion is strictly `atol=0, rtol=0`: every other job leaves
`KORNIA_TEST_OPTIMIZER` empty, and `conftest.py` then deselects any test whose name contains
`compile` or `dynamo`. The CUDA tolerance therefore covers local GPU runs only and is never
exercised by CI.

## How did AI help?

Claude Code (Opus 5) did most of the work here: isolating the two causes stage by stage, running the
eager-vs-compiled sweeps across three torch versions and two devices, building and then rejecting two
wider variants (a single-code-path version and a reciprocal-divisor version), and drafting the change
and the test. I directed the investigation and reviewed the result. After rebasing onto current
`main` it re-ran every number above against torch 2.14.0, which CI's dynamo job moved to in the
meantime.

Everything asserted above comes from an executed run rather than from reasoning, and the numbers are
reproducible with the commands listed. Three intermediate conclusions were wrong and were caught by
measurement rather than review: that the change had no eager cost on CUDA (it adds two aten ops);
that the divisor form could be rewritten so eager and capture agree on both devices (it cannot -- the
attempt fixed CUDA and broke eight CPU tests); and that `--isolate-half-precision` would skip the new
test (it is what makes it run on CUDA -- the default is to skip it).


ducha-aiki:
## Approved at `75b351b3` — the "coincidence" claim is true, and I executed it rather than taking it on trust

Reviewed against merge-base `ca1f2f7b`, on torch 2.9.1 (the half legs' version) and 2.14.0 (the dynamo job's), CPU, macOS arm64, plus an MPS spot check.

### The two new bfloat16 manifest lines

This morning's read of your red CI was "seven recovered xfails, two new regressions". Your `75b351b3` says the two were a coincidence. Measured, `test_translation` at bfloat16, seed 0, torch 2.9.1 (2.14.0 gives the same picture):

| shape | `main` grid vs the correctly rounded bf16 grid | `main` warp error | this PR's grid vs correctly rounded | this PR's warp error | correctly rounded bf16 grid + the same bf16 warp | float64 warp, narrowed to bf16 only at the end |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| (4, 5) `shape0` | 1.95e-3 | **0** (pass) | 0 | 3.9e-3 | 3.9e-3 | 3.9e-3 |
| (2, 6) `shape1` | 2.93e-3 | 2.9e-3 | 0 | 2.9e-3 | 2.9e-3 | 1.95e-3 |
| (4, 3) `shape2` | 1.95e-3 | **0** (pass) | 0 | 3.9e-3 | 3.9e-3 | 3.9e-3 |
| (5, 7) `shape3` | 3.9e-3 | 1.17e-2 | 0 | 7.8e-3 | 7.8e-3 | 6.1e-3 |

The tolerance is `atol=1e-4`. The last two columns are the point: **no bfloat16 warped grid passes any of the four shapes**, not the correctly rounded one and not the one you get by doing the whole warp in float64 and rounding once at the very end. `main`'s grid is one ulp *off* the correctly rounded value at `shape0`/`shape2`, and that error happens to make the transformed sample positions land on exact pixel indices, which is why those two cells passed. This PR's grid is bit-identical to the correctly rounded bf16 grid at all four shapes. So the two new manifest lines are right, and the CHANGELOG sentence describing them is accurate.

One heads-up, not a blocker, because it is a property of the test rather than of this PR: `test_translation` draws `torch.rand`, and at bfloat16 whether the one-ulp interpolation residue crosses a rounding boundary depends on the draw. Over 300 seeds at head, torch 2.9.1: `shape0` passes 12 times, `shape2` 52, `shape1` 8 (that one was already in the manifest on `main`), `shape3` 0. The CI profile seeds each node deterministically, so the manifest is stable there; I ran `test_homography_warper.py` + `test_grid.py` at head under `--xfail-known-failures --known-failure-profile=cpu-bfloat16` (130 passed / 8 xfailed) and `cpu-float16` (133 / 5), no XPASS. The unseeded `pixi run test-half` will XPASS `shape2` about one draw in six, which is the kind of thing that then gets blamed on a manifest edit. Seeding that test, or giving it a fixed patch, is a separate one-line change; I am not asking for it here.

### Everything else verified

- **Fail-on-base is real:** `test_normalized_meshgrid_half_dtype_matches_compile` copied onto `main`, `KORNIA_TEST_OPTIMIZER=inductor`, torch 2.14.0: 4 failed; at head 4 passed, and the whole `test_grid.py` under inductor is 53 passed.
- **Non-half dtypes are untouched:** a (dtype, shape, sum, abs-sum) fingerprint of `create_meshgrid` and `create_meshgrid3d` over float32/float64/int32/int64/uint8, normalized and pixel, at (1,1), (1,7), (3,1), (0,4), (4,5), (257,300), (2,2049) is identical to `main`. Only the normalized float16/bfloat16 cells move, as the CHANGELOG says.
- **Scriptable and traceable:** `torch.jit.script` and `torch.jit.trace` outputs equal eager at bfloat16, float16 and float32 on both torch versions.
- **MPS:** head at (257, 300) normalized, float16/bfloat16/float32, `PYTORCH_ENABLE_MPS_FALLBACK` unset: bit-identical to CPU.
- `test_grid.py` on 2.9.1, float32+float64: 64 passed. Both half legs are green on CI with `--verify-known-failures`, which is the authoritative check on the manifests.

### Two calls that are the maintainer's, with my recommendation

1. **`Refs #4030` -> `Fixes #4030`.** #4030 is the eager/compiled divergence at half precision as observed on CPU; #4036 closed the pixel ramp and this closes the normalization, 0 differing elements on CPU. The CUDA float16 one-ulp residue is the scalar-vs-tensor divisor lowering difference that #4195 already tracks separately. I would close #4030 with this PR.
2. **CHANGELOG under `Breaking changes`:** agree, for the reason you give; eager bfloat16/float16 normalized grids change value, and the entry says what used to happen.

Needs the `CHANGELOG.md` rebase (it conflicts with today's merges; the source and manifest files merge cleanly), and then this is mergeable from my side.

ducha-aiki (approved):

Issue #4214, opened 2026-09-04
CI: no job sets timeout-minutes, so a hung test burns the full 6-hour default and reports as an unexplained red


ducha-aiki:
### Describe the bug

No job in `.github/workflows/` sets `timeout-minutes`:

```console
$ grep -rn "timeout-minutes" .github/workflows/
$
```

so every job inherits GitHub's 6-hour default. Today that turned a single hung test into a six-hour burn on a PR that had nothing to do with it, and reported it as an unexplained red rather than as a timeout.

**The observed instance** — `half-precision (float16) / tests (3.11, 2.9.1)` on #4162, [run 33828540585](https://github.com/kornia/kornia/actions/runs/33828540585/job/100886437358):

```text
status:   cancelled          (not "failure")
started:  2026-09-04T02:12:13Z
completed:2026-09-04T08:12:43Z          6 h 0 m 30 s
step 11 "Run tests" -> cancelled;  steps 12 "Publish known-failure receipt" and 13 "Upload candidate manifest" -> skipped
```

The whole `Run tests` step produced one 21,428-second gap with no output:

```text
02:15:33  tests/feature/test_responces_local_features.py ........x.......x.   [ 58%]
08:12:40  ##[error]The operation was canceled.
```

i.e. it wedged inside the first test of `tests/feature/test_scale_space_detector.py` and never emitted another character.

### Why this is worth a guard rail rather than a shrug

- **The norm for that job is 6–8 minutes.** Every other run of the float16 leg on 2026-09-04 finished in that window (`33828089149`, `33831292379`, `33831333600`, `33837154369`, `33838831649`, `33839561035`, `33840321694`, `33854832657`). This is a hang, not slowness, and not the AMD-runner cost from #4154 — the runner here was the *faster* `INTEL(R) XEON(R) PLATINUM 8573C`.
- **It is invisible as a timeout.** The job conclusion is `cancelled`, the receipt and candidate-manifest steps are `skipped`, and `gh pr checks` renders it as a plain `fail` next to a duration. Diagnosing it took reading the raw job log for timestamp gaps. A `timeout-minutes` breach at least names itself.
- **It blocks the PR and costs six runner-hours.** `collector` fails on `failure` **or** `cancelled`, so the PR is red until someone notices and re-runs.
[...]

PR #4232, opened 2026-09-05
fix(contrib): keep distance_transform gradients finite

### LINKED ISSUE #4183: distance_transform returns NaN gradients wherever its convolution hits zero (log(0) through nan_to_num)
## Summary

`kornia.contrib.distance_transform` returns **NaN** gradients for any input that has a region far enough from a boundary pixel for the cascade's convolution to return exactly zero — which includes the ordinary case of a sparse binary mask, and every all-zero input. The forward result is correct; only the backward is poisoned, silently, with no warning.

## Reproduction

`main` @ `ee6b713f`, torch 2.9.1, CPU:

```python
import torch
from kornia.contrib import distance_transform

torch.manual_seed(0)
x = (torch.rand(1, 1, 16, 16) > 0.9).float().requires_grad_()
distance_transform(x, kernel_size=3).sum().backward()
print(torch.isnan(x.grad).sum().item(), "of", x.grad.numel())   # 193 of 256
```

| input (float64) | NaN entries in the gradient |
|---|---|
| random binary mask, 16×16, ~10% ones | 175 / 256 |
| one-hot 4×4 (single boundary pixel) | 15 / 16 |
| all-zeros 4×4 | 16 / 16 |

float32 on the same random mask: 193 / 256. All NaN, no Inf.

## Mechanism

`kornia/contrib/distance_transform.py`, both the 2D and the 3D impl:

```python
cdt = filter2d(boundary, kernel, border_type="replicate")
cdt = -h * torch.log(cdt)
cdt = torch.nan_to_num(cdt, nan=0.0, posinf=0.0, neginf=0.0)
```

Where the convolution returns exactly `0`, `log` gives `-inf`, and `nan_to_num` cleans the *forward* value to `0`. Its backward, however, multiplies the incoming gradient by `isfinite(input)`, so the `inf` derivative of `log` at zero meets a zero multiplier and yields `0 * inf = NaN`. Isolated:

```python
x = torch.zeros(1, 1, 4, 4, dtype=torch.float64, requires_grad=True)
cdt = filter2d(x, torch.ones(1, 3, 3, dtype=torch.float64), border_type="replicate")
torch.nan_to_num(-0.35 * torch.log(cdt), nan=0.0, posinf=0.0, neginf=0.0).sum().backward()
print(x.grad.flatten()[:3])   # tensor([nan, nan, nan], dtype=torch.float64)
```

## Why the existing tests do not catch it

`tests/contrib/test_conv_distance_transformer.py::test_gradcheck` uses `torch.ones(B, C, H, W)` for both the 2D and the 3D case. With an all-ones image the convolution is positive everywhere, the `log` never sees zero, and the defective branch is never entered. The regression test for this needs an input with an actual empty region — a sparse mask or a one-hot image.

## Candidate fix

Guarding the `log` instead of cleaning up after it removes every NaN and leaves the forward **byte-identical**. Measured on the three inputs in the table above, against the library's own output (`torch.equal`, float64):

```python
positive = cdt > 0
cdt = -h * torch.log(torch.where(positive, cdt, torch.ones_like(cdt)))
cdt = torch.where(positive, cdt, torch.zeros_like(cdt))
```

| input | forward byte-identical to current | NaN grads now | NaN grads with the guard |
|---|---|---|---|
| random mask | yes | 193 | **0** |
| one-hot | yes | 15 | **0** |
| all zeros | yes | 16 | **0** |

Note `mask = cdt > 0` is computed *after* this line, on the cleaned value, so a naive `cdt.clamp_min(eps)` before the `log` is **not** a valid alternative: it turns those positions into `-h*log(eps) > 0` and flips them into the mask, changing the forward. Whatever the fix, the forward has to be shown byte-identical.

## Classification

**Broken output** (gradients), reachable from ordinary usage of a differentiable op. Not related to #4152 — verified present and identical on both sides of #4158, which changes nothing here.

## Environment

kornia @ `ee6b713f`, torch 2.9.1, Python 3.11.14, macOS/arm64, CPU.

Jeevang1-epic:

Fixes NaN gradients in `kornia.contrib.distance_transform` when an intermediate convolution evaluates to exactly zero.

Previously, both the 2D and 3D implementations evaluated:

```python
cdt = -h * torch.log(cdt)
cdt = torch.nan_to_num(cdt, nan=0.0, posinf=0.0, neginf=0.0)
```

When `cdt == 0`, `torch.log(0)` produces `-inf`. `torch.nan_to_num` repairs the forward value, but the backward still passes through the singular derivative of `log` at zero, which can produce NaN gradients.

The fix avoids evaluating `log` at inactive positions:

```python
positive = cdt > 0
cdt = -h * torch.log(torch.where(positive, cdt, torch.ones_like(cdt)))
cdt = torch.where(positive, cdt, torch.zeros_like(cdt))
```

This is applied to both `_distance_transform_2d_impl` and `_distance_transform_3d_impl`.

The zero-convolution positions therefore follow a finite inactive branch instead of passing through `log(0)`, while positive convolution values retain the existing computation.

The existing `torch.nan_to_num(...)` step is intentionally retained so handling of exceptional non-finite inputs and the surrounding behavior remain unchanged.

The patch does not clamp the convolution result before `log`; doing so would turn zero positions into positive distance values and change the existing mask/forward semantics.

### Regression coverage

A dedicated regression test was added for both:

- 2D input: `(1, 1, 4, 4)`
- 3D input: `(1, 1, 4, 4, 4)`

and for:

- all-zero inputs
- one-hot inputs

across the repository's `float32` / `float64` CPU test matrix.

The regression verifies that:

- the forward output is finite;
- the all-zero case remains exactly zero;
- backward completes successfully;
- `sample.grad` is produced;
- the gradient shape exactly matches the input shape;
- all gradient entries are finite.

The regression was also checked against the unpatched implementation, where the affected cases fail because the gradients contain NaNs.

I additionally compared patched and pre-fix forward results on representative sparse, one-hot, and all-zero inputs using exact `torch.equal` comparisons. The forward results remain unchanged while the affected gradients become finite.

### Compatibility with #4158

This was checked after rebasing onto the current `main`, which already contains #4158.

The two fixes address separate failure modes in the same function:

- #4158 handles kernel working precision / `exp(-dist / h)` underflow;
- this change prevents the later logarithm from being evaluated where the convolution result is zero.

The #4158 behavior, kernel construction, working-dtype handling, `_check_h_range`, offsets, ONNX-safe operations, and public API are unchanged by this patch.

Fixes #4183.

### Regression test

```text
pixi run test-module tests/contrib/test_conv_distance_transformer.py::TestConvDistanceTransform::test_zero_convolution_has_finite_gradients --device=cpu --dtype=float32,float64

8 passed
```

### Full distance-transform test module

```text
pixi run test-module tests/contrib/test_conv_distance_transformer.py --device=cpu --dtype=float32,float64

42 passed
```

### Lint / formatting

```text
pixi run lint

ruff check    Passed
ruff format   Passed
```

### Type checking

```text
pixi run typecheck
```

Result:

```text
exit code 0
```

The command reports one existing deprecation diagnostic in `kornia/feature/lightglue.py` for `torch.cuda.amp.custom_fwd`; it is unrelated to this change.

### Pre-commit

Targeted pre-commit was run for:

```text
kornia/contrib/distance_transform.py
tests/contrib/test_conv_distance_transformer.py
```

The applicable formatting, lint, merge-conflict, large-file, codespell, and license-header checks pass. On Windows, the mixed-line-ending hook normalized the two touched files when required.

### Git / scope checks

The branch was rebased onto current `upstream/main`.

The final source change is limited to:

```text
kornia/contrib/distance_transform.py
tests/contrib/test_conv_distance_transformer.py
```

with:

```text
2 files changed, 24 insertions(+), 2 deletions(-)
```

`git diff --check` exits successfully.

## How did AI help?

I used AI-assisted tooling to help inspect the existing implementation and test coverage, reproduce the reported backward failure, compare the patch against the issue's requested scope, and organize the validation steps.

I reviewed the resulting source and test changes myself, verified the regression against the pre-fix behavior, checked exact forward-output preservation, confirmed finite gradients after the fix, rebased onto the merged #4158 implementation, and reran the relevant tests, lint, type checking, pre-commit, and Git scope checks before preparing this pull request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -351,6 +351,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+
 * `kornia.io.load_image` and `write_image` work on the kornia_rs that a plain `pip install kornia`
   resolves. kornia_rs 0.1.11 moved its image readers and writers from the package root into
   `kornia_rs.io`, and kornia kept calling the root, so on 0.1.11 and newer `load_image` raised
@@ -362,6 +363,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   so instead of mislabelling the tensor. The nightly PyPI job now imports the installed wheel from
   outside the checkout (it used to import the source tree) and round-trips an image instead of only
   importing. (#4326)
+
+* `distance_transform` no longer returns NaN gradients when the cascade's convolution is exactly zero, including
+  sparse masks and all-zero inputs; the existing forward output is unchanged. (#4232)
+
+
 * `SemanticSegmentation.visualize` works for CUDA, MPS and half-precision models. It indexed the
   CPU-drawn colormap with the mask's `argmax`, which raises for CUDA and MPS masks, and recognised a
   softmax head with `torch.allclose(sum, 1)` at float32-sized default tolerances, which a float16 or
--- a/kornia/contrib/distance_transform.py
+++ b/kornia/contrib/distance_transform.py
@@ -76,7 +76,9 @@ def _distance_transform_2d_impl(image: torch.Tensor, kernel_size: int, h: float)
 
     for i in range(n_iters):
         cdt = filter2d(boundary, kernel, border_type="replicate")
-        cdt = -h * torch.log(cdt)
+        positive = cdt > 0
+        cdt = -h * torch.log(torch.where(positive, cdt, torch.ones_like(cdt)))
+        cdt = torch.where(positive, cdt, torch.zeros_like(cdt))
         cdt = torch.nan_to_num(cdt, nan=0.0, posinf=0.0, neginf=0.0)
 
         mask = cdt > 0
@@ -108,7 +110,9 @@ def _distance_transform_3d_impl(image: torch.Tensor, kernel_size: int, h: float)
 
     for i in range(n_iters):
         cdt = filter3d(boundary, kernel, border_type="replicate")
-        cdt = -h * torch.log(cdt)
+        positive = cdt > 0
+        cdt = -h * torch.log(torch.where(positive, cdt, torch.ones_like(cdt)))
+        cdt = torch.where(positive, cdt, torch.zeros_like(cdt))
         cdt = torch.nan_to_num(cdt, nan=0.0, posinf=0.0, neginf=0.0)
 
         mask = cdt > 0
--- a/tests/contrib/test_conv_distance_transformer.py
+++ b/tests/contrib/test_conv_distance_transformer.py
@@ -173,6 +173,24 @@ class TestConvDistanceTransform(BaseTester):
         sample3d = torch.ones(B, C, D, H, W, device=device, dtype=torch.float64, requires_grad=True)
         self.gradcheck(kornia.contrib.distance_transform, (sample3d,))
 
+    @pytest.mark.parametrize("shape", [(1, 1, 4, 4), (1, 1, 4, 4, 4)], ids=["2d", "3d"])
+    @pytest.mark.parametrize("has_signal", [False, True], ids=["all-zero", "one-hot"])
+    def test_zero_convolution_has_finite_gradients(self, shape, has_signal, device, dtype):
+        sample = torch.zeros(*shape, device=device, dtype=dtype)
+        if has_signal:
+            sample[(0, 0, *((1,) * (len(shape) - 2)))] = 1.0
+        sample.requires_grad_(True)
+
+        output = kornia.contrib.distance_transform(sample)
+        assert torch.isfinite(output).all()
+        if not has_signal:
+            assert torch.equal(output, torch.zeros_like(output))
+
+        output.sum().backward()
+        assert sample.grad is not None
+        assert sample.grad.shape == sample.shape
+        assert torch.isfinite(sample.grad).all()
+
     def test_loss_grad(self, device, dtype):
         B, C, H, W = 1, 1, 32, 32
         sample1 = torch.rand(B, C, H, W, device=device, dtype=dtype, requires_grad=True)


ducha-aiki:
## Approved at `bb978c08` once a CHANGELOG entry is added — verified on torch 2.9.1 and 2.14.0

Reviewed against merge-base `ca1f2f7b`, CPU.

**Verified**

- **Fail-on-base:** the new `test_zero_convolution_has_finite_gradients` copied onto `main` -> 8 failed on both torch versions (2d/3d × all-zero/one-hot × float32/float64), matching #4183's three reproductions. Head: 42 passed on both versions, and 22 passed under each CI half profile (`cpu-float16`, `cpu-bfloat16`) with no XPASS.
- **Forward is bit-identical:** `torch.equal` to `main` over 36 cases (2D and 3D; random, all-zero, one-hot and all-one inputs; `kernel_size` 3/5/7; `h` 0.35/0.1). Where the convolution is exactly zero the old path produced `-h * log(0) = inf` which `nan_to_num` mapped to `0`; the new path writes `0` directly, and elsewhere the `where` is the identity.
- Gradient: `log` now only ever sees the substituted `1` at inactive positions, so its derivative there is finite and the outer `where` zeroes it; active positions are untouched. Keeping `nan_to_num` afterwards is harmless.

**The one ask:** there is no `CHANGELOG.md` entry. Add a bullet under `## Unreleased` / `### Bug fixes` citing `(#4232)`: `distance_transform` returned NaN gradients wherever the cascade's convolution was exactly zero (sparse masks, all-zero inputs), forward unchanged. `Fixes #4183` is already in the body and registered.

**CI:** the `half-precision (float16)` job on your run hung twice (attempt 1 03:32 -> 09:32 UTC, attempt 2 12:32 -> 18:32, both cancelled at the 6-hour cap, both stalled on entering `tests/feature/test_scale_space_detector.py`, a file this PR does not touch). That is #4214, which hits some Intel runners; the bfloat16 leg and everything else is green. The red `collector` is not yours; it gets re-run after the CHANGELOG push.

Jeevang1-epic:
@ducha-aiki  Thanks for the review! I’ve added the requested CHANGELOG entry and updated the branch. All 40 checks are green now. Ready for your review whenever you have time.

ducha-aiki:
Delta re-review at `0eca0d62` — **APPROVED, merge-ready**. The only outstanding ask was the CHANGELOG entry; `0eca0d62` adds it under `### Bug fixes` in the Unreleased section (`+6`, nothing else touched), the code and test files are byte-identical to the approved `bb978c08`, CI is 40/40 green including the float16 leg that hung twice before (#4214), and GitHub reports `MERGEABLE`. Nothing left for you to do.

ducha-aiki (approved):

Issue #4177, opened 2026-09-02
W1b: `Boxes.from_tensor(..., validate_boxes=True)` never validates the vertex modes; `'vertices'` deforms a non-rectangular or reordered quadrilateral instead of rejecting it


ducha-aiki:
## Summary

`Boxes.from_tensor(boxes, mode, validate_boxes=True)` documents `validate_boxes` as *"check if boxes are valid rectangles or not. Valid rectangles are those with width and height >= 1 (>= 2 when mode ends with '_plus' suffix)"* (`kornia/geometry/boxes.py:524` at `e0969438`). The check exists only for the `'xyxy'`, `'xyxy_plus'` and `'xywh'` modes (`_boxes_to_quadrilaterals`, `boxes.py:139-143`). The two vertex modes are never validated, and `'vertices'` goes further: its exclusive-to-inclusive import subtracts one from fixed vertex slots — `x` of vertices 1 and 2, `y` of vertices 2 and 3 — so a non-rectangular or non-canonically-ordered quadrilateral is silently deformed instead of rejected.

## Reproduction

```python
import torch
from kornia.geometry.boxes import Boxes

q = torch.tensor([[[0., 0.], [9., 0.], [3., 7.], [0., 1.]]])            # not a rectangle
Boxes.from_tensor(q, mode="vertices", validate_boxes=True).data          # accepted
Boxes.from_tensor(q, mode="vertices_plus", validate_boxes=True).data     # accepted, stored verbatim

inv = torch.tensor([[[5., 5.], [1., 5.], [1., 1.], [5., 1.]]])          # axis-aligned, vertices in reversed order
Boxes.from_tensor(inv, mode="vertices", validate_boxes=True).to_tensor("xywh")

Boxes.from_tensor(torch.tensor([[5., 5., 1., 1.]]), mode="xyxy", validate_boxes=True)    # the xy* modes do reject
```

## Observed

```
vertices      -> data = [[[0., 0.], [8., 0.], [2., 6.], [0., 0.]]]     # fourth vertex collapsed onto the first, no error
vertices_plus -> data = [[[0., 0.], [9., 0.], [3., 7.], [0., 1.]]]     # no error
inverted 'vertices' -> xywh [[0., 0., 6., 6.]]                          # the box spans x=1..5; width 6 comes from subtracting at the wrong slots
xyxy          -> ValueError: Some boxes have negative widths or 0.
```

[...]

PR #4243, opened 2026-09-05
fix(geometry): reject non-finite bbox coordinates

### LINKED ISSUE #4238: NaN coordinates pass validate_bbox and Boxes.from_tensor validation
Two bbox validation paths accept NaN coordinates: `validate_bbox` can return `True`, and `Boxes.from_tensor(..., validate_boxes=True)` can construct a box with NaN extents. This lets invalid coordinates pass the validation boundary and propagate into downstream geometry.

## Reproduction

```python
import torch
from kornia.geometry.bbox import validate_bbox

vertices = torch.tensor([
    [[0.0, 0.0], [float("nan"), 0.0], [1.0, 1.0], [0.0, 1.0]]
])
print(validate_bbox(vertices))
# True

xyxy = torch.tensor([[0.0, 0.0, float("nan"), 1.0]])
boxes = Boxes.from_tensor(xyxy, mode="xyxy", validate_boxes=True)
print(boxes.to_tensor("xyxy"))
# tensor([[nan, 0., nan, 1.]])
```

## Expected behavior

NaN coordinates should not be certified as valid. In the eager validation paths shown above, `validate_bbox` should return `False`, and `Boxes.from_tensor(..., validate_boxes=True)` should reject the input with a meaningful error. Preserve the existing opt-out/export policy rather than adding data-dependent validation to paths that intentionally bypass it.

## Cause

- [bbox.py](https://github.com/kornia/kornia/blob/3f102790/kornia/geometry/bbox.py#L80-L94) computes opposite-edge differences and rejects only when a difference is greater than `1e-4`. A NaN difference does not satisfy that comparison, so validation falls through to `True`.
- [boxes.py](https://github.com/kornia/kornia/blob/3f102790/kornia/geometry/boxes.py#L140-L145) rejects widths/heights only when they are `<= 0`. A NaN extent does not satisfy that comparison either, so construction succeeds despite `validate_boxes=True`.

The bbox/Boxes batch-4a convention audit and its current pins did not cover this case. This is distinct from the already tracked inclusive-width, vertex-order, rank, and list-padding issues.

## Suggested regression scope

Cover a NaN in each coordinate position, invalid rows mixed with valid rows, and the supported eager input layouts/modes. Check coordinate finiteness rather than relying only on derived widths: a non-finite origin must not escape because its extent was supplied separately. Include infinities in the finite-input policy tests as well; the executed reproductions above establish the NaN cases specifically. Keep valid finite inputs and the explicit validation opt-out behavior unchanged.

## Verification

Both reproductions executed on Kornia `0.9.0rc1`, commit `3f102790`, PyTorch `2.14.0`, Python `3.11.14`, macOS CPU, float32. CUDA, MPS, and export behavior were not tested for this finding. The existing convention/wart selection across conversions, bbox/boxes, and geometry transforms remains green: 343 passed, 17 xfailed, 2 skipped on CPU float32/float64.

09Catho:

`validate_bbox` now returns `False` for NaN and infinite coordinates. Eager `Boxes.from_tensor` validation now raises a clear `ValueError` for the same inputs in the `xyxy`, `xyxy_plus`, and `xywh` modes. The existing `validate_boxes=False` opt-out remains unchanged.

Fixes #4238.

## What did you check?

The new regressions cover every coordinate position, NaN/positive infinity/negative infinity, mixed valid and invalid rows, and both unbatched and batched inputs. The focused geometry tests also retain the existing JIT coverage.

```text
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/geometry/test_bbox.py tests/geometry/test_boxes.py --device=cpu --dtype=float32,float64
190 passed in 2.12s

ruff check kornia/geometry/bbox.py kornia/geometry/boxes.py tests/geometry/test_bbox.py tests/geometry/test_boxes.py
All checks passed!

ruff format --check kornia/geometry/bbox.py kornia/geometry/boxes.py tests/geometry/test_bbox.py tests/geometry/test_boxes.py
4 files already formatted

python -m py_compile kornia/geometry/bbox.py kornia/geometry/boxes.py tests/geometry/test_bbox.py tests/geometry/test_boxes.py
git diff --check
```

Pixi-based full pre-commit and typecheck were not run because Pixi is not installed in the local environment.

## How did AI help?

AI helped inspect the issue and repository guidance, then draft the implementation and regression tests. I reproduced both failures on the merge base, reviewed the resulting diff, and ran the checks listed above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -411,6 +411,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   branch, used to raise: `_compute_zero_padding3d` defined a `(k - 1) // 2` helper and then returned the
   full kernel sizes, so the padded volume did not match the kernel it was convolved with. (#4241, #4242)
 
+* Eager bbox validation now rejects NaN and infinite coordinates: `validate_bbox` returns `False` and
+  `Boxes.from_tensor(..., validate_boxes=True)` raises `ValueError` for the `xyxy`, `xyxy_plus` and `xywh`
+  modes, instead of accepting them as valid geometry (closes #4238). (#4243)
+
 * `HyNet` and `SOSNet` now run in half precision, on CPU and on GPU, and no longer return NaN for a
   degenerate patch (closes #4224). Two defects sat on the same line. On CPU both raised
   `NotImplementedError: "avg_pool3d_out_frame" not implemented for 'Half'` (and the `'BFloat16'` spelling):
--- a/kornia/geometry/bbox.py
+++ b/kornia/geometry/bbox.py
@@ -70,6 +70,9 @@ def validate_bbox(boxes: torch.Tensor) -> bool:
     if not (len(boxes.shape) in [3, 4] and boxes.shape[-2:] == torch.Size([4, 2])):
         return False
 
+    if not torch.isfinite(boxes).all():
+        return False
+
     if len(boxes.shape) == 4:
         boxes = boxes.reshape(-1, 4, 2)
 
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -139,6 +139,8 @@ def _boxes_to_quadrilaterals(boxes: torch.Tensor, mode: str = "xyxy", validate_b
 
         # Value validation reads the data, which graph capture cannot do; skip it under export.
         if validate_boxes and not is_exporting():
+            if not torch.isfinite(boxes).all():
+                raise ValueError("Some boxes have non-finite coordinates.")
             if (width <= 0).any():
                 raise ValueError("Some boxes have negative widths or 0.")
             if (height <= 0).any():
--- a/tests/geometry/test_bbox.py
+++ b/tests/geometry/test_bbox.py
@@ -123,6 +123,23 @@ class TestBbox2D(BaseTester):
         assert expanded.stride(0) == 0
         assert validate_bbox(expanded) is True
 
+    def test_convention_validate_bbox_rejects_non_finite_coordinates_4238(self, device, dtype):
+        # Pin kornia#4238: NaN and infinite coordinates must not pass validation,
+        # including when an invalid row is mixed with a valid row.
+        boxes = torch.tensor(
+            [
+                [[0.0, 0.0], [4.0, 0.0], [4.0, 4.0], [0.0, 4.0]],
+                [[1.0, 1.0], [5.0, 1.0], [5.0, 5.0], [1.0, 5.0]],
+            ],
+            device=device,
+            dtype=dtype,
+        )
+        for coordinate_index in range(8):
+            for non_finite in [float("nan"), float("inf"), float("-inf")]:
+                invalid_boxes = boxes.clone()
+                invalid_boxes[1].reshape(-1)[coordinate_index] = non_finite
+                assert validate_bbox(invalid_boxes) is False
+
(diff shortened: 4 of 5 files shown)


ducha-aiki:
## Approved at `8d93dced` — both #4238 reproductions now reject; one CHANGELOG sentence to fold into the rebase

Reviewed against merge-base `acb861c3`, torch 2.14.0, CPU, macOS arm64. All 38 checks green at head. Copilot's one inline comment is right; I verified it rather than its verdict.

**Verified**

- **Fail-on-base:** the four new `_4238` pins copied onto `main` -> 4 failed (the opt-out pin passes on both sides, as a policy pin should).
- Head: `test_bbox.py` + `test_boxes.py`, cpu, float32+float64 -> 190 passed. `uvx ruff@0.16.4` check/format clean.
- `validate_bbox` returns `False` for a NaN under `torch.jit.script` as well as eager, and for `inf` at float16; the check sits before the rank-4 reshape, so the `(B, N, 4, 2)` layout is covered.
- The `is_exporting()` gate is preserved: the new value check is inside the existing `validate_boxes and not is_exporting()` block, so export graphs gain no data-dependent branch, which is what #4238 asked for.
- The bbox/boxes files merge cleanly over #4245 and #4253; only `CHANGELOG.md` conflicts.

**Scope I checked and am not asking you to widen**, recorded so nobody reopens #4238 for it:

- `Boxes.from_tensor(..., mode="vertices" / "vertices_plus", validate_boxes=True)` still accepts NaN vertices. That is consistent with the documented contract ("this flag has no validation effect for vertex modes", tracked in #4177), so it belongs there, not here.
- `Boxes3D.from_tensor(..., validate_boxes=True)` accepts NaN, and `validate_bbox3d` raises an `AssertionError` ("must be cube") on a NaN instead of returning `False`. #4238 is 2D-only; I am filing that as its own issue.

**The one ask, cosmetic:** Copilot's point on the CHANGELOG entry stands. It names only `validate_bbox`, but the user-visible change is equally `Boxes.from_tensor(..., validate_boxes=True)` now raising `ValueError` for the `xy*` modes. One sentence, with `closes #4238` in it, while you rebase `CHANGELOG.md` anyway.

09Catho:
Rebased onto main (head now `a5778242`) to resolve the CHANGELOG.md conflict, and folded the CHANGELOG entry into one sentence that names both `validate_bbox` and `Boxes.from_tensor(..., validate_boxes=True)`, with "closes #4238" included. Re-ran the focused suite after the rebase: `tests/geometry/test_bbox.py` + `tests/geometry/test_boxes.py`, cpu, float32+float64 -> 307 passed, 30 xfailed. `ruff check` and `ruff format --check` both clean.

ducha-aiki (approved):
Delta re-review: approved at `a5778242`. Compared the old and new PR patches against their respective merge bases: production changes are identical, the regression tests are preserved, and the only incidental test-patch difference is an already-resolved blank-line cleanup. The CHANGELOG now correctly limits the claim to eager validation and the xyxy/xyxy_plus/xywh modes.

Re-ran `tests/geometry/test_bbox.py tests/geometry/test_boxes.py --device=cpu --dtype=float32,float64` in the rebased checkout (torch 2.14.0): 307 passed, 30 expected failures. The prior substantive approval stands. GitHub reports mergeable; wait for CI on this new head before merging.

ducha-aiki (approved):

Issue #4244, opened 2026-09-05
Boxes.pad, unpad, and clamp fail on valid unbatched boxes


ducha-aiki:
Reproduced 2026-09-05 at
`acb861c31e51cc4cfd375fe484978827801ef77e`, PyTorch 2.14.0, Python 3.11, CPU.
Searches for `Boxes unbatched` and `Boxes pad clamp` found no existing issue describing
this failure. #4017 covers different cases: tuple clamp bounds, trim, and fast translation.

## Reproduction

Run from the checkout root:

```python
import torch
from kornia.geometry.boxes import Boxes

for name, args in [
    ("pad", (torch.tensor([[1., 2., 3., 4.]]),)),
    ("unpad", (torch.tensor([[1., 2., 3., 4.]]),)),
    ("clamp", (torch.tensor([[0., 0.]]), torch.tensor([[5., 5.]]))),
]:
    for batched in [False, True]:
        source = torch.tensor([[1., 2., 5., 4.]])
        boxes = Boxes.from_tensor(source[None] if batched else source, mode="xyxy")
        try:
            print(name, batched, getattr(boxes, name)(*args).data)
        except Exception as error:
            print(name, batched, type(error).__name__, str(error))
```

## Observed

For the valid unbatched container (`data.shape == (1, 4, 2)`):

- `pad` and `unpad`: `RuntimeError: output with shape [1, 4] doesn't match the broadcast shape [1, 1, 4]`.
- `clamp`: `IndexError: too many indices for tensor of dimension 2`.

Adding a leading singleton batch dimension to the same source makes all three calls work,
without changing the padding or bound tensors. The padding produces an offset of `[1, 3]`,
unpadding `[-1, -3]`, and these clamp bounds leave the box unchanged.

## Expected

These operations should support the class's documented unbatched representation and preserve
its rank, matching the corresponding singleton-batch operation. In-place calls should preserve
the established object and storage behavior. Include asymmetric coordinates, multiple boxes,
both mutation modes for clamp, and dtype/device fixtures in regression tests.

## Classification

Broken-output crash on valid class input. Idiom 3: the accepted class shape contract is wider
than the method's implicit batched indexing/broadcast assumptions. A focused fix is welcome;
stage 4b documents the currently supported batched path without changing runtime behavior.

## Disposition

n/a — single symbol

This is a valid-input crash report, not a proposal to alias or consolidate the methods.
The three entry points are grouped because they share the same unsupported container shape
[...]

Issue #4020, opened 2026-08-26
Per-op convention blocks: geometry.bbox / geometry.boxes (batch 4)


ducha-aiki:
Tracking issue for conventions batch 4: per-op `Convention:` blocks and pinning tests for the bounding-box surface — `kornia/geometry/bbox.py` (10 functions) and `kornia/geometry/boxes.py` (`Boxes`, `Boxes3D`, `VideoBoxes`). Same method as batch 3 (#3941): document actual behavior including warts, pin every precise claim with a `test_convention_*` / `test_wart_*` test, change no runtime behavior, and route every behavior finding to its own issue.

## Sub-batches (one PR each)

| PR | scope | symbols |
|---|---|---|
| 4a | representation & conversion core — the five modes, inclusive `+1` arithmetic, vertex order, dtype policy | `Boxes` (canonical block), `Boxes.from_tensor`, `Boxes.to_tensor`, `get_boxes_shape`, `merge`, `index_put`, `validate_bbox`, `infer_bbox_shape` |
| 4b | geometry ops | `transform_bbox`, `nms`, `Boxes.transform_boxes` / `transform_boxes_`, `translate`, `pad`, `unpad`, `clamp`, `trim`, `compute_area`, `filter_boxes_by_area` |
| 4c | masks, generators, 3D, video | `bbox_to_mask`, `bbox_to_mask3d`, `bbox_generator`, `bbox_generator3d`, `validate_bbox3d`, `infer_bbox_shape3d`, `Boxes.to_mask`, `Boxes3D` (canonical block) and its methods, `VideoBoxes` (three undocumented methods) |

Class rule: the class docstring carries the canonical block; every method that depends on it gets a one-line pointer; a method gets its own block only where it adds or deviates; pure accessors get nothing.

## Behavior findings from the up-front audit (filed 2026-08-26, none repaired here)

Convention conflicts / pairs (each carries a `## Disposition`):
- #4008 — `nms` is the only exclusive-arithmetic op in `kornia.geometry.bbox` (KEEP SEPARATE, proposed)
- #4009 — `Boxes.to_tensor('vertices')` fed to the free functions reads one pixel larger per axis (window, with #3934)
- #4012 — `Boxes(...)` raises on integer input while `from_tensor` casts (window)
- #4013 — `validate_bbox` returns `False`, `validate_bbox3d` raises (window)
[...]

PR #4245, opened 2026-09-05
docs(geometry): Convention blocks for bbox/boxes (batch 4b: geometry operations)

ducha-aiki:
## Summary

Documents and pins the eleven bbox/Boxes geometry operations in conventions batch 4b. Callers can see endpoint-restoration behavior, exclusive NMS IoU, padding order, in-place storage behavior, clipping, and area filtering directly in the API docs.

- Add convention blocks and class-anchor pointers for `transform_bbox`, `nms`, and `Boxes.{transform_boxes, transform_boxes_, translate, pad, unpad, clamp, trim, compute_area, filter_boxes_by_area}`.
- Correct the unbatched transformation-matrix claim, the NMS return description, the unpad summary, and the fictitious trim description. Clarify that area filtering zeroes coordinates and preserves cardinality, with inclusive thresholds.
- Add 14 convention/wart test methods, including strict xfails for the area mismatch (#4010) and the newly filed unbatched-container crashes (#4244). The latter covers one and two unbatched boxes, pad, unpad, and both clamp mutation modes. Empty-container transform aliasing is pinned separately.
- Keep existing repairs for #4011, #4019, and #4059 intact. No obsolete warnings or wart expectations are reintroduced.

Runtime tokens and ASTs are unchanged after stripping docstrings/comments. Existing test ASTs are unchanged; all pins are additive. The behavior defects are documented, not repaired here.

Part of #4020; follows merged batch 4a #4060. Batch 4c remains separate.

## Findings and disposition

| Issue | What this batch documents | Classification / disposition |
|---|---|---|
| #4008 | NMS uses exclusive xyxy IoU | Convention difference; KEEP SEPARATE — intentional variant |
| #4010 | Polygon area differs from inclusive width × height, affecting filtering | Existing behavior mismatch; n/a — single symbol |
| #4017 | trim, tuple clamp bounds, and fast translation are unimplemented | API-completeness gap; n/a — single symbol |
| #4244 | Valid unbatched Boxes fail in pad, unpad, and clamp | Valid-input crash, idiom 3; n/a — single symbol |

## Validation

- Full `pixi run pre-commit-all`: passed.
- Both bbox/boxes files on CPU float32/float64: 224 passed, 18 expected strict xfails.
- CPU float16/bfloat16: 212 passed, 18 xfailed, the same 12 failure IDs as freshly measured main.
- MPS float32/float16: 213 passed, 18 xfailed, 8 skipped, the same 3 failure IDs as freshly measured main.
- Full doctests: 646 passed, 15 skipped. Changed-module doctests: 15 passed.
- Sphinx `-W` HTML build: passed; all eleven API entries render their convention block or pointer.
- Mutation checks reject inverted expectations and non-inclusive filter boundaries; the #4244 pins reject a simulated repair / become strict XPASSes as appropriate.
- No known-failure manifest edits. CUDA and alternate PyTorch versions were not run.

Executed from the worktree root with verified import resolution, PyTorch 2.14.0. The half-precision failures are observed baseline failures, not regressions introduced by this documentation/test change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Documented bbox/Boxes geometry-operation conventions and added executable pins for transforms,
+  NMS, padding, clipping, and area filtering, including known unbatched-input limitations. (#4245)
+
 * Documentation pages now provide search-specific titles and descriptions, canonical tutorial links, and a
   canonicalized redirect from the retired highlighted-features page. (#4226)
 
--- a/kornia/geometry/bbox.py
+++ b/kornia/geometry/bbox.py
@@ -544,7 +544,7 @@ def transform_bbox(
         transformed cyclic order.
 
     Args:
-        trans_mat: The transformation matrix to be applied, with shape :math:`(B, 3, 3)`.
+        trans_mat: The transformation matrix to be applied, with supported shape :math:`(B, 3, 3)`.
             For boxes shaped :math:`(N, 4)`, ``B`` is one or ``N``. For boxes shaped
             :math:`(B, N, 4)` or :math:`(B, N, 4, 2)`, it is one or the boxes' ``B``.
         boxes: The boxes to be transformed with a common shape of :math:`(N, 4)` or batched as :math:`(B, N, 4)`, the
@@ -610,7 +610,7 @@ def nms(boxes: torch.Tensor, scores: torch.Tensor, iou_threshold: float) -> torc
 
     .. warning::
         This differs from the inclusive coordinate arithmetic used by several other
-        :mod:`kornia.geometry.bbox` operations and is tracked in
+        :mod:`kornia.geometry.bbox` operations and is documented in
         `#4008 <https://github.com/kornia/kornia/issues/4008>`_.
 
     Example:
--- a/kornia/geometry/bbox.py
+++ b/kornia/geometry/bbox.py
@@ -538,21 +538,26 @@ def transform_bbox(
 ) -> torch.Tensor:
     r"""Apply a transformation matrix to a box or batch of boxes.
 
+    Convention:
+        ``xyxy`` and ``xywh`` inputs are transformed through endpoint pairs. Coordinate
+        restoration sorts their endpoints after a flip; polygon vertices retain their
+        transformed cyclic order.
+
     Args:
-        trans_mat: The transformation matrix to be applied with a shape of :math:`(3, 3)`
-            or batched as :math:`(B, 3, 3)`.
+        trans_mat: The transformation matrix to be applied, with shape :math:`(B, 3, 3)`.
+            For boxes shaped :math:`(N, 4)`, ``B`` is one or ``N``. For boxes shaped
+            :math:`(B, N, 4)` or :math:`(B, N, 4, 2)`, it is one or the boxes' ``B``.
         boxes: The boxes to be transformed with a common shape of :math:`(N, 4)` or batched as :math:`(B, N, 4)`, the
             polygon shape of :math:`(B, N, 4, 2)` is also supported.
         mode: The format in which the boxes are provided. If set to 'xyxy' the boxes are assumed to be in the format
             ``xmin, ymin, xmax, ymax``. If set to 'xywh' the boxes are assumed to be in the format
             ``xmin, ymin, width, height``
-        restore_coordinates: In case the boxes are flipped, adding a post processing step to restore the
-            coordinates to a valid bounding box. Enabled by default (``None`` behaves as ``True``); pass
-            ``False`` to keep the raw transformed corners (the pre-2022 behavior, which yields invalid
-            boxes under flips).
+        restore_coordinates: Reorder endpoints after a flipped ``xyxy`` or ``xywh`` transform.
+            Enabled by default (``None`` behaves as ``True``); pass ``False`` to preserve raw
+            transformed endpoints. Polygon vertices retain their transformed order.
 
     Returns:
-        The set of transformed points in the specified mode
+        The transformed boxes in the specified mode.
 
     """
     if not isinstance(mode, str):
@@ -589,13 +594,24 @@ def transform_bbox(
 def nms(boxes: torch.Tensor, scores: torch.Tensor, iou_threshold: float) -> torch.Tensor:
     """Perform non-maxima suppression (NMS) on tensor of bounding boxes according to the intersection-over-union (IoU).
 
-    Args:
-        boxes: tensor containing the encoded bounding boxes with the shape :math:`(N, (x_1, y_1, x_2, y_2))`.
-        scores: tensor containing the scores associated to each bounding box with shape :math:`(N,)`.
-        iou_threshold: the throshold to discard the overlapping boxes.
+    Convention:
+        Boxes use exclusive ``xyxy`` coordinates. IoU area is
+        ``(x_2 - x_1) * (y_2 - y_1)`` and the result contains kept indices rather
+        than a boolean mask.
 
-    Return:
-        A tensor mask with the indices to keep from the input set of boxes and scores.
+    Args:
+        boxes: tensor containing exclusive ``xyxy`` bounding boxes with shape
+            :math:`(N, 4)`, ordered as ``(x_1, y_1, x_2, y_2)``.
+        scores: tensor containing the scores associated to each bounding box with shape :math:`(N,)`.
+        iou_threshold: the threshold to discard the overlapping boxes.
+
+    Returns:
+        Indices of the boxes kept from the input set, ordered by descending score.
+
+    .. warning::
+        This differs from the inclusive coordinate arithmetic used by several other
+        :mod:`kornia.geometry.bbox` operations and is tracked in
+        `#4008 <https://github.com/kornia/kornia/issues/4008>`_.
 
     Example:
         >>> boxes = torch.tensor([
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -263,6 +263,10 @@ class Boxes:
         `#4012 <https://github.com/kornia/kornia/issues/4012>`_. With ``validate_boxes=True``, vertex modes remain
         unvalidated, and ``'vertices'`` also subtracts one from fixed vertex positions, potentially deforming the
         input rather than rejecting it; this is tracked in `#4177 <https://github.com/kornia/kornia/issues/4177>`_.
+        The unimplemented ``trim``, ``translate(method='fast')``, and tuple-bound ``clamp`` paths are tracked in
+        `#4017 <https://github.com/kornia/kornia/issues/4017>`_. The pad, unpad, and clamp operations fail for
+        unbatched containers even though the class accepts :math:`(N, 4, 2)` data; this is tracked in
+        `#4244 <https://github.com/kornia/kornia/issues/4244>`_.
 
     """
 
@@ -452,9 +456,9 @@ class Boxes:
         ``self`` after adding those two values to every vertex. This operation
         supports only batched :math:`(B, N, 4, 2)` containers.
 
-        .. warning::
-            Unbatched containers raise a broadcasting error; this limitation is
-            tracked in `#4244 <https://github.com/kornia/kornia/issues/4244>`_.
+        Note:
+            Padded :class:`~kornia.augmentation.RandomCrop` uses this method
+            and therefore requires batched boxes.
 
         Args:
             padding_size: Per-batch padding in ``(left, right, top, bottom)``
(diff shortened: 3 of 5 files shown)


ducha-aiki (commented):
Reviewed commit b2d0fdc169661cf64d8df78bac435d8b542276f9. No actionable findings.

The documentation and additive tests match the intended scope. I independently verified that both library modules have identical ASTs after removing docstrings, so this PR introduces no runtime changes.

Validation from the PR worktree, with import resolution verified, using Python 3.11.14 and PyTorch 2.14.0:

- Both geometry test files on CPU float32/float64: 210 passed, 10 expected strict xfails.
- Changed-module doctests: 15 passed.
- `git diff --check`: passed.

Copilot's exception-type concern does not reproduce for the committed test inputs. With two unbatched boxes, both clamp mutation modes raise `RuntimeError` during the broadcast comparison, before boolean indexing can raise `IndexError`. The existing `pytest.raises(RuntimeError)` and restricted strict-xfail therefore work as written.

All reported GitHub checks are green. I did not independently rerun CUDA, MPS, half precision, the full test suite, or the Sphinx build.

ducha-aiki:
Reviewed commit b2d0fdc169661cf64d8df78bac435d8b542276f9 (base `acb861c3`).

**Verdict: changes requested.** Two of the new pins codify an accident of the current implementation as a contract, one of them contradicting the reproduction in its own tracking issue, and one new Convention claim is false for empty containers. The docstring prose is otherwise accurate and renders correctly on the RTD preview. Runtime is unchanged: I re-checked that both library modules have identical ASTs after stripping docstrings.

Everything below was executed from a worktree at the PR head with `kornia.__file__` verified, Python 3.11, PyTorch 2.14.0, CPU.

### P2 — the #4244 clamp pins do not cover the failure the issue reports

Issue #4244's own reproduction is a single unbatched box, and it records `clamp` failing with `IndexError: too many indices for tensor of dimension 2`. The exception type depends on `N`, because the `(1, 4, 4)` bound tensor broadcasts against the `(N, 4)` coordinate view only when `N` is 1 or 4:

| unbatched `N` | `clamp` (both mutation modes) |
|---|---|
| 1 | `IndexError` |
| 2 | `RuntimeError` (broadcast) |
| 4 | `IndexError` |

`pad`/`unpad` raise `RuntimeError` for every `N`.

The PR picks `N=2` and then restricts the strict xfail with `raises=RuntimeError` and asserts `pytest.raises(RuntimeError)` in the wart pin. So Copilot's comment was dismissed on the committed inputs, but it was right about the filed case: the pins pass only because they avoid the reproduction in #4244. A repair that fixes the broadcast path but leaves the boolean-indexing path broken would XPASS the `N=2` xfail and be declared done while the issue's own snippet still crashes.

Ask: parametrize the two #4244 tests over `N in (1, 2)` (or at least add the `N=1` case), and accept `(RuntimeError, IndexError)` in both the wart pin and the xfail's `raises=`. Update the comment that says two boxes avoid a "singleton accident": the singleton case is a *different* failure, not an accident to avoid. Optionally note the `N`-dependence on #4244.

### P2 — `transform_bbox`: the `(3, 3)` rejection is incidental, not a convention

The new test `test_convention_transform_bbox_requires_a_batched_matrix_and_restores_vector_endpoints` and its comment ("A matrix must carry a batch dimension") pin a `ValueError` for a `(3, 3)` matrix. That error comes from `transform_points` comparing `trans_mat.shape[0]` with `boxes.shape[0]`, so it fires only when the leading box dimension is not 3. With exactly three boxes (or `B == 3`), a `(3, 3)` matrix is accepted and applied to every box:

```python
m = torch.tensor([[-1., 0, 10], [0, 1, 0], [0, 0, 1]])
transform_bbox(m, torch.tensor([[1., 1, 3, 2]] * 3))   # works, same result as m[None]
transform_bbox(m, torch.tensor([[1., 1, 3, 2]] * 2))   # ValueError
```

`Boxes.transform_boxes` accepts `(3, 3)` by design for unbatched and singleton-batch containers (it unsqueezes the matrix), so the two entry points disagree, and the pre-PR docstring promised `(3, 3)` for `transform_bbox` too. Documenting `(B, 3, 3)` as the supported shape is the right conservative call, but a `test_convention_*` name plus a `pytest.raises` turns the accident into a contract that a later `transform_points` fix (or a `transform_bbox` unsqueeze) would break.

Ask: either drop the `pytest.raises` from that test and keep only the endpoint-restoration assertions, or move it into a `test_wart_*` with a note that the rejection is shape-dependent, and record the `(3, 3)` inconsistency with `Boxes.transform_boxes` on #4017 or #4020 so it has a home. The docstring can say "a `(3, 3)` matrix is not supported" rather than implying it is always rejected.

### P3 — `compute_area` convention block explains nothing

> The shoelace result uses stored inclusive vertices and can differ from the product returned by `get_boxes_shape`, whose extents use inclusive arithmetic.

Both halves say "inclusive", so the reader never learns why the two disagree, and the class-level block on `Boxes` calls the same thing an "exclusive area convention". The actual relation is that the shoelace formula measures the span *between* the stored inclusive vertices, `(xmax - xmin) * (ymax - ymin) = (w - 1) * (h - 1)`, while `get_boxes_shape` reports `w` and `h` with the `+ 1`. Suggest stating that formula; it also makes the "2-by-1 box has area zero" pin self-explanatory.

### P3 — `transform_boxes_` Convention block is false for an empty container

`_transform_boxes` returns its input tensor untouched when there are zero boxes, so for `Boxes(torch.empty(1, 0, 4, 2))` (and `(0, 4, 2)`) the in-place call returns `self` **and** `data` still aliases the pre-call reference; the non-inplace `transform_boxes` also hands back a clone whose `data` *is* the input's tensor. The augmentation base calls `transform_boxes_` on every batch, including chips with no annotations, so this is a reachable case, and the new pin only covers `N=1`. Add an "except for an empty container" clause, and consider recording the non-inplace aliasing on #4020 as a wart.

### P3 — the unbatched `pad` crash is reachable from `AugmentationSequential`

```python
seq = AugmentationSequential(RandomCrop((6, 6), padding=1, p=1.0), data_keys=["input", "bbox_xyxy"])
seq(torch.rand(1, 3, 8, 8), torch.tensor([[1., 1., 4., 4.]]))   # RuntimeError from Boxes.pad
```

The same `(N, 4)` tensor goes through `RandomHorizontalFlip` fine, so a user whose pipeline works until they add a padded crop gets a broadcast error three frames deep. Neither the new warnings nor #4244 mention it. Suggest one sentence in the `pad` warning naming the crop path, and a note on #4244 so the eventual fix adds an augmentation pin.

### P3 — `trim`: the deleted table was the only spec of `correspondence_preserve`

The removed docstring stated the intended contract (`True` keeps shape and drops all-zero layers, `False` compacts per batch). #4017 only says the method "has a 20-line docstring describing" it, so after this merges the semantics live in `git log -p` alone. Paste the two-mode description into #4017 before this lands, or keep it in the docstring under an "Intended behavior" heading.

### Nits

- `transform_boxes_` renders its Convention block *after* the auto-generated "Return type: Boxes" line on the preview page; the docstring has no `Returns:` section for the typehint extension to attach to. A one-line `Returns:` (as `translate` has) should restore the usual order.
- The two #4244 tests repeat the same two-box literal; a small helper would keep the `N` parametrization above from duplicating it again.
- Batch 4a put every tracked issue for `Boxes` in the single class-level `.. warning::` and left methods with the one-line pointer. This PR adds #4017 and #4244 only as per-method warnings (three copies of the #4244 sentence) and leaves the class block without them; whichever pattern wins, the AGENTS.md grep sweep has to clear every copy when #4244 is fixed, so one line in the class block plus the pointers is cheaper.
- The `nms` warning says "tracked in #4008", but #4008's disposition is KEEP SEPARATE (intentional variant, cross-reference only). "Documented in" avoids implying a repair is pending.
- `test_wart_clamp_tuple_bounds_and_outside_box_behavior_4017` passes `validate_boxes=False` for a box that validates, and both it and the `transform_boxes` in-place test rebuild the fixture from scratch to check non-mutation; `original = boxes.data.clone()` up front says the same thing in one line.
- `test_convention_transform_bbox_xywh_preserves_the_caller_tensor` re-pins what `test_transform_boxes_wh` already asserts with a stronger (flip) case.

### Verified

- Both test files, CPU float32/float64: 210 passed, 10 xfailed (strict), matching the PR description.
- NMS pin: exclusive IoU is 1/4 versus inclusive 4/9 at threshold 0.3, so the pin discriminates; kept-index order `[1, 0]` is descending score (also `[1, 2, 0]` on a three-box case).
- `pad`/`unpad`: only `left`/`top` are applied, in place, batched `(B, 4)`; unbatched raises `RuntimeError` for `N` in {1, 2}.
- `clamp` on a wholly-outside box collapses every vertex to the boundary corner; tuple bounds raise `NotImplementedError`.
- `translate` on an *unbatched* container succeeds with a `(1, 2)` size, so the "supports only batched containers" sentence is correctly absent from that method.
- `compute_area` returns `(N,)` unbatched and `(B, N)` batched.
- `transform_bbox` with `(N, 3, 3)` matrices and `(N, 4)` boxes applies one matrix per box as documented.
- RTD preview: every changed entry renders its Convention block or pointer; issue links resolve.
- Not run: CUDA, MPS, half precision, other PyTorch versions.

ducha-aiki:
Addressed the latest review in `56fa5ca5`.

- Both #4244 pins now cover N=1 and N=2 and accept `IndexError` / `RuntimeError`. Supported-reference errors still become `AssertionError`, so the strict xfail cannot hide them.
- Removed the shape-dependent `(3,3)` rejection from the convention pin and the redundant identity-only xywh pin.
- Clarified the axis-aligned area formula and empty-container tensor-alias exception, with an empty copy/in-place pin and the missing `Returns:` section.
- Consolidated warnings, added padded-crop reachability, and applied the fixture/wording simplifications.
- Preserved the intended trim description on #4017; added the N-dependent exceptions and augmentation reproduction on #4244; recorded the matrix and empty-alias findings on #4020.

Validation: 224 CPU float32/float64 passes and 18 expected strict xfails. CPU half and MPS have exactly the same failure IDs as the main baseline. Changed-module doctests (15), full pre-commit, and Sphinx `-W` pass. Runtime tokens remain unchanged and pre-PR tests remain intact. Independent re-review found no remaining actionable findings.

ducha-aiki:
Re-reviewed the delta `b2d0fdc1..56fa5ca5` (base `acb861c3`), from a worktree at the PR head with `kornia.__file__` verified, Python 3.11, PyTorch 2.14.0.

**Verdict: approved, good to merge.** Both P2 asks are applied as requested, every P3 has landed, and the three issue comments (#4244, #4017, #4020) say what the reply claims. No new findings.

### What I re-checked

- **#4244 pins now include the issue's own case.** Both tests are parametrized over `num_boxes in (1, 2)` and accept `(RuntimeError, IndexError)`. I re-ran the exception probe at the head: `clamp` raises `IndexError` for N=1 and N=4 and `RuntimeError` for N=2 and N=3, both mutation modes; `pad`/`unpad` raise `RuntimeError` for every N. The comment's "indexing failure for one box, broadcasting failure for two" matches. The reference-must-succeed guard still converts a failing batched reference into an `AssertionError`, which the strict xfail's `raises=` tuple does not cover, so a broken reference cannot hide behind the xfail.
- **`transform_bbox` (3, 3) rejection is gone from the pin.** The test keeps only the endpoint-restoration assertions and the docstring says "supported shape", which does not imply a rejection. The shape-dependent acceptance is recorded on #4020.
- **`compute_area` block.** Verified on a `[1, 1, 4, 3]` xyxy box: shoelace area 2, `get_boxes_shape` gives height 2 and width 3, so `(w - 1) * (h - 1)` and `w * h` are the right two expressions.
- **Empty-container aliasing.** Both `(0, 4, 2)` and `(1, 0, 4, 2)`: `transform_boxes_` returns `self` with `data` unchanged in identity, and `transform_boxes` returns a new wrapper whose `data` is the input tensor. The new `test_wart_transform_boxes_empty_copy_aliases_input_4020` asserts exactly that in all four cells.
- **Prose claims in the consolidated class block:** `translate(method='fast')` raises `NotImplementedError`; padded `RandomCrop` inside `AugmentationSequential` fails with an `(N, 4)` box tensor and succeeds with `(B, N, 4)`, so the `pad` note is accurate.
- **AGENTS.md grep:** `#4017` and `#4244` now appear exactly once each in `kornia/`, in the `Boxes` class block, so the eventual fixes have one line to clear.
- **Runtime unchanged:** both library modules have identical ASTs to `acb861c3` after stripping docstrings.
- **Tests, CPU float32/float64:** 224 passed, 18 xfailed (strict), matching the description.
- **CPU float16/bfloat16:** 12 failures, the same 12 node IDs as on `acb861c3`; with `--xfail-known-failures` and the matching profile both dtypes are fully green (117 passed / 12 xfailed, 111 passed / 18 xfailed). No manifest line names either file, so the reparametrization needs no manifest edit.
- **MPS float32** (fallback env var unset): 112 passed, 8 skipped, 9 xfailed, no failures.
- **Pinned Ruff 0.16.4:** check and format clean on the four changed files.
- **RTD preview:** the `transform_boxes_` Convention block now precedes the return information. The preview shows "Return type" ahead of "Returns" for every method on the page, including `translate` and `compute_area`, so that order is the theme's default and my earlier nit was mis-stated; nothing left to change there.

Not run: CUDA, other PyTorch versions, the full docs `-W` build (the RTD check is green).

Issue #4229, opened 2026-09-04
torch.clamp's boundary gradient changed between 2.9.1 and 2.14.0, so kornia's clamp-as-gradient-guard pattern silently fails on the older half of its supported range


ducha-aiki:
## Summary

`torch.clamp`'s gradient **at the clamp bound itself** changed between torch 2.9.1 and 2.14.0: it used to pass
the incoming gradient through (`1.0`), and now returns `0.0`. kornia declares `torch>=2.5.1` and its PR CI runs
**three** torch versions (2.5.1 and 2.14.0 in the main matrix, 2.9.1 on the half-precision legs), so the two ends
of the supported range disagree.

That matters because kornia repeatedly uses `clamp` as a *gradient* guard directly in front of an operator whose
own derivative is unbounded exactly at the clamp bound — `clamp(min=0.0).sqrt()`, `clamp(-1, 1).acos()`,
`clamp(-1, 1).asin()`. Wherever the bound **is** the singular point, the guard silently works on 2.14 and does
**not** work on 2.5.1 / 2.9.1: the clamp hands the unbounded derivative straight through and the result is `inf`
or `nan`.

Measured, same machine, CPU:

| | torch 2.5.1 | torch 2.9.1 | torch 2.14.0 |
| --- | --- | --- | --- |
| `clamp(x, min=0.0)` grad at `x = 0` | `1.0` | `1.0` | **`0.0`** |
| `clamp(w, -1, 1)` grad at `w = 1` | `1.0` | `1.0` | **`0.0`** |
| `clamp(x, min=0.0).sqrt()` grad at `x = 0` | `inf` | `inf` | `0.0` |
| `clamp(w, -1, 1).acos()` grad at `w = 1` | `-inf` | `-inf` | `0.0` |
| bare `acos` grad at `w = 1` | `-inf` | `-inf` | `-inf` |
| bare `asin` grad at `x = 1` | `inf` | `inf` | `inf` |

The last two rows are there to head off the obvious wrong diagnosis: **`acos`/`asin`/`sqrt` did not change**.
Only `clamp`'s treatment of the closed boundary did.

## A reachable instance: NaN gradients from half-precision descriptor matching

`kornia/feature/matching.py:51`

```python
dm = d1_sq.repeat(1, d2.size(0)) + d2_sq.repeat(1, d1.size(0)).t() - 2.0 * d1 @ d2.t()
dm = dm.clamp(min=0.0).sqrt()
```

[...]

PR #4246, opened 2026-09-05
fix(geometry): correct quaternion_to_axis_angle's identity gradient for negative and non-unit quaternions

### LINKED ISSUE #4237: quaternion_to_axis_angle: wrong identity gradient for negative and non-unit quaternions
`quaternion_to_axis_angle` returns a finite but incorrect gradient at identity quaternions other than `(1, 0, 0, 0)`. In particular, `(-1, 0, 0, 0)` is the same unit identity rotation but its vector-component derivatives have the wrong sign. Scaled identity inputs also have the wrong derivative magnitude; the function explicitly permits non-unit quaternions.

## Reproduction

```python
import torch
from kornia.geometry.conversions import quaternion_to_axis_angle

for w in (1.0, -1.0, 2.0, -2.0):
    q = torch.tensor([w, 0.0, 0.0, 0.0], dtype=torch.float64, requires_grad=True)
    jac = torch.autograd.functional.jacobian(quaternion_to_axis_angle, q)
    delta = torch.tensor([0.0, 1e-6, 0.0, 0.0], dtype=q.dtype)
    numerical = (
        quaternion_to_axis_angle(q + delta) - quaternion_to_axis_angle(q - delta)
    ) / (2e-6)
    print(w, jac[0, 1].item(), numerical[0].item())

q = torch.tensor([-1.0, 0.0, 0.0, 0.0], dtype=torch.float64, requires_grad=True)
torch.autograd.gradcheck(quaternion_to_axis_angle, (q,))  # Jacobian mismatch
```

Observed, rounded:

| Real component `w` | Autograd `d output_x / d input_x` | Central difference |
| --- | --- | --- |
| 1 | 2 | 2 |
| -1 | 2 | -2 |
| 2 | 2 | 1 |
| -2 | 2 | -1 |

`gradcheck` passes for `w=1` and fails for each of the other three inputs.

## Expected behavior and cause

For fixed nonzero `w` and a vector part approaching zero, the implemented surrounding map has first-order expansion `axis_angle = (2 / w) * vector_part`. The derivative at the zero vector part should agree with this limit. Both signs of the unit identity must work.

The zero-vector branch instead uses the constant `k_neg = 2.0 * torch.ones_like(sin_theta)` at [conversions.py](https://github.com/kornia/kornia/blob/3f102790/kornia/geometry/conversions.py#L982-L984), independently of `w`. This gives the correct value (zero) but the correct derivative only when `w=1`.

The NaN-gradient repair for #3949 made this branch finite, but the current `test_convention_gradients_at_the_identity_are_finite_3949` only tests the positive unit identity. Its analytic-gradient assertion therefore does not catch this case. The current convention block promises the analytic limit at identity.

## Impact and regression coverage

This is a backward-correctness defect: a valid unit quaternion can produce the opposite gradient direction in pose optimization even though the forward rotation is correct. It should be handled as a broken-output repair, with regression checks for both unit identity signs, non-unit identities, and mixed batches. The all-zero quaternion is a separate degenerate-input policy question; it is not needed to reproduce this issue.

## Verification

Reproduced on Kornia `0.9.0rc1`, commit `3f102790`, PyTorch `2.14.0`, Python `3.11.14`, macOS CPU, float64. Central differences and `torch.autograd.gradcheck` independently expose the mismatch. CUDA and MPS were not checked. The existing convention/wart selection across conversions, bbox/boxes, and geometry transforms remains green: 343 passed, 17 xfailed, 2 skipped on CPU float32/float64.

mohansree14:
## What

`quaternion_to_axis_angle` returns a **finite but wrong gradient** at every identity-like quaternion except the exact positive unit identity `(1,0,0,0)`.

| Real component `w` | Autograd (before) | Correct (central-difference) |
|---|---|---|
| 1 | 2 | 2 ✅ |
| -1 | 2 | **-2** ❌ (wrong sign) |
| 2 | 2 | **1** ❌ (wrong magnitude) |
| -2 | 2 | **-1** ❌ (wrong sign and magnitude) |

`(-1,0,0,0)` represents the **same physical rotation** as `(1,0,0,0)` (quaternions have a double cover: `q` and `-q` are identical rotations), so a gradient with the opposite sign there means pose-optimization could push a parameter in the wrong direction purely depending on which sign of the identity it started from — a real, silent backward-correctness bug.

## Root cause

The zero-vector-part branch (added by the earlier `#3949` NaN-gradient fix) hardcoded the derivative as the constant `k = 2.0`. That's only the `w = 1` case of the function's true analytic limit near any identity-like point: `k = 2 / w`. Every other `w` got the wrong slope. The existing `#3949` regression test only ever exercised `w = 1` — the one point where the wrong constant and the correct formula happen to coincide — so it structurally couldn't have caught this.

## Fix

Replace the constant `2.0` with `2.0 / cos_theta`, guarded at `cos_theta == 0` the same way the function's existing `sin_theta` division is already guarded (substitute a safe argument via `torch.where`, so autograd never divides by zero). `w = 0` only reaches this branch for the fully degenerate all-zero quaternion `(0,0,0,0)` — not a valid rotation; a genuine half-turn has `w = 0` with a *nonzero* vector part and takes the other branch, unaffected.

That guard keeps the degenerate all-zero input's previous harmless `(0,0,0)` forward value with a finite gradient rather than introducing a new `inf`/`nan` — though its specific gradient *value* does change, from the old constant `(0,2,2,2)` to `(0,0,0,0)`. That's not a regression: no gradient is analytically correct at a point with no well-defined limit (different approach paths give different answers), and the issue itself notes this is "a separate degenerate-input policy question."

## Testing

- The issue's own `w ∈ {1,-1,2,-2}` reproduction table now matches central differences exactly.
- `gradcheck` now passes at the negative identity `(-1,0,0,0)` (was failing).
- Forward value verified byte-identical against the previous implementation over 500 random inputs, including several forced onto unit, negative-unit, and scaled-identity cases.
- `torch.compile(fullgraph=True)` compiles cleanly (`backend="eager"` — no C++ toolchain available locally to exercise `inductor` directly).
- Extended the existing shared `_IDENTITY_GRADIENT_CASES` table (which already checks exact gradient values and mixed-batch behavior, not just finiteness — the right infrastructure for this) with the three previously-wrong cases, and added a dedicated test for the all-zero degenerate input's finiteness.
- Both new/extended tests verified to **fail on the unfixed code** (exact 3.0 mismatch matching the expected-vs-wrong delta) and pass on the fix.
- Full `tests/geometry/test_conversions.py` (392 tests): identical pre-existing failure set with and without this change, across float16/bfloat16/float32/float64 — none in the touched test classes.
- Doctests: 34 passed. `ruff check` / `ruff format` clean.
- `grep -rnE "#4237|issues/4237" kornia/ tests/` per `AGENTS.md`'s documented-defect gate — the one `kornia/` hit is unambiguous past tense ("that constant gave the wrong sign").

## Not yet verified

No `pixi`, CUDA, or MPS hardware in my local environment — verified with a throwaway `uv` venv (torch, CPU) instead of the project's usual toolchain.

Fixes #4237

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -620,6 +620,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   so identical inputs can differ at the ulp level between calls unless
   `torch.use_deterministic_algorithms(True)` is set. (#4254)
 
+* `quaternion_to_axis_angle` returns the correct gradient at every identity-like quaternion, not only the
+  positive unit identity `(1, 0, 0, 0)` (#4237). Its zero-vector-part branch used the constant `k = 2.0`,
+  which is only the `w = 1` case of the true analytic limit `k = 2 / w` (`w` = the real component); every
+  other identity-like input got a wrong gradient. That included the negative unit identity `(-1, 0, 0, 0)`
+  -- the same physical rotation as `(1, 0, 0, 0)` under the double cover -- which got the **wrong sign**
+  (`+2` instead of `-2`), and any non-unit "identity" such as `(2, 0, 0, 0)` (a scale this function
+  explicitly permits), which got the **wrong magnitude** (`+2` instead of `+1`). This was a genuine
+  backward-correctness defect, not a missing edge case: a valid unit quaternion could produce the opposite
+  gradient direction in pose optimization depending purely on which sign of the identity it started from.
+  The prior regression test for the `#3949` NaN-gradient fix only exercised `w = 1`, the one point where
+  the wrong constant and the correct formula happen to agree, and so could not have caught this. The
+  division needed for `2 / w` is guarded at `w = 0` the same way the function's existing `sin_theta`
+  division already is, so the fully degenerate all-zero quaternion `(0, 0, 0, 0)` -- not a valid rotation
+  -- keeps its previous harmless `(0, 0, 0)` value with a finite gradient rather than picking up a new
+  `inf`/`nan`; its gradient value itself changes (from the old constant `(0, 2, 2, 2)` to `(0, 0, 0, 0)`),
+  which is not a regression since no gradient is analytically correct at a point with no well-defined
+  limit. The forward value is unaffected everywhere -- verified byte-identical against the previous
+  implementation over 500 random inputs, including several forced onto unit, negative-unit, and scaled
+  identities.
+
 * Non-maxima suppression with a window larger than `(7, 7)` no longer builds a `(k*k, 1, k, k)` one-hot
   convolution to gather each neighbour into its own channel. On CPU in half precision that convolution
   has no vectorized kernel and falls back to `_slow_conv2d_forward`, where it accounted for 99% of a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -633,12 +633,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the wrong constant and the correct formula happen to agree, and so could not have caught this. The
   division needed for `2 / w` is guarded at `w = 0` the same way the function's existing `sin_theta`
   division already is, so the fully degenerate all-zero quaternion `(0, 0, 0, 0)` -- not a valid rotation
-  -- keeps its previous harmless `(0, 0, 0)` value with a finite gradient rather than picking up a new
-  `inf`/`nan`; its gradient value itself changes (from the old constant `(0, 2, 2, 2)` to `(0, 0, 0, 0)`),
-  which is not a regression since no gradient is analytically correct at a point with no well-defined
-  limit. The forward value is unaffected everywhere -- verified byte-identical against the previous
-  implementation over 500 random inputs, including several forced onto unit, negative-unit, and scaled
-  identities.
+  -- keeps its `(0, 0, 0)` forward value. Its gradient there changes, and improves: it was `(0, 2, 2, 2)`
+  on torch 2.14 but already `(nan, 2, 2, 2)` on torch <= 2.9.1, where `atan2(0, 0)`'s derivative with
+  respect to its second argument -- a `0 / 0` -- returns `nan`. That `atan2` is now shielded at this one
+  input as well, so the gradient is `(0, 0, 0, 0)` and finite on every supported torch version. No
+  gradient is analytically correct at a point with no well-defined limit, so the changed value is not a
+  regression. The forward value is unaffected everywhere: the `atan2` shield deliberately takes its value
+  from the unshielded expression, because routing `cos_theta` through `torch.where` hands `atan2` a
+  contiguous tensor instead of a stride-4 view and can select a different kernel. Verified byte-identical
+  against the previous implementation over 2000 random inputs in `float64`, `float32`, `float16` and
+  `bfloat16`, including several forced onto unit, negative-unit, scaled and all-zero identities.
 
 * Non-maxima suppression with a window larger than `(7, 7)` no longer builds a `(k*k, 1, k, k)` one-hot
   convolution to gather each neighbour into its own channel. On CPU in half precision that convolution
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -632,17 +632,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   The prior regression test for the `#3949` NaN-gradient fix only exercised `w = 1`, the one point where
   the wrong constant and the correct formula happen to agree, and so could not have caught this. The
   division needed for `2 / w` is guarded at `w = 0` the same way the function's existing `sin_theta`
-  division already is, so the fully degenerate all-zero quaternion `(0, 0, 0, 0)` -- not a valid rotation
-  -- keeps its `(0, 0, 0)` forward value. Its gradient there changes, and improves: it was `(0, 2, 2, 2)`
-  on torch 2.14 but already `(nan, 2, 2, 2)` on torch <= 2.9.1, where `atan2(0, 0)`'s derivative with
-  respect to its second argument -- a `0 / 0` -- returns `nan`. That `atan2` is now shielded at this one
-  input as well, so the gradient is `(0, 0, 0, 0)` and finite on every supported torch version. No
-  gradient is analytically correct at a point with no well-defined limit, so the changed value is not a
-  regression. The forward value is unaffected everywhere: the `atan2` shield deliberately takes its value
-  from the unshielded expression, because routing `cos_theta` through `torch.where` hands `atan2` a
-  contiguous tensor instead of a stride-4 view and can select a different kernel. Verified byte-identical
-  against the previous implementation over 2000 random inputs in `float64`, `float32`, `float16` and
-  `bfloat16`, including several forced onto unit, negative-unit, scaled and all-zero identities.
+  division already is. That gate is `~pos` -- the mask that actually selects the branch -- and not merely
+  `w != 0`: `2.0 / t` lowers to `t.reciprocal() * 2`, whose backward is `-grad * result**2`, and for a
+  rotation just short of a half turn `w` is small but non-zero, so `(1 / w)**2` overflows to `inf` and
+  meets the exact `0.0` that the unselected branch receives as `0 * inf` -> `nan`. In `float16` that
+  would have been every rotation within ~0.45 degrees of 180 -- ordinary inputs, not degenerate ones.
+  The `2 / w` coefficient is also detached, because on the branch that selects it the vector part is
+  zero, so `d(out)/dw` is exactly zero there and computing it through `-2 / w**2` only reintroduces the
+  same overflow.
+
+  The fully degenerate all-zero quaternion `(0, 0, 0, 0)` -- not a valid rotation -- keeps its
+  `(0, 0, 0)` forward value. Its gradient there changes, and improves: it was `(0, 2, 2, 2)` on torch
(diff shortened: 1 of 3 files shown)


ducha-aiki:
## Changes requested at `81680550` — the fix is right; the new pin fails on the CI floor because of `atan2`, not `clamp`

Reviewed against merge-base `601b5a4a`, torch 2.9.1 and 2.14.0, CPU. Copilot's one inline comment is correct and is folded in below.

### What is right

- **`k = 2/w` is the correct limit and the three new table rows discriminate.** Central differences in float64 vs autograd at the four identity-like points, on both torch versions: `main` returns `[0, 2, 2, 2]` at all four; head returns `[0, -2, -2, -2]`, `[0, 1, 1, 1]`, `[0, -1, -1, -1]` for `w = -1, 2, -2`, matching the finite-difference derivative exactly. Fail-on-base is those three rows, on both versions.
- **Forward is untouched:** `torch.equal` to `main` over 2000 random quaternions plus the forced identities and half turns, in float64/32/16/bf16.

### The blocker: `test_convention_quaternion_to_axis_angle_zero_quaternion_gradient_is_finite_4237` fails on every torch 2.5.1 leg and on the bfloat16 leg (2.9.1)

```
AssertionError: kornia#4237: zero quaternion has a non-finite gradient: tensor([nan, 0., 0., 0.], dtype=torch.float64)
```

Your 2.14 passes it; the CI floor does not. The `nan` sits in `w`'s slot and does not come from your guarded division: it is **`atan2(0, 0)`'s backward**. At the all-zero quaternion `sin_theta` has already been substituted to `0` and `cos_theta` is `0`, so the `two_theta` line evaluates `atan2(0, 0)`, whose derivative w.r.t. the second argument is `-s / (s² + c²) = 0/0`:

```
torch 2.9.1   d/dc atan2(s, c) at (0, 0) = nan
torch 2.14.0  d/dc atan2(s, c) at (0, 0) = 0.0
```

Same shape of problem as #4229 (a boundary derivative that differs across kornia's supported torch range), different operator. It also means the CHANGELOG's "keeps its previous harmless ... finite gradient rather than picking up a new inf/nan" is not true on 2.5.1/2.9.1: `main`'s gradient at `(0, 0, 0, 0)` there is already `[nan, 2, 2, 2]` (`[0, 2, 2, 2]` only on 2.14).

**Fix: shield `atan2`'s input at that one point, the way `pos` already shields the `sqrt`.** I measured two forms.

The obvious one

```python
w_nonzero = cos_theta != 0.0
# atan2(0, 0) has a nan backward on torch <= 2.9.1; feed it (0, 1) at the all-zero quaternion only
safe_cos_for_atan2 = torch.where(pos | w_nonzero, cos_theta, torch.ones_like(cos_theta))
two_theta = 2.0 * torch.where(
    cos_theta < 0.0, torch.atan2(-sin_theta, -safe_cos_for_atan2), torch.atan2(sin_theta, safe_cos_for_atan2)
)
```

fixes the gradient (`[0, 0, 0, 0]` at the zero quaternion on 2.9.1 and 2.14, every other gradient unchanged) **but moves the forward by one ulp on a fraction of random float32/float64 inputs on CPU**. That is not the substitution: `cos_theta` is a stride-4 view of the input, and `torch.where` hands `atan2` a contiguous tensor, which takes a different (vectorised) kernel path. `torch.equal(torch.atan2(s, q[..., 0]), torch.atan2(s, q[..., 0].contiguous()))` is `False` on both versions with no kornia involved.

The form that keeps the forward bit-identical takes the value from the original expression and the gradient through the shielded one (the #4228 pattern):

```python
def _two_theta(cos: torch.Tensor) -> torch.Tensor:
    return 2.0 * torch.where(cos_theta < 0.0, torch.atan2(-sin_theta, -cos), torch.atan2(sin_theta, cos))

safe_cos_for_atan2 = torch.where(pos | w_nonzero, cos_theta, torch.ones_like(cos_theta))
two_theta_safe = _two_theta(safe_cos_for_atan2)
# value from the unshielded expression (bit-identical to before), gradient through the shielded one
two_theta = two_theta_safe + (_two_theta(cos_theta).detach() - two_theta_safe.detach())
```

Measured over 2000 random quaternions plus the forced identities, torch 2.9.1 and 2.14.0: forward `torch.equal` to your head in float64/32/16/bf16; gradients equal wherever both are finite; the zero-quaternion gradient is `[0, 0, 0, 0]` on both versions; exactly one fewer non-finite gradient row than head on 2.9.1 at every dtype, and that row is the zero quaternion. Cost is one extra `atan2` pair. Either form is fine by me; if you take the first, say in the CHANGELOG that the forward can move by one ulp on CPU.

### Two prose fixes

1. **Copilot is right about the code comment** at `conversions.py:990`: "keep that degenerate input's previous harmless (0, zero-gradient) behavior". The previous gradient there was `[0, 2, 2, 2]` (2.14) / `[nan, 2, 2, 2]` (≤ 2.9.1), not zero. Your CHANGELOG already says the gradient *changes*; make the comment agree with it, and name the `atan2` reason next to the shield so the next reader does not simplify it away.
2. **CHANGELOG:** replace "keeps its previous ... finite gradient rather than picking up a new inf/nan" with what holds across the range: on torch ≤ 2.9.1 the old gradient at `(0, 0, 0, 0)` was already `nan` in `w`; with the `atan2` shield it is finite on every supported version.

### Housekeeping

- `CHANGELOG.md` conflicts with today's merges; the source and test files merge cleanly.
- The `half-precision (float16)` job on your run has been running since 14:45 UTC and will be cancelled at the 6-hour cap. That is #4214 (the leg hangs on some Intel runners), not this PR. The bfloat16 leg's failure *is* the pin above.
- Fail-on-base for the shielded pin is the 2.5.1/2.9.1 result; a local 2.14 run passes it before and after and proves nothing.

mohansree14:
Thanks — the `atan2` diagnosis was exactly right, and the second form you measured is what's in `a21e5bd`. Three things changed since `81680550`; the third one you haven't seen and it matters most.

**1. The `atan2` shield, widened from `w == 0` to the whole `~pos` branch.**

Your fix works, but the same `0 * inf` reaches the gradient a second way: `atan2`'s backward multiplies by `1 / (sin_theta**2 + cos_theta**2)`, which is `inf` wherever `w**2` *underflows*, not only where it's exactly zero — `w = 1e-30` in float32, `|w| < 2.4e-4` in float16. Those also produce a `nan` on `main`. Since `two_theta` reaches the output only through `k_pos`, which `k`'s final `where` masks on `pos`, shielding the entire masked branch changes no live gradient and closes both. I used the straight-through form (value from the unshielded expression, gradient through the shielded one), so the forward is bit-identical and no CHANGELOG ulp caveat is needed.

**2. A float16 regression in the version you reviewed.**

This one is on me. The `2 / w` limit is only selected on the zero-vector-part branch, but the division was evaluated for every element and gated on `w != 0` rather than `~pos`. `2.0 / t` lowers to `t.reciprocal() * 2`, backward `-grad * result**2` — so for a rotation just short of a half turn, `w` is small but non-zero, `(1/w)**2` overflows to `inf`, and the exact `0.0` the unselected branch receives meets it as `0 * inf`:

```
main       fp16 [0.003, 1.0, 0.0, 0.0] -> [-2.0, 0.0059, 3.13, 3.13]
81680550   fp16 [0.003, 1.0, 0.0, 0.0] -> [ nan, 0.0059, 3.13, 3.13]
```

In float16 that's every rotation within ~0.45 degrees of 180 — ordinary inputs, finite on `main`. Now gated on `~pos`. The `2 / w` coefficient is also detached, since on the branch that selects it the vector part is zero, so `d(out)/dw` is exactly zero and computing it through `-2 / w**2` only reintroduces the overflow. A parametrized test sweeps both directions across float16/32/64 and fails on base at eight float16 inputs.

**3. The two prose fixes and the conflict.**

The `conversions.py` comment and the CHANGELOG now both say the gradient at `(0,0,0,0)` was `(0, 2, 2, 2)` on 2.14 and already `(nan, 2, 2, 2)` on torch <= 2.9.1. Rebased onto `main`; the CHANGELOG conflict is resolved.

**Measured** on torch 2.7.0 CPU — inside the <= 2.9.1 band, so it reproduces the CI floor (`d/dc atan2(0,0) = nan` there) — over 2000 random quaternions plus forced unit, negative-unit, scaled, all-zero and near-half-turn inputs: forward `torch.equal` to the previous head in float64/float32/float16/bfloat16; gradients equal wherever both are finite; **zero** non-finite gradient rows at every dtype against 1 / 3 / 15 (f64 / f32 / f16) before. The #4237 identity pins are unchanged (`2, -2, 1, -1`). Fail-on-base reproduces your CI message verbatim.

CI is 40/40 green, including all six 2.5.1 legs and bfloat16. The float16 leg finished in 9m29s this time, which supports #4214 being a runner issue rather than this PR.

One trade worth your call: the straight-through form evaluates `atan2` four times per call instead of two, to keep the forward bit-identical. If you'd rather pay a one-ulp forward move on CPU float32/float64 for the cheaper version, say so and I'll switch.

ducha-aiki (approved):
Re-review of exact head a21e5bd33ed5574cb544e47c3b5195f6bdc6330d. The prior supported-version NaN is resolved: atan2 is shielded on the whole zero-vector branch, and the 2/w limit is evaluated only where that branch is selected. The identity gradients now have the correct sign and magnitude for w = -1, 2, and -2, while the near-half-turn and all-zero cases retain finite gradients. The forward-value preservation and mixed-batch coverage are also adequate.

Local validation verified the isolated checkout import and passed the complete CPU float32/float64 conversions file: 627 passed, 2 skipped, 12 expected xfails. All 40 hosted checks are green and GitHub reports the head mergeable.

Issue #4261, opened 2026-09-05
Undeclared optional imports: remove flash_attn / xformers / basicsr / huggingface_hub / safetensors / torchvision, declare boxmot and segmentation_models_pytorch as extras


ducha-aiki:
Follow-up to #4259 (declared dependencies). This one is about the imports that are declared **nowhere**: they sit behind `try: import` / `LazyLoader` and the user only meets them as an `ImportError` at call time. The install page and README list no extras at all.

Inventory on `main` @ 92e5fe8d (AST walk of `kornia/` and `tests/`):

| package | where | how guarded | tested in CI |
|---|---|---|---|
| `flash_attn` | `kornia/feature/lightglue.py:33-41` | `try: import` | no (never installed) |
| `xformers` | `kornia/feature/dedode/transformer/layers/{attention,block,swiglu_ffn}.py` | `try: import` | no (never installed) |
| `basicsr` | `kornia/contrib/super_resolution.py:190-202` | `LazyLoader` | no test at all |
| `huggingface_hub`, `safetensors` | `kornia/models/kimi_vl/builder.py:38-55`, `kornia/models/siglip2/builder.py:36-52` | `try: import` with a good message | `importorskip` in an integration test |
| `torchvision` | `tests/feature/test_aliked.py:81-105` | `skipif` | no (never installed, so the parity test has never run in CI) |
| `boxmot` | `kornia/contrib/boxmot_tracker.py` | `LazyLoader`, `autodoc_mock_imports` | no test at all |
| `segmentation_models_pytorch` | `kornia/models/segmentation/segmentation_models.py` | `LazyLoader`, `autodoc_mock_imports` | no test at all |

The question for each is "can the import go away?" and the answer splits cleanly: six can be removed, two are real integrations that should be declared as extras.

## Remove (the import is dead, or the code behind it is small)

**`flash_attn`** – `kornia/feature/lightglue.py:38` sets `FLASH_AVAILABLE = FlashCrossAttention or hasattr(F, "scaled_dot_product_attention")`, and `Attention.forward` (line 179-189) takes the SDPA branch whenever `has_sdp` is true. SDPA exists on every supported torch (floor 2.5.1), so the `self.flash_(...)` branch and the einsum fallback at line 195-200 are unreachable. Delete the import, the two dead branches and `FLASH_AVAILABLE`; keep the `flash` config key as the `enable_flash_sdp` switch it already is.

[...]

PR #4287, opened 2026-09-05
fix(feature): remove the dead flash_attn path from LightGlue

ducha-aiki:
## What changed

`kornia/feature/lightglue.py` probed for the optional `flash_attn` package at import time and used
the probe result to pick between three code paths in `Attention.forward`: an SDPA path, a
`flash_attn`-backed path, and an `einsum` fallback for torch builds without
`scaled_dot_product_attention`. `F.scaled_dot_product_attention` exists on every torch version
kornia supports (floor 2.5.1), so `has_sdp` was always `True` and the other two branches were dead
code guarded behind a package kornia never declares as a dependency (it isn't in `pyproject.toml`
under any extra).

Removed:
- The `try/except ModuleNotFoundError` import of `flash_attn.modules.mha.FlashCrossAttention` and
  the module-level `FLASH_AVAILABLE` flag.
- `Attention.__init__`'s "FlashAttention is not available..." warning (it could never fire), the
  `self.flash_ = FlashCrossAttention()` construction, and `self.has_sdp`.
- `Attention.forward`'s `flash_attn` branch and its `einsum` fallback branch.
- The three remaining `FLASH_AVAILABLE` reads outside `Attention` (`CrossBlock.__init__`,
  `LightGlue.pruning_min_kpts`), simplified since the flag was always `True`.

Kept unchanged:
- `Attention.__init__(allow_flash)`'s signature and the `torch.backends.cuda.enable_flash_sdp(allow_flash)`
  call, in its original position (now the last statement, since the `has_sdp` guard around it was
  always true).
- The `flash` config key and `TransformerLayer(..., flash=...)` / `CrossBlock(..., flash=...)`
  signatures.
- The CUDA half-cast SDPA branch (`if self.allow_flash and q.device.type == "cuda": ... x.half() ...`)
  byte-for-byte in behaviour — `self.enable_flash` is renamed to `self.allow_flash` (it held the same
  value as the constructor argument once `FLASH_AVAILABLE` is always true) but the branch body is
  untouched.

## Why

Part of the dependency-cleanup effort tracked in kornia#4261: removing optional imports kornia
never declared as a dependency, where the guarded code paths were unreachable given kornia's actual
supported-torch floor.

## What was verified

Run from the worktree root with `/Users/oldufo/dev/kornia/.venv/bin/python`
(torch 2.14.0; `kornia.__file__` confirmed to resolve to this worktree before every run):

- `python -m pytest tests/feature/test_lightglue.py -q`
  → before the change: `30 passed, 1 skipped`; after adding the new `Attention` test:
  `31 passed, 1 skipped`. The pre-existing tests are byte-identical and pass unchanged, confirming
  the surviving SDPA path is the one CI always exercised.
- `python -m pytest tests/test_api_surface.py -q` → `13 passed`. `FLASH_AVAILABLE` was never listed
  in `tests/api_surface.json` and `lightglue.py` has no `__all__`, so no inventory change was
  needed.
- `python -m pytest --doctest-modules kornia/feature/lightglue.py -q` → `no tests ran` (the module
  has no doctests).
- `grep -rn "flash_attn\|FLASH_AVAILABLE\|FlashCrossAttention" kornia tests docs` → empty.
- `python -m ruff check kornia/feature/lightglue.py` → `All checks passed!`
- `git add -A && pre-commit run --all-files` → all hooks passed (ruff check, ruff format,
  codespell, license header, etc.), no files modified by the hooks.

New test added in `tests/feature/test_lightglue.py`:
`test_attention_allow_flash_matches_on_cpu` instantiates `Attention(allow_flash=False)` and
`Attention(allow_flash=True)` and checks they produce `assert_close`-identical outputs on CPU for
the same random `q`/`k`/`v`, both with no mask and with a boolean mask that leaves one query row
fully masked (forcing the `nan_to_num` path to fire). This pins that both constructor settings take
the same SDPA branch on CPU and that mask/`nan_to_num` handling survived the refactor.

## What is not covered

- No CUDA-specific run was performed in this environment (none available); the CUDA half-cast
  branch's code path is unchanged byte-for-byte from before, and CI's existing CPU-only
  `test_lightglue.py` suite is the regression witness for the SDPA path, as scoped by the task
  brief.
- No change was made to `pyproject.toml` / packaging metadata since `flash_attn` was never listed
  there.

Refs #4261.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -293,7 +293,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug fixes
 
 * `kornia.feature.LightGlue` no longer probes for `flash_attn`; the SDPA path it always used is now
-  the only one. (#XXXX)
+  the only one. (#4287)
 
 * `RandAugment`, `AutoAugment`, `TrivialAugment` and `AugMix` no longer silently upcast half-precision
   batches to `float32`. `OperationBase.forward` — the shared gate every auto-augment op routes through —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -292,6 +292,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `kornia.feature.LightGlue` no longer probes for `flash_attn`; the SDPA path it always used is now
+  the only one. (#XXXX)
+
 * `RandAugment`, `AutoAugment`, `TrivialAugment` and `AugMix` no longer silently upcast half-precision
   batches to `float32`. `OperationBase.forward` — the shared gate every auto-augment op routes through —
   moved its `batch_prob` mask to `input.device` but not `input.dtype`, and since `batch_prob` is `float32`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -293,7 +293,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug fixes
 
 * `kornia.feature.LightGlue` no longer probes for `flash_attn`; the SDPA path it always used is now
-  the only one. (#4287)
+  the only one. `kornia.feature.lightglue.Attention.enable_flash`, the public attribute that recorded
+  whether the probe had succeeded, was renamed to `Attention.allow_flash` and now simply mirrors the
+  constructor argument of the same name. (#4287)
 
 * `RandAugment`, `AutoAugment`, `TrivialAugment` and `AugMix` no longer silently upcast half-precision
   batches to `float32`. `OperationBase.forward` — the shared gate every auto-augment op routes through —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -293,9 +293,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug fixes
 
 * `kornia.feature.LightGlue` no longer probes for `flash_attn`; the SDPA path it always used is now
-  the only one. `kornia.feature.lightglue.Attention.enable_flash`, the public attribute that recorded
-  whether the probe had succeeded, was renamed to `Attention.allow_flash` and now simply mirrors the
-  constructor argument of the same name. (#4287)
+  the only one. `kornia.feature.lightglue.Attention.enable_flash`, the attribute that combined the
+  constructor argument with the probe result (and so always equalled the argument, because SDPA is
+  present on every supported torch), was renamed to `Attention.allow_flash`. (#4287)
 
 * `RandAugment`, `AutoAugment`, `TrivialAugment` and `AugMix` no longer silently upcast half-precision
   batches to `float32`. `OperationBase.forward` — the shared gate every auto-augment op routes through —
--- a/kornia/feature/lightglue.py
+++ b/kornia/feature/lightglue.py
@@ -30,16 +30,6 @@ from kornia.core.download import hf_url, load_state_dict_from_url
 from kornia.core.utils import is_exporting
 from kornia.feature.laf import laf_to_three_points, scale_laf
 
-try:
-    from flash_attn.modules.mha import FlashCrossAttention
-except ModuleNotFoundError:
-    FlashCrossAttention = None
-
-if FlashCrossAttention or hasattr(F, "scaled_dot_product_attention"):
-    FLASH_AVAILABLE = True
-else:
-    FLASH_AVAILABLE = False
-
 
 def math_clamp(x, min_, max_):  # type: ignore
     """Clamp a value to lie within [min, max]."""
@@ -146,17 +136,8 @@ class Attention(nn.Module):
 
     def __init__(self, allow_flash: bool) -> None:
         super().__init__()
-        if allow_flash and not FLASH_AVAILABLE:
-            warnings.warn(
-                "FlashAttention is not available. For optimal speed, consider installing flash-attn.",
-                stacklevel=2,
-            )
-        self.enable_flash = allow_flash and FLASH_AVAILABLE
-        self.has_sdp = hasattr(F, "scaled_dot_product_attention")
-        if allow_flash and FlashCrossAttention:
-            self.flash_ = FlashCrossAttention()
-        if self.has_sdp:
-            torch.backends.cuda.enable_flash_sdp(allow_flash)
+        self.allow_flash = allow_flash
+        torch.backends.cuda.enable_flash_sdp(allow_flash)
 
     def forward(
         self, q: torch.Tensor, k: torch.Tensor, v: torch.Tensor, mask: Optional[torch.Tensor] = None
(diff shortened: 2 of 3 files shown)


ducha-aiki:
**Approve after fixes pushed.** The removed `flash_attn` / einsum branches were unreachable on every supported torch, the surviving SDPA path is byte-identical to `main` on CPU, and the one thing worth fixing (the new test leaving a process-global backend flag behind) is pushed in `573de1a2`.

### Verified

Run from the worktree root with the repo `.venv` (torch 2.14.0), `kornia.__file__` checked to name the PR checkout every time.

- Full test file, api-surface and the neighbouring feature tests, at `ab14072b`:
  ```
  python -m pytest tests/feature/test_lightglue.py -q            -> 31 passed, 1 skipped
  python -m pytest tests/test_api_surface.py tests/feature/test_lightglue_onnx.py tests/feature/test_matching.py -q
                                                                  -> 108 passed, 1 skipped
  ```
- Behaviour witness `main` vs PR head, `torch.equal` on 28 tensors: `Attention(False/True)` with and without a
  boolean mask that fully masks one query row (float64), a random-weight `LightGlue(features=None, n_layers=2)`
  forward with `flash=False` and `flash=True` (every output key), and `pruning_min_kpts` for cpu/cuda/mps devices
  under both `flash` settings. Result: `torch.equal mismatches: []`; `pruning_min_kpts(cuda)` is still `1536`
  with `flash=True` and `1024` with `flash=False`, `-1` for cpu/mps, matching what `FLASH_AVAILABLE=True` gave.
- Leftovers: `grep -rnE "flash_attn|FLASH_AVAILABLE|has_sdp|FlashCross|flash_\b|enable_flash|flash-attn" kornia tests docs pyproject.toml pixi.toml .github`
  returns only the kept `torch.backends.cuda.enable_flash_sdp(allow_flash)` call. `docs/source/conf.py`
  `autodoc_mock_imports` never listed `flash_attn`. `import warnings` is still used (line 655).
- `uvx ruff@0.16.4 check` / `format --diff` on the three files: clean (the venv ruff is 0.14.10, so the pinned
  version was used). `pre-commit run --files <the three files>`: all hooks pass.
- Discrimination: with `main`'s `kornia/feature/lightglue.py` swapped in, the new
  `test_attention_allow_flash_matches_on_cpu` still passes (`1 passed`). That is expected and the PR body says
  as much: the test pins that both constructor settings share the SDPA path on CPU and that the boolean-mask /
  `nan_to_num` handling survives, not that the dead branches existed. The `torch.equal` witness above is the
  actual no-change evidence.

### Findings

1. `tests/feature/test_lightglue.py:161` — the new test called `Attention(False)` then `Attention(True)`, so it
   always left `torch.backends.cuda.flash_sdp_enabled()` at `True` regardless of what it found. Probe: setting
   the flag to `False` and running the test left it `True`. Wrapped the body in `try/finally` restoring the
   snapshot; the same probe now reports `found False -> after False`. Fixed in `573de1a2`.
2. `CHANGELOG.md:296` — the entry said `Attention.enable_flash` "recorded whether the probe had succeeded" and
   "now simply mirrors the constructor argument", which implies it used to differ. It was
   `allow_flash and FLASH_AVAILABLE` with `FLASH_AVAILABLE` always `True`, so it always equalled the argument.
   Reworded to say that. Fixed in `573de1a2`.
3. Question for the author, not fixed: `enable_flash` -> `allow_flash` is a rename of a readable attribute with no
   functional effect (`self.enable_flash = allow_flash` would have been a one-line change). Nothing in kornia,
   tests or docs reads it, and `Attention` is not exported from `kornia.feature`, so I left it; if you would
   rather avoid any user-visible rename in a "Bug fixes" entry, reverting to `enable_flash` is trivial.
4. Pre-existing, out of scope: `Attention.__init__` itself flips the process-global `enable_flash_sdp` flag on
   `main` too, and the existing `_make_lightglue(flash=False)` fixtures in `test_lightglue.py` and
   `test_xfeat.py` leave the flag `False` for the rest of a session (probed). Any CUDA test run after them uses
   a non-flash SDPA kernel. This PR does not extend that wart; a follow-up could move the switch into
   `forward` via `torch.nn.attention.sdpa_kernel` or drop it.

### Copilot review

- `tests/feature/test_lightglue.py:177` "snapshot and restore the global flash-SDPA flag": VALID on mechanics
  (the test did leave the flag at `True` when it found `False`), though the constructor-level mutation is
  pre-existing and the existing fixtures already flip it session-wide. Restored via `try/finally` in
  `573de1a2`.

### Not covered

- No CUDA device here, so the half-cast `allow_flash and q.device.type == "cuda"` branch was not executed; its
  body is unchanged from `main` and the witness covers every CPU path. No MPS run either (the code has no MPS
  branch).
- The pretrained-weights test (`features="disk"`) stayed skipped; no downloads were made. No docs build.

ducha-aiki:
**Final review: Approve.** Full re-review of the PR at head `573de1a2` (three files, `+52/-42` against `origin/main` `e8111a61`), not a delta. Every earlier finding is fixed in the code, the surviving SDPA path is bit-identical to `main` on CPU, nothing references the removed names, and CI is green except for one hung float16 job that I restarted.

`kornia/feature/lightglue.py` loses the `flash_attn` import probe, `FLASH_AVAILABLE`, `has_sdp`, the `flash_attn` and `einsum` branches of `Attention.forward`, and the three `FLASH_AVAILABLE` reads in `CrossBlock.__init__` / `pruning_min_kpts`. `F.scaled_dot_product_attention` exists on every torch kornia supports (floor 2.5.1), so all of that was dead. `Attention.enable_flash` (always equal to the constructor argument) becomes `allow_flash`; the `enable_flash_sdp(allow_flash)` call and the CUDA half-cast branch are unchanged. One new CPU test pins that both constructor settings share the SDPA path (with and without a fully-masked row) and restores the process-global flash-SDPA flag it flips. CHANGELOG bullet under `Unreleased / Bug fixes` cites #4287 and states the rename correctly.

### Verified at `573de1a2`

Worktree root, `.venv` python (torch 2.14.0), `kornia.__file__` naming the checkout in every block.

```
pytest tests/feature/test_lightglue.py --dtype=float32,float64         -> 48 passed, 1 skipped
pytest tests/feature/test_lightglue.py --dtype=float16,bfloat16        -> 41 passed, 8 skipped
pytest tests/test_api_surface.py tests/feature/test_lightglue_onnx.py tests/feature/test_matching.py tests/feature/test_xfeat.py
                                                                        -> 136 passed, 23 skipped
uvx ruff@0.16.4 check / format --diff on both .py files                 -> clean
pre-commit run --files (3 files)                                        -> all hooks pass
uvx ty check kornia/feature/lightglue.py                                -> 1 diagnostic, the pre-existing custom_fwd deprecation at line 44 (untouched)
```

- Witness, `origin/main`'s `lightglue.py` loaded side by side with the head module, `torch.equal` on every output: `Attention(False/True)` with and without the boolean mask (float64); random-init `LightGlue(features=None, n_layers=2, input_dim=64)` forward for `flash=False` and `flash=True` (every output key, tensors, lists and the scalar `stop`); `pruning_min_kpts` for cpu/cuda/mps under both settings (`-1 / 1024 / -1` and `-1 / 1536 / -1`). Result: `torch.equal mismatches: []`. Loading `main`'s module raised no warning either, confirming the removed "FlashAttention is not available" warning could never fire.
- Flag restore: `enable_flash_sdp(False)`, run the new test, flag is still `False` afterwards (earlier round: it was left `True`).
- Leftovers: `grep -rnE "flash_attn|FLASH_AVAILABLE|has_sdp|FlashCross|enable_flash\b" kornia tests docs pyproject.toml pixi.toml .github` → nothing; `import warnings` and `KORNIA_CHECK` still have users in the file, so no unused import survives the removal.
- CI on the head: 39/42 checks green, the two `-/-` rows are the RTD and pre-commit.ci contexts that never report here. The `half-precision (float16) / tests (3.11, 2.9.1)` job had been running 2h21m against a 3m18s bfloat16 twin, which is the Intel-runner float16 hang (kornia#4214), not this PR; I cancelled run 34011794927 and re-ran the job.

### Findings

None blocking. Two notes carried over, both deliberately out of scope for this PR:

1. `Attention.__init__` flips the process-global `torch.backends.cuda.enable_flash_sdp` on `main` too, and the existing `_make_lightglue(flash=False)` fixtures leave it `False` for the rest of a CUDA session. The new test is the only one that restores it. A follow-up could move the switch into `forward` via `torch.nn.attention.sdpa_kernel`.
2. `enable_flash` → `allow_flash` is a readable-attribute rename in a "Bug fixes" entry. Nothing in kornia, tests or docs reads it and `Attention` is not exported from `kornia.feature`, so I am fine with it as written.

### Not covered

No CUDA device: the `allow_flash and q.device.type == "cuda"` half-cast branch was not executed (body unchanged from `main`). No MPS run (no MPS branch exists). The pretrained-weights test stayed skipped; no downloads. No docs build.

PR #4288, opened 2026-09-05
fix(feature): drop the dead `xformers` paths from DeDoDe's vendored DINOv2

ducha-aiki:
## What changed

Three vendored DINOv2 files under `kornia/feature/dedode/transformer/layers/` tried to import
`xformers.ops` at module load and kept a second, xformers-backed implementation behind the
resulting `XFORMERS_AVAILABLE` flag. kornia has never declared `xformers` as a dependency and it
is installed in no CI environment, so those imports always failed and only the pure-PyTorch
fallbacks ever ran. Importing `kornia.feature` even emitted a `logger.warning("xFormers not
available")` for a package the user was never asked to install.

- `attention.py`: the `try: from xformers.ops import fmha, memory_efficient_attention, unbind`
  block and `XFORMERS_AVAILABLE` are gone. `MemEffAttention.forward(x, attn_bias=None)` keeps its
  signature and its class name (the vendored DINOv2 builders select it by name), rejects a
  non-`None` `attn_bias` with `NotImplementedError("attn_bias is not supported")`, and otherwise
  computes exactly what the previous non-xformers branch computed (`super().forward(x)`).
  The rejection used to be `ValueError("xFormers is required for nested tensors usage")` — a path
  that could only ever raise, since `attn_bias` was only ever produced by the removed list path.
- `block.py`: the import, the `xFormers not available` warning, the module logger that existed
  only for that warning, `XFORMERS_AVAILABLE`, and the helpers used only by the nested-tensor list
  path (`get_branges_scales`, `add_residual`, `attn_bias_cache`, `get_attn_bias_and_cat`,
  `drop_add_residual_stochastic_depth_list`) are removed, along with
  `NestedTensorBlock.forward_nested`. `NestedTensorBlock.forward(x_or_x_list)` keeps its name and
  its Tensor behaviour; a list now raises `TypeError` saying the nested-tensor list path was
  removed (it previously failed a `KORNIA_CHECK`, i.e. `kornia.core.exceptions.BaseError`).
  `add_residual` also called `xformers.ops.scaled_index_add` in its `scaling_vector` branch, so
  that branch was a `NameError` waiting to happen.
- `swiglu_ffn.py`: the import is gone and `SwiGLUFFNFused` subclasses `SwiGLUFFN` directly, which
  is what the `SwiGLU = SwiGLUFFN` fallback already made it resolve to. Its class name and
  constructor signature are unchanged.
- `kornia/feature/dedode/transformer/layers/__init__.py` is untouched; every exported name is the
  same.

## Why

Part of #4261: kornia optionally imports packages it never declares as dependencies, which makes
the behaviour of the library depend on whatever happens to be in the user's environment, hides
untested code paths from CI, and (here) prints a warning about a package nobody asked for. The
surviving path is the one CI has always exercised.

## What was verified

All commands run from the branch worktree with `/Users/oldufo/dev/kornia/.venv/bin/python`
(torch 2.14.0); `kornia.__file__` was printed and confirmed to name the worktree checkout.

- Byte-identity witness. Same script run before and after the edits, saving keypoints/scores/
  descriptions of `kornia.feature.DeDoDe(detector_model="L", descriptor_model="G",
  amp_dtype=torch.float32)` (which is the configuration that actually instantiates the vendored
  DINOv2 ViT-L/14 encoder) in eval mode, on CPU float32, `torch.manual_seed(0)` before both model
  construction and input creation, on a `1x3x64x64` `torch.randn` image with `n=64`:

  ```
  python - <before.pt path> < witness.py     # on e8111a61
  python - <after.pt path>  < witness.py     # on this branch
  python - < compare.py                      # torch.equal on each tensor
  ```

  Result: `keypoints (1, 64, 2) torch.equal: True`, `scores (1, 64) torch.equal: True`,
  `descriptions (1, 64, 256) torch.equal: True`, `all equal: True`. The DINOv2 ViT-L/14 backbone
  used pretrained weights (downloaded to the torch hub cache); the VGG stem and the DeDoDe
  detector/descriptor heads were seeded random init, as `DeDoDe(...)` (not `from_pretrained`)
  builds them. The "before" run also printed `xFormers not available`; the "after" run does not.

- `python -m pytest tests/feature/test_dedode.py -q` → `4 passed, 2 skipped`. The pre-existing
  `TestDeDoDe` class is unchanged (it is `@pytest.mark.skip`-ed as unmaintained on main) and
  `TestConvRefiner` is untouched. Two new tests were added:
  `TestNestedTensorBlock` asserts `NestedTensorBlock(dim=8, num_heads=2)` on a `(1, 5, 8)` tensor
  returns shape `(1, 5, 8)` and is not the identity, and that a list input raises `TypeError`;
  `TestMemEffAttention::test_attn_bias_raises` pins the new `NotImplementedError`. Both new
  exception pins fail on the base revision (`BaseError` and `ValueError` respectively), so they
  discriminate.
- `python -m pytest tests/feature/ tests/test_api_surface.py -q` → `735 passed, 98 skipped`.
- `python -m pytest --doctest-modules kornia/feature/dedode/transformer/layers/ -q` → `no tests
  ran` (these vendored layer modules carry no doctest examples);
  `--doctest-modules kornia/feature/dedode/ -q` → `1 skipped` (the `DeDoDe` example is the
  weights-download-gated one).
- `grep -rn "xformers\|XFORMERS\|xFormers" kornia tests docs benchmarks` → no matches.
- `git add -A && pre-commit run --all-files` → all hooks Passed.
- `uvx "ty>=0.0.9,<0.0.10" check --python .venv kornia/feature/dedode` → All checks passed.

## What is not covered

- Nothing here was run on CUDA or MPS, and no half-precision run was done. The change is a
  deletion of never-executed branches plus one delegation, so device and dtype behaviour is the
  Tensor path's, unchanged.
- The witness covers `DeDoDe` end to end with the `G` descriptor (the only builder that
  instantiates DINOv2). The `B` descriptor does not use these layers at all.
- The removed list path cannot be tested, on this branch or on main, without installing
  `xformers`; it has never been exercised by kornia's tests.
- Both module-level `logger`s are gone: `block.py`'s existed only for the `xFormers not available`
  warning, and `attention.py`'s was never used.
- One level up, `DinoVisionTransformer.forward_features` in `dinov2.py` still branches on
  `isinstance(x, list)` into `forward_features_list`; that path raised before this change (the
  `KORNIA_CHECK` on `XFORMERS_AVAILABLE`) and now raises `TypeError` from the block. Removing the
  branch itself is a follow-up under #4261.
- A user who had installed `xformers` by hand loses the nested-tensor list path. kornia never
  declared, documented, or tested it, which is why this is filed under bug fixes rather than
  breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -292,6 +292,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* DeDoDe's vendored DINOv2 no longer probes for `xformers`; the pure-PyTorch path it always ran is the
+  only one. (#XXXX)
+
 * `RandAugment`, `AutoAugment`, `TrivialAugment` and `AugMix` no longer silently upcast half-precision
   batches to `float32`. `OperationBase.forward` — the shared gate every auto-augment op routes through —
   moved its `batch_prob` mask to `input.device` but not `input.dtype`, and since `batch_prob` is `float32`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -293,7 +293,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug fixes
 
 * DeDoDe's vendored DINOv2 no longer probes for `xformers`; the pure-PyTorch path it always ran is the
-  only one. (#XXXX)
+  only one. (#4288)
 
 * `RandAugment`, `AutoAugment`, `TrivialAugment` and `AugMix` no longer silently upcast half-precision
   batches to `float32`. `OperationBase.forward` — the shared gate every auto-augment op routes through —
--- a/kornia/feature/dedode/transformer/layers/attention.py
+++ b/kornia/feature/dedode/transformer/layers/attention.py
@@ -32,14 +32,6 @@ from torch import Tensor, nn
 logger = logging.getLogger("dinov2")
 
 
-try:
-    from xformers.ops import fmha, memory_efficient_attention, unbind
-
-    XFORMERS_AVAILABLE = True
-except ImportError:
-    XFORMERS_AVAILABLE = False
-
-
 class Attention(nn.Module):
     """Implement the standard Multi-Head Self-Attention mechanism.
 
@@ -97,31 +89,22 @@ class Attention(nn.Module):
 
 
 class MemEffAttention(Attention):
-    """Implement a memory-efficient version of Multi-Head Self-Attention using xFormers."""
+    """Implement Multi-Head Self-Attention with the signature the vendored DINOv2 blocks call.
+
+    The computation is the one of :class:`Attention`. The extra ``attn_bias`` argument is kept for
+    signature compatibility with the DINOv2 model builders that select this class, and must be ``None``.
+    """
 
     def forward(self, x: Tensor, attn_bias=None) -> Tensor:  # type: ignore[no-untyped-def]
         """Run this DeDoDe module forward.
 
         Args:
             x: Input token sequence with shape :math:`(B, N, C)`.
-            attn_bias: Optional attention bias from xFormers for nested tensor support.
+            attn_bias: Unsupported; must be ``None``.
 
         Returns:
             Attention output with shape :math:`(B, N, C)`.
         """
-        if not XFORMERS_AVAILABLE:
-            if attn_bias is not None:
-                raise ValueError("xFormers is required for nested tensors usage")
-            return super().forward(x)
-
-        B, N, C = x.shape
-        qkv = self.qkv(x).reshape(B, N, 3, self.num_heads, C // self.num_heads)
-
-        q, k, v = unbind(qkv, 2)
-
-        x = memory_efficient_attention(q, k, v, attn_bias=attn_bias)
-        x = x.reshape([B, N, C])
-
-        x = self.proj(x)
-        x = self.proj_drop(x)
-        return x
+        if attn_bias is not None:
+            raise NotImplementedError("attn_bias is not supported")
+        return super().forward(x)
--- a/kornia/feature/dedode/transformer/layers/attention.py
+++ b/kornia/feature/dedode/transformer/layers/attention.py
@@ -85,10 +85,11 @@ class Attention(nn.Module):
 
 
 class MemEffAttention(Attention):
-    """Implement Multi-Head Self-Attention with the signature the vendored DINOv2 blocks call.
+    """Implement Multi-Head Self-Attention under the class name the vendored DINOv2 builders select.
 
-    The computation is the one of :class:`Attention`. The extra ``attn_bias`` argument is kept for
-    signature compatibility with the DINOv2 model builders that select this class, and must be ``None``.
+    The computation is the one of :class:`Attention`. The class name is kept because the DINOv2 model
+    builders reference it; the extra ``attn_bias`` argument is kept so :meth:`forward` matches the upstream
+    DINOv2 signature. Nothing in kornia passes it, and it must be ``None``.
     """
 
     def forward(self, x: Tensor, attn_bias=None) -> Tensor:  # type: ignore[no-untyped-def]
--- a/kornia/feature/dedode/transformer/layers/attention.py
+++ b/kornia/feature/dedode/transformer/layers/attention.py
@@ -25,11 +25,7 @@
 #   https://github.com/facebookresearch/dino/blob/master/vision_transformer.py
 #   https://github.com/rwightman/pytorch-image-models/tree/master/timm/models/vision_transformer.py
 
-import logging
-
 from torch import Tensor, nn
-
-logger = logging.getLogger("dinov2")
 
 
 class Attention(nn.Module):
--- a/kornia/feature/dedode/transformer/layers/block.py
+++ b/kornia/feature/dedode/transformer/layers/block.py
@@ -25,29 +25,15 @@
 #   https://github.com/facebookresearch/dino/blob/master/vision_transformer.py
 #   https://github.com/rwightman/pytorch-image-models/tree/master/timm/layers/patch_embed.py
 
-import logging
-from typing import Any, Callable, Dict, List, Tuple
+from typing import Callable
 
 import torch
 from torch import Tensor, nn
 
-from kornia.core.check import KORNIA_CHECK
-
-from .attention import Attention, MemEffAttention
+from .attention import Attention
 from .drop_path import DropPath
 from .layer_scale import LayerScale
 from .mlp import Mlp
-
-logger = logging.getLogger("dinov2")
-
-
-try:
-    from xformers.ops import fmha, index_select_cat, scaled_index_add
-
-    XFORMERS_AVAILABLE = True
-except ImportError:
-    logger.warning("xFormers not available")
-    XFORMERS_AVAILABLE = False
 
 
 class Block(nn.Module):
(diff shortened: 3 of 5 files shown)


ducha-aiki:
**Approve after fixes pushed.** The three vendored layer files lose only code that was unreachable without `xformers`; every surviving Tensor path is bit-identical to `main` on CPU, the two new exception pins fail on `main`, and the one thing I changed is a docstring that misattributed why `attn_bias` survives.

### Verified

All from the worktree root with `.venv/bin/python` (torch 2.14.0), `kornia.__file__` printed and naming the checkout under test.

- Behaviour witness, `main` vs `97954ef0`, same seeds and weights, `torch.equal` on every tensor: `MemEffAttention` (with and without `attn_bias=None`), `NestedTensorBlock` in eval, in train with `drop_path=0.2` (stochastic-depth branch) and `drop_path=0.05` (`DropPath` branch), `SwiGLUFFNFused` (weights, output and MRO), and `vit_small(patch_size=14, img_size=56)` random-init `forward_features` for both `ffn_layer="mlp"` and `"swiglufused"` (`x_norm_clstoken`, `x_norm_patchtokens`, `x_prenorm`, state-dict keys):

  ```
  all tensors equal: True
  ```

  Only the exception types moved, as the body says: `attn_bias` → `ValueError` becomes `NotImplementedError`; a list into the block or into `DinoVisionTransformer.forward_features` → `BaseError` becomes `TypeError`; a non-Tensor non-list (`ndarray`) → bare `AssertionError` becomes the same `TypeError`. The `main` run printed `xFormers not available`, the PR run did not.
- Discrimination: with `main`'s three layer files checked out over the PR tree, `-k "MemEff or NestedTensor"` → `2 failed, 1 passed` (both `*_raises` pins fail with the old exception types). Restored; tree clean.
- `pytest tests/feature/test_dedode.py` → `4 passed, 2 skipped` (the `TestDeDoDe` slow tests); `--dtype=float16,bfloat16` on the new tests → `6 passed`. `pytest tests/feature/ tests/test_api_surface.py` → `735 passed, 98 skipped`. `tests/core/test_weights_prefetch.py` → `84 passed`. `--doctest-modules kornia/feature/dedode/` → `1 skipped`.
- `uvx ruff@0.16.4 check` / `format --diff` on the four changed files → clean (venv ruff is 0.14.10, so I used the pinned one). `uvx "ty>=0.0.9,<0.0.10" check --python .venv kornia/feature/dedode` → All checks passed.
- `grep -rn "xformers\|XFORMERS\|xFormers\|memory_efficient_attention\|fmha\|scaled_index_add\|index_select_cat\|forward_nested\|attn_bias_cache" kornia tests docs pyproject.toml pixi.toml .github benchmarks` → no matches; `autodoc_mock_imports` never listed it; `tests/api_surface.json` and `docs/source/*.rst` contain no `dedode` entry, so the removed module-level helpers were in no inventory. CHANGELOG entry sits under `## Unreleased` / `### Bug fixes` and cites #4288.

### Findings

1. `kornia/feature/dedode/transformer/layers/attention.py:88-91` — the new `MemEffAttention` docstring said `attn_bias` is "kept for signature compatibility with the DINOv2 model builders that select this class". The builders only reference the class by name (`partial(Block, attn_class=MemEffAttention)`), and `Block.forward` calls `self.attn(self.norm1(x))` with one argument; `grep -rn attn_bias kornia/` finds no caller. Reworded: the name is kept for the builders, the argument for the upstream DINOv2 `forward` signature, and nothing in kornia passes it. Fixed in `dd207a27`.
2. PR body, "What is not covered": it says `attention.py` "still defines an unused module-level `logger`" and is "left alone", but `97954ef0` (the head commit at review time) removed it. The body is stale on that bullet; the code is right.
3. Question, not fixed (outside this PR's files, and already misleading on `main`): `kornia/feature/dedode/transformer/dinov2.py:296` `forward_features` still documents "or list of tensors" and dispatches a list to `forward_features_list`, which now can only run `prepare_tokens_with_masks` per item and then hit the block's `TypeError`. The author defers this to #4261; I agree that is the place, but the docstring line could go with it.

Note for #4261's plan: the issue text says `MemEffAttention.forward` "becomes SDPA". This PR deliberately delegates to `Attention.forward` (manual softmax) instead, which is what makes the witness bit-identical; SDPA would not be. Worth reflecting in the tracking issue.

### Copilot review

- "Unsupported paths fail fast with explicit exceptions": VALID, witnessed above (`NotImplementedError` and `TypeError`, raised before any attention compute; the `forward_features` list path still runs token preparation first, which is the deferred item 3).
- "`SwiGLUFFNFused` consistently inherits from `SwiGLUFFN` while keeping DINOv2 hidden-dim rounding": VALID, MRO and `w12`/`w3` shapes identical on both sides, and the `(int(h * 2 / 3) + 7) // 8 * 8` rounding matches upstream DINOv2.
- "Approval recommended", no inline comments: consistent with what I found.

### Not covered

CUDA, MPS and CUDA/MPS half precision (CPU float16/bfloat16 only). The body's end-to-end `DeDoDe(detector_model="L", descriptor_model="G")` witness needs the 1.2 GB DINOv2 ViT-L/14 checkpoint, which I did not download; my witness covers the same layers through a random-init `vit_small` instead. The removed list path itself, which needs `xformers` on either side. Docs build.

ducha-aiki:
**Final review: Approve.** Full re-review at head `dd207a27` (five files, `+63/-175` against `origin/main` `e8111a61`), not a delta. The three vendored layer files lose only code that was unreachable without `xformers`; every surviving Tensor path is bit-identical to `main`, the two new exception pins discriminate, nothing in the tree references the removed names, and CI is fully green.

- `attention.py`: the `xformers.ops` probe, `XFORMERS_AVAILABLE` and the unused module logger are gone. `MemEffAttention` keeps its name (the DINOv2 builders select it by name) and the `attn_bias` argument (upstream signature); it delegates to `Attention.forward` and raises `NotImplementedError` on a non-`None` bias. The docstring says exactly that, and no longer claims the builders pass the argument.
- `block.py`: the probe, the `xFormers not available` warning and its logger, and the five list-path helpers plus `NestedTensorBlock.forward_nested` are removed. `forward` accepts a Tensor and raises `TypeError` for anything else (a list used to fail a `KORNIA_CHECK`, an ndarray a bare `AssertionError`).
- `swiglu_ffn.py`: `SwiGLUFFNFused` subclasses `SwiGLUFFN` directly, which is what the `SwiGLU = SwiGLUFFN` fallback already resolved to; the DINOv2 `(2h/3 + 7) // 8 * 8` hidden rounding is unchanged.
- Tests: `TestMemEffAttention.test_attn_bias_raises`, `TestNestedTensorBlock.test_tensor_input` / `test_list_input_raises`, all `BaseTester` classes on the injected `device`/`dtype`. CHANGELOG bullet under `Unreleased / Bug fixes` cites #4288.

### Verified at `dd207a27`

```
pytest tests/feature/test_dedode.py --dtype=float32,float64     -> 7 passed, 4 skipped (TestDeDoDe is skip-marked on main)
pytest tests/feature/test_dedode.py --dtype=float16,bfloat16    -> 7 passed, 4 skipped
pytest tests/feature/ tests/test_api_surface.py                 -> 735 passed, 98 skipped
pytest --doctest-modules kornia/feature/dedode/                 -> 1 skipped (download-gated DeDoDe example)
uvx ruff@0.16.4 check / format --diff (4 .py files)             -> clean
pre-commit run --files (5 files)                                -> all 9 hooks pass
uvx ty check kornia/feature/dedode                              -> All checks passed
```

- Witness, `origin/main`'s three layer files (plus `main`'s `dinov2.py` re-pointed at them) loaded as a sibling package, same seeds, `torch.equal` on every tensor: `MemEffAttention` with and without `attn_bias=None` (float64); `NestedTensorBlock` at `drop_path` 0.0/0.2/0.05 in eval and train (stochastic-depth and `DropPath` branches, seeded); `SwiGLUFFNFused` weights, output and MRO (`SwiGLUFFNFused, SwiGLUFFN, Module` on both sides, `w12` `(96, 16)`, `w3` `(16, 48)`); random-init `vit_small(patch_size=14, img_size=56)` `forward_features` for `ffn_layer="mlp"` and `"swiglufused"`, state-dict keys and every output tensor. Result: `torch.equal mismatches: []`. Loading `main`'s package printed `xFormers not available`; the head did not.
- Exception types moved exactly as the body says: `attn_bias` `ValueError -> NotImplementedError`; list `BaseError -> TypeError`; ndarray `AssertionError -> TypeError`.
- Discrimination: with `main`'s three files checked out over the head, `-k "MemEff or NestedTensor"` gives `2 failed, 1 passed` (both `*_raises` pins fail on the old exception types); restored, tree clean.
- Leftovers: `grep -rnE "xformers|XFORMERS|xFormers|memory_efficient_attention|fmha|scaled_index_add|index_select_cat|forward_nested|attn_bias_cache" kornia tests docs pyproject.toml pixi.toml .github benchmarks` → no match. `layers/__init__.py` is untouched, so every exported name survives; `tests/api_surface.json` and `docs/source/*.rst` have no `dedode` entry.
- CI on the head: 41/43 green; the two `-/-` rows are the RTD and pre-commit.ci contexts that never report.
- PR body: the one stale bullet (it said `attention.py` "still defines an unused module-level `logger`"; the head removes it) is corrected in the body now.

### Findings

None blocking. One note carried over, out of scope by the author's own plan: `dinov2.py:296` `forward_features` still documents "or list of tensors" and dispatches a list to `forward_features_list`, which can now only run `prepare_tokens_with_masks` per item before hitting the block's `TypeError`. Removing that branch and the docstring line is the #4261 follow-up; note for that issue that `MemEffAttention.forward` deliberately delegates to `Attention.forward` (manual softmax), not SDPA, which is what keeps the witness bit-identical.

### Not covered

CUDA and MPS (CPU only, float16/bfloat16 included). The end-to-end `DeDoDe(detector_model="L", descriptor_model="G")` witness in the body needs the 1.2 GB ViT-L/14 checkpoint, which I did not download; the random-init `vit_small` witness above covers the same layers. The removed list path itself, which needs `xformers` on either side. Docs build.

PR #4289, opened 2026-09-05
test(feature): replace the torchvision deform_conv2d parity test with a hardcoded literal

ducha-aiki:
## What changed

`tests/feature/test_aliked.py::TestDeformConv2d::test_matches_torchvision` compared
`kornia.feature.aliked.deform_conv2d.deform_conv2d` against `torchvision.ops.deform_conv2d`,
guarded by `@pytest.mark.skipif(not _has_torchvision(), ...)`. kornia has never declared
`torchvision` as a dependency, and no CI environment installs it, so this test has never run
there — only ever locally, by hand, by whoever happened to have torchvision on their machine.

- Deleted `_has_torchvision()` and `test_matches_torchvision`.
- Added `test_matches_torchvision_reference` on the same `TestDeformConv2d` class (now a
  `testing.base.BaseTester` subclass, for `self.assert_close`), parametrized over
  `use_mask in (False, True)` and driven by the suite's injected `device`/`dtype` fixtures.
  It builds deterministic float64 CPU inputs (`B=1, C_in=2, H=W=4, C_out=2, kH=kW=3, padding=1,
  stride=1`) via `_torchvision_reference_inputs()` (`torch.linspace`/`torch.arange`/`sin`/`cos`/
  `sigmoid`, no RNG), casts them to the fixture's `device`/`dtype`, and compares kornia's output
  against one of two hardcoded `float64` literals (`_EXPECTED_NO_MASK`, `_EXPECTED_WITH_MASK`,
  32 numbers each) generated once offline with torchvision itself — see the comment above
  `_EXPECTED_NO_MASK` for the exact generation snippet and versions
  (`torch==2.14.0`, `torchvision==0.29.0`, CPU wheels, Python 3.11.0).
  - float64-on-CPU compares at `rtol=0, atol=1e-9` (the literal is quoted to 10 decimals and
    kornia's float64 result lands within 5e-11 of it); every other device/dtype combination
    (including MPS, which skips float64) uses `BaseTester`'s default per-dtype tolerance.
  - One extra assertion recomputes kornia's result at float32 (regardless of which dtype the
    test instance is parametrized over) and checks it against the float64 literal at
    `atol=1e-5`, to catch a float32 precision regression.
- No other torchvision references remain in this file; `test_gradients`,
  `test_zero_offset_matches_regular_conv`, and `test_output_shape` are untouched.

## Why

Part of #4261: kornia optionally imports packages it never declares as dependencies, which
hides untested code paths from CI. Here, the hidden path was an entire test — CI has silently
never exercised `deform_conv2d` against the torchvision reference it claims to match, on any
platform, ever. The hardcoded literal makes the comparison run everywhere, at both float32 and
float64, without adding a runtime or CI dependency.

## What was verified

All commands run from the branch worktree with `/Users/oldufo/dev/kornia/.venv/bin/python`
(torch 2.14.0, no torchvision installed); `kornia.__file__` was printed and confirmed to name
`/Users/oldufo/dev/kornia/.claude/worktrees/deps-audit/kornia/__init__.py`.

- Literal generation, in a throwaway venv (`uv venv /tmp/tv && uv pip install --python
  /tmp/tv/bin/python torch torchvision`, CPU wheels): resolved `torch==2.14.0`,
  `torchvision==0.29.0`. Ran a snippet that builds the exact same inputs as
  `_torchvision_reference_inputs()` and calls `torchvision.ops.deform_conv2d` with
  `torch.set_printoptions(precision=10)`; the printed `(1, 2, 4, 4)`, 32-element tensors for
  `use_mask=False` and `use_mask=True` were copied verbatim into `_EXPECTED_NO_MASK` /
  `_EXPECTED_WITH_MASK`.
- `pytest tests/feature/test_aliked.py -k DeformConv2d -q --dtype=float32,float64` →
  `8 passed, 30 deselected`.
- `pytest tests/feature/test_aliked.py -q --dtype=float32,float64` (whole file) →
  `35 passed, 3 skipped` (the 3 skips are pre-existing `need --runslow option to run` skips
  elsewhere in this file, unrelated to this change).
- `pytest tests/test_api_surface.py -q` → `13 passed` (no public API changed; test-only diff).
- `grep -rn torchvision tests/feature/test_aliked.py` → only the new helper/test names and the
  generation-snippet comment; no `import torchvision` remains in this file. (Other test files —
  `tests/geometry/transform/test_affine.py`, `tests/augmentation/test_torch_export.py` — still
  reference torchvision; out of scope for this task, per the brief.)
- `git add -A && pre-commit run --all-files` → all hooks Passed (ruff-format reflowed the two
  literal lists to one value per line; content unchanged).
- Perturbation self-check: temporarily changed `_EXPECTED_NO_MASK`'s first value from
  `1.3524878443` to `1.9524878443` and reran
  `pytest tests/feature/test_aliked.py -k test_matches_torchvision_reference -q
  --dtype=float32,float64`. Result: `2 failed, 2 passed` — both `use_mask=False` cases
  (`cpu-float32-False`, `cpu-float64-False`) failed with `Mismatched elements: 1 / 32 (3.1%)`
  against the corrupted value; the two `use_mask=True` cases were unaffected, as expected. The
  file was then restored to the original literal and the full suite re-verified green
  (`35 passed, 3 skipped`) before committing.

## What is not covered

- No CUDA or MPS run was done (this environment has neither); the float64 assertion path is
  skipped on MPS by design (`pytest.skip("MPS does not support float64")`, matching the sibling
  tests in this class), and non-CPU/float64 combinations fall through to `BaseTester`'s default
  per-dtype tolerance rather than the tighter `1e-9` used for CPU float64.
- float16/bfloat16 pass against the literal on CPU with `BaseTester`'s default tolerances
  (measured max deviation 1.0e-3 and 9.2e-3, at most 0.63 of the allowed bound); no CUDA/MPS
  half-precision run.
- The literal was generated with a single (torch, torchvision) version pair; it is not
  cross-checked against other torchvision releases.

### DIFF
--- a/tests/feature/test_aliked.py
+++ b/tests/feature/test_aliked.py
@@ -23,22 +23,12 @@ from kornia.feature.aliked.deform_conv2d import deform_conv2d
 
 from testing.base import BaseTester
 
-
-def _has_torchvision() -> bool:
-    try:
-        import torchvision.ops  # noqa: F401
-
-        return True
-    except ImportError:
-        return False
-
-
 # ---------------------------------------------------------------------------
 # deform_conv2d tests
 # ---------------------------------------------------------------------------
 
 
-class TestDeformConv2d:
+class TestDeformConv2d(BaseTester):
     """Tests for the pure-PyTorch deform_conv2d implementation."""
 
     @pytest.mark.parametrize("dtype", [torch.float32, torch.float64])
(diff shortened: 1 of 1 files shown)


ducha-aiki:
**Approve after fixes pushed.** The new literal test is deterministic, device/dtype-clean and discriminating; I tightened the float64 bound and made the generation snippet standalone in `71ae2abc`.

### Verified

Head `befba424` reviewed, then `71ae2abc` pushed. All runs from the worktree root with the project venv (torch 2.14.0, no torchvision), `kornia.__file__` confirmed to name the worktree.

```
pytest tests/feature/test_aliked.py -q --dtype=float32,float64   -> 35 passed, 3 skipped
pytest tests/feature/test_aliked.py -q --dtype=float16,bfloat16  -> 35 passed, 3 skipped
uvx ruff@0.16.4 check / format --diff tests/feature/test_aliked.py -> clean; pre-commit run --all-files -> all Passed
```

- Inputs are `linspace`/`arange`/`sin`/`cos`/`sigmoid`, no RNG, so the literal is not torch-version-fragile. Compared tensors go through `.to(device=device, dtype=dtype)`; the only CPU tensors are the deliberate float32 guard.
- Distance of kornia's output from the literal, max abs: float64 `5.0e-11`, float32 `2.3e-7`, float16 `1.0e-3`, bfloat16 `9.2e-3`. Against `BaseTester`'s per-dtype defaults the half runs sit at 0.49-0.62 of the allowed bound (float32 at 0.004), so the half legs pass with headroom.
- The test fails when the implementation is wrong. Four perturbations of `kornia/feature/aliked/deform_conv2d.py` (restored afterwards), `-k test_matches_torchvision_reference --dtype=float32,float64`: offset sign flipped -> 4 failed; mask multiply dropped -> 2 failed (only `use_mask=True`, as it should); `off_y` read from the odd channels -> 4 failed; border validity `<` changed to `<=` -> 4 failed.
- The float32 guard at head really is `rtol=0, atol=1e-5` (the ledger's deferred minor is closed).
- `_has_torchvision` and `test_matches_torchvision` have no remaining references in `tests/`, `testing/`, `conftest.py`, `.github/`; no known-failure manifest names `test_aliked`; no `import torchvision` remains in the file. Test-only diff, no CHANGELOG entry expected and none added.
- Not regenerated locally: torchvision is not importable in the venv and the brief forbids installing it. The controller's independent regeneration (max diff 5e-11) agrees with my measured kornia-vs-literal float64 distance of 5e-11, which is the literal's own 10-decimal rounding.

### Findings

1. `tests/feature/test_aliked.py:205` - the float64 CPU comparison used `rtol=1e-6, atol=1e-6`, about 2e4x looser than the reference (10-decimal literal, measured max diff 5e-11); a float64 arithmetic bug below ~1e-6 would have passed. Fixed in `71ae2abc`: `rtol=0, atol=1e-9` (20x headroom), with a comment stating the measured numbers. The PR body's "float64-on-CPU compares at `rtol=1e-6, atol=1e-6`" is now stale; read it as `rtol=0, atol=1e-9`.
2. `tests/feature/test_aliked.py:92-95` - the generation snippet was not runnable (see Copilot below). Fixed in `71ae2abc` by inlining the five input lines; I diffed the inlined lines against `_torchvision_reference_inputs`' body and they are byte-identical.
3. Question, not changed: float64 on CUDA falls through to `BaseTester`'s `(1e-5, 1e-5)` default rather than the tight bound. There is no reason a CUDA float64 result should differ from the literal by more than ~1e-12, so the `device.type == "cpu"` condition could simply go; I left it because I could not run CUDA. Nit in the same spirit: the float32 guard is a bare `assert torch.allclose(...)`, so a failure prints no mismatch summary, and it recomputes the same CPU float32 result in every dtype/device instance. Harmless, left as is.

### Copilot review

- `tests/feature/test_aliked.py:95`, snippet calls `_torchvision_reference_inputs()` unqualified: VALID. It is a `TestDeformConv2d` staticmethod, and importing the test module into the throwaway venv is not an option either (it pulls `kornia` and `pytest`, which that venv does not have), so I inlined the inputs and said why. Fixed in `71ae2abc`.

### Not covered

- No CUDA or MPS run; the MPS float64 skip and the CUDA fall-through were read, not executed.
- The literal was not regenerated here (no torchvision, not installed); this review relies on the controller's independent regeneration plus the 5e-11 agreement above.
- The half margins (0.49-0.62 of tolerance) were measured on one Apple arm64 CPU; the residual is dominated by half-quantised inputs, so it should be stable across the Intel/AMD Linux legs, but that is inference, not a measurement.

ducha-aiki:
**Final review: Approve.** Full re-review at head `71ae2abc` (one test file, `+146/-33` against `origin/main` `e8111a61`), not a delta. The replacement test is deterministic, device/dtype-clean, tight where it can be and discriminating against four different implementation bugs; the only torchvision reference left is the generation comment, and CI is fully green.

`TestDeformConv2d` becomes a `BaseTester`; the skipif-guarded live comparison against `torchvision.ops.deform_conv2d` (never run in CI, torchvision is not installed anywhere) is replaced by `test_matches_torchvision_reference[use_mask]`, driven by the injected `device`/`dtype` fixtures. Inputs are `linspace`/`arange`/`sin`/`cos`/`sigmoid` (no RNG); the two 32-element float64 literals were generated once with `torch==2.14.0`/`torchvision==0.29.0`, and the comment above them is a standalone snippet that inlines the same five input lines (byte-identical to `_torchvision_reference_inputs`). CPU float64 compares at `rtol=0, atol=1e-9`; everything else uses the `BaseTester` per-dtype defaults; a float32 recomputation is guarded against the float64 literal at `atol=1e-5` in every instance.

### Verified at `71ae2abc`

Worktree root, `.venv` python (torch 2.14.0, no torchvision), `kornia.__file__` naming the checkout.

```
pytest tests/feature/test_aliked.py --dtype=float32,float64    -> 35 passed, 3 skipped (pre-existing --runslow skips)
pytest tests/feature/test_aliked.py --dtype=float16,bfloat16   -> 35 passed, 3 skipped
uvx ruff@0.16.4 check / format --diff                           -> clean; pre-commit run --files -> all hooks pass
```

- Distance of kornia's output from the literal (max abs, and fraction of the applicable bound):

  | dtype | no mask | with mask | fraction of bound |
  |---|---|---|---|
  | float64 | 5.0e-11 | 4.9e-11 | 0.05 of `1e-9` |
  | float32 | 2.3e-7 | 1.3e-7 | 0.004 |
  | float16 | 1.0e-3 | 1.0e-3 | 0.47 / 0.63 |
  | bfloat16 | 9.2e-3 | 4.9e-3 | 0.42 / 0.24 |

  The float64 residual is the literal's own 10-decimal rounding, so `atol=1e-9` keeps 20x headroom while still catching a float64 arithmetic error below the old `1e-6`.
- Discrimination, each perturbation of `kornia/feature/aliked/deform_conv2d.py` applied, run at `--dtype=float32,float64`, reverted (tree clean afterwards): offset sign flipped → 4 failed; `off_y` read from the odd channels → 4 failed; border validity `<` → `<=` → 4 failed; mask multiply dropped → 2 failed, exactly the `use_mask=True` pair.
- Leftovers: `_has_torchvision` and `test_matches_torchvision` have no reference left in `tests/`, `testing/`, `conftest.py`, `.github/`; no known-failure manifest names `test_aliked`; the only other torchvision import in the tree (`tests/geometry/transform/test_affine.py:724`) is out of this PR's scope as the body says. Test-only diff, so no CHANGELOG entry is expected.
- CI on the head: 39/41 green; the two `-/-` rows are the RTD and pre-commit.ci contexts that never report.
- PR body: the two stale statements (`rtol=1e-6, atol=1e-6` for CPU float64; "float16/bfloat16 were not exercised") are corrected in the body now.

### Findings

None blocking. One nit left as is: float64 on CUDA falls through to `BaseTester`'s `(1e-5, 1e-5)` default rather than the `1e-9` bound. Nothing in the kernel is device-dependent (gather plus bilinear weights), so the `device.type == "cpu"` condition could probably go, but I could not run CUDA to prove it and the default is still a valid parity check.

### Not covered

No CUDA or MPS run: the MPS float64 skip and the CUDA fall-through were read, not executed. The literal was not regenerated here (torchvision not installed, and not installed for this review); the 5e-11 agreement above is the evidence that the committed numbers are the torchvision output rounded to 10 decimals. Half margins were measured on one Apple arm64 CPU; the residual is dominated by half-quantised inputs, so it should hold on the Linux legs, and the green float16/bfloat16 CI jobs on this head agree.

Issue #4291, opened 2026-09-05
kornia.contrib.super_resolution builders raise TypeError: SuperResolution never implements the abstract from_config (since #3426)


ducha-aiki:
## What happens

Every builder in `kornia/contrib/super_resolution.py` fails at construction on `main` (e8111a61):

```python
>>> from kornia.contrib.super_resolution import SmallSRBuilder
>>> SmallSRBuilder.build(pretrained=False)
TypeError: Can't instantiate abstract class SuperResolution with abstract method from_config
```

`SuperResolution(ModelBase, ONNXExportMixin)` never implements `from_config`, which `ModelBase` declares `@abstractmethod` (`kornia/models/base.py:134`). Both `SmallSRBuilder.build` (`super_resolution.py:255`) and `RRDBNetBuilder.build` (`super_resolution.py:213`) instantiate `SuperResolution(...)` directly, so the whole public super-resolution entry point has been unreachable since the models refactor in #3426 (2025-12-30) made `from_config` abstract. `SuperResolution`, `SmallSRBuilder` and `RRDBNetBuilder` are all exported from `kornia.contrib` and documented.

Nothing in `tests/` calls either builder, which is why CI never noticed. `tests/models/test_small_sr.py` exercises the underlying `SmallSRNetWrapper` only.

## Fix

Either give `SuperResolution` a `from_config` (the other `ModelBase` subclasses show the shape: a config dataclass + a classmethod that builds the wrapper), or stop subclassing `ModelBase` if the config machinery is not wanted here. Then add one non-slow test per builder that constructs with `pretrained=False` and runs a `(1, 3, 16, 16)` forward, so the entry point is pinned.

Found while vendoring RRDBNet for #4261 (the `RRDBNetBuilder` test in that PR has to monkeypatch around this).

Posted on behalf of @ducha-aiki by Claude (Claude Fable 5.1).

adimalkar:
I'd like to pick this up — one question on direction first, since the fix section leaves it open.

Reproduced on current `main` (`763462e`):

```
>>> from kornia.contrib.super_resolution import SmallSRBuilder
>>> SmallSRBuilder.build(pretrained=False)
TypeError: Can't instantiate abstract class SuperResolution without an
implementation for abstract method 'from_config'
```

**Which direction would you prefer?**

**(a) Give `SuperResolution` a `from_config`.** The eight existing implementations
[...]

PR #4292, opened 2026-09-05
feat(models): vendor RRDBNet and drop the `basicsr` optional import

ducha-aiki:
Part of #4261 (optional imports kornia never declared).

## What changed

- **New** `kornia/models/rrdbnet.py` vendors `ResidualDenseBlock`, `RRDB` and `RRDBNet` from
  [BasicSR](https://github.com/XPixelGroup/BasicSR) (`basicsr/archs/rrdbnet_arch.py` plus the three
  helpers it uses from `arch_util.py`). Constructor arguments and every module/parameter name are
  identical to upstream (`conv_first`, `body.{i}.rdb{1,2,3}.conv{1..5}`, `conv_body`, `conv_up1`,
  `conv_up2`, `conv_hr`, `conv_last`), so the published Real-ESRGAN checkpoints still load with
  `strict=True`.
- `kornia/contrib/super_resolution.py`: `RRDBNetBuilder.build` now constructs the vendored
  `RRDBNet` directly. The four variants keep their exact hyperparameters
  (`RealESRGAN_x4plus` / `RealESRNet_x4plus`: 23 blocks, scale 4; `RealESRGAN_x4plus_anime_6B`:
  6 blocks, scale 4; `RealESRGAN_x2plus`: 23 blocks, scale 2).
- `kornia/core/external.py`: `basicsr = LazyLoader("basicsr")` deleted.
- `RRDBNet` is exported from `kornia.models` and documented in `docs/source/models.rst`.
- The ONNX export survey's skip reason for `contrib.RRDBNetBuilder.build` no longer mentions the
  missing package (the case stays skipped for the unrelated reason that `SuperResolution` is
  abstract, exactly like `contrib.SmallSRBuilder.build`).
- **New** `tests/models/test_rrdbnet.py` (19 tests). The module was previously untested; it is not
  added to the coverage `omit` list.

## Why

`basicsr`'s last PyPI release is 1.4.2 (2022) and it imports
`torchvision.transforms.functional_tensor`, removed in torchvision 0.17 — so this code path was
already broken on any modern stack, and nothing in CI noticed because it had no test. kornia never
declared `basicsr` in `pyproject.toml`, so it was an undeclared optional import rather than a real
extra.

BasicSR is Apache-2.0, "Copyright 2018-2022 BasicSR Authors" — the same licence as kornia. The
vendored file carries the attribution in its module docstring, and `docs/source/models.rst` repeats
it for readers of the rendered API reference.

## What was verified

Interpreter: repo `.venv` (torch 2.14.0), all commands from the worktree root.

- **The vendored code is bitwise-identical to upstream.** The real upstream file was downloaded and
  executed with only its `ARCH_REGISTRY` decorator and its `arch_util` import stubbed out, then A/B'd
  against the vendored class over four configs (scale 1/2/4, and the 64-feature 6-block anime
  configuration). For each: identical `state_dict` keys, **bitwise-identical initial weights** under
  the same seed (so the two consume the RNG in the same order), **bitwise-identical forward output**,
  and `load_state_dict(..., strict=True)` succeeding in **both** directions.
  `ALL IDENTICAL: True`.
- `python -m pytest tests/models/test_rrdbnet.py -q` → `18 passed, 1 skipped in 0.85s`
- `KORNIA_TEST_OPTIMIZER=inductor python -m pytest tests/models/test_rrdbnet.py -q` →
  `19 passed, 1 skipped in 8.97s` (the skip is `test_gradcheck`, deliberately skipped: RRDBNet is a
  deep convolutional generator and a float64 gradcheck of it is prohibitively slow)
- `python -m pytest tests/models/ tests/contrib/ tests/core/test_weights_prefetch.py -q` →
  `444 passed, 109 skipped in 23.07s`
- `python -m pytest --doctest-modules kornia/models/rrdbnet.py kornia/contrib/super_resolution.py`
  → `1 passed`
- `python -m pytest tests/test_api_surface.py -q` → `13 passed` (no inventory change needed;
  `kornia.models` is not a tracked key and additions never fail this test)
- `git add -A && pre-commit run --all-files` → all hooks **Passed**
- `ty check kornia` → `Found 1 diagnostic`, pre-existing and unrelated
  (`kornia/feature/lightglue.py:54`, deprecated `torch.cuda.amp.custom_fwd`)
- `sphinx-build -b html -W --keep-going docs/source …` on a clean output directory →
  `build succeeded.`, **0 warnings**, and the new `RRDBNet` section renders
- `grep -rn basicsr kornia tests docs` → only the required BasicSR attribution in the vendored
  module and in `docs/source/models.rst`, plus the upstream source URL cited in the test file. No
  dependency reference remains.

## What is not covered

- **No pretrained-weight load test.** Verifying `RRDBNetBuilder.build(..., pretrained=True)` end to
  end means downloading a ~64 MB Real-ESRGAN checkpoint, which the offline test suite deliberately
  avoids (the `small_sr` pretrained tests skip on a cold cache for the same reason). The
  checkpoint-compatibility claim is instead pinned two ways that need no network: a hardcoded
  `state_dict` key snapshot in `tests/models/test_rrdbnet.py`, and the bitwise upstream A/B described
  above, which round-trips a `state_dict` between the upstream class and the vendored one with
  `strict=True`.
- **`RRDBNetBuilder.build` still cannot return.** `SuperResolution` subclasses `ModelBase`, whose
  `from_config` is abstract and unimplemented, so instantiating it raises
  `TypeError: Can't instantiate abstract class SuperResolution with abstract method from_config`.
  That is a pre-existing defect, unrelated to `basicsr` — `SmallSRBuilder.build` hits it too, and
  the export survey already records it as a KORNIA BUG. Fixing it is out of scope here, so the
  builder test stubs the wrapper out and asserts on the `RRDBNet` the builder constructs (variant,
  scale, block count, `num_grow_ch`, `eval()` state, and the pre/post-processors `build` passes to
  the wrapper). That bug is tracked as #4291, and the builder test monkeypatches around it; once
  #4291 is fixed the stub can be dropped and the returned `SuperResolution` asserted on directly.
- **Two deliberate deviations from upstream source.** (1) Upstream's hand-written `pixel_unshuffle`
  helper is expressed as `torch.nn.functional.pixel_unshuffle`, which the A/B above shows returns a
  byte-identical tensor. Upstream guards the helper with a bare `assert`; kornia's Ruff config
  enables flake8-bandit and `kornia/` contains zero `assert` statements, while
  `torch.nn.functional.pixel_unshuffle` is already used elsewhere in the tree
  (`kornia/models/kimi_vl/model.py`). (2) `RRDBNet` rejects a `scale` outside `{1, 2, 4}` with a
  `ValueError` at construction; upstream lets `scale=3` fall through to the x4 path and silently
  returns a 4x output. The module docstring states both deviations. All other changes are type
  hints, Google docstrings and `from __future__ import annotations`; the arithmetic is untouched.
- CPU only locally (float32, float64, float16 and bfloat16); no CUDA or MPS run.


ducha-aiki:
Follow-up noted during the final review of the series, for after this merges rather than a blocker: checkpoint compatibility of the vendored `RRDBNet` currently rests on the state-dict key snapshot plus the reviewer's offline A/B against upstream `basicsr` (bitwise-identical init, forward, and `strict=True` interchange in both directions). No released Real-ESRGAN `.pth` has been loaded through it in a test, and the key snapshot pins names, not shapes. A `@pytest.mark.slow` test that downloads `RealESRGAN_x2plus.pth` (~64 MB) into the `weights/` cache, loads it with `strict=True`, and pins one output value on a fixed 8x8 input would close that gap. It needs the `RRDBNetBuilder` wrapper fixed first (kornia#4291), or a direct `RRDBNet` + `load_state_dict` path in the test.

ducha-aiki:
**Approve after fixes pushed.** The vendored `RRDBNet` is bitwise-faithful to BasicSR and `basicsr` is gone from every registry, config and doc; I pushed `5cca044f` for the two Copilot items and for a test suite that did not yet pin the arithmetic it vendors.

### Verified

Head `7ad6666d` checked out; every command from the worktree root with the repo `.venv` (torch 2.14.0), `kornia.__file__` confirmed to name the worktree.

- Upstream A/B, rebuilt independently: `rrdbnet_arch.py` and the three `arch_util.py` helpers fetched from BasicSR `master`, executed with only the `ARCH_REGISTRY` decorator stubbed, then compared with the vendored class over scale 1/2/4 and the 64-feat/6-block config under one seed:
  ```
  pixel_unshuffle scale=2 torch.equal: True   scale=4 torch.equal: True
  {scale 4|2|1, tiny} and {scale 4, 64 feat, 6 blocks}: keys True weights True forward True
  strict=True load_state_dict in both directions: ok        ALL IDENTICAL: True
  ```
- The four builder configs match `inference_realesrgan.py` line for line (23/64/32 x4, 23/64/32 x4, 6/64/32 x4, 23/64/32 x2); `main` never mapped the SRVGG variants, so nothing is mis-routed to RRDBNet.
- `grep -rn basicsr kornia tests docs pyproject.toml pixi.toml .github README.md` -> only the attribution comment in the test file; `autodoc_mock_imports` never listed it. `kornia.models`/`kornia.contrib` are not `api_surface.json` keys (`tests/test_api_surface.py`: 13 passed).
- `pytest tests/models/test_rrdbnet.py --dtype=float32,float64` and `--dtype=float16,bfloat16`: 33 passed, 1 skipped each on the original head; `tests/models/ tests/contrib/ tests/core/test_weights_prefetch.py`: 453 passed, 109 skipped after the fix; doctest 1 passed; ruff check, `uvx ruff@0.16.4 format --diff` and `pre-commit run --all-files` clean.
- `export_support.json`: `merge._reason("skipped", <new string>)` returns it verbatim, so the hand edit is what the merge would write; the 23-block generator has 16.70M parameters, so "17M" holds. The 2 h regeneration itself was not run.
- `SuperResolution` is still abstract, the builder test still monkeypatches around it, and neither the CHANGELOG entry, the PR body nor `models.rst` claims `RRDBNetBuilder.build` returns (the rst flags the `TypeError` and points at #4291).

### Findings

1. `kornia/models/rrdbnet.py:174` (Copilot 1): `RRDBNet(3, 3, scale=3)(torch.rand(1, 3, 8, 8)).shape` was `(1, 3, 32, 32)`, a silent x4 for a documented-invalid scale. Fixed in `5cca044f`: `ValueError` at construction, `Raises:` section, and the module docstring now lists this as the second deliberate deviation from upstream (it was claiming "arithmetic unchanged" only, which stays true). Test `test_exception_rejects_unsupported_scale[0|3|8]`.
2. `tests/models/test_rrdbnet.py:198` (Copilot 2): `KORNIA_TEST_OPTIMIZER=inductor ... --dtype=float16,bfloat16 -k dynamo` failed both half legs on the original head (`Greatest absolute difference: 0.0009765625 ... (up to 0.0001 allowed)`), so the hard-coded 1e-4 was tighter than the dtype allows. Fixed in `5cca044f`: `assert_close` defaults plus `cudnn_tf32_follows_option`, as in `tests/feature/test_sosnet.py`. Note this would not have surfaced in CI because `tests/models` is not in the dynamo leg (#4311).
3. The suite pinned names and the inner 0.2, nothing else: with the RRDB residual changed to 0.3, the `default_init_weights(..., 0.1)` scale set to 1.0, the LeakyReLU slope set to 0.1, or `interpolate` switched to bilinear, all 19 tests stayed green (each perturbation applied to the worktree copy, run, reverted). Fixed in `5cca044f`: an RNG-free numerical pin generated from the upstream code (cosine-ramp weights keyed on the sorted `state_dict`, so no seed-dependent init; literals and the generation snippet in the docstring; passes at all four CPU dtypes with default tolerances), an `RRDB` residual pin, and a recorder asserting the `0.1` init call. Re-running all eight perturbations (including the two the old suite caught and the new validation) now fails at least one test each.

### Copilot review

- `rrdbnet.py:178` scale validation: VALID, reproduced (`scale=3` -> x4 output), fixed in `5cca044f`.
- `test_rrdbnet.py:207` tolerances / `cudnn_tf32_follows_option`: VALID, reproduced (2 half-precision inductor failures), fixed in `5cca044f`.

### Not covered

- CPU only: no CUDA or MPS run, so the `cudnn_tf32_follows_option` leg is untested here.
- No pretrained checkpoint load (64 MB download); the existing follow-up comment on this PR already covers that gap, and the new numerical pin does not replace it (it pins arithmetic, not the released weights).
- The `sphinx -W` build and the export-support regeneration were not rerun.
- Sibling note: the `test/declared-imports-pin` branch's `tests/test_declared_imports.py` docstring lists `basicsr` among the seven undeclared imports the audit found; that is a historical statement and stays true after this merges, but whichever PR lands second must not resurrect the `basicsr = LazyLoader(...)` line in `kornia/core/external.py` during conflict resolution.

ducha-aiki:
**Final review: Approve.** Full re-review at head `5cca044f` (nine files, `+617/-20` against `origin/main` `e8111a61`), not a delta. The vendored `RRDBNet` is bitwise-faithful to BasicSR `master` on every config I built, the numerical pin regenerates exactly from upstream, the suite now fails on every one of eight arithmetic/layout perturbations, `basicsr` is gone from every registry and doc, and CI is fully green.

- New `kornia/models/rrdbnet.py`: `ResidualDenseBlock`, `RRDB`, `RRDBNet` and `_default_init_weights` from BasicSR (Apache-2.0, attribution in the module docstring and in `models.rst`). Module and parameter names are upstream's, so released Real-ESRGAN checkpoints load `strict=True`. Two documented deviations: `F.pixel_unshuffle` for the hand-written helper (byte-identical) and a `ValueError` for `scale` outside `{1, 2, 4}` (upstream silently upsamples by 4).
- `RRDBNetBuilder.build` constructs the vendored class for the four Real-ESRGAN variants with the same hyperparameters; `basicsr = LazyLoader("basicsr")` is deleted; `RRDBNet` is exported from `kornia.models` and documented; the export-survey skip text no longer names the package.
- `tests/models/test_rrdbnet.py`: smoke/cardinality/exception/module/state-dict/strict round-trip tests, an RNG-free numerical pin against upstream (cosine-ramp weights), residual-scaling pins, an init-scale recorder, `test_dynamo` with `cudnn_tf32_follows_option` and default tolerances, and builder selection tests that stub the abstract `SuperResolution` wrapper (kornia#4291).

### Verified at `5cca044f`

```
pytest tests/models/test_rrdbnet.py --dtype=float32,float64                  -> 45 passed, 1 skipped
pytest tests/models/test_rrdbnet.py --dtype=float16,bfloat16                 -> 45 passed, 1 skipped
KORNIA_TEST_OPTIMIZER=inductor ... -k test_dynamo --dtype=float32,float64,float16,bfloat16 -> 4 passed
pytest tests/models/ tests/contrib/ tests/core/test_weights_prefetch.py tests/test_api_surface.py -> 466 passed, 109 skipped
pytest --doctest-modules kornia/models/rrdbnet.py kornia/contrib/super_resolution.py -> 1 passed
uvx ruff@0.16.4 check / format --diff (6 .py files) -> clean; pre-commit run --files (9 files) -> all hooks pass
uvx ty check kornia/models/rrdbnet.py kornia/contrib/super_resolution.py -> All checks passed
```

- Upstream A/B, rebuilt independently: `rrdbnet_arch.py` and the three `arch_util.py` helpers fetched from BasicSR `master` today, executed with only the `ARCH_REGISTRY` decorator stubbed, same seed, over tiny scale 4/2/1 and the 64-feat/6-block anime config: state-dict keys, initial weights and forward output `torch.equal` on all four; `strict=True` `load_state_dict` succeeds in both directions; upstream `pixel_unshuffle` equals `F.pixel_unshuffle` for factors 2 and 4. `ALL IDENTICAL: True`. Upstream `RRDBNet(3, 3, scale=3)` on an 8x8 input returns `(1, 3, 32, 32)`; the vendored class raises `scale must be 1, 2 or 4, got 3`.
- The three `test_numerical_matches_upstream` literals (sum, first pixel, last pixel for scale 4/2/1) regenerated from the upstream class with the test's own `_cosine_ramp_weights` are identical to the committed values, digit for digit.
- Discrimination, each perturbation of `kornia/models/rrdbnet.py` applied, suite run, reverted (tree clean afterwards): RRDB residual 0.2→0.3 → 2 failed; RDB residual 0.2→0.3 → 2 failed; init scale 0.1→1.0 → 1 failed; LeakyReLU slope 0.2→0.1 → 3 failed; nearest→bilinear upsampling → 3 failed; scale-2 unshuffle factor 2→4 → 3 failed; scale validation removed → 3 failed; `conv_hr` renamed → 1 failed. Last round the first five of these passed untouched.
- `RRDBNetBuilder`'s four configs match `inference_realesrgan.py` (23/64/32 x4, 23/64/32 x4, 6/64/32 x4, 23/64/32 x2); the x4plus generator has 16,697,987 parameters, so the survey's "17M" holds.
- `grep -rn basicsr kornia tests docs pyproject.toml pixi.toml .github README.md uv.lock` → only the two attribution/source comments in the test file; `autodoc_mock_imports` never listed it; `kornia.models` / `kornia.contrib` are not `api_surface.json` keys, so the addition needs no inventory change. CHANGELOG bullet under `Unreleased / Added` cites #4292.
- CI on the head: 42/44 green; the two `-/-` rows are the RTD and pre-commit.ci contexts that never report.
- PR body: the "one deliberate deviation" bullet is updated to name the second one (scale validation) and the local dtype coverage line is corrected.

### Findings

None blocking. Open follow-up, already recorded on this PR: a `@pytest.mark.slow` test that loads a released Real-ESRGAN `.pth` with `strict=True` and pins one output value would close the last gap between "keys and arithmetic match upstream" and "the published weights work"; it needs #4291 or a direct `RRDBNet` + `load_state_dict` path.

### Not covered

CPU only; the `cudnn_tf32_follows_option` leg is untested here (no CUDA), no MPS. No pretrained checkpoint download. The `sphinx -W` build and the 2 h export-support regeneration were not rerun (the JSON edit is what `merge._reason("skipped", ...)` would write, checked last round). Sibling note unchanged: whichever of #4292 and #4307 lands second must not resurrect the `basicsr = LazyLoader(...)` line in `kornia/core/external.py` during conflict resolution; #4307's current head already has it removed.

PR #4293, opened 2026-09-05
Load safetensors checkpoints with pure torch and kornia's own downloader

ducha-aiki:
## What changed

`kornia/models/kimi_vl/builder.py` and `kornia/models/siglip2/builder.py` try-imported
`huggingface_hub.hf_hub_download` and `safetensors.safe_open` to fetch and read
`model.safetensors`. Neither package is declared anywhere in kornia's dependencies, so
`KimiVLBuilder.from_pretrained_hf()` and `SigLip2Builder.from_pretrained_hf()` raised `ImportError`
on a plain `pip install kornia`. Both now go through kornia's own machinery:

* **`kornia.core.load_safetensors(path, device="cpu")`** (new module `kornia/core/safetensors.py`) —
  a pure-torch reader for the format: eight bytes of little-endian header length, a JSON header of
  that length, one contiguous byte buffer the header indexes into. It covers
  `F64 F32 F16 BF16 I64 I32 I16 I8 U8 BOOL` and rejects any other dtype by name. The file is
  memory-mapped **privately** (`mmap.ACCESS_COPY`), which is what keeps `torch.frombuffer` from
  warning about a non-writable buffer while never writing a page back to disk; every tensor is then
  copied out (`.to(device, copy=True)`), so the returned state dict owns its storage and the file is
  closed by the time the call returns. Zero-size tensors take a `torch.empty` branch, because
  `frombuffer` rejects a count of 0. The entries must tile the byte buffer exactly — the format
  requires it to be entirely indexed and hole-free, and two entries claiming `[0, 8]` and `[4, 8]`
  both pass a per-entry range check while returning values one of them was never given. Every error
  is a `ValueError` naming the file, so a caller who hits a corrupt cache entry knows what to delete.

* **`kornia.core.download_file_from_url(url, *, file_name, model_dir, progress)`** (new, in
  `kornia/core/download.py`) — the sibling of `load_state_dict_from_url` for a checkpoint torch
  cannot unpickle. It reuses `_cached_file_path`, `_prefetch_with_retry`, `_drop_failed_download`
  and `_SleepBudget`, so it shares the same cache, the same fallback-URL list with the cache name
  pinned to the first URL, the same retry/backoff on transient failures and rate limits, and the
  same `Downloading: "<url>"` line on stderr rather than stdout. It returns the cached path instead
  of a state dict.

* **`hf_url`** now also accepts a full `owner/name` repository id (a repository *name* cannot
  contain a `/`, which is what separates the two forms). SigLIP2 downloads from `google/…`, so a
  kornia-org-only URL builder could not serve it. A bare name still resolves under `kornia/`, and
  every existing caller passes a bare name.

* **`kornia.core.download_hf_file(repo, filename, *, model_dir=None, progress=True)`** (new) — the
  two-decision wrapper a Hub file needs: the `resolve/main` URL, and a cache name that carries the
  repository id. The second is not optional. **Every** HF repository publishes its single-shard
  checkpoint under the name `model.safetensors`, and the download cache is one flat directory, so
  caching by URL basename would have made SigLIP2 load whichever of the two checkpoints — or
  whichever SigLIP2 variant — was fetched first. The name becomes
  `kornia--kimi-vl-a3b-instruct-vision--model.safetensors`; both spellings of a kornia repo (`hardnet`
  and `kornia/hardnet`) resolve to one URL and therefore to one cache entry.

* Both builders' `_download_weights` is now `load_safetensors(download_hf_file(repo,
  "model.safetensors", model_dir=cache_dir))`. Public signatures are unchanged and `cache_dir` keeps
  its meaning ("the directory the checkpoint is cached in").

* `tests/core/test_weights_prefetch.py`: the docstring no longer names `huggingface_hub`, and the
  call-site scan now matches `download_file_from_url` and `download_hf_file` as well as
  `load_state_dict_from_url`, so a future model that downloads through either is caught the same way.
  The two builders are listed in `_DOWNLOAD_CALL_ALLOWLIST` (the mechanism the file already uses for
  `kornia/models/base.py`) rather than in `NOT_PREFETCHED`: SigLIP2 takes the repository as an
  argument, so there is no fixed variant list to enumerate, and `NOT_PREFETCHED` entries must be
  reachable from a registry. That allowlist now has the staleness check `NOT_PREFETCHED` already had
  — every key must match a module the scan really finds — which immediately caught a pre-existing
  dead entry (`kornia/core/__init__.py` re-exports the names and calls none of them), now removed.

* `tests/integration/models/test_kimi_vl.py` reads official shards with `load_safetensors`, so it no
  longer `importorskip`s `safetensors.torch`.

### Behaviour change: where the checkpoints are cached

`cache_dir=None` used to mean the HuggingFace cache
(`~/.cache/huggingface/hub/models--<owner>--<name>/snapshots/<sha>/model.safetensors`) and now means
torch's hub directory (`<torch hub dir>/checkpoints/<owner>--<name>--model.safetensors`) — the same
cache every other kornia checkpoint uses and the one CI restores. An explicit `cache_dir` moves too:
it used to hold the same `models--<owner>--<name>/snapshots/<sha>/…` tree and now holds the flat
`<owner>--<name>--model.safetensors` file. Either way, users who already had `huggingface_hub`
installed and the checkpoint cached re-download it once after upgrading: 854 MB for KimiVL, ~1.5 GB
for SigLIP2. This is stated in the CHANGELOG entry under *Breaking changes*.

`download_file_from_url` and `download_hf_file` also reject a `file_name`/`filename` that is not a
bare filename (a path separator, an absolute path, `""`, `.` or `..`) with `ValueError`, so a caller
cannot write outside the cache directory; the pre-existing `load_state_dict_from_url` keeps torch's
unchecked behaviour.

Part of the optional-import cleanup in #4261.

## Why

kornia never declared either package. The import was optional in form only: there is no code path in
which these two builders work without it, and the "pip install huggingface_hub safetensors" message
asked users to install a hard dependency by hand. The format is nine lines of specification and the
download machinery already existed, so removing both imports costs one small module and buys these
two checkpoints the retrying, rate-limit-aware, quarantining cache that every other kornia checkpoint
already uses — and the CI `weights/` cache along with it.

## What was verified

All from the worktree root, with `.venv/bin/python` (torch 2.14.0); `kornia.__file__` confirmed to be
the worktree copy.

| Command | Result |
| --- | --- |
| `pytest tests/core/test_safetensors.py tests/core/test_download.py tests/core/test_weights_prefetch.py tests/models/test_kimi_vl.py tests/models/test_siglip2.py tests/integration/models/test_kimi_vl.py tests/test_api_surface.py -q` | `259 passed, 1 skipped` (the skip is the integration test: `KIMI_VL_WEIGHTS_DIR` unset) |
| `pytest tests/core tests/models/test_kimi_vl.py tests/models/test_siglip2.py tests/integration/models/test_kimi_vl.py tests/test_api_surface.py -q` | `451 passed, 9 skipped` |
| `pytest --doctest-modules kornia/core/safetensors.py kornia/core/download.py kornia/models/{kimi_vl,siglip2}/builder.py -q` | `3 passed, 4 skipped` (the skips are the `+SKIP` network examples) |
| `git add -A && pre-commit run --all-files` | all hooks `Passed` |
| `uvx ty@0.0.9 check kornia` | 1 diagnostic, pre-existing and unrelated (`torch.cuda.amp.custom_fwd` in `kornia/feature/lightglue.py`) |
| `grep -rn "huggingface_hub\|safetensors" kornia tests docs` | no `huggingface_hub` anywhere; `safetensors` only in the new module, its callers and tests, and prose about the file format. No file imports either package. |

The reader was additionally checked against a real 854 MB `model.safetensors` (443 bfloat16 tensors,
the published Kimi-VL vision checkpoint) that was **already in the local HF cache** — nothing was
downloaded. All 443 tensors match `safetensors.torch.load_file` exactly (same key order, dtype,
shape, `torch.equal`). This was a one-off snippet, not a test: no test imports the `safetensors`
package, since it is present in the dev environment as a transitive dependency and a test using it
would pass locally while proving nothing about the install kornia declares.

New tests: `tests/core/test_safetensors.py` builds safetensors files **by hand** from the
specification (JSON header + raw `tensor.view(torch.uint8)` bytes) and asserts the round trip for
float32, bfloat16, int64 and a zero-size tensor, every supported dtype, the requested device, that
nothing warns, that returned tensors can be mutated and do not view the closed mapping, and that
fourteen malformed files each raise `ValueError` — including overlapping ranges, a gap between
ranges, and trailing bytes no tensor accounts for. `tests/core/test_download.py` gains a class that
downloads through real `file://` URLs (the same `torch.hub.download_url_to_file` path a remote URL
takes, so cache path, bytes on disk and cache hit are all exercised end to end): caching under
`model_dir`, a second call that transfers nothing, `file_name` keeping two same-named checkpoints
apart, fallback to a second source, a partial transfer not blocking the next source, retry on a 429,
the failure message naming cause and path, and stdout staying empty. `download_hf_file` gets its own
class: URL and cache name for a full `owner/name` id and for a bare kornia repo name, and an
end-to-end pair of real transfers proving two repositories publishing `model.safetensors` land in two
cache entries.

## What is not covered

* **No real checkpoint was downloaded.** Kimi-VL is 854 MB and SigLIP2 ~1.5 GB; neither is fetched by
  any PR-matrix job, and the builders' unit tests patch `_download_weights`. The end-to-end path
  (`from_pretrained_hf` → HTTP → cache → reader → `load_state_dict`) is therefore covered as its two
  halves — a real `file://` transfer through the downloader, and the reader against a real
  bit-identical 854 MB checkpoint from the local cache — not as one live run.
* The two checkpoints are **not** added to the CI prefetch list. They were not prefetched before this
  change either, and SigLIP2's repository is a caller-supplied argument with no fixed variant list.
* `pixi run typecheck` / `pixi run build-docs` were not run (pixi is not usable in this worktree);
  `ty` was run directly at the pinned version, and the two new `autofunction` entries in
  `docs/source/core.rst` were not rendered.


ducha-aiki:
**Approve after fixes pushed** — the reader and the downloader hold up under hostile input and against the reference implementation; I pushed one Copilot-flagged validation gap, one reader nit and a CHANGELOG relocation as `1bbd599b`.

### Verified

Head `e0ea2e5b` checked out in a worktree, `kornia.__file__` confirmed to be the worktree copy, `.venv/bin/python` with torch 2.14.0.

```
pytest tests/core/test_safetensors.py tests/core/test_download.py tests/core/test_weights_prefetch.py \
  tests/models/test_kimi_vl.py tests/models/test_siglip2.py tests/test_api_surface.py -q   -> 259 passed (266 after the fix)
pytest tests/core/test_safetensors.py --dtype=float16,bfloat16 -q                            -> 39 passed (the file takes no dtype fixture, so this is the same 39)
pytest --doctest-modules kornia/core/safetensors.py kornia/core/download.py kornia/models/{kimi_vl,siglip2}/builder.py -> 3 passed, 4 skipped
uvx ruff@0.16.4 check / format --diff on the 11 changed .py files                            -> clean; pre-commit on the changed files -> all Passed
grep -rnE "huggingface_hub|hf_hub_download|snapshot_download|safe_open|import safetensors" kornia tests docs pyproject.toml pixi.toml .github -> no hits
git diff origin/main...HEAD -- tests/api_surface.json | grep -c '^-[^-]'                     -> 0 (five additions, no removal)
```

Reader witness, built by hand from the spec in a stdin snippet (not with the `safetensors` package): 16 tensors in one file — every one of the ten dtypes, a `shape: []` scalar, `(0,)` and `(3,0,2)` empties, a tensor saved from a non-contiguous origin, a 3-byte `U8` first so every later offset is unaligned, a `__metadata__` entry and a header padded with 5 spaces. `torch.equal` and dtype match on all 16 with zero warnings; in-place `add_`/`fill_` on two results changed neither the file nor any sibling tensor, and all storages are distinct. The same file read by the `safetensors` package (present transitively in the dev venv; nothing was installed) matches the PR reader exactly, and a file written by `safetensors.torch.save_file` is read exactly by the PR reader. 19 malformed variants (truncated data, header length past EOF, zero header, empty file, overlap, gap, unknown/F8/U16 dtype, shape–offset size mismatch, `2**62 x 2**62` shape, empty tensor mid-buffer, non-UTF-8 header, float offsets, offset `2**70`) each raise a `ValueError` naming the file; the reference implementation agrees on overlap/gap and on accepting the padded header.

Copilot repro: `download_file_from_url(file_url, file_name="../escaped.bin", model_dir=cache)` wrote `cache/../escaped.bin`, an absolute `file_name` wrote to that absolute path, and `download_hf_file("kornia/x", "../../esc2.bin")` produced `cache/kornia--x--../../esc2.bin`. The pre-existing `load_state_dict_from_url(file_name="../escaped.pth")` does the same, and torch's own `load_state_dict_from_url` is `filename = file_name` joined unchecked.

Discrimination: the new `test_no_allowlisted_module_has_stopped_downloading` fails when main's dead `kornia/core/__init__.py` allowlist entry is re-added (1 failed); the call-site regex matches both builders, so the safetensors path is now visible to the prefetch scan. The 7 tests I added fail on the pre-fix sources and pass after.

### Findings

1. `kornia/core/download.py:838` (`download_file_from_url`): `file_name` reached `_cached_file_path` unvalidated, so a separator or `..` wrote outside `model_dir` (repro above). Fixed in `1bbd599b`: `ValueError` unless `file_name` is a bare filename (`os.path.basename(file_name) == file_name` and not `""`/`.`/`..`), documented on both `download_file_from_url` and `download_hf_file` (a filename with a `/` — a file in a repo subdirectory — is now rejected instead of nesting a directory under the cache), with 6 tests. `load_state_dict_from_url` still accepts such a `file_name`; that is pre-existing and torch-compatible, so I left it — question for the author whether to tighten it too.
2. `CHANGELOG.md`: the cache relocation forces a re-download and sat under `### Added`; it also only described `cache_dir=None`, while an explicit `cache_dir` moves too (`hf_hub_download` laid out `cache_dir/models--<owner>--<name>/snapshots/<sha>/model.safetensors`; the PR writes `cache_dir/<owner>--<name>--model.safetensors`). Fixed in `1bbd599b`: moved to `### Breaking changes` with the old and new layouts for both cases; the `Added` bullet points at it. The PR body's "Behaviour change" section should say the same about an explicit `cache_dir`.
3. `kornia/core/safetensors.py:125`: `data_offsets: [False, True]` was accepted as `[0, 1]` (a `bool` is an `int`), while the shape check next to it deliberately rejects `[True]`. Harmless in effect, but inconsistent with its own comment and with the reference reader. Fixed in `1bbd599b` with a parametrize case.
4. Nit, left open: `download_file_from_url` calls `_drop_failed_download(cache_path)` after a failed `_prefetch_with_retry`, but torch's `download_url_to_file` writes to a temp file and only moves it on success, so nothing is ever at `cache_path` there; the branch and `test_a_partial_transfer_does_not_block_the_next_source` (which mocks a write to `dst` before raising) defend against a torch that does not exist. Not wrong, just dead — the author can decide whether to keep it as a belt.
5. Design note, no action: `hf_hub_download` honoured `HF_TOKEN`/`HF_ENDPOINT`/`HF_HUB_OFFLINE` and could serve a gated repo; `download_hf_file` is an anonymous `resolve/main` GET. Both current checkpoints are public, so this only matters if someone passes a gated `model_name` to `SigLip2Builder`, which then fails with an HTTP 401 inside the `RuntimeError` rather than a hint about authentication.

### Copilot review

- `download_file_from_url()` joins an unvalidated `file_name` into the cache dir: **VALID** — reproduced (`cache/../escaped.bin` written); fixed in `1bbd599b`. The pre-existing helper and torch behave the same, which is why it had precedent, but a new public function should not.
- Suppressed item, hard-coded `model.safetensors` in the SigLIP2 `FileNotFoundError` message: **VALID but cosmetic** — the constant `_WEIGHTS_FILE` exists two lines up; left as is.

### Not covered

No real checkpoint was downloaded (854 MB / ~1.5 GB, disk-limited), so `from_pretrained_hf` end to end and the byte equality of the resulting state dict against main's `safe_open` path are not run here; the code paths agree on repo id, filename and `strict=True`, and the reader is bit-exact against the reference on the fixtures above. CUDA/MPS `device=` only checked on CPU. The docs build was not run (`currentmodule:: kornia.core` is present, no `nitpicky`, so the new `autofunction` entries should resolve). The integration test skips without `KIMI_VL_WEIGHTS_DIR`.

ducha-aiki:
**Final review: Approve.** Full re-review at head `1bbd599b` (fourteen files, `+1248/-77` against `origin/main` `e8111a61`), not a delta. The pure-torch reader is bit-exact against the reference implementation in both directions and rejects every malformed file I could build; the downloader shares the existing retry/cache machinery, keys same-named checkpoints apart and now refuses to write outside the cache; both builders are declared-dependency-only; CI is fully green.

- New `kornia/core/safetensors.py`: `load_safetensors(path, device)` reads the 8-byte length, JSON header and byte buffer per the specification; ten dtypes accepted by name, `__metadata__` ignored, every entry validated (dtype, non-bool non-negative int shape, two non-bool int offsets, range inside the buffer, size = numel × itemsize), then the buffer must be tiled exactly with no overlap or gap. Private `mmap` (`ACCESS_COPY`) so `frombuffer` does not warn; every tensor is copied out, so the file is closed on return. Every error is a `ValueError` naming the file.
- `kornia/core/download.py`: `hf_url` accepts `owner/name`; `_hf_cache_file_name` folds the repo id into the cache name (`kornia--kimi-vl-a3b-instruct-vision--model.safetensors`); `download_file_from_url` is `load_state_dict_from_url` minus the unpickle, reusing `_cached_file_path`, `_prefetch_with_retry`, `_SleepBudget` and `_drop_failed_download`, and rejects a `file_name` that is not a bare filename; `download_hf_file` composes the two.
- Both builders' `_download_weights` become `load_safetensors(download_hf_file(...))`; the `huggingface_hub`/`safetensors` try-imports, their error strings and the SigLIP2 `import huggingface_hub` probe are gone; the SigLIP2 `FileNotFoundError` message now names the cache path.
- Inventory: five `kornia.core` additions, zero removals. Docs: a *Weights* section in `core.rst`. CHANGELOG: an *Added* bullet and a *Breaking changes* bullet giving the old and new cache layouts for both `cache_dir=None` and an explicit `cache_dir`.
- Tests: hand-built-from-spec safetensors fixtures (round trips, every dtype, device, no warnings, storage ownership, twenty-odd malformed variants), `file://` end-to-end download tests (cache hit, `file_name` isolation, fallback, 429 retry, failure message, stdout silence, six `file_name` rejections), `download_hf_file` URL/cache-name cases, builder tests pinning the new call chain, and a prefetch scan that now covers the new call sites and fails on a stale allowlist entry.

### Verified at `1bbd599b`

```
pytest tests/core/test_safetensors.py tests/core/test_download.py tests/core/test_weights_prefetch.py
       tests/models/test_kimi_vl.py tests/models/test_siglip2.py tests/test_api_surface.py
       tests/integration/models/test_kimi_vl.py                   -> 266 passed, 1 skipped (KIMI_VL_WEIGHTS_DIR unset)
pytest tests/core/test_safetensors.py --dtype=float16,bfloat16   -> 40 passed
pytest --doctest-modules (safetensors.py, download.py, both builders) -> 3 passed, 4 skipped (+SKIP network examples)
uvx ruff@0.16.4 check / format --diff (11 .py files)             -> clean; pre-commit run --files (14 files) -> all hooks pass
uvx ty check (download.py, safetensors.py, both builders)         -> All checks passed
grep -rnE "huggingface_hub|hf_hub_download|snapshot_download|safe_open|import safetensors|from safetensors" kornia tests docs pyproject.toml pixi.toml .github -> no hit
git diff origin/main...HEAD -- tests/api_surface.json | grep -c '^-[^-]' -> 0
```

- Reader witness, file built by hand from the specification (not with the `safetensors` package): 15 tensors — every accepted dtype, a `shape: []` scalar, `(0,)` and `(3, 0, 2)` empties, a tensor saved from a non-contiguous origin, a 3-byte `U8` first so every later offset is unaligned, a `__metadata__` entry, a header padded with 5 spaces. `torch.equal`, dtype and key order match on all 15 with zero warnings; in-place `add_` on a result changes neither the file nor a re-read; all storages are distinct. Cross-check against the `safetensors` package (present transitively in the dev venv, nothing installed): values identical in both directions, a package-written file is read exactly, and the only difference is that the package returns keys in hash order (it varied across three runs) while kornia's reader keeps header order.
- 19 malformed variants (truncated data, header length past EOF, zero header, empty file, overlap, gap, `F8_E4M3`, `U16`, shape/offset size mismatch, `2**62 x 2**62` shape, non-UTF-8 header, float offsets, `[False, True]` offsets, `[True]` shape, offset `2**70`, header that is a list, entry that is not an object, trailing bytes, negative dim) each raise a `ValueError` whose message names the file. An empty tensor mid-buffer with a padded header loads.
- Copilot's traversal repro on the fixed head: `file_name` of `../escaped.bin`, an absolute path, `""`, `.`, `..` and `sub/x.bin` all raise `ValueError` before any transfer; `download_hf_file("kornia/x", "../../esc2.bin")` likewise; a bare `ok.bin` lands at `<model_dir>/ok.bin` and nothing appears outside the cache.
- CI on the head: 42/44 green; the two `-/-` rows are the RTD and pre-commit.ci contexts that never report.
- PR body: the *Behaviour change* section now also states that an explicit `cache_dir` moves layout, and names the new `file_name` validation.

### Findings

None blocking. Carried over as author's-choice items, none of which I would hold the merge on:

1. `load_state_dict_from_url` still joins an unchecked `file_name` (torch's own behaviour); the new public functions refuse it, the old one keeps compatibility.
2. `download_file_from_url` calls `_drop_failed_download` after a failed attempt, but torch's `download_url_to_file` writes to a temp file and only moves it on success, so nothing is ever at that path; the branch and its test are a belt against a torch that does not exist.
3. `hf_hub_download` honoured `HF_TOKEN`/`HF_ENDPOINT`; the new path is an anonymous `resolve/main` GET, so a gated repo passed to `SigLip2Builder` fails with an HTTP 401 inside the `RuntimeError` rather than an auth hint. Both shipped checkpoints are public.
4. Cosmetic: SigLIP2's `FileNotFoundError` text hard-codes `model.safetensors` two lines below the `_WEIGHTS_FILE` constant.

### Not covered

No real checkpoint download (854 MB / ~1.5 GB, disk-limited), so `from_pretrained_hf` end to end is covered as its two halves (real `file://` transfers through the downloader; the reader bit-exact against the reference on the fixtures above and, per the body, against the cached 854 MB Kimi-VL file). `device=` only on CPU. No docs build (`currentmodule:: kornia.core` is present and no `nitpicky`, so the five `autofunction` entries should resolve). The integration test skips without `KIMI_VL_WEIGHTS_DIR`.

Issue #4283, opened 2026-09-05
W1b: `get_optical_pose_base` / `homography_i_H_ref` — documented public functions that always raise `NotImplementedError`


ducha-aiki:
`get_optical_pose_base` and `homography_i_H_ref` in `kornia/geometry/camera/pinhole.py` have full docstrings, Args/Returns blocks and a `.. math::` block describing the 12-vector pinhole model — and both bodies validate their input and then `raise NotImplementedError`. This is the same shape as #4017 in the bbox surface: advertised, documented, always raising.

`get_optical_pose_base` carries `# TODO: where is rtvec_to_pose?` — its dependency was removed from torchgeometry years ago. `homography_i_H_ref` is dead because it calls `get_optical_pose_base`.

Neither name is exported: they are in no `__all__`, are not reachable as `kornia.geometry.camera.X`, and are on no docs page (see #4275, #4268). Their in-source documentation is the only thing that advertises them, and it advertises functions that cannot be called.

## Repro

```python
import torch, inspect
from kornia.geometry.camera.pinhole import get_optical_pose_base, homography_i_H_ref

# the documented (N, 12) pinhole vector
v = torch.tensor([[100.0, 100.0, 4.0, 3.0, 6.0, 8.0, 0., 0., 0., 0., 0., 0.]])
for name, fn, args in (("get_optical_pose_base", get_optical_pose_base, (v,)),
                       ("homography_i_H_ref", homography_i_H_ref, (v, v))):
    doc = (inspect.getdoc(fn) or "").splitlines()
    print(f"{name:<22} docstring: {doc[0]}")
    try:
        fn(*args)
    except Exception as e:
        print(f"{'':<22} call     : {type(e).__name__}: {str(e)!r}")
```

## Observed

```
get_optical_pose_base  docstring: Compute extrinsic transformation matrices for pinholes.
                       call     : NotImplementedError: ''
homography_i_H_ref     docstring: Homography from reference to ith pinhole.
                       call     : NotImplementedError: ''
```

## Expected

Two coherent outcomes:

[...]

Issue #4282, opened 2026-09-05
W1b: `unproject_points_z1` — the documented `(..., 1)` extension raises for `N > 1`


ducha-aiki:
`unproject_points_z1(points, extension)` documents `extension` with shape `(..., 1)`. Its guard is written `elif extension.shape[0] > 1: extension = extension[..., None]`, which unsqueezes on the *batch* size rather than checking the rank. The result is exactly backwards:

- the documented `(N, 1)` shape works for `N = 1` and **raises** for `N > 1`;
- an undocumented `(N,)` shape works for `N > 1` and raises for `N = 1`.

`unproject_points_orthographic`, the sibling with the same signature, compares *ranks* — the right predicate — and accepts both shapes for both sizes.

## Repro

```python
import torch, inspect
from kornia.geometry.camera import unproject_points_orthographic, unproject_points_z1

print("docstring Args:", [l.strip() for l in inspect.getdoc(unproject_points_z1).splitlines() if l.strip().startswith("extension")][0])
p1, p2 = torch.tensor([[3.0, 6.0]]), torch.tensor([[3.0, 6.0], [15.0, 20.0]])
print("N=1, extension (1, 1)  ->", unproject_points_z1(p1, torch.tensor([[3.0]])).tolist())
for name, pts, ext in (("N=2, extension (2, 1) [documented]", p2, torch.tensor([[3.0], [5.0]])),
                       ("N=1, extension (1,)   [undocumented]", p1, torch.tensor([3.0])),
                       ("N=2, extension (2,)   [undocumented]", p2, torch.tensor([3.0, 5.0]))):
    try:
        print(f"{name} ->", unproject_points_z1(pts, ext).tolist())
    except Exception as e:
        print(f"{name} ->", type(e).__name__ + ":", str(e)[:80])
print("unproject_points_orthographic, (2, 1):", unproject_points_orthographic(p2, torch.tensor([[5.0], [6.0]])).tolist())
print("unproject_points_orthographic, (2,)  :", unproject_points_orthographic(p2, torch.tensor([5.0, 6.0])).tolist())
```

## Observed

```
docstring Args: extension: torch.Tensor representing the extension (depth) of the points to unproject with shape (..., 1).
N=1, extension (1, 1)  -> [[9.0, 18.0, 3.0]]
N=2, extension (2, 1) [documented] -> RuntimeError: Sizes of tensors must match except in dimension 2. Expected size 2 but got size 
[...]

Issue #4281, opened 2026-09-05
W1b: `PinholeCamera` / `warp_frame_depth` / `StereoCamera` — an empty batch (`B = 0`) is rejected


ducha-aiki:
kornia's degenerate-shape convention (the #4115 family) is **empty in → empty out**, with validation parity: a `B = 0` input is rejected only where the math is undefined. The free functions on this surface already follow it — `project_points`, `depth_to_3d` and `depth_to_3d_v2` all return an empty tensor of the right shape for `B = 0`. Three entry points beside them do not, each for a different accidental reason:

1. **`PinholeCamera(B=0)`** → `ValueError: Arguments shapes must match`. The predicate is `all(data.shape[0] for ...)`, which tests truthiness rather than equality (see #4266), so a batch size of `0` is falsy and a perfectly consistent empty batch is rejected with a message about mismatched shapes.
2. **`warp_frame_depth(B=0)`** → `ZeroDivisionError: integer division or modulo by zero` — an unguarded arithmetic crash, not a validation error.
3. **`StereoCamera(B=0)`** → `StereoException: Expected T_x * f_x to be negative. Got tensor([])`. `torch.all` of an empty tensor is `True`, so the guard's *complement* fires on an empty batch.

## Repro

```python
import torch
from kornia.geometry.camera import PinholeCamera, StereoCamera, project_points
from kornia.geometry.depth import depth_to_3d, depth_to_3d_v2, warp_frame_depth

K3 = torch.tensor([[100.0, 0.0, 4.0], [0.0, 100.0, 3.0], [0.0, 0.0, 1.0]])[None]
# the same surface already returns empty for empty on the free functions
print("project_points(B=0)        :", tuple(project_points(torch.zeros(0, 1, 3), K3).shape))
print("depth_to_3d(B=0)           :", tuple(depth_to_3d(torch.zeros(0, 1, 2, 3), K3.repeat(0, 1, 1)).shape))
print("depth_to_3d_v2(B=0)        :", tuple(depth_to_3d_v2(torch.zeros(0, 4, 5), K3.repeat(0, 1, 1)).shape))

try:
    PinholeCamera(torch.zeros(0, 4, 4), torch.zeros(0, 4, 4), torch.zeros(0), torch.zeros(0))
except Exception as e:
    print("PinholeCamera(B=0)         :", type(e).__name__ + ":", str(e)[:70])
try:
[...]

Issue #4279, opened 2026-09-05
W1b: `PinholeCamera.from_parameters` — `height` and `width` are 0 for every batch element after the first


ducha-aiki:
`PinholeCamera.from_parameters(..., batch_size=B)` fills `height` and `width` only for element 0. The builder does `height_tmp[..., 0] += height` on a `(B,)` zero tensor, so every element after the first keeps `0`. Every other parameter (`fx`, `fy`, `cx`, `cy`, `tx`, `ty`, `tz`) is broadcast correctly, so the camera looks healthy until something reads its image size.

## Repro

```python
import torch
from kornia.geometry.camera import PinholeCamera

cam = PinholeCamera.from_parameters(
    fx=torch.tensor([100.0, 200.0]), fy=torch.tensor([100.0, 50.0]),
    cx=torch.tensor([4.0, 6.0]), cy=torch.tensor([3.0, 2.0]),
    height=6, width=8,
    tx=torch.tensor([1.0, 2.0]), ty=torch.tensor([0.0, 0.0]), tz=torch.tensor([0.0, 0.0]),
    batch_size=2, device=torch.device("cpu"), dtype=torch.float32)
print("height:", cam.height.tolist())
print("width :", cam.width.tolist())
print("fx    :", cam.fx.tolist(), " (the other parameters are filled correctly)")
print("tx    :", cam.tx.tolist())
```

## Observed

```
height: [6.0, 0.0]
width : [8.0, 0.0]
fx    : [100.0, 200.0]  (the other parameters are filled correctly)
tx    : [1.0, 2.0]
```

## Expected

`height` and `width` are filled for the whole batch — `height_tmp += height` (or `torch.full((batch_size,), height, ...)`), matching every other parameter. For `batch_size=1`, the only case that works today, the result is byte-identical, so this is a focused fix and is welcome as a PR. A regression test that builds a `batch_size=2` camera and asserts `height.tolist() == [6.0, 6.0]` fails before the fix.

## Classification

**Broken output** — a silently wrong attribute for every batch element after the first. Idiom 3 (the `batch_size` argument's documented contract is not what the code implements).

## Disposition

n/a — single symbol.

## Environment

kornia 0.9.0rc1 @ `92e5fe8d` (main), torch 2.14.0, python 3.11, cpu. Found during the camera conventions audit (batch 5, tracking issue #4262); the batch documents the behavior below as it is and pins it, and does not change it.

Posted on behalf of @ducha-aiki by Claude (Opus 5).

PR #4294, opened 2026-09-05
docs(geometry): Convention blocks for camera (batch 5a: projection core)

ducha-aiki:
## Summary

Documents and pins the camera projection core from the batch 5 audit: what `PinholeCamera`'s frames are, where a
pixel centre sits, the two things "depth" means on this surface, and what each projection entry point does at
`z = 0`. Fifteen `Convention:` blocks, eleven pointer lines back to the `PinholeCamera` anchor, twelve
`.. warning::` blocks linking the eight open issues below, plus 27 additive test pins.

empty for all four modified `kornia/` files. No existing test or doctest was modified; every pin is new. The only `docs/`
edit is the `geometry.camera.perspective.rst` meta description, which said "world coordinates" where the rewritten
`unproject_points` docstring on that page now says camera-frame (5a adds no rst directives; the later sub-batches do).

Part of #4262. Changelog: entry added under `## Unreleased` -> `### Added` in this PR.

## Scope

| symbol | what it got |
| --- | --- |
| `PinholeCamera` (anchor) | canonical `Convention:` block: `K` layout, world→camera `extrinsics`, integer pixel centres and `cx = (W-1)/2`, depth = camera-frame `z` vs ray length; class warning for #4263/#4264/#4265/#4266/#4281 |
| `PinholeCamera.scale` / `.scale_` | `Convention:` (new camera vs in place, `cx' = s·cx`, shared `extrinsics`, int size → float only for a float factor, `scale_` left half-scaled when it raises) + warnings #4263/#4264/#4265 |
| `PinholeCamera.project` | pointer + `Convention:` (rank ≥ 2 or `IndexError`, empty points, undivided `K (R X + t)` at camera-frame `z = 0`) + #4266/#4267 |
| `PinholeCamera.clone` / `.unproject` / `.intrinsics_inverse` | pointer |
| `PinholeCamera.from_parameters` | pointer + `Convention:` (every parameter, image size included, broadcast over the batch — #4279 was fixed on `main` by #4297 while this PR was open; float sizes) |
| `pixel2cam` / `cam2pixel` | pointer + `Convention:` (4x4 inverse/projection, the guard predicates as written — `pixel2cam`'s depth guard as repaired by #4314 — `x / (z + eps)` per dtype incl. the zero-numerator and float16 cases) + #4266/#4267 |
| `pinhole_matrix` / `inverse_pinhole_matrix` | pointer + `Convention:` (12-vector layout, `AssertionError`) + `.. note::` + #4268 |
| `get_optical_pose_base` / `homography_i_H_ref` | warning #4283 |
| `project_points` / `unproject_points` | pointer + `Convention:` (camera frame, rank ≥ 2 or `ValueError`, `z = 0` and behind-camera, `normalize` = ray length) + #4266/#4267 |
| `project_points_z1` / `unproject_points_z1` / `dx_project_points_z1` | `Convention:` (canonical `z = 1` is normalized, not pixels, and differs from `project_points` below its `1e-8` threshold; `±inf`/`nan` per component at `z = 0`; `extension` = camera `z`, both shapes accepted since #4355; the `(..., 2, 3)` Jacobian) + #4267 |
| `unproject_points_orthographic` | `Convention:` (both extension shapes accepted) |

## Pins

`tests/geometry/camera/test_pinhole.py` — 6 convention, 12 wart:

- `test_convention_extrinsics_are_world_to_camera`, `test_convention_unproject_returns_the_world_point`,
  `test_convention_clone_is_a_deep_copy`, `test_convention_intrinsics_inverse_is_the_exact_inverse`
- `test_convention_from_parameters_fills_every_batch_element_4279` — regression pin for the #4297 fix (it was a
  strict `xfail` until #4297 merged)
- `test_convention_pixel2cam_rejects_multi_channel_depth_4266` — regression pin for the #4314 depth guard
- `test_wart_scale_rescales_the_principal_point_by_the_half_pixel_rule_4263` (#4263)
- `test_wart_scale_shares_extrinsics_with_the_source_4264` (#4264 — also the constructor aliasing, the `tx` setter and `scale_` writing through the caller's tensors)
- `test_wart_scale_inplace_rejects_integer_image_size_4265` (#4265 — int factor succeeds, float factor raises after the intrinsics are already scaled; asserts the post-state)
- `test_wart_constructor_accepts_3x3_intrinsics_whose_projection_is_garbage_4266`,
  `test_wart_project_rejects_an_unbatched_point_with_indexerror_4266`,
  `test_wart_pixel2cam_guard_admits_a_3x3_inverse_4266`,
  `test_wart_cam2pixel_guard_admits_a_3x3_projection_4266` (#4266)
- `test_wart_project_and_project_points_disagree_at_z_zero_4267`,
  `test_wart_cam2pixel_epsilon_makes_the_singular_divide_finite_4267` (#4267 — the float16 overflow is the
  float16 branch of the same node, so the blocking `cpu-float16` job runs it in its own cell)
- `test_wart_pinhole_matrix_perturbs_every_entry_4268` (#4268)
- `test_wart_constructor_rejects_an_empty_batch_4281` (#4281 — mixed: the constructor `ValueError` is the wart, `project_points(B = 0) -> (0, 1, 2)` is the convention that survives the repair)
- `test_wart_dead_legacy_functions_always_raise_4283` (#4283)

`tests/geometry/camera/test_perspective.py` — 2 convention, 1 wart:
`test_convention_integer_pixel_centres_put_the_principal_point_at_w_minus_one_half`,
`test_convention_normalize_makes_depth_the_ray_length`,
`test_wart_project_points_skips_the_divide_at_z_zero_4267` (#4267).

`tests/geometry/camera/test_projections.py` — 5 convention, 1 wart:
`test_convention_dx_project_points_z1_matches_autograd`,
`test_convention_project_points_z1_differs_below_perspective_epsilon`,
`test_convention_dx_orthographic_is_the_scalar_du_dx_not_the_full_jacobian`,
`test_convention_orthographic_drops_and_restores_the_z_axis`,
`test_convention_unproject_points_orthographic_accepts_both_extension_shapes`,
`test_wart_project_points_z1_zero_depth_is_component_dependent_4267` (#4267).
The `unproject_points_z1` shape contract is already pinned on `main` by #4355's parametrized
`test_unproject_depth_shapes_4282`, which this PR leaves untouched.

Every wart pin asserts the **current** behavior, names its issue in the test name, and carries a comment saying it
is not a contract and is deleted when the issue is repaired.

## Issues documented here

| --- | --- | --- |
| #4263 `scale` rescales `cx` by `s·cx` while every grid uses integer pixel centres | semantic contract (idiom 3) | MERGE IN WINDOW — behavior change (all four scale sites at once) |
| #4264 `scale` shares `extrinsics`; `scale_` and the setters write the caller's tensors | broken output, input side effect (idiom 6) | n/a — single symbol |
| #4265 `scale_` raises `RuntimeError` on the int64 size the constructor accepts | broken output, crash on documented input (idiom 2) | n/a — single symbol |
| #4266 shape guards that do not enforce what their messages say (six remain; #4314 repaired the `pixel2cam` depth guard) | semantic contract (idioms 3, 5) | n/a — single symbol |
| #4267 five projection entry points, four answers at `z = 0` | semantic contract, recorded conflict (idioms 1, 4) | KEEP SEPARATE — intentional variant; the one-policy sweep is a window item |
| #4268 the 12-vector API returns `eye(4) + eps` and is exported nowhere | semantic contract (idiom 1) + API completeness | MERGE IN WINDOW — behavior change (canonical: the class pair) |
| #4281 an empty batch (`B = 0`) is rejected by an accidental predicate | semantic contract, degenerate shapes (idioms 3, 4) | n/a — single symbol; linked to #4115 |
| #4283 `get_optical_pose_base` / `homography_i_H_ref` always raise `NotImplementedError` | API completeness | n/a — single symbol |

#4279 and #4282 were documented and pinned as warts when this PR opened; #4297 and #4355 fixed them on `main` in the
meantime, so their warnings and wart pins are gone and the surviving pins record the repaired behavior.

## Docstring errors fixed

- `unproject_points` — the `point_2d` Args line and the Returns line called the output **world** coordinates. The
  function takes no extrinsics and returns camera-frame coordinates; both now say so, and the block names
  `PinholeCamera.unproject` as the world-frame counterpart.
- `PinholeCamera` — the `intrinsics` / `extrinsics` Args lines promised `(B, 4, 4)` without saying that the
  constructor's predicate also admits a `(B, 3, 3)` (and a rank-2 `(4, 4)`); they now state the predicate applied.
- `pinhole_matrix` / `inverse_pinhole_matrix` / `scale_pinhole` — "This method is going to be deprecated in version
  0.2" (0.2 is long past and nothing was deprecated) replaced by a `.. note::` recording the current state and #4268.
- `PinholeCamera.intrinsics_inverse` — typo `instrisics`.

## Docs-only proof

For each of the four modified `kornia/geometry/camera/*.py` files, parse both revisions, pop the leading string
`Expr` from every `Module` / `ClassDef` / `FunctionDef` / `AsyncFunctionDef`, `ast.dump`, and diff:

```
diff <(git show <base>:$f | strip_docstrings) <(strip_docstrings < $f)

EMPTY  kornia/geometry/camera/pinhole.py
EMPTY  kornia/geometry/camera/perspective.py
EMPTY  kornia/geometry/camera/projection_z1.py
EMPTY  kornia/geometry/camera/projection_orthographic.py
```

## Validation

At `66d0f724` (merge of `main` `f4532f39`; torch 2.14.0; `kornia.__file__` printed from the branch worktree):

- `pixi run lint` / pre-commit.ci — passed; the pinned `ruff@0.16.6` check and format are clean
- docs: the `docs (3.11)`, `Check documentation links` and Read the Docs checks are green on the branch; the
  `camera|pinhole|perspective|projection` filter over the build log's warning/error lines is empty
- `pytest tests/geometry/camera --device=cpu --dtype=float32,float64` — 289 passed, 1 skipped
- `--dtype=float16 --xfail-known-failures --known-failure-profile=cpu-float16` — 133 passed, 4 skipped, 13 xfailed;
  `bfloat16` likewise — 124 passed, 2 skipped, 24 xfailed. No `test_wart_*` / `test_convention_*` node is xfailed or
  skipped by a manifest, and no manifest names a camera node
- `--device=mps --dtype=float32` (with `PYTORCH_ENABLE_MPS_FALLBACK` unset) — 140 passed, 10 skipped, 0 failed
- every pin was mutation-checked: each claim was flipped, the node was observed to fail, and the flip reverted; the
  round-1 review re-executed every quantified sentence one parameter away from the fixture (int vs float scale
  factor, `scale_` post-state for int height vs int width, the three sign classes at `z = 0`, float16 overflow)

`PinholeCamera.scale` / `scale_` / `scale_pinhole` / `PinholeModel.scale` (MERGE IN WINDOW, #4263/#4264/#4265) ·
`pinhole_matrix` / `inverse_pinhole_matrix` (MERGE IN WINDOW, #4268) ·
`project_points` / `PinholeCamera.project` / `project_points_z1` / `cam2pixel` / `Z1Projection.project`
(KEEP SEPARATE; the singular-input conflict is #4267) ·
`PinholeCamera.unproject` / `unproject_points` / `pixel2cam` / `depth_to_3d_v2` (KEEP SEPARATE — different frames;
the "world coordinates" Returns line is the docstring fix above).

## Notes for reviewers

- **Pointer targets.** `geometry.camera.pinhole.rst` sets `currentmodule:: kornia.geometry.camera.pinhole`, so
  `:class:`~kornia.geometry.camera.PinholeCamera`` renders as dead text with no build warning. Every pointer uses
  the resolvable `~kornia.geometry.camera.pinhole.PinholeCamera`; the rendered line is unchanged and now links.
  Each added xref was checked against `docs/build/html/objects.inv` and as a real `<a href>` in the built page.
- `get_optical_pose_base` and `homography_i_H_ref` sit outside the nominal 5a symbol list but are covered here,
  because #4283 is their issue and they live in the same module.
- `:class:`~kornia.core.exceptions.ShapeError`` resolves nowhere (`kornia.core.exceptions` is on no rst page). It is
  pre-existing — the merged bbox batch uses the same reference in `boxes.py` and `bbox.py` — so it is a repo-wide
  follow-up rather than a 5a change.
- `test_convention_intrinsics_inverse_is_the_exact_inverse` is the one node to watch on the Linux 2.5.1 / 2.9.1
  legs: exactness was verified on torch 2.14 / macOS only. It skips float16 with a visible reason.
- `dx_project_points_orthographic` returns `(..., 1)` — the single partial `du/dx` its own math states — while
  `dx_project_points_z1` returns the full `(..., 2, 3)` Jacobian. Both are pinned; the asymmetry between two
  identically prefixed names is deliberate current behavior, not an issue.
- The audit's candidate idiom "axis-order slip hidden by symmetric test data" is not ratified yet and appears in no
  docstring here; it belongs to #4269, which is out of 5a's scope.


ducha-aiki:
Pushed `04573e7f`: the anchor's functional-API shape sentence now names `unproject_meshgrid`'s `(B, 3, 3)`; found by the 5c audit (#4271).

Re-executing the whole list on this branch (torch 2.14.0, cpu) showed the `(*, 3, 3)` claim held for two of the six functions, not five:

| function | `(3, 3)` | `(2, 3, 3)` | `(2, 1, 3, 3)` | `(2, 2, 3, 3)` |
| `project_points` | ok | ok | ok | ok |
| `unproject_points` | ok | ok | ok | ok |
| `depth_to_3d` | `ShapeError` | ok | `ShapeError` | `ShapeError` |
| `depth_to_3d_v2` | `ShapeError` | ok | ok | `RuntimeError` |
| `unproject_meshgrid` | `ShapeError` | ok | ok | `RuntimeError` |
| `depth_to_normals` | `ShapeError` | ok | `ShapeError` | `ShapeError` |

So the sentence now splits: `project_points` / `unproject_points` take a `(*, 3, 3)` `K`, and the four depth functions need it batched as `(B, 3, 3)`. Docstring-only: the docstring-stripped AST of `kornia/geometry/camera/pinhole.py` is identical to `e8111a61`'s. `pixi run lint`, `pixi run doctest` (650 passed, 11 skipped) and a clean `pixi run build-docs` are green, with no `pinhole` warning in the build log.

ducha-aiki (commented):
Reviewed in a separate detached worktree at 04573e7f. I recommend correcting the two P2 documentation findings before merging; the P3 scaling qualification is a smaller correction. These are inaccuracies introduced in the new convention blocks, not requests to repair the underlying tracked runtime behavior.

Validation: the full CPU camera suite at float32/float64 passed (246 passed, 2 skipped, 2 xfailed); the new convention/wart pins at CPU float16/bfloat16 passed (50 passed, 6 skipped, 2 xfailed); those pins at MPS float32/float16 passed (52 passed, 4 skipped, 2 xfailed); all 13 doctests in the four changed library modules passed. Stripping docstrings and comparing ASTs confirms no runtime changes in all four modules. Imports were verified to resolve to the review worktree.

CI currently reports an XFeat numerical comparison failure in tests/feature/test_xfeat.py::TestXFeatModel::test_dynamo[cpu-float32-inductor], outside the changed files. I have not reproduced it against the base and am not attributing its cause. The other test matrix jobs, pre-commit and hosted docs build passed. I reused existing dependencies and did not run a full local docs build or install packages given the available disk space.

Submitted as a comment review because the authenticated account is also the PR author.

kornia/geometry/camera/projection_z1.py
```diff
@@ -41,6 +41,18 @@ def project_points_z1(points_in_camera: torch.Tensor) -> torch.Tensor:
         If this is not the case, the points will be projected to the canonical plane, but the resulting
         points will be behind the camera and causing numerical issues for z == 0.
 
+    Convention:
+        - the input is a **camera-frame** point and the output its position on the canonical ``z = 1`` plane,
+          which is a normalized coordinate rather than a pixel: applying a ``K`` to it with
+          :func:`~kornia.geometry.conversions.denormalize_points_with_intrinsics` gives the pixel that
+          :func:`~kornia.geometry.camera.perspective.project_points` returns.
+        - the ``z > 0`` precondition above is not validated: a point with ``z = 0`` returns ``inf`` rather
+          than raising.
```
ducha-aiki: [P2] Qualify the projection equivalence and singular result

The new block says applying K gives the result of project_points and that z=0 returns inf, but the singular cases being documented do not satisfy either statement. At identity K, project_points_z1([[0., 0., 0.]]) returns [[nan, nan]], and [[0., 2., 0.]] returns [[nan, inf]]; project_points returns finite coordinates for both. The difference also includes small positive depths: for float64 [[1., 2., 1e-9]], Z1 returns approximately [[1e9, 2e9]], while project_points returns [[1., 2.]] because its homogeneous conversion skips the divide when abs(z) <= 1e-8. Please qualify the ordinary-projection equivalence and describe zero-depth outputs as component-dependent Inf/NaN. A zero numerator and a near-zero positive depth case would cover what the new all-nonzero singular pin misses.

kornia/geometry/camera/projection_z1.py
```diff
@@ -66,6 +78,18 @@ def unproject_points_z1(
...
         \begin{bmatrix} u \\ v \end{bmatrix} \cdot w
 
+    Convention:
+        - ``extension`` is the camera-frame ``z`` of the unprojected point: the canonical point is multiplied
+          by it and it becomes the third component.
+          :meth:`~kornia.sensors.camera.projection_model.Z1Projection.unproject` names the same argument
+          ``depth``.
+        - the guard reads the **batch size** of ``extension`` rather than its rank, so the documented
+          ``(..., 1)`` shape raises for more than one point while an undocumented ``(N,)`` shape works there.
+          :func:`~kornia.geometry.camera.unproject_points_orthographic` compares ranks and accepts both.
```
ducha-aiki: [P2] Describe the extension failure by its first dimension

The documented (..., 1) extension does not fail simply when there is more than one point. For points shaped (1, 2, 2), an extension shaped (1, 2, 1) succeeds and returns (1, 2, 3), despite containing two points. Conversely, points shaped (2, 1, 2) with extension (2, 1, 1) fail. The actual trigger is extension.shape[0] > 1, so the current wording only describes the special (N, 2) case tested below and misleads users with batched point sets. Please state that restriction explicitly and cover a multidimensional example while retaining the underlying wart.

kornia/geometry/camera/pinhole.py
```diff
@@ -272,6 +316,19 @@ def intrinsics_inverse(self) -> torch.Tensor:
     def scale(self, scale_factor: torch.Tensor) -> "PinholeCamera":
         r"""Scale the pinhole model.
 
+        Convention:
+            - returns a **new** camera whose focal lengths, principal point and image size are multiplied by
+              ``scale_factor``: ``fx' = s * fx`` and ``cx' = s * cx``, the half-pixel rule.
+            - the new camera **shares** its ``extrinsics`` tensor with the source, so writing ``tx`` / ``ty`` /
+              ``tz`` on either camera moves both; the intrinsics are cloned. :meth:`clone` is the deep copy.
+            - an integer ``height`` / ``width`` is promoted to floating point, unlike :meth:`scale_`.
```
ducha-aiki: [P3] Qualify integer-size behavior by the factor dtype

Promotion is not unconditional: with int64 height/width, scale(torch.tensor([2])) preserves int64, and scale_(torch.tensor([2])) succeeds (height 6 becomes 12). The promotion/in-place casting-error distinction holds for a floating scale factor such as the tested 0.5. Please qualify this line and the corresponding scale_ convention at lines 359-360 so that integer image sizes alone are not described as sufficient to trigger promotion or failure.

ducha-aiki:
## Review of #4294 (batch 5a: projection core) at `04573e7f`

**Verdict: changes requested.** The runtime claim holds (docstring-stripped AST of the four `kornia/` files is identical to `e8111a61`), every new pin passes on cpu float16/bfloat16/float32/float64 and on mps float32 with `PYTORCH_ENABLE_MPS_FALLBACK` unset, and no manifest needs an entry. What needs work is prose: four sentences in the new Convention blocks are false one parameter away from the fixture, and one pin escapes the float16 leg. Everything below was executed on the branch worktree (torch 2.14.0, `kornia.__file__` printed from the worktree).

### CI

- The red `dynamo / tests (3.11, 2.14.0)` job is `tests/feature/test_xfeat.py::TestXFeatModel::test_dynamo[cpu-float32-inductor]`: 4 of 4096 elements off by 2.3e-4 against a 1e-4 atol. This PR touches nothing in `kornia/feature`, and the same job is green on the 24 other runs of that workflow today, so it is a flake of that test; a re-run should clear it.
- GitHub reports the PR as `CONFLICTING` after today's ten commits on `main`, although `git merge-tree origin/main docs/convention-blocks-camera-a` is clean locally. Rebase anyway: a dirty PR runs no `pull_request` workflow, so the current green predates `main` moving.

### P1: false or over-general sentences

1. **`cam2pixel` Convention (pinhole.py ~L876) says the `and` slip "leaves the shape clause dead, so only the rank is tested".** The clause is live and inverted, not dead: the guard raises iff `rank != 3 and trailing == (4, 4)`. Executed: `cam2pixel(pts, torch.eye(4))` raises the guard's own `ValueError("... Bx4x4. Got torch.Size([4, 4])")`, while `cam2pixel(pts, torch.eye(3))` passes it. The `pixel2cam` block two functions up describes the same idiom correctly ("raises exactly when ... not rank 4 **and** its second dimension is 1"); make `cam2pixel` say the same. #4266's body inherits the "dead" wording and should be corrected with it.
2. **`scale` / `scale_` / class warning: "an integer height/width is promoted to floating point" and "raises RuntimeError" hold only for a floating-point `scale_factor`.** Executed on an int64-size camera: `cam.scale(2)` and `cam.scale(torch.tensor([2]))` return int64 `[12]`; `cam.scale_(2)` succeeds. Only `0.5`, `2.0`, `tensor([2.])` promote or raise. Qualify both sentences, and give the #4265 pin an int-factor arm so it discriminates.
3. **`scale_` raises after it has written the intrinsics.** The body scales `fx, fy, cx, cy` in place and only then hits `self.height *= s`. Executed: after `scale_(0.5)` raises, the caller's `K` reads fx=50, cx=2, cy=1.5 while height/width are still 6/8. The block should say the caller's camera is left half-scaled, and `test_wart_scale_inplace_rejects_integer_image_size_4265` should assert that post-state, otherwise a #4265 repair that raises before mutating is indistinguishable from one that mutates and then raises.
4. **`project_points_z1` Convention and warning: "a point with z = 0 returns inf".** Executed: `[[1, 2, 0]]` gives `[inf, inf]`, `[[-1, 2, 0]]` gives `[-inf, inf]`, `[[0, 2, 0]]` gives `[nan, inf]`, `[[0, 0, 0]]` gives `[nan, nan]`. A caller who filters with `isinf` misses the on-axis case. Say "`+inf`, `-inf` or `nan` per component"; #4318 already phrases the identical fact correctly for `Z1Projection.project`, so the two should agree. The pin probes only `(1, 2, 0)`; add a negative and an on-axis coordinate.

### P1: test matrix

5. **`test_wart_cam2pixel_overflows_to_inf_at_z_zero_in_float16_4267` skips the float16 cell and builds float16 tensors inside the float32 cell.** The blocking `cpu-float16` job therefore reports this float16 pin as skipped, and `_is_subprocess_isolated_test` keys on the `dtype` param, so under `--isolate-half-precision` the half kernel runs unisolated in the float32 subprocess (kornia#4126). Fold it into `test_wart_cam2pixel_epsilon_makes_the_singular_divide_finite_4267` with an `if dtype == torch.float16: assert isinf else: assert_close(1e14)` branch and delete both skips.

### P2

6. `inverse_pinhole_matrix` warning: "a zero focal length gives a large finite number" is `inf` in float16 (`1/(0 + 1e-6)` exceeds 65504). The `cam2pixel` block in this same diff carves out float16 for the same pattern; do it here too.
7. `scale_pinhole` still carries the "going to be deprecated in version 0.2 in favour of :attr:`kornia.PinholeCamera.scale()`" note that this PR rewrote on its two siblings; #4268 lists it among the six legacy names.
8. `docs/source/geometry.camera.perspective.rst` line 5 still describes the module as "transforming 3D world coordinates into 2D image plane coordinates and vice versa", which the rewritten `unproject_points` docstring on that very page now contradicts.
9. Typo `instrisics` in `intrinsics_inverse` (Copilot's note; the PR touches that docstring).
10. The anchor's "(B, 3, 3)" for the four depth functions is right, but `depth.py`'s Args still say `(3, 3)` and `(*, 3, 3)` until #4317 lands. Fine if 5a merges first and 5c follows; just do not let 5c stall.
11. `test_wart_cam2pixel_epsilon_makes_the_singular_divide_finite_4267` pins `1e14` at `z = 0`, which a guarded divide (`x / where(|z| < eps, eps, z)`) reproduces exactly, so the pin never observes the "eps biases every small z" mechanism the docstring names. A `z = 1e-12` point (5e13 today vs 1e14 guarded) would.
12. Cross-cutting for 5a to 5d: the pin comments cite "audit labels 5a-pp-13", "the batch-5 worktree" and similar. Those resolve to nothing in this repository (none of the batch-4 pins on `main` carry them), so the recorded "source" is a private artefact. Keep the executed snippet and the commit hash, drop the labels or spell out what each measured.

Checked and clean: every remaining Convention claim I probed (empty point set through `PinholeCamera.project`, the four depth functions rejecting an unbatched `K`, `project_points` broadcasting a `(3, 3)` `K`, `from_parameters` size dtype, the `pixel2cam` three-channel silent path, the `pinhole_matrix` eps perturbation, the non-rendering of the legacy 12-vector API), the strict xfail for #4279, CHANGELOG placement, and all Sphinx targets except the pre-existing `ShapeError` one.

ducha-aiki:
## Final review of #4294 at `564f6792`

**Verdict: approve once the two test nits below are fixed; nothing in the prose blocks.** All five P1 items of the round-1 review (issuecomment-5558769956) and the six P2s are resolved, and the branch has absorbed the four `main` fixes that landed on top of its pins (#4297 → #4279, #4314 → the `pixel2cam` depth guard, #4355 → #4282, plus the `kornia_rs` floor bump) correctly. Everything below was executed on a fresh worktree at `564f6792` (torch 2.14.0, `kornia.__file__` printed from the worktree).

### Verified

- **Runtime unchanged.** Docstring-stripped `ast.dump` of `pinhole.py`, `perspective.py`, `projection_z1.py` and `projection_orthographic.py` is identical to `origin/main` (`f4532f39`).
- **Tests.** `tests/geometry/camera`: cpu float32+float64 293 passed / 1 skipped; cpu float16 with `--xfail-known-failures --known-failure-profile=cpu-float16` 135 passed / 4 skipped / 13 xfailed; cpu bfloat16 likewise 126 / 2 / 24; mps float32 (`PYTORCH_ENABLE_MPS_FALLBACK` unset) 142 passed / 10 skipped, 0 failed. No `test_wart_*` / `test_convention_*` node is xfailed or skipped by any manifest, and no manifest names a camera node.
- **Round-1 asks, re-executed on the head.** (1) `cam2pixel`: the guard raises on an unbatched `(4, 4)`, passes `(1, 3, 3)` into `transform_points`, and a rank-3 coordinate tensor raises `IndexError` — the block now says exactly that, and #4266's body already carries the same wording. (2) `scale`/`scale_`: `cam.scale(2)` on an int64 size returns int64 `[12]`, `cam.scale_(2)` succeeds, only a float factor promotes or raises; the pin's int-factor arm discriminates. (3) `scale_` post-state: int `height` + float `width` → raises with `fx=50, cx=2`, `height=[6]`, `width=[8.0]`; float `height` + int `width` → `height=[3.0]` already scaled, `width=[8]`. Both sentences of the new bullet are true and the pin asserts `K`, `cam.intrinsics`, `height`, `width` afterwards. (4) `project_points_z1` at `z = 0`: `[1,-2,0] → [inf,-inf]`, `[0,2,0] → [nan,inf]`, `[1,0,0] → [inf,nan]`; the pin covers all three sign classes with `isposinf`/`isneginf`/`isnan` masks. (5) The float16 `cam2pixel` overflow now lives in the float16 cell of the folded pin; the float32 branch also pins `z = 1e-12 → 5e13` (the eps-bias mechanism, round-1 P2 11). Also: `inverse_pinhole_matrix` on a zero focal gives `1e6` (float32/64), `1003520` (bfloat16), `inf` (float16) — matches the new sentence; the `scale_pinhole` note and the `perspective.rst` meta description are fixed; `instrisics` is gone; no `audit label` / `full_audit` / `batch-5 worktree` string survives in `kornia/` or `tests/`.
- **Absorbed `main` fixes.** `grep -rnE "#4279|issues/4279|#4282|issues/4282" kornia/ tests/` hits only the `#4279` regression pin's comment; both warnings and both wart pins are gone, `from_parameters` on `batch_size=2, dtype=float64` gives `height=[6., 6.]` in float64, and `pixel2cam` rejects a rank-5, a rank-3 and a 3-channel depth with the `Bx1xHxW` `ValueError`.
- **Merge commit `7f6542ba`.** The diff against `origin/main` is purely additive docstring/test hunks plus one CHANGELOG bullet; nothing from `main` was reverted.
- CI: pre-commit.ci, docs, doc links, import surface and CodeQL(actions) green on `564f6792`; `Tests on CPU (PR)` was still pending when I posted this.

### P2: test nits

1. **`test_unproject_depth_shape_mismatch` pins an accident.** It asserts that `unproject_points_z1(points (2,1,2), depth (1,2,1))` raises `RuntimeError("Sizes of tensors must match")`. That error comes from the final `torch.cat`, not from any kornia guard, and it fires only because the batch dims happen to disagree after broadcasting: `depth (2,2,1)` against the same points broadcasts silently to `(2, 2, 3)`. So the function does not reject mismatched leading dims; the test name and the docstring sentence it backs ("accepted when their leading dimensions match") would both be read as a rejection contract that does not exist. Either drop the test or rename it to say it pins the broadcast (`(2,2,1)` → `(2,2,3)`) and let the `(1,2,1)` case go — pinning a `torch.cat` message string is fragile across torch versions in any case.
2. **Revert the rename of `main`'s `test_unproject_depth_shapes_4282`.** #4355 added that test on `main` with the issue suffix; this PR drops it to `test_unproject_depth_shapes` while keeping the suffix on its own closed-issue pin `test_convention_from_parameters_fills_every_batch_element_4279`. A pin keeps naming its issue after the fix (AGENTS.md), and a docs PR should not rename a regression test it did not write. With that revert, `test_convention_unproject_points_z1_accepts_both_extension_shapes` is a subset of the `(1, 2)`/`(2, 3)` × `column_depth` cells of that parametrized test and can go.

### P3

3. **PR body is stale.** It still says "No existing test or doctest was modified; every pin is new" (see 2), lists the four deleted wart pins (`..._4279`, `..._pixel2cam_returns_silently_..._4266`, `..._overflows_to_inf_..._4267`, `..._4282`) under Pins, counts "fourteen `.. warning::` blocks", and keeps #4279 and #4282 as rows in "Issues documented here". Squash-merge takes the body as the commit message, so refresh it before merging.

ducha-aiki:
Pushed `66d0f724` addressing the three items of the final review above (issuecomment-5576404451):

1. `test_unproject_depth_shape_mismatch` is gone — it pinned a `torch.cat` error that fires only for the chosen `(1, 2, 1)` depth (a `(2, 2, 1)` depth broadcasts silently to `(2, 2, 3)`), so it recorded no contract of `unproject_points_z1`.
2. `main`'s `test_unproject_depth_shapes_4282` (from #4355) keeps its name; the PR no longer touches any existing test. `test_convention_unproject_points_z1_accepts_both_extension_shapes` is removed as a subset of that parametrized test's `(1, 2)` / `(2, 3)` × `column_depth` cells.
3. PR body refreshed: 27 pins listed as they are now, twelve warnings / eight open issues, #4279 and #4282 moved out of the issues table with a note that #4297 and #4355 fixed them while the PR was open, the `scale_pinhole` "out of scope" reviewer note dropped (it is fixed here), the validation section re-run at this head.

Re-run at `66d0f724` from the branch worktree (torch 2.14.0): `tests/geometry/camera` cpu float32+float64 289 passed / 1 skipped; cpu float16 (profile `cpu-float16`) 133 passed / 4 skipped / 13 xfailed; cpu bfloat16 124 / 2 / 24; mps float32 with `PYTORCH_ENABLE_MPS_FALLBACK` unset 140 passed / 10 skipped / 0 failed. Pinned `ruff@0.16.6` check and format clean. `git diff origin/main -- tests/geometry/camera/test_projections.py` deletes no line.

ducha-aiki (commented):
Final review at 66d0f724baecb476253443c6b671fb95ab148c20: no remaining actionable findings. The previous projection-equivalence/singular-output, extension-shape, and integer-scale-factor findings have been addressed. The extension documentation now reflects the fix merged in #4355.

Validation in an isolated detached worktree, using PyTorch 2.14.0 and verifying the imported Kornia path:

- Full CPU camera suite, float32/float64: 289 passed, 1 skipped.
- Convention/wart tests, CPU float16/bfloat16: 50 passed, 4 skipped.
- Convention/wart tests, MPS float32: 27 passed.
- Doctests in all four changed library modules: 13 passed.
- Docstring-stripped ASTs match the merge base in all four library files. Existing test functions are unchanged; all 27 test pins are additive.

The code and documentation review is clear to merge once required checks pass. Read the Docs and pre-commit.ci pass, but five GitHub Actions checks remain pending at review time. I did not run CUDA, other PyTorch versions, or a full local documentation build; the exact-equality inverse pin therefore remains unverified on those other backends/versions.

Submitted as a comment review because the authenticated account is also the PR author.

PR #4295, opened 2026-09-05
fix(feature): lazy-import onnxruntime and numpy in OnnxLightGlue

### LINKED ISSUE #4260: `import kornia` loads onnxruntime eagerly through kornia.feature.lightglue_onnx (bypasses LazyLoader, ~20 ms)
## What happens

`kornia/feature/lightglue_onnx/lightglue.py:28-33` imports `onnxruntime` (and `numpy`) at module level:

```python
try:
    import numpy as _np  # noqa: ICN001
    import onnxruntime as _ort
except ImportError:
    _np = None  # type: ignore
    _ort = None
```

That module is reachable from a plain `import kornia`, so on any machine that has onnxruntime installed, importing kornia loads the ORT native extension for a class (`OnnxLightGlue`) nobody instantiated.

Measured on `main` @ 92e5fe8d, torch 2.14.0, onnxruntime 1.29.0, `python -X importtime -c "import kornia"`:

```
import time:       247 |      19895 |   onnxruntime          (cumulative, µs)
import time:     18742 |      18742 |   onnxruntime.capi.onnxruntime_pybind11_state
import time:       363 |     684244 | kornia
```

So ~20 ms of a 684 ms import (torch itself is 579 ms of that), and `import kornia` behaves differently depending on whether an optional package is present. It is the only ONNX consumer that bypasses the lazy loader: `kornia/onnx/module.py`, `kornia/onnx/sequential.py` and `kornia/core/mixin/onnx.py` all go through `from kornia.core.external import onnxruntime as ort`.

## Fix

- Replace the `try: import` block with the `LazyLoader` handles from `kornia.core.external` (`numpy`, `onnxruntime`), matching `kornia/onnx/module.py`.
- Keep the "onnxruntime is not installed" check in `OnnxLightGlue.__init__` (e.g. `importlib.util.find_spec("onnxruntime") is not None`, or let the loader's `ImportError` surface) so the error message stays at construction time rather than at first `forward`.
- Add a pin: a subprocess test that runs `python -c "import kornia, sys; assert 'onnxruntime' not in sys.modules"` (skip when onnxruntime is not installed, otherwise the assertion is vacuous). That is the check that would have caught this, and it also guards the rest of the LazyLoader registry.

Found while auditing declared vs. imported dependencies; the rest of that audit is tracked separately.

ducha-aiki:
## What changed

`kornia/feature/lightglue_onnx/lightglue.py` had a module-level
`try: import numpy as _np; import onnxruntime as _ort` block. Because
`kornia.feature` imports `OnnxLightGlue` at package-import time, plain
`import kornia` eagerly loaded onnxruntime (and numpy) whenever onnxruntime
was installed, even for users who never touch `OnnxLightGlue`.

- Replaced the try/except block with the standard kornia lazy-loading
  pattern already used by every other ONNX consumer (see
  `kornia/onnx/module.py`):
  ```python
  from kornia.core.external import numpy as np
  from kornia.core.external import onnxruntime as ort
  ```
  and renamed the three call sites (`_ort.InferenceSession`, `_np.float32`)
  to `ort.`/`np.`.
- In `OnnxLightGlue.__init__`, replaced the `_ort is not None` /
  `_np is not None` checks (which no longer exist once the eager try/except
  is gone) with `importlib.util.find_spec("onnxruntime") is not None` /
  `find_spec("numpy") is not None`, preserving the exact same
  `KORNIA_CHECK` error messages ("onnxruntime is not installed." /
  "numpy is not installed.") when either dependency is missing.
- Added a subprocess-based regression test,
  `test_import_kornia_does_not_import_onnxruntime`, to `tests/test_import.py`
  (next to the two existing subprocess import guards, and deliberately not in
  `tests/feature/test_lightglue_onnx.py`, whose module-level `importorskip("onnx")`
  would gate it on an unrelated package): it `importorskip`s onnxruntime
  (so it's not vacuous), spawns a fresh interpreter with
  `cwd=<repo root>` (derived from `kornia.__file__`), `timeout=300` and
  `check=False` (so a crashing child shows its stderr), running
  `import kornia, sys; print('onnxruntime' in sys.modules)`, and asserts
  the printed value is `False`.
- Updated a now-stale comment in `tests/test_import.py` that described
  onnxruntime as something `kornia.feature.lightglue_onnx` "imports
  eagerly when installed" — that's no longer true after this fix; the
  comment now explains the pre-import is defensive/no-op-safe rather than
  load-bearing.
- Added a `CHANGELOG.md` entry under `## Unreleased` / `### Bug fixes`.

No behavior change other than *when* onnxruntime/numpy get imported:
`OnnxLightGlue(...)` still raises the identical error messages at
construction if either dependency is missing.

Closes #4260

## Why

Every other ONNX consumer in kornia (`kornia/onnx/module.py`) already
routes through `kornia.core.external`'s `LazyLoader` so that importing
`kornia` doesn't pull in optional heavy dependencies. `lightglue_onnx`
was the one remaining holdout doing a real top-level `import onnxruntime`,
adding measurable import-time cost (see verification below) for every
`import kornia` when onnxruntime happens to be installed, regardless of
whether `OnnxLightGlue` is ever used.

## What was verified

All commands run from the worktree root with
`/Users/oldufo/dev/kornia/.venv/bin/python` (torch 2.14.0, onnxruntime
installed locally so the pin is not vacuous). `kornia.__file__` was
confirmed to resolve to
`/Users/oldufo/dev/kornia/.claude/worktrees/deps-audit/kornia/__init__.py`
before running anything.

- `python -X importtime -c "import kornia" 2>&1 | grep -c onnxruntime`
  — **before** (eager import, reverted temporarily via `git stash`): `9`
  matching importtime lines.
  **after** (this fix): `0`.
- `pytest tests/feature/test_lightglue_onnx.py -q` — `2 passed, 8 skipped`
  (the 8 skips are the pre-existing CUDA/slow-marked tests, unaffected by
  this change; unrelated to onnxruntime availability).
- Discrimination check: temporarily restored the old eager
  `try: import numpy as _np; import onnxruntime as _ort` block via
  `git stash` and re-ran just the new pin
  (`test_import_kornia_does_not_import_onnxruntime`) — it **failed** as
  expected (`AssertionError: True` — the child process reported
  `onnxruntime` was in `sys.modules`). Then restored the fix via
  `git stash pop` and confirmed the pin passes again (`2 passed, 8 skipped`
  including this test).
- `pytest --doctest-modules kornia/feature/lightglue_onnx/ -q` —
  `1 skipped` (the one doctest in `utils/download.py` downloads a real
  ONNX checkpoint and is skipped by the repo's `KORNIA_DOCTEST_DOWNLOAD`
  gate by default, unrelated to this change; not touched by this diff).
- `pytest tests/test_api_surface.py -q` — `13 passed` (no public API
  surface changed by this fix).
- `pytest tests/feature -q` — `720 passed, 98 skipped` (broader regression
  sweep of the whole feature module; only pre-existing, unrelated
  warnings such as `torch.jit.script` deprecation and a `PytestUnknownMarkWarning`
  for `pytest.mark.jit`).
- `pytest tests/test_import.py tests/test_doctest_downloads.py tests/core/test_weights_prefetch.py -q`
  — `108 passed` (the import-warning guard and the download-guard tests
  that reference `lightglue_onnx` all still pass; the stale comment in
  `tests/test_import.py` was updated for accuracy).
- `git add -A && pre-commit run --all-files` — all hooks passed
  (`ruff check`, `ruff format`, `codespell`, license header, etc.), run
  twice (once before, once after the CHANGELOG edit).

## What is not covered

- No CUDA hardware is available in this environment, so the
  `test_pretrained_weights_cuda` / CUDA branch of `test_forward` /
  `test_exception` in `tests/feature/test_lightglue_onnx.py` were not
  exercised (they were already skipped before this change and this diff
  does not touch CUDA-specific code paths).
- `pixi run typecheck` was not run through pixi; `uvx --from 'ty>=0.0.9,<0.0.10' ty check`
  on `kornia/feature/lightglue_onnx/lightglue.py` against the project `.venv` reports
  "All checks passed". The change only touches import statements and a
  boolean-returning `find_spec(...) is not None` check, both patterns already used
  elsewhere in kornia (`kornia/onnx/module.py`, `kornia/core/utils.py:xla_is_available`).


ducha-aiki:
**Approve after fixes pushed** — the lazy import is correct and matches `kornia/onnx/module.py`, the construction-time error is unchanged, and the only real defect was where the regression pin lived; I moved it (`4c2aeb31`).

### Verified

All runs from the worktree root with the venv python (torch 2.14.0, onnxruntime 1.29.0, onnx 1.22.0 installed); `kornia.__file__` named the worktree in every block.

Witness, `main` (`e8111a61`) vs PR head, same snippet in a fresh interpreter:

```
python -c "import sys, kornia; print('onnxruntime' in sys.modules, 'numpy' in sys.modules)"
main:    True  True      -X importtime | grep -c onnxruntime -> 9
PR head: False True      -X importtime | grep -c onnxruntime -> 0
import kornia wall, 3 runs: main 0.664/0.659/0.669 s; PR 0.649/0.651/0.644 s (~15 ms, matches #4260's ~20 ms)
```

`numpy` stays in `sys.modules` on both sides — torch imports it and it is a hard dependency — so the "and numpy" half of the title is real for the module but unobservable from `import kornia`. The CHANGELOG only claims onnxruntime, which is right.

Missing-dependency behaviour, `sys.modules["onnxruntime"] = None` then `OnnxLightGlue(device="cpu")`: identical on both sides — `kornia.core.exceptions.BaseError`, message `onnxruntime is not installed.` (byte-identical, diffed).

Discrimination: with `origin/main`'s `lightglue.py` swapped in, the pin fails (`AssertionError: import kornia loaded onnxruntime: True`); restored, it passes.

`tests/feature/test_lightglue_onnx.py` 1 passed / 8 skipped (CUDA, `--runslow`); `tests/test_import.py` 3 passed; `tests/onnx/` 35 passed, 0 skipped; `--doctest-modules kornia/feature/lightglue_onnx/` 1 skipped (download gate, the only doctest there calls `download_onnx_from_url`, not `OnnxLightGlue`). `uvx ruff@0.16.4 check` / `format --diff` clean; `pre-commit run --files` on the four files all passed; `uvx --from 'ty>=0.0.9,<0.0.10' ty check kornia/feature/lightglue_onnx/lightglue.py` "All checks passed" (the PR body lists `ty` as not run). `grep -rnE "#4260|issues/4260" kornia/ tests/ docs/` has no hit in `kornia/`; `grep "^import onnx\|^from onnx\|^import onnxruntime\|^import numpy\|^from numpy" kornia/` finds no remaining eager site outside `kornia/core/external.py`, so "Closes #4260" is not an over-claim — the issue names this one file.

### Findings

1. `tests/feature/test_lightglue_onnx.py:25` — the new pin sat below `pytest.importorskip("onnx")` although it needs only onnxruntime; nothing in that file uses `onnx` at all (its tests need onnxruntime). CI installs `--extra dev`, which today carries `onnx`, so the pin ran; the sibling `build/optional-extras` branch moves `onnx` into a `kornia[onnx]` extra, and the pin would then depend on which extra a leg installs. Moved it to `tests/test_import.py`, next to the two existing subprocess import guards, with `timeout=300`, `check=False` plus a returncode assertion (so a failing child shows its stderr instead of a bare `CalledProcessError`), the `cwd=repo_root` shadowing kept, and `tests/feature/test_lightglue_onnx.py` restored to `main`'s content. Re-ran the discrimination check from the new location: fails on main's `lightglue.py`, passes on the PR. Fixed in `4c2aeb31`. Note for the PR body: the test now lives in `tests/test_import.py`, not `tests/feature/test_lightglue_onnx.py`.

2. Question, not fixed: a *broken* onnxruntime install (spec present, `import onnxruntime` raises) used to hit `_ort = None` and raise `onnxruntime is not installed.` at construction. With `find_spec` the check passes, the weights are downloaded, and `ort.InferenceSession` goes through `LazyLoader._load`, whose default `InstallationMode.ASK` calls `input()` — in a non-interactive process that is an `EOFError` (reproduced with a stub package on `sys.path`). This is exactly what every other ONNX consumer does today, so it is the pattern the issue asked for rather than a regression of this PR; flagging in case the maintainer wants `OnnxLightGlue.__init__` to touch `ort.get_available_providers()` before downloading.

3. Latent, not fixed: under `--doctest-modules`, `LazyLoader._load` short-circuits for non-dev loaders, so `OnnxLightGlue(...)` in a doctest would raise `AttributeError: 'NoneType' object has no attribute 'InferenceSession'` (reproduced by appending `--doctest-modules` to `sys.argv`). No doctest instantiates it and the same holds for `kornia/onnx/module.py`, so nothing to do here; recording it for whoever adds a LightGlue-ONNX example.

### Copilot review

- `test_lightglue_onnx.py:25`, pin gated by the unrelated `importorskip("onnx")`: VALID, see finding 1 — moved to `tests/test_import.py`.
- `:49`, subprocess test without `timeout`: VALID — the moved test uses `timeout=300` like its neighbours.

### Not covered

CUDA paths (`test_pretrained_weights_cuda`, the CUDA branch of `test_forward`/`test_exception`) and the `--runslow` CPU weight downloads were not run; no GPU here and the checkpoints are not cached. The stacked #4307 deletes this pin from `tests/feature/test_lightglue_onnx.py` in favour of `tests/test_declared_imports.py`; after this move it will need to drop it from `tests/test_import.py` instead (or keep both, they do not conflict).

ducha-aiki:
**Final review: Approve.** Full re-review at head `4c2aeb31` (three files, `+46/-14` against `origin/main` `e8111a61`), not a delta. `import kornia` no longer loads onnxruntime, the construction-time error is byte-identical, the pin lives where no unrelated `importorskip` can gate it and fails on `main`'s file, and CI is fully green.

`kornia/feature/lightglue_onnx/lightglue.py` replaces its module-level `try: import numpy / onnxruntime` with the `LazyLoader` handles from `kornia.core.external` (the pattern `kornia/onnx/module.py` already uses); `OnnxLightGlue.__init__` keeps the two `KORNIA_CHECK`s with the same messages, now on `importlib.util.find_spec`; the three call sites become `ort.InferenceSession` / `np.float32`. `tests/test_import.py` gains `test_import_kornia_does_not_import_onnxruntime` (importorskip on onnxruntime, fresh interpreter with `cwd=repo_root`, `timeout=300`, `check=False` plus a returncode assertion) next to the two existing subprocess guards, and its stale comment about the eager import is rewritten. CHANGELOG bullet under `Unreleased / Bug fixes` cites #4295.

### Verified at `4c2aeb31`

Worktree root, `.venv` python (torch 2.14.0, onnxruntime 1.29.0 installed), `kornia.__file__` naming the checkout.

```
pytest tests/test_import.py tests/feature/test_lightglue_onnx.py tests/onnx tests/test_api_surface.py -> 52 passed, 8 skipped (CUDA / --runslow)
pytest --doctest-modules kornia/feature/lightglue_onnx/ -> 1 skipped (download-gated example in utils)
uvx ruff@0.16.4 check / format --diff -> clean; pre-commit run --files (3 files) -> all 9 hooks pass
uvx ty check kornia/feature/lightglue_onnx/lightglue.py -> All checks passed
```

- Fresh interpreter, `import kornia`: `'onnxruntime' in sys.modules` → `False` (numpy stays loaded, as it must: torch imports it and it is a hard dependency); `-X importtime | grep -c onnxruntime` → `0` (was `9` on `main` last round); `import kornia` wall 0.656/0.663/0.657 s.
- Missing-dependency behaviour: with `sys.modules["onnxruntime"] = None`, `OnnxLightGlue(device="cpu")` raises `kornia.core.exceptions.BaseError('onnxruntime is not installed.')`, the same type and text as `main`.
- Discrimination: with `origin/main`'s `lightglue.py` checked out over the head, the pin fails with `AssertionError: import kornia loaded onnxruntime`; restored, tree clean.
- No eager `import onnx|onnxruntime|onnxscript|numpy` remains under `kornia/` outside `kornia/core/external.py`, and no `#4260` reference survives in `kornia/`, so "Closes #4260" is not an over-claim.
- CI on the head: 41/43 green; the two `-/-` rows are the RTD and pre-commit.ci contexts that never report.
- PR body: the bullet that placed the test in `tests/feature/test_lightglue_onnx.py` and the "ty not run" bullet are corrected in the body now.

### Findings

None blocking. Two latent notes, both the existing pattern rather than this PR's doing:

1. A *broken* onnxruntime install (spec present, import raises) used to hit `_ort = None` and fail fast at construction; now `find_spec` passes, the weights download, and `LazyLoader._load` under the default `InstallationMode.ASK` calls `input()`, which in a non-interactive process is an `EOFError`. Every other ONNX consumer behaves the same way today. If that ever matters, touching `ort.get_available_providers()` in `__init__` before the download would restore the early failure.
2. Under `--doctest-modules`, `LazyLoader._load` short-circuits for non-dev loaders, so an `OnnxLightGlue(...)` doctest would see `None.InferenceSession`. No doctest instantiates it; recorded for whoever adds one.

Sibling note: the stacked #4307 keeps exactly one onnxruntime import pin, `tests/test_declared_imports.py::test_import_kornia_does_not_load_optional_dependencies`, and drops this file's copy in its merge; the two do not conflict in either order.

### Not covered

CUDA paths (`test_pretrained_weights_cuda`, the CUDA branch of `test_forward` / `test_exception`) and the `--runslow` weight downloads; no GPU here and the checkpoints are not cached. No docs build.

PR #4297, opened 2026-09-05
fix(geometry): fill height and width for the whole batch in PinholeCamera.from_parameters

### LINKED ISSUE #4279: W1b: `PinholeCamera.from_parameters` — `height` and `width` are 0 for every batch element after the first

thegoodengineer:

`PinholeCamera.from_parameters(..., batch_size=B)` filled `height` and `width` only for element `0`:

```python
height_tmp = torch.zeros(batch_size, device=device, dtype=dtype)
height_tmp[..., 0] += height
```

Every other parameter (`fx`, `fy`, `cx`, `cy`, `tx`, `ty`, `tz`) is broadcast over the batch, so the camera looks healthy until something reads its image size:

```text
height: [6.0, 0.0]
width : [8.0, 0.0]
fx    : [100.0, 200.0]
tx    : [1.0, 2.0]
```

Both are now built with `torch.full((batch_size,), ...)`, matching the rest of the constructor. `batch_size=1`, the only case that worked before, is unchanged.

Also adds `TestPinholeCamera::test_from_parameters`, parametrized over `batch_size` in `(1, 2, 5)`, asserting every parameter including `height` and `width` over the whole batch.

Closes #4279

## What did you check?

The new test on unmodified `main` fails exactly where the issue says it should, and passes at `batch_size=1`:

```text
4 failed, 2 passed, 38 deselected
FAILED test_from_parameters[cpu-float32-2] - AssertionError: Tensor-likes are not close!
FAILED test_from_parameters[cpu-float32-5]
FAILED test_from_parameters[cpu-float64-2]
FAILED test_from_parameters[cpu-float64-5]
```

`tests/geometry/test_depth_warper.py` builds `batch_size=2` cameras through this helper, so it exercises the changed path; it passes unchanged, since `DepthWarper` takes its output size explicitly rather than from the camera.

```text
$ pytest tests/geometry/camera tests/geometry/test_depth_warper.py tests/geometry/test_depth.py --dtype=float32,float64
================== 332 passed, 1 skipped, 12 warnings in 4.29s ==================

All checks passed!
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -376,6 +376,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   model's dtype instead of coming back as float32. `OnnxLightGlue` without `onnxruntime` raises an
   `ImportError` that names `pip install "kornia[onnx]"`, like the lazy-loader handles do, instead of a
   bare `BaseError`. (#4301)
+* `PinholeCamera.from_parameters` now fills `height` and `width` for the whole batch. It built them with
+  `height_tmp[..., 0] += height` on a `(B,)` zero tensor, so every batch element after the first kept `0`
+  while `fx`, `fy`, `cx`, `cy`, `tx`, `ty` and `tz` were broadcast correctly, and the camera looked healthy
+  until something read its image size. `batch_size=1`, the only case that worked, is unchanged. (#4279)
 * `unproject_meshgrid` now returns the documented `(*, H, W, 3)` shape when `W` is 1. It squeezed the
   meshgrid with a bare `.squeeze()`, which drops the width axis along with the leading batch axis whenever
   `W == 1`, so `unproject_meshgrid(1, 3, K)` and `unproject_meshgrid(3, 1, K)` returned the same shape with
--- a/kornia/geometry/camera/pinhole.py
+++ b/kornia/geometry/camera/pinhole.py
@@ -424,11 +424,9 @@ class PinholeCamera:
         extrinsics[..., 0, -1] += tx
         extrinsics[..., 1, -1] += ty
         extrinsics[..., 2, -1] += tz
-        # create image hegith and width
-        height_tmp = torch.zeros(batch_size, device=device, dtype=dtype)
-        height_tmp[..., 0] += height
-        width_tmp = torch.zeros(batch_size, device=device, dtype=dtype)
-        width_tmp[..., 0] += width
+        # create image height and width, one entry per batch element
+        height_tmp = torch.full((batch_size,), height, device=device, dtype=dtype)
+        width_tmp = torch.full((batch_size,), width, device=device, dtype=dtype)
         return self(intrinsics, extrinsics, height_tmp, width_tmp)
 
 
--- a/tests/geometry/camera/test_pinhole.py
+++ b/tests/geometry/camera/test_pinhole.py
@@ -388,6 +388,32 @@ class TestPinholeCamera(BaseTester):
         assert pinhole.rotation_matrix.shape == (batch_size, 3, 3)
         assert pinhole.translation_vector.shape == (batch_size, 3, 1)
 
+    @pytest.mark.parametrize("batch_size", (1, 2, 5))
+    def test_from_parameters(self, batch_size, device, dtype):
+        height, width = 6, 8
+        fx = torch.arange(1, batch_size + 1, device=device, dtype=dtype) * 100.0
+        fy = fx / 2.0
+        cx = torch.full((batch_size,), width / 2, device=device, dtype=dtype)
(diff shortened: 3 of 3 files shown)


ducha-aiki (dismissed):
Reviewed at `3dcd8d43`: code approved.

The change fills height and width across the entire batch. All 6 focused float32/float64 cases pass. At merge-base 3ea797d4, B=2 produces height [6, 0] and width [8, 0]; the fix produces the requested dimensions for both cameras. A B=1 state comparison is byte-identical. No code findings. Current main merges the source/tests cleanly, but CHANGELOG.md conflicts; please resolve that mechanical conflict and wait for the outstanding float16 job/collector. Existing camera-documentation work must retire its old #4279 behavior pins when rebased after this fix.

Validation ran in isolated checkouts with the repository Python environment (torch 2.14.0, CPU); CUDA was not tested locally.

ducha-aiki (dismissed):

thegoodengineer:
Thanks for the approval, and for the rebase onto `0743e29e`. Main moved again after it (`#4143`, `#4243`), so this had gone `CONFLICTING` and would have stopped reporting CI; rebased onto `217b92e6`, new head `7273df27`. The only conflict was the `CHANGELOG.md` entry, both kept.

`kornia/geometry/camera/pinhole.py` and `tests/geometry/camera/test_pinhole.py` are byte-identical across both rebases, so nothing in the approved diff changed. The push dismissed the approval automatically, sorry about that; it was a rebase only.

Re-ran at the new base: `tests/geometry/camera` + `tests/geometry/test_depth_warper.py` at float32+float64, 223 passed, 1 skipped.

One note on the earlier red, since it was not this diff. The `collector` failure came from `half-precision (float16) / tests (3.11, 2.9.1)` being CANCELLED at exactly six hours, hung in `tests/feature/test_scale_space_detector.py`. The same leg hung the same way on #4300 and passed on every other run of these branches, and locally the touched files at `--dtype=float16,bfloat16` produce only failures already listed in `testing/half_precision_xfails/`, with the new test passing 6/6. It looks like an intermittent hang in that file on the 2.9.1 float16 leg, which is #4214's symptom.

ducha-aiki:
Delta at `7273df27` — still **APPROVED, merge-ready**.

The push is a rebase onto `main` `217b92e6` (after #4243). The code and test hunks are the same as the approved ones: `from_parameters` builds `height_tmp`/`width_tmp` with `torch.full((batch_size,), ...)` instead of a zero tensor with only `[..., 0]` filled, and `test_from_parameters` pins all of `height`, `width`, `fx`…`tz` for `batch_size` 1, 2, 5. `git diff origin/main...7273df27 -- CHANGELOG.md` adds four lines under `### Bug fixes` and deletes nothing. At this head `tests/geometry/camera/test_pinhole.py` is 43 passed, 1 skipped (float32/float64, torch 2.14.0). CI is 40/40 green and the PR is `MERGEABLE`. Nothing left to do on your side.

ducha-aiki (approved):

PR #4233, opened 2026-09-05
fix: guard zero-distance gradients in _cdist

cris-pat-37:
## What does this pull request do?

Fixes a gradient issue in `_cdist` when the squared distance is exactly zero.

The manual `_cdist` fallback currently uses `dm.clamp(min=0.0).sqrt()`. On older supported PyTorch versions, `torch.clamp` passes the gradient through at the boundary, causing `sqrt(0)` to produce NaN gradients. This particularly affects the half-precision fallback path.

The implementation replaces zero values with a safe positive value before `sqrt`, while preserving the exact zero-distance output.

## Related issue or discussion

Refs #4229

## What did you check?

- `pixi run test-module tests/feature/test_matching.py` — 96 passed, 11 skipped
- `pixi run lint` — passed
- `git diff --check` — passed
- Regression test covers `float16` and `bfloat16` with exact zero distances and verifies finite gradients.

## Scope

This PR only changes `_cdist` and its regression test.

## How did AI help?

AI was used to help investigate the reported PyTorch compatibility issue, review the implementation, and identify the appropriate regression test. I manually reviewed the changes and verified them using Kornia's test and lint tooling.

PR #4299, opened 2026-09-05
fix(geometry): finite gradients for solve_cubic at its acos boundary

### LINKED ISSUE #4290: solve_cubic's D<=0 branch differentiates acos at its own domain boundary, giving NaN gradients for repeated/near-repeated real roots
## Summary

`solve_cubic`'s `D <= 0` branch (three-real-roots case, Cardano's trigonometric solution) computes `acos(R / sqrt(-Q3))` without guarding the fact that `d(acos)/dx = -1/sqrt(1-x^2)` is unbounded at `x = +-1`. The branch condition guarantees the ratio lies in `[-1, 1]` (`D = Q3 + R^2 <= 0` implies `R^2 <= -Q3`), but a cubic with a repeated or near-repeated real root — a routine, not exotic, configuration reached through `solve_quartic`'s resolvent cubic whenever the original quartic has a repeated or close real root pair — pushes the ratio to exactly that boundary. The forward value is fine there; the gradient is not.

Confirmed via `torch.autograd.set_detect_anomaly(True)`, not inferred by reading:

```
Function 'AcosBackward0' returned nan values in its 0th output.
  File ".../kornia/geometry/solvers/polynomial_solver.py", line 185, in solve_cubic
    theta_D_zero = torch.acos(R[mask_D_zero] / torch.sqrt(-Q3[mask_D_zero]))
```

Same failure shape as `quaternion_exp_to_log`/`euler_from_quaternion`'s `acos`/`asin` boundary (#4007, fixed in #4228) — different call site, not covered by that fix.

## Reproduction

```python
import torch
import kornia

torch.autograd.set_detect_anomaly(True)

# (x-2)^2 * (x-3) * (x+1) -- roots -1, 2, 2, 3. This is kornia's OWN existing test fixture
# (tests/geometry/solvers/test_polynomial_solver.py, TestQuarticSolver.test_solve_quartic,
# "Case 3: Double Roots") -- it is already in the suite and already passes, because that
# test only checks the forward value and never calls .backward(). It fails immediately if you do:
coeffs = torch.tensor([[1.0, -6.0, 9.0, 4.0, -12.0]], requires_grad=True)
roots = kornia.geometry.solve_quartic(coeffs)
roots.sum().backward()  # RuntimeError: Function 'AcosBackward0' returned nan values
```

**Measured scale, not just a single instance:** 20,000 randomly generated real-rooted quartics (4 input distributions — fully random, symmetric ±r/±s pairs, exact double-root pairs, near-repeated clusters), run through `solve_quartic` end to end:

| Distribution | Trials | Non-finite gradient or disconnected graph |
| --- | ---: | ---: |
| All distributions combined | 20,000 | 2,145 (10.7%) |
| Exact double-root pairs only | 5,000 | 1,765 (35.3%) |

This is reachable through any caller of `solve_quartic`/`solve_cubic` inside an autograd graph — differentiable PnP/P3P, essential-matrix decomposition, or any minimal solver built on these primitives — whenever the input is at or near a repeated-root degeneracy, which real camera configurations (near-coplanar points, near-symmetric point sets) produce routinely, not just as a synthetic edge case.

## Root cause

`kornia/geometry/solvers/polynomial_solver.py:182` (line numbers as of `main` at the time of filing):

```python
if torch.any(mask_D_zero):
    theta_D_zero = torch.acos(R[mask_D_zero] / torch.sqrt(-Q3[mask_D_zero]))
```

A **plain `torch.clamp(ratio, -1, 1)` before the `acos` call does not fix this.** Verified directly: clamping only guards the *value* against exceeding the domain — it does nothing about differentiating a function whose *derivative* diverges at a value that is already legitimately inside the domain (`acos'(1.0)` is `-inf` regardless of how you arrived at exactly `1.0`). Re-running the reproduction above with a naive clamp still reproduces `AcosBackward0 returned nan values` identically.

## Suggested fix

The pattern already established in this repo for the identical failure shape (#4228): route the boundary through `.acos()` on a *detached* copy for the value, and through `.acos()` on a substituted safe argument for the gradient, so autograd never differentiates `acos` at `+-1` at all.

```python
ratio_D_zero = R[mask_D_zero] / torch.sqrt(-Q3[mask_D_zero])
ratio_D_zero = torch.clamp(ratio_D_zero, min=-1.0, max=1.0)
at_boundary_D_zero = ratio_D_zero.abs() >= 1.0
safe_ratio_D_zero = torch.where(at_boundary_D_zero, torch.zeros_like(ratio_D_zero), ratio_D_zero)
theta_D_zero = torch.where(
    at_boundary_D_zero, ratio_D_zero.detach().acos(), safe_ratio_D_zero.acos()
)
```

**Verified, not asserted:**
- Forward values are byte-identical to the unpatched version on every one of the 5 existing `TestCubicSolver`/`TestQuarticSolver` fixtures, plus every failing case surfaced by the random search.
- The two pre-existing gradcheck-passing fixtures (distinct-root quartic, biquadratic) still produce identical, finite gradients after the patch — no regression.
- Re-running the same 20,000-trial search with the fix applied: failures drop from 2,145/20,000 to **1,269/20,000**.

## Known residual — explicitly not covered by this fix

The fix above does not eliminate every failure in the same random search. A second, distinct issue remains: some inputs (first instance found: two simultaneous exact double-root pairs, e.g. roots `[3.1022, 3.1022, 4.0217, 4.0217]`) raise `element 0 of tensors does not require grad and does not have a grad_fn` — a disconnected autograd graph rather than a `nan`, and `set_detect_anomaly` does not name a specific backward function for it, meaning the graph breaks somewhere in the forward pass (candidate: the `D > 0` branch's `torch.pow(x, 1/3)` cube root, which has its own unbounded derivative at `x = 0` — not confirmed). Naming this now rather than either silently omitting it or folding an unverified second fix into this issue.

## Relationship to existing issues

- **Not the same as #4229** ("`torch.clamp`'s boundary gradient changed between torch versions"). That issue's two `polynomial_solver.py` sites (`torch.sqrt(torch.clamp(R_sq, min=0.0))` and `torch.sqrt(torch.clamp(radicand, min=0.0))`) were investigated directly as part of finding this: 20,000 random real-rooted quartics never reached either of them as the actual failure cause — the `acos` site above is the real mechanism. #4229's confirmed-reachable instance (`kornia/feature/matching.py`) is separately already covered by PR #4233.
- **Same shape as #4007 / PR #4228**, different call site (`solve_cubic` vs. the quaternion conversions), not covered by that fix.

## Environment

kornia `main`, torch 2.11.0, CPU, python 3.11.

tritsystem:

Fixes `solve_cubic`'s `D <= 0` (three-real-roots, Cardano trigonometric) branch returning `nan` gradients for a repeated or near-repeated real root.

The branch computes `theta_D_zero = acos(R / sqrt(-Q3))`. The branch condition guarantees `|R/sqrt(-Q3)| <= 1` (`D = Q3 + R^2 <= 0` implies `R^2 <= -Q3`), but `acos`'s own derivative, `-1/sqrt(1-x^2)`, is unbounded at `x = +-1`, and a repeated or near-repeated root pushes the ratio to exactly that boundary. Confirmed via `torch.autograd.set_detect_anomaly(True)`, not inferred: `AcosBackward0 returned nan values` at exactly this line.

This is not a rare edge case — a double-root quartic reaches it through `solve_quartic`'s resolvent cubic routinely. Measured on 20,000 random real-rooted quartics: **10.7% hit a non-finite gradient or a disconnected graph overall, 35.3% for exact double-root pairs specifically.** The repro in the added test reuses kornia's own existing test fixture (`TestQuarticSolver`, "Case 3: Double Roots", `[1,-6,9,4,-12]` → roots `-1,2,2,3`) — already in the suite, already passing, only because that test never calls `.backward()`.

A plain `torch.clamp(ratio, -1, 1)` before the `acos` call does **not** fix this — verified directly, it still reproduces the identical `nan`. Clamping only guards the *value* against exceeding the domain; it does nothing about differentiating a function whose *derivative* diverges at a value already legitimately inside the domain.

## Fix

Same pattern already established in this repo for the identical failure shape (#4228, `quaternion_exp_to_log`/`euler_from_quaternion`, tracked in #4007): route the boundary through `.acos()` on a detached copy for the value, and through `.acos()` on a substituted safe argument for the gradient, so autograd never differentiates `acos` at `±1`.

Closes #4290. Not the same as #4229 — that issue's two `polynomial_solver.py` sites were investigated as part of finding this (20,000 trials never reached either as the actual cause); #4229's confirmed-reachable instance (`kornia/feature/matching.py`) is separately already covered by #4233.

## What did you check?

- Forward values byte-identical to the unpatched version on all 5 existing `TestCubicSolver`/`TestQuarticSolver` fixtures, plus every failing case from the random search.
- Both pre-existing gradcheck-passing fixtures (distinct-root quartic, biquadratic) unchanged after the patch — no regression.
- Re-ran the same 20,000-trial random search with the fix applied: failures drop from 2,145/20,000 to 1,269/20,000. The residual (a distinct, disconnected-graph failure in a different code path, e.g. two simultaneous exact double-root pairs) is named explicitly in the linked issue and intentionally not addressed here — a different mechanism (`set_detect_anomaly` doesn't attribute it to a specific backward function), out of scope for this fix.
- New test added: `TestCubicSolver.test_convention_gradient_is_finite_at_the_acos_boundary_4290`, RED against unpatched `main` (`AcosBackward0 returned nan values`), GREEN with the fix.

CC: @ducha-aiki

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -335,6 +335,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the only one. `kornia.feature.lightglue.Attention.enable_flash`, the attribute that combined the
   constructor argument with the probe result (and so always equalled the argument, because SDPA is
   present on every supported torch), was renamed to `Attention.allow_flash`. (#4287)
+* `solve_cubic`'s `D <= 0` (three-real-roots) branch no longer returns `nan` gradients for a
+  repeated or near-repeated real root (#4290). It differentiates `acos(R / sqrt(-Q3))`, whose own
+  derivative `-1/sqrt(1-x^2)` is unbounded at `x = +-1`; the branch condition guarantees the ratio
+  lies in `[-1, 1]`, but a repeated or near-repeated root pushes it to exactly that boundary,
+  where the value is correct but the derivative diverges. A double-root quartic reaches this
+  through `solve_quartic`'s resolvent cubic routinely, not as an edge case: measured on 20,000
+  random real-rooted quartics, 10.7% hit a non-finite gradient or a disconnected graph overall,
+  35.3% for exact double-root pairs specifically. Same failure shape as the `acos`/`asin`
+  boundary in `quaternion_exp_to_log`/`euler_from_quaternion` (#4007, fixed in #4228) — this is a
+  different call site, not covered by that fix, using the same guard: differentiate a substituted
+  safe argument, take the value from a detached copy of the real (possibly boundary) argument.
+  Forward values are unchanged; re-running the same 20,000-trial search with the fix applied
+  drops the failure count to 1,269 — a real, separate residual (a disconnected-graph case with
+  two simultaneous double-root pairs) remains and is tracked in #4290, not fixed here.
 
 * `RandAugment`, `AutoAugment`, `TrivialAugment` and `AugMix` no longer silently upcast half-precision
   batches to `float32`. `OperationBase.forward` — the shared gate every auto-augment op routes through —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -336,7 +336,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   constructor argument with the probe result (and so always equalled the argument, because SDPA is
   present on every supported torch), was renamed to `Attention.allow_flash`. (#4287)
 * `solve_cubic`'s `D <= 0` (three-real-roots) branch no longer returns `nan` gradients for a
-  repeated or near-repeated real root (#4290). It differentiates `acos(R / sqrt(-Q3))`, whose own
+  repeated or near-repeated real root (#4290, #4299). It differentiates `acos(R / sqrt(-Q3))`, whose own
   derivative `-1/sqrt(1-x^2)` is unbounded at `x = +-1`; the branch condition guarantees the ratio
   lies in `[-1, 1]`, but a repeated or near-repeated root pushes it to exactly that boundary,
   where the value is correct but the derivative diverges. A double-root quartic reaches this
--- a/kornia/geometry/solvers/polynomial_solver.py
+++ b/kornia/geometry/solvers/polynomial_solver.py
@@ -179,7 +179,22 @@ def solve_cubic(coeffs: torch.Tensor) -> torch.Tensor:
     mask_D_zero_solutions = (a_D_zero <= 0) & (a_Q_zero != 0)
 
     if torch.any(mask_D_zero):
-        theta_D_zero = torch.acos(R[mask_D_zero] / torch.sqrt(-Q3[mask_D_zero]))
+        # d(acos)/dx = -1/sqrt(1-x^2) is unbounded at x = +-1. The branch condition (D <= 0)
+        # guarantees |ratio_D_zero| <= 1 (D = Q3 + R^2 <= 0 implies R^2 <= -Q3), but a repeated
+        # or near-repeated real root pushes the ratio to exactly that boundary, where the
+        # *value* is fine but the *derivative* diverges -- same shape as the acos/asin boundary
+        # in quaternion_exp_to_log/euler_from_quaternion (#4007, fixed in #4228). A plain
+        # `.clamp(-1, 1)` does not help here: it only guards the value, not the diverging
+        # derivative of a value already inside the domain. Route the boundary through `.acos()`
+        # on a detached copy for the value and through `.acos()` on a substituted safe argument
+        # for the gradient, so autograd never differentiates `acos` at +-1 at all. (#4290)
+        ratio_D_zero = R[mask_D_zero] / torch.sqrt(-Q3[mask_D_zero])
+        ratio_D_zero = torch.clamp(ratio_D_zero, min=-1.0, max=1.0)
+        at_boundary_D_zero = ratio_D_zero.abs() >= 1.0
+        safe_ratio_D_zero = torch.where(at_boundary_D_zero, torch.zeros_like(ratio_D_zero), ratio_D_zero)
+        theta_D_zero = torch.where(
+            at_boundary_D_zero, ratio_D_zero.detach().acos(), safe_ratio_D_zero.acos()
+        )
         sqrt_Q_D_zero = torch.sqrt(-Q[mask_D_zero])
         x0_D_zero = 2 * sqrt_Q_D_zero * torch.cos(theta_D_zero / 3.0) - b_a_3[mask_D_zero]
         x1_D_zero = 2 * sqrt_Q_D_zero * torch.cos((theta_D_zero + 2 * _PI) / 3.0) - b_a_3[mask_D_zero]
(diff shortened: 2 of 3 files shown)


ducha-aiki (changes requested):
Review at `f309d327`: implementation accepted as a finite boundary-gradient convention; documentation follow-up requested.

The shield removes the reproduced NaNs and preserves forward values. Focused polynomial-solver tests: 117 passed on CPU / torch 2.14.0; current CI is green, including the older torch matrix. Older torch was not run locally.

At a repeated root the root-wise derivative is undefined, so the detached boundary path is a finite surrogate, not the mathematical root Jacobian. This is not a demand to implement an impossible finite derivative: please state that convention in the public solve_cubic docstring (and note its effect through solve_quartic), so differentiable callers do not interpret finite output as a correct derivative for arbitrary root losses. The existing internal comment explains the mechanism but the public API does not describe the limitation. Suggested wording: "At the acos boundary, backward suppresses the derivative of the acos argument to keep gradients finite. Repeated-root derivatives are undefined; this is a surrogate convention, not a mathematical Jacobian."

Two mechanical release-note items: add PR #4299 to the CHANGELOG reference (currently only issue #4290), and retire the resolved #4290 tag in the production comment during the required issue-number closeout search; the test pin can keep its issue number. I recommend approval once these documentation edits are made.

ducha-aiki:
Delta re-review at `a62504b2` — **APPROVED**, merge-ready (CI 40/40 green, `MERGEABLE`).

Everything the 09-06 review asked for landed: the surrogate-gradient convention is now in `solve_cubic`'s docstring with a cross-reference from `solve_quartic`, the `#4290` tag is gone from the production comment (`grep -rnE "#4290|issues/4290" kornia/` is empty at this head; the test keeps naming it, as pins should), and the CHANGELOG entry cites the PR. Focused run at this head: `tests/geometry/solvers` 60 passed on float32/float64; the float16/bfloat16 failures in that file (8) are the same 8 `TestQuarticSolver` cases that fail on `main`, none new.

Two things you should know, neither blocks this PR:

1. **#4329 (opened today) is a second fix for the same issue**, with a different shield (`acos` at `ratio` clamped by one `eps`, value restored through a detached difference). I ran each PR's tests against the other's code: your pin passes on #4329's code and #4329's three gradient tests pass on yours, so the two post-images agree on every case either of you pinned. This PR came first, was reviewed, and answered its review; I have recommended on #4329 that it be closed in favour of this one, with its forward-value pin invited as a follow-up.
2. **While checking both PRs on a random batch I found a separate, pre-existing NaN in `solve_cubic`'s batched backward — filed as #4334.** A three-real-root row gets `nan` gradients whenever the same batch holds a one-real-root row, because the `D > 0` branch evaluates `sqrt(D)` and `-Q / AD` over every `|R| > 0` row. Your fix and #4329's are both correct for what they claim and neither touches that branch, so it stays a separate issue; `solve_quartic` inherits it. If you want it, it is yours — it is the same file and the fix shape is a one-mask change.

ducha-aiki (approved):

PR #4300, opened 2026-09-05
fix(enhance): return a finite image at jpeg_quality=0

### LINKED ISSUE #4205: jpeg_codec_differentiable returns all-NaN at jpeg_quality=0, and the tests sample it at random
### Describe the bug

`jpeg_codec_differentiable` returns an **all-NaN** image when `jpeg_quality` is exactly `0`, on every device, with a NaN gradient to match. The docstring documents the supported range as `[0, 100]` (`kornia/enhance/jpeg.py:519`), so `0` is a documented input, not a misuse.

The cause is an unguarded division in `_jpeg_quality_to_scale` (`kornia/enhance/jpeg.py:236-241`):

```python
scale = _differentiable_polynomial_floor(
    torch.where(
        compression_strength < 50,
        5000.0 / compression_strength,   # <- inf at quality 0
        200.0 - 2.0 * compression_strength,
    )
)
```

At `compression_strength == 0` the taken branch is `5000.0 / 0` -> `inf`, and the polynomial floor of `inf` propagates NaN through the whole codec.

### Reproduction steps

```python
import torch, kornia

img = torch.rand(2, 3, 32, 32)
q = torch.tensor([0.0, 50.0], requires_grad=True)
out = kornia.enhance.jpeg_codec_differentiable(img, q)
print(out.isnan().flatten(1).any(1).tolist())   # [True, False]
out.nansum().backward()
print(q.grad.tolist())                          # [nan, 0.000528...]
```

Verified identical on `cpu` and `mps`, torch 2.14.0, kornia at `1e322cb1`. It is elementwise, so only the batch entries whose quality is `0` are poisoned. Quality `1` is fine (`5000/1 = 5000`), so the defect is a single point.

### Why this matters beyond the edge case: it makes CI intermittently red

`tests/enhance/test_jpeg.py` draws the quality with **`torch.randint(low=0, high=100, ...)` at 16 sites**, unseeded, so every run has a real chance of drawing a `0` and producing NaN. Tests that only assert a shape survive it; tests that compare values do not — for example `TestDiffJPEG::test_keyword_argument` calls the codec twice with the *same* inputs and asserts closeness, which fails because `NaN != NaN`:

```
Mismatched elements: 3072 / 6144 (50.0%)
Greatest absolute difference: nan at index (1, 0, 0, 0) (up to 1e-05 allowed)
```

The 50.0% is one whole image of the `B=2` batch — the entry that drew quality `0`.

Measured rate, 20 consecutive plain CPU runs of that one file:

```
$ for i in $(seq 1 20); do pytest tests/enhance/test_jpeg.py --device=cpu --dtype=float32 -q; done
run 10: 1 failed, 11 passed
FAILED 1 / 20
```

So roughly **5% of runs of this file go red on any job**, CPU included — it is not MPS-specific. I hit it in a full-suite MPS run while re-recording the strict known-failure baseline for #4202, where an unlisted failure is a hard gate failure; it passed on the immediately following identical run.

### Expected behavior

Two things, and I think both are wanted:

1. `jpeg_quality=0` should return a finite, heavily-compressed image rather than NaN — clamp the strength away from zero before the division, or document and validate the range as `[1, 100]` and reject `0`. Whichever is chosen, the docstring and the code should agree; today they do not.
2. `tests/enhance/test_jpeg.py` should not sample the failing value at random. Drawing from `low=1` (or seeding the draws) removes the flake independently of the fix, and a dedicated pinned case for the boundary is better coverage than a 1-in-100 lottery spread over 16 tests.

Worth noting for whoever picks this up: `RandomJPEG` defaults to `jpeg_quality=50.0` and samples a continuous uniform, so hitting exactly `0.0` through the augmentation has measure zero. The integer `randint` draws in the tests are what make it reachable in practice.

### Environment

```
kornia 1e322cb1 | torch 2.14.0 | python 3.11 | OS macOS 26.5.1 arm64 | MPS=True
```
Reproduced on both the `cpu` and `mps` devices; the defect is in device-independent Python arithmetic.

thegoodengineer:

`jpeg_codec_differentiable` returned an **all-NaN** image at `jpeg_quality=0`, with a NaN gradient to match. `0` is inside the `[0, 100]` range the function documents *and* validates with `KORNIA_CHECK`, so this is a documented input, not a misuse.

The cause is the unguarded division in `_jpeg_quality_to_scale`: the `< 50` branch computes `5000.0 / compression_strength`, which is `inf` at `0`, and the polynomial floor of `inf` is `NaN`.

A quality of `0` is now treated as `1`, the strongest compression. That is what libjpeg does with the same formula (it clamps quality to a minimum of `1` before the division), so the value the codec produces is the standard's own answer for "compress as hard as possible" rather than an invented one. The guard uses `torch.where` rather than `clamp` so the gradient at exactly `1` stays well defined and independent of the torch version, given #4229. Both docstrings now say what `0` does.

The second half of the issue is CI flakiness. `tests/enhance/test_jpeg.py` drew the quality with unseeded `torch.randint(low=0, high=100, ...)` at 16 sites, so every run had a real chance of drawing `0`; value-comparing tests such as `test_keyword_argument` then failed on `NaN != NaN`, measured at roughly 1 run in 20 on plain CPU. The draws now start at `1`, and the boundary gets three pinned tests instead of a 1-in-100 lottery spread over sixteen.

Closes #4205

## What did you check?

Two of the three new tests fail on unmodified `main`, which is the fail-on-base:

```text
4 failed, 2 passed, 23 deselected
FAILED test_quality_zero_is_finite[cpu-float32]           - assert tensor(False)
FAILED test_quality_zero_is_finite[cpu-float64]
FAILED test_quality_zero_matches_quality_one[cpu-float32] - Greatest relative difference: nan
FAILED test_quality_zero_matches_quality_one[cpu-float64]
```

`test_quality_zero_only_affects_its_own_batch_entry` passes on base as well: the defect is elementwise, so a `0` in the batch never poisoned its batch-mates. It is there to keep the guard elementwise, not as a fail-on-base case, and I would rather say so than imply otherwise.

Quality `>= 1` is unaffected. I compared `main` against this branch over qualities `1, 25, 49, 50, 51, 75, 99, 100` at float32 and float64 on a `(3, 3, 32, 48)` batch, forward and the quality gradient: `torch.equal` on all 18 tensors. The gradient at quality `50` is the same `0.0015030280919745564` before and after.

```text
$ pytest tests/enhance --dtype=float32
==================== 357 passed, 15 skipped, 8 warnings in 10.81s ====================

$ pytest tests/enhance/test_jpeg.py --dtype=float32,float64
================================ 29 passed in 1.54s ================================

$ pytest --doctest-modules kornia/enhance/jpeg.py
================================ 2 passed in 1.39s =================================

All checks passed!
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -354,8 +354,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `jpeg_codec_differentiable` no longer returns an all-`NaN` image at `jpeg_quality=0`, a value inside the
   `[0, 100]` range it documents and validates. `_jpeg_quality_to_scale` divides `5000` by the quality on the
   `< 50` branch, which is `inf` at `0`, and the polynomial floor of `inf` is `NaN`, poisoning the image and
-  its gradient. A quality of `0` is now treated as `1`, the strongest compression, which is what libjpeg does
-  before the same division. Output and gradients for every quality `>= 1` are byte-identical. The tests drew
+  its gradient. A quality of exactly `0` is now given the scale of quality `1`, which is the table libjpeg
+  gives quality `0`. The guard is that one point: a fractional quality in `(0, 1)` was already finite, keeps
+  its own larger scale, and is untouched, so this endpoint is not the limit of the formula from above.
+  Output and gradients for every quality `> 0` are byte-identical. The tests drew
   the quality with `torch.randint(low=0, high=100)` unseeded at sixteen sites, so roughly one run in twenty of
   `tests/enhance/test_jpeg.py` went red on any job; the draws now start at `1` and the boundary has pinned
   cases of its own. (#4205)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -351,6 +351,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `jpeg_codec_differentiable` no longer returns an all-`NaN` image at `jpeg_quality=0`, a value inside the
+  `[0, 100]` range it documents and validates. `_jpeg_quality_to_scale` divides `5000` by the quality on the
+  `< 50` branch, which is `inf` at `0`, and the polynomial floor of `inf` is `NaN`, poisoning the image and
+  its gradient. A quality of `0` is now treated as `1`, the strongest compression, which is what libjpeg does
+  before the same division. Output and gradients for every quality `>= 1` are byte-identical. The tests drew
+  the quality with `torch.randint(low=0, high=100)` unseeded at sixteen sites, so roughly one run in twenty of
+  `tests/enhance/test_jpeg.py` went red on any job; the draws now start at `1` and the boundary has pinned
+  cases of its own. (#4205)
 
 * `kornia.io.load_image` and `write_image` work on the kornia_rs that a plain `pip install kornia`
   resolves. kornia_rs 0.1.11 moved its image readers and writers from the package root into
--- a/kornia/enhance/jpeg.py
+++ b/kornia/enhance/jpeg.py
@@ -228,17 +228,26 @@ def _jpeg_quality_to_scale(
 
     Args:
         compression_strength (torch.Tensor): Compression strength ranging from 0 to 100. Any shape is supported.
+            A strength of 0 is treated as 1, matching libjpeg, which clamps the quality to a minimum of 1
+            before this same division.
 
     Returns:
         scale (torch.Tensor): Scaling factor to be applied to quantization matrix. Same shape as input.
 
     """
+    # ``5000 / 0`` is ``inf`` and the polynomial floor of ``inf`` is ``NaN``, which poisons the whole codec
+    # for a documented input. libjpeg clamps the quality to a minimum of 1 before this division, so do the
+    # same. ``torch.where`` rather than ``clamp`` keeps the gradient at exactly 1 well defined and
+    # independent of the torch version (see #4229).
+    strength: torch.Tensor = torch.where(
+        compression_strength < 1.0, torch.ones_like(compression_strength), compression_strength
+    )
     # Get scale
     scale: torch.Tensor = _differentiable_polynomial_floor(
         torch.where(
-            compression_strength < 50,
-            5000.0 / compression_strength,
-            200.0 - 2.0 * compression_strength,
+            strength < 50,
+            5000.0 / strength,
+            200.0 - 2.0 * strength,
         )
     )
     return scale
@@ -517,6 +526,7 @@ def jpeg_codec_differentiable(
     Args:
         input: the RGB image to be coded.
         jpeg_quality: JPEG quality in the range :math:`[0, 100]` controlling the compression strength.
+          A quality of 0 is treated as 1, the strongest compression, matching libjpeg.
         quantization_table_y: quantization table for Y channel. Default: `None`, which will load the standard
           quantization table.
         quantization_table_c: quantization table for C channels. Default: `None`, which will load the standard
--- a/kornia/enhance/jpeg.py
+++ b/kornia/enhance/jpeg.py
@@ -228,19 +228,20 @@ def _jpeg_quality_to_scale(
 
     Args:
         compression_strength (torch.Tensor): Compression strength ranging from 0 to 100. Any shape is supported.
-            A strength of 0 is treated as 1, matching libjpeg, which clamps the quality to a minimum of 1
-            before this same division.
+            A strength of exactly 0 is given the scale of strength 1, matching libjpeg, which gives quality 0
+            the quality-1 table. Fractional strengths in (0, 1) are left alone and scale as the formula says.
 
     Returns:
         scale (torch.Tensor): Scaling factor to be applied to quantization matrix. Same shape as input.
 
     """
     # ``5000 / 0`` is ``inf`` and the polynomial floor of ``inf`` is ``NaN``, which poisons the whole codec
-    # for a documented input. libjpeg clamps the quality to a minimum of 1 before this division, so do the
-    # same. ``torch.where`` rather than ``clamp`` keeps the gradient at exactly 1 well defined and
-    # independent of the torch version (see #4229).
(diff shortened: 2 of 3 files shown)


ducha-aiki (changes requested):
Review at `ebd690eb`.

[P2] Keep the guard limited to the zero boundary (`kornia/enhance/jpeg.py:243`). The condition compression_strength < 1.0 also changes valid fractional qualities in (0,1), although this differentiable API accepts continuous [0,100] inputs. With torch.manual_seed(7), a random (1,3,32,32) image, quality 0.5 and all-ones custom Y/C quantization tables, the merge base differs from quality 1 by max_abs 0.4738097; this head makes the outputs exactly equal. That changes previously finite outputs beyond the stated quality-zero repair.

Please narrow the guard to zero and pin a sub-unit fractional quality, or explicitly propose/document the wider behavior change for maintainer agreement. The q=0 fix itself is verified, with finite output/gradient and zero quality gradient at the boundary. Focused JPEG tests: 29 passed on CPU / torch 2.14.0. I did not run older torch locally; the float16 2.9.1 CI leg was still pending. A CHANGELOG-only rebase is also needed.

thegoodengineer:
Good catch, and the measurement is reproducible on my side: with `manual_seed(7)`, a `(1, 3, 32, 32)` image, all-ones Y/C tables, quality `0.5` versus quality `1` gave `max_abs 0.4738096594810486` on the merge base and `0.0` at `ebd690eb`. My `compression_strength < 1.0` was carried over from libjpeg's integer-quality clamp without accounting for the fact that this API takes continuous values, so it swallowed a range that was never broken.

Narrowed to the zero point, new head `4b7082da`:

```python
strength = torch.where(compression_strength == 0.0, torch.ones_like(compression_strength), compression_strength)
```

`0.4738096594810486` is back for quality `0.5`, and quality `0` is still finite with a `0` gradient.

**Verification that nothing else moved.** I diffed this branch against the merge base over 78 tensors: qualities `0.001, 0.25, 0.5, 0.75, 0.999, 1, 25, 49, 49.5, 50, 51, 99, 100`, each with the default tables and with all-ones tables, forward and the quality gradient, at float32 and float64. All 78 are `torch.equal` to the merge base, and none of them was NaN there. Only the exact zero changes.

Worth noting for the record, since it surprised me while checking: with the *default* quantization tables the sub-unit qualities were already indistinguishable from quality `1` (the standard tables times a scale that large saturate to the same result), which is why the wide guard slipped past my earlier A/B. It only shows with custom tables, so the new pin uses all-ones tables as yours did.

**Pinned.** `test_sub_unit_quality_is_untouched` asserts quality `0.5` stays finite and does not collapse onto quality `1`. It fails at `ebd690eb` and passes here.

**Documented.** Both docstrings now say quality `0` is a defined endpoint rather than the limit of the formula from above, since a quality in `(0, 1)` compresses harder than `0` does:

> A quality of exactly 0 is treated as 1, matching libjpeg, which gives quality 0 the quality-1 table. This endpoint is not the limit of the formula as the quality approaches 0 from above: a fractional quality in (0, 1) is left alone and compresses harder still.

The CHANGELOG says the same and now claims byte-identity for every quality `> 0` rather than `>= 1`.

Rebased onto `217b92e6`; both CHANGELOG entries kept, and `jpeg.py` and `test_jpeg.py` are byte-identical across the rebase.

`tests/enhance` at float32: 358 passed, 15 skipped. `test_jpeg.py` at float32+float64: 31 passed. At float16+bfloat16: 30 passed with one failure, `test_cardinality[cpu-bfloat16]`, which is already in `testing/half_precision_xfails/cpu_bfloat16.txt:107` and is not mine. Doctests pass, ruff check and format clean.

One unrelated observation while I was here: the `half-precision (float16) / tests (3.11, 2.9.1)` leg on this PR was CANCELLED at exactly six hours, hung in `tests/feature/test_scale_space_detector.py`, which is what failed the `collector`. The same leg hung the same way on #4297 and passed on every other run of these branches, so it looks like a real intermittent hang in that file on the 2.9.1 float16 leg rather than anything in this diff. That is #4214's symptom exactly. Happy to open a separate issue with both job links if useful.

ducha-aiki:
Delta re-review at `4b7082da` — **APPROVED**, one nit.

The P2 from the review at `ebd690eb` landed exactly: the guard is now `compression_strength == 0.0` (`jpeg.py:243-245`) instead of `< 1.0`, and `test_sub_unit_quality_is_untouched` pins it (2 failed at `ebd690eb`, passes here). The rest of the delta is docstring, comment and CHANGELOG wording.

Re-executed at this head, torch 2.14.0: `jpeg_quality=0` gives a finite forward, a finite image gradient and `q.grad = 0` on CPU float32/float64/float16 and MPS float32, all NaN on `main`. Every `q > 0` output and `q`-gradient (14 qualities, both table sets) is `torch.equal` to `main` — 160/160 on CPU, 56/56 on MPS. A `[0, 50]` batch: row 1 is bit-identical to the lone call on CPU; on MPS it differs by 1 ulp, exactly as on `main`, and equals row 1 of a `[50, 50]` batch, so that is MPS batch-size rounding, not contamination. Fail-on-base 4 failed / 4 passed as the body says, and a `nan_to_num` suppression plant is rejected by the `q`-gradient finiteness line and the quality-1 equality pin. The flake this closes: 0/20 failing runs of `tests/enhance/test_jpeg.py` at head vs 1/20 on `main` (`test_keyword_argument[cpu-float32]`, the issue's exact symptom; it is also the red MPS leg on #4314 and #4140 right now). Doctest and `pre-commit` pass; CI 40/40 green; `MERGEABLE`.

Nit: the CHANGELOG entry cites `(#4205)` only; the file's convention is the PR number too, `(#4205, #4300)`. Not blocking.

One follow-up that is not this PR's: at float16, fractional qualities below ~0.08 are NaN on `main` and here alike (`5000 / q` overflows), so the new "(0, 1) is left alone" sentence holds only partly at float16. Pre-existing; worth an issue.

I am clearing the changes-requested state from the `ebd690eb` round with an approving review.

ducha-aiki (approved):
Approved at `4b7082da`: the `ebd690eb` round's P2 landed (guard narrowed to `== 0.0` with its pin), q=0 finite on CPU and MPS, every q>0 output `torch.equal` to main, CI 40/40, `MERGEABLE`.

PR #4296, opened 2026-09-05
build(deps): drop dead declared dependencies and fix manifest housekeeping

ducha-aiki:
## What changed

- `pyproject.toml:28` — removed `"packaging"` from the runtime `dependencies` list.
  Evidence it's dead: `grep -rn "packaging" kornia/` matches only
  `kornia/core/_compat.py:66`, a comment noting that `_compat.py` parses
  versions by hand *instead of* using `packaging.version` (so dynamo can
  constant-fold it) — no actual import anywhere under `kornia/`.
- `pyproject.toml:43` — removed `"numpy<3"` from the `dev` extra.
  `numpy<3` in `dev` was the only numpy ceiling in the manifest; it
  applied to contributor environments only, capped a major that does not
  exist, and the runtime line deliberately carries no ceiling, so it
  constrained nothing.
- `pyproject.toml:50` — removed `"pytest-cov"` from `dev`. Evidence: CI's
  `coverage.yml` and the `AGENTS.md`-documented workflow both run
  `coverage run -m pytest ...` directly, and `grep -rn "pytest-cov"
  .github/` returns nothing — no workflow or config invokes
  `pytest --cov`.
- `pyproject.toml:53` — removed `"ruff"` from `dev`. Evidence: `ruff` is
  pinned and installed by the `astral-sh/ruff-pre-commit` hook in
  `.pre-commit-config.yaml` (currently `rev: v0.16.4`); the dev-extra copy
  was an unpinned, redundant second source of the same tool.
- `pyproject.toml:54` — expanded the `"setuptools"` comment from `# For
  torch.compile` to `# torch.compile / triton on py3.11 + torch 2.5.1,
  which does not declare it there`.
- `pixi.toml:79` — dropped the nonexistent `--extra x` flag from the CUDA
  `install-docs` task (`uv sync --extra dev --extra docs --extra x → uv
  sync --extra dev --extra docs`). `pyproject.toml` has no `x` extra
  group, and on the `uv` the repo's tasks resolve to today (0.9.28)
  `uv export --extra x` fails with `Extra 'x' is not defined in the
  project's optional-dependencies table`, so the CUDA `install-docs` task
  could never have run.
- `.pre-commit-config.yaml` — removed the `requirements-txt-fixer` hook.
  Evidence: `find . -iname "requirements*.txt"` (repo-wide) finds nothing;
  the hook had no file to ever act on.
- `tests/enhance/test_histogram.py` — removed `from packaging import
[...]

Issue #4259, opened 2026-09-05
Dependency audit: `packaging` is dead at runtime, 3 unused dev entries, diffusers/transformers cost 28 packages for one slow test, 3 unused docs entries


ducha-aiki:
Audit of every declared dependency against what the code actually imports, on `main` @ 92e5fe8d. Method: AST walk of all `import`s in `kornia/`, `tests/`, `testing/`, `docs/`, `benchmarks/`; the `LazyLoader` registry in `kornia/core/external.py`; `git log -S` on every manifest line; transitive closure computed from `uv.lock`; `python -X importtime`.

`pyproject.toml` is the only manifest (no requirements files, no conda recipe). Sizes from the lock (Linux resolution):

| surface | direct | closure |
|---|---|---|
| runtime `dependencies` | 4 | 23 (19 of them torch's own CUDA/triton/sympy tail) |
| `[dev]` | 15 | +51 |
| `[docs]` | 17 | +55 |

## Runtime

- **`torch`, `kornia_rs`**: required. `kornia_rs` is used by exactly one file, `kornia/io/io.py`, imported eagerly by `kornia/__init__.py`.
- **`numpy>=1.21`**: declared in #3726 because torch does not declare numpy (verified: torch 2.14.0 has no numpy dependency in the lock) and `kornia.io` needs it (kornia_rs is numpy-in / numpy-out; `image_to_tensor`, `tensor_to_image`, `Image.from_numpy`). Legitimate; kornia's line is the only reason numpy is installed at all.
- **`packaging`: dead.** Nothing under `kornia/` imports it; `kornia/core/_compat.py` parses `torch.__version__` by hand on purpose (dynamo constant-folds it, `packaging.version` does not). The only users in the repo are `tests/enhance/test_histogram.py:20,58` (a `skipif torch < 1.9`, dead since the floor moved to 2.5.1 in #4197) and `.github/scripts/check_torch_env.py` (a CI script that has packaging through pytest/onnxruntime anyway). It has been carried since #2232 (2023) through two manifest migrations (#3307, #3396).

## `[dev]`

| entry | consumer | verdict |
|---|---|---|
| `pytest`, `coverage`, `pytest-timeout` | CI (`--timeout=30` in `.github/actions/coverage`) | keep |
| `pytest-cov` | none: the coverage action runs `coverage run -m pytest` directly. pytest-cov 7 + numpy 2.4 also breaks conftest import locally | drop |
[...]

PR #4301, opened 2026-09-06
build(deps): declare optional extras `onnx` and `sd`; remove BoxMotTracker; import-free SegmentationModelsBuilder

### LINKED ISSUE #4310: SegmentationModelsBuilder default activation="softmax" maps to nn.Softmax() with no dim (implicit-dimension UserWarning on every call)
## What happens

`kornia/models/segmentation/segmentation_models.py` (`SegmentationModelsBuilder.build`) defaults to `activation="softmax"` and passes it through to `segmentation_models_pytorch`, which maps that string to `nn.Softmax()` with no `dim`. PyTorch's implicit-dimension heuristic resolves it to `dim=1` for 4-D output, which is the intended per-pixel class softmax, but every forward emits

```
UserWarning: Implicit dimension choice for softmax has been deprecated. Change the call to include dim=X as an argument.
```

so any user of `kornia[segmentation]` (extra added in #4301) gets a deprecation warning on every call, and the behaviour depends on a heuristic PyTorch has said it will remove.

Nothing in CI sees it today: `segmentation_models_pytorch` is installed nowhere in the matrix, and `tests/models/test_segmentation_models.py` is `importorskip`-guarded.

## Fix

Pass `activation="softmax2d"` (smp's name for `nn.Softmax(dim=1)`) as the default, or build the head with an explicit `nn.Softmax(dim=1)` after `smp.create_model(..., activation=None)`. Add a `pytest.warns`-free assertion to the existing smoke test so the warning cannot come back once the extra is installed somewhere.

Found during the review of #4301.

### LINKED ISSUE #4320: BoxMotTracker targets the boxmot 10.x API; no boxmot installable with torch>=2.5.1 exposes DeepOCSORT(model_weights=...)
## Summary

`kornia.contrib.boxmot_tracker.BoxMotTracker` is written against the **boxmot 10.x** API and cannot be
constructed with any `boxmot` release that installs next to kornia's `torch>=2.5.1` floor. Its default
`tracker_model_name="DeepOCSORT"` names a class that only boxmot 10.x exported, and the constructor
kwargs it passes (`model_weights=`, `device=`, `fp16=`) are 10.x's signature.

Found while reviewing #4301, which declares `tracking = ["boxmot"]` and resolves it to boxmot 23.0.0 in
`uv.lock`; the `pip install "kornia[tracking]"` path that PR documents therefore installs a boxmot the
wrapper cannot use.

## Evidence (wheel inspection, no install)

`kornia/contrib/boxmot_tracker.py` (main, `94882c52` is the last touch):

```python
self.tracker = getattr(boxmot, tracker_model_name)(      # default "DeepOCSORT"
    model_weights=Path(...), device=device, fp16=fp16, **kwargs,
)
```

Per-version `boxmot/__init__.py` exports and `DeepOcSort.__init__` signature, read from the PyPI wheels:

| boxmot | top-level export | constructor | installable with torch>=2.5.1? |
| --- | --- | --- | --- |
| 10.0.83 (2024-09) | `DeepOCSORT` | `(model_weights: Path, device, fp16, per_class=False, ...)` | no: pins `numpy==1.26.4`, `torchvision>=0.17.1,<0.18` (torch 2.2) |
| 11.0.9 – 17.0.0 | `DeepOcSort` (renamed) | `(reid_weights: Path, device, half, ...)` | yes (11–12 still pin `numpy==1.26.4`) |
| 22.0.0 – 23.0.0 | only `BoxMOT`, `Detector`, `ReIDModel`; `__getattr__` raises `AttributeError` for anything else | `DeepOcSort(reid_model=None, ..., **BaseTracker kwargs)` | yes |

So `BoxMotTracker()` on any installable boxmot raises
`AttributeError: module 'boxmot' has no attribute 'DeepOCSORT'`; passing
`tracker_model_name="DeepOcSort"` on 11–17 then fails with `unexpected keyword argument 'model_weights'`
(the kwargs became `reid_weights=` / `half=`).

Nothing in `tests/` exercised the wrapper against a real boxmot until #4301's
`tests/contrib/test_boxmot_tracker.py::test_lazy_loader_resolves_tracker_class`, which asserts
`boxmot.DeepOCSORT` is a class and so **fails (does not skip)** as soon as somebody installs the
extra — that failure is this issue.

## What needs deciding

1. Rewrite `BoxMotTracker` against the current boxmot API (`boxmot.BoxMOT` / `Detector` / `ReIDModel`
   pipeline, `DeepOcSort(reid_model=...)`), and add a real smoke test that constructs the tracker with a
   dummy `ObjectDetector` so the constructor kwargs are pinned, not just the class lookup.
2. Until then, either drop the `tracking` extra from #4301 or say on the installation page that the
   wrapper does not work with current boxmot.
3. Whichever way, `docs/source/models.rst` shows `tracker_model_name="DeepOCSORT"` as the example value.

Refs #4261, #4301.

ducha-aiki:
## What changed

Two features wrap third-party packages that kornia imports lazily but never declared as
dependencies: `kornia.onnx` / ONNX export / `OnnxLightGlue` and `StableDiffusionDissolving`. Users
had no supported way to install them, and the `dev` extra carried `diffusers` and `transformers`
for every contributor. Two more wrappers, `BoxMotTracker` and `SegmentationModelsBuilder`, were
first declared as `tracking` / `segmentation` extras in this PR; the review found that neither could
work as written (#4320, #4310), and the maintainer's decision was to remove the tracker and make the
segmentation builder import-free instead of declaring extras for them.

**`pyproject.toml`** — two new extras:

```toml
onnx = ["onnx", "onnxruntime", "onnxscript"]
sd = ["diffusers", "transformers"]
```

`dev` now lists `"kornia[onnx]"` instead of the three onnx packages, and no longer lists
`diffusers` or `transformers` (the dependabot-managed `transformers<5.11` pin goes away with it).
`docs` is untouched.

**`kornia/core/external.py`** — `LazyLoader.__init__` takes `extra: Optional[str] = None`. When set,
all three "Optional dependency 'X' is not installed" messages (the `ASK` prompt, the `ASK`-declined
`ImportError`, the `RAISE` `ImportError`) end with `Install it with: pip install "kornia[<extra>]".`
instead of the generic "Please install it to use this functionality."; loaders without an extra keep
the old text. `extra=` is set on `onnx`/`onnxruntime` (`onnx`) and `diffusers` (`sd`).
`InstallationMode` behaviour is unchanged. The unused `transformers = LazyLoader("transformers")` is
deleted (nothing in `kornia/` imported it), and so are the `boxmot` and `segmentation_models_pytorch`
handles, whose only consumers are gone (below).

**`OnnxLightGlue` without `onnxruntime`** raises an `ImportError` that names `pip install "kornia[onnx]"`, like the
`LazyLoader` handles, instead of a bare `KORNIA_CHECK` `BaseError`; the installation page lists the class under the
`onnx` extra, and its prose now describes the default `ask`/`auto`/`raise` `installation_mode` behaviour rather than
promising an `ImportError` unconditionally.

**`BoxMotTracker` removed (breaking).** `kornia/contrib/boxmot_tracker.py` is deleted, with its
`kornia.contrib` export, `tests/contrib/test_boxmot_tracker.py`, its `docs/source/models.rst`
section, its `autodoc_mock_imports` entry and its export-support survey row. The class called the
boxmot 10.x API (`DeepOCSORT(model_weights=..., device=..., fp16=...)`); no boxmot release that
installs next to `torch>=2.5.1` provides it, so it could not be constructed on any supported stack
(#4320). Users track with boxmot directly on kornia detector output. `kornia.contrib` is the
experimental tier, and the removal is acknowledged in `tests/api_surface_removals.json` for both
`kornia.contrib` and `kornia.contrib.boxmot_tracker`; `.github/scripts/check_import_surface.py
--base-ref origin/main` exits 0.

**`SegmentationModelsBuilder` import-free (breaking).** `build()` used to take
`model_name`/`encoder_name`/`encoder_weights`/`in_channels`/`classes`/`activation`/`**kwargs`, import
smp lazily, instantiate the architecture and look up the encoder's preprocessing itself. It is now

```python
SegmentationModelsBuilder.build(model: nn.Module, preproc_params: dict | None = None, name: str = "segmentation_model") -> SemanticSegmentation
```

where `preproc_params` is the dict `smp.encoders.get_preprocessing_params(encoder_name)` returns
(`input_space`, `input_range`, `mean`, `std`); `None` means identity preprocessing. Kornia no longer
imports smp anywhere. `get_preprocessing_pipeline` keeps its semantics (BgrToRgb, range rescale, mean/std
`Normalize`) and adds a `KORNIA_CHECK` naming a missing key. The `[0, 255]` rescale is now
`kornia.enhance.Rescale(255.0)`, a multiply by a constant every dtype stores exactly, instead of
`Normalize(std=1/255)`, a division by a stored reciprocal that bfloat16 rounds to a ~254.0 multiplier
(0.5 -> 127.0 instead of 127.5); float32 outputs of that step move by at most one ulp.
`SemanticSegmentation` gains the `__init__(model, pre_processor, post_processor, name=None)` and
placeholder `from_config` its siblings (`ObjectDetector`, `EdgeDetector`) have: it was abstract, so the
old `build()` could never have returned one even with smp installed (the export-support survey had
recorded that as a kornia bug). Its `visualize_output` now moves the CPU-drawn colormap to the mask's
device *and dtype* before the gather; indexing the CPU colormap with a CUDA or MPS argmax raised, which the
MPS job surfaced the moment the class became constructible, and the float32 colormap made every non-float32
model's visualization float32. Its softmax probe tolerates `C * eps(dtype)` instead of `allclose`'s float32-sized
default, which rejected every float16 and bfloat16 model; a raw-logit head is still refused, and the docs say
`visualize` needs a softmax head (the smp examples build one with `activation="softmax2d"`). `save` is not
advertised on the `SemanticSegmentation` docs entry: every container's `save` raises because
`_save_outputs` writes float images to `.png` (#4322, pre-existing). The docstring example imports smp under
`# doctest: +SKIP`;
`docs/source/models.rst` shows the new usage and now documents `SemanticSegmentation`.

**CI** — `.github/actions/coverage/action.yml` gains an optional `extra-install` input (default
`''`) and a guarded step after `Install dependencies`. `.github/workflows/coverage.yml` passes
`extra-install: sd` from the `slow` job only, which is the only job that runs
`tests/filters/test_dissolving.py`.

**Tests**

* `tests/core/test_external.py` — a `RAISE`-mode `LazyLoader("definitely_not_a_module_xyz",
  extra="sd")` raises an `ImportError` containing `kornia[sd]`; the no-extra loader keeps the
  generic message; the three declared loaders carry the right extra.
* `tests/models/test_segmentation_models.py` — rewritten as a `BaseTester` class with the
  `device`/`dtype` fixtures and no optional package: a `Conv2d(3, 2, 1) + Softmax(dim=1)` stand-in
  network; `build()` returns a `SemanticSegmentation` in eval mode with the given name, batch and
  list forward shapes/finiteness/dtype, `None` params give `nn.Identity` preprocessing, and the
  preprocessing is pinned numerically: for `{"input_space": "BGR", "input_range": [0, 255], mean,
  std}` the expected tensor is computed by hand in float64 on the CPU (channel flip, x255,
  (x-mean)/std; MPS has no float64, so the output is brought over for the comparison) and compared
  within `8 * max(eps(dtype), eps(float32))`: the pipeline's mean/std/1-over-255 constants are float32
  tensors, so a float64 input is only float32-accurate, and the three chained ops accumulate ~2 eps
  in the half dtypes against a default rtol of ~1 eps; missing keys raise with the `KORNIA_CHECK`
  message pinned in full; unsupported `input_space` and `input_range` raise; `from_config` raises
  `NotImplementedError`; `visualize` runs on the batch and the per-image list path and its result
  stays on the input's device (float32 only, its softmax detection is `torch.allclose` at default
  tolerance).
* `tests/filters/test_dissolving.py` — module-level `pytest.importorskip("diffusers")`, and an autouse
  fixture doing the same on `tests/augmentation/test_augmentation.py::TestRandomDissolving`, so a local
  `--runslow` run skips rather than hitting the `LazyLoader` prompt now that `diffusers` is out of `dev`.
* `tests/models/test_segmentation_models.py` also pins `visualize` in every dtype (result dtype and device),
  that a raw-logit head is still refused, and an ONNX round-trip of the BGR/255/ImageNet preprocessing
  pipeline on the CPU (`dynamo=True`, opset 18, onnxruntime vs eager). `tests/feature/test_lightglue_onnx.py`
  pins the `kornia[onnx]` hint with `onnxruntime` hidden.

**Docs** — an "Optional extras" `list-table` on `docs/source/get-started/installation.rst` (`onnx`,
`sd`, `dev`, `docs`), plus one sentence that `pillow` is needed for `output_type="pil"`. One pointer
line in the README's installation section, and a parenthetical on the multi-framework page. The
`api.rst` / `introduction.rst` one-liners for `kornia.models` no longer promise tracking builders.

**`uv.lock`** regenerated from `main`'s lock with `uv lock --no-upgrade` for the remaining extras.

**Rebase** — `origin/main` is merged in twice: `763462e4` (after #4287–#4296 and #4304 landed) as
`d45761f6`, and `3ea797d4` (after #4302 and #4306 landed) as `703a1a5b`. Conflicts both times were
`pyproject.toml` (main's `dev`/`docs` lists plus this PR's `kornia[onnx]` and the two new extras; clean in
the second merge), the `kornia/core/external.py` registry tail (main's, without `basicsr` and, since #4302,
without `requests`, plus the three `extra=` hints) and `uv.lock` (main's lock relocked with `uv lock
--no-upgrade`; the second relock is 119 packages, and its diff against main is only the extra markers and
`provides-extras`, with no package added, removed or moved). `docs` no longer lists `onnxruntime` since
#4304, so the earlier nit about `docs` naming the bare onnx packages is moot.

**Export-support survey** — the `contrib.BoxMotTracker` row is dropped; the two
`models.segmentation` cases in `cases_models.py` use the new API and their snapshot rows were
re-measured with the harness (`harness.run_case` + `probe_compile.run_case`, normalized with
`merge.py`'s helpers) on torch 2.14.0; their `note` says so, since the rest of the snapshot is
torch 2.9.1. Result: ONNX ok, `torch.export` ok, compile ok with 2 graph breaks for the wrapped
stand-in; all ok for the identity container.

## Why

`Refs #4259` step C, `Refs #4261`. Closes #4320 (the class is removed rather than rewritten).
Closes #4310 (kornia no longer passes any `activation` to smp; the user builds the head).

## What was verified

* `uv lock --no-upgrade` from `main`'s lock → resolved 126 packages (was 144 with boxmot/smp);
  `uv lock --check` clean. `git diff --stat origin/main -- uv.lock` → `40 insertions(+), 30
  deletions(-)` with no package added or removed; `rich` stays at 14.3.3. Versus the previous head,
  `4 insertions(+), 513 deletions(-)`.
* `get_preprocessing_pipeline` witness: for the `[0, 1]` dicts (RGB/[0,1]/0.5, RGB/[0,1]/None, BGR/[0,1]/None)
  the outputs are `torch.equal` to `main`'s implementation with the same module sequence; the `[0, 255]`
  path differs by design since `62f54cd2` (exact in every dtype; at most one float32 ulp from the old
  division, 76 % of elements move on a random input).
* `pytest tests/models/test_segmentation_models.py -q --dtype=float32,float64,float16,bfloat16` →
  `48 passed` (with `tests/feature/test_lightglue_onnx.py`); `--device=mps --dtype=float32` with
  `PYTORCH_ENABLE_MPS_FALLBACK` unset → `18 passed, 1 skipped` (the off-CPU export). Against the previous
  `visualize_output`, the dtype-parametrized `test_visualize` fails at float64, float16 and bfloat16.
  The new `test_preprocessing_255_rescale_is_exact` pins `0/0.5/1 -> 0/127.5/255` with `torch.equal` in
  every dtype and fails on the old `Normalize(std=1/255)` at float32, float64 and bfloat16.
* Export survey re-run for the two `models.segmentation` cases on `62f54cd2` (harness + `probe_compile`,
  from the worktree root): ONNX ok, `torch.export` ok, compile ok with 2 graph breaks for the builder case,
  all ok for the container — the snapshot rows are unchanged.
* `pytest tests/test_api_surface.py tests/test_import.py tests/core/test_external.py tests/models
  tests/contrib tests/filters/test_dissolving.py -q` → `378 passed, 113 skipped`.
* `pytest --doctest-modules kornia/models/segmentation/` → `1 passed` (the smp example is skipped).
* `python -c "import kornia, kornia.contrib, kornia.models.segmentation"` in a subprocess → neither
  `boxmot` nor `segmentation_models_pytorch` in `sys.modules`; `kornia.contrib.BoxMotTracker` and
  the two `kornia.core.external` handles are gone.
* `grep -rni "boxmot" kornia/ tests/ docs/ README.md pyproject.toml pixi.toml .github/ benchmarks/`
  → only the two rows in `tests/api_surface_removals.json`.
* `.github/scripts/check_import_surface.py --base-ref origin/main` → exit 0, notices only.
* `uvx ruff@0.16.4 check` / `format --diff` on the touched files → clean; `pre-commit run
  --all-files` → all hooks pass; `uvx ty@0.0.9 check` on the four touched library files → `All
  checks passed!`.

## What is not covered

* The CI YAML change cannot be executed locally; it was validated by parsing and by review only.
  `coverage.yml` is `workflow_call`-only and is invoked solely from `scheduled_test_cpu.yml`, whose
  triggers are `push` on `main`, `schedule`, `workflow_dispatch` and `workflow_call` — there is no
  `pull_request` trigger, so the new `extra-install` step does **not** run on this PR. It first
  executes post-merge on `main` or on the nightly schedule.
* The conditional step runs `uv sync --extra dev --extra <extra>` rather than `uv pip install
  ".[<extra>]"`: `uv pip install .` would replace the editable install with a built wheel and ignores
  `UV_PROJECT_ENVIRONMENT` on the `py312`/`py313` environments; `uv sync` is also what every `pixi`
  install task in this repo uses.
* No test runs an actual smp network through the builder (smp is in no environment by design); the
  stand-in covers the container and the preprocessing, and the docs example is the smp path.
* Docs build not run locally (no docs deps installed); the new `autoclass` for `SemanticSegmentation`
  and the `:class:` cross-reference to it were checked by reading only.
* The preprocessing pipeline exports from the CPU only: `Normalize` (and `Rescale`) keep their constants as
  plain tensor attributes that `.to(device)` does not move, so `torch.export` of an accelerator-resident
  pipeline fails on a device mismatch (#4323, pre-existing; the new export test skips off-CPU for it).
* The `mean`/`std` constants stay float32 tensors, so a float64 model gets float32-accurate mean/std
  normalization (pre-existing). Storing them as float64 would fix that in eager but change the ONNX
  graph's output dtype, because the export path uses the tensors as-is; left alone.


ducha-aiki:
Pre-merge check for the new `extra-install` step: `coverage.yml` is `workflow_call`-only and its caller `scheduled_test_cpu.yml` has no `pull_request` trigger, so the step does not run on this PR. I dispatched it manually on this branch instead: https://github.com/kornia/kornia/actions/runs/34005644407. The `slow` job there is the one that installs `kornia[sd]` and runs `tests/filters/test_dissolving.py`; the thing to check in its log is that the dissolving tests *run* rather than skip (the new module-level `importorskip("diffusers")` would turn a failed extra install into a silent skip).

ducha-aiki:
**Changes requested** — the `onnx`, `sd` and `segmentation` extras, the `LazyLoader` hint, the CI plumbing and the lock all check out and I pushed the two docs fixes; what remains is that the `tracking` extra installs a `boxmot` the wrapper cannot use, so the installation page's "enables `BoxMotTracker`" is false today (kornia#4320) and the maintainer has to decide what to do with that row.

### Verified

Head `ddf7f257` matched `headRefOid`; reviewed the full diff, then pushed `63f1ced5` on top.

```
uv 0.9.28: uv lock --check            -> Resolved 144 packages, rc=0 (cache 9.8G before and after)
lock diff: added boxmot 23.0.0, segmentation-models-pytorch 0.5.0 + their deps; rich 14.3.3 -> 15.0.0
           (boxmot 23 requires rich>=15); no other version moved, no entry removed
lock: kornia.dev = coverage numpy onnx onnxruntime onnxscript pillow ... (self-ref kornia[onnx] resolved)
pytest tests/core/test_external.py tests/contrib/test_boxmot_tracker.py tests/models/test_segmentation_models.py -rs
           -> 8 passed, 2 skipped, both reasons name the extra (boxmot, smp not installed here)
pytest tests/filters/test_dissolving.py --collect-only -> 5 collected; tests/onnx --collect-only -> 35
pytest tests/test_api_surface.py -> 13 passed
discrimination: main's kornia/core/external.py + new tests/core/test_external.py -> 6 failed, 1 passed
ruff 0.14.10 (venv) and uvx ruff@0.16.4: check clean, 5 files already formatted
grep "external.transformers|transformers = LazyLoader" kornia/ tests/ -> nothing; only diffusers.py uses `diffusers`
```

The dispatched run 34005644407 (head `ddf7f257`) concluded `success`. Job `coverage / slow (float32)` log:
step `Install optional extra sd` ran `pixi run -e py311 uv sync --extra dev --extra sd`, installed 21 packages
including `diffusers==0.39.0` and `transformers==5.10.1`, and `tests/filters/test_dissolving.py .....` — five
passes, not skips. Job `coverage / overall (float32)` installed `onnx==1.22.0 onnxruntime==1.29.0 onnxscript==0.7.1`
through `dev -> kornia[onnx]` and ran `tests/onnx/*` (all dots), so moving the onnx packages behind the extra did
not turn `tests/onnx` into skips in the default legs.

`sd = ["diffusers", "transformers"]` is the right pair: `StableDiffusionPipeline.from_pretrained` needs
`transformers` for the CLIP text encoder although `diffusers` does not declare it. `segmentation_models_pytorch`
0.5.0 (wheel inspected, not installed) exports `Unet` and `encoders.get_preprocessing_params` and `Unet` takes
`activation=`, so the builder's calls are valid at the locked version.

### Findings

1. `pyproject.toml:75` `tracking = ["boxmot"]` + `docs/source/get-started/installation.rst` row "``tracking`` —
   enables `BoxMotTracker`". `kornia/contrib/boxmot_tracker.py:104` calls
   `getattr(boxmot, "DeepOCSORT")(model_weights=Path(...), device=..., fp16=...)`, which is the **boxmot 10.x**
   API. From the PyPI wheels (no install): 10.0.83 exports `DeepOCSORT(model_weights, device, fp16, ...)` but pins
   `numpy==1.26.4` and `torchvision<0.18` (torch 2.2), so it cannot coexist with kornia's `torch>=2.5.1`;
   11.0.9–17.0.0 rename it to `DeepOcSort(reid_weights, device, half, ...)`; 22.0.0 and 23.0.0 (what the lock
   resolves) export only `BoxMOT`, `Detector`, `ReIDModel` and their `__getattr__` raises `AttributeError` for
   anything else. So `pip install "kornia[tracking]"` gives a `BoxMotTracker()` that raises
   `AttributeError: module 'boxmot' has no attribute 'DeepOCSORT'`. The new smoke test
   `test_lazy_loader_resolves_tracker_class` asserts `boxmot.DeepOCSORT` is a class, so it will **fail, not skip**,
   for the first person who installs the extra — good as a pin, but the docs row promises something that does not
   work. Filed as kornia#4320 with the per-version table. Left open: either drop `tracking` from this PR until the
   wrapper is rewritten, or keep the extra and say on the row that the wrapper is broken against current boxmot
   (pointing at #4320). I did not want to make that call for you. The wrapper defect itself predates this PR.
2. `docs/source/models.rst:81` and `:147` still told users `pip install segmentation_models_pytorch` and
   `pip install boxmot`, two lines below the classes whose install path this PR moved to the extras. Fixed in
   `63f1ced5` (both now `pip install "kornia[...]"`).
3. `docs/source/get-started/installation.rst:82` — Copilot's item, see below. Fixed in `63f1ced5`: `dev` and
   `docs` are separate rows, `docs` says "on top of ``dev``" with `pip install -e ".[dev,docs]"`, matching the
   `pixi install-docs` task (`uv sync --extra dev --extra docs`).
4. Nit, not changed: `docs` still lists bare `onnx`, `onnxruntime` where `dev` now says `kornia[onnx]`. Harmless
   because every docs install also pulls `dev`, and the PR body says `docs` is deliberately untouched.
5. Question, not changed: the `sd` extra drops the `transformers<5.11` cap. That cap was the dependabot pin
   from #4161 rather than a compatibility bound (the older `<4.57` from #3332 was), and the run above passed
   with 5.10.1, so dropping it looks right, but nothing here exercised `transformers>=5.11`.

PR-body correction: "the wrappers cannot rot silently once somebody installs the extra" — for `tracking` it has
already rotted; the test will report that as a failure (see 1).

### Copilot review

- `installation.rst:82`, `dev`, `docs` row lists both environments but installs only `.[dev]` — **VALID**: the
  row's command does not install `docs`. Fixed in `63f1ced5` by splitting the row (finding 3).

### Not covered

- No `boxmot` or `segmentation_models_pytorch` install (boxmot pulls torchvision, pandas, scikit-learn, timm on a
  disk with 4.9 GB free); the two extras' smoke tests skipped here. The boxmot API facts come from reading the
  wheels, not from running the wrapper.
- No `sphinx -W` build: `docs/source/conf.py` runs `generate_examples.main()` unconditionally (four pretrained
  models) with no env var to skip it. The two changed `.rst` files went through pre-commit only; targets
  (`docs/source/onnx.rst`, `OnnxLightGlue`, the two `autoclass` entries in `models.rst`) exist.
- `test_dissolving.py` downloads `runwayml/stable-diffusion-v1-5` into `weights/`; the dispatched slow job passed,
  I did not re-run it locally.
- `uv lock --check` was run, not a fresh `uv lock --no-upgrade`.

ducha-aiki:
**Reworked after review; new head `064b334f`.** Per the maintainer's decision, the two wrappers this PR first declared extras for are handled differently:

* **`BoxMotTracker` is removed**: module, `kornia.contrib` export, test, docs section, survey row, `autodoc_mock_imports` entry and the `tracking` extra. It called the boxmot 10.x API, which no boxmot installable next to `torch>=2.5.1` provides (#4320), so it never worked on a supported stack. Breaking change in the CHANGELOG; acknowledged in `tests/api_surface_removals.json`.
* **`SegmentationModelsBuilder` no longer imports smp.** New signature: `build(model: nn.Module, preproc_params: dict | None = None, name: str = "segmentation_model") -> SemanticSegmentation`, with `preproc_params` as returned by `smp.encoders.get_preprocessing_params(...)`. `get_preprocessing_pipeline` is byte-identical (`torch.equal` witness on four dicts). `SemanticSegmentation` gains the `__init__`/`from_config` its siblings have; it was abstract, so the old `build()` could never have returned one. The `segmentation` extra is gone. The test is a `BaseTester` class on a `Conv2d + Softmax` stand-in with the preprocessing pinned by hand, green at float16/bfloat16 too.
* `uv.lock` regenerated from `main`'s: boxmot, smp and their closure gone, `rich` back at 14.3.3.

PR body updated; `Closes #4320` and `Closes #4310`.

ducha-aiki:
**Approve after fixes pushed** — delta re-review of `63f1ced5..064b334f` (BoxMotTracker removal, import-free `SegmentationModelsBuilder`); this supersedes the earlier "Changes requested" (the dead `tracking` extra is gone), and the one real defect left, a float64-only test failure that had every float64 CI leg red, is fixed in `a238faf1`.

### Verified

Head matched `064b334f`; the whole diff against `origin/main` was read in the context of the new commit.

- Removal is complete. `grep -rniE "boxmot|BoxMot" kornia/ tests/ docs/ README.md pyproject.toml pixi.toml .github/ benchmarks/ uv.lock CHANGELOG.md` hits only `tests/api_surface_removals.json` and the CHANGELOG bullet; `segmentation_models_pytorch|\bsmp\b` hits only the docstring example, `models.rst`/`base.py` prose telling the user to build the smp network, two test comments and the CHANGELOG. `kornia.contrib` is not an inventory key in `tests/api_surface.json`, so the removals file is the right mechanism; `.github/scripts/check_import_surface.py --base-ref origin/main` exits 0 and reports both rows as acknowledged. `kornia.core.external` has no `boxmot`/`segmentation_models_pytorch` handle; neither package lands in `sys.modules` after `import kornia, kornia.contrib, kornia.models.segmentation`.
- `get_preprocessing_pipeline` witness, built independently: `origin/main`'s function (exec'd from `git show` with the smp import stubbed) vs the new one on 5 dicts (BGR/255/imagenet, RGB/1/0.5, RGB/1/None, BGR/1/None, RGB/255/mean-only) x float32/float64/float16/bfloat16 — `torch.equal` on every pair, identical module sequence and `repr`. Only difference: a missing key now raises kornia's `BaseError` naming the key instead of a bare `KeyError`.
- `SemanticSegmentation.__init__`/`from_config` match `EdgeDetector` line for line (`name: Optional[str]`, `model.eval()`, `NotImplementedError` pointing at the builder); no other subclass exists to affect. Device/dtype: `Normalize` keeps its constants as plain float32 attributes (not buffers), and `kornia.enhance.normalize` moves them with `torch.as_tensor(..., device=data.device, dtype=data.dtype)` per call, so the pipeline follows the input on any device (reasoned from the code; CPU only here).
- Tests: `test_segmentation_models.py` 16 passed default, 22 passed `--dtype=float16,bfloat16`; `tests/test_api_surface.py tests/test_import.py tests/core/test_external.py tests/models tests/contrib tests/filters/test_dissolving.py` → 378 passed, 113 skipped; `--doctest-modules kornia/models/segmentation/` → 1 passed (smp lines skipped). Perturbations: dropping the BGR flip → 3 failed; dropping the `/255` rescale → 3 failed; dropping the `KORNIA_CHECK` → **0 failed** before my fix, 5 failed after (finding 2).
- `uv lock --check` clean (uv 0.9.28; `~/.cache/uv` 9.8G before and after); vs `main`: 40+/30−, 126 packages on both sides, no package added/removed, no version change, `rich` 14.3.3, lock's kornia extras = `dev`/`docs`/`onnx`/`sd` with `sd = [diffusers, transformers]`.
- CHANGELOG: two `### Breaking changes` bullets say what used to happen and what happens now; `### Added` names only `onnx` and `sd`; `installation.rst` table = `onnx`, `sd`, `dev`, `docs` = `pyproject.toml`. `export_support.json`: the two re-measured rows carry the same 15 keys as their 1338 neighbours and `generate_export_support.py` reads only `onnx`/`export`/`compile`/`graph_breaks`; the BoxMotTracker row is gone. PR title/body, `Closes #4320`/`Closes #4310` (#4310 is moot: kornia no longer passes any `activation` to smp) and the four posted comments are accurate against the code, apart from the `1/256` claim below.
- `uvx ruff@0.16.4 check`/`format --diff` clean on the touched files; `pre-commit run --all-files` all hooks pass; `uvx ty@0.0.9 check` on the two library files clean.

### Findings

1. `tests/models/test_segmentation_models.py:99` — the by-hand pin used `8 * eps(dtype)`, but the pipeline builds its `mean`/`std`/`1/255` constants as float32 tensors, so a float64 input is only float32-accurate (measured 1.3 float32 eps). The test fails at float64, and that is the single failure (1 failed / 9260 passed) in every red `tests (…, float64, …)` job on this PR. The author ran only the default float32 and the two half dtypes. Fixed in `a238faf1`: `tol = 8 * max(eps(dtype), eps(float32))`; 34 passed at float32/float64/float16/bfloat16.
2. `tests/models/test_segmentation_models.py:120,135` — `pytest.raises(Exception, match=missing)` also matches the bare `KeyError('std')` the un-checked dict lookup raises, so the four missing-key tests plus the `build()` one did not pin the new `KORNIA_CHECK` at all (perturbation: 16 passed with the check deleted). Fixed in `a238faf1`: match the full message `preproc_params is missing the key '<key>'`.
3. Same file, comment at line 96, and the PR body twice ("bfloat16 rounds the pipeline's `1/255` reciprocal to exactly `1/256`" / "divides by `bf16(1/255) == 1/256`"): false. `torch.tensor(1/255).to(torch.bfloat16)` is `0.003936767578125`, so the rescale multiplies by ~254.0 (0.4 % low; `0.5 -> 127.0`, not `128.0`); float16 stores `0.0039215087890625` and is exact. The real reason for the custom tolerance is that the three-op chain accumulates up to ~2 eps (measured 1.9 float32 / 1.3 float16 / 1.4 bfloat16) against a half-precision default rtol of ~1 eps. Comment rewritten with the measured values in `a238faf1`; treat this as the correction to the PR body. The "multiply by 255 instead" follow-up still holds (255 is exact in bfloat16).
4. Question, not changed: the float32 constants mean the preprocessing is float32-accurate for float64 models too (pre-existing, byte-identical to `main`, and outside this PR's "no behaviour change" promise). A follow-up could keep `mean`/`std` as Python lists so `Normalize` materialises them in the input dtype.

### Copilot review

No Copilot review on the delta. Its one earlier item (`installation.rst` `dev`, `docs` row) was VALID and was fixed in `63f1ced5` last round; the row is still correct after this commit.

### Not covered

No docs build (`autoclass` for `SemanticSegmentation` and the `:class:` cross-reference read only); no CUDA/MPS run; smp is installed nowhere, so the docstring/`models.rst` example (`activation="softmax2d"`) was checked against smp's API by reading, not by running; the `coverage.yml` `extra-install` step still cannot run on a PR; PR CI on `a238faf1` had not finished when this was posted.

ducha-aiki:
**Final review: Approve after fixes pushed.** Full re-review of the whole diff at `a238faf1`, then `74c00481` pushed for the one thing that was actually wrong: the `tests-mps` job on `a238faf1` was red with two failures in the new segmentation test file, one of which is a real library defect this PR made reachable. Everything else — the `onnx` and `sd` extras, the `LazyLoader` hint, the `BoxMotTracker` removal, the import-free `SegmentationModelsBuilder`, the CI plumbing, the lock, the docs — holds up.

### What the PR does, re-read at `74c00481`

- `pyproject.toml`: extras `onnx = [onnx, onnxruntime, onnxscript]` and `sd = [diffusers, transformers]`; `dev` lists `kornia[onnx]` and drops `diffusers`/`transformers<5.11`; `docs` untouched.
- `kornia/core/external.py`: `LazyLoader(..., extra=)`; the three "not installed" messages end with `pip install "kornia[<extra>]"` when set; `onnx`/`onnxruntime` → `onnx`, `diffusers` → `sd`; the unused `transformers`, `boxmot` and `segmentation_models_pytorch` handles are deleted.
- `BoxMotTracker` removed (module, `kornia.contrib` export, test, docs section, survey row, `autodoc_mock_imports` entry), acknowledged in `tests/api_surface_removals.json` for both module paths; it called the boxmot 10.x API that no boxmot installable next to `torch>=2.5.1` provides (#4320).
- `SegmentationModelsBuilder.build(model, preproc_params=None, name=...)` wraps a user-built network; kornia imports smp nowhere; `get_preprocessing_pipeline` is unchanged apart from a `KORNIA_CHECK` naming a missing key; `SemanticSegmentation` gains the `__init__`/`from_config` its siblings have and, as of `74c00481`, a `visualize_output` that gathers the colormap on the mask's device.
- CI: `coverage` action gains `extra-install`; the `slow` job passes `sd`. Docs: "Optional extras" table (`onnx`, `sd`, `dev`, `docs` as separate rows), README pointer, `models.rst` rewritten for the new API and documenting `SemanticSegmentation`, `onnx.rst` quickstart uses the extra. Export survey: the two segmentation cases re-measured, the tracker row gone. CHANGELOG: one *Added* and three *Breaking changes* bullets, each saying what used to happen.

### Verified at `a238faf1`, re-run on `74c00481` where the fix touches it

```
pytest tests/models/test_segmentation_models.py --dtype=float32,float64,float16,bfloat16 -> 34 passed (after fix; 22+22 before)
env -u PYTORCH_ENABLE_MPS_FALLBACK pytest tests/models/test_segmentation_models.py --device=mps --dtype=float32
                                                                    -> 2 failed, 14 passed on a238faf1; 16 passed on 74c00481
pytest tests/core/test_external.py tests/test_api_surface.py tests/test_import.py tests/models tests/contrib tests/filters/test_dissolving.py
                                                                    -> 378 passed, 113 skipped; tests/core+models+contrib after fix -> 701 passed, 116 skipped
pytest tests/filters/test_dissolving.py --collect-only -> 5 collected; tests/onnx --collect-only -> 35 collected
pytest --doctest-modules kornia/models/segmentation/ kornia/core/external.py -> 1 passed (smp lines skipped)
uv 0.9.28: uv lock --check -> Resolved 126 packages, rc 0; ~/.cache/uv 9.8G unchanged
.github/scripts/check_import_surface.py --base-ref origin/main -> rc 0, both BoxMotTracker rows acknowledged, rest informational
uvx ruff@0.16.4 check / format --diff (9 files) -> clean; git add -A && pre-commit run --all-files -> 13 hooks pass, tree clean
uvx ty check kornia/core/external.py kornia/models/segmentation/ -> All checks passed
```

- Lock: no `[[package]]` block added or removed and zero version lines changed against `main`; `rich` stays at 14.3.3; kornia's recorded extras are `dev`, `docs`, `onnx = [onnx, onnxruntime, onnxscript]`, `sd = [diffusers, transformers]`.
- Removal sweep: `grep -rniE boxmot` over `kornia/ tests/ docs/ README.md pyproject.toml pixi.toml .github/ benchmarks/ uv.lock` hits only `tests/api_surface_removals.json` and the CHANGELOG; `segmentation_models_pytorch|\bsmp\b` hits only prose, the `+SKIP` docstring example, `models.rst` and two test comments; no `external.transformers` anywhere; `autodoc_mock_imports` is empty. After `import kornia, kornia.contrib, kornia.models.segmentation`, neither `boxmot` nor `segmentation_models_pytorch` nor `diffusers` is in `sys.modules` (`onnxruntime` is, because this branch predates #4295).
- `get_preprocessing_pipeline` witness, `main`'s function exec'd from `git show` with the smp import stubbed vs the head, five dicts (BGR/255/imagenet, RGB/1/0.5, RGB/1/None, BGR/1/None, RGB/255/mean-only) × float32/float64/float16/bfloat16: identical module sequences and `torch.equal` on every output; the only difference is a missing key raising `BaseError: preproc_params is missing the key '<key>'` instead of a bare `KeyError`.
- The dispatched `coverage` run 34005644407 from the earlier round still stands as the evidence that the `extra-install: sd` step installs `diffusers`/`transformers` and runs the five dissolving tests rather than skipping them; the step cannot run on a PR (`coverage.yml` is `workflow_call`-only from a workflow with no `pull_request` trigger).
- CI on `a238faf1`: 41/45 green, `tests-mps` red (the two failures below), `collector` red as its aggregate; the two `-/-` rows are the RTD and pre-commit.ci contexts that never report. `74c00481` was pushed as this review was written, so its checks are pending.

### Findings

1. **Fixed in `74c00481`.** `kornia/models/segmentation/base.py:178` — `visualize_output` indexed the CPU-drawn colormap (`torch.rand(..., generator=g_cpu)`) with the mask's argmax, so `visualize` on a CUDA or MPS output raised `RuntimeError: indices should be either on cpu or on the same device as the indexed tensor`. Reproduced locally on MPS (`test_visualize[mps]`), and CUDA has the same rule. The code is pre-existing, but until this PR `SemanticSegmentation` was abstract and could not be instantiated, so this is the first head on which the path runs. Fix: `colors.to(semantic_mask.device)` before the gather, with a comment; `test_visualize` now also asserts the result's device and covers the per-image list path (which draws a colormap per mask) against the batch result. CHANGELOG bullet extended by one clause.
2. **Fixed in `74c00481`.** `tests/models/test_segmentation_models.py:87` — `test_preprocessing_bgr_255` built its by-hand float64 reference on the test device (`torch.tensor(mean, device=device, dtype=torch.float64)`, `x.to(torch.float64)`), which MPS refuses. The reference now lives on the CPU and `out` is brought over for the comparison; the assertions on `out.shape/dtype/device` are unchanged.
3. Not changed, pre-existing and outside the "byte-identical pipeline" promise: the preprocessing constants are float32 tensors, so a float64 model gets float32-accurate preprocessing, and bfloat16 stores `1/255` as `0.0039367675` (a ~254.0 multiplier). Multiplying by 255 instead would be exact in every dtype; one-line follow-up.
4. Nit, not changed: the `docs` extra still lists bare `onnx`, `onnxruntime` where `dev` now says `kornia[onnx]`; harmless because every docs install also pulls `dev`, and the body says `docs` is deliberately untouched.

### Copilot review

Its one item (`installation.rst` row installing only `dev`) was fixed in `63f1ced5` and the split `dev`/`docs` rows are still right at this head. No Copilot review on the later commits.

### Not covered

No CUDA run; the colormap fix was verified on MPS and by reading torch's indexing rule for CUDA. No `sphinx -W` build (the `autoclass` for `SemanticSegmentation` and the `:class:` cross-reference read only; `docs/source/conf.py` runs the example generators unconditionally, which need the pretrained models). smp is installed nowhere, so the docstring/`models.rst` example was checked against smp 0.5.0's API by reading. `test_dissolving.py` downloads `runwayml/stable-diffusion-v1-5`; the dispatched slow job passed, I did not re-run it locally. PR CI on `74c00481` had not finished when this was posted.

ducha-aiki:
**Final review after the rebase: Approve after fixes pushed.** Two commits on top of `74c00481`: the merge of `origin/main` (`763462e4`) as `d45761f6`, and `62f54cd2`, which closes the one open finding from the morning's reviews (the bfloat16 `1/255` rescale). The PR is `MERGEABLE` again and the PR body is updated for both.

### Conflict resolution (`d45761f6`)

- `pyproject.toml` — main's `dev`/`docs` lists (after #4296 and #4304) with this PR's `kornia[onnx]` in `dev`, `diffusers`/`transformers<5.11` gone, and the `onnx`/`sd` extras appended. The branch-vs-merge-base and merged-vs-main diffs are the same eight lines. Since #4304 dropped `onnxruntime` from `docs`, the nit about `docs` naming the bare onnx packages is moot.
- `kornia/core/external.py` — the branch's file (with the `extra=` parameter and `_install_hint`) whose registry tail is main's without `basicsr` (#4292) plus the three hints. My first pass took main's whole file and lost the parameter; the test run caught it (`TypeError: unexpected keyword argument 'extra'`) before anything was pushed, and the amended merge is what went out. Both-way diff check clean.
- `uv.lock` — main's, relocked with `uv lock --no-upgrade` (uv 0.9.28, 121 packages). Diff vs main: `diffusers`/`transformers` leave `dev`, `onnx`/`onnxruntime`/`onnxscript`/`diffusers`/`transformers` gain their extra markers, `provides-extras` gains `onnx`, `sd`; no package block added or removed, no version moved.

### The rescale fix (`62f54cd2`)

`get_preprocessing_pipeline` used `Normalize(mean=0, std=1/255)` for `input_range: [0, 255]`, a division by a stored reciprocal. bfloat16 stores `1/255` as `0.003936767578125`, so the step multiplied by ~254.0 and mapped `0.5` to `127.0`. It is now `Denormalize(mean=torch.tensor([0.0]), std=torch.tensor([255.0]))`: a multiply by a constant every dtype holds exactly, with tensor constants because `denormalize` rejects Python floats under `torch.onnx.is_in_onnx_export()`.

- Probed before editing: `0.5 -> 127.5` and `1 -> 255` in float32/float64/float16/bfloat16 (old code: `127.49999`, `127.49999`, `127.5`, `127.0`); dynamo ONNX export + onnxruntime run of a BgrToRgb/rescale/Normalize/conv/softmax pipeline max diff 4.7e-10, output dtype float32; `torch.export` ok; `torch._dynamo.explain` 0 graph breaks for both old and new. On a random float32 input the old and new step outputs differ in 76 % of elements by at most one ulp (1.5e-5 at magnitudes 128–256), which is why the CHANGELOG bullet now says the float32 outputs of that step move by at most one ulp rather than "byte-identical".
- New pin `test_preprocessing_255_rescale_is_exact`: `torch.equal` on `0/0.5/1 -> 0/127.5/255` in every dtype; with the old line restored it fails at float32, float64 and bfloat16 (float16 stores `1/255` closely enough to round to `127.5`).
- Left as is: the `mean`/`std` constants remain float32 tensors, so a float64 model gets float32-accurate mean/std. Float64 constants would fix eager but change the ONNX graph's output dtype (the export path uses the tensors as-is), so that is a design question, not a one-liner. Body updated to say so.

### Verified at `62f54cd2`

```
pytest tests/models/test_segmentation_models.py --dtype=float32,float64,float16,bfloat16 -> 38 passed
env -u PYTORCH_ENABLE_MPS_FALLBACK pytest tests/models/test_segmentation_models.py --device=mps --dtype=float32 -> 17 passed
pytest tests/models/test_segmentation_models.py tests/core/test_external.py tests/test_api_surface.py tests/test_import.py tests/contrib
       tests/filters/test_dissolving.py tests/core/test_lazyloader.py --dtype=float32,float64,float16,bfloat16
       -> 294 passed, 75 skipped, 1 failed: tests/contrib/test_hist_match.py::test_histmatch[cpu-bfloat16], a manifest entry
          (testing/half_precision_xfails/cpu_bfloat16.txt:115) that fails identically on the primary checkout; untouched by this PR
pytest --doctest-modules kornia/models/segmentation/ kornia/core/external.py -> 1 passed
.github/scripts/check_import_surface.py --base-ref origin/main -> rc 0, both BoxMotTracker rows acknowledged
export survey, models.segmentation cases, harness + probe_compile run from the worktree root -> ONNX ok / export ok / compile ok-breaks:2 for the builder, all ok for the container: snapshot rows unchanged
uvx ruff@0.16.4 check / format --diff, uvx ty check kornia/models/segmentation/ -> clean; git add -A && pre-commit run --all-files -> all hooks pass
```

The rest of the diff (extras, `LazyLoader` hint, BoxMotTracker removal, import-free builder, CI plumbing, docs) is byte-identical to `74c00481`, so the morning's verification of it stands.

### Not covered

CI on `62f54cd2` had not started when this was posted. No CUDA run; no `sphinx -W` build; smp is installed nowhere, so the docstring example is checked by reading. The `coverage.yml` `extra-install` step still cannot run on a PR; the dispatched run from the first round stands as its evidence.

ducha-aiki:
**Re-review after #4302 and #4306 merged: Approve after fixes pushed.** Two commits on top of `62f54cd2`: `703a1a5b` merges `origin/main` (`3ea797d4`), and `9f7a23d0` closes the findings below. Head `9f7a23d0` is `MERGEABLE`; the PR body is updated for both. #4307 is re-stacked on this head (`0ad0aa11`) and stays the one that merges last.

### Conflict resolution (`703a1a5b`)

- `kornia/core/external.py` — the branch's registry tail (`onnx`/`diffusers`/`onnxruntime` with `extra=`) without `requests`, which #4302 removed; the merged-vs-main diff is exactly the `extra` parameter, `_install_hint` and the three hints.
- `uv.lock` — main's, relocked with `uv lock --no-upgrade` (uv 0.9.28, 119 packages); diff vs main is the `onnx`/`sd` extra markers and `provides-extras`, no package added, removed or moved. `pyproject.toml` merged clean.
- Every other file's branch-vs-merge-base diff equals its merged-vs-main diff (checked per file). The CHANGELOG diff vs main is this PR's bullets only.

### Findings, verified on `703a1a5b` and fixed in `9f7a23d0`

1. **`kornia/models/segmentation/base.py` — `visualize` rejected every float16/bfloat16 model.** The softmax probe was `torch.allclose(mask.sum(C), 1)` at default tolerances (rtol 1e-5); a half-precision sum of `C` probabilities is off by ~`C * eps`, so it fell into the "Only muliclass segmentation is supported" `ValueError` (20/20 draws on MPS for both half dtypes; a coin flip per draw on CPU with 3 classes). The earlier round's test dodged it with a "float32 only" comment. The tolerance now scales with `C * finfo(dtype).eps`; a raw-logit head is still refused (pinned in every dtype with a scaled 1x1 conv).
2. **Same function — the colormap followed the mask's device but not its dtype**, so a float64/float16/bfloat16 model's visualization came back float32. `colors.to(device=..., dtype=...)`; `test_visualize` is now dtype-parametrized and pins `vis.dtype`, and fails on the previous code at float64, float16 and bfloat16.
3. **`tests/augmentation/test_augmentation.py::TestRandomDissolving` had no `importorskip("diffusers")`** while `test_dissolving.py` did, so a local `--runslow` run in a fresh `dev` environment (which no longer installs `diffusers`) hit the `LazyLoader` prompt: `OSError: pytest: reading from stdin while output is captured!`. Autouse fixture added; with `diffusers` hidden the two tests report `SKIPPED ... pip install "kornia[sd]"`.
4. **`docs/source/get-started/installation.rst` promised "raises an `ImportError` naming the extra"**, but the default `installation_mode` is `ASK`: on a terminal the object offers to pip-install, with stdin closed it raises `EOFError`, and the `ImportError` only comes in `raise` mode or after declining. The prose now says that, and names the config knob.
5. **`OnnxLightGlue`, listed under the `onnx` extra on that page, pre-empted the hint** with `KORNIA_CHECK(find_spec("onnxruntime") ...)` → `BaseError: onnxruntime is not installed.` (from #4295). It now raises `ImportError('... pip install "kornia[onnx]"')`; pinned with `find_spec` monkeypatched.
6. **The smp examples built raw-logit networks** (`smp.Unet(..., classes=2)` in the docstring, no activation) while `visualize` needs probabilities, so copying the docstring and calling `visualize` raised. Both examples now use `activation="softmax2d"`, the container's `model` argument and `models.rst` say `visualize` needs a softmax head.
7. **`Denormalize(mean=tensor([0.]), std=tensor([255.]))` → `kornia.enhance.Rescale(255.0)`**: same values in every dtype (`torch.equal` on CPU/MPS x float32/float16/bfloat16 per the verifier; the exactness pin still passes), one module instead of two tensor constants, and the `Rescale` kornia's other preprocessors already use. The comment claiming the tensor constants "keep the step ONNX-exportable" was misleading: the pipeline exports from the CPU, but not from an accelerator, because `Normalize` (and `Rescale`) hold their constants as plain attributes that `.to(device)` does not move — `torch.export` fails with `FakeTensorDeviceMismatchError` on an MPS-resident pipeline. Pre-existing on `main`, filed as #4323; the new `test_preprocessing_onnx_export` (dynamo, opset 18, onnxruntime vs eager) runs on CPU and skips off-CPU citing it.
8. **`SemanticSegmentation.save` was newly advertised (`:members: forward, visualize, save`) but raises on every input**: `_save_outputs` writes the float visualization to `.png`, which `write_image` refuses (`NotImplementedError: Unsupported file extension: .png for float32 image`). Reproduced; it is `ModelBase` code shared by all four containers, so filed as #4322 and `save` dropped from the members list rather than fixed here.
9. Smaller: the `KORNIA_CHECK` in `get_preprocessing_pipeline` was documented as raising `Exception` (it is `BaseError`); the `conf.py` comment said "every LazyLoader'd package left the documented API" (diffusers/onnx are still reached lazily; the Sphinx placeholder in `external.py` is why nothing needs mocking); the CHANGELOG now has a `### Bug fixes` bullet for the visualize/OnnxLightGlue fixes instead of burying them in a Breaking-changes bullet.

Left as is, with reasons: the `from_config` stub duplicated across containers stems from `ModelBase.from_config` being `@abstractmethod` (fixing that is #4291's territory); `ImageSequential` for the three preprocessing modules keeps a CPU copy of the last output (an `ImageSequential` property shared with `_hf_models`, out of scope); `kornia[sd]` no longer bounds `transformers` (the `<5.11` was a dependabot range pin, not a known break; the lock pins 5.10.1 for CI); the export-survey `ref: ""` for the new class resolves on the next docs build.

### Verified at `9f7a23d0`

```
pytest tests/models/test_segmentation_models.py tests/feature/test_lightglue_onnx.py --dtype=float32,float64,float16,bfloat16 -> 48 passed, 8 skipped (CUDA / --runslow)
env -u PYTORCH_ENABLE_MPS_FALLBACK pytest tests/models/test_segmentation_models.py --device=mps --dtype=float32 -> 18 passed, 1 skipped (off-CPU export, #4323)
pytest tests/models tests/contrib tests/core tests/feature/test_lightglue_onnx.py tests/test_api_surface.py tests/test_import.py -> 817 passed, 125 skipped
(on 703a1a5b) pytest tests/core/test_external.py tests/models/test_segmentation_models.py tests/test_import.py tests/test_api_surface.py tests/contrib tests/io tests/onnx
       tests/filters/test_dissolving.py tests/core/test_lazyloader.py --dtype=float32,float64,float16,bfloat16
       -> 354 passed, 75 skipped, 1 failed: tests/contrib/test_hist_match.py::test_histmatch[cpu-bfloat16], the manifest entry that fails identically on main
discrimination: previous visualize_output + new tests -> test_visualize fails at float64/float16/bfloat16; diffusers hidden -> TestRandomDissolving 2 skipped
pytest --doctest-modules kornia/models/segmentation/ kornia/feature/lightglue_onnx/lightglue.py -> 1 passed (smp lines skipped)
export survey, models.segmentation cases via harness.run_case + probe_compile.run_case -> status ok / compile ok-breaks:2 (builder), ok / ok (container): snapshot rows unchanged
.github/scripts/check_import_surface.py --base-ref origin/main -> rc 0; uv lock --check -> rc 0
uvx ruff@0.16.4 check / format; uvx ty@0.0.9 check kornia/models/segmentation/ kornia/feature/lightglue_onnx/lightglue.py -> clean
git add -A && pre-commit run --all-files (after the merge commit) -> all hooks pass, tree clean
```

### Not covered

CI on `9f7a23d0` had not started when this was posted; the `703a1a5b` run had the pre-test jobs green and the CPU matrix pending. No CUDA; no `sphinx -W` build; smp is installed nowhere, so `activation="softmax2d"` was checked against smp's `Activation` by reading. The `coverage.yml` `extra-install` step still cannot run on a PR; the dispatched run from the first round stands.

PR #4253, opened 2026-09-05
docs(geometry): Convention blocks for bbox/boxes (batch 4c: masks, generators, 3D, video)

ducha-aiki:
## Summary

Documents and pins the sixteen mask, generator, 3D, and video symbols of conventions batch 4c, plus the `Boxes3D` class anchor. Callers can read the size-argument order, the three rounding policies, the inclusive extents, the 3D vertex order, the export-mode defaults, and the `VideoBoxes` storage contract directly in the API docs.

- Add the `Boxes3D` Convention block and anchor pointers on `get_boxes_shape`, `from_tensor`, `to_tensor`, `to_mask`, `transform_boxes`, `transform_boxes_`; own blocks for `Boxes.to_mask`, `VideoBoxes`, and the six `kornia.geometry.bbox` functions `validate_bbox3d`, `infer_bbox_shape3d`, `bbox_to_mask`, `bbox_to_mask3d`, `bbox_generator`, `bbox_generator3d`.
- Correct the documented facts that execution contradicted: `validate_bbox3d` checks equal edge extents, not cubes; `Boxes3D.to_mask` returns `(N, depth, height, width)`, not `(N, depth, width, height)`; `Boxes.to_mask` returns `(N, height, width)`, not `(N, width, height)`; `Boxes3D.from_tensor` validation is "positive in the mode's convention", not ">= 2 for _plus"; both `to_mask` doctest equivalence comments named the wrong mode; `get_boxes_shape` omitted depth; the generators described "the masked image". After review: `VideoBoxes` no longer lists `pad` and `to` as copying methods (they update `self` in place, with `unpad` and `type`); the `Boxes3D` anchor no longer says `validate_bbox3d` requires unbatched input; `bbox_to_mask3d` warns that a box covering or overhanging a full axis fills the whole volume (#4255).
- Add 43 convention/wart test nodes per dtype (85 per two-dtype run), all additive, including strict xfails for the 3D generator extent (#4018), the rank-4 3D helper crash, the `Boxes3D.to_tensor` default-mode split, the `VideoBoxes` inherited-method breakage, and the `bbox_to_mask3d` full-axis fill (#4255). A copy-versus-in-place pin covers the `VideoBoxes` inherited methods. The CUDA/export mask path is pinned against the loop path on every device by patching the module's `is_exporting` predicate.
- Re-executed every audit row that #4112, #4189, #4184, #4218, and #4138 changed since the up-front audit; nothing obsolete is re-documented.

[...]

Issue #4250, opened 2026-09-05
W1b: `bbox_to_mask3d` returns float32 `(B, 1, D, H, W)` while `bbox_to_mask` and `Boxes3D.to_mask` keep the input dtype with no channel axis


ducha-aiki:
The three mask builders already differ in argument order (#4014) and rounding policy (#4015). They also disagree on the output dtype and rank: `bbox_to_mask3d` ends with `.float()` and keeps a singleton channel axis from its intermediate broadcast, so it returns `float32` `(B, 1, D, H, W)` for any input dtype (float64, float16, bfloat16, int64 all executed), while `bbox_to_mask` returns `(B, H, W)` in the input dtype (int64 in gives int64 out) and `Boxes3D.to_mask` returns `(N, D, H, W)` in the box dtype.

Reproduced 2026-09-05 at `1eea5835`, PyTorch 2.14.0, CPU and MPS.

## Reproduction

```python
import torch
from kornia.geometry.bbox import bbox_to_mask, bbox_to_mask3d
from kornia.geometry.boxes import Boxes3D

cuboid = torch.tensor([[[1., 1., 1.], [3., 1., 1.], [3., 3., 1.], [1., 3., 1.],
                        [1., 1., 2.], [3., 1., 2.], [3., 3., 2.], [1., 3., 2.]]], dtype=torch.float64)
square = torch.tensor([[[1., 1.], [3., 1.], [3., 2.], [1., 2.]]], dtype=torch.float64)
for label, out in (("bbox_to_mask3d", bbox_to_mask3d(cuboid, (4, 5, 6))),
                   ("Boxes3D.to_mask", Boxes3D(cuboid).to_mask(4, 5, 6)),
                   ("bbox_to_mask", bbox_to_mask(square, 6, 5))):
    print(label, tuple(out.shape), out.dtype)
```

## Observed

```text
bbox_to_mask3d (1, 1, 4, 5, 6) torch.float32
Boxes3D.to_mask (1, 4, 5, 6) torch.float64
bbox_to_mask (1, 5, 6) torch.float64
```

## Expected

[...]

Issue #4248, opened 2026-09-05
Bug: `infer_bbox_shape3d` / `bbox_to_mask3d` accept `(B, N, 8, 3)` through `validate_bbox3d` and then crash or return wrong-shaped values


ducha-aiki:
`validate_bbox3d` documents and accepts rank-4 `(B, N, 8, 3)` input (it reshapes to `(B*N, 8, 3)` internally), but its two callers in the same module do not reshape: `infer_bbox_shape3d` runs `torch.index_select(boxes, 1, ...)` and `bbox_to_mask3d` indexes `boxes[:, 4, 2]` on the original tensor, so the box axis is read as the vertex axis. This is the 3D twin of #4180, which #4218 fixed for `infer_bbox_shape` / `bbox_to_mask` by rejecting rank 4 with `ShapeError`.

Reproduced 2026-09-05 at `1eea5835` (main after #4245), PyTorch 2.14.0, Python 3.11, CPU and MPS.

## Reproduction

Run from the checkout root:

```python
import torch
from kornia.geometry.bbox import validate_bbox3d, infer_bbox_shape3d, bbox_to_mask3d

cuboid = torch.tensor([[[1., 1., 1.], [3., 1., 1.], [3., 3., 1.], [1., 3., 1.],
                        [1., 1., 2.], [3., 1., 2.], [3., 3., 2.], [1., 3., 2.]]])
for n in (1, 3, 8):
    boxes = cuboid.expand(n, 8, 3)[None].contiguous()  # (1, n, 8, 3)
    print(n, "validate:", validate_bbox3d(boxes))
    for name, call in (("infer", lambda: infer_bbox_shape3d(boxes)), ("mask", lambda: bbox_to_mask3d(boxes, (4, 5, 5)))):
        try:
            out = call()
            print("  ", name, [tuple(t.shape) for t in out] if isinstance(out, tuple) else tuple(out.shape))
        except Exception as error:
            print("  ", name, type(error).__name__, str(error)[:70])
```

## Observed

```text
1 validate: True
   infer RuntimeError INDICES element is out of DATA bounds, id=1 axis_dim=1
   mask IndexError index 4 is out of bounds for dimension 1 with size 1
3 validate: True
   infer RuntimeError INDICES element is out of DATA bounds, id=5 axis_dim=3
   mask IndexError index 4 is out of bounds for dimension 1 with size 3
8 validate: True
   infer [(1, 3), (1, 3), (1, 3)]
[...]

Issue #4015, opened 2026-08-26
W1b: three rounding policies for fractional box coordinates — `bbox_to_mask` (inclusive on floats), `Boxes.to_mask` (round, half-open), `bbox_to_mask3d` (truncate)


ducha-aiki:
All three mask builders agree on integer vertices and disagree on fractional ones, because each has its own undocumented rounding policy: `bbox_to_mask` compares the raw floats inclusively (`xx >= x_min & xx <= x_max`), `Boxes.to_mask` exports the exclusive `xyxy` box, `.round()`s it and slices half-open, and `bbox_to_mask3d` `.long()`-truncates. Boxes come out of `transform_boxes` with fractional coordinates as a matter of course.

## Repro

```python
import torch
from kornia.geometry.bbox import bbox_to_mask, bbox_to_mask3d
from kornia.geometry.boxes import Boxes

v = torch.tensor([[[1., 1.], [4., 1.], [4., 3.], [1., 3.]]])
f = torch.tensor([[[1.4, 1.4], [3.6, 1.4], [3.6, 2.6], [1.4, 2.6]]])
print("integer box:    bbox_to_mask", bbox_to_mask(v, 6, 6).sum().item(), " Boxes.to_mask", Boxes(v).to_mask(6, 6).sum().item())
print("fractional box: bbox_to_mask", bbox_to_mask(f, 6, 6).sum().item(), " Boxes.to_mask", Boxes(f).to_mask(6, 6).sum().item())
f3 = torch.tensor([[[1.4, 1.4, 1.4], [3.6, 1.4, 1.4], [3.6, 2.6, 1.4], [1.4, 2.6, 1.4],
                    [1.4, 1.4, 2.6], [3.6, 1.4, 2.6], [3.6, 2.6, 2.6], [1.4, 2.6, 2.6]]])
print("fractional 3D:  bbox_to_mask3d", bbox_to_mask3d(f3, (6, 6, 6)).sum().item())
```

## Observed

```
integer box:    bbox_to_mask 12.0  Boxes.to_mask 12.0
fractional box: bbox_to_mask 2.0   Boxes.to_mask 12.0      <- 6x apart for the same box
fractional 3D:  bbox_to_mask3d 12.0                        <- z 1..2, y 1..2, x 1..3 after truncation
```

[...]

PR #4303, opened 2026-09-06
fix(geometry): bbox_to_mask3d intersects axis ranges for a full/overhanging-axis box

### LINKED ISSUE #4255: Bug: `bbox_to_mask3d` fills the whole volume when a box spans or overhangs a full axis
`bbox_to_mask3d` builds its mask as a union of three axis slabs and then recovers the intersection with `all` reductions. When a box covers an entire output axis, the union intermediate is `True` everywhere, every reduction comes back `True`, and the function fills the whole volume, losing the bounds on the other two axes. A box that overhangs the volume on one axis, which is the normal state of a box after a crop or a translation, triggers the same failure. `Boxes3D.to_mask` on the same box fills the intended region on both its loop and export paths.

Reproduced 2026-09-05 at `1eea5835`, PyTorch 2.14.0, CPU.

## Reproduction

```python
import torch
from kornia.geometry.bbox import bbox_to_mask3d
from kornia.geometry.boxes import Boxes3D

size = (4, 4, 5)  # depth, height, width
for label, xyzxyz in {
    "x spans 0..4 (full width)": [0., 1., 1., 4., 2., 2.],
    "y spans 0..3 (full height)": [1., 0., 1., 2., 3., 2.],
    "z spans 0..3 (full depth)": [1., 1., 0., 2., 2., 3.],
    "x overhangs -1..5": [-1., 1., 1., 5., 2., 2.],
    "interior 1..2": [1., 1., 1., 2., 2., 2.],
}.items():
    b = Boxes3D.from_tensor(torch.tensor([xyzxyz]), "xyzxyz_plus")
    print(label, int(bbox_to_mask3d(b.data, size).sum()), int(b.to_mask(*size).sum()))
```

## Observed

```text
x spans 0..4 (full width)  80 20
y spans 0..3 (full height) 80 16
z spans 0..3 (full depth)  80 16
x overhangs -1..5          80 20
interior 1..2              8 8
```

The volume has 80 voxels. Every full-axis case fills all of them; only the interior box matches the method.

## Expected

The intersection of the three slabs, as `Boxes3D.to_mask` computes it: 20, 16, 16, 20 and 8 voxels. The three-way `|` followed by `all` reductions should be a three-way `&` of the broadcast slabs, which is what the export path of `Boxes3D.to_mask` already does.

## Cause

In `kornia/geometry/bbox.py`, `bbox_to_mask3d` computes `m = z_slab | y_slab | x_slab` and then `cond1 = m.all(dim=3).all(dim=2)` and its two siblings. For a slab `X` that covers every index, `all_{z,y}(Z | Y | X) = X | Zfull | Yfull` is `True` for every `x`, and the same happens to the other two conditions, so the product is `True` on the whole grid. For interior boxes the `all` reductions recover the slabs exactly, which is why the existing interior-box tests pass.

## Classification

Broken output. A wrong mask for a valid, in-range box. Not a rounding or convention question: the inclusive-truncating policy (#4015) and the `float32` channel-axis output (#4250) are separate.

## Disposition

**Fix directly, not in the window.** The canonical result is unambiguous and `Boxes3D.to_mask`'s export path already implements it; the repair changes no output for any box that the current code handles correctly. The conventions batch 4c PR (#4253) documents the current behavior in the `bbox_to_mask3d` warning and pins it: `test_wart_bbox_to_mask3d_fills_the_whole_volume_when_a_box_spans_an_axis_<this issue>` records today's output, and a strict `xfail` on the intersection result turns into an XPASS when this is fixed.

Found by a review of #4253 during the bbox/boxes conventions audit (#4020); the audit changes no runtime behavior.

tritsystem:

Fixes #4255: `bbox_to_mask3d` fills the whole volume when a box spans or overhangs a full axis, instead of the intersection.

The old implementation built the mask as a union of three axis slabs (`|`) and tried to recover the intersection with a three-way `all()` reduction/product. That recovery breaks the moment any one slab covers a whole output axis: the union goes all-true on that axis and the reduction can no longer see the other two slabs' bounds. A box that overhangs the volume on one axis — the normal state of a box after a crop or a translation, not an edge case — triggers the same failure.

Computes the intersection directly with `&` instead, which needs no recovery step: each broadcast slab already carries its own axis's bound at every position, so the elementwise intersection is exactly the answer. `Boxes3D.to_mask`'s export path already does this; `bbox_to_mask3d` now matches it exactly.

Closes #4255.

## What did you check?

- All 4 parametrized cases from the issue (full width/height/depth, overhang) now match `Boxes3D.to_mask` exactly (20, 16, 16, 20 voxels), not 80.
- Interior boxes and the docstring's own worked example are byte-identical before and after — this only changes behavior for boxes that cover or overhang a full axis.
- Updated the two tests #4253 added for this exact issue: flipped the `strict=True` `xfail` on `test_convention_bbox_to_mask3d_intersects_the_axis_ranges_for_a_full_axis_box_4255` to a normal passing test (it would otherwise fail the suite as an unexpected XPASS), and corrected the parametrized "wart pin" test's stale `== 80.0` assertions to `== intersection` now that the behavior it was pinning is fixed.
- Removed the now-resolved `#4255` paragraph from the docstring's `.. warning::` block; the other tracked warts in that block (#4015, #4250, #4013, #4248) are unrelated and untouched.


ducha-aiki (changes requested):
Review at `c64ea775` — implementation verified; documentation follow-up requested.

The mask implementation is correct: an exhaustive sweep of 58,320 ordered integer cuboids over a (4,4,5) volume matched a direct intersection oracle at head with zero mismatches; the merge base mismatched 31,347 cases. Focused bbox_to_mask3d tests passed (14 CPU float32/float64 cases), and all current GitHub checks are green.

Two closeout items remain:

1. `kornia/geometry/boxes.py:1451-1453` still publicly warns that bbox_to_mask3d fills the whole volume for a full/overhanging axis. This is false after the patch: remove that sentence and its #4255 link. The required issue-number closeout search also finds #4255 in the new comment at `kornia/geometry/bbox.py:434`; keep the explanation, but retire the resolved-issue reference under kornia/. The test pins can retain their issue references.
2. The CHANGELOG entry cites issue #4255 but does not cite PR #4303. Please add the PR number.

No production-code change is requested; I recommend approval after these documentation edits.

ducha-aiki:
Delta re-review at `8cdd3549` — **APPROVED, merge-ready**. The delta is exactly the three asks: the `Boxes3D.to_mask` warning loses its now-false "fills the whole volume" sentence and keeps the unrelated #4015 one, the `kornia#4255` tag is gone from the production comment in `bbox.py` (the explanation stays), and the CHANGELOG entry cites `(#4255, #4303)`. `grep -rnE "#4255|issues/4255" kornia/` is empty at this head; the one `tests/` hit is your pin naming its issue, as it should. Code is unchanged from the reviewed `c64ea775`. CI 38 green with the last leg in progress at the time of writing; `MERGEABLE`.

ducha-aiki (dismissed):

PR #4305, opened 2026-09-06
fix(geometry): solve the quad-interp Hessian in float32 for float16 input

### LINKED ISSUE #4257: ConvQuadInterp3d float16 backward returns NaN on CPU (Cramer solve division)
## Summary

`ConvQuadInterp3d` / `conv_quad_interp3d` return finite coordinates and values in float16 on CPU, but the gradient with respect to the input contains NaN over part of the volume. bfloat16 and float32 are clean. Present on `main` (`1eea5835`); #4254 does not touch `kornia/geometry/subpix/` and reproduces it identically.

## Reproduction

```python
import torch
from kornia.geometry.subpix import ConvQuadInterp3d

torch.manual_seed(0)
for dt in (torch.float16, torch.bfloat16, torch.float32):
    x = torch.rand(1, 1, 5, 40, 40, dtype=dt, requires_grad=True)
    coords, vals = ConvQuadInterp3d(strict_maxima_bonus=0.0)(x)
    g = torch.autograd.grad(coords.sum() + vals.sum(), x)[0]
    print(dt, coords.isnan().any().item(), vals.isnan().any().item(), g.isnan().sum().item(), "/", g.numel())
```

torch 2.14.0, CPU, macOS arm64:

```
torch.float16 False False 614 / 8000
torch.bfloat16 False False 0 / 8000
torch.float32 False False 0 / 8000
```

`torch.autograd.detect_anomaly()` names `DivBackward0` from `_solve_cramer_sym3x3` (`kornia/geometry/subpix/spatial_soft_argmax.py`, the `/ safe_det` divisions that produce `sx`, `sy`, `ss`), reached through `conv_quad_interp3d`.

Through `ScaleSpaceDetector(minima_are_also_good=True)` with `BlobDoG` on a 96×99 float16 image, the input gradient has 2161 NaN entries of 9504. The count is identical on `main` and on #4254, on both the stacked and the separate refinement path.

## Likely mechanism, not yet verified

The determinant of the 3×3 Hessian is a product of three second derivatives and sits around 1e-6 to 1e-9 for a [0, 1] image, in or below float16's subnormal range. `solved = det.abs() > eps` admits those entries, and the backward of `num / safe_det` multiplies by `1 / safe_det**2`, which overflows to inf in float16 and becomes NaN in the reduction. The forward divides once and stays finite. The default `eps=1e-7` of `_solve_cramer_sym3x3` is below `torch.finfo(torch.float16).tiny` (6.1e-5), so the threshold admits subnormal determinants in float16. A check is to compare `det.abs()` at the NaN positions against `torch.finfo(torch.float16).tiny`.

## Notes

- `tests/feature/test_scale_space_detector.py::TestScaleSpaceDetector::test_joint_extrema_matches_separate_refinement` compares gradients in every dtype but float16 for this reason (#4254); once fixed, that guard can go.
- Whether CUDA float16 is affected is unverified (no CUDA here).

ethanstoner:
Closes #4257.

## What was wrong

`_solve_cramer_sym3x3` returned NaN gradients for `float16` input.

The Hessian determinant is a product of three second derivatives, so for a `[0, 1]` response it lands around 1e-4 and below. That still clears the `eps = 1e-7` gate, so `solved` admits it. The forward divides by it once and stays finite, which is why coordinates and values look fine. The backward of `num / safe_det` scales by `1 / safe_det**2`, and that square is not representable in `float16` (`finfo(float16).tiny` is 6.1e-5), so the gradient overflowed and reduced to NaN.

`bfloat16` keeps float32's exponent range, which is why it was never affected.

## The fix

Promote the solve to float32 for `float16` input and cast the shifts back. `bfloat16` and `float32` keep their existing path. This mirrors what #4231 already did for the half-precision meshgrid.

## Evidence

The issue's reproduction, torch 2.6.0, CPU, Windows:

| dtype | NaN in input gradient, before | after |
| --- | --- | --- |
| float16 | **514 / 8000** | **0 / 8000** |
| bfloat16 | 0 / 8000 | 0 / 8000 |
| float32 | 0 / 8000 | 0 / 8000 |

The issue reports 614/8000 on torch 2.14 / macOS arm64; the count differs with the RNG but the defect is the same.

Isolating `_solve_cramer_sym3x3` on a diagonal Hessian shows the mechanism directly, and that the `eps` gate is not the thing keeping these out:

| det | passes eps gate | float16 grad, before | float16 grad, after | float32 grad |
| 1.25e-1 | yes | -12.0 | -12.0 | -12.0 |
| 1.00e-3 | yes | -300.0 | -300.25 | -300.00006 |
| 1.25e-4 | yes | **NaN** | -1201.0 | -1200.0002 |
| 8.00e-6 | yes | **NaN** | -7496.0 | -7500.0 |
| 1.00e-6 | yes | **NaN** | -29984.0 | -30000.0 |

## Tests

`test_float16_backward_is_finite` is the pin. It fails on `main` with `assert tensor(False)` and passes with the change.

I also removed the `float16` guard in `test_joint_extrema_matches_separate_refinement`, which cited this issue. Worth being explicit: **that test passes either way on torch 2.6.0 CPU**, so its removal is a cleanup rather than a second pin. It may well have been load-bearing on the platform the issue was filed from.

Per `AGENTS.md` I ran `grep -rnE "#4257|issues/4257" kornia/ tests/`; the guard above was the only reference and there are none left.

`tests/geometry/subpix/` plus `tests/feature/test_scale_space_detector.py`, CPU:

| dtype | main | this branch |
| --- | --- | --- |
| float32 | 1 failed / 211 passed | 1 failed / 212 passed |
| float16 | 15 failed / 196 passed | 15 failed / 197 passed |
| bfloat16 | 13 failed / 198 passed | 13 failed / 199 passed |

The extra pass in each column is the new test. I captured the float16 failure lists on both sides and diffed them: identical, so no regressions. Those are the known `cpu-float16` manifest failures. `ruff check` and `ruff format --diff` are clean on the three files.

## Not verified

CUDA float16, and MPS. I have no CUDA or Apple device here, so the new test's behaviour on those backends is untested, though the change is dtype-based rather than device-based. I have not run the full suite, only the two directories above.

## AI usage

Claude Code wrote the patch and the measurements under my direction; I reviewed the diff, ran everything above myself, and can answer questions on it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -381,6 +381,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   constructor argument with the probe result (and so always equalled the argument, because SDPA is
   present on every supported torch), was renamed to `Attention.allow_flash`. (#4287)
 
+* `ConvQuadInterp3d` / `conv_quad_interp3d` no longer return NaN gradients for `float16` input. The
+  Hessian determinant `_solve_cramer_sym3x3` divides by is a product of three second derivatives, so for
+  a `[0, 1]` response it lands around 1e-4 and below and still clears the `eps` gate. The forward divides
+  by it once and stays finite, but the backward scales by `1 / det**2`, which `float16` cannot represent
+  (`finfo(float16).tiny` is 6.1e-5), so the gradient overflowed and reduced to NaN over part of the
+  volume. The solve is now promoted to `float32` for `float16` input and the shifts cast back, as #4231
+  already does for the half-precision meshgrid. `bfloat16` keeps `float32`'s exponent range and was never
+  affected; `bfloat16` and `float32` outputs are unchanged. (#4305)
+
 * `RandAugment`, `AutoAugment`, `TrivialAugment` and `AugMix` no longer silently upcast half-precision
   batches to `float32`. `OperationBase.forward` — the shared gate every auto-augment op routes through —
   moved its `batch_prob` mask to `input.device` but not `input.dtype`, and since `batch_prob` is `float32`
--- a/kornia/geometry/subpix/spatial_soft_argmax.py
+++ b/kornia/geometry/subpix/spatial_soft_argmax.py
@@ -652,6 +652,19 @@ def _solve_cramer_sym3x3(
         systems (``|det| > eps``).  Outputs for unsolved entries are numerically
         meaningless and should be discarded by the caller.
     """
+    # float16 cannot carry this solve. The determinant is a product of three
+    # second derivatives, so for a [0, 1] response it lands around 1e-4 and
+    # below — still above ``eps``, so ``solved`` admits it. The forward divides
+    # by it once and stays finite, but the backward of ``num / safe_det``
+    # scales by ``1 / safe_det**2``, and that square is not representable in
+    # float16 (finfo.tiny is 6.1e-5), so the gradient becomes inf and reduces
+    # to NaN. bfloat16 keeps float32's exponent range and is unaffected.
+    in_dtype = dxx.dtype
+    if in_dtype == torch.float16:
+        dxx, dyy, dss = dxx.float(), dyy.float(), dss.float()
+        dxy, dxs, dys = dxy.float(), dxs.float(), dys.float()
+        r0, r1, r2 = r0.float(), r1.float(), r2.float()
+
     cf00 = dyy * dss - dys * dys  # cofactor M00
     cf01 = dxy * dss - dys * dxs  # cofactor M01
     cf02 = dxy * dys - dyy * dxs  # cofactor M02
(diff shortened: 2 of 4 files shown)


ducha-aiki (dismissed):
Review at `ecdd6686`.

No actionable code findings. Promoting the Hessian solve for float16 fixes the demonstrated backward failure: the same seed-zero reproduction has 614 NaN input-gradient cells on the merge base and none at head. The constant-solution helper probe also has an exact-zero derivative after the fix.

Focused finite-backward and scale-space tests: 61 passed, 12 skipped on CPU / torch 2.14.0. Current CI is green, including half-precision torch 2.9.1 and MPS; older torch was not run locally. Source/tests merge cleanly with current main, but CHANGELOG.md conflicts. Please resolve that mechanical conflict and ensure the rebased head is green. Code approved.

ducha-aiki (approved):

Issue #4308, opened 2026-09-06
W1b: `undistort_points_kannala_brandt` — the principal point returns `nan` in float16, because both singularity guards underflow


ducha-aiki:
`undistort_points_kannala_brandt` returns `nan` for a point **at the principal point** when `params` is `float16`.

The function guards its two singular spots with hardcoded constants:

```python
th = rth.clamp(min=1e-16).sqrt()   # Newton start
...
denom = rth + eps                  # eps = 1e-8, final radial rescale
mag = radius_undistorted.abs() / denom
```

Both constants underflow to exactly `0.0` in `float16` (`1e-8` is below the smallest float16 subnormal, `~5.96e-8`; `1e-16` is far below it). The whole body runs in `params.dtype`, so with `float16` `params` the guards are no guards at all: at the principal point `rth == 0`, `denom == 0 + 0 == 0`, `radius_undistorted == tan(0) == 0`, and the result is `0.0 / 0.0 == nan` in both components.

`float32`, `float64` and `bfloat16` all return the exact origin — `bfloat16` survives because it keeps the float32 exponent range, so the underflow is specific to `float16`. The sibling `undistort_points_affine` has no such constants and returns the origin in every dtype.

## Repro

```python
import torch
from kornia.geometry.camera.distortion_affine import undistort_points_affine
from kornia.geometry.camera.distortion_kannala_brandt import undistort_points_kannala_brandt

for dtype in (torch.float64, torch.float32, torch.bfloat16, torch.float16):
    params = torch.tensor([100.0, 100.0, 4.0, 3.0, 0.1, 0.01, 0.001, 0.0001], dtype=dtype)
    principal_point = torch.tensor([4.0, 3.0], dtype=dtype)   # == (cx, cy)
    print(dtype, undistort_points_kannala_brandt(principal_point, params).tolist())

for dtype in (torch.float32, torch.float16):                  # the same on mps
    params = torch.tensor([100.0, 100.0, 4.0, 3.0, 0.1, 0.01, 0.001, 0.0001], dtype=dtype, device="mps")
    principal_point = torch.tensor([4.0, 3.0], dtype=dtype, device="mps")
    print("mps", dtype, undistort_points_kannala_brandt(principal_point, params).tolist())

# only params.dtype matters: the body casts the points to params.dtype and casts back at the end
params = torch.tensor([100.0, 100.0, 4.0, 3.0, 0.1, 0.01, 0.001, 0.0001], dtype=torch.float32)
[...]

Issue #4286, opened 2026-09-05
W1b: `distort_points` / `undistort_points` — `torch.compile(fullgraph=True)` and `torch.jit.script` failures on the camera surface


ducha-aiki:
Five compatibility gaps on the camera surface, recorded so they are on the record rather than rediscovered. This is not a `W1b` convention finding — it is filed under the compile-first policy, and nothing in batch 5 chases it.

- **`torch.compile(fullgraph=True)` fails on `distort_points` and `undistort_points`** with `Unsupported: Data-dependent branching`. The break is the `torch.any(dist[..., 12] != 0)` tilt test — a host-side read of a tensor value — which is the exact read the existing `is_exporting()` guard already works around for ONNX.
- **`torch.jit.script` fails on `depth_to_3d_v2`, `warp_frame_depth` and `PinholeCamera.project`.** The first two fail inside `F.normalize`'s overload resolution; the third fails because TorchScript infers `self` as a `Tensor`.

`project_points`, `unproject_points` and `reproject_disparity_to_3D` are fullgraph-clean and scriptable; `distort_points` and `undistort_points` script fine.

## Repro

```python
import torch
from kornia.geometry.calibration import distort_points, undistort_points
from kornia.geometry.camera import PinholeCamera
from kornia.geometry.depth import depth_to_3d_v2, warp_frame_depth

K3 = torch.tensor([[100.0, 0.0, 4.0], [0.0, 100.0, 3.0], [0.0, 0.0, 1.0]])[None]
K4 = torch.eye(4)[None]; K4[0, :3, :3] = K3[0]
pts = torch.tensor([[[1.0, 2.0]]])
dist = torch.tensor([[0.1, 0.01, 0.001, 0.0001]])

for name, fn, args in (("distort_points", distort_points, (pts, K3, dist)),
                       ("undistort_points", undistort_points, (pts, K3, dist))):
    try:
        torch.compile(fn, fullgraph=True)(*args)
        print(f"torch.compile(fullgraph=True) {name:<18}: ok")
    except Exception as e:
        print(f"torch.compile(fullgraph=True) {name:<18}: {type(e).__name__}: {str(e).splitlines()[0][:90]}")

for name, fn in (("depth_to_3d_v2", depth_to_3d_v2), ("warp_frame_depth", warp_frame_depth),
                 ("PinholeCamera.project", PinholeCamera.project)):
    try:
        torch.jit.script(fn)
        print(f"torch.jit.script              {name:<18}: ok")
[...]

Issue #4285, opened 2026-09-05
W1b: `undistort_points` — the fixed-point iteration diverges silently outside the valid radius


ducha-aiki:
`undistort_points` inverts the Brown-Conrady model with a fixed-point iteration (`num_iters=5` by default). Outside the model's radius of convergence the iteration **diverges**, silently, and more iterations make it worse: at `k1 = 0.5`, three focal lengths off axis, the round trip is off by `288` px at the default 5 iterations and by `1338` px at 200.

Near the principal point the same call is exact. Nothing in the docstring says where the boundary is, and nothing in the code detects that the iteration is running away.

A user hit this in practice in **#3273** (closed 2025-12-29): undistorting keypoints at the corners of a 3840×2160 image with a real calibration returned invalid coordinates while `undistort_image` on the same calibration looked fine. The synthetic case below is the same phenomenon in four lines.

## Repro

```python
import torch, inspect
from kornia.geometry.calibration import distort_points, undistort_points

K = torch.tensor([[100.0, 0.0, 4.0], [0.0, 100.0, 3.0], [0.0, 0.0, 1.0]])[None]  # fx = fy = 100
d = torch.tensor([[0.5, 0.0, 0.0, 0.0]])                                          # k1 = 0.5
for name, p in (("near the principal point (5, 4)", torch.tensor([[[5.0, 4.0]]])),
                ("3x focal length off axis (304, 3)", torch.tensor([[[304.0, 3.0]]]))):
    err5 = (undistort_points(distort_points(p, K, d), K, d) - p).abs().max().item()
    err200 = (undistort_points(distort_points(p, K, d), K, d, num_iters=200) - p).abs().max().item()
    print(f"{name:<36} round-trip err: {err5} (5 iters, the default)   {err200} (200 iters)")
print("num_iters default:", inspect.signature(undistort_points).parameters["num_iters"].default)
print("docstring mentions a validity domain:", "valid" in (inspect.getdoc(undistort_points) or "").lower())
```

## Observed

```
near the principal point (5, 4)      round-trip err: 0.0 (5 iters, the default)   0.0 (200 iters)
3x focal length off axis (304, 3)    round-trip err: 287.78857421875 (5 iters, the default)   1337.788330078125 (200 iters)
num_iters default: 5
docstring mentions a validity domain: False
```

## Expected

[...]

PR #4312, opened 2026-09-06
docs(geometry): Convention blocks for camera (batch 5b: distortion and calibration)

### LINKED ISSUE #4324: tilt_projection flattens every leading axis to (-1, 3, 3), so distort/undistort fail on multi-axis batches once tau != 0
## Summary

`kornia.geometry.calibration.tilt_projection` documents `taux`/`tauy` of shape `(*, 1)` and a result of shape `(*, 3, 3)`, but its body reshapes both angles to `(-1,)` and stacks the result as `(-1, 3, 3)`, so every leading axis is flattened into one. The three callers (`distort_points`, `undistort_points`, `undistort_image`) then broadcast that flattened `(N, 3, 3)` against points that still carry their `(*, N, 2)` layout, and fail as soon as the tilt branch is taken on a batch with more than one leading axis.

## Reproduction

Executed on `docs/convention-blocks-camera-c` at `1a96bfd1` (same code as `main` for these files), torch 2.14.0, cpu float32:

```python
import torch
from kornia.geometry.calibration import tilt_projection, undistort_image, distort_points

tilt_projection(torch.zeros(2, 3, 1), torch.zeros(2, 3, 1)).shape   # torch.Size([6, 3, 3]), documented (*, 3, 3) -> (2, 3, 3, 3)
tilt_projection(torch.zeros(2, 1), torch.zeros(2, 1)).shape         # torch.Size([2, 3, 3]) -- only the single-axis case survives

img = torch.rand(2, 3, 1, 5, 6)
K = torch.tensor([[3., 0, 3], [0, 2, 2], [0, 0, 1]]).expand(2, 3, 3, 3).contiguous()
d14 = torch.zeros(2, 3, 14); d14[..., 12] = 0.01; d14[..., 13] = 0.02

undistort_image(img, K, torch.zeros(2, 3, 4)).shape    # (2, 3, 1, 5, 6)  -- 4 coefficients: fine
undistort_image(img, K, torch.zeros(2, 3, 14)).shape   # (2, 3, 1, 5, 6)  -- 14 coefficients, zero tilt: fine
undistort_image(img, K, d14)                            # RuntimeError: The size of tensor a (3) must match the size of tensor b (6) at non-singleton dimension 1
undistort_image(img[0], K[0], d14[0]).shape             # (3, 1, 5, 6)     -- one leading axis with the same tilt: fine
```

`distort_points` on `(2, 3, N, 2)` points with the same `d14` raises the same `RuntimeError`, and `undistort_points` raises `ValueError: Input batch size must be the same for both tensors or 1. Got torch.Size([6, 3, 3]) and torch.Size([2, 3, 5, 2])`. Under ONNX export (`is_exporting()` true) the tilt branch is taken for every coefficient vector, so there even a 4-coefficient multi-axis batch fails.

## Expected

`tilt_projection` keeps the leading dimensions of its inputs: `(2, 3, 1)` angles return `(2, 3, 3, 3)`, and the three callers accept the same multi-axis batches on the tilt branch that they already accept off it. Batch 5b (#4312) documents the multi-axis batch form on `undistort_image` and needs to scope that sentence to the non-tilt branch until this is fixed.

## Context

Found while reviewing #4312 (batch 5b of #4262). Distinct from #4276, which is about *which* matrix the forward branch returns; this is about its batch shape. Both live in the same function, so one PR may reasonably fix both.

ducha-aiki:
## Summary

Documents and pins the distortion and calibration surface from the batch 5 audit: which entry points take
normalized `z = 1` points and which take pixels, the one place the OpenCV coefficient layout is written down,
what `K` and `new_K` each do in the forward and the inverse map, and what the two iterative inverses converge
to. Eleven `Convention:` blocks — one per symbol — nine `.. warning::` blocks linking the six issues below,
and 37 additive test pins.

empty for all five modified `kornia/` files. No existing test assertion and no doctest was modified; every pin
is new (the single deleted line under `tests/` is an import extended in place). No `docs/` page is touched.
Part of #4262, with a changelog entry under `## Unreleased` -> `### Added`. Independent of #4294 (sub-batch
5a): the pixel-centre convention is stated inline ("the top-left centre is `(0, 0)`") rather than by xref to the
`PinholeCamera` anchor block, so either order works.

## Scope

| symbol | what it got |
| --- | --- |
| `distort_points_affine` | `Convention:` (normalized `z = 1` point + `[fx, fy, cx, cy]` in, pixels out; `calibration.distort_points` is the pixel/`K` counterpart) |
| `undistort_points_affine` | `Convention:` (pixels in, normalized out; closed form — one subtraction and one division per axis, no iteration; zero `fx`/`fy` gives an infinity carrying the numerator's sign, or `nan` exactly on the principal-point coordinate) |
| `dx_distort_points_affine` | `Convention:` (the `(..., 2, 2)` Jacobian, rows `(u, v)`, columns `(x, y)`; equals `diag(fx, fy)` and does not depend on the point) |
| `distort_points_kannala_brandt` | `Convention:` (`params = [fx, fy, cx, cy, k0, k1, k2, k3]`, first four the affine part, `k0..k3` multiplying `θ², θ⁴, θ⁶, θ⁸`) |
| `undistort_points_kannala_brandt` | `Convention:` (a fixed Gauss-Newton budget that is not a parameter and has no convergence test; the round trip reaches the rounding floor in `float32` and stops at about `1e-8` in `float64`, the bias of the `1e-8` rescale guard; the three guard constants by value, returning the origin rather than `nan` where they are representable) + warning #4308 |
| `dx_distort_points_kannala_brandt` | `Convention:` (shape and layout only) + warning #4277 |
| `calibration.distort_points` | `Convention:` (pixel in, pixel out; the coefficient layout's single home; lengths 4, 5, 8, 12, 14 accepted and zero-padded to 14, every other length `ValueError`; `new_K` normalizes and `K` denormalizes; the leading-dimension rule scoped to eager and the non-tilt path, #4324) + warnings #4276, #4286 |
| `calibration.undistort_points` | `Convention:` (the mirrored intrinsics roles; a fixed-point iteration of `num_iters` steps that improves down to the working dtype's rounding floor, which `float16` reaches at the default count; the leading-dimension rule scoped to eager and the non-tilt path, #4324) + warnings #4285 (a two-point cycle outside the valid radius, so the answer depends on the parity of `num_iters`), #4276, #4286 |
| `undistort_image` | `Convention:` (the leading-dimension rule and the one legacy exception, scoped to eager and the non-tilt path, #4324; the map is `distort_points` on `create_meshgrid`'s integer-centre grid; `align_corners=True` is baked in; zero coefficients give a close but not byte-identical image) + warning #4276 |
| `tilt_projection` | `Convention:` (`R = Ry(tauy) @ Rx(taux)`, `Pz` built from `R`'s third column, both branches exactly `eye(3)` at zero angles, the inverse branch inverts `Pz R` and is the one `undistort_points` applies; every leading axis of the angles is flattened to `(numel, 3, 3)`, #4324) + warning #4276 |
| `solve_pnp_dlt` | `Convention:` (the `(B, 3, 4)` output is the world-to-camera `[R | t]`, the same direction as `PinholeCamera.extrinsics` and the inverse of the camera pose; `weights` scales each point's two rows; the guards and what they raise) |

The #4276 and #4286 explanations are written once (on `tilt_projection` and `distort_points`) and the other
sites carry one-sentence pointers. No pointer line is used: no 5b symbol is a method of an anchored class.

## Pins

37 new pins — 25 `test_convention_*` (two of them strict `xfail`s) and 12 `test_wart_*` (counting a
parametrized function once per parameter set).

`tests/geometry/camera/test_distortion.py` — 7 convention (one a strict `xfail`), 3 wart:
`test_convention_distort_points_affine_takes_normalized_z1_points`,
`test_convention_undistort_points_affine_is_the_exact_inverse`,
`test_convention_dx_distort_points_affine_matches_autograd`,
`test_convention_undistort_points_kannala_brandt_round_trip_closes`,
`test_convention_undistort_points_kannala_brandt_closes_far_off_axis`,
`test_convention_principal_point_undistorts_to_the_origin`,
`test_wart_principal_point_undistorts_to_nan_in_float16_4308` (#4308),
`test_wart_rescale_guard_caps_float64_round_trip_4308` (#4308, the float64 side of the same constant),
`test_wart_dx_distort_points_kannala_brandt_disagrees_with_autograd_4277` (#4277),
`test_convention_dx_distort_points_kannala_brandt_matches_autograd_4277` — **strict `xfail`**, the autograd
value the #4277 repair must produce.

`tests/geometry/calibration/test_distort.py` — 8 convention (one a strict `xfail`), 3 wart:
`test_convention_accepts_4_5_8_12_14_coefficients_and_rejects_6`,
`test_convention_zero_coefficients_are_a_no_op_at_dtype_tolerance`,
`test_convention_zero_coefficients_are_bit_exact_on_these_points`,
`test_convention_new_K_normalizes_and_K_denormalizes`,
`test_convention_agrees_with_distort_points_affine_on_the_same_camera`,
`test_convention_tilt_projection_zero_angles_are_the_identity`,
`test_convention_tilt_projection_inverse_branch_inverts_pz_times_r`,
`test_wart_tilt_projection_forward_is_pz_times_r_transpose_4276` (#4276),
`test_wart_tilt_projection_flattens_leading_axes_4324` (#4324, both branches), and the **strict `xfail`**
`test_convention_tilt_projection_branches_are_inverses_4276` (#4276).

`tests/geometry/calibration/test_undistort.py` — 6 convention, 6 wart:
`test_convention_undistort_points_inverts_distort_points`, `test_convention_new_K_denormalizes_and_K_normalizes`,
`test_convention_more_iterations_shrink_the_round_trip_residual`,
`test_convention_zero_coefficients_are_close_but_not_byte_identical`,
`test_convention_accepts_both_batch_conventions`, `test_convention_resamples_with_align_corners_true`,
`test_wart_distort_undistort_round_trip_breaks_with_tilt_4276` (#4276),
`test_wart_fixed_point_cycles_outside_valid_radius_4285` (#4285: `res(5) == res(7)`, `res(6) == res(8)`,
`res(6) > res(5) > 1`),
`test_wart_multi_axis_tilt_batch_fails_4324` (#4324; `distort_points` and `undistort_points`, eager tilt and
export), `test_wart_unbatched_export_undistort_fails_4324` (#4324),
`test_wart_multi_axis_tilt_image_fails_4324` (#4324, eager tilt and export).

`tests/geometry/calibration/test_pnp.py` — 4 convention:
`test_convention_returns_the_world_to_camera_extrinsics`, `test_convention_world_to_camera_translation_sign`,
`test_convention_rejects_4x4_intrinsics`, `test_convention_planar_world_points_raise`.

Every wart pin asserts the **current** behavior, names its issue in the test name, and says in a comment that it
is deleted when the issue is repaired. Every pin was mutation-checked: the claim was flipped, the node observed
to fail, and the flip reverted.

## Issues documented here

| --- | --- | --- |
| #4276 `tilt_projection`'s forward branch returns `Pz Rᵀ` where OpenCV returns `Pz R`, so `distort_points` and `undistort_image` are wrong whenever `τ ≠ 0` | broken output (idiom 4) | n/a — single symbol |
| #4277 `dx_distort_points_kannala_brandt` is not the Jacobian of `distort_points_kannala_brandt`; its own test and doctest pin the wrong value | broken output (idiom 4 on the `nan` at the origin) | n/a — single symbol |
| #4285 `undistort_points`' fixed-point iteration diverges silently outside the valid radius, and more iterations make it worse | semantic contract (idiom 4) | n/a — single symbol |
| #4286 `torch.compile(fullgraph=True)` and `torch.jit.script` gaps on the distortion path | compatibility | n/a — single symbol; one issue, five sites |
| #4308 `undistort_points_kannala_brandt` returns `nan` at the principal point in `float16`, because all three singularity guards underflow; the same `1e-8` rescale guard caps the `float64` round trip at about `1e-8` | broken output, dtype-scoped (idiom 2 — a constant that does not inherit the input's dtype) | n/a — single symbol |
| #4324 `tilt_projection` flattens every leading axis to `(-1, 3, 3)`, so `distort_points`, `undistort_points` and `undistort_image` fail on multi-axis batches whenever the tilt path is taken, which under ONNX export is always | broken output (shape) | n/a — single symbol, three callers |

#4308 and #4324 were found while writing and reviewing this sub-batch and filed with repros re-executed against
this PR's base; the `float64` side of #4308 is recorded on the issue.

## Docstring errors fixed

- `distort_points` and `undistort_image` — `Radial (k_1, k_2, k_3, k_4, k_4, k_6)` had `k_4` twice; it is
  `k_5`, as the module page and `undistort_points` already spell it.
- `distort_points` Returns — "Undistorted 2D points"; it is the forward map, so "Distorted 2D points".
- `tilt_projection` Returns — "Inverse tilt projection matrix" for **both** branches; now "Tilt projection
  matrix, or the inverse tilt projection matrix when `return_inverse` is True".
- `dx_distort_points_affine` and `dx_distort_points_kannala_brandt` — both summaries said "Compute the
  derivative of the x distortion with respect to the x coordinate", which describes a scalar, and both Returns
  lines said shape `(..., 2)`; execution gives `(..., 2, 2)`. The affine one now says it is that model's
  Jacobian; the Kannala-Brandt one is deliberately neutral, because the warning below says it is not one.
- `dx_distort_points_kannala_brandt` carried `.. math:: ∂u/∂x = diag(f_x, f_y)`, copied verbatim from the
  affine function. That is the affine model's Jacobian and this function does not return it, so the false block
  is **deleted**; the Convention block states the real layout and the warning states #4277.
- `distort_points` / `undistort_points` `new_K` Args — both said "Intrinsic camera matrix of the distorted
  image ... you may additionally scale and shift the **result**". Executed with zero coefficients and
  `new_K = 2 K`: `distort_points` **halves** its output and `undistort_points` **doubles** it. Both entries now
  state which matrix normalizes and which denormalizes, and which end a different matrix rescales.
- `undistort_points_affine`'s `.. math::` read `[u;v] - [c_x;c_y] · F⁻¹`, which is neither conformable nor what
  the code computes (`F⁻¹([u;v] - [c_x;c_y])`); reordered with an explicit parenthesis.

## Docs-only proof

For each of the five modified `kornia/` files, parse both revisions, pop the leading string `Expr` from every
`Module` / `ClassDef` / `FunctionDef` / `AsyncFunctionDef`, `ast.dump`, and diff
(`diff <(git show e8111a61:$f | strip_docstrings) <(strip_docstrings < $f)`) — **EMPTY for all five**:
`calibration/distort.py`, `calibration/undistort.py`, `calibration/pnp.py`, `camera/distortion_affine.py`,
`camera/distortion_kannala_brandt.py`.

## Validation

- `pixi run lint`, `pixi run pre-commit-all` — passed
- `pixi run doctest` — 650 passed, 11 skipped (no `>>>`/`...` line appears in the `kornia/` diff)
- clean `pixi run build-docs` (`rm -rf docs/build/html` first) — `build succeeded`, and the
  `distort|calibration|pnp|tilt|undistort` filter over the build log's warning/error lines is empty; every xref
  this change adds was checked against `docs/build/html/objects.inv` and as a real `<a href>` in the built page
- `pytest tests/geometry/camera/test_distortion.py tests/geometry/calibration --dtype=float32,float64` —
  190 passed, 3 skipped, 4 xfailed. At `--dtype=float16,bfloat16` the failure-**ID set** is exactly the 29 IDs
  of `main`, and with the CI manifests (`--xfail-known-failures --known-failure-profile=cpu-float16` /
  `cpu-bfloat16`) nothing fails; at `--device=mps --dtype=float32,float16` (with `PYTORCH_ENABLE_MPS_FALLBACK`
  unset) exactly the 15 mps IDs of `main` — nothing new, nothing lost on either device

`calibration.distort_points` / `calibration.undistort_points` (KEEP SEPARATE — forward/inverse pair; the `τ`
defect is #4276) · `distort_points_affine` / `calibration.distort_points` (KEEP SEPARATE — same answer, different
input domain and parametrization; pinned by `test_convention_agrees_with_distort_points_affine_on_the_same_camera`)
· `distort_points_affine` / `undistort_points_affine` (KEEP SEPARATE — forward/inverse pair) · the D-row
`undistort_image` with no `distort_image` (recorded, no issue).

## Notes for reviewers

- **`undistort_image` with zero coefficients is not byte-identical to its input.** The audit expected it to be;
  execution refutes it for 0 of 300 random cameras, for `K = eye(3)`, and in float64. The distortion *map* is
  likewise the pixel grid only up to rounding — bit-exact on the audit's camera, false on 275 of 500 random
  ones — so the prose says so and the bit-exact case is pinned separately on representable points.
- **`tilt_projection`'s forward branch returns `Pz Rᵀ`**, and `docs/source/geometry.calibration.rst` documents
  OpenCV's `Pz R`. The page is the *intended* form, so it is left unedited: it becomes correct on its own when
  #4276 is repaired. The `tilt_projection` warning says so in prose.
- **#4285 is documented on `undistort_points` only, and its pin records the parity, not a value.** Outside the
  valid radius the fixed point is a two-point cycle (residual 287.8 px at every odd count, 1337.8 px at every
  even one, on the pin's camera), so the pin asserts `res(5) == res(7)`, `res(6) == res(8)` and
  `res(6) > res(5) > 1` and flips when a divergence guard lands, without settling the correct value. **Three
  other Convention sentences are deliberately unpinned** and were executed twice each instead:
  `solve_pnp_dlt`'s `weights` scaling, its `BaseError` on too few points, and `undistort_points_affine`'s
  zero-`fx` behavior. Pins for them are a follow-up.
- **The batch sentences on `distort_points`, `undistort_points` and `undistort_image` describe eager execution
  on the non-tilt path.** With non-zero tilt, or always under ONNX export, two leading axes fail with the #4324
  `RuntimeError`, and `undistort_points`' legacy unbatched form fails with a `ValueError`; both are pinned.
  #4345 fixes #4324 and, when it merges, deletes those pins and the four `#4324` prose sites here.
- **`tests/geometry/calibration/test_distort.py::TestDistortPoints::test_jit[cpu-float16]` is a pre-existing
  unseeded flake** in a body that is byte-identical at the base commit. It draws `torch.rand` for `K` and
  `new_K` with no `manual_seed`, so a draw with a tiny `fx` blows the float16 tolerance — measured 1 red in 5
  full-gate runs, and green in the runs recorded above. This PR does not touch it; seeding it is a follow-up.
- **`undistort_points_kannala_brandt`'s output dtype follows `points`, not `params`** (the reason float16 points
  with float32 params escape #4308). The wart pin asserts it; the block does not yet state it — a follow-up.
- **The existing `test_dx_distort_points_kannala_brandt` and that function's doctest both pin the wrong Jacobian
  and are left untouched**, per the batch rule that a documentation PR is additive. The #4277 fix must re-derive
  both; the strict `xfail` beside them becomes the new assertion.

Posted on behalf of @ducha-aiki by Claude (Opus 5); revised after review by Claude (Claude Fable 5.1).


ducha-aiki:
## Review of #4312 (batch 5b: distortion and calibration) at `c0b50ad7`

**Verdict: changes requested.** The docs-only claim holds (docstring-stripped AST of the five `kornia/` files identical to `e8111a61`, doctests unchanged), the suite is green on cpu at all four dtypes and on mps float32/float16 with the fallback unset, and the headline claims (forward tilt branch is `Pz @ R.T`, the inverse branch inverts `Pz @ R`, the docs page writes `Pz @ R`, `fullgraph=True` breaks on the tilt test, the `new_K`/`K` roles, the coefficient-length rule) all reproduce. Six sentences in the new prose are false as written, one of them because of a defect that was not yet filed. Everything below was executed on the branch worktree (torch 2.14.0).

### Blocking

1. **`undistort_image` Convention: leading dims "may be empty, a single batch axis, or several axes deep" is false on the tilt branch.** Executed with image `(2, 3, C, H, W)`, `K (2, 3, 3, 3)`, `dist (2, 3, 14)`: with 4 coefficients or with 14 and zero tilt the call returns `(2, 3, C, H, W)`; with `tau = (0.01, 0.02)` it raises `RuntimeError: The size of tensor a (3) must match the size of tensor b (6)`. The same shapes with one leading axis work. Root cause is `tilt_projection` reshaping its angles to `(-1,)`, so `tilt_projection(zeros(2, 3, 1), zeros(2, 3, 1))` returns `(6, 3, 3)` where its own Returns line promises `(*, 3, 3)`; `distort_points` and `undistort_points` fail the same way. I filed that as **#4324**. Scope the sentence to the non-tilt branch (and under ONNX export the tilt branch is always taken), add the `(*, 3, 3)` caveat to `tilt_projection`'s block, and pin the multi-axis + tilt case as a wart for #4324.
2. **`undistort_points_kannala_brandt` Convention: the round-trip accuracy "is set by the model and the point rather than by the working dtype, and a caller cannot raise it".** The PR's own test comments record 1.91e-06 (float32) vs 2.03e-08 (float64) at `(3, 0)`, a 100x gap, so a caller raises it by switching dtype. Say what `undistort_points` says: the step count is fixed, the residual bottoms out at the working dtype's rounding floor.
3. **`undistort_points` Convention: "At the default step count that accuracy is set by the camera and the coefficients rather than by the working dtype".** In float16 the residual is 1.56e-02 (one ulp at 54) at 5 steps and still at 50; the sibling test comment already says so. Qualify to float32/float64.
4. **`undistort_points_kannala_brandt` names "the Newton denominator and the final radial rescale" as the two guards that underflow.** The source has three constants: `rth.clamp(min=1e-16)` on the Newton start, `d_thd + 1e-12` on the Newton denominator, `rth + 1e-8` on the rescale. All three underflow in float16, but at the principal point `d_thd = 1` regardless, so only the `1e-8` produces the `nan`. Name them precisely (#4308 already lists all three); the test comment names a different pair ("1e-16 on the Newton start, 1e-8 on the final rescale"), so the two currently disagree.
5. **Test comments in `test_undistort.py` say "raising num_iters to 50 does not change the answer" (round-trip pin) and "50 steps change nothing" (`test_convention_more_iterations_shrink_the_round_trip_residual`).** Executed on the pin's own camera: float64 residual 4.62e-05 at 5 steps, 7.5e-10 at 10, 0.0 at 50; float32 3.81e-05, 1.9e-06, 1.9e-06. That contradicts the Convention text the tests exist to pin. The float64 5-step arm passes only because the tester's float64 atol (1e-5) plus rtol at |x| = 54 swallows 4.6e-5; moving the second point to `[-16, 3]` would fail it. Fix the comments and either raise the step count or state the tolerance dependence.
6. **`test_wart_dx_distort_points_kannala_brandt_disagrees_with_autograd_4277`: `assert not allclose(analytic.T, autograd)` is vacuous.** With `fx == fy == 100` the analytic matrix is exactly symmetric, so the line re-asserts the one above it and does not evidence the docstring's "transposing it does not close the gap". `fy = 50` discriminates (analytic `[[22.06, -35.78], [-17.89, 37.86]]` vs autograd `[[84.64, -4.49], [-2.24, 45.69]]`), the same reason this PR uses `taux != tauy` elsewhere.

### P2

7. `test_distort.py` bit-exact pin comment: "when `(u - cx) / fx` is exactly representable" is not why it passes; `-0.2` and `0.2` are not representable in float32/float16/bfloat16, the product just rounds back. State it as an executed fact on these four points.
8. `:func:\`...\`\'s forward branch` renders fine in Sphinx but shows the backslash in `help()`; rephrase ("the forward branch of ..."). `dx_distort_points_affine` keeps the `∂u/∂x` label on what the new text calls the full Jacobian (both from Copilot; both real).
9. Eight docstring sites embed test-function names (`distort.py` ~L42-44, L90-93; `undistort.py` ~L73; `distortion_kannala_brandt.py` ~L134, L217-218). No merged Convention block on `main` does that; a rename silently stales the rendered docs while the AGENTS.md issue-number grep already finds the blast radius. Keep the issue links, drop the names. The #4276 warning is written out four times and #4286 twice; the batch-4 pattern is one full explanation plus one-sentence pointers.
10. Six docstrings point at "the Convention block on `PinholeCamera`", which exists only in #4294. Merge 5a first, as planned, so the sentence is true when this renders.
11. Cross-cutting for 5a to 5d: pin comments cite "audit labels 5b-xx-NN", "the batch-5b worktree" and `full_audit.py`, none of which exist in this repository or its history. Replace the worktree with the commit hash and drop or spell out the labels.
12. Optional: `_pz_r`'s `R` half is a line-for-line copy of `tilt_projection`'s own construction, which is what makes the atol=0 pin hold. `testing.geometry.linalg.euler_angles_to_rotation_matrix(-taux, -tauy, 0)[:, :3, :3]` is `torch.equal` to it on every cell and would make the reference independent of the function under test.

Checked and clean: OpenCV reference values in `test_opencv_all_coeff` carry non-zero tau, so the "reproduces OpenCV" claim is not vacuous; `solve_pnp_dlt` direction, weights, too-few-points and 4x4 rejection; the affine zero-focal `inf`/`nan` sentence; the fp16 principal-point `nan`; the strict xfails for #4276 and #4277; ruff at the pinned 0.16.4; CHANGELOG placement.

ducha-aiki:
Two additions to the review above, both executed on the branch worktree (torch 2.14.0, float64):

7. **The #4285 warning on `undistort_points` ("raising `num_iters` makes the answer worse instead of better") is a parity artefact, not a trend.** On #4285's own probe (`fx = fy = 100`, `k1 = 0.5`, point `(304, 3)`) the round-trip residual by `num_iters` is 1: 288.0, 2: 1338.1, 3: 287.8, 4: 1337.8, 5: 287.8, 6: 1337.8, 7: 287.8, 20: 1337.8, 21: 287.8, 199: 287.8, 200: 1337.8. Outside the valid radius the fixed point is a two-point cycle, so the answer depends on whether the count is odd or even and no count converges; the issue's "1338 px at 200" is the even branch. Say that instead, and this *is* pinnable without settling the correct value: `res(5) == res(7)`, `res(6) == res(8)`, `res(6) > res(5) > 1` records today's behaviour and flips when a divergence guard lands. I am correcting #4285 in parallel.
8. **The Convention blocks describe the eager surface only; the ONNX branch is worse than #4324.** With `is_exporting()` true the tilt branch is taken for every coefficient vector, so `undistort_points(rand(1, 2), eye(3), rand(4))`, the legacy unbatched form the block documents as accepted, raises `ValueError: Input batch size must be the same ...`, and `distort_points` on `(2, 3, 5, 2)` points with `(2, 3, 3, 3)` `K` raises the #4324 `RuntimeError` even with four coefficients. AGENTS.md treats ONNX as its own compatibility surface; scope the batch sentences on `distort_points`, `undistort_points` and `undistort_image` to eager and cite #4324, which I have extended with these two cases.

One policy note rather than an ask, since the batch rule is "document and pin, fix in the window": the review fork prototyped #4276 as the one-token `Pz @ R.transpose(-1, -2)` to `Pz @ R` and reports that with it `distort_points` reproduces the repo's own `test_opencv_all_coeff` data to 6e-05 px (today 611 px) and the only new failures are the two wart pins and the strict xfail flipping. If you would rather take that fix here than carry four warnings and three pins for one day, the pin to add is `assert_close(distort_points(U, K, D), P, atol=1e-3)` beside `test_opencv_all_coeff`. Your call; the review stands either way.

ducha-aiki:
## Final review of #4312 at `0184b310` (delta `d2fe9507`, plus fixes)

**Verdict: APPROVE — merge-ready once CI is green.** The delta resolves every ask from the round-1 review and the follow-up, and I re-executed each one rather than reading it. One new false sentence turned up in the rewritten Kannala-Brandt block; I fixed it on the branch (`0beae601`) together with three nits and merged `main` (`0184b310`) so the stale "conflicting" flag clears and CI actually runs on the head. Everything below was executed on the branch (torch 2.14.0, cpu and mps).

### Round-1 asks, all resolved on `d2fe9507`

1. `undistort_image` leading dims — scoped to eager and the non-tilt path on all three blocks, `tilt_projection`'s block and Returns line now say `(taux.numel(), 3, 3)`, #4324 cited; pinned by `test_wart_tilt_projection_flattens_leading_axes_4324` (both branches), `test_wart_multi_axis_tilt_batch_fails_4324` (both ops, eager tilt and export), `test_wart_unbatched_export_undistort_fails_4324` and `test_wart_multi_axis_tilt_image_fails_4324`. Re-executed: `(2, 3, 5, 2)` points fail with tilt and under export, the unbatched form works for `distort_points` in both modes and fails for `undistort_points` in both, exactly as the two blocks now say.
2. / 3. The two dtype-independence claims are gone; `undistort_points` now states the float16 floor at the default count and the float32/float64 improvement, matching the recorded residuals.
4. The three guard constants are named by value, and the warning names the `1e-8` rescale as the one that produces the `nan`; source confirms (`denom = rth + eps`, `mag = |tan θ| / denom`).
5. The "50 steps change nothing" comments are replaced by the tolerance explanation (`4.62e-05` passes through `atol + rtol·|x|`).
6. `fy = 50` in the #4277 wart pin, with a symmetry guard on the analytic matrix; the literals match my round-1 values.
7.–9., 11. Bit-exact comment rewritten and the test renamed `_on_these_points`; backslash-apostrophe and `∂u/∂x` gone; no test name, audit label, worktree or `full_audit.py` reference survives in `kornia/` or `tests/` (grep is empty); #4276 is explained once on `tilt_projection` and #4286 once on `distort_points`, the other sites carry one-sentence pointers.
10. The `PinholeCamera` xref was removed rather than kept: the pixel-centre rule is stated inline, so the PR no longer depends on #4294 rendering first.
12. `_pz_r` builds `R` with `testing.geometry.linalg.euler_angles_to_rotation_matrix`; the `atol=0` forward pin still passes on cpu float32/float64/float16/bfloat16 and mps float32/float16, so the two constructions are bit-equal.
Follow-up 7 (#4285 parity): the warning now says two-point cycle and odd/even, and `test_wart_fixed_point_cycles_outside_valid_radius_4285` pins it. On its camera: 5 → 287.789, 6 → 1337.789, 7 → 287.788, 8 → 1337.788, 50 → 1337.788, 51 → 287.788.
Follow-up 8 (ONNX branch): scoped on all three blocks and pinned through a monkeypatched `is_exporting`.

### New finding, fixed in `0beae601`

**`undistort_points_kannala_brandt`: "the residual eventually bottoms out at the working dtype's rounding floor" was false in float64.** The float64 round trip stops at about `1e-8`, forty million ulps above the floor, and the cause is the `1e-8` rescale guard: `mag = |tan θ| / (rth + 1e-8)` scales every result by `1 - 1e-8 / r`. Re-executing the function with only that constant set to `0.0`:

| point | float64, `eps = 1e-8` | float64, `eps = 0` | predicted `|x| · 1e-8 / r` | float32 |
| `(0.5, 0.25)` | 9.55e-09 | 5.6e-17 | 8.9e-09 | 2.98e-08 |
| `(3, 0)` | 2.03e-08 | 0.0 | 1.0e-08 | 1.91e-06 |
| `(0.01, 0)` | 1.00e-08 | 1.7e-18 | 1.0e-08 | 1.12e-08 |

The block now says float32 reaches the rounding floor and float64 stops at about `1e-8` because of that guard, and `test_wart_rescale_guard_caps_float64_round_trip_4308` pins it (`1e-9 < residual < 1e-7` on `(3, 0)`; mutation-checked: with the guard at `0.0` it fails on `assert 1e-09 < 0.0`). Recorded on #4308 as the float64 side of the same constant (https://github.com/kornia/kornia/issues/4308#issuecomment-5576626966): a dtype-aware `eps` alone does not fix it, the additive term has to go.

Also in `0beae601`: "only a scalar or one leading batch dimension" → "only unbatched inputs or one leading batch dimension" on `distort_points`; the four #4324 wart pins carry the same "delete when repaired" note as the other wart pins, so #4345's author finds them. The PR body is refreshed for the delta: the six-xref claim, the pin list (37 now, the rename, the #4285 pin that the body said did not exist by design), the #4324 row, and the validation counts.

### Verified at `0184b310`

- Docs-only: docstring-stripped AST of all five `kornia/` files identical to `e8111a61`.
- `tests/geometry/camera/test_distortion.py tests/geometry/calibration` at `--dtype=float32,float64`: 190 passed, 4 skipped, 4 xfailed. `--dtype=float16` / `bfloat16` with the CI manifests: no failures. Raw failure-ID sets at `float16,bfloat16` (29) and mps `float32,float16` (15, fallback unset) identical to `main` `f4532f39`.
- ruff 0.16.4 and `pre-commit run --all-files` clean; no line over 120.
- `sphinx -W` on the merged tree: `build succeeded`; all eleven blocks and nine warnings render, every `:func:` in them is a live link (checked the HTML, since a dead short-form target emits no warning), every issue link resolves.
- `git diff origin/main HEAD` touches only the PR's ten files; the `#4312` CHANGELOG entry survives the merge.

### Coordination

- **#4345** (fixes #4324, currently blocked) must delete the four `_4324` pin functions and the four `#4324` prose sites here when it lands; `grep -rnE "#4324|issues/4324" kornia/ tests/` is the gate.
- **#4347** adds `test_convention_out_of_radius_fixed_point_is_a_two_cycle_4285`, the same claim as `test_wart_fixed_point_cycles_outside_valid_radius_4285` here. This one runs under the device/dtype fixtures and the repo helper; whichever merges second drops its copy.
- Optional: the delta dropped the "that exception is not a broadcast" sentence from `undistort_image`; the `ValueError` on `B = 2` with unbatched `K` is still pinned by `test_convention_accepts_both_batch_conventions`, so nothing is lost either way.

Issue #4269, opened 2026-09-05
W1b: `reproject_disparity_to_3D` — `u` and `v` are swapped, and the existing real-data pin encodes the swap


ducha-aiki:
`StereoCamera.reproject_disparity_to_3D` and the module-level `reproject_disparity_to_3D` (they share one body) unbind the pixel meshgrid as `v, u = torch.unbind(uv, dim=-1)`. `create_meshgrid(..., normalized_coordinates=False)` returns `(x, y)`, so `uv[..., 0]` is the **column** and `uv[..., 1]` is the **row**. The code therefore feeds the column index into `v` and the row index into `u`, and the returned `X` is computed from the row while `Y` is computed from the column — the opposite of `cv2.reprojectImageTo3D`, whose semantics this function was added to provide (#2042, closed 2025-12-29).

Every image gets wrong 3D coordinates, silently. What comes out is the OpenCV answer with the two pixel indices **transposed**: kornia's value at `(row, col)` is what `cv2.reprojectImageTo3D` puts at `(col, row)` — exactly `X = (row - cx) Z / fx`, `Y = (col - cy) Z / fy` with `Z = fx · tx / d`. `Q`'s own focal swap (row 0 carries `fy·b` and row 1 `fx·b`) compensates for the homogeneous divide, so `X` is still scaled by `1/fx` and `Y` by `1/fy`; what moved is only which *index* feeds which coordinate. A symmetric camera does **not** make it cancel: on a square image with `fx == fy` and `cx == cy`, transposing the index pair is a no-op only where `row == col`, so the diagonal agrees with OpenCV and nothing else does — 0.2 apart on the 5 × 5 rig in the repro below.

Three independent reproductions below: a synthetic asymmetric grid against the hand-computed OpenCV formula; the axis-dependence check (`X` must vary with the column and `Y` with the row, and they do the reverse); and **the repo's own real-data test fixture**, whose stored ground truth is the swapped value — `tests/geometry/camera/test_stereo.py::_RealTestData` stores a `(1, 10, 1, 1)` disparity (10 rows, 1 column, though its comment says "first 10 cols of 1 row"), and `test_reproject_disparity_to_3D_real` passes *because of* the bug.

## Repro

```python
import sys, os, torch
from kornia.geometry.camera import StereoCamera
from kornia.utils import create_meshgrid

def stereo(fx=100.0, fy=100.0, cx=4.0, cy=3.0, tx=0.5):
[...]

PR #4298, opened 2026-09-05
fix(geometry): keep the W axis in unproject_meshgrid when W is 1

thegoodengineer:
## What does this pull request do?

`unproject_meshgrid` squeezed the grid it builds with a bare `.squeeze()`. `create_meshgrid` returns `(1, H, W, 2)`, so that call is meant to drop the leading batch axis, but it drops *every* size-1 axis. At `W == 1` the width axis goes with it, the grid broadcasts across a phantom width, and the documented `(*, H, W, 3)` contract breaks:

```text
unproject_meshgrid(H=1, W=3): (1, 1, 3, 3)
unproject_meshgrid(H=3, W=1): (1, 1, 3, 3)   same shape, different values
depth_to_3d(W=1)    shape: (1, 3, 3, 1)
depth_to_3d_v2(W=1) shape: (1, 3, 3, 3)      should be (1, 3, 1, 3)
warp_frame_depth(W=1) shape: (1, 2, 4, 4)    should be (1, 2, 4, 1)
```

`H = 1` is unaffected, because the batch axis absorbs the squeeze, and `H = W = 1` is correct by accident. `W = 1` is a degenerate but valid shape (a single image column, a 1-pixel-wide crop) and nothing raised.

The squeeze is now `squeeze(0)`, with a comment saying why the axis is named explicitly. Three regression tests are added, parametrized over degenerate and ordinary sizes: the grid shape itself, `depth_to_3d_v2` against `depth_to_3d` (the sharper pin, since the two agree everywhere else), and the `warp_frame_depth` output shape.

## Related issue or discussion

Closes #4278

## What did you check?

The new tests on unmodified `main` fail at `W = 1` and only there, which is exactly the reported shape:

```text
6 failed, 16 passed, 109 deselected
FAILED test_unproject_meshgrid_degenerate_sizes[cpu-float32-3-1] - assert torch.Size([2, 1, 3, 3]) == (2, 3, 1, 3)
FAILED test_depth_to_3d_v2_degenerate_sizes[cpu-float32-3-1]    - assert torch.Size([2, 3, 3, 3]) == (2, 3, 1, 3)
FAILED test_shape_degenerate_sizes[cpu-float32-4-1]             - assert torch.Size([1, 2, 4, 4]) == (1, 2, 4, 1)
```

[...]

Issue #4280, opened 2026-09-05
W1b: `depth_from_plane_equation` — the singularity guard is a no-op exactly at the singularity


ducha-aiki:
`depth_from_plane_equation` guards its division by the ray-plane dot product with

```python
denom = torch.where(mask, eps * torch.sign(denom), denom)     # mask = denom.abs() < eps
```

`torch.sign(0.) == 0.`, so at `denom == 0` — the exact singularity the guard exists for — the epsilon is multiplied by zero and the guard is a no-op. A ray exactly parallel to the plane returns `inf`.

The masked-guard *shape* is right (unlike an `x + eps` that perturbs every result, this one only touches the masked branch — and it works: a denominator of `2.4e-09`, inside the mask but non-zero, is clamped to `±eps` and returns `±2e8`). Only the worst case is missed.

A grazing ray is not an exotic input: it is every pixel on the horizon of a ground plane.

## Repro

```python
import torch
from kornia.geometry.depth import depth_from_plane_equation

K = torch.tensor([[100.0, 0.0, 4.0], [0.0, 100.0, 3.0], [0.0, 0.0, 1.0]])[None]
px = torch.tensor([[[4.0, 3.0]]])            # the principal point: its ray is (0, 0, 1)
# a plane whose normal is exactly perpendicular to that ray -> n . ray == 0 exactly
print("grazing ray, denom == 0 exactly :", depth_from_plane_equation(torch.tensor([[0.0, 1.0, 0.0]]), torch.tensor([[2.0]]), px, K).item())
print("torch.sign(0.0)                 :", torch.sign(torch.tensor(0.0)).item(), "<- the guard multiplies eps by this")
# a denominator just inside the mask, but non-zero: the guard does clamp it
n = torch.tensor([[0.0, 1.0, 2.384185791015625e-09]])
print("denom in (0, eps)               :", depth_from_plane_equation(n, torch.tensor([[2.0]]), px, K).item())
print("denom in (-eps, 0)              :", depth_from_plane_equation(-n, torch.tensor([[2.0]]), px, K).item())
```

## Observed

```
grazing ray, denom == 0 exactly : inf
torch.sign(0.0)                 : 0.0 <- the guard multiplies eps by this
denom in (0, eps)               : 200000000.0
denom in (-eps, 0)              : -200000000.0
```

## Expected

The sign is taken with a zero-safe fallback, so `denom == 0` is clamped like every other point inside the mask:

```python
[...]

PR #4317, opened 2026-09-06
docs(geometry): Convention blocks for camera (batch 5c: depth and stereo)

ducha-aiki:
## Summary

Documents and pins the depth and stereo surface from the batch 5 audit: which functions read `depth` as a
camera-frame `z` and which read it as a ray length, when the point cloud is `(B, 3, H, W)` and when it is
`(B, H, W, 3)`, which frame the depth belongs to in each of the two depth warps, what the rectified `Q` matrix
is and which pixel index feeds which output coordinate. Nine `Convention:` blocks — two of them the
`DepthWarper` and `StereoCamera` anchors — seven pointer lines into those anchors, eight `.. warning::` blocks linking the eight open issues below, and 37 additive test pins.

empty for both modified `kornia/` files. No existing test assertion and no doctest was modified; the `tests/`
diff has **zero** removed lines, so `_RealTestData` and every pre-existing test are byte-identical. Two rst
pages are touched: `geometry.depth.rst` gains two directives and `geometry.camera.stereo.rst` two corrected
sentences. Part of #4262, with a changelog entry under `## Unreleased` -> `### Added`.

Independent of #4294 (sub-batch 5a): the `PinholeCamera` anchor block lands there and this branch only links it
by xref, and those xrefs resolve against today's `main`, so either order works.

## Scope

| symbol | what it got |
| --- | --- |
| `DepthWarper` | **anchor** `Convention:` (depth in the source frame, image in the destination frame; the two-step API and what each method raises before `compute_projection_matrix`; `_dst_proj_src = K_dst @ (E_dst @ inverse(E_src))`, inverted and not transposed; `grid` versus what `warp_grid` returns; `align_corners=True`) + warnings #4273 and #4267 (the two warps agree away from `z = 0` and split outright at it) |
| `StereoCamera` | **anchor** `Convention:` (two rectified `(B, 3, 4)` projection matrices; `tx = -P_right[0, 3] / fx`; which focal sits in which `Q` row and how it cancels in the divide; `Q` and `-Q` give the same points wherever the homogeneous divide runs (`|W| > 1e-8`) and opposite ones where it is skipped, and the rst page's matrix is this `Q` verbatim under the page's own `tx`; the rst page's matrix is the same expression at the opposite baseline; disparity is `(B, H, W, 1)`) + warnings #4269, #4270 (+#4281 inline), #4275 |
| `unproject_meshgrid` | `Convention:` (the ray at depth 1 in `(B, H, W, 3)`, which is what makes it `depth_to_3d_v2`'s `xyz_grid` cache; `(B, 3, 3)` in, singleton extra leading dims pass through, a non-singleton one is unsupported — it raises `RuntimeError` or, when it happens to match `W`, silently broadcasts a different camera onto each column) + warning #4271 |
| `depth_to_3d_v2` | `Convention:` (channels-last `(*, H, W, 3)`; equal to `depth_to_3d` after `permute(0, 2, 3, 1)`; `xyz_grid`) + warning #4271 |
| `depth_to_3d` | `Convention:` (channels-first `(B, 3, H, W)`; the integer-centre pixel grid; `normalize_points` as ray length; an integer depth is promoted to `camera_matrix`'s dtype) |
| `depth_to_normals` | `Convention:` (`d/dx × d/dy` then normalize, so a fronto-parallel plane is `(0, 0, 1)` and `+z` points away from the camera; the two in-plane signs) |
| `depth_from_plane_equation` | `Convention:` (Hessian form in the camera frame; the result is a `(B, N)` list of depths, not a map; the masked signed clamp, and positive `eps` at an exactly zero denominator — the #4280 repair, now a regression pin) |
| `warp_frame_depth` | `Convention:` (depth on `dst`, image on `src`, `src_trans_dst` maps dst into src; `grid_sample` with `align_corners=True` and zero padding both baked in, and that padding is a zero extension the bilinear kernel blends with, not a mask; output is `(B, D, H, W)`) + warnings #4273, #4281 |
| `depth_from_disparity` | `Convention:` (`baseline * focal / disparity`; each of the two is a python `float` or a `(1,)` tensor — a 0-dim, a `(B,)` and a python `int` are all rejected; the sign is unchecked) + warning #4272 |
| `depth_warp`, `DepthWarper.compute_projection_matrix`, `.compute_subpixel_step`, `.warp_grid`, `.forward` | pointer line to the `DepthWarper` anchor |
| `StereoCamera.reproject_disparity_to_3D`, module-level `reproject_disparity_to_3D` | pointer line to the `StereoCamera` anchor |

## Pins

37 new pins — 23 `test_convention_*` (two of them strict `xfail`s) and 14 `test_wart_*`. Each wart pin's issue
is the number in its own name.

`tests/geometry/test_depth.py` — 18 convention (one strict `xfail`), 7 wart: `test_convention_pixel_origin_is_the_integer_centre`, `test_convention_integer_depth_follows_intrinsics_dtype`,
`test_convention_depth_to_3d_and_v2_agree_up_to_layout`, `test_convention_depth_to_3d_v2_keeps_the_w_one_axis_4278` (regression pin for the #4298 repair, covering `depth_to_normals` and `warp_frame_depth` shapes at `W = 1` too), `test_convention_xyz_grid_bypasses_the_camera_matrix`, `test_convention_returns_b_h_w_3_and_equals_depth_to_3d_v2_at_depth_one`,
`test_convention_normals_point_away_from_the_camera_and_x_tracks_the_column`, `test_convention_subpixel_border_blends_with_zero_padding`, `test_convention_src_trans_dst_moves_the_sampling_point`,
`test_convention_projection_matrix_is_k_dst_times_dst_trans_src`, `test_convention_warp_grid_requires_compute_projection_matrix_first`, `test_convention_forward_returns_b_c_h_w_and_align_corners_defaults_to_true`,
`test_convention_depth_warp_is_byte_identical_to_depth_warper`, `test_convention_warp_frame_depth_matches_depth_warper_without_being_bitwise_equal`, `test_convention_plane_z_equals_two_gives_depth_two`,
`test_convention_depth_from_plane_equation_clamps_a_tiny_denominator`, `test_convention_depth_from_plane_equation_clamps_the_singularity_4280` (regression pin for the #4348 repair, sign-aware: `+2e8` for both normal signs);
`test_wart_unproject_meshgrid_rejects_unbatched_intrinsics_4271`, `test_wart_unproject_meshgrid_extra_camera_axis_broadcasts_over_columns_4271`, `test_wart_warp_frame_depth_rejects_an_empty_batch_4281`,
`test_wart_warp_frame_depth_and_depth_warper_name_the_depth_frame_oppositely_4273`, `test_wart_warp_frame_depth_and_depth_warper_split_at_zero_transformed_depth_4267`, `test_wart_zero_disparity_gives_a_finite_depth_4272`,
`test_wart_depth_from_disparity_rejects_a_batched_baseline_4272`; and the **strict `xfail`** `test_convention_unproject_meshgrid_error_names_the_shape_the_caller_passed_4271`.

`tests/geometry/camera/test_stereo.py` — 5 convention (one strict `xfail`), 7 wart: `test_convention_reproject_disparity_takes_bhw1`, `test_convention_reproject_method_and_free_function_are_byte_identical`,
`test_convention_tx_is_the_right_translation_over_fx_and_fixes_q`, `test_convention_q_signs_only_cancel_away_from_a_zero_homogeneous_coordinate`; `test_wart_reproject_disparity_reads_u_from_the_row_index_4269`,
`test_wart_reproject_disparity_x_varies_with_the_row_and_y_with_the_column_4269`, `test_wart_stereo_rejects_differing_principal_points_4270`, `test_wart_stereo_accepts_a_batch_with_one_positive_tx_fx_4270`,
`test_wart_stereo_tx_zero_collapses_every_point_to_the_origin_4270`, `test_wart_stereo_accepts_a_four_by_four_pair_4270`, `test_wart_stereo_rejects_an_empty_batch_4281`; and the **strict `xfail`**
`test_convention_reproject_disparity_uses_the_column_as_u_4269` — the OpenCV-indexed answer the #4269 repair must produce.

Every wart pin asserts the **current** behavior and says in a comment that it is deleted when the issue is
repaired. The 37 pins of the first push were mutation-checked — 108 live mutations in total, zero unexpected survivors, and each of
the then four strict `xfail`s flipped as a whole claim to a strict `XPASS`; the pins added or rewritten in the review rounds (integer-depth dtype, extra-axis broadcast, subpixel padding, `Q` sign at `W = 0`, the `z = 0` warp split, the float16 `inf` arm) were executed on every cpu dtype and on mps but not mutation-checked. Two pins are rewrites of a first version
that a mutation *survived*: the projection-matrix pin proved nothing about operand order (a pure `x` translation
commutes with `K`, `0.0` apart — it now uses a rotation, `4.0` apart), and the `W = 1` "tiled column" reading
was an artefact of a constant depth (the real structure is an outer product whose diagonal is the column
`depth_to_3d` returns).

## Issues documented here

| --- | --- | --- |
| **#4269** `reproject_disparity_to_3D` unbinds the meshgrid as `v, u`, so `X` comes from the row index and `Y` from the column: kornia's value at `(row, col)` is `cv2.reprojectImageTo3D`'s at `(col, row)` | broken output (candidate idiom 7 — an axis-order slip hidden by symmetric test data) | n/a — single symbol; the method is a one-line forwarder |
| #4270 four `StereoCamera.__init__` guards do not enforce the documented contract (differing `cx` rejected, `torch.all` on the `tx * fx` sign, `tx = 0` accepted, a `shape[:1]` check that can never fire) | semantic contract (idiom 3; idiom 4 on `tx = 0`) | n/a — single symbol, four sites |
| #4271 `unproject_meshgrid`'s documented `(3, 3)` `K` passes the guard and then raises deeper in, with a message naming a shape the caller never passed | semantic contract (idiom 3) | n/a — single symbol |
| #4272 `depth_from_disparity` divides by `disparity + 1e-8`, so a zero disparity returns `5e9`; and "float or tensor" means a `(1,)` tensor only | semantic contract (idioms 1 and 3) | n/a — single symbol |
| #4273 `warp_frame_depth` and `DepthWarper` are the same warp under opposite frame names | semantic contract (idiom 3) | **MERGE IN WINDOW — behavior change**; canonical is `warp_frame_depth`'s `src_trans_dst` naming |
| #4275 public names missing from their rst pages and `__all__`s | API completeness | n/a — single symbol; item 1's depth half is done here |
| #4278 (**closed** by #4298 on `main`, absorbed by the merge) `unproject_meshgrid`'s bare `.squeeze()` destroyed the `W = 1` axis | was broken output on a degenerate-but-valid shape | the three wart pins and the warnings are gone; `test_convention_depth_to_3d_v2_keeps_the_w_one_axis_4278` is the regression pin |
| #4280 (**closed** by #4348 on `main`, absorbed by the merge) `depth_from_plane_equation`'s clamp was `eps * torch.sign(denom)`, a no-op exactly at the singularity | was broken output (idiom 1) | the strict `xfail` became the passing regression pin `test_convention_depth_from_plane_equation_clamps_the_singularity_4280` |
| #4267 five `z`-divides, four answers at `z = 0` — here as the split between `warp_frame_depth` (`project_points` skips the divide at `abs(z) <= 1e-8`) and `DepthWarper` (`cam2pixel` divides by `z + 1e-12`) | semantic contract (idioms 1 and 4) | KEEP SEPARATE per the issue; the warp-level consequence is documented on `DepthWarper` and pinned here |
| #4281 an empty batch (`B = 0`) is rejected by three entry points, each for a different accidental reason | semantic contract, degenerate shapes (idiom 3, surfacing as idiom 4) | n/a — single symbol, three sites; linked to #4115 |

Two issue bodies were corrected from this branch's execution and re-PATCHed: #4269's opening paragraph and
#4278, which gained `depth_to_normals` as its third public consequence. Both are in the notes below.

## Docstring errors fixed

All eleven audit DOCS-WRONG rows, each corrected sentence re-executed:

- `depth_to_3d`'s `.. note::` said it is "an alternative implementation of `depth_to_3d`" — a copy of the note on `depth_to_3d_v2`, describing itself. Replaced with the pointmap-RFC §5 statement (`depth_to_3d_v2` is the newer of the two, both layouts kept pending a coordinated decision, no `DeprecationWarning` is emitted — executed, `catch_warnings(record=True)` returns `[]`).
- `unproject_meshgrid` Args `K (3, 3)` -> `(B, 3, 3)` and Return `(*, H, W, 3)` -> `(B, H, W, 3)`; `warp_frame_depth` Return `(B, 3, H, W)` -> `(B, D, H, W)` (a `D = 5` image returns `(1, 5, 4, 5)`) and its "the warped tensor in the source frame" -> "`image_src` resampled onto the destination pixel grid".
- `DepthWarper` Args `pinholes_dst` -> `pinhole_dst` (the actual parameter name); `forward`'s Shape `(N, C, H, W)` -> `(B, C, H, W)` with the two missing `Input:` lines; and `depth_warp`'s "See :class:`~kornia.geometry.warp.DepthWarper`" pointed at a module that does not exist — it is now the pointer line to `kornia.geometry.depth.DepthWarper`, checked as a real `<a href>`.
- `StereoCamera.reproject_disparity_to_3D` (the method) documented `(B, 1, H, W)` — the layout the shared body rejects with `StereoException`; corrected to `(B, H, W, 1)`, which the free function already had right. And `depth_from_disparity`'s "float/tensor" -> "a python `float` or a tensor of shape `(1,)`" in both Args.
- Beyond the list: `StereoCamera`'s **class** Args called the two inputs "the intrinsic matrix", contradicting `__init__`'s own correct "rectified projection matrix of shape `(B, 3, 4)`" in the same file.

## `docs/` changes

`geometry.depth.rst` gains `.. autofunction:: depth_warp` and `.. autoclass:: DepthWarper` `:members:` — both
were in `__all__` and on no page (#4275, item 1's depth half); all four members render.
`geometry.camera.stereo.rst`, two sentences: the `cx` sentence said a differing `cx` "is taken into account
here", but a right camera with `cx = 5.0` raises `StereoException`, so it now says what the `Q[3, 3]` term is
for, that the constructor rejects such a pair today, and links #4270. The second states how the page's `Q`
relates to the one `StereoCamera` builds; the note below gives the executed relation.

## Docs-only proof and validation

For each modified `kornia/` file, parse both revisions, pop the leading string `Expr` from every `Module` /
`ClassDef` / `FunctionDef` / `AsyncFunctionDef`, `ast.dump`, and diff against `main` `607d4534` — **EMPTY for both**
(`kornia/geometry/depth.py`, `kornia/geometry/camera/stereo.py`). The `tests/` diff has 0 removed lines and the
`kornia/` diff contains no `>>>` or `...` line. `git merge-tree` against that `main` is clean.

- pinned Ruff (`0.16.6`) check and format clean; the 6 doctests of the two modules pass
- clean `sphinx -W` build (`rm -rf docs/build` first) — `build succeeded`, zero warnings; `geometry.depth.html` renders `depth_warp` and all four `DepthWarper` members, and every `PinholeCamera` / issue link in the two pages is a real anchor.
- `tests/geometry/test_depth.py tests/geometry/camera/test_stereo.py`: cpu `--dtype=float32,float64` **233 passed, 4 xfailed**; cpu float16 with `--xfail-known-failures --known-failure-profile=cpu-float16` 111 passed / 3 skipped / 7 xfailed and cpu bfloat16 100 / 1 / 20 (the xfails are the manifest lines plus the two strict `xfail`s); `--device=mps --dtype=float32` (with `PYTORCH_ENABLE_MPS_FALLBACK` unset) 114 passed / 5 skipped / 2 xfailed; the pins touched in the review rounds also on mps float16. Hosted CI was fully green at `26ddb21e`.

`depth_to_3d` / `depth_to_3d_v2` (**MERGE IN WINDOW**, re-executed: permute-equal under an asymmetric `K`,
`H != W`, a batch-2 differing `K`, f32/f64 and integer depth — and, since #4298, permute-equal at `W = 1` as well) · `warp_frame_depth` / `DepthWarper` (**MERGE IN WINDOW — behavior change**, #4273) ·
`depth_warp` / `DepthWarper` (**KEEP SEPARATE — intentional variant**: byte-identical functional wrapper,
`torch.equal` in every dtype cell) · `StereoCamera.reproject_disparity_to_3D` / the module-level function
(**KEEP SEPARATE — intentional variant**: a one-line forwarder; the docstring and export gaps are single-symbol).

## Notes for reviewers

- **The rst page's `Q` is the class's `Q` under the page's own `tx`.** The page's `P_1` carries `fx * tx` in its last column, so its `tx = P_1[0, 3] / fx` is negative on every rig the constructor validates as intended (the guard rejects only a positive product, so `tx = 0` and a batch with one positive product pass, #4270), and evaluating the page's matrix at that value gives the class's `Q` bit for bit (`torch.equal`, fx = 100, fy = 50, cx = 4, cy = 3, P_right[0, 3] = -50). The class's `tx` attribute is the negation of that symbol; substituting it into the page's matrix gives neither `Q` nor `-Q`, because the last row carries no `tx`. An overall sign cancels in the homogeneous divide wherever it runs, so `Q` and `-Q` return the same points except at `W = 0`, where the numerator comes back unchanged and the two are opposite. Both the page and the `StereoCamera` block now say this.
- **A one-line correction went to the 5a branch from this surface's execution** (`04573e7f`, on #4294): that anchor claimed a `(*, 3, 3)` camera matrix for the whole functional API, and executing all six named functions shows it is false for four of them — `depth_to_3d`, `depth_to_3d_v2`, `unproject_meshgrid` and `depth_to_normals` all need a batched `(B, 3, 3)`. The sentence there is now split accordingly.
- **The `_RealTestData` stereo pin passes only because its layout is transposed, and it is left untouched**, per the batch rule that a documentation PR is additive. Its stored literals are correct OpenCV output for a one-row, ten-column strip (hand `cv2` formula: 8.8e-5 away); the fixture's `permute` stores that strip as ten rows of one column, which the swap cancels. #4269's fix re-lays the fixture and keeps every number (as #4366 does); the anchor's warning says so, and the strict `xfail` beside it is the OpenCV-indexed answer that repair must produce.
- **#4269's body sentence about square images was corrected after execution.** The old claim — that the error cancels when `fx == fy` and `cx == cy` on a square image — is false: the relation is a pure index transpose, so only the diagonal agrees (`0.2` apart on a `5 x 5` rig, `0.3` on a `7 x 7`). What execution does settle is `Q`'s own focal swap: rows 0/1 carry `fy`/`fx` against a divide by `-fy * disparity`, which *compensates* for the scaling, so the `unbind` change alone is the whole fix.
- **`unproject_meshgrid` documents `(B, 3, 3)`, not `(*, 3, 3)`.** A bare `(*, 3, 3)` is false in both directions: a rank-2 `K` raises, and a non-singleton extra leading dimension raises `RuntimeError`. Only *singleton* extras pass through — `(2, 1, 3, 3)` -> `(2, 1, H, W, 3)` — and that is pinned. #4271 is about the guard and its message, which stay open.
- **`warp_frame_depth` and `DepthWarper` split outright at `z = 0`.** With `K = I`, unit depth and a `(+1, 0, -1)` translation every destination pixel reaches `z = 0` in the other view: `project_points` skips the divide, so `warp_frame_depth` returns the image shifted by one column (`[2, 3, 0]`), while `cam2pixel` divides by `z + 1e-12` and `DepthWarper` sends the same pixel to a `1e12` coordinate (what `grid_sample` then returns is backend-dependent: zeros on cpu float32, nan in float16, `1e24` in one row on mps, and torch 2.5.1's aarch64 CPU kernel segfaults on it, pytorch/pytorch#24823, so the pin asserts the grid and never samples through it). The `DepthWarper` warning states the agreement for `z` away from zero only and links #4267, which already catalogues those two divide policies.
- **`warp_frame_depth` and `DepthWarper` are not "never bitwise equal".** On an integer-`K` identity fixture they are `torch.equal`. What execution supports over six configurations is that the two build the sampling *grid* by different routes (`depth_to_3d_v2` + `project_points` versus `pixel2cam` + `cam2pixel`), and wherever the two grids come out bit-identical so do the images. They agree at float32/float64 tolerance (`4.02e-07` / `2.90e-12` residual on the pin's fixture); float16 and bfloat16 are **skipped with measured reasons** (`1.46e-03`, `1.95e-02` — outside `assert_close`'s tolerances), and the byte-identity sibling `depth_warp` == `DepthWarper` runs in every dtype.
- **`depth_to_normals` was found to be a third #4278 consequence** while writing the docstrings — a `(1, 1, 4, 1)` depth returned a `(1, 3, 4, 4)` normal map. The issue body gained that row; #4298 repaired all three and the wart pins left with the merge from `main`. `depth_to_normals` at `H = 1` is device-dependent (all-zero on cpu, `[-1, -1, -1]` on mps, `nan` in float16); that is unfiled, and the pin deliberately asserts only its shape.
- **Three Convention sentences are deliberately unpinned** and were executed twice each instead: `depth_from_disparity`'s negative-disparity bullet, `depth_to_3d`'s `normalize_points` ray-length reading, and the `StereoCamera` `1 / fx` scaling clause. Pins for them are a follow-up. One pin, `test_convention_projection_matrix_is_k_dst_times_dst_trans_src`, reads the private `warper._dst_proj_src` — the only way to state the class docstring's own `P_src^{dst} = K_dst * T_src^{dst}` equation as an assertion.
- **`depth_from_disparity` in float16.** `zeros(..., dtype=float16) + 1e-8` is exactly `0`, so the "large finite depth" wart does not occur there: the divide is by zero and the result is `inf`, on cpu and mps. The warning says so and the float16 arm of the #4272 pin asserts the `inf` instead of skipping.
- Three sibling sub-batch PRs each add a changelog bullet as the first entry under `## Unreleased` -> `### Added`; whichever merges later rebases with a one-hunk conflict.

## Review rounds

- `1a96bfd1` — first review (issuecomment-5558770187): approve after the integer-depth sentence names `camera_matrix`'s dtype.
- `26ddb21e` — merge from `main` plus the Codex round-1/round-2 asks: the four documentation findings (matching-width broadcast, integer-depth promotion, subpixel zero padding, `Q`/`-Q` at `W = 0`) corrected with matching pins; the #4278/#4280 warts retired after #4298/#4348 landed.
- `089daad9` — the two round-3 asks: the `z = 0` warp split (#4267) and the float16 `inf` of a zero disparity, plus provenance trailers for the round-2 pins.
- `2bd698d5` — the #4267 split pin no longer calls `grid_sample` on the `1e12` grid: torch 2.5.1's aarch64 CPU kernel segfaults once one normalized coordinate exceeds `2**31` (killed both macOS 2.5.1 legs at `c77ec04c`).
- `d344bd0b` — round 6 (a second `/code-review` pass at `2bd698d5`, every finding re-executed): the real-data fixture's literals are correct OpenCV output and only its layout is transposed; the `tx * fx` guard rejects only a positive product; `depth_to_3d_v2` Args/Return are `(*, 3, 3)` / `(*, H, W, 3)` because the leading dims broadcast; `unproject_meshgrid`'s singleton pass-through is filed under #4271, not a convention; the two bare `:meth:` xrefs resolved to the module-level function and are now fully qualified; `_check_disparity_tensor`'s docstring says `(B, H, W, 1)` and its message is #4374; the #4273 pin list names the wart pin only; `depth_warp` exposes `align_corners` only; the #4281 pin comment names `transform_points`.
- `c77ec04c` — round 4 (a `/code-review` pass, every finding verified by execution): the page's `Q` and the class's `Q` are the same matrix under the page's own `tx = P_1[0, 3] / fx` (the class's `tx` attribute is its negation); the empty-batch `ZeroDivisionError` comes from `transform_points`, not the pixel normalization; `depth_to_3d_v2`'s Args/Return say `(B, 3, 3)` / `(B, H, W, 3)` and the `xyz_grid` path, which never reads `camera_matrix`, is pinned; the `Q`-sign pin uses a `1 x 1` map so it no longer encodes the #4269 swap; the #4271 strict `xfail` XPASSes under either repair outcome; the #4280 regression pin asserts the sign; `TestDepthWarper` in `test_depth.py` is `TestDepthWarperConventions` (no node-name collision with `test_depth_warper.py`); ledger residue dropped from test comments.

Posted on behalf of @ducha-aiki by Claude (Opus 5); refreshed after the review rounds by Claude (Claude Fable 5.1).


ducha-aiki:
## Review of #4317 (batch 5c: depth and stereo) at `1a96bfd1`

**Verdict: approve once one sentence is corrected.** All 44 checks are green, every Convention claim I re-executed on the branch worktree (torch 2.14.0) holds, and the three headline findings reproduce exactly as documented: `reproject_disparity_to_3D` reads `X` from the row index, `unproject_meshgrid` loses the `W = 1` axis, and `warp_frame_depth` / `DepthWarper` name the depth frame oppositely while agreeing to tolerance and never bitwise. The rst arithmetic also checks out: the page's `Q` at `tx -> -tx` is the class's `Q` entry for entry (`[0, 0] = -50`, `[0, 3] = +200`, `[2, 3] = -5000`, last row `[0, 0, -fy, 0]` on both).

### Required

1. **`depth_to_3d` Convention: "an integer depth map is promoted: an int64 depth returns a float32 point cloud".** The promotion target is `camera_matrix`'s dtype, not float32. Executed: int64 depth with a float64 `K` returns float64 from both `depth_to_3d` and `depth_to_3d_v2`, and with a float16 `K` returns float16. `test_convention_integer_depth_returns_a_float32_point_cloud` only ever builds a float32 `K`, so it cannot see this. Say "promoted to the dtype of `camera_matrix`" and add a float64-`K` arm (rename the test accordingly).

### P2

2. Copilot's note on `unproject_meshgrid`: the Returns line says `(B, H, W, 3)` while the block explains that singleton leading dims pass through as `(2, 1, H, W, 3)`. The block is the more precise statement; fine as is, or add "plus any singleton leading dims" to Returns.
3. Six docstrings point at the Convention block on `PinholeCamera`, which exists only in #4294. Merge 5a first (this PR also corrects `depth.py`'s `(3, 3)` Args that 5a's anchor already contradicts, so the two belong together).
4. Cross-cutting for 5a to 5d: pin comments cite "audit labels 5c-st-16" and "this worktree"; nothing in this repository resolves them. Keep the snippet and a commit hash, drop or spell out the labels.
5. `test_convention_depth_from_plane_equation_clamps_the_singularity_4280` puts a `pytest.skip` inside a strict xfail for float16. That works (skip wins), but a one-line `if dtype == float16: pytest.skip` before the mark's effect is easy to misread as an XPASS risk; a comment saying "skip wins over strict xfail" would save the next reader a check.

Checked and clean: `DepthWarper.grid` is `(0, 0, 1)` at pixel `(0, 0)` and `(4, 3, 1)` at `(3, 4)`; `warp_frame_depth` really calls `grid_sample(..., align_corners=True)` with default padding; the `depth_from_disparity` `(1,)`-only rule and the python-int rejection; the `depth_from_plane_equation` `sign(0) = 0` no-op; the `B = 0` `ZeroDivisionError`; `depth_warp` bitwise equal to the class on both fixtures; the new `geometry.depth.rst` entries; CHANGELOG placement. GitHub currently reports the PR as `CONFLICTING` against today's `main` although a local `merge-tree` is clean; rebase so a real CI run backs the green.

kornia/geometry/depth.py
```diff
@@ -63,17 +63,56 @@ def unproject_meshgrid(
...
+          function's ``xyz_grid`` cache.
+        - the pixels are the integer pixel centres that :func:`~kornia.geometry.grid.create_meshgrid`
+          enumerates, described in the Convention block on
+          :class:`~kornia.geometry.camera.pinhole.PinholeCamera`: pixel ``(0, 0)`` is centred at ``(0, 0)``.
+          ``camera_matrix`` is the :math:`(3, 3)` ``[[fx, 0, cx], [0, fy, cy], [0, 0, 1]]`` of that grid and
+          there are no extrinsics, so the rays are in the **camera** frame.
+        - ``camera_matrix`` is batched: the supported form is :math:`(B, 3, 3)`, which returns
+          :math:`(B, H, W, 3)`. Extra **singleton** leading dimensions pass through unchanged -- a
+          :math:`(2, 1, 3, 3)` intrinsics returns :math:`(2, 1, H, W, 3)` -- while a non-singleton extra
+          leading dimension raises ``RuntimeError``, and an unbatched :math:`(3, 3)` raises ``ShapeError``
```
ducha-aiki: [P2] Describe accidental broadcasting for extra camera dimensions

A non-singleton extra dimension does not reliably raise: `unproject_meshgrid(2, 3, torch.eye(3).expand(2, 3, 3, 3).clone())` succeeds and returns `(2, 1, 2, 3, 3)`. The extra camera axis broadcasts against the pixel width, so callers can silently get a camera-per-column interpretation instead of the promised exception. Please describe extra leading dimensions as unsupported with shape-dependent accidental broadcasting, and include a matching-width case in the pin; the existing `(2, 2, 3, 3)` fixture only exercises a mismatch.

kornia/geometry/depth.py
```diff
@@ -150,8 +221,26 @@ def depth_to_3d(depth: torch.Tensor, camera_matrix: torch.Tensor, normalize_poin
...
+          where that function loses the width axis
+          (`#4278 <https://github.com/kornia/kornia/issues/4278>`_).
+        - ``u`` and ``v`` are the column and the row of the integer pixel centres that
+          :func:`~kornia.geometry.grid.create_meshgrid` enumerates, described in the Convention block on
+          :class:`~kornia.geometry.camera.pinhole.PinholeCamera`: pixel ``(0, 0)`` is centred at ``(0, 0)``.
+          ``camera_matrix`` is the :math:`(3, 3)` ``[[fx, 0, cx], [0, fy, cy], [0, 0, 1]]`` of that grid and
+          there are no extrinsics, so the points are in the **camera** frame.
+        - ``normalize_points=True`` reads ``depth`` as the Euclidean ray length from the camera centre instead
+          of as ``z``, so the returned point has that norm rather than that ``z``.
+        - an integer ``depth`` map is promoted: an ``int64`` depth returns a ``float32`` point cloud.
```
ducha-aiki: [P2] Qualify integer-depth promotion by the intrinsics dtype

The new unconditional float32 guarantee is false with float64 intrinsics: `depth_to_3d(torch.ones(1, 1, 2, 3, dtype=torch.int64), torch.eye(3, dtype=torch.float64)[None]).dtype` is `torch.float64`. The new test fixes the camera to float32 in every dtype cell, so it cannot detect this. Please state that the result follows arithmetic promotion involving the camera matrix, and qualify float32 as the tested float32-intrinsics case. Otherwise callers cannot rely on this documented output dtype when combining integer sensor depth with double-precision calibration.

kornia/geometry/depth.py
```diff
@@ -288,6 +421,44 @@ def warp_frame_depth(
...
+          (``image_src``), and ``src_trans_dst`` maps destination-frame points into the source frame. With
+          ``fx = fy = 1`` and a unit depth, a :math:`+1` translation in ``x`` therefore samples ``image_src``
+          one pixel to the right of each destination pixel.
+        - the unprojection is :func:`~kornia.geometry.depth.depth_to_3d_v2` and the reprojection
+          :func:`~kornia.geometry.camera.perspective.project_points`, so the pixel grid, the camera frame and
+          the two readings of ``depth`` selected by ``normalize_points`` are the ones stated in that function's
+          Convention block.
+        - the sampling is ``grid_sample`` with ``align_corners=True`` and the default
+          ``padding_mode="zeros"``, both baked in: the function exposes neither, so a destination pixel whose
+          source pixel falls outside ``image_src`` comes back as 0.
```
ducha-aiki: [P2] Account for partial border interpolation with zero padding

Zero padding does not make every sample outside the image zero: the default bilinear interpolation blends in-bounds neighbors with padded zeros. For an all-ones `(1,1,2,3)` image, unit depth, identity intrinsics and `src_trans_dst[0,0,3] = 0.25`, the last column samples at x=2.25 and returns 0.75, not 0. Please describe the zero extension and interpolation rather than promising an out-of-bounds zero mask; the current integer-translation pin misses this common subpixel case.

kornia/geometry/camera/stereo.py
```diff
@@ -51,9 +51,72 @@ def __init__(self, msg: str, *args: Any, **kwargs: Any) -> None:
...
+          shape :math:`(B, 3, 4)`: ``[[fx, 0, cx, 0], [0, fy, cy, 0], [0, 0, 1, 0]]`` for the left camera, and
+          the same matrix with ``-tx * fx`` in the last column for the right one. The constructor requires the
+          two to be equal outside that last column.
+        - the baseline is read back from that column as ``tx = -P_right[0, 3] / fx``, and :attr:`Q` is built
+          from ``fx``, ``fy``, ``cx_left``, ``cy`` and that ``tx``. Note which focal length sits in which row:
+          ``Q[0, 0]`` carries ``fy`` and ``Q[1, 1]`` carries ``fx``, while the homogeneous divide is by
+          ``-fy * disparity``, so the two cancel and the first output coordinate ends up scaled by ``1 / fx``
+          and the second by ``1 / fy``.
+        - a point cloud is a homogeneous transform by :attr:`Q` followed by the divide by ``W``, so an
+          overall sign on :attr:`Q` cancels: :attr:`Q` and its negation return the same points. The matrix
```
ducha-aiki: [P2] Qualify Q sign equivalence at zero homogeneous depth

The implementation does not always divide by the homogeneous coordinate: `convert_points_from_homogeneous` leaves the numerator unchanged when `abs(W) <= 1e-8`. On the documented fx=fy=100, cx=4, cy=3, tx=0.5 rig, a zero `(1,1,1,1)` disparity returns `[200,150,-5000]` with `Q` and `[-200,-150,5000]` with `-Q`. Zero disparity is accepted, so the new unconditional equivalence can lead callers to change their output by flipping the matrix sign. Please qualify the claim to coordinates where the homogeneous division is performed and describe the singular case; update the same claim in `geometry.camera.stereo.rst` as well.

ducha-aiki (commented):
Found four reproducible inaccuracies in the new depth/stereo convention documentation; please correct these statements and their coverage before merging. These findings concern newly added documentation, not the existing runtime defects this PR intentionally records.

Validation on the PR checkout: both changed library files have identical ASTs to the base after removing docstrings. The two focused test files on CPU float32/float64: 205 passed, 8 xfailed. New convention/wart pins on CPU float16/bfloat16: 62 passed, 5 skipped, 7 xfailed; on MPS float32: 33 passed, 4 xfailed. CUDA and the docs build were not rerun.

ducha-aiki (commented):
Re-reviewed at `2332b5a1`. The merge from main introduces an additional integration blocker: the new wart tests still assert defects that main has now fixed.

On this exact checkout, the CPU float32 selection `tests/geometry/test_depth.py tests/geometry/camera/test_stereo.py -k "4278 or 4280"` produces **7 failures**, including two strict XPASSes. The #4278 cases now preserve the width axis, and the #4280 grazing-ray case returns finite `2e8`; the documentation and assertions still promise the former broken behavior. Please invert the repaired wart assertions, remove the resolved strict-xfail markers, and update the corresponding #4278/#4280 warnings and release-note wording. Keep dtype-specific limitations where they still apply.

The four documentation findings in the earlier review are also unchanged by this merge: accidental matching-width camera broadcasting, integer-depth promotion with float64 intrinsics, fractional zero-padding interpolation, and Q/-Q behavior when homogeneous division is skipped. Those still need the requested prose and regression coverage.

Validation used the existing project environment through Pixi, with `kornia.__file__` verified inside the isolated PR checkout. No runtime edits were made.

ducha-aiki (commented):
Re-reviewed the follow-up at `26ddb21e`. The findings in my earlier review are now resolved: the four inaccurate statements have matching corrected coverage, the repaired #4278/#4280 wart assertions and strict xfails are retired, and the public depth documentation no longer describes those repaired defects.

Validation on this exact head: full depth/stereo files on CPU float32/float64 passed (231 passed, 4 expected xfails); convention/wart selections on CPU float16/bfloat16 passed (61 passed, 5 skipped, 4 expected xfails) and on MPS float32 passed (33 passed, 2 expected xfails). `git diff --check` is clean. Both changed library files have identical runtime ASTs to current main after stripping docstrings. Imports resolved to the isolated PR checkout.

No remaining actionable findings from this re-review. The new head still needs hosted CI completion and an eligible external approval. Coordinate with #4366: whichever change lands second must update the #4269 stereo axis-swap pins and warning. CUDA and the docs build were not rerun locally.

kornia/geometry/depth.py
```diff
@@ -342,8 +466,54 @@ class DepthWarper(nn.Module):
...
+        :func:`~kornia.geometry.depth.warp_frame_depth` performs the same warp under the **opposite** naming.
+        The frame this class calls ``src``, the one holding the depth, is that function's ``dst``
+        (``depth_dst``), and the image this class takes as ``patch_dst`` is that function's ``image_src``;
+        reading "dst" as "dst" across the two APIs gives the inverse warp. The two also differ arithmetically:
+        they build the sampling grid by different routes -- :func:`~kornia.geometry.depth.depth_to_3d_v2` and
+        :func:`~kornia.geometry.camera.perspective.project_points` here, ``pixel2cam`` and ``cam2pixel`` there
+        -- so the grid, and with it the resampled image, can differ in the last bits. Where the two grids come
+        out bit-identical, so do the images. In float32 and float64 the two agree at the working dtype's
+        tolerance; in float16 and bfloat16 the gap is wider than that tolerance, which is why the pin below
+        states the claim for the two single- and double-precision dtypes only. Tracked as
```
ducha-aiki: [P2] Exclude singular projections from the warp-equivalence guarantee

The difference can be much larger than the dtype tolerance when transformed points reach `z = 0`. With `image = arange(1., 7.).reshape(1, 1, 2, 3)`, unit depth, identity intrinsics and a rigid transform translating by `(1, 0, -1)`, `warp_frame_depth` returns rows `[2, 3, 0]` and `[5, 6, 0]`, while `DepthWarper` with the corresponding camera pair returns all zeros, in both float32 and float64. `project_points` skips the homogeneous division near zero; `cam2pixel` divides by `z + 1e-12`. Please qualify the last-bit/tolerance claim to nonsingular projections and document this difference so callers do not assume that switching APIs preserves these pixels.

kornia/geometry/depth.py
```diff
@@ -593,10 +777,30 @@ def depth_from_disparity(
...
+          ``ShapeError``, and a python ``int`` is rejected by the type check, so one value is shared by the
+          whole batch.
+        - ``disparity`` is :math:`(*, H, W)` and the result has its shape. Its sign is not checked, so a
+          negative disparity gives a negative depth.
+
+    .. warning::
+        The epsilon is inside the arithmetic -- the divisor is ``disparity + 1e-8`` -- instead of selecting a
+        branch, so a zero disparity, which is what a stereo matcher writes where it found no match, returns a
+        large finite depth rather than ``inf``, and that value is a function of the epsilon rather than of the
+        camera. Tracked as `#4272 <https://github.com/kornia/kornia/issues/4272>`_ and pinned by
```
ducha-aiki: [P2] Qualify finite zero-disparity output by dtype

This warning promises a finite result without the dtype limitation already acknowledged by the new test. `depth_from_disparity(torch.zeros(1, 1, 2, 3, dtype=torch.float16), 0.5, 100.0)` returns `inf` throughout: the `1e-8` denominator underflows to zero in float16. The same call in float32 returns about `5e9`. This matters when callers design invalid-depth handling from the documented zero-disparity behavior. Please state the float16 exception and restrict the finite-result claim to cases where the denominator and result are representable, as the new plane-depth documentation already does.

ducha-aiki (commented):
Reviewed `26ddb21e`. The earlier four documentation findings and the #4278/#4280 integration failures are addressed. I found two additional numerical guarantees in the new documentation that need qualification; both are reproduced in the inline comments. These concern the new prose, not a request to change runtime behavior in this documentation PR.

Validation on the isolated PR checkout, using the project environment through Pixi and verifying the imported path:

- Full depth/stereo test files, CPU float32/float64: 231 passed, 4 expected xfails.
- Convention/wart tests, CPU float16/bfloat16: 61 passed, 5 skipped, 4 expected xfails.
- Convention/wart tests, MPS float32/float16: 62 passed, 4 skipped, 4 expected xfails.
- The affected modules' doctests: 6 passed.
- Runtime ASTs match the base after stripping docstrings; `git diff --check` is clean.

CUDA and a local rendered documentation build were not checked.

ducha-aiki:
## Re-review of #4317 (batch 5c: depth and stereo) at `089daad9`

**Verdict: APPROVE.** The head is the round-2 delta `26ddb21e` (main merged, the four round-1 documentation findings corrected, the #4278/#4280 warts retired once #4298 and #4348 landed) plus one commit I pushed, `089daad9`, which resolves the two findings still open from the 18:46 review. Every claim below was re-executed on a worktree at this head (torch 2.14.0); the whole diff against `main` `607d4534` was re-read, not just the delta.

### What `089daad9` changes
1. **`DepthWarper` warning.** "In float32 and float64 the two agree at the working dtype's tolerance" now holds only while the transformed `z` stays away from zero. At `z = 0` the two split outright: `project_points` skips the divide at `abs(z) <= 1e-8`, so `warp_frame_depth` returns image content shifted by one column (`[[2, 3, 0], [5, 6, 0]]` on the reported fixture), while `cam2pixel` divides by `z + 1e-12` and sends the same pixel to a `1e12` coordinate. Linked to #4267, which already catalogues exactly those two divide policies, and pinned by `test_wart_warp_frame_depth_and_depth_warper_split_at_zero_transformed_depth_4267`. The pin asserts `warp_frame_depth`'s value and the grid coordinate, not what `grid_sample` makes of a coordinate that far out: that answer is backend-dependent (zeros on cpu float32/float64/bfloat16, `nan` on cpu float16, `1e24` in row 1 on mps float32). A `z = 0.5` arm checks the two still agree one half unit away.
2. **`depth_from_disparity` warning.** The finite `5e9` is a float32/float64/bfloat16 fact; in float16 the `1e-8` itself rounds to zero (`zeros(1, 1, 1, 1, dtype=float16) + 1e-8 == 0`), so the divide is by zero and the result is `inf`, on cpu and mps alike. The float16 arm of `test_wart_zero_disparity_gives_a_finite_depth_4272` now asserts that `inf` instead of skipping.
3. The extra-axis broadcast pin is now listed in the #4271 warning and carries the same "Pins the CURRENT behavior … delete when repaired" trailer and generation snippet as its siblings; the subpixel-padding and `Q`-sign pins gain their snippets. One float16 trap found on the way: a literal `1e9` compared against a float16 grid is itself `inf`, and `inf > inf` is `False`, so that comparison is done in float32.

### Verified at this head
- Runtime ASTs of both `kornia/` files identical to `main` after stripping docstrings; the `tests/` diff has zero removed lines; `git merge-tree` against `main` clean; pinned Ruff 0.16.6 check and format clean.
- `tests/geometry/test_depth.py` + `tests/geometry/camera/test_stereo.py`: cpu float32/float64 **233 passed, 4 xfailed**; cpu float16 with `--known-failure-profile=cpu-float16` 111 passed / 3 skipped / 7 xfailed; cpu bfloat16 100 / 1 / 20 (manifest lines plus the two strict `xfail`s); mps float32 114 passed / 5 skipped / 2 xfailed (`PYTORCH_ENABLE_MPS_FALLBACK` unset); the pins touched in the review rounds also on mps float16 (7 passed). The 6 doctests of both modules pass.
- Clean `sphinx -W` build after `rm -rf docs/build`: zero warnings; `geometry.depth.html` renders `depth_warp` and all four `DepthWarper` members; every `PinholeCamera` and `#4270` link in the two pages is a real anchor.
- Round-1/2 asks all resolved by execution: int64 depth follows `camera_matrix`'s dtype (float16, float32 and float64 each verified; the pin now runs in every dtype cell); a `(2, 3, 3, 3)` camera broadcasts a different `cx` onto each column, `(2, 1, 2, 3, 3)` out; a quarter-pixel translation returns `0.75` at the last column; `Q` vs `-Q` are `torch.equal` at `d = 10` and `d = 0.01` in both dtypes and exact opposites at `d = 0` (`[200, 150, -5000]`).
- `unproject_meshgrid(4, 1, K)` is `(1, 4, 1, 3)` and the grazing ray returns `2e8`, so the #4278/#4280 retirement matches `main`; the AGENTS grep leaves no `#4278`/`#4280` hit in `kornia/`, and the two `tests/` hits are the regression pins.

### Merge notes
- **#4366** (fixes #4269) is currently `CONFLICTING` and touches `stereo.py` and `test_stereo.py`. Whichever lands second drops the two `_4269` wart pins, turns the strict `xfail` into a plain pin, rewrites the `StereoCamera` warning and re-derives `_RealTestData`.
- The pointer lines cite the `PinholeCamera` Convention block that #4294 adds; the xrefs resolve on today's `main` either way, so order is free, but #4294 is approved and should go first.
- PR body refreshed to this head (36 pins = 22 convention incl. 2 strict `xfail`s + 14 wart; 8 warnings; #4278/#4280 rows marked closed; #4267 row added; validation numbers current).

ducha-aiki:
### Addendum: round 4 at `c77ec04c` (approval stands)

A `/code-review` pass over `26ddb21e` returned ten findings; I re-executed each on the worktree and pushed the fixes that survived as `c77ec04c`. Three were the items already closed by `089daad9`. The rest:

- **`Q` framing was backwards, on the page and in the class block.** Under the page's own definition (`P_1[0, 3] = fx * tx`, so `tx = P_1[0, 3] / fx`, which the constructor requires to be negative), the class builds the page's matrix *verbatim*: `torch.equal` at fx = 100, fy = 50, cx = 4, cy = 3, P_right[0, 3] = -50. What differs is only the symbol: the class's `tx` attribute is that quantity's negation. Substituting the attribute's value into the page's matrix gives neither `Q` nor `-Q` (the last row carries no `tx`), which is what the old "opposite baseline" sentence was groping at. Both texts now say it that way.
- **#4281 attribution.** The empty-batch `ZeroDivisionError` is raised in `transform_points` (`linalg.py:224`, `0 // 0` in the batch-repeat count), before any pixel normalization runs. The warning and the pin comment now name it.
- **`depth_to_3d_v2` contradicted itself**: the Convention block said a batch dimension is required while Args/Return still said `(*, 3, 3)` / `(*, H, W, 3)`; and the "rejected inside `unproject_meshgrid`" sentence is only true without `xyz_grid`. With `xyz_grid`, `camera_matrix` is never read, so a bare `(3, 3)` is silently accepted (`(1, 3, 5, 3)`, bit-equal to the batched call). Args/Return corrected, the sentence conditioned, and the bypass pinned (`test_convention_xyz_grid_bypasses_the_camera_matrix`).
- **`#4272` wording**: the finite value is `baseline * focal / 1e-8`, set by the epsilon *as much as* by the camera, not "rather than".
- **Pin hygiene.** The `Q`-sign pin used a `(1, 1, 2, 1)` map whose expected literal encoded the #4269 swap, so it would go red under #4366 without naming the issue; it now uses a `1 x 1` map, which reads the same under either index order. The #4271 strict `xfail` wrapped `pytest.raises`, so a repair that *accepts* `(3, 3)` would have left it `DID NOT RAISE` forever; it now XPASSes under either outcome. The #4280 regression pin compared `.abs()`, which cannot see the documented `+eps`; it now asserts `+2e8` for both normal signs. The #4278 regression pin also covers `depth_to_normals` and `warp_frame_depth` shapes at `W = 1`, which the retired wart pins had covered. The normals pin is named for what it asserts (`+z` points away from the camera), and the second `TestDepthWarper` class is `TestDepthWarperConventions` so it does not share a node name with `tests/geometry/test_depth_warper.py`. Round-1's cross-cutting ask is now actually done: no `audit label`, `ledger` or "on this base" text survives in the test comments.

Not changed, on purpose: the fold of the two regression pins into #4298/#4348's own tests (the AGENTS naming convention wants the issue number in the pin's name, and the pins now assert something those tests do not, the sign and the W = 1 siblings); the `.clone()` and duplicate-call cleanups (67 ms of added test time across 70 cells).

Verified at `c77ec04c`: cpu float32/float64 **235 passed, 4 xfailed**; cpu float16 profile 112 / 3 skipped / 7 xfailed; bfloat16 101 / 1 / 20; mps float32 115 / 5 / 2; touched pins on mps float16 10 passed; the two strict `xfail`s still XFAIL; 6 doctests; docstring-stripped ASTs identical to `main`; zero removed lines in `tests/` against `main`; pinned Ruff clean; `sphinx -W` zero warnings. PR body refreshed (37 pins = 23 convention incl. 2 strict `xfail`s + 14 wart).

ducha-aiki:
### Addendum: round 5 at `2bd698d5` (approval stands, CI was red at `c77ec04c`)

The `Tests on CPU (PR)` run for `c77ec04c` failed both **macOS torch 2.5.1** legs (3.11 and 3.12) with `Fatal Python error: Segmentation fault` inside `tests/geometry/test_depth.py`, and the collector went red with them. The faulthandler trace names the pin I added in `089daad9`, `test_wart_warp_frame_depth_and_depth_warper_split_at_zero_transformed_depth_4267`, at its `singular(depth, image)` line, i.e. `DepthWarper.forward` → `F.grid_sample`. The `26ddb21e` run had passed those legs, so the regression was mine.

Reproduced on an arm64 Mac with a torch 2.5.1 / Python 3.11 venv importing this worktree (rc -11 = SIGSEGV):

| `grid[..., 0]` | `grid[..., 1]` | torch 2.5.1 aarch64 CPU |
|---|---|---|
| `2147483000.0` | `-1.0` | zeros |
| `2147483648.0` (2^31) | `-1.0` | **segfault** |
| `1e12` | `-1.0` | **segfault** |
| `-1e12` | `-1.0` | zeros |
| `1e12` | `1e12` | zeros |
| `inf` | `-1.0` | **segfault** |
| `nan` | `-1.0` | nan |

So the vectorized 2D CPU kernel dies when exactly one normalized coordinate overflows int32 while the other is in range, which is precisely the grid `DepthWarper` builds at transformed `z = 0` (`x` of order `1e12`, `y` in `[-1, 2e12]`). The x86 and Windows 2.5.1 legs and every ≥ 2.9.1 leg are fine. Upstream this is the open pytorch/pytorch#24823 (expanded from #19826, which diagnosed the vectorized-kernel crash).

Fix pushed as `2bd698d5`: the pin drops the `grid_sample` call, whose answer was backend-dependent anyway (zeros / nan / `1e24` per the comment), and keeps the two assertions that actually carry the #4267 split, `warp_frame_depth`'s `[[2, 3, 0], [5, 6, 0]]` and `DepthWarper`'s `> 1e9` grid coordinate. The comment records the crash boundary and the upstream issue. Verified after the change:

- `tests/geometry/test_depth.py` on torch 2.5.1 arm64, cpu float32+float64: 183 passed, 2 xfailed (was: segfault)
- same file on torch 2.14.0 cpu float32+float64: 183 passed, 2 xfailed
- the changed node on torch 2.14.0: cpu float16 / bfloat16 with the CI profiles pass; mps float32+float16 pass
- pre-commit on the file green; PR body updated (bullet on the #4267 split and the history list)

Not a CI surface but for the record: on torch 2.5.1 the same node fails in cpu float16/bfloat16 with `grid_sampler_2d_cpu not implemented for Half`, as do 32 other warp tests in the file; that is the 2.5.1 kernel gap, not the pin.

Library-level consequence, noted on #4267: on torch 2.5.1 aarch64 a `DepthWarper` whose transformed points reach `z ≈ 0` does not return garbage, it takes the process down.

ducha-aiki:
### Addendum: round 6 at `d344bd0b` (approval stands)

A second `/code-review 4317 high` pass, run against `2bd698d5`, returned ten findings. I re-executed every one on the worktree before touching anything; nine are real and are fixed in `d344bd0b`, one is a style call I am reporting rather than applying. None changes a test outcome: the library edits are AST-identical to `2bd698d5` with docstrings stripped, and every pin passes as before.

**Fixed (each re-verified):**

- **The #4269 warning blamed the wrong thing.** It said the real-data test "stores the swapped values as its ground truth, so a fix has to update that test". Executed on the fixture as stored (`float64`): current code vs the stored cloud 8.8e-5; corrected `u, v` unbind on the stored layout 0.726; corrected unbind with the disparity re-laid as `(1, 1, 10, 1)` and the cloud as `(1, 1, 10, 3)` back to 8.8e-5; the hand `cv2.reprojectImageTo3D` formula for a one-row, ten-column strip vs the stored numbers 8.8e-5. The literals are correct OpenCV output; only the fixture's `permute` is transposed. Reworded in the `StereoCamera` warning and the strict-xfail comment. The #4269 body's "re-derive the point cloud" item would have replaced a real OpenCV check with a self-pin; corrected there in issuecomment-5592413761. #4366 already does the layout fix.
- **"the constructor requires that product to be negative, so `tx` is negative for every accepted rig"** (rst page and class block) was false: the guard is `torch.all(tx_fx > 0)`-style, so `tx_fx = 0` is accepted (`tx = -0.0`) and a `[-50, +50]` batch is accepted (`tx = [0.5, -0.5]`), which this PR's own #4270 pins exercise. Both sites now say the guard rejects only a positive product.
- **`depth_to_3d_v2` Args/Return** went back to `(*, 3, 3)` / `(*, H, W, 3)`: executed, `(4, 4)` depth + `(2, 3, 3)` K gives `(2, 4, 4, 3)` (the docstring's own Example), `(2, 3, 4, 4)` + `(2, 1, 3, 3)` gives `(2, 3, 4, 4, 3)`, `(2, 4, 4)` + `(2, 1, 3, 3)` gives `(2, 2, 4, 4, 3)`. The round-4 `(B, ...)` rewrite was the regression; the Return now states the broadcast.
- **`unproject_meshgrid`'s singleton pass-through was written as a convention** while being pinned as a #4271 wart the endorsed repair (rank-3 guard) removes. Now described as an accident under the warning; the Return no longer advertises "preserved singleton leading dimensions".
- **Both bare `:meth:` `reproject_disparity_to_3D` roles resolved to the module-level function.** Sphinx's non-refspecific lookup tries `module.name` before `module.class.name`; the built page's hrefs confirmed it, with no warning under `-W`. Fully qualified now; the rebuilt page links the method.
- **`_check_disparity_tensor`'s docstring still said `(B, 1, H, W)`** three lines under the corrected method docstring. Fixed. Its message ("Expected dimension 1 ... to be 1" for a `shape[-1]` check, never naming `(B, H, W, 1)`, missing space before `Got`) is runtime and is filed as #4374; the pin comment that said "no issue was filed" now cites it.
- **The #4273 "pinned by" list** named `test_convention_warp_frame_depth_matches_depth_warper_without_being_bitwise_equal`, which calls both APIs positionally and carries no `4273`, so the closing grep never finds it and the rename passes through it. It is the agreement pin; the list now names the `_4273` wart pin only, and "all three" is attached to the right clause.
- **`depth_warp` "is the functional form ... equal bit for bit"** omitted that it forwards `align_corners` only: `DepthWarper.__init__(pinhole_dst, height, width, mode, padding_mode, align_corners)` vs `depth_warp(..., align_corners=True)`. Both sites now say so.
- **The #4281 pin comment** attributed the empty-batch `ZeroDivisionError` to `normalize_pixel_coordinates`; the traceback ends in `transform_points` (`linalg.py:224`, `0 // 0`), which is what the docstring corrected in round 4 says. Comment fixed.

**Reported, not applied:** the warning blocks enumerate 17 private test names (8 in `stereo.py`, 10 mentions in `depth.py`), where `main` has only the glob form `test_convention_* / test_wart_*` on `PinholeCamera`. A rename or parametrize silently falsifies the rendered reference and the issue grep does not catch it. #4318 does the same, so this is a batch-wide choice; I left the names in place pending a decision.

**Not findings:** the `depth_to_3d` deprecation-note rewording is accurate (#3142 withdrew the decorator); the three one-line wart fixes the pass suggested stay out because this PR is runtime-free by design.

**Verified at `d344bd0b`:** cpu float32+float64 for both files 235 passed / 4 xfailed on torch 2.14.0 and on torch 2.5.1 arm64; float16 profile 112 / 3 skipped / 7 xfailed; bfloat16 101 / 1 / 20; mps float32 115 / 5 / 2; 6 doctests; pre-commit green; `sphinx -W` from the worktree tree zero warnings. PR body updated (the `tx` bullet, the `_RealTestData` bullet, the history list).

Issue #4275, opened 2026-09-05
W1b: `DepthWarper` / `BrownConradyModel` / `reproject_disparity_to_3D` — public names missing from their rst pages and `__all__`s


ducha-aiki:
Four export / documentation gaps found by parsing the repo's own rst pages at run time and comparing them against the live `__all__`s. AGENTS.md requires every public symbol to appear on its rst page.

1. **`DepthWarper` and `depth_warp`** are in `kornia.geometry.depth.__all__` and on no page — `docs/source/geometry.depth.rst` carries the other seven names.
2. **`BrownConradyModel`, `KannalaBrandtK3` and `Orthographic`** are in `kornia.sensors.camera.__all__` and absent from `docs/source/sensors.camera.rst`, whose note still says "the only supported model is Pinhole". (That note is *more* accurate than the code — see #4284 — but the classes are exported and undocumented either way.)
3. **The module-level `reproject_disparity_to_3D`** is the mirror image: it is on `docs/source/geometry.camera.stereo.rst`, is defined in `kornia.geometry.camera.stereo`, and is in **no** `__all__`, so it is not reachable as `kornia.geometry.reproject_disparity_to_3D`.
4. **`pinhole_matrix` and `inverse_pinhole_matrix`** (and the rest of the legacy 12-vector API, #4268) are on no docs page at all.

Prior art: **#2414** ("Improve camera API docs", closed 2025-12-29) asked for exactly the `sensors.camera` half of (2).

## Repro

```python
import os, re, kornia
import kornia.geometry.depth
import kornia.sensors.camera
import kornia.geometry.camera.stereo as stereo_mod

D = re.compile(r"^\s*\.\.\s+auto(?:function|class|method|data|attribute)::\s+([A-Za-z_][\w.]*)")
C = re.compile(r"^\s*\.\.\s+currentmodule::\s+([\w.]+)")

def rst_names(rel, module=None):
    names, cur = [], None
    for line in open(os.path.join("docs", "source", rel), encoding="utf-8"):
        m = C.match(line)
        if m: cur = m.group(1); continue
        m = D.match(line)
        if m and (module is None or cur == module): names.append(m.group(1))
    return names

for rel, mod, all_ in (("geometry.depth.rst", None, kornia.geometry.depth.__all__),
                       ("sensors.camera.rst", "kornia.sensors.camera", kornia.sensors.camera.__all__)):
    page = rst_names(rel, mod)
    print(f"{rel}: page={page}")
[...]

Issue #4274, opened 2026-09-05
W1b: `kornia.sensors.camera` — a second camera type system that duplicates `kornia.geometry.camera`


ducha-aiki:
kornia ships two camera type systems. `kornia.geometry.camera` takes plain tensors; `kornia.sensors.camera` takes `Vector2` / `Vector3` and raises `AttributeError: 'Tensor' object has no attribute 'z'` on a tensor — the request in **#2708** (closed 2025-12-29). `docs/source/sensors.camera.rst` says the new one "will subsume `kornia.geometry.camera`" (the design discussion is **#2011**, closed 2023-06-19).

Executed on the same parameters, the Pinhole paths of the two systems agree exactly — projection, unprojection, the `K` matrix, the `scale` rule (including the same half-pixel principal-point convention, #4263) and the affine distortion. So today a user has two APIs that compute the same numbers with different types, one of which is documented as the future of the other, and neither page says which to use for new code.

What makes the "will subsume" claim not yet actionable: every non-Pinhole model in `kornia.sensors.camera` constructs and then raises `NotImplementedError` (#4284). There is nothing to subsume with.

## Repro

```python
import torch, inspect, kornia
from kornia.geometry.camera import PinholeCamera, project_points, unproject_points, distort_points_affine
from kornia.geometry.vector import Vector2, Vector3
from kornia.sensors.camera import CameraModel, CameraModelType
from kornia.sensors.camera.distortion_model import AffineTransform
from kornia.image import ImageSize

K3 = torch.tensor([[[100.0, 0.0, 4.0], [0.0, 100.0, 3.0], [0.0, 0.0, 1.0]]])
m = CameraModel(ImageSize(6, 8), CameraModelType.PINHOLE, torch.tensor([100.0, 100.0, 4.0, 3.0]))
try:
    m.project(torch.tensor([1.0, 2.0, 4.0]))
except Exception as e:
    print("PinholeModel.project(Tensor)      :", type(e).__name__ + ":", e)
print("PinholeModel.project(Vector3)     :", m.project(Vector3(torch.tensor([[1.0, 2.0, 4.0]]))).data.tolist(),
      " geometry project_points:", project_points(torch.tensor([[1.0, 2.0, 4.0]]), K3).tolist())
print("PinholeModel.unproject(Vector2, d):", m.unproject(Vector2(torch.tensor([[29.0, 53.0]])), torch.tensor([2.0])).data.tolist(),
[...]

Issue #4267, opened 2026-09-05
W1b: `project_points` — five projection entry points, four different answers at `z = 0`


ducha-aiki:
`kornia.geometry.camera` has **five** entry points that divide by the camera-frame `z`, and at the singular input `(x, y, z) = (1, 2, 0)` they return **four different answers**. Away from the singularity all five agree exactly.

| site | policy | answer at `z = 0` |
|---|---|---|
| `project_points` | masked divide (`abs(z) <= 1e-8` → skip the divide), K applied **after** | `[[104.0, 203.0]]` |
| `PinholeCamera.project` | the same mask, K applied **before** | `[[100.0, 200.0]]` |
| `project_points_z1` | plain divide | `[[inf, inf]]` |
| `Z1Projection.project` (`kornia.sensors`) | plain divide | `[[inf, inf]]` |
| `cam2pixel` | `x / (z + 1e-12)` | `[[[[1.0e14, 2.0e14]]]]` (`inf` in float16) |

Two of these are worth separating out. `PinholeCamera.project` and `project_points` are documented as the same projection and are the container/free-function pair; they apply `K` on opposite sides of the masked divide, so they **disagree exactly on the singular branch** — the divergence is the defect, not either value. And `cam2pixel`'s `+ 1e-12` is not a guard: it enters the arithmetic, dominates the result below `z ≈ 1e-10`, and biases every finite output (relative error `1e-06` already at `z = 1e-6`).

Separately, all five project points **behind** the camera silently: `z = -4` gives a finite pixel with no warning anywhere in the family.

## Repro

```python
import torch
from kornia.geometry.camera import PinholeCamera, cam2pixel, project_points, project_points_z1
from kornia.sensors.camera.projection_model import Z1Projection
from kornia.geometry.vector import Vector3

K4 = torch.eye(4)[None]; K4[0, 0, 0] = K4[0, 1, 1] = 100.0; K4[0, 0, 2], K4[0, 1, 2] = 4.0, 3.0
K3 = K4[:, :3, :3].contiguous()
cam = PinholeCamera(K4.clone(), torch.eye(4)[None], torch.tensor([6]), torch.tensor([8]))
p = torch.tensor([[1.0, 2.0, 0.0]])          # z = 0 exactly: the singular input

print("project_points          :", project_points(p, K3).tolist())
print("PinholeCamera.project   :", cam.project(p).tolist())
print("project_points_z1       :", project_points_z1(p).tolist())
[...]

PR #4318, opened 2026-09-06
docs(sensors): Convention blocks for camera (batch 5d: sensors.camera)

### LINKED ISSUE #4284: W1b: `BrownConradyModel` / `KannalaBrandtK3` / `Orthographic` — the `sensors.camera` models construct and then raise `NotImplementedError`
`BrownConradyModel`, `KannalaBrandtK3` and `Orthographic` are exported from `kornia.sensors.camera.__all__` and construct without complaint. Every operation on them raises `NotImplementedError` with an empty message: `project`, `matrix`, and the underlying `BrownConradyTransform.distort`, `KannalaBrandtK3Transform.distort` and `OrthographicProjection.project`. Only the `AffineTransform` + `Z1Projection` (Pinhole) path is implemented.

`CameraModel`'s own class docstring makes this worse: its `Example` block constructs all three of them, so the rendered documentation demonstrates models that cannot do anything. (`docs/source/sensors.camera.rst`'s note — "At the moment, the only supported model is Pinhole" — is the accurate one, and the three classes are missing from that page entirely; see #4275.)

This is the #4017 pattern again, and it is what makes the rst's "will subsume `kornia.geometry.camera`" claim (#4274) not yet actionable: the working equivalents already exist next door as `kornia.geometry.camera.distort_points_kannala_brandt`, `kornia.geometry.calibration.distort_points` (Brown-Conrady) and `project_points_orthographic`.

## Repro

```python
import torch
from kornia.sensors.camera import CameraModel, CameraModelType
from kornia.sensors.camera.distortion_model import BrownConradyTransform, KannalaBrandtK3Transform
from kornia.sensors.camera.projection_model import OrthographicProjection
from kornia.geometry.vector import Vector2, Vector3
from kornia.image import ImageSize

bc = CameraModel(ImageSize(6, 8), CameraModelType.BROWN_CONRADY, torch.tensor([1.0] * 12))
kb = CameraModel(ImageSize(6, 8), CameraModelType.KANNALA_BRANDT_K3, torch.tensor([1.0] * 8))
orth = CameraModel(ImageSize(6, 8), CameraModelType.ORTHOGRAPHIC, torch.tensor([100.0, 100.0, 4.0, 3.0]))
print("all three construct:", type(bc).__name__, "|", bc.params.tolist()[:4], kb.params.tolist()[:4], orth.params.tolist())
pt = Vector3(torch.tensor([[1.0, 2.0, 4.0]]))
p2 = Vector2(torch.tensor([[0.5, 0.25]]))
for name, call in (("BROWN_CONRADY.project", lambda: bc.project(pt)),
                   ("KANNALA_BRANDT_K3.project", lambda: kb.project(pt)),
                   ("ORTHOGRAPHIC.project", lambda: orth.project(pt)),
                   ("ORTHOGRAPHIC.matrix()", lambda: orth.matrix()),
                   ("BrownConradyTransform().distort", lambda: BrownConradyTransform().distort(torch.tensor([[1.0] * 12]), p2)),
                   ("KannalaBrandtK3Transform().distort", lambda: KannalaBrandtK3Transform().distort(torch.tensor([[1.0] * 8]), p2)),
                   ("OrthographicProjection().project", lambda: OrthographicProjection().project(pt))):
    try:
        call(); print(f"{name:<36}: ok")
    except Exception as e:
        print(f"{name:<36}: {type(e).__name__}: {str(e)!r}")
```

## Observed

```
all three construct: CameraModel | [1.0, 1.0, 1.0, 1.0] [1.0, 1.0, 1.0, 1.0] [100.0, 100.0, 4.0, 3.0]
BROWN_CONRADY.project               : NotImplementedError: ''
KANNALA_BRANDT_K3.project           : NotImplementedError: ''
ORTHOGRAPHIC.project                : NotImplementedError: ''
ORTHOGRAPHIC.matrix()               : NotImplementedError: ''
BrownConradyTransform().distort     : NotImplementedError: ''
KannalaBrandtK3Transform().distort  : NotImplementedError: ''
OrthographicProjection().project    : NotImplementedError: ''
```

## Expected

Two coherent outcomes:

1. **Implement the three models** by delegating to the working `kornia.geometry.camera` functions — `distort_points_kannala_brandt` / `undistort_points_kannala_brandt`, `kornia.geometry.calibration.distort_points` for Brown-Conrady, and `project_points_orthographic` / `unproject_points_orthographic`. The parameter layouts already line up (the `CameraModelType` guard enforces 12 / 8 / 4 parameters respectively), so this is mostly plumbing plus `Vector2` / `Vector3` wrapping, and it is welcome as a PR.
2. **Stop exporting them** until they work: remove the three names from `__all__`, drop them from `CameraModel`'s docstring example, and keep the rst note as the single statement of what is supported.

Doing neither leaves three exported classes whose only observable behaviour is `NotImplementedError` with an empty message. At minimum, the `raise NotImplementedError` sites should carry a message naming the model and pointing at the geometry twin.

Batch 5 documents the current state on each class and links here; it changes no behavior.

## Classification

**API completeness** — advertised-but-unimplemented, the #4017 pattern, with a class docstring demonstrating the unusable models. Idiom: —.

## Disposition

n/a — single symbol. One issue for the `kornia.sensors.camera` models; the two dead functions in `kornia/geometry/camera/pinhole.py` with the same shape are #4283, because the decision there is tied to the legacy 12-vector API instead.

## Environment

### LINKED ISSUE #4313: W1b: `Z1Projection.unproject` — a python-scalar depth builds a CPU tensor and fails on every accelerator
`Z1Projection.unproject` documents `depth: torch.Tensor | float` and its own doctest passes the python
`int` `3`. When `depth` is a python scalar the method promotes it with

```python
if isinstance(depth, (float, int)):
    depth = torch.Tensor([depth])
```

`torch.Tensor([depth])` is a **CPU float32** constructor: it ignores the device and the dtype of `points`.
On cpu the mismatch is invisible; on every accelerator the very next line — `points.x * depth` — raises
`RuntimeError: Expected all tensors to be on the same device`. So the documented and doctested scalar form
of the API works on exactly one device.

The same line also silently widens the dtype: with a `float16` (or `bfloat16`) `Vector2`, the float32 depth
promotes the product, and the returned `Vector3` is `float32`. The tensor form of `depth` has neither
problem, so the two documented spellings of the same argument do not behave the same.

Prior art for the idiom: **#3562** (`get_shear_matrix2d` built its `None` defaults with `torch.tensor([0.0])`
and failed on a GPU `center`), closed 2026-04-20.

## Repro

```python
import torch
from kornia.geometry.vector import Vector2

pts_cpu = Vector2(torch.tensor([[0.25, 0.5]]))
print("cpu, tensor depth  :", Z1Projection().unproject(pts_cpu, torch.tensor([4.0])).data.tolist())
print("cpu, python depth  :", Z1Projection().unproject(pts_cpu, 4).data.tolist())

half = Vector2(torch.tensor([[0.25, 0.5]], dtype=torch.float16))
print("cpu float16 dtype  :", Z1Projection().unproject(half, 4).data.dtype)

if torch.backends.mps.is_available():
    pts = Vector2(torch.tensor([[0.25, 0.5]], device="mps"))
    print("mps, tensor depth  :", Z1Projection().unproject(pts, torch.tensor([4.0], device="mps")).data.tolist())
    try:
        Z1Projection().unproject(pts, 4)
    except Exception as e:
        print("mps, python depth  :", type(e).__name__ + ":", e)
```

## Observed

```
cpu, tensor depth  : [[1.0, 2.0, 4.0]]
cpu, python depth  : [[1.0, 2.0, 4.0]]
cpu float16 dtype  : torch.float32
mps, tensor depth  : [[1.0, 2.0, 4.0]]
mps, python depth  : RuntimeError: Expected all tensors to be on the same device, but found at least two devices, mps:0 and cpu!
```

## Expected

The scalar depth should be materialised on the device and in the dtype of `points`, so that all four cells
above return `[[1.0, 2.0, 4.0]]` and the `float16` cell stays `float16`:

```python
if isinstance(depth, (float, int)):
    depth = torch.tensor([depth], device=points.data.device, dtype=points.data.dtype)
```

That is a one-line, behavior-preserving change on cpu (it only fixes the accelerator and half-precision
cells), so it is a focused fix and welcome as a PR. `CameraModelBase.unproject` is typed
`depth: torch.Tensor` and so is unaffected; only the direct `Z1Projection.unproject` entry point and its
doctest exercise the scalar form.

## Classification

**Broken output** — idiom 2 (a constant/buffer that does not inherit the input's device). The dtype
widening is the same line's second effect.

## Disposition

n/a — single symbol.

## Environment

kornia 0.9.0rc1 @ `e8111a61` (main), torch 2.14.0, python 3.11, cpu and mps (Apple silicon). Found during
the camera conventions audit (batch 5, tracking issue #4262); the batch documents the behavior below as it
is and pins it, and does not change it.

### LINKED ISSUE #4316: W1b: `CameraModelBase` — the base constructor validates neither the length nor the rank of `params`
`CameraModelBase.__init__` documents `params: Camera parameters of shape (B, 4) for PINHOLE Camera, (B, 12)
for Brown Conrady, (B, 8) for Kannala Brandt K3` and then validates none of it:

```python
self._params = params
```

The typed constructors do validate. `CameraModel` and the `PinholeModel` / `BrownConradyModel` /
`KannalaBrandtK3` / `Orthographic` subclasses check `params.shape[-1] != N or len(params.shape) > 2` and
raise `ValueError`, which is what the class-level Convention block describes. `CameraModelBase` is public —
it is exported from `kornia.sensors.camera.__all__`, it is the documented base class, and its own docstring
example constructs one directly — so the documented precondition is enforced on one path and simply absent
on the other.

Two things follow. A **too-short** vector constructs and then fails far from the constructor, inside
`AffineTransform.distort`, with an `IndexError` that names neither `params` nor the camera model. A
**rank-3** `(B, 1, N)` vector — the exact shape the typed constructors reject, because there is no
`(B, N, 4)` multi-camera form — constructs, projects without complaint, and silently returns a `(1, 1, 2)`
result. Neither outcome is the documented contract.

## Repro

```python
import torch

from kornia.geometry.vector import Vector3
from kornia.image import ImageSize
from kornia.sensors.camera import CameraModel, CameraModelType
from kornia.sensors.camera.camera_model import CameraModelBase
from kornia.sensors.camera.distortion_model import AffineTransform

point = Vector3(torch.tensor([[1.0, 2.0, 4.0]]))

# 1. a THREE-element params vector: the base constructor accepts it, project() fails deep inside.
short = CameraModelBase(AffineTransform(), Z1Projection(), ImageSize(6, 8), torch.ones(3))
print("short  constructed, params.shape =", tuple(short.params.shape))
try:
    short.project(point)
except Exception as e:
    print("short  project      :", type(e).__name__ + ":", e)

# 2. a (1, 1, 4) params tensor: rejected by the typed constructors, accepted and USED here.
rank3 = CameraModelBase(AffineTransform(), Z1Projection(), ImageSize(6, 8), torch.ones(1, 1, 4))
print("rank3  constructed, params.shape =", tuple(rank3.params.shape))
print("rank3  project      :", tuple(rank3.project(point).data.shape), rank3.project(point).data.flatten().tolist())

# 3. the typed constructors reject both of those parameter tensors.
for params in (torch.ones(3), torch.ones(1, 1, 4)):
    try:
        CameraModel(ImageSize(6, 8), CameraModelType.PINHOLE, params)
    except Exception as e:
        print(f"typed  {tuple(params.shape)!s:<12}:", type(e).__name__ + ":", e)
```

## Observed

```
short  constructed, params.shape = (3,)
short  project      : IndexError: index 3 is out of bounds for dimension 1 with size 3
rank3  constructed, params.shape = (1, 1, 4)
rank3  project      : (1, 1, 2) [1.25, 1.5]
typed  (3,)        : ValueError: params must be of shape (B, 4) for PINHOLE Camera
typed  (1, 1, 4)   : ValueError: params must be of shape (B, 4) for PINHOLE Camera
```

## Expected

Either of two coherent outcomes; the current state is neither.

1. **Validate in the base**, so that the documented precondition is the enforced one on every path.
   `CameraModelBase.__init__` grows the rank check the typed constructors already apply — `params` is
   `(N,)` or `(B, N)`, never higher-rank — and, where the distortion/projection pair fixes a length, the
   length check too. Both cells above then raise `ValueError` from the constructor, the way
   `CameraModel(ImageSize(6, 8), CameraModelType.PINHOLE, ...)` already does. This is the behavior a reader
   of the `__init__` docstring expects, and it turns a deep `IndexError` and a silent wrong-rank result into
   one message at the call site.

2. **Say it is unchecked.** If `CameraModelBase` is deliberately the escape hatch for a distortion /
   projection pair that has no `CameraModelType` — the only reason to construct one directly — then the
   `__init__` docstring should stop stating a shape it does not enforce, and say instead that `params` is
   stored as given and that shape validation lives in the typed constructors.

Option 1 is preferred: the shape statement is already in the docstring, and the check is the same two
comparisons the four subclasses each spell out. Option 2 is the smaller change and would at least make the
documentation true.

## Classification

**Semantic contract** — idiom 5 (a documented precondition that is never validated). Same family as
**#4266**, where `PinholeCamera`'s seven shape guards do not enforce what their messages say; here the base
class has the message in its docstring and no guard at all.

## Disposition

n/a — single symbol.

## Environment

kornia 0.9.0rc1 @ `e8111a61` (main), torch 2.14.0, python 3.11, cpu. Found during the camera conventions
audit (batch 5, tracking issue #4262); the batch documents the behavior above as it is and pins it, and does
not change it.

ducha-aiki:
## Summary

Documents and pins `kornia.sensors.camera`, the `Vector`-typed camera type system: what `params` holds for each
`CameraModelType` and who validates it, what `matrix()` returns and how it differs from what `PinholeCamera`
stores, what `depth` means, and the mathematical Pinhole mapping it shares with
`kornia.geometry.camera`, including its camera-axis broadcasting, projection rounding and zero/near-zero depth differences, while every other model raises. Six `Convention:` blocks, seven pointer lines back to
the anchor, nine `.. warning::` blocks, four issues linked, 14 additive test pins, and three `autoclass`
entries plus a rewritten note and description on `docs/source/sensors.camera.rst`.

The anchor for this type system is `CameraModelBase`; every other symbol points at it rather than restating it.
Two sentences additionally point at the `PinholeCamera` Convention block for the pixel grid, which #4294
(sub-batch 5a) put on `main` on 2026-09-08.

Documentation and tests only, with zero runtime change: an AST comparison after stripping docstrings is empty
for all three modified `kornia/` files. No existing test assertion and no doctest was modified; every pin is
new (the three deleted lines under `tests/` are import statements extended in place). Part of #4262, with a
changelog entry under `## Unreleased` → `### Added`.

**Rebased 2026-09-10 onto `main` `f36ab6de`** (merge commit `5e214b21` + fix commit `3ced4c71`). Two of the
issues this PR originally pinned as warts were repaired on `main` in the meantime — #4340 closed #4313
(python-scalar `depth` now built on the device and dtype of `points`) and #4369 closed #4316
(`CameraModelBase.__init__` now validates the rank and length of `params`) — so on the merged tree the two
wart pins failed in every non-vacuous cell (`_4316` in all 4 cpu dtypes and both MPS dtypes; `_4313` in cpu
`float16`/`bfloat16` and both MPS dtypes). Those two pins, the three `.. warning::` blocks that documented the
repaired behaviour and the two CHANGELOG clauses are gone; the `params` Convention bullet now states the
base-class check and the `Z1Projection.unproject` Convention block states the promotion. The AGENTS grep for
`#4313|issues/4313|#4316|issues/4316` over `kornia/` and `tests/` is empty. `main`'s own pins for the two
fixes (`TestCameraModelBaseParamsValidation`, `test_unproject_scalar_depth`) are kept next to this PR's.

## Scope

| symbol | what it got |
| --- | --- |
| `CameraModelBase` | the **anchor** `Convention:` block: `Vector`-typed in and out (a raw `Tensor` raises `AttributeError`); the per-`CameraModelType` `params` length and who enforces it (the typed constructors and, since #4369, the base constructor itself); `matrix()`/`K()` as the `(*, 3, 3)` intrinsics; `project`/`unproject` as the distortion∘projection composition with `depth` = camera-frame `z`; the shared Pinhole mapping, approximate projection agreement away from `abs(z) <= 1e-8`, the divergent zero/near-zero depth handling, the different camera-axis broadcasting of batched `(B, 4)` intrinsics over `(B, N)` point clouds, and unprojection with `normalize=False` + warning #4284 |
| `CameraModelBase.matrix` / `.K` / `.project` / `.unproject`, `PinholeModel`, `PinholeModel.matrix`, `CameraModel` | the seven pointer lines back to the anchor |
| `PinholeModel.scale` | `Convention:` (a **new** model with `fx' = s·fx` and `cx' = s·cx`, the half-pixel rule `PinholeCamera.scale` also applies) + warning #4263 (both deviations) |
| `BrownConradyModel`, `KannalaBrandtK3`, `Orthographic` | one warning each: construction succeeds, then `project`/`unproject`/`matrix` raise, naming the site each raises from (#4284) |
| `AffineTransform.distort` | `Convention:` (normalized `z = 1` in, pixels out, `params = [fx, fy, cx, cy]`; the same values as `distort_points_affine`, from which the two type systems' separation is #4274; `undistort` is closed-form) |
| `AffineTransform.undistort` | `Convention:` (pixels in, normalized out, `(u − cx)/fx`; cross-reference to the `distort` block for the layout) |
| `BrownConradyTransform`, `KannalaBrandtK3Transform` | one warning each (#4284) naming the implemented counterpart — `calibration.distort_points` is the Brown-Conrady *model* in pixel space, `distort_points_kannala_brandt` is an interface equivalent |
| `Z1Projection.project` | `Convention:` (`xy / z`, no epsilon and no validation; `z = 0` gives an infinity, `nan` on an axis whose numerator is also zero; a point behind the camera gives a finite coordinate) + warning #4267 |
| `Z1Projection.unproject` | `Convention:` (`depth` is the camera-frame `z`, so the third coordinate of the result *is* the depth passed in — not a ray length; a python-scalar `depth` is promoted on the device and in the dtype of `points`, the behaviour #4340 fixed) |
| `OrthographicProjection` | warning #4284, naming `project_points_orthographic` as the implemented equivalent |

## Pins

14 new pins — 6 `test_convention_*` and 8 `test_wart_*`, **0 strict `xfail`s**.

`tests/sensors/camera/test_camera_model.py` — 5 convention, 5 wart:
`test_convention_projection_matches_geometry_camera_project_points`,
`test_convention_unproject_takes_the_camera_frame_z_as_depth`,
`test_convention_matrix_is_the_three_by_three_intrinsics`,
`test_convention_params_length_is_fixed_per_camera_model_type`,
`test_convention_shared_or_paired_intrinsics_match_geometry` (shared `(4,)` intrinsics over a point cloud and one point per camera with `(B, 4)`, both `torch.equal` to the geometry functions),
`test_wart_projection_differs_from_geometry_at_small_depth_4267` (#4267; parametrized over zero, small positive/negative depth, and both threshold boundaries),
`test_wart_batched_intrinsics_differ_from_geometry_4274` (#4274; `B == N == 2` runs in both APIs with different camera associations, `B = 2, N = 3` raises here and runs there),
`test_wart_scale_rescales_the_principal_point_by_the_half_pixel_rule_4263` (#4263),
`test_wart_scale_turns_the_image_size_fields_into_tensors_4263` (#4263),
`test_wart_the_three_non_pinhole_models_construct_and_then_raise_4284` (#4284).

`tests/sensors/camera/test_distortion_model.py` — 1 convention, 1 wart:
`test_convention_distort_is_the_affine_map_and_undistort_is_its_exact_inverse`,
`test_wart_brown_conrady_and_kannala_brandt_transforms_are_placeholders_4284` (#4284).

`tests/sensors/camera/test_projection_model.py` — 2 wart:
`test_wart_project_divides_by_z_with_no_guard_4267` (#4267), `test_wart_orthographic_projection_is_a_placeholder_4284`
(#4284).

Every wart pin asserts the **current** behavior, names its issue in the test name, and says in a comment that
it is deleted when the issue is repaired. The original 11 pins were mutation-checked: the claim was flipped, the node
observed to fail, and the flip reverted.

**There is no strict `xfail` here**, deliberately. One is honest only where an issue has settled a single
intended behavior *and* the current behavior differs from it in every collected cell. #4263 (the resize rule),
#4267 (three coherent `z = 0` policies) and #4284 (implement the models, or stop exporting them) each leave
the choice open.

## Issues documented here

| --- | --- | --- |
| #4274 `kornia.sensors.camera` is a second camera type system that duplicates `kornia.geometry.camera` | duplication (W1b) | **KEEP SEPARATE — intentional variant**; the `Vector` typing differs; the Pinhole mapping is shared, but projection rounding, zero/near-zero depth policies and the camera-axis broadcasting of batched intrinsics differ, now covered explicitly and recorded on the issue (issuecomment-5616135040) |
| #4284 `BrownConradyModel` / `KannalaBrandtK3` / `Orthographic` construct and then raise `NotImplementedError` | missing implementation behind a public export | n/a — one issue, three classes and three raise sites |
| #4263 `PinholeCamera.scale` — the principal-point rescale rule contradicts the library's pixel grid | broken output | `PinholeModel.scale` is that issue's fourth site; the `ImageSize`-field observation is recorded in its comment thread (issuecomment-5556356093), not its body, and the warning says so |
| #4267 five projection entry points, four different answers at `z = 0` | semantic contract | n/a — `Z1Projection.project` is the fifth site |

#4313 and #4316 were found while writing this sub-batch and filed with repros re-executed against this PR's
original base; both were fixed on `main` before this PR merged (#4340, #4369) and their pins and warnings
were dropped in the 2026-09-10 rebase. This PR also closes **gap (2) of #4275** — the `sensors.camera` half;
its other three gaps are untouched.

## Documentation errors fixed

- `docs/source/sensors.camera.rst`'s `.. note::` said "At the moment, the only supported model is Pinhole. We
  plan to add Kannala-Brandt, orthographic, and other models in the future." Those three models are exported
  and importable: they construct, and then every call raises. The note now says that, and links #4284.
- `CameraModelBase`'s class docstring listed "Distortion is of 3 types" and "Projection is of 2 types" as if
  all five worked. Each entry now names its implementing class, or says it is a placeholder that raises.
- `CameraModelBase.__init__`'s Args documented `params` as `(B, 4)` / `(B, 12)` / `(B, 8)` and validated none
  of it (#4316). #4369 rewrote that Args line and added the check on `main`; the anchor's Convention bullet
  now states the per-distortion length the base reads and that the unbatched `(N,)` form is accepted.
- `PinholeModel.scale` documented neither of its deviations. The field type of the returned `image_size`
  follows the **argument** type — a tensor factor gives 0-dim floating tensors, a python `int` keeps `int`
  fields, a python `float` gives floats — and the method's own doctest shows only the `int` case. The warning
  is scoped to the tensor factor and states the other two.
- `Z1Projection.project` documented no `z = 0` behavior at all. `Vector3([[0., 2., 0.]])` gives `[nan, inf]`:
  an axis whose numerator is also zero computes `0 / 0`, so the block states `nan` as well as `inf`.
- `CameraModelBase.project` / `unproject` now write the composition out with its real first argument
  (`self.distortion.distort(self.params, self.projection.project(points))` and the reverse).

## `docs/source/sensors.camera.rst`

`.. autoclass:: BrownConradyModel`, `KannalaBrandtK3` and `Orthographic` (`:members:`) join `PinholeModel`; the
stale note is replaced as above; the "will subsume `kornia.geometry.camera`" warning **stays** and gains one
clause saying the two are kept separate by design today, share the mathematical Pinhole mapping with documented projection rounding and depth-threshold exceptions, and that this
is recorded in #4274. The `.. meta::` description's "only the Pinhole model" sentence is aligned with the note (`6a81e6e0`); both code samples and the section
structure are byte-identical.

## Docs-only proof

For each of the three modified `kornia/` files (`camera_model.py`, `distortion_model.py`,
`projection_model.py`), parse both revisions, pop the leading string `Expr` from every `Module` / `ClassDef` /
`FunctionDef` / `AsyncFunctionDef`, `ast.dump` and diff — **EMPTY for all three**. The `kornia/` diff contains
no `>>>` or `...` line.

## Validation (at `ca7f286a`, before the rebase)

- `pixi run lint` and `pixi run pre-commit-all` (after `git add -A`) — passed; `pixi run doctest` — 650
  passed, 11 skipped
- clean `pixi run build-docs` (`rm -rf docs/build/html` first) — `build succeeded`, and the
  `sensors|camera_model|distortion_model|projection_model` filter over the build log's warning/error lines is
  empty; the whole log's only warning is an unrelated `libpng warning: iCCP`. Every xref this change adds was
  checked against `objects.inv` and as a real `<a href>` in the built page
- `pytest tests/sensors/camera --dtype=float32,float64` — 62 passed, 29 skipped; the same at
  `--device=mps --dtype=float32,float16` with `PYTORCH_ENABLE_MPS_FALLBACK` unset — 62 passed, 29 skipped
- `--dtype=all` — 123 passed, 55 skipped, **1 failed**:
  `test_camera_model.py::TestPinholeCamera::test_project_unproject[cpu-float16-2]`. That test draws
  `torch.rand` with no `manual_seed` and compares a float16 round trip at the default tolerance; its body is
  byte-identical at the base commit and this PR changes nothing executable. It is the only node that has ever
  failed across this branch's runs, and it fails a different parametrization each time

## Duplication ledger

One row touched: `geometry.camera` / `sensors.camera` — **KEEP SEPARATE, intentional variant**. Executed on the
Pinhole fixtures: `unproject` equals `unproject_points`, `matrix()` equals the 3×3 `K`, `scale` equals
`PinholeCamera.scale`, and `AffineTransform().distort` equals `distort_points_affine`, all `torch.equal`
on the selected representable literals. Projection is byte-identical on the `z=4` fixture only; general
nonzero depths can differ through rounding (direct division versus reciprocal multiplication). At
`abs(z) <= 1e-8` geometry skips division while sensors divides unconditionally; this is a substantive
policy difference, not merely rounding. The new pin exercises zero, small positive/negative depth, and
the threshold boundaries; a non-power-of-two depth checks approximate agreement away from the threshold. Batched `(B, 4)` intrinsics over `(B, N)` point clouds broadcast along different axes in the two APIs (the point axis here, the camera axis there), pinned by `test_wart_batched_intrinsics_differ_from_geometry_4274`. The rst's "will subsume" line is a tiering decision for the
coordinated window, not a merge verdict, and is recorded in #4274.

## Notes for reviewers

- **`BrownConradyTransform`, `KannalaBrandtK3Transform` and `OrthographicProjection` are on no rst page**, so
  their new `.. warning::` blocks are source-only. Adding them is #4275's gap-(2) sibling question rather than
  a decision for this PR; their xref targets all resolve, so nothing breaks if they are added later.
- **The page's `.. meta::` description** carried the same "only the Pinhole model" staleness the `.. note::` had;
  fixed in `6a81e6e0`.
- **`AffineTransform.undistort` is the closed-form inverse, not a bit-exact one.** Over 2000 seeded random
  `(point, params)` draws in cpu float32 the round trip is bit-exact in 544 and differs in 1456, by up to
  3.9e-06. The docstring therefore says "closed-form", and the pin — whose name still contains "exact
  inverse" — is scoped to the representable literals it uses, where `atol = rtol = 0` is honest; its comment
  carries the full seeded recipe.
- **`test_project_unproject` is a pre-existing unseeded half-precision flake** (see Validation). Seeding it is
  a follow-up; this PR does not touch it, per the batch rule that a documentation PR is additive.
- **Ordering against #4294 (sub-batch 5a)** is settled: #4294 merged on 2026-09-08, so the two
  `PinholeCamera` block pointers now point at a block that exists on `main`.
- **Collision ahead: #4377** (implements the three placeholder models, fixes #4284) is open and
  `CONFLICTING`. Whichever of the two lands second drops this PR's four `_4284` pins, the seven `#4284`
  prose sites in `kornia/sensors/camera/` and the rst note, and re-checks the `matrix()` raise site.

## Validation after the rebase (2026-09-10, head `3ced4c71`, torch 2.14.0)

- `pytest tests/sensors/camera --dtype=float32,float64,float16,bfloat16` — 163 passed, 55 skipped, 1 failed:
  the pre-existing unseeded `test_project_unproject[cpu-float16-*]` flake, which fails 1 run in 6 on `main`
  `f36ab6de` as well
- `--device=mps --dtype=float32,float16` with `PYTORCH_ENABLE_MPS_FALLBACK` unset — 82 passed, 29 skipped
- `pytest --doctest-modules kornia/sensors/camera` — 11 passed
- pinned `ruff` (0.16.6) check and format clean on the touched files
- AST docs-only proof re-run against `main` `f36ab6de` — empty for all three `kornia/` files; the `kornia/`
  diff contains no `>>>` or `...` line
- `sphinx -W -b html` on the merged tree — see the review comment for the result

## Equivalence review correction

The anchor, sensors page, changelog, and test comments now distinguish the shared mathematical mapping
from floating-point identity and the different behavior at `abs(z) <= 1e-8`. The added test covers
`z = 0, ±1e-9, ±1e-8`; the existing convention test also exercises `[5, 2, 3]` with approximate comparison.
Runtime code remains unchanged.

Validation of the correction (torch 2.14.0, verified worktree imports):

- Full CPU float32/float64 sensors/camera suite: **92 passed, 29 skipped**.
- Convention/wart/scalar-depth selection across all four CPU dtypes: **68 passed**.
- The same selection on MPS float32/float16: **34 passed**.
- Camera doctests: **11 passed**; full `pixi run pre-commit-all`: **passed**.
- Sphinx HTML build with `-W`: **passed**; runtime AST after removing docstrings: **unchanged**.
- The full MPS float32/float16 run gave **91 passed, 29 skipped, 1 failed** in the unchanged
  `test_project_unproject[mps-float16-5]`. Its half-precision round-trip failure reproduces with seed 4
  on both the original PR head `3ced4c71` and this correction, with the same error. It is not caused
  by the documentation/test additions. CUDA and the full repository test suite were not run.

## Batching review correction and review round at `6a81e6e0` (2026-09-10)

`f5394499` answered the second Codex finding: the anchor, the page, the changelog and two new pins now state that batched
`(B, 4)` intrinsics broadcast along the point axis here and the camera axis in `kornia.geometry.camera`, and that the
correspondence holds for shared intrinsics and for one point per camera. `6a81e6e0` (review round, see the review
comment) fixed three nits: the `scale` warning's "0-dim floating tensors" is scoped to a floating-point factor
(`torch.tensor(2)` gives int64 fields), the params-length convention pin no longer pins the inconsistent shape
spelling of the `ValueError` message, and the rst `.. meta::` description no longer says only Pinhole is supported.

Validation at `6a81e6e0` (torch 2.14.0, `kornia.__file__` verified in the branch worktree):

- `pytest tests/sensors/camera --dtype=float32,float64,float16,bfloat16` — 200 passed, 55 skipped; per-dtype half
  runs with the CI profiles — float16 50 passed, bfloat16 49 passed + 1 xfailed
- `--device=mps --dtype=float32,float16`, `PYTORCH_ENABLE_MPS_FALLBACK` unset — the convention/wart/validation
  selection 64 passed; the full run at `f5394499` was 99 passed, 29 skipped, 1 failed in the unseeded
  `test_project_unproject[mps-float16-5]`, which fails 1 run in 6 on `main` `f36ab6de` as well
- doctest 11 passed; pinned ruff 0.16.6 check and format clean; AST docs-only proof identical to `main` for all three
  `kornia/` files; no line over 120 characters
- merge with `main` `749a1c17` clean; on the merged tree cpu float32/float64 100 passed, the AGENTS grep for
  `#4313|#4316` empty, `sphinx -W` zero warnings, and every xref this PR adds is a live link on the rendered
  `sensors.camera.html` (the only dead ones are pre-existing: `Vector2`/`Vector3` in signatures and the `:mod:` refs)

Posted on behalf of @ducha-aiki by Codex (gpt-6-astra) and Claude (Fable 5.1).


ducha-aiki:
## Review of #4318 (batch 5d: sensors.camera) at `ca7f286a`

**Verdict: approve.** All 44 checks are green, the diff is docstrings, one rst page and pins, and every claim I re-executed on the branch worktree (torch 2.14.0) holds: `PinholeModel.project` / `unproject` are `torch.equal` to `project_points` / `unproject_points` on the same `K`, `matrix()` is the `(3, 3)` layout carrying the batch axis, the three non-pinhole models construct and then raise an empty `NotImplementedError` from the three sites the pin names, `CameraModelBase` validates no `params` shape, and the python-scalar `depth` builds a CPU float32 tensor. The `Z1Projection.project` block is the one place in batch 5 that states the `z = 0` answer completely ("to `nan` on an axis whose numerator is zero as well"); #4294's `project_points_z1` block should be brought in line with it, which I asked for there.

### P2 (none blocking)

1. Cross-cutting for 5a to 5d: pin comments cite "audit labels 5d-sc-01" and "this worktree", which resolve to nothing in this repository. Keep the executed snippet and a commit hash, drop or spell out the labels.
2. The `CameraModelBase` block points at the Convention block on `PinholeCamera`, which exists only in #4294; merge 5a first, as planned.
3. The `Z1Projection.unproject` warning says "verified on MPS"; CUDA raises the same device-mismatch error for the same reason, so "on any non-CPU device (executed on MPS)" reads less like an MPS-only caveat.

Checked and clean: `docs/source/sensors.camera.rst` now renders `BrownConradyModel`, `KannalaBrandtK3` and `Orthographic` (the docs job passed), the `:doc:` target `/geometry.camera` exists, the `scale` int/float/tensor arms of `test_wart_scale_turns_the_image_size_fields_into_tensors_4263` match the source, the `CameraModelType` enum values, and CHANGELOG placement. GitHub reports the PR as `CONFLICTING` against today's `main` while a local `merge-tree` is clean; rebase so the green comes from a real run.

ducha-aiki (commented):
Integration re-review of `ca7f286a` against current main (`f36ab6de`): the branch needs to absorb both merged repairs before it is ready.

- **#4316 / parameter validation:** transplanting this PR's sensor test files onto current main makes `test_wart_camera_model_base_validates_no_params_shape_4316` fail for CPU float32 and float64: the base constructor now correctly rejects `(3,)` affine parameters. Replace the obsolete wart with the repaired contract and update the warning/parameter-validation prose and PR description.
- **#4313 / scalar depth:** the transplanted `test_wart_unproject_with_a_python_scalar_depth_builds_a_cpu_tensor_4313` fails on MPS float32 because the expected device-mismatch exception no longer occurs. Retire that wart and its two public warnings, retaining positive scalar-depth device/dtype coverage.

Measured in an isolated current-main worktree with verified import origin: targeted CPU selection **2 failed, 2 passed**; targeted MPS selection **1 failed**. These are integration results, not results on the unchanged PR branch. GitHub also reports a conflict. No fresh full-suite or docs build was run.

The non-pinhole placeholder documentation additionally needs coordination with #4377 if that implementation lands first. This is a comment review because this account owns the PR; eligible external approval is still required after the refresh.

ducha-aiki:
## Re-review of #4318 at `3ced4c71` (rebased onto `main` `f36ab6de`, 2026-09-10)

**Verdict: approve.** The 2026-09-06 approval at `ca7f286a` (issuecomment-5558770294) stood on a head that `main` had since falsified in two places, so this is a delta review of what changed between that head and this one, and the fix was applied on the branch.

### What `main` did to the approved head

Two of the six issues this PR pinned as warts were repaired on `main` after the review:

| issue | fix on `main` | this PR's pin on the merged tree |
| --- | --- | --- |
| #4313 python-scalar `depth` built a CPU float32 tensor | #4340 (`1f735028`): `torch.as_tensor([depth], device=points.data.device, dtype=points.data.dtype)` | `test_wart_unproject_with_a_python_scalar_depth_builds_a_cpu_tensor_4313` **fails** in cpu `float16`/`bfloat16` (dtype no longer widened) and in both MPS dtypes (no `RuntimeError`); passes vacuously in cpu `float32`/`float64` |
| #4316 `CameraModelBase.__init__` validated no `params` shape | #4369 (`0272024f`): `_validate_params` raises `ValueError` on rank ∉ {1, 2} or the wrong length for the distortion | `test_wart_camera_model_base_validates_no_params_shape_4316` **fails** in all 4 cpu dtypes and both MPS dtypes (`ValueError: params must be of shape (B, 4) or (4,) for AffineTransform; got shape (3,)`) |

Both executed on the merged tree before any edit: `pytest tests/sensors/camera --dtype=float32,float64,float16,bfloat16` gave 7 failed / 165 passed (6 of the 7 are these two pins, the seventh is the pre-existing unseeded `test_project_unproject[cpu-float16-*]` flake, which also fails 1 run in 6 on `main` itself), and `--device=mps --dtype=float32,float16` gave 4 failed / 82 passed, all four from these two pins. The AGENTS grep for `#4313|issues/4313|#4316|issues/4316` on the merged tree found 3 prose sites in `kornia/sensors/camera/` (two warnings on `CameraModelBase`, one on `Z1Projection.unproject`) plus the two pins — the "stale docstring after a pin flip" defect, one PR removed.

The test files also conflicted textually: `main`'s `TestCameraModelBaseParamsValidation` and `test_unproject_scalar_depth` land in the same two files, at the same insertion points, with overlapping imports.

### The delta on the branch (`5e214b21` merge + `3ced4c71` fix)

- Merge resolved keeping both sides: `main`'s import superset, `main`'s two new test classes/methods appended after this PR's pins.
- Dropped the two repaired wart pins; dropped the three `.. warning::` blocks (`CameraModelBase` ×2, `Z1Projection.unproject` ×1); dropped the two clauses from the CHANGELOG bullet.
- The `params` Convention bullet no longer says the base "validates neither the length nor the rank"; it now states that `CameraModelBase.__init__` applies the same check on the direct path and reads the length off the distortion type (4 / 12 / 8), which is what `_validate_params` does. The `Z1Projection.unproject` Convention block gained a bullet stating the promotion on the device and dtype of `points`, which `main`'s `test_unproject_scalar_depth` pins.
- Nothing else in the PR moved: 6 Convention blocks, 7 pointer lines, 9 warnings, 4 issues (#4263, #4267, #4274, #4284), 11 pins (5 convention, 6 wart), 0 strict xfails.

### Verified on `3ced4c71` (torch 2.14.0)

- `pytest tests/sensors/camera --dtype=float32,float64,float16,bfloat16` — 163 passed, 55 skipped, 1 failed (the `test_project_unproject` float16 flake, pre-existing and reproduced on `main`)
- `--device=mps --dtype=float32,float16`, `PYTORCH_ENABLE_MPS_FALLBACK` unset — 82 passed, 29 skipped
- `pytest --doctest-modules kornia/sensors/camera` — 11 passed
- AST-after-docstring-strip of the three `kornia/` files identical to `main` `f36ab6de`; no `>>>`/`...` line in the `kornia/` diff
- pinned `ruff` 0.16.6 check + format clean; the AGENTS grep for #4313/#4316 over `kornia/` and `tests/` empty; the 4263/4267/4274/4284 sites unchanged
- `sphinx -W -b html` on the merged tree from the worktree (`PYTHONPATH` set so `kornia.__file__` names it) — zero warnings; the rendered `sensors.camera.html` carries no `#4313`/`#4316` reference, both new Convention bullets, and live links for the `PinholeCamera` pointers (the block #4294 added) and `AffineTransform`
- `git merge-tree` against `main` clean; the PR body has been refreshed to match (counts, the issues table, a "Rebased 2026-09-10" section).

### Still open from the 2026-09-06 review (P2, non-blocking)

1. Pin comments cite "audit labels 5d-sc-NN" and "this worktree", which resolve to nothing public. Batch-wide; not touched here.
2. The 17-private-test-names-in-public-warnings question raised on #4317 round 6 applies to the seven `_4284` mentions here too; awaiting the user's call, not touched.

### Collision to schedule

#4377 (implements Brown-Conrady, Kannala-Brandt and orthographic, fixes #4284; `CONFLICTING` against `main`) touches all three source files and all three test files this PR touches. Whichever lands second drops the four `_4284` pins, the seven `#4284` prose sites and the rst note, and re-checks the `matrix()` raise site — #4377 says it "adds intrinsic matrix support for all camera models", so that third site goes too. Merging #4318 first keeps the rebase on the side that has to change anyway.

ducha-aiki (commented):
Re-reviewed head 3ced4c71 against the previously reviewed ca7f286a and current main f36ab6de. Verdict: code-review clear.

The new merge and follow-up resolve both integration findings from the prior review: the obsolete #4313 scalar-depth wart and both public warnings are removed in favor of the positive device/dtype convention already tested on main, and the obsolete #4316 no-validation wart/warning is removed while the Convention block now describes the base-class rank/length validation. A grep over kornia and tests finds no remaining #4313/#4316 pin or warning, and merge-tree against current main is clean. The remaining #4316 CHANGELOG entry belongs to merged #4369 and correctly records that repair.

From an isolated checkout with kornia.__file__ verified, the full sensors/camera CPU float32/float64 selection passed (82 passed, 29 skipped), and the focused CPU float16/bfloat16 scalar-depth/parameter-validation selection passed (26 passed). I found no actionable issue in the delta. Hosted CI is still running and an eligible external approval is still required because this account owns the PR. If #4377 lands first, its #4284 implementation will require the already identified prose/pin cleanup before this PR merges.

kornia/sensors/camera/camera_model.py
```diff
@@ -104,12 +104,57 @@ class CameraModelBase:
...
+        - :meth:`project` is ``self.distortion.distort(self.params, self.projection.project(points))`` and
+          :meth:`unproject` the reverse,
+          ``self.projection.unproject(self.distortion.undistort(self.params, points), depth)``. ``depth`` is
+          the camera-frame ``z``: :class:`~kornia.sensors.camera.projection_model.Z1Projection` multiplies
+          the :math:`z = 1` point by it, so the third coordinate of the result is the ``depth`` that was
+          passed in, and not a Euclidean ray length.
+        - on the Pinhole path the numbers are those of :doc:`kornia.geometry.camera </geometry.camera>`:
+          :meth:`project` matches :func:`~kornia.geometry.camera.perspective.project_points` and
+          :meth:`unproject` matches :func:`~kornia.geometry.camera.perspective.unproject_points` on the ``K``
+          built from the same ``[fx, fy, cx, cy]``, so the pixels are on the integer-centre grid described in
```
ducha-aiki: [P2] Qualify the claimed equivalence with geometry.project_points

This newly documented equivalence is false for valid inputs. With `params=[100.,50.,4.,3.]`, `PinholeModel.project(Vector3([[1.,2.,0.]]))` returns `[inf,inf]`, whereas `project_points` with the same K returns `[104.,103.]`; at `z=1e-9` the sensors result is about `[1e11,1e11]` but geometry still returns `[104.,103.]`. Geometry masks `abs(z)<=1e-8`, while Z1Projection divides unconditionally. Even away from that threshold the byte-identical claim in the new test comment/PR description fails: CPU float32 `[5.,2.,3.]` gives x=170.66665649414062 versus 170.6666717529297 because the implementations use direct division versus reciprocal multiplication. The new equality pin only exercises `z=4`, which hides both differences. Please describe the common mathematical pinhole mapping with the threshold/rounding exceptions, qualify the corresponding rst and PR claims, and add non-power-of-two and zero/near-zero coverage; no runtime change is needed.

ducha-aiki (commented):
Reviewed all eight changed files at `3ced4c71f71cce6d44974ffdbac750a931af90ef` personally, without subagents, in an isolated detached worktree.

**Verdict: needs documentation correction before merge — one P2 finding.** The new anchor and page overstate numerical equivalence between the two pinhole APIs. The attached inline finding includes reproduced counterexamples at zero depth, small nonzero depth, and ordinary positive depth. Passing the deliberately exactly representable fixture does not establish the advertised general equivalence.

Minimal reproduction on this head (CPU float32, torch 2.14.0):

```python
import torch
from kornia.geometry.camera import project_points
from kornia.geometry.vector import Vector3
from kornia.image import ImageSize
from kornia.sensors.camera import PinholeModel

cam = PinholeModel(ImageSize(6, 8), torch.tensor([100., 50., 4., 3.]))
for xyz in ([1., 2., 0.], [1., 2., 1e-9], [5., 2., 3.]):
    p = torch.tensor([xyz], dtype=torch.float32)
    print(cam.project(Vector3(p)).data.tolist(),
          project_points(p, cam.matrix()).tolist())
```

The first pair is `[[inf, inf]]` versus `[[104., 103.]]`; the second is approximately `[[1e11, 1e11]]` versus `[[104., 103.]]`; the third differs in x by about `1.53e-5`. The runtime difference already exists; the defect introduced here is documenting a stronger equivalence than the implementation provides.

Validation, with the interpreter explicitly selected and `kornia.__file__` verified inside the review worktree:

- Full sensors/camera CPU float32/float64: **82 passed, 29 skipped**.
- Added convention/wart tests plus scalar-depth regression, CPU float16/bfloat16: **24 passed**.
- Full sensors/camera MPS float32/float16: **82 passed, 29 skipped**.
- Camera module doctests: **11 passed**.
- All three changed runtime files have identical ASTs to merge-base `f36ab6de` after stripping docstrings; `git diff --check` passes.
- Local merge-tree against fetched main `c1500866` is clean. No remaining #4313/#4316 runtime warning was found; the earlier integration findings are resolved.

- Fresh Sphinx HTML build with `-W`: **passed, no warning/error lines**. Verified the three added class entries and the rendered PinholeCamera/project_points cross-references.

No code was changed. CUDA and the full repository test suite were not run. This is posted as a comment review because the signed-in account owns the PR.

kornia/sensors/camera/camera_model.py
```diff
@@ -104,12 +104,63 @@ class CameraModelBase:
...
+          passed in, and not a Euclidean ray length.
+        - the Pinhole path uses the same mathematical camera mapping as
+          :doc:`kornia.geometry.camera </geometry.camera>` with ``K`` built from the same ``[fx, fy, cx, cy]``.
+          However, :meth:`project` divides directly by ``z``, whereas
+          :func:`~kornia.geometry.camera.perspective.project_points` multiplies by its reciprocal and skips
+          the divide when ``abs(z) <= 1e-8`` (compared in the working dtype). Results can differ by rounding
+          away from that threshold and differ substantially at or below it: ``[1, 2, 0]`` yields infinities
+          here but finite pixels there. See `#4267 <https://github.com/kornia/kornia/issues/4267>`_.
+          :meth:`unproject` corresponds to :func:`~kornia.geometry.camera.perspective.unproject_points`
+          with ``normalize=False`` and depth shaped as ``(*, 1)`` there instead of ``(*,)`` here.
```
ducha-aiki, line 150: [P2] Qualify the correspondence for batched point clouds

The stated correspondence also needs a camera-axis qualification. With `params.shape == (B, 4)` and pixels shaped `(B, N, 2)`, sensors multiplies/subtracts its `(B,)` intrinsic components directly against `(B, N)` coordinates, so they align with the point axis; geometry inserts a singleton point axis and applies each camera's intrinsics along the batch axis. For two cameras with `params=[[100,50,4,3],[200,80,10,20]]`, a `(2,2,2)` pixel tensor filled with `[104,103]`, and unit depth, sensors returns x coordinates `[[1,.47],[1,.47]]`, whereas `unproject_points(pixels, depth[...,None], cam.matrix())` returns `[[1,1],[.47,.47]]`. Both calls succeed, but their camera associations differ. With `B=2, N=3`, sensors raises a shape error while geometry succeeds. Projection has the same distinction, even at exactly representable `z=4`.

Please scope the correspondence to shared-intrinsics/one-point-per-camera cases, document the batching difference, and add a multi-camera, multi-point pin with unequal intrinsics. The existing equivalence pins use one unbatched camera and cannot detect this. I reproduced the behavior on both this head and merge-base `f36ab6de`; the introduced issue is the documentation claim, so no runtime change is needed here.

ducha-aiki (commented):
Reviewed all eight changed files at `163fbc79474ee6f5d52f8fabd8a80df61c6da259` personally, without subagents, in a detached worktree.

**Verdict: one P2 documentation correction remains before merge.** The attached finding concerns batched point clouds: wrapping points in Vector objects and changing the depth shape does not preserve the camera-axis association between the two APIs. When the camera and point counts happen to match, both calls succeed with different results, making this a consequential qualification for the new Convention block.

The previous projection finding is addressed: the anchor, page, changelog and test now distinguish the shared formula from rounding and zero/near-zero depth behavior. I re-executed the zero, near-zero, threshold-boundary, just-above-threshold and non-power-of-two examples. The new tests correctly pin the correction.

Minimal reproduction for the remaining finding (CPU float32, torch 2.14.0):

```python
import torch
from kornia.geometry.camera import unproject_points
from kornia.geometry.vector import Vector2
from kornia.image import ImageSize
from kornia.sensors.camera import PinholeModel

params = torch.tensor([[100., 50., 4., 3.],
                       [200., 80., 10., 20.]])
cam = PinholeModel(ImageSize(6, 8), params)
pixels = torch.tensor([[[104., 103.], [104., 103.]],
                       [[104., 103.], [104., 103.]]])
depth = torch.ones(2, 2)

print(cam.unproject(Vector2(pixels), depth).data[..., 0])
# tensor([[1.0000, 0.4700], [1.0000, 0.4700]])
print(unproject_points(pixels, depth[..., None], cam.matrix())[..., 0])
# tensor([[1.0000, 1.0000], [0.4700, 0.4700]])
```

The same result reproduces in a separate worktree at merge-base `f36ab6de`, with the explicit interpreter and `kornia.__file__` verified in both trees. This is existing runtime behavior that the newly added correspondence needs to describe accurately.

Validation on the reviewed head, with imports verified inside the worktree:

- Full sensors/camera CPU float32/float64: **92 passed, 29 skipped**.
- Full sensors/camera MPS float32/float16, MPS fallback unset: **92 passed, 29 skipped**.
- CPU float16/bfloat16 convention/wart/scalar-depth/parameter-validation selection: **56 passed**.
- Camera module doctests: **11 passed**.
- Full `pixi run pre-commit-all`: **passed**, worktree remains clean.
- Fresh Sphinx HTML build with `-W`: **passed**; verified the added class entries, the rendered Convention block and its geometry cross-references.
- All three changed runtime files have identical ASTs to the merge base after stripping docstrings. `git diff --check` passes.
- Re-executed all nine placeholder-model operations and verified the documented exception sites/messages. No obsolete #4313/#4316 warning remains.
- Local merge-tree against fetched main `749a1c17` is clean.

No tracked files were changed. CUDA and the full repository test suite were not run. GitHub Actions checks are still queued; Read the Docs and pre-commit.ci report success. This is a COMMENT review because the signed-in account owns the PR.

ducha-aiki:
## Review of #4318 at `6a81e6e0` (2026-09-10, after the two Codex rounds)

**Verdict: approve; merge when CI is green.** Reviewed the whole diff against `main` `749a1c17` by hand, without subagents, in the branch worktree (torch 2.14.0, `kornia.__file__` naming the worktree). Since the re-review at `3ced4c71` (issuecomment-5614918243) two Codex findings landed as `163fbc79` (projection is not byte-identical to `project_points`: reciprocal multiplication versus direct division, plus the `abs(z) <= 1e-8` pass-through) and `f5394499` (batched `(B, 4)` intrinsics broadcast along the point axis here and the camera axis there). Both are correct and both are docs/pins only; I re-executed each, pushed `6a81e6e0` with three nits and refreshed the PR body.

### Re-executed
- `convert_points_from_homogeneous` masks `abs(z) > 1e-8` and multiplies by `1 / z`; `Z1Projection.project` divides. `[5, 2, 3]` on cpu float32: 170.66665649414062 here versus 170.6666717529297 there (the figures in the test comment); float64 differs in the last digit as well. `[1, 2, 0]`: `inf` here, `[104, 103]` there.
- `denormalize_points_with_intrinsics` / `normalize_points_with_intrinsics` `unsqueeze(-2)` the intrinsics when the points carry more dims; `AffineTransform` multiplies `(B,)` against `(B, N)`. `B == N == 2` associates the intrinsics with different points; `B = 2, N = 3` raises `RuntimeError` here and runs there.
- On `main` after #4387/#4381, `PinholeCamera.scale` still applies `cx' = s * cx` (only the shape guards moved) and `camera_matrix` is still the `[..., :3, :3]` block; `distort_points_kannala_brandt` still takes `[fx, fy, cx, cy, k0..k3]` after #4368; `main`'s `conversions.py` delta is axis-angle only.
- #4263's comment thread carries the `ImageSize` observation the warning cites; #4284's Expected asks for a message naming the model, which is what the empty-message asserts guard.

### Fixed in `6a81e6e0`
1. The `PinholeModel.scale` warning said a tensor factor gives "0-dim floating tensors"; `torch.tensor(2)` gives int64 ones. Scoped to a floating-point factor.
2. `test_convention_params_length_is_fixed_per_camera_model_type` pinned the inconsistent shape spelling of the `ValueError` message (`(B, 4)` versus `B, 8`) in its two explicit matches; a convention pin should not fail on a message-format cleanup. They now match the model name only; the measured spellings stay in the comment.
3. The rst `.. meta::` description still said "currently supporting only the Pinhole model"; the body had deferred it as a follow-up, but it is the same file and the same staleness as the note.

### Verified on `6a81e6e0`
- cpu `tests/sensors/camera --dtype=float32,float64,float16,bfloat16`: 200 passed, 55 skipped; per-dtype half runs with the CI profiles: float16 50 passed, bfloat16 49 passed + 1 xfailed. mps float32/float16: the 64-node convention/wart/validation selection green after the edits; the full run at `f5394499` was 99 passed, 29 skipped, 1 failed in the unseeded `test_project_unproject[mps-float16-5]`, which fails 1 run in 6 on `main` `f36ab6de` in a base worktree.
- doctest 11 passed; pinned ruff 0.16.6 check + format clean; AST-after-docstring-strip identical to `main` for all three `kornia/` files; no `>>>` in the `kornia/` diff; no line over 120.
- Merge with `main` `749a1c17` is clean (only `CHANGELOG.md` overlaps). On the merged tree: cpu float32/float64 100 passed; the AGENTS grep for `#4313|#4316` is empty; `sphinx -W` zero warnings; on the rendered `sensors.camera.html` every xref this PR adds is a live link, including `perspective.project_points` / `unproject_points`, `PinholeCamera.scale` / `.camera_matrix`, `distort_points_affine`, `distort_points_kannala_brandt`, `project_points_orthographic` and `calibration.distort_points`. The only dead xrefs on the page are pre-existing (`Vector2` / `Vector3` in signatures, the `:mod:` refs).
- #4274's body still says the two Pinhole paths "agree exactly"; the three measured exceptions are now recorded there (issuecomment-5616135040) with the pin names.

### Still open, unchanged
- Pin comments cite audit labels and "this worktree" (batch-wide); the 17-private-test-names-in-public-warnings question from #4317 round 6 — both the user's call.
- #4377 (fixes #4284) collides with the four `_4284` pins, the seven prose sites and the rst note; whichever lands second drops them.

CI at `f5394499` was still queued 20 minutes after its push; `6a81e6e0` restarts it. `3ced4c71` was fully green.

PR #4319, opened 2026-09-06
fix(feature): register RenderingDeFMO's times as a buffer so .to() moves it

tritsystem:
## Problem

`RenderingDeFMO` (used by `DeFMO`) holds its rendering time-steps as a plain Python attribute, not a registered buffer, so `nn.Module.to()` never touches it:

```python
from kornia.feature.defmo import RenderingDeFMO
m = RenderingDeFMO().half()
print(next(m.parameters()).dtype, m.times.dtype)
```
```
torch.float16  torch.float32
```

`forward` compensates by re-deriving `times`'s **device** from the input on every call, but never its **dtype**:

```python
times = self.times.to(latent.device).unsqueeze(0).repeat(latent.shape[0], 1)
```

So a half-precision module crashes on the very first call:

```python
latent = torch.rand(1, 2048, 4, 4, dtype=torch.float16)
m(latent)
```
```
RuntimeError: Input type (torch.FloatTensor) and weight type (torch.HalfTensor)
should be the same or input should be a MKLDNN tensor and weight is a dense tensor
```

This is the exact same bug shape as #4069/#4079 (`SIFTDescriptor.gk`, `PatchDominantGradientOrientation.weighting`, `PatchAffineShapeEstimator.weighting`) in a different class — `#4069` explicitly scoped to those three, and `#4068` covered `DenseSIFTDescriptor` separately; neither mentions `defmo.py`.

## Change

`register_buffer("times", ..., persistent=False)`, same convention and rationale as #4079: `times` is fully determined by `tsr_steps` (derived state, not learned), and `persistent=False` keeps `state_dict()` keys unchanged so existing checkpoints keep loading under `strict=True`. `forward` now casts into a **local** variable rather than depending on the caller having matched dtype, mirroring the local-cast pattern #4079 established (not rebinding `self`).

## Behaviour

Unchanged in the normal path. Byte-identical (`torch.equal`) to the pre-fix module with the same seed and input:

```
fixed  shape: torch.Size([1, 24, 4, 64, 64]) sum: 201291.984375
original shape: torch.Size([1, 24, 4, 64, 64]) sum: 201291.984375
byte-identical: True
```

## Tests

Added `TestRenderingDeFMOTimesBuffer` (`tests/feature/test_defmo.py`), mirroring `TestSIFTDescriptorKernelBuffer`'s style from #4079:
- `test_to_moves_times` — `.to(device, torch.float16)` carries `times` along
- `test_times_stays_out_of_state_dict` — non-persistent, checkpoints unaffected
- `test_forward_does_not_mutate_the_module`
- `test_half_precision_forward_no_longer_crashes` — the actual reported defect
- `test_matches_pre_fix_output_in_the_normal_float32_path` — pins the exact checksum above

Confirmed failing-then-passing: `test_to_moves_times` and `test_half_precision_forward_no_longer_crashes` fail with the exact RuntimeError above against the pre-fix code, pass after.

`ruff check` / `ruff format --check` clean on both changed files.

## AI-assistance disclosure

This fix was found via an automated static-analysis tool (arch-finder, a personal project) that flags a tensor stashed as a plain `self.<attr>` in `__init__` and corroborates it by finding a later method papering over the gap by re-deriving that attribute's device/dtype from its own arguments — the same shape as #4069/#4079. Claude (Anthropic) was used to write and verify the reproduction, the fix, and the tests, including the byte-identical comparison against the pre-fix module and confirming the tests fail on the pre-fix code before passing on the fix. The finding, the reproduction, and the test results were independently verified by the human researcher via direct execution before this PR was opened, not taken on the tool's or the model's word alone.


ducha-aiki (changes requested):
Review at `5bff982e` — implementation verified; documentation follow-up requested.

The buffer fix is correct and preserves checkpoint compatibility. At head, .half() moves times to float16, registers it as a buffer, keeps it out of state_dict, and a half forward returns shape (1,24,4,64,64), dtype float16. At merge-base e8111a61, times remains an unregistered float32 attribute and the same forward raises RuntimeError: expected scalar type Float but found Half. TorchScript half forward also succeeds. A normal float32 forward comparison is byte-identical to base.

The focused new test class passed 4 tests with the slow case skipped by marker; that half-forward case was also executed directly. Current CI is green.

One required follow-up: add an Unreleased / Bug fixes changelog entry citing #4319. This changes a user-visible half-precision runtime error into working output. No code changes are requested; I recommend approval after the release note.

ducha-aiki:
Delta re-review at `231c9b55` — **APPROVED**; one mechanical step left. The delta is the requested CHANGELOG entry only (`+9` under `### Bug fixes`), code and tests unchanged from `5bff982e`.

GitHub now reports the branch `CONFLICTING`/`DIRTY`, so no `pull_request` workflow ran on this head (that is what the `pre-commit.ci - pr: error during mergeable check` is). A `git merge-tree` of your head against `main` is clean here — the "conflict" is GitHub's CHANGELOG-adjacency detection — but a dirty PR runs no CI, so please rebase onto `main` (or merge it in) and push; once the new head is green this can merge.

tritsystem:
Thanks for the approval and for tracking down exactly what's going on with the CONFLICTING/DIRTY status — appreciate you confirming `git merge-tree` comes back clean.

I can't push a rebase from my end here: bringing the branch up to date with `main` pulls in two unrelated workflow-file changes (`coverage.yml`, `scheduled_test_pypi_package.yml`), and my push token doesn't carry the `workflow` OAuth scope needed to push those, so both a plain `git push` and the fork-sync API reject it — same situation as the CHANGELOG-only conflict on #4210 a couple of days ago, just with workflow files in the mix this time instead of only `CHANGELOG.md`.

Would you (or anyone with the right scope) be able to rebase/merge `main` into this branch from your side, the way #4210 got resolved via the web merge editor? Code and tests are unchanged from `5bff982e` — the only delta needed is bringing the branch current so CI can actually run. Happy to redo anything if it's easier for you to have me try a different approach instead.

tritsystem:
Rebased onto current `main` and pushed — the branch is `MERGEABLE` now (was `CONFLICTING`/dirty), and the full CI matrix is running for the first time on this head.

Only `CHANGELOG.md` actually conflicted (three adjacent-entry ordering conflicts, purely additive on both sides — resolved by keeping both sides' entries in list order, no wording changed). `defmo.py` and `test_defmo.py` merged cleanly with no conflicts. Code and tests are unchanged from `5bff982e`/`231c9b55`.

Re: the `workflow`-scope push failure mentioned above — this push went through fine from `tritsystem/kornia` (my own fork), even though it also pulled in the two workflow-file changes from `main` (`coverage.yml`, `scheduled_test_pypi_package.yml`). Not sure why the earlier attempt hit the scope error and this one didn't; possibly a difference in exactly what was being pushed. Flagging in case it's useful context, not asserting a root cause I haven't verified.

Will report back once the full matrix finishes.

tritsystem:
Thanks for the review — the requested Unreleased/Bug fixes changelog entry citing #4319 has been added (it's already on this branch, in the commit right after your review). Happy to adjust the wording if it needs anything else. Re-requesting review when convenient.

ducha-aiki:
Delta at `79ce4abb` — still **APPROVED, merge-ready**.

The last two commits are the CHANGELOG resync and a merge of `origin/main`. The code+test patch-id is `f9403116…` both at the approved `231c9b55` and here, so the buffer fix and its tests are untouched; the CHANGELOG diff against `main` adds the entry under `### Bug fixes` and deletes nothing. CI is 40/40 green and the PR is now `MERGEABLE`, which was the one mechanical step left.

The PR still shows "changes requested" from the review at `5bff982e`; every item from that round landed at `231c9b55`, so I am clearing it with an approving review.

ducha-aiki (approved):
Approved at `79ce4abb`: code+test patch-id identical to the approved `231c9b55`, CI 40/40 green, `MERGEABLE`. This clears the stale changes-requested state from the `5bff982e` round, whose items all landed.

ducha-aiki (dismissed):

PR #4326, opened 2026-09-06
fix(io): follow kornia_rs 0.1.11's move of image I/O into kornia_rs.io

### LINKED ISSUE #4325: kornia.io is broken on kornia_rs >= 0.1.11 (the version `pip install kornia` resolves): image I/O moved into `kornia_rs.io`
`pip install kornia` in a fresh environment resolves the newest `kornia_rs` (0.1.14, released 2026-05-19), and on that version `kornia.io` no longer works:

```
>>> import kornia
>>> kornia.io.load_image("x.jpg", kornia.io.ImageLoadType.RGB8)
AttributeError: module 'kornia_rs' has no attribute 'read_image_jpegturbo'. Did you mean: 'read_image_jpeg'?
```

`write_image` fails the same way for every container (`write_image_png_u8`, `write_image_tiff_*`, `write_image_png_u16`), and PNGs whose header is not RGB fail on `read_image_png_u8`. Only the RGB-PNG path survives, because the deprecated root-level `read_image_any` still exists (it now warns and points to `kornia_rs.io.read_image`).

## Cause

kornia_rs 0.1.11 (2026-05-04) moved every `read_image_*`/`write_image_*` function from the package root into the `kornia_rs.io` submodule, with the same names and signatures. `kornia/io/io.py` calls them on the root module, and `pyproject.toml` declares `kornia_rs>=0.1.9` with no upper bound. The lock pins 0.1.10, so CI never runs the new layout, and nothing in the repo exercises what users actually get:

| | kornia_rs | `load_image("x.jpg")` |
|---|---|---|
| `uv sync` (lock) | 0.1.10 | works |
| `pip install kornia` (0.8.3, 2026-05-19) | 0.1.14 | `AttributeError` |
| `pip install .` from `main` (`0743e29e`) | 0.1.14 | `AttributeError` |

Verified by installing `main` into a fresh Python 3.12 venv with only the declared runtime dependencies, and by diffing `dir(kornia_rs)` across 0.1.10, 0.1.11, 0.1.12, 0.1.13 and 0.1.14 (`kornia_rs.io` holds the whole old API from 0.1.11 on; 0.1.10 has no `io` attribute at all).

## Why nobody noticed

* The nightly `Test PyPI package` workflow does `pip install kornia` and then only prints `kornia.__version__`; a bare import never touches kornia_rs functions, so it has been green since May.
* Dependabot has not proposed a kornia-rs bump since #2830 (2024): the `pip` ecosystem has nothing to bump on an unpinned `>=0.1.9`, and the lock is not on its radar.
* The dependency audit (#4259, #4261, the twelve PRs merged today) looked at what kornia imports, not at whether the declared version range still matches the imported API. This is the one thing that pass missed.

## Fix

Resolve the namespace once at import (`kornia_rs.io` when it has the readers, the root otherwise), keep `read_image_any` for 0.1.9/0.1.10 and use `kornia_rs.io.read_image` on newer versions (verified byte-identical on RGB/mono/RGBA PNG, JPEG and TIFF), bump the lock to 0.1.14 so CI runs the layout users get, and make the nightly PyPI job round-trip an image through `write_image`/`load_image` instead of only importing. PR follows.

ducha-aiki:
Closes #4325.

## What was broken

`pip install kornia` resolves the newest `kornia_rs` (0.1.14). kornia_rs 0.1.11 (2026-05-04) moved the image readers and writers from the package root into `kornia_rs.io`, with the same names and signatures, and left at the root only a deprecated `read_image_any` plus the uint8 JPEG/PNG writers. 0.1.12 and 0.1.13 have no `read_image_jpegturbo` at all; 0.1.14 has it back in `io`. `kornia/io/io.py` called everything on the root, so on 0.1.11 and newer:

```
>>> kornia.io.load_image("x.jpg", kornia.io.ImageLoadType.RGB8)
AttributeError: module 'kornia_rs' has no attribute 'read_image_jpegturbo'. Did you mean: 'read_image_jpeg'?
```

Every JPEG load, every PNG load that is not plain RGB, and every TIFF, 16-bit or float32 `write_image` failed this way (8-bit JPEG/PNG writes still worked). The lock pins 0.1.10, so CI never saw it, and the nightly PyPI job only imports kornia. Released 0.8.3 is affected too.

## Change

* `pyproject.toml`: `kornia_rs>=0.1.14` (was `>=0.1.9`). The five released layouts between 0.1.9 and 0.1.14 do not share a call site that works everywhere (0.1.12/0.1.13 have no libjpeg-turbo reader at all), and a per-version resolver would have kept a floor that no CI job exercises. Recorded under *Breaking changes*.
* `kornia/io/io.py`: binds `kornia_rs.io` at import and calls it directly; `kornia_rs.io.read_image` replaces the deprecated root `read_image_any`. The decoder also reads 16-bit PNG/TIFF and float TIFF, which 0.1.10 rejected at decode: `ImageLoadType.UNCHANGED` returns them as `uint16`/`float32`, and the 8-bit load types raise a `NotImplementedError` that names the decoded dtype instead of mislabelling the tensor. `_color_mode` replaces five copies of the mono/rgb expression.
* `uv.lock`: `kornia-rs` 0.1.10 → 0.1.14, so every CI job runs the same layout users get, and the lock now equals the floor.
* `tests/io/test_io_image.py`: the set of `kornia_rs.io` functions `io.py` calls is derived from the source with an `ast` walk and checked against the installed kornia_rs (so a new call site cannot go unchecked), uint8 round trips through `jpg`/`png`/`tiff` with pixel equality for the lossless ones, call-time lookup (patching `kornia_rs.io` is seen), and the 16-bit/float decode contract. The RGBA fixture writes through `kornia_rs.io` instead of the deprecated root function. On unfixed `main` with kornia_rs 0.1.14 the existing `test_smoke` already fails with the `AttributeError` above.
* `.github/workflows/scheduled_test_pypi_package.yml`: both steps now run from `${{ runner.temp }}`. Until now `python` ran from the checkout root, so `import kornia` resolved the source tree: the 2026-09-06 log shows `kornia-0.8.3` installed and `Kornia version: 0.9.0rc1` printed, i.e. the job never tested the wheel. After the import check it round-trips an image through `write_image`/`load_image` for the three writable containers, with pixel equality for png/tiff. **This step is red against the released 0.8.3 until the next release ships this fix**; that is the signal the job was missing since May. Drop the step if a red nightly for that window is not acceptable.

## History

The first cut (`71af284a`) kept the `>=0.1.9` floor with a resolver over the namespaces; its review (comment below) found that the resolver missed 0.1.12/0.1.13, that the nightly job imported the checkout, and that the lock bump widened `load_image`'s inputs without a contract. `3f96097c` fixed those with a per-name resolver verified on 0.1.10/0.1.12/0.1.13/0.1.14; the maintainer then chose to raise the floor instead, which `HEAD` does.

## Verified

Against a kornia install built from this branch in a fresh py3.12 venv with kornia_rs 0.1.14: `tests/io`, `tests/core/test_module.py`, `tests/test_declared_imports.py` and `pytest --doctest-modules kornia/io`, 62 passed under `-W error::DeprecationWarning`. `ty check kornia` reports the same 146 diagnostics as `main`; pre-commit is clean on the changed files.


ducha-aiki:
## Review of `71af284a` (first cut): changes requested, fixes pushed as `HEAD`

Every finding below was re-executed against real kornia_rs wheels (0.1.10, 0.1.12, 0.1.13, 0.1.14 in separate venvs) or the nightly job's log before it was acted on. All are fixed on the branch; the PR body was rewritten to match.

**Defects**

1. **0.1.12 and 0.1.13 were still broken.** The resolver chose `kornia_rs.io` only if it had `read_image_jpegturbo`; those two releases have the full `io` namespace except that one function, so they were routed to the package root and `test_smoke` failed with the same `AttributeError` as before (`_rs_io is kornia_rs`, 9 of 28 tests/io red on 0.1.12). Both satisfy `kornia_rs>=0.1.9` and have wheels for every supported Python. Fix: per-name, call-time lookup over (`io`, root) and a generic-decoder fallback for JPEG; verified 39 passed on 0.1.12 and 0.1.13.
2. **The nightly PyPI job never tested the wheel.** Both steps ran `python` from the checkout root, so `sys.path[0]` was the source tree: the 2026-09-06 log shows `Successfully installed kornia-0.8.3 kornia-rs-0.1.14` followed by `Kornia version: 0.9.0rc1`. The first cut's claim that the new step "will be red against 0.8.3" was therefore false; it would have gone green while every `pip install kornia` user kept the bug. Fix: `working-directory: ${{ runner.temp }}` on both steps, `kornia.__file__` printed, pixel equality for png/tiff. The step is now genuinely red against 0.8.3 until a release ships the fix.
3. **Moving the lock to 0.1.14 widened what `load_image` accepts without a contract.** 0.1.11+ decodes 16-bit PNG/TIFF and float TIFF, which 0.1.10 rejected at decode, so non-uint8 arrays reached `_convert_image_type`: `RGB32` on a float TIFF raised a bare `Validation condition failed`, `RGB8` returned a float32 tensor labelled RGB8, `GRAY8` on a uint16 PNG raised `Unknown data type: torch.uint16` from `rgb_to_grayscale`. Fix: `UNCHANGED` returns the decoded dtype (pinned bit-exact for uint16 png/tiff and float32 tiff); the 8-bit load types raise a `NotImplementedError` that names the dtype and points at `UNCHANGED`.
4. **Wording overstated the defect and the fix.** "Every `write_image` raised" was false (the uint8 JPEG/PNG writers stayed at the root on 0.1.11+), and "works again on 0.1.11 and newer" was false for 0.1.12/0.1.13. CHANGELOG, docstrings, workflow comment and PR body now state the real scope: JPEG loads, non-RGB PNG loads, TIFF/16-bit/float32 writes; all five layouts.

**Robustness / maintainability**

5. The any-format reader was bound at import while the other calls resolved at call time, so patching `kornia_rs.read_image_any` was invisible to `load_image`. All lookups are now at call time; pinned by a test that patches the writer and checks the spy is hit.
6. The test's `USED` tuple duplicated the call sites by hand. It is now derived from `io.py` with an `ast` walk over `_rs_io.<name>` attributes and `_rs_io.find("<name>")` arguments, so a new call site is checked automatically.
7. The nightly round trip only compared shapes; it now checks `torch.equal` for the lossless containers, matching the in-repo test.
8. Five copies of the mono/rgb `mode` expression collapsed into `_color_mode`; the `ndim == 2` arm was dead because `write_image` already adds the channel axis.
9. Comments in `_load_image_to_tensor` and on PNG colour type 4 described `read_image_any` and claimed gray+alpha falls through "as RGB"; verified it raises on both layouts. Rewritten.
10. Found by the doctest run afterwards, not by the review: the resolver's `__getattr__` turned `inspect`'s `__wrapped__` probe into an `ImportError` and broke `pytest --doctest-modules kornia/io` collection. Dunder lookups now raise `AttributeError`; pinned.

**Left open, on purpose:** the `>=0.1.9` floor is exercised only by the resolver tests and the by-hand runs in the PR body, not by a CI cell. The simpler shape is to raise the floor to 0.1.14 (wheels exist for cp310–cp314) and delete the resolver; that changes what users may have installed, so it is a maintainer decision rather than part of this fix.

Verdict after the fixes: approve. `tests/io` 31/39/39/63 passed on kornia_rs 0.1.10/0.1.12/0.1.13/0.1.14, doctests, `tests/test_declared_imports.py`, `ty` (146 diagnostics, same as `main`) and pre-commit clean. CI on the amended head is still queued at the time of posting.

ducha-aiki:
Maintainer decision: raise the floor. `HEAD` requires `kornia_rs>=0.1.14`, deletes the resolver and its layout tests, binds `kornia_rs.io` directly, and records the floor bump under *Breaking changes*. Lock now equals the floor, so CI exercises exactly what is declared. Re-verified on 0.1.14: 62 passed (tests/io, test_module, declared-imports pin, io doctests) with deprecation warnings as errors; `ty` unchanged at 146.

PR #4327, opened 2026-09-06
test: drop version guards that the torch>=2.5.1 floor made dead; CUDA pixi tasks stop reinstalling torchvision

ducha-aiki:
Follow-up to the dependency cleanup (#4259, #4261): three leftovers seen during the final pass, all dead under the current floors (`requires-python >= 3.11`, `torch >= 2.5.1`, `kornia_rs` a hard runtime dependency).

* `tests/io/test_io_image.py`: the module wrapped `import kornia_rs` in a `try`, and `available_package()` skipped the whole `TestIoImage` class unless Python >= 3.7, torch >= 1.10 and kornia_rs imported. None of those can be false any more; the guard, the `sys`/`torch_version_ge` imports it needed and the `skipif` are gone. `kornia_rs` is a plain import now, as in `conftest.py`.
* `tests/filters/test_dissolving.py` and `TestRandomDissolving` in `tests/augmentation/test_augmentation.py`: both classes were skipped on `torch <= 2.0.1` ("transformers clip model needs distributed tensor"). Same guard, two files; both decorators and the now-unused `torch_version_le` imports are removed. The classes stay `@pytest.mark.slow` behind the `diffusers` `importorskip`.
* `pixi.toml`: the CUDA `install`/`install-docs` tasks reinstalled `torch torchvision` from the cu121 index. torchvision is not a kornia dependency (the last test that used it went in #4289) and nothing in the repo imports it, so the task was installing an undeclared package into the CUDA environment. Now `torch` only.

Not touched, on purpose: the other dead version guards in `tests/` (`torch_version_le(1, 9, 1)` in eight files, `torch_version_lt(1, 11, 0)`, `torch_version_lt(2, 0, 0)` twice, the `py3.8` skips in `tests/models/test_tiny_vit.py` and `tests/geometry/test_conversions.py`, and the `< 2.2`/`< 2.3`/`< 2.4` CPU-half and ONNX guards). They are the same kind of dead code; this PR is limited to the three items called out in the review.

Verified: `tests/io` 21 passed; the two dissolving classes still collect and skip without `--runslow`; `pixi.toml` parses and the CUDA task reads `uv sync --extra dev && uv pip install --reinstall torch --index pytorch-cu121`; pre-commit clean on the changed files.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -75,8 +75,8 @@ platforms = ["linux-64", "linux-aarch64", "win-64"]
 # ------------------------------------------------------------------------------
 
 [feature.cuda.tasks]
-install = { cmd = "uv sync --extra dev && uv pip install --reinstall torch torchvision --index pytorch-cu121", description = "Install dev dependencies (CUDA PyTorch)" }
-install-docs = { cmd = "uv sync --extra dev --extra docs && uv pip install --reinstall torch torchvision --index pytorch-cu121", description = "Install dev + docs (CUDA PyTorch)" }
+install = { cmd = "uv sync --extra dev && uv pip install --reinstall torch --index pytorch-cu121", description = "Install dev dependencies (CUDA PyTorch)" }
+install-docs = { cmd = "uv sync --extra dev --extra docs && uv pip install --reinstall torch --index pytorch-cu121", description = "Install dev + docs (CUDA PyTorch)" }
 # CUDA test tasks
 test-cuda = { cmd = "uv run pytest tests/", env = { KORNIA_TEST_DEVICE = "cuda" }, description = "Run tests on CUDA" }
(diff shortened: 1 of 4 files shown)


ducha-aiki (commented):
Reviewed commit 41e30ba1a94d6df611a661b40af9c49489fd91f8. No actionable findings.

The removed guards are redundant under the declared Python >=3.11, torch >=2.5.1 and mandatory kornia_rs requirements. The optional diffusers and slow-test gates remain intact. An AST scan found no executable torchvision imports in kornia/ or tests/, and both edited CUDA task definitions parse correctly.

Validation: ran tests/io, tests/filters/test_dissolving.py and tests/augmentation/test_augmentation.py::TestRandomDissolving against the exact PR checkout through Pixi, with CPU float32 and torch 2.14.0: 21 passed, 7 skipped. Verified kornia.__file__ resolves to that checkout. CUDA installation and slow model execution were not exercised on this Mac; the full CI matrix was still pending when checked.

PR #4328, opened 2026-09-06
refactor: import Self, ParamSpec, TypedDict and NotRequired from typing instead of typing_extensions

ducha-aiki:
Follow-up to the dependency cleanup (#4259). `kornia/` imported `typing_extensions` in five files without declaring it; `tests/test_declared_imports.py` tolerated that as the single `IMPLICIT_ALLOWED` entry on the grounds that torch requires it. With `requires-python >= 3.11` every name kornia used is in the standard library, so the import can simply go:

| file | name | in `typing` since |
|---|---|---|
| `kornia/augmentation/auto/operations/base.py` | `Self` | 3.11 |
| `kornia/augmentation/container/ops.py` | `ParamSpec` | 3.10 |
| `kornia/feature/keynet.py`, `kornia/feature/scale_space_detector.py` | `TypedDict` | 3.8 |
| `kornia/feature/adalam/core.py` | `TypedDict`, `NotRequired` | 3.11 |

`IMPLICIT_ALLOWED` becomes an empty `frozenset` with a comment saying why it should stay empty; the test that checks its premise (torch must require every entry) is kept for the day someone adds one.

What this does and does not buy: nothing changes in what gets installed, torch still pulls `typing_extensions` in. What changes is that kornia no longer imports a package on another package's word, so the declared-imports pin has no exception left.

Verified: 725 tests over the five modules' suites (`test_keynet`, `test_scale_space_detector`, `test_matching` for AdaLAM, `test_auto_operation`, `augmentation/container`) plus `tests/test_declared_imports.py` pass; `ty check kornia` reports the same 146 diagnostics as `main`; pre-commit clean.

### DIFF
--- a/kornia/augmentation/auto/operations/base.py
+++ b/kornia/augmentation/auto/operations/base.py
@@ -15,12 +15,11 @@
 # limitations under the License.
 #
 
-from typing import Callable, Dict, List, Optional, Tuple, TypeVar
+from typing import Callable, Dict, List, Optional, Self, Tuple, TypeVar
 
 import torch
 from torch import nn
 from torch.distributions import Bernoulli, RelaxedBernoulli
-from typing_extensions import Self
 
 from kornia.augmentation.base import _AugmentationBase
 
--- a/kornia/augmentation/container/ops.py
+++ b/kornia/augmentation/container/ops.py
@@ -17,11 +17,10 @@
 
 import copy
 from abc import ABCMeta, abstractmethod
-from typing import Any, Callable, Dict, Generic, List, Optional, Type, TypeVar, Union
+from typing import Any, Callable, Dict, Generic, List, Optional, ParamSpec, Type, TypeVar, Union
 
 import torch
 from torch import nn
-from typing_extensions import ParamSpec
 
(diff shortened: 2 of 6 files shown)


ducha-aiki (commented):
Reviewed commit 8d1891a11bf7a856b763664eb785aab87f1ef3e3. No actionable findings.

All four typing names are available at the declared Python >=3.11 minimum. Compared base/head runtime type-hint resolution for OperationBase.train/eval and make_input_only_sequential, and resolved hints plus required/optional key metadata for AdalamConfig, KeyNet_conf and Detector_config: they match semantically. Removing the now-unused declared-import exception is consistent with the source changes.

Validation: tests/test_declared_imports.py and tests/augmentation/test_auto_operation.py passed (29 tests) against the exact PR checkout on Python 3.11.14 / torch 2.14.0, with kornia.__file__ confirmed to resolve to that checkout. git diff --check was clean. Also confirmed the standard-library symbols exist on Python 3.12 and 3.13. The full affected suite and full runtime tests on those additional Python versions were not rerun.

Issue #4322, opened 2026-09-06
ModelBase.save writes float visualizations to .png, which write_image rejects: every container's save raises


ducha-aiki:
## Summary

`ModelBase.save` (inherited by `SemanticSegmentation`, `ObjectDetector`, `EdgeDetector`, `DepthEstimation`) hands the float output of `visualize` to `write_image` under a hard-coded `.png` name, and `write_image` only writes `.png` for `uint8`/`uint16` images. So every container's `save` raises on its first `_save_outputs` call, the directory is created and left empty, and nothing is written.

## Reproduce

```python
import tempfile, torch
from torch import nn
from kornia.models.segmentation import SegmentationModelsBuilder

model = SegmentationModelsBuilder.build(nn.Sequential(nn.Conv2d(3, 3, 1), nn.Softmax(dim=1)), None)
model.save(torch.rand(1, 3, 6, 6), directory=tempfile.mkdtemp())
# NotImplementedError: Unsupported file extension: .png for float32 image   (kornia/io/io.py)
```

float64 / float16 inputs fail one step earlier with `Unsupported image dtype`. Observed on kornia at #4301's head (`9f7a23d0`, torch 2.14.0, CPU); `kornia/models/base.py::_save_outputs` is unchanged on `main`, so the other three containers are the same.

## Where

- `kornia/models/base.py::_save_outputs` — `write_image(os.path.join(directory, f"...png"), out)` with `out` float.
- Every `visualize` returns float (`draw_rectangle` / `add_weighted` / the colormap gather), so no caller can reach a `uint8` write.

## Fix options

1. Convert in `_save_outputs`: `(out.clamp(0, 1) * 255).round().to(torch.uint8)` for floating inputs, then write `.png`.
2. Or choose the extension by dtype (`.tiff`/`.exr` for float), which changes what users get on disk.

Option 1 keeps the documented "saves a PNG" behaviour. It needs a test per container that actually writes into a `tmp_path` and reads the file back; today no test calls `save` on any of the four (which is how this survived). #4301 stops advertising `save` on the `SemanticSegmentation` docs entry until this is fixed.

Refs #4301.

Posted on behalf of @ducha-aiki by Claude (Claude Fable 5.1).

PR #4330, opened 2026-09-06
fix(enhance): register Normalize/Denormalize/Rescale constants as buffers

### LINKED ISSUE #4323: Normalize/Denormalize/Rescale constants are not buffers: .to(device) leaves them on the CPU and export from an accelerator fails
## Summary

`kornia.enhance.Normalize`, `Denormalize` and `Rescale` keep their constants (`mean`, `std`, `factor`) as plain tensor attributes, not buffers. `module.to(device)` therefore does not move them, and exporting a module that lives on an accelerator fails inside `torch.export` with a device mismatch, although the same module exports fine from the CPU and runs fine in eager (a 0-d / broadcastable CPU tensor combines with a CUDA or MPS tensor in eager, but `torch.export` traces with fake tensors that refuse the mix).

## Reproduce

```python
import torch
from kornia.models.segmentation import SegmentationModelsBuilder

params = {"input_space": "BGR", "input_range": [0, 255], "mean": [0.485, 0.456, 0.406], "std": [0.229, 0.224, 0.225]}
pipe = SegmentationModelsBuilder.get_preprocessing_pipeline(params).to("mps")   # or "cuda"
x = torch.rand(1, 3, 6, 6, device="mps")
pipe(x)                                                     # eager: fine
print(pipe[2].mean.device)                                  # cpu -- .to() did not move it
torch.onnx.export(pipe, (x,), dynamo=True, opset_version=18)
# FakeTensorDeviceMismatchError: Expected all tensors to be on the same device, but found at least two devices, mps:0 and cpu!
```

Seen on torch 2.14.0 / MPS at #4301's head (`9f7a23d0`); `Normalize`'s storage is unchanged on `main`. The `mean=None, std=None` variant of that pipeline (a `Normalize(mean=tensor(0.), std=tensor(1.))` with 0-d constants) fails to export even from the CPU with an `IndexError` in `kornia/enhance/normalize.py` (the export branch indexes `mean[0]`), which is a second, independent defect on the same lines.

## Suggested fix

Register the constants with `self.register_buffer(name, value, persistent=False)`: `.to()` / `.cuda()` then move them, `torch.export` sees one device, and `persistent=False` keeps them out of `state_dict`, so checkpoints saved before and after the change stay loadable in both directions. Add a `tests/onnx` case that moves a `Normalize` / `Rescale` to the test `device` before exporting, so the accelerator path is pinned (today's export tests and the export-support survey run on the CPU only); #4301's new `test_preprocessing_onnx_export` skips off-CPU for exactly this reason and should lose that skip with the fix.

Refs #4301.

dchaudhari7177:
Fixes #4323.

`mean`, `std` and `factor` were plain tensor attributes, so `.to(device)` did not move them — a module built on the CPU and moved to an accelerator kept CPU constants.

Eager tolerates that (a broadcastable CPU tensor combines with a CUDA or MPS one), which is why it went unnoticed. `torch.export` does not: it traces with fake tensors and refuses the mix, so exporting a preprocessing pipeline living on an accelerator failed while the same pipeline exported fine from the CPU.

I have no accelerator here, so I reproduced it through dtype, which `.to()` propagates the same way:

```python
n = Normalize(torch.zeros(3), torch.ones(3)).to(torch.float64)
n.mean.dtype        # before: torch.float32   <- .to() did nothing
                    # after:  torch.float64
```

## Two decisions worth review

**`persistent=False`.** These are constructor arguments, not learned state. A persistent buffer would add them to `state_dict()`, so every existing checkpoint would report unexpected keys — for values the constructor already supplies. With `persistent=False` the constants move with `.to()` and `state_dict()` stays `{}`, so nothing about loading changes. There's a test pinning that.

**`Denormalize` now coerces scalars.** A float cannot be a buffer, and `Denormalize.__init__` accepted `Union[torch.Tensor, float]` and stored it raw. It now coerces, which `Normalize` already did with the same argument.

That is a visible change: `Denormalize.__repr__` shows tensors now.

```
before:  Denormalize(mean=[0.5], std=[0.1])
after:   Denormalize(mean=tensor([0.5000]), std=tensor([0.1000]))
```

`TestDenormalize::test_smoke` asserted the old string, so it is updated — to exactly the form `TestNormalize::test_smoke` was already using (`Normalize(mean=tensor([[0.5000]]), std=tensor([[0.1000]]))`). So this makes the two classes agree rather than inventing a third convention. Flagging it explicitly since it is the only user-visible change beyond the fix.

## Tests

Thirteen new tests in `TestNormalizeConstantsAreBuffers`. **Six fail on `main`:**

```
FAILED test_constants_are_registered_buffers      - Normalize registered no buffers
FAILED test_to_moves_the_constants[Normalize]     - Normalize.mean stayed torch.float32 after .to(float64)
FAILED test_to_moves_the_constants[Denormalize]   - Denormalize.mean stayed torch.float32 after .to(float64)
FAILED test_to_moves_the_constants[Denormalize-float] - Denormalize-float.mean is not a tensor
FAILED test_to_moves_the_constants[Rescale]       - Rescale.factor stayed torch.float32 after .to(float64)
FAILED test_denormalize_coerces_scalars
```

One detail I'd point at: `test_to_moves_the_constants` asserts on the attributes **by name**, not by iterating `named_buffers()`. My first version iterated, and it passed on `main` vacuously — a module that registered no buffers has nothing to iterate, which is precisely the bug. Worth the extra parametrize.

```
pytest tests/enhance        367 passed, 15 skipped
ruff 0.16.4 check / format  clean   (matching .pre-commit-config.yaml)
```

Not addressed here: #4322 (`ModelBase.save`) touches the same `kornia.models` pipelines but is an unrelated failure in `write_image`, so I left it separate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,17 +135,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   tests — a physical M1 on macOS 26 compiles them fine. kornia still supports torch 2.5.1, so the
   in-tree MPS workarounds stay. (#4202)
 
-### Fixed
-
-* `Normalize`, `Denormalize` and `Rescale` register their constants (`mean`, `std`, `factor`) as
-  non-persistent buffers instead of plain attributes, so `.to(device)` moves them with the module.
-  Previously they stayed on the CPU: eager tolerates the mix, but `torch.export` traces with fake
-  tensors and refused it, so exporting a preprocessing pipeline from an accelerator failed while
-  the same pipeline exported fine from the CPU. `Denormalize` now coerces a scalar `mean`/`std` to
-  a tensor, as `Normalize` already did, which changes its `__repr__` to match `Normalize`'s. The
-  buffers are non-persistent, so `state_dict()` is unchanged and existing checkpoints still load.
-  (#4323)
-
 ### Breaking changes
 
 * `kornia_rs>=0.1.14` is required; the floor used to be 0.1.9. kornia_rs 0.1.11 relocated its image I/O
@@ -369,6 +358,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed `unproject_points_z1` depth shape handling for singleton and multi-axis batches,
   accepting both trailing-singleton and flat depth tensors. (#4355)
+
+* `Normalize`, `Denormalize` and `Rescale` register their constants (`mean`, `std`, `factor`) as
+  non-persistent buffers instead of plain attributes, so `.to(device)` moves them with the module.
+  Previously they stayed on the CPU: eager tolerates the mix, but `torch.export` traces with fake
+  tensors and refused it, so exporting a preprocessing pipeline from an accelerator failed while
+  the same pipeline exported fine from the CPU. `Denormalize` now coerces a scalar `mean`/`std` to
+  a 1-D tensor, as `Normalize` already did, which changes its `__repr__` to match `Normalize`'s and
+  lets a scalar `Denormalize` reach the ONNX export branch instead of raising `IndexError` there.
+  The buffers are non-persistent, so `state_dict()` is unchanged and existing checkpoints still
+  load. (#4323, #4330)
 
 * `RenderingDeFMO` (used by `DeFMO`) no longer crashes on a half-precision forward pass. Its rendering
   time-steps (`times`) were a plain Python attribute, not a registered buffer, so `nn.Module.to()` never
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   tests — a physical M1 on macOS 26 compiles them fine. kornia still supports torch 2.5.1, so the
   in-tree MPS workarounds stay. (#4202)
 
+### Fixed
+
+* `Normalize`, `Denormalize` and `Rescale` register their constants (`mean`, `std`, `factor`) as
+  non-persistent buffers instead of plain attributes, so `.to(device)` moves them with the module.
+  Previously they stayed on the CPU: eager tolerates the mix, but `torch.export` traces with fake
+  tensors and refused it, so exporting a preprocessing pipeline from an accelerator failed while
+  the same pipeline exported fine from the CPU. `Denormalize` now coerces a scalar `mean`/`std` to
+  a tensor, as `Normalize` already did, which changes its `__repr__` to match `Normalize`'s. The
+  buffers are non-persistent, so `state_dict()` is unchanged and existing checkpoints still load.
+  (#4323)
+
 ### Breaking changes
 
 * `kornia_rs>=0.1.14` is required; the floor used to be 0.1.9. kornia_rs 0.1.11 relocated its image I/O
--- a/kornia/enhance/normalize.py
+++ b/kornia/enhance/normalize.py
@@ -77,8 +77,17 @@ class Normalize(nn.Module):
         if isinstance(std, (tuple, list)):
             std = torch.tensor(std)[None]
 
-        self.mean = mean
-        self.std = std
+        # Buffers, not plain attributes: `.to(device)` has to move them, or a
+        # module living on an accelerator keeps CPU constants. Eager tolerates
+        # the mix (a broadcastable CPU tensor combines with a CUDA/MPS one),
+        # which is why this went unnoticed, but `torch.export` traces with fake
+        # tensors and refuses it.
+        #
+        # persistent=False: these are constructor arguments, not learned state.
+        # Putting them in `state_dict()` would make every existing checkpoint
+        # report unexpected keys, for values the constructor already supplies.
+        self.register_buffer("mean", mean, persistent=False)
+        self.register_buffer("std", std, persistent=False)
 
     def forward(self, input: torch.Tensor) -> torch.Tensor:
         """Normalize an input tensor channel-wise with this module's statistics.
@@ -209,8 +218,18 @@ class Denormalize(nn.Module):
     def __init__(self, mean: Union[torch.Tensor, float], std: Union[torch.Tensor, float]) -> None:
         super().__init__()
 
-        self.mean = mean
-        self.std = std
+        # A float has to become a tensor before it can be a buffer. This is
+        # also what `Normalize` already does with the same argument, so the two
+        # classes now agree on what they store.
+        if not isinstance(mean, torch.Tensor):
+            mean = torch.tensor(mean)
+        if not isinstance(std, torch.Tensor):
+            std = torch.tensor(std)
+
+        # See Normalize: buffers so `.to(device)` moves them; non-persistent so
+        # they stay out of `state_dict()`.
+        self.register_buffer("mean", mean, persistent=False)
+        self.register_buffer("std", std, persistent=False)
 
     def forward(self, input: torch.Tensor) -> torch.Tensor:
         """Restore scale/offset from a tensor normalized by mean and std.
--- a/kornia/enhance/normalize.py
+++ b/kornia/enhance/normalize.py
@@ -218,12 +218,18 @@ class Denormalize(nn.Module):
     def __init__(self, mean: Union[torch.Tensor, float], std: Union[torch.Tensor, float]) -> None:
         super().__init__()
 
-        # A float has to become a tensor before it can be a buffer. This is
-        # also what `Normalize` already does with the same argument, so the two
-        # classes now agree on what they store.
-        if not isinstance(mean, torch.Tensor):
+        # A float has to become a tensor before it can be a buffer. Wrap a
+        # scalar in a list so it becomes 1-D, exactly as `Normalize` does: the
+        # ONNX export branch indexes `mean.shape[0]`, which a 0-d tensor would
+        # turn into `IndexError: tuple index out of range`.
+        if isinstance(mean, (int, float)):
+            mean = torch.tensor([mean])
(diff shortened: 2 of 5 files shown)


dchaudhari7177:
The red `tests-mps` leg is `tests/enhance/test_jpeg.py::TestDiffJPEG::test_keyword_argument[mps-float32]`, which this PR cannot reach.

- The diff touches only `kornia/enhance/normalize.py`, `kornia/enhance/rescale.py`, their test file and the changelog.
- `kornia/enhance/jpeg.py` has **zero** references to `Normalize`, `Denormalize` or `Rescale`, and imports neither module.
- The test compares `jpeg_codec_differentiable(input=img, jpeg_quality=q)` against `jpeg_codec_differentiable(img, q)` — the same tensors through the same function twice — so "Tensor-likes are not close" means the MPS kernel was non-deterministic between the two calls, not that an input changed.

I do not have rerun rights on the repo, so I could not confirm by re-running the job. If a maintainer re-runs it I expect it to go green; happy to push an empty commit to retrigger if that is easier.

`tests/enhance` is 367 passed / 15 skipped locally, and the other 38 checks on this PR are green.

ducha-aiki:
Reviewed at `6007be24` by execution — **CHANGES REQUESTED**, all three asks small; the fix itself is right and the issue's repro confirms it.

**Verified.** The 7 new tests fail on the merge-base (`7 failed, 54 passed`) and pass at the head (`61 passed, 7 skipped`, `tests/enhance/test_normalize.py` + `test_rescale.py`). More usefully, since you had no accelerator: #4323's own repro on this Mac (torch 2.14.0, MPS) — `SegmentationModelsBuilder.get_preprocessing_pipeline(params).to("mps")` then `torch.onnx.export(..., dynamo=True, opset_version=18)` — raises `TorchExportError` on `main` and exports at this head. (`torch.export.export` alone tolerates the CPU constants on 2.14/MPS; it is the ONNX dynamo path in the issue that refuses them.) You were also right about the red `tests-mps` leg: it is `TestDiffJPEG::test_keyword_argument[mps-float32]`, the `jpeg_quality=0` non-finite flake tracked in #4205, unrelated to this diff.

**Asks:**

1. **`Denormalize` scalar coercion: make it 1-D, not 0-d.** Copilot's point holds, executed: `Denormalize(0.5, 0.1)` now stores `tensor(0.5)` (shape `[]`), and under the ONNX-export flag the export branch's `mean.shape[0]` raises `IndexError: tuple index out of range` where `main` raised the clear `ValueError: Only torch.Tensor is accepted when converting to ONNX.` #4323 already names this 0-d `IndexError` as a second defect on the same lines. `Normalize` coerces a scalar with `torch.tensor([mean])` (1-D); do the same here — then the two classes really do agree, and export of a scalar `Denormalize` works instead of failing worse.
2. **Drop the now-false skip in `tests/models/test_segmentation_models.py::test_preprocessing_onnx_export`.** Its `if device.type != "cpu": pytest.skip(...)` block says "`Normalize` keeps mean/std as plain tensor attributes that `.to(device)` does not move" — this PR makes that sentence false, and #4323 asks for exactly that skip to go. I removed the four lines at your head and ran the test with `--device=mps --dtype=float32`: **1 passed**. That is the accelerator export pin the issue asks for, so removing the skip gives it to you for free.
3. **CHANGELOG heading.** The entry is under a new `### Fixed` heading; the file uses `Added` / `Breaking changes` / `Bug fixes` (`AGENTS.md`). Move it under `### Bug fixes`.

Not asked: `persistent=False` is the right call for the reasons you give, and `test_to_moves_the_constants` asserting by attribute name rather than iterating `named_buffers()` is exactly the right shape (the iterating version is vacuous on `main`).

dchaudhari7177:
All three done, and I could execute this round — a CPU torch 2.5.0 turned up on this machine, so these are run results rather than reasoning.

**1. `Denormalize` scalar coercion is now 1-D.** Your reading is right and I reproduced the exact failure. `denormalize()`'s export branch only reshapes when `mean.dim() == 1`, so a 0-d constant slips past it and dies on `mean.shape[0]`:

```
0-d (before this fix): IndexError: tuple index out of range
1-D (after this fix):  OK -> (1, 1, 4, 4)
```

(`denormalize(x, mean, std)` under a patched `torch.onnx.is_in_onnx_export`.) `Denormalize` now uses `torch.tensor([mean])` for an `int`/`float`, matching `Normalize` exactly, and falls back to `torch.tensor(mean)` for anything else non-tensor so a tuple/list still lands 1-D:

```python
Denormalize(0.5, 0.1).mean  ->  tensor([0.5000])  torch.Size([1])
Normalize(0.5, 0.1).mean    ->  tensor([0.5000])  torch.Size([1])
```

So the two classes agree on shape, not just on type — and export of a scalar `Denormalize` now works instead of failing worse than `main`.

**2. Stale skip removed.** The four lines in `test_preprocessing_onnx_export` are gone. Their stated reason ("`Normalize` keeps mean/std as plain tensor attributes that `.to(device)` does not move") is exactly what this branch falsifies, and #4323 asks for the skip to go. Thanks for running it on MPS — I have no accelerator, so that leg is yours and CI's.

**3. CHANGELOG** moved under the existing `### Bug fixes`; the `### Fixed` heading is gone. I also extended the entry to say the scalar coercion is to a *1-D* tensor and that it unblocks the export branch, since that is now part of the fix rather than a side effect.

**Ran locally** (CPU, float32): `tests/enhance/test_normalize.py` + `test_rescale.py` → **61 passed, 7 skipped**, matching your numbers; `tests/models/test_segmentation_models.py` → **18 passed, 1 skipped** (the ONNX test is the skip — no `onnx` here). `ruff check` / `ruff format` clean at v0.16.4. AGENTS.md's `grep -rnE "#4323|issues/4323" kornia/ tests/` returns nothing.

Agreed on `persistent=False` and on the by-attribute-name assertion.

ducha-aiki:
Delta re-review at `0898cb87` — **APPROVED**.

All three asks landed and were re-executed here (torch 2.14.0, CPU + MPS):

- A scalar `Denormalize(0.5, 0.25)` now stores `mean`/`std` as `(1,)` buffers, moves to MPS, and the legacy `torch.onnx.export(dynamo=False)` path succeeds where `main` raised `ValueError` (and where the 0-d version raised Copilot's `IndexError`).
- The skip on `test_segmentation_models.py::test_preprocessing_onnx_export` is gone; 1 passed on CPU and on `--device=mps`. One correction to my earlier note: that test is not itself a regression pin — it passes on base MPS too because `disable_features=True` sidesteps the mismatch. The `tests/enhance` pins are the discriminating ones (7 failed on base, pass at head).
- CHANGELOG entry sits under `### Bug fixes`, no lines of `main`'s deleted.

Also checked: the #4323 repro verbatim fails on base (`FakeTensorDeviceMismatchError`) and exports at head; the buffers are non-persistent, so `state_dict()` is empty on both sides and `load_state_dict(strict=True)` works base→head and head→base; `torch.jit.script` behaves identically to base; 214 passed / 28 skipped over the four CPU dtypes, 56 passed on MPS, dynamo 2 passed.

One mechanical step: GitHub reports the branch `CONFLICTING`, so no test workflow ran at this head. `git merge-tree --write-tree origin/main 0898cb87` is clean — it is the CHANGELOG-adjacency trap — but a dirty PR runs no CI. Please rebase onto current `main` and push; once the checks are green this is merge-ready.

ducha-aiki:
Rebased this branch onto current `main` (`072aff2c`) for you, `0898cb87` -> `a6aab3d0`, at the maintainer's request, so CI can run: GitHub had marked it `CONFLICTING` from CHANGELOG adjacency although a real merge was clean.

Checked before pushing: the rebase applied without conflicts; the code+test patch-id excluding `CHANGELOG.md` is identical before and after; the CHANGELOG diff against `main` adds exactly the same lines as before and deletes nothing; `pre-commit --all-files` passes; the enhance normalize tests and the segmentation ONNX export test, 125 passed, 14 skipped (float32/float64, torch 2.14.0). Pushed with an exact `--force-with-lease` against `0898cb87`. If you have local work on this branch, `git fetch` and reset to `a6aab3d0` before continuing. Nothing else changed; the approval stands and this is merge-ready once the checks are green.

ducha-aiki (dismissed):

ducha-aiki (commented):
Delta re-review at `3fe2b943` — the implementation and regression coverage remain approved; one release-note reference is still missing.

The current head preserves the reviewed buffer changes, scalar `Denormalize` coercion, and removal of the stale accelerator skip. Focused validation passes: 112 passed / 14 skipped in `test_normalize.py` + `test_rescale.py`, and the segmentation preprocessing ONNX export test passes. All 40 reported CI checks are green, and GitHub reports the branch mergeable.

Please cite this PR in the CHANGELOG entry. The new entry still ends only with the issue reference `(#4323)`; repository policy requires the PR number for user-visible changes. Please make it `Fixes #4323. (#4330)` (or otherwise include `#4330`). No code changes are requested.

ducha-aiki (approved):
Delta re-review at `215ffa800085635b28c7b0273738ea71689ca692` — **APPROVED**. The prior release-note request is resolved and the new scalar/export changes are correct.

The CHANGELOG entry now cites both `#4323` and `#4330` under the existing `Bug fixes` heading. `Denormalize` wraps Python scalar constants as one-dimensional tensors, matching `Normalize` and avoiding the ONNX branch's `mean.shape[0]` failure. Removing the accelerator skip is appropriate now that the preprocessing constants are registered buffers.

Local validation at this exact head passed: 61 passed / 7 skipped across `tests/enhance/test_normalize.py` and `tests/enhance/test_rescale.py`, plus the segmentation preprocessing ONNX export test passed. GitHub reports the branch mergeable; 31 checks are successful and 6 are still running or queued, so merge should wait for those remaining checks to complete successfully.

ducha-aiki (approved):

PR #4331, opened 2026-09-06
fix(models): make ModelBase.save actually write its visualizations

### LINKED ISSUE #4322: ModelBase.save writes float visualizations to .png, which write_image rejects: every container's save raises

dchaudhari7177:
Fixes #4322.

`save`/`_save_outputs` handed `visualize`'s float output to `write_image` under a hard-coded `.png` name. PNG is `uint8`/`uint16` only — float32 is TIFF-only, other float dtypes are rejected outright — so every container's `save` (`SemanticSegmentation`, `ObjectDetector`, `EdgeDetector`, `DepthEstimation`) raised on its first write and left the directory it had just created empty.

```
NotImplementedError: Unsupported file extension: .png for float32 image
```

## There are two bugs on this path, not one

Fixing the dtype uncovers the next one immediately. The containers document their inputs as `(B, 3, H, W)` and passed the batch straight through, but `write_image` takes `(3, H, W)`, `(1, H, W)` or `(H, W)`. The backend rejects the rank with a message that names neither the rank nor the shape:

```
TypeError: argument 'image': 'ndarray' object is not an instance of 'ndarray'
```

So both are fixed here. The dtype fix alone still leaves `save` broken for the shape the containers say they accept, which I don't think is worth shipping as a "fix".

- **Float → `uint8`**, the same conversion `ImageModule` already does before handing a tensor to PIL. Clamped to `[0, 1]` first: a visualization that overshoots would otherwise wrap and put black where it should be white. Integer images pass through untouched.
- **A 4-D output is written as one file per batch item.** A batch of one still gets an index suffix, so a caller does not have to guess between `name.png` and `name_0.png`.

After the fix, a 2-image batch through `SemanticSegmentation.save` produces the six files it should — `_src`, `_mask` and `_overlay` for each item — and they load back as `(3, 8, 8) uint8`.

## Why this shipped, and what the tests do about it

**Every existing test in `test_model_base.py` patches `write_image`.** A mock accepts any dtype and any rank, so nothing ever checked that what the containers hand it is something that can actually be written.

The ten new tests call the real `write_image`. **Nine fail on `main`:**

```
FAILED test_a_float_visualization_is_written[dtype0]  - Unsupported file extension: .png for float32 image
FAILED test_a_float_visualization_is_written[dtype1]  - Unsupported image dtype: torch.float64
FAILED test_a_float_visualization_is_written[dtype2]  - Unsupported image dtype: torch.float16
FAILED test_a_batch_is_written_as_one_file_per_item
FAILED test_a_batch_of_one_is_still_indexed
FAILED test_the_written_file_is_a_readable_image
FAILED test_values_survive_the_conversion
FAILED test_out_of_range_values_are_clamped_not_wrapped
FAILED test_save_outputs_writes_real_files_too
```

Two go past "a file exists" and read the PNG back:

- `test_values_survive_the_conversion` pins `0.0 / 1.0 / 0.5 / 0.25` → `0 / 255 / 128 / 64`, so a conversion that silently rescaled would fail.
- `test_out_of_range_values_are_clamped_not_wrapped` pins `1.5` and `-0.5` → `255` and `0`. This is the one that fails if someone drops the clamp as redundant.

```
pytest tests/models/test_model_base.py    21 passed
ruff 0.16.4 check / format                clean
```

One judgement call: I converted to `uint8` rather than switching the extension to `.tiff` (which `write_image` does support for float32). These are pictures for a human to look at — the method is `visualize` — so a viewable PNG seemed more useful than a lossless TIFF. Easy to flip if you disagree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   tests — a physical M1 on macOS 26 compiles them fine. kornia still supports torch 2.5.1, so the
   in-tree MPS workarounds stay. (#4202)
 
+### Fixed
+
+* `ModelBase.save` and `_save_outputs` write the visualizations they are given instead of raising.
+  They handed `visualize`'s float output to `write_image` under a hard-coded `.png` name, but PNG is
+  `uint8`/`uint16` only, so every container's `save` (`SemanticSegmentation`, `ObjectDetector`,
+  `EdgeDetector`, `DepthEstimation`) raised on its first write and left an empty directory behind.
+  Float images are now converted to `uint8`, clamped to `[0, 1]` first so an out-of-range
+  visualization does not wrap. A batched `(B, 3, H, W)` output -- the shape the containers document
+  -- is written as one file per item rather than passed whole to `write_image`, which takes
+  `(3, H, W)` and rejected the rank with a message naming neither. (#4322)
+
 ### Breaking changes
 
 * `kornia_rs>=0.1.14` is required; the floor used to be 0.1.9. kornia_rs 0.1.11 relocated its image I/O
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,17 +130,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   tests — a physical M1 on macOS 26 compiles them fine. kornia still supports torch 2.5.1, so the
   in-tree MPS workarounds stay. (#4202)
 
-### Fixed
-
-* `ModelBase.save` and `_save_outputs` write the visualizations they are given instead of raising.
-  They handed `visualize`'s float output to `write_image` under a hard-coded `.png` name, but PNG is
-  `uint8`/`uint16` only, so every container's `save` (`SemanticSegmentation`, `ObjectDetector`,
-  `EdgeDetector`, `DepthEstimation`) raised on its first write and left an empty directory behind.
-  Float images are now converted to `uint8`, clamped to `[0, 1]` first so an out-of-range
-  visualization does not wrap. A batched `(B, 3, H, W)` output -- the shape the containers document
-  -- is written as one file per item rather than passed whole to `write_image`, which takes
-  `(3, H, W)` and rejected the rank with a message naming neither. (#4322)
-
 ### Breaking changes
 
 * `kornia_rs>=0.1.14` is required; the floor used to be 0.1.9. kornia_rs 0.1.11 relocated its image I/O
@@ -381,6 +370,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the quality with `torch.randint(low=0, high=100)` unseeded at sixteen sites, so roughly one run in twenty of
   `tests/enhance/test_jpeg.py` went red on any job; the draws now start at `1` and the boundary has pinned
   cases of its own. (#4205)
+* `ModelBase.save` and `_save_outputs` write the visualizations they are given instead of raising.
+  They handed `visualize`'s float output to `write_image` under a hard-coded `.png` name, but PNG is
+  `uint8`/`uint16` only, so every container's `save` (`SemanticSegmentation`, `ObjectDetector`,
+  `EdgeDetector`, `DepthEstimation`) raised on its first write and left an empty directory behind.
+  Float images are now converted to `uint8`, clamped to `[0, 1]` first so an out-of-range
+  visualization does not wrap. A batched `(B, 3, H, W)` output -- the shape the containers document
+  -- is written as one file per item rather than passed whole to `write_image`, which takes
+  `(3, H, W)` and rejected the rank with a message naming neither. (#4322)
 
 * `kornia.io.load_image` and `write_image` work on the kornia_rs that a plain `pip install kornia`
   resolves. kornia_rs 0.1.11 moved its image readers and writers from the package root into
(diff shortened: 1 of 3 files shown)


ducha-aiki:
Reviewed at `be97f860` by execution — **APPROVE on one CHANGELOG edit**.

**Verified.** The 9 new tests fail on the merge-base (`9 failed, 12 passed`) and pass at the head (`21 passed`, `tests/models/test_model_base.py`), against the real `write_image`, which is the point: every pre-existing test in that file mocks it. The two read-back pins (`0/1/0.5/0.25 → 0/255/128/64`, and `1.5/-0.5 → 255/0`) are the kind that survive a later "the clamp is redundant" edit. `kornia/` carries no `#4322` tag; the test file names it, as a pin should. The uint8-over-TIFF judgement call is the right one for a method named `visualize`.

**Ask:** the CHANGELOG entry sits under a new `### Fixed` heading; the file's existing headings are `Added` / `Breaking changes` / `Bug fixes` (`AGENTS.md`). Move it under `### Bug fixes`.

**Not asked (Copilot's point, checked):** `_to_writable_png` passes `int64` / `int32` / `bool` through, and `write_image` then raises `NotImplementedError: Unsupported image dtype` — the same outcome as before this PR, so nothing regresses, and every container's `visualize` returns float. If you want the docstring exact, "Integer images pass through untouched" could say "`uint8`/`uint16` images pass through untouched"; optional.

dchaudhari7177:
Both done.

**CHANGELOG** moved under the existing `### Bug fixes`; the new `### Fixed` heading is gone.

**Docstring:** took the optional edit, and made it a bit wider than your suggestion because "Integer images pass through untouched" was wrong in two directions — `bool` passes through too, and of the ones that pass, only `uint8`/`uint16` are ones `write_image` can actually serialize. It now reads:

> Non-float images pass through untouched, so a `uint8` or `uint16` visualization is written as-is.

That matches the code (`if not image.is_floating_point(): return image`) without implying the pass-through is a guarantee of writability. As you checked, `int64`/`int32`/`bool` still reach the same `NotImplementedError` they reached before this PR, so nothing regresses.

`ruff check` / `ruff format` clean at v0.16.4.

ducha-aiki:
Delta re-review at `88ef8d05` — **APPROVED**.

Both edits landed: the CHANGELOG entry moved under `### Bug fixes` (nothing of `main`'s deleted), and the docstring now says exactly what the code does ("Non-float images pass through untouched, so a `uint8` or `uint16` visualization is written as-is"; `int64`/`bool` still raise `NotImplementedError`, as on base). Re-run here: 21 passed at head, the 9 real-`write_image` pins still fail on base, `tests/models` non-slow 329 passed / 92 skipped.

Same mechanical step as #4330: GitHub marks the branch `CONFLICTING` (merge-tree is clean, it is the CHANGELOG adjacency), so no CI ran at this head. Rebase onto current `main`, push, and it is merge-ready once green.

ducha-aiki:
Rebased this branch onto current `main` (`072aff2c`) for you, `88ef8d05` -> `d8f1df62`, at the maintainer's request, so CI can run: GitHub had marked it `CONFLICTING` from CHANGELOG adjacency although a real merge was clean.

Checked before pushing: the rebase applied without conflicts; the code+test patch-id excluding `CHANGELOG.md` is identical before and after; the CHANGELOG diff against `main` adds exactly the same lines as before and deletes nothing; `pre-commit --all-files` passes; tests/models save tests, 21 passed (float32/float64, torch 2.14.0). Pushed with an exact `--force-with-lease` against `88ef8d05`. If you have local work on this branch, `git fetch` and reset to `d8f1df62` before continuing. Nothing else changed; the approval stands and this is merge-ready once the checks are green.

ducha-aiki (approved):

PR #4332, opened 2026-09-06
fix(core): let download_file_from_url validate and re-fetch a cache entry

### LINKED ISSUE #4309: download_file_from_url: a truncated-but-2xx transfer poisons the cache entry permanently
## What happens

`kornia.core.download.download_file_from_url` (added in #4293, used by `download_hf_file` for the KimiVL and SigLIP2 safetensors checkpoints) returns the cached path on a cache hit and never re-validates the file. The transfer underneath is `torch.hub.download_url_to_file`, which moves the temp file into place atomically but does not check the body against `Content-Length`. A connection that drops after a 2xx status therefore leaves a truncated file in the cache; every later call returns it as a hit, `load_safetensors` raises `ValueError` naming the path (the header or `data_offsets` no longer fit the file), and `from_pretrained_hf()` fails until the user deletes the file by hand.

`load_state_dict_from_url` does not have this problem: its quarantine machinery (`_discard_cache_entry` / `_settle_quarantine`) is driven by *load* failures, so a corrupt `.pth` gets re-fetched once. A download-only function has no load step to hook into, which is why #4293 left `download_file_from_url` without quarantine and documented it in the docstring.

## Fix

Cheapest: have the two safetensors callers catch `ValueError` from `load_safetensors`, delete the cached path, and retry the download once (mirroring the `.pth` path's one re-attempt). Better: give `download_file_from_url` an optional `validate: Callable[[str], None]` that the caller passes (`load_safetensors` header parse), so the quarantine lives with the download like it does for state dicts. Either way, one test with a `file://` source that serves a truncated body first and the full body second.

Found during the final review of the #4259/#4261 series.

dchaudhari7177:
Fixes #4309.

`torch.hub.download_url_to_file` moves its temp file into place atomically but does not check the body against `Content-Length`, so a connection that drops after a 2xx leaves a truncated file in the cache. Every later call returned it as a hit, `load_safetensors` raised naming the path, and `from_pretrained_hf()` failed until the file was deleted by hand.

Measured end to end against a poisoned cache entry:

```
without validate:  returned 90 bytes   (still poisoned, forever)
with validate:     returned 102 bytes  (re-fetched, loads)
```

## The approach

I took the issue's "better" option — an optional `validate: Callable[[str], None]` — rather than catching `ValueError` in the two safetensors callers. The reason is the one #4293 already documented: `load_state_dict_from_url` hangs its quarantine off the *load* step, and a download-only function has no load step. `validate` **is** that missing step, so the quarantine can live with the download instead of being reimplemented per caller.

Raising from `validate` is then treated exactly as a load failure is, reusing the existing machinery unchanged: `_discard_cache_entry` renames the entry aside, later sources are tried, the discarded source is re-fetched once, and `_settle_quarantine` decides on the *outcome of the call*.

**Renaming rather than deleting matters as much here as it does there.** A validator that is wrong about a multi-gigabyte checkpoint must not destroy it — there's a test asserting that a file which never validates raises `RuntimeError` and leaves the *original* cache entry in place.

**Without `validate` the behaviour is exactly as before.** There's a test pinning that too, so it stays a choice rather than becoming an accident.

## `check_safetensors`

The validator the builders pass. Rather than write a second header parser, I factored the header half of `load_safetensors` into `_read_and_check_header` and shared it — the length prefix, the JSON header, every entry's dtype and byte range, and that those ranges cover the buffer exactly once. So the two cannot drift.

Header-only, so it stays cheap on a large checkpoint. That matters because `validate` runs on **cache hits** too, not just after a transfer.

## Tests

Six new tests. **Five fail on `main`:**

```
FAILED test_a_poisoned_cache_entry_is_refetched                      - unexpected kwarg 'validate'
FAILED test_a_good_cache_entry_is_not_refetched                      - unexpected kwarg 'validate'
FAILED test_a_source_that_serves_a_bad_file_falls_through_to_the_next - unexpected kwarg 'validate'
FAILED test_a_file_that_never_validates_raises_and_keeps_the_original - unexpected kwarg 'validate'
FAILED test_check_safetensors_rejects_a_truncated_checkpoint          - cannot import check_safetensors
```

The sixth — `test_without_validate_the_poisoned_entry_is_returned` — passes on both, deliberately: it pins the opt-out.

```
pytest tests/core            408 passed, 13 skipped
ruff 0.16.4 check / format   clean
```

One thing worth a maintainer's opinion: `validate` is opt-in, so `download_file_from_url` called without it still has the old failure mode. I left it that way because the function is public and a mandatory validator would be a breaking change — but if you'd rather `download_hf_file` defaulted to `check_safetensors` for `.safetensors` files, that's a one-line change and I'm happy to make it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,16 +135,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   tests — a physical M1 on macOS 26 compiles them fine. kornia still supports torch 2.5.1, so the
   in-tree MPS workarounds stay. (#4202)
 
-### Fixed
-
-* `download_file_from_url` and `download_hf_file` take an optional `validate=` callable, and
-  `kornia.core.check_safetensors` is the one the KimiVL and SigLIP2 builders pass. A transfer cut
-  short after a 2xx status leaves a truncated file in the cache, which was then returned as a cache
-  hit on every later call -- `from_pretrained_hf()` failed until the user deleted it by hand.
-  `validate` gives a download-only call the quarantine `load_state_dict_from_url` gets from its load
-  step: a rejected entry is moved aside, the next source is tried, and the discarded source is
-  re-fetched once. Without `validate` the behaviour is unchanged. (#4309)
-
 ### Breaking changes
 
 * `kornia_rs>=0.1.14` is required; the floor used to be 0.1.9. kornia_rs 0.1.11 relocated its image I/O
@@ -368,6 +358,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed `unproject_points_z1` depth shape handling for singleton and multi-axis batches,
   accepting both trailing-singleton and flat depth tensors. (#4355)
+
+* `download_file_from_url` and `download_hf_file` take an optional `validate=` callable, and
+  `kornia.core.check_safetensors` is the one the KimiVL and SigLIP2 builders pass. A transfer cut
+  short after a 2xx status leaves a truncated file in the cache, which was then returned as a cache
+  hit on every later call -- `from_pretrained_hf()` failed until the user deleted it by hand.
+  `validate` gives a download-only call the quarantine `load_state_dict_from_url` gets from its load
+  step: a rejected entry is moved aside, the next source is tried, and the discarded source is
+  re-fetched once. Without `validate` the behaviour is unchanged. (#4309, #4332)
 
 * `RenderingDeFMO` (used by `DeFMO`) no longer crashes on a half-precision forward pass. Its rendering
   time-steps (`times`) were a plain Python attribute, not a registered buffer, so `nn.Module.to()` never
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   tests — a physical M1 on macOS 26 compiles them fine. kornia still supports torch 2.5.1, so the
   in-tree MPS workarounds stay. (#4202)
 
+### Fixed
+
+* `download_file_from_url` and `download_hf_file` take an optional `validate=` callable, and
+  `kornia.core.check_safetensors` is the one the KimiVL and SigLIP2 builders pass. A transfer cut
+  short after a 2xx status leaves a truncated file in the cache, which was then returned as a cache
+  hit on every later call -- `from_pretrained_hf()` failed until the user deleted it by hand.
+  `validate` gives a download-only call the quarantine `load_state_dict_from_url` gets from its load
+  step: a rejected entry is moved aside, the next source is tried, and the discarded source is
+  re-fetched once. Without `validate` the behaviour is unchanged. (#4309)
+
 ### Breaking changes
 
 * `kornia_rs>=0.1.14` is required; the floor used to be 0.1.9. kornia_rs 0.1.11 relocated its image I/O
--- a/docs/source/core.rst
+++ b/docs/source/core.rst
@@ -19,7 +19,8 @@ Helpers for fetching and reading pretrained weights. :func:`load_state_dict_from
 is the one models load a ``torch.load``-able checkpoint with;
 :func:`download_hf_file` (or :func:`download_file_from_url`) and
 :func:`load_safetensors` are the two halves of the same job for a
-``.safetensors`` checkpoint.
+``.safetensors`` checkpoint; :func:`check_safetensors` is the ``validate=``
+callable that lets the download half reject a truncated cache entry.
 
 .. autofunction:: hf_url
 
@@ -30,3 +31,5 @@ is the one models load a ``torch.load``-able checkpoint with;
 .. autofunction:: download_hf_file
 
 .. autofunction:: load_safetensors
+
+.. autofunction:: check_safetensors
--- a/kornia/core/__init__.py
+++ b/kornia/core/__init__.py
@@ -30,7 +30,7 @@ from .mixin import (
 )
 from .module import ImageModule, ImageSequential
 from .ops import eye_like, vec_like
-from .safetensors import load_safetensors
+from .safetensors import check_safetensors, load_safetensors
 from .tensor_wrapper import TensorWrapper
 
 __all__ = [
@@ -41,6 +41,7 @@ __all__ = [
     "ONNXMixin",
     "ONNXRuntimeMixin",
     "TensorWrapper",
+    "check_safetensors",
     "download_file_from_url",
     "download_hf_file",
     "external",
--- a/kornia/core/download.py
+++ b/kornia/core/download.py
@@ -23,6 +23,7 @@ import os
 import sys
 import time
 import warnings
+from collections.abc import Callable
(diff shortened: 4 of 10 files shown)


ducha-aiki:
Reviewed at `875758cf` by execution — **CHANGES REQUESTED**: the PR breaks two existing tests, which is why 29 CI legs are red.

**What is red.** Every `tests (...)` leg fails on the same two nodes:

```
FAILED tests/models/test_siglip2.py::TestSigLip2Builder::test_download_weights_uses_kornias_downloader - AssertionError: expected call not found.
FAILED tests/models/test_kimi_vl.py::TestKimiVLBuilder::test_download_weights_uses_kornias_downloader - AssertionError: expected call not found.
```

Reproduced locally at your head (`2 failed` with `-k download_weights`). Both tests `assert_called_once_with(...)` on `download_hf_file` with the old argument list, and this PR adds `validate=check_safetensors` to the call. The body's `pytest tests/core  408 passed` did not cover them; the builders are under `tests/models`. Update the two expectations to include `validate=check_safetensors` — that is the correct new contract, and it pins that the builders actually pass the validator, which nothing else does.

**Verified, and good.** The 5 new tests in `tests/core/test_download.py` fail on the merge-base (`5 failed, 86 passed`) and pass at the head (`91 passed`); the opt-out pin passes on both, as intended. Reusing `_discard_cache_entry` / `_settle_quarantine` rather than a second quarantine, renaming aside rather than deleting, and factoring the header parse into `_read_and_check_header` so `check_safetensors` cannot drift from `load_safetensors` are all the right calls. `kornia/` carries no `#4309` tag.

**Two more asks, small:**

1. `check_safetensors` is a new public name in `kornia.core.__all__`; add `.. autofunction:: check_safetensors` next to `load_safetensors` in `docs/source/core.rst` (`AGENTS.md`: every new public function goes on its rst page).
2. The CHANGELOG entry is under a new `### Fixed` heading; the file uses `Added` / `Breaking changes` / `Bug fixes`. Move it under `### Bug fixes`.

**On the open question** (should `download_hf_file` default to `check_safetensors` for `.safetensors` names): keep it opt-in as you have it. The two kornia callers pass it explicitly, the function is public, and a validator that runs on every cache hit is something a caller should choose. If a maintainer wants the default, it is the one-line change you describe and can come later.

dchaudhari7177:
Thanks — all three addressed at the new head.

**The two red tests.** You're right that `pytest tests/core` never reached them; the builders live under `tests/models`. Both `assert_called_once_with` expectations now carry `validate=<module>.check_safetensors`, referenced through the builder module so the assertion pins the exact object the builder passes rather than a re-import that could drift:

```python
download.assert_called_once_with(
    "google/siglip2-base-patch16-224",
    "model.safetensors",
    model_dir=str(tmp_path),
    validate=siglip2_builder.check_safetensors,
)
```

Same shape in `tests/models/test_kimi_vl.py`. That is the new contract, and as you note it is now the only place asserting the builders actually wire the validator in.

**Docs.** `.. autofunction:: check_safetensors` added to `docs/source/core.rst` after `load_safetensors`, and the section's prose now names it: "`check_safetensors` is the `validate=` callable that lets the download half reject a truncated cache entry."

**CHANGELOG.** Entry moved out of the new `### Fixed` heading and under the existing `### Bug fixes`; the `### Fixed` heading is gone.

**On the default:** agreed, keeping it opt-in. Both kornia callers pass it explicitly, and making a validator run on every cache hit is a caller's choice to make.

I don't have an accelerator or a torch install here, so the two updated expectations are CI's to confirm — they are pure-mock assertions with no device path, and `ruff check` / `ruff format` (v0.16.4, the pre-commit pin) are clean.

ducha-aiki:
Delta re-review at `bc4a305d` — **APPROVED**.

All three asks landed: `test_kimi_vl.py:89` and `test_siglip2.py:91` now assert `validate=<builder>.check_safetensors`, which is exactly the wiring the 29 red legs were missing; `check_safetensors` is on `docs/source/core.rst` (prose at line 22 and an `autofunction` at 35); the CHANGELOG entry is under `### Bug fixes` with nothing of `main`'s deleted. Re-run here: 134 passed across `tests/core` and the builder tests at head; the new pins are 7 failed on base for the right reasons. `validate` is keyword-only in both functions, and there is no other caller or signature-sensitive mock in the tree. Copilot's "Failed to download" wording point stands as an optional nit.

Same mechanical step as your other two: GitHub marks the branch `CONFLICTING` (merge-tree is clean — CHANGELOG adjacency), so no CI ran at this head. Rebase onto current `main`, push, and it is merge-ready once green. Any order among #4330, #4331 and #4332 merges cleanly; the three entries just stack under `### Bug fixes`.

ducha-aiki:
Rebased this branch onto current `main` (`072aff2c`) for you, `bc4a305d` -> `d4675d73`, at the maintainer's request, so CI can run: GitHub had marked it `CONFLICTING` from CHANGELOG adjacency although a real merge was clean.

Checked before pushing: the rebase applied without conflicts; the code+test patch-id excluding `CHANGELOG.md` is identical before and after; the CHANGELOG diff against `main` adds exactly the same lines as before and deletes nothing; `pre-commit --all-files` passes; tests/core and the two builder download tests, 169 passed (float32/float64, torch 2.14.0). Pushed with an exact `--force-with-lease` against `bc4a305d`. If you have local work on this branch, `git fetch` and reset to `d4675d73` before continuing. Nothing else changed; the approval stands and this is merge-ready once the checks are green.

ducha-aiki (dismissed):

ducha-aiki (commented):
Delta re-review at current head `2eb90dee` — **APPROVE after one CHANGELOG edit**.

The branch was force-pushed back from the maintainer-rebased `d4675d73`, which dismissed the prior approval, but the code-and-test patch is unchanged: the non-CHANGELOG diff from each head's merge base has the same stable patch ID (`a6679615`). The merge tree against current `main` is clean and GitHub reports `MERGEABLE` with all 41 checks green.

I re-ran `tests/core` plus both builder downloader assertions at float32/float64: 479 passed, 11 skipped. The two mock expectations still pin the exact `validate=check_safetensors` wiring, and the public docs/export remain present.

One repository-contract edit was missed in the earlier round: the CHANGELOG entry at line 365 cites only issue #4309. Add this PR number too: `(#4309, #4332)`. No code changes are otherwise needed.

ducha-aiki (approved):
Delta re-review at `a6dfb1a0d7f9e2743ee4f5855f65988e683793d3` — **APPROVED**. The prior CHANGELOG citation request is resolved, and the reviewed implementation remains sound after the rebase.

The `Bug fixes` entry now cites both issue `#4309` and PR `#4332`. The downloader's one-refetch-per-cache-path/process bound remains intentional: a real transfer spends the budget even when the replacement is rejected, preventing a persistently bad source from triggering a multi-gigabyte download on every builder call. Offline failures that transfer nothing still release the budget, as documented by `_settle_quarantine`.

Focused validation at this exact head passed: 138 downloader/safetensors/builder tests and 146 API-surface tests. GitHub reports the branch mergeable; 33 checks are successful and 5 remain running or queued at this snapshot, so merge should wait for those checks to complete successfully.

#4362 implements the same issue and public download-validation surface and must not also merge. Useful extra validator coverage from that later PR can be considered separately, but #4332 is complete as submitted.

PR #4335, opened 2026-09-06
fix(contrib): make the super_resolution builders constructible again

### LINKED ISSUE #4291: kornia.contrib.super_resolution builders raise TypeError: SuperResolution never implements the abstract from_config (since #3426)

adimalkar:
Fixes #4291

Thanks for the steer on direction, @ducha-aiki — went with (a).

## Problem

Both public builders raised at construction, so the whole `kornia.contrib.super_resolution`
entry point was unreachable:

```
>>> from kornia.contrib.super_resolution import SmallSRBuilder
>>> SmallSRBuilder.build(pretrained=False)
TypeError: Can't instantiate abstract class SuperResolution without an
implementation for abstract method 'from_config'
```

**Two defects were stacked here, not one.** Implementing `from_config` alone does not fix it —
`SuperResolution` also defines no `__init__`, so once the class is concrete the builders'
arguments fall through to `nn.Module.__init__`:

```
TypeError: SuperResolution.__init__() got an unexpected keyword argument 'pre_processor'
```

Both `SmallSRBuilder.build` and `RRDBNetBuilder.build` pass `pre_processor=`, `post_processor=`
and `name=`.

## Changes

- **`SuperResolutionConfig`** — minimal, as requested: the fields are exactly the arguments the
  two builders already take (`model_name`, `pretrained`, `upscale_factor`, `image_size`).
- **`SuperResolution.from_config`** — dispatches to `SmallSRBuilder` for the lightweight variant
  and `RRDBNetBuilder` for the RRDB variants, matching the shape of the eight existing
  `ModelBase` implementations.
- **`SuperResolution.__init__`** — matching `SemanticSegmentation`, which has the identical
  model / pre / post / name wrapper contract.

Keeping `ModelBase` rather than dropping it preserves `save`/`_save_outputs` and the ONNX export
mixin, and keeps super-resolution inside the `ModelBase.save` fix in #4331.

## Before / after

```
on main:      SmallSRBuilder.build -> TypeError: Can't instantiate abstract class...
              RRDBNetBuilder.build -> TypeError: Can't instantiate abstract class...

with the fix: SmallSRBuilder.build -> OK   forward (1, 3, 16, 16) -> (1, 3, 672, 672)
              RRDBNetBuilder.build -> OK   forward (1, 3, 16, 16) -> (1, 3, 64, 64)
```

## Tests

`tests/contrib/test_super_resolution.py`, non-slow, five cases:

- one per builder, constructing with `pretrained=False` and running a `(1, 3, 16, 16)` forward
- `from_config` dispatch for both families, parametrized
- `from_config` rejects an unknown `model_name`

The `RRDBNetBuilder` case means #4261 can drop its monkeypatch. Nothing in `tests/` called either
builder before, which is why CI never caught this.

## Incidental

The fix also clears **7 of the 8** mypy errors in the module — mypy had been reporting
`Cannot instantiate abstract class "SuperResolution"` at both builder call sites on `main`. The
one remaining (`Signature of "save" incompatible with supertype "ModelBaseMixin"`) predates this
change and is left alone.

CHANGELOG entry added under *Unreleased → Bug fixes*.

## Note on local verification

I do not have `pixi` installed, so I ran `ruff check`, `ruff format --check` and `mypy` directly
rather than `pixi run pre-commit-all`. Both ruff checks pass on the changed files. If the hook
set covers anything beyond those, CI will be the real check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -493,7 +493,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `RRDBNetBuilder.build()` both raised `TypeError` at construction — the whole public
   super-resolution entry point had been unreachable since the models refactor made `from_config`
   abstract. It now has a `SuperResolutionConfig` and a `from_config` that dispatches to either
-  builder family, and both builders are covered by tests. (#4291)
+  builder family, and both builders are covered by tests. Fixes #4291. (#4335)
 
 * `kornia.io.load_image` and `write_image` work on the kornia_rs that a plain `pip install kornia`
   resolves. kornia_rs 0.1.11 moved its image readers and writers from the package root into
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -488,6 +488,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   now built in `float32` and `float16`/`bfloat16` results are byte-identical to `float32`/`float64`.
   `RandomErasing` and `RandomCutMixV2` build their masks through it. (#4336)
 
+* `kornia.contrib.super_resolution` builders construct again. `SuperResolution` never implemented
+  `ModelBase`'s abstract `from_config` and defined no `__init__`, so `SmallSRBuilder.build()` and
+  `RRDBNetBuilder.build()` both raised `TypeError` at construction — the whole public
+  super-resolution entry point had been unreachable since the models refactor made `from_config`
+  abstract. It now has a `SuperResolutionConfig` and a `from_config` that dispatches to either
+  builder family, and both builders are covered by tests. (#4291)
+
 * `kornia.io.load_image` and `write_image` work on the kornia_rs that a plain `pip install kornia`
   resolves. kornia_rs 0.1.11 moved its image readers and writers from the package root into
   `kornia_rs.io`, and kornia kept calling the root, so on 0.1.11 and newer `load_image` raised
--- a/docs/export_support/cases_models.py
+++ b/docs/export_support/cases_models.py
@@ -672,7 +672,6 @@ A(
             model=torch.nn.Identity(), pre_processor=torch.nn.Identity(), post_processor=torch.nn.Identity(), name="x"
         ),
         [IMG],
-        note="KORNIA BUG: abstract (ModelBase.from_config not implemented, no __init__) -> TypeError",
     )
 )
 A(
@@ -681,7 +680,6 @@ A(
         "contrib",
         lambda x: K.contrib.SmallSRBuilder.build("small_sr", pretrained=False),
         [IMG],
-        note="KORNIA BUG: builder instantiates abstract SuperResolution -> TypeError at super_resolution.py:262",
     )
 )
 A(
@@ -690,7 +688,7 @@ A(
         "contrib",
         None,
         [],
-        skip="builds a 17M-parameter Real-ESRGAN generator and would hit the same abstract SuperResolution TypeError",
+        skip="builds a 17M-parameter Real-ESRGAN generator; too heavy for the survey",
     )
 )
 
--- a/docs/source/contrib.rst
+++ b/docs/source/contrib.rst
@@ -195,3 +195,18 @@ KMeans
 
 .. autoclass:: KMeans
     :members:
+
+Super Resolution
+----------------
+
+.. autoclass:: SuperResolution
+    :members:
+
+.. autoclass:: SuperResolutionConfig
+    :members:
+
+.. autoclass:: SmallSRBuilder
+    :members:
+
+.. autoclass:: RRDBNetBuilder
+    :members:
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -80,10 +80,13 @@ The `RRDBNet` class is the Residual-in-Residual Dense Block generator behind ESR
 It is a plain ``nn.Module`` that upsamples a batched ``(B, 3, H, W)`` image by a factor of 1, 2 or 4,
 and its module and parameter names match the reference implementation, so the published Real-ESRGAN
 checkpoints load with ``strict=True``. ``kornia.contrib.super_resolution.RRDBNetBuilder`` configures
-it for the released Real-ESRGAN variants and downloads their weights, but it currently raises
-``TypeError`` because the ``SuperResolution`` wrapper it returns never implements the abstract
-``from_config`` method it inherits (`kornia#4291 <https://github.com/kornia/kornia/issues/4291>`_);
-until that is fixed, construct ``RRDBNet`` directly and load the checkpoint with ``load_state_dict``.
+it for the released Real-ESRGAN variants and downloads their weights, and returns a
+``SuperResolution`` wrapper ready for inference::
+
+    from kornia.contrib.super_resolution import RRDBNetBuilder
+
+    model = RRDBNetBuilder.build("RealESRNet_x4plus")
+    upscaled = model(images)  # (B, 3, H, W) -> (B, 3, 4H, 4W)
 
 The architecture is vendored from `BasicSR <https://github.com/XPixelGroup/BasicSR>`_ (Apache-2.0,
 Copyright 2018-2022 BasicSR Authors); no extra package is required to use it.
--- a/kornia/contrib/__init__.py
+++ b/kornia/contrib/__init__.py
@@ -39,7 +39,7 @@ from .image_stitching import ImageStitcher
 from .kmeans import KMeans
 from .lambda_module import Lambda
 from .object_detection import ObjectDetector, RTDETRDetectorBuilder
-from .super_resolution import RRDBNetBuilder, SmallSRBuilder, SuperResolution
+from .super_resolution import RRDBNetBuilder, SmallSRBuilder, SuperResolution, SuperResolutionConfig
 
 __all__ = [
     "CombineTensorPatches",
@@ -55,6 +55,7 @@ __all__ = [
     "RTDETRDetectorBuilder",
     "SmallSRBuilder",
     "SuperResolution",
+    "SuperResolutionConfig",
     "TinyViT",
     "combine_tensor_patches",
     "compute_padding",
--- a/kornia/contrib/super_resolution.py
+++ b/kornia/contrib/super_resolution.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 #
 
+from dataclasses import dataclass
 from typing import Any, List, Optional, Tuple, Union
 
 import torch
@@ -41,6 +42,32 @@ _URLs = {
 
 
 # TODO: support patching -> SR -> unpatching pipeline
+@dataclass
+class SuperResolutionConfig:
+    """Configuration to construct a :class:`SuperResolution` model.
+
+    The fields mirror the arguments accepted by :class:`RRDBNetBuilder` and
+    :class:`SmallSRBuilder`, so a config can drive either builder.
+
+    Args:
+        model_name: Name of the model variant to build. RRDB variants are
+            ``"RealESRGAN_x4plus"``, ``"RealESRNet_x4plus"``,
+            ``"RealESRGAN_x4plus_anime_6B"`` and ``"RealESRGAN_x2plus"``;
+            the lightweight variant is ``"small_sr"``.
+        pretrained: If ``True``, load pretrained weights for the selected variant.
+        upscale_factor: Integer scale used by the lightweight variant. Ignored by
+            the RRDB variants, whose scale is fixed by ``model_name``.
+        image_size: Optional fixed input size for the lightweight variant. When
+            provided, output metadata is configured as ``image_size * upscale_factor``.
+
+    """
+
+    model_name: str = "small_sr"
+    pretrained: bool = True
+    upscale_factor: int = 3
+    image_size: Optional[int] = None
+
+
 class SuperResolution(ModelBase, ONNXExportMixin):
     """SuperResolution is a module that wraps an super resolution model."""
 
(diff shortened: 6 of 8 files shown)


ducha-aiki:
## Review at `1fe9e648` by execution — CHANGES REQUESTED

The construction fix is right, and every before/after claim in the body held verbatim. What blocks it is that the new `__init__` leaves the ONNX path — the thing the body says it preserves — unreachable, plus three loose ends the issue thread and the AGENTS rules call for.

**Verified**

| Claim | Result |
| --- | --- |
| Base: both builders raise the abstract-class `TypeError` | Held. Planting only a `from_config` on base gives the second `TypeError: ... unexpected keyword argument 'pre_processor'` — the "two stacked defects" description is exact. |
| Head: `(1,3,16,16)` → `(1,3,672,672)` small, `(1,3,64,64)` RRDB | Held. The 672 is the fixed `ResizePreProcessor(224, 224)` then ×3, not 16×42 — a 32×32 input also gives 672. |
| `from_config` dispatches both families; unknown name raises | Held; `ValueError` from `RRDBNetBuilder.build`. Config defaults equal `SmallSRBuilder.build`'s exactly; `@dataclass` like `SamConfig`. |
| `pretrained=True` reaches the downloader | Held (stubbed `CachedDownloader.download_to_cache`, right URL, nothing downloaded). |
| New tests | 9 passed float32+float64 (8.2 s); 9 passed float16+bfloat16 (28 s); 5 passed MPS float32. Fail-on-base: the two builder tests fail with the right `TypeError`; the rest cannot import on base. `pre-commit` and `ty` clean. |
| `save` at head | Fails in `write_image` on a float32 PNG — that is #4331's defect, independent of this PR. |

**Asks**

1. **P1 — `SuperResolution.__init__` never initialises `input_image_size` / `output_image_size` / `pseudo_image_size`.** The class body only annotates them (`super_resolution.py:48-50`) and `to_onnx` reads them (`:156-158`). At head `RRDBNetBuilder.build(pretrained=False).to_onnx()` raises `AttributeError: 'SuperResolution' object has no attribute 'input_image_size'`; `SmallSRBuilder` with `image_size=None` sets only `pseudo_image_size`, so it fails the same way. Set the three to `None` in `__init__`. With that three-line change both families export and onnxruntime agrees with eager (max |diff| ≤ 2.8e-6). Pin it: at minimum `assert RRDBNetBuilder.build(pretrained=False).pseudo_image_size is None` (fails at head), better an `importorskip("onnx")` + `to_onnx(save=False)` on the small builder.
2. **P2 — `docs/source/models.rst:82-86`** still says the builder "currently raises `TypeError` ... (kornia#4291); until that is fixed, construct `RRDBNet` directly". False once this merges; replace with the working usage.
3. **P2 — `tests/models/test_rrdbnet.py:322-333`**: drop the monkeypatch stub, as its own comment instructs ("Once #4291 is fixed, drop the stub and assert on the returned `SuperResolution` directly"). Verified: with the stub removed and the assertions moved onto the returned `SuperResolution` (`isinstance`, `.model`, `.name`, `.pre_processor`, `.post_processor`), the 5 Builder tests pass at head. The `super_resolution_module` import goes with it.
4. **P2 — export and document `SuperResolutionConfig`**: it is not in the module's `__all__` nor in `kornia.contrib`'s, and `docs/source/contrib.rst` has no super-resolution section at all. Add one with `autoclass` entries for `SuperResolution`, `SuperResolutionConfig`, `SmallSRBuilder`, `RRDBNetBuilder`, the way `SamConfig`/`EfficientViTConfig` are exported and documented.
5. **P2 — CHANGELOG** cites `(#4291)`, the issue; the file cites the PR number. Make it `... Fixes #4291. (#4335)`.
6. nit — the config docstring repeats the builder's "`image_size * upscale_factor`" formula, but the code does `image_size * 3` (`upscale_factor=2, image_size=32` gives `output_image_size == 96`, and the forward is 448 because of the fixed 224 resize). Pre-existing, but the PR restates it in new prose; the one-token fix belongs here rather than documenting the wrong formula twice.
7. nit — a typo'd `model_name` falls through to `RRDBNetBuilder`, whose message lists the four RRDB names and omits `small_sr`. Raise from `from_config` with all five.
8. nit — `ModelBase[SuperResolutionConfig]` like `Sam`/`RTDETR`/`EfficientViT`, so `from_config`'s parameter type matches the generic.
9. nit — `docs/export_support/cases_models.py:669-693` still carries three "KORNIA BUG: abstract SuperResolution" notes and a `skip=`; rewrite the notes so the next survey run probes them (the snapshot can wait for that rerun).

Nothing else in flight collides. CUDA and real weight downloads were not exercised.

adimalkar:
Thanks for running it rather than reading it — and the P1 catch is fair: my `__init__` broke the ONNX path the body claims it preserves. All nine addressed, pushed at `83314b5` (rebased onto `main`; five PRs had landed since).

**P1 — ONNX metadata.** Confirmed: `input_image_size` / `output_image_size` / `pseudo_image_size` were class-body annotations only, and `to_onnx` reads all three. Set to `None` in `__init__`; the builders still overwrite what they know. Both families export now:

```
RRDB:    to_onnx(save=False) -> OK
SmallSR: to_onnx(save=False) -> OK
```

Pinned three ways, as you suggested and a bit beyond: a per-builder assertion that the three are initialised, an explicit `RRDBNetBuilder ... is None` test (fails at head), and an `importorskip("onnx")` + `to_onnx(save=False)` test for both families.

**P2 — `docs/source/models.rst`.** The "currently raises `TypeError` ... construct `RRDBNet` directly" paragraph is replaced with the working usage.

**P2 — `tests/models/test_rrdbnet.py`.** Stub dropped as its comment instructed; the assertions now run against the returned `SuperResolution` (`isinstance`, `.model`, `.name`, `.pre_processor`, `.post_processor`). The `super_resolution_module` import stays — it is what the `isinstance` check uses. 27 pass.

**P2 — export and document.** `SuperResolutionConfig` added to the module's `__all__` and to `kornia.contrib`'s, and `docs/source/contrib.rst` gains a Super Resolution section with `autoclass` for all four public names.

**P2 — CHANGELOG.** Now `... Fixes #4291. (#4335)`.

**nit — the `image_size * upscale_factor` formula.** I read "the one-token fix belongs here" as fixing the code rather than the prose, so `SmallSRBuilder` now computes `image_size * upscale_factor` instead of the hard-coded `* 3`; `upscale_factor=2, image_size=32` gives `output_image_size == 64`, and both docstrings are true. Say the word if you would rather keep the old metadata and correct the docs instead — it is a one-line flip either way, and it does change `output_image_size` for any `upscale_factor != 3`.

**nit — typo'd `model_name`.** `from_config` now raises with all five names before falling through.

**nit — the generic.** `ModelBase[SuperResolutionConfig]`.

**nit — `cases_models.py`.** The three "KORNIA BUG: abstract SuperResolution" notes are gone and the `RRDBNetBuilder` skip now gives its real reason (17M-parameter model, too heavy for the survey) rather than naming the fixed bug.

Verification on the rebased branch: 11 pass in the new file, 27 in `test_rrdbnet.py`, 88 across `tests/contrib`, `ruff check`/`format` clean, and `mypy` on the module is down to the one pre-existing `save` override you attributed to #4331.

Two things I did not do: CUDA and real weight downloads, same as your run. `save` at head still fails in `write_image` on a float32 PNG — leaving that to #4331 as you said.

ducha-aiki (changes requested):
Delta re-review at `83314b50` — CHANGES REQUESTED.

The requested construction, export-initialisation, public export/docs, RRDB builder cleanup, error handling, and CHANGELOG fixes are present. Focused validation passes: 15 passed in the new super-resolution suite (including both ONNX exports) and 5 passed in the RRDB builder tests; CI on this head is green. One new metadata assertion, however, pins a value that disagrees with the model it describes.

`SmallSRBuilder.build(pretrained=False, upscale_factor=2, image_size=32)` now sets `output_image_size == 64` at `kornia/contrib/super_resolution.py:353`, and `tests/contrib/test_super_resolution.py:90` asserts only that assignment. The wrapper's `ResizePreProcessor(224, 224)` at line 345 still resizes every input to 224 before the x2 network, so a real `(1, 3, 32, 32)` forward returns `(1, 3, 448, 448)`, not `(1, 3, 64, 64)`. I also exported this exact builder: the ONNX input is `(-1, 3, 32, 32)` and the inferred output is `(1, 3, 448, 448)`, while the object reports 64 and passes that false static size to `ONNXExportMixin.to_onnx`.

Please make the metadata and actual wrapper behavior agree, then pin the real forward/export shape rather than only the attribute assignment. Two coherent options are to make `image_size` configure the resize preprocessor, or to keep the fixed 224 preprocessor and derive the included-preprocessor output as `224 * upscale_factor` (with the raw-model export handled separately). The branch also needs a rebase for CI after this fix; GitHub reports it conflicting, though the current merge-tree operation itself is clean.

adimalkar:
You are right, and it was my change that introduced it — the previous round swapped one wrong number for another. Pushed at `5f6ac6e`, rebased onto `main`.

Reproduced your finding exactly before touching anything:

```
SmallSRBuilder.build(pretrained=False, upscale_factor=2, image_size=32)
  object reports output_image_size : 64
  real forward (1, 3, 32, 32)      : (1, 3, 448, 448)
```

The `ResizePreProcessor(224, 224)` resizes every input before the network, so the output side is `224 * upscale_factor` and does not depend on `image_size` at all. `image_size * 3` was wrong before my change; `image_size * upscale_factor` was wrong after it. Pinning the attribute rather than the model is exactly what let that through — a fair hit.

**Took the second of your two options** — keep the fixed preprocessor, derive the included-preprocessor output as `224 * upscale_factor` — because it is metadata-only and changes no runtime behaviour. `image_size` is documented as declaring the size fed to the wrapper, not as reconfiguring the resize, so making it drive the preprocessor (option one) would have changed what existing callers get from a forward. Happy to switch if you would rather `image_size` became load-bearing; that is the more invasive of the two and felt like your call rather than mine.

Now:

```
upscale_factor=2, image_size=32   reports 448   real (1, 3, 448, 448)
upscale_factor=3, image_size=None reports 672   real (1, 3, 672, 672)
upscale_factor=4, image_size=64   reports 896   real (1, 3, 896, 896)
```

`output_image_size` is set on both branches now, so it is known whether or not `image_size` is given. The 224 is named once (`_SMALL_SR_INPUT_SIZE`) at the preprocessor and the assignment, so the relationship is visible where it matters. Both docstrings stop promising `image_size * upscale_factor` and say what `image_size` actually does.

**Pinned the model, not the attribute**, per your ask — the weak assertion is gone, replaced by two tests:

- parametrized over `(2, 32)`, `(3, None)`, `(4, 64)`: runs a real forward and asserts `out.shape[-1] == model.output_image_size`, so the two cannot drift apart again
- an ONNX test that exports and reads the output dim off the graph. For the config that previously advertised 64, the graph now declares `[1, 3, 448, 448]` and the object agrees.

Also rebased as you asked; the CHANGELOG conflict is resolved keeping both sides, and I checked the diff against `main` has zero deletions so nothing upstream was dropped.

14 pass in the super-resolution suite, 27 in `test_rrdbnet.py`, `ruff check`/`format` clean.

ducha-aiki (dismissed):
Delta re-review at `5f6ac6e7` — **APPROVED in substance; rebase before merge**.

The metadata mismatch from the previous review is fixed. `SmallSRBuilder` now derives `output_image_size` from the fixed 224-pixel preprocessor and the requested upscale factor, and the new regression runs the real wrapper for factors 2, 3, and 4 before comparing both output axes with that metadata. The ONNX regression also verifies that the graph's static output side agrees with the wrapper.

Focused re-run at this exact head: the three eager shape cases plus the ONNX shape case passed (`4 passed`). All 41 hosted checks are green. A merge-tree against current `main` is clean locally, but GitHub still reports this head `CONFLICTING`; refresh or rebase it before merging and let the rebased checks complete.

adimalkar:
Rebased onto `8f29230`; head is now `789ab4c` and GitHub reports the branch mergeable.

**Conflict.** It was a real one, not a stale-mergeability artifact: `#4233` (`_cdist`) and `#4336` (`bbox_to_mask`) landed entries at the same position in *Unreleased → Bug fixes* as this PR's. I kept both sides — `git diff origin/main -- CHANGELOG.md` shows 0 deleted lines, so no upstream entry was dropped in the resolution.

**The diff you approved is unchanged.** Excluding `CHANGELOG.md`, the patch at `789ab4c` is byte-identical to the one at `5f6ac6e`:

```
git diff f4532f3..5f6ac6e -- . ':(exclude)CHANGELOG.md'   # approved head
git diff origin/main..789ab4c -- . ':(exclude)CHANGELOG.md'   # rebased head
# identical
```

Same 8 files, +254/−34. Locally, `tests/contrib/test_super_resolution.py` and `tests/models/test_rrdbnet.py` give 41 passed, 1 skipped on the rebased head.

**One red check, and I don't think it's this PR.** `tutorials / tutorials (3.12)` fails in `data_augmentation_segmentation.ipynb`:

```
UnidentifiedImageError: cannot identify image file 'causevic16semseg3.png'
```

That notebook fetches its sample image without checking the response:

```python
bytesio = io.BytesIO(requests.get(url).content)   # no raise_for_status()
with open(filename, "wb") as outfile:
    outfile.write(bytesio.getbuffer())
```

So a non-200 or truncated body from `github.com/kornia/data/raw/main/causevic16semseg3.png` gets written to disk as-is and PIL fails on it. Fetching that URL now returns `200`, `image/png`, a valid 1143×287 PNG — the asset is fine, the fetch wasn't. The notebook imports `kornia.augmentation`, nothing from `kornia.contrib.super_resolution`, and this PR touches no download path. (I did check whether `#4332` was implicated, since it landed in the same rebase and changes `download_file_from_url` — it isn't; the notebook uses bare `requests`.)

The other **40 checks are green**. Two things need your button, since I have neither admin rights to re-run the job nor a way to re-approve:

1. a re-run of `tutorials (3.12)`;
2. a re-approval — the force-push dismissed your approval under the stale-review rule.

Happy to push an empty commit instead if you'd rather retrigger that way, though it would dismiss the approval again, so the re-run seems cheaper.

ducha-aiki (approved):
Delta re-review at `789ab4c4` — approved; rerun the tutorial check before merge.

The branch is now rebased onto current main and GitHub reports it mergeable. Excluding CHANGELOG conflict placement, the code/docs/test patch has the same stable patch ID as the previously approved `5f6ac6e7` delta. The super-resolution metadata and actual wrapper/export shapes remain aligned as approved.

Validation at this exact head: 41 tests passed and 1 skipped across `tests/contrib/test_super_resolution.py` and `tests/models/test_rrdbnet.py`; `git diff --check` is clean; `import kornia` resolved to the isolated PR worktree. Forty hosted checks are green. The failed tutorial log shows `data_augmentation_segmentation.ipynb` raising `PIL.UnidentifiedImageError` on its downloaded `causevic16semseg3.png` sample, outside the paths changed here. Please rerun that check and merge once it is green.

adimalkar:
Thanks for the re-approval — and for independently confirming the patch-id equality; that was the thing I most wanted a second pair of eyes on.

On *"rerun that check and merge once it is green"*: I can't do either. `gh run rerun` gives me `Must have admin rights to Repository`, and I have no merge button here.

That said, I don't think the rerun gates anything. The PR is currently:

```
mergeable: MERGEABLE   mergeStateStatus: UNSTABLE   reviewDecision: APPROVED
```

`UNSTABLE` rather than `BLOCKED` means every **required** check passes — `collector` and `pre-commit.ci - pr` are both green — and the only red mark, `tutorials (3.12)`, carries no required badge. So this is mergeable as it stands; the red X is cosmetic.

If you'd rather merge on a fully green board, the one lever I do have is toggling the PR to draft and back to ready: `pr_test_cpu.yml` fires on `ready_for_review`, so that reruns the whole suite, and since it adds no commit it won't dismiss your approval the way the rebase did. Happy to do that on your word — I didn't want to flip the PR's state unasked.

PR #4336, opened 2026-09-06
fix(geometry): bbox_to_mask position grid is exact at any image size

tritsystem:

`bbox_to_mask`'s position grid is built via `torch.arange(..., dtype=boxes.dtype)` — exact pixel indices at whatever precision the caller's `boxes` happen to be. `float16` can only represent integers exactly up to 2048; past that, consecutive positions collapse onto the same value, so a `float16` `boxes` on an image taller or wider than 2048px silently mismasks rows or columns near the collapse.

```python
import torch, kornia

boxes32 = torch.tensor([[[1000., 1000.], [4090., 1000.], [4090., 3000.], [1000., 3000.]]])
boxes16 = boxes32.half()

mask32 = kornia.geometry.bbox_to_mask(boxes32, width=4096, height=4096)
mask16 = kornia.geometry.bbox_to_mask(boxes16, width=4096, height=4096)

print(mask32.double().sum().item())  # 6185091.0 (correct)
print(mask16.double().sum().item())  # 6188182.0 -- an entire extra row wrongly included
```

Same bug class as the already-merged `rotary-embedding-torch#50` (integer positions silently colliding under float16 past 2048) and the general "a position/index grid must stay fixed-precision regardless of the caller's own dtype" pattern used elsewhere in the codebase.

## Fix

Build `yy`/`xx` at a fixed `float32` regardless of `boxes.dtype`. `float32`-vs-`boxes.dtype` comparisons promote to `float32` automatically (PyTorch's own type promotion), so nothing downstream loses precision, and the returned mask still casts back to `boxes.dtype` exactly as documented.

Found via a static architectural-bug scanner (a personal tool) flagging position/index grids built at a variable, potentially-low-precision dtype instead of a fixed one — the same shape as the already-merged `rotary-embedding-torch#50`.

## What did you check?

- `float16` output now matches a `float32` reference **exactly** (byte-identical boolean mask) for a box spanning both axes past the 2048 collapse threshold.
- `float32`/`float64` inputs are byte-identical to before (promotion only goes up for `float64`; `float32` was already the built dtype).
- Added `test_bbox_to_mask_matches_float32_reference_for_float16_boxes_on_a_large_image`. Confirmed **red on unpatched code** (mismatched mask vs. the float32 reference) and **green with the fix**.
- Full `TestBbox2D` suite: 21 passed, no regressions.
- `ruff check` / `ruff format --check` clean on both changed files.

## How did AI help?

A static scanner flagged the position-grid-at-variable-dtype pattern; I reproduced the exact float16 mismasking on real coordinates, traced the root cause to the `2048` integer-precision boundary, wrote the fix and the regression test, and verified fail-on-unpatched / pass-with-fix by running the code directly. I reviewed the root cause, the fix, and the test before opening this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -352,6 +352,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug fixes
 
 
+* `bbox_to_mask` built its pixel-position grid in the box dtype. With `float16` boxes (and
+  `bfloat16` boxes) on images wider or taller than 2048 px (256 px for `bfloat16`) consecutive
+  pixel positions collapsed, so rows and columns near the collapse were mismasked; the grid is
+  now built in `float32` and `float16`/`bfloat16` results are byte-identical to `float32`/`float64`.
+  `RandomErasing` and `RandomCutMixV2` build their masks through it. (#4336)
+
 * `kornia.io.load_image` and `write_image` work on the kornia_rs that a plain `pip install kornia`
   resolves. kornia_rs 0.1.11 moved its image readers and writers from the package root into
   `kornia_rs.io`, and kornia kept calling the root, so on 0.1.11 and newer `load_image` raised
--- a/kornia/geometry/bbox.py
+++ b/kornia/geometry/bbox.py
@@ -313,6 +313,12 @@ def bbox_to_mask(boxes: torch.Tensor, width: int, height: int) -> torch.Tensor:
     Note:
         It is currently non-differentiable.
 
+    Note:
+        The pixel-position grid is built in ``float32`` regardless of ``boxes.dtype``, so
+        ``float16``/``bfloat16`` boxes are masked exactly even on images wider or taller than
+        the integer-exact range of those dtypes (2048 px for ``float16``, 256 px for
+        ``bfloat16``). The result is still returned in ``boxes.dtype``.
+
     Examples:
         >>> boxes = torch.tensor([[
         ...        [1., 1.],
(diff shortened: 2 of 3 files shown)


ducha-aiki:
## Review at `3c8e7bae` by execution — APPROVE after two small edits

The defect is real and the fix is the right one; the bfloat16 half of the same defect is fixed by it too but not pinned, and the entry for the CHANGELOG is missing.

**Verified**

| Claim | Result |
| --- | --- |
| Body repro | Held verbatim: base float16 mask sum 6188182 vs the float32 6185091; head equal (`torch.equal`). Same on MPS. |
| float32/float64 byte-identical | 64 seeded random boxes, sha256 of the mask bytes identical base vs head (also MPS float32). |
| float16 at head matches the float32 reference | Exactly; on base 20281 pixels differ on CPU. |
| float64 precision kept | A float32 grid vs float64 boxes promotes to float64 (a near-integer float64 box gives sum 8, not the 16 a float32 comparison would give). |
| Pin fails on base | 1 failed on base at float32/float64, at the half dtypes, and on MPS; passes at head. |
| Suites | `test_bbox.py` 90 passed / 6 xfailed; half dtypes 45 / 6 (base 44 / 6); MPS 45 passed; `RandomErasing` / `RandomCutMixV2` callers identical counts base vs head; dynamo 4 passed; `pre-commit` 12 hooks passed. |
| `bbox_to_mask3d` | Does not share the defect: it builds a `long` grid and truncates with `.long()`; measured equal at float16 on a 4096² volume. `Boxes.to_mask` / `Boxes3D.to_mask` use integer aranges too. |

**Asks**

1. **P2 — CHANGELOG entry.** None at head. Under `## Unreleased` / `### Bug fixes`, citing `(#4336)`: "`bbox_to_mask` builds its pixel grid in `float32` regardless of the box dtype. With `float16` boxes (and `bfloat16` boxes) on images wider or taller than 2048 px (256 px for `bfloat16`) the grid was built in the box dtype and consecutive pixel positions collapsed, so rows and columns near the collapse were mismasked; `float32`/`float64` results are byte-identical. `RandomErasing` and `RandomCutMixV2` build their masks through it."
2. **P2 — pin bfloat16 too.** The new test hardcodes `.half()` and takes only `device`, so `--dtype=bfloat16` re-runs the float16 case and the bfloat16 branch of the same defect has no pin. Measured on base: 6386913 vs 6458100 with box (1000, 1000)–(4080, 3072) on 4096², or 60501 vs 61004 with (100, 100)–(300, 400) on a 512² image; head equal. Parametrize the half dtype (`[torch.float16, torch.bfloat16]`, `boxes32.to(half_dtype)`) and use coordinates exactly representable in both: 1000, 3072, 4080 are; 4090 rounds to 4096 and 3000 to 3008 in bfloat16 (measured), so the current box would silently test a different rectangle there. Rename the test accordingly.
3. nit — one `Note:` line in the `bbox_to_mask` docstring that the grid is built in float32, so half-precision boxes are masked exactly on any image size.
4. nit — the in-code comment says the comparisons "promote to float32"; with float64 boxes they promote to float64. "The wider of the two."

CUDA not exercised.

ducha-aiki (approved):
Delta re-review at `ea8f1e1b` — APPROVED; rerun the cancelled float16 job before merge.

Both requested edits landed. The regression test now covers float16 and bfloat16 with coordinates exactly representable in both formats, and the CHANGELOG documents the two precision thresholds and cites #4336. The docstring/comment accurately describe promotion with float64 boxes. Focused `bbox_to_mask` validation passes 24/24, including both new half-dtype cases, and the bfloat16, MPS, typecheck, dynamo, and ordinary CPU CI jobs are green.

The red status comes from a cancelled job: the float16 job ran for exactly six hours and was cancelled, so the collector failed. It produced no failing-test log and is consistent with the known #4214 Intel runner hang. Please rerun that job; I found no code issue in this delta.

PR #4337, opened 2026-09-06
fix(augmentation): register fill_value/sigma as buffers, not plain attrs

tritsystem:

`RandomChannelDropout.fill_value` and `RandomGaussianBlurGenerator.sigma` (when passed as a `Tensor`) were plain, non-`register_buffer`'d tensor attributes — the same shape as #4079 (`SIFTDescriptor.gk`) and #4319 (`RenderingDeFMO.times`), found via the same static architectural scanner.

Unlike #4319, **neither is a crash risk**: both re-derive device *and* dtype in a single `.to()` call inline on every use (`fill_value.to(device=..., dtype=...)`; `sigma.to(device=..., dtype=...)`), so `.half()`-ing the module never produced a dtype mismatch. I confirmed this by direct execution before writing the fix. This is a buffer-hygiene fix (`state_dict()` / `named_buffers()` / TorchScript-export visibility), not a fix for a crash.

## Fix

`register_buffer(..., persistent=False)`, same convention as #4079/#4319 — `persistent=False` keeps `state_dict()` keys unchanged. `RandomGaussianBlurGenerator.sigma` can also be a plain `(min, max)` tuple of Python floats with no device/dtype at all; `register_buffer` requires a `Tensor`, so that case is correctly left as a plain attribute.

## What did you check?

- New tests confirm buffer registration (`named_buffers()`), that `.to()`/`.half()` correctly moves the buffer, and — for `sigma` — that the tuple case is correctly left as a plain attribute (not silently registered as an empty buffer or similar).
- Full `TestRandomChannelDropout` and `TestRandomGaussianBlur` suites pass, no regressions.
- `ruff check` / `ruff format --check` clean on all changed files.

## How did AI help?

A static scanner (a personal tool, the same one behind #4319) flagged both of these as the plain-attribute-not-buffer shape. I verified by direct execution that unlike #4319's crash, both re-derive device and dtype together and never crash, wrote the register_buffer fix following #4079/#4319's exact convention, added the tests, and ran the full suites before opening this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -357,6 +357,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `Module.to()`/`.half()`/`.cuda()` and appear in `named_buffers()` like the rest of the module's
   state; `state_dict()` keys are unchanged. Neither crashed before this change (both re-derived
   device and dtype inline on every call), so this is a hygiene fix, not a bug fix for a crash.
+  (#4337)
 
 * `kornia.io.load_image` and `write_image` work on the kornia_rs that a plain `pip install kornia`
   resolves. kornia_rs 0.1.11 moved its image readers and writers from the package root into
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -357,6 +357,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `Module.to()`/`.half()`/`.cuda()` and appear in `named_buffers()` like the rest of the module's
   state; `state_dict()` keys are unchanged. Neither crashed before this change (both re-derived
   device and dtype inline on every call), so this is a hygiene fix, not a bug fix for a crash.
+  One user-visible effect: after `.half()`/`.bfloat16()` the fill value is now held in that dtype,
+  so an input already in the module's dtype is unaffected, while a `float32` input fed to a
+  half-converted `RandomChannelDropout(fill_value=0.3)` is filled with `0.300048828125` (float16)
+  or `0.30078125` (bfloat16) rather than `0.3` -- the same buffer semantics as `Normalize`.
   (#4337)
 
 * `kornia.io.load_image` and `write_image` work on the kornia_rs that a plain `pip install kornia`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -352,6 +352,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug fixes
 
 
+* `RandomChannelDropout.fill_value` and `RandomGaussianBlurGenerator.sigma` (when passed as a
+  Tensor) are registered as non-persistent buffers instead of plain attributes, so they move with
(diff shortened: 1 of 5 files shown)


ducha-aiki:
## Review at `0b91bd0b` by execution — APPROVE after one CHANGELOG sentence

**Verified.** On `main`, `.to("mps")` / `.half()` plus a forward never crashed for either module across 8 device/dtype combinations — the attributes just stayed CPU float32 and were re-derived per call, exactly as the body says. At head the buffers move with the module, `state_dict()` keys are unchanged (empty for both; a strict empty load round-trips), `named_buffers()` lists them, CPU and MPS float32/float16 forwards pass, and deepcopy/pickle work, including inside `AugmentationSequential`. `register_buffer` before `Module.__init__` raises `AttributeError` (executed), and both `__init__`s call `super().__init__()` first, so the ordering is right. A tuple `sigma` stays a plain attribute; float64, `requires_grad` and MPS tensor sigmas all construct. Fail-on-base: 4 failed (`'fill_value' in {}`, `'sigma' in {}`, two dtypes), pass at head. Suites: 29 passed float32/float64; half dtypes 15 passed each (base 12); MPS 15 passed; the slow dynamo test passes with `--runslow`, and the 1-ulp inductor noise on the blur is identical on `main`. CHANGELOG under `### Bug fixes` with the PR number, nothing deleted; `pre-commit` 12 hooks passed.

**Asks**

1. **P2 — one user-visible effect the entry omits.** After `.half()` / `.bfloat16()` the fill value is now held in that dtype, so a float32 input fed to a half-converted `RandomChannelDropout(0.3)` is filled with 0.300048828125 (float16) or 0.30078125 (bfloat16) instead of 0.3 (measured). Standard buffer semantics, the same as `Normalize`, so a disclosure rather than an objection: add a sentence to the CHANGELOG entry saying inputs in the module's dtype are unaffected and a float32 input to a half-converted module gets the half-rounded value.
2. nit — `test_augmentation.py:4553-4557`: `input_tensor` exists only to read its `.device`; compare the buffer's device the way the generator test does, and assert the forward still fills with the value after the move (`aug(torch.zeros(1, 3, 3, 3, device=device, dtype=dtype))` selects 0.3 in the dropped channel), so the pin covers use, not only registration.
3. nit — `test_random_generator.py:1562`: the module-level imports can go to the top of the file with the others rather than inside each test.
4. nit — the comments cite "#4079/#4319" as the package convention; `plain_uniform.py:73-76` registers its factors as persistent buffers, so the package is not uniform. Not for this PR to fix, just do not claim uniformity.

The TorchScript-visibility claim is unobservable here: `jit.script` of the generator raises on both trees. CUDA not exercised.

ducha-aiki (approved):
Delta re-review at `46b70e1b` — APPROVED.

The requested user-visible dtype disclosure is now explicit in the CHANGELOG, including the measured float16/bfloat16 rounding behavior. The follow-up tests also exercise the moved `fill_value` in a real forward and the parent-module `_apply()` path for the Gaussian blur generator, while keeping both buffers non-persistent. Focused CPU validation passes (2 channel-dropout buffer cases and 3 Gaussian-generator buffer cases), the diff is clean, and all 40 reported CI checks are green. I found no remaining actionable issue.

ducha-aiki (approved):

PR #4338, opened 2026-09-06
fix(solvers): stop solve_cubic's D>0 branch poisoning other rows' gradients

### LINKED ISSUE #4334: solve_cubic: batched gradients are NaN for three-real-root rows whenever the batch also holds a one-real-root row (D>0 branch computes sqrt(D) and -Q/AD on every |R|>0 row)
## Summary

`solve_cubic`'s batched backward is `nan` for every three-real-root row (`D < 0`, `R != 0`) as soon as the same batch also holds a one-real-root row (`D > 0`, `Q != 0`). Each row differentiates fine on its own; the contamination comes from the `D > 0` branch, which evaluates `sqrt(D)` and then `-Q / AD` over **every** row with `|R| > 1e-16`, not only the `D > 0` rows. `solve_quartic` inherits it through its resolvent cubic, so a batch mixing a four-real-root quartic with a two-real-root one gets `nan` gradients for the four-root row.

This is independent of #4290: it hits well-conditioned, distinct roots, and #4299 / #4329 (which fix the `acos` boundary inside the `D <= 0` branch) do not change it. Found while checking those two PRs against a random batch of repeated-root cubics: 26,282 of 50,000 rows had non-finite gradients on `main` and on both heads, and every one of them was finite when run alone.

## Reproduce

```python
import torch
from kornia.geometry.solvers import solve_cubic, solve_quartic

three = [1., -7., 14., -8.]   # (x-1)(x-2)(x-4): D < 0, three real roots, R != 0
one   = [1.,  0.,  1., -2.]   # (x-1)(x^2+x+2): D > 0, one real root, Q != 0

for rows in ([three], [three, one]):
    c = torch.tensor(rows, dtype=torch.float64, requires_grad=True)
    solve_cubic(c).sum().backward()
    print(len(rows), "row(s): grad of the three-root row =", c.grad[0].tolist())
# 1 row(s):  [-7.0, -1.0, -1.9e-16, -1.7e-16]
# 2 row(s):  [nan, nan, nan, nan]

q4 = [1., -10., 35., -50., 24.]   # (x-1)(x-2)(x-3)(x-4)
q2 = [1., 0., 0., 0., -16.]       # x^4 - 16: two real roots
for rows in ([q4], [q4, q2]):
    c = torch.tensor(rows, dtype=torch.float64, requires_grad=True)
    solve_quartic(c).sum().backward()
    print(len(rows), "row(s): grad of the four-root quartic =", c.grad[0].tolist())
# 1 row(s):  [-10.0, -1.0, 0.0, 0.0, 0.0]
# 2 row(s):  [nan, nan, nan, nan, nan]
```

Same in float32. The forward values are correct in every case; only the gradient is poisoned.

## Mechanism

`kornia/geometry/solvers/polynomial_solver.py`, `D > 0` branch on `main` (`ea07adac`):

```python
mask_R_positive = R_abs > 1e-16
if torch.any(mask_R_positive):
    AD[mask_R_positive] = torch.pow(R_abs[mask_R_positive] + torch.sqrt(D[mask_R_positive]), 1 / 3)
    ...
    BD[mask_R_positive] = -Q[mask_R_positive] / AD[mask_R_positive]
x0_D_positive = AD[mask_D_positive] + BD[mask_D_positive] - b_a_3[mask_D_positive]
```

`mask_R_positive` is a condition on `R`, not on `D`. For a `D < 0` row it computes `sqrt(D) = nan`, so `AD = nan` for that row. The row is never *read* (only `mask_D_positive` rows reach `x0_D_positive`), but `BD[...] = -Q / AD` is still a node in the graph, and `DivBackward0` for the unused row is `0 / nan = nan`, which lands on `Q.grad` and from there on every coefficient. `torch.autograd.detect_anomaly()` names exactly that node: `Function 'DivBackward0' returned nan values in its 0th output.`

The existing `test_gradcheck` fixture `(2x+1)(x-2)(x+3)` has `R == 0`, so it is excluded by `mask_R_positive` and never sees this; a single-row test cannot see it either.

## Suggested fix

Restrict the `D > 0` computations to `mask_D_positive & mask_R_positive` rows (index once, compute on the compacted rows, scatter back), so no `sqrt` of a negative `D` is ever built. A pin: a two-row batch of the two cubics above, `backward()`, assert finite — it fails on `main` and cannot pass by accident. Adding the quartic pair as well covers `solve_quartic`'s resolvent path.

Refs #4290, #4299, #4329, #4229 (the two `sqrt(clamp(...))` sites it lists in `solve_quartic` are on the same branch structure).

## Environment

kornia `main` at `ea07adac`, torch 2.14.0, CPU, macOS.

dchaudhari7177:
Closes #4334.

The `D > 0` branch selected its rows with `abs(R) > 1e-16` — a condition on `R`, not on `D` — so it evaluated `sqrt(D)` on `D < 0` rows too. Those rows are never read back out of `AD`/`BD`, but `-Q / nan` stays in the graph, `DivBackward0` returns `nan` for them, and that lands on `Q.grad` and from there on every coefficient.

One line: intersect the mask with `mask_D_positive`.

**Verified by execution** (CPU, torch 2.5.0). The issue's repro, before → after:

```
solve_cubic,   1 row : [-7.0, -1.0, -1.9e-16, -1.7e-16]
solve_cubic,   2 rows: [nan, nan, nan, nan]        ->  [-7.0, -1.0, -1.9e-16, -1.7e-16]
solve_quartic, 2 rows: [nan, nan, nan, nan, nan]   ->  [-10.0, -1.0, 0.0, 0.0, 0.0]
```

At scale, a seeded random batch (float64):

| | `main` | this branch |
|---|---|---|
| `solve_cubic`, 50 000 rows | 12 312 rows with non-finite grad | **0** |
| `solve_quartic`, 20 000 rows | 5 684 | **0** |

The forward pass is untouched: `solve_cubic` on a 5 000-row batch is `torch.equal` to evaluating the same rows one at a time.

**Tests.** Three pins, in the shape the issue suggests — a mixed two-row batch, asserting both finiteness *and* that the batched gradient equals the single-row one, so a later fix that merely suppresses `nan` does not pass. Plus a forward-value pin, which passes on `main` on purpose: the values were always right, and that is what a gradient fix must not disturb. On `main`: **4 failed, 2 passed**; at this head: **6 passed**. Whole file, `--dtype=float32,float64`: **123 passed**.

**One thing worth flagging.** The issue's quartic pair uses `x^4 - 16` as the second row, and that polynomial has non-finite gradients *on its own* — `[nan, nan, nan, 0.0, 0.0]`, with or without this change. That is a separate defect in `solve_quartic`'s own `sqrt` sites, which looks like the #4229 territory the issue mentions, so I scoped the quartic assertion to row 0 rather than pin an unrelated bug. Happy to take that one next if you want it.

`ruff check` / `ruff format` clean at v0.16.4. `grep -rnE "#4334|issues/4334" kornia/ tests/` hits only the new test pins. CHANGELOG entry under `### Bug fixes`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -410,6 +410,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   form and reshapes internally, but both callers index dim 1 as the vertex axis, so the box axis was
   read as the vertices: an out-of-bounds error with one box, and three `(1, 3)` tensors -- one value
   per coordinate rather than per box -- with eight. (#4248, #4351)
+* `solve_cubic` and `solve_quartic` return finite gradients for a row whose neighbours in the
+  same batch take a different branch. The `D > 0` branch selected its rows with `abs(R) > 1e-16`
+  alone, so it evaluated `sqrt(D)` on `D < 0` rows too. Those rows are never read back, but the
+  resulting `-Q / nan` stays in the graph and its backward returns `nan`, which reaches every
+  coefficient: a three-real-root cubic differentiated correctly on its own and gave `nan` as soon
+  as a one-real-root cubic shared the batch, and `solve_quartic` inherited it through its
+  resolvent cubic. The forward values are unchanged. (#4334, #4338)
 
 * `RenderingDeFMO` (used by `DeFMO`) no longer crashes on a half-precision forward pass. Its rendering
   time-steps (`times`) were a plain Python attribute, not a registered buffer, so `nn.Module.to()` never
--- a/kornia/geometry/solvers/polynomial_solver.py
+++ b/kornia/geometry/solvers/polynomial_solver.py
@@ -214,7 +214,10 @@ def solve_cubic(coeffs: torch.Tensor) -> torch.Tensor:
         AD = torch.zeros_like(R)
         BD = torch.zeros_like(R)
         R_abs = torch.abs(R)
-        mask_R_positive = R_abs > 1e-16
+        # Intersect with mask_D_positive: sqrt(D) on a D <= 0 row is nan, and
+        # although such a row is never read out of AD/BD, `-Q / nan` stays in
+        # the graph and its backward poisons every coefficient's gradient.
+        mask_R_positive = (R_abs > 1e-16) & mask_D_positive
         if torch.any(mask_R_positive):
             AD[mask_R_positive] = torch.pow(R_abs[mask_R_positive] + torch.sqrt(D[mask_R_positive]), 1 / 3)
             mask_R_positive_ = R < 0
--- a/testing/half_precision_xfails/cpu_float16.txt
+++ b/testing/half_precision_xfails/cpu_float16.txt
@@ -398,6 +398,7 @@ call	builtins.AssertionError	tests/geometry/liegroup/test_so3.py::TestSo3::test_
 call	builtins.AssertionError	tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_convention_gradient_is_finite_for_a_pure_biquadratic_4229[cpu-float16-coeffs0-expected0-expected_grad0]
 call	builtins.AssertionError	tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_convention_gradient_is_finite_for_a_pure_biquadratic_4229[cpu-float16-coeffs1-expected1-expected_grad1]
 call	builtins.AssertionError	tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_convention_gradient_is_finite_for_a_pure_biquadratic_4229[cpu-float16-coeffs2-None-expected_grad2]
+call	builtins.AssertionError	tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_convention_gradient_does_not_leak_across_batch_rows_4334[cpu-float16]
 call	builtins.AssertionError	tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_random[cpu-float16]
 call	builtins.AssertionError	tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_solve_quartic[cpu-float16-coeffs0-expected_solutions0]
 call	builtins.AssertionError	tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_solve_quartic[cpu-float16-coeffs3-expected_solutions3]
--- a/tests/geometry/solvers/test_polynomial_solver.py
+++ b/tests/geometry/solvers/test_polynomial_solver.py
@@ -506,10 +506,7 @@ class TestQuarticSolver(BaseTester):
         mixed = torch.tensor([four, two], device=device, dtype=dtype, requires_grad=True)
         solver.solve_quartic(mixed).sum().backward()
 
-        # Row 0 only. Row 1's own gradients are not this fix's to guarantee: they depend on
-        # solve_quartic's separate sqrt sites (#4229), which are torch-version sensitive --
-        # `x^4 - 16` alone is non-finite on torch <= 2.9.1 and finite on 2.14, where clamp
-        # zeroes the boundary gradient. Asserting over the whole batch would pin that moving
-        # target. What #4334 is about is row 0 being poisoned by row 1's presence.
+        # Check row 0 for the cross-batch contamination from #4334. Row 1's separate
+        # zero-radicand gradient convention was fixed in #4339 and is covered above.
         assert bool(torch.isfinite(mixed.grad[0]).all()), mixed.grad
         self.assert_close(mixed.grad[0], alone.grad[0])
--- a/tests/geometry/solvers/test_polynomial_solver.py
+++ b/tests/geometry/solvers/test_polynomial_solver.py
@@ -124,6 +124,33 @@ class TestCubicSolver(BaseTester):
         cubic_roots.sum().backward()
         assert bool(torch.isfinite(cubic_coeffs.grad).all()), cubic_coeffs.grad
 
+    def test_convention_gradient_does_not_leak_across_batch_rows_4334(self, device, dtype):
+        # #4334: the D > 0 branch keyed its work off `abs(R) > 1e-16` alone, so it evaluated
+        # sqrt(D) on D < 0 rows too. Those rows are never read back, but `-Q / nan` stays in
+        # the graph and DivBackward0 returns nan, which reaches every coefficient. The result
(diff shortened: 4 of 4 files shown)


ducha-aiki:
## Review at `4d8da2c8` by execution — APPROVE after one CI fix and one CHANGELOG edit

The one-line mask intersection is the right and complete fix for #4334. What blocks it is the red `cpu-float16` leg, which is a pre-existing float16 *forward* defect the new quartic pin happens to stand on.

**Verified**

| Claim | Result |
| --- | --- |
| Issue repro | `[(x-1)(x-2)(x-4), (x-1)(x²+x+2)]`: `[nan×4]` on base → the single-row values at head, float32 and float64, torch 2.14.0 and 2.5.1. Quartic pair likewise. |
| At scale | My own seed: cubic 19,526 / 50,000 non-finite rows → 0; quartic 12,774 / 20,000 → 0. |
| Forward untouched | `torch.equal` to `main` on the seeded batches. |
| Batch contamination | An 8-row quartic batch with one row per branch is finite and equal to the single-row gradients at head; on `main` 3 of 7 rows are NaN. |
| Pins | Fail-on-base 4 failed / 2 passed (the forward-value pin passes on base on purpose, as the body says); head 6 passed. Whole file 123 passed at float32/float64; 60 passed on MPS. |
| #4299 interplay | Its `acos` boundary handling is untouched; the file's suite passes at both heads. |
| CHANGELOG | Under `### Bug fixes`, no lines of `main`'s deleted. AGENTS grep: `#4334` appears only in the new pins. `pre-commit` clean. |

**Asks**

1. **P1 — the blocking `half-precision (float16)` leg is red** on `TestQuarticSolver::test_convention_gradient_does_not_leak_across_batch_rows_4334[cpu-float16]` (`AssertionError: tensor([[nan, nan, 0., nan, 0.]])`). Deterministic, and not something this fix can address: on `main` the float16 *forward* of `solve_quartic` is already `[inf, -inf, 0, 0]` for `(x-1)(x-2)(x-3)(x-4)` and `[nan×4]` for `x⁴-16` (resolvent overflow; `zero_tol=1e-12` rounds to 0 in float16), so there is no finite gradient to protect. The file's other float16 quartic failures are recorded in `testing/half_precision_xfails/cpu_float16.txt` (lines 398–402); add one line next to them, tab-separated like its neighbours:
   `call	builtins.AssertionError	tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_convention_gradient_does_not_leak_across_batch_rows_4334[cpu-float16]`
   bfloat16 passes, so nothing goes into `cpu_bfloat16.txt`. A manifest line is better than a `pytest.skip`: it stays a strict xfail that flips to XPASS and forces its own deletion the day someone fixes the float16 forward.
2. **P2 — CHANGELOG** cites only the issue; the file's convention is the PR number, `(#4334, #4338)` or "Closes #4334. (#4338)".
3. nit — the comment at `test_polynomial_solver.py:452-456` ("`x⁴ - 16` has non-finite gradients on its own … that this change does not touch") is true on torch ≤ 2.9.1 only; on 2.14 `clamp` zeroes the boundary gradient and `x⁴-16` alone is finite. It also goes stale the moment #4339 lands. Whichever of the two merges second rewords it; the row-0-only scope of the assertion is still the right call here.

**Merge notes.** #4339 touches the same file and both PRs append to `TestQuarticSolver`, so `git merge-tree` of the two heads conflicts in the test file (solver and CHANGELOG auto-merge). The combined tree passes: 127 passed on torch 2.14 and 2.5.1. This one first. And a heads-up, not for this PR: while measuring the quartic gradients I found that `solve_quartic`'s forward is wrong for ~28 % of general quartics and most biquadratics on `main` (the `argmax` over `R_sq` candidates ranks `solve_cubic`'s zero placeholders) — filed as #4346. It should land after #4338/#4339 to avoid a three-way conflict, if you want it.

dchaudhari7177:
Thanks for the execution review — all three addressed in `189a36eb`.

**P1 — the red `cpu-float16` leg.** Added the manifest line exactly as you specified, tab-separated, alphabetically among the file's other `TestQuarticSolver` entries (now line 398):

```
call	builtins.AssertionError	tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_convention_gradient_does_not_leak_across_batch_rows_4334[cpu-float16]
```

Verified it takes effect rather than merely being present:

```
pytest ...::test_convention_gradient_does_not_leak_across_batch_rows_4334 \
  --device=cpu --dtype=float16 --xfail-known-failures --known-failure-profile=cpu-float16
1 xfailed
```

Agreed on the reasoning for a manifest line over a `pytest.skip` — it stays strict, so it XPASSes and forces its own deletion once the float16 forward is fixed. `cpu_bfloat16.txt` untouched, since bfloat16 passes.

**P2 — CHANGELOG.** Now `(#4334, #4338)`.

**nit — the `x⁴ - 16` comment.** You're right that it was only true on torch ≤ 2.9.1. Rather than restate a version-specific fact that #4339 would immediately stale, the comment now says why row 1 is out of scope at all — its gradients depend on the separate `#4229` sqrt sites, which are torch-version sensitive (non-finite on ≤ 2.9.1, finite on 2.14 where `clamp` zeroes the boundary gradient), so asserting over the whole batch would pin a moving target. The row-0-only scope is unchanged.

Local: 123 passed on the file at float32/float64; the float16 node xfails under the profile.

Noted on the merge order — happy for this to go first, and I'll rebase #4339 onto it once this lands rather than leaving you the test-file conflict.

ducha-aiki (dismissed):
Delta re-review at `cf5e03b4` — **APPROVE once the remaining CI job completes**.

Both prior asks landed. The new `cpu_float16.txt` entry matches the exact quartic node and runs as a strict xfail (`1 xfailed` locally with the focused profile); the CHANGELOG now cites both #4334 and #4338. The version-sensitive comment was also corrected without widening the assertion beyond the row whose cross-batch contamination this PR fixes.

Re-run at this head: `tests/geometry/solvers/test_polynomial_solver.py --device=cpu --dtype=float32,float64` → 123 passed. The focused float16 node xfails as recorded. GitHub reports `MERGEABLE`; 38 checks are green and only `half-precision (float16)` is still running, with no failed checks.

Merge this before #4339: both append to `TestQuarticSolver`, and #4339 should rebase afterward and retain both PRs' manifest entries.

ducha-aiki (dismissed):
Approved at `da058828` after the maintainer-authorized recovery.

Rebased onto main f4532f39, preserving the cubic batch-gradient fix and its regression/manifest entries. Solver tests: 123 passed at float32/float64; local CPU float16 profile: 57 passed, 6 expected xfails. The old four-hour stalled run was cancelled; new-head CI has restarted.

PR #4339, opened 2026-09-06
fix(solvers): guard solve_quartic's sqrt gradients at a zero radicand

dchaudhari7177:
Fixes the two `solve_quartic` `sqrt(clamp(...))` sites that #4229 lists. Found while writing the pins for #4334 — the issue's own quartic repro uses `x^4 - 16` as its second row, and that polynomial has non-finite gradients *on its own*, which is a different defect from the batch leak #4334 describes.

**The pattern does not guard what it looks like it guards.** `d(sqrt)/dx` is unbounded at 0, and on torch below 2.14 `clamp` passes the incoming gradient straight through at the bound rather than zeroing it. Measured here on the torch this machine has:

```
sqrt(clamp(x, min=0.0)) grad at x = 0  ->  inf      (torch 2.5.0+cpu)
```

kornia declares `torch>=2.5.1`, so the guard is a no-op across the older half of the supported range.

**What reaches it.** A quartic with no `x^3` and no `x^2` term puts `R_sq` exactly on 0:

| | `main` | this branch |
|---|---|---|
| `x^4 - 16` | `[nan, nan, nan, 0, 0]` | `[finite]`, roots `±2` |
| `x^4 - 1` | `[nan, nan, nan, 0, 0]` | `[finite]`, roots `±1` |
| `(x^2-1)(x^2-4)` | finite | finite (unchanged) |
| `x^4 + 1` | finite | finite (unchanged) |

**The fix** substitutes a safe radicand under the `sqrt` and selects with `torch.where`, so autograd never differentiates `sqrt` at 0. That is not a new idiom — `solve_quadratic`, twenty lines up in the same module, already does exactly this for its own discriminant, with a comment saying why. This makes the three root-finders consistent.

**Forward is untouched.** On a seeded 20 000-row random batch, `solve_quartic`'s output is `np.array_equal` to `main`'s. Single-row gradient sweep over 3 000 random rows: 0 non-finite on both, as expected — random coefficients essentially never land `R_sq` on exactly 0, which is why this needed a structured pin rather than a fuzz test.

**Tests:** parametrized over the two biquadratics, asserting finite gradients *and* pinning the forward roots so a later fix that repairs the gradient by moving the value is caught. On `main`: **4 failed**. At this head: whole file, `--dtype=float32,float64`, **121 passed**.

Independent of #4338 (#4334) — that one is the `D > 0` branch in `solve_cubic`, this one is `solve_quartic`'s own `sqrt` sites. They touch different lines and can go in either order.

`ruff check` / `ruff format` clean at v0.16.4. CHANGELOG entry under `### Bug fixes`.


ducha-aiki:
## Review at `10388c20` by execution — CHANGES REQUESTED (one CI fix, one missing pin, one docstring)

The safe-radicand + `torch.where` form is right and matches `solve_quadratic`'s idiom. Measured on both ends of the supported torch range.

**Verified**

| Claim | Result |
| --- | --- |
| `sqrt(clamp(x, 0))` gradient at 0 | `inf` on torch 2.5.1, `0.0` on 2.14.0 — the #4229 mechanism, measured. |
| `x⁴-16`, `x⁴-1` | On 2.5.1: `[nan, nan, nan, 0, 0]` on base → `[0, -0.5, 0, 0, 0]` at head, roots ±2 / ±1. On 2.14: value-identical to `main` (the guard is a no-op there). |
| Forward untouched | `torch.equal` to `main` on 20,000 seeded rows. |
| Pins | 4 failed on base under torch 2.5.1; **4 passed on base under 2.14** — inherent, the 2.5.1 / 2.9.1 CI legs carry the discrimination. Whole file 121 passed at float32/float64. |
| CHANGELOG | Under `### Bug fixes`, nothing deleted. `pre-commit` clean. |

**Asks**

1. **P1 — the blocking `half-precision (float16)` leg is red** on both parametrizations of `test_convention_gradient_is_finite_for_a_pure_biquadratic_4229[cpu-float16-…]` (`AssertionError: tensor([[nan, nan, nan, nan, 0.]])`). Same cause as on #4338: `solve_quartic`'s float16 *forward* is already `[nan×4]` for `x⁴-16` on `main` (`zero_tol` underflow), so nothing a gradient guard can do. Add two lines to `testing/half_precision_xfails/cpu_float16.txt` next to lines 398–402, tab-separated like their neighbours:
   `call	builtins.AssertionError	tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_convention_gradient_is_finite_for_a_pure_biquadratic_4229[cpu-float16-coeffs0-expected0]`
   `call	builtins.AssertionError	tests/geometry/solvers/test_polynomial_solver.py::TestQuarticSolver::test_convention_gradient_is_finite_for_a_pure_biquadratic_4229[cpu-float16-coeffs1-expected1]`
   bfloat16 passes. A manifest line rather than a skip, for the reason given on #4338. If ask 2 adds a third parametrization, measure it in float16 before adding its line.
2. **P2 — the second `sqrt` site (`polynomial_solver.py:331-338`, the `R ≈ 0` fallback) is changed but has no pin.** It is reachable: `(x²+1)²` = `[1, 0, 2, 0, 1]` gives `R_sq < 0 → R = 0 → radicand == 0` exactly; on torch 2.5.1 the gradient is `[nan×5]` on base and `[0×5]` at head (measured). Add it to the parametrization with an all-zero expected output (the "no real roots" case needs a small tweak to the `expected` handling — compare `roots[0]` to zeros rather than sorting the first two).
3. **P2 — docstring convention**, the way #4299 was asked to state it for `solve_cubic`: at `R_sq == 0` (pure biquadratic) and at the `radicand == 0` fallback, backward suppresses the diverging `sqrt` derivative, so the gradient is finite but is not the root Jacobian. The CHANGELOG's "returns finite gradients" is accurate; the docstring is where a user learns what "finite" means here.
4. nit — replace the finiteness-only assertion with a value pin of what the guard produces, `[[0.0, -0.5, 0.0, 0.0, 0.0]]` for both biquadratics (identical on 2.5.1 / 2.14, measured). It fails under a "detach everything" pseudo-fix, which the current assertion does not. One caveat, and the reason I am not asking for the "analytic" value: `solve_quartic`'s forward is discontinuous at exactly these points on `main` — perturb the `x³` coefficient of `x⁴-16` by `1e-6` and the ±2 roots vanish, perturb the `x` coefficient by `1e-6` and they move by `1.1e-2` — so no finite-difference derivative exists to pin against. That is a separate forward defect, filed as #4346; the value pin here documents a convention, and says so.
5. nit — the comments at `test_polynomial_solver.py:427-429` and `polynomial_solver.py:308-311`: add that on torch ≥ 2.14 `clamp` already zeroes the boundary gradient, which is why the pin passes on base on those legs.

**Merge notes.** Conflicts with #4338 in `TestQuarticSolver` (both append); the combined tree passes 127 on torch 2.14 and 2.5.1. #4338 first; whoever lands second rewords #4338's row-0-only comment. #4346 (the forward defect) should come after both.

dchaudhari7177:
Addressed at `cb0a2208`, rebased onto current `main` first (the CHANGELOG-adjacency conflict, resolved by re-inserting the same block under `### Bug fixes` with nothing of `main`'s deleted).

**1 — float16 manifest.** Added, but with different node IDs than quoted: ask 2 adds a third parametrization *and* the value pin adds an `expected_grad` argument, so the IDs are now `coeffs0-expected0-expected_grad0`, `coeffs1-expected1-expected_grad1` and `coeffs2-None-expected_grad2`. Three lines, tab-separated, next to the existing `test_solve_quartic` entries. I measured the third in float16 before adding its line, as asked: `[1, 0, 5, 0, 4]` is `[nan × 4]` forward and `[nan × 5]` backward there, same `zero_tol` underflow.

**2 — second sqrt site.** Added, but **not** with `(x^2+1)^2` = `[1, 0, 2, 0, 1]`. That case does not discriminate: on base it is already `[0, 0, 0, 0, 0]`, not `[nan × 5]`. With no real roots every output entry is the hard-zero placeholder, so backward never reaches `E` and the unguarded `sqrt` is not on the path. Measured on torch 2.5.0+cpu:

| coeffs | base grad | head grad |
| --- | --- | --- |
| `[1, 0, 2, 0, 1]` — `(x^2+1)^2` | `[0, 0, 0, 0, 0]` | `[0, 0, 0, 0, 0]` |
| `[1, 0, 5, 0, 4]` — `(x^2+1)(x^2+4)` | `[nan, nan, nan, nan, nan]` | `[0, 0, 0, 0, 0]` |

So I searched for cases that do reach it: 40,000 seeded integer quartics, keeping those whose base backward is non-finite, whose head backward is finite, and whose forward is `torch.equal` between the two. 214 hit, and `(x^2+1)(x^2+4)` is the smallest no-real-roots one. That is what the parametrization uses. If you would rather pin the shape you named, `[1, 0, 2, 0, 1]` can go in alongside it, but it passes on base and would not catch a regression.

The `expected` handling grew a `None` case for it, as you asked — all four roots compared to the zero placeholder rather than sorting the first two.

**3 — docstring.** Added as a second `.. note::` on `solve_quartic`, matching the wording #4299 settled on for `solve_cubic`: both `sqrt` boundaries named, finite-but-not-the-root-Jacobian stated, forward unaffected.

**4 — value pin.** Replaced the finiteness-only assertion with `[[0.0, -0.5, 0.0, 0.0, 0.0]]` for both biquadratics and `[[0.0, 0.0, 0.0, 0.0, 0.0]]` for the new case, with your caveat as the comment: convention, not Jacobian, because of the forward discontinuity in #4346.

**5 — comments.** Both sites now say that torch >= 2.14 already zeroes the boundary gradient and that the 2.5.1 / 2.9.1 legs carry the discrimination. The fallback site also names its reachable case.

**Verification** (torch 2.5.0+cpu, Python 3.12): `tests/geometry/solvers/test_polynomial_solver.py` — 174 passed at float32/float64/float16, the 9 float16 failures being the 6 already in the manifest plus the 3 new lines. Pinned `ruff` check and format clean. Two things I could not run locally: `--verify-known-failures` refuses this machine (`Python mismatch: expected '3.11', found '3.12'`), so the manifest lines are validated by CI, not by me; and `tests/test_half_precision_ci.py::test_failure_recorder_refuses_nodes_fed_by_eager_rng` fails identically on a clean `main` checkout here, so it is pre-existing and unrelated.

Merge order note still stands — #4338 first, and I will reword its row-0-only comment if this lands second.

ducha-aiki (commented):
Delta re-review at `cb0a2208` — the algorithmic and test asks are addressed; **one small documentation correction remains**.

The second `sqrt` site now has a discriminating no-real-roots pin, all three cases pin the produced gradient values, the three float16 nodes are recorded, and the public docstring explains the finite surrogate convention. Re-run at this head: the solver file passes 123 tests at float32/float64; the three focused CPU-float16 nodes all xfail under the profile as intended. GitHub has no failed checks; only the long-running float16 job remains.

At `kornia/geometry/solvers/polynomial_solver.py:263`, the new docstring contains an actual ASCII BEL control byte, so runtime inspection returns ``"R \x07pprox 0"`` instead of ``"R \\approx 0"``. I reproduced this with `inspect.getdoc(solve_quartic)` from the PR worktree. Replace the control byte with the literal `\approx` sequence; the function already uses a raw docstring. Please also keep the generated half-precision manifest in node-ID order by moving the new `test_convention...` lines before `test_random`.

After #4338 merges, rebase this branch onto it because the two PRs conflict where they append to `TestQuarticSolver`; retain #4338's float16 manifest line alongside these three.

ducha-aiki (approved):
Approved at `2b69e6a4` after the maintainer-authorized recovery.

Rebased onto main f4532f39. Corrected the BEL control byte to a valid single-backslash Sphinx math expression and sorted the added float16 manifest entries. Solver tests: 123 passed at float32/float64; local CPU float16 profile: 55 passed, 8 expected xfails. The final follow-up only corrects documentation escaping. This clears the remaining review requests. Coordinate the solver-test merge with #4338; both remain independent PRs.

ducha-aiki (approved):

PR #4340, opened 2026-09-07
fix(sensors): preserve device and dtype for scalar depth in Z1Projection.unproject

### LINKED ISSUE #4313: W1b: `Z1Projection.unproject` — a python-scalar depth builds a CPU tensor and fails on every accelerator

li-lizhe:
## Problem

`Z1Projection.unproject` documents `depth: torch.Tensor | float` and its own doctest passes a python `int` `3`. When `depth` is a python scalar the method promotes it with:

```python
depth = torch.Tensor([depth])
```

`torch.Tensor([depth])` is a **CPU float32** constructor: it ignores the device and the dtype of `points`. On CPU the mismatch is invisible; on every accelerator the very next line — `points.x * depth` — raises `RuntimeError: Expected all tensors to be on the same device`. So the documented and doctested scalar form of the API works on exactly one device.

The same line also silently widens the dtype: with a `float16` (or `bfloat16` / `float64`) `points` tensor, the depth is promoted to `float32` and the downstream computation runs in mixed precision (half-precision inputs now stay in their dtype, previously widened to float32).

## Root cause

`torch.Tensor([depth])` does not propagate `device` or `dtype` from the caller.

## Fix

Use `torch.as_tensor([depth], device=points.data.device, dtype=points.data.dtype)` so the scalar depth matches the points' device and dtype.

## Verification

- **Ascend NPU (910B, torch_npu)**: old code raises `Expected all tensors to be on the same device. Expected NPU tensor`; patched code returns NPU tensors and preserves float32/bfloat16 dtype.
- **CPU float64**: patched code preserves float64 dtype (old code widened to float32).
- **Tensor-depth path**: unchanged (not affected by the scalar branch).
- **CPU float32 path**: returns identical values (regression check).

Fixes #4313

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -352,6 +352,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug fixes
 
 
+* `Z1Projection.unproject` now materialises a python `int`/`float` `depth` on the device and
+  in the dtype of `points`; it used to build a CPU float32 tensor, which raised `RuntimeError`
+  on every accelerator and widened float16/bfloat16 results to float32. (#4340)
+
 * `kornia.io.load_image` and `write_image` work on the kornia_rs that a plain `pip install kornia`
   resolves. kornia_rs 0.1.11 moved its image readers and writers from the package root into
   `kornia_rs.io`, and kornia kept calling the root, so on 0.1.11 and newer `load_image` raised
--- a/kornia/sensors/camera/projection_model.py
+++ b/kornia/sensors/camera/projection_model.py
@@ -73,7 +73,7 @@ class Z1Projection:
 
         """
         if isinstance(depth, (float, int)):
-            depth = torch.Tensor([depth])
+            depth = torch.as_tensor([depth], device=points.data.device, dtype=points.data.dtype)
         return Vector3.from_coords(points.x * depth, points.y * depth, depth)
 
 
--- a/kornia/sensors/camera/projection_model.py
+++ b/kornia/sensors/camera/projection_model.py
@@ -59,7 +59,7 @@ class Z1Projection:
 
         Args:
             points: Vector2 representing the points to unproject.
-            depth: torch.Tensor representing the depth of the points to unproject.
+            depth: a :class:`torch.Tensor` of shape ``(B,)``, or a python scalar for a single point.
 
         Returns:
             Vector3 representing the unprojected points.
--- a/tests/sensors/camera/test_projection_model.py
+++ b/tests/sensors/camera/test_projection_model.py
@@ -83,21 +83,15 @@ class TestProjection(BaseTester):
         accelerator and widened float16/bfloat16 results to float32.
         """
         projection = Z1Projection()
-        points = Vector2(
-            torch.tensor([[0.25, 0.5]], device=device, dtype=dtype)
-        )
+        points = Vector2(torch.tensor([[0.25, 0.5]], device=device, dtype=dtype))
         expected = projection.unproject(
             points,
             torch.tensor([4.0], device=device, dtype=dtype),
         ).data
         for depth in (4, 4.0):
             out = projection.unproject(points, depth)
-            assert out.data.device.type == device.type, (
-                f"expected device {device.type}, got {out.data.device.type}"
-            )
-            assert out.data.dtype == dtype, (
-                f"expected dtype {dtype}, got {out.data.dtype}"
-            )
+            assert out.data.device.type == device.type, f"expected device {device.type}, got {out.data.device.type}"
+            assert out.data.dtype == dtype, f"expected dtype {dtype}, got {out.data.dtype}"
             self.assert_close(out.data, expected, atol=0.0, rtol=0.0)
         # Also verify a vector depth still works unchanged
         out_vec = projection.unproject(points, torch.tensor([4.0], device=device, dtype=dtype))
--- a/tests/sensors/camera/test_projection_model.py
+++ b/tests/sensors/camera/test_projection_model.py
@@ -74,3 +74,31 @@ class TestProjection(BaseTester):
             ).data,
             expected,
         )
+
+    def test_unproject_scalar_depth(self, device, dtype):
+        """Regression test: scalar depth must preserve device and dtype.
+
+        PR #4340: Z1Projection.unproject used to build a CPU float32 tensor
+        for python int/float depth, which raised RuntimeError on every
+        accelerator and widened float16/bfloat16 results to float32.
+        """
+        projection = Z1Projection()
+        points = Vector2(
+            torch.tensor([[0.25, 0.5]], device=device, dtype=dtype)
+        )
+        expected = projection.unproject(
+            points,
+            torch.tensor([4.0], device=device, dtype=dtype),
+        ).data
+        for depth in (4, 4.0):
+            out = projection.unproject(points, depth)
+            assert out.data.device.type == device.type, (
+                f"expected device {device.type}, got {out.data.device.type}"
+            )
+            assert out.data.dtype == dtype, (
+                f"expected dtype {dtype}, got {out.data.dtype}"
+            )
+            self.assert_close(out.data, expected, atol=0.0, rtol=0.0)
+        # Also verify a vector depth still works unchanged
+        out_vec = projection.unproject(points, torch.tensor([4.0], device=device, dtype=dtype))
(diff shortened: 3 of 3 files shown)


ducha-aiki:
## Review at `7232e719` by execution — CHANGES REQUESTED (the fix is right; it needs its pin and its CHANGELOG line)

**Verified.** On base, `Z1Projection().unproject(points_on_mps, 3)` raises `Expected all tensors to be on the same device ... mps:0 and cpu!` for int and float depth at float32/float16/bfloat16; at head it returns `mps:0` tensors in the points' dtype, `torch.equal` to the tensor-depth path in all 14 CPU+MPS cells, and CPU float16/bfloat16 no longer widen to float32. The tensor-depth path is untouched. The doctest passes; `torch.compile(fullgraph=True)` is clean; `pre-commit` green. The `[depth]` → `(1,)` shape is the right call: a 0-d would change the doctest repr and the `data` shape. Pre-existing and out of scope: a scalar depth with B > 1 points fails on base and head alike (`Vector3.from_coords` stacks, no broadcasting), so the pin below uses B == 1. Ascend NPU and CUDA were not exercised here; CPU and MPS were.

**CI.** The red `collector` is the float16 leg hanging for 6 h (#4214, an Intel-runner hang); the same file under #4341 passed that leg in 5 m 45 s. A rerun will clear it.

**Asks**

1. **P1 — regression test.** In `tests/sensors/camera/test_projection_model.py::TestProjection`, next to `test_unproject`: `points = Vector2(torch.tensor([[0.25, 0.5]], device=device, dtype=dtype))`, `expected = projection.unproject(points, torch.tensor([4.0], device=device, dtype=dtype)).data`, then for `depth in (4, 4.0)` assert `out.device.type == device.type`, `out.dtype == dtype`, and `self.assert_close(out, expected, atol=0.0, rtol=0.0)`. Measured: fails on `main` in 5 of 7 device/dtype cells (cpu-float16, cpu-bfloat16, the three MPS cells) and passes at head 7/7. Compare `device.type`, not `device` (`mps` vs `mps:0`).
2. **P1 — CHANGELOG.** `CHANGELOG.md` is untouched. Under `## Unreleased` / `### Bug fixes`: "`Z1Projection.unproject` now materialises a python `int`/`float` `depth` on the device and in the dtype of `points`; it used to build a CPU float32 tensor, which raised `RuntimeError` on every accelerator and widened float16/bfloat16 results to float32. (#4340)".
3. **P2 — docstring.** `projection_model.py:64` says `depth: torch.Tensor representing the depth`; the signature is `torch.Tensor | float`. Say "a `torch.Tensor` of shape `(B,)`, or a python scalar for a single point", which also records the B > 1 limitation.
4. nit — body: "half-precision inputs now stay in their dtype (previously widened to float32)" is the accurate statement; "identical values" holds for float32 only.

**Merge order, two couplings the author cannot see from the diff.**

- **#4341 contains this commit** as its first of two. Merge this one first and rebase #4341, or this PR becomes a zero-diff PR.
- **Our open docs PR #4318 pins this defect as a wart** (`test_wart_unproject_with_a_python_scalar_depth_builds_a_cpu_tensor_4313`) and documents it in two `.. warning::` blocks (`camera_model.py`, `projection_model.py`). Planted that test file onto this head: it fails in 4 cells (`torch.float16 == torch.float32` on cpu-float16/bfloat16; `DID NOT RAISE RuntimeError` on mps). That is the expected shape of a fix landing under a wart pin, and the #4318 side is ours to adjust: once this merges, #4318 drops the wart test (or inverts it into the contract test of ask 1), deletes both warning blocks, and rewords its CHANGELOG line, so that `grep -nP "#4313(?!\d)|issues/4313(?!\d)" kornia/` is empty. Nothing for you to do there beyond asks 1–4.

li-lizhe:
@ducha-aiki Thank you for the thorough review! All asks have been addressed:

1. **P1 — Regression test**: Added `test_unproject_scalar_depth` in `tests/sensors/camera/test_projection_model.py::TestProjection`. Verifies that python int/float depth preserves device type and dtype, with zero-tolerance comparison to the tensor-depth path.
2. **P1 — CHANGELOG**: Added entry under `## Unreleased` / `### Bug fixes` documenting the fix.
3. **P2 — Docstring**: Updated `depth` parameter to "a :class:\`torch.Tensor\` of shape ``(B,)``, or a python scalar for a single point."
4. **nit — Body**: Updated the PR body to say "half-precision inputs now stay in their dtype (previously widened to float32)" instead of "identical values".

Commits pushed to the `fix/z1-unproject-scalar-depth-device` branch (32401b26).

ducha-aiki (approved):
Delta re-review at `be47f837` — APPROVED.

All requested follow-ups landed: the scalar-depth regression test compares python `int` and `float` depths with the tensor path at zero tolerance, the public docstring records the single-point scalar contract, and the CHANGELOG documents the old accelerator failure and half-dtype widening with the PR number. Focused validation passes 6 float32/float64 cases plus 2 float16/bfloat16 scalar-regression cases. All 40 reported CI checks are green, and GitHub reports the branch mergeable. I found no remaining actionable issue.

Merge this before #4341, whose current head still contains this PR's original commit.

ducha-aiki (approved):

Issue #4264, opened 2026-09-05
W1b: `PinholeCamera.scale` — the result shares `extrinsics` with the source, and `scale_` / the `tx` setters write into the caller's tensors


ducha-aiki:
`PinholeCamera.scale()` is documented as returning a scaled camera, and it does clone the intrinsics — but it hands the **same** `extrinsics` tensor to the new object. Writing `scaled.tx = 7` therefore changes the source camera too.

The same aliasing runs the other way through the constructor: `PinholeCamera(K, E, h, w)` stores the caller's `K` and `E` by reference, so `scale_()` and the `tx` / `ty` / `tz` setters write into the tensors the caller passed in. `scale_`'s trailing underscore signals in-place *on the camera object*; it also mutates the caller's data. `clone()` is the only deep copy on the class.

## Repro

```python
import torch
from kornia.geometry.camera import PinholeCamera

K = torch.eye(4)[None]; K[0, 0, 0] = K[0, 1, 1] = 100.0; K[0, 0, 2], K[0, 1, 2] = 4.0, 3.0
E = torch.eye(4)[None]
h, w = torch.tensor([6]), torch.tensor([8])

# (a) scale() shares `extrinsics` with the source
cam = PinholeCamera(K.clone(), E.clone(), h.clone(), w.clone())
scaled = cam.scale(torch.tensor([0.5]))
print("scaled is cam                       :", scaled is cam)
print("scaled.extrinsics is cam.extrinsics :", scaled.extrinsics is cam.extrinsics)
print("intrinsics share storage            :", scaled.intrinsics.data_ptr() == cam.intrinsics.data_ptr())
scaled.tx = 7.0
print("source cam.tx after `scaled.tx = 7` :", cam.tx.item())

# (b) the constructor stores the caller's tensors; scale_ writes through them
Kc, Ec = K.clone(), E.clone()
c2 = PinholeCamera(Kc, Ec, torch.tensor([6.0]), torch.tensor([8.0]))
c2.scale_(0.5)
print("caller's K[0,0,2] after scale_(0.5) :", Kc[0, 0, 2].item())

# (c) the tx/ty/tz setters write into the caller's extrinsics
Kc2, Ec2 = K.clone(), E.clone()
c3 = PinholeCamera(Kc2, Ec2, h.clone(), w.clone())
c3.tx = 5.0
print("caller's E[0,:,3] after `cam.tx = 5`:", Ec2[0, :, 3].tolist())
```

## Observed

```
[...]

Issue #4266, opened 2026-09-05
W1b: `PinholeCamera` — seven shape guards that do not enforce what their messages say


ducha-aiki:
`kornia/geometry/camera/pinhole.py` has a family of shape guards whose predicates do not enforce what their own error messages promise. A validator passing is not a shape guarantee for the code behind it — the caller gets an object that constructs cleanly and then produces garbage or a `RuntimeError` from deep inside an unrelated call.

1. **`_check_valid_params` `and` / `or` slip** (`pinhole.py:64`, which even carries the author's comment *"Shouldn't this be an OR logic than AND?"*): the message says `Bx4x4` or `BxNx4x4`, the predicate accepts **any** rank-3 or rank-4 tensor. A `(1, 3, 3)` intrinsics plus a `(1, 3, 4)` extrinsics constructs, `fx` reads correctly, and `.project` returns a `(1, 1)` tensor of nonsense. `(1, 2, 2)` and `(1, 5, 5)` also pass.
2. **`_check_valid` truthiness** (`pinhole.py:57`): "Arguments shapes must match" is checked with `all(data.shape[0] for ...)`, which tests that each batch size is *non-zero*, not that they are *equal*. A batch-2 intrinsics with a batch-1 extrinsics constructs and `.project` silently broadcasts.
3. **`BxNx4x4` is advertised and unusable**: the validator's own message names it, a `(1, 2, 4, 4)` camera constructs, and `.project` raises `RuntimeError`.
4. **`PinholeCamera.project` documents `(*, 3)`** and raises `IndexError: tuple index out of range` on a `(3,)` input. The free function `project_points` raises `ValueError: Input must be at least a 2D tensor` on the same input — two exception types for one contract.
5. **`pixel2cam` / `cam2pixel` have the same `and` slip** (`if not len(depth.shape) == 4 and depth.shape[1] == 1`): the second clause is dead, so a 2-channel depth or a `(B, 3, 3)` projection matrix passes the guard and fails much later with an unrelated message.
6. **`pixel2cam`'s `intrinsics_inv` check is rank-only** (`if not len(intrinsics_inv.shape) == 3`): it never inspects the trailing `4x4` its own message promises, so a `(B, 3, 3)` inverse — the shape every free function on this surface takes — has rank 3, passes, and fails inside `transform_points` with `ValueError: Last input dimensions must differ by one unit`.
[...]

PR #4341, opened 2026-09-07
fix(geometry): promote dtype in PinholeCamera.scale_ for int64 height/width

### LINKED ISSUE #4265: W1b: `PinholeCamera.scale_` — raises `RuntimeError` on the int64 `height` / `width` the constructor accepts
The `PinholeCamera` constructor accepts `height` / `width` as int64 tensors — that is what the class's own example builds — and `scale()` handles them by promoting the result to float. Its in-place twin `scale_()` writes the float result back into the int64 tensor and raises.

So a camera built exactly as documented cannot be scaled in place, and the two twins have two different dtype policies for the same attribute.

## Repro

```python
import torch

E = torch.eye(4)[None]

def cam():  # exactly what the class docstring's own example builds: int64 height/width
    return PinholeCamera(K.clone(), E.clone(), torch.tensor([6]), torch.tensor([8]))

print("ctor accepts int64 height, dtype   :", cam().height.dtype)
print("scale(0.5).height dtype / value    :", cam().scale(torch.tensor([0.5])).height.dtype, cam().scale(torch.tensor([0.5])).height.tolist())
try:
    print("scale_(0.5).height                 :", cam().scale_(0.5).height.tolist())
except Exception as e:
    print("scale_(0.5) raises                 :", type(e).__name__ + ":", e)
f = PinholeCamera.from_parameters(torch.tensor([100.0]), torch.tensor([100.0]), torch.tensor([4.0]), torch.tensor([3.0]), 6, 8, torch.tensor([0.0]), torch.tensor([0.0]), torch.tensor([0.0]), 1, torch.device("cpu"), torch.float32)
print("from_parameters height dtype/value :", f.height.dtype, f.height.tolist())
```

## Observed

```
ctor accepts int64 height, dtype   : torch.int64
scale(0.5).height dtype / value    : torch.float32 [3.0]
scale_(0.5) raises                 : RuntimeError: result type Float can't be cast to the desired output type Long
from_parameters height dtype/value : torch.float32 [6.0]
```

## Expected

`scale_` behaves like `scale`: `self.height = self.height * scale_factor` instead of `self.height *= scale_factor`, so the attribute is rebound to the promoted float tensor rather than written into the int64 storage. For float `height` / `width` — the only inputs that work today — the result is byte-identical, so this is a focused fix that is welcome as a PR. A regression test that builds the camera with int64 sizes (as the class docstring does) and calls `scale_` fails before the fix.

Note that the fix interacts with #4264: with the tensors rebound rather than written in place, `scale_` stops mutating the caller's `height` / `width` too.

## Classification

**Broken output (crash on documented input).** Idiom 2 — a result whose dtype does not follow the input's, and two dtype policies for one attribute.

## Disposition

n/a — single symbol.

## Environment

li-lizhe:
## Problem

`PinholeCamera` constructor accepts int64 `height`/`width` — that is what the class's own example builds. `scale()` handles this by promoting the result to float. Its in-place twin `scale_()` writes the float result back into the int64 tensor and raises:

```
RuntimeError: result type Float can't be cast to the desired output type Long
```

So a camera built exactly as documented cannot be scaled in place, and the two twins have two different dtype policies for the same attribute.

## Root cause

`self.height *= scale_factor` (in-place) writes a float result into int64 storage. `self.height = self.height * scale_factor` (rebind) creates a new float tensor.

## Fix

Change `self.height *= scale_factor` → `self.height = self.height * scale_factor` (and the same for `width`), so the attribute is rebound to the promoted float tensor, matching `scale()`'s dtype policy.

**Side-effect note**: The byte-identical claim holds for values but not for side effects. `scale_` no longer mutates the `height`/`width` tensors the caller passed to the constructor; `intrinsics` are still updated in place. The only in-tree `.scale_(` caller works on a `clone()`, so nothing in kornia breaks, but this is user-visible and documented in the updated docstring. Issue #4264 stays open for the intrinsics write-through.

## Verification

- **Ascend NPU (910B, torch_npu)**: old code raises `Result type DT_FLOAT can't be cast to the desired output type DT_INT64`; patched code returns float32 height/width on NPU.
- **CPU int64**: patched code returns float32 (old code raised).
- **CPU float32/NPU float32**: unchanged (regression checks).

**Note**: This PR is stacked on #4340 (contains its commit as first of two). Once #4340 merges, this should be rebased onto `main` with only the #4341 commit.

Fixes #4265

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -402,6 +402,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   in the dtype of `points`; it used to build a CPU float32 tensor, which raised `RuntimeError`
   on every accelerator and widened float16/bfloat16 results to float32. (#4340)
 
+* `PinholeCamera.scale_` now rebinds ``height``/``width`` to the promoted result instead of
+  writing into their storage, so a camera built with int64 sizes (as the class example does) no
+  longer raises ``RuntimeError: result type Float can't be cast to the desired output type Long``
+  and matches :meth:`scale`. As a consequence ``scale_`` no longer mutates the ``height``/``width``
+  tensors the caller passed to the constructor; ``intrinsics`` are still updated in place. (#4341)
+
 * `kornia.io.load_image` and `write_image` work on the kornia_rs that a plain `pip install kornia`
   resolves. kornia_rs 0.1.11 moved its image readers and writers from the package root into
   `kornia_rs.io`, and kornia kept calling the root, so on 0.1.11 and newer `load_image` raised
--- a/kornia/geometry/camera/pinhole.py
+++ b/kornia/geometry/camera/pinhole.py
@@ -295,6 +295,10 @@ class PinholeCamera:
     def scale_(self, scale_factor: Union[float, torch.Tensor]) -> "PinholeCamera":
         r"""Scale the pinhole model in-place.
 
+        The intrinsics are updated in place while ``height`` and ``width``
+        are rebound to the floating-point scaled size, so an integer image
+        size is promoted as in :meth:`scale`.
+
         Args:
             scale_factor: a torch.Tensor with the scale factor. It has
               to be broadcastable with class members. The expected shape is
@@ -302,6 +306,11 @@ class PinholeCamera:
 
         Returns:
             the camera model with scaled parameters.
+
+        .. note::
+            As a consequence of the rebinding, ``scale_`` no longer mutates
+            the ``height``/``width`` tensors the caller passed to the
+            constructor; ``intrinsics`` are still updated in place.
 
         """
         # scale the intrinsic parameters
@@ -313,7 +322,6 @@ class PinholeCamera:
         self.height = self.height * scale_factor
         self.width = self.width * scale_factor
         return self
-
     def project(self, point_3d: torch.Tensor) -> torch.Tensor:
         r"""Project a 3d point in world coordinates onto the 2d camera plane.
 
--- a/kornia/geometry/camera/pinhole.py
+++ b/kornia/geometry/camera/pinhole.py
@@ -310,8 +310,8 @@ class PinholeCamera:
         self.intrinsics[..., 0, 2] *= scale_factor
         self.intrinsics[..., 1, 2] *= scale_factor
         # scale the image height/width
-        self.height *= scale_factor
-        self.width *= scale_factor
+        self.height = self.height * scale_factor
+        self.width = self.width * scale_factor
         return self
 
     def project(self, point_3d: torch.Tensor) -> torch.Tensor:
--- a/kornia/geometry/camera/pinhole.py
+++ b/kornia/geometry/camera/pinhole.py
@@ -322,6 +322,7 @@ class PinholeCamera:
         self.height = self.height * scale_factor
         self.width = self.width * scale_factor
         return self
+
     def project(self, point_3d: torch.Tensor) -> torch.Tensor:
         r"""Project a 3d point in world coordinates onto the 2d camera plane.
 
--- a/tests/geometry/camera/test_pinhole.py
+++ b/tests/geometry/camera/test_pinhole.py
@@ -486,6 +486,34 @@ class TestPinholeCamera(BaseTester):
         self.assert_close(pinhole_scale.height, pinhole.height * scale_val, atol=1e-4, rtol=1e-4)
         self.assert_close(pinhole_scale.width, pinhole.width * scale_val, atol=1e-4, rtol=1e-4)
 
+    def test_scale_inplace_int64_size(self, device, dtype):
+        """Regression test: scale_ must promote int64 height/width to float.
+
+        PR #4341: PinholeCamera.scale_ used to write the scaled result into
+        the int64 storage, which raised RuntimeError on every accelerator.
+        """
+        import kornia
+        batch_size = 1
+        fx, fy, cx, cy = 1.0, 2.0, 4.0, 3.0
+        tx, ty, tz = 1.0, 2.0, 3.0
+        s = torch.tensor([0.5], device=device, dtype=dtype)
+
+        intrinsics = self._create_intrinsics(batch_size, fx, fy, cx, cy, device=device, dtype=dtype)
+        extrinsics = self._create_extrinsics(batch_size, tx, ty, tz, device=device, dtype=dtype)
+        # Build camera with int64 height/width (as the class docstring example does)
+        height = torch.tensor([6], device=device)
+        width = torch.tensor([8], device=device)
+
+        cam = kornia.geometry.camera.PinholeCamera(intrinsics, extrinsics, height, width)
+        expected = cam.clone().scale(s)
+        out = cam.scale_(s)
+        assert out is cam, "scale_ must return self"
+        assert cam.height.is_floating_point(), "height must be promoted to float"
+        assert cam.height.dtype == expected.height.dtype, "height dtype must match scale()"
+        self.assert_close(cam.height, expected.height, atol=0.0, rtol=0.0)
+        self.assert_close(cam.width, expected.width, atol=0.0, rtol=0.0)
+        self.assert_close(cam.intrinsics, expected.intrinsics, atol=0.0, rtol=0.0)
+
     def test_pinhole_camera_project_and_unproject(self, device, dtype):
         batch_size = 5
         n = 2  # Point per batch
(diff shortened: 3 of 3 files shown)


ducha-aiki:
## Review at `996c2c1d` by execution — CHANGES REQUESTED (the fix is right; the side-effect change must be stated, and the PR is stacked on #4340)

**Verified.** On base, `scale_` on a camera built with int64 `height`/`width` raises `RuntimeError: result type Float can't be cast to the desired output type Long` in 14/14 device/dtype cells; at head it returns float sizes on the input device, and `scale_` and `scale` agree `torch.equal` on `height`, `width` and `intrinsics` in 28/28 cells (int64 and float sizes, tensor factor). Focused suites at head: 73 passed float32/float64; the 14 half-dtype failures are identical to `main` (0 under the known-failure profiles); MPS 36 passed. `pre-commit` green. CPU and MPS only; Ascend NPU and CUDA not exercised.

**The "byte-identical for float" claim is true for values and false for side effects.** Measured with `h = torch.tensor([480.])`, `cam = PinholeCamera(K, E, h, w)`, `cam.scale_(2.0)`: on base `cam.height is h` and the caller's `h` reads `[960.]` afterwards; at head `cam.height is h` is `False` and `h` stays `[480.]`. `intrinsics` are still written in place on both (`K[0, 0, 2]` → 8.0, `cam.intrinsics is K`). Issue #4265 itself flags this interaction with #4264. The only in-tree `.scale_(` caller works on a `clone()`, so nothing in kornia breaks, but it is user-visible and has to be written down.

**Asks**

1. **P1 — regression test.** In `tests/geometry/camera/test_pinhole.py::TestPinholeCamera`, next to `test_pinhole_camera_scale_inplace`: build the camera the way the class docstring does (intrinsics/extrinsics in `dtype`, `height = torch.tensor([6], device=device)`, `width = torch.tensor([8], device=device)` — int64, no dtype), then `expected = cam.clone().scale(s)` with `s = torch.tensor([0.5], device=device, dtype=dtype)`, `out = cam.scale_(s)`, and assert `out is cam`, `cam.height.is_floating_point()`, `cam.height.dtype == expected.height.dtype`, `assert_close` with zero tolerance on `height`, `width` and `intrinsics`. Measured: fails on `main` 7/7 cells with the issue's `RuntimeError`, passes at head 7/7.
2. **P1 — CHANGELOG**, naming the side-effect change. Under `## Unreleased` / `### Bug fixes`: "`PinholeCamera.scale_` now rebinds `height`/`width` to the promoted result instead of writing into their storage, so a camera built with int64 sizes (as the class example does) no longer raises `RuntimeError: result type Float can't be cast to the desired output type Long` and matches `scale()`. As a consequence `scale_` no longer mutates the `height`/`width` tensors the caller passed to the constructor; `intrinsics` are still updated in place. (#4341)". If the maintainer reads the aliasing change as breaking, the same sentence goes under `### Breaking changes`; either way the old behaviour is stated.
3. **P1 — PR body.** Qualify the byte-identical claim as above, and say `intrinsics` write-through is unchanged, so #4264 stays open.
4. **P2 — scope.** The head contains #4340's commit `7232e719`, so the diff also touches `kornia/sensors/camera/projection_model.py` for #4313, which #4265 does not cover. Rebase onto `main` with only your commit once #4340 merges (or say in the body that this is stacked on #4340). As it stands the two cannot both merge unchanged.
5. **P2 — docstring.** `pinhole.py:296` says only "Scale the pinhole model in-place." Add that the intrinsics are updated in place while `height`/`width` are rebound to the floating-point scaled size, so an integer image size is promoted as in `scale`.
6. nit — `scale_(0.5)` with a python float on an int64 size yields `float32` even on a float64 camera, while a tensor factor yields the factor's dtype. `scale()` has the same policy today, so not required for #4265; tying the size to `self.intrinsics.dtype` would make one policy.

**Merge order.** Our open docs PR #4294 pins this defect as a wart. Planted its `test_pinhole.py` onto this head: `test_wart_scale_inplace_rejects_integer_image_size_4265` fails in all 5 cells (`DID NOT RAISE RuntimeError`), and the height/width leg of `test_wart_scale_shares_extrinsics_with_the_source_4264` fails too (the caller's `height` now stays `[6.0]`); its extrinsics and intrinsics legs still hold. Five prose sites in `pinhole.py` at #4294's head say the opposite of this fix. That side is ours: once this merges, #4294 drops the #4265 pin, trims the #4264 pin to its surviving legs, and rewords those sentences, so that `grep -nP "#4265(?!\d)|issues/4265(?!\d)" kornia/` is empty. After merge #4265 is fully fixed; #4264 stays open with a changed shape (intrinsics write-through, sizes rebound), and #4266 / #4263 are untouched, as expected.

li-lizhe:
@ducha-aiki Thank you for the detailed review! All asks have been addressed:

1. **P1 — Regression test**: Added `test_scale_inplace_int64_size` in `tests/geometry/camera/test_pinhole.py::TestPinholeCamera`. Builds a camera with int64 height/width, verifies scale_ promotes to float, dtype matches scale(), and intrinsics are also updated correctly.
2. **P1 — CHANGELOG**: Added entry under `## Unreleased` / `### Bug fixes` documenting the fix and the side-effect change (height/width rebound, intrinsics in-place).
3. **P1 — PR body**: Updated with the side-effect qualification (scale_ no longer mutates passed height/width tensors; intrinsics still write-through; #4264 stays open).
4. **P2 — Scope**: Added note that this PR is stacked on #4340 (contains its commit as first of two). After #4340 merges, this should be rebased onto main.
5. **P2 — Docstring**: Updated `scale_` docstring to document the rebinding behavior with a .. note:: block.

Commits pushed to the `fix/pinhole-scale-inplace-dtype` branch (5f1a61c6).

ducha-aiki (commented):
Delta re-review at `cf1a5309` — the #4341 implementation is approved, but do not merge this stacked head.

The regression test, CHANGELOG disclosure, PR-body qualification, and `scale_` docstring all address the prior review. Focused scale validation passes 6 float32/float64 cases plus 2 float16/bfloat16 regression cases, and all 40 reported CI checks are green. I found no remaining code issue in the #4341 delta.

The current branch still contains #4340's original commit `7232e719` but not #4340's later regression test, docstring, or CHANGELOG commit. Merging this head first would therefore land #4340's production change without its completed coverage and documentation. Merge #4340, then rebase #4341 onto that result so this PR contains only the pinhole change; the simulated merge of the two current heads is clean.

ducha-aiki (approved):
Approved at `c6fa5511` after the maintainer-authorized recovery.

Rebased onto main f4532f39 and dropped the original scalar-depth commit already completed by merged #4340. The PR now contains only the reviewed pinhole change and its tests/docs/release note; the sensor implementation matches main. Focused int64 scale tests: 2 passed. This clears the earlier ordering hold.

ducha-aiki (changes requested):
Re-review at `c6fa5511` against current main `1af0be0e`: #4340 is merged and the earlier ordering hold is cleared, but #4294 has now introduced an integration follow-up.

[P1] Update the newly merged integer-size/aliasing pins and conventions on rebase. Applying this PR's two height/width assignment changes to current main makes both `test_wart_scale_inplace_rejects_integer_image_size_4265` and `test_wart_scale_shares_extrinsics_with_the_source_4264` fail at CPU float32; both pass on unmodified main. The former expects the old RuntimeError, while the latter still expects caller-owned height/width tensors to be mutated. Replace the integer-size wart with the successful promotion contract and adjust only the image-size aliasing assertions in the #4264 pin. Update the class, `scale`, and `scale_` convention blocks in `kornia/geometry/camera/pinhole.py`, which currently describe the old exception and partial post-state. Keep the unrelated intrinsics write-through behavior pinned.

This is required alongside resolving the current conflict and obtaining green CI. The numerical fix and its prior approval remain sound; the newly merged tests/docs need to move with it. Validation used current main and then the exact two-line implementation delta in a scratch worktree, with the import path verified.

li-lizhe:
## Addressed #4341 review (P1): rebased + updated integration tests and docstrings

Rebased onto current main (which now contains #4294 and #4340). Updated the two newly merged integration tests that were pinned to the old broken behavior:

- **`test_wart_scale_inplace_rejects_integer_image_size_4265`**: Replaced the `RuntimeError` expectation with the successful promotion contract — `scale_` now promotes int64 height/width to float, matching `scale()` exactly.
- **`test_wart_scale_shares_extrinsics_with_the_source_4264`**: Adjusted the image-size aliasing assertions to reflect that `scale_` now rebinds `height`/`width` (caller's tensors are no longer mutated) while keeping the intrinsics write-through pinned.
- Updated the class convention block in `PinholeCamera` to document the rebinding behavior.
- Updated `scale()` docstring: removed the "unlike scale_" qualifier since both methods now promote.

CI should be green now that all wart tests are updated for the fixed behavior.

PR #4342, opened 2026-09-07
fix(geometry): accept (*, 3) in axis_angle_to_rotation_matrix

### LINKED ISSUE #3955: axis_angle_to_rotation_matrix accepts only (N, 3) although its own guard says (*, 3); aa2R(R2aa(R)) fails for every rank but 3

thegoodengineer:

`axis_angle_to_rotation_matrix` raised `Input size must be a (*, 3) tensor` from its own guard, but only rank-2 `(N, 3)` actually worked. The body did `wxyz.unbind(dim=1)` and `.view(-1, 3, 3)`, so anything else got past the guard and failed three frames down with a raw torch or CPython error:

```text
aa2R(zeros(3,))      -> IndexError: Dimension out of range (expected to be in range of [-1, 0], but got 1)
aa2R(zeros(2, 5, 3)) -> ValueError: too many values to unpack (expected 3)
aa2R(zeros(1, 1, 3)) -> ValueError: not enough values to unpack (expected 3, got 1)
```

Every sibling in the module already accepts `(*, 3)`, including `rotation_matrix_to_axis_angle`, so the pair did not compose:

```text
R(3, 3)       -> R2aa (3,)       -> aa2R IndexError
R(1, 3, 3)    -> R2aa (1, 3)     -> aa2R (1, 3, 3)
R(2, 5, 3, 3) -> R2aa (2, 5, 3)  -> aa2R ValueError
```

The `(3,)` row is `rotation_matrix_to_axis_angle`'s own first doctest output, so the documented result of one function could not be fed to the function documented as its inverse.

This takes option 1 from the issue, which is the one the issue recommends: widen the implementation rather than narrow the guard. Four rank-2 assumptions go: `unbind(dim=1)` becomes `unbind(dim=-1)`, the row stack becomes `dim=-2`, the Taylor branch reshapes to the input's own leading shape instead of `view(-1, 3, 3)`, and the branch mask broadcasts with `[..., None, None]` instead of `view(-1, 1, 1)`.

This is additive. No input that works today changes its output, and shapes that used to raise now return.

Closes #3955

## What did you check?

`(N, 3)` is byte-identical. I diffed this branch against the merge base over 34 tensors: batch sizes 1, 2, 7, 64 at float32 and float64, the general branch, the Taylor branch, a batch straddling the `theta2 > 1e-6` boundary, the empty `(0, 3)` batch, and the input gradient. All 34 `torch.equal`.

New shapes behave:

```text
aa2R(zeros(3,))       -> (3, 3)
aa2R(zeros(2, 5, 3))  -> (2, 5, 3, 3)
aa2R(zeros(2, 3, 4, 3)) -> (2, 3, 4, 3, 3)
nested == flat reshaped: True    (general branch and Taylor branch)
round trip through R2aa, (2, 5, 3) float64: max err 6.9e-09
aa2R(zeros(1, 4)) still raises ValueError from the guard
```

**Pins moved with the fix.** `grep -rnE "#3955|issues/3955" kornia/ tests/` was the starting point:

- the `.. warning::` block in the docstring documenting the rank restriction is removed, and the Convention block gains the leading-batch-dimension bullet its siblings have
- `rotation_matrix_to_axis_angle`'s Convention bullet said its output "cannot be fed straight back"; it now says it can
- `test_convention_accepts_any_leading_batch_dimensions_3955` was `xfail(strict=True)` with a comment saying a fix should force the mark out. The mark is gone and the body is strengthened: a shape assertion alone would not catch a fix that flattens the leading dimensions but unbinds the wrong axis, so each rank is compared against the same data through the rank-2 path
- `test_wart_only_rank_2_input_is_accepted_3955`, which pinned the three raw unpacking errors, is deleted, since it pinned behavior that no longer exists. Its own comment says it is not a contract that those ranks keep raising
- `_runs_without_raising`, whose docstring says "Retire this helper together with the #3955 pins", is deleted, and the one surviving comment that referenced it is reworded
- `test_convention_low_angle_taylor_branch_is_rank_agnostic_3955` is added, because the two branches reshape separately and a rank fix has to hold on both sides of the mask boundary

Both new tests fail on unmodified `main`:

```text
FAILED test_convention_accepts_any_leading_batch_dimensions_3955[cpu] - IndexError: Dimension out of range
FAILED test_convention_low_angle_taylor_branch_is_rank_agnostic_3955[cpu] - ValueError: too many values to unpack
```

```text
$ pytest tests/geometry/test_conversions.py --dtype=float32,float64
============= 614 passed, 2 skipped, 14 xfailed, 48 warnings in 12.41s =============

====== 2603 passed, 50 skipped, 25 xfailed, 4 xpassed, 120 warnings in 66.11s ======

$ pytest --doctest-modules kornia/geometry/conversions.py
================================ 34 passed in 0.45s ================================

$ ruff check / ruff format on both changed files
All checks passed!
```


ducha-aiki:
## Review at `accee2ab` by execution — APPROVE after one small edit

The widening is correct and additive; the one thing in the way is the new Taylor-branch test, which is the reason the blocking `tests-mps` leg is red.

**Verified**

| Claim | Result |
| --- | --- |
| `(N, 3)` byte-identical to base | A/B over 75 seeded cases (N ∈ {0, 1, 2, 7, 64} × general / Taylor / straddling batch × CPU float32/float64/float16/bfloat16 + MPS float32): 75/75 forward `torch.equal`; gradients `torch.equal` at float32/float64/bfloat16/MPS. The 8 float16 Taylor/straddle gradient cases are NaN at identical positions on base and head — pre-existing, not this PR's. |
| New shapes | `(3,)`, `(2,5,3)`, `(2,3,4,3)`, `(1,1,3)`, `(0,3)`, `(0,5,3)` all `torch.equal` to flatten-then-reshape at all four dtypes; round trip through `rotation_matrix_to_axis_angle` at `(2,5,3)` float64 6.6e-9; the guard still rejects `(1, 4)`. `(3, 3)` was already accepted on base as three axis-angles, consistent with the siblings. |
| Pins fail on base | Planted on `main`: 2 failed for exactly the body's reasons (`IndexError` on `(3,)`, `ValueError` on `(2,5,3)`), 2 pass at head. Dropping any one of the four body changes fails both tests, so they discriminate. The strengthened test compares values against the rank-2 path, not shapes; the Taylor test uses `1e-9` (theta² ≈ 3e-18) and a straddling batch. |
| AGENTS grep for #3955 | 0 hits in `kornia/` at head; all 13 pins on `main` handled (strict xfail dropped, wart pin and `_runs_without_raising` deleted, the line-6448 comment reworded, the `.. warning::` removed, the `rotation_matrix_to_axis_angle` bullet flipped). |
| Suites | `test_conversions.py` float32+float64: 614 passed / 2 skipped / 14 xfailed vs `main` 615 / 2 / 15 — exactly −3 wart cells, +1 xfail→pass, +1 new. Half-dtype failing set identical to `main` (8 pre-existing). 34 doctests pass; `pre-commit` clean. `torch.compile(fullgraph=True)` clean on the new ranks. Callers (`epipolar/scene.py`, `imgwarp.py`) pass rank-2 and never relied on the raise. |

**Asks**

1. **P1 — `tests/geometry/test_conversions.py:2321-2337`** (`test_convention_low_angle_taylor_branch_is_rank_agnostic_3955`): the three `dtype=torch.float64` literals (lines 2328, 2333, 2334) raise `TypeError: Cannot convert a MPS Tensor to float64` on MPS, and that single test is the red `tests-mps` leg at this head (reproduced locally: 1 failed, 10 passed, 10 skipped on `--device=mps`). Take `(self, device, dtype)` and use `dtype=dtype` in all three constructors — verified 4 passed at float16/bfloat16/float32/float64 on CPU and 1 passed on MPS. The Taylor branch is still exercised at every dtype: `1e-9` gives theta² below `1e-6` everywhere.
2. **P2 — CHANGELOG** ends with `(#3955)`, the issue; the file's convention is the PR number, `(#4342)`, with the issue in prose. Also one of the two blank lines after the entry can go.
3. nit — `TestAngleAxisToRotationMatrix` has no `test_dynamo` while its sibling classes do; since the body now uses `reshape(list(...))` and `[..., None, None]`, a `test_dynamo(self, device, dtype, torch_optimizer)` on a `(2, 5, 3)` straddling input would give the rank path compile coverage. Coverage, not a fix.

**Merge notes.** #4162 also edits `conversions.py`/`test_conversions.py` (quaternion region only); merge-tree of the two heads is clean. CHANGELOG adjacency with the rest of the queue as usual. CUDA not exercised.

ducha-aiki (changes requested):
Re-review at `deb65b5b`: the implementation remains sound, but the two prior asks are still present after the rebase.

[P1] `tests/geometry/test_conversions.py::test_convention_low_angle_taylor_branch_is_rank_agnostic_3955` still hardcodes float64 three times and does not accept the dtype fixture. MPS does not support float64, so this remains the failing `tests-mps` job. Please add `dtype` to the test signature and construct all three tensors with it.

[P2] The Unreleased CHANGELOG entry still cites only issue #3955; include this PR, #4342.

The production implementation is unchanged beyond the rebase. CI currently reports the MPS failure and a pending float16 job; the prior substantive review remains valid.

thegoodengineer:
Both asks done, plus the nit. New head `620c6c23`, rebased onto `29e05314`.

**P1 — the dtype fixture.** `test_convention_low_angle_taylor_branch_is_rank_agnostic_3955` now takes `(self, device, dtype)` and builds all three tensors with it. You were right that `1e-9` keeps the Taylor branch selected everywhere: theta² is about `3e-18`, well under the `1e-6` boundary in every supported dtype, so the branch under test is still the one exercised at each. The comment says so, to stop a later reader "fixing" the literal.

Passes 4/4 for that test and 9/9 across the three `#3955` cells at float32, float64, float16 and bfloat16 on CPU. I cannot run the MPS leg locally, so that one is on CI.

**P2 — CHANGELOG.** Now ends `... only shapes that used to raise now return (closes #3955). (#4342)`. The double blank line you mentioned had already been collapsed by an earlier rebase.

**nit 3 — dynamo coverage.** Added `TestAngleAxisToRotationMatrix::test_dynamo` on the straddling `(2, 5, 3)` input, matching the sibling classes' signature. Its point is the rank path specifically, since the body now reads the input's own rank in `reshape(list(axis_angle.shape[:-1]) + [3, 3])` and `[..., None, None]`.

One honest limitation on that last one: this machine has no MSVC, so `inductor` cannot build (`InvalidCxxCompiler: Compiler: cl is not found`) and I could not run the test as CI will. I checked the substance with the eager backend instead, at float32 and float64: `0` graph breaks, `fullgraph=True` traces, and the compiled result is `torch.equal` to eager at `(2, 5, 3, 3)`. The inductor leg is CI's to confirm.

`test_conversions.py` at float32+float64: 620 passed, 2 skipped, 11 xfailed. The two half-dtype failures in that file (`test_axis_angle_to_rotation_matrix[cpu-float16]` and `[cpu-bfloat16]`) are already listed in `testing/half_precision_xfails/`, so the failing set is unchanged.

ducha-aiki (dismissed):
Re-review of exact head 620c6c2372d6adb764f6eccf68aa2e35dd0eb501. Both prior blockers are resolved: the low-angle rank pin now uses the injected dtype instead of hardcoded float64, and the release note cites #4342. The added compile case straddles the Taylor/general branch boundary at rank three and is appropriate coverage for the reshape change. I found no remaining blocker.

Local validation verified the isolated checkout import and passed the focused #3955/rank selections across CPU dtypes. All 40 hosted checks, including MPS and dynamo, are green, and GitHub reports the head mergeable.

Issue #4013, opened 2026-08-26
W1b: `validate_bbox` returns `False` while `validate_bbox3d` raises; the 2D shape/mask helpers never validate, the 3D ones always do


ducha-aiki:
The 2D and 3D validators have different contracts for the same job, and their callers follow suit: `infer_bbox_shape3d` and `bbox_to_mask3d` call `validate_bbox3d` (which raises `AssertionError`), whereas `infer_bbox_shape` and `bbox_to_mask` validate nothing (`bbox_to_mask` used to call `validate_bbox` and discard the bool — the comment at `bbox.py:216-219` records why the dead call was removed).

## Repro

```python
import torch
from kornia.geometry.bbox import validate_bbox, validate_bbox3d, infer_bbox_shape, infer_bbox_shape3d

bad2 = torch.tensor([[[0., 0.], [2., 0.], [2., 2.], [0., 3.]]])      # not a rectangle
print("validate_bbox ->", validate_bbox(bad2))
print("infer_bbox_shape ->", [t.tolist() for t in infer_bbox_shape(bad2)])

bad3 = torch.zeros(1, 8, 3); bad3[0, 1, 0] = 2.; bad3[0, 2, 0] = 3.  # not a cuboid
for fn in (validate_bbox3d, infer_bbox_shape3d):
    try:
        fn(bad3); print(fn.__name__, "-> ok")
    except AssertionError as e:
        print(fn.__name__, "-> AssertionError:", str(e)[:50])
```

## Observed

```
validate_bbox -> False
infer_bbox_shape -> [[3.0], [3.0]]                 <- garbage for a non-rectangle, no error
validate_bbox3d -> AssertionError: Boxes must have be cube, while get different widths
infer_bbox_shape3d -> AssertionError: ...
```

## Expected

One contract for both dimensionalities: validators either return a bool (and helpers call them and raise a `ValueError` with a corrective message) or raise themselves — but the same for 2D and 3D. Making `validate_bbox` raise turns a currently-returning call into an error; making `validate_bbox3d` return silently un-raises current callers. Both are behavior changes → window.

## Classification

**Semantic contract / sibling asymmetry.** Idiom 3 (a guard whose behavior does not match the contract its sibling documents); idiom 4 for `infer_bbox_shape` on a non-rectangle (garbage instead of raising).

## Disposition

[...]

PR #4343, opened 2026-09-07
fix(geometry): reject non-finite coordinates in the 3D box paths

### LINKED ISSUE #4258: NaN coordinates pass Boxes3D.from_tensor validation, and validate_bbox3d raises AssertionError on them
`Boxes3D.from_tensor(..., validate_boxes=True)` accepts NaN coordinates, and `validate_bbox3d` raises an `AssertionError` on them instead of returning `False`. This is the 3D counterpart of #4238, which #4243 fixes for the 2D `validate_bbox` / `Boxes.from_tensor` paths only.

## Reproduction

```python
import torch
from kornia.geometry.bbox import validate_bbox3d
from kornia.geometry.boxes import Boxes3D

nan = float("nan")
b = Boxes3D.from_tensor(torch.tensor([[0.0, 0.0, 0.0, nan, 4.0, 4.0]]), mode="xyzxyz", validate_boxes=True)
print(torch.isnan(b.data).any())
# tensor(True)   -- constructed, NaN in the stored vertices

hexahedron = torch.tensor([[[0., 0., 0.], [4., 0., 0.], [4., 4., 0.], [0., 4., 0.],
                            [0., 0., 4.], [4., 0., 4.], [4., 4., 4.], [0., nan, 4.]]])
validate_bbox3d(hexahedron)
# AssertionError: Boxes must have be cube, while get different heights tensor([[5., 5., 5., nan]]).
```

## Expected

Mirror #4243: `validate_bbox3d` returns `False` for non-finite coordinates (it is a predicate, so an `AssertionError` is the wrong failure mode regardless), and `Boxes3D.from_tensor(..., validate_boxes=True)` raises a `ValueError` naming the non-finite coordinates, with the `validate_boxes=False` opt-out and the export gate left as they are.

## Verification

Executed on `main` at `b327e7e6` with #4243 (`8d93dced`) applied, torch 2.14.0, Python 3.11, macOS CPU, float32. #4243 does not touch the 3D paths, so the result is the same without it.

thegoodengineer:

The 3D counterpart of #4243, which fixed the 2D `validate_bbox` / `Boxes.from_tensor` paths only.

`Boxes3D.from_tensor(..., validate_boxes=True)` accepted non-finite coordinates. An `inf` passed the positive-extent checks outright (`inf - 0 > 0`) and a `NaN` passed them because every comparison against `NaN` is `False`, so the box was constructed with non-finite vertices. `validate_bbox3d` let the `NaN` reach its `allclose` extent comparisons, which see `nan != nan` and raise:

```text
AssertionError: Boxes must have be cube, while get different heights tensor([[5., 5., 5., nan]])
```

That is a predicate raising, and naming the wrong defect. `validate_bbox3d` now returns `False` for non-finite coordinates, matching `validate_bbox`, and `from_tensor` raises the same `ValueError("Some boxes have non-finite coordinates.")` the 2D path raises. The `validate_boxes=False` opt-out and the export gate are unchanged.

The finite check sits after the `is_exporting()` gate, with the other value checks, since reading the data is exactly what that gate exists to skip.

**One thing the issue does not mention, which the fix has to handle.** `validate_bbox3d` has four internal callers, and every one of them calls it for its raise and discards the return value:

```text
kornia/geometry/bbox.py:272             infer_bbox_shape3d
kornia/geometry/bbox.py:430             bbox_to_mask3d
kornia/geometry/transform/crop3d.py:370 src_box
kornia/geometry/transform/crop3d.py:371 dst_box
```

On `main` all four raise `AssertionError` for a non-finite box, because the `allclose` comparison raises. Turning that case into a bare `False` would have made them silently accept `NaN` and return `NaN` extents. So the four sites convert the `False` themselves and keep raising, and `test_convention_consumers_still_raise_on_non_finite_coordinates_4258` pins that. The 2D fix had no equivalent problem because `validate_bbox` has no in-tree callers.

Closes #4258

## What did you check?

The new pins follow the shape of the ones #4243 added, sweeping every coordinate index against `nan`, `inf` and `-inf`, in both layouts. The `from_tensor` sweep carries a valid row alongside the invalid one; the `validate_bbox3d` pin is `B=1`, so its check is batch-wide. They fail on unmodified `main`:

```text
10 failed, 2 passed, 335 deselected
FAILED test_convention_validate_bbox3d_rejects_non_finite_coordinates_4258[cpu-float32] - AssertionError: Boxes must have be cube, while get different widths
FAILED test_convention_from_tensor_rejects_non_finite_coordinates_4258[cpu-float32-xyzxyz] - Failed: DID NOT RAISE ValueError
  (and the same for xyzxyz_plus, xyzwhd, at both dtypes)
```

The two opt-out tests pass on base as they should, since `validate_boxes=False` was never the problem.

Existing `#4013` pins (`test_wart_validate_bbox_returns_false_where_validate_bbox3d_raises_4013`, `test_wart_validate_bbox3d_checks_equal_edge_extents_not_cuboids_4013`) use finite inputs, so they are unaffected and still pass. The Convention block gains one sentence for the new `False` result.

```text
$ pytest tests/geometry/test_bbox.py tests/geometry/test_boxes.py tests/geometry/transform/test_crop3d.py --dtype=float32,float64
======================= 358 passed, 28 xfailed, 2 warnings in 14.68s =======================

============ 2610 passed, 50 skipped, 26 xfailed, 4 xpassed, 120 warnings in 75.11s ============

$ ruff check / ruff format on all changed files
All checks passed!
```


ducha-aiki:
## Review at `cea7df89` by execution — APPROVE after two small edits

A faithful 3D mirror of #4243: same `ValueError("Some boxes have non-finite coordinates.")`, same `isfinite().all()` placed first inside the `validate_boxes and not is_exporting()` gate. Every behaviour claim in the body held on base vs head.

**Verified**

| Claim | Result |
| --- | --- |
| `Boxes3D.from_tensor` accepts non-finite on base | Sweep of 108 cases (3 modes × 2 layouts × 6 coordinates × nan/inf/-inf, valid row alongside): base constructed 78 with non-finite vertices and mislabelled 30; head raises the new `ValueError` 108/108. |
| `validate_bbox3d` raises on base, returns `False` at head | 144 cases (rank-3, rank-4, B=2 variant): base `AssertionError` 144/144; head `False` 144/144. |
| The four internal callers still raise | `infer_bbox_shape3d`, `bbox_to_mask3d`, crop3d `src_box`/`dst_box` (and `crop_and_resize3d`) all still raise `AssertionError` at head. The exception **type is unchanged**; only the message moved from "Boxes must have be cube, while get different heights …" to "Boxes must have finite coordinates …". No caller outside `kornia/geometry` is affected (augmentation only uses `Boxes3D` as an annotation). |
| Export gate | `torch.export` graphs identical on base and head (68 nodes; NaN passes through under export, as intended). Under `torch.compile` the validated path gains one graph break (3 → 4), the same pattern #4243 landed in 2D; `fullgraph` already failed on base; `validate_boxes=False` stays at 0 breaks. |
| Pins fail on base | Planted on `main`: **10 failed / 2 passed** at float32+float64 (the body says 8/2 — it omits the consumers pin), all for the right reasons; 12 passed at head. |
| Suites | `test_bbox` + `test_boxes` + `test_crop3d`: float32/64 head 358 passed vs `main` 346 (+12, the new pins); half dtypes 22 failed on both with byte-identical sets; MPS 178 vs 172. `pre-commit` all hooks pass. The #4013 pins pass and no sentence in them is falsified. AGENTS grep for #4258: `kornia/` empty, `tests/` only the two new pins' own comments. CHANGELOG has no `-` lines, under `### Bug fixes`. |

**Asks**

1. **P2 — the consumers pin does not cover the two `crop3d` sites.** `test_convention_consumers_still_raise_on_non_finite_coordinates_4258` exercises `infer_bbox_shape3d` and `bbox_to_mask3d` only, although the body says all four. Verified: with the `crop3d.py` hunk reverted, the pin still passes, and `crop_by_boxes3d(vol, nan_src, dst)` / `crop_and_resize3d(vol, nan_box, size)` return all-NaN crops silently. Add `pytest.raises(AssertionError, match="finite")` for those two calls (the `src_box` case; the `dst_box` case is caught earlier by `infer_bbox_shape3d`).
2. **P2 — CHANGELOG** line 361 ends `(#4258)`, the issue; the file cites the PR number. Make it `(closes #4258). (#4343)` like the #4243 entry, and mention that the four callers' error message changed.
3. nit — body counts: 10/2, not 8/2; and "valid row alongside" is not true of the `validate_bbox3d` pin (B=1). Harmless, the check is batch-wide.
4. nit — stale prose: the `bbox_to_mask3d` Convention (`bbox.py:371`) still says `validate_bbox3d` raises for an invalid box; `from_tensor`'s `validate_boxes:` argument text and the class Convention bullet mention only positive-extent rejection (the same omission 2D has after #4243, so fix it here for 3D and, if you like, the 2D one in passing).

CUDA not exercised; the local half-dtype run compared failure sets only.

ducha-aiki (changes requested):
Re-review at `a9bfd4b9`: the implementation remains correct and all 40 CI checks are green, but the two prior review asks are unchanged after the rebase.

[P2] `tests/geometry/test_bbox.py::test_convention_consumers_still_raise_on_non_finite_coordinates_4258` still exercises only `infer_bbox_shape3d` and `bbox_to_mask3d` while claiming all four callers. Please add assertions for the two crop3d source-box paths (`crop_by_boxes3d` and `crop_and_resize3d`) that require the finite-coordinate error. The previous review demonstrated that reverting the crop3d hunk leaves the current pin green while those paths can return NaN output.

[P2] The Unreleased CHANGELOG still ends with issue #4258. Cite this PR as (#4343), and mention that the four callers keep raising `AssertionError` but their message now names non-finite coordinates.

This re-review inspected the rebase delta and current CI; the earlier execution evidence remains applicable to the unchanged implementation.

thegoodengineer:
All four addressed. New head `6f748e21`, rebased onto `29e05314`.

**P2 — the consumers pin now covers all four callers.** You were right, and I reproduced the gap before fixing it: with the `crop3d.py` hunk reverted, `crop_by_boxes3d(vol, nan_src, dst)` and `crop_and_resize3d(vol, nan_box, size)` both returned without raising, and the pin stayed green. The reason is the one you gave: `crop_by_boxes3d` sizes its output from the **dst** box, so `infer_bbox_shape3d` never sees the non-finite **src** box, and the `_crop_by_boxes3d_to_size` guard is the only thing standing between it and the sampler.

Both calls are pinned now, with the valid-box counterparts alongside. The pin discriminates the hunk: reverting `crop3d.py` alone turns it red (`Failed: DID NOT RAISE AssertionError`), and it passes with it.

One correction to my own claim while checking: I could not get a single-vertex non-finite `src_box` to produce all-NaN output on my inputs, only the missing raise. I did not chase the difference, since the raise is what the guard is for and what the pin now asserts.

**P2 — CHANGELOG.** Now `... unchanged (closes #4258). (#4343)`, and it says the four callers keep raising the same `AssertionError` with a message that now names the non-finite coordinates instead of reporting mismatched cube extents.

**nit 4 — stale prose.** Fixed in both dimensions:
- `infer_bbox_shape3d` and `bbox_to_mask3d` Convention blocks said validation raises rather than returning `False` full stop; they now scope that to shape and extent failures and name the non-finite case as the one `validate_bbox3d` reports as `False` and they raise for.
- `Boxes3D.from_tensor`'s `validate_boxes:` text and the `Boxes3D` Convention bullet mentioned only positive-extent rejection; both now lead with the non-finite rejection.
- The 2D counterparts (`Boxes.from_tensor`'s argument text and the `Boxes` Convention bullet) carried the same omission left by #4243, corrected in passing as you offered.

**nit 3 — body counts.** Corrected: the fail-on-base is 10 failed / 2 passed, not 8/2, because the consumers pin also fails there; and "valid row alongside" describes the `from_tensor` sweep, not the `validate_bbox3d` pin, which is `B=1`. The PR body is updated.

`test_bbox` + `test_boxes` + `test_crop3d` at float32+float64: 360 passed, 24 xfailed. Doctests for both modules: 15 passed. Ruff clean.

ducha-aiki (dismissed):
Re-review of exact head 6f748e211376da0e59406fbd8d356d73cb78ee20. The two missing crop3d consumer pins now exercise non-finite source boxes through crop_by_boxes3d and crop_and_resize3d, and the release note cites #4343 while describing the preserved exception type and corrected message. The additional public prose accurately records the new finite-coordinate validation. I found no remaining blocker.

Local validation verified the isolated checkout import and passed the focused #4258 selections across bbox and Boxes on CPU float32/float64: 12 passed. The author also verified that reverting the crop3d guard makes the new consumer pin fail. All 40 hosted checks are green and GitHub reports the head mergeable.

Issue #4276, opened 2026-09-05
W1b: `tilt_projection` — the forward branch returns `Pz Rᵀ` where OpenCV returns `Pz R`


ducha-aiki:
`tilt_projection(taux, tauy)` returns `Pz · Rᵀ`. OpenCV's `computeTiltProjectionMatrix` (`modules/calib3d/src/distortion_model.hpp`) returns `Pz · R`. The `return_inverse=True` branch **is** the OpenCV inverse (`inv(Pz · R)`, matching to `1.2e-07`), so the two branches of the same function are not inverses of each other: `forward @ inverse` is not the identity.

The consequence is silent and only shows up when the 13th/14th distortion coefficients (`τx`, `τy`) are non-zero:

- `distort_points` and `undistort_points` stop being inverses — a round trip is off by `43.34` px on the probe input, exactly `0.0` with `τ = 0`.
- Measured against the repo's **own** OpenCV pin (`tests/geometry/calibration/test_undistort.py::TestUndistortPoints::test_opencv_all_coeff`, whose `expected` came from `cv2.undistortPoints`): `undistort_points` reproduces it to `1.2e-04`, while `distort_points` applied to OpenCV's answer misses the original points by `610.6` px. With `τ` zeroed in both directions the same round trip closes to `6.1e-05`.
- `undistort_image` builds its map with `distort_points`, so it inherits the defect.

`undistort_points` is unaffected because it uses the inverse branch, which is correct. There is no OpenCV-valued test on `distort_points` at all, which is why this survived.

## Repro

```python
import torch
from kornia.geometry.calibration import distort_points, tilt_projection, undistort_points

taux, tauy = torch.tensor([0.1]), torch.tensor([0.2])
fwd = tilt_projection(taux, tauy)
inv = tilt_projection(taux, tauy, return_inverse=True)

# OpenCV computeTiltProjectionMatrix (calib3d/src/distortion_model.hpp): R = Ry(tauy) @ Rx(taux);
# Pz built from R's third column; the returned matrix is Pz @ R.
c = torch.cos; s = torch.sin
one, zero = torch.ones_like(taux), torch.zeros_like(taux)
Rx = torch.stack([one, zero, zero, zero, c(taux), s(taux), zero, -s(taux), c(taux)], -1).reshape(-1, 3, 3)
Ry = torch.stack([c(tauy), zero, -s(tauy), zero, one, zero, s(tauy), zero, c(tauy)], -1).reshape(-1, 3, 3)
[...]

PR #4345, opened 2026-09-07
fix(calibration): preserve tilt projection batch dimensions

### LINKED ISSUE #4324: tilt_projection flattens every leading axis to (-1, 3, 3), so distort/undistort fail on multi-axis batches once tau != 0

victorwon2001:

Preserves leading batch dimensions in `tilt_projection` instead of flattening them, keeps the tilt coefficient singleton axis explicit in both point callers, and adds the batch dimension required by `transform_points` for unbatched inverse tilt. Regression coverage includes multi-axis and singleton-axis batches, export-path four-coefficient calls, legacy unbatched `undistort_points`, and multi-axis `undistort_image`; the separate #4276 forward tilt matrix arithmetic is unchanged.

Fixes #4324

## What did you check?

Focused CPU float32/float64 calibration tests passed with 73 tests, including the existing JIT test and the new tilt/export regressions. Full pre-commit and Ruff checks passed; `ty check kornia` completed with one deprecation warning in `kornia/feature/lightglue.py` and no errors.

## How did AI help?

AI was used to implement the scoped batch-shape fix, add focused regression tests, and run the validation above. The final diff was checked for scope, and the unrelated #4276 numerical behavior was intentionally left unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -562,6 +562,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   abstract. It now has a `SuperResolutionConfig` and a `from_config` that dispatches to either
   builder family, and both builders are covered by tests. Fixes #4291. (#4335)
 
+* `tilt_projection` preserves its documented leading batch dimensions, so `distort_points`,
+  `undistort_points`, and `undistort_image` no longer fail on multi-axis batches when tilt distortion
+  is applied. (#4345)
+
 * `kornia.io.load_image` and `write_image` work on the kornia_rs that a plain `pip install kornia`
   resolves. kornia_rs 0.1.11 moved its image readers and writers from the package root into
   `kornia_rs.io`, and kornia kept calling the root, so on 0.1.11 and newer `load_image` raised
--- a/kornia/geometry/calibration/distort.py
+++ b/kornia/geometry/calibration/distort.py
@@ -35,9 +35,9 @@ def tilt_projection(taux: torch.Tensor, tauy: torch.Tensor, return_inverse: bool
           :func:`~kornia.geometry.calibration.undistort_points` applies, and it is what reproduces OpenCV's
           ``undistortPoints`` on this repository's own reference values.
         - ``return_inverse=False`` returns ``Pz @ R.T``, so the two branches are not inverses of each other.
-        - Scalar angles return :math:`(3, 3)`. For any non-scalar angle shape, all input dimensions
-          are flattened and the result is ``(taux.numel(), 3, 3)``; multiple batch axes are not preserved.
-          This is tracked as `#4324 <https://github.com/kornia/kornia/issues/4324>`_.
+        - Scalar angles return :math:`(3, 3)`. For non-scalar angles, a trailing singleton angle-component
+          axis is consumed and every leading batch dimension is preserved; without that trailing singleton,
+          the full input shape is treated as the batch shape.
 
     .. warning::
         OpenCV's ``computeTiltProjectionMatrix``, which this implementation cites, returns ``Pz @ R`` for the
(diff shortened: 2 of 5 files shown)


ducha-aiki (commented):
The batch-shape fix is correct at `118aa5e3`. `tilt_projection` now preserves every documented leading dimension, both point callers keep the coefficient axis explicit, and the unbatched inverse path is adapted to `transform_points`. The focused float32/float64 calibration run passed 73 tests. I also exported multi-axis `distort_points` and `undistort_points` through both ONNX exporters and ran all four graphs in ONNX Runtime; each returned `(2, 3, 5, 2)` and matched eager output exactly. The full CI matrix is green.

One required release artifact is missing: this changes public behavior in `kornia/geometry/calibration/distort.py:42` and fixes crashes in all three public callers, but the PR has no `CHANGELOG.md` entry. Please add a Bug fixes entry citing `#4345`. With that addition, this is ready to approve.

ducha-aiki (changes requested):
Delta re-review at `d26df006` — **CHANGES REQUESTED for current-main integration**.

The earlier release-note ask landed, and the calibration implementation remains correct. However, #4312 merged after this head and GitHub now reports conflicts in both calibration test files. That merge deliberately documented and pinned the exact #4324 behavior this PR repairs, so retaining those hunks during conflict resolution would leave the public docs claiming that tilt flattens or rejects multi-axis batches and would keep tests that expect this fixed code to fail.

Please rebase and retire every #4324 wart/prose item as part of the conflict resolution:

- `kornia/geometry/calibration/distort.py:38-40,57-59,126-129` on current `main` still says `tilt_projection` flattens leading dimensions and that `distort_points` supports at most one batch axis on the tilt/export path.
- `kornia/geometry/calibration/undistort.py:57-60,184-191` says the same for `undistort_points` and `undistort_image`.
- `tests/geometry/calibration/test_distort.py:324-329` expects `(2, 3, 1)` angles to flatten to `(6, 3, 3)`.
- `tests/geometry/calibration/test_undistort.py:400-429,477-490` expects the multi-axis and unbatched export calls repaired here to raise.

The branch's new positive tests already cover the replacement behavior, so delete or invert the newly merged wart pins rather than resolving around them. Per the repository's documented-issue gate, `git grep -n -E '#4324|issues/4324' -- kornia/ tests/` should have no surviving stale source prose when this closes #4324. Re-run the combined calibration files after the rebase; the pre-conflict head itself has 40 green checks.

ducha-aiki (approved):
Re-review of `451e69ceb50cd563d013976734077e97bb4f652b`.

The follow-up removes the obsolete #4324 public prose and failure pins while retaining positive multi-axis and export-path coverage. Complete calibration tests passed 142 CPU float32/float64 cases, with 1 skip and 2 expected xfails for the separate forward-tilt issue. No #4324 source references remain, the release note is present, current-main trial merge is clean and hosted checks pass. This is code-review clear. Coordinate with #4384 so its Pz @ R correction survives alongside this batch-shape fix; fresh integration review is needed after the first lands.

PR #4347, opened 2026-09-07
test: pin undistort_points out-of-radius two-cycle

victorwon2001:
Related to #4285

Adds a convention regression test for the out-of-radius behavior reported in #4285. The test round-trips the maintainer-verified float64 probe through `distort_points` and `undistort_points` and records the fixed-point iteration's two-cycle: iterations 5 and 7 have matching residuals, iterations 6 and 8 have matching residuals, and the even branch is larger than the odd branch while both remain invalid.

No absolute residual value is pinned, and there is no change to `undistort_points`, `undistort_image`, their public behavior, or documentation.

Related to #4285

## What did you check?

```text
$ pixi run test-module tests/geometry/calibration/test_undistort.py::TestUndistortPoints::test_convention_out_of_radius_fixed_point_is_a_two_cycle_4285 --device=cpu --dtype=float64
$ pixi run test-module tests/geometry/calibration/test_undistort.py --device=cpu --dtype=float64
```

The focused regression test and the full calibration undistortion test module both passed on CPU float64.

## How did AI help?

AI was used to inspect the relevant implementation and test conventions, add the focused regression test, run the requested checks, and inspect the final diff.

### DIFF
--- a/tests/geometry/calibration/test_undistort.py
+++ b/tests/geometry/calibration/test_undistort.py
@@ -18,6 +18,7 @@
 import pytest
 import torch
 
+from kornia.geometry.calibration.distort import distort_points
 from kornia.geometry.calibration.undistort import undistort_image, undistort_points
 
 from testing.base import BaseTester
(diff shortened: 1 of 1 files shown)


ducha-aiki (commented):
The regression itself is sound: at `6bf05986`, the new test records the stable odd/even two-cycle and passes in the focused CPU float64 run. The completed cross-platform CI jobs are green; the float16 job was cancelled at the six-hour limit and contains no production-code delta from this test-only PR.

One metadata blocker remains before merge: the first line of the PR body says `Fixes #4285`, so GitHub will close #4285 even though this PR intentionally only records the current invalid behavior and does not resolve the divergence. Please change that line to `Related to #4285` (matching the later section of the body). Once that is corrected, this is ready.

ducha-aiki (approved):
The final metadata correction is now applied: the PR body says Related to #4285, so the test-only pin will keep the unresolved distortion issue open. Code is unchanged from the reviewed head 6bf05986. The cancelled float16 check and its collector have been rerun; merge once required checks pass.

ducha-aiki (approved):

PR #4348, opened 2026-09-07
fix(depth): make depth_from_plane_equation's grazing-ray guard fire at the singularity (closes #4280)

### LINKED ISSUE #4280: W1b: `depth_from_plane_equation` — the singularity guard is a no-op exactly at the singularity

dchaudhari7177:
Closes #4280.

## The bug

`depth_from_plane_equation` guards its division by the ray-plane dot product:

```python
zero_mask = denom.abs() < eps
denom = torch.where(zero_mask, eps * torch.sign(denom), denom)
```

`torch.sign(0.) == 0.`, so at `denom == 0` -- the exact singularity the guard exists for -- the epsilon is multiplied by zero and the division still runs against zero. A ray exactly parallel to the plane returns `inf`.

The masked *shape* of the guard is right: unlike an `x + eps` it perturbs nothing outside the mask, and it already works for small non-zero denominators (`2.4e-09` is clamped to `eps` and returns `2e8`). Only the worst case was missed, and a grazing ray is not an exotic input -- it is every pixel on the horizon of a ground plane.

## The fix

`torch.copysign(full_like(denom, eps), denom)`. It has no zero hole, and it keeps the sign that the small non-zero denominators already got, so nothing outside `denom == 0` changes.

## Before and after

The repro from the issue, on CPU / float32:

```
                          main       this PR
grazing ray, denom == 0    inf     200000000.0
denom = 2.4e-09       2e8         2e8            (unchanged)
```

## Tests

Two added to `TestDepthFromPlaneEquation`:

- `test_grazing_ray_is_finite` -- the principal point's ray against a perpendicular plane normal, so the dot product is exactly zero. **Fails on main** with `AssertionError: grazing ray returned [[inf]]`.
- `test_small_denominators_keep_their_sign` -- two rays whose dot products differ only in sign must return equal-magnitude, opposite-sign depths, so the copysign branch cannot quietly collapse onto one side. Passes on main too; it is there to hold the part of the guard that already worked.

```
git stash (fix removed):
  tests/geometry/test_depth.py -k "grazing_ray_is_finite or small_denominators_keep_their_sign"
  1 failed, 1 passed

with the fix:
  same selection                      2 passed
  tests/geometry/test_depth.py       70 passed
```

`uvx ruff@0.16.4 check` and `format` clean. CHANGELOG entry added under the existing `### Bug fixes` heading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -363,8 +363,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   near-singular guard was `eps * torch.sign(denom)`, and `torch.sign` is zero at zero, so at the exact
   singularity the epsilon was multiplied away and the division still ran against zero, returning `inf`.
   A grazing ray is not an exotic input: it is every pixel on the horizon of a ground plane. The guard
-  now uses `torch.copysign`, which has no such hole and keeps the sign the small non-zero denominators
-  already got. (#4280)
+  now picks its sign with a comparison, which has no such hole and keeps the sign the small non-zero
+  denominators already got. `torch.copysign` reads the same but is not exportable, and this function
+  is in the documented ONNX export surface, so the comparison form is pinned in
+  `tests/onnx/test_export_coverage.py`. The finite-depth promise is bounded by the dtype: the default
+  `eps=1e-8` is below float16 resolution, so half-precision callers must pass a representable `eps`.
+  (#4280, #4348)
 
 * `RenderingDeFMO` (used by `DeFMO`) no longer crashes on a half-precision forward pass. Its rendering
   time-steps (`times`) were a plain Python attribute, not a registered buffer, so `nn.Module.to()` never
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -359,6 +359,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed `unproject_points_z1` depth shape handling for singleton and multi-axis batches,
   accepting both trailing-singleton and flat depth tensors. (#4355)
 
+* `depth_from_plane_equation` returns a finite depth for a ray exactly parallel to the plane. The
+  near-singular guard was `eps * torch.sign(denom)`, and `torch.sign` is zero at zero, so at the exact
+  singularity the epsilon was multiplied away and the division still ran against zero, returning `inf`.
+  A grazing ray is not an exotic input: it is every pixel on the horizon of a ground plane. The guard
+  now uses `torch.copysign`, which has no such hole and keeps the sign the small non-zero denominators
+  already got. (#4280)
+
 * `RenderingDeFMO` (used by `DeFMO`) no longer crashes on a half-precision forward pass. Its rendering
   time-steps (`times`) were a plain Python attribute, not a registered buffer, so `nn.Module.to()` never
   moved it; `forward` re-derived `times`'s device from the input on every call but never its dtype, so
--- a/kornia/geometry/depth.py
+++ b/kornia/geometry/depth.py
@@ -271,12 +271,17 @@ def depth_from_plane_equation(
     denom = torch.sum(rays * plane_normals_exp, dim=-1)  # (B, N)
     denom_abs = torch.abs(denom)
     zero_mask = denom_abs < eps
-    # copysign rather than `eps * sign(denom)`: `sign` is zero at zero, so the
+    # The guard was `eps * sign(denom)`, and `sign` is zero at zero, so the
     # multiplication cancelled the guard at the exact singularity it exists for
-    # and a ray parallel to the plane returned inf. copysign has no such hole,
-    # and keeps the branch's sign for the small non-zero denominators the guard
-    # already handled.
-    denom = torch.where(zero_mask, torch.copysign(torch.full_like(denom, eps), denom), denom)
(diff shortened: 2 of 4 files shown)


ducha-aiki (changes requested):
Changes requested at `88c24793`:

- **P1 — `kornia/geometry/depth.py:279` regresses the documented ONNX export surface.** `depth_from_plane_equation` is listed in `docs/export_support/cases_geomB.py`, and the current implementation exports through both `torch.onnx.export(..., dynamo=False)` and `dynamo=True`. At this head the same minimal wrapper fails both ways: the legacy exporter raises `UnsupportedOperatorError: aten::copysign`, while the dynamo exporter raises `ConversionError` because there is no ONNX function for `prims.signbit`. The zero-safe `torch.where(denom < 0, -1, 1)` formulation used by #4352 exports through both paths and produced `[[2e8]]` in ONNX Runtime, so please use an exportable formulation and pin export.
- **P1 — the two new tests fail the supported CPU float16 job.** `tests/geometry/test_depth.py:539` and `:556` both still return `inf`: the default `eps=1e-8` rounds to zero in float16, and the asserted `2 / 1e-8` result is also outside float16's finite range. This is the two-failure cause in job 101746040484 and reproduced locally. Either make the singular fallback policy dtype-aware if the finite-output promise is intended for half precision, or choose a representable `eps`/offset for the regression and qualify the CHANGELOG claim accordingly; the blocking half suite must remain green.
- `CHANGELOG.md:364` cites the issue (`#4280`) rather than this pull request. Change it to `#4348`.

#4352 is a duplicate of the same issue and line. Since this PR opened first and has the stronger positive/negative regression coverage, I suggest keeping this one and incorporating its sibling's ONNX-safe arithmetic.

ducha-aiki (approved):
Delta re-review at `23d51bc2` — **APPROVED**.

All three prior asks are addressed. The singular guard now uses ONNX-exportable `torch.where` arithmetic while preserving the sign of small negative denominators; the half-precision regressions choose a representable epsilon and the CHANGELOG explicitly bounds the default-epsilon promise by dtype; and the release note cites both #4280 and #4348. The new export case exercises the exact zero-denominator branch and runs the generated graph in ONNX Runtime.

Focused re-run at this exact head: the two numerical regressions passed on CPU float16, bfloat16, float32, and float64 (`8 passed`), and the grazing-ray dynamo ONNX/Runtime case passed (`1 passed`). The merge-tree against current `main` is clean and all 41 hosted checks are green.

PR #4349, opened 2026-09-07
fix(camera): clone extrinsics in PinholeCamera.scale (closes #4264)

### LINKED ISSUE #4264: W1b: `PinholeCamera.scale` — the result shares `extrinsics` with the source, and `scale_` / the `tx` setters write into the caller's tensors

dchaudhari7177:
Closes #4264 -- the half the issue scopes as "a focused fix that is welcome as a PR".

## The bug

`scale()` clones the intrinsics and then hands `self.extrinsics` to the new camera by reference:

```python
return PinholeCamera(intrinsics, self.extrinsics, height, width)
```

The constructor stores what it is given, so the returned camera shares translation state with the source:

```
scaled is cam                       : False
scaled.extrinsics is cam.extrinsics : True     <- the bug
intrinsics share storage            : False
source cam.tx after `scaled.tx = 7` : 7.0      <- the source moved
```

## The fix

One `.clone()`, for the same reason the intrinsics already have one. The returned camera's numbers are byte-identical to today's, so nothing downstream changes.

## Scope

Only `scale()`. The issue's second half -- the constructor storing the caller's `K` / `E` by reference, so `scale_()` and the `tx` / `ty` / `tz` setters write through them -- is a change to the aliasing contract of every `PinholeCamera(...)` call site, and the issue puts that choice in the coordinated Repair and Deprecation window. Batch 5 documents it as it is. Untouched here.

## Tests

`test_pinhole_camera_scale_does_not_alias_the_source` asserts the two tensors are distinct objects with distinct storage, then mutates the scaled camera and asserts the source is unchanged.

```
without the fix:  1 failed   (assert on the source camera's tx)
with the fix:     tests/geometry/camera/test_pinhole.py -k scale   3 passed
                  tests/geometry/camera/test_pinhole.py           29 passed, 1 skipped
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -436,6 +436,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   step: a rejected entry is moved aside, the next source is tried, and the discarded source is
   re-fetched once. Without `validate` the behaviour is unchanged. (#4309, #4332)
 
+* `PinholeCamera.scale` no longer shares its `extrinsics` tensor with the source camera. The
+  intrinsics were cloned, the extrinsics were handed to the new object by reference, and the
+  constructor stores what it is given, so `scaled.tx = 7` moved the camera `scale()` was called on.
+  The returned camera's numbers are unchanged. (#4264, #4349)
+
 * `RenderingDeFMO` (used by `DeFMO`) no longer crashes on a half-precision forward pass. Its rendering
   time-steps (`times`) were a plain Python attribute, not a registered buffer, so `nn.Module.to()` never
   moved it; `forward` re-derived `times`'s device from the input on every call but never its dtype, so
--- a/kornia/geometry/camera/pinhole.py
+++ b/kornia/geometry/camera/pinhole.py
@@ -49,14 +49,16 @@ class PinholeCamera:
           :func:`~kornia.geometry.camera.perspective.unproject_points` and the ``normalize_points`` flags of
           :func:`~kornia.geometry.depth.depth_to_3d` and :func:`~kornia.geometry.depth.depth_to_3d_v2` read it
           as the Euclidean ray length instead, so the unprojected point has that norm rather than that ``z``.
-        - the class stores the tensors it is constructed from instead of copying them: :meth:`scale` returns a
-          new camera that **shares** ``extrinsics`` with the source, and :meth:`scale_` and the ``tx`` / ``ty``
-          / ``tz`` setters write into the caller's tensors. :meth:`clone` is the only deep copy.
+        - the class stores the tensors it is constructed from instead of copying them, so :meth:`scale_` and
+          the ``tx`` / ``ty`` / ``tz`` setters write into the caller's tensors. :meth:`scale` is the exception:
+          it returns a new camera that owns both its ``intrinsics`` and its ``extrinsics``. :meth:`clone` is
+          the deep copy of an existing camera.
 
     .. warning::
         :meth:`scale` and :meth:`scale_` rescale the principal point as ``cx' = s * cx`` — the half-pixel rule —
         which disagrees with the integer pixel centres above; it is tracked as a coordinated repair in
-        `#4263 <https://github.com/kornia/kornia/issues/4263>`_. The shared storage is
+        `#4263 <https://github.com/kornia/kornia/issues/4263>`_. The write-through to the caller's tensors
+        that remains on :meth:`scale_` and the setters is
         `#4264 <https://github.com/kornia/kornia/issues/4264>`_, the in-place :meth:`scale_` failure on an
         integer ``height`` / ``width`` with a floating-point scale factor
         `#4265 <https://github.com/kornia/kornia/issues/4265>`_, the shape
@@ -321,16 +323,15 @@ class PinholeCamera:
         Convention:
             - returns a **new** camera whose focal lengths, principal point and image size are multiplied by
               ``scale_factor``: ``fx' = s * fx`` and ``cx' = s * cx``, the half-pixel rule.
-            - the new camera **shares** its ``extrinsics`` tensor with the source, so writing ``tx`` / ``ty`` /
-              ``tz`` on either camera moves both; the intrinsics are cloned. :meth:`clone` is the deep copy.
+            - the new camera owns its ``intrinsics`` and its ``extrinsics``: both are cloned, so writing
+              ``tx`` / ``ty`` / ``tz`` on the returned camera leaves the source where it was.
             - with a floating-point ``scale_factor``, an integer ``height`` / ``width`` is promoted to floating
               point, unlike :meth:`scale_`. An integer factor preserves the integer image-size dtype.
 
         .. warning::
             The ``cx' = s * cx`` rule disagrees with the integer pixel centres the rest of the library
             enumerates; it is tracked as a coordinated repair in
-            `#4263 <https://github.com/kornia/kornia/issues/4263>`_. The shared ``extrinsics`` are
-            `#4264 <https://github.com/kornia/kornia/issues/4264>`_.
+            `#4263 <https://github.com/kornia/kornia/issues/4263>`_.
 
         Args:
             scale_factor: a torch.Tensor with the scale factor. It has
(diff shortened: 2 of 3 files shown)


ducha-aiki (commented):
The code change is correct at `e832ae80`: `scale()` now gives the returned camera independent extrinsics, the new mutation regression fails on the merge-base and passes at the head, and the focused float32/float64 run passed 2/2. The completed cross-platform CI jobs are green; the float16 job hit the existing six-hour timeout.

Please update the release-note citation at `CHANGELOG.md:362` from the issue number `(#4264)` to this pull request, `(#4349)`, as required for changelog entries. With that mechanical edit, this is ready to approve.

ducha-aiki (approved):
Approved at `eec3cb28` after the maintainer-authorized recovery.

Rebased onto main f4532f39, preserving the extrinsics-clone fix, and added PR #4349 to the CHANGELOG citation. Focused aliasing regression: 1 passed. This clears the remaining review request.

ducha-aiki (changes requested):
Re-review at `eec3cb28` against current main `1af0be0e`: the clone fix remains correct, but there is a new integration blocker after #4294 merged.

[P1] Update the newly landed aliasing pin and conventions before merging. Applying this PR's implementation hunk to current main makes `TestPinholeCamera::test_wart_scale_shares_extrinsics_with_the_source_4264[cpu-float32]` fail at `tests/geometry/camera/test_pinhole.py:756` (`scaled.extrinsics is cam.extrinsics`). The same test passes on unmodified main. The class and `scale` docstrings in `kornia/geometry/camera/pinhole.py` now also explicitly promise shared extrinsics and describe #4264 as unresolved. Rebase onto current main, invert the scale-aliasing part of that pin to verify isolation, and update those sentences/warnings. Preserve the separate constructor/setter/in-place-intrinsics behavior that this patch does not change.

The existing head's green CI predates these main additions and therefore does not verify this combination. This is a merge-integration follow-up, not a rejection of the clone arithmetic. Validation: current-main CPU float32 wart pins pass; current main plus this implementation hunk reproduces the one failure above, using the primary Python environment with the worktree import path verified.

dchaudhari7177:
Addressed, rebased onto current `main` (`8f292308`).

**The integration blocker.** You were right — #4294 landed a wart pin and a set of docstring sentences that describe the aliasing as the contract, so applying this patch to current `main` made `test_wart_scale_shares_extrinsics_with_the_source_4264` fail at the `scaled.extrinsics is cam.extrinsics` assertion.

That pin covers four legs of #4264 in one test, and this patch only repairs one of them. So rather than delete it, I split it along that line:

- The `scale()` leg — the part this patch fixes — is removed from the pin and is now asserted the other way up in `test_pinhole_camera_scale_does_not_alias_the_source`.
- The other three legs are untouched and still pinned: the constructor storing the caller's `intrinsics` / `extrinsics` by reference, the `tx` setter writing into the caller's extrinsics, and in-place `scale_` rewriting the caller's intrinsics and image size. The test is renamed to `test_wart_constructor_and_scale_inplace_write_through_to_the_caller_4264` so the name matches what is left, and its comment points at the regression test for the repaired leg.

**Isolation check, as requested.** Reverting only the implementation hunk (`self.extrinsics.clone()` back to `self.extrinsics`):

```
1 failed, 1 passed
FAILED tests/geometry/camera/test_pinhole.py::TestPinholeCamera::test_pinhole_camera_scale_does_not_alias_the_source[cpu-float32]
```

The retained wart pin passes either way, which is the point — it no longer depends on the aliasing. With the fix in place both pass.

**Docstrings.** Three sentences promised shared extrinsics and are now corrected, and the two `#4264` citations that referred to `scale()` are re-scoped rather than dropped:

- the class Convention block: the store-by-reference sentence now names `scale_` and the setters as the write-through paths, with `scale()` called out as the exception that owns both its tensors.
- the class warning: "The shared storage is #4264" is now "The write-through to the caller's tensors that remains on `scale_` and the setters is #4264" — the issue is not fully closed by this patch and the docs should not imply it is.
- `scale()`'s own Convention block and warning: the shares/`#4264` sentence is gone, replaced by the cloning statement. `scale_`'s warning keeps its `#4264` citation, which is still accurate.

**Verification on `a19a84c3`.** Full `tests/geometry/camera/test_pinhole.py`: 47 passed / 1 skipped on both `--dtype float32` and `--dtype float64` (cpu). `ruff@0.16.6` check and format (the version pinned in `.pre-commit-config.yaml`): clean. CHANGELOG entry cites `(#4264, #4349)`.

ducha-aiki (approved):
Approved at `a19a84c3`. The current head resolves the integration request from the previous review: `scale()` now has independent intrinsics and extrinsics, the repaired leg is covered by the positive regression, and the remaining constructor/setter/`scale_` write-through behavior stays explicitly pinned under #4264. The class and method conventions now describe that split accurately.

Validation: the full pinhole camera file passes on CPU float32/float64 (93 passed, 1 skipped), and the repaired alias test plus retained wart test pass across float16, bfloat16, float32 and float64 (8 passed). An independent mutation probe confirmed that setting `scaled.tx` leaves the source pose unchanged. Hosted CI is green and the head integrates cleanly with current main `8f292308`. The import was verified inside the isolated checkout.

ducha-aiki (approved):

PR #4350, opened 2026-09-07
fix(subpix): make iterative_quad_interp3d's max_candidates cap per image (closes #4256)

### LINKED ISSUE #4256: iterative_quad_interp3d: max_candidates cap is shared across the batch, so one image's refinement depends on its batch-mates
## Summary

`iterative_quad_interp3d(..., max_candidates=K)` (and therefore `IterativeQuadInterp3d` / `AdaptiveQuadInterp3d` in `patch` mode) applies the candidate cap across the **whole batch**, not per image. The `topk` at `kornia/geometry/subpix/spatial_soft_argmax.py` ranks the flattened `(B*C)` candidate list:

```python
if max_candidates is not None and N > max_candidates:
    cand_vals = inp[bc_idx, d_idx, h_idx, w_idx]  # (N,) over every image in the batch
    _, keep = torch.topk(cand_vals, k=max_candidates)
```

So an image's refined keypoints depend on which other images share its batch. A quiet image next to a high-contrast one gets zero refined candidates. The docstring ("only the top-`max_candidates` NMS maxima ... are processed") reads as a per-image budget, and `ScaleSpaceDetector` treats `num_features` per image, so the shared cap is a defect rather than a documented contract.

## Reproduction (main `1eea5835`, torch 2.14.0, CPU)

```python
import torch
from kornia.geometry.subpix.spatial_soft_argmax import iterative_quad_interp3d

torch.manual_seed(0)
x = torch.rand(2, 1, 5, 40, 40)
x[1] *= 10                                   # image 1 outranks every image-0 candidate
cb, vb = iterative_quad_interp3d(x, max_candidates=50)
c0, v0 = iterative_quad_interp3d(x[:1], max_candidates=50)
c1, v1 = iterative_quad_interp3d(x[1:], max_candidates=50)
print(torch.equal(cb[:1], c0), torch.equal(vb[:1], v0))   # False False
print(torch.equal(cb[1:], c1), torch.equal(vb[1:], v1))   # True  True
```

Image 0 is refined at 147 positions on its own and at 0 positions inside the batch; image 1 is unaffected.

## Why it matters now

#4254 batches the positive and negative DoG refinement into one `(2B)` call. With today's semantics that call is unsafe when a cap is set, because `-response` competes with `+response` for the same `K` slots, so the PR keeps two separate calls whenever `max_candidates` is not `None` and its equivalence test skips the capped cells. That is the right workaround for the current function, but it leaves the configuration the docstring recommends for CPU (`max_candidates = num_features * 5`) on the slow path.

## Proposed fix

Rank within each `bc` segment: e.g. sort candidates by `(bc_idx, -value)` and keep `rank_within_segment < max_candidates`, or `topk` on a `(BC, N_max)` padded matrix with `-inf` fill. Once the cap is per image, the stacked call in `ScaleSpaceDetector._refine_minmax` is exact for every built-in refiner with or without a cap, and the `max_candidates` clause there can be deleted. A pin: batched output for image `i` equals the single-image output for every `i`, with a batch where one image outranks the other.

Found while reviewing #4254.

dchaudhari7177:
Closes #4256.

## The bug

`max_candidates` ranked the flattened `(B*C)` candidate list:

```python
cand_vals = inp[bc_idx, d_idx, h_idx, w_idx]  # (N,) over every image in the batch
_, keep = torch.topk(cand_vals, k=max_candidates)
```

so the cap was shared across the batch and an image's refined keypoints depended on which other images it travelled with. The issue's repro, on main:

```
image 0 batched == alone : False
```

`x[1] *= 10` makes every candidate in image 1 outrank every candidate in image 0, and image 0 comes back unrefined. This reaches `IterativeQuadInterp3d` and `AdaptiveQuadInterp3d` in `patch` mode too.

## The fix

Rank within each row: sort by response, then stably by row, so each row keeps descending-response order, and a positional rank inside the row selects the top `max_candidates` of that image. That is the same ranking the global `topk` applied, taken one image at a time.

`N <= max_candidates` remains the untouched fast path, because if the whole batch is under the bound no single row can be over it.

## Why this is a defect rather than a documented contract

The docstring reads as a per-image budget ("only the top-`max_candidates` NMS maxima ... are processed"), and `ScaleSpaceDetector` treats `num_features` per image.

## Tests

Two, in a new `TestIterativeQuadInterp3dMaxCandidates`:

- `test_cap_is_per_image_not_per_batch` -- the issue's repro as an assertion: each image refined alone must equal that image refined in the batch. **Fails on main** (`assert False`).
- `test_cap_still_bounds_the_work_per_image` -- a per-image cap must still cap. Counts positions whose coordinates moved off the grid that `max_candidates=0` leaves, and asserts at most K per image. Passes on main too; it is there so the fix cannot become "no cap at all".

```
without the fix:   1 failed, 1 passed
with the fix:
  tests/geometry/subpix/test_iterative_quad_interp_pyhesaff.py    15 passed
  tests/geometry/subpix/                                         125 passed
  tests/feature/test_scale_space_detector.py                      86 passed, 6 skipped
```

The new path is sort-based and index-only, so it stays traceable for `torch.compile`; I have not measured the AMD dynamo leg.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -359,6 +359,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed `unproject_points_z1` depth shape handling for singleton and multi-axis batches,
   accepting both trailing-singleton and flat depth tensors. (#4355)
 
+* `iterative_quad_interp3d`'s `max_candidates` cap is now a per-image budget rather than one shared
+  across the batch. The `topk` ranked the flattened `(B*C)` candidate list, so an image's refined
+  keypoints depended on which other images shared its batch: a quiet image next to a high-contrast
+  one got none. This also reaches `IterativeQuadInterp3d` and `AdaptiveQuadInterp3d` in `patch` mode.
+  The docstring already read as a per-image budget. A negative `max_candidates` now raises
+  `ValueError`; it previously raised a `topk` `RuntimeError`, and under the per-image ranking it
+  would otherwise have silently disabled refinement instead. (#4256, #4350)
+
 * `RenderingDeFMO` (used by `DeFMO`) no longer crashes on a half-precision forward pass. Its rendering
   time-steps (`times`) were a plain Python attribute, not a registered buffer, so `nn.Module.to()` never
   moved it; `forward` re-derived `times`'s device from the input on every call but never its dtype, so
--- a/kornia/geometry/subpix/spatial_soft_argmax.py
+++ b/kornia/geometry/subpix/spatial_soft_argmax.py
@@ -1108,14 +1108,26 @@ def iterative_quad_interp3d(
     # few hundred features are ultimately needed.  The per-candidate patch gather
     # (random memory access into a multi-MB volume) is cache-miss dominated on CPU;
     # reducing N here gives a proportional speedup of the iteration loop below.
+    # The cap is per image, not over the flattened batch: a global topk would make
+    # one image's refined keypoints depend on which other images share its batch,
+    # so a quiet image next to a high-contrast one would get none. N <= the cap is
+    # the fast path, because then no row can be over it either.
     if max_candidates is not None and N > max_candidates:
         cand_vals = inp[bc_idx, d_idx, h_idx, w_idx]  # (N,) pre-refinement responses
-        _, keep = torch.topk(cand_vals, k=max_candidates)
+        # Sort by response, then stably by row: within each row the candidates
+        # stay in descending-response order, so a positional rank inside the row
+        # is the same ranking the global topk used, taken one image at a time.
+        by_value = torch.argsort(cand_vals, descending=True, stable=True)
+        grouped = by_value[torch.argsort(bc_idx[by_value], stable=True)]
+        counts = torch.bincount(bc_idx, minlength=B * C)
+        row_start = torch.cumsum(counts, 0) - counts
+        rank = torch.arange(N, device=device) - row_start[bc_idx[grouped]]
+        keep = grouped[rank < max_candidates]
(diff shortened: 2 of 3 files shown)


ducha-aiki (changes requested):
Review at `f105c84c` — the per-row ranking is correct and the issue repro is fixed; **two small edits remain**.

I re-ran the focused subpixel file (15 passed) and an independent two-image/two-channel probe. On base, the quiet image differs when batched; at this head, both images are bitwise equal to their single-image calls for coordinates and values. All 40 GitHub checks are green and the head is `MERGEABLE`.

At `kornia/geometry/subpix/spatial_soft_argmax.py:1115-1125`, a negative `max_candidates` changed from an immediate `torch.topk` `RuntimeError` on base to a silent no-op: `rank < -1` selects nothing and the function returns every candidate unrefined. Reject negative values explicitly (preferably `ValueError`) and pin that exception so an invalid top-K budget cannot silently disable refinement.

The CHANGELOG entry at line 363 cites only issue #4256. Per the repository CHANGELOG contract, add this PR number too: `(#4256, #4350)`.

ducha-aiki (approved):
Delta re-review at `fac4dbab` — **APPROVED**.

Both requested edits landed. Negative `max_candidates` values now raise a targeted `ValueError` before ranking, with coverage for `-1` and `-50`; the docstring defines zero and negative behavior; and the CHANGELOG cites both #4256 and #4350 while explaining the compatibility change from the old `topk` error.

The full max-candidates regression group passed at this exact head (`4 passed`), including the per-image isolation and work-cap pins. The merge-tree against current `main` is clean and all 40 hosted checks are green.

PR #4351, opened 2026-09-07
fix(bbox): reject rank-4 input in infer_bbox_shape3d and bbox_to_mask3d (closes #4248)

### LINKED ISSUE #4248: Bug: `infer_bbox_shape3d` / `bbox_to_mask3d` accept `(B, N, 8, 3)` through `validate_bbox3d` and then crash or return wrong-shaped values

dchaudhari7177:
Closes #4248.

## The bug

`validate_bbox3d` documents and accepts rank-4 `(B, N, 8, 3)`, reshaping to `(B*N, 8, 3)` internally. Neither caller reshapes, and both index dim 1 as the vertex axis:

```python
# infer_bbox_shape3d
torch.index_select(boxes, 1, torch.tensor([1, 2, 5, 6]))
# bbox_to_mask3d
boxes[:, 4, 2]
```

so the box axis is read as the vertices. On main:

```
n=1  validate: True   infer -> IndexError      mask -> IndexError
n=3  validate: True   infer -> IndexError      mask -> IndexError
n=8  validate: True   infer -> three (1, 3) tensors, one value per coordinate
                      mask  -> RuntimeError
```

The eight-box case is the bad one: it returns rather than raising.

## The fix

`KORNIA_CHECK_SHAPE(boxes, ["N", "8", "3"])` / `["B", "8", "3"]` at the top of each caller, which is what #4218 did for the 2D twin of this (#4180).

**Added, not substituted.** #4218 could drop `validate_bbox` because its boolean was discarded and it never raised. `validate_bbox3d` is not that: it raises when the four edges parallel to an axis disagree, so it stays and the shape gate goes in front of it.

`validate_bbox3d` itself is unchanged and still accepts rank 4, since it reshapes correctly. The docstrings now say where the rejection lives and why.

## Tests

The repository already pinned both halves of this: a wart test asserting the broken behaviour, and a `strict=True` xfail convention test asserting the behaviour it should have. The fix flips both, so they collapse into the one convention test, which keeps the `validate_bbox3d(boxes) is True` assertion since that half of the mismatch is deliberate.

```
tests/geometry/test_bbox.py -k 4248                              2 passed
tests/geometry/test_bbox.py                                     47 passed, 1 xfailed
tests/geometry/test_bbox.py + tests/geometry/test_boxes.py     163 passed, 12 xfailed
pytest --doctest-modules kornia/geometry/bbox.py                  7 passed
```

`grep -rnE "#4248|issues/4248" kornia/` is now empty -- the three docstring references described the breakage and have been rewritten to describe the rejection instead, including the one in `boxes.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -359,6 +359,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed `unproject_points_z1` depth shape handling for singleton and multi-axis batches,
   accepting both trailing-singleton and flat depth tensors. (#4355)
 
+* `infer_bbox_shape3d` and `bbox_to_mask3d` reject rank-4 `(B, N, 8, 3)` input with a `ShapeError`,
+  the way `infer_bbox_shape` and `bbox_to_mask` have since #4218. `validate_bbox3d` accepts the rank-4
+  form and reshapes internally, but both callers index dim 1 as the vertex axis, so the box axis was
+  read as the vertices: an out-of-bounds error with one box, and three `(1, 3)` tensors -- one value
+  per coordinate rather than per box -- with eight. (#4248, #4351)
+
 * `RenderingDeFMO` (used by `DeFMO`) no longer crashes on a half-precision forward pass. Its rendering
   time-steps (`times`) were a plain Python attribute, not a registered buffer, so `nn.Module.to()` never
   moved it; `forward` re-derived `times`'s device from the input on every call but never its dtype, so
--- a/kornia/geometry/bbox.py
+++ b/kornia/geometry/bbox.py
@@ -99,7 +99,7 @@ def validate_bbox(boxes: torch.Tensor) -> bool:
 
 
 def validate_bbox3d(boxes: torch.Tensor) -> bool:
-    """Validate that a 3D box has equal inclusive edge extents along each axis, raising when it does not.
+    r"""Validate that a 3D box has equal inclusive edge extents along each axis, raising when it does not.
 
     Convention:
         Vertices use inclusive coordinates in the order front-top-left, front-top-right, front-bottom-right,
@@ -356,7 +356,7 @@ def bbox_to_mask(boxes: torch.Tensor, width: int, height: int) -> torch.Tensor:
 
 
 def bbox_to_mask3d(boxes: torch.Tensor, size: tuple[int, int, int]) -> torch.Tensor:
-    """Convert 3D bounding boxes to masks. Covered area is 1. and the remaining is 0.
+    r"""Convert 3D bounding boxes to masks. Covered area is 1. and the remaining is 0.
 
     Convention:
         ``size`` is ``(depth, height, width)`` and the mask comes back as :math:`(B, 1, depth, height, width)` in
--- a/kornia/geometry/bbox.py
+++ b/kornia/geometry/bbox.py
@@ -113,9 +113,9 @@ def validate_bbox3d(boxes: torch.Tensor) -> bool:
 
     .. warning::
         :func:`validate_bbox` returns ``False`` where this function raises; that inconsistency is tracked in
-        `#4013 <https://github.com/kornia/kornia/issues/4013>`_. Rank-4 input passes this check but breaks
-        :func:`infer_bbox_shape3d` and :func:`bbox_to_mask3d`, tracked in
-        `#4248 <https://github.com/kornia/kornia/issues/4248>`_.
+        `#4013 <https://github.com/kornia/kornia/issues/4013>`_. Rank-4 input passes this check, but
+        :func:`infer_bbox_shape3d` and :func:`bbox_to_mask3d` reject it with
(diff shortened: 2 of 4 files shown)


ducha-aiki (commented):
Review at `9107b38b` — **APPROVE after one CHANGELOG edit**.

The shape gates are in the right layer: `validate_bbox3d` keeps its documented rank-4 behavior, while the two callers that index dimension 1 as the vertex axis now reject rank 4 before that indexing. An independent A/B probe reproduced the old one-box exceptions and the eight-box `(1, 3)` wrong result; both callers now raise `ShapeError` for both cardinalities. The focused file passes 89 tests at float32/float64 (2 expected xfails), the `#4248` source prose is cleared while the test pin remains, all 40 GitHub checks are green, and the head is `MERGEABLE`.

The CHANGELOG entry at line 363 cites only issue #4248. Add the PR number as required by the repository contract: `(#4248, #4351)`. No code changes are otherwise needed.

ducha-aiki (approved):
Delta re-review at `849e9fcb` — **APPROVED**.

The only remaining ask is resolved: the Bug fixes entry now cites both #4248 and #4351. The code and regression are unchanged in substance, and the focused rank-4 rejection pin passed at float32 and float64 (`4 passed`). The merge-tree against current `main` is clean and all 40 hosted checks are green.

PR #4352, opened 2026-09-07
fix(geometry): guard denom == 0 in depth_from_plane_equation to prevent infinite depth

### LINKED ISSUE #4280: W1b: `depth_from_plane_equation` — the singularity guard is a no-op exactly at the singularity

thegoodengineer:

\depth_from_plane_equation\ guarded its division by the ray-plane dot product with:
\\\python
denom_abs = torch.abs(denom)
zero_mask = denom_abs < eps
denom = torch.where(zero_mask, eps * torch.sign(denom), denom)
\\\
When \denom == 0\ (for instance, a grazing ray perpendicular to the plane normal), \	orch.sign(0.) == 0.\, multiplying \ps\ by zero and causing the guard to be a no-op. The division then produced \inf\.

This change uses a zero-safe fallback for \sign\ so that \denom == 0\ is clamped to \ps\ (or \-eps\ if negative), matching other values within \(-eps, eps)\ and ensuring the computed depth remains finite.

Closes #4280

## What did you check?

- Added a regression test \	est_grazing_ray_singularity\ in \	ests/geometry/test_depth.py\ checking that grazing rays with \denom == 0\ produce finite depth values matching \plane_offsets / eps\.
- Ran \pytest tests/geometry/test_depth.py::TestDepthFromPlaneEquation --device=cpu --dtype=float32,float64\ (all 19 tests pass).
- Ran \
uff check\ on modified files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -356,6 +356,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `depth_from_plane_equation` now guards rays perpendicular to the plane normal (`denom == 0`). The
+  previous guard multiplied `eps` by `torch.sign(denom)`, which evaluates to `0.0` at `denom == 0` and
+  returned `inf`. The guard now uses a zero-safe sign fallback so grazing rays are clamped to a finite
+  depth. (#4280)
 * `RenderingDeFMO` (used by `DeFMO`) no longer crashes on a half-precision forward pass. Its rendering
   time-steps (`times`) were a plain Python attribute, not a registered buffer, so `nn.Module.to()` never
   moved it; `forward` re-derived `times`'s device from the input on every call but never its dtype, so
--- a/kornia/geometry/depth.py
+++ b/kornia/geometry/depth.py
@@ -271,7 +271,8 @@ def depth_from_plane_equation(
     denom = torch.sum(rays * plane_normals_exp, dim=-1)  # (B, N)
     denom_abs = torch.abs(denom)
     zero_mask = denom_abs < eps
-    denom = torch.where(zero_mask, eps * torch.sign(denom), denom)
+    sign = torch.where(denom < 0, denom.new_tensor(-1.0), denom.new_tensor(1.0))
+    denom = torch.where(zero_mask, eps * sign, denom)
 
     # Compute depth from plane equation
     depth = plane_offsets / denom  # plane_offsets: (B, 1), denom: (B, N) -> depth: (B, N)
--- a/tests/geometry/test_depth.py
+++ b/tests/geometry/test_depth.py
@@ -534,3 +534,13 @@ class TestDepthFromPlaneEquation(BaseTester):
             eps=1e-6,
             atol=1e-4,
         )
+
+    def test_grazing_ray_singularity(self, device, dtype):
+        K = torch.tensor([[100.0, 0.0, 4.0], [0.0, 100.0, 3.0], [0.0, 0.0, 1.0]], device=device, dtype=dtype)[None]
+        px = torch.tensor([[[4.0, 3.0]]], device=device, dtype=dtype)
+        plane_normals = torch.tensor([[0.0, 1.0, 0.0]], device=device, dtype=dtype)
+        plane_offsets = torch.tensor([[2.0]], device=device, dtype=dtype)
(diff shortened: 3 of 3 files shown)


ducha-aiki (commented):
This duplicates #4348: both close #4280, change the same denominator guard, and add the same grazing-ray regression; #4348 opened about five hours earlier. I recommend closing this PR and consolidating on #4348.

There is useful value to carry over: the zero-safe `torch.where` implementation at `kornia/geometry/depth.py:274` remains ONNX-exportable. I verified both legacy and dynamo ONNX export and ONNX Runtime output (`[[2e8]]`), whereas #4348's `torch.copysign` formulation breaks both exporters.

If this PR is retained instead, it is not mergeable yet: `tests/geometry/test_depth.py:544` fails the required CPU float16 job because `eps=1e-8` underflows to zero in float16 and `2 / 1e-8` cannot be represented finitely. The test/promise needs a representable numerical range or a dtype-aware policy, and `CHANGELOG.md:362` must cite this PR (`#4352`) rather than the issue.

thegoodengineer:
Closing this as superseded, as suggested. #4348 landed first and is on `main`, #4280 is closed, and its final form already uses the comparison-based `torch.where` guard rather than `torch.copysign`, so the ONNX exportability point is covered there too. The float16 test and CHANGELOG citation issues go away with the PR. Thanks for the review.

PR #4354, opened 2026-09-07
fix(geometry): keep zero-element tensors out of grid_sample in empty 2D warps

### LINKED ISSUE #4032: warp_affine with dsize=(0, w) crashes on MPS with an internal grid_sample assertion

Swapnil-gautam:

`warp_affine`, `warp_perspective` and `remap` crashed on MPS whenever the destination was empty, with an internal PyTorch assert:

```
RuntimeError: [srcBuf length] > 0 INTERNAL ASSERT FAILED at ".../mps/OperationUtils.mm":551,
please report a bug to PyTorch. Placeholder tensor is empty!
```

`_empty_warp_output_2d` already substituted a connected 1x1 stand-in when **`src`** was empty, but built the sampling grid straight from `dsize`, so a zero-sized destination reached `grid_sample` as a zero-element grid. Probing the MPS backend directly (matrix below) shows it rejects **any** zero-element `grid_sample` operand before torch 2.14 — including a zero-element grid against a 1x1 source, and a zero-sized batch — so guarding only the source is not sufficient:

| src / grid (torch 2.9.1, MPS) | result |
| --- | --- |
| normal src, zero-H grid | `Placeholder tensor is empty!` |
| **1x1 src, zero-H grid** | `Placeholder tensor is empty!` |
| zero-batch src + grid | `Placeholder tensor is empty!` |
| 1x1 src, 1x1 grid | ok |

The empty path now samples connected 1x1 stand-ins on **both** operands — every zero dimension clamped to one, non-zero dimensions preserved so `grid_sample` still validates batch, dtype, device and mode — and expands the sampled result to the empty destination, which keeps the autograd links to `src` and `transform`. `grid_sample` therefore never receives a zero-element operand on any backend. Outputs, gradients and the documented empty-source policy are unchanged on CPU and CUDA.

The zero-sized **batch** crash is not in the report but is the same defect and is fixed by the same change; it is covered by the new `batch` parametrization.

PyTorch fixed its MPS backend by 2.14, where all of these return an empty tensor — but kornia supports `torch>=2.0`, so the workaround is still required.

**On the test skips.** `_skip_if_mps_empty_grid_sample` skipped three tests on MPS entirely. They cannot simply be un-skipped: all three call `.backward()`, and `aten::grid_sampler_2d_backward` is unimplemented on MPS before torch 2.14, so the blocking MPS job (torch 2.9.1, no `PYTORCH_ENABLE_MPS_FALLBACK`) would fail on that unrelated gap. Instead the helper is narrowed to exactly that condition and renamed, and the forward half of the contract — the half this PR fixes — is covered on every backend by new forward-only tests for the warps and for `remap`.

Fixes #4032

## What did you check?

Local, CPU, on this commit:

```text
$ python -m pytest tests/geometry/transform/test_imgwarp.py -q
147 passed

$ python -m pytest tests/geometry/transform/ -q
531 passed, 15 skipped, 1 xfailed

$ ruff check / ruff format --check / ty check   # changed files
All checks passed!
```

Also checked `torch.jit.script(warp_affine)` through the empty-destination path (shape and gradients intact).

MPS was verified on hosted Apple Silicon runners (`macos-15`), on torch **2.9.1** and latest, with **no** `PYTORCH_ENABLE_MPS_FALLBACK`, mirroring the blocking MPS job:

- before the change, torch 2.9.1: the report's repro asserts, as filed
- after the change, torch 2.9.1: `warp_affine` `(0, 4)`, `(3, 0)`, `(0, 0)`, `warp_perspective`, `remap` and the batched case all return the empty result; `tests/geometry/transform/test_imgwarp.py --device=mps` → 95 passed, 0 failed
- after the change, torch latest: same, 115 passed, 0 failed

`grep -rnE "#4032|issues/4032" kornia/` is empty; the only references are the new test pins.

`_empty_warp_output_3d` has the same shape of guard, but MPS has no `grid_sampler_3d` at all (the 3-D empty-destination tests are `NotImplementedError` entries in `testing/known_failure_xfails/mps_float32.txt`), so it is out of scope here and its manifest lines are untouched.

## How did AI help?

Investigation and implementation were done with Claude Code. The MPS behaviour was not assumed: the crash, the probe matrix above, and the post-fix results were all produced by running on hosted Apple Silicon runners, since I do not have a Mac. The CPU results above were run locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -356,6 +356,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `warp_affine`, `warp_perspective` and `remap` crashed on MPS for an empty destination -- a `dsize` with a
+  zero dimension, or zero-sized `remap` maps -- with an internal
+  `[srcBuf length] > 0 INTERNAL ASSERT FAILED ... Placeholder tensor is empty!` from PyTorch. The MPS backend
+  rejects *any* zero-element `grid_sample` operand before torch 2.14, including a zero-element grid sampled
+  against a non-empty source, which is what the empty-destination path built. That path now samples a connected
+  1x1 stand-in and expands the result to the requested empty shape, so `grid_sample` never receives a
+  zero-element operand on any backend and an empty warp costs the same whatever the non-zero side of `dsize` is.
+  The empty path also resolves its output batch the way each operation's non-empty path does, so a singleton
+  transform or map batch broadcasts identically and a mismatched batch is rejected rather than silently
+  returning the wrong cardinality. Outputs, autograd links and the documented empty-source policy are unchanged
+  on CPU and CUDA. (#4032, #4354)
+
 * Fixed `unproject_points_z1` depth shape handling for singleton and multi-axis batches,
   accepting both trailing-singleton and flat depth tensors. (#4355)
 
--- a/kornia/geometry/transform/imgwarp.py
+++ b/kornia/geometry/transform/imgwarp.py
@@ -103,7 +103,7 @@ def _empty_warp_output_2d(
     align_corners: bool,
     grid_dtype: torch.dtype,
     fill_value: Optional[torch.Tensor] = None,
-    expand_transform_batch: bool = False,
+    transform_batch_broadcast: str = "none",
     allow_fill: bool = True,
     operand: str = "transform",
 ) -> torch.Tensor:
@@ -111,6 +111,12 @@ def _empty_warp_output_2d(
 
     ``operand`` names the second tensor in the error messages. ``remap`` has no ``transform``
     parameter — it delegates here with its stacked maps — so it must not be told about one.
+
+    ``transform_batch_broadcast`` names the caller's singleton-batch rule, so the empty
+    destination broadcasts and rejects exactly as its non-empty path does:
+    ``"none"`` requires equal batches (``warp_perspective``), ``"src_when_larger"`` expands a
+    singleton transform batch only to a source batch above one (``warp_affine``), and ``"src"``
+    expands it to the source batch unconditionally, a zero batch included (``remap``).
     """
     if src.device != transform.device:
         raise RuntimeError(f"Expected src and {operand} on the same device, got {src.device} and {transform.device}.")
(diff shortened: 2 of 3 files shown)


ducha-aiki (changes requested):
Review at `4c33cf3c` by execution — **CHANGES REQUESTED**.

The MPS workaround fixes the reported zero-element `grid_sample` path, and the focused CPU file passes (262 tests at float32/float64). Two edge cases in the stand-in construction still need correction:

1. **P2 — zero-batch cardinality and validation are lost** at `kornia/geometry/transform/imgwarp.py:131-146`. Clamping each batch to one lets mismatched zero batches reach the return at line 156, despite the helper's stated goal of preserving batch behavior. A valid `remap` broadcast demonstrates the wrong result: with `image.shape == (0, 2, 3, 4)` and maps shaped `(1, 0, 4)`, the non-empty-map path returns batch 0, while this head's empty-map path returns `(1, 2, 0, 4)`. In the reverse invalid case (image batch 1, map batch 0), the non-empty path raises during broadcast but the empty path silently returns `(0, 2, 0, 4)`. The same 0/1 and 1/0 mismatch is silently accepted by both matrix warps for an empty destination. Preserve singleton broadcast to the source batch, explicitly reject the invalid zero-batch combinations, and add the `batch=0` coverage the PR body says exists; the new test currently parametrizes only 1 and 2 (`test_imgwarp.py:53`).

2. **P2 — work is proportional to the nonzero side of an empty output** at `imgwarp.py:142-154`. Keeping `max(dsize[i], 1)` makes `grid_sample` materialize a sampled tensor along an axis that will be discarded. A bounded CPU probe with `warp_affine(..., dsize=(0, 2_000_000))` took 0.00034 s on base versus 0.0198 s here and added about 24 MB RSS; the cost grows without bound although the result has zero elements. Sample a true 1x1 spatial stand-in and then expand to the requested empty shape; mode/dtype/device checks do not require retaining the requested spatial extent. Add a large-aspect empty regression so this remains constant-work.

Finally, the PR created a second `### Bug fixes` heading at `CHANGELOG.md:138` while the existing one is at line 369. Move this entry under the existing heading and cite the PR number as well as the issue (`#4032, #4354`).

Both probes imported `/private/tmp/kornia-review-Oj3X6c/{base,pr-4354}/kornia/__init__.py`, so they compared the intended revisions.

ducha-aiki (approved):
Delta re-review at `b0fc3224` — **APPROVED**.

The two substantive edge cases from the previous review are fixed. The empty path now resolves each operation's actual singleton-batch rule before replacing zero-sized operands, so zero batches retain their cardinality and invalid 0/1 combinations reject consistently with the non-empty path. It samples a true `(1, 1)` spatial stand-in and only then expands to the requested empty shape, keeping work constant when the other output axis is large. The CHANGELOG entry is under the existing Bug fixes heading and cites #4032 and #4354.

Focused validation at this exact head passed 152 empty-destination/map/source cases on CPU float32 and float64. Additional batch/channel/fill probes covered zero, singleton, and multi-item batches and zero through RGB channel counts; supported cases preserved output shape and connected gradients. The merge-tree against current `main` is clean and all 40 hosted checks are green.

PR #4356, opened 2026-09-08
fix(geometry): use math.pi in rad2deg/deg2rad to preserve full precision

### LINKED ISSUE #3937: rad2deg/deg2rad: float32 kornia.constants.pi is cast to the input dtype — integer inputs divide by 3, float64 loses ~7 digits
`kornia.geometry.conversions.rad2deg` and `deg2rad` divide by `kornia.constants.pi` after casting that constant to the **input** tensor's dtype (`conversions.py:97` and `:127`: `180.0 * tensor / pi.to(tensor.device).type(tensor.dtype)`). `pi` is created as `torch.tensor(3.14159265358979323846)` (`kornia/constants.py:25`), which materialises as a **float32** tensor. Two consequences follow: for an **integer** input tensor the constant is cast to int64 and becomes literally `3`, so the conversion is off by ~5 %; for a **float64** input the constant is upcast from its already-truncated float32 value, so ~7 significant digits are silently lost. `angle_to_rotation_matrix` calls `deg2rad` and inherits the integer bug directly. `torch.rad2deg` / `torch.deg2rad` are correct in both cases.

## Repro

```python
import math
import torch
from kornia.constants import pi
from kornia.geometry.conversions import rad2deg, deg2rad, angle_to_rotation_matrix

print("pi.dtype, pi.item():", pi.dtype, pi.item())
print("pi.type(torch.int64):", pi.type(torch.int64))
print("pi.type(torch.float64).item():", pi.type(torch.float64).item(), "| math.pi:", math.pi)

# (1) integer input
print("kornia rad2deg int:", rad2deg(torch.tensor([1, 2, 3])))
print("torch  rad2deg int:", torch.rad2deg(torch.tensor([1, 2, 3])))
print("kornia deg2rad int:", deg2rad(torch.tensor([180, 90])))
print("torch  deg2rad int:", torch.deg2rad(torch.tensor([180, 90])))

# (2) downstream: angle_to_rotation_matrix with an integer angle
print("angle_to_rotation_matrix(tensor([90])) :", angle_to_rotation_matrix(torch.tensor([90])))
print("angle_to_rotation_matrix(tensor([90.])):", angle_to_rotation_matrix(torch.tensor([90.])))

# (3) float64 precision
x64 = torch.tensor(math.pi, dtype=torch.float64)
print("kornia rad2deg f64 err:", rad2deg(x64).item() - 180.0)
print("torch  rad2deg f64 err:", torch.rad2deg(x64).item() - 180.0)
d64 = torch.tensor(180.0, dtype=torch.float64)
print("kornia deg2rad f64:", deg2rad(d64).item(), "| torch:", torch.deg2rad(d64).item(), "| math.pi:", math.pi)
print("kornia deg2rad f64 abs err:", abs(deg2rad(d64).item() - math.pi))
```

## Observed

```
pi.dtype, pi.item(): torch.float32 3.1415927410125732
pi.type(torch.int64): tensor(3)
pi.type(torch.float64).item(): 3.1415927410125732 | math.pi: 3.141592653589793
kornia rad2deg int: tensor([ 60., 120., 180.])
torch  rad2deg int: tensor([ 57.2958, 114.5916, 171.8873])
kornia deg2rad int: tensor([3.0000, 1.5000])
torch  deg2rad int: tensor([3.1416, 1.5708])
angle_to_rotation_matrix(tensor([90])) : tensor([[[ 0.0707,  0.9975],
         [-0.9975,  0.0707]]])
angle_to_rotation_matrix(tensor([90.])): tensor([[[-4.3711e-08,  1.0000e+00],
         [-1.0000e+00, -4.3711e-08]]])
kornia rad2deg f64 err: -5.008956179608504e-06
torch  rad2deg f64 err: 0.0
kornia deg2rad f64: 3.1415927410125732 | torch: 3.141592653589793 | math.pi: 3.141592653589793
kornia deg2rad f64 abs err: 8.742278012618954e-08
```

## Expected

- `rad2deg(torch.tensor([1, 2, 3]))` should be `[57.2958, 114.5916, 171.8873]`, matching `torch.rad2deg`. Observed `[60., 120., 180.]` is exactly what you get from dividing by 3 instead of π — the output dtype is already promoted to float32, so nothing about the integer input forces an integer constant.
- `angle_to_rotation_matrix(torch.tensor([90]))` should equal `angle_to_rotation_matrix(torch.tensor([90.]))` (the quarter-turn matrix). Observed is the matrix for `90 * 3 / 180 = 1.5` rad ≈ 85.94°.
- `deg2rad(torch.tensor(180., dtype=torch.float64))` should be `3.141592653589793` (what `torch.deg2rad` returns). Observed `3.1415927410125732` carries an absolute error of `8.74e-08` — about **7 significant digits of the 16 that float64 provides**. Symmetrically `rad2deg(π, f64)` is off 180° by `-5.0e-06` where `torch.rad2deg` is exact.

Note that `rad2deg(deg2rad(x))` still round-trips cleanly, because the same biased constant appears in the numerator and denominator and cancels — which is why this has gone unnoticed.

Suggested fix: delegate to `torch.rad2deg` / `torch.deg2rad`, or at minimum build the constant at the input's dtype from a float64 source rather than casting a float32 tensor.

To size the blast radius on the common float32 path, here is the exact comparison snippet (fixed seed, four input ranges spanning six orders of magnitude — absolute deltas are scale-dependent and not meaningful here, so this reports ULP distance and relative error):

```python
import torch
from kornia.geometry.conversions import rad2deg, deg2rad

def cmp(name, fn_k, fn_t, x):
    a, b = fn_k(x), fn_t(x)
    same = (a == b).float().mean().item()
    rel = (a - b).abs() / b.abs().clamp(min=torch.finfo(torch.float32).tiny)
    ulp = (a.view(torch.int32) - b.view(torch.int32)).abs().max().item()
    print(f"{name:>24}: bit-identical {same * 100:5.1f}%  max rel {rel.max().item():.3e}  max ULP {ulp}")

g = torch.Generator().manual_seed(0)
ranges = {
    "U(-10,10)": torch.rand(100000, generator=g) * 20 - 10,
    "U(-1e3,1e3)": torch.rand(100000, generator=g) * 2e3 - 1e3,
    "U(-1e6,1e6)": torch.rand(100000, generator=g) * 2e6 - 1e6,
    "U(0,1)": torch.rand(100000, generator=g),
}
for n, x in ranges.items():
    cmp("rad2deg " + n, rad2deg, torch.rad2deg, x)
for n, x in ranges.items():
    cmp("deg2rad " + n, deg2rad, torch.deg2rad, x)
```

```
       rad2deg U(-10,10): bit-identical  52.2%  max rel 1.869e-07  max ULP 2
     rad2deg U(-1e3,1e3): bit-identical  52.1%  max rel 1.869e-07  max ULP 2
     rad2deg U(-1e6,1e6): bit-identical  52.2%  max rel 1.867e-07  max ULP 2
          rad2deg U(0,1): bit-identical  51.9%  max rel 1.864e-07  max ULP 2
       deg2rad U(-10,10): bit-identical  57.1%  max rel 1.675e-07  max ULP 2
     deg2rad U(-1e3,1e3): bit-identical  55.6%  max rel 1.673e-07  max ULP 2
     deg2rad U(-1e6,1e6): bit-identical  55.2%  max rel 1.676e-07  max ULP 2
          deg2rad U(0,1): bit-identical  55.9%  max rel 1.676e-07  max ULP 2
```

So on float32 roughly half the values stay bit-identical and the rest move by **at most 2 ULP** (max relative error ~1.9e-07, i.e. about one float32 epsilon) — stable across all four ranges. Meanwhile the integer and float64 paths go from wrong to correct.

## Classification

**Clearly-broken-output bug**, not a semantic/default-convention change. The integer path returns numbers that are wrong by ~5 % under any reading of "convert radians to degrees" — there is no convention under which π equals 3 — and the float64 path is a plain precision defect against an exact reference. Per the reference-core lifecycle (Part I), this class may land early as an ordinary bugfix rather than waiting for the coordinated Repair & Deprecation window. The ≤2 ULP float32 movement measured above is the only behavior change on the common path.

## Environment

kornia 0.9.0rc1 @ `3358ef16`, torch 2.9.1, cpu.

li-lizhe:
fix(geometry): use math.pi in rad2deg/deg2rad to preserve full precision

Fixes #3937

## Problem

`kornia.geometry.conversions.rad2deg` / `deg2rad` convert the float32
constant `pi` to the input tensor's dtype via:

```python
pi.to(tensor.device).type(tensor.dtype)
```

This has two precision bugs:

1. **Integer inputs** — `pi` is truncated to `3` (its float32 value cast to
   `int64`), so results are ~5% off (e.g. `rad2deg([1,2,3])` returns
   `[60,120,180]` instead of `[57.3, 114.6, 171.9]`).
2. **float64 inputs** — the constant is already truncated to float32 precision
   before the division, so ~7 significant digits are silently lost and can
   never be recovered by upcasting.

## Root cause

The constant `180.0` is a Python double and always exact, but `pi` is a
float32 tensor that gets force-cast to the input dtype. Integer casting
truncates to `3`; float64 casting loses the float32→float64 recoverable
precision.

## Fix

Use the built-in `math.pi` (Python double, full precision) for the constant,
and drop the `pi.to(device).type(tensor.dtype)` cast. The constant is now
always exact, independent of input dtype. `180.0` was already a Python
double, so the two are now consistent.

```python
# before
return 180.0 * tensor / pi.to(tensor.device).type(tensor.dtype)
# after
return 180.0 * tensor / math.pi
```

## Verification (Ascend NPU)

- `rad2deg(torch.tensor([1,2,3]))` → `[57.2958, 114.5916, 171.8873]`
  (before: `[60,120,180]`)
- `rad2deg(torch.tensor(math.pi, dtype=torch.float64))` → `180.0` (<1e-10 err)
- Results match NumPy `np.degrees` ground truth; device/dtype placement
  preserved.

### DIFF
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -24,7 +24,6 @@ from typing import Optional
 import torch
 import torch.nn.functional as F
 
-from kornia.constants import pi
 from kornia.core._compat import deprecated
 from kornia.core.check import KORNIA_CHECK, KORNIA_CHECK_SHAPE
 from kornia.core.utils import _inverse_3x3_closed_form, _torch_inverse_cast, is_compiling
--- a/kornia/geometry/conversions.py
+++ b/kornia/geometry/conversions.py
@@ -17,6 +17,7 @@
 
 from __future__ import annotations
 
+import math
 import warnings
 from typing import Optional
 
@@ -113,7 +114,7 @@ def rad2deg(tensor: torch.Tensor) -> torch.Tensor:
     if not isinstance(tensor, torch.Tensor):
         raise TypeError(f"Input type is not a torch.Tensor. Got {type(tensor)}")
 
-    return 180.0 * tensor / pi.to(tensor.device).type(tensor.dtype)
+    return 180.0 * tensor / math.pi
 
 
 def deg2rad(tensor: torch.Tensor) -> torch.Tensor:
@@ -146,7 +147,7 @@ def deg2rad(tensor: torch.Tensor) -> torch.Tensor:
     if not isinstance(tensor, torch.Tensor):
         raise TypeError(f"Input type is not a torch.Tensor. Got {type(tensor)}")
 
-    return tensor * pi.to(tensor.device).type(tensor.dtype) / 180.0
+    return tensor * math.pi / 180.0
 
 
 def pol2cart(rho: torch.Tensor, phi: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:


ducha-aiki (changes requested):
[P1] Keep the in-tree `pi` consumer importable. Removing `pi` from `kornia.geometry.conversions` at `kornia/geometry/conversions.py:27` leaves `kornia/feature/siftdesc.py:28` importing that name. On this head, a clean `import kornia` raises `ImportError: cannot import name 'pi'`, which explains the matrix-wide test failures. Either keep the compatibility import or update every consumer in the same change; #4358 already takes the latter route and includes the regression/documentation cleanup, so consolidating this duplicate into #4358 is preferable.

[P2] Retire the #3937 failure pins and stale defect prose together with the fix. The warnings at `kornia/geometry/conversions.py:127-132` still say `deg2rad` returns the old truncated result, and the existing strict xfail/wart tests still encode the broken behavior. Once the import error is repaired, those pins will fail or misdocument the implementation. This user-visible bug fix also needs a `CHANGELOG.md` entry citing #4356 if this PR remains open.

Validation: `import kornia` fails from the PR worktree at `4d59910cb087e189b9662f8cb7a6209e309243f9`; the worktree import path was verified before the exception. No prior comments or formal reviews exist on this head.

li-lizhe:
Closing this PR as a duplicate — #4358 by @Hrithikcrick already covers the same rad2deg/deg2rad precision fix with all consumer updates (import path, regression tests, obsolete warnings, CHANGELOG). Per the maintainer's review recommendation, consolidating into #4358 is the cleaner path. Thank you for the review!

PR #4357, opened 2026-09-08
fix: correct quartic real root selection

### LINKED ISSUE #4346: solve_quartic returns wrong real roots for ~28% of general quartics: the y-selection argmax ranks solve_cubic's zero placeholders, and the R≈0 fallback uses an absolute tolerance on sqrt(R_sq)
`solve_quartic` returns wrong real roots for a large share of ordinary inputs. Two independent defects, both in the Ferrari step after the resolvent cubic; neither is touched by #4338 or #4339, which only guard gradients, and neither is seen by the test suite because `TestQuarticSolver::test_random` builds every polynomial from four *real* roots.

Measured on `main` `217b92e6`, torch 2.14.0, CPU, float64 unless stated, against `numpy.roots` (a returned entry counts as a real root only if some numpy real root is within 1e-5 of it; "spurious" is a nonzero entry that matches no real root):

| coefficients (leading 1, others uniform in [-5, 5]) | rows | rows with a missing real root | spurious entries |
| --- | --- | --- | --- |
| general | 2000 | 558 (28 %) | 1750 |
| `x⁴ + b x³ + e` | 2000 | 1003 (50 %) | 1003 |
| `x⁴ + b x³ + e`, `e < 0` | 2000 | 2000 (100 %) | 2000 |
| biquadratic `x⁴ + c x² + e` | 2000 | ~55 % | 1042 |
| `x⁴ + d x + e` with `d = ±1e-6` | 2000 | 1886 (94 %), worst error 5.6e-3 | 1886 |
| `x⁴ + d x + e`, `d` uniform | 2000 | 0 | 0 |

Clean repros (the first two behave the same in float32; the 1e-6 ones need float64, float32's `zero_tol=1e-6` happens to catch them):

```python
import torch
from kornia.geometry.solvers import solve_quartic
solve_quartic(torch.tensor([[1., 2., 0., 0., -16.]], dtype=torch.float64))
# tensor([[ 0., -0.,  0., -2.]])      real roots are -2.760555 and 1.638345; -2 is not a root (16-16-16 = -16)
solve_quartic(torch.tensor([[1., 0., -4., 0., -5.]], dtype=torch.float64))
# tensor([[ 2.,  0.,  0., -2.]])      x⁴ - 4x² - 5 = (x²-5)(x²+1), roots ±2.236068; ±2 are not roots
solve_quartic(torch.tensor([[1., 0., 0., 1e-6, -16.]], dtype=torch.float64))
# tensor([[0., 0., 1.988574, -1.988574]])   roots are ±2 to 1e-7; a 1e-6 perturbation moved them by 1.1e-2
solve_quartic(torch.tensor([[1., 1e-6, 0., 0., -16.]], dtype=torch.float64))
# tensor([[0., -0., 0., -1e-6]])      the ±2 roots of x⁴ - 16 vanish for any b in [1e-6, 0.1]
```

## Defect 1 — the "pick y maximising R²" step ranks `solve_cubic`'s zero placeholders as candidates

`solve_cubic` returns `0` in the slots of non-real roots (documented). `solve_quartic` then does

```python
R_sq_candidates = 0.25 * A_sq.unsqueeze(-1) - B.unsqueeze(-1) + y_roots
best_idx = torch.argmax(R_sq_candidates, dim=-1, keepdim=True)
```

over all three slots, so a placeholder `y = 0` competes with the real resolvent root. Whenever the resolvent has one real root and it is negative, the placeholder wins (`A²/4 - B + 0 > A²/4 - B + y`), Ferrari continues with `y = 0`, and both quadratics are wrong. For `x⁴ + 2x³ - 16` the resolvent is `y³ + 64y + 64`, `solve_cubic` gives `[-0.985, 0, 0]`, the candidates are `[0.015, 1, 1]`, and `argmax` picks a placeholder.

Every one of the 558 general failures above has exactly this shape (checked row by row). Masking the candidates by "is a root of the resolvent" (`|p(y)| ≤ tol`, `-inf` otherwise) before the `argmax` takes the general case to **0 missing / 0 spurious** on the same 2000 rows. `solve_cubic` returning a real-root mask alongside the roots, or the quartic recomputing it from the residual, are both fine; the argmax must not see placeholders.

## Defect 2 — the `R ≈ 0` fallback keys on an absolute `zero_tol` of `R`, so rounding noise in `R_sq` sends `R ≈ 1e-8` down the division path

```python
R = torch.sqrt(torch.clamp(R_sq, min=0.0))
mask_R_small = torch.abs(R) < zero_tol          # 1e-6 float32, 1e-12 float64
E[mask_R_large] = (A*y - 2*C) / (4*R)
```

For a biquadratic (or a near-biquadratic) `R_sq` is analytically 0 but comes out of the cubic as `±1e-16`; `sqrt` turns `1e-16` into `1e-8`, which is far above `1e-12`, and `E` is computed as `(≈0)/(4e-8)` — arbitrary. That is the ±1.9886 result for `x⁴ + 1e-6 x - 16` and the ~55 % biquadratic failure rate that survives the defect-1 fix (588 / 2000 still missing with placeholders masked). The threshold has to be relative to the coefficient scale and applied to `R_sq`, not to `R` after a `sqrt`.

## Test gap

`test_random` constructs `(x-r1)(x-r2)(x-r3)(x-r4)` from `randn` roots, so every polynomial has four real roots and a resolvent with three real roots; neither branch above is ever exercised. The pin for this issue should draw coefficients directly (or two real roots plus a complex pair) and compare with `numpy.roots` real parts, plus the four literals above.

## Scope

No in-tree caller (`git grep solve_quartic kornia/` hits only the exports), so this is the public API only. Found while reviewing #4339's biquadratic pins: the finite-difference derivative at `x⁴ - 16` could not be measured because the forward is discontinuous there. #4338 (batch leak in `solve_cubic`, #4334) and #4339 (`sqrt` guards) are gradient fixes and stay valid; whoever takes this should land it after them to avoid three-way conflicts in `polynomial_solver.py` and `TestQuarticSolver`.

victorwon2001:

Filters Ferrari resolvent candidates by a scaled cubic residual so `solve_cubic` zero placeholders cannot win the `R^2` selection. Valid candidates are ranked by `R^2`, while rows with no candidate inside the validation threshold use the smallest scaled residual; `float16` and `bfloat16` quartic inputs evaluate Ferrari intermediates in `float32` and cast the roots back to the input dtype.

Cancellation-level `R^2` values are snapped to zero relative to the scale of `A^2 / 4`, `B`, and the selected resolvent root before the gradient-safe square root. The linear-term division and constant-term Ferrari `E` forms are compared by normalized coefficient reconstruction error in every actual Ferrari compute dtype, with the constant-term form forced at `R == 0`; float32 additionally keeps the guarded one-step refinement of the selected resolvent root. The public `(B, 4)` real-root/zero-placeholder contract and cubic fallback are unchanged.

Fixes #4346

## What did you check?

On Python 3.11 / PyTorch 2.9.1, the complete CPU float32/float64 polynomial-solver module, both strict CPU half-precision solver profiles, and the repository Ruff lint task pass. No fresh CUDA, MPS, or full-suite run was performed locally for this follow-up.

## How did AI help?

AI helped reproduce the remaining float64 factorization and biquadratic cancellation cases, calibrate the coefficient-scale-relative `R^2` threshold, extend the existing `E` reconstruction selection to all Ferrari compute dtypes, add the focused regressions, and verify the solver and half-precision profiles described above.


ducha-aiki (changes requested):
[P1] Make the resolvent validity tolerance work for half dtypes. `zero_tol` is `1e-12` for every dtype except float32 at `kornia/geometry/solvers/polynomial_solver.py:268`, and the new filter applies it at lines 306-318. In float16/bfloat16 the normalized residual of a genuine resolvent root is ordinarily around `1e-3`, so all three candidates can become invalid and `argmax([-inf, -inf, -inf])` silently selects slot 0. For `torch.tensor([[1., -10., 35., -50., 24.]], dtype=torch.bfloat16)`, whose roots are 1, 2, 3, 4, the prior head returns finite approximations `[1.1172, 1.1172, 3.875, 3.875]`; this head returns `[-inf, 0, 0, inf]`. Please use a dtype-aware residual tolerance/fallback and add a half-precision regression that catches non-finite or dropped roots.

The hosted half jobs corroborate this: float16 has three numerical failures plus two strict XPASS manifest entries, and bfloat16 has two numerical failures. Update the known-failure manifests for any genuinely repaired nodes rather than leaving the suite red. This PR also conflicts with #4339 in the same fallback block; merge #4338, then #4339, rebase this change, and rerun the solver/gradient coverage so their guards are preserved. Add a `CHANGELOG.md` bug-fix entry citing #4357.

Validation at `844d6d953a5c2a8d5125e28d9dc800f49c14ab9d`: 127 CPU float32/float64 solver tests passed (4 skipped); focused hosted and local half runs reproduced the failures above. The head merges cleanly with #4338 and conflicts with #4339. No prior comments or formal reviews exist.

ducha-aiki (changes requested):
Reviewed exact head `c91a91a9bae01a9fd9e8d5237e746a6e54521317`.

[P1] The new R-near-zero conditioning/selection heuristic at kornia/geometry/solvers/polynomial_solver.py:365-381 still returns non-roots for ordinary float32 inputs. The literal [1., -2.4491360, 1.1192101, -2.0354464, -3.8231988] returns approximately [2.5948343, -0.7535356, 0, 0]; the true real roots are approximately 2.552848 and -0.782185. The returned nonzero values have absolute polynomial residuals 0.9765 and 0.2836. An independent 10,000-row comparison against NumPy found 16 float32 root-set misses and 37 bad-residual rows (float64 had no misses). Correct the conditioning choice and add this literal as a regression. The earlier half-intermediate/tolerance issue and release note are fixed and hosted checks are green, but current-main conflicts also remain. Coordinate with the overlapping #4375.

ducha-aiki (changes requested):
Re-review of exact head `0115b437`.

The prior P1 is fixed. The literal `[1., -2.4491360, 1.1192101, -2.0354464, -3.8231988]` now returns `[2.5528479, -0.7821853]` with float32 residuals below 5e-7 (main: non-roots with residual 1.1), and an independent 10,000-row N(0, 2) sweep against `numpy.roots` goes from 2453 misses / 2457 bad-residual rows on main to 0 / 0 here in both float32 and float64. Every family in #4346's table is clean, float16/bfloat16 give no non-finite rows (main had 25) and return the right roots for `(x-1)(x-2)(x-3)(x-4)`, both regression literals are pinned and fail on main, and all 38 hosted checks are green.

Two precision issues remain before this can carry `Fixes #4346`:

- [P2] `polynomial_solver.py:393-411`: the selection between the division form and the constant-term identity for `E` is gated on `coeffs.dtype == torch.float32`; float64 always takes `E = sqrt(y^2/4 - D)`, which cancels catastrophically when the two quadratic factors share a constant term. For `(x^2 + 3x + 2)(x^2 - 5x + 2.000001)` in float64 main gives residual 5.2e-14 and this head 1.5e-8 (relative root error 8.9e-16 -> 4.2e-10). That is a regression on an input main got right. Making the selection unconditional restores 8.9e-16 with the sweeps and the test file unchanged in my experiment; please apply it for every dtype and pin this literal.
- [P2] `polynomial_solver.py:359-374`: #4346's second defect asks for a threshold relative to the coefficient scale applied to `R_sq`. This head removes the threshold entirely; `E` no longer divides by `R`, but `R = sqrt(R_sq)` still enters both quadratics' linear terms, so for a biquadratic, where `R_sq` is analytically 0 but numerically eps-level noise, the roots move by about `R/2`. On the issue's biquadratic family (2000 rows) float32 max relative root error is 3.0e-4 (`[1, 0, 2.9696312, 0, -2.3985832]` returns `+/-0.81259, +/-0.81308` against the true `+/-0.8128379`, residual 1.7e-3) and float64 2.4e-8. Not a regression, but the prescribed snap of eps-level `R_sq` to zero brings float32 to about 1.6e-6 and needs your calibration of the constant.
- [P3] The docstring does not say that float16/bfloat16 inputs are solved in float32 and cast back, while the release note advertises it; one sentence in the notes.

On the competing #4375: both heads are 0 / 0 on the full-precision sweeps and both share the two P2s above, but #4375 has no half-precision path (float16 returns `[5, 0, 5, 0]` for roots 1..4, 49 misses / 120 bad rows on a 5000-row sweep; bfloat16 1306 misses), so this PR is the one to finish.

Validation (isolated worktree, import origin verified): `test_polynomial_solver.py` CPU float32/float64 141 passed, 8 skipped; cpu-float16 and cpu-bfloat16 profiles 65 passed, 2 xfailed each with no XPASS, so the manifest removals are consistent; MPS float32 68 passed; ruff clean; `#4346` grep clean in `kornia/`. Not run: CUDA, full suite.

ducha-aiki (changes requested):
Re-review of exact head `4ace02f9`.

Both prior P2s are addressed. The `E`-form selection is now unconditional, so the float64 shared-constant-term regression is gone: `(x^2+3x+2)(x^2-5x+2.000001)` returns residual 1.2e-14 / relative root error 1.1e-15 (main 1.2e-14, previous head 1.5e-8), and the new `test_E_reconstruction_precision_float64` fails on `0115b437` and passes here. The relative `R^2` snap brings the float32 biquadratic literal from 3.4e-4 to 1.2e-9 relative error. Independent 2000-row sweeps (general N(0, 2), four real roots, real pair x complex pair, biquadratic, shared-constant-term; `numpy.roots` reference) are 0 misses / 0 bad residuals in float32 and float64 for every family except the one below; float16/bfloat16 have zero non-finite rows in all families and return `[1, 2, 3, 4]` exactly for `(x-1)(x-2)(x-3)(x-4)`; gradcheck passes in float64 on four non-degenerate literals and the autograd Jacobian matches both the implicit-function analytic Jacobian (7e-14) and central finite differences. The P3 docstring sentence is in.

One item is left before `Fixes #4346` is accurate:

- [P2] `kornia/geometry/solvers/polynomial_solver.py:365-370`: the snap tolerance `0.5 * eps * scale` is too tight for float64. On the issue's own literal `[1, 0, -4, 0, -5]` in float64, `solve_cubic` returns `y = -3.999999999999999` (one ulp off), `R_sq = 8.881784197001252e-16` is exactly one ulp of `B = 4` and lands one bit *above* `snap_tol = 8.881784197001251e-16`, so `R = 2.98e-8` enters both quadratics and the roots come back as `+-2.2360679924` instead of `+-2.2360679775` (relative error 6.7e-9, residual 4e-7) -- defect 2's mechanism, eight digits lost. In the 2000-row biquadratic family float64 has 93 rows with scaled residual above 1e-9 (max 1.5e-8) while every other family sits at 1e-13..1e-15. Re-running the same module with only the multiplier changed: `2*eps` leaves 4 rows, `4*eps` gives 0 rows / max residual 3.7e-16 / literal error 2e-16 with no change in any other float64 family; in float32 `4*eps` worsens the general family (max scaled residual 2.3e-6 -> 7.9e-5) and `8*eps` adds a bad row, so keep `0.5*eps` there (the float32-only refinement already recomputes `R_sq` to exactly 0 on this input). A dtype-aware multiplier, or extending the one-step resolvent refinement to float64, both work; please pin `[1, 0, -4, 0, -5]` in float64 at about 1e-12 relative. Note that the float64 leg of `test_biquadratic_R_sq_relative_snap` passes on main and on `0115b437` (atol 5e-6), so only its float32 leg currently pins the delta.

- [P3, pre-existing, follow-up issue rather than this PR] Near-perfect-square quartics in float32: when the resolvent has a near-double root, float32 `solve_cubic` loses it, the best valid candidate has `R_sq < 0`, `R` is clamped to 0 and the constant-term `E` radicand is negative, so the output is the real roots of `x^2 + A/2 x + y/2`. `(x^2+2x+5)(x^2+2x+5.01)` float32 returns `[1.0012, -3.0012, 1.0012, -3.0012]` (residual 64); main returns `[0, -2, 0, -2]` (residual 25) and #4375 matches this head; float64 is correct everywhere. On a 2000-row near-square sweep the three heads are equal (300/642 no-real-root rows spurious). A cheap guard is "best valid `R_sq < -tol` means the largest resolvent root was lost", but that is outside #4346's scope.

On #4375: this PR remains the one to finish. #4375 at `eee57a2a` has no half-precision path (float16 returns `[5, 0, 5, 0]` for roots 1..4 and has 19/242/12 non-finite rows in the four-real / real-x-complex / shared-constant families), keeps the float64 shared-constant-term regression (1 miss / 123 bad rows, 1.5e-8 on the literal) and the float32 biquadratic error of 3.4e-4.

Validation (isolated worktree, import origin printed and checked in every snippet): `tests/geometry/solvers` CPU float32/float64 199 passed, 9 skipped; cpu-float16 and cpu-bfloat16 profiles 93 passed, 2 xfailed each with no XPASS; MPS float32 96 passed; pinned ruff check and format clean; `#4346` grep clean in `kornia/`, `tests/` and `docs/`; the new tests fail on main (13 of 15, the two passes being the regression pin and the float64 leg noted above). Not run: CUDA, full suite.

ducha-aiki (approved):
Re-review of exact head `ad248967` (delta since `4ace02f9`, whole diff re-read). Thanks for the quick turnaround. The last P2 is fixed, and I have no blocking findings left, so this is ready to land.

**Prior P2: fixed.** `polynomial_solver.py:371` now uses `4 * eps * scale` for float64 and keeps `0.5 * eps` otherwise. On the issue literal `[1, 0, -4, 0, -5]` in float64, the head returns `+-2.2360679775` (relative error 7.9e-16, residual 1.1e-14). `4ace02f9` returned `2.2360679924` (6.7e-9, residual 4.0e-7). The new `test_biquadratic_R_sq_relative_snap_float64_issue_literal` pins it: with the PR's test file it is the only failure out of 51 quartic tests on the `4ace02f9` solver, it also fails on main, and it passes here.

To check that nothing else moved, I compared the roots bitwise between `ad248967` and `4ace02f9` on the same seeded sweep: 8 families × 2000 rows × 4 dtypes (general N(0, 2), four real roots, real pair × complex pair, biquadratic, shared constant term, double root, two complex pairs, near-zero leading coefficient). The only rows that differ are 93 float64 biquadratic rows, the ones the fix targets. Every float32, float16 and bfloat16 output is bitwise identical to `4ace02f9`, and so is every other float64 family.

| 2000-row families, `numpy.roots` / constructed-root reference | main | `4ace02f9` | **`ad248967`** | #4375 `1f2b3633` |
| f64 `[1,0,-4,0,-5]` relative root error | returns `+-2` | 6.7e-9 | **7.9e-16** | 6.7e-9 |
| f64 biquadratic: rows with scaled residual > 1e-9 (max) | 471 (0.82) | 93 (1.5e-8) | **0 (3.7e-16)** | 376 (2.7e-8) |
| f64 shared constant term: miss / bad rows (max) | 0 / 0 (5.5e-15) | 0 / 0 | **0 / 0 (5.5e-15)** | 1 / 123 (1.6e-8) |
| f32 biquadratic: rows > 1e-4 (max) | 471 (0.82) | 0 (1.4e-7) | **0 (1.4e-7)** | 302 (4.6e-4) |
| f32 `[1,0,2.9696312,0,-2.3985832]` relative error | no roots | 1.2e-9 | **1.2e-9** | 3.5e-4 |
| f16 / bf16 non-finite rows (7 families, 14k rows) | 2095 / 77 | 0 / 0 | **0 / 0** | 0 / 0 |
| f16 / bf16 non-finite rows, seeded 50k rows in [-20, 20] | 1121 / 2 | 0 / 0 | **0 / 0** | 0 / 0 |
| #4346 table families (6 × 2000), f64 / f32 rows missing a real root | 4969 / 4648 | 0 / 0 | **0 / 0** | 0 / 11 |
| #4474 near-square f32: spurious rows out of 642 | 300 | 300 | **300** | 300 |

Two optional notes. Neither blocks the merge.

**[P3, optional] A near-biquadratic `sqrt(eps)` accuracy floor in float64, shifted by the wider snap.** When `R_sq` is genuinely a few eps (say `x^4 + c x^2 + d x + e` with `|d|` in 1e-8..1e-6), `R = sqrt(R_sq)` is only accurate to about `sqrt(eps)` whether or not the snap fires. The `4*eps` budget therefore trades accuracy between neighbouring inputs. On 20k-row sweeps, counting rows with a scaled residual above 1e-9:
- `|d| <= 1e-10`: 1116 rows become 0.
- `|d|` in 1e-8..1e-6: 4577 rows become 5741, and the max residual goes from 2.7e-8 to 5.4e-8.
- `|d|` in 1e-10..1e-4: 1980 rows become 2077.

For comparison, main has 9810 rows (0.82) on the middle family and #4375 has 5972 (2.8e-8). Our earlier note that `4*eps` changed no other float64 family only covered the five families we had swept. In a quick experiment outside the library, one Newton step on the returned nonzero roots against the original quartic took the middle family to 0 rows (max 2.1e-15). That would be a reasonable follow-up, but it is not needed for #4346.

**[P3, pre-existing, follow-up rather than this PR] float16 never reaches the cubic fallback.** `zero_tol = 1e-12` (`polynomial_solver.py:280`) rounds to 0 in float16, so `abs(a) < zero_tol` is never true. As a result, `solve_quartic(torch.tensor([[0., 1., -6., 11., -6.]], dtype=torch.float16))` returns `[nan, -inf, nan, nan]`. Main returns all-nan, #4375 behaves the same, and bfloat16/float32 return the cubic roots `1, 2, 3`. In the near-zero leading coefficient family this is 2000 of 2000 float16 rows on every head. Because the Ferrari path is already promoted to float32, deciding the fallback in float32 would also fix this, but it is outside #4346.

Also checked:
- **Duplicate:** this PR is now strictly ahead of #4375. They share the same half-precision promotion, and #4375 still has the float64 shared-constant regression and no `R^2` threshold.
- **#4474:** unchanged, as expected. The literal still returns `[1.0012, -3.0012, 1.0012, -3.0012]` (residual 64), and the near-square sweep is still 300/642 on main and on both PRs.

**Verified on this head** (isolated worktree, import origin printed in every probe):
- **Tests:** `tests/geometry/solvers`
  - CPU float32/float64: 200 passed, 10 skipped.
  - cpu-float16 and cpu-bfloat16 known-failure profiles: 93 passed, 12 skipped, 2 xfailed each, no XPASS.
  - MPS float32: 96 passed, 11 skipped.
  - `test_polynomial_solver.py` on torch 2.5.1, float32/float64: 145 passed, 10 skipped.
- **MPS sweep:** float32 and float16 bad-row counts on MPS match CPU in every family.
- **Gradients:** float64 `gradcheck` passes on six non-degenerate literals (roots 1..4, `[1,2,0,0,-16]`, the float32 P1 literal, shared constant term, real × complex pair, near-biquadratic `d=1e-3`) on torch 2.14 and 2.5.1. The autograd Jacobian matches the implicit-function Jacobian `-[r^4..1]/p'(r)` to within 7e-14 (3.7e-9 for the near-biquadratic).
- **Continuity:** the forward is now continuous at the `x^4 - 16` and `x^4 - 1` pins. Roots move 5e-7 under 1e-6 coefficient perturbations, against 2.0 on main.
- **Compile:** `torch.compile` matches eager bitwise in float64 and float16, and to within 1 ulp (1.2e-7) in float32.
- **Batching:** no cross-row contamination. Mixed-family batches differ from per-family batches only by 1–2 ulp in one float32 row.
- **Housekeeping:**
  - pinned ruff 0.16.6 `check` and `format --diff` are clean.
  - `grep -rnE "#4346|issues/4346"` over `kornia/`, `tests/`, `testing/` and `docs/` is clean; the stale "forward is discontinuous (#4346)" test comment was rewritten.
  - The CHANGELOG entry is under `Unreleased / Bug fixes` and cites #4357.
- **Branch and CI:** mergeable, 17 commits behind main (main only touched two comments in this file since the merge base). CI is 40/40 green.
- **Not run:** CUDA, full suite.

PR #4358, opened 2026-09-08
Fix rad2deg and deg2rad precision for integer and float64 inputs

### LINKED ISSUE #3937: rad2deg/deg2rad: float32 kornia.constants.pi is cast to the input dtype — integer inputs divide by 3, float64 loses ~7 digits

Hrithikcrick:
Fixes #3937

## Summary

- Fix `rad2deg` and `deg2rad` for integer tensor inputs.
- Preserve float64 precision using Python's high-precision `math.pi`.
- Use basic tensor arithmetic instead of `torch.rad2deg` / `torch.deg2rad` to preserve ONNX export compatibility.
- Update the existing #3937 regression tests.
- Remove obsolete warnings related to the fixed behavior.
- Update `kornia.feature.siftdesc` to import `pi` directly from `kornia.constants`.
- Update half-precision known-failure lists because the affected tests now pass.

## Testing

- `pytest tests/geometry/test_conversions.py -k "3937" -q`
  - 6 passed

- `pytest tests/geometry/test_conversions.py -q`
  - 393 passed, 1 skipped, 8 xfailed

- `pytest tests/augmentation/test_onnx_export.py -k "RandomAffine or RandomRotation" -q`
  - 5 passed

- `pixi run -e default lint`
  - ruff check passed
  - ruff format passed

- GitHub Actions
  - 39 checks passed


ducha-aiki (commented):
The conversion implementation and its regression coverage look correct, and this is the preferable version of the overlapping #4356 fix. Two documentation edits remain before merge: update the comments at `tests/geometry/test_conversions.py:2554-2571`, which still say all three operations are broken and marked strict xfail even though this patch removes that mark and fixes the outputs; and add the required `CHANGELOG.md` bug-fix entry citing #4358.

Validation at `eabe5de5f0d96efd68c12a5d9c287d84dd5a244b`: `import kornia` resolves to the PR worktree; 620 CPU float32/float64 conversion tests passed (2 skipped, 12 expected xfails); all 39 hosted checks are green. `pi` was never in `conversions.__all__` or the API-surface inventory, and the in-tree SIFT consumer now imports it directly from `kornia.constants`. No prior comments or formal reviews exist.

Hrithikcrick:
Addressed the requested follow-up changes: updated the stale #3937 regression comments and added the CHANGELOG.md bug-fix entry for #4358. Verified with the #3937 tests and lint; all pass.

ducha-aiki (dismissed):
Delta re-review at `8576fec8`.

Both requested follow-ups are resolved: the #3937 regression comments now describe the repaired behavior, and `CHANGELOG.md` has the required Bug fixes entry citing #4358. The implementation remains the previously reviewed angle-conversion fix; the merge from upstream places it on current `main` without conflicts.

Validation at this exact head: 620 CPU float32/float64 conversion tests passed (2 skipped, 12 expected xfails), `git diff --check` is clean, all 40 hosted checks are green, and `import kornia` resolved to the isolated PR worktree.

PR #2358, opened 2023-04-29
Update ruff configs

johnnv1:
with the objective of having a clear/minimal config for ruff, this patch cleans up some configs and enables some new rules

Rules activated:
- C4 (flake8-comprehensions) - https://pypi.org/project/flake8-comprehensions/
- BLE (blind except) - https://pypi.org/project/flake8-blind-except/
- RUF (ruff specifics rules) - https://beta.ruff.rs/docs/rules/#ruff-specific-rules-ruf

Rules to be considered (#1227):
- B (bugbear): https://beta.ruff.rs/docs/rules/#flake8-bugbear-b
- ISC (flake8-implicit-str-concat): https://beta.ruff.rs/docs/rules/#flake8-implicit-str-concat-isc
- RET (flake8-return): https://beta.ruff.rs/docs/rules/#flake8-return-ret
- SIM (simplify): https://beta.ruff.rs/docs/rules/#flake8-simplify-sim
- TCH (flake8-type-checking): https://beta.ruff.rs/docs/rules/#flake8-type-checking-tch
- PGH (pygrep-hooks): https://beta.ruff.rs/docs/rules/#pygrep-hooks-pgh

PR #1218, opened 2021-08-23
LoFTR inference basic PR [WIP]

ducha-aiki:
### Description
This is very basic adaptation of the LoFTR (https://github.com/zju3dv/LoFTR)  for kornia.
Things done: keep only inference part, remove dependency on einops and config library, download weights via torch.hub

Missing things:

- code restructuring to a single file
- using kornia operations whenever possible
- tests
- code style fix

### Status
**Work in progress*

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or new feature that would cause existing functionality to change)
- [ ] New tests added to cover the changes
- [ ] Docstrings/Documentation updated

### PR Implementer
This is a small checklist for the implementation details of this PR.

If there are any questions regarding code style or other conventions check out our
[summary](https://github.com/kornia/kornia/blob/master/CONTRIBUTING.rst).

- [ ] Did you discuss the functionality or any breaking changes before ?
- [ ] **Pass all tests**: did you test in local ? `make test`
- [ ] Unittests: did you add tests for your new functionality ?
- [ ] Documentations: did you build documentation ? `make build-docs`
- [ ] Implementation: is your code well commented and follow conventions ? `make lint`
- [ ] Docstrings & Typing: has your code documentation and typing ? `make mypy`
- [ ] Update notebooks & documentation if necessary

### KorniaTeam
<details>
  <summary>KorniaTeam workflow</summary>

  - [ ] Assign correct label
  - [ ] Assign PR to a reviewer
  - [ ] Does this PR close an Issue? (add `closes #IssueNumber` at the bottom if
        not already in description)

</details>

### Reviewer
<details>
  <summary>Reviewer workflow</summary>

  - [ ] Do all tests pass? (Unittests, Typing, Linting, Documentation, Environment)
  - [ ] Does the implementation follow `kornia` design conventions?
  - [ ] Is the documentation complete enough ?
  - [ ] Are the tests covering simple and corner cases ?

</details>

PR #4359, opened 2026-09-08
fix(feature): CoarseMatching training-mode gt-padding crash + dtype leak

tritsystem:
## Summary

`CoarseMatching.get_coarse_match`'s training-only ground-truth-padding branch (`self.training=True`) crashed unconditionally: `zip()` was called with 4 separate 2-element lists (`[b_ids, spv_b_ids]`, `[i_ids, spv_i_ids]`, `[j_ids, spv_j_ids]`, `[mconf, mconf_gt]`) instead of 2 separate 4-element lists, so each resulting 4-tuple failed to unpack into the loop's two variables (`x, y`) — `ValueError: too many values to unpack (expected 2)` on every single call. No existing test exercised this path at all (grepped the whole suite for `CoarseMatching`/`get_coarse_match`: zero hits), which is why it went uncaught.

Restructured the `zip()` to pair each real tensor with its ground-truth counterpart, matching what the surrounding code clearly intends. Once reachable, a second real bug surfaces: `mconf_gt = torch.zeros(..., device=_device)` has no `dtype=`, so it defaults to `float32` regardless of `mconf`'s own dtype (sliced straight from `conf_matrix`, which can be `float16`/`bfloat16` under autocast) — the `torch.cat` that follows silently type-promotes the whole padded `mconf` to `float32`. Fixed by building `mconf_gt` in `mconf.dtype`.

## Test plan

- [x] Added `TestCoarseMatching` with tests exercising the training-mode gt-padding branch directly (no pretrained weights needed)
- [x] Verified red-then-green: reverting just the source fix makes 3/4 new tests fail with the exact `ValueError` above; restoring it passes all 4, parametrized over `float32`/`float16`/`bfloat16`
- [x] Added a dedicated eval-mode test confirming that path (which never touches this branch) is unaffected
- [x] `ruff check` clean


ducha-aiki (commented):
Reviewed at `67201ceddc1423d811394a7e82f8ac314fb86298` — the implementation and regression coverage look correct; one repository-contract edit and CI completion remain.

The training branch now zips the four prediction tensors against the four ground-truth tensors, fixing the unconditional `ValueError` from unpacking a four-element tuple into `(x, y)`. Creating `mconf_gt` in `mconf.dtype` also prevents the concatenation from widening half-precision confidences. I ran `tests/feature/test_loftr.py -k CoarseMatching` from the PR worktree with the repository Python: 4 passed, and the import resolved to this exact worktree.

Please add a `CHANGELOG.md` entry under `Unreleased` → `Bug fixes` citing `#4359`, as required for this user-visible fix. The float16 CI job is also still running at this snapshot; merge should wait for it to finish successfully. No code changes are otherwise requested.

ducha-aiki:
Reviewed at `d7835ccd`. `origin/main` (`8c71df9a0`) is already an ancestor of the head, so the merge tree is the head; nothing on `main` has made any part of this redundant (`coarse_matching.py` last changed in #3432, 2025-12-31; `tests/feature/test_loftr.py` in #4135).

**The production fix is right.** I verified all of the following.

**1. The crash reproduces on `main` and is gone here.** Minimal call, run from each worktree root with `kornia.__file__` asserted:

```python
cfg = {"thr": 0.01, "border_rm": 0, "train_coarse_percent": 0.5,
       "train_pad_num_gt_min": 2, "match_type": "dual_softmax", "dsmax_temperature": 0.1}
d = {"hw0_c": (4, 4), "hw1_c": (4, 4), "hw0_i": (32, 32), "hw1_i": (32, 32),
     "spv_b_ids": torch.zeros(20, dtype=torch.long),
     "spv_i_ids": torch.randint(0, 16, (20,)), "spv_j_ids": torch.randint(0, 16, (20,))}
cm = CoarseMatching(cfg); cm.train()
cm.get_coarse_match(torch.rand(1, 16, 16), d)
```

- base: `ValueError: too many values to unpack (expected 2)` — on torch **2.14.0 and on the 2.5.1 floor** (checked with a `--target` overlay).
- merge tree: returns `M' = 8` matches (6 sampled predictions + 2 gt pads), `gt_mask.sum() == 2`, on both torch versions.

**2. The restructured `zip()` is byte-identical to upstream LoFTR.** Both expressions on the same tensors:

```
n outputs: 4 4
b identical: True   i identical: True   j identical: True   mconf identical: True (float16 both)
```

Git archaeology confirms the provenance: the correct `map(lambda x, y: ..., *zip([b_ids, spv_b_ids], [i_ids, spv_i_ids], ...))` form landed in #1218 and survived #1777; **#2358 ("Update ruff configs", 2023-07-09) rewrote it as a generator expression and dropped the `*` that did the transpose**. So this has been dead for ~2 years, which matches "zero hits for `CoarseMatching`/`get_coarse_match` in `tests/` + `testing/`" — re-grepped on base, 0.

**3. The dtype leak is real and the fix is minimal.** A/B of every returned tensor's dtype, with and without `dtype=mconf.dtype`, over float64/float32/bfloat16/float16:

| input dtype | `mconf` without fix | `mconf` with fix | other 7 outputs |
|---|---|---|---|
| float64 | float64 | float64 | unchanged |
| float32 | float32 | float32 | unchanged |
| bfloat16 | **float32** | bfloat16 | unchanged |
| float16 | **float32** | float16 | unchanged |

Exactly 2 of 32 measured output dtypes move, and only `mconf` — `mkpts0_c`/`mkpts1_c` were already pinned by the trailing `.to(dtype=conf_matrix.dtype)`. The leak is also bidirectional under a non-default default dtype: with `torch.set_default_dtype(torch.float64)` and a **float32** `conf_matrix`, base+zip-fix returns `mconf` as **float64**; with the fix it stays float32. Device handling is correct too — the fixed branch on MPS (`env -u PYTORCH_ENABLE_MPS_FALLBACK`) for float32 and float16 gives `mconf.device == mps:0`, `M' = 8` in both.

**4. Regression sweep.** `tests/feature --device=cpu --dtype=float32`: base `738 passed, 97 skipped`; merge tree `742 passed, 97 skipped`. Exactly +4, zero newly-red nodes. `test_loftr.py` also passes at `--dtype=float16` and `--dtype=bfloat16` (4 each), and under the 2.5.1 overlay. `uvx ruff@0.16.7 check` / `format --diff` clean. Fragment type `fixed` is correct; no issue referenced, so no prose blast radius. **Dynamo:** `torch._dynamo.explain(cm.get_coarse_match)` on the merge tree reports `graph_count=5, graph_break_count=4`, all 4 from the pre-existing `aten.nonzero.default` at `torch.where(mask_v)` line 236 — the diff adds no new `.item()`, no new `len()` (the `len(data["spv_b_ids"])` was already there) and no new data-dependent construct.

## Findings

**[P2] The new tests do not pin the fix. 6 of 8 one-token mutants survive, including removing the training branch entirely.**

Mutating `kornia/feature/loftr/utils/coarse_matching.py` and re-running `pytest tests/feature/test_loftr.py --device=cpu --dtype=float32` (baseline `4 passed, 7 skipped`):

| # | mutation | result | what actually changes |
|---|---|---|---|
| M1 | revert the `zip()` fix to the 4x2 form | **3 failed** killed | crash |
| M2 | drop `dtype=mconf.dtype` | **2 failed** killed | `mconf` float16/bf16 -> float32 |
| M3 | swap `spv_b_ids` / `spv_i_ids` in the gt list | **4 passed** SURVIVES | `b_ids = [0,0,0,0,0,0,8,6]` for a batch of **1** — out-of-range batch indices that then index `data["scale0"][b_ids]` and the fine-level unfold |
| M4 | swap `spv_i_ids` / `spv_j_ids` | **4 passed** SURVIVES | `i_ids` tail `[8,6]` -> `[4,6]`; source/target coarse indices transposed for every gt pad |
| M5 | `torch.cat([y[gt_pad_indices], x[pred_indices]])` (gt first) | **4 passed** SURVIVES | `gt_mask` moves from the tail to the head |
| M6 | `y[pred_indices]` instead of `y[gt_pad_indices]` | **4 passed** SURVIVES | `M' = 12` instead of `8`, `gt_mask.sum() = 6` instead of `2` |
| M7 | `torch.zeros` -> `torch.ones` for `mconf_gt` | **4 passed** SURVIVES | `gt_mask.sum() = 0`; `len(mconf) = 8` instead of `6` — the gt pads stop being filtered out |
| M8 | `if self.training:` -> `if False:` | **4 passed** SURVIVES | **no gt padding at all**: `M' = 9`, `gt_mask.sum() = 0`, `len(mconf) = 9` |

Every one of the six is non-equivalent, with the numbers above. M8 is the diagnostic one: the whole hunk this PR touches can be deleted and all 4 tests still pass. The reason is that the only real assertion —

```python
assert out["mconf"].dtype == dtype
```

— holds for a reason independent of the branch: `mconf` is sliced from `conf_matrix`, so if the branch never runs the dtype is trivially right. And `assert out["b_ids"].shape == out["i_ids"].shape == out["j_ids"].shape` is a tautology — those three are built from the same index tensors on every path. So M1 is killed only by the crash, M2 only by the dtype, and nothing pins the *pairing*, which is precisely what the fix changes.

A version that kills **8/8** (verified) needs (a) an asymmetric coarse grid so `i` and `j` index ranges are distinguishable, (b) a deterministic `conf_matrix` so the pred/pad split is known rather than drawn, and (c) positional assertions on the gt tail:

```python
class TestCoarseMatching:
    CFG = {"thr": 0.01, "border_rm": 0, "train_coarse_percent": 0.5,
           "train_pad_num_gt_min": 2, "match_type": "dual_softmax", "dsmax_temperature": 0.1}
    H0C, W0C, H1C, W1C = 4, 4, 3, 4   # L = 16, S = 12 -> i/j ranges are distinguishable
    N_PRED, N_GT = 5, 20

    def _data(self, device):
        L, S = self.H0C * self.W0C, self.H1C * self.W1C
        return {
            "hw0_c": (self.H0C, self.W0C), "hw1_c": (self.H1C, self.W1C),
            "hw0_i": (self.H0C * 8, self.W0C * 8), "hw1_i": (self.H1C * 8, self.W1C * 8),
            "spv_b_ids": torch.zeros(self.N_GT, dtype=torch.long, device=device),
            "spv_i_ids": torch.full((self.N_GT,), L - 1, dtype=torch.long, device=device),
            "spv_j_ids": torch.full((self.N_GT,), S - 1, dtype=torch.long, device=device),
        }

    @pytest.mark.parametrize("dtype", [torch.float32, torch.float16, torch.bfloat16])
    def test_training_gt_padding(self, device, dtype):
        L, S = self.H0C * self.W0C, self.H1C * self.W1C
        conf = torch.zeros(1, L, S, device=device, dtype=dtype)
        for k in range(self.N_PRED):            # exactly N_PRED mutual-NN matches, no RNG
            conf[0, k, k] = 0.5 + 0.01 * k
        cm = CoarseMatching(self.CFG).to(device)
        cm.train()
        out = cm.get_coarse_match(conf, self._data(device))

        n_train = int(max(L, S) * self.CFG["train_coarse_percent"])            # 8
        n_pad = max(n_train - self.N_PRED, self.CFG["train_pad_num_gt_min"])   # 3
        assert out["b_ids"].shape[0] == self.N_PRED + n_pad                    # kills M6, M8
        assert (out["b_ids"] == 0).all()                                       # kills M3
        assert (out["i_ids"] < L).all() and (out["j_ids"] < S).all()           # kills M3, M4
        assert out["i_ids"][: self.N_PRED].tolist() == list(range(self.N_PRED))# kills M5
        assert out["j_ids"][: self.N_PRED].tolist() == list(range(self.N_PRED))
        assert (out["i_ids"][self.N_PRED :] == L - 1).all()                    # kills M4, M5
        assert (out["j_ids"][self.N_PRED :] == S - 1).all()
        assert not out["gt_mask"][: self.N_PRED].any()                         # kills M5, M7
        assert out["gt_mask"][self.N_PRED :].all()
        assert out["mconf"].numel() == self.N_PRED                             # kills M7
        assert out["mconf"].dtype == dtype                                     # kills M2
        assert out["mkpts0_c"].shape == (self.N_PRED, 2)
```

Measured against the same eight mutants (baseline `3 passed`): M1 3 failed, M2 2 failed, M3 3 failed, M4 3 failed, M5 3 failed, M6 3 failed, M7 3 failed, M8 3 failed — **8/8 killed** — and it passes unmodified on the merge tree (3 passed; also 3 passed under the 2.5.1 overlay). It also removes the `torch.manual_seed` below.

**[P3] `test_eval_path_unaffected` cannot fail.** The entire hunk is inside `if self.training:`, and the test only asserts `out["mconf"].dtype == torch.float32` for a float32 input. All eight mutants leave it green, including M8. Either drop it or make it assert something the branch would break — `gt_mask.sum() == 0` and `M' == num_matches_pred` — so it documents "eval takes no gt padding".

**[P3] The test class is outside the repo's test convention, so the fixed branch never runs on an accelerator.** `AGENTS.md` asks for `BaseTester` from `testing.base` and the injected `device`/`dtype` fixtures; `TestCoarseMatching` uses neither (1 of only 3 non-`BaseTester` `class Test*` in `tests/feature/`, out of 101). Under `--device=cuda` or the blocking `tests-mps` leg the test silently runs on CPU, and it always runs its own hardcoded float16/bfloat16 parametrization regardless of `--dtype`. That matters here specifically because this is a *device-and-dtype* fix. I confirmed by hand that the fixed branch is correct on MPS; CI will never confirm it. Taking `device` from the fixture costs nothing.

**[P3] `torch.manual_seed(0)` without `restore_torch_rng`.** The new tests reseed the global CPU generator and do not restore it, shifting draws in every test file collected afterwards. `conftest.py` has the opt-in `restore_torch_rng` fixture for exactly this (#4446). Bare `manual_seed` is widespread (854 occurrences in `tests/`) so this is not a regression in kind — but the branch can be exercised with no RNG at all (snippet above), which is strictly better.

**[P3] The new comment hardcodes a line number.** `"sliced straight from conf_matrix (line 237)"` — correct today, guaranteed to rot. The fact carries the whole argument without the number; four lines of comment for a one-token change is also more than this needs.

## What is left to do

Replace the assertions in `TestCoarseMatching` with ones that fail when the pairing, the ordering, the pad count or the zero-confidence convention is wrong (8/8 mutants killed, snippet above, verified), and take `device` from the fixture. The production diff itself I would merge as-is — correct, minimal, upstream-faithful, green on all 44 CI legs including `dynamo`, `tests-mps` and both half-precision legs, and it changes nothing outside the training branch (738 -> 742 passed in `tests/feature`, no node newly red). Nice piece of archaeology finding a two-year-old ruff rewrite as the cause.

ducha-aiki (changes requested):
Requesting changes for the remaining test-coverage finding in the existing review: assert the ground-truth pairing, ordering, padding count and zero-confidence behavior, and use the device fixture. The production fix itself was assessed as correct.

Recording the verdict from the [existing review comment](https://github.com/kornia/kornia/pull/4359#issuecomment-5711867791) as a formal GitHub review on `d7835ccd087c737feddd3f2090de31273613b06d`. This records the existing assessment; it is not a new test run.

ducha-aiki (approved):
Review of `1c3a40eeabd5a3e47c2c208eef20c21d0149e3d8`.

The previous test-coverage blocker is resolved. The regression now uses a deterministic asymmetric grid and asserts the prediction/ground-truth pairing, ordering, pad count, zero-confidence convention, dtype, and eval behavior. It also inherits `BaseTester` and uses the injected device fixture, so the affected branch runs on accelerator jobs. The production change remains the minimal upstream-faithful repair: it zips the four prediction tensors with their four ground-truth counterparts and creates `mconf_gt` in `mconf.dtype`.

Validation on the exact head, with imports resolved to the `/tmp/kornia-pr4359` worktree: `tests/feature/test_loftr.py -k CoarseMatching` passed all 4 cases on CPU and all 4 on MPS; the parameterized training test covered float32, float16, and bfloat16 on each device. The merge tree against current `origin/main` is conflict-free, and every hosted check is green. No full local suite or CUDA run was performed.

PR #4360, opened 2026-09-08
fix(enhance): image_histogram2d bin centers built at image.dtype

tritsystem:
## Summary

Same bug class as the already-merged bbox_to_mask fix (#4336). `image_histogram2d` built its auto bin centers with `torch.arange(n_bins, dtype=image.dtype)`, and a half-precision dtype cannot represent the bin indices exactly, so distinct bins landed on the same center and returned numerically identical output. Measured on the old construction (`min=0, max=1`): `bfloat16` loses 63 of 256 centers at the default `n_bins`, and `float16` loses 511 of 2048 and 1535 of 4096.

The centers are now built at `torch.promote_types(image.dtype, torch.float32)`: `float32` for `float16`/`bfloat16` images, and a `float64` image keeps `float64` (a hard-coded `float32` would downgrade it). The result is cast back to the image's dtype only when the centers were auto-built and the image is floating point. Integer images and an explicitly passed `centers` behave exactly as they did before this PR; a reviewer found that an earlier version of this PR broke both, so there are tests for each.

Not done, on purpose: `pdf` is still normalized after `hist` is cast to the image dtype, as on `main`. Normalizing before the cast would fix the all-zero `pdf` a large `float16` image can give (which `main` also gives), but it changes half-precision `pdf` values and is a separate change.

## Test plan

- [x] `test_large_n_bins_float16_centers_not_collapsed`: auto centers against explicitly supplied `float32` centers on the same `float16` image.
- [x] `test_auto_centers_preserve_input_dtype`: `float16`/`bfloat16` images still return their own dtype (`hist` and `pdf`).
- [x] New, each failing on the previous head of this PR: `test_integer_image_is_not_cast_into_an_integer_histogram` (a `uint8` image whose bin counts exceed 255 gave `6` where `main` gives `262.645`), `test_float64_image_keeps_float64_bin_centers` (`n_bins=1000` gives a non-power-of-two bandwidth, where float32 centers move the histogram by about 1e-4), `test_explicit_centers_keep_the_promoted_dtype` (float32 image with float64 centers stays float64).
- [x] Production mutants killed by that file: centers back to `image.dtype`, hard-coded `float32` centers, the `is_floating_point` guard dropped, the `auto_centers` guard dropped, and the cast dropped.
- [x] `tests/enhance` and `tests/test_half_precision_ci.py`, CPU: `float32` 490 passed / 16 skipped, `float64` 490 / 16, `float16` 487 / 19, `bfloat16` 487 passed / 17 skipped / 2 failed. Those 2 (`test_jpeg::test_cardinality`, `test_zca::test_dim_args`) fail on unmodified `main` too. `torch.jit.script(image_histogram2d)` still compiles. `ruff check` and `ruff format --check` clean.
- [ ] Not tested: MPS (no Apple hardware here), so the `bfloat16`-on-MPS skip is only relaxed for torch >= 2.14 on the reviewer's report and the blocking MPS job will show whether that holds; GPU.


ducha-aiki (changes requested):
[P1] Preserve the auto-center output dtype. The float32 centers introduced at `kornia/enhance/histogram.py:248` promote `u`, `kernel_values`, and therefore both public return tensors to float32. Before this patch, a float16 image with `centers=None` returned float16 `hist`/`pdf`, and bfloat16 returned bfloat16; this head returns float32 for both dtypes and for both `return_pdf` paths. Keep the internal calculation in float32, then restore the established image dtype at the public return boundary, and add a dtype assertion so the regression is covered.

[P1] Make the new reference test runnable on MPS. `tests/enhance/test_histogram.py:119` constructs the explicit centers as float64 on the injected device. MPS does not support float64, and the hosted MPS job fails this exact test with `TypeError: Cannot convert a MPS Tensor to float64 dtype`. Use an MPS-supported reference construction or an explicit platform skip while retaining the numerical pin.

This user-visible numerical fix also needs a `CHANGELOG.md` bug-fix entry citing #4360.

Validation at `d0b919639b328d5fc07a3d32071e6796e64c6c7d`: 103 CPU float32/float64 and 103 CPU float16/bfloat16 histogram tests passed; a direct base/head probe confirmed the half/bfloat return-dtype regression. Hosted CI is otherwise green, with MPS and its collector red from the test above. No prior comments or formal reviews exist.

ducha-aiki:
Reviewed at `9e15dbdd`. `origin/main` (`8c71df9a0`) is already an ancestor of the head, so the merge tree is the head; nothing to reconcile. All 44 CI checks green, `ruff@0.16.7 check` / `format --diff` clean on both touched files, no line over 120 chars, no public names added, and the PR closes no issue, so there is no issue-prose blast radius to sweep.

## What I verified

**The defect is real, and larger than the PR states.** The PR's story is "`float16` can't represent integers past 2048". That is half of it: `min + bandwidth * (arange + 0.5)` runs the *multiply* at `image.dtype` too, so the collapse starts much earlier. Distinct auto-constructed centers lost, `min=0.0, max=1.0`, on `main`:

| n_bins | 256 | 512 | 1024 | 2048 | 4096 |
|---|---|---|---|---|---|
| float16 | 0 | 0 | 0 | **511** | **1535** |
| bfloat16 | **63** | **191** | **575** | **1471** | **3391** |

bfloat16 already collapses **63 of 256** centers at the *default* `n_bins=256`. End to end with default args (`min=0, max=255, n_bins=256`, gaussian, 32x32 image), against a float64 reference: `main` bfloat16 max abs hist error **4.075** with **65** adjacent bin pairs numerically identical; this PR **0.0493** with **2**. float16: **0.375 -> 0.00654**.

**Red-then-green.** The PR's test file against `main`'s production code: `test_large_n_bins_float16_centers_not_collapsed` fails with **3058 / 4096 bins (74.7%) mismatched, greatest absolute difference 4.464** (the PR body's "~5.99" does not reproduce on torch 2.14 CPU with `manual_seed(0)`).

**float32/float64 A/B sweep**, 48 configs (2 dtypes x 6 `(min,max,n_bins)` including non-dyadic bandwidths x 4 kernels), comparing `hist` and `pdf` bit-for-bit: **36/48 identical**. All 24 float32 configs are bit-identical. All 12 that move are float64 — see [P1] #2.

**Mutation testing**, `tests/enhance/test_histogram.py` at `--dtype=float32` and `--dtype=float16` (unmutated baseline 59 passed):

| mutant | result |
|---|---|
| M1 `dtype=torch.float32` -> `dtype=image.dtype` (revert the core fix) | **killed** |
| M2 bin-center offset `+ 0.5` -> `+ 0.0` | **killed** |
| M3 bin-center offset `+ 0.5` -> `+ 1.0` (half-bin shift) | **killed** |
| M4 drop `hist = hist.to(image.dtype)` | **killed**, 4 failures |
| M5 `hist.to(image.dtype)` -> `hist.to(torch.float32)` | **killed**, 7 failures |
| M6 `dtype=torch.float32` -> `dtype=torch.float64` | survives — but that mutant is an *improvement*, see [P1] #2 |

**Regression sweep**, whole `tests/enhance/`, CPU, base vs merge tree:

| dtype | base | merge tree |
|---|---|---|
| float32 | 377 passed, 15 skipped | 382 passed, 15 skipped |
| float64 | 377 passed, 15 skipped | 382 passed, 15 skipped |
| float16 | 375 passed, 17 skipped | 380 passed, 17 skipped |
| bfloat16 | 2 failed, 375 passed | 2 failed, 380 passed |

The 2 bfloat16 failures (`test_jpeg.py::TestDiffJPEG::test_cardinality`, `test_zca.py::TestZCA::test_dim_args[...-1]`) are identical on both trees and covered by `cpu_bfloat16.txt`; with `--xfail-known-failures --known-failure-profile=cpu-bfloat16` the merge tree is `380 passed, 2 xfailed`. **No manifest change is needed** — `grep -n histogram testing/half_precision_xfails/*.txt testing/known_failure_xfails/*.txt` is empty on both trees, and the PR correctly touches neither.

**MPS (blocking leg).** `tests/enhance/test_histogram.py --device=mps --dtype=float32`, `PYTORCH_ENABLE_MPS_FALLBACK` unset: base 48 passed / 6 skipped, merge tree **51 passed / 8 skipped, 0 failures**.

## Findings

**[P1] `hist = hist.to(image.dtype)` silently truncates *and integer-wraps* the histogram for an integer-dtype image.** New regression from the second commit, not present on `main`. `image_histogram2d` has no dtype guard and its `max: float = 255.0` default invites `uint8` input; on `main` an integer image promotes through `arange + 0.5` to float32 and the KDE values come back as floats. Now they are cast back down:

```python
img = (torch.rand(1, 64, 64, generator=torch.Generator().manual_seed(3)) * 255).to(torch.uint8)
hist, _ = kornia.enhance.image_histogram2d(img, 0.0, 255.0, 16)
# main: torch.float32  [239.645, 262.645, 268.665, 243.669, 254.061, 238.208, 235.804, 267.525]
# PR:   torch.uint8    [239,      6,       12,      243,     254,     238,     235,     11    ]
```

`262.645 -> 6` and `268.665 -> 12` are mod-256 wraps — not rounding, silent corruption. `int32` input is merely truncated (`4.254 -> 4`), and `return_pdf=True` then normalizes the truncated counts, so the pdf moves too (`0.0581 -> 0.0536`, `0.0795 -> 0.0893` on a `1x8x8` uint8 image, `n_bins=16`). Guarding the cast with `image.is_floating_point()` restores byte-identical `main` behaviour here (verified).

**[P1] The fix loses float64 bin-center precision — the opposite of its stated purpose.** `min` and `bandwidth` are Python floats, which are wrapped numbers and do not promote, so `min + bandwidth * (torch.arange(n_bins, dtype=torch.float32) + 0.5)` yields a **float32** `centers` for *every* input dtype, float64 included. On `main` a float64 image got float64 centers:

```
n_bins=100:  max |centers_PR - centers_main| = 4.77e-08
n_bins=1000: max |centers_PR - centers_main| = 7.63e-08
```

That propagates: **12 of the 48** A/B configs move, all float64 with a non-dyadic bandwidth (the power-of-two `n_bins` cases are identical only by accident — `1/256` is exact in both):

| config (float64) | max abs dhist | max rel dhist |
|---|---|---|
| `min=0, max=1, n_bins=1000`, triangular | 1.90e-04 | **7.25%** |
| `min=0, max=1, n_bins=1000`, epanechnikov | 2.30e-04 | **7.25%** |
| `min=0, max=255, n_bins=300`, epanechnikov | 7.51e-05 | 0.145% |
| `min=0, max=255, n_bins=300`, triangular | 5.74e-05 | 0.146% |

Nothing pins this — which is exactly why M6 survived. The one-token fix is a compute dtype that is a *promotion*, not a constant:

```python
centers = min + bandwidth * (
    torch.arange(n_bins, device=image.device, dtype=torch.promote_types(image.dtype, torch.float32)) + 0.5
)
```

`promote_types(float16, float32) == float32`, `promote_types(float64, float32) == float64`, `promote_types(uint8, float32) == float32`. I executed this: all 59 `test_histogram.py` nodes stay green at float32/float64/float16/bfloat16, the float64 values are restored bit-for-bit, and `torch.jit.script(kornia.enhance.image_histogram2d)` still compiles — this module is JIT-scriptable and has no `from __future__ import annotations`, so that mattered.

**[P2] The changelog fragment's central claim is false, and the return-dtype change it hides is user-visible and incompatible.** The fragment says the outputs are cast back "so this precision fix does not itself change the function's return dtype". For the `centers=None` float16/bfloat16 path that is true. But the cast is unconditional, so it also fires on the **explicit-`centers=` path**, where `main` deliberately returned the *promoted* dtype:

```python
img = torch.rand(1, 8, 8, dtype=torch.float32)
c   = torch.linspace(0, 1, 16, dtype=torch.float64)
hist, pdf = kornia.enhance.image_histogram2d(img, 0.0, 1.0, 16, centers=c, return_pdf=True)
# main: hist.dtype=torch.float64  pdf.dtype=torch.float64
# PR:   hist.dtype=torch.float32  pdf.dtype=torch.float32
```

A user who supplies float64 centers to buy precision now gets it thrown away at the return boundary. With the uint8/int32 case above, the return dtype *does* change for two input classes, both undocumented and untested. Either scope the cast to the auto-constructed-centers, floating-point case (so the fragment's claim becomes true and `fixed` is right), or keep the unconditional cast and file a **`breaking`** fragment stating old and new. As written, `4360.fixed.md` is the wrong type for what the code does.

**[P3] The new test's docstring states a number the second commit falsified.** It says "fixed, ~9e-7 (ordinary float32 rounding noise)". Once `hist = hist.to(image.dtype)` landed, both sides are cast to float16 and the computation is bit-identical, so the measured `max |hist_auto - hist_explicit|` is **exactly 0.0**. The "unfixed ~5.99" is also off — I measure **4.464**. Re-measure or drop the numbers.

**[P3] The MPS bfloat16 skip is unnecessary on the torch the blocking MPS leg pins.** `tests-mps` runs torch 2.14.0, and `image_histogram2d` on a bfloat16 MPS tensor works there (`hist.dtype == torch.bfloat16`, all finite, verified locally with `PYTORCH_ENABLE_MPS_FALLBACK` unset). The `pytest.skip("bfloat16 has limited/no MPS support")` drops half the new dtype coverage on the only MPS leg. If it is needed for the 2.5.1 floor, gate it on `torch_version_ge` and say so.

**[P3] Casting before the pdf normalization forecloses a free win.** `hist` is cast to `image.dtype` *before* `normalization = sum(hist) + eps`, so the pdf is normalized in low precision. For a 384x384 float16 image with `n_bins=8`, `kernel="uniform"`, the counts exceed float16's range, `normalization` becomes `inf`, and `pdf == [0.0] * 8` — on `main` **and** on this PR, so not a regression. Moving the cast after the normalization (executed) gives `pdf == [0.0995, 0.1331, 0.1340, 0.1338, 0.1335, 0.1335, 0.1332, 0.0992]` and keeps every existing test green. Since this PR is already moving that cast, it is the natural place.

**[P3] Twelve lines of first-person narrative comment in library code.** Two sentences — what the compute dtype is and why it is not `image.dtype` — carry the same information.

## Suggested shape

Ran end to end; passes `tests/enhance/` at float32 / float64 / float16 / bfloat16 with no new failures and keeps `torch.jit.script` working:

```python
    if centers is None:
        # Build the bin-center grid at a dtype that can represent the bin indices exactly.
        # image.dtype alone collapses distinct centers onto one value (bfloat16 loses 63 of 256
        # centers at the default n_bins; float16 loses 1535 of 4096); a hardcoded float32 would
        # instead downgrade a float64 image. promote_types does both.
        compute_dtype = torch.promote_types(image.dtype, torch.float32)
        centers = min + bandwidth * (torch.arange(n_bins, device=image.device, dtype=compute_dtype) + 0.5)
    ...
    if return_pdf:
        normalization = torch.sum(hist, dim=-1, keepdim=True) + eps
        pdf = hist / normalization
        if image.is_floating_point():          # never cast a KDE result into an integer dtype
            hist = hist.to(image.dtype)
            pdf = pdf.to(image.dtype)
```

## What is left to do

1. Stop the return cast firing on integer `image.dtype` and add a pin — a `uint8` image whose bin counts exceed 255 is the discriminating input; `main` gives `262.645`, this PR gives `6`.
2. Use `torch.promote_types(image.dtype, torch.float32)` for the centers and add a float64 pin with a non-dyadic bandwidth (`min=0.0, max=1.0, n_bins=1000`, triangular) — a float32-centers implementation is 7.25% off there, so the pin discriminates. That is also what kills the currently-surviving M6.
3. Decide the explicit-`centers=` contract and make the fragment match the code. Pin whichever you pick.
4. Fix or remove the `~9e-7` / `~5.99` numbers in the test docstring and the PR body.
5. Drop or version-gate the MPS bfloat16 skip; trim the comment block; optionally move the cast past the pdf normalization.

1–3 block; 4–5 are nits. Everything else — the diagnosis, the mutant-killing pins, the manifest hygiene, the MPS leg — is in good shape, and the underlying bug is well worth fixing.

tritsystem:
Thank you, this was a careful review and the two P1s and the P2 were all right. I reproduced each on the previous head (`9e15dbd`) before changing anything:

- **uint8 wrap:** `main` gives `float32 [239.645, 262.645, 268.665, ...]`, that head gave `uint8 [239, 6, 12, ...]`. `int32` came back as `int32`.
- **float64:** the float64 histogram sums moved off `main` (for example `1024.0000000000` became `1023.9999096277` at `n_bins=1000`, triangular). The power-of-two `n_bins=256` case was unchanged, as you said.
- **Explicit centers:** a float32 image with float64 `centers=` returned float64 on `main` and float32 on that head.

Pushed c5989b5 (also merges current `main`):

1. The centers use `torch.promote_types(image.dtype, torch.float32)`, and the cast back to the image dtype now applies only to auto-built centers on a floating-point image. Integer images and explicit `centers` are byte-identical to `main` again (checked with the same probes). `torch.jit.script` still compiles.
2. Three new tests, each of which fails on `9e15dbd` (and passes here): the uint8 case (bin counts above 255), the float64 case with `n_bins=1000`, and the explicit-centers dtype. I also ran production mutants against the whole file: centers back to `image.dtype`, hard-coded float32, the `is_floating_point` guard dropped, the `auto_centers` guard dropped, and the cast dropped are all killed. The float64 mutant you noted as surviving is now killed too.
3. The fragment is rewritten to match the code and stays a `fixed` fragment, since integers and explicit `centers` no longer change. It now also says `bfloat16` loses 63 of 256 centers at the default `n_bins`; I re-counted that and your table on the old construction and both reproduce.
4. The `~9e-7` and `~5.99` figures are gone from the test docstring and the PR description (the description had gone stale in other ways too and is rewritten). The 12-line comment is now 3 lines plus a small helper with a docstring, which also keeps `image_histogram2d` under ruff's complexity limit.
5. The MPS `bfloat16` skip is now `torch_version_lt(2, 14, 0)`. I have no Apple hardware, so I am relying on your check that it works on 2.14; the blocking MPS job will confirm it.

Not adopted: moving the cast after the `pdf` normalization. The all-zero `pdf` on a large `float16` image also happens on `main`, so it is not a regression here, and the change alters half-precision `pdf` values, so I would rather it be a separate PR than ride along with a dtype fix. Happy to open one if you want it.

`--dtype=bfloat16` still shows 2 failures in `tests/enhance` (`test_jpeg::test_cardinality`, `test_zca::test_dim_args`); I confirmed they fail identically on unmodified `main`.

ducha-aiki (approved):
The revised dtype handling preserves integer-image behavior, float64 center precision, explicit-center promotion, and the established floating return dtype while fixing collapsed half-precision centers. The new tests discriminate the prior regressions and include MPS coverage. Focused `ImageHistogram2d` validation passed (93 tests); the remote CI matrix is green. No remaining blocking findings.

PR #4361, opened 2026-09-08
fix(geometry): render_gaussian2d pixel-coordinate linspace built at dtype

tritsystem:
## Summary

Same bug class as the already-merged bbox_to_mask fix (#4336): `xs`/`ys` were built via `torch.linspace(..., dtype=dtype)`, where `dtype` comes from the caller's `mean` tensor. `float16` can only exactly represent integers up to 2048 — confirmed directly: `torch.linspace(0, 2199, 2200, dtype=torch.float16)` has only 2124 distinct values, not 2200. For `width`/`height` beyond that, distinct pixel coordinates silently collapse onto the same value, distorting the rendered Gaussian's shape near the collapse region.

I checked a sibling case in the same submodule (`spatial_soft_argmax.py`'s `conv_quad_interp3d`/`iterative_quad_interp3d`) and deliberately did **not** file a fix for it: those functions directly *return* a coordinate tensor, so building it at float32 and casting back to `float16` at the end reproduces byte-identical output to building it directly at `float16` (confirmed empirically) — a genuine, unfixable limitation of that return contract, not a construction-order bug. `render_gaussian2d` is different: `xs`/`ys` feed into smooth Gaussian arithmetic whose *returned* density values don't need to represent large integers at all, and computing `xs` at float32 before the Gaussian math gives a materially different, more correct result (max abs diff ~0.33 on a `[0,1]`-range density) than computing directly at `float16` — confirmed this is a real, fixable case before touching the code.

Fixed by building `xs`/`ys` at `float32` for `float16`/`bfloat16` inputs (other dtypes keep their own precision), letting the subsequent arithmetic promote naturally, and casting the two normalized 1-D Gaussian vectors back to `dtype` before the outer product, so there is no full-size float32 intermediate.

## Test plan

- [x] `test_large_grid_peak_not_distorted`: 16 cases (`float16`/`bfloat16` x pixel/normalized coordinates x width/height axis x two mean positions, 2049.3 and 1500.2). It compares the rendered peak against `mean`'s own *actual stored* value (post dtype rounding), so only the coordinate grid is under test. Each case asserts the heatmap is not all zero, that the expected pixel is a maximum, and that every tied maximum lies within one pixel of it (a collapsed grid smears the maximum over a plateau, which the exact-pixel check alone accepts).
- [x] Red-then-green: 16 of 16 pass with the fix; against the unfixed code 10 of 16 fail. The 6 that still pass on the unfixed code are `float16` in normalized mode (barely affected there, about 1 px) and `bfloat16` in normalized mode at 2049.3 (1 px; it is 7 px at 1500.2).
- [x] Mutations killed by that test: full revert, `bfloat16` dropped from the dtype check, normalized `xs`, normalized `ys`, pixel `xs`, pixel `ys` (height), and dropping the final casts.
- [x] `tests/geometry/subpix` and `tests/test_half_precision_ci.py`, CPU `float32`: 212 passed, 9 skipped. `ruff check` and `ruff format --check` clean on the changed files.
- [ ] Not covered: GPU/MPS, and `torch.compile` of `render_gaussian2d` beyond the existing dynamo test. `test_normalized_coordinates[bfloat16]` and `test_pixel_coordinates[bfloat16]` already fail on unmodified `main` under `--dtype=bfloat16`; they are unrelated to this change.


ducha-aiki (changes requested):
The production change fixes the reported large-grid peak error while preserving the public output dtype. Two test-maintenance failures and the release note remain:

- The inserted regression shifts the existing eager `torch.randn` call in `tests/geometry/subpix/test_dsnt.py` from line 73 to line 96, but `testing/half_precision_eager_rng.py:191` still inventories line 73. Every hosted test matrix therefore fails `test_eager_rng_audit_matches_exact_inventory`; update the audited line to 96.
- `tests/geometry/subpix/test_dsnt.py:85-86` unconditionally creates the float64 comparison tensors on the injected device. The MPS job fails because MPS has no float64 support. Use a supported comparison dtype or skip only that loop case on MPS.
- Add the required `CHANGELOG.md` bug-fix entry citing #4361.

Validation at `732efd2c24d68f2fb03c3be24bc5495d01f132dc`: 15 CPU float32/float64 DSNT tests passed; the focused large-width probe peaks at 2049 on base versus the stored-mean expectation 2050, and at 2050 on this head. The hosted CPU and MPS logs reproduce the inventory and float64-device failures above. No prior comments or formal reviews exist.

ducha-aiki:
## What I verified

Measured on the merge tree (`2ab3943`, which already contains `origin/main` `8c71df9a0`) against a `main` worktree, CPU, torch 2.14.0, Apple M1. `kornia.__file__` asserted in every measurement block.

**The defect is real.** The `torch.linspace` precision loss at `width = 2200`:

| dtype | distinct coords / 2200 | max coordinate error |
|---|---|---|
| float16, pixel | 2124 | 2.0 px |
| bfloat16, pixel | 650 | 15.0 px |
| float16, normalized | 2200 | — |
| bfloat16, normalized | 495 | 0.00697 (~7.7 px) |

Propagated to `render_gaussian2d`, comparing the rendered peak against a float64 render driven by the *stored* half value of `mean` (so the input's own rounding is factored out), 10 mean positions per width, `std=2.0`, pixel coordinates:

| dtype | W=300 | W=512 | W=1024 | W=2048 | W=2200 | W=4096 |
|---|---|---|---|---|---|---|
| float16 **base** | 0/10 | 0/10 | 0/10 | 0/10 | 6/10 (1 px) | 4/10 (2 px) |
| bfloat16 **base** | 6/10 (1 px) | 4/10 (2 px) | 6/10 (3 px) | 8/10 (6 px) | 8/10 (4 px) | 8/10 (12 px) |
| both, **this PR** | 0/10 | 0/10 | 0/10 | 0/10 | 0/10 | 0/10 |

**No collateral movement at wide precision.** 44 cases (float32 and float64 x `normalized_coordinates` both ways x 11 sizes including 1x1, odd/even, 257x256): byte hash of every output **identical** on both trees (`61c7804d...`).

**Half-precision outputs do all move**, as expected: over 56 half cases, 54 changed bitwise; 49 got closer to the float64 reference, 5 unchanged in max relative error, 2 marginally farther (both ~5e-4, single-ULP noise). Monotone improvement.

**The `testing/half_precision_eager_rng.py` edit is correct and is not a count change.** It moves the one line-pinned entry for the `torch.randn` in `TestSpatialSoftmax2d`'s fixture parameter list from 73 to 98 — the new test inserts exactly 25 lines above it, and `grep -n torch.randn tests/geometry/subpix/test_dsnt.py` confirms 98 on the merge tree vs 73 on base. Path / call name / count / node prefixes / classification unchanged. `tests/test_half_precision_ci.py`: **62 passed**. Harness proven to discriminate: restoring `73` makes it go red with `audited entries with no matching call site (stale line?): [('tests/geometry/subpix/test_dsnt.py', 73, 'torch.randn')]`.

**Regression sweep**, `tests/geometry/subpix/` in full:
- `--dtype=float32`: base **129 passed**, PR **130 passed**.
- `--dtype=float16,bfloat16`: base **26 failed / 177 passed**, PR **26 failed / 178 passed**; the two failure lists `diff` byte-identically. The two `test_dsnt.py` entries are the pre-existing `cpu_bfloat16` manifest lines 314-315, which still legitimately fail — under `--xfail-known-failures --known-failure-profile=cpu-bfloat16` that is `6 passed, 2 xfailed` (not XPASS), so no manifest line goes stale.
- MPS (`--device=mps --dtype=float32`, `PYTORCH_ENABLE_MPS_FALLBACK` unset): base 7 passed, PR **8 passed**. The new test is collected once as `[cpu]`/`[mps]`, so it does not multiply the half/MPS legs.
- `KORNIA_TEST_OPTIMIZER=inductor`: 11 passed, 1 skipped.
- `torch.autograd.gradcheck` on float64 still `True`; grad dtypes preserved for fp16/fp32/fp64.
- `ruff@0.16.7` check and `format --diff` clean on all three source files; no line over 120 chars.

The `compute_dtype = torch.float32 if dtype in (torch.float16, torch.bfloat16) else dtype` construction is the established repo idiom (`kornia/contrib/distance_transform.py::_compute_dtype`, `kornia/feature/scale_space_detector.py:139`, `kornia/augmentation/random_generator/_2d/rectangle_earase.py:94`), so the shape of the fix is right. `changelog.d/4361.fixed.md` exists and `fixed` is correct — a precision improvement within a dtype, not an incompatible change. No public name added. The PR closes no issue, so no issue-prose blast radius; `conventions.rst`'s only `float16`/`2049 -> 2048` sentence is about augmentation mask dtype, unrelated.

## Findings

**[P2] Three non-equivalent mutants survive the new test.** The production diff changes four things; the single new pin exercises exactly one of them (`float16`, pixel coordinates, **width** axis). Each mutation, then `tests/geometry/subpix/test_dsnt.py --device=cpu --dtype=float32`:

| mutation | result | proof it is non-equivalent |
|---|---|---|
| `compute_dtype = dtype` (full revert) | **killed** (1 failed) | — |
| `torch.linspace(0, width, width, ...)` (endpoint off-by-one) | **killed** (2 failed) | — |
| drop `.to(dtype)` | **killed** (`assert torch.float32 == torch.float16`) | — |
| `if dtype is torch.float16` — **bfloat16 dropped from the tuple** | **8 passed — SURVIVES** | bf16 pixel W=4096: 8/10 peaks wrong, worst **12 px**; W=2048: 8/10, 6 px; W=300: 6/10, 1 px. bf16 normalized W=2200: worst **10 px**, worst relative error **0.32**. |
| normalized branch back to `dtype=dtype` | **8 passed — SURVIVES** | bf16 `normalized_coordinates=True`, W=2200, 120 (mu, std) combinations: worst peak displacement **10 px**, worst relative error **0.3228**, vs **5 px / 0.0042** unmutated (`torch.linspace(-1, 1, 2200, dtype=torch.bfloat16)` has 495 distinct values). |
| `ys` (height) pixel branch back to `dtype=dtype` | **8 passed — SURVIVES** | H=2200, W=10, mean `[5.0, 2049.3]`: fp16 peak_y **2049** vs float64 **2050**; bf16 peak_y **2051** vs float64 **2048**. |

The last one matters for the changelog too: the fragment claims the fix covers `height`, but nothing pins the height axis, so a future edit touching only `ys` ships silently. Please extend the pin to cover (a) `bfloat16`, (b) `normalized_coordinates=True`, (c) the height axis — e.g. parametrize over `(dtype, normalized, axis)` rather than looping two dtypes inside one test body.

**[P2] The cast placement costs ~9x runtime and 1.74x peak memory on the workload this PR is about, and it is avoidable.** `(gauss_y.unsqueeze(-1) * gauss_x.unsqueeze(-2)).to(dtype)` materializes the full `(*, H, W)` outer product at float32 and then allocates the half copy. `B=64, H=W=1024`, float16, `torch.set_num_threads(4)`, 21 iterations after 5 warmups:

| | median | min | peak RSS delta |
|---|---|---|---|
| base | 58.66 ms | 16.84 ms | 127.0 MB |
| this PR | 453.93 ms | 151.98 ms | 220.4 MB |

(`B=8, 512x512`: 2.67 -> 3.76 ms. Small sizes unaffected.)

Only the 1-D coordinate vectors need the wider dtype; the outer product does not. Casting the two 1-D vectors instead:

```python
gauss_x = gauss_x.to(dtype)
gauss_y = gauss_y.to(dtype)
return gauss_y.unsqueeze(-1) * gauss_x.unsqueeze(-2)
```

Applied exactly that on the merge tree: **64.51 ms median / 16.44 ms min / 101.2 MB peak RSS** — back to base, and *below* base on memory — all **8 tests still pass**, and the full peak sweep above is still **0/10 wrong at every width for both fp16 and bf16**. The entire correctness benefit is retained. Per `AGENTS.md` ("For performance changes, run the same benchmark on the relevant base revision and changed branch"), please take the cheaper placement or justify the current one with a measurement.

**[P3] The changelog fragment understates the fix by a dtype and by ~8x in threshold.** It says "for `width`/`height` beyond 2048 on a float16 input". bfloat16 has an 8-bit significand, so its pixel linspace collapses from `width ~ 257`, not 2049 — measured on base, `bfloat16` at **W=300** already puts 6/10 peaks on the wrong pixel, and `torch.linspace(0, 2199, 2200, dtype=torch.bfloat16)` has **650** distinct values against float16's 2124. A bfloat16 user at 512x512 reading this fragment would conclude they were never affected; on base they had 4/10 peaks wrong with a 2 px worst error. Please name `bfloat16` and its own threshold, and say that every float16/bfloat16 output changes bitwise (54 of 56 sampled cases) — that is what a downstream user with a stored regression baseline will hit.

**[P3] Test style.** The 15-line narrative docstring belongs in the PR body or a two-line comment; nearby tests use one-line docstrings. The in-body `for dtype in (...)` loop should be `@pytest.mark.parametrize` so a failure names the dtype. `torch.float16` is hardcoded rather than taken from the injected `dtype` fixture — defensible here since the test is dtype-targeted and I confirmed it collects once and passes on CPU and MPS, but it will not skip cleanly on an XLA/TPU backend without fp16. The name says `float16` while the test also runs a float32 control arm.

**[P3] Idiom duplication.** That `compute_dtype` expression is now the fourth open-coded copy in `kornia/` (`kornia/contrib/distance_transform.py` already wraps it as a private `_compute_dtype`). Not this PR's job, but a shared `kornia/utils` helper is overdue.

## What is left to do

1. Pin bfloat16, `normalized_coordinates=True`, and the height axis — the three surviving mutants.
2. Move the `.to(dtype)` onto `gauss_x`/`gauss_y` before the outer product, or post a benchmark defending the current placement.
3. Widen the changelog fragment to name bfloat16 and its real threshold, and to say half-precision outputs change bitwise.

The production change is correct as far as it goes, and the audit-file line move is exactly right — happy to approve once the pins cover the branches the fix touches.

One housekeeping note: the PR body carries the `Generated with Claude Code` footer, which this repo does not use. Please drop it.

victorwon2001:
**kornia/geometry/subpix/dsnt.py:145** — The new production comment says the coordinate vectors are created at a fixed float32 regardless of dtype, but compute_dtype is float32 only for float16/bfloat16 and otherwise remains the original dtype.

The comment misdocuments float32/float64 behavior and can mislead future maintenance or precision reasoning around this numerical path.

The comment states fixed float32 regardless of dtype immediately before a conditional expression that preserves non-half dtypes.

`file:tests/geometry/subpix/test_dsnt.py:head`

> width = 2200
>         mu = 2049.3
>         for dtype in (torch.float16, torch.float32):
>             mean = torch.tensor([[mu, 5.0]], dtype=dtype, device=device)
>             std = torch.tensor([[2.0, 2.0]], dtype=dtype, device=device)
>             heatmap = kornia.geometry.subpix.render_gaussian2d(mean, std, (10, width), False)
>             assert heatmap.dtype == dtype
>             peak_x = int(heatmap[0, 5].argmax())
>             expected = round(mean[0, 0].item())
>             assert peak_x == expected, f"dtype={dtype}: peak at pixel {peak_x}, expected {expected}"

`comment:5711876388`

> **[P2] Three non-equivalent mutants survive the new test.** The production diff changes four things; the single new pin exercises exactly one of them (`float16`, pixel coordinates, **width** axis). Each mutation, then `tests/geometry/subpix/test_dsnt.py --device=cpu --dtype=float32`:
> 
> | mutation | result | proof it is non-equivalent |
> |---|---|---|
> | `compute_dtype = dtype` (full revert) | **killed** (1 failed) | — |
> | `torch.linspace(0, width, width, ...)` (endpoint off-by-one) | **killed** (2 failed) | — |
> | drop `.to(dtype)` | **killed** (`assert torch.float32 == torch.float16`) | — |
> | `if dtype is torch.float16` — **bfloat16 dropped from the tuple** | **8 passed — SURVIVES** | bf16 pixel W=4096: 8/10 peaks wrong, worst **12 px**; W=2048: 8/10, 6 px; W=300: 6/10, 1 px. bf16 normalized W=2200: worst **10 px**, worst relative error **0.32**. |
> | normalized branch back to `dtype=dtype` | **8 passed — SURVIVES** | bf16 `normalized_coordinates=True`, W=2200, 120 (mu, std) combinations: worst peak displacement **10 px**, worst relative error **0.3228**, vs **5 px / 0.0042** unmutated (`torch.linspace(-1, 1, 2200, dtype=torch.bfloat16)` has 495 distinct values). |
> | `ys` (height) pixel branch back to `dtype=dtype` | **8 passed — SURVIVES** | H=2200, W=10, mean `[5.0, 2049.3]`: fp16 peak_y **2049** vs float64 **2050**; bf16 peak_y **2051** vs float64 **2048**. |
> 
> The last one matters for the changelog too: the fragment claims the fix covers `height`, but nothing pins the height axis, so a future edit touching only `ys` ships silently. Please extend the pin to cover (a) `bfloat16`, (b) `normalized_coordinates=True`, (c) the height axis — e.g. parametrize over `(dtype, normalized, axis)` rather than looping two dtypes inside one test body.

`file:kornia/geometry/subpix/dsnt.py:diff`

> @@ -175,4 +184,4 @@ def render_gaussian2d(
>      gauss_x = gauss_x / (gauss_x.sum(dim=-1, keepdim=True) + 1e-8)
>      gauss_y = gauss_y / (gauss_y.sum(dim=-1, keepdim=True) + 1e-8)
>  
> -    return gauss_y.unsqueeze(-1) * gauss_x.unsqueeze(-2)
> +    return (gauss_y.unsqueeze(-1) * gauss_x.unsqueeze(-2)).to(dtype)

`comment:5711876388`

> **[P2] The cast placement costs ~9x runtime and 1.74x peak memory on the workload this PR is about, and it is avoidable.** `(gauss_y.unsqueeze(-1) * gauss_x.unsqueeze(-2)).to(dtype)` materializes the full `(*, H, W)` outer product at float32 and then allocates the half copy. `B=64, H=W=1024`, float16, `torch.set_num_threads(4)`, 21 iterations after 5 warmups:
> 
> | | median | min | peak RSS delta |
> |---|---|---|---|
> | base | 58.66 ms | 16.84 ms | 127.0 MB |
> | this PR | 453.93 ms | 151.98 ms | 220.4 MB |
> 
> (`B=8, 512x512`: 2.67 -> 3.76 ms. Small sizes unaffected.)
> 
> Only the 1-D coordinate vectors need the wider dtype; the outer product does not. Casting the two 1-D vectors instead:
> 
> ```python
> gauss_x = gauss_x.to(dtype)
> gauss_y = gauss_y.to(dtype)
> return gauss_y.unsqueeze(-1) * gauss_x.unsqueeze(-2)
> ```
> 
> Applied exactly that on the merge tree: **64.51 ms median / 16.44 ms min / 101.2 MB peak RSS** — back to base, and *below* base on memory — all **8 tests still pass**, and the full peak sweep above is still **0/10 wrong at every width for both fp16 and bf16**. The entire correctness benefit is retained. Per `AGENTS.md` ("For performance changes, run the same benchmark on the relevant base revision and changed branch"), please take the cheaper placement or justify the current one with a measurement.

`file:changelog.d/4361.fixed.md:head`

> `kornia.geometry.subpix.render_gaussian2d` no longer distorts the rendered Gaussian's peak
> location for `width`/`height` beyond 2048 on a float16 input. The internal pixel-coordinate
> linspace used to be built at `dtype`; float16 can only exactly represent integers up to 2048
> (`torch.linspace(0, 2199, 2200, dtype=torch.float16)` has only 2124 distinct values, not 2200),
> so distinct coordinates silently collapsed together near the top of a large grid. The coordinate
> grid and downstream Gaussian computation now run at float32 internally whenever the caller's
> dtype is float16/bfloat16, with the returned heatmap explicitly cast back to that original dtype
> at the end.

`comment:5711876388`

> **[P3] The changelog fragment understates the fix by a dtype and by ~8x in threshold.** It says "for `width`/`height` beyond 2048 on a float16 input". bfloat16 has an 8-bit significand, so its pixel linspace collapses from `width ~ 257`, not 2049 — measured on base, `bfloat16` at **W=300** already puts 6/10 peaks on the wrong pixel, and `torch.linspace(0, 2199, 2200, dtype=torch.bfloat16)` has **650** distinct values against float16's 2124. A bfloat16 user at 512x512 reading this fragment would conclude they were never affected; on base they had 4/10 peaks wrong with a 2 px worst error. Please name `bfloat16` and its own threshold, and say that every float16/bfloat16 output changes bitwise (54 of 56 sampled cases) — that is what a downstream user with a stored regression baseline will hit.

`file:kornia/geometry/subpix/dsnt.py:head`

> # Create coordinates vectors at a fixed float32, regardless of `dtype` -- float16 can only
>     # exactly represent integers up to 2048, so for width/height beyond that, building the
>     # pixel-coordinate linspace directly at a float16 `dtype` collapses distinct coordinates
>     # together (confirmed directly: torch.linspace(0, 2199, 2200, dtype=torch.float16) has only
>     # 2124 distinct values, not 2200), silently distorting the rendered Gaussian's shape near
>     # the collapse point. `dist_x_sq`/`dist_y_sq`/`gauss_x`/`gauss_y` below promote to float32
>     # automatically once combined with these (ordinary PyTorch type promotion); the final
>     # returned heatmap is cast back to `dtype` at the very end, where it belongs -- a smooth
>     # density in [0, 1] loses nothing meaningful by rounding to float16 only once, at output.
>     compute_dtype = torch.float32 if dtype in (torch.float16, torch.bfloat16) else dtype
>     if normalized_coordinates:

Extend the regression pin to bfloat16, normalized coordinates, and the height axis; move the cast onto the normalized 1-D Gaussian vectors before the outer product; correct the bfloat16 changelog scope; and make the compute-dtype comment match the actual conditional behavior.

SHA: `2ab394300681c09f1141120666bddb230c476020`

ducha-aiki (changes requested):
## [P2] Make the normalized half-precision regression non-vacuous

In tests/geometry/subpix/test_dsnt.py:82-96 the normalized cases assign the long-axis sigma to both axes. On the 10-pixel short axis there is then no grid point near mean zero, so the half-precision heatmap rounds entirely to zero and line[expected] == line.max() passes as 0 == 0. On this head the bfloat16 normalized/x case reports expected index 2049, argmax 0, max 0, with all 2200 entries tied. Use an axis-appropriate normalized standard deviation and assert a nonzero unique peak or the actual argmax so the normalized-coordinate fix is exercised.

The implementation and cast placement otherwise address the prior findings. Please also qualify the fragment's claim that all float16/bfloat16 outputs change bitwise: small 1x1, 2x2, and 5x7 cases remain byte-identical, so this is true only for affected wide grids.

tritsystem:
Thanks, the [P2] was right. I reproduced it before changing anything: on the previous head the normalized `bfloat16` case gives expected index 2049, argmax 0, max 0, and every entry of the 10-pixel-axis heatmap is zero, so `line[expected] == line.max()` passed as `0 == 0`.

Pushed 6064bd9 (also merges current `main`, which was 57 commits ahead):

- **Sigma per axis.** The normalized cases now use a 6-pixel sigma expressed in each axis's own normalized units, and assert `line.max() > 0`.
- **The zero-guard alone was not enough.** I checked the new test against the unfixed code, and the 4 normalized cases still passed there. A collapsed grid does not move the peak to one wrong pixel, it smears the maximum over a plateau that still contains the expected pixel. So the test now also requires every tied maximum to be within 1 px of the expected one. On the fix the plateau is within 1 px at every mean I tried; on the unfixed code `bfloat16` normalized is 1 px off at mean 2049.3 but 7 px off at 1500.2, so the test now runs at both means (2049.3 and 1500.2).
- **Result:** 16 of 16 pass with the fix, 10 of 16 fail on unfixed `main`. The 6 that still pass on unfixed code are `float16` in normalized mode (about 1 px, barely affected there) and `bfloat16` normalized at 2049.3. I do not think a test can honestly fail those.
- **Mutants** (`compute_dtype = dtype`, `bfloat16` dropped from the check, normalized `xs`, normalized `ys`, pixel `xs`, pixel `ys`/height, final casts removed): all 7 killed. That includes the three that the comment on 2ab3943 reported as surviving (`bfloat16` dropped, normalized branch reverted, height axis reverted). That comment was also written against an older head; 27efb56 already casts the two 1-D vectors instead of the outer product, corrects the compute-dtype comment, and names `bfloat16` in the fragment.

Where I did not follow the review: you noted that 1x1, 2x2 and 5x7 outputs stay byte-identical. I compared the actual outputs of this branch against unmodified `main` on random means and sigmas (both coordinate modes, 4 trials each, `float16` and `bfloat16`): 1x1 is identical in 8/8, 2x2 in 1-2 of 8, and 5x7, 16x16, 64x64 and larger differ in 8/8. The Gaussian arithmetic now runs in `float32`, not only the coordinates, so nearly every output changes bitwise. I kept the fragment's warning about regenerating baselines and made it say exactly that. If you saw identical 5x7 output for particular inputs, I would like to know which, because I could not reproduce it.

Also: `ruff check` and `ruff format --check` are clean on the changed files, and I dropped the "Generated with Claude Code" footer from the description as the project does not use it. With `--dtype=bfloat16`, `test_normalized_coordinates` and `test_pixel_coordinates` fail on unmodified `main` as well as here; I left them alone. I did not test GPU or MPS.

ducha-aiki (approved):
The normalized half-precision regression now uses axis-specific standard deviations, rejects all-zero heatmaps, and checks the maximum plateau, so it exercises both coordinate modes and axes. The fragment accurately qualifies the bitwise impact. Focused `TestRenderGaussian2d` validation passed (20 tests). No remaining blocking findings.

PR #4362, opened 2026-09-08
fix(core): let download_file_from_url quarantine a corrupt cache entry

### LINKED ISSUE #4309: download_file_from_url: a truncated-but-2xx transfer poisons the cache entry permanently

aminmiral:
Fixes #4309.

`torch.hub.download_url_to_file` does not check the body against `Content-Length`, so a connection dropping after a 2xx leaves a truncated file in the cache. `download_file_from_url` returned it as a hit on every later call, and `from_pretrained_hf()` failed on it until the file was deleted by hand.

Took the second of the two options in the issue. `download_file_from_url` and `download_hf_file` accept an optional `validate` callable, applied to the cached file before it is returned; a rejection goes through the same quarantine as a load failure in `load_state_dict_from_url`. `kornia.core.validate_safetensors` is that check for a `.safetensors` file — the header parse of `load_safetensors`, now shared as `_read_header`, without the mapping that follows — and both builders pass it.

Two places where mirroring `load_state_dict_from_url` was not quite right, both noted in the commit message: the single-URL case needs its re-attempt pass or the discard empties the path for a source already served the bad file, and a fresh transfer that `validate` refuses is dropped even with no source left to use the path, since that rejection is a verdict on the file rather than one of the ambiguous load failures that function has to weigh.

Tests: seven in `TestValidateRejectsACorruptCacheEntry` and eight for `validate_safetensors`, covering a truncated entry being refetched, the entry returned untouched without `validate`, a valid entry not refetched, restoration when nothing can replace it, a rejected fresh transfer not being left behind, what the failure message reports, and rejection of a truncated body, truncated header, empty file and an HTML page served with a 200.

`tests/core`, `tests/models`, `tests/test_api_surface.py` and `tests/scripts/test_check_import_surface.py` pass (654 passed, 19 skipped); ruff clean. `validate_safetensors` is added to `docs/source/core.rst` and to `tests/api_surface.json`.


ducha-aiki (commented):
Reviewed at `af354220b7dfce47d7f5bc0cb3567a9fa9ffc3e7` — this substantially duplicates #4332, so the two implementations must not merge independently.

The documentation check is independently red. Sphinx reports `kornia/core/safetensors.py`, `validate_safetensors` docstring line 20: “Explicit markup ends without a blank line; unexpected unindent.” Add a blank line after the doctest at source line 260 so the closing delimiter does not terminate the explicit block immediately.

Focused validation otherwise passed: 147 downloader/safetensors/builder tests and 146 API-surface tests. Compared with #4332, this branch adds useful dedicated safetensors-validator coverage and drops a rejected fresh transfer even for the final source; those improvements can be folded into the earlier PR if desired. The shared one-refetch-per-cache-path/process budget is intentional storm control and is not a defect. I recommend closing #4362 as the later duplicate after any desired consolidation.

aminmiral:
Closing — #4332 landed the same approach first. Same `validate` callable, same re-attempt pass, same extracted header parse.

Two small things in this branch that #4332 does not cover; I'll open a follow-up against main rather than leave this one open:

* `_drop_failed_download` is gated on `more_sources`, so a *fresh* transfer that `validate` refuses stays in the cache when there is one URL — which is every `download_hf_file` caller. `from_pretrained_hf()` on a cold cache with a connection cut after a 2xx fails *and* leaves the poison; recovery waits for the next call. A `validate` rejection is a verdict on the file, unlike the ambiguous load failures that gate exists for, so those bytes can go unconditionally.
* The `refetch_note` swap from `load_state_dict_from_url` is not carried over, so when the re-attempt fails offline the `RuntimeError` reports the `URLError` rather than the rejection that names the truncation.

PR #4363, opened 2026-09-08
fix(augmentation): RandomMixUpV2 silently upcasts float16/bfloat16 input

tritsystem:
## Summary

`apply_transform`'s blend weight (`lam = params["mixup_lambdas"]...to(input.device)`) was moved to `input.device` but never cast to `input.dtype`. `MixupGenerator`'s own device/dtype tracking is independent of the real runtime input — its `forward()` derives dtype from the constructor's raw `lambda_val` argument, which defaults to `None` (confirmed directly: a fresh generator's tracked dtype stays `torch.float32` regardless of what dtype it's later called with). `input * (1 - lam)` then silently upcasts a `float16` input to `float32` through ordinary PyTorch type promotion the instant `lam` isn't already the same dtype — reproduced directly: `RandomMixUpV2` on a `float16` batch returns a `float32` batch.

Fixed by casting `lam` to `input.dtype` (not just `input.device`) in `apply_transform`.

**Note on scope**: while diagnosing this I found `MixupGenerator.forward()`'s own dtype-tracking mechanism (`_extract_device_dtype([self.lambda_val])`) is itself questionable — and the same `_extract_device_dtype([self.<constructor arg>])` pattern appears in ~20 other generator files (`cutmix.py`, `posterize.py`, `translate.py`, `affine.py`, etc.). I have **not** individually verified whether any of those have an analogous bug — some (like `posterize.py`) coerce their output to a fixed dtype regardless, so may not be affected the same way. Flagging this here in case it's worth a dedicated audit, but keeping this PR scoped to the one directly confirmed, reproduced issue.

## Test plan

- [x] Added a test parametrized over `float16`/`float32` confirming output dtype matches input dtype
- [x] Verified red-then-green
- [x] `bfloat16` deliberately excluded from the test: `kornia.constants.DType` has no `BFLOAT16` member at all, so the labels path independently crashes with `KeyError: 'BFLOAT16'` regardless of this fix — a real, separate, pre-existing bug, out of scope here
- [x] Full `tests/augmentation/test_augmentation_mix.py` suite (51 tests) passes, no regressions
- [x] `ruff check` clean


ducha-aiki (changes requested):
Reviewed at `3d690af7a206e0299edacb233fc70fb352a5d001` — the dtype fix is correct, but the strict half-precision baseline and release note must be updated before merge.

Casting `mixup_lambdas` to the runtime input's device and dtype prevents `RandomMixUpV2` from widening float16/bfloat16 image tensors. Focused validation passes: 9 `RandomMixUpV2` tests from `tests/augmentation/test_augmentation_mix.py`.

The float16 CI failure is a real baseline-maintenance failure caused by this fix. These two entries now strict-XPASS and must be removed from `testing/half_precision_xfails/cpu_float16.txt`:

```
tests/augmentation/test_augmentation_mix.py::TestRandomMixUpV2::test_random_mixup_lam0[cpu-float16]
tests/augmentation/test_augmentation_mix.py::TestRandomMixUpV2::test_random_mixup_same_on_batch[cpu-float16]
```

Keep the `test_random_mixup_p1` entry, which did not XPASS. Please also add a `CHANGELOG.md` entry under `Unreleased` → `Bug fixes` citing `#4363`.

tritsystem:
Closing this — #4471 (merged a few hours ago) makes the exact same one-line fix to `mixup.py` (`.to(input.device, dtype=input.dtype)`), so this is now redundant. Confirmed by diffing directly rather than assuming: the change is byte-for-byte identical to what this PR proposed.

Thanks for the earlier review feedback regardless — the xfail-list housekeeping and CHANGELOG-fragment pointers were useful for the sibling PRs (#4360, #4364) even though this specific fix landed via a different route.

PR #4364, opened 2026-09-08
fix(augmentation): RandomPlanckianJitter silently upcasts float16 input

tritsystem:
## Summary

Found via a systematic sweep of every `Random*` augmentation class for dtype-preservation, run after fixing the analogous `RandomMixUpV2` bug (#4363) — same symptom class, different mechanism.

`apply_transform` indexed the registered `pl` buffer (the Planckian-locus color coefficient table) with `.to(device=input.device)` but no `dtype=`. A registered buffer follows the whole `nn.Module`'s `.to()` calls, but stays at its own construction-time dtype (`float32`) if the caller never explicitly `.to(dtype=...)`'s the augmentation module itself — which is the normal way these are used (call `aug(x)` directly on a `float16` batch, never touching the module's own dtype). `input * r_w` then silently upcasts a `float16` input to `float32` through ordinary PyTorch type promotion the instant `r_w`/`b_w` (sliced from the buffer) aren't already the same dtype — reproduced directly.

Fixed by casting the buffer to `input.dtype` (not just `input.device`) before indexing it.

## Context: found while auditing a systemic lead

While fixing #4363 I flagged that the same `_extract_device_dtype([self.<constructor-arg>])` pattern appears in ~20 other generator files, unverified. Rather than auditing those one at a time, I ran a broader, more direct sweep: constructed every `kornia.augmentation.Random*` class and called it on a `float16` batch, checking output dtype. Out of 45 successfully-constructed-and-called augmentations, exactly 2 leaked: this one, and `RandomDissolving`. `RandomDissolving`'s leak comes from an internal Stable Diffusion pipeline — a real, different, heavier case (likely has genuine numerical-stability considerations for running a diffusion UNet in float16, not a simple tensor-dtype fix) — not touched by this PR, flagging it separately rather than rushing a fix there.

## Test plan

- [x] Added a test parametrized over `float16`/`float32` confirming output dtype matches input dtype
- [x] Verified red-then-green
- [x] Full `TestPlanckianJitter` suite (6 tests) passes, `ruff check` clean


ducha-aiki (changes requested):
Reviewed at `62478a6f552e525c7fee2de777739e7ff01073b5` — the dtype fix is correct, but both strict half-precision baselines and the release note must be updated before merge.

Casting the Planckian coefficient table to the runtime input dtype prevents `RandomPlanckianJitter` from widening float16/bfloat16 outputs. Focused validation passes: 6 `PlanckianJitter` tests from `tests/augmentation/test_augmentation.py`.

Both half-precision CI jobs fail because the fix makes all four existing Planckian tests strict-XPASS. Remove the four `TestPlanckianJitter` entries (`test_planckian_jitter_batch`, `test_planckian_jitter_blackbody`, `test_planckian_jitter_cied`, and `test_planckian_jitter_same_on_batch`) from each of `testing/half_precision_xfails/cpu_float16.txt` and `testing/half_precision_xfails/cpu_bfloat16.txt`. Please also add a `CHANGELOG.md` entry under `Unreleased` → `Bug fixes` citing `#4364`.

ducha-aiki:
Reviewed at `d066441e` on the merge tree with `origin/main` (`8c71df9a0`); the branch is already rebased, so `git merge origin/main` reports "Already up to date". torch 2.14.0, CPU.

Thanks for this — the diagnosis is correct and the fix is the right one. The problem is that **#4578 fixes the same defect with strictly stronger tests**, and the two cannot both land. Detail below, then the tie-break.

## What I verified

**The defect is real on `main`.** `K.RandomPlanckianJitter(p=1.0)` returns `float32` for both `float16` and `bfloat16` input; `float32`/`float64` keep their dtype. On the merge tree all four inputs come back in their own dtype.

**The fix fixes it, and only it.** A 40-cell A/B sweep (2 modes x 4 input dtypes x {no module cast, `.to()` each of the 4 float dtypes}), hashing the `float64`-promoted output bytes on base and on the merge tree: **18 of 40 cells change, and all 18 are dtype changes** — no value drift in any cell whose dtype is unchanged. `float32`/`float64` output is bit-identical.

**`torch.compile(fullgraph=True)` still captures the op** and preserves dtype for `float32`/`float16`/`bfloat16` — the graph-capture concern the in-code comment documents is not regressed by adding `dtype=`.

**The 8 deleted half-precision manifest lines are all correct.** Each of the 4 `TestPlanckianJitter` nodes fails on base with `AssertionError: The values for attribute 'dtype' do not match: torch.float3...` at both dtypes, and passes on the merge tree:

| tree | `--dtype=float16` | `--dtype=bfloat16` |
|---|---|---|
| base `8c71df9a0` | 4 failed | 4 failed |
| merge tree | 6 passed | 6 passed |

**Regression sweep, whole `tests/augmentation/`:**

| leg | base | merge tree |
|---|---|---|
| `--dtype=float32` | 3370 passed, 3405 skipped, 30 xfailed, 5 xpassed | **3372** passed, 3405 skipped, 30 xfailed, 5 xpassed |
| `--dtype=float16` (cpu-float16 profile) | 5 failed, 3132 passed, 229 xfailed | 5 failed, **3138** passed, **225** xfailed |
| `--dtype=bfloat16` (cpu-bfloat16 profile) | — | 3316 passed, 82 xfailed, 0 failed |

The 5 `float16` failures are byte-for-byte the same nodes on both trees — pre-existing local macOS noise, not this PR. No newly-red node anywhere. `ruff@0.16.7` clean; module doctest passes.

## Findings

**[P2] A non-equivalent mutant survives every leg of this PR's tests.** Mutant: cast the buffer only for half inputs —

```python
coeffs = self.pl.to(
    device=input.device,
    dtype=input.dtype if input.dtype in (torch.float16, torch.bfloat16) else self.pl.dtype,
)[params["idx"].long()]
```

Genuinely non-equivalent: under it, a `float32` input through a module cast with `.to(torch.float64)` returns `float64`, where this PR returns `float32` (measured on the mutated merge tree). Running each PR's whole Planckian test set:

| | float32 | float16 | bfloat16 | float64 |
|---|---|---|---|---|
| this PR (#4364) | SURVIVED | SURVIVED | SURVIVED | SURVIVED |
| #4578 | KILLED | KILLED | KILLED | KILLED |

The gap is a dropped case. `main` pinned the module-cast direction with `assert K.RandomPlanckianJitter(p=1.0).to(device=device, dtype=dtype)(image).dtype == dtype`; this PR deletes that line and replaces it with nothing that exercises a module cast *wider* than the input. #4578 replaces it with a dedicated `test_convention_random_planckian_jitter_wider_module_dtype_4574`, which is what kills the mutant on all four legs.

Seven further one-token mutants (revert to device-only; hardcode `dtype=torch.float32`; hardcode `float64`; swap the `coeffs[:, 0]`/`coeffs[:, 1]` gain columns; drop `clamp(max=1.0)`; force row 0; scale `r_w` by 2.0) are all killed at the suite level on both PRs, so this is the one hole.

**[P2] The change is incomplete: the module still documents the wart it just fixed.** `grep -rnE "#4574|issues/4574" kornia/ tests/ docs/` on the merge tree still hits `kornia/augmentation/_2d/intensity/planckian_jitter.py:137`. Lines 133-137 are a rendered `.. warning::` that now states the opposite of the shipped behaviour:

> The table is not cast to the input's dtype, so the output takes the promoted dtype of the input and the table. With the table left at `float32`, a `float16` or `bfloat16` input comes back as `float32`; casting the module to the input's own dtype keeps it. Tracked in `#4574`.

Every clause is false after this patch, and the second advertises a workaround the patch removes. Per AGENTS.md a surviving `kornia/` hit for the closed issue number means the change is incomplete. The block arrived via the rebase (added by #4566, batch 6c, `30dfbf711`); #4578 deletes it.

**[P2 in kind] The `.to(float64)` incompatibility carries only a `fixed` fragment.** Measured:

```
BASE  .to(float64) on a float32 input -> torch.float64
PR    .to(float64) on a float32 input -> torch.float32
```

A user-visible incompatible change to a non-half dtype, and the docstring quoted above is exactly where the old contract was written down. `changelog.d/4364.fixed.md` is the only fragment and does not mention the module-cast direction. #4578 hit the same thing in review and now ships `4578.breaking.md` alongside `4578.fixed.md`.

**[P3] `TestPlanckianJitter::test_output_preserves_input_dtype` ignores the injected `dtype` fixture.** It takes only `device` and hardcodes `@pytest.mark.parametrize("input_dtype", [torch.float16, torch.float32])`, against the AGENTS.md rule; it never covers `bfloat16` or `float64`. In its favour: because the dtypes are hardcoded it *does* pin float16 preservation on the default CPU-`float32` leg — which is a genuine advantage over #4578, and I am porting it there rather than losing it.

**[P3] `changelog.d/4364.fixed.md` never names issue #4574,** so merging it would not close the issue.

## Duplicate tie-break versus #4578

The **executable** post-image is identical. `self.pl.to(device=input.device, dtype=input.dtype)` (here) and `self.pl.to(input)` (#4578) are the same operation, and the 40-cell A/B sweep produced **0 differing cells out of 40** between the two merge trees (same aggregate hash `c09d356f4e9a0683`; base `796d5a0bcf07f5ae`). The only remaining production difference is that #4578 also deletes the stale `.. warning::` block.

They cannot both land: `git merge-tree --write-tree refs/remotes/pr/4364 refs/remotes/pr/4578` exits 1 with content conflicts in `planckian_jitter.py`, `test_augmentation.py` and `test_conventions_intensity_values.py`.

So the tie-break is test quality, and it goes to **#4578**:

| | #4364 | #4578 |
|---|---|---|
| half-only-cast mutant (non-equivalent) | survives on all 4 legs | killed on all 4 legs |
| module cast *wider* than input pinned | no | yes |
| `assert_close` in `..._scales_red_and_blue_then_clamps_above` | keeps `out.to(dtype)`, so the pin cannot see a dtype regression | drops the cast, so `assert_close` checks dtype too |
| new pin uses the injected `dtype` fixture | no (hardcoded `[float16, float32]`) | yes (all four dtypes per leg) |
| convention census | 69 -> 69 (rename, one case dropped) | 69 -> 70 (one dropped, two added) |
| stale `#4574` docstring warning removed | no | yes |
| `breaking` fragment for the `.to(float64)` reversal | no | yes |

**Closing this in favour of #4578**, and porting the one thing this PR does better — a dtype-preservation assertion that runs on the default `float32` leg — into #4578 so nothing is lost. The manifest deletions are byte-identical in both PRs and verified correct.

This is not a rejection of the work: your diagnosis was right, you found it a week before the other PR was opened, and the measurement that the two are executably identical is the reason the decision comes down to the test suite rather than the fix. Please do keep sending these — #4359, #4360 and #4361 from the same batch are all live and being reviewed on their own merits.

Issue #4249, opened 2026-09-05
Bug: `VideoBoxes` breaks inherited `Boxes` methods — `get_boxes_shape`/`to_mask` raise `TypeError`, indexing drops `temporal_channel_size`


ducha-aiki:
`VideoBoxes` (exported and documented by #4138) overrides `to_tensor(self, mode=None)` without the `as_padded_sequence` keyword that the inherited `Boxes.get_boxes_shape` and `Boxes.to_mask` pass, so both raise `TypeError` on a `VideoBoxes`. Separately, `Boxes.__getitem__` builds the result with `type(self)(...)` and never sets `temporal_channel_size`, so `video_boxes[0].to_tensor()` raises `AttributeError`. Methods that copy through `clone()` (`transform_boxes`, `translate`, `clamp`, `filter_boxes_by_area`, `pad`, `unpad`, `merge`, `index_put`, `to`, `type`) keep the temporal size and work.

Reproduced 2026-09-05 at `1eea5835`, PyTorch 2.14.0, Python 3.11, CPU and MPS.

## Reproduction

```python
import torch
from kornia.geometry.boxes import VideoBoxes

frame = torch.tensor([[[1., 1.], [3., 1.], [3., 3.], [1., 3.]]])  # (1, 4, 2)
video = VideoBoxes.from_tensor(frame.view(1, 1, 1, 4, 2).expand(2, 3, 1, 4, 2).contiguous())  # (B=2, T=3)
for label, call in (
    ("get_boxes_shape", lambda: video.get_boxes_shape()),
    ("to_mask", lambda: video.to_mask(4, 5)),
    ("[0].to_tensor", lambda: video[0].to_tensor()),
    ("clamp keeps T", lambda: video.clamp(torch.zeros(6, 2), torch.full((6, 2), 2.0)).temporal_channel_size),
):
    try:
        print(label, "->", call())
    except Exception as error:
        print(label, "->", type(error).__name__, str(error)[:80])
```

## Observed

```text
get_boxes_shape -> TypeError VideoBoxes.to_tensor() got an unexpected keyword argument 'as_padded_sequence'
to_mask -> TypeError VideoBoxes.to_tensor() got an unexpected keyword argument 'as_padded_sequence'
[0].to_tensor -> AttributeError 'VideoBoxes' object has no attribute 'temporal_channel_size'
clamp keeps T -> 3
```

## Expected

[...]

PR #4365, opened 2026-09-08
fix(boxes): let VideoBoxes.to_tensor accept as_padded_sequence (closes #4176)

### LINKED ISSUE #4176: Bug: `VideoBoxes.get_boxes_shape()` raises `TypeError` — inherited body passes `as_padded_sequence` to the `to_tensor` override

dchaudhari7177:
Closes #4176. Clears the `get_boxes_shape`/`to_mask` half of #4249; the indexing half stays open with its own pin.

`Boxes.get_boxes_shape` and `Boxes.to_mask` both call `self.to_tensor(mode, as_padded_sequence=True)`. `VideoBoxes` overrides `to_tensor` with only `(self, mode)`, so both raised `TypeError` on every `VideoBoxes` — including the containers `AugmentationSequential` builds for video box inputs (`kornia/augmentation/container/augment.py:635`).

## Change

The override now accepts and forwards `as_padded_sequence`. That keyword only changes the result of a list-backed container, and a `VideoBoxes` built from a `(B, T, N, 4, 2)` tensor is not list-backed, so **no call that works today changes its answer** — a `TypeError` becomes a result. `Boxes.clip` and `Boxes.clamp` pass the same keyword and are inherited too, so they are covered by the same change.

`get_boxes_shape` returns `(B, T, N)`, with the temporal axis restored the way `to_tensor` already restores it, rather than the flattened `(B*T, N)`. The batch-4c pin asserts only element counts, so either layout clears it; this one matches the rest of the class.

## Pins

- `test_convention_inherited_shape_and_mask_work_on_the_temporal_wrapper_4249` was a **strict xfail**; it now passes and loses the marker. Verified red on the merge base:
  `AssertionError: VideoBoxes.to_tensor() got an unexpected keyword argument 'as_padded_sequence'`.
- `test_wart_inherited_methods_break_on_the_temporal_wrapper_4249` is renamed to `test_wart_indexing_drops_the_temporal_size_4249` and narrowed to the surviving wart: `Boxes.__getitem__` builds the result with `type(self)(...)` and never sets `temporal_channel_size`, so a sliced wrapper's `to_tensor` raises `AttributeError`.
- The class `.. warning::` is updated to state only the indexing wart.

## What I ran

- `tests/geometry/test_boxes.py --device=cpu --dtype=float32`: **117 passed, 10 xfailed** at this head.
- The convention pin alone on the merge base: **1 failed**; at this head it passes.
- `grep -rnE "#4176|issues/4176" kornia/ tests/` — no surviving hit in `kornia/`.
- `ruff@0.16.6 check` and `format --check` clean on both touched files.

Not run here: MPS, CUDA and half-precision legs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -487,6 +487,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   transform or map batch broadcasts identically and a mismatched batch is rejected rather than silently
   returning the wrong cardinality. Outputs, autograd links and the documented empty-source policy are unchanged
   on CPU and CUDA. (#4032, #4354)
+* `VideoBoxes.get_boxes_shape()` and `VideoBoxes.to_mask()` no longer raise `TypeError`. Both are
+  inherited from `Boxes` and pass `as_padded_sequence=True` to `to_tensor`, but the `VideoBoxes`
+  override declared only `mode`, so every call on a `VideoBoxes` failed -- including the containers
+  `AugmentationSequential` builds for video box inputs. The override now accepts and forwards the
+  keyword; it only changes a list-backed container, and a `VideoBoxes` built from a
+  `(B, T, N, 4, 2)` tensor is not one, so no working call changes its result. Indexing a
+  `VideoBoxes` still drops `temporal_channel_size`, which is the remaining half of #4249.
+  (#4176, #4365)
 
 * Fixed `unproject_points_z1` depth shape handling for singleton and multi-axis batches,
   accepting both trailing-singleton and flat depth tensors. (#4355)
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -1154,8 +1154,8 @@ class VideoBoxes(Boxes):
                 changes the result of a list-backed container; a
                 :class:`VideoBoxes` built from a :math:`(B, T, N, 4, 2)` tensor
                 is not list-backed, so the value makes no difference there. The
-                keyword is accepted because the inherited :meth:`get_boxes_shape`,
-                :meth:`to_mask`, :meth:`clip` and :meth:`clamp` all pass it.
+                keyword is accepted because the inherited :meth:`get_boxes_shape`
+                and :meth:`to_mask` pass it.
 
         Returns:
             Tensor shaped :math:`(B, T, \ldots)` where :math:`T` is
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -1093,9 +1093,8 @@ class VideoBoxes(Boxes):
           original coordinates or dtype.
 
     .. warning::
-        :meth:`get_boxes_shape` and :meth:`to_mask` raise ``TypeError`` because the :meth:`to_tensor` override
-        does not accept the ``as_padded_sequence`` keyword they pass, and indexing returns a wrapper without
-        :attr:`temporal_channel_size`, so its :meth:`to_tensor` raises ``AttributeError``. Both are tracked in
+        Indexing returns a wrapper without :attr:`temporal_channel_size`, so its :meth:`to_tensor` raises
+        ``AttributeError``; this is the remaining half of
         `#4249 <https://github.com/kornia/kornia/issues/4249>`_. The inert ``validate_boxes`` flag is part of
         `#4177 <https://github.com/kornia/kornia/issues/4177>`_.
 
@@ -1143,21 +1142,26 @@ class VideoBoxes(Boxes):
         out.temporal_channel_size = temporal_channel_size
         return out
 
-    def to_tensor(self, mode: Optional[str] = None) -> torch.Tensor | list[torch.Tensor]:  # type: ignore[override]
+    def to_tensor(
+        self, mode: Optional[str] = None, as_padded_sequence: bool = False
+    ) -> torch.Tensor | list[torch.Tensor]:
         r"""Cast :class:`VideoBoxes` to a tensor with the temporal axis restored.
-
-        The ``as_padded_sequence`` keyword of :meth:`Boxes.to_tensor` is not
-        accepted; see the warning on the class.
 
         Args:
             mode: Output box format forwarded to :meth:`Boxes.to_tensor`. When
                 ``None``, uses the stored mode (``vertices_plus`` by default).
+            as_padded_sequence: Forwarded to :meth:`Boxes.to_tensor`. It only
+                changes the result of a list-backed container; a
+                :class:`VideoBoxes` built from a :math:`(B, T, N, 4, 2)` tensor
+                is not list-backed, so the value makes no difference there. The
+                keyword is accepted because the inherited :meth:`get_boxes_shape`,
+                :meth:`to_mask`, :meth:`clip` and :meth:`clamp` all pass it.
 
         Returns:
             Tensor shaped :math:`(B, T, \ldots)` where :math:`T` is
             :attr:`temporal_channel_size`.
         """
-        out = super().to_tensor(mode, as_padded_sequence=False)
+        out = super().to_tensor(mode, as_padded_sequence=as_padded_sequence)
         if isinstance(out, torch.Tensor):
             return out.view(-1, self.temporal_channel_size, *out.shape[1:])
         # If returns a list of boxes.
--- a/tests/geometry/test_boxes.py
+++ b/tests/geometry/test_boxes.py
@@ -1910,16 +1910,14 @@ class TestVideoBoxes(BaseTester):
         assert transformed.temporal_channel_size == 3
         self.assert_close(transformed.to_tensor(), boxes, atol=0.0, rtol=0.0)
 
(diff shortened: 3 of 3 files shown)


ducha-aiki (commented):
Reviewed at `a086eec86433224efaea953586af23738c5ff187` — the production fix is correct and this PR includes the CHANGELOG and consumer coverage missing from the older #4192, but one public docstring claim needs correction.

In `kornia/geometry/boxes.py:1122`, the new `as_padded_sequence` documentation says inherited `get_boxes_shape`, `to_mask`, `clip`, and `clamp` all pass the keyword. The actual call sites are `get_boxes_shape` and the two `to_mask` paths only. `Boxes` has no `clip` method, and `clamp` operates directly on `_data` without calling `to_tensor`. Please narrow that sentence to the two real consumers so the public API docs do not advertise nonexistent behavior.

The implementation itself matches the already reviewed #4192 fix: forwarding the keyword repairs both inherited consumers without changing ordinary tensor-backed `to_tensor` output. Local validation at this exact head passed the whole boxes file on CPU float32/float64: 222 passed / 20 xfailed, with the import resolving to the PR worktree.

One non-blocking test improvement remains from the #4192 review: the new mask assertion checks only element count and the final spatial dimensions, so it would accept either the current flattened-time shape `(B*T, N, H, W)` or a future restored-time shape `(B, T, N, H, W)`. Pin the full current shape (`(6, 1, 4, 5)` for this fixture) to make that convention explicit.

#4365 and #4192 solve the same #4176 defect and should not both merge. Because this PR already supplies the required release note and consumer test, I recommend keeping #4365 after the doc correction and closing the stale #4192 duplicate.

dchaudhari7177:
Addressed in `17dc34c9`, rebased onto current `main` (`a7e2db6b`, including #4372's `Boxes.clamp` change). The CHANGELOG conflict is resolved with every entry kept.

- **Docstring.** `VideoBoxes.to_tensor` now names only `get_boxes_shape` and `to_mask` as the inherited consumers. You are right that `Boxes` has no `clip` and that `clamp` works on `_data`, which is still the case after #4372.
- **Mask shape (non-blocking).** `test_convention_inherited_shape_and_mask_work_on_the_temporal_wrapper_4249` now asserts `mask.shape == (6, 1, 4, 5)`, which pins the current flattened-time `(B*T, N, H, W)` convention.
- **CHANGELOG.** While rebasing I found the entry cited `(#4176, #4361)`. #4361 doesn't exist; it now reads `(#4176, #4365)`.

`tests/geometry/test_boxes.py`: 268 passed, 4 xfailed on CPU float32/float64. With `main`'s `boxes.py` restored, the convention test fails. `ruff@0.16.6` check and format are clean. No CUDA run.

Agreed on #4192: it should not merge alongside this.

ducha-aiki (approved):
Delta re-review of #4365 at `17dc34c9` (was `a086eec8`; now rebased onto `a7e2db6b`). **Approve.** Thanks for the quick turnaround. All three items from the last review are fixed, and the fix itself still holds up on the new base.

**Prior asks, checked on this head**

- **Docstring consumers: fixed.** In `Boxes`, only two places call `to_tensor(..., as_padded_sequence=True)`: `get_boxes_shape` (`kornia/geometry/boxes.py:343`) and `to_mask` (`boxes.py:873`). The CPU loop and the vectorized path of `to_mask` share that one call. `Boxes` has no `clip`, and `clamp` still works on `_data` after #4372. The new `as_padded_sequence` entry (`boxes.py:1153-1158`) names exactly those two methods, and the class `.. warning::` now covers only the indexing wart.
- **Full mask-shape pin: fixed.** `assert mask.shape == (6, 1, 4, 5)` now pins the flattened-time layout. The old assertion (`shape[-2:] == (4, 5)` plus a `numel` check) also accepted a restored-time `(2, 3, 1, 4, 5)` mask. The new one rejects it.
- **CHANGELOG: fixed.** The entry sits under `## Unreleased` / `### Bug fixes` and cites `(#4176, #4365)`. Good catch on the `#4361` typo. It says what used to happen (`TypeError`) and what happens now.

**[P3] The PR description still makes the `clip`/`clamp` claim.** The "Change" section still says "`Boxes.clip` and `Boxes.clamp` pass the same keyword and are inherited too, so they are covered by the same change." That is the claim you just removed from the docstring. Please trim it, since the description can end up in the squash commit message.

**[P3] Optional: pin the `get_boxes_shape` layout as well.** The description argues for `(B, T, N)`, and #4176 asked for that layout. The code does return it: `(2, 3, 2)` in my probe, equal elementwise to the flattened `Boxes` reference. But `test_convention_inherited_shape_and_mask_work_on_the_temporal_wrapper_4249` still checks only `heights.numel() == widths.numel() == 6`, so a flattened `(6, 1)` result would also pass. Adding `assert heights.shape == widths.shape == (2, 3, 1)` next to the mask assertion would pin both layouts. The class now has two of them: the extents keep the time axis and the mask flattens it. A one-line bullet in the `VideoBoxes` Convention block would record that too. Not blocking.

**Verified on this head** (kornia imported from the PR worktree, torch 2.14.0)
- `tests/geometry/test_boxes.py`: **268 passed, 4 xfailed** on CPU float32+float64. **135 passed, 5 skipped, 2 xfailed** on MPS float32. `-k TestVideoBoxes`: **10 passed** each on CPU float16 and bfloat16 (known-failure profiles), and **10 passed** on torch 2.5.1 CPU float32.
- Base check: I copied this head's test file onto `main` (`dc0af1a4`). The convention pin **fails** there (`AssertionError: VideoBoxes.to_tensor() got an unexpected keyword argument 'as_padded_sequence'`). The narrowed indexing wart pin passes on both, as it should, since that half of #4249 is unchanged.
- Probe (`B=2, T=3, N=2`, CPU and MPS). On `main`, both `get_boxes_shape` and `to_mask` raise `TypeError`. On this head:
  - `get_boxes_shape` returns `(2, 3, 2)` and equals the flattened `Boxes` result elementwise.
  - `to_mask(9, 10)` returns `(6, 2, 9, 10)`, `torch.equal` to the flattened `Boxes.to_mask`. The vectorized branch (forced through `is_exporting`) gives the same mask, and the source `data` is left untouched.
  - `to_tensor(mode)` and `to_tensor(mode, as_padded_sequence=True)` are identical for `xyxy`, `xywh`, `vertices_plus` and the default mode.
- `grep -rnE "#4176|issues/4176" kornia/ tests/`: no hit in `kornia/`. The one hit in `tests/` is the updated wart-pin comment. The remaining `#4249` references (class warning, wart pin) correctly describe the open indexing half.
- Doctests in `kornia/geometry/boxes.py`: 8 passed. `ruff@0.16.6` check and format are clean on both files. `git merge-tree` against current `main` is clean; the 4 commits it is behind don't touch geometry.

#4192 fixes the same bug with the same one-line change. It should be closed in favour of this PR.

Issue #2042, opened 2022-11-29
[Feature] Add `reprojectImageTo3d`


edgarriba:
## 🚀 Feature

Implement opencv `reprojectImageTo3d` (or similar name) to `kornia.geometry.stereo` (or `kornia.stereo`)
https://docs.opencv.org/3.4/d9/d0c/group__calib3d.html#ga1bc1152bd57d63bc524204f21fde6e02

## Motivation

This can be easily be vectorized which torchdynamo can bootstrap in a high order magnitude.
______________________________________________________________________

#### Consider also to contribute to Kornia universe projects :)

<sub>

- [**Tutorials**](https://github.com/kornia/tutorials): our repository containing the tutorials.

</sub>

AnimeshMaheshwari22:
Hi. So the input for this would be disparity map right?

edgarriba:
@AnimeshMaheshwari22 yes

edgarriba:
If anyone is interested in this, feel free to re-open the ticket. Thanks!

PR #4366, opened 2026-09-08
fix(stereo): stop reproject_disparity_to_3D transposing the pixel indices (closes #4269)

### LINKED ISSUE #4269: W1b: `reproject_disparity_to_3D` — `u` and `v` are swapped, and the existing real-data pin encodes the swap

dchaudhari7177:
Closes #4269.

`create_meshgrid(..., normalized_coordinates=False)` returns `(x, y)`, so `uv[..., 0]` is the **column** and `uv[..., 1]` the **row**. The shared body unbound it as `v, u`, feeding the column into `v` and the row into `u`, so `X` was computed from the row and `Y` from the column. Every returned point was the value belonging to its transposed pixel: kornia's answer at `(row, col)` was what `cv2.reprojectImageTo3D` puts at `(col, row)` — the OpenCV semantics this function was added to provide (#2042).

A square rig with `fx == fy` and `cx == cy` agrees only where `row == col`, which is why this was silent rather than absent.

## Verification

On an asymmetric 3×5 rig (`cx=4 != cy=3`), before / after, against the hand-computed OpenCV formula:

```
                     base                          fixed = opencv
(row 0, col 2)  [-0.2,  -0.05, 5.0]            [-0.1,  -0.15, 5.0]
(row 1, col 0)  [-0.15, -0.15, 5.0]            [-0.2,  -0.1,  5.0]
(row 2, col 4)  [-0.1,   0.05, 5.0]            [ 0.0,  -0.05, 5.0]

X along a row (col varies):  base [-0.2, -0.2, -0.2, -0.2, -0.2]   fixed [-0.2, -0.15, -0.1, -0.05, 0.0]
X down a column (row varies): base [-0.2, -0.15, -0.1]             fixed [-0.2, -0.2, -0.2]
```

All 15 pixels match the OpenCV formula at this head.

## The real-data fixture

`_RealTestData` stored ten points as **ten rows of one column**, while its own comment ("first 10 cols of 1 row") and its ground truth describe **one row of ten columns**. That transpose is exactly why `test_reproject_disparity_to_3D_real` passed against the swapped code.

Laying the fixture out as its comment says makes the fixed code reproduce the stored ground truth to **9.2e-5**. **No ground-truth number is edited** — only the shape the same literals are arranged in.

## What I ran

| | base | head |
|---|---|---|
| `test_stereo.py` float32 | 6 failed, 12 passed | **18 passed** |
| `test_stereo.py` float64 | — | **18 passed** |
| `test_stereo.py` float16 | 3 failed, 12 passed | 3 failed, 15 passed |

The three float16 failures are `test_reproject_disparity_to_3D_real` and are **identical on base and head** — real-data precision at half, untouched by this change. The new test passes at float16.

`ruff@0.16.6 check` / `format --check` clean. Not run here: MPS, CUDA.

## Note on impact

This changes output for every non-square input, so it is a behaviour change as well as a fix — the previous values were the transposed-pixel answer. Flagging in case you would rather file it under Breaking changes than Bug fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -396,8 +396,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   answer at `(row, col)` was what `cv2.reprojectImageTo3D` puts at `(col, row)`, the OpenCV
   semantics this function was added to provide (#2042). A square rig with `fx == fy` and
   `cx == cy` agrees only on the diagonal, so the error was silent rather than absent. Output
-  changes for every non-square input. The real-data regression fixture stored ten points as ten
-  rows of one column while its own comment and its ground truth describe one row of ten columns,
+  changes at every pixel with `row != col`, square inputs included. The real-data regression
+  fixture stored ten points as ten rows of one column while its own comment and its ground truth
+  describe one row of ten columns,
   which is why it passed against the swapped code; it is now laid out as the comment says, and
   the fixed code reproduces the unchanged ground-truth values. #4317's two wart pins for this defect
   are retired and its strict-xfail convention test is now an ordinary passing regression. (#4269, #4366)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -399,7 +399,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   changes for every non-square input. The real-data regression fixture stored ten points as ten
   rows of one column while its own comment and its ground truth describe one row of ten columns,
   which is why it passed against the swapped code; it is now laid out as the comment says, and
-  the fixed code reproduces the unchanged ground-truth values. (#4269, #4366)
+  the fixed code reproduces the unchanged ground-truth values. #4317's two wart pins for this defect
+  are retired and its strict-xfail convention test is now an ordinary passing regression. (#4269, #4366)
 
 * `RandomRain` with `same_on_batch=True` now samples the same number of rain drops for all samples in the batch. (#4453)
 * `RandomRain` now rejects drop heights equal to the image height and absolute drop widths equal to the image
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -387,6 +387,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `RandomThinPlateSpline(scale=0.0)` now generates zero control-point displacement instead of
   raising an error when constructing a degenerate uniform distribution. (#4463)
+
+* `reproject_disparity_to_3D` (the `StereoCamera` method and the module-level function, which share
+  one body) no longer transposes the two pixel indices. The pixel meshgrid was unbound as
+  `v, u = torch.unbind(uv, dim=-1)`, but `create_meshgrid(normalized_coordinates=False)` returns
+  `(x, y)`, so the column fed `v` and the row fed `u`: `X` was computed from the row and `Y` from
+  the column. Every returned point was the value belonging to its transposed pixel -- kornia's
+  answer at `(row, col)` was what `cv2.reprojectImageTo3D` puts at `(col, row)`, the OpenCV
+  semantics this function was added to provide (#2042). A square rig with `fx == fy` and
+  `cx == cy` agrees only on the diagonal, so the error was silent rather than absent. Output
+  changes for every non-square input. The real-data regression fixture stored ten points as ten
+  rows of one column while its own comment and its ground truth describe one row of ten columns,
+  which is why it passed against the swapped code; it is now laid out as the comment says, and
+  the fixed code reproduces the unchanged ground-truth values. (#4269, #4366)
+
 * `RandomRain` with `same_on_batch=True` now samples the same number of rain drops for all samples in the batch. (#4453)
 * `RandomRain` now rejects drop heights equal to the image height and absolute drop widths equal to the image
   width with the documented validation error, instead of allowing boundary-sized drops to reach an internal
--- a/kornia/geometry/camera/stereo.py
+++ b/kornia/geometry/camera/stereo.py
@@ -80,9 +80,7 @@ class StereoCamera:
           enumerates, described in the Convention block on
           :class:`~kornia.geometry.camera.pinhole.PinholeCamera`.
         - ``u`` is the **column** index and ``v`` the **row** index, as in ``cv2.reprojectImageTo3D``:
-          :math:`X = (u - c_x) Z / f_x` and :math:`Y = (v - c_y) Z / f_y`. The two were transposed until
-          `#4269 <https://github.com/kornia/kornia/issues/4269>`_, so output changes for any
-          non-square input.
+          :math:`X = (u - c_x) Z / f_x` and :math:`Y = (v - c_y) Z / f_y`.
 
     .. warning::
         Several of the constructor guards do not enforce the contract above. A differing ``cx`` is
--- a/kornia/geometry/camera/stereo.py
+++ b/kornia/geometry/camera/stereo.py
@@ -399,8 +399,13 @@ def reproject_disparity_to_3D(disparity_tensor: torch.Tensor, Q_matrix: torch.Te
 
     uv = create_meshgrid(rows, cols, normalized_coordinates=False, device=device, dtype=dtype)
     uv = uv.expand(batch_size, -1, -1, -1)
-    v, u = torch.unbind(uv, dim=-1)
-    v, u = torch.unsqueeze(v, -1), torch.unsqueeze(u, -1)
+    # create_meshgrid(normalized_coordinates=False) returns (x, y), so uv[..., 0] is the
+    # column and uv[..., 1] is the row. u is the column and v the row, as in
+    # cv2.reprojectImageTo3D, whose semantics this function provides (#2042). Unbinding
+    # them the other way round fed the row into u and the column into v, which transposed
+    # the two pixel indices in the result.
+    u, v = torch.unbind(uv, dim=-1)
+    u, v = torch.unsqueeze(u, -1), torch.unsqueeze(v, -1)
     uvd = torch.stack((u, v, disparity_tensor), 1).reshape(batch_size, 3, -1).permute(0, 2, 1)
     points = transform_points(Q_matrix, uvd).reshape(batch_size, rows, cols, 3)
 
--- a/kornia/geometry/camera/stereo.py
+++ b/kornia/geometry/camera/stereo.py
@@ -79,16 +79,10 @@ class StereoCamera:
         - the pixels are the integer pixel centres that :func:`~kornia.geometry.grid.create_meshgrid`
           enumerates, described in the Convention block on
           :class:`~kornia.geometry.camera.pinhole.PinholeCamera`.
-
-    .. warning::
-        :meth:`~kornia.geometry.camera.stereo.StereoCamera.reproject_disparity_to_3D` unbinds that pixel grid
-        as ``v, u``, while :func:`~kornia.geometry.grid.create_meshgrid` returns it as ``(x, y)``, so ``X`` is
-        computed from the **row** index and ``Y`` from the column index -- the opposite of
-        ``cv2.reprojectImageTo3D``, whose semantics this function was added to provide. The repository's own
-        real-data test passes only because its fixture lays a one-row, ten-column strip out as ten rows of one
-        column, which the swap cancels; its stored numbers are correct OpenCV output for that strip, so a fix
-        corrects the layout and keeps every literal. Tracked as
-        `#4269 <https://github.com/kornia/kornia/issues/4269>`_.
+        - ``u`` is the **column** index and ``v`` the **row** index, as in ``cv2.reprojectImageTo3D``:
+          :math:`X = (u - c_x) Z / f_x` and :math:`Y = (v - c_y) Z / f_y`. The two were transposed until
+          `#4269 <https://github.com/kornia/kornia/issues/4269>`_, so output changes for any
+          non-square input.
 
     .. warning::
         Several of the constructor guards do not enforce the contract above. A differing ``cx`` is
--- a/tests/geometry/camera/test_stereo.py
+++ b/tests/geometry/camera/test_stereo.py
@@ -124,7 +124,11 @@ class _RealTestData(BaseTester):
             ],
             device=device,
             dtype=dtype,
-        ).permute(0, 2, 3, 1)
+        )
+        # The literal is already (B, rows=1, cols=10, 1). It used to be permuted to
+        # (B, 10, 1, 1) -- ten rows of one column -- which contradicts the comment above
+        # and the ground truth below, and made this test pass only because the pixel
+        # indices were swapped inside reproject_disparity_to_3D (#4269).
         return disp.expand(batch_size, -1, -1, -1)
 
     @staticmethod
@@ -147,7 +151,8 @@ class _RealTestData(BaseTester):
             ],
             device=device,
             dtype=dtype,
-        )
+        ).permute(0, 2, 1, 3)
+        # Same ten points, laid out as the one row of ten columns the comment describes.
 
         return pc.expand(batch_size, -1, -1, -1)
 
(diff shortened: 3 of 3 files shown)


ducha-aiki (commented):
Reviewed at `5cf4d0dc`. The axis-order fix is correct: `create_meshgrid` returns `(x, y)`, so unbinding as `u, v` restores column/row semantics. The asymmetric formula test and relaid real-data fixture cover the prior false-green shape. Local stereo validation passed all 36 CPU float32/float64 cases; the six new-regression half cases pass, while the six real-data half failures match existing known-failure manifests. All 40 hosted checks are green. The import resolved to the isolated PR checkout.

This head is still GitHub-CONFLICTING with current main (the merge-tree conflict is in `CHANGELOG.md`), so rebase and preserve the #4366 entry before merge. Also coordinate with #4317: its current head still documents and pins #4269's old swapped behavior; after both changes are combined, those wart tests/warning must be retired and the strict xfail made a normal passing contract. This is a sequencing requirement because #4317 has not merged.

ducha-aiki (changes requested):
Re-reviewed integration at `5cf4d0dc` after #4317 merged into main (`f36ab6de`). The axis correction remains sound, but the previously conditional follow-up is now required before merging.

[P1] Include the newly merged #4269 test and documentation cleanup in the rebase. Applying this PR's source/test diff to current main (excluding the CHANGELOG conflict) and running `tests/geometry/camera/test_stereo.py --device=cpu --dtype=float32,float64 -k 4269` produces **6 failures**: the two wart tests fail for both dtypes and `test_convention_reproject_disparity_uses_the_column_as_u_4269` becomes a strict XPASS for both. Remove or invert the two obsolete wart assertions, promote the convention test to an ordinary passing regression, and update the `StereoCamera` warning that still describes the axis swap as present. Keep the corrected real-data layout and its unchanged ground-truth numbers.

This is an integration finding against current main, not a claim that the PR's older branch CI failed. GitHub currently reports a conflict. Please rebase with these updates and rerun the stereo tests; the earlier numerical review stands. Validation used an isolated worktree with its import path verified; no full-suite or CUDA run was performed.

dchaudhari7177:
Rebased on current `main` (`d8b0a0c`-era CHANGELOG conflict resolved, the #4366 entry preserved) and the #4317 follow-up is folded in — `2fcaf79`.

Your integration finding reproduced exactly: applying the fix to a tree carrying #4317's pins gives 6 failures, the two wart tests for both dtypes and the convention test as a strict XPASS for both. All three are addressed:

| #4317 artefact | now |
| --- | --- |
| `test_wart_reproject_disparity_reads_u_from_the_row_index_4269` | deleted; its `fx != fy` whole-map arm moved into the convention test, so a reading that swapped the focal lengths as well as the indices still cannot reproduce it |
| `test_wart_reproject_disparity_x_varies_with_the_row_and_y_with_the_column_4269` | inverted into `test_convention_reproject_disparity_x_varies_with_the_column_4269` |
| `test_convention_reproject_disparity_uses_the_column_as_u_4269` | strict xfail removed, ordinary passing regression |

I inverted the axis-dependence pin rather than deleting it: it is the only assertion here that names which *index* each output coordinate depends on rather than a literal, so it survives any change of fixture and is worth keeping as a contract. The literal-value pin was redundant with the convention test once its `fy = 50` arm moved across, so that one is gone.

`StereoCamera`'s warning describing the swap as present is removed, and the Convention block now states the contract positively (`u` is the column, `v` the row, as in `cv2.reprojectImageTo3D`) with a note that output changes for non-square input.

The corrected real-data layout and its unchanged ground-truth numbers are untouched, as you asked.

**Verification:** `tests/geometry/camera/test_stereo.py` — 102 passed on cpu for float32 and float64, no xfail/xpass left in the file. `ruff` 0.16.6 (the version `.pre-commit-config.yaml` pins) check and format clean. No CUDA run.

ducha-aiki (changes requested):
Delta re-review at `2fcaf790` (rebase of `5cf4d0dc`). The axis correction stands and the #4317 follow-up is folded in as asked; one repo-convention item blocks merge.

**[P2] `kornia/geometry/camera/stereo.py:84` still links `#4269`.** AGENTS.md requires `grep -rnE "#4269|issues/4269" kornia/` to come back empty before a change that closes #4269 merges, and this head returns one hit: the new Convention bullet ends with ``The two were transposed until `#4269 <https://github.com/kornia/kornia/issues/4269>`_, so output changes for any non-square input.`` The sentence is historical rather than a live defect note, so the contract is not falsified, but the closed issue stays anchored in the API docs and the completeness grep will keep flagging it. Please drop the issue link from the docstring (the CHANGELOG entry already carries the history) -- ending the bullet after the two formulas is enough.

**[P3] "output changes for any non-square input" (docstring line 85, CHANGELOG line 392) understates the change.** Transposing the index pair is a no-op only where `row == col`, so square inputs change too: on a 5x5 map with `fx == fy == 100`, `cx == cy == 3`, base and head differ in 20 of 25 cells, agreeing only on the diagonal. Suggest "output changes at every pixel with `row != col`".

**CI:** the two red macOS 2.5.1 legs each fail only on `tests/contrib/test_super_resolution.py::TestSuperResolutionBuilders::test_to_onnx_exports[RRDBNetBuilder] - Failed: Timeout (>120.0s)` (`1 failed, 9988 passed`); that is the known RRDBNet ONNX export timeout on those runners, unrelated to this PR, and a rerun is all it needs.

**Verified on this head** (isolated worktree, `kornia.__file__` checked in every snippet): `test_stereo.py` 102 passed on CPU float32/float64 with no xfail or xpass left; cpu-float16 48 passed / 3 xfailed and cpu-bfloat16 36 passed / 15 xfailed with only pre-existing #4153 manifest lines and no XPASS; MPS float32 51 passed. An independent probe on a 4x9 random disparity map (`fx=120, fy=80, cx=7, cy=2`) matches `Q @ [u=col, v=row, d, 1]` to 2.8e-17 on this head and is off by up to 0.92 on base; the module-level function and the method agree, and float64 gradcheck passes. The four touched test families (`..._uses_the_column_for_x`, `..._real`, both `test_convention_*_4269`) fail on base (16 failures) and pass here. Prior asks: both wart pins retired, the strict xfail promoted, the real-data fixture relaid with every ground-truth literal unchanged, the #4366 CHANGELOG entry under `Unreleased / Bug fixes` citing `(#4269, #4366)` and stating the old behaviour. `ruff@0.16.6` check/format clean. No CUDA run.

dchaudhari7177:
Both addressed in `ac551baf`, rebased onto current `main` (`a7e2db6b`). The CHANGELOG conflict with #4459/#4463/#4372 is resolved with every entry kept.

- **[P2]** The `#4269` link is gone from the Convention bullet, which now ends after the two formulas. `grep -rnE "#4269|issues/4269" kornia/` comes back empty.
- **[P3]** The CHANGELOG entry now says output changes *at every pixel with `row != col`, square inputs included*. The docstring sentence that carried the understatement was the one removed for P2.

`tests/geometry/camera/test_stereo.py`: 102 passed on CPU float32/float64. `ruff@0.16.6` check and format are clean on `stereo.py`. No CUDA run. Noted on the macOS 2.5.1 RRDBNet ONNX timeout; the rebase re-runs CI anyway.

ducha-aiki (approved):
Delta re-review at `ac551baf` (rebase of `2fcaf790` onto `a7e2db6b` plus the new commit `ac551baf`). Both items from the last review are addressed and I found nothing new, so approving. Thanks for the quick turnaround.

- **[P2] closed-issue link:** gone. `grep -rnE "#4269|issues/4269" kornia/` is empty on this head. The Convention bullet now ends after the two formulas. The four hits left in `tests/geometry/camera/test_stereo.py` (lines 131, 309, 452, 468) are history comments on passing regressions, with no xfail or wart pin left for #4269.
- **[P3] wording:** the CHANGELOG entry now says output changes "at every pixel with `row != col`, square inputs included". That matches a 5x5 square rig (`fx == fy == 100`, `cx == cy == 3`): base differs from the OpenCV reference in 20 of 25 cells, none of them on the diagonal, and this head differs in 0.

**Verified on this head** (isolated worktrees, `kornia.__file__` checked in every snippet):
- **Rebase:** content-identical apart from `ac551baf`. The net `kornia/` + `tests/` patch of `8a34324f` against `a7e2db6b` is byte-identical to that of `2fcaf790` against its old base, and main did not touch either stereo file in between. `ac551baf` only removes the docstring sentence and rewords the CHANGELOG.
- **CHANGELOG:** against `a7e2db6b` the branch only adds lines (0 removed): the single #4366 entry under `Unreleased / Bug fixes`. `git merge-tree` against current main (`dc0af1a4`) merges cleanly, and the merged CHANGELOG differs from main's only by that entry, so no main entry is lost or duplicated.
- **Tests:** `test_stereo.py` passes 102 on CPU float32/float64. cpu-float16 gives 48 passed / 3 xfailed and cpu-bfloat16 36 passed / 15 xfailed, with no XPASS. The head test file copied onto main fails 16 tests there (`_real`, `_uses_the_column_for_x`, and both `test_convention_*_4269`) and passes the other 86.
- **Probe:** on a random 4x9 disparity map with `fx=120, fy=80, cx=7, cy=2`, the output matches `Q @ [col, row, d, 1]` to 1.8e-15 on this head and is off by up to 1.03 on base. The method and the module-level function agree exactly, and float64 gradcheck passes.
- **Lint and PR state:** `ruff@0.16.6` check and format are clean on both changed files. CI is 40/40 green, `closingIssuesReferences` is just #4269, and there is no competing PR. No CUDA or MPS run on this delta.

PR #4367, opened 2026-09-08
fix(core): drop a transfer validate refused, and report the rejection

aminmiral:
Two follow-ups to #4332, both in `download_file_from_url`.

**A fresh transfer that `validate` refused is kept.** `_drop_failed_download` is gated on `more_sources`, so when no source is left to use the emptied path the refused bytes stay in the cache. That is every `download_hf_file` caller, since it passes one URL: `from_pretrained_hf()` on a cold cache with a connection cut after a 2xx fails *and* leaves the truncated file behind, and only the next call clears it.

The gate is right in `load_state_dict_from_url` — a load failure is ambiguous, and deleting on one would sometimes destroy an intact multi-gigabyte checkpoint behind a bad `map_location`. Reaching the same branch here means the transfer succeeded and `validate` refused what it wrote, which is a verdict on the file itself. Dropped unconditionally; the gate in `load_state_dict_from_url` is untouched.

**The `refetch_note` swap is missing.** When the re-attempt fails offline, `last_exc` is a `URLError` on top of the rejection that names the truncation, so the `RuntimeError` reported the network and pointed the caller at an entry that is intact and, by then, restored. Now reports the rejection with the refetch failure as context, as `load_state_dict_from_url` does.

Two tests, both failing on the parent commit. `tests/core`, `tests/models` and `tests/test_api_surface.py` pass (570 passed, 18 skipped); ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -436,6 +436,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   step: a rejected entry is moved aside, the next source is tried, and the discarded source is
   re-fetched once. Without `validate` the behaviour is unchanged. (#4309, #4332)
 
+* A fresh transfer that `validate` rejects is now removed even when no other source could have used
+  the emptied path -- which is every `download_hf_file` caller, since it passes one URL. Those bytes
+  arrived during the call and were refused, so unlike the ambiguous load failures
+  `load_state_dict_from_url` weighs, keeping them would end the call having added a poisoned entry to
+  a cache that had none. An offline re-fetch also reports the validator's rejection rather than the
+  network error stacked on top of it, so the message names the file rather than an entry that is
+  intact and, by then, restored. (#4367)
+
 * `RenderingDeFMO` (used by `DeFMO`) no longer crashes on a half-precision forward pass. Its rendering
   time-steps (`times`) were a plain Python attribute, not a registered buffer, so `nn.Module.to()` never
   moved it; `forward` re-derived `times`'s device from the input on every call but never its dtype, so
--- a/kornia/core/download.py
+++ b/kornia/core/download.py
@@ -533,8 +533,12 @@ def _drop_failed_download(path: str) -> None:
     returns the path to the state the call found, which is also the empty path the
     next source needs in order to be fetched at all.
 
-    Only called while a later source remains; see :func:`load_state_dict_from_url`
-    for why the final source keeps what it wrote.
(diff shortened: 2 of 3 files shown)


ducha-aiki (changes requested):
The implementation fixes both reported follow-ups correctly, and the new tests cover cold-cache cleanup and the preserved validation cause. Two documentation updates remain before merge:

- **P2 — `kornia/core/download.py:802-810`:** the public `download_file_from_url` docstring still says a validation rejection is treated “exactly” like a load failure: quarantined and re-fetched once. This patch creates a different contract for a freshly transferred file: `_drop_failed_download` deletes it unconditionally, and a one-URL call ends without re-fetching it. Describe cached-entry quarantine and fresh-transfer deletion separately. The helper docstring around line 524 also still says `_drop_failed_download` is only called while a later source remains, which is no longer true.
- **P2 — `CHANGELOG.md:434-437`:** record this follow-up: a fresh transfer rejected by `validate` is removed even with one source, and an offline re-fetch error reports the original validator rejection. Cite #4367.

Validation at `3afd4117c5b6a376657cfa6d741b6c4c313f21fb`: all 93 tests in `tests/core/test_download.py` passed, `git diff --check` is clean, all 39 hosted checks are green, and `import kornia` resolved to the isolated PR worktree.

aminmiral:
Both docs items done in 324d1db.

- `download_file_from_url` now describes the two cases separately: a rejected **cache entry** is quarantined, re-fetched once and restored if nothing replaces it; a rejected **fresh transfer** is deleted outright with no re-fetch, so a one-URL call — every `download_hf_file` caller — raises with the cache empty rather than poisoned.
- `_drop_failed_download`'s "only called while a later source remains" now names both callers and why they differ: a load failure does not establish the bytes are bad, a `validate` rejection does.
- CHANGELOG entry added citing #4367.

`tests/core/test_download.py`: 93 passed. Doctests, ruff check/format and `git diff --check` clean.

ducha-aiki (approved):
Re-review of exact head 51887840c9f45dcdb7cffb0ef4fe70bb3a2b836a. The requested documentation now separates rejected cached entries from rejected fresh transfers, the helper documentation names both caller contracts, and the release note records the fresh-transfer deletion and preserved validator cause with #4367. I found no remaining correctness blocker.

Local validation verified the isolated checkout import and passed all 93 downloader tests. All 40 hosted checks are green and GitHub reports the head mergeable.

ducha-aiki (approved):

PR #4368, opened 2026-09-08
fix(geometry): correct Kannala-Brandt Jacobian

### LINKED ISSUE #4277: W1b: `dx_distort_points_kannala_brandt` — the analytic Jacobian is wrong, and its test and doctest pin the wrong value
`dx_distort_points_kannala_brandt` returns a matrix that is not the Jacobian of `distort_points_kannala_brandt`. Autograd and central finite differences agree with each other to three decimal places and both disagree with the analytic result — at the probe point by `62.58` in absolute terms, on entries of order `10^2`. Transposing the analytic matrix does not fix it, so it is not an index-order slip.

The two sibling Jacobians in the same family, `dx_distort_points_affine` and `dx_project_points_z1`, agree with autograd to `0.0`. At the origin the function returns `[[nan, nan], [nan, nan]]`, and the source carries `# TODO: return identity matrix if a and b are zero`.

**The existing test and the docstring example both pin the wrong value.** `tests/geometry/camera/test_distortion.py::TestDistortionKannalaBrandt::test_dx_distort_points_kannala_brandt` asserts `[[1191.5316162109375, 282.3212890625], [282.3212890625, 1615.0135498046875]]`, which is exactly what the analytic code returns; autograd at the same input gives `[[1221.1802978515625, 341.61865234375], [341.61865234375, 1733.6082763671875]]`. The `>>>` example in the docstring prints the analytic value for a third parameter set, where the analytic/finite-difference ratio on the diagonal is `[0.926, 0.5201]`.

## Repro

```python
import torch
from kornia.geometry.camera import (
    distort_points_affine, distort_points_kannala_brandt, dx_distort_points_affine,
    dx_distort_points_kannala_brandt, dx_project_points_z1, project_points_z1)

def fd_jac(f, p, h=1e-4):
    p = p.double()
    cols = []
    for i in range(p.shape[-1]):
        e = torch.zeros_like(p); e[i] = h
        cols.append((f((p + e).float()).double() - f((p - e).float()).double()) / (2 * h))
    return torch.stack(cols, dim=-1)

# (a) the input of the existing test, tests/geometry/camera/test_distortion.py
#     ::TestDistortionKannalaBrandt::test_dx_distort_points_kannala_brandt
p = torch.tensor([1.0, 2.0])
par = torch.tensor([600.0, 600.0, 319.5, 239.5, 0.1, 0.2, 0.3, 0.4])
print("analytic dx_distort_points_kannala_brandt:", dx_distort_points_kannala_brandt(p, par).tolist())
print("the value the existing test asserts      :", [[1191.5316162109375, 282.3212890625], [282.3212890625, 1615.0135498046875]])
print("autograd jacobian                        :", torch.autograd.functional.jacobian(lambda q: distort_points_kannala_brandt(q, par), p).tolist())
print("central finite differences               :", fd_jac(lambda q: distort_points_kannala_brandt(q, par), p).tolist())

# (b) a second input; the disagreement is not a conditioning artefact
p2 = torch.tensor([0.5, 0.25])
par2 = torch.tensor([100.0, 100.0, 4.0, 3.0, 0.1, 0.01, 0.001, 0.0001])
a = dx_distort_points_kannala_brandt(p2, par2)
g = torch.autograd.functional.jacobian(lambda q: distort_points_kannala_brandt(q, par2), p2)
print("analytic :", a.tolist())
print("autograd :", g.tolist())
print("finite d :", fd_jac(lambda q: distort_points_kannala_brandt(q, par2), p2).tolist())
print("max abs analytic - autograd              :", (a - g).abs().max().item(),
      "| after transposing the analytic:", (a.T - g).abs().max().item())

# (c) the docstring example of dx_distort_points_kannala_brandt uses these params
dpar = torch.tensor([1000.0, 1000.0, 320.0, 280.0, 0.1, 0.01, 0.001, 0.0001])
ad = dx_distort_points_kannala_brandt(p, dpar); fdd = fd_jac(lambda q: distort_points_kannala_brandt(q, dpar), p)
print("docstring input: analytic", ad.tolist())
print("docstring input: finite d", fdd.tolist())
print("diagonal ratio analytic/fd               :", [round(ad[0,0].item()/fdd[0,0].item(), 4), round(ad[1,1].item()/fdd[1,1].item(), 4)])

# (d) the sibling Jacobians in the same family agree with autograd
apar = torch.tensor([100.0, 100.0, 4.0, 3.0])
print("dx_distort_points_affine vs autograd     :",
      (dx_distort_points_affine(p2, apar) - torch.autograd.functional.jacobian(lambda q: distort_points_affine(q, apar), p2)).abs().max().item())
p3 = torch.tensor([3.0, 6.0, 9.0])
print("dx_project_points_z1 vs autograd         :",
      (dx_project_points_z1(p3) - torch.autograd.functional.jacobian(project_points_z1, p3)).abs().max().item())
print("dx_distort_points_kannala_brandt at (0,0):", dx_distort_points_kannala_brandt(torch.tensor([0.0, 0.0]), par2).tolist())
```

## Observed

```
analytic dx_distort_points_kannala_brandt: [[1191.5316162109375, 282.3212890625], [282.3212890625, 1615.0135498046875]]
the value the existing test asserts      : [[1191.5316162109375, 282.3212890625], [282.3212890625, 1615.0135498046875]]
autograd jacobian                        : [[1221.1802978515625, 341.61865234375], [341.61865234375, 1733.6082763671875]]
central finite differences               : [[1224.9755859375, 341.796875], [347.900390625, 1733.3984375]]
analytic : [[22.0633487701416, -35.7770881652832], [-35.7770881652832, 75.72897338867188]]
autograd : [[84.64064025878906, -4.488434791564941], [-4.488434791564941, 91.373291015625]]
finite d : [[84.64813232421875, -4.5013427734375], [-4.482269287109375, 91.36199951171875]]
max abs analytic - autograd              : 62.577293395996094 | after transposing the analytic: 62.577293395996094
docstring input: analytic [[486.05072021484375, -213.55731201171875], [-213.55731201171875, 165.7147216796875]]
docstring input: finite d [[524.90234375, -137.02392578125], [-135.498046875, 318.603515625]]
diagonal ratio analytic/fd               : [0.926, 0.5201]
dx_distort_points_affine vs autograd     : 0.0
dx_project_points_z1 vs autograd         : 0.0
dx_distort_points_kannala_brandt at (0,0): [[nan, nan], [nan, nan]]
```

## Expected

The analytic Jacobian is re-derived so it matches autograd and central differences, and the origin case returns something finite (the source's own TODO proposes the identity; `distort_points_kannala_brandt` itself takes an affine branch at `radius_sq <= 1e-8`, so the Jacobian of *that* branch is the consistent answer).

Two things must move in the same change, and this is what makes it more than a one-line fix:

1. **`test_dx_distort_points_kannala_brandt`'s `expected` must be re-derived.** It currently asserts the wrong matrix and passes because of the bug.
2. **The docstring `>>>` example must be re-derived**, or the doctest fails.

Batch 5 leaves both untouched — the "additive pins only" rule is what keeps a documentation batch reviewable — and adds beside them a `test_wart_*` pinning today's analytic value and a strict `xfail` `test_convention_*` asserting the autograd value, so the fix flips the pair. A correctness fix here is welcome as a PR on those terms.

Note that `dx_distort_points_kannala_brandt` is not used inside kornia today, so nothing downstream changes.

## Classification

**Broken output** — an analytic derivative that disagrees with both autograd and finite differences, pinned by its own test. Idiom 4 on the `nan` at the origin; the wrong Jacobian itself is none of 1–6.

## Disposition

n/a — single symbol.

## Environment

Hrithikcrick:
﻿Fixes #4277

## Summary

- Re-derive `dx_distort_points_kannala_brandt` so it matches the derivative of `distort_points_kannala_brandt`.
- Return the affine Jacobian `diag(fx, fy)` for the small-radius branch, including the origin.
- Evaluate half/bfloat16 Jacobian arithmetic in float32 and cast back for numerical stability.
- Update the existing expected values and doctest output.
- Replace the old wart/xfail behavior with passing regression coverage and remove the stale bfloat16 known-failure entry.
- Add a changelog entry.

## Validation

- Analytic Jacobian checked against PyTorch autograd and central finite differences.
- Random float64 checks matched autograd to machine precision.
- `float16`, `bfloat16`, `float32`, and `float64` targeted tests pass.
- Full `tests/geometry/camera/test_distortion.py` run: 55 passed, 3 skipped.
- Doctests pass.
- TorchScript matches eager for all four dtypes.
- `torch.export` succeeds with zero error.
- `torch.compile(..., backend="eager", fullgraph=True)` matches eager for all four dtypes.
- Kornia lint passes and `git diff --check` is clean.

Note: the default Inductor backend could not be exercised locally on Windows because the MSVC `cl` compiler is not installed; the failure was environmental rather than a graph-capture error.


Hrithikcrick:
Update: the fresh CI run completed successfully.

All 40 checks now pass, including the Dynamo/Inductor job and the required collector check. The earlier XFeat numerical mismatch appears to have been transient; no unrelated XFeat changes were made in this PR.

ducha-aiki (commented):
The re-derived Jacobian is correct at `4f4b3550`: the focused dx tests pass for CPU float16/bfloat16/float32/float64 (12 passed), and the registered graph-capture case passes ONNX Runtime, torch.export, and torch.compile with zero graph breaks. Hosted CI is green.

Two small maintenance items remain before merge:

- `CHANGELOG.md:367` cites issue #4277 but not this PR; please add #4368 to the release-note citation.
- `tests/geometry/camera/test_distortion.py:174-175` still describes the Kannala-Brandt function as “NOT the Jacobian” and points to #4277. That surviving control-test comment contradicts the repaired behavior; update it alongside the repaired issue pin.

The import path was verified inside the isolated PR checkout.

Hrithikcrick:
Addressed the review feedback in 469e91fe: added #4368 to the CHANGELOG entry and updated the stale Kannala-Brandt Jacobian comment. The focused distortion tests and pre-commit checks pass locally. Thanks for the review!

ducha-aiki (dismissed):
Re-review of exact head 6826c87a47ddd3187368168c02cd25911ae7fe9e. Both follow-ups are resolved: the release note cites #4368, and the surviving affine control-test prose now describes the repaired Kannala-Brandt Jacobian accurately. The production Jacobian remains unchanged from the previously validated implementation.

Local validation verified the isolated checkout import and passed the complete distortion file on CPU float32/float64: 55 passed, 3 skipped. The earlier float16/bfloat16, ONNX Runtime, torch.export, and torch.compile evidence still applies to the unchanged implementation. All 40 hosted checks are green and GitHub reports the head mergeable.

PR #4370, opened 2026-09-08
fix(camera): handle zero-radius Kannala-Brandt undistortion

### LINKED ISSUE #4308: W1b: `undistort_points_kannala_brandt` — the principal point returns `nan` in float16, because both singularity guards underflow

victorwon2001:

`undistort_points_kannala_brandt` now handles an exact zero distorted radius with a structural mask instead of hardcoded epsilon guards. This makes the principal point return the exact origin in float16 and removes the additive `1e-8` radial-rescale bias that limited float64 round-trip precision, while preserving the existing 10 Gauss-Newton iterations and step denominator.

The convention documentation and regressions were updated accordingly, including mixed float16-points/float32-params behavior and representative float64 round trips.

Fixes #4308

## What did you check?

The focused principal-point and float64 precision regressions pass on CPU; the full float32/float64 camera-distortion module and both repository CPU half-precision profiles also pass with only their existing manifest xfails. MPS-specific cases were selected but skipped because this Linux host has no MPS device.

```text
$ pixi run --environment default test-module tests/geometry/camera/test_distortion.py --device=cpu --dtype=float16,float32,float64,bfloat16 -k 'principal_point_undistorts_to_the_origin or float64_round_trip_precision_4308'
$ pixi run --environment default test-module tests/geometry/camera/test_distortion.py --device=cpu --dtype=float32,float64
$ pixi run --environment default test-module tests/geometry/camera/test_distortion.py --known-failure-profile=cpu-float16 --xfail-known-failures
$ pixi run --environment default test-module tests/geometry/camera/test_distortion.py --known-failure-profile=cpu-bfloat16 --xfail-known-failures
$ pixi run --environment default typecheck
```

## How did AI help?

AI was used to inspect the affected implementation, callers, tests, and repository conventions; implement the zero-radius masking change; update the regression coverage and documentation; and run the checks above. The final diff was inspected for scope, compatibility, and unintended files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -374,6 +374,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   construct and then fail with an `IndexError` from inside `AffineTransform.distort`, naming neither
   `params` nor the camera; a rank-3 `(B, 1, N)` tensor used to construct, project, and silently return a
   `(1, 1, 2)` result. Both now raise `ValueError` from the constructor. (#4316, #4369)
+* `undistort_points_kannala_brandt` no longer collapses representable `float16` points next to the principal point
+  to the origin, and the exact principal-point path now has finite autograd gradients. The zero-radius decision is
+  made from the unsquared normalized coordinates, and the `float16` radius uses `float32` intermediates so its
+  squared value does not underflow; nonzero radial rescaling remains epsilon-free. (#4308, #4370)
 
 * `warp_affine`, `warp_perspective` and `remap` crashed on MPS for an empty destination -- a `dsize` with a
   zero dimension, or zero-sized `remap` maps -- with an internal
--- a/kornia/geometry/camera/distortion_kannala_brandt.py
+++ b/kornia/geometry/camera/distortion_kannala_brandt.py
@@ -163,9 +163,15 @@ def undistort_points_kannala_brandt(distorted_points_in_camera: torch.Tensor, pa
     un = (x - cx) / fx
     vn = (y - cy) / fy
 
-    rth2 = un * un + vn * vn
-    rth = rth2.sqrt()
-    nonzero_radius = rth > 0
+    nonzero_radius = (un != 0) | (vn != 0)
+    safe_un = torch.where(nonzero_radius, un, torch.ones_like(un))
+    safe_vn = torch.where(nonzero_radius, vn, torch.zeros_like(vn))
+    radius_dtype = safe_un.dtype
+    if radius_dtype == torch.float16:
+        safe_un = safe_un.float()
+        safe_vn = safe_vn.float()
+    rth = (safe_un * safe_un + safe_vn * safe_vn).sqrt().to(radius_dtype)
+    rth = torch.where(nonzero_radius, rth, torch.zeros_like(rth))
 
     safe_rth = torch.where(nonzero_radius, rth, torch.ones_like(rth))
     th = safe_rth.sqrt()
--- a/kornia/geometry/camera/distortion_kannala_brandt.py
+++ b/kornia/geometry/camera/distortion_kannala_brandt.py
@@ -119,21 +119,10 @@ def undistort_points_kannala_brandt(distorted_points_in_camera: torch.Tensor, pa
         - the inverse is a fixed number of Gauss-Newton steps rather than a closed form: the step count is not
           a parameter and there is no convergence test, so the round trip through
           :func:`distort_points_kannala_brandt` closes only to the accuracy that iteration has reached. The
-          step count cannot be raised by a caller. In ``float32`` the residual reaches the rounding floor; in
-          ``float64`` it stops at about ``1e-8`` on the normalized plane, because the final radial rescale
-          divides by ``r + 1e-8`` rather than ``r`` and so scales every result by ``1 - 1e-8 / r``. That is the
-          ``float64`` side of `#4308 <https://github.com/kornia/kornia/issues/4308>`_.
-          :func:`~kornia.geometry.camera.undistort_points_affine` is the closed-form contrast.
-        - three small constants guard the Newton start (``1e-16``), Newton denominator (``1e-12``), and final
-          radial rescale (``1e-8``), so a point at the principal point comes back as the origin rather than
-          ``nan`` -- as long as those constants are representable in the dtype of ``params``, which is the
-          dtype the whole body runs in.
-
-    .. warning::
-        All three guard constants underflow to zero in ``float16``. At the principal point the unguarded Newton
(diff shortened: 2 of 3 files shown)


ducha-aiki (changes requested):
[P1] The zero-radius mask is computed after the radius has already underflowed. In float16, a representable non-principal point can therefore be classified as the principal point and silently collapsed to the origin. On `8132f75e`, using params `[100, 50, 4, 3, .1, .01, .001, .0001]` and the nearest representable point to the right of the centre, `[4.00390625, 3]`, the normalized x coordinate is `3.9041e-05`. Its square underflows to zero, `nonzero_radius` becomes false, and `undistort_points_kannala_brandt` returns `[0, 0]` instead of approximately `[3.9041e-05, 0]`.

Please derive zero-ness from the unsquared normalized components and compute the radius without half-precision underflow, then cover the nearest representable off-centre float16 point. Masking before the singular square root would also prevent exact-origin gradients from remaining NaN; the current patch returns a finite value there but still produces NaN point and parameter gradients. See `kornia/geometry/camera/distortion_kannala_brandt.py:166-172`.

[P2] This user-visible bug fix has no Unreleased changelog entry. Add one under Bug fixes citing #4308 and PR #4370.

The ordinary CPU float32/float64 distortion suite passes (53 passed, 1 skipped, 2 expected xfails), and hosted CI is green. Validation used the isolated PR checkout with its import path verified; the blocker is the uncovered near-origin behavior above.

ducha-aiki (approved):
Re-review of exact head 0a50ad763b9cbab3fa78c7d9620c820ff6b36966. The previous blockers are resolved: the zero-radius decision now uses the unsquared normalized components, float16 radius arithmetic is widened before squaring, the exact-origin path is shielded before the singular operations, and the release note cites both the issue and this PR. The new tests pin the nearest representable off-centre float16 point and the point/parameter Jacobians at the principal point.

Local validation in the isolated checkout verified the import origin and passed the 12 focused principal-point, near-origin, and float64 round-trip selections (8 passed, 4 dtype skips). Hosted ordinary precision, MPS, typecheck, docs, and pre-commit checks are green. The float16 job is still running, and the only completed test failure is the known XFeat inductor tolerance mismatch in an unrelated file, so CI still needs completion and a clean rerun before merge.

PR #4372, opened 2026-09-08
fix(geometry): support unbatched boxes in Boxes.pad, unpad and clamp

### LINKED ISSUE #4244: Boxes.pad, unpad, and clamp fail on valid unbatched boxes

thegoodengineer:

`Boxes` documents and constructs an unbatched `(N, 4, 2)` container, but `pad`, `unpad` and `clamp` all assumed the batched `(B, N, 4, 2)` indexing, so valid unbatched boxes crashed:

```text
pad   False RuntimeError output with shape [1, 4] doesn't match the broadcast shape [1, 1, 4]
unpad False RuntimeError output with shape [1, 4] doesn't match the broadcast shape [1, 1, 4]
clamp False IndexError    too many indices for tensor of dimension 2
pad   True  torch.Size([1, 1, 4, 2])   [3.0, 8.0, 6.0, 8.0, 6.0, 9.0, 3.0, 9.0]
```

Adding a leading singleton batch axis to the same boxes made all three work, without changing the padding or bound tensors.

`pad` and `unpad` broadcast the padding as `(B, 1, 1)`, which does not fit the unbatched `(N, 4)` coordinate view. `clamp` materialized its bounds with `repeat(1, _data.size(1), 4)`, reading dim 1 as the box count, because the masked assignment it used needs a bound tensor of exactly the mask's shape.

A small helper shapes a per-image column for whichever rank the container holds, and `clamp` broadcasts its bounds through `maximum`/`minimum` instead of materializing them. That is elementwise-equivalent to the masked assignment including for `NaN` (`nan < bound` is `False`, and `maximum` returns `NaN`), and it is what unties the method from the batched indexing.

An unbatched container carries a single image, so a per-image tensor with more than one row is now rejected with a message naming the row count, rather than silently broadcasting against the box axis when the two happen to match.

Closes #4244

## What did you check?

The repository already had the pin for this, as a strict xfail. `test_convention_unbatched_geometry_operations_match_singleton_batch_4244` compares the unbatched result against the singleton-batch reference and checks the identity and non-mutation semantics of both `clamp` modes; it loses its mark here. `test_wart_unbatched_geometry_operations_raise_4244`, which pinned the three raises, is deleted, since it pinned behavior that no longer exists. The class Convention block and the three method docstrings drop the limitation sentence.

On unmodified `main` the convention pin fails in all 8 cells, and passes in all 16 here:

```text
BASE: 8 failed, 111 deselected
FAILED ...match_singleton_batch_4244[cpu-float32-2-pad-None]     - RuntimeError: output with shape [2, 4] ...
FAILED ...match_singleton_batch_4244[cpu-float32-2-clamp-False]  - RuntimeError: The size of tensor a (2) must match ...

HEAD: 16 passed, 210 deselected
```

Batched results are unchanged. I diffed this branch against the merge base over 40 tensors: float32 and float64, `(B, N)` of `(1,1)`, `(2,3)`, `(4,1)`, `(3,5)`, all of `pad`, `unpad`, `clamp(inplace=False)`, `clamp(inplace=True)`, plus the wholly-outside box whose collapse onto the boundary the docstring calls out. All 40 `torch.equal`.

```text
========== 2745 passed, 53 skipped, 18 xfailed, 4 xpassed, 132 warnings in 72.09s ==========

$ pytest tests/geometry/test_boxes.py tests/geometry/test_bbox.py --dtype=float32,float64
==================== 311 passed, 8 xfailed, 2 warnings in 8.73s ====================

$ pytest --doctest-modules kornia/geometry/boxes.py
================================ 8 passed in 0.42s ================================

$ ruff check / ruff format on both changed files
All checks passed!
```

One note on scope: `RandomCrop` is the in-tree consumer of `pad`/`unpad`, so I ran the crop augmentation subset too. It reports 3 failures, but the identical 3 fail on the merge base on this machine with an unrelated `OSError: pytest: reading from stdin while output is captured!`, so they are environmental rather than from this change. CI is the better signal there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -388,7 +388,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `depth_to_normals` now raises `ShapeError` when `H < 2` or `W < 2`, since surface normals need two
   tangent directions. Previously, singleton axes could yield zero or non-finite normals, and empty
   spatial dimensions failed inside padding. Inputs with both dimensions at least 2 are unchanged. (#4458)
-
+* `Boxes.pad`, `Boxes.unpad` and `Boxes.clamp` accept the unbatched `(N, 4, 2)` container the class
+  documents and constructs. All three assumed the batched `(B, N, 4, 2)` indexing: `pad` and `unpad`
+  broadcast the padding as `(B, 1, 1)`, which does not fit the unbatched `(N, 4)` coordinate view and
+  raised `RuntimeError: output with shape [1, 4] doesn't match the broadcast shape [1, 1, 4]`, and `clamp`
+  materialized its bounds with `repeat(1, _data.size(1), 4)` and raised `IndexError: too many indices for
+  tensor of dimension 2`. Adding a leading singleton batch axis to the same boxes made all three work.
+  They now broadcast against whichever rank the container holds and match the singleton-batch result;
+  an unbatched container carries one image, so a per-image tensor with more than one row is rejected.
+  Batched results are byte-identical, including for a non-finite bound: `clamp` stays comparison-based, so
+  a `NaN` bound leaves the coordinate alone rather than propagating into it (closes #4244). (#4372)
 * `RandAugment`'s `m` guard is exclusive at both ends, but its docstring and its error message
   both named the closed interval `[0, 30]`, so a user who asked for the maximum strength the
   message advertised got an exception saying `30` was in range. Both now read `(0, 30)`; the
--- a/kornia/geometry/boxes.py
+++ b/kornia/geometry/boxes.py
@@ -267,9 +267,7 @@ class Boxes:
         unvalidated, and ``'vertices'`` also subtracts one from fixed vertex positions, potentially deforming the
         input rather than rejecting it; this is tracked in `#4177 <https://github.com/kornia/kornia/issues/4177>`_.
         The unimplemented ``trim``, ``translate(method='fast')``, and tuple-bound ``clamp`` paths are tracked in
-        `#4017 <https://github.com/kornia/kornia/issues/4017>`_. The pad, unpad, and clamp operations fail for
-        unbatched containers even though the class accepts :math:`(N, 4, 2)` data; this is tracked in
-        `#4244 <https://github.com/kornia/kornia/issues/4244>`_.
+        `#4017 <https://github.com/kornia/kornia/issues/4017>`_.
 
     """
 
@@ -449,6 +447,22 @@ class Boxes:
         obj._data = _data
         return obj
 
+    def _broadcast_over_vertices(self, values: torch.Tensor) -> torch.Tensor:
+        """Shape a per-image ``(B, k)`` column so it broadcasts over ``self._data[..., i]``.
+
+        The batched container indexes as ``(B, N, 4)``, so the column needs one axis for the box
+        count. The unbatched ``(N, 4, 2)`` form carries a single implicit image and indexes as
+        ``(N, 4)``, so its ``(1, k)`` column already broadcasts and must not gain that axis.
+        """
+        if self._is_batched:
+            return values[..., None, :]
+        if values.size(0) != 1:
+            raise RuntimeError(
+                f"Unbatched (N, 4, 2) boxes carry a single image, so a per-image tensor must have one row. "
+                f"Got {values.size(0)}."
+            )
+        return values
+
     def pad(self, padding_size: torch.Tensor) -> Boxes:
         """Pad every box in place.
 
@@ -456,8 +470,10 @@ class Boxes:
 
         ``padding_size`` is ordered as ``(left, right, top, bottom)``. Only
         ``left`` and ``top`` change the coordinate origin; this method returns
-        ``self`` after adding those two values to every vertex. This operation
-        supports only batched :math:`(B, N, 4, 2)` containers.
+        ``self`` after adding those two values to every vertex. Both the batched
+        :math:`(B, N, 4, 2)` and the unbatched :math:`(N, 4, 2)` container are
+        supported; the unbatched form carries a single image, so ``padding_size``
+        must have exactly one row.
 
         Note:
             Padded :class:`~kornia.augmentation.RandomCrop` uses this method
@@ -470,8 +486,9 @@ class Boxes:
         """
         if not (len(padding_size.shape) == 2 and padding_size.size(1) == 4):
             raise RuntimeError(f"Expected padding_size as (B, 4). Got {padding_size.shape}.")
-        self._data[..., 0] += padding_size[..., None, :1].to(device=self._data.device)  # left padding
-        self._data[..., 1] += padding_size[..., None, 2:3].to(device=self._data.device)  # top padding
+        offset = padding_size.to(device=self._data.device)
+        self._data[..., 0] += self._broadcast_over_vertices(offset[..., :1])  # left padding
+        self._data[..., 1] += self._broadcast_over_vertices(offset[..., 2:3])  # top padding
         return self
 
     def unpad(self, padding_size: torch.Tensor) -> Boxes:
@@ -481,8 +498,10 @@ class Boxes:
 
         ``padding_size`` is ordered as ``(left, right, top, bottom)``. Only
         ``left`` and ``top`` change the coordinate origin; this method returns
-        ``self`` after subtracting those two values from every vertex. This
-        operation supports only batched :math:`(B, N, 4, 2)` containers.
+        ``self`` after subtracting those two values from every vertex. Both the
+        batched :math:`(B, N, 4, 2)` and the unbatched :math:`(N, 4, 2)` container
+        are supported; the unbatched form carries a single image, so
+        ``padding_size`` must have exactly one row.
 
         Args:
             padding_size: Per-batch padding in ``(left, right, top, bottom)``
@@ -491,8 +510,9 @@ class Boxes:
         """
         if not (len(padding_size.shape) == 2 and padding_size.size(1) == 4):
             raise RuntimeError(f"Expected padding_size as (B, 4). Got {padding_size.shape}.")
-        self._data[..., 0] -= padding_size[..., None, :1].to(device=self._data.device)  # left padding
-        self._data[..., 1] -= padding_size[..., None, 2:3].to(device=self._data.device)  # top padding
+        offset = padding_size.to(device=self._data.device)
+        self._data[..., 0] -= self._broadcast_over_vertices(offset[..., :1])  # left padding
+        self._data[..., 1] -= self._broadcast_over_vertices(offset[..., 2:3])  # top padding
         return self
 
     def clamp(
@@ -509,7 +529,9 @@ class Boxes:
             Bounds must be tensors with one ``(x, y)`` pair per batch element.
             Every vertex is clamped independently, so a box wholly outside the
             bounds collapses onto the nearest boundary instead of being removed.
-            This operation supports only batched :math:`(B, N, 4, 2)` containers.
+            Both the batched :math:`(B, N, 4, 2)` and the unbatched
+            :math:`(N, 4, 2)` container are supported; the unbatched form carries
+            a single image, so the bounds must have exactly one row.
 
         Coordinates below ``topleft`` are raised to the lower bound and
         coordinates above ``botright`` are lowered to the upper bound. The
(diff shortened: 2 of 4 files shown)


ducha-aiki (changes requested):
The unbatched shape fix works, but `clamp` changes an existing batched edge case.

[P2] `kornia/geometry/boxes.py:567-568`: `maximum`/`minimum` is not equivalent to the old comparison-based masked assignment when a bound is NaN. With a finite batched box, `topleft=[[nan, 0]]`, and `botright=[[10, 10]]`, current main leaves all coordinates finite because comparisons to NaN are false; this head turns every x coordinate into NaN. This contradicts the claim that existing batched behavior is preserved. Please preserve the comparison semantics with broadcasted `torch.where` (or equivalent) and pin this batched case.

[P2] Add the public `AugmentationSequential(RandomCrop(..., padding=1), data_keys=["input", "bbox_xyxy"])` regression requested in #4244's maintainer follow-up. The current tests cover only direct `Boxes` calls, so the user-facing adapter path is still unpinned.

[P2] The CHANGELOG ends with issue #4244; also cite this PR as #4372.

Local evidence at `892f5eff`: 220 passed / 6 expected xfails for `test_boxes.py` on CPU float32/float64; the public RandomCrop repro succeeds. The NaN-bound behavior was compared directly with current main, and imports were verified inside the intended checkouts. Hosted CI has no failures but float16 is still running.

thegoodengineer:
You found a real bug, and my "byte-identical" claim was wrong. All three addressed. New head `747ccef9`, rebased onto `29e05314`.

**P2 — the NaN bound.** Confirmed exactly as you described, on your inputs:

```text
BASE  [[nan, 0]] topleft -> [1.0, 2.0, 4.0, 2.0, 4.0, 3.0, 1.0, 3.0]   finite
HEAD  [[nan, 0]] topleft -> [nan, 2.0, nan, 2.0, nan, 3.0, nan, 3.0]   NaN on the x axis
```

`maximum`/`minimum` propagate a NaN operand; the masked assignment they replaced does not, because every comparison against NaN is False and the coordinate is simply not written. My comment asserted the two agreed "NaN included" — I had checked NaN in the *data*, where they do agree, and never NaN in the *bound*, where they do not. My A/B probe only ever generated finite bounds, so it could not have caught this.

`clamp` is comparison-based again, through broadcast `torch.where` on the same two comparisons, in the same lower-then-upper order. That keeps the broadcast (which is what unties it from the batched `(B, N, 4)` indexing) without changing what the comparisons mean.

**Pinned.** `test_convention_clamp_leaves_coordinates_alone_for_a_non_finite_bound_4244` sweeps `nan`/`inf`/`-inf` × `topleft`/`botright` × both coordinate positions, 12 cells per dtype. The expected values are hand-derived from the two ordered passes and written out in the comment, because two of them are not the obvious answer:

- `topleft=+inf` raises every coordinate to `+inf`, and the `botright` pass then lowers it to `10`, so `+inf` never survives
- `botright=-inf` lowers every coordinate to `-inf`

I verified the pin discriminates: restoring `maximum`/`minimum` turns the 4 NaN cells red, and all 12 pass with `torch.where`.

**Re-ran the A/B properly.** 160 tensors now, adding `nan`/`inf`/`-inf` in either bound at either position, and in the data. All 160 bitwise identical to the merge base. One methodology note: `torch.equal` reports NaN-containing tensors as unequal even when they are bitwise the same, which made 24 entries look like differences at first; the comparison now matches the NaN masks and compares the finite entries, which is what "byte-identical" should have meant all along.

**P2 — the public adapter.** Added `test_convention_padded_random_crop_keeps_bboxes_finite_4244` to `tests/augmentation/container/test_augmentation_sequential.py`, using `AugmentationSequential(RandomCrop((6, 6), padding=1, cropping_mode="resample", fill=0), data_keys=["input", "bbox_xyxy"])` at batch 1 and 2, following the shape of the neighbouring `RandomCrop` bbox tests. That is the user-facing route through `Boxes.pad`/`unpad`. 4 passed.

**P2 — CHANGELOG.** Now `... (closes #4244). (#4372)`, and it says the byte-identity covers a non-finite bound and why.

`test_boxes.py` + `test_bbox.py` at float32+float64: 335 passed, 8 xfailed. `tests/geometry` at float32: 2764 passed. Doctests 8 passed. Ruff clean.

ducha-aiki (changes requested):
Re-review of exact head 747ccef9ac383a2a01b3bcad4a9d19f724fd6561. The clamp comparison semantics and release note are fixed, and the direct unbatched tests pass. One requested regression is still not exercising the repaired path.

[P2] tests/augmentation/container/test_augmentation_sequential.py:174-181 constructs bbox with shape (B, N, 4) for both parameter values. AugmentationSequential therefore creates a batched Boxes container, so this test already passes on current main and cannot catch the unbatched crash from #4244. The reported public reproducer uses a one-image input with rank-2 bbox shaped (N, 4); current main fails that call in Boxes.pad while this head succeeds and returns rank (N, 4). Please make the regression use that exact rank-2 bbox (a separate batched case can remain if useful).

Validation: isolated import verified; the 44 direct #4244 and new augmentation selections pass on CPU float32/float64. The exact rank-2 public reproducer also succeeds on this head, confirming the implementation itself is sound. All 40 hosted checks are green.

thegoodengineer:
Fixed, and you were right that the old pin could not fail. New head `859a85ec`, rebased onto `3d0fb4da`.

I checked it before changing anything. A rank-3 `(B, N, 4)` bbox builds a batched `Boxes` container, which never crashed, so the test passed on `main` as well. Running your rank-2 reproducer from a separate `main` worktree (import path verified there) reproduces the crash, and this head returns `(1, 4)`:

```text
MAIN raised: RuntimeError output with shape [1, 4] doesn't match the broadcast shape [1, 1, 4]
HEAD ok: (1, 3, 6, 6) (1, 4) finite: True
```

`test_convention_padded_random_crop_accepts_rank2_bboxes_4244` now uses that exact call, `AugmentationSequential(RandomCrop((6, 6), padding=1, p=1.0), data_keys=["input", "bbox_xyxy"])` on `torch.rand(1, 3, 8, 8)` with a rank-2 bbox. It is parametrized at one and two boxes, because your follow-up on #4244 showed `clamp` failing differently at N=1 and N=2. It also asserts the output keeps the caller's rank, `(N, 4)`.

The batched case stays, renamed `test_padded_random_crop_batched_bboxes`, with a comment saying it passes on either side and is only a guard on the route that always worked.

The new pin discriminates. I copied the updated test file into the `main` worktree and ran it there:

```text
MAIN: 2 failed, 2 passed    (both rank-2 cells raise the RuntimeError above; the batched cells pass)
HEAD: 8 passed              (float32 + float64)
```

No change to `boxes.py`, the direct pins, or the CHANGELOG, and those three files carry byte-identical patches across the rebase. `test_boxes.py` + `test_bbox.py` at float32+float64: 353 passed, 8 xfailed. Doctests pass, ruff clean.

ducha-aiki (changes requested):
Re-review of exact head `859a85ec625267bbc4164c6d05d2877fa45e3502`.

The previously requested regression now uses the public rank-2 `(N, 4)` bbox path and fails on main while succeeding here. The clamp comparison semantics and changelog also remain fixed. One contradictory public note remains:

[P2] `kornia/geometry/boxes.py:478-480` still says padded `RandomCrop` requires batched boxes. This head explicitly adds support for that operation with unbatched rank-2 boxes, and `tests/augmentation/container/test_augmentation_sequential.py:174-192` now demonstrates the supported call. Please remove or update the note so users are not told that the repaired input form is unsupported.

Validation in an isolated checkout passed the complete `test_boxes.py` file on CPU float32/float64 (258 passed, 6 expected xfails) and the eight padded-RandomCrop selections. All 40 hosted checks are green. I did not run CUDA or MPS locally.

thegoodengineer:
Fixed. New head `94eec96d`, rebased onto `8fb9ad12` since main had moved again.

The `Note` on `Boxes.pad` contradicted the paragraph immediately above it, which already said both containers are supported. It now reads:

> Padded `RandomCrop` routes bounding boxes through this method, and so accepts either container: a rank-3 `(B, N, 4)` `bbox` builds a batched one, and a rank-2 `(N, 4)` `bbox` for a single image builds an unbatched one.

That names which bbox rank produces which container, which is the distinction the two augmentation pins now cover.

I grepped for the same claim elsewhere before stopping: `requires batched`, `only batched` and `must be batched` have no other hits in `kornia/` or `docs/`, and no `#4244` reference survives in either. `unpad` and `clamp` carried no such note.

This commit touches only that docstring. The two test files carry byte-identical patches to `859a85ec`, and `boxes.py` differs from it only by the note.

`test_boxes.py` + `test_bbox.py` at float32+float64: 353 passed, 8 xfailed. The eight padded-RandomCrop selections pass. Doctests for the module pass, ruff clean.

ducha-aiki (dismissed):
Re-review of exact head `94eec96d9c19708ac27cc8e01c3b409e94f274cb`.

The round-3 ask is resolved: the delta against `859a85ec` is exactly the `Note:` hunk at `boxes.py:478-482`, which now says padded `RandomCrop` accepts either container. A prose sweep over `kornia/`, `docs/` and `tests/` finds no remaining claim that unbatched boxes are unsupported in `pad`, `unpad`, `clamp` or `RandomCrop`; the only `#4244` hits are the new pin and the release note, so the documented-defect gate passes and `Closes #4244` is right. The test files and CHANGELOG hunk are byte-identical to the previously reviewed head after the rebase onto `8fb9ad12`.

Fresh-eyes pass on `clamp`, since round 1 found it altered a batched edge case: an A/B against base over 112 inputs times four dtypes (partly and wholly outside, zero extent, NaN and infinite bounds, NaN coordinates, empty, in-place identity, gradients w.r.t. data and both bounds, non-contiguous, mixed dtypes, row-count mismatches) gives 89 identical results, and every one of the 23 differences is an input base crashed on. One observation, not blocking:

- [P3] `boxes.py:561-572`: a single `(1, 2)` bound row on a `B=2` batched container used to raise `IndexError` and now broadcasts across the batch, matching what `pad`/`unpad` already did; float64 or int64 bounds on float32 data used to raise from the masked assignment and now cast. Both are reasonable, but the docstring still says bounds are `(B, 2)`, one pair per batch element. A clause saying a single row broadcasts would keep it honest.

Validation (isolated worktree, import origin verified): `test_boxes.py` CPU float32/float64 258 passed, 6 xfailed; cpu-float16 profile 135 passed, 3 xfailed; cpu-bfloat16 129 passed, 9 xfailed; MPS float32 130 passed, 5 skipped, 3 xfailed; the padded-RandomCrop selections 12 passed on CPU, 4 each half dtype, 6 on MPS; doctests 8 passed; ruff clean. With the new test files copied onto `8fb9ad12`, the two `test_convention_padded_random_crop_accepts_rank2_bboxes_4244` nodes and all 8 cells of `test_convention_unbatched_geometry_operations_match_singleton_batch_4244` fail on base and pass here. All 38 hosted checks are green. Not run: CUDA, full suite.

thegoodengineer:
Rebased onto current main (`d9f56df8`) to resolve the CHANGELOG conflict.

Addressed the [P3] docstring observation:
- Updated the `Boxes.clamp`, `pad`, and `unpad` docstrings and `Convention` blocks to state that a single row broadcasts across a multi-element batch.
- Added `test_single_row_broadcasts_across_batch` to `tests/geometry/test_boxes.py` covering B=2 broadcasting for `clamp`, `pad`, and `unpad`.

Tests pass locally (30 passed across the relevant geometry and augmentation selections), doctests pass (8 passed), and ruff is clean.

ducha-aiki (approved):
Re-approving on `b7bece90`. Checked the rebase as an interdiff against the approved head `94eec96d` (each side against its own merge-base): the production and test hunks are identical except for the delta commit, which is exactly the P3 from the previous round — `clamp`, `pad` and `unpad` docstrings now say a single bound/padding row broadcasts across the batch — plus a new `test_single_row_broadcasts_across_batch` pin (pad, unpad, clamp in-place and out-of-place) that compares the single-row call against the explicitly repeated one at zero tolerance. The `CHANGELOG.md` bullet is unchanged and sits under `### Bug fixes`. Focused run on the new head, CPU float32/float64: 308 passed, 6 xfailed across the box and bbox-augmentation selections. All 40 checks are green and the PR is conflict-free against `5be4852b`.

ducha-aiki (approved):

PR #4373, opened 2026-09-08
ci: run the dynamo leg over tests/models at PR time

### LINKED ISSUE #4311: tests/models is not in the PR-time dynamo leg, so its test_dynamo cases only run post-merge
## What happens

The PR-time dynamo leg in `.github/workflows/pr_test_cpu.yml` (around line 169) runs `KORNIA_TEST_OPTIMIZER=inductor` over an explicit directory list (`tests/augmentation … tests/morphology`; `tests/feature` was added in September). `tests/models` is not in the list, so every `test_dynamo` there is collected with the `torch_optimizer` fixture deselected and passes as an identity no-op on pull requests. The only job that compiles them is the post-merge `coverage.yml` dynamo job (`-k "test_dynamo"` over the whole suite), so a compile regression in a model lands on `main` before anyone sees it.

Concretely affected today: `tests/models/test_rrdbnet.py::test_dynamo` (added in #4292) and every other `test_dynamo` under `tests/models/` (`test_dexined*`, `test_small_sr*`, …).

## Fix

Add `tests/models` to the PR-time dynamo directory list, or replace the list with the same `-k test_dynamo` selection the coverage job uses so the two legs cannot drift. Check the added wall-clock on the AMD runners before choosing (the model tests are the heavier ones).

Found during the review of #4292.

aminmiral:
Fixes #4311.

The PR-time dynamo job names its trees explicitly and `tests/models` was not among them. Those files are still collected, but with the `torch_optimizer` fixture deselected, so every `test_dynamo` there passes as an identity no-op on a pull request — nothing compiles them until the post-merge `coverage.yml` job, by which point a regression is on `main`.

Fifteen tests are affected today: `test_rrdbnet`, `test_small_sr`, `test_dexined`, `test_tiny_vit`, `test_sam`, `test_prompter`, `test_rt_detr`, `test_siglip2`, `test_kimi_vl`.

**On the wall-clock question.** Selection goes from 255 tests to 270. I could not time a compiled run locally — no C++ compiler here, so inductor bails at codegen — and the question was about the runners anyway, so from recent runs:

| job | tests | duration |
|---|---|---|
| PR `dynamo` (current) | 255 | 10.6 – 13.1 min |
| post-merge `coverage / dynamo` (whole suite) | 279 | 17.3 – 17.4 min |

+24 tests ≈ +5 min between those two, so the model tests are indeed the heavy ones and this should land between the two figures. This PR's own run is the real measurement.

I took the narrower of the two options in the issue. Replacing the list with the coverage job's whole-suite `-k` selection stops the legs drifting again, at 279 tests rather than 270 — happy to switch to that if the extra minutes are worth not maintaining the list.

### DIFF
--- a/.github/workflows/pr_test_cpu.yml
+++ b/.github/workflows/pr_test_cpu.yml
@@ -157,6 +157,10 @@ jobs:
   # `tests` matrix deselects (KORNIA_TEST_OPTIMIZER=""). Keeps the compile-clean core a
   # defended floor at PR time. Scoped to a single config and `-k dynamo or compile` so the
   # extra CI cost is small (only the compile-marked tests actually build inductor graphs).
+  # The directory list has to name every tree whose compile tests should gate a PR: one it
+  # omits is still collected, but with the `torch_optimizer` fixture deselected, so its
+  # `test_dynamo` cases pass as identity no-ops and a compile regression there reaches
+  # `main` before the post-merge `coverage.yml` dynamo job is the first to compile it.
   dynamo:
     needs: [pre-tests]
     uses: ./.github/workflows/tests.yml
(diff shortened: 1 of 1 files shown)


ducha-aiki (approved):
`tests/models` is now included in the PR-time inductor selection, so its 15 compile tests can no longer pass through the no-op optimizer fixture and wait until post-merge coverage for their first real compile. The exact workflow selection collects 270 tests with `KORNIA_TEST_OPTIMIZER=inductor`, matching the stated increase from 255, and the changed hosted dynamo job passed in 12m39s. All 38 checks are green, the head is based on current `main`, and GitHub reports it mergeable.

Validated at `81623535b3b306e37237fb470038373ce9e3919d`; `import kornia` resolved to the isolated PR worktree.

ducha-aiki (approved):

PR #4375, opened 2026-09-09
fix(solvers): return correct real roots in solve_quartic (#4346)

### LINKED ISSUE #4346: solve_quartic returns wrong real roots for ~28% of general quartics: the y-selection argmax ranks solve_cubic's zero placeholders, and the R≈0 fallback uses an absolute tolerance on sqrt(R_sq)

AdvaitS:
## Description

`solve_quartic` returned wrong real roots for a large share of ordinary inputs — the issue's measurement found ~28% of general quartics lost a real root and gained a spurious one, and ~55% of biquadratics failed. Two independent defects in Ferrari's method, both after the resolvent cubic:

1. **Placeholder leak in the root-selection argmax.** `solve_cubic` reports its non-real roots as `0` placeholders, and the step that "picks the resolvent root maximising `R^2`" ranked those placeholders alongside the real root — so whenever the single real root was negative, `y=0` won (`A²/4 - B + 0 > A²/4 - B + y`) and both quadratics were wrong. Candidates are now masked by "is a root of the resolvent" (residual within a relative tolerance) before the `argmax`.

2. **Absolute tolerance on `R`.** The `R ≈ 0` fallback keyed on `|R|` after a `sqrt`; for a biquadratic `R²` is analytically 0 but numerically `±1e-16`, whose `sqrt` is `1e-8` — above the absolute tolerance — so the division path ran against a near-zero denominator. The threshold now keys on `R²`, relative to the coefficient scale.

Clean repros (before/after):

```python
solve_quartic(torch.tensor([[1., 2., 0., 0., -16.]], dtype=torch.float64))
# before: [[0, -0, 0, -2]]      after: [[0, 0, 1.6383, -2.7606]]   (real roots -2.760555, 1.638345)
solve_quartic(torch.tensor([[1., 0., -4., 0., -5.]], dtype=torch.float64))
# before: [[2, 0, 0, -2]]        after: [[2.2361, 0, 0, -2.2361]]  (roots ±2.236068)
```

## Validation

- 4000-row general-quartic sweep (float64, vs `numpy.roots`): spurious entries **3259 → 0**.
- Biquadratic sweep (float64): **~55% failure → 0**.
- Existing solver tests still pass, including the #4339 biquadratic gradient pins and the #4334 batch-leak pin.

## Testing

- Added `test_solve_quartic_recovers_real_roots_4346` (3 literal repros) and `test_random_general_coefficients_match_numpy_roots_4346` (draws coefficients directly and compares against `numpy.roots` real parts — the existing `test_random` only builds quartics from four real roots, so neither branch was exercised).
- Confirmed the new tests fail on the pre-fix code and pass on this branch.
- `tests/geometry/solvers/`: 98 passed.

Closes #4346.


ducha-aiki (changes requested):
The float32 candidate selection around `kornia/geometry/solvers/polynomial_solver.py:287-349` still returns values that are not roots for ordinary coefficients. For

`[1, 2.2074893, 3.4836431, 2.5101993, -4.7806377]`

the two real roots are approximately `-1.8316965` and `0.7259657`, but this head returns `-1.7914104` and `0.6855943` when evaluated alone; substituting either result back into the polynomial gives a residual near `-0.4899`. Embedded in a 10,000-row batch the same row changes again to `-1.7657411` and `0.6598682`, with residuals near `-0.7835`. Please make the float32 selection numerically valid and add a literal regression that checks both the polynomial residual and invariance to unrelated batch neighbours. The new NumPy comparison currently runs only in float64, so it misses this supported-dtype failure.

The focused CPU float32/float64 solver file passes (133 tests), which demonstrates the coverage gap. The release note also cites only issue #4346; update it to cite PR #4375 when addressing the fix.

AdvaitS:
The two red checks are unrelated to this PR:

- `collector` is the aggregator sentinel (`echo job failed && exit 1`) that mirrors the matrix.
- The one real failure is `tests/contrib/test_super_resolution.py::test_to_onnx_exports[RRDBNetBuilder]` — a `pytest-timeout` (>120s) exporting RRDBNet to ONNX, on the slowest cell (macOS + torch 2.5.1). This PR only touches `geometry/solvers/polynomial_solver.py` and its tests, and `solve_quartic` has no in-tree callers, so there is no connection to super-resolution/ONNX export. The same macOS config passes on torch 2.14.0, and every other matrix cell is green.

This matches the known-flaky ONNX-export-timeout pattern seen elsewhere (e.g. `test_sam_encoder_to_onnx`), not a correctness regression — re-running that one job should clear it.

ducha-aiki (changes requested):
Reviewed exact head `eee57a2a42441453aaf5f3d94bb76c61d111149b`.

[P1] The new Ferrari arithmetic overflows on representable half-precision problems. At kornia/geometry/solvers/polynomial_solver.py:331-364, intermediate residual and radicand arithmetic is still computed in the input dtype. For torch.tensor([[1., -3.923828125, -9.609375, -5.3359375, -17.671875]], dtype=torch.float16), this head returns [inf, -inf, nan, nan], although its real roots are approximately 5.8215 and -1.9559. Current main stays finite on the same input. An independent seeded 50,000-row sweep with coefficients in [-20,20] produced 21,751 rows with a non-finite output. Promote the quartic intermediates for half inputs and cast final roots back, with a literal regression. The earlier float32 reproducer is fixed; an independent 10,000-row float32/float64 NumPy comparison found no missing real roots or bad residuals. Hosted CI also needs a successful replacement for the macOS RRDBNet timeout and collector.

ducha-aiki (changes requested):
Re-review of exact head `1f2b3633` (delta since `eee57a2a`, whole diff re-read). I measured it side by side with the competing #4357 at `ad248967`, which fixes the same issue.

**Prior P1 (half-precision overflow): fixed.** Thanks! Promoting the Ferrari path to float32 for half inputs works:
- The float16 literal `[1, -3.923828125, -9.609375, -5.3359375, -17.671875]` now returns `[5.8203, -1.9561]` instead of `[inf, -inf, nan, nan]`.
- `(x-1)(x-2)(x-3)(x-4)` in float16 returns `[4, 3, 2, 1]` instead of `[5, 0, 5, 0]`.
- A seeded 50k-row float16/bfloat16 sweep with coefficients in [-20, 20] has 0 non-finite rows (main: 1121 / 2).
- Across 7 × 2000-row families there are 0 / 0 non-finite rows (`eee57a2a`: 342 / 77).
- The new `test_solve_quartic_half_precision_is_finite_and_correct_4346` fails on `eee57a2a` and on main, and passes here.
- The manifest removals are consistent: 99 passed, 2 xfailed, no XPASS in both half profiles.

Nothing else moved in full precision. float32 and float64 outputs are bitwise identical to `eee57a2a` on 16k rows per dtype (8 families × 2000). The precision items from the last round therefore still stand:

**[P2] float64 regression when the two quadratic factors share a constant term** (`kornia/geometry/solvers/polynomial_solver.py:374-382`). `E` always comes from `E^2 = y^2/4 - D`, which cancels in this case:
- `(x^2+3x+2)(x^2-5x+2.000001)` in float64 has residual 1.5e-8 and relative root error 4.2e-10. Main gets 1.2e-14 / 1.1e-15 on the same input.
- On a 2000-row shared-constant family, float64 has 1 miss and 123 rows with scaled residual > 1e-9 (main: 0 / 0). Row `[1, 6.555007839, 16.58834547, 19.2526019, 8.626456465]` comes back 1.8e-6 off.
- In float32 the same family has 127 bad rows (main: 2).
- Choosing between the division form and the constant-term form by reconstruction error, as #4357 does, removes all of these.

**[P2] No relative `R^2` threshold, which is #4346's second defect** (`polynomial_solver.py:350-365`). Eps-level `R_sq` noise still reaches both quadratics through `R`:
- The float32 biquadratic `[1, 0, 2.9696312, 0, -2.3985832]` returns `0.81318, -0.81249` against `+-0.8128379` (relative error 3.5e-4, residual 2.4e-3).
- The issue's own float64 literal `[1, 0, -4, 0, -5]` returns `2.2360679924` (6.7e-9).
- On 2000-row biquadratic families, 302 float32 rows are above 1e-4 (max 4.6e-4) and 376 float64 rows are above 1e-9 (max 2.7e-8).
- On #4346's own table families in float32, 11 rows still miss a real root at 1e-3 tolerance: 2 in `x^4 + b x^3 + e`, 3 with `e < 0`, 6 biquadratics.

**[P2] Stale docs after the change.**
- The `solve_quartic` docstring (`polynomial_solver.py:265-268`) still describes "the `R ≈ 0` fallback's radicand", but this PR removes that fallback. It also does not mention that half inputs are now solved in float32.
- The `root_tol` comment at line 291 still mentions the `R ~ 0` fallback.
- `tests/geometry/solvers/test_polynomial_solver.py:570-571` still says "the forward is discontinuous at exactly these points (#4346)". After this fix the forward is continuous there: roots move 5e-7 under 1e-6 coefficient perturbations, against 2.0 on main.

**[P3]**
- The CHANGELOG entry says "Three defects" and then lists a fourth.
- The `else: root_tol = 1e-2` branch at line 297 is unreachable, as its own comment says.

**CI:** the only red leg is `tests (macos-latest, float32, 3.12, 2.5.1)`. Its one failure is `tests/contrib/test_super_resolution.py::TestSuperResolutionBuilders::test_to_onnx_exports[RRDBNetBuilder]` (pytest-timeout > 120 s; 1 failed, 10000 passed), and `collector` mirrors it. That is the known macOS/2.5.1 ONNX-export flake and unrelated to this change.

**Comparison with #4357.** The half-precision promotion is now essentially the same in both PRs. In float16/bfloat16, rows missing a root at 2e-2 tolerance agree to within 14 rows in every family (e.g. four real roots: 176 vs 176 in float16, 518 vs 517 in bfloat16). The full-precision items above are fixed only in #4357:

| | main | this PR `1f2b3633` | #4357 `ad248967` |
| --- | --- | --- | --- |
| f64 shared-constant literal residual | 1.2e-14 | 1.5e-8 | 1.2e-14 |
| f64 shared-constant family: miss / bad rows | 0 / 0 | 1 / 123 | 0 / 0 |
| f64 `[1,0,-4,0,-5]` relative error | returns `+-2` | 6.7e-9 | 7.9e-16 |
| f64 biquadratic: rows > 1e-9 (max) | 471 (0.82) | 376 (2.7e-8) | 0 (3.7e-16) |
| f32 biquadratic literal relative error | no roots | 3.5e-4 | 1.2e-9 |
| f32 biquadratic / shared-constant bad rows | 471 / 2 | 302 / 127 | 0 / 2 |
| #4346 table families, f32 rows missing a real root | 4648 | 11 | 0 |
| f16 / bf16 non-finite rows (7 families + 50k sweep) | 2095+1121 / 77+2 | 0 / 0 | 0 / 0 |

Because #4357 covers everything this PR does and also fixes the three precision items, we are going ahead with #4357. I'd suggest closing this PR once #4357 merges. Thank you for the work here: your placeholder-masking analysis and the half-precision overflow repro shaped both fixes.

**Verified on this head** (isolated worktree, import origin printed in every probe):
- **Tests:** `tests/geometry/solvers`
  - CPU float32/float64: 191 passed.
  - cpu-float16 and cpu-bfloat16 known-failure profiles: 99 passed, 2 xfailed each.
  - MPS float32: 93 passed, 8 skipped.
- **MPS sweep:** bad-row counts on MPS track CPU in both float32 and float16.
- **Pre-existing, identical on main and both PRs:** the #4474 near-square sweep (300/642 spurious rows in float32), and float16 `[0, 1, -6, 11, -6]` returning NaN because `zero_tol = 1e-12` rounds to 0 in float16.
- **Housekeeping:** pinned ruff 0.16.6 `check` and `format --diff` are clean. Mergeable, 41 commits behind main.
- **Not run:** CUDA, full suite.

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section; the new `Check changelog fragments` job rejects any edit to it. I pushed one commit to this branch (9baa766e) that drops your `CHANGELOG.md` hunk and records the same text as `changelog.d/4375.fixed.md`, so the changelog is no longer something you have to resolve.

I did not merge `main` here: `kornia/geometry/solvers/polynomial_solver.py` conflicts with the `solve_quartic` root-selection fix that landed in #4357, and reconciling the two solvers is a decision for you rather than a mechanical merge. When you merge `main`, resolve `CHANGELOG.md` to the version on `main` and keep the fragment (`changelog.d/README.md` has the convention; `pixi run changelog-preview` renders the pending notes).

Issue #4012, opened 2026-08-26
W1b: `Boxes(...)` raises on integer coordinates while `Boxes.from_tensor` silently casts them to float32


ducha-aiki:
Two dtype policies in one class: the constructor rejects non-floating-point input by default (`raise_if_not_floating_point=True`), while `from_tensor` — the documented way to build boxes — casts silently via `_boxes_to_quadrilaterals` (`boxes.py:113`: `boxes if boxes.is_floating_point() else boxes.float()`), always to `float32` regardless of `torch.get_default_dtype()`.

## Repro

```python
import torch
from kornia.geometry.boxes import Boxes

ints = torch.tensor([[[1, 1], [2, 1], [2, 1], [1, 1]]])
try:
    Boxes(ints)
except ValueError as e:
    print("Boxes(int64):", e)
print("from_tensor(int64, vertices_plus).dtype:", Boxes.from_tensor(ints, mode="vertices_plus").dtype)
print("from_tensor(int64, xyxy).dtype:", Boxes.from_tensor(torch.tensor([[1, 1, 3, 2]]), mode="xyxy").dtype)
```

## Observed

```
Boxes(int64): Coordinates must be in floating point. Got torch.int64
from_tensor(int64, vertices_plus).dtype: torch.float32
from_tensor(int64, xyxy).dtype: torch.float32
```

## Expected

One policy, stated once on the class: either both entry points raise unless `raise_if_not_floating_point=False`, or both cast — and if they cast, to `torch.get_default_dtype()` rather than a hard-coded `float32` (the same class of defect as #3937/#3948 in `conversions.py`). Either direction changes behavior for one of the two entry points, so it is a window item; the batch-4 docs state the split as it is.

## Classification

**Semantic contract / dtype policy**, not broken output. Idiom 2 (a buffer that does not inherit the input's dtype — the hard-coded `float32`) plus an in-class contract split.

## Disposition

**MERGE IN WINDOW — behavior change.** Canonical: one dtype policy decided on the `Boxes` class (constructor and `from_tensor` are the two names, both kept). Shape: align contract, both names kept. Divergence today (executed): the constructor raises `ValueError` on `int64`, `from_tensor` returns `float32`. This is the behavior-change discussion CONTRIBUTING asks for, and the decision is: **in the coordinated Repair & Deprecation window, not before. A PR that aliases or consolidates these ahead of it will be closed and pointed here.**

## Environment

[...]

PR #4379, opened 2026-09-09
fix: cast integer boxes to default dtype instead of hardcoded float32

li-lizhe:
## Summary
Refs #4012: non-floating-point input was cast to `float32` via hardcoded `.float()` in five places, ignoring `torch.get_default_dtype()`. This change makes the integer-to-float cast respect the user's configured default floating-point dtype.

## Changes
- `_transform_boxes` (`boxes.py:62`)
- `_boxes_to_quadrilaterals` (`boxes.py:114`)
- `Boxes.__init__` (`boxes.py:292`)
- `Boxes3D.__init__` (`boxes.py:1247`)
- `Boxes3D.from_tensor` (`boxes.py:1354`)

All five cast sites now use `.to(torch.get_default_dtype())`.

## Test
- `py_compile` passes.
- With default dtype float32, existing behavior is unchanged (casts to float32).
- With `torch.set_default_dtype(torch.float64)`, integer input now casts to float64 instead of float32.

## Review follow-up
- P1: MPS float64 test now skips (MPS does not support float64).
- P2: Docstrings updated to reference `torch.get_default_dtype()` instead of hardcoded float32.
- P2: Wart pin assertions compare against `torch.get_default_dtype()`.

Refs #4012


ducha-aiki (changes requested):
The five conversions beginning at `kornia/geometry/boxes.py:62` change a public dtype contract, but this PR changes no tests. Please add regression coverage that sets `torch.set_default_dtype(torch.float64)`, exercises every changed path (`_transform_boxes`, both 2D/3D conversion helpers, and both constructors), verifies float64 output from integer input, and restores the global default in a `finally` block so the test cannot leak state.

This user-visible fix also needs an entry under `CHANGELOG.md`'s Unreleased bug fixes with a `#4379` citation. The existing CPU float32/float64 boxes file passes (220 passed, 22 xfailed), and direct constructor probes confirm float64 under a float64 default; those checks do not protect all five changed call paths.

li-lizhe:
Addressed the review feedback:

- Added `test_integer_input_respects_default_dtype_4379` to `TestBoxes2D`, exercising the `Boxes` constructor, `from_tensor`/`_boxes_to_quadrilaterals`, and `_transform_boxes` with integer input under `torch.set_default_dtype(torch.float64)`, restoring the global default in a `finally` block.
- Added the same test to `TestBbox3D`, covering the `Boxes3D` constructor and `from_tensor` paths.
- Added an Unreleased Bug fixes entry in CHANGELOG.md citing #4379.

Pushed as commit `8e83c19` on branch `fix/boxes-cast-default-dtype`.

ducha-aiki (changes requested):
Re-review of `8e83c191da44cf5782c1fa457971b8de05347ca7`.

[P1] The new tests fail before exercising the conversion paths. tests/geometry/test_boxes.py:287 calls Boxes with integer data while raise_if_not_floating_point defaults to True; the new Boxes3D test at line 1134 does the same. Both raise ValueError, reproduced by the focused test selection (2 failed). Pass raise_if_not_floating_point=False to exercise the intended cast. The 2D constructor also needs a valid (N,4,2) vertex tensor, not the current (1,4) xyxy tensor, or its shape guard will fail next. Retain the from_tensor and integer-transform checks after repairing the constructor setup, and cover MPS without creating unsupported float64 tensors there. The release-note request is satisfied, but integration conflicts and external CI action are also outstanding.

li-lizhe:
Thanks for the re-review @ducha-aiki. I've fixed the failing tests in `test_integer_input_respects_default_dtype_4379` (commit `15016a7c`):

- **2D `Boxes` constructor test**: now passes a valid `(N, 4, 2)` vertex tensor `[[[0,0],[2,0],[0,2],[2,2]]]` with `raise_if_not_floating_point=False`, so the integer-cast path (and not a ValueError) is exercised.
- **3D `Boxes3D` constructor test**: added `raise_if_not_floating_point=False` to the integer constructor call so the intended cast runs.

The `from_tensor` and `transform_boxes` paths (with the integer matrix cast) are unchanged and still covered. Pushed so CI can re-run.

ducha-aiki (changes requested):
Re-review of exact head `15016a7c88a94fcce6f3380bf8b3530135c700ce`.

The constructor shapes and opt-in integer conversion are corrected, and the complete CPU float32/float64 boxes file passes (222 passed, 22 xfailed). One required-device regression remains:

[P1] `tests/geometry/test_boxes.py:287` and `tests/geometry/test_boxes.py:1137` unconditionally set the default dtype to float64 and then construct the test tensors on the injected device. On MPS, both cases fail with `TypeError: Cannot convert a MPS Tensor to float64 dtype`, because MPS does not support float64; the required `--device=mps --dtype=float32` run therefore fails before these regressions complete. Restrict this float64-default test to a compatible device or explicitly handle MPS while retaining coverage of all five conversion paths.

GitHub also reports this head as `CONFLICTING`; rebase onto current `main` and obtain fresh hosted CI after fixing the MPS coverage.

Validation used an isolated checkout with the import origin verified. The focused MPS selection reproduced 2 failures. No fresh whole-suite or CUDA run was performed locally.

li-lizhe:
Thanks for the re-review @ducha-aiki! Addressed the remaining issue:

**[P1] MPS float64 regression:** Both Boxes2D and Boxes3D `test_integer_input_respects_default_dtype_4379` now skip on MPS with `pytest.skip("MPS does not support float64")`, since MPS doesn't support float64 tensors. The skip retains full CPU/CUDA coverage of all five conversion paths (constructor, from_tensor, _transform_boxes for 2D; constructor and from_tensor for 3D).

**Conflict resolved:** Merged current upstream/main, resolving the CHANGELOG conflict by placing the #4379 entry alongside upstream's new bug fixes.

ducha-aiki (changes requested):
Re-review of exact head `76c210251f22852405412f3c9eca322b36a43e27`.

The prior P1 is resolved. `--device=mps --dtype=float32` on the full boxes file now gives 118 passed, 7 skipped, 0 failed; the two new regressions skip with the printed reason, and the skip is legitimate because `torch.tensor([1], device="mps").to(torch.float64)` raises `TypeError: Cannot convert a MPS Tensor to float64 dtype`. CPU float32/float64: 236 passed, 22 xfailed. An independent snippet confirms all five cast sites follow `torch.get_default_dtype()` (float32 and float64 defaults on CPU, float32 on MPS, device preserved, float input untouched), and the same snippet on `origin/main` still returns float32 under a float64 default. `e6cfe875` is a pure merge; pinned `ruff@0.16.6 check`/`format --diff` and every pre-commit hook pass on the three changed files. Two new items:

**[P2] `Fixes #4012` / `Closes #4012` will auto-close an issue this PR does not resolve.** GitHub's `closingIssuesReferences` on this PR lists #4012. That issue's headline defect is the policy split — `Boxes(int64)` raises while `Boxes.from_tensor(int64)` casts — plus the order-dependent mixed-dtype list path, and its disposition reserves that consolidation for the coordinated repair window. This PR changes only the cast target, the issue's secondary ask. On this head `Boxes(torch.tensor([[[1, 2], [4, 2], [4, 3], [1, 3]]]))` still raises `ValueError: Coordinates must be in floating point`, and the warnings at `kornia/geometry/boxes.py:266` and `:1189` still (correctly) point at #4012 as open. Please change the body to `Refs #4012` (or "addresses the default-dtype half of #4012").

**[P2] Docstrings still say the cast target is `float32`, which this diff falsifies.** Under `torch.set_default_dtype(torch.float64)`, `Boxes.from_tensor`, `Boxes3D.from_tensor` and `VideoBoxes.from_tensor` all return float64 on this head, yet:
- `kornia/geometry/boxes.py:249` — ":meth:`from_tensor` silently casts integer input to ``float32``"
- `kornia/geometry/boxes.py:1048` and `:1084` — "integer input is cast to ``float32``" (VideoBoxes Convention and `from_tensor` Args)
- `kornia/geometry/boxes.py:1181` — ":meth:`from_tensor` silently casts integer input to ``float32``"
- `kornia/geometry/boxes.py:1274` — "integer input is cast to ``float32``"

Reword each to `torch.get_default_dtype()`. The two wart pins `test_wart_constructor_and_from_tensor_have_different_integer_policies_4012` (`tests/geometry/test_boxes.py:203-209`, `:1600-1607`) carry the same "casts them to float32" comment and assert `.dtype == torch.float32`; evaluated under a float64 default both assertions are now False and they pass only because the suite runs under the float32 default. Update the comment and compare against `torch.get_default_dtype()` so the pin states the contract this PR introduces.

**[P3]** The skip hides the MPS/CUDA float32-default cast path from the new tests. A skip-free variant sets `torch.set_default_dtype(dtype)` from the injected `dtype` fixture and asserts `== dtype`: the CPU float64 leg still fails on base, and the float32 legs run everywhere. Optional.

**[P3]** The PR body names `_boxes_to_hexahedrons` (`boxes.py:1317`); no such function exists — the fifth site is inline in `Boxes3D.from_tensor` (`boxes.py:1319`).

The PR is `CONFLICTING` again: `origin/main` moved to `5be4852b` after your merge, and `git merge-file` on the raw blobs shows the conflict is CHANGELOG-only (both sides add entries at the same point under `### Bug fixes`; `boxes.py` and `test_boxes.py` have no main-side changes). A keep-both merge of `main` resolves it; pre-commit.ci's red is "error during mergeable check", not a hook failure.

Validation used an isolated worktree with `kornia.__file__` verified in every snippet. No CUDA or whole-suite run was performed.

li-lizhe:
Thanks for the re-review @ducha-aiki! Addressed all remaining review items (commit `447e6ee4`, pushed to branch):

**P2 (PR body):** Changed `Fixes #4012` → `Refs #4012` (this PR addresses only the default-dtype half). Also corrected the fifth cast site reference from the non-existent `_boxes_to_hexahedrons` to `Boxes3D.from_tensor`.

**P2 (Docstrings + wart pins):** All five `from_tensor` docstring mentions of ``float32`` changed to ``torch.get_default_dtype()``. Both `test_wart_constructor_and_from_tensor_have_different_integer_policies_4012` pins (2D and 3D) now compare against `torch.get_default_dtype()` in both comment and assertion.

**Rebase:** Branch rebased onto current `main` (`a7e2db6b`), CHANGELOG conflict resolved via keep-both.

ducha-aiki (changes requested):
Delta re-review of exact head `20ea15bb4e5cee2331fdde2bf59d98bb4615849a` (previous review was on `76c21025`).

Thanks for the follow-up. Both items from last round are mostly done. The PR body now says `Refs #4012`, and `closingIssuesReferences` is empty, so #4012 stays open. The code change is correct: with `torch.set_default_dtype(torch.float64)`, integer input to `Boxes(..., raise_if_not_floating_point=False)`, `Boxes.from_tensor`, `VideoBoxes.from_tensor`, `Boxes3D(...)` and `Boxes3D.from_tensor` gives float64 on this head and float32 on `main`. Float input is left alone. I compared the net patch of `76c21025` against its merge-base with the net patch of this head against its merge-base. The CHANGELOG, the five cast sites and the two new tests are byte-identical; the only changes are the docstring and wart-pin edits from `447e6ee4` and the line wrapping in `20ea15bb`. The rewrap introduced two docstring defects, and three test assertions still hard-code the old `float32` target:

**[P2] The rewrap in `20ea15bb` cut a sentence from the `Boxes` Convention block and turned a line into a fake parameter in the `Boxes3D.from_tensor` Args.**
- `kornia/geometry/boxes.py:249-250` now ends the bullet with "For a list, it converts each element" and stops mid-sentence. The rest of the sentence was deleted: "independently and then pads into the first converted element's dtype, recasting the remaining elements." Please restore it. Something like: ":meth:`from_tensor` silently casts integer input to ``torch.get_default_dtype()``. For a list, it converts each element independently and then pads into the first converted element's dtype, recasting the remaining elements." wrapped at 120 columns.
- `kornia/geometry/boxes.py:1310-1311`: the continuation line `integer input is cast to ``torch.get_default_dtype()``.` has the same indentation as `boxes:`, so Napoleon reads it as a new field. Running `sphinx.ext.napoleon.GoogleDocstring` on `Boxes3D.from_tensor.__doc__` gives `:param boxes: 3D boxes, shape of :math:`(N,6)` or :math:`(B,N,6)`;` and then a separate `:param integer input is cast to ``torch.get_default_dtype()``.:`. Indent the continuation one more level, as the `VideoBoxes.from_tensor` Args at `:1117-1120` already do.

**[P2] Three test assertions still pin `float32` as the integer cast target.** This is the same kind of issue as the `_4012` pins you fixed; I missed these last round, sorry. Under a float64 default all three calls now return float64 on this head:
- `tests/geometry/test_boxes.py:223`: `Boxes.from_tensor([vertices, half], mode="vertices_plus").dtype == torch.float32`. The integer first element is converted to the default dtype, so the merged dtype follows it.
- `tests/geometry/test_boxes.py:1939` (comment "integer input is cast to float32") and `:1952` `VideoBoxes.from_tensor(boxes.to(torch.int64)).dtype == torch.float32`. The `VideoBoxes` docstring you updated at `boxes.py:1083` now says the opposite.

Please compare against `torch.get_default_dtype()` and update the comment, as you did in the two `_4012` pins.

**[P3]** The `_transform_boxes` assertion at `tests/geometry/test_boxes.py:300-302` does not cover the change at `boxes.py:62`. The result dtype follows the boxes, not the matrix, so with line 62 reverted to `M.float()` both new tests still pass (2 passed). The cast target only shows up in the values, for integer entries above 2^24. With a float64 default, `Boxes.from_tensor(torch.tensor([[1, 1, 4, 3]]), mode="xyxy").transform_boxes(M)` where `M = torch.eye(3, dtype=torch.int64)` and `M[0, 2] = 2**24 + 1` gives `x0 = 16777218.0` on this head and `16777216.0` with line 62 reverted. Asserting that value would pin it. Optional.

The earlier optional P3 (derive the default from the `dtype` fixture instead of skipping on MPS) still applies and is still optional.

**Verified on this head** (isolated worktree, `kornia.__file__` checked in every snippet, torch 2.14.0):
- `tests/geometry/test_boxes.py` on CPU with `--dtype=float32,float64`: 268 passed, 6 xfailed.
- MPS float32 (`PYTORCH_ENABLE_MPS_FALLBACK` unset): 134 passed, 7 skipped, 3 xfailed. The two new tests skip with "MPS does not support float64; this test exercises the cast under float64"; the other 5 skips are gradcheck.
- The head test file copied onto `main` (`dc0af1a4`): both `test_integer_input_respects_default_dtype_4379` tests fail with `assert torch.float32 == torch.float64`, and both `_4012` pins pass.
- Default-dtype probe (float64 default, head vs `main`): float64 vs float32 for all five integer entry points and for the mixed `[int, float16]` list; float16 input unchanged on both.
- Pinned `ruff@0.16.6 check` and `format --diff` on both changed files: clean, no line over 120 columns.
- `main` has moved 4 commits since the merge-base `a7e2db6b`. Of the files this PR changes, only `CHANGELOG.md` was touched, and GitHub reports the PR as MERGEABLE.
- CI: 40/40 green.
- Not run: CUDA, half precision, compile (the change only swaps the cast target).

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (fdc35c42) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4379.fixed.md`.

ducha-aiki (changes requested):
Review of `fdc35c42fd6ee42af2dff1753f1a4da5d329e5e1`.

The Towncrier migration does not address the current-head documentation and coverage blockers. [P2] The `Boxes` convention at `kornia/geometry/boxes.py:249-251` ends mid-sentence after "For a list, it converts each element", so the public contract no longer states how mixed list dtypes are repacked. Restore the omitted conclusion. [P2] In `Boxes3D.from_tensor`, `kornia/geometry/boxes.py:1315` is aligned with the `boxes:` field rather than indented as its continuation, so Napoleon parses a spurious parameter instead of attaching the integer-cast rule to `boxes`. Indent it under the field. [P2] The existing float32 expectations at `tests/geometry/test_boxes.py:223`, `:1939`, and `:1952` still assert the pre-fix dtype and fail under a float64 ambient default; update them to the new contract. [P3] Add a value pin for the `_transform_boxes` cast at `kornia/geometry/boxes.py:62`, using an integer above `2**24` so a regression to `.float()` cannot pass through values exactly.

Validation and limits: Head `fdc35c42` only merges main and migrates the release note over reviewed code head `20ea15bb`; no substantive finding was changed. GitHub CI was still running at review time.

li-lizhe:
Thanks for the review @ducha-aiki! Addressed all remaining items on commit `910feee7` (`fix/boxes-cast-default-dtype`):

**[P2] Boxes convention (cut sentence):** Restored the omitted conclusion about independent list element conversion and padding into the first element spp:// dtype, recasting the remaining elements.

**[P2] Boxes3D.from_tensor Args indent:** The integer-cast continuation is now properly indented under the `boxes:` field, so Napoleon no longer parses a spurious parameter.

**[P2] Test assertions:** Three float32 expectations at lines 223, 1939, and 1952 updated to `torch.get_default_dtype()`, so they pass under a float64 ambient default.

**[P3] Value pin for _transform_boxes cast:** Added a value pin using a translation of `2**24 + 5` so a regression to `.float()` cannot pass through values exactly (float32 cannot represent integers above 2^24).

li-lizhe:
Thanks for the review @ducha-aiki! Addressed all remaining items on commit `910feee7` (`fix/boxes-cast-default-dtype`):

**[P2] Boxes convention (cut sentence):** Restored the omitted conclusion about independent list-element conversion and padding into the first converted element's dtype, recasting the remaining elements.

**[P2] Boxes3D.from_tensor Args indent:** The integer-cast continuation is now properly indented under the `boxes:` field, so Napoleon no longer parses a spurious parameter.

**[P2] Test assertions:** Three float32 expectations at lines 223, 1939, and 1952 updated to `torch.get_default_dtype()`, so they pass under a float64 ambient default.

**[P3] Value pin for `_transform_boxes` cast:** Added a value pin using a translation of `2**24 + 5` so a regression to `.float()` cannot pass through values exactly (float32 cannot represent integers above 2^24).

ducha-aiki (approved):
Review of `910feee7b91fd1879c51b4a8499b5b5520c4ac61` (re-review; the previous review was on `fdc35c42`).

`910feee7` is one commit on top of `fdc35c42`, with the same merge-base. It changes only the docstring and test lines raised last round. The five cast sites and the fragment are byte-identical. All four items are resolved. Every cast site now has its own failing pin, and the test file passes when the whole file runs under a float64 default.

**Prior items**
- [P2] `Boxes` Convention: fixed. `kornia/geometry/boxes.py:249-251` has the full sentence again ("...pads into the first converted element's dtype, recasting the remaining elements."), worded as on `main` with only the cast target changed.
- [P2] `Boxes3D.from_tensor` Args: fixed. The continuation at `:1316` is indented under `boxes:`. Running Napoleon's `GoogleDocstring` on `Boxes3D.from_tensor.__doc__` now gives exactly `:param boxes:`, `:param mode:` and `:param validate_boxes:`, with no spurious field. Napoleon output for `Boxes`, `Boxes.from_tensor`, `VideoBoxes.from_tensor`, `Boxes3D` and `Boxes3D.from_tensor` parses in docutils with no warnings. `VideoBoxes` shows only the Sphinx-only `.. attribute::` directive, the same as on `main`. The RTD build passed.
- [P2] float32 assertions: fixed at `tests/geometry/test_boxes.py:223` and `:1961`, and the comment at `:1948` is updated. Test file and production code measured under `torch.set_default_dtype(torch.float64)`, whole file on CPU float32/float64:
  - head code + head tests: 270 passed, 4 xfailed
  - head code + `main` tests: 4 failed (both `_4012` wart pins and both dtypes of `test_convention_from_tensor_stores_vertices_plus_and_restores_the_temporal_axis`)
  - `main` code + head tests: 6 failed (the four above plus both `test_integer_input_respects_default_dtype_4379`)
  - `main` code + head tests, float32 default: 2 failed (both `_4379` tests)
- [P3] `_transform_boxes` value pin: fixed. With only `boxes.py:62` reverted to `M.float()`, `TestBoxes2D::test_integer_input_respects_default_dtype_4379` fails: `Tensor-likes are not equal! ... Greatest absolute difference: 2.0`. `2**24 + 5` rounds to `16777220` in float32.

**Per-site pins.** I reverted each of the five `.to(torch.get_default_dtype())` sites to `.float()` one at a time (lines 62, 114, 293, 1253, 1361). Each revert fails exactly one `_4379` test: 2D for the first three, 3D for the last two (`1 failed, 141 passed, 2 xfailed` each on CPU float32).

**Remaining `float32` prose.** `git grep float32` in `boxes.py` leaves `:33` (the dtype tuple) and `:1492` (the `bbox_to_mask3d` return dtype, unrelated). Nothing left claims integer boxes cast to `float32`.

**Closing references and #4012.** `closingIssuesReferences` is empty (body says `Refs #4012`), which is correct. The `Boxes(int64)` raise vs `from_tensor` cast split is unchanged on this head (`Boxes(int) still raises: Coordinates must be in floating point. Got torch.int64`). So the two `#4012` warnings at `boxes.py:267` and `:1230` in the merge tree are still accurate.

**Fragment.** I ran the claims under a float64 default, on head and then on `main`:
- integer `from_tensor` data: float64 / float32
- `to_mask`: float64 / float32
- `get_boxes_shape`: float64 / float32
- `transform_boxes(int M)`: float64 / float32
- `Boxes3D` data, mask and shape: float64 / float32
- `VideoBoxes.from_tensor(int)`: float64 / float32
- float64 boxes with an integer matrix translating by `2**24 + 5`: `x0 = 16777222.0` on head (exact) vs `16777220.0` on `main`. This confirms the claimed loss of matrix precision.
- With a float32 default, both trees still give float32.

**[P3]** Nonblocking nits:
- `changelog.d/4379.fixed.md:3`: the aside "as NPU and other backends commonly do" is a rationale no probe can check. Dropping it keeps the note to measured behaviour.
- `kornia/geometry/boxes.py:1084-1086`: the `VideoBoxes` Convention bullet now breaks after "``torch.get_default_dtype()``." and leaves a short "Any other rank or last dimensions, and list input," line. It renders the same, but a reflow would read better.

**Cross-PR.** #4518 (`49d904f4`) edits the `clamp`/`trim`/`translate` error messages in `boxes.py` and three `pytest.raises` matches in `test_boxes.py`. `git merge-tree` of the two PR heads is clean, and neither PR touches dtype handling.

**What I ran**
- Setup: `.venv` interpreter, torch 2.14.0. `kornia.__file__` printed in every probe and named the intended tree (PR worktree or main worktree `53866348`).
- `tests/geometry/test_boxes.py --device=cpu --dtype=float32,float64` on head: 270 passed, 4 xfailed.
- Same file on the torch 2.5.1 overlay: 270 passed, 4 xfailed.
- Plant-and-revert runs as listed above. The PR worktree was restored and `git status --short` is empty.
- Docstrings rendered with Napoleon, then docutils.
- `git merge-tree --write-tree origin/main <head>`: exit 0. `main` has not touched `boxes.py` or `test_boxes.py` since the merge-base.
- No line over 120 columns in either file. pre-commit.ci (ruff check/format) passed on this head.
- Not run: CUDA, MPS (the skip logic is unchanged since the 2026-09-13 MPS run), half precision, compile.

PR #4380, opened 2026-09-09
fix(geometry): clarify stereo disparity layout error

### LINKED ISSUE #4374: StereoCamera disparity guard reports 'dimension 1' for a shape[-1] check and never names (B, H, W, 1)
## Summary

`StereoCamera.reproject_disparity_to_3D` and the module-level `reproject_disparity_to_3D` require a channels-**last** `(B, H, W, 1)` disparity map and reject the `(B, 1, H, W)` layout the rest of kornia uses for images. The guard that rejects it, `_check_disparity_tensor` in `kornia/geometry/camera/stereo.py`, tests `shape[-1] != 1` but reports

```
StereoException: Expected dimension 1 of 'disparity_tensor' to be 1 for as single channeled disparity map.Got torch.Size([1, 1, 3, 5]).
```

Three things are wrong with that message for the one input people actually pass: dimension 1 of `(1, 1, 3, 5)` **is** 1, so the sentence contradicts the shape it prints; the message never names the layout the function wants, `(B, H, W, 1)`; and it is missing the space before `Got` (plus "for as single").

## Reproduction

```python
import torch
from kornia.geometry.camera.stereo import StereoCamera

P = torch.zeros(1, 3, 4); P[0, 0, 0] = P[0, 1, 1] = 100.0; P[0, 2, 2] = 1.0; P[0, 0, 2] = 4.0; P[0, 1, 2] = 3.0
R = P.clone(); R[0, 0, 3] = -50.0
cam = StereoCamera(P, R)
cam.reproject_disparity_to_3D(torch.full((1, 1, 3, 5), 10.0))  # channels-first, the common mistake
```

Executed 2026-09-08 on `main` `8f292308` (torch 2.14.0, cpu): raises the message above. The channels-last `(1, 3, 5, 1)` input returns `(1, 3, 5, 3)`.

## Expected

The message names the check that fired and the layout that is wanted, for example:

```
Expected 'disparity_tensor' to be channels-last (B, H, W, 1) with a single channel in the last dimension. Got torch.Size([1, 1, 3, 5]).
```

The `_check_disparity_tensor` docstring said `(B, 1, H, W)` until #4317 corrected it; the method docstring said the same and is corrected there too. This issue is the remaining runtime half, kept out of #4317 because that batch is documentation-only.

## Pins

`test_convention_reproject_disparity_takes_bhw1` in `tests/geometry/camera/test_stereo.py` (added by #4317) pins the accepted and rejected layouts and matches on the current message text, so the fix updates that `match=` in the same change. Batch tracker: #4262.

NaCr05:

Correct the shared disparity validation message to describe the required channels-last `(B, H, W, 1)` shape and report the received shape. A channels-first input such as `(1, 1, 3, 5)` previously produced an error about dimension 1 even though that dimension was already 1.

Add regression coverage for both `StereoCamera.reproject_disparity_to_3D` and the module-level function, using channels-first and multi-channel inputs. Update the existing layout convention test introduced by #4317 to match the new message, and replace its obsolete comment with a regression note for #4374.

Fixes #4374.

#4317 has merged. This branch includes upstream main at `c1500866e9103c67e75cd9f44206c1de49141ce1`; the follow-up in `0482d18158cc55230b78b01bafb3abb8229aae32` synchronizes its existing assertion and comment with the corrected message. The published head is `64b43eecc671aa5c51ec33984411c38e60c7d3a7`, including the subsequent main synchronization and changelog conflict resolution.

## What did you check?

Windows 11, Python 3.14.6, PyTorch 2.14.0+cpu. The following checks were run in the original fix branch after integrating the upstream changes and applying the review follow-up:

```text
pixi run --locked -e default test-module tests/geometry/camera/test_stereo.py --device=cpu --dtype=float32,float64 --tb=short -q -rx -p no:cacheprovider
60 passed, 2 xfailed

pixi run --locked -e default pre-commit-all
All 12 hooks passed

git diff --check
Passed
```

Both expected failures are the existing strict #4269 u/v convention cases. Before the follow-up, the existing layout test reproduced the two regex failures on CPU float32/float64 when the original PR was combined with the merged #4317.

GPU tests, the full repository suite, and the full documentation build were not run locally.

## How did AI help?

Codex investigated the issue and related PR, prepared the implementation, regression tests and review follow-up, and ran the failure reproduction, focused tests and full pre-commit checks.


ducha-aiki (changes requested):
The message change at `kornia/geometry/camera/stereo.py:283` is correct, but #4317 merged after this head and added an assertion for the old text at `tests/geometry/camera/test_stereo.py:329`. A clean trial merge into current `main` reproduces two failures (CPU float32 and float64) because that regex still expects `to be 1 for as single channeled disparity map`; its adjacent comment also still says #4374 is unresolved. Please update that existing convention test and prose in this PR.

This head's own focused cases pass (8 passed), while the current-main integration selection reports 2 failed, 30 passed, and 2 xfailed.

NaCr05:
Thanks for reviewing this and checking the integration with #4317.

Addressed in `0482d18158cc55230b78b01bafb3abb8229aae32`. The branch now includes upstream main at `c1500866e9103c67e75cd9f44206c1de49141ce1`, with the intervening changelog conflict resolved in head `64b43eecc671aa5c51ec33984411c38e60c7d3a7`. The follow-up updates `test_convention_reproject_disparity_takes_bhw1` to match the `channels-last shape (B, H, W, 1)` message and replaces the obsolete comment with a regression note for #4374. The existing shape and numerical assertions remain intact.

I reproduced the two regex failures before the test update. After integrating the upstream changes and applying the follow-up in the original fix branch, the complete `tests/geometry/camera/test_stereo.py` file passed on CPU float32/float64: **60 passed, 2 xfailed**. Both expected failures are the existing #4269 u/v convention cases. Full pre-commit and `git diff --check` also passed. These checks used Windows, Python 3.14.6 and PyTorch 2.14.0+cpu; I did not run GPU tests or the full repository suite.

Codex prepared the follow-up and ran the local verification. Could you take another look when convenient?

ducha-aiki (approved):
Re-review of `64b43eecc671aa5c51ec33984411c38e60c7d3a7`.

The follow-up updates the existing channels-last convention assertion and its obsolete comment, retaining numerical coverage. Complete stereo tests passed 60 CPU float32/float64 cases with 2 expected xfails. Current-main trial merge is clean and hosted checks pass. The previous integration blocker is resolved.

ducha-aiki (approved):

PR #4314, opened 2026-09-06
fix(geometry): validate pixel2cam depth shape before transformation

KehaoXu:
### What does this pull request do?

Fix the depth shape guard in `pixel2cam` so it rejects inputs outside the documented `Bx1xHxW` shape. Previously, four-dimensional inputs bypassed the channel check: three-channel depth could silently scale each camera coordinate separately, while scalar and one-dimensional inputs raised `IndexError` from the guard.

Add parametrized regression coverage for scalar, one-, three-, four-, and five-dimensional inputs, including two- and three-channel depth. Valid depth inputs keep the same transformation path.

### Related issue or discussion

Addresses the depth-input guard in #4266 only. The other camera shape guards remain outside this PR. #4294 documents the existing behavior and may need coordination if it lands first.

### What did you check?

On Windows, Python 3.13.9 and PyTorch 2.14.0+cpu:

```text
pytest tests/geometry/camera/test_pinhole.py --device=cpu --dtype=float32,float64 -p no:cacheprovider -q
49 passed, 1 skipped

pytest tests/geometry/camera/test_pinhole.py -k invalid_depth_shape --device=cpu --dtype=float16,bfloat16 -p no:cacheprovider -q
12 passed, 38 deselected

Before the fix, the new float32/float64 cases produced 8 failures and 4 passes.
Ruff check, Ruff format --check, git diff --check: passed.
```

`python -m pre_commit run --all-files`: all applicable hooks passed (requirements.txt hook skipped because there were no matching files). The first run normalized checkout line endings; the second run passed. Checkout-only line-ending changes outside this patch were then restored; the content diff remains limited to the three listed files.

### How did AI help?

Codex helped investigate the issue and related PRs, prepare the patch and regression tests, and run the local checks. The contributor reviewed the final diff before submission.

PR #4381, opened 2026-09-09
fix(geometry): validate pixel2cam inverse intrinsics shape

NaCr05:

Validate the full `(B, 4, 4)` shape of `pixel2cam`'s `intrinsics_inv` before transforming pixel coordinates. The previous rank-only guard admitted `(B, 3, 3)` matrices that failed later with an unrelated error, and non-square `(B, 3, 4)` / `(B, 5, 4)` matrices that could return two / four coordinate components instead of three.

Keep the existing `ValueError` message and valid-input transformation path. Replace the old wart test with parametrized rejection tests, add numerical coverage for matching and broadcast matrix batches, and update the affected docstring and changelog. The numerical fixture uses `fx=2, fy=4, cx=4, cy=3`: pixel `(6, 11)` at depth `2` gives camera point `(2, 4, 2)`.

Addresses **only the `pixel2cam` inverse-intrinsics guard** in #4266. Please keep #4266 open for its remaining guards and the separate `PinholeCamera` layout decision. The depth-input guard was already fixed in #4314.

## What did you check?

Locally on Windows 11, Python 3.14.6 and PyTorch 2.14.0+cpu, using the repository Pixi tasks and the isolated worktree's virtual environment:

```text
pixi run --locked -e default test-module tests/geometry/camera/test_pinhole.py::TestPixel2Cam -k "invalid_intrinsics_shape_4266 or intrinsics_batch_broadcast" --device=cpu --dtype=float32,float64 --tb=short -q -p no:cacheprovider
# Before the guard fix: 12 failed, 14 passed, 27 deselected.

pixi run --locked -e default test-module tests/geometry/camera/test_pinhole.py tests/geometry/test_depth_warper.py --device=cpu --dtype=float32,float64 --tb=short -q -rs -p no:cacheprovider
# After the fix: 143 passed, 1 skipped (the slow pixel2cam gradcheck).

pixi run --locked -e default pre-commit-all
# All hooks passed.

pixi run --locked -e default typecheck
# Exit 0; one custom_fwd deprecation warning in kornia/feature/lightglue.py.

git diff --check
# Passed.
```

A separate CPU probe checked 20 invalid-shape cases across float32/float64: all now raise the intended input error, including the rectangular matrices. Six valid configurations retained the expected numerical result, including a single inverse matrix broadcast across two image batches. The full repository suite, GPU tests and slow gradcheck were not run locally.

## How did AI help?

Codex assisted with issue and related-PR inspection, reproduction, the code/test/docstring changes, and local validation. The final diff was checked for the narrow scope described above.


ducha-aiki (dismissed):
The full `(B, 4, 4)` check at `kornia/geometry/camera/pinhole.py:859` rejects the previously admitted square and rectangular matrices before `transform_points`, while preserving the intended single-matrix broadcast. The focused invalid-shape and numerical broadcast coverage passes on CPU float32/float64 (20 cases), and hosted CI is green. I found no correctness blocker in this head.

I did not rerun the slow gradcheck, GPU backends, or the full suite locally.

PR #4382, opened 2026-09-09
fix: respect align_corners in elastic transform grid

### LINKED ISSUE #4235: [Bug]: elastic_transform2d is not identity for zero displacement with align_corners=False
## What happened?

`elastic_transform2d` does not preserve the input image when the displacement field is zero and `align_corners=False` is used.

A zero displacement field represents an identity deformation, so the output should be equal to the input up to floating-point tolerance.

The current implementation constructs the sampling grid using `create_meshgrid(h, w)` and then passes it to `torch.nn.functional.grid_sample` with the selected `align_corners` value.

Since `create_meshgrid` produces the corner-aligned normalized `[-1, 1]` grid, this grid is not consistent with the coordinate convention used by `grid_sample` when `align_corners=False`.

As a result, even with an exactly zero displacement field, the image is resampled at shifted coordinates.

### Expected behavior

With:

python
noise = torch.zeros(B, 2, H, W)

the transformation should be an identity transformation:
elastic_transform2d(image, noise, ...) == image
within numerical tolerance.

Actual behavior

On the current main checkout, using a 4x4 deterministic input and a zero displacement field produces a maximum absolute difference of:
tensor(11.2500)
So the input is substantially changed even though the displacement is exactly zero.

This behavior is specific to the align_corners=False coordinate convention; the same corner-aligned grid is consistent with align_corners=True.

### Minimal reproduction

```python
import torch
from kornia.geometry.transform import elastic_transform2d

image = torch.arange(16.0).reshape(1, 1, 4, 4)
noise = torch.zeros(1, 2, 4, 4)

output = elastic_transform2d(
    image,
    noise,
    kernel_size=(3, 3),
    sigma=(1.0, 1.0),
    alpha=(1.0, 1.0),
)

print("Input:")
print(image)

print("\nOutput:")
print(output)

print("\nMax absolute difference:")
print((output - image).abs().max())
```

### Environment

Kornia 0.9.0rc1 | PyTorch 2.14.0+cpu | Python 3.13.11 | Windows

### Anything else?

_No response_

3901984509cry:
## Summary

- Build the elastic transform identity grid with the same coordinate convention as `grid_sample`.
- Preserve identity behavior for zero displacement with both values of `align_corners`.
- Generate reduced-precision grids in float32 for CPU compatibility.
- Add regression tests.

Fixes #4235

## Tests

- CPU elastic transform tests: 72 passed
- CUDA elastic transform tests: 18 passed, 36 skipped
- Geometry transform tests: 561 passed, 15 skipped, 1 xfailed
- pre-commit checks passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -410,6 +410,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Make calibration `distort_points` and `undistort_points` tilt checks compatible with
   `torch.compile(fullgraph=True)`, preserving eager behavior. (#4391)
 * Canny hysteresis preserves the input dtype, avoiding a convolution dtype mismatch for half-precision images. (#4393)
+* `elastic_transform2d` now builds its identity sampling grid with the requested `align_corners`
+  convention, so a zero displacement field preserves the input image. (closes #4235). (#4382)
 
 * `rad2deg` and `deg2rad` now handle integer tensor inputs correctly and preserve
   float64 precision. `angle_to_rotation_matrix` inherits the corrected conversion,
--- a/kornia/geometry/transform/elastic_transform.py
+++ b/kornia/geometry/transform/elastic_transform.py
@@ -23,7 +23,6 @@ import torch.nn.functional as F
 from kornia.core.check import KORNIA_CHECK_IS_TENSOR, KORNIA_CHECK_SHAPE
 from kornia.filters import filter2d_separable
 from kornia.filters.kernels import get_gaussian_kernel1d
-from kornia.geometry.grid import create_meshgrid
 
 __all__ = ["elastic_transform2d"]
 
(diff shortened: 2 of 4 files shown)


3901984509cry:
All CI checks are now passing, including the CPU bfloat16 baseline. Could a maintainer please review this PR when available?

ducha-aiki (changes requested):
`CHANGELOG.md:367` cites only the fixed issue as `(closes #4235)`. Repository policy requires each user-visible change to cite its pull request number, so please change this entry to cite `#4382` (it may mention #4235 as well).

The implementation itself looks correct: the full elastic-transform file passes on CPU float16, bfloat16, float32, and float64 (72 cases total), and hosted CI is green. I did not rerun CUDA or MPS locally.

3901984509cry:
Thanks for the review. I updated the changelog entry to reference PR #4382 and rebased the branch onto the latest upstream main. All 38 CI checks are passing now. Could you please re-review it?

ducha-aiki (approved):
Re-review of `9c1d9365bb7c79efbe547333de82f70875eee6be`.

The requested CHANGELOG citation now names #4382. The identity grid uses the requested align_corners convention, with a float32 grid for reduced-precision inputs before casting back. Complete elastic-transform tests passed 36 CPU float32/float64 cases in this re-review; the earlier review covered all four CPU dtypes. Current-main trial merge is clean and hosted checks pass. No remaining blocker. #4403 targets the same underlying defect; merge one implementation and re-evaluate the duplicate.

PR #4388, opened 2026-09-09
Fix parent fixture teardown for isolated half-precision tests

### LINKED ISSUE #4149: --isolate-half-precision: mixed multi-file CUDA runs error with "previous item was not torn down properly" on unrelated non-half tests

Con-Benksl:

When a normal test is followed by an isolated half-precision test from the same module, pytest may retain the module's fixtures for that next item. The isolation hook runs the next item in a subprocess but never unwinds the parent's fixture stack. A following test from another module then fails with `previous item was not torn down properly`.

Unwind the parent's pending fixtures for the next item after reporting the subprocess result. Capture finalizer exceptions as teardown reports, and propagate `KeyboardInterrupt` and `pytest.exit`. The isolated item's fixtures still run only in its subprocess. This deliberately avoids invoking the full parent teardown hook, whose plugins can depend on setup state that the isolated item never created.

Fixes #4149.

## What did you check?

- Added two pytester regressions using the real isolation hook and real subprocesses. Both fail on the unchanged main branch and pass with the fix. They check fixture ordering across modules and correct reporting of a failing parent finalizer.
- `pixi run test-module tests/test_conftest.py tests/test_half_precision_ci.py -q`: **78 passed**.
- `pixi run pre-commit-all`: passed all hooks.
- `pixi run typecheck`: passed; reports the existing `custom_fwd` deprecation warning in LightGlue.
- Additional subprocess checks exercise `-x` with a failing parent finalizer, `KeyboardInterrupt`, and `pytest.exit`.

Validated on macOS with Python 3.13.13, PyTorch 2.14.0, and pytest 9.0.3. The regression uses CUDA/float16 selection metadata without allocating CUDA tensors, so it verifies pytest's isolation protocol on a CPU host. CUDA kernels and the full GPU suite were not run.

## How did AI help?

Codex investigated the fixture lifecycle, implemented the fix and regression tests, and ran the checks above. A separate agent reviewed the change before submission.


ducha-aiki (dismissed):
Reviewed head ae88fbff. The isolated protocol now performs the parent-side setup-state transition that the default pytest protocol would have performed, reports finalizer exceptions as teardown errors, and honors stop/fail state by unwinding the full stack. The regression covers both successful and failing module finalizers and verifies that the next module can set up. In an isolated source checkout with the subprocess import path pinned to this PR, both new pytester integration cases passed; hosted CPU, half, MPS, Dynamo, typecheck and tutorial jobs are green. I found no actionable issues. I did not run an actual CUDA device or the whole suite.

PR #4390, opened 2026-09-10
Keep list padding empty in box masks

### LINKED ISSUE #4252: Bug: `Boxes.to_mask` on a list-backed object fills pixel (0, 0) for every padding row
`Boxes.to_mask` exports with `to_tensor("xyxy", as_padded_sequence=True)`, and the zero padding row of a list-backed object exports as the exclusive `xyxy` box `[0, 0, 1, 1]` under the inclusive `+1` convention (the "unit box" padding behavior recorded on the `Boxes` anchor since #4060). The mask then marks pixel `(0, 0)` in the channel of every padding entry, so the returned `(B, N_max, H, W)` mask carries a spurious pixel per padded slot instead of an empty channel. This is the first concrete consumer of the padded-row value; it is the same family as #4179 (padding rows exported as boxes after indexing).

Reproduced 2026-09-05 at `1eea5835`, PyTorch 2.14.0, CPU and MPS.

## Reproduction

```python
import torch

box = torch.tensor([[[1., 1.], [2., 1.], [2., 2.], [1., 2.]]])
boxes = Boxes([box, torch.cat([box, box + 2.0])])      # first sample has one box, second has two
mask = boxes.to_mask(5, 5)
print(mask.shape, "padding channel:", mask[0, 1].nonzero().tolist())
print(boxes.to_tensor("xyxy", as_padded_sequence=True)[0, 1].tolist())
```

## Observed

```text
torch.Size([2, 2, 5, 5]) padding channel: [[0, 0]]
[0.0, 0.0, 1.0, 1.0]
```

## Expected

An empty mask channel for a padding entry (or no channel at all, if `to_mask` should honor the list cardinality the way `to_tensor` does without `as_padded_sequence`). Either choice depends on the padded-row value contract, which is not settled: the batch-4a audit recorded it as a design candidate because the padded row's value after mode conversion is unspecified. Batch 4c documents the pixel on `to_mask` and pins it (`test_wart_to_mask_fills_the_origin_pixel_for_list_padding_rows_4252`).

## Classification

Broken output for list-backed input (a padding artifact surfaces as data). Idiom 3, inherited from the padded-row conversion rather than from `to_mask` itself.

## Disposition

n/a — single symbol

The fix is coupled to the padded-row contract and to #4179; whoever settles that contract should close both.

Con-Benksl:

`Boxes.to_mask()` currently treats the padded slots of a variable-length list as real boxes: zero padding fills the origin pixel, and transformed padding can fill a different pixel. Use the existing per-image padding counts to give those slots empty bounds before either mask path fills them. Real point boxes still cover one pixel, and the stored coordinates and `to_tensor()` exports are unchanged.

Update the documented convention and its regression pin to cover an empty image, unequal list lengths, real point boxes, and transformed padding in both mask paths.

Fixes #4252 for list-backed containers that retain their padding metadata. The separate `__getitem__` metadata-loss issue (#4179, with PR #4194) is unchanged.

## What did you check?

- The 8 regression cases fail on the base revision with spurious padding pixels. With the fix, `pixi run test-module tests/geometry/test_boxes.py tests/geometry/test_bbox.py --device=cpu --dtype=float32,float64 -q` reports 317 passed and 24 existing expected failures.
- CPU `to_mask` tests with the project's known-failure profiles: 19 passed for float16 and 19 passed for bfloat16.
- The official tensor-input export case plus list-padding and transformed-list cases: ONNX checker/ONNX Runtime and `torch.export` execution match eager output. Ordinary tensor output and real box pixels are bitwise unchanged against the base; only the padding pixels disappear. Default `torch.compile` matches eager and retains the base's 1/2/3 graph breaks for those cases. An additional CPU `fullgraph=True` check hits the same existing dynamic-slice Inductor failure on both revisions.
- `pixi run pre-commit-all` passes all 12 hooks; `pixi run typecheck` exits successfully with the existing `lightglue.py` deprecation warning.

Checks ran locally on CPU with PyTorch 2.14.0; CUDA was not available for validation.

## How did AI help?

Codex investigated the issue, implemented the change and regression tests, and ran the checks. A separate Codex agent reviewed the fix, call sites, padding metadata handling, and test evidence before submission.


ducha-aiki (dismissed):
Reviewed head d255980c. Resetting each list row padding tail after coordinate conversion correctly prevents both the loop and vectorized mask paths from interpreting transformed padding as boxes, while leaving the stored box tensor unchanged. Coverage includes empty rows, real zero-area point boxes, translated padding, and both execution paths. In an isolated source checkout with kornia.__file__ verified, the complete CPU float32/float64 boxes file passed (226 passed, 22 expected xfails). Hosted CPU, half, MPS, Dynamo, typecheck and tutorial jobs are green. I found no actionable issues; I did not run CUDA or the whole suite.

PR #4389, opened 2026-09-09
Fix TorchScript compilation of depth projections

Con-Benksl:
## What does this pull request do?

`torch.jit.script` cannot compile `unproject_meshgrid` because its `F.normalize` call passes an integer norm order where TorchScript expects a float. This also prevents scripting `depth_to_3d_v2` and `warp_frame_depth`, which call the shared helper.

Pass `p=2.0`, following the existing camera projection implementation. Add scripting regressions for the three entry points, including normalized rays and cached grids, and record the fix in the changelog.

## Related issue or discussion

Related to #4286: addresses its depth projection scripting failures. The other findings in that issue are outside the scope of this PR.

## What did you check?

- The final eight regression cases fail on the unchanged base with the norm-order type error. Before syncing upstream, the modified branch passed all 175 depth/perspective tests on CPU float32/float64. After merging upstream `c150086`, the full depth test file passes with 191 passed and 2 expected failures on CPU float32/float64.
- The new scripting cases also pass on CPU float16 and bfloat16: four cases per dtype. Warping is compared with eager execution, preserving its existing interpolation precision.
- `pixi run pre-commit-all` passed. `pixi run typecheck` exited successfully with the existing LightGlue deprecation diagnostic. `git diff --check` passed.
- Eight export cases, covering normalized/unnormalized rays and cached grids, passed ONNX checker/runtime, `torch.export`, and `torch.compile` checks in independently executed base and patched worktrees. All had zero graph breaks; eager outputs were exactly equal between revisions.

Environment: Python 3.13.13, PyTorch 2.14.0. Tests use synthetic tensors; no model downloads, GPU or full-repository test suite was run. PyTorch emits its TorchScript deprecation warning.

## How did AI help?

OpenAI Codex implemented and tested the change autonomously. A separate agent reviewed the shared call path and final diff, independently reran the export comparisons, and checked scripting save/load and numerical parity. That review also replaced an overly strict half-precision identity assertion with a comparison against eager output.

PR #4384, opened 2026-09-09
fix(calibration): correct forward sensor tilt projection

UnbearableFate:
## What does this pull request do?

Non-zero sensor tilt made `distort_points` disagree with OpenCV by over 600 pixels on the existing 14-coefficient reference, and prevented distortion/undistortion round trips. The forward matrix now uses `Pz @ R`, as defined by OpenCV's `computeTiltProjectionMatrix`; the inverse formula is retained.

The existing independent OpenCV reference now exercises forward distortion. Tests that documented the defect now assert the correct matrix and point round trips, and the obsolete #4276 API warnings are removed. Non-zero tilt also corrects the sampling map used by `undistort_image`; zero-tilt behavior is unchanged.

## Related issue or discussion

Fixes #4276.

Merge commit `e38f81a821af4d8e08dfdf901945d1a893695ba0` integrates main at `a7e2db6bef0482ae4fdea4cbc776e3c8568c608f` on top of the maintainer's reviewed head `bdcf2e4d`. The only manual conflict resolution is in `CHANGELOG.md`, preserving both sides' entries. Calibration source and tests remain identical to `7cf133b`, retaining #4345's batch-shape fix and #4391's tensor-only compile/export path alongside this PR's forward matrix correction.

## What did you check?

Fresh validation of the merged tree at `e38f81a`, with interpreter and worktree import origin checked:

- Complete calibration CPU float32/float64: 144 passed, 1 existing float64-only PnP pin skipped.
- Full repository pre-commit, Pixi typecheck, and diff whitespace checks passed; the worktree is clean.
- The resolved tree differs from Git's automatic merge tree only in `CHANGELOG.md`; its changelog diff against current main adds this PR's two-line entry without removing any main line.

Local environment: Pixi, Python 3.12.12, PyTorch 2.14.0+cpu. CUDA/MPS, the full repository matrix, Inductor/export, and documentation execution were not rerun for this changelog conflict resolution. Earlier numerical, fullgraph, export, and gradient validation on the unchanged calibration files is recorded in the [September 11 audit](https://github.com/kornia/kornia/pull/4384#issuecomment-5634739637); those results are not new runs on this merge commit. Fresh hosted CI is running.

The original implementation was also reproduced before the runtime fix: all 8 updated CPU float32/float64 regression cases failed, including the independent OpenCV forward reference. The existing low-precision inverse-reference tolerance is unchanged.

## How did AI help?

[...]

PR #4391, opened 2026-09-10
fix(calibration): avoid data-dependent tilt checks during compile

UnbearableFate:

`distort_points` and `undistort_points` raise `Unsupported: Data-dependent branching` under `torch.compile(fullgraph=True)` because their non-zero tilt check reads tensor values on the host. Explicitly recognize compilation using the existing `is_compiling` helper and take the same unconditional tilt path used by export. The eager and scripted coefficient checks remain unchanged.

Add fullgraph regressions for both functions with 4 and 14 coefficients. The tests check real compiler output and assert `CompileCounter.frame_count == 1` while reusing the graph with zero and non-zero tilt. Update the affected docs and changelog.

Addresses the two calibration fullgraph failures in #4286. The depth scripting failures are handled separately by #4389; this does not close the entire issue.

Integration: #4345 fixes batch shapes and #4384 fixes forward tilt arithmetic in the same functions. This PR targets current main; both independent limitations are retained/documented and may require rebasing depending on merge order.

## What did you check?

- New fullgraph regressions: all 8 CPU float32/float64 cases fail before the fix and pass afterwards; all 8 float16/bfloat16 cases also pass with Inductor.
- Calibration suite on CPU float32/float64: 138 passed, 1 skipped, 2 existing #4276 xfails; compile tests are collected separately with `KORNIA_TEST_OPTIMIZER=inductor`.
- 20 deterministic eager cases across all supported coefficient lengths and float32/float64 are byte-identical between the baseline and patched calibration code (independent processes/import paths).
- Both functions pass `torch.export` and ONNX Runtime parity with 14 non-zero tilt coefficients; calibration doctests: 4 passed.
- Full `pixi run pre-commit-all` passed. `ty check kornia` exited successfully with the unchanged LightGlue deprecation warning. `git diff --check` passed.

Environment: Python 3.12.3, PyTorch 2.14.0+cpu. No CUDA/MPS or full-repository test matrix was run locally.

## How did AI help?

OpenAI Codex (GPT-6) reproduced, implemented, and validated the change. Lower-cost agents screened other issues and monitored submitted PRs. The parent checked the final diff, baseline failures, single-graph capture, eager equivalence, and export compatibility.


ducha-aiki (dismissed):
Reviewed head 7ef50332. The capture guard routes both calibration functions through the tensor-only tilt path during compilation while preserving the eager and TorchScript branch, and the tests exercise four- and fourteen-coefficient inputs while changing tilt values without recompilation. The #4286 stale warnings are removed and the release note records the behavior change. In an isolated source checkout with kornia.__file__ verified, all 8 CPU float32/float64 fullgraph/Inductor cases passed. Hosted CPU, half, MPS, Dynamo, typecheck and tutorial jobs are green. I found no actionable issues; I did not run CUDA, ONNX, or the whole suite.

Issue #4272, opened 2026-09-05
W1b: `depth_from_disparity` — disparity 0 returns 2.5e9, and "float/tensor" means a tensor of shape `(1,)`


ducha-aiki:
Two problems in `depth_from_disparity`, both visible from its own docstring.

1. **`baseline * focal / (disparity + 1e-8)`.** The epsilon is inside the arithmetic, so a disparity of exactly `0` — the standard encoding for "no match" / sky / infinity in every stereo pipeline — returns `2.5e9` rather than `inf` or a masked value, and biases every small-disparity result. In float16 it saturates to `inf` anyway, so the epsilon does not even buy what it was added for. A negative disparity produces a silent negative depth.
2. **"float/tensor" is not true.** The Args say `baseline: float/tensor` and `focal: float/tensor`, but a tensor must be exactly shape `(1,)` — a 0-dim tensor and a `(2,)` tensor both raise `ShapeError` — and a python `int` is rejected outright with a `BaseError` that names `float` and `torch.Tensor` as the only accepted types.

## Repro

```python
import torch, inspect
from kornia.geometry.depth import depth_from_disparity

doc = inspect.getdoc(depth_from_disparity)
print("docstring Args:", [l.strip() for l in doc.splitlines() if l.strip().startswith(("baseline", "focal"))])
d = torch.tensor([[[[1.0, 2.0, 0.0]]]])
b, f = torch.tensor([0.5]), torch.tensor([50.0])
print("depth at disparity [1, 2, 0]   :", depth_from_disparity(d, b, f).tolist())
print("float16 at disparity 0         :", depth_from_disparity(d.half(), b.half(), f.half()).tolist())
print("negative disparity -1          :", depth_from_disparity(torch.tensor([[[[-1.0]]]]), b, f).tolist())
for name, bb, ff in (("0-dim tensor", torch.tensor(0.5), torch.tensor(50.0)),
                     ("(2,) baseline", torch.tensor([0.5, 0.5]), f),
                     ("python float/int", 0.5, 50)):
    try:
        depth_from_disparity(d, bb, ff)
        print(f"{name:<17}: accepted")
    except Exception as e:
        print(f"{name:<17}:", type(e).__name__ + ":", str(e).replace("\n", " ")[:80])
```

## Observed

```
docstring Args: ['baseline: float/tensor containing the distance between the two lenses.', 'focal: float/tensor containing the focal length.']
[...]

PR #4392, opened 2026-09-10
Accept scalar camera parameters in depth_from_disparity

Con-Benksl:

`depth_from_disparity` rejects Python integer camera parameters and zero-dimensional tensors, although both represent the same scalar baseline or focal length as the accepted float and `(1,)` tensor forms. Widen the two parameter guards to accept them. Keep boolean and non-scalar inputs rejected, and document the accepted shapes.

The arithmetic is unchanged. Tests cover all combinations of integer, float, scalar-tensor and `(1,)` tensor parameters for unbatched and batched disparity maps, invalid inputs, and gradients through scalar tensors.

Related to #4272, item 2 only. This does not change the epsilon, zero/negative disparity behavior, or add per-batch camera-parameter broadcasting. The convention documentation and wart test added by #4317 now reflect the accepted scalar forms, while retaining the unsupported per-batch calibration checks.

## What did you check?

- Before changing the guards, the focused class reports 49 failed / 39 passed; the new parameter forms and scalar-tensor gradcheck fail at the old guards.
- `pixi run test-module tests/geometry/test_depth.py --device=cpu --dtype=float32,float64 -q`: 252 passed, 2 xfailed on main `c1500866`, including float64 gradchecks for scalar and `(1,)` tensor parameters.
- The focused `TestDepthFromDisparity` class with `--dtype=float16,bfloat16`: 92 passed.
- A separate `ty` check with `invalid-argument-type` explicitly enabled accepts Python integer baseline/focal arguments through the public `float | Tensor` annotations. A string-argument negative control produces the expected type error.
- Independent baseline and patched worktrees: 32 accepted-input cases are bitwise unchanged across four dtypes, including zero and negative disparities. Two existing export survey cases and three new scalar cases pass ONNX checker/Runtime, `torch.export` execution and Inductor `torch.compile(fullgraph=True)` execution, with no graph breaks. The new cases are rejected on the baseline. Direct TorchScript retains its existing unsupported `type(baseline)` diagnostic on both revisions.
- Full `pixi run pre-commit-all`: all 12 hooks passed. `pixi run typecheck`: successful exit with the existing LightGlue deprecation warning.

Validated locally on CPU, Python 3.13.13 / PyTorch 2.14.0. CUDA/MPS and the full repository suite were not run.

## How did AI help?

Codex investigated the issue, implemented the guards and tests, and ran the checks. A separate Codex agent reviewed the change, call sites, compatibility and validation evidence before submission.


ducha-aiki (changes requested):
Reviewed head e9ef6357. Two integration/API issues need correction:

- kornia/geometry/depth.py:591-593 documents and accepts Python integers at runtime, but the public annotations remain float | torch.Tensor. Static users therefore cannot call the newly supported API without a type error. Please add int to both baseline and focal annotations.
- This head conflicts with current main after #4317. The resolution must update the newly landed depth_from_disparity Convention text and test_wart_depth_from_disparity_rejects_a_batched_baseline_4272 (kornia/geometry/depth.py around current-main line 800 and tests/geometry/test_depth.py around line 1142): they currently say/assert that scalar tensors and Python ints are rejected. Keeping those pins makes the documented/tested behavior contradict this patch.

The runtime arithmetic and guards otherwise look sound. In an isolated branch checkout, the complete CPU float32/float64 depth file passed (204 tests), with source origin verified. Hosted checks on this head are green, but they predate the current-main conflict resolution.

Con-Benksl:
Synchronized with main `f36ab6de` after #4317, preserving the existing branch history. The Convention/Args now list Python integers and scalar tensors. Removed their obsolete rejection assertions from the wart test, retaining the unsupported per-batch calibration checks; the existing parameterized regression covers all four accepted forms. CPU depth tests: 252 passed, 2 xfailed; the disparity class in float16/bfloat16: 92 passed. Full pre-commit and typecheck pass.

For the annotation point, `float | torch.Tensor` already accepts integer arguments under Python’s numeric typing rule. The project’s [Ruff PYI041 rule](https://docs.astral.sh/ruff/rules/redundant-numeric-union/) removes `int` from an `int | float` parameter union. I verified this against the public function using typed `int` variables and `ty check --error invalid-argument-type --error no-matching-overload --error invalid-assignment --error unresolved-import --error unresolved-attribute`: integer baseline/focal calls pass. A string-argument negative control fails at this function with `Expected int | float | Tensor`, so the check is not relying on the repository’s default ignored argument diagnostic. The annotations retain the lint-compatible form; the documentation explicitly includes `int`.

This follow-up and the independent review were performed by Codex agents. No CUDA/MPS or full repository suite was run.

ducha-aiki (approved):
Re-review of `f0016077e7cf26064038a39b53afc31640223098`.

The scalar-parameter follow-up correctly updates the Convention/Args and removes obsolete rejection assertions while retaining the unsupported per-batch calibration checks. I retract the earlier annotation objection: Python numeric typing accepts int arguments for float parameters, so the existing float | Tensor annotation is appropriate; adding int would be redundant. The full depth file passed 252 CPU float32/float64 cases with 2 expected xfails. Current-main trial merge is clean and hosted checks pass.

PR #4393, opened 2026-09-10
Fix Canny hysteresis for half-precision images

Con-Benksl:

Canny's hysteresis loop converts its weak and strong edge masks to float32. With float16 or bfloat16 input, this promotes the next edge map to float32: a second iteration then raises a convolution dtype mismatch, while a one-iteration result silently returns the wrong dtype.

Keep both masks in the input dtype. Tests cover a blank image and promotion of a weak edge adjacent to a strong edge, through both the function and module APIs. The thresholding and propagation rules are unchanged.

Related to #4153, the Canny dtype failures only.

Remove five recovered half-precision manifest entries. The two `test_magnitude_hyst` entries now reach their existing magnitude-tolerance assertions, so retain those entries as `AssertionError` rather than `RuntimeError`. No numerical tolerances are changed.

## What did you check?

- New regression on the unmodified main: 8 failed / 8 passed. Half-precision blank images fail the output-dtype assertion; weak-edge inputs raise the convolution dtype mismatch.
- Full `tests/filters/test_canny.py` on CPU float32/float64: 93 passed, including the existing gradcheck.
- Full Canny file with the corresponding strict known-failure profile, on both PyTorch 2.9.1 and 2.14.0: float16 has 43 passed / 4 xfailed; bfloat16 has 44 passed / 3 xfailed. Separate baseline runs and assertion traces confirm which manifest entries this fix recovers.
- New regression on MPS float32: 4 passed.
- Independent baseline/patched checks on 16 image/dtype cases: patched edges match an eight-neighbor graph traversal of each input's thresholded edge map; the eight previously supported float32/float64 cases retain bitwise-identical magnitude and edge outputs.
- Full `pixi run pre-commit-all`: all 12 hooks passed. `pixi run typecheck`: successful exit with the existing LightGlue deprecation diagnostic.

Local tests used Python 3.13.13 on macOS arm64. CUDA and the full repository suite were not run. This does not add ONNX support to Canny.

## How did AI help?

Codex investigated, implemented and tested the fix. Separate Codex agents reviewed the change, callers, issue ownership and validation evidence; independent numerical checks compared the original and patched implementations.


ducha-aiki (approved):
Reviewed head f9ddc7d1. The two mask casts now preserve the input dtype through every hysteresis iteration, and the regression covers both function/module entry points plus the multi-iteration weak-edge promotion path. Local validation from an isolated source checkout with kornia.__file__ verified passed the full CPU float32/float64 Canny file (93 passed) and the new CPU float16/bfloat16 selection (8 passed). Hosted CPU, half-precision, MPS, Dynamo, typecheck and tutorial jobs are green. I found no actionable issues. I did not run CUDA or the whole repository suite.

ducha-aiki (approved):

PR #4394, opened 2026-09-10
Fix fft_conv for CPU half-precision inputs

Con-Benksl:

`fft_conv` currently passes CPU float16 and bfloat16 tensors directly to `torch.fft.rfftn`, which rejects both dtypes. Promote the two FFT operands to float32 on that path and cast the cropped result back to the input dtype. Kernel normalization and spatial padding retain their existing order and precision; other devices and input dtypes retain their existing path.

Add a regression against `filter2d` covering per-batch kernels, convolution/correlation, normalization, and same/valid padding. Remove the 37 recovered FFT test entries from each CPU half-precision manifest and document the CPU computation dtype.

Related to #4153. This addresses the `fft_conv` failures only; the broader half-precision tracker remains open.

## What did you check?

Local checks ran on macOS arm64 / Python 3.13.13. Both PyTorch 2.9.1 (the version pinned by CPU half CI) and 2.14.0 were tested; this is not a Linux or CUDA validation claim.

- New regression: **16 half-precision failures / 16 normal-precision passes before**, **32 passes after**.
- Full `tests/filters/test_filters.py`: **234 passed, 3 skipped** with float32/float64 on 2.14.0, including float64 gradchecks. Each CPU half profile on each Torch version: **120 passed, 3 skipped**. All 74 removed entries were individually confirmed as strict XPASS before removal.
- Existing `test_dynamo` cases with `KORNIA_TEST_OPTIMIZER=inductor`: **16 passed** across four CPU dtypes. New regression on MPS float32: **8 passed**.
- Separate baseline/patch runs: 32 float32/float64 cases have bitwise-identical outputs and input/kernel gradients; two integer-input cases retain their float32 output and exact values. Sixteen half-precision cases match spatial filtering in both forward and input/kernel gradients.
- `torch.export` construction and actual exported-module execution: **12 cases passed** across four dtypes and three filtering variants; the six existing float32/float64 cases also pass on the baseline. Direct TorchScript compilation still rejects the existing global set guard on both revisions.
- `pixi run pre-commit-all` and `pixi run typecheck` passed. Typecheck reports the existing LightGlue `custom_fwd` deprecation warning.

## How did AI help?

OpenAI Codex investigated the failure, implemented the fix and regression, and ran the checks. A separate Codex agent reviewed the code, the exact manifest removals, and the baseline/patch evidence.


ducha-aiki (dismissed):
Re-review of `4c072a2ccf604fa23d98a12ae901dcae4782fdf5`.

The CPU-only float16/bfloat16 promotion is confined to FFT computation and casts the output back to the input dtype; ordinary precision and accelerator paths retain their existing behavior. The new spatial-filter comparison covers padding, convolution/correlation and normalization, and the repaired half-runtime-failure entries are removed. The full FFT-convolution class passed 182 CPU tests across float16, bfloat16, float32 and float64, with 1 skip. Code review is clear. Hosted bfloat16 CI is still red in TestSOSNet.test_jit (and collector), rather than the FFT tests, so merge remains blocked until successful CI; I have not established whether that SOSNet failure reproduces on base.

ducha-aiki (approved):
Re-review of exact head `63d54067ef8823d131e8d65d0d5632b457406452` after integrating main.

The PR-specific implementation, focused comparison test, and precision documentation are unchanged from the prior reviewed head. The merge preserved the latest upstream fixes and both release-note histories while retaining only the intended removals of the repaired FFT-convolution half-precision failures. CPU float16 and bfloat16 are promoted only for the FFT computation and cast back to the caller's dtype; ordinary precision and accelerator behavior remain unchanged. Code review is clear.

The full `TestFilter2D_fftconv` class passed 186 CPU cases across float16, bfloat16, float32, and float64, with three skips, in an isolated checkout with its import origin verified. All 41 hosted checks are green. I did not run the whole suite, CUDA, or MPS locally.

PR #4403, opened 2026-09-10
Fix #4235: elastic_transform2d identity mapping with align_corners=False

### LINKED ISSUE #4235: [Bug]: elastic_transform2d is not identity for zero displacement with align_corners=False

dhruvv16-hash:
Fixes #4235

### Description

This PR fixes a bug in `elastic_transform2d` where a zero displacement field does not result in an identity transformation when `align_corners=False`.

The issue was caused by `elastic_transform2d` always using `create_meshgrid(normalized_coordinates=True)` to generate the base grid, which assumes the `align_corners=True` convention (`[-1, 1]` mapped to outer pixel centers). When this grid was passed into `F.grid_sample` with `align_corners=False`, the coordinate meaning shifted, resulting in the image being slightly resampled at shifted coordinates.

I fixed this by explicitly generating the grid according to the selected `align_corners` flag so that it's consistent with `grid_sample` expectations. 

Includes tests that verify identity mappings for both `align_corners=False` and `align_corners=True`.

### DIFF
--- a/kornia/geometry/transform/elastic_transform.py
+++ b/kornia/geometry/transform/elastic_transform.py
@@ -124,7 +124,14 @@ def elastic_transform2d(
 
     # Warp image based on displacement matrix
     _, _, h, w = image.shape
-    grid = create_meshgrid(h, w, device=image.device).to(image.dtype)
+    
+    if align_corners:
+        grid = create_meshgrid(h, w, device=image.device).to(image.dtype)
+    else:
+        grid = create_meshgrid(h, w, normalized_coordinates=False, device=image.device).to(image.dtype)
+        grid[..., 0] = (grid[..., 0] + 0.5) * 2.0 / max(w, 1) - 1.0
+        grid[..., 1] = (grid[..., 1] + 0.5) * 2.0 / max(h, 1) - 1.0
+        
     warped = F.grid_sample(
         image, (grid + disp).clamp(-1, 1), align_corners=align_corners, mode=mode, padding_mode=padding_mode
     )
--- a/kornia/geometry/transform/elastic_transform.py
+++ b/kornia/geometry/transform/elastic_transform.py
@@ -124,14 +124,14 @@ def elastic_transform2d(
 
     # Warp image based on displacement matrix
     _, _, h, w = image.shape
-    
+
     if align_corners:
         grid = create_meshgrid(h, w, device=image.device).to(image.dtype)
     else:
         grid = create_meshgrid(h, w, normalized_coordinates=False, device=image.device).to(image.dtype)
         grid[..., 0] = (grid[..., 0] + 0.5) * 2.0 / max(w, 1) - 1.0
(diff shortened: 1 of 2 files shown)


ducha-aiki (changes requested):
Reviewed exact head `eb1782b96a95dc2848ad04350d0a583540f5a2b0`.

The identity-grid correction looks sound at ordinary precision, and the complete elastic-transform file passed 36 CPU float32/float64 tests. This targets the same defect as #4382, which has already addressed its review requests; I recommend consolidating on that PR rather than merging both implementations.

[P2] If this PR is retained, add the required Unreleased Bug fixes entry citing #4403. This changes public elastic-transform behavior, but the current diff has no CHANGELOG entry. Hosted float16 CI is also cancelled and collector is red, so successful replacement checks are needed.

Nonblocking precision limit: both this PR and #4382 still show zero-displacement differences for sufficiently wide float16/bfloat16 images because normalized sampling coordinates are quantized. Comparison with current main and #4382 did not establish a new regression specific to this implementation; that limitation should be handled separately.

dhruvv16-hash:
Closing as this was addressed and merged in #4382. Thank you!

PR #4405, opened 2026-09-10
fix(liegroup): finite gradients at the identity for every exp and log

### LINKED ISSUE #4404: All SO(3)/SE(3)/SE(2) exp and log return nan gradients at the identity
## Summary

Every `exp` and `log` in `kornia.geometry.liegroup` returns **`nan` gradients at the identity**, on every dtype
including `float64`. The forward values are correct; only the backward pass is poisoned. The zero rotation vector
is the standard initialisation for pose optimisation, so this is the first backward pass of a typical
`So3.exp(...)`-based optimisation loop.

`So2` is the exception — it goes through `atan2` and has no singular branch.

| | forward at identity | backward at identity |
| --- | --- | --- |
| `So3.exp` | correct | **`nan`** |
| `So3.log` | correct | **`nan`** (also at a half turn, `real = 0`) |
| `Se3.exp` | correct | **`nan`** |
| `Se3.log` | correct | **`nan`** |
| `Se2.exp` | correct | **`nan`** |
| `So2.exp` / `So2.log` | correct | finite |

```python
import torch
from kornia.geometry.liegroup import So3, Se3, Se2
from kornia.geometry.quaternion import Quaternion

for dt in (torch.float64, torch.float32, torch.float16):
    v = torch.zeros(1, 3, dtype=dt, requires_grad=True)
    So3.exp(v).matrix().sum().backward()
    print(dt, "So3.exp", v.grad.flatten().tolist())
```

```
torch.float64 So3.exp [nan, nan, nan]
torch.float32 So3.exp [nan, nan, nan]
torch.float16 So3.exp [nan, nan, nan]
```

Same shape for the other four:

```python
q = torch.tensor([[1., 0., 0., 0.]], dtype=torch.float64, requires_grad=True)
So3(Quaternion(q)).log().sum().backward()            # [nan, nan, nan, nan]

q = torch.tensor([[0., 1., 0., 0.]], dtype=torch.float64, requires_grad=True)
So3(Quaternion(q)).log().sum().backward()            # [nan, nan, nan, nan]  (half turn)

v = torch.zeros(1, 6, dtype=torch.float64, requires_grad=True)
Se3.exp(v).matrix().sum().backward()                 # [nan] * 6

v = torch.zeros(1, 3, dtype=torch.float64, requires_grad=True)
Se2.exp(v).matrix().sum().backward()                 # [0., 0., nan]
```

## Mechanism

`torch.where` differentiates **both** branches and multiplies the unselected one by zero. Where that branch's
value is a `0/0` or its derivative is unbounded, `0 * nan` and `0 * inf` are `nan`, and the poison reaches the
input even though the value came from the safe branch. This is the same class as the merged #4338
(`solve_cubic`'s `D > 0` branch poisoning other rows) and #4233.

Per site:

- `kornia/geometry/liegroup/so3.py:132` — `b_large = torch.sin(theta_half) / theta` is `0/0` at `theta = 0`,
  and `b = torch.where(small_mask, b_small, b_large)` walks it anyway.
- `kornia/geometry/liegroup/so3.py:150-156` — three singularities in one expression: `sqrt` at `vec_sq = 0`,
  the division by `theta`, and `acos(real)` at `real = ±1`, all in the branch that is *not* selected at the
  identity; symmetrically, the small-angle branch `2 * vec / real` divides by zero at a half turn, which is the
  branch not selected there.
- `kornia/geometry/liegroup/se3.py:180-187` — `theta = dot(omega, omega).sqrt()` has an unbounded derivative at
  `omega = 0`, and `V` divides by `theta**2` and `theta**3`.
- `kornia/geometry/liegroup/se3.py:201` — `clamp_min(1e-12).sqrt()`. Per #4229, `clamp` is not a gradient guard
  on torch < 2.14 (it passes the incoming gradient straight through at its bound), and `1e-12` underflows to `0`
  in `float16` regardless, so `sqrt`'s boundary derivative is reached on every supported version.
- `kornia/geometry/liegroup/se2.py:191-192` — `so2.z.imag / theta` and `(1 - so2.z.real) / theta` are `0/0` at
  `theta = 0`.

## Why it was not caught

`test_gradcheck` is an empty `# TODO: implement me` stub in all three test classes —
`tests/geometry/liegroup/test_so3.py:50`, `tests/geometry/liegroup/test_se3.py:68`,
`tests/geometry/liegroup/test_se2.py:97`. Nothing in the suite differentiates these functions anywhere, let
alone at the identity.

## Suggested fix

The version-independent pattern already accepted in #4233, #4338 and #4228: substitute a safe argument into the
branch that gets differentiated, so the singular expression is never evaluated at its singular point. The
substituted value is discarded by the `where`, so every forward stays byte-identical.

I have this working locally for all five functions — forwards byte-identical across 64 random inputs per
function plus small-angle and sub-`eps` batches, `gradcheck` passing at generic points, and the identity
gradients agreeing with central differences (for `So3.log` at the identity, with the differences taken along the
unit sphere rather than through the ambient 4-space, where the perturbed quaternion leaves the manifold and the
other branch's formula takes over). Happy to open the PR, along with real `test_gradcheck` implementations for
the three stubs.

## Environment

kornia `main` @ `d6a2c13`, torch 2.7.0, python 3.11, Windows, CPU.
Related: #4229 (`clamp` is not a portable gradient guard), #4338 and #4233 (same `torch.where` poisoning class),
#4007 / #4228 (the same defect in `quaternion_exp_to_log`).

mohansree14:
## What

Fixes #4404. `So3.exp`, `So3.log`, `Se3.exp`, `Se3.log` and `Se2.exp` all returned `nan` gradients at the
identity — on **every dtype, `float64` included** — while their forward values were correct. `So3.log` was `nan`
at a half turn (`real = 0`) too. `So2` was already fine and is untouched.

The identity is the standard initialisation for pose optimisation, so this was the first backward pass of a
typical `So3.exp(...)` loop.

Before, on `main` @ `d6a2c13`, torch 2.7.0, CPU, `float64`:

| | forward | backward |
| --- | --- | --- |
| `So3.exp` @ `v = 0` | correct | `[nan, nan, nan]` |
| `So3.log` @ identity | correct | `[nan, nan, nan, nan]` |
| `So3.log` @ half turn | correct | `[nan, nan, nan, nan]` |
| `Se3.exp` @ `v = 0` | correct | `[nan] * 6` |
| `Se3.log` @ identity | correct | `[nan] * 7` |
| `Se2.exp` @ `v = 0` | correct | `[0., 0., nan]` |

After, same inputs: every one finite, at `float64`, `float32` and `float16`.

## Why it happens

`torch.where` differentiates **both** branches and multiplies the unselected one by zero. Where that branch holds
a `0/0` or an unbounded derivative, `0 * nan` and `0 * inf` are `nan`, and the poison reaches the input even
though the value came from the safe branch. Same class as the merged #4338 and #4233.

The five sites and what is singular in each are listed in #4404. Two are worth calling out here:

- **`So3.log` is singular at both ends, symmetrically.** At the identity the large-angle branch's `sqrt`,
  division by `theta` and `acos(±1)` all diverge — and that is the branch `where` does *not* select there. At a
  half turn the small-angle branch's `2 * vec / real` divides by zero, and that is the branch not selected
  *there*. Both needed guarding.
- **`Se3.log`'s `clamp_min(1e-12)` was never a gradient guard.** Per #4229, `clamp` passes the incoming gradient
  straight through at its bound on torch < 2.14, and `1e-12` underflows to `0` in `float16` on every version, so
  `sqrt`'s unbounded derivative at `0` was reached regardless. The floor is kept on the branch that is selected,
  so every `theta_sq > 0` stays byte-identical.

## The fix

The pattern already accepted in #4233, #4338 and #4228: substitute a safe argument into the branch that gets
differentiated, so the singular expression is never evaluated at its singular point. The `where` discards the
substituted value, so the forward cannot move.

## Verification

Local, torch 2.7.0, CPU. `pixi`, CUDA and MPS were not available to me.

- **Forwards are byte-identical**, not merely close: `torch.equal` against `main` over 64 random inputs per
  function (seeded), plus a 16-input small-angle batch (`1e-4`) and a 16-input batch below `So3.exp`'s own `eps`
  floor (`1e-7`), in `float64`. `So3.exp`, `So3.log`, `Se3.exp`, `Se3.log`, `Se2.exp` — all seven captured
  tensors equal.
- **Gradients away from the singular points are unchanged**: byte-identical for `So3.exp` and `So3.log`; `Se3.exp`
  differs by `8.9e-16` (~2 ULP in `float64`, from the reassociation `theta**2` → `safe_theta**2`).
- **`gradcheck` passes** at generic points for `So3.exp`, `So3.log`, `Se3.exp` and `Se2.exp` — these are the
  `test_gradcheck` stubs this PR fills in, all three of which were `# TODO: implement me` and are the reason this
  went unnoticed.
- **The identity gradients match central differences** in `float64` for `So3.exp` (`[0, 0, 0]`), `Se3.exp`
  (`[1, 1, 1, 0, 0, 0]`), `Se2.exp` (`[0, 0, 0]`) and `So3.log` at a half turn (`[-2, 0, π, π]`).
- `tests/geometry` at `float32` passes in full; `tests/geometry/liegroup` passes at `float32` and `float64`.
  `ruff check` and `ruff format --check` are clean.

### One value worth reviewing rather than taking on trust

`So3.log` at the identity comes out as `d(omega_x)/dq_x = 2`, and **central differences through the ambient
4-space disagree — they say `0`.** They are measuring a different function: a perturbed `(1, h, 0, 0)` is not a
unit quaternion, so it lands in the large-angle branch, where `acos(1) = 0` zeroes the result. Along the unit
sphere — `q(h) = (√(1-h²), h, 0, 0)`, the only path that stays a rotation — the difference quotient is
`2.000000017` at `h = 1e-4` and `2.000000083` at `h = 1e-5`, which is the value autograd now returns. It is also
exactly what the selected branch `2 * vec / real` differentiates to, unchanged by this PR; only the `nan` from
the *other* branch is gone. `tests/geometry/liegroup/test_so3.py` pins it with that reasoning written down.

The `float16` legs of `tests/geometry/liegroup` have 36 pre-existing failures on `main` (`ComplexHalf`,
`test_ortho` tolerances); this PR neither adds to them nor fixes them.


ducha-aiki (dismissed):
Reviewed exact head `0b280d32db5b531245b8b812966b676067721a64`.

The five singular-branch guards preserve the selected forward computations and make the previously unsafe inactive expressions finite at the intended identity/half-turn cases. Generic gradchecks cover normal paths. The full CPU float32/float64 liegroup suite passed 738 tests; focused identity/gradcheck coverage across all four CPU dtypes passed 19 tests with 1 bfloat16 Se2 skip. No introduced code-review blocker found. Hosted float16 CI was cancelled and the collector is red, so successful replacement checks are still required before merge.

Nonblocking follow-up: Se2.exp already collapses pure translations when theta==0 because its selected a coefficient is zero rather than the analytic limit one. The identity translation Jacobian is therefore zero. This behavior also exists on base and is not introduced here; it should be handled separately from these NaN guards.

ducha-aiki (dismissed):
Re-approving after pushing a merge of `main` into this branch. GitHub reported the branch as conflicting on `CHANGELOG.md` only, because this PR and #4447 both insert a new first bullet under `### Bug fixes`; the merge commit keeps both bullets and changes nothing else (every other path is tree-identical to git's own merge result). The previous approval on the pre-merge head is [here](https://github.com/kornia/kornia/pull/4405#pullrequestreview-5178443011); it was dismissed automatically by the push. The full matrix is running on the merge commit.

ducha-aiki (dismissed):
Re-approving on `767563f1`. That commit is a pure merge of `origin/main` (5be4852b) into the previously approved head, pushed to unblock the `CHANGELOG.md` adjacent-insertion conflict that yesterday's nine merges re-created: every path except `CHANGELOG.md` is tree-identical to `git merge-tree`, and the `CHANGELOG.md` lines added over main equal the branch's own added lines over its merge-base (no main line removed). Nothing about the reviewed change moved; CI is re-running on the merge commit.

ducha-aiki (dismissed):
Re-approving on `9f2caff2`. Merging #4460, #4461, #4462 and #4452 into `main` brought back the `CHANGELOG.md` adjacent-insertion conflict with the approved head `767563f1`. I pushed `9f2caff2`, a keep-both merge of `origin/main` (`dc0af1a4`). It is a pure merge: every path except `CHANGELOG.md` is tree-identical to `git merge-tree`, and the non-blank lines it adds to `CHANGELOG.md` are exactly this branch's own lines. Nothing else changed since the approval. It merges once the matrix on this commit is green.

PR #4406, opened 2026-09-10
docs(agents): clamp is not a portable gradient guard

### LINKED ISSUE #4229: torch.clamp's boundary gradient changed between 2.9.1 and 2.14.0, so kornia's clamp-as-gradient-guard pattern silently fails on the older half of its supported range

mohansree14:
## What

Closes #4229 by writing its finding into `AGENTS.md`'s precision section — item 4 of that issue's suggested
scope, and the last one still open.

The other items already landed, by other contributors:

- item 2, `kornia/feature/matching.py` — #4233
- item 3, both `polynomial_solver.py` sites — #4338 and #4339
- item 1 is the principle this bullet records; #4007's instance is #4228

## The bullet

The mechanism is easy to rediscover the expensive way, because a CPU float32 run on a recent torch shows nothing:

> `torch.clamp` is not a portable gradient guard: its derivative **at the bound** passes the incoming gradient
> through (`1.0`) on torch 2.5.1 and 2.9.1 and returns `0.0` on 2.14.0. Wherever the bound is also the singular
> point — `clamp(min=0).sqrt()`, `clamp(-1, 1).acos()`, `clamp(-1, 1).asin()` — the clamp bounds the value only,
> so the unbounded derivative still reaches the backward pass as `inf` or `nan` on the older half of the
> supported range while 2.14 masks it. […] the 2.5.1 leg is what discriminates.

It also carries the half-precision corollary, which #4229 does not spell out and which I hit while auditing the
"positive floor — safe on every version" list: a floor below the dtype's smallest subnormal is not a guard
either, because it underflows to zero and the clamp becomes a no-op. `clamp(min=1e-12)` is exactly `clamp(min=0)`
in `float16` (fp16's smallest subnormal is ~6e-8).

And it names the fix that does not depend on the torch version — substitute a safe argument into the expression
that gets differentiated, take the value from a `.detach()`ed copy or the other arm of a `torch.where` — which is
what #4233, #4338, #4339 and #4228 all converged on independently.

## The grep gate

Per `AGENTS.md`, a change that closes a documented issue has to leave `grep -rnE "#4229|issues/4229" kornia/
tests/` clean of anything that reads as open work. Three hits survive:

- `kornia/enhance/jpeg.py:242` and `kornia/geometry/solvers/polynomial_solver.py:319` — both are retrospective in
  substance (they explain why the code does *not* use clamp), so following the form accepted in #4184 I reworded
  each to say it points at a measurement rather than at a tracking issue, rather than deleting a citation that is
  still useful to a reader.
- `tests/geometry/solvers/test_polynomial_solver.py:459` — a pin, which the guide says goes on naming its issue.

No `CHANGELOG.md` entry: this is contributor guidance, with no user-visible behaviour change. Happy to add one if
you would rather it were recorded there too.

## Verification

Documentation and comments only — no executable change. `ruff check` and `ruff format --check` clean on both
touched modules, both still import, and `tests/geometry/solvers/test_polynomial_solver.py -k 4229` passes (3
passed). The two `clamp` claims in the bullet are the ones measured in #4229's table; I reproduced the pre-2.14
column locally on torch 2.7.0 (`clamp(x, min=0.0)` grad at `x = 0` is `1.0`, `clamp(min=0).sqrt()` grad at `0` is
`inf`), and the `float16` underflow claim directly (`torch.tensor(0.0, dtype=torch.float16).clamp(min=1e-12)` is
`0.0`).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -99,6 +99,15 @@ when more than one virtualenv is in play: name the interpreter explicitly rather
 - MPS does not support float64 gradcheck. MPS autocast can also change the effective dtype; inspect nearby tests before changing tolerances or skips.
 - The blocking MPS job runs `--device=mps --dtype=float32 --xfail-known-failures`. Its exact strict-xfail baseline is `testing/known_failure_xfails/mps_float32.txt` and is tracked in #4159. A fix removes its manifest line; a rename or reparametrization updates the node ID; a new failure is fixed or explicitly documented before being added. The manifest contract requires a full-suite run, without `-k` or a partial path.
 - TF32 matmul is disabled by default; `--tf32` enables it. cuDNN convolutions still use PyTorch's TF32 default. Tests marked `tf32` are xfailed at collection unless `--tf32` is passed, and that marker is non-strict, so a default run reports neither their failure nor their recovery.
+- `torch.clamp` is not a portable gradient guard: its derivative **at the bound** passes the incoming gradient
+  through (`1.0`) on torch 2.5.1 and 2.9.1 and returns `0.0` on 2.14.0. Wherever the bound is also the singular
+  point -- `clamp(min=0).sqrt()`, `clamp(-1, 1).acos()`, `clamp(-1, 1).asin()` -- the clamp bounds the value only,
+  so the unbounded derivative still reaches the backward pass as `inf` or `nan` on the older half of the supported
+  range while 2.14 masks it. A floor below the dtype's smallest subnormal is no guard either: `clamp(min=1e-12)`
+  underflows to `0` in `float16`. Substitute a safe argument into the expression that gets differentiated and take
+  the value from a `.detach()`ed copy, or from the other arm of a `torch.where`, instead; that is
+  version-independent by construction. A CPU float32 run on a recent torch does not exercise any of this -- the
+  2.5.1 leg is what discriminates. The measurements and the audit of the pattern are in #4229.
 - Preserve device and dtype rather than creating implicit CPU or default-dtype tensors. Use the injected `device` and `dtype` fixtures in tests.
 
 ## Library preferences
--- a/kornia/enhance/jpeg.py
+++ b/kornia/enhance/jpeg.py
@@ -239,7 +239,8 @@ def _jpeg_quality_to_scale(
     # for a documented input. libjpeg gives quality 0 the quality-1 table, so give it the quality-1 scale.
     # The guard is exactly the zero point: a fractional quality in (0, 1) is already finite, keeps its own
     # (larger) scale, and is not touched. ``torch.where`` rather than ``clamp`` also keeps the gradient at
-    # every unguarded quality independent of the torch version (see #4229).
+    # every unguarded quality independent of the torch version -- clamp's derivative at its own bound is not
+    # portable across the supported torch range, as measured in #4229.
     strength: torch.Tensor = torch.where(
         compression_strength == 0.0, torch.ones_like(compression_strength), compression_strength
     )
--- a/kornia/enhance/jpeg.py
+++ b/kornia/enhance/jpeg.py
@@ -240,7 +240,7 @@ def _jpeg_quality_to_scale(
     # The guard is exactly the zero point: a fractional quality in (0, 1) is already finite, keeps its own
     # (larger) scale, and is not touched. ``torch.where`` rather than ``clamp`` also keeps the gradient at
     # every unguarded quality independent of the torch version -- clamp's derivative at its own bound is not
-    # portable across the supported torch range, as measured in #4229.
+    # portable across the supported torch range, as measured in PR #4406.
     strength: torch.Tensor = torch.where(
         compression_strength == 0.0, torch.ones_like(compression_strength), compression_strength
     )
--- a/kornia/geometry/solvers/polynomial_solver.py
+++ b/kornia/geometry/solvers/polynomial_solver.py
@@ -316,7 +316,7 @@ def solve_quartic(coeffs: torch.Tensor) -> torch.Tensor:
     R_sq = torch.gather(R_sq_candidates, -1, best_idx).squeeze(-1)
 
     # `clamp(min=0).sqrt()` does not guard the gradient: d(sqrt)/dx is unbounded at 0, and on
-    # torch < 2.14 clamp passes the incoming gradient straight through at the bound (#4229), so
+    # torch < 2.14 clamp passes the incoming gradient straight through at the bound, as measured in #4229, so
     # R_sq == 0 -- a biquadratic such as x^4 - 16 -- gave inf and then nan. Substitute a safe
     # radicand instead, as solve_quadratic above already does, so sqrt is never differentiated at 0.
     # On torch >= 2.14 clamp already zeroes the boundary gradient, so the pins for this guard pass
--- a/kornia/geometry/solvers/polynomial_solver.py
+++ b/kornia/geometry/solvers/polynomial_solver.py
@@ -316,7 +316,7 @@ def solve_quartic(coeffs: torch.Tensor) -> torch.Tensor:
     R_sq = torch.gather(R_sq_candidates, -1, best_idx).squeeze(-1)
 
     # `clamp(min=0).sqrt()` does not guard the gradient: d(sqrt)/dx is unbounded at 0, and on
-    # torch < 2.14 clamp passes the incoming gradient straight through at the bound, as measured in #4229, so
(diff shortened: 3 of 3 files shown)


ducha-aiki (changes requested):
Reviewed exact head `82bb1fc37746db0a333fe5a7fffbad174f91d6cd`.

[P2] Reconcile this closing PR with the documented-issue gate in AGENTS.md. The body says Closes #4229, but the prescribed issue-number grep still finds it at kornia/enhance/jpeg.py:243 and kornia/geometry/solvers/polynomial_solver.py:319. The gate explicitly says a surviving kornia/ hit makes a closing change incomplete; it does not currently exempt retrospective citations. The reworded explanations are accurate and useful, but move their measurement citation to contributor documentation or otherwise remove these source hits, or use Refs #4229 and keep the issue open. This is a repository-policy blocker, not a correctness objection to the guidance. The focused #4229 solver selection passed 3 tests; hosted checks are green.

mohansree14:
Thanks — agreed, that's a policy blocker and the gate reads it correctly.

Pushed `72d4901`. Both surviving `kornia/` hits now cite the delivering PR instead of the closing issue:

- `kornia/enhance/jpeg.py:243` — `as measured in #4229` → `as measured in PR #4406`
- `kornia/geometry/solvers/polynomial_solver.py:319` — same substitution

The explanations are otherwise unchanged, and the measurement is still reachable from the source. The full audit stays in the AGENTS.md bullet (outside the `kornia/`/`tests/` grep scope), and the pin in `tests/geometry/solvers/test_polynomial_solver.py:459` keeps its `#4229` name, which the gate allows.

`grep -rnE "#4229|issues/4229" kornia/` now returns nothing, so `Closes #4229` is clean. `ruff check` and `ruff format --check` pass on both files.

mohansree14:
@ducha-aiki this is ready for another look when you have a moment — I can't re-request review from a fork. Thanks!

ducha-aiki (approved):
Re-review of exact head `72d49014fee376b7a38f55fe0ec5e94913c5d475`.

The prior closing-issue blocker is fixed: both production comments now cite PR #4406, and the prescribed `#4229|issues/4229` search has no hits under `kornia/`. The remaining issue references are appropriately in contributor guidance and its regression pin. The guidance accurately captures the supported-version gradient behavior. The focused solver selection passed on CPU float32/float64 (6 cases) and MPS float32 (3 cases), and all 40 hosted checks are green. I found no blocking issues.

PR #4447, opened 2026-09-11
fix(augmentation): make the RandAugment and AutoAugment range guards say what they enforce

### LINKED ISSUE #4443: W1b: `RandAugment` rejects both endpoints its own message advertises for `m`, and never validates `n`
`RandAugment`'s docstring says "m: magnitude for all the augmentations, ranged from [0, 30]" and its guard raises
``ValueError: Expect `m` in [0, 30]. Got 0.`` — a message that names a closed interval and then rejects both of
its endpoints. The guard at `rand_augment.py:81` is `if m <= 0 or m >= 30`, so the accepted values are `1..29`. A user
who reads either the docstring or the error message and writes `m=30` for maximum strength gets an exception whose
text says `30` is in range.

`n` has the opposite problem: it is validated nowhere. `n=0` constructs and is a silent no-op, returning the input
untouched with no operation applied; `n=20` constructs and is silently clamped to the policy length, because
`randperm` samples without replacement from the 15-entry default policy. Neither is reported.

The magnitude bin has the same shape one level down. `AutoAugment`'s policy tuples take an integer bin index, which
`linspace(lo, hi, 11)[m:m+2]` maps to a range, so the valid bins are `0..9`; bin `10` produces a raw `IndexError`
naming an internal tensor rather than a message about the policy the caller wrote.

## Repro

```python
import torch
from kornia.augmentation.auto.rand_augment.rand_augment import default_policy


torch.manual_seed(0)
x = torch.rand(2, 3, 16, 16)

for m in (-1, 0, 1, 15, 29, 30, 31):
    try:
        AA.RandAugment(n=2, m=m)
        print(f"m={m:3d}  OK")
    except Exception as e:
        print(f"m={m:3d}  {type(e).__name__}: {e}")

print("default_policy length:", len(default_policy))
for n in (0, 1, 2, 5, 15, 20):
    torch.manual_seed(0)
    aug = AA.RandAugment(n=n, m=10)
    out = aug(x)
    print(f"n={n:3d}  constructed OK, ops applied per forward call = {len(aug._params)}"
          f"  output_is_input = {bool((out == x).all())}")

try:
    AA.AutoAugment(policy=[[("rotate", 1.0, 10)]])(x)
except Exception as e:
    print("AutoAugment magnitude bin 10 ->", type(e).__name__ + ":", e)
```

## Observed

```
m= -1  ValueError: Expect `m` in [0, 30]. Got -1.
m=  0  ValueError: Expect `m` in [0, 30]. Got 0.
m=  1  OK
m= 15  OK
m= 29  OK
m= 30  ValueError: Expect `m` in [0, 30]. Got 30.
m= 31  ValueError: Expect `m` in [0, 30]. Got 31.
default_policy length: 15
n=  0  constructed OK, ops applied per forward call = 0  output_is_input = True
n=  1  constructed OK, ops applied per forward call = 1  output_is_input = False
n=  2  constructed OK, ops applied per forward call = 2  output_is_input = False
n=  5  constructed OK, ops applied per forward call = 5  output_is_input = False
n= 15  constructed OK, ops applied per forward call = 15  output_is_input = False
n= 20  constructed OK, ops applied per forward call = 15  output_is_input = False
AutoAugment magnitude bin 10 -> IndexError: index 11 is out of bounds for dimension 0 with size 11
```

## Expected

For `m`, the message and the docstring agree with the guard. The guard is exclusive, so both should read `(0, 30)`
— that is a documentation and message fix and changes no behavior. Widening the guard to accept `0` and `30`
instead would be a behavior change: `m=0` would become a legal no-op and `m=30` a legal maximum, which is a
reasonable contract but a different one.

For `n`, a validation message. `n <= 0` is a caller mistake — a `RandAugment` that applies nothing is never what
was meant — and `n > len(policy)` is a request the sampler cannot honour; both should say so rather than being
absorbed. For `AutoAugment`'s magnitude bin, the same: an out-of-range bin should name the policy entry and the
`0..9` range instead of surfacing an index error from `linspace`.

None of these change a currently successful call: every value that works today keeps working.

## Classification

**Semantic contract** (the `m` message) and **API completeness** (the missing `n` and bin guards) — a documented
range that the guard contradicts, beside two arguments with no guard at all. Idiom: 3 for `m` (a guard whose
predicate does not match the documented contract) and 5 for `n` (a documented precondition never validated).

## Disposition

`n/a — single symbol`. The `m` message and docstring live in
`kornia/augmentation/auto/rand_augment/rand_augment.py`, and the `n` guard belongs beside them; the `AutoAugment`
bin guard is the same two-line shape one file over. A focused fix is welcome as a PR. Batch 6 documents the
accepted `1..29` range and the unvalidated `n` on the class and pins them, and does not change them.

## Environment

change it.

adimalkar:
Fixes #4443.

Reproduced the issue's script verbatim on `d6a2c138` (the commit it cites), then again after the change. Every value that constructed before still constructs.

## `m`

The guard is `m <= 0 or m >= 30`; the docstring and the message both named `[0, 30]`. Both now read `(0, 30)`. The guard itself is untouched, so this is a documentation and message fix with no behaviour change — the direction the issue prefers. Widening the guard to admit `0` and `30` is the other reasonable contract, and I have not taken it; say the word if you would rather have that and I will swap them.

## `n`

Now checked against the policy actually in use, so a caller-supplied two-entry policy reports `[1, 2]` rather than the default's `[1, 15]`.

```
n= 0  ValueError: Expect `n` in [1, 15], the number of sub-policies. Got 0.
n=20  ValueError: Expect `n` in [1, 15], the number of sub-policies. Got 20.
n= 1..15  unchanged
```

## `AutoAugment`'s magnitude bin

Two findings here changed how I implemented it, and both are worth your eye.

**A guard at the dispatch point would have broken working calls.** `auto_contrast`, `invert` and `equalize` take the magnitude as `_` and never index the scale. `("equalize", 1.0, 10)` works today. Gating every op in `compose_subpolicy_sequential` would reject it, which the issue's own "none of these change a currently successful call" rules out. So the check lives where the magnitude is consumed: one `_magnitude_bin` helper in `autoaugment/ops.py`, used by the eleven ops that index a scale. The three that ignore it stay ungated, and there is a test pinning that.

**Negative bins were silently wrong, which the issue does not mention.** Python indexing wraps, so on `main`:

```
rotate bin  -1  -> magnitude_range (30.0, -30.0), reversed; fails later inside
                   RandomRotation with "degrees[0] should be smaller than degrees[1]"
rotate bin  -2  -> constructs fine, silently a different range
rotate bin -11  -> constructs fine, silently a different range
rotate bin -12  -> IndexError
```

A wrong augmentation range with no error is worse than the reported `IndexError`. `0 <= bin <= last` catches all of it in one rule:

```
AutoAugment bin  10 -> ValueError: Expect magnitude bin for `rotate` in [0, 9]. Got 10.
AutoAugment bin  -1 -> ValueError: Expect magnitude bin for `rotate` in [0, 9]. Got -1.
```

The bound is derived as `magnitudes.shape[0] - 2` rather than hardcoded to 9, so an op with a different scale length reports its own bound.

## Tests

Seven new tests, 19 cases, in `TestRandAugment` and `TestAutoAugment`:

- `m` accepted at 1/15/29 and rejected at -1/0/30/31, matching on the new message
- `n` rejected at 0, -1 and `len(policy) + 1`, and bounded by a supplied policy
- magnitude bin rejected at 10, 11, -1, -2, naming the op
- bin endpoints 0 and 9 accepted
- the three magnitude-ignoring ops still accept bin 10

**12 of the 19 fail on `main` and pass here.** The other 7 pass both ways by design — they are the guards proving nothing that worked before stopped working.

## Verification

`tests/augmentation/` gives **160 failed, 2174 passed** both with this change and on clean `main` — byte-identical counts. Those 160 are a pre-existing `OSError: pytest: reading from stdin while output is captured` in `_test_sequential` and friends, unrelated to this change; I stashed and re-ran to confirm rather than assume. `test_auto_operation.py` alone goes from `3 failed, 21 passed` to `3 failed, 40 passed`, the same 3.

`ruff check`, `ruff format --check` and `mypy` clean on all changed files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -379,6 +379,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `RandAugment`'s `m` guard is exclusive at both ends, but its docstring and its error message
+  both named the closed interval `[0, 30]`, so a user who asked for the maximum strength the
+  message advertised got an exception saying `30` was in range. Both now read `(0, 30)`; the
+  accepted values are unchanged. `n` was validated nowhere: `n=0` constructed a `RandAugment`
+  that applied nothing, and `n` above the policy length was silently clamped, because the
+  sampler draws without replacement. It is now checked against the policy. `AutoAugment`'s
+  magnitude bin indexes two adjacent points of an 11-point scale, so `9` is the last usable
+  bin; a larger one raised a raw `IndexError` naming an internal tensor, and a negative one
+  wrapped silently onto a reversed range. Out-of-range bins now name the operation and the
+  valid range. Operations that ignore the magnitude entirely keep accepting any bin. (#4443)
+
 * `pixel2cam` now validates the full `Bx4x4` shape of `intrinsics_inv`, rejecting invalid matrix sizes
   before they cause unrelated transformation errors or return the wrong number of coordinate components. (#4381)
 * `Boxes.to_mask` leaves list-padding channels empty, including after coordinate transforms,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -388,7 +388,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   magnitude bin indexes two adjacent points of an 11-point scale, so `9` is the last usable
   bin; a larger one raised a raw `IndexError` naming an internal tensor, and a negative one
   wrapped silently onto a reversed range. Out-of-range bins now name the operation and the
-  valid range. Operations that ignore the magnitude entirely keep accepting any bin. (#4443)
+  valid range. Operations that ignore the magnitude entirely keep accepting any bin. (#4447)
 
 * `pixel2cam` now validates the full `Bx4x4` shape of `intrinsics_inv`, rejecting invalid matrix sizes
   before they cause unrelated transformation errors or return the wrong number of coordinate components. (#4381)
--- a/kornia/augmentation/auto/autoaugment/ops.py
+++ b/kornia/augmentation/auto/autoaugment/ops.py
@@ -17,6 +17,8 @@
 
 """AutoAugment operation wrapper."""
 
+from typing import Tuple
+
 import torch
 
 from kornia.augmentation.auto.operations import (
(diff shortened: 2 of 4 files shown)


ducha-aiki (changes requested):
Reviewed exact head `9085bc242b508b0b29aa92904c3e5eb832165902`.

[P2] Change the new changelog entry at CHANGELOG.md:390 from the issue citation (#4443) to the delivering PR (#4447). The issue can remain separately. The repository requires each user-visible Unreleased entry to cite its PR. The range guards and tests otherwise look correct, including the custom-policy n bound and magnitude-ignoring operations. The focused auto-operation file passed 43 CPU float32 cases; hosted checks are green.

adimalkar:
Done in `da9638a` — the entry now cites #4447. #4443 stays as the issue.

That is the only change; the source and tests are untouched from the head you reviewed (`9085bc2`), so the `43` CPU float32 cases in `test_auto_operation.py` and the rest of the diff are as you saw them.

ducha-aiki (approved):
Re-review of exact head `0468921fd4ff9461debeaa13a8a9bddd6e8b16f9`.

The requested changelog correction is present: the Unreleased bug-fix entry now cites the delivering PR, #4447. The range guards and tests remain correct, including the caller-supplied policy bound for `n`, rejection of negative magnitude bins, and preservation of operations that ignore magnitude. Code review is clear.

The complete `test_auto_operation.py` file passed 60 CPU float32/float64 cases in an isolated checkout with its import origin verified. All 40 hosted checks are green. I did not run the whole suite, CUDA, or MPS locally.

PR #4449, opened 2026-09-11
test: make Sequential test_forward deterministic

### LINKED ISSUE #4446: test: `TestSequential::test_forward` is unseeded, so its float16 failing set depends on collection order
`tests/augmentation/container/test_sequential_aug.py::TestSequential::test_forward` draws its inputs and its `random_apply` selection from torch's global generator without seeding, and its float16 tolerance is tight enough that whether a given parametrization passes depends on the draw. Which IDs fail therefore depends on what ran before it in the same process: the file alone and the full `tests/augmentation` directory fail different members of the same parametrized family. The CPU half-precision manifest isolates each node's RNG, so CI is stable; any unseeded run (`pixi run test-half`, a directory run, a `-k` run) is not. The batch-6 convention pins added to the same directory consume the global generator and therefore moved this test's failing set until they were wrapped in an autouse fixture that restores the RNG state after each pin; that fixture is a workaround, not the fix.

## Repro

```bash
cd <checkout> && .venv/bin/python -m pytest tests/augmentation/container/test_sequential_aug.py --dtype=float16 -q -p no:cacheprovider -rf --color=no | grep -E "^FAILED|failed"
# versus the same test inside a full-directory run:
.venv/bin/python -m pytest tests/augmentation --dtype=all -q -p no:cacheprovider -rf --color=no | grep "test_sequential_aug.py::TestSequential::test_forward"
```

## Observed

```
file alone (float16):
FAILED tests/augmentation/container/test_sequential_aug.py::TestSequential::test_forward[cpu-float16-1]
FAILED tests/augmentation/container/test_sequential_aug.py::TestSequential::test_forward[cpu-float16-10]
FAILED tests/augmentation/container/test_sequential_aug.py::TestSequential::test_forward[cpu-float16-True]
FAILED tests/augmentation/container/test_sequential_aug.py::TestSequential::test_forward[cpu-float16-False]
4 failed, 69 passed

full tests/augmentation directory (--dtype=all), the same test:
FAILED tests/augmentation/container/test_sequential_aug.py::TestSequential::test_forward[cpu-float16-10]
FAILED tests/augmentation/container/test_sequential_aug.py::TestSequential::test_forward[cpu-float16-False]
FAILED tests/augmentation/container/test_sequential_aug.py::TestSequential::test_forward[cpu-float16-True]
FAILED tests/augmentation/container/test_sequential_aug.py::TestSequential::test_forward[cpu-float16-random_apply1]
```

## Expected

The test seeds its own draw (`torch.manual_seed` inside the body, or a seeded generator for its inputs) so that its float16 outcome is a property of the test and not of the collection order, and the `cpu-float16` known-failure manifest entries for it are then either stable or removed with a tolerance fix.

## Classification

Test determinism — an unseeded random-input test whose half-precision result depends on the RNG state left by earlier tests. Idiom: none (test hygiene, not library behavior).

## Disposition

`n/a — single symbol`. Focused fix, welcome as a PR: seed the test, re-derive its manifest lines.

## Environment

kornia 0.9.0rc1 @ `d6a2c138` (main), torch 2.14.0, python 3.11, cpu. Found while adding the batch-6 convention pins (tracking issue #4407).

AyushSrivastava1818:

Make `TestSequential::test_forward` deterministic by isolating and seeding PyTorch's RNG with `torch.random.fork_rng()` and `torch.manual_seed(0)`. This prevents its CPU float16 results from depending on test collection order.

The float16 manifest was also updated to reflect the stable deterministic xfail set.

Fixes #4446.

## What did you check?

```text
$ pixi run test-module tests/augmentation/container/test_sequential_aug.py --dtype=float16
2 passed, 5 xfailed

$ pixi run test-module tests/augmentation --dtype=all
```

The target `test_forward` produced the exact same results in both isolated and full-directory runs:

5 xfailed, 2 passed.

The broader directory run also reported unrelated pre-existing failures in `tests/augmentation/container/`; none were from `TestSequential::test_forward`.

## How did AI help?

AI was used to inspect the reported issue, identify the RNG-state dependency, and suggest a minimal test-only fix. The resulting changes and manifest were manually reviewed and verified by running the test in both isolated and full-directory collection contexts to confirm identical results.


ducha-aiki (changes requested):
Reviewed exact head `7f13e59c937fa56e2b6d634ec1da011f04fd7d20`.

[P1] The new float16 manifest entries at testing/half_precision_xfails/cpu_float16.txt:31 and :34 do not match the required Linux/PyTorch 2.9.1 full-suite result. The hosted verifier reports strict XPASS for both TestSequential.test_forward[cpu-float16-1] and TestSequential.test_forward[cpu-float16-random_apply2], leaving the half job and collector red. Remove/regenerate those entries using the required full cpu-float16 profile and rerun. The deterministic test change itself looks reasonable. Local float32 file coverage passed 72 tests; the Darwin/PyTorch 2.14 half profile had 67 passes and 5 xfails, which cannot establish the Linux baseline. The separate red macOS job reports an RRDBNet ONNX timeout.

AyushSrivastava1818:
@ducha-aiki Updated the CPU float16 manifest as requested and pushed the follow-up commit.

The Linux/PyTorch 2.9.1 float16 check is now passing. The two stale `test_forward` xfail entries were removed, while the remaining expected failures are preserved.

The only remaining CI failure is the unrelated macOS/Python 3.12/PyTorch 2.5.1 `RRDBNetBuilder` ONNX export timeout.

ducha-aiki (dismissed):
Re-review of `d12b914e480a09aaa6a31bd744f7e829e916931a`.

The two stale float16 manifest entries from the preceding review are removed, and both required Linux/PyTorch 2.9.1 half-precision verifier jobs now pass on this head. The deterministic test change and revised manifest are code-clear. The affected file passed 81 CPU float32/float64 cases locally with import origin verified in the isolated review checkout.

Merge remains gated by the macOS/Python 3.12/PyTorch 2.5.1 RRDBNetBuilder ONNX export timeout and the failed collector. Those need successful replacement checks; this approval does not waive CI. No fresh full-suite, CUDA or MPS run was performed locally.

ducha-aiki (approved):
Re-approving on `8f91573f`. The branch had fallen 115 commits behind and conflicted with `main` in `tests/augmentation/container/test_sequential_aug.py`: #4452 appended `TestConventionImageSequential` right after `test_forward`, which this branch re-indents under `fork_rng`. I pushed `8f91573f`, a merge of `origin/main` (`dc0af1a4`) that keeps both sides. It takes main's file with this branch's `test_forward` body; main had not touched `test_forward` since the merge-base. The merged tree's diff against `main` equals, line for line, the branch's diff against its merge-base: the seeded `test_forward` plus the one `cpu_float16.txt` manifest line (`random_apply3` → `True`). Nothing else changed.

I could not settle one thing locally. On torch 2.14 the seeded float16 failing set is `{1, (1, 2), 10, True, False}` on both the approved head and the merge, which differs from the manifest. The blocking half-precision legs run torch 2.9.1, where this head's float16 leg was green, so the merge-commit run is the check that counts. The seed makes the set deterministic per torch version, not across versions.

PR #4451, opened 2026-09-11
fix(augmentation): enforce strict RandomRain drop bounds

### LINKED ISSUE #4434: W1b: `RandomRain`'s drop-size guard admits `drop_height == H` under a "less than" message
`RandomRain` validates its drawn drop size at `kornia/augmentation/_2d/intensity/random_rain.py:75-86`. Both guards compare with `<=` while their messages say "less than": the height check admits `drop_height == H` under the message "Height of drop should be greater than zero and less than image height", and the width check admits `drop_width == W` under "Width of drop should be less than image width". The drawing code needs the strict inequality — the start row is scaled by `H - h - 1`, which is `-1` when `h == H` — so the value the guard lets through is exactly the value that then indexes past the end of the tensor. A caller who asks for a drop as tall as the image gets an `IndexError` naming an internal dimension instead of the guard's own message, which is one off from being correct.

That would be a narrow edge case if the default did not walk into it. `drop_height=(5, 20)` draws integers from 5 through 19, so on any image with `H <= 19` some seeds draw a drop taller than the image (the guard fires, with a message the caller did not cause by passing anything) and some draw one exactly as tall (the guard passes and the indexing fails). Both outcomes are a crash on a valid image with default arguments, on a fraction of seeds that depends on the batch, the channel count and the input, while the failure mode does not. At `H = 20` and above the default is clean.

The axis assignment was executed and is correct — this is not an axis-order slip. `drop_height` is checked and drawn against `shape[2]` and `drop_width` against `shape[3]`, verified by running each argument against both orientations of the same non-square image, and the drop itself is drawn height-along-rows and width-along-columns, verified by printing the lit pixel coordinates. An earlier reading of the sweep, that `16x32` fails while `32x16` never does because the two axes are swapped, is refuted by exactly that: `32x16` is clean because its height is 32, which is above the top of the default draw.

## Repro

```python
import collections, torch, kornia.augmentation as K

def rain(H, W, dh, dw, fixture="zeros"):
    torch.manual_seed(0)
    x = torch.zeros(1, 1, H, W) if fixture == "zeros" else torch.rand(2, 3, H, W)
    return K.RandomRain(number_of_drops=(1, 1), drop_height=(dh, dh), drop_width=(dw, dw), p=1.0)(x)

# 1. the guard's predicate against its own message, on both axes
for h in (11, 12, 13):
    try:
        rain(12, 30, h, 1); print(f"drop_height={h} on H=12: ok")
    except Exception as e:                                          # noqa: BLE001
        print(f"drop_height={h} on H=12: {type(e).__name__}: {e}")
for w in (29, 30, 31):
    try:
        rain(12, 30, 1, w); print(f"drop_width={w}  on W=30: ok")
    except Exception as e:                                          # noqa: BLE001
        print(f"drop_width={w}  on W=30: {type(e).__name__}: {e}")

# 2. the axis assignment, executed: which image dimension bounds which argument,
#    and along which axis is the drop actually drawn?
print()
for dh, dw in ((20, 1), (1, 20)):
    for H, W in ((12, 30), (30, 12)):
        try:
            rain(H, W, dh, dw); print(f"drop_height={dh} drop_width={dw} on {H}x{W}: ok")
        except Exception as e:                                      # noqa: BLE001
            print(f"drop_height={dh} drop_width={dw} on {H}x{W}: {type(e).__name__}: {str(e)[:42]}...")
for dh, dw in ((6, 1), (1, 6)):
    out = rain(12, 30, dh, dw)
    print(f"drop_height={dh} drop_width={dw} on 12x30: lit (row,col) = {(out[0, 0] > 0).nonzero().tolist()}")

# 3. what the default range draws, and how often it breaks a short image
print()
torch.manual_seed(0)
d = torch.cat([K.RandomRain(p=1.0).forward_parameters((2, 3, 64, 64))["drop_height_factor"] for _ in range(200)])
print("default drop_height=(5, 20) draws integers in", (int(d.min()), int(d.max())))
for (H, W), fixture in (((16, 32), "rand"), ((32, 16), "rand"), ((16, 32), "zeros"),
                        ((32, 16), "zeros"), ((19, 19), "rand"), ((20, 20), "rand")):
    c = collections.Counter()
    for seed in range(100):
        torch.manual_seed(seed)
        x = torch.rand(2, 3, H, W) if fixture == "rand" else torch.zeros(1, 1, H, W)
        try:
            K.RandomRain(p=1.0)(x); c["ok"] += 1
        except Exception as e:                                      # noqa: BLE001
            c[type(e).__name__] += 1
    shape = "torch.rand(2,3,H,W)" if fixture == "rand" else "torch.zeros(1,1,H,W)"
    print(f"default RandomRain on {H}x{W}, {shape}, seeds 0-99: {dict(c)}")
```

## Observed

```
drop_height=11 on H=12: ok
drop_height=12 on H=12: IndexError: index 12 is out of bounds for dimension 0 with size 12
drop_height=13 on H=12: BaseError: Height of drop should be greater than zero and less than image height.
drop_width=29  on W=30: ok
drop_width=30  on W=30: IndexError: index 30 is out of bounds for dimension 1 with size 30
drop_width=31  on W=30: BaseError: Width of drop should be less than image width.

drop_height=20 drop_width=1 on 12x30: BaseError: Height of drop should be greater than zero...
drop_height=20 drop_width=1 on 30x12: ok
drop_height=1 drop_width=20 on 12x30: ok
drop_height=1 drop_width=20 on 30x12: BaseError: Width of drop should be less than image wi...
drop_height=6 drop_width=1 on 12x30: lit (row,col) = [[0, 8], [1, 8], [2, 8], [3, 8], [4, 8], [6, 9]]
drop_height=1 drop_width=6 on 12x30: lit (row,col) = [[1, 7], [1, 8], [1, 9], [1, 10], [1, 11], [2, 13]]

default drop_height=(5, 20) draws integers in (5, 19)
default RandomRain on 16x32, torch.rand(2,3,H,W), seeds 0-99: {'ok': 56, 'BaseError': 32, 'IndexError': 12}
default RandomRain on 32x16, torch.rand(2,3,H,W), seeds 0-99: {'ok': 100}
default RandomRain on 16x32, torch.zeros(1,1,H,W), seeds 0-99: {'IndexError': 7, 'ok': 80, 'BaseError': 13}
default RandomRain on 32x16, torch.zeros(1,1,H,W), seeds 0-99: {'ok': 100}
default RandomRain on 19x19, torch.rand(2,3,H,W), seeds 0-99: {'ok': 82, 'IndexError': 18}
default RandomRain on 20x20, torch.rand(2,3,H,W), seeds 0-99: {'ok': 100}
```

## Expected

`<` in both guards at `random_rain.py:75-86`, so that the predicate matches the message it raises and the rejected value is the one the drawing code cannot handle. That is a focused fix and welcome as a PR.

The default is a second decision. Leaving `drop_height=(5, 20)` as it is means `RandomRain()` keeps failing on any image shorter than 20 pixels, with a clearer message but still a hard failure the caller cannot avoid without overriding the argument. The two coherent options are to clamp the drawn drop size to the image (a behavior change for callers who currently get an exception) or to keep raising and document the minimum image height in the `Args` block, which today says only "height of the drop in image(same for each drops in one image)". The audit recommends the guard fix now and the documented minimum with it; changing what the default draws is a behavior change.

## Classification

**Broken output** — a guard whose predicate is one value wider than its own message, letting the exact value through that the code below it cannot index. Idiom: 3 (a guard whose predicate does not match the documented contract). Explicitly **not** idiom 7: the axis assignment was executed and is correct.

## Disposition

`n/a — single symbol`. Batch 6 documents the behavior above as it is and pins it. The `<=` to `<` change is a focused fix, welcome as a PR; altering the default drop-height range is a behavior change and belongs in the coordinated Repair & Deprecation window.

## Environment

Jeevang1-epic:

Fixes the `RandomRain` drop-size boundary validation reported in #4434.

The existing guards accepted the boundary values:

- `drop_height == H`
- `abs(drop_width) == W`

even though the corresponding error messages describe strict bounds.

Those equality cases could therefore pass validation and fail later during drawing with an internal `IndexError`.

This PR changes the validation so that the implementation matches the documented contract:

- `drop_height < H`
- `abs(drop_width) < W`

The default `RandomRain` sampling behavior remains unchanged.

Regression coverage was added for:

- `drop_height == H`
- `drop_width == W`
- `drop_width == -W`
- a non-square image case to verify height/width orientation
- immediately in-bounds values that must continue to succeed

Fixes #4434

## What did you check?

Focused `RandomRain` tests:

```text
13 passed
```

Also verified:

- the branch is based directly on the current `upstream/main`
- the worktree is clean
- `git diff --check` passes
- there are no conflict markers
- only the intended `RandomRain` implementation and regression-test files are changed

## How did AI help?

AI tooling assisted with repository analysis, reproducing the boundary behavior, regression-test planning, implementation review, and validation guidance.

I reviewed the resulting changes and ran the focused regression checks locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -378,6 +378,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `RandomRain` now rejects drop heights equal to the image height and absolute drop widths equal to the image
+  width with the documented validation error, instead of allowing boundary-sized drops to reach an internal
+  `IndexError`. (#4451)
 * `pixel2cam` now validates the full `Bx4x4` shape of `intrinsics_inv`, rejecting invalid matrix sizes
   before they cause unrelated transformation errors or return the wrong number of coordinate components. (#4381)
 * `Boxes.to_mask` leaves list-padding channels empty, including after coordinate transforms,
--- a/kornia/augmentation/_2d/intensity/random_rain.py
+++ b/kornia/augmentation/_2d/intensity/random_rain.py
@@ -32,8 +32,10 @@ class RandomRain(IntensityAugmentationBase2D):
     Args:
         p: probability of applying the transformation.
         number_of_drops: number of drops per image
-        drop_height: height of the drop in image(same for each drops in one image)
-        drop_width: width of the drop in image(same for each drops in one image)
+        drop_height: Height of the drop in the image (same for each drop in one image). Sampled values must be
+            greater than zero and strictly smaller than the input image height.
+        drop_width: Width of the drop in the image (same for each drop in one image). The absolute sampled value
+            must be strictly smaller than the input image width.
     Shape:
         - Input: :math:`(C, H, W)` or :math:`(B, C, H, W)`
         - Output: :math:`(B, C, H, W)`
--- a/kornia/augmentation/_2d/intensity/random_rain.py
+++ b/kornia/augmentation/_2d/intensity/random_rain.py
@@ -74,14 +74,13 @@ class RandomRain(IntensityAugmentationBase2D):
         KORNIA_CHECK(image.shape[1] in {3, 1}, "Number of color channels should be 1 or 3.")
         KORNIA_CHECK(
             bool(
-                torch.all(params["drop_height_factor"] <= image.shape[2])
-                and torch.all(params["drop_height_factor"] > 0)
+                torch.all(params["drop_height_factor"] < image.shape[2]) and torch.all(params["drop_height_factor"] > 0)
             ),
             "Height of drop should be greater than zero and less than image height.",
         )
 
         KORNIA_CHECK(
-            bool(torch.all(torch.abs(params["drop_width_factor"]) <= image.shape[3])),
+            bool(torch.all(torch.abs(params["drop_width_factor"]) < image.shape[3])),
             "Width of drop should be less than image width.",
         )
         modeified_img = image.clone()
--- a/tests/augmentation/test_augmentation.py
+++ b/tests/augmentation/test_augmentation.py
@@ -5359,6 +5359,37 @@ class TestRandomRain(BaseTester):
 
             assert err_msg in str(errinfo)
 
+    @pytest.mark.parametrize(
+        "drop_height,drop_width,error_message",
+        [
+            (
+                (5, 5),
+                (1, 1),
+                "Height of drop should be greater than zero and less than image height.",
+            ),
+            ((1, 1), (7, 7), "Width of drop should be less than image width."),
+            ((1, 1), (-7, -7), "Width of drop should be less than image width."),
+        ],
+    )
+    def test_drop_size_boundaries(self, drop_height, drop_width, error_message, device, dtype):
+        from kornia.core.exceptions import BaseError
+
+        input_data = torch.zeros(1, 3, 5, 7, device=device, dtype=dtype)
+        aug = RandomRain(p=1.0, drop_height=drop_height, drop_width=drop_width, number_of_drops=(1, 1))
+
+        with pytest.raises(BaseError) as errinfo:
+            aug(input_data)
+
+        assert str(errinfo.value) == error_message
+
+    def test_drop_size_immediately_inside_boundaries(self, device, dtype):
+        input_data = torch.zeros(1, 3, 5, 7, device=device, dtype=dtype)
+        aug = RandomRain(p=1.0, drop_height=(4, 4), drop_width=(6, 6), number_of_drops=(1, 1))
+
+        output_data = aug(input_data)
+
+        assert output_data.shape == input_data.shape
+
     def test_zero_probability(self, device):
         input_data = torch.rand(10, 3, 8, 8, device=device)
         aug = RandomRain(p=0.0, drop_height=(2, 3), drop_width=(2, 3), number_of_drops=(1, 3))


ducha-aiki (changes requested):
Reviewed exact head `96304c3ed896b17de40a587ebbf273967d0b1ee2`.

[P2] Add an Unreleased Bug fixes entry citing #4451. Boundary-sized RandomRain drops now raise the documented validation error instead of an internal IndexError, so this is a user-visible correction covered by the repository changelog requirement. The strict bounds and regression coverage otherwise look correct, including negative-width drops. Exact-head TestRandomRain passed 25 CPU float32/float64 cases; hosted checks are green.

ducha-aiki (dismissed):
Re-review of exact head `13cd4a423545dd01edc9f4facf85c86bc74837d4`.

The requested Unreleased bug-fix entry is now present and cites #4451. The strict height and absolute-width bounds match the indexing performed by `RandomRain`, the immediately-inside cases remain accepted, and the regression covers positive and negative boundary widths. Code review is clear.

Focused validation at this head passed 51 CPU cases across float16, bfloat16, float32, and float64. Imports were verified inside the isolated checkout. The hosted float16 full-suite job reached GitHub's six-hour limit without reporting a test failure, so the collector failed as a consequence; a successful replacement run is still needed before merge. I did not run the whole suite, CUDA, or MPS locally.

Jeevang1-epic:
Thanks for the re-review and approval.

I see the PR is now blocked by a CHANGELOG.md conflict after main moved. Since the current exact head is already approved and updating the branch would create a new head and invalidate the existing review, would you prefer that I refresh the branch and resolve the changelog conflict, or should I leave the approved head untouched for you to handle?

Happy to follow whichever approach you prefer.

ducha-aiki (dismissed):
Re-approving after pushing a merge of `main` into this branch. GitHub reported the branch as conflicting on `CHANGELOG.md` only, because this PR and #4447 both insert a new first bullet under `### Bug fixes`; the merge commit keeps both bullets and changes nothing else (every other path is tree-identical to git's own merge result). The previous approval on the pre-merge head is [here](https://github.com/kornia/kornia/pull/4451#pullrequestreview-5184245014); it was dismissed automatically by the push. The full matrix is running on the merge commit.

Issue #4407, opened 2026-09-11
Per-op convention blocks: kornia.augmentation (batch 6)


ducha-aiki:
Tracking issue for conventions batch 6: per-class `Convention:` blocks and pinning tests for the public `kornia.augmentation` surface — the 84 exported classes plus the `auto` and `container` subpackages. Same method as batch 3 (#3941), batch 4 (#4020) and batch 5 (#4262): document actual behavior including warts, pin every precise claim with a `test_convention_*` / `test_wart_*` test, change no runtime behavior, and route every behavior finding to its own issue. This batch also carries the two consumer-contract items the plan assigned to it: the serialization / resume contract (pickle, `deepcopy`, `state_dict` per class) and the randomness contract (what seeds what, the draw device and dtype, `same_on_batch`, `DataLoader` workers, RNG consumption order), stated once on the Conventions & Pitfalls page and pinned.

The whole surface was audited up front before any PR opened, at `d6a2c138` with torch 2.14.0, on cpu and on mps: a mechanical sweep over all 84 classes (rank / `keepdim`, dtype policy, draw device, `p` / `p_batch` / `same_on_batch`, RNG consumption, `params=` replay, `transform_matrix` / `inverse`, serialization, `torch.compile` / `torch.jit.script`), a library-wide rotation-direction sweep, and four section audits — 257 executed fact rows over 89 symbols, 173 `OK`, 18 docstring defects, 37 behavior defects, 24 unenforced accidents, from about 6,200 probe-output lines. Each section table was reviewed and re-reviewed against its raw output line by line; every issue below carries a `## Repro` re-executed against `d6a2c138` and a verbatim `## Observed` block.

## Sub-batches (one PR each)

| PR | scope | symbols |
|---|---|---|
[...]

Issue #4061, opened 2026-08-28
W1b: `Boxes` conversion silently corrupts sub-unit extents — normalized-coordinate boxes do not round-trip


ducha-aiki:
`Boxes` stores extents inclusively and materializes vertices with `_boxes_to_polygons`, which places the top-right corner at `xmin + width - 1` (`boxes.py:88-92`). That `- 1` is only meaningful for integer pixel extents of at least one unit. For `0 < extent < 1` the "top-right" vertex lands **left** of "top-left" and the "bottom-right" vertex lands **above** "top-right": the stored quadrilateral is inverted on both axes and is no longer in the documented clockwise TL/TR/BR/BL order. `to_tensor` then re-derives the box with `amin`/`amax`, so it recovers a different — and larger — box than the caller supplied.

The affected modes are exactly the three that convert: `'xyxy'`, `'xywh'` and `'vertices'`. The `'_plus'` modes are unaffected because their `+ 1` cancels the `- 1`.

This is not an exotic input. It is *every* box in normalized `[0, 1]` coordinates, and it reaches the public augmentation API: `AugmentationSequential(..., data_keys=["bbox_xywh"])` corrupts such boxes through an otherwise identity pipeline. `validate_boxes=True` does not catch it, because `0.5 > 0` passes the only check there is.

## Repro

```python
import torch
import kornia.augmentation as K
from kornia.geometry.boxes import Boxes

cases = {
    "xyxy":          torch.tensor([[0.1, 0.1, 0.6, 0.9]]),
    "xywh":          torch.tensor([[0.1, 0.1, 0.5, 0.8]]),
    "vertices":      torch.tensor([[[0.1, 0.1], [0.6, 0.1], [0.6, 0.9], [0.1, 0.9]]]),
    "vertices_plus": torch.tensor([[[0.1, 0.1], [0.6, 0.1], [0.6, 0.9], [0.1, 0.9]]]),
}
for mode, t in cases.items():
    out = Boxes.from_tensor(t, mode=mode, validate_boxes=True).to_tensor(mode)
    print(f"{mode:14s} {[round(v, 2) for v in t.flatten().tolist()]} -> "
          f"{[round(v, 2) for v in out.flatten().tolist()]}")

print("stored:", Boxes.from_tensor(cases["xyxy"], mode="xyxy").data.round(decimals=2).tolist())

# end to end, through an identity pipeline
img = torch.rand(1, 3, 8, 8)
[...]

PR #4454, opened 2026-09-11
Fix: (augmentation)RandomMosaic returns (W, H) on a non-square input, and the same axis swap reaches start_ratio_range

### LINKED ISSUE #4438: W1b: `RandomMosaic` returns `(W, H)` on a non-square input, and the same axis swap reaches `start_ratio_range`
`RandomMosaic` composes a mosaic from a batch and, with the default `output_size=None`, is documented to keep the
input size. It does keep the number of pixels, but it transposes them: a `(B, C, 6, 8)` batch comes back
`(B, C, 8, 6)`. The same swap reaches three further surfaces that all trace to one root cause — the unbatched
`(C, H, W)` path, the `start_ratio_range` slots, and the output shape itself, which changes with the draw because
the swap only happens when the augmentation fires.

`kornia/augmentation/random_generator/_2d/mosaic.py:101-106` hands `input_sizes[0]`, the height, to
`bbox_generator` as its `width` argument and `input_sizes[1]`, the width, as its `height`, so the source crop is
`H` wide and `W` tall; `crop_dst` two lines below uses the correct order. Line 99 of the same file computes
`start_corner_x = start_corner_factor[:, 0] * batch_shape[-2]`, scaling the slot it calls `x` by the height, so
`start_ratio_range`'s first entry moves the box down rather than across. An explicit `output_size` is honoured
correctly as `(H, W)`, which is why the swap is invisible to anyone who passes one.

The class doctest uses a `224x224` input and a `(300, 300)` output size, and every `RandomMosaic` fixture in
`tests/augmentation/test_augmentation_mix.py` and `tests/augmentation/container/test_augmentation_sequential.py`
is square too — `8x8`, `100x100`, `224x224` with `output_size=(300, 300)`. Nothing in the repository looks at a
non-square mosaic, so the swap has never had a chance to fail a check. A caller who feeds one gets a transposed
image and, at `p < 1.0`, a module whose output shape is not a function of its input shape.

## Repro

```python
import torch


for (h, w) in ((6, 8), (8, 6), (5, 7), (4, 4)):
    torch.manual_seed(0)
    out = K.RandomMosaic(p=1.0)(torch.rand(4, 3, h, w))
    print(f"input (H, W) = ({h}, {w})  ->  output {tuple(out.shape)}")

# an explicit output_size IS honoured as (H, W)
torch.manual_seed(0)
print("output_size=(6, 10) on 5x7   ->  output", tuple(K.RandomMosaic(p=1.0, output_size=(6, 10))(torch.rand(4, 3, 5, 7)).shape))

# the shape of one module's output depends on whether the augmentation fires
for p in (0.0, 1.0):
    torch.manual_seed(0)
    print(f"p={p}  ->  output", tuple(K.RandomMosaic(p=p)(torch.rand(4, 3, 6, 8)).shape))

# the unbatched (C, H, W) path carries the same swap
for keepdim in (False, True):
    torch.manual_seed(0)
    print(f"(C, H, W) = (1, 6, 8), keepdim={keepdim}  ->  RandomMosaic  ",
          tuple(K.RandomMosaic(p=1.0, keepdim=keepdim)(torch.rand(1, 6, 8)).shape),
          " RandomMixUpV2 ", tuple(K.RandomMixUpV2(p=1.0, keepdim=keepdim)(torch.rand(1, 6, 8)).shape))

# the source box handed to the crop: 5 wide x 7 tall on a 6x8 input
torch.manual_seed(0)
aug = K.RandomMosaic(p=1.0)
aug(torch.rand(4, 3, 6, 8))
src, dst = aug._params["src"][0], aug._params["dst"][0]
print("src[0] =", [[round(v, 4) for v in pt] for pt in src.tolist()])
print("dst[0] =", [[round(v, 4) for v in pt] for pt in dst.tolist()])

# start_ratio_range: the slot the code calls x is scaled by H
for (h, w) in ((6, 8), (3, 11)):
    for r in (0.0, 0.5, 1.0):
        torch.manual_seed(0)
        a = K.RandomMosaic(p=1.0, start_ratio_range=(r, r))
        a(torch.rand(4, 3, h, w))
        print(f"HxW={h}x{w}  start_ratio_range=({r}, {r})  ->  src[0] top-left =",
              [round(v, 4) for v in a._params["src"][0, 0].tolist()])
```

## Observed

```
input (H, W) = (6, 8)  ->  output (4, 3, 8, 6)
input (H, W) = (8, 6)  ->  output (4, 3, 6, 8)
input (H, W) = (5, 7)  ->  output (4, 3, 7, 5)
input (H, W) = (4, 4)  ->  output (4, 3, 4, 4)
output_size=(6, 10) on 5x7   ->  output (4, 3, 6, 10)
p=0.0  ->  output (4, 3, 6, 8)
p=1.0  ->  output (4, 3, 8, 6)
(C, H, W) = (1, 6, 8), keepdim=False  ->  RandomMosaic   (1, 1, 8, 6)  RandomMixUpV2  (1, 1, 6, 8)
(C, H, W) = (1, 6, 8), keepdim=True  ->  RandomMosaic   (1, 8, 6)  RandomMixUpV2  (1, 6, 8)
src[0] = [[3.4557, 5.2725], [8.4557, 5.2725], [8.4557, 12.2725], [3.4557, 12.2725]]
dst[0] = [[0.0, 0.0], [7.0, 0.0], [7.0, 5.0], [0.0, 5.0]]
HxW=6x8  start_ratio_range=(0.0, 0.0)  ->  src[0] top-left = [0.0, 0.0]
HxW=6x8  start_ratio_range=(0.5, 0.5)  ->  src[0] top-left = [3.0, 4.0]
HxW=6x8  start_ratio_range=(1.0, 1.0)  ->  src[0] top-left = [6.0, 8.0]
HxW=3x11  start_ratio_range=(0.0, 0.0)  ->  src[0] top-left = [0.0, 0.0]
HxW=3x11  start_ratio_range=(0.5, 0.5)  ->  src[0] top-left = [1.5, 5.5]
HxW=3x11  start_ratio_range=(1.0, 1.0)  ->  src[0] top-left = [3.0, 11.0]
```

## Expected

`output_size=None` returns `(B, C, H, W)` for a `(B, C, H, W)` input — the size the docstring already promises —
and `start_ratio_range`'s first slot scales the horizontal start by `W` and the second the vertical start by `H`,
matching their names and matching `bbox_generator`'s `(x_start, y_start, width, height)` signature. The fix is the
argument order at `:101-106` plus the two factors at `:99-100`; the commented-out
`output_size = input_sizes if self.output_size is None else self.output_size` at `:82` is the shape half of the
same line of thought.

This changes the output of every existing non-square caller, including anyone who has compensated for the swap by
passing a transposed `output_size`, so it is a behavior change rather than a silent correction.

## Classification

**Broken output** — a transposed image and a draw-dependent output shape on any non-square input, with no error.
Idiom: 7 (an axis-order slip hidden by symmetric test data) — the class doctest is square, so the defect has never
been visible to an example.

## Disposition

`n/a — single symbol`. All four surfaces are `RandomMosaic` and its generator, fixed by one change to
`kornia/augmentation/random_generator/_2d/mosaic.py`; a focused fix is welcome as a PR. Batch 6 documents the
current `(W, H)` default on the class and pins it, and does not change it.

## Environment

change it.

msparsh13:

Fixes `RandomMosaic` behavior for non-square inputs and preserves padded bounding boxes when the augmentation is not applied.

`RandomMosaic` was swapping the height and width when `output_size=None`. The mosaic generator was passing `H` as the width and `W` as the height, causing a `(B, C, H, W)` input to produce `(B, C, W, H)`. The same ordering issue also affected the source crop and `start_ratio_range`. This change uses the correct `(W, H)` ordering in the mosaic generator.

The change also fixes bbox passthrough when `p=0`. Even when the augmentation was skipped, `bbox_xyxy` tensors were converted to `Boxes`. For padded boxes represented as `[0, 0, 0, 0]`, that conversion produces `[-1, -1, 1, 1]`. The no-op path now returns the original bbox tensor before conversion, preserving the padding.

Fixes  #4438

## What did you check?

* Added regression coverage for non-square `RandomMosaic` inputs.
* Checked that `output_size=None` preserves `(H, W)`.
* Checked `start_ratio_range` uses width for the horizontal component and height for the vertical component.
* Checked unbatched `(C, H, W)` inputs.
* Checked `bbox_xyxy` passthrough with `p=0`, including padded `[0, 0, 0, 0]` boxes.
* Ran the augmentation test suite.

```text
$ uv run pytest -q ./tests/augmentation
```

## How did AI help?

Used AI to help investigate the failing regression and trace the bbox conversion path. The implementation, root cause, and proposed changes were checked manually, and the tests were run locally.

_Please let me know if I’ve misunderstood any of the expected conventions or if anything in the implementation should be adjusted._

### DIFF
--- a/kornia/augmentation/random_generator/_2d/mosaic.py
+++ b/kornia/augmentation/random_generator/_2d/mosaic.py
@@ -96,13 +96,13 @@ class MosaicGenerator(RandomGeneratorBase):
         start_corner_factor = _adapted_rsampling(
             (batch_size, 2), self.start_ratio_range_sampler, same_on_batch=False
         ).to(device=_device, dtype=_dtype)
-        start_corner_x = start_corner_factor[:, 0] * batch_shape[-2]
-        start_corner_y = start_corner_factor[:, 1] * batch_shape[-1]
+        start_corner_x = start_corner_factor[:, 0] * batch_shape[-1] ##  batch has shape is [B, C, H, W]
+        start_corner_y = start_corner_factor[:, 1] * batch_shape[-2]
         crop_src = bbox_generator(
             start_corner_x,
             start_corner_y,
-            start_corner_x.clone().fill_(input_sizes[0]),
-            start_corner_y.clone().fill_(input_sizes[1]),
+            start_corner_x.clone().fill_(input_sizes[1]),
+            start_corner_y.clone().fill_(input_sizes[0]),
         )
         crop_dst = torch.tensor(
             [[[0, 0], [input_sizes[1] - 1, 0], [input_sizes[1] - 1, input_sizes[0] - 1], [0, input_sizes[0] - 1]]],
--- a/kornia/augmentation/random_generator/_2d/mosaic.py
+++ b/kornia/augmentation/random_generator/_2d/mosaic.py
@@ -96,7 +96,7 @@ class MosaicGenerator(RandomGeneratorBase):
         start_corner_factor = _adapted_rsampling(
             (batch_size, 2), self.start_ratio_range_sampler, same_on_batch=False
         ).to(device=_device, dtype=_dtype)
-        start_corner_x = start_corner_factor[:, 0] * batch_shape[-1] ##  batch has shape is [B, C, H, W]
+        start_corner_x = start_corner_factor[:, 0] * batch_shape[-1]  ##  batch has shape is [B, C, H, W]
         start_corner_y = start_corner_factor[:, 1] * batch_shape[-2]
         crop_src = bbox_generator(
             start_corner_x,
--- a/tests/augmentation/test_augmentation_mix.py
+++ b/tests/augmentation/test_augmentation_mix.py
@@ -403,6 +403,53 @@ class TestRandomMosaic(BaseTester):
         # batch dimension must be preserved regardless of how many samples are selected.
         assert out_image.shape[0] == input.shape[0]
 
+    def test_p0_passthrough(self, device, dtype):
+        torch.manual_seed(76)
+
+        f = RandomMosaic(p=0.0, data_keys=["input", "bbox_xyxy"])
+
+        input = torch.randn((4, 3, 32, 32), device=device, dtype=dtype)
+        boxes = torch.tensor(
+            [
+                [[2.0, 3.0, 10.0, 12.0], [0.0, 0.0, 0.0, 0.0]],
+                [[4.0, 5.0, 14.0, 15.0], [1.0, 2.0, 3.0, 4.0]],
+                [[6.0, 7.0, 16.0, 17.0], [0.0, 0.0, 0.0, 0.0]],
+                [[8.0, 9.0, 18.0, 19.0], [2.0, 3.0, 4.0, 5.0]],
+            ],
+            device=device,
+            dtype=dtype,
+        )
+
+        input_copy = input.clone()
+        boxes_copy = boxes.clone()
+
+        out_image, out_boxes = f(input, boxes)
+
+        self.assert_close(out_image, input_copy)
+        self.assert_close(out_boxes, boxes_copy)
+
+    def test_partial_batch_passthrough(self, device, dtype):
+        torch.manual_seed(76)
+
+        f = RandomMosaic(
+            output_size=(32, 32),
+            p=0.5,
+            data_keys=["input", "bbox_xyxy"],
+        )
+
+        input = torch.randn((12, 3, 32, 32), device=device, dtype=dtype)
+        boxes = torch.zeros((12, 2, 4), device=device, dtype=dtype)
+
+        output, _ = f(input, boxes)
+
+        assert output.shape == input.shape
+
+        to_apply = f._params["batch_prob"] > 0.5
+        untouched = ~to_apply
+
+        if untouched.any():
+            self.assert_close(output[untouched], input[untouched])
+
 
 class TestRandomJigsaw(BaseTester):
     def test_smoke(self, device, dtype):
--- a/tests/augmentation/test_augmentation_mix.py
+++ b/tests/augmentation/test_augmentation_mix.py
@@ -430,26 +430,64 @@ class TestRandomMosaic(BaseTester):
 
     def test_partial_batch_passthrough(self, device, dtype):
         torch.manual_seed(76)
-
+    
         f = RandomMosaic(
-            output_size=(32, 32),
             p=0.5,
             data_keys=["input", "bbox_xyxy"],
         )
-
-        input = torch.randn((12, 3, 32, 32), device=device, dtype=dtype)
+    
+        input = torch.randn((12, 3, 6, 8), device=device, dtype=dtype)
         boxes = torch.zeros((12, 2, 4), device=device, dtype=dtype)
-
+    
         output, _ = f(input, boxes)
+    
+        assert output.shape == input.shape
+    
+        to_apply = f._params["batch_prob"] > 0.5
+        untouched = ~to_apply
+    
+        if untouched.any():
+            self.assert_close(output[untouched], input[untouched])
+    
+    def test_non_square_default_output_size(self, device, dtype):
+        input = torch.randn((4, 3, 6, 8), device=device, dtype=dtype)
+
+        f = RandomMosaic(p=1.0)
+
+        output = f(input)
 
         assert output.shape == input.shape
 
-        to_apply = f._params["batch_prob"] > 0.5
-        untouched = ~to_apply
+    def test_start_ratio_range_uses_width_for_x_and_height_for_y(self, device, dtype):
+        input = torch.randn((4, 3, 6, 8), device=device, dtype=dtype)
 
-        if untouched.any():
-            self.assert_close(output[untouched], input[untouched])
+        f = RandomMosaic(
+            p=1.0,
+            start_ratio_range=(0.5, 0.5),
+        )
 
+        f(input)
+
+        top_left = f._params["src"][0, 0]
+
+        expected = torch.tensor([4.0, 3.0], device=device, dtype=dtype)
+
+        self.assert_close(top_left, expected)
+
+    @pytest.mark.parametrize("keepdim", [False, True])
+    def test_unbatched_non_square(self, keepdim, device, dtype):
+        input = torch.randn((1, 6, 8), device=device, dtype=dtype)
+
+        f = RandomMosaic(
+            p=1.0,
+            keepdim=keepdim,
+        )
+
+        output = f(input)
+
+        expected_shape = (1, 6, 8) if keepdim else (1, 1, 6, 8)
+
+        assert output.shape == expected_shape
 
 class TestRandomJigsaw(BaseTester):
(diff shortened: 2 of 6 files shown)


tests/augmentation/test_augmentation_mix.py
```diff
@@ -403,6 +403,53 @@ def test_p(self, p, device, dtype):
...
+                [[8.0, 9.0, 18.0, 19.0], [2.0, 3.0, 4.0, 5.0]],
+            ],
+            device=device,
+            dtype=dtype,
+        )
+
+        input_copy = input.clone()
+        boxes_copy = boxes.clone()
+
+        out_image, out_boxes = f(input, boxes)
```
ducha-aiki, line 426: [P1] Account for the bfloat16 failure in both new tests

Both `test_p0_passthrough[cpu-bfloat16]` and `test_partial_batch_passthrough[cpu-bfloat16]` fail before reaching the new behavior: `MixAugmentationBaseV2.forward` calls `DType.get(in_tensor.dtype)`, whose enum has no BFLOAT16 entry. This is reproduced locally and is the pair of new failures in the required hosted bfloat16 verifier. Please resolve this supported-test-profile mismatch (or deliberately record the existing unsupported path through the repository's required known-failure process), and obtain a passing verifier. Leaving these newly parametrized cases unaccounted for blocks the suite.

kornia/augmentation/random_generator/_2d/mosaic.py
```diff
@@ -96,13 +96,13 @@ def forward(self, batch_shape: Tuple[int, ...], same_on_batch: bool = False) ->
...
-        start_corner_y = start_corner_factor[:, 1] * batch_shape[-1]
+        start_corner_x = start_corner_factor[:, 0] * batch_shape[-1]  ##  batch has shape is [B, C, H, W]
+        start_corner_y = start_corner_factor[:, 1] * batch_shape[-2]
         crop_src = bbox_generator(
             start_corner_x,
             start_corner_y,
-            start_corner_x.clone().fill_(input_sizes[0]),
-            start_corner_y.clone().fill_(input_sizes[1]),
+            start_corner_x.clone().fill_(input_sizes[1]),
+            start_corner_y.clone().fill_(input_sizes[0]),
```
ducha-aiki, line 105: [P2] Commit non-square regression coverage for the axis correction

The PR description says it adds non-square, default-output-size, start-ratio and unbatched coverage, but the only two added tests both use 32x32 inputs; the partial-batch test also sets a square output size. Those tests do not exercise this correction and would not catch reverting all four generator changes. Please add asymmetric image/coordinate cases that assert the default (H,W) output, x scaled by W and y by H, and the unbatched keepdim behavior described in #4438.

ducha-aiki (changes requested):
The generator correction works on the non-square examples I exercised: 6x8 and 3x11 inputs preserve their dimensions, and a fixed 0.5 start ratio yields top-left coordinates (4,3) and (5.5,1.5). The current diff still needs the regression coverage described in the PR and a resolution for its newly failing bfloat16 tests, as detailed inline.

Please also add the required Unreleased CHANGELOG.md entry citing #4454, explaining the old swapped dimensions and the new behavior. The description's `Fixes #4407` would close the entire augmentation-conventions tracking issue; this focused fix should reference that umbrella and close #4438 only.

Validation: 13 selected RandomMosaic CPU float32/float64 cases passed in the isolated PR checkout, with import origin verified. The two new RandomMosaic cases reproduce `KeyError: 'BFLOAT16'` locally and in the required hosted bfloat16 job (103414849761), which leaves the collector red. No full-suite, CUDA or MPS run was performed locally.

tests/augmentation/test_augmentation_mix.py
msparsh13: @ducha-aiki I checked the failure on main and confirmed it reproduces there as well. It looks like the direct fix would be adding BFLOAT16 to DType (the to_torch() branch already exists), but since DType is shared across the library, I’m not sure if you’d want that included in this PR or handled separately.

Would you prefer me to add the BFLOAT16 support here, or keep it separate?

ducha-aiki (changes requested):
Re-review of exact head `c8454ddf`.

Status of the prior asks:

- Asymmetric regression coverage: resolved. The 6x8 default-output, start-ratio `(4, 3)` and unbatched `keepdim` cases are there and all 12 new nodes fail on main.
- Closing scope: resolved, `Fixes #4438` only.
- CHANGELOG: the entry is now under `## Unreleased` / `### Bug fixes` but still cites no `(#4454)`. Not resolved as asked.
- bfloat16 failures: not resolved, and now 6 nodes are red on the cpu-bfloat16 profile instead of 2; 5 of them die at `DType.get` with `KeyError: 'BFLOAT16'`.

New findings on this head:

- [P1] `test_start_ratio_range_uses_width_for_x_and_height_for_y` compares the generator's `_params["src"]`, which the generator always emits in float32, against a tensor built from the `dtype` fixture. It fails on float64 (`The values for attribute 'dtype' do not match: torch.float32 != torch.float64`), float16, bfloat16 and MPS; that is every float64 leg in the hosted matrix. Compare in the generator's dtype.
- [P1] `pyproject.toml` gains an unrelated `[dependency-groups] dev = ["ruff>=0.16.7"]` block and `uv.lock` is regenerated (83 lines), dropping the `sys_platform != 'darwin'` markers on the nvidia-* dependencies. Neither belongs in this fix; please revert both.
- [P2] The `mix/base.py` `bbox_xyxy` p=0 passthrough is outside #4438, guards one of the three box keys (`bbox_xywh` still returns `[-1, -1, 2, 2]`), and does nothing for partial batches, where the untouched rows still come back as `[0, 0, 1, 1]`; `test_partial_batch_passthrough` discards the boxes output so it cannot see this. The root cause is #4061. Split it out as its own complete fix if you want to pursue it.
- [P3] The `##  batch has shape is [B, C, H, W]` comment in `mosaic.py`.

On your question about adding `BFLOAT16` to `DType`: yes, it is the right fix, but not in this PR. `DType.get(torch.bfloat16)` raises `KeyError` on main and `to_torch` has no bfloat16 branch either (it raises `ValueError`), so it is a four-line enum change with no other consumers. However `testing/half_precision_xfails/cpu_bfloat16.txt` carries 27 strict-xfail lines for mix tests that fail with exactly that `KeyError`; with the enum in place 21 of them pass, which fails the strict verifier, and 6 flip to real numerical failures (MixUp output dtype, CutMix and Mosaic values). That change needs its own PR with the manifest edit, merged first.

The practical situation: #4459 lands the byte-identical `mosaic.py` fix with tests that are green on float32/float64, both half profiles and MPS, and all its hosted checks pass. I have asked its author only to move the release-note heading. Unless you object, the cleanest path is to close this PR as superseded by #4459 and reopen the bbox passthrough as a separate, complete fix referencing #4061.

Validation (isolated worktree, import origin verified): CPU float32 mosaic selection passes; the new start-ratio test fails on CPU float64 as described; cpu-bfloat16 profile 6 red nodes; the 12 new nodes fail on `8fb9ad12`. Not run: full suite, CUDA.

Issue #4204, opened 2026-09-03
RANSAC tests abort the pytest process on MPS on GitHub-hosted Apple-silicon runners


ducha-aiki:
### Summary

Running `tests/geometry/test_ransac.py` against the `mps` device on a GitHub-hosted Apple-silicon
runner **aborts the pytest process** (`Fatal Python error: Aborted`, exit 134). It is not an
exception, so nothing can catch, xfail or pin it: the whole job dies wherever it happens to land.
This is the reason #4202 skips that file on MPS at collection time.

### What was observed

Three full-suite runs on GitHub-hosted runners, torch 2.14.0, Python 3.13, `KORNIA_TEST_DEVICE=mps`,
float32, one pytest process:

| run | selection | image | outcome |
|---|---|---|---|
| A | `--deselect tests/geometry/test_ransac.py` | `macos-15-arm64` | **completed**, 14 failed / 8280 passed, 10m28s |
| B | `--deselect …::TestRANSACEssential` | `macos-15-arm64` | **aborted**, in `F.pad` (`test_affine.py::test_downsizeAA`), *after* `test_ransac.py` finished |
| C | none | `macos-26-arm64` | **aborted**, in `_multiply_deg_one_poly` (`test_ransac.py::TestRANSACEssential::test_forward_enforces_essential_constraint`) |

Run C's traceback:

```
Fatal Python error: Aborted
  File "kornia/geometry/epipolar/essential.py", line 87 in _multiply_deg_one_poly
  File "kornia/geometry/epipolar/essential.py", line 254 in _null_to_Nister_solution_script
  File "kornia/geometry/epipolar/essential.py", line 406 in null_to_Nister_solution
  File "kornia/geometry/epipolar/essential.py", line 75 in run_5point
  File "kornia/geometry/epipolar/essential.py", line 791 in find_essential
  File "kornia/geometry/ransac.py", line 215 in estimate_model_from_minsample
  File "kornia/geometry/ransac.py", line 379 in forward
```

Run B's, for contrast — a plain pad, in a test that has nothing to do with RANSAC:

```
Fatal Python error: Aborted
  File "torch/nn/functional.py", line 5849 in pad
  File "kornia/filters/filter.py", line 138 in filter2d
  File "kornia/filters/filter.py", line 205 in filter2d_separable
  File "kornia/filters/gaussian.py", line 116 in gaussian_blur2d
  File "kornia/geometry/transform/affwarp.py", line 665 in resize
[...]

PR #4455, opened 2026-09-11
fix(core): route MPS SVD and svdvals through CPU at 8192 elements or more

### LINKED ISSUE #4201: MPS: torch 2.14 batched linalg.svd/svdvals/lstsq fail at >= 8192 input elements, blocking RANSAC on Apple silicon

vimalyad:
torch 2.14's MPS `linalg.svd` and `linalg.svdvals` fail to build their Metal
pipeline state object once a batched input holds **8192 elements or more**,
regardless of the per-matrix shape. On Apple silicon that made
`find_fundamental`, `find_homography_dlt(..., solver="svd")` and `solve_pnp_dlt`
raise `Failed to created pipeline state object` at realistic batch sizes.

**Change.** `_torch_svd_cast` and `_torch_linalg_svdvals` now detect an MPS input
at or above that ceiling and decompose on the CPU, moving results back to the
input's device. This is the approach suggested in the issue thread; chunking was
ruled out there. Below the ceiling the code path is unchanged, and because
`.cpu()` and `.to(device)` are ordinary differentiable ops the autograd graph
stays intact. The bound is inlined rather than named because `_torch_svd_cast` is
TorchScript-compiled via `zca_mean`, and TorchScript cannot close over a module
global.

**Verification.** `tests/core/test_helpers.py` passes: 40 cases on CPU, and on
`--device=mps --dtype=float32` the SVD and svdvals boundary cases pass.
`test_batch_at_mps_ceiling` and `test_svdvals_at_mps_ceiling` pin exactly 8192
elements (512 4x4 matrices) and assert the CPU fallback is taken and outputs
return on the input device. `solve_pnp_dlt` at B=512, N=6 now returns finite
results on MPS, where it previously raised.

**Scope.** `find_essential` is *not* fixed: `run_5point` calls
`torch.linalg.eigvals`, which has no MPS kernel, so it raises at every batch
size for an unrelated reason. The release note and `TESTING.md` have been
corrected to say so rather than claiming all SVD-based solvers work.

Fixes #4201

### DIFF
--- a/kornia/core/utils.py
+++ b/kornia/core/utils.py
@@ -268,6 +268,12 @@ def _torch_histc_cast(input: torch.Tensor, bins: int, min: Union[float, bool], m
     return torch.histc(input.to(dtype), bins, min, max).to(input.dtype)
 
 
+# torch 2.14's MPS ``linalg.svd`` raises "Failed to created pipeline state object" -- a Metal shader
+# compilation failure, not memory pressure -- once a batched input holds more than this many elements,
+# whatever the per-matrix shape. ``svdvals`` and ``lstsq`` share the ceiling; eigen/QR/LU/cholesky do not.
+_MPS_SVD_MAX_NUMEL = 8192
+
+
 def _torch_svd_cast(input: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
     """Make torch.svd work with other than fp32/64.
 
@@ -278,6 +284,9 @@ def _torch_svd_cast(input: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor, to
     NOTE: in torch 1.8.1 this function is recommended to use as torch.linalg.svd
 
     For numerical stability, fp32 inputs are promoted to fp64 (except on MPS where fp64 is unsupported).
+
+    An MPS input past the shader-compilation ceiling is decomposed on the CPU and moved back, which
+    is what keeps the batched minimal solvers behind ``RANSAC`` working on Apple silicon.
     """
     if is_mps_tensor_safe(input):
         dtype = torch.float32
@@ -286,7 +295,14 @@ def _torch_svd_cast(input: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor, to
     else:
         dtype = _normalize_to_float32_or_float64(input.dtype)
 
-    out1, out2, out3H = torch.linalg.svd(input.to(dtype))
+    x = input.to(dtype)
+    if is_mps_tensor_safe(x) and x.numel() > _MPS_SVD_MAX_NUMEL:
+        # SVD is per-matrix, so decomposing the whole batch on the CPU gives the same result; the
+        # casts back to the MPS device keep the autograd graph intact.
+        U, S, Vh = torch.linalg.svd(x.cpu())
+        out1, out2, out3H = U.to(x.device), S.to(x.device), Vh.to(x.device)
+    else:
+        out1, out2, out3H = torch.linalg.svd(x)
     # Since kornia requires torch>=2.5.1, we can always use .mH
     out3 = out3H.mH
     return (out1.to(input.dtype), out2.to(input.dtype), out3.to(input.dtype))
--- a/kornia/core/utils.py
+++ b/kornia/core/utils.py
@@ -268,12 +268,6 @@ def _torch_histc_cast(input: torch.Tensor, bins: int, min: Union[float, bool], m
     return torch.histc(input.to(dtype), bins, min, max).to(input.dtype)
 
 
-# torch 2.14's MPS ``linalg.svd`` raises "Failed to created pipeline state object" -- a Metal shader
-# compilation failure, not memory pressure -- once a batched input holds more than this many elements,
-# whatever the per-matrix shape. ``svdvals`` and ``lstsq`` share the ceiling; eigen/QR/LU/cholesky do not.
-_MPS_SVD_MAX_NUMEL = 8192
-
-
 def _torch_svd_cast(input: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
     """Make torch.svd work with other than fp32/64.
 
@@ -296,7 +290,11 @@ def _torch_svd_cast(input: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor, to
         dtype = _normalize_to_float32_or_float64(input.dtype)
 
     x = input.to(dtype)
-    if is_mps_tensor_safe(x) and x.numel() > _MPS_SVD_MAX_NUMEL:
+    # torch 2.14's MPS ``linalg.svd`` raises "Failed to created pipeline state object" -- a Metal
+    # shader compilation failure, not memory pressure -- once a batched input holds more than 8192
+    # elements, whatever the per-matrix shape. The bound is inlined rather than named because this
+    # function is scripted (via ``zca_mean``) and TorchScript cannot close over a module global.
+    if is_mps_tensor_safe(x) and x.numel() > 8192:
         # SVD is per-matrix, so decomposing the whole batch on the CPU gives the same result; the
         # casts back to the MPS device keep the autograd graph intact.
         U, S, Vh = torch.linalg.svd(x.cpu())
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -379,6 +379,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* SVD-based solvers work on MPS for batches of any size. torch 2.14's MPS `linalg.svd` fails to build
+  its Metal pipeline state object once the input holds more than 8192 elements, so `RANSAC`,
+  `find_fundamental`, `find_essential` and `find_homography_dlt(..., solver="svd")` raised
+  "Failed to created pipeline state object" at any realistic batch; such a batch is now decomposed on
+  the CPU and moved back (closes #4201). (#4394)
 * `pixel2cam` now validates the full `Bx4x4` shape of `intrinsics_inv`, rejecting invalid matrix sizes
   before they cause unrelated transformation errors or return the wrong number of coordinate components. (#4381)
 * `Boxes.to_mask` leaves list-padding channels empty, including after coordinate transforms,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -383,7 +383,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   its Metal pipeline state object once the input holds more than 8192 elements, so `RANSAC`,
   `find_fundamental`, `find_essential` and `find_homography_dlt(..., solver="svd")` raised
   "Failed to created pipeline state object" at any realistic batch; such a batch is now decomposed on
-  the CPU and moved back (closes #4201). (#4394)
+  the CPU and moved back (closes #4201). (#4455)
 * `pixel2cam` now validates the full `Bx4x4` shape of `intrinsics_inv`, rejecting invalid matrix sizes
   before they cause unrelated transformation errors or return the wrong number of coordinate components. (#4381)
 * `Boxes.to_mask` leaves list-padding channels empty, including after coordinate transforms,
--- a/tests/core/test_helpers.py
+++ b/tests/core/test_helpers.py
@@ -206,6 +206,31 @@ class TestSvdCast:
         tol_val: float = 1e-1 if dtype == torch.float16 else 1e-3
         assert_close(a, u @ torch.diag_embed(s) @ v.transpose(-2, -1), atol=tol_val, rtol=tol_val)
 
+    def test_batch_above_mps_ceiling(self, device, dtype):
+        # torch 2.14's MPS SVD raises above 8192 input elements (#4201), far below the hypothesis
+        # count RANSAC's batched minimal solvers use; 1000 * 3 * 3 = 9000 elements clears it.
+        torch.manual_seed(0)
+        a = torch.randn(1000, 3, 3, device=device, dtype=dtype)
+        u, s, v = _torch_svd_cast(a)
+
+        assert u.device == a.device
+        assert s.device == a.device
+        assert v.device == a.device
+        # Both half dtypes round three factors back before the reconstruction contracts them.
+        tol_val: float = 1e-1 if dtype in (torch.float16, torch.bfloat16) else 1e-3
+        assert_close(a, u @ torch.diag_embed(s) @ v.transpose(-2, -1), atol=tol_val, rtol=tol_val)
+
+    def test_gradient_above_mps_ceiling(self, device, dtype):
+        # Decomposing the batch elsewhere must not detach it from the graph.
+        torch.manual_seed(0)
+        a = torch.randn(1000, 3, 3, device=device, dtype=dtype, requires_grad=True)
+        _, s, _ = _torch_svd_cast(a)
+
+        s.sum().backward()
+        assert a.grad is not None
+        assert a.grad.device == a.device
+        assert not a.grad.isnan().any()
+
 
 class TestSolveCast:
     def test_smoke(self, device, dtype):


kornia/core/utils.py
```diff
@@ -286,7 +289,18 @@ def _torch_svd_cast(input: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor, to
     else:
         dtype = _normalize_to_float32_or_float64(input.dtype)
 
-    out1, out2, out3H = torch.linalg.svd(input.to(dtype))
+    x = input.to(dtype)
+    # torch 2.14's MPS ``linalg.svd`` raises "Failed to created pipeline state object" -- a Metal
+    # shader compilation failure, not memory pressure -- once a batched input holds more than 8192
+    # elements, whatever the per-matrix shape. The bound is inlined rather than named because this
+    # function is scripted (via ``zca_mean``) and TorchScript cannot close over a module global.
+    if is_mps_tensor_safe(x) and x.numel() > 8192:
```
ducha-aiki: [P1] Include exactly 8192 elements in the MPS fallback

The failure starts at the boundary itself. On local MPS with PyTorch 2.14.0, `_torch_svd_cast(torch.randn(512, 4, 4, device='mps'))` still raises `RuntimeError: Failed to created pipeline state object ... XPC_ERROR_CONNECTION_INTERRUPTED`, because `512 * 4 * 4 == 8192` does not enter this branch. The adjacent 511-matrix case succeeds on MPS, and 513 matrices succeed via the new CPU fallback. Please include equality in the guard, correct the corresponding boundary prose, and add an exact-boundary regression; the new 9000-element test cannot catch this hole.

ducha-aiki (changes requested):
The CPU fallback preserves output placement and gradients in the exercised cases, and the helper remains scriptable. One boundary case still reproduces the reported failure and needs correction before merging.

Validation at this head in an isolated checkout: tests/core/test_helpers.py passed 62 CPU float32/float64 cases; the SVD class passed 3 MPS float32 cases. Additional eager and scripted MPS probes passed for 511 and 513 matrices of shape 4x4 (8176 and 8208 elements), including reconstruction and finite singular-value gradients. Exactly 512 such matrices still fail, as detailed inline. Imports were verified against the review checkout. The hosted macOS/PyTorch 2.5.1 job also needs a successful replacement for its RRDBNetBuilder ONNX export timeout. No full-suite or CUDA run was performed locally.

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (a1a4c0dd) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4455.fixed.md`.

ducha-aiki (changes requested):
Review of `a1a4c0dd0f2dc63fbecbb101c8cc542af9b79309`.

[P1] Include the exact 8192-element boundary in the MPS fallback. `kornia/core/utils.py:297` still uses `x.numel() > 8192`, but torch 2.14's MPS SVD fails at 8192 itself: a `(512, 4, 4)` batch takes the MPS branch and still raises the Metal pipeline-state error. Change the guard to include equality, correct `kornia/core/utils.py:294` and `changelog.d/4455.fixed.md:2`, and add a regression at exactly 8192 elements.

Validation and limits: Head `a1a4c0` only merges main and migrates the release note; the reviewed SVD implementation and its 9000-element test are unchanged. The boundary was measured on MPS torch 2.14: 511 matrices succeed, 512 fail, and 513 succeed only through the CPU fallback.

vimalyad:
Fixed in 4f48cf8d: the element-count guard used `>` instead of `>=`, so exactly 8192 elements (e.g. 512 4x4 matrices) fell through to the MPS path and hit the bug, while the 8208-element case happened to be caught. The guard now checks `>= 8192`, so that boundary routes through the CPU fallback too. Worth re-running the 512x4x4 probe to confirm it now passes.

The RRDBNetBuilder ONNX export timeout on the macOS/2.5.1 job is unrelated to this change — none of the pushed commits touch that path. Will look into it separately rather than fold it into this PR.

ducha-aiki (changes requested):
Review of `4f48cf8dec82c776e7081c77d0a9a431f5c39118`.

The prior [P1] is resolved. The guard is now `x.numel() >= 8192` (`kornia/core/utils.py:299`), and `test_batch_at_mps_ceiling` pins the exact boundary. On MPS, whatever the per-matrix shape, every 8192-element batch now succeeds with outputs on the input device, where main raised. TorchScript, `torch.compile`, gradients and half dtypes all behave. Two scope problems remain: the closing keyword and the release note both claim more than the change fixes.

**Boundary on MPS, torch 2.14.0, main vs head** (`_torch_svd_cast`, reconstruction error on success):

| input | numel | main | head |
|---|---|---|---|
| `(511, 4, 4)` | 8176 | ok | ok |
| `(512, 4, 4)` | 8192 | **raises** (pipeline state object) | ok, recon 2.4e-06 |
| `(513, 4, 4)` | 8208 | raises | ok |
| `(2047, 2, 2)` / `(2048, 2, 2)` | 8188 / 8192 | ok / **raises** | ok / ok |
| `(127, 8, 8)` / `(128, 8, 8)` | 8128 / 8192 | ok / **raises** | ok / ok |
| `(1024, 2, 4)`, `(1024, 8, 1)`, `(32, 16, 16)`, `(8, 32, 32)`, `(2, 64, 64)` | 8192 | all raise | all ok |
| `(1, 91, 91)`, `(91, 91)`, `(128, 128)` | 8281-16384 | ok | ok (now routed to CPU explicitly; torch already stages these on CPU) |

The cap applies to the whole tensor's element count. The batch-1 and unbatched rows succeed on main above 8192, so the ceiling hits the batched kernel only, as the #4201 correction says. Routing those extra shapes to the CPU is harmless. Multi-dim batches `(16, 32, 4, 4)` and `(4, 128, 4, 4)` also work at head.

**[P2] `Fixes #4201` would close the issue while its `svdvals` half is still broken.** #4201's title covers batched `svd`/`svdvals`/`lstsq`. kornia has no `lstsq` call site, but `_torch_linalg_svdvals` (`kornia/core/utils.py:311-325`) is still unguarded. It feeds `solve_pnp_dlt` (`kornia/geometry/calibration/pnp.py:180`). Measured on MPS at head:

```
_torch_linalg_svdvals (911, 3, 3) numel=8199 RAISE Failed to created pipeline state object
solve_pnp_dlt B=64  N=6 (svdvals numel=1152, svd numel=9216)  ok (64, 3, 4)   # main: RAISE
solve_pnp_dlt B=512 N=6 (svdvals numel=9216, svd numel=73728) RAISE Failed to created pipeline state object
```

The PR body already offers to extend the change. Please do: the same `>= 8192` guard and CPU round trip in `_torch_linalg_svdvals`, plus a boundary test like `test_batch_at_mps_ceiling`. Otherwise change `Fixes #4201` to a non-closing reference. The #4201 grep on the merge tree also hits `TESTING.md:319-320` ("That ceiling is why `RANSAC`'s batched minimal solvers still raise on MPS"). After this PR the fundamental minimal solver no longer hits the ceiling, so that sentence needs updating in the same change.

**[P2] `changelog.d/4455.fixed.md` overstates the fix.** "SVD-based solvers work on MPS for batches of any size" is contradicted by `solve_pnp_dlt` above. `find_essential` is listed as fixed, but it still raises on MPS at every batch size, because `run_5point` calls `torch.linalg.eigvals`, which has no MPS kernel (pinned in `testing/known_failure_xfails/mps_float32.txt`). Only the error changes:

```
                 main                                        head
B=16    NotImplementedError aten::_linalg_eigvals    NotImplementedError aten::_linalg_eigvals
B=1024  RuntimeError Failed to created pipeline...   NotImplementedError aten::_linalg_eigvals
```

`find_fundamental` and `find_homography_dlt(..., solver="svd")` are genuinely fixed. `find_fundamental` at B=1024 raised on main. At head it matches native-MPS results computed in 256-chunks exactly (max diff 0.0). `find_homography_dlt(solver="svd")` at B=128 is a silent fix worth naming: on main the `RuntimeError` is caught, it warns "SVD did not converge" and returns `torch.empty` garbage (max |dH| vs CPU 3.6e+03). At head it matches CPU to 1.5e-05. Please drop `find_essential` from the list, or say only its SVD stage is fixed, and scope the lead sentence to what works.

**[P3]** Stale "above 8192" wording remains in the PR title, in the `tests/core/test_helpers.py:211` comment ("raises above 8192 input elements"), and in the PR body. The body also describes the test as "1024 3×3 matrices — 9216 elements" and "37 tests"; the test uses 1000 x 3 x 3 = 9000 and the file has 64 CPU cases.

**What I ran** — `.venv` python 3.11, torch 2.14.0 (plus the 2.5.1 overlay), `kornia.__file__` checked against the PR and main worktrees in every probe; MPS runs with `PYTORCH_ENABLE_MPS_FALLBACK` unset.
- `tests/core/test_helpers.py`, CPU float32/float64: 64 passed (2.14.0) and 64 passed (2.5.1). CPU float16 with the `cpu-float16` profile: 36 passed, 3 skipped. CPU bfloat16 with the `cpu-bfloat16` profile: 35 passed, 3 skipped, 1 xfailed.
- `TestSvdCast` on `--device=mps --dtype=float32`: 4 passed (2.14.0), 4 passed (2.5.1).
- Plant-and-revert, PR worktree: guard set back to `x.numel() > 8192`. `test_batch_at_mps_ceiling[mps-float32]` FAILED with `RuntimeError: Failed to created pipeline state object` and the other 3 passed; reverted, worktree clean. The spy assertion `seen == ["cpu"]` would still fail under `>` on an MPS device without the ceiling, such as the hosted runner.
- TorchScript: `torch.jit.script(_torch_svd_cast)` on MPS at `(511|512|513, 4, 4)` and `(2048, 2, 2)`: all ok, outputs on mps, recon ≤ 2.4e-06. `torch.compile(backend="eager", fullgraph=True)` at the same boundary shapes: ok.
- Dtype/device: float16 and bfloat16 MPS inputs at `(512, 4, 4)` return float16/bfloat16 tensors on mps, with the same reconstruction error as native `(511, 4, 4)`: 2.2e-03 (fp16) and 1.9e-02 (bf16).
- Gradients through the fallback: a gauge-invariant loss touching U, S and V, MPS vs CPU reference, at `(512, 4, 4)`, `(1000, 3, 3)` and `(128, 8, 9)`. Grad on mps, finite, max |g_mps − g_cpu| = 9.1e-06 / 6.7e-06 / 6.7e-05 (max |g| ≈ 9-14).
- Merge tree with `origin/main` is clean. CI job `tests-mps` (macos-15) concluded success at this head and ran `tests/core/test_helpers.py`.
- Not run: `RANSAC` end-to-end on MPS (its homography path uses `torch.linalg.qr`, which is unsafe on this machine) and CUDA.

ducha-aiki (changes requested):
Review of `298728bc`. All three prior items are done: `_torch_linalg_svdvals` is guarded at `kornia/core/utils.py:326` with `test_svdvals_at_mps_ceiling`, the fragment now says `find_essential` still raises and why, and `TESTING.md:319-325` is rewritten.

The measurements back the whole scope. On MPS at torch 2.14.0, main raises for every `numel >= 8192` (`8192`/`8193` as `(N,1,1)`, `(512,4,4)`, `(513,4,4)`, `(2048,2,2)`, `(911,3,3)`) and the head returns on `mps:0` at all of them with reconstruction error <= 2.4e-06. `solve_pnp_dlt` at `B=64/512/1024` went from raising to finite `(B,3,4)` on device; `find_fundamental` at `B=1024` is fixed; `find_homography_dlt(solver="svd")` at `B=128`/`B=1024` went from a "SVD did not converge" warning plus `torch.empty` garbage to real values; `find_essential` still raises `aten::_linalg_eigvals` at every batch size, as the fragment says. Every guard is pinned on MPS: dropping the svdvals guard fails `test_svdvals_at_mps_ceiling`, changing `>=` to `>` fails both boundary pins, and dropping the svd guard fails three. TorchScript, `torch.compile(fullgraph=True)`, half dtypes and gradients (`max |g_mps - g_cpu| = 0.0` for svdvals) all hold on 2.14.0 and on the 2.5.1 floor.

Two things block merge.

### [P2] `.agents/skills/kornia-developer/SKILL.md` does not belong to this PR

It arrived in `39824fd6b` alongside the svdvals guard and is a byte-identical copy of the already-tracked `.claude/skills/kornia-developer/SKILL.md`:

```
$ git diff --name-status origin/main...<head>
A	.agents/skills/kornia-developer/SKILL.md
M	TESTING.md
A	changelog.d/4455.fixed.md
M	kornia/core/utils.py
M	tests/core/test_helpers.py

$ git ls-tree -r origin/main --name-only | grep -iE "SKILL.md|^\.agents/"
.claude/skills/kornia-developer/SKILL.md

$ diff <(git show <head>:.agents/skills/kornia-developer/SKILL.md) \
       <(git show origin/main:.claude/skills/kornia-developer/SKILL.md)   # empty, 92 lines
```

Ninety-two lines of agent instructions unrelated to MPS SVD, in a second location that will drift from the first, and nothing in the repo references `.agents` at all. Please drop the file and the directory.

### [P2] `changelog.d/+migration-024.added.md:12-15` still carries the sentence you fixed in `TESTING.md`

It says the batched ceiling "keeps `RANSAC`'s batched minimal solvers off the device ([#4201])". That fragment is pending, so it renders into the same release notes as `4455.fixed.md` and the two bullets will contradict each other about the issue this PR closes. `grep -rnE "#4201|issues/4201" kornia/ tests/ docs/ testing/ TESTING.md changelog.d/` is how it turns up, and it is the only remaining stale hit.

### [P3] `changelog.d/4455.fixed.md` breaks the fragment convention twice

`changelog.d/README.md:16-17` says "Write a short description without a heading or an outer bullet. Towncrier adds the heading, bullet, and PR link." The file starts with `* ` and ends with `(#4455)`, so `towncrier build --draft` renders:

```
- * Batched SVD no longer fails on MPS at 8192 elements or more. ...
    (closes #4201). (#4455) ([#4455](https://github.com/kornia/kornia/pull/4455))
```

Drop the leading `* `, the trailing `(#4455)`, and the parenthetical `(closes #4201)` — the closing keyword belongs in the PR body, which already has it. The `Check changelog fragments` job catches none of this.

### [P3] The unbatched exemption in that paragraph is wrong, and the guard is better than advertised

`TESTING.md:316-318` (and the same clause in `+migration-024`) says "A single unbatched matrix is unaffected at any size". On main, torch 2.14 MPS `svd` and `svdvals` raise for an unbatched matrix whose `numel` is *exactly* 8192, and succeed on either side of it — 3/3 trials each, `PYTORCH_ENABLE_MPS_FALLBACK` unset:

```
(128,64) (64,128) (512,16) (1024,8) (2048,4) (8192,1) (1,8192) (1,128,64)  numel=8192 -> RAISE
(128,63) 8064 ok   (129,64) 8256 ok   (64,129) 8256 ok   (8191,1) ok   (8193,1) ok
(150,60) 9000 ok   (200,50) 10000 ok  (256,64) 16384 ok  (1,128,128) 16384 ok
```

So `numel() >= 8192` is not merely harmless for unbatched inputs — it is *necessary* for the exactly-8192 case, and the head fixes `(1, 128, 64)` too. While editing the paragraph, drop the "unaffected at any size" claim and soften the word "batched" in the `kornia/core/utils.py:294-296` comment. The over-coverage costs nothing: `(91,91)` 3.94 -> 3.33 ms, `(128,128)` 5.18 -> 4.53 ms, `(512,512)` 57.9 -> 55.3 ms, because torch was already staging those on the CPU.

Also worth a line in `_torch_linalg_svdvals`' docstring, which only got an inline comment while `_torch_svd_cast` got a NOTE. On the 2.5.1 floor the guard leaves it asymmetric on MPS — `(512,4,4)` works through the CPU route while `(511,4,4)` still raises `NotImplementedError: aten::_linalg_svd.U`. Both raise on main, so it is an improvement, but it is surprising enough to name.

### CI — not your fault

`half-precision (float16) (3.11, 2.9.1)` was *cancelled*, not failed: it hit the 45-minute job timeout (08:10:07 -> 08:55:26) inside `tests/feature/test_scale_space_detector.py` on an `INTEL(R) XEON(R) PLATINUM 8573C` runner. That is the known Intel-only float16 hang (#4214); `tests/core/test_helpers.py` had already completed. `collector` is the aggregator reflecting it. A rerun is a coin flip.

### `Fixes #4201` is now accurate

Both kornia SVD entry points are guarded, `grep -rn lstsq kornia/` finds no call site, and the issue's "two `RANSAC` pins" no longer exist — `tests/geometry/test_ransac.py:35` skips via `pytest.mark.mps_process_abort` (#4204) and `testing/known_failure_xfails/mps_float32.txt` holds only `eigvals` and `yuv` entries, so there is nothing to delete there.

Not run: `RANSAC` end to end (its 4-point homography path calls `torch.linalg.qr` at `kornia/geometry/homography.py:261`, which is unsafe on this machine, and #4204 aborts the process) and CUDA.


ducha-aiki (changes requested):
Re-review of `e9c94a87dd5b209ad93fe78dce2b6854bec9e943`.

The duplicated skill file is gone and the migration fragment now acknowledges the repaired RANSAC paths. Two previously reported documentation defects remain in the current head.

### [P2] Remove the false unbatched exemption everywhere this PR edits it — `TESTING.md:316`

`TESTING.md:316-318` and `changelog.d/+migration-024.added.md:18` still say a single unbatched matrix is unaffected at any size, while the measured torch 2.14 MPS behavior is that unbatched matrices with exactly 8192 elements fail too (`(128,64)`, `(64,128)`, `(8192,1)`, and `(1,8192)` all raise; sizes on either side succeed). The implementation's shape-agnostic `x.numel() >= 8192` guard correctly fixes those inputs, so the public docs understate the fix and contradict the reason the guard has no batch-rank check. Drop the exemption and change the `kornia/core/utils.py:294-295` / `:325-326` comments from "batched input" to "input".

### [P2] Let towncrier supply the bullet and PR link — `changelog.d/4455.fixed.md:1`

The fragment still starts with `* ` and ends with both `(closes #4201)` and `(#4455)`, despite `changelog.d/README.md` requiring plain fragment prose because towncrier adds the outer bullet and PR link. The rendered release note therefore contains a nested bullet and a duplicate manual/automatic #4455 citation. Remove the leading marker and trailing parentheticals; keep `Fixes #4201` in the PR body.

Validation: all eight focused CPU guard tests passed on float32 and float64; the prior MPS measurements still establish the implementation behavior. The only hosted red job failed before environment setup after three Pixi HTTP 500 responses, so it is unrelated but still needs a successful rerun.

vimalyad:
Both remaining items are addressed in `5f530ce1`.

**[P2] false unbatched exemption.** Confirmed on an M5 Pro, torch 2.14.0, float32 — your
measurement reproduces exactly:

| shape | elements | result |
|---|---|---|
| `(128, 64)`, `(64, 128)`, `(8192, 1)`, `(1, 8192)` | 8192 | raises |
| `(127, 64)`, `(64, 127)`, `(8191, 1)`, `(1, 8191)` | 8128 / 8191 | ok |
| `(129, 64)`, `(64, 129)`, `(8193, 1)`, `(1, 8193)` | 8256 / 8193 | ok |

The exemption is gone from `TESTING.md` (both the table row and the prose) and from
`changelog.d/+migration-024.added.md`, and the comments at `kornia/core/utils.py:294` and
`:325` now say "input" rather than "batched input".

While checking it I think the boundary has a simpler explanation than "8192 elements", and
I've added it to `TESTING.md` since it justifies why the guard counts elements and ignores
batch rank: it is a **byte** count. 8192 float32 elements are exactly 32768 bytes, which is
the MPS threadgroup limit, and torch's own fallback — the
`matrix too large to stage in MPS threadgroup memory ... falling back to CPU` path — triggers
on *greater than* that limit. So an input sitting exactly on the limit neither falls back nor
compiles, which is why 8191 and 8193 both pass and only 8192 fails.

That also predicts the limit moves with dtype rather than element count. I could only test one
other dtype usefully: float16 SVD fails on MPS at every size I tried, including well under the
limit, so it looks like a separate defect rather than the same ceiling. float64 isn't available
on MPS at all. Happy to drop that paragraph if you'd rather the doc stayed with the measured
element count.

**[P2] changelog fragment.** Leading `* ` and the trailing `(closes #4201)` / `(#4455)` are
removed, so towncrier supplies the bullet and the PR link. `Fixes #4201` stays in the PR body.

On the red job: agreed, it is the three Pixi HTTP 500s before environment setup, and the
`collector` failure is just mirroring it. I can't trigger a rerun from outside the repo, but
this push creates a new head, so fresh runs should start on their own.

Tests: 40 in `tests/core/test_helpers.py` on CPU, and the 26 MPS-selected ones pass on real
Apple hardware.

ducha-aiki (changes requested):
Re-review of `5f530ce1455c827d3b6706802aec76a06f19024b`.

The previous implementation and documentation blockers are resolved: both SVD entry points now include the exact 8192-element boundary, the tests pin the CPU fallback, the false unbatched exemption is gone, and the towncrier fragment format is fixed. One user-visible description remains inaccurate.

**[P2] Describe the homography failure as silent invalid output — `changelog.d/4455.fixed.md:3`**

Lines 3–4 group `find_homography_dlt(..., solver="svd")` with `find_fundamental` and `solve_pnp_dlt` and say all three raised the MPS pipeline-state exception. `find_homography_dlt` catches that `RuntimeError` at `kornia/geometry/homography.py:244-246`, emits `SVD did not converge`, and returns an uninitialized `torch.empty` tensor. Callers could therefore consume arbitrary homographies rather than see the exception. This silent-corruption behavior is the most consequential part of that fix and should not be described as a hard failure. Please distinguish it in the fragment, for example by saying the first and third functions raised while `find_homography_dlt` warned and returned invalid output.

Validation on the exact head, with imports resolved to the `/tmp/kornia-pr4455` worktree: the focused SVD helper tests passed 10 CPU float32/float64 cases and 5 MPS float32 cases. The `#4201` audit leaves only the updated explanatory docs and the three regression-test pins; there are no stale hits under `kornia/`. The merge tree against current `origin/main` is conflict-free, and every hosted check is green. No CUDA or full local suite was run.

ducha-aiki (approved):
Re-reviewed the latest delta. The changelog now accurately distinguishes the `find_homography_dlt` caught-exception/uninitialized-output path from the APIs that raise. The prior blocker is resolved, focused CPU/MPS coverage passed, and current CI is green. I found no unresolved code findings.

PR #4452, opened 2026-09-11
docs(augmentation): Convention blocks (batch 6a: shared contract and containers)

ducha-aiki:
## Summary

Documents the shared augmentation and container contracts with executable convention tests and issue-linked regressions for current limitations. `AugmentationBase2D` anchors the standard 2D contract; mix augmentations and 3D operations have explicitly separate scopes. `AugmentationSequential` anchors data-key dispatch, masks, boxes, matrices, and inverse behavior. Part of #4407, with an Unreleased changelog entry for #4452.

The documentation distinguishes probability gates, rank preservation, sampling from returned-parameter placement, seed reproducibility from parameter replay, serialization, and configuration-dependent dtype/compile support. Container guidance covers mask erasing, interpolation and precision, lossy box inversion, nested matrix limitations, patch selection/padding, video layouts, and dictionary dispatch. Qualitative guarantees link to tracking issues instead of relying on class counts or exhaustive exception lists.

This changes documentation, tests, test helpers, and test audit metadata. All 11 modified library files have identical runtime ASTs after removing docstrings compared with merge base `a7e2db6b`. Known runtime defects remain tracked in their linked issues.

## Coverage

Five augmentation test files exercise the documented contracts. The tests cover saved-parameter stability, probability-gate RNG order, mask/crop behavior, nested transforms, and per-operation empty-batch output shapes. Expected failures specify their exception types. Device/dtype kernel probes guard unsupported gradient and padding operations; the dissolving replay test uses a mock and downloads no checkpoint. The eager-RNG audit inventory matches the current source.

## Validation

Local Python 3.11:

- PyTorch 2.14.0, CPU float32/float64: 949 passed, 100 skipped, 17 xfailed across the five affected test files and the half-precision CI harness.
- PyTorch 2.5.1, CPU float32/float64: 885 passed, 102 skipped, 17 xfailed across the five affected test files.
- PyTorch 2.14.0, convention tests on MPS float32/float16: 150 passed, 17 xfailed.
- PyTorch 2.14.0, convention tests on CPU float16 and bfloat16 with focused known-failure profiles: each 92 passed, 9 xfailed.
- A final 3D equalization fixture adjustment was rechecked across all the above device/dtype combinations.
- Full pre-commit hooks: passed.
- Clean Sphinx HTML build with warnings as errors: passed; the PatchMix API anchor renders.
- Changed-library doctests: 10 passed.

[...]

PR #4456, opened 2026-09-11
fix(auto): refuse an inverse the drawn sub-policy cannot undo

### LINKED ISSUE #4442: W1b: `AutoAugment` / `RandAugment` / `TrivialAugment` expose an `inverse` that silently returns its input
The three `auto` composers inherit `inverse` from `ImageSequentialBase`, and `hasattr(aug, "inverse")` is `True` on
all three. The method inverts the geometric part of the drawn sub-policy and leaves everything else alone. Because
the sub-policy is drawn per forward call and most `auto` policies are dominated by intensity operations —
`equalize`, `posterize`, `solarize`, `color`, `contrast`, `brightness`, `sharpness`, `auto_contrast`, `invert` —
many draws contain no invertible geometry at all, and for those `inverse` returns its argument unchanged. It
reports no error and no warning, so a caller who round-trips an augmented batch back to the original frame gets the
augmented batch back and no indication that nothing happened.

Reading a round-trip error on its own does not show this: a large error is also what a genuinely wrong inverse
produces. The forward displacement is the control. Printing `fwd_moved = max|forward(x) - x|` beside
`roundtrip_max_err = max|inverse(forward(x)) - x|` makes the reading unambiguous — when the two are equal to the
last digit, `inverse` returned exactly what it was given. The digits themselves are a property of the sub-policy
each seed happens to draw and are not the pinnable fact; the **equality** is.

The neighbouring APIs refuse instead of pretending. `MixAugmentationBaseV2.inverse` raises
`RuntimeError: Inverse for <class> is not supported.` for all five mix classes, and `AugmentationSequential` raises
`NotImplementedError` for a 3D child even when the child is a flip that inverts exactly.

## Repro

```python
import torch


torch.manual_seed(0)
x = torch.rand(2, 3, 8, 6)   # H = 8, W = 6
MAKE = (
    ("AutoAugment", lambda: AA.AutoAugment("imagenet")),
    ("RandAugment", lambda: AA.RandAugment(n=2, m=10)),
    ("TrivialAugment", lambda: AA.TrivialAugment()),
)
print("has_inverse:", {n: hasattr(m(), "inverse") for n, m in MAKE})

no_ops = 0
for seed in (9, 11):
    for name, make in MAKE:
        torch.manual_seed(seed)
        aug = make()
        out = aug(x)
        back = aug.inverse(out)
        fwd = (out - x).abs().max().item()          # how far the forward moved the image
        err = (back - x).abs().max().item()         # how far the round trip is from the input
        no_ops += err == fwd
        print(f"seed={seed:2d}  {name:15s} fwd_moved={fwd:.5f}  roundtrip_max_err={err:.5f}"
              f"  inverse_returned_its_input={err == fwd}")
print(f"rows where inverse returned its input unchanged: {no_ops} of 6")

# what the neighbouring APIs do with the same impossible request
try:
    K.RandomMixUpV2(p=1.0).inverse()
except Exception as e:
    print("RandomMixUpV2.inverse()            ->", type(e).__name__ + ":", e)
torch.manual_seed(0)
seq = K.AugmentationSequential(K.RandomHorizontalFlip3D(p=1.0), data_keys=["input"])
v = torch.rand(1, 1, 4, 6, 8)
try:
    seq.inverse(seq(v))
except Exception as e:
    print("AugmentationSequential(3D).inverse  ->", type(e).__name__ + ":", e)
```

## Observed

```
has_inverse: {'AutoAugment': True, 'RandAugment': True, 'TrivialAugment': True}
seed= 9  AutoAugment     fwd_moved=0.83691  roundtrip_max_err=0.84211  inverse_returned_its_input=False
seed= 9  RandAugment     fwd_moved=0.11564  roundtrip_max_err=0.13503  inverse_returned_its_input=False
seed= 9  TrivialAugment  fwd_moved=0.99413  roundtrip_max_err=0.99413  inverse_returned_its_input=True
seed=11  AutoAugment     fwd_moved=0.83691  roundtrip_max_err=0.84211  inverse_returned_its_input=False
seed=11  RandAugment     fwd_moved=0.99707  roundtrip_max_err=0.99707  inverse_returned_its_input=True
seed=11  TrivialAugment  fwd_moved=0.28416  roundtrip_max_err=0.28416  inverse_returned_its_input=True
rows where inverse returned its input unchanged: 3 of 6
RandomMixUpV2.inverse()            -> RuntimeError: Inverse for RandomMixUpV2 is not supported.
AugmentationSequential(3D).inverse  -> NotImplementedError: The support for 3d inverse operations are not yet supported. You are welcome to file a PR in our repo.
```

## Expected

Two coherent outcomes.

1. **`inverse` refuses what it cannot invert** — the composer raises when the drawn sub-policy contains a
   non-invertible operation, matching what the mix classes and the 3D container path already do. A caller then
   learns immediately that an `auto` pipeline is not round-trippable.
2. **`inverse` documents itself as geometry-only** — the docstring states that only the rigid part of the drawn
   sub-policy is undone, that intensity operations are left in place, and that on a draw with no geometry the
   method is an identity. A caller can then decide whether that is useful.

The audit recommends the first: the method is reached through the generic `inverse` contract of
`ImageSequentialBase`, where a geometric 2D class either inverts for real or refuses — the three crop classes in
`cropping_mode="slice"` raise `NotImplementedError` saying inverse is only applicable for resample cropping mode,
rather than faking one. Silently returning the input is the third option, and it is the one that makes the current
behavior a trap. Either way, the count of no-op rows above is a property of the drawn sub-policies, not a
constant — rerunning with other seeds moves it.

## Classification

**Broken output** — a request the API cannot serve returns a plausible-looking wrong answer instead of an error.
Idiom: 4 (a degenerate case silently producing garbage). The method is inherited from `ImageSequentialBase`, so the
container-level rule is the broader question; this issue is about the three `auto` classes that advertise it.

## Disposition

`n/a — single symbol`. `ImageSequentialBase.inverse` is where the three composers get the method, and one change
there covers all of them; a focused fix is welcome as a PR. Batch 6 documents the geometry-only, silently-identity
behavior on the three classes and pins it, and does not change it.

## Environment

change it.

mohansree14:
Closes #4442.

`AutoAugment`, `RandAugment` and `TrivialAugment` inherit `inverse` from `ImageSequentialBase`, which undoes the geometric part of the drawn sub-policy and leaves everything else alone. `auto` policies are dominated by intensity operations and the sub-policy is redrawn on every forward pass, so many draws hold no invertible geometry at all — and for those, `inverse` returned its argument unchanged, reporting neither an error nor a warning.

The issue's repro makes the reading unambiguous by printing the forward displacement beside the round-trip error: when the two are equal to the last digit, `inverse` returned exactly what it was given. On `main`, 3 of its 6 rows do that.

## Change

Refuse instead, matching what the neighbouring APIs already do — the five mix classes raise `RuntimeError: Inverse for <class> is not supported.`, and the three crop classes in `cropping_mode="slice"` raise rather than fake an inverse. This is the issue's recommended outcome (1).

The guard sits on `PolicyAugmentBase.inverse`, where all three composers get the method:

```
RuntimeError: Inverse for RandAugment is not supported: the drawn sub-policy applied
RandomSolarize, RandomSharpness, which cannot be inverted. Only geometric operations
are invertible, and the sub-policy is redrawn on every forward pass.
```

A mixed draw is refused too: undoing only its geometry is the same trap, one digit closer to the input and no more round-trippable.

## What does not change

- **A geometry-only draw still inverts.** Verified byte-identical to `main`: nine round trips across the three composers and three seeds, `torch.equal(patched, main) == True`.
- **The nested path is untouched.** `AugmentationSequential` reaches a child composer through `InputSequentialOps.inverse` → `inverse_inputs`, not through `inverse`, so `test_sequential` and the container suite are unaffected. The container-level rule is the broader question the issue raises; this PR is scoped to the three `auto` classes that advertise the method.
- **The un-run policy error is preserved.** `inverse` before any forward pass still raises the original `ValueError: No parameters available for inversing, ...`, not the new refusal.

## Verification

```
tests/augmentation/test_auto_operation.py -k inverse
  on main:      3 failed, 1 passed   (DID NOT RAISE RuntimeError, one per composer)
  with the fix: 4 passed

tests/augmentation/container + tests/augmentation/test_auto_operation.py
  539 passed, 51 skipped
```

`ruff check` and `ruff format --check` clean; `mypy` reports no error in the changed module.

### DIFF
--- a/kornia/augmentation/auto/base.py
+++ b/kornia/augmentation/auto/base.py
@@ -20,6 +20,7 @@ from typing import Any, Dict, Iterator, List, Optional, Tuple, Union, cast
 import torch
 from torch import nn
 
+from kornia.augmentation._2d.geometric.base import GeometricAugmentationBase2D
 from kornia.augmentation.auto.operations.base import OperationBase
 from kornia.augmentation.auto.operations.policy import PolicySequential
 from kornia.augmentation.container.base import ImageSequentialBase, TransformMatrixMinIn
@@ -118,6 +119,58 @@ class PolicyAugmentBase(ImageSequentialBase, TransformMatrixMinIn):
             if not module.is_intensity_only():
                 return False
         return True
+
+    def _non_invertible_ops(self, params: List[ParamItem]) -> List[str]:
+        """Name every drawn operation that :meth:`inverse` cannot undo.
+
+        Args:
+            params: Parameters recorded by a forward pass.
+
+        Returns:
+            Class names of the drawn operations that are not geometric, in execution order.
+        """
+        names: List[str] = []
+        for (_, module), param in zip(self.get_forward_sequence(params), params):
+            subpolicy = cast(PolicySequential, module)
+            subparams = cast(Optional[List[ParamItem]], param.data)
+            for _, operation in subpolicy.get_forward_sequence(subparams):
+                operation = cast(OperationBase, operation)
+                if not isinstance(operation.op, GeometricAugmentationBase2D):
+                    names.append(operation.op.__class__.__name__)
+        return names
+
+    def inverse(
+        self, input: torch.Tensor, params: Optional[List[ParamItem]] = None, extra_args: Optional[Dict[str, Any]] = None
+    ) -> torch.Tensor:
+        """Undo the drawn sub-policy, or refuse when it cannot be undone.
+
+        Only geometric operations are invertible. A policy draw that contains an intensity operation is
+        not round-trippable, and inverting it would return a tensor that still carries that operation --
+        for a draw with no geometry at all, the input unchanged. This raises instead, as
+        :meth:`kornia.augmentation.MixAugmentationBaseV2.inverse` already does for the mix classes.
+
+        Args:
+            input: Tensor produced by a forward pass.
+            params: Parameters used during that forward pass. Defaults to the cached ones.
+            extra_args: Optional per-input-type overrides.
+
+        Returns:
+            The inverse-transformed tensor.
+
+        Raises:
+            RuntimeError: The drawn sub-policy contains a non-invertible operation.
+        """
+        if params is None:
+            params = self._params
+        if params is not None:
+            non_invertible = self._non_invertible_ops(params)
+            if non_invertible:
+                raise RuntimeError(
+                    f"Inverse for {self.__class__.__name__} is not supported: the drawn sub-policy applied "
+                    f"{', '.join(non_invertible)}, which cannot be inverted. Only geometric operations are "
+                    "invertible, and the sub-policy is redrawn on every forward pass."
+                )
+        return super().inverse(input, params, extra_args=extra_args)
 
     def forward_parameters(self, batch_shape: torch.Size) -> List[ParamItem]:
         """Generate per-module parameters for one policy forward pass.
--- a/tests/augmentation/test_auto_operation.py
+++ b/tests/augmentation/test_auto_operation.py
@@ -215,3 +215,62 @@ def test_operation_preserves_input_dtype(device, low_dtype):
         assert out.dtype == low_dtype, f"{type(op).__name__}: {out.dtype} != {low_dtype}"
         checked += 1
     assert checked, f"no op could run in {low_dtype} on this build"
+
+
+# AutoAugment sub-policies are ``(name, probability, magnitude_bin)``; the other two are
+# ``(name, min_magnitude, max_magnitude)``. RandAugment and TrivialAugment allow only one
+# operation per sub-policy, so a mixed draw has to come from two of them -- and cannot
+# happen at all under TrivialAugment, which draws exactly one operation.
+_AA_GEOMETRIC, _AA_INTENSITY = ("translate_x", 1.0, 5), ("solarize", 1.0, 5)
+_GEOMETRIC, _INTENSITY = ("translate_x", -0.5, 0.5), ("solarize", 0.0, 1.0)
+
+
+@pytest.mark.parametrize(
+    ("intensity_only", "mixed", "geometry_only"),
+    [
+        (
+            lambda: AutoAugment(policy=[[_AA_INTENSITY]]),
+            lambda: AutoAugment(policy=[[_AA_GEOMETRIC, _AA_INTENSITY]]),
+            lambda: AutoAugment(policy=[[_AA_GEOMETRIC]]),
+        ),
+        (
+            lambda: RandAugment(n=1, m=15, policy=[[_INTENSITY]]),
+            lambda: RandAugment(n=2, m=15, policy=[[_GEOMETRIC], [_INTENSITY]]),
+            lambda: RandAugment(n=1, m=15, policy=[[_GEOMETRIC]]),
+        ),
+        (
+            lambda: TrivialAugment(policy=[[_INTENSITY]]),
+            None,
+            lambda: TrivialAugment(policy=[[_GEOMETRIC]]),
+        ),
+    ],
+    ids=["AutoAugment", "RandAugment", "TrivialAugment"],
+)
+def test_inverse_refuses_a_non_invertible_draw(intensity_only, mixed, geometry_only, device, dtype):
+    # ``inverse`` can only undo geometric operations. The sub-policy is redrawn on every
+    # forward pass, so a draw holding an intensity operation used to come back through
+    # ``inverse`` with that operation still applied -- and on a draw with no geometry at
+    # all, as the input unchanged, with no error and no warning. It refuses now, the way
+    # ``MixAugmentationBaseV2.inverse`` already does.
+    x = torch.rand(2, 3, 8, 6, device=device, dtype=dtype)
+
+    aug = intensity_only()
+    with pytest.raises(RuntimeError, match="is not supported"):
+        aug.inverse(aug(x))
+
+    if mixed is not None:
+        # half-undoing a mixed draw is the same trap, so it is refused too
+        aug = mixed()
+        with pytest.raises(RuntimeError, match="is not supported"):
+            aug.inverse(aug(x))
+
+    # a geometry-only draw still inverts, unchanged
+    aug = geometry_only()
+    aug.inverse(aug(x))
+
+
+def test_inverse_without_a_forward_pass_still_reports_missing_params(device, dtype):
+    # the refusal must not shadow the pre-existing error for an un-run policy
+    x = torch.rand(2, 3, 8, 6, device=device, dtype=dtype)
+    with pytest.raises(ValueError, match="No parameters available"):
+        RandAugment(n=1, m=5).inverse(x)


ducha-aiki (changes requested):
Reviewed `1584092b18c07657a8dc2bcc7504286371f1ba77`.

The guard correctly refuses partial inversion when the drawn policy includes an operation it cannot invert. Focused validation passed 28 CPU float32 auto-operation cases and 16 inverse-selected cases across all four CPU floating dtypes, in an isolated checkout with import origin verified.

Two release-readiness corrections remain:
- [P2] Add the required Unreleased bug-fix entry citing #4456. The new RuntimeError behavior affects AutoAugment, RandAugment and TrivialAugment and needs a release note.
- [P2] tests/augmentation/test_auto_operation.py:267 calls inverse for the geometry-only case but never asserts the returned result. Use an exactly invertible geometric draw (or an appropriate tolerance) and assert the round trip so the retained inverse behavior is actually covered.

The required macOS/PyTorch 2.5.1 job also remains red in the RRDBNetBuilder ONNX export timeout, outside the changed files, with a failed collector. Successful replacement CI is needed. No fresh full-suite, CUDA or MPS run was performed locally.

mohansree14:
Thanks for the review. Both P2 items are addressed in `aac7e337`.

**CHANGELOG entry.** Added under `Unreleased` → `Bug fixes`, citing #4456 and naming all three policies affected by the new `RuntimeError`.

**The unasserted geometry-only inverse** (`tests/augmentation/test_auto_operation.py:267`). It now asserts the round trip:

```python
ramp = 0.5 * (
    torch.linspace(0, 1, 8, device=device, dtype=dtype)[:, None]
    + torch.linspace(0, 1, 6, device=device, dtype=dtype)
).expand(2, 3, 8, 6)
params = aug.forward_parameters(ramp.shape)
kept = aug.inverse(aug(torch.ones_like(ramp), params=params), params=params) > 1 - 1e-4
assert kept.any(), "the drawn geometry pushed the whole image out of frame"
round_trip = aug.inverse(aug(ramp, params=params), params=params)
assert_close(round_trip[kept], ramp[kept], rtol=1e-4, atol=1e-4)
```

On the "exactly invertible draw" half of your suggestion: I could not construct one through the policy API. `rotate` rejects a non-symmetric range (`rand_augment/ops.py:76`), so 180° cannot be pinned, and `translate_*`/`shear_*` always sample from the range — for AutoAugment through the magnitude-bin table, which for `rotate` tops out at ±30°. There is no flip in any of the three name registries either. So the only exactly invertible draws available are degenerate zero-magnitude ones, which assert nothing.

Instead the assertion removes the two sources of round-trip error rather than absorbing them into a tolerance:

- **Interpolation error** — a ramp is an affine function of the coordinates, so bilinear resampling reproduces it exactly. Random noise does not survive two resamplings; a ramp does.
- **Border loss** — the pixels the translation pushed out of frame cannot be recovered by any correct inverse, so they are excluded. Running the *same* draw and the same round trip over `ones` marks which pixels stayed in frame, with no assumption about the transform itself.

What is left is genuine inverse error, and the tolerance is tight enough to see it.

Validation for the assertion itself, CPU float32:

- Interior error over 300 seeds: AutoAugment 1.8e-07, RandAugment 1.2e-07, TrivialAugment 4.3e-06 — all well inside `atol=1e-4`. `kept` was never empty.
- Teeth check: substituting a no-op inverse (returning the augmented tensor) fails the assertion on 50/50 draws.
- `tests/augmentation/test_auto_operation.py` — 28 passed. `ruff check` and `ruff format --check` clean.

Note that RandAugment draws a zero magnitude on roughly a quarter of forwards, because `OperationBase._random_flip` multiplies the magnitude by a boolean rather than negating it (`operations/base.py:82-87`). Those draws make the test's geometry-only leg a no-op assertion rather than a failure, and the remaining draws still exercise a real inversion. That looks like a separate defect in the symmetric-magnitude sampler — happy to open an issue if you agree it is one, but I have left it out of this PR.

**CI.** The macOS / torch 2.5.1 failure was the RRDBNetBuilder ONNX export timeout, outside the changed files. The push re-triggered the full matrix on `aac7e337`; it is running now.

Still no CUDA or MPS run on my side — CPU only.

ducha-aiki (changes requested):
Delta re-review of `6fdecfa7154b818b6b5da83ae79e908f788444cf` (previous review was on `1584092b`).

Thanks for the follow-up. Both earlier asks are addressed. The `(#4456)` entry is under `Unreleased` → `Bug fixes` and says what used to happen and what happens now. The geometry-only leg now checks the round-trip value. I like the ramp plus `ones` mask: in float32 it catches a no-op inverse on 200/200 seeds for AutoAugment and TrivialAugment, and on 149/200 for RandAugment. The RandAugment misses are the zero-magnitude draws you described, which #4495 removes. Both merges of `main` are clean: each merge commit's tree is identical to `git merge-tree` of its parents, and the only `CHANGELOG.md` lines the branch adds are its own four. One problem came in with the new assertion, though, and it turns the two half-precision CI legs red.

**[P2] The geometry-only round-trip assertion fails on every run under float16 and bfloat16, so the two required half-precision CI legs are red.** `tests/augmentation/test_auto_operation.py:318` uses the threshold `> 1 - 1e-4`. When a tensor is compared with a Python scalar, the scalar is cast to the tensor's dtype, and `1 - 1e-4` becomes exactly `1.0` in float16 and in bfloat16. A round trip over `ones` never goes above 1, so `kept` is empty and the test stops at `assert kept.any(), "the drawn geometry pushed the whole image out of frame"`. The CI jobs show exactly that: `half-precision (float16) / tests (3.11, 2.9.1)` and `half-precision (bfloat16) / tests (3.11, 2.9.1)` each fail `test_inverse_refuses_a_non_invertible_draw[cpu-{float16,bfloat16}-{AutoAugment,RandAugment,TrivialAugment}]` (3 failed, 9642 / 9704 passed), and I get the same six failures locally with `--xfail-known-failures --known-failure-profile=cpu-float16` / `cpu-bfloat16`. The threshold is not the only issue. With a dtype-aware one (`1 - 8 * torch.finfo(dtype).eps`, which left `kept` non-empty on 100/100 seeds), the worst interior error over 100 seeds is 1.0e-3 to 3.9e-3 in float16 and 7.8e-3 to 3.5e-2 in bfloat16. In float32 it is at most 1.8e-7. So `atol=1e-4` cannot hold in half precision either. Two possible fixes: run the value leg in float32/float64 only (for example, build `ramp` as float32 when the fixture dtype is a half dtype), keeping the refusal legs on every dtype; or scale both the threshold and the tolerance with the dtype. Don't add these nodes to the known-failure manifests, because this is a test bug, not a library defect.

**[P3] A gated-off intensity operation still triggers the refusal, and the message says it was applied.** `_non_invertible_ops` names every operation in the drawn sub-policy, including ones whose probability gate skipped them for every sample. For example, `AutoAugment(policy=[[("translate_x", 1.0, 5), ("solarize", 0.0, 5)]])` records `Solarize_1` with `batch_prob == [0., 0.]`, so the forward pass was pure geometry and the base inverse was exact. Head still raises `...the drawn sub-policy applied RandomSolarize, which cannot be inverted`. With the `imagenet` policy's 0.4 to 0.8 probabilities this happens on a real share of draws. Refusing there is conservative and consistent with the issue, so this does not block. Still, either skip operations whose recorded `batch_prob` is all zero, or change "applied" to "drew".

**CI:** the two half-precision legs above are red. When I checked, 19 other jobs had passed and 18 were still pending, with no other failures.

**Verified on this head** (isolated worktree, `kornia.__file__` checked):
- **Tests on head:** `tests/augmentation/test_auto_operation.py` passed on CPU float32+float64 (68), MPS float32 (45 passed, 2 skipped) and torch 2.5.1 CPU float32+float64 (68). On CPU float16 and bfloat16 the three geometry legs fail as described. `tests/augmentation/container` passed on CPU float32 (699 passed, 233 skipped).
- **Tests on `main`:** the PR's test file, copied onto `main`, fails all six `test_inverse_refuses_a_non_invertible_draw` cases on CPU float32/float64 with `DID NOT RAISE`, and passes the missing-params test.
- **Geometry-only inverse unchanged:** 60 inverse outputs (3 composers × 20 seeds) are `torch.equal` to `main`.
- **Combined with #4495:** with its magnitude-sign fix applied, the inverse tests still pass (5 unseeded runs).
- **Merge state:** a merge with current `main` (#4452) is clean. #4452 adds no Convention block or pin for the `auto` inverse, and `grep -rnE "#4442|issues/4442" kornia/ tests/` has no hits on head or on `main`.
- **Style:** `ruff@0.16.6` check and format are clean on the changed files.

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (71bc0a8c) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4456.fixed.md`.

ducha-aiki (changes requested):
Review of `71bc0a8ca332fdd3be7e4d8f4ada1040a3b08585`.

Re-review of `71bc0a8c`: the new head only performs the Towncrier merge, so the prior test failure remains, and there is a nonblocking opportunity to make the conservative inverse gate more precise.

[P2] `tests/augmentation/test_auto_operation.py:318` compares half tensors with `1 - 1e-4`, which rounds to exactly `1.0` in float16 and bfloat16. `kept` is therefore empty and line 319 fails in all six required half-job variants. Run the value round-trip leg only in float32/float64, or use a half-safe construction and dtype-scaled tolerance; do not add these nodes to the known-failure manifests.

[P3] `kornia/augmentation/auto/base.py:136` rejects every intensity operation present in the sampled policy without checking its recorded `batch_prob`. For `AutoAugment(policy=[[("translate_x", 1.0, 5), ("solarize", 0.0, 5)]])`, solarize is skipped for every row and the result is geometry-only, but `inverse()` raises while claiming it was applied. Consider ignoring operations whose recorded gate is zero for all rows and adding this regression, or make the conservative drawn-policy restriction explicit in the error wording. The half-precision test is the blocking request.

mohansree14:
Thanks for tracking the half-precision failure down to the threshold, and for the fragment move. Both items are addressed in `525c5a6a`.

**[P2] Half-precision round trip.** I took your first option. The geometry-only value check now builds `ramp` in float32 when the fixture dtype is float16 or bfloat16, so both `1 - 1e-4` and `atol=1e-4` mean what they say. The refusal checks still run on the fixture dtype, and nothing was added to the known-failure manifests.

**[P3] Gated-off intensity op.** I went with skipping instead of rewording. `_non_invertible_ops` now ignores an operation whose recorded `batch_prob` is zero for every row, since that operation left the input untouched. If any row applied it, it still refuses, so "applied" in the message is accurate now. `test_inverse_ignores_an_operation_its_gate_skipped` uses your example: `[("translate_x", 1.0, 5), ("solarize", 0.0, 5)]` inverts, and the same policy with solarize at `1.0` raises. The test fails against the previous guard. The fragment notes the gate rule.

**Verified** (CPU, torch 2.14.0):
- On the previous head, float16 and bfloat16 reproduce the same six failures as CI.
- `tests/augmentation/test_auto_operation.py`: float32 and float64 70 passed; `--xfail-known-failures --known-failure-profile=cpu-float16` 47 passed, 1 xfailed; `cpu-bfloat16` 48 passed.
- `tests/augmentation/container` float32: 737 passed, 233 skipped, 3 xfailed.
- `ruff check` and `ruff format --diff` are clean on the changed files.

Not run: CUDA, MPS or torch 2.5.1. The macOS / torch 2.5.1 legs may still time out in the `RRDBNetBuilder` ONNX export test, which is outside the changed files.

ducha-aiki (approved):
Review of `525c5a6abec8fc35fd054d9c12ca881aad13fe55`.

Both items from the last review are fixed. The geometry-only value leg now runs in float32 when the fixture is a half dtype, and the refusal legs still use the fixture dtype. `_non_invertible_ops` now skips an operation whose recorded `batch_prob` is zero on every row. I checked that this matches what the forward pass actually did: it doesn't let through an inverse that should be refused, and it doesn't refuse one that is exact. The only change from the previous head is the commit that makes these two fixes (`changelog.d/4456.fixed.md`, `kornia/augmentation/auto/base.py`, `tests/augmentation/test_auto_operation.py`).

**Prior [P2], half-precision round trip: resolved.**
- Focused CPU half runs with the manifest profile (`--xfail-known-failures`; `--verify-known-failures` only accepts the full `tests/` target):
  - `--dtype=float16 --known-failure-profile=cpu-float16`: 47 passed, 1 xfailed. The xfail is `TestAutoAugment::test_reproduce[cpu-float16]` (#4153), and `main` xfails it too.
  - `--dtype=bfloat16 --known-failure-profile=cpu-bfloat16`: 48 passed.
- The previous head's test file on this source still fails under float16+bfloat16 (5 failed, 3 passed).
- CI now runs both `half-precision (float16|bfloat16) / tests (3.11, 2.9.1)` with the full-suite `--verify-known-failures`, and both pass on `525c5a6a`. No manifest was edited.

**Prior [P3], gated-off intensity op: resolved, and the rule is exact.** `batch_prob` comes from `_AugmentationBase.__batch_prob_generator__`. It is exactly 0 or 1 for each row: `p == 0` gives `zeros`, and `0 < p < 1` gives `(rand < p)`. The `RelaxedBernoulli` that `OperationBase` installs goes to `_p_gen`/`_p_batch_gen`, which nothing reads. So "all zero" really does mean the op never touched the tensor. I ran 7 `AutoAugment` sub-policies × B∈{1,4} × 6 seeds × {float32, float16}, 168 draws in all, on the PR and on `main`:

| sub-policy | recorded gate (B=4) | PR | main |
|---|---|---|---|
| `translate_x 1.0, solarize 0.0` | sol `[0,0,0,0]` | inverts (interior err 1.8e-7) | inverts |
| `translate_x 1.0, solarize 1.0` | sol `[1,1,1,1]` | raises `applied RandomSolarize` | inverts silently |
| `translate_x 1.0, solarize 0.5` | sol `[1,0,1,0]` | raises | inverts silently |
| same, seed with sol `[0,0,0,0]` | | inverts | inverts |
| `translate_x 0.0, solarize 1.0` | tx `[0,0,0,0]` | raises | inverts silently |
| `solarize 0.0, equalize 0.0` | both zero | inverts (err 0.0) | inverts |
| `rotate 1.0, color 0.0, posterize 0.0` | intensity zero | inverts | inverts |
| `solarize 0.0, translate_y 1.0` | sol zero | inverts | inverts |

- **Forward unchanged:** all 168 forward outputs are `torch.equal` between the PR and `main`.
- **Accepted inverses unchanged:** all 104 draws the PR accepts give inverse outputs `torch.equal` to `main`.
- **Gated-off op really is a no-op:** over 20 seeds × {float32, float16, bfloat16}, `translate_x 1.0, solarize 0.0` and `rotate 1.0, color 0.0, posterize 0.0` give forward and inverse outputs `torch.equal` to the geometry-only policy with the same seed.
- **Mixed gates still refuse:** a gate that fires on any row still raises.
- **RandAugment and TrivialAugment:** their ops are built with probability `1.0` (`rand_augment/ops.py`), so the new skip can't change their behaviour.

**New test has teeth.**
- `test_inverse_ignores_an_operation_its_gate_skipped` fails on float32 and float64 with the previous head's `base.py` put back.
- With `main`'s `base.py`, all 8 refusal/gate cases fail and the missing-params test passes.

**[P3] The public `inverse` docstring still describes the old rule.** `kornia/augmentation/auto/base.py:154` says "A policy draw that contains an intensity operation is not round-trippable … This raises instead". `:168` says `RuntimeError: The drawn sub-policy contains a non-invertible operation.` Both are rendered through `.. automethod:: inverse` in `docs/source/augmentation.auto.rst`. After this commit, `AutoAugment(policy=[[("translate_x", 1.0, 5), ("solarize", 0.0, 5)]])` does contain an intensity operation and inverts without raising. The helper docstring and the fragment were updated to "applied". Suggest matching wording here, e.g. "…that applied an intensity operation to any sample…". This does not block.

**What I ran**
- **Interpreter:** repository `.venv` Python 3.11, torch 2.14.0, `kornia.__file__` checked in every probe (PR worktree and main worktree at `53866348`). The torch 2.5.1 overlay printed `2.5.1` and the PR worktree.
- **Test runs:**
  - `tests/augmentation/test_auto_operation.py` CPU float32+float64: 70 passed.
  - The same file on torch 2.5.1: 70 passed.
  - Half runs: as above.
  - `-k inverse` over all four dtypes, 20 repeated runs: 20/20 passed each time.
  - `tests/augmentation/container` CPU float32: 737 passed, 233 skipped, 3 xfailed.
- **Probes and plants:** the gate probes and the source plants described above.
- **Merge:** `git merge-tree --write-tree origin/main <head>` is clean. `closingIssuesReferences` is exactly #4442. `git grep -nE "#4442|issues/4442"` on the merge tree over `kornia tests docs` has no hits.
- **Not run:** CUDA; MPS locally (the CI `tests-mps` leg passed); a local ruff run (no ruff in the venv; `pre-commit.ci` and `typecheck` passed).

ducha-aiki (approved):

PR #4457, opened 2026-09-12
fix(augmentation): make Gaussian illumination serializable

### LINKED ISSUE #4435: W1b: `RandomGaussianIllumination` cannot be pickled or `torch.save`d
`RandomGaussianIllumination.__init__` defines its transform as a local function at `kornia/augmentation/_2d/intensity/gaussian_illumination.py:163` and stores it on the instance at `:177` as `self._fn`; `apply_transform` at `:187` is a one-line forward to it. A closure defined inside `__init__` has no importable qualified name, so `pickle` cannot serialize the instance, and `torch.save` fails the same way because it pickles the module. `copy.deepcopy` works, which is why the problem does not show up in a single-process script.

The practical consequence is a `DataLoader` with `num_workers > 0` under the `spawn` start method — the default on macOS and Windows — which has to pickle whatever the dataset or transform closes over. Every other 2D augmentation in `kornia.augmentation` survives that; this one does not, and the error a user sees names a private local function rather than the class.

The obvious fix, making `_apply_transform` a normal method the way the two sibling illumination classes do, is not a drop-in. The closure is load-bearing: this class defines its own public `compile()` at `:189`, which rebinds `self._fn = torch.compile(self._fn, ...)`, and the closure at `:163` is exactly what that hook replaces. Neither `RandomLinearIllumination` nor `RandomLinearCornerIllumination` has `_fn` or defines its own `compile()`, so neither is a template for the rewrite. Any fix has to keep the `compile()` hook working — for instance by keeping the method as the picklable default and letting `compile()` set a separate, optional compiled attribute — and the equivalence of such a rewrite was not executed here, so no byte-identical claim is made for it.

## Repro

```python
import copy, io, pickle, torch, kornia.augmentation as K

aug = K.RandomGaussianIllumination(p=1.0)
print("type(aug._fn):", type(aug._fn).__name__, "| qualname:", aug._fn.__qualname__)
print("RandomGaussianIllumination defines its own compile():", "compile" in vars(K.RandomGaussianIllumination))
for label, call in (("pickle.dumps ", lambda: pickle.dumps(aug)),
                    ("torch.save   ", lambda: torch.save(aug, io.BytesIO())),
                    ("copy.deepcopy", lambda: copy.deepcopy(aug))):
    try:
        call(); print(f"{label}: ok")
    except Exception as e:                                          # noqa: BLE001
        print(f"{label}: {type(e).__name__}: {e}")
for cls in (K.RandomLinearIllumination, K.RandomLinearCornerIllumination):
    sib = cls(p=1.0)
    pickle.dumps(sib)
    print(f"pickle {cls.__name__}: ok | has _fn: {hasattr(sib, '_fn')} | "
          f"defines its own compile(): {'compile' in vars(cls)}")
```

## Observed

```
type(aug._fn): function | qualname: RandomGaussianIllumination.__init__.<locals>._apply_transform
RandomGaussianIllumination defines its own compile(): True
pickle.dumps : AttributeError: Can't pickle local object 'RandomGaussianIllumination.__init__.<locals>._apply_transform'
torch.save   : AttributeError: Can't pickle local object 'RandomGaussianIllumination.__init__.<locals>._apply_transform'
copy.deepcopy: ok
pickle RandomLinearIllumination: ok | has _fn: False | defines its own compile(): False
pickle RandomLinearCornerIllumination: ok | has _fn: False | defines its own compile(): False
```

## Expected

`pickle.dumps` and `torch.save` succeed on a `RandomGaussianIllumination` instance, as they do on every other 2D augmentation, and the class keeps the public `compile()` hook it has today. That is a focused fix and welcome as a PR; it needs a pickle round-trip test and a check that `compile()` still rebinds what it is meant to rebind, rather than a straight hoist of the closure into a method.

## Classification

**API completeness** — a public `nn.Module` that cannot be serialized, failing with a message about a private local function. Idiom: none.

## Disposition

`n/a — single symbol`. Batch 6 documents the behavior above as it is and pins it. The fix is focused and welcome as a PR, with the `compile()` constraint above as part of it. The shape of the defect does not match any of the seven idioms the audit works with and is raised for ratification as a new one: a closure stored on `self` makes an `nn.Module` unserializable — the state exists and behaves correctly in-process, and only crosses a process boundary badly.

## Environment

NaCr05:

RandomGaussianIllumination currently stores an __init__-local function on the instance, so pickle and torch.save fail even before the first forward pass. Move the same computation to a private module-level callable while retaining the existing _fn dispatch and compile() hook.

Add regression coverage for uncompiled module round trips before/after forward, restored parameter replay, CHW/BCHW layouts, input non-mutation with fixed batch selections, and actual compiled-graph execution after restoration. Strengthen the existing Inductor test to compare against eager output with replayed parameters.

The serialization scope is uncompiled instances and compilation after restoration; this does not add persistence of backend-generated compiled callables or caches.

Fixes #4435.

Related to #4407; the broader audit remains open.

## What did you check?

Windows 11, Python 3.14.6, PyTorch 2.14.0+cpu. Import origins were verified against the intended worktree.

- Before the fix: the new CPU float32 round-trip tests produced 8 pickle/torch.save failures and 4 successful deepcopy controls.
- Focused Gaussian/Linear/LinearCorner illumination tests, CPU float32/float64: **66 passed**.
- New round-trip, input-preservation, and counting-backend tests across CPU float32/float64/float16/bfloat16: **68 passed**. The counting backend verifies one graph is compiled and actually executed twice after restoring each module.
- Manual Windows DataLoader check with num_workers=1 and multiprocessing_context="spawn": failed to pickle the transform on base 8fb9ad12; passed on this branch with the worker importing this checkout and outputs matching the parent.
- Full pre-commit checks: **passed**.
- Type checking: **exit 0**, with one deprecation warning in unchanged lightglue.py.

Commands for the passing repository checks:
```powershell
pixi run -e py314 test-module tests/augmentation/test_augmentation.py -k 'GaussianIllumination or LinearIllumination or LinearCornerIllumination' --device=cpu --dtype=float32,float64 -q
$env:KORNIA_TEST_OPTIMIZER = 'eager'
pixi run -e py314 test-module tests/augmentation/test_augmentation.py::TestRandomGaussianIllumination -k 'serialization_roundtrip or input_preserved or compile_after_serialization' --device=cpu --dtype=float32,float64,float16,bfloat16 -q
pixi run -e py314 typecheck --python .venv-py314
```

Default Inductor execution is blocked locally by InvalidCxxCompiler: Compiler: cl is not found. The same failure was reproduced on base; the custom backend checks dispatch and numerical equivalence, not Inductor code generation. Full-suite and accelerator validation remain for CI.

## How did AI help?

Codex assisted with investigation, implementation, regression tests, and the checks listed above. The serialization failures were reproduced before the implementation change, and the branch was checked afterward.


NaCr05:
Could a maintainer please rerun the failed jobs for this PR? The macOS / Python 3.12 / PyTorch 2.5.1 job timed out after 120 seconds in `TestSuperResolutionBuilders::test_to_onnx_exports[RRDBNetBuilder]`, causing the collector to fail. The other 38 checks passed.

I tried rerunning the failed jobs, but GitHub denied the request due to repository permissions. Thanks!

ducha-aiki (dismissed):
Reviewed exact head `b099f02c`.

Correct and minimal: the `__init__`-local closure becomes the module-level `_apply_gaussian_illumination`, the `_fn` dispatch and the `compile()` hook are unchanged, and forward output is byte-identical to main on CPU float32/float64 and MPS float32 for the same params, including a mixed `batch_prob` batch. Pickle and `torch.save` round trips replay bit-exactly; a spawn-context `DataLoader` with two workers fails on main and passes here with the worker importing the reviewed tree. The strengthened dynamo test genuinely compiles (one Inductor kernel, zero graph breaks, compiled output equal to eager). `Fixes #4435` names a single-symbol issue and is fully covered; #4407 is only "Related to"; the documented-defect grep is empty; the release note cites #4457.

Non-blocking:

- [P3] Pickling a compiled instance still fails, now with a misleading "not the same object" `PicklingError`. The PR scopes this out explicitly, and the sibling custom `compile()` hooks in `RandomGaussianBlur` and `ColorJitter` share it, whereas `nn.Module.compile()`-based augmentations stay picklable because `Module.__getstate__` drops `_compiled_call_impl`. A small `__getstate__` that drops the compiled callable would close the gap; fine as a follow-up.
- [P3] `torch_optimizer` is used only as a gate in the new compile test, and `_roundtrip` falls through to `deepcopy` for an unknown mode instead of failing.

Validation (isolated worktree, import origin verified): 20 new nodes (16 pickle/`torch.save` round trips, 4 compile-after-restore) fail on `8fb9ad12` and pass here; the 8 deepcopy nodes are controls that pass on both. Illumination selection CPU float32/float64 66 passed; cpu-float16 and cpu-bfloat16 profiles 20 passed, 1 pre-existing xfail each; MPS float32 21 passed; ruff clean. Not run: full suite, CUDA, the torch 2.5.1 floor.

ducha-aiki (dismissed):
Re-approving on `baeccb06`. That commit is a pure merge of `origin/main` (5be4852b) into the previously approved head, pushed to unblock the `CHANGELOG.md` adjacent-insertion conflict that yesterday's nine merges re-created: every path except `CHANGELOG.md` is tree-identical to `git merge-tree`, and the `CHANGELOG.md` lines added over main equal the branch's own added lines over its merge-base (no main line removed). Nothing about the reviewed change moved; CI is re-running on the merge commit.

ducha-aiki (dismissed):
Re-approving on `f6f1db92`. I pushed two commits on top of the approved head `baeccb06`:

1. `f492f43e` is a keep-both merge of `origin/main` (`dc0af1a4`), made for the recurring `CHANGELOG.md` adjacent-insertion conflict. It is a pure merge: every path except `CHANGELOG.md` is tree-identical to `git merge-tree`, and the non-blank lines it adds to `CHANGELOG.md` are exactly this branch's own lines.
2. `f6f1db92` is a docs-only reconciliation with #4452, which landed after the approval. #4452's Convention prose cites #4435 in three places: `docs/source/get-started/conventions.rst` names `RandomGaussianIllumination` as unpicklable, `docs/source/augmentation.base.rst` cites #4435 for configuration-dependent pickling, and the `AugmentationBase2D` docstring says some augmentations are not picklable. This branch makes all three false. On the merge I pickled every public `kornia.augmentation` module I could construct: 82 classes, including `RandomGaussianIllumination`, `AugmentationSequential` and `ImageSequential`. Only the `auto` policies (`AutoAugment`, `RandAugment`, `TrivialAugment`) still fail, which is #4469. So the two docs pages now cite #4469 for the caveat that remains, and the `AugmentationBase2D` bullet drops the pickling clause. After this, `grep -rnE "#4435|issues/4435" kornia/ docs/ tests/` is empty, both RST files parse without new docutils warnings, and the `GaussianIllumination` tests pass (42 on CPU float32/float64).

This merges once the matrix on `f6f1db92` is green.

Issue #4278, opened 2026-09-05
W1b: `unproject_meshgrid` — `.squeeze()` destroys the `W = 1` axis, corrupting `depth_to_3d_v2` and `warp_frame_depth`


ducha-aiki:
`unproject_meshgrid` calls `.squeeze()` on the meshgrid it builds. With `W = 1` that removes the width axis, and the grid then broadcasts across a phantom width — so `unproject_meshgrid(1, 3, K)` and `unproject_meshgrid(3, 1, K)` return the **same shape** `(1, 1, 3, 3)` with different values, violating the documented `(*, H, W, 3)` contract. `H = 1` is unaffected (the batch axis absorbs the squeeze), and `H = W = 1` is correct by accident.

Three public functions inherit it:

- **`depth_to_3d_v2` with `W = 1`** returns `(1, 3, 3, 3)` where `depth_to_3d` returns the correct `(1, 3, 1, 3)`; the values are the correct column tiled three times. This is the one input at which the two implementations, otherwise `torch.equal` under asymmetric `K`, `H ≠ W`, batching and float64, disagree.
- **`warp_frame_depth` with `W = 1`** returns a `(1, 2, 4, 4)` image for a `(1, 2, 4, 1)` input — a silently wrong-width result. `H = 1` is fine.
- **`depth_to_normals` with `W = 1`** returns a `(1, 3, 4, 4)` normal map for a `(1, 1, 4, 1)` depth, with its single column of normals laid along the phantom width. `H = 1` and `H = W = 1` are fine.

`W = 1` is a degenerate but perfectly valid shape (a single image column, a 1-pixel-wide crop), and nothing raises.

The `.squeeze()` is on the line that #2709 (closed 2025-05-07) and #2970 (closed 2025-12-29) touched when `unproject_meshgrid` was made batched; see also #4271 for the docstring half.

## Repro

```python
import torch
from kornia.geometry.depth import depth_to_3d, depth_to_3d_v2, depth_to_normals, unproject_meshgrid, warp_frame_depth

K = torch.tensor([[[100.0, 0.0, 4.0], [0.0, 100.0, 3.0], [0.0, 0.0, 1.0]]])

print("unproject_meshgrid(H=1, W=3):", tuple(unproject_meshgrid(1, 3, K).shape))
print("unproject_meshgrid(H=3, W=1):", tuple(unproject_meshgrid(3, 1, K).shape), " <- same shape, documented (*, H, W, 3)")
[...]

PR #4458, opened 2026-09-12
fix(geometry): reject degenerate spatial sizes in depth_to_normals

### LINKED ISSUE #4398: `depth_to_normals` returns all-zero "unit normals" when `H = 1` or `W = 1`, and `nan` in float16
`depth_to_normals` documents a unit normal per pixel. When either spatial axis has length 1 the result is a zero vector in float32 and float64, and `nan` in float16 — silently, with the documented output shape.

## Repro

```python
import torch
from kornia.geometry.depth import depth_to_normals

K = torch.tensor([[[100.0, 0.0, 1.0], [0.0, 100.0, 0.0], [0.0, 0.0, 1.0]]])
for shape in [(1, 1, 1, 3), (1, 1, 3, 1), (1, 1, 1, 1), (1, 1, 2, 3)]:
    for dt in (torch.float32, torch.float16, torch.float64):
        n = depth_to_normals(torch.full(shape, 2.0, dtype=dt), K.to(dt))
        print(shape, str(dt)[6:], tuple(n.shape), n.norm(dim=1).flatten().float().tolist())
```

Executed on `main` `147ecc4e`, torch 2.14.0 (cpu; mps gives the same values):

| depth shape | float32 | float16 | float64 |
|---|---|---|---|
| `(1, 1, 1, 3)` (H = 1) | `(1, 3, 1, 3)`, norms `[0, 0, 0]` | `nan` everywhere | norms `[0, 0, 0]` |
| `(1, 1, 3, 1)` (W = 1) | `(1, 3, 3, 1)`, norms `[0, 0, 0]` | `nan` everywhere | norms `[0, 0, 0]` |
| `(1, 1, 1, 1)` | `(1, 3, 1, 1)`, norm `[0]` | `nan` | `[0]` |
| `(1, 1, 2, 3)` (control) | norms all `1.0` | `1.0` | `1.0` |

A flat plane at `H = 3` returns `[0, 0, 1]` at the centre, so the sign convention itself is fine; only the degenerate sizes are wrong.

## Mechanism

`spatial_gradient` along a length-1 axis is identically zero, so one of the two tangent vectors is zero, `torch.linalg.cross` returns zero, and `F.normalize(normals, dim=-1, p=2)` divides `0` by `max(‖0‖, 1e-12)`: `0` in float32/float64, and in float16 the `1e-12` floor underflows to `0`, giving `0 / 0 = nan`. Nothing checks `H >= 2` / `W >= 2`; `KORNIA_CHECK_SHAPE(depth, ["B", "1", "H", "W"])` accepts any positive size.

## Classification

**Broken output** — a zero "unit normal" (or `nan`) with the documented shape and no error. Idiom 4 (a degenerate input silently produces garbage instead of raising) and idiom 2 (the `1e-12` normalisation floor is not scaled to the working dtype).

## Disposition

n/a — single symbol. The focused fix is to reject `H < 2 or W < 2` with a `ShapeError` (a normal field needs two tangent directions); if a zero normal is the intended answer for the degenerate case, say so in the docstring and use a dtype-scaled floor so float16 returns zeros too. Either way the `test_convention_*` pin goes on the chosen contract.

Found in the batch-5 audit (row `5c-nm-05`, tracking issue #4262), recorded there as unfiled; the `W = 1` case originally reached this function through #4278's squeeze and now reaches it on its own after #4298.

NaCr05:

`depth_to_normals` currently accepts singleton spatial axes even though a surface normal requires two tangent directions. These inputs can produce zero or non-finite normals; empty spatial dimensions fail later inside padding.

This implements the rejection policy suggested in the issue: raise `ShapeError` with the actual dimensions when `H < 2` or `W < 2`, before unprojection and spatial gradients. The normal computation for valid sizes is unchanged. The docstring and changelog describe the minimum dimensions, and the older #4278 pin keeps its projection/warping assertions while the new #4398 pin covers the normal-size contract.

Fixes #4398.

## What did you check?

Local environment: Windows, Python 3.14.6, PyTorch 2.14.0+cpu.

- Tests first: all 12 new invalid-size cases failed on the unmodified implementation; 12 valid-size controls passed.
- Full depth module, CPU float32/float64, including gradchecks and full-graph `torch.compile` tests using `aot_eager`: **304 passed, 2 xfailed**.
- Focused CPU float16 and bfloat16 checks with their known-failure profiles: **25 passed each**. These cover invalid dimensions, 2x2/2x3/3x2 planes, both depth meanings, shared/per-batch intrinsics, and the retained #4278 projection/warping pin. The profiles emit Linux/Windows platform compatibility warnings locally.
- Depth module docstring examples: **6 passed**.
- Full repository pre-commit checks: **passed**.
- Type check: **exit 0**, with one deprecation warning in the unchanged `kornia/feature/lightglue.py`.

```text
KORNIA_TEST_OPTIMIZER=aot_eager pixi run -e py314 test-module tests/geometry/test_depth.py --device=cpu --dtype=float32,float64 --runslow -q
pixi run -e py314 test-module tests/geometry/test_depth.py --device=cpu --dtype=float16 --xfail-known-failures --known-failure-profile=cpu-float16 -k 'require_two_spatial_axes or minimum_spatial_size_plane or keeps_the_w_one_axis_4278' -q
pixi run -e py314 test-module tests/geometry/test_depth.py --device=cpu --dtype=bfloat16 --xfail-known-failures --known-failure-profile=cpu-bfloat16 -k 'require_two_spatial_axes or minimum_spatial_size_plane or keeps_the_w_one_axis_4278' -q
pixi run -e py314 test-module --doctest-modules kornia/geometry/depth.py -q
pixi run -e py314 typecheck --python .venv-py314
```

Local validation did not cover GPU execution, Inductor code generation, or the full repository test suite.

## How did AI help?

Codex assisted with investigation, implementation, tests, documentation, and the checks above. The new regression tests were run before the fix to verify the failure. Half-precision plane-test data were checked against both upstream main and the patch and adjusted for representable coordinates without relaxing the repository's default tolerances.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -378,9 +378,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   unnormalize as an exact pixel index. Refs #4030. (#4231)
 
 ### Bug fixes
+
 * `depth_to_normals` now raises `ShapeError` when `H < 2` or `W < 2`, since surface normals need two
   tangent directions. Previously, singleton axes could yield zero or non-finite normals, and empty
-  spatial dimensions failed inside padding. Inputs with both dimensions at least 2 are unchanged. (PR pending)
+  spatial dimensions failed inside padding. Inputs with both dimensions at least 2 are unchanged. (#4458)
 
 * `RandAugment`'s `m` guard is exclusive at both ends, but its docstring and its error message
   both named the closed interval `[0, 30]`, so a user who asked for the maximum strength the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -378,6 +378,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   unnormalize as an exact pixel index. Refs #4030. (#4231)
 
 ### Bug fixes
+* `depth_to_normals` now raises `ShapeError` when `H < 2` or `W < 2`, since surface normals need two
+  tangent directions. Previously, singleton axes could yield zero or non-finite normals, and empty
+  spatial dimensions failed inside padding. Inputs with both dimensions at least 2 are unchanged. (PR pending)
 
 * `RandAugment`'s `m` guard is exclusive at both ends, but its docstring and its error message
   both named the closed interval `[0, 30]`, so a user who asked for the maximum strength the
--- a/kornia/geometry/depth.py
+++ b/kornia/geometry/depth.py
@@ -26,6 +26,7 @@ import torch.nn.functional as F
 from torch import nn
 
 from kornia.core.check import KORNIA_CHECK, KORNIA_CHECK_IS_TENSOR, KORNIA_CHECK_SHAPE
+from kornia.core.exceptions import ShapeError
 from kornia.filters.sobel import spatial_gradient
 from kornia.geometry.grid import create_meshgrid
 
@@ -263,9 +264,10 @@ def depth_to_3d(depth: torch.Tensor, camera_matrix: torch.Tensor, normalize_poin
 
 
 def depth_to_normals(depth: torch.Tensor, camera_matrix: torch.Tensor, normalize_points: bool = False) -> torch.Tensor:
-    """Compute the normal surface per pixel.
+    r"""Compute the normal surface per pixel.
 
     Convention:
+        - both spatial dimensions must be at least 2: a surface normal requires two tangent directions.
         - the normal is the cross product of the two spatial gradients of the unprojected point cloud, taken in
           the order ``d/dx`` cross ``d/dy`` and then normalized to unit length, so a fronto-parallel plane gets
           the normal ``(0, 0, 1)``: ``+z`` points **away** from the camera, along the viewing direction, and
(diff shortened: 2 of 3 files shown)


ducha-aiki (approved):
Reviewed exact head `685c90ba`.

The guard matches the disposition in #4398: `H < 2 or W < 2` raises `ShapeError` with the actual dimensions, placed right after the existing `KORNIA_CHECK_SHAPE` calls and before `depth_to_3d_v2`, so nothing is allocated first. `depth_to_3d`, `depth_to_3d_v2` and `warp_frame_depth` are correct at `W = 1` (the retained #4278 pin covers them) and are rightly out of scope. The valid path is byte-identical to main: 16 configurations (float32/float64, four asymmetric shapes, both `normalize_points`) give `torch.equal` outputs and gradients. `Fixes #4398` is right for a single-symbol issue; the documented-defect grep is empty in `kornia/`; the release note cites #4458 with old and new behaviour.

Compile: `dynamo.explain` on `depth_to_normals` reports one graph and zero breaks, and `fullgraph=True` works both static and `dynamic=True`; this PR adds the first dynamo test for the module, and all four new nodes pass under inductor.

Validation (isolated worktree, import origin verified): all 24 `test_convention_normals_require_two_spatial_axes_4398` nodes fail on `8fb9ad12` and pass here; the 24-node valid-size plane test is a value control with asymmetric K and H != W. `tests/geometry/test_depth.py` CPU float32/float64 with `KORNIA_TEST_OPTIMIZER=inductor` 304 passed, 2 xfailed; cpu-float16 and cpu-bfloat16 profiles 35 passed, 1 pre-existing xfail each; MPS float32 35 passed, 1 skipped; ruff clean. Not run: full suite, CUDA.

Unrelated and pre-existing, not for this PR: `ShapeError.__init__` is not dynamo-traceable, so under `fullgraph=True` any `ShapeError` raise (this guard and every `KORNIA_CHECK_SHAPE`) surfaces as `torch._dynamo.exc.Unsupported`. Identical on main.

PR #4459, opened 2026-09-12
fix(augmentation): correct RandomMosaic H/W axes

### LINKED ISSUE #4438: W1b: `RandomMosaic` returns `(W, H)` on a non-square input, and the same axis swap reaches `start_ratio_range`

Hrithikcrick:

Fixes #4438 by correcting the height/width axis ordering in `RandomMosaic`.

For non-square inputs with `output_size=None`, the augmentation now preserves `(H, W)` instead of returning `(W, H)`.

`start_ratio_range` now scales x by width and y by height, and the source crop passes width and height to `bbox_generator` in the correct order.

## Regression coverage

- Non-square batched input preserves `(H, W)`.
- `start_ratio_range=(0.5, 0.5)` on `H=6, W=8` produces `(x, y) = (4, 3)`.
- Unbatched `(C, H, W)` input is covered with both `keepdim=False` and `keepdim=True`.
- Full `test_augmentation_mix.py`: 53 passed.
- Pre-commit: passed.
- `git diff --check`: clean.

## Related issue

Fixes #4438

Related to #4407.

## Note

#4454 was opened while this fix was being prepared. This PR keeps the scope limited to the #4438 axis-order defect and includes the asymmetric and unbatched regression coverage requested for this behavior.


ducha-aiki (changes requested):
Reviewed exact head `968f208b`.

The generator fix is right and complete for every surface #4438 lists: `(4, 3, 6, 8)` and `(4, 3, 8, 6)` inputs keep their shape with `output_size=None`, `start_ratio_range=(0.5, 0.5)` on 6x8 gives top-left `(4, 3)` where main gives `(3, 4)`, the unbatched path keeps `(1, 6, 8)` / `(1, 1, 6, 8)` for both `keepdim` values, and a pixel-encoded batch of three distinct images matched the `src`/`permutation`-predicted source pixel at every output location for both aspect ratios. All four new nodes fail on main and pass here. `Fixes #4438` is the only closing reference and #4407 is "Related to", which is right.

One change requested:

- [P2] `CHANGELOG.md:160`: the entry is under a newly invented `### Fixed` heading inside `## Unreleased`, which already has a `### Bug fixes` section (around line 384). Move it there, and say what used to happen (the default output came back as `(W, H)` and `start_ratio_range` scaled x by H and y by W), since #4438 classifies this as a behaviour change.

Non-blocking:

- [P3] The three new tests take no `device`/`dtype` fixtures, so they only ever run on CPU float32; one uses `torch.testing.assert_close` instead of `self.assert_close`, and two are shape-only. Taking the fixtures is the house convention. If you do, note that the generator's `src` is emitted in the generator's own float32, so compare in that dtype rather than the fixture's.

Duplicate: #4454 fixes the same issue with a byte-identical `mosaic.py` hunk, but bundles unrelated `pyproject.toml`/`uv.lock` changes and an out-of-scope `mix/base.py` change, and its fixture-parametrized tests are red on float64, the half profiles and MPS. This PR is the one to merge once the release note is moved; I have asked the other author to close #4454 as superseded.

Validation (isolated worktree, import origin verified): mosaic selection CPU float32/float64 13 passed; cpu-float16 profile 8 passed, 1 xfailed; cpu-bfloat16 5 passed, 4 xfailed (existing manifest lines, no new reds); MPS float32 9 passed; ruff clean; all 38 hosted checks green. Not run: full suite, CUDA.

Two pre-existing defects, identical on main and not for this PR, which I will track separately: `mix/mosaic.py:134-135` offsets bbox tiles with `H*i` for x and `W*j` for y while `_compose_images` places tile `(i, j)` at `x = W*i, y = H*j`, so on non-square inputs the boxes land in the wrong place; and every mix augmentation raises `KeyError: 'BFLOAT16'` on bfloat16 input because `DType` in `kornia/constants.py` has no such member.

Hrithikcrick:
Thanks for the detailed review!

I've addressed the requested P2 changelog change in the latest head `17a2fe1b`:

- moved the `RandomMosaic` entry into the existing `### Bug fixes` section
- documented that non-square inputs previously returned `(W, H)`
- documented that `start_ratio_range` previously scaled x by height and y by width
- rebased onto the latest `main` and resolved the `CHANGELOG.md` conflict while preserving the newer upstream entries

Local `test_augmentation_mix.py`: 53 passed.
All 40 hosted checks are now green.

The P3 test-fixture suggestion is noted as non-blocking, so I've kept this update scoped to the requested change.

Whenever convenient, could you please take another look at the latest head?

Thanks again!

ducha-aiki (dismissed):
Re-reviewed exact head `17a2fe1b` (rebase of `968f208b`).

The `mosaic.py` generator hunk and the three regression tests are byte-identical to the previously reviewed head; only the release note moved. The P2 is resolved: the entry now sits under the existing `### Bug fixes` heading in `## Unreleased` (`CHANGELOG.md:384-386`) and states both old behaviours (`(W, H)` output and `start_ratio_range` scaling x by H / y by W). The invented `### Fixed` heading is gone.

Re-verified independently on this head with a pixel-encoded batch of three images on `(4, 3, 6, 8)` and `(4, 3, 8, 6)`: the output keeps the input shape, every output pixel matches the `permutation`/`src`-predicted source pixel (0 mismatches, CPU float32 and MPS float32), and `start_ratio_range=(0.5, 0.5)` gives top-left `(4, 3)` on 6x8 and `(3, 4)` on 8x6. On `main` the same probe returns the swapped shape for every case. All four new nodes fail on `main` and pass here.

Validation (isolated worktree, import origin verified): mosaic selection CPU float32/float64 13 passed; cpu-float16 profile 8 passed, 1 xfailed; cpu-bfloat16 profile 5 passed, 4 xfailed (existing manifest lines, no new reds); MPS float32 9 passed; full `test_augmentation_mix.py` CPU float32 53 passed; pinned ruff clean; 40/40 hosted checks green. Not run: CUDA, full suite.

The P3 (device/dtype fixtures, `self.assert_close`) was left as-is, which is fine; it stays a nit.

Out of scope, for the record: the bounding-box tile offsets in `mix/mosaic.py:134-135` are still swapped on non-square input; that is tracked separately in #4468 and is not this PR's job.

Approving. This supersedes #4454, which is closed.

PR #4453, opened 2026-09-11
fix(augmentation): sample same number of drops in RandomRain when same_on_batch=True (#4448)

topqaz:
## What does this pull request do?

Fixes `RandomRain` with `same_on_batch=True` to draw the same number of rain drops across all samples in the batch.
Previously, `RainGenerator.forward` omitted `same_on_batch` when calling `_adapted_rsampling` for `number_of_drops_factor`, causing each sample in the batch to receive a different drop count despite `same_on_batch=True`.

## Related issue or discussion

Fixes #4448

## What did you check?

Added `test_same_on_batch` in `TestRandomRain` to verify that `number_of_drops_factor`, `drop_height_factor`, `drop_width_factor`, `coordinates_factor`, and batch outputs are identical across the batch when `same_on_batch=True`.

```text
$ pytest tests/augmentation/test_augmentation.py -k "TestRandomRain"
10 passed, 395 deselected in 4.23s
$ ruff check kornia/augmentation/random_generator/_2d/random_rain.py tests/augmentation/test_augmentation.py
All checks passed!
$ ruff format --check kornia/augmentation/random_generator/_2d/random_rain.py tests/augmentation/test_augmentation.py
2 files already formatted

PR #4460, opened 2026-09-12
fix(augmentation): preserve patch coverage and batch shape in PatchSequential

### LINKED ISSUE #4421: W1b: `PatchSequential` augments `B x C` patch rows, and its `padding` modes change image and batch size
`PatchSequential` extracts a `grid_size` grid of patches from each image and applies its augmentations patch-wise. Its `forward` hands the **image** shape to `forward_parameters` (`kornia/augmentation/container/patch.py:559`), and `forward_parameters` reads `batch_shape[0] * batch_shape[1]` as the number of patch rows (`patch.py:288`). For an image batch those two axes are `B` and `C`, not `B` and `n_patches`, so the container draws and applies parameters for `B * C` rows of a tensor that has `B * n_patches` of them.

Where `B * C` exceeds the patch-row count the call dies with a raw `IndexError` from the indexing, which is why a `1 x 1` grid works on a single-channel image and crashes on a three-channel one. Where it is smaller the call succeeds and augments only the leading rows: on a `2 x 2` grid with `B = 2, C = 3` the last two patches of the second sample are left untouched and nothing reports it. The shipped docstring example uses `grid_size=(2, 2)` with a 3-channel input, which is exactly the silent-partial case.

The `padding` argument in the same file has a second axis problem. `patch.py:53-54` documents `same` as padding so that every pixel is covered and `valid` as trimming the redundant border. Executed, `same` shrinks the image and `valid` leaves it unchanged — and on a non-square image `valid` with a `4 x 4` grid returns a batch of 3 for an input batch of 2, so a container that is documented to preserve the batch changes it.

## Repro

```python
import torch
from kornia.augmentation import PatchSequential
from kornia.contrib import extract_tensor_patches

torch.manual_seed(42)
x = torch.rand(2, 3, 8, 8)          # B = 2, C = 3, an 8 x 8 image
x68 = torch.rand(2, 3, 6, 8)        # B = 2, C = 3, a non-square 6 x 8 image

# 1. a 1 x 1 grid is one patch per sample, but the row count is B * C
for c in (1, 2, 3):
    try:
        out = PatchSequential(K.RandomHorizontalFlip(p=1.0), grid_size=(1, 1), patchwise_apply=False)(x[:, :c])
        print(f"grid (1,1), C={c}: out {tuple(out.shape)}")
    except Exception as e:
        print(f"grid (1,1), C={c}: {type(e).__name__}: {e}")

# 2. where it does not crash it augments only the first B*C of B*n_patches rows
out = PatchSequential(K.RandomHorizontalFlip(p=1.0), grid_size=(2, 2), patchwise_apply=False)(x)
pin = extract_tensor_patches(x, window_size=(4, 4), stride=(4, 4))
pout = extract_tensor_patches(out, window_size=(4, 4), stride=(4, 4))
flipped = [bool(torch.allclose(pout[b, i], torch.flip(pin[b, i], dims=[-1]))) for b in range(2) for i in range(4)]
print("grid (2,2), B=2, C=3: per-patch flipped:", flipped)
print("  patch rows the op touched:", sum(flipped), "of", len(flipped), "(B*C =", 2 * 3, ")")

# 3. padding changes the image size and the batch size
for grid, pad, src in (((3, 3), "same", x), ((3, 3), "valid", x), ((4, 4), "valid", x68), ((5, 5), "valid", x68)):
    try:
        out = PatchSequential(K.RandomHorizontalFlip(p=1.0), grid_size=grid, padding=pad, patchwise_apply=False)(src)
        print(f"grid {grid} padding={pad!r} on {tuple(src.shape)} -> {tuple(out.shape)}")
    except Exception as e:
        print(f"grid {grid} padding={pad!r} on {tuple(src.shape)} -> {type(e).__name__}: {e}")
```

## Observed

```
grid (1,1), C=1: out (2, 1, 8, 8)
grid (1,1), C=2: IndexError: index 2 is out of bounds for dimension 0 with size 2
grid (1,1), C=3: IndexError: index 2 is out of bounds for dimension 0 with size 2
grid (2,2), B=2, C=3: per-patch flipped: [True, True, True, True, True, True, False, False]
  patch rows the op touched: 6 of 8 (B*C = 6 )
grid (3, 3) padding='same' on (2, 3, 8, 8) -> (2, 3, 7, 7)
grid (3, 3) padding='valid' on (2, 3, 8, 8) -> (2, 3, 8, 8)
grid (4, 4) padding='valid' on (2, 3, 6, 8) -> (3, 3, 6, 6)
grid (5, 5) padding='valid' on (2, 3, 6, 8) -> ZeroDivisionError: integer modulo by zero
```

`patchwise_apply=True` under-applies further — 4 of the 8 patch rows are touched on the same fixture. On MPS the crash is an `AcceleratorError` carrying the same message and a device-scheduling-dependent index.

## Expected

`forward` passes the extracted-patch shape `(B, n_patches, C, h, w)` to `forward_parameters`, so the drawn row count is `B * n_patches` and every patch of every sample is augmented. `padding="same"` preserves or grows the spatial size and `padding="valid"` trims it, as `patch.py:53-54` says; neither changes the batch dimension. A `grid_size` that does not divide the input is rejected with a kornia message rather than a `ZeroDivisionError` or a raw `reshape` error.

## Classification

**Broken output** — silent partial augmentation on the grids that work, a raw `IndexError` on the grids that do not, and a batch size that depends on the padding mode. Idiom: 7 (an axis-order slip hidden by symmetric test data — a square grid with `C == n_patches` conceals it, and the class's own docstring example is the silent-partial configuration).

## Disposition

`n/a — single symbol`. Both halves live in `kornia/augmentation/container/patch.py` and can be fixed together. Batch 6 documents the current behavior on `PatchSequential` and pins it; the repair is a focused fix, welcome as a PR.

## Environment

NaCr05:

`PatchSequential` could leave trailing patches unaugmented, skip later modules, or change the number of images during reconstruction. For example, a two-image RGB batch on a 2x2 grid only flipped 6 of its 8 patches. Parameters now use the extracted `(B, N, C, h, w)` layout, fixed-location modules cover every sample, and the complete selected sequence runs on every applicable patch.

`forward_parameters` accepts both `(B, C, H, W)` image shapes and `(B, N, C, h, w)` patch shapes. A shared shape-only padding calculation derives patch dimensions without allocating an image. The usual replay idiom remains supported:

```python
params = seq.forward_parameters(x.shape)
out = seq(x, params=params)
```

Padding has an explicit contract: `same` temporarily zero-pads to the grid and removes the added border, preserving H/W; `valid` crops the remainder and keeps the cropped output. Both preserve B/C. Odd padding/cropping goes to the right/bottom. Geometric children may move the temporary zero border into the retained image, as documented and tested. Reconstruction checks the patch count and preserves the explicit batch dimension; patch write-back retains the input dtype when a mix operation promotes to its parameter dtype.

### Behaviour changes

The changelog's **Breaking changes** section records these examples as `(B,C,H,W)`:

| Input / grid / mode | Before | After |
| --- | --- | --- |
| `(2,3,8,8)` / `(3,3)` / `same` | `(2,3,7,7)` | `(2,3,8,8)` |
| `(2,3,8,8)` / `(3,3)` / `valid` | `(2,3,8,8)` | `(2,3,6,6)` |
| `(2,3,6,8)` / `(4,4)` / `valid` | `(3,3,6,6)` | `(2,3,4,8)` |

`restore_from_patches` defaults to `self.grid_size` instead of `(4,4)` and raises `ValueError` for a mismatched patch count. Parameters cover `B*N` patch rows instead of `B*C`, so fixed-seed outputs and RNG consumption can change. A three-dimensional image still raises `ValueError`.

Regression coverage includes per-patch values, operation order, parameter sharing, four-dimensional/five-dimensional parameter equivalence, replay without re-sampling, asymmetric padding, invalid shapes/grids/counts, gradients, non-contiguous inputs, and mixed parameter/image precision. The old integration test's invalid `K.color` reference and broad exception handler silently bypassed execution. It now executes supported configurations, includes both `(2,3,24,24)` and non-divisible `(2,3,23,25)` inputs, and checks the full output shape for each padding mode. Nested colour operations use brightness/contrast-only `ColorJiggle` to avoid coupling container replay to HSV's float16 black-pixel NaNs. CPU RNG state is restored after tests.

Fixes #4421.

Related to #4452, still open at inspected head `f44693f6`. Its convention pins were reconciled with this repair in an isolated checkout. The combined PatchSequential and augmentation-base files passed **945 tests, with 360 skips and 2 existing xfails**, including the complete RNG fingerprint test.

Whichever PR lands second needs to reconcile:

- Remove the strict xfail from `test_convention_patch_sequential_augments_every_patch_row`.
- Replace the B*C wart assertions with successful RGB 1x1-grid execution and all 12/8 patches transformed on 2x3/2x2 grids.
- Replace padding wart expectations with `(2,1,8,8)` / `(2,1,6,8)` for its `same` cases, and `(2,1,6,6)` / `(2,1,4,8)` / `(2,1,6,6)` for its `valid` cases.
- Update `_RNG_FINGERPRINTS["PatchSequential"]` in `tests/augmentation/test_base.py` from `[("rand", (12,))]` to `[("rand", (16,))]`: its four images on a 2x2 grid need 16 angle draws. Keep the existing four-dimensional `(4,3,6,8)` input shape. The old 12-draw pin fails with this repair; the updated pin passes.
- Keep the repaired padding/coverage Convention prose, remove the obsolete #4421 warning, and retain #4452's shared-contract reference and image-only API note.

Non-divisible images are supported through padding/cropping; empty `valid` patches and invalid grids are rejected. Broader parent-container nesting, annotation/geometric inverse support, and the pre-existing empty-batch reshape error remain outside this fix.

## What did you check?

Windows CPU, Python 3.14.6, PyTorch 2.14.0+cpu, using the Pixi py314 environment. Worktree import origins and the interpreter were verified.

| Check | Result |
| --- | --- |
| All `tests/augmentation/container/`, float32/float64, including slow tests | 1,368 passed, 364 skipped |
| PatchSequential, float16, existing CPU-half profile | 424 passed, 182 skipped, 1 existing xfail |
| PatchSequential, bfloat16, existing CPU-half profile | 425 passed, 182 skipped |
| `patch.py` doctests, including four-dimensional replay | 4 passed |
| Source typecheck | Passed; one existing LightGlue deprecation warning |
| Full `pre-commit-all` | Passed |
| #4452 integration: PatchSequential and augmentation-base tests, float32/float64, including slow | 945 passed, 360 skipped, 2 existing xfails |

Initial numerical regressions produced 54 failures on main `8fb9ad1`; both MixUp half-dtype cases failed before the write-back repair. The review follow-up tests produced **31 failures and 3 passing controls** on the reviewed head `fd51449d`; all 34 then passed after the compatibility/error-type changes.

A CPU Dynamo probe compared `fd51449d` with this revision in separate worktrees. In both padding modes, replay through `seq.forward` captured one graph with zero breaks and matched eager output under `torch.compile(..., backend="eager", fullgraph=True)`. Compiling the container with automatic parameter generation also matched eager output, retaining the same three graph breaks on both revisions. This probe used the eager backend and did not validate Inductor.

The expanded legacy matrix explicitly skips 360 unsupported combinations in the float32/float64 runs. No known-failure entries were added or changed. Local half-profile metadata differs from the recorded Linux/Python 3.11/PyTorch 2.9.1 baseline. CUDA, MPS and Inductor were not run locally for this follow-up.

## How did AI help?

OpenAI Codex inspected the implementation and review, drafted and revised the tests and repair, ran the reported checks, and prepared this PR. Numerical expectations use independent slices or explicit tensors, and the parameter-entry-point regression supplies patch dimensions independently. Compatibility and before/after compilation checks used separate worktrees.


NaCr05:
Could a maintainer please rerun the failed jobs for this PR?

The [macOS / Python 3.12 / PyTorch 2.5.1 job](https://github.com/kornia/kornia/actions/runs/34674277959/job/103501287026) timed out after 120 seconds in `TestSuperResolutionBuilders::test_to_onnx_exports[RRDBNetBuilder]`, causing the collector to fail. The other 38 checks passed.

I tried rerunning the failed jobs, but GitHub denied the request due to repository permissions. Thanks!

ducha-aiki (changes requested):
Reviewed exact head `fd51449d`.

The fix is correct and I could reproduce every #4421 claim on main and its repair here: a 2-image RGB batch on a 2x2 grid with `RandomHorizontalFlip(p=1.0)` flips 6 of 8 patches on main (4 of 8 with `patchwise_apply=True`) and 8 of 8 here; the `(1, 1)` grid `IndexError` is gone; `same`/`valid` padding now preserve H/W and B respectively; 21 `random_apply x patchwise_apply x same_on_batch` combinations replay bitwise with the CPU RNG state untouched; float16/bfloat16 MixUp write-back works; `dynamo.explain` shows no graph breaks; and the two documented ONNX export cases match ORT to 0.0 on both trees. The old integration test really did bypass execution (`hasattr(kornia.augmentation, "color")` is `False`, so every parametrization returned inside the broad `except`).

What blocks merge is the bookkeeping of a behaviour change, not the code:

- [P2] `CHANGELOG.md:382-385`: the entry sits under `### Bug fixes` and only states the new behaviour, but outputs change for inputs that used to work. Verified old -> new on the same fixtures: `same`/(3,3) on `(2,3,8,8)` gave `(2,3,7,7)`, now `(2,3,8,8)`; `valid`/(3,3) on `(2,3,8,8)` returned the image unchanged, now `(2,3,6,6)`; `valid`/(4,4) on `(2,3,6,8)` gave `(3,3,6,6)` (batch 2 -> 3), now `(2,3,4,8)`. Two further API changes are unrecorded: `restore_from_patches` default `grid_size` changed from `(4, 4)` to `self.grid_size` and now raises on a wrong patch count, and `forward_parameters` rejects a 4-D shape (next item). Move the shape/batch contract to `### Breaking changes` with the old shapes, and list the other two.
- [P2] `patch.py:338-339`: the generic container replay idiom `params = seq.forward_parameters(x.shape); seq(x, params=params)` worked on main and now raises `ValueError: Expected patch shape (B, rows * columns, C, h, w). Got torch.Size([2, 3, 8, 8])`. Users cannot easily compute the 5-D shape themselves because `h, w` depend on `compute_padding`, which takes a tensor. Since 4-D and 5-D are unambiguous, accept a `(B, C, H, W)` shape and derive the padded patch shape from it; if you prefer the strict 5-D contract, record it as a breaking change and show the supported replay in the `forward_parameters` docstring.

Non-blocking:

- [P3] A 3-D input now raises `ShapeError`, which is not a `ValueError` subclass, where main raised `ValueError` from `extract_tensor_patches`; either keep `ValueError` like the file's other new guards or mention it in the release note.
- [P3] `same` padding lets a geometric child pull the temporary zero border into the kept image (`test_padding_with_flip` pins `[3, 2, 1, 0, 5]` for a 3x5 input). That is the documented mode's correct consequence, but the `padding` docstring should say so, or users will report it.
- [P3] The rewritten `test_forward` passes on main unchanged because its only fixture shape `(2, 3, 24, 24)` divides the grid, so it cannot observe either defect; fine as smoke coverage, but the description's "checks the batch dimension" oversells it. A non-divisible shape such as `(2, 3, 23, 25)` with a per-padding expected shape would give it teeth.
- [P3] `B = 0` still dies with a raw reshape `RuntimeError` on both trees; pre-existing, noted so the follow-up is not lost.

Validation (isolated worktree, import origin verified): `test_patch_sequential.py` CPU float32/float64 with `--runslow` 444 passed; cpu-float16 profile 228 passed, 1 pre-existing xfail; cpu-bfloat16 229 passed; MPS float32 228 passed; 57 nodes of the new regression class fail on `8fb9ad12`; ruff clean. Not run: full suite, CUDA.

Maintainer note: our open batch-6a docs PR #4452 pins the old behaviour in three PatchSequential tests plus the `_RNG_FINGERPRINTS["PatchSequential"]` entry in `tests/augmentation/test_base.py`, which calls `forward_parameters((4, 3, 6, 8))` and would raise here; whichever lands second reconciles those and the `patch.py` warning prose.

ducha-aiki (dismissed):
Re-reviewed exact head `558e30a6` (delta over `fd51449d`: "retain image-shape parameter replay") in an isolated worktree with import origin verified.

Both blocking asks are resolved:

- CHANGELOG: `### Breaking changes` under `## Unreleased` now records the old-vs-new shapes for `same`/(3,3) on (2,3,8,8) (`(2,3,7,7)` -> `(2,3,8,8)`), `valid`/(3,3) (`(2,3,8,8)` -> `(2,3,6,6)`), `valid`/(4,4) on (2,3,6,8) (`(3,3,6,6)` -> `(2,3,4,8)`), the `restore_from_patches` default and count check, the dual `forward_parameters` shapes, and the B*N vs B*C RNG-consumption change; the Bug-fixes entry cross-references it. I re-measured each old/new value on `origin/main` and here.
- `forward_parameters(image.shape)`: accepted again. Across all 24 `random_apply x patchwise_apply x same_on_batch x padding` combinations the 4-D and 5-D shapes produce identical params (indices and every data key, atol=rtol=0), leave the CPU RNG in the same state, replay without drawing, and give bitwise-equal outputs; `fd51449d` still raises on the same call. The doctest example passes.

Non-blocking follow-ups from last round are also done: 3-D inputs raise `ValueError`, the padding docstring names the zero-border movement, and `test_forward` gained `(2, 3, 23, 25)`.

Remaining nits (no action required for merge):

- [P3] `tests/augmentation/container/test_patch_sequential.py:63-95`: the `valid` branch of `test_forward` still passes on `main` because main's crop-6/pad-6 path also yields `(2, 3, 22, 24)` (with a 12-pixel zero border). Asserting the border is non-zero for a `torch.rand` input would make it discriminate; the regression class already pins the real behaviour, so this is cosmetic.
- The PR was CONFLICTING on `CHANGELOG.md` only (main's #4453/#4451/#4458 entries sit in the same hunk); `patch.py` and the test file merge cleanly. I pushed that keep-both merge as `dc9f01a1` (every other path tree-identical to `git merge-tree`) and this approval is on that commit.
- CI on this head: 36 green; the `float16` half-precision leg was cancelled (known runner hang) so `collector` shows red - it needs a rerun, not a change. Locally the cpu-float16 profile is 424 passed / 1 pre-existing xfail and cpu-bfloat16 425 passed.

Validation here: `test_patch_sequential.py` CPU float32/float64 `--runslow` 831 passed; cpu-float16 / cpu-bfloat16 profiles green; MPS float32 424 passed; all eight #4421 claims reproduce on `main` and hold here; 35 of the delta-new nodes fail on `main` with the head test file; `dynamo.explain` 1 graph / 0 breaks and `torch.compile(seq.forward, fullgraph=True)` equal to eager; ONNX (`dynamo=True`) matches ORT to 0.0 on four export cases, identical to the previous head; ruff clean. Not run: CUDA, full suite.

Maintainer note: #4452's three PatchSequential pins, its strict xfail, the `patch.py` "Tracked in #4421" warning/Convention prose, and `_RNG_FINGERPRINTS["PatchSequential"]` (`(12,)` -> `(16,)`, verified on this head with `(4, 3, 6, 8)`) all need the reconciliation the PR body lists, whichever lands second. #4462 only collides on CHANGELOG and its plasma test passes on this head.

PR #4462, opened 2026-09-12
fix(augmentation): make RandomPlasma params replay deterministic

### LINKED ISSUE #4445: W1b: `RandomPlasma*` do not replay bitwise from `params=` — the fractal is drawn in `apply_transform`
`forward(x, params=aug._params)` replays a previous draw bitwise on every public augmentation class except the three `RandomPlasma*` classes. Their `_params` hold only the sampled `roughness` / `intensity` / `shade_*` factors; the plasma fractal itself is drawn inside `apply_transform` from torch's global generator on every call, so replaying with the stored parameters gives a different image unless the global seed is reset to the same value first. The class docstrings describe `params` the same way as every other class and do not say that the replay is partial.

## Repro

```python

torch.manual_seed(0)
x = torch.rand(2, 3, 6, 8)
for cls in (K.RandomPlasmaBrightness, K.RandomPlasmaContrast, K.RandomPlasmaShadow, K.RandomBrightness):
    aug = cls(p=1.0)
    y1 = aug(x)
    y2 = aug(x, params=aug._params)
    torch.manual_seed(3); y3 = aug(x, params=aug._params)
    torch.manual_seed(3); y4 = aug(x, params=aug._params)
    print(f"{cls.__name__:24s} replay bitwise={torch.equal(y1, y2)}  same-global-seed replays equal={torch.equal(y3, y4)}  params={sorted(aug._params)}")
```

## Observed

```
RandomPlasmaBrightness   replay bitwise=False  same-global-seed replays equal=True  params=['batch_prob', 'forward_input_shape', 'intensity', 'roughness']
RandomPlasmaContrast     replay bitwise=False  same-global-seed replays equal=True  params=['batch_prob', 'forward_input_shape', 'roughness']
RandomPlasmaShadow       replay bitwise=False  same-global-seed replays equal=True  params=['batch_prob', 'forward_input_shape', 'roughness', 'shade_intensity', 'shade_quantity']
RandomBrightness         replay bitwise=True  same-global-seed replays equal=True  params=['batch_prob', 'brightness_factor', 'forward_input_shape']
```

## Expected

Either the fractal field is drawn in `generate_parameters` and stored in `_params` (so `params=` replays bitwise like every other class, at the cost of a larger parameter dict), or the three classes document that `params=` restores the factors only and that a bitwise replay additionally needs the global seed. The first outcome matches the contract the anchor documents for the rest of the package; the audit recommends it.

## Classification

**Semantic contract** — the `params=` replay contract holds on 66 of 69 concrete classes and silently does not on these three. Idiom: 5 (a documented precondition never validated: the docstring's `params` contract is not honoured).

## Disposition

`n/a — single symbol` (one family, three classes sharing one generator path). The batch documents the exception on the anchor and pins it as a wart; moving the draw into `generate_parameters` is a focused fix, welcome as a PR.

## Environment

Hrithikcrick:

Fixes #4445 by making `RandomPlasmaBrightness`, `RandomPlasmaContrast`, and `RandomPlasmaShadow` replay bitwise from stored `params=`.

The plasma fractal is now generated during parameter generation and stored in `_params`, so `apply_transform` no longer draws a new fractal from the global RNG on every replay.

## Regression coverage

- Covers all three `RandomPlasma*` augmentations.
- Changes the global RNG state before replay to verify that stored params are sufficient for deterministic output.
- `TestRandomPlasma --dtype=all`: 24 passed.
- Plasma doctests: 3 passed.
- `tests/augmentation/test_augmentation.py --dtype=float32,float64`: 628 passed, 92 skipped, 21 xfailed, 7 xpassed.
- Pre-commit: passed.
- `git diff --check`: clean.

## Note

Opening as a draft only to add the Unreleased changelog entry with the final PR number before marking it ready for review.


ducha-aiki (changes requested):
Reviewed exact head `a11da8cd`.

The replay fix works: the fractal is drawn in `generate_parameters` and stored as `params["plasma"]`, and the #4445 repro is now bitwise for all three classes, on float32/float64/float16/bfloat16 and MPS, through `AugmentationSequential` and `VideoSequential`, and under `torch.compile(fullgraph=True)`. Partial-batch rows (`p=0.5`, B=3) match single-sample replays exactly. `Fixes #4445` is right, the release note cites #4462, and the new pin fails on main.

One change requested:

- [P2] `kornia/augmentation/_2d/intensity/plasma.py:70,128,192`: `B, C, H, W = shape` regresses `PatchSequential` for all three classes. `PatchSequential.forward` hands children a three-element `(B*C, H, W)` shape (the #4421 defect). Main's plasma code ignored the shape and drew from `image.shape` in `apply_transform`, so it ran; on this head every mode crashes. Reproduced from the worktree root:

  ```python
  K.PatchSequential(K.RandomPlasmaBrightness(p=1.0), grid_size=(2, 2), patchwise_apply=False)(torch.rand(2, 3, 8, 8))
  # main: (2, 3, 8, 8)
  # head: ValueError: not enough values to unpack (expected 4, got 3)
  ```

  Same for Contrast and Shadow, and for `patchwise_apply=True` with either `random_apply`. Nothing in `test_patch_sequential.py` covers plasma. Either draw the fractal in a way that tolerates the container's shape (and add a `PatchSequential(RandomPlasma*)` pin), or wait for the #4421 fix in #4460, which changes what the container passes, and re-check on top of it.

Non-blocking:

- [P3] `same_on_batch=True` still draws an independent fractal per sample (pre-existing, but the draw now lives where `same_on_batch` is honoured for the roughness factor, so a `(1, C, H, W)` draw expanded to `B` would close it cheaply).
- [P3] The CHANGELOG entry is a single 141-character line; wrap it and say what used to happen (the fractal was redrawn from the global RNG on every call).

Notes: the fractal now follows the generator's device/dtype and is cast with `.to(image)`, the same convention as the other generator draws, so a float64 input without `.to(torch.float64)` gets a float32-quantised fractal; and with `p < 1` the head draws `B` fractals where main drew only the selected rows, so seeded outputs under `p < 1` differ from previous releases. Both acceptable for a bug fix, worth a sentence in the release note.

Validation (isolated worktree, import origin verified): the 6 new nodes fail on `8fb9ad12` and pass here; plasma selection CPU float32/float64 passed, cpu-float16 and cpu-bfloat16 profiles 6 passed each, MPS float32 6 passed, doctests 3 passed, ruff clean. Not run: full suite, CUDA.

ducha-aiki (dismissed):
Approving on `60d569c3`, a pure merge of `origin/main` (5be4852b) into head `ca65e18b` that I pushed to clear the `CHANGELOG.md` conflict (every other path tree-identical to `git merge-tree`; the CHANGELOG lines added over main are the branch's own), so the CI can run.

Reviewed exact head `ca65e18b` (delta over `a11da8cd`).

The P2 from the previous round is fixed. `_generate_plasma` now accepts the three-element shapes `PatchSequential` hands its children today, and `_plasma_for_input` crops the stored fractal to the patch, so every container mode runs and replays bitwise. Verified from an isolated worktree with an independent probe: 3 classes x `patchwise_apply` x `random_apply` x `same_on_batch` x `padding` = 48 combinations, all return the input shape and replay bitwise from `params=` (main: 24 replay-different, 24 crash on `padding="valid"`); the new `test_patch_sequential_plasma.py` fails on `a11da8cd` with the old `ValueError` and passes here. The core fix is unchanged: `params=` / `forward_parameters` replay is bitwise for all three classes on CPU float32/float64/float16/bfloat16, MPS float32, through `AugmentationSequential`/`VideoSequential`, and under `torch.compile(fullgraph=True)`; `test_params_replay_4445` fails on main. A side benefit: `VideoSequential(same_on_frame=True)` now really applies the same fractal to every frame (main drew a fresh one per frame). Also checked the tree with #4460's `patch.py` on top: 48/48 and all tests pass, so merge order between the two does not matter.

Approving. Two housekeeping items before merge:

- The PR is conflicting on `CHANGELOG.md` only (main added the #4453/#4451/#4458 entries at the same `### Bug fixes` insertion point; `tests/augmentation/test_augmentation.py` merges cleanly). I pushed that merge; while you are there: the entry is a single 141-character line; wrap at 120 and say what used to happen (the fractal was redrawn from the global RNG on every call, so `params=` replay was not reproducible).
- Collision with #4452: its `test_wart_random_plasma_replay_4445` asserts replay is *not* bitwise and its `kornia/augmentation/_2d/base.py` prose says the same, "tracked in #4445". Whichever of the two merges second has to drop that pin and rewrite the sentence (RandomDissolving remains the only sample-during-apply exception); I will handle it on the #4452 side if this lands first.

Non-blocking, pre-existing: `same_on_batch=True` still draws an independent fractal per sample; and `kornia.contrib.diamond_square` fails for spatial sizes below 3 px (`(2,2)` raises `TypeError: rand(): ... got float`), which is why main's `PatchSequential(RandomPlasma*, padding="valid")` dies on a 9x10 image. This head dodges it only because the fractal is drawn at the full-image size the container currently passes; after #4460 small patches will hit it again. That is a `diamond_square` bug and deserves its own issue.

Validation: new tests CPU float32/float64 21 passed, cpu-float16 and cpu-bfloat16 profiles 15 passed each, MPS float32 15 passed; ruff 0.16.6 clean. Not run: full suite, CUDA.

PR #4463, opened 2026-09-12
fix(augmentation): accept zero scale in RandomThinPlateSpline

### LINKED ISSUE #4416: W1b: `RandomThinPlateSpline(scale=0.0)` raises a raw torch distribution error
`RandomThinPlateSpline` displaces five fixed control points by a uniform draw over `(-scale, scale)`. At `scale=0.0` that interval is empty, and the `torch.distributions.Uniform` constructed from it rejects its own arguments. The caller sees a message about a `LessThan(upper_bound=0.0)` constraint and a parameter named `low`, neither of which appears in the class's signature or docstring.

`scale=0.0` is not a nonsense value. It is the obvious way to express "no warp" — the setting a user reaches for when disabling an augmentation in a sweep, when annealing a magnitude to zero, or when writing a baseline against which the warped run is compared. The smallest non-zero scale works and produces the expected control-point offset, so the failure is a boundary case, not a general one.

The repro below also runs the seven other magnitude-driven 2D geometric classes at their own zero magnitude, each at its own default `align_corners`. None of them raises. Five return the input unchanged; `RandomPerspective` and `RandomElasticTransform` return a displaced image, which is the separate identity-warp defect tracked in #3904 and filed for this batch alongside this issue, not a second reading of what a zero magnitude means. `RandomThinPlateSpline` is the only one of the eight that refuses the value.

## Repro

```python
x = torch.zeros(1, 1, 6, 8); x[0, 0, 1, 2] = 1.0
for scale in (0.0, 1e-06, 0.2):
    torch.manual_seed(0)
    try:
        aug = K.RandomThinPlateSpline(scale=scale, p=1.0)
        aug(x)
        d = (aug._params["dst"] - aug._params["src"]).abs().max().item()
        print(f"scale={scale!r:8s} -> ok, max|dst - src| control-point offset = {round(d, 6)}")
    except Exception as e:
        print(f"scale={scale!r:8s} -> {type(e).__name__}: {e}")

# what the other magnitude-driven geometric classes do at their own zero magnitude,
# each at its own default align_corners
print("--- zero magnitude on the other magnitude-driven 2D geometric classes ---")
ramp = torch.arange(48., dtype=torch.float32).reshape(1, 1, 6, 8) / 48.
for label, make in (
    ("RandomRotation(degrees=0.)          ", lambda: K.RandomRotation(degrees=(0., 0.), p=1.)),
    ("RandomAffine(degrees=0.)            ", lambda: K.RandomAffine(degrees=(0., 0.), p=1.)),
    ("RandomShear(shear=0.)               ", lambda: K.RandomShear(shear=(0., 0.), p=1.)),
    ("RandomTranslate(translate_x=0.)     ", lambda: K.RandomTranslate(translate_x=(0., 0.), p=1.)),
    ("RandomPerspective(distortion=0.)    ", lambda: K.RandomPerspective(distortion_scale=0., p=1.)),
    ("RandomElasticTransform(alpha=(0,0)) ", lambda: K.RandomElasticTransform(alpha=(0., 0.), p=1.)),
    ("RandomRotation90(times=0)           ", lambda: K.RandomRotation90(times=(0, 0), p=1.)),
):
    torch.manual_seed(0)
    try:
        o = make()(ramp)
        print(f"{label} -> ok,     max|out - in| = {round((o - ramp).abs().max().item(), 6)}")
    except Exception as e:
        print(f"{label} -> {type(e).__name__}: {str(e).splitlines()[0]}")
```

## Observed

```
kornia 0.9.0rc1 torch 2.14.0
scale=0.0      -> ValueError: Expected parameter low (Tensor of shape ()) of distribution Uniform(low: -0.0, high: 0.0) to satisfy the constraint LessThan(upper_bound=0.0), but found invalid values:
-0.0
scale=1e-06    -> ok, max|dst - src| control-point offset = 1e-06
scale=0.2      -> ok, max|dst - src| control-point offset = 0.164609
--- zero magnitude on the other magnitude-driven 2D geometric classes ---
RandomRotation(degrees=0.)           -> ok,     max|out - in| = 0.0
RandomAffine(degrees=0.)             -> ok,     max|out - in| = 0.0
RandomShear(shear=0.)                -> ok,     max|out - in| = 0.0
RandomTranslate(translate_x=0.)      -> ok,     max|out - in| = 0.0
RandomPerspective(distortion=0.)     -> ok,     max|out - in| = 0.734375
RandomElasticTransform(alpha=(0,0))  -> ok,     max|out - in| = 0.734375
RandomRotation90(times=0)            -> ok,     max|out - in| = 0.0
```

## Expected

Two coherent outcomes:

1. **`scale=0.0` yields the identity warp** — the degenerate interval is special-cased to a zero displacement, so a zero magnitude means no deformation, as it does for the five neighbours in the observed run that return the input unchanged.
2. **kornia raises its own error**, naming `scale`, stating that it must be strictly positive, and saying what to use instead.

The audit recommends (1), because a zero magnitude reads as "off" rather than as a mistake, because it is what a caller sweeping `scale` downward expects at the end of the sweep, and because it is the only outcome consistent with the seven neighbours in the observed run, none of which rejects its own zero. Either way the current message must go: it names internals of a distribution the caller never constructed.

Whichever is chosen, it is a focused fix in the TPS parameter generator and welcome as a PR. Open PR #3944 touches the same class's sampling grid and is worth reading alongside it, though it does not address the degenerate `scale`.

## Classification

**Semantic contract** — the documented `scale` range has an endpoint the class cannot accept, and the rejection surfaces as a torch-internal message rather than a kornia one. Idiom: 4 (degenerate/singular input silently producing garbage — here producing a raw exception instead, with the same root: the degenerate case has no branch).

## Disposition

`n/a — single symbol`. One class, one parameter, one boundary. Focused fix, welcome as a PR. Batch 6 documents the strictly-positive `scale` requirement on the class and pins the current raise; it does not change it.

## Environment

trueoneplusone:

`RandomThinPlateSpline(scale=0.0)` raises while constructing `Uniform(-0.0, 0.0)`, before an image or parameters can be processed. Handle this endpoint with zero control-point displacement, preserving the existing sampling path for nonzero scales.

The regression tests cover empty/single/multiple batches, both `same_on_batch` settings, parameter dtype/device, image identity with `align_corners=True`, parameter replay, and the identity image gradient. The docstring and changelog describe the newly accepted endpoint.

This is limited to the parameter-generation failure. The separate `align_corners=False` sampling-grid issue remains tracked in #3928 and PR #3944; accepting zero displacement does not repair that grid. The identity test follows the existing TPS tests in skipping float16 because `get_tps_transform` already produces NaN weights for coincident control points in that dtype. Zero-displacement parameter tests still run for float16.

Fixes #4416.

## What did you check?

Windows, Python 3.13.5, PyTorch 2.10.0, kornia_rs 0.1.14. Used a local virtual environment and direct Python commands because Pixi is not installed.

```text
# Before the fix: all 16 new float32/float64 cases fail at Uniform(-0.0, 0.0).
python -m pytest tests/augmentation/test_augmentation.py -k "RandomThinPlateSpline and zero_scale" --device=cpu --dtype=float32,float64
16 failed

# After the fix:
python -m pytest tests/augmentation/test_augmentation.py -k RandomThinPlateSpline --device=cpu --dtype=float32,float64
34 passed, 16 skipped, 708 deselected

python -m pytest tests/augmentation/test_augmentation.py -k "RandomThinPlateSpline and zero_scale" --device=cpu --dtype=float16,bfloat16
14 passed, 2 skipped, 742 deselected

ty check kornia/augmentation/_2d/geometric/thin_plate_spline.py
All checks passed!

python -m pre_commit run --all-files
All 12 hooks passed.

git diff --cached --check
No output.
```

The existing class-scoped fixture emits one Pytest deprecation warning on pytest 9.1.1. Full-repository `ty check kornia` emits 144 diagnostics in this environment; its concise output is byte-for-byte identical on the base revision and this branch. A baseline worktree also confirmed the existing float16 identity NaNs. With seed 123, the positive-scale generated parameters were byte-identical between base and branch, and negative scale still raises the existing `ValueError`.

## How did AI help?

Codex investigated the issue, implemented the patch and regression tests, ran the reported checks, and compared behavior against a separate baseline worktree. A separate Codex agent reviewed the diff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -380,7 +380,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug fixes
 
 * `RandomThinPlateSpline(scale=0.0)` now generates zero control-point displacement instead of
-  raising an error when constructing a degenerate uniform distribution.
+  raising an error when constructing a degenerate uniform distribution. (#4463)
 
 * `RandAugment`'s `m` guard is exclusive at both ends, but its docstring and its error message
   both named the closed interval `[0, 30]`, so a user who asked for the maximum strength the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -379,6 +379,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `RandomThinPlateSpline(scale=0.0)` now generates zero control-point displacement instead of
+  raising an error when constructing a degenerate uniform distribution.
+
 * `RandAugment`'s `m` guard is exclusive at both ends, but its docstring and its error message
   both named the closed interval `[0, 30]`, so a user who asked for the maximum strength the
   message advertised got an exception saying `30` was in range. Both now read `(0, 30)`; the
--- a/kornia/augmentation/_2d/geometric/thin_plate_spline.py
+++ b/kornia/augmentation/_2d/geometric/thin_plate_spline.py
@@ -31,7 +31,8 @@ class RandomThinPlateSpline(AugmentationBase2D):
     .. image:: _static/img/RandomThinPlateSpline.png
 
     Args:
-        scale: the scale factor to apply to the destination points.
+        scale: the non-negative scale factor to apply to the destination points.
+            Zero leaves the control points unchanged.
         align_corners: Interpolation flag used by ``grid_sample``.
         mode: Interpolation mode used by `grid_sample`. Either 'bilinear' or 'nearest'.
         same_on_batch: apply the same transformation across the batch.
@@ -69,7 +70,7 @@ class RandomThinPlateSpline(AugmentationBase2D):
             "align_corners": align_corners,
             "padding_mode": SamplePadding.get(padding_mode),
         }
-        self.dist = torch.distributions.Uniform(-scale, scale)
+        self.dist = None if scale == 0 else torch.distributions.Uniform(-scale, scale)
 
     def generate_parameters(self, shape: Tuple[int, ...]) -> Dict[str, torch.Tensor]:
         B, _, _, _ = shape
@@ -84,7 +85,9 @@ class RandomThinPlateSpline(AugmentationBase2D):
             dtype=dtype,
         ).expand(B, 5, 2)
 
-        if self.same_on_batch:
+        if self.dist is None:
+            noise = torch.zeros_like(src)
+        elif self.same_on_batch:
             noise = self.dist.rsample((1, 5, 2)).to(device=device, dtype=dtype)
             noise = noise.expand(B, 5, 2)
         else:
--- a/tests/augmentation/test_augmentation.py
+++ b/tests/augmentation/test_augmentation.py
@@ -5502,6 +5502,35 @@ class TestRandomThinPlateSpline(CommonTests):
 
         assert any(d > 0 for d in diffs)
 
+    @pytest.mark.parametrize("batch_size", [0, 1, 4])
+    @pytest.mark.parametrize("same_on_batch", [False, True])
+    def test_zero_scale_parameters(self, batch_size, same_on_batch, device, dtype):
+        aug = self._augmentation_cls(scale=0.0, same_on_batch=same_on_batch, p=1.0)
+        aug.set_rng_device_and_dtype(device=device, dtype=dtype)
+
+        params = aug.generate_parameters((batch_size, 3, 6, 8))
+
+        assert params["dst"].shape == (batch_size, 5, 2)
+        assert params["dst"].device == device
+        assert params["dst"].dtype == dtype
+        self.assert_close(params["dst"], params["src"], atol=0, rtol=0)
+
+    @pytest.mark.parametrize("same_on_batch", [False, True])
+    def test_zero_scale_identity(self, same_on_batch, device, dtype):
+        if dtype == torch.float16:
+            pytest.skip("get_tps_transform is numerically unstable in float16 (produces NaN)")
+        # align_corners=False has a separate sampling-grid defect tracked in #3928.
+        aug = self._augmentation_cls(scale=0.0, align_corners=True, same_on_batch=same_on_batch, p=1.0)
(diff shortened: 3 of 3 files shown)


ducha-aiki (dismissed):
Reviewed exact head `86215081`.

Correct and tightly scoped: `scale == 0` skips the `Uniform(-0.0, 0.0)` construction and `generate_parameters` emits zero displacement via `zeros_like(src)`, so the nonzero path is byte-identical to main (same `dst` for the same seed). Across 42 configurations (cpu and mps, all four dtypes, both `same_on_batch`, B in {0, 1, 3}) the `scale=0.0` params match `scale=1e-6` in keys, shapes, dtype, device and strides, replay through `params=` is bit-exact, and the identity holds with `align_corners=True` (6e-8 in float32, exact in float64). The remaining `align_corners=False` offset is the pre-existing #3928 grid defect and is correctly left out. The float16 skip on the identity test follows the precedent in `tests/geometry/transform/test_thin_plate_spline.py`; base already produces NaN TPS weights in float16 at small nonzero scales, so it is not this branch. #4416 is fully covered; the documented-defect grep is empty; the release note cites #4463.

One nit, not blocking:

- [P3] The docstring now says `scale` is non-negative, but `scale=-0.1` still surfaces the raw torch `LessThan` constraint message. A kornia-named guard would complete the contract the docstring states. Outside #4416's scope.

Validation (isolated worktree, import origin verified): all 16 new float32/float64 nodes fail on `8fb9ad12` with the `Uniform(low: -0.0, high: 0.0)` error and pass here; the ThinPlateSpline selection under cpu-float16 15 passed / 10 skipped, cpu-bfloat16 17 / 8, MPS float32 17 passed; doctest and ruff clean. Not run: full suite, CUDA. No dynamo test exists for this class, so compile coverage of the new branch is unverified by the suite.

ducha-aiki (approved):
Re-approving on `b805f1db`. This head is the author's merge of `main` (d9f56df8) into the approved head `86215081`: every path except `CHANGELOG.md` is tree-identical to `git merge-tree` of the two parents, and the `CHANGELOG.md` result carries exactly the branch's own bullet on top of main's. All 40 checks are green on it; nothing else changed since the approval above.

PR #4465, opened 2026-09-12
feat(augmentation): add opt-in geometry audit reports

### LINKED ISSUE #3854: Feature: augmentation audit reports for geometric provenance and round-trip error
### Feature request

Add an opt-in **augmentation audit report** that records geometric provenance for a composed Kornia augmentation pipeline and measures how well transformed spatial data can be mapped back to its source coordinates.

### Motivation

In spatial-AI workflows, the same augmentation may affect images, masks, boxes, keypoints, depth, and camera geometry. Pipelines can run successfully while introducing subtle inconsistencies: a non-invertible crop, keypoints outside the valid region, a mask transformed with unsuitable interpolation, or accumulated round-trip drift.

Kornia already exposes much of the underlying augmentation metadata. A small, unified audit layer could turn that metadata into an actionable debugging and validation artifact without changing normal execution.

### Proposed API (illustrative)

```python
out, report = pipeline.audit(
    image=image,
    mask=mask,
    keypoints=keypoints,
    check_roundtrip=True,
)

print(report.summary())
report.to_json("augmentation-audit.json")
```

The report could include:

- ordered transform names and sampled parameters;
- composed forward/inverse transformation matrices when available;
- which inputs were transformed and with what interpolation/padding policy;
- valid-region coverage and counts of boxes/keypoints moved out of frame;
- invertibility status, including the first non-invertible operation;
- round-trip error statistics for points and optionally sampled image coordinates;
- deterministic seed/device/dtype metadata needed to reproduce the run;
- warnings based on configurable thresholds.

### Why this fits Kornia

This would make differentiable augmentation pipelines easier to trust in robotics, 3D vision, medical imaging, and multimodal training. It could also provide a machine-readable artifact for CI: teams could fail a test when median point round-trip error exceeds a threshold or valid coverage drops unexpectedly.

### Suggested scope for an MVP

1. Support `AugmentationSequential` with images plus keypoints/boxes.
2. Reuse existing sampled parameters and transformation matrices rather than recomputing transforms.
3. Return a lightweight Python dataclass with `summary()` and `to_dict()`/JSON serialization.
4. Add tests for an invertible affine pipeline and a pipeline containing a non-invertible crop.

A later extension could compare audit reports across releases or visualize source/transformed/round-tripped control points.

trueoneplusone:

Adds an experimental, opt-in `kornia.contrib.audit(sequence, ...)` utility that runs an `AugmentationSequential` once and returns its ordinary outputs alongside a typed geometry report. It helps detect spatial-label inconsistencies without changing ordinary forward execution.

- Records captured leaf execution order, including nested built-in sequences and repeated sampled operations, detached parameters and effective flags, shapes, dtype/device, and composed coordinate matrices.
- Inverse-maps the actual returned keypoints and boxes, reporting per-batch finite error counts, mean/median/maximum pixel errors, and out-of-frame/nonfinite counts.
- Incorporates `RandomCrop` prepadding according to the returned image branch. Ambiguous mixed shape-changing operations are reported as unsupported when cached matrices cannot certify every image row.
- Keeps the pipeline forward in the caller's autocast context, then disables autocast for diagnostic matrix composition and inverse mapping.
- Treats missing/out-of-order capture, opaque policy containers, custom sequential subclasses, non-rigid operations, and singular/nonfinite matrices conservatively instead of reporting false success.
- Provides `summary()`, JSON-safe `to_dict()`/`to_json()`, configurable warning thresholds, and a runnable contrib documentation example.

The current scope is one nonempty BCHW image followed by optional batched masks, boxes, and keypoints. Dictionary/ragged/unbatched inputs, video/patch/3D containers, and modules registered under multiple aliases are explicitly unsupported. Reporting is an eager diagnostic API and adds snapshot/matrix costs only when called.

Closes #3854. Image-area coverage, image reconstruction, camera/depth auditing, and report visualization remain outside this change.

## Changes after review

- Moved the implementation from stable `kornia.augmentation` into `kornia.contrib.augmentation_audit` as requested. The public entry point is now the free function `kornia.contrib.audit(sequence, ...)`.
- Removed `AugmentationSequential.audit`, the stable augmentation exports, and the `tests/api_surface.json` entries.
- Moved the tests to `tests/contrib/test_augmentation_audit.py` and the documentation to `docs/source/contrib.rst`.
- Fixed lazy transform-matrix materialization under autocast. The matrix is first read in the caller's autocast context so keypoint/box outputs remain exactly equal to a normal forward; only later diagnostic arithmetic disables autocast.
- Added a `RandomResizedCrop` CPU bfloat16 regression that failed before the fix because audited keypoints differed from normal forward output by up to 0.125 pixels.
- Merged current `main` and resolved the CHANGELOG-only conflict. The final diff against `main` contains five expected files, all in contrib/docs/tests plus the changelog.

## What did you check?

On Windows with Python 3.13.5 and PyTorch 2.10.0, with the import origin verified as this worktree:

```text
python -m pytest tests/contrib/test_augmentation_audit.py tests/test_api_surface.py --device=cpu --dtype=float32,float64 -q
342 passed

python -m pytest tests/contrib/test_augmentation_audit.py --device=cpu --dtype=float16 --xfail-known-failures --known-failure-profile=cpu-float16 -q
103 passed

python -m pytest tests/contrib/test_augmentation_audit.py --device=cpu --dtype=bfloat16 --xfail-known-failures --known-failure-profile=cpu-bfloat16 -q
103 passed

python -m ruff check kornia/contrib/augmentation_audit.py kornia/contrib/__init__.py tests/contrib/test_augmentation_audit.py
All checks passed!

python -m ty check kornia/contrib/augmentation_audit.py
All checks passed!

python -m pre_commit run --all-files
All hooks passed.
```

A focused Sphinx build of the new contrib section passes with warnings treated as errors. The complete local Sphinx build could not start because this environment lacks `sphinx_autodoc_typehints`; GitHub and Read the Docs CI provide the full documentation build.

The previously red macOS Python 3.12 / PyTorch 2.5.1 leg timed out in the existing `RRDBNetBuilder` ONNX export test after 10,127 other tests passed. It did not exercise this feature. The new push starts a fresh CI matrix.

## How did AI help?

Codex assisted with implementation, test design, documentation, review-response changes, and local validation. Review findings were reproduced before the corresponding fixes. These checks do not replace maintainer review or CI.

### DIFF
--- a/kornia/augmentation/container/audit.py
+++ b/kornia/augmentation/container/audit.py
@@ -1,0 +1,461 @@
+# LICENSE HEADER MANAGED BY add-license-header
+#
+# Copyright 2018 Kornia Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Opt-in diagnostics for the geometry of an executed augmentation pipeline."""
+
+from __future__ import annotations
+
+import json
+import math
+from dataclasses import dataclass, fields, is_dataclass
+from enum import Enum
+from functools import partial
+from typing import TYPE_CHECKING, Any, Literal, TypeVar, cast
+
+import torch
+from torch import Tensor, nn
+
+from kornia.augmentation._2d.geometric.base import GeometricAugmentationBase2D
+from kornia.augmentation._2d.geometric.crop import RandomCrop
+from kornia.augmentation._2d.intensity.base import IntensityAugmentationBase2D
+from kornia.augmentation._3d.base import AugmentationBase3D
+from kornia.constants import DataKey
+from kornia.geometry.boxes import Boxes
+from kornia.geometry.keypoints import Keypoints
+
+from .image import ImageSequential
+from .ops import DataType
+from .params import ParamItem
+from .patch import PatchSequential
+from .video import VideoSequential
+
+if TYPE_CHECKING:
+    from .augment import AugmentationSequential
+
+__all__ = ["AugmentationAuditReport", "AugmentationAuditStep", "SpatialAudit"]
+
+_T = TypeVar("_T")
+_BOX_MODES = {DataKey.BBOX: "vertices_plus", DataKey.BBOX_XYXY: "xyxy_plus", DataKey.BBOX_XYWH: "xywh"}
+
+
+def _snapshot(value: _T) -> _T:
+    if isinstance(value, Tensor):
+        return cast(_T, value.detach().clone())
+    if isinstance(value, ParamItem):
+        return cast(_T, ParamItem(value.name, _snapshot(value.data)))
+    if isinstance(value, dict):
+        return cast(_T, {key: _snapshot(item) for key, item in value.items()})
+    if isinstance(value, (tuple, list)):
+        return cast(_T, type(value)(_snapshot(item) for item in value))
+    return value
+
+
+def _json_value(value: Any) -> Any:
+    if isinstance(value, Tensor):
+        return {
+            "values": _json_value(value.detach().cpu().tolist()),
+            "shape": list(value.shape),
+            "dtype": str(value.dtype),
+            "device": str(value.device),
+        }
+    if isinstance(value, ParamItem):
+        return {"name": value.name, "data": _json_value(value.data)}
+    if is_dataclass(value) and not isinstance(value, type):
+        return {field.name: _json_value(getattr(value, field.name)) for field in fields(value)}
+    if isinstance(value, Enum):
+        return value.name
+    if isinstance(value, dict):
+        return {str(_json_value(key)): _json_value(item) for key, item in value.items()}
+    if isinstance(value, (list, tuple)):
+        return [_json_value(item) for item in value]
+    if isinstance(value, float) and not math.isfinite(value):
+        return None
+    if value is None or isinstance(value, (str, bool, int, float)):
+        return value
+    if isinstance(value, (torch.dtype, torch.device)):
+        return str(value)
+    raise TypeError(f"Audit metadata of type {type(value).__name__} cannot be serialized to JSON.")
+
+
+@dataclass(frozen=True)
+class AugmentationAuditStep:
+    """Snapshot of one executed leaf operation, including repeated occurrences.
+
+    ``matrix`` maps this occurrence's local input pixel coordinates to its output
+    coordinates, including RandomCrop prepadding. The full pipeline mapping is
+    stored separately on the report. ``None`` means that the
+    operation has no supported 2D matrix, not that it is an identity transform.
+    Parameters and flags are detached copies; ``occurrence`` is zero-based for
+    each qualified module name. Shapes describe the image path.
+    """
+
+    name: str
+    occurrence: int
+    module: str
+    input_shape: tuple[int, ...]
+    output_shape: tuple[int, ...]
+    params: dict[str, Any]
+    flags: dict[str, Any]
+    matrix: Tensor | None
+    unsupported_reason: str | None
+
+
+@dataclass(frozen=True)
+class SpatialAudit:
+    """Per-batch diagnostics for one returned box or keypoint input.
+
+    Each point or box is one item. A box is outside if any corner is outside
+    the inclusive pixel-center rectangle ``[0, width-1] x [0, height-1]``.
+    Nonfinite items are counted separately, not counted as outside. Fractions
+    use the total item count; empty inputs have a NaN fraction.
+
+    Round trips inverse-map the *actual returned coordinates* with the recorded
+    composed matrix. Errors are Euclidean point distances or symmetric Hausdorff
+    distances between the four box corners (independent of corner ordering).
+    Tensor box exports can lose orientation by taking an axis-aligned envelope;
+    this loss is included in the measurement. ``Boxes`` objects retain corners.
+    Statistics cover only finite errors, in pixels; a batch with no valid errors
+    has NaN statistics and zero ``roundtrip_valid_count``. Diagnostics use float32
+    for half inputs, otherwise the image dtype.
+    """
+
+    input_index: int
+    data_key: str
+    metric: Literal["euclidean", "corner_hausdorff"]
+    count: Tensor
+    out_of_frame: Tensor
+    nonfinite: Tensor
+    out_of_frame_fraction: Tensor
+    roundtrip_valid_count: Tensor
+    roundtrip_mean: Tensor
+    roundtrip_median: Tensor
+    roundtrip_max: Tensor
+
+
+@dataclass(frozen=True)
+class AugmentationAuditReport:
+    """Geometry provenance and spatial diagnostics from one augmentation call.
+
+    ``params`` is a detached snapshot of native ``ParamItem`` values for replay
+    with the same pipeline configuration. JSON is a diagnostic export, not a
+    parameter deserializer. Reports retain neither image pixels nor autograd
+    graphs. Tensor fields are independent snapshots, not read-only tensors.
+
+    ``geometry_status`` is ``unsupported`` when any executed operation lacks a
+    supported matrix, ``singular`` when any composed matrix cannot be inverted,
+    or ``available``. Inspect the per-batch ``invertible`` mask and spatial
+    errors rather than interpreting availability as alignment success.
+    ``inverse_matrix`` has NaNs for singular/nonfinite batches. No image inverse
+    is evaluated: coordinate invertibility does not imply pixel reconstruction,
+    particularly after cropping, interpolation or intensity changes.
+    """
+
+    inputs: list[dict[str, Any]]
+    output_shape: tuple[int, ...]
+    steps: list[AugmentationAuditStep]
+    params: list[ParamItem]
+    configured_extra_args: dict[DataKey, dict[str, Any]]
+    geometry_status: Literal["available", "unsupported", "singular"]
+    matrix: Tensor | None
+    inverse_matrix: Tensor | None
+    invertible: Tensor
+    spatial: list[SpatialAudit]
+    warnings: list[str]
+    roundtrip_tolerance: float
+    out_of_frame_tolerance: float
+    image_reconstruction_evaluated: bool = False
+
+    def to_dict(self) -> dict[str, Any]:
+        """Export JSON-safe metadata; tensor values include dtype/device/shape, and NaNs become None."""
+        return cast(dict[str, Any], _json_value(self))
+
+    def to_json(self, *, indent: int | None = 2) -> str:
+        """Serialize the report as strict JSON (nonfinite diagnostic values become null)."""
+        return json.dumps(self.to_dict(), indent=indent, allow_nan=False)
+
+    def summary(self) -> str:
+        """Describe matrix availability and diagnostic warnings without claiming image invertibility."""
+        return (
+            f"Augmentation audit: {len(self.steps)} operations, geometry {self.geometry_status}, "
+            f"{len(self.spatial)} spatial inputs. Image reconstruction not evaluated."
+            + ("\n" + "\n".join(self.warnings) if self.warnings else "")
+        )
+
+
+def _leaves(sequence: ImageSequential, prefix: str = "", seen: set[int] | None = None) -> list[tuple[str, nn.Module]]:
+    seen = set() if seen is None else seen
+    result: list[tuple[str, nn.Module]] = []
+    for name, module in sequence._modules.items():
+        if module is None:
+            continue
+        path = f"{prefix}.{name}" if prefix else name
+        if id(module) in seen:
+            raise ValueError("audit does not support a module registered under multiple names.")
+        seen.add(id(module))
+        if isinstance(module, (VideoSequential, PatchSequential, AugmentationBase3D)):
+            raise ValueError("audit supports 2D image pipelines, not video, patch or 3D augmentations.")
+        if isinstance(module, ImageSequential):
+            result.extend(_leaves(module, path, seen))
+        else:
+            result.append((path, module))
+    return result
+
+
+def _coordinates(value: DataType, key: DataKey) -> Tensor:
+    if (isinstance(value, Boxes) and key not in _BOX_MODES) or (
+        isinstance(value, Keypoints) and key != DataKey.KEYPOINTS
+    ):
+        raise ValueError("audit Boxes/Keypoints objects must match their data key.")
+    if isinstance(value, (Boxes, Keypoints)):
+        if value._N is not None:
+            raise ValueError("audit does not support ragged Boxes or Keypoints constructed from lists.")
+        return value.data
+    if not isinstance(value, Tensor):
+        raise ValueError("audit requires batched tensors, Boxes or Keypoints; ragged inputs are unsupported.")
+    if key in _BOX_MODES:
+        # Match the container's public representation, including axis-aligned tensor exports.
+        return Boxes.from_tensor(value, mode=_BOX_MODES[key], validate_boxes=False).data
+    return value
+
+
+def _capture(
+    module: nn.Module, inputs: tuple[Any, ...], output: Any, *, name: str, steps: list[AugmentationAuditStep]
+) -> None:
+    image = inputs[0]
+    params = _snapshot(getattr(module, "_params", {}))
+    flags = _snapshot(getattr(module, "flags", {}))
+    matrix = None
+    reason = None
+    if isinstance(module, (GeometricAugmentationBase2D, IntensityAugmentationBase2D)):
+        # Access before detaching: lazy matrix materialization must preserve ordinary autograd behavior.
+        matrix = _snapshot(module.transform_matrix)
+        if matrix is not None and isinstance(module, RandomCrop):
+            padding = params["padding_size"].to(matrix)
+            translation = torch.eye(3, device=matrix.device, dtype=matrix.dtype).expand_as(matrix).clone()
+            applied = (params["batch_prob"] > 0.5).to(matrix)
+            translation[:, 0, 2] = padding[:, 0] * applied
+            translation[:, 1, 2] = padding[:, 2] * applied
+            matrix = matrix @ translation
+        if matrix is not None and matrix.shape != (image.shape[0], 3, 3):
+            matrix = None
+            reason = "operation produced a matrix with an unsupported shape"
+        elif matrix is None:
+            reason = "operation did not produce a 2D matrix"
+    elif type(module) is nn.Identity:
+        matrix = torch.eye(3, device=image.device, dtype=image.dtype).expand(image.shape[0], -1, -1).clone()
+    else:
+        reason = "non-rigid or unknown operation; coordinate propagation is not certified by a 2D matrix"
+    steps.append(
+        AugmentationAuditStep(
+            name,
+            sum(step.name == name for step in steps),
+            type(module).__name__,
+            tuple(image.shape),
+            tuple(output.shape),
+            params,
+            flags,
+            matrix,
+            reason,
+        )
+    )
+
+
+def _matrices(steps: list[AugmentationAuditStep], image: Tensor) -> tuple[Tensor | None, Tensor | None, Tensor]:
+    dtype = torch.float64 if image.dtype == torch.float64 else torch.float32
+    identity = torch.eye(3, device=image.device, dtype=dtype).expand(image.shape[0], -1, -1)
+    matrix = identity.clone()
+    for step in steps:
+        if step.matrix is None or step.matrix.shape != matrix.shape:
+            return None, None, torch.zeros(image.shape[0], device=image.device, dtype=torch.bool)
+        matrix = step.matrix.to(matrix) @ matrix
+    finite = matrix.isfinite().all(dim=(-2, -1))
+    inverse, info = torch.linalg.inv_ex(torch.where(finite[:, None, None], matrix, identity))
+    valid = finite & (info == 0) & inverse.isfinite().all(dim=(-2, -1))
+    inverse = torch.where(valid[:, None, None], inverse, torch.full_like(inverse, float("nan")))
+    return matrix, inverse, valid
+
+
+def _spatial_audit(
+    source: Tensor,
+    result: Tensor,
+    key: DataKey,
+    index: int,
+    shape: tuple[int, ...],
+    inverse: Tensor | None,
+    dtype: torch.dtype,
+) -> SpatialAudit:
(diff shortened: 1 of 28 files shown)


ducha-aiki (changes requested):
Reviewed exact head `0223eea9`.

Credit where due: ordinary `AugmentationSequential.forward` is untouched (the `augment.py` diff is one import plus the `audit` method), same-seed forward outputs are byte-identical to main, `torch._dynamo.explain` gives the same single graph on both trees, and the core mapping is right on the cases I checked by hand on a 6x10 image: `RandomCrop` with asymmetric padding yields the exact expected shift and window, crop plus forced-parameter affine round-trips to 1e-16, a forced-singular affine reports `singular` with a per-row `invertible` mask, and a 17-op sweep marks every non-rigid op `unsupported`. Exports, `api_surface.json`, CHANGELOG and the rst `currentmodule` are consistent, and all 61 new nodes fail on main.

Four blockers:

- [P1] `audit.py:199-215, 398-400`: policy containers give a false success. `RandAugment`, `AutoAugment` and `TrivialAugment` are `ImageSequentialBase` but not `ImageSequential`, so `_leaves` keeps them as a single leaf, and the container runs them through `transform_inputs`, never `__call__`, so the forward hook never fires. `AugmentationSequential(RandAugment(n=2, m=10), data_keys=["input", "keypoints"]).audit(img, kp)` returns zero steps, an identity matrix and `geometry_status="available"` after the policy moved the image, and on seeds where it also moves the keypoints the report still says success. Either recurse into policy containers or classify them `unsupported`.
- [P1] `audit.py:246-252`: `RandomCrop` with `p < 1` gives a false alarm. The prepadding translation is multiplied by `applied`, but the container pads every row and returns the padded top-left window for unapplied rows, so their true mapping is the pure pad translation. With `RandomCrop((4, 8), padding=(1, 2), p=0.5)` and a batch where no row is applied, keypoints move from `(3, 3)` to `(4, 5)` and the report says "round-trip error exceeds 0.001 pixels" plus a warning. Dropping the multiplier makes the error 0. Not covered by a test (the mixed-batch pin uses a flip, the crop pin uses `p=1`).
- [P1] `tests/augmentation/container/test_audit.py:75,139,150`: the blocking half-precision jobs are red; 3 failures each on the cpu-float16 and cpu-bfloat16 profiles, reproduced locally. Two are test dtype mismatches against the code's deliberate float32 diagnostics, one is a 1e-4 box tolerance the container cannot meet at half. Run the whole file under both profiles, not a `-k` subset.
- [P1] `docs/source/augmentation.container.rst:30,71`: the `-W` docs job is red because both new section underlines are one character short (29-character title over 28 tildes, 23 over 22 carets).

Also:

- [P2] An empty `steps` list always yields `available`. Make the report detect leaves that were registered but never fired, rather than special-casing the policy containers only.
- [P3] The rst should say what happens to policy containers; and the content-loss warning keys on `"Crop" in class name`, so a `Resize` or `LongestMaxSize` downscale gets none.

Scope note for the maintainers rather than the author: #3854 is a sanctioned request and its MVP list is covered, but the feature is built entirely on public data and adds a sizeable exported surface. Worth deciding whether it lives here or as an example, and whether only `AugmentationAuditReport` needs exporting from `kornia.augmentation`.

Validation (isolated worktree, import origin verified): `test_audit.py` CPU float32/float64 208 passed; cpu-float16 31 passed, 3 failed; cpu-bfloat16 likewise 3 failed; ruff clean; no import cycle. Timing of the ordinary forward was too noisy on a loaded machine to report. Not run: full suite, CUDA, MPS, the sphinx build itself (the CI log was read instead).

ducha-aiki:
Addendum: head `e04f8df7` was pushed while the review of `0223eea9` above was being written. Its delta is `docs/source/augmentation.container.rst` (both underlines now match their titles) and `tests/augmentation/container/test_audit.py` (13 lines), so it addresses the docs and half-precision blockers. `kornia/augmentation/container/audit.py` is unchanged between the two heads, so the two behavioural P1s (policy containers reporting a false success, and the `RandomCrop` `p < 1` false alarm) and the P2 still apply to the current head as written.

ducha-aiki (approved):
Reviewed exact head `f38fe0dc` (delta since `0223eea9`/`e04f8df7`: `21eb3cda` plus the merge commit), whole diff re-read.

All four blockers and the P2 from the previous round are resolved, each reproduced independently in an isolated worktree:

- Policy containers: `AugmentationSequential(RandAugment(n=2, m=10), data_keys=["input", "keypoints"]).audit(...)` now reports `geometry_status="unsupported"`, `matrix=None`, `invertible` all False and the capture warning (`selected ['RandAugment_0']; captured []`); same for `AutoAugment`, `TrivialAugment`, a policy nested inside an `ImageSequential`, and a custom `AugmentationSequential` subclass. An empty pipeline or an unselected leaf still certifies, so the old P2 is covered by the same check.
- `RandomCrop` with `p < 1`: I re-measured with a pixel landmark instead of the label. In every case where a matrix is reported (`slice`/`resample`, none/mixed/all rows applied, batch of one, non-shape-changing crop), `matrix @ landmark` equals the pixel the landmark actually landed on. The remaining `sqrt(5)` round-trip error is on rows where the image is returned untouched but the container pads the keypoints to `(4, 4)` — that is a real label/image mismatch on `main` (reproduced on the base worktree with plain `forward`), so the PR body's rebuttal is right and my earlier "drop the multiplier" ask was wrong on the label side; the mapping now follows the pixels, which is the correct contract for an audit. The shape-changing mixed `slice` batch is reported `unsupported`, which is honest (the unapplied row really is a random window of the padded image).
- Half precision: whole file, `--xfail-known-failures`, cpu-float16 98 passed, cpu-bfloat16 98 passed. MPS float32 98 passed. CPU float32/float64 with `tests/test_api_surface.py`: 336 passed. All 98 new float32 nodes fail on `main` (`no attribute 'audit'`).
- Docs: both underlines now match their titles (29/29, 23/23).

Also checked: `f38fe0dc` is a pure merge (`git merge-tree 21eb3cda d9f56df8` reproduces its tree exactly, CHANGELOG included; the three added CHANGELOG lines are the branch's own); the audit-off forward is byte-identical to `main` for flip / affine+brightness / `RandomCrop p=0.5` / nested `random_apply` pipelines, RNG state included, and `torch._dynamo.explain` still gives one graph; nested `AugmentationSequential` with keypoints and `Boxes` certifies with round-trip error ≤ 2.4e-6 and replays exactly; pinned ruff and ty are clean; `merge-tree` against #4452 is clean.

Two non-blocking notes:

- `audit.py:260-263`: the shape-changing mixed `slice` batch could be certified rather than marked unsupported, since `crop_by_indices` uses `params["src"]` for every row and `get_perspective_transform(src, dst)` composed with the pad translation is the true per-row map. Fine as a follow-up.
- The container defect this audit exposes — `RandomCrop` with `p < 1` pads keypoints/boxes on unapplied rows while returning the untouched image (`apply_transform_keypoint` pads unconditionally at `crop.py:191`) — is pre-existing and outside this PR; it deserves its own issue so the "real inconsistency" bullet in the rst can eventually point at it.

Scope remains a maintainer call (in-tree feature vs example, and whether the three report types need to be exported from `kornia.augmentation`); nothing in the code blocks it.

Not run: full suite, CUDA, the Sphinx build itself.

ducha-aiki (changes requested):
That was AI review, and now human one - let's add it to kornia.contrib.augmentation_audit first, and then see how it goes.

kornia/contrib/augmentation_audit.py
```diff
@@ -0,0 +1,538 @@
...
+        | (result[..., 1] < 0)
+        | (result[..., 1] > shape[-2] - 1)
+    )
+    if key in _BOX_MODES:
+        outside = outside.any(-1)
+    counts = torch.full((batch,), count, device=result.device, dtype=torch.long)
+    errors = torch.full((batch, count), float("nan"), device=result.device, dtype=dtype)
+    if inverse is not None and count:
+        flat = result.reshape(batch, -1, 2)
+        homogeneous = torch.cat((flat, torch.ones_like(flat[..., :1])), dim=-1) @ inverse.transpose(-1, -2)
```
ducha-aiki, line 397: [P2] Disable autocast for the diagnostic arithmetic

The inverse-mapping matmul runs under the caller's autocast context: `torch.no_grad()` does not disable autocast, and converting operands to float32 does not prevent matmul from being reduced to bfloat16. An empty pipeline therefore reports a round-trip failure even though it returns the image and keypoints exactly unchanged.

Reproduced on CPU with PyTorch 2.14.0 at `f38fe0dc`:

```python
import torch

image = torch.zeros(1, 1, 32, 48)
points = torch.tensor([[[20.01, 15.03]]])
pipeline = K.AugmentationSequential(data_keys=["input", "keypoints"])

with torch.autocast("cpu", dtype=torch.bfloat16):
    (out_image, out_points), report = pipeline.audit(image, points)

assert torch.equal(out_image, image)
assert torch.equal(out_points, points)
print(report.spatial[0].roundtrip_max)
# tensor([0.0316])
print(report.warnings)
# ['Input 1: round-trip error exceeds 0.001 pixels.']
```

Outside autocast the error is zero; rerunning just the diagnostic calculation with autocast disabled also gives zero. Matrix composition in `_matrices` has the same exposure: adding `nn.Identity()` makes `report.matrix` bfloat16 for float32 inputs.

Keep the ordinary forward in the caller's execution context, but disable autocast around diagnostic matrix composition and inverse mapping, including the crop snapshot arithmetic. Add an autocast regression with an empty/identity pipeline and fractional float32 coordinates.

kornia/augmentation/container/audit.py
```diff
@@ -0,0 +1,538 @@
...
+    # saving them on module.flags. `params` is a separate bound argument.
+    flags = _snapshot(
+        {**getattr(module, "flags", {}), **{key: value for key, value in kwargs.items() if key != "params"}}
+    )
+    matrix = None
+    reason = None
+    if isinstance(module, (GeometricAugmentationBase2D, IntensityAugmentationBase2D)):
+        # Access before detaching: lazy matrix materialization must preserve ordinary autograd behavior.
+        matrix = _snapshot(module.transform_matrix)
+        if matrix is not None and isinstance(module, RandomCrop):
```
ducha-aiki: [P2] Handle mixed application for shape-changing operations beyond RandomCrop

The image-mapping correction/rejection is applied only to `RandomCrop`. With a `Resize` and explicitly supplied mixed `batch_prob`, the shape-mismatch fallback returns the resized image branch for every row, while the skipped row's cached matrix and keypoints remain unchanged. The audit reports an identity mapping and zero round-trip error for that row, missing the image/keypoint mismatch.

Reproduced at `f38fe0dc`:

```python
import torch
from kornia.augmentation.container.params import ParamItem

image = torch.zeros(2, 1, 8, 10)
image[:, 0, 2, 4] = 1  # Landmark at (x=4, y=2).
points = torch.tensor([[[4., 2.]], [[4., 2.]]])
resize = K.Resize((4, 5), resample="nearest", p=0.5)
pipeline = K.AugmentationSequential(
    resize, data_keys=["input", "keypoints"]
)
torch.manual_seed(0)
params = resize.forward_parameters(image.shape)
params["batch_prob"] = torch.tensor([0., 1.])

(out_image, out_points), report = pipeline.audit(
    image, points, params=[ParamItem("Resize_0", params)]
)
print(torch.nonzero(out_image[0, 0]))  # tensor([[1, 2]]) => (x=2, y=1)
print(out_points[0])                  # tensor([[4., 2.]])
print(report.matrix[0])               # Identity.
print(report.spatial[0].roundtrip_max) # tensor([0., 0.])
print(report.geometry_status)        # available
```

The only warning is the generic downsampling/content-loss warning. Plain forward with the same parameters returns exactly the same images and keypoints, so this finding concerns the audit's false negative.

Scope: `Resize` normally samples one batch-wide gate. This reproduction requires explicit mixed parameters through the accepted `params` API; it does not occur from the normal Resize sampler. Existing crop tests already exercise analogous mixed parameter overrides.

Extend the shape-change handling to these operations, or report their geometry as unsupported when the cached per-row matrix cannot describe the returned image. Add an explicit-parameters Resize regression using a pixel landmark.

ducha-aiki (commented):
Follow-up correctness review of exact head `f38fe0dc798ad75d1a6fee69eebfe2d009397276`. Two additional P2 findings are attached inline: autocast introduces diagnostic errors on an unchanged pipeline, and explicit mixed application of Resize can produce an undetected image/keypoint mismatch.

Validation: 336 audit and public API tests passed on CPU float32/float64 with PyTorch 2.14.0; both counterexamples were reproduced separately in an isolated worktree with the import origin verified. Full-suite and GPU runs were not repeated.

ducha-aiki (changes requested):
Delta re-review of exact head `4b3e6693`. The delta since `f38fe0dc` is one commit: `audit.py` +25/-6, the rst +5 and `test_audit.py` +29. I re-read the whole diff too.

Thanks for the quick turnaround on the two inline findings. Both hold up under probing. The shape-change fix is solid, and the autocast fix is right for the diagnostics, but it breaks output equality in one path (below). The main request from the last round is still open, though.

**[P2] The move to `kornia.contrib.augmentation_audit` was not done, so this stays at changes requested.** The request was to land the feature in `kornia.contrib.augmentation_audit` first. On `4b3e6693` it is still entirely in `kornia.augmentation`:

- `kornia/augmentation/container/audit.py` (the whole implementation);
- the `AugmentationSequential.audit` method and its `.audit` import in `container/augment.py`;
- the three report types re-exported from `kornia/augmentation/container/__init__.py` and `kornia/augmentation/__init__.py`;
- three names added to the `kornia.augmentation` stable-core inventory in `tests/api_surface.json`, which would put them under the removal guard;
- the `automethod` and the ~100-line section in `docs/source/augmentation.container.rst`;
- `tests/augmentation/container/test_audit.py`;
- a CHANGELOG entry that describes `AugmentationSequential.audit`.

None of this needs to stay in `kornia.augmentation`. `audit(sequence, *args, ...)` is already a free function. It registers forward hooks on the leaves only for the duration of the call and removes them in `finally`, so the containers carry no hooks. I copied `audit.py` unchanged to `kornia/contrib/augmentation_audit.py` (only the relative imports made absolute). `from kornia.contrib.augmentation_audit import audit` then audited a flip pipeline correctly, and `import kornia` still worked with an eager re-export from `kornia/contrib/__init__.py`. The move is therefore:

- the module plus `docs/source/contrib.rst`;
- tests under `tests/contrib/`;
- dropping the method, the `kornia.augmentation` exports and the `api_surface.json` entries;
- rewording the CHANGELOG line.

The current code does add zero cost when the audit is off. On a merge with current `main`, the ordinary `forward` of four pipelines is byte-identical to `main`: outputs, input gradients, sampled params and RNG state give the same SHA-256. The pipelines were flip/affine/jitter, `RandomCrop` p=0.5 + `RandomResizedCrop`, nested `random_apply` + `Resize`, and vflip/`CenterCrop`/`PadTo`, each with mask, keypoints and boxes.

**[P1] `audit.py:312-313`: the autocast fix makes `audit` return different labels than `forward` for lazily built matrices.** The hook now reads `module.transform_matrix` inside `torch.autocast(enabled=False)`. `RandomResizedCrop` (default `cropping_mode="slice"`), `Resize`, the flips and the intensity ops defer their matrix until first access. So in an audit run the hook is what builds and caches it: without autocast, and without the `is_autocast_enabled()` cast in `generate_transformation_matrix`. The container then moves the keypoints and boxes with that cached matrix. A plain `forward` under the same autocast builds it inside autocast. The docstring and the rst promise that outputs have "the ordinary forward structure, values and gradients", and that no longer holds:

```python
img = torch.rand(2, 3, 32, 48)
pts = torch.tensor([[[20.01, 15.03], [40.7, 3.3]], [[7.77, 29.1], [33.3, 11.1]]])
aug = K.AugmentationSequential(K.RandomResizedCrop((20, 30), scale=(0.5, 1.0)), data_keys=["input", "keypoints"])
with torch.autocast("cpu", dtype=torch.bfloat16):
    torch.manual_seed(1); _, fpts = aug(img, pts)
    torch.manual_seed(1); (_, apts), rep = aug.audit(img, pts)
(fpts - apts).abs().max()  # 4b3e6693: 0.125   f38fe0dc: 0
```

MPS float16 autocast gives 0.031. Images and RNG state stay equal; only the labels differ. The same run with `cropping_mode="resample"` (eager matrix) and with `CenterCrop` is equal. So are `Resize` (lazy) on this input and the flips, but only because their matrix arithmetic happens to survive bfloat16 here.

The fix is one line: read the matrix in the caller's context and disable autocast only for the arithmetic after it:

```python
matrix = _snapshot(module.transform_matrix)
with torch.autocast(device_type=image.device.type, enabled=False):
    if matrix is not None and isinstance(module, RandomCrop):
        ...
```

With that change the probe gives 0 for all six cases. The file still passes (193 CPU float32/float64, 101 cpu-bfloat16), and the empty/`Identity` autocast regression still reports zero error. Please add a regression next to `test_diagnostics_disable_cpu_autocast` that compares `audit` outputs with `forward` outputs under autocast for a default `RandomResizedCrop`. A flip is not enough, since it is exact. That test is also CPU-only with no `device` fixture, so MPS/CUDA autocast are not pinned.

**Resolved: autocast diagnostic error (inline P2 #1).** The Codex counterexample now gives zero error and no warning, on both CPU bfloat16 and MPS float16 autocast: empty pipeline, `Identity`, fractional float32 points, `report.matrix` float32. It also holds on torch 2.5.1, where `torch.autocast("mps", enabled=False)` is accepted. Both new autocast tests fail on the `f38fe0dc` `audit.py`.

Some round-trip warnings remain under autocast for real operations. They are genuine, not artifacts: plain `forward` on `main` under CPU bfloat16 autocast already moves keypoints away from the float32 result by 0.14 px (hflip), 0.15 px (affine) and 0.058 px (resize). A sentence in the rst saying that autocast itself quantizes label propagation would stop users reading those warnings as audit noise (P3).

**Resolved: mixed application of shape-changing ops (inline P2 #2).** I ran a pixel-landmark sweep of 81 cases with explicit `batch_prob`:

- ops: `RandomResizedCrop` slice/resample, `Resize`, `LongestMaxSize`, `CenterCrop` slice/resample, `RandomCrop` slice/resample with padding, and `PadTo`;
- gates: mixed `[0, 1, 0]`, all-applied and none-applied;
- each plain, followed by a flip, and nested in `ImageSequential`.

No case reports `available` with a matrix that disagrees with where the landmark pixel landed. Every mixed shape-changing case is `unsupported`, except `RandomCrop` resample (certified, 0 px, matrix matches the pixels). `PadTo` is static and always certifies correctly. The all-applied and none-applied cases certify at 0 px, except none-applied `RandomCrop` with padding. That case reports the real sqrt(5) label/image mismatch from #4473.

Note that every shape-changing op here gates the whole batch (`p=1.0, p_batch=p`, including `RandomCrop`), so mixed batches only come from explicit params. `unsupported` is conservative for the resample-mode crops, whose labels matched the pixels per row; certifying those could be a follow-up. `test_shape_changing_mixed_application_is_unsupported` fails on the `f38fe0dc` code.

**Interaction with #4485** (open; stops `RandomCrop` from padding the annotations of skipped rows, the #4473 defect). It merges cleanly with this PR, and `test_audit.py` + `test_random_crop.py` pass together (285). After #4485, the bullet at `docs/source/augmentation.container.rst:103-106` becomes stale: "A skipped crop can currently leave an image unchanged while padding its labels…". On the combined tree the none-applied round-trip error is 0. Whichever PR lands second should update or remove that bullet. The `_crop_matrix` logic follows the pixels, so it needs no change.

**CI:** the red `tests (macos-latest, float32, 3.12, 2.5.1)` leg is the known flake `tests/contrib/test_super_resolution.py::TestSuperResolutionBuilders::test_to_onnx_exports[RRDBNetBuilder]` (pytest-timeout >120 s). Everything else in that job passed (10127 passed), and `collector` is red only because of that leg. GitHub reports a conflict, but it is CHANGELOG-only: `git merge-tree` against current `main` (including #4452/#4459/#4460/#4461/#4462/#4463) is clean in code, and the merged tree differs from `main` by exactly this PR's 8 files.

**Verified on this head:**

- Head: `test_audit.py` + `tests/test_api_surface.py` CPU float32/float64 340 passed; cpu-float16 101 passed; cpu-bfloat16 101 passed (both with `--xfail-known-failures`); MPS float32 101 passed; torch 2.5.1 CPU float32/float64 193 passed.
- On a throwaway merge with `main` `dc0af1a4`: `tests/augmentation/container` + `tests/test_api_surface.py` CPU float32/float64 1712 passed, 464 skipped, 5 xfailed; `test_audit.py` cpu-float16 101, cpu-bfloat16 101, MPS float32 101, torch 2.5.1 193 passed; the byte-identity digest above.
- The 4 new delta nodes fail with the `f38fe0dc` `audit.py`.
- Probes: autocast (12 pipelines × CPU bfloat16 / MPS float16, `audit` vs `forward` equality, RNG and round-trip), the 81-case landmark sweep, the #4485 combination, and a contrib-module feasibility check.
- Pinned ruff 0.16.6 check/format clean on the changed Python files.
- No author replies on the inline threads.

Not run: ty, the Sphinx build, CUDA, the full suite.

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (69b27b9d) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4465.added.md`.

ducha-aiki (dismissed):
Review of `69b27b9dd6ec4ca371a52fc9c3320c5674621810`.

The maintainer-requested move to kornia.contrib.augmentation_audit and the previous lazy-matrix/autocast-output defect are resolved on this head. The matrix is captured before autocast is disabled, and the release note uses changelog.d/4465.added.md. No remaining current-head blocker found.

Integration note: docs/source/contrib.rst currently describes skipped RandomCrop padding labels. That is accurate against current main, but #4485 repairs it. If #4485 lands first, remove or qualify that example when integrating this PR; if this PR lands first, include the documentation cleanup with #4485. This is an interaction between two queue entries, not a reason to retain the resolved contrib/autocast change requests.

Validation: re-review of the changed implementation and documentation; earlier numerical review evidence was retained. The author reports 342 CPU float32/float64 passes and 103 passes under each half profile; these are author results, not fresh local runs from this sweep. Required hosted CI must finish green.

ducha-aiki:
#4485 has landed, so I pushed two commits to this branch:

- **9f8d94a1** merges current `main`, which includes the #4485 fix and its `RandomCrop` docstring. It merged cleanly.
- **44fe48d0** updates the audit docs to match. The `docs/source/contrib.rst` note said a skipped crop "can currently leave an image unchanged while padding its labels". That is no longer true for a crop skipped for the whole batch. The note now says whole-batch skip leaves images and labels unchanged (#4473). It keeps the case that still happens: manually supplied parameters that apply the crop to only some rows can return a transformed image with untransformed labels on the skipped rows. The matching test comment in `test_crop_matrix_tracks_actual_pixels` is updated too. The test also now checks for zero round-trip error when the crop is applied to or skipped for the whole batch.

Checks (Linux, worktree import verified):
- `tests/contrib/test_augmentation_audit.py`, cpu+cuda, float32/float64: 379 passed
- CPU float16 / bfloat16 with the known-failure profiles: 103 passed each
- The new zero-error check fails in 12 cases against the pre-#4485 `crop.py`, so it covers the fix
- pre-commit on the changed files passes

PR #4471, opened 2026-09-12
fix(augmentation): add bfloat16 support to mix augmentations (#4467)

### LINKED ISSUE #4467: DType lacks BFLOAT16 support in the image-mixing forward path
## Updated diagnosis from the #4452 review

The failing family includes `PatchMix` in addition to `RandomMosaic`, `RandomMixUpV2`, `RandomCutMixV2`, and `RandomJigsaw`: those classes use the `MixAugmentationBaseV2.forward` dtype-enum path. This is not a guarantee that every subclass fails. Transplantation uses a different forward path; a mask-only `RandomTransplantation(...)(mask, data_keys=["mask"])` preserves bfloat16 in the reproduced case, and the 3D variant shares that transplantation path.

The enum/precision repair remains tracked here and in #4471. #4452 removes the universal mix-family claim and links this issue instead. Applied MixUp can additionally promote half inputs; a skipped `p=0` call need not do so.

The original report is preserved below as historical context; the qualifications above supersede conflicting statements.

<details>
<summary>Original report</summary>

## Describe the bug

`kornia.constants.DType` has no `BFLOAT16` member, so `DType.get(torch.bfloat16)` and `DType.to_torch(torch.bfloat16)` both raise `KeyError: 'BFLOAT16'`. Every mix augmentation goes through `DType.get(in_tensor.dtype)` in `MixAugmentationBaseV2.forward`, so `RandomMosaic`, `RandomMixUpV2`, `RandomCutMixV2` and `RandomJigsaw` crash on a bfloat16 input instead of augmenting it. This is a production defect, not a test-only one: bfloat16 is a supported test dtype and `float16` works on the same path.

## Reproduction steps

```python
import torch
from kornia.constants import DType

DType.get(torch.bfloat16)        # KeyError: 'BFLOAT16'
DType.to_torch(torch.bfloat16)   # KeyError: 'BFLOAT16'
K.RandomMosaic(p=1.0)(torch.rand(4, 3, 8, 8, dtype=torch.bfloat16))  # KeyError: 'BFLOAT16'
```

`main` at `8fb9ad12`, torch 2.14, CPU.

## Expected behavior

`DType` enumerates every floating dtype kornia runs tests on; bfloat16 input is augmented in bfloat16, as float16 is.

## Notes

The enum change is four lines, but `testing/half_precision_xfails/cpu_bfloat16.txt` carries 27 strict-xfail lines for mix tests whose recorded failure is exactly this `KeyError`. With the member added, 21 of them pass (the strict verifier then fails) and 6 flip to real numerical failures: `RandomMixUpV2` output dtype comes back float32, and `RandomCutMixV2` / `RandomMosaic` values are not close. The fix therefore needs its own PR that adds the member, deletes the 21 manifest lines, and either fixes or re-records the 6 that become real failures.

Surfaced while reviewing #4454 and #4459, both of which hit the `KeyError` in new bfloat16-parametrized `RandomMosaic` tests.

</details>

msparsh13:
**What does this pull request do?**

- Adds BFLOAT16 support to kornia.constants.DType, fixing KeyError: 'BFLOAT16' for mix augmentations when the input tensor uses torch.bfloat16.

- This enables `RandomMosaic`, `RandomMixUpV2`, `RandomCutMixV2`, and `RandomJigsaw `to run with `bfloat16 `inputs instead of failing during dtype resolution. The affected CPU bfloat16 strict-xfail entries are also updated now that these tests can execute.

Some `RandomMosaic `**bounding-box numerical checks required updated expected values because bfloat16 has lower precision than float32.** The test expectations were adjusted to reflect the valid bfloat16 results rather than loosening the numerical checks broadly.

**Related issue or discussion**

Fixes #4467.

**What did you check?**
 `pixi run test-module tests/augmentation/test_augmentation_mix.py`
- Ran the mix augmentation tests with CPU bfloat16.
- Verified the affected augmentations no longer fail with KeyError: 'BFLOAT16'.
- Updated the bfloat16 `RandomMosaic `numerical expectations for the precision differences in bounding-box coordinates.
Verified the affected tests pass with the updated expectations.

**How did AI help?**

AI was used to help investigate the dtype and numerical-precision failures, identify the affected code paths, and review the test changes. The resulting changes and test behavior were manually checked against the implementation and test output.

~~GitHub Actions currently reports a test failure that does not reproduce in the local test run.~~


ducha-aiki (changes requested):
Reviewed head `eed62ece` (first review).

Thanks -- the enum member and the two dtype fixes are right, the 27 + 6 manifest lines are gone and both half-precision CI legs are green. One regression and a CHANGELOG fix before merge.

**P1 -- `kornia/augmentation/_2d/mix/cutmix.py:126`: float64 lambda is now float32.** `lam_val = w * h / (width * height)` dropped the `.to(input.dtype)` casts, so the lambda is computed by int64 true-divide (float32) for every dtype and only upcast at the `torch.stack`. On base a float64 label got a float64 lambda. Reproduced with `RandomCutMixV2(p=1.0, num_mix=3, data_keys=["input","class"])` on a `6x3x15x17` float64 batch (same seeds on base and head): base `0.1215686274509804`, head `0.12156862020492554`; label tensors differ by up to `3.6e-8`. Power-of-two sizes such as the tests' `16x16` are exact in float32, which is why the suite does not see it. This keeps your float32 computation for the half dtypes and is byte-identical to base for float32/float64 (I ran it: 49/49/95 passed, outputs equal to base):

```python
calc_dtype = input.dtype if input.dtype in (torch.float32, torch.float64) else torch.float32
lam_val = w.to(calc_dtype) * h.to(calc_dtype) / (width * height)
lam = 1 - lam_val if self.use_correct_lambda else lam_val
```

and delete the bare `lam.to(device=..., dtype=...)` on lines 128-131 -- its result is discarded, the effective cast is the one inside `torch.stack`.

**P2 -- CHANGELOG.md:1370-1373.** (a) Cite the PR, `(#4471)`, not the issue. (b) The bullet is indented under the previous SIFT entry; make it a top-level `*` under `### Bug fixes`. (c) "Bounding-box transformations also handle mixed dtypes without failing at the matrix multiplication step" is not in this diff -- with only the enum member patched onto base, `RandomMosaic` with `bbox_xyxy` on bfloat16 already succeeds. Drop the sentence. (d) Mention the user-visible float16 change: `RandomMixUpV2` on float16 input used to return a float32 image (the float32 lambda promoted the product); it now returns float16.

**P3.** The PR body says expected values were adjusted, but the diff loosens tolerances (`test_numerical` boxes to `rtol=1e-2, atol=5e-2` for both half dtypes, mixup lam to `1e-3` for float16). That is fine -- bfloat16 boxes at coordinate 8 have a 0.0625 ULP and I measure 3.7e-2 drift vs float32 -- but say so in the body; the float16 mosaic manifest line is removed by the tolerance alone.

Verified: issue repro fails on base and passes on head; head file green on CPU float32/float64 (95), float16 profile (49), bfloat16 profile (49); base shows exactly 27 + 6 xfails for the deleted lines; head tests fail on base (27 KeyError, 5 assertion); f32/f64 byte-identical to base for MixUpV2, Jigsaw, Mosaic+bbox and for CutMixV2 except the float64 case above; `grep -rnE "#4467|issues/4467" kornia/ tests/` is empty; ruff 0.16.6 clean. No conflict with #4459's hunks in the same test file.

ducha-aiki (changes requested):
Delta re-review of `eed62ece` → `6caaab7d`. While I was writing it you pushed `f8e65846`, a rebase onto current `main` (`dc0af1a4`). Its net patch against `dc0af1a4` is identical to `6caaab7d`'s against `a7e2db6b`, apart from CHANGELOG hunk context, and GitHub now reports it as `MERGEABLE`. So everything below applies to `f8e65846`.

Thanks for the quick turnaround. Both findings from the last review are fixed. **P1** (`cutmix.py:127`): the lambda is computed in `calc_dtype` again and the stray `lam.to(...)` is gone. A float64 label on a `6x3x15x17` batch gives `0.8784313725490196` on head and on main again. Across 12 CutMixV2 and 3 MixUpV2 cases at float32 and float64, with float and int64 labels and both `use_correct_lambda` settings, the output is byte-identical to main. **P2** (CHANGELOG): the entry is now a top-level bullet, cites `(#4471)`, drops the bbox sentence and mentions the float16 MixUp dtype change. What's left comes from #4452, which was merged after your original branch point and is now in your base.

**[P2] The rebased head still carries #4467 references and two half-precision skips that this PR makes stale.** The rebase applied cleanly, so all three stayed behind unnoticed (checked on `f8e65846`):
- `kornia/augmentation/_2d/base.py:56-57` (the Convention block from #4452) still ends with "Individual augmentations may have dtype-specific behavior; see their own docs and `#4467 <…/issues/4467>`_". Once this PR closes #4467, drop the link (the sentence can stay without it).
- `tests/augmentation/container/test_augmentation_sequential.py:742-743` skips `test_wart_mix_children_pass_annotations_through_4493` at bfloat16 with "Tracked in #4467: the mix forward path has no bfloat16 DType". With the skip removed, a merge of this PR into `main` passes at bfloat16/float16/float32/float64 (4 passed). On main it fails at bfloat16 with `KeyError: 'BFLOAT16'`.
- `tests/augmentation/test_base.py:336-337` skips `test_mix_augmentation_blends_cpu_params_with_accelerator_input` at both half dtypes because "RandomMixUpV2 promotes half inputs because mixup_lambdas are float32". Your `mixup.py` change fixes exactly that. With the skip removed it passes at float16 and bfloat16 on that merge (2 passed) and fails on main (`dtype … do not match: torch.float32 != …`).

Please delete both skips and drop the #4467 link. After that, `grep -rnE "#4467|issues/4467" kornia/ tests/` should come back empty.

**[P3]**
- Nothing pins the float64 lambda fix. I put the `eed62ece` line (`w * h / (width * height)`) back on head and the mix tests still pass at float64 (57 passed), because the `16x16` fixtures are exact in float32. You could add a small test: `RandomCutMixV2(p=1.0, data_keys=["input", "class"])` on a float64 `15x17` batch with float64 labels, checking that the label lambda equals `w * h / 255` computed in float64.
- CHANGELOG: the same enum fix also unblocks `PatchMix`, which the issue's updated diagnosis lists. `PatchMix(p=1.0)` on bfloat16 raises `KeyError` on main and returns bfloat16 on head, so it could be named too. Also say what float16 `RandomMixUpV2` used to return (a float32 image). The diff against `main` also has two unrelated whitespace edits: it deletes the blank line between the #4459 and #4463 entries, and adds a second blank line before `## :rocket: [0.6.11]`. Please revert both.
- Carried over: the PR body still says the expected values were adjusted "rather than loosening the numerical checks". The diff loosens tolerances, which is fine, but the body should say so.

**Verified on this head:** CPU `test_augmentation_mix.py` on head: 99 passed at float32/float64, 53 passed under the `cpu-float16` profile and 53 under `cpu-bfloat16` (`--xfail-known-failures`), with zero xfails. On main the same file gives 6 and 27 xfails, so all 33 manifest deletions still hold. I got the same counts on a local merge of `6caaab7d` into main `dc0af1a4`. Also on that merge, the mix/Mosaic/Jigsaw/CutMix subsets of `test_base.py`, `tests/augmentation/container` and `test_bbox.py` pass at float32/float64 (26) and under both half profiles (19 each). The CutMixV2/MixUpV2 byte-identity probe above matches on head and on the merge. Two float16 changes against main: MixUpV2 returns a float16 image (it was float32), and CutMixV2 with `use_correct_lambda=True` and float16 labels now returns the correctly rounded lambda (it was off by up to `2.4e-4`, now equal to the float64 value cast to float16). `ruff` 0.16.6 check and format are clean on the four changed Python files. CI on `6caaab7d` is 38/38 jobs green, including both half-precision legs. Open #4491 also edits `cutmix.py`, but only the `cut_size` docstring, and it merges cleanly with this head.

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (9f876cf3) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4471.fixed.md`.

ducha-aiki (changes requested):
Review of `9f876cf3fabf0be0dbc3a3260b32632d72bed79e`.

The numerical changes and `changelog.d/4471.fixed.md` are sound, but the documented-issue cleanup is still incomplete.

[P2] `kornia/augmentation/_2d/base.py:57`, `tests/augmentation/test_base.py:337`, and `tests/augmentation/container/test_augmentation_sequential.py:743` still describe or skip the exact #4467 dtype defects this PR closes. This leaves half/bfloat16 integration coverage disabled and violates the repository rule that a closing change must re-check every `#NNNN`/`issues/NNNN` hit. Remove the source citation and both stale skips, then run the exposed nodes with the `cpu-float16` and `cpu-bfloat16` known-failure profiles. Earlier merged-tree probing showed all four exposed cases pass here and three fail on main.

msparsh13:
investigated the failing TestSequential::test_forward[cpu-float16-*] cases. The same 5 failures reproduce on the base main (1, (1, 2), 10, True, and False), so these appear to be pre-existing rather than introduced by this PR.

The failures are in the existing CPU-float16 reproducibility check, not in the BF16 DType changes covered by this PR.

ducha-aiki (changes requested):
Review of `d41e526859e26d70f39c7c40bba6c3f46697fd31`.

The last review's P2 is fixed. The `#4467` link is gone from `_2d/base.py`, both stale skips are gone and the exposed nodes pass. The P3 float64 pin is in too: `test_random_cutmix_float64_lambda` fails when the old `w * h / (width * height)` line is put back (`Greatest absolute difference: 2.875e-08`) and passes on head. Three things still block the merge. One is new: the MixUp dtype fix makes the float16 CI leg fail. The other two come from merging main, which brought in #4505 and #4537.

**[P2] Float16 CI leg: the failing node is caused by this PR, not pre-existing.** The job fails with `1 failed, 10144 passed, 563 xfailed, 5 xpassed`. The failing node is `tests/augmentation/container/test_sequential_aug.py::TestSequential::test_forward[cpu-float16-random_apply3]`, i.e. `random_apply=(2,)`. The same job on `9f876cf3` failed on the same node on a different CPU (EPYC 7763 there, EPYC 9V74 now). Main's scheduled float16 leg on the base commit `53866348` passes with `10137 passed, 569 xfailed, 5 xpassed`.

The five failures in your comment (`1`, `(1, 2)`, `10`, `True`, `False`) are a different set. I get exactly those five on both main and head locally (torch 2.14, arm64), with identical failure sets. Under `--xfail-known-failures --known-failure-profile=cpu-float16` both trees leave the same two, `[cpu-float16-1]` and `[cpu-float16-random_apply2]`. Those are specific to my local torch/arch, and CI main passes them. None of that covers `random_apply3`.

Mechanism: in that CI run the container applied `RandomMixUpV2` first, then three `ColorJiggle`s, `MedianBlur` and `RandomAffine` (the params are in the assertion message). On main, MixUpV2 promoted float16 to float32, so the ColorJiggles ran in float32. After the `mixup.py` change they run in float16. There ColorJiggle produces NaN, and `reproducibility_test` compares with `equal_nan=False`. I replayed that chain with the logged params (seed-0 `randn(1, 3, 30, 30)` float16):

| step | main | PR |
|---|---|---|
| after MixUpV2 | float32 | float16 |
| after ColorJiggle 1 / 2 / 3 (NaN count) | 0 / 0 / 0 | 0 / 462 / 468 |
| final | float32, 0 NaN | float16, 2457 NaN |

The float16 ColorJiggle NaN is an older defect that main's float32 promotion happened to hide. So recording it is the right move, not fixing ColorJiggle here. Add this line to `testing/half_precision_xfails/cpu_float16.txt`, in byte order right after `...test_forward[cpu-float16-True]`:

```
call	builtins.AssertionError	tests/augmentation/container/test_sequential_aug.py::TestSequential::test_forward[cpu-float16-random_apply3]
```

**[P2] `tests/augmentation/container/test_augmentation_sequential.py:737-741`: the merge put back the pre-#4505 wart name and comment.** Merge `48114b27` resolved this hunk by keeping your branch's header. Main has `test_mix_children_dispatch_annotation_keys_4493` with the "Regression (#4493)" comment; this PR renames it back to `test_wart_mix_children_pass_annotations_through_4493`. The restored comment says the container "hides both" and the keys "come back unchanged". The body, which is main's, asserts the opposite: `RandomMosaic` boxes are transformed (`assert not torch.equal(out_boxes, boxes)`) and every unsupported key raises `NotImplementedError`. A `test_wart_*` name is what the AGENTS.md grep uses to find open documented defects, and #4493 is closed. Restore main's name and four-line comment, and delete only the two skip lines. The node passes at float16, bfloat16, float32 and float64 either way (4 passed).

**[P2] `README.md:99` and `docs/source/get-started/precision.rst:50-54` still cite #4467.** Both came in with #4537, after the last review. They say `RandomJigsaw`, `RandomCutMixV2`, `RandomMixUpV2` and `RandomMosaic` "raise `KeyError: 'BFLOAT16'`", with a link to #4467, and list `kornia.augmentation` known failures as `205 / 83`. `git grep -nE "#4467|issues/4467"` on the head (which equals the merge tree) returns exactly these two hits; `kornia/`, `tests/` and `testing/` are clean. The count column is a manifest count. `grep -c tests/augmentation/` gives 205/83 on main and 199/56 on head, so 200/56 once the line above is added. Update both counts. Replace the bfloat16 note, e.g. "bfloat16: mostly 3D-augmentation gradient checks (28 of 56 entries are `RandomMotionBlur3D`/`RandomRotation3D` backward)". Leave the dated CI results table in `precision.rst` (commit `ca5021eb`) as it is.

**[P3]** `tests/augmentation/test_base.py:407`: removing the skip left an empty first line in the function body. Ruff 0.16.6 accepts it; delete it anyway.

**Verified on this head**
- Production diff vs main: `DType.BFLOAT16`, the `calc_dtype` lambda in cutmix, the MixUpV2 dtype cast and the base.py prose. It matches `9f876cf3`'s net patch.
- Fragment claims: on main, `RandomMosaic`, `RandomMixUpV2`, `RandomCutMixV2`, `RandomJigsaw` and `PatchMix` raise `KeyError: 'BFLOAT16'` at bfloat16, and float16 MixUpV2 returns float32. On head all five return bfloat16, and MixUpV2 returns float16 for float16 input.
- The two re-enabled `test_base.py` nodes fail at float16 and bfloat16 once `constants.py` and `mixup.py` are reverted to main (2 failed). The re-enabled #4493 node passes at bfloat16 even with main's `constants.py`, so its skip was already stale after #4505, and removing it is still correct.

**What I ran** (`.venv` python 3.11, torch 2.14.0; `kornia.__file__` printed in each probe and naming the PR or main worktree)
- `pytest tests/augmentation/test_augmentation_mix.py tests/augmentation/test_base.py tests/augmentation/container/test_augmentation_sequential.py tests/augmentation/container/test_sequential_aug.py --device=cpu`:
  - `--dtype=float32,float64`: head 494 passed, 11 xfailed; main 493 passed.
  - `--dtype=float16 --xfail-known-failures --known-failure-profile=cpu-float16`: head 2 failed / 288 passed / 23 xfailed; main 2 failed / 280 passed / 29 xfailed. Failure sets identical (the two local-only `test_forward` nodes above).
  - `--dtype=bfloat16` with `cpu-bfloat16`: head 305 passed / 8 xfailed; main 275 passed / 35 xfailed.
- The xfail sets that disappear equal the removed manifest lines exactly (6 float16, 27 bfloat16), with no new xfails.
- Plant-and-revert: old cutmix lambda line (new pin fails); main's `constants.py` + `mixup.py` (re-enabled nodes fail). The worktree was restored afterwards.
- CI logs: PR runs `34881958418` (`d41e5268`) and `34826480445` (`9f876cf3`), main scheduled run `34927215586` (`53866348`).
- `ruff` 0.16.6 `check` and `format --diff` on the seven changed Python files: clean.
- `git merge-tree --write-tree` against main and the listed open PRs: all clean.

ducha-aiki (approved):
Review of `76435b1c`.

All three asks from the last round are in. The `random_apply3` line is at `cpu_float16.txt:33` in byte order, main's `test_mix_children_dispatch_annotation_keys_4493` name and four-line comment are back with only the two skip lines removed, and the README / `precision.rst` counts are numerically right: I recounted `grep -c tests/augmentation/` on the head and get exactly **200 / 56** (main is 205 / 83). The sub-claims check out too — 108 float16 entries are `TestRandomCutMixGen`, and 28 of the 56 bfloat16 entries are `RandomMotionBlur3DBackward` / `RandomRotation3DBackward`.

**bfloat16 support is real, not just non-crashing.** On main, `RandomMixUpV2`, `RandomCutMixV2`, `RandomJigsaw`, `RandomMosaic` and `PatchMix` all raise `KeyError: 'BFLOAT16'`; on the head all five return bfloat16 on CPU with every element finite. Stronger: with integer-valued inputs the bfloat16 Jigsaw and Mosaic outputs are byte-equal to the float64 ones (`5a74b2960020`, `21aa11cda438`) and contain zero values absent from the input, so the rearrangement is correct rather than merely finite; and the bfloat16 MixUp output is bitwise equal to a recomputed `x*(1-lam) + x_perm*lam` (max deviation 0.0). float32 and float64 are bitwise unchanged from main across all 7 probe cases per dtype; float16 changes in exactly the two places the PR intends (MixUp image dtype float32→float16, and the `use_correct_lambda=True` label lambda).

Manifest discipline holds. Planting the 27 deleted bfloat16 lines back gives 27 strict XPASS, and the 6 deleted float16 lines give 6. The added float16 line reports `XPASS(strict)` here on macOS arm64 — the same platform divergence as the two local-only `test_forward` nodes, which also fail on main — but your `half-precision (float16)` leg runs `--verify-known-failures` over the whole `tests/` target on an `AMD EPYC 9V74` and is green (`10320 passed, 557 xfailed`), which is the authoritative proof that the addition and all 33 deletions are exactly right there. Both manifests stay byte-sorted.

The standing #4467 reconciliation item is **discharged**: `grep -rnE "#4467|issues/4467" kornia/ tests/ docs/ README.md testing/` is empty on the head, and also empty on the merge tree against current main (which now contains both #4538 and #4566). Main has exactly 4 hits and this PR removes all 4.

**Approving, with one merge-order item and three nits.**

### [P2 — landing order] This collides with #4550 in `cpu_bfloat16.txt`, and your README count depends on the resolution

It is a real git conflict, not a silent one. Merging both onto `30dfbf711`:

```
$ git merge pr/4471 && git merge pr/4550
CONFLICT (content): Merge conflict in testing/half_precision_xfails/cpu_bfloat16.txt
UU testing/half_precision_xfails/cpu_bfloat16.txt
```

Resolving it "keep both" — your 27 deletions plus #4550's four `..._4468[cpu-bfloat16-shapeN]` additions — makes the bfloat16 leg red: `4 failed, 54 passed`, all four as strict XPASS. Dropping #4550's four lines gives `58 passed` and a bfloat16 augmentation count of **56**, exactly what this PR writes into the README.

So the clean order is **this PR first**; #4550 then just removes its four lines and needs no README change. I have asked for that order there. If #4550 lands first, this PR must delete those four lines in its conflict resolution too, or both the CI leg and the `200 / 56` count are wrong.

### [P3] The float16 tolerance branch is not needed

`tests/augmentation/test_augmentation_mix.py:62-66`. With `rtol, atol = 1e-4, 1e-4` for float16 the node passes on all four dtypes here (`4 passed`) — the float16 failure on main was the float32/float16 dtype mismatch your `mixup.py` fix removes, not a tolerance miss. Please delete the `if/else` and go back to `rtol=1e-4, atol=1e-4`.

### [P3] `atol=5e-2` is looser than the data needs

`tests/augmentation/test_augmentation_mix.py:444-447`. `rtol, atol = 1e-2, 1e-2` passes both half dtypes; at `1e-4` the actual misses are `0.001953125` (float16) and `0.03125` (bfloat16).

### [P3] The fragment over-claims CutMix

"preserve input half-precision dtypes in MixUp and **CutMix**" — CutMix's image *and* label dtypes are unchanged from main at float16/float32/float64 (I hashed them on a `4x3x15x17` batch with fixed seeds). What changes for CutMix is the lambda *value* at float16 with `use_correct_lambda=True` (label hash `40c8302cebb7` → `c3fad163743a`). The fragment also does not mention that MixUp used to return a float32 image. Suggested wording: bfloat16 now works for the mix augmentations; `RandomMixUpV2` on a float16 input used to return a float32 image and now returns float16; and `RandomCutMixV2` computes the lambda in float32 for the half dtypes, so a float16 label gets the correctly rounded value.

Pins verified: `calc_dtype = torch.float32` mutation fails `test_random_cutmix_float64_lambda`; reverting the mixup cast fails the re-enabled `test_mix_augmentation_blends_cpu_params_with_accelerator_input` at float16 and bfloat16; deleting `DType.BFLOAT16` gives 27 failures in the mix file at bfloat16. `BFLOAT16 = 4` appends without renumbering 0-3, so the `docs/export_support/cases_aug.py` value round-trip stays compatible. Suites: mix + base + both containers at float32/float64 → 502 passed (main 501, the +1 being the new pin); bfloat16 profile 309 passed / 8 xfailed (main 279 / 35). Doctests 6 passed, ruff clean, `closingIssuesReferences` exactly `[4467]`.

PR #4470, opened 2026-09-12
Fix remaining PinholeCamera validation guards (#4266)

dhruvv16-hash:
Fixes #4266

This PR resolves the remaining \PinholeCamera\ validation issues outlined in #4266:
- \_check_valid\ now checks for batch size equality across all arguments, rather than testing their batch sizes for truthiness.
- \project\ and \unproject\ now raise the same \ValueError\ as \project_points\ when given a rank-1 point tensor, rather than throwing a bare \IndexError\.

Regression tests in \	ests/geometry/camera/test_pinhole.py\ (the \	est_wart_*\ ones pinning this behavior) have been updated to assert the correct error messages and empty batch construction.

(Note: \_check_valid_params\ and the \pixel2cam\ / \cam2pixel\ checks were already fixed in earlier PRs, so they were left untouched here.)

PR #4475, opened 2026-09-12
fix(camera): unify singular-depth projection policy

victorwon2001:

The current implementation applies guarded singular-depth behavior to the Z1 projection paths and changes `cam2pixel` to use `eps` as a branch threshold rather than adding it to every denominator. The policy-independent `PinholeCamera.project` review regression is also corrected so regular-depth projection preserves full 4x4 intrinsics, higher-rank camera batching, and point dtype.

The complete singular-depth contract is not yet settled: maintainer review still requires an explicit decision covering singular and near-singular depths, Kannala-Brandt/fisheye behavior, NaN depth, and points behind the camera. The repository's towncrier convention is followed with the incompatible behavior documented in `changelog.d/4475.breaking.md` rather than by editing `CHANGELOG.md`.

Related: #4267

## What did you check?

The frozen candidate's affected camera/depth CPU float32+float64 modules passed (492 passed, 1 skipped); targeted float16/bfloat16 regressions passed (4 passed), changed-source doctests passed (11 passed), and focused `ty` and Ruff 0.16.6 checks passed. The breaking fragment passes focused pre-commit and the Towncrier draft preview renders it under Breaking changes; singular/near-singular, NaN-depth, behind-camera, and Kannala-Brandt semantics still await a maintainer policy decision for #4267.

## How did AI help?

AI was used to inspect the projection and changelog review findings, preserve the policy-independent camera corrections, rewrite the changelog fragment to match the repository's breaking-change convention, and verify its formatting and Towncrier rendering.


ducha-aiki (changes requested):
First review of head `2925afa7`.

Thanks for taking this on. The Z1 Jacobian and the `torch.where` guard look clean, and I checked their gradients on torch 2.5.1 and 2.14. My main concern is the policy choice itself. #4267 lists three coherent policies (return `inf`, raise, or keep the mask) and leaves the choice open, together with the question of what to do with points behind the camera. This PR picks "keep the mask" and applies it to all five sites. At the Z1 and `kornia.sensors` sites that turns today's `inf`/`nan` into a finite pass-through. The issue files that pass-through under idiom 4 (a singular input silently giving a plausible value), and a downstream `isfinite` check can no longer catch it. This is an API decision more than a bug fix. Findings 2–4 below are concrete costs of it, so it's worth settling on #4267 which policy we want before iterating on the details. Finding 1 is independent of the policy.

**[P1] `PinholeCamera.project` now ignores every intrinsics entry except `fx, fy, cx, cy`, and pairs cameras with the wrong points for rank > 3 input** (`kornia/geometry/camera/pinhole.py:431-432`). `docs/source/get-started/camera-conventions.rst:237-244` documents that the whole `4x4` takes part: `project` multiplies the full `intrinsics @ extrinsics`, and "a non-zero `intrinsics[0, 3]` shifts every projected `u` by `intrinsics[0, 3] / z`". `project_points` goes through `denormalize_points_with_intrinsics`, which reads only the diagonal and the principal point. Measured on a regular point `X = (1, 2, 4)` with translation `(0.1, -0.2, 0.3)` and `fx = fy = 100, cx = 4, cy = 3`:

| intrinsics | main `u` | head `u` | head `unproject(project(X), z)` |
|---|---|---|---|
| canonical | 29.581 | 29.581 | `[1.0, 2.0, 4.0]` |
| skew `K[0, 1] = 5` | 31.674 | 29.581 | `[0.91, 2.0, 4.0]` |
| `K[0, 3] = 7` | 31.209 | 29.581 | `[0.93, 2.0, 4.0]` |
| `K[2, 2] = 2` | 14.791 | 29.581 | `[1.086, 2.065, 1.85]` |

`unproject` still inverts the full `K @ E`, so the round trip breaks on head. Batching regresses too. With `B = 2` cameras (different `fx`) and points shaped `(2, 2, 5, 3)`, `project` differs from a per-camera loop by 0.0 on main and by **75.7 px** on head: the `(B, 2)` intrinsics are unsqueezed once and broadcast against the `H` axis instead of the camera axis. Points shaped `(2, 3, 5, 3)` raise `RuntimeError: The size of tensor a (3) must match the size of tensor b (2)` on head and work on main. The output dtype also changes: a float64 camera with float32 points returns float64 on head and float32 on main (`transform_points` restores the point dtype). Here is an ordering that keeps `transform_points`' batching and dtype handling and still applies K after the guarded divide:

```python
xy = convert_points_from_homogeneous(transform_points(self.extrinsics, point_3d))
return convert_points_from_homogeneous(transform_points(self.intrinsics, convert_points_to_homogeneous(xy)))
```

I checked it against the per-camera loop (max difference 0.0), with skew and `K[2, 2]` (matches main) and at `z = 0` (`[[104, 203]]`). It still cannot produce the `intrinsics[0, 3] / z` term, so `camera-conventions.rst` needs an update either way. Please add a test with batched rank-4 points and a non-canonical `intrinsics`.

**[P1] `KannalaBrandtK3.project` now puts rays at 90° off-axis on the 45° ring** (`kornia/sensors/camera/projection_model.py:57-58` via `project_points_z1`). A fisheye camera has a correct finite answer at `z = 0`, because Kannala-Brandt distortion works on the angle `atan(r)`. Pass-through of `(x, y)` is a pinhole convention. Setup: `fx = fy = 300, cx = 320, cy = 240`, `k = (0.1, 0.01, 0, 0)`, direction `(0.6, 0.8, z)`. The analytic θ = π/2 pixel is `(689.72, 732.96)`.

| `z` | main | head |
|---|---|---|
| `1e-3` | `(689.35, 732.47)` | `(689.35, 732.47)` |
| `2e-8` | `(689.72, 732.96)` | `(689.72, 732.96)` |
| `1e-8`, `1e-9` | `(689.72, 732.96)` | `(470.63, 440.84)` |
| `0` | `nan` | `(470.63, 440.84)` |

`(470.63, 440.84)` is the pixel for θ = atan(1) = 45°. Main was exact for `0 < z <= 1e-8` and honest (`nan`) at 0. Head gives a plausible wrong pixel for a ray that a >180° fisheye does see. `BrownConradyModel.project` likewise goes from `nan` to a finite `(179, 353)` at `z = 0`.

**[P2] `cam2pixel` still applies K before the mask, so at `z = 0` the namespace still gives two answers and `DepthWarper` and `warp_frame_depth` still split** (`kornia/geometry/camera/pinhole.py:894`, `kornia/geometry/depth.py:529-530`). `cam2pixel` receives the fused `P = K T` and masks `w` after K has been applied. That is exactly what the old `PinholeCamera.project` did, and #4267 calls that divergence "the defect". At the issue's audit point `(1, 2, 0)` with `fx = fy = 100, cx = 4, cy = 3`, head gives `[[104, 203]]` for `project_points` and `PinholeCamera.project` but `[[100, 200]]` for `cam2pixel`. I reran the fixture of `test_warp_frame_depth_and_depth_warper_agree_at_zero_transformed_depth_4267` with `K = [[1, 0, 1], [0, 1, 0.5], [0, 0, 1]]` instead of `I`. `warp_frame_depth` returns `[[2, 3, 0], [5, 6, 0]]` and `DepthWarper` returns `[[0.5, 1, 1.5], [2.5, 3.5, 4.5]]` (the two agree at `z = 0.5`). Three things follow:
- The new `DepthWarper` docstring sentence ("Their projection routes now share the same strict ``abs(z) > 1e-8`` divide guard, including at ``z = 0``") and the CHANGELOG's "share the same singular-depth policy" are false for any K with a non-zero principal point.
- The regression test passes only because `K = I` makes `cx = cy = 0`.
- `Fixes #4267` would close the issue with two answers still in place and the behind-camera question undecided.

A fused `4x4` does not let `cam2pixel` add `cx` back after the mask. Under the mask policy, `DepthWarper` can only agree with `project_points` if it passes K separately, which is one more reason to decide the policy first.

**[P2] A NaN depth now becomes a finite pixel in six more public APIs.** `abs(nan) > 1e-8` is `False`, so a NaN `z` takes the pass-through branch:
- `project_points_z1([[1, 2, nan]])` was `[[nan, nan]]` and is now `[[1, 2]]`.
- `Z1Projection.project` changes the same way.
- `dx_project_points_z1` now returns the identity Jacobian.
- `PinholeModel.project` returns `[[104, 203]]`, `BrownConradyModel.project` `(179, 353)` and `KannalaBrandtK3.project` `(63.1, 121.3)`.

`project_points`/`convert_points_from_homogeneous` already did this on main; the PR spreads it, and nothing documents or pins it. If the mask stays, `mask = ~(torch.abs(z) <= eps)` lets NaN propagate through the divide branch. I checked it keeps `z = 0` → `(x, y)`, `z = 4` → `x / z`, and the gradients of both branches.

**[P2] CHANGELOG** (`CHANGELOG.md:391-394`). The entry cites `(#4267)`; it needs this PR's number, `(#4475)`. It changes the output of five public entry points plus `DepthWarper`, and it changes a default argument: `cam2pixel(eps=1e-12)` becomes `1e-8`, and a caller-supplied `eps` changes meaning from an additive term to a threshold. That belongs under `### Breaking changes` and should say what used to happen:
- `project_points_z1`, `Z1Projection.project` and the `kornia.sensors` models returned `±inf` (`nan` on a zero numerator) at `z = 0` and `x / z` for `0 < |z| <= 1e-8`.
- `PinholeCamera.project` returned the undivided `K (R X + t)`, e.g. `[[100, 200]]` where it now returns `[[104, 203]]`.
- `cam2pixel` returned `x / (z + 1e-12)`: about `1e14` at `z = 0`, `inf` in float16, and a bias on every finite depth.

**Verified on this head**
- **Tests** (torch 2.14.0 unless noted), the six touched test modules:
  - cpu float32 + float64: 743 passed, 19 skipped, 2 xfailed.
  - float16 (`cpu-float16` known-failure profile): 363 passed, 11 xfailed. bfloat16: 365 passed, 12 xfailed.
  - mps float32: 365 passed.
  - torch 2.5.1 cpu float32 + float64: 743 passed. On 2.5.1 the half-precision warp tests fail with `grid_sampler_2d_cpu not implemented for Half`, the same as on main.
- **Base check:** head's test files run against main give 36 failed / 707 passed. The one renamed pin that passes on main, `test_projection_masks_singular_depth_4267`, covers `project_points`, whose behaviour did not change.
- **Before/after table:** 9 entry points × `z ∈ {0, ±1e-9, ±1e-8, ±2e-8, 1e-3, -4, nan}` × float32/float64/float16, on main and head:
  - Outputs change only for `|z| <= 1e-8` (in float16 also `±2e-8`, which rounds to 0) and for NaN `z`.
  - The only change at `±2e-8` in float32/float64 is `cam2pixel` dropping its bias (`4.99975e9` → `5e9`). `1e-3` and `-4` are unchanged everywhere.
- **Gradients:** float64 values and gradients are identical on torch 2.5.1 and 2.14. `gradcheck` passes inside each branch: `z = 0, ±5e-9` with finite-difference eps `1e-10`, and `±3e-8, 1e-3, 0.5`.
- **Compile/export:** `torch.compile(fullgraph=True)` and ONNX (`dynamo=True`, opset 18, onnxruntime) match eager for `project_points_z1`, `dx_project_points_z1`, `cam2pixel` and `PinholeCamera.project` on an input with a `z = 0` row. `torch.jit.script` works for the three functions.
- **Other checks:**
  - Doctests of the six modules: 25 passed.
  - `ruff@0.16.6` check/format is clean.
  - `grep -rnE "#4267|issues/4267" kornia/ tests/` leaves no hits in `kornia/`; the seven hits in `tests/` are the new regression tests.
- **Overlap:** no other PR targets #4267. #4470 (already conflicting) edits `PinholeCamera.project` a few lines from this change.
- **CI and merge state:** 40/40 green, mergeable, 4 commits behind main (augmentation only).

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (b5adfad6) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4475.fixed.md`.

ducha-aiki (changes requested):
Review of `b5adfad6fa637fa681205c28c212c61f0d16e367`.

The Towncrier migration commit does not address the current-head blockers. [P1] `PinholeCamera.project` now delegates to `project_points` at `kornia/geometry/camera/pinhole.py:431-432`; that helper intentionally ignores skew and only reads `fx`, `fy`, `cx`, and `cy`, while the old `K @ E` path honoured the complete stored 4x4 intrinsics. It also changes broadcasting for rank-4 point batches. A camera with nonzero skew therefore projects to the wrong pixel and no longer round-trips through `unproject`. Preserve the full intrinsics semantics while applying the singular-depth guard. [P1] Reusing the Z1 pass-through at `kornia/sensors/camera/projection_model.py:58` also changes the Kannala-Brandt path at `abs(z) <= 1e-8`: a near-singular 3-D ray is first replaced by raw `(x, y)` and then fisheye-distorted onto a roughly 45-degree ring, rather than approaching the ray's fisheye limit. The singular policy must be defined at the complete camera-model level, not imposed before every distortion model. [P2] This intentionally changes public results at singular depth and changes `cam2pixel`'s default/finite-depth arithmetic, so `changelog.d/4475.fixed.md` must be a `breaking` fragment and describe the old and new outputs.

Validation and limits: Head `b5adfad6` adds only a merge of main and the Towncrier migration over reviewed implementation `2925afa7`; the substantive patch is unchanged. Previous reproductions measured the skewed Pinhole result as 29.581 instead of 31.674 and the Kannala-Brandt singular result as `(470.6, 440.8)` instead of `(689.7, 733.0)`.

ducha-aiki (changes requested):
Review of `f9c754b32f8027595e6ef0df40e355c29929fc1d`.

`f9c754b3` is titled `docs(changelog)`, but it also rewrites `PinholeCamera.project`, updates the `DepthWarper` docstring and adds two tests. The earlier `b5adfad6` review was written against the delegating version. The Pinhole regression is fixed, and every factual claim in the new breaking fragment checks out. The Kannala-Brandt singular-depth regression and the NaN-depth change are untouched since `2925afa7`.

**Resolved**

- **Pinhole full intrinsics, batching, dtype** (`kornia/geometry/camera/pinhole.py:433-445`). Regular-depth output is identical to main. Setup: `X = (1, 2, 4)`, `t = (0.1, -0.2, 0.3)`, `fx = fy = 100`, `cx = 4`, `cy = 3`.

  | intrinsics | main `u` | head `u` |
  |---|---|---|
  | canonical | 29.581 | 29.581 |
  | skew `K[0, 1] = 5` | 31.674 | 31.674 |
  | `K[0, 3] = 7` | 31.209 | 31.209 |
  | `K[2, 2] = 2` | 14.791 | 14.791 |

  - With `B = 2` cameras (different `fx`, skew), points shaped `(2, 2, 5, 3)` and `(2, 3, 5, 3)` differ from a per-camera loop by 0.0 on both trees, including when `z = 0` rows are mixed in.
  - A float64 camera with float32 points returns float32 on both.
  - The two new `_4475` tests fail when `b5adfad6`'s `pinhole.py` is planted back (3 failures) and pass on head.
- **Release note.** `changelog.d/4475.breaking.md` replaces the `fixed` fragment. Its claims match measurements (main → head):
  - `project_points_z1` at `z = 0`: `[inf, inf]` (`[nan, inf]` when the numerator is zero) → `(x, y)`.
  - `project_points_z1` at `z = 5e-9`: `[2e8, 4e8]` → `[1, 2]`.
  - `PinholeCamera.project` at `(1, 2, 0)`: `[100, 200]` → `[104, 203]` in float32, float64 and float16.
  - `cam2pixel` default `eps`: `1e-12` → `1e-8`.
  - `cam2pixel` at `(1, 2, 0)`: `[1e14, 2e14]` in float32/float64 and `[inf, inf]` in float16 on main, `[100, 200]` on head.
  - `cam2pixel` bias at `z = 1e-3` (float64): `100003.9999` → `100004.0`.
- **Closing references and docs.** `closingIssuesReferences` is empty (the body says `Related: #4267`). The `DepthWarper` docstring now states that the two warps still split at `z = 0` with a principal point. Re-measured with `K = [[1, 0, 1], [0, 1, 0.5], [0, 0, 1]]`: `warp_frame_depth` returns `[[2, 3, 0], [5, 6, 0]]` and `DepthWarper` returns `[[0.5, 1, 1.5], [2.5, 3.5, 4.5]]`, as documented.

**Still open**

**[P1] Kannala-Brandt still maps near-singular rays to the 45° ring** (`kornia/sensors/camera/projection_model.py:58`, unchanged). Setup: `fx = fy = 300`, `cx = 320`, `cy = 240`, `k = (0.1, 0.01, 0, 0)`, direction `(0.6, 0.8, z)`. The analytic θ = π/2 pixel is `(689.72, 732.96)`.

| `z` | main | head |
|---|---|---|
| `2e-8` | `(689.72, 732.96)` | `(689.72, 732.96)` |
| `1e-8` | `(689.72, 732.96)` | `(470.63, 440.84)` |
| `1e-9` | `(689.72, 732.96)` | `(470.63, 440.84)` |
| `0` | `nan` | `(470.63, 440.84)` |

`(470.63, 440.84)` is exactly the θ = 45° pixel. The fragment's "#4267 remains open for fisheye" does not cover this: the PR itself turns a correct answer into a plausible wrong one for `0 < z <= 1e-8`. Either keep `KannalaBrandtK3` off the Z1 pass-through, or define the singular case on the angle.

**[P2] NaN depth still becomes a finite output, and the fragment implies NaN semantics are unchanged** (`kornia/geometry/camera/projection_z1.py:61-65`). `abs(nan) > 1e-8` is `False`, so NaN takes the pass-through branch.

| API | main | head |
|---|---|---|
| `project_points_z1([[1, 2, nan]])` | `[[nan, nan]]` | `[[1, 2]]` |
| `dx_project_points_z1` | contains `nan` | identity Jacobian |
| `KannalaBrandtK3.project` (setup above) | `nan` | `(497.44, 594.89)` |

`PinholeCamera.project` stays `nan` on both trees, because the extrinsics multiply propagates the NaN. The fragment says #4267 "remains open for … NaN-depth … semantics", but this PR already changes them. Either use `mask = ~(torch.abs(z) <= 1e-8)` so NaN propagates, or list the change in the breaking fragment.

**Policy.** The pass-through at Z1 and in `kornia.sensors` is still one of the three options #4267 leaves open, and the PR body acknowledges this. It needs a maintainer decision on #4267 before merge.

**[P3] `PinholeCamera.project` is about 3x slower at regular depth** (`pinhole.py:435-445`). Every call now evaluates both branches: three matrix transforms, three homogeneous conversions and a `torch.where`. Medians of 21 runs, float32 CPU, 4 threads, two alternating runs per tree:

| N | main | head |
|---|---|---|
| 1e3 | 0.072 / 0.079 ms | 0.211 / 0.217 ms |
| 1e5 | 2.13 / 2.54 ms | 7.11 / 7.42 ms |
| 1e6 | 17.6 / 18.6 ms | 56.4 / 59.7 ms |

Nonblocking, but it is worth sharing the camera-frame transform between the two branches.

**CI.** Both macOS torch 2.5.1 legs fail only `tests/contrib/test_super_resolution.py::TestSuperResolutionBuilders::test_to_onnx_exports[RRDBNetBuilder]` with `Timeout (>120.0s)` (1 failed, 10383 passed). The run tested merge commit `f7ab898` against `3c37d390` (#4499), which predates #4515. On current main that parameter is skipped on darwin with torch < 2.9. The `collector` failure follows from those two legs.

**What I ran** (torch 2.14.0 / py3.11; `kornia.__file__` checked to name the PR worktree or the main worktree in every probe)

- **Probes:** same heredoc on head and main for regular-depth intrinsics, rank-4 batching, dtype, Kannala-Brandt, fragment claims, NaN and timing.
- **Touched test modules, cpu float32+float64:** 750 passed, 19 skipped.
- **Half precision** (`test_pinhole.py`, `test_projections.py`, `test_depth.py`, known-failure profiles): float16 293 passed, 9 xfailed; bfloat16 295 passed, 10 xfailed.
- **torch 2.5.1:**
  - `test_pinhole.py`, `test_projections.py` and `test_projection_model.py`, float32+float64: 310 passed, 12 skipped.
  - `PinholeCamera.project` gradients with skew and `K[0, 3] = 7` at `z ∈ {0, ±5e-9, ±4}` are finite, and `gradcheck` passes (also on 2.14).
- **Compile:** `torch.compile(fullgraph=True)` of `PinholeCamera.project` on input with `z = 0` rows matches eager (max diff 0.0).
- **Plant-and-revert:** `b5adfad6`'s `pinhole.py` fails the new rank-4 test (float32/float64) and the dtype test. Main's `pinhole.py` fails the two `_4267` singular-depth pins.
- **Merge:** `git merge-tree` against main is clean. With the merged versions of `pinhole.py`, `depth.py`, `linalg.py` and their tests overlaid, the camera, depth, linalg and sensors suites give 920 passed, 19 skipped.

ducha-aiki:
The #4267 policy is now decided: https://github.com/kornia/kornia/issues/4267#issuecomment-5678077316. This PR already takes that direction (a masked divide with no additive epsilon), so it can land as the implementation once it covers the points below. My review of `f9c754b3` otherwise stands.

1. **Rebase.** The head now conflicts with `main` in `kornia/geometry/camera/perspective.py` (#4501 added a conventions link there).
2. **[P1] Kannala-Brandt.** Compute `θ = atan2(‖(x, y)‖, z)` from the camera-frame ray instead of sending it through the Z1 pass-through. At `fx = fy = 300`, `c = (320, 240)`, `k = (0.1, 0.01, 0, 0)` and ray `(0.6, 0.8, z)`, `z = 1e-8`, `1e-9` and `0` must give the θ = 90° pixel `(689.72, 732.96)`, not the 45° pixel `(470.63, 440.84)`.
3. **[P2] NaN depth.** Select the divided branch with `~(torch.abs(z) <= eps)`, so `nan` stays `nan`. On this head, `project_points_z1([[1, 2, nan]])` returns `[[1, 2]]`, where `main` returns `[[nan, nan]]`.
4. **One intrinsics ordering.** `cam2pixel` must follow `project_points` (mask, then `K`), so the two depth warps agree at `z = 0` with a nonzero principal point. With `K = [[1, 0, 1], [0, 1, 0.5], [0, 0, 1]]`, this head still gives `warp_frame_depth` → `[[2, 3, 0], [5, 6, 0]]` and `DepthWarper` → `[[0.5, 1, 1.5], [2.5, 3.5, 4.5]]`. With that settled, the `DepthWarper` docstring caveat and its #4267 pin can go.
5. **Tests and docs.** Replace the seven `_4267` wart pins listed in the decision with convention tests, and rewrite the `#4267` prose to state the rule. Then `Fixes #4267` is appropriate.
6. **Out of scope here.** Behind-camera points keep today's projection arithmetic. Validity masks and what the warps sample at invalid pixels are #4555 and should not be pre-empted in this PR.

Nonblocking: share the camera-frame transform between the two branches of `PinholeCamera.project`; evaluating both made it about 3x slower at regular depth.

ducha-aiki:
@victorwon2001 you have been blocked for posting unsolicited reviews in other people PRs and ignoring maintainer requests to not do it 
https://github.com/kornia/kornia/pull/3935#issuecomment-5740442066
It also looks like you are using AI, but do not disclose that for your comments, which is against our policy.

PR #4484, opened 2026-09-13
fix(augmentation): avoid retaining images for lazy transform matrices

### LINKED ISSUE #4482: Lazily computed transform_matrix keeps the last input batch alive, and pickles carry it
Augmentations that compute `transform_matrix` lazily keep a reference to the last input batch until the matrix is read. That covers every 2D intensity augmentation, both flips, `Resize`/`LongestMaxSize`/`SmallestMaxSize`, and `RandomResizedCrop` in slice mode. In a training loop that never reads the matrix, each module pins one full batch (on GPU, device memory), and `pickle` / `copy.deepcopy` / `torch.save` of the module carry that batch (main `a7e2db6b`, torch 2.14.0, CPU):

```python
import pickle
import torch

flip = K.RandomHorizontalFlip(p=1.0)
flip(torch.rand(8, 3, 256, 256))
len(pickle.dumps(flip)), flip._lazy_matrix_args[0].shape   # (6293307, torch.Size([8, 3, 256, 256]))
flip.transform_matrix
len(pickle.dumps(flip))                                    # 1854
```

**Cause.** #3820 ("lazy transform matrix for all matrix-ignoring ops") stores `(input, params, flags)` in `_lazy_matrix_args` (`kornia/augmentation/_2d/base.py:89-96`, written through `_commit_state` in `kornia/augmentation/base.py:237-238`) so that `transform_matrix` can be computed on first access. Only the input's shape, dtype and device are needed for the matrices these classes produce, but the whole tensor is kept.

**Expected.** The lazy state keeps only what `compute_transformation` needs (shape, device, dtype, and the params). For example, keep a zero-size or meta-device stand-in rather than the batch, so memory use and serialized size don't depend on the last input.

Found while reviewing #4452, whose new serialization text says `pickle` and `copy.deepcopy` carry `_params` and `transform_matrix`. They also carry the batch.

NaCr05:

Built-in augmentations that defer their transformation matrix no longer keep the last image batch alive just to compute that matrix later. In a local horizontal-flip reproduction with an `(8, 3, 256, 256)` float32 batch, the input becomes collectible without reading the matrix, and the unread module's pickle shrinks from 6,293,295 bytes on the original base (`a7e2db6`) to 1,601 bytes on the export-fixed head before this documentation rebase.

The pending state stores shape, dtype and device alongside the existing parameters. A one-element expanded tensor supplies that metadata only when the matrix is requested. Matrix computation remains lazy, and parameter gradient connections are preserved. Shapes are stored as plain tuples so PyTorch 2.5's non-strict exporter can snapshot pending state after an eager forward; `torch.Size` in that state caused the earlier CI failure.

The optimization is attached to audited method implementations. Overrides of `transform_tensor`, `generate_transformation_matrix`, `compute_transformation`, or `identity_matrix` retain the original input-based lazy path, including wrappers using `functools.wraps`. An unchanged inherited implementation uses compact state. This preserves custom transforms that need pixel values, although those custom implementations may still retain their input.

Adds regression coverage and keeps existing serialized pending states containing a real tensor readable. Documentation uses main's existing Serialization subsection, with matching base-class Convention blocks and Conventions & Pitfalls guidance. All four override methods are named, and built-in metadata state is distinguished from the input retained by custom lazy subclasses.

Fixes #4482.

## What did you check?

For the export regression, unmodified main passed the `LongestMaxSize` ONNX test while the previous PR head failed in the same Linux environment. The new horizontal-flip, invert and longest-side-resize regression cases all failed before the tuple-shape correction and pass afterward. They exercise an eager call followed by non-strict export with an unread matrix, and compare exported outputs for both the original and a second input.

Validation after rebasing onto `upstream/main` at `ca5021e` and integrating the documentation:

- **Windows, Python 3.14.6, PyTorch 2.14.0+cpu:** lazy-state tests, augmentation base tests, deterministic `torch.export`, and ONNX export coverage: **245 passed, 8 xfailed** on CPU float32/float64.
- **Linux / WSL2, Python 3.13.11, PyTorch 2.5.1+cpu:** lazy-state tests, augmentation base tests, and ONNX export coverage: **236 passed, 8 xfailed** on CPU float32/float64. The worktree and interpreter were verified in both environments.
- Full Sphinx HTML build with warnings treated as errors: passed in Python UTF-8 mode. The rendered pages keep Serialization under Some Further Notes, contain all four override methods, and have neither the extra headings nor stale #4482 links. Docs-only dependencies were installed outside the repository environment.
- Full pre-commit checks, typecheck and `git diff --check`: passed. Source/test syntax trees with docstrings excluded match the reviewed head; this update changes documentation, not executable behavior.

Validation before this documentation rebase, with the imported worktree path and interpreter verified:

- **Linux / WSL2, Python 3.13.11, PyTorch 2.5.1+cpu:** lazy-state tests, augmentation base tests, and the full ONNX export-coverage file: **160 passed**. The relevant ONNX dependencies match the failed CI job: ONNX 1.22.0, ONNX Runtime 1.29.0, ONNX Script 0.7.1. This reproduces the core CI versions, not the complete runner image.
- **Windows, Python 3.14.6, PyTorch 2.14.0+cpu:** lazy-state and augmentation base tests: **146 passed**; deterministic `torch.export` and ONNX export-coverage suites: **23 passed**; selected legacy ONNX export/runtime cases: **8 passed**.
- Focused lazy-state file under each CPU float16/bfloat16 profile: **44 passed** per profile. Each run includes the three CPU float32 export cases and a float64 gradcheck. These local environments differ from the Linux/Python 3.11/PyTorch 2.9.1 half-precision CI profiles; no manifests were changed.
- Dynamo tests: **2 passed**. Flip/invert/resize probes with `torch.compile(backend="eager", fullgraph=True)` each produced one graph and zero breaks, with matching images and matrices. These are correctness checks, not Inductor performance measurements.
- Verified input collection, compact pickle size, and successful loading and matrix evaluation of a trusted unread-state pickle generated on untouched main. The regression suite also covers content-dependent overrides, copied method attributes, inherited implementations, parameter replay, gradient connections, serialization, changing shapes, and container routes.
- Full pre-commit checks, typecheck, and `git diff --check`: passed.

Metadata bookkeeping has an overhead; prior CPU timing pairs varied substantially, so this change does not claim a throughput improvement. CUDA/MPS and the full remote matrix are left to CI.

## How did AI help?

Codex assisted with source inspection, implementation, regression design, documentation, and validation. It reproduced the export failure against untouched main, traced the failing non-strict capture path, and verified the new tests fail before the correction. Validation scope and limitations are reported above.


ducha-aiki (changes requested):
First review of head `0146e15f`. Thanks @NaCr05, this is a careful fix. The code change holds up everywhere I probed it: matrices are bitwise unchanged, the input is released, and pickles shrink. What blocks the merge is the rebase onto #4452 and the #4482 text that #4452 left in the library and docs.

**[P2] Rebase needed, and #4452's #4482 wording has to change with it.** GitHub reports the PR as conflicting. `git merge-tree origin/main 0146e15f` stops in `docs/source/augmentation.base.rst`. This PR adds the `Lazy matrices and saved state` section and a new `Custom geometric transformations` heading directly above the paragraph ("For a rigid **geometric** augmentation, ...") that #4452 rewrote. `CHANGELOG.md` also needs a keep-both merge. The rest auto-merges, including `kornia/augmentation/_2d/base.py`. But on the merged tree, `grep -rnE "#4482|issues/4482" kornia/ docs/` still finds four places that describe this bug as current behaviour:

- `kornia/augmentation/_2d/base.py`, `AugmentationBase2D` Convention: "saved state can retain data from the last call (#4482)"
- `kornia/augmentation/_2d/base.py`, `RigidAffineAugmentationBase2D` Convention: "retained state has the serialization limitations in #4482"
- `docs/source/augmentation.base.rst`, `Serialization`: "lazy matrix state can retain the last input batch in a pickle or deepcopy (#4482)"
- `docs/source/get-started/conventions.rst`, "Serializing an augmentation": "Lazy matrix state can retain the last input batch, increasing serialized size until the matrix is read (#4482)"

After this PR those statements are false for every built-in lazy class, and they would link to a closed issue. When you rebase, please:

- Fold your new section into main's existing `Serialization` subsection, replacing the #4482 sentence there, rather than adding two new `---` headings inside "Custom Augmentation Classes".
- Reword the two Convention bullets and the `conventions.rst` bullet. Built-in lazy matrices keep only shape, dtype and device. A subclass that overrides a matrix-path method keeps the input until the matrix is read.
- Name the four methods whose override selects the old path (`transform_tensor`, `generate_transformation_matrix`, `compute_transformation`, `identity_matrix`). The current text only says "another matrix-path method".

**[P3] Per-call cost of the implementation check (optional).** `apply_func` builds four bound methods and compares their marker on every forward, although the answer only changes with the class or an instance-level override. Single-thread timings on `RandomHorizontalFlip(p=1.0)` with a `(1, 3, 32, 32)` input, main vs this head, measured twice:

| | main | this head |
|---|---|---|
| forward | 17.5 µs | 19.0 µs (+8%) |
| forward + `transform_matrix` | 37.2 µs | 41.2 µs |

The check plus the metadata tuple alone measures 1.0 µs. #3820 added the lazy path for speed, so it could be worth caching the result per `type(self)`, plus a `self.__dict__` check for instance overrides. The absolute cost is negligible at realistic batch sizes, so this is optional.

**CI:** the only red leg on this head is macOS py3.12 / torch 2.5.1. It is the known `tests/contrib/test_super_resolution.py::TestSuperResolutionBuilders::test_to_onnx_exports[RRDBNetBuilder]` 120 s timeout, which also turns `collector` red. The previous head's failure was `tests/onnx/test_export_coverage.py::test_onnx_export_matches_eager[LongestMaxSize]` on ubuntu py3.13 / torch 2.5.1, and your third commit fixes it (details below).

**Verified on this head** (torch 2.14.0 unless noted, CPU):

- **The export fix.** On py3.13, torch 2.5.1's dynamo is unavailable, so the ONNX exporter falls back to non-strict `torch.export`. That path snapshots module attributes through pytree, which rebuilds a `torch.Size` as a `tuple`. With `tuple(in_tensor.shape)` put back to `in_tensor.shape`, the three `test_export_with_unread_matrix` cases fail on torch 2.5.1 with `ValueError: Node type mismatch; expected <class 'tuple'>, but got <class 'torch.Size'>`. They pass on 2.14 either way, so the 2.5.1 CI legs are what enforce this pin. After a non-strict export, `transform_matrix` still reads correctly on both torch versions. `tests/onnx/test_export_coverage.py` passes 14/14 on 2.14 and on 2.5.1.
- **Which classes take the compact path.** Walking every `RigidAffineAugmentationBase2D` subclass finds 43 lazy classes: 37 intensity classes (`Normalize`/`Denormalize` included), both flips, `Resize`/`LongestMaxSize`/`SmallestMaxSize` and slice-mode `RandomResizedCrop`. `RandomDissolving` was not exercised because it needs `diffusers`. All 43 store `_InputMetadata`; none fall back to the tensor. `_3d` has no lazy path and is untouched.
- **Bitwise parity with main.** For all 43 classes, I compared `transform_matrix` with `torch.equal` against main, checking dtype and `requires_grad` too:
  - float32, float64, float16 and bfloat16
  - `p=1` and `p=0.5` at B=6, with `same_on_batch` both ways
  - reading the matrix twice (the same object both times)
  - a second forward with a different B and H×W, and an unbatched `keepdim` input
  - `inverse()` on the flips and the three resizes
  - CPU autocast at bfloat16

  All 2101 matrix and inverse fields are identical. The same holds for:
  - an `AugmentationSequential` chain (hflip, ColorJiggle, vflip, RandomAffine, Resize, with mask, boxes and keypoints) plus its `inverse`
  - a slice-`RandomResizedCrop` chain
  - `transformation_matrix_mode="silent"`
  - `VideoSequential`

  Image outputs differ only for `RandomPlasma*`, which #4462 changed on main, and for `ColorJiggle` at float16, which differs between two runs on main as well.
- **What is stored, and the marker mechanism.** The pending state is `(_InputMetadata(shape, dtype, device), params, flags)`, and the placeholder input is built with `torch.empty(()).expand(shape)` only when the matrix is read. No weakrefs are involved: the marker is a function attribute holding the function itself, compared against the bound method's `__func__`. A plain override and a `functools.wraps` override both fall back to the old path, and the tests cover both.
- **`torch.compile`.** `torch._dynamo.explain` graph and break counts match main for hflip, Resize, ColorJiggle and slice RRC. `fullgraph=True` with the eager and inductor backends, `dynamic=None` and `dynamic=True`, over three input shapes gives the same results on both trees. On torch 2.5.1 a compiled forward does not record state on either tree, so that is not new.
- **Serialization.** Pickle size after an unread `(8, 3, 256, 256)` forward, main vs this head:

  | class | main | this head |
  |---|---|---|
  | hflip | 6,293,307 B | 1,609 B |
  | ColorJiggle | 6,298,051 B | 6,353 B |
  | slice RRC | 6,297,182 B | 5,484 B |
  | LongestMaxSize | 6,295,356 B | 3,658 B |

  The input is garbage-collected on head. Pickles written on main load on this head and give identical matrices. A bonus: `copy.deepcopy` after a forward on a non-leaf input that requires grad raises `RuntimeError` on main and works here.
- **Tests.**
  - `tests/augmentation/test_lazy_matrix.py`: 84 passed at float32/float64 on torch 2.14 and on 2.5.1; 44/44 at float16 and at bfloat16 with the `cpu-float16`/`cpu-bfloat16` profiles; MPS float32 40 passed, 1 skipped; `test_dynamo` with `KORNIA_TEST_OPTIMIZER=inductor` passed.
  - The same file copied onto main: 28 of 84 fail. These are the input-release, serialized-size, grad-graph-release and container-release pins.
  - Augmentation suites (`test_base`, `container/`, `_2d/`, `test_torch_export`, `test_onnx_export`, `test_augmentation`): 1707 passed.
  - `ruff@0.16.6` `check` and `format --diff` are clean.

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (2395bfd9) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4484.fixed.md`.

ducha-aiki (approved):
Review of `2395bfd9d59c6f0992ff6e192ec69dbaca32aba6`.

Re-reviewed current head `2395bfd9`. The previous documentation conflict and the five stale #4482 statements are resolved. Metadata-only lazy state is limited to implementations explicitly audited as reading shape, dtype, and device; custom overrides retain the input-backed path. Coverage includes old-state serialization, non-strict export on PyTorch 2.5.1, custom pixel-dependent overrides, parameter gradients, half precision, and Dynamo. The release note is now `changelog.d/4484.fixed.md`, with no `CHANGELOG.md` edit. I found no actionable defect. Approve; merge after required checks finish green.

Validation limits: this follow-up used source/delta inspection and retained prior numerical review evidence; the author’s new test counts and hosted checks were inspected, not rerun locally during this sweep.

PR #4485, opened 2026-09-13
fix(augmentation): preserve skipped RandomCrop annotations

### LINKED ISSUE #4473: RandomCrop with padding and p<1 shifts keypoints and boxes on rows where the crop was not applied
`RandomCrop` with `padding` set and `p < 1` shifts the keypoints and boxes of *unapplied* rows by the padding offset while returning those images untouched (main `5be4852b`, torch 2.14.0, CPU):

```python
import torch
from kornia.augmentation import AugmentationSequential, RandomCrop
x = torch.zeros(1, 1, 6, 10)
kp = torch.tensor([[[3.0, 2.0]]])
seq = AugmentationSequential(RandomCrop((4, 8), padding=(1, 2), p=0.0), data_keys=["input", "keypoints"])
out = seq(x, kp)
torch.equal(out[0], x)   # True  -- image not cropped, not padded
out[1]                   # tensor([[[4., 4.]]])  -- keypoint moved by (+1, +2)
```

The image side of `RandomCrop` applies the pad only on rows where `batch_prob` fires, but `apply_transform_keypoint` (`kornia/augmentation/_2d/geometric/crop.py:191`) pads unconditionally, so on an unapplied row the label no longer corresponds to the pixels. The same offset reaches boxes. `p=1.0` is consistent; the bug is only visible with `p < 1` (or a batch where some rows are skipped).

**Expected:** the keypoint/box pad translation is applied only on rows where the crop was applied, mirroring the image path, so a label round-trip through `inverse` (or an identity check on unapplied rows) holds.

Found by the geometry audit in #4465, whose `RandomCrop(p=0.5)` round-trip error of `sqrt(5)` on unapplied rows is exactly this `(1, 2)` offset; the audit follows the pixels, which is the right contract, so this is a container defect rather than an audit defect.

NaCr05:

A skipped padded `RandomCrop` currently leaves images and masks unchanged but translates keypoints and boxes. For example, `padding=(1, 2), p=0` moves a keypoint from `(3, 2)` to `(4, 4)` even though the image and transform matrix are unchanged.

Clone annotations inside the transformed crop branch before padding. `pad()` mutates its input, so this keeps the original coordinates available to the non-transform branch during probability blending. The public in-place padding methods are unchanged. The RandomCrop docstring now explains whole-batch probability and skipped annotation behavior.

The regressions compare position-coded pixels, masks and explicit coordinates; cover padding variants, applied/skipped batches, same-size mixed forward masks, bbox representations and ranks, object metadata, parameter replay, label inverse, a subsequent flip, gradients and Dynamo.

Fixes #4473, which was found by the audit in #4465.

The fix covers the skipped-image/shifted-annotation defect reported in #4473, including public whole-batch `p` behavior and same-size mixed forward selection. As confirmed in review, manually supplied mixed masks with shape-changing crops and mixed inverse have separate image-blending and size/unpadding defects to track independently. Those cases remain outside this fix. Crop inverse also cannot reconstruct discarded image content.

## What did you check?

Windows, Python 3.14.6, PyTorch 2.14.0+cpu. Base and patch imports were checked against their respective worktrees.

After rebasing onto `upstream/main` at `ca5021e` for this review update:

- RandomCrop, the augmentation containers, Boxes and Keypoints: **1,855 passed, 469 skipped, 10 xfailed** on CPU float32/float64, including the new regressions.
- Full `pre-commit-all`, `typecheck`, and `git diff --check` passed.
- The crop implementation and regression file are unchanged from the reviewed head. The CHANGELOG keeps main's entries and now describes the previous left/top annotation shift with both #4473 and #4485.

Original implementation validation before the rebase:

- New tests on untouched `a7e2db6`: **52 failed, 40 passed**. On the fix: **92 passed**, including float64 gradcheck for image, keypoint and box inputs.
- RandomCrop, the augmentation containers, Boxes and Keypoints: **1,438 passed, 107 skipped, 7 xfailed** (includes the new tests).
- Focused float16 and bfloat16 runs: **47 passed each**, using the repository's CPU half profiles. These include two float64 gradcheck cases; no manifest entries were changed. Profile metadata differs from this local Windows/Torch environment.
- Dynamo with the eager backend: **2 passed**. RandomCrop docstring examples: **10 passed**.
- Full `pre-commit-all`, `typecheck`, and `git diff --check` passed.

```text
pixi run -e py314 test-module tests/augmentation/container/test_random_crop.py --device=cpu --dtype=float32,float64
pixi run -e py314 test-module tests/augmentation/test_augmentation.py::TestRandomCrop tests/augmentation/container/ tests/geometry/test_boxes.py tests/geometry/test_keypoints.py --device=cpu --dtype=float32,float64
pixi run -e py314 typecheck
```

The mixed-forward pixel test uses nearest sampling to isolate row selection from half-precision bilinear errors at padded edges. The same bilinear image errors were reproduced on untouched base and patch; other tests retain default bilinear sampling. CUDA, MPS and Inductor were not tested locally.

## How did AI help?

Codex assisted with tracing the mutation, designing and implementing the regressions, the focused fix, documentation and local validation. The failing baseline tests, fixed-coordinate pixel checks, base/patch interpolation comparison and checks above were executed before submission.


ducha-aiki (changes requested):
First review of head `8f475055`.

Thanks @NaCr05, this is a clean, minimal fix with a careful test file. `Keypoints.pad` and `Boxes.pad` work in place. On main, `apply_transform_keypoint` / `apply_transform_box` padded the container that `apply_non_transform_*` then returned, so the skipped branch of the blend was already shifted. Cloning before the pad removes that aliasing, and the fix is per row: `transform_keypoints` / `transform_boxes` still `torch.where`-blend each row, and now the skipped rows get their untouched coordinates. I could not find any path where the fix changes something it should not. The only asks are about how the PR is linked.

**[P2] This fixes #4473 as reported, so please close it with `Fixes #4473` and cite it in the CHANGELOG.** #4473 is about rows where the crop is not applied: the image comes back untouched but the keypoints and boxes move by the left/top padding. After this PR I could not build such a row on any path:
- a skipped call, which is what `p < 1` gives, since `p` is `p_batch` here;
- a hand-built mixed `batch_prob=[1, 0, 1]` with a same-size crop, where every row's keypoint lands on its own pixel landmark (row 1 is off by `(1, 2)` on main);
- `VideoSequential(RandomCrop(..., p=0))`;
- the skipped-call `inverse` round trip (error `2.0` on main, `0.0` here).

The two cases you left out are separate defects, not leftovers of #4473:
- A mixed batch with a crop that changes the shape takes the `_blend_by_prob` fallback, which returns the cropped image for every row, skipped rows included. The skipped image is changed there, and other shape-changing augmentations (`Resize`) behave the same way.
- A mixed `inverse` raises `shape mismatch` in `GeometricAugmentationBase2D.inverse_inputs`. `inverse_keypoints` / `inverse_boxes` skip `unpad` for the applied rows unless `batch_prob.all()` (`crop.py:319`, `crop.py:337`). This fails the same way on main.

Both need hand-edited params, because `p` gates the whole batch. We'll track them in their own issue, so they need not hold up closing #4473. In `CHANGELOG.md`, please cite the issue and say what used to happen, for example: "`RandomCrop` with `padding` or `pad_if_needed` no longer shifts keypoints and bounding boxes by the left/top padding when the crop is skipped; they previously moved while the image and mask stayed unchanged. (#4473, #4485)".

**[P3] On the first of those two cases, this PR changes which output is inconsistent.** Setup: `RandomCrop((20, 26), padding=(1, 2), p=0.5)` on a 24x32 batch, a hand-set `batch_prob=[0, 1, 0, 1]`, `resample="nearest"`, and one bright pixel per row as a landmark. The fallback returns a transformed image for the skipped rows 0 and 2 too. On main row 0's keypoint followed its pixel (pixel `(13, 12)`, keypoint `(13, 12)`) in both modes, and so did row 2 in `resample` mode, but only because of the aliasing this PR removes. In `slice` mode row 2 was already inconsistent on main (pixel `(19, 7)`, keypoint `(19, 13)`). On this head the skipped rows get their input keypoints back in both modes, e.g. `(12, 10)` against a pixel at `(13, 12)`. The all-skipped batch, the case `p` can actually produce, is now consistent in both modes (keypoint == pixel on every row, where main was off by `(1, 2)`). No change is needed here: the real defect is the image fallback, and the follow-up issue should cover this case so a later fix handles images and labels together.

**Merge state:** GitHub reports a conflict only in `CHANGELOG.md`. `crop.py` has not changed on main since your base, and the source diff applies cleanly to main `dc0af1a4`.

**Interaction with #4465:** there is no code conflict. With both PRs applied on main, `tests/augmentation/container/test_audit.py` passes with and without this fix (193 passed each). Its crop test computes the expected error from the returned outputs, so it pins no `sqrt(5)`. On the natural reproduction from #4465 (seed 1, `RandomCrop((4, 8), padding=(1, 2), p=0.5)`), `report.spatial[0].roundtrip_max` drops from `2.236` to `0.0`. The only thing that goes stale is the #4465 PR description, which calls that residual intentional.

**CI:** `tests (macos-latest, float32, 3.11, 2.5.1)` failed only on `tests/contrib/test_super_resolution.py::TestSuperResolutionBuilders::test_to_onnx_exports[RRDBNetBuilder]`, a `Timeout (>120.0s)` (1 failed, 10072 passed). That is a known flake on this leg and has nothing to do with this change. `collector` is red only because of that leg.

**Verified on this head** (torch 2.14.0, Apple Silicon, unless noted):
- **New test file:** 92 passed on CPU float32/float64. Copied onto main `dc0af1a4`, 52 failed and 40 passed. The 40 are the controls: the applied path, the `p=1` box/object cases, gradcheck, and padding with no left/top offset.
- **Other runs of the new file:**
  - `KORNIA_TEST_OPTIMIZER=inductor` dynamo test: 2 passed here, 2 failed on main.
  - Half precision: 47 passed each for float16 and bfloat16 with the `cpu-float16` / `cpu-bfloat16` profiles.
  - MPS float32: 45 passed, 2 skipped (float64 gradcheck).
  - torch 2.5.1: 92 passed.
- **Each clone is tested on its own:** reverting only the keypoint clone fails 20 tests, and reverting only the box clone fails 26 (float32).
- **Existing suites on main plus this fix:** `tests/augmentation/container/`, `tests/augmentation/test_base.py`, `test_augmentation.py::TestRandomCrop`, `tests/geometry/test_boxes.py` and `test_keypoints.py` give 1991 passed, 469 skipped, 20 xfailed, 0 xpassed on CPU float32/float64. None of the #4452 pins change. Crop-related container and `TestRandomCrop` tests on MPS and in half precision also pass.
- **Bitwise sweep against main:** B=3, `image+mask+keypoints+bbox_xyxy`, 288 configs. The grid covered:
  - `cropping_mode`: slice and resample;
  - `p`: 0, 0.5 and 1;
  - `padding`: None, `1`, `(1, 2)`, `(1, 2, 3, 0)`, `pad_if_needed`, and a same-size crop with `(1, 2)`;
  - `same_on_batch`: on and off;
  - 4 seeds each.

  Image, mask, the next RNG draw and the input tensors are bitwise identical between head and main. Keypoints, boxes and their `inverse` differ only on the 120 skipped calls, where main shifted them and this head returns the input bitwise. With `p=1` the labels are bitwise identical.
- **Other checks:**
  - Gradients: gradcheck passes for `p=0` and `p=1`.
  - Cost: the extra clone is not measurable (16×3×256² images, 2000 keypoints, 200 boxes, single thread, both trees 4.8–5.2 ms min). As a side effect, a direct `crop.transform_keypoints(k, ...)` no longer mutates the caller's `Keypoints`; on main it did.
  - `RandomCrop3D` has no keypoint/box handlers, so this bug cannot occur there.
  - `grep -rnE "#4473|issues/4473" kornia/ tests/` has no hits on main or on this head.
  - `ruff 0.16.6` check and format are clean, and the `crop.py` doctest passes.

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (3a2a8095) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4485.fixed.md`.

ducha-aiki (approved):
Review of `3a2a809505a35175444bdc92748a1dd48c0ba40e`.

Re-reviewed current head `3a2a8095`. The previous issue-linking and release-note requests are resolved: the PR closes #4473 and `changelog.d/4485.fixed.md` states the old and new behavior. Cloning before the in-place `Keypoints.pad` and `Boxes.pad` calls preserves the skipped branch without changing those public in-place methods. The tests cover skipped/applied paths, replay, inverse, boxes, keypoints, gradients, Dynamo, and half precision. I found no actionable defect. Approve; merge after the refreshed required checks finish green.

Validation limits: this follow-up used source/delta inspection and retained prior numerical review evidence; the author’s new test counts and hosted checks were inspected, not rerun locally during this sweep.

PR #4486, opened 2026-09-13
fix(augmentation): pass RandomBoxBlur's flag to box_blur by keyword and document it

### LINKED ISSUE #4433: W1b: `RandomBoxBlur`'s `normalized` flag is passed into `box_blur`'s `separable` argument
`RandomBoxBlur` takes a `normalized` constructor argument, documented at `kornia/augmentation/_2d/intensity/box_blur.py:35` as "if True, L1 norm of the kernel is set to 1". It stores that value in `flags["normalized"]` and forwards it positionally at `box_blur.py:72` as the fourth argument of `kornia.filters.box_blur` — which is `separable`, not `normalized`. `box_blur` has no `normalized` parameter at all.

So the flag controls whether the blur is computed as two 1D passes or one 2D pass. It does not control normalization: the kernel is L1-normalized in both settings, and the `normalized=False` output at an interior pixel is exactly the mean of the input window, not its sum. The only difference between the two settings is the float rounding of the separable versus the non-separable path.

A caller who sets `normalized=False` expecting unnormalized box sums gets means, with no error, and a caller who wanted the separable path has no documented way to ask for it.

## Repro

```python
from kornia.filters import box_blur

print("kornia.filters.box_blur signature:", inspect.signature(box_blur))
print("'normalized' is a parameter of box_blur:", "normalized" in inspect.signature(box_blur).parameters)

torch.manual_seed(0)
x = torch.rand(2, 3, 6, 8)
t = K.RandomBoxBlur((3, 3), normalized=True, p=1.0)(x.clone())
f = K.RandomBoxBlur((3, 3), normalized=False, p=1.0)(x.clone())
print("max|class(normalized=True) - class(normalized=False)|:", (t - f).abs().max().item())
print("max|class(normalized=True)  - box_blur(..., separable=True) |:",
      (t - box_blur(x, (3, 3), "reflect", separable=True)).abs().max().item())
print("max|class(normalized=False) - box_blur(..., separable=False)|:",
      (f - box_blur(x, (3, 3), "reflect", separable=False)).abs().max().item())
print("normalized=False output at (2,2):", round(f[0, 0, 2, 2].item(), 6),
      "| mean of the 3x3 input window:", round(x[0, 0, 1:4, 1:4].mean().item(), 6),
      "| sum of that window:", round(x[0, 0, 1:4, 1:4].sum().item(), 6))
```

## Observed

```
kornia.filters.box_blur signature: (input: 'torch.Tensor', kernel_size: 'tuple[int, int] | int', border_type: 'str' = 'reflect', separable: 'bool' = False) -> 'torch.Tensor'
'normalized' is a parameter of box_blur: False
max|class(normalized=True) - class(normalized=False)|: 1.1920928955078125e-07
max|class(normalized=True)  - box_blur(..., separable=True) |: 0.0
max|class(normalized=False) - box_blur(..., separable=False)|: 0.0
normalized=False output at (2,2): 0.463115 | mean of the 3x3 input window: 0.463115 | sum of that window: 4.168036
```

## Expected

Two coherent outcomes:

1. **Make the flag real.** `kornia.filters.box_blur` grows a `normalized` parameter, and `RandomBoxBlur` passes `normalized` to it by keyword. This changes the output of every current `normalized=False` caller, who today gets means.
2. **Expose what the argument actually does.** `RandomBoxBlur` forwards by keyword to `separable=` and renames or replaces `normalized` accordingly, with the docstring sentence at `box_blur.py:35` corrected to say that the kernel is always L1-normalized.

The audit recommends (2), because it matches the executed behavior and keeps existing outputs byte-identical. Either way, the call at `box_blur.py:72` should pass its arguments by keyword: the whole defect is a positional forward into a parameter list that does not line up.

## Classification

**Semantic contract** — a public flag whose name and documentation describe normalization while it selects a separable convolution. Idiom: 3 (a guard or parameter whose effect does not match its documented contract).

## Disposition

`n/a — single symbol`. Batch 6 documents the current behavior as it is and pins it. Outcome (2) is a focused fix — one keyword at `box_blur.py:72` plus the `Args` entry — and welcome as a PR; outcome (1) changes existing output and belongs in the coordinated Repair & Deprecation window.

## Environment

dchaudhari7177:

`RandomBoxBlur` forwarded its `normalized` flag **positionally** as the fourth argument of `kornia.filters.box_blur`, which is `separable`. So the flag picks the separable or the single 2D pass, the kernel is L1-normalized either way, and `normalized=False` returns window means, not the sums its `Args` entry implies.

This is outcome (2) from the issue, scoped as its disposition describes: one keyword call and the `Args` entry.

- `apply_transform` calls `box_blur(input, kernel_size, border_type=..., separable=flags["normalized"])`, so the parameter lists can no longer silently misalign.
- The `Args` entry says what the flag does: it selects the implementation, and both settings return the window mean and differ only by float rounding.

Outputs are byte-identical. The argument keeps its name; renaming or deprecating it, or making normalization real (outcome 1), is left to the Repair & Deprecation window as the issue says.

Closes #4433 (tracking #4407).

## What did you check?

Two tests in `TestRandomBoxBlur`, which now uses `BaseTester`:

- `test_normalized_selects_the_separable_path[True|False]`: the class output equals `box_blur(..., separable=normalized)` exactly (`atol=0`).
- `test_output_is_the_window_mean_for_either_setting[True|False]`: the interior pixel equals the 3×3 window mean for both settings.

Both also pass on `main`, since the change keeps behaviour. They pin the contract the new docstring states, so reverting to a sum, or wiring the flag to anything but `separable`, fails them.

```text
$ python -m pytest tests/augmentation/test_augmentation.py -k TestRandomBoxBlur --device=cpu --dtype=float32,float64
10 passed
$ python -m pytest kornia/augmentation/_2d/intensity/box_blur.py --doctest-modules
1 passed
```

`ruff@0.16.6` check and format are clean. `grep -rnE "#4433|issues/4433" kornia/` is empty. No CUDA/MPS run.

## How did AI help?

I used Claude Code to draft the change, tests and description. I checked the `box_blur` signature, ran the tests and doctest above, and confirmed the outputs match `main`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -381,6 +381,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `RandomBoxBlur`'s `normalized` argument is documented as what it does. It was described as "if True, L1 norm
+  of the kernel is set to 1", but it was forwarded positionally into `kornia.filters.box_blur`'s `separable`
+  parameter: it chooses between the separable and the single 2D pass, and the kernel is L1-normalized either
+  way, so `normalized=False` returns window means, not sums. The call now passes `border_type` and
+  `separable` by keyword; outputs are unchanged. (#4433, #4486)
+
 * `RandomMosaic` now preserves `(H, W)` for non-square inputs when `output_size=None`; it previously
   returned `(W, H)`. `start_ratio_range` now scales x by width and y by height; it previously scaled
   x by height and y by width. (#4459)
--- a/kornia/augmentation/_2d/intensity/box_blur.py
+++ b/kornia/augmentation/_2d/intensity/box_blur.py
@@ -32,7 +32,9 @@ class RandomBoxBlur(IntensityAugmentationBase2D):
         kernel_size: the blurring kernel size.
         border_type: the padding mode to be applied before convolving.
           The expected modes are: ``constant``, ``reflect``, ``replicate`` or ``circular``.
-        normalized: if True, L1 norm of the kernel is set to 1.
+        normalized: selects the implementation of :func:`kornia.filters.box_blur`: ``True`` computes the blur
+          as two 1D passes (``separable=True``), ``False`` as one 2D pass. The kernel is L1-normalized either
+          way, so both return the mean of each window, never its sum, and differ only by float rounding.
         same_on_batch: apply the same transformation across the batch.
         p: probability of applying the transformation.
         keepdim: whether to keep the output shape the same as input (True) or broadcast it
@@ -69,4 +71,4 @@ class RandomBoxBlur(IntensityAugmentationBase2D):
     def apply_transform(
         self, input: Tensor, params: Dict[str, Tensor], flags: Dict[str, Any], transform: Optional[Tensor] = None
     ) -> Tensor:
-        return box_blur(input, flags["kernel_size"], flags["border_type"], flags["normalized"])
+        return box_blur(input, flags["kernel_size"], border_type=flags["border_type"], separable=flags["normalized"])
--- a/tests/augmentation/test_augmentation.py
+++ b/tests/augmentation/test_augmentation.py
@@ -4778,11 +4778,30 @@ class TestRandomElasticTransform(BaseTester):
                 assert labels_transformed[to_apply].ne(labels[to_apply]).any()
 
 
-class TestRandomBoxBlur:
+class TestRandomBoxBlur(BaseTester):
     def test_smoke(self, device, dtype):
         img = torch.rand(1, 1, 2, 2, device=device, dtype=dtype)
         aug = RandomBoxBlur(p=1.0)
         assert img.shape == aug(img).shape
+
+    @pytest.mark.parametrize("normalized", [True, False])
+    def test_normalized_selects_the_separable_path(self, normalized, device, dtype):
+        from kornia.filters import box_blur
+
+        torch.manual_seed(0)
+        img = torch.rand(2, 3, 6, 8, device=device, dtype=dtype)
+        out = RandomBoxBlur((3, 3), normalized=normalized, p=1.0)(img.clone())
+
+        expected = box_blur(img, (3, 3), border_type="reflect", separable=normalized)
+        torch.testing.assert_close(out, expected, rtol=0.0, atol=0.0)
+
+    @pytest.mark.parametrize("normalized", [True, False])
+    def test_output_is_the_window_mean_for_either_setting(self, normalized, device, dtype):
+        torch.manual_seed(0)
+        img = torch.rand(1, 1, 6, 8, device=device, dtype=dtype)
+        out = RandomBoxBlur((3, 3), normalized=normalized, p=1.0)(img.clone())
+
+        self.assert_close(out[0, 0, 2, 2], img[0, 0, 1:4, 1:4].mean())
 
 
 class TestPadTo(BaseTester):


ducha-aiki (dismissed):
First review of head `467c7686`. Thanks, this is a clean, correctly scoped fix for #4433. Approving. The only thing needed before merge is a merge of `main`, and the conflict is in `CHANGELOG.md` only.

What I checked:

- **Production change.** `box_blur.py:74` now passes `border_type=` and `separable=` by keyword, so the call no longer relies on the two parameter lists lining up. Outputs are unchanged. The public `normalized` keyword keeps its name and its default, so nothing breaks and the `Bug fixes` entry is the right place (it cites #4486). The name is still misleading, but renaming it or adding a `separable` alias belongs in the Repair & Deprecation window, as the issue's disposition says. Closing #4433 with this scope matches that disposition.
- **The docstring is accurate.** I compared `box_blur(separable=True)` with `separable=False` over the `constant`/`reflect`/`replicate`/`circular` borders and kernels `(3,3)`, `(3,5)`, `(5,3)`, `(4,4)`, `(7,7)` on 2x3x16x19 inputs. Across all of them the two paths never match bitwise, but they differ only by rounding. The largest gap on CPU is 4.9e-4 in float16, 3.9e-3 in bfloat16, 3.3e-7 in float32 and 3.3e-16 in float64. On MPS it is 4.9e-4 / 3.9e-3 / 2.4e-7 (float16/bfloat16/float32). That is about one ULP at the half-precision output scale. A `(1,7)` kernel is bitwise identical. With integer-valued float64 input the paths still differ, by up to 2.1e-14, so they are not mathematically distinct operations. In practice the flag is a speed switch. On CPU float32 1x3x256x256, separable vs 2D takes 2.0 vs 2.3 ms at 3x3, 5.2 vs 32.6 ms at 15x15 and 15.3 vs 171 ms at 31x31.
- **The tests pin the wiring.** Both tests pass on `main` too, which is expected for a behaviour-preserving change. So I checked them by mutating the forwarding line on this head and running `TestRandomBoxBlur` on CPU float16/bfloat16/float32/float64. Every mutation fails the tests:
  - swap to `separable=not normalized`: 8 failed
  - drop the flag, which always takes the 2D path: 4 failed
  - always separable: 4 failed
  - return window sums when `normalized=False`: 8 failed
  - ignore `border_type`: 8 failed

  I also checked that the separable and 2D outputs differ bitwise for all 500 seeds tried on the test's 2x3x6x8 input in every dtype, so the `atol=0` comparison cannot pass by accident. The window-mean test is a useful independent value pin that does not use `box_blur` as its own oracle.
- **No leftover references.** `grep -rnE "#4433|issues/4433" kornia/ tests/` is empty on this head and on `main`. #4452 added no Convention block or wart pin for this issue, and no other `box_blur` caller forwards positionally into `separable` (the `guided.py` calls already use `separable=`).

**[P3] Optional polish.**
- `test_augmentation.py:4796`: the class now derives from `BaseTester`, so `self.assert_close(out, expected, rtol=0.0, atol=0.0)` fits the file's convention. It honours explicit zero tolerances.
- `test_augmentation.py:4789`: the `from kornia.filters import box_blur` inside the test can move to the module imports.
- The `Args` entry could add one clause saying that `True` (the default) is much faster for large kernels. That is the practical reason to choose between the settings.

**Verified on this head:** `TestRandomBoxBlur` passes: 10 on CPU float32/float64, 5 on CPU float16 and 5 on CPU bfloat16 (each with its known-failure profile), and 10 on MPS float32/float16. The module doctest passes (1), and ruff 0.16.6 check/format is clean on both changed files. The new tests also pass when copied onto `main` (20 across the four CPU dtypes). You can see the mutation and path-equivalence probes above. CI shows 40/40 green.

ducha-aiki (dismissed):
Re-approving on `6b248dfd`. To clear the `CHANGELOG.md`-only conflict (`main` added entries at the top of the same `### Bug fixes` list), I pushed `6b248dfd`, a keep-both merge of `origin/main` (`dc0af1a4`) into the head `467c7686` I reviewed above. It is a pure merge: every path except `CHANGELOG.md` is tree-identical to `git merge-tree`, and the non-blank lines it adds to `CHANGELOG.md` are exactly this branch's own. The review above stands unchanged. If you push again, please pull this merge first.

PR #4487, opened 2026-09-13
fix(geometry): let transform_points take an empty transform batch with points

### LINKED ISSUE #4466: transform_points raises ZeroDivisionError on an empty transform batch with a non-empty point axis
## Describe the bug

`transform_points` raises `ZeroDivisionError` when the transform batch is empty and the points carry a non-empty point axis. #4386 made empty camera batches work, and the early return in `transform_points` covers an empty **point set** (`points.shape[-2] == 0`), but an empty **batch** with `N > 0` points still reaches the `repeat_interleave` line, where `int(points_1.shape[0] // trans_01.shape[0])` divides `0 // 0`.

`PinholeCamera.project` inherits it: a `B=0` camera projects `(0, 3)` points fine but crashes on `(0, 1, 3)`.

## Reproduction steps

```python
import torch
from kornia.geometry.linalg import transform_points

trans = torch.eye(4).expand(0, 4, 4)
transform_points(trans, torch.zeros(0, 3))      # (0, 3), fine: early return on shape[-2] == 0
transform_points(trans, torch.zeros(0, 5, 3))   # ZeroDivisionError: integer division or modulo by zero

cam = PinholeCamera(trans, trans, torch.zeros(0), torch.zeros(0))
cam.project(torch.zeros(0, 3))                  # (0, 2)
cam.project(torch.zeros(0, 1, 3))               # ZeroDivisionError
```

Observed on `main` at `8fb9ad12`, torch 2.14, CPU. The crash is in `kornia/geometry/linalg.py` at the `torch.repeat_interleave(..., repeats=int(points_1.shape[0] // trans_01.shape[0]))` line.

## Expected behavior

An empty batch is a valid input, as #4386 established for the camera paths: the result should be an empty tensor of shape `(0, 5, 3)` / `(0, 1, 2)`, the same as when the transform has `B=1` and the points have `B=0` (which already returns `(0, 5, 3)`).

## Notes

Found while re-reviewing #4450, which is not the cause (identical on base and on that PR's head). Fix sketch: extend the early return to `points_1.shape[0] == 0` when the leading dims match, or guard the `repeats` computation so `0 // 0` never happens.

dchaudhari7177:

`transform_points` raised `ZeroDivisionError` for an empty transform batch whose points carry a non-empty point axis, for example `(0, 4, 4)` transforms with `(0, 5, 3)` points. Validation accepts the pair, and the early return only covers an empty point *set* (`N == 0`), so the call reached

```python
torch.repeat_interleave(trans_01, repeats=int(points_1.shape[0] // trans_01.shape[0]), dim=0)  # 0 // 0
```

The division is now skipped when the transform batch is empty (`repeats = 0`). Validation already guarantees the point batch is empty too. The rest of the function runs unchanged, so the result is `(0, 5, 3)` in the caller's dtype, the same shape a `B=1` transform already returned for those points. `PinholeCamera.project` inherited the crash for an empty camera batch on `(0, N, 3)` points and now returns `(0, N, 2)`.

I kept the division on the normal path rather than adding a second early return, so the empty-batch output still goes through `bmm` and the dtype restoration like any other call.

Closes #4466.

## What did you check?

- `tests/geometry/test_linalg.py::TestTransformPoints::test_transform_points_empty_batch_with_points`, parametrized over `D ∈ {2, 3}` and points `(0, 1, D)` / `(0, 5, D)`: shape, dtype, and parity with the `B=1` transform.
- `tests/geometry/camera/test_pinhole.py::...::test_project_an_empty_batch_with_a_point_axis_4466`: `(0, 3)` → `(0, 2)` and `(0, 1, 3)` → `(0, 1, 2)`.

```text
$ python -m pytest tests/geometry/test_linalg.py tests/geometry/camera/test_pinhole.py -k "empty or 4466" --device=cpu --dtype=float32
main: 5 failed, 3 passed      # the 4 new linalg cases + the pinhole case, all ZeroDivisionError
head: 8 passed
$ python -m pytest tests/geometry/test_linalg.py tests/geometry/camera/test_pinhole.py --device=cpu --dtype=float32,float64
347 passed, 1 skipped
```

`ruff@0.16.6` check and format are clean. No CUDA/MPS/ONNX run.

## How did AI help?

I used Claude Code to locate the division, draft the guard and tests, and write this description. I reproduced the crash on `main` with the issue's snippet and ran the tests above on both trees.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -381,6 +381,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `transform_points` returns an empty result for an empty transform batch whose points carry a non-empty
+  point axis, such as `(0, 4, 4)` transforms with `(0, 5, 3)` points. It used to raise `ZeroDivisionError`
+  expanding the transforms (`0 // 0`), and `PinholeCamera.project` inherited the crash for an empty camera
+  batch on `(0, N, 3)` points. (#4466, #4487)
+
 * `RandomMosaic` now preserves `(H, W)` for non-square inputs when `output_size=None`; it previously
   returned `(W, H)`. `start_ratio_range` now scales x by width and y by height; it previously scaled
   x by height and y by width. (#4459)
--- a/kornia/geometry/linalg.py
+++ b/kornia/geometry/linalg.py
@@ -220,8 +220,11 @@ def transform_points(trans_01: torch.Tensor, points_1: torch.Tensor) -> torch.Te
     points_1 = points_1.reshape(-1, points_1.shape[-2], points_1.shape[-1])
     trans_01 = trans_01.reshape(-1, trans_01.shape[-2], trans_01.shape[-1])
     # We expand trans_01 to match the dimensions needed for bmm. repeats input division is cast
-    # to integer so onnx doesn't record the value as a tensor and get a device mismatch
-    trans_01 = torch.repeat_interleave(trans_01, repeats=int(points_1.shape[0] // trans_01.shape[0]), dim=0)
+    # to integer so onnx doesn't record the value as a tensor and get a device mismatch. An empty
+    # transform batch has nothing to repeat, and the validation above guarantees an empty point
+    # batch alongside it, so skip the division rather than computing ``0 // 0``.
+    repeats = points_1.shape[0] // trans_01.shape[0] if trans_01.shape[0] > 0 else 0
+    trans_01 = torch.repeat_interleave(trans_01, repeats=int(repeats), dim=0)
     # to homogeneous
     points_1_h = convert_points_to_homogeneous(points_1)  # BxNxD+1
     # `torch.bmm` requires both operands to share a scalar type. Callers may pass coordinate
--- a/tests/geometry/camera/test_pinhole.py
+++ b/tests/geometry/camera/test_pinhole.py
@@ -911,6 +911,16 @@ class TestPinholeCamera(BaseTester):
         )
         assert empty.shape == (0, 1, 2)
 
+    def test_project_an_empty_batch_with_a_point_axis_4466(self, device, dtype):
+        # Regression for kornia#4466: an empty camera batch projected (0, 3) points but raised
+        # ZeroDivisionError on (0, N, 3), inside transform_points.
+        trans = torch.eye(4, device=device, dtype=dtype).expand(0, 4, 4)
+        empty = torch.zeros(0, device=device, dtype=dtype)
+        camera = kornia.geometry.camera.PinholeCamera(trans, trans, empty, empty)
+
+        assert camera.project(torch.zeros(0, 3, device=device, dtype=dtype)).shape == (0, 2)
+        assert camera.project(torch.zeros(0, 1, 3, device=device, dtype=dtype)).shape == (0, 1, 2)
+
     @pytest.mark.parametrize("batch_sizes", [(1, 2, 1, 1), (0, 1, 0, 0)])
     def test_constructor_rejects_mismatched_batch_sizes_4281(self, batch_sizes, device, dtype):
         with pytest.raises(ValueError, match="Arguments shapes must match"):
--- a/tests/geometry/test_linalg.py
+++ b/tests/geometry/test_linalg.py
@@ -63,6 +63,21 @@ class TestTransformPoints(BaseTester):
         assert out.shape == points.shape
         self.assert_close(out, points)
 
+    @pytest.mark.parametrize("num_dims", [2, 3])
+    @pytest.mark.parametrize("points_shape", [(0, 1), (0, 5)])
+    def test_transform_points_empty_batch_with_points(self, num_dims, points_shape, device, dtype):
+        # An empty transform batch with a non-empty point axis divided 0 // 0 while expanding the
+        # transforms (kornia#4466); it must return the same empty shape as a B=1 transform does.
+        points = torch.zeros(*points_shape, num_dims, device=device, dtype=dtype)
+        empty_trans = torch.eye(num_dims + 1, device=device, dtype=dtype).expand(0, -1, -1)
+        single_trans = torch.eye(num_dims + 1, device=device, dtype=dtype)[None]
+
+        out = kgl.transform_points(empty_trans, points)
+
+        assert out.shape == points.shape
+        assert out.dtype == dtype
+        assert out.shape == kgl.transform_points(single_trans, points).shape
+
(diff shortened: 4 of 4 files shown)


ducha-aiki (dismissed):
First review of head `abcfdbc4`. Thanks, this is a clean, minimal fix: the guard only changes the `repeats` value when the transform batch is empty, so the non-empty path is unchanged, and the empty case still runs through `bmm` and the dtype restore. I found nothing blocking. Two optional notes are below.

**[P3] The new comment claims more than validation guarantees** (`kornia/geometry/linalg.py:223-225`). "The validation above guarantees an empty point batch alongside it" is true for 3-D inputs. It is not true for the `(*, B, N, D)` inputs the reshape accepts, because validation only compares `shape[0]`. `transform_points(torch.eye(4).expand(2, 0, 4, 4), torch.zeros(2, 3, 5, 3))` passes validation and flattens to 0 transforms for 6 point sets. On base that raised `ZeroDivisionError`. On this head it raises `RuntimeError: Expected size for first two dimensions of batch2 tensor to be: [6, 4] but got: [0, 4]` from `bmm`. The input is invalid either way, so behaviour is fine. Only the comment is too strong. Something like "for matching leading dimensions the point batch is empty too" would be accurate.

**[P3] The 4-D broadcast path the camera helpers use has no pin.** `pixel2cam` and `cam2pixel` call `transform_points(K[:, None], (B, H, W, 3))`. With `B=0` both raised `ZeroDivisionError` on base and now return `(0, H, W, 3)` / `(0, H, W, 2)`. The PR fixes them too, but the new tests only cover the 3-D `(0, N, D)` shape. You could add one `(0, 1, D+1, D+1)` / `(0, H, W, D)` case to `test_transform_points_empty_batch_with_points`, or a `pixel2cam` case. The CHANGELOG entry could also mention these helpers next to `PinholeCamera.project` (`unproject` is fixed the same way).

**CI:** the two red legs (macOS py3.11 and py3.12 with torch 2.5.1) each have exactly one failure: `tests/contrib/test_super_resolution.py::TestSuperResolutionBuilders::test_to_onnx_exports[RRDBNetBuilder]` hit `Timeout (>120.0s)`. That is the known macOS/2.5.1 flake (10030 passed otherwise), and `collector` is red only because those legs failed. The PR is `CONFLICTING` on `CHANGELOG.md` only (no other file overlaps with `main`), and a routine keep-both merge will clear it.

**Verified on this head**
- Shape matrix, head vs base, D=2 and D=3. Transforms `(0)` with points `(0,1,D)` / `(0,5,D)` now return `(0,N,D)` (base: `ZeroDivisionError`). Transforms `(0)` with `(0,0,D)`, `(1)` with `(0,5,D)`, and `(1)` with unbatched `(5,D)` behave the same on both. Transforms `(0)` with points `(1,5,D)` / `(2,5,D)`, `(2)` with `(0,5,D)`, and unbatched `(D+1,D+1)` with `(0,5,D)` still raise `ValueError` on both. 4-D inputs: `(0,2)`/`(0,2,5)`, `(2,0)`/`(2,0,5)` and `(0,1)`/`(0,3,5)` now return the input shape (base: `ZeroDivisionError`).
- dtype and device: mixed float64/float32, float32/float16 and float16/float64 transforms/points on an empty batch return the points' dtype. MPS float32 returns `(0,5,3)` on `mps:0`.
- Gradient: with `requires_grad` transforms `(0,4,4)` and points `(0,5,3)`, the output has a `grad_fn`, and `.sum().backward()` gives grads of shape `(0,4,4)` and `(0,5,3)`.
- `PinholeCamera` with `B=0`: `project` on `(0,1,3)` / `(0,5,3)` returns `(0,1,2)` / `(0,5,2)`, and `unproject` on `(0,1,2)` / `(0,5,2)` returns `(0,1,3)` / `(0,5,3)` (base: `ZeroDivisionError`).
- `torch.compile`: `dynamo.explain` shows 1 graph and 0 breaks. With `fullgraph=True`, over 5 non-empty `(B, N)` shapes, head and base compile the same number of frames and ops (5 frames / 95 ops static, 3 frames / 67 ops dynamic). The empty batch now compiles (base: `Unsupported: Observed exception`). Inductor `fullgraph=True` is fine on both paths.
- ONNX (TorchScript exporter, opset 17, dynamic B and N): the graph is node-for-node identical to base (101 nodes, same op histogram), and onnxruntime matches eager at `(2,5)` and `(3,7)`.
- The PR's new tests fail on base (10 failed, all `ZeroDivisionError`, cpu float32 and float64) and pass on head.
- `tests/geometry/test_linalg.py` and `tests/geometry/camera/test_pinhole.py`:
  - cpu float32 and float64: 347 passed, 1 skipped (same on torch 2.5.1)
  - cpu float16 with the known-failure profile: 176 passed, 3 skipped, 7 xfailed
  - cpu bfloat16 with the known-failure profile: 177 passed, 2 skipped, 7 xfailed
  - MPS float32: 166 passed, 20 skipped
  - `TestTransformPoints` with `KORNIA_TEST_OPTIMIZER=inductor`: 35 passed
- `tests/geometry/{camera,test_depth.py,epipolar,transform}` at cpu float32: 1189 passed on head, 1188 on base. The only difference is the new pinhole test, and both trees show the same 4 xpasses.
- `ruff@0.16.6` check and format are clean. No line is over 120 columns. The `#4466` grep hits only the two new tests and the CHANGELOG, with no stale prose in `kornia/`. `closingIssuesReferences` is only #4466, and no duplicate PR exists.

ducha-aiki (dismissed):
Re-approving on `ec5054cf`. To clear the `CHANGELOG.md`-only conflict (`main` added entries at the top of the same `### Bug fixes` list), I pushed `ec5054cf`, a keep-both merge of `origin/main` (`dc0af1a4`) into the head `abcfdbc4` I reviewed above. It is a pure merge: every path except `CHANGELOG.md` is tree-identical to `git merge-tree`, and the non-blank lines it adds to `CHANGELOG.md` are exactly this branch's own. The review above stands unchanged. If you push again, please pull this merge first.

ducha-aiki (dismissed):

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (9a87c38d) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4487.fixed.md`.

ducha-aiki (approved):
Re-approving on the changelog-fragment merge commit; the code diff against main is unchanged.

PR #4488, opened 2026-09-13
fix(contrib): let diamond_square generate sides below 3 px

### LINKED ISSUE #4472: diamond_square crashes for spatial sizes below 3 px with an unrelated error, and PatchSequential(RandomPlasma*, padding="valid") hits it
`kornia.contrib.diamond_square` fails for any spatial size below 3 px (main `5be4852b`, torch 2.14.0, CPU):

```python
from kornia.contrib import diamond_square
diamond_square((1, 1, 1, 1))   # ValueError: math domain error
diamond_square((1, 1, 2, 2))   # TypeError: rand(): argument 'size' failed to unpack the object at pos 3
diamond_square((1, 1, 2, 3))   # TypeError: rand(): argument 'size' failed to unpack the object at pos 3
diamond_square((1, 1, 3, 3))   # ok -> (1, 1, 3, 3)
```

Neither error names the size constraint. The `(1, 1)` case comes from `math.log2` on a non-positive intermediate; the `2 x k` cases come from a float landing in the `torch.rand` size tuple after the power-of-two step computation.

**Why it matters:** `PatchSequential(RandomPlasmaBrightness(...), padding="valid")` on a 9x10 image with a `(2, 2)` grid dies on exactly this call on `main` today (24 of 48 container mode combinations in the #4462 review), because the plasma generator is drawn at the patch size. #4462 sidesteps it by drawing the fractal at the full-image size the container currently passes and cropping; once #4460 passes the true patch shape to the children, patches under 3 px will hit it again.

**Expected:** either support sizes below 3 px (a 1x1 or 2x2 fractal is just the corner draws) or raise a `ValueError` that states the minimum size, before any arithmetic runs. Found during the review of #4462.

dchaudhari7177:

`kornia.contrib.diamond_square` failed for any spatial side below 3 px:

```text
(1, 1, 1, 1)  ValueError: math domain error                  # log2(1 - 1)
(1, 1, 2, 2)  TypeError: rand(): argument 'size' failed ...  # depth -1, seed size 3.0
```

The seed grid size came from `2 ** ceil(log2(side - 1)) + 1`. At 1 px that is `log2(0)`. At 2 px it gives a grid of 2, a recursion depth of `-1`, and a float seed size.

It now uses `max(side, 3)`, the smallest grid the algorithm builds (two corners and a midpoint). The existing `img[..., :width, :height]` slice then cuts it to the requested size, so a 1 or 2 px side is just its corner draws, as the issue suggested. For a side of 3 px or more, `max(side, 3) == side`, so the arithmetic, the random draws and the output are unchanged.

Closes #4472 (found in the #4462 review; relevant once #4460 passes the true patch shape).

## What did you check?

- `test_spatial_sizes_below_three` over `(1, 1)`, `(2, 2)`, `(2, 3)`, `(3, 2)`, `(1, 5)`, `(2, 17)`: shape and finiteness.
- `test_plasma_augmentation_on_a_two_pixel_image`: `RandomPlasmaBrightness(p=1.0)` on a `(1, 1, 2, 2)` image.
- **Unchanged outputs for sides ≥ 3 px:** with `torch.manual_seed(0)` and `(B, C) = (2, 3)`, `diamond_square` on `main` and on this branch are `torch.equal` for `(3, 3)`, `(3, 4)`, `(4, 3)`, `(5, 9)`, `(9, 5)`, `(16, 17)`, `(33, 20)` and `(64, 64)`.

```text
$ python -m pytest tests/contrib/test_diamond_square.py --device=cpu --dtype=float32
main: 7 failed, 2 passed
head: 9 passed            # 18 passed across float32 + float64
```

`ruff@0.16.6` check and format are clean. No CUDA/MPS run.

## How did AI help?

I used Claude Code to trace the arithmetic for 1 and 2 px sides, draft the change and tests, and write this description. I ran the tests above and the main-versus-branch output comparison myself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -381,6 +381,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `kornia.contrib.diamond_square` generates fractals with a spatial side below 3 px. A 1 px side raised
+  `ValueError: math domain error` and a 2 px side a `TypeError` from a float in the `torch.rand` size, so
+  `RandomPlasmaBrightness`, `RandomPlasmaContrast` and `RandomPlasmaShadow` failed on such images. A small
+  side is now drawn on the 3 px grid and sliced; outputs for sides of 3 px and more are bit-identical.
+  (#4472, #4488)
+
 * `RandomMosaic` now preserves `(H, W)` for non-square inputs when `output_size=None`; it previously
   returned `(W, H)`. `start_ratio_range` now scales x by width and y by height; it previously scaled
   x by height and y by width. (#4459)
--- a/kornia/contrib/diamond_square.py
+++ b/kornia/contrib/diamond_square.py
@@ -206,9 +206,11 @@ def diamond_square(
     for x in output_size[:-2]:
         num_samples *= x
 
-    # compute the image seed
-    p2_width: float = 2 ** math.ceil(math.log2(width - 1)) + 1
-    p2_height: float = 2 ** math.ceil(math.log2(height - 1)) + 1
+    # compute the image seed. The smallest grid the algorithm builds is 3 px per side (two corners
+    # and a midpoint), so a smaller side is generated at 3 px and sliced down below: a 1 or 2 px
+    # side is just its corner draws. Sizes from 3 px up are unaffected.
+    p2_width: int = 2 ** math.ceil(math.log2(max(width, 3) - 1)) + 1
+    p2_height: int = 2 ** math.ceil(math.log2(max(height, 3) - 1)) + 1
     recursion_depth: int = int(min(math.log2(p2_width - 1) - 1, math.log2(p2_height - 1) - 1))
     seed_width: int = (p2_width - 1) // 2**recursion_depth + 1
     seed_height: int = (p2_height - 1) // 2**recursion_depth + 1
--- a/tests/contrib/test_diamond_square.py
+++ b/tests/contrib/test_diamond_square.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 #
 
+import pytest
 import torch
 
 import kornia
@@ -46,3 +47,18 @@ class TestDiamondSquare(BaseTester):
         )
         self.assert_close(out.min(), expected_min)
         self.assert_close(out.max(), expected_max)
+
+    @pytest.mark.parametrize("spatial_size", [(1, 1), (2, 2), (2, 3), (3, 2), (1, 5), (2, 17)])
+    def test_spatial_sizes_below_three(self, spatial_size, device, dtype):
+        # Sides under 3 px used to fail with `math domain error` or a float in the `rand` size
+        # (kornia#4472).
+        torch.manual_seed(0)
+        output_size = (2, 3, *spatial_size)
+        out = kornia.contrib.diamond_square(output_size, device=device, dtype=dtype)
+        assert out.shape == output_size
+        assert torch.isfinite(out).all()
+
+    def test_plasma_augmentation_on_a_two_pixel_image(self, device, dtype):
+        img = torch.rand(1, 1, 2, 2, device=device, dtype=dtype)
+        out = kornia.augmentation.RandomPlasmaBrightness(p=1.0)(img)
+        assert out.shape == img.shape


ducha-aiki (dismissed):
First review of head `f67bd5a9`.

Thanks for this. The fix is small and correct. Clamping each side to 3 px before the power-of-two step makes a side below 3 px equal to the 3 px grid, sliced. For every size of 3 px or more the arithmetic is unchanged, so both the output and the RNG consumption match `main` bit for bit. I checked both claims by running them, see below. I found nothing blocking, only two optional notes.

**[P3] The new tests only check shape and finiteness, not the value the CHANGELOG describes.** `tests/contrib/test_diamond_square.py:51-58` would also pass if a small side returned a constant map. I tried that: I planted `img = torch.full_like(img, 0.5)` for any side below 3, and all 9 tests in the file still passed. The CHANGELOG says "drawn on the 3 px grid and sliced", and one assertion can pin that, including RNG consumption:

```python
h, w = spatial_size
torch.manual_seed(0)
small = kornia.contrib.diamond_square((2, 3, h, w), device=device, dtype=dtype)
torch.manual_seed(0)
ref = kornia.contrib.diamond_square((2, 3, max(h, 3), max(w, 3)), device=device, dtype=dtype)
self.assert_close(small, ref[..., :h, :w], rtol=0.0, atol=0.0)
```

With your `spatial_size` list, this passes on this head on CPU (float32/float64/float16/bfloat16, 24 cases) and on MPS (float32/float16, 12 cases). It fails in all 6 cases against the constant-map plant. You could also test the case that originally surfaced this: `PatchSequential(RandomPlasmaShadow(p=1.0), grid_size=(2, 2), padding="valid", patchwise_apply=False)` on a `5x5` image, which gives `2x2` patches. It crashes on `main` and works with this PR.

**[P3] The code comment overstates the corner case.** `kornia/contrib/diamond_square.py:209-211` says "a 1 or 2 px side is just its corner draws". That is only true for a `1x1` output. A 2 px side keeps the seed's middle row or column. A 1 px side along an axis of 3 px or more keeps the interpolated midpoints, e.g. columns 1 and 3 of a `1x5`. Over 2000 draws, a `1x5` map has mean 0.30 while its corner pixel has mean 0.17. Something like "a 1 or 2 px side is the leading rows/columns of that 3 px grid" would be accurate.

The PR merges cleanly into `main` except for the usual `CHANGELOG.md` conflict. Merged locally, the diff against `main` is exactly this PR's three files, and the entry lands under `## Unreleased` / `### Bug fixes`.

**Verified on this head** (and on a local merge with `main` `dc0af1a4` for the augmentation checks, torch 2.14.0 unless noted):
- The new tests fail on `main`: with the head test file copied onto `main`, 14 of 18 fail on CPU float32/float64 (`math domain error` for a 1 px side, the `rand()` float-size `TypeError` for 2 px), and all 18 pass on this head. `tests/contrib/test_diamond_square.py`: 9/9 passed each on CPU float16 and bfloat16 (known-failure profiles), 9/9 on MPS float32, and 27 passed on torch 2.5.1 CPU float32/float64 (together with `test_patch_sequential_plasma.py`).
- Sizes of 3 px or more are bit-identical to `main`. I compared SHA-1 of the output and of the CPU (and MPS) RNG state after the call: all 961 `(H, W)` pairs for 3..33 x 3..33 on CPU float32, plus a 9x8 odd/even non-square grid on CPU float64/float16/bfloat16 and MPS float32/float16. Each was run with default roughness, a per-sample roughness tensor, and `normalize_range=(0, 1)`, B=2, C=3. That is 3963 cases with 0 differences. Through the augmentations, `RandomPlasmaBrightness/Contrast/Shadow` on 3..13 x {3,4,5,8,9,13} gave identical output, stored `params["plasma"]` and RNG state for 198 cases, 0 differences.
- Small sizes: 1x5, 2x7, 5x1, 1x1, 2x2, 1x2, 2x1, 2x17, 17x2 and 1x33 all equal the 3 px grid sliced, with the same RNG state afterwards. B=4, C=3, scalar and per-sample roughness, on CPU float32/float64/float16/bfloat16 and MPS float32/float16. Shape, dtype and device are preserved and outputs are finite. The same holds for zero-size sides (0x5, 5x0, 0x0), which now return an empty tensor instead of raising. The output is not constant: over 2000 samples every size gave 2000 distinct maps, and the top-left pixel has the same distribution at every size (mean about 0.166 from 1x1 up to 33x33). With `normalize_range`, 1x5, 2x2 and 5x1 span [0, 1] in float32/float16/bfloat16. A 1x1 map normalizes to the lower bound, since min-max has nothing to stretch.
- Replay: all three `RandomPlasma*` classes on 1x1, 1x5, 2x7, 5x1, 2x2 and 3x3 images give a bitwise `params=` replay under a different seed, a fresh forward still draws different noise, and the stored plasma has the image size. On `main` every one of these below 3 px raises.
- `PatchSequential` with each plasma class, 4 `patchwise_apply`/`random_apply` combinations, `same`/`valid`, 8 image sizes and 3 grids gave 528 combinations. On `main`, all 348 whose patches have a side below 3 px crash, in both padding modes. With this PR, 0 of 528 fail. With `padding="valid"`, `5x5`/`(2, 2)`, `4x7`/`(2, 3)` and `3x3`/`(3, 3)` now run in float16/32/64.
- On the merged tree, `test_diamond_square.py`, `container/test_patch_sequential_plasma.py`, `container/test_patch_sequential.py`, `test_base.py` and `test_augmentation.py::TestRandomPlasma` give: 1012 passed on CPU float32/float64, 535 on float16, 539 on bfloat16 (known-failure profiles), and 526 on MPS float32 (base 519, +7 new). Inductor: 15 passed. Doctests for `diamond_square.py` and `plasma.py`: 3 passed.
- `grep -rnE "#4472|issues/4472" kornia/ tests/` finds only the new test comment, and nothing on `main` (no Convention note or wart pin to update). `closingIssuesReferences` is only #4472, which is fully covered. No competing PR for #4472. `ruff@0.16.6` check and format are clean on the changed files. CI 40/40 green, including typecheck.

ducha-aiki (dismissed):
Re-approving on `4016d7f6`. To clear the `CHANGELOG.md`-only conflict (`main` added entries at the top of the same `### Bug fixes` list), I pushed `4016d7f6`, a keep-both merge of `origin/main` (`dc0af1a4`) into the head `f67bd5a9` I reviewed above. It is a pure merge: every path except `CHANGELOG.md` is tree-identical to `git merge-tree`, and the non-blank lines it adds to `CHANGELOG.md` are exactly this branch's own. The review above stands unchanged. If you push again, please pull this merge first.

Issue #4430, opened 2026-09-11
W1b: 2D intensity augmentations split four ways on out-of-`[0, 1]` input; ten collapse to zeros


ducha-aiki:
Every 2D intensity augmentation in `kornia.augmentation` is written against the same `[0, 1]` input assumption, and none of them agrees on what happens when a caller leaves that range. Pushing one seeded base image, scaled into `[0, 2]` and shifted into `[-1, 0]`, through all of them splits the family four ways: one group clamps the result back into `[0, 1]` on both fixtures, `RandomPlanckianJitter` clamps the upper end only, a second group passes the out-of-range values straight through, and `RandomEqualize` raises a raw index error out of its 256-bin histogram.

The split by itself is a documentation problem: a caller feeding mean/std-normalized tensors through a pipeline has no way to predict which member will renormalize the data and which will leave it alone. What makes it more than that is the second result of the same run — on the `[-1, 0]` image, ten of these classes return an **all-zero tensor**, with no warning and no error. The image is destroyed and the pipeline continues. `RandomPosterize` is a third variation: it reaches `[0, 1]` through a `uint8` round-trip that floors and wraps rather than clamping, so a `[0, 2]` ramp comes back with four times the distinct levels its `bits` argument asks for.

None of this is stated on the classes. The same shape of problem in `kornia.color` was accepted as a defect in #3997, and #1163 is a user running into the range precondition of `equalize` back when it was still enforced.

## Repro

```python
import textwrap, torch, kornia.augmentation as K

N = dict(mean=torch.tensor([0.5]), std=torch.tensor([0.5]))
SPECS = [
    ("ColorJiggle", (0.2, 0.2, 0.2, 0.1), {}), ("ColorJitter", (0.2, 0.2, 0.2, 0.1), {}),
    ("Denormalize", (), N), ("Normalize", (), N), ("RandomAutoContrast", (), {}),
    ("RandomBoxBlur", (), {}), ("RandomBrightness", ((1.5, 1.5),), {}),
    ("RandomChannelDropout", (), {}), ("RandomChannelShuffle", (), {}),
    ("RandomContrast", ((1.5, 1.5),), {}), ("RandomEqualize", (), {}), ("RandomErasing", (), {}),
    ("RandomGamma", ((2.0, 2.0), (1.0, 1.0)), {}), ("RandomGaussianBlur", ((3, 3), (1.0, 1.0)), {}),
[...]

PR #4489, opened 2026-09-13
fix(enhance): name the [0, 1] range when equalize cannot index an input

### LINKED ISSUE #4431: W1b: `RandomEqualize` raises a raw index error on any input outside `[0, 1]`
`RandomEqualize` and the `kornia.enhance.equalize` it delegates to assume the input lies in `[0, 1]`. Neither public docstring says so, and nothing validates it. An image outside that range does not produce a kornia error; it produces a raw indexing failure out of the 256-bin histogram, whose message names an internal bin index and gives the caller nothing to act on. A negative input fails the same way with a negative index.

This used to be enforced. `_scale_channel` in `kornia/enhance/adjust.py` still carries the two `ValueError`s a caller would want — `Values in the input torch.Tensor must greater or equal to 0.0. Found …` — and that is the message the reporter of #1163 saw in 2021; the helper is now unreferenced except from a docstring. The vectorized replacement `_scale_channel_batched` dropped the check in #3842 to remove two `.item()` device syncs and make the op `torch.compile` fullgraph-safe. Two claims made in that change do not hold today. Its commit message says out-of-range input "is handled without the check" and its comment at `adjust.py:967` says such values are "clamped into the 256-bin range below"; the clamp at `adjust.py:972` applies to the histogram accumulation only, while the LUT lookup at `adjust.py:991` indexes with the unclamped `scaled.long()`, which is what raises. The same commit message says "the `[0, 1]` expectation is documented on `equalize`"; the range appears in the private helper's docstring, not in the public one.

Removing a device sync from the hot path is a good reason not to restore an `.item()`-based check. It is not a reason to leave the precondition undocumented and unvalidated: a tensor-level `KORNIA_CHECK`, or the clamp the comment already claims, both keep the graph intact.

## Repro

```python
from kornia.enhance import equalize

x = torch.linspace(0, 1, 64).reshape(1, 1, 8, 8)
for tag, img in (("[0, 1]", x), ("[0, 1.0001]", x * 1.0001), ("[0, 2]", x * 2.0), ("[-1, 0]", x - 1.0)):
    torch.manual_seed(0)
    try:
        y = K.RandomEqualize(p=1.0)(img.clone())
        print(f"RandomEqualize on {tag:<12} -> out min/max {y.min().item():.6g}/{y.max().item():.6g}, "
              f"{y.unique().numel()} distinct levels")
    except Exception as e:                                  # noqa: BLE001
        print(f"RandomEqualize on {tag:<12} -> {type(e).__name__}: {e}")
try:
    equalize(x * 2.0)
except Exception as e:                                      # noqa: BLE001
    print(f"kornia.enhance.equalize on [0, 2]  -> {type(e).__name__}: {e}")

for obj in (K.RandomEqualize, equalize):
    doc = inspect.getdoc(obj) or ""
    print(f"public docstring of {obj.__name__:<16} mentions a value range: "
          f"{any(s in doc for s in ('[0, 1]', '[0,1]', '0 and 1'))}")
```

## Observed

```
RandomEqualize on [0, 1]       -> out min/max 0/1, 64 distinct levels
RandomEqualize on [0, 1.0001]  -> out min/max 0/1.0001, 64 distinct levels
RandomEqualize on [0, 2]       -> RuntimeError: index 259 is out of bounds for dimension 1 with size 256
RandomEqualize on [-1, 0]      -> RuntimeError: index -255 is out of bounds for dimension 1 with size 256
kornia.enhance.equalize on [0, 2]  -> RuntimeError: index 259 is out of bounds for dimension 1 with size 256
public docstring of RandomEqualize   mentions a value range: False
public docstring of equalize         mentions a value range: False
```

## Expected

A kornia-level error naming the `[0, 1]` precondition, of the kind `_scale_channel` still raises, and a docstring on `RandomEqualize` and on `equalize` that states it. Either fix keeps `torch.compile` fullgraph working:

1. a tensor-level `KORNIA_CHECK` on the input range, with no `.item()` and therefore no device sync; or
2. make the code do what the comment at `adjust.py:967` already claims, by clamping the gather index at `adjust.py:991` the way the histogram index is clamped at `:972`. That silently accepts out-of-range input, so it has to be paired with a documented statement that values are clamped into `[0, 1]`.

The audit prefers (1): silent clamping is the behavior `RandomPosterize` already has, and it destroys data without telling the caller. Whichever is chosen, `kornia.enhance.equalize` needs the same treatment, since the class is a thin wrapper and the failure comes from the function. `RandomEqualize3D` has the same unvalidated precondition through `equalize3d` and is filed separately in this batch.

## Classification

**Broken output** — a precondition documented nowhere public, surfacing as a raw `RuntimeError` about an internal bin index on an input the caller has no reason to think is invalid. Idiom: 5 (a documented precondition never validated).

## Disposition

`n/a — single symbol`. Batch 6 documents the behavior above as it is and pins it. Restoring a validation message is a focused fix and welcome as a PR, provided it does not reintroduce the `.item()` sync that #3842 removed. The wider `[0, 1]` policy for the whole 2D intensity family is a separate decision, filed in this batch.

## Environment

### LINKED ISSUE #4432: W1b: `RandomEqualize3D`: undocumented `[0, 1]` precondition and a draw-dependent identity near 256 voxels
`RandomEqualize3D`'s docstring states neither a value range nor the resolution of the histogram it equalizes with, and two things follow from that.

First, a volume outside `[0, 1]` does not raise a kornia error. It raises a raw index error out of the 256-bin histogram, naming an internal bin index; a negative volume fails the same way with a negative index. This is the 3D counterpart of the `RandomEqualize` crash filed separately in this batch — both go through the same helper in `kornia/enhance/adjust.py`, whose range check was removed in #3842.

Second, on a valid `[0, 0.5]` volume the operator looks like an identity on a small enough input and is not one on a realistic volume. Equalization here uses a 256-bin histogram over the whole per-channel volume, and its LUT step is a truncating integer division by 255, so a volume with few enough voxels per channel yields a step of zero and comes back untouched, while a larger one is stretched across the full `[0, 1]` range. A caller checking the operator on a small test volume can therefore conclude it does nothing, then get a full-range rescale in production — which is how #780 was reported for the 2D op.

The boundary itself is not a clean threshold, and this is the part a docstring must not overstate. A five-seed sweep at one voxel count per size gives an exact identity at 255 voxels per channel on every seed, a result that depends on the draw at both 256 and 257, and no identity at all at 512. Five seeds support "255 and below is an identity" and "512 is reliably not"; they do not support any statement about which side of the line 256 and 257 fall on, because the draw decides.

## Repro

```python

# 1. is a [0, 0.5] volume left unchanged?
for shape in ((1, 1, 3, 5, 7), (1, 1, 5, 7, 9), (1, 3, 5, 7, 9)):
    torch.manual_seed(0)
    x = torch.rand(*shape) * 0.5
    y = K.RandomEqualize3D(p=1.0)(x.clone())
    print(f"{str(shape):<16} {shape[2] * shape[3] * shape[4]:>4} voxels/channel  "
          f"in=({x.min():.4f},{x.max():.4f}) out=({y.min():.4f},{y.max():.4f}) maxdiff={(y - x).abs().max():.6f}")

# 2. where does the identity stop?  five fresh seeds per size, one channel, n voxels
print()
for n in (255, 256, 257, 512):
    diffs = []
    for seed in range(5):
        torch.manual_seed(seed)
        x = torch.rand(1, 1, 1, 1, n) * 0.5
        diffs.append(round((K.RandomEqualize3D(p=1.0)(x.clone()) - x).abs().max().item(), 6))
    print(f"voxels={n:<4} maxdiffs(seed 0-4)={diffs} identity={sum(d == 0.0 for d in diffs)}/5")

# 3. the value-range precondition
print()
for tag, lo, hi in (("[0.0, 1.0]", 0.0, 1.0), ("[0.0, 1.5]", 0.0, 1.5), ("[-0.5, 0.5]", -0.5, 0.5)):
    torch.manual_seed(0)
    x = torch.rand(1, 1, 5, 7, 9) * (hi - lo) + lo
    try:
        y = K.RandomEqualize3D(p=1.0)(x)
        print(f"{tag:<12} -> out=({y.min():.4f},{y.max():.4f})")
    except Exception as e:                                      # noqa: BLE001
        print(f"{tag:<12} -> {type(e).__name__}: {e}")
```

## Observed

```
(1, 1, 3, 5, 7)   105 voxels/channel  in=(0.0006,0.4985) out=(0.0006,0.4985) maxdiff=0.000000
(1, 1, 5, 7, 9)   315 voxels/channel  in=(0.0006,0.4985) out=(0.0000,1.0000) maxdiff=0.606997
(1, 3, 5, 7, 9)   315 voxels/channel  in=(0.0000,0.4999) out=(0.0000,1.0000) maxdiff=0.606997

voxels=255  maxdiffs(seed 0-4)=[0.0, 0.0, 0.0, 0.0, 0.0] identity=5/5
voxels=256  maxdiffs(seed 0-4)=[0.0, 0.0, 0.501837, 0.0, 0.0] identity=4/5
voxels=257  maxdiffs(seed 0-4)=[0.50269, 0.0, 0.501938, 0.0, 0.500997] identity=2/5
voxels=512  maxdiffs(seed 0-4)=[0.743052, 0.744825, 0.759465, 0.750291, 0.735924] identity=0/5

[0.0, 1.0]   -> out=(0.0000,1.0000)
[0.0, 1.5]   -> RuntimeError: index 293 is out of bounds for dimension 1 with size 256
[-0.5, 0.5]  -> RuntimeError: index -104 is out of bounds for dimension 1 with size 256
```

## Expected

The docstring states the `[0, 1]` input precondition, says that equalization uses a 256-bin histogram over the whole per-channel volume, and says that a volume with no more than 255 voxels per channel is returned unchanged — without promising a particular behavior for the sizes just above that, which the draw decides. The precondition should also be validated, so that an out-of-range volume gets a kornia-level message instead of a bin index; the same constraint as in the 2D case applies, namely that the check must not reintroduce the `.item()` device sync #3842 removed.

Whether the truncating step is itself the right rule is a separate question from documenting it. Making small volumes equalize like large ones would change the output of every current caller with a small volume, so it is not part of this issue.

## Classification

**Semantic contract** — an unvalidated `[0, 1]` precondition surfacing as a raw index error, plus a size-dependent identity regime that no docstring mentions. Idiom: 5 (a documented precondition never validated).

## Disposition

`n/a — single symbol`. Batch 6 documents the behavior above as it is and pins it. Stating the precondition and the histogram resolution in the docstring is a focused fix and welcome as a PR; validating the range is the same fix as for `RandomEqualize` and should land with it.

## Environment

dchaudhari7177:

`kornia.enhance.equalize` / `equalize3d`, and the `RandomEqualize` / `RandomEqualize3D` wrappers, failed on out-of-range input with a raw `RuntimeError: index 259 is out of bounds for dimension 1 with size 256`. `_scale_channel_batched` clamps the histogram index, but gathers the LUT with the **unclamped** `scaled.long()`. The comment claiming out-of-range values are "clamped into the 256-bin range below" was wrong.

**The check.** It runs right after scaling, through the existing `_assert_async_value_check` (`torch._assert_async`), so there is no `.item()` and no device sync, as #3842 requires:

```python
_assert_async_value_check(
    ((scaled > -1.0) & (scaled < 256.0)).all(),
    "equalize expects input values in [0, 1]. Scale the image into that range first, for example image / 255.0 for 8-bit data.",
)
```

It validates exactly the domain the gather can index (`scaled.long()` in `[0, 255]`), not a strict `[0, 1]`. So every input that equalized before still does, including the issue's `x * 1.0001` case. Only inputs that already crashed change, and they now get a message naming the range. A strict `[0, 1]` check would be the wider policy decision #4430 reserves for the window. The misleading comment is corrected.

**Docstrings (#4432 and #4431).**
- `equalize`: the `[0, 1]` precondition, the 256-bin histogram, the error, and that the check is skipped on MPS (as for `adjust_gamma`).
- `equalize3d`: the same, plus one histogram over the whole per-channel volume, and that ≤ 255 voxels per channel comes back unchanged, without promising anything just above 255, which the values decide.
- `RandomEqualize` / `RandomEqualize3D`: the precondition and the error, pointing at the functions.

Closes #4431. Closes #4432. Tracking #4407.

## What did you check?

New tests in `tests/enhance/test_adjust.py`:
- `TestEqualize::test_out_of_range_input_names_the_range` (`x*2`, `x-1`), CPU-only like the other `_assert_async` tests.
- `TestEqualize::test_input_the_lookup_can_index_is_still_accepted` (`x*1.0001`).
- `TestEqualize::test_dynamo_fullgraph`: `torch.compile(equalize, fullgraph=True, backend="eager")` matches eager.
- `TestEqualize3D::test_out_of_range_input_names_the_range` (`[0, 1.5]`, `[-0.5, 0.5]`).
- `TestEqualize3D::test_at_most_255_voxels_per_channel_is_unchanged`.

```text
$ python -m pytest tests/enhance/test_adjust.py -k TestEqualize --device=cpu --dtype=float32
main: 4 failed, 16 passed, 2 skipped     # the four out-of-range cases (raw index error)
head: 20 passed, 2 skipped               # 38 passed, 4 skipped across float32 + float64
$ python -m pytest kornia/enhance/adjust.py kornia/augmentation/_2d/intensity/equalize.py kornia/augmentation/_3d/intensity/equalize.py --doctest-modules
28 passed
$ python -m pytest tests/augmentation -k Equalize --device=cpu --dtype=float32
20 passed, 4 skipped, 1 xfailed, 1 xpassed   # identical on main (existing Windows print-precision markers)
```

`ruff@0.16.6` check and format are clean. No CUDA/MPS run: on CUDA the assert is asynchronous, and on MPS it is skipped by design.

## How did AI help?

I used Claude Code to trace the gather, choose the check domain, draft the change, docstrings and tests, and write this description. I reproduced the issue's crashes on `main` and ran the tests and doctests above on both trees.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -381,6 +381,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `kornia.enhance.equalize`, `equalize3d`, `RandomEqualize` and `RandomEqualize3D` raise a `RuntimeError` naming
+  the `[0, 1]` input range for values the 256-bin lookup cannot index, instead of a raw
+  `index 259 is out of bounds for dimension 1 with size 256` from the gather. The check uses
+  `torch._assert_async`, so it adds no device sync and `torch.compile` fullgraph still works, and inputs that
+  equalized before (including values a hair above 1) are unchanged. The docstrings now state the range, the
+  256-bin histogram, and that a 3D volume of at most 255 voxels per channel comes back unchanged.
+  (#4431, #4432, #4489)
+
 * `RandomMosaic` now preserves `(H, W)` for non-square inputs when `output_size=None`; it previously
   returned `(W, H)`. `start_ratio_range` now scales x by width and y by height; it previously scaled
   x by height and y by width. (#4459)
--- a/kornia/augmentation/_2d/intensity/equalize.py
+++ b/kornia/augmentation/_2d/intensity/equalize.py
@@ -39,7 +39,8 @@ class RandomEqualize(IntensityAugmentationBase2D):
         - Output: :math:`(B, C, H, W)`
 
     .. note::
-        This function internally uses :func:`kornia.enhance.equalize`.
+        This function internally uses :func:`kornia.enhance.equalize`, which expects the input in
+        :math:`[0, 1]` and raises a ``RuntimeError`` naming that range for values it cannot equalize.
 
     Examples:
         >>> rng = torch.manual_seed(0)
--- a/kornia/augmentation/_3d/intensity/equalize.py
+++ b/kornia/augmentation/_3d/intensity/equalize.py
@@ -37,7 +37,10 @@ class RandomEqualize3D(IntensityAugmentationBase3D):
         - Output: :math:`(B, C, D, H, W)`
 
     Note:
-        Input tensor must be float and normalized into [0, 1] for the best differentiability support.
+        Input tensor must be float and normalized into [0, 1]: values outside it that the 256-bin lookup
+        cannot index raise a ``RuntimeError`` naming the range. This uses :func:`kornia.enhance.equalize3d`,
+        which equalizes each channel's whole volume from one 256-bin histogram, so a volume with no more
+        than 255 voxels per channel is returned unchanged.
         Additionally, this function accepts another transformation tensor (:math:`(B, 4, 4)`), then the
         applied transformation will be merged int to the input transformation tensor and returned.
 
--- a/kornia/enhance/adjust.py
+++ b/kornia/enhance/adjust.py
@@ -964,9 +964,15 @@ def _scale_channel_batched(input: torch.Tensor) -> torch.Tensor:
     n = shape[0] * shape[1]
     scaled = input.reshape(n, -1) * 255.0  # (N, P)
 
-    # Input is expected in [0, 1] (see the docstring). Out-of-range values are clamped into the
-    # 256-bin range below rather than raising: the previous ``.item()`` range check forced two
-    # device syncs and broke ``torch.compile`` fullgraph for no correctness benefit on valid input.
+    # Input is expected in [0, 1]. The histogram index below is clamped, but the LUT lookup indexes
+    # with the unclamped ``scaled.long()``, which is out of bounds unless ``scaled`` is in (-1, 256).
+    # Check that domain without ``.item()``, so there is no device sync and fullgraph still compiles;
+    # inputs the lookup can index keep working exactly as before.
+    _assert_async_value_check(
+        ((scaled > -1.0) & (scaled < 256.0)).all(),
+        "equalize expects input values in [0, 1]. Scale the image into that range first, "
+        "for example image / 255.0 for 8-bit data.",
+    )
 
     # Per-plane 256-bin histogram matching ``torch.histc(x, 256, 0, 255)`` bin placement.
     bins = torch.clamp((scaled * (256.0 / 255.0)).floor().long(), 0, 255)
@@ -1051,6 +1057,12 @@ def equalize(input: torch.Tensor) -> torch.Tensor:
     Returns:
         Equalized image torch.Tensor with shape :math:`(*, C, H, W)`.
 
+    .. note::
+       The input is expected in :math:`[0, 1]`, and each channel is equalized from a 256-bin histogram.
+       Values the 256-bin lookup cannot index (outside roughly :math:`[0, 1]`) raise a ``RuntimeError``
+       naming the range. The check runs on CPU and CUDA (via ``torch._assert_async``); on MPS it is
+       skipped, as for :func:`adjust_gamma`.
+
     Example:
         >>> x = torch.rand(1, 2, 3, 3)
         >>> equalize(x).shape
@@ -1073,6 +1085,14 @@ def equalize3d(input: torch.Tensor) -> torch.Tensor:
 
     Returns:
         Equalized volume with shape :math:`(B, C, D, H, W)`.
+
+    .. note::
+       The input is expected in :math:`[0, 1]`, and each channel's whole :math:`(D, H, W)` volume is
+       equalized from one 256-bin histogram. The lookup step is an integer division by 255, so a volume
+       with no more than 255 voxels per channel is returned unchanged; just above that, whether it changes
+       depends on the values. Values the 256-bin lookup cannot index (outside roughly :math:`[0, 1]`)
+       raise a ``RuntimeError`` naming the range. The check runs on CPU and CUDA (via
+       ``torch._assert_async``); on MPS it is skipped.
 
     """
     # Scales each channel independently (each (D, H, W) volume), batched over (B, C).
--- a/tests/enhance/test_adjust.py
+++ b/tests/enhance/test_adjust.py
@@ -885,6 +885,28 @@ class TestEqualize(BaseTester):
         inputs = torch.ones(bs, channels, height, width, device=device, dtype=torch.float64)
         self.gradcheck(kornia.enhance.equalize, (inputs,), fast_mode=False)
 
+    @pytest.mark.parametrize("scale, shift", [(2.0, 0.0), (1.0, -1.0)])
+    def test_out_of_range_input_names_the_range(self, scale, shift, device, dtype):
+        # kornia#4431: an input the 256-bin lookup cannot index used to fail with a raw
+        # "index 259 is out of bounds" from the gather.
+        if device.type != "cpu":
+            pytest.skip("value asserts are synchronous only on CPU (async on CUDA, skipped on MPS)")
+        x = torch.linspace(0, 1, 64, device=device, dtype=dtype).reshape(1, 1, 8, 8) * scale + shift
+        with pytest.raises(RuntimeError, match=r"expects input values in \[0, 1\]"):
+            kornia.enhance.equalize(x)
+
+    def test_input_the_lookup_can_index_is_still_accepted(self, device, dtype):
+        # The check covers exactly the values that crashed, so a hair above 1 keeps working.
+        x = torch.linspace(0, 1, 64, device=device, dtype=dtype).reshape(1, 1, 8, 8) * 1.0001
+        assert kornia.enhance.equalize(x).shape == x.shape
+
+    def test_dynamo_fullgraph(self, device, dtype):
+        # The range check must not reintroduce a graph break (#3842 removed an ``.item()`` check).
+        x = torch.rand(2, 3, 8, 8, device=device, dtype=dtype)
+        torch._dynamo.reset()
+        compiled = torch.compile(kornia.enhance.equalize, fullgraph=True, backend="eager")
+        self.assert_close(compiled(x), kornia.enhance.equalize(x))
+
     @pytest.mark.skip(reason="args and kwargs in decorator")
     def test_jit(self, device, dtype):
         batch_size, channels, height, width = 1, 2, 3, 3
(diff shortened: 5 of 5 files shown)


ducha-aiki (dismissed):
First review of head `177fbe0d`.

Thanks, this is a clean fix. The check validates `scaled` in `(-1, 256)`, which is exactly the set of values `scaled.long()` can index. It is computed on the same tensor the gather uses, so the domain matches in every dtype. It goes through the existing `_assert_async_value_check`, so there is no `.item()` and no graph break. I ran both trees across four dtypes, and the only inputs that behave differently are ones that crashed on `main`, with one exception covered in the first P3 below. For example, float16 `x * 255.99/255` crashed on `main` with `index 256` because of float16 rounding, and on this head it gets the new message. The PR also fixes something the description doesn't claim. On `main`, `torch.compile(equalize, fullgraph=True)` with the inductor backend **aborts the whole process** on out-of-range CPU input (`libc++abi: terminating due to uncaught exception ... index out of bounds: 0 <= tmp17 < 256LL`, exit 134). On this head the same call raises a catchable `RuntimeError` with the new message. Both issues are fully covered. #4431 asked for option 1, a tensor-level check plus public docstrings, and gets that with the policy question left to #4430. #4432 asked for the range, the whole-volume 256-bin histogram and the at-most-255-voxel identity with no promise above that, and gets all three plus the validation. The only merge conflict is `CHANGELOG.md`.

**[P3] NaN input is a new rejection on aarch64 CPU, which the CHANGELOG entry doesn't mention.** On `main`, `NaN.long()` saturates to `0` on arm64. An image with a NaN pixel therefore equalized silently: the NaN went into bin 0 and was mapped to `lut[0]`. Every dtype printed `OK out[min,max]=(0,1) nan=False`. On this head, `NaN > -1.0` is `False`, so the same input raises `equalize expects input values in [0, 1]` in eager mode, under inductor, and on torch 2.5.1. `±inf` crashed on `main` already. I expect x86 to have crashed on NaN too, because `cvttsd2si` gives `INT64_MIN`, but I did not run that. Rejecting NaN is the right outcome. The problem is the entry's sentence "inputs that equalized before ... are unchanged", which doesn't hold for a NaN-containing image on Apple Silicon or Linux aarch64. Suggest adding "and inputs containing NaN or ±inf now raise too", and optionally mentioning NaN in the message.

**[P3] Docstring precision (optional).**
- `equalize3d` and `RandomEqualize3D` say a volume of at most 255 voxels per channel is "returned unchanged". The step-0 path returns `(x * 255) / 255`, which is equal only up to rounding. With `(2, 3, 3, 5, 17)` in float32, 26 of 1530 voxels differ by 2.98e-8, and in float16, 7 voxels differ by 2.4e-4. The new test passes because it uses `assert_close`. "Unchanged up to floating-point rounding" would be accurate.
- The `RandomEqualize` / `RandomEqualize3D` notes say out-of-range values raise, but not that the check is skipped on MPS. On MPS you still get the raw `AcceleratorError: gather: index 439 is out of bounds` on both trees. They also don't say that the whole batch is checked regardless of `p`. The intensity base runs `apply_transform` on every row before blending, so `RandomEqualize(p=0.0)(x * 2)` raises on both trees. One clause pointing at the function's note would cover both points.

**[P3] `test_dynamo_fullgraph` hard-codes `backend="eager"`** rather than using the injected `torch_optimizer` fixture. It passes on `main` as well, which is fine for a guard. The fullgraph path is already guarded under inductor by `TestRandomEqualizeAlternative::test_dynamo`, so this is optional.

**CI:** the one red leg, macOS / py3.12 / torch 2.5.1, fails only `tests/contrib/test_super_resolution.py::TestSuperResolutionBuilders::test_to_onnx_exports[RRDBNetBuilder]`: `Timeout (>120.0s)` (1 failed, 10031 passed). This is the known timeout flake, and `collector` fails as a consequence. It is unrelated to this change.

**Verified on this head:**
- `tests/enhance/test_adjust.py -k Equalize`:
  - CPU float32+float64: 38 passed, 4 skipped.
  - CPU float16 (`cpu-float16` profile): 20 passed, 2 skipped. bfloat16 (`cpu-bfloat16` profile): 20 passed, 2 skipped.
  - MPS float32: 14 passed, 8 skipped.
  - torch 2.5.1 CPU float32: 20 passed, 2 skipped.
- Copying the PR's test file onto `main`: 8 failed, all 8 `test_out_of_range_input_names_the_range` cases (2D and 3D, float32+float64), and 30 passed.
- `tests/augmentation -k Equalize` on CPU float32: 20 passed, 4 skipped, 1 xfailed, 1 xpassed, the same markers as on `main`.
- Doctests on the three touched modules: 28 passed.
- `KORNIA_TEST_OPTIMIZER=inductor`: equalize dynamo tests 4/4 on head and 2/2 on `main`.
- `torch._dynamo.explain` on `equalize`, `equalize3d` and `RandomEqualize(p=0.5)`: 1 graph and 0 breaks on both trees, and on torch 2.5.1 for `equalize`.
- Inductor and eager-backend compiled output matches eager on valid input (max diff 0.0).
- In-process interleaved A/B of the check's CPU eager cost: +0.6% at `(8, 3, 224, 224)` and +4.6% at `(1, 3, 512, 512)`.
- Grep for `#4431|#4432|#4430` finds no hits in `kornia/` on either tree, and only the two new test comments in `tests/`. #4452 added no equalize wart pins that would need updating.
- No duplicate PRs. `ruff@0.16.6` check and format are clean.

ducha-aiki (dismissed):
Re-approving on `3f3edfb8`. To clear the `CHANGELOG.md`-only conflict (`main` added entries at the top of the same `### Bug fixes` list), I pushed `3f3edfb8`, a keep-both merge of `origin/main` (`dc0af1a4`) into the head `177fbe0d` I reviewed above. It is a pure merge: every path except `CHANGELOG.md` is tree-identical to `git merge-tree`, and the non-blank lines it adds to `CHANGELOG.md` are exactly this branch's own. The review above stands unchanged. If you push again, please pull this merge first.

PR #4490, opened 2026-09-13
docs(augmentation): make three Args blocks match their constructors

### LINKED ISSUE #4437: W1b: `ColorJitter`'s `Args` documents a nonexistent argument and omits `order`
`ColorJitter`'s `Args` block lists `silence_instantiation_warning: if True, silence the warning at instantiation.` at `kornia/augmentation/_2d/intensity/color_jitter.py:48`. No such parameter exists in the constructor, and `grep -rn silence_instantiation_warning` over the repository finds the name in exactly three docstrings — `color_jitter.py:48`, `brightness.py:38` and `gaussian_blur.py:46` — and in no code. A reader following the documentation writes a call that raises `TypeError`.

The same block omits the argument that is actually there. `order` is a real constructor parameter, and it is load-bearing: passing a fixed order makes `apply_transform` iterate a static Python sequence instead of the drawn `order` tensor, which is what makes the class `torch.compile` fullgraph-safe (`color_jitter.py:108-118`, `:151-156`). Nothing in the rendered documentation says the argument exists.

`order` also leaves a trap in the replay state. The parameter generator draws an `order` key into `_params` whether or not the constructor fixed one, and with a fixed order the drawn key is ignored: replaying with `_params["order"]` reversed changes nothing, while the same edit on a default instance changes the output. `_params` is otherwise the complete replay state of an augmentation, so a key that is present, drawn, and silently disregarded is the kind of thing a caller building a replay pipeline will get wrong. `ColorJiggle`, the sibling this class points at, has no mismatch in either direction.

## Repro

```python
import inspect, re, torch, kornia.augmentation as K

def documented(cls):
    doc = inspect.getdoc(cls) or ""
    block = re.search(r"Args:\n(.*?)\n\s*(?:Shape|Returns|Examples?):", doc, re.S)
    return [m.group(1) for m in re.finditer(r"^\s{4}(\w+):", block.group(1), re.M)]

for cls in (K.ColorJitter, K.ColorJiggle, K.RandomBrightness, K.RandomGaussianBlur):
    sig = list(inspect.signature(cls.__init__).parameters)[1:]
    doc = documented(cls)
    print(f"{cls.__name__}: documented but absent from the signature {[a for a in doc if a not in sig]} | "
          f"in the signature but undocumented {[a for a in sig if a not in doc]}")

print()
torch.manual_seed(0)
x = torch.rand(2, 3, 6, 8)
for order in ((0, 1, 2, 3), None):
    torch.manual_seed(7)
    aug = K.ColorJitter(0.2, 0.2, 0.2, 0.1, p=1.0, order=order)
    out = aug(x.clone())
    params = dict(aug._params)
    print(f"order={str(order):<12} drawn _params['order'] = {params['order'].tolist()}")
    params["order"] = torch.flip(params["order"], (0,))
    replay = aug(x.clone(), params=params)
    print(f"{'':<19}replaying with that key reversed changes the output: "
          f"{not torch.equal(out, replay)}  max|diff| {(out - replay).abs().max().item():.6g}")
```

## Observed

```
ColorJitter: documented but absent from the signature ['silence_instantiation_warning'] | in the signature but undocumented ['order']
ColorJiggle: documented but absent from the signature [] | in the signature but undocumented []
RandomBrightness: documented but absent from the signature ['silence_instantiation_warning'] | in the signature but undocumented []
RandomGaussianBlur: documented but absent from the signature ['silence_instantiation_warning'] | in the signature but undocumented []

order=(0, 1, 2, 3) drawn _params['order'] = [0, 3, 2, 1]
                   replaying with that key reversed changes the output: False  max|diff| 0
order=None         drawn _params['order'] = [0, 3, 2, 1]
                   replaying with that key reversed changes the output: True  max|diff| 0.0378586
```

## Expected

The `Args` block matches the signature: `silence_instantiation_warning` is removed from `color_jitter.py:48` — and, in the same pass, from `brightness.py:38` and `gaussian_blur.py:46`, where the same phantom entry appears — and `order` is documented, including that it fixes the application order and is what makes the class fullgraph-compilable.

`_params["order"]` is a separate, smaller decision: either the generator stops drawing the key when the constructor fixed an order, or the docstring states that with `order=` set the key is present but not used. The first changes the RNG consumption of an existing call, so the audit recommends documenting it here and treating the draw as its own change.

## Classification

**API completeness** — public documentation that names an argument the code does not accept and omits one the code does, plus a replay key that is drawn and ignored. Idiom: none.

## Disposition

`n/a — single symbol`. Batch 6 documents the current behavior as it is and pins it. The docstring correction is a focused fix and welcome as a PR, and it should cover all three classes carrying the phantom entry.

## Environment

dchaudhari7177:

The docstring fix from #4437, across all three classes carrying the phantom entry:

- `ColorJitter`, `RandomBrightness` and `RandomGaussianBlur` no longer document `silence_instantiation_warning`. None of them accepts it, and it appeared in no code, so a call written from the docs raised `TypeError`.
- `ColorJitter` now documents `order`: indices into (brightness, contrast, saturation, hue) where a subset applies only those; `None` draws a random order per call; a fixed order makes the transform `torch.compile` fullgraph-safe. It also states the replay trap from the issue: with `order` set, the generator still draws `_params["order"]` and that entry is ignored, including on replay. This takes the issue's recommendation to document the key rather than stop drawing it, since stopping would change RNG consumption.

No code changes.

Closes #4437 (tracking #4407).

## What did you check?

`tests/augmentation/test_args_docstrings.py` (new) parses each class's `Args` block and asserts it lists exactly the constructor's parameters, for `ColorJitter`, `ColorJiggle` (the sibling the issue uses as the control), `RandomBrightness` and `RandomGaussianBlur`.

```text
$ python -m pytest tests/augmentation/test_args_docstrings.py
main: 3 failed, 1 passed      # ColorJiggle already matched
head: 4 passed
$ python -m pytest kornia/augmentation/_2d/intensity/{color_jitter,brightness,gaussian_blur}.py --doctest-modules
3 passed
```

`grep -rn silence_instantiation_warning kornia/` is empty. `ruff@0.16.6` check and format are clean.

## How did AI help?

I used Claude Code to draft the docstring edits, the test and this description. I checked the `order` handling in `apply_transform` against the new wording and ran the test on both trees.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -381,6 +381,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* The `Args` blocks of `ColorJitter`, `RandomBrightness` and `RandomGaussianBlur` no longer document a
+  `silence_instantiation_warning` argument that none of them accepts, and `ColorJitter` now documents its
+  `order` argument: a fixed (sub)set of brightness/contrast/saturation/hue indices that makes the transform
+  `torch.compile` fullgraph-safe, with the drawn `_params["order"]` entry ignored when it is set.
+  (#4437, #4490)
+
 * `RandomMosaic` now preserves `(H, W)` for non-square inputs when `output_size=None`; it previously
   returned `(W, H)`. `start_ratio_range` now scales x by width and y by height; it previously scaled
   x by height and y by width. (#4459)
--- a/kornia/augmentation/_2d/intensity/brightness.py
+++ b/kornia/augmentation/_2d/intensity/brightness.py
@@ -35,7 +35,6 @@ class RandomBrightness(IntensityAugmentationBase2D):
     Args:
         brightness: the brightness factor to apply
         clip_output: if true clip output
-        silence_instantiation_warning: if True, silence the warning at instantiation.
         same_on_batch: apply the same transformation across the batch.
         p: probability of applying the transformation.
         keepdim: whether to keep the output shape the same as input (True) or broadcast it
--- a/kornia/augmentation/_2d/intensity/color_jitter.py
+++ b/kornia/augmentation/_2d/intensity/color_jitter.py
@@ -45,11 +45,14 @@ class ColorJitter(IntensityAugmentationBase2D):
         contrast: The contrast factor to apply.
         saturation: The saturation factor to apply.
         hue: The hue factor to apply.
-        silence_instantiation_warning: if True, silence the warning at instantiation.
         same_on_batch: apply the same transformation across the batch.
         p: probability of applying the transformation.
         keepdim: whether to keep the output shape the same as input (True) or broadcast it
                  to the batch form (False).
+        order: a fixed application order, as indices into (brightness, contrast, saturation, hue); a subset
+          applies only those. ``None`` (the default) draws a random order on every call. A fixed order makes
+          the transform ``torch.compile`` fullgraph-safe. The parameter generator still draws an ``order``
+          entry into ``_params``, and with a fixed order that entry is ignored, including on replay.
     Shape:
         - Input: :math:`(C, H, W)` or :math:`(B, C, H, W)`, Optional: :math:`(B, 3, 3)`
         - Output: :math:`(B, C, H, W)`
--- a/kornia/augmentation/_2d/intensity/gaussian_blur.py
+++ b/kornia/augmentation/_2d/intensity/gaussian_blur.py
@@ -43,7 +43,6 @@ class RandomGaussianBlur(IntensityAugmentationBase2D):
         p: probability of applying the transformation.
         keepdim: whether to keep the output shape the same as input (True) or broadcast it
                  to the batch form (False).
-        silence_instantiation_warning: if True, silence the warning at instantiation.
 
     Shape:
         - Input: :math:`(C, H, W)` or :math:`(B, C, H, W)`, Optional: :math:`(B, 3, 3)`
--- a/tests/augmentation/test_args_docstrings.py
+++ b/tests/augmentation/test_args_docstrings.py
@@ -1,0 +1,39 @@
+# LICENSE HEADER MANAGED BY add-license-header
+#
+# Copyright 2018 Kornia Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import inspect
+import re
+
+import pytest
+
+import kornia.augmentation as K
+
+
+def _documented_args(cls):
+    doc = inspect.getdoc(cls) or ""
+    block = re.search(r"Args:\n(.*?)\n\s*(?:Shape|Returns|Examples?|\.\. note::|Note)", doc, re.S)
+    assert block, f"{cls.__name__} has no Args block"
+    return [match.group(1) for match in re.finditer(r"^\s{4}(\w+):", block.group(1), re.M)]
+
+
+@pytest.mark.parametrize("cls", [K.ColorJitter, K.ColorJiggle, K.RandomBrightness, K.RandomGaussianBlur])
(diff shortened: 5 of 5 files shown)


ducha-aiki (dismissed):
First review of head `bbe238f2`. Thanks, this is a clean, well-scoped fix: all three phantom `silence_instantiation_warning` entries named in #4437 are gone, `order` is documented, and every sentence of the new `order` text holds up when you run it. `Closes #4437` matches the issue's full scope (the three docstrings plus `order`, with the replay key documented rather than changed, which is what the issue recommends). Nothing blocks the merge. The notes below are optional.

**[P3] `order` wording leaves out repeats, the empty tuple and the error case** (`kornia/augmentation/_2d/intensity/color_jitter.py:52`). "A subset applies only those" is true, but the constructor only checks `set(order) <= {0, 1, 2, 3}`. So `order=(0, 0)` applies brightness twice (the output matches two chained `adjust_brightness_accumulative` calls), `order=()` returns the input unchanged, and `order=(0, 1, 9)` raises `ValueError`. One extra clause would cover it, for example "repeated indices apply that adjustment again; an index outside 0..3 raises `ValueError`".

**[P3] The replay sentence is now documented behaviour, but no test pins it.** The docstring says the drawn `_params["order"]` is ignored when `order` is fixed. I checked it: on a fixed-order instance, reversing `_params["order"]` or deleting the key leaves replay unchanged. On a default instance, the reversed key changes the output (max |diff| ≈ 0.038) and a missing key raises `KeyError: 'order'`. A few lines in `TestColorJitter` would catch a later generator change that makes this sentence stale, for example one that stops drawing the key when `order` is fixed.

**[P3] Outside this PR's scope:** the same signature-vs-`Args` comparison, run over every public class in `kornia.augmentation`, finds three more phantom arguments. Each one raises `TypeError` when passed. `RandomThinPlateSpline` documents `mode`, but the constructor takes `padding_mode`. `RandomCutMixV2` documents `height` and `width`. `Denormalize` documents `same_on_batch`. #4437 doesn't cover these, so the closing keyword is correct and the four-class parametrization is fine as it is. They should get their own issue.

**Merge state:** GitHub reports a conflict. It is CHANGELOG-only: `main` added #4462/#4461/#4460 entries at the top of the same `### Bug fixes` list. A keep-both merge resolves it. The code files don't overlap with the four commits `main` gained since `a7e2db6b`.

**Verified on this head:**
- Docstrings vs signatures: the new `order` entry matches `order: Optional[Sequence[int]] = None`. The three removals match constructors that take no such argument. No other `Args` line changed, and the diff has no code changes.
- Behaviour claims (torch 2.14.0, CPU): `order=(0,)` output equals `adjust_brightness_accumulative` with the drawn factor alone. `torch.compile(..., fullgraph=True, backend="eager")` succeeds with `order=(0, 1, 2, 3)` and fails with `order=None` (data-dependent `UserError`). On torch 2.5.1, fullgraph compile succeeds for `(0, 1, 2, 3)` and `(2, 0)`. Replay results are as described above.
- `tests/augmentation/test_args_docstrings.py`: 4 passed on head. Copied onto `main` (`dc0af1a4`) without the docstring fix: 3 failed (`ColorJitter`, `RandomBrightness`, `RandomGaussianBlur`) and 1 passed (`ColorJiggle`). With the PR's `kornia/` and `tests/` changes applied on top of `main`: 4 passed.
- Doctests of `color_jitter.py`, `brightness.py` and `gaussian_blur.py`: 3 passed, on head and on head's changes applied to `main`.
- Render check: I ran the three docstrings through `sphinx.ext.napoleon` (`GoogleDocstring`) and then docutils. `order` renders as a single `:param order:` field with the `` ``None`` ``, `` ``torch.compile`` `` and `` ``_params`` `` literals intact. The only docutils messages are the Sphinx-only `:math:`/`:func:` roles, and base shows the same set.
- `grep -rnE "#4437|issues/4437" kornia/ tests/`: no hit in `kornia/`. The only hit in `tests/` is the new test's comment. `main` has no #4437 pin that would need flipping.
- `ruff@0.16.6` check and format are clean on the four changed Python files, and no changed line exceeds 120 columns. The CHANGELOG entry is under `## Unreleased` / `### Bug fixes` and cites the PR.
- No duplicate PRs for #4437.

ducha-aiki (dismissed):
Re-approving on `279183aa`. To clear the `CHANGELOG.md`-only conflict (`main` added entries at the top of the same `### Bug fixes` list), I pushed `279183aa`, a keep-both merge of `origin/main` (`dc0af1a4`) into the head `bbe238f2` I reviewed above. It is a pure merge: every path except `CHANGELOG.md` is tree-identical to `git merge-tree`, and the non-blank lines it adds to `CHANGELOG.md` are exactly this branch's own. The review above stands unchanged. If you push again, please pull this merge first.

Issue #3934, opened 2026-08-10
W1b: bbox width/height arithmetic is inclusive (+1) — semantic repair candidate for the coordinated deprecation window


ducha-aiki:
## Behavior

`kornia.geometry.bbox` computes box extents with inclusive (+1) arithmetic, e.g. `infer_bbox_shape` returns `width = x_br - x_tl + 1` (`kornia/geometry/bbox.py:140-141`, same pattern in the 3D variants at lines 95-105 and 184-190). `validate_bbox` (lines 60-61) writes the same `+1` terms but is **not** in scope for the semantic convention repair: they cancel algebraically inside its width and height comparisons. That invariance holds in exact arithmetic, but finite-precision rounding can change the result; in particular, float16 boxes with spans below the dtype's resolution can validate under the inclusive calculation while failing the exclusive comparison. The separate `validate_bbox` contract wart is #4013.

A degenerate box with identical corners therefore has width/height 1, not 0.

This is a classic convention trap: COCO, torchvision, and albumentations use exclusive arithmetic (`width = x_br - x_tl`), so round-tripping boxes between kornia and those ecosystems silently shifts sizes by one pixel. Code keeps running while meaning changes — the worst class of geometry bug. Previously discussed in #1142 (closed Q&A, 2022); this issue is the dedicated, tracked repair candidate required by the convention-repair protocol (one issue per finding, never a closed umbrella).

## Protocol

1. **Document first**: the upcoming bbox/boxes convention-block batch (see ROADMAP) documents the current inclusive behavior including this wart, with `test_convention_*` pinning tests. A documented + pinned wart is interim spec.
2. **Repair versioned**: changing the arithmetic is a semantic default change, so it batches into the coordinated Repair & Deprecation window — deprecation cycle per the [stability policy](https://kornia.readthedocs.io/en/latest/get-started/stability.html) (promise 2: no silent semantic changes), landing at a minor-release boundary, with an old-vs-new conformance vector pair and one migration note.
3. Do **not** fix this ad hoc: any interim PR changing the +1 arithmetic outside the window will silently break downstream users.

Relates to #1142 (original report). Roadmap context: #3933.

Posted on behalf of [@ducha-aiki](https://github.com/ducha-aiki) by Codex (gpt-5.6-sol).

ducha-aiki:
Two corrections to this issue's scope, both executed against `4ab79c78`.

[...]

PR #4461, opened 2026-09-12
Reject mismatched input counts in many-to-many augmentation

Shy7777:
## What does this pull request do?

`ManyToManyAugmentationDispather` silently truncates a call when the number of input bundles differs from the number of augmentations. For example, two image/mask augmentation sequences given three bundles return only two results.

This raises a `ValueError` naming both counts before any augmentation runs or its replay parameters change. Matching calls still support independent bundles with different `data_keys`, including image-only and image/mask inputs. Empty-to-empty dispatch remains valid.

## Related issue or discussion

Related to #4422. This implements the focused input-count check; it does not add the issue's proposed empty-constructor or uniform-data-key restrictions. The documentation and test pins in #4452 describe the current truncation and will need to account for the runtime fix when that PR is integrated.

## What did you check?

The two non-blocking review suggestions are addressed: the empty-dispatcher test now points to #4422, and the changelog citation follows the surrounding style. The 21 focused tests and full pre-commit checks were rerun after these edits.

The new mismatch cases fail on the original code: 8 failures and 13 passes across float32/float64. After the fix, all 21 focused tests pass.

- Full container tests on CPU with float32/float64 and `KORNIA_TEST_OPTIMIZER=aot_eager`: 962 passed, 102 skipped.
- New dispatcher cases under the float16 and bfloat16 CPU profiles: 7 passed for each dtype, with no xfails.
- Dispatcher doctests: 2 passed.
- Full `pixi run pre-commit-all`: passed.
- `pixi run typecheck --python .venv`: passed with an existing deprecation warning in unchanged LightGlue code.

These checks used the project's Pixi environment on macOS with Python 3.11 and PyTorch 2.14. No GPU execution or model downloads were needed.

[...]

PR #4491, opened 2026-09-13
fix(augmentation): document RandomCutMixV2's cut_size and reject a minimum of 1

### LINKED ISSUE #4439: W1b: `RandomCutMixV2.cut_size` is the inverse of its name, and `(1.0, 1.0)` builds a negative-extent box
`RandomCutMixV2`'s docstring says `cut_size` is "controlling the minimum and maximum cut ratio from [0, 1]". A
reader takes that to mean the cut covers between `cut_size[0]` and `cut_size[1]` of each side, so a larger
`cut_size` cuts more. It does the opposite: `cut_size` is the clamp applied to the Beta-sampled mixing coefficient
`lambda`, and the cut side is `floor(sqrt(1 - lambda) * side)`. Raising `cut_size` raises `lambda`, which shrinks
the cut. On a `6x8` image, `cut_size=(0.1, 0.1)` cuts a `6 x 4` box and `cut_size=(0.9, 0.9)` cuts a `1 x 0` one.

The end of that range is not merely degenerate. `random_generator/_2d/cutmix.py:130-137` clamps the draw and takes
`cutmix_rate = sqrt(1.0 - cutmix_betas)`, so `cut_size=(1.0, 1.0)` gives `cutmix_rate = 0`, a cut height and width
of zero, and — because `bbox_generator` places the far corner at `start + size - 1` — a box whose extents are both
`-1`. That box is built, stored in `_params["crop_src"]` and handed on to `Boxes` and `bbox_to_mask` without a
check; on some draws one of its corners is at a negative coordinate, outside the image. No exception is raised and
the augmentation silently becomes an identity with `lam = 1.0`.

The value a user is most likely to write for "cut the whole image" is exactly the one that cuts nothing.

## Repro

```python
import torch


torch.manual_seed(0)
x = torch.rand(4, 3, 6, 8)                        # H = 6, W = 8
label = torch.tensor([0, 1, 2, 3])

for cs in ((0.1, 0.1), (0.5, 0.5), (0.9, 0.9), (1.0, 1.0)):
    torch.manual_seed(0)
    aug = K.RandomCutMixV2(num_mix=1, p=1.0, cut_size=cs, use_correct_lambda=True, data_keys=["input", "class"])
    out, lab = aug(x, label)
    box = aug._params["crop_src"][0, 0]            # vertices (x, y), clockwise from top-left
    dx = (box[1, 0] - box[0, 0]).item()
    dy = (box[2, 1] - box[1, 1]).item()
    print(f"cut_size={cs}  cut (dx, dy) = ({dx}, {dy})  min_y = {box[:, 1].min().item()}"
          f"  lam = {round(lab[0][0, 2].item(), 4)}  image_changed = {bool((out != x).any())}")
    print(f"    crop_src[0, 0] = {[[round(v, 1) for v in pt] for pt in box.tolist()]}")

# the degenerate box is not a one-seed accident, and on some draws it leaves the image
for seed in range(8):
    torch.manual_seed(seed)
    aug = K.RandomCutMixV2(num_mix=1, p=1.0, cut_size=(1.0, 1.0), use_correct_lambda=True, data_keys=["input", "class"])
    aug(x, label)
    boxes = aug._params["crop_src"][0]
    extents = sorted({(round((b[1, 0] - b[0, 0]).item()), round((b[2, 1] - b[1, 1]).item())) for b in boxes})
    print(f"cut_size=(1.0, 1.0)  seed={seed}  extents over the batch = {extents}"
          f"  min coordinate = {boxes.min().item()}")
```

## Observed

```
cut_size=(0.1, 0.1)  cut (dx, dy) = (6.0, 4.0)  min_y = 0.0  lam = 0.2708  image_changed = True
    crop_src[0, 0] = [[0.0, 0.0], [6.0, 0.0], [6.0, 4.0], [0.0, 4.0]]
cut_size=(0.5, 0.5)  cut (dx, dy) = (4.0, 3.0)  min_y = 0.0  lam = 0.5833  image_changed = True
    crop_src[0, 0] = [[1.0, 0.0], [5.0, 0.0], [5.0, 3.0], [1.0, 3.0]]
cut_size=(0.9, 0.9)  cut (dx, dy) = (1.0, 0.0)  min_y = 0.0  lam = 0.9583  image_changed = True
    crop_src[0, 0] = [[2.0, 0.0], [3.0, 0.0], [3.0, 0.0], [2.0, 0.0]]
cut_size=(1.0, 1.0)  cut (dx, dy) = (-1.0, -1.0)  min_y = 0.0  lam = 1.0  image_changed = False
    crop_src[0, 0] = [[3.0, 1.0], [2.0, 1.0], [2.0, 0.0], [3.0, 0.0]]
cut_size=(1.0, 1.0)  seed=0  extents over the batch = [(-1, -1)]  min coordinate = 0.0
cut_size=(1.0, 1.0)  seed=1  extents over the batch = [(-1, -1)]  min coordinate = 0.0
cut_size=(1.0, 1.0)  seed=2  extents over the batch = [(-1, -1)]  min coordinate = -1.0
cut_size=(1.0, 1.0)  seed=3  extents over the batch = [(-1, -1)]  min coordinate = 1.0
cut_size=(1.0, 1.0)  seed=4  extents over the batch = [(-1, -1)]  min coordinate = 3.0
cut_size=(1.0, 1.0)  seed=5  extents over the batch = [(-1, -1)]  min coordinate = 0.0
cut_size=(1.0, 1.0)  seed=6  extents over the batch = [(-1, -1)]  min coordinate = 2.0
cut_size=(1.0, 1.0)  seed=7  extents over the batch = [(-1, -1)]  min coordinate = -1.0
```

## Expected

Two halves, and they can land separately.

The naming half is a documentation fix and is settled: `cut_size` is the `[min, max]` clamp applied to the
Beta-sampled `lambda`, the cut side is `floor(sqrt(1 - lambda) * side)`, and a larger `cut_size` therefore yields a
smaller cut. Renaming the argument would be a breaking change for every existing caller and is not what the audit
recommends; saying what it does is.

The degenerate-box half is a validation fix. A clamp range that can produce `cutmix_rate = 0` should either be
rejected at construction with a kornia-level message, or clamped so the generated box is never inverted. A box with
`dx = dy = -1` reaching `Boxes` and `bbox_to_mask` is an unvalidated precondition, not a documented outcome — the
inclusive `start + size - 1` arithmetic that produces it is the same one tracked in #3934.

## Classification

**Semantic contract** (the `cut_size` naming — the documented meaning of a public constructor argument is the
inverse of its behavior) and **API completeness** (the degenerate box at `cut_size=(1.0, 1.0)` — a clamp range that
produces `dx = dy = -1` is accepted and passed to `Boxes` with no guard). Idiom: 4 (degenerate input silently
producing garbage rather than an error).

## Disposition

`n/a — single symbol`. Both halves are `RandomCutMixV2` and its generator. The docstring correction and a guard on
the degenerate clamp are a focused fix, welcome as a PR. Batch 6 documents the current `cut_size` meaning on the
class and pins it, and does not change it.

## Environment

change it.

dchaudhari7177:

Both halves of #4439.

**Naming (docs).** `cut_size` on `RandomCutMixV2` and `CutmixGenerator` is now documented as the `[min, max]` clamp on the Beta-sampled mixing coefficient `lambda`, with the cut side `floor(sqrt(1 - lambda) * side)`. So a larger `cut_size` gives a **smaller** cut. The argument keeps its name, as the issue recommends.

**Degenerate clamp (validation).** A minimum of `1.0` forces `lambda = 1`, so `cutmix_rate = 0` and the box has `dx = dy = -1` (inverted extents that sometimes sit at negative coordinates), and the augmentation silently becomes an identity. `CutmixGenerator.make_samplers` now raises `ValueError` for `cut_size[0] >= 1`, naming the cause:

```
`cut_size` clamps the mixing coefficient lambda, and a minimum of 1 forces lambda = 1, which cuts nothing
(cut side = floor(sqrt(1 - lambda) * side)). A larger cut_size gives a smaller cut, so lower the minimum. Got [1.0, 1.0].
```

This rejects the clamp at construction, the first of the issue's two options. I didn't clamp box sizes to at least 1 px instead: the same zero-size arithmetic is how non-applied samples (`p` misses, `cutmix_rate = 0`) flow today, so a floor there would change the output of every current call. It also doesn't touch the inclusive `start + size - 1` corner (#3934). The default `[0, 1]` and any minimum below 1 are unaffected; the check uses `float()` once at sampler construction, not per call.

Closes #4439 (tracking #4407; box arithmetic tracked in #3934).

## What did you check?

`tests/augmentation/test_random_generator.py`:
- `test_invalid_param_combinations` gains `cut_size=[1.0, 1.0]` for both `same_on_batch` values;
- `test_a_minimum_cut_size_of_one_is_rejected_by_name_4439`;
- `test_a_larger_cut_size_gives_a_smaller_cut_4439` pins the documented direction (0.1 cuts wider than 0.9). It passes on `main` too, since it pins the contract the new docstring states.

```text
$ python -m pytest tests/augmentation/test_random_generator.py -k Cutmix --device=cpu --dtype=float32
main: 3 failed, 341 passed      # the [1.0, 1.0] cases (x2) and the named rejection
head: 344 passed                # 688 across float32 + float64
$ python -m pytest tests/augmentation/test_augmentation_mix.py -k CutMix --device=cpu --dtype=float32
8 passed
```

Doctests on both modules pass. `ruff@0.16.6` check and format are clean. No CUDA/MPS run.

## How did AI help?

I used Claude Code to trace `cut_size` through the generator, draft the docstrings, the guard and the tests, and write this description. I reproduced the inverted box on `main` and ran the tests above on both trees.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -381,6 +381,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `RandomCutMixV2` and `CutmixGenerator` document `cut_size` as what it is: the `[min, max]` clamp on the
+  Beta-sampled mixing coefficient `lambda`, where the cut side is `floor(sqrt(1 - lambda) * side)`, so a larger
+  `cut_size` gives a smaller cut. It was described as the "minimum and maximum cut ratio". A minimum of `1.0` is
+  now rejected with a `ValueError`: it forced `lambda = 1`, built an inverted zero-size box and silently made the
+  augmentation an identity. (#4439, #4491)
+
 * `RandomMosaic` now preserves `(H, W)` for non-square inputs when `output_size=None`; it previously
   returned `(W, H)`. `start_ratio_range` now scales x by width and y by height; it previously scaled
   x by height and y by width. (#4459)
--- a/kornia/augmentation/_2d/mix/cutmix.py
+++ b/kornia/augmentation/_2d/mix/cutmix.py
@@ -48,8 +48,10 @@ class RandomCutMixV2(MixAugmentationBaseV2):
         num_mix: cut mix times.
         beta: hyperparameter for generating cut size from beta distribution.
             Beta cannot be set to 0 after torch 1.8.0. If None, it will be set to 1.
-        cut_size: controlling the minimum and maximum cut ratio from [0, 1].
-            If None, it will be set to [0, 1], which means no restriction.
+        cut_size: the ``[min, max]`` clamp, within [0, 1], applied to the Beta-sampled mixing coefficient
+            ``lambda``, not the fraction of the image to cut. The cut side is ``floor(sqrt(1 - lambda) * side)``,
+            so a larger ``cut_size`` gives a *smaller* cut, and a minimum of 1 is rejected because it would cut
+            nothing. If None, it will be set to [0, 1], which means no restriction.
         same_on_batch: apply the same transformation across the batch.
             This flag will not maintain permutation order.
         keepdim: whether to keep the output shape the same as input (True) or broadcast it
--- a/kornia/augmentation/random_generator/_2d/cutmix.py
+++ b/kornia/augmentation/random_generator/_2d/cutmix.py
@@ -42,8 +42,10 @@ class CutmixGenerator(RandomGeneratorBase):
         num_mix (int): number of images to mix with. Default is 1.
         beta (torch.Tensor, optional): hyperparameter for generating cut size from beta distribution.
             If None, it will be set to 1.
-        cut_size (torch.Tensor, optional): controlling the minimum and maximum cut ratio from [0, 1].
-            If None, it will be set to [0, 1], which means no restriction.
+        cut_size (torch.Tensor, optional): the ``[min, max]`` clamp, within [0, 1], applied to the
+            Beta-sampled mixing coefficient ``lambda``. The cut side is ``floor(sqrt(1 - lambda) * side)``,
+            so a larger ``cut_size`` gives a *smaller* cut. The minimum must be below 1: at 1, ``lambda`` is
+            always 1 and nothing is cut. If None, it will be set to [0, 1], which means no restriction.
 
     Returns:
         params Dict[str, torch.Tensor]: parameters to be passed for transformation.
@@ -88,6 +90,12 @@ class CutmixGenerator(RandomGeneratorBase):
             self._cut_size = torch.as_tensor(self.cut_size, device=device, dtype=dtype)
 
         _joint_range_check(self._cut_size, "cut_size", bounds=(0, 1))
+        if float(self._cut_size[0]) >= 1.0:
+            raise ValueError(
+                f"`cut_size` clamps the mixing coefficient lambda, and a minimum of 1 forces lambda = 1, which "
+                f"cuts nothing (cut side = floor(sqrt(1 - lambda) * side)). A larger cut_size gives a smaller "
+                f"cut, so lower the minimum. Got {self._cut_size.tolist()}."
+            )
 
         self.beta_sampler = Beta(self._beta, self._beta)
         self.prob_sampler = Bernoulli(torch.tensor(float(self.p), device=device, dtype=dtype))
--- a/tests/augmentation/test_random_generator.py
+++ b/tests/augmentation/test_random_generator.py
@@ -1487,6 +1487,7 @@ class TestRandomCutMixGen(RandomGeneratorBaseTests):
             (200, 200, 1, torch.tensor([0.0, 1.0]), None),
             (200, 200, 1, None, torch.tensor([-1.0, 1.0])),
             (200, 200, 1, None, torch.tensor([0.0, 2.0])),
+            (200, 200, 1, None, torch.tensor([1.0, 1.0])),
         ],
     )
     @pytest.mark.parametrize("same_on_batch", [True, False])
@@ -1498,6 +1499,23 @@ class TestRandomCutMixGen(RandomGeneratorBaseTests):
                 beta=(beta.to(device=device, dtype=dtype) if isinstance(beta, (Tensor)) else beta),
                 cut_size=(cut_size.to(device=device, dtype=dtype) if isinstance(cut_size, (Tensor)) else cut_size),
             )(torch.Size([8, 3, height, width]), same_on_batch=same_on_batch)
+
+    def test_a_minimum_cut_size_of_one_is_rejected_by_name_4439(self, device, dtype):
+        # kornia#4439: cut_size clamps lambda, so (1.0, 1.0) forced lambda = 1 and built an inverted,
+        # zero-size box that silently made the augmentation an identity.
+        with pytest.raises(ValueError, match="forces lambda = 1, which cuts nothing"):
+            CutmixGenerator(cut_size=torch.tensor([1.0, 1.0], device=device, dtype=dtype))
+
+    def test_a_larger_cut_size_gives_a_smaller_cut_4439(self, device, dtype):
+        side = []
+        for cut_size in ((0.1, 0.1), (0.9, 0.9)):
+            torch.manual_seed(0)
+            params = CutmixGenerator(cut_size=torch.tensor(cut_size, device=device, dtype=dtype), p=1.0)(
+                torch.Size([4, 3, 60, 80])
+            )
+            box = params["crop_src"][0, 0]
+            side.append(float(box[:, 0].max() - box[:, 0].min()))
+        assert side[0] > side[1]
 
     def test_random_gen(self, device, dtype):
         torch.manual_seed(42)


ducha-aiki (dismissed):
First review of head `443b4591`.

Thanks @dchaudhari7177, this is a clean fix for both halves of #4439. The docstring now states what the generator actually computes, and the guard catches the one configuration that turns the augmentation into a silent identity every time. I have no blocking findings. There are two optional notes below, and the branch needs a routine merge of `main` (the conflict is in `CHANGELOG.md` only).

**[P3] The guard covers `lambda = 1` exactly, but a minimum just below 1 can still cut nothing.** `random_generator/_2d/cutmix.py:93`. The cut is empty in a dimension whenever `sqrt(1 - lambda) * side < 1`, so a minimum above `1 - 1/side**2` still gives the inverted `(-1, -1)` box, `lam = 1.0`, and an unchanged image. I ran this on head:

```
cut_size=(0.99, 1.0)     on 6x8:   extents=[(-1, -1)] lam=[1.0, 1.0, 1.0, 1.0] image_changed=False
cut_size=(0.9999, 0.9999) on 64x64: extents=[(-1, -1)] lam=[1.0, 1.0, 1.0, 1.0] image_changed=False
```

The image size is only known in `forward`, so a constructor check can't catch these, and I think stopping at the endpoint is a reasonable scope. #3934 tracks the box arithmetic, and your PR body already explains why `p`-miss rows keep their zero-size boxes. Still, the new docstring sentence "a minimum of 1 is rejected because it would cut nothing" can be read as saying that any minimum below 1 cuts something. One more clause would fix that, e.g. "the cut is also empty whenever `sqrt(1 - lambda) * side < 1`, so a minimum close to 1 cuts nothing on small images." A related edge: the check runs in `make_samplers`, so dtype rounding shifts where it fires. `RandomCutMixV2(cut_size=(0.99999, 1.0))` constructs fine, but `.half()` then raises this `ValueError`, because `0.99999` rounds to `1.0` in float16. That's consistent (float16 would clamp `lambda` to 1 anyway), just worth knowing.

**[P3] The direction test could also pin the formula.** `test_a_larger_cut_size_gives_a_smaller_cut_4439` checks only that `0.1` cuts wider than `0.9`, and as you note it passes on `main`. Any monotone mapping would satisfy it. The docstring now promises `floor(sqrt(1 - lambda) * side)`, so one exact value would hold it to that. For example, `cut_size=(0.36, 0.36)` on `60x80` must give a `64x48` cut: vertex extents `63` and `47`, because `bbox_generator` puts the far corner at `start + size - 1`.

**Docstring math checked.** `CutmixGenerator.forward` computes `(sqrt(1 - lambda) * p_mask * side).floor()`: floor, not round or `int`. On `60x80` I compared the box size (vertex extent + 1) with `math.floor(math.sqrt(1 - lam) * side)` for 12 values of `lambda` in `[0, 0.99]`. Eleven match exactly. The only miss is `lambda = 0.99` (box `7x5`, exact math `8x6`), and that comes from float32 rounding: `1 - 0.99` lands just below `0.01`. The cut region in `RandomCutMixV2`'s output covers exactly `w * h` pixels (4200 / 2352 / 450 for `lambda = 0.1 / 0.5 / 0.9`). So the documented formula is what gets pasted.

**Behaviour change and release note.** Nothing that did useful work before is rejected now. `(1, 1)`, `[1.0, 1.0]` and `torch.tensor([1.0, 1.0])` all produced an identity on `main` (inverted box, `lam = 1.0`), and all now raise `ValueError` from both `CutmixGenerator(...)` and `RandomCutMixV2(...)` constructors, since `set_rng_device_and_dtype` runs `make_samplers` in `__init__`. `ValueError` matches `_joint_range_check` on the line above. The check runs in `make_samplers`, not in `forward`, so compiled forwards are unaffected. The CHANGELOG entry says what used to happen and what happens now, and it sits under `Bug fixes` alongside similar "now rejects" entries (`RandomRain`, #4461).

**Verified on this head**
- `tests/augmentation/test_random_generator.py -k CutMix`: 688 passed on CPU float32+float64. 344 passed on CPU float32 with torch 2.5.1. The two new tests pass on CPU float16 and bfloat16. With the known-failure profiles, `test_random_generator.py` + `test_augmentation_mix.py -k CutMix` gave 242 passed / 110 xfailed (float16) and 344 passed / 8 xfailed (bfloat16). The same selection on MPS float32: 352 passed.
- `tests/augmentation/test_augmentation_mix.py -k CutMix`: 15 passed on CPU float32+float64, and 8 passed on CPU float32 with `KORNIA_TEST_OPTIMIZER=inductor`.
- Base check: I copied the test diff onto `main` (`dc0af1a4`). `test_invalid_param_combinations[...cut_size7]` (both `same_on_batch`) and `test_a_minimum_cut_size_of_one_is_rejected_by_name_4439` fail there (3 failed, 341 passed), matching your PR body. The direction test passes on both, as stated.
- Doctests for both modules pass. `ruff@0.16.6` check and format are clean.
- `grep -rnE "#4439|issues/4439" kornia/ tests/` finds no hits in `kornia/` on head or on current `main`. The only hit is the new test comment, so no Convention block or wart pin needs updating.
- Overlap: open #4471 (bfloat16 mix support) touches `_2d/mix/cutmix.py` at `apply_transform_class` (~L119). This PR changes only the docstring at L48-54 of that file, so the hunks don't overlap.

ducha-aiki (dismissed):
Re-approving on `8d80f68e`. To clear the `CHANGELOG.md`-only conflict (`main` added entries at the top of the same `### Bug fixes` list), I pushed `8d80f68e`, a keep-both merge of `origin/main` (`dc0af1a4`) into the head `443b4591` I reviewed above. It is a pure merge: every path except `CHANGELOG.md` is tree-identical to `git merge-tree`, and the non-blank lines it adds to `CHANGELOG.md` are exactly this branch's own. The review above stands unchanged. If you push again, please pull this merge first.

PR #4492, opened 2026-09-13
fix(augmentation): clarify RandomBoxBlur flag semantics

### LINKED ISSUE #4433: W1b: `RandomBoxBlur`'s `normalized` flag is passed into `box_blur`'s `separable` argument

kv-248:

`RandomBoxBlur(normalized=False)` currently selects a non-separable blur, although its docstring says the flag controls kernel normalization. Both modes actually use L1-normalized kernels.

Correct the argument documentation and pass the existing flag explicitly as `separable=`, preserving the constructor name, per-call overrides, and numerical behavior. Add a regression that checks both per-call modes against the corresponding filter with exact equality.

Fixes #4433 using the behavior-preserving documentation/forwarding approach suggested in the issue.

## What did you check?

- New behavior-preservation cases passed on the original implementation (4 CPU cases).
- Focused `TestRandomBoxBlur` checks passed after the change: 6 cases, CPU float32/float64, Python 3.11.9 and PyTorch 2.8.0+cpu.
- Full `pre-commit run --all-files` and `git diff --check`.
- No CUDA or full test-suite run. Pixi was unavailable; checks used isolated Python environments.

## How did AI help?

GPT-5.6 Sol implemented the documentation, forwarding change, and tests. The coordinating agent reviewed source behavior and the diff, caught and corrected a flag-override issue, and ran the full pre-commit checks.


ducha-aiki (commented):
First review of head `af9c2ecc`. Thank you for picking up #4433. The change is correct, but it duplicates #4486, which fixes the same issue and was opened earlier (07:17 UTC vs 10:01 UTC today). We plan to merge #4486, so I recommend closing this one.

How the two PRs compare, so the choice is transparent:

- **Production change: equivalent.** Both PRs forward `flags["normalized"]` as `separable=` by keyword. This PR keeps `border_type` positional, like the `guided.py` callers, and #4486 names both keywords. Both keep the public `normalized` keyword, so neither is a breaking change, and outputs do not change.
- **Tests: equivalent in strength.** I mutated the forwarding line and ran `TestRandomBoxBlur` on CPU float16/bfloat16/float32/float64. Your `test_normalized_override_selects_separable` fails every mutation:
  - swap to `separable=not normalized`: 8 failed
  - drop the flag: 4 failed
  - always separable: 4 failed
  - return window sums when `normalized=False`: 4 failed
  - ignore `border_type`: 8 failed

  #4486's tests fail on all five as well. The test is unseeded, but the separable and 2D outputs differed bitwise in all 500 seeds I tried on its 2x3x7x9 / `(3,5)` input, so it will not flake. Your test also covers the per-call override path and a non-square kernel, which #4486's tests do not. #4486, for its part, adds an independent window-mean value pin for both settings.
- **Documentation: #4486 is more complete, and this decided it.** Your `Args` entry at `box_blur.py:35` is accurate. #4486's entry also says what callers need to know: both settings return the window mean, never the sum, and they differ only by float rounding. I measured that gap at up to about 1 ULP (for example 3.3e-7 in float32 and 4.9e-4 in float16, never bitwise equal). #4486 also links `kornia.filters.box_blur`, and its CHANGELOG entry quotes the old description next to what the flag actually does.

With equivalent code and equally strong tests, the documentation made the difference, and #4486 was also opened first.

**CI:** no test workflow has run on this PR yet. `kornia.kornia` is `ACTION_REQUIRED` and `pre-commit.ci` shows `ERROR`. The PR is also 25 commits behind `main` and has a CHANGELOG-only conflict.

**Verified on this head:** `TestRandomBoxBlur` passes: 6 on CPU float32/float64, 3 on CPU float16 and 3 on CPU bfloat16 (each with its known-failure profile), and 6 on MPS float32/float16. The module doctest passes, ruff 0.16.6 check/format is clean, and the new test passes on `main` (12 across the four CPU dtypes, as expected for a behaviour-preserving change). You can see the mutation runs above. `grep -rnE "#4433|issues/4433" kornia/ tests/` is empty.

Issue #4441, opened 2026-09-11
W1b: `PolicySequential` bypasses `OperationBase.forward_parameters`, so one policy entry has two meanings


ducha-aiki:
The `auto` sub-package describes a policy entry as `(op_name, min_mag, max_mag)` and every composer takes the same
tuple. Two of them do not interpret it the same way. `RandAugment` draws its parameters through
`OperationBase.forward_parameters`, which applies the operation's own `_magnitude_fn` — for `ShearX` that is
`_process_magnitude`, `magnitude * 180`, converting the policy's shear fraction into the degrees
`RandomShear` expects. `AutoAugment` and `TrivialAugment` route through `PolicySequential`, whose
`forward_parameters` calls `module.op.forward_parameters(batch_shape)` at
`kornia/augmentation/auto/operations/policy.py:157` — `module.op` is the wrapped `RandomShear`, so the call reaches
neither the `mag` override nor `_magnitude_fn`.

The same entry `("shear_x", -0.3, 0.3)` therefore produces a shear of about 52 degrees under `RandAugment` and
about 0.2 degrees under `TrivialAugment`, a factor of 180 — `TrivialAugment`'s shear is a visually undetectable
no-op. The second half of the same bypass is `_update_probability_gen`: the per-operation `_probability` that
`OperationBase` maintains is never consulted on the `PolicySequential` path either, which is why those entries
survive into `state_dict()` and change nothing when loaded (filed separately in this batch).

## Repro

```python
import torch
import kornia.augmentation.auto as AA
from kornia.augmentation.auto.operations import ops


POLICY = [[("shear_x", -0.3, 0.3)]]
torch.manual_seed(0)
x = torch.rand(8, 3, 16, 16)

torch.manual_seed(0)
ra = AA.RandAugment(n=1, m=29, policy=POLICY)
ra(x)
print("RandAugment(n=1, m=29)  shear_x =",
      [round(v, 4) for v in ra._params[0].data[0].data["shear_x"].flatten().tolist()])

torch.manual_seed(0)
ta = AA.TrivialAugment(policy=POLICY)
ta(x)
print("TrivialAugment          shear_x =",
      [round(v, 4) for v in ta._params[0].data[0].data["shear_x"].flatten().tolist()])

# the same op object, the two parameter paths side by side
op = ops.Rotate(initial_magnitude=90.0, initial_probability=1.0, magnitude_range=(0.0, 90.0), symmetric_megnitude=False)
torch.manual_seed(0)
[...]

PR #2181, opened 2023-01-26
[Feat] Initiate AutoAugment modules

shijianjian:
Planning on integrating AutoAugment-ish modules into the library.

Planning for this PR:

- General ideas:
  - [x] Each operation shall be differentiate w.r.t. magnitude and probability.
       - NOTE: Apart from `Equalize`.

- Grab and use policies:
  - [x] AutoAugment
  - [x] RandAugment
  - [x] TrivialAugment

- New Augmentations:
  - [x] RandomShear
  - [x] RandomTranslate
  - [x] RandomAutoContrast 

The upcoming feature:

  - [x] Support multi-data key transformation.
        - Will make `AugmentationSequential` consume `PolicyAugmentBase`.

- Augmentation search:
  - [ ] AugNet https://github.com/cedricrommel/augnet
  - [ ] Faster AutoAugment https://github.com/moskomule/dda/tree/master/faster_autoaugment
  - [ ] Augerino https://github.com/g-benton/learning-invariances/tree/master/augerino

PR #4495, opened 2026-09-13
fix(auto): symmetric magnitudes must negate, not zero

### LINKED ISSUE #4440: W1b: `OperationBase`'s symmetric magnitude multiplies by a bool, so half of every `RandAugment` batch gets 0
`kornia.augmentation.auto.rand_augment.ops` builds `rotate`, `shear_x`, `shear_y`, `translate_x` and `translate_y`
with `symmetric_megnitude=True`, and each of those factories rejects an asymmetric range first
(`if min_mag != -max_mag: raise ValueError(...)`, `auto/rand_augment/ops.py:76-83` for `rotate`), so a `RandAugment`
policy entry written `("rotate", -30.0, 30.0)` promises rotation in both directions.

It never rotates in the negative direction. `OperationBase._init_magnitude_fn`'s `_random_flip`
(`auto/operations/base.py:82-86`) draws `flip = torch.rand((x.shape[0],), device=x.device) > 0.5` — a **bool**
tensor — and returns `fn(x) * flip`. Multiplying by a bool keeps the magnitude where `flip` is true and sets it to
**zero** where it is false, instead of negating it. So those five operations draw from `{0, +mag}` rather than
`{-mag, +mag}`: the negative half of the advertised range is unreachable, and roughly half of every batch receives
magnitude zero, which for a geometric operation is the identity. At `B = 256` the zero fraction is close to one
half and reads as a weaker augmentation; at smaller batch sizes it is a lottery, and a small batch can come out
entirely unaugmented for that operation.

## Repro

```python
import torch
from kornia.augmentation.auto.operations.base import OperationBase
CASES = [("rotate", -30.0, 30.0, "degrees"), ("shear_x", -0.3, 0.3, "shear_x"), ("shear_y", -0.3, 0.3, "shear_y"),
         ("translate_x", -0.5, 0.5, "translate_x"), ("translate_y", -0.5, 0.5, "translate_y")]

def drawn(name, lo, hi, key, m, B, seed):
    torch.manual_seed(seed)
    aug = AA.RandAugment(n=1, m=m, policy=[[(name, lo, hi)]])
    aug(torch.rand(B, 3, 16, 16))
    return aug._params[0].data[0].data[key].flatten()

for B, m, seed in ((256, 29, 7), (64, 1, 7)):
    print(f"--- B={B}, m={m}, seed={seed} ---")
    for name, lo, hi, key in CASES:
        v = drawn(name, lo, hi, key, m, B, seed)
        print(f"{name:12s} unique={[round(u, 4) for u in v.unique().tolist()]}"
              f"  zero_frac={(v == 0).float().mean():.3f}  neg_frac={(v < 0).float().mean():.3f}")

print("--- a whole small batch can be the identity: m=15, B=2 ---")
for seed in (7, 0, 3):
    v = drawn("rotate", -30.0, 30.0, "degrees", 15, 2, seed)
    print(f"seed={seed}  degrees={[round(u, 4) for u in v.tolist()]}  zero_frac={(v == 0).float().mean():.3f}")

print("--- which of the three policy classes reaches _random_flip ---")
torch.manual_seed(7); x = torch.rand(64, 3, 16, 16)
for label, aug in (("RandAugment", AA.RandAugment(n=1, m=29, policy=[[("rotate", -30.0, 30.0)]])),
                   ("TrivialAugment", AA.TrivialAugment(policy=[[("rotate", -30.0, 30.0)]])),
                   ("AutoAugment bin=0", AA.AutoAugment(policy=[[("rotate", 1.0, 0)]])),
                   ("AutoAugment bin=9", AA.AutoAugment(policy=[[("rotate", 1.0, 9)]]))):
    aug(x)
    d = aug._params[0].data[0].data["degrees"]
    flags = [m.symmetric_megnitude for m in aug.modules() if isinstance(m, OperationBase)]
    print(f"{label:18s} symmetric_megnitude={flags}  degrees min={d.min():.4f} max={d.max():.4f}"
          f"  zero_frac={(d == 0).float().mean():.3f}")
```

## Observed

```
--- B=256, m=29, seed=7 ---
rotate       unique=[0.0, 29.0]  zero_frac=0.484  neg_frac=0.000
shear_x      unique=[0.0, 52.2]  zero_frac=0.465  neg_frac=0.000
shear_y      unique=[0.0, 52.2]  zero_frac=0.465  neg_frac=0.000
translate_x  unique=[0.0, 0.4833]  zero_frac=0.484  neg_frac=0.000
translate_y  unique=[0.0, 0.4833]  zero_frac=0.484  neg_frac=0.000
--- B=64, m=1, seed=7 ---
rotate       unique=[0.0, 1.0]  zero_frac=0.516  neg_frac=0.000
shear_x      unique=[0.0, 1.8]  zero_frac=0.500  neg_frac=0.000
shear_y      unique=[0.0, 1.8]  zero_frac=0.500  neg_frac=0.000
translate_x  unique=[0.0, 0.0167]  zero_frac=0.516  neg_frac=0.000
translate_y  unique=[0.0, 0.0167]  zero_frac=0.516  neg_frac=0.000
--- a whole small batch can be the identity: m=15, B=2 ---
seed=7  degrees=[0.0, 0.0]  zero_frac=1.000
seed=0  degrees=[15.0, 0.0]  zero_frac=0.500
seed=3  degrees=[0.0, 0.0]  zero_frac=1.000
--- which of the three policy classes reaches _random_flip ---
RandAugment        symmetric_megnitude=[True]  degrees min=0.0000 max=29.0000  zero_frac=0.516
TrivialAugment     symmetric_megnitude=[True]  degrees min=0.1085 max=29.6658  zero_frac=0.000
AutoAugment bin=0  symmetric_megnitude=[False]  degrees min=-29.9259 max=-24.0319  zero_frac=0.000
AutoAugment bin=9  symmetric_megnitude=[False]  degrees min=24.0179 max=29.8717  zero_frac=0.000
```

## Expected

`unique = [-mag, 0.0, mag]` at most, with `neg_frac` near `0.5` and `zero_frac` near zero — the symmetric range the
factories validate for. The shape of the fix is to make `flip` a sign rather than a mask, for instance
`torch.where(torch.rand(...) > 0.5, 1.0, -1.0)`, so `fn(x) * flip` negates.

The blast radius is `RandAugment` alone, which the last block of the repro settles by execution over the three
policy classes. Only `RandAugment` both builds its operations with `symmetric_megnitude=True` and draws them
through `OperationBase.forward_parameters`, the path that applies `_random_flip`. `TrivialAugment` reuses the same
`rand_augment.ops` factories so the flag is `True` there too, but draws through `PolicySequential` and never
reaches `_random_flip`; its one-sided magnitudes have a different cause, filed separately in this batch.
`AutoAugment` builds every geometric operation with `symmetric_megnitude=False`
(`auto/autoaugment/ops.py:48, 59, 70, 81, 92`) and reaches negative magnitudes through signed bins: bin `0` of
`rotate` draws inside `[-30, -24]`. Fixing `_random_flip` leaves both of them byte-identical, and changes every
current `RandAugment` result — strictly more augmentation, in the opposite direction half the time — so it is a
behavior change. `Rotate`'s own guard (`symmetric_megnitude and magnitude_range[0] < 0` raises) assumes the range
passed down is the non-negative half and stays correct under the fix.

## Classification

**Broken output** — five operations that validate for a symmetric range produce a one-sided one, and about half of
every batch is an identity for them. Idiom: 3 (a guard whose predicate does not match the documented contract —
the factories validate `-min_mag == max_mag` for a symmetry the magnitude function does not implement).

## Disposition

**MERGE IN WINDOW — behavior change.** Canonical: `_random_flip` in `kornia/augmentation/auto/operations/base.py`,
where all five `rand_augment.ops` operations get their sign. Every `RandAugment` result changes, so the decision
belongs in the coordinated Repair & Deprecation window rather than in a drive-by PR. Batch 6 documents the current
one-sided draw on `OperationBase` and pins it, and does not change it.

## Environment

change it.

### LINKED ISSUE #4494: [Bug]: symmetric_megnitude zeroes the magnitude instead of negating it, so Rotate/Shear/Translate never go negative
### What happened?

`OperationBase._init_magnitude_fn` multiplies the sampled magnitude by a **boolean** mask instead of a sign, so `symmetric_megnitude=True` zeroes half the magnitudes rather than negating them.

https://github.com/kornia/kornia/blob/main/kornia/augmentation/auto/operations/base.py#L82-L87

```python
def _random_flip(fn):
    def f(x):
        flip = torch.rand((x.shape[0],), device=x.device) > 0.5
        return fn(x) * flip          # <-- bool, so this is `m` or `0`, never `-m`
    return f
```

`flip` is a `torch.bool` tensor, so `fn(x) * flip` promotes to `m * 1` or `m * 0`. The intended `-m` branch never happens.

Two consequences:

1. **The augmentations are one-sided.** `Rotate`, `ShearX`, `ShearY`, `TranslateX` and `TranslateY` all default to `symmetric_megnitude=True` and are the only ops that use it. Images are only ever rotated clockwise, only ever sheared and translated in the positive direction. That is not AutoAugment/RandAugment/TrivialAugment as specified, and it halves the geometric diversity of every policy that draws one of these five.
2. **Roughly half of all draws are silently a no-op.** A zero magnitude means the operation runs but does nothing, so a policy that picks `rotate` leaves the image unchanged about half the time.

The docstring says the opposite of what the code does — `symmetric_megnitude: if to randomly assign the magnitude as negative or not` (`operations/ops.py:414`) — and the callers are built around that reading: `Rotate.__init__` rejects `magnitude_range[0] < 0` when symmetric, and `rand_augment.ops.rotate` requires `min_mag == -max_mag` then passes `(0.0, max_mag)`. Both only make sense if the sign is applied afterwards.

**Expected:** magnitudes distributed symmetrically about zero, with no spike at zero.

### Minimal reproduction

```python
import torch
from kornia.augmentation.auto.operations import ops

torch.manual_seed(0)
for cls, factor in [(ops.Rotate, "degrees"), (ops.ShearX, "shear_x"), (ops.TranslateX, "translate_x")]:
    op = cls(None, 1.0)  # symmetric_megnitude=True by default
    m = torch.cat([op.forward_parameters(torch.Size([64, 3, 8, 8]))[factor] for _ in range(40)])
    print(f"{cls.__name__:10s} neg={(m < 0).sum().item():4d} zero={(m == 0).sum().item():4d} pos={(m > 0).sum().item():4d}")
```

Actual, 2560 draws each:

```
Rotate     neg=   0 zero=1258 pos=1302     # min=0.000  max=29.975, never below 0
ShearX     neg=   0 zero=1238 pos=1322
TranslateX neg=   0 zero=1259 pos=1301
```

Expected: `neg` and `pos` each near half, `zero` near 0.

### Environment

- Kornia version: 0.9.0rc1 (also present on `main`)
- PyTorch version: 2.14.0+cpu
- Python version: 3.12.7
- OS / device (CPU, CUDA, ROCm, MPS): Windows 11, CPU

### Anything else?

Applying a sign instead of a mask gives the documented distribution:

```python
sign = torch.where(torch.rand((x.shape[0],), device=x.device) > 0.5, 1.0, -1.0).to(x.dtype)
return fn(x) * sign
```

With that change, same script and seed:

```
Rotate     neg=1258 zero=   0 pos=1302     # min=-29.995  max=29.975
ShearX     neg=1238 zero=   0 pos=1322
TranslateX neg=1259 zero=   0 pos=1301
```

The explicit `.to(x.dtype)` matters — `torch.where` on Python floats produces `float32` and would otherwise silently upcast a half-precision magnitude.

Note this is a behaviour change for anyone with a seeded pipeline, since the sampled magnitudes move.

Found while writing the geometry-only inverse round-trip test in #4456 — the zero-magnitude draws made a share of that test's assertions vacuous. Not fixed there, it is unrelated to that PR's guard. Happy to send a PR with a regression test on the magnitude distribution if that is wanted.

mohansree14:
Closes #4494.
Closes #4440.

### What

`OperationBase._init_magnitude_fn._random_flip` multiplies the sampled magnitude by a **`torch.bool`** mask:

```python
flip = torch.rand((x.shape[0],), device=x.device) > 0.5
return fn(x) * flip          # `m` or `0` -- never `-m`
```

The boolean promotes to `1` or `0`, so the intended `-m` branch never happens.

`symmetric_megnitude=True` is set by `Rotate`, `ShearX`, `ShearY`, `TranslateX` and `TranslateY`, the only five ops that use the flag. All of them therefore ran in the positive direction only — only counter-clockwise, only right, only down — and were a silent no-op on roughly half their draws. The multiply-by-bool has been there since the ops were added in #2181, so these ops never drew a negative magnitude.

**Who this reaches.** The five ops used directly (`symmetric_megnitude=True` is their default) and `RandAugment`, which samples through `OperationBase` with `op.forward_parameters(batch_shape, mag=mag)`. `AutoAugment` builds the five ops with `symmetric_megnitude=False` and gets its negative values from signed magnitude bins, so it is unaffected. `TrivialAugment` samples through `PolicySequential.forward_parameters`, which bypasses `OperationBase._magnitude_fn`, so it is unaffected here and is still one-sided; that is tracked separately in #4441.

### Why this is the intended behaviour, not a deliberate gate

- The argument documents it: `symmetric_megnitude: if to randomly assign the magnitude as negative or not` (`operations/ops.py`).
- `Rotate.__init__` raises when `magnitude_range[0] < 0` and `symmetric_megnitude` is set — a lower bound that must be non-negative only makes sense if the sign is applied afterwards.
- `rand_augment.ops.rotate` requires `min_mag == -max_mag`, then passes `(0.0, max_mag)` through. The symmetric half of the range is dropped on the assumption that something restores it.

A separate probability gate already exists (`batch_prob` in `OperationBase.forward`), so the zeroing is not serving that purpose either.

### The fix

```python
sign = torch.where(torch.rand((x.shape[0],), device=x.device) > 0.5, 1.0, -1.0)
return fn(x) * sign.to(x.dtype)
```

The explicit `.to(x.dtype)` matters: `torch.where` on Python float scalars produces `float32`, which would silently upcast a float16 or bfloat16 magnitude — the same promotion #4210 fixed in `OperationBase.forward`.

### Evidence

2560 draws per op, before and after, same seed:

| op | before (neg / zero / pos) | after (neg / zero / pos) |
|---|---|---|
| `Rotate` | 0 / 1258 / 1302 | 1258 / 0 / 1302 |
| `ShearX` | 0 / 1238 / 1322 | 1238 / 0 / 1322 |
| `TranslateX` | 0 / 1259 / 1301 | 1259 / 0 / 1301 |

`Rotate`'s observed range goes from `[0.000, 29.975]` to `[-29.995, 29.975]` against its `(0.0, 30.0)` magnitude range.

### Test

`test_symmetric_magnitude_negates_rather_than_zeroing` covers all five ops: over 512 seeded draws each it asserts both signs appear and that no draw was zeroed. It fails on all five parametrisations without the fix (`AssertionError: degrees: no negative magnitude in 512 draws`) and passes with it.

### Validation

- `tests/augmentation/` — 2519 passed, 3399 skipped, 24 xfailed, 5 xpassed. CPU float32 only; no CUDA or MPS run locally.
- `ruff check`, `ruff format --check` clean; `mypy` reports nothing new in the changed file.

### Compatibility

This changes which magnitudes a **seeded** pipeline draws, but narrowly: the RNG stream, the ops picked and the rows that already drew `+m` are bit-identical, and the next `torch.rand` after sampling matches. Only the rows that used to get `0` now get `-m`. The distribution moves to the documented one, and the operations were not doing what the policies specify before, so I read that as the point of the fix rather than a cost of it. Flagging it for the release note either way.

Release note: `changelog.d/4495.fixed.md`.


ducha-aiki (changes requested):
First review of head `8391a86d`. Thanks, this is the right fix: a sign instead of a bool mask, one `torch.rand` per call as before, and the new test fails on `main` for all five ops. The code change is good to go. The two items below are about what the release note and the PR say the change reaches.

**[P2] The release note names `AutoAugment` and `TrivialAugment`, but neither goes through `_random_flip`, so neither changes.** `CHANGELOG.md:395-399` says the bug affected "every `AutoAugment`, `RandAugment` and `TrivialAugment` policy that drew one of them", and that seeded magnitudes change. The PR body says the same. `RandAugment` is the only composer that reaches the flip, because it calls `op.forward_parameters(batch_shape, mag=mag)` (`auto/rand_augment/rand_augment.py:169`).
- `AutoAugment` builds all five geometric ops with `symmetric_megnitude=False` (`auto/autoaugment/ops.py:75` and following). It gets its negative values from signed magnitude bins.
- `TrivialAugment` reuses the `rand_augment.ops` factories, so the flag is `True` there. But it samples through `PolicySequential.forward_parameters`, which calls `module.op.forward_parameters(batch_shape)` (`auto/operations/policy.py:157`). That skips `OperationBase` completely: no `mag`, no `_magnitude_fn`, no flip. So `TrivialAugment` is still one-sided on this head. That is the separate bug #4441.

I sampled 300 policy draws at `B=8` with seed 0, on this head and on `main`:

```
                 head (neg / zero / pos)                       main (neg / zero / pos)
TrivialAugment   Rotate 0/0/224, ShearX 0/0/328, TranslateX 0/0/160   identical
AutoAugment      Rotate 104/0/432, ShearX 0/0/64                      identical
RandAugment      Rotate 157/0/131, ShearX 195/0/189                   Rotate 0/157/131, ShearX 0/195/189
```

Thirty seeded forward passes of the default `AutoAugment()` and `TrivialAugment()` give the same output checksums on both trees. Only `RandAugment` differs (38925.32 vs 39783.46). Could you reword the entry so it names `RandAugment` and the five ops used directly (`symmetric_megnitude=True` is their default)? It could also say `TrivialAugment` is unaffected and still tracked in #4441. Two smaller wording points:
- Drop "again". The multiply-by-bool has been there since the ops were added in #2181, so these ops never drew a negative magnitude.
- Say exactly what moves for a seeded pipeline. The RNG stream, the ops picked and the rows that already had `+m` are all bit-identical. The next `torch.rand` after sampling matches on both trees for all five ops. Only the rows that used to get `0` now get `-m`.

**[P2] Please also close #4440.** #4440 reports the same bug, `_random_flip` multiplying by a bool, with the same suggested fix. It was filed two days before #4494 and its repro isolates `RandAugment` as the only affected composer. Right now only #4494 is in `closingIssuesReferences`, so #4440 would stay open after merge. Adding `Closes #4440` to the body is enough. Neither #4440 nor #4494 is cited anywhere in `kornia/` or `tests/` on this head or on `main`, so there is no pin or Convention block to update.

**[P3] Rotation direction in the test comment.** `tests/augmentation/test_auto_operation.py:274` (and the PR body) says the ops ran "only clockwise". In kornia, positive `degrees` rotate counter-clockwise on screen. With `Rotate` at `degrees=+45` on a 21×21 image, a bright pixel at (row 10, col 18) lands at (4, 16): it moves up, not down. So the old behaviour was "only counter-clockwise". "Only right, only down" is correct for the translations: `+3` px moves the pixel to col 20 and to row 13.

**Heads-up on #4456** (same author, both open): both PRs append a test at the end of `tests/augmentation/test_auto_operation.py` and add an entry at the top of `### Bug fixes`. Whichever merges second will get a textual conflict in both files, and keeping both sides resolves it. I applied #4456's `auto/base.py` and a keep-both merge of its tests onto this head. Everything passes: 52 tests each on CPU float32 and float64, 50 passed and 2 skipped on MPS float32. #4456's geometry-only round trip now sees negative `RandAugment` translations too, and still holds.

**Verified on this head**
- `tests/augmentation/test_auto_operation.py`:
  - CPU float32+float64: 65 passed.
  - CPU float16 (`--xfail-known-failures`, `cpu-float16`): 47 passed, 1 xfailed.
  - CPU bfloat16: 48 passed.
  - MPS float32: 46 passed, 2 skipped.
  - `KORNIA_TEST_OPTIMIZER=inductor`: 48 passed.
  - torch 2.5.1: 48 passed.
  - Doctests under `kornia/augmentation/auto`: 3 passed.
- I copied the new test onto `main`, where it fails on all five parametrisations (`degrees: no negative magnitude in 512 draws`, and the same for the other four).
- Distribution, 2560 draws per op:
  - `Rotate`: 1258/0/1302 on this head vs 0/1258/1302 on `main`, range `[-29.995, 29.975]`.
  - `ShearX` and `ShearY`: 1310/0/1250.
  - `TranslateX` and `TranslateY`: 1258/0/1302.
- Flip dtype: `_magnitude_fn` keeps float16, bfloat16, float32 and float64 on CPU (torch 2.14.0 and 2.5.1), and float16, bfloat16 and float32 on MPS.
- Learned magnitude: the gradient of `Rotate(15.0)`'s `_magnitude` is now (#pos − #neg) (−4 for 2 positive and 6 negative rows), not #pos, so it flows through the sign as expected.
- `same_on_batch`: all five ops hard-code `same_on_batch=False`, and the sign was already drawn per row, so batch semantics do not change.
- `ruff@0.16.6 check` and `format --diff` are clean on the changed files.

CI was still running when I checked: 27 green and 12 in progress or queued, none red.

ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (283606a2) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4495.fixed.md`.

ducha-aiki (changes requested):
Review of `283606a28c838412ae53fe299c7f0c805a4fa629`.

Re-review of `283606a2`: the merge only moved the release note to Towncrier, so the previous content findings remain.

[P2] `changelog.d/4495.fixed.md:1-5` still says this changes AutoAugment and TrivialAugment. It does not: AutoAugment constructs the affected operations with `symmetric_megnitude=False` (`kornia/augmentation/auto/autoaugment/ops.py:75` and following), while TrivialAugment samples through `PolicySequential.forward_parameters` (`kornia/augmentation/auto/operations/policy.py:157`) and bypasses `OperationBase._magnitude_fn`. Only direct uses of the five operations and RandAugment reach this fix. Update the fragment and PR body accordingly; TrivialAugment remains tracked separately in #4441.

[P2] The PR body still closes only #4494. Add `Closes #4440`, which reports the same `_random_flip` bug and would otherwise remain open.

[P3] `tests/augmentation/test_auto_operation.py:274` still says the prior positive rotations were clockwise; positive Kornia rotations are counter-clockwise. Correct the comment while updating the release prose.

mohansree14:
Thanks for the careful re-review, and for moving the entry to a fragment. All three items are addressed in `8cb54dd8`.

**[P2] Release-note scope.** `changelog.d/4495.fixed.md` now names the five ops used directly and `RandAugment` as the only paths the fix reaches. It says `AutoAugment` (built with `symmetric_megnitude=False`) and `TrivialAugment` (samples through `PolicySequential.forward_parameters`) are unaffected, with `TrivialAugment` still tracked in #4441. "Again" is gone, and it states what moves for a seeded pipeline: the RNG stream, the ops picked and the rows that already drew `+m` are unchanged, and only the rows that drew `0` now draw `-m`. The PR body is updated the same way.

**[P2] #4440.** Added `Closes #4440` to the body. It now shows up in the closing references next to #4494.

**[P3] Rotation direction.** The comment at `tests/augmentation/test_auto_operation.py:274` now says counter-clockwise, and it notes that `RandAugment` is the only composer that samples through the flag. The PR body is corrected too.

No code change in this push: only the fragment, a test comment and the PR body.

ducha-aiki (dismissed):
Review of `8cb54dd833ff8b7030b158875b4d0f61f6392aa7`.

This push changes only `changelog.d/4495.fixed.md`, one test comment and the PR body. The net production patch in `kornia/` matches `8391a86d` line for line, the head where the code was already verified. All three items from review 5196222134 are resolved, and I checked every factual claim in the new fragment against both trees. Approving.

**Prior items**

- [P2] Fragment scope: resolved. The fragment now says the fix reaches the five ops used directly and `RandAugment`, and that `AutoAugment` and `TrivialAugment` do not change. I ran the same seeded probe in the PR worktree and the main worktree: 300 `forward_parameters` draws at `B=8` with seed 0, counting neg/zero/pos per parameter key across the policy's ops, then the next 3 `torch.rand` values, then 30 seeded forward passes on a fixed `4x3x32x32` input.

  ```
                     head                                          main
  AutoAugment()      degrees 104/0/432  sum=9705.519145            identical, tail rand identical
                     fwd checksum 181760.961566                    181760.961566
  TrivialAugment()   degrees 0/0/224  translate_x 0/176/160        identical, tail rand identical
                     fwd checksum 172837.637991                    172837.637991
  RandAugment(2,10)  degrees 157/0/131  shear_x 195/320/189        degrees 0/157/131  shear_x 0/515/189
                     translate_x 174/288/162                       translate_x 0/462/162
                     tail rand [0.63706, 0.613156, 0.664355]       identical
                     fwd checksum 159618.737440                    163406.454200
  ```

  `AutoAugment` and `TrivialAugment` are byte-identical: the counts, the sums, the RNG tail and the forward checksums all match. For `RandAugment`, the positive counts and the RNG tail are identical on both trees. The only change is that the rows that drew `0` now draw `-m`, for example `degrees` 157 zeros become 157 negatives. The zeros left under `shear_x` on head come from the other ops' `shear_x = 0` rows. That confirms the fragment's sentence "the random stream, the ops picked and the rows that already drew `+m` are unchanged; only the rows that drew `0` now draw `-m`". For the direct ops, all five default to `symmetric_megnitude=True` (`kornia/augmentation/auto/operations/ops.py:424, 463, 503, 543, 577`), so "which is their default" holds. With seed 0 and 40 × `B=64` draws, `Rotate` gives 1258/0/1302 on head and 0/1258/1302 on main. `ShearX` and `ShearY` give 1310/0/1250 on head and 0/1310/1250 on main. The translations give 1258/0/1302. The next `torch.rand` matches on both trees for every op. The code paths agree with this. `AutoAugment` passes `symmetric_megnitude=False` (`kornia/augmentation/auto/autoaugment/ops.py:75-119`). `TrivialAugment` goes through `PolicySequential.forward_parameters`, which calls `module.op.forward_parameters` (`kornia/augmentation/auto/operations/policy.py:157`) and never reaches `_magnitude_fn`. `RandAugment` calls `op.forward_parameters(batch_shape, mag=mag)` (`kornia/augmentation/auto/rand_augment/rand_augment.py:169`).
- [P2] `Closes #4440`: resolved. GraphQL `closingIssuesReferences` now returns exactly #4440 and #4494, and this PR fully fixes both. #4440's stated blast radius is `RandAugment` alone, and its expected output is `neg_frac≈0.5, zero_frac≈0`. #4441, the one-sided `TrivialAugment` bug, is correctly left out and is only cited in the fragment. `git grep -nE "#4440|issues/4440|#4494|issues/4494|#4441|issues/4441"` over the merge tree (`git merge-tree --write-tree origin/main` with this head) has no hits in `kornia/`, `tests/` or `docs/`, so there is no documented-defect prose or pin to update.
- [P3] Rotation direction: resolved. `tests/augmentation/test_auto_operation.py:274` now says "only counter-clockwise". The new sentence "`RandAugment` is the only composer that samples through it" matches the call sites listed above.

**Whole production diff re-read** (`kornia/augmentation/auto/operations/base.py:82-87`): each call still makes exactly one `torch.rand((B,), device=x.device)`. `torch.where(..., 1.0, -1.0)` is cast with `.to(x.dtype)` before the multiply, so half-precision magnitudes are not upcast. The numeric literals are the sign values ±1.0 and the 0.5 fair-coin threshold, which was already there. Nothing else changes.

**What I ran**

- Environment: `.venv` python 3.11, torch 2.14.0. In every probe, `kornia.__file__` named the tree under test: the PR worktree or the main worktree.
- `pytest tests/augmentation/test_auto_operation.py --device=cpu --dtype=float32,float64` on head: 65 passed.
- Plant-and-revert: I replaced `kornia/augmentation/auto/operations/base.py` with main's version in the PR worktree. `test_symmetric_magnitude_negates_rather_than_zeroing` then failed on all 5 parametrisations (`degrees: no negative magnitude in 512 draws`, and the same for `shear_x`, `shear_y`, `translate_x` and `translate_y`). The file is restored and the worktree is clean.
- The seeded policy and direct-op sampling probe above, run identically in both trees.
- I did not rerun the float16/bfloat16, MPS, inductor and torch 2.5.1 checks on this push, because the production patch is unchanged since `8391a86d`, where they all passed.
- Merge: `git merge-tree` against main 53866348 is clean. Against open #4456 (`525c5a6a`) there is a content conflict in `tests/augmentation/test_auto_operation.py`: both PRs append a test at the end of the file. Keeping both sides resolves it, as noted in the first review.

ducha-aiki (approved):
Review of `c74bf65b2d2cec47b1270936fd29f74325b27ad6`.

Re-approval after the merge with `main` that I pushed to this branch. Merged #4456 also appended a test to the end of `tests/augmentation/test_auto_operation.py`, so I kept both tests. The merge's net diff against `main` is exactly this PR's own patch; the only differences are blank-line context positions.

On the merge commit, torch 2.14.0, CPU: `tests/augmentation/test_auto_operation.py` float32/float64 75 passed; float16 (`cpu-float16` profile) 52 passed, 1 xfailed; bfloat16 (`cpu-bfloat16`) 53 passed. `ruff@0.16.7` check and format are clean. The earlier approval of the content stands.

PR #4505, opened 2026-09-14
fix(augmentation): dispatch mix annotation keys in AugmentationSequential (#4493)

### LINKED ISSUE #4493: W1b: AugmentationSequential silently passes mask, box and keypoint keys through mix children
`AugmentationSequential` mixes the image through a mix child and hands every other data key back unchanged, so masks, boxes and keypoints silently desynchronize from the mixed image. The same keys raise `NotImplementedError` when the mix class is called directly, and `RandomMosaic` transforms boxes directly, so the container both hides an error and drops a working handler.

## Observed (`main` `a7e2db6b`, torch 2.14.0, CPU)

| child | key | direct call | inside `AugmentationSequential` |
|---|---|---|---|
| `RandomMixUpV2`, `RandomCutMixV2`, `RandomJigsaw`, `PatchMix` | `mask`, `bbox_xyxy`, `keypoints` | `NotImplementedError` | returned unchanged, image mixed |
| `RandomMosaic` | `bbox_xyxy` | boxes transformed | returned unchanged, image mixed |
| `RandomMosaic` | `mask`, `keypoints` | `NotImplementedError` | returned unchanged |
| `RandomMixUpV2`, `RandomCutMixV2` | `class` | labels returned | `NotImplementedError` from the container |

Only `RandomTransplantation` has a mask branch in the container.

## Mechanism

`MaskSequentialOps.transform`, `BoxSequentialOps.transform` and `KeypointSequentialOps.transform` in `kornia/augmentation/container/ops.py` dispatch on `GeometricAugmentationBase2D`, `GeometricAugmentationBase3D`, `_AugmentationBase`, the containers and `OperationBase`, with a `RandomTransplantation` special case for masks. A `MixAugmentationBaseV2` child matches none of them and reaches the trailing `return input`. `ClassSequentialOps.transform` is the only handler that checks for the mix base, and it raises.

## Repro

```python

torch.manual_seed(0)
x = torch.rand(2, 3, 16, 16)
boxes = torch.tensor([[[0.0, 0.0, 2.0, 2.0]], [[1.0, 1.0, 3.0, 3.0]]])
mask = torch.arange(2.0).reshape(2, 1, 1, 1).expand(2, 1, 16, 16).clone()

out = K.AugmentationSequential(K.RandomMosaic(p=1.0), data_keys=["input", "bbox_xyxy", "mask"])(x, boxes, mask)
print(torch.equal(out[0], x), torch.equal(out[1], boxes), torch.equal(out[2], mask))  # False True True

print(torch.equal(K.RandomMosaic(p=1.0, data_keys=["input", "bbox_xyxy"])(x, boxes)[1], boxes))  # False
K.RandomMixUpV2(p=1.0)(x, mask, data_keys=["input", "mask"])  # NotImplementedError
```

## Expected

Either the container raises for a key the mix child cannot transform, as the direct call does, or it dispatches to the child's own handler where one exists (`RandomMosaic` boxes). Silent passthrough of an annotation next to a mixed image is the same failure shape as #4420 for non-rigid children.

## Docs

The `AugmentationSequential` docstring note said mix augmentations "can only be working with input/image data key"; #4452 rewrites it to state the passthrough and link here. Batch 6d (#4407) covers the mix classes' own data-key contracts.

Yi-111-a:
## Summary

Fixes #4493.

`MaskSequentialOps`, `BoxSequentialOps` and `KeypointSequentialOps` had no branch for `MixAugmentationBaseV2`, so a mix child inside `AugmentationSequential` mixed the image and silently returned mask / box / keypoint keys unchanged. Direct calls raise `NotImplementedError` for unsupported keys, and `RandomMosaic` already transforms boxes — the container hid both.

This change dispatches those keys to the mix child's own handlers (`transform_mask` / `transform_boxes` / `transform_keypoint`) with the same params as the image step:

- supported keys work (e.g. `RandomMosaic` boxes)
- unsupported keys raise `NotImplementedError`, matching a direct call
- `RandomTransplantation` mask handling is unchanged (still matched first)
- `class` / `label` still raises from `ClassSequentialOps` (unchanged; needs a separate `dtype` param path)

Also updates the container docstring note and replaces the #4493 wart pin with a regression test.

## Test plan

- [x] Local repro from #4493: mosaic boxes move inside the container; mix+mask/bbox/keypoints raise
- [x] Mosaic boxes match `transform_boxes` on the container's `xyxy_plus` Boxes path
- [x] `RandomTransplantation` still works through `AugmentationSequential`
- [x] Existing crop+flip+mosaic bbox reproducibility still holds
- [ ] CI: `tests/augmentation/container/test_augmentation_sequential.py::TestConventionAugmentationSequential::test_mix_children_dispatch_annotation_keys_4493`


ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (cbcc5ea3) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4505.fixed.md`.

ducha-aiki (approved):
Review of `cbcc5ea35fde38b08dba83977b80d05f6508091f`.

Reviewed current head `cbcc5ea3`. The new dispatch branches use the image step's recorded parameters, preserve the existing special handling for `RandomTransplantation`, transform `RandomMosaic` boxes through its native handler, and surface `NotImplementedError` for annotation types a mix child cannot handle instead of silently desynchronizing them. I found no actionable defect. The user-visible behavior change has `changelog.d/4505.fixed.md` and no `CHANGELOG.md` edit. I ran the new regression on CPU float32/float64 and the focused mosaic/transplantation selection: all four cases passed from an isolated worktree with the import path verified. Approve; merge after required CI finishes green.

PR #4506, opened 2026-09-14
fix: correct docstring Args for RandomThinPlateSpline, RandomCutMixV2, Denormalize

### LINKED ISSUE #4496: docs(augmentation): RandomThinPlateSpline, RandomCutMixV2 and Denormalize document constructor arguments that do not exist
Three more augmentation docstrings document constructor arguments that do not exist. This is the same defect as #4437 (`ColorJitter`, `RandomBrightness`, `RandomGaussianBlur`), which #4490 fixes. #4437 does not list these three classes. They turned up when the reviewer of #4490 ran the same signature-vs-`Args` comparison over every public class in `kornia.augmentation`.

| Class | Documented in `Args` | Constructor | Location on `main` (`dc0af1a4`) |
| --- | --- | --- | --- |
| `RandomThinPlateSpline` | `mode: Interpolation mode used by grid_sample` | `padding_mode` (undocumented); there is no `mode` | `kornia/augmentation/_2d/geometric/thin_plate_spline.py:37` |
| `RandomCutMixV2` | `height`, `width` ("the width of the input image", both) | neither exists; `data_keys` is undocumented | `kornia/augmentation/_2d/mix/cutmix.py:44-45` |
| `Denormalize` | `same_on_batch` | `mean, std, p, keepdim` only | `kornia/augmentation/_2d/intensity/denormalize.py:38` |

## Repro

```python

for cls, kw in (
    (K.RandomThinPlateSpline, dict(mode="bilinear")),
    (K.RandomCutMixV2, dict(height=4)),
    (K.RandomCutMixV2, dict(width=4)),
    (K.Denormalize, dict(mean=0.0, std=1.0, same_on_batch=True)),
):
    try:
        cls(**kw)
        print(cls.__name__, "accepted")
    except TypeError as e:
        print(cls.__name__, "TypeError:", e)
```

## Observed (kornia `dc0af1a4`, torch 2.14.0, CPU)

```
RandomThinPlateSpline TypeError: RandomThinPlateSpline.__init__() got an unexpected keyword argument 'mode'
RandomCutMixV2 TypeError: RandomCutMixV2.__init__() got an unexpected keyword argument 'height'
RandomCutMixV2 TypeError: RandomCutMixV2.__init__() got an unexpected keyword argument 'width'
Denormalize TypeError: Denormalize.__init__() got an unexpected keyword argument 'same_on_batch'
```

## Expected

Each `Args` block names exactly the constructor's parameters. The fix is docs-only:
- `RandomThinPlateSpline`: document `padding_mode` in place of `mode`.
- `RandomCutMixV2`: drop `height`/`width` and document `data_keys`.
- `Denormalize`: drop `same_on_batch`.

A good regression test is to extend the signature-vs-docstring test that #4490 adds (`tests/augmentation/test_args_docstrings.py`) to these classes.

abhi-0203:
## Summary

Fixes three augmentation classes whose docstrings document constructor arguments that do not exist.

| Class | Documented | Actual | Fix |
|-------|-----------|--------|-----|
| `RandomThinPlateSpline` | `mode` | `padding_mode` | Renamed to `padding_mode` with correct description |
| `RandomCutMixV2` | `height`, `width` | Neither exists | Removed both; added missing `data_keys` |
| `Denormalize` | `same_on_batch` | Hardcoded to `True` in `super().__init__()` | Removed from docs |

Reproduced the failure with the exact code from the issue:

```python
for cls, kw in (
    (K.RandomThinPlateSpline, dict(mode="bilinear")),
    (K.RandomCutMixV2, dict(height=4)),
    (K.Denormalize, dict(mean=0.0, std=1.0, same_on_batch=True)),
):
    cls(**kw)  # all raise TypeError
```

Closes #4496


ducha-aiki (changes requested):
Review of `3c89172aec968965171cd40bf31a747257748829`.

The three Args corrections match the constructor signatures, and tests/augmentation/test_args_docstrings.py passes all 4 tests.

[P2] Add the required per-PR release note, for example changelog.d/4506.fixed.md. This corrects user-facing API documentation, so it needs a fragment under the Towncrier policy introduced in #4513; the current diff has none. Do not edit CHANGELOG.md. The existing green checks predate this requirement and do not establish compliance. See changelog.d/README.md.

ducha-aiki (changes requested):
Re-review of `3c89172a` against `30dfbf711` (main).

The head has not moved since the 2026-09-14 review, but main has: **#4538 (batch 6b) and #4566 (batch 6c) have landed and already fixed two of this PR's three hunks.** Two of the three are now redundant, and the redundant TPS hunk is the only thing making this PR conflict.

### Executed on `30dfbf711`

```
RandomThinPlateSpline    PASS  doc-only=[] sig-only=[]
RandomCutMixV2           FAIL  doc-only=['height', 'width'] sig-only=['data_keys']
Denormalize              PASS  doc-only=[] sig-only=[]
```

(signature-vs-`Args` comparison, the same one #4496 used; `kornia.__file__` checked into a detached `30dfbf711` worktree, torch 2.14.0, CPU)

- `RandomThinPlateSpline` — main's Args block already reads ``padding_mode: Padding used by ``grid_sample``: 'zeros', 'border' or 'reflection'.`` and the constructor really takes `padding_mode` (`RandomThinPlateSpline(padding_mode="border")` → `SamplePadding.BORDER`). **Already fixed; this hunk is a no-op.**
- `Denormalize` — main's Args block no longer lists `same_on_batch`. **Already fixed; this hunk is a no-op.**
- `RandomCutMixV2` — still documents `height` and `width`, which do not exist (`TypeError: RandomCutMixV2.__init__() got an unexpected keyword argument 'height'`), and still omits the real `data_keys`. **This hunk is the only remaining content, and it is correct.**

### [P2] Drop the two redundant hunks — the TPS one is the sole merge conflict

`git merge-tree --write-tree 30dfbf711 <head>` conflicts in exactly one file:

```
CONFLICT (content): Merge conflict in kornia/augmentation/_2d/geometric/thin_plate_spline.py
```

`denormalize.py` and `cutmix.py` auto-merge. So reducing this PR to the `cutmix.py` hunk both removes the conflict and leaves all of its surviving value. Please rebase onto main and keep only `cutmix.py`.

### [P2] Add the regression pin #4496 asked for

`tests/augmentation/test_args_docstrings.py` exists on main but is parametrized over only the four #4437 classes:

```python
@pytest.mark.parametrize("cls", [K.ColorJitter, K.ColorJiggle, K.RandomBrightness, K.RandomGaussianBlur])
```

That is why this defect survived, and it means the fix currently has no guard — the docstring can drift straight back. #4496 prescribed the fix for this explicitly: "A good regression test is to extend the signature-vs-docstring test that #4490 adds (`tests/augmentation/test_args_docstrings.py`) to these classes."

Adding `K.RandomCutMixV2` to that list is a pin that genuinely discriminates: it **fails on main** and passes on the head (shown above). Adding `K.RandomThinPlateSpline` and `K.Denormalize` passes on main either way, so those two are regression guards rather than pins of this change — still worth adding, but please note the distinction rather than presenting all three as pins.

One optional hygiene note while you are in that file: the Args-block regex terminates only on
`Shape|Returns|Examples?|.. note::|Note`, so on main it captures straight past `Return:`,
`Convention:` and `RandomCutMixV2`'s `Inputs:` block, stopping at `.. note::` / `Returns:`
respectively. That is harmless today — I ran the repo's exact regex against all three classes and it
extracts the right names — because those trailing blocks use bullets rather than `name:` entries. But
it means a future `Convention:` line written as `name: description` would be silently misread as a
constructor argument. Adding `Return|Convention|Inputs` to the alternation is a one-line hedge.

### [P2] Changelog fragment — still outstanding from the last round

There is still no `changelog.d/4506.fixed.md`. This corrects user-facing API documentation, so it needs a fragment under the Towncrier policy (#4513). Do not edit `CHANGELOG.md`. See `changelog.d/README.md`.

### Closing keyword

`Closes #4496` stays accurate: #4496's scope is exactly these three classes, and with main having fixed two, the `cutmix.py` hunk finishes it.

**Verdict: changes requested.** Rebase onto main, keep only the `cutmix.py` hunk, add the extended `test_args_docstrings.py` pin, and add the fragment.

Issue #1528, opened 2022-01-15
The RandomPosterize class raises RuntimeError because of conflicting devices.


RElbers:
### Describe the bug

Using the _RandomPosterize_ class on the GPU device raises the following exception:

```
  File "<...>\Python39\lib\site-packages\kornia\enhance\adjust.py", line 509, in _right_shift
    return (input * 255).to(torch.uint8) / (2 ** shift).to(input.dtype) / 255.0
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

It seems like it's caused by the _bits_ variable not being placed on the same device. It's probably fixed just by placing something like the below code in the _posterize_ function.

```
if bits.device != input.device:
    bits = bits.to(input.device)
```

### Reproduction steps

```bash
import torch
from kornia.augmentation import RandomPosterize
from kornia.enhance import posterize

# On CPU works
transform = RandomPosterize(bits=6, p=1.0)
input = torch.rand(size=(1, 3, 128, 128),  dtype=torch.float32)
out = transform(input)

# On GPU fails with: RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
transform = RandomPosterize(bits=6, p=1.0).to('cuda:0')
input = torch.rand(size=(1, 3, 128, 128), device='cuda:0', dtype=torch.float32)
out = transform(input)

# The posterize function fails, because bits and input are on different devices
input = torch.rand(size=(1, 3, 128, 128), device='cuda:0', dtype=torch.float32)
bits = torch.tensor([6], device='cpu', dtype=torch.int32)
out = posterize(input, bits)
```

### Expected behavior

The `RandomPosterize` class should work on all devices.

### Environment
Using Kornia 0.6.2

```shell
PyTorch version: 1.10.1+cu113
Is debug build: False
CUDA used to build PyTorch: 11.3
ROCM used to build PyTorch: N/A

OS: Microsoft Windows 10 Pro
GCC version: (x86_64-posix-seh, Built by strawberryperl.com project) 8.3.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: N/A

Python version: 3.9.4 (tags/v3.9.4:1f2e308, Apr  6 2021, 13:40:21) [MSC v.1928 64 bit (AMD64)] (64-bit runtime)
Python platform: Windows-10-10.0.19041-SP0
Is CUDA available: True
CUDA runtime version: 11.5.119
[...]

PR #4507, opened 2026-09-14
fix(augmentation): build shear ranges on the requested device and dtype

### LINKED ISSUE #4415: W1b: `RandomShear` and `RandomAffine(shear=...)` cannot be moved off CPU unless `shear` is a 4-tuple
`ShearGenerator.make_samplers` builds its range tensor by stacking two values. The x half comes from `_range_bound(...)`, which places its result on CPU unless it is handed a device — and the call site does not hand it one. The y half, in the float and 2-tuple branches, is `torch.tensor([0, 0], device=device, dtype=dtype)`, already on the requested device. `torch.stack` of the two then raises. The 4-tuple branch stacks two `_range_bound` results, both on CPU, so it survives.

The consequence is that `K.RandomShear(5.0).to("cuda")` and `K.RandomAffine(degrees=0., shear=(5., 5.)).to("cuda")` fail at construction-time device placement — before any input is seen — while the same classes with a 4-tuple `shear` work, and the control in the same run, `RandomRotation`, moves without complaint. The failure is device-generic, not MPS-specific: nothing in the failing line is backend-dependent, and `_range_bound`'s own default is `torch.device("cpu")` whatever device was requested. The same defect has been reported and fixed for other generators (#3704, #1528). The repro below was run on cpu and on mps, the non-CPU backend available on the audit machine, with `PYTORCH_ENABLE_MPS_FALLBACK` unset so no missing kernel is masked.

## Repro

```python
print("kornia", kornia.__version__, "torch", torch.__version__, "| mps available:", torch.backends.mps.is_available())
for label, make in (
    ("RandomShear(5.0)              ", lambda: K.RandomShear(5.0)),
    ("RandomShear((5., 5.))         ", lambda: K.RandomShear((5., 5.))),
    ("RandomShear((-5., 2., 5., 10.))", lambda: K.RandomShear((-5., 2., 5., 10.))),
    ("RandomAffine(shear=5.0)       ", lambda: K.RandomAffine(degrees=0., shear=5.0)),
    ("RandomAffine(shear=(5., 5.))  ", lambda: K.RandomAffine(degrees=0., shear=(5., 5.))),
    ("RandomAffine(shear=(-5.,2.,5.,10.))", lambda: K.RandomAffine(degrees=0., shear=(-5., 2., 5., 10.))),
    ("RandomRotation(5.0)  [control]", lambda: K.RandomRotation(5.0)),
):
    for dev in ("cpu", "mps"):
        try:
            make().to(dev)
            print(f"{label} .to({dev!r}) -> ok")
        except Exception as e:
            print(f"{label} .to({dev!r}) -> {type(e).__name__}: {e}")
```

## Observed

```
kornia 0.9.0rc1 torch 2.14.0 | mps available: True
RandomShear(5.0)               .to('cpu') -> ok
RandomShear(5.0)               .to('mps') -> RuntimeError: Passed CPU tensor to MPS op
RandomShear((5., 5.))          .to('cpu') -> ok
RandomShear((5., 5.))          .to('mps') -> RuntimeError: Passed CPU tensor to MPS op
RandomShear((-5., 2., 5., 10.)) .to('cpu') -> ok
RandomShear((-5., 2., 5., 10.)) .to('mps') -> ok
RandomAffine(shear=5.0)        .to('cpu') -> ok
RandomAffine(shear=5.0)        .to('mps') -> RuntimeError: Passed CPU tensor to MPS op
RandomAffine(shear=(5., 5.))   .to('cpu') -> ok
RandomAffine(shear=(5., 5.))   .to('mps') -> RuntimeError: Passed CPU tensor to MPS op
RandomAffine(shear=(-5.,2.,5.,10.)) .to('cpu') -> ok
RandomAffine(shear=(-5.,2.,5.,10.)) .to('mps') -> ok
RandomRotation(5.0)  [control] .to('cpu') -> ok
RandomRotation(5.0)  [control] .to('mps') -> ok
```

with the traceback, on any of the failing rows (absolute path prefixes abridged):

```
  File "kornia/augmentation/base.py", line 114, in to
    self.set_rng_device_and_dtype(device, dtype)
  File "kornia/augmentation/base.py", line 177, in set_rng_device_and_dtype
    self._param_generator.set_rng_device_and_dtype(device, dtype)
  File "kornia/augmentation/random_generator/base.py", line 64, in set_rng_device_and_dtype
    self.make_samplers(device, dtype)
  File "kornia/augmentation/random_generator/_2d/shear.py", line 75, in make_samplers
    _shear = torch.stack(
             ^^^^^^^^^^^^
RuntimeError: Passed CPU tensor to MPS op
```

## Expected

Both halves of the stack are built on the requested device: pass `device=`/`dtype=` through to `_range_bound`, or build the zero half with `torch.zeros_like(...)` of the `_range_bound` result so the two always agree. That is a one-line change in `kornia/augmentation/random_generator/_2d/shear.py` and it is welcome as a focused PR.

Two things worth doing in the same change. First, the fix should be verified on CUDA as well as MPS — the failing line has no backend-specific code, so the expectation is that both are affected, and that expectation should be executed rather than assumed. Second, the sibling generators are worth a sweep: any generator that stacks a `_range_bound` result with a device-placed constant has the same shape of defect, which is how #3704 and #1528 arose.

## Classification

**API completeness** — `.to(device)` is part of every `nn.Module`'s contract, and two of the eighteen 2D geometric classes raise on it for two of the three accepted `shear` forms. Idiom: 2 (constants/buffers not inheriting dtype/device).

## Disposition

`n/a — single symbol`. One generator; `RandomShear` and `RandomAffine(shear=...)` both route through it, so one change fixes both. Focused fix, welcome as a PR. Batch 6 documents that the shear generator's ranges stay on CPU and pins the behavior; it does not change it.

## Environment

dchaudhari7177:

`ShearGenerator.make_samplers` and the shear branch of `AffineGenerator.make_samplers` built the x shear range with `_range_bound(...)` but never passed `device`/`dtype`. `_range_bound` defaults to CPU and `torch.get_default_dtype()`. With a scalar or 2-tuple `shear`, the y range is `torch.tensor([0, 0], device=device, dtype=dtype)`, so `torch.stack` mixed a CPU tensor with one on the target device, and `RandomShear(5.0).to("cuda")` raised at construction. A 4-tuple `shear` stacked two CPU results, so it did not raise, but it left both ranges on CPU.

The fix passes `device=device, dtype=dtype` to the three shear `_range_bound` calls in each generator. I changed both generators because `RandomAffine` does not go through `ShearGenerator`: `AffineGenerator` has its own copy of the same block. That copy is the reason both classes in the issue fail the same way.

Fixes #4415. Same class of defect as #3704 and #1528.

## What did you check?

No non-CPU device is available here, so the device half is CI's to confirm. The dtype half has the same cause and is visible on CPU: after `.to(dtype=torch.float64)`, a 4-tuple `shear` kept float32 sampler bounds. The new `TestShearRangeDeviceDtype` checks the device **and** dtype of every shear sampler bound for `RandomShear` and `RandomAffine`, for all three `shear` forms and for float32/float64. On a CUDA or MPS leg the device assertion covers the crash in the issue directly.

```text
$ python -m pytest tests/augmentation/test_random_generator.py -k TestShearRangeDeviceDtype --device=cpu --dtype=float32
# generators at upstream main (fix stashed), new test present
2 failed, 10 passed      # the 4-tuple float64 case, RandomShear and RandomAffine
# with the fix
12 passed

$ python -m pytest <every test file that uses RandomShear or RandomAffine> --device=cpu --dtype=float32
1055 passed, 1295 skipped, 7 xfailed

$ uvx ruff@0.16.6 format --check / check   # the pinned pre-commit version
All checks passed!
```

## How did AI help?

I used an AI assistant to trace the call path, which is how the second copy in `AffineGenerator` turned up, and to draft the test and this description. I ran the red/green check and the test files above myself and read the diff before pushing.


ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (60d7a63a) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4507.fixed.md`.

ducha-aiki (changes requested):
Review of `60d7a63a7e11f5e1a9494b01ef770c495154adbf`.

The production fix is correct and the Towncrier fragment is in the right place, but one new test makes the required MPS job fail.

[P1] `tests/augmentation/test_random_generator.py:876` parametrizes `target_dtype` with `torch.float64` independently of the injected device. On MPS, all six float64 variants fail in the generator constructors because MPS does not support float64, before the assertions run. Use the repository's injected `dtype` fixture (so the MPS job supplies float32), or explicitly skip the float64 cases on MPS. Keep the sampler-bound assertions: they are stronger than a smoke-only `.to()` check. Then rerun the MPS job and the two CPU half profiles.

dchaudhari7177:
Thanks for the review, and for moving the entry into the fragment.

Before fixing the MPS parametrization I rebased onto `main` and found #4523 already covers this. `_shear_bound` now passes `device`/`dtype` to both `_range_bound` calls, which is the same fix, and `TestShearSamplerPlacement` in `tests/augmentation/test_parameter_constants.py` checks the same sampler bounds on both generators for the scalar, 2-tuple and 4-tuple forms, using the injected `dtype` fixture so MPS is fine. I copied my `TestShearRangeDeviceDtype` onto current `main` without the production change and all 12 cases pass, so it adds nothing now.

Closing as superseded by #4523. #4415 can be closed against that PR.

PR #4508, opened 2026-09-14
fix(augmentation): preserve shear sampler device and dtype

Hrithikcrick:

Fixes device and dtype propagation for shear ranges in both `ShearGenerator` and `AffineGenerator`.

`_range_bound(...)` defaults to CPU/default dtype when `device` and `dtype` are not provided. For scalar and 2-tuple shear values, this caused the x shear range to remain on CPU while the zero y range was created on the requested device, making `.to("cuda")` / `.to("mps")` fail.

This change forwards the requested `device` and `dtype` to `_range_bound(...)` in both generators.

## Tests

Added regression coverage for:

- `RandomShear`
- `RandomAffine(..., shear=...)`
- scalar shear
- 2-tuple shear
- 4-tuple shear


ducha-aiki (commented):
Review of `2f7d95c1e6972e5b7f10ec65a4b17e6148f0a64d`.

This is a duplicate of #4507: both PRs make the same two production changes by passing the requested device and dtype into the shear `_range_bound` calls in `AffineGenerator` and `ShearGenerator`. #4507 opened 12 minutes earlier, closes #4415, already has the required Towncrier fragment, and asserts the device and dtype of every sampler bound rather than only checking that `.to(...)` returns. Please close this PR in favor of #4507. If maintainers choose this branch instead, it first needs `changelog.d/4508.fixed.md` and an issue-closing link; it currently has neither.

Hrithikcrick:
Thanks for the review and clarification. I agree that #4507 covers the same fix and was opened earlier. Closing this PR in favor of #4507.

PR #4509, opened 2026-09-14
fix(augmentation): honour an explicit mask resample on the 2D geometric forward path

### LINKED ISSUE #4419: W1b: `extra_args[DataKey.MASK]`'s `resample` override is discarded on the 2D geometric classes
`AugmentationSequential`'s `extra_args` is documented as the per-data-key control: *"to control the behaviour for each datakeys. By default, masks are handled by nearest interpolation strategies"* (`container/augment.py:85-86`). "By default" reads as an overridable default, and the dictionary the container installs when the caller passes nothing is exactly `{DataKey.MASK: {'resample': Resample.NEAREST, 'align_corners': None}}` — the same two keys a caller would write to change it.

For `DataKey.MASK` on a 2D geometric augmentation the `resample` half of that dictionary never reaches the sampler. `RigidAffineAugmentationBase2D.apply_transform_mask` reads the merged flags and then overwrites them — `kornia/augmentation/_2d/geometric/base.py:107` does `flags["resample"] = Resample.get("nearest")` after `transform_masks` merged the caller's entry, and `:230` repeats it on the inverse path. The `align_corners` half has no such line and does reach the sampler, so the same dictionary is half honoured and half dropped, with no warning either way. `RandomElasticTransform` does not go through that base and honours the `resample` entry, which is what makes the discard visible as a divergence rather than a policy.

Two further facts about `extra_args` are undocumented and worth stating with the fix: a caller's dictionary **replaces** the default wholesale rather than merging into it (so `{MASK: {'align_corners': True}}` silently drops the `NEAREST` entry), and `DataKey.IMAGE` honours both entries while `DataKey.KEYPOINTS` honours neither.

## Repro

```python
import torch
from kornia.augmentation import AugmentationSequential
from kornia.constants import DataKey, Resample

torch.manual_seed(42)
x = torch.rand(2, 3, 6, 8)
m = torch.zeros(2, 1, 6, 8)
m[:, :, 1:4, 2:6] = 1.0

def mask_out(build, extra):
    torch.manual_seed(0)
    seq = AugmentationSequential(build(), data_keys=["input", "mask"], extra_args=extra)
    return seq(x, m)[1]

CASES = [
    ("RandomAffine",       lambda: K.RandomAffine(degrees=(45.0, 45.0), p=1.0)),
    ("RandomPerspective",  lambda: K.RandomPerspective(distortion_scale=0.5, p=1.0)),
    ("RandomRotation",     lambda: K.RandomRotation(degrees=(45.0, 45.0), p=1.0)),
    ("RandomTranslate",    lambda: K.RandomTranslate(translate_x=(0.5, 0.5), translate_y=(0.5, 0.5), p=1.0)),
    ("Resize",             lambda: K.Resize((8, 12))),
    ("RandomElasticTransform", lambda: K.RandomElasticTransform(p=1.0)),
]
bilinear = {DataKey.MASK: {"resample": Resample.BILINEAR, "align_corners": None}}
ac_true  = {DataKey.MASK: {"resample": Resample.NEAREST, "align_corners": True}}

print("class                     | resample=BILINEAR override | align_corners=True override")
for name, build in CASES:
    base = mask_out(build, None)
    r = (mask_out(build, bilinear) - base).abs().max().item()
    a = (mask_out(build, ac_true) - base).abs().max().item()
    print(f"{name:25s} | {r!r:26s} | {a!r}")

print()
seq = AugmentationSequential(K.RandomAffine(degrees=(45.0, 45.0), p=1.0), data_keys=["input", "mask"])
print("default extra_args        :", seq.extra_args)
seq2 = AugmentationSequential(K.RandomAffine(degrees=(45.0, 45.0), p=1.0), data_keys=["input", "mask"],
                              extra_args={DataKey.MASK: {"align_corners": True}})
print("user dict replaces default:", seq2.extra_args)
```

## Observed

```
class                     | resample=BILINEAR override | align_corners=True override
RandomAffine              | 0.0                        | 0.0
RandomPerspective         | 0.0                        | 1.0
RandomRotation            | 0.0                        | 0.0
RandomTranslate           | 0.0                        | 1.0
Resize                    | 0.0                        | 0.0
RandomElasticTransform    | 0.4551059603691101         | 0.0

default extra_args        : {<DataKey.MASK: 1>: {'resample': <Resample.NEAREST: 0>, 'align_corners': None}}
user dict replaces default: {<DataKey.MASK: 1>: {'align_corners': True}}
```

The same `0.0` was executed on all 18 2D geometric classes: `RandomAffine`, `CenterCrop`, `RandomCrop` (both `slice` and `resample` cropping modes), `RandomHorizontalFlip`, `PadTo`, `RandomPerspective`, `LongestMaxSize`, `Resize`, `SmallestMaxSize`, `RandomResizedCrop`, `RandomRotation`, `RandomRotation90`, `RandomShear`, `RandomTranslate` and `RandomVerticalFlip` drop it; `RandomElasticTransform` honours it; `RandomFisheye` and `RandomThinPlateSpline` raise `NotImplementedError` on a `MASK` key before any override is read. `RandomResizedCrop` rejects `align_corners=True` with `ValueError: align_corners option can only be set with the interpolating modes…`.

## Expected

Two coherent outcomes:

1. **Honour the override.** Delete the two `flags["resample"] = Resample.get("nearest")` lines and let the merged flags stand — nearest stays the default because the container installs it as the default, and a caller who asks for bilinear gets bilinear, as the docstring implies and as `DataKey.IMAGE` and `RandomElasticTransform` already behave.
2. **Refuse the override.** Keep masks pinned to nearest and reject a `resample` entry under `DataKey.MASK` with a message, rather than accepting and dropping it.

The audit recommends (1): the discard is not a deliberate mask policy, since the neighbouring `align_corners` entry in the same dictionary reaches the sampler and changes the mask, and one class in the same container honours both. Either way the `extra_args` docstring needs the replace-not-merge rule and the per-data-key scope added.

## Classification

**Broken output** — a documented override is merged and then overwritten, so a caller's `resample` choice is silently discarded on the mask path. Idiom: 3 (a guard whose predicate does not match the documented contract).

## Disposition

`n/a — single symbol`. Batch 6 documents the current behavior on `AugmentationSequential.extra_args` and pins it, and does not change it. The repair is a focused fix in `kornia/augmentation/_2d/geometric/base.py` — welcome as a PR. Related history: #2002 (the elastic `mode`/`resample` naming, closed) and #2037 (nearest in nested containers, closed).

## Environment

dchaudhari7177:

On the forward path, `GeometricAugmentationBase2D.apply_transform_mask` set `flags["resample"] = Resample.get("nearest")` **after** `transform_masks` had already merged the caller's kwargs into the flags. So `extra_args[DataKey.MASK]["resample"]` was accepted and then dropped on every class that uses the base mask path.

I took the issue's option (1), with one refinement. The issue suggests deleting the overwrite outright, but on its own that would change the default. The module's own `resample` is its **image** mode (bilinear for `RandomAffine`), so a direct `transform_masks` call, or a user `extra_args` dict that leaves `resample` out, would start resampling masks bilinearly. Instead:

- `transform_masks` supplies `resample=nearest` only when the caller did not pass one, then defers to the base implementation.
- `apply_transform_mask` uses the merged flags as given. That also removes the in-place mutate-and-restore of `flags` for `resample`.

This matches what `inverse_masks` **already** did. It forces nearest in place, but `inverse_inputs` then merges `kwargs` back over it, so an explicit value won on inverse and only forward discarded it. Measured on `main` with `{MASK: {"resample": BILINEAR}}` against the default dict: forward delta `0.0` on all of `RandomAffine`/`RandomRotation`/`RandomPerspective`/`RandomTranslate`/`RandomShear`/`Resize`/`RandomResizedCrop`/`LongestMaxSize`/`SmallestMaxSize`, while inverse delta was `0.34`–`0.48` on the same classes.

Docs: the `.. warning::` blocks on `GeometricAugmentationBase2D` and `AugmentationSequential` are replaced. The `extra_args` description now says the entry is honoured in both directions on the base mask path and that nearest stays the default without it. The replace-not-merge sentence and the `RandomElasticTransform` note are kept.

Fixes #4419.

## What did you check?

Pins, per AGENTS.md (`grep -rnE "#4419|issues/4419" kornia/ tests/` has no hits left in `kornia/`):
- `test_wart_extra_args_mask_resample_override_is_discarded_4419` → renamed `test_extra_args_mask_override_reaches_the_sampler_4419`. The affine assertion is inverted (`== 0.0` → `> 0.0`). Its align_corners, elastic and replace-not-merge assertions are unchanged.
- The strict `xfail` `test_convention_extra_args_mask_resample_override_is_honoured` (XPASSed with the fix) → `test_extra_args_mask_resample_is_honoured_and_nearest_stays_the_default_4419`, parametrized over six classes. On a checkerboard mask it asserts that a bilinear override gives non-binary output **and** that the default dict and a dict without `resample` both stay binary.
- New `test_direct_transform_masks_call_without_resample_stays_nearest_4419`: a direct call on `RandomAffine` (image `resample` bilinear) still returns a binary mask and leaves `aug.flags["resample"]` untouched.

```text
$ python -m pytest tests/augmentation/container/test_augmentation_sequential.py -k 4419 --device=cpu --dtype=float32
# base.py at upstream main, new tests present
7 failed, 1 passed      # the direct-call guard passes on both sides by design
# with the fix, float32 and float64
16 passed

# robustness of the parametrized test: seeds 0-19 x float32/float64 x the six classes -> no failures

$ python -m pytest tests/augmentation --device=cpu --dtype=float32
2675 passed, 3410 skipped, 30 xfailed, 5 xpassed   # the 5 XPASS are the pre-existing Windows print-precision / CenterCropAlternative markers

$ uvx ruff@0.16.6 check / format --check ; doctest on the two touched modules
All checks passed!
```

Not verified locally: CPU float16/bfloat16. My torch is 2.5.0, where `grid_sampler_2d_cpu` has no Half kernel, so these classes fail at `upstream/main` too. The CI half legs on the pinned 2.9.1 are the real check. MPS/CUDA are CI's as well.

## How did AI help?

I used an AI assistant to trace how `kwargs` merge into `flags` on the forward and inverse paths, to write the probe that measured both directions across the geometric classes, and to draft the tests and this description. I ran the red/green and suite checks above and reviewed the diff, including the choice not to simply delete the overwrite.


ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (9f50b2bc) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4509.fixed.md`.

ducha-aiki (approved):
Review of `9f50b2bcd4fbc1c092e94ed19d14de41f0861242`.

Reviewed current head `9f50b2bc`. The fix puts the nearest-neighbor mask default in `GeometricAugmentationBase2D.transform_masks` only when the caller did not supply `resample`, so an explicit container override now reaches `apply_transform` while direct calls and partial override dictionaries keep the old nearest default. It also stops mutating the module flags on the forward path. I found no correctness regressions. The user-visible change is recorded in `changelog.d/4509.fixed.md` and `CHANGELOG.md` is untouched. I ran the complete `-k 4419` regression set from an isolated worktree after verifying `kornia.__file__`: 16 passed on CPU float32/float64. Approve; merge after the refreshed required matrix finishes green.

PR #4510, opened 2026-09-14
docs(augmentation): RandomElasticTransform's alpha is (x, y)

### LINKED ISSUE #4418: W1b: `RandomElasticTransform`'s `alpha` doc has x and y swapped
`RandomElasticTransform`'s `Args` says `alpha` is "the scaling factor that controls the intensity of the deformation in the y and x directions, respectively". The code applies `alpha[0]` to the x displacement and `alpha[1]` to the y displacement — the opposite order. The confusion is easy to acquire because the two neighbouring parameters really are `(y, x)`: `kernel_size` and `sigma` both index row first, and `sigma` is documented correctly with the same sentence shape.

The result is that `alpha=(4., 0.)` scales the horizontal displacement while the docstring promises the vertical one, and a caller who wants an anisotropic warp gets it on the wrong axis. A symmetric `alpha=(a, a)` cannot see the difference, and that is what the class's own tests pass, which is why the sentence has survived.

The delegated function, `kornia.geometry.transform.elastic_transform2d`, already states the correct order ("in the x and y directions, respectively") and carries a `Convention:` block recording that channel 0 of `noise` is the x displacement; that half was corrected in the batch-1 conventions work (#3926). Only the augmentation class's copy of the sentence is still wrong.

## Repro

```python
import torch, inspect, kornia, kornia.augmentation as K
# a strictly positive, asymmetric image, so the centre of mass is well defined
img = torch.arange(1., 49.).reshape(1, 1, 6, 8)

def com(t):
    t = t[0, 0]
    r = torch.arange(6.)[:, None].expand_as(t); c = torch.arange(8.)[None, :].expand_as(t)
    s = t.sum()
    return (round((t * r).sum().item() / s.item(), 4), round((t * c).sum().item() / s.item(), 4))

def run(alpha, ch):
    noise = torch.zeros(1, 2, 6, 8)
    if ch is not None:
        noise[:, ch] = 1.0                      # a constant +1 in ONE noise channel
    aug = K.RandomElasticTransform(kernel_size=(3, 3), sigma=(1., 1.), alpha=alpha,
                                   padding_mode="border", p=1.)
    return com(aug(img, params={"noise": noise, "batch_prob": torch.tensor([True])}))

print("input                                     com(row, col) =", com(img))
print("no displacement, alpha=(0., 0.)           com(row, col) =", run((0., 0.), None))
print("noise channel 0 (x) = +1, alpha=(4., 0.)  com(row, col) =", run((4., 0.), 0))
print("noise channel 0 (x) = +1, alpha=(0., 4.)  com(row, col) =", run((0., 4.), 0))
print("noise channel 1 (y) = +1, alpha=(4., 0.)  com(row, col) =", run((4., 0.), 1))
print("noise channel 1 (y) = +1, alpha=(0., 4.)  com(row, col) =", run((0., 4.), 1))
print("--- RandomElasticTransform docstring ---")
print("\n".join(l for l in inspect.getdoc(K.RandomElasticTransform).splitlines()
                if "alpha" in l or "sigma" in l or "directions" in l))
print("--- elastic_transform2d docstring ---")
print("\n".join(l for l in inspect.getdoc(kornia.geometry.transform.elastic_transform2d).splitlines()
                if "alpha" in l or "sigma" in l or "directions" in l))
```

## Observed

```
kornia 0.9.0rc1 torch 2.14.0
input                                     com(row, col) = (3.4524, 3.7143)
no displacement, alpha=(0., 0.)           com(row, col) = (3.5068, 3.727)
noise channel 0 (x) = +1, alpha=(4., 0.)  com(row, col) = (3.381, 3.5)
noise channel 0 (x) = +1, alpha=(0., 4.)  com(row, col) = (3.5068, 3.727)
noise channel 1 (y) = +1, alpha=(4., 0.)  com(row, col) = (3.5068, 3.727)
noise channel 1 (y) = +1, alpha=(0., 4.)  com(row, col) = (2.5, 3.625)
--- RandomElasticTransform docstring ---
    sigma: The standard deviation of the Gaussian in the y and x directions,
      respectively. Larger sigma results in smaller pixel displacements.
    alpha: The scaling factor that controls the intensity of the deformation
      in the y and x directions, respectively.
--- elastic_transform2d docstring ---
      ``sigma``) and scaled by ``alpha``
    kernel_size: the size of the Gaussian kernel, in the y and x directions, respectively.
    sigma: The standard deviation of the Gaussian in the y and x directions,
      respectively. Larger sigma results in smaller pixel displacements.
    alpha : The scaling factor that controls the intensity of the deformation
      in the x and y directions, respectively.
    >>> sigma = torch.tensor([4., 4.], requires_grad=True)
    >>> image_hat = elastic_transform2d(image, noise, (3, 3), sigma)
    >>> alpha = torch.tensor([16., 32.], requires_grad=True)
    >>> image_hat = elastic_transform2d(image, noise, (3, 3), alpha=alpha)
```

The two rows that move are `alpha=(4., 0.)` with x-noise, which shifts the centre of mass along the **column** axis (`3.727 → 3.5`), and `alpha=(0., 4.)` with y-noise, which shifts it along the **row** axis (`3.5068 → 2.5`). The two crossed combinations leave the image at the no-displacement baseline: `alpha[0]` scales x only and `alpha[1]` scales y only.

## Expected

`RandomElasticTransform`'s `alpha` line reads "in the x and y directions, respectively", matching the code and matching the same line on `elastic_transform2d`. The `sigma` and `kernel_size` lines stay as they are — those really are `(y, x)`, and the asymmetry between the two parameters is worth stating on the class rather than smoothing over, since it is the reason the wrong sentence looked plausible.

Docs-only change, no behavior change, welcome as a PR.

## Classification

**Semantic contract** — the documented axis order for `alpha` is the reverse of the implemented one, on a parameter whose only asymmetric use is the one the doc misdescribes. Idiom: 7 (an axis-order slip hidden by symmetric test data — the class's own tests pass a symmetric `alpha`; the asymmetric coverage that exists is on the function, added with the batch-1 blocks).

## Disposition

`n/a — single symbol`. One wrong sentence on one class. Note for whoever picks it up: the audit row this came from also named `elastic_transform2d`'s copy of the sentence as wrong; the fresh run above shows it is already correct, so the change is to the augmentation class only. Batch 6 documents the executed `(x, y)` order for `alpha` against the `(y, x)` order for `sigma` and `kernel_size`, and pins both.

## Environment

dchaudhari7177:

`RandomElasticTransform`'s `Args` block said `alpha` controls the deformation "in the y and x directions, respectively". `elastic_transform2d` does the opposite: `alpha[0]` scales `disp_x` (noise channel 0) and `alpha[1]` scales `disp_y`. So `alpha=(4., 0.)` warps horizontally while the docstring promised vertically. The functional already documents `(x, y)`, so only the augmentation class's sentence changes.

The wrong order is easy to pick up because the two neighbouring parameters really are `(y, x)` (`ksize_y, ksize_x = kernel_size`, `sigma_y, sigma_x = sigma[:, 0], sigma[:, 1]`). The new sentence therefore calls out that `alpha` runs the other way. I also added "in the y and x directions, respectively" to the class's `kernel_size`, matching the functional's wording, since that was the one parameter of the three with no order stated.

Fixes #4418. As the disposition notes, `elastic_transform2d`'s copy is already correct and is unchanged.

## What did you check?

`test_alpha_is_x_then_y` pins the order through the augmentation class instead of through prose. The image is constant down each column and both noise channels are `+1`. A y displacement cannot change such an image and an x displacement must:

```text
alpha=(4., 0.) vs alpha=(0., 0.): max|delta| = 7.0
alpha=(0., 4.) vs alpha=(0., 0.): max|delta| = 0.0
```

It passes before and after this change, since the code is not changing. It exists so the docstring and the behaviour can't drift apart again unnoticed.

```text
$ python -m pytest tests/augmentation/test_augmentation.py -k TestRandomElasticTransform --device=cpu --dtype=float32,float64
14 passed
$ python -m pytest tests/augmentation/test_args_docstrings.py
4 passed
All checks passed!
```

## How did AI help?

I used an AI assistant to cross-check the parameter order in `elastic_transform2d` and to draft the test and this description. I ran the probe and the tests above and reviewed the wording.


ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (05f2fa6a) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4510.fixed.md`.

ducha-aiki (approved):
Review of `05f2fa6a6d98d6b5dd26156b00c2e048f90df8a9`.

The corrected alpha order agrees with elastic_transform2d: alpha[0] scales horizontal displacement and alpha[1] scales vertical displacement, whereas kernel_size and sigma use (y, x). The axis-discriminating regression exercises that contract through the augmentation. The Towncrier fragment is present.

Validation: TestRandomElasticTransform across CPU float32/float64/float16/bfloat16: 28 passed. Fresh hosted CI remains a merge gate.

Issue #4436, opened 2026-09-11
W1b: `RandomAutoContrast.clip_output` is a no-op


ducha-aiki:
`RandomAutoContrast` takes a public `clip_output` constructor argument, documented at `kornia/augmentation/_2d/intensity/auto_contrast.py:31` as "if true clip output" and defaulting to `True`. `apply_transform` computes `normalize_min_max(input)` and then, at `:60`, clamps that into `[0, 1]` when the flag is set.

The clamp cannot do anything. `kornia.enhance.normalize_min_max` rescales each per-sample, per-channel plane onto `[0, 1]` by construction, so its output is already inside the interval the clamp would enforce. Setting the flag either way gives the identical tensor on a constant image, on an in-range image, on an image scaled into `[0, 2]` and on one shifted into `[-1, 0]` — the last two being the cases a caller would reach for the flag to protect against.

The argument therefore promises out-of-range protection that a user might rely on when feeding non-normalized data, and provides none, because there is nothing to protect against. Worth noting alongside it: on a constant image, the operator returns zeros for both settings, which is `normalize_min_max`'s behavior for a zero-range plane and is not what "auto contrast" suggests.

## Repro

```python
import torch, kornia.augmentation as K
from kornia.enhance import normalize_min_max

torch.manual_seed(0)
base = torch.rand(2, 3, 6, 8)
for tag, x in (("constant 0.5", torch.full((2, 3, 6, 8), 0.5)), ("[0,1] image", base),
               ("[0,2] image", base * 2.0), ("[-1,0] image", base - 1.0)):
    torch.manual_seed(0); t = K.RandomAutoContrast(clip_output=True, p=1.0)(x.clone())
    torch.manual_seed(0); f = K.RandomAutoContrast(clip_output=False, p=1.0)(x.clone())
    print(f"{tag:<13} clip_output=False min/max {f.min().item():.6g}/{f.max().item():.6g}   "
          f"max|clip_output=True - clip_output=False| = {(t - f).abs().max().item()}")
torch.manual_seed(0)
print("max|RandomAutoContrast(x) - kornia.enhance.normalize_min_max(x)|:",
      (K.RandomAutoContrast(p=1.0)(base.clone()) - normalize_min_max(base)).abs().max().item())
n = normalize_min_max(base * 2.0)
print("normalize_min_max already returns min/max:", round(n.min().item(), 6), round(n.max().item(), 6))
```

## Observed

```
[...]

PR #4511, opened 2026-09-14
docs(augmentation): say RandomAutoContrast's clip_output has no effect

dchaudhari7177:

This is option (2) from #4436, the docstring correction that the disposition welcomes as a PR now. It does not remove the argument, which is the `MERGE IN WINDOW` half.

`RandomAutoContrast.apply_transform` clamps `normalize_min_max(input)` to `[0, 1]` when `clip_output` is set. But `normalize_min_max` computes `(x - min) / (max - min + eps)` per sample and channel, which is already inside `[0, 1]`, so the clamp cannot change anything. `clip_output: if true clip output` now reads as having no effect on the output. It gives the reason, mentions that this holds for inputs outside `[0, 1]` too, and says the argument is kept for signature compatibility. The class note also records that a constant channel comes back as zeros, which the issue flagged as not what "auto contrast" suggests.

Refs #4436. It should stay open for the removal in the Repair & Deprecation window.

## What did you check?

I checked that the no-op claim holds beyond float32 before writing it into the docstring. Six input ranges (`[0,1]`, `[0,2]`, `[-1,0]`, `[-5000,5000]`, a `1e-6`-wide range, and a constant image), 4×3×16×16, each dtype:

```text
float16  max|clip_output=True - clip_output=False| = 0.0   (output min >= 0, max <= 1)
bfloat16 0.0
float32  0.0
float64  0.0
```

`test_clip_output_does_not_change_the_output` pins that with an exact comparison (`rtol=atol=0`) on four of those ranges, plus the `[0, 1]` bounds, so a future change to the transform can't make the documented claim false without the test noticing.

```text
$ python -m pytest tests/augmentation/test_augmentation.py -k TestRandomAutoContrast --device=cpu --dtype=float32,float64,float16,bfloat16
17 passed, 1 skipped
$ python -m pytest tests/augmentation/test_args_docstrings.py --doctest-modules kornia/augmentation/_2d/intensity/auto_contrast.py
4 passed
All checks passed!
```

## How did AI help?

I used an AI assistant to draft the docstring, the dtype sweep and the test. I ran the sweep and the tests above and checked the claim against `normalize_min_max`'s formula.


ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (2f85843b) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4511.fixed.md`.

ducha-aiki (approved):
Review of `2f85843b5c678ed537ddfd5257bedd4d6b1e5815`.

The docstring now describes the actual normalize_min_max/clamp behavior and the constant-channel result. The signature remains unchanged, and Refs #4436 correctly leaves the future argument-removal decision open. The per-PR Towncrier fragment is present.

Validation: TestRandomAutoContrast across CPU float32/float64/float16/bfloat16: 17 passed, 1 skipped. Code/docstring and the fragment were inspected. Fresh hosted CI remains a merge gate.

ducha-aiki (approved):

Issue #4417, opened 2026-09-11
W1b: the three crop classes disagree about an `int` `size`, with two unhelpful messages


ducha-aiki:
`CenterCrop`, `RandomCrop` and `RandomResizedCrop` all take a `size` naming a square or rectangular output, and all three accept a `(h, w)` tuple. Only `CenterCrop` accepts a bare `int`. `RandomCrop(4)` raises an `AssertionError` whose message describes a `torch.Tensor` argument with shape `(B, 2)` — a tensor the caller never passed and a shape that has nothing to do with `4`. `RandomResizedCrop(4)` raises `TypeError: object of type 'int' has no len()` from inside the constructor, naming nothing at all.

Both messages leak an internal representation rather than naming the argument that was wrong. A caller who has just used `CenterCrop(4)` successfully and swaps in `RandomCrop(4)` has to read the source to discover that the fix is `(4, 4)`, and that `RandomResizedCrop` needs the same.

## Repro

```python
import torch, kornia, kornia.augmentation as K
print("kornia", kornia.__version__, "torch", torch.__version__)
x = torch.arange(96., dtype=torch.float32).reshape(2, 1, 6, 8)
for cls in (K.CenterCrop, K.RandomCrop, K.RandomResizedCrop):
    torch.manual_seed(0)
    try:
        print(f"{cls.__name__}(4)  -> {tuple(cls(4, p=1.0)(x).shape)}")
    except Exception as e:
        print(f"{cls.__name__}(4)  -> {type(e).__name__}: {e}")
    torch.manual_seed(0)
    print(f"{cls.__name__}((4, 4)) -> {tuple(cls((4, 4), p=1.0)(x).shape)}")
```

## Observed

```
kornia 0.9.0rc1 torch 2.14.0
CenterCrop(4)  -> (2, 1, 4, 4)
CenterCrop((4, 4)) -> (2, 1, 4, 4)
RandomCrop(4)  -> AssertionError: If `size` is a torch.Tensor, it must be shaped as (B, 2). Got torch.Size([2, 1]) while expecting torch.Size([2, 2]).
RandomCrop((4, 4)) -> (2, 1, 4, 4)
RandomResizedCrop(4)  -> TypeError: object of type 'int' has no len()
RandomResizedCrop((4, 4)) -> (2, 1, 4, 4)
```

## Expected

One behavior across the three classes. Two coherent outcomes:

[...]

PR #4512, opened 2026-09-14
fix(augmentation): name size in the int-size errors of RandomCrop and RandomResizedCrop

dchaudhari7177:

This is the error-message half of #4417, which the disposition welcomes now. Whether the three crops should agree on accepting an `int` is left to the window.

Before, on upstream `main`:

```text
RandomCrop(4)(x)     -> AssertionError: If `size` is a torch.Tensor, it must be shaped as (B, 2). Got torch.Size([2, 1]) while expecting torch.Size([2, 2]).
RandomResizedCrop(4) -> TypeError: object of type 'int' has no len()
```

After:

```text
RandomCrop(4)(x)     -> AssertionError: `size` must be a (height, width) pair of integers or a (B, 2) tensor. Got 4.
RandomResizedCrop(4) -> TypeError: `output_size` (`size` on RandomResizedCrop) must be a (height, width) tuple of 2 positive integers. Got 4 of type int.
```

- `CropGenerator.forward`: a non-tensor `size` that does not expand to `(B, 2)` gets its own message. The tensor branch and its existing message are unchanged.
- `ResizedCropGenerator.__init__`: a type check comes before `len(output_size)`. The generator's parameter is `output_size`, so the message names both spellings.

Exception types (`AssertionError`, `TypeError`) and the point where each raises are deliberately unchanged: `RandomCrop` still raises on the first forward and `RandomResizedCrop` at construction. Moving `RandomCrop`'s check to construction would be a behaviour change, and unifying the types belongs with the signature decision.

Refs #4417. It stays open for the `size`-type unification in the window.

## What did you check?

`TestCropIntSizeMessages` pins both new messages and that the tensor-shape message is untouched.

```text
$ python -m pytest tests/augmentation/test_random_generator.py -k TestCropIntSizeMessages --device=cpu --dtype=float32
# crop.py at upstream main
2 failed, 1 passed       # the two int-size message tests; the tensor-message guard passes on both sides
# with the change
3 passed

$ python -m pytest tests/augmentation --device=cpu --dtype=float32
2671 passed, 3410 skipped, 31 xfailed, 5 xpassed   # the 5 XPASS are pre-existing Windows print-precision markers

All checks passed!
```

## How did AI help?

I used an AI assistant to trace where each error is raised and to draft the messages, tests and this description. I ran the red/green and suite checks above and reviewed the diff.


ducha-aiki:
Kornia switched to per-PR changelog fragments in #4513 (towncrier), and `CHANGELOG.md` no longer has a hand-edited Unreleased section, so every open PR that touched it became conflicting. To spare you a rebase I pushed a merge commit to this branch (430a2df0) that:

- merges the current `main`;
- moves your entry, text unchanged, into `changelog.d/4512.fixed.md`.

ducha-aiki (approved):
Review of `430a2df062ff02b1e79d80940c9dfb18053fc206`.

The new int-size diagnostics name the public argument and preserve the existing exception types and raising points. The tensor-size error is unchanged. The Towncrier fragment is correctly named and CHANGELOG.md is untouched.

Validation: all 9 new message cases pass across CPU float32/float64/float16/bfloat16. The wider crop-generator selection gives 61 passed and 4 half-precision failures; the identical selection on current main reproduces those same four failures (52 passed), so they are not introduced by this PR. Import origins and the explicit project interpreter were verified. Fresh hosted CI remains a merge gate.

PR #4500, opened 2026-09-13
fix(camera): own PinholeCamera constructor tensors

victorwon2001:
## What does this pull request do?

`PinholeCamera` now clones `intrinsics`, `extrinsics`, `height`, and `width` at construction, so `scale_()` and the translation setters mutate only camera-owned storage. It also removes redundant copies in `clone()` and `scale()`, updates the ownership documentation and changelog, and replaces the #4264 wart pin with positive regression coverage while preserving the #4263 principal-point rule and #4265 integer-image-size failure behavior.

## Related issue or discussion

Fixes #4264

## What did you check?

Focused ownership regressions pass on CPU for float16, bfloat16, float32, and float64; the complete float32/float64 `TestPinholeCamera` suite, type checking, pre-commit hooks, and the warnings-as-errors documentation build also pass. The broader all-dtype class run still has two low-precision `project/unproject` tolerance failures at float16 and bfloat16; diagnostic comparison found no result difference from constructor cloning for the same inputs.

## How did AI help?

AI was used to implement the constructor ownership change, update the regression tests and documentation, and run the verification reported above.

PR #4513, opened 2026-09-14
Avoid changelog merge conflicts with per-PR fragments

ducha-aiki:
We are all tired  of merge conflicts because of the Changelog. So we will instead use a temp change log directory, per PR, and the main changelog is assembled automatically at release time.

The part below written by Codex (gpt-6-astra) on my behalf, explaining the details. 

## How does it work
Changelog conflicts happen because several PRs add entries to the same lines under `Unreleased`. This switches to one small file per PR, such as `changelog.d/4491.fixed.md`, so unrelated changes can merge independently.

Contributors write the release note in that file. Maintainers run Towncrier when preparing a release to combine the files into `CHANGELOG.md` and remove the consumed fragments. The headings stay **Added**, **Bug fixes**, and **Breaking changes**.

- `pixi run changelog-preview` prints pending notes without changing files.
- `pixi run changelog-build --version X.Y.Z` assembles a release without an interactive prompt, then stages the changelog and consumed-fragment deletions for review.
- Contributor instructions, agent guidance, import-surface diagnostics, and the PR template explain the new convention. CI checks fragment names and rendering on every PR. Ordinary PRs must add a fragment and leave `CHANGELOG.md` alone. A maintainer can apply `no-changelog` for internal changes or release PRs; filename and rendering validation still run.

Most of the diff moves the 177 existing unreleased entries into fragments. Their content and PR references are preserved (directional cross-references now name the category); published changelog history is unchanged. Towncrier runs as a pinned, isolated development tool and adds no runtime dependency.

Existing PRs that edit `CHANGELOG.md` will need a one-time update to add a fragment instead. Generated changelogs belong only in release PRs, otherwise the same conflicts would return.

## What did you check?

- `pixi run test-module tests/scripts/test_check_changelog.py -q` passed (11 tests).

- `pixi run pre-commit-all` and `pixi run toml-fmt-check` passed.
- `pixi run test-module tests/scripts/test_check_import_surface.py` passed.
- Checked all 177 migrated entries; updated five cross-references so sorting cannot make “above” or “below” misleading.
- Tested preview and release generation in a temporary Git checkout, including closed stdin: correct insertion point, all 177 notes preserved, fragments consumed, historical releases and README retained. Also checked Finder metadata, invalid categories, and multi-PR deduplication.

The library test suite was not run because no library code changes. This tooling-only PR uses the `no-changelog` exemption.

## How did AI help?

Codex implemented the workflow, migrated the existing entries, and checked preview and release generation.


ducha-aiki:
## Review: approve, with a few non-blocking follow-ups

Verified locally at b2c782d3 (the checkout, not the PR description):

- **Migration fidelity.** Parsed the 177 `* ` entries under `## Unreleased` on `main` and the 177 `- ` bullets that `pixi run changelog-preview` renders, normalised whitespace, compared per category: Added 24/24, Bug fixes 133/133, Breaking changes 20/20, nothing missing, nothing extra. Only the order differs: Towncrier sorts within a category (numbered fragments by number, orphans last, alphabetically by text), so the previous newest-first order is gone. Acceptable for release notes.
- **Real release build.** In a throwaway worktree, `towncrier build --version 0.9.0 --yes` with two extra future-style fragments `4513.added.md` / `4513.fixed.md`: a dated `## [0.9.0] - …` section lands after the marker, the marker and header survive, all 177 migration fragments are `git rm`'d, `changelog.d/README.md` is kept, and the numbered fragments render as `([#4513](https://github.com/kornia/kornia/pull/4513))` ahead of the orphan entries.
- **Name validation.** `fix-thing.fixed.md` is rejected (`Issue name 'fix-thing' does not match the configured pattern, '[0-9]+'`); `+fix-thing.fixed.md` is accepted as an orphan, matching what the README and CONTRIBUTING say to do before a PR number exists.
- **Claimed checks reproduce.** `pixi run pre-commit-all` (all hooks pass, tree clean afterwards), `pixi run toml-fmt-check`, `tests/scripts/test_check_import_surface.py` (77 passed).
- **No stragglers.** `grep -rn CHANGELOG` over `.github/`, `docs/`, `tests/`, `*.toml`, `*.md` finds only the references this PR rewrote; no workflow or docs page still reads or path-filters on `CHANGELOG.md`.

Follow-ups. None blocks the merge; 1 and 2 are the ones I would do before the first release.

1. **`changelog-build` needs `--yes`, or a documented recovery.** Towncrier writes and stages `CHANGELOG.md` *before* the "Is it okay if I remove those files?" prompt. Verified: `towncrier build --version 0.9.0 </dev/null` exits with `Aborted!`, leaves `M  CHANGELOG.md` staged plus all 177 fragments on disk, and a rerun refuses with `ValueError: It seems you've already produced newsfiles for this version?`. Any non-interactive run (CI, a script, an agent shell) lands there. Since step 1 of the README already has the releaser review the preview, baking `--yes` into the pixi task is the simplest fix; otherwise document `git restore --staged --worktree CHANGELOG.md` as the recovery and tell the releaser to review with `git diff --cached`.
2. **25 open PRs edit `CHANGELOG.md` today** (#4512, #4511, #4510, #4509, #4507, #4505, #4500, #4495, #4487, #4485, #4484, #4475, #4471, #4465, #4456, #4455, #4389, #4379, #4375, #4371, #4165, #4034, #3968, #3945, #3931). Their hunks sit inside the region this PR deletes, so each becomes CONFLICTING the moment this merges, and a conflicting PR runs no CI. Each needs the same one-time edit: drop the `CHANGELOG.md` hunk, add `changelog.d/<PR>.<type>.md` with the same text. Worth one paragraph in `changelog.d/README.md`, since the PR body is the only place that says it today. One trap to call out for whoever does the sweep: a local `merge=union` attribute on `CHANGELOG.md` (which this checkout has in `.git/info/attributes`) makes a keep-both merge "succeed" by resurrecting the whole 1527-line block under the marker, and the next release would then print every entry twice. Retire that attribute the day this lands.
3. **`ignore = []` is load-bearing, not a no-op.** Towncrier's strict fragment-name mode is `config.ignore is not None`, so the empty list is what makes `4520.bugfix.md` an error instead of a silently dropped file. Verified both ways. Keep it, but a Finder-created `changelog.d/.DS_Store` (gitignored, invisible to `git status`) aborts both tasks with `Invalid news fragment name: .DS_Store`. Suggest `ignore = [".DS_Store"]` plus a comment saying the key enables strict mode.
4. **`package = ""` does not disable the version lookup.** Towncrier substitutes `[project].name`, so `towncrier build --draft` without `--version` tries `import kornia` from the source tree inside the isolated tool env and dies with `No module named 'torch'` instead of the usage error. In an env where kornia imports it would take `kornia.__version__` (`0.9.0rc1`) and, for a non-draft build, consume every fragment under that heading. The `name` key and the three `showcontent = true` lines are likewise inert (byte-identical draft without them). Drop the dead keys and keep `--version` in the task description as the contract.
5. **Nothing enforces fragment presence or naming.** No workflow or pre-commit hook runs towncrier, so a `+name` fragment nobody renamed, or a `4500.add.md`, merges green and only fails when the releaser runs the build. `towncrier check --compare-with origin/main` works with this config (checked), so a CI job in the shape of `import-surface.yml` with a `no-changelog` label bypass would close the gap. Note `check` skips itself whenever `CHANGELOG.md` is in the diff and does not validate names, so pair it with a draft build. Follow-up, since the policy already allows omitting a fragment.
6. **Doc nits.** The PR template names only `<PR>.<type>.md` and omits the `+name.<type>.md` fallback; CONTRIBUTING says rename "after opening the PR" while the README says "before merging"; neither says how to credit a multi-PR entry (30 migrated entries cite two or three PRs) or that identical text under two numbers merges into one `(#a, #b)` bullet, or that the filename must be a PR number since `issue_format` links to `/pull/`. The permanent `## Unreleased` heading above the marker will read as an always-empty section on GitHub after the first release; folding the pointer into the intro paragraph avoids that. Migrated entries keep plain `(#NNNN)` while new ones get links, so the next release section mixes styles once.

ducha-aiki:
## Re-review of a5bf48db: approve

Re-read the whole diff against `main`, not only the delta. Every follow-up from the first review is addressed, and the new CI policy check is sound.

Verified locally:

- **Migration still faithful.** 177 fragments render to 177 bullets; per category Added 24/24, Bug fixes 133/133, Breaking changes 20/20. The five entries that now differ from `main` are exactly the five reworded fragments, which only replace "above"/"below" cross-references with the category name (correct, since towncrier reorders).
- **Non-interactive release build works.** `pixi run changelog-build --version 0.9.0 </dev/null` in a throwaway worktree: exits 0, no prompt, stages the dated section plus 177 fragment deletions, keeps the README and marker. The header now folds the pointer above `****` with no `## Unreleased` heading, so the first release will not leave an empty section behind.
- **Strict naming kept, Finder tolerated.** With `ignore = [".DS_Store"]`, a stray `changelog.d/.DS_Store` no longer aborts the preview, while `4520.bugfix.md` is still rejected (`Invalid news fragment name`).
- **Policy script behaves as the PR itself requires.** From this checkout with `--base-ref origin/main --pr-number 4513`: without the label flag it reports both "no changed changelog.d/4513.… fragment" and "ordinary PRs must not edit CHANGELOG.md" and exits 1; with `--skip-presence` it prints "Changelog fragments are valid." The PR carries the `no-changelog` label, which exists on the repo, so the new job should pass once GitHub's queue drains (it was still `queued` when I posted this).
- **Tests and hooks.** `tests/scripts/test_check_changelog.py` (11) plus `test_check_import_surface.py` (77) pass, `pixi run pre-commit-all` passes on a clean tree, and ruff 0.16.6 is clean on the two new Python files. The workflow uses the same `checkout@v7` / `setup-python@v7` / `setup-pixi@v0.10.2` / `pixi v0.40.0` as the other 20 workflows.
- **Dead config gone.** `package`, `name`, and the three `showcontent = true` lines are removed; the draft is unchanged.

Two small notes, neither worth another round:

- A PR that only fixes a typo in *another* PR's fragment fails the presence check and needs the label. That follows from "a fragment named for its own number" and is documented well enough, just something for reviewers to expect.
- `.DS_Store` is now listed twice, in `[tool.towncrier].ignore` and in the script's `IGNORED_FILES`. If one list grows, the other has to follow.

Issue #4503, opened 2026-09-13
normal_transform_pixel3d and normalize_homography3d have no align_corners, so warp_affine3d(align_corners=False) is wrong (3D half of #3904)


VanniLeonardo:
### What happened?

This is the 3D half of #3904, which #3945 fixes for 2D.

`normal_transform_pixel3d` and `normalize_homography3d` hardcode the corner-aligned normalization, where voxel centres `0` and `size - 1` map to `-1` and `+1`. Neither takes an `align_corners` argument. `warp_affine3d` then hands the matrix built that way to `grid_sample(align_corners=...)`, which reads `-1` and `+1` as the outer voxel edges when the flag is `False`. Mixing the two conventions applies a spurious sub-pixel scale and shift.

A whole-voxel translation is the clearest case, because it should move the volume by exactly one voxel under either convention. At `align_corners=False` it does not.

This is the same shape of defect as the 2D one, and the 2D figure on current `main` is in the reproduction below for comparison. #3945 threads `align_corners` through `create_meshgrid`, `normal_transform_pixel`, `normalize_homography` and `denormalize_homography`, but deliberately leaves the 3D functions alone, because the 3D test expectations need their own review.

### Minimal reproduction

```python
import inspect, torch
from kornia.geometry.conversions import normal_transform_pixel, normal_transform_pixel3d, normalize_homography3d
from kornia.geometry.transform import warp_affine3d, warp_affine

print("2D normal_transform_pixel  :", inspect.signature(normal_transform_pixel))
print("3D normal_transform_pixel3d:", inspect.signature(normal_transform_pixel3d))
print("3D normalize_homography3d  :", inspect.signature(normalize_homography3d))

d, h, w = 2, 3, 4
vol = torch.arange(float(d * h * w)).view(1, 1, d, h, w)
M = torch.eye(4)[None][:, :3].clone()
M[:, 0, 3] = 1.0                                   # shift by +1 voxel along x
expected = torch.zeros_like(vol)
expected[..., 1:] = vol[..., :-1]
for ac in (True, False):
    out = warp_affine3d(vol, M, (d, h, w), align_corners=ac)
    print(f"warp_affine3d  +1 voxel, align_corners={ac!s:5s} max|out - expected| = {(out - expected).abs().max():.6f}")

img = torch.arange(12.0).view(1, 1, 3, 4)
M2 = torch.tensor([[[1.0, 0.0, 1.0], [0.0, 1.0, 0.0]]])
[...]

PR #4514, opened 2026-09-14
fix(geometry): reorder the 3D sampling grid to (x, y, z) in homography_warp3d

### LINKED ISSUE #4502: homography_warp3d feeds a (d, x, y) grid to grid_sample, which reads (x, y, z), so warp_perspective3d fails on an identity warp
### What happened?

`create_meshgrid3d` emits its channels in `(z, x, y)` order, but `torch.nn.functional.grid_sample` reads a 5D grid's last dimension as `(x, y, z)`. `homography_warp3d` passes the grid straight to `grid_sample`, so the two disagree and every warp built on that path samples the wrong voxel.

The effect is not subtle and it is not an `align_corners` question. An identity homography on a 4x4x4 cube comes back with a maximum deviation of `45.0` at `align_corners=True` and `55.125` at `False`, on an input whose values run 0 to 63.

`warp_perspective3d` and `homography_warp3d` are affected. `warp_affine3d` is not, because it builds its grid with `F.affine_grid`, which already produces `(x, y, z)`.

The 2D `create_meshgrid` is correct, so this is a mismatch between the two functions rather than a library-wide convention. Feeding `grid_sample` a hand-built `(x, y, z)` grid with the same normalization reproduces the input exactly under both `align_corners` settings, which isolates the channel order as the whole of the defect.

Nothing pins this today. `warp_perspective3d`'s only numerical test asserts that an unbatched matrix gives the same answer as a batched one, and both are wrong in the same way, so it passes.

### Minimal reproduction

```python
import torch
import torch.nn.functional as F
from kornia.geometry import create_meshgrid, create_meshgrid3d
from kornia.geometry.transform import warp_perspective3d, warp_affine3d, homography_warp3d

# 1. the channel order create_meshgrid3d emits, against the 2D convention
g3 = create_meshgrid3d(2, 3, 4, normalized_coordinates=False)   # (1, D, H, W, 3)
g2 = create_meshgrid(3, 4, normalized_coordinates=False)        # (1, H, W, 2)
print("2D create_meshgrid  [h=2, w=3] ->", g2[0, 2, 3].tolist())
print("3D create_meshgrid3d[d=1, h=2, w=3] ->", g3[0, 1, 2, 3].tolist())

# 2. what that costs: an identity warp on a 4x4x4 cube
n = 4
vol = torch.arange(float(n ** 3)).view(1, 1, n, n, n)
I4 = torch.eye(4)[None]
for ac in (True, False):
    p = warp_perspective3d(vol, I4, (n, n, n), align_corners=ac)
    a = warp_affine3d(vol, I4[:, :3], (n, n, n), align_corners=ac)
    h = homography_warp3d(vol, I4, (n, n, n), align_corners=ac)
    print(f"align_corners={ac!s:5s}  warp_perspective3d {(p - vol).abs().max():9.4f}"
          f"   homography_warp3d {(h - vol).abs().max():9.4f}   warp_affine3d {(a - vol).abs().max():.3e}")

# 3. a hand-built (x, y, z) grid, same normalization, is exact
zs, ys, xs = torch.meshgrid(*[torch.arange(float(n))] * 3, indexing="ij")
for ac in (True, False):
    if ac:
        grid = torch.stack([2 * xs / (n - 1) - 1, 2 * ys / (n - 1) - 1, 2 * zs / (n - 1) - 1], -1)[None]
    else:
        grid = torch.stack([(2 * xs + 1) / n - 1, (2 * ys + 1) / n - 1, (2 * zs + 1) / n - 1], -1)[None]
    out = F.grid_sample(vol, grid, align_corners=ac)
    print(f"hand-built (x, y, z) grid, align_corners={ac!s:5s} -> {(out - vol).abs().max():.4f}")
```

Observed:

```
2D create_meshgrid  [h=2, w=3] -> [3.0, 2.0]
3D create_meshgrid3d[d=1, h=2, w=3] -> [1.0, 3.0, 2.0]     grid_sample wants [3.0, 2.0, 1.0]
align_corners=True   warp_perspective3d   45.0000   homography_warp3d   45.0000   warp_affine3d 3.815e-06
align_corners=False  warp_perspective3d   55.1250   homography_warp3d   55.1250   warp_affine3d 0.000e+00
hand-built (x, y, z) grid, align_corners=True  -> 0.0000
hand-built (x, y, z) grid, align_corners=False -> 0.0000
```

Expected: an identity homography reproduces the input, and `warp_perspective3d` agrees with `warp_affine3d` given the same identity, exactly as the 2D pair do.

### Environment

- Kornia version: 0.9.0rc1 (current `main`)
- Python version: 3.11.14
- OS / device: Linux, CPU

### Anything else?

Whichever side is changed, the fix needs a sweep of the `create_meshgrid3d` callers, since anything that consumes the grid as `(z, x, y)` today would have to move with it.

This is separate from the `align_corners` defect in the 3D normalization, which I am filing alongside it. That one affects `warp_affine3d`; this one does not.

Found while verifying the 3D follow-up to #3904 that came out of the #3945 review. Investigated with Claude Code; every figure above is from an executed run on current `main`.

VanniLeonardo:

`homography_warp3d`, and therefore `warp_perspective3d`, sampled every voxel from the wrong place. `create_meshgrid3d` produces its channels in kornia's documented `(d, x, y)` order, which `normalize_pixel_coordinates3d` and `conv_soft_argmax3d` depend on. `homography_warp3d` handed that grid straight to `warp_grid3d` and `torch.nn.functional.grid_sample`, both of which read the last axis as `(x, y, z)`. Nothing converted between the two.

On `main`, an identity homography on a 4x4x4 `arange` volume deviates by 45.0 at `align_corners=True` and 55.125 at `False`, and a +1 translation along `x` moves slices instead of columns. The only numerical test compared an unbatched call against a batched one, and both were wrong the same way.

The fix is one line at that call site: reorder the grid to `(x, y, z)` before the homography is applied. `create_meshgrid3d` and its other consumers are untouched, so their public order does not change. `warp_affine3d` builds its grid with `F.affine_grid` and was never affected.

After the change an identity warp is exact at `align_corners=True` on cubes, non-cubes and singleton axes, a +1 translation along each of `x`, `y`, `z` moves exactly the right axis, and `warp_perspective3d` agrees with `warp_affine3d` given the same affine matrix. The `align_corners=False` identity still deviates. That residual is the 3D normalization convention tracked in #4503, and it is pinned here as a strict `xfail` so it turns loud when #4503 lands.

| dsize | `main`, `True` | this PR, `True` | this PR, `False` (#4503) |
|---|---|---|---|
| (4, 4, 4) | 45.0 | 0.0 | 55.125 |
| (2, 3, 4) | 9.0 | 0.0 | 20.125 |
| (1, 5, 3) | 10.5 | 0.0 | 10.5 |

The `homography_warp3d` Convention block now states the `(x, y, z)` order the homography acts in and that the conversion happens there.

Fixes #4502. Unblocks verifying #4503 through `warp_perspective3d`.

## What did you check?

```text
$ pytest tests/geometry/transform/test_imgwarp3d.py --dtype=float32,float64
127 passed, 6 skipped, 4 xfailed
$ pytest tests/geometry/transform/test_imgwarp3d.py --dtype=float16  --xfail-known-failures --known-failure-profile=cpu-float16
59 passed, 3 skipped, 9 xfailed
$ pytest tests/geometry/transform/test_imgwarp3d.py --dtype=bfloat16 --xfail-known-failures --known-failure-profile=cpu-bfloat16
62 passed, 3 skipped, 6 xfailed
$ pytest tests/ --dtype=float32
10417 passed, 3735 skipped, 45 xfailed, 11 xpassed
$ pytest tests/ --dtype=float16  --verify-known-failures --known-failure-profile=cpu-float16
9814 passed, 3819 skipped, 570 xfailed, exit 0
$ pytest tests/ --dtype=bfloat16 --verify-known-failures --known-failure-profile=cpu-bfloat16
9868 passed, 3772 skipped, 563 xfailed, exit 0
$ pytest --doctest-modules kornia/geometry/transform/imgwarp.py
9 passed
$ python docs/export_support/cases_geomA.py out.json transform.warp_perspective3d transform.homography_warp3d
transform.warp_perspective3d  ok      transform.homography_warp3d  ok
```

`torch.compile(fullgraph=True)` and `torch.jit.trace` of `warp_perspective3d` both match eager bit for bit, with zero graph breaks. Pre-commit passes on the changed files; `pixi run typecheck` reports only the pre-existing diagnostic in `kornia/feature/lightglue.py`.

Mutation check: with the reorder line removed, 11 of the 13 new `TestWarpPerspective3d` nodes fail at float32 and the two #4503 xfails stay xfail; the two that still pass are the projective-row and gradcheck tests, which do not depend on the order.

Two fixture notes, so nobody has to rediscover them. The agreement and translation tests use sizes that are one more than a power of two, so the corner-aligned scale `2 / (size - 1)` is exact and the two grid routes agree bit for bit at float16 and bfloat16 too; a size-4 axis makes them round differently by up to 0.023 at bfloat16, which says nothing about channel order. The agreement test also avoids singleton axes because torch's CPU `F.affine_grid` has no half-precision kernel for a size-1 dimension, and that is the reference, not the subject.

## How did AI help?

Claude Code did the investigation, the change, the tests and this description. I directed the work, reviewed every line, and every figure above comes from a run I executed on this branch.


ducha-aiki (approved):
Review of `0b42422f8b92fa03621aa0e23742ce28c1ac545a`.

No actionable findings. Reordering the `create_meshgrid3d` output at `kornia/geometry/transform/imgwarp.py:1830` correctly converts Kornia's `(d, x, y)` convention to the `(x, y, z)` convention consumed by `warp_grid3d` and 5-D `grid_sample`, while leaving the shared grid helper and its other consumers unchanged. The tests cover identity on non-cubic and singleton volumes, all three translation axes, an affine oracle, a projective matrix, the functional entry point, and gradients. The release note is correctly carried by `changelog.d/4514.fixed.md`.

Validation and limits: Current head is mergeable. Focused CPU float32/float64 run from the PR worktree imported `/private/tmp/kornia-triage-60Oa1g/wt-4514/kornia/__init__.py`: 127 passed, 6 skipped, 4 expected xfails. GitHub has 38 successful checks; the only test failure is the unrelated macOS torch-2.5.1 `RRDBNetBuilder` ONNX timeout, with collector failing only as its aggregate.

PR #4515, opened 2026-09-14
test: skip macOS RRDBNet export and prefetch reference data

ducha-aiki:
The RRDBNet builder's ONNX export exceeds its 120-second test timeout on the macOS runners under torch 2.5.1. The cost is the TorchScript exporter torch < 2.9 defaults to, not the platform: on one Apple Silicon machine the `RealESRGAN_x4plus_anime_6B` export takes 6.4 s under torch 2.14 (dynamo exporter) and 131.9 s under 2.5.1, and every CI timeout was on the `macos-latest × 2.5.1` legs while the macOS 2.14 and MPS legs passed. The parameter is therefore skipped only on darwin with `torch_version_lt(2, 9, 0)`, retaining SmallSR export coverage everywhere and RRDBNet coverage on every other leg.

The `data` fixture's reference tensors were missing from the CI prefetch, so matrix jobs downloaded them live through GitHub blob redirects without retries. Prefetch the five distinct reference files, use commit-pinned raw URLs, and route fixture downloads through Kornia's existing retry helper. The URL table moves out of `conftest.py` into `testing/reference_data.py` (public `TEST_DATA_URLS`) and is enumerated in `WEIGHT_REGISTRIES`, so the existing two-way prefetch guard in `tests/core/test_weights_prefetch.py` covers it without a dedicated test; the `dexined` entry reuses the library's own source list. `tests/tracking/test_planar_tracking.py` now reads the fixture instead of an unpinned `blob/main` URL that aliased the prefetched cache entry, and a new test holds `docs/generate_examples.py`'s copy of the `data_test` commit hash to the table.

Validation on macOS, Python 3.11, CPU float32:
- torch 2.14: super-resolution module 14 passed (RRDBNet export runs, 5.6 s); torch 2.5.1: the RRDBNet parameter is skipped with the exporter reason.
- `tests/core/test_weights_prefetch.py`, `tests/models/test_dexined.py`, `tests/test_conftest.py`: 110 passed, 1 skipped. The docs-hash test was checked to fail on a drifted hash.
- `tests/tracking/test_planar_tracking.py` with slow tests: 14 passed. Fixture-consuming matcher tests with slow tests: 11 passed.
- All five reference files prefetched by the script's exact call into a fresh cache and resolved by the fixture under the same names.
- `pixi run pre-commit-all`, `git diff --check` and `pixi run typecheck` (one pre-existing deprecation warning in `kornia/feature/lightglue.py`) passed.

Linux/Windows CI has not been run locally. No changelog fragment: this change affects tests and CI only.

Posted on behalf of [@ducha-aiki](https://github.com/ducha-aiki) by Codex (gpt-6); revised on behalf of @ducha-aiki by Claude (Fable 5.1).


ducha-aiki:
## Review of 7ba85190

**Verdict: one change requested.** The prefetch half is correct and verified; the macOS skip is gated on the wrong axis and drops coverage from four legs that pass today.

### P1: the skip should key on the exporter, not the platform

Every `test_to_onnx_exports[RRDBNetBuilder]` timeout I could find is on `macos-latest × torch 2.5.1` (py3.11 and py3.12): 8 of the 12 most recent failed `pr_test_cpu` runs plus the two scheduled failures (34804576631, 34779403220). The macOS torch 2.14 legs and the MPS leg (macos-15, 2.14) pass in the same runs.

Measured on one Apple Silicon machine, same checkout, same `to_onnx(save=False)` call:

| torch | `torch.onnx.export` default | RRDBNet anime_6B export |
|---|---|---|
| 2.14.0 | `dynamo=True` | 6.4 s |
| 2.5.1 | `dynamo=False` (TorchScript exporter) | 131.9 s |

`to_onnx` never passes `dynamo`, so the 2.5.1 leg walks the TorchScript exporter and the 120 s budget is what it overruns. `sys.platform == "darwin"` therefore skips the case on mac 2.14 × {3.11, 3.12, 3.13} and on the MPS leg, all of which currently exercise it in ~6 s, and the reason string blames the platform for what the exporter does.

Suggested gate (the dynamo default flipped in torch 2.9):

```python
from kornia.core._compat import torch_version_lt
...
marks=pytest.mark.skipif(
    sys.platform == "darwin" and torch_version_lt(2, 9, 0),
    reason="TorchScript exporter (torch < 2.9) exceeds the 120-second timeout on macOS runners",
),
```

Keep the darwin term: the Linux and Windows 2.5.1 legs pass, and CI logs carry no `--durations`, so I could not measure their margin. Either way please adjust the reason string to name the exporter; the comment block above the test already explains the node-count cost, and this is the same mechanism.

### P2: the "HTTP 503 setup errors" motivation is not in CI

The body says the `data` fixture failed with 503s. I grepped the failed jobs of the 12 most recent failed `pr_test_cpu` runs and the two scheduled failures for `HTTP Error 503`: zero hits. The only network setup errors in that window are `tests/io/test_io_image.py` (DNS `Errno 8` on an MPS leg in run 34824678526, a read timeout on a mac 2.5.1 leg in run 34793842309), a different URL and outside this PR. The prefetch change stands on its own (below), but please link the run that showed the 503s so the motivation is checkable. Not blocking.

### Verified

- Prefetch path: ran the script's exact call, `load_state_dict_from_url(url, map_location=cpu, file_name=name)`, for the five new entries into a fresh hub dir. All five load as dicts and land under `checkpoints/<name>`; every `_TEST_DATA_URLS` entry (minus `dexined`) then resolves to a cached file by basename, which is what `torch.hub.load_state_dict_from_url` looks up. `conftest.py` pins `torch.hub.set_dir("weights/")`, matching the script's `-t weights/` and the restored cache path, and the cache key hashes the script so this PR forces a fresh prefetch.
- The five raw URLs return 200 (2.2 MB to 4.3 MB each).
- `tests/core/test_weights_prefetch.py`: 92 passed locally, including the 7 new `test_fixture_data_is_prefetched` cases and the orphan check.
- Docs note: `docs/generate_examples.py` fetches `knchurch_disk.pt` at the same sha via the `github.com/.../raw/` spelling; same basename, so the docs build hits the prefetched file. Docstring update is accurate.
- `from conftest import _TEST_DATA_URLS` follows the existing pattern in `tests/test_conftest.py` and `tests/test_half_precision_ci.py`.

Unrelated aside: scheduled run 34804576631 also failed `tests/feature/test_sosnet.py::TestSOSNet::test_jit[cpu-float64]` on ubuntu / py3.11 / torch 2.6.0. Not this PR.

ducha-aiki:
Applied in 633951bd7, on top of 7ba85190.

**From my review above**

- The skip is now `sys.platform == "darwin" and torch_version_lt(2, 9, 0)`, with the reason naming the TorchScript exporter. Checked in-process on this Mac: torch 2.14 runs the case (5.6 s), torch 2.5.1 skips it with that reason.

**From the Copilot review**

- *Importing `conftest` as a module* and *the leading underscore*: applied together. The table now lives in `testing/reference_data.py` as `TEST_DATA_URLS` (plus `DATA_TEST_SHA`); `conftest.py` and the guard test both import it from there. The `conftest` import did follow two existing precedents (`tests/test_conftest.py`, `tests/test_half_precision_ci.py`) and resolves to the one `sys.modules["conftest"]` pytest already loaded, so the "duplicate module state" concern was theoretical, but an ordinary module is the cleaner home either way.
- *`_load_prefetch_script()` re-executes the script per parameter*: not applied, it is already wrapped in `functools.lru_cache(maxsize=1)` (`tests/core/test_weights_prefetch.py:93`) and its docstring says so. The parametrized test the comment worried about is gone anyway (next item).

**From a second pass over the staged tree**

- The fixture table is enumerated in `WEIGHT_REGISTRIES`, so `test_checkpoint_is_prefetched_or_exempt`, `test_prefetched_urls_match_the_library` and the orphan check cover it without the dedicated `test_fixture_data_is_prefetched` and its `dexined` special case. `dexined` now reuses `kornia.filters.dexined.url`, so the fixture shares the library's cache entry, fallback mirror and guard instead of spelling a third URL for the same file.
- `tests/tracking/test_planar_tracking.py` fetched `loftr_outdoor_and_homography_data.pt` from an unpinned `blob/main` URL through bare `torch.hub`. With the prefetch in place that call would have been served the pinned bytes from cache by basename, silently. It now uses the `data` fixture (`loftr_homo`) and casts into a copy, since the fixture is session-scoped and `test_real` rebinds `image0`/`image1`.
- `docs/generate_examples.py` carries its own copy of the `disk` commit hash. A new test, `test_docs_reference_data_matches_the_fixture_table`, resolves every `data_test` reference in that script (literal or f-string field) and holds it to the table; verified to fail when the hash is altered.

Validation: `pixi run pre-commit-all`, `git diff --check`, `pixi run typecheck` (one pre-existing deprecation warning in `lightglue.py`); `tests/core/test_weights_prefetch.py` + `tests/models/test_dexined.py` + `tests/test_conftest.py` 110 passed / 1 skipped; `tests/tracking/test_planar_tracking.py` with slow tests 14 passed; fixture-consuming matcher tests 11 passed; `tests/contrib/test_super_resolution.py` 14 passed on torch 2.14. PR body updated to match.

Issue #4426, opened 2026-09-11
W1b: `set_rng_device_and_dtype` moves the `p` gate, not the sampled augmentation parameters


ducha-aiki:
`augmentation.base.rst` ("Random Reproducibility") and `_BasicAugmentationBase`'s own docstring (`kornia/augmentation/base.py:55-56`) say that random parameters are sampled on the CPU with `torch.get_default_dtype()` independently of the input's device, and then: *"To change this behaviour, use `set_rng_device_and_dtype`."* The first half is correct. The second half names a method that does not do it.

`set_rng_device_and_dtype` reaches the generator — the generator reports the new device and dtype afterwards — but on most classes the only entry in `_params` that actually changes is `batch_prob`, the Bernoulli draw behind `p` / `p_batch`. The sampled augmentation parameters keep coming out on CPU in `float32`: after `set_rng_device_and_dtype(torch.device('cpu'), torch.float64)`, `RandomAffine`'s `angle`, `scale`, `shear_x`, `shear_y`, `translations` and `center` are all still `torch.float32` while `batch_prob` is `torch.float64`. The device half behaves the same way on an accelerator.

The escape hatch that does work is the one the documentation does not mention: `torch.set_default_dtype` **before constructing** the augmentation, which is read once at construction time and applies to every key. A caller cannot even rely on `set_rng_device_and_dtype` being a harmless no-op: a minority of classes do move some of their generator keys, and a handful raise outright when asked to move to an accelerator — `RandomShear` fails inside `make_samplers`, and `RandomLinearIllumination`, `RandomLinearCornerIllumination` and `RandomRain` fail in their own ways. Moving the module with `module.to(device)` is a third, different behavior again: it moves whatever generator buffers a class happens to have registered, which is not the same set.

## Repro

```python
import torch
import kornia.augmentation as K

def key_dtypes(aug, shape=(4, 3, 6, 8)):
    torch.manual_seed(0)
    p = aug.forward_parameters(torch.Size(shape))
    return {k: str(v.dtype) for k, v in p.items() if torch.is_tensor(v)}

a = K.RandomAffine(degrees=(45.0, 45.0), p=1.0)
a.set_rng_device_and_dtype(torch.device("cpu"), torch.float64)
print("generator reports        :", a._param_generator.device, a._param_generator.dtype)
print("RandomAffine key dtypes  :", key_dtypes(a))

r = K.RandomRotation(degrees=(45.0, 45.0), p=1.0)
[...]

Issue #3913, opened 2026-08-08
torch.compile of RandomPerspective: CUDA illegal memory access on L4; GuardOnDataDependentSymNode on MPS


ducha-aiki:
Found by the Stage-2 W3 benchmark suites (#3911 / PR #3912) while benchmarking augmentations through the class API (parameter sampling included in the compiled graph).

**Symptoms (same root cause, different failure depth):**
- **MPS + CPU (Apple Silicon, torch 2.9.1):** `torch.compile(RandomPerspective(0.5, p=1.0))` warmup fails with `GuardOnDataDependentSymNode` — dynamo cannot guard on runtime-sampled perspective parameters. Reported cleanly by the harness NOTE.
- **CUDA (NVIDIA L4, Debian 10):** compilation *succeeds* but the generated kernel triggers `CUDA error: an illegal memory access was encountered` during the warmup call — inductor emits out-of-bounds indexing for the data-dependent parameter graph. The fault is asynchronous and poisons the CUDA context; before the harness learned to drain the stream inside the warmup try-block it surfaced two ops later at an innocent `RandomResizedCrop.to(device)`. Exact torch/CUDA versions to be posted from the box in the PR #3912 thread.

**Context:** the geometry flagship (`warp_perspective` as a pure functional op, static shapes/params) compiles and runs fine on the same box — the failure is specific to compiling the augmentation forward, where `forward_parameters` sampling and value-dependent transform math are part of the traced graph. `RandomResizedCrop` (both devices) and `ColorJiggle` (InductorError) also fail to compile; RandomPerspective is the only one observed to miscompile into a faulting kernel.

**Repro:**
```
python benchmarks/augmentation/flagship.py --batches 1 --size 256 --device cuda --compile
# harness exits: FATAL: CUDA context poisoned during torch.compile warmup of 'RandomPerspective' ...
# workaround: --skip-compile-ops RandomPerspective
```

**Why it matters (agent-era plan):** S5 intends to certify/advertise torch.compile cleanliness for core ops — that claim must exclude (or fix) the augmentation param-generation path first. Candidate Stage-3 fix direction: make the parameter generators compile-safe (clamped/masked tensor ops instead of data-dependent indexing), which would fix all three ops at once. Likely also worth a minimal upstream pytorch/inductor repro for the L4 miscompile once versions are confirmed.

ducha-aiki:
[...]

PR #4517, opened 2026-09-14
Fix CUDA compilation with CPU geometric augmentation parameters

### LINKED ISSUE #4516: RandomAffine and RandomPerspective fail CUDA compilation with numeric constructor arguments
`torch.compile` fails for ordinary `RandomAffine` and `RandomPerspective` module calls on CUDA when they are constructed with Python-valued arguments and moved with `.to("cuda")`. This affects library users directly, independently of the benchmark harness. These existing constructors and device-migration calls should work without requiring users to replace Python values with CUDA tensors.

Reproduced on Kornia `3c37d390` / `0.9.0rc1`, PyTorch `2.14.0+cu130` (commit `08187d9e0fba026dc8217405802ab5381dc88d90`), Python 3.13.12, Linux/WSL2, NVIDIA RTX 4090, CUDA 13.0, driver 595.95.

### Reproduction

Run each transform in a fresh process because the perspective failure poisons the CUDA context:

```python
import torch

torch.set_num_threads(4)
x = torch.rand(1, 3, 256, 256, device="cuda")
aug = K.RandomAffine(degrees=30.0, translate=(0.1, 0.1), scale=(0.8, 1.2), p=1.0).to("cuda")
# In a separate process, substitute:
# aug = K.RandomPerspective(0.5, p=1.0).to("cuda")
y = torch.compile(aug)(x)
torch.cuda.synchronize()
```

With `CUDA_LAUNCH_BLOCKING=1`, affine raises `RuntimeError: CUDA driver error: invalid argument` / `CUDA_ERROR_INVALID_VALUE`; perspective raises an illegal-memory-access error. Affine also reproduces at batches 8, 32 and 128. Eager execution and the `eager` and `aot_eager` compile backends succeed.

The failure was initially exposed by this benchmark command, which calls the same public augmentation modules:

```bash
python benchmarks/augmentation/flagship.py --batches 1,8,32,128 --size 256 --device cuda --compile --json aug_cuda.json
```

### Diagnosis

The samplers move to CUDA, but these generators use `_extract_device_dtype` on constructor ranges to choose the returned parameter placement. With only Python values this defaults to CPU. Affine therefore returns its center and other parameters on CPU; perspective returns both corner tensors on CPU. The augmentation then moves them to the image device.

On the tested PyTorch build, Inductor mishandles this mixed-device graph: the generated wrapper allocates a CPU buffer and passes that buffer directly to a Triton CUDA kernel without a host-to-device copy. Affine fails on its center transform; perspective fails on corner-coordinate processing. This is not merely a graph break or unsupported random sampling.

This smaller PyTorch-only reproduction also fails with the affine driver error:

```python
import torch

def f(x):
    center = torch.tensor([256, 256], dtype=torch.float32).view(1, 2) / 2.0 - 0.5
    center = center.expand(x.shape[0], -1)
    matrix = torch.eye(3, device=x.device).repeat(x.shape[0], 1, 1)
    matrix[:, :2, 2] = center.to(x)
    return matrix, center

x = torch.rand(1, 3, 256, 256, device="cuda")
torch.compile(f)(x)
torch.cuda.synchronize()
```

Upstream report for the PyTorch-only reproduction: https://github.com/pytorch/pytorch/issues/196969.

### Verified workaround and Kornia fix

Passing `degrees=torch.tensor(30.0, device="cuda")` or `distortion_scale=torch.tensor(0.5, device="cuda")` makes the respective transform compile. Both passed batches 1, 8, 32 and 128 at 256×256, with two calls per configuration and maximum absolute error below 1.3e-4 against eager replay of the compiled parameters.

Fix PR: https://github.com/kornia/kornia/pull/4517. For Python-valued constructor ranges, the two generators return parameters on their configured RNG device, so the existing `.to("cuda")` call avoids the problematic transfers automatically. The PR preserves explicit tensor-valued range placement and the existing returned dtype policy, and tests end-to-end compilation with fresh sampling and parameter replay. Broader RNG migration and precision inconsistencies remain separate.

Related: #3913 (earlier perspective compile failures on CUDA/MPS) and #4426 (sampler versus returned parameter placement). The diagnosis here is measured on the stack above and does not establish the cause of every historical backend failure in #3913.

Written by Codex (gpt-6) on behalf of @ducha-aiki

ducha-aiki:
`torch.compile(RandomAffine(...))(cuda_images)` and `RandomPerspective` can fail when Inductor reuses a CPU geometry constant in a CUDA kernel without transferring it. The generators now construct those constants by stacking scalar factories, preserving numeric and tensor-range placement, returned precision, and RNG consumption. Both transforms work with moved or unmoved samplers, CPU tensor ranges, and an ambient CUDA default device.

Module migration now validates `.to()` arguments before changing augmentation samplers. Random generators use `_apply` so direct `.to()`, convenience methods, and container moves update their registered buffers and samplers together. Omitted device/dtype arguments preserve their current values, invalid integer-dtype requests leave samplers untouched, and CUDA device canonicalization no longer rebuilds samplers twice. Plain tensor constructor attributes retain their existing placement precedence.

Closes #4516. Related: #4426 and #3913. The upstream compiler defect remains tracked in https://github.com/pytorch/pytorch/issues/196969. This compiler workaround covers **RandomAffine and RandomPerspective**; RandomShear, RandomThinPlateSpline, RandomErasing, CenterCrop, and other generators are outside its scope.

Validation through Pixi tasks on an RTX 4090, Python 3.13.12 / PyTorch 2.14.0+cu130:

- **5,081 passed**, with 4,226 skips and 16 expected failures, across focused CPU/CUDA float32/float64 augmentation base, 2D/3D generator, affine/perspective, and backward suites.
- **168 fullgraph cases passed** across CPU/CUDA, float32/float64, batches 1/4, affine and both perspective methods, seven placement/migration configurations, and two fresh calls per case. Sampling and application are captured together; explicit parameter outputs support eager replay even when older Torch suppresses `_params` state updates. The earlier direct-module version of this matrix also passed all 168 cases.
- A CPU FX guard rejects lifted geometry constants and indexed scalar fills. It catches the original tensor-constant implementation and the pre-review indexed-fill variant. The explicit-sampling test wrapper still reproduces the original CUDA failure against the merge base.
- **384 exact A/B parameter-and-RNG comparisons** against merge base `3c37d390`, including CPU/CUDA, all four floating dtypes, empty/shared batches, and half-precision coordinates at 257×2049.
- CPU float16 and bfloat16 profiles: **294 passed and 2 expected failures each**. Focused ONNX export/runtime: **4 passed**.
- Full pre-commit, typecheck, and diff checks passed.

On Python 3.11 / PyTorch 2.5.1+cpu: **571 float32/float64 tests**, **42 migration/parameter checks per half dtype**, and **6 representative fullgraph cases** passed. Container compilation is skipped on 2.5 because its existing union-type checks cannot be traced, verified on the pre-review revision. A broader half-precision run has the same 22 failures before and after these fixes: unsupported CPU grid sampling and existing numerical assertions.

CUDA validation is local; no GPU CI runner is configured. The CPU FX guard provides a structural regression signal in CI but cannot replace CUDA execution. MPS runtime was unavailable. The container uses its public tensor-only mode (`disable_features=True`). Eager perspective generator overhead from the constant-construction workaround remains; the stack rewrite does not establish a speedup over the previous PR head.

Written by Codex (gpt-6) on behalf of @ducha-aiki


ducha-aiki:
## Code review — #4517

Reviewed the diff against its merge base (`3c37d390`) with A/B worktrees on CUDA (RTX 4090, torch 2.14.0+cu130). The findings below marked *verified* are reproductions run on both revisions, not speculation.

**What holds.** The PR's own claim is correct: numeric ranges + an explicit `.to('cuda')` now compile cleanly. I confirmed `fullgraph=True` for `RandomAffine`, `RandomPerspective` (basic and area-preserving), `p=0.5`, `same_on_batch`, 4-element `scale`, `.cuda()`, and inside `AugmentationSequential` — all of which fail on the merge base. Recompilation guards behave (moving the aug back to CPU after compiling re-specializes). `tests/augmentation/test_base.py`, `test_random_generator.py` and `test_perspective_rand.py` pass on CPU and CUDA (1108 passed).

---

### 1. Regression: `torch.set_default_device("cuda")` now crashes where it previously worked

`kornia/augmentation/random_generator/_2d/affine.py:170` (and `perspective.py:81`)

Verified A/B:

```python
torch.set_default_device("cuda")
aug = K.RandomAffine(degrees=30., translate=(.1, .1), scale=(.8, 1.2), p=1.0)
torch.compile(aug)(torch.rand(4, 3, 16, 19, device="cuda"))
```

* merge base `3c37d390`: **compiles OK**
* PR `db5606d5`: `RuntimeError: CUDA driver error: invalid argument / Could not find allocation for given ptr ... CUDA_ERROR_INVALID_VALUE`

Before the change, `_extract_device_dtype` returned `torch.empty(0).device` — i.e. `cuda:0` under the ambient default — so parameters were already on the GPU. Now the un-migrated CPU sampler wins, parameters land on CPU, and the Inductor CPU-constant transfer is re-armed. This is the exact crash the PR sets out to fix, reintroduced in a different configuration. No test covers `set_default_device`.

### 2. Tensor-valued constructor ranges still crash, and a new test pins that split as intended

`affine.py:170`, `tests/augmentation/test_perspective_rand.py:82`

Verified on the PR branch:

```python
torch.compile(K.RandomAffine(degrees=torch.tensor([10., 20.]), translate=(.1, .1),
                             scale=(.8, 1.2), p=1.0).to("cuda"))(cuda_x)
# RuntimeError: CUDA driver error: invalid argument
```

Same for `torch.compile(K.RandomPerspective(torch.tensor(0.5), p=1.0).to('cuda'))`.

Root cause: `AffineGenerator.degrees` is a plain attribute, not a buffer or `Parameter`, so `.to('cuda')` never moves it. `aug._param_generator.degrees.device` stays `cpu` while `degree_sampler.low.device` is `cuda:0`, and the `not any(isinstance(value, torch.Tensor) ...)` gate hands precedence back to `_extract_device_dtype`, forcing CPU forever. Tensor ranges are a documented, type-hinted constructor form; `test_affine_tensor_range_keeps_placement` now asserts this configuration is correct behavior, locking in something that cannot be compiled on CUDA.

### 3. Altitude: the fix is in the generator, not at the CPU→CUDA transfer

Verified identical on base and PR:

```python
torch.compile(K.RandomAffine(degrees=30., translate=(.1, .1), p=1.0))(torch.rand(4, 3, 16, 19, device="cuda"))
# RuntimeError: CUDA driver error: invalid argument
```

`kornia/augmentation/_2d/base.py:63` documents that "parameter sampling normally uses CPU defaults independently of the image device" — i.e. *not* calling `.to()` is the documented default, and it is also the most common way users compile an augmentation. The transfer that actually trips Inductor is `torch.as_tensor(params[...], device=input.device)` at `kornia/augmentation/_2d/geometric/affine.py:132-138` and `params[...].to(input)` at `perspective.py:95`. Fixing placement in 1 of 4 reachable configurations (moved + numeric) leaves un-moved, tensor-range, and ambient-default-device all broken. Worth considering whether the transfer site is the better place to fix this once.

### 4. `_device` and `_dtype` can now name a pair that doesn't exist

`affine.py:169`

`_dtype` is still taken from the ambient default while `_device` comes from the sampler:

```python
aug = K.RandomAffine(30.).to("mps")
torch.set_default_dtype(torch.float64)
aug(x)   # TypeError: Cannot convert a MPS Tensor to float64
```

from `torch.tensor([width, height], device=_device, dtype=_dtype)` (`affine.py:201`) and `perspective.py:88`. Pre-PR this worked because `_device` was CPU. The PR body acknowledges "the default returned floating dtype must be supported on the selected device" but adds no guard, no clamp to the sampler dtype, and no test; the blocking MPS leg runs `--dtype=float32`, so CI cannot catch it.

### 5. A dtype-only `.to()` silently un-does the device move

`kornia/augmentation/base.py:118`

Verified: `aug = K.RandomAffine(30., p=1.0).to('cuda')` → sampler and params `cuda:0`. Then `aug.to(torch.float64)` → sampler `cpu`, `aug.device` `cpu`, params `cpu`. `_BasicAugmentationBase.to` passes `device=None` from `_parse_to` straight into `set_rng_device_and_dtype`, which defaults `None` to CPU. This bug pre-dates the PR, but coupling returned placement to the sampler makes it user-visible: `model.to('cuda')` followed by `model.double()` now silently loses GPU placement and brings the `CUDA_ERROR_INVALID_VALUE` crash back under `torch.compile`. No test covers dtype-only `.to()` after a device move.

### 6. The conventions rewrite narrows a caveat that is still true package-wide

`docs/source/get-started/conventions.rst:296`

The deleted text ("Numeric ranges or tensor-valued constructor ranges can determine the returned device/dtype") still describes every generator this PR did not touch. Verified on the PR branch:

* `K.RandomRotation(degrees=30., p=1.0).to('cuda').forward_parameters(...)` → `degrees` on `cpu`, `batch_prob` on `cuda:0`
* `K.RandomResizedCrop((8,8), p=1.0).to('cuda')` → `src`, `dst`, `input_size`, `output_size` all `cpu`

The new sentence claims only "explicit CPU tensor ranges" produce sample-here/return-there, so a reader concludes numeric-range generators now all follow the sampler. AGENTS.md calls this page "the canonical statement of all of this".

### 7. 3D siblings untouched — new 2D/3D asymmetry

`kornia/augmentation/random_generator/_3d/affine.py:178`

Verified: `K.RandomAffine3D(...).to('cuda').forward_parameters((2,3,8,16,19))` returns every geometry key (`translations`, `center`, `scale`, `angles`, `sxy`..`szy`) on `cpu` while `batch_prob` is `cuda:0`; `RandomPerspective3D` likewise. They don't crash under Inductor today, but the 2D and 3D siblings now disagree about what `.to(device)` means, and the latent CPU→CUDA constant transfer per forward remains. Same untouched pattern in `_3d/rotation.py:78`, `_3d/motion_blur.py:102`, `_3d/crop.py:70`.

### 8. Prose left contradicting the page the PR edited

AGENTS.md requires re-checking every `#NNNN` prose hit in a fix's blast radius. `grep -rnE '#4426|issues/4426' kornia/` returns four surviving hits, none re-checked here:

* `kornia/augmentation/_2d/base.py:63-66` — "returned parameter placement is a separate concern; see #4415 and #4426"
* `kornia/augmentation/base.py:180-184` — "returned parameters can be cast to a different device/dtype"
* `docs/source/augmentation.base.rst:233` — "Returned parameter placement is separate from sampling placement"

For `RandomAffine`/`RandomPerspective` with numeric ranges those are no longer true.

### 9. The regression test can never run on CUDA in CI

`tests/augmentation/test_perspective_rand.py:68`

`conftest.py:410-413` drops every item whose name contains `dynamo` or `compile` unless `KORNIA_TEST_OPTIMIZER` is set, and the only job that sets it (`.github/workflows/pr_test_cpu.yml:167-176`) runs `ubuntu-latest` with no GPU — `grep -rlE 'self-hosted|gpu' .github/workflows/` returns nothing. So `test_dynamo_numeric_parameters` executes only on CPU, where `aug.to(device=cpu)` makes the assertion vacuous and the Inductor CPU-constant bug does not exist. #4516 has no guard against re-regression.

### 10. Test quality nits

`tests/augmentation/test_perspective_rand.py:80`

`TestGeometricTensorRangeDevice` calls `set_rng_device_and_dtype(cpu, float32)`, which is a no-op at default settings — `_BasicAugmentationBase.__init__` already ran it, and `RandomGeneratorBase.set_rng_device_and_dtype` early-exits when nothing changed. On `--device=cpu` the test asserts that CPU tensor ranges produce CPU parameters, which is true no matter what the generator does. The case that actually exercises the precedence rule is a CUDA sampler with a CPU tensor range. Also:

* the `RandomAffine` coverage lives in `test_perspective_rand.py` rather than beside `TestRandomAffine` in `tests/augmentation/test_augmentation.py`
* `self.assert_close(value, value[:1].expand_as(value))` (line 54) is vacuously true for `center`/`start_points` and for `batch_size=1`

### 11. Reuse: the idiom is copy-pasted and reaches through the sampler internals

`perspective.py:81`, `affine.py:170`

`RandomGeneratorBase.set_rng_device_and_dtype` (`random_generator/base.py:62-66`) already stores `self.device`, so both sites could read that instead of reaching into a distribution's `low` tensor — which silently depends on `make_samplers` having run and on that sampler having been built from a device tensor. Better still, `_extract_device_dtype` (`kornia/core/utils.py:96`) could take a `default_device=` argument, letting all ~30 generators (the 3D siblings, `RotationGenerator`, `CropGenerator`, `PlainUniformGenerator`) adopt the behavior instead of hand-copying a comment that says "as in AffineGenerator".

### 12. Efficiency

`affine.py:210-211`

`sx = torch.tensor([0] * batch_size, device=_device, dtype=_dtype)` now builds a `batch_size`-element Python list on the host and copies it to the GPU on every call for any `RandomAffine` without shear — at batch 128 that is 256 list elements plus two H2D transfers per forward. `torch.zeros(batch_size, device=_device, dtype=_dtype)` is the existing idiom used two lines up (187, 199) and does no host work. Line 201 and `perspective.py:88-89` are the same pattern. Separately, `_extract_device_dtype(ranges)` at line 169 still allocates `torch.empty(0)` whose device is then discarded on the numeric path — under `set_default_device('cuda')` that is a discarded CUDA allocation per forward.

### 13. Changelog type

`changelog.d/4517.fixed.md`

The PR body says this "deliberately changes numeric parameter placement after sampler migration, including when an ambient default device differs from the sampling device", a documented convention was rewritten, and a convention pin in `tests/augmentation/test_base.py:926` had to be inverted. User code doing `aug.to('cuda'); aug._params['angle'].numpy()` now raises, and the `set_default_device('cuda')` case moves parameters the other way. Per AGENTS.md that reads as `breaking`, and the fragment should say what used to happen as well as what happens now.

---

Findings 1–3 are the blocking ones. 1 is a straight regression with an A/B reproduction; 2 and 3 bound how much of #4516 this actually closes.

ducha-aiki:
Applied the review fixes in 2e0b2585.

- **1–3:** Replaced the sampler-placement workaround with compiler-safe geometry constant construction. Numeric and CPU tensor ranges now compile with moved or unmoved samplers, and the ambient CUDA default regression is removed. The existing parameter-transfer sites and output-placement rules are preserved.
- **4, 6–8:** Restored the original device/dtype conventions and #4426 test pin. This removes the new MPS float64 placement issue and 2D/3D policy asymmetry. Rechecked the #4426 references: their sampling-versus-returned-placement caveats remain accurate; that broader issue is still open. Added default-device/default-dtype tests.
- **5:** Both augmentation and generator `.to()` preserve omitted device/dtype arguments. Added direct sampler assertions and an end-to-end dtype-only migration compile case.
- **10:** Tensor-precedence tests now put the sampler on the selected device while keeping explicit range tensors on CPU. Shared-sampling assertions exclude constant centers/corners and singleton batches. Kept the tests in this file because the existing `TestRandomAffine` is already in `test_perspective_rand.py`.
- **11–13:** Removed both distribution-internal device overrides. Used factories for center/corner/sign/zero constants, including `torch.zeros` for absent affine shear. No new placement policy remains, so the rewritten changelog correctly stays a `fixed` fragment.
- **9 remains an infrastructure limitation:** No GPU CI runner is configured. The expanded tests run under the existing optimizer fixture; all 84 CUDA fullgraph cases were verified locally on the RTX 4090. CPU CI cannot substitute for CUDA runtime coverage.

Verification: **2,176 CPU / 2,170 CUDA** focused tests; **84 CPU / 84 CUDA** fullgraph tests; **192 exact A/B parameter-and-RNG comparisons** against `3c37d390`; **285 passed + 2 expected failures** for each CPU half profile; **4 ONNX export/runtime tests**; full pre-commit, typecheck, and diff checks passed. Also ran the exact global CUDA-default reproduction successfully for all three transform variants. MPS runtime was unavailable locally.

The container compile case uses the public `disable_features=True` tensor-only mode. Its default image-conversion wrapper separately fails fullgraph tracing on both base and changed revisions; the ordinary eager container path is also tested. Updated the PR description to describe the final implementation and these validation limits.

Written by Codex (gpt-6) on behalf of @ducha-aiki

ducha-aiki:
## Review of #4517 (head `2e0b2585`)

**The fix works for what #4516 reports.** I compiled `RandomAffine` and `RandomPerspective` with `fullgraph=True` on CUDA in 12 variants and all of them passed. The new tests pass on CPU and CUDA (766 passed, float32/float64), and the affine and perspective ONNX export tests still pass. Compiling with dynamic shapes behaves the same as on the merge base. I tested on an RTX 4090 with torch 2.14.0+cu130 and checked `kornia.__file__` in every run.

The problem is scope. The workaround changes two call sites, but the same CPU-constant pattern exists in other augmentations, and they still crash on this branch.

### Main issues

1. **Other augmentations still crash on CUDA the same way.** These all raise `RuntimeError: CUDA driver error: invalid argument` on this branch:
   - `ShearGenerator` (`random_generator/_2d/shear.py:107`) has the exact `torch.tensor([width, height], ...).view(1, 2) / 2.0 - 0.5` line this PR rewrote in affine. `torch.compile(K.RandomShear((-10., 10.), p=1.0), fullgraph=True)(torch.rand(4, 3, 16, 19, device="cuda"))` crashes. `RandomAffine` with shear passes.
   - `RandomThinPlateSpline.generate_parameters` (`_2d/geometric/thin_plate_spline.py:82`) builds `src = torch.tensor([[[-1, -1], ...]]).expand(B, 5, 2)`. It crashes with `fullgraph=True` once you call `torch.cuda.synchronize()`.
   - `RandomErasing` (`rectangle_earase.py:159`, `values`) and `CenterCrop` (`center_crop_generator`, `crop.py:356`) crash even without `fullgraph`.

   Rather than fixing one generator at a time, I'd suggest one shared constant-construction helper used by every generator (including the 3D affine and crop ones), or a fix at the shared parameter-transfer layer. If this PR stays narrow, the changelog should say which transforms are covered, so users don't take it as "geometric augmentations now compile on CUDA".

2. **The regression test can't fail in CI.** `test_dynamo_parameter_generation` only reproduces #4516 on CUDA. CI has no GPU runner, and the test is skipped on MPS. If you revert `affine.py` and `perspective.py` to `torch.tensor(...)`, every CPU job still passes. A structural check that runs on CPU would catch that, for example capturing the FX graph and asserting that no lifted `_tensor_constant` feeds a device transfer.

3. **`test_ambient_default_device` leaks a torch function mode into the rest of the session** (`tests/augmentation/test_perspective_rand.py:106`). It restores with `torch.set_default_device(torch.get_default_device())`. When no default device was set, that call doesn't clear anything. It pushes a permanent `DeviceContext('cpu')`:
   ```python
   >>> torch.overrides._get_current_function_mode_stack()
   []
   >>> torch.set_default_device(torch.get_default_device())
   >>> torch.overrides._get_current_function_mode_stack()
   [<torch.utils._device.DeviceContext object at ...>]
   ```
   After that, every torch call in the process goes through Python `__torch_function__`, and later dynamo tests trace under a mode real users never have. Use `with torch.device(device):`, or restore with `torch.set_default_device(None)` when nothing was set before.

4. **`_BasicAugmentationBase.to` changes the samplers before `super().to()` validates the arguments** (`augmentation/base.py:119-122`). The ordering predates this PR, but the PR edits this function:
   - `aug = K.RandomAffine(30.0); aug.to(torch.int64)` raises `TypeError: nn.Module.to only accepts floating point or complex dtypes`. Afterwards, though, `aug.dtype == aug._param_generator.dtype == torch.int64`, so later forwards sample with integer samplers.
   - The new `None` fallbacks aren't needed. `super().to()` already calls the overridden `_apply`, which gets the missing device or dtype from `self.device` and `self.dtype` (`torch.nn.Module.to(aug, torch.float64)` alone keeps `cuda:0`). A spy on `make_samplers` shows `aug.to("cuda")` building the samplers twice, first with `device("cuda")` and then with `device("cuda", 0)`. Dropping the call before `super()` removes the duplicate work and fixes the int-dtype problem.

5. **`RandomGeneratorBase.to` still doesn't fully move the generator** (`random_generator/base.py:79`). It has no `_apply` override and never calls `super().to()`. As a result, `g.cuda()` and `g.double()` do nothing to the samplers (`K.RandomAffine(30.0)._param_generator.cuda()` leaves `device == cpu`). Also, `K.RandomRotation((10., 20.))._param_generator.to("cuda")` reports `cuda` while its `degrees` buffer stays on CPU. Most of this predates the PR, but the changelog's statement about moving generators suggests it works.

### Smaller points

6. **The two workaround comments contradict each other.** `perspective.py:83-84` says indexed fills "can still make it reuse the returned CPU corner buffer directly in CUDA kernels". Affine builds `center` with exactly that (`torch.empty` plus `center[0, 0] = width`). It only works because `/ 2.0 - 0.5` happens to allocate a new buffer. A refactor that removes the arithmetic, or a different Inductor fusion decision, would bring the crash back in affine with no CPU-side signal. A shared helper (point 1) would also fix this.
7. **Eager CPU slowdown in perspective.** `start_points` now takes 3 factories + stack + view, and `pts_norm` takes ones + 2 sliced `__setitem__`, where there used to be 2 `torch.tensor` calls. Generator forward at (8, 3, 64, 64) on one CPU thread: **21.7 µs → 27.3 µs (+26%)**. Affine is unchanged (17.7 vs 17.1 µs). Building `pts_norm` with a single stack would avoid the two indexed writes.
8. **Existing utility.** The TL/TR/BR/BL corner stack re-implements `kornia.geometry.bbox.bbox_generator`, which already builds these inclusive `start + size - 1` corners by stacking. `bbox_generator(zero, zero, torch.full((), width), torch.full((), height))` returns the same `(1, 4, 2)` tensor. That's safer than a flat 8-element stack, where the order is easy to get wrong.
9. **`test_default_dtype_after_sampler_move` doesn't check what its comment describes** (`test_perspective_rand.py:79`). The comment says "a float64 default must not force float64 sampling on MPS", but the test only checks the dtype of the returned parameters. A regression that rebuilt the samplers at float64 would still pass on CPU and CUDA. Asserting `aug._param_generator.dtype == torch.float32` and checking the sampler's `low.dtype` would cover it.
10. **Test placement.** The affine tests (the `affine` param of `TestGeometricParameterDevice` and `TestGeometricTensorRangeDevice.test_affine_tensor_range_keeps_placement`) live in `test_perspective_rand.py`. Someone changing `AffineGenerator` won't look there, and runs like `test-module tests/augmentation/test_augmentation.py -k Affine` skip them.

**Not verified:** MPS, the torch 2.5.1 leg, and CPU float16/bfloat16.

ducha-aiki:
Applied the latest review fixes in bb7ca4bf.

- **1:** Kept this PR focused on RandomAffine and RandomPerspective. The changelog and PR body now explicitly name the other reported transforms that remain outside this workaround.
- **2:** Added a CPU FX regression guard for lifted geometry constants and indexed scalar fills. Verified that it rejects the merge-base tensor constructors and the pre-review indexed-fill code. CUDA runtime coverage remains local, but CPU CI can now detect those regressions.
- **3, 9:** Replaced the global default-device mutation with a scoped device context and asserted that the torch-function mode stack is restored. The ambient-dtype test now checks the generator dtype and actual sampler bound dtype/device.
- **4–5:** Delegated `.to()` validation to Module.to and synchronized samplers through `_apply`. Invalid integer dtypes leave samplers unchanged; device canonicalization rebuilds them once; generator `.to()`, convenience methods, and parent-module moves update registered buffers and samplers together. Added regressions that failed before these changes.
- **6–7:** Both generators now use scalar stacks without indexed fills, and their workaround comments agree. Perspective signs also use a stack. The single-thread CPU generator benchmark at (8,3,64,64) measured perspective at 28.2 µs versus 28.1 µs on the previous PR head and 20.7 µs on the merge base (IQR about 1 µs). The eager overhead remains; this rewrite does not establish a speedup.
- **8:** Did not substitute bbox_generator because it subtracts in the tensor dtype and changes existing half-precision coordinates. With float16 width 2049, the existing far corner is 2048 but the suggested bbox call yields 2047; bfloat16 width 257 similarly gives 256 versus 255. The source now explains the required subtraction-before-casting order.
- **10:** Moved shared placement/compilation coverage into `test_geometric_parameter_device.py`, and placed the affine tensor-range test beside `TestRandomAffineGen` in `test_random_generator.py`.

Also made eager replay in the fullgraph regression independent of `_params` side effects: parameters are generated and returned explicitly inside the compiled function. This passes on PyTorch 2.5.1, whose existing export guard suppresses those state updates. The revised wrapper still reproduces the original merge-base CUDA crash. Its container case skips Torch 2.5's existing inability to trace union-type checks, reproduced on the pre-review head.

Verification: **5,081 focused CPU/CUDA tests**, **168 fullgraph cases**, **294 passed + 2 expected failures per CPU half profile**, **384 exact A/B parameter/RNG comparisons**, **4 ONNX checks**, full pre-commit, typecheck, and diff checks. On PyTorch 2.5.1 CPU: **571 float32/float64 tests**, **42 migration/parameter checks per half dtype**, and **6 fullgraph cases** passed. The broader 2.5.1 half run reproduces the same 22 failures on the pre-review head, involving unsupported grid sampling and existing numerical assertions. MPS runtime and GPU CI remain unavailable. Independent code review found no remaining actionable issues.

Written by Codex (gpt-6) on behalf of @ducha-aiki

ducha-aiki:
Extended the #4517 fix in follow-up #4523 (head `69e6069b`). A shared helper now covers the additional 2D/3D geometry, erasing, crop/resize, TPS, illumination and mix generators. The wider audit also fixed CUDA sampler migration in shear and mixed-device sign generation in both linear-illumination generators.

Verification includes 6,200 focused CPU/CUDA tests, 509 combined regression tests (340 Inductor cases), 160 moved/unmoved generator-transfer cases, CPU half-precision profiles, ONNX/export checks, and PyTorch 2.5.1 coverage. The CPU graph guard rejects all 18 affected generator cases on the merged base. Exact A/B parameter/RNG results and the intentionally corrected shear half-precision sampling are detailed in the new PR.

The additional CUDA audit passed 17 other augmentations. Existing crop, ColorJiggle, Mosaic and CutMix fullgraph limitations were reproduced on the base and are documented in #4523. MPS runtime was unavailable.

Written by Codex (gpt-6) on behalf of @ducha-aiki

PR #4518, opened 2026-09-14
fix: clarify unimplemented Boxes error messages

victorwon2001:
Refs #4017

Adds corrective messages to the three intentionally unimplemented `Boxes` paths tracked by #4017: tuple bounds passed to `Boxes.clamp`, `Boxes.trim`, and `Boxes.translate(method="fast")`. Their existing `NotImplementedError` types and triggering conditions are preserved; the messages now describe the supported `(B, 2)` tensor bounds, the unimplemented `trim` operation, and the supported `warp` translation method respectively.

The existing #4017 regression tests now assert the important parts of those messages, and a bug-fix changelog fragment records the user-visible error improvement. Supported tensor clamp behavior, `method="warp"`, public signatures, and the unknown translation-method branch are unchanged.

Refs #4017

## What did you check?

```text
pixi run --environment default test-module tests/geometry/test_boxes.py -k "4017" --device=cpu --dtype=float32,float64
pixi run --environment default test-module tests/geometry/test_boxes.py --device=cpu --dtype=float32
pixi run --environment default typecheck
pixi run --environment default pre-commit-all
pixi run --environment default changelog-preview
```

The focused regression cases and full Boxes module pass, and the final tree passes type checking, all pre-commit hooks, and the changelog preview.

## How did AI help?

AI was used to inspect the affected code paths, implement the scoped error-message and regression-test changes, add the changelog fragment, and run the verification commands above.


ducha-aiki (changes requested):
Review of `49d904f4ed9120ace9d57404502db10fc8e4117b`.

The code change is correct. The three bare `NotImplementedError`s now carry messages. The exception type and the conditions that trigger each raise are unchanged, and every alternative a message names works when called. The pins now use `match=`, and they fail against main's `boxes.py`. One thing blocks merging: the PR body closes #4017, but this PR only does the interim part of that issue.

**[P2] `Fixes #4017` would close an issue this PR does not fully fix.** The body starts with `Fixes #4017` and later says `Refs #4017`. GitHub uses the first one: `closingIssuesReferences` = `[4017]`. #4017's Expected section asks for each of the three paths to be either implemented or removed. It calls corrective messages the step to take *until then*, because removal is an API change and needs a deprecation window. After this PR:
- `trim()` still always raises (`boxes.py:606`).
- `clamp` is still typed `Optional[torch.Tensor | tuple[int, int]]` (`boxes.py:524-525`).
- The `translate` docstring still says `method: "warp" or "fast".` (`boxes.py:985`).

In the merge tree of this PR with main, `kornia/geometry/boxes.py:270` still says the unimplemented `trim`, `translate(method='fast')` and tuple-bound `clamp` paths "are tracked in #4017". The `test_wart_*_4017` pins also still cite it. If the issue closes, that warning points at a closed issue while the defect is still there. Please change the first line to `Refs #4017`, and make sure no other closing keyword is left in the body.

**[P3] The unknown-method branch still raises with an empty message.** `translate` at `boxes.py:999`: `b.translate(torch.tensor([[1., 2.]]), method="bogus")` -> `NotImplementedError('')`. The PR says it leaves this branch alone on purpose, and #4017 does not list it, so this does not block. Since the neighbouring branch now has a message, the same kind of message here would finish the method, e.g. ``f"`Boxes.translate` supports only method='warp'; got {method!r}."``, keeping the exception type.

Verified on the PR worktree (torch 2.14.0, CPU):

| call | result |
|---|---|
| `clamp((0, 0), (4, 4))` | ``NotImplementedError('`Boxes.clamp` accepts `topleft` and `botright` as `(B, 2)` torch.Tensor bounds; got topleft=tuple and botright=tuple.')`` |
| `clamp()` / `clamp(tensor, None)` / `clamp([0,0],[4,4])` | same message, with `NoneType`/`NoneType`, `Tensor`/`NoneType`, `list`/`list`: accurate in each case |
| `clamp` with `(B, 2)` tensors, B=2 with `(1, 2)` broadcast, unbatched `(1, 2)` | ok, values clamped (so the alternative the message names works) |
| `trim()` | ``NotImplementedError('`Boxes.trim` is not implemented.')``: names no alternative, and there is none |
| `translate(size, method="fast")` | ``NotImplementedError("`Boxes.translate(method='fast')` is not implemented; use `method='warp'` instead.")`` |
| `translate(size, method="warp")`, default method, unbatched | ok, `[[2,4],[5,4],[5,5],[2,5]]` for box `xyxy (1,2,5,4)` + `(1,2)` |
| `except NotImplementedError` on `trim()` | caught, `type(e) is NotImplementedError` |

`Boxes3D` has no `NotImplementedError` sites on main. `VideoBoxes` inherits these three methods, so the messages apply to it too. The only other raise in the module, `_merge_box_list` at `boxes.py:46`, already had a message. The `fixed` fragment is accurate: the three paths, `NotImplementedError` kept, tensor bounds and `warp` named where they apply.

Overlap with other open PRs: none. `git merge-tree` of this PR with #4379 (`910feee7`, integer-box dtype) and with #4194 (`f90db50c`, `__getitem__` padding metadata) is clean both times. Neither of them edits `clamp`, `trim` or `translate`.

**What I ran**
- `.venv` python 3.11, torch 2.14.0; `kornia.__file__` pointed into the PR worktree.
- `pytest tests/geometry/test_boxes.py --device=cpu --dtype=float32,float64`: 268 passed, 4 xfailed.
- Plant-and-revert, step 1: I replaced `kornia/geometry/boxes.py` with main's copy and ran `-k 4017`. 4 failed, all `Regex pattern did not match`.
- Plant-and-revert, step 2: I put back only the bare `translate(method="fast")` raise and ran `test_wart_trim_and_fast_translate_are_unimplemented_4017[cpu-float32]`. It failed with `Expected regex: 'fast.*not implemented.*warp'`, `Actual message: ''`, so the fast-translate `match=` catches a regression on its own. I restored the file afterwards; worktree status is clean.
- Probe table above.
- `git merge-tree --write-tree origin/main <head>`: clean. `git grep -nE "#4017|issues/4017"` on that tree: `kornia/geometry/boxes.py:270` and `tests/geometry/test_boxes.py:688,710`.
- No `testing/` manifest entries for these tests.
- CI job logs: the only failure is `##[error]Unexpected HTTP response: 500` in `prefix-dev/setup-pixi`.

ducha-aiki:
## Status update — this branch went `CONFLICTING` today, and the resolution is three lines

#4575 ("resolve Ruff return-flow warnings") merged as `6cb89631d` and its RET505 pass flattened the exact `if/elif/else` chain in `Boxes.translate` that this PR rewrites. GitHub now reports `CONFLICTING` against `main` (`ecbcad561`) in one file, `kornia/geometry/boxes.py`, one hunk.

Because the branch is `CONFLICTING`, **no workflow runs on it** — the green checks you see are stale and predate #4575.

### The conflict

```
        if method == "fast":
<<<<<<< HEAD
            raise NotImplementedError(
                "`Boxes.translate(method='fast')` is not implemented; use `method='warp'` instead."
            )
        elif method == "warp":
            pass
        else:
=======
            raise NotImplementedError
        if method != "warp":
>>>>>>> ecbcad561
            raise NotImplementedError
```

Your `elif method == "warp": pass` / `else:` shape is precisely what RET505 removed on main. Keep main's flattened control flow and carry your message into it:

```python
        if method == "fast":
            raise NotImplementedError(
                "`Boxes.translate(method='fast')` is not implemented; use `method='warp'` instead."
            )
        if method != "warp":
            raise NotImplementedError
```

That is behaviourally identical to both sides — `fast` raises with your message, anything other than `warp` raises bare, `warp` falls through — and it stays RET505-clean, so it will not be re-flattened by the next lint sweep.

### While you are in there

Your PR title is "clarify unimplemented Boxes error messages" and the body says the unknown-method branch is deliberately unchanged. That is a defensible scope line, but it does leave `translate` with one informative raise and one bare one three lines apart. If you want the file to read consistently:

```python
        if method != "warp":
            raise NotImplementedError(f"Unknown `method` {method!r}; expected 'warp' or 'fast'.")
```

Entirely optional — say so and I will not press it.

### The previous round's ask is already discharged — disregard it

The last review asked you to change `Fixes #4017` to `Refs #4017`. You already did: the GraphQL
`closingIssuesReferences` for this PR is empty, so #4017 will correctly stay open when this merges.
Nothing to do there.

Once you push the rebase, CI will run for real for the first time since #4575; ping me and I will re-review the delta.

ducha-aiki (approved):
Re-reviewed `a8ef2d4c25` against current main `3a6a1e333`: APPROVE.

The previous blocker is resolved: the PR body uses `Refs #4017`, preserving the issue for the unimplemented operations. The merge keeps main's flattened control flow and the three actionable exception messages without changing exception types or supported operations. I withdraw the earlier optional unknown-method-message request: that branch is explicitly outside this change's scope and is not a regression. No open review findings remain.

Validation on the clean merge with main: `tests/geometry/test_boxes.py --device=cpu --dtype=float32,float64` — 270 passed, 4 xfailed. The imported module was verified to come from the review worktree. Current-head CI was green at review time.

PR #4519, opened 2026-09-14
fix(augmentation): keep shear ranges on target device

### LINKED ISSUE #4415: W1b: `RandomShear` and `RandomAffine(shear=...)` cannot be moved off CPU unless `shear` is a 4-tuple

Hrithikcrick:
## Summary

Fixes #4415 by passing the requested `device` and `dtype` into the shear `_range_bound(...)` calls used by both `ShearGenerator` and `AffineGenerator`.

This prevents scalar and 2-tuple shear configurations from mixing CPU tensors with tensors on the requested device, and also keeps the 4-tuple case on the correct device/dtype.

Regression coverage is included for `RandomShear` and `RandomAffine` with scalar, 2-tuple, and 4-tuple shear inputs.

## Validation

- `python -m pytest tests/augmentation/test_perspective_rand.py -q` → `165 passed`
- focused regression tests → `6 passed`
- `git diff --check` → clean
- pre-commit.ci formatting applied on the pushed branch


Hrithikcrick:
Closing this PR because current main now addresses the same shear device/dtype issue through the shared _shear_bound helper. Thanks!

PR #4539, opened 2026-09-14
docs(augmentation): refresh RandomPerspective3D's stale doctest output

VanniLeonardo:

`RandomPerspective3D`'s docstring example was printing the output of the 3D sampling-grid defect that #4514 fixed, so `pixi run doctest` fails on `main` right now. This restores the example to what the class actually returns. The numbers are a permutation of the old ones, which is exactly the signature of the `(d, x, y)` against `(x, y, z)` grid mix-up #4514 removed.

This is my regression from #4514 and I am sorry for the noise.

The new values are the correct ones rather than merely the current ones. The invariant #4514 established holds on this example's own class:

```
RandomPerspective3D(distortion_scale=0., align_corners=True)   max|out - in| = 0.0
RandomPerspective3D(distortion_scale=0., align_corners=False)  max|out - in| = 0.875
```

A zero distortion is now exactly the identity at `align_corners=True`. The residual at `False` is the 3D normalization convention tracked in #4503 and is unrelated to this change.

### Why CI did not catch it on #4514

`.github/workflows/docs.yml`, which is the only job that runs `pytest --doctest-modules`, is path-filtered:

```yaml
paths:
  - 'docs/**'
  - '*.rst'
  - '.github/workflows/docs.yml'
```

#4514 touched `kornia/`, `tests/` and `changelog.d/` only, so the docs job never ran on it. The failure surfaced on #3945, which edits a `.rst` file and therefore does run the job, after it merged `main`.

I have not changed that filter here, since when the doctests run is your call and adding `kornia/**` would run the docs job on most PRs. If you would like it changed, or the doctest run moved next to the test jobs, I am happy to do it in a separate PR.

Follow-up to #4514. Unblocks the `docs (3.11)` job on #3945.

## What did you check?

```text
655 passed, 14 skipped

$ pytest --doctest-modules kornia/augmentation/_3d/geometric/perspective.py
1 passed

$ pytest tests/augmentation --dtype=float32,float64 -k "Perspective3D or perspective"
470 passed, 16 skipped
```

On `main` the same full doctest run is `1 failed, 650 passed, 18 skipped`, failing on this one example. Pre-commit passes on both changed files.

## How did AI help?

Claude Code found the root cause, produced and verified the corrected values, and drafted this description. I reviewed it and ran the checks above.


VanniLeonardo:
Fixed, the fragment is now `changelog.d/4539.fixed.md` and the check passes locally.

Worth flagging the reason I got it wrong, since it will catch other contributors. `changelog.d/README.md` says:

> If there is no PR number yet, use a unique name such as `+fix-camera-shape.fixed.md`, then rename it to the PR number before merging.

But `check_changelog.py` matches `(?:[0-9]+|\+migration-[0-9]+)\.(?:added|fixed|breaking)\.md`, so any `+name` other than `+migration-<number>` is a hard CI failure rather than something you rename later. I followed the README and hit the error.

Two ways to reconcile them, whichever you prefer. Either the README drops that paragraph and tells people to open the PR first and name the fragment after it, or the pattern widens to allow a general orphan prefix and the presence check keeps requiring the numbered file before merge. Happy to send that as a separate one-line docs PR if it is useful.

ducha-aiki (approved):

Issue #4478, opened 2026-09-13
AugmentationSequential mask round trip rounds integer labels, casts every mask to the last mask's dtype, and fails on B = 0


ducha-aiki:
`AugmentationSequential` converts masks to a floating working dtype and back, and that round trip corrupts masks in three ways (main `a7e2db6b`, torch 2.14.0, CPU):

```python
import torch
import kornia.augmentation as K

flip = lambda: K.AugmentationSequential(K.RandomHorizontalFlip(p=1.0), data_keys=["input", "mask"])

# 1. integer labels are rounded to the image dtype, even though no interpolation happens
for image_dtype, labels in ((torch.float32, [16777216, 16777217]), (torch.float16, [2048, 2049, 26001]),
                            (torch.bfloat16, [256, 257, 1203])):
    mask = torch.tensor(labels).reshape(1, 1, 1, -1)
    flip()(torch.rand(1, 3, 1, len(labels), dtype=image_dtype), mask)[1].flip(-1)
# float32  [16777216, 16777217] -> [16777216, 16777216]  (int64)
# float16  [2048, 2049, 26001]  -> [2048, 2048, 26000]
# bfloat16 [256, 257, 1203]     -> [256, 256, 1200]

# 2. every mask output is cast to the dtype of the last mask argument
semantic = torch.tensor([0, 2, 3, 5]).reshape(1, 1, 1, 4)
seq = K.AugmentationSequential(K.RandomHorizontalFlip(p=1.0), data_keys=["input", "mask", "mask"])
_, a, b = seq(torch.rand(1, 3, 1, 4), semantic, semantic > 0)
a.dtype, a.flatten().tolist()   # torch.bool, [True, True, True, False] -- labels 2, 3, 5 are gone

# 3. an empty batch with a mask fails before any op runs
flip()(torch.rand(0, 3, 4, 4), torch.rand(0, 1, 4, 4))
# IndexError: index 0 is out of bounds for dimension 0 with size 0
```

A mask placed before the image (in `data_keys` order or dictionary insertion order) is also converted with the previous call's image dtype, or `float32` on a fresh container. A `float64` mask with values like `1 + 2**-30` loses them even when the image is `float64`.

**Cause** (`container/augment.py`):
- `_preproc_mask` (`:598-608`) casts masks to `self.input_dtype`. That is the image dtype recorded by the most recent image argument, not necessarily from this call, and the cast is lossy for integer labels outside the dtype's exact-integer range.
[...]

PR #3968, opened 2026-08-11
fix: align Kimi-VL with Moonshot implementation

TomasGuija:
## Which lane? *(check one — see [Contributing Guide](https://github.com/kornia/kornia/blob/main/CONTRIBUTING.md#which-lane-is-your-contribution))*
- [ ] 🟢 **Green lane** — test-first bug fix / verified docs fix / [`help wanted`](https://github.com/kornia/kornia/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22help%20wanted%22) issue / benchmark results. No prior issue needed.
- [x] 🟠 **Discuss-first** — feature or behavior change; a maintainer confirmed the scope on the linked issue.

Fixes: #3879

## What does this PR do?

This PR better aligns Kornia’s Kimi-VL implementation with Moonshot’s official [configuration](https://huggingface.co/moonshotai/Kimi-VL-A3B-Instruct/blob/398eede0903cd983a2bfa0cc634e9ac1d843f375/config.json) and [model implementation](https://huggingface.co/moonshotai/Kimi-VL-A3B-Instruct/blob/398eede0903cd983a2bfa0cc634e9ac1d843f375/modeling_kimi_vl.py).

It addresses discrepancies in:

1. MoonViT configuration values.
2. Learned positional embeddings and runtime interpolation.
3. Two-dimensional rotary positional embeddings (RoPE).
4. MoonViT activation handling.
5. Projector patch-merging order.
6. Existing unit and integration test compatibility.

> This is an initial implementation opened early for discussion; I plan to continue reviewing implementation details and add targeted tests based on maintainer feedback.

## Test log

```powershell
python -m pytest tests/models/test_kimi_vl.py tests/integration/models/test_kimi_vl.py -q
```

```text
16 passed
```
## Local comparison with Moonshot

The following local script was run unchanged on upstream `main` and this branch. It was used for manual verification and is not included in the PR.

```python
from pathlib import Path

import torch
from safetensors import safe_open
from transformers import AutoConfig
from transformers.dynamic_module_utils import get_class_from_dynamic_module

from kornia.models.kimi_vl import KimiVLBuilder
from kornia.models.kimi_vl.config import _kimi_vl_a3b_instruct_config

REPO = "moonshotai/Kimi-VL-A3B-Instruct"
REVISION = "398eede0903cd983a2bfa0cc634e9ac1d843f375"
OFFICIAL_SHARD = Path("checkpoints/kimi-vl/model-00001-of-00007.safetensors")
[...]

PR #4540, opened 2026-09-14
fix(augmentation): disable antialiasing for masks

### LINKED ISSUE #4479: Resize(antialias=True) blurs masks in AugmentationSequential, creating values that are not labels
`Resize(..., antialias=True)` inside `AugmentationSequential` applies the anti-aliasing Gaussian blur to masks before the nearest-neighbour downscale. Masks then contain values that are not labels, integer masks lose and invent labels, and boolean masks change coverage (main `a7e2db6b`, torch 2.14.0, CPU):

```python
import torch

yy, xx = torch.meshgrid(torch.arange(12), torch.arange(16), indexing="ij")
blocks = ((yy // 2) + (xx // 2)) % 2   # 2x2 blocks, so a plain nearest 2x downscale keeps them exactly
for name, mask in (("float {2, 3}", (2 + blocks).float()), ("int64 {2, 3}", 2 + blocks), ("bool", blocks.bool())):
    for antialias in (False, True):
        seq = K.AugmentationSequential(K.Resize((6, 8), antialias=antialias), data_keys=["input", "mask"])
        out = seq(torch.rand(1, 3, 12, 16), mask[None, None])[1]
```

```
antialias=False float {2, 3} values=[2.0, 3.0]                          mean=2.500
antialias=True  float {2, 3} values=[2.0, 2.1065, 2.1903, 2.8097, 2.8935] mean=2.500
antialias=False int64 {2, 3} values=[2.0, 3.0]                          mean=2.500
antialias=True  int64 {2, 3} values=[1.0, 2.0]                          mean=1.979
antialias=False bool         values=[0.0, 1.0]                          mean=0.500
antialias=True  bool         values=[0.0, 1.0]                          mean=0.979
```

**Cause.** The mask path (`GeometricAugmentationBase2D.apply_transform_mask`) forces `resample` to nearest but reuses the image `flags`. `Resize.apply_transform` (`_2d/geometric/resize.py:104-112`) passes `antialias=flags["antialias"]` to `resize`, so the blur runs on the mask too.

**Expected.** Masks keep their label set under `Resize`. Anti-aliasing is image-only, for example by clearing `antialias` on the mask path the way `resample` is overridden.

Found while reviewing #4452, whose new Convention text says masks are resampled "without introducing intermediate labels" on every 2D geometric augmentation.

Hrithikcrick:
## Summary

Closes #4479.

Disable antialiasing on the geometric mask path so `Resize(antialias=True)` continues to antialias images without filtering segmentation masks.

This preserves mask labels and dtype while masks continue to use nearest-neighbour resampling.

Regression coverage includes float, int64, and boolean masks.

## Validation

- regression test: 3 passed
- `tests/augmentation/container/test_augmentation_sequential.py`: 81 passed, 50 skipped
- `git diff --check`: clean


ducha-aiki (changes requested):
Review of `47995d231d3137b12613c6dc574a97a516221a08`.

On the default mask path the change does what #4479 asks. Masks come out of `Resize(antialias=True)` with the same labels they went in with, and images are still antialiased. The updated pin fails on main for all three mask dtypes. `Resize` is the only augmentation with an `antialias` flag: `LongestMaxSize` and `SmallestMaxSize` inherit it fixed at `False`, and `RandomResizedCrop` has none. So the base-class override reaches nothing else. The merge is clean on top of #4509, and the new block follows #4509's `resample` default, with one gap (first finding). Two things need fixing before merge.

**[P2] An explicit mask `antialias` is now silently overwritten, which regresses main and contradicts the #4509 contract in the same class.** `kornia/augmentation/_2d/geometric/base.py:121-122` sets `kwargs["antialias"] = False` unconditionally. Three lines above, #4509 made `resample` default to nearest only when the caller did not pass one, and the class `Note` says "An explicit value is honoured in both directions". The mask-path defaults should work the same way. A soft mask (probability map) resampled bilinearly is a legitimate opt-in to antialiasing, and main honours it. With `K.Resize((6, 8), antialias=True)`, a 12x16 float checkerboard mask, and `extra_args={DataKey.MASK: {"resample": BILINEAR, "align_corners": True, "antialias": True}}`, the mask output was compared against `resize(mask, (6, 8), "bilinear", align_corners=True, antialias=...)`:

```
                                          main                     PR head
extra_args antialias=True               == aa: True  (11 values)  == aa: False (2 values)
extra_args without antialias key        == aa: True  (11 values)  == aa: False (2 values)
extra_args antialias=False              == no-aa: True            == no-aa: True
aug.transform_masks(..., antialias=True, resample=BILINEAR)
                                        == aa: True               == aa: False
```

Suggested fix, matching the `resample` line above it:

```python
if "antialias" not in kwargs and "antialias" in (self.flags if flags is None else flags):
    kwargs["antialias"] = False
```

With that guard planted in the PR worktree, the PR's pin still passes (`test_augmentation_sequential.py`: 156 passed, 100 skipped). The explicit rows go back to `== aa: True`, and the default container path still gives mask values `[2.0, 3.0]`. Please add a row that pins it: `extra_args={DataKey.MASK: {..., "antialias": True}}` on a float mask equals `resize(..., antialias=True)`, like #4509's explicit-resample guard.

Even with the guard, the second row changes behaviour: an explicit bilinear mask resample without an `antialias` key used to inherit the image's `antialias=True`, and now does not. That is consistent with how `resample` is handled, but the fragment should say so (see below).

**[P2] The #4479 defect prose survives in `kornia/` and `docs/`.** In the merge tree of `origin/main` and this head, `git grep -nE "#4479|issues/4479"` still finds three places that describe the bug as current:

- `kornia/augmentation/_2d/geometric/base.py:48-51` (Convention block): "container mask processing has dtype- and operator-specific limitations; see #4478 and #4479". Drop #4479 here. The `Note` at lines 55-59 is the right place to add that masks are not antialiased unless `antialias` is passed explicitly.
- `kornia/augmentation/container/augment.py:128-130`: "``Resize(antialias=True)`` filters masks before sampling and can change labels (#4479)". Remove the `Resize` clause and keep the padding fill-value caveat.
- `docs/source/get-started/conventions.rst:225-229`: "``Resize(antialias=True)`` filters masks before sampling, producing fractional labels or changing integer and boolean masks (#4479)". Same edit.

The only hit in `tests/` is the flipped pin itself, which is correct. Our open docs PR #4538 carries the same three passages unchanged from main, so whichever PR lands second has to reconcile them.

**[P3] The changelog fragment overstates the fix and omits the opt-in path.** `changelog.d/4540.fixed.md` says "preserving mask labels and dtype". Dtype was already preserved on main. With the production change reverted in the PR worktree, all six parametrizations get past `assert out_mask.dtype == mask_dtype` and fail only at `assert torch.equal(out_mask, expected_mask)`. Suggested wording: "`Resize(antialias=True)` no longer antialiases masks, so their label set is preserved; pass `antialias=True` in `extra_args[DataKey.MASK]` (or to `transform_masks`) to filter soft masks explicitly." After the guard above, the last clause is also true.

**[P3] Nothing pins the image side of the PR body's claim that `Resize(antialias=True)` "continues to antialias images".** The test only compares masks. Measured on both trees, the image output equals `resize(img, (6, 8), "bilinear", align_corners=True, antialias=True)` (`True`). Asserting `with_antialias(image, mask)[0]` is close to `resize(image, (6, 8), "bilinear", align_corners=True, antialias=True)` would pin it.

**What I ran**

- Environment: repo `.venv` python 3.11, torch 2.14.0. `kornia.__file__` printed as the PR worktree or main worktree in every probe.
- Pin on PR head: `pytest tests/augmentation/container/test_augmentation_sequential.py -k 4479 --device=cpu --dtype=float32,float64` gave 6 passed.
- Plant-and-revert: with main's `base.py` restored in the PR worktree, the same run gave 6 failed, all at `torch.equal(out_mask, expected_mask)`. Worktree restored clean afterwards.
- Neighbouring tests: `test_augmentation_sequential.py` + `tests/augmentation/test_base.py` at float32/float64 gave 311 passed, 100 skipped, 8 xfailed. `tests/augmentation/test_augmentation.py -k "Resize or MaxSize"` gave 18 passed, 1 skipped, 1 xfailed.
- Half precision: `test_augmentation_sequential.py` with `--xfail-known-failures`, profile `cpu-float16` gave 70 passed, 50 skipped, 11 xfailed; `cpu-bfloat16` gave 78 passed, 51 skipped, 2 xfailed.
- Other runs: this file has no compile tests (`KORNIA_TEST_OPTIMIZER=inductor -k "4479 or compile or dynamo"` selects only the three float32 pin rows, 3 passed); the change is a Python dict edit outside the traced kernel. On the torch 2.5.1 overlay (prints `2.5.1`, PR worktree), `-k 4479` gave 6 passed.
- A/B probe (same heredoc in both worktrees): image-vs-`resize` equality, default container mask values (main `[1.9999996, 2.1065, 2.1903, 2.8097, 2.8935, ...]`, PR `[2.0, 3.0]`), `extra_args` with and without `antialias`, direct `transform_masks` with an `antialias=True` kwarg, `self.flags["antialias"]` not mutated (still `True`), and inverse mask values (main fractional, PR `[2.0, 3.0]`).
- Merge: `git merge-tree --write-tree origin/main <head>` exits 0.
- CI: `gh pr checks` shows 41 pass.

Hrithikcrick:
Thanks for the detailed review. Addressed the requested changes:

- mask antialiasing is now disabled only by default; an explicit `antialias` value is honored
- added regression coverage for explicit mask `antialias=True`
- added coverage that image antialiasing remains enabled
- removed the resolved #4479 wart text from the API/container docs
- updated the changelog to describe the explicit soft-mask opt-in path

Validation:
- `test_augmentation_sequential.py`: 82 passed, 50 skipped
- `git diff --check`: clean

ducha-aiki (approved):
Review of `e21e7fbd0baf967d9c2427cde3478e940f3f6b8c`.

This is a re-review of the delta from `47995d23` to `18ee6bd2`, plus a one-line maintainer edit on top. All four findings from the [previous review](https://github.com/kornia/kornia/pull/4540#pullrequestreview-5207101292) are fixed, and I verified each one:

- **[P2] explicit mask `antialias`:** `base.py` now uses the `"antialias" not in kwargs` guard, so an explicit value is honoured. The new `test_resize_mask_explicit_antialias_override_4479` pins it. To check that the pin catches the bug, I put the previous head's `base.py` back in the PR worktree: both float32/float64 rows then fail and the six default-path rows still pass. The file was restored afterwards.
- **[P2] #4479 prose:** `git grep -nE "#4479|issues/4479" -- kornia docs tests` finds nothing in the merge tree of `origin/main` and this head. The class `Note` now documents the mask default and the opt-in.
- **[P3] fragment:** it now states the label-set fix and the soft-mask opt-in, without the dtype claim.
- **[P3] image side:** the existing pin now also compares the image against `resize(..., antialias=True)`.

**Maintainer edit `e21e7fbd`.** The hosted run on `18ee6bd2` was red on 5 jobs. The half-precision bfloat16 job and the four Windows jobs all failed `tests/test_half_precision_ci.py::test_eager_rng_audit_matches_exact_inventory` with `audited entries with no matching call site (stale line?): [('tests/augmentation/container/test_augmentation_sequential.py', 48, 'torch.randn') ×2]`. The new `from kornia.geometry.transform import resize` import moved `test_mixup`'s collection-time `torch.randn` parametrize from line 48 to line 49. The RNG inventory in `testing/half_precision_eager_rng.py` records exact lines, so I pushed the matching `48` → `49` change. Before the edit, that test fails locally 1/4; after it, `tests/test_half_precision_ci.py` passes 62/62. Any future edit above that line needs the same bump.

The bfloat16 job also failed `tests/feature/test_sosnet.py::TestSOSNet::test_jit[cpu-bfloat16]` with a TorchScript `builtin cannot be used as a value` error in `Conv2d._conv_forward`. I don't think this PR causes it:
- the PR does not touch `kornia/feature` or TorchScript code;
- the same job passed on main's scheduled run and on #3968's run from the same base;
- it did not reproduce locally, either in a full `tests/` bfloat16 run on a torch 2.9.1 overlay (the CI job's torch) or in a focused run on 2.14.0.

I read it as a TorchScript flake. If it comes back on the new run, it deserves an issue of its own.

**What I ran** (repo `.venv`, python 3.11; `kornia.__file__` printed as the PR worktree):

- `test_augmentation_sequential.py` + `tests/augmentation/test_base.py`, CPU float32/float64, torch 2.14.0: 313 passed, 100 skipped, 8 xfailed.
- `test_augmentation_sequential.py` with `--xfail-known-failures`: profile `cpu-float16` gave 71 passed, 50 skipped, 11 xfailed; `cpu-bfloat16` gave 79 passed, 51 skipped, 2 xfailed.
- `tests/augmentation/test_augmentation.py -k "Resize or MaxSize"`, float32: 9 passed, 1 skipped, 1 xfailed.
- `-k 4479`, float32/float64: 8 passed on torch 2.5.1 and 8 passed on 2.14.0. Also 4 passed on MPS float32 and 4 passed with `KORNIA_TEST_OPTIMIZER=inductor`.
- Full `tests/` suite, bfloat16, torch 2.9.1 overlay, `--xfail-known-failures --known-failure-profile=cpu-bfloat16` on `18ee6bd2`: 10191 passed, 562 xfailed, 5 xpassed (non-strict), 2 failed. One failure is the RNG audit fixed above. The other is `test_check_torch_env`, which flags the overlay's duplicate torch install and is an artefact of the overlay. SOSNet `test_jit` passed.
- Merge: `git merge-tree --write-tree origin/main <head>` exits 0. A simulated merge of main + this PR with our open #4538 is also clean, and leaves no #4479 hit, so #4538 owes nothing further for #4479.
- CI on `e21e7fbd`: all 42 checks pass, including the half-precision float16/bfloat16 jobs (SOSNet `test_jit[cpu-bfloat16]` passed this time) and all Windows legs.

Approving; ready to merge.

Issue #4525, opened 2026-09-14
perf(morphology): the default `engine="unfold"` is 8x slower than `engine="convolution"` on MPS (identical output), and the docstring's speed claim is inverted on CPU


ducha-aiki:
## Problem

`dilation`, `erosion`, `opening`, `closing`, `gradient`, `top_hat`, `bottom_hat` default to `engine="unfold"`: the image is unfolded into a 6-D `(B, C, H, W, kh, kw)` window tensor and reduced with `max`/`min` over the window axes. On MPS that reduction is a weak kernel (a reduction whose innermost dim is 3–7 wide runs 2–5x slower than the CPU; details and standalone repros in #4522), so the whole morphology family is 1.5–2.1x **slower** than the M1 CPU at every shape up to 1024², while `engine="convolution"` is 8x faster on MPS and gives identical output for flat structuring elements.

| op, 256², k7 | CPU unfold | CPU convolution | MPS unfold | MPS convolution | max abs diff between engines |
|---|---:|---:|---:|---:|---:|
| `dilation` | 37.7 ms | 52.0 ms | 58.7 ms | **7.2 ms** | 0.0 |
| `erosion` | 34.5 ms | 38.1 ms | 56.2 ms | **7.0 ms** | 0.0 |
| `dilation`, 512², k7 | 141.5 ms | 214.0 ms | 221.6 ms | **23.1 ms** | 0.0 |

So the CPU prefers `unfold` and accelerators prefer `convolution`, and the docstring says the opposite of the CPU result: "convolution is faster and less memory hungry, and unfold is more stable numerically".

## Proposal

- Pick the engine by device when the caller does not specify one (`unfold` on CPU, `convolution` on CUDA/MPS), or
- keep the default but document which engine wins where, with numbers.

Related: #2948 / #3001 (memory of the unfold engine at large kernels), #4521, #4522.

Repro:

```python
import torch, torch.utils.benchmark as bench, kornia.morphology as KM
def us(fn, dev):
    s = torch.mps.synchronize if dev == "mps" else (lambda: None); fn(); s()
    return bench.Timer(stmt="fn(); s()", globals={"fn": fn, "s": s}).blocked_autorange(min_run_time=0.4).median * 1e6
for dev in ("cpu", "mps"):
    im = torch.rand(1, 3, 256, 256, device=dev); ker = torch.ones(7, 7, device=dev)
[...]

PR #4544, opened 2026-09-14
fix(morphology): cast conv weight/bias to input dtype in engine=convolution

### LINKED ISSUE #4541: [Bug]: kornia.morphology dilation/erosion crash with engine="convolution" when tensor.dtype != kernel.dtype
### What happened?

`kornia.morphology.dilation` and `erosion` (and everything built on them — `opening`, `closing`,
`gradient`, `top_hat`, `bottom_hat`) raise a `RuntimeError` when called with `engine="convolution"`
if the input tensor's dtype differs from the kernel's dtype. This is a realistic scenario since
kernels are typically built with `torch.ones(k, k)`, which defaults to `float32`, while the image
tensor may be `float64` (or `float16`/`bfloat16`).

The default `engine="unfold"` works fine with the exact same inputs — PyTorch's elementwise
arithmetic auto-promotes dtypes there — so this is specifically a bug in the `engine="convolution"`
path, not a documented limitation. The `engine` docstring only describes it as a speed/numerical-
stability tradeoff, not a dtype-compatibility one.

Error:
RuntimeError: Input type (double) and bias type (float) should be the same

Root cause: `_neight2channels_like_kernel` and the conv `bias` in `kornia/morphology/morphology.py`
are built using `kernel.dtype`, not `tensor.dtype`. `F.conv2d` requires input/weight/bias to match
dtype exactly (unlike plain tensor arithmetic, which promotes), so it hard-errors instead of
producing a result.

### Minimal reproduction

```python
import torch
from kornia.morphology import dilation, erosion

tensor = torch.rand(1, 3, 5, 5, dtype=torch.float64)
kernel = torch.ones(3, 3)  # default dtype: float32

dilation(tensor, kernel, engine="convolution")
# RuntimeError: Input type (double) and bias type (float) should be the same

erosion(tensor, kernel, engine="convolution")
# same error
```

### Environment

- Kornia version: 0.9.0rc1
- PyTorch version: 2.14.0+cpu
- Python version: 3.11.14
- OS / device (CPU, CUDA, ROCm, MPS): Windows-10-10.0.26200-SP0, CPU (CUDA/ROCm unavailable, MPS unavailable)

### Anything else?

I have worked on the fix and will open a PR shortly. Loosely related: #4525 also touches
engine="convolution" behavior (perf comparison) but doesn't hit this dtype path since it uses
matching dtypes.

aryan32134hello:

`kornia.morphology.dilation`/`erosion` (and everything built on them — `opening`, `closing`, `gradient`, `top_hat`, `bottom_hat`) crashed with `engine="convolution"` whenever the input tensor's dtype differed from the kernel's dtype — a realistic case since kernels are typically built with `torch.ones(k, k)` (float32 default) while the image tensor may be float64/float16/bfloat16.

Root cause: `_neight2channels_like_kernel` and the conv `bias` were built from `kernel.dtype`, not the input tensor's dtype. `F.conv2d` requires input/weight/bias to match dtype exactly (unlike plain tensor arithmetic, which promotes — this is why `engine="unfold"` never hit this). This casts both to `output.dtype` (the input tensor's dtype) before the `F.conv2d` call, so the result always preserves the input tensor's dtype regardless of what dtype the kernel was built in.

Adds `changelog.d/+morphology-convolution-dtype.fixed.md`.

Fixes #4541 

## What did you check?

Added `test_convolution_engine_dtype_mismatch` to `test_dilation.py` and `test_erosion.py`, covering `tensor.dtype != kernel.dtype` with `engine="convolution"`. Confirmed it fails on the unfixed code (via `git stash`) and passes with the fix, across float32/float64/float16.

```text
$ pixi run test-module tests/morphology/test_dilation.py tests/morphology/test_erosion.py --dtype=float32,float64,float16
53 passed in 5.11s

$ pixi run pre-commit-all
12 hooks passed (fix end of files, trim trailing whitespace, check yaml, check toml, check for merge conflicts, check for added large files, mixed line ending, pretty format json, ruff check, ruff format, codespell, Add license header)

$ pixi run typecheck
Found 1 diagnostic (pre-existing, unrelated — kornia/feature/lightglue.py:44, a torch.cuda.amp.custom_fwd deprecation warning untouched by this change)
```


ducha-aiki (changes requested):
Review of `90029a100cbbe6dfef7a6c1b6f5bc803ee6d2d12`.

The fix is right. `dilation` and `erosion` are the only places that call `F.conv2d`, and `opening`, `closing`, `gradient`, `top_hat` and `bottom_hat` all go through them, so casting the eye weight and the bias to `output.dtype` covers the whole family. Blocking issue: the new erosion test compares random input against `engine="unfold"` with a fixed `1e-3` tolerance. On an arm64 Mac in float32, that fails in about half of unseeded runs.

**[P2] `tests/morphology/test_erosion.py:174` (and the same pattern in `test_dilation.py:187`): the new test is flaky on macOS arm64 float32.**
In float32, the macOS convolution path has error near the geodesic border, where the padding is ±`max_val` = ±1e4. This is already documented in `test_kernel` in the same file: "~3.9e-4 absolute / ~1.5e-3 relative error on macOS's Apple-backend float32 conv path". Those older tests use fixed inputs. The new test uses `torch.rand`, so a border minimum close to 0 gets a tolerance of about `1e-3`, while the engines differ by up to `2.4e-3`. The error is already on main (same-dtype float32 `torch.ones(3, 3)` kernel on a `(1, 2, 5, 5)` input: unfold vs convolution differ by `1.4e-3` for erosion). The PR's test is what makes it show up. Measured on an M1 CPU, torch 2.14.0:

- `pytest tests/morphology --device=cpu --dtype=float32`, run 20 times: 9 failures, all `TestErode::test_convolution_engine_dtype_mismatch[cpu-float32]`. Sample failure: `Greatest absolute difference: 0.0013924241065979004 at index (0, 1, 4, 1) (up to 0.001 allowed)`.
- The test body over 3000 seeds with the PR's `atol=1e-3, rtol=1e-3`: erosion float32 fails 1231/3000 and dilation float32 fails 47/3000. float64, float16 and bfloat16 fail 0/3000.

The 5 macOS float32 CI legs passed. Every fresh process draws different inputs, so a green run does not prove the test is stable. The explicit `atol`/`rtol` also brings back the kind of hardcoded override that #4088 removed.

The test only needs to show that a mismatched kernel dtype gives the same result as a matching one, and it can do that exactly. Compare against the same convolution call with `kernel.to(dtype)`. That is the same computation after the cast, so the result is bitwise equal. On this machine, `torch.equal` held for 16000/16000 cases (float32↔float64, float16/bfloat16 tensor with a float64/float32 kernel) and 8000/8000 with the pairing below. Pick the other dtype so that it never needs float64. Right now `other_dtype` is always float64 for float32/float16/bfloat16 inputs, so the MPS branch always skips and the test never runs on MPS:

```python
other_dtype = torch.float16 if dtype == torch.float32 else torch.float32
tensor = torch.rand(1, 2, 5, 5, device=device, dtype=dtype)
kernel = torch.ones(3, 3, device=device, dtype=other_dtype)
result = erosion(tensor, kernel, engine="convolution")
assert result.dtype == dtype
self.assert_close(result, erosion(tensor, kernel.to(dtype), engine="convolution"))
```

Putting a non-flat `structuring_element` in a third dtype into the same test would also pin the bias cast. On main, a float64 structuring element with a float32 tensor and float32 kernel failed with `Input type (float) and bias type (double) should be the same`. On the PR it works, with a max diff of `4.5e-4` against unfold, which is the same backend error.

**[P3] The two engines now return different dtypes for the same call.**
`engine="unfold"` still promotes through `output + neighborhood`. `engine="convolution"` now returns the input dtype. The fragment only talks about `engine="convolution"`, so it is accurate. Measured on the PR across all 7 ops (conv dtype / unfold dtype):

| tensor | kernel | convolution | unfold |
|---|---|---|---|
| float16 | float32 | float16 | float32 |
| bfloat16 | float16 | bfloat16 | float32 |
| float32 | float64 | float32 | float64 |

Values agree after casting: max diff 0.0 in float16/bfloat16/float64. In float32 the max diff is ≤ 2.1e-3, which is the backend border error above and is the same on main for matching dtypes. Building `neighborhood` in `tensor.dtype` once would make both engines return the input dtype and replace the two casts. That changes what the default engine returns for mixed dtypes, so it belongs in a follow-up with its own fragment, not in this PR.

**Checked, no finding**
- On main, all 12 mismatched pairs of {float16, bfloat16, float32, float64} raise `RuntimeError: Input type (...) and bias type (...) should be the same` for all 7 ops with `engine="convolution"`. The 4 matching pairs work. On the PR, all 16 pairs work for all 7 ops and return the input dtype.
- The new tests fail without the fix: restoring main's `morphology.py` makes both fail for each of float32, float64, float16 and bfloat16 (2 failed per dtype).
- Half precision and `max_val`: with the default `max_val=1e4` there are no NaNs in float16 for either engine with a float16 or float32 kernel. With `max_val=1e5` and a float16 tensor or float16 kernel, both engines raise `value cannot be converted to type c10::Half without overflow` from `F.pad`/`index_put`, on main and on the PR. So the cast does not open a silent `-inf` → `0 * inf` NaN path.
- Integer kernels: an `int64` kernel used to fail with the conv engine and now matches unfold, up to the same backend border error. A `bool` kernel gives wrong values in both engines, because `-max_val` stored in a bool becomes `True` = +1. That was already true of unfold on main and is out of scope here.
- Gradients: for `gradient(...)` with a float64 tensor, float32 kernel and float32 `structuring_element`, forward and `se.grad` match unfold exactly, `x.grad` differs by `8.9e-16`, and grad dtypes follow the inputs. `torch.autograd.gradcheck` passes for convolution-engine `dilation`/`erosion` with a float64 tensor and float32 kernel.
- `torch.compile(fullgraph=True, backend="eager")` of `opening(..., engine="convolution")` with float32 input and a float64 kernel has no graph break and is bitwise equal to eager.
- `closingIssuesReferences` is exactly #4541. In the merge tree, `#4541` only appears in the two test comments. No other open PR or issue covers this; #4525 (engine perf) does not touch the dtype path. The fragment's claims hold for `engine="convolution"`.

**What I ran**: `.venv` interpreter with torch 2.14.0, and a torch 2.5.1 overlay. `kornia.__file__` was checked in every probe (PR worktree vs main worktree).
- PR, `pytest tests/morphology --device=cpu --dtype=float32,float64`: 101 passed on the first run, 1 failed / 100 passed on a rerun (the flaky node).
- `--dtype=float16 --xfail-known-failures --known-failure-profile=cpu-float16`: 54 passed. Same for bfloat16: 54 passed.
- torch 2.5.1, `pytest tests/morphology/test_dilation.py tests/morphology/test_erosion.py --dtype=float64,float16`: 36 passed.
- Restored main's `morphology.py` in the PR worktree and ran `-k dtype_mismatch` per dtype: 2 failed ×4. The PR worktree is back to clean.
- Probes for the dtype matrix and engine comparison, `max_val`/integer kernels, gradients and compile, on both trees.
- `git merge-tree` against current main: clean.

aryan32134hello:
Hi @ducha-aiki, Thanks for the thorough review — really appreciate the detailed measurements.

Fixed the [P2] flakiness in 535726a33. The test now compares engine="convolution" with a mismatched kernel against the same convolution call with the kernel pre-cast to the input dtype (kernel.to(dtype)), instead of comparing against engine="unfold". Since that's the same computation, the results are bitwise equal, so I dropped the atol/rtol override and switched to self.assert_close(). I also changed other_dtype to the float16/float32 pairing so it never needs float64 — that removes the always-skip on MPS.

Verified locally on Windows/x86, torch 2.14.0:
- The new test body over 3000 seeds: 0/3000 failures for both dilation and erosion across float32/float64/float16/bfloat16 (was ~1231/3000 for erosion float32 with the old comparison).
- Full tests/morphology suite green across all four dtypes.

Left the morphology.py fix unchanged, and kept the two-engine dtype-return difference ([P3]) as-is for a follow-up per your note.

ducha-aiki:
@aryan32134hello I don't see push in this repo, are you sure it is in correct branch?

aryan32134hello:
Hi @ducha-aiki — it's pushed and on the right branch. The PR head is now 535726a331a8210f06fa17239cf6ab119a31a839 on aryan32134hello:fix-morphology-convolution-dtype, and the commit list here shows all three commits (the fix, the changelog rename, and the test rework 535726a33). It touches tests/morphology/test_dilation.py and test_erosion.py.

Could you try a refresh? Happy to re-push if it still isn't showing for you.

ducha-aiki (approved):
Review of `535726a3`. Both prior items are fixed and this is ready to merge.

The flakiness is gone, and the contrast is clean. On the old head `90029a10` the two nodes failed **6 of 20** fresh `pytest` processes (`--device=cpu --dtype=float32 -k dtype_mismatch`) and were **skipped** on `--device=mps --dtype=float32`. On this head they fail **0 of 30** of the same runs and both **pass** on MPS. Comparing against the same-dtype convolution call makes it exact rather than merely tighter: over 5000 fresh `torch.rand` draws per dtype, `torch.equal(op(t, k), op(t, k.to(dtype)))` holds 5000/5000 for `dilation` and `erosion` in float32, float64, float16 and bfloat16 — 40 000 comparisons, zero mismatches — and `result.dtype == dtype` every time. The hardcoded `atol`/`rtol` and the cross-engine comparison are both gone.

Both casts are pinned independently. Restoring main's `morphology.py` fails 8 nodes (2 tests × 4 dtypes); casting the weight but not the bias fails 4; casting the bias but not the weight fails 4. The bias cast survives as a pin because `F.conv2d` rejects the dtype mismatch, so the all-ones kernel's zero bias does not weaken it.

The production cast re-checks out across the whole family: 112 `(op, tensor dtype, kernel dtype)` combinations per engine — all 7 ops against all 16 pairs of {float16, bfloat16, float32, float64} — give no errors, the input's dtype for `engine="convolution"`, and finite values for both engines; on MPS, 21 combinations preserve dtype and device and are bitwise equal to the same-dtype call. `gradcheck` passes for convolution-engine `dilation` with a float64 tensor and float32 kernel, and `torch.compile(fullgraph=True)` of `opening(..., engine="convolution")` with a float64 kernel is bitwise equal to eager.

I also re-ran it on `30dfbf711` merged with your head (merge is clean), since #4566 landed after your CI:

```
$ pytest tests/morphology -q --device=cpu --dtype=float32,float64      -> 101 passed
$ pytest tests/morphology -q --device=cpu --dtype=float16 --xfail-known-failures \
      --known-failure-profile=cpu-float16                              -> 54 passed
```

Nothing in #4566's blast radius touches morphology, so this one is unaffected by it. On torch 2.5.1: 148 passed on cpu float32+float64+float16 and 47 passed / 7 skipped on mps.

One optional item, not blocking: adding a `structuring_element` in a third dtype would pin the bias cast on the value as well as on the dtype check. I confirmed it works — float32 tensor, float32 kernel, float64 `structuring_element`, `engine="convolution"` returns float32 and is bitwise equal to casting `se` to float32 up front. The engine dtype divergence noted last time (`unfold` promotes, `convolution` preserves) is unchanged and still belongs in its own PR; the fragment correctly scopes itself to `engine="convolution"`.

`Fixes #4541` is accurate: `closingIssuesReferences` is exactly #4541, its whole scope is this crash, and its repro works at head. `ruff@0.16.7`, `--doctest-modules`, the fragment check and the towncrier render are all clean.


ducha-aiki (approved):

PR #4545, opened 2026-09-14
Vectorize KMeans centroid update to remove per-cluster host syncs

### LINKED ISSUE #4533: perf(contrib): KMeans updates centroids with a per-cluster `nonzero` + `index_select` + `mean` loop (160 host syncs per fit at k=8); use scatter/bincount
## Problem

`KMeans.fit` (`kornia/contrib/kmeans.py:180–186`) recomputes centroids as

```python
for index in range(self.num_clusters):
    selected = torch.nonzero(cluster_assignment == index).squeeze()
    selected = torch.index_select(X, 0, selected)
    ...
    current_centers[index] = selected.mean(dim=0)
```

`nonzero` is a host round trip on every accelerator (470 µs per call on MPS for 4k elements, #4522), so a fit with k=8 and 20 iterations does 160 of them plus 160 `index_select` and 160 `mean`. At 4096×32 the fit takes 29 ms on the M1 CPU and **94 ms on MPS**; at 65536×32 MPS wins (269 ms vs 460 ms) because the distance matrix dominates.

## Fix

One `index_add_` of `X` into a `(k, D)` zeros tensor by `cluster_assignment`, one `bincount(cluster_assignment, minlength=k)` for the counts, and a `where` on `count == 0` for the empty-cluster re-seed. No per-cluster loop, no sync per cluster; the empty-cluster random reseed can stay a masked `torch.randint` gather. (#2809 / #2861 already moved the distance to `kornia.geometry.linalg.euclidean_distance`; this is the other half.)

prithvirajrh:
## What

`KMeans.fit` updated centroids with a Python `for` loop over clusters, using `torch.nonzero` + `index_select` per cluster. Each call has a data-dependent output shape, which forces a host sync on every cluster of every iteration.

This replaces it with a one-hot assignment matrix and a single matmul to get per-cluster sums, plus `one_hot.sum(0)` for counts (kept in `int64`, exact). The sum accumulates in at least float32: `X.dtype` itself overflows to `inf` in float16 past 65504 and loses precision in bfloat16, for perfectly ordinary inputs like float16 image color quantization - caught in review, see below. The empty-cluster fallback (reinit to a random point) is computed unconditionally with `torch.where`, so that branch never depends on a data-dependent Python bool either.

## A note on the approach

I first tried `scatter_add_` into the `(num_clusters, D)` destination, since that's the natural reading of "vectorize with scatter". It's correct, but benchmarking showed it's actually slower than the original loop on MPS - many threads end up racing to accumulate into very few destination rows, and that specific access pattern seems to be a weak spot for the MPS scatter kernel. Routing the same reduction through a one-hot matmul instead avoids that entirely and is fast on both backends.

## Benchmark

My own numbers, N=20000, D=32, k=8, 20 iterations, best of 8 runs:

| device | before | after |
|---|---|---|
| CPU | 74.1ms | 50.7ms (~1.5x) |
| MPS | 71.9ms | 36.3ms (~2x best; avg 206.8ms -> 61.6ms) |

Independently re-measured in review (best of 8, `mps.synchronize` inside the timed region): MPS wall clock at the same N/D/K went 110.6/133.5ms -> 44.9/43.6ms. On CPU, review's more thorough sweep across problem sizes found the win shrinks toward noise at larger N, since the O(N*K*D) distance computation (unchanged by this PR) comes to dominate - worth knowing if you came here for a CPU speedup at scale. This is primarily an MPS/host-sync fix.

## Behavior changes worth knowing

- A fit that hits an empty cluster may now produce different (but equally valid) centers for the same seed - the random reseed now draws every iteration instead of only when a cluster is empty, so the RNG stream advances differently.
- `fit` no longer mutates a `cluster_centers` tensor passed in at construction (previously written into row-by-row).
- Calling `fit` a second time no longer warm-starts from the first call's result, as a consequence of the same change.

None of these were documented behavior before; all three are called out in the changelog fragment.

## Verified

- All existing and new `tests/contrib/test_kmeans.py` tests pass on CPU and MPS across float32/float64/float16/bfloat16, including `test_dynamo` (torch.compile)
- Two new regression tests cover exactly the review findings: an empty cluster reseeding to a real data point, and per-cluster sums large enough to overflow float16 if accumulated in the input dtype (parametrized across both cases from review)
- Full `tests/contrib/` suite: 204 passed, 0 regressions
- `ruff check` / `ruff format` clean, `ty check` shows no new diagnostics
- `torch.export` and ONNX export, which failed on `main` for this case, now succeed
- Changelog fragment at `changelog.d/4545.fixed.md`

Fixes #4533


ducha-aiki (changes requested):
Review of `8e8ad6b7007621d920557556c4bf7c9bbc0f88d0`.

The vectorization does what #4533 asks. The per-cluster `nonzero`/`index_select` loop is gone, float64 fits are bit-identical to main, and a fixed-iteration fit now traces as one graph and exports. The new update does its sums and counts in `X.dtype`, though. For float16 inputs that overflows on ordinary data, where main's `mean` stayed finite, and for bfloat16 it biases the centroids.

**[P1] float16 centroids become `inf`/`NaN`, and bfloat16 centroids drift, because sums and counts accumulate in the half dtype** (`kornia/contrib/kmeans.py:186-188,195`)

`one_hot.t() @ X` sums up to N points in float16, which overflows past 65504. `one_hot.to(X.dtype).sum(0)` also counts in the half dtype. In float16 that count becomes `inf` above 65504. In bfloat16 it is no longer exact well before that. Main's per-cluster `selected.mean(0)` avoided both problems: on CPU its float16 error is identical to computing in float32 (numbers below). Measured end to end with `KMeans(...).fit` on CPU with torch 2.14.0; torch 2.5.1 gives the same result:

| input | main | head |
| --- | --- | --- |
| float16 64x64 RGB in [0, 255], K=4 | centers finite, `[99.5, 59.53, 75.44]` | `[inf, inf, inf]` |
| float16 3000 points around 50, K=2 | `[[47.6, 49.9], [52.4, 50.2]]` | `[[inf, inf], [49.7, 45.6]]`, all 3000 points in one cluster |
| float16 512x512 RGB in [0, 1], K=2 | `[[0.248, 0.506, 0.500], [0.748, 0.493, 0.500]]` | `[[0.338, 0.316, 0.644], [nan, nan, nan]]` |
| float16, K=1, `50 + arange(3200) % 4` reshaped to (1600, 2), one iteration (exact mean `[51, 52]`) | `[51.0, 52.0]` | `[inf, inf]` |
| bfloat16, same input | `[51.0, 52.0]` | `[51.0, 51.75]` |

Here is the error of a single centroid update against the float64 mean of the same half-rounded data, with the assignment held fixed. The formulas are copied from each tree.

| dtype, data | main | head | head count error |
| --- | --- | --- | --- |
| float16, N=4096, [0, 255] | 0.0573 | inf | 0 |
| float16, N=200000, [0, 1] | 0.000175 | 0.502 | inf |
| bfloat16, N=3000, [50, 53] | 0.0245 | 0.272 | 4 |
| bfloat16, N=200000, [0, 1] | 0.00173 | 0.00274 | 232 |

Colour quantisation of a float16 image is a normal use of `KMeans`, and it is exactly this input. A minimal fix keeps the one-hot matmul, counts in int64 and accumulates half inputs in float32:

```python
one_hot = torch.nn.functional.one_hot(cluster_assignment, num_classes=self.num_clusters)
cluster_counts = one_hot.sum(0)  # int64, exact
work = X.float() if X.dtype in (torch.float16, torch.bfloat16) else X
cluster_sums = one_hot.to(work.dtype).t() @ work
...
means = (cluster_sums / cluster_counts.clamp(min=1).unsqueeze(1).to(work.dtype)).to(X.dtype)
```

I applied this in the PR worktree and reverted it afterwards. Its errors match main's for every row above, for example float16 N=4096 [0, 255] = 0.0573 and bfloat16 N=3000 = 0.0245. All the pins below pass with it. `tests/contrib/test_kmeans.py` passes for float32, float64, float16 and bfloat16 (24 passed), and so does `test_dynamo` under inductor. The dispatch counts and the `torch.export` result are unchanged from head. Its float32 step time stays within 1.1-1.4x of head's (table in "What I ran").

**[P2] Missing regression tests for the paths this PR rewrites**

`tests/contrib/test_kmeans.py` runs the new lines but never checks their values in a case that can distinguish implementations. `test_module` uses three well-separated 500-point blobs and compares with `atol=2`. No test produces an empty cluster or a large cluster sum, so CI's float16 and bfloat16 legs pass with the overflow above. Please add these tests, using the injected `device`/`dtype`:

- **Empty cluster.** Use `x = [[0,0],[1,0],[0,1],[1,1]]` with both initial centers at `[0.5, 0.5]` and `max_iterations=1`. Every point ties to cluster 0, so cluster 1 is empty. Assert `centers[0] == x.mean(0)` and that `centers[1]` equals some row of `x`. The pin passes on main and on head in all four dtypes. With `random_points` swapped for `previous_centers` (a keep-the-old-centre bug), it fails with `[[0.5, 0.5], [0.5, 0.5]]`.
- **Large per-cluster sums.** Use `x = 50 + (arange(3200) % 4).reshape(1600, 2)`, K=1 and one iteration, and expect exactly `[[51, 52]]`. A second variant uses 400 rows of 200 plus 400 rows of 20, initial centers at 190 and 30, and expects `[[200]*3, [20]*3]`. Both pass on main in every dtype. On head the first gives float16 `[inf, inf]` and bfloat16 `[51.0, 51.75]`, and the second gives float16 `[inf, 20.0]`. Both pass with the fix above.

**[P2] `# Tracked in #4533.` in the new source comment** (`kornia/contrib/kmeans.py:185`)

This PR closes #4533 (`closingIssuesReferences` = [#4533]). On the merge tree, `git grep -nE "#4533|issues/4533" -- kornia tests docs` returns this line as its only hit. After merge the code will point at a closed issue with the phrasing kornia uses for open defects, which trips the closing-issue grep in AGENTS.md. Please drop the reference.

The same comment says `scatter_add_` was "~100x slower on MPS", while the PR body says "~10x slower than the original loop". Please make them agree, or drop the number.

**[P3] Stale graph-capture note** (`docs/export_support/cases_models.py:1126`)

The `contrib.KMeans.fit` case note still says "fit() uses torch.nonzero + data-dependent `if selected.shape[0] == 0` per cluster". That case (`tolerance=None`, `max_iterations=5`) changes behaviour with this PR. On main, non-strict `torch.export` fails with `GuardOnDataDependentSymNode ... Eq(u0, 0)`. On head it succeeds, and `torch.compile` with `CompileCounter` captures one 82-op frame instead of 7 ops. Please update the note. The JSON snapshot is regenerated in bulk and can wait.

**[P3] Changelog fragment**

`fixed` is the right type among `added`/`fixed`/`breaking` for a performance defect. Two things to change in the text:

1. Most of it is implementation detail (one-hot, matmul, GEMM). A single user-facing sentence would do, for example "`KMeans.fit` no longer synchronises with the device once per cluster per iteration; the centroid update is vectorised."
2. "Faster on both CPU and MPS": for the update step alone I measured 1.5-4.7x on CPU. The whole fit, single-threaded, is 10-20% faster at N ≤ 20000 and within noise at N=65536 and 262144, because the N×K×D distance dominates. I did not measure MPS.

A user-visible side effect is not mentioned. The empty-cluster reseed now draws `num_clusters` indices from the global RNG on every iteration, where main drew one index only when a cluster was empty. With the same seed, fits that hit an empty cluster produce different centers than before. The global RNG stream after any fit also differs: `torch.rand(1)` after a 5-iteration fit with no empty cluster gives 0.7773 on main and 0.5603 on head. Worth one clause in the fragment.

**[P3] `index_add_` is much faster than one-hot on CPU at larger K**

This does not block the PR. For a float32 update step (median of 30, 4 threads):

| N, D, K | main loop | head | `index_add_` |
| --- | --- | --- | --- |
| 65536, 3, 16 | 3.50 ms | 1.85 ms | 0.63 ms |
| 16384, 3, 128 | 7.52 ms | 2.97 ms | 0.19 ms |

`index_add_` is the option #4533 proposed. The PR's MPS finding was for `scatter_add_` at K=8, so a device-conditional choice or an MPS measurement at larger K may be worth a look.

**What I ran**

- **Setup.** `.venv` Python 3.11, torch 2.14.0. The torch 2.5.1 overlay was used for the float16 probe and for `test_kmeans.py`. `kornia.__file__` was printed in every probe and named the PR worktree or the main worktree (`53866348`) as intended.
- **Tests on head.** `tests/contrib/test_kmeans.py`, CPU: float32/float64 12 passed; float16 and bfloat16, each with its known-failure profile, 6 passed each; `test_dynamo` with `KORNIA_TEST_OPTIMIZER=inductor` 2 passed; torch 2.5.1 float32/float64 12 passed.
- **Equality vs main.** Same heredoc in both trees: `KMeans(K, None, 1e-4, 50, seed)`, seeds 0-3, configs (1000,3,2), (5000,8,32), (20000,16,3), (257,64,5), K=1 (100,1,4), duplicated points (1000 rows in {0..3}², K=10), and (3000,2,2) around 50.
  - **float64:** all 24 fits without an empty cluster give identical labels and bit-identical centers.
  - **float32:** 21 of 24 have identical labels, max center difference ≤ 7.7e-6. In the other 3, sub-ULP differences in the matmul sum send the trajectory elsewhere (max center difference 0.09).
  - **Empty clusters (duplicates config):** main and head both reseed to a data point and never produce NaN. The reseeded values differ by construction, since the RNG draws differ.
  - **Half precision:** float16 gave non-finite centers for all 4 seeds of the around-50 config.
- **Host syncs.** A `TorchDispatchMode` op count over a 20-iteration fit at 4096×32, K=8:
  - main: `nonzero` 160, `index_select` 160, `mean` 160.
  - head: no `nonzero` or `index_select`, and one `mm`, `index` and `where` per iteration. It also has 40 `_local_scalar_dense`, all from `F.one_hot`'s CPU-only range checks. On MPS, `one_hot` does no host-side check: an out-of-range index did not raise eagerly and surfaced later as an asynchronous device error.
  - The `tolerance` exit check `center_shift**2 < tolerance` still syncs once per iteration, as it does on main.
- **CPU whole-fit timing.** 1 thread, 20 iterations, run main, head, main, head. Medians in ms:

  | N, D, K | main | head |
  | --- | --- | --- |
  | 1000, 2, 3 | 2.43 / 2.44 | 2.14 / 2.01 |
  | 4096, 32, 8 | 24.5 / 26.6 | 21.6 / 19.8 |
  | 20000, 32, 8 | 139 / 158 | 125 / 127 |
  | 65536, 3, 16 | 400 / 402 | 387 / 440 |
  | 16384, 3, 128 | 728 / 727 | 693 / 742 |
  | 262144, 3, 8 | 869 / 877 | 946 / 820 |

- **Step timing with the suggested fix.** float32, median of 30, 4 threads, head vs fix in ms: 0.09 vs 0.08 (4096,32,8); 0.28 vs 0.33 (20000,32,8); 1.85 vs 2.00 (65536,3,16); 2.97 vs 4.07 (16384,3,128); 3.28 vs 3.68 (262144,3,8).
- **Merge and CI.** `git merge-tree --write-tree origin/main <head>` is clean. CI: 41 checks pass, including the float16/bfloat16 half-precision legs, dynamo, tests-mps and typecheck.
- **Not verified:** MPS or CUDA timing, and MPS/CUDA numerics of the suggested fix.

prithvirajrh:
Thanks for the thorough review, and for catching that overflow - I only tested small N and never hit it.

- **P1 (float16/bfloat16 overflow):** fixed as suggested - counts stay exact via int64, sums accumulate in float32 whenever the input is float16/bfloat16, converted back to `X.dtype` only for the final centers. Verified against every case in your table (all finite/correct now), plus a couple of my own with N=200000.
- **P2 (missing tests):** added both cases you described (empty-cluster reseed, large per-cluster sums in both variants). I checked they actually catch the bug - ran them against the pre-fix code first and got the same `inf`/wrong-cluster results you reported, then confirmed they pass with the fix.
- **P2 (`Tracked in #4533`):** dropped, agreed it'd go stale on merge.
- **Benchmark number mismatch:** fixed the code comment to match the PR description's ~10x.
- **P3 (docs/export_support note):** updated to describe the current behavior.
- **P3 (changelog):** rewritten to one user-facing sentence, a more honest CPU/MPS breakdown (MPS is the real win, CPU shrinks toward noise at larger N), and a note about the RNG-stream side effect on empty-cluster reseeding.
- **P3 (`index_add_` at larger K):** noted, didn't act on it since you flagged it as non-blocking - happy to look at a device-conditional choice in a follow-up if useful.

All pushed. Full `tests/contrib/` suite (204 tests) plus lint/typecheck clean locally.

ducha-aiki (changes requested):
Review of `83808a95`.

**The P1 is fixed, in both halves**, and I re-verified it by construction and by measurement. Counts stay int64 and are exact at N=262144 (where a float16 count is `inf`); the sums accumulate in float32 for half inputs. Against the pre-fix revision `8e8ad6b7`, on `(50 + arange(3200) % 4).reshape(1600, 2)` with `KMeans(1, None, None, 1, 0)` (exact mean `[51, 52]`):

| dtype | `8e8ad6b7` | this head | `main` |
| --- | --- | --- | --- |
| float16 | `[inf, inf]` | `[51.0, 52.0]` | `[51.0, 52.0]` |
| bfloat16 | `[51.0, 51.75]` | `[51.0, 52.0]` | `[51.0, 52.0]` |
| float16, N=200000 in [0,1] | `nan` | err 1.6e-04 | err 1.6e-04 |
| bfloat16, N=200000 | err 3.1e-03 | err 8.2e-04 | err 8.2e-04 |

Same on MPS. Empty clusters never produce NaN (`clamp(min=1)` makes 0/0 unreachable) in any of the four dtypes on CPU or MPS. `# Tracked in #4533` is gone and `grep -rnE "#4533|issues/4533" kornia/ tests/ docs/` is empty in the merge tree. The syncs are gone: a `TorchDispatchMode` count over a 20-iteration fit at 4096x32, K=8 goes from `nonzero` 160 / `index_select` 160 / `mean` 160 to zero of each, and MPS wall clock at N=20000, D=32, K=8 goes 110.6/133.5 ms to 44.9/43.6 ms (best of 8, `mps.synchronize` inside the timed region, two interleaved rounds). float64 fits are bit-identical to main over 18 configurations; float32 differs sub-ULP with identical labels in 17 of 18. `torch.export` and `torch.onnx.export(dynamo=True)` both fail on main and both succeed here (ONNX: 136 nodes, `onnx.checker` clean, onnxruntime within 2.4e-07 of eager). Peak RSS is *lower* on the head (755 vs 868 MB at N=262144, K=256) — the `(N,K,D)` distance tensor already dominates the `(N,K)` one-hot.

**The float16 CI leg is not yours.** Its conclusion is `cancelled`, not `failure`: the job log stops inside `tests/feature/test_scale_space_detector.py` at 58% (14:41:07) and is cancelled 36 minutes later (15:17:49) on `Intel(R) Xeon(R) 6973P-C` — the Intel-only float16 hang in #4214. `tests/contrib` was never reached. `collector` reflects it. No manifest line is needed: `tests/contrib/test_kmeans.py` is 9 passed at float16 and 9 at bfloat16 with the profile flags.

One P2 and four P3s, all one-liners.

### [P2] The large-sum pin catches only the float16 half of the defect

`tests/contrib/test_kmeans.py:135`. Last round named `bfloat16 [51.0, 51.75]` as one of the two things the test had to catch. It does not. With `8e8ad6b7`'s `kornia/contrib/kmeans.py` restored under these tests: float16 2 failed, **bfloat16 9 passed**. The pre-fix bfloat16 error is 0.25 and `BaseTester`'s default bfloat16 tolerance is `7.8e-3 * 52 + 7.8e-3 = 0.4134`. Sweeping N over 1600, 6400, 20000 and 60000, the pre-fix error is 0.25 or exactly 0 — never above tolerance. Mutating only `kmeans.py:190` back to `cluster_counts = one_hot.to(X.dtype).sum(0)` also passes all four dtypes, so count exactness is unpinned too.

Both means are exact in every dtype on this head, so tightening the assert is enough:

```python
self.assert_close(kmeans.cluster_centers, expected.to(device=device, dtype=dtype), rtol=0.0, atol=0.0)
```

Applied in a worktree: this head passes float32/float64/float16/bfloat16; `8e8ad6b7` fails float16 2/2 and bfloat16 1/2.

### [P3] Two user-visible behaviour changes are in no fragment

`kornia/contrib/kmeans.py:202` rebinds `current_centers` instead of writing into `self._cluster_centers` row by row, which has two observable consequences:

- `fit` no longer mutates the caller's `cluster_centers` tensor. `sc = torch.tensor([[0., 0.], [1., 1.]]); KMeans(2, sc, None, 3, 0).fit(X)` leaves `sc = [[-5.5, -5.5], [10.5, 10.5]]` on main and `[[0., 0.], [1., 1.]]` here.
- A second `fit()` no longer warm-starts from the first result. `km.fit(A); km.fit(B)` with A far from B gives `[1.0, 3.0]` on main and `[0.5, 2.5]` here.

Both are improvements, but the second silently changes results. Please add a clause to `changelog.d/4545.fixed.md` next to the RNG one — or a `breaking` fragment if warm-starting was ever intended.

### [P3] The rewritten export note overclaims and contradicts its own snapshot row

`docs/export_support/cases_models.py:1126`. `torch._dynamo.explain` gives `graph_count=2, break_count=1` on this head as well as on main — the remaining break is `torch.manual_seed` at `kmeans.py:68`, not the centroid update, so "traces and exports as a single graph" is wrong for the `compile` column. And `docs/source/_data/export_support.json` still records `contrib.KMeans.fit` as `export: fail`, `onnx: fail`, `graph_breaks: 2`, `where: kornia/contrib/kmeans.py:185`, so until the snapshot is regenerated the page renders a note asserting a capability beside three columns denying it. Something like "...no data-dependent control flow, so `torch.export` and the ONNX export now succeed; the one remaining graph break is `torch.manual_seed` in `__init__`" holds under both the current and the regenerated JSON.

### [P3] `num_clusters != cluster_centers.shape[0]` now fails opaquely, and asynchronously on MPS

`KMeans(3, torch.randn(5, 2), None, 2, 0).fit(torch.randn(50, 2))` ran on main and returned `(5, 2)` centers. Here it raises `RuntimeError: Class values must be smaller than num_classes.` out of `F.one_hot` on CPU — and on MPS it raises nothing at that call and instead aborts a later, *unrelated* operation with `torch.AcceleratorError: scatter: index 4 is out of bounds for dimension with size 3`, because `one_hot` skips its host-side range check there. A `KORNIA_CHECK(cluster_centers.shape[0] == num_clusters, ...)` in `__init__`, next to the existing `num_clusters != 0` check, makes both deterministic.

### [P3] The PR body is stale

It still names the fragment `changelog.d/+kmeans-vectorize-fit.fixed.md` — the file is correctly `changelog.d/4545.fixed.md` — and the approach note, benchmark table and "Verified" list all describe the pre-fix revision.

**Not blocking, for a follow-up:** no test distinguishes float64 accumulation from float32 (mutating `kmeans.py:191` to `work_dtype = torch.float32` unconditionally passes all four dtypes); I checked the float64 bit-identity to main by hand instead, over 18 fits. The `index_add_` option from last round is still open, and the `tolerance` exit check still syncs once per iteration exactly as on main.

Suites on the head: `tests/contrib/` float32 204 passed / 17 skipped, float64 203 passed / 18 skipped; `test_kmeans.py` float16 and bfloat16 9 passed each with their profiles; MPS float32 9 passed; `KORNIA_TEST_OPTIMIZER=inductor` 10 passed. Merge is clean against `30dfbf711`. `ruff@0.16.7` check and format clean. Not verified: CUDA.


prithvirajrh:
Thanks for verifying the P1 fix so thoroughly, and for catching the rest.

- **P2 (bfloat16 not actually caught):** you're right, the tolerance was too loose. Tightened to `rtol=0.0, atol=0.0` and confirmed against the actual pre-fix revision: 3 of the 4 parametrized cases now fail there as expected (the case that still passes was never the one you flagged).
- **P3 (mutation / warm-start):** added both to the changelog fragment, next to the RNG note. Didn't file a `breaking` fragment - neither was documented behavior, and both changes are strict improvements (not mutating a caller's tensor is generally the right default), but agreed they're worth disclosing.
- **P3 (export note overclaims):** corrected to your wording - the remaining break is `torch.manual_seed` in `__init__`, not the centroid update.
- **P3 (opaque/async failure on shape mismatch):** added the `KORNIA_CHECK` in `__init__` plus a test for it.
- **PR body stale:** rewritten from scratch to describe the current state, including your independently-measured MPS numbers and the more honest CPU-at-scale finding.
- **`index_add_` at larger K:** still open for a follow-up, agreed not blocking here.

All pushed. Full suite still green locally (204 passed), lint/typecheck clean.

ducha-aiki (approved):
Review of `b302df1e7dfbd609018d02cd7f5fb7e2e9acc496` — **APPROVE**.

All remaining blockers from the previous review are resolved. The large-sum regression is exact and now catches the bfloat16 count/sum bug, the two observable `fit` behavior changes are disclosed, the export note matches the actual graph break, and a constructor check gives deterministic behavior when `num_clusters` disagrees with the supplied centers. The earlier half-precision accumulation and empty-cluster issues remain fixed.

Validation: `tests/contrib/test_kmeans.py` passed across CPU float16, bfloat16, float32, and float64 (36 passed), and the Inductor dynamo case passed. The merge tree against current `main` is clean and all 42 reported checks are green. No actionable findings remain.

PR #4546, opened 2026-09-15
fix(augmentation): honour the sampler device in linear illumination

### LINKED ISSUE #4536: bug(augmentation): RandomLinearIllumination fails on MPS/CUDA with a device mismatch — `directions` stays on CPU in LinearIlluminationGenerator
## Bug

```python
x = torch.rand(8, 3, 256, 256, device="mps")   # same on cuda
K.RandomLinearIllumination(p=1.0).to(x.device)(x)
# RuntimeError: Expected all tensors to be on the same device, but found at least two devices, mps:0 and cpu!
```

Traceback ends in `kornia/augmentation/random_generator/_2d/linear_illumination.py:114` (`gradient = sign * gain_factor * gradient`) via `augmentation/base.py:169 generate_parameters`. Cause: `forward` takes `_device, _dtype = _extract_device_dtype([self.gain, self.sign])` — the Python constructor tuples, i.e. CPU — instead of the device that `make_samplers(device, dtype)` received from `.to(...)`. `make_samplers` does move the gain and sign ranges to the requested device, so `sign` is sampled on MPS, while `gain_factor` is `.to(_device)` (CPU) and the ramps are built on CPU; the product at line 114 mixes the two. (`torch.where` at line 85 survives only because its 0-d CPU operands count as scalars.) `RandomLinearCornerIllumination` and `RandomGaussianIllumination` had the same symptom and were fixed in #3704; the plain linear generator was left out. Same class as #4426 (sampler migration / returned parameter placement); the open shear PRs for #4415 (#4507, #4519) do not touch this file.

## Fix

Take `_device, _dtype` from the generator's recorded sampler device (what `make_samplers` received) rather than from `self.gain`/`self.sign`, as the corner variant does after #3704, and add `RandomLinearIllumination` to the device test that covers the other two.

Found while timing 45 augmentations on MPS for #4521; it is the only 2-D intensity augmentation in that set that does not run on MPS at all.

rishabhsai:

`LinearIlluminationGenerator.forward` and `LinearCornerIlluminationGenerator.forward` took their placement from `_extract_device_dtype([self.gain, self.sign])`, but `self.gain` and `self.sign` are the plain Python tuples handed to the constructor, so that helper skipped them and fell back to the default device and dtype. The generators therefore ignored the device and dtype that `make_samplers` had just been given: moving either augmentation to CUDA or MPS, or calling `set_rng_device_and_dtype` as the class docstring tells you to, still returned a CPU `float32` `gradient`, so a `float64` pipeline silently got `float32` parameters and every call copied the gradient across devices in `apply_transform`. They now read `self.device` and `self.dtype`, which is what `GaussianIlluminationGenerator` already does. #4523 rewrote the sign construction in this file on 2026-09-14 and added `tests/augmentation/test_parameter_constants.py`; it says it preserved the existing returned-parameter placement, so it kept the `_extract_device_dtype` lookup and pinned the CPU result in a test, and that placement is what this changes.

Two notes on the issue text. The crash in its repro no longer reproduces on `main`: `apply_transform` does `input.add(params["gradient"].to(input))` (kornia/augmentation/_2d/intensity/linear_illumination.py:131 and :238), which copies the CPU parameters onto the input device, and #4523 rewrote the lines the traceback points at. What is left is the placement and dtype contract, which is what this PR claims. The issue also says the corner variant was fixed in #3704; it still used `_extract_device_dtype` on `main`, so both generators are changed here. Output on the default CPU `float32` path is byte-identical, checked with `torch.equal` against `main` under the same seed.

Fixes #4536

## What did you check?

`TestIlluminationPlacement` replaces `TestLinearIlluminationPlacement`, asserts the generator's own device and dtype instead of CPU, and covers the Gaussian generator too, so all three illumination generators share one contract. It fails on `main` with the test kept and the two-line source change reverted, on both axes:

```text
$ pytest tests/augmentation/test_parameter_constants.py -k TestIlluminationPlacement --dtype float64 -q
E       assert torch.float32 == torch.float64
FAILED ...::test_moved_sampler[cpu-float64-linear] - assert torch.float32 == torch.float64
FAILED ...::test_moved_sampler[cpu-float64-linear-corner] - assert torch.float32 == torch.float64
2 failed, 1 passed, 42 deselected in 0.09s

$ pytest tests/augmentation/test_parameter_constants.py -k TestIlluminationPlacement --device mps --dtype float32 -q
E       AssertionError: assert device(type='cpu') == device(type='mps', index=0)
2 failed, 1 passed, 42 deselected in 0.46s
```

With the fix:

```text
$ pytest tests/augmentation -k "illumination or parameter_constants" --dtype float32,float64 -q
130 passed, 4 skipped, 12075 deselected in 5.31s

$ pytest tests/augmentation -q
2877 passed, 3403 skipped, 30 xfailed, 5 xpassed in 18.50s

$ KORNIA_TEST_OPTIMIZER=inductor pytest tests/augmentation/test_parameter_constants.py -q
116 passed, 2 skipped in 92.31s

$ pytest --doctest-modules kornia/augmentation/ -q
84 passed in 1.10s
```

`pre-commit run --files` on the three changed files passes, and `ty check kornia` reports nothing new for `linear_illumination.py`. The CPU `float16`/`bfloat16` runs are unchanged: the one failure there, `TestRandomGaussianIllumination::test_smoke`, fails identically on `main` and is already listed in `testing/half_precision_xfails/`.

## How did AI help?

I used Claude to work through the generator and write the change and the test. I read the diff, confirmed the repro at the generator level on MPS and on `float64`, ran the red and green output above myself, and checked the default CPU `float32` output is byte-identical to `main` before trusting it.


ducha-aiki (dismissed):
Review of `e855005099b5dd2039a07e2a31f49e9d10d28f34`.

The change is needed on top of #4523 and does not duplicate it. #4523 changed how the sign is built, and that already stops the #4536 crash on main. But #4523 kept `_extract_device_dtype([self.gain, self.sign])`, which ignores the plain tuples and falls back to CPU and the default dtype. So on main, `gradient` still comes back on CPU in float32 whatever device and dtype the samplers were moved to. The PR switches both generators to `self.device`/`self.dtype`, as `GaussianIlluminationGenerator` already does. The new test fails on main. CPU float32 output is bit-identical to main, and the half-precision results only differ by rounding.

What happens on main vs this PR (same probe run in each worktree):

| probe | main | PR |
|---|---|---|
| `K.RandomLinearIllumination(p=1.0).to("mps")(x)`, x on mps | output mps, `_params["gradient"]` **cpu float32** | output mps, gradient mps float32 |
| same for `RandomLinearCornerIllumination` | gradient **cpu float32** | gradient mps float32 |
| generator `.to(device="meta")` then forward | `NotImplementedError: Cannot copy out of meta tensor; no data!` (sign drawn on meta, copied to CPU) | gradient on meta |
| `aug.set_rng_device_and_dtype(cpu, float64)`, `forward_parameters` | gradient **float32** | gradient float64 |
| generator `.to(dtype=float16 / bfloat16)` | gradient float32 | gradient float16 / bfloat16, all finite |

The crash from #4536 no longer happens on main, as the PR body says. What remains is the placement: every call copies the full B×C×H×W gradient from the device to CPU and back. So closing #4536 is correct.

Values, compared across the two trees: 2 generators × {f32, f64, f16, bf16} × `same_on_batch` {False, True} × shapes {(4,3,16,19), (0,3,5,5), (2,1,1,7), (3,2,2,2)}, same seed.
- max|main − PR| is 0.0 for f32, ≤2.5e-8 for f64, ≤1.3e-4 for f16 and ≤8.5e-4 for bf16. The half-precision differences come from computing in half precision instead of float32 and are below two ULPs at these magnitudes.
- The global RNG state after the call is identical in both trees for every case.
- The only errors are on the empty batch for the corner generator (`ValueError: Invalid input tensor, it is empty.` from `normalize_min_max`). They occur on main too.

**[P3] `directions_sampler` is not moved** — `kornia/augmentation/random_generator/_2d/linear_illumination.py:70` and `:158`

`UniformDistribution(0, 4, validate_args=False)` is built from Python ints, so it stays on CPU in float32. After `.to(dtype=torch.float64)`, `g.directions_sampler.low` reports `cpu torch.float32`, while `gain_sampler.low` reports float64. On an accelerator, directions are therefore still drawn from the CPU RNG and copied host-to-device (B int8 values) on every forward. The result is correct and the gradient ends up on the right device, so this is optional. Building the bounds with `device=device, dtype=dtype` in `make_samplers` would remove that last copy.

**[P3] Test comment is still incomplete** — `tests/augmentation/test_parameter_constants.py:31-32`

The edited comment says only the illumination generators follow the sampler device/dtype. After `.to(dtype=torch.float64)`, `ResizeGenerator` (`id="resize"`) also returns float64 parameters, while the other non-illumination entries return float32. Suggested wording: "…except for the illumination and resize generators…". Or drop the list of exceptions, since no test in this file asserts CPU placement any more.

No other open PR touches these files. The PR does not touch any code or docs that name #4536. The general placement prose in `docs/source/get-started/conventions.rst` and `kornia/augmentation/base.py` (#4426) stays accurate. The changelog claims (CPU/default-dtype result, float32 under a float64 pipeline, a copy on every call) match the probes above.

**What I ran**

Setup:
- Interpreter: `.venv` Python 3.11, torch 2.14.0. `kornia.__file__` was checked in every probe: PR worktree for PR runs, main worktree for main runs.
- torch 2.5.1 overlay for the floor run.
- Only CPU pytest runs, plus one tiny MPS probe per tree (no MPS fallback env var, alarm-guarded).

Tests on the PR worktree:
- `pytest tests/augmentation/test_parameter_constants.py --device=cpu --dtype=float32,float64`: 64 passed, 4 skipped (CUDA-only).
- Plant-and-revert of `linear_illumination.py` to main's version, `-k TestIlluminationPlacement`: 2 failed (`test_moved_sampler[cpu-float64-linear]` and `[cpu-float64-linear-corner]`, both `assert torch.float32 == torch.float64`), 4 passed (the Gaussian cases already pass on main). Source restored; worktree clean.
  - The device half of the pin is exercised by the `tests-mps` CI leg, which passes. On main, the MPS probe above gives a CPU gradient.
- `test_parameter_constants.py` plus `TestRandom{Linear,LinearCorner,Gaussian}Illumination`, CPU `--dtype=float16` and `--dtype=bfloat16`, each with `--xfail-known-failures --known-failure-profile=cpu-<dtype>`: 75 passed, 2 skipped, 1 xfailed for each dtype. The xfail is the known `TestRandomGaussianIllumination::test_smoke`.
- `KORNIA_TEST_OPTIMIZER=inductor pytest tests/augmentation/test_parameter_constants.py -k "linear or gaussian"` (f32, f64): 21 passed.
- torch 2.5.1: `test_parameter_constants.py -k Illumination --dtype=float32,float64,float16,bfloat16`: 12 passed.
- `test_augmentation.py` and `test_random_generator.py`, `-k Illumination` (f32, f64): 66 passed.

Not verified: CUDA runtime (no CUDA here). CI and the author's inductor run cover CPU compile; the compiled CUDA transfer path of the moved generator was not run.

ducha-aiki:
Maintainer edit `2c7fe204`: fixes both P3s from the review above.

**Directions sampler.** Both generators now build their `directions_sampler` bounds on the sampler device, so directions are drawn from the device RNG and no longer copied from the host on every forward. The bounds stay float32 whatever the generator dtype. That differs from the review's suggestion of `dtype=dtype`, which turned out to be unsafe: on MPS, half-precision `torch.rand` can return exactly `1.0` (torch 2.14: 24 652 in 1e8 draws for float16, 195 271 for bfloat16; float32 gives none, and CPU gives none in any dtype on 2.14 or 2.5.1). `Uniform(0, 4)` then gives exactly 4, which truncates to direction 4. With the `dtype=dtype` variant, `LinearIlluminationGenerator` on MPS returned 42 (float16) and 381 (bfloat16) all-zero gradient rows out of 204 800. `LinearCornerIlluminationGenerator` would index `base[4]` on a size-4 axis. With float32 bounds the same probe gives 0 all-zero rows in all three dtypes, and both the gradient and the sampler are on `mps:0`.

**Test comment.** Now reads "…except for the resize and illumination generators, which follow the device and dtype the generator was moved to." After `.to(dtype=torch.float64)`, `resize`, `linear`, `linear-corner` and `gaussian` are the only `GENERATORS` entries that return float64 parameters.

**New pin** `TestIlluminationPlacement::test_directions_sampler`: moves each linear generator to `meta`/float16 and asserts that the sampler bounds are on `meta` and in float32. Using `meta` lets the CPU legs check placement too. Plant-and-revert of `linear_illumination.py` to `e8550050`: both cases fail with `assert 'cpu' == 'meta'`.

**What I ran** (`.venv` Python 3.11, torch 2.14.0; `kornia.__file__` checked against the intended worktree in each probe)
- Compared with `e8550050` in a separate worktree, same seed: 2 generators × {not moved, f32, f64, f16, bf16} × `same_on_batch` {False, True} × shapes {(4,3,16,19), (2,1,1,7), (3,2,2,2), (64,1,5,5)}, 3 calls each. All 80 cases are bit-identical, and the global RNG state afterwards is identical in each case. On CPU the draw is float32 in both trees.
- `test_parameter_constants.py`, CPU f32/f64: 66 passed, 4 skipped (CUDA-only).
- The same file plus the illumination tests in `test_augmentation.py`, CPU `--dtype=float16` and `--dtype=bfloat16` with `--xfail-known-failures --known-failure-profile=cpu-<dtype>`: 77 passed, 2 skipped, 1 xfailed (the known `TestRandomGaussianIllumination::test_smoke`) for each dtype.
- `KORNIA_TEST_OPTIMIZER=inductor`, `test_parameter_constants.py -k "linear or gaussian or Illumination"` (f32, f64): 23 passed.
- `test_augmentation.py` and `test_random_generator.py`, `-k Illumination` (f32, f64): 66 passed.
- torch 2.5.1 overlay, `test_parameter_constants.py -k Illumination` (f32, f64, f16, bf16): 14 passed.
- MPS probe (no fallback env var): `RandomLinearIllumination` and `RandomLinearCornerIllumination` moved to MPS give output and `_params["gradient"]` on `mps:0`. The generators in f32/f16/bf16 return finite gradients on MPS in the requested dtype.
- `pre-commit run --all-files` and `ty check` on the module: clean.

Not verified: CUDA (no CUDA here), including whether CUDA half-precision `torch.rand` can also return 1.0. The float32 bounds avoid that question on every backend.

PR #4547, opened 2026-09-15
perf(losses): reduce total_variation over a flattened dim

### LINKED ISSUE #4526: perf(losses): total_variation reduces with `sum(dim=(-2, -1))`, a trailing-two-dims reduction that MPS runs 24x slower than CPU; `flatten(-2).sum(-1)` is identical and 4x faster there
## Problem

`kornia.losses.total_variation` sums its absolute differences with `.sum(dim=(-2, -1))`. On MPS a reduction over the trailing two dims is a pathological path: `x.sum(dim=(-2,-1))` on (1,3,1024,1024) takes 5.05 ms vs 0.21 ms on the M1 CPU, while `x.sum()` and `x.sum(dim=-1)` on the same tensor take 0.5 ms on MPS (standalone numbers in #4522). The loss is therefore slower on MPS than on CPU at 1024² even though its elementwise part is faster there.

| (1,3,1024,1024) | CPU | MPS |
|---|---:|---:|
| `total_variation` today | 8.5 ms | 11.7 ms |
| same with `.flatten(-2).sum(-1)` | 8.5 ms | **3.0 ms** |

Max abs difference between the two: 0.0.

## Fix

Replace the two `.sum(dim=(-2, -1))` with `.flatten(-2).sum(-1)` (or `.sum(-1).sum(-1)`); the `reduction` argument from #1815 is unaffected.

```python
import torch, kornia.losses as L
t = torch.rand(1, 3, 1024, 1024)
tv = (t[..., :, 1:] - t[..., :, :-1]).abs().flatten(-2).sum(-1) + (t[..., 1:, :] - t[..., :-1, :]).abs().flatten(-2).sum(-1)
print((L.total_variation(t) - tv).abs().max().item())  # 0.0
```

dchaudhari7177:

`total_variation` reduced both absolute-difference maps with `sum(dim=(-2, -1))` (or `mean(dim=(-2, -1))`). As #4526 measures, MPS handles a reduction over the two trailing dims badly: the loss is slower there than on the M1 CPU, and `flatten(-2).sum(-1)` brings it from 11.7 ms to 3.0 ms at 1024².

Both maps are now flattened once and reduced over `dim=-1`. I changed the `mean` branch as well as `sum`. It has the same two-dim reduction, and the only difference between the two branches is the reducer. The `reduction` argument, the integer-to-float handling for `mean`, and the output shapes are unchanged.

Fixes #4526.

## What did you check?

First I checked that the new form gives the same values as the old one, using `torch.equal` on CPU:

```text
float16 / float32 / float64 × shapes (1,3,1024,1024) (2,3,4,5,6) (3,1,7) (4,4) (2,3,1,9) (2,3,9,1) × sum/mean
  -> identical everywhere (the H=1 / W=1 `mean` rows are NaN on both sides: the mean of an empty map)
channels_last (8,3,64,64) sum -> identical
int64 input sum -> identical
```

`flatten(-2)` has to copy a channels-last or permuted input, so `test_tv_matches_the_two_dim_reduction` compares `total_variation` against the `dim=(-2, -1)` form for contiguous, channels-last and permuted inputs under both reductions. It guards the equivalence, not a bug, so it passes before and after this change.

```text
$ python -m pytest tests/losses/test_total_variation.py --device=cpu --dtype=float32,float64,float16
65 passed
All checks passed!
```

I don't have an MPS device, so the speedup is the issue's measurement, not mine. On CPU the two forms time the same (15.9 ms vs 15.4 ms at 1×3×1024², 20-run mean), so CPU users lose nothing.

## How did AI help?

I used an AI assistant to draft the equivalence probe, the test and this description. I ran the probe and the tests above and reviewed the diff.


ducha-aiki (approved):
Review of `16b0bed1d4c5efa2909243a3e7f38fa3da581c85`.

The change fixes #4526. It gives the same values and gradients as main, bit for bit, on the inputs people normally pass, and I measured the MPS speedup the PR body quotes from the issue on torch 2.14: 12.0 ms to 3.1 ms at `(1,3,1024,1024)`. The new test compares against the old reduction and passes on main too, as the body says. That is enough for a perf change. Two nonblocking notes follow.

**[P3] `changelog.d/4547.fixed.md:1`, `kornia/losses/total_variation.py:63`: "identical values" is only true for some layouts.**
I compared main and the PR with `torch.equal` on forward values and on input gradients from `sum`/`mean`. The grid was 12 shapes (`(C,H,W)`, `(B,C,H,W)`, rank 5 and 7, `(4,4)`, H=1, W=1, `(0,3,8,8)`, `(2,0,8,8)`, `(2,3,0,5)`, `(1,1,1,1)`, `(1,3,1024,1024)`) × {fp16, bf16, fp32, fp64} × {contiguous, channels_last, `[..., ::2, ::2]`, `transpose(-1,-2)`}, plus int32/int64/uint8. 17 of the 326 cases differ, all of them fp32/fp64 transposed inputs. The differences are rounding: at most about 1.5 ULP relative, for example `(1,3,1024,1024)` fp32 transposed `sum` differs by `0.0625`. Gradients match in all 326 cases, and fp16/bf16 values match everywhere. torch 2.5.1 shows the same pattern: only fp32/fp64 transposed values differ, and no gradient does. On MPS the contiguous `sum` also moves by 1 ULP (`0.0625` at 1024², `0.0039` at `(8,3,256,256)`). The test already compares with `assert_close` tolerances, so it is fine. Suggested wording for the fragment and the code comment: "the same values up to floating-point summation order".

**[P3] `kornia/losses/total_variation.py:65-66`: `flatten(-2)` copies when H and W are transposed in memory, which roughly doubles the CPU forward time.**
`torch.rot90(x, 1, [-2, -1])` produces this layout: it is non-contiguous, with strides `(3145728, 1048576, 1, 1024)`. The subtraction keeps that layout, so each `flatten(-2)` makes a full copy. The body's "CPU users lose nothing" holds for contiguous and channels_last inputs, but not for this one. Timings are medians of interleaved runs in one process, 30-60 reps after warmup, 4 threads:

| CPU, 4 threads | main | PR | PR/main |
|---|---:|---:|---:|
| `(1,3,1024,1024)` contiguous sum fwd | 5.92 ms | 5.78 ms | 0.98 |
| `(1,3,1024,1024)` channels_last sum fwd | 5.91 ms | 5.80 ms | 0.98 |
| `(1,3,1024,1024)` transposed sum fwd | 5.79 ms | 11.82 ms | 2.04 |
| `(1,3,1024,1024)` transposed mean fwd | 6.27 ms | 14.79 ms | 2.36 |
| `(8,3,256,256)` transposed sum fwd | 2.75 ms | 4.51 ms | 1.64 |
| `rot90` of `(1,3,1024,1024)` fwd | 5.35 ms | 12.03 ms | 2.25 |
| `(1,3,1024,1024)` transposed sum fwd+bwd | 39.8 ms | 46.0 ms | 1.16 |
| contiguous fwd+bwd, 64² / 256² / 1024² | | | 1.03 / 1.01 / 1.01 |

The single-thread runs look the same: 0.93-1.08 on contiguous/channels_last and 1.73-2.01 on transposed inputs at 256²/1024².

`.sum(-1).sum(-1)`, the alternative the issue mentions, avoids the copy. CPU transposed: 4.24 ms, against 3.78 ms for `dim=(-2,-1)` and 6.89 ms for flatten. MPS contiguous: 2.50 ms, against 2.55 ms for flatten. MPS transposed: 2.90 ms, against 5.08 ms for flatten. But it does not match main bit for bit even on contiguous fp32/fp64 inputs (`1.2e-4` on an fp32 `(2,3,64,65)` sum), and flatten does. I think keeping flatten is reasonable; this is only a note about the transposed-layout cost.

**Other checks, no finding**
- MPS speedup, torch 2.14, interleaved medians of 24 runs with `mps.synchronize` inside the timed region:
  - `(1,3,1024,1024)`: sum 12.01 → 3.08 ms, mean 12.06 → 3.03 ms.
  - `(8,3,256,256)`: sum 3.56 → 1.55 ms.
  - Transposed 1024²: 11.87 → 5.03 ms.
  - So the fragment's "avoids a slow reduction path on MPS" is accurate. `fixed` is the right fragment type, since there is no perf category. The PR commits no benchmark numbers, so the `benchmarks/README.md` contract does not apply.
- Edge cases: an empty batch or zero channels gives shape `(0,3)` / `(2,0)` on both. H=0, H=1 or W=1 with `mean` gives NaN on both.
- `torch.compile(fullgraph=True)`:
  - Matches eager for forward and fp64 gradients, with `dynamic=False` and `dynamic=True`.
  - Frame counts are the same as main (3 static / 1 dynamic over three shapes); the graph gains 2 ops for the flattens.
  - `torch.jit.script` of the function and of `TotalVariation` still works, and the scripted output equals eager.
- Closing reference is exactly #4526. The merge tree (`5bb9f0d5`) has no `#4526`/`issues/4526` hits in `kornia/`, `tests/` or `docs/`. `git merge-tree` against current main exits 0.

**What I ran**
- Setup: repo `.venv` Python 3.11 with torch 2.14.0, plus the torch 2.5.1 overlay for the equality grid and the test file. Every probe printed `kornia.__file__` and named the intended tree (PR worktree or main worktree).
- Equality grid: the same stdin probe in both worktrees, results saved and compared with `torch.equal` on values, gradients and strides.
- Timing: main's and the PR's `total_variation.py` taken from `git show` and loaded into one process, with interleaved A/B runs on CPU (1 and 4 threads) and MPS (`PYTORCH_ENABLE_MPS_FALLBACK` unset).
- `pytest tests/losses/test_total_variation.py`:
  - `--device=cpu --dtype=float32,float64`: 44 passed on torch 2.14, 44 passed on 2.5.1.
  - `--dtype=float16 --xfail-known-failures --known-failure-profile=cpu-float16`: 23 passed.
  - `--dtype=bfloat16 --xfail-known-failures --known-failure-profile=cpu-bfloat16`: 21 passed, 2 xfailed. Both xfails are existing manifest entries (`test_tv_on_3d`/`test_tv_on_4d` pred0) and still fail, so no XPASS.
  - `KORNIA_TEST_OPTIMIZER=inductor -k dynamo`: 1 passed.
- Plant-and-revert: with main's `total_variation.py` swapped into the PR worktree, `test_tv_matches_the_two_dim_reduction` passes (18 passed across fp32/fp64/fp16, 6 passed on bf16). It is a parity test by design. Worktree restored, `git status` clean.

PR #4548, opened 2026-09-15
perf(color): subsample yuv420 chroma with avg_pool2d

### LINKED ISSUE #4524: perf(color): rgb_to_yuv420 subsamples chroma with a 6-D unfold+mean that is 9x slower than CPU on MPS; avg_pool2d gives the same result 35x faster
## Problem

`kornia.color.rgb_to_yuv420` builds the half-resolution chroma planes as

```python
yuvimage[..., 1:3, :, :].unfold(-2, 2, 2).unfold(-2, 2, 2).mean((-1, -2))
```

a mean over two window dims of a 6-D strided view. On MPS that reduction is the whole cost of the op (`aten::mean.dim` 82.8 ms of 84.7 ms at 1024²), and it is 9x slower than the CPU kernel; the op ends up 7.8x slower on MPS than on the M1 CPU while every other colour conversion wins on MPS at that size.

| shape | CPU today | MPS today | CPU with `avg_pool2d` | MPS with `avg_pool2d` |
|---|---:|---:|---:|---:|
| (1,3,1024,1024) | 12.7 ms | **88.8 ms** | 4.6 ms | **2.5 ms** |
| (8,3,256,256) | 6.4 ms | 42.7 ms | – | – |

## Fix

`F.avg_pool2d(yuv[..., 1:3, :, :], 2)` is the same 2x2 box mean: max abs difference against the current output is 3.0e-8 at 1024² (float32). It is also 2.8x faster on the CPU. Same treatment applies to `rgb_to_yuv422` if it uses the same unfold pattern.

```python
import torch, torch.nn.functional as F, kornia.color as KC
img = torch.rand(1, 3, 1024, 1024)
_, uv = KC.rgb_to_yuv420(img)
uv2 = F.avg_pool2d(KC.rgb_to_yuv(img)[:, 1:], 2)
print((uv - uv2).abs().max().item())  # 2.98e-08
```

Precedent for choosing the MPS-friendly op: #3914 / #3916 (`rgb_to_hsv` max/min-with-indices vs amax/amin).

dchaudhari7177:

`rgb_to_yuv420` computed the half-resolution chroma as `unfold(-2, 2, 2).unfold(-2, 2, 2).mean((-1, -2))`, a mean over two window dims of a 6-D strided view. #4524 shows this mean is almost the whole cost of the op on MPS. It isn't cheap on CPU either. The chroma is now `F.avg_pool2d(uv, 2)`, which computes the same 2×2 box mean directly.

`avg_pool2d` takes 3-D/4-D input, and `rgb_to_yuv420` accepts `(*, 3, H, W)`, so the leading dims are folded into one batch and restored afterwards. The fold uses plain `int` lists rather than `torch.Size` arithmetic, because TorchScript refuses the latter and `test_jit` scripts this function.

`rgb_to_yuv422` is unchanged. Its `unfold(-1, 2, 2).mean(-1)` reduces over a single dim, and the slow path the issue measured is the two-dim reduction. I didn't want to change it without a measurement showing it is slow as well.

Fixes #4524.

## What did you check?

Equivalence against the old expression on CPU:

```text
shapes (3,4,6) (2,3,4,6) (2,5,3,8,4) (0,3,4,4) (1,3,1024,1024)
float16 / bfloat16  max|diff| = 0.0
float32             max|diff| <= 3.0e-08
float64             max|diff| <= 5.6e-17
```

Zero-height input raised before (`unfold`: "maximum size for tensor at dimension 2 is 0") and still raises, so no new error cases appear.

CPU timing of the chroma step at 1×3×1024², 20-run mean: **20.4 ms → 1.8 ms**. I have no MPS device, so the MPS numbers are the issue's.

`test_leading_dims_match_per_image` covers the fold: a `(2, 3, 3, 4, 6)` batch must match per-image calls. I checked that it catches a real bug by flipping the batch order before the final reshape. That fails only this test, and the other 36 `Yuv420` tests pass, since none of them compares values across leading dims. I also updated a comment in `test_unit_subsampling` that referred to the library's `unfold`.

```text
$ python -m pytest tests/color/test_yuv.py --device=cpu --dtype=float32,float64,float16
329 passed        (includes test_jit and test_dynamo)
$ python -m pytest tests/color --device=cpu --dtype=float32
444 passed, 2 skipped
$ python -m pytest --doctest-modules kornia/color/yuv.py
12 passed
All checks passed!
```

## How did AI help?

I used an AI assistant to draft the equivalence probe, the TorchScript-compatible fold, the test and this description. I ran the probe, the mutation check and the tests above and reviewed the diff.


ducha-aiki (approved):
Review of `530bbe7d6d9099998f5df781aa006be0e191a604`.

The rewrite is a faithful drop-in. I ran the same probe in the PR worktree and the main worktree on torch 2.14.0 and on the torch 2.5.1 floor. Output shapes, dtypes, contiguity, error types and values match on every input below. `rgb_to_yuv422` is left alone, and on CPU that holds up: its one-dim `unfold(-1, 2, 2).mean(-1)` is not slower than `avg_pool2d(uv, (1, 2))` (1024²: 2.28 ms vs 2.41 ms; 8×256²: 1.22 ms vs 1.56 ms, bit-identical output). The new test is a parity pin, which fits a perf change: it passes on main's `yuv.py` and fails under a batch-order mutation.

[P3] `changelog.d/4548.fixed.md`: "same values, several times faster on CPU and MPS" overstates the CPU gain for the function the note names, and "same values" is only true up to rounding. I measured interleaved medians in one process (head `rgb_to_yuv420` vs main's expression on the same `rgb_to_yuv` output):

| shape, dtype | 1 thread main → head | 4 threads main → head |
|---|---:|---:|
| (1,3,256,256) f32 | 0.63 → 0.21 ms (2.95x) | 0.34 → 0.16 ms (2.06x) |
| (8,3,256,256) f32 | 6.57 → 3.68 ms (1.79x) | 3.87 → 2.74 ms (1.41x) |
| (1,3,1024,1024) f32 | 11.23 → 4.63 ms (2.43x) | 5.97 → 3.30 ms (1.81x) |
| (1,3,1080,1920) f32 | 22.48 → 9.44 ms (2.38x) | 11.42 → 7.09 ms (1.61x) |
| (1,3,1024,1024) f16 | 16.01 → 7.92 ms (2.02x) | 10.79 → 7.14 ms (1.51x) |

The chroma step alone is 2.1-3.4x faster. I could not reproduce the PR body's "20.4 ms → 1.8 ms" at 1×3×1024²: I measured 4.52 → 1.39 ms at 4 threads and 10.13 → 3.04 ms single-threaded. Float32 outputs differ from main by up to 3.0e-8 and float64 by up to 5.6e-17. Float16 and bfloat16 match bit for bit. Suggested wording: "matches the previous output to floating-point rounding; about 1.5-2.5x faster on CPU and, per #4524, much faster on MPS". Nonblocking.

**Equivalence vs main** (same heredoc in each worktree, outputs saved and diffed):
- Forward, function and `RgbToYuv420` module, float64/float32/float16/bfloat16, shapes `(3,4,6)`, `(1,3,4,6)`, `(2,3,4,6)`, 5-D `(2,5,3,8,4)`, 6-D `(2,1,2,3,4,6)`, empty batch `(0,3,4,4)`, empty inner leading dim `(2,0,3,4,4)`, `(1,3,256,254)`, `(3,2,2)`: max |diff| on Y is 0.0 everywhere. On UV it is 5.6e-17 (f64), 3.0e-8 (f32), 0.0 (f16/bf16). The numbers are the same on torch 2.14.0 and 2.5.1, so the `avg_pool2d` half kernels exist on CPU at the floor.
- Odd H or W (`(1,3,5,4)`, `(1,3,4,5)`) raises `ShapeError` on both trees.
- Zero H or W (`(1,3,0,4)`, `(1,3,4,0)`) raises `RuntimeError` on both trees. Only the message changes, from `unfold`'s "maximum size for tensor at dimension 2 is 0" to `avg_pool2d`'s "Expected 3D or 4D ... got [1, 2, 0, 4]". No test matches on the message. Raising instead of returning empty is pre-existing on main.
- Non-contiguous inputs all match main to ≤3.0e-8: permuted NHWC→NCHW, channels_last, spatially strided `[:, :, ::2, 1::2]`, batch-strided `[::2]` and `expand`ed. uint8 input gives float32 output on both, with diff 0.0.
- Input gradients (f64/f32 at `(3,4,6)`, `(2,5,3,8,4)`; f16/bf16 at the same shapes) have max |diff| 0.0 against main on both torch versions. `gradcheck` on a 5-D f64 input passes on both.
- `torch.jit.script(rgb_to_yuv420)` on 3-D, 5-D and empty-batch inputs, and `torch.jit.script(RgbToYuv420())` on a 5-D input, give output equal to eager on both torch versions.
- `torch.compile(fullgraph=True)` with a counting backend on 2.14 compiles cleanly. Over five shapes (varying batch, rank and H/W) it builds 5 graphs with `dynamic=False` and 4 with `dynamic=True`, the same counts as main. Inductor output matches eager.
- `RgbToYuv420` has `ONNX_EXPORTABLE = False`, so ONNX is not affected. The production diff has no bare literals beyond the channel count 2, the 2×2 kernel, and `// 2`.
- Merged tree: no `#4524` or `issues/4524` hits in kornia/, tests/, docs/ or testing/. No yuv420 entries in the known-failure manifests. `closingIssuesReferences` is exactly #4524, which is covered in full.

**What I ran** — `.venv` Python 3.11 with torch 2.14.0, plus the torch 2.5.1 overlay. `kornia.__file__` was printed in every probe and named the intended worktree.
- PR worktree, `tests/color/test_yuv.py`:
  - torch 2.14.0: `--dtype=float32,float64` 222 passed; `--dtype=float16 --xfail-known-failures --known-failure-profile=cpu-float16` 115 passed; `--dtype=bfloat16` with the matching profile 115 passed; `KORNIA_TEST_OPTIMIZER=inductor -k "dynamo or jit"` 12 passed.
  - torch 2.5.1: the same float32/float64, float16 and bfloat16 runs gave 222, 115 and 115 passed.
- `pytest --doctest-modules kornia/color/yuv.py`: 12 passed.
- Plant-and-revert in the PR worktree:
  - main's `kornia/color/yuv.py` with the PR's tests: 222 passed. The new `test_leading_dims_match_per_image` is a parity pin.
  - `uv.flip(0)` inserted before the final reshape: only `test_leading_dims_match_per_image` fails (1 failed, 114 passed at f32).
  - Worktree restored clean.
- CI: 41 checks pass on `530bbe7d` (39 check runs `success`). Merge base is current main `53866348`, and `git merge-tree` is clean.
- Not verified: MPS timings (the fragment's MPS claim rests on #4524's measurements).

PR #4550, opened 2026-09-15
fix(augmentation): offset RandomMosaic boxes by W along x and H along y

### LINKED ISSUE #4468: RandomMosaic composes bounding boxes with the tile x/y offsets swapped on non-square inputs
## Describe the bug

`RandomMosaic` places the bounding boxes of each tile with the height and width swapped, so on a non-square input the boxes no longer sit on the image content they describe.

In `kornia/augmentation/_2d/mix/mosaic.py`, `_compose_images` builds each column by concatenating the `j` loop along `dim=-2` (height) and then concatenates the `i` loop along `dim=-1` (width), so tile `(i, j)` lands at `x = W * i`, `y = H * j`. The box path in `_compose_boxes` (lines 134-135) offsets the same tile with

```python
_offset[idx, 0] = batch_shapes_idx[:, -2] * i - src_box_idx[:, 0, 0]   # x offset uses H
_offset[idx, 1] = batch_shapes_idx[:, -1] * j - src_box_idx[:, 0, 1]   # y offset uses W
```

For square inputs the two agree, which is why the existing tests never see it. On a 6x8 input (tiles of 3x4 when `output_size=None`) the right-hand tile's box is shifted by 3 in x instead of 4, and the lower tile's by 4 in y instead of 3. Keypoints are unaffected because the keypoint path raises `NotImplementedError`.

## Reproduction steps

Run `RandomMosaic(p=1.0, data_keys=["input", "bbox_xyxy"])` on a batch of `(B, 3, 6, 8)` images whose pixels encode their source image and position, with a box per image covering a marked patch; compare the output box against the location where that patch's pixels actually appear in the composed image. The x offset of the second column is 3 where the pixels sit at 4.

`main` at `8fb9ad12`. The generator-side axis swap #4438 (fixed in #4459 / #4454) is separate: this swap is in the box composition and remains after that fix.

## Expected behavior

Box offsets use `batch_shapes[:, -1] * i` for x and `batch_shapes[:, -2] * j` for y, matching `_compose_images`, and a non-square regression pins a box against the pixel content of the composed image.

Refs #4407.

dchaudhari7177:

`_compose_images` concatenates the `j` loop along the height and the `i` loop along the width, so tile `(i, j)` starts at `x = W * i`, `y = H * j`. `apply_transform_boxes` offset the same tile's boxes by `batch_shapes[:, -2] * i` in x (H) and `batch_shapes[:, -1] * j` in y (W). The two axes are swapped back, with a comment naming the composition they must match. On a square input H == W, so nothing changes there. That is also why `test_numerical` (8×8) never saw the bug and still passes unchanged.

Fixes #4468.

## What did you check?

`test_boxes_follow_tile_content_on_non_square_input_4468` checks box positions against pixel content, as the issue suggests. Each image carries a 2×2 marked patch with a box drawn on it. The crop starts at each tile's centre (`start_ratio_range=(0.5, 0.5)`) and every patch straddles that centre. That way all four tile copies of every patch are partly in view and no box gets clamped to a border sliver. Every output box must then cover marked pixels only. Without the straddle, boxes from copies outside the crop end up as 1-px slivers on the border and the check can't tell a real bug from clamping, which is how my first attempt failed on square inputs too.

Red before green, stashing only `mosaic.py`:

```text
$ python -m pytest tests/augmentation/test_augmentation_mix.py -k 4468 --device=cpu --dtype=float32,float64
without the fix: 6 failed, 2 passed   ((6,8), (8,6), (4,12) fail; (8,8) passes)
with the fix:    8 passed
```

A wider probe with the same design (5 seeds, 3 images each) found 30 misplaced boxes per non-square shape before the fix, for `(6,8) (8,6) (10,14) (4,12)`, and 0 after. `(8,8)` gave 0 both times.

```text
$ python -m pytest tests/augmentation/test_augmentation_mix.py tests/augmentation/container/test_augmentation_sequential.py tests/augmentation/test_base.py --device=cpu --dtype=float32,float64
412 passed, 102 skipped, 8 xfailed
$ python -m pytest --doctest-modules kornia/augmentation/_2d/mix/mosaic.py
1 passed
All checks passed!
```

## How did AI help?

I used an AI assistant to trace the axis order through `_compose_images`, design the pixel-content check (including the straddling-patch fix to my first probe) and draft this description. I ran the probes and the tests above and reviewed the diff.


ducha-aiki (changes requested):
Review of `1320ee2eba915be7cf8dd63c8fcfb9066145231a`.

The swap is correct. `_compose_images` concatenates the `i` loop along the width and the `j` loop along the height, so tile `(i, j)` starts at `x = W*i`, `y = H*j`. The head's box offsets now match that for every tile, grid, cropping mode and entry point I tried, and they are byte-identical to main when `H == W`. The new test fails on main for every non-square shape and passes on the square one, so it is not caught by the symmetric-input trap. The only thing blocking merge is CI.

**[P2] The new test turns the blocking CPU bfloat16 job red; the half-precision manifest needs its 4 nodes** (`testing/half_precision_xfails/cpu_bfloat16.txt`)

`RandomMosaic` does not run in bfloat16 yet (#4467): `DType` has no `BFLOAT16`, which is why `TestRandomMosaic::test_numerical` and `test_p` are already listed in that manifest. The new parametrized test hits the same `KeyError`, but none of its nodes are listed. The PR's `half-precision (bfloat16)` job (conclusion `failure`) ends with:

```
FAILED ...TestRandomMosaic::test_boxes_follow_tile_content_on_non_square_input_4468[cpu-bfloat16-shape0] - KeyError: 'BFLOAT16'
  (same for shape1, shape2, shape3)
= 4 failed, 10191 passed, 3776 skipped, 562 xfailed, 5 xpassed ...
```

I get the same result locally with `--dtype=bfloat16 --xfail-known-failures --known-failure-profile=cpu-bfloat16`: 4 failed, 5 passed, 4 xfailed. With `--dtype=float16` all 12 selected tests pass and 1 xfails, so no float16 lines are needed. The fix is to add these lines next to the existing `TestRandomMosaic` entries, keeping the file sorted:

```
call	builtins.KeyError	tests/augmentation/test_augmentation_mix.py::TestRandomMosaic::test_boxes_follow_tile_content_on_non_square_input_4468[cpu-bfloat16-shape0]
call	builtins.KeyError	tests/augmentation/test_augmentation_mix.py::TestRandomMosaic::test_boxes_follow_tile_content_on_non_square_input_4468[cpu-bfloat16-shape1]
call	builtins.KeyError	tests/augmentation/test_augmentation_mix.py::TestRandomMosaic::test_boxes_follow_tile_content_on_non_square_input_4468[cpu-bfloat16-shape2]
call	builtins.KeyError	tests/augmentation/test_augmentation_mix.py::TestRandomMosaic::test_boxes_follow_tile_content_on_non_square_input_4468[cpu-bfloat16-shape3]
```

This collides with the open #4471, which adds bfloat16 to `DType` and deletes the `TestRandomMosaic` bfloat16 manifest lines. I applied #4471's `kornia/` changes on top of this head in the PR worktree (then reverted them): all 4 new bfloat16 nodes pass. Whichever PR merges second has to reconcile. If #4550 lands first, #4471 must also delete these 4 lines, or they fail as strict XPASS. If #4471 lands first, this PR needs no manifest lines at all.

**Verification of the fix** (probe on the PR worktree and the main worktree, float64). Each source image `k` gets pixel value `k*H*W + y*W + x + 1`. The permutation is forced to be distinct per output (identity in column 0, as the generator produces), so every output pixel decodes to a unique source image and position. The crop origins are integers and are set per case, including `x != y`. The boxes sit at tile corners (`[0,0,2,2]`, `[W-2,H-2,W,H]`, `[W-3,0,W,2]`) plus one interior box `[1,2,4,3]`. For each output box, the expected box is the bounding box of the output pixels that decode to its source image and source box. Boxes that are fully and partly in view are counted separately.

Totals over 5 crop origins for each case: `(0,0)`, `(W/2,H/2)`, `(W-1,1)`, `(1,H-1)`, `(W/4,3H/4)`.

| shape (H,W) | grid | mode | main: wrong boxes / checked (max dev) | head |
|---|---|---|---|---|
| (6,8) | (2,2) | slice / resample | 68/112 (2.0) | 0/112 |
| (6,8) | (2,3), (3,2) | slice / resample | 102/168 (2.0) | 0/168 |
| (6,8) | (1,2) / (2,1) | slice / resample | 12/34, 8/26 | 0 |
| (8,6) | (2,2) | slice / resample | 76/116 (2.0) | 0/116 |
| (8,6) | (2,3), (3,2), (1,2), (2,1) | both | 114/174, 114/174, 20/38, 6/26 | 0 |
| (4,12) | (2,2), (2,3), (3,2) | both | 60/116, 90/174, 90/174 (8.0) | 0 |
| (4,12) | (1,2), (2,1) | both | 8/36, 10/28 | 0 |
| (8,8) | all five grids | both | 0 | 0 |

Two more checks used the generator's own parameters: `start_ratio_range=(0.5,0.5)`, 5 seeds, B=16, keeping only outputs with distinct permutations.

- Standalone `RandomMosaic(data_keys=["input","bbox_xyxy"])`: main has 168 of 224 visible boxes wrong at both (6,8) and (4,12); head has 0.
- `AugmentationSequential(RandomMosaic(...), data_keys=["input","bbox_xyxy"])` (reachable since #4505): the same numbers, 168/224 on main and 0/224 on head. This run uses the container's inclusive `xyxy_plus` convention.
- Square (8,8) gives 0 on both trees, for both entry points.

Other checks:

- **Partial batch:** `batch_prob=[1,0,1,0]` gives correct boxes for applied rows, and the non-applied rows keep their input boxes.
- **Keypoints:** they still raise `NotImplementedError` both standalone and through the container, so they are not affected by the swap. The `bbox` and `bbox_xywh` keys go through the same `apply_transform_boxes` and get the same fix.

**The new test fails without the fix, on both axes.** Restoring main's `mosaic.py` in the PR worktree gives `6 failed, 2 passed`: shapes (6,8), (8,6) and (4,12) fail in float32 and float64, and (8,8) passes. Undoing only one axis of the fix is also caught: keeping the wrong y offset fails 1 of 4 shapes, and keeping the wrong x offset fails 2 of 4.

**Fragment.** `changelog.d/4550.fixed.md` is accurate: on non-square inputs the boxes were placed on the wrong tile position, and square inputs were unaffected. `closingIssuesReferences` is exactly #4468, which this PR fully fixes. `git grep -nE "#4468|issues/4468"` on the merge tree over `kornia/`, `tests/` and `docs/` finds nothing.

**What I ran** (`.venv` python 3.11, torch 2.14.0; `kornia.__file__` printed and matched the intended tree in each run)
- PR worktree: `pytest tests/augmentation/test_augmentation_mix.py --device=cpu --dtype=float32,float64` gave 107 passed. `tests/augmentation/container/test_augmentation_sequential.py tests/augmentation/test_base.py` gave 307 passed, 100 skipped, 8 xfailed. The `mosaic.py` doctest gave 1 passed.
- Same with torch 2.5.1: `-k Mosaic` gave 21 passed.
- Half precision, `-k Mosaic` with the cpu-float16 profile: 12 passed, 1 xfailed. With the cpu-bfloat16 profile: 4 failed (the new nodes), 5 passed, 4 xfailed.
- Plant-and-revert of `mosaic.py` (full revert and each axis alone), as described above. The PR worktree is clean afterwards.
- The pixel-decoding probe described above, run from the main and PR worktrees.
- `git merge-tree --write-tree origin/main <head>` exits 0 (no conflicts).

dchaudhari7177:
Added the four nodes in `testing/half_precision_xfails/cpu_bfloat16.txt`, next to the existing `TestRandomMosaic` entries. The file is ordered by node id, and that order still holds.

`pytest tests/augmentation/test_augmentation_mix.py -k Mosaic --device=cpu --dtype=bfloat16 --xfail-known-failures --known-failure-profile=cpu-bfloat16` now gives 5 passed, 8 xfailed (it was 4 failed, 5 passed, 4 xfailed before).

On the collision with #4471: if this lands first, #4471 has to drop these four lines along with the other `TestRandomMosaic` ones, or they will report strict XPASS. If #4471 lands first, I'll rebase and take the lines back out.

ducha-aiki (dismissed):
Review of `8946d239`.

The manifest ask is done correctly. The four lines byte-match what `--collect-only` prints, the file is still `LC_ALL=C`-sorted on the node-ID field, and I confirmed both directions: with the lines the nodes report `4 xfailed`, with them stripped `4 failed` each with `KeyError: 'BFLOAT16'`. float16 needs nothing (`4 passed`); float32/float64 `8 passed`.

I re-verified the fix from pixel content rather than from the code. First the composition law itself: I rebuilt each output row's super-image as *tile (i, j) at x = W·i, y = H·j*, cropped it at `params["src"]` and compared to `out_image` — 0 mismatches over shapes (6,8), (8,6), (4,12), (8,8) × grids (2,2), (2,3), (3,2), (3,3) × 4 seeds, on both trees. So the fix's comment is accurate and `batch_shapes[:, -2]`/`[:, -1]` really are H/W (`MosaicGenerator` sets `batch_shapes = batch_shape[1:]`).

Then the boxes, with a distinct marker value per source, requiring every returned box to cover a non-empty, single-valued, all-positive region: **head 0 bad of 4600 checked**, over 6 shapes × 6 grids (including (3,3) and the degenerate (1,2)/(2,1)) × both cropping modes × 5 seeds, batch 4. Main is wrong on every non-square shape — e.g. 120 of 180 at (4,12) with grid (3,3) — and correct on (8,8) on both trees, so the square-input trap is avoided. With `p=0.5`, so applied and skipped rows sit in the same batch, over 30 seeds × B=6: head 0 bad of 376 and no skipped row's boxes move; main 188 of 376. Same picture at float16, float32 and float64. Restoring main's `mosaic.py` gives `6 failed, 2 passed`, and each axis alone is caught (x-only wrong → 4 failed, y-only → 2 failed).

**Approving, with one test-strength nit and the merge-order item.**

### [P3] The new test is symmetric under a tile mis-assignment, so it cannot catch one

Every source gets the same box `[cx-1, cy-1, cx+1, cy+1]` and the same marker position, so the *set* of returned boxes is invariant under a permutation of the tile-to-source mapping. Swapping `i` and `j` on `mosaic.py:136-137` changes `out_box`, yet your test still reports `8 passed`. Give each source a slightly different patch — e.g. `image[k, 0, cy-1:cy+1, cx-1:cx+1+(k % 2)]` with the box widened to match — and the same mutation fails (8 bad boxes of 96 in my probe). That closes the symmetry without changing the straddle design.

### [Merge order] #4471 first

Your note on #4471 is right, and it is a real git conflict rather than a silent one: merging both onto current main leaves `testing/half_precision_xfails/cpu_bfloat16.txt` as `UU` with 2 markers. Keeping both edits makes the bfloat16 leg red with 4 strict XPASS; dropping your four lines gives `58 passed` and a bfloat16 augmentation count of 56, which is the number #4471 writes into the README. So the cheaper order is **#4471 first** and this PR then simply drops the four lines. I have asked for that order there.

Suites: full mix file float32/float64 107 passed; bfloat16 profile 26 passed / 31 xfailed (27 pre-existing + your 4); float16 profile 51 passed / 6 xfailed; containers + `test_base.py` 307 passed; `mosaic.py` doctest 1 passed; ruff clean, no line over 120. `changelog.d/4550.fixed.md` is accurate, `closingIssuesReferences` = `[4468]`, and `grep -rnE "#4468|issues/4468"` is empty.

PR #4551, opened 2026-09-15
test(camera): draw project/unproject focal lengths and depth away from zero

### LINKED ISSUE #4399: test_project_unproject (sensors.camera) draws its depth from `torch.rand`, so the half-precision verdict depends on the draw
`tests/sensors/camera/test_camera_model.py::TestPinholeCamera::test_project_unproject` builds its points with `torch.rand((batch_size, 3))` and its intrinsics with `torch.rand(batch_size, 4)`, unseeded, then asserts `unproject(project(p), p.z) ≈ p` at the fixture's dtype tolerance. In float16 and bfloat16 the verdict depends on the draw.

## Evidence (`main` `147ecc4e`, torch 2.14.0, cpu)

Five runs of the node alone with `--dtype=float16,bfloat16`:

```
6 passed
2 failed, 4 passed   # float16 miss 0.00195 (atol 0.001), bfloat16 miss 0.129 (atol 0.0078)
1 failed, 5 passed   # bfloat16 miss 0.0156
1 failed, 5 passed   # bfloat16 miss 0.0254
6 passed
```

A 40-seed sweep of the same computation (batch 5, `torch.rand` points, pinhole `CameraModel`):

| dtype | draws over the test tolerance | worst miss (at min z) |
|---|---|---|
| float16 (atol 1e-3) | 13 / 40 | 0.0161 (z = 0.049) |
| bfloat16 (atol 7.8e-3) | 25 / 40 | 0.1055 (z = 0.049) |

## Mechanism

`project` divides by `z` and `unproject` multiplies by it; with `z ~ U(0, 1)` the round trip's relative error in half precision grows like `1 / z` and is unbounded. The test's claim is true for well-conditioned depth; the fixture does not provide one.

## Why it matters for CI

The half-precision CI legs seed each node deterministically, so in CI the test is not flaky: the draw for `[cpu-bfloat16-2]` happens to fail and is listed in `testing/half_precision_xfails/cpu_bfloat16.txt` (line 563), while the other five nodes drew passing samples and are not listed. That manifest line records a seed, not a defect; `pixi run test-half` (unseeded) shows the real picture above. Same class as #3943 in `tests/geometry/test_conversions.py`.

## Fix

Sample the depth away from zero (e.g. `points[..., 2] += 1.0`, i.e. `z ∈ [1, 2)`), keep the dtype tolerances, and delete the `cpu_bfloat16.txt` line in the same PR after running `--known-failure-profile=cpu-bfloat16 --verify-known-failures` on the file.

dchaudhari7177:

`TestPinholeCamera::test_project_unproject` draws its intrinsics and points from `torch.rand`. Its half-precision verdict therefore depends on the draw, and the one seeded bfloat16 node that happened to fail was being carried by `testing/half_precision_xfails/cpu_bfloat16.txt`.

The test now keeps `fx, fy` and `z` in `[1, 2)` (`params[:, :2] += 1.0`, `points[..., 2] += 1.0`), and the manifest line for `[cpu-bfloat16-2]` is removed. The in-place shifts don't change how many random numbers the test consumes, so every other draw stays the same.

**This differs from the fix the issue proposes, and I measured why.** Shifting only the depth made the round trip *worse* on this model. Pinhole `unproject` computes `(u - cx) / fx * z`, so a focal length near zero amplifies the rounding in `u`, and a larger `z` scales that error up. Replaying the test's computation with `BaseTester.assert_close`'s dtype tolerances (batch sizes 1/2/5):

```text
draws over tolerance                    float16    bfloat16   float32   float64
current fixture          (300 draws)     50         51         3         0
z in [1, 2) only         (300 draws)    117        119         3         0
fx, fy in [1, 2) only    (1200 draws)     4         18         0         0
fx, fy and z in [1, 2)   (1200 draws)     0          0         0         0
```

The current fixture also fails float32 on about 1 draw in 100, which the focal-length shift fixes as well. The tolerances are unchanged.

Fixes #4399.

## What did you check?

Under CI's seeded profile (`--known-failure-profile ... --xfail-known-failures`):

```text
before  cpu-bfloat16: 2 passed, 1 xfailed   ([cpu-bfloat16-2] via the manifest)
after   cpu-bfloat16: 3 passed              (manifest line removed)
before  cpu-float16:  3 passed
after   cpu-float16:  3 passed
```

Unseeded, 5 consecutive runs of the node on `--dtype=float16,bfloat16,float32,float64` gave 12 passed each time.

I did not run `--verify-known-failures`, because it requires the full `tests/` target. The CPU half-precision jobs will verify that the manifest is complete with the line removed.

No changelog fragment: this changes a test and its xfail manifest only, so could it get the `no-changelog` label?

```text
$ uvx ruff@0.16.6 check / format --check tests/sensors/camera/test_camera_model.py
All checks passed!
```

## How did AI help?

I used an AI assistant to write the draw sweeps that compared the candidate fixtures and to draft this description. I ran the sweeps and the profile runs above and reviewed the diff.


ducha-aiki (changes requested):
Review of `6902e500ec14cc1a1d27680fb5ba0a4624c9a7a9`.

Moving `fx, fy` as well as `z` into `[1, 2)` is the right diagnosis. My sweep confirms that the fix #4399 suggested (shifting only `z`) makes things worse. Removing the `[cpu-bfloat16-2]` manifest line is required: I put the line back on the PR head and the node reports `XPASS(strict)`. The change also hides no untracked defect. The `z = 0` behaviour belongs to #4267, which is already covered by the `CameraModelBase` docstring and pinned by `test_wart_projection_differs_from_geometry_at_small_depth_4267`. A focal length near zero is just a degenerate camera. One gap remains: the round trip still fails on some draws.

**[P2] The round trip still depends on the draw in both half dtypes** (`tests/sensors/camera/test_camera_model.py:97-99`, `:106`)

The new comment says the property "holds there on every draw (#4399)", and the PR closes #4399. #4399 is about exactly this: the unseeded (`pixi run test-half`) verdict depending on the draw. I replayed the test body (same `_make_rand_data` draw, same shifts, `BaseTester.assert_close` defaults) over `torch.manual_seed(0..N-1)`:

```text
torch 2.14.0, PR fixture, draws over tolerance
batch  float16     bfloat16     float32   float64
1      -           1/20000      -         -
2      -           4/20000      -         -
5      6/20000     21/20000     0/20000   0/20000
worst miss (batch 5): float16 0.001953125 (atol 0.001), bfloat16 0.015625 (atol 0.0078), float32 2.38e-07
```

Reproduction: batch 5, bfloat16, seed 36 on torch 2.14.0 gives `Greatest absolute difference: 0.015625 at index (1, 0) (up to 0.0078 allowed)`. Float16 seed 4468 gives `0.001953125 (up to 0.001 allowed)`. On the torch 2.5.1 floor the failing seeds are different but the rate is similar: 2/5000 in bfloat16 and 1/5000 in float16 at batch 5.

The worst miss is exactly `2 * torch.finfo(dtype).eps` in all three dtypes:

- `u = fx * x / z + cx` lands in `[1, 4)`, where it is rounded to one ULP.
- `unproject` scales that rounding by `z / fx`, which is below 2.
- The default half-precision tolerances are about 1 eps, so this fixture cannot be tolerance-proof by drawing values alone.

Two ways to make the claim true:

1. Keep the fixture and floor the tolerance at the bound. Use the default tolerances, but `atol = max(default_atol, 4 * torch.finfo(dtype).eps)`. I measured that variant at 0/20000 failures at batch 5 in float16, bfloat16, float32 and float64. Add a comment giving the `ULP(u) * z / fx` bound.
2. Keep the default tolerances, reword the comment to say the miss rate is reduced rather than eliminated, and use `Refs #4399` instead of `Fixes`.

Option 1 is what closing the issue needs.

For reference, the fixture comparison at 6000 draws per dtype (batch sizes 1/2/5 × 2000 seeds each):

```text
                 float16   bfloat16   float32   float64
current fixture  984       1018       3         0
z in [1, 2) only 2381      2426       14        0
fx, fy only      2         24         0         0
fx, fy and z     0         7          0         0
```

The float32 rate on the current fixture is 3 in 6000 here, not the body's "about 1 draw in 100". This does not affect the change.

**Release notes.** This is a test-only change, and the body explains why there is no fragment. "Check changelog fragments" fails with `no changed changelog.d/4551.(added|fixed|breaking).md fragment found`. I applied the `no-changelog` label, so that check should pass on its next run.

**What I ran** (`.venv` python 3.11, torch 2.14.0; `kornia.__file__` checked to be the PR worktree or main worktree in each probe; torch 2.5.1 overlay where noted):

- The seed sweeps above, run as heredoc probes from the PR worktree root. They replicate the test body, and the library code is identical on main and the PR because the diff touches only tests.
- PR worktree, `test_project_unproject`, `--dtype=bfloat16 --xfail-known-failures --known-failure-profile=cpu-bfloat16`: 3 passed. The same run with the manifest line planted back: 1 failed `[XPASS(strict)]`, 2 passed. The planted line was reverted and `git status` is clean.
- Main worktree, same profile: 2 passed, 1 xfailed (`[cpu-bfloat16-2]`).
- PR worktree, `--known-failure-profile=cpu-float16`: 3 passed. The same node on torch 2.5.1 in the bfloat16 profile: 9 passed (`-k test_project_unproject`).
- PR worktree, `tests/sensors/camera/test_camera_model.py`: float32,float64 104 passed, 7 skipped; cpu-bfloat16 profile 52 passed, 4 skipped; cpu-float16 profile 52 passed, 4 skipped.
- `git merge-tree --write-tree origin/main` at `53866348`: clean. The only `#4399` hit in the merged tree is the new test comment.

dchaudhari7177:
Took option 1 in `2d6de84`: kept the fixture and floored the tolerance at the bound, so `Fixes #4399` is accurate.

- `rtol, atol = _DTYPE_PRECISIONS[dtype]`, then `assert_close(..., rtol=rtol, atol=max(atol, 4 * torch.finfo(dtype).eps))`. That leaves the default tolerance in place for float32 and float64 and only raises it in the half dtypes, where the default is about 1 eps.
- The comment now gives the `ULP(u) * z / fx` bound and no longer says the default tolerance holds on every draw.

I re-ran your sweep here on torch 2.5.0+cpu: the test body over `torch.manual_seed(0..4999)` for batch sizes 1, 2 and 5 in each dtype.

```text
              default atol   floored atol   (draws over tolerance, 15000 per dtype)
float16       1              0
bfloat16      6              0
float32       0              0
float64       0              0
```

`tests/sensors/camera/test_camera_model.py`: float32,float64 104 passed, 7 skipped; the cpu-bfloat16 and cpu-float16 profiles each 52 passed, 4 skipped. ruff 0.16.6 format and check are clean.

ducha-aiki (approved):
Review of `2d6de84f`. Option 1 is in and it holds up. **Approving**, with one prose correction.

I re-measured with the real test method rather than a replay, over `manual_seed(0..2999)` × batch 1/2/5: **0 failures in 9000 draws per dtype**, in float16, bfloat16, float32 and float64. I then went past seeds and swept 2 000 000 fixture rows per half dtype directly (6 000 000 compared components): **0 over the floored tolerance**, against 155 (float16) and 339 (bfloat16) over the *default* atol — so the flake was real and the floor removes it rather than reducing it. Worst miss `2.5 * eps`, and `max(diff / allowed) = 0.50`, i.e. a 2x margin.

I also checked the case the tolerance rule makes tightest, since `atol + rtol*|x|` is tightest at the smallest element: restricted to the ~4000 components with `|x| < 1e-3`, where the verdict is atol-only, the worst miss is `0.48 eps` (float16) and `0.13 eps` (bfloat16) — 0 over tolerance. So the floor is safe at the smallest magnitude the fixture can draw, not only the typical one.

The floor is correctly scoped: `4*eps` exceeds the default atol only for float16 (`0.00391 > 0.001`) and bfloat16 (`0.03125 > 0.0078`); float32 (`4.77e-7 < 1e-5`) and float64 are untouched. All four injectable dtypes are keys of `_DTYPE_PRECISIONS`, so the bare `[dtype]` index cannot `KeyError`. Removing `[cpu-bfloat16-2]` is required: planted back on the head it reports `1 failed [XPASS(strict)], 2 passed`, while main gives `2 passed, 1 xfailed`. No other manifest line references `test_camera_model.py` in either profile. Seeding is clean — `_make_rand_data` draws `torch.rand(batch_size, 4)` on **CPU** then `.to(device)`, and the test adds no `manual_seed`, so it neither reseeds CUDA/MPS nor leaks state. `Fixes #4399` is accurate now.

The floor does not make the pin decorative: planted faults in `AffineTransform.undistort` still fail `TestPinholeCamera::test_project_unproject` — an fx/fy swap and a cx/cy swap broadly, a 0.5% scale error 6 of 12 nodes, 0.05% still 6 of 12. Worth knowing for later: at bfloat16 the floored atol is ~3% of the compared values, so a planted `* 1.02` systematic error is caught at float16, float32 and float64 but *not* at bfloat16. That is the price of option 1, which is what I asked for, not a reason to change it.

### [P3] The comment's arithmetic is wrong in three details, though the conclusion is right

`tests/sensors/camera/test_camera_model.py:97-108`:

- `u` lands in **[0, 3)**, not `[1, 4)`. `cx` and `cy` are not shifted, so they stay in `[0, 1)`, and `fx * x / z` stays below 2. Measured over 200 000 rows: `u` in `[0.0037, 2.887]`, `v` in `[0.0016, 2.906]`.
- Because `u` can sit in `[2, 3)`, `ULP(u)` reaches `2 * eps`, and with `z / fx` reaching 2 the comment's own factors give a bound of `4 * eps` — exactly the floor, i.e. **no margin as written**. The measured worst case is `2.5 * eps`, which is where the real margin comes from. Please state the bound as `4 * eps` and the measurement as `2.5 * eps`, so the floor is visibly above what is observed rather than equal to what is derived.
- "Keep both in `[1, 2)`" is really `[1, 2]`: `torch.rand(...).to(dtype)` rounds onto 1.0 in the half dtypes, so `fx`, `fy` and `z` all reach exactly 2.0 here (and `z / fx` reaches 2.0).

No code change needed, and the `no-changelog` label is already on. Suites: `tests/sensors/camera/test_camera_model.py` float32/float64 104 passed / 7 skipped; each half profile 52 passed / 4 skipped; ruff clean; `grep -rnE "#4399|issues/4399"` finds only the new comment.

PR #4552, opened 2026-09-15
test(calibration): give TestDistortPoints.test_jit a well-conditioned camera

### LINKED ISSUE #4400: test_distort.py::TestDistortPoints::test_jit[cpu-float16] compares `nan` to `nan` on ~1 in 40 draws
`tests/geometry/calibration/test_distort.py::TestDistortPoints::test_jit` compares `distort_points` eager against `torch.jit.script` on unseeded `torch.rand` inputs: points, a random `K`, a random `new_K` and four random coefficients. In float16 the random `K` (entries in `[0, 1)`, so `fx, fy < 1`) plus the polynomial distortion overflows on some draws, both outputs become `nan`, and `assert_close` (`equal_nan=False`) fails even though eager and scripted agree bitwise.

## Evidence (`main` `147ecc4e`, torch 2.14.0, cpu float16)

Eight runs of the node alone: 2 failed, 6 passed. Failing output:

```
Mismatched elements: 2 / 2 (100.0%)
Greatest absolute difference: nan at index (0, 0, 0) (up to 0.001 allowed)
```

Seed sweep 0..39 of the same inputs: one draw (seed 32, `fx = 0.414`, `fy = 0.648`) gives eager `[nan, nan]`, and `torch.equal(eager.view(torch.int16), scripted.view(torch.int16))` is `True` for it. The property the test states (script == eager) holds on every draw; the fixture makes the comparison undefined on about 1 in 40.

The node is not in `testing/half_precision_xfails/cpu_float16.txt`, so the seeded CI leg has been drawing a passing sample; `pixi run test-half` (unseeded) sees the failure. Same class as #3943.

## Fix

Use a well-conditioned fixture (`fx, fy` of order 100, principal point inside the image, coefficients ≤ 0.1, points in `[0, 1)`), or pass `equal_nan=True` if the intent is only "script matches eager". Seeding alone would hide the conditioning problem behind one lucky draw. (`torch.jit.script` also emits a `FutureWarning` on torch 2.14; that is the separate scripting-deprecation question and not part of this issue.)

dchaudhari7177:

`TestDistortPoints::test_jit` compared eager `distort_points` against `torch.jit.script` on a random `K`, `new_K` and four random coefficients. With `fx, fy < 1`, the distortion polynomial overflows float16 on some draws. Eager and scripted then both return `nan`, and `assert_close` (`equal_nan=False`) fails even though the two agree.

The fixture now follows the issue's first suggestion. `K` and `new_K` come from the file's existing `_k_asymmetric` camera (`fx = fy = 100`, principal point `(4, 3)`), the coefficients are `0.1 * torch.rand(1, 4)`, and the points stay in `[0, 1)`. I didn't use `equal_nan=True`, because a draw where both sides are `nan` compares nothing. The comparison stays strict.

Fixes #4400.

## What did you check?

Replaying the test's inputs for 400 seeds per dtype, counting draws where eager returns `nan` and where `BaseTester.assert_close(eager, scripted)` fails:

```text
                old fixture (nan / fail)   new fixture (nan / fail)
float16         45 / 45                    0 / 0
bfloat16         1 / 1                     0 / 0
float32          0 / 0                     0 / 0
float64          0 / 0                     0 / 0
```

The new coefficients still exercise the distortion. For one draw, the distorted point is `(0.7891, 0.2942)` against `(0.7576, 0.2793)` with zero coefficients, so the test isn't comparing an identity.

```text
$ python -m pytest tests/geometry/calibration/test_distort.py -k "TestDistortPoints and test_jit" --device=cpu --dtype=float16,bfloat16,float32,float64
8 unseeded runs: 4 passed each
$ ... --dtype=float16 --known-failure-profile=cpu-float16 --xfail-known-failures   -> 1 passed
$ ... --dtype=bfloat16 --known-failure-profile=cpu-bfloat16 --xfail-known-failures -> 1 passed
$ python -m pytest tests/geometry/calibration/test_distort.py --device=cpu --dtype=float16,float32,float64
75 passed
All checks passed!
```

The node has no line in either half-precision manifest, so there's nothing to delete there.

No changelog fragment, since this changes a test fixture only. Could it get the `no-changelog` label?

## How did AI help?

I used an AI assistant to write the seed sweep comparing the two fixtures and to draft this description. I ran the sweep and the tests above and reviewed the diff.


ducha-aiki (changes requested):
Review of `d985f8b469b45abcd9084dc7bea818ad0dc8e7aa`.

The new fixture removes the `nan` comparisons (#4400), and `_k_asymmetric` is a good choice of camera. The problem is that the points stay in `[0, 1)` pixels. With `cx = 4`, `cy = 3` and `fx = fy = 100`, that puts the normalized coordinates at `|x|, |y| < 0.04`. There `r2 < 0.0025` and `r4 < 6e-6`, so the radial terms fall below the comparison tolerance, and the scripted-vs-eager check no longer covers them. No manifest line refers to this node in either half-precision profile, so there is nothing to change there.

**[P2] `test_jit` no longer sees `k1` or `k2`** (`tests/geometry/calibration/test_distort.py:334-337`)

The old test and the new one both pass 4 coefficients (`k1, k2, p1, p2`), so the set of coefficient kinds is unchanged. What changed is whether each coefficient can move the output past `assert_close`. To measure that, I compared eager `distort_points` with the same call with one coefficient zeroed, which is what a scripted path that dropped that term would return. 1000 seeds per dtype:

```text
torch 2.14.0         old fixture                       PR fixture
                     nan/fail   k1   k2   p1   p2      nan/fail  k1   k2   p1   p2
float16              93/93      868  822  887  897     0/0       0    0    953  964
bfloat16             5/5        811  815  860  861     0/0       0    0    703  730
float32              0/0        986  970  985  980     0/0       917  0    1000 1000
float64              0/0        998  982  991  994     0/0       974  0    1000 999
```

The body's "the new coefficients still exercise the distortion" holds for the tangential terms only.

Plant-and-revert in the PR worktree: I made the scripted path drop `k2` (`if torch.jit.is_scripting(): k2 = torch.zeros_like(k2)` in `distort_points`). The PR's `test_jit` still passed on `--dtype=float16,bfloat16,float32,float64` (4 passed). Changing only the points to `100 * torch.rand(1, 1, 2, device=device, dtype=dtype)` makes the same plant fail on all four dtypes (4 failed, e.g. `Greatest absolute difference: 3.10 ... (up to 1e-05 allowed)`).

Suggested fix: `points = 100 * torch.rand(1, 1, 2, device=device, dtype=dtype)`, keeping `_k_asymmetric` and `0.1 * torch.rand(1, 4)`.

- That gives `x` in `[-0.04, 0.96]`.
- Over 2000 seeds per dtype it produced 0 non-finite outputs and 0 eager/scripted mismatches.
- It caught a dropped coefficient at these rates: k1/k2/p1/p2 at 1876/1652/1975/1979 (float16), 1449/1133/1871/1863 (bfloat16), 1971/1863/1997/1997 (float32).
- The worst-case corner (points at 100 px, every coefficient 0.1) stays finite in float16, with `max |out| = 207.8`, and every dropped coefficient is caught there.
- Unplanted, 5 unseeded runs gave 4 passed each.

Update the comment's last sentence to match.

**Release notes.** This is a test-only change, and the body explains why there is no fragment. "Check changelog fragments" fails with `no changed changelog.d/4552.(added|fixed|breaking).md fragment found`. I applied the `no-changelog` label, so that check should pass on its next run.

**What I ran** (`.venv` python 3.11, torch 2.14.0; `kornia.__file__` checked to be the PR worktree in each probe; torch 2.5.1 overlay where noted):

- Main worktree, `TestDistortPoints::test_jit --dtype=float16`, 20 unseeded runs: 1 failed, 19 passed. This is the #4400 flake.
- The seed sweeps and the plant above, run as heredoc probes and pytest from the PR worktree root. Both the plant and the fixture edit were reverted, and `git status` is clean.
- PR worktree, `tests/geometry/calibration/test_distort.py`: float32,float64 51 passed; cpu-float16 profile 27 passed; cpu-bfloat16 profile 27 passed. `test_jit` on torch 2.5.1, all four dtypes: 4 passed.
- `git grep test_distort.py origin/main -- testing/`: no manifest entries.
- `git merge-tree --write-tree origin/main` at `53866348`: clean. The only `#4400` hit in the merged tree is the new test comment.

dchaudhari7177:
Applied your suggestion in `6f7712f`: `points = 100 * torch.rand(1, 1, 2, ...)`, still with `_k_asymmetric` and `0.1 * torch.rand(1, 4)`. The comment now says why the points span 100 px.

Checked on torch 2.5.0+cpu:

- **Plant and revert.** I zeroed one coefficient at a time in the scripted path only (`if torch.jit.is_scripting(): dist = ...` after the pad in `distort_points`) and ran `test_jit` on float16, bfloat16, float32 and float64. Dropping k1, k2, p1 or p2 each gives **4 failed**. Unplanted, it's 4 passed. `distort.py` is restored and `git status` shows only the test file.
- **Seed sweep**, 2000 seeds per dtype with the new fixture: 0 non-finite outputs and 0 eager/scripted mismatches under the dtype's default tolerance, in all four dtypes.
- `tests/geometry/calibration/test_distort.py`: float32,float64 51 passed; cpu-float16 and cpu-bfloat16 profiles 27 passed each. ruff 0.16.6 format and check are clean.

ducha-aiki (changes requested):
Review of `6f7712fe`.

The points change does what it was supposed to. The #4400 flake is real and gone — main's `test_jit` at float16 fails **4 of 30** unseeded runs, the head fails **0 of 25** across all four dtypes — and both coefficients are observable again. I planted scripted-only faults one at a time: dropping k1 fails all four dtypes, dropping k2 fails 3-4, dropping p1 or p2 fails all four, zeroing both tangential terms fails all four, and a 1% k1 perturbation fails float64 every time. The k2 plant that passed 4/4 last round is now a failure, so that ask is discharged. A 2000-seed sweep on the new fixture gives 0 non-finite outputs per dtype and detection rates k1/k2/p1/p2 of 1876/1652/1975/1979 (float16) and 1971/1863/1997/1997 (float32).

**The CI red is not yours.** `half-precision (float16) / tests (3.11, 2.9.1)` has conclusion **`cancelled`**, not `failure`: its `Run tests` step printed its last line at 09:36:05 inside `tests/feature/test_scale_space_detector.py` at 58% and then went silent for 43 minutes until the run timed out at 10:19:03, on an `Intel(R) Xeon(R) 6973P-C` runner. That is the Intel-only float16 hang in #4214, and this PR only touches `tests/geometry/calibration/test_distort.py`. `collector` is just the aggregator reflecting it. A rerun should clear both.

One blocker, and it is partly my fault.

### [P2] `K` and `new_K` are now the same matrix, so `test_jit` can no longer see a scripted-only intrinsics swap — main's fixture could

I prescribed `new_K = _k_asymmetric(device, dtype)` last round without checking that argument, so this is my miss as much as yours. `distort_points` normalizes with `new_K` and denormalizes with `K`, so the two play opposite roles, and with `K == new_K` a scripted path that mixed them up returns the identical answer. I planted `if torch.jit.is_scripting(): K, new_K = new_K, K` after the `new_K` validation:

```
your fixture:                       test_jit -> 4 passed   (float16, bfloat16, float32, float64)
new_K = 0.9 * _k_asymmetric(...):   test_jit -> 4 failed   (all four dtypes)
```

That one-line change costs nothing and is slightly better on every other axis: over 2000 seeds per dtype, 0 non-finite outputs; coefficient detection 1893/1712/1979/1984 (float16) and 1976/1887/1998/1997 (float32) — the same or better than the current fixture; the K/new_K swap caught 2000/2000 in every dtype; 0 of 10 unseeded `test_jit` runs failing; and 27 passed under each half profile.

The eager role distinction *is* pinned by `test_convention_new_K_normalizes_and_K_denormalizes`, but nothing in the file would catch a **scripted-only** swap — which is exactly what `test_jit` exists for. Please also mention the distinct `new_K` in the comment.

### [P3] The comment's last clause reads as unconditional

"k1 and k2 still move the output past the tolerance" is true per-draw at 93%/83% in float16 and 72%/57% in bfloat16 — the scripted-only k2 drop fails bfloat16 in only 2 of 5 unseeded runs, while float32 and float64 catch it every time (my sweep: k2 observable in 1652/2000 at float16, 1133/2000 at bfloat16, 1863/2000 at float32). That is fine as coverage — a huge improvement on the prior `0/4` — but "k1 and k2 move the output past the tolerance on most draws, and on essentially every float32/float64 draw" would match what I measure.

Suites: `tests/geometry/calibration/test_distort.py` float32/float64 51 passed; cpu-float16 profile 27 passed; cpu-bfloat16 profile 27 passed. No manifest line references this node in either profile. ruff clean; `grep -rnE "#4400|issues/4400"` finds only the new comment; `closingIssuesReferences` = `[4400]` and `Fixes #4400` is accurate. The `no-changelog` label is on.


dchaudhari7177:
Done: `new_K = 0.9 * _k_asymmetric(device, dtype)`, and the comment now names the distinct `new_K` and states the k1/k2 claim as measured (most draws, and essentially every float32/float64 draw).

With your plant (`if torch.jit.is_scripting(): K, new_K = new_K, K` after the `new_K` validation), `test_jit` fails 4/4 across float16/bfloat16/float32/float64. Without it, three runs give 4 passed each. `test_distort.py` float32/float64: 51 passed.

ducha-aiki (approved):
Re-review of `6059946ea3d5`.

The JIT fixture now keeps `new_K` distinct from `K`, so it observes their opposite normalization/denormalization roles as well as every distortion coefficient. The proposed scripted-only `K, new_K` swap is caught in all four dtypes, while the unmodified test remains finite and passes in float16, bfloat16, float32 and float64. The comment now states the measured coefficient sensitivity accurately.

Validation: `pytest tests/geometry/calibration/test_distort.py::TestDistortPoints::test_jit --device=cpu --dtype=float16,bfloat16,float32,float64` (4 passed); clean real merge tree against `origin/main` `3a6a1e333a39`.

PR #4556, opened 2026-09-15
fix(geometry): honor align_corners in the 3D coordinate normalization

### LINKED ISSUE #4503: normal_transform_pixel3d and normalize_homography3d have no align_corners, so warp_affine3d(align_corners=False) is wrong (3D half of #3904)

VanniLeonardo:

The 3D half of #3904. `normal_transform_pixel3d` and `normalize_homography3d` always normalized with the corner-aligned mapping, and `create_meshgrid3d` always built a corner-aligned grid, whatever `align_corners` the 3D warps went on to pass to `grid_sample`. Mixing the two conventions applies a spurious sub-voxel scale and shift, so at `align_corners=False` a whole-voxel translation through `warp_affine3d` did not move the volume by a whole voxel, and an identity homography through `warp_perspective3d` or `homography_warp3d` did not reproduce its input. Those two default to `False`, so plain calls were affected.

This threads `align_corners` through exactly as #3945 did in 2D. `normal_transform_pixel3d`, `normalize_homography3d` and `create_meshgrid3d` gain `align_corners: bool = True`, last in the signature, so every existing call is bit-identical. The `False` branch is the half-pixel mapping, scale `2 / size` and offset `(1 - size) / size`, written as a single division so eager and graph capture round once and agree bitwise, and it needs no singleton guard because a size of 1 already lands on the centre. `warp_affine3d`, `warp_perspective3d` and `homography_warp3d` forward their own flag to the normalization and the grid they build.

Measured on `main` and on this head, as `max|out - in|` or `max|out - expected|`:

| call | `main`, `True` | `main`, `False` | this PR, `True` | this PR, `False` |
|---|---|---|---|---|
| `warp_affine3d`, +1 voxel along x, 2x3x4 | 4.8e-07 | 6.6667 | 4.8e-07 | 1.9e-06 |
| `warp_perspective3d`, identity, 4x4x4 | 0.0 | 55.125 | 0.0 | 0.0 |
| `homography_warp3d`, identity, 2x3x4 | 0.0 | 20.125 | 0.0 | 2.4e-07 |
| `RandomPerspective3D(distortion_scale=0.)`, `linspace(0, 1, 96).view(1, 1, 4, 4, 6)` | 4.8e-07 | 0.875 | 4.8e-07 | 9.5e-07 |

Every non-zero `this PR` cell is `grid_sample` accumulation noise rather than a residual of the fix. On this volume it is 4 ULP of the data at `True` and 8 ULP at `False`, and that ratio is unchanged when the ramp is rescaled 8x or 100x, so the exact figure depends on the torch build (these are torch 2.9.1; ducha-aiki measures 1 ULP on 2.14.0). The `before` figures are convention errors and are build-independent, which is why the fragment states the `after` cells as a bound.

The `align_corners=True` path did not change. I saved the outputs of all six functions on `main` and on this head for a random 4x4 homography and compared them bitwise: identical in every case.

### Behaviour change on default-argument paths

`warp_perspective3d`, `homography_warp3d`, `affine3d`, `rotate3d`, `crop_and_resize3d`, `crop_by_boxes3d`, `RandomAffine3D`, `RandomRotation3D` and `RandomPerspective3D` default to `align_corners=False`, so calls that pass no flag return different voxels (`rotate3d(x, 10., 10., 10.)` on a 2x3x4 `arange` volume changes by 2.80, `affine3d` with a small shear and translation by 2.73). Their transform matrices are unchanged. The outputs now match the `align_corners=True` result up to out-of-bounds handling, instead of carrying the sub-voxel shift. `RandomAffine3D` and `RandomRotation3D` are the one qualification: the old defect is a per-axis dilation about the volume centre, which commutes with a centred rotation or a uniform scale, so on a *cubic* volume their default draw is unchanged (3.1e-05 on 4x4x4, float noise) while `translate` (8.94) and `shears` (2.15) break the commutation. `warp_affine3d`, `center_crop3d` and `crop_by_transform_mat3d` default to `True` and are unchanged on their default path; they change only for an explicit `False`. These are all the call sites: the only other one, `kernels_geometry.py`, passes `True`, and `CenterCrop3D` and `RandomCrop3D` default to `True`. The `breaking` fragment carries the figures.

### Tests that moved, and why

Two hardcoded literals were generated under the defective default and are replaced with the values the `align_corners=True` path already produced, with the pre-fix literal and generation snippet kept in a comment: `TestCropAndResize3D::test_crop`, whose new values are the ones `test_crop_batch` pins for the same boxes at `True`, and `TestRandomRotation3D::test_random_rotation`, whose new values equal the `True` output for the same seed to within 5.4e-07 while its transform matrix is untouched.

Two convention pins asserted that the two settings *differ* on an in-bounds crop (`TestCenterCrop3D` and `TestCropByBoxes3D`). A crop box is given in voxel coordinates, so both conventions now resolve it to the same voxel centres, exactly as the 2D crop pins did in #3945. They now assert the two agree and both equal the integer slice, and pin the default value on the signature instead.

`TestCropAndResize3D::test_gradcheck` used integer box corners. Trilinear sampling is not differentiable where a sample lands exactly on a voxel centre, and with integer corners every sample does under either setting; it passed before only because the defective default sampled off-centre. Fractional corners fix it and it passes under both settings.

The strict `xfail` that #4514 left for this issue is folded into the identity test, now parametrized over both settings. New pins mirror the 2D ones: `create_meshgrid3d` under both conventions, `normal_transform_pixel3d` at `False` applied to voxel centres, `normalize_homography3d` at `False` on the identity including size 1 and against the explicit composition, a whole-voxel translation at `False` through both warps, and eager-versus-`torch.export` bitwise agreement at the five sizes where the two-step offset used to diverge, with all three spatial dims dynamic so the depth and height offsets are exercised too.

One float16 node is recorded in the known-failure manifest, `test_convention_identity_reproduces_input[cpu-float16-dsize2-False]` on `(1, 5, 3)`, off by 1.5e-3 on a zero voxel. Its `True` sibling `TestHomographyWarper3D::test_normalize_homography_identity` is already recorded for the same reason.

Prose: the conventions page no longer lists the 3D warps as an exception, the `normalize_homography` warning no longer says the 3D function lacks the parameter, and the `crop_and_resize3d` bullet claiming the default "interpolates instead" of the exact slice is corrected; its `align_corners` bullet says the flag changes nothing on an in-bounds box, since the function takes no `padding_mode`. `grep -rnE "#4503|issues/4503" kornia/ docs/` is empty; the remaining hits are test pins that name the issue.

Fixes #4503. Follows #3945 and #4514.

## What did you check?

```text
$ pytest tests/ --dtype=float32                                        10902 passed, 3737 skipped
$ pytest tests/ --dtype=float64                                        10906 passed, 3732 skipped
$ pytest tests/ --dtype=float16  --verify-known-failures --known-failure-profile=cpu-float16
                                                                        10296 passed, exit 0
$ pytest tests/ --dtype=bfloat16 --verify-known-failures --known-failure-profile=cpu-bfloat16
                                                                        10357 passed, exit 0
$ pytest --doctest-modules kornia/                                        657 passed, 14 skipped
$ KORNIA_TEST_OPTIMIZER=inductor pytest <3D files, tests/augmentation/_3d> -k "dynamo or compile or export or trace"
                                                                           69 passed
$ python docs/export_support/cases_geomA.py out.json transform.warp_perspective3d transform.homography_warp3d transform.warp_affine3d transform.warp_grid3d
                                                                            5 ok
```

`torch.compile(fullgraph=True)` and `torch.jit.trace` of `warp_perspective3d`, `warp_affine3d` and `create_meshgrid3d` have zero graph breaks under both conventions; bitwise at `True`, within 1.9e-06 at `False` for the warps (the grid itself is bitwise), which is `grid_sample` kernel noise on a 0..23 volume rather than the normalization. Pre-commit passes on all changed files; `pixi run typecheck` reports only the pre-existing diagnostic in `kornia/feature/lightglue.py`.

## How did AI help?

Claude Code did the investigation, the change, the tests and this description. I directed the work, reviewed every line, and every figure above comes from a run I executed on this branch against a `main` worktree.


ducha-aiki (changes requested):
Review of `8e84551e49d084b5e3a872304e3d4f423d8b7121`.

The fix is correct. It threads `align_corners` through `normal_transform_pixel3d`, `normalize_homography3d` and `create_meshgrid3d` exactly as #3945 did in 2D, with the same signatures and defaults. The `False` mapping is the half-voxel one:
- voxel centres `0` and `size - 1` land at `-1 + 1/size` and `1 - 1/size`;
- `-0.5` and `size - 0.5` land at `±1`;
- a size of 1 lands on the centre.

`create_meshgrid3d(..., align_corners=False)` equals the pixel grid mapped through `normal_transform_pixel3d(..., align_corners=False)`. It also equals `F.affine_grid(identity, align_corners=False)` after the `(d, x, y)` reorder, and its `x`/`y` channels equal the 2D `create_meshgrid`. The `True` path is untouched. The measurement table reproduces, the moved literals and flipped crop pins are justified, and every new pin fails on main. One P2 blocks, in the breaking fragment. The rest are P3s.

**[P2] The breaking fragment and the PR body omit `affine3d` and `rotate3d`, and both change on their default path.**

`kornia.geometry.transform.affine3d` (`kornia/geometry/transform/affwarp.py:196`) and `rotate3d` (`affwarp.py:328`) both default to `align_corners=False` and forward it to `warp_affine3d`. Plain calls therefore return different voxels now. On a `(1, 1, 2, 3, 4)` `arange` volume:

```
rotate3d(x, 10., 10., 10.)                          max|head - main| = 2.80
affine3d(x, [[1, .1, 0, .3], [0, 1, 0, .2], [0, 0, 1, .1]])   max|head - main| = 2.73
```

`changelog.d/4556.breaking.md:1-3` lists `warp_perspective3d`, `homography_warp3d`, `warp_affine3d`, `crop_and_resize3d`, `crop_by_boxes3d` and `center_crop3d`. Lines 18-21 add `RandomAffine3D` and `RandomRotation3D`, which reach the change *through* `warp_affine3d` and `affine3d`. The two public functions in between are missing, and so is `crop_by_transform_mat3d`, which changes only for an explicit `False`. The PR body's "Behaviour change on default-argument paths" section also omits `crop_and_resize3d` and `crop_by_boxes3d`, although the fragment has them. These are the only other call sites: `grep` over `kornia/` for callers of the six functions finds exactly `affwarp.py:244`/`398`, `crop3d.py:384`/`429`, the five 3D augmentations and `kernels_geometry.py:208`. The last one passes `align_corners=True`, so motion kernels are unchanged. `CenterCrop3D` and `RandomCrop3D` default to `True` and are also unchanged.

Suggested fix: add `affine3d` and `rotate3d` to the default-path sentence, and `crop_by_transform_mat3d` to the explicit-`False` list. Optionally add a `rotate3d` row to the table.

**[P3] `changelog.d/4556.fixed.md:4` says the translation "was off by 6.67 voxels".** The 6.67 is `max|out - expected|` in intensity units on the `arange` volume, not a displacement in voxels. Please say "differed from the exact shift by up to 6.67 on a 2x3x4 `arange` volume", as the breaking fragment does. In the same vein, `breaking.md:18-19` says "up to 1.5 and 1.7 on a 0..23 volume at 10 degrees". Over seeds 0-19, `RandomRotation3D(10.)` and `RandomAffine3D(10.)` on that volume change by up to 2.99 and 3.43. Either name the seed or drop "up to".

**[P3] `test_normal_transform3d_capture_matches_eager_bitwise` exercises only the `x` offset.** `tests/geometry/test_conversions.py:3669` makes only dim 4 (width) dynamic, so depth 2 and height 4 are static powers of two, where every offset is exact. I checked what the pin catches by changing the capture-branch offsets back to two steps:
- the two-step `tx_t = one / width_t - one` fails the 5 `False` rows (good);
- `ty_t = one / height_t - one` and `tz_t = one / depth_t - one` pass all 10 rows.

With all three dims dynamic, that `ty`/`tz` plant diverges at sizes 3, 11, 31 and 251. The production code is right: with all three axes dynamic and no plant, `normal_transform_pixel3d` and `create_meshgrid3d` agree bitwise with eager at `(3, 11, 31)`, `(251, 31, 3)`, `(11, 4051, 1)`, `(1, 1, 1)` and `(7, 5, 9)` under both settings. Suggested change: make dims 2, 3 and 4 dynamic, and use the size for all three at runtime, for example `torch.zeros(1, 1, size, size, size)` capped to keep 4051 cheap, or a mixed shape.

**[P3] `docs/source/get-started/conventions.rst:200-205` puts "The 3-D warps are consistent." inside the `.. warning::` block that opens at line 191.** It is no longer a warning, so it renders in a warning box next to the `remap` caveat. Move it out of the admonition, for example after the "Where the two settings genuinely differ is out-of-bounds sampling" paragraph. The `align_corners` defaults table above still lists 2D functions only. A 3D row would help, because the 3D defaults are split: `warp_affine3d`, `center_crop3d` and `crop_by_transform_mat3d` default to `True`; `warp_perspective3d`, `homography_warp3d`, `affine3d`, `rotate3d`, `crop_and_resize3d` and `crop_by_boxes3d` default to `False`.

**Checked and fine**

- **Bit-identity of `True`.** I saved 214 outputs on main and on this head and compared them with `torch.equal`:
  - `normal_transform_pixel3d` (default and explicit `True`), `create_meshgrid3d` (normalized and pixel), all four float dtypes;
  - `normalize_homography3d`, `warp_affine3d`, `warp_perspective3d`, `homography_warp3d` (normalized and pixel coordinates), `crop_by_transform_mat3d`, `rotate3d(align_corners=True)` and `center_crop3d`, batch 3 with random near-identity projective homographies, float32/float64;
  - sizes `(4,4,4)`, `(2,3,5)`, `(1,5,3)`, `(3,1,4)`, `(7,2,1)`, `(1,1,1)`, `(5,9,6)`, with source and destination sizes differing.

  0 mismatches.
- **Eager against the capture branch at `False`.** I forced `is_compiling()` to take the tensor branch eagerly, for widths 1-2999 plus 4051, 8191, 16383, 16385, 65535 and 100003. The two branches agree bitwise in float32, float64, float16 and bfloat16, and each entry equals the 2D `normal_transform_pixel(..., align_corners=False)` one. Compiled with `fullgraph=True, dynamic=True`, `warp_perspective3d` at `False` has an identity residual of 1.9e-06, matching the PR body. The compiled `create_meshgrid3d` is bitwise equal to eager.
- **Measurement table on main against head** (`max|out - in|` or `max|out - expected|`):
  - `warp_affine3d` +1 voxel on `(2, 3, 4)` at `False`: 6.6667 → 0.0;
  - identity `warp_perspective3d` on `(4, 4, 4)` at `False`: 55.125 → 0.0;
  - identity `homography_warp3d` on `(2, 3, 4)` at `False`: 20.125 → 2.4e-07;
  - default `warp_perspective3d` +1 voxel on `(2, 3, 4)`: 17.5 → 2.4e-07;
  - `RandomPerspective3D(distortion_scale=0.)` on a 4x4x6 linspace ramp: 0.875 → 1.2e-07;
  - `True` rows unchanged.
- **Moved literals.**
  - `TestCropAndResize3D::test_crop`: the new values `[16, 19, 24, 27, 48, 51, 56, 59]` are the integer slice `16z + 4y + x` at `z ∈ {1, 3}`, `y ∈ {0, 2}`, `x ∈ {0, 3}`, and equal main's `align_corners=True` output.
  - `TestRandomRotation3D::test_random_rotation`: the new literal is within 5.5e-05 of the head output (its 4-decimal rounding). At seed 0, head's default output is within 5.1e-07 of `align_corners=True`.
  - Out-of-bounds claim: with random homographies, `True` and `False` outputs agree to 1e-06 under `zeros` and `border` padding and differ under `reflection` (0.57 / 0.65). So "only out-of-bounds handling" is accurate.
- **Gradcheck.** With integer box corners, `gradcheck` on `crop_and_resize3d` fails under both settings, for an `arange` and a random volume at both output sizes. With `+0.25` it passes in all 8 combinations. An interior integer box passes on `arange` but fails on a random volume, so the comment's reason is the right one; `arange` hides it only in the interior because it is linear.
- **Plant-and-revert.** With main's `conversions.py`, `grid.py`, `imgwarp.py` and `crop3d.py` restored under this head's tests, 52 nodes fail at float32:
  - every new pin (for the parametrized whole-voxel and identity pins, every `False` row);
  - both moved literals;
  - both flipped crop pins.

  The worktree was restored afterwards.
- **Issue references.** In the merge tree of `origin/main` and this head, `git grep -nE "#4503|issues/4503"` finds nothing in `kornia/` or `docs/`. The `tests/` hits are the new pins and comments. The strict xfail from #4514 is folded in correctly: its two `False` rows now pass as plain parametrizations.
- **Fragments.** One PR may add both `breaking` and `fixed` (changelog.d README), and the format matches #3945's. No new public names, so no `.rst` or `api_surface.json` change is needed. `ruff@0.16.7 format --diff` and `check` are clean on the 15 changed Python files.
- **Pre-existing, not from this PR:** `torch.jit.script(warp_affine3d)` and `torch.jit.script(warp_perspective3d)` fail on main too. The three changed helpers script on head, and a scripted `normal_transform_pixel3d(..., False)` equals eager.

**Policy (separate from the technical verdict)**

This changes default outputs of `RandomAffine3D`, `RandomRotation3D` and `RandomPerspective3D`. Those are batch-6d symbols under the #4407 repair-window hold, whose rule is "change no runtime behavior". #4412 (the split `align_corners` default) is on that batch's list. #3945 merged the 2D half as an explicit maintainer exception, on the grounds that #3904 is not on the batch-6 list. #4503 is not on it either, but this PR needs the same maintainer call before it merges.

**What I ran** (repo `.venv`, python 3.11, torch 2.14.0; `kornia.__file__` printed as the PR or main worktree in every probe):

- The six touched test files, `tests/augmentation/_3d` and `test_affine.py`, CPU float32/float64: 1339 passed, 22 skipped, 18 xfailed.
- Every test file that references a 3D warp, crop, meshgrid or 3D augmentation (20 files), float32/float64: head 3527 passed, 47 xfailed, 7 xpassed (non-strict); main 3439 passed, 51 xfailed, 7 xpassed. The 4 fewer xfails are the removed strict xfail's rows.
- The same 20 files with `--xfail-known-failures`:
  - profile `cpu-float16`: head 1828 passed, 123 xfailed, 2 failed; main 1780 passed, 124 xfailed, 2 failed;
  - profile `cpu-bfloat16`: head 1857 passed, 121 xfailed, 3 failed; main 1808 passed, 123 xfailed, 3 failed.

  The failures are the same nodes on both trees (`test_deg2rad`; bfloat16 also `TestHomographyWarper::test_translation`), macOS arm64 against the Linux manifest. There is no new `XPASS(strict)`. The new manifest node `test_convention_identity_reproduces_input[cpu-float16-dsize2-False]` fails unmasked by 1.46e-3 at a zero voxel, and it is the only half node of the new pins that fails.
- `KORNIA_TEST_OPTIMIZER=inductor`, touched files plus `tests/augmentation/_3d`, `-k "dynamo or compile or export or trace or jit"`: 81 passed, 3 skipped, 1 xfailed on both trees.
- Torch 2.5.1 overlay (prints `2.5.1`), new and changed pins, float32/float64: 288 passed, 1 xfailed.
- MPS float32: `test_imgwarp3d.py`, `test_crop3d.py`, `test_grid.py` and `test_random_rotation_3d.py` gave 133 passed, 14 skipped, 1 xfailed; `test_homography_warper.py -k 3d` gave 74 passed.
- Doctests: `--doctest-modules` on `kornia/augmentation/_3d`, `imgwarp.py`, `crop3d.py`, `affwarp.py`, `grid.py` and `conversions.py` gave 73 passed.
- Merge: `git merge-tree --write-tree origin/main <head>` exits 0.
- Simulated collisions with open PRs that touch the same files, on main + this head:
  - #4538, #4540, #4471 and #3981 stay clean. #4538 has no 3D content and leaves no `#4503` hit.
  - #3979, #3993 and #3996 conflict in the same files as against main alone.
  - #3944 gains one extra adjacent-line conflict in `testing/half_precision_xfails/cpu_float16.txt`, since both touch the `test_imgwarp3d` / `test_thin_plate_spline` boundary.
- CI at 10:35 UTC: 11 check runs succeeded, plus the pre-commit.ci and RTD statuses. 28 were still queued or in progress, and none had failed.

ducha-aiki:
Review of `d57037f9`. **All five asks from the last round are addressed**, and I re-verified each independently rather than reading the delta. Technically this is ready; the #4407 repair-window exception is the remaining gate and is not mine to decide — see the bottom.

**The [P2] fragment list is now complete.** I enumerated the surface myself, comparing main against this head at *default* arguments. Changing: `warp_perspective3d` 55.125, `homography_warp3d` 20.125, `affine3d` 2.72711, `rotate3d` 2.79631, `crop_and_resize3d` 9.16667, `crop_by_boxes3d` 6.14197, and `RandomAffine3D` / `RandomRotation3D` / `RandomPerspective3D`. Bit-identical: `warp_affine3d`, `center_crop3d`, `crop_by_transform_mat3d`, `normal_transform_pixel3d`, `normalize_homography3d`, `create_meshgrid3d`, `normalize_pixel_coordinates3d`, `denormalize_pixel_coordinates3d`, `warp_grid3d`, and `CenterCrop3D`, `RandomCrop3D`, the three flips, `RandomEqualize3D` and `RandomMotionBlur3D` over 20 seeds each. `crop_by_transform_mat3d` at an explicit `False` moves by 11.6667. That is exactly the two lists the fragment now carries. The other two `create_meshgrid3d` call sites (`contrib/distance_transform.py:102`, `geometry/subpix/spatial_soft_argmax.py:511`) both pass `normalized_coordinates=False`, so the flag is inert there, and neither passes a seventh positional argument.

The three P3 figures reproduce: `warp_affine3d`'s +1-voxel shift is off by 6.66667 at `False` on main and 0.0 here; `RandomRotation3D(10.)` on the 0..23 volume differs by 1.72928 at seed 0 and 2.987 over seeds 0-19; `RandomAffine3D(10.)` by 1.54936 and 3.4272. The capture pin now bites: planting the two-step form on `ty_t`/`tz_t` only, leaving `tx_t` correct, gives **5 failed / 5 passed** — every `False` row, greatest absolute difference 5.96e-08 at index `(0, 2, 3)`. Last round that same plant passed all ten. And the 3-D paragraph is out of the `.. warning::` in the merge tree, with the two 3D rows added to the defaults table.

Also re-checked: **52 nodes fail** with the merge base's four production files restored under these tests, including both moved literals *and* both softened convention pins (`TestCenterCrop3D::test_convention_align_corners_default_is_true`, `TestCropByBoxes3D::test_convention_align_corners_default_is_false`), so the flipped asserts were not weakened into decoration. The manifest addition fails unmasked by 0.00146484375 against a 0.001 tolerance and xfails with the profile; the two xfails that disappear came from the in-source strict mark, not the manifest, so no deletion is owed, and there is **0 XPASS on either tree** at float16 or bfloat16 with identical FAILED sets. 98 `align_corners=True` outputs are bitwise equal to main; eager and the forced capture branch are bitwise equal at 10 sizes up to 100003 in all four dtypes; `create_meshgrid3d(..., False)` matches `F.affine_grid(identity, False)` to one float32 ULP; `torch.jit.script` works on all three helpers at `False`. `#4503` has zero hits in `kornia/` and `docs/` in the merge tree of current main (after #4566 and #4538), and main's stale 3-D-exception warning on the conventions page is correctly removed by the merge. CI is 40 check runs plus `pre-commit.ci` and RTD, all green, and the merge is clean.

Two P3s, both prose.

### [P3] The `RandomPerspective3D` row in the breaking fragment does not reproduce

`changelog.d/4556.breaking.md:20`, and the same pair in the PR body's table. It says `0.866 -> 9.5e-07` on a "4x4x6 ramp". Every [0,1]-normalized 4x4x6 ramp I tried gives **0.875 -> 1.19e-07**: `linspace(0, 1, 96)`, `arange(96)/95`, and a `linspace(0, 1, n)` ramp along D, H or W alone, at float32 and float64, with shapes `(1,1,4,4,6)`, `(1,3,4,4,6)` and `(2,1,4,4,6)`, identically for every seed 0-19 (`distortion_scale=0.` makes the draw irrelevant). An unnormalized `arange(96)` ramp gives `83.125 -> 7.6e-06` instead. I measured 0.875 -> 1.2e-07 last round too and did not catch that it disagreed with the table — my miss as much as yours. Please either name the exact construction or use the measured pair.

### [P3] `kornia/geometry/transform/crop3d.py:46` claims a difference I could not produce

"align_corners: `False` by default; it only changes out-of-bounds handling". Over a 20-box sweep at float64 on a 4x4x4 volume — x shifted by -1.5 through +1.5 and scaled so the box runs past both edges — `max|True - False|` for `crop_and_resize3d` never exceeded 1.03e-05, which is the float32 rounding of the normalize-and-invert chain (the `True` path returns 16.000000476 for the integer box). The function takes no `padding_mode`, so the reflection-padding case the 2D sentence rests on cannot arise here. "It changes nothing on an in-bounds box, since the box is given in voxel coordinates" is what the code supports.

### Policy — unchanged from last round, and still the gate

This changes default-path outputs of three batch-6d classes — `RandomAffine3D` (0.0506), `RandomRotation3D` (0.0513) and `RandomPerspective3D` (0.7063), measured over 20 seeds on a normalized 4x4x6 ramp — while `CenterCrop3D`, `RandomCrop3D`, the three flips, `RandomEqualize3D` and `RandomMotionBlur3D` are bit-identical. On the function side it changes the default path of `warp_perspective3d`, `homography_warp3d`, `affine3d`, `rotate3d`, `crop_and_resize3d` and `crop_by_boxes3d`, and `crop_by_transform_mat3d` only at an explicit `False`.

Batch 6d's rule is "change no runtime behavior" and #4412 (the split `align_corners` default) is on its list. #3945 merged the 2D half as an explicit exception because #3904 is not on the batch-6 list; #4503 is not on it either. **The same maintainer call is needed before this merges** — I am recording the blast radius precisely so it can be made, not making it.

### What I ran

Repo `.venv`, python 3.11, torch 2.14.0; `kornia.__file__` printed in every probe. `tests/augmentation/_3d` plus the five touched geometry files, CPU float32 and float64: 795 passed, 11 skipped, 13 xfailed each. `--doctest-modules` on `grid.py`, `conversions.py`, `imgwarp.py`, `crop3d.py`: 48 passed, including the new `create_meshgrid3d` example. `tests/test_api_surface.py`: 147 passed (no new public names). Half precision on the six touched files: head 46 xfailed at float16 and 48 at bfloat16, against 47 and 50 on main. Moved literals: the `RandomRotation3D` literal is head's output to 5.47e-05 against a 1e-4 atol, head's default equals head's `True` to 5.07e-07, the old literal was main's output to 5.07e-05, new-vs-old is 0.2043, and the transform matrix is bit-identical to main; `TestCropAndResize3D::test_crop` returns exactly `[16, 19, 24, 27, 48, 51, 56, 59]` at the default, which main produced only at `True`. `ruff@0.16.7` clean on all ten changed Python files. Not re-verified this round (unchanged code, green in CI): the torch 2.5.1 overlay, MPS, and the inductor leg.


ducha-aiki:
## Re-review — #4556, head `f572a4a6`

The docs-only delta does what its commit message says, and I re-verified the blast-radius list that the #4407 decision rests on. **This PR remains technically ready and blocked only on the maintainer's repair-window call, not on any code finding.** One stale number is worth one more push.

Merge onto `ecbcad561` is clean. CI **42/42 SUCCESS**, `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` on approval only.

**The delta contains no executable change** — proven rather than read: stripping docstrings from both revisions of `crop3d.py` gives an **identical `ast.dump`**; without stripping it differs. `git diff --stat d57037f9..f572a4a6` is 2 files, +3/−2.

### Both claims in the new commit, checked by execution

**"Names the RandomPerspective3D ramp" — the *before* value is true.** `linspace(0, 1, 96).view(1, 1, 4, 4, 6)` is a valid 4×4×6 ramp (96 = 4·4·6) and gives exactly **0.875** on main, for every seed 0–5 at `p=0.5` and `p=1.0` (`distortion_scale=0.` makes the draw irrelevant). Naming the input was the right fix — it is what made the row falsifiable.

**"Drops the crop3d out-of-bounds claim" — the dropped statement was false, so dropping it loses nothing.** This was the one I was watching for, since dropping a *true* warning is as wrong as keeping a false one. The removed text said `align_corners` "only changes out-of-bounds handling" for `crop_and_resize3d`. On the merge tree I swept 240 in-bounds boxes (random fractional corners, shapes (4,4,4)/(3,5,7)/(6,4,5), `bilinear` and `nearest`, float32+float64): worst `max|True − False|` = **7.7e-07**. Then 240 deliberately **out-of-bounds** boxes: worst = **1.0e-06** — also pure float noise.

The mechanism: the function exposes no `padding_mode` (it reaches `crop_by_transform_mat3d`'s `"zeros"`), and `grid_sample`'s zero/border padding is applied in *pixel* space, so once the normalization follows the flag the two conventions agree out of bounds too. Only `reflection` padding would differ, and this API cannot select it. Your new bullet is true, and the PR body's stated reason ("takes no `padding_mode`") is the right one — if anything it under-claims, since it would hold without the "in-bounds" qualifier.

---

## [P3] The `RandomPerspective3D` row's *after* cell is wrong by 8×

`changelog.d/4556.breaking.md:20` reads `linspace(0, 1, 96).view(1, 1, 4, 4, 6)` | 0.875 | **9.5e-07**. On that exact construction the head gives **1.1920929e-07**, in float32 and float64 alike. The figure `9.5367432e-07` reproduces on `linspace(0, 8, 96)` — i.e. the "after" cell still comes from a differently scaled ramp.

The PR body's four-column version of the row is worse: its `main, True` and `this PR, True` cells say `4.8e-07` where the named ramp gives **9.3132257e-10**, so 3 of its 4 cells do not reproduce on the input it names.

Set the fragment cell to `1.2e-07`, and the PR-body row to `9.3e-10 | 0.875 | 9.3e-10 | 1.2e-07`.

## [P3] The fragment over-states the augmentation blast radius on cubic volumes — in your favour

`breaking.md:22-23` says `RandomAffine3D`/`RandomRotation3D` "sample different voxels for non-identity transforms". On a **cubic** volume with their default (rotation-only, centred) parameters they do not:

| shape | `max|base − head|`, seeds 0/1/7, both dtypes |
| --- | --- |
| `(4,4,4)` | **3.6e-07 / 4.2e-07** — i.e. unchanged |
| `(2,3,4)` | 0.135 / 0.096 |
| `(4,4,6)` | 0.036 / 0.037 |

The mechanism, verified: the old defect is a per-axis dilation about the volume centre, which commutes with a centred rotation and a centred uniform scale, so on a cube they cancel (`RandomAffine3D` rotation-only 2.4e-07, `+scale` **bitwise 0**) while `translate` (0.191) and `shears` (0.032) break the commutation. `RandomPerspective3D` changes on every shape (0.68 / 0.70 / 0.73).

This narrows the practical blast radius, so it is worth saying: "…sample different voxels except where the sampled transform commutes with the old per-axis dilation (a centred rotation or uniform scale on a cubic volume)".

---

## The blast-radius list re-confirms on the current head

Default arguments, pristine `ecbcad561` vs the merge tree, `torch.equal`, float32 and float64 agreeing on every verdict (144 + 198 + 116 saved tensors compared):

| symbol | verdict | `max|base−head|` |
| --- | --- | --- |
| `warp_perspective3d` | CHANGED | 15.07 |
| `homography_warp3d` | CHANGED | 16.92 |
| `affine3d` | CHANGED | 2.727 |
| `rotate3d` | CHANGED | 2.796 |
| `crop_and_resize3d` | CHANGED | 10.5 |
| `crop_by_boxes3d` | CHANGED | 10.5 |
| `warp_affine3d` | BIT-IDENTICAL | 0 |
| `center_crop3d` | BIT-IDENTICAL | 0 |
| `crop_by_transform_mat3d` | BIT-IDENTICAL | 0 |
| `RandomAffine3D` | CHANGED | 0.135 (0 on a cube) |
| `RandomRotation3D` | CHANGED | 0.096 (0 on a cube) |
| `RandomPerspective3D` | CHANGED | 0.735 |
| `CenterCrop3D`, `RandomCrop3D`, `RandomHorizontalFlip3D`, `RandomVerticalFlip3D`, `RandomDepthicalFlip3D`, `RandomEqualize3D`, `RandomMotionBlur3D`, `RandomTransplantation3D` | BIT-IDENTICAL | 0 (18 cases each) |

**One correction to my own earlier note: eight other 3D classes are bit-identical, not seven** — `RandomTransplantation3D` is the eighth; my previous round's harness errored on it and I re-probed it correctly here.

**Explicit-flag matrix**, all nine functions, both dtypes: bit-identical at explicit `True` for all nine; changed at explicit `False` for all nine (`warp_affine3d` 2.73, `center_crop3d` 3.5, `crop_by_transform_mat3d` 3.5). Exactly what `breaking.md:1-5` and `:25-26` assert. Every `transform_matrix` of every 3D augmentation is bit-identical, as the fragment claims.

**Merge safety against the 12 main commits this PR has not merged.** Its base is `1590df487`, so its own green CI predates batches 6b/6c. `git diff --name-only 1590df487 ecbcad561` touches **none** of this PR's production files, and the full `tests/augmentation` suite — including main's new `test_conventions_geometric_*.py` — gives byte-identical pass/skip/xfail/xpass counts on base and on the merge tree (**3332 / 3405 / 30 / 5 on both**). No convention pin from 6b/6c collides.

`tests/geometry` full: head 3100 passed / 10 xfailed vs base 3051 / 12 — the +49/−2 are this PR's new pins and the folded-in strict xfail; 6 xpassed on both, no new XPASS. Six touched test files at float32+float64: 1242 passed, 12 skipped, 18 xfailed. Doctests on the 5 changed modules: 58 passed. `grep -rnE "#4503|issues/4503" kornia/ docs/` → empty; the `tests/` hits are the pins, each naming the issue. Pinned `ruff@0.16.7` clean.

The fragment's remaining figures reproduce: `RandomRotation3D(10.)` on the 0..23 volume gives **1.7293** at seed 0 and **2.987** over seeds 0–19; `RandomAffine3D(10.)` gives **1.5494** and **3.4272** — all four match to the printed digits.

**Mutation testing deliberately not re-run**, with a proof it cannot apply: the delta is AST-equal after docstring stripping, so no pin could discriminate it. Last round's evidence stands on this head — 52 nodes fail when the merge base's four production files are restored under these tests, including both moved literals and both flipped convention pins.

---

**Still not approving**, and to be clear about why: nothing here is a code finding. This changes default-path output for `RandomAffine3D`/`RandomRotation3D`/`RandomPerspective3D` plus `warp_perspective3d`, `homography_warp3d`, `affine3d`, `rotate3d`, `crop_and_resize3d`, `crop_by_boxes3d`, and it needs the same #4407 repair-window exception #3945 got for 2D. That decision is @ducha-aiki's, and the list above is what it rests on.

VanniLeonardo:
Thanks. I merged main (the branch had gone stale, one conflict in `testing/half_precision_xfails/cpu_float16.txt` where #4565 removed the TPS line next to ours, resolved by keeping main's removal and our line) and fixed both P3s.

P3 on the RandomPerspective3D row. Our numbers disagree and I think the row was asking for a figure that cannot be pinned. On torch 2.9.1 the named ramp gives 9.5367432e-07 at float32 and 1.1920929e-07 at float64, so your 1.19e-07 is what I measure at float64, not float32. The reason the two of us differ is that this residual is not a residual of the fix, it is grid_sample accumulation noise. On my build it is exactly 4 ULP of the data at `True` and 8 ULP at `False`, and that ratio is unchanged when the ramp is rescaled by 8 or by 100:

```
scale   align_corners=True            align_corners=False
1       4.7683716e-07  (4 ULP)        9.5367432e-07  (8 ULP)
8       3.8146973e-06  (4 ULP)        7.6293945e-06  (8 ULP)
100     4.5776367e-05  (3.8 ULP)      9.1552734e-05  (7.7 ULP)
```

Your 9.3132257e-10 is 2^-30, which is one float32 ULP at about 0.01, i.e. the spacing of this ramp. So both of your figures look like one ULP of something on torch 2.14.0 against four and eight on 2.9.1. Pinning either would put a build-dependent number in a changelog. The fragment now marks those cells *rounding* with a bound of 2e-06 and one sentence saying they are accumulation noise that scales with the input. The `before` figures are convention errors, are build-independent, and are unchanged. The PR body keeps the four-column table with the measured torch 2.9.1 values and carries the same explanation, naming your 2.14.0 measurement.

Same treatment for the `+1 voxel` row, which said "exact" and is 1.9e-06 here. That was the same over-claim in the other direction.

P3 on the cubic volumes. You are right and it reproduces. Rotation-only `RandomAffine3D` and `RandomRotation3D` on `(4, 4, 4)` are unchanged (3.1e-05 and 2.7e-05 over seeds 0, 1, 7 at float32 and float64, which is float noise on an `arange(64)` volume), and so is `RandomAffine3D` with `scale` (3.8e-06), while `translate` gives 8.94 and `shears` 2.15 on the same cube. On `(2, 3, 4)` and `(4, 4, 6)` everything changes. The fragment now says they sample different voxels "except where the sampled transform commutes with the old per-axis dilation, as a centred rotation or a uniform scale does on a cubic volume", and gives the cube figures.

Re-verified on the merge tree, since it now carries 12 commits of main that its green CI predated: six touched test files at float32 and float64 1246 passed, full `tests/augmentation` 3444 passed with main's new 6b/6c convention pins, `tests/geometry` 3113 passed, both half-precision known-failure legs exit 0 (11090 and 11141 passed), doctests on the five changed modules 58 passed, `tests/test_half_precision_ci.py` and `tests/test_api_surface.py` 211 passed. Every figure in the table reproduces on the merged tree against `c643312a`.

On the repair-window call, understood, and thank you for recording the blast radius that precisely. Nothing further from me until that decision.

ducha-aiki (dismissed):
Re-reviewed the latest delta. The inaccurate noise figures are now stated as bounds and the cubic-volume qualification is explicit. Current CI is green and I found no unresolved technical findings. Approval is for the code; merging still requires the explicit maintainer decision for the #4407 repair-window exception.

ducha-aiki:
Good work, and let's wait for the repair window (week or so)

ducha-aiki:
Merged `origin/main` (088586663) into this branch and reconciled main's #4503 pins. New head: b7fe7c6d2.

**Conflict in `docs/source/get-started/conventions.rst`:** #4634 had refined the "3-D warps are the other exception" warning paragraph, the one this PR deletes as fixed. I kept the deletion and your new "The 3-D warps are consistent" paragraph. I carried one fact from main's refinement into your paragraph, because it still holds after the fix: `homography_warp3d` builds its sampling grid in `float32` (`create_meshgrid3d` without a dtype), so a `float64` volume is reproduced only to about `1e-6`. An identity `warp_perspective3d` on a float64 4x4x4 `arange` still leaves 1.25e-6 at `align_corners=True`. Main's new serialization sections come through unchanged.

**Reconciliation (b7fe7c6d2):** #4634 landed after this branch's base. It documented #4503 as an open defect, and the merged tree went red on it:
- `tests/augmentation/test_conventions_3d.py`: `test_wart_random_perspective3d_identity_align_corners_false_4503` asserted a deviation > 1 (now 1.9e-6), and the strict-xfail `test_convention_random_perspective3d_identity_align_corners_false` XPASSed. I inverted the wart into `test_convention_random_perspective3d_identity_under_either_align_corners_4503`, which checks identity under both settings, and dropped the xfail duplicate. The new pin fails on origin/main (float32 and float64) and passes here.
- `GeometricAugmentationBase3D` and `RandomPerspective3D` docstrings: removed the "at `align_corners=False` the applied warp differs ... tracked in #4503" wording. They now describe both settings as following the matrix. The float32-grid precision caveat stays: at `/64` scale, float64 is off by 4e-8 at False and 5e-9 at True. `grep -rnE "#4503|issues/4503" kornia/ docs/` is now empty. The `tests/` hits are your own pins plus the one above.

**Ran (torch 2.14, CPU):**
- This PR's test files + `tests/augmentation/_3d` + `test_conventions_3d.py` / `test_conventions_geometric_*.py`, float32/float64: 1488 passed, 12 skipped, 16 xfailed.
- Same set in float16 and bfloat16 with `--xfail-known-failures --known-failure-profile=cpu-<dtype>`: the only failures are `test_deg2rad` and bfloat16 `TestHomographyWarper::test_translation`. They fail the same way on an origin/main worktree, so it's macOS vs. the Linux manifest, not this PR.
- Doctests on the touched modules: 49 passed. Ruff check/format (pinned version): clean.
- Both-way check: `origin/main..HEAD` equals this PR's diff plus the two carried rst lines and the reconciliation commit.

ducha-aiki:
Coordination note from the `geometry.transform` convention-scope trim (#4788). There is no change request here.

I checked which #4503 prose and pins in `kornia/geometry/transform/` and `tests/geometry/transform/` this PR has to cover when it lands. It covers all of them:

- `crop_and_resize3d` block (`crop3d.py`: "reproducing the exact integer-voxel slice requires `align_corners=True`"): rewritten here.
- `test_crop3d.py` `TestCenterCrop3D::test_convention_align_corners_default_is_true` and `TestCropByBoxes3D::test_convention_align_corners_default_is_false`, which assert the defect: rewritten here as a signature default plus agreement.
- `warp_affine3d`, `warp_perspective3d`, `homography_warp3d` blocks: rewritten here.
- `affine3d`, `rotate3d`, `crop_by_boxes3d` blocks: these only state the `False` default, which stays true after the fix, so no edit is needed.
- `test_imgwarp3d.py` strict xfail: folded here.

#4788 deliberately leaves all of these hunks to this PR. `git merge-tree` of the two heads is clean, and `tests/geometry/transform` passes on the merged tree (float32/float64: 1243 passed).

One optional nit, for consistency with the scope rule #4788 applies: the "selects the normalization ... so ... under either setting" sentence now appears in three 3-D blocks, and the conventions page already states it. The "Pre-#4503 literal" comments in `test_crop3d.py` are history. Either can be trimmed later, so neither blocks this PR.

Posted on behalf of @ducha-aiki by Claude (Opus 5.5).

ducha-aiki:
Heads-up for the `RandomPerspective3D` docstring hunk: the float64-through-a-float32-grid behaviour described in the first Convention bullet is now tracked as a defect in #4776, and #4791 adds a separate one-line wart bullet for it (left outside your hunk to avoid a conflict). When you next rebase, that bullet's "the sampling grid is built by `create_meshgrid3d` in float32 ... exact to `float64` roundoff" clause can shrink to a pointer to #4776, so the fact is stated once.

Posted on behalf of @ducha-aiki by Claude (Opus 5.5).

ducha-aiki:
Addendum to the #4776 heads-up above. The same applies to the `conventions.rst` hunk. The new sentence "The grid is built in ``float32``, so a ``float64`` volume is reproduced only to about ``1e-6``." describes the #4776 defect but carries no issue link, so the AGENTS.md closing grep (`#4776|issues/4776`) would miss it when #4776 is fixed. Please end it with `` (`#4776 <https://github.com/kornia/kornia/issues/4776>`_)`` when you next rebase. It doesn't block this PR. #4788 already carries the #4776 wart on the `warp_perspective`/`warp_perspective3d` docstrings, and `git merge-tree` with it is clean.

Posted on behalf of @ducha-aiki by Claude (Opus 5.5).

ducha-aiki:
Coordination note for the #4776 fix in #4795 (approved at `b5096dca4`). No change is requested here yet.

#4556 merges cleanly with `main`, but it conflicts with #4795 in four places, all mechanical. Whichever PR lands second resolves them as follows:

- **`docs/source/get-started/conventions.rst`:** this PR rewrites the 3-D paragraph but keeps "is built in ``float32``, so a ``float64`` volume is reproduced only to about ``1e-6``". Once #4795 lands that sentence is false, so drop it.
- **`RandomPerspective3D`, first Convention bullet:** it should say the identity "reproduces its input under either `align_corners` setting up to floating-point roundoff".
- **`homography_warp3d`:** keep both `dtype=` (#4795) and `align_corners=` (this PR) in the `create_meshgrid3d` call; it accepts both.
- **`tests/geometry/transform/test_imgwarp3d.py`:** keep both new tests.

Posted on behalf of @ducha-aiki by Claude (Opus 5.5).

Issue #4504, opened 2026-09-13
remap ignores align_corners in its pixel-coordinate normalization, so an identity map does not reproduce the input


VanniLeonardo:
### What happened?

`remap` normalizes its pixel maps with `normalize_pixel_coordinates`, which hardcodes the corner-aligned convention where pixel centres `0` and `size - 1` map to `-1` and `+1`. It then samples with `grid_sample(align_corners=align_corners)`, whose own default resolves to `False`, where `-1` and `+1` are the outer pixel edges instead.

Mixing the two conventions applies a spurious sub-pixel scale and shift, so an identity pixel map does not reproduce the input. On a 4x4 `arange` image the deviation is `11.25` out of a 0 to 15 range, and that is the behaviour you get with no `align_corners` argument at all.

This is the same root cause as #3904 and #3928. #3945 fixes it for `create_meshgrid`, `normal_transform_pixel`, `normalize_homography`, `denormalize_homography`, `warp_perspective`, `warp_affine`, `homography_warp`, `warp_image_tps` and `HomographyWarper`. `remap` is left out of that PR because fixing it means adding a public `align_corners` argument to `normalize_pixel_coordinates` and `denormalize_pixel_coordinates`, which is wider than the linked issues. It is currently documented as the remaining exception on the Conventions page.

### Minimal reproduction

```python
import inspect, torch
from kornia.geometry import create_meshgrid
from kornia.geometry.conversions import normalize_pixel_coordinates
from kornia.geometry.transform import remap

print("normalize_pixel_coordinates:", inspect.signature(normalize_pixel_coordinates))

img = torch.arange(16.0).view(1, 1, 4, 4)
pix = create_meshgrid(4, 4, normalized_coordinates=False)   # identity pixel map
map_x, map_y = pix[..., 0], pix[..., 1]
for ac in (True, False, None):
    out = remap(img, map_x, map_y, align_corners=ac)
    print(f"remap identity map, align_corners={str(ac):5s} max|out - in| = {(out - img).abs().max():.4f}")
```

Observed on current `main`:

```
normalize_pixel_coordinates: (pixel_coordinates, height, width, eps=1e-08)

remap identity map, align_corners=True  max|out - in| = 0.0000
remap identity map, align_corners=False max|out - in| = 11.2500
remap identity map, align_corners=None  max|out - in| = 11.2500     <- the default
```

Expected: an identity pixel map reproduces the input under every setting, and the two conventions agree except in out-of-bounds handling.

### Environment

[...]

PR #4558, opened 2026-09-15
fix(feature): sample SOLD2 line descriptors with the reference grid normalization

### LINKED ISSUE #4554: SOLD2 samples line descriptors with a corner-aligned grid under align_corners=False, drifting up to one image pixel from the reference implementation
### What happened?

kornia's SOLD2 port samples line descriptors at slightly different positions from the reference implementation the pretrained weights were trained with, and the gap grows toward the bottom-right of the image.

`WunschLineMatcher.forward` builds its sampling grid with `keypoints_to_grid` (`kornia/feature/sold2/sold2.py:331`), which normalizes through `normalize_pixel_coordinates`, i.e. the corner-aligned mapping `2p / (size - 1) - 1`. It then reads the descriptor map with `F.grid_sample(..., align_corners=False)`, which interprets `[-1, 1]` as the outer pixel edges. The reference implementation ([cvg/SOLD2 `sold2/misc/geometry_utils.py`](https://github.com/cvg/SOLD2/blob/main/sold2/misc/geometry_utils.py)) normalizes with `keypoints * 2 / img_size - 1` and also samples at `grid_sample`'s default `align_corners=False`, so its grid and its sampler use the same convention.

The two normalizations differ by `2p / (W (W - 1))` in normalized units, which is `p / (W - 1)` image pixels: zero at the origin and one full image pixel at the far edge, independent of the descriptor stride. It has been this way since the port landed in #1844.

Measured on a 480x640 image with the default `grid_size=8`, so a 60x80 descriptor map. Columns are the descriptor-map coordinate that `grid_sample` actually reads:

| keypoint `(i, j)` | kornia reads | upstream reads | difference, desc px | difference, image px |
|---|---|---|---|---|
| (0, 0) | (-0.500, -0.500) | (-0.500, -0.500) | 0 | 0 |
| (240, 320) | (39.563, 29.563) | (39.500, 29.500) | +0.063 | +0.50 |
| (479, 639) | (79.500, 59.500) | (79.375, 59.375) | +0.125 | +1.00 |
| (100, 639) | (79.500, 12.026) | (79.375, 12.000) | +0.125 / +0.026 | +1.00 / +0.21 |

At the far edge kornia's sample sits exactly half a descriptor pixel past the last centre, so half its bilinear weight lands in zero padding while upstream's does not. On a unit-norm random descriptor map the sampled norms at `(479, 639)`, `(100, 639)` and `(479, 300)` come out `0.250`, `0.486`, `0.468` for kornia against `0.391`, `0.625`, `0.625` for upstream. `F.normalize` afterwards undoes the scale, so what survives is the shift itself and a different blend of neighbouring descriptors along the right and bottom borders.

The magnitude is small, at most one image pixel and one eighth of a descriptor pixel, so this is a fidelity deviation from the reference weights rather than a gross failure. It is the same corner-aligned-versus-edge mismatch as #3904, on a code path #3945 did not touch.

### Minimal reproduction

```python
import torch
from kornia.feature.sold2.sold2 import keypoints_to_grid

def upstream_grid(kp, img_size):  # cvg/SOLD2, sold2/misc/geometry_utils.py
    n = kp.shape[-2]
    g = kp.float() * 2.0 / torch.tensor(img_size, dtype=torch.float32) - 1.0
    return g[..., [1, 0]].view(-1, n, 1, 2)

H, W, g = 480, 640, 8
hd, wd = H // g, W // g
kp = torch.tensor([[0.0, 0.0], [240.0, 320.0], [479.0, 639.0]])          # ij, image pixels

def where_grid_sample_reads(grid):                                     # align_corners=False inverse
    return (grid[0, :, 0, :] + 1) * torch.tensor([wd, hd]) / 2 - 0.5

print(where_grid_sample_reads(keypoints_to_grid(kp, (H, W))))
print(where_grid_sample_reads(upstream_grid(kp, (H, W))))
```

Observed:

```
tensor([[-0.5000, -0.5000], [39.5625, 29.5625], [79.5000, 59.5000]])   # kornia
tensor([[-0.5000, -0.5000], [39.5000, 29.5000], [79.3750, 59.3750]])   # upstream
```

Expected: the two agree, since the weights were trained with the second.

### Environment

- Kornia version: 0.9.0rc1 (current `main`, `1590df48`)
- Python version: 3.11.14
- OS / device: Linux, CPU

### Anything else?

The fix is one line and self-contained: have `keypoints_to_grid` compute `keypoints * 2 / img_size - 1` as upstream does, instead of calling `normalize_pixel_coordinates`. It should not wait for #4504, and it should not use the `align_corners` parameter that issue proposes either. The half-pixel mapping `(2p + 1) / W - 1` is not what upstream does; it differs from upstream's formula by a constant half image pixel, and the target here is the convention the weights were trained under, not the pixel-centre convention. `keypoints_to_grid` is a module-level helper, not exported and not tested, so this is not a public API change.

Upstream also samples line scores and descriptors in two further places (`line_matching.py` lines 163 to 174 and 246 to 249). kornia's port has only the one site, so nothing else inherits this.

Found while surveying the callers of `normalize_pixel_coordinates` for #4504. Investigated with Claude Code; every figure above is from an executed run on current `main`.

VanniLeonardo:

`keypoints_to_grid` in `kornia/feature/sold2/sold2.py` normalized keypoints with the corner-aligned `normalize_pixel_coordinates` (`2p / (size - 1) - 1`) and then `WunschLineMatcher` sampled the descriptor map with `F.grid_sample(..., align_corners=False)`. The reference implementation (cvg/SOLD2, `sold2/misc/geometry_utils.py`) normalizes with `keypoints * 2 / img_size - 1` and samples with the same `align_corners=False`, so kornia's grid drifted from the one the pretrained weights were trained with by `p / (size - 1)` image pixels, zero at the origin and one full image pixel at the far edge. At the far edge that is half a descriptor pixel past the last centre, so half of the bilinear weight landed in zero padding.

This replaces the normalization with the reference formula. It is a module-level helper, not exported and not in `tests/api_surface.json`, so there is no public API change. Integer keypoints are promoted to float as upstream does. The docstring now says which convention it implements and why it is not `normalize_pixel_coordinates`.

Measured on a 480x640 image with `grid_size=8`, as the descriptor-map coordinate `grid_sample` reads:

| keypoint `(i, j)` | before | after (= upstream) |
|---|---|---|
| (0, 0) | (-0.5, -0.5) | (-0.5, -0.5) |
| (240, 320) | (39.5625, 29.5625) | (39.5, 29.5) |
| (479, 639) | (79.5, 59.5) | (79.375, 59.375) |
| (100, 639) | (79.5, 12.026) | (79.375, 12.0) |

This is the fix the issue proposes. It does not use the half-pixel `align_corners` mapping that #4504 will add to `normalize_pixel_coordinates`, because that differs from upstream's formula by a constant half image pixel and the target here is the convention the weights were trained under.

New tests in `TestKeypointsToGrid`: smoke, exceptions, integer promotion, bitwise agreement with the upstream formula plus the table above as a literal, a cell-centre pin (the image pixel at the centre of a descriptor cell reads exactly that descriptor under `align_corners=False`, on power-of-two sizes so it is exact at half precision), and dynamo. With the test file on `main`'s tree, `test_convention_matches_reference_normalization_4554`, `test_convention_cell_centre_reads_its_own_descriptor_4554` and `test_integer_keypoints_are_promoted` fail; the other pins pass on both trees.

Fixes #4554. Found while surveying the callers of `normalize_pixel_coordinates` for #4504.

## What did you check?

```text
$ pytest tests/feature/test_sold2.py --dtype=float32,float64,float16,bfloat16     17 passed, 26 skipped
$ KORNIA_TEST_RUNSLOW=true pytest tests/feature/test_sold2.py --dtype=float32       9 passed, 4 skipped
$ KORNIA_TEST_OPTIMIZER=inductor pytest tests/feature/test_sold2.py -k "TestKeypointsToGrid and dynamo"
                                                                                    1 passed
$ pytest --doctest-modules kornia/feature/sold2                                     2 passed, 1 skipped
$ pixi run typecheck                                                                All checks passed
```

`WunschLineMatcher` runs end to end on random segments and descriptors and returns the same shape and dtype as before. Pre-commit passes on both changed files. `kornia.__file__` named the branch worktree in every run, and the `main` comparison ran from a `main` worktree.

## How did AI help?

Claude Code did the investigation, the change, the tests and this description. I directed the work, reviewed every line, and every figure above comes from a run I executed on this branch against a `main` worktree.


ducha-aiki (dismissed):
Review of `7dfab8f0`. Verified end to end: this is the right fix, correctly pinned. **Approving**; four nits below, none blocking.

### What I reproduced

On a main worktree the four keypoints of your table read `(-0.5,-0.5)`, `(39.5626,29.5626)`, `(79.5,59.5)`, `(79.5,12.0261)` and on this branch `(-0.5,-0.5)`, `(39.5,29.5)`, `(79.375,59.375)`, `(79.375,12.0)` — your table exactly, and `torch.equal(head_grid, upstream_grid)` is `True`.

I checked the reference itself rather than the quote. cvg/SOLD2 `sold2/misc/geometry_utils.py::keypoints_to_grid` is `keypoints.float() * 2. / torch.tensor(img_size) - 1.` then `[..., [1, 0]].view(-1, n_points, 1, 2)`, and every `grid_sample` in `sold2/model/line_matching.py` (62-63, 201, 209, 370-372 — the inference path, not just the loss) is called with no `align_corners`, i.e. PyTorch's default `False`. Path, function name and `.float()` in your description are all accurate. The paper is SOLD², *Self-supervised Occlusion-aware Line Description and Detection*, Pautrat\*, Lin\*, Larsson, Oswald, Pollefeys, CVPR 2021 (arXiv:2104.03362). Aside: the *issue* cites upstream lines 163-174 and 246-249; on upstream `main` the sites are 60-63, 196-209 and 368-372. The substantive claim holds.

### That it is an improvement, not just a change

With a coordinate-encoding descriptor map the read position on main is off by up to `0.120 desc px = 0.960 image px`, on this branch by `0.000000`. Descriptor cosine against the reference over 400 random line points: main min `0.9486` / mean `0.9874`; branch bitwise identical. Cell-centre identity: main max abs err `0.0927`, branch `1e-6`. On a synthetic 200-line pair (100 true + 100 decoys) through the full `WunschLineMatcher`, true positives go `95 → 97/100` and false positives `20 → 18/100`. Translation consistency at 8/128/0.5/3.25 px: the branch is exactly equivariant (`min 1.000000`), main drifts to `0.999942` at 128 px.

Robustness bonus worth knowing: the new mapping reduces to `p/grid_size - 0.5` independently of `img_size`, so `WunschLineMatcher` reconstructing `img_size` from `desc.shape[2:] * grid_size` (`sold2.py:191-192`) no longer affects the read — `480x640`, `256x320` and `32x64` all give `desc x 12.000000` for image pixel 100. Main's `/(S-1)` form did depend on it.

### Pin quality

Reverting `sold2.py` to `90650596d` with the tests kept fails exactly the three tests you name (`3 failed, 2 passed, 8 skipped`). Of 13 one-token mutations of the new code, 12 die — including `2/(size-1)`, `+1` for `-1`, swapped `img_size` axes, a dropped `[1,0]` reorder, x-uses-height, swapped output components, `*1.0` for `*2.0`, a dropped `-1.0`, `.double()`, `.half()`, and `KORNIA_CHECK_SHAPE` removal. dtype/device preservation is identical to main for every input dtype on CPU and MPS, and the new code allocates no tensor at all. No API surface change (`keypoints_to_grid` is absent from `tests/api_surface.json` and from the package `__init__`), and `kornia/feature/sold2/` has exactly one `grid_sample` pair, so the scope is complete.

Nits:

1. `changelog.d/4558.fixed.md` and the docstring say the change is felt "near the right and bottom borders". It is not confined there — the drift is `p/(size-1)` image pixels, so half an image pixel at the image *centre*. In my run 6 of 200 match decisions change, and four are 24-118 px from the right/bottom borders (one is at rows 0-2, cols 89-121). Suggest "match decisions can change anywhere except at the top-left origin; the shift grows from zero there to one image pixel at the far edge."

2. `sold2.py:349-350` is the only mutation I could not kill: dropping `if not xy.is_floating_point(): xy = xy.float()` leaves all 17 tests green, because `int64 * 2.0` already promotes. The one configuration where the branch is observable is `torch.set_default_dtype(torch.float64)`, and there it *hurts*:

   ```
   int64 kp under default_dtype=float64 -> grid dtype torch.float32
   grid_sample(float64 desc, that grid) -> RuntimeError: expected scalar type Double but found Float
   without the .float() branch:            torch.float64 -> grid_sample OK
   ```

   `sample_line_points` (`sold2.py:255-256`) always yields float, so the branch is unreachable from the library's own path. I would delete the two lines. Relatedly, `test_integer_keypoints_are_promoted`'s `assert grid.is_floating_point()` also passes on main, so that test fails there only via its value assertion — the promotion in its name is not actually pinned.

3. `sold2.py:337-338` ("the image pixel at the centre of a descriptor cell reads that cell exactly") mixes conventions by half a pixel. Under kornia's documented pixel-centre convention (`docs/source/get-started/conventions.rst:53-55`) cell *i* pools image pixels `g*i … g*(i+1)-1`, whose centre `g*(i+0.5)-0.5` reads `i - 0.125`, not `i`. Executed at `H=W=32, g=4`: the docstring's `p = g*(i+0.5)` reads `[0.0, 1.0, 2.0, ...]` while the cell centre reads `[-0.125, 0.875, 1.875, ...]`. The docstring rules out `normalize_pixel_coordinates` but not the other candidate: this formula is exactly `+0.5` image pixels from #4504's proposed `(2p+1)/W - 1` at *every* `p`. Please say so, in the shape used at `kornia/feature/xfeat.py:262-265`, so the next convention sweep does not "fix" it.

4. `tests/feature/test_sold2.py:103,105`: `pytest.raises(Exception)` without `match=`. It does kill the removal of `KORNIA_CHECK_SHAPE` (neither malformed input raises on its own), but `Exception` would also swallow an unrelated `IndexError`; the check raises `ShapeError`, and `match=` is already the majority style in `tests/feature` (13 vs 12).

On the fragment class: `fixed` is defensible and I am not asking you to change it. For the record `changelog.d/3945.breaking.md` covers the same defect class, but `4540.fixed.md` (`Resize(antialias=True)` no longer antialiasing masks — a larger silent output change) and `4495.fixed.md` are the closer and more recent precedent, and your fragment already states old and new behaviour.

Runs: `--dtype=float32,float64` → 9 passed / 14 skipped; float16 and bfloat16 under their profiles → 5 passed / 8 skipped each; `KORNIA_TEST_OPTIMIZER=inductor -k dynamo` → 1 passed; `KORNIA_TEST_RUNSLOW=true` → 9 passed / 4 skipped; `--doctest-modules kornia/feature/sold2` → 2 passed / 1 skipped. `tests/test_half_precision_ci.py` + `tests/test_api_surface.py` → 209 passed (the eager-RNG inventory needs no edit — the new `torch.rand` at `test_sold2.py:143` is in a method body, which the visitor does not descend into). No manifest entries for SOLD2 in any of the three manifests. `ruff@0.16.7` clean, max line 119, Apache headers present, and `typing.Tuple` correctly kept without `from __future__ import annotations` for this TorchScript-scriptable module. `grep -rnE "#4554|issues/4554"` finds only the two pins, and `Fixes #4554` is accurate and complete.


ducha-aiki (approved):

PR #4561, opened 2026-09-15
Fix float16 HSV saturation and gradients at black pixels

### LINKED ISSUE #4560: `rgb_to_hsv` returns NaN saturation for black pixels in float16, and `adjust_hue` / `ColorJiggle` / `ColorJitter` spread it
`kornia.color.rgb_to_hsv` computes the saturation as `deltac / (max_rgb + eps)` with `eps = 1e-8` (`kornia/color/hsv.py:61`). In `float16` the smallest subnormal is about `6e-8`, so `1e-8` rounds to `0.0`, and a black pixel (`max_rgb == 0`, `deltac == 0`) divides `0 / 0` and gets `NaN` saturation. `float32`, `float64` and `bfloat16` are finite for the same input. Everything built on the HSV round trip inherits it: `kornia.enhance.adjust_hue` on a black pixel is `NaN` in every channel, and `ColorJiggle` / `ColorJitter` with any non-zero `hue` turn every black pixel of a `float16` image into `NaN` — a black border or a masked-out region is enough. The same on `mps`.

## Repro

```python
from kornia.color import rgb_to_hsv
from kornia.enhance import adjust_hue

for dev in ["cpu", "mps"]:
    for dt in [torch.float16, torch.bfloat16, torch.float32]:
        if dev == "mps" and dt is torch.bfloat16:
            continue
        z = torch.zeros(1, 3, 1, 1, device=dev, dtype=dt)
        hsv, hue = rgb_to_hsv(z), adjust_hue(z, 0.1)
        img = torch.rand(1, 3, 6, 8, device=dev, dtype=dt)
        img[..., :2, :3] = 0.0                                   # a black patch
        torch.manual_seed(0)
        out = K.ColorJiggle(0.0, 0.0, 0.0, 0.1, p=1.0)(img)
        if dev == "mps":
            torch.mps.synchronize()
        print(f"{dev} {str(dt):16s} hsv(black)={hsv.flatten().tolist()} adjust_hue(black) nan={hue.isnan().any().item()} "
              f"ColorJiggle nan={int(out.isnan().sum())}/{out.numel()}")
```

## Observed

```
cpu torch.float16    hsv(black)=[0.0, nan, 0.0] adjust_hue(black) nan=True ColorJiggle nan=12/144
cpu torch.bfloat16   hsv(black)=[0.0, 0.0, 0.0] adjust_hue(black) nan=False ColorJiggle nan=0/144
cpu torch.float32    hsv(black)=[0.0, 0.0, 0.0] adjust_hue(black) nan=False ColorJiggle nan=0/144
mps torch.float16    hsv(black)=[0.0, nan, 0.0] adjust_hue(black) nan=True ColorJiggle nan=12/144
mps torch.float32    hsv(black)=[0.0, 0.0, 0.0] adjust_hue(black) nan=False ColorJiggle nan=0/144
```

`torch.finfo(torch.float16)`: `eps = 9.77e-04`, `tiny = 6.10e-05`, smallest subnormal `5.96e-08`; `torch.tensor(1e-8, dtype=torch.float16)` is `0.0`. A grey pixel at `1e-3` is finite (`[0.0, 0.0, 0.0010]`), so the failure is exactly the `max_rgb == 0` case.

## Expected

A black pixel has saturation `0` in every dtype. The guard has to be an argument that is representable in the input's dtype — for example `torch.where(max_rgb == 0, ones, max_rgb)` as the divisor (the pattern the same function already uses for `deltac` at `hsv.py:63`), or an `eps` floored at the dtype's smallest normal — so that the division never sees `0 / 0`. `adjust_hue`, `ColorJiggle` and `ColorJitter` then return the black pixel unchanged, as they do in `float32`.

## Classification

**Broken output** — `NaN` from a valid input. Idiom: 1 (`x + eps` used as a singularity guard, with an `eps` that underflows in the dtype; see #4162 and the `eps` floor pattern).

## Disposition

`n/a — single symbol` — one line in `rgb_to_hsv`; the three callers need no change. Focused fix, welcome as a PR.

## Environment

kornia 0.9.0rc1 @ `90650596` (main), torch 2.14.0, python 3.11, cpu and mps (Apple M1, `PYTORCH_ENABLE_MPS_FALLBACK` unset). Found while writing the 2D intensity convention pins of the augmentation conventions audit (batch 6, tracking issue #4407, sub-batch 6c): the `ColorJiggle` / `ColorJitter` pins skip `float16` on a fixture that contains zeros because of this, and cite this issue as the reason.

Shy7777:

Black pixels currently produce NaN saturation in float16 because the default `eps` rounds to zero. Hue adjustment then spreads those NaNs into the image, including ordinary black padding or masked regions.

Use a safe saturation divisor at black pixels before dividing, while retaining `max_rgb + eps` for nonblack pixels. This keeps the black-pixel backward pass finite too, including when a small nonzero `eps` would overflow the reciprocal in float16. The tests cover black and dark nonblack pixels, custom `eps`, `adjust_hue`, `ColorJiggle`, and `ColorJitter`; the existing half-precision skip is removed.

The fix also recovers 27 existing float16 augmentation tests. Remove their stale expected-failure entries, leaving the remaining baseline and all test assertions unchanged.

Fixes #4560.

## What did you check?

- The new float16 regressions fail on the unchanged source: 6 failed, 1 passed.
- HSV, hue adjustment, and black-pixel augmentation regressions on CPU, float16/bfloat16/float32/float64: 153 passed.
- The same selection on MPS, float16/float32: 72 passed, 5 skipped because MPS does not support the float64 gradchecks.
- Existing ColorJiggle/ColorJitter tests on CPU, float32/float64: 60 passed, 1 skipped, 2 expected failures.
- HSV compilation tests with `aot_eager`, float16/float32/float64: 6 passed.
- Full `pixi run pre-commit-all` and changelog preview passed.
- The complete Linux CPU float16 CI profile on PyTorch 2.9.1 reported 10,323 passed and 27 strict XPASS results, with no ordinary test failures. Those 27 recovered cases are the only entries removed from the known-failure manifest.

Tested with Python 3.11 and PyTorch 2.14.0. CUDA was not tested locally.

## How did AI help?

AI assistance was used for the implementation and regression tests. The checks listed above were run locally against this branch.


ducha-aiki (changes requested):
Review of `1159170f`. The fix itself is right and the test quality is high — I could not break it. Two blocking items, both from #4566 landing after your base.

**The production change is correct and verified on the torch floor.** `kornia/color/hsv.py:62` uses `torch.where`, not `clamp`, so the unbounded derivative never reaches backward: `s_divisor = torch.where(max_rgb == 0, torch.ones_like(max_rgb), max_rgb + eps)` has two finite branches. Checked on a fresh `torch==2.5.1` arm64 venv — kornia's floor, where `clamp`-based guards still leak `inf`/`nan` — forward and `autograd.grad` are finite at black, grey, white and the zero-max pixel `(0, -0.5, -0.5)` in float16, bfloat16, float32 and float64, while main gives `[0.0, inf, 0.0]` / `grad_finite=False` there. `tests/color/test_hsv.py` + `tests/enhance/test_adjust.py` are `163 passed` per dtype under 2.5.1.

**No ordinary pixel moves.** Over an exhaustive 17³ = 4913-pixel grid on `[0, 1]` the element-wise diff against main is exactly **1 pixel of 4913, float16 hsv only** (the black pixel, nonfinite 1→0); float32/float64/bfloat16 forward, round-trip and gradient hashes are bitwise identical, and `rgb_to_hls` is untouched. The wrappers go from 12/144 to 0/144 float16 NaNs with random-image output hashes bitwise identical in all four dtypes. All three one-token mutations of the new line are killed in all four dtypes (`== 0`→`!= 0`: 14/9/9/9 failures; `ones_like`→`zeros_like`: 7 each; dropping `+ eps`: 1-2 each). Manifest discipline is exact: all 27 deleted lines are in the 3 files they name, the merged-tree float16 run gives `267 passed, 4 xfailed, 0 failed, 0 xpassed`, and the xfail count across `tests/augmentation` drops 235→208 — exactly 27. Reverting the production file with the tests kept fails 6 nodes in float16 *and* 1 in float32, so the pin is not float16-only.

Blocking:

### [P2] #4566 added pins that this fix falsifies

Merging `30dfbf711` with your head and running the float16 leg gives 6 failures:

```
$ pytest tests/augmentation/test_conventions_intensity_values.py \
         tests/augmentation/test_conventions_intensity_ops.py \
         -q --device=cpu --dtype=float16 --xfail-known-failures --known-failure-profile=cpu-float16
6 failed, 324 passed
  test_wart_hsv_path_black_pixel_is_nan_in_float16_4560[cpu-float16-RandomHue]
  ...[RandomSaturation] ...[ColorJiggle] ...[ColorJiggleSaturation] ...[ColorJitter]
      assert [False, False] == [True, True]          test_conventions_intensity_values.py:1941
  test_convention_random_hue_keeps_out_of_range_extremes[cpu-float16]
      assert bool(collapsed.isnan().any())           :976
```

float32 (337 passed), float64 (337), bfloat16 (335) and MPS float32 (465) are green, so this is float16-only. Please rebase onto main, invert `test_wart_hsv_path_black_pixel_is_nan_in_float16_4560` into a convention test — a black pixel and the zero-max pixel `(0, -0.5, -0.5)` are finite in **every** dtype now, with the head giving saturation `0.5` for the latter where main gave `5e7` in float32/float64 and `inf` in float16 — drop the float16 branch at `:976`, and rewrite the generation-snippet comment blocks above both.

### [P2] Four `kornia/` warnings still describe the bug

`grep -rnE "#4560|issues/4560" kornia/ tests/`: `saturation.py:60-62`, `hue.py:61-63` **and the Convention clause at `hue.py:59`** ("comes back as zeros, or as NaN in ``float16``" — only the "zeros" half survives), `color_jiggle.py:86-90`, `color_jitter.py:98-101`. AGENTS.md treats a surviving `kornia/` hit for a closed issue as an incomplete change. Five `tests/` skip/`continue` reasons citing #4560 also need re-checking in the same PR: `test_conventions_intensity_values.py:240`, `:400`, `:447`, `:846`, `:959-963`.

### Informational

The float32 gradient at an exactly-black pixel changes from `[0, 2, -2]` to `[0.3333, 1.3805, -0.7139]` — which is exactly main's *float64* value, so the change harmonises float32 with float64 rather than regressing it. Worth a clause in the fragment. Also `test_dark_red_saturation` passes with and without the fix; it is a guard against a blanket-eps-floor alternative, not a pin on this one.

**Merge-order note:** this PR and #4578 both delete adjacent lines in `testing/half_precision_xfails/cpu_float16.txt` and conflict in either order — whichever lands second needs a keep-both manifest merge. #4583 is independent.


ducha-aiki:
I merged current `main` into this branch and did the #4566 reconciliation — head is now `002098b0`, and the PR is **no longer `CONFLICTING`**. It was `DIRTY`, so it was running no CI at all. **This was our debt, not yours.**

### The conflict: the known-failure manifest

`cpu_float16.txt` conflicted because `main` moved under you (#4471 landed 33 manifest edits). I resolved it by set arithmetic rather than by hand-picking hunks:

```
merge base  527 lines
this branch 500   (removes 27, adds 0)
main        522   (removes 6, adds 1)
overlap between the two removal sets: 0
merged      495 = main's 522 minus your 27
```

No line is removed twice and none of your 27 is one main also touched, so the merge is exact.

### The 6c work

**Inverted the wart pin** into `test_convention_hsv_path_zero_max_pixel_is_finite_4560`, keeping its five-class parametrization (the two saturation-only configurations included, since `ColorJiggle`'s saturation step is an HSV round trip too).

**Added a gradient pin**, `test_convention_rgb_to_hsv_black_pixel_gradient_is_finite_4560`. Your guard is a `torch.where` and not a clamp, which is the right choice and the reason it holds in backward — AGENTS.md documents the clamp trap under #4229, where the value is bounded but the unbounded derivative still reaches the backward pass. Nothing in the forward pins would have caught a clamp, so that property deserved its own pin.

**Rewrote four docstrings** (`saturation.py`, `hue.py`, `color_jiggle.py`, `color_jitter.py`) and **deleted five float16 skip branches** in the convention suite that existed only to route around the NaN. `grep -rnE "#4560|issues/4560" kornia/ docs/` is clean.

**Two further pins encoded the NaN and had to move with it** — these are the ones a plain merge would have left silently wrong, since git merges them cleanly:

- the #4430 collapse census excused `ColorJitter` from the expected set in float16; with the fix it collapses like every other dtype, so it goes back **in** rather than being excused.
- `test_convention_random_hue_keeps_out_of_range_extremes` asserted `isnan()` for the zero-max pixel in float16; it now asserts zeros in every dtype.

### Verification

| Leg | Result |
| --- | --- |
| `tests/augmentation` + `tests/color` + `tests/enhance`, float32 | 4102 passed |
| same, float64 | 4100 passed |
| same, bfloat16 with the profile | 4031 passed, 0 failed |
| convention + hsv suites, **MPS** float32 | 194 passed |
| convention + hsv + adjust, **torch 2.5.1** | 355 passed |

**Your fix also clears four float16 nodes that are still red on `main`.** Running the same float16 command on a pristine `main` worktree and on this one:

```
main       : 5 failed  (test_forward[1], [random_apply2], [random_apply3],
                        test_same_on_frame[random_apply1-BCTHW-aug0], [BTCHW-aug0])
this branch: 1 failed  (test_forward[random_apply3] only)
```

**On that last one, a warning so nobody deletes the wrong line.** `test_forward[cpu-float16-random_apply3]` reports a strict XPASS here — but it does so on **pristine `main` too**, with no PR applied. It is pre-existing manifest drift and not yours, so I deliberately left the line in place. The manifest's header says `# os: Linux` and the blocking half jobs are Linux-only; this node passes on macOS arm64 and the Linux leg was green when #4471 added it, so the platforms genuinely disagree. Removing it to make a local run quiet would turn the Linux leg red.

Your `hsv.py` change and changelog fragment are untouched.

ducha-aiki:
#4578 merged this morning and broke this branch in `testing/half_precision_xfails/cpu_float16.txt`. I resolved and pushed the merge as `f6f375e3`, so it is MERGEABLE again and CI is running.

**The conflict was mechanical.** #4578 fixed the Planckian Jitter dtype wart and therefore deleted four `TestPlanckianJitter` entries from the manifest; this branch deletes 27 entries for the HSV nodes it repairs. Both sides only delete, and the sets are disjoint:

```
base entries        522
this branch deletes  27
main deletes          4
overlap               0
either side adds      0
resolved            491
```

So the resolution is the union of the deletions. I asserted the disjointness and the no-additions property in the script rather than eyeballing it, so a non-mechanical case would have stopped rather than silently merged.

**Verified against reality, which is the part that matters for a strict-xfail manifest.** A deleted line whose node still fails, or a retained line whose node now passes, reds the blocking leg. On the merged tree:

```
tests/augmentation --device=cpu --dtype=float16 \
    --xfail-known-failures --known-failure-profile=cpu-float16
  1 failed, 3199 passed, 3434 skipped, 198 xfailed, 5 xpassed
```

The single failure is `test_sequential_aug.py::TestSequential::test_forward[cpu-float16-random_apply3]`, `XPASS(strict)`. **It is not caused by this merge** — I ran the same node on pristine `main` on this machine and it `XPASS(strict)`es there too. It is the known macOS-local flip of a Linux manifest entry: `random_apply3` draws its input, so whether it passes depends on the draw, and this machine's draw differs from the Linux runner's. No newly-red node relative to `main`.

That said, it is the same node your `half-precision (float16)` leg was already red on before this merge (`1 failed, 10704 passed`, same `XPASS(strict)`, issue #4153). So it is worth deciding separately: either the `cpu_float16.txt` line for `random_apply3` goes, or the node gets seeded so its outcome stops depending on the draw. **Re-running the job is not the answer** — it is a coin flip, not a flake in the usual sense. That is not a request on this PR; the entry predates it.

Nothing else changed: `git diff origin/main HEAD` is this branch's own work plus the manifest resolution.

ducha-aiki (changes requested):
Re-review of `f6f375e33c7b6e1fbe5387c0ac4597e933c50dc5` after the mechanical merge with current `main`.

The production HSV fix remains correct and the merge preserves the disjoint #4578 manifest deletions. One stale known-failure entry now blocks the branch.

### [P1] Remove or deliberately stabilize the strict-XPASS entry — `testing/half_precision_xfails/cpu_float16.txt:30`

The blocking Linux float16 job deterministically passed `TestSequential::test_forward[cpu-float16-random_apply3]`, but the manifest still requires `call builtins.AssertionError`. The strict known-failure harness therefore turns that recovery into the branch's only test failure (`1 failed, 10812 passed, 525 xfailed`), and `collector` fails with it. The exact focused node also reproduces as strict XPASS locally under the profile. Remove this manifest row if the recovered outcome is now the contract, then rerun the full profile as required by the manifest rules; if the node is intentionally nondeterministic, make its input deterministic before retaining or updating the row.

ducha-aiki (approved):
Re-review of `24096e73f7563760adb10505b40befff443d2deb`: **APPROVE**. The remaining review blockers are resolved.

Merged main `a5d724265` without rewriting the contributor's history. The intensity census now includes both repaired float16 operations: main repaired RandomSnow through HLS, and this branch repairs ColorJitter through HSV. Keeping either side's old exclusion would have hidden one repaired operation, so the merge removes both exclusions and both expected-set adjustments.

Removed the stale manifest entry for `TestSequential::test_forward[cpu-float16-random_apply3]`. This is supported by the canonical Linux x86_64 / torch 2.9.1 job 105168747898: 10,812 tests passed and the sole failure was this exact strict XPASS. The case wraps RNG in `fork_rng()` and sets `torch.manual_seed(0)` before constructing its input and augmentation, so this is not a decision based on an unseeded run.

Validation on the merged head: the census passed at CPU float16 and bfloat16 (2 tests), ColorJitter/RandomSnow repair checks passed (6 tests), the exact recovered sequential node passed, and manifest infrastructure passed 63 tests. The worker's broader focused sequential selection passed 7 tests. `pixi run pre-commit-all` passed every hook, and diff/conflict-marker checks are clean.

No open code-review findings remain. The full canonical Linux float16 profile must pass in the fresh CI run; it has not been claimed as locally verified on this macOS arm64 / torch 2.14 host.

PR #4562, opened 2026-09-15
test(conversions): pin the 3-D dtype guard too

rawsun007:
The [P3] item from your review of #3974, which merged without it.

`test_convention_non_default_float_and_complex_dtypes_are_accepted_3959` only called `normal_transform_pixel`, and the guard exists twice. I reproduced your measurement before writing anything: reverting the probe in `normal_transform_pixel3d` alone, leaving the 2-D one intact, keeps `tests/geometry/test_conversions.py` green at 419 passed on this tree, while `normal_transform_pixel3d(2, 4, 5, dtype=torch.float8_e4m3fn)` raises.

With the 3-D call added, that same mutation fails the `accepted_dtype3` cell with `ValueError ... Got torch.float8_e4m3fn.` at `conversions.py:2346`.

The assertion mirrors the 2-D one: dtype, shape, device, then the value widened to `complex128` and compared on cpu, with the 3-D expectation carrying the `2.0` depth scale that `depth=2` produces.

Verification: `tests/geometry/test_conversions.py` 419 passed, 1 skipped, 7 xfailed on cpu; `--device=mps -k NormalTransformPixel` 35 passed, 1 skipped; `ruff check` and `ruff format --check` clean.

Refs #3959, which stays open for the four functions neither PR touches.

No changelog fragment: this is test-only, which `changelog.d/README.md` lets omit one under the `no-changelog` label. Nothing observable to a user changed.

Assisted-by: Claude Opus 5 (Claude Code), under my account. I have not read the diff line by line myself yet.


ducha-aiki (changes requested):
Review of `5f016c8e`. The pin holds up, and I am asking for one small change before this merges.

I reproduced your measurement independently. Dropping the probe in `normal_transform_pixel3d` alone (`kornia/geometry/conversions.py:2346-2348`, reduced to allowlist-only rejection) makes the 3-D call raise while the 2-D call still succeeds:

```
normal_transform_pixel   float8 OK     torch.float8_e4m3fn
normal_transform_pixel3d float8 RAISE  ValueError dtype must be a floating point or complex type
```

`tests/geometry/test_conversions.py` at the merge base is blind to that regression — `419 passed, 1 skipped, 7 xfailed` — and with your added lines the `accepted_dtype3` cell fails (`1 failed, 418 passed`). So the new coverage is real, and it is exactly the `float8_e4m3fn` cell: the one dtype that reaches the probe rather than the allowlist fast path. The rejection half for the 3-D function was already pinned by `test_convention_integer_dtype_rejection_is_unconditional_3959`, so acceptance is precisely the gap, and the PR claims no more than that.

The value assertion is not decorative either: mutating `sz` from `2.0 / (depth - 1.0)` to `3.0 / (depth - 1.0)` gives 14 failures here against 10 on the base file. And the expected diagonal `[0.5, 2/3, 2.0]` for `(depth=2, height=4, width=5)` is three distinct values, so a permutation between the `x`, `y` and `z` slots would be caught too — that is the relabeling hole the 3-D function's own convention block warns about. Your reported `419 passed` and the MPS `35 passed, 1 skipped` both reproduce.

**[P2] Please delete `changelog.d/4562.fixed.md`.** This is a test-only change, so it should not carry a fragment. The file currently reads "The acceptance test for the `normal_transform_pixel` dtype guard now covers the 3-D function as well…", which Towncrier would publish as a *Bug fixes* bullet in the release notes — telling users a test improved, when no behavior they can observe changed. `changelog.d/README.md:30-36` covers exactly this case with the `no-changelog` label, and the explanation already in your PR body is what that label needs. I have applied the label, so CI's presence check is satisfied; the fragment just needs to go, otherwise it will be consumed into the next release's notes.

Minor: please add `Refs #3959` to the body. The prose "Relates to #3959" creates no cross-reference, and it would be useful for that thread to record that two of its clauses now have pins while the other four functions remain open.

With the fragment removed this is good to go — CI is fully green on this head (40/40, including `tests-mps` and all four macOS/Windows torch legs).


rawsun007:
Both done.

Fragment deleted in `af6c2b6` — `changelog.d/README.md:32-34` says exactly what you say, and with the `no-changelog` label applied the presence check is satisfied. The body now records why there is no fragment, and `Relates to` is now `Refs #3959` so the cross-reference actually lands on that thread.

`tests/geometry/test_conversions.py` still 419 passed, 1 skipped, 7 xfailed after the removal.

Thanks for re-deriving the mutation independently rather than taking the numbers on trust — the `sz` mutation at 14 against 10 is a sharper statement of what the value assertion buys than the one I put in the body, and the point that the three diagonal entries are distinct so a slot permutation is caught is not something I had said out loud.

Written by Claude Opus 5 in Claude Code under @rawsun007's account; he has not read the diff line by line yet.

ducha-aiki (approved):
## Re-review — #4562, head `af6c2b6f` — **approving**

Both asks from the last round are discharged: the changelog fragment is gone and `no-changelog` is applied. CI is 39/39 SUCCESS and the merge onto `ecbcad561` is clean — I rebuilt it rather than trusting the label, and the post-merge diff is exactly your 18 added lines in `tests/geometry/test_conversions.py`, nothing else.

**The one re-check I owed you: does the added 3-D block actually discriminate?** Yes.

I planted the failure mode your comment names — an allowlist in the **3-D** guard only, i.e. replacing the probe at `kornia/geometry/conversions.py:2346-2348` with an unconditional raise for anything outside the literal tuple, leaving `normal_transform_pixel` untouched:

```
FAILED tests/geometry/test_conversions.py::TestNormalTransformPixel::
  test_convention_non_default_float_and_complex_dtypes_are_accepted_3959[cpu-accepted_dtype3]
  - ValueError: dtype must be a floating point or complex type. Got torch.float8_e4m3fn.
1 failed, 10 passed
```

The traceback lands on `conversions.py:2346`, reachable only through the `normal_transform_pixel3d` call you added — so the new lines are exactly what catches it, and without them this mutation would have gone unnoticed. Your inline comment ("an allowlist planted in that one alone would leave this test green if it only called the 2-D function") is literally correct.

Restored and re-ran clean: `11 passed` at `--dtype=float32,float64`.

Good pin — it asserts the value, not just dtype and shape, and the float8 cell is the one that makes the probe-vs-allowlist distinction a decision rather than an accident.

PR #4565, opened 2026-09-15
fix(geometry): make TPS kernel finite in float16

### LINKED ISSUE #4557: get_tps_transform: _kernel_distance eps=1e-8 underflows to 0 in float16, so every TPS warp is NaN
`kornia.geometry.transform.thin_plate_spline._kernel_distance` evaluates `0.5 * d2 * (d2 + eps).log()` with a fixed `eps = 1e-8`. The smallest float16 subnormal is about `6e-8`, so `torch.tensor(1e-8, dtype=torch.float16)` is `0.0`. Every control point has zero squared distance to itself, so the kernel matrix diagonal becomes `0 * log(0) = 0 * -inf = NaN`, the solve in `get_tps_transform` propagates it, and every float16 thin-plate-spline warp is NaN. float32 and bfloat16 are unaffected (bfloat16 keeps float32's exponent range).

The consequence for the public API is that `RandomThinPlateSpline` returns an all-NaN image for any float16 input, on CPU and on MPS, for either `align_corners` value, even with `scale=0`, which is the identity warp.

## Repro

```python
import torch, kornia
from kornia.geometry.transform.thin_plate_spline import _kernel_distance

for dt in (torch.float16, torch.bfloat16, torch.float32):
    print(dt, _kernel_distance(torch.tensor([0.0, 0.5, 2.0], dtype=dt)), torch.tensor(1e-8, dtype=dt).item())

for dev in ("cpu", "mps"):
    for ac in (False, True):
        out = K.RandomThinPlateSpline(scale=0, align_corners=ac, p=1)(torch.ones(1, 1, 3, 3, dtype=torch.float16, device=dev))
        print(dev, "align_corners", ac, "finite" if torch.isfinite(out).all() else "NaN")
```

Observed with kornia 0.9.0rc1, torch 2.14.0, macOS arm64, `PYTORCH_ENABLE_MPS_FALLBACK` unset:

```
torch.float16 tensor([    nan, -0.1733,  0.6934], dtype=torch.float16) 0.0
torch.bfloat16 tensor([-0.0000, -0.1729,  0.6914], dtype=torch.bfloat16) 1.0011717677116394e-08
torch.float32 tensor([-0.0000, -0.1733,  0.6931]) 9.99999993922529e-09
cpu align_corners False NaN
cpu align_corners True NaN
mps align_corners False NaN
mps align_corners True NaN
```

## Where it is currently recorded

- The `get_tps_transform` docstring says "float16 currently produces NaN weights" without a tracking link.
- #4538 adds a `.. warning::` on `RandomThinPlateSpline` describing the same underflow; it should link this issue.
- `testing/half_precision_xfails/cpu_float16.txt` carries `tests/geometry/transform/test_thin_plate_spline.py::TestWarpImage::test_identity_warp_align_corners[cpu-float16-1]`, and #4153 lists that file with one float16 failure. Neither names the cause.

## Suggested fix

The bound is also the singular point, so a larger clamp or a dtype-aware `eps` only moves the problem. Substitute a safe argument into the expression that gets differentiated and take the value from the other arm of a `torch.where`, as `AGENTS.md` prescribes for `clamp`-at-the-singularity patterns:

```python
def _kernel_distance(squared_distances: torch.Tensor) -> torch.Tensor:
    safe = torch.where(squared_distances > 0, squared_distances, torch.ones_like(squared_distances))
    return torch.where(squared_distances > 0, 0.5 * squared_distances * safe.log(), torch.zeros_like(squared_distances))
```

This keeps `r^2 log r^2 -> 0` at `r = 0` exactly, in every dtype, and is version-independent by construction. It changes float32/float64 results by at most the `eps` term, which is well inside the existing tolerances. After the fix, delete the manifest line above, drop the `float16` skip in the batch 6b convention tests, and remove the two prose notes.

Hrithikcrick:
Fixes #4557.

Fixes the float16 NaN failure in the thin-plate-spline kernel.

`_kernel_distance` previously evaluated `0.5 * d2 * log(d2 + 1e-8)`. In float16, `1e-8` rounds to zero, causing zero squared distances to evaluate `0 * log(0)` and produce NaN.

The kernel now uses a safe positive value for the logarithm at zero and explicitly returns zero for that branch, matching the mathematical limit of `r^2 log(r)` as `r -> 0`.

The change also removes the stale float16 known-failure entry, TPS convention-test skip, and documentation warning describing the resolved issue.

## What did I check?

- `tests/geometry/transform/test_thin_plate_spline.py` on CPU float16: 35 passed, 6 skipped
- `tests/geometry/transform/test_thin_plate_spline.py` on CPU float32/float64: 80 passed, 2 skipped
- TPS convention tests on CPU float16/float32: 8 passed
- pre-commit checks on all changed files
- `git diff --check`

## ONNX verification

Reviewer re-verification of `transform.warp_points_tps[at_centers]` using torch 2.14.0, onnx 1.22.0, onnxruntime 1.29.0, and onnxscript 0.7.1:

- main (`ecbcad561`): `mismatch` ? `out0: NaN pattern differs (torch 0, ort 10 NaNs)`
- PR merge tree: `ok` ? no eager/ORT mismatch; the exported graph includes `Where` and `Log`, and repeated runs were deterministic

This confirms that the zero-distance guard also resolves the previously recorded ONNX mismatch for this export-support case.


ducha-aiki (changes requested):
Review of `f99deafd`.

Verified the fix end to end on `30dfbf711` merged with the head (merge is clean, `ec45b2285`). #4557 reproduces on main and is fixed here; on the merged tree `test_conventions_geometric_dense.py` is 13 passed in each of float16, bfloat16, float32 and float64, the full TPS suite passes in float16 (35 passed, 6 skipped), and `grep -rnE "#4557|issues/4557" kornia/ tests/ docs/ testing/ README.md` finds nothing. Reverting only the production file inside the merged tree fails 3 nodes, so the forward fix is pinned.

**The implementation is the right one, and I am taking it over #3944's approach** (separate comment there). It is the `torch.where` safe-argument pattern rather than a `clamp` inside the differentiated expression, and I confirmed that at the singularity:

```
grad of _kernel_distance at d2=0:   main -9.210340e+00   head 0.000000e+00   (finite, fp32 and fp64)
fp16 inputs [0, 5.96e-08 (smallest subnormal), 1.19e-07, 6.10e-05, ...] -> all finite: main False, head True
```

Note the bound: float16's smallest subnormal is `5.96e-08`, not `finfo.tiny = 6.10e-05` (which is the smallest *normal*) — the head needs no eps at all, so neither bound applies to it. Dropping `+1e-8` also makes small distances strictly *more* accurate, since the eps used to dominate them: at `d2=1e-12`, main returns `-9.2103e-12` and the head `-1.3816e-11`, which is exact. Normal-path perturbation is safely inside tolerance: float32 max abs diff `1.49e-08` on kernel weights and `1.91e-06` on the warped image (rel `5.1e-7`) against `BaseTester`'s float32 `(1e-4, 1e-5)`; float64 `6.09e-13` (`2.53e-09` at zero-distance query points) against `(1e-5, 1e-5)`; bfloat16 bitwise identical. Manifest handling is correct: the deleted node XFAILs on main and PASSES here, it is the only TPS entry in either manifest, and a sweep of `tests/geometry/transform tests/augmentation` under both half profiles gives the same failure set as main with no new strict XPASS. TorchScript, `torch.export` and `torch.compile(fullgraph=True)` all work on the new kernel; `_kernel_distance` is private, absent from `__all__`, and no caller passes `eps=`, so dropping the parameter is safe.

Two blocking asks, then three small ones. Neither blocker disputes the implementation.

### [P2] Add a gradient pin

`safe = torch.where(...)` at `kornia/geometry/transform/thin_plate_spline.py:52` — the line that makes this version-independent rather than forward-only — is not pinned at all. Two one-token mutations survive the entire suite while turning the `d2 == 0` gradient into `nan`:

| mutation | float16 | float32 | grad at `d2=0` |
|---|---|---|---|
| `torch.ones_like` → `torch.zeros_like` | 48 passed | 52 passed | **nan** |
| `safe = squared_distances` | 48 passed | 52 passed | **nan** |
| swap the outer `where` arms | 32 failed | 36 failed | killed |
| `> 0` → `>= 0` | 9 failed | 12 failed | killed |

Both survivors leave the forward output at `0.0` because the outer `where` masks the bad arm, so no forward-only test can catch them — and the last two rows show the harness does report failures. One assertion that `_kernel_distance`'s gradient is finite for an input containing an exact `0.0`, in float32/float64, closes this.

Please target `_kernel_distance` directly rather than the public path. Through `get_tps_transform` the NaN is currently hidden on torch 2.14, because `_pair_square_euclidean`'s existing `clamp(min=0)` has derivative `0.0` at the bound there (I measured `0.0`); on the 2.5.1/2.9.1 leg that derivative is `1.0` and the NaN would propagate publicly. I could not run the 2.5.1 leg, so treat that half as unverified — which is exactly why the pin should not depend on it.

### [P2] `kornia/augmentation/_2d/geometric/thin_plate_spline.py:45` is now stale

The surviving `Convention:` bullet still says the zero-scale identity holds "with float32, float64, and bfloat16 inputs". That enumeration excluded float16 *because* of #4557, which is why the `.. warning::` you deleted sat directly beneath it. On the merged tree float16 satisfies the claim better than the bfloat16 already listed. Probe: `RandomThinPlateSpline(scale=0, align_corners=ac, p=1)` on `torch.arange(48., dtype=dt).reshape(1,1,6,8)/48`, reporting `(out - img).abs().max()`:

```
float16  align_corners=True   maxerr=0.000e+00   bitwise identical
float16  align_corners=False  maxerr=1.221e-04
bfloat16 align_corners=True   maxerr=1.953e-03   <- already listed
bfloat16 align_corners=False  maxerr=9.766e-04   <- already listed
```

float16 beats bfloat16 at both settings, so please add `float16` to the list. This is the last of #4557's four follow-ups, and the reason the `Fixes #4557` claim is not yet fully discharged.

### [P3] The changelog fragment begins with a UTF-8 BOM

```
$ xxd changelog.d/4565.fixed.md | head -1
00000000: efbb bf46 6978 2066 6c6f 6174 3136 204e  ...Fix float16 N
$ xxd changelog.d/4540.fixed.md | head -1
00000000: 6052 6573 ...                            `Res
```

It passes `check_changelog.py`, but it renders an invisible stray character at the start of the released bullet. Please strip it.

### [P3] Two leftover references to the removed eps guard

`docs/export_support/cases_geomA.py:268` and the generated `docs/source/_data/export_support.json:4906` both say `"... zero distance) -> exercises the eps guard in _kernel_distance"`. There is no eps guard any more.

### [P3] Minor

Dropping `require_tps` from `test_convention_dense_shape_and_matrix_interface` records nothing — that test passes even with the NaN kernel (`PASSED ...[cpu-float16-tps]` with the production file reverted). Fine as a re-enable, but an `isfinite` assertion would make it earn the un-skip.

Two things I checked that are *not* problems: the `get_tps_transform` bullet now ending `dtype- and backend-dependent` without punctuation matches the other bullets in that block, and `gradcheck` failing exactly at `d2=0` happens identically on main and on this head (inherent to the kernel's singularity, not a regression).

Separately, `docs/source/get-started/precision.rst:58` hard-codes `43 / 58` manifest entries for `kornia.geometry.transform` while the real counts are `42 / 54` on main and `41 / 54` here. That drift pre-dates this PR so it is not yours to fix, but this is the change that decrements the count — I will file a follow-up.


ducha-aiki (changes requested):
## Re-review — #4565, head `372cfeeb`

**Both of my P2s from the last round are fully discharged**, and the new gradient pin is a good one — it kills exactly the two mutations that survived at `f99deafd`. Merge onto `ecbcad561` is clean (`07b1d5de9`).

One blocker, and it is two lines.

## [P1] The new test hardcodes `float64` while taking the injected `device`, so it hard-fails on MPS

`tests/geometry/transform/test_thin_plate_spline.py:71-73`. The blocking `tests-mps / tests (3.13, 2.14.0)` leg is **red** on this head:

```
FAILED tests/geometry/transform/test_thin_plate_spline.py::TestTransformParameters::
  test_kernel_distance_zero_gradient[mps-grad_dtype1]
  - TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64
1 failed, 10522 passed
```

Reproduced locally: `1 failed, 1 passed` on the node, `1 failed, 44 passed, 10 skipped` on the whole file at `--device=mps --dtype=float32`.

The in-house idiom — 58 call sites of `device.type == "mps"` in `tests/`, e.g. `tests/geometry/test_linalg.py:102`, `test_keypoints.py:152`, `test_boxes.py:285`, `tests/enhance/test_zca.py:91`:

```python
def test_kernel_distance_zero_gradient(self, device, grad_dtype):
    if device.type == "mps" and grad_dtype == torch.float64:
        pytest.skip("MPS does not support float64")
    squared_distances = ...
```

Verified with that patch applied: MPS node `1 passed, 1 skipped`; MPS both files `44 passed, 11 skipped, 0 failed`; CPU float32/float64/float16 unchanged at `156 passed, 8 skipped`. The float32 leg still runs on MPS, so the pin keeps its device coverage rather than being skipped away. (Not pushed — yours to apply.)

## [P2] The export snapshot still says `mismatch` for a row this PR *fixes*

You updated the one cosmetic field in `docs/source/_data/export_support.json` and left the two the change actually falsifies. The `transform.warp_points_tps[at_centers]` record still carries:

```json
"onnx": "mismatch",
"onnx_detail": "runs, but the result differs from eager: out0: NaN pattern differs (torch 0, ort 10 NaNs)"
```

I ran the real probe on both trees with the same toolchain (torch 2.14.0 / onnx 1.22.0 / onnxruntime 1.29.0 / onnxscript 0.7.1):

```
main (ecbcad561): status=mismatch  error="out0: NaN pattern differs (torch 0, ort 10 NaNs)"   <- reproduces the committed string verbatim
head (merged):    status=ok        error=None      (ops include Where + Log; repeated twice, deterministic)
```

So the `torch.where` guard removes the eager/ORT divergence — this PR **fixes an ONNX mismatch** and the rendered docs page still shows the row with a `mismatch` badge (1 of 16 in the snapshot). Hand-editing those fields when a code change falsifies them is established precedent: `da4e565c8` (#4292) did exactly that for three fields. A full regeneration is ~2h and the merge refuses mixed-revision groups, so a targeted edit is the practical route. Set `"onnx": "ok"` and `"onnx_detail": ""` in that record, and put the A/B measurement in the PR body — it is a nice result and it should not be invisible.

## [P3] Two `_kernel_distance` mutations still survive — both pre-existing

Across `test_thin_plate_spline.py` + `test_conventions_geometric_dense.py` + `test_parameter_constants.py` + `test_augmentation_sequential.py` (baseline `181 passed, 54 skipped`): `0.5 * …` → `1.0 * …` survives, and `torch.zeros_like` → `torch.ones_like` in the false arm survives — the latter changes the kernel value at `r=0` from `0` to `1`. The scale mutation also survives on pristine main (`179 passed`), so nothing pins `_kernel_distance`'s *value*, and that is not your regression. But since you are already in this test class, one line closes it:

```python
assert_close(_kernel_distance(tensor([0., 1., e**2])), tensor([0., 0., e**2]))
```

## [P3] Carried over, not yours

`docs/source/get-started/precision.rst:58` hard-codes `43 / 58` manifest entries for `kornia.geometry.transform`; the real counts are `42 / 54` on main and `41 / 54` here. Pre-existing drift; this PR is only the decrement.

---

## Both previous asks: discharged

**My gradient-pin P2.** The two mutations that survived the whole suite at `f99deafd` are now killed by `test_kernel_distance_zero_gradient`:

| mutation | at `f99deafd` | at `372cfeeb` |
| --- | --- | --- |
| `torch.ones_like` → `torch.zeros_like` (safe arg) | SURVIVED | **2 failed** (`assert tensor(False)`, both dtypes) |
| `safe = squared_distances` | SURVIVED | **2 failed** |
| restore `0.5 * d2 * (d2 + 1e-8).log()` | — | **4 failed** |
| production file reverted to main | — | **3 failed** (conventions, 4 dtypes) |

Targeting the private `_kernel_distance` rather than the public path is the right call, as asked.

**My `thin_plate_spline.py:45` P2**, and the new docstring sentence is true by execution. `RandomThinPlateSpline(scale=0, p=1)`, max abs error vs input on the merge tree:

| dtype | `align_corners=True` | `False` |
| --- | --- | --- |
| float16 | **0.000e+00** (bitwise) | 1.221e-04 |
| bfloat16 | 1.953e-03 | 9.766e-04 |
| float32 | 5.96e-08 | 1.49e-08 |
| float64 | 0.0 | 2.78e-17 |

All finite, and float16 beats the already-listed bfloat16 at both `align_corners` settings — so the four-dtype enumeration holds.

**BOM stripped.** `head -c 16 changelog.d/4565.fixed.md | xxd` → `4669 7820 666c …` (`Fix fl…`), no `efbbbf`, matching `4568.fixed.md` / `4569.breaking.md`. Text unchanged and still true.

## Also verified

- `grep -rnE "#4557|issues/4557" kornia/ tests/` → no hits, and the same across `docs/ testing/ changelog.d/ benchmarks/`. No `eps` / `kernel epsilon` prose survives in either TPS file or `cases_geomA.py`.
- The `cases_geomA.py:268` hunk is correct and necessary — "exercises zero-distance handling" is accurate for the new code, and the JSON `note` edit mirrors it exactly as a regeneration would (`run_case` copies `note` straight from the case dict). No test or CI job validates the snapshot, so that edit is consistency-only.
- The `test_conventions_geometric_dense.py` un-skip now earns itself. With the production file reverted inside the merge tree: `test_convention_dense_shape_and_matrix_interface[cpu-float16-tps]` → `assert tensor(False)`; conventions file `3 failed, 49 passed` across 4 dtypes. `require_tps` is fully removed with no orphan callers.
- **Manifest handling is right.** `tests/geometry/transform tests/augmentation --device=cpu --dtype=float16 --xfail-known-failures --known-failure-profile=cpu-float16`: head `5 failed, 3647 passed, 3461 skipped, 272 xfailed, 5 xpassed`; main `5 failed, 3641 passed, 3464 skipped, 273 xfailed, 5 xpassed`. **Identical failure set and identical xpassed set** — the 5 failures (including the known macOS `random_apply3` strict-XPASS) are pre-existing, and the single manifest deletion is exactly accounted for (273→272).
- Counts: CPU `--dtype=float32,float64,float16` both files `156 passed, 8 skipped`; `--dtype=bfloat16` `53 passed, 3 skipped`. Pinned `ruff@0.16.7` clean, 5 files already formatted.

Fix the MPS skip and the two JSON fields and I will approve.

Hrithikcrick:
Thanks for the re-review. I’ve addressed the two remaining requested changes:

- added the MPS-specific skip for the float64 gradient case while keeping the float32 MPS coverage
- updated `transform.warp_points_tps[at_centers]` in the export-support snapshot from ONNX `mismatch` to `ok` and cleared the stale `onnx_detail`

I also added the ONNX A/B verification result to the PR description.

All CI checks are now passing (42/42).

Could you please take another look when you get a chance?

ducha-aiki:
## Re-review — head `bbb0cbac`

**Both blockers from `372cfeeb` are discharged**, and I re-measured each from scratch rather than taking the delta on trust. Merge onto `8c71df9a0` is clean (`bcf11f14d`). CI is 42/42 at this head, including `tests-mps / tests (3.13, 2.14.0)` (job `105072604285`, `success`, 12m20s).

One blocker remains, and it is one line. I graded it **P3 last round and said I would approve without it — that grade was wrong**, and a third surviving mutant I found this round makes the gap much bigger than I described. Apologies for the extra trip.

## [P2] `_kernel_distance`'s value is pinned by nothing — three one-token mutants survive

Test set: `tests/geometry/transform/test_thin_plate_spline.py` + `tests/augmentation/test_conventions_geometric_dense.py` + `tests/augmentation/test_parameter_constants.py` + `tests/augmentation/container/test_augmentation_sequential.py`, `--device=cpu --dtype=float32,float64,float16`. Baseline on the merge tree: **`11 failed, 466 passed, 164 skipped`** — the 11 are pre-existing float16 manifest entries in `test_augmentation_sequential`, identical on `main`.

| mutation of `_kernel_distance` | result | non-equivalence (float64) |
| --- | --- | --- |
| safe arg `torch.ones_like` -> `torch.zeros_like` | **13 failed** — killed | |
| `safe = squared_distances` | **13 failed** — killed | |
| swap the outer `where` arms | **120 failed** — killed | |
| production file reverted to `main` | **15 failed** — killed | |
| outer `> 0` -> `>= 0` | 11 failed — *equivalent* (value and gradient at `d2=0` unchanged; not a finding) | |
| **`0.5 *` -> `1.0 *`** | **11 failed — SURVIVES** | `get_tps_transform` kernel weights move by **1.21e-01** |
| **false arm `torch.zeros_like` -> `torch.ones_like`** | **11 failed — SURVIVES** | `warp_points_tps` at the kernel centers moves by **2.42e-01** |
| **`safe.log()` -> `safe.log1p()`** | **11 failed — SURVIVES** | warped image moves by **1.28e-01** on a `[0,1]` image — 1281x the suite's own `atol=1e-4`; free query points move `3.81e-02` |

Raw values, `_kernel_distance([0, 1, e^2])` in float64:

```
head       [0.0, 0.0,      7.389056]   <- correct: 0.5 * r^2 * log(r^2)
0.5->1.0   [0.0, 0.0,     14.778112]
zeros->ones[1.0, 0.0,      7.389056]
log->log1p [0.0, 0.346574,  7.857995]
```

**Why nothing catches them.** Every value pin in the TPS suite is invariant to the kernel function phi:

- `test_no_warp`, `test_affine_only`, `TestWarpImage::test_warp` (zoom by 2) and the `align_corners` convention tests are all **purely affine**, so the kernel weights are exactly zero and phi never reaches the result. Measured for `0.5 -> 1.0`: `max|d warp_image| = 1.3e-15` while `max|d kernel_weights| = 1.2e-01`.
- `TestWarpPoints::test_warp` checks the **interpolation property at the control points**, which holds for *any* phi by construction of the linear system.
- The two assertions this PR adds or changes are **finiteness only**: `assert torch.isfinite(grad).all()` and `assert torch.isfinite(output).all()`.

So no test anywhere evaluates the TPS kernel at a point that is neither a control point nor part of a purely affine map. The `torch.zeros_like` false arm is a line *this PR introduces*, and the value it produces — `0`, the limit of `r^2 log r` — is what makes the fix *correct* rather than merely *finite*; nothing checks it.

**One assertion closes it**, verified end to end rather than prescribed blind. Added to `TestTransformParameters` on the merge tree (needs `import math`):

```python
def test_kernel_distance_values(self, device, dtype):
    # 0.5 * r^2 * log(r^2): 0 at r^2 = 0 (the limit) and at r^2 = 1; e^2 at r^2 = e^2.
    d2 = torch.tensor([0.0, 1.0, math.e**2], device=device, dtype=dtype)
    expected = torch.tensor([0.0, 0.0, math.e**2], device=device, dtype=dtype)
    self.assert_close(_kernel_distance(d2), expected)
```

Measured with that test present:

```
head, --device=cpu --dtype=float32,float64,float16,bfloat16   9 passed   (exact, max abs diff 0.0 in all four)
head, --device=mps --dtype=float32                            6 passed
0.5 -> 1.0                    4 failed, 5 passed   (all four dtypes)
zeros_like -> ones_like       4 failed, 5 passed
log() -> log1p()              4 failed, 5 passed
production file = main        1 failed, 8 passed   (float16 only -> it doubles as the #4557 regression pin)
```

It takes the injected `dtype` fixture, so it does not repeat last round's MPS problem; `e^2` rounds exactly through the expression in float16 (`7.390625`) and bfloat16 (`7.375`), so there is no tolerance fiddling.

## Both previous asks: discharged

### [P1] MPS float64 — fixed

`test_thin_plate_spline.py:71-73` now guards with the in-house `device.type == "mps"` idiom. Verified with `env -u PYTORCH_ENABLE_MPS_FALLBACK`:

```
node,  --device=mps --dtype=float32   1 passed, 1 skipped   (SKIPPED: MPS does not support float64)
node,  --device=cpu --dtype=float32,float64,float16   2 passed
both files, --device=mps --dtype=float32   44 passed, 11 skipped, 1 deselected, 0 failed   (was 1 failed, 44 passed, 10 skipped)
both files, --device=cpu --dtype=float64   56 passed
```

The float32 leg still runs on MPS, so the pin keeps its device coverage. CI agrees.

### [P2] Export snapshot — fixed, and re-measured rather than trusted

Ran the real harness (`harness.run_cases(..., only=["transform.warp_points_tps[at_centers]"])`) on both trees with torch 2.14.0 / onnx 1.22.0 / onnxruntime 1.29.0 / onnxscript 0.7.1:

```
base 8c71df9a0 : status=mismatch  error="out0: NaN pattern differs (torch 0, ort 10 NaNs)"
                 ops = [Add, Gather, Log, MatMul, Max, Mul, ReduceSum, Slice, Transpose, Unsqueeze]
merge tree     : status=ok        error=None
                 ops = [Add, Gather, Greater, Log, MatMul, Max, Mul, ReduceSum, Slice, Transpose, Unsqueeze, Where]
```

The base run reproduces the committed `onnx_detail` string **verbatim** under a newer toolchain than the snapshot header claims, which is good evidence the flip is the code change and not the toolchain. `merge._norm_onnx` maps `status="ok"` -> `("ok", "")`, and `where` stays `""` for a non-fail row, so the two edited fields are exactly what a regeneration would write. Snapshot mismatch count 16 -> 15; nothing in `docs/` hardcodes that count.

Blast radius on the snapshot: the other three TPS cases, re-run on the merge tree, keep their committed values — `get_tps_transform` still `export-fail` -> `fail` (no ONNX lowering for `linalg.solve`), `warp_points_tps` `ok`, `warp_image_tps` `ok`. The one-row edit is the complete set of snapshot changes this PR causes.

## Core claim, re-verified from scratch at `bbb0cbac`

- **float16 finiteness.** `_kernel_distance([0, 5.96e-08, 1.19e-07, 6.10e-05, 1e-3, 0.5, 1.0, 2.0])` in float16: base returns `nan` for the first element, head `0.0`. Over a 20 060-element sweep, float16 NaNs go **43 -> 0**.
- **Gradient at the singularity.** `d(_kernel_distance)/d(d2)` at `d2 = 0`: base `-9.210340` in float32 and float64, head `0.0`. This is `torch.where` safe substitution, not a `clamp`, so it is torch-floor independent by construction.
- **Public path.** `RandomThinPlateSpline(scale=0, p=1)` on a float16 image: base NaN for both `align_corners`; head finite, max abs error `0.000e+00` (`align_corners=True`, bitwise identical) and `1.221e-04` (`False`).
- **Not bit-identical in float32/float64, and that is the point.** Removing `+1e-8` changes values: float64 `20050/20060` kernel values differ (max abs `5.0e-09`), float32 `783/20060` (max `2.98e-08`), bfloat16 `37/20060`, float16 `0/20060` among the finite entries. Max *relative* change `0.60`, entirely in the `logspace` tail where the eps used to dominate — the head is the exact one there. On a random non-degenerate 12-point TPS, float32 outputs move base<->head by up to `1.217e-04` (image), `6.01e-05` (points), `4.01e-05` (kernel weights) — inside that configuration's own float32 error, measured against float64 as `7.293e-04` on base and `7.633e-04` on head (means `1.079e-05` vs `1.113e-05`). No accuracy regression; a different rounding path.
- **Manifest.** The deleted line names `TestWarpImage::test_identity_warp_align_corners[cpu-float16-1]`, which on `main` is `1 failed — Tensor-likes are not close!` and on the merge tree `1 passed`. It is the only TPS entry in any manifest. Both half legs are `success`.
- **Regression sweep**, `tests/geometry/transform tests/augmentation --device=cpu --dtype=float32`: head `3971 passed, 3420 skipped, 32 xfailed, 5 xpassed, 0 failed`; base `3969 / 3420 / 32 / 5 / 0`. Identical skip/xfail/xpass sets; the `+2` are the new gradient test's two dtypes.
- **Portability.** `torch.jit.script`, `torch.export.export` and `torch.compile(fullgraph=True)` all work on the new kernel. `_kernel_distance` has exactly two call sites, both in-module, neither passing `eps=`; private, absent from `__all__`.
- **Blast radius.** `grep -rnE "#4557|issues/4557" kornia/ tests/ docs/ testing/ changelog.d/ benchmarks/ README.md` -> no hits. No `eps guard` / `kernel epsilon` prose survives for TPS. The `cases_geomA.py` note and the JSON `note` match character for character.
- **Hygiene.** `changelog.d/4565.fixed.md` starts `4669 7820` — BOM gone — and `fixed` is the right type. Pinned `ruff@0.16.7`: all checks passed, 5 files already formatted. Doctests on both changed modules: 4 passed.

## [P3] Carried over, not yours

`docs/source/get-started/precision.rst:58` still hardcodes `43 / 58` manifest entries for `kornia.geometry.transform`; the real counts are `42 / 54` on `main` and `41 / 54` here. I said last round I would file a follow-up and have not yet — that is on me.

## [P3] Minor

- The fragment says "Fix float16 NaNs…", the headline, but dropping `+1e-8` also perturbs float32/float64 slightly (numbers above). A half-clause — "…and removes the epsilon bias from float32/float64 kernel values" — would stop a user being surprised by a `1e-4`-scale float32 diff after upgrading.
- Nothing in either TPS test file carries a `dynamo`/`compile` marker: `-k "dynamo or compile"` deselects all 56 nodes, so the green `dynamo` leg says nothing about this change. I checked `torch.compile(fullgraph=True)` by hand instead. Pre-existing, not yours.
- The ONNX section in the PR description renders `->` as `?` in two bullets (mojibake). Description only.

Add the value assertion and I will approve on sight; everything else here is verified and correct.

Hrithikcrick:
Thanks for the detailed re-review. I added the direct `_kernel_distance` value regression using the injected `device` and `dtype` fixtures.

It now pins:
- `d2 = 0` -> `0`
- `d2 = 1` -> `0`
- `d2 = e^2` -> `e^2`

This covers the kernel value itself in addition to the existing gradient/finiteness regression.

Local CPU checks across float16, bfloat16, float32, and float64 pass, along with pre-commit.

Thanks again for catching the remaining gap.

ducha-aiki (approved):
Re-review of `b635708892da6269424321bd4ea41f9c6f7866b5`.

The last open request is resolved. `test_kernel_distance_values` independently pins the TPS kernel at squared distances `0`, `1`, and `e**2`, including both zero-valued cases and the nonzero `e**2` result. It catches the previously surviving coefficient, zero-substitution, and `log1p` changes while complementing the existing singular-gradient regression.

Validation: the kernel-value and zero-gradient tests passed 6/6 on CPU across float16, bfloat16, float32, and float64. Hosted CI is 42/42 green, and GitHub reports the head mergeable. The earlier MPS and export-snapshot blockers remain resolved. I found no unresolved correctness issue.

